PR #959, opened 2016-09-29
RF: Attempt to shrink the install+get combo (thought experiment)

mih:
TL;DR: `get` obtains any (and any number of dataset components, files, directories, datasets); `install` obtains ONE dataset from a (remote) source.

Size-wise this is not much less code than before. However, I believe the get implementation is cleaner  now. In particular the separation of clone and annexget operations.

More could be done re helper functions shared by install and get. There is potential for RF into Dataset() and some shrinkage. But not in this PR!

Finished (?) TODO:

- [x] bring `get` to (practically) 100% coverage
- [x] bring `install` to (practically) 100% coverage
- [x] consider changing `--fulfill` into a binary flag `--fulfill-datasets` that needs to be set when recursive `get` should also fulfill matching file handles that pop up in the form of new subdatasets that get installed along the way. I believe something like this is necessary to be able to express: give me, recursively, content for all file handles that are already there, but not for anything new
- [x] documentation needs heavy pass-over
- [x] unify naming or make it clear what is the difference since to yoh `--get-data` and `--fulfill` seems to accomplish the same
- [x] answer to Yarik how in this scheme I would accomplish what he asked in https://github.com/datalad/datalad/pull/959#issuecomment-250915262 -- "fulfill/get"  data for all datasets (including sub-datasets) which are already installed (or pointed to directly) while not installing any new dataset
- [ ] converge on [API we all like with hopefully minimal amount of API breakage](https://docs.google.com/spreadsheets/d/1bi3AQK5yMKwTA_6FN1oCPnIGASy9VmN8LbARr15XMKE/edit?usp=sharing)

PR #990, opened 2016-10-11
TST+BRK: How super is resolved with symlinks, needs resolution

mih:
Stumbled upon this question when looking for holes in the tests. Please share your thoughts.

### DIFF
--- a/datalad/distribution/dataset.py
+++ b/datalad/distribution/dataset.py
@@ -344,13 +344,16 @@ class Dataset(object):
         else:
             return was_once_installed
 
-    def get_superdataset(self, topmost=False):
+    def get_superdataset(self, topmost=False, stringent=True):
         """Get the dataset's superdataset
 
         Parameters
         ----------
         topmost : bool, optional
           Return the topmost super-dataset. Might then be the current one.
+        stringent : bool, optional
+          Be stringent in testing -- if superdataset is not containing sub-dataset
+          we came from as a submodule, it is not its super
 
         Returns
         -------
@@ -362,7 +365,7 @@ class Dataset(object):
         path = self.path
         sds_path = path if topmost else None
         while path:
-            par_path = opj(path, pardir)
+            par_path = normpath(opj(path, pardir))
             sds_path_ = GitRepo.get_toppath(par_path)
             if sds_path_ is None:
                 # no more parents, use previous found
@@ -372,6 +375,12 @@ class Dataset(object):
             # sds = Dataset(sds_path_)
             # if not sds.id:
             #     break
+
+            if stringent:
+                # verify that we contain this sub-dataset
+                submodules = GitRepo(sds_path_).get_submodules(sorted_=False)
+                if relpath(path, sds_path_) not in [s.path for s in submodules]:
+                    break
 
             # That was a good candidate
             sds_path = sds_path_
@@ -384,11 +393,14 @@ class Dataset(object):
             # None was found
             return None
 
-        if realpath(self.path) != self.path:
+        # and it is a subdirectory, not some fancy symlink somewhere outside
+        if realpath(self.path) != self.path and \
+            realpath(self.path).startswith(with_pathsep(realpath(sds_path))):
             # we had symlinks in the path but sds_path would have not
             # so let's get "symlinked" version of the superdataset path
             sds_relpath = relpath(sds_path, realpath(self.path))
             sds_path = normpath(opj(self.path, sds_relpath))
+
         return Dataset(sds_path)
 
     def get_containing_subdataset(self, path, recursion_limit=None):
--- a/datalad/distribution/tests/test_dataset.py
+++ b/datalad/distribution/tests/test_dataset.py
@@ -12,6 +12,8 @@
 import os
 import shutil
 from os.path import join as opj, abspath, normpath
+from os.path import realpath
+from os.path import pardir
 
 from ..dataset import Dataset, EnsureDataset, resolve_path, require_dataset
 from datalad.api import create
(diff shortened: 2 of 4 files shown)


yarikoptic:
It is different so that it is a virtual Yarik adding the label. I was still sleeping then.
I think it might be the waffle.io I was trying out... Might need to disable now

bpoldrack:
Hm. I think that means, that `GitRepo.get_toppath` should not resolve symlinks. But I'm currently not sure, whether this approach will fire back in a slightly different setup using symlinks. Any thought on this approach?

yarikoptic:
`GitRepo.get_toppath` -- I think I have "fixed" it to try to figure out the one which would be based on PWD ... in ad55e45ce2fdfcd609103015d781a776676dd325  which is a part of Michael's #959 I believe

mih:
But the question is: what do we want it to do? The subdataset in question is not actually underneath the super dataset.

yarikoptic:
but what is a super-dataset of symlinked dataset? e.g. if I have

    ./super1/ds1
    ./super2/ds1 -> ../super1/ds1

and I am under ./super2/ds1 -- which one should be considered a superdataset?

if super1 -- we need to follow resolved path, if super2 -- we should follow non resolved (similarly to how I did for get_toppath).

bpoldrack:
Well, I think if you are under `./super2/ds1` it should return `super2` and if you are under `./super1/ds1` it should retunr `super1`. What would be a situation where you possibly want a different result?

bpoldrack:
And: As @mih pointed out in line 131 of `test_dataset.py` it yields `None` currently. (Not sure, whether this is still the case with your fix). And that means, `get_superdataset` resolved the symlink, went up and found no dataset anymore, therefore returning `None`.

yarikoptic:
> Well, I think if you are under ./super2/ds1 it should return super2 and if you are under ./super1/ds1 it > should retunr super1. What would be a situation where you possibly want a different result?

;-) we are organizing super/sub datasets now using submodules mechanism, correct? ;)  so, if we stay true to our concept, super2 is not is not a superdataset of ds1 since there it is merely a symlink to somewhere, where it is (potentially) a proper sub-dataset.  super2 has no clue how to `get/install` ds1 (no submodule) -- it just kinda comes available if it is available in the original symlinked location.

mih:
Are we still talking about the same thing? In my test there is a proper super/sub relationship -- that is ignored. I think this is a problem.

mih:
And IMHO this is one case of a more general problem where we assume dataset hierarchy to be reflected on the filesystem. We don't have the same problem looking down, because we can ask the repo for it's submodules. But upwards we make inappropriate shortcuts. I think.

yarikoptic:
> In my test there is a proper super/sub relationship
I will read the code later whenever have spare time... once again -- if only PR description had a nice description so code reading was not necessary to understand what the PR is about...  ;)

> And IMHO this is one case of a more general problem where we assume dataset hierarchy to be reflected on the filesystem. We don't have the same problem looking down, because we can ask the repo for it's submodules. But upwards we make inappropriate shortcuts. I think.

They are appropriate under assumption of a hierarchical layout.  If it is not hierarchical -- then we are indeed in trouble, and need to store the path of the super dataset e.g. within `.git/config` so we could just look up in the config.  That should then work for both (I guess).

bpoldrack:
> super2 has no clue how to get/install ds1 (no submodule)
I think Michael is right - this is not true in his setup.

Thinking about it I realized this is indeed a more general issue. We probably don't even need to ask for it's superdataset to get in trouble, but just instantiate `Dataset` both ways. `Dataset` itself resolves symlinks and this would lead to forget where we came from, meaning both objects would be the same. Originally this was intentional but with a usecase like this, it's troublesome.

Would you guys generally agree to change `Dataset` to not resolve symlinks when instantiated and just use `normpath` instead of `realpath` in the constructor? That way we would generally be able to reflect a situation like this. That would make it harder to figure out, whether two datasets are equal (representing the same repository), but I think we hardly need such a comparison outside test code.
Any fears regarding this?

mih:
I have no fear. Someone needs to be brave enough to make that change, discover that the tests are not complete, open a PR, and face the echo. I can do that, unless somebody beats me too it ;-)

yarikoptic:
@bpoldrack  since we agreed awhile back we indeed should avoid `realpath` (and also `abspath` which some times leads to resolving links)

yarikoptic:
I think that all the symlinks will put us into grave or at least cause lots of problems... and having a submodule which is also actually just a symlink outside somewhere (now that I think I have read the test correctly) is one of those "interesting" cases I would say.  I have pushed into this PR few commits -- cherry picked one I was talking about and additional where I have introduced check for super to have original one as a submodule, and then also what I thought could be correct additional condition for post-treatment of the deduced super's path... but that broke some other test ... feel free to fix more if you see clearly what is happening or scrape my commits.

But overall -- is this a use-case for flat hierarchies or ???

mih:
This was not to promote a use case, but came from a real-life observation. I produced such a configuration for an analysis repo, because `annex init` was taking so long that I simply symlinked the source. Which worked fine for almost anything, but led to the discovery that the present behavior is unexpected and untested.

yarikoptic:
If we are to allow such use-case, we should may be just somehow directly support it in `install` with e.g. `--beyond-reckless` ;)

datalad/distribution/dataset.py
```diff
@@ -384,11 +393,14 @@ def get_superdataset(self, topmost=False):
             # None was found
             return None
 
-        if realpath(self.path) != self.path:
+        # and it is a subdirectory, not some fancy symlink somewhere outside
+        if realpath(self.path) != self.path and \
+            realpath(self.path).startswith(with_pathsep(realpath(sds_path))):
```
yarikoptic: FTR:  in https://github.com/datalad/datalad/pull/1011/files#diff-3947667888c8ff9d4bfc6c26f2be70f9L387 I overhauled this a bit -- apparently I didn't need this check here when `normpath` is used.

mih:
Do you guys come to any conclusion on this one?

mih:
Conclusion: no conclusion.

PR #995, opened 2016-10-12
Addressing direct mode

### LINKED ISSUE #982: Central switch to engage direct mode
We should have a config switch to engage direct mode for all installed and created repositories.

Once we have it, there should be a travis test run that exercises the entire battery in direct mode.

This should give us an initial idea how far we got off the track already and to make sure we dont diverge any further.

@bpoldrack Do you want to give that a try?

bpoldrack:
General idea:
In order to solve direct mode issues provide a "central switch" (see #982) to let datalad create annex repos in direct mode. This is done by a new config item `datalad.repo.direct`, that can be set via env variable or config files. Then have a dedicated matrix entry for Travis, that makes use of this switch. This means, that - for now - this build will fail! But: At least we can make sure to not add even more failures by PRs to come and fix the issue over time. ~~So, failing of this travis build should NOT prevent this PR from being merged.~~
Changed: We have now 3 extra travis builds (one per python version), that are running in direct mode, but are allowed to fail. So travis doesn't consider this to be an overall failure.
Fixing these issues is a lot of things and won't be done in a single PR. Depending on the failures, there are different reasons like:
- GitPython's git calls, that can't be easily modified yet whenever we are in a direct mode repo
- Not using `AnnexRepo` to invoke a git call, but either `Runner` or just `GitRepo` (ignoring the fact, that it actually is an annex)
- wrong assertions in test code or test code, that was correct but isn't anymore, since new "central switch" works slightly different
- broken commit refs comparisons, since `origin/master` becomes `origin/annex/direct/master` for example
- ...

Furthermore, also introduced a config item `datalad.repo.version` to set a default for annex repository version. This also leads to realizing, that we should go for version 6 by default eventually. But this is kind of different story and doesn't mean we don't have to care for direct mode, I think. Especially given, that one way or another, we could have to deal with version 5 (and therefore direct mode) nevertheless.
- [x] Closes #982 
- [x] Add a direct mode matrix entry for Travis
- [x] fix AutoMagicIO (two tests are somehow broken by this PR even when not using direct mode switch)
- [x] some basic fixes for direct mode
- [x] make GitPython use _GIT_COMMON_OPTIONS of `GitRepo` class and therefore deal with direct mode
- [x] fix python 3

Final note: This PR does not solve all the issues, but provides the mechanisms to use when solving them. Whenever git is called correctly via an `AnnexRepo` instance it should be able to deal with direct mode now, even when GitPython is used. However, not even all tests and methods of `AnnexRepo` and `GitRepo` are properly adapted by now. But hopefully there are all the means to not make it worse during further development of datalad and to solve the issues, that are already there within PR(s) to come.


bpoldrack:
@yarikoptic : What's the 'in progress' label for?

bpoldrack:
Since everything appears to fail atm, I want to point out:
The only thing currently failing OUTSIDE the direct mode build is a test of the `AutoMagicIO`. Not clear to me yet, what is the issue.

bpoldrack:
@yarikoptic : The failure in `test_auto.py` appears to happen when `git config -z -l` is called, while you patched `sys.stdout` and `sys.stderr`, if I get it right. This probably happens, since `AnnexRepo.__init__` calls `ConfigManager` now. The point is: Apparently line 129 in `test_auto.py` (`verify_load( ...`) leads to a (probably unneccessary) instantiation of `AnnexRepo`. But I don't get all the mocking, patching and redirecting entirely atm, so may be you can have look?

yarikoptic:
> So, failing of this travis build should NOT prevent this PR from being merged.

It should -- we don't want to have master failing. BUT we can allow it to fail, see e.g. http://stackoverflow.com/questions/20905969/travisci-how-to-allow-failures-for-environment-variable

so we would always run it but it would not trigger the entire build to be considered 'failing'.  Then we could merge it any time when nothing else gets broken

yarikoptic:
looking into that auto test failing test issue now.  note that most probably it would be the same issue if ran under ipython (that is what that replacement of out/err is there to mimic)

yarikoptic:
This would workaround it then we would overload stderr as well when running, so not the crippled stderr which that tests sets up would be used, and we would not "trip the wire" BUT I am not sure yet if that is the desired behavior

```shell
$> git diff datalad/config.py
diff --git a/datalad/config.py b/datalad/config.py
index 4c9ab9b..4dbb7c5 100644
--- a/datalad/config.py
+++ b/datalad/config.py
@@ -155,7 +155,7 @@ class ConfigManager(object):
                     stdout, self._store, replace=False)
 
         if not self._dataset_only:
-            stdout, stderr = self._run(['-z', '-l'], log_stderr=False)
+            stdout, stderr = self._run(['-z', '-l'], log_stderr=True)
             self._store = _parse_gitconfig_dump(
                 stdout, self._store, replace=True)
```

bpoldrack:
Thanks!

datalad/support/gitrepo.py
```diff
@@ -70,7 +70,32 @@
...
+        self._git_persistent_options = []
+        self._git_temp_options = ()
+
+    def _git_options_get(self):
+        return self._git_persistent_options + [i for i in self._git_temp_options]
+
+    def _git_options_set(self, val):
+        self._git_temp_options = val
+
+    _git_options = property(_git_options_get, _git_options_set)
```
bpoldrack: Sadly, it doesn't work that way, because this leads to modifying `_git_options` only if there were any git options passed at all. Whenever the original class (`Git`) creates a call without any git options passed, `__call__` will not be called and therefore `_git_persistent_options` are ignored. We need  `_git_options` to return something different, not being set different.

yarikoptic:
Just a note : while working on this one targetting direct mode, keep in mind that sooner (than later) we will probably want to do the same for 'annex version 6' repositories, which Joey recommends as an ultimate replacement for direct mode.  Some bugs got fixed by joey yesterday, so hopefully soon I will provide updated snapshot packages (packaging needs some tuning atm)

bpoldrack:
Yes, I'm aware.

bpoldrack:
GitPython released a new version, including my patch. Therefore 2.1.0 is now a requirement.
Ready to merge from my point of view.

datalad/tests/test_utils_testrepos.py
```diff
@@ -33,12 +34,10 @@ def _test_BasicAnnexTestRepo(repodir):
     ok_file_under_git(trepo.path, 'test.dat')
     ok_file_under_git(trepo.path, 'INFO.txt')
     ok_file_under_git(trepo.path, 'test-annex.dat', annexed=True)
-    if not trepo.repo.is_crippled_fs():
-        ok_broken_symlink(pathjoin(trepo.path, 'test-annex.dat'))
+    ok_(trepo.repo.file_has_content('test-annex.dat') is False)
```
yarikoptic: I have got surprised that this line marked not covered... well -- for me it just fails altogether:

```shell
(git)hopa:~datalad/datalad[nf-direct-mode]git
$> git describe
0.4-80-gec22586
changes on filesystem:                                                                                          
 datalad/tests/test_utils_testrepos.py | 1 +
(venv-tests) 2 16182.....................................:Fri 28 Oct 2016 08:16:26 PM EDT:.
(git)hopa:~datalad/datalad[nf-direct-mode]git
$> nosetests datalad/tests/test_utils_testrepos.py
E.E
ERROR: datalad.tests.test_utils_testrepos.test_BasicAnnexTestRepo_random_location_generated
    self.test(*self.arg)
  File "/home/yoh/proj/datalad/datalad/datalad/tests/test_utils_testrepos.py", line 47, in test_BasicAnnexTestRepo_random_location_generated
    _test_BasicAnnexTestRepo(None)  # without explicit path -- must be generated
  File "/home/yoh/proj/datalad/datalad/datalad/tests/test_utils_testrepos.py", line 28, in _test_BasicAnnexTestRepo
    trepo = BasicAnnexTestRepo(repodir)
  File "/home/yoh/proj/datalad/datalad/datalad/tests/utils_testrepos.py", line 44, in __init__
    self.repo = self.REPO_CLASS(path)
  File "/home/yoh/proj/datalad/datalad/datalad/support/annexrepo.py", line 155, in __init__
    create=create, git_opts=git_opts)
  File "/home/yoh/proj/datalad/datalad/datalad/support/gitrepo.py", line 492, in __init__
    self.repo.git._persistent_git_options = self._GIT_COMMON_OPTIONS
AttributeError: 'Git' object has no attribute '_persistent_git_options'

```

yarikoptic: doh -- nevermind -- I still had GitPython 2.0.9

yarikoptic:
smaug  crashed again, restarted dockers, but it seems that something 'has changed', so failures might be unrelated to this PR changes... yet to pin down

mih:
Cool, thanks!

Just skimmed through the failures -- looks like the winter won't get boring....

mih:
Needs a bit of conflict resolution, and then it can go straight in IMHO.

yarikoptic:
FWIW just pushed updated smmap and python-git (2.1.0) modules to Debian and NeuroDebian so we should be all good on that end indeed now

Issue #997, opened 2016-10-13
datalad search tries to create a cache in a local (has nothing to do with annex/datalad) repo


yarikoptic:
```shell
(git)hopa:~datalad/datalad[master]git
$> datalad search bids
2016-10-13 00:40:57,565 [INFO   ] Loading and caching local meta-data... might take a few seconds 
Traceback (most recent call last):
  File "/home/yoh/proj/datalad/datalad/venv-tests/bin/datalad", line 11, in <module>
    load_entry_point('datalad', 'console_scripts', 'datalad')()
  File "/home/yoh/proj/datalad/datalad/datalad/cmdline/main.py", line 282, in main
    cmdlineargs.result_renderer(ret, cmdlineargs)
  File "/home/yoh/proj/datalad/datalad/datalad/metadata/search.py", line 361, in result_renderer_cmdline
    for location, r in res:
  File "/home/yoh/proj/datalad/datalad/datalad/metadata/search.py", line 201, in __call__
    os.makedirs(cache_dir)
  File "/home/yoh/proj/datalad/datalad/venv-tests/lib/python2.7/os.py", line 157, in makedirs
    mkdir(name, mode)
OSError: [Errno 20] Not a directory: '/home/yoh/proj/datalad/datalad/.git/datalad/cache'
```
IMHO if a current directory/repo is not a datalad dataset, shouldnt try to search within it since it wouldn't find anything -- should get to the logic asking to install central repo, or just bail
PS in above it fails since I do have .git/datalad which is readonly/root-owned (for reasons to catch such cases)

mih:
Without this permission issue it still doesn't work:

````
datalad/git (git)-[master] % datalad search bids
2016-10-14 08:39:02,602 [INFO   ] Loading and caching local meta-data... might take a few seconds 
Traceback (most recent call last):
  File "/home/mih/hacking/datalad/git/bin/datalad", line 9, in <module>
    load_entry_point('datalad', 'console_scripts', 'datalad')()
  File "/home/mih/hacking/datalad/git/datalad/cmdline/main.py", line 282, in main
    cmdlineargs.result_renderer(ret, cmdlineargs)
  File "/home/mih/hacking/datalad/git/datalad/metadata/search.py", line 361, in result_renderer_cmdline
    for location, r in res:
  File "/home/mih/hacking/datalad/git/datalad/metadata/search.py", line 204, in __call__
    ignore_cache=False)
[...]

Issue #986, opened 2016-10-08
RFC: do not autogenerate a Dataset id if not defined. Generate one during create (only)


yarikoptic:
Currently I can do in any directory:

```shell
hopa:/tmp
$> python -c 'from datalad.distribution.dataset import Dataset; print(Dataset(".").id)'
54e5a0aa-8d05-11e6-95c9-8019340ce7f2
```
which imho makes little sense... and in general -- I thought we could discriminate between a Dataset and non-Dataset by checking for their id, and if None - consider it to not correspond to a dataset

edit 1: well -- ATM the only way to initiate a dataset in existing git/annex repo seems to use `create --force` but not sure if what `--force` does doesn't have side-effects... seems it doesn't and just indeed does only initialization of the dataset if git or annex repo is already there.  May be we should add explicit `--init-only` which, unlike `--force` wouldn't try to create one?

bpoldrack:
I'm not sure, what implications it would have for metadata if there wasn't an `id` in this case. May be it makes sense to have `None` here. @mih?

But as far as it regards this aspect:
> I thought we could discriminate between a Dataset and non-Dataset by checking for their id, and if 
> None - consider it to not correspond to a dataset

How about `Dataset.is_installed()` or `Dataset.repo is None`? Why would you need a third method of figuring out, whether or not there is a dataset? Or do I misunderstand the purpose of your wanted check via `Dataset.id`?

mih:
Implication would be no metadata unless there is an ID. I think that was
merged already. Or at least there is an issue with this resolution.

On Oct 10, 2016 13:14, "Benjamin Poldrack" <notifications@github.com> wrote:

I'm not sure, what implications it would have for metadata if there wasn't
an id in this case. May be it makes sense to have None here. @mih
<https://github.com/mih>?

But as far as it regards this aspect:

I thought we could discriminate between a Dataset and non-Dataset by
checking for their id, and if
None - consider it to not correspond to a dataset

How about Dataset.is_installed() or Dataset.repo is None? Why would you
need a third method of figuring out, whether or not there is a dataset? Or
do I misunderstand the purpose of your wanted check via Dataset.id?

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/datalad/datalad/issues/986#issuecomment-252589711>, or mute
the thread

Issue #949, opened 2016-09-28
[wishlist] do not initiate dataset (i.e. add .datalad/config) in aggregate


yarikoptic:
leave initiation of a dataset solely to Dataset.create, and in aggregate just skip (with a warning) those without id yet.

mih:
#965 was merged (thanks!), but the issue was not addressed in full. I think `get_metadata()` also needs to ignore such datasets.

yarikoptic:
ok... indeed probably it should

mih:
With the merge of #1050 this is now fully addressed.

PR #987, opened 2016-10-08
ENH+BF: save --super-datasets, save subds, save [files] avoiding saving index etc

yarikoptic:
Closes #472

PR #1010, opened 2016-10-14
Dataset ID and metadata handling #986 #949 #997

mih:
ATTENTION: This should not be touch until after #1050 is in!

Looking into this I found that we have to decide how to deal with plain repositories. I was under the impression that we are moving towards caring for datalad datasets, not any repository. But given the recent misconceptions:

What about the case where a dataset, contains a repository submodule (no dataset), which in turn contains a dataset.

IMHO the only clean way is to ignore anything from the submodule downward (including the submodule). If we ignore it in the metadata, we should likely ignore it in things like `get_subdataset()`, to avoid proliferation of dataset flavors. If we don't want to ignore it, we have to be able to describe it, hence plain repos will keep getthing an ID ... IWO things stay the way they are and issues  like #986 are invalid.

If we want to ignore it, how can we tell a subdataset from a submodule without having it installed?

Likewise, what if I have a plain repository that contains submodules which are dataset? While the top is not a dataset? Do we have to consider it to be one nevertheless, because it would make sense to search in it? In this case the resolution of #997 becomes more complicated than it seemed at first.

Decision please.

### DIFF
--- a/datalad/metadata/search.py
+++ b/datalad/metadata/search.py
@@ -125,7 +125,11 @@ class Search(Interface):
         try:
             ds = require_dataset(dataset, check_installed=True, purpose='dataset search')
             if ds.id is None:
-                raise NoDatasetArgumentFound
+                raise NoDatasetArgumentFound(
+                    "There seems to be a repository which lacks a DataLad's "
+                    "dataset id, thus not considered to be a dataset. "
+                    "You can use 'datalad create --force %s' command to "
+                    "'initiate' that repository as a DataLad dataset" % ds.path)
         except NoDatasetArgumentFound:
             exc_info = sys.exc_info()
             if dataset is None:
--- a/datalad/metadata/search.py
+++ b/datalad/metadata/search.py
@@ -124,6 +124,8 @@ class Search(Interface):
                   match, dataset)
         try:
             ds = require_dataset(dataset, check_installed=True, purpose='dataset search')
+            if ds.id is None:
+                raise NoDatasetArgumentFound
         except NoDatasetArgumentFound:
             exc_info = sys.exc_info()
             if dataset is None:
--- a/datalad/distribution/create.py
+++ b/datalad/distribution/create.py
@@ -11,6 +11,7 @@
 """
 
 import logging
+import uuid
 
 from os import listdir
 from os.path import isdir, realpath, relpath
@@ -236,13 +237,16 @@ class Create(Interface):
             for nt in native_metadata_type:
                 tbds.config.add('datalad.metadata.nativetype', nt)
 
-        # record the ID of this repo for the afterlife
+        # record an ID for this repo for the afterlife
         # to be able to track siblings and children
         id_var = 'datalad.dataset.id'
         if id_var in tbds.config:
             # make sure we reset this variable completely, in case of a re-create
             tbds.config.unset(id_var, where='dataset')
-        tbds.config.add(id_var, tbds.id, where='dataset')
+        tbds.config.add(
+            id_var,
+            tbds.id if tbds.id is not None else uuid.uuid1().urn.split(':')[-1],
+            where='dataset')
 
         # save everthing
         tbds.repo.add('.datalad', git=True)
--- a/datalad/distribution/dataset.py
+++ b/datalad/distribution/dataset.py
@@ -10,7 +10,6 @@
 """
 
 import logging
-import uuid
 from functools import wraps
 from os.path import abspath
 from os.path import commonprefix
@@ -151,15 +150,11 @@ class Dataset(object):
         Returns
         -------
         str
-          This is either a stored UUID, or if there is none: the UUID of the
-          dataset's annex, or a new generated UUID.
+          This is either a stored UUID, or `None`.
         """
         if self._id is None:
             # if we have one on record, stick to it!
             self._id = self.config.get('datalad.dataset.id', None)
-            if self._id is None:
-                # fall back on self-made ID
-                self._id = uuid.uuid1().urn.split(':')[-1]
         return self._id
 
     @property
--- a/datalad/distribution/tests/test_dataset.py
+++ b/datalad/distribution/tests/test_dataset.py
@@ -298,17 +298,19 @@ def test_require_dataset(path):
 @with_tempfile(mkdir=True)
 def test_dataset_id(path):
     ds = Dataset(path)
+    assert_equal(ds.id, None)
+    ds.create()
     dsorigid = ds.id
     # ID is always a UUID
     assert_equal(ds.id.count('-'), 4)
     assert_equal(len(ds.id), 36)
-    # creating a new object for the same path (no ID on record)
-    # does not yields the same ID
+    # creating a new object for the same path
+    # yields the same ID
     newds = Dataset(path)
     assert_false(ds is newds)
-    assert_true(ds.id != newds.id)
+    assert_equal(ds.id, newds.id)
     # recreating the dataset does NOT change the id
-    ds.create(no_annex=True)
+    ds.create(no_annex=True, force=True)
     assert_equal(ds.id, dsorigid)
     # even adding an annex doesn't
     ds.create(force=True)
--- a/datalad/metadata/__init__.py
+++ b/datalad/metadata/__init__.py
@@ -72,13 +72,15 @@ def _get_base_dataset_metadata(ds_identifier):
     """Return base metadata as dict for a given ds_identifier
     """
 
-    return {
+    meta = {
         "@context": "http://schema.org/",
-        "@id": ds_identifier,
         "type": "Dataset",
         # increment when changes to meta data representation are done
         "dcterms:conformsTo": "http://docs.datalad.org/metadata.html#v0-1",
     }
+    if ds_identifier is not None:
+        meta["@id"] = ds_identifier
+    return meta
 
 
 def _get_implicit_metadata(ds, ds_identifier=None, subdatasets=None):
@@ -144,12 +146,11 @@ def _get_implicit_metadata(ds, ds_identifier=None, subdatasets=None):
     subdss = []
     # we only want immediate subdatasets
     for subds in subdatasets:
-        subds_id = subds.id
         submeta = {
             'location': relpath(subds.path, ds.path),
             'type': 'Dataset'}
-        if not subds_id.startswith('_:'):
-            submeta['@id'] = subds_id
+        if subds.id:
+            submeta['@id'] = subds.id
         subdss.append(submeta)
     if len(subdss):
         if len(subdss) == 1:
--- a/datalad/metadata/__init__.py
+++ b/datalad/metadata/__init__.py
@@ -146,6 +146,8 @@ def _get_implicit_metadata(ds, ds_identifier=None, subdatasets=None):
     subdss = []
     # we only want immediate subdatasets
     for subds in subdatasets:
+        if subds.id is None:
+            continue
         submeta = {
             'location': relpath(subds.path, ds.path),
             'type': 'Dataset'}
--- a/datalad/metadata/__init__.py
+++ b/datalad/metadata/__init__.py
@@ -261,7 +261,7 @@ def get_metadata(ds, guess_type=False, ignore_subdatasets=False,
     # for any subdataset that is actually registered (avoiding stale copies)
     for subds in subdss:
         subds_path = relpath(subds.path, ds.path)
-        if ignore_cache and subds.is_installed():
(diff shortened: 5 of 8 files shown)


datalad/metadata/search.py
```diff
@@ -124,6 +124,8 @@ def __call__(match,
                   match, dataset)
         try:
             ds = require_dataset(dataset, check_installed=True, purpose='dataset search')
+            if ds.id is None:
+                raise NoDatasetArgumentFound
```
yarikoptic: I would have added a message that repository is there but it is not a dataset since missing an ID

mih: Why? We decided to care about datasets and not repositories. Otherwise I should also acknowledge the existence of a directory. IMHO an appropriate message is generated few lines below. This particular exception never sees the light of day anyway.

yarikoptic:
pushed a "fix" to my comment. 

I would concur that probably it is better to not treat an arbitrary repository as a full-featured dataset, since it is simply not so.  But indeed it makes many things complicated, since we would theoretically fail when dealing with plain repositories which are not datasets  during installation, etc.  I wonder if we could consider every repository as a potential dataset, just not initialized (no id), and some operations (such as search) blow whenever it is not initialized?

oops... kids came -- need to run, will try to reply more later

yarikoptic:
I guess the cleanest way would be to consider a dataset without id (i.e. pure repository) to not be a dataset, so we have clean boundaries between "datasets" and their "super-repositories" and possible pure sub-modules (e.g. code/ of some kind, which in turn might have its own dataset as a submodule).  Otherwise I am afraid many things (primarily while going "upward" in hierarchy) would become complicated and/or broken.  E.g. if we include some dataset as a submodule to a regular git repo.  search -d^ wouldn't work correctly if we follow to the top (into that repository) and try to search from there (it has no meta-data).
Information about either it is a sub-dataset or sub-module is there -- in `datalad/meta`.  If it has a record (subdirectory with meta.json) for submodule -- it is a sub-dataset, otherwise -- a sub-module.  But during aggregation, we check all sub-modules for presence of id and meta-data to be aggregated.  
So we just need to make sure that meta-data is up-to-date.  And for that we need to aggregate meta during  `save` (https://github.com/datalad/datalad/issues/860), install, uninstall, and when we create a dataset (in both cases aggregating within, unconditionally in create and in save could be smarter) and then updating its aggregated information in super (if created or changed), we should have consistent idea on what is a dataset and what not, and all get_super/subdataset should stop at datasets boundaries.  So it would require additional work but IMHO doable.  FWIW, I would actually quite appreciate such automatic aggregation and update.  The only concern is that updating aggregated information in super-dataset, might be somewhat a "touchy" topic I guess?  But imho it is not too bad if only corresponding to subdataset of interest meta-data files get committed.  Alternative could be not even to update super's copy of meta-data automagically -- if created/saved as a sub-dataset, it will get some meta-data added at that point, since committing .gitmodules anyways.

`datalad ls -r` in turn could add an indicator either give submodule is a dataset (currently just says if it is git or annex repo).

some commands do not really care if what they operate on is a dataset (with id) or just a repo, e.g. install, uninstall.  So it seems to me that concern is primarily about commands which deal with meta-data, so they could rely on meta-data to decide.

Makes sense?

yarikoptic:
BTW, apparently in https://github.com/datalad/datalad/pull/987/files#diff-3947667888c8ff9d4bfc6c26f2be70f9L370 I had already to enable such check as well (that superdataset is actually a dataset with an id) so I could stop getting to the top.

mih:
Ok, complex reasoning that seems to make sense upon first read, but it is not clear to me what pratical consequences it might have. I think before I can continue we need consensus on this yes/no question:

Will `Dataset.get_subdataset()` results include instances for submodules that have no ID, hence are not datasets?

If no, when and how is it possible to decide if a submodule is a dataset? (You seem to suggest that it is a meta data based decision, which implies that a basic mechanism will now depends on up-to-date meta data -- I consider this a bold move!

So if the answer is no, that change needs to be implemented first. And it will result in a number of implied changes IMHO.

If the answer is yes, the name of this method needs to change, and every use in the code base needs to be reevaluated whether datasets or any submodule are valid output.

yarikoptic:
The answer was "no"

mih:
Great, now we need the answer to the "how is it possible" part. ;-)

yarikoptic:
Regarding"how" - I think you understood my suggestion correctly. It should be quite doable I think. Just needs data aggregation bring triggered by commands making or removing datasets...I see few possible gotchas but it might work

mih:
Cool. So let me put that in other words and continue the argument:

With such a change, we are no longer using the submodule functionality to register subdatasets. If that is the case, why should we keep using submodules to locate submodules, or store info on from where to download them?

My point: If we make this change, it should be accompanied by and end of any submodule dependency in datalad, because it becomes obsolete. Sure thing users can still use submodules for anything they want, but datalad does not.

The counter argument: At present *any* basic handling of datalad datasets can be achieved without datalad. A switch to relying on datalad metadata as required information will put a permanent end to this. 

Are we good on this one?

@bpoldrack ?

bpoldrack:
Hold on a second. Trying to catch with the weekend work of you guys ... ;-)

yarikoptic:
> With such a change, we are no longer using the submodule functionality to register subdatasets

why?  although it would somewhat duplicate (path) information contained in submodules, it would not replace it (although potentially could eventually, but ATM not sure if we should pursue it).  We could still rely on git's mechanisms of dealing with submodules as e.g. installing them (thus using url field in gitmodules, which could point outside of local hierarchy, and we could harvest meta-data from online repos), guaranteeing their correct status according to the commit information contained in git for that Subproject (they call it so in the commit IIRC) -- somewhat anal while maintaining big hierarchies but useful, etc.  As far as I see it, it would only supplement with meta-information to provide a check if submodule is a dataset.

mih:
I don't agree with this statement.

By making "being a submodule" a necessary but insufficient condition to be considered a subdataset, we are simply making it more complicated (for us, and for the user), without any visible gain, unless we go further.

Users cannot use plain git commands to manipulate repositories anymore, they have to use datalad (or have to mimic what datalad does internally). We complicate the internal handling by having two distinct locations for information that now need to be kept in sync, and we need to additionally handle all case when they are not (people can still manipulate the repos...).

At the same time we are stuck with the way submodules work (do not track a branch, unless we shoehorn that onto them, only one download URL, where we could have more). Moreover, we have quite a bit of code that makes sure that we don't use submodules the way they are intented to be used (.git dir placement, ....).

I am not saying that it is easy to drop submodules, but I am saying that just adding more components that need to be managed will make things complicated and we should think how much more we should do to actually get something out of it. Right now it looks as if we are changing the deifnition based on a first idea and see how it goes.

bpoldrack:
> The counter argument: At present any basic handling of datalad datasets can be achieved without 
> datalad. A switch to relying on datalad metadata as required information will put a permanent end to this.

> Are we good on this one?

No, I'm not good with it and I agree to @mih's point of view, that this approach leads to the effect, that people cannot deal with datalad datasets without datalad any more. And I'm strongly opposing it.
Datalad is supposed to make it easier, while still allowing for everything to be done via git/git-annex. We are on a track of losing this, if we distinguish between a "valid" dataset and a git repository. To me a git repo should be a dataset. Always. Being a submodule should be sufficient to be considered a subdataset. The entire thing should still make sense when datalad is gone for some reason. I think this is an important point of design. I don't see why we would need to not treat a repository as something not being a dataset and I think this would do a lot of harm to the entire perception of datalad. A subdataset might not be published along with its superdataset or may be even excluded from meta data (I don't think that makes sense, but well ...), but this kind of things should depend on some configuration instead of on a distinction between a git repo and a dataset.
I really hope we can agree on this as the fundament to derive a decision about IDs from.

yarikoptic:
Started to compose a lengthy summary summarizing the points, but scraped it since I think it will not be productive -- we should just discuss it interactively (hangout?).  Overall I think we are again trying to fit multiple things into one bucket -- the one and only kind of a `Dataset`, whenever we have at least 3 different types (a `Dataset` which points to non-existing path, a `Dataset` which points to a git or annex which has no idea about datalad, a `Dataset` which datalad created and about which it could have additional meta-data such as `id` etc -- I will call it a `datalad dataset` for this post).  Any of the first two could be transformed into a `datalad dataset` (and technically vise-versa).  Current implementation tries to wash out that difference by generating ids on the fly to have all 3 of them look the same.

Indeed datalad is supposed to be just an *enhancement* on top of plain git/annex repos, and basic operations on those datasets should be possible without datalad (obviously anything meta-data related could be a prerogative of datalad).  But at times we need to be able clearly determine the boundaries  among submodules where a datalad (super)dataset (as a collection of datasets) ends and regular git/annex repositories begins, but some times we want to cross those boundaries to provide those *enhancements* since we can do it.  Sample use cases:

- having a datalad dataset as a submodule of a regular git (e.g. software) project.  When going up (e.g. for `search -d^`) we need to be able to stop before escaping into (super-)project
- having a regular repository (e.g. 3rd party git repo), which we might not even have control over (i.e. can't add meta), as a submodule -- e.g. under some `bids_dataset/code/3rdpartydeps/software1`.  We do want to install it with `datalad install -r` but I am not sure that we want to aggregate its info since it has none.  Having said that I acknowledge that it is not exactly correct -- we could extract some kind of an `id` for it (e.g. left-most first commit) uniquely identifying it, and which would not change.  But IMHO we then should still make it clear that it is not a `datalad dataset`.

Altogether, I think we need to make it clear(er) on what a `Dataset` is (FTR: according to glossary it is indeed any git/annex repo) and be able to tell apart those 3 types since some operations could/should operate across all of those types, but some (ATM I see only `search -d^` and possibly `aggregate-metadata`) should respect the boundaries of `datalad dataset` s .

yarikoptic:
FTR (to myself including):  While working on another PR (guess which? ;)) by mistake I have ran
```datalad install -d. --if-dirty=ignore ///openfmri/ds00000{1,4,7} somepath``` within datalad source project code.  And that is where on purpose I have made `.git/datalad` available but be a file owned by root (so no evil things could happen to it).    On one hand, if I ran this one intensionally, I could say I wanted intensionally to add those as submodules to current repository (without making it into a full fledged Datalad dataset).    On another -- since in this case it was clearly an error -- I would have preferred it to just fail since '-d.' points not to a datalad dataset...

bpoldrack:
Re @yarikoptic 's "FTR comment":
Well, if I type `sudo rm -rf ~` I also want it to fail in case it wasn't intentional and to not fail if it was intentional. ;-)
The thing is: If you run something valid by accident, then it's an accident. We can't avoid that. The only way to make it less likely, is a very clear API, that doesn't allow for too much valid input and doesn't try too heavily to make sense of everything. Sounds familiar, right? ;-)

yarikoptic:
@bpoldrack try zsh

bpoldrack:
I do. But still missing the point. What do you mean?

yarikoptic:
At least in my case, might be configuration, i do get prompt to confirm such big kills

bpoldrack:
Yeah - ok.
So, that's what you want for a case like the one you mentioned? What would be the criteria to ask for confirmation? A missing `.datalad` in superdataset? Might be a way to deal with non-datalad datasets in general. They are valid for datalad, but whenever they are becoming part of a dataset hierarchy, we ask for confirmation.

yarikoptic:
yeah... something like that... so we could call those all to be a "dataset"

1. `empty dataset` -- a Dataset which points to non-existing path
2. `non-initialized dataset` (or `git dataset`?) -- a Dataset which points to a git or annex which has no idea about datalad -- i.e. missing an id (so might have `.datalad` subdir with other stuff)
3. `datalad dataset` -- a Dataset which datalad created and about which it could have additional meta-data such as id etc.  I would decide on being a `datalad dataset` by it having an id.

Then we could even add an option to any of our traversal (recursion) commands (we have already `existing` to stop on pretty much any not-yet-installed dataset) to e.g. also stop recursion on `non-initialized` dataset, which we could deduce from aggregated metadata in case of not-yet installed datasets.  When going up decision is easier to make, but we could by default bubble up until top `datalad dataset` but allow (option) to go beyond into `non-initialized` (still needs to be a git repo) parents too.

mih:
I will close this one now. With the merge of #1050 the starting point has changed.

Regarding the meta data with can now aggregate "over" non-dataset submodules and still track any proper datasets below.

Regarding meta data *read* access: This is now independent of any actual submodule relationship, but relies on `aggregate_metadata` to do the right thing.

As a result, the pieces are less intertwined.

I will go through the diff of this PR and check what still hasn't made it into master at this point and propose a new PR with the remaining pieces, if any.

Issue #335, opened 2016-01-14
Publish to github with specialremote for data access


mih:
Here is the protocol of setting up a fairly convenient state to aid public data consumption. This protocol assumes that you only push to a server that shall host the data, and do not access the local machine from the server (as this will often not be possible).

Start with regular local annex repo.

```
git init
git annex init "thebeginning"
git annex add .
git commit
```

On the remote server create new Git repo. We use a non-bare repo. Using a bare repo is possible and makes some of the steps below unnecessary. However, having a work tree on the server allows for additional use cases (browsing, etc.)

```
# --shared all is likely what you want, but double-check before running
git init --shared=all
# with modern Git the next is unnecessary and can be replaced with the line afterwards
#git checkout -b ignore
git config receive.denyCurrentBranch updateInstead
mv .git/hooks/post-update.sample .git/hooks/post-update
git update-server-info
```

On the local machine add the new remote. Do it twice, once for push access via SSH, and once for anonymous pull access via HTTP. And push the content.

```
git remote add httpdata HTTPURL/.git
git remote add sshdata SSHURL
git push sshdata git-annex
git push sshdata master
```

Now go back to the remote, and init the annex.

```
# with the modern Git setup the following is not required
#git checkout master
git annex init "public data"
```

Now the remote repo can be enabled as a special remote on the client-side. Once done, sync the state with the remote.

```
git annex initremote datasrc type=git location=HTTPURL/.git autoenable=true
# at this point the http remote is no longer needed
git remote remove httpdata
git fetch sshdata
git annex merge
```

Now the whole thing can be pushed to github. Create a repo on github, and add it as a remote to the local repo. Then push

```
git push github master
git push github git-annex
```

At this point, anybody can clone from github and get a local annex repo with an automatically enabled special remote.

```
git clone https://github.com/SOMETHING mine
git annex init
```

At this point the `datasrc` did not receive any files yet, hence none can be annex-get.  The necessary workflow is as follows. In the original repo do:

```
git annex copy SOMEFILE --to sshdata
git push github git-annex
```

This is sufficient to make a file available and the availability known. In the anonymously cloned repo now do:

```
git pull
git annex get SOMEFILE
```

and it will be downloaded via HTTP from `datasrc`.

The above is verified and works. The rest is anticipation and extrapolation:

Further workflow for maintaining `sshdata`:

```
git annex sync sshdata
```

Further workflow for publishing to github

```
git push github git-annex
git push github master
```

[...]

PR #1011, opened 2016-10-14
RF+ENH+BF: #959 reincarnated -- with  old install API (-s SOURCE PATH*)

yarikoptic:
See #959 for many discussions on the topic.

- [x] logic in install on handling PATH/SOURCE is still the old one, needs to be RFed as well
- [x] tests must be fixed
- [x] additional tests (may be as well mocked up) for install to invoke correct actions (e.g. calling get)
- [x] Issues which came up while testing #959 (I think all were addressed by @mih or by me by now)
- [x] Should close #1000 
- [x] Should close #1001 
- [x] Decide possibly how to avoid having `_`-prefixed argument in `Get.__call__`  (discussed with @bpoldrack , and agreed that if needs changing, could be done later)
- [x] Provide documentation outlining similarities and differences between `install` and `get` commands
- [x] add `-J` (jobs) option to `install` to pass to `get`

For future PRs
- Unify options between `get` and `install` APIs more  addressing those commented out ones which are specific only to get or to install currently

https://github.com/mih/datalad/pull/13 shows changes on top of original (and improved) rf-getinstall from @mih

PR #1018, opened 2016-10-15
NF: `create_sibling_github` (see gh-335)

mih:
This is a working preview. Some aspects will be later test(ed/able). But
I do not plan to test the actual interaction with the Github API.

Many possible features are missing (customizable repo name templating,
...). It is unlikely that this PR will get them, because my focus is on
addressing the core of #335

TODO for #335:

- [x] enhance `add_sibling` to be able to configure a special-remote (type git)
- [x] enhance `add_sibling` to configure a publish dependency on another sibling
- [x] make `publish` to honor a publish dependency a TODO in #1037 

When this is done, one can create two siblings (one regular, one on github), and afterwards simply `datalad publish github` to have all data and handles moved to a server and afterward all handles to github with a autoenabled special remote, so anyone installing from github gets data too.


yarikoptic:
tried on a plain datasets.datalad.org  clone with few datasets fetched... it seems to try to enable github remote for annex -- but it shouldn't since it should actually be annex=ignore
```shell
(git)hopa:/tmp/datasets.datalad.org[master]git
$> datalad create-sibling-github --github-organization datalad-collection-1 -r --existing=reconfigure datasets.datalad.org
...
i> 
2016-10-19 16:16:02,825 [ERROR  ] Failed to run ['git', '-c', 'receive.autogc=0', '-c', 'gc.auto=0', 'annex', 'enableremote', 'github'] under u'/tmp/datasets.datalad.org/crcns/aa-2'. Exit code=1. out=enableremote github failed
|  err=git-annex: enableremote: 1 failed
|  
2016-10-19 16:16:02,825 [INFO   ] Failed to enable annex remote github, could be a pure git 
2016-10-19 16:16:02,877 [ERROR  ] Failed to run ['git', '-c', 'receive.autogc=0', '-c', 'gc.auto=0', 'annex', 'enableremote', 'github'] under u'/tmp/datasets.datalad.org/kaggle/melbourne-university-seizure-prediction'. Exit code=1. out=enableremote github failed
|  err=git-annex: enableremote: 1 failed
|  
2016-10-19 16:16:02,877 [INFO   ] Failed to enable annex remote github, could be a pure git 
'git://github.com/datalad-collection-1/datasets.datalad.org.git' configured as sibling 'github' for <Dataset path=/tmp/datasets.datalad.org>
'git://github.com/datalad-collection-1/datasets.datalad.org-crcns.git' configured as sibling 'github' for <Dataset path=/tmp/datasets.datalad.org/crcns>
'git://github.com/datalad-collection-1/datasets.datalad.org-crcns-aa-2.git' configured as sibling 'github' for <Dataset path=/tmp/datasets.datalad.org/crcns/aa-2>
'git://github.com/datalad-collection-1/datasets.datalad.org-kaggle.git' configured as sibling 'github' for <Dataset path=/tmp/datasets.datalad.org/kaggle>
'git://github.com/datalad-collection-1/datasets.datalad.org-kaggle-melbourne-university-seizure-prediction.git' configured as sibling 'github' for <Dataset path=/tmp/datasets.datalad.org/kaggle/melbourne-university-seizure-prediction>
```

yarikoptic:
how difficult would be to implement 'replace' strategy for existing?

yarikoptic:
gy -- with such a huge patch (on top of #1011) 
```
--- a/datalad/distribution/utils.py
+++ b/datalad/distribution/utils.py
@@ -269,7 +269,12 @@ def _get_flexible_source_candidates(src, base_url=None):
         else:
             base_path = ri.path
             base_suffix = ''
-        ri.path = normpath(opj(base_path, src, base_suffix))
+        if ri.hostname == 'github.com':
+            # It must be a flat hierarchy
+            assert(base_suffix == '')
+            ri.path = "%s-%s" % (base_path, src)
+        else:
+            ri.path = normpath(opj(base_path, src, base_suffix))
 
     src = str(ri)
 
```
we should be able to install/get those published from github (see e.g. a limited collection of those which I have installed prior adding a sibling at  https://github.com/datalad-collection-1)

yarikoptic:
I would still prefer if double slash `--` was used to separate levels in a flat hierarchy (as we hypothesized in the proposal).  This is more likely to avoid any collision (e.g. between `n1/n2-suf`, `n1/n2/suf`, and `n1-n2/suf`).  Moreover we could place that check/restriction during `create` to not allow `--` (or make it interactive to get a confirmation to allow) in the names.

mih:
Thanks for the fixes!

> it seems to try to enable github remote for annex -- but it shouldn't since it should actually be annex=ignore

Will try to figure out where to pass this -- in any case annex sees it eventually and sets this by itself.

> how difficult would be to implement 'replace' strategy for existing?

Depends on what "replace" should do? If it is to wipe out an existing repo, it requires user/passwd auth (token is not enough). Otherwise it has full access and can destroy anything. Not sure how many warning signs I would want to go with that -- it is not only a repo that is gone in this case (which could be brought back easily in many cases), but also issues, wikis, access configurations, ....).

> I would still prefer if double slash -- was used to separate levels in a flat hierarchy (as we hypothesized in the proposal). This is more likely to avoid any collision (e.g. between n1/n2-suf, n1/n2/suf, and n1-n2/suf). Moreover we could place that check/restriction during create to not allow -- (or make it interactive to get a confirmation to allow) in the names.

This is all doable when any templating is introduced. I for one will *not* use any double dashes, and not use the current simplistic implementation either. Each repo is (and should be) standalone on github, a user does not need to care of any *possibly* existing hierarchy, any particular repo can be part of any number of hierarchies at the same time, and github does a good job of linking interconnected subdatasets to each other already.

mih:
Related: #1035

mih:
Seems to work. Will merge this now and @bpoldrack will finish the job as part of #1037 (thanks!).

yarikoptic:
> I for one will not use any double dashes, and not use the current simplistic implementation either

Hm... Then for whom this implementation is if neither of us will/could use it? :-)
Or on another side - if you don't use anyways it lets use -- and then adjust through templating :-)

If for sfn, if not crcns we could demo having the same repos on datasets.d.o and github, and have them usable/installable, it would be great

yarikoptic:
As for replace - could force push existing and then force push removal of old unique branches?

datalad/distribution/tests/test_create_github.py
```diff
@@ -0,0 +1,61 @@
...
+from nose.tools import assert_raises, assert_in, assert_true, assert_false, \
+    assert_not_in, assert_equal
+from nose import SkipTest
+
+
+try:
+    import github as gh
+except ImportError:
+    # make sure that the command complains too
+    assert_raises(MissingExternalDependency, create_sibling_github, 'some')
```
yarikoptic: Lovely! Although I would not exercise assert right at import time -  make it into a new test_ function right here and nose will pick it up I i guess.

datalad/distribution/create_sibling_github.py
```diff
@@ -0,0 +1,373 @@
...
+    @staticmethod
+    def result_renderer_cmdline(res, args):
+        from datalad.ui import ui
+        res = assure_list(res)
+        if args.dryrun:
+            ui.message('DRYRUN -- Anticipated results:')
+        if not len(res):
+            ui.message("Nothing done")
+        else:
+            for d, url, existed in res:
```
yarikoptic: I think output is nowhere described... Btw shouldn't it be an ordered (if order matters) dict with d as a key?  I guess I would be to look also into regular create-sibling..  beasts should be similar

datalad/distribution/create_sibling_github.py
```diff
@@ -0,0 +1,352 @@
...
+        reponame=Parameter(
+            args=('reponame',),
+            metavar='REPONAME',
+            doc="""Github repository name. When operating recursively,
+            a suffix will be appended to this name for each subdataset""",
+            constraints=EnsureStr()),
+        recursive=recursion_flag,
+        recursion_limit=recursion_limit,
+        sibling_name=Parameter(
+            args=('--sibling-name',),
```
mih: I don't agree. It makes no sense to do it this way for this command.

datalad/distribution/create_sibling_github.py
mih: Fine, but the nature of this one is different. The old one doesn't even have --dryrun. one thing at a time.

mih:
If you need to see double-dashes at once or nothing at all, go for it. This is just one piece in a bigger picture, and to me it is not yet the time to work on this because those hours are likely wasted. But I am not here to judge others ;-)

PR #1022, opened 2016-10-16
ENH (tiny): decide that dataset not-installed if path does not even exist

### LINKED ISSUE #1012: RFC: Dataset.is_installed should not leak errors
```
> /home/yoh/picts/mris/heudiconv/bin/heudiconv(1188)add_to_datalad()
-> if not ds.is_installed():
(Pdb) p ds
<Dataset path=/home/yoh/picts/mris/dartmouth-phantoms/bids_test4-20161014/bids-output>
(Pdb) n
ERROR: <class 'git.exc.NoSuchPathError'>: /home/yoh/picts/mris/dartmouth-phantoms/bids_test4-20161014/bids-output
INFO: Failed to detect a valid repo at /home/yoh/picts/mris/dartmouth-phantoms/bids_test4-20161014/bids-output
> /home/yoh/picts/mris/heudiconv/bin/heudiconv(1189)add_to_datalad()
-> lgr.info("Initiating %s", ds)
```

I think it would be logical to swallow error messages -- after all we are asking if is installed. agreed?

yarikoptic:
I think it should Closes #1012

### DIFF
--- a/datalad/distribution/dataset.py
+++ b/datalad/distribution/dataset.py
@@ -336,11 +336,15 @@ class Dataset(object):
         bool
         """
         # do early check manually if path exists to not even ask git at all
-        was_once_installed = self.path is not None and \
-                             exists(self.path) and \
-                             self.repo is not None
+        exists_now = exists(self.path)
 
-        if was_once_installed and not exists(self.repo.repo.git_dir):
+        was_once_installed = None
+        if exists_now:
+            was_once_installed = self.path is not None and \
+                                 self.repo is not None
+
+        if not exists_now or \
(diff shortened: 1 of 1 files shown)


mih:
I don't think this is OK.

A dataset that existed once, and is now completely gone, will continue to carry a `repo` instance with this change. The condition needs to go into the conditional just below the one you altered.

mih:
That should work. Feel free to merge when the tests pass.

yarikoptic:
fixed logic, tests passed (besides nd70 now disabled), so merging. thanks

PR #1023, opened 2016-10-16
NF: RFC822-based meta data format targeting manual meta data entry scenarios

mih:
Which invents as little as possible. Objectives:

1. Compact
2. Suitable for hand-editing (we have two JSON-based ones already)
3. Flexible enough to be able to capture pretty much anything I could think of

Choosing Debian's RFC822 felt natural. Pretty much all these files are composed by hand (that's proof enough).

The idea is to merge the various ideas from Debian's metadata files into a single spec file (control, copyright, ...). If necessary, we could have multiple files of course, but it doesn't look like we need it (yet).

At this point, the implementation represents the minimal technical infrastructure to support this format on a demo level, and to get the discussion going.

### DIFF
--- a/datalad/metadata/parsers/datalad_rfc822.py
+++ b/datalad/metadata/parsers/datalad_rfc822.py
@@ -57,22 +57,34 @@ class MetadataParser(BaseMetadataParser):
         for header, dataladterm in \
                 (('name', 'name'),
                  ('license', 'license'),
-                 ('maintainer', 'author'),
-                 #('homepage', 'citation'),
-                 #('Funding', 'foaf:fundedBy'),
-                 ('description', 'description')):
+                 ('author', 'author'),
+                 ('maintainer', 'doap:maintainer'),
+                 ('audience', 'doap:audience'),
+                 ('homepage', 'doap:homepage'),
+                 ('version', 'doap:Version'),
+                 ('funding', 'foaf:fundedBy'),
+                 ('issue-tracker', 'bug-database'),
+                 ('cita-as', 'citation'),
+                 ('doi', 'sameAs'),
+                 ('description', None)):
             if not header in spec:
                 continue
             content = spec[header]
-            if header in ('description', 'license'):
-                title, content = _beautify_multiline_field(content)
+            if header == 'description':
+                short, long = _beautify_multiline_field(content)
+                meta['doap:shortdesc'] = short
+                meta['description'] = long
+            elif header == 'license':
                 # TODO if title looks like a URL, use it as @id
-                if title:
-                    meta[dataladterm] = [title, content]
+                label, desc = _beautify_multiline_field(content)
+                if label:
+                    meta[dataladterm] = [label, desc]
                 else:
-                    meta[dataladterm] = content
-            elif header in ('maintainer',):
+                    meta[dataladterm] = desc
+            elif header in ('maintainer', 'author'):
                 meta[dataladterm] = _split_list_field(content)
+            elif header == 'doi':
+                meta[dataladterm] = 'http://dx.doi.org/{}'.format(content)
             else:
                 meta[dataladterm] = content
 
--- a/datalad/metadata/parsers/datalad_rfc822.py
+++ b/datalad/metadata/parsers/datalad_rfc822.py
@@ -1,0 +1,78 @@
+# emacs: -*- mode: python; py-indent-offset: 4; tab-width: 4; indent-tabs-mode: nil -*-
+# ex: set sts=4 ts=4 sw=4 noet:
+# ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+#
+#   See COPYING file distributed along with the datalad package for the
+#   copyright and license terms.
+#
+# ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+"""Parser for RFC822-based metadata specifications
+
+This is inspired by (and very similiar to) Debian's package meta data format.
+The main difference is that information spread across multiple files in Debian
+packages, is concentrated in one file.
+
+The main advantage of this format is that it is proven to be hand-editable,
+i.e. can be composed from scratch, by hand, in an editor -- with a good
+chance of producing syntax-compliant content with the first attempt.
+"""
+
+import rfc822
+from os.path import join as opj
+from datalad.metadata.parsers.base import BaseMetadataParser
+from datalad.interface.base import dedent_docstring
+
+
+def _split_list_field(content):
+    return [i.strip() for i in content.split(',') if i.strip()]
+
+
+def _beautify_multiline_field(content):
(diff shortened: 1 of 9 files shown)


mih:
Note to self: https://github.com/edumbill/doap/wiki/JSON-LD-@context-for-package.json

mih:
FTR: Build failures are unrelated to the actual changes (NITRC rejects connections...)

datalad/metadata/parsers/datalad_rfc822.py
```diff
@@ -0,0 +1,78 @@
...
+            content = '{}{}{}'.format(
+                content,
+                ' ' if len(content) and l != '.' and content[-1] != '\n' else '',
+                l if l != '.' else '\n')
+    return title, content
+
+
+class MetadataParser(BaseMetadataParser):
+    _metadata_compliance = "http://docs.datalad.org/metadata.html#v0-1"
+    _core_metadata_filenames = [opj('.datalad', 'control')]
```
yarikoptic: I think naming it 'control' (I guess inspired by `debian/control`) is imho suboptimal for many reasons, one of which is that meaning of this file would not be understood by non-Debian folks.  Even Debian folks, like yours truly, might feel violated since they only know that 'debian/rules' and 'debian/control's with 'debian/' prefix.  What about `.datalad/meta.rfc822`?

datalad/metadata/parsers/datalad_rfc822.py
```diff
@@ -0,0 +1,78 @@
...
+    def _get_metadata(self, ds_identifier, meta, full):
+        spec = rfc822.Message(open(self.get_core_metadata_filenames()[0]))
+
+        # loop over all recognized headers and translate them
+        for header, dataladterm in \
+                (('package', 'name'),
+                 ('license', 'license'),
+                 ('maintainer', 'author'),
+                 #('homepage', 'citation'),
+                 #('Funding', 'foaf:fundedBy'),
```
yarikoptic: IMHO if we are to provide a "datalad"'s own meta-data format, then it should be a superset of all fields we value, which would include both homepage, and funding. So why not to keep it in?

bpoldrack:
No objections.

mih:
Ok, as feeback was on the technical side I assume the general verdict was: right direction

yarikoptic:
well - tech sides aside, I do see value in supporting non-json metadata format!  but just to make clear -- introducing it nohow would help in creating debian packages for datalad datasets since we would anyways need to generate produce description from packages from  aggregated `meta.json`... but may be then something like this format could become a substitute for meta.json format?  i.e. then could natively be edited (instead of being aggregated)?  or do you foresee difficulty of serializing some entries into it from whatever other formats we would be aggregating from?

yarikoptic:
> FTR: Build failures are unrelated to the actual changes (NITRC rejects connections...)

you can restart entire build by clicking a button on travis... will do for this one

yarikoptic:
FWIW -- travis failures are genuine -- py3

datalad/metadata/parsers/datalad_rfc822.py
mih: Done

datalad/metadata/parsers/datalad_rfc822.py
mih: Time will come.

mih: Time will come.

mih:
Downscaled the goals of this PR. If you get a chance check the test and recommend additional fields. The task is to map them to terms.

Remember: the goal of this one is primarily to be hand-editable, so please no non-essential imposed structure -- just because it is possible.

yarikoptic:
> recommend additional fields

well -- those which we support in aggregated data and I commented on ;)  we will be giving a demo on monday, so might be nice to show all the 'funding' search

mih:
Last think before merge will be a little blurb about this in the docs.

datalad/metadata/parsers/datalad_rfc822.py
```diff
@@ -0,0 +1,92 @@
...
+                (('name', 'name'),
+                 ('license', 'license'),
+                 ('author', 'author'),
+                 ('maintainer', 'doap:maintainer'),
+                 ('audience', 'doap:audience'),
+                 ('homepage', 'doap:homepage'),
+                 ('version', 'doap:Version'),
+                 ('funding', 'foaf:fundedBy'),
+                 ('issue-tracker', 'bug-database'),
+                 ('cita-as', 'citation'),
```
yarikoptic: cite-as

yarikoptic: Is there a way we could somehow standardize it so it could be machine parseable for duecredit?

datalad/metadata/parsers/tests/test_rfc822.py
```diff
@@ -0,0 +1,91 @@
...
+  ],
+  "doap:shortdesc": "Basic summary",
+  "foaf:fundedBy": "BMBFGQ1411, NSF 1429999",
+  "license": [
+    "CC0",
+    "The person who associated a work with this deed has dedicated the work to the public domain by waiving all of his or her rights to the work worldwide under copyright law, including all related and neighboring rights, to the extent allowed by law.\\nYou can copy, modify, distribute and perform the work, even for commercial purposes, all without asking permission."
+  ],
+  "name": "studyforrest_phase2",
+  "sameAs": "http://dx.doi.org/10.5281/zenodo.48421",
+  "type": "Dataset"
```
yarikoptic: Citation is missing

datalad/metadata/parsers/datalad_rfc822.py
yarikoptic: Also for funded by

mih: There are exactly 537 known ways to standardize. UMEGAYA has proposed one with less than 25  subfields. But I see no benefit for anything other than a DOI and custom text, hence very low on my list of things to work on.

datalad/metadata/parsers/tests/test_rfc822.py
mih: Jep, that is the bug you pointed out above.

datalad/metadata/parsers/datalad_rfc822.py
yarikoptic: I like doi and tolerate free text. But could we avoid need for magic/parsing to get those of this field? Especially if multiple references

mih: Not sure I can follow. What magic parsing?

datalad/metadata/parsers/tests/test_rfc822.py
mih: Done.

mih:
Docs are in.

mih:
Will stop here.

mih:
We support arbitrary content in aggregate.

PR #1029, opened 2016-10-17
RF+ENH: dedicated [devel] + requirements-devel.txt mode of installation 

yarikoptic:
Regular requirements.txt will now attempt to install as it should without -e (but that would entail copying entire working directory).  So if you are developing -- using requirements-devel.txt which would also make sure that you have ALL needed for development requirements (including sphinx, useful nose plugins etc)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -92,7 +92,7 @@ before_install:
 
 install:
   # Install standalone build of git-annex for the recent enough version
-  - travis_retry sudo apt-get install git-annex-standalone zip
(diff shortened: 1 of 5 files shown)


yarikoptic:
upgraded buildbot config to demand pip>=8 although jessie has only 1.5.8 but I think we will be fine since on those which we care about ppl will install via apt-get or would develop on, thus would use requirements-devel.txt

yarikoptic:
any objections/comments or I will merge it soon

PR #1033, opened 2016-10-19
BF: SSH shared connections issues

### LINKED ISSUE #826: `create-publication-target-sshwebserver` does not reuse connection
````
% datalad create-publication-target-sshwebserver localhost:/tmp/dest rmt
2016-09-14 15:43:47,342 [INFO   ] Connecting ... (create_publication_target_sshwebserver.py:209)
mih@localhost's password: 
2016-09-14 15:43:50,244 [INFO   ] Creating target dataset some at /tmp/dest (create_publication_target_sshwebserver.py:239)
2016-09-14 15:43:50,338 [INFO   ] Adjusting remote git configuration (create_publication_target_sshwebserver.py:287)
2016-09-14 15:43:50,357 [INFO   ] Enabling git post-update hook ... (create_publication_target_sshwebserver.py:307)
2016-09-14 15:43:50,430 [INFO   ] Uploading web interface to /tmp/dest (create_publication_target_sshwebserver.py:321)
mih@localhost's password: 
````

password requested twice.

### LINKED ISSUE #865: install  asks for password twice -- controlpath config is setup but controlpath is not established
Might be related to #826 

```

hopa:/tmp
$> datalad -l 1 install smaug:/home/yoh/mris/GA/24AUG16GA_24AUG16GA_24AUG16GA/HEAD_CLINICAL_LIBRARIES_20160824_095910_981000
...
2016-09-20 10:19:55,705 [Level 5] Detected ssh ri (network.py:317)
2016-09-20 10:19:55,705 [DEBUG  ] Git clone from smaug:/home/yoh/mris/GA/24AUG16GA_24AUG16GA_24AUG16GA/HEAD_CLINICAL_LIBRARIES_20160824_095910_981000 to /tmp/HEAD_CLINICAL_LIBRARIES_20160824_095910_981000 (gitrepo.py:504)
Enter passphrase for key '/home/yoh/.ssh/id_rsa': 
yoh@smaug's password: 
2016-09-20 10:20:00,518 [DEBUG  ] Git clone completed (gitrepo.py:507)
...
2016-09-20 10:20:00,523 [DEBUG  ] Running: ['git', '-c', 'receive.autogc=0', '-c', 'gc.auto=0', 'annex', 'init', '--debug'] (cmd.py:352)
Enter passphrase for key '/home/yoh/.ssh/id_rsa': 
Enter passphrase for key '/home/yoh/.ssh/id_rsa': 
yoh@smaug's password: 
```

and at that stage we do have config already setup to use controlpath but the controlpath is not there yet
```
hopa:/tmp
$> cat HEAD_CLINICAL_LIBRARIES_20160824_095910_981000/.git/config 
[core]
	repositoryformatversion = 0
	filemode = true
	bare = false
	logallrefupdates = true
[remote "origin"]
	url = smaug:/home/yoh/mris/GA/24AUG16GA_24AUG16GA_24AUG16GA/HEAD_CLINICAL_LIBRARIES_20160824_095910_981000
	fetch = +refs/heads/*:refs/remotes/origin/*
	annex-ssh-options = -o ControlMaster=auto -S /home/yoh/.config/datalad/smaug
[branch "master"]
	remote = origin
	merge = refs/heads/master
[annex]
	uuid = 64fbdbca-31f1-410e-a7d9-95dd9230edaf
	version = 5

$> ls -l /home/yoh/.config/datalad/smaug
ls: cannot access '/home/yoh/.config/datalad/smaug': No such file or directory
```

### LINKED ISSUE #1032: SSH ControlPath in inappropriate location
````
In [5]: ssh = sm.get_connection('localhost:/tmp/dummy')
In [6]: ssh.open()
mih@localhost's password: 
In [7]: ssh.open()
In [8]: ssh.ctrl_path
Out[8]: '/home/mih/.config/datalad/localhost'
````

ssh_config(5) says:

> It is recommended that any ControlPath used for opportunistic connection sharing include at least %h, %p, and %r (or alternatively %C) and be placed in a directory that is not writable by other users.

This is not true in our default setup.

The placement of the socket might also be related to the issues of SSH connections not being reused #826 #865 (of which at least the former is still present in master as of now). Moreover, the control master socket is no config file. I could imagine people having their `.config` in a VCS.

The socket could go into `~/cache/datalad/sockets` which would be made `700`. Actual target directory should come from a config setting.

One more: We do `ControlPersist=yes`. Are we 100% positive that under any circumstances we will not leave behind something that could be used for (un)intentional evil? Or should we at least use something like `ControlPersist=1h`?

@bpoldrack?

### PR DESCRIPTION
**Pull request by bpoldrack**
Fixes issues with shared SSH connections. Led to asking for credentials several times.
- Cloning and  `datalad add-sibling` didn't correctly use this mechanic
- directory used for SSH control masters now depends on config item `datalad.locations.cache`
- Deal with existing but not running control masters
- let annex know about setting a remote's url
- time out for control master: 15 minutes

Closes #826 
Closes #865 
Closes #1032 

- [x] make socket location configurable

Note: More on the topic of publishing and siblings to come in another PR. Also more testing.

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -297,6 +297,21 @@ class AnnexRepo(GitRepo):
                              " -S %s" % c.ctrl_path)
             writer.release()
 
+    def set_remote_url(self, name, url, push=False):
+        """Overrides method from GitRepo in order to set
+        remote.<name>.annex-ssh-options in case of a SSH remote."""
+
+        super(AnnexRepo, self).set_remote_url(name, url, push=push)
+        from datalad.support.network import is_ssh
+        if is_ssh(url):
+            c = ssh_manager.get_connection(url)
+            writer = self.repo.config_writer()
+            writer.set_value("remote \"%s\"" % name,
+                             "annex-ssh-options",
+                             "-o ControlMaster=auto"
+                             " -S %s" % c.ctrl_path)
+            writer.release()
+
     def __repr__(self):
         return "<AnnexRepo path=%s (%s)>" % (self.path, type(self))
 
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -1315,6 +1315,26 @@ class GitRepo(object):
             else:
                 return None
 
+    def set_remote_url(self, name, url, push=False):
+        """Set the URL a remote is pointing to
+
+        Sets the URL of the remote `name`. Requires the remote to already exist.
+
+        Parameters
+        ----------
+        name: str
+          name of the remote
+        url: str
+        push: bool
+          if True, set the push URL, otherwise the fetch URL
+        """
+
+        cmd = ["git", "remote", "set-url"]
+        if push:
+            cmd.append("--push")
+        cmd += [name, url]
+        return self._git_custom_command('', cmd)
+
     def get_branch_commits(self, branch, limit=None, stop=None, value=None):
         """Return GitPython's commits for the branch
 
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -495,6 +495,14 @@ class GitRepo(object):
         path : str
         """
 
+        if is_ssh(url):
+            cnct = ssh_manager.get_connection(url)
+            cnct.open()
+            # TODO: with git <= 2.3 keep old mechanism:
+            #       with rm.repo.git.custom_environment(GIT_SSH="wrapper_script"):
+            env = {'GIT_SSH_COMMAND': "ssh -S %s" % cnct.ctrl_path}
+        else:
+            env = None
         ntries = 5  # 3 is not enough for robust workaround
         for trial in range(ntries):
             try:
@@ -502,6 +510,7 @@ class GitRepo(object):
                 self.repo = self.cmd_call_wrapper(gitpy.Repo.clone_from,
                                                   url,
                                                   path,
+                                                  env=env,
                                                   odbt=default_git_odbt)
                 lgr.debug("Git clone completed")
                 break
--- a/datalad/support/sshconnector.py
+++ b/datalad/support/sshconnector.py
@@ -125,7 +125,7 @@ class SSHConnection(object):
         if not exists(self.ctrl_path):
             # set control options
             ctrl_options = ["-o", "ControlMaster=auto",
-                            "-o", "ControlPersist=yes"] + self.ctrl_options
+                            "-o", "ControlPersist=15m"] + self.ctrl_options
             # create ssh control master command
             cmd = ["ssh"] + ctrl_options + [self.host, "exit"]
 
--- a/datalad/support/sshconnector.py
+++ b/datalad/support/sshconnector.py
@@ -14,6 +14,7 @@ git calls to a ssh remote without the need to reauthenticate.
 """
 
 import logging
+from os.path import exists
 from subprocess import Popen
 from shlex import split as sh_split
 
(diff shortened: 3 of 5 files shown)

### CODE REVIEW

File: datalad/support/gitrepo.py [hunk 1]
```diff
@@ -495,13 +495,22 @@ def clone(self, url, path):
...
+        else:
+            env = None
         ntries = 5  # 3 is not enough for robust workaround
         for trial in range(ntries):
             try:
                 lgr.debug("Git clone from {0} to {1}".format(url, path))
                 self.repo = self.cmd_call_wrapper(gitpy.Repo.clone_from,
                                                   url,
                                                   path,
+                                                  env=env,
```
mih: I think this is a problem. Not here but in the runner implementation. This will cause the entire process environment to be replaced with the above minimal environment definition. That might lead to unexpected things. One solution would be to copy os.environ, ammend it with the above setting, and pass it on.

datalad/support/sshconnector.py
```diff
@@ -111,15 +112,16 @@ def open(self):
         connection, if it is not there already.
         """
 
-        # set control options
-        ctrl_options = ["-o", "ControlMaster=auto", "-o", "ControlPersist=yes"] + self.ctrl_options
-        # create ssh control master command
-        cmd = ["ssh"] + ctrl_options + [self.host, "exit"]
+        if not exists(self.ctrl_path):
```
mih: What if the socket exists, but it is no longer listenting? Can this happen?

datalad/support/gitrepo.py
bpoldrack: Nope. This is custom for the git call. `Runner` isn't really used here to call `git`. It just wraps the function call (which points into GitPython) in order to be able to protocol those calls.

datalad/support/sshconnector.py
```diff
@@ -185,7 +187,7 @@ def socket_dir(self):
         if self._socket_dir is None:
             # TODO: centralize AppDirs (=> datalad.config?)
             from appdirs import AppDirs
-            self._socket_dir = AppDirs('datalad', 'datalad.org').user_config_dir
+            self._socket_dir = AppDirs('datalad', 'datalad.org').user_cache_dir
```
mih: It should be a subdirectory, and get `700` permsions. Subdir name could come from a config setting that has the above as a part. I think one like this is already defined.

datalad/support/sshconnector.py
bpoldrack: I don't think so. But we probably should ask the ssh executable anyway indeed. Will figure it out.

datalad/support/gitrepo.py
mih: Hmm, I must be reading the code wrong.

bpoldrack: Probably. ;-)
Here `Runner.__call__` is called with an callable as argument. This is different from calling an actual command that way.

datalad/support/sshconnector.py
yarikoptic: It happens to me from time to time (after resume from sleep?) that I have a socket file but it is stable - attempt to login just hangs until I remove it. But I am not sure if anything could be done besides some check with a timeout

bpoldrack: Yes, we can! :-)
`ssh -O check` ...

yarikoptic:
Would you be kind to change title/description of the PR to summarize the changes?  I see that multiple aspects are tuned but not sure which are critical to accomplish the mission.  Since I do try to collate a usable changelog now for releases, it is very helpful to have descriptive merge messages which come when you merge a PR.

bpoldrack:
All of them are critical of course! ;-)
Yes, will provide a better description. First need to be sure, what needs to be addressed.

mih:
Works for me now.

Needs directory permissions fixed still.

datalad/support/gitrepo.py
mih: OK, was missing that part.

datalad/support/annexrepo.py
```diff
@@ -297,6 +297,21 @@ def add_remote(self, name, url, options=None):
...
             writer.release()
 
+    def set_remote_url(self, name, url, push=False):
+        """Overrides method from GitRepo in order to set
+        remote.<name>.annex-ssh-options in case of a SSH remote."""
+
+        super(AnnexRepo, self).set_remote_url(name, url, push=push)
+        from datalad.support.network import is_ssh
+        if is_ssh(url):
+            c = ssh_manager.get_connection(url)
```
yarikoptic: theoretically such constructs should be within try/finally
```python
try:
  get connection
finally:
  release
```
otherwise if something booms in writer but caught upstairs and "handled" we might be having a dangling thing with unknown consequences

bpoldrack: I'm not sure, what exactly you are talking about.

1. What do you mean by "caught upstairs and handled"? If "upstairs" handles such an exception, how is this problematic? Or do you mean some `except Exception"? If this leads to an issue, I'd consider this except-clause the issue!

2. If `writer = self.repo.config_writer()` causes an Exception, then `writer.release()` will also fail (`AttributeError`)

yarikoptic: ok then you enclose further:
```python
try:
            writer.set_value("remote \"%s\"" % name,
                             "annex-ssh-options",
                             "-o ControlMaster=auto"
                             " -S %s" % c.ctrl_path)
finally:
            writer.release()
```
so if set_value fails (I don't know how and don't want to know if it could or not -- just taking no chances), then writer would release ... although I must say that not sure if that would be a good thing -- does it save at that point (then better just to fail -- which would be your code now), or just removes some lock on the `.git/config` (then my try/final)

yarikoptic:
will merge for now ;) try/finally is a small thing

Issue #758, opened 2016-09-06
Publish fails on subdataset with detached HEAD. Should it ?


debanjum:
Me and Yarik think it should probably give an error message or something instead of crashing like it currently did. Checking out master fixed it for now

``` sh
         2016-09-02 18:37:01,510 [INFO   ] Publishing dataset <Dataset path=/home/debanjum/datasets.datalad.org/openfmri/ds000200> to sibling public-server ... (publish.py:270)
         2016-09-02 18:37:02,546 [WARNING] Nothing was pushed. (publish.py:48)
         2016-09-02 18:37:03,674 [WARNING] Nothing was pushed. (publish.py:48)
         Traceback (most recent call last):
           File "/home/debanjum/datalad/venv-tests/bin/datalad", line 9, in <module>
             load_entry_point('datalad==0.2.3', 'console_scripts', 'datalad')()
           File "/home/debanjum/datalad/datalad/cmdline/main.py", line 245, in main
             ret = cmdlineargs.func(cmdlineargs)
           File "/home/debanjum/datalad/datalad/interface/base.py", line 285, in call_from_parser
             return cls.__call__(**kwargs)
           File "/home/debanjum/datalad/datalad/distribution/publish.py", line 235, in __call__
             published_, skipped_ = ds_.publish(to=to, recursive=recursive, **pkw)
           File "/home/debanjum/datalad/datalad/distribution/dataset.py", line 510, in apply_func
             return f(**kwargs)
           File "/home/debanjum/datalad/datalad/distribution/publish.py", line 235, in __call__
             published_, skipped_ = ds_.publish(to=to, recursive=recursive, **pkw)
           File "/home/debanjum/datalad/datalad/distribution/dataset.py", line 510, in apply_func
             return f(**kwargs)
           File "/home/debanjum/datalad/datalad/distribution/publish.py", line 176, in __call__
             track_remote, track_branch = ds.repo.get_tracking_branch()
           File "/home/debanjum/datalad/datalad/support/gitrepo.py", line 1432, in get_tracking_branch
             branch = self.get_active_branch()
           File "/home/debanjum/datalad/datalad/support/gitrepo.py", line 758, in get_active_branch
             return self.repo.active_branch.name
           File "/home/debanjum/datalad/venv-tests/local/lib/python2.7/site-packages/git/repo/base.py", line 670, in active_branch
             return self.head.reference
[...]

Issue #481, opened 2016-05-25
publish should push also tags and branches


yarikoptic:
realized today that while publishing we didn't push any tags or branches.
Not sure what default behavior should be but may be 'publish' command should allow to specify the list of `push` targets to set in the config for that remote in every dataset it works on?  or may be just to allow to pass `--all --tags` to push call BUT that would not persist for other calls

yarikoptic:
probably should be done in 'prepare' stage.
Could be set per branch within .git/config, e.g. I have in pymvpa git:
```
[remote "gh-pymvpa-hyper"]
	url = https://github.com/PyMVPA/PyMVPA-hyper
	push = refs/tags/upstream/*
	push = refs/heads/master
	push = refs/heads/conhyp
```

yarikoptic:
yeap... I think it should be done in prepare stage, BUT then publish code should either use native 'git push' command or load and pass those references to be pushed to GitPython's push. Currently just hard codes master and git-annex for annex repos

yarikoptic:
For now I manually had to do to repositories on smaug to be published
```
git config --add remote.datalad-public.push 'refs/tags/*'; git config --add remote.datalad-public.push 'refs/heads/*'
```
so I would say that `create_pub...` should do that by default, and either allow to tune the list of those refs to be pushed, or just have an option such as `--do-not-add-push-refs` which would instruct to not do it.
It must be careful though in `--existing=reconfigure` mode to not change those settings if already defined... or may be would be safer to not change those at all in `reconfigure` mode -- I will just fix up manually those lacking them for now

yarikoptic:
Anybody is working on this one @mih or @bpoldrack ?

mih:
Looking into this now. I agree that this needs to be configurable. I agree that this needs to be done in a git-native way.

Looking at `publish` it seems a precondition for this is to stop calling 'push' with specific refspecs.

When we are doing that, we could consider offloading more duties to git annex by making more use of `sync`. this could take care of all kinds of conflicts, and data transfer in one go. We still would want to perform uni-directional updates, but `--no-pull` should do that. All that for annex repos only, of course.

After annex-sync is done (or for any other non-annex repo), we can run a simple `git push <remote>` to push the remaining configured 
items to send.

Any comments @bpoldrack

bpoldrack:
Not ATM. I agree.

mih:
#1275 is where we are in the discussion right now.

Issue #925, opened 2016-09-27
RFC: create-sibling --wanted=manual


yarikoptic:
So, finally got to play with git meta-data to ease publishing with the content.  It could be quite sweet, e.g.:

``` shell
> git annex wanted datalad-public "not metadata=distribution-restrictions=*"
> git annex copy --to=datalad-public --auto -J6
```

worked out splendidly for http://datasets.datalad.org/?dir=/labs/haxby/raiders - what should have not been published, was not ;)  So, also in light of #820 discussion, I wondered if we just should provide option `--wanted` which would default to `manual` (thus nothing should get copied by default if I got it right), but could as well be `backup` to a complete copy online, or `client` (for all content in working dir but not under "/archive" etc, see [1] for more groups), or `manual` to not copy anything automagically (and thus should be overridden upon specific publish invocation by passing options to annex or specifying PATH(s).  But the point is that at `create-sibling` time we would instrument such a policy, thus possibly avoiding  any manual interactions upon subsequent "publish".

@mih -- are you using preferred content already for your repos?

[1] https://git-annex.branchable.com/preferred_content/standard_groups/

bpoldrack:
Generally I like the approach to set such "policies" to provide a default configuration for `publish` behaviour. But since you mentioned to put this into `create_sibling`: I think this should rather go into `add_sibling` (which is called by `create_sibling`), since this is not about setting up the repo to publish to, but about configuring a remote.

yarikoptic:
Nope. Because those policies iirc are set per remote within git-annex branch for that remote uuid, ie they are omnipresent across all clones. So you don't want to go flipping them back and forth from different clones - anarchy would come.

yarikoptic:
Ok, I will address this one by providing `-w,--wanted` which could be either
- one of the [standard annex groups](https://git-annex.branchable.com/preferred_content/standard_groups/)
- few of our custom groups
    - `backup-unrestricted` -- `not metadata=distribution-restrictions=*`
    - `unique` -- to push derived data (the one present only locally) mostly, but somewhat simplistic since would be just `copies=1` (might consult @joeyh on optimal one).  
   - `unique-unrestricted` -- as unique but with constrained restriction, which we could advise on while defining a sibling if there is any file with 'distribution-restrictions'
- and if not matching any of those -- taken as a [preferred content](https://git-annex.branchable.com/preferred_content/) expression  (or should we be kosher and make those separate?  `--wanted-group` and `--wanted` staying more in line with annex?)

[...]

PR #1037, opened 2016-10-20
Rework publish Part I

bpoldrack:
First part of reworking publish. Will stop here for now and focus on #1091 first, since this will make some things more clear and easy.

@datalad/developers : I'd vote for merge to get the fixes into master.

@yarikoptic : Regarding #481 - you can setup `git push` configuration via `add-sibling` and `create-sibling` now. I'm aware there's more to that issue in general, but this has to wait for the RF'ing.
Furthermore it would be nice to confirm, that #1063 isn't happening anymore with this branch.

TODO:
- [x] fix several SSH issues
- [x] make publish honor dependencies (see #1018)
- [x] address #481 (partially done; this is a more general thing)
- [x] #1063 (check: might be solved already)
- [x] consider #758 
- [x] respect `remote.*.annex-ignore`


yarikoptic:
please look into addressing  #481 as well!

The pairing change should happen in create-sibling which should have option and setup branches/refs on what to be pushed, and probably the data to be pushed (#925).  But even without it, fixing #481 but removing custom list of refs to push, this would be of help since now I need to run git push manually after each publish whenever I do set proper list of references to be pushed.

datalad/interface/common_opts.py
```diff
@@ -148,4 +148,14 @@
...
+    action='append',
+    constraints=EnsureStr() | EnsureNone())
+
+publish_by_default = Parameter(
+    args=("--publish-by-default",),
+    metavar='REFSPEC',
+    doc="""add a refspec to be published to this sibling by default if nothing
+    specified.""",
+    constraints=EnsureStr() | EnsureNone(),
     action='append')
```
yarikoptic: I believe that it might need multiple values eg I push all refs and all tags, but sketching that manually is really not user friendly... I thought i had had somewhere an issue to describe common possible publication schemes. But I guess I will try and see first

bpoldrack: You can give that option multiple times. Each will result in an entry.

yarikoptic: as discussed briefly during hangout -- information about what to be pushed shouldn't go to `.datalad/config` since that one contains persistent/committed to git information and we are just talking about specific siblings (remotes) of a specific dataset (clone), so imho should be just in `.git/config`,.

But there specifically -- how do I to specify to push both all refs and all tags?? should then at least be `nargs="+"` and me to specify lengthy `--publish-by-default=refs/tags/* --publish-by-default=refs/heads/*`, or how?

datalad/support/exceptions.py
```diff
@@ -117,6 +117,25 @@ def __str__(self):
...
+        self.repo = repo
+        self.branch = branch
+        self.branches = available_branches
+        if msg is None:
+            self.msg = "branch '{0}' missing in {1}." \
+                       "".format(self.branch, self.repo)
+            if self.branches:
+                self.msg += " Available branches: {0}".format(self.branches)
+        else:
+            self.msg = msg
```
yarikoptic: why not just to call super and pass msg to it as the first positional argument?

```python
In [1]: e = Exception("messaga")

In [2]: e.message
Out[2]: 'messaga'
```
currently exception would have both .msg and empty .message I guess

datalad/interface/common_opts.py
bpoldrack: `--publish-by-default=refs/tags/* --publish-by-default=refs/heads/*` should work, yes.
Did you notice `action='append'` or do I get sth wrong?

datalad/distribution/publish.py
```diff
@@ -256,49 +282,90 @@ def __call__(path=None, dataset=None, to=None, recursive=False, since=None,
...
+                           recursion_limit=recursion_limit,
+                           git_opts=git_opts,
+                           annex_opts=annex_opts,
+                           annex_copy_opts=annex_copy_opts)
+
+            lgr.info("Publishing {0} to {1}".format(ds, dest_resolved))
+
+            # we now know where to push to:
+            # TODO: what to push? default: git push --mirror if nothing configured?
+            # consider also: --follow-tags, --tags, --atomic
```
yarikoptic: as we discussed during one of the hangouts before (see in notes from 09/15/2016), and I thought agreed upon, best would be just to rely on settings within `.git/config`.  That would eliminate such questions here in publish.  There is really no need to reinvent the wheel/duplicate settings within publish or what could be set directly within git.  That would then "natively" provide resolution to #481 (if git config was setup accordingly by e.g. create-sibling, see #925)

datalad/interface/common_opts.py
yarikoptic: I indeed missed append, thanks

datalad/distribution/publish.py
```diff
@@ -49,16 +51,17 @@ def _log_push_info(pi_list):
...
 
-    This makes the current state of a dataset available to a sibling of the
-    dataset. That sibling needs to be known to the dataset before.
-    `create_publication_target` commands are meant to set up such a sibling, but
-    generally you can publish to any sibling added to a dataset.
-    Publishing may or may not include subdatasets.
-    By default file handles are published without their actual content,
-    but they can be published including the content, of course.
+    This makes the last saved state of a dataset available to a sibling
+    or special remote data store of the dataset which must already exist
```
yarikoptic: Oh... Shouldn't we now introduce some new term to improve clarity, e.g. 'bastard', when taking about special remotes? :-)

datalad/distribution/publish.py
```diff
@@ -256,49 +282,91 @@ def __call__(path=None, dataset=None, to=None, recursive=False, since=None,
...
                 if not isinstance(ds.repo, AnnexRepo):
-                    raise RuntimeError(
-                        "Cannot publish content of something, that is not "
-                        "part of an annex. ({0})".format(ds))
+                    # incomplete, since `git push` was done already:
+                    raise IncompleteResultsError(
+                        (published, skipped),
+                        failed=publish_files,
+                        msg="Cannot publish content of something, that is not "
+                            "an annex. ({0})".format(ds))
```
yarikoptic: Why you can't? It is just that you don't have to do anything special to make it happen, bit you can publish them.
My main concern is this exception while publishing recursively set of datasets since if which are git. IMHO this exception will be a big obstacle and Skype be replaced with a simple warning at most

datalad/distribution/publish.py
```diff
@@ -256,49 +282,91 @@ def __call__(path=None, dataset=None, to=None, recursive=False, since=None,
...
+                if ds.config.get('remote.{}.annex-ignore', False):
+                    # Q: Do we need a --force option here? annex allows to
+                    # ignore the ignore setting
+                    raise IncompleteResultsError(
+                        (published, skipped),
+                        failed=publish_files,
+                        msg="Sibling '{0}' of {1} is configured to be ignored "
+                            "by annex. No content was published."
+                            % (dest_resolved, ds))
```
yarikoptic: This one too... May be I am not entirely clear on the flow and logic here - but in both cases wouldn't it interrupt publishing as soon as it encounters one of those cases?

datalad/distribution/tests/test_publish.py
```diff
@@ -206,7 +207,6 @@ def test_publish_with_data(origin, src_path, dst_path, sub1_pub, sub2_pub):
     # TODO: Figure out, when to fetch things in general; Alternatively:
     # Is there an option for push, that prevents GitPython from failing?
     source.repo.fetch("target")
-
```
yarikoptic: So there were quite a few changes in publish behavior, but none of those effects were teste since I don't see any test being added?

mih:
In the interest of fewer dangling PRs, I will merge this now, and will help deal with any potential fallout if necessary.

yarikoptic:
ok... I was testing it as of speaking on http://datasets-dev.datalad.org ... so far didn't run into a fallout besides pushing into the wrong remote (didn't specify --to datalad-public-dev and forgot that still had our main one available... but I guess that is "by design" and user error ;) )

PR #1042, opened 2016-10-21
Use install in log and add space

yarikoptic:
While on the phone


yarikoptic:
Travis didn't run... But that should be OK

PR #1052, opened 2016-10-24
RF: Sacrifice a few log lines to be able to configure logging via config

mih:
At present only for once switch, but many additional switches are
conceivable

### DIFF
--- a/datalad/__init__.py
+++ b/datalad/__init__.py
@@ -9,21 +9,17 @@
 """DataLad aims to expose (scientific) data available online as a unified data
 distribution with the convenience of git-annex repositories as a backend."""
 
-from .log import lgr
 
 # Other imports are interspersed with lgr.debug to ease troubleshooting startup
 # delays etc.
-lgr.log(5, "Instantiating config")
 from .config import ConfigManager
 cfg = ConfigManager()
 
-lgr.log(5, "Instantiating ssh manager")
 from .support.sshconnector import SSHManager
 ssh_manager = SSHManager()
 
 import atexit
 atexit.register(ssh_manager.close, allow_fail=False)
-atexit.register(lgr.log, 5, "Exiting")
 
 from .version import __version__
 
@@ -54,6 +50,7 @@ _test_states = {
 
 
 def setup_package():
+    from .log import lgr
     import os
 
     # To overcome pybuild overriding HOME but us possibly wanting our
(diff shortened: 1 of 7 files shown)


datalad/__init__.py
```diff
@@ -9,21 +9,17 @@
...
 
-from .log import lgr
 
 # Other imports are interspersed with lgr.debug to ease troubleshooting startup
 # delays etc.
-lgr.log(5, "Instantiating config")
 from .config import ConfigManager
 cfg = ConfigManager()
 
-lgr.log(5, "Instantiating ssh manager")
```
yarikoptic: couldn't we keep this one already since config is already imported and now we could import log?

asking because all those log lines were important/used to optimize import times, so if we are to prune -- we should prune as little as sufficient

datalad/log.py
```diff
@@ -125,7 +125,10 @@ def __init__(self, use_color=None, log_name=False, log_pid=False):
 
     def _get_format(self, log_name=False, log_pid=False):
         # TODO: config log.timestamp=True
-        return (("" if not int(os.environ.get("DATALAD_LOG_TIMESTAMP", True)) else "$BOLD%(asctime)-15s$RESET ") +
+        from datalad import cfg
+        from datalad.config import anything2bool
+        show_timestamps = anything2bool(cfg.get('datalad.log.timestamp', True))
```
yarikoptic: BTW -- thought about it meanwhile -- should we may be default to False, i.e. remove those time stamps since by default they are just polluting the screen and not used much?

mih: Love you for saying that! 100% in favor.

datalad/__init__.py
mih: Done

datalad/log.py
mih: Testing the change now.

datalad/__init__.py
```diff
@@ -106,5 +107,3 @@ def teardown_package():
     lgr.debug("Teardown tests. " + msg)
     for path in _TEMP_PATHS_GENERATED:
         rmtemp(path, ignore_errors=True)
-
-lgr.log(5, "Done importing main __init__")
```
yarikoptic: and this one was redundant or why removed?

mih: This was an attention test. I can assert that you are performing on (at least) average level. Congrats!

mih:
Seems to work.

datalad/__init__.py
```diff
@@ -54,6 +53,7 @@ def test(package='datalad', **kwargs):
 
 
 def setup_package():
+    from .log import lgr
```
yarikoptic: those aren't needed strictly speaking, right? since now there is a `from .log import lgr` above it

PR #1050, opened 2016-10-25
RF meta data handling to disentangle the pieces

mih:
Currently `get_metadata()` is also doing half of `aggregate_metadata()`.
The plan is to have the `get` part to work the exact same way, whether
or not (some) subdatasets are installed. It will look into the given
dataset only, and not traverse *into* possible installed subdatasets.
All that is known to it, is locally aggregated meta data.

TODO:

- [x] RF read access to aggregated meta data into a regular parser
      implementation
- [x] RF out the option to have `get_metadata()` do magic to bypass
      aggregated data and go straight into a subdataset
- [x] instead make sure that a prior `aggregate_metadata()` call
      can quickly and reliably perform a (partial) update of subdataset
      meta data

This will change the way and location where decisions are made. It will
be irrelevant for `get_metadata()` whether some component actually exists.
This investigation, and the decision on how to deal with it will be left to
the aggregation stage. This will make it easier and more comprehensible to
switch to an on-demand meta data update.

### DIFF
--- a/datalad/crawler/pipelines/tests/test_openfmri.py
+++ b/datalad/crawler/pipelines/tests/test_openfmri.py
@@ -262,14 +262,14 @@ def test_openfmri_pipeline1(ind, topurl, outd, clonedir):
     eq_(len(commits_l['incoming']), 3)
     eq_(len(commits['incoming-processed']), 6)
     eq_(len(commits_l['incoming-processed']), 4)  # because original merge has only 1 parent - incoming
-    eq_(len(commits['master']), 11)  # all commits out there -- dataset init, crawler init + 3*(incoming, processed, merge)
-    eq_(len(commits_l['master']), 5)
+    eq_(len(commits['master']), 12)  # all commits out there -- dataset init, crawler init + 3*(incoming, processed, meta data aggregation, merge)
+    eq_(len(commits_l['master']), 6)
 
     # Check tags for the versions
     eq_(out[0]['datalad_stats'].get_total().versions, ['1.0.0', '1.0.1'])
     # +1 because original "release" was assumed to be 1.0.0
     eq_([x.name for x in repo.repo.tags], ['1.0.0', '1.0.0+1', '1.0.1'])
-    eq_(repo.repo.tags[0].commit.hexsha, commits_l['master'][-3].hexsha)  # next to the last one
+    eq_(repo.repo.tags[0].commit.hexsha, commits_l['master'][-4].hexsha)  # next to the last one
     eq_(repo.repo.tags[-1].commit.hexsha, commits_l['master'][0].hexsha)  # the last one
 
     def hexsha(l):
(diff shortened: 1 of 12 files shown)


mih:
If the tests past this is worth a look @yarikoptic @bpoldrack

mih:
Those commit counting tests ... :-(

yarikoptic:
Yeah... Those are "strict" but somewhat make sense generally

mih:
So the reason for the failing pipeline commits is that aggregate is called one additional time per "crawl" (see screenshot). It seems to be related to a second aggregate run on a switch to the master branch. But I am not 100% positive that I understand the algorithm that leads to this. My gut wants to tell me that the first aggregate run records the version (commit) the meta data is based on in the meta data. The result is then committed and tagged. This implies that the actual last commit and the one on record in the meta data are different. In any case, all should be good at this point. But then aggregate is ran once again, naturally finds the version mistmatch and another commit is made. This could go on forever, of course.

I don't know enough about how the crawling is meant to work, but I believe the second aggregate run could just be disabled. In this case it would only have 3 additional commits.

But in general I think the crawler doesn't want aggregate to commit anything. So we would need a `--nosave`. Because by default, I think, aggregate should commit.

image

mih:
Ok, so I implemented a `save=bool` option. But disabling the commit in the crawler leads to

````
RuntimeError: Repository /tmp/datalad_temp_test_openfmri_pipeline1kOsVZr is dirty.  Finalize your changes before running this pipeline
````

Dunno what it wants from me.

yarikoptic:
On October 29, 2016 5:49:50 AM EDT, Michael Hanke <notifications@github.com> wrote:
>Ok, so I implemented a `save=bool` option. But disabling the commit in
>the crawler leads to
>
>````
>RuntimeError: Repository
>/tmp/datalad_temp_test_openfmri_pipeline1kOsVZr is dirty.  Finalize
>your changes before running this pipeline
>````
>
>Dunno what it wants from me.
>
>-- 
>You are receiving this because you were mentioned.
>https://github.com/datalad/datalad/pull/1050#issuecomment-257082368

Crawler by default refuses to work in a dirty repository. Saved my butt few times already

So we are enjoying the chicken and egg problem now I guess. Shouldn't version/commit to be then recorded by the super dataset only? If you have this dataset installed, you can always check its version from git

mih:
That would be a possible approach. I'll think about it, and see how complicated it would get. My worry is to not get back to this mix of reporting what is aggregated and what could be replaced with info from an actually installed dataset. The goal of this PR was to make metadata more independent of what is installed or not.

mih:
All good now.

yarikoptic:
Oh... Was about to try it today... So, with current version I can run aggregate twice and second time it wouldn't generate a new commit (since no data should get changed) right?

mih:
Yes, take a look at the changes in the tests. Except for an additional dedicated aggregate-commit nothing should change. If that is not true, we need more tests (would not be surprised).

Things are still a bit in flux. With the change in contexts it could make sense to compact the graph already in aggregate -- my first exploration of file-based meta data will tell (I hope)....

PR #1058, opened 2016-10-25
When `add` failed due to insufficient permissions it reports ("unknown reason")

mih:
@agramfort maybe you can past the log from your server? Thanks!

### DIFF
--- a/datalad/interface/common_opts.py
+++ b/datalad/interface/common_opts.py
@@ -13,7 +13,8 @@
 __docformat__ = 'restructuredtext'
 
 from datalad.support.param import Parameter
-from datalad.support.constraints import EnsureInt, EnsureNone, EnsureStr
+from datalad.support.constraints import EnsureInt, EnsureNone, EnsureStr, \
+    EnsureChoice
 
 
 dataset_description = Parameter(
@@ -32,6 +33,15 @@ recursion_limit = Parameter(
     metavar="LEVELS",
     constraints=EnsureInt() | EnsureNone(),
     doc="""limit recursion into subdataset to the given number of levels""")
+
+shared_access_opt = Parameter(
+    args=('--shared-access',),
+    metavar='MODE',
+    constraints=EnsureChoice(
+        'false', 'true', 'umask', 'group', 'all',
+        'world', 'everybody', '0xxx') | EnsureNone(),
+    doc="""configure shared access to a dataset, see `git init --shared`
+    documentation for complete details on the supported scenarios""")
 
 super_datasets_flag = Parameter(
     args=("-S", "--super-datasets",),
--- a/datalad/interface/common_opts.py
+++ b/datalad/interface/common_opts.py
@@ -13,8 +13,7 @@
 __docformat__ = 'restructuredtext'
 
 from datalad.support.param import Parameter
-from datalad.support.constraints import EnsureInt, EnsureNone, EnsureStr, \
-    EnsureChoice
+from datalad.support.constraints import EnsureInt, EnsureNone, EnsureStr
 
 
 dataset_description = Parameter(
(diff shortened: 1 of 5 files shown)


mih:
Made it work by penetrating the non-functional general mechanism sideways with a rubber spoon.

mih:
@agramfort you could try this branch and do

````
datalad create --shared-access group
````

That should do it.

**Comment by agramfort** [comment]
works like a charm !

**Comment by yarikoptic** [comment]
Damn thing didn't post my reply via email. There is --shared for create-sibling already but there it is too flexible imho and we should unify between the two since it is the same option passed to git init call in both cases

**Comment by agramfort** [comment]
I did manage to add my files and clone the repo on my laptop but when trying to get a file I get:

```
datalad get sample_submission.csv
2016-10-25 18:43:03,024 [INFO   ] Getting 1 items of dataset <Dataset path=/Users/alex/[snip]> ...

Tried to get 1 file. Failed to get 1.
sample_submission.csv ... failed. ((Note that these git remotes have annex-ignore set: origin))
```

any idea?

**Comment by yarikoptic** [comment]
@agramfort probably upon a failed (for some reason) attempt to retrieve stuff from a remote, annex set annex-ignore property on that remote -- just edit your `.git/config` there, remove for that remote where we know annex **is** present and retry please.

**Comment by agramfort** [comment]
I removed the  annex-ignore = true from my local .git/config but it puts it
back automatically and the "datalad get" does not complete.

what should I try?

**Comment by agramfort** [comment]
ok problem solved... my sys admin accepted to update git annex for all our machines :)

**Comment by yarikoptic** [comment]
@agramfort -- it should have just failed out loud if annex was older than 6.20160808.  otherwise it should have been "usable" ;)

File: datalad/interface/common_opts.py [hunk 1]
```diff
@@ -33,6 +33,13 @@
     constraints=EnsureInt() | EnsureNone(),
     doc="""limit recursion into subdataset to the given number of levels""")
 
+shared_access_opt = Parameter(
+    args=('--shared-access',),
```
yarikoptic: so what about just `--shared` to be inline with `create-sibling`, @mih?

mih: I think that would be short-sighted, given the optimizations that are yet to come. Shared access (here), shared annex (we talked about). I cannot foresee if they would be incremental, or orthogonal. Hence I went for a special long option for a special case.

yarikoptic: Well... I don't see how they ever really need to be different, since in both cases we are taking about "create", just that here it is local and the other one on a remote location.

mih: I thought so. But I'd like to avoid another "create_sibling" situation. We had to shorten it, because it was the only one. But now there are two -- but the command name never got longer again. These things can accumulate to an API structure that communicates more "history" than necessary. I'd prefer to first have the features implemented, not just envisioned.

yarikoptic: I don't see how command name choice directly relates to the option name here in question, but ok, if you see them needing to be differentiated, I can silence my whining for a bit longer ;-)

mih: No no. I don't mind also renaming the current ``--shared`` to something more informative! ;-)

anikfal:
> Doable with `datalad create --git-opts " --shared=group"`

Is it possible to apply `--shared=group` on an already created dataset?

yarikoptic:
> > Doable with `datalad create --git-opts " --shared=group"`
> 
> Is it possible to apply `--shared=group` on an already created dataset?

you can try and see but I would not think it would do all the needed dances. More of dances to do only for `git` could be found at https://stackoverflow.com/questions/3242282/how-to-configure-an-existing-git-repo-to-be-shared-by-a-unix-group but then also there are dances to do for git-annex and there is more of that, so frankly I would not even try.  I would start from a fresh repo which I would initialize into that shared mode first.

PR #1066, opened 2016-10-28
ENH+RF:  datalad add  progress indication + passing --jobs  into annex

yarikoptic:
as @agramfort pointed out -- ATM `datalad add BUNCHOFFILES` just stays silent for a while and then spits out its long list.  This PR adds a progress bar (based on file sizes, but also with total count) and exposes `--jobs` option, allowing to make checksumming faster

Resulted also in some refactorings etc ;)  Let's see now if anything got broken


yarikoptic:
If no objections or recommendations, will merge in few days

PR #1068, opened 2016-10-28
Choice 2  BF+RF:  do not log outputs at all for git config calls

### LINKED ISSUE #1060: Prevent ConfigManager from dumping config into the log (with all secrets)

yarikoptic:
Alternative to #1067 

Closes #1060


mih:
I like this one.

PR #1073, opened 2016-10-30
DOC: JSON-LD context info

mih:


mih:
Will merge now. For the next bits to work, I need this in master.

PR #1074, opened 2016-10-30
NF+RF+ENH: More comprehensive and configurable meta data support

mih:
Broadband improvements:

- [x] A custom context enables us to flexibly expand or adjust it, without having (much) impact on meta data that is already circulating around. And it is also more compact in the datasets ...
- [x] Going to versioned meta data item IDs allows us to distinguish between instances of the same dataset
- [x] config variable `datalad.metadata.parsers.enable` to replace `metadata.nativetypes`
- [x] config variable `datalad.metadata.parsers.disable` for "negative selection"  of parser, i.e. "all but these two"
- [x] annex information now from a separate parser that can be enabled/disabled
- [x] robustify `search` to keep working on mixture of old and new meta data
- [x] each meta data item now comes with a `describedby` property that identifies the responsible parser (with version) to enable flexible and reliable filtering later on
- [x] simple draft of a parser for file-based meta data that can report file name, size, mimetype and encoding
- [x] `search` can now search for datasets _and_ files (datasets by default)
- [x] switch to version 0.2 as meta data standard to reflect the changes
- [x] enhance BIDS meta data parser to produce file-based meta data as well
- [x] enhance `aggregate_metadata` to honor disabled parser in superdatasets -- this way users can configure selective meta data _removal_ in superdatasets, independent of the presence of the entire subtree, or without having to modify any subdataset.
- [x] any non-essential (e.g. file-based) meta data is separated into by-parser files that are (optionally, and by default) annexed.
- [ ] make optional annexed meta data files compressed by default -- no point in diffs anyway, so can be optimized for download
- [ ] RF `search` to make room for alternative search algorithms
- [ ] Implement support for SPARQL queries in `search`
- [ ] add bacon


yarikoptic:
Shouldn't then parsers be adjusted to use only or own schema instead of some terms eg for fundedby from outside, or am I mixing up things?

mih:
There are two things to this:

1. Technically, the parsers already use "our" schema with this change.

2. We can make things look simpler by defining "aliases" of essential terms (like "foaf:fundedBy" -> "fundedBy").

The latter we should be able to do in our newly created context that extends schema.org. If things work out as planned, graph-compacting should preserve the aliases.

So: yes, it is in the works. But no, we will not define any really new terms, just give convenience names that map to existing terms. So a machine will still be able to easily infer that "our" fundedBy" is the same as "foaf:fundedBy".

Likely we also want to define a scope for things like neurolex... but one thing at a time...

mih:
@yarikoptic I pushed a context proposal. Please note that it is not possible to have the context and the matching unit tests in one PR, because term resolution relies on the context to come from schema.datalad.org.

It is possible to use a local context by default, but that would mean that older installations would not look for new contexts.

mih:
If there are no objections re the context only, I will push this file into master so that we can enjoy the tests properly.

mih:
Prepared such a PR in #1075

mih:
@yarikoptic @bpoldrack file-based meta data is next. If you don't want gigantic PRs please take a look and act! ;-)

yarikoptic:
- so this version does generate a new commit every time I rerun aggregate-metadata, ie suffers from chicken/egg problem.  I don't think we should fall into this trap.  If we can't achieve "self containing" all the information (current version of myself as the git commit), we shouldn't pretend like we can.

- I re-aggregated info (on portion of sub-datasets, and then on top one, using `datalad aggregate-metadata --guess-native-type  --if-dirty ignore` and before that with -r from top until crashing for not this code problem, so did manualy in indi, etc with non -recursive invocation on top after) within `/srv/datasets-dev.datalad.org/www` on falkor (I believe you do have access to it @mih. I will make it group writeable now so you could play yourself) and then installed it locally -- search returns just top dataset hit:

```shell
hopa:/tmp
$> datalad install http://datasets-dev.datalad.org
[INFO   ] Installing dataset at /tmp/datasets-dev.datalad.org from http://datasets-dev.datalad.org 
<Dataset path=/tmp/datasets-dev.datalad.org>
(venv-tests) 2 16760.....................................:Tue 01 Nov 2016 02:18:35 PM EDT:.
hopa:/tmp
$> cd datasets-dev.datalad.org 
crcns/  dicoms/  indi/  kaggle/  labs/  neurovault/  nidm/  openfmri/
(venv-tests) 2 16761.....................................:Tue 01 Nov 2016 02:18:39 PM EDT:.
(git)hopa:/tmp/datasets-dev.datalad.org[master]
$> datalad search --regex -r\* .
[INFO   ] Loading and caching local meta-data... might take a few seconds 
.:
 @id: 4cc7e1fd67d269be388ddbe05882d939dd734619
 Type: Dataset
 conformsTo: http://docs.datalad.org/metadata.html#v0-1
 isVersionOf: @id=6d69ca68-7e85-11e6-904c-002590f97d84
 modified: 2016-10-11T00:26:47-04:00
```

yarikoptic:
and since we haven't even discussed file-based meta-data anyhow yet, it might be more sensible to create a separate PR for this, even though it would be sitting on top of this one.  To ease review, we could push this branch into the main repo, and base that PR to aim to be merged into it (so we could see only specific to it changes)

mih:
Trying to comprehend your observation. There should be no chicken and egg something, and I didn't see such behavior yet in my tests.

mih:
There is already a separate branch for the file-based meta data.

mih:
So has the code example you gave something to do with the error description? Is that the input, or the result? I'd like to reproduce, but my parser fails on:

> on portion of sub-datasets, and then on top one, using datalad aggregate-metadata --guess-native-type --if-dirty ignore and before that with -r from top until crashing for not this code problem, so did manualy in indi, etc with non -recursive invocation on top after

yarikoptic:
```shell
falkor:/srv/datasets-dev.datalad.org/www
*$> datalad --version  2>&1 | head -n 1
datalad 0.4.dev107
(venv-tests)(git)/mnt/btrfs/manual-snapshots/srv-20161101/datasets.datalad.org/www:[master]
#76 !159 [0].....................................:Tue Nov 01 14:38:36:.
falkor:/srv/datasets-dev.datalad.org/www
*$> cd indi
total 4
4 datapackage.json  0 fcon1000/
(venv-tests)(git)/mnt/btrfs/manual-snapshots/srv-20161101/datasets.datalad.org/www/indi:[master]
#77 !160 [0].....................................:Tue Nov 01 14:38:50:.
falkor:/srv/datasets-dev.datalad.org/www/indi
$> datalad aggregate-metadata --guess-native-type  --if-dirty ignore -r
[INFO   ] aggregating meta data for <Dataset path=/srv/datasets-dev.datalad.org/www/indi/fcon1000/AnnArbor_a>
[INFO   ] aggregating meta data for <Dataset path=/srv/datasets-dev.datalad.org/www/indi/fcon1000/AnnArbor_b>
...
[INFO   ] aggregating meta data for <Dataset path=/srv/datasets-dev.datalad.org/www/indi>
(venv-tests)(git)/mnt/btrfs/manual-snapshots/srv-20161101/datasets.datalad.org/www/indi:[master]
#78 !161 [0].....................................:Tue Nov 01 14:39:05:.
falkor:/srv/datasets-dev.datalad.org/www/indi
$> cd ../
total 4
0 balsa/  0 dicoms/     0 indi/    0 labs/        0 nidm/      0 test/
0 crcns/  4 index.html  0 kaggle/  0 neurovault/  0 openfmri/
(venv-tests)(git)/mnt/btrfs/manual-snapshots/srv-20161101/datasets.datalad.org/www:[master]
#79 !162 [0].....................................:Tue Nov 01 14:39:09:.
falkor:/srv/datasets-dev.datalad.org/www
*$> git show | head -n 1
commit 3c0a30b47922980bca0896bee3e3e1077e71a383
(venv-tests)(git)/mnt/btrfs/manual-snapshots/srv-20161101/datasets.datalad.org/www:[master]
#80 !163 [0].....................................:Tue Nov 01 14:39:20:.
falkor:/srv/datasets-dev.datalad.org/www
$> datalad aggregate-metadata --guess-native-type  --if-dirty ignore
[INFO   ] aggregating meta data for <Dataset path=/srv/datasets-dev.datalad.org/www/crcns>
[INFO   ] aggregating meta data for <Dataset path=/srv/datasets-dev.datalad.org/www/dicoms>
[INFO   ] aggregating meta data for <Dataset path=/srv/datasets-dev.datalad.org/www/indi>
[INFO   ] aggregating meta data for <Dataset path=/srv/datasets-dev.datalad.org/www/kaggle>
[INFO   ] aggregating meta data for <Dataset path=/srv/datasets-dev.datalad.org/www/labs>
[INFO   ] aggregating meta data for <Dataset path=/srv/datasets-dev.datalad.org/www/neurovault>
[INFO   ] aggregating meta data for <Dataset path=/srv/datasets-dev.datalad.org/www/nidm>
[INFO   ] aggregating meta data for <Dataset path=/srv/datasets-dev.datalad.org/www/openfmri>
[INFO   ] aggregating meta data for <Dataset path=/srv/datasets-dev.datalad.org/www>
(venv-tests)(git)/mnt/btrfs/manual-snapshots/srv-20161101/datasets.datalad.org/www:[master]
#81 !164 [0].....................................:Tue Nov 01 14:39:31:.
falkor:/srv/datasets-dev.datalad.org/www
$> git show | head -n 1
commit bc6d44989ccfa325d3acaa84acf6d675e285b6fa
(venv-tests)(git)/mnt/btrfs/manual-snapshots/srv-20161101/datasets.datalad.org/www:[master]
#82 !165 [0].....................................:Tue Nov 01 14:39:34:.
falkor:/srv/datasets-dev.datalad.org/www
$> datalad aggregate-metadata --guess-native-type  --if-dirty ignore
[INFO   ] aggregating meta data for <Dataset path=/srv/datasets-dev.datalad.org/www/crcns>
[INFO   ] aggregating meta data for <Dataset path=/srv/datasets-dev.datalad.org/www/dicoms>
[INFO   ] aggregating meta data for <Dataset path=/srv/datasets-dev.datalad.org/www/indi>
[INFO   ] aggregating meta data for <Dataset path=/srv/datasets-dev.datalad.org/www/kaggle>
[INFO   ] aggregating meta data for <Dataset path=/srv/datasets-dev.datalad.org/www/labs>
[INFO   ] aggregating meta data for <Dataset path=/srv/datasets-dev.datalad.org/www/neurovault>
[INFO   ] aggregating meta data for <Dataset path=/srv/datasets-dev.datalad.org/www/nidm>
[INFO   ] aggregating meta data for <Dataset path=/srv/datasets-dev.datalad.org/www/openfmri>
[INFO   ] aggregating meta data for <Dataset path=/srv/datasets-dev.datalad.org/www>
(venv-tests)(git)/mnt/btrfs/manual-snapshots/srv-20161101/datasets.datalad.org/www:[master]
#83 !166 [0].....................................:Tue Nov 01 14:39:41:.
falkor:/srv/datasets-dev.datalad.org/www
$> git show | head -n 1
commit a569caac14aca2e129de7cef2e3f14661a3110de
(venv-tests)(git)/mnt/btrfs/manual-snapshots/srv-20161101/datasets.datalad.org/www:[master]
```

mih:
Thanks. Found a simple test case. Basically two things done in reverse. Will fix.

mih:
Suggestions for further tests?

yarikoptic:
Did you "real life" tested it on that datasets-dev  collection?  I bet it could give you more soil to dig in ;)

mih:
No, i only have a fraction of that on my laptop. I can only work more or less offline for most of the time. I have to rely on test cases. If anyone has suggestions on what to test, please dont hesitate.

If there is a concrete list of things to tackle I will take a shot at creating test cases.

I should be able to come back to this in a week or so...

mih:
@yarikoptic Looked into possible glitches that you could have observed in your testing. If you reaggregated only partially, but include the top-level superdataset, all non-aggregated datasets will vanish from your search results. The reason for that is the very core of this PR. We switched the context. The runtime optimizations implemented in search assume a single context -- a condition that you may have invalided.

If you don't like that, one (not me, though) could theoretically implement more expensive checks and conversions that can unify items from individual contexts. But the whole point of this PR is to make such switches unnecessary in the future. So if you reaggregate everything it should work just fine.

From what I can see using my clone of the base dataset it does.

Context homogenization is possible, and the meta data item structure currently implemented affords that, but given that *all* our meta data comes from our own parsers, this is an exercise for a time when one of us a bored. Any takers? ;-)

mih:
Created #1086 to keep track of it.

From my POV this is done.

yarikoptic:
On November 2, 2016 3:11:10 PM EDT, Michael Hanke <notifications@github.com> wrote:
>@yarikoptic Looked into possible glitches that you could have observed
>in your testing. If you reaggregated only partially, but include the
>top-level superdataset, all non-aggregated datasets will vanish from
>your search results. The reason for that is the very core of this PR.
>We switched the context. The runtime optimizations implemented in
>search assume a single context -- a condition that you may have
>invalided.
>
>If you don't like that, one (not me, though) could theoretically
>implement more expensive checks and conversions that can unify items
>from individual contexts. But the whole point of this PR is to make
>such switches unnecessary in the future. So if you reaggregate
>everything it should work just fine.
>
>From what I can see using my clone of the base dataset it does.
>
>Context homogenization is possible, and the meta data item structure
>currently implemented affords that, but given that *all* our meta data
>comes from our own parsers, this is an exercise for a time when one of
>us a bored. Any takers? ;-)
>
>-- 
>You are receiving this because you were mentioned.
>https://github.com/datalad/datalad/pull/1074#issuecomment-257968847

Quick one - no need to overcomplicate it. I did partial one only because aggregating all was taking too long, and some failed because of other reasons and I had no time to fix-up... But we will need to sync the release with reaggregation and ideally detect wherever old style meta it's available to demand datalad update, 
That is why I thought it would be nice to introduce if not there a version check so we could also demand upgrade of datalad if new version of meta is detected

datalad/metadata/aggregate.py
```diff
@@ -157,16 +158,19 @@ def _within_metadata_store(ds, guess_native_type, metapath):
...
+    # subdatasets
+    ds_identifier = ds.repo.repo.head.commit.hexsha
     for m in meta:
-        if not is_implicit_metadata(m):
+        if not m.get('@id', None) == ds_identifier:
             continue
-        for prop in ('dcterms:modified', 'version'):
-            if prop in m:
-                del m[prop]
+        m['@id'] = 'THISDATASET!'
```
yarikoptic: pardon me -- what does it mean? i.e. why **this super dataset** doesn't have its own ID?
```shell
(git)hopa:/tmp/datasets-dev.datalad.org[master].datalad/meta
$> git grep THISDA
crcns/meta.json:    "isPartOf": "THISDATASET!",
dicoms/meta.json:    "isPartOf": "THISDATASET!",
indi/meta.json:    "isPartOf": "THISDATASET!",
kaggle/meta.json:    "isPartOf": "THISDATASET!",
labs/meta.json:    "isPartOf": "THISDATASET!",
meta.json:    "@id": "THISDATASET!",
neurovault/meta.json:    "isPartOf": "THISDATASET!",
nidm/meta.json:    "isPartOf": "THISDATASET!",
openfmri/meta.json:    "isPartOf": "THISDATASET!",
```

mih:
> it seems that instead of the "version" field, now we do have additional record with id corresponding to the latest commit, and isVersionOf link to the dataset record itself. What is the benefit here? that we could have the same 'version' be present in multiple datasets?

Pretty much is an acceptable paraphrasing of the PR description ;-)

This goes back to *your* feature request and a ~1h discussion in some hotel room. Except that it is not about *same* version, but about *different* versions of the same dataset. Rational is now in the description of the meta data standard update.

> per each sibling we have a separate record (referenced to by id in availableFrom), I guess for 'standardization' of some kind since I do not think we would benefit from reuse

No significant change from previous. If you don't like this info, you can look forward to https://github.com/datalad/datalad/pull/1077/commits/b0ed365a7f046e688b26edd2d4a95321085fce81 -- which makes it easy for anyone to en/disable this.

> when we accept this one we would need to release right when we merge (I can't introduce fixes and (recrawl) without reaggregating. we can't reaggregate until this version is released). So it might be worth "playing" with it before making this step

I guess you are talking about `search` not being functional for some combinations of old, older, and new, plus outdated caches, etc. I pushed a change that provides remedy for that. With an outdated cache, and an old copy of the base dataset you can now get this:

````
% datalad search '.' -r '*'
<snip>
dartmouth-phantoms/bids_test3:
 @id: e8d2d492-859c-11e6-95c9-8019340ce7f2
 @type: Dataset
 availableFrom: @id=008d6e24-df90-48b0-841b-2415e61ef96a, @id=58081f52-cd9d-4e8c-a34a-41905fa2ef26, @id=e48ce580-d0e2-45f0-8586-63d1b3486c11
 http://purl.org/dc/terms/conformsTo: http://docs.datalad.org/metadata.html#v0-1
 http://purl.org/dc/terms/isPartOf: 85c4465e-7aea-11e6-a5d5-002590f97d84
 http://purl.org/dc/terms/modified: 2016-09-28T12:59:39-04:00
 location: dartmouth-phantoms/bids_test3
 version: b5b39f2934347d8b03796bfe9d14f7d74c6f4602
<snip>
````

If that is not enough #1086 has the path to a brighter future.

> there is some THISDATASET!

If you got this from `get_metadata()` it is a bug. If you saw this in a file -- not a bug. This is how we internally refer to the dataset that contains an aggregated meta data file when it refers to itself. Unavoidable as the ID is unknown at the point of storage.

> we should depict the version of meta-data "beast" we have. so why not to boost

Done.

> So it might be worth "playing" with it before making this step

So, I'll merge #1074 back in to afternoon (someone requested it to be kept separate) and continue to march on. Is there a foreseeable end point other than "pain is too much, let's end this"? ;-)

**Comment by yarikoptic** [comment]
"I'll merge #1074 back in to afternoon"
A bit confused... Anyways - of I could search old and new data - that would be awesome and then I wouldn't need to wait and just well merge this one after verifying that search works on old one (iirc there was some test doing the search on canonical ///)

File: datalad/metadata/tests/test_base.py [hunk 2]
```diff
@@ -271,7 +281,7 @@ def assert_names(res, names, path=clone.path):
     )
 
     # more tests on returned paths:
-    assert_names(clone.search('datalad'), ['.', 'sub', 'sub/subsub'])
+    #assert_names(clone.search('datalad'), ['.', 'sub', 'sub/subsub'])
```
yarikoptic: So what they are now since test is disabled?

mih:
Honestly, I don't mind keeping this open for longer. Lots of things need to be discovered. But it makes no sense to keep to meta data PRs open at the same time, when they get closer and closer to each other.

docs/source/metadata.rst
```diff
@@ -110,18 +113,42 @@ JSON-LD meta data format
...
 The following sections describe details and changes in the meta data
 specifications implemented in datalad.
 
+.. _0.2:
+
+v0.2
+----
+
+* Use a datalad-specific JSON-LD context to streamline make further
+  extensions.
```
yarikoptic: making

datalad/metadata/tests/test_base.py
mih: Some, forgot to remove comment. Fixed. Will push soon.

yarikoptic:
pushed a little commit 8189487de6bc707c43dee582b4c440f0a58f8551 to use consts for the metadata version so we could at least in theory do check/warnings

mih:
Thanks.

mih:
Due to popular demand now all meta data related changes in a single PR.

yarikoptic:
Wow. Thanks for our and detailed changelog! I will try to give it a ride today
Is there noteable impact on run time? Or if you try on our mighty dataset

mih:
@yarikoptic thanks. When you try it out, please keep in mind that running with `guess=True` is not supposed to give a one size fits all solution. The idea is to be able to turn things on and off they way you want them to be.

yarikoptic:
On November 5, 2016 9:25:07 AM EDT, Michael Hanke <notifications@github.com> wrote:
>@yarikoptic thanks. When you try it out, please keep in mind that
>running with `guess=True` is not supposed to give a one size fits all
>solution. The idea is to be able to turn things on and off they way you
>want them to be. 
>
>-- 
>You are receiving this because you were mentioned.
>https://github.com/datalad/datalad/pull/1074#issuecomment-258611357

Ideally I don't even want to guess... How could I avoid guessing, Ie  what to do  within datasets we have already? Within those automatically crawled I could easily adjust crawlers to say that for those datasets I need to do x and y. Could you please give me an example eg for openfmri where some are bids and some are not - what types of meta i should enable and where/how as you envision it

mih:
> Ideally I don't even want to guess... How could I avoid guessing, Ie  what to do  within datasets we have already? Within those automatically crawled I could easily adjust crawlers to say that for those datasets I need to do x and y. Could you please give me an example eg for openfmri where some are bids and some are not - what types of meta i should enable and where/how as you envision it

At present we have no way to say "gimme all" without `guess=True`. We can enable parsers one by one via `datalad.metadata.parsers.enable`. And disable individual ones via `datalad.metadata.parsers.disable` even after having guess'ed them. You can set these config vars in the datasets, but you can also set them in your user config to achieve a homogeneous result. Additionally, some parsers support additional config variables to tailor their output. They should be documented in `datalad/interface/common_cfg.py` (`datalad.metadata.` prefix).

I believe it should not hurt to enable parsers that do not actually match a particular dataset. If it does, I'd consider this a bug

Right now I do not want to prescribe to one way to do it. We need more experience how the machinery reacts to a lot of meta data from many parsers. But the configurability is getting better and we should be able to cope with what lies ahead.

mih:
First observations on aggregation performance with all bells and whistles enabled: uuuuuultra slow.

At this point I blame AnnexRepo for doing one git-annex call per file key request (#1076).

mih:
Merged #1076 -- speed up by some largish factor! thanks.

mih:
FTR: Everything enabled yields 71MB of meta data for all openfmri datasets combined until ds0200. That is 7.2 MB gzipped. Biggest chunk comes from ds0115 with 27MB alone. Aggregation took 150s.

mih:
FTR: Current search implementation is incapable of handling this amount of metadata. Caching for 20min with 100% CPU and 1.3 GB resident memory usage. It yield the desired results eventually, and a subsequent invocation has a subsecond runtime, but still, such overhead requires some kind of mitigation.

Pickled cache is 51 MB btw. 11 MB if gzipped.

mih:
With 147k meta data dicts it seems to take on the order of 3 min to shove them into an RDFlib graph (no proper triple store backend).

yarikoptic:
Please have a look at biocaddies use cases ([2.2 Top-down use cases](https://docs.google.com/document/d/1hVcYRleE6-dFfn7qbF9Bv1Ohs1kTF6a8OwWUvoZlDto/edit?usp=sharing))

mih:
OK, time to close. Keeping it open, did not prevent duplication of work in other lines of work. So let's not keep pretending.

PR #1075, opened 2016-10-31
ENH: Extended context from #1074

mih:


yarikoptic:
+1 to merge as soon as tests pass

Issue #1076, opened 2016-11-01
AnnexRepo methods should support list arguments whenever git-annex does too


mih:
Example: `AnnexRepo.get_file_key()`

`lookupkey` can eat a list of files, but in Python I need to have an output loop that invokes git-annex once per file.

PR #1077, opened 2016-11-01
Explore file-based metadata

mih:
At present just playing with toys to explore the problem space. 

Lives on top of #1074 for now.

### DIFF
--- a/datalad/metadata/parsers/fileinfo.py
+++ b/datalad/metadata/parsers/fileinfo.py
@@ -9,7 +9,6 @@
 """Parser for generic file-based information (mime types, etc...)
 """
 
-from datalad.metadata import _get_base_metadata_dict
 from datalad.support.annexrepo import AnnexRepo
 from datalad.metadata.parsers.base import BaseMetadataParser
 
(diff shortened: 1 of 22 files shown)


datalad/metadata/parsers/fileinfo.py
```diff
@@ -0,0 +1,52 @@
...
+        ds_meta = _get_base_metadata_dict(dsid)
+        parts = []
+        if not self.has_metadata():
+            return meta
+        # shortcuts
+        cfg = self.ds.config
+        repo = self.ds.repo
+        cfg_section = self.cfg_section
+        files = repo.get_annexed_files()
+        # TODO RF to do this with one annex call
```
bpoldrack: Note: For this TODO, an issue would be nice, telling that we need AnnexRepo.get_file_key() to work on a list and in "batch mode".

bpoldrack:
Do we consider files in annex only, because they have a key? Or is there another reason to exclude files in git?

datalad/metadata/parsers/fileinfo.py
mih: There is #1076

mih:
> Do we consider files in annex only, because they have a key? Or is there another reason to exclude files in git? 

This is just a simple thing to generate some file-based meta data. Many improvements are possible, but this PR is about figuring out the general issues and principles.

mih:
Before someone is wondering. The "disable some parsers but guess the others" feature is already implemented. Will push when tests are done.

mih:
Those are in already.

mih:
Merging with #1074

PR #1088, opened 2016-11-03
BF+RF+ENH(TST) add (GitRepo and Dataset)

yarikoptic:
- (**change in behavior**) `datalad add` was useless in pure git repositories, unless explicit `--to-git` was added, which to me made little sense. So I have changed it so it does `git add` in pure git repos, instead of returning failure
- GitRepo.add didn't return anything, at least with recent git 2.9.3 unless --verbose option is added
- parsing of the add output was not working correctly for obscure filenames with ' in them
- `datalad add` was returning (I think) incorrect results in case of paths across multiple datasets were provided, since the last lucky `ds` value was used. Since output was not really tested much -- that bug was lingering happily around.  I have added some testing

P.S. now I just need to remember what the heck I was doing that I got to this issue ;)  ah, right -- I was under ~/datalad and was trying to save a bogus file so I could trigger regeneration of metadata cache. but was surprised that command seemingly succeeded but file was not added to git at all.

### DIFF
--- a/datalad/distribution/add.py
+++ b/datalad/distribution/add.py
@@ -14,6 +14,8 @@ import logging
 
 from os import curdir
 from os.path import isdir
+from os.path import join as opj
+from os.path import relpath
 from collections import defaultdict
 
 from datalad.interface.base import Interface
@@ -345,6 +347,10 @@ class Add(Interface):
                     )
             return_values = None  # to avoid mis-use
 
+        # XXX or we could return entire datasets_return_values, could be useful
+        # that way.  But then should be unified with the rest of commands, e.g.
+        # get etc
+        return_values_flat = []
         for dspath, return_values in datasets_return_values.items():
             if save and len(return_values):
                 # we got something added -> save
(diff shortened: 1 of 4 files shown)


yarikoptic:
@datalad/developers any objections regarding change in behavior?

datalad/distribution/add.py
```diff
@@ -336,17 +345,34 @@ def __call__(
...
+            if save and len(return_values):
+                # we got something added -> save
+                # everything we care about at this point should be staged already
+                Save.__call__(
+                    message='[DATALAD] added content',
+                    dataset=ds,
+                    auto_add_changes=False,
+                    recursive=False)
+            # TODO: you feels that this is some common logic we already have somewhere
+            dsrelpath = relpath(dspath, dataset.path)
```
bpoldrack: I share your feeling :)

At least it's used in `get` already (line 156 and following). But basically this is needed whenever we recurse and want to report file paths.

bpoldrack:
Sadly, coverage plugin currently doesn't load anything . Other than that: I agree.

yarikoptic:
Email didn't go out...

Yeap.. reported to codecov, they are on top of it (some github API change)

Since we aware in agreement, will merge

yarikoptic:
On November 3, 2016 11:59:59 AM EDT, Benjamin Poldrack <notifications@github.com> wrote:
>Sadly, coverage plugin currently doesn't load anything . Other than
>that: I agree.
>
>-- 
>You are receiving this because you authored the thread.
>https://github.com/datalad/datalad/pull/1088#issuecomment-258186870

Yeap..  reported to codecov, they are on top of it (some github API change)

Since we aware in agreement, will merge

PR #1091, opened 2016-11-04
RF: Repos

### LINKED ISSUE #1028: Dataset singletons
It is @bpoldrack 's dream to have those.  And flyweight pattern is probably the way to go:
http://zqpythonic.qiniucdn.com/data/20060526140550/index.html#id17 by establishing a weakref directionary, so there will be no guarantee to always get the same dataset (of which you could access thousands in a good session) but at least the same if you still have it referenced somewhere

bpoldrack:
This PR provides some structural changes, especially a singleton approach for the classes `GitRepo`, `AnnexRepo` and `Dataset`.
It doesn't make full use of it or implement everything, that would now be easier, but lays the ground.
Besides some fixes there are three main points:

1. `GitRepo`, `AnnexRepo` and `Dataset` now consistently use `ConfigManager` to read any config item.
It removes the use of any other config reader within the repo classes. This is to achieve several things. We need to be able to read datalad config items anyway, which we can't easily do with GitPython's git config reader for example. It's now consistent how to access any config item across the entire datalad code, which is a value of its own, I think. And we thereby make sure there's no conflict when accessing files and other, possibly outdated, instances providing the same item. ATM it might be a little bit slower, since we don't have a really good strategy yet on when to reload `ConfigManager` (see #1044). But this is to be solved anyway and therefore is no reason to keep the chaos from my point of view.
Note, that `ConfigManager`'s instance is bound to `GitRepo.config`, i.e. `Dataset` uses the `ConfigManager` from its underlying repo.

2. `GitRepo`, `AnnexRepo` and `Dataset` are now flyweights. For the important aspects this means they are basically singletons. ;)
`GitRepo` and therefore `AnnexRepo` are using their `realpath` as the key, which resolves symlinks. So, for any physical repository on the file system, there's only one instance at a time. In opposition `Dataset` uses just the canonical path of what you use to request an instance, meaning it doesn't resolve symlinks and it possibly uses relative paths. Therefore two `Dataset` instances might not be the same while both of them are pointing to the same physical repository and using the same `AnnexRepo`/`GitRepo` instance to access it.
The flyweight pattern is realized by its own metaclass `Flyweight`. See docs within `datalad/support/repo.py ` for further details.
You can use `GitRepo`, `AnnexRepo` and `Dataset` mostly in the very same way as before. What used to be a call of the constructor ( `GitRepo(whatever=option)` for example ) is turned into a call to `Flyweight.__call__` and will return an instance of the class you expect. Still this will need some more work, I think to figure out, what should happen if an instance is requested, that already exists, but with different options to the "constructor" call. But for now this issue is a theoretical one and it should be the topic of another PR. However, if this theoretical conflict becomes a reality, this probably simply points to something not very convincing within the code, using these classes, I think.

3. Superclass
There is a superclass `Repo` now, that is supposed to be a container for methods, which should be available to both repo classes while implemented differently. This may sound strange at first, since this could be done by adding them to `GitRepo` and override them in `AnnexRepo`. But:
As @mih  pointet out in issue #1100  there is a need for things like `file_has_content` to be implemented for `GitRepo` as well, while their sheer existence requires an idea of annex. So, from my point of view it would be strange to have them defined in `GitRepo`. While it technically isn't necessary, I think it's the right thing to collect those things at a single place. `Repo` is that place.

While this PR doesn't change much on how to use these classes, I need you guys, @mih , @yarikoptic , to have a close look at the concept and to be aware of how it works. Some tests for example used to implicitly rely on those classes NOT being singletons, reinstanciating them over and over again.

Finally, I'm aware there have to be more and/or better tests, but right now, it's more important to get this into master ASAP, since it becomes more inconvenient to make use of this with every additional commit to master. We need to base the logic, we implement elsewhere, on that concept and harvest the benefit of these "singletons". PR with another general change to `AnnexRepo` is coming and will provide better coverage.

- Closes #1028

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ language: python
 
 python:
   - 2.7
-  - 3.3
   - 3.4
 
 cache:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ This is a very high level and scarce summary of the changes between releases.
 We would recommend to consult log of the [DataLad git repository](http://github.com/datalad/datalad)
 for more details ATM.
 
-## 0.4.1 (Oct ??, 2016) -- will be better than ever
+## 0.4.2 (Nov ??, 2016) -- will be better than ever
 
 ???
 
@@ -20,6 +20,37 @@ for more details ATM.
 ### Enhancements and new features
 
 ???
+
+
+## 0.4.1 (Nov 10, 2016) -- CA release
+
+Requires now GitPython >= 2.1.0
+
+### Fixes
+
+- [save]
+     - to not save staged files if explicit paths were provided
+- improved (but not yet complete) support for direct mode
+- [update] to not crash if some sub-datasets are not installed
+- do not log calls to `git config` to avoid leakage of possibly 
+  sensitive settings to the logs
+
+### Enhancements and new features
+
+- New [rfc822-compliant metadata] format
+- [save]
+    - -S to save the change also within all super-datasets
+- [add] now has progress-bar reporting
+- [create-sibling-github] to create a :term:`sibling` of a dataset on
+  github
+- [OpenfMRI] crawler and datasets were enriched with URLs to separate
+  files where also available from openfmri s3 bucket
+  (if upgrading your datalad datasets, you might need to run
+  `git annex enableremote datalad` to make them available)
+- various enhancements to log messages
+- web interface
+    - populates "install" box first thus making UX better over slower
+      connections
 
 
 ## 0.4 (Oct 22, 2016) -- Paris is waiting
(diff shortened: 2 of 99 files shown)


yarikoptic:
ok -- that fix didn't provide ultimate remedy.  But I do think it is that ConfigManager just keeping some files or pipes open or smth like that. Also note that on dockers there are complaints about too many open files which I am not sure if not a new side-effect here.  I will poke around more a bit more...

yarikoptic:
that is where all those fids etc go:
```shell
$> px | grep git.*cat | nl | tail
   307	yoh      28739  0.0  0.0  15100  3060 pts/38   S+   13:27   0:00  |       \_ git -c receive.autogc=0 -c gc.auto=0 cat-file --batch
   308	yoh      28954  0.0  0.0  15100  3124 pts/38   S+   13:27   0:00  |       \_ git -c receive.autogc=0 -c gc.auto=0 cat-file --batch-check
   309	yoh      28961  0.0  0.0  15100  3060 pts/38   S+   13:27   0:00  |       \_ git -c receive.autogc=0 -c gc.auto=0 cat-file --batch
   310	yoh      28976  0.0  0.0  15100  3168 pts/38   S+   13:27   0:00  |       \_ git -c receive.autogc=0 -c gc.auto=0 cat-file --batch-check
   311	yoh      28977  0.0  0.0  15100  3132 pts/38   S+   13:27   0:00  |       \_ git -c receive.autogc=0 -c gc.auto=0 cat-file --batch
   312	yoh      29160  0.0  0.0  15100  3088 pts/38   S+   13:27   0:00  |       \_ git -c receive.autogc=0 -c gc.auto=0 cat-file --batch-check
   313	yoh      29161  0.0  0.0  15100  3140 pts/38   S+   13:27   0:00  |       \_ git -c receive.autogc=0 -c gc.auto=0 cat-file --batch
   314	yoh      29165  0.0  0.0  15100  3084 pts/38   S+   13:27   0:00  |       \_ git -c receive.autogc=0 -c gc.auto=0 cat-file --batch-check
   315	yoh      29166  0.0  0.0  15100  3036 pts/38   S+   13:27   0:00  |       \_ git -c receive.autogc=0 -c gc.auto=0 cat-file --batch
   316	yoh      29691  0.0  0.0  12708   488 pts/33   S+   13:27   0:00      \_ grep --color=auto -d skip git.*cat
```
so -- I think it is the singleton repos, which do not die off, thus still occupy memory and all their attributes, such as batched git-annex'es lurking in the background.

What is the idea here -- when singleton repo should die off?  whenever it is not bound to any non gc-ed variable?  also it would be nice if comments stayed in english so I could understand them as well

bpoldrack:
FTR: Current build failures happend during setup due to an issue on neurodebian.ovgu.de

bpoldrack:
I agree: #1169 should be merged first, so I can care for the conflicts.

bpoldrack:
@yarikoptic : nd80 failure during `pip install`. Nevertheless, actual tests were running and passed.

mih:
@bpoldrack  can you please force-push to bounce the tests. Given the potential for disaster, it would be good to see them all pass.

mih:
OK, fingers crossed.

PR #1094, opened 2016-11-04
ENH openfmri - crawl also extracted datasets where available

yarikoptic:
Includes a number of enhancements around in the code


yarikoptic:
no objections so merging

PR #1097, opened 2016-11-07
ENH: Allow to look up keys for lists of files (Closes #1076)

### LINKED ISSUE #1076: AnnexRepo methods should support list arguments whenever git-annex does too

bpoldrack:
This is a quickfix for what @mih requested.

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -736,57 +736,69 @@ class AnnexRepo(GitRepo):
                                        annex_options=['--'] + git_cmd,
                                        **kwargs)
 
-    @normalize_path
-    def get_file_key(self, file_):
+    @normalize_paths
+    def get_file_key(self, files):
         """Get key of an annexed file.
 
         Parameters
         ----------
-        file_: str
-            file to look up
+        files: str or list
+            file(s) to look up
 
         Returns
         -------
-        str
-            keys used by git-annex for each of the files
+        str or list
+            keys used by git-annex for each of the files;
+            in case of a list an empty string is returned if there was no key
+            for that file
         """
 
-        cmd_str = 'git annex lookupkey %s' % file_  # have a string for messages
+        if len(files) > 1:
+            return self._batched.get('lookupkey',
+                                     git_options=self._GIT_COMMON_OPTIONS,
+                                     path=self.path)(files)
+        else:
+            files = files[0]
+            # single file
+            # keep current implementation
+            # TODO: This should change, but involves more RF'ing and an
+            # alternative regarding FileNotInAnnexError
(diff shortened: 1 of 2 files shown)


datalad/support/annexrepo.py
```diff
@@ -736,57 +736,69 @@ def proxy(self, git_cmd, **kwargs):
...
-        str
-            keys used by git-annex for each of the files
+        str or list
+            keys used by git-annex for each of the files;
+            in case of a list an empty string is returned if there was no key
+            for that file
         """
 
-        cmd_str = 'git annex lookupkey %s' % file_  # have a string for messages
+        if len(files) > 1:
```
yarikoptic: In all other cases there is a switch option for using batched mode, and it is never decided by the number of files. If I see it right you would get now completely different behavior in how special cases reported back. At least with nothing in docstring

datalad/support/annexrepo.py
```diff
@@ -736,57 +736,69 @@ def proxy(self, git_cmd, **kwargs):
...
-                                         filename=file_)
+            try:
+                out, err = self._run_annex_command('lookupkey',
+                                                   annex_options=[files],
+                                                   expect_fail=True)
+            except CommandError as e:
+                if e.code == 1:
+                    if not exists(opj(self.path, files)):
+                        raise IOError(e.code, "File not found.", files)
+                    # XXX you don't like me because I can be real slow!
```
yarikoptic: If that is what was causing Michael slowdown, it means that he was going the special cases handling... How those are reported in batched mode?

yarikoptic: Aha, they are reported by empty string. So differently. I just wonder if Michael would ever pass a list with a single item causing bon batched mode handling and would be up for a surprise
 So IMHO batched mode must be an option not an internal decision

bpoldrack:
The thing with an optional batch mode is, that `annex lookupkey` on a list will just abort on the first item not in annex. So without batch mode but allowing for a list will result in not so nice behaviour. We could raise in that case, but I think to have a result for each item passed is better.

Raising on single items of a list is something I don't like. It does not allow the caller to deal with a partially invalid list. So, as I mentioned in the comment at the beginning - the way we report in the single item case should change to be consistent.

Agree?

datalad/support/annexrepo.py
bpoldrack: FTR: I don't think this was slowing down Michael. He probably just noticed `get_indexed_files()` when looking into that function and left a note ;-)
I'm pretty sure, the slowing down was caused by all the unnecessary `annex` calls.

PR #1103, opened 2016-11-09
BF: make sure custom command versions are comparable versions

mih:


yarikoptic:
A bit confused on what this PR accomplishes besides allowing to give a version value itself I guess, may be because unittest didn't test new features behavior? (I only see the check if the class, which I guess was in effect before as well, or am I wrong?)

mih:
Make sure _custom command versions_ are comparable versions. Which was not the case and all such comparisons in the code were string comparisons and lead to faulty behavior.

Issue #1107, opened 2016-11-11
`update` assumes configured remote -- it should not


mih:
This is particularly painful when I want to update a tree of datasets recursively, but the top-level one has no remote... fails right at the start.

````
$ git remote -v
$ datalad update
[INFO   ] Updating dataset '.' ... 
[ERROR  ] 'NoneType' object has no attribute 'name' [gitrepo.py:fetch:1214] (AttributeError) 
$ datalad --dbg update
[INFO   ] Updating dataset '.' ... 
Traceback (most recent call last):
....
  File "/home/mih/datalad/datalad/distribution/update.py", line 127, in __call__
    repo.fetch(remote=name, all_=fetch_all)
  File "/home/mih/datalad/datalad/support/gitrepo.py", line 1214, in fetch
    tb = self.repo.active_branch.tracking_branch().name
AttributeError: 'NoneType' object has no attribute 'name'
()
> /home/mih/datalad/datalad/support/gitrepo.py(1214)fetch()
-> tb = self.repo.active_branch.tracking_branch().name
(Pdb) 
````

PR #1108, opened 2016-11-11
BF: Help `update` cope with dataset with no remotes (fixes gh-1107)

mih:


mih:
Ok, things could be done even better. Agreed. In any case, this quick fix makes it possible to use update at all in these cases. More urgent than ultimate beauty.

PR #1113, opened 2016-11-11
RF: Higher-level abstraction helper for command that do thing to dataset content items

mih:
For example `get` and `unlock` are highly similar in terms of what they should be doing to figure out which items to `get` or to `unlock`. Yet they use hardly the same lines of code.

Will try to identify and RF some of the common bits into more generic helpers.

Feel free to merge any time, despite possibly open TODOs. I do not intent to work on this for two months. Only low hanging fruit.

- [x] `unlock` can work recursively and across multiple disjoint datasets -- like `get`
- [x] `get` and `unlock` use more of the same codebase
- [x] common path input argument normalization
- [X] decide whether to allow `get` to work with an empty path argument (i.e. "get this") -> #1119

### DIFF
--- a/datalad/distribution/get.py
+++ b/datalad/distribution/get.py
@@ -20,6 +20,7 @@ from os.path import relpath
 
 from datalad.interface.base import Interface
 from datalad.interface.utils import get_paths_by_dataset
+from datalad.interface.utils import get_normalized_path_arguments
 from datalad.interface.common_opts import recursion_flag
 from datalad.interface.common_opts import git_opts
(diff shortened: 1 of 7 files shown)


mih:
If this gets no negative feedback, it is doable to apply the same (or similar) paradigm to `uninstall`. This should help to address some of the oddities of `uninstall` that were recently reported.

mih:
#1118 now anticipates that this will see a soonish merge.

mih:
Will merge in 2 hours.

PR #1121, opened 2016-11-16
ENH: INDI, s3 etc

yarikoptic:
Was working to provide crawling of indi datasets... still somewhat WiP.  Submitting to test if our buildbots came back up ok after power outage


mih:
My understanding is still limited, but I see nothing that looks like a showstopper. Feel free to merge.

PR #1123, opened 2016-11-18
ENH: better use recentish tqdm and then no need for custom options

yarikoptic:
Discovered during demos that global pbar did not behave properly.  There seems to be
a number of issues within older tqdm and fresh one addresses them.  So we just need a
fresh tqdm, but I do not think we need to hard demand it until it is present everywhere

I have submitted request to have pkg updated in debian, after we get it there and in neurodebian -- we will boost versioned dependency

Paired with a fix (workaround) for a problem resolving domains with underscores in recent requests  (https://github.com/kennethreitz/requests/issues/3683#issuecomment-261947670), fix for which is WiP


yarikoptic:
woohoo -- all tests "pass" again, so if no objections I will merge it within a few hours so we could get master back in a good shape

PR #1125, opened 2016-11-22
BF: ls does output all original boto constructs back now for s3:// urls

yarikoptic:


yarikoptic:
minor, thus just merging

PR #1126, opened 2016-11-22
ENH: BIDS -- read README as the description if none provided in dataset_description.json

yarikoptic:
as the bible says!  well -- somewhat.

Here I read README only if original Description is not provided.  BIDS somewhat says "in addition".
So I wondered if some alternative solution should be used, e.g.

1. append content README to description even if description was provided (there are a few openfmri datasets which have both, README and description;  more datasets have no descriptions but do have README)
2. add a new field (e.g. `description-long`) to our schema to contain content of README.

What do you think @mih ?


mih:
FTR: I looked at it, and I have no objections. But I will not merge it, because I will have to resolve all the conflicts too ;-)

PR #1134, opened 2016-11-25
BF: Direct Mode 

### LINKED ISSUE #1131: GitRepo.add_submodule() is incompatible with direct mode
In its core it calls `git submodule`, which does not work:

````
Failed to run ['git', '-c', 'receive.autogc=0', '-c', 'gc.auto=0', 'submodule', 'add', '--name', 'some/what/deeper', './some/what/deeper', './some/what/deeper'] under '/tmp/datalad_temp_test_create_subnjhdlspb'. Exit code=1. out= err=fatal: /usr/lib/git-core/git-submodule cannot be used without a working tree.
````

bpoldrack:
Make sure, `AnnexRepo.__init__()` is aware, whenever a repo was created in direct mode without this being specified in its parameters.

No tests, since they are already there ;)

Closes #1131 , I think.

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -236,12 +236,21 @@ class AnnexRepo(GitRepo):
             else:
                 raise RuntimeError("No annex found at %s." % self.path)
 
+        self._direct_mode = None  # we don't know yet
+
+        # If we are in direct mode already, we need to make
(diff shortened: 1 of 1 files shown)


mih:
Doesn't like it....

mih:
In this context #1143 is relevant.

PR #1122, opened 2016-11-25
datalad ls should test for uninstalled subdatasets

yarikoptic:
While making sure that uninstall replaces deinited subdatasets with an empty dir I noticed this:

````
/tmp/some/deep (git)-[master] % datalad ls -r
.  unknown
/tmp/some/deep (git)-[master] % cd ..
/tmp/some (git)-[master] % datalad ls -r
.      [annex]  master  - 2016-11-17/17:42:41  OK
deep   not installed
````

In the first case it doesn't even acknowledge to be in a dataset, and gives the same output as for a directory that is not even under git. I think it should give some indication of being in a dataset (minimum), and better say that it is at an uninstalled subdataset.

### DIFF
--- a/datalad/interface/ls.py
+++ b/datalad/interface/ls.py
@@ -15,6 +15,7 @@ import sys
 import time
 from os.path import exists, lexists, join as opj, abspath, isabs, getmtime
 from os.path import curdir, isfile, islink, isdir, dirname, basename, split, realpath
+from os.path import relpath
 from os import listdir, lstat, remove, makedirs
 import json as js
 import hashlib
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by yarikoptic** [comment]
woohoo -- `git hub pull attach 1122` worked!

So -- **disclaimer**:  `ls` needs to go through heavy overhaul for its 'interactive' use sooner or later.  So I just did quick and dirty fix for now, which should work, e.g.:
```shell
(git)hopa:~/datalad/openfmri[master]
$> datalad ls 123* ds00000[12]
123  dir
123.txt  file
ds000001  not installed
ds000002   [annex]  master  2.0.1-19-gaf97459 2016-11-04/17:18:02  OK

$> datalad -C ds000001 ls .
.  not installed
```

mih:
Your fix smells like it wants to use RF'ed code from interface.utils. Something to keep in mind for the coming bigger RF...

mih:
In any case, feel free to merge when travis is done.

yarikoptic:
what portion of interface.utils do you see it using? `get_paths_by_dataset` ?

PR #1156, opened 2016-12-03
Bunch of helpers and RF towards more uniform command implementations

mih:
Practically salvaging pieces from my failed attempt to RF `save`.

### DIFF
--- a/datalad/distribution/uninstall.py
+++ b/datalad/distribution/uninstall.py
@@ -19,21 +19,18 @@ from os.path import curdir
 from os.path import exists
 from os.path import relpath
 from os.path import join as opj
-from datalad.support.exceptions import InsufficientArgumentsError
 from datalad.support.param import Parameter
 from datalad.support.constraints import EnsureStr, EnsureNone
 from datalad.distribution.dataset import Dataset, EnsureDataset, \
     datasetmethod, require_dataset
 from datalad.interface.base import Interface
+from datalad.interface.base import report_result_objects
 from datalad.interface.common_opts import if_dirty_opt
 from datalad.interface.common_opts import recursion_flag
 from datalad.interface.common_opts import recursion_limit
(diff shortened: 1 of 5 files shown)


datalad/distribution/uninstall.py
```diff
@@ -169,7 +117,6 @@ class Drop(_CommandBase):
 
     """
     _action = 'drop'
-    _passive = 'dropped'
```
yarikoptic: Note that over Travis rub fails seems due to use of this one somewhere

datalad/interface/base.py
```diff
@@ -304,3 +308,53 @@ def call_from_parser(cls, args):
...
+        return
+    msg = "{n} {obj} {action}:\n".format(
+        obj='items were' if len(res) > 1 else 'item was',
+        n=len(res),
+        action=cls._passive)
+    for item in res:
+        if isinstance(item, Dataset):
+            msg += "Dataset: %s\n" % item.path
+        else:
+            msg += "File: %s\n" % item
```
yarikoptic: How to extend for other types of objects and reports? Eg in case of crawl and ls?

datalad/interface/base.py
```diff
@@ -304,3 +308,53 @@ def call_from_parser(cls, args):
...
             ui.error("\nInterrupted by user while doing magic: %s" % exc_str(exc))
             sys.exit(1)
+
+    @staticmethod
+    def _prep(
+            path=None,
+            dataset=None,
+            recursive=False,
+            recursion_limit=None,
+            mark_recursive=False):
```
yarikoptic: Could you please add some basic doc to describe its purpose?

mih:
Ah, sorry -- didn't realized how many depths remain in unpushed commits -- not worth it right now. Will close.

PR #1027, opened 2016-12-04
add --to-git pass explicitly -o annex.largefiles=

yarikoptic:
if annex v6 is used, then `git add` will pretty much do `git annex add`, so we need say explicitly to not commit to annex for that `git add` call

### DIFF
--- a/datalad/distribution/tests/test_add.py
+++ b/datalad/distribution/tests/test_add.py
@@ -126,7 +126,9 @@ def test_add_recursive(path):
 
     added2 = ds.add(opj('dir', 'testindir2'), recursive=True, to_git=True)
     # added to git, so parsed git output record
-    eq_(added2, [{'success': True, 'file': _path_('dir/testindir2')}])
+    eq_(added2, [{'file': _path_('dir/testindir2'), 'command': u'add',
+                  'note': u'non-large file; adding content to git repository',
+                  'success': True}])
     assert_in('testindir2', Dataset(opj(path, 'dir')).repo.get_indexed_files())
 
     # We used to fail to add to pure git repository, but now it should all be
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -614,7 +614,7 @@ class AnnexRepo(GitRepo):
         return expected_downloads, fetch_files
 
     @normalize_paths
-    def add(self, files, git=False, backend=None, options=None, commit=False,
+    def add(self, files, git=None, backend=None, options=None, commit=False,
             msg=None, dry_run=False,
             jobs=None,
             git_options=None, annex_options=None, _datalad_msg=False):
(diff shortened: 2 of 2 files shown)


yarikoptic:
gy -- was my bad -- should have read at least the name of the test to grasp situation better.  Pushing the "fix" to see how it reflects across all the tests

yarikoptic:
Hm, unfortunately my approach of just using `git annex add` all the time doesn't work ATM -- annex simply ignores (doesn't do anything) for files under git control:

```shell
$> git status
On branch master

	modified:   newfiletest.dat

Untracked files:

	123.dat

# annex just ignores file under git  control and only deals with a new file
# which goes into git
$> git annex add -c 'annex.largefiles=nothing' --json .
{"command":"add","note":"non-large file; adding content to git repository","success":true,"file":"123.dat"}

$> git status
On branch master
Changes to be committed:

	new file:   123.dat

	modified:   newfiletest.dat

```
and then if we introduce modifications to file annex added to git -- again ignores it
```shell
$> echo 124 >| 123.dat; git status 
On branch master
Changes to be committed:

	new file:   123.dat

	modified:   123.dat
	modified:   newfiletest.dat

$> git annex add -c 'annex.largefiles=nothing' --json .
```

And then, we can't use just "git add" even in new v6 mode since then we loose all the --json --batch interactions annex provides us with and we rely upon.   But we would need some centralized point of entry for 'add'  e.g. in case of mixed repo where some files are under git some under annex (v6 or not).
@joeyh -- what do you think?  could may be 'annex add' also handle adding to 'git' regular/under git control files?  may be optionally (`--handle-git` or alike)) so not to break compatibility if so desired.

joeyh:
I worry a bit about `git annex add` slipping toward a complete reimplementation of `git add`, which is not its intent. However, it already staged modifications to non-large files in v6 mode and indirect mode, and its man page seems to say it will add modified files, so let's make it always do so.

mih:
Great, that will simplify the implementation of datalad's `save` by quite a bit.

yarikoptic:
Hi @joeyh , please buzz whenever this feature/functionality hits git-annex

joeyh:
Yaroslav Halchenko wrote:
> Hi @joeyh , please buzz whenever this feature/functionality hits git-annex

This landed in f61508aed47bb915529f35f2281407c0fc613dfe / 6.20161210

-- 
see shy jo

yarikoptic:
Thank you @joeyh .  NeuroDebian now carries  6.20161211+gitgc3ab3c6 so we will give it a shot!

yarikoptic:
I thought that there is one additional test to fix, but may be that was it.
I expect dockers to fail since they don't have up to date annex.  So if travis passes, I will upgrade dockers (which would render other PRs to fail) so we could quickly finalize this PR and get it merged to get other PRs "off the hook".  Thus quick review would be appreciated @datalad/developers

yarikoptic:
ok -- annexes updated all around, tests pass all around -- merging

Issue #1143, opened 2016-11-26
Switch everything to v6 repos ASAP


mih:
I played with git annex on windows for a while, and it led me back to the realization that we need to switch all logic to v6 mode repositories ASAP. The possibility to have mixtures of locked an unlocked files, the option to have adjusted branch whose content needs to get `annex sync`'ed at some point put an additional level of complexity on the program logic -- which we are completely ignoring now. I think the longer we wait with this switch, the more we have to redo again.

For the record, here are a few pointers with essential information:

- https://git-annex.branchable.com/tips/unlocked_files/
- https://git-annex.branchable.com/git-annex-adjust/
- https://git-annex.branchable.com/git-annex-sync/

We already have a few bugs with infos and issues:

https://github.com/datalad/datalad/issues?q=v6+label%3Aannex-v6

mih:
I am increasingly under the impression that we should ship v6 repos in unlocked state from datasets.datalad.org.

FTR: https://git-annex.branchable.com/forum/converting_a_v5_repo_to_an_always_unlocked_v6_repo/

At the same time we need to expose `annex.thin` via `install`.

mih:
FTR: In v6 unlocked repo mode (and in any direct mode repository too!), we need to run:

````
git annex sync --no-push --no-pull
````

before we perform any checks on whether "master" is of a particular state. The above command syncs all potential changes in the "adjusted" (or "annex/direct/...." branch into the actual master -- without talking to remotes.

mih:
Still playing with it to get some real life impression. I am doing this (https://git-annex.branchable.com/forum/converting_a_v5_repo_to_an_always_unlocked_v6_repo/) with a single repo (40k files with some 700GB of data). The `sync` part is taking ages. It passes all files individually through the clean smudge filter on commit -- we are talking hours of runtime.

Here is an impression of the commands running during the commit stage:

````
git-annex smudge --clean 'SOMEFILE'
git --git-dir=.git --work-tree=. --literal-pathspecs check-attr -z --stdin annex.backend annex.numcopies annex.largefiles --
git --git-dir=.git --work-tree=. --literal-pathspecs cat-file --batch
git --git-dir=.git --work-tree=. --literal-pathspecs cat-file --batch-check=%(objectname) %(objecttype) %(objectsize)
````

yarikoptic:
Iirc 40k files is already a notable burden even for regular git  manipulations
-- 
Sent from a phone which beats iPhone.

[...]

PR #1158, opened 2016-12-05
ENH: Annex V6 default build added to Travis

bpoldrack:
Addressing #1143 as agreed on in hangout.

Should add builds with V6 as default.
When we started having a look at these, should change to V6 for all builds instead of dedicated ones.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,7 @@ env:
   matrix:
     - DATALAD_REPO_DIRECT=1
     - DATALAD_REPO_DIRECT=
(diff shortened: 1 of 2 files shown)


mih:
Ha, I didn't even realize we had that switch already. It is not in `datalad/interface/common_cfg.py`. I wonder how much more I am missing...

Anyways, the errors promise to provide some entertainment for the winter time!

Thanks.

bpoldrack:
My bad - introduced it with the direct mode switch, but forgot to add it to `common_cfg`.

Added both of them now.

Issue #444, opened 2016-05-20
Need for housekeeping command?


mih:
Datalad downloads tarballs in the background and caches them. This could lead to situations where a user want to know how much storage is needednfor the cache and maybe empty it.

Should we have a housekeeping command that can provide this info and perform cleanups?

Maybe even can kill cached files that are no longer needed (determined by some measure)?

yarikoptic:
There are even multiple levels of what could be considered a cache here..  we should indeed discuss, please tag. Meanwhile, there is clean command already iirc.
-- 
Sent from a phone which beats iPhone.

hanke:
Yes, there is `clean`, but nothing beyond that.

mih:
Superseded by #1371.

PR #1160, opened 2016-12-05
RF+ENH: clean to clean up .git/annex/tmp, do recursive traversal, use -d

yarikoptic:
slightly related to https://github.com/datalad/datalad/issues/444

### DIFF
--- a/datalad/interface/clean.py
+++ b/datalad/interface/clean.py
@@ -19,6 +19,15 @@ from ..support.param import Parameter
 from ..support.annexrepo import AnnexRepo
 from ..cmdline.helpers import get_repo_instance
 from ..consts import ARCHIVES_TEMP_DIR
+from ..consts import ANNEX_TEMP_DIR
+
+from datalad.support.constraints import EnsureNone
+from datalad.distribution.dataset import Dataset
+from datalad.distribution.dataset import EnsureDataset
+from datalad.distribution.dataset import require_dataset
+from datalad.distribution.dataset import datasetmethod
+from datalad.interface.common_opts import recursion_flag
+from datalad.interface.common_opts import recursion_limit
 
 from logging import getLogger
 lgr = getLogger('datalad.api.clean')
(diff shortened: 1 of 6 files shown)


datalad/interface/clean.py
```diff
@@ -35,30 +44,64 @@ class Clean(Interface):
...
+                in current working directory""",
+            constraints=EnsureDataset() | EnsureNone()),
         # TODO:  --info  -- which performs dry run just summarizing what is to be cleaned up
         # TODO: Python only???
-        annex=Parameter(
-            doc="""annex instance to use"""
-            #constraints=EnsureStr() | EnsureNone()
-        ),
+        what=Parameter(
+            args=("--what",),
```
yarikoptic: better name would be welcome

datalad/interface/clean.py
```diff
@@ -35,30 +44,64 @@ class Clean(Interface):
     """
 
     _params_ = dict(
+        dataset=Parameter(
+            args=("-d", "--dataset"),
+            # TODO: this probably changes to install into the dataset (
```
bpoldrack: Don't copy&paste TODOs/discussions from `install` ... ;-)

datalad/interface/clean.py
```diff
@@ -35,30 +44,64 @@ class Clean(Interface):
...
+             "temporary annex", ("file", "files")),
+        ]:
+            lgr.info("Considering to clean %s:%s", dataset, dirpath)
+            if not ((what is None) or (flag in what)):
+                continue
+            topdir = opj(dataset.path, dirpath)
+            paths = glob(opj(topdir, '*'))
+            if paths:
+                pl = len(paths) > 1
+                pwd = getpwd()
```
bpoldrack: I doubt, that it makes sense to use `pwd` here. What if I call it via python API ( `Dataset('some/where').clean(...)` )? Why would pwd matter?

yarikoptic: old code I forgot to RF since it provided sensible output ;)
I guess it is worth discussing what our UI should strive to output -- I often get overwhelmed with full paths polluting my terminal while working within some reasonably lengthy path and already knowing that I am in that bloody dataset.  So here it was providing output relative to curdir since that is what you expect to get most of the time (e.g. when you do `ls file` you don't get a full path to the file). and if was outside of that path -- would have provided a full path.  `datalad ls` also does it to some degree... but indeed we might want to strive to unify/optimize display of such output to the user

mih:
I would prefer if we tried to replicate the pattern of input structure and preprocessing implemented in `uninstall` and friends. I.e. use the helper functions for sorting any number of inputs upfront and act upon the detected dataset.

yarikoptic:
`_CommandBase` approach seems to be sensible indeed.  but I guess then we should better RF not only clean (which ATM doesn't even take paths) but other commands as well?

mih:
Of course, but this is not always easy. I don't think all at once is feasible.

yarikoptic:
sure, but if not **all** then **some** and it would make more sense then (`>` just clean, which has no paths argument atm, and if it would -- only paths to datasets would make sense) anyways ;)

yarikoptic:
I will merge this one meanwhile and https://github.com/datalad/datalad/issues/1161 for the rest

PR #1165, opened 2016-12-09
WIP: ENH: Physionet dataset crawler

debanjum:


yarikoptic:
- did you crawl it already somewhere on smaug to have a look?
- create at least a syntax/smoke test that pipelines load fine.  check others like test_balsa IIRC for hints on how to do so

yarikoptic:
moved to https://github.com/datalad/datalad-crawler/pull/16

PR #1185, opened 2017-01-16
Homogenization of `update` API

mih:
Update to bring `update` more in line with the more recently refurbished commands.

- [x] Unify with other commands and say "what to update" rather than "from where" as the primary argument(s)

### DIFF
--- a/datalad/distribution/update.py
+++ b/datalad/distribution/update.py
@@ -23,7 +23,6 @@ from datalad.support.gitrepo import GitRepo
 from datalad.support.param import Parameter
 from datalad.utils import knows_annex
 
-from .dataset import Dataset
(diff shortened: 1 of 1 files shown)


datalad/distribution/update.py
```diff
@@ -35,74 +36,77 @@ class Update(Interface):
...
 
     """
+    # TODO: adjust docs to say:
+    # - update from just one sibling at a time
 
     _params_ = dict(
+        path=Parameter(
+            args=("path",),
+            metavar="PATH",
+            doc="path/name of the component to be dropped",
```
bpoldrack: c&p?
"the component to be dropped" ... That's surprising. ;-)

PR #1169, opened 2016-12-21
RF save/add/create/get

mih:
High-level summary of the desired changes -- after staring two weeks at the code, trying to come up with something that makes more sense, is easier to test, and minimizes the places in the code where similar things happen:

- `save` will no longer do half of the work of `add` -- it will only operate on content that was "added" by other means before. that being said, once the clean version is implemented we can and should reevaluate the scope again
- `save` will no longer do half of the work of `uninstall` and `remove`. only paths to things that presently exist on the file system will be dealt with -- no convenience deinit of vanished submodules, no (possibly dangerous) `git rm` of vanished files anymore. maybe there will be a `fix-this-mess` command later on, and maybe it will triggered by `save`, but clean implementation first
- "surgical" save: only path that we given explicitly (or implicitely via something like `all_changes=True` will be saved. Even if other changes were staged already.
- three layers of functionality are envisioned:
   1. toplevel API: throw it any number of paths in any number of datasets in any order and it will do the right thing with the respective datasets (flags will enable modes such as "also save this one in the superdataset", or "record all changes to known components")
   2. intermediate: give it a dictionary with datasets to touch and the paths to safe in them, and it will figure out the order of processing, discover possibly required intermediate datasets to handle in order to fulfill a request and do the right thing
   3. low: given a single dataset and a list of paths to save, it will save that content in this single dataset -- regardless of the type of content (file, directory, subdataset state)

To be addressed issues:
- [x] #1175 (parts of)
- [x] #1174 
- [x] #1173 
- [x] #1172 
- [x] #1141 
- [x] #1140 
- [x] #1109 
- [x] #1106 (50% of it)
- [x] #1096 
- [x] #1016 
- [x] #1004 

Also contains:

- [x] #1119

PR #1189, opened 2017-01-17
ASV based  benchmarking of DataLad

yarikoptic:
Current state was published to gh-pages: http://datalad.github.io/datalad/
Left smaug running them for ALL commits... not sure if a good idea since currently we have 100 commits nicely sampling previous history and can easily rerun more new benchmarks for them using "EXISTING" argument to "asv run"

TODOs (if we decide to use ASV)
- [x] add .asv submodule with the results (for now just local + smaug clones)
- [ ] automate running benchmarks as the part of the buildbot process (on master, not on PRs to not pull in junk) and pushing results
- [ ] point docs to gh-pages
???

Future TODOs
- Create more benchmarks for basic operations (install, add, get, ...)
- Create few environments with different git annexes. Current just whatever annex is installed on my laptop and smaug. But for these sample benchmarks doesn't matter anyways


mih:
Very cool, very useful. Feel free to merge when tests pass.

I don't get what you mean by "add .asv submodule with the results" -- I guess I know too little about what ASV can do.

yarikoptic:
Re "add .asv submodule with the results"  -- it is there now ;) https://github.com/datalad/.asv

mih:
Failing test is the Git version issue that I reported in #1169 -- if #1169 is merged this test is relaxed.

bpoldrack:
Nice!

From looking at the results, I wonder whether there is a way to include not just `cpu` but the number of cores/CPUs ...

yarikoptic:
On January 18, 2017 8:20:20 AM EST, Benjamin Poldrack <notifications@github.com> wrote:
>Nice!
>
>From looking at the results, I wonder whether there is a way to include
>not just `cpu` but the number of cores/CPUs ... 
>
>-- 
>You are receiving this because you authored the thread.
>https://github.com/datalad/datalad/pull/1189#issuecomment-273473221

Do you mean into description of the machine?

bpoldrack:
Yes!

yarikoptic:
ok -- let me merge it as is for now so you could play with it and add new exciting benchmarks. Then we will put it into automation

Issue #1187, opened 2017-01-16
shipped within Debian package manpages are formatted for wider terminal


yarikoptic:
```shell
$> man datalad | cat
<standard input>:4: warning [p 1, 1.3i]: cannot adjust line
<standard input>:93: warning [p 2, 0.7i]: cannot adjust line
<standard input>:93: warning [p 2, 0.8i]: cannot adjust line
datalad(1)                  General Commands Manual                 datalad(1)

SYNOPSIS
        datalad [-h] [-l LEVEL] [-p {condor}] [--version] [--dbg] [--idbg] [-C
       PATH]
               {search,aggregate-metadata,create,install,get,add,publish,unin‐
       stall,update,create-sibling,create-sibling-github,add-sibling,rewrite-
       urls,unlock,save,export,test,crawl,crawl-init,ls,clean,add-archive-con‐
       tent,download-url,create-test-dataset}
               ...

DESCRIPTION
       DataLad  provides  a  unified data distribution with the convenience of
       git-annex
       repositories as a backend.  DataLad command line tools allow to manipu‐
       late
...
```

mih:
Hmm, didn't know one could format for a specific width. It sounds more like the content cannot be adjusted for any normal width. Maybe if we hack the list of commands to use commas with spaces as separators it would be able to?

PR #1190, opened 2017-01-17
BF: Nicer manpages (fixes gh-1187)

mih:
Could be done in many different ways, but this does it in a minimal way

### DIFF
--- a/datalad/cmdline/helpers.py
+++ b/datalad/cmdline/helpers.py
@@ -61,7 +61,7 @@ class HelpAction(argparse.Action):
             helpstr = parser.format_help()
         # better for help2man
         # for main command -- should be different sections. And since we are in
-        # heavy output messaging mode...
+        # heavy output massaging mode...
         if "commands for dataset operations" in helpstr.lower():
             opt_args_str = '*Global options*'
             pos_args_str = '*Commands*'
--- a/datalad/cmdline/main.py
+++ b/datalad/cmdline/main.py
@@ -18,6 +18,7 @@ lgr.log(5, "Importing cmdline.main")
 import argparse
 import sys
 import textwrap
+import shutil
 from importlib import import_module
 
 import datalad
@@ -176,19 +177,22 @@ def setup_parser(
 
     # create command summary
     cmd_summary = []
+    console_width = shutil.get_terminal_size()[0] \
+        if hasattr(shutil, 'get_terminal_size') else 80
+
     for i, grp in enumerate(interface_groups):
         grp_descr = grp[1]
         grp_cmds = grp_short_descriptions[i]
 
         cmd_summary.append('\n*%s*\n' % (grp_descr,))
         for cd in grp_cmds:
-            cmd_summary.append('  - %-20s %s'
-                               % ((cd[0] + ':',
+            cmd_summary.append('  %s\n%s'
+                               % ((cd[0],
                                   textwrap.fill(
                                       cd[1].rstrip(' .'),
-                                      75,
-                                      #initial_indent=' ' * 4,
-                                      subsequent_indent=' ' * 8))))
+                                      console_width - 5,
+                                      initial_indent=' ' * 6,
+                                      subsequent_indent=' ' * 6))))
     # we need one last formal section to not have the trailed be
     # confused with the last command group
     cmd_summary.append('\n*General information*\n')
@@ -199,7 +203,7 @@ def setup_parser(
     Detailed usage information for individual commands is
     available via command-specific --help, i.e.:
     datalad <command> --help"""),
-                         75, initial_indent='', subsequent_indent=''))
+                         console_width - 5, initial_indent='', subsequent_indent=''))
     parts['datalad'] = parser
     lgr.log(5, "Finished setup_parser")
     if return_subparsers:
--- a/datalad/interface/__init__.py
+++ b/datalad/interface/__init__.py
@@ -12,6 +12,7 @@
 
 __docformat__ = 'restructuredtext'
 
+# ORDER MATTERS FOLKS!
 
 # the following should be series of import definitions for interface implementations
 # that shall be exposed in the Python API and the cmdline interface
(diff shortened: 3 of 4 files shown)


yarikoptic:
oh -- problem wasn't really (just in) the listing of commands.  It is in too wide lines in DESCRIPTION and below.  So
```shell
$> zcat /usr/share/man/man1/datalad.1.gz | grep -A4 DESCRIPTION
.SH DESCRIPTION
DataLad provides a unified data distribution with the convenience of git\-annex
.br
repositories as a backend.  DataLad command line tools allow to manipulate
.br
```
due to indentation renders as

```
DESCRIPTION
       DataLad  provides  a  unified data distribution with the convenience of
       git-annex
       repositories as a backend.  DataLad command line tools allow to manipu‐
       late
```

those warnings weren't my primary warning

mih:
@yarikoptic Thanks for the hint. I pushed another fix.

While we are at it, maybe someone has a suggestion on how to redo the command list in `datalad.1` -- in tight terminals it doesn't look that well.

mih:
OSX failure is unrelated.

yarikoptic:
Probably the only way is not so it at all (just say COMMAND) and rely on consise list under in description

mih:
Sounds acceptable to me. However, that list in description isn't a beauty either.

mih:
Did the COMMAND change, looking at additional changes now

mih:
Now with extra sweetness.

yarikoptic:
I think it is much better now, but options descriptions is still wrapped not so neatly:
```
...
         --version       show the program's version and license information
         --dbg           enter Python debugger when uncaught exception happens
         --idbg          enter IPython debugger when uncaught  exception  hap‐
       pens
         -C  PATH          run  as if datalad was started in <path> instead of
       the
                               current working  directory.  When  multiple  -C
       options
                               are  given,  each  subsequent  non-absolute  -C
       <path> is
                               interpreted  relative  to  the   preceding   -C
       <path>. This
                               option  affects the interpretations of the path
       names
                               in that they are made relative to the working
                               directory caused by the -C option

```
anything could be done without major pain?

also at some point (not necessarily in this PR) we should probably sort those lists of commands within each section -- now it seems to be pretty much arbitrary:
```
       Commands for dataset operations

         create
             Create a new dataset from scratch
         install
             Install a dataset from a (remote) source
         get
             Get any dataset content (files/directories/subdatasets)
         add
             Add files/directories to an existing dataset
         publish
             Publish a dataset to a known sibling
         uninstall
             Uninstall subdatasets
         drop
             Drop file content from datasets
         remove
             Remove components from datasets
         update
             Update a dataset from a sibling
         create-sibling
             Create dataset(s)'s sibling (e.g., on a web server)
```

mih:
re order of things: they are in the exact order in which we list them in `datalad/interface/__init__.py` -- folowing a divine plan!

mih:
re option description formatting: A little tricky due to the flexibility of the argparse help output depending on the amount of text and length of the option labels. Would prefer to keep this for later entertainment.

bpoldrack:
re divine plan: Please, have a comment on this within `datalad/interface/__init__.py`.

yarikoptic:
almost perfect, thanks @mih!

look though at `man git` and see how sub-sections are demarkated:
```
.SH "HIGH\-LEVEL COMMANDS (PORCELAIN)"
.sp
We separate the porcelain commands into the main commands and some ancillary user utilities\&.
.SS "Main porcelain commands"
.PP
\fBgit-add\fR(1)
.RS 4
```

is there a chance we could gain those .SS's for "Commands for meta data handling" etc? ;)

mih:
sure

mih:
if you look at the synopsis of create you can see that there is more todo. suggestion for that?

yarikoptic:
eh -- indeed.  I have tried non-braking space in between `--some-option` and its `VALUE` but then it start braking at `--some` and `option` etc.  So the only way I think is to see if we could provide manual/improved formatting is to fool argparse probably.... here is a first sketch which seems to do it:
```diff
diff --git a/formatters.py b/formatters.py
index 8ae88693..a06966c1 100644
--- a/formatters.py
+++ b/formatters.py
@@ -56,8 +56,12 @@ class ManPageFormatter(argparse.HelpFormatter):
     def _mk_synopsis(self, parser):
         self.add_usage(parser.usage, parser._actions,
                        parser._mutually_exclusive_groups, prefix='')
-        usage = self._format_usage(None, parser._actions,
-                                   parser._mutually_exclusive_groups, '')
+        try:
+            self._width -= 8
+            usage = self._format_usage(None, parser._actions,
+                                       parser._mutually_exclusive_groups, '')
+        finally:
+            self._width += 8
         # replace too long list of commands with a single placeholder
         usage = re.sub(r'\n\s*{.*}\n\s*', ' COMMAND ', usage, flags=re.MULTILINE)
         usage = usage.replace('%s ' % self._prog, '')
```

mih:
I think latest change makes it look good enough.

mih:
Bottom line: This is nicer enough. Unless it broke something, I will stop here.

yarikoptic:
it is indeed nice enough (thus merging), although I would have gone for my adhoc adjustment to wrapping since currently it would still wrap in the middle of options and thus making it not clear if that is the hyphen for wrapping or a dash within option name, 
```shell
              [-D DESCRIPTION] [-d PATH] [--no-annex] [--nosave] [--annex-ver‐
              sion  ANNEX_VERSION]  [--annex-backend ANNEX_BACKEND] [--native-
              metadata-type   LABEL]   [--if-dirty    COMMAND   ]   [--shared-
```

yarikoptic:
d'oh - was about to merge but then decided to check datalad.1 -- some itemization seems to be broken:
```
   Commands for meta data handling
       search Search  within  available in datasets' meta data aggregate-meta‐
              data Aggregate meta data of a dataset for later query
```

mih:
That is not broken that is how man/troff do it.

yarikoptic:
> That is not broken that is how man/troff do it.
I guess I wasn't clear - search and aggregate-meta now we're clashed into a single paragraph

mih:
Indeed, wtf. Thanks for persistence. Will check,

mih:
@yarikoptic fix in #1201

PR #1201, opened 2017-01-19
BF: Don't ignore commands with dashes

mih:


yarikoptic:
If works for you - works for me! Can't check ATM

mih:
I can work so gooood...

Issue #1199, opened 2017-01-19
`update` needs attention


mih:
Update needs an RF to bring it in line with the more recently refurbished commands. Please share you ideas on what needs to be done. Further work needs decisions on the following @bpoldrack @yarikoptic:

- [ ] update --merge fails without giving a clear message (conflicts in .gitmodules) why #1525
- [x] Implement `--reobtain-data`, possibly based on #1203
- [x] Decide whether `--merge` should become ON by default
- [x] Give some (any) indication of available updates when run without `--merge` -- at present it will simply say nothing, and even say nothing (and do nothing) if there are actual updates (#1219)
- [ ] Process trees of datasets and save potentially updated states accordingly (see comment at the end of #1370)
- [x] Unify with other commands and say "what to update" rather than "from where" as the primary argument(s) #1185 
- [x] Simplify complexity by just supporting an update from one sibling at a time, instead of am ordered list. It could be done for a list, but complexity explodes when considering trees of datasets at once.
- [ ] if a path to a file or directory *within* a dataset is given as input: shall we fail and ask for root paths only, or shall we inform that the associated dataset will be updated? MIH: leaning towards the latter.
- [ ] decide whether `--fetch-all` combined with `--merge` should remain "get and apply all available update", or whether `--fetch-all`should literally mean just that and `--merge` would only merge what is indicated by `--sibling` (MIH: the latter is cleaner)
- [x] for updating an annex repo switch away from `annex merge` to `annex sync --no-push --no-pull` in order to properly handle managed branches (e.g. in v6 repos)
- [x] decide on return value(s). MIH: Seems JSON-like output makes most sense. Proposal:
    * which dataset was updated
    * from which sibling (report actual name in case of requested default)
    * state/commit/tag? updated from and to, in case of merge-enabled update
- [x] Resolution for #1007 
- [x] Resolution for #953
- [x] User reporting #868 
- [x] Multi-branch support #793
- [x] No tags fetched #712 
- [x] #705
- [x] #1306 
- [x] #1307 
- [ ] #1313 more strategies for `-reobtain-data`

mih:
OK, guys. Silence is approval. Consider questions and conditional statements above as imperative. Now in execution phase.

mih:
See #1304

yarikoptic:
well -- I trust you @mih ;) and had no chance yet to give a proper consideration to the list which managed to sneak in unnoticed (I guess when I was again shoveling the damn snow) ;)

[...]

PR #1203, opened 2017-01-20
RF: `publish`

mih:
This refurbishes `publish` in the following ways:

- [X] provide a generic replacement of the `--since` functionality that was built into `publish`, and RF it to make use of it. This  could be used in other places too. Such as "`get` data for all files that were changed `--since`". Which could be used for the presently not-implemented `update --reobtain-data` (ping #1199).
- [X] reimplement `publish` to perform upfront checks and sorting without recursive self-invocation at the top-level. Recursive calls are still used to handle publication dependencies, but only a cheap helper call is used for this purpose
- [X] let `publish` look harder to figure out if anything needs to be pushed at all
- [x] sift through the function signature to see what kind of values/object should be passed to yield most flexibility
- [x] reevaluate where in the logic publication dependencies should be detected and acted upon

### DIFF
--- a/datalad/distribution/tests/test_publish.py
+++ b/datalad/distribution/tests/test_publish.py
@@ -168,7 +168,7 @@ def test_publish_recursive(origin, src_path, dst_path, sub1_pub, sub2_pub):
 
     # test for publishing with  --since.  By default since no changes, nothing pushed
     res_ = publish(dataset=source, recursive=True)
-    eq_(set(r.path for r in res_[0]), set([]))
+    eq_(set(r.path for r in res_[0]), set())
 
     # still nothing gets pushed, because orgin is up to date
     res_ = publish(dataset=source, recursive=True, since='HEAD^')
--- a/datalad/distribution/tests/test_publish.py
+++ b/datalad/distribution/tests/test_publish.py
@@ -18,6 +18,7 @@ from datalad.dochelpers import exc_str
 from datalad.utils import chpwd
 from datalad.support.gitrepo import GitRepo
 from datalad.support.annexrepo import AnnexRepo
+from datalad.support.exceptions import InsufficientArgumentsError
 
 from nose.tools import ok_, eq_, assert_false, assert_is_instance
 from datalad.tests.utils import with_tempfile, assert_in, with_tree,\
@@ -33,6 +34,18 @@ from datalad.tests.utils import skip_if_no_module
 from datalad.tests.utils import ok_clean_git
 from datalad.tests.utils import swallow_logs
 from datalad.tests.utils import assert_not_in
+
+
+@with_testrepos('submodule_annex', flavors=['local'])
+def test_invalid_call(origin):
+    ds = Dataset(origin)
+    ds.uninstall('subm 1', check=False)
+    # nothing
+    assert_raises(ValueError, publish, '/notthere')
+    # known, but not present
+    assert_raises(ValueError, publish, opj(ds.path, 'subm 1'))
+    # --since without dataset is not supported
+    assert_raises(InsufficientArgumentsError, publish, since='HEAD')
 
 
 @with_testrepos('submodule_annex', flavors=['local'])  #TODO: Use all repos after fixing them
@@ -61,7 +74,8 @@ def test_publish_simple(origin, src_path, dst_path):
 
     # don't fail when doing it again
     res = publish(dataset=source, to="target")
-    eq_(res, ([source], []))
+    # and nothing is pushed
+    eq_(res, ([], []))
 
     ok_clean_git(src_path, annex=False)
     ok_clean_git(dst_path, annex=False)
@@ -108,7 +122,7 @@ def test_publish_recursive(origin, src_path, dst_path, sub1_pub, sub2_pub):
     # subdatasets have no remote yet, so recursive publishing should fail:
     with assert_raises(ValueError) as cm:
         publish(dataset=source, to="target", recursive=True)
-    assert_in("No sibling 'target' found", exc_str(cm.exception))
+    assert_in("Unknown target sibling 'target'", exc_str(cm.exception))
 
     # now, set up targets for the submodules:
     sub1_target = GitRepo(sub1_pub, create=True)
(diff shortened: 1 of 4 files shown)


datalad/interface/utils.py
```diff
@@ -667,3 +667,84 @@ def _discover_trace_to_known(path, trace, spec):
...
+    given paths are replaced by the set of actually modified paths within them.
+
+    Only commited changes are considered!
+
+    Parameters
+    ----------
+    content_by_ds : dict
+      Per-dataset path specifications, as produced ,for example, by
+      `Interface._prep()`
+    refds : Dataset
```
yarikoptic: if we are just using `.path` and `.repo` of the refds -- why bother passing (creating) `Dataset` and not just pass `Repo` (which has `.path`)?

datalad/interface/utils.py
```diff
@@ -667,3 +667,84 @@ def _discover_trace_to_known(path, trace, spec):
...
+    # life is simple: we diff the base dataset, and kill anything that
+    # does not start with something that is in the diff
+    # we cannot really limit the diff paths easily because we might get
+    # or miss content (e.g. subdatasets) if we don't figure out which ones
+    # are known -- and we don't want that
+    diff = refds.repo.repo.commit().diff(since)
+    # get all modified paths (with original? commit) that are still
+    # present
+    modified = dict((opj(refds.path, d.b_path),
+                    # talk to @yaricoptic about the next line
```
yarikoptic: that guy doesn't exist! ;)

yarikoptic: I got it now!  but meanwhile here is probably a handy demo case:
```shell
$> d=/tmp/d; rm -rf $d; mkdir $d; cd $d; git init; echo 1 > modified; echo 2 > removed; echo 1 > renamed; git add modified removed renamed; git commit -m 'initial commit'; echo 11 >> modified; rm removed; echo 33 > added; git mv renamed renamed2; git add added; git commit -m "removed added etc" -a  ; 
Initialized empty Git repository in /tmp/d/.git/          
[master (root-commit) 7a2f2c1] initial commit
 3 files changed, 3 insertions(+)
 create mode 100644 modified
 create mode 100644 removed
 create mode 100644 renamed
[master 28fd456] removed added etc
 4 files changed, 2 insertions(+), 1 deletion(-)
 create mode 100644 added
 delete mode 100644 removed
 rename renamed => renamed2 (100%)

$> datalad install -d . ///labs/haxby/raiders; cd raiders; git reset --hard HEAD^; cd ..; git commit -m 'downgrade raiders' -a;
[INFO   ] Installing subdataset from 'http://datasets.datalad.org/labs/haxby/raiders' at: http://datasets.datalad.org/labs/haxby/raiders 
[INFO   ] Saving possible changes to <Dataset path=/tmp/d> - installed subdataset: raiders 
                                                                                                                          
[INFO   ] Saving possible changes to <Dataset path=/tmp/d> - installed subdataset: raiders 
<Dataset path=/tmp/d/raiders>
```

so let's check:
```
$> python -c 'from git import Repo; r = Repo("."); print [((d.a_path, bool(d.a_blob)), (d.b_path, bool(d.b_blob))) for d in r.commit().diff("HEAD^^")]'
[((u'.gitmodules', True), (u'.gitmodules', False)), ((u'raiders', True), (u'raiders', False))]
```
so it seems that b's blob is null... and altogether:
```shell
$> python -c 'from git import Repo; r = Repo("."); print [((d.a_path, bool(d.a_blob)), (d.b_path, bool(d.b_blob))) for d in r.commit().diff("HEAD^^^")]' 
[((u'.gitmodules', True), (u'.gitmodules', False)), ((u'added', True), (u'added', False)), ((u'modified', True), (u'modified', True)), ((u'raiders', True), (u'raiders', False)), ((u'removed', False), (u'removed', True)), ((u'renamed2', True), (u'renamed', True))]
```

So if the question was -- `why b` -- then what I have missed in the (absent) documentation of `_get_changed_datasets` is that it returns a dictionary for paths with values being previous states, so I could pass those as "since" into the corresponding sub-dataset `publish` invocations, and here you also pass recursively into `filter_unmodified`. That was "commented" on as 
```# if we were returned a dict, we got subds_prev_hexsha```
in `publish` logic.  Also in that function I was explicitly checking for no renames (so paths were the same on a and b) for now, so you might improve upon that here while at it ;)

File: datalad/interface/utils.py (hunk 1 above)
Comment (bpoldrack) [inline datalad/interface/utils.py]: I think the toplevel implementations should just operate on `Dataset` instances. Whoever uses these should not need to know much about anything underneath `Dataset` from my point of view.

Comment (yarikoptic) [inline datalad/interface/utils.py]: this is an internal function as far as I see it -- so why waste cycles creating `Dataset` instances even if they are more or less lightweight?

Comment (bpoldrack) [inline datalad/interface/utils.py]: The function is to be used by toplevel API functions, which operate on `Dataset` instances. They don't need to create it in order to pass it. They already have it.

**Comment by mih** [comment]
@bpoldrack @yarikoptic The basic concept of the new implementation is in. The, slightly adjusted, tests pass locally for me -- not sure why no test run is attempted for this PR.

Please check whether the test adjustments, and consequently the change in behavior, make sense to you. I left a few more TODOs based on your previous feedback and additional observations in the top-level description. Unless you are spotting an issue, I consider this to be, roughly, the initial RF step. Once this is merged, We could add the missing bits from #1197.

**Comment by yarikoptic** [comment]
re tests run: interesting!  may be something changed in travis API or backend so travis stopped picking those up?  our buildbot says (in response to this PR I think): `No changes found`.  Is this an issue which got "PR attached" from command line tool or fully crafted on github?  could you please push one more commit of some kind into it?

File: datalad/distribution/tests/test_publish.py [hunk 3]
```diff
@@ -61,7 +74,8 @@ def test_publish_simple(origin, src_path, dst_path):
 
     # don't fail when doing it again
     res = publish(dataset=source, to="target")
-    eq_(res, ([source], []))
+    # and nothing is pushed
+    eq_(res, ([], []))
```
yarikoptic: not that I am against such a change (return only published items) but note that this would diverge from how/what e.g. `install` and probably other API commands return -- they are trying to fulfill the promise, i.e. return requested item to be installed even if it was installed before already.  The reason for that was I believe to be able to have e.g. a Python script which you could just rerun and use those returned objects.
I wonder now if we should introduce here and elsewhere some additional option (e.g. `--affected-only`) which would switch between two behaviors. not sure though if that is feasible or not.

datalad/distribution/tests/test_publish.py
```diff
@@ -152,25 +166,26 @@ def test_publish_recursive(origin, src_path, dst_path, sub1_pub, sub2_pub):
     eq_(list(sub2_target.get_branch_commits("git-annex")),
         list(sub2.get_branch_commits("git-annex")))
 
-    # test for publishing with  --since.  By default since no changes, only current pushed
+    # test for publishing with  --since.  By default since no changes, nothing pushed
     res_ = publish(dataset=source, recursive=True)
-    # only current one would get pushed
-    eq_(set(r.path for r in res_[0]), {src_path})
+    eq_(set(r.path for r in res_[0]), set([]))
```
yarikoptic: nit picking --  `set([]) == set()`

datalad/interface/tests/test_utils.py
```diff
@@ -249,3 +250,65 @@ def test_interface_prep():
...
+        assert_equal([r], p)
+
+    orig_base_commit = ds.repo.repo.commit()
+    # nothing was modified compared to the status quo, output must be empty
+    assert_equal({}, filter_unmodified(spec, ds, orig_base_commit))
+
+    # modify one subdataset
+    added_path = opj(subb.path, 'added')
+    with open(added_path, 'w') as f:
+        f.write('test')
```
yarikoptic: just FYI, I think it could make use of create_tree to make it a  one liner
```python
create_tree(subb.path, {'added': 'test'})
```

datalad/interface/utils.py
yarikoptic: my email reply didn't post... I was talking about recursive invocation within this function:
```
filter_unmodified(mod_subs, Dataset(subds_path), modified[subds_path])
```
so it would possibly create those datasets whenever there is no really any use for them since only repo is used.

yarikoptic:
it feels like this one is composed from commits collected from other PRs -- may be that is why CI didn't run the tests since it ran on those commits

datalad/interface/utils.py
mih: Don't worry. That is one of the TODOs.

datalad/distribution/tests/test_publish.py
mih: I don't think we need two modes, but I see your point. However, we should discuss the issue of return values on a more general level. At present there is no clear concept, and the issue of what items are return is just one, the format/nature of return values (JSON-like, ...) is another. For this PR, I'd prefer to have the return values reflect what was actually done. In a second step I'd like to uniformize return values (and handling in renderers) across all or most commands.

mih:
Not sure about 14.04 build bot failure. Smaug seems to be down.

datalad/distribution/tests/test_publish.py
mih: Done

mih:
Move return value issue into #1218

mih:
Time to. take the leap. I expect the next PR on this to come within the next few days.

PR #1206, opened 2017-01-23
Fixing SSH shared connection issue

### LINKED ISSUE #1127: asks for ssh password for every dataset while installing recursively:
version 0.4.1
```shell
$> datalad install -r -s smaug:/mnt/btrfs/dbic/inbox/bids dbic-bids
[INFO   ] Installing dataset at /data/dbic-bids_/dbic-bids from smaug:/mnt/btrfs/dbic/inbox/bids 
ECDSA host key for IP address '129.170.233.9' not in list of known hosts.
yoh@smaug's password: 
...
[INFO   ] Installing subdataset <Dataset path=/data/dbic-bids_/dbic-bids/dbic> 
ECDSA host key for IP address '129.170.233.9' not in list of known hosts.
yoh@smaug's password: 
```
and so on... you could try yourself on our canonical dataset  (and this one with current master)

```shell
hopa:/tmp
$> datalad install -r falkor:/srv/datasets.datalad.org/www
[INFO   ] Installing dataset at /tmp/www from falkor:/srv/datasets.datalad.org/www 
Enter passphrase for key '/home/yoh/.ssh/id_rsa': 
Enter passphrase for key '/home/yoh/.ssh/id_rsa': 
[INFO   ] Installing subdataset <Dataset path=/tmp/www/crcns> 
Enter passphrase for key '/home/yoh/.ssh/id_rsa': 
[INFO   ] Installing subdataset <Dataset path=/tmp/www/dicoms> 
Enter passphrase for key '/home/yoh/.ssh/id_rsa': 
```

bpoldrack:
- BF/ENH: Don't close SSH connections, that were not opened by this instance of SSHManager

- Workaround: Unclear circumstances led to AnnexRepo.__del__ accessing NoneType object. Just ignore and log it.

Closes #1127

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -249,7 +249,6 @@ class AnnexRepo(GitRepo, RepoInterface):
         try:
             if hasattr(self, '_batched') and self._batched is not None:
                 self._batched.close()
-            super(AnnexRepo, self).__del__()
         except TypeError as e:
             # Workaround:
             # most likely something wasn't accessible anymore; doesn't really
@@ -259,6 +258,11 @@ class AnnexRepo(GitRepo, RepoInterface):
             # thing to happen, since we check for things being None herein as
             # well as in super class __del__;
             # At least log it:
+            lgr.debug(exc_str(e))
+        try:
+            super(AnnexRepo, self).__del__()
+        except TypeError as e:
+            # see above
             lgr.debug(exc_str(e))
 
     def _set_shared_connection(self, remote_name, url):
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -246,15 +246,27 @@ class AnnexRepo(GitRepo, RepoInterface):
         self._batched = BatchedAnnexes(batch_size=batch_size)
 
     def __del__(self):
-        if hasattr(self, '_batched') and self._batched is not None:
-            self._batched.close()
-        super(AnnexRepo, self).__del__()
+        try:
+            if hasattr(self, '_batched') and self._batched is not None:
+                self._batched.close()
+            super(AnnexRepo, self).__del__()
+        except TypeError as e:
+            # Workaround:
+            # most likely something wasn't accessible anymore; doesn't really
+            # matter since we wanted to delete it anyway.
+            #
+            # Nevertheless, in some cases might be an issue and it is a strange
+            # thing to happen, since we check for things being None herein as
+            # well as in super class __del__;
+            # At least log it:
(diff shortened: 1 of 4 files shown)


datalad/support/annexrepo.py
```diff
@@ -246,15 +246,27 @@ def __init__(self, path, url=None, runner=None,
...
+            # Workaround:
+            # most likely something wasn't accessible anymore; doesn't really
+            # matter since we wanted to delete it anyway.
+            #
+            # Nevertheless, in some cases might be an issue and it is a strange
+            # thing to happen, since we check for things being None herein as
+            # well as in super class __del__;
+            # At least log it:
+            lgr.debug(exc_str(e))
+            pass
```
yarikoptic: while at it 
- I would have wrapped both into two separate try/excepts -- although not even sure either we need to super it up here -- I guess we should
- no need for a `pass` so why having it?

datalad/support/tests/test_sshconnector.py
```diff
@@ -80,15 +80,30 @@ def test_ssh_open_close(tfile1):
...
 
     manager.close()
 
-    ok_(not exists(opj(manager.socket_dir, 'localhost')))
-    ok_(not exists(opj(manager.socket_dir, 'datalad-test')))
+    still_exists_1 = exists(opj(manager.socket_dir, 'localhost'))
+    still_exists_2 = exists(opj(manager.socket_dir, 'datalad-test'))
+
+    ok_(existed_before_1 == still_exists_1)
+    ok_(existed_before_2 == still_exists_2)
```
yarikoptic: why not `eq_` so we could actually see what is failing whenever it would?

datalad/support/annexrepo.py
bpoldrack: Yes, we should super it up, since what `GitRepo.__del__` does is valid for `AnnexRepo` too.

`pass` is a left over indeed. Will remove it.

datalad/support/tests/test_sshconnector.py
bpoldrack: Agree.

datalad/support/annexrepo.py
bpoldrack: What's the benefit of having two separate wrappings? The log via `exc_str()` should tell us everything anyway or do I miss something?

yarikoptic: if there is an exception thrown in the first line(s) the super wouldn't be even called if there is just a single try/except handler. so to robustify

bpoldrack: Well, okay. Done.

PR #1227, opened 2017-01-25
Fix fixes and improvements around github sibling creation

mih:


mih:
Ok, as this is little more than actually making it work without access tokens I'll merge this now.

PR #1230, opened 2017-01-25
RF: Uniform and optional --sibling-name for `create-sibling*`

### LINKED ISSUE #1229: create-sibling: always derive default sibling name from hostname
It is very annoying that by default no remote is configured unless a name is given.

Let's derive a default sibling name from the hostname of the SSHURL. This is save, we already error in case such a remote exists.

mih:
This fixes #1229

### DIFF
--- a/datalad/distribution/create_sibling.py
+++ b/datalad/distribution/create_sibling.py
@@ -14,12 +14,11 @@ __docformat__ = 'restructuredtext'
 
 import logging
 import datalad
-from os.path import join as opj, abspath, basename, relpath, normpath, dirname
+from os.path import join as opj, basename, relpath, normpath, dirname
 from distutils.version import LooseVersion
 from glob import glob
 
 from datalad.support.network import RI, URL, SSHRI
-from datalad.support.gitrepo import GitRepo
 from datalad.support.param import Parameter
 from datalad.dochelpers import exc_str
 from datalad.support.constraints import EnsureStr, EnsureNone, EnsureBool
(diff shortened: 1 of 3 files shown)


datalad/distribution/create_sibling.py
```diff
@@ -336,23 +346,22 @@ def __call__(sshurl, target=None, target_dir=None,
...
-                         publish_by_default=publish_by_default,
-                         publish_depends=publish_depends)
+        # add the sibling(s):
+        lgr.debug("Adding the siblings")
+        if target_url is None:
+            target_url = sshurl
+        if target_pushurl is None and sshurl != target_url:
+            target_pushurl = sshurl
+        AddSibling()(dataset=ds,
+                     name=sibling_name,
```
yarikoptic: Here you go - why but just a `name` since the command is ivory about a sibling? And then actually uniform with add-sibling ;-)

mih:
My reason for the sibling prefix was uniformity. I was about to propose the same for `update` (which has `--sibling`). Maybe we can apply the same compromise and use `--name` for the `create-sibling*` commands, but have `-s` as the uniform short option name.

yarikoptic:
just out loud thinking
```shell
$> git grep -e 'args.*[^-]-s\>' 
datalad/distribution/get.py:            args=("-s", "--source",),
datalad/distribution/install.py:            args=("-s", "--source"),
datalad/distribution/update.py:            args=("-s", "--sibling",),
datalad/metadata/search.py:            args=('-s', '--search'),

$> git grep -e 'args.*[^-]-n\>'
datalad/distribution/get.py:            args=("-n", "--no-data",),
datalad/interface/crawl.py:        #            args=("-n", "--dry-run"),
```

so those two `-s` (pretty much the same thing) are to point to siblings (and anything else annex might know and we don't)... so indeed would be kinda uniform but somewhat counterintuitive why not  `-n` since a `name`.  But I guess that is how it should be to not conflict with `get -n`

mih:
And there is also `-s` for `install`, which is conceptually the same thing, and also wouldn't work well as `-n`.

PR #1232, opened 2017-01-27
General `create-sibling` RF and few enhancements

### LINKED ISSUE #1233: Fragile `create_sibling()` and `add_sibling()` interaction
If I see this correctly, a recursive `create_sibling()` will generate target path names, and then hand over to `add_sibling` to rediscover all subdatasets, and reimagine the URLs that match the generated target paths -- which were created with different code. A comment in there promises a centralization of this functionality. 

I don't like that much.

I propose the following change: RF `add_sibling()` to offer a helper function that performs most of its functionality, but bypasses all dataset discovery, and accepts a data structure that contains all relevant information, including the target URL.

@bpoldrack?

mih:
So far, mostly sorting checks to fail as early (and as cheap) as possible, plus some RF to use more common code. And now also various minor fixes and simplifications. Brought back previously disabled tests.

If you are happy about what you see feel free to click the merge button.

- [x] Fix #1233 
- [x] Implements --since to create siblings only for added subdatasets

### DIFF
--- a/datalad/distribution/create_sibling.py
+++ b/datalad/distribution/create_sibling.py
@@ -59,7 +59,8 @@ def _create_dataset_sibling(
         shared,
         remote_git_version,
         publish_depends,
-        publish_by_default):
+        publish_by_default,
+        as_common_datasrc):
     localds_path = ds.path
     if not replicate_local_structure:
         ds_name = localds_path.replace("/", "-")
@@ -427,7 +428,8 @@ class CreateSibling(Interface):
                 shared,
                 remote_git_version,
                 publish_depends,
-                publish_by_default)
+                publish_by_default,
+                as_common_datasrc)
             if not path:
                 # nothing new was created
                 continue
--- a/datalad/distribution/create_sibling.py
+++ b/datalad/distribution/create_sibling.py
@@ -258,19 +258,32 @@ class CreateSibling(Interface):
                 sorted(datasets.keys(), key=lambda x: x.count('/')):
             current_ds = datasets[current_dspath]
             if not replicate_local_structure:
-                path = target_dir.replace("%NAME",
-                                          current_dspath.replace("/", "-"))
+                ds_name = current_dspath.replace("/", "-")
+                path = target_dir.replace("%NAME", ds_name)
             else:
                 # TODO: opj depends on local platform, not the remote one.
                 # check how to deal with it. Does windows ssh server accept
                 # posix paths? vice versa? Should planned SSH class provide
                 # tools for this issue?
                 # see gh-1188
-                path = normpath(opj(target_dir,
-                                    relpath(current_dspath,
-                                            start=ds.path)))
+                ds_name = relpath(current_dspath, start=ds.path)
+                path = normpath(opj(target_dir, ds_name))
 
-            lgr.info("Creating target dataset {0} at {1}".format(current_dspath, path))
+            # construct a would-be ssh url based on the current dataset's path
+            sshri.path = path
+            ds_sshurl = sshri.as_str()
+            # configure dataset's git-access urls
+            ds_target_url = target_url.replace('%NAME', ds_name) \
+                if target_url else ds_sshurl
+            # push, configure only if needed
+            ds_target_pushurl = None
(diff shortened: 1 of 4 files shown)


datalad/distribution/create_sibling.py
```diff
@@ -146,46 +148,65 @@ class CreateSibling(Interface):
...
                  existing='error', shared=False, ui=False,
                  as_common_datasrc=None,
                  publish_by_default=None,
                  publish_depends=None):
 
-        if sshurl is None:
-            raise ValueError("""insufficient information for target creation
-            (needs at least a dataset and a SSH URL).""")
+        # there is no point in doing anything further
+        not_supported_on_windows(
```
bpoldrack: This is superfluous. `SSHManager` instantiation raises this before we get here.

datalad/distribution/create_sibling.py
```diff
@@ -146,46 +148,86 @@ class CreateSibling(Interface):
...
-                             purpose='creating a sibling')
-
-        assert(ds is not None and sshurl is not None and ds.repo is not None)
-
-        # determine target parameters:
+        # check the login URL
         sshri = RI(sshurl)
-
         if not isinstance(sshri, SSHRI) \
                 and not (isinstance(sshri, URL) and sshri.scheme == 'ssh'):
```
bpoldrack: While at it: condition should be replaceable by `not datalad.support.network.is_ssh()` now.

datalad/distribution/create_sibling.py
```diff
@@ -146,46 +148,86 @@ class CreateSibling(Interface):
...
+        assert(ds is not None and sshurl is not None and ds.repo is not None)
+
+        # TODO apply a future --since
+
+        # dataset instances
+        datasets = {p: Dataset(p) for p in content_by_ds}
+
+        # find datasets with existing remotes with the target name
+        remote_existing = [p for p in datasets
+                           if name in datasets[p].repo.get_remotes()]
```
bpoldrack: Not sure, but as a hint: If we run into issues here with real data, we might need an `AnnexRepo.get_remotes()` (overriding the one from `GitRepo`) in order to filter out special remotes.

datalad/distribution/create_sibling.py
mih: Cool. Note that I only moved this up....

datalad/distribution/create_sibling.py
mih: OK, will do. If "should" means "that works".

datalad/distribution/create_sibling.py
mih: I don't get this. If we have an AnnexRepo instance, this will be called anyways. If we don't have one, we cannot have special remotes. What am I missing?

datalad/distribution/create_sibling.py
bpoldrack: :-) I used it with the entire connection sharing mechanic.
Therefore "should" means: "If I didn't miss an important point".

datalad/distribution/create_sibling.py
bpoldrack: You miss, that `AnnexRepo` currently is just using the method inherited from `GitRepo`. It's all fine herein, but we might need to change this within `AnnexRepo`.

mih: OK, so nothing that needs to go into this change.

datalad/distribution/create_sibling.py
mih: Done.

datalad/distribution/create_sibling.py
bpoldrack: Yes, just wanted to save you time, in case to try with an setup, that would trigger that issue. ;-)

mih: @bpoldrack an issue for that would be good. This information will soon be lost in this PR otherwise.

mih:
FTR: All untested lines correspond to code that was just moved, but was there before.

mih:
FTR: This will not see further changes, unless required by something in #1237

mih:
Alrighty, life goes on. Merge.

PR #1234, opened 2017-01-28
just to test is all buildbots are healthy now - do not merge

yarikoptic:


yarikoptic:
@datalad/developers   nd14.04 is healthy again.  Had to manually install setuptools of some previous release before doing 'pip install -e .' which would otherwise install most recent one which leads to kaboom

PR #1237, opened 2017-01-29
Continuation of #1232, focus on add-sibling, create-sibling*, publish interaction

### LINKED ISSUE #335: Publish to github with specialremote for data access

mih:
This PR sits on top of #1232

- [x] first and foremost this fixes #335
- [ ] investigate #1256

### DIFF
--- a/datalad/distribution/create_sibling.py
+++ b/datalad/distribution/create_sibling.py
@@ -60,7 +60,6 @@ def _create_dataset_sibling(
         target_pushurl,
         existing,
         shared,
-        remote_git_version,
         publish_depends,
         publish_by_default,
         as_common_datasrc):
@@ -166,7 +165,7 @@ def _create_dataset_sibling(
 
     # check git version on remote end
     lgr.info("Adjusting remote git configuration")
-    if remote_git_version and remote_git_version >= "2.4":
+    if ssh.get_git_version() and ssh.get_git_version() >= LooseVersion("2.4"):
         # allow for pushing to checked out branch
         try:
             ssh("git -C {} config receive.denyCurrentBranch updateInstead".format(
@@ -180,8 +179,8 @@ def _create_dataset_sibling(
                   " Version detected on server: %s\nSkipping configuration"
                   " of receive.denyCurrentBranch - you will not be able to"
                   " publish updates to this repository. Upgrade your git"
-                  " and run with --existing=reconfigure"
-                  % remote_git_version)
+                  " and run with --existing=reconfigure",
+                  ssh.get_git_version())
 
     # enable metadata refresh on dataset updates to publication server
     lgr.info("Enabling git post-update hook ...")
@@ -435,8 +434,6 @@ class CreateSibling(Interface):
                 'git-annex',
                 msg='on the remote system')
 
-        remote_git_version = CreateSibling.get_remote_git_version(ssh)
-
         # loop over all datasets, ordered from top to bottom to make test
         # below valid (existing directories would cause the machinery to halt)
         # But we need to run post-update hook in depth-first fashion, so
@@ -457,7 +454,6 @@ class CreateSibling(Interface):
                 target_pushurl,
                 existing,
                 shared,
-                remote_git_version,
                 publish_depends,
                 publish_by_default,
                 as_common_datasrc)
@@ -516,24 +512,6 @@ class CreateSibling(Interface):
                           "\nError: %s\nSkipping ..." % (path, exc_str(e)))
                 return False
         return True
-
-    @staticmethod
-    def get_remote_git_version(ssh):
-        try:
-            # options to disable all auto so we don't trigger them while testing
-            # for absent changes
-            out, err = ssh("git version")
-            assert out.strip().startswith("git version")
-            git_version = out.strip().split()[2]
-            lgr.debug("Detected git version on server: %s" % git_version)
-            return LooseVersion(git_version)
-
-        except CommandError as e:
-            lgr.warning(
-                "Failed to determine git version on remote.\n"
-                "Error: {0}\nTrying to configure anyway "
-                "...".format(exc_str(e)))
-        return None
 
     @staticmethod
     def create_postupdate_hook(path, ssh, dataset):
--- a/datalad/distribution/create_sibling.py
+++ b/datalad/distribution/create_sibling.py
@@ -38,6 +38,7 @@ from datalad.support.constraints import EnsureChoice
 from datalad.support.exceptions import InsufficientArgumentsError
 from datalad.support.network import RI
 from datalad.support.network import is_ssh
+from datalad.support.sshconnector import sh_quote
 from datalad.support.param import Parameter
 from datalad.utils import make_tempfile
 from datalad.utils import not_supported_on_windows
@@ -102,7 +103,7 @@ def _create_dataset_sibling(
         # TODO: Is this condition valid for != '.' only?
         path_exists = True
         try:
-            out, err = ssh(["ls", remoteds_path])
+            out, err = ssh("ls {}".format(sh_quote(remoteds_path)))
         except CommandError as e:
             if "No such file or directory" in e.stderr and \
                     remoteds_path in e.stderr:
@@ -119,9 +120,9 @@ def _create_dataset_sibling(
                 return
             elif existing == 'replace':
                 # enable write permissions to allow removing dir
-                ssh(["chmod", "+r+w", "-R", remoteds_path])
+                ssh("chmod +r+w -R {}".format(sh_quote(remoteds_path)))
                 # remove target at path
-                ssh(["rm", "-rf", remoteds_path])
+                ssh("rm -rf {}".format(sh_quote(remoteds_path)))
                 # if we succeeded in removing it
                 path_exists = False
             elif existing == 'reconfigure':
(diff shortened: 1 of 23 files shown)


mih:
Rebased on #1241 (remote SSH invocation still an issue #1248)

datalad/distribution/create_sibling.py
```diff
@@ -379,9 +380,14 @@ def __call__(sshurl, name=None, target_dir=None,
         # dataset instances
         datasets = {p: Dataset(p) for p in content_by_ds}
 
+        # make sure dependencies are valid
+        for d in datasets.values():
+            _check_deps(d.repo, publish_depends)
```
yarikoptic: `publish_depends` sounds like a property of each particular dataset.  I could have some datasets depending on some remotes and some which don't.  Here, testing is done of all resolved datasets against the same set of "publish_depends"... shouldn't publish_depends be stored in the datalad's config, probably not persistent (thus within .git/config) for the dataset?

docs/source/cmdline.rst
```diff
@@ -61,3 +61,11 @@ Miscellaneous commands
...
    generated/man/datalad-ls
    generated/man/datalad-test
+
+Plumbing commands
+=================
+
+.. toctree::
+   :maxdepth: 1
+
+   generated/man/datalad-sshrun
```
yarikoptic: probably  generated/man/datalad-create-test-dataset  could migrate here

datalad/distribution/create_sibling.py
yarikoptic: ah -- nevermind me here ... I thought I was looking at publish at that point ;)  for create-sibling obviously this logic would be appropriate ;)

datalad/support/tests/test_annexrepo.py
```diff
@@ -1029,8 +1031,8 @@ def test_annex_ssh(repo_path, remote_1_path, remote_2_path):
     rm2 = AnnexRepo(remote_2_path, create=False)
 
     # check whether we are the first to use these sockets:
-    socket_1 = opj(ssh_manager.socket_dir, 'datalad-test')
-    socket_2 = opj(ssh_manager.socket_dir, 'localhost')
+    socket_1 = opj(ssh_manager.socket_dir, get_connection_hash('datalad-test'))
+    socket_2 = opj(ssh_manager.socket_dir, get_connection_hash('localhost'))
```
yarikoptic: I feel like having `ssh_manager.get_connection_hash` doing this opj could be of help ;)

datalad/support/tests/test_sshconnector.py
```diff
@@ -167,3 +170,22 @@ def test_ssh_copy(sourcedir, sourcefile1, sourcefile2):
             eq_(content, fp.read())
 
     ssh.close()
+
+
+@skip_ssh
+def test_ssh_compound_cmds():
+    ssh = SSHManager().get_connection('ssh://localhost')
+    out, err = ssh('[ 1 = 2 ] && echo no || echo success')
```
yarikoptic: so -- could we pass a string with `'` in it? with `"`?  with both?

datalad/distribution/create_sibling.py
```diff
@@ -548,9 +538,12 @@ def create_postupdate_hook(path, ssh, dataset):
         # collate content for post_update hook
         hook_content = '\n'.join(['#!/bin/bash', 'git update-server-info', make_log_dir, json_command])
 
-        with make_tempfile(content=hook_content) as tempf:  # create post_update hook script
-            ssh.copy(tempf, hook_remote_target)             # upload hook to dataset
-        ssh(['chmod', '+x', hook_remote_target])            # and make it executable
+        with make_tempfile(content=hook_content) as tempf:
+            # create post_update hook script
+            # upload hook to dataset
+            ssh.copy(tempf, hook_remote_target)
```
bpoldrack: I have no idea why (or whether) this worked before, but this is one of the things currently failing (even manually), since the directory to copy to ('hooks') doesn't exist on the remote end.
So, similar to `make_log_dir = ... logs_remote_dir` in line 524 a `make_hooks_dir = ... hooks_remote_dir` would be required, I guess.

datalad/distribution/tests/test_publish.py
```diff
@@ -258,3 +252,69 @@ def test_publish_with_data(origin, src_path, dst_path, sub1_pub, sub2_pub):
     # before
     eq_({sub1.path, sub2.path},
         set(result_paths))
+
+
```
bpoldrack: missing `@skip_ssh`

mih: Done

datalad/support/tests/test_sshconnector.py
mih: You can feed it any valid Python string. As the docs say, quoting (via. sh_quote) is up to the user. In any case, this string is passed as the one (and single) last argument to the SSH call -- no matter how many unmatched quotes it contains. If it does, then the call on the remote will fail, of course -- but it will be on the remote.

datalad/support/tests/test_annexrepo.py
mih: Seems only relevant for these tests. Actual use doesn't see it.

docs/source/cmdline.rst
mih: True, done.

datalad/distribution/create_sibling.py
mih: Question is: why doesn't it exist (anymore). 'hooks/' is created by `git init`.

datalad/support/tests/test_sshconnector.py
yarikoptic: great -- I just was hinting on it would be great if it was tested at least a bit here on that regard

mih:
@yarikoptic @bpoldrack Still three tests failing because of too old Git on travis. No idea. It is unlikely that I will manage to revisit this PR within a week (maybe two). Please help out, if you can.

bpoldrack:
FTR: I'm on it. The issue in all three cases is the same. It's a `git push` call, which somehow manages to ignore `GIT_SSH_COMMAND` and therefore is calling `ssh` instead of `datalad sshrun`.

bpoldrack:
Ok, the three failures are fixed with the commit in PR #1263, I'm pretty sure.
But it leads to an `AssertionError`, due to a wrong assertion from my point of view:

in line 258ff in `datalad/distribution/tests/test_target_ssh.py` we currently assume, that `.git/config` is modified only if the local system-git is new enough to understand the updateInstead setting for recieve. This doesn't make sense. First of all, we don't use system-git, but the bundled one and second the _remote_ git has to understand that setting, not the local one. So just remove the condition in line 258.

datalad/distribution/publish.py
```diff
@@ -292,14 +353,19 @@ def __call__(
 
         if dataset and since:
```
bpoldrack: I think this is the issue regarding `since` (Issue #1256 ):
There is no default for `since`. If it is `None`, the commit to consider should be the latest on the remote (last published), right @yarikoptic , @mih ?

bpoldrack: To be more clear: By ignoring `since` if it is `None`, it implicitly defaults to local(!) HEAD, instead of the remote state, I think.

bpoldrack: Forget my comments. Talkin' alternative facts ... believe me - a second ago, it did make sense. :-D

bpoldrack:
Since #1256 (open task) can't be reproduce ATM - merge.

Issue #1240, opened 2017-01-30
Make sure bundled Git is used on remote systems (not just localhost in travis)


mih:
The resolve this issue the work in #1182 must be extended to properly detect Git for SSH-based remote execution. We will frequently encounter the problem that system Git is outdated (think "stable" production servers). At present this is already hurting test efforts, as I cannot have `localhost` and `datalad-test` be two pretend-different hosts and have bundled Git be used on both of them.

Issue #1192, opened 2017-01-18
create-sibling should fail sooner than later when remote host has no annex


yarikoptic:
We do require git-annex on remote box if publishing git annex repositories, so we could 'git annex init' them there to get uuid.  If we are creating with -r, it just keeps failing for all of them instead of ideally first sensing that annex is not available and quitting without performing any action (so we could rerun it after having annex installed and have no side-effects of partially configured repositories which would not be reconfigured)

Sample error:
```
[INFO   ] Creating target dataset bids/dbic/dbic-animals/derivatives/mriqc at /inbox/BIDS/dbic/dbic-animals/derivatives/mriqc 
[ERROR  ] Initialization of remote git annex repository failed at /inbox/BIDS/dbic/dbic-animals/derivatives/mriqc.
| Error: CommandError: command '['ssh', '-o', u'ControlPath=/home/mvdoc/.cache/datalad/sockets/rolando.cns.dartmouth.edu', u'rolando.cns.dartmouth.edu', "'git'", "'-C'", u"'/inbox/BIDS/dbic/dbic-animals/derivatives/mriqc'", "'annex'", "'init'"]' failed with exitcode 1
| Failed to run ['ssh', '-o', u'ControlPath=/home/mvdoc/.cache/datalad/sockets/rolando.cns.dartmouth.edu', u'rolando.cns.dartmouth.edu', "'git'", "'-C'", u"'/inbox/BIDS/dbic/dbic-animals/derivatives/mriqc'", "'annex'", "'init'"] under None. Exit code=1. out= err=git: 'annex' is not a git command. See 'git --help'.
|  [cmd.py:run:370]
``

mih:
And do this as part of a general remote sensing step that also gets a git version....

mih:
Once #1241 is merged, this will be trivial to implement.

mih:
Fix in #1237

PR #1241, opened 2017-01-30
NF: Remote git version sensing; SSH execution env conditions (fixes gh-1240)

mih:
NF: Remote git version sensing; SSH execution env conditions (fixes gh-1240)
    
With this change, remote SSH calls will try to use a bundles Git version if any exists (or can be found next to git-annex). This should work for explicit calls to git, as well as any implicit  calls made by other (proxy) commands.

This required some changes in the way we call SSH. Most notably we need to send commands as a single string (not a list), and we need to quote path arguments prior to sending them to SSH.

TODO:

- [x] Open SSH connection on demand
- [x] New `SSHConnection.is_open()`
- [x] RF `SSHConnection.__call__()` to accept a single command string. This is also the way Git will send commands along
- [x] Auto-sensing of Git version, Git-annex version and installation location, as will as detection of bundles Git installations
- [x] Proper control path filename encoding of host, user and port, following SSH recommendations
- [x] Use hash as control path filename to not give away information on shared locations
- [x] Implementation of an `sshrun` command to be used by us as `GIT_SSH_COMMAND`
- [x] RF to be able to generate control paths in the tests too. All failing tests are due to inappropriate assumptions on the control path value after RF
- [x] When `CommandError` reaches the top level, datalad as a whole fails with the same error and exit code as the underlying error.
- [ ] #1247

### DIFF
--- a/datalad/distribution/create_sibling.py
+++ b/datalad/distribution/create_sibling.py
@@ -38,6 +38,7 @@ from datalad.support.constraints import EnsureChoice
 from datalad.support.exceptions import InsufficientArgumentsError
 from datalad.support.network import RI
 from datalad.support.network import is_ssh
+from datalad.support.sshconnector import sh_quote
 from datalad.support.param import Parameter
 from datalad.utils import make_tempfile
 from datalad.utils import not_supported_on_windows
@@ -102,7 +103,7 @@ def _create_dataset_sibling(
         # TODO: Is this condition valid for != '.' only?
         path_exists = True
         try:
-            out, err = ssh(["ls", remoteds_path])
+            out, err = ssh("ls {}".format(sh_quote(remoteds_path)))
         except CommandError as e:
             if "No such file or directory" in e.stderr and \
                     remoteds_path in e.stderr:
@@ -119,9 +120,9 @@ def _create_dataset_sibling(
                 return
             elif existing == 'replace':
                 # enable write permissions to allow removing dir
-                ssh(["chmod", "+r+w", "-R", remoteds_path])
+                ssh("chmod +r+w -R {}".format(sh_quote(remoteds_path)))
                 # remove target at path
-                ssh(["rm", "-rf", remoteds_path])
+                ssh("rm -rf {}".format(sh_quote(remoteds_path)))
                 # if we succeeded in removing it
                 path_exists = False
             elif existing == 'reconfigure':
@@ -133,7 +134,7 @@ def _create_dataset_sibling(
 
         if not path_exists:
             try:
-                ssh(["mkdir", "-p", remoteds_path])
+                ssh("mkdir -p {}".format(sh_quote(remoteds_path)))
             except CommandError as e:
                 lgr.error("Remotely creating target directory failed at "
                           "%s.\nError: %s" % (remoteds_path, exc_str(e)))
@@ -167,8 +168,8 @@ def _create_dataset_sibling(
     if remote_git_version and remote_git_version >= "2.4":
         # allow for pushing to checked out branch
         try:
-            ssh(["git", "-C", remoteds_path] +
-                ["config", "receive.denyCurrentBranch", "updateInstead"])
+            ssh("git -C {} config receive.denyCurrentBranch updateInstead".format(
+                sh_quote(remoteds_path)))
         except CommandError as e:
             lgr.error("git config failed at remote location %s.\n"
                       "You will not be able to push to checked out "
@@ -469,9 +470,8 @@ class CreateSibling(Interface):
         for path in remote_repos_to_run_hook_for[::-1]:
             # Trigger the hook
             try:
-                ssh(
-                    ["cd '" + _path_(path, ".git") + "' && hooks/post-update"],
-                    wrap_args=False  # we wrapped here manually
+                ssh("cd {} && hooks/post-update".format(
+                    sh_quote(_path_(path, ".git")))
                 )
             except CommandError as e:
                 lgr.error("Failed to run post-update hook under path %s. "
@@ -482,9 +482,9 @@ class CreateSibling(Interface):
 
     @staticmethod
     def init_remote_repo(path, ssh, shared, dataset, description=None):
-        cmd = ["git", "-C", path, "init"]
-        if shared:
-            cmd.append("--shared=%s" % shared)
+        cmd = "git -C {} init{}".format(
+            sh_quote(path),
+            " --shared='{}'".format(sh_quote(shared)) if shared else '')
         try:
             ssh(cmd)
         except CommandError as e:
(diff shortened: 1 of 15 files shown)


datalad/distribution/create_sibling.py
```diff
@@ -102,7 +103,7 @@ def _create_dataset_sibling(
         # TODO: Is this condition valid for != '.' only?
         path_exists = True
         try:
-            out, err = ssh(["ls", remoteds_path])
+            out, err = ssh("ls {}".format(sh_quote(remoteds_path)))
```
yarikoptic: Couldn't we just sh_quote every item in the list and join it, while maintaining allowing for lists as an argument?

mih:
@yarikoptic  @bpoldrack Please have a look. Meanwhile I am merging this into #1237 to be able to fix #1192

datalad/distribution/create_sibling.py
mih: Sadly not. We need to be able to prepend the PATH var adjustment to the final command. I could not figure out how to do that with a list -- short of merging the list into a string, and get into all kind of issue.

mih: In any case, this implementation should work. I would not mind a better version with list support -- but I couldn't do it. At the same time I am suffering quite a bit from outdated systrem Git version. So I rather have it this way now, than some better way later.

mih:
The last piece missing for this to be complete is to have Git prefix its commands with the proper PATH variable definitions (like already done in `__call__()`). For that I see no other reason than to have a wrapper script that takes the 2 or 4 arguments that Git will send `GIT_SSH_COMMAND`, prepends the remote command with the variable definitions, and executes it.

The status quo is simply to add the control master option to a regular SSH call.

I am torn regarding the implementation of such script. We should always have Python, but who knows whether Git can find it. I guess we have to install a proper shell script with the right permissions... Thoughts? @bpoldrack @yarikoptic

mih:
Maybe it can be `datalad sshproxy` -- datalad itself should be in the path on a user system -- at least in the session were datalad is used.

mih:
Ok, had quick chat with @bpoldrack -- doesn't seem to be a completely stupid idea. Will give it a try...

datalad/distribution/create_sibling.py
yarikoptic: `merging the list into a string, and get into all kind of issue`
yeah -- I had that in mind (merging) and trying to avoid issues by quoting each argument accordingly.  What issues you foresee?

yarikoptic:
FWIW: For now we required datalad on remote host only for publishing to a remote host and generating webui meta-information.  So if publishing without enhanced webui -- it could just work without having datalad.  Requiring datalad installed on remote host might be quite a heavy requirement imho.  Probably the most resilient way is indeed to generate on the remote end (for a session) some temporary (bash) runner which would simply adjust environment, and execute given commands?
```shell
hopa:/tmp
$> ssh localhost /tmp/runner.sh echo "bumba 'dumba'"
running echo bumba dumba...
bumba dumba
done ...

$> cat /tmp/runner.sh 
#!/bin/bash

export PATH=/usr/lib/git-annex.linux/bin:$PATH
echo "running $@..."
"$@"
echo "done ..."
```

yarikoptic:
eh -- my idea doesn't quite work for any non-trivial commands:
```
$> ssh localhost /tmp/runner.sh '[ 1 = 2 ] && echo no || echo success'
running [ 1 = 2 ]...
done ...
no
2 10089.....................................:Mon 30 Jan 2017 03:05:00 PM EST:.
hopa:/tmp
$> ssh localhost /tmp/runner.sh '[ 2 = 2 ] && echo no || echo success' 
running [ 2 = 2 ]...
done ...
no
```

mih:
Not sure whether we are talking about the same thing. Nothing here requires datalad on a remote server.

yarikoptic:
ah sorry -- I have misunderstood then why such command is needed

datalad/support/sshconnector.py
```diff
@@ -51,29 +53,32 @@ class SSHConnection(object):
...
         """
         self._runner = None
 
-        # TODO: This may actually also contain "user@host".
-        #       So, better name instead of 'host'?
-        self.host = host
-        self.ctrl_path = ctrl_path + ":" + port if port else ctrl_path
-        self.port = port
+        from datalad.support.network import SSHRI
+        if hasattr(sshri, 'scheme') and not sshri.scheme == 'ssh':
```
bpoldrack: As mentioned before: There is `is_ssh()` to shorten this ;-)

mih:
OK, with the current state we should be able to do `GIT_SSH_COMMAND="datalad sshrun"` for any invocation of Git, and we should be able to get our managed ssh connections reused. And we should eb able to `git config --add core.sshCommand "datalad sshrun` upon `create` of our datasets, to get this be default for any Git command, even outside datalad.

Except, I doesn't work -- it seems the communication with the underlying `ssh` process is missing something.

````
% GIT_TRACE=1 GIT_SSH_COMMAND="datalad sshrun" git fetch --all
12:36:00.696917 git.c:371               trace: built-in: git 'fetch' '--all'
Fetching localhost
12:36:00.697324 run-command.c:350       trace: run_command: 'fetch' '--append' 'localhost'
12:36:00.697555 exec_cmd.c:116          trace: exec: 'git' 'fetch' '--append' 'localhost'
12:36:00.698807 git.c:371               trace: built-in: git 'fetch' '--append' 'localhost'
12:36:00.699263 run-command.c:350       trace: run_command: 'datalad sshrun' 'mih@localhost' 'git-upload-pack '\''/tmp/remote/this'\'''
12:36:00.699644 run-command.c:209       trace: exec: '/bin/sh' '-c' 'datalad sshrun "$@"' 'datalad sshrun' 'mih@localhost' 'git-upload-pack '\''/tmp/remote/this'\'''
````

This looks good, but it hangs right there. Looking at the `git-upload-pack` process:

````
% strace -p 27854
strace: Process 27854 attached
read(0, 
````

If I get this right, it is reading from stdin and gets nothing. I guess we must pass stdin of the `ssrun` process through to `ssh`.

**Comment by mih** [comment]
FYI: core issues resolved (thanks @bpoldrack), push will come tonight, once minor issues are fixed too

File: datalad/support/sshconnector.py [hunk 3]
```diff
@@ -123,7 +138,16 @@ def __call__(self, cmd):
 
         # TODO: pass expect parameters from above?
         # Hard to explain to toplevel users ... So for now, just set True
```
bpoldrack: May be move the old comment to line 146 (right before `runner.run(...`), since this is, what it is referring to.

datalad/support/sshconnector.py
mih: done

datalad/support/sshconnector.py
mih: done

mih:
I consider this done. It also swallowed #1246 

Now moving over to #1237 (will rebase it onto this one for easier comprehension)...

mih:
This now has #1248 (but it doesn't seem to work yet), plus a few more fixes.

mih:
OK, closing this one to reduce the magnitude of the mess. #1237 has the entire package.

PR #1246, opened 2017-01-31
ENH: Make input stream available

### LINKED ISSUE #1245: Enhance `Runner` to be able to connect a process to `stdin`
This is needed for #1241. At present there seems to be no way to hand filedescriptor through as `stdin` of a process.

bpoldrack:
Closes #1245

### DIFF
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -238,7 +238,7 @@ class Runner(object):
 
     def run(self, cmd, log_stdout=True, log_stderr=True, log_online=False,
             expect_stderr=False, expect_fail=False,
-            cwd=None, env=None, shell=None):
+            cwd=None, env=None, shell=None, stdin=None):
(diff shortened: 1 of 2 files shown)


datalad/cmd.py
```diff
@@ -238,7 +238,7 @@ def decide_to_log(v):
 
     def run(self, cmd, log_stdout=True, log_stderr=True, log_online=False,
             expect_stderr=False, expect_fail=False,
-            cwd=None, env=None, shell=None):
+            cwd=None, env=None, shell=None, inputstream=None):
```
mih: Let's do "stdin" to be consistent with the other args that use ..._stderr and so on.

mih:
Merged into #1241 already.

mih:
This is in #1237 and better comes with this context.

Issue #1254, opened 2017-02-03
save should use PWD as reference dataset if none is given and PWD contains one


yarikoptic:
```shell
(git)smaug:/mnt/btrfs/datasets/datalad/crawl/labs/tarr/face_place[master]git
$> git status
On branch master
Changes to be committed:
  (use "git reset HEAD <file>..." to unstage)

        new file:   README3_0.pdf
        new file:   README3_0.txt
        new file:   datapackage.json

(git)smaug:/mnt/btrfs/datasets/datalad/crawl/labs/tarr/face_place[master]git
$> datalad save -S -m "Added README and basic meta-data for the dataset" -a
cached/staged changes:
 README3_0.pdf    |  1 +
 README3_0.txt    | 57 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 datapackage.json | 14 ++++++++++++++
```
those are new files which were staged to be added (may be that matters that they didn't exist in index before)

using current master 0.4.1-330-gc3f4ad0

yarikoptic:
rright -- probably I had to specify the paths for files I had to save (despite -a) -- then it kinda works (different case already):
```shell
(git)smaug:/mnt/btrfs/datasets/datalad/crawl/labs/tarr/face_place[master]git
$> datalad save -S -m "fixed up json" -a datapackage.json
Exception IOError: IOError("Lock at u'/mnt/btrfs/datasets/datalad/crawl/openfmri/ds000210/.git/index.lock' could not be obtained",) in <bound method AnnexRepo.__del__ of <AnnexRepo path=/mnt/btrfs/datasets/datalad/crawl/openfmri/ds000210 (<class 'datalad.support.annexrepo.AnnexRepo'>)>> ignored
Total: 100%|███████████████████████████████████████████████████████████████████| 939/939 [00:00<00:00, 7.05KB/s]
Saved state: fc494114731ef116d50f297c081f122fef881104 for <Dataset path=/mnt/btrfs/datasets/datalad/crawl/labs/tarr/face_place>
Saved state: 6deb77618d1652f9b782939e0a01f64e3d78b583 for <Dataset path=/mnt/btrfs/datasets/datalad/crawl/labs/tarr>
Saved state: 9c88abf5a9e89a11a91cdffc2161b67d16908d5f for <Dataset path=/mnt/btrfs/datasets/datalad/crawl/labs>
[...]

PR #1261, opened 2017-02-06
BF: `save` works on PWD (again); fixes gh-1254

mih:


yarikoptic:
so `TERRIFIC!` would disappear whenever tests pass? ;)

PR #1263, opened 2017-02-06
BF: Use (bundled) git with GitPython, too.

bpoldrack:
While trying to debug remaining failures in PR #1237 , discovered we completely missed to use bundled git with GitPython.
Probably fixes these failures, since they are occuring while calling `git push` via GitPython, which then ignores `GIT_SSH_COMMAND` on Travis (that env. variable was introduced only in git 2.3), therefore wasn't calling `datalad sshrun`, but plain `ssh`, which in return led to not using bundled git on the remote end.

### DIFF
--- a/datalad/__init__.py
+++ b/datalad/__init__.py
@@ -12,6 +12,14 @@ distribution with the convenience of git-annex repositories as a backend."""
 
 # Other imports are interspersed with lgr.debug to ease troubleshooting startup
 # delays etc.
+
+# If there is a bundled git, make sure GitPython uses it too:
+from datalad.cmd import GitRunner
+_git_runner = GitRunner()
+if _git_runner._GIT_PATH:
+    import os
+    os.environ['GIT_PYTHON_GIT_EXECUTABLE'] = os.path.join(_git_runner._GIT_PATH, 'git')
+
 from .config import ConfigManager
 cfg = ConfigManager()
 
--- a/datalad/__init__.py
+++ b/datalad/__init__.py
@@ -15,10 +15,11 @@ distribution with the convenience of git-annex repositories as a backend."""
 
 # If there is a bundled git, make sure GitPython uses it too:
 from datalad.cmd import GitRunner
-_git_runner = GitRunner()
-if _git_runner._GIT_PATH:
+GitRunner._check_git_path()
+if GitRunner._GIT_PATH:
     import os
-    os.environ['GIT_PYTHON_GIT_EXECUTABLE'] = os.path.join(_git_runner._GIT_PATH, 'git')
+    os.environ['GIT_PYTHON_GIT_EXECUTABLE'] = \
+        os.path.join(GitRunner._GIT_PATH, 'git')
 
 from .config import ConfigManager
 cfg = ConfigManager()


bpoldrack:
FTR: Current failures due to connection timeout (nitrc.org)

yarikoptic:
:-/ confirming it -- I did restart one of the travis jobs and it succeeded on the 2nd try.  I will alert Christian may be -- or what should be our take from this?  This test is minor/too specific, but still would be nice to have it running to guarantee that crawler could still get that data.  but on the other hand it is large unspecific to our main development so ideally all those could/should be separated... but they also serve as an "integration" tests battery.  for now -- if you could amend last commit and re-push to trigger retesting, would be great @bpoldrack

datalad/__init__.py
```diff
@@ -12,6 +12,14 @@
 
 # Other imports are interspersed with lgr.debug to ease troubleshooting startup
 # delays etc.
+
+# If there is a bundled git, make sure GitPython uses it too:
+from datalad.cmd import GitRunner
+_git_runner = GitRunner()
+if _git_runner._GIT_PATH:
+    import os
+    os.environ['GIT_PYTHON_GIT_EXECUTABLE'] = os.path.join(_git_runner._GIT_PATH, 'git')
```
yarikoptic: why this is not just a part of where it is relevant (gitrepo.py?)?

bpoldrack: Because it isn't relevant just in gitrepo.py. ;-)
First of all, I think there's still some code, especially in the tests, that uses GitPython directly without instantiating `GitRepo`.
But more importantly, if we access `GitRepo`'s underlying GitPython objects, git is called by GitPython, not by our `Runner`. So there's nothing the `GitRepo` instance on top of it could do about it, except setting this environment variable in `GitRepo`s constructor. And this is pointless, since it nevertheless would be a universal setting, which has nothing to do with a certain repository. Plus: We would do it over and over again for no reason. Plus: We loose consistency, when using GitPython directly. So I don't see no advantage, but only disadvantages in moving this to `GitRepo`.

bpoldrack: Edited above, since I was arguing wrt to this being part of `GitRepo`'s constructor (which would be even worse). But that's probably not what you meant anyway. ;-)

yarikoptic: hm, ok... for completness 
- indeed there are some rogue imports
```shell
$> git grep -e 'import git *$' -e 'from git import'
datalad/crawler/nodes/annex.py:from git import Repo
datalad/distribution/rewrite_urls.py:    from git import GitConfigParser
datalad/support/annexrepo.py:from git import InvalidGitRepositoryError
datalad/support/tests/test_annexrepo.py:import git
datalad/support/tests/test_annexrepo.py:from git import GitCommandError
datalad/tests/test_testrepos.py:import git
datalad/tests/utils.py:import git
```

my concern is a slight but additional impact on the start time of 'datalad' cmdline tools.  From quick timing measurements it adds ~10-20ms (I guess be cause it needs to sense external git etc) to overall 380ms of running `datalad --help`. Seems to be little but it all adds up and I did spend quite a bit of time trying to cut it down.
So I guess 
- for now it would be ok to add it here indeed
- I wonder if we could add some test if that setting is in effect so later on if I come to optimize things up again, I don't ruin it?
- may instead of initiating unused GitRunner, just call `GitRunner._check_git_path()` -- yeap, it is protected API but at least it would carve off may be a few ns of dealing with possible protocols etc?

bpoldrack: Re start time: I think that is ill-measured. As far as I'm aware, there is no datalad cmdline tool, that wouldn't import `gitrepo.py` anyway. Even `datalad --help` needs to import all the toplevel functions, which in turn are importing `gitrepo` directly or via `annexrepo` or via `dataset` ... I don't think there would be an actual difference, the user could possibly experience. I think this is only valid for an artificial measure point with no meaning regarding the actual use of any command.

Re test: Well, but what would it actually test? At what point the setting should be in effect according to that test? I'm not opposing a proper test, I'm just not sure, what's actually appropriate.

Re using protected API: Well, okay. Will do it that way.

PR #1267, opened 2017-02-07
BF: import exc_str at top, fix key within log msg

yarikoptic:


bpoldrack:
I'm confused by the BUG label on a PR ...
Otherwise: No objections. :-)

yarikoptic:
BUG label in a sense that it was fixing one ;)  thanks

PR #1268, opened 2017-02-08
BF: direct mode

### LINKED ISSUE #1180: ok_clean_git() broken in direct mode

````
    self.test(*self.arg)
    ok_clean_git(ds.path)
    ok_(repo.head.is_valid())
AssertionError: None
````

### LINKED ISSUE #1208: Cannot use `submodule add` in direct mode
From tests

````
CommandError: CommandError: command '['git', '-c', 'receive.autogc=0', '-c', 'gc.auto=0', '-c', 'core.bare=False', 'submodule', 'add', 'file:///tmp/datalad_temp_testrepo_WZ5ySr', 'subm 1']' failed with exitcode 1
Failed to run ['git', '-c', 'receive.autogc=0', '-c', 'gc.auto=0', '-c', 'core.bare=False', 'submodule', 'add', 'file:///tmp/datalad_temp_testrepo_WZ5ySr', 'subm 1'] under '/tmp/datalad_temp_testrepo_gMCtaQ'. Exit code=1. out= err=fatal: $program_name cannot be used without a working tree.
````

### PR DESCRIPTION
**Pull request by bpoldrack**
- [x] Workaround direct mode issue with git 2.11.0 as pointed out in #1208 and #1179
- [x] fix `ok_clean_git` and `Repo.dirty` wrt to direct mode and (partially) V6 (#1180, #1132)
- [x] `Repo.add_submodule` works in direct mode and V6
- no failure for direct mode build left in `datalad/tests`
- no failure for direct mode build left in `datalad/support/tests`
- locally direct mode build says `FAILED (SKIP=36, errors=89, failures=29)` for me on `master` vs. `FAILED (SKIP=44, errors=42, failures=38)` with this PR.

Closes #1208 
Closes #1180

### CODE REVIEW

**Comment by bpoldrack** [comment]
There is more work to do of course, but I think this can go into master already.

File: datalad/tests/utils.py [hunk 1]
```diff
@@ -147,58 +147,92 @@ def ok_clean_git_annex_proxy(path):
...
-    ok_(exists(path))
-    ok_(exists(join(path, '.git')))
-    if annex:
-        ok_(exists(join(path, '.git', 'annex')))
-    repo = git.Repo(path)
+    if isinstance(path, AnnexRepo):
+        if annex is None:
+            annex = True
+        # if `annex` was set to False, but we find an annex => fail
+        assert_is(annex, True)
```
yarikoptic: god bless Python (at least Python2):
```shell
$> python -c 'x=True; True=False; print(x is True, bool(x))'
(False, True)
```
;)  since unlikely to happen I would let it be as is, but I would have just `ok_(annex)` since it can't be None any longer

datalad/tests/utils.py
```diff
@@ -147,58 +147,92 @@ def ok_clean_git_annex_proxy(path):
...
+    else:
+        # 'path' is an actual path
+        try:
+            r = AnnexRepo(path, init=False, create=False)
+            if annex is None:
+                annex = True
+            # if `annex` was set to False, but we find an annex => fail
+            assert_is(annex, True)
+        except Exception:
+            # Instantiation failed => no annex
```
yarikoptic: or may be "no nothing", then following `GitRepo` would raise exception as well.  Since this is a test helper -- it should just throw AssertionErrors, so below call should also be try/excepted ideally with a meaninfull informative AssertionError msg in the handler ;)

datalad/tests/utils.py
```diff
@@ -147,58 +147,92 @@ def ok_clean_git_annex_proxy(path):
...
+            r = GitRepo(path, init=False, create=False)
+            if annex is None:
+                annex = False
+            # explicitly given GitRepo instance doesn't make sense with
+            # 'annex' True
+            assert_is(annex, False)
+
+    if annex and r.is_direct_mode():
+        ok_(not r.dirty)
```
yarikoptic: and that is all there to test in direct mode? 
 I thought we discussed it and agreed that even though it might be more computationally intensive -- it would be ok to do those checks in the tests.
I bet those checks like on existence of r.path and .git/annex should still hold at least, so why they aren't common?

datalad/tests/utils.py
```diff
@@ -147,58 +147,92 @@ def ok_clean_git_annex_proxy(path):
...
-        chpwd(cwd)
-
-    assert_in(
-        "nothing to commit", out[0],
-        msg="git-status output via proxy not plausible: %s" % out[0]
-    )
+    if isinstance(path, AnnexRepo):
+        ar = path
+    else:
+        ar = AnnexRepo(path)
```
yarikoptic: strangely reported to be not covered by tests anyhow ... I guess codecov issue?

I would recommend to use pythonic construct in such cases:
```python
ar = path if isinstance(path, AnnexRepo) else AnnexRepo(path)
```
it is
- more concise
- removes duplication (`ar =`) which regardless how small could become a pain

datalad/support/tests/test_gitrepo.py
```diff
@@ -414,10 +414,10 @@ def test_GitRepo_fetch(test_path, orig_path, clone_path):
 
     fetched = clone.fetch(remote='origin')
     # test FetchInfo list returned by fetch
-    eq_([u'origin/master', u'origin/new_branch'],
+    eq_([u'origin/' + clone.get_active_branch(), u'origin/new_branch'],
         [commit.name for commit in fetched])
 
-    ok_clean_git(clone.path, annex=False)
+    ok_clean_git(clone.path)
```
yarikoptic: so, before we were explicitly testing that fresh clone is not an annex (since no annex commands were ran).  That behavior was tested explicitly with `annex=False`.  Current new default is `annex=None` which allows either if I have got it right. So it relaxes this (and other similar) test, what for?

datalad/interface/tests/test_save.py
```diff
@@ -81,7 +81,9 @@ def test_save(path):
     with open(opj(subds.path, "some_file.tst"), "w") as f:
         f.write("something")
     subds.add('.')
-    ok_clean_git(subds.path, annex=isinstance(ds.repo, AnnexRepo))
+    ok_clean_git(subds.path, annex=isinstance(subds.repo, AnnexRepo))
```
yarikoptic: so you think itwas a bug/typo before on testing subds being the same annex'ness as ds?

datalad/support/annexrepo.py
```diff
@@ -314,7 +317,22 @@ def dirty(self):
...
+            # {"status":"?","file":"filename"}
+            # ? -- untracked
+            # D -- deleted
+            # M -- modified
+            # A -- staged
+            # T -- type changed/unlocked
+            if not list(result):
+                return False
+            else:
+                return True
```
yarikoptic: did you mean to say something like `return any(result)`?

datalad/support/annexrepo.py
```diff
@@ -548,9 +566,37 @@ def set_direct_mode(self, enable_direct_mode=True):
             # adjust git options for plain git calls on this repo:
             # Note: Not sure yet, whether this solves the issue entirely or we
             # still need 'annex proxy' in some cases ...
-            if 'core.bare=False' not in self._GIT_COMMON_OPTIONS:
+
+            lgr.debug("detected git version: %s" % external_versions['cmd:git'])
+
+            if external_versions['cmd:git'] == '2.11.0':
```
yarikoptic: what do we know about this issue?  was it fixed in `2.11.1`?

datalad/tests/utils.py
yarikoptic: aha -- all uses of this helper were removed! so may be the entire helper then better be gone?

bpoldrack: Right, first changed it to ease RF'ing and then melted into ok_clean_git and forgot to remove.

datalad/interface/tests/test_save.py
bpoldrack: That's not what that parameter actually tested. In case `ds` wasn't an annex, it simply wouldn't have checked for `subds` being an annex. At least, that's what `ok_clean_git` did, when the assertion here was written. And that doesn't make sense.

bpoldrack: Furthermore, what you suggest is a matter of line 78: `subds = ds.create('subds')`, which has no parameter `no_annex=True`. If this is to be tested, then it has to be done in a unittest for `create`, not here.

datalad/support/annexrepo.py
bpoldrack: What a shame. I agree. :-)

datalad/support/annexrepo.py
bpoldrack: Don't know, but this is the point, of course. Forgot to change it to `>=`

datalad/support/tests/test_gitrepo.py
bpoldrack: This is not, what was done before. `ok_clean_git` never did that. It just checked for the existence of `.git/annex` if `annex` was True. Nothing more.

But more important: The test is running on all testrepos. So it might be an annex. It's not important whether annex commands were run, but `ok_clean_git` needs to know it is allowed to be an annex, cause this might change, what that helper has to do. (direct mode and V6)

bpoldrack: In other words: `ok_clean_git` was originally assuming, that the determination of whether or not a repo is clean does not depend on whether or not it is an annex. `annex=` added just a little additional test. But this fundamental assumption is not true.

datalad/tests/utils.py
bpoldrack: Well, okay.

datalad/tests/utils.py
bpoldrack: No, that's not all. Of course it needs implementation for all the other parameters.
Just didn't do it yet, since they are not used in `datalad/support/tests`. So this state of fixing `ok_clean_git` allowed to fix those, since now the actual failures are not hidden anymore by that simple failure of `ok_clean_git` throughout the entire test battery.

As for testing the existence of r.path and .git/annex: It's done ;-) If it wasn't there instantiation of AnnexRepo/GitRepo would fail.

datalad/tests/utils.py
bpoldrack: Nope. I need `True` in `annex`. I'm not interested in whether `bool(annex)` is `True`.
`bool("whatever the fuck")` is also `True`. Including `bool("False")` - so you'd expect `ok_clean_git(somepath, annex="False")` to behave like `ok_clean_git(somepath, annex=True)`. Which btw could be f** up the very same way, you described. ;-)

Bottom line: god bless Python for you can destroy everything with a single line.

datalad/tests/utils.py
yarikoptic: > As for testing the existence of r.path and .git/annex: It's done ;-)

my comment was about the fact that those conditions are explicitly tested also in `else:` but not here.  Either they should be tested uniformly or not tested at all explicitly relying on those calls above as you have mentioned.  Otherwise it would be for the next developer to "figure it out" why it is so

Re absent testing in direct mode -- please at least add a big fat "TODO: " comment so at least there is some indication for future generations that we did have good intentions here but just lacked time/inspiration to do that ;)

yarikoptic:
LGTM as to me, thanks @bpoldrack for all the answers. just needs a tiny change to document absent testing and resolve new conflicts I guess

datalad/tests/utils.py
bpoldrack: Agree. Removed the tests in `else:`

TODO added to docstring and a comment (to be detected by PyCharm for example) and `ok_clean_git`  raises `NotImplementedError` now whenever it is called with those additional parameters on an annex in direct mode. 
Fat enough? :-)

bpoldrack:
@datalad/developers :

Please note, that you can run `DATALAD_REPO_DIRECT=1 nosetests -s -v datalad/support datalad/tests` once this is merged. Any failure therein is happening on YOUR branch! ;-)

mih:
OK, as there are only happy campers left, I am merging this. Thanks!

PR #1269, opened 2017-02-08
BF/ENH: Make use of remote.<name>.annexurl

bpoldrack:
Solves the `annex copy` issue, we discussed during last hangout, when `git annex` doesn't use the provided pushurl to talk to a remote.

Implementing joey's solution.

See:
http://git-annex.branchable.com/bugs/annex_ignores_pushurl_and_uses_only_url_upon___34__copy_--to__34__

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -1874,7 +1874,10 @@ class AnnexRepo(GitRepo, RepoInterface):
           name of the remote
         url: str
         push: bool
-          if True, set the push URL, otherwise the fetch URL
+          if True, set the push URL, otherwise the fetch URL;
+          if True, additionally set annexurl to `url`, to make sure annex uses
+          it to talk to the remote, since access via fetch URL might be
+          restricted.
         """
 
         if push:
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -1863,6 +1863,28 @@ class AnnexRepo(GitRepo, RepoInterface):
         else:
             return None
 
+    def set_remote_url(self, name, url, push=False):
+        """Set the URL a remote is pointing to
+
+        Sets the URL of the remote `name`. Requires the remote to already exist.
+
+        Parameters
+        ----------
+        name: str
+          name of the remote
+        url: str
+        push: bool
+          if True, set the push URL, otherwise the fetch URL
+        """
+
+        if push:
+            # if we are to set a push url, also set 'annexUrl' for this remote,
+            # in order to make git-annex use it, when talking to the remote.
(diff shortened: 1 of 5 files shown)


mih:
Note: This requires reverting a "fix" in #1237 in publish.py (l.158)

datalad/support/annexrepo.py
```diff
@@ -1863,6 +1863,28 @@ def get_description(self, uuid=None):
         else:
             return None
 
+    def set_remote_url(self, name, url, push=False):
+        """Set the URL a remote is pointing to
+
+        Sets the URL of the remote `name`. Requires the remote to already exist.
```
yarikoptic: since providing custom `__doc__` already anyways (and now just `@borrowdoc`) -- why not to note how this one differs from the parent's? i.e. that it sets annexurl in this case

otherwise LGTM

bpoldrack: Yeah - okay

yarikoptic:
re reverting:  shouldn't it sense configuration for that purpose?  i.e. I have a bunch of repos now which have no field set

bpoldrack:
Re reverting:
@mih : Yes, agree. But this is based on master, so I can't revert right now. I'll wait for #1237 being merged.

@yarikoptic : May be we should have a little script, fixing existing repos instead?
Otherwise `publish` would have to write the config if it senses a missing `annexurl`. Not sure whether we want that, since it might be seen as weird, that `publish` writes to the to-be-published repo.

bpoldrack:
Alternatively, `publish` could stick to the current approach, but pass `remote.<name>.annexurl` instead of pushurl via -c if it doesn't find a configured annexurl, but does find a pushurl.
What do you think, @yarikoptic , @mih ?

mih:
I think we should configure `annexurl`. We don't want annex to fail when not used through datalad.

bpoldrack:
I agree to that argument, but `publish` still seems kind of a strange place to "fix" repos, that lack this config.
May be it simply is time for a `datalad fsck` or sth.?

mih:
I didn't mean to have `publish` fix it, but to not have `create-sibling` create the problem to begin with.

bpoldrack:
Okay, sure.
My point is @yarikoptic 's comment:

>  I have a bunch of repos now which have no field set

bpoldrack:
And btw: make `create_sibling` configuring it is what this PR achieves, since `create_sibling` uses `add_sibling`, which in return calls `repo.set_remote_url` for the push url.

bpoldrack:
Note: For now I went for `publish` to check for remotes, that are not configured as expected.

yarikoptic:
> Alternatively, publish could stick to the current approach, but pass remote.<name>.annexurl instead of pushurl via -c if it doesn't find a configured annexurl, but does find a pushurl.

sounds sensible to me!

PR #1271, opened 2017-02-09
RF: Plain move of a test battery into the module it belongs

mih:


yarikoptic:
buildbot didn't run for some reason, but looks LGTM and travis is happy

PR #1277, opened 2017-02-10
WIP: ENH/BF: Rework everthing 'dirty' - Part II

### LINKED ISSUE #781: Enhance GitRepo.dirty() to be as flexible as GitPython's is_dirty() repo function.
So we can avoid calling it directly, as done in `save()`.

### LINKED ISSUE #1133: *Repo.repo.is_dirty() is broken in direct_mode
Which in turn breaks `save`. The culprit is:

````
cmdline: git -c receive.autogc=0 -c gc.auto=0 diff --abbrev=40 --full-index --raw
  stderr: 'fatal: This operation must be run in a work tree'
````

### LINKED ISSUE #1170: *Repo.commit should report ValueError if instructed to commit untracked content
Currently it looks like this:

````
Failed to run ['git', '-c', 'receive.autogc=0', '-c', 'gc.auto=0', 'commit', '-m', 'set of new files', '/tmp/datalad_temp_clone_url_g9n4z4od/one.txt', '/tmp/datalad_temp_clone_url_g9n4z4od/two.txt'] under '/tmp/datalad_temp_clone_url_g9n4z4od'. Exit code=1. out= err=error: pathspec '/tmp/datalad_temp_clone_url_g9n4z4od/one.txt' did not match any file(s) known to git.
error: pathspec '/tmp/datalad_temp_clone_url_g9n4z4od/two.txt' did not match any file(s) known to git.
````

without such exception users (as in higher-level functions) would need to test upfront whether something is known (which can be pretty expensive) or would not be able to tell what kind error happened.

Affects the implementation of #1169

### LINKED ISSUE #1180: ok_clean_git() broken in direct mode

### LINKED ISSUE #1289: we need a better concept than staged/unstaged
As discussed in last hangout, in direct mode there are no unstaged changes.
In general the fact, that we cannot distinguish staged and unstaged changes in direct mode should lead to not use this idea with annex repos at all.

But while trying to rework the Repo classes' `dirty()` methods and the `ok_clean_git` test helper in PR #1277 this is raising questions. There is at least one important place, we use this: `interface/utils.py` in `save_dataset`. There we want to commit if we have either explicitly given paths to commit or there is something staged. As of https://github.com/datalad/datalad/pull/1277/commits/d59cbd5eb4b3f678e654623899a78b6ae901fc29 I keep the current behaviour in indirect mode, therefore still allowing for the parameter `working_tree` for `dirty()` (triggers a warning instead of an exception).
By now this means kind of a correct behaviour, since `dirty(index=True, working_tree=False)` returns `True` whenever there is something staged. It's just different what is staged in direct mode and that's okay - it's direct mode, it should behave like direct mode.

BUT: While the technical behaviour might be okay-ish, it's troublesome to provide it that way. We encourage the idea of distinguishing, what for an annex isn't a valid distinction. As long as we have it that way, the user of an `AnnexRepo` has to be aware of what that means in different annex modes. I don't see us (or even future contributors) having this is mind all the time.

So, I think we need a concept of how to catagorize possible changes in an annex repository in a more fitting way. The minimum would be to rename the parameters (originally coming from GitPython) from `index, working_tree` to `staged, unstaged`, which is still troublesome but at least better than talking about a working tree, that isn't there.
What else can you think of? Perhaps consider especially the situation described in `save_dataset`. Any other conceptional approach on what kind of changes to commit than "staged" ones? Please note, that it has not necessarily to fit into `dirty()`. Once we have an idea, what we are looking for, we can have a dedicated method to do so, of course.

### LINKED ISSUE #1291: Let it be OK to call GitRepo.commit() when there is nothing to commit
See  #1289 for the story.

### PR DESCRIPTION
**Pull request by bpoldrack**
Intense testing/development for `AnnexRepo.status()` and `XXXRepo.dirty()`
Making sure this works in v5 direct mode, v5 indirect mode, v6, v6 adjusted branch.
In all cases includes submodules and even mixed hierarchies of plain git, and different annex type repos. 

- [x] Closes #781 
- [x] Closes #1180 
- [x] Closes #1133
- [x] Closes #1291 
- [x] Closes #1170
- [ ] Closes #1289 (depends on what else waits to be discovered herein)
- ~make submodule recursion despite of direct mode submodules also work when started at a plain git~
This is going to be a new PR on top of it, since it involves more than initially expected
- [x] move calls to annex-adjust, annex-lock, annex-unlock to AnnexRepo
- [ ] document everything
- [ ] provide proper documentation: what (compound) operations on annexes we can have reliable assumptions about (before state vs after state) and what are these assumptions.

### CODE REVIEW

File: datalad/crawler/nodes/annex.py [hunk 1]
```diff
@@ -336,7 +336,7 @@ def __init__(self, path=None,
         self._providers = Providers.from_config_files()
         self.yield_non_updated = yield_non_updated
 
-        if (not allow_dirty) and self.repo.dirty:
+        if (not allow_dirty) and self.repo.dirty():
```
yarikoptic: I think I have expressed it before but could reiterate -- I see no reason or beauty in making `dirty` into a method whenever all the other similar methods are called with `is_` prefix within our Repo classes:
```shell
(git)hopa:~datalad/datalad[pr-1294]git
$> git grep -A1 '@property' -- datalad/support/*repo.py     
datalad/support/annexrepo.py:    @property
datalad/support/annexrepo.py-    def dirty(self):
--
datalad/support/annexrepo.py:    @property
datalad/support/annexrepo.py-    def default_backends(self):
--
datalad/support/annexrepo.py:    @property
datalad/support/annexrepo.py-    def uuid(self):
--
datalad/support/gitrepo.py:    @property
datalad/support/gitrepo.py-    def config(self):
--
datalad/support/gitrepo.py:    @property
datalad/support/gitrepo.py-    def dirty(self):
--
datalad/support/gitrepo.py:    @property
datalad/support/gitrepo.py-    def count_objects(self):
changes on filesystem:                                                                                                                                  
 datalad/distribution/publish.py  | 23 ++++++++++++++++++++++-
 datalad/interface/common_opts.py |  5 +++++
(dev) 2 12000.....................................:Tue 14 Feb 2017 01:55:22 PM EST:.
(git)hopa:~datalad/datalad[pr-1294]git
$> git grep -B1 'def is_' -- datalad/support/*repo.py  
datalad/support/annexrepo.py-    @classmethod
datalad/support/annexrepo.py:    def is_valid_repo(cls, path, allow_noninitialized=False):
--
datalad/support/annexrepo.py-
datalad/support/annexrepo.py:    def is_direct_mode(self):
--
datalad/support/annexrepo.py-
datalad/support/annexrepo.py:    def is_crippled_fs(self):
--
datalad/support/annexrepo.py-    @normalize_paths
datalad/support/annexrepo.py:    def is_under_annex(self, files, allow_quick=True, batch=False):
--
datalad/support/annexrepo.py-    @normalize_paths(serialize=True)
datalad/support/annexrepo.py:    def is_available(self, file_, remote=None, key=False, batch=False):
--
datalad/support/gitrepo.py-    @classmethod
datalad/support/gitrepo.py:    def is_valid_repo(cls, path):
--
datalad/support/gitrepo.py-
datalad/support/gitrepo.py:    def is_with_annex(self, only_remote=False):
```
and then I do not see why not to leave `dirty` to be a property as a lightweight shim to `is_dirty` call, thus making this PR easier to review (now full of `.dirty` -> `.dirty()`) and code which possibly uses it already outside not needing to change

bpoldrack: The point is to make it clear to the user, that there are parameters and that it is not necessarily obvious, what to consider dirty. Having a property suggests there is a notion of "dirty" that always makes sense and is the same for an annex repo and a git repo independent of the mode of operation. And this is a troublesome suggestion, which is not true.

In that sense, "all the other similar methods" actually are not that similar. Their meaning doesn't change, there simply are no parameters to tune, what these properties mean or are supposed to return.

However, I agree on renaming to `is_dirty()` to fit the pattern. If you need a shortcut property, pointing to the default values in order to save some bytes - well, feel free to build your one-liner. But I bet this will be used in an invalid way and lead to complaining on how this is confusing ;-)

bpoldrack: And: If you want that property, be aware you save just 5 Bytes by `dirty` instead of `is_dirty()`. You need to make sure, that you use it sufficiently often to compensate for the entire property definition! :-D

yarikoptic: it is not about "size" ;)
regarding all the rest of arguments ("makes sense" etc) -- scrolling pages of `.dirty -> .dirty()` changes invalidates them. The only valid to me is somewhat of a nuisance to have both `is_dirty` and `dirty` but I can leave with that :-P

bpoldrack: You won't have less changes to scroll, if I rename it to `is_dirty`. And I will not use that property, you request. I can provide it, if you insist, but I will not go through the code, figuring out, where you might want to see the use of `dirty` and where `is_dirty` is appropriate.
If your actual concern is the scrolling, don't use the f*** phone to review. ;-)
Come on - that one time scrolling in a PR review can't be an argument on how the code should look like.

yarikoptic: brr.... why to rename `dirty` to `is_dirty()` if good old `dirty` would be == `is_dirty()`???

bpoldrack:
@yarikoptic : Requirements changed. `AnnexRepo.GIT_ANNEX_MIN_VERSION` is set to be brandnew. I need the new feature. Until the builds can have a current git-annex-standalone, they will fail.

bpoldrack:
@yarikoptic : While everything is passing on Travis ATM, the buildbots all fail. From a quick look, I think none of them has a recent annex as required. Is that the case and can we have that?

yarikoptic:
yes -- buildbots do not upgrade automagically ATM.    upgraded now -- please push some ammended or new commit

bpoldrack:
Thank you!

yarikoptic:
oh -- something was quickly merged into master demanding brand new annex? (thus osx failures)

nd90 failure makes no sense to me, ie cannot be related to this PR changes -- I thought it must be upgrade effects

bpoldrack:
> something was quickly merged into master demanding brand new annex? (thus osx failures)

No, nothing I'm aware of. What is requiring brand new annex is this PR. And: OSX buildbot says it uses brand new. Plus: The failures stating git-annex doesn't know a backend called 'MD5E' doesn't make sense even with a reasonably aged git-annex, right?
Don't get it ATM.

bpoldrack:
Update on the OS X failure:
The failing command is a `git -c core.bare=False commit -m "something" .datalad` in a direct mode annex.
Interestingly this command succeeds with the system-git `git version 2.5.4 (Apple Git-61)`, but fails with the bundled git `git version 2.9.3` ("This operation must be run in a work tree").

bpoldrack:
@yarikoptic : OS X is fixed, but now suddenly all the nd buildbots (except 16_04) have a one year old annex version installed ... What's going on?

yarikoptic:
@bpoldrack AFAIK all buildbots should be top-notch now ;-)

bpoldrack:
Thanks, @yarikoptic .
FTR: As of https://github.com/datalad/datalad/pull/1277/commits/d9b5e1452b7d726f0af184abc0b327df319c3f9a, all tests are passing. Travis failure is just, that NFS build took too long. I have no idea why the Travis builds sometimes take double the time, but that's a Travis issue not ours.

yarikoptic:
Restarted that build again...

yarikoptic:
And now everything is in conflict with master, 😁

yarikoptic:
so with this PR the entire battery started to run much slower overall taking eg. ~35 min vs ~23 min on the first env in the matrix.  Running on nfsmount seems have managed to reach the cut off of 50min (was originally at around 40min... heh).   So it seems that we might benefit from making our tests a bit faster (now #1383), but also looking into the slowness brought up by this improved dirty checking

yarikoptic:
actually -- we started to reach 50min with that nfs mount env even on other PRs so we better fix it up regardless one way or another :-(

bpoldrack:
FTR: Current failure in direct mode didn't happen with `git-annex 6.20170220+gitg75a15e1ad-1~ndall+1`, but it does (again) with `git-annex 6.20170307+gitg24ade8a25-1~ndall+1`.

bpoldrack:
Closing in favor of #1450

PR #1292, opened 2017-02-14
BF: fix what's wrong with save_dataset

bpoldrack:
Stumbled upon a bug in `save_dataset`, which led to swallowing all exceptions raised by git commit.

Does this ring a bell in relation to other issues, @datalad/developers ?

PR is intended to show the issue and find the implications. Need to fix this outside of PR #1277 , since otherwise there are too many different things at once to solve ...

### DIFF
--- a/datalad/distribution/install.py
+++ b/datalad/distribution/install.py
@@ -570,8 +570,16 @@ def _save_installed_datasets(ds, installed_datasets):
         "s" if len(paths) > 1 else "", paths_str)
     lgr.info("Saving possible changes to {0} - {1}".format(
         ds, msg))
-    ds.save(
-        files=paths + ['.gitmodules'],
-        message='[DATALAD] ' + msg,
-        all_changes=False,
-        recursive=False)
+    try:
+        ds.save(
+            files=paths + ['.gitmodules'],
+            message='[DATALAD] ' + msg,
+            all_changes=False,
+            recursive=False)
+    except ValueError as e:
+        if "did not match any file(s) known to git" in str(e):
+            # install doesn't add; therefore save call might included
+            # not yet added paths.
+            pass
+        else:
+            raise
--- a/datalad/interface/utils.py
+++ b/datalad/interface/utils.py
@@ -361,7 +361,7 @@ def save_dataset(
                 lgr.debug(
                     "Was instructed to commit %s files but repository is not dirty",
                     files)
-            elif 'no changes added to commit':
+            elif 'no changes added to commit' in str(e):
                 lgr.info(
                     'Nothing to save')
             else:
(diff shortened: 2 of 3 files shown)


mih:
Why not merge? Should be merged ASAP.

bpoldrack:
Not sure yet, whether an actual fix will be done herein. Otherwise: Yes, a fix should be merged ASAP.

mih:
Hmm, in my eyes this IS the fix. The actual improvement needed is to remove this entire try/except construct and have `commit()` do all this.

mih:
From what I can see in the now failing tests (minus the direct mode stuff), would all go away when commit would not log errors, that would not needed to be swallowed, and the error checking would go inside. An immediate merge would break master, but only in the sense that it would make the breakage known, not actually introducing any.

bpoldrack:
> would break master, but only in the sense that it would make the breakage known, not actually introducing any.

Yes.

> would all go away when commit would not log errors, that would not needed to be swallowed, and the error checking would go inside

Not entirely, sadly. This is why I wanted to separate this PR out to see what's actually related to this bug and what might be an incomplete or buggy implementation of the changes to commit().

mih:
We need a bot to assign this tag based on content automatically...

bpoldrack:
:-)

To be a little bit more clear:
There is a second bug in the condition for the second case - it's not (only?) "no changes added ..", but (also?) "nothing added ..."
If I correct this, `save_dataset` doesn't only raise the correct exceptions, it then also swallows the intended ones. But this is leading to the next issue, where `datalad.distribution.tests.test_install.test_failed_install_multiple` for example, results in a lot of failures, trying to explicitly commit unstaged changes ("not known to git").

bpoldrack:
But may be this "correction" is indeed incorrect. I need to figure this out in detail ...

bpoldrack:
Update: Follow up failures seem to come from within `install` which tries to save `installed_items`, that are not staged yet.

bpoldrack:
Okay, decided to catch it from `_save_installed_datasets`, since this is the traceback:
```
distribution/install.py  Install.__call__
distribution/install.py _handle_and_return_installed_items
distribution/install.py _save_installed_datasets
interface/save.py Save.__call__
interface/utils.py save_dataset_hierarchy
interface/utils.py save_dataset
=> exception due to git commit failure ("did not match any file(s) known to git")
```

While ignoring this apparently is okay for install, it certainly isn't in general for save, I think. See next commit.

bpoldrack:
Please note, that changes to `save_dataset` will go into `GitRepo.commit()` via PR #1277 . Nevertheless, this one can go into master. The other one isn't ready yet.

bpoldrack:
%$§"(/"$% !

At a first glance - no idea why that direct mode test fails on OS X, while it passes in all other builds. git versions seem to be fine. Need to dive into it on that box.

PR #1296, opened 2017-02-14
NF: Basic support for annex-native metadata manipulation and query

mih:


mih:
Seems we have this issue: http://stackoverflow.com/questions/41145607/some-way-to-disable-ivar-auto-cross-referencing-of-attribute-names

mih:
Happy days!

yarikoptic:
that was fast! ;)
also I am confused about link referencing a bug with napoleon

mih:
Try putting:

````
Parameters
---------------
myfirstarg: path
````

in the docs. It will try to crossref `path` and fail, because we have it multiple times.

bpoldrack:
I guess the confusion is, whether we should care for that issue or you found a way to avoid it or how it is related to this PR ... ;-)

mih:
That is my secret. Get your own problems....!

yarikoptic:
Who is `it` @mih? sphinx for http://docs.datalad.org/en/latest/generated ?

PR #1298, opened 2017-02-14
BF: while creating remote sibling, first try to rmdir if exists

### LINKED ISSUE #1297: create-sibling analysis if remote dataset exist is flawed if super-ds was pushed before (nested layout only)
ATM it just checks if a directory (or a file for that sake) corresponding to the subdataset exists on remote side, and if not -- silently(**!**) skips (actually first says "Creating... ;)") without doing anything.  If we push a new state of the super dataset first, it (git?) creates empty subdirectory for it, so path "exists" and thus either causes Error or gets skipped:
```shell
hopa:/tmp/test
$> builtin cd /tmp; rm -rf test test-target; datalad create test; cd test; datalad create-sibling -s test-target localhost:/tmp/test-target; datalad publish --to=test-target ; datalad create -d . sub; datalad publish --to=test-target; datalad create-sibling -s test-target -r --existing=skip localhost:/tmp/test-target
...
[INFO   ] Publishing <Dataset path=/tmp/test> to test-target 
[INFO   ] Publishing data of dataset <Dataset path=/tmp/test> ... 
1 item was published:
Dataset: /tmp/test

[INFO   ] Skipping dataset with an already configured sibling 'test-target': ['/tmp/test'] 
[INFO   ] Connecting ... 
[INFO   ] Creating target dataset /tmp/test/sub at /tmp/test-target/sub 
(dev) 2 12112.....................................:Tue 14 Feb 2017 03:54:33 PM EST:.
hopa:/tmp/test
$> ls /tmp/test-target 
sub/
(dev) 2 12113.....................................:Tue 14 Feb 2017 03:54:48 PM EST:.
hopa:/tmp/test
$> ls -a /tmp/test-target/sub
./  ../
```
so as you can see -- empty directory remains, no remote sibling is actually created.  datalad -- current master 0.4.1-460-gc3c52a9c .  will submit a quick PR addressing it

yarikoptic:
Also adjusted logging a bit to not be misleading and more informative

Closes #1297

### DIFF
--- a/datalad/distribution/create_sibling.py
+++ b/datalad/distribution/create_sibling.py
@@ -92,7 +92,7 @@ def _create_dataset_sibling(
         ds_target_pushurl = target_pushurl.replace('%RELNAME', ds_name) \
             if target_pushurl else ds_sshurl
 
-    lgr.info("Creating target dataset {0} at {1}".format(
+    lgr.info("Considering to create a target dataset {0} at {1}".format(
         localds_path, remoteds_path))
     # Must be set to True only if exists and existing='reconfigure'
     # otherwise we might skip actions if we say existing='reconfigure'
@@ -112,13 +112,29 @@ def _create_dataset_sibling(
                 raise  # It's an unexpected failure here
 
         if path_exists:
+            _msg = "Target path %s already exists." % remoteds_path
+            # path might be existing but be an empty directory, which should be
+            # ok to remove
+            try:
+                lgr.debug(
+                    "Trying to rmdir %s on remote since might be an empty dir",
+                    remoteds_path
+                )
+                # should be safe since should not remove anything unless an empty dir
+                ssh("rmdir {}".format(sh_quote(remoteds_path)))
+                path_exists = False
+            except CommandError as e:
+                # If fails to rmdir -- either contains stuff no permissions
+                _msg += " And it fails to rmdir (%s)." % (e.stderr.strip(), )
+
+        if path_exists:
             if existing == 'error':
-                raise RuntimeError(
-                    "Target directory {} already exists.".format(
-                        remoteds_path))
+                raise RuntimeError(_msg)
             elif existing == 'skip':
+                lgr.info(_msg + " Skipping")
                 return
             elif existing == 'replace':
+                lgr.info(_msg + " Replacing")
                 # enable write permissions to allow removing dir
                 ssh("chmod +r+w -R {}".format(sh_quote(remoteds_path)))
                 # remove target at path
@@ -126,6 +142,7 @@ def _create_dataset_sibling(
                 # if we succeeded in removing it
                 path_exists = False
             elif existing == 'reconfigure':
+                lgr.info(_msg + " Will only reconfigure")
                 only_reconfigure = True
             else:
                 raise ValueError(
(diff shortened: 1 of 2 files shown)


datalad/distribution/create_sibling.py
```diff
@@ -92,7 +92,7 @@ def _create_dataset_sibling(
         ds_target_pushurl = target_pushurl.replace('%RELNAME', ds_name) \
             if target_pushurl else ds_sshurl
 
-    lgr.info("Creating target dataset {0} at {1}".format(
+    lgr.info("Considering to create a target dataset {0} at {1}".format(
```
mih: "Honestly considering"?

datalad/distribution/create_sibling.py
```diff
@@ -112,20 +112,37 @@ def _create_dataset_sibling(
...
+                    remoteds_path
+                )
+                # should be safe since should not remove anything unless an empty dir
+                ssh("rmdir {}".format(sh_quote(remoteds_path)))
+                path_exists = False
+            except CommandError as e:
+                # If fails to rmdir -- either contains stuff no permissions
+                _msg += " And it fails to rmdir (%s)." % (e.stderr.strip(), )
+
+        if path_exists:
```
bpoldrack: Why a second block with the very same condition (see line 114)?

bpoldrack:
Unconditionally trying  to `rmdir` seems to be troublesome to me.
I say `--existing=skip` and find, that I deleted something on the remote filesystem?
What if I'm not the only one to use that remote location and the thing I just deleted, contains commits I don't have?

datalad/distribution/create_sibling.py
yarikoptic: Who said that we are honest here?  but we are indeed considering (or analyzing or ...), but not yet creating really

yarikoptic:
I have also a bit of a touchy feeling about `rmdir` but imho it should be safe -- it will not delete anything but an empty directory right before creating a dataset there.  so I do not see possibility/logic for that directory to not exist later on, resulting in only new great things to come to that remote host ;)

datalad/distribution/create_sibling.py
bpoldrack: How about "preparing"?

bpoldrack:
Well if it is empty - fine. But is it? How do you know? Do I miss something?

yarikoptic:
rmdir does not remove a file or a non-empty directory -- it just fails then

alternative would be to do smth like `ls -a` on that path and analysing output (1 line -- must be a file -- error, 2 lines == '.', '..' -- our case -- rmdir, any other -- error since either not empty or I don't know what it is)...  BUT then I would need to call rmdir anyways before proceeding, and do similar analysis.  So -- is such `ls -a` safeguard valuable? not sure

datalad/distribution/create_sibling.py
mih: "Attempt" -- used in other places too.

bpoldrack:
Rrrright! I forgot `rmdir` isn't doing that anyway. Sorry.

So, should be okay as is, I guess.

datalad/distribution/create_sibling.py
yarikoptic: It is not attempting anything yet -- just doing some kind of analysis while considering to create.   "Preparing" is a viable possibility... but not better enough worth rerunning the tests ;)  Since otherwise is ok, I will merge and then will let others to improve wording

PR #1299, opened 2017-02-15
ENH: --inherit for add/create-sibling, RF: publish --missing=fail|skip|inherit; wanted for annex; apparently some BF for ls

### LINKED ISSUE #1295: create-sibling --inherit-settings
As we have discussed with @mih, we might want to lessen my and hopefully others pain while creating and publishing new sub-datasets to the siblings for which their super-datasets have everything set already.   We need it scalable and flexible (e.g. some sub-sets of datasets could have different policy from others, like I have for ///labs/haxby).

For `url` and `pushurl` -- in case of non-hierarchical setups gets tricky, I am not yet having a good grip, but could be detected/marked somehow.  For what to push/upload - somewhat of two possible ways
1. **custom config variable in super-** 
First we thought about having some additional config `remote.*.datalad-publish` which would be recorded within super's `.git/config` so that sub-dataset could just take them and use while configuring itself...   thinking about it though, "I don't like it" since it would not accommodate ad-hoc/custom setups I might have by interacting directly with `git config` and `git annex wanted` to achieve necessary effect.    Those two sets of settings would diverge and lead to chaos.  So I would prefer `2.`:
2. **inherit all relevant settings**, i.e. rely more on git/annex settings instead of ad-hoc ones.

What we really need is to inherit (depends on resolution of how publish behave, my view is here: https://github.com/datalad/datalad/issues/1274#issuecomment-279747603)
- what to push: `--get-all remote.SIBLING.push`
- what to upload: all non-empty values of `git annex wanted SIBLING` and if set to `standard` or `groupwanted`: `git annex group`, and in addition if `groupwanted` only: `git annex groupwanted`. 
- and only then inherit also any additional relevant custom variables (e.g.`datalad-publish-depends` and `datalad-publish-wanted` in https://github.com/datalad/datalad/issues/1274#issuecomment-279755509)

### Difficulties with 2.

- `shared` setting -- I don't think it is easy/viable to figure out from an existing one of the super-dataset... I could be wrong

I guess since I am the one mudding the waters and needing it most ATM, I could take a try at it

edit 1: added `datalad-publish-depends` into the list of vars to inherit
edit 2: added `Difficulties` pointing to `--shared` setting.
edit 3: +1c to those who wish all datasets to be annexes.  We cannot inherit "wanted" settings from the super dataset if it is just a pure git... heh... boomer!

yarikoptic:
Closes #1295 

- [x] get feedback from fella developers
- [x] inherit annex wanted settings
- [x] may be move some of introduced code into functions to improve modularity etc

### DIFF
--- a/datalad/distribution/add_sibling.py
+++ b/datalad/distribution/add_sibling.py
@@ -30,6 +30,9 @@ from datalad.interface.common_opts import recursion_flag
 from datalad.interface.common_opts import as_common_datasrc
 from datalad.interface.common_opts import publish_depends
 from datalad.interface.common_opts import publish_by_default
+from datalad.interface.common_opts import annex_wanted_opt
+from datalad.interface.common_opts import annex_group_opt
+from datalad.interface.common_opts import annex_groupwanted_opt
 from datalad.distribution.dataset import EnsureDataset, Dataset, \
     datasetmethod, require_dataset
 from datalad.support.exceptions import CommandError
@@ -105,6 +108,9 @@ class AddSibling(Interface):
         as_common_datasrc=as_common_datasrc,
         publish_depends=publish_depends,
         publish_by_default=publish_by_default,
+        annex_wanted=annex_wanted_opt,
+        annex_group=annex_group_opt,
+        annex_groupwanted=annex_groupwanted_opt,
         # TODO: inherit_settings
     )
 
@@ -113,7 +119,8 @@ class AddSibling(Interface):
     def __call__(name=None, url=None, dataset=None,
                  pushurl=None, recursive=False, fetch=False, force=False,
                  as_common_datasrc=None, publish_depends=None,
-                 publish_by_default=None):
+                 publish_by_default=None,
+                 annex_wanted=None, annex_group=None, annex_groupwanted=None):
 
         # TODO: Detect malformed URL and fail?
 
@@ -299,6 +306,16 @@ class AddSibling(Interface):
                             'Not configuring "%s" as a common data source, '
                             'URL protocol is not http or https',
                             name)
+                if annex_wanted:
+                    repo.set_wanted(name, annex_wanted)
+                if annex_group:
+                    repo.set_group(name, annex_group)
+                if annex_groupwanted:
+                    if not annex_group:
+                        raise ValueError("To set groupwanted, you need to provide annex_group option")
+                    repo.set_groupwanted(annex_group, annex_groupwanted)
+
+
 
             successfully_added.append(repo_name)
 
--- a/datalad/distribution/add_sibling.py
+++ b/datalad/distribution/add_sibling.py
@@ -36,6 +36,7 @@ from datalad.interface.common_opts import annex_groupwanted_opt
 from datalad.distribution.dataset import EnsureDataset, Dataset, \
     datasetmethod, require_dataset
 from datalad.support.exceptions import CommandError
+from datalad.support.exceptions import InsufficientArgumentsError
 
 
 lgr = logging.getLogger('datalad.distribution.add_sibling')
@@ -312,10 +313,9 @@ class AddSibling(Interface):
                     repo.set_group(name, annex_group)
                 if annex_groupwanted:
                     if not annex_group:
-                        raise ValueError("To set groupwanted, you need to provide annex_group option")
+                        raise InsufficientArgumentsError(
+                            "To set groupwanted, you need to provide annex_group option")
                     repo.set_groupwanted(annex_group, annex_groupwanted)
-
-
 
             successfully_added.append(repo_name)
 
--- a/datalad/distribution/add_sibling.py
+++ b/datalad/distribution/add_sibling.py
@@ -105,6 +105,7 @@ class AddSibling(Interface):
         as_common_datasrc=as_common_datasrc,
         publish_depends=publish_depends,
         publish_by_default=publish_by_default,
+        # TODO: inherit_settings
     )
 
     @staticmethod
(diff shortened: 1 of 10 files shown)


yarikoptic:
lovely tests catch awkward double `\\`'ing spaces within ssh RIs for the remote... diidn't figure out where it is coming from

yarikoptic:
@mih and @bpoldrack -- it is conceptually ready and seems to work. feedback?
will deal with failing tests if any tomorrow

yarikoptic:
BTW - as for wanted -- decided just to inherit them from super datasets which are annexes... which adds a cent to have supers annexes nearly always

datalad/distribution/publish.py
```diff
@@ -255,7 +256,10 @@ class Publish(Interface):
         recursion_limit=recursion_limit,
         git_opts=git_opts,
         annex_opts=annex_opts,
-        annex_copy_opts=annex_copy_opts
+        annex_copy_opts=annex_copy_opts,
+        # "interferes" with skip_failing which is actually "skip_missing"
+        # TODO: unify as --missing=(skip|inherit) ?
```
bpoldrack: ATM it's `skip_missing` indeed. But `skip_failing` is also used, when there was no sibling specified and we can not figure out one automatically. In that case `inherit` wouldn't work, since we don't know what to inherit from. With proper documentation, `inherit` could then fall back to `skip` behavior, but we then also need a third value `fail`, I guess.

datalad/distribution/publish.py
```diff
@@ -343,11 +348,25 @@ def __call__(
...
+                    if not superds:
+                        raise RuntimeError(
+                            "%s has no super-dataset to inherit settings for the remote %s"
+                            % (ds, to)
+                        )
+                    # XXX due to difference between create-sibling and create-sibling-github
+                    # would not be as transparent to inherit for -github
+                    lgr.info("Will try to create a sibling inheriting settings from %s", superds)
+                    # XXX explicit None as sshurl for now
+                    ds.create_sibling(None, name=to, inherit_settings=True)
```
bpoldrack: I'm sceptical. :-)
Shouldn't it be `ds.add_sibling(..., inherit_settings=True)`? Actually creating a sibling on the remote from within `publish` seems a little bit harsh to me. While it might make sense in some cases, it would require different access permissions than before in other cases. Also: What's the policy regarding parameter `existing` in that case?
I think inheriting settings for the remote configuration locally is fine, but creating a fitting remote repo is something different.

bpoldrack: May be we just need more detailed values for `inherit` to distinguish whether we should try to create or just add that sibling.

datalad/support/annexrepo.py
```diff
@@ -513,6 +513,31 @@ def _run_annex_command(self, annex_cmd, git_options=None, annex_options=None,
...
+        their output while ignoring some messages
+
+        Parameters
+        ----------
+        **kwargs: all passed into _run
+        """
+        out, err = self._run_annex_command(
+            *args, **kwargs
+        )
+        lines = out.rstrip('\n').split('\n')
```
bpoldrack: Nope. `'\n'` is platform dependent.
```python
lines = out.splitlines()
```

yarikoptic: thanks!

datalad/distribution/publish.py
yarikoptic: overall idea here was to improve UX -- I got tired of remembering either I have new sub-datasets or not before running `publish`, and then invoking `create-sibling` manually in case if there were and publish failed...  I had to remember parameters for `create-sibling` invocation etc -- just doesn't scale.  Hence I have decided to address at least this common use-case where I have a hierarchy of datasets, which I could publish over and over again efficiently even if I get new sub-datasets without remembering again and again how to create those remote again and again.
Such scheme indeed might fail in case if there is two sets of hierarchies which try to sync/publish  to the common one -- that is where `existing=skip` would be necessary I guess.  For now `existing`  here indeed remained to its safest default and I did not proxied any more settings into `create_sibling` -- for any of those scenarios better to use it directly/consciously.
Having said all that I am yet to see this piece "in action" since yesterday had to revert to separate `create-sibling` invocation anyways and `publish` is hard to tame for me (see #1329) ATM

datalad/distribution/add_sibling.py
```diff
@@ -298,6 +306,16 @@ def __call__(name=None, url=None, dataset=None,
                             'Not configuring "%s" as a common data source, '
                             'URL protocol is not http or https',
                             name)
+                if annex_wanted:
+                    repo.set_wanted(name, annex_wanted)
+                if annex_group:
+                    repo.set_group(name, annex_group)
+                if annex_groupwanted:
+                    if not annex_group:
+                        raise ValueError("To set groupwanted, you need to provide annex_group option")
```
yarikoptic: thanks!

datalad/distribution/publish.py
yarikoptic: so, semantically it is "skip missing" not "failing" since not that some publish of a sub-dataset has failed (which can happen) and we would like to skip it then.
So do you think I should tackle RF `--skip-failing` + `--inherit-settings` into `--missing=(fail|skip|inherit)` within this PR?  I guess I could as well do it

bpoldrack: Yes, makes sense to me.

yarikoptic:
oki doki 

- publish now got --missing=fail|skip|inherit
- moved relevant inheriting logic into add-sibling, so it got --inherit too now.  Doesn't have a dedicated test though but test ran all the way from the top by publish should check majority of the conditioning ;)
- also renamed --inherit-settings into just --inherit to stay inline with `--missing=inherit`

so if tests pass and no objections -- wouldn't mind a merge. real-life testing would show more

yarikoptic:
for me few tests fail ATM due to change in publish behavior, but for CI might be just fine, so if tests pass and no other comments -- a candidate for merge

bpoldrack:
Will merge for now. See #1344 .

PR #1302, opened 2017-02-15
ENH(TST): test annex meta-data manipulations with trickier values

yarikoptic:
was familiarizing myself

### DIFF
--- a/datalad/support/tests/test_annexrepo.py
+++ b/datalad/support/tests/test_annexrepo.py
@@ -50,6 +50,7 @@ from datalad.tests.utils import with_tempfile
 from datalad.tests.utils import with_tree
 from datalad.tests.utils import create_tree
 from datalad.tests.utils import with_batch_direct
+from datalad.tests.utils import assert_dict_equal as deq_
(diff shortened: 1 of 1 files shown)


datalad/support/tests/test_annexrepo.py
```diff
@@ -1617,30 +1618,32 @@ def test_AnnexRepo_metadata(path):
...
         'mike': ['awesome']}},
         ar.get_metadata('up.dat'))
-    ar.set_metadata('.', reset={'tag': 'one'}, purge=['mike'], recursive=True)
+    # Use trickier tags (spaces, =)
+    ar.set_metadata('.', reset={'tag': 'one and= '}, purge=['mike'], recursive=True)
     playfile = opj('d o"w n', 'd o w n.dat')
     target = {
         'up.dat': {
-            'tag': ['one']},
+            'tag': ['one and= ']},
```
yarikoptic: not to myself -- here using name "tag" is similar to using `annex metadata --tag`/`--untag`, so we do not need extending interface. nicely done @mih

yarikoptic:
any objections/desires?

Issue #868, opened 2016-09-20
update -r should may be more friendly and less scarry?


yarikoptic:
I have a "partially" installed our super dataset, upgrade seems to require me to specify the sibling name (suboptimal) since I guess there is also other remotes (e.g. datalad-archives)

``` shell
(git)hopa:/tmp/datasets.datalad.org[master]
$> datalad update --merge -r
2016-09-20 10:54:14,651 [ERROR  ] No merge strategy for multiple remotes implemented yet. [update.py:__call__:11(venv-tests) 2 19554 ->1......(venv-tests) 2 195(ve(((((((venv-tests) 2 19554 

$> datalad update --merge -r origin
2016-09-20 10:54:26,393 [INFO   ] Updating dataset '/tmp/datasets.datalad.org' ... (update.py:115)
2016-09-20 10:54:26,416 [INFO   ] Applying changes from tracking branch... (update.py:144)
2016-09-20 10:54:26,448 [INFO   ] Updating f6ee804..65157dd
| Fast-forward
|  .datalad/meta/crcns/meta.json  | 12 +++---
|  .datalad/meta/dicoms/meta.json | 91 +++++++++++++++++++++++++++++++++++++++---
|  .datalad/meta/nidm/meta.json   |  6 ++-
|  dicoms                         |  2 +-
|  4 files changed, 97 insertions(+), 14 deletions(-)
|  (update.py:161)
2016-09-20 10:54:26,449 [INFO   ] Updating dataset '/tmp/datasets.datalad.org/crcns' ... (update.py:115)
2016-09-20 10:54:26,473 [INFO   ] Applying changes from tracking branch... (update.py:144)
2016-09-20 10:54:26,498 [INFO   ] Already up-to-date.
|  (update.py:161)
2016-09-20 10:54:26,500 [INFO   ] Updating dataset '/tmp/datasets.datalad.org/crcns/aa-1' ... (update.py:115)
2016-09-20 10:54:26,550 [INFO   ] Applying changes from tracking branch... (update.py:144)
2016-09-20 10:54:26,581 [INFO   ] Already up-to-date.
|  (update.py:161)
2016-09-20 10:54:26,583 [INFO   ] Updating annex ... (update.py:164)
2016-09-20 10:54:26,648 [INFO   ] merge git-annex ok
|  (update.py:167)
2016-09-20 10:54:26,648 [WARNING] 'origin' not known to dataset /tmp/datasets.datalad.org/crcns/aa-2.
[...]

Issue #712, opened 2016-08-14
datalad update (without --fetch-all) doesn't fetch tags from its origin (the only sibling)


yarikoptic:
``` shell
(git)smaug:/tmp/ds000017B[master]
$> git tag -d 1.1.0
Deleted tag '1.1.0' (was 4278439)
(venv-tests)3 10495.....................................:Sun 14 Aug 2016 03:45:32 PM EDT:.
(git)smaug:/tmp/ds000017B[master]git
$> datalad update
2016-08-14 15:45:36,103 [INFO   ] Updating dataset '/tmp/ds000017B' ... (update.py:115)
(venv-tests)3 10496.....................................:Sun 14 Aug 2016 03:45:36 PM EDT:.
(git)smaug:/tmp/ds000017B[master]git
$> git tag         
1.0.0
(venv-tests)3 10497.....................................:Sun 14 Aug 2016 03:45:38 PM EDT:.
(git)smaug:/tmp/ds000017B[master]git
$> datalad update --fetch-all
2016-08-14 15:45:43,110 [INFO   ] Updating dataset '/tmp/ds000017B' ... (update.py:115)
(venv-tests)3 10498.....................................:Sun 14 Aug 2016 03:45:43 PM EDT:.
(git)smaug:/tmp/ds000017B[master]git
$> git tag                   
1.0.0
1.1.0
```

but I guess may be that is "by design", so we just need to identify how update should work

yarikoptic:
while at it -- should not puke errors about annex special remotes when --fetch-all:

```
$> datalad update --fetch-all
2016-09-29 09:51:02,679 [INFO   ] Updating dataset '/home/yoh/datalad/openfmri/ds000030' ... (update.py:115)
2016-09-29 09:51:02,724 [ERROR  ] No option 'url' in section: u'remote "datalad-archives"' [ConfigParser.py:get:340] (NoOptionError) (main.py:270)
```

mih:
On the original issue: I think `add-sibling` should configure what to fetch, and what not.

Issue #705, opened 2016-08-12
datalad update didn't actually update superdataset with new state for a child


yarikoptic:
``` shell
(git)smaug:/mnt/datasets/datalad/test-crawled2/all-5/indi/fcon1000[master]git
$> datalad update
2016-08-11 23:51:16,550 [INFO   ] Updating dataset '/mnt/datasets/datalad/test-crawled2/all-5/indi/fcon1000' ... (update.py:115)
(venv-tests)2 10875.....................................:Thu 11 Aug 2016 11:51:16 PM EDT:.
(git)smaug:/mnt/datasets/datalad/test-crawled2/all-5/indi/fcon1000[master]git
$> ls
AnnArbor_a/  Beijing_Zang/       Durham_Madden/  Milwaukee_a/  NewYork_Test-Retest_Reliability/  Ontario/     Pittsburgh/
AnnArbor_b/  Berlin_Margulies/   ICBM/           Milwaukee_b/  NewYork_a/                        Orangeburg/  Queensland/
Atlanta/     Cambridge_Buckner/  Leiden_2180/    Munchen/      NewYork_a_ADHD/                   Oulu/        SaintLouis/
Baltimore/   Cleveland\ CCF/     Leiden_2200/    NewHaven_a/   NewYork_b/                        Oxford/      Taipei_a/
Bangor/      Dallas/             Leipzig/        NewHaven_b/   Newark/                           PaloAlto/    Taipei_b/


(venv-tests)2 10876.....................................:Thu 11 Aug 2016 11:51:18 PM EDT:.
(git)smaug:/mnt/datasets/datalad/test-crawled2/all-5/indi/fcon1000[master]git
$> git status
On branch master
Your branch is behind 'origin/master' by 1 commit, and can be fast-forwarded.
  (use "git pull" to update your local branch)
nothing to commit, working directory clean
```

where `git pull --ff-only` did it

``` shell
$> git pff
Updating 9829816..b486fc3
Fast-forward
 Baltimore | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)
changes on filesystem:
 Baltimore | 2 +-
(venv-tests)2 10879.....................................:Thu 11 Aug 2016 11:51:31 PM EDT:.
(git)smaug:/mnt/datasets/datalad/test-crawled2/all-5/indi/fcon1000[master]git
$> 
changes on filesystem:                                                                                                           
 Baltimore | 2 +-
```

mih:
[...]

PR #1304, opened 2017-02-15
RF update (focus: single dataset operations)

mih:
`update` is now practically a frontend for `annex sync`.

Please comment @bpoldrack @yarikoptic 

- [x] implements `--reobtain-data`
- [x] adds missing tests: #705, #712, #868 
- [x] verify safe operation when upstream removes subdatasets that are installed locally
- [x] can upgrade a GitRepo to an AnnexRepo if needed #1307
- [x] NF: AnnexRepo.sync(), reimplemented AnnexRepo.merge_annex() with it.

From my PoV this step is complete. There is more TODO, (#1199) but there is no need to wait for it .

### DIFF
--- a/datalad/distribution/tests/test_update.py
+++ b/datalad/distribution/tests/test_update.py
@@ -9,28 +9,20 @@
 
 """
 
-import os
-from os.path import join as opj, abspath, basename
+from os.path import join as opj
 from ..dataset import Dataset
-from datalad.api import update, install, uninstall
-from datalad.utils import chpwd
+from datalad.api import install
 from datalad.utils import knows_annex
 from datalad.support.gitrepo import GitRepo
 from datalad.support.annexrepo import AnnexRepo
 
-from nose.tools import ok_, eq_, assert_false, assert_is_instance
-from datalad.tests.utils import with_tempfile, assert_in, with_tree,\
+from nose.tools import eq_, assert_false, assert_is_instance
+from datalad.tests.utils import with_tempfile, assert_in, \
     with_testrepos, assert_not_in
 from datalad.tests.utils import SkipTest
-from datalad.tests.utils import assert_cwd_unchanged, skip_if_on_windows
-from datalad.tests.utils import assure_dict_from_str, assure_list_from_str
 from datalad.tests.utils import create_tree
-from datalad.tests.utils import assert_not_in
-from datalad.tests.utils import assert_raises
 from datalad.tests.utils import ok_file_has_content
-from datalad.tests.utils import skip_if_no_module
-from datalad.tests.utils import ok_clean_git, swallow_outputs
-from datalad.tests.utils import ok_file_has_content
+from datalad.tests.utils import ok_clean_git
 
 
 @with_testrepos('submodule_annex', flavors=['local'])  #TODO: Use all repos after fixing them
--- a/datalad/distribution/tests/test_update.py
+++ b/datalad/distribution/tests/test_update.py
@@ -170,3 +170,8 @@ def test_newthings_coming_down(originpath, destpath):
     # no merge, only one sibling, no parameters should be specific enough
     ds.update()
     assert(knows_annex(ds.path))
+    # no branches appeared
+    eq_(ds.repo.get_branches(), ['master'])
+    # now merge
+    ds.update(merge=True)
+    eq_(sorted(ds.repo.get_branches()), ['git-annex', 'master'])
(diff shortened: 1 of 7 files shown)


mih:
Hmm, 'thighting"... sounds good, will do that more

mih:
Ok, this works better than what we had before. I will merge this some time this weekend.

datalad/distribution/tests/test_update.py
```diff
@@ -146,3 +143,135 @@ def test_update_fetch_all(src, remote_1, remote_2):
...
+    eq_(ds.repo.get_branches(), ['master'])
+    # now merge, and get an annex
+    ds.update(merge=True)
+    assert_in('git-annex', ds.repo.get_branches())
+    assert_is_instance(ds.repo, AnnexRepo)
+    # should be fully functional
+    testfname = opj(ds.path, 'load.dat')
+    assert_false(ds.repo.file_has_content(testfname))
+    ds.get('.')
+    ok_file_has_content(opj(ds.path, 'load.dat'), 'heavy')
```
bpoldrack: I'm a bit confused here. 'load.dat' was originally added to a plain git repo in line 152, right?
Where is the point it became an annexed file, that we are able to annex-get?

datalad/distribution/update.py
```diff
@@ -109,82 +104,67 @@ def __call__(
...
+        all_=fetch_all,
+        prune=True)  # prune to not accumulate a mess over time
+
+    if not merge:
+        return
+    # we need to check whether we need to convert this dataset to
+    # annex, would would be the case when we presently have a git repo
+    # and the recent fetch brought evidence for a remote annex
+    if not isinstance(repo, AnnexRepo) and knows_annex(repo.path):
+        lgr.info("Init annex at '%s' prior merge.", repo.path)
```
bpoldrack: I think, this should be done be re-requesting `ds.repo`, since that property already has a check for whether it became an annex. This check in `Dataset.repo` should use `knows_annex` though, since it currently is just looking for a local git-annex' branch.

datalad/support/annexrepo.py
```diff
@@ -1066,16 +1091,59 @@ def enable_remote(self, name):
...
+          configured
+        """
+        args = []
+        for label, arg in (('push', push),
+                           ('pull', pull),
+                           ('commit', commit),
+                           ('content', content)):
+            args.append('--{}{}'.format('' if arg else 'no-', label))
+        for label, arg in (('all', all), ('fast', fast)):
+            if arg:
```
bpoldrack: lines 1143 - 1146 are equivalent to:
```python
from datalad.support.gitrepo import to_options
args.extend(to_options(all=all, fast=fast))
```

Just saying - we should use it more often ;-)

bpoldrack: For the more complex thing above (`--no-something` vs `--something`) it could also be used:

```python
to_options(push=push, no_push=not push, ...)
```

datalad/distribution/tests/test_update.py
mih: Indeed. Evolution of that test somewhat distorted the original idea. Conveniently, `ds.repo.file_has_content(testfname)` returns False for a file that is guaranteed to have its content with it. So we can leave it as is!

datalad/distribution/update.py
mih: I assume you meant something like this:

````
diff --git a/datalad/distribution/update.py b/datalad/distribution/update.py
index 49a03c28..fe814705 100644
--- a/datalad/distribution/update.py
+++ b/datalad/distribution/update.py
@@ -145,7 +145,9 @@ def _update_repo(ds, remote, merge, fetch_all, reobtain_data):
     # and the recent fetch brought evidence for a remote annex
     if not isinstance(repo, AnnexRepo) and knows_annex(repo.path):
         lgr.info("Init annex at '%s' prior merge.", repo.path)
-        repo = AnnexRepo(repo.path, create=False)
+        # re-requesting the repo will notice the type change
+        ds._repo = None
+        repo = ds.repo
     lgr.info("Merging updates...")
     if isinstance(repo, AnnexRepo):
         if reobtain_data:
````

datalad/support/annexrepo.py
mih: Did the first. For the second I would need to be convinced that a causal reader has an easier time decoding with it wants to say compared to the present implementation.

mih:
Well, that went south quickly

https://travis-ci.org/datalad/datalad/jobs/203121067

I will revert the changes and repush. When there is more time, we can investigate why exactly the predicted behavior was not the observed one.

mih:
So in summary, I reverted all changes that I made in response to Ben's comments. I will merge this now, as it is still better than what it was. I won't have time to get to this again this week and I don't won't to have it dangling...

Issue #1275, opened 2017-02-10
publish should publish current branch (possibly even make it available on remote by default)


yarikoptic:
it was kinda cool that things were actually made easier for me even while dealing with code repository:
```shell
(git)hopa:~/picts/mris/heudiconv[mvdoc-fixup-protocols]git
$> datalad create-sibling -s smaug smaug:proj/heudiconv
[INFO   ] Connecting ... 
ECDSA host key for IP address '129.170.233.9' not in list of known hosts.
[INFO   ] Creating target dataset /home/yoh/picts/mris/heudiconv at proj/heudiconv 
[INFO   ] Adjusting remote git configuration 
[INFO   ] Enabling git post-update hook ... 

$> datalad --dbg publish -r --to smaug                                                 
[INFO   ] Publishing <Dataset path=/home/yoh/picts/mris/heudiconv> to smaug 
ECDSA host key for IP address '129.170.233.9' not in list of known hosts.
1 item was published:
Dataset: /home/yoh/picts/mris/heudiconv

```

so I thought -- holy cow, I am done "publishing" my code to a new host... but not really since current branch wasn't pushed, not to say that it wasn't made the "default" one.
I think, for a simplistic use-case, we actually might want publish to reflect current branch on the remote end... eg. in this case I would have been a happy camper right away ;)

mih:
How do we decide that we have a simple case?

We cannot always take the active branch (think direct mode). What if next time you publish the branch is different? You want previous branch published too? Or replace with the new one? What about master? Never publish if you are on a different branch?

Ultimately the question is: Is there any subset of the `git push` interface that we do not wish to generalize to work across any repositories?

I'd prefer if we first try to get a consistent implementation of the basic functionality done, before we implement the fancy ones.

mih:
Precondition: #1212

yarikoptic:
#1212 marked fixed. Ran into this issue again

mih:
Conclusion to this issue is unclear to me (as is the actual problem description). See unanswered questions in https://github.com/datalad/datalad/issues/1275#issuecomment-278928873

PR #1308, opened 2017-02-16
RF: Do not place default push config on publish anymore

mih:
Instead temporarily amend the push config to push the current branch, or rather the underlying branch in case of managed branches.

We may want to have this behavior to be conditioned on an explicit switch, or to only be active when there is no push config at all. Which would help with the issue that it presently cannot handle wildcard expressions for push config.

ATM it makes sense to me to do it like this.

This is work towards a resolution of gh-1275

### DIFF
--- a/datalad/distribution/publish.py
+++ b/datalad/distribution/publish.py
@@ -11,6 +11,7 @@
 """
 
 import logging
+import re
 from os.path import curdir
 
 from datalad.interface.base import Interface
@@ -129,21 +130,50 @@ def _publish_dataset(ds, remote, refspec, paths, annex_copy_options):
         ds.repo.fetch(remote=remote)
         ds.repo.merge_annex(remote)
 
-    # we now know where to push to:
     # Note: git's push.default is 'matching', which doesn't work for first
     # time publication (a branch, that doesn't exist on remote yet)
     # But if we want to respect remote.*.push entries, etc. we need to
     # not pass a specific refspec (like active branch) to `git push`
     # by default.
+    # hence we amend any existing config on the fly
     # TODO: what else to push by default?
     # consider also: --follow-tags, --tags, --atomic
-    lgr.debug("Push to '%s'", remote)
-    pushvar = 'remote.{}.push'.format(remote)
-    if not ds.config.get(pushvar, None):
-        # no default push targets configured for this remote, leave our mark
-        ds.config.add(pushvar, 'master', where='local')
+    # make sure we push
+    things2push = []
+    current_branch = ds.repo.get_active_branch()
+    if current_branch:  # possibly make this conditional on a switch
+        # TODO: this should become it own helper
         if isinstance(ds.repo, AnnexRepo):
-            ds.config.add(pushvar, 'git-annex', where='local')
+            # annex could manage this branch
+            if current_branch.startswith('annex/direct') \
+                    and ds.config.getbool('annex', 'direct', default=False):
+                # this is a "fake" annex direct mode branch
+                # we want to publish the underlying branch
+                current_branch = current_branch[12:]
+            match_adjusted = re.match(
+                'adjusted/(.*)\([a-z]*\)',
+                current_branch)
+            if match_adjusted:
+                # adjusted/master(...)
+                current_branch = match.group(1)
+        things2push.append(current_branch)
+    if isinstance(ds.repo, AnnexRepo):
+        things2push.append('git-annex')
+    # check that all our magic found valid branches
+    things2push = [t for t in things2push if t in ds.repo.get_branches()]
+    # check that we don't ask to push things that are already configured
+    # -> would cause error
+    # TODO need to find a way to properly do this, when wildcards are used
+    # in the push configuration variable
+    things2push = [t for t in things2push
+                   if t not in ds.config.get('remote.{}.push'.format(remote), [])]
+    # now we know what to push where
+    lgr.debug("Attempt to push '%s' to sibling '%s'", things2push, remote)
+    # configure targets for the next push call
+    # those will be incremental to any remote..push setting in the config
+    ds.repo.repo.git(
+        c=['remote.{}.push={}'.format(remote, thing)
+           for thing in things2push])
     _log_push_info(ds.repo.push(remote=remote))
 
     published.append(ds)
--- a/datalad/distribution/tests/test_publish.py
+++ b/datalad/distribution/tests/test_publish.py
@@ -81,6 +81,7 @@ def test_publish_simple(origin, src_path, dst_path):
 
     # 'target/master' should be tracking branch at this point, so
     # try publishing without `to`:
+    # MIH: Nope, we don't automatically add this anymore
 
     # some modification:
     with open(opj(src_path, 'test_mod_file'), "w") as f:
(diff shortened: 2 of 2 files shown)


bpoldrack:
What's the issue with wildcards? How does it fail?

I agree on the desired RF'ing to move parts of it elsewhere, but for now: LGTM

bpoldrack:
Regarding the failure: I guess, the push setting should change from `remote.<name>.push=master` to `remote.<name>.push=master:master`

mih:
any such spec is possible, but also `push=refs/heads/*` so a test using `fnmatch` (or simething like that) would be needed.

bpoldrack:
I don't get it yet. Even if it is `push=refs/heads/*` it would be the same logic: change to `push=refs/heads/*:refs/heads/*`. What do you need `fnmatch` for?

mih:
Test if `git-annex` is already covered by this expression, otherwise failure could occur. It does occure for sure when you give `git-annex` twice. Not sure if also when once directly and once indirectly.

bpoldrack:
Ah - okay.

mih:
@yarikoptic what do you think?

datalad/distribution/publish.py
```diff
@@ -129,21 +130,50 @@ def _publish_dataset(ds, remote, refspec, paths, annex_copy_options):
...
+    # check that we don't ask to push things that are already configured
+    # -> would cause error
+    # TODO need to find a way to properly do this, when wildcards are used
+    # in the push configuration variable
+    things2push = [t for t in things2push
+                   if t not in ds.config.get('remote.{}.push'.format(remote), [])]
+    # now we know what to push where
+    lgr.debug("Attempt to push '%s' to sibling '%s'", things2push, remote)
+    # configure targets for the next push call
+    # those will be incremental to any remote..push setting in the config
```
yarikoptic: thank you for the comment since I was wondering about that!  Ideally, we should have had a test for it that ;-)

mih: It is tested implicitly already. Proper test will come with more comprehensive support for expressions.

mih:
OK, the test failure in here is a real concern. We have to come up with a more general strategy on publication. What do we publish under what conditions, how much of that do we want to manifest in the config and in what way?

mih:
Elaborating on the issue:

The test failure makes it obvious that the current approach doesn't work: Even when we specify exactly what to push via config, it would be refused in some cases with `--set-upstream`. In addition, there is a potential `push.default` setting (changes with Git v2.0) that alters the behavior of what is being pushed (current default `upstream`, but `matching` might be more relevant in our case).

The first issue alone seems to indicate that we either

- make push target more permanent in the config
- figure out exactly what temporary config we need to synthesize in order keep doing what we are doing

In both cases it would help a lot to first nail a (max) two-sentence-description of what our default publication paradigm is.

Please discuss your ideas, thanks!

bpoldrack:
> current default upstream, but matching might be more relevant in our case
I think `current` is what we probably want as default behavior, since we don't necessarily publish to upstream, right?

May be I get that failure somehow wrong, but my impression is, that this is not about what to push, but where to push to. If there's no upstream (yet) and we specify only the local refs to push, but not the corresponding remote ones, it fails because `git push` doesn't know what's the remote ref to push to. So we either need to change `push.default` or make it explicit in the push settings as I mentioned above.

Sadly, I fail to reproduce the failure ATM, since it fails even earlier (and different) for me. But I guess I need to fully understand what exactly is going wrong under what circumstances and why. Did you try to reproduce a simple setup manually, that fails in the same way, @mih?

bpoldrack:
My conclusion for now would be:

We keep pushing WHAT you worked out already, we just need to make sure, WHERE we are pushing to. The only thing that's missing is the remote's refspec, that is figured by `git push` and therefore relies on a default strategy. In order to not mess with the user's setting for his workflows not involving datalad publish, I'd either:
- additionally pass `-c push.default=current`
- pass explicit remote refspec per push setting ( `-c remote.<name>.push="whatever:whatever"`)

bpoldrack:
BTW: Manually tried to specify `remote.<name>.push=git-annex` as well as `refs/heads/*`. Doesn't seem to conflict. No complaining and worked.

mih:
Please post the entire command. I tried lots of things, and could not figure out a solution that works in all scenarios -- not just the failing one.

datalad/distribution/publish.py
```diff
@@ -129,21 +130,50 @@ def _publish_dataset(ds, remote, refspec, paths, annex_copy_options):
...
+                    and ds.config.getbool('annex', 'direct', default=False):
+                # this is a "fake" annex direct mode branch
+                # we want to publish the underlying branch
+                current_branch = current_branch[12:]
+            match_adjusted = re.match(
+                'adjusted/(.*)\([a-z]*\)',
+                current_branch)
+            if match_adjusted:
+                # adjusted/master(...)
+                current_branch = match.group(1)
```
bpoldrack: Just noted:
```python
current_branch = match.group(1)
```
Should probably be:
```python
current_branch = match_adjusted.group(1)
```
Right?

mih: True.

mih:
> - pass explicit remote refspec per push setting ( -c remote.<name>.push="whatever:whatever")

we do half of that at present (not enough). We do not pass the second half of the push variable content in order to not override any existing setting. We would need a more comprehensive treatment than just setting this to be "matching".

> - additionally pass -c push.default=current

This would mean that we want to push the current branch regardless of what is configured.

mih:
To clarify: this issue is about _what_ to push. The question where to push to is secondary and can be resolved more easily once it is not what to push. The problem at hand happens when it is absolutely clear where to push to (remote is given as input), and a default logic would need to decide what to push to a blank remote.

yarikoptic:
I can merge so well

bpoldrack:
What did I miss, @yarikoptic ? Since when the failure is gone?

yarikoptic:
I must confess that I have made a mistake... I saw
- that tests passed (which is wrong) -- I guess because ssh tests run only on travis and it was hidden on the top (that is why my comments in hangout now make little sense I guess ;) )
- that I have managed to Ok (some version of) original changes
- that discussion has died off, and thought that you the issue was just for further development.

I think in general
- publish should not store any changes into config unless explicitly requested by the user (hence was my comment on analog with `-u`).  That is why the idea of changes made sense to me.
- we should do less magic -- rely more on git/git-annex config settings than our custom ones/heuristics
- but complimentary to it, I think we will not be able to react in some smart way to all those possible git settings affecting push, which also evolved organically through the years. So it might be easier/more productive to establish default behaviour via having ...
- all configuration on what should be pushed generally be done by create,add-sibling (see e.g. https://github.com/datalad/datalad/issues/925#issuecomment-254226073), so then user could, if desired, adjust it to his/her liking by editing .git/config.  The only exception could have been a git-annex branch, which I believe is handled by 'git annex sync' anyways, so actually nothing to do for us. ATM with #1229 I have just exposed native 'annex *wanted/group' settings to provide at least some interface to the data files portion of the puzzle.

So I think that ideal resolution here would be to adjust default behavior of `create,add-sibling` "to do the right thing" from the beginning:
- provide convenient choices to setup all needed push targets correctly, including covering git-annex branch) and let publish just do mere invocation of push/sync commands.  indeed we might loose some magical "push to tracking/current/etc" but we would actually have it if user sets it up accordingly
- we might want to add an option (or two ;-) ) to `publish` to be able to say explicitly what needs to be pushed for this particular invocation overloading .git/config settings this way (similar to how you could do `git push REMOTE REFSPECs`)

or am I missing something?

yarikoptic:
in this particular case it seems the critical piece somewhat is that git behavior of `git push -c remote.remote-False.push=master REMOTE` is different from just having that setting set in the config (then pushes fine) or if refspec provided in the command line positional args (also pushes fine)...

PR #1310, opened 2017-02-16
BF+ENH(DOC): fill in the blank, DeprecatedError for clone invocation of GitRepo

### LINKED ISSUE #1309: RFed behavior of GitRepo(path, url=)   is not working "correctly"
it used to be that we used `GitRepo(path, url=orig_url)` to clone.  With flywheel RF behavior changed but confusingly:
1. doc still says
```If `url` is given, a clone is created at `path`.```
2. there is a check in the code
```python
        if url is not None:
            RuntimeError("RF: url passed to init()")
```
which actually does nothing since it is not `raised`!
3. The only happens if I do call `GitRepo(path, url=orig_url)` is that new repository is created at the path.

yarikoptic:
Closes #1309

### DIFF
--- a/datalad/support/exceptions.py
+++ b/datalad/support/exceptions.py
@@ -54,6 +54,33 @@ class MissingExternalDependency(RuntimeError):
         return to_str
 
 
+class DeprecatedError(RuntimeError):
+    """To raise whenever a deprecated entirely feature is used"""
+    def __init__(self, new=None, version=None, msg=''):
+        """
+
+        Parameters
+        ----------
+        new : str, optional
+          What new construct to use
+        version : str, optional
+          Since which version is deprecated
+        kwargs
+        """
+        super(DeprecatedError, self).__init__()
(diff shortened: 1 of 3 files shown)


bpoldrack:
Well, agree on the doc corrections. Thanks.

Regarding the exception: Actually was meant to stay a little after RF PR was merged, to ease the identifying of misuse due to familiarization. I think it's time to remove `url` parameter altogether finally.
But if you prefer to still have it - I'm fine with that.

yarikoptic:
> I think it's time to remove url parameter altogether finally.

as I have whined elsewhere, ideally we should start to provide proper deprecation cycles... here -- at least this exception would provide information on what users/us should do to accommodate new interface. so may be let's keep it until at least 0.6 release ;)

yarikoptic:
ok -- since tests passed all around, I will merge now so no incorrect behavior goes unnoticed ;)

PR #1316, opened 2017-02-20
ENH: cmdline entrypoint supports cmds that are generators

mih:


bpoldrack:
Nice.

mih:
Likely not needed, though. Our future internal success evaluation would have that loop anyways.

mih:
Confirmed now, this needs to be done in a different place.

PR #1317, opened 2017-02-20
RF: Find dataset to update in PWD if nothing is given

mih:


datalad/distribution/update.py
```diff
@@ -88,6 +88,10 @@ def __call__(
         if dataset and not path:
             # act on the whole dataset if nothing else was specified
             path = dataset.path if isinstance(dataset, Dataset) else dataset
+        if not dataset and not path:
+            # try to find a dataset in PWD
+            dataset = require_dataset(
+                None, check_installed=True, purpose='updating')
```
yarikoptic: FWIW -- since I see it as a quite common case/use -- would be great to have at least a basic test for it (reported as not covered atm)

PR #1324, opened 2017-02-22
And I was on my knees begging... but I was not heard

yarikoptic:


datalad/distribution/tests/test_update.py
```diff
@@ -89,6 +94,18 @@ def test_update_simple(origin, src_path, dst_path):
...
 
+@with_tempfile
+@with_tempfile
+def test_update_git_smoke(src_path, dst_path):
+    # Apparently was just failing on git repos for basic lack of coverage, hence this quick test
+    ds = Dataset(src_path).create(no_annex=True)
+    target = install(dst_path, source=src_path)
+    create_tree(ds.path, {'file.dat': '123'})
+    ds.add('file.dat')
+    target.update(recursive=True, merge=True)
```
bpoldrack: Why `recursive`?

datalad/distribution/update.py
```diff
@@ -164,9 +165,19 @@ def _update_repo(ds, remote, merge, fetch_all, reobtain_data):
...
-            repo.pull(remote=remote)
+        if active_branch == (None, None):
+            # I guess we need to fetch, and then let super-dataset to update
+            # into the state it points to for this submodule, but for now let's
+            # just blow I guess :-/
+            lgr.warning(
+                "No active branch in %s, fetching and not changing state",
+                repo
+            )
+            repo.fetch(remote=remote)
```
bpoldrack: We already fetched at the beginning of this function.

bpoldrack:
Totally unfair! Very dishonest people! Sad!


Not sure, what to do in case of a detached HEAD and `merge=True`.
I guess, we shouldn't try to deal with it from within `_update_repo` as you did, but instead detect this in line 135 of `update.py` and then go for `git submodule update` instead of `_update_repo`, no?

datalad/distribution/tests/test_update.py
yarikoptic: why not? ;-)  grew organically while me trying to hit that failing logic nail, and I was ready to add a sub-dataset

I just know that in this setting it was failing before, and passes now

datalad/distribution/update.py
yarikoptic: thanks! adjusted...

yarikoptic:
I think we should 
- first fix up install (iirc as we have discussed before) so it stays in the branch when checking out some previous state of it, not the current HEAD.  This should minimize possibility of arriving to detached heads
- detached head could happen for various reasons and may be we should better be safe than sorry, i.e. to not force 'submodule update' on it which might allow to "loose" track of some committed outside of the branch content.  IMHO just issuing warning so ppl and may be instructing (how to) fix up for it would be wiser.  With fix in install then possibility of this happening should minimize.

bpoldrack:
Good point. I (mostly) agree.
I guess instead of a warning we ultimately need a bit RF'ing to allow it to be reported `skipped`. But warning is fine for now - let's see how @mih 's playing with return values and success evaluation evolves.

yarikoptic:
ok -- ready for a merge... if nobody or no further comments -- I will do in an hour

Issue #1319, opened 2017-02-20
"new" behavior of git-annex 6.20170220+gitgbfb38eece - upon publish new commit on remote git-annex


yarikoptic:
Was checking new version and one test (only) started to fail:
```
$> DATALAD_TESTS_SSH=1 nosetests -s -v datalad/distribution/tests/test_publish.py:test_publish_simple
                                                                                                                                              
FAIL
Versions: cmd:annex=6.20170220+gitgbfb38eece-1~ndall+1 cmd:git=2.11.0 cmd:system-git=2.11.0 tqdm=4.11.2

======================================================================
FAIL: datalad.distribution.tests.test_publish.test_publish_simple
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/yoh/proj/datalad/datalad/datalad/tests/utils.py", line 734, in newfunc
    t(*(arg + (uri,)), **kw)
  File "/home/yoh/proj/datalad/datalad/datalad/tests/utils.py", line 564, in newfunc
    return t(*(arg + (filename,)), **kw)
  File "/home/yoh/proj/datalad/datalad/datalad/tests/utils.py", line 564, in newfunc
    return t(*(arg + (filename,)), **kw)
  File "/home/yoh/proj/datalad/datalad/datalad/distribution/tests/test_publish.py", line 99, in test_publish_simple
    list(source.repo.get_branch_commits("git-annex")))
[...]

PR #1327, opened 2017-02-22
BF: Correct assertion to check what we are actually interested in.

bpoldrack:
See #1319

### DIFF
--- a/datalad/distribution/tests/test_publish.py
+++ b/datalad/distribution/tests/test_publish.py
@@ -95,8 +95,8 @@ def test_publish_simple(origin, src_path, dst_path):
     ok_clean_git(dst_path, annex=None)
     eq_(list(target.get_branch_commits("master")),
         list(source.repo.get_branch_commits("master")))
-    eq_(list(target.get_branch_commits("git-annex")),
-        list(source.repo.get_branch_commits("git-annex")))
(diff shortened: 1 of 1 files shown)


datalad/distribution/tests/test_publish.py
```diff
@@ -95,8 +95,9 @@ def test_publish_simple(origin, src_path, dst_path):
     ok_clean_git(dst_path, annex=None)
     eq_(list(target.get_branch_commits("master")),
         list(source.repo.get_branch_commits("master")))
-    eq_(list(target.get_branch_commits("git-annex")),
-        list(source.repo.get_branch_commits("git-annex")))
+    assert(all(
+        [commit in list(target.get_branch_commits("git-annex"))
+         for commit in list(source.repo.get_branch_commits("git-annex"))]))
```
yarikoptic: I would have used then set not a list and it's  .difference

bpoldrack:
Better, @yarikoptic ?

yarikoptic:
"good enough" ;)   I would have also added a comment about this so later we don't wonder wtf... will do directly after merge

yarikoptic:
btw with this -- uploaded that new annex to stock neurodebian @bpoldrack

PR #1333, opened 2017-02-23
BF/RF: instead of using -c remote..push, pass refspecs into push invocation

yarikoptic:
should fix failing test
  see discussions after merge in https://github.com/datalad/datalad/pull/1308

includes an "easter egg" fix ;)

### DIFF
--- a/datalad/distribution/publish.py
+++ b/datalad/distribution/publish.py
@@ -38,7 +38,7 @@ lgr = logging.getLogger('datalad.distribution.publish')
 # TODO: make consistent configurable output
 
 
-def _log_push_info(pi_list):
+def _log_push_info(pi_list, log_nothing=True):
     from git.remote import PushInfo as PI
 
     error = False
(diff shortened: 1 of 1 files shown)


mih:
And we are full circle. This is where we started to fix the problem that existing config is ignored when push is called with specific refspecs.

yarikoptic:
Ah... 😁
Did you try to read the ones from config and provide them all then as well?

yarikoptic:
Or may be just push twice ;-)

mih:
Many things were tried, but it needs one brave man to do it right this one
time. I am right behind you!

yarikoptic:
ok -- I will see if I could make an executive decision on what behavior from publish we want ;)  but for now, to avoid broken master we then probably just merge this one, or reset --hard previous merge I have done.  any input on which way? (either is fine with me)

yarikoptic:
ok -- if desire was to push both detected and prescribed, I have for now made two pushes -- if tests pass and no big objections would merge it to bring master back in order -- two more PRs are staggering now ;)

bpoldrack:
Made issue #1339 for further discussion and to link this PR as well as the other to preserve the state of discussion.

I'll merge for the sake of a clean master.

PR #1334, opened 2017-02-23
DON'T MERGE: DEMO: decorator styles

bpoldrack:
Here is a single file demo script. Just import it.

Note, that the entire point is not how to generally use `wrapt.decorator`, but the function `my_wraps_replacement`, which is a decorator supposed to replace `functools.wraps`. Therefore the only difference between `my_functools_decorator` and `my_wrapt_decorator` is the replacement of `@wraps(func)` by `@my_wraps_replacement(func)`. The goal is to not have to rewrite our decorators.

If you don't want to have a closer look, this is the important part of the output:

difference on simple decorated functions: we got the correct signature
```
old function:
ArgSpec(args=[], varargs='args', keywords='kwargs', defaults=None)
new function:
ArgSpec(args=['one', 'two', 'three'], varargs=None, keywords=None, defaults=(None,))
```

A decorated `__call__` method of a class, used with `@datasetmethod` (including adaption).
We got the correct signature for `__call__` as well as for `Dataset.command`:

```
full-scale additional decorator:
full-scale __call__ method:
ArgSpec(args=['what', 'dataset', 'ever'], varargs=None, keywords=None, defaults=(None, None))
wrapped with wrapt
actual __call__ method:
dataset: None
what: this
ever: ever
full-scale Dataset method:
ArgSpec(args=['what', 'dataset', 'ever'], varargs=None, keywords=None, defaults=(None, None))
wrapped with wrapt
actual __call__ method:
dataset: <Dataset path=/does/not/matter>
what: this
ever: ever
```

If you have any suggestions what else to test/verify or sth like that - let me now, @mih , @yarikoptic

### DIFF
--- a/datalad/support/deco_demo.py
+++ b/datalad/support/deco_demo.py
@@ -1,0 +1,229 @@
+from datalad.distribution.dataset import Dataset, datasetmethod
+from functools import wraps
+from wrapt import decorator
+from inspect import getargspec
+
+
+# the "classical" functools decorator:
+def my_functools_decorator(func):
+
+    @wraps(func)
+    def new_func(*args, **kwargs):
+        print "wrapped with functools"
+        return func(*args, **kwargs)
+
+    return new_func
+
+
+# old-style decorated function:
+@my_functools_decorator
+def old_style_function(one, two, three=None):
+    print "old_style_function"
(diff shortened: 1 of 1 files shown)


mih:
Looks great!

PR #1337, opened 2017-02-24
BF: do not fetch remotes without url (with warning unless special remote

### LINKED ISSUE #1336: datalad update --fetch-all shouldn't try to fetch special remotes (or any remote without url for that matter)
```shell
$> datalad update --fetch-all        
[INFO   ] Updating dataset '/tmp/ds000030' ... 
[ERROR  ] No option 'url' in section: u'remote "datalad-archives"' [ConfigParser.py:get:340] (NoOptionError)
$> echo $?
1
```

yarikoptic:
Closes #1336

### DIFF
--- a/datalad/distribution/tests/test_update.py
+++ b/datalad/distribution/tests/test_update.py
@@ -131,6 +131,10 @@ def test_update_fetch_all(src, remote_1, remote_2):
         f.write("different file load")
     rmt2.add("second.txt", git=True, commit=True, msg="Add file to git.")
 
+    # Let's init some special remote which we couldn't really update/fetch
+    ds.repo.init_remote(
+        'datalad',
+        ['encryption=none', 'type=external', 'externaltype=datalad'])
     # fetch all remotes
     ds.update(fetch_all=True)
 
--- a/datalad/distribution/tests/test_update.py
+++ b/datalad/distribution/tests/test_update.py
@@ -132,9 +132,10 @@ def test_update_fetch_all(src, remote_1, remote_2):
     rmt2.add("second.txt", git=True, commit=True, msg="Add file to git.")
 
     # Let's init some special remote which we couldn't really update/fetch
-    ds.repo.init_remote(
-        'datalad',
-        ['encryption=none', 'type=external', 'externaltype=datalad'])
+    if not os.environ.get('DATALAD_TESTS_DATALADREMOTE'):
+        ds.repo.init_remote(
+            'datalad',
+            ['encryption=none', 'type=external', 'externaltype=datalad'])
     # fetch all remotes
     ds.update(fetch_all=True)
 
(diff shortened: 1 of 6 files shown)


bpoldrack:
I'd rather see us use `AnnexRepo.get_remotes(exclude_special_remotes=True)` for that, then implementing it twice.

yarikoptic:
oh -- forgot about this one, and I thought that I have commented on (or may be we just talked about during the hangout).  here it is done in the base class to avoid complexity of making it as effective (warnings if no url but not special remote) and without explicit checking for the class type at the point of invocation (which would be ugly).  

meanwhile -- fixed one broken with additional init test, should be ok to merge unless strongly against it ;)

datalad/distribution/tests/test_update.py
```diff
@@ -131,6 +131,11 @@ def test_update_fetch_all(src, remote_1, remote_2):
         f.write("different file load")
     rmt2.add("second.txt", git=True, commit=True, msg="Add file to git.")
 
+    # Let's init some special remote which we couldn't really update/fetch
+    if not os.environ.get('DATALAD_TESTS_DATALADREMOTE'):
```
bpoldrack: Actually, we agreed on not directly using `os.environ` anymore, but use `ConfigManager` to obtain `datalad.tests.dataladremote` instead (should be available via `datalad.api.cfg`, I think). And have this documented in `common_cfg.py`.

mih:
Exactly the same changes are necessary to `add-sibling` which needs to test in the same way, and does it, of course, in a slightly different way. We really need to standardize on a common function. If the current standard getter function is not fit for this purpose, it should change. We already have an incomprehensible number of special cases.

mih:
IMHO this patch should go into `GitRepo.get_remotes()`, which should get something like a `exclude_nourl` argument (also added to `AnnexRepo.get_remotes()`).

datalad/distribution/tests/test_update.py
yarikoptic: took me a while to find where we keep those guys under `datalad.interface.common_cfg`, will add this flag which was used only in travis and some tests

yarikoptic: btw -- I am still confused -- are we using **any** of those config settings for tests ???  `git grep '\.get.*datalad.test'` returned nothing to the point

yarikoptic:
@mih -- added/used new option for get_remotes as you have suggested just with a different name consistent with the other option name in the GitRepo class invocation ;)

yarikoptic:
I am really not sure what kind of coolaid coverage/codecov is having but I would like it too: 
```Merging #1337 into master will increase coverage by 4.42%.``` ;-)

yarikoptic:
@bpoldrack - has I sufficiently addressed your requested changes so we could merge it? ;)

datalad/distribution/tests/test_update.py
mih: There is no additional harm in ignoring it again. But you are the one adding most of the environment switches. And you don't use the config mechanism hence the situation you observe will never change.

mih:
Ok, I will merge this -- and we will finally have 132% coverage!

PR #1340, opened 2017-02-24
ENH: Make direct mode tests for already cleaned modules non-optional

bpoldrack:


bpoldrack:
Hm. Despite on build failed during `pip install`, it generally seems to work, but:
The build this is about, apparently ran `datalad/support` only, instead of `datalad/tests` _and_ `datalad/support`.
Any ideas, @yarikoptic , @mih?

yarikoptic:
something is weird: look at that newly added run:
https://travis-ci.org/datalad/datalad/jobs/204985254
it never completed according to log ending with

```datalad.support.tests.test_annexrepo.test_AnnexRepo_get_outofspace ... ```

I will restart it again

yarikoptic:
`eval` didn't help... so it looks like the first argument it associates with 'coverage' options (thus reports coverage for datalad/tests) while passes only the 2nd one to nosetests

yarikoptic:
ah damn -- now I have noted that you hae removed 'datalad' entirely -- that is indeed an option to coverage which package to cover -- keep it and just add taht new variable with whih submodules to test (i.e. `datalad` 2nd time in case of full run), and remove `eval` ;-)

yarikoptic:
or I will do it now to expedite....

bpoldrack:
Ah - okay. That's confusing. How did `datalad` make it into `nosetests` before, if this was passed to coverage? And shouldn't it be the same for both then?
However, it's running the tests it is supposed to now and coverage for this build is not of particular interest, I guess. Thus: Merging.

yarikoptic:
See my comments above... Before it was picking up all tests it could find in current directory... So may be we also should change default one to be . not datalad to guarantee that we don't miss any among examples etc... I will do and check in my other PR where to not run some tests for non master

PR #1342, opened 2017-02-24
ENH: trying to skip many matrix definitions if requested from PR

yarikoptic:
also adjusted to run all possible tests


yarikoptic:
ok -- seems to work and half of the matrix runs run only a few seconds, so theoretically should cut average travis run time in half or so.  please merge if do not find any oddities so we could check if works as prescribed when not in PR

PR #1346, opened 2017-02-27
ENH: crawl openneuro bucket, various little fix ups while crawling it ;)

yarikoptic:


yarikoptic:
This PR is indeed messy but at least now passes existing tests ;)  would like to merge it as is for now ;-)

PR #1348, opened 2017-02-28
WIP: RF: new decorator approach

bpoldrack:
DON'T MERGE

This is preparing the evaluation of return values using a decorator and related stuff by developing the basic mechanisms. It will be integrated in PR #1350 

- [x] first, make use of the replacement for `functools.wraps` in order to allow for additional decorators for our commands
- [x] provide a rudimentary decorator for dealing with return values
- [x] solve python 3 issues
- [x] enhance `eval_results` to be able to receive its own arguments
- [x] additional decorator to build the docstrings right there instead of during import of `datalad.api`
- [ ] RF to use build_doc everywhere and remove that part from `datalad.api._generate_api`

### DIFF
--- a/datalad/api.py
+++ b/datalad/api.py
@@ -105,32 +105,25 @@ def _generate_func_api():
                 if always_render:
                     globals()[get_api_name(intfspec)] = intf__
 
-
-def _fix_datasetmethod_docs():
-    """Fix up dataset methods docstrings which didn't get proper docs
-    """
-    from six import PY2
-    for attr in dir(Dataset):
-        try:
-            func = getattr(Dataset, attr)
-            orig_func = getattr(func, '__orig_func__')
-        except AttributeError:
-            continue
-        if PY2:
-            func = func.__func__
-        orig__doc__ = func.__doc__
-        if orig__doc__ and orig__doc__.strip():  # pragma: no cover
-            raise RuntimeError(
-                "No meaningful docstring should have been assigned before now. Got %r"
-                % orig__doc__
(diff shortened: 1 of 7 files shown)


bpoldrack:
Changed to WIP again, since PY3 will take a little bit more time than expected

bpoldrack:
Note: build_doc decorator approach doesn't really work yet. If we do it in the part that's executed when decorated function is called, we do it over and over again and the docstring is available after first execution only.
If we do it in the part that's executed when the decorator is applied, the class we need to read from is being build in this very moment and not available from the module yet.

bpoldrack:
Re my comment above:

Solved, I think. For building the docs, we need to decorate the class instead of its `__call__` method.

bpoldrack:
Except for RF'ing to make use of these solutions, that's it for now, I think.
Since this PR is for showing the mechanics and will go into #1350 , this might be a good point to have a look at the diff and check out how it works and is supposed to be used, @mih , @yarikoptic .

bpoldrack:
Note, that failure is again a single build which terminated during setup. Seems to happen to random travis builds ATM.

yarikoptic:
note that it needs merge conflicts addressed

mih:
Given the enormity of #1350 s diff. I am not absolutely confident that all of these changes went into #1350. @bpoldrack can you give a second opinion? If they are all in, we could close this.

bpoldrack:
Second opinion: I'm confident it's all in there.

PR #1349, opened 2017-02-28
BF+RF: publish - warrant "wanted", annex_copy_opts is not annex_copy_options

### LINKED ISSUE #1274: publish seems to publish data "by default" now?
So -- did manage to create remote sibling (see #1273 ), noted btw that none of the 'push' settings aren't yet set in .git/config for that remote (thought that we have had that settled already, I guess I was wrong), and decided to publish
```shell
(venv)mvdoc@smaug /mnt/btrfs/dbic/inbox/bids (master) $ cat Wheatley/Beau/.git/config            
[core]
        repositoryformatversion = 0
        filemode = true
        bare = false
        logallrefupdates = true
[submodule "1007_personality"]
        url = /mnt/btrfs/dbic/inbox/bids-test/Wheatley/Beau/1007_personality
[remote "rolando"]
        url = ssh://rolando.cns.dartmouth.edu/inbox/BIDS/Wheatley/Beau
        fetch = +refs/heads/*:refs/remotes/rolando/*
(venv)mvdoc@smaug /mnt/btrfs/dbic/inbox/bids (master) $ cd -
/mnt/btrfs/dbic/inbox/bids/Wheatley/Beau/1007_personality
(venv)mvdoc@smaug /mnt/btrfs/dbic/inbox/bids/Wheatley/Beau/1007_personality (master) $ datalad publish --to=rolando -r                                                                                                
[INFO   ] Publishing <Dataset path=/mnt/btrfs/dbic/inbox/bids/Wheatley/Beau/1007_personality> to rolando 
[INFO   ] Publishing data of dataset <Dataset path=/mnt/btrfs/dbic/inbox/bids/Wheatley/Beau/1007_personality> ... 
```
and now staring at the terminal (oh those lovely progress bars -- where are you?) and data meanwhile is flowing...
here actually it is intended for data to eventually being transfered but I do not think I have set it up for that and feels like logic in datalad some where is flawed... I will check whenever it is done copying on some other repository I guess or there is some obvious explanation?

### LINKED ISSUE #1293: publish PATH doesn't copy PATH if nothing was pushed
```shell
$> builtin cd /tmp/; chmod +w -R test-publish*; rm -rf test-publish*; datalad create test-publish; cd test-publish; echo "123" > 123.dat; datalad add 123.dat; datalad create-sibling localhost:/tmp/test-published; datalad publish --to=localhost;
...
$> echo 124 > 124.dat; datalad add 124.dat; git annex sync
... # pushes all the branches etc

$> datalad publish 124.dat
Nothing was published
# and indeed so
$> file -L ../test-published/124.dat; ls -ld ../test-published/124.dat
../test-published/124.dat: cannot open `../test-published/124.dat' (No such file or directory)
lrwxrwxrwx 1 yoh yoh 116 Feb 14 10:18 ../test-published/124.dat -> .git/annex/objects/V4/80/MD5E-s4--df8b712c4fe20a0df933819665770165.dat/MD5E-s4--df8b712c4fe20a0df933819665770165.dat
```

### LINKED ISSUE #1329: unclear decision making of publish on what to actually publish
- (note) can't make it just rerun annex copy (may be with new options) since it seems to not get there since "nothing to publish"
- just keeps publishing the same sub-dataset (among many) over and over again for some reason
- it seems that '--since' is not quite functional since even if I specify some really old commit -- it just keeps publishing the same dataset over again
- seems to fail to detect that I am already in the dataset and complains " Modification detection (--since) without a base dataset is not supported"

here is a full protocol of the struggle:
```shell
$ datalad --version
datalad 0.4.1.dev533

(dev)user@smaug /mnt/btrfs/dbic/inbox/bids (master) $ datalad publish --to=rolando -r --annex-copy-opts ' --fast'
[INFO   ] Publishing <Dataset path=/mnt/btrfs/dbic/inbox/bids> to rolando
No handlers could be found for logger "git.remote"
[INFO   ] Publishing <Dataset path=/mnt/btrfs/dbic/inbox/bids/gitds3> to rolando
[INFO   ] Publishing <Dataset path=/mnt/btrfs/dbic/inbox/bids/gitds1/gitds2/annexds> to rolando
[INFO   ] Publishing data of dataset <Dataset path=/mnt/btrfs/dbic/inbox/bids/gitds1/gitds2/annexds> ...
3 items were published:
Dataset: /mnt/btrfs/dbic/inbox/bids
Dataset: /mnt/btrfs/dbic/inbox/bids/gitds3
Dataset: /mnt/btrfs/dbic/inbox/bids/gitds1/gitds2/annexds

(dev)user@smaug /mnt/btrfs/dbic/inbox/bids (master) $ datalad publish --to=rolando -r --annex-copy-opts ' --fast'
[INFO   ] Publishing <Dataset path=/mnt/btrfs/dbic/inbox/bids> to rolando
[INFO   ] Publishing <Dataset path=/mnt/btrfs/dbic/inbox/bids/gitds1/gitds2/annexds> to rolando
[INFO   ] Publishing data of dataset <Dataset path=/mnt/btrfs/dbic/inbox/bids/gitds1/gitds2/annexds> ...
2 items were published:
Dataset: /mnt/btrfs/dbic/inbox/bids
Dataset: /mnt/btrfs/dbic/inbox/bids/gitds1/gitds2/annexds

(dev)user@smaug /mnt/btrfs/dbic/inbox/bids (master) $ datalad publish --to=rolando -r --annex-copy-opts ' --fast'
[INFO   ] Publishing <Dataset path=/mnt/btrfs/dbic/inbox/bids/gitds1/gitds2/annexds> to rolando
[INFO   ] Publishing data of dataset <Dataset path=/mnt/btrfs/dbic/inbox/bids/gitds1/gitds2/annexds> ...
1 item was published:
Dataset: /mnt/btrfs/dbic/inbox/bids/gitds1/gitds2/annexds

(dev)user@smaug /mnt/btrfs/dbic/inbox/bids (master) $ datalad publish --to=rolando -r --annex-copy-opts ' --fast'
[INFO   ] Publishing <Dataset path=/mnt/btrfs/dbic/inbox/bids/gitds1/gitds2/annexds> to rolando
[INFO   ] Publishing data of dataset <Dataset path=/mnt/btrfs/dbic/inbox/bids/gitds1/gitds2/annexds> ...
1 item was published:
Dataset: /mnt/btrfs/dbic/inbox/bids/gitds1/gitds2/annexds

(dev)user@smaug /mnt/btrfs/dbic/inbox/bids (master) $ datalad publish --to=rolando -r --annex-copy-opts ' --fast' --since=
[INFO   ] Publishing <Dataset path=/mnt/btrfs/dbic/inbox/bids/gitds1/gitds2/annexds> to rolando
[INFO   ] Publishing data of dataset <Dataset path=/mnt/btrfs/dbic/inbox/bids/gitds1/gitds2/annexds> ...
1 item was published:
Dataset: /mnt/btrfs/dbic/inbox/bids/gitds1/gitds2/annexds


(dev)user@smaug /mnt/btrfs/dbic/inbox/bids (master) $ datalad publish --to=rolando -r --annex-copy-opts ' --fast' --since=7d95f4d31afaf0ec26cc010c06ed2d404706732f
[ERROR  ] Modification detection (--since) without a base dataset is not supported [publish.py:__call__:282] (InsufficientArgumentsError)
usage: datalad publish [--version] [-h] [-l LEVEL] [--pbs-runner {condor}]
                       [-d DATASET] [--to LABEL] [--since SINCE]
                       [--skip-failing] [--inherit-settings] [-r]
                       [--recursion-limit LEVELS] [--git-opts STRING]
                       [--annex-opts STRING] [--annex-copy-opts STRING]
                       [PATH [PATH ...]]
(dev)user@smaug /mnt/btrfs/dbic/inbox/bids (master) $ datalad publish --to=rolando -r --annex-copy-opts ' --fast' --since=7d95f4d31afaf0ec26cc010c06ed2d404706732f -d .
[INFO   ] Publishing <Dataset path=/mnt/btrfs/dbic/inbox/bids/gitds1/gitds2/annexds> to rolando 
[INFO   ] Publishing data of dataset <Dataset path=/mnt/btrfs/dbic/inbox/bids/gitds1/gitds2/annexds> ...
1 item was published:  
Dataset: /mnt/btrfs/dbic/inbox/bids/gitds1/gitds2/annexds

```

yarikoptic:
+ copy invocation independent on either anything to push

Closes #1329 ATM partially, TODO
- [x] see to that inability to detect dataset (last item)

Closes #1293 -- should copy if I provide the path although everything was pushed already
- [x] I thought that upon anything copied we should redo the push of the git-annex branch since now remote wouldn't be aware of the new copy IIRC.  BUT as test showed, at least on new versions of the remotes (init'ed with recent annex) that is not needed -- they get their git-annex branch also updated. I guess related to https://github.com/datalad/datalad/issues/1319 

Closes #1274 -- should not copy data by default now, unless paths are provided or remote has 'wanted' setting.

edit 1+3:
Formalized/changed/new behavior: to publish a dataset under `dspath` with all of its content (i.e. invoke `annex copy` on the entire dataset), there are two ways:
1. `datalad publish --to=remote .` as to instruct explicitly to publish the "root" directory of the dataset.
2. Define `wanted` for the remote (via add-sibling or create-sibing) and then can use `datalad publish --to=remote dspath` should consult `wanted` setting of the dataset and perform `copy --auto`

edit 2:
I think it might be a changed behavior now although not clear since to me two tests contradicted each other before:
- `copy_to` will return only paths to the content actually copied (datasets, or paths to files).  I.e. second invocation of exactly the same `copy_to` should return no filenames.  This also would make it more consistent with invocation when `annex copy` is called with `--fast` 

TODO:
- [x] test precisely that of 1.  ATM I believe there is no explicit test like that to do 'out of CWD' invocation pointing to the path
- [x] edit 4:  verify/make sure that `publish -r subdir` publishes submodules under `subdir` with the data.


mih:
I cannot provide proper feedback right now, but please keep in mind that all this code should preferably be ripped out completely and be replaced by a smallish frontend to `annex sync`. Any other development direction is a waste of time IMHO.

yarikoptic:
"it depends" -- if you still like to provide "publish path/to/files" functionality, iirc `sync` alone would not be sufficient anyways.  Changes I am pursuing are just to make what I think you envisioned work (i.e. fix bugs and logic just a bit), and tests to assure that it does ;)  so if a big RF would come, we could still do it.  But I think it would be valuable to create use-cases spreadsheet (actually we started already) so we just identify how we want it to behave.

datalad/distribution/publish.py
```diff
@@ -62,13 +63,23 @@ def _publish_dataset(ds, remote, refspec, paths, annex_copy_options):
...
+    #    4. push non-annex branch(es)
+    # 5. copy data to the remote if paths are provided or it wants something generally
 
     published, skipped = [], []
+
+    # pre-treat paths.  If path points to entire dataset, and not its root directory
+    # i.e. dataset or dataset/  vs dataset/. - we do not invoke copying all the data
+    # TODO: move into a helper function and test
+    paths = filter(
+        lambda p: p.rstrip(dirsep) != ds.repo.path.rstrip(dirsep),
```
bpoldrack: Apart from that, I see the intention of the filtering, but I think this is leading to inconsistency of API and is somewhat confusing. Now passed `paths` are to be published but not if they point to a dataset's root dir, but then it's nevertheless published if it ends with `[/].`? At least provide proper documentation in `Publish._params_` please.

datalad/support/tests/test_annexrepo.py
```diff
@@ -1134,7 +1134,8 @@ def test_annex_copy_to(origin, clone):
     repo.get("test-annex.dat")
     # now it has:
     eq_(repo.copy_to("test-annex.dat", "target"), ["test-annex.dat"])
-    eq_(repo.copy_to(["INFO.txt", "test-annex.dat"], "target"), ["test-annex.dat"])
+    # and will not be copied again since it was already copied
+    eq_(repo.copy_to(["INFO.txt", "test-annex.dat"], "target"), [])
```
bpoldrack: This is problematic, I think. You changed to use `_run_annex_command_json` within `AnnexRepo.copy_to`. But this one tends to swallow failures and to just return empty list instead. This might be fine in general but in combination with that change to the test, we don't test whether it works with a list passed into it. It could always fail on lists now and we wouldn't notice.

datalad/distribution/publish.py
yarikoptic: agreed -- changed to `ds.path` and provided some docs. will push shortly

datalad/support/tests/test_annexrepo.py
yarikoptic: 1. we should move to use `--json` mode anyways -- I have some unfinished branch already to introduce progress reporting
2. tests still pass -- I can "fix" or refactor envisioning only the cases covered by tests.  So "swallow failures" behavior was never tested I guess?
3. if there is something is not working in `--json` mode we must not just stay with non-json mode, but ask joey to provide desired behavior.  Now there is  http://git-annex.branchable.com/bugs/copy_does_not_reflect_some_failed_copies_in_--json_output/?updated   -- what other errors you experienced besides non-existing files ? note that exit status is non-0 with --json or not, and in our case prior code would fail before even calling out to git-annex:
```shell
$> python -c 'from datalad.api import Dataset; ds = Dataset("."); ds.repo.copy_to(["nonexisting"], "neurosmaug")'
  File "/home/yoh/proj/datalad/datalad/datalad/support/gitrepo.py", line 296, in newfunc
    result = func(self, files_new, *args, **kwargs)
  File "/home/yoh/proj/datalad/datalad/datalad/support/annexrepo.py", line 1994, in copy_to
    self.get_file_key(files[0])
  File "/home/yoh/proj/datalad/datalad/datalad/support/gitrepo.py", line 296, in newfunc
    result = func(self, files_new, *args, **kwargs)
  File "/home/yoh/proj/datalad/datalad/datalad/support/annexrepo.py", line 953, in get_file_key
    raise IOError(e.code, "File not found.", files)
IOError: [Errno 1] File not found.: 'nonexisting'
```
so whatever errors you expect to be caught -- we better hard-coded expected behavior in the tests!

yarikoptic: ok -- I see the specific case you might be having in mind -- we are pretty much ignoring non-0 exit status of --json mode exit... so we shouldn't ?

yarikoptic: I could make it behave as before (in non --json mode) I guess, at least for now for consistency

yarikoptic: funny enough, current master's publish just reports that nothing was published, which is factually correct ;)  but doesn't fail!!!  direct invocation to `repo.copy_to` does raise an CommandError

yarikoptic: I guess all the roads lead to "returned values" issue, and I am yet to checkout @mih's PR on that.  So, for now, I guess, I will just try to make it behave as master does and hardcode that  behavior in a test

yarikoptic: not sure if overall to fail is the way to go, but here you are ;) dae8de81f49b9bec864deae354575fedee10fe36

yarikoptic: in c4c3a8187eb00e43804eeacdaf1447d2c68a587d added a (mocked up) test for the exception being thrown, and changed exception to be IncompleteResultsError.  I think final resolution could/should wait until we decide on return values

yarikoptic:
Heh, making `/.` handling is a bit trickier than I thought due to all the pre-treatment of paths.  Not sure if I should push forward my quick and dirty attempt unless we agree that it the semantic we want to pursue.  But then if we just reserve publishing with path pointing to a dataset to be the publishing without data, how to publish entire dataset with the data without setting up `wanted` for the remote?  and definetely we can't stay with current implementation which would copy all of the data, since then how to publish without? ;)  either we need to use this 'path/' presentation (somewhat similar to rsync) or add some option.  ideas?

datalad/support/tests/test_annexrepo.py
bpoldrack: Well, actually that's not exactly what I meant. (Although I'm fine with that). I'm concerned about a possible non-zero exit - yes. But this could happen by mistransforming a passed list for example.
Or just respecting the first entry or not existing files, etc.
Your change in that single test line means, that we no longer test, whether we can successfully pass a list, because the only assertion involving a list of files is this one and expects an empty result. So, theoretically the function could be unable to deal with a list at all and simply always return an empty result when a list was passed. We wouldn't notice now. So, what I actually wanted to see is just an invocation, that uses a list and expects something in the results. ;-)

And since the `_run_annex_json` behaves differently I'm not sure how it now behaves on not existing files or files not in annex, etc. Before we just got a non-zero exit from annex and the respective `CommandError`. We didn't need to test this explicitly, we simply relied on git-annex' non-zero exit. Now, we have no clue what happens in such cases and I think, git-annex itself behaves differently with `--json`. So actually we need to test for it now. Your mockup doesn't do that.

So just two test lines, please:
 - pass a list, that should result in something else than doing nothing
 - try to copy a not existing file
A file not in annex we already tried to copy (`INFO.txt`), so that's fine, if we have make sure that the list call works at all.

yarikoptic: thanks @bpoldrack -- good arguments... will address them whenever get to this 
as for --json and non-existing files -- see http://git-annex.branchable.com/bugs/copy_does_not_reflect_some_failed_copies_in_--json_output/  which is relevant here... I guess while processing stderr coming from `--json` we need also to catch those :-/

bpoldrack: Yes, we already started to deal with that kind of stderr parsing in `_run_annex_command_json`, I think, but it starts getting messy without a proper design for that ...

bpoldrack:
Re `/.` handling:

Actually I don't think we urgently need this distinction. If you want to publish the dataset itself, you can `-d` option instead `path` (plus `--recursive` may be). It's relevant only if we want to publish several datasets without data, that are not related to each other in a way expressible via `-d` and `-r`.
Additionally: If we want to publish the data if `path` is a directory, but not if it's also a submodule, then this might lead to even more confusion.

yarikoptic:
@bpoldrack -- addressed your demand for testing a list being passed (via a gentle mocking too ;)) see https://github.com/datalad/datalad/pull/1349/commits/14db1f7a50142edbf1562eda548e726ab37d9a5a#diff-4ebc4f0a30935b2a6339fd61d99a8a1fR1141

But overall just arrived to show that [we do copy data](https://github.com/datalad/datalad/pull/1349/commits/03f856e3c54246ed7a3d61405b968a9a9803feeb#diff-70bbcdc6e3b21e440b6d8a4ec991ebeeR218) while publishing sub-datasets **by default**.  My wild guess it is because of all the sorting of paths and assigning by default either `.` or may be full dataset path into the sorted paths -- which leads to the publish call for the sub-dataset with the path specified as if we wished to publish entire thing with the data... heh.

yarikoptic:
ok - this smells like a partial success (nd90 error is due to gitpython exc issue again... bleh -- irrelevant)

overall I feel that we are in the same situation as we used to be with the "magical install" before we split it into install/get with clear(er) semantics.  Now with publish we are trying to mix "predefined rules" (my invocation of 'annex copy --auto') with explicit, but not scalable implicit semantic (upload data pointed by paths).  Why not scalable?  tell me how to "publish" without copying data all sub-datasets under path `ds/subdir`?

I feel that most probably introduce an option explicitly stating the mode:
   - "with data" (paths are considered to be the ones to publish data for) (kinda like a current one)
   -  "auto" -- publish according to the set 'wanted' rules, and nothing if no rule set (also current one if no paths provided)
   -  "no data", or some explicit expression (`--annex-copy-opts`) on what data to publish.  

Either "auto" or "no data" should be default.  Providing paths shouldn't be defining the mode (as it is now) since that is somewhat orthogonal from the intent -- paths merely provide the boundary/targets for what to publish.
@mih, @bpoldrack -- what do you think?

datalad/support/tests/test_annexrepo.py
yarikoptic: all implemented and tested now (yet to push one more tiny fix and get to the last TODO ;) )

yarikoptic:
ok -- I think this one is ready for the merge @datalad/developers .  Future RF extending interface could make it a bit more explicit.  I call current behavior a "hybrid" one ;-)

bpoldrack:
Re `Dataset` not hashable:
That's not direct mode. It's the PY3 build!

datalad/distribution/tests/test_publish.py
```diff
@@ -315,13 +367,21 @@ def test_publish_depends(
...
     ok_file_has_content(opj(src_path, 'probe1'), 'probe1')
     for p in (target1_path, target2_path, target3_path):
-        assert_false(exists(opj(p, 'probe1')))
+        assert_false(lexists(opj(p, 'probe1')))
     # publish to a standalone remote
     source.publish(to='target3')
-    # no others are affected
+    ok_(lexists(opj(target3_path, 'probe1')))
+    # but it has no data copied
+    ok_(not exists(opj(target3_path, 'probe1')))
```
bpoldrack: We need a `file_has_content` here. `exists` is true without content on "crippled filesystems".

bpoldrack: Otherwise I'm fine with a merge, but this kind of assertions are unnecessarily adding effort to make it pass in direct mode.

yarikoptic: +1 for file_has_content
this assertion was to test that data was not copied (can't just rely on return values), so it on purpose

yarikoptic:
ok, addressed test, if no further comments -- would appreciate a merge or will merge in the morning

PR #1350, opened 2017-02-28
Return value revolution

### LINKED ISSUE #819: New general option `on_failure` to indicate action when something goes wrong
Currently, we sometimes fail, sometimes we continue. I guess it is time for making it explicit and to stick to it.

mih:
Not yet complete (see list below), but should give a reasonable impression of the approach. Basic idea: commands `yield` report dictionaries. And wrapper act upon this information in whatever desired way.

Sits on top of #1348.

````
% datalad update --merge 2>/dev/null
{"action": "update", "path": "/tmp/local", "status": "ok", "type": "dataset"}
````

which is actually implemented as a generator all the way up to the entrypoint -- through layers of decorators.

The interesting parts of the diff (beyond #1348) are in `datalad/interface/utils.py` and `datalad/distribution/update.py`.

Change summary:

- [x] global command line switch to select output format
- [x] line-based 'json' output format (like git-annex)
- [x] 'simple' output format for "STATUS: PATH" lines, where PATH is relative whenever a reference dataset is available
- [x] supported commands return generators that are actually executed in main.py
- [x] centralized error detection that guarantees an exception if any error occured
- [x] BF: `update` did not fail on unavailable paths as argument
- [x] customize exception generation to be less dumb
- [x] centralize logging by shipping messages in the status dicts (actually: message template, variable, logger instance or func)
- [x] reimplement forced 'cmdline' rendering of API_ function with the new facilities to get rid of these shadow functions
- [x] convert actual return value nature on a by-call basis. Standard should be json dict sequence/generator.
- [x] provide `return_value` helper module with:
      - function to give a default dict
      - function to convert git-annex json record into our own structure
- [x] implement new constraint to test whether a dict has a key with a certain value 
- [x] implement general command line options to filter results by `type` and by `status` (using new constraint above)
- [x] implement new general kwarg for all commands to accept a filter function for results (to fit a passed (compound) constraint) (name: `filter_results`)
- [x] implement new general kwarg for all commands to switch return values from generator to sequence (spec: `return_type`, `{'generator', 'list', 'item-or-list'}`)
- [x] implement flexible transformation of return values from result dicts to other types (like Datasets)
- [x] implement general `--on-failure` switch (fixes #819)
- [x] new test helpers: `assert_status`, `assert_message`, `assert_in_results`, `assert_not_in_results`, `assert_result_count`, `assert_fields_equal`
- [ ] #1380 support for result summaries
 
Commands converted:

- [x] create (possibly TODO: report addition to a superdataset)
- [ ] install -- this needs a volunteer, you guessed it
- [x] get -- needs workaround for dirty interfacing to `install`, otherwise needs RF for old-school recursion in helper functions)
- [ ] add (complicated due to the old-school recursion in helper functions, needs substantial RF)
- [ ] publish (seems straightforward)
- [x] uninstall (needs additional RF, PY2 version of Ben's class detector cannot deal with more than one class per module)
- [x] drop (needs additional RF, PY2 version of Ben's class detector cannot deal with more than one class per module)
- [x] remove (needs additional RF, PY2 version of Ben's class detector cannot deal with more than one class per module)
- [x] update (cmd itself is done, but needs to pass though results from `get`)
- [ ] create-sibling (seems doable, unclear if RF needed to pass `add-sibling` results through)
- [ ] create-sibling-github (doable)
- [ ] add-sibling (maybe better wait till after general RF; convoluted already)
- [ ] rewrite-urls (does not seem to be used anywhere -> low priority; but otherwise simple)
- [x] unlock
- [x] save
- [ ] export (seems doable, but also rather useless)
- search (will not be attempted, waiting for metadata RF)
- aggregate-metadata (will not be attempted, waiting for metadata RF)
- test (not needed)
- crawl, crawl-init (no idea if it is feasible to attempt this, better be done by Yarik, also crawl-init about to be gone #1155)
- ls (will not be attempted, does not fit paradigm as of now, but could become relevant after RF)
- [x] clean (looks straightforward)
- [ ] add-archive-content
- [ ] download-url (seems straightforward, custom stop_on_failure needs to be RF'ed away)
- [ ] create-test-dataset
- sshrun (out of scope, needs to stay as it is)

### CODE REVIEW

**Comment by mih** [comment]
Rebased on master

**Comment by mih** [comment]
Test seem to pass now. Remaining failure is due to the unresolved PY3 issue from #1348.

**Comment by bpoldrack** [comment]
Note:
#1348 now contains drafts to provide arguments for `eval_results` (see last point in TODOs here) and to provide another decorator `build_doc`, which generates the docstrings for the `__call__` methods when they are build, so we don't need to do it when importing `datalad.api` and have it available no matter how we got to know the functions.

File: datalad/interface/utils.py [hunk 1]
```diff
@@ -761,3 +765,105 @@ def filter_unmodified(content_by_ds, refds, since):
...
+    set of error a compound exception will be raised that includes information
+    on all individual errors that occurred.
+
+    Status messages will be logged automatically, by default the following
+    association of result status and log channel will be used: 'ok' (debug),
+    'notneeded' (debug), 'impossible' (warning), 'error' (error).  Logger
+    instances included in the results are used to capture the origin of a
+    status report.
+    """
+    want_generator = dlcfg.getbool('datalad.api', 'return-generator', False)
```
yarikoptic: mental note to myself possibly
- either all decorated commands return a generator, or just return a list. all the time
- we will need travis setup to test correct internal operation in both modes
- since decision is made once for the entire process, usefulness is probably only for command line invocations of separate datalad commands 
- I do not think there is an easy way (if any?) to make it 'run time' configurable per a command invocation by adding some kwarg to the function which decorator would consume and either return a generator or the list of results... actually, why not?
```python
$> python /tmp/f.py
<generator object gen_func at 0x7f7d584f08c0>
[1, 2]

$> cat /tmp/f.py

def gen_func():
   yield 1
   yield 2

def dynamic(f, what):
   res = f()
   return res if what == 'generator' else list(res)

print dynamic(gen_func, 'generator')
print dynamic(gen_func, 'list')

```
so probably it could be made configurable at run time by decoration checking some kwarg which could augment default behavior... or I am missing smth

datalad/utils.py
```diff
@@ -21,6 +21,7 @@
 import platform
 import gc
 import glob
+import wrapt
 
 from contextlib import contextmanager
 from functools import wraps
```
yarikoptic: do we still need this decorator if we are to use wrapt one?

datalad/distribution/update.py
```diff
@@ -98,44 +100,70 @@ def __call__(
...
                 continue
             if not sibling_ and len(remotes) == 1:
                 # there is only one remote, must be this one
                 sibling_ = remotes[0]
             if not sibling_ and len(remotes) > 1 and merge:
                 lgr.debug("Found multiple siblings:\n%s" % remotes)
-                raise NotImplementedError(
+                res['status'] = 'impossible'
+                res['error'] = NotImplementedError(
                     "Multiple siblings, please specify from which to update.")
```
yarikoptic: while at it -- can include a list of those multiple siblings here I guess ;-)

datalad/distribution/update.py
```diff
@@ -154,13 +183,17 @@ def _update_repo(ds, remote, merge, fetch_all, reobtain_data):
         if reobtain_data:
             # get all annexed files that have data present
             lgr.info('Recording file content availability to re-obtain update files later on')
-            reobtain_data = repo.get_annexed_files(with_content_only=True)
+            reobtain_data = \
+                [opj(ds.path, p)
```
yarikoptic: why do we need full path here if later it is used in  `ds.get` anyways which would probably strip those off directly or indirectly (by annex)?

datalad/interface/utils.py
bpoldrack: I introduced dedicated keyword arguments for eval_results in #1348 and am about to integrate it in this one. So, yes, the mechanic for such dynamic tuning of behavior is already built.

datalad/utils.py
bpoldrack: Actually ATM we don't need it. But having it provides an easy way to improve all existing decorators without the need to rewrite them. So I'd vote for keeping it.

datalad/interface/utils.py
```diff
@@ -761,3 +765,105 @@ def filter_unmodified(content_by_ds, refds, since):
...
+    `yields` individual results, and 2) "list mode" that returns a sequence of
+    results. The behavior can be selected via the
+    `datalad.api.return-generator` configuration variable. Default is "list
+    mode".
+
+    This decorator implements common functionality for result rendering/output,
+    error detection/handling, and logging (TODO).
+
+    Result rendering/output can be triggered via the
+    `datalad.api.result-render-mode` configuration variable. Supported modes
```
mih: re by-call: done now

datalad/interface/utils.py
```diff
@@ -761,3 +765,105 @@ def filter_unmodified(content_by_ds, refds, since):
...
+    are: 'json' (one object/dict per line, like git-annex), 'simple' (TODO,
```
mih: Or I remove that statement...

datalad/distribution/update.py
mih: Ah, four days ago... no longer at it... SAD ;-)

datalad/distribution/update.py
mih: because we test later on if that thing still exists after upgrade? I don't think it represents a bottleneck of the entire operation in any case.

mih:
FYI: the only thing todo in this PR is to test one nested invocation of commands (update->get). All other open points in the TODO list are either dependent on all commands being fully converted, or I have no idea how to do them better than the status quo (exception generation).

mih:
I give up on `get` for now -- too many hack in the way it talks with `install` -- nightmare to replace the return values with something uniform. Looking for another command...

yarikoptic:
> I give up on get for now -- too many hacks in the way it talks with install -- nightmare to replace the return values with something uniform. Looking for another command...

my 1c: well... many of other commands talk to other commands in various ways already too -- if you wouldn't be able to figure it out for `get` now, unlikely anyone would later and then we might then introduce yet another approach for handling return values which would not generalize beyond "simple" commands.

mih:
hehe, wise words... check the actual signature of `get` .. especially towards the end and you know exactly what I am talking about. This is not talking to other commands, this is groping...

yarikoptic:
so... wasn't the idea of ultimate return value revolution to provide such needed flexibility?  in this particular case `_return_datasets` was used only in a single spot, so easy to track intent/ramifications

mih:
Indeed, so I'll just fix up `install` to become a generator quickly. Plus
some minor changes. What could go wrong....

mih:
Travis test runs core dump, some buildbots (but not all) too. Obviously I cannot replicate it locally....

yarikoptic:
You can login to Smaug and enter one of the buildbot dockers where core dump is happening

mih:
I want to highlight one of today's changes for you guys: This PR enables flexible, by-call, configuration of how commands shall behave:

- raise or not raise on incomplete results
- compact one-item list results to a single item
- transform results from status dicts to actual objects (like datasets)

All this can now be configures in the classes, and be overriden on a by-call basis. This allows us to have commands behave in substantially different ways in the Python API v the cmdline. But also to pull out information for testing without having to fish it out of the logging output.

mih:
@yarikoptic @bpoldrack Apart form the following issues this PR is done IMHO:

1) this weird test crash on travis, OSX, and nd90 (but nowhere else, including my local machine that is nd90+ and also runs SSH tests etc...) -- I was hoping we could make an attempt to dissect this today
2) not all commands have been converted. However, many have been, this PR is complicated enough, and all interfaces (Python API, cmdline) and interface bridges are exercised in the PR (i.e. `install(get)` with old-style outside, `update(get)` both new-style). I will leave further conversions to the moment when we touch the remaining commands next.

mih:
Ah forgot one: `get` needs RF for its `tailored` renderer.

mih:
Nice, conflict with master. We reached the end of the line for this one.

mih:
I am reverting the merge of recent master, as it breaks this PR (see bottom of #1370).

mih:
Story continues at #1393

PR #1351, opened 2017-03-01
General config overrides mechanism and `-c` option for the cmdline

mih:
Addresses #958 

At present we have no simple way to test the few lines in `main.py`.

### DIFF
--- a/datalad/cmdline/main.py
+++ b/datalad/cmdline/main.py
@@ -101,6 +101,11 @@ def setup_parser(
         parser.add_argument(
             '--idbg', action='store_true', dest='common_idebug',
             help="enter IPython debugger when uncaught exception happens")
+    parser.add_argument(
+        '-c', action='append', dest='cfg_overrides', metavar='KEY=VALUE',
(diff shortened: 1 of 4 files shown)


mih:
Buildbot failure is due to the way the tests are executed there, i.e.

````
python -c 'import nose; nose.main()' -s -v datalad
````

it considers `-c` as `sys.argv[0]` and now there is an actual `-c` option in datalad. Can we use the same test invocation in the buildbots that we use on travis, i.e. via a call to `nosetests`? @yarikoptic

mih:
These changes are now a requirement for #1350.

yarikoptic:
how come travis is happy and all buildbots aren't?

mih:
@yarikoptic the answer is in the message above. And in #1353

mih:
#1353 now merged into this one.

mih:
OK, travis and buildbots (minus the ones with general trouble) are OK. Good to merge. Count down started. Veto?

mih:
Gracie!

PR #1353, opened 2017-03-01
BF: Don't use unecessary and unreliable sys.argv[0]

bpoldrack:
Addresses buildbot failures of #1351 

@yarikoptic : Please note this change.

### DIFF
--- a/datalad/cmdline/tests/test_main.py
+++ b/datalad/cmdline/tests/test_main.py
@@ -111,7 +111,22 @@ def test_subcmd_usage_on_unknown_args():
 
 
 def check_incorrect_option(opts, err_str):
-    stdout, stderr = run_main((sys.argv[0],) + opts, expect_stderr=True, exit_code=2)
+    # The first line used to be:
+    # stdout, stderr = run_main((sys.argv[0],) + opts, expect_stderr=True, exit_code=2)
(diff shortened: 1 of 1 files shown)


bpoldrack:
Okay, for one thing there's sth different with PY3 again (I f** hate it) ;-)

But more important: @yarikoptic ths failures in nd90 and osx buildbots are the very same as in #1277 .
Something (or someone?) messed with them.

yarikoptic:
nd* buildbots were recently upgraded, otherwise nobody touched them, so it might be that some new release of our depends changed the behavior.  

comment says:

> But sys.argv[0] actually isn't used by main at all

which is imho not exactly correct. Those args passed into the argparser entirely -- so it is for it to "treat" them, and I am not sure how stripping the very first argument there would affect it -- I guess it might result in swallowing next one treating as the command which was executed.  I wondered, since this is just for the tests, if we shouldn't just make it "datalad" and be done

```
-    stdout, stderr = run_main((sys.argv[0],) + opts, expect_stderr=True, exit_code=2)
+    stdout, stderr = run_main(("datalad",) + opts, expect_stderr=True, exit_code=2)
```

I have tested locally -- tests pass

mih:
This is fixed and part of #1351 now.

PR #1354, opened 2017-03-01
BF(workaround): safeguard import of h5py in auto 

yarikoptic:
since git could have messed with str encoding handlers

should resolve nd90 weird error (although would also skip some tests I guess)


yarikoptic:
nd90 is "happy" again, so I will just merge ;-)

PR #1355, opened 2017-03-02
BF: Don't overcomplain long lines in terminal output

bpoldrack:
We are formatting the tested help output to the terminal size if we know it, which is the case in PY3 only.
(See datalad/cmdline/main.py:182)
Nevertheless we tested against a fixed width of 80. Therefore the test would fail on python 3 if ran in a terminal wider than 80 characters.
Changed the test to fit the actual implementation.

### DIFF
--- a/datalad/cmdline/tests/test_main.py
+++ b/datalad/cmdline/tests/test_main.py
@@ -90,13 +90,18 @@ def test_help_np():
     # none of the lines must be longer than 80 chars
     # TODO: decide on   create-sibling and possibly
     # rewrite-urls
+    import shutil
+    accepted_width = shutil.get_terminal_size()[0] \
+        if hasattr(shutil, 'get_terminal_size') else 80
+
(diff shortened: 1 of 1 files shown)


bpoldrack:
Note: Travis failure unrelated. Happens in setup not in the tests.

datalad/cmdline/tests/test_main.py
```diff
@@ -90,13 +90,18 @@ def test_help_np():
     # none of the lines must be longer than 80 chars
     # TODO: decide on   create-sibling and possibly
     # rewrite-urls
+    import shutil
+    accepted_width = shutil.get_terminal_size()[0] \
+        if hasattr(shutil, 'get_terminal_size') else 80
```
yarikoptic: "not sure" what we are "fixing" with this. we shouldn't test for less than 80... and not sure if valuable to test for > 80

PR #1356, opened 2017-03-04
BF: delay decision making on dataset availability for --since [Was Question: what am I doing wrong?  non-reproducible]

yarikoptic:
WAS: fetch fails when ran during the test but works fine if I run manually after... am I doing smth wrong or a bug?  testing also if our buildbot back online

Now -- delays decision making on dataset availability (also part of the #1349), tests for that, and improves error msg reporting in case of wrong since value

### DIFF
--- a/datalad/distribution/publish.py
+++ b/datalad/distribution/publish.py
@@ -302,11 +302,6 @@ class Publish(Interface):
             annex_opts=None,
             annex_copy_opts=None):
 
-        if since and not dataset:
-            raise InsufficientArgumentsError(
-                'Modification detection (--since) without a base dataset '
-                'is not supported')
-
         if dataset and not path:
             # act on the whole dataset if nothing else was specified
             path = dataset.path if isinstance(dataset, Dataset) else dataset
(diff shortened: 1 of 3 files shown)


bpoldrack:
I think it's a bug. The message is generated right at the beginning of call, which means the dataset has to be explicitly specified to operate with `--since`. This means:

1. It probably works, if you use `Dataset(p1).publish(...)` instead
2. Moving this check beneath the two following trials to determine the base dataset should fix it (and would be reasonable I think)

yarikoptic:
I now don't know what has happened that night, but IIRc it was some obscure failure from "fetch" call, unrelated to this logic -- it was happening during `create_sibling`, so having no chance to get to the publish.  I am not sure why I haven't pasted that output into the PR text :-/   well -- now I am getting a normal failure in deciding on the dataset, and then cryptic failure (apparently GitPython manages to not record/report stderr from git call) -- so I just decided for us to test explicitly the value of since and raise ValueError exception if that value found to be incorrect. Otherwise test should be passing now.... really not sure what was happening then -- hopefully I will run into it again to troubleshoot properly (meanwhile I am fixing bloody dockers... grrr)

yarikoptic:
@datalad/developers - review, merge, or stay silent! ;)

PR #1360, opened 2017-03-06
UX: Have a proper error message (Closes #1359)

### LINKED ISSUE #1359: Incomprehensible error message of `remove`
````
$ datalad remove src/aligned 
[ERROR  ]  [uninstall.py:__call__:401] (AssertionError) 
````

but the actual remove operation works like charm.

bpoldrack:


bpoldrack:
Failures: See PR #1277

yarikoptic:
so I managed to bring some older versions of dockers up I guess for buildbot  -- I am upgrading them inplace again and then will rerun in those which failed

bpoldrack:
Thanks! And: Good morning! :-)

yarikoptic:
oh well --= upgraded but lazy to go restart them all one by one... travis passes and the line seems to be covered (although not sure if message is -- interested to know if python evaluates string even if no assertion failure)so let it be

PR #1364, opened 2017-03-07
found this db.py no longer used -- kill it

yarikoptic:


mih:
Still hooked into the docs somehow, hence failure.

yarikoptic:
yeap, thanks, fixed for that... hopefully this time ;)

PR #1368, opened 2017-03-10
DOC: instructions on how to enter the buildbot environment for troubleshooting

yarikoptic:


yarikoptic:
some docs is better than no docs, so merging

PR #1369, opened 2017-03-10
BF: use swallow_output in 2 failing spots whenever nosetests ran without -s

yarikoptic:
I test to run nosetests without -s nowadays, and it seems (this PR tests run will show ;-) ) to work but few tests required workaround.


yarikoptic:
anyone against it??? ;)  or I would merge

PR #1370, opened 2017-03-14
ENH: when installing/getting a submodule, make checkout branch to point to that commit

### LINKED ISSUE #937: 'datalad install subds' installs into a detached HEAD
as super/sub-datasets progress-forward, current version of a subds might be ahead of what installed locally super-dataset knows about.  Then when 'datalad install subds' is performed, it  installs that subdataset rightfully at the known position, but because its master (remote one) is ahead, it installs into a detached HEAD:

``` shell
hopa:~/datalad/openfmri/ds000030
$> git br
* (HEAD detached at 3d9fc7e)
  git-annex
  master

$> git lg master
*   890f095 - (origin/master, origin/HEAD, master) Merge branch 'incoming-processed' (24 hours ago) [Yaroslav Halchenko]
|\  
| *   15f8bcf - [DATALAD] Added files from extracted archives (24 hours ago) [Yaroslav Halchenko]
| |\  
| | * 19feafd - [DATALAD] removed: 25 (26 hours ago) [Yaroslav Halchenko]
| * |   cff7f40 - [DATALAD] Added files from extracted archives (26 hours ago) [Yaroslav Halchenko]
| |\ \  
| | |/  
```

which of cause could be treated as a feature or a curse in terms of usability. I wonder if  we should
- add `--updated` to install so it wouldn't install version as known to super, but rather just proceeds to current state/master and possibly marking itself as 'changed' in the eyes of the super.
- should install always 'reset' master (or some other known remote branch name which known position is a predecessor of) to the position?

those all should be considered also in light that may be this RF should be joined with RF to support versioning specification for install calal

yarikoptic:
Closes #937 
but now I wonder if we should place more safe-guards, e.g. to not perform it if previous position of the branch is not available within any remote, thus signalling that it probably has local changes, thus we shouldn't just 'loose' them.

### DIFF
--- a/datalad/distribution/utils.py
+++ b/datalad/distribution/utils.py
@@ -38,6 +38,7 @@ from datalad.utils import swallow_logs
 from datalad.utils import rmtree
 from datalad.utils import knows_annex
 from datalad.utils import with_pathsep as _with_sep
+from datalad.utils import unique
 
 from .dataset import Dataset
 
@@ -214,7 +215,43 @@ def _install_subds_from_flexible_source(ds, sm_path, sm_url, reckless):
     # do fancy update
     if sm_path in ds.get_subdatasets(absolute=False, recursive=False):
         lgr.debug("Update cloned subdataset {0} in parent".format(subds))
+        # TODO: move all of that into update_submodule ??
+        # TODO: direct mode ramifications?
+        branch = ds.repo.get_active_branch()
+        # track branch originally cloned
+        subrepo = subds.repo
+        branch_hexsha = subrepo.get_hexsha(branch)
         ds.repo.update_submodule(sm_path, init=True)
+        updated_branch = subrepo.get_active_branch()
+        if branch and (not updated_branch or updated_branch == (None, None)):
+            # got into 'detached' mode
+            # trace if current state is a predecessor of the branch_hexsha
+            lgr.debug(
+                "Detected detached HEAD after updating submodule %s which was "
+                "in %s branch before", subds.path, branch)
+            detached_hexsha = subrepo.get_hexsha()
+            if subrepo.get_merge_base(
+                    [branch_hexsha, detached_hexsha]) == detached_hexsha:
+                # TODO: config option?
+                # in all likely event it is of the same branch since
+                # it is an ancestor -- so we could update that original branch
+                # to point to the state desired by the submodule, and update
+                # HEAD to point to that location
+                lgr.info(
+                    "Submodule HEAD got detached. Resetting branch %s to point "
+                    "to %s. Original location was %s",
+                    branch, detached_hexsha[:8], branch_hexsha[:8]
+                )
+                branch_ref = 'refs/heads/%s' % branch
+                subrepo.update_ref(branch_ref, detached_hexsha)
+                assert(subrepo.get_hexsha(branch) == detached_hexsha)
+                subrepo.update_ref('HEAD', branch_ref, symbolic=True)
+                assert(subrepo.get_active_branch() == branch)
+            else:
+                lgr.warning(
+                    "%s has a detached HEAD since original branch %s has another common ancestor with %s",
+                    subrepo.path, branch, detached_hexsha[:8]
+                )
     else:
         # submodule is brand-new and previously unknown
         ds.repo.add_submodule(sm_path, url=clone_url)
@@ -327,7 +364,7 @@ def _get_flexible_source_candidates_for_submodule(ds, sm_path, sm_url=None):
             sm_url,
             remote_url if remote_url else ds.path)
 
-    return clone_urls
+    return unique(clone_urls)
 
 
 def _clone_from_any_source(sources, dest):
--- a/datalad/distribution/utils.py
+++ b/datalad/distribution/utils.py
@@ -205,6 +205,7 @@ def _install_subds_from_flexible_source(ds, sm_path, sm_url, reckless):
 
     # now loop over all candidates and try to clone
     subds = Dataset(opj(ds.path, sm_path))
+    got_installed = not subds.is_installed()
     try:
         clone_url = _clone_from_any_source(clone_urls, subds.path)
     except GitCommandError as e:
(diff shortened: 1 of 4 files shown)


datalad/distribution/utils.py
```diff
@@ -327,7 +364,7 @@ def _get_flexible_source_candidates_for_submodule(ds, sm_path, sm_url=None):
             sm_url,
             remote_url if remote_url else ds.path)
 
-    return clone_urls
+    return unique(clone_urls)
```
yarikoptic: spotted it going through identical URLs -- so fixed with our helper

datalad/distribution/utils.py
```diff
@@ -214,7 +215,43 @@ def _install_subds_from_flexible_source(ds, sm_path, sm_url, reckless):
     # do fancy update
     if sm_path in ds.get_subdatasets(absolute=False, recursive=False):
         lgr.debug("Update cloned subdataset {0} in parent".format(subds))
+        # TODO: move all of that into update_submodule ??
+        # TODO: direct mode ramifications?
```
yarikoptic: started to move into update_submodule but that one is defined in GitRepo, and so far nothing decorated at AnnexRepo level.  @bpoldrack -- should it work currently in direct mode somehow?

yarikoptic:
@bpoldrack @mih -- what do you think? should I implement more safe-guards or should be ok as is for now?

mih:
There is indeed the potential to loose something. Especially, if we start to prune things automatically, as suggested in #1371 

I think we should have protection against that.

yarikoptic:
ha -- looking back at the code -- I think we are safe since this function is to install (clone) new sub-datasets so there is nothing too loose since we didn't have anything before running it.  Nevertheless I have pushed f6f8d699066a8f6fd97236de9f8ce8cb101d6ad6 to
- fix a bug of me taking branch name from the super-dataset not sub-dataset ;)
- adding an additional check and not perform this smart branch update whenever dataset
  was somehow pre-installed first.

This whole logic/adjustment I guess is also relevant to 'update' action, which should be actually the one worrying about loosing stuff ;)  but I did not find other uses of `update_submodule` so I guess it is done differently in the update, and could be the next thing to tackle for consistent with this change behavior

mih:
The reason why you don't find any other use of that function is that `update` is really dumb (I mentioned it repeatedly). While it works recursively, it treats all datasets it acts upon absolutely independently. I will simple wrack the ship in any more than trivial update scenario. It should not cause any real harm, but it would leave behind a dataset hierarchy that has lots of dirty-submodule-superdataset relationships.

`update` needs an update that makes it figure out what changed, and what needs to be saved where, compiles that into a `content_by_ds` dict and gives it to `save_dataset_hierarchy`.

datalad/distribution/utils.py
bpoldrack: I'm too late obviously, but for FTR: Yes, it should. `AnnexRepo ` makes sure, that all the git calls are done with `-c core.bare=False` automatically in direct mode. It needs to override it only if this isn't sufficient, but in most cases it is.

mih:
#1350 is now broken after having merged this PR. The last line of the log snippet below documents the origin of the failure. That line was added in this PR and accesses the `repo` attribute of the to-be-installed dataset -- which is `None` in this case.

I am not sure why this is happening at all. I would appreciate any hints. Thanks.

````
DATALAD_LOG_LEVEL=debug PATH=~/hacking/datalad/git/bin:$PATH PYTHONPATH=~/hacking/datalad/git DATALAD_TESTS_SSH=1 nosetests3 -s datalad/distribution/tests/test_install.py:test_install_skip_failed_recursive
....
[DEBUG  ] Determined class of decorated function: <class 'datalad.distribution.get.Get'> 
[DEBUG  ] Resolved input path arguments: ['/tmp/datalad_temp_test_install_skip_failed_recursiveuj9ipxvl'] 
[INFO   ] Obtaining <Dataset path=/tmp/datalad_temp_test_install_skip_failed_recursiveuj9ipxvl>  recursively 
[DEBUG  ] Failed to detect a valid repo at /tmp/datalad_temp_test_install_skip_failed_recursiveuj9ipxvl/subm 1 
[DEBUG  ] Failed to detect a valid repo at /tmp/datalad_temp_test_install_skip_failed_recursiveuj9ipxvl/subm 1 
[DEBUG  ] Retrieving a dataset from URL: /tmp/datalad_temp_testrepo_bd4ykdz9/subm 1 
[DEBUG  ] Failed to retrieve from URL: /tmp/datalad_temp_testrepo_bd4ykdz9/subm 1 
[DEBUG  ] Retrieving a dataset from URL: file:///tmp/datalad_temp_testrepo_from62uw 
[DEBUG  ] Failed to retrieve from URL: file:///tmp/datalad_temp_testrepo_from62uw 
[DEBUG  ] Failed to detect a valid repo at /tmp/datalad_temp_test_install_skip_failed_recursiveuj9ipxvl/subm 1 
[WARNING] Target /tmp/datalad_temp_test_install_skip_failed_recursiveuj9ipxvl/subm 1 already exists and is not an installed dataset. Skipped. 
[DEBUG  ] Original failure:
| Cmd('git') failed due to: exit code(128)
|   cmdline: git clone -v file:///tmp/datalad_temp_testrepo_from62uw /tmp/datalad_temp_test_install_skip_failed_recursiveuj9ipxvl/subm 1
|   stderr: 'fatal: destination path '/tmp/datalad_temp_test_install_skip_failed_recursiveuj9ipxvl/subm 1' already exists and is not an empty directory.' [gitrepo.py:clone:576] 
[DEBUG  ] Update cloned subdataset <Dataset path=/tmp/datalad_temp_test_install_skip_failed_recursiveuj9ipxvl/subm 1> in parent 
[DEBUG  ] Failed to detect a valid repo at /tmp/datalad_temp_test_install_skip_failed_recursiveuj9ipxvl/subm 1 
[DEBUG  ] Failed to detect a valid repo at /tmp/datalad_temp_test_install_skip_failed_recursiveuj9ipxvl/subm 1 
[ERROR  ] Installation of subdatasets <Dataset path=/tmp/datalad_temp_test_install_skip_failed_recursiveuj9ipxvl/subm 1> failed with exception: 'NoneType' object has no attribute 'get_active_branch' [utils.py:_install_subds_from_flexible_source:146] 
````

yarikoptic:
sorry -- can't find that code within 0.4.1-688-g55689abc of your branch -- 146th line of that file is just "return subds" for me

mih:
Yes, i reverted it.

PR #1372, opened 2017-03-15
ENH: tqdm - use leave=True instead of explicit clean call to avoid spurious new lines

yarikoptic:
There is still seems to be some spurious new lines, but otherwise output
seems to behave better

Example:

before:

```
$> datalad install -g -J3 sub001/anatomy
[INFO   ] Getting 1 items of dataset <Dataset path=/tmp/raiders> ...
[INFO   ] Actually getting 10 files

<Dataset path=/tmp/raiders>
```
after:
```
$> datalad install -g -J3 sub001/anatomy
[INFO   ] Getting 1 items of dataset <Dataset path=/tmp/raiders> ...
[INFO   ] Actually getting 10 files
<Dataset path=/tmp/raiders>
```

TODO
- [x] just disable "interactive" progress bars by default while running tests, make them "optional" so we could still smoke test them.  This is to avoid destroying output such as names of the tests being printed out... and there is still some spurious newline in some cases which extends output

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,6 +49,7 @@ matrix:
     - DATALAD_TESTS_PROTOCOLREMOTE=1
     - DATALAD_TESTS_DATALADREMOTE=1
     - DATALAD_LOG_OUTPUTS=1
+    - DATALAD_TESTS_UI_BACKEND=console
   - python: 2.7
     env:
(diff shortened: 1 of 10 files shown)


mih:
Great. Much appreciated!

Issue #1374, opened 2017-03-16
our example was broken and we didn't spot it + some awkward intollerance to spaces by ssh connector


yarikoptic:
on travis we have
```
$ if [[ ${TMPDIR:-} =~ ".* .*" ]]; then tools/testing/run_doc_examples; else echo "not running since has spaces"; fi

not running since has spaces
```
even if we didn't define TMPDIR at all -- unlikely travis itself did ... not sure WTF and so far failed to figure out why magic fails when running under TMPDIR with spaces:
```
[INFO   ] Getting 1 items of dataset <Dataset path=/tmp/te st/datalad_demo_bob.GCDi/myanalysis> ... 
Got nothing new
+ datalad create-sibling --recursive -s public 'localhost:/tmp/te st/datalad_demo_testpub.6Qm5'
[INFO   ] Connecting ... 
command-line line 0: garbage at end of line; "st/datalad_demo_bob.GCDi/.cache/datalad/sockets/ef2d633e66d966938fb3423864aa2d8e".
[ERROR  ] Cannot open SSH connection to localhost [sshconnector.py:__call__:119] (RuntimeError) 
http_proxy=http://localhost:3182 TMPDIR=/tmp/te\ st bash -x   9.88s user 1.96s system 79% cpu 14.926 total
```

failed to replicate in an independent call -- then things work out just fine

PR #1375, opened 2017-03-16
RF+BF: save -- revive all_changes to die properly, and -a -> -u, all_changes -> all_updated

yarikoptic:
BF: demo script which was apparently broken for a while   see #1374 

- fail if `save -a` is used since its behavior has changed to fulfill the promise of the least surprise
  - get proper deprecation error with description instead of just behaving differently now

```
$> datalad save -a                                                                                                                
[ERROR  ] RF: all_changes option passed to the save is deprecated since version 0.5.0. Use all_updated  option where fits and/or datalad add instead. [save.py:__call__:142] (DeprecatedError) 
```
- I know that `save -a` is somewhat consistent with `git commit -a` but
  -  we might want to revive `-a` operation at some point, then it would be all kosher without confusion and tears
  - it is `git add -u` to add updated files into the index so we will be consistent with it instead ;)

### DIFF
--- a/datalad/distribution/install.py
+++ b/datalad/distribution/install.py
@@ -574,7 +574,7 @@ def _save_installed_datasets(ds, installed_datasets):
         ds.save(
             files=paths + ['.gitmodules'],
             message='[DATALAD] ' + msg,
-            all_changes=False,
+            all_updated=False,
             recursive=False)
     except ValueError as e:
         if "did not match any file(s) known to git" in str(e):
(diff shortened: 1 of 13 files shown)


mih:
I will not merge this, as it will conflict with #1350. Resolution would bring #1350 over the magic 100 commit mark.

yarikoptic:
Don't worry about 100 commits like even if you don't want to squash any of yours, and I could help with a merge. I would really like to get current state released and avoid surprises

PR #1377, opened 2017-03-16
BF: test example running and some work toward handling of socket paths with spaces in them

yarikoptic:
Let's see if travis is happy.

handling of to socket files paths with spaces in them is still TODO:  http://git-annex.branchable.com/bugs/cannot___40__or_how__63____41___to_pass_socket_path_with_a_space_in_its_path_via_annex-ssh-options/


yarikoptic:
1. woohoo from https://travis-ci.org/datalad/datalad/jobs/211898447#L2420
```
ControlPath "/var/tmp/sym link/datalad_demo_bob.mbhS/.cache/datalad/sockets/61e3791cde6802cbec22210eb6b5c00d.Y5QqaEOcXyUqozX2" too long for Unix domain socket
```
which is quite a boomer!  From resourceful https://github.com/ansible/ansible/issues/11536#issuecomment-153030743  on Mac it is 104 characters including trailing `\0`.  Since some trailer also added, I will just reduce has size to ... 8!  I really doubt we would get any collision even in that tiny multibillion choices space ;)

2. While looking at the code -- hash is nohow uses "current repo" or "current process" ID, so I see possible collision of reusing/removing the socket initiated by another datalad process going to the same host... not kosher imho

3. and I will need to disable it to be ran in direct mode for now: https://travis-ci.org/datalad/datalad/jobs/211898455
```
[ERROR] Failed to run ['git', '-c', 'receive.autogc=0', '-c', 'gc.auto=0', 'submodule', 'update', '--checkout', '--init', '--', u'src/forrest_structural'] under '/tmp/datalad_demo_alice.WnTJ/bobs_analysis'. Exit code=1. out= err=fatal: $program_name cannot be used without a working tree.
```
although that one fails also a test:
```shell
  File "/home/travis/build/datalad/datalad/datalad/support/tests/test_annexrepo.py", line 1044, in test_annex_ssh
    ok_(exists(socket_1))

AssertionError: None
```
which I guess now I have to fix ... heh... although a weird one since on travis it passes in indirect mode but now fails for me even in it on my laptop... weird

bpoldrack:
Let's not return to disabling any test in direct mode that used to pass, please!
We need to figure it out.

Furthermore at a first glance this reminds me of a failure, I fixed in PR #1277 . May be yours simply has to wait for that one, which is going to be ready next week.

yarikoptic:
I am not certain that it was ever working on direct mode, but sure better to keep it if we could make it work

yarikoptic:
@bpoldrack it seems that the fix caused the test to break? https://travis-ci.org/datalad/datalad/jobs/212724582#L1758

bpoldrack:
Hm. A relation to my fix doesn't make sense to me.
I can't reproduce it either. Furthermore at a first glance the "failure" looks reasonable, since it refuses to drop something, that seems to be available from github only, which is marked as "unsafe".
Looking into it ...

bpoldrack:
Also: It did not happen on Travis, when I pushed my fix to my fork.

bpoldrack:
@yarikoptic could you push again, please? I'm suspecting it's just a connection issue with github, which happens from time to time. In that case, the failing of `git annex drop` is kind of correct and we need to adapt that test to account for it.

yarikoptic:
ok -- I have clicked "Restart build" on travis view.  Don't you have that button available too @bpoldrack ?

bpoldrack:
Never noticed, but yes: I have it! ;-)
Sorry.

yarikoptic:
seems to be ok -- only nfs one timed out as usual

bpoldrack:
Okay. Made an issue for that, since I think it's a more general issue.
Thus merging.

PR #1378, opened 2017-03-17
ENH+RF: _LOG_NAME, _LOG_NAMES, _LOG_NAMESRE to control what log entries to print

yarikoptic:
thought to do it for a while but never got to it... now need to troubleshoot that ssh connector business and do not want to scroll or grep through tons of logs to find relevant lines.. so it will be something like

```DATALAD_LOG_LEVEL=2 DATALAD_LOG_NAMESRE=.*ssh do_whatever ```

and only log lines from modules having anything to do with ssh should get printed ;)

### DIFF
--- a/datalad/interface/common_cfg.py
+++ b/datalad/interface/common_cfg.py
@@ -147,6 +147,18 @@ definitions = {
             'title': 'Used for control the verbosity of logs printed to '
                      'stdout while running datalad commands/debugging'}),
     },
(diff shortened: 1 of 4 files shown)


mih:
Nice!

bpoldrack:
FTR:

Take this, @yarikoptic:  :1st_place_medal:

yarikoptic:
Glad you liked it

Issue #1384, opened 2017-03-17
Do not use setuptools entry_points mechanism


yarikoptic:
It seems to add around 200ms to startup time, i.e. makes our scripts start time twice or three times slower. And in principle we don't quite need those entry points discovery etc - may be only on windows.  So on *nix'es we could replace them with  simple python scripts which would simply invoke those functions specified as entry points.

yarikoptic:
bloody setuptools!  I found no reliable way to "override" its behavior regarding installation of the scripts.  Mere import of the setuptools is sufficient to "augment" distutils behavior and lead to creation of those "script runners" even on linux:

1. simple setuputils scripts spec - already creates its shim
```shell
$> deactivate; rm -rf /tmp/testproj; mkdir /tmp/testproj; cd /tmp/testproj; echo -e 'from setuptools import setup;\nsetup(scripts=["script1"])' > setup.py; echo -e '#!/usr/bin/env python\nprint("#RUN")' > script1; virtualenv venv; source venv/bin/activate; python setup.py install >/dev/null; which script1 | xargs cat ; 
Running virtualenv with interpreter /usr/bin/python2
New python executable in /tmp/testproj/venv/bin/python2
Also creating executable in /tmp/testproj/venv/bin/python
Installing setuptools, pkg_resources, pip, wheel...done.
warning: install_lib: 'build/lib.linux-x86_64-2.7' does not exist -- no Python modules to install

zip_safe flag not set; analyzing archive contents...
#!/tmp/testproj/venv/bin/python
# EASY-INSTALL-SCRIPT: 'UNKNOWN==0.0.0','script1'
__requires__ = 'UNKNOWN==0.0.0'
__import__('pkg_resources').run_script('UNKNOWN==0.0.0', 'script1')
```
2. with distuils it is all nice and dandy:
```shell
$> deactivate; rm -rf /tmp/testproj; mkdir /tmp/testproj; cd /tmp/testproj; echo -e 'from distutils.core import setup;\nsetup(scripts=["script1"])' > setup.py; echo -e '#!/usr/bin/env python\nprint("#RUN")' > script1; virtualenv venv; source venv/bin/activate; python setup.py install >/dev/null; which script1 | xargs cat ;          
Running virtualenv with interpreter /usr/bin/python2
New python executable in /tmp/testproj/venv/bin/python2
Also creating executable in /tmp/testproj/venv/bin/python
Installing setuptools, pkg_resources, pip, wheel...done.
[...]

PR #1385, opened 2017-03-17
OPT: An ugly workaround to avoid generation of setuptools shims on non-Windows systems

yarikoptic:
Closes: #1384


yarikoptic:
uff... done. not sure if it had notable impact on duration of our tests but it for sure makes running datalad --help notably faster ;)

yarikoptic:
if no objections are heard -- I will merge it in later today

PR #1386, opened 2017-03-18
OPT: shave off ~10sec from test_make_studyforrest_mockup runtime by running 2 loops instead of 3

yarikoptic:
actually I thought it would go down from 70 to newly measured 40 sec but apparently it is just around 10sec... still notable improvement especially since it is just a smoke test


yarikoptic:
well, if objections would come -- we could revert

mih:
No posthoc objections from me ;-)

PR #1395, opened 2017-03-20
OPT: use eatmydata for some installations to hopefully shave off a few seconds

yarikoptic:


yarikoptic:
failures are unrelated and fixed in other PRs. seems to shave off few seconds -- nfs based test passed even ;)

Issue #1396, opened 2017-03-20
Speeding up the tests


mih:
The test run time is getting too long. @yarikoptic has started slimming the test battery.

Another angle to approach this problem would be to gradually move to a more layered testing approach, where we don't always run the rather heavy top-level commands with their substantial condition checking penalties, but use intermediate commands (from future RF) that execute functionality under a more constrained set of requirements.

yarikoptic:
Duplicate of https://github.com/datalad/datalad/issues/1383
?

mih:
Indeed.

PR #1397, opened 2017-03-20
NF: `clone` a faster `install` for most cases

mih:
Check `datalad/distribution/tests/test_clone.py` for converted install tests that represent the scope of `clone`. Major change is dramatically reduced flexibility wrt input arguments (see #1396). This leads to about ~10% of speed up for our bread-and-butter call, i.e. install one dataset (recursive or not) from one source.

It looks like we could switch to `clone` for pretty much all use in the tests, except for the test of `install`. Even if `install` represents only a fraction of the actual runtime, we are calling it a lot, so it should make things a bit faster.

Here is a demo:

````
In [1]: %run clone_speed_demo.py

# disregard the first run -- testrepos are built here (see below for the code)
In [2]: %timeit test_clone_speed()
1 loop, best of 3: 3.63 s per loop

In [3]: %timeit test_install_speed()
1 loop, best of 3: 4.02 s per loop

In [4]: %timeit test_install_speed()
1 loop, best of 3: 4.04 s per loop

In [5]: %timeit test_clone_speed()
1 loop, best of 3: 3.63 s per loop

In [6]: %timeit test_clone_speed()
1 loop, best of 3: 3.68 s per loop

In [8]: %timeit test_clone_speed()
1 loop, best of 3: 3.64 s per loop
````

Here is the code. It performs a recursive install to get a lower bound estimate. Both commands use `get` for the recursion part, albeit in a slightly different way. Not going recursive yields a ~12% speed increase.

````
from datalad.tests.utils import with_testrepos
from datalad.tests.utils import with_tempfile
from datalad.api import clone
from datalad.api import install
from datalad.api import Dataset


@with_testrepos('.*nested_submodule.*', flavors=['local-url', 'network', 'local'])
@with_tempfile(mkdir=True)
def test_clone_speed(src, path):
    Dataset(path)

    clone(src, path, recursive=True,
          result_xfm='datasets', return_type='item-or-list')


@with_testrepos('.*nested_submodule.*', flavors=['local-url', 'network', 'local'])
@with_tempfile(mkdir=True)
def test_install_speed(src, path):
    Dataset(path)

    install(path, source=src, recursive=True)
````

### DIFF
--- a/datalad/api.py
+++ b/datalad/api.py
@@ -89,7 +89,7 @@ def _generate_func_api():
             # TODO: BEGIN to be removed, when @build_doc is applied everywhere
             spec = getattr(intf, '_params_', dict())
             api_name = get_api_name(intfspec)
-            if api_name not in ('update', 'save', 'create', 'unlock', 'clean', 'drop', 'uninstall', 'get'):
+            if api_name not in ('update', 'save', 'create', 'unlock', 'clean', 'drop', 'uninstall', 'get', 'clone'):
                 # FIXME no longer using an interface class instance
                 # convert the parameter SPEC into a docstring for the function
                 update_docstring_with_parameters(
--- a/datalad/distribution/clone.py
+++ b/datalad/distribution/clone.py
@@ -238,8 +238,6 @@ class Clone(Interface):
             # TODO generator: yield when `add` is converted
             dataset.add(dest_path, save=True, ds2super=True)
 
-        # TODO make sure `get` does this too
-        # handle description arg
         _handle_possible_annex_dataset(
             destination_dataset,
             reckless,
--- a/datalad/distribution/clone.py
+++ b/datalad/distribution/clone.py
@@ -6,9 +6,7 @@
 #   copyright and license terms.
 #
 # ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
-"""High-level interface for dataset (component) installation
-
-"""
+"""Plumbing command for dataset installation"""
 
 
 import logging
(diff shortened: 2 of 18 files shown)

### CODE REVIEW

**Comment by yarikoptic** [comment]
Not sure if we sold be solving performance issues by expanding API... I am already confused by having both install and clone

**Comment by bpoldrack** [comment]
Since you are familiar with git, `clone` should not exactly be a confusing thing. So, if you want to clone, use `clone. If you want a lot of guessing magic and simultaneously add things to superdatasets and so on, use `install`.

**Comment by mih** [comment]
One more datapoint: For `nosetests3 -s datalad/tests/test_testrepos.py:test_make_studyforrest_mockup` alone I get an ~8% speedup (best of three iterations) by doing `s/install/clone/g`. In absolute numbers this is ~10s.

**Comment by yarikoptic** [comment]
```
Since you are familiar with git, clone should not exactly be a confusing thing. So, if you want to clone, use clone. If you want a lot of guessing magic and simultaneously add things to superdatasets and so on, use install.
```
that is my point -- how would I know if I need any magic or not when I am cloning/installing a remote source?  also moving testing of install in various scenarios as it is done now into testing through a new command could potentially leave install less tested and thus magic break more often.

mih:
Please don't sweat over this PR. If needed I can move it to my clone.

yarikoptic:
I feel deficient of having no "my clone" to move stuff to ;-)  seriously -- what would be the "function in line" of the new API command?  is it for a pro-user (or tests) to avoid using `install` when he could just use `clone` because it is faster?  or... ?

mih:
@bpoldrack test failure for a direct mode test is the last issue. could you please have a quick look. there still seems to be a necessity that i am not aware off. thanks.

PR #1387, opened 2017-03-19
ENH/BF: crawler - support few more diff modes, stage only non empty filenames

yarikoptic:
no split 117 ds in openfmri

collection of changes aggregated from updating openfmri dataset

submitted for testing -- if broke nothing, will merge for imminent release

Closes #1394

PR #1398, opened 2017-03-20
BF(workaround): run only subset of tests under NFS while testing PR - full in master

yarikoptic:
we need to make testing of PRs usable again quickly, but should still strive to lighten up testing so master could pass tests again

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -93,9 +93,15 @@ matrix:
     - TMPDIR="/tmp/d i r"
     - _DATALAD_NONPR_ONLY=1
   - python: 2.7
-    # Test under NFS mount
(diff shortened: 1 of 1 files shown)


yarikoptic:
failed test was fixed in just merged #1387, so should be good to go unless we decide against it. @mih @bpoldrack ?

bpoldrack:
I'm fine with that one. I guess if there are fundamental issues with NFS, this should be sufficient to point us to them. But then: I have no experience whatsoever with NFS - may be I'm not the one to judge ...

yarikoptic:
ok -- since no objections, merging since timing out test keeps annoying ;)

PR #1408, opened 2017-03-21
NF: Result summary renderer (more or less useful demo in `get`)

mih:


mih:
Only failing test is the direct mode issue. Merging.

PR #1409, opened 2017-03-21
RF: return values (#1350 continued)

### LINKED ISSUE #819: New general option `on_failure` to indicate action when something goes wrong

### LINKED ISSUE #1110: Confusing communication and inconsequent, and undesired behavior of `install`
````
$ datalad install -d . --reckless -s ../aligned src/aligned
Total: 100%|███████████████████████████████████████████████████████▉| 36.0M/36.0M [00:00<00:00, 62.9MB/s]
[INFO   ] Installing subdataset from '../aligned' at: ../aligned 

[INFO   ] Saving possible changes to <Dataset path=/home/kbenz/taste2_data/collection/testballoon> - installed subdatasets: src/aligned 
<Dataset path=/home/kbenz/taste2_data/collection/testballoon/src/aligned>
````

The first info line is wrong (seems to state the source twice), and scared the shit out of my -- this was on real data ....

After the command (which basically did everything right) I am left with this:

````
$ git diff HEAD..HEAD~1
diff --git a/src/aligned b/src/aligned
deleted file mode 160000
index 9bde30d..0000000
--- a/src/aligned
+++ /dev/null
@@ -1 +0,0 @@
-Subproject commit 9bde30de94b4d787f44ed8be6708320709a0032d
$ git status
On branch master
Changes to be committed:

	modified:   .gitmodules
````

So the subdataset is commited, the module config is not.

Moreover:

````
$ git diff HEAD
diff --git a/.gitmodules b/.gitmodules
index 982a8ad..7069686 100644
--- a/.gitmodules
+++ b/.gitmodules
@@ -9,4 +9,4 @@
        url = ./src/aligned
 [submodule "src/aligned"]
        path = src/aligned
-       url = ./src/aligned
+       url = /some/abs/path/collection/aligned
````

There are two things to this:

1. I gave a relative path to install from, but an absolute path was recorded. I don't think that is a good idea.
2. you can see that a submodule was registered at that location before -- which I had accidentally `rm`'ed before running this command.

Auto-saving had commit that change before installation, but only like so:

````
$ git diff HEAD~2..HEAD~1 -- src/
diff --git a/src/aligned b/src/aligned
deleted file mode 120000
index c800611..0000000
--- a/src/aligned
+++ /dev/null
@@ -1 +0,0 @@
-../../aligned/
\ No newline at end of file
````

I am not sure what should actually happen in this. It seems somewhat correct behavior, as in I could simply reobtain the deleted subdataset, but I wanted to bring it up nevertheless,

### LINKED ISSUE #1173: RF install to use `add` for registering install submodules
Currently this is rather convoluted. Once #1172 is resolved, `install` should stop using its various little helpers and plainly `add` a submodule to its parent, once the core installation operation is completed.

Not sure exactly where this needs to plug into the current implementation, but `datalad/distribution/utils.py:_save_installed_datasets()` might be it.

### LINKED ISSUE #1469: `get` fails on unavailable subdataset with incomprehensible message
````
% datalad -l debug get raw/structural
[DEBUG  ] Resolved input path arguments: ['/tmp/tnt/raw/structural'] 
[DEBUG  ] Failed to detect a valid repo at /tmp/tnt/code/ofhelpers 
[DEBUG  ] Failed to detect a valid repo at /tmp/tnt/raw/phase1 
[DEBUG  ] Failed to detect a valid repo at /tmp/tnt/raw/phase2 
[DEBUG  ] Failed to detect a valid repo at /tmp/tnt/raw/structural 
[DEBUG  ] Failed to detect a valid repo at /tmp/tnt/raw/structural 
[DEBUG  ] Failed to detect a valid repo at /tmp/tnt/raw/structural 
[INFO   ] Installing subdataset <Dataset path=/tmp/tnt/raw/structural> 
[DEBUG  ] Failed to detect a valid repo at /tmp/tnt/raw/structural 
[DEBUG  ] Retrieving a dataset from URL: /home/mih/forrest/collection/tnt/raw/structural 
[DEBUG  ] Failed to retrieve from URL: /home/mih/forrest/collection/tnt/raw/structural 
[DEBUG  ] Retrieving a dataset from URL: /home/mih/forrest/collection/structural 
[DEBUG  ] Failed to retrieve from URL: /home/mih/forrest/collection/structural 
[DEBUG  ] Failed to detect a valid repo at /tmp/tnt/raw/structural 
[WARNING] Target /tmp/tnt/raw/structural already exists and is not an installed dataset. Skipped. 
[DEBUG  ] Original failure:
| Cmd('/usr/lib/git-annex.linux/git') failed due to: exit code(128)
|   cmdline: /usr/lib/git-annex.linux/git clone -v /home/mih/forrest/collection/structural /tmp/tnt/raw/structural
|   stderr: 'fatal: repository '/home/mih/forrest/collection/structural' does not exist
| ' [cmd.py:wait:291] 
[DEBUG  ] Update cloned subdataset <Dataset path=/tmp/tnt/raw/structural> in parent 
[DEBUG  ] Failed to detect a valid repo at /tmp/tnt/raw/structural 
[DEBUG  ] Failed to detect a valid repo at /tmp/tnt/raw/structural 
[ERROR  ] 'NoneType' object has no attribute 'get_active_branch' [utils.py:_install_subds_from_flexible_source:223] (AttributeError) 
````

The actual issue is that the subdataset has a relative path "URL" and is not available at the location of `origin`.

Solution will come as part of #1468

### PR DESCRIPTION
**Pull request by mih**
Basic idea: commands `yield` report dictionaries. And wrappers act upon this information in whatever desired way.

Sits on top of #1348. Present tail of the procedure is in #1409 

Change summary:

- [x] global command line switch to select output format
- [x] line-based 'json' output format (like git-annex)
- [x] 'simple' output format for "STATUS: PATH" lines, where PATH is relative whenever a reference dataset is available
- [x] supported commands return generators that are actually executed in main.py
- [x] centralized error detection that guarantees an exception if any error occured
- [x] BF: `update` did not fail on unavailable paths as argument
- [x] customize exception generation to be less dumb
- [x] centralize logging by shipping messages in the status dicts (actually: message template, variable, logger instance or func)
- [x] reimplement forced 'cmdline' rendering of API_ function with the new facilities to get rid of these shadow functions
- [x] convert actual return value nature on a by-call basis. Standard should be json dict sequence/generator.
- [x] provide `return_value` helper module with:
      - function to give a default dict
      - function to convert git-annex json record into our own structure
- [x] implement new constraint to test whether a dict has a key with a certain value 
- [x] implement general command line options to filter results by `type` and by `status` (using new constraint above)
- [x] implement new general kwarg for all commands to accept a filter function for results (to fit a passed (compound) constraint) (name: `filter_results`)
- [x] implement new general kwarg for all commands to switch return values from generator to sequence (spec: `return_type`, `{'generator', 'list', 'item-or-list'}`)
- [x] implement flexible transformation of return values from result dicts to other types (like Datasets)
- [x] implement general `--on-failure` switch (fixes #819)
- [x] new test helpers: `assert_status`, `assert_message`, `assert_in_results`, `assert_not_in_results`, `assert_result_count`, `assert_fields_equal`
- [x] #1380 support for result summaries
 
Commands converted:

- [x] create (possibly TODO: report addition to a superdataset)
- [x] install -- RF'ed to remove redundant code; split core functionality into a `clone` command (fully converted); high-level condition checking in `install` itself still needs conversion
- [x] get
- [x] add (complicated due to the old-school recursion in helper functions, needs substantial RF)
- publish (seems straightforward, but has quite some diff to master, not attempted)
- [x] uninstall (needs additional RF, PY2 version of Ben's class detector cannot deal with more than one class per module)
- [x] drop (needs additional RF, PY2 version of Ben's class detector cannot deal with more than one class per module)
- [x] remove (needs additional RF, PY2 version of Ben's class detector cannot deal with more than one class per module)
- [x] update (cmd itself is done, but needs to pass though results from `get`)
- [ ] create-sibling (seems doable, unclear if RF needed to pass `add-sibling` results through)
- [ ] create-sibling-github (doable)
- add-sibling (better waits till after general RF; convoluted already, calling for `sibling` command)
- [ ] rewrite-urls (does not seem to be used anywhere -> low priority; but otherwise simple)
- [x] unlock
- [x] save
- export (seems doable, but also rather useless)
- search (will not be attempted, waiting for metadata RF)
- aggregate-metadata (will not be attempted, waiting for metadata RF)
- test (not needed)
- crawl, crawl-init (no idea if it is feasible to attempt this, better be done by Yarik, also crawl-init about to be gone #1155)
- ls (will not be attempted, does not fit paradigm as of now, but could become relevant after RF)
- [x] clean (looks straightforward)
- [ ] add-archive-content
- [ ] download-url (seems straightforward, custom stop_on_failure needs to be RF'ed away)
- [ ] create-test-dataset
- sshrun (out of scope, needs to stay as it is)

Copied from #1468:

- [x] Various fixes by @yarikoptic 
- [x] `install` is a new-style generator command now
- [x] fixes #1469 
- [x] fixes #1173
- [x] fixes #1110
- [x] some work towards #875 
- [x] `install` no longer uses any low-level helpers directly, instead it sits rather cleanly on top of `clone` and `get`, and performs argument inspection and switching logic only. It seems that this is now at a stage where we could think about adding more convenience functionality to `install` without much pain (e.g. #1347, #1146)
- [x] code cleanup: unreachable parts in `install`, plus a duplicate (and buggy) implementation of a `clone` helper removed
- [x] Remove support for old-style "if_dirty" handling from `install` #1175
- [x] helpers unique to `get` now consolidated in `get.py`
- [x] Extend result filtering API to be able to pass a command's `kwargs`
- [x] Attempt to implement result filters for input argument based filtering in `install`
- [x] Figure out what kind of result rendering we want #1476
- [ ] Possibly disable "implicit" result report for any command by default (at least via the cmdline API) -- this is more a thought than a plan for now and likely out of scope for this PR
- [x] must have fix for #1474
- [x] would be nice to get #1475 sorted out
- [x] make `get` report on things that annex is silent about (whole dirs, files in Git, ...)

### CODE REVIEW

**Comment by mih** [comment]
Except for the direct mode fix (promised by @bpoldrack ) and the fallout from merging the current master just now -- this is done from my POV. Switching to meta data now.

**Comment by mih** [comment]
@yarikoptic Here is the test that fails after merging master. This is in the code you recently added for creating adhoc branches. This could be the symptom of a more general problem, but given that you recently looked into this bit maybe you have an immediate insight.

````
FAIL: datalad.distribution.tests.test_install.test_install_skip_failed_recursive
    self.test(*self.arg)
  File "/home/travis/build/datalad/datalad/datalad/tests/utils.py", line 734, in newfunc
    t(*(arg + (uri,)), **kw)
  File "/home/travis/build/datalad/datalad/datalad/tests/utils.py", line 564, in newfunc
  File "/home/travis/build/datalad/datalad/datalad/distribution/tests/test_install.py", line 694, in test_install_skip_failed_recursive
    result[0]['message'][2])
AssertionError: "destination path '/tmp/datalad_temp_test_install_skip_failed_recursivelw23T3/subm 1' already exists and is not an empty directory" not found in "'NoneType' object has no attribute 'get_active_branch' [utils.py:_install_subds_from_flexible_source:147]"
````

**Comment by mih** [comment]
FTR: Inflated test runtime is a consequence of running most former `install` tests also for `clone`. I will take a shot at removing duplicates.

**Comment by mih** [comment]
@yarikoptic I would be great if you could make an attempt to convert the `install` implementation in this PR to a generator. It is already reduced to the aspects that aren't covered by `clone` (i.e. mainly scenario determination and invocation of corresponding lower-level commands). All underlying commands are already generators, so their results can be inspected (if necessary) or simply passed outside. There are now interfaces to render individual results and summaries. However, returning status dicts by default will unavoidably change the way `install` feels. It may well be that `install` would become a special case in that it hard-codes a result filter to only ever report installed datasets, and combines that with a result transformation that yields actual dataset instances. This would be fine, as there is no need to use install internally when full result reports are needed, as a `clone/get` combo call could provide those. Please keep in mind that in this case (hard-coded filter, hard-coded transformation) there is actually little need to do any RF, because that is the status quo, minus the fact that it is not a generator. But if results come out in a custom format, we cannot act on them in generic wrapper code anyways, so little would be lost. Please have a look.

**Comment by yarikoptic** [comment]
@mih 
- re test_install_skip_failed_recursive  -- see https://github.com/mih/datalad/pull/18  ;-)
- re the rest of the install -- would not have any quality time for that any time soon -- and it would take me too long to get acquainted.  As I have commented in that PR, I am not quite sure on the assumptions those helper functions have (when to fail)

**Comment by mih** [comment]
Merged master again -- conflicts getting more and more difficult. Apart from the direct mode failure (that is already fixed elsewhere according to @bpoldrack ) this thing is done. It would be great to get this merged. I am switching my focus to metadata management and will base this off of this branch.

**Comment by yarikoptic** [comment]
yeah -- I will git it a run tomorrow or day after and then let's merge indeed BUT @bpoldrack -- is there a chance to cherry-pick the fix for the direct mode into this branch since we don't want to merge something which would be breaking the other PRs after

**Comment by yarikoptic** [comment]
@bpoldrack could you please look at the failing direct mode related test/example?

**Comment by yarikoptic** [comment]
@mih trying to familiarize myself (finally) with this PR changes.  Great to see some things working much better now (e.g. "the distribution guarantee for install" if I call it second time with get_data -- it does get data etc -- just great) 
even though it is not there yet -- how do you envision output of `install` command to look like?  will it just output the paths (in `path` field) or datasets?

**Comment by mih** [comment]
Glad you like it!

> how do you envision output of install command to look like? will it just output the paths (in path field) or datasets?

We have to distinguish cmdline API and Python API. The former is set up to execute a class-specific custom renderer by default (could be switched to another with a global option). I am pretty much indifferent what it should output by default. Ideally, it would be a behavior that would also make sense for other commands. At present our `simple` renderer comes closest (status: path [relative to base dataset if one was given]). Just returning paths makes sense. It could be combined with `--report-status` and friends to get only the paths one is interested in.

For the Python API the situation is different. If we ever want to use `install` inside another command, it should be made to return result dicts. We could decide to add a config that specifies a default result filter/transformer. Check the code for the `create` command to see how this is done (top of the class). Given that we have these levers, I don't think we should make `install` an outlier by forcing it into one particular type of return values in all cases.

File: datalad/distribution/clone.py [hunk 1]
```diff
@@ -0,0 +1,249 @@
...
+        # generate candidate URLs from source argument to overcome a few corner cases
+        # and hopefully be more robust than git clone
+        candidate_sources = _get_flexible_source_candidates(source)
+        for source_ in candidate_sources:
+            try:
+                lgr.info("Attempting to clone dataset from '%s' to '%s'",
+                         source_, dest_path)
+                # TODO this is a shame. we should be able to instruct GitRepo
+                # how to behave wrt logging, and not stick a pillow in its face
+                with swallow_logs():
```
mih: FTR: The swallowing was in-place before (it just appears as an addition due to other RF). I merely added the comment.

datalad/distribution/clone.py
```diff
@@ -0,0 +1,249 @@
...
+                    GitRepo.clone(
+                        path=dest_path, url=source_, create=True)
+                break  # do not bother with other sources if succeeded
+            except GitCommandError as e:
+                lgr.debug("Failed to clone from URL: %s (%s)",
+                          source_, exc_str(e))
+                lgr.debug("Wiping out unsuccessful clone attempt at: %s",
+                          dest_path)
```
mih: Thanks!

datalad/distribution/create.py
```diff
@@ -264,29 +280,31 @@ def __call__(
...
-    def result_renderer_cmdline(res, args):
+    def custom_result_renderer(res, **kwargs):
         from datalad.ui import ui
-        if res is None:
+        if res.get('action', None) == 'create' and \
+               res.get('status', None) == 'ok' and \
+               res.get('type', None) == 'dataset':
+            ui.message("Created dataset at {}.".format(res['path']))
+        else:
             ui.message("Nothing was created")
```
mih: Check `datalad --help` all these options (result filtering, ...) are global and uniform for any command (assuming new-style).

mih:
> - yet to grasp the whole thing, but not sure if ever would

`eval_result` is pretty much the only place you need to see in order to grasp what it does in essence

> I would have preferred if at least 'install' was also completed to follow the scheme since it is probably the most used functionality ATM

If you specify what the desired behavior is in terms of result properties (which status to report, which property to report, in which format) I can give it a try. ATM we only have a long list scenarios with custom defined behavior for each. I could not clearly and immediately see how to map those onto the new facilities, without ending up with something that is as custom as the present implementation.

datalad/distribution/create.py
mih: Snippet from `--help`:

````
  --output-format {simple,json,tailored}
                        select format for returned command results. 'simple'
                        give one line per result using a '<status>: <path>'
                        syntax; 'json' renders a JSON object with all
                        properties for each result (one per line); 'tailored'
                        enables a command-specific rendering style that is
                        typically tailored to human consumption (no result
                        output otherwise).
  --report-status {success,failure,ok,notneeded,impossible,error}
                        constrain command result report to records matching
                        the given status. 'success' is a synonym for 'ok' OR
                        'notneeded', 'failure' stands for 'impossible' OR
                        'error'.
  --report-type {dataset,file}
                        constrain command result report to records matching
                        the given type. Can be given more than once to match
                        multiple types.
  --on-failure {ignore,continue,stop}
                        when an operation fails: 'ignore' and continue with
                        remaining operations, the error is logged but does not
                        lead to a non-zero exit code of the command;
                        'continue' works like 'ignore', but an error causes a
                        non-zero exit code; 'stop' halts on first failure and
                        yields non-zero exit code. A failure is any result
                        with status 'impossible' or 'error'.
````

Comment (yarikoptic) [inline datalad/distribution/create.py]: ah... I was running `datalad get --help` as to see options pertaining to that command.  I guess, ideally, we should adjust our `datalad COMMAND --help` to state that there could also be `datalad` options , i.e. have something like
```
Usage: datalad [DATALAD_OPTIONS] clone [--version] [-h] [-l LEVEL] [--pbs-runner {condor}]
                     [-d DATASET] [-D DESCRIPTION] [--reckless]
                     SOURCE [PATH]
....
* DataLad options*
   See `datalad --help` for the full description of the options available to any datalad command [not sure if we might want to at least list them all over here]
```
and may be strip away `--log-level` and `--pbs-runner` to be replicated in `COMMAND --help`.

yarikoptic:
> If you specify what the desired behavior is in terms of result properties (which status to report, which property to report, in which format) I can give it a try. 

Well -- I guess it should 
- be somewhat consistent with what `create`, `uninstall` and `remove` return since they all operate on datasets and can operate recursively (but `create`)
- per each argument (path) given to install have a record returned
  - thus probably have `return_type = 'item-or-list'`
- consider most common use-cases  how it would be invoked
  - e.g. in Python it is handy now to be able to say `ds = install('///openfmri/ds000001'); ds.get('sub-01')`
  - `create` also returns that created dataset as its return value, which makes it immediately usable
  - so may be for `install` by default, even with `recursive` we just return Datasets matching provided arguments (see above), and then allow to switch to return full list with statuses?  if some failure has happened, and we allowed to ignore failures, we return None for that "item".

makes sense, as at least for the default behavior?

mih:
> per each argument (path) given to install have a record returned thus probably have return_type = 'item-or-list'

What should happen if you install with data? There are two actions to be reported for the same path (`install`, and `get`), ignoring for now that git-annex also contributes results for files in this case. And ignoring for now that in the general case it could be quite hard to connect a subset of results with individual input arguments.

Given my current interpretation of the other requirements you gave I see two approaches: 1) Implement facilities to perform inference on a result list (no generator behavior seems possible, as we would need to choose one result over another depending on which ones were to appear (in general we get lower-level command results first)), or 2) keep the current code as is, and simply collect the result dicts from lower-level commands (add the few missing ones from `install` itself and return a list. That would make install somewhat compatible with the new decorator, despite not becoming a generator.

I am leaning towards 2), as I don't see a good way how to approach 1) in a way that is valid in the general case, and also for other commands.

mih:
I realized that it might be not very obvious why this is difficult. The reason is that with the implemented design decision making on what to report/render/transform is made *outside* the respective command, meaning that all individual commands can do is to "hide" results by not relaying them outside at all (something that would make them a lot less flexible to combine with other commands). Therefore all decision making regarding aspects like "results that correspond to input arguments" needs to be implemented in a generic fashion outside individual commands, hence guaranteeing that any sensible implementation needs to be generic and valid for any command.

That being said here is an alternative proposal that might be close enough to what you envisioned. `install`, by default, could use this configuration (pretty much identical to `create`):

````
return_type = 'item-or-list'
result_xfm = 'datasets'
result_filter = EnsureKeyChoice('action', ('install',)) & \
                      EnsureKeyChoice('status', ('ok', 'notneeded'))
````

IOW, it returns (a list of) Dataset instances (or a single Dataset item) for any dataset that gets newly installed, or is found requested (by any direct or indirect means) but is already present.

This leads to the situation that simple `install` calls that just ask to install a single dataset, yield a simple, single Dataset instance in return. Likewise, if you ask for two (sub-)datasets to be installed, you get a list of two Datasets instances.

However, it is important to keep in mind that this is not the same as *"per each argument (path) given to install have a record returned"*. Example: I ask it to install a subdataset in a dataset that requires two additional, intermediate levels of subdatasets to be install before it can actuall install the requested one.
In this case I get three Dataset instances returned. There is not even a guarantee that any of the results corresponds to an input argument. If in the previous example there is a typo in the path argument of the subdataset, it could easily happen that we get the two intermediate datasets reported as installed, but not the actually requested one (because it doesn't exist under this path). With `on_failure='ignore'` this will lead to a two-Dataset list return value (by default it errors, of course).

yarikoptic:
yeap -- this last one is very close! But ...

> The reason is that with the implemented design decision making on what to report/render/transform is made outside the respective command, meaning that all individual commands can do is to "hide" results by not relaying them outside at all (something that would make them a lot less flexible to combine with other commands). 

I think that in our desire to make interface as flexible as possibly desired for our own "integration" we should try to keep it simple and "mentally accessible" for users by default. 

To my naive taste connection of `input argument` -> `result item` seems to be easy to grasp and would eliminate need to analyze results for simple use cases such as `install(url, recursive=True)` which would in proposed solution spit out a possibly long list of installed datasets.  What I have requested was "install the dataset from url with all of its subdatasets" and I would like to get that dataset.  If I would be interested in getting access to subdatasets, I could use `get_subdatasets`.  I might not even know or care about having sub-datasets (e.g. whenever we finally provide HCP dataset with a subdataset or two per each subject).  I do respect though that probably it is not that easy to accomplish with current `result_filter` approach which probably lacks association with input arguments, i.e. with what was actually requested... or does it?

And then, since this behavior could be easily masked away with alternative `result_filter`, we (or anyone really into digging into the installation "progress") could ask to obtain those records for all actions which were performed.

> If in the previous example there is a typo in the path argument of the subdataset, it could easily happen that we get the two intermediate datasets reported as installed, but not the actually requested one (because it doesn't exist under this path).

`None` is then returned for that path, despite some intermediates were installed as well, and thus "_per each argument (path) given to install have a record returned_" could be fulfilled and be factually correct -- _the requested dataset was not installed_.  So "generator" behavior (as seen from outside) could be provided at the level of iterating over input arguments, although probably at a cost of loosing some parallelization (or just more complicated logic) since order would need to be maintained if we run those installs in parallel. Rendering of the intermediate results could keep on going interim (e.g. "installed dataset interim1", "installed dataset interim1/interim2", "failed to install dataset inter1/interim2/typo") if we decide to render them in favor over `lgr.info` msgs

yarikoptic:
@mih does it (my description above) make sense? ;)

mih:
@yarikoptic Makes sense -- did make sense even before that already ;-)

I was thinking on how to achieve that though without custom logic in install. I think I have a plan, but it needs a bit more testing. Will report once I have a conclusion.

mih:
See #1468 for work towards a solution for install along the lines suggested by @yarikoptic -- aiming for a non-install-specific solution.

yarikoptic:
conflicts are because of me I guess so I will take a look now

mih:
Merci!

mih:
FTR: #1476 is a candidate for merge in here.

I will not touch `publish` in this PR: version in master is substantially different.

mih:
FTR: Test failures are a timeout, and this (no idea):

````
FAIL: datalad.tests.test_tests_utils.test_get_resolved_values
    self.test(*self.arg)
  File "/home/travis/build/datalad/datalad/datalad/tests/test_tests_utils.py", line 145, in test_get_resolved_values
    eq_(flavors, _get_resolved_flavors(flavors))
AssertionError: ['networkish', 'local'] != ['local']
````

Issue #1418, opened 2017-03-22
save -u  commits moved annexed files without adjusting their symlinks


yarikoptic:
whenever annexed files are moved around, git annex adjusts paths in .git/hooks/pre-commit.
and that is happening just fine after the move if we just do 'git commit' or 'datalad save'

But if I do `datalad save -u` those with incorrect symlinks are committed and nothing gets fixed upon commit (I do  not see 'git commit' running, so I guess that is done via GitPython's commit?) and we ending up with broken symlinks all around... heh

correct of plain `save` (doesn't do any 'git annex add')
```shell
$> builtin cd /tmp/; chmod +w -R /tmp/testl; rm -rf /tmp/testl; datalad create /tmp/testl && cd /tmp/testl && mkdir d && echo 123 > d/123 && datalad add d/123 && git mv d/123 123 && datalad save && ls -l 123      
[INFO   ] Creating a new annex repo at /tmp/testl 
Created dataset at /tmp/testl.                                                                                                                                                                            
W: git-annex repositories not (yet) supported in the prompt
Added /tmp/testl/d/123                                                                                                                                                                                    
Saved state: d565e942c8f9cd7aa4e9ed8b3daf22f1a206ac61 for <Dataset path=/tmp/testl>
lrwxrwxrwx 1 yoh yoh 108 Mar 22 14:29 123 -> .git/annex/objects/pF/Zf/MD5E-s4--ba1f2511fc30423bdbb183fe33f3dd0f/MD5E-s4--ba1f2511fc30423bdbb183fe33f3dd0f
```

incorrect operation of save -u:
```shell
$> chmod +w -R /tmp/testl; rm -rf /tmp/testl; datalad create /tmp/testl && cd /tmp/testl && mkdir d && echo 123 > d/123 && datalad add d/123 && git mv d/123 123 && datalad save -u && ls -l 123 
[INFO   ] Creating a new annex repo at /tmp/testl 
Created dataset at /tmp/testl.                                                                                                                                                                            
W: git-annex repositories not (yet) supported in the prompt
Added /tmp/testl/d/123                                                                                                                                                                                    
Saved state: b5c7e86e4d3e6f67479469152020e0c5dc5f8bb1 for <Dataset path=/tmp/testl>
[...]

PR #1420, opened 2017-03-23
BF+ENH: *Repo.add(updates=False), do not call git annex add without files upon save

### LINKED ISSUE #1419: save recursive saves untracked files in submodules
while trying to fix up #1418 discovered that  it still does it .... heh... I guess manifestation of old behavior (set in stone with a test! ;) ) or that is somehow intentional @mih?

this is the test in question within test_save:
```python
    states = [d.repo.get_hexsha() for d in (ds, subds, subsubds)]
    testfname = opj('sub', 'subsub', 'saveme2')
    with open(opj(ds.path, testfname), 'w') as f:
        f.write('I am in here!')
    assert_true(ds.save(all_updated=True, recursive=True))
    newstates = [d.repo.get_hexsha() for d in (ds, subds, subsubds)]
    for old, new in zip(states, newstates):
        assert_not_equal(old, new)
```
so it checks that save indeed happens when we create a new file deep in hierarchy

yarikoptic:
I saw no other way but to ask 'git add -u --dry-run' for what files to be added before calling 'git annex add' which should not called without specific files path, since it would just add everything, which lead to #1419.  IMHO we better operate slower but correctly.

This is my 2nd attempt (see https://github.com/datalad/datalad/compare/master...yarikoptic:bf-save-u if interested, but that one was 'shortsighted' although I have spent probably half a day on it learning intricacies of the save), and I did try to keep it minimal.  While at it I have fixed the tests which still hoped to get stuff automagically added.  I guess we are to introduce '-a' option back sooner than later ;-)

But while analyzing the 'save', I was converging on an idea that we might not need/want '-u' since we will not be able to provide consistent interface.  Please see the newly added 'save' spreadsheet in the API Use-cases convergence google doc.  E.g. SUC6.  So I am starting to wonder if we should just always save all staged and non-staged (but tracked) changes?

Unlikely I will be up whenever the tests fail, so if you could contribute to fixing them -- would be most welcome!
Closes #1419

FTR  Joey agreed that having 'annex add -u' is sensible thus there is http://git-annex.branchable.com/todo/annex_add___40__-u__124__--update__41___mode/ now  but probably will not happen right away, thus this solution might still be sensible for now

### DIFF
--- a/datalad/interface/save.py
+++ b/datalad/interface/save.py
@@ -162,6 +162,8 @@ class Save(Interface):
         if all_updated:
             # and we do this by replacing any given paths with the respective
             # datasets' base path
+            # XXX this implies no "boundary" possibly imposed by specified paths!
+            #  see our use-case spreadsheet -- TODO
             for ds in content_by_ds:
                 content_by_ds[ds] = [ds]
 
--- a/datalad/interface/tests/test_save.py
+++ b/datalad/interface/tests/test_save.py
@@ -12,6 +12,7 @@
 
 __docformat__ = 'restructuredtext'
 
+import os
 from os.path import join as opj
 from datalad.utils import chpwd
 
@@ -112,7 +113,10 @@ def test_recursive_save(path):
     with open(newfile_name, 'w') as f:
         f.write('some')
     # saves the status change of the subdataset due to the subsubdataset addition
-    assert_equal(ds.save(all_updated=True), [ds])
+
+    res = ds.save(all_updated=True)
+    assert ds.repo.dirty
+    #assert_equal(res, [ds])
 
     # make the new file known to its dataset
     # with #1141 this would be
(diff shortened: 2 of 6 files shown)


yarikoptic:
tried also to "fix" #1418 by avoiding providing the full path to the 'commit' call but sure thing failed since other "logics" depend on commit being called with the path of the dataset apparently -- I guess it has some additional effects I am not fully aware of yet ... heh

yarikoptic:
heh -- first buildbot has finished with "FAILED (SKIP=56, errors=2, failures=6)" and they all look like there is more of reliance on 'add even untracked files by default' behavior... heh -- should that become a default ;-)

mih:
@bpoldrack also made a few statements regarding a default "add/save" behavior based on technical constraints in direct/v6 mode. I agree we should reevaluate our targeted behavior in this light.

yarikoptic:
FWIW: 6.20170525+gitge1cf095ae-1~ndall+1 of annex already carries 'annex add -u' ... I wonder though in the light of the all the recent changes what remains pertinent to this case here?  may be at least some tests could be adopted/tried to see if we still have the same issue and may be to start using 'annex add -u'? @mih and @bpoldrack  -- what do you think?

yarikoptic:
@mih is this one entirely superseded now?

mih:
@yarikoptic I am no 100% sure, but I believe many previously problematic cases are no longer relevant, because we are now passing much more info into the function in question to be able to make better decisions. It could be closed, I think.

yarikoptic:
Ok

PR #1421, opened 2017-03-23
Advanced magic in `crawl` test, need exorcist ;-)

mih:
I am trying to debug a test failure in the crawler for the new `subdataset` command. It "fails" because it doesn't return subdatasets for a Dataset that is an empty directory (hence not installed). However, the old Dataset.get_subdatasets() happily reports the presence of subdatasets. Here is the situation using the old code:
````
> /home/mih/hacking/datalad/git/datalad/interface/crawl.py(150)__call__()
-> subdatasets = Dataset(os.curdir).get_subdatasets(recursive=recursive)
(Pdb) ds=Dataset(os.curdir)
(Pdb) ds.path
'/tmp/datalad_temp_test_crawl_api_recursiveVTk5BI'
(Pdb) Dataset(os.curdir).get_subdatasets(recursive=recursive)
['path1', 'path1/path1_1', 'path2', 'path_to_fail']
(Pdb) import os
(Pdb) os.listdir(ds.path)
[]
````

So the beast is not there, yet it has subdatasets. I assume that is GitPython still remembering things it should have forgotton. The new code, doesn't use GitPython functionality, so it returns an empty list -- as also the old code should have done IMHO.

Somehow, not sure how to tackle this.

In any case, the new code is in #1423

### DIFF
--- a/datalad/interface/crawl.py
+++ b/datalad/interface/crawl.py
@@ -147,6 +147,8 @@ class Crawl(Interface):
                 # into the subdatasets' crawl.  We will collect all of them here so we might later
                 # also introduce automatic commits when super-dataset got successfully updated
                 subdatasets = Dataset(os.curdir).get_subdatasets(recursive=recursive)
+                if len(subdatasets):
+                    assert Dataset(os.curdir).is_installed()
 
                 lgr.info("Crawling %d subdatasets", len(subdatasets))
                 output = [output]

### CODE REVIEW

**Comment by mih** [comment]
Pushed an assertion that should show that things are broken. With correct behavior of `get_subdatasets` I get this test failure:

````
FAIL: datalad.interface.tests.test_crawl.test_crawl_api_recursive
    self.test(*self.arg)
  File "/home/mih/hacking/datalad/git/datalad/tests/utils.py", line 863, in newfunc
    reraise(*exc_info)
  File "/home/mih/hacking/datalad/git/datalad/tests/utils.py", line 839, in newfunc
    func(*args, **kwargs)
  File "/usr/lib/python2.7/dist-packages/mock/mock.py", line 1305, in patched
    return func(*args, **keywargs)
  File "/home/mih/hacking/datalad/git/datalad/tests/utils.py", line 564, in newfunc
  File "/home/mih/hacking/datalad/git/datalad/interface/tests/test_crawl.py", line 84, in test_crawl_api_recursive
    assert_equal(output, [[]]*4 + [None])  # for now output is just a list of outputs
AssertionError: Lists differ: [[]] != [[], [], [], [], None]

Second list contains 4 additional elements.
First extra element 1:
[]

- [[]]
+ [[], [], [], [], None]
````

**Comment by mih** [comment]
Just to be clear, I need advice on how to adjust the test, and an opinion on implications on crawler behavior.

**Comment by bpoldrack** [comment]
> So the beast is not there, yet it has subdatasets. I assume that is GitPython still remembering things it should have forgotton.

I don't think so. Look at this:
`test_crawl.py` line 64
```python
@patch('datalad.distribution.dataset.Dataset.get_subdatasets',  # return_value=['path1', 'path2'])
    side_effect=[
        ['path1', 'path1/path1_1', 'path2', 'path_to_fail'],
        # ATM we will not recursively crawl within sub-datasets
        # ['path1_1'],  # sub-dataset of the first sub-dataset
        #[],  # so it would get crawled and have no further sub-datasets
        # [],  # no sub-datasets in path2
        # will fail to return anything for path_to_fail
    ]
)
```

mih:
Hmm, so I just adjust the patched object and make it return invalid values again? I don't get it.

bpoldrack:
Well, I don't get that test either. Just wanted to point out, that `get_subdatasets` wasn't broken somehow remembering removed submodules, but it was patched to return that shit instead.

bpoldrack:
I'd try to apply that patch to the new one instead, hoping that the applied side effect will fix the test ;-)

yarikoptic:
Will look at code when get to work - smells like I had a mock to avoid something

mih:
Gracias!

yarikoptic:
oh -- this was probably my flagman of the mocked testing! ;)  E.g. with this test I have the goal was to test the logic of recursive crawling, without actually doing any expensive crawling/interaction with git. The entire test runs in 460ms, tests 33% of the datalad codebase, and excercises/test if recursive crawling works out as it should.  As at that point get_subdatasets was returning just paths -- that is what mocked (side_effects) return values were.   None of those ever physically ever existed!  Now that it would return Dataset objects, I will see what I should do, but probably that assert you have added would be only a stumbling point or irrelevant.

mih:
So this was useless apparenty. Issue in #1423 still stands.

yarikoptic:
@bpoldrack :

> Well, I don't get that test either. 

I would advise to have a second look at it , and all the patch decorations predefined -- it establishes a scenario to test code without running those underlying calls to test specific logic under assumption that mocked out interfaces work correctly.  Talk by Gary Bernhardt might be informative as well: https://www.youtube.com/watch?v=RAxiiRPHS9k

mih:
I get that. This particular test however, does not test that the interface that it "calls" in this particular way does actually return values like the mockup plants their. Just renaming the mocked function alone would have made the test pass -- no need to adjust the actual call in the crawler code. An actual crawler run would fail -- maybe still does -- it is not tested. Just mocked. In this particular test -- dunno if that is tested elsewhere properly.

yarikoptic:
Yes -- that is why I said "to test the logic", not the interface, which might or might not be (but should have been) tested elsewhere indeed separately.

Issue #1422, opened 2017-03-23
Need ability to execute Git command via our helpers (also in direct mode) without all of GitPython


mih:
During `subdatasets` command RF. I find myself needing to run `git submodules` but I cannot use custom_git_command, because I need to avoid Repo instantiation. I see no way to do this while retaining all  of the support for direct mode. @bpoldrack

mih:
Related PR in #1423

bpoldrack:
I think the cleanest way would be to RF GitRepo to have a lazy loading `repo` (GitPython `Repo`) attribute. That way instantiation would be cheap and as long as `get_submodules` and the likes don't use GitPython, that command should remain quick. But this might take a bit of time to discover side effects in the tests ...

Alternatively there would be a quick solution for `subdatasets` command. It would be kind of a code duplication and not automatically 'updated' whenever `GitRepo` is improved. But it's quickly available.
This is, what it would look like:

In `_parse_git_submodules(dspath)` you need to look into `dspath/git/config` and figure out, whether it is in direct mode (see `AnnexRepo._is_direct_mode_from_config`). If so just add `-c core.bare=False` to your `git submodule` call.

mih:
Cool. Any reason not to add this option unconditionally? Seem to not have side-effects on regular repos.

mih:
Hmmm:

````
/tmp/direct (git)-[annex/direct/master] % git -c core.bare=False submodule
fatal: $program_name cannot be used without a working tree.
````

bpoldrack:
The reason is in that 'seems' ;-)
I really don't know under what circumstances there would be side effects. Plus: Sadly it doesn't work with all commands under all circumstances. Sometimes we still need `annex-proxy` or even more complex workarounds

bpoldrack:
@mih: You're using recent git, right? I forgot.

There is this bug, that this option's priority over `.git/config` isn't repected. Workaround in `AnnexRepo`: Write it to `.git/config` before the call and restore old value afterwards.

See `AnnexRepo._git_custom_command`

mih:
Too bad, that makes one call to `git config` to figure out whether this is needed and two more in direct mode to mitigate. Thanks anyways.

mih:
What about this?

````
git --work-tree=. submodule
````

Seems to work, requires no config ping pong.

bpoldrack:
Hm. Sounds like it should work for submodule commands indeed.

bpoldrack:
[...]

Issue #1412, opened 2017-03-22
datalad add is obnoxious (runs git/annex add 2 times) and WRONG (adds to annex) in disregarding .gitattributes


yarikoptic:
discovered while trying to use 0.5.0
```shell
hopa:/tmp/test2
$> cat .gitattributes; echo 128 > 128.txt; DATALAD_LOG_LEVEL=1 datalad add 128.txt 2>&1 | grep 'Running.*git.*add'
* annex.largefiles=(not(mimetype=text/*))                                                                         
[DEBUG] Running: ['git', '-c', 'receive.autogc=0', '-c', 'gc.auto=0', 'add', '--dry-run', '-N', '--ignore-missing', '--verbose', '128.txt']
Total:   0%|          | 0.00/4.00 [00:00<?, ?B/s][DEBUG] Running: ['git', '-c', 'receive.autogc=0', '-c', 'gc.auto=0', 'annex', 'add', '--debug', '--json', '-c', 'annex.largefiles=anything', '128.txt']
[DEBUG] Running: ['git', '-c', 'receive.autogc=0', '-c', 'gc.auto=0', 'add', '--dry-run', '-N', '--ignore-missing', '--verbose', '128.txt']
[DEBUG] Running: ['git', '-c', 'receive.autogc=0', '-c', 'gc.auto=0', 'annex', 'add', '--debug', '--json', '128.txt']
(dev) 2 14279 [2].....................................:Wed 22 Mar 2017 09:39:30 AM EDT:.
hopa:/tmp/test2
$> ls -ld 128.txt            
lrwxrwxrwx 1 yoh yoh 116 Mar 22 09:39 128.txt -> .git/annex/objects/X9/77/MD5E-s4--650a1c9c9baa20730b4fcfdbe4cdc135.txt/MD5E-s4--650a1c9c9baa20730b4fcfdbe4cdc135.txt
```

so you can see that first it 'git add's in dry, then explicitly adds into annex by specifying largefiles=anything, and then again adds it dry, and then again, finally calls what imho should be the correct call to annex

mih:
FTR: The first is `add`, the second wave is from `save`. We recently looked at that too.

yarikoptic:
I am looking into it ATM at least to make it work correctly (may be still obnoxiously ;) )

Issue #1411, opened 2017-03-22
Strip `Dataset` of methods and expose them as (plumbing) commands


mih:
The `Dataset` class is an outlier in our API, because it is only available in Python. As a result its methods are more "special" than our usual commands that are bound to the `Dataset` class. Moreover, the functionality they provide (e.g. `get_subdatasets`) is not available through the cmdline API.

We should carefully evaluate which methods could be exposed as regular commands.

#1423 
#1424

mih:
Related to #1400

yarikoptic:
are you saying that `.get()`, `.publish()` and other methods of the Dataset should be gone??
IMHO Dataset is not the outlier -- it is the core!  ;-)

mih:
Those are the commands, they are only bound as methods in addition to being commands.

mih:
`Dataset.register_sibling` is totally useless.
`get_subdatasets` should be a command.

Not yet sure about `get_containing_subdataset`.

yarikoptic:
I know that those are commands.
ah -- you are saying to unify more -- i.e. that majority of Dataset methods should be similar commands.  

- `get_subdatasets` might indeed be useful. ATM it is just `datalad ls` which could reveal that information to the user to some degree
- `get_containing_subdataset` also not sure since it is primarily for internal usage

mih:
Yup. PR for a `subdataset` command is coming.

mih:
Done from my POV

PR #1423, opened 2017-03-23
RF+NF: Dataset.get_subdatasets() -> standalone `subdatasets`

mih:
In response to #1411

This implementation is a lot faster, as it completely avoids instantiation of repositories, but wraps `git submodule` instead. Moreover, it exposes more information in a more directly accessible way.

It uses the new command API, so all transformation and filter mechanisms can be applied based on arbitrary subdataset properties.

The old `Dataset.get_subdatasets()` is left in place, as it is still used a lot in the tests. All occurrences in the actual code base are RF'ed. Use in test triggers a warning identifying the caller to
streamline further RF.

The new `subdatasets` command as advertised as a plumbing command in the docs.

Tests in `test_subdatasets.py` have been extended to near 100% coverage, only one safeguard line is missing.

TODO:

- [x] make it work in direct mode #1422 
- [x] figure out how crawler should behave without broken `get_subdataset` behavior #1412 
- [x]  base branch still needs to fixes from https://github.com/mih/datalad/pull/18

### DIFF
--- a/datalad/api.py
+++ b/datalad/api.py
@@ -89,7 +89,7 @@ def _generate_func_api():
             # TODO: BEGIN to be removed, when @build_doc is applied everywhere
             spec = getattr(intf, '_params_', dict())
             api_name = get_api_name(intfspec)
-            if api_name not in ('update', 'save', 'create', 'unlock', 'clean', 'drop', 'uninstall', 'get', 'clone'):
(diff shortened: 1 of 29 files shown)


mih:
Test runtime down to ~30min.

mih:
Now all good on PY3 too. Merging.

Issue #1401, opened 2017-03-20
Reduce number of `git config` runs?


yarikoptic:
In light of #1383 decided to whine about may be considering to make git config calls less frequent?  now that we are using flyweight pattern for all the datasets/*Repos, we have centralized `.config` access and could track when/if any change was made thus requiring or not reload of configuration upon next read.
Here is eg. a sample to get a taste of abundance of those calls:
```shell
(git)hopa:~datalad/datalad[opt-tests]git
$> DATALAD_LOG_TIMESTAMP=1 DATALAD_LOG_LEVEL=DEBUG python -m nose -s -v datalad/interface/tests/test_utils.py:test_save_hierarchy 2>&1 | tools/dtime | sort -n | grep git.*config | nl
...
     7	    5 2017-03-20 10:05:03,773 [DEBUG] Running: ['git', 'config', '-z', '-l']
     8	    5 2017-03-20 10:05:03,928 [DEBUG] Running: ['git', 'config', '--local', 'annex.backends', 'MD5E']
     9	    5 2017-03-20 10:05:03,974 [DEBUG] Running: ['git', 'config', '-z', '-l']
...
   294	  116 2017-03-20 10:05:20,278 [DEBUG] Running: ['git', 'config', '-z', '-l']
   295	  125 2017-03-20 10:05:22,925 [DEBUG] Running: ['git', 'config', '-z', '-l']
   296	  164 2017-03-20 10:05:19,365 [DEBUG] Running: ['git', 'config', '-z', '-l']
   297	  165 2017-03-20 10:05:17,883 [DEBUG] Running: ['git', 'config', '-z', '-l']
   298	  257 2017-03-20 10:05:03,427 [DEBUG] Running: ['git', 'config', '-z', '-l']
DATALAD_LOG_TIMESTAMP=1 DATALAD_LOG_LEVEL=DEBUG python -m nose -s -v  2>&1  7.04s user 4.01s system 55% cpu 20.085 total
```
so not necessarily that (2nd column) is the timing for running those commands (forgot how to enable that timing protocolling... did we ever finish it? ;-)), but it gives an idea at least on a number of those calls -- almost 300 in the duration of that test.  Even with 5 msec for each, that is 1.5sec of the total run time of the test.

mih:
Looked at this when I created #1044. Not that simple.

mih:
Will give it a try another try.

bpoldrack:
[...]

PR #1427, opened 2017-03-23
ENH: Configuration file mtime tracking to reduce `git config` call frequency

mih:
Fixes gh-1401
Fixes gh-1044


mih:
Well, there seems to be more to it...

mih:
Hmm, so the 16.04 `git config` doesn't have `--show-origin`?

mih:
Bounced buildbot tests to see if nonsense OSX logs are reproducible.

mih:
So OSX fails reliably. Given the nature of the error it seems that the entire approach is not working and the config is fundamentally broken. Not sure what to do, I have no juice to investigate this on OSX. I could disable support for mtime tracking on OSX and be done.

yarikoptic:
I thought I have replied -- I guess again did it in email though already know that github wouldn't post those reliably for me ...  now that the problem showed itself I recalled  -- that on some OS/file systems mtime would have precision up to a second.  I think I saw it on linuxes as well. And that is what is happening on OSX:
```shell
datalads-imac:~ yoh$ touch blah; python -c 'import os, time; t1=os.stat("blah").st_mtime; time.sleep(0.5); os.system("touch blah"); t2=os.stat("blah").st_mtime; print t1, t2, t2-t1 '; rm blah
1491099043.0 1491099043.0 0.0
datalads-imac:~ yoh$ touch blah; python -c 'import os, time; t1=os.stat("blah").st_mtime; time.sleep(0.5); os.system("touch blah"); t2=os.stat("blah").st_mtime; print t1, t2, t2-t1 '; rm blah
1491099044.0 1491099045.0 1.0
```
so -- slept 0.5 sec which would some times then result in 0 and some times in 1 sec diff between mtimes.  So decision should be made not based on OS but on filesystem's properties.

mih:
Ah, thanks! That makes sense now. I have to think about where we would want to include this kind of detection.

mih:
From http://stackoverflow.com/questions/19059877/python-os-path-getmtime-time-not-changing:

Some statistics on the time resolution of various common file systems:

FAT32: 2s
ext3: 1s
exFAT: 10ms
NTFS: 100ns
ext4: 1ns

mih:
Hoping that my last commit makes OSX happy again, here are the stats (refer to #1401 for comparison):

````
DATALAD_LOG_TIMESTAMP=1 DATALAD_LOG_LEVEL=DEBUG python -m nose -s -v datalad/interface/tests/test_utils.py:test_save_hierarchy 2>&1 | tools/dtime | sort -n | grep git.*config | nl 

     1	   17 2017-04-02 09:24:21,783 [DEBUG] Running: ['git', 'config', '-z', '-l', '--show-origin']
     2	   17 2017-04-02 09:24:22,718 [DEBUG] Running: ['git', 'config', '-z', '-l', '--show-origin']
     3	   17 2017-04-02 09:24:23,413 [DEBUG] Running: ['git', 'config', '--local', 'annex.backends', 'MD5E']
...
   170	  113 2017-04-02 09:24:34,005 [DEBUG] Running: ['git', 'config', '-z', '-l', '--show-origin']
   171	  116 2017-04-02 09:24:32,437 [DEBUG] Running: ['git', 'config', '-z', '-l', '--show-origin']
   172	  226 2017-04-02 09:24:19,930 [DEBUG] Running: ['git', 'config', '-z', '-l', '--show-origin']
DATALAD_LOG_TIMESTAMP=1 DATALAD_LOG_LEVEL=DEBUG python -m nose -s -v  2>&1  16,37s user 4,50s system 61% cpu 33,940 total
````

Almost doubled the number of calls again, but should now work even in FAT32 (dataset on USB stick...).

mih:
Well, now OSX is OK (travis too), but all others fail due to a strange timeout -- which I cannot replicate locally. All healthy with the builtbots @yarikoptic ?

yarikoptic:
at least yesterday seemed to be ok in https://github.com/datalad/datalad/pull/1444 which I will restart once more now -- if so, that means that buildbots are more or less healthy and then we will go from there

mih:
Ok, in this case we have total linux buildbot failures, local is good travis is good. Buildbot failure is timeout.

yarikoptic:
tried rebuilding one -- worked just fine.... so pushed a minor change to trigger full rebuild

Issue #1428, opened 2017-03-23
create-sibling might want to make remote repo use current branch as the default


yarikoptic:
ATM it is just always 'master', but if we are working in another branch and want that one to be default I need manually to go to the server and check it out after push, and run hook (post-update) manually (if it is a datalad web frontend). Example dataset http://datasets.datalad.org/?dir=/workshops/nipype-2017/ds000114

Although it would be more of a magic to switch a branch upon publish, at least during `create-sibling` IMHO target location should have current local branch checked out.  Or if at publish time -- we would need to verify that current local branch/hexsha on local dataset corresponds to the same branch/hexsha on remote location and then checkout on remote end, and rerun post-update.  Actually we might as well just migrate necessary for web-frontend actions into 'post-checkout' hook

mih:
This issue is describing one of many aspects that require a general overhaul of the `create-sibling` command. This overhaul effort is being tracked by the meta-issue #6149. I am closing this issue here, after having added it to that TODO list.

Issue #1424, opened 2017-03-23
RF `Datalad.recall_state() -> `load` command


mih:
Interesting use cases:

- Reset an entire tree of datasets to a known previous state. Could become a helper from some kind of `undo` #748 
- Switch an entire tree of datasets to a certain branch

yarikoptic:
sounds useful indeed ;)  but I guess after all the other revolutions settle down?

yarikoptic:
We might need to record a branch name associated with submodule state somewhere

mih:
@yarikoptic branch name is indeed an issue. We could start adding this info to .gitmodules. The `subdatasets` command already allows for storing additional info. We would also need to make sure that on `save` this info is updated if needed.

If we have branch info in `.gitmodules` we could `git reset --hard` that branch to the state recorded in the parent dataset (or error, if there is no such branch).

More tricky is the treatment of subdatasets that were added after the target state. We would need to go through and uninstall the respective subtrees. `diff` can discover those already.

yarikoptic:
Let me digest/explore a bit about storing branch name within .gitmodules since initial reaction was -- it is a recipe for disaster...  

In #748 I need(ed) a helper for "local" undo's, pretty much like reflog, which seems to record also the branch information.  Those "states" were not envisioned to be shared with anyone (the same with  reflog), so it was totally fine to store branches information

yarikoptic:
hm, it seems that in general that is already "official" way git submodules support branches:
```shell
$> git submodule add --branch buga ../ds2 ds2
Cloning into '/tmp/ds3/ds2'...
done.
# there was also git commit I think

$> ls ds2
123@

$> cat .gitmodules 
[submodule "ds2"]
	path = ds2
	url = ../ds2
	branch = buga
```
and `datalad install -r` already worked out nicely with pointer(s) to specific branches.  So indeed, probably again under-used (and thus possibly buggy) feature of git submodule's, but there is already branch specification reserved in .gitmodules.
The question would be though, when do we add this branch'ing information?  In the case of already crawled datasets I think in 100% of the cases it is the 'master' branch is assumed. But if we implement this handling of multiple repos in sync, I can see myself annotating submodules in .gitmodules with branches for 'incoming', etc, so when we get to organizing some really large dataset (like HCP, but with archives to be extracted) I would then use it.
[...]

PR #1431, opened 2017-03-24
DOC+ENH: write up on construction of a 'derived' dataset and adding it to larger distribution

yarikoptic:
Just thought to formalize my notes from creating http://datasets.datalad.org/?dir=/workshops/nipype-2017/ds000114 

Also adjusted script running those tests... I am really not sure about the purpose of shunit2 magic/stubs in there, may be @mih would explain.

- Ideally should get #1428 fixed soonish and remove custom stub
- perspective 'load' #1424 command could be used to checkout into a detached branch (although I should have just based on actual version checkout... heh -- next time)
- publish still behaves flaky -- it seems that after all this it manages not to push most recent update to top dataset and requires additional 'datalad publish'... although worked just fine until I started to correct for that first issue

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -163,7 +163,7 @@ script:
   # Run javascript tests
   - grunt test --verbose
   # Run doc examples if no spaces in the TMPDIR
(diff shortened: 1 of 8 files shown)


yarikoptic:
@bpoldrack -- could you please check what is not working in direct mode in my demo script? thanks in advance

PR #1443, opened 2017-03-26
ENH: bring coverage back to the one in master even without running all the environments in travis

yarikoptic:


yarikoptic:
there are still some lines I believe which might be covered in master but not in PRs but largely should be better and PRs should have "improved coverage" from on ;)  I do not expect much of conflicts with other PRs, so merging

PR #1446, opened 2017-03-31
NF: datalad-zenodo special remote to push off the data files to zenodo

### LINKED ISSUE #23: CERN's Zenodo
Provides portal to upload software/data with versioning etc. See
https://zenodo.org/search?p=980__a:dataset
so might be useful to keep in mind at least for "crawling" and exposing via annex, ideally -- for publishing to ;)

yarikoptic:
Not yet sure if we want actually implement such type of publishing through "git-annex special remote".
More later -- for now just want to see if I broke anything

- [ ] investigate possibility to upload with subdirectories structure.  So far it seems to be not possible (see edit 1 below)
- for final "publish" command:
  - [ ] should add regular (`web` remote) urls for each file, since published version could have access to them
  - [ ] we might want to upload also the git repository itself as a tarball. Then we could easily implement (one time) "datalad clone" from it, which would fetch/extract that tarball and then be able to fetch the files as usual
- [ ] atm we (ab)use 'STATE' which annex allows to associate with key/remote, to store the file id as zenodo knows it, but may be we would be better off by providing `datalad-zenodo` urls, similar to datalad-archives url for that purpose. not sure yet
- [x] fixes #23

edit 1: I thought that zenodo can provide files from subdirs, but apparently that was incorrect perception -- what I thought was subdirs, was listing of the content of .zip file (e.g. for github projects).  Still waiting on a possible reply from zenodo folks about this feature. If there is no way, the closest solution could be to upload with some file for that key original name + some additional suffix corresponding to beginningof the key checksum to disambiguate possible collisions.  E.g. instead of `MD5E-s10025785--0e790e210a275d76e3f2f4d895df6358.nii.gz` it would become `sub-02_T1w_0e790e21.nii.gz`.  I don't think it would be likely to get a collision but we could add some additional sanity check (at a cost of even slower interactions)


yarikoptic:
Failing buildbot on OSX (failed to lock reports by GitPython) smells with some singletons-causing-trouble behavior... yet to look inside

yarikoptic:
ok, @mih -- pushed resolved conflicts. But when reviewing the diff saw

```
+        if not ui.yesno(
+            """\
+Published on Zenodo datasets acquire publicly visible URL, doi, etc. Also, 
+depending on the access_type, files could be fetched from regular URLs which 
+we will assign to every file.  But you will not be able to introduce any further
+changes to the uploaded files""",
+            title='Are you sure you want to finalize/publish your Zenodo dataset?'
+        ):
+            return
```
which now brings memory of another limitation -- similar to figshare that once "published", I believe it could not be changed.  May be things changed since then?

oumaimahajji:
Hello, we are currently working on implementing Zenodo as a new special remote for git-annex. We are using the AnnexRemote library to do this, but we have noticed while researching, that there was a plan to make a Zenodo export here for Datalad. 

Is there a reason why you didn't implement this remote or any other technical difficulties that hindered this idea?

adswa:
Hey, those are exciting plans! From our end, I think it was mainly a lack of time that has yet prevented us to approach a Zenodo special remote.

I'm not 100% sure, but someone may have done the Zenodo roundtrip for CONP and could share their experience (tagging @glatard - sorry if I am mistaken here).

oumaimahajji:
Thanks for the response, I'm currently looking into how they implemented the crawler for Zenodo and if the functionalities are similar to the ones we aim to have.

glatard:
Hi @oumaimahajji and @adswa, the CONP crawlers are at https://github.com/CONP-PCNO/conp-dataset/blob/master/scripts/Crawlers/BaseCrawler.py The Zenodo sub-class is at https://github.com/CONP-PCNO/conp-dataset/blob/master/scripts/Crawlers/ZenodoCrawler.py

@joeyzhou98 could provide more information if needed.

adswa:
thx @glatard!

PR #1450, opened 2017-04-04
BF: direct mode/v6 status PR against return value revolution

### LINKED ISSUE #781: Enhance GitRepo.dirty() to be as flexible as GitPython's is_dirty() repo function.

### LINKED ISSUE #1133: *Repo.repo.is_dirty() is broken in direct_mode

### LINKED ISSUE #1170: *Repo.commit should report ValueError if instructed to commit untracked content

### LINKED ISSUE #1180: ok_clean_git() broken in direct mode

### LINKED ISSUE #1289: we need a better concept than staged/unstaged

### LINKED ISSUE #1291: Let it be OK to call GitRepo.commit() when there is nothing to commit

bpoldrack:
This replaces PR #1277 .

Intense testing/development for `AnnexRepo.status()` and `XXXRepo.dirty()`
Making sure this works in v5 direct mode, v5 indirect mode, v6, v6 adjusted branch.
In all cases includes submodules and even mixed hierarchies of plain git, and different annex type repos. The latter can only fully be achieved with a `status` method at the level of `Dataset`, which will also be a datalad command. Since the things already in here should be merged ASAP, this will be completed by a PR on top of this one. Therefore the following original tasks are to be completed in that next PR, too:
- [ ] ~make submodule recursion despite of direct mode submodules also work when started at a plain git~
- [ ] ~document everything~
- [ ] ~provide proper documentation: what (compound) operations on annexes we can have reliable assumptions about (before state vs after state) and what are these assumptions.~

- [x] Closes #781 
- [x] Closes #1180 
- [x] Closes #1133
- [x] Closes #1291 
- [x] Closes #1170
- [x] Closes #1289 (depends on what else waits to be discovered herein)
- [x] move calls to annex-adjust, annex-lock, annex-unlock to AnnexRepo

Note:
In the former PR (#1277) there was an unfinished discussion on the question of whether to provide a `dirty` property and to rename the method to `is_dirty`. I wont do it herein. This becomes clearer with the full shaped `status` at `Dataset` level and a test helper `ok_status`, which I think should replace the enhanced usages of `ok_clean_git`.

Note2:
This also merges master into rf-returnvalues!


yarikoptic:
> In the former PR (#1277) there was an unfinished discussion on the question of whether to provide a dirty property and to rename the method to is_dirty. I wont do it herein. This becomes clearer with the full shaped status at Dataset level and a test helper ok_status, which I think should replace the enhanced usages of ok_clean_git.

I really don't get it why not herein....  why now we need to
1. break the API compatibility with anything which could have used `.dirty` property so it becomes `.dirty()`
2. make this diff longer than needed since many changes are just that `dirty` -> `dirty()`
I really don't get it

mih:
PY3 failure is a forgotton? `print`.

yarikoptic:
re `print` statement -- I felt like it was left there on purpose to crash and alert to incomplete yet smth ;)  so I didn't fix it in my "attempt to help" PR

bpoldrack:
@datalad/developers : ATM I have no idea what is going on. Mow a lot of SSH related tests are failing. At the state the both of you commented on last, this didn't happen. But reverting the last two commits doesn't prevent it from happen. So - seems unrelated.

yarikoptic:
https://github.com/datalad/datalad/issues/1456

Issue #1459, opened 2017-04-07
cannot install over ssh a subdataset with space in its name


yarikoptic:
```
hopa:/tmp
$> rm -rf testds*; datalad create --no-annex testds; datalad create -d testds "sub ds"; ls -l testds/sub\ ds -a 
[INFO   ] Creating a new git repo at /tmp/testds 
Created dataset at /tmp/testds.
[INFO   ] Creating a new annex repo at /tmp/testds/sub ds 
Created dataset at /tmp/testds/sub ds.                                                                                                                  
total 16
drwx------ 4 yoh yoh 4096 Apr  7 16:58 ./
drwx------ 5 yoh yoh 4096 Apr  7 16:58 ../
drwx------ 2 yoh yoh 4096 Apr  7 16:58 .datalad/
drwx------ 9 yoh yoh 4096 Apr  7 16:58 .git/

$> datalad install -r -s localhost:/tmp/testds testds-clone                                                                   
[INFO   ] Installing dataset at /tmp/testds-clone from localhost:/tmp/testds 
[INFO   ] Installing subdataset <Dataset path=/tmp/testds-clone/sub ds> 
[WARNING] Target /tmp/testds-clone/sub ds already exists and is not an installed dataset. Skipped. 
[ERROR  ] 'NoneType' object has no attribute 'get_active_branch' [utils.py:_install_subds_from_flexible_source:223] (AttributeError) 

$> datalad install -r -s /tmp/testds testds-clone-local           
[INFO   ] Installing dataset at /tmp/testds-clone-local from /tmp/testds 
[INFO   ] Installing subdataset <Dataset path=/tmp/testds-clone-local/sub ds> 
[INFO   ] Obtaining <Dataset path=/tmp/testds-clone-local/sub ds> underneath . recursively 
2 installed items are available at
<Dataset path=/tmp/testds-clone-local>
<Dataset path=/tmp/testds-clone-local/sub ds>
```

edit 1: issue remains in returnvalues branch where some fixes were implemented already

PR #1460, opened 2017-04-07
RF: SSHRI should not provide string form with escaped characters

yarikoptic:
Should address #1459 

Also with this I am confessing that I have "downgraded" the version of git-annex-standalone as we ship from NeuroDebian (now it is at a revision right before it broke for us).  So travis should get all happy

I have uploaded the most recent broken version to neurodebian -devel so we could "play" with it


yarikoptic:
hm, buildbots seems to be building something but not really reporting back here... heh

yarikoptic:
woohoo -- tests pass again!  Since pass, minor changes and no comments -- will merge

Issue #1456, opened 2017-04-06
ssh interactions are broken with fresh annex 6.20170403+gitgdbe23d695-1~ndall+1


yarikoptic:
which I have managed to upload although I thought I have tested it locally first

I think (actually now certain since succeeds with `6.20170317+gitg8cd473c71-1~ndall+1` and fails with the `6.20170317+gitgfaecd73f3-1~ndall+1`) it is due to
```
commit faecd73f327bee5067998dd6d398033fb09a26a6
Author: Joey Hess <joeyh@joeyh.name>
Date:   Fri Mar 17 16:02:47 2017 -0400

    Support GIT_SSH and GIT_SSH_COMMAND
    
    They are handled close the same as they are by git. However, unlike git,
    git-annex sometimes needs to pass the -n parameter when using these.
    
    So, this has the potential for breaking some setup, and perhaps there ought
    to be a ANNEX_USE_GIT_SSH=1 needed to use these. But I'd rather avoid that
    if possible, so let's see if anyone complains.
    
    Almost all places where "ssh" was run have been changed to support the env
    vars. Anything still calling sshOptions does not support them. In
    particular, rsync special remotes don't. Seems that annex-rsync-transport
    already gives sufficient control there.
    
    (Fixed in passing: Remote.Helper.Ssh.toRepo used to extract
    remoteAnnexSshOptions and pass them to sshOptions, which was redundant
    since sshOptions also extracts those.)
    
    This commit was sponsored by Jeff Goeke-Smith on Patreon.
```
~and I guess we should just move from passing them via -c remote.*.annex-ssh-options~

if that is so, following the commit msg -- @joeyh know that you did break someone's workflow and I am complaining!   The issue is that we can't just use env variables to point to the socket paths (e.g. during sync), since those would be different for different remotes and multiple remotes might be communicated in a single git annex call.
@joeyh - please advise on how to proceed

joeyh:
I'm not clear how faecd73f327bee5067998dd6d398033fb09a26a6 broke
datalad. Is datalad setting GIT_SSH/GIT_SSH_COMMAND when it runs
git-annex?

-- 
see shy jo

yarikoptic:
yeap.  We set 'GIT_SSH_COMMAND' to `datalad sshrun` (if not already in the environment)

joeyh:
Yaroslav Halchenko wrote:
> yeap. We set 'GIT_SSH_COMMAND' to datalad sshrun (if not already in the
> environment)

[...]

PR #1461, opened 2017-04-07
RF+ENH+BF: attempts to fix up for recent git-annex changes

### LINKED ISSUE #1458: sshrun does not actually pass stdin? (or doesn't spitout received stdout?)
compare

```shell
$> echo 123 | ssh yoh@localhost cat 
123
$> echo 123 | datalad sshrun yoh@localhost cat 
$>
```

yarikoptic:
- [x] for the goodness of it -- report system version of ssh
- [x] `sshrun` now supports `-n`
- [x] more and improved (more consistent vars) logging of CMD execution. By default now would not add a new line with cwd to each run.  Use DATALAD_LOG_CMD_CWD variable
- [x] do not close SSHConnection if it wasn't opened by it
- [x] fixes #1458 in not passing stdin upon initial invocation of ssh command to some other check orwhatnot
- [x] yet to make it work with recent annex ;-)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ env:
     # will be used in the matrix, where neither other variable is used
     - BOTO_CONFIG=/tmp/nowhere
     - DATALAD_TESTS_SSH=1
+    - DATALAD_LOG_CMD_ENV=GIT_SSH_COMMAND
     - TESTS_TO_PERFORM=
     - NOSE_OPTS=-s
     # Special settings/helper for combined coverage from special remotes execution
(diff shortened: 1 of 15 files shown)


yarikoptic:
might be just the choice of travis build hosts or my changes made our execution slower so we started to hit the sealing again e.g. when logging too much (so suggests "improved logging of CMD execution" change although shouldn't be that sever

yarikoptic:
@bpoldrack in 1719faa I introduced at the level of the sshconnection "do not close if I wasn't the one which opened" logic for the socket (which somewhat compliments the logic in SSHManager which is solely based on the presence of the socket files).  And in e29def9 I presumably fixed up the tests demanding ssh manager then to actually close the connections it created since otherwise once test started to fail which expected that connection itself would close itself even if socket existed and was active before (as created by the other tests).  Could you please either verify that my logic is correct or state why it is not. Cheers!

yarikoptic:
not sure yet if problems with recent annex (#1456) aren't also due to annex, so not sure if should be handled as part of this PR.  @datalad/developers -- would appreciate review, improvements and merge

yarikoptic:
oh it is so green and want to be merged... if noone cares to do so -- I will

PR #1468, opened 2017-04-19
Last mile, hopefully.

### LINKED ISSUE #1110: Confusing communication and inconsequent, and undesired behavior of `install`

### LINKED ISSUE #1173: RF install to use `add` for registering install submodules

### LINKED ISSUE #1469: `get` fails on unavailable subdataset with incomprehensible message

mih:
Some discussion on what is happening here can be found in #1409 and #1470 

TODO:

- [x] Various fixes by @yarikoptic 
- [x] `install` is a new-style generator command now
- [x] fixes #1469 
- [x] fixes #1173
- [x] fixes #1110
- [x] some work towards #875 
- [x] `install` no longer uses any low-level helpers directly, instead it sits rather cleanly on top of `clone` and `get`, and performs argument inspection and switching logic only. It seems that this is now at a stage where we could think about adding more convenience functionality to `install` without much pain (e.g. #1347, #1146)
- [x] code cleanup: unreachable parts in `install`, plus a duplicate (and buggy) implementation of a `clone` helper removed
- [x] Remove support for old-style "if_dirty" handling from `install` #1175
- [x] helpers unique to `get` now consolidated in `get.py`
- [x] Extend result filtering API to be able to pass a command's `kwargs`
- [x] Attempt to implement result filters for input argument based filtering in `install`
- [ ] Figure out what kind of result rendering we want #1476
- [ ] Possibly disable "implicit" result report for any command by default (at least via the cmdline API) -- this is more a thought than a plan for now and likely out of scope for this PR
- [x] must have fix for #1474
- [x] would be nice to get #1475 sorted out
- [x] make `get` report on things that annex is silent about (whole dirs, files in Git, ...)


mih:
@yarikoptic @bpoldrack Remaining test failures are due to #1471 (our PY2 test setup is no longer valid), and due to NITRC being down for the second day in a row now.

mih:
Pending any issues that may pop up from the last few commits the basic machinery is in. It would be a good time for a review @yarikoptic @bpoldrack The only thing missing is the particular kind of result rendering we want to see from `install`, but I consider this secondary to a review of the general approach.

yarikoptic:
while I thought I could easily resolve it, kinda got lost in the logic (with all the loggers passing and popping) so decided just to whine:  if install fails to install something it seems to complain only about the first failed item, twice in logging at WARNING level and then also with ERROR and does not render result for successfully installed item:
```
$> datalad install missing1 missing2 ///labs/haxby/raiders          
[INFO   ] Cloning dataset from 'http://datasets.datalad.org/labs/haxby/raiders' to '/tmp/raiders' 
[WARNING] path does not exist: /tmp/missing1 
[WARNING] path does not exist: /tmp/missing1 
[ERROR  ] could not perform all requested actions: Command did not complete successfully [{'action': 'get', 'path': '/tmp/missing1', 'status': 'impossible', 'message': ('path does not exist: %s', '/tmp/missing1')}] [utils.py:generator_func:1061] 
```
actually in this case didn't even report the successful installation of one of the datasets (the raiders one).  Expected:
- (optionally) warning about each failed installation
- rendering of results for each accomplished (failed or not) installation

yarikoptic:
while trying it in Python I saw that
- (PY1) `install.__doc__` (as `install?` in ipython says) mentions `on_failure` but lacks it from signature (so have no idea on default wihch seems to be 'continue')
- (PY2) if I specify `on_failure='ignore'` return value does not include failed ones... Expected: may be I had misunderstanding but I thought the point was to have those failed ones also in the result in such cases (as None's)
- (PY3) if provided a list of urls (with a single one) a non-list single Dataset instance is returned.  moreover if a list is provided, where only 1 succeeded -- that succeeded one is returned (without being wrapped in the list). Expected: type (list or a single one) matching the input (list of things if list is provided, and a single one if a singe path/url).  Similar behavior is done already within `@normalize_path` so GitRepo and AnnexRepo where applicable have that behavior. Otherwise -- it would be difficult to handle it programatically since output type would be changing
- (PY4) it seems that order of items in the return list doesn't match order of input items -- again would complicate (and/or confuse) usage programmatically
```ipython
In [31]: r = install(['///labs/haxby/raiders', 'failed'], on_failure='skip', result_xfm=None)
[WARNING] path does not exist: /tmp/failed 
[WARNING] path does not exist: /tmp/failed 

In [32]: print r
[{'status': 'notneeded', 'source_url': '///labs/haxby/raiders', 'action': 'clone', 'path': '/tmp/raiders', 'type': 'dataset', 'message': ("dataset %s was already cloned from '%s'", <Dataset path=/tmp/raiders>, 'http://datasets.datalad.org/labs/haxby/raiders')}, {'action': 'get', 'path': '/tmp/failed', 'status': 'impossible', 'message': ('path does not exist: %s', '/tmp/failed')}]

In [34]: r = install(['failed', '///labs/haxby/raiders'], on_failure='skip', result_xfm=None)
[WARNING] path does not exist: /tmp/failed 
[WARNING] path does not exist: /tmp/failed 

In [35]: print r
[{'status': 'notneeded', 'source_url': '///labs/haxby/raiders', 'action': 'clone', 'path': '/tmp/raiders', 'type': 'dataset', 'message': ("dataset %s was already cloned from '%s'", <Dataset path=/tmp/raiders>, 'http://datasets.datalad.org/labs/haxby/raiders')}, {'action': 'get', 'path': '/tmp/failed', 'status': 'impossible', 'message': ('path does not exist: %s', '/tmp/failed')}]
```

datalad/distribution/install.py
```diff
@@ -178,106 +191,123 @@ def __call__(
...
-                    if ds_.is_installed():
-                        installed_items.append(ds_)
-                    else:
-                        lgr.warning("%s was not installed", ds_)
-
-            return Install._handle_and_return_installed_items(
-                ds, installed_items, failed_items, save)
+                for r in Get.__call__(
+                        to_get,
+                        # TODO should pass-through description, not sure why disabled
```
bpoldrack: I think it wasn't actually disabled, but outcommented as a reminder to implement that option. And the destiny of a reminder is to become a forgotten artifact. ;-)

datalad/distribution/install.py
```diff
@@ -178,106 +191,123 @@ def __call__(
...
             # obviously pointless input combination
-            raise ValueError(
-                "installation `source` and destination `path` are identical. "
-                "If you are trying to add a subdataset simply use `save` %s".format(
-                    path))
+            yield get_status_dict(
+                'install', path=path, status='impossible', logger=lgr,
+                source_url=source, refds=refds_path,
+                message="installation `source` and destination `path` are identical. "
+                "If you are trying to add a subdataset simply the `add` command")
```
bpoldrack: Missing word in the message: "use".

bpoldrack:
Generally: LGTM. I like the simplification of that $/&%§/ command. ;-)

Re (PY1): The common command options are not in the signature, yes. We can't change that. But there might be a way to figure out the default for the command and put it in the docstring too.

Re (PY3): This probably is a general issue. May be we could change the behavior of `item-or-list` to not only depend on the number of items in the result but also in the input? Or we simply have another one `input-type`?

mih:
Thanks for the reviews!

PY2: Indeed, that should be changed. Thanks!

PY3: that is a choice. You can set `return_type='list'` and have your suggested behavior. When I tally the use cases in our test, the install-a-single-dataset case if by far the most frequent one, hence I went for this default.

PY4: that is simply maintained previous behavior,. My own dislike is documented in #1470 with reasons, and two possible way to approach it.

PY5: I consider this to be a bug -- no idea yet why...

datalad/support/gitrepo.py
```diff
@@ -1051,7 +1055,7 @@ def get_active_branch(self):
         except TypeError as e:
             if "HEAD is a detached symbolic reference" in str(e):
                 lgr.debug("detached HEAD in {0}".format(self))
-                return None, None
+                return None
```
yarikoptic: I believe that this double `None` was done on purpose to distinguish case of having no branch (e.g. new repo, no commit yet) from having a detached branch 

remainders:
```
datalad/distribution/get.py-    ds.repo.update_submodule(sm_path, init=True)
datalad/distribution/get.py-    updated_branch = subrepo.get_active_branch()
datalad/distribution/get.py:    if branch and (not updated_branch or updated_branch == (None, None)):
--
datalad/distribution/update.py-        # handle merge in plain git
datalad/distribution/update.py-        active_branch = repo.get_active_branch()
datalad/distribution/update.py:        if active_branch == (None, None):
```

yarikoptic:
To facilitate testing etc, let's proceed by merging ;)

mih:
Uh, that came quickly! Thanks!

PR #1477, opened 2017-04-24
ENH+BF: Variety of little fixes and enhancement accumulated during recent recrawling of openfmri

yarikoptic:


mih:
Didn't inspect the changes in detail, but it doesn't seem to conflict which the other things that are going on. So +1 from me.

datalad/distribution/publish.py
```diff
@@ -75,33 +126,59 @@ def _publish_dataset(ds, remote, refspec, paths, annex_copy_options):
...
+    # if not annex_ignore:
+    #     if annex_uuid is None:
+    #         # most probably not yet 'known' and might require some annex
+    knew_remote_uuid = None
+    if isinstance(ds.repo, AnnexRepo):
+        try:
+            ds.repo.get_wanted(remote)  # could be just checking config.remote.uuid
+            knew_remote_uuid = True
+        except CommandError:
+            knew_remote_uuid = False
```
yarikoptic: this is the point I want to get improved, but not sure if would get time any time soon.... so probably will merge as is for now

PR #1478, opened 2017-04-24
BF: to get master testable again

yarikoptic:
- [x] disable `line-profiler` in setup.py to avoid pulling ipython 6.x not compatible with python < 3.4


yarikoptic:
some travis runs  (e.g. among https://travis-ci.org/datalad/datalad/builds/225419887) failed due to #1479

PR #1485, opened 2017-05-03
Return values: last half mile

mih:
TODO:

- [x] Give summary according to #1476
- [x] converted `add` to a new-style command
- [ ] figure out what to do with the `ValueError` thrown by `Interface._prep()` if non-dataset paths are given. This should rather turn into an "error" result to get a result for each input argument.
- [x] make `add` talk about all non-actions like `get`. For example: currently a `datalad add .` in a clean dataset gives no result whatsoever
- [x] be able to `add` and `save` things that are already staged #1413

PR #1498, opened 2017-05-07
This is PR #1485 (just going into master this time)

mih:


mih:
Screw it. Will open yet another one with all changes.

Issue #1493, opened 2017-05-05
uninstall -r fails if pure git repo is in the path


yarikoptic:
```
$> datalad uninstall -r labs
uninstall(ok): /tmp/datasets.datalad.org/labs/haxby/raiders (dataset)
[WARNING] no annex in dataset
action summary:
  drop (impossible: 1)
  uninstall (ok: 1)
[ERROR  ] could not perform all requested actions: Command did not complete successfully [{'action': 'drop', 'path': '.', 'status': 'impossible', 'message': 'no annex in dataset'}] [utils.py:generator_func:1082]
```
expected -- shouldn't try to drop anything where there is no annex. datalad 0.5.1.dev288 (michael's rf-returnvalues)

current master seems to be ok:
```
$> datalad uninstall -r labs
[INFO   ] skip dropping files in <Dataset path=/tmp/datasets.datalad.org/labs/haxby>, no annex
[INFO   ] skip dropping files in <Dataset path=/tmp/datasets.datalad.org/labs>, no annex
2 items were uninstalled:
Dataset: /tmp/datasets.datalad.org/labs/haxby
Dataset: /tmp/datasets.datalad.org/labs

```

mih:
Agreed, bug.

mih:
Option `-r` is not necessary to produce the error.

Issue #1492, opened 2017-05-05
we lost ability to install uninstalled/uninstalledsub/sub


yarikoptic:
With 0.4.1 we still could (0.5.0 already couldn't):
```
(git)smaug:/tmp/datasets.datalad.org[master]git
$> datalad install labs/haxby/raiders/
[INFO   ] Installing subdataset <Dataset path=/tmp/datasets.datalad.org/labs> in order to get /tmp/datasets.datalad.org/labs/haxby/raiders
[INFO   ] Installing subdataset <Dataset path=/tmp/datasets.datalad.org/labs/haxby> in order to get /tmp/datasets.datalad.org/labs/haxby/raiders
[INFO   ] Installing subdataset <Dataset path=/tmp/datasets.datalad.org/labs/haxby/raiders>
1 installed item is available at
<Dataset path=/tmp/datasets.datalad.org/labs/haxby/raiders>
```
current bleeding edge:
```
$> datalad install labs/haxby/raiders/
[WARNING] path does not exist: /tmp/datasets.datalad.org/labs/haxby/raiders
[ERROR  ] could not perform all requested actions: Command did not complete successfully [{'action': 'get', 'path': '/tmp/datasets.datalad.org/labs/haxby/raiders', 'status': 'impossible', 'message': ('path does not exist: %s', '/tmp/datasets.datalad.org/labs/haxby/raiders')}] [utils.py:generator_func:1082]
```

yarikoptic:
moreover it was working correctly even while requesting datafiles -- was so lovely:
```shell
(git)smaug:/tmp/datasets.datalad.org[master]git
$> datalad install -g labs/haxby/raiders/sub001/anatomy
[INFO   ] Installing subdataset <Dataset path=/tmp/datasets.datalad.org/labs> in order to get /tmp/datasets.datalad.org/labs/haxby/raiders/sub001/anatomy
[INFO   ] Installing subdataset <Dataset path=/tmp/datasets.datalad.org/labs/haxby> in order to get /tmp/datasets.datalad.org/labs/haxby/raiders/sub001/anatomy
[INFO   ] Installing subdataset <Dataset path=/tmp/datasets.datalad.org/labs/haxby/raiders> in order to get /tmp/datasets.datalad.org/labs/haxby/raiders/sub001/anatomy
[INFO   ] Getting 1 items of dataset <Dataset path=/tmp/datasets.datalad.org/labs/haxby/raiders> ...
[INFO   ] Actually getting 14 files
[...]

Issue #1476, opened 2017-04-24
Common logging and result report concept


mih:
The aim is to agree on one and then to write it up in here.

### Q1: When to just log vs yield a result

[this is about command implementations; the situation in low-level code is different. I'd argue that low-level code should only ever trigger debug messages (maybe distinguish different levels). It generally cannot decide or know whether a low-level error is actually a problem. We see this in numerous scary messages that are not that meaningful for a user]

Results can always easily be logged by including a logger in the result dict. Pure logging (without yielding a result) should only happen for progress info or decision making that may not lead to any result, e.g. a download attempt. Any result-oriented reporting (e.g. "i couldn't do what you requested", or "I got this") should happen in the form of a yielded result.

### Q2: Reporting individual results vs summaries?

Reporting summarizes comes with the need to aggregate across mutliple results. This aggregation practically prevents the ability to immediately yield intermediate results (a function that yields can only return `None`).  Summary rendering should focus on cmdline use (where we relay result dicts to the top-level anyways, and do not need to rely on custom return values (like in the Python API).

Command summaries might be highly custom, but maybe there is a meaningful default summary implementation. Something like

````
Performed actions:
  install (ok: 200, notneeded: 3, error: 1)
  get (impossible: 12876)
````

### Q3: Default format of reporting individual results?

It should be uniform across all commands. Proposal:

````
% datalad --output-format simple install -s ~/forrest/collection/phase2
install(notneeded): /tmp/phase2 [dataset <Dataset path=/tmp/phase2> was already cloned from '/home/mih/forrest/collection/phase2']
````

more formal: <**action label**>(<**status label**>): <**result path**> [**result annotation/message/note**]

This is semi-structured and at the same time somewhat human readable.

### Q4: What individual results to report?

Some commands (install/get) could make use of the principle "ignore implicit results by default" in order to avoid overwhelming the user. We have to keep in mind that yielded results are logged if desired, but that doesn't imply that they have to be rendered as a result at the same time. In general, it seems difficult to come up with a common principle to decide what to render and what not.

mih:
@yarikoptic @bpoldrack Please have a look!

yarikoptic:
Sounds reasonable BUT so far the most unpleasant experience with datalad for me is not when it reports too much by default, and not even some wires it might report but when it does something for a while (minutes) without saying anything..  oops plane is taxing.. Going offline

mih:
FTR: With current improvements in #1468 we get this:

````
[...]

Issue #1413, opened 2017-03-22
datalad add does not save files if they were staged before


yarikoptic:
have a situation (0.5.0)
```shell
hopa:/tmp/test2
$> git st
On branch master
Changes to be committed:
  (use "git reset HEAD <file>..." to unstage)

	new file:   123.txt
	new file:   125.txt

$> datalad add *
Nothing was added (consider --recursive if that is unexpected)
```
so it seems that whatever analysis it did, wasn't really complete as to decide either save (well -- commit) was necessary

yarikoptic:
seems to be just that wrong in general
```shell
$> ls -lta
total 4
drwx------  9 yoh  yoh   300 Mar 22 09:52 .git/
drwx------  4 yoh  yoh   100 Mar 22 09:52 ./
drwxrwxrwt 91 root root 7200 Mar 22 09:52 ../
-rw-------  1 yoh  yoh    42 Mar 22 09:36 .gitattributes
drwx------  2 yoh  yoh    80 Mar 22 09:35 .datalad/
(dev) 2 14305 [2].....................................:Wed 22 Mar 2017 09:52:24 AM EDT:.
hopa:/tmp/test2
$> cat .gitattributes; echo 128 > 129.txt; datalad add 129.txt; cp ~/picts/l_hires.jpg .; datalad add l_hires.jpg; datalad add *      
* annex.largefiles=(not(mimetype=text/*))  
Added /tmp/test2/129.txt                                                                                                                                
Added /tmp/test2/l_hires.jpg                                                                                                                            
Nothing was added (consider --recursive if that is unexpected)
```

bpoldrack:
Well, I think that was (partly) intentional, since we wanted to commit only things, that we have added.
And since there was nothing to add (to stage), we don't commit anything.
In such a situation `save` is what you want, I guess.

yarikoptic:
yeah... but in this case I am also instructing explicitly to add/commit those files. For me (user) it doesn't matter if they were staged or not before.  Similar to `git commit FILES` -- it would commit those files regardless if they were staged or not before

mih:
Looking into it. ATM still the case in #1485 -- but I like that the new default output is more informative already. 

````
mih@meiner /tmp % datalad create addtest
[INFO   ] Creating a new annex repo at /tmp/addtest 
create(ok): /tmp/addtest (dataset)                                                                                    
mih@meiner /tmp % cd addtest
mih@meiner /tmp/addtest (git)-[master] % ls
[...]

PR #1499, opened 2017-05-07
PR#1485 reborn against master, plus more fixes

mih:
- [x] Give summary according to #1476
- [x] converted `add` to a new-style command
- [ ] figure out what to do with the `ValueError` thrown by `Interface._prep()` if non-dataset paths are given. This should rather turn into an "error" result to get a result for each input argument.
- [x] make `add` talk about all non-actions like `get`. For example: currently a `datalad add .` in a clean dataset gives no result whatsoever
- [x] be able to `add` and `save` things that are already staged #1413 
- [x] calling `drop` in pure-git repo during recursive uninstall is no longer an issue #1493 
- [x] can install `uninstalled/uninstalledsub/sub` again #1492

### DIFF
--- a/datalad/api.py
+++ b/datalad/api.py
@@ -77,7 +77,7 @@ def _generate_func_api():
             # TODO: BEGIN to be removed, when @build_doc is applied everywhere
             spec = getattr(intf, '_params_', dict())
             api_name = get_api_name(intfspec)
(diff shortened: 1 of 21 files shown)


mih:
@yarikoptic linux buildbots seems to be broken (all timeout due to inactivity)

mih:
Travis is happy, OSX buildbot is happy. Given that master still fails and this has fixes, I'll merge now to see where the remaining issues lie.

PR #1509, opened 2017-05-10
Some speed-up for get_paths_by_dataset

### LINKED ISSUE #1505: `get_paths_by_dataset()` is MAJOR source of slowdown!
@bpoldrack @yarikoptic This is the climax of the OPT saga of the past few days:

Let me introduce you to: `get_paths_by_dataset()`, a function that runs every time a path is passed into almost any top-level command. Now watch this (with our big collection):

````
% time DATALAD_LOG_LEVEL=debug \
  python -c "import datalad.api; from datalad.interface.utils import get_paths_by_dataset; get_paths_by_dataset(['$PWD'], recursive=True)"
<snip>
DATALAD_LOG_LEVEL=debug python -c   1,31s user 1,10s system 18% cpu 12,744 total

# and now for fun and games, we give the same path twice
mih@meiner /tmp/datasets.datalad.org (git)-[master] % time DATALAD_LOG_LEVEL=debug \
  python -c "import datalad.api; from datalad.interface.utils import get_paths_by_dataset; get_paths_by_dataset(['$PWD', '$PWD'], recursive=True)"
<snip>
DATALAD_LOG_LEVEL=debug python -c   2,14s user 2,09s system 16% cpu 25,093 total
````

So this is not only slow, it is also stupid. I can say that, I wrote it.

I see no way how to make the individual pieces faster. If we ask for the full list of datasets in a tree, we have to get them at some point (which does not mean we should carefully evaluate when to ask for such a list). However, the logic purely within that function could be improved. In particular wrt the wasteful rediscovery when multiple input paths are given (even beyond a plain `unique(paths)`).

I won't have time to work on this now, but if you ever feel something is slow, consider throwing some computer science in this approximate direction first!

### PR DESCRIPTION
**Pull request by bpoldrack**
Closes #1505 

In the light of #1506 it probably doesn't make sense to spend more time on this right now. Nevertheless some obvious waste of time is removed. Basically this reduces the calls to `subdatasets`. There is one little change in semantics: Before paths, that were in the input several times, were reported several times in the output. Since the output is a dict with the discovered datasets as key, this doesn't make sense to me. Changing this behavior seems to not break anything, so ...

Marked as WIP, while waiting for the tests ...

### DIFF
--- a/datalad/interface/utils.py
+++ b/datalad/interface/utils.py
@@ -510,7 +510,7 @@ def get_paths_by_dataset(paths, recursive=False, recursion_limit=None,
     # paths that don't exist (yet)
     unavailable_paths = []
     nondataset_paths = []
-    for path in paths:
+    for path in unique(paths):
         if not lexists(path):
             # not there yet, impossible to say which ds it will actually
             # be in, if any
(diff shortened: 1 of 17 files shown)

### CODE REVIEW

File: datalad/interface/utils.py [hunk 1]
```diff
@@ -525,24 +525,53 @@ def get_paths_by_dataset(paths, recursive=False, recursion_limit=None,
...
+            else:
+                # we figured out the dataset previously, so we can spare some
+                # effort by not calling ds.subdatasets or
+                # ds.get_containing_subdataset. Instead we just need
+                # get_dataset_root, which is cheaper
+                if dspath != get_dataset_root(d):
+                    # if the looked up path isn't the default value,
+                    # it's a 'fixed' entry for an unavailable dataset (see above)
+                    unavailable_paths.append(path)
+                    continue
```
yarikoptic: note that above logic is not covered by tests and tests got one consistent failure already
NB didn't try to follow the logic yet... would prefer it first to be fixed/tested

bpoldrack: Actually it is tested. That's why it failed. Fix pushed.

mih:
Now conflicts. Could be resolved, or we wait for the PR (coming today) that includes this fix and takes it to the next level...

bpoldrack:
Closing in favor for the PR coming today, which includes this fix and takes it to the next level.

yarikoptic:
so which PR came to replace this one @bpoldrack ?

mih:
#1506. But speed is coming after correctness.

PR #1515, opened 2017-05-11
WIP: TST: What's going on in the sym-link build?

bpoldrack:


mih:
I think this can be closed now.

Issue #1389, opened 2017-03-19
save subdspaths  does nothing


yarikoptic:
```shell
(venv-tests)3 10596.....................................:Sun 19 Mar 2017 11:17:47 AM EDT:.
(git)smaug:/mnt/datasets/datalad/crawl/openfmri[master]git
$> git diff ds000017[AB]
diff --git a/ds000017A b/ds000017A
index caec2d1..c69868f 160000
--- a/ds000017A
+++ b/ds000017A
@@ -1 +1 @@
-Subproject commit caec2d1df5e8ddabcee7916871880f8bf3889b0f
+Subproject commit c69868f434390bbde4422344e18a311bff85e924
diff --git a/ds000017B b/ds000017B
index 64f9195..76551c8 160000
--- a/ds000017B
+++ b/ds000017B
@@ -1 +1 @@
-Subproject commit 64f9195de0005f5aa511988fa19aabdbc37c8979
+Subproject commit 76551c816fbd13c0a9277f60a1791620de67c632
changes on filesystem:                                                                                                           
 ds000017A | 2 +-
 ds000017B | 2 +-
(venv-tests)3 10597.....................................:Sun 19 Mar 2017 11:17:51 AM EDT:.
(git)smaug:/mnt/datasets/datalad/crawl/openfmri[master]git
$> datalad save -m "updated A and B datasets for 17 for their final version - superseeeded by pure 017" ds000017[AB]
changes on filesystem:                                                                                                           
 ds000017A | 2 +-
 ds000017B | 2 +-
```

so those two sub-datasets progressed forward, but if I specify them to save command -- it just does nothing.  Have to use `-u` option to save their adjusted state

yarikoptic:
can anyone of save experts look at it @datalad/developers ?  keep hitting it

mih:
Fix in #1506

Issue #1217, opened 2017-01-25
Provide `diff` command


mih:
We need `diff`. We currently use GitPython's diff, but it is kinda heavy and not flexible enough (e.g. direct mode submodules). We also need `status`, but `status` is pretty much a diff (ignoring the patch generation) of the workdir against HEAD. Ideally, we would like to have enough flexibility to also be able to do as little as `ls` (with recursion).

Plan (based on #1506):

- [x] wrap `git diff --raw SOMETHING`, where SOMETHING is anything that Git supports
- [x] provide results via a generator that yields annotated paths (`git diff` readily informs us about the pre/post nature of any path...)
- [x] support recursion in the standard fashion
- [ ] re-use `AnnexRepo.get_status()` tests

mih:
Any plans anyone to make this happen? @yarikoptic @bpoldrack

yarikoptic:
Fwiw Myself I just use gitpython's diff in the crawler... What I at times feel need for is diff which works also for annexed files content with some configurable diff per type of file... But that is our of scope of this over I guess.

mih:
I came up with a basic wrapper around `git diff` that yields annotated paths. Will turn into a proper command next. PR will follow.

mih:
From #1506:

````
mih@meiner /tmp/datasets.datalad.org (git)-[master] % datalad --on-failure continue diff --revision HEAD~2..HEAD~1 -r
modified: .datalad/meta/crcns/meta.json(file)
[WARNING] fatal: bad revision 'HEAD~2..HEAD~1' [diff(/tmp/datasets.datalad.org/TEST2)] 
modified: crcns/aa-2(dataset)
modified: crcns/aa-1/.datalad/meta/meta.json(file)
modified: crcns/aa-2/.datalad/meta/meta.json(file)
added: crcns/aa-3/.datalad/meta/meta.json(file)
modified: crcns/ac-1/.datalad/meta/meta.json(file)
modified: crcns/ac-2/.datalad/meta/meta.json(file)
<snip>
added: crcns/hc-2/ec016.582/ec016.582.dat.tar.gz-b(file)
added: crcns/hc-2/ec016.582/ec016.582.dat.tar.gz-c(file)
added: crcns/hc-2/ec016.582/ec016.582.mpg.tar.gz(file)
added: crcns/hc-2/ec016.582/ec016.582.tar.gz(file)
modified: crcns/hc-3/.datalad/meta/meta.json(file)
[...]

PR #1506, opened 2017-05-12
Uniform, efficient and explicit input path handling

### LINKED ISSUE #1518: `ds.create(path)` must refuse to create at `path` if `ds` knows about a subdataset at this location already
See hangout debugging.

mih:
Things got better, but we are still suffering from some fuzzy semantics and scattered processing of input paths. This is an attempt to change that, and with it a number of related issues.

The principle goal is to concentrate any path handling in a single command.

- [x] `Interface._prep()` RF to `annotate_paths` plumbing command (generator) and centralize the reporting logic for `unavailable_paths` in there. Except for `get` and `remove` all other commands could delegate reporting to it. Would compact client code quite a bit. `get` and `remove` could fish these out of the yielded results quite easily.
- [x] add another simple helper that can turn validated results into an oldschool `content_by_ds` dictionary to ease phased RF
- [x] reconsider exception/error handling #1200, but also see next
- [x]  figure out what to do with the ValueError thrown by Interface._prep() if non-dataset paths are given. This should rather turn into an "error" result to get a result for each input argument.
- [x] possible OPT for `get_paths_by_dataset()` #1505 
- [x] explicitly identify paths that were given as input, in contrast to paths that were discovered (e.g. due to recursion)
- [x] explicitly distinguish between paths pointing to a dataset vs. paths pointing into a dataset. For example, `ds.add(subds)` could mean add `subds` as a subdataset to `ds`, but it could also mean add everything in `subds` to `subds` and then add `subds` as a subdataset to `ds`. So far this is decided by secret negotiations in the back alley. This may require a change to our approach to path normalization (`get_normalized_path_arguments()`), which is presently done upfront, and therefore prevents us from using semantic markup like `subdir/.` or `subdir/` (like rsync) to indicate that we are talking about the content of a dir, rather than the dir itself.
- [x] possible further OPT once actual use cases for `unavailable_paths` are analyzed (also see #1505, and below) 
- [x] avoid manual passed `path`s for log (formated) message, always annotate log automatically 
- [x] enable all commands that get input paths to also accepted "annotated" paths (dicts like our result dicts), so we can avoid rediscovering what something is (file, dir, dataset), and also avoid more expensive things like "what is the superdataset".
- [x] converted `create` to fix #1518 with no additional custom test
- [x] fixes: `ds.create(path)` must refuse to create at `path` if `ds` knows about a subdataset at this location already #1518
- [x] fixes: `get_paths_by_dataset()` is MAJOR source of slowdown! #1505 
- [x] fixes: save recursive saves untracked files in submodules #1419 (or anywhere)
- [x] fixes: save subdspaths does nothing #1389 
- [x] give `add` a little extra speed boost #1517 
- [x] `subdatasets` needs mode to query only a subset of the subdatasets #1529
- [x] fixes: remove refuses to remove #1530
- [x] rewrite `get` to avoid unecessary dataset hierarchy traversal
- [x] rewrite `clean` with path annotation
- [x] rewrite `drop` with path annotation
- [x] rewrite `uninstall` with path annotation
- [x] removed obsolete `handle_dirty_datasets()`
- [x] initial draft of a `diff` command #1217 
- [x] generic `--since` functionality #1202 (this still needs investigating what exactly it is supposed to do and how it should be used. in any case, path annotation supports seomthing like this now)
- [x] default error mode is 'continue' now
- [x] cmdline interface no longer vommits `IncompleteResultsError`
- optimize `annotate_paths()`


mih:
Pushed rebase on master

mih:
Coverage report is here: https://codecov.io/gh/datalad/datalad/compare/master...keep/diff

yarikoptic:
`datalad subdatasets -r` 

1. without `--bottomup` (as of 0.5.1-454-gbbed867f in this 'keep' branch) doesn't "generate" but just collects them all first and then spits out.  IMHO should be possible to implement in a generator fashion in both cases

2. with `--bottomup` seems to be not bottom uping really in some cases

```shell
$> datalad subdatasets -r --bottomup
subdataset(ok): crcns (dataset)     
subdataset(ok): dicoms (dataset)
subdataset(ok): indi (dataset)
subdataset(ok): kaggle (dataset)
subdataset(ok): labs/haxby/raiders (dataset)
subdataset(ok): labs/haxby (dataset)
subdataset(ok): labs/mitchell (dataset)
subdataset(ok): labs (dataset)
subdataset(ok): neurovault (dataset)
subdataset(ok): nidm (dataset)
subdataset(ok): openfmri/ds000001 (dataset)
subdataset(ok): openfmri/ds000002 (dataset)
subdataset(ok): openfmri/ds000003 (dataset)
...
```
here it is when just a subset of subdatasets is installed:
```shell
$> datalad ls -r | grep -v 'not installed'
.                   [git]  master  - 2016-11-08/15:04:15  OK
labs                [git]  master  - 2016-11-07/23:35:03  OK
labs/haxby          [git]  master  - 2016-10-11/00:01:37  OK
labs/haxby/raiders  [annex]  master  - 2016-09-27/10:32:34  OK
openfmri            [git]  master  - 2016-11-07/23:06:04  OK
openfmri/ds000105   [annex]  master  2.0.1-13-g6bb31fde 2016-11-03/17:03:51  OK
```

mih:
Yeah, bug. I know why. But note that this is not this PR, but master.

mih:
From my POV this is ready to merge, and should go in ASAP. Two test failures remain:

- doc test in crawler

````
Document: crawler/basics
------------------------
Exception occurred:
  File "/home/travis/virtualenv/python2.7.13/lib/python2.7/site-packages/sphinx/util/logging.py", line 331, in filter
    raise SphinxWarning(record.msg % record.args)
TypeError: not enough arguments for format string
The full traceback has been saved in /tmp/sphinx-err-pE3imU.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!
make: *** [doctest] Error 1
make: Leaving directory `/home/travis/build/datalad/datalad/docs'
The command "PYTHONPATH=$PWD $NOSE_WRAPPER make -C docs html doctest" exited with 2.
````

I am not able to reproduce this locally. @yarikoptic do you have any clue what might be happening? Locally I see this:

````
Document: crawler/basics
------------------------
1 items passed all tests:
   7 tests in default
7 tests in 1 items.
7 passed and 0 failed.
Test passed.

Document: generated/datalad.utils
---------------------------------
1 items passed all tests:
   5 tests in default
5 tests in 1 items.
5 passed and 0 failed.
Test passed.

Doctest summary
===============
   12 tests
    0 failures in tests
    0 failures in setup code
    0 failures in cleanup code
build succeeded.
````

- the other failure is limited to direct mode and @bpoldrack was working on it. Any ETA for a fix?

File: datalad/interface/diff.py [hunk 1]
```diff
@@ -0,0 +1,276 @@
...
+    'D': 'deleted',
+    'M': 'modified',
+    'R': 'renamed',
+    'T': 'typechange',
+    'U': 'unmerged',
+    'X': 'unknown_potentialbug',
+}
+
+
+def _translate_status(label, ap):
```
bpoldrack: DOC: What is `label`? Seems to be a dictionary, which is somewhat strange for a "label". What input is expected here?
`ap`? An annotated path, I guess?

datalad/interface/diff.py
```diff
@@ -0,0 +1,276 @@
...
+    if label[0] in ('C', 'R'):
+        ap['perc_similarity'] = float(label[1:])
+        label = label[0]
+    ap['state'] = state_map[label]
+
+
+def _translate_type(mode, ap, prop):
```
bpoldrack: Also DOC: What `mode` are we talking about? Where does it come from?

datalad/interface/diff.py
```diff
@@ -0,0 +1,276 @@
...
+    elif mode == stat.S_IFDIR | stat.S_IFLNK:
+        ap[prop] = 'dataset'
+    elif stat.S_ISDIR(mode):
+        # not sure if this can happen at all
+        ap[prop] = 'directory'
+    else:
+        ap[prop] = 'file'
+
+
+def _parse_git_diff(dspath, diff_thingie=None, paths=None,
```
bpoldrack: And DOC: Not sure, what a `diff_thingie` is supposed to be. Also: a `path` seems to be considered only if it is a raw input, right?


Generally, one can find out what those helpers are for, but it requires to go through the entire code of the helper itself and at least some usages of them to figure it out. This might be okayish for a complete review (still makes it harder), but within a week one needs to do it again to recall ... ;-)

datalad/interface/diff.py
```diff
@@ -0,0 +1,276 @@
...
+
+
+@build_doc
+class Diff(Interface):
+    """Report changes of dataset component between revisions.
+    """
+    # TODO describe properties that are reported
+
+    # make the custom renderer the default one, as the global default renderer
+    # does yield meaningful output for this command
```
bpoldrack: I assume this is meant to read "does NOT yield meaningful output"?

datalad/interface/diff.py
```diff
@@ -0,0 +1,276 @@
...
+            path=None,
+            dataset=None,
+            revision='HEAD',
+            staged=False,
+            ignore_subdatasets='none',
+            recursive=False,
+            recursion_limit=None):
+        if not dataset and not path:
+            # act on the whole dataset if nothing else was specified
+            dataset = curdir
```
bpoldrack: Not sure, whether I missed something recently - what happend to `require_dataset`?

datalad/interface/diff.py
mih: any label (see right above), it is not a dict

datalad/interface/diff.py
mih: From the only place this function is called... ;-)

datalad/interface/diff.py
mih: Can I avoid the documentation when I copy the function body inline? These are really just internal helper, in order to understand what they are doing and how exactly they can be used one really has to go through the whole code. Whatever I write in there will not be the whole truth, because I don't know it. In particular with `diff_thingie`: this goes as is to Git, it can be whatever Git can eat and Git will do with it whatever it feels like that day. I could copy the Git documentation, but it seems worthless as it will be outdated the next minute.

datalad/interface/diff.py
mih: Yes, will go away anyways -- this is already the case now.

datalad/interface/diff.py
mih: Good point, what happened to it? ;-)

datalad/interface/diff.py
bpoldrack: Well, so it's any label, git-diff can digest to figure out what to compare to? That's a reasonable one-liner, I think. ;-)

bpoldrack: Regarding "internal use only":
Yes, I get that I don't strongly insist. It's just, that it is often hard to RF or adapt commands, when it takes so much time to figure out, what certain blocks are supposed to do - whether thy are in helper functions or inline.

mih: re: `diff_thingie` -- see you agree now, it is a diff thingie

datalad/interface/diff.py
bpoldrack: I see. But then I don't get this line:
```python
ap['perc_similarity'] = float(label[1:])
```
If `label` is the one character key for `state_map`, what is `float(label[1:])` supposed to do?

mih: you can resist it, but in the end you have to read the `git diff` docs. According to that there are these to special cases.

bpoldrack: Actually, it says there three cases: `M` might have that score, too.

mih: Indeed! Nice catch, will fix it.

mih:
Look:

````
mih@meiner /tmp/datasets.datalad.org (git)-[master] % datalad annotate-paths -d . --modified-since HEAD^1..HEAD -r
annotate_path(): TEST2 (dataset)
annotate_path(): TEST2/mike2 (file)
````

I think I might rename the option to just `modified`.

mih:
Next I will try to RF `publish` to use the new generic `--modified` functionality. I am sure there is something that it does totally wrong right now. I would be very useful to have some real use cases from @yarikoptic that are currently broken and can be used to come up with a better way.

datalad/interface/diff.py
mih: Done

mih:
@yarikoptic all tests but the direct mode ones pass now. Could you please have another look @bpoldrack? Don't worry about performance, just correct behavior. Thanks

datalad/interface/diff.py
mih: Now I remember, `require_dataset` create a `Dataset` instance (not necessarily needed), and it also walks the path upward. This would lead to path annotations above any given request or the PWD -- not desired.

mih:
@yarikoptic buildbots seems to be defunc right now

yarikoptic:
there was some network awkwardness... I have restarted one -- seems not getting stuck.  so whenever you push again should work fine

mih:
@bpoldrack direct mode seems fine now, thanks!

mih:
@bpoldrack if you check the coverage report, you'll see that a bunch of code in annexrepo and gitrepo no longer runs anywhere in the tests. This will need a closer look re whether we should have dedicated tests for it, or whether we should RF it out. Cannot say from my POV.

mih:
Now working on RF'ing `publish` again, but do not hesitate to merge when ya feel like it...

bpoldrack:
Re coverage:
Most of uncovered lines seem to be special case or failure treatment, that isn't easy to test, but still makes sense from my point of view. I think we should look at it while fixing the *Repo classes for v6 and have dedicated build (like we have for direct mode). Only then we have the complete picture to decide whether to keep or where to move some of these parts.

bpoldrack:
All tests passing is a good point in time to merge the beast in my opinion.

PR #1521, opened 2017-05-14
TEMP: troubleshoot wtf with travis access to neurodebian

yarikoptic:


yarikoptic:
issue disappeared when we started to troubleshoot it

PR #1523, opened 2017-05-15
Attempt to have our tests run in stages by travis

mih:
Whole test batteries could fail in their entirety much earlier, and
as such free up test worker to run tests in other PRs. Currently,
we are waiting on each other quite a bit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -73,6 +73,7 @@ matrix:
     - https_proxy=
     - _DATALAD_NONPR_ONLY=1
   - python: 2.7
+    # not idea what the purpose of this one is, disable for now?
     env:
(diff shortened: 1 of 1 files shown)


mih:
Half-worked. I wanted to have the 'allowed_failure' tests run in a third stage, as we don't care too much about them -- yet they consume as much time is the other ones. So there is no point in running them when more critical tests already failed.

bpoldrack:
I don't quite get your second comment, @mih. According to the diff you didn't add any `stage` entry to the jobs in `allow_failure`. So what do you mean by "half-worked"?

mih:
I tried, but it is not valid to do so.

yarikoptic:
so is this effort to continue or RiP for now?

mih:
ok, screw it (for now)

PR #1526, opened 2017-05-18
BF: do not whine whenever removing a dataset which is not a subdataset

yarikoptic:
did
```
datalad install --reckless -s raiders raiders-clone
```
so it wasn't added to that superdataset. but upon remove benign message made no sense:
```shell
$> datalad remove --nocheck raiders-clone             
[ERROR  ] Found multiple subdatasets with registered path raiders-clone:
| []
| There should be only one. [uninstall.py:__call__:405] (AssertionError) 
```
so I just made that logic conditional on being a subdataset

### DIFF
--- a/datalad/distribution/create_sibling.py
+++ b/datalad/distribution/create_sibling.py
@@ -682,13 +682,13 @@ git update-server-info
 # DataLad
 #
 # (Re)generate meta-data for DataLad Web UI and possibly init new submodules
-dsdir={path}
-logfile=$dsdir/{WEB_META_LOG}/{log_filename}
+dsdir="{path}"
+logfile="$dsdir/{WEB_META_LOG}/{log_filename}"
 
 mkdir -p "$dsdir/{WEB_META_LOG}"  # assure logs directory exists
 
 ( which datalad > /dev/null \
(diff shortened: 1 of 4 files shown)


mih:
FYI: In #1506 I will make substantial modifications to `remove`, I will cherry-pick this commit and run from there.

mih:
Ah, no. I misread. This is actually going into 0.5. So ignore me.

mih:
Do you happen to have a test? Would be handy for making sure my RF didn't introduce anything like this.

yarikoptic:
fair enough -- I will add a test!

yarikoptic:
@mih -- added a basic smoke test -- should fail with unpatched code due to assertion error within it

yarikoptic:
one failure is due to those spaces -- CPed commit from master into this one (agaainst 0.5.x). The other one is from sphinx and I just hope it would go away, if not -- will check but it is unnlikely to be related to these changes

yarikoptic:
ok -- it has a test now, and should help resolve failures of other PRs (with "sym link" temp dir), so I will merge it for our common good

PR #1535, opened 2017-06-01
0.5.x changes (sphinx bugfix, crawling tune ups. etc) merged into master

yarikoptic:


yarikoptic:
failing travis is just the locking issue... I bet it might as well be gone with #1531 ;-)

Issue #1495, opened 2017-05-06
datalad publish --since is not in effect really (with -r)


yarikoptic:
I think it wasn't like that before at some point but I could be wrong.  
The idea of having `--since` was to minimize amount of work needed depending on what submodules actually got changed and how.  But I guess ATM publish having received -r is happily first queries for all present submodules only later making analysis that actually may be there is nothing to publish at all or it could be just a single submodule.   Actually not even sure if --since is in effect since now that everything is published I still see it going through submodules and trying to push some data where data might needed to be pushed:
```shell
(git)smaug:/mnt/btrfs/datasets/datalad/crawl[master]
$> datalad publish --to=datalad-public --missing=inherit -r
[INFO   ] Publishing <Dataset path=/mnt/btrfs/datasets/datalad/crawl/workshops/nipype-2017/ds000114> data to datalad-public 
...
$> datalad --version
datalad 0.5.1.dev95
...
```

yarikoptic:
it hurts so bad

mih:
In #1544 `publish` is now using the common modification detection from `annotate-paths`. Please check how you feel about that.

yarikoptic:
usecase:
- you have a collection of datasets which you published before
- by default to publish you need to publish only changes from the previous state
- so it used to look at the common commit with the remote, and take that point as the argument for --since

mih:
As we discussed and agreed upon during telco on June 9 2017: `since=''` should mean: check the diff between HEAD and the branch which a matching name on the remote, and only act upon that. Perform the modification detection on the base dataset, and use the result to filter out subdataset, in order to avoid having to process them all.

mih:
`since` logic is in place now with `create_sibling` in #1544

yarikoptic:
Yeap, that is what we agreed upon and then realized that for publish we needed some way to say "from the last commit point" how it used to be before when since was set to default None

mih:
This should work now, if not please give demo/test.

yarikoptic:
ok... got it now... so the default behaviour changed... if before, `since=None` was to trigger "since last publish", now I need to use `since=''`. Doing doc fix, and tune ups for your review

Issue #1480, opened 2017-04-25
publish should publish (annex copy) data first before pushing


yarikoptic:
the reason is that otherwise for datasets with data published, webui would not get triggered to crawl web meta-data for the web ui, and thus reporting wrong sizes.

Issue #1426, opened 2017-03-23
publish still "doesn't publish" (without -r)


yarikoptic:
e.g. I am in the top of our crawled mega-dataset, and trying to publish it after I have added workshops/ subdataset:

```shell
(git)smaug:/mnt/btrfs/datasets/datalad/crawl[master]git
$> datalad publish --to=datalad-public
[INFO   ] Publishing <Dataset path=/mnt/btrfs/datasets/datalad/crawl> to datalad-public
yoh@datasets.datalad.org's password:
Permission denied, please try again.
yoh@datasets.datalad.org's password:
No handlers could be found for logger "git.remote"
1 item was published:
Dataset: /mnt/btrfs/datasets/datalad/crawl


$> git lg
* 4ea5f63 - (HEAD -> master) [DATALAD] aggregated meta data (22 minutes ago) [Yaroslav Halchenko]
* c544eac - aggregated meta-data for the workshops which now includes nipype 2017 (23 minutes ago) [Yaroslav Halchenko]
* 2699b2f - [DATALAD] added content (31 minutes ago) [Yaroslav Halchenko]
* 9a87f45 - (datalad-public/master) [DATALAD] aggregated meta data (4 days ago) [Yaroslav Halchenko]
* d581162 - [DATALAD] auto-saved changes (4 days ago) [Yaroslav Halchenko]
```

so as you can see datalad-public/master is still  at older point

NB1 adding -r makes action to take long time, so I guess it does its "lets get entire hierarchy just in case" although there were no changes to other datasets but 'workshops' since the last push.
But it seems to do the right thing
```shell
$> datalad publish --to=datalad-public -r
[INFO   ] Publishing <Dataset path=/mnt/btrfs/datasets/datalad/crawl/workshops> to datalad-public
yoh@datasets.datalad.org's password:
Permission denied, please try again.
yoh@datasets.datalad.org's password:
[INFO   ] Publishing data of dataset <Dataset path=/mnt/btrfs/datasets/datalad/crawl/workshops/nipype-2017/ds000114> ...
[INFO   ] Publishing <Dataset path=/mnt/btrfs/datasets/datalad/crawl> to datalad-public
[INFO   ] Publishing data of dataset <Dataset path=/mnt/btrfs/datasets/datalad/crawl/dicoms/dartmouth-phantoms/Philips-20131217> ...
[...]

PR #1544, opened 2017-06-07
RF `publish` in many ways

mih:
- [x] `publish` need to become a new-style command using path annotation
- [x] #1495 --since might be not in effect at all
- [x] #1480 bug: publish should publish (annex copy) data first before pushing
- [x] #1426 bug: doesn't publish
- [x] remove horrible `Dataset.get_containing_subdataset()`


mih:
Nice segfault in tests, awesome!

datalad/distribution/tests/test_publish.py
```diff
@@ -268,24 +293,22 @@ def test_publish_with_data(origin, src_path, dst_path, sub1_pub, sub2_pub, dst_c
     source.repo.add_remote("target", dst_path)
 
     # now, set up targets for the submodules:
-    sub1_target = GitRepo(sub1_pub, create=True)
+    # the need to be annexes, because we want to be able to copy data to them
+    # further down
+    sub1_target = AnnexRepo(sub1_pub, create=True)
```
yarikoptic: I think that was on purpose -- is should be ok to have superdataset to be a pure git, but subdatasets - annex. and publish the whole collection

mih: True, but this is about the subdataset.

mih:
The current state of this PR completes the conversion to new-style commands in that all old helpers could be removed. The few remaining commands are even older and never used the old-new helpers, or it does not make sense for them to use such code at all. Here is the list:

- `add-archive-content`
- `add-sibling`
- `aggregate-metadata`
- `crawl-init`
- `crawl`
- `create-sibling-github`
- `create-test-dataset`
- `download-url`
- `export`
- `ls`
- `move`
- `rewrite-urls`
- `search`
- `sshrun`
- `test`

yarikoptic:
example fails now @mih

yarikoptic:
so wrapt is hunting us down .... note that on travis it installs it from pypi the most recent 1.10.10 whenever in debian we have only 1.9.0 ... might be worth exploring the downgrade for a quick and dirty workaround...?

mih:
And it starts to fail from Debian 8 (with wrapt 1.8, works) to Debian 9
(with wrapt 1.9, broken).

yarikoptic:
probably picked up somewhere else among failures but here is what I get when testing within docker using python-dbg (chasing that gc issue... heh):
```
ERROR: datalad.distribution.tests.test_publish.test_invalid_call
    self.test(*self.arg)
  File "/tmp/datalad/datalad/tests/utils.py", line 762, in newfunc
    t(*(arg + (uri,)), **kw)
  File "/tmp/datalad/datalad/tests/utils.py", line 592, in newfunc
  File "/tmp/datalad/datalad/distribution/tests/test_publish.py", line 47, in test_invalid_call
    ds.uninstall('subm 1', check=False)
  File "/usr/local/lib/python2.7/dist-packages/wrapt/wrappers.py", line 561, in __call__
    args, kwargs)
  File "/tmp/datalad/datalad/distribution/dataset.py", line 403, in apply_func
    return f(**kwargs)
  File "/usr/local/lib/python2.7/dist-packages/wrapt/wrappers.py", line 522, in __call__
    args, kwargs)
  File "/tmp/datalad/datalad/interface/utils.py", line 511, in eval_func
  File "/usr/local/lib/python2.7/dist-packages/wrapt/wrappers.py", line 522, in __call__
    args, kwargs)
  File "/tmp/datalad/datalad/interface/utils.py", line 499, in return_func
    results = list(results)
  File "/tmp/datalad/datalad/interface/utils.py", line 490, in generator_func
IncompleteResultsError: Command did not complete successfully [{'status': 'error', 'orig_request': 'subm 1', 'registered_subds': True, 'raw_input': True, 'parentds': '/tmp/datalad_temp_testrepo_vY4iwT', 'action': 'uninstall', 'path': '/tmp/datalad_temp_testrepo_vY4iwT/subm 1', 'message': 'refusing to uninstall current or parent directory', 'type': 'dataset', 'refds': '/tmp/datalad_temp_testrepo_vY4iwT'}]

```

mih:
re uninstall error: a number of these were fixed in #1575

mih:
Published a rebase on current master. Now with fewer commits. Let's see how it performs. If it fails again, I will factor out a few more.

mih:
FTR: This failure is not specific to the test where the segfault usually occurs. When that one is disabled the next one fails: https://travis-ci.org/datalad/datalad/jobs/244968746

mih:
FTR: Moving the publish test to an ealier point in the test run does NOT change the failure pattern. That makes a general leakage as the cause more unlikely, and something publish-related more likely.

mih:
Rebased on dev-master.

PR #1545, opened 2017-06-07
WiP fix up some broken tests when ran in full

yarikoptic:
not yet sure what to do if anything about the nfsmount test(s) since they just time out these days :-/

before merge should remove TEMP commit


mih:
Ah, thanks a lot! The NFS tests are weired. I have the impression that they stall somehow. It is somewhat random when it happens, but it doesn't seem to be a general slow down. Am I wrong?

yarikoptic:
I think it is general slowdown -- there is no message that there were no output after X minutes of inactivity -- all processes are terminated around 49  min at random spots (as far as they manage to get)

mih:
If you remove the TEMP commit, this could be merged!

yarikoptic:
ok -- will remove TEMP commit and merge locally to 0.5.x and master and push directly to avoid all the travis roundtripping.

Issue #1546, opened 2017-06-08
datalad search blows hard


yarikoptic:
current master... was giving a demo :(

```shell
$> datalad search zebra
[ERROR  ] __call__() got an unexpected keyword argument 'result_renderer' [base.py:call_from_parser:352] (TypeError) 

$> datalad --version
datalad 0.5.1.dev541
```

mih:
It is missing im the exclusion list in Interface.base.  Simple fix. Cannot
do right now. On the road...

mih:
Managed to push #1547

PR #1547, opened 2017-06-08
BF: `search` is old-style (fixes gh-1546)

mih:


yarikoptic:
it took much longer on a "sym link" tmpdir run and timed out -- otherwise works for me locally -- THANKS!  and pity that apparently we don't have a test for it to catch :-/

PR #1551, opened 2017-06-09
RF: benchmarks -- shortened names and moved into separate modules

yarikoptic:
Now published already at http://datalad.github.io/datalad/ -- I reran them completely, so old results are gone, and should be easier to navigate.  running for more commits now


mih:
Sweet, thanks.

PR #1552, opened 2017-06-09
WiP NF XNAT/NITRC-IR support

yarikoptic:
TODOs

- [ ] come up with a saner default for file names/paths
- [ ] figure out if when deciding to either create a subdataset, first crawl it to see if there is any file to be crawled.  Some datasets (on central.xnat) seems to be just empty, so kinda waste
- [ ] aggregate with information about datasets which is exposed from regular NITRC (there is TODO comment on that)
  - [ ] per file meta-data (whenever we get there)
- [ ] make it drop files by default (nothing to extract and we might be short of space kinda)
- [ ] provide extractor for the meta-data
- [ ] publish NITRC-IR superdataset (public datasets for now)
- [ ] publish CENTRAL.xnat superdataset

in the future (probably after ultimate db is done) -- also crawl 
- [ ] HCP
- [ ] private ones from central and NITRC which do not have any private info in the file names/meta-data (@chaselgrove)

### DIFF
--- a/datalad/cmdline/main.py
+++ b/datalad/cmdline/main.py
@@ -343,7 +343,9 @@ def main(args=None):
                 # behave as if the command ran directly, importantly pass
                 # exit code as is
                 if exc.msg:
-                    os.write(2, exc.msg.encode() if isinstance(exc.msg, text_type) else exc.msg)
+                    os.write(
+                        2,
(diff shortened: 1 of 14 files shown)


yarikoptic:
moved to https://github.com/datalad/datalad-crawler/pull/9

Issue #1549, opened 2017-06-09
datalad install -r fails for studyforrest dataset -- asks for username (private?)


yarikoptic:
```shell
$> datalad install -r https://github.com/psychoinformatics-de/studyforrest-data
[INFO   ] Cloning dataset from 'https://github.com/psychoinformatics-de/studyforrest-data' to '/mnt/btrfs/datasets/datalad/crawl-misc/studyforrest/studyforrest-data' 
install(ok): /mnt/btrfs/datasets/datalad/crawl-misc/studyforrest/studyforrest-data (dataset)
[INFO   ] Installing <Dataset path=/mnt/btrfs/datasets/datalad/crawl-misc/studyforrest/studyforrest-data> recursively 
[INFO   ] Cloning dataset from 'https://github.com/psychoinformatics-de/studyforrest-data/aligned' to '/mnt/btrfs/datasets/datalad/crawl-misc/studyforrest/studyforrest-data/aligned' 
[ERROR  ] Failed to clone data from any candidate source URL: ['https://github.com/psychoinformatics-de/studyforrest-data/aligned', 'https://github.com/psychoinformatics-de/studyforrest-data/aligned/.git'] 
[INFO   ] Cloning dataset from 'https://github.com/psychoinformatics-de/studyforrest-data-aligned.git' to '/mnt/btrfs/datasets/datalad/crawl-misc/studyforrest/studyforrest-data/aligned' 
[INFO   ] Submodule HEAD got detached. Resetting branch master to point to 72b577b3. Original location was 47e53498 
install(ok): aligned (dataset) [Installed subdataset <Dataset path=/mnt/btrfs/datasets/datalad/crawl-misc/studyforrest/studyforrest-data/aligned>]
[INFO   ] Cloning dataset from 'https://github.com/psychoinformatics-de/studyforrest-data-aligned.git/src/phase1' to '/mnt/btrfs/datasets/datalad/crawl-misc/studyforrest/studyforrest-data/aligned/src/phase1' 
[ERROR  ] Failed to clone data from any candidate source URL: ['https://github.com/psychoinformatics-de/studyforrest-data-aligned.git/src/phase1', 'https://github.com/psychoinformatics-de/studyforrest-data-aligned.git/src/phase1/.git'] 
[...]

PR #1556, opened 2017-06-11
Enhance `subdatasets` command to set values in .gitmodules

mih:
While discussing #1549 we came up with the need to label subdatasets as 'no-auto-install' (or something like this). `subdatasets` already reports all values from `.gitmodules`, but we lack the ability to set them conveniently.

This PR adds this feature.

- [x] dissolve `rewrite_urls` when `subdatasets` get's the ability to digest value template strings.


datalad/distribution/subdatasets.py
```diff
@@ -257,10 +303,36 @@ def _get_submodules(dspath, fulfilled, recursive, recursion_limit,
...
+            submodule_section = 'submodule "{}"'.format(sm['gitmodule_name'])
+            # first deletions
+            for dprop in assure_list(delete_property):
+                parser.remove_option(submodule_section, dprop)
+                # also kick from the info we just read above
+                sm.pop('gitmodule_{}'.format(dprop), None)
+            # and now setting values
+            for sprop in assure_list(set_property):
+                prop, val = sprop
+                if '{' in val:
```
yarikoptic: Isn't { a legit character for a filename/path? I just wonder if we should just rely on some custom prefix to say that it is a template and should go through format

yarikoptic: Not to say that we might want to have some property which would contain { in its value and might not want to format it here

mih: Any suggestions for a prefix? I don't care, I just used the same mechanism that we use for --output-format.

mih: OK, added additional requirement for a template to be wrapped in `<>` and contain `{` before expansion is attempted`.

PR #1558, opened 2017-06-11
TST: Let NFS tests fail -- they just take too long

mih:
There is no information in them failing constantly, , while we know exactly why. Moreover, they make master look broken for no good reason.


mih:
I am merging this immediately, as it is only an internal issue and can be modified again at any time.

PR #1559, opened 2017-06-12
Various little improvements of `siblings`

### LINKED ISSUE #1484: rewrite-description
Our annex descriptions for the siblings/remotes aren't consistent, and I would like to change descriptions for all datasets which we host on `datasets.datalad.org` to be public URLs pointing to that website (not personal SSHRIs), so ppl could easily 'find' them.

we have `rewrite-urls` which adjusts .gitmodules to rewrite urls.  I wanted to adjust git annex descriptions for the annexes.  I see interface which takes
- `[--sibling NAME]`  by default would adjust of the local one, but can adjust for the remote ones as well
- `--dry-run` (ATM we have `--dryrun` in `create_sibling_github` and used to have `dry-run` for crawler, and that is inline with git) just to visualize
- `--recursive` (with other recursive options we have)
- `desc` which could be just a freetext or some kind of RI (SSHRI or URL) allowing for the same rewrites as `rewrite-urls` does

am I missing anything @mih and @bpoldrack ?  ideally should somehow be used/inline with default description create and create-sibling (at the moment none is custom specified, so annex creates).  so ideally should be some kind of option I guess somewhere they could `inherit`

mih:
Log:

- [x] tiny convenience bits
- [x] report annex description for all remotes with annexes and UUIDs
- [x] enable modification of annex descriptions (fixes #1484)
- [x] ability to query the local repo for infos under the label `here` (uses the same keys as for a proper annex remote):
````
mih@meiner ...tasets.datalad.org/labs/haxby/raiders (git)-[master] % datalad --output-format json_pp siblings -s here
{
  "action": "query-sibling",
  "annex-bare": "false",
  "annex-description": "mih@meiner:/tmp/datasets.datalad.org/labs/haxby/raiders",
  "annex-uuid": "7203baa3-f472-4bca-b886-753e2a9da1df",
  "annex-version": "5",
  "available_local_disk_space": 23273848256,
  "name": "HERE",
  "path": "/tmp/datasets.datalad.org/labs/haxby/raiders",
  "refds": "/tmp/datasets.datalad.org/labs/haxby/raiders",
  "status": "ok",
  "type": "sibling"
}
````

### DIFF
--- a/datalad/distribution/tests/test_siblings.py
+++ b/datalad/distribution/tests/test_siblings.py
@@ -10,7 +10,9 @@
 """
 
 from os.path import join as opj, basename
-from datalad.api import install, siblings
+from datalad.api import create
+from datalad.api import install
+from datalad.api import siblings
 from datalad.support.gitrepo import GitRepo
 from datalad.support.exceptions import InsufficientArgumentsError
 
@@ -173,3 +175,20 @@ def test_siblings(origin, repo_path):
             ok_(pushurl.endswith(basename(repo.path)))
         eq_(url, r['url'])
         eq_(pushurl, r['pushurl'])
+
+
+@with_tempfile(mkdir=True)
+def test_here(path):
+    # few smoke tests regarding the 'HERE' sibling
+    ds = create(path)
+    res = ds.siblings(
+        'query',
+        on_failure='ignore')
+    assert_status('ok', res)
+    assert_result_count(res, 1)
+    assert_result_count(res, 1, name='HERE')
+    here = res[0]
+    eq_(ds.repo.uuid, here['annex-uuid'])
+    assert_in('annex-description', here)
+    assert_in('annex-bare', here)
+    assert_in('available_local_disk_space', here)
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: datalad/distribution/tests/test_siblings.py [hunk 1]
```diff
@@ -173,3 +175,20 @@ def test_siblings(origin, repo_path):
...
+        'query',
+        on_failure='ignore')
+    assert_status('ok', res)
+    assert_result_count(res, 1)
+    assert_result_count(res, 1, name='HERE')
+    here = res[0]
+    eq_(ds.repo.uuid, here['annex-uuid'])
+    assert_in('annex-description', here)
+    assert_in('annex-bare', here)
+    assert_in('available_local_disk_space', here)
```
bpoldrack: May be one test, that uses 'HERE' as the name of the remote to query? Right now this tests only that it is used for output. But should be usable for input, too, right?

mih: This is coming later, working on being able to modify these things now.

mih: Done now for description setting.

PR #1567, opened 2017-06-12
Various fixes to get basic functionality running on windows

mih:
Biggest pieces is to make the global SSHManager optional.

Some proof:

image

### DIFF
--- a/datalad/__init__.py
+++ b/datalad/__init__.py
@@ -25,12 +25,16 @@ from .config import ConfigManager
 cfg = ConfigManager()
 
 from .log import lgr
-lgr.log(5, "Instantiating ssh manager")
-from .support.sshconnector import SSHManager
-ssh_manager = SSHManager()
-
 import atexit
-atexit.register(ssh_manager.close, allow_fail=False)
+from datalad.utils import on_windows
+if not on_windows:
+    lgr.log(5, "Instantiating ssh manager")
+    from .support.sshconnector import SSHManager
+    ssh_manager = SSHManager()
+    atexit.register(ssh_manager.close, allow_fail=False)
+else:
+    ssh_manager = None
+
 atexit.register(lgr.log, 5, "Exiting")
 
 from .version import __version__
--- a/datalad/support/network.py
+++ b/datalad/support/network.py
@@ -24,6 +24,7 @@ from collections import OrderedDict
 from os.path import abspath, isabs
 from os.path import join as opj
 from os.path import dirname
+from ntpath import splitdrive as win_splitdrive
 
 from six import string_types
 from six import iteritems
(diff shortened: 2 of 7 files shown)


mih:
Not so simple, is it....

yarikoptic:
is there an immediate demand to make our beast work on windows?

mih:
What does the grand proposal say? ;-)

mih:
NFS failure due to known issue unrelated to these changes.

mih:
Before this is merged, it need manual confirmation that it actually works on windows. It should not break other platforms, but the fixes I made are done completely blind -- hope to get to testing tonight.

yarikoptic:
if we are supporting windows now, we should establish CI on it, e.g. using appveyor: here is an example https://github.com/duecredit/duecredit/blob/master/appveyor.yml .

mih:
re windows CI: let's wait till at least one test ever passed during manual invocation....

mih:
@bpoldrack @yarikoptic new screenshot -- major progress. `create` (incl. subdatasets) works, `install` (incl. subdatasets) works. 

In `get` we are having a windows issue. The German message says that the log file that it tries to remove is still "in use" by another process. I assume this is the downloader.

mih:
If the tests pass, I'll merge this.

yarikoptic:
re downloader -- "not sure".  that ds000001 could use indeed datalad-archives, but there is also urls per each file which should be just fetched by git-annex curl/wget... you might want to see more of the traceback -- SET DATALAD_EXC_STR_TBLIMIT=20  or so? ;)

datalad/support/network.py
```diff
@@ -282,11 +283,24 @@ def __call__(self, url):
...
+        'file': PathRI,
+        'datalad': DataLadRI
+    }
+    # go in exotic mode if this is an absolute windows path
+    win_split = win_splitdrive(ri)
+    # we need a drive and a path, otherwise this could be a false positive
+    if win_split[0] and win_split[1]:
+        # OMG we got something from windows
+        lgr.log(5, "Detected file ri")
+        return TYPES['file']
```
yarikoptic: not necessarily from windows -- could be some local masocist ;-)
```
hopa:~
$> touch c:\\bugi\dugi
2 19687.....................................:Tue 13 Jun 2017 04:53:35 PM EDT:.
hopa:~
$> ls -ld c:\\bugi\dugi
-rw------- 1 yoh yoh 0 Jun 13 16:53 c:\\bugidugi
```
but the conclusion holds ;)

datalad/__init__.py
```diff
@@ -25,12 +25,16 @@
...
 import atexit
-atexit.register(ssh_manager.close, allow_fail=False)
+from datalad.utils import on_windows
+if not on_windows:
+    lgr.log(5, "Instantiating ssh manager")
+    from .support.sshconnector import SSHManager
+    ssh_manager = SSHManager()
+    atexit.register(ssh_manager.close, allow_fail=False)
+else:
+    ssh_manager = None
```
yarikoptic: IIRC There could be a cmdline ssh on Windows -- why should we just rule it out at once?

PR #1578, opened 2017-06-14
TST/BF: Skip inode tests if we could not change the inode.

bpoldrack:
Test for `AnnexRepo` with changed inode should not fail on NFS anymore.

This simply skips those tests, if the test's setup to achieve that inode change didn't work.


datalad/support/tests/test_gitrepo.py
```diff
@@ -1105,7 +1105,10 @@ def test_GitRepo_flyweight_monitoring_inode(path, store):
     # recreate
     shutil.copytree(store, path, symlinks=True)
     new_inode = os.stat(path).st_ino
-    assert_not_equal(old_inode, new_inode)
+
+    if old_inode == new_inode:
+        raise SkipTest("inode did not change. Nothing to test for.")
```
yarikoptic: oh perils of duplication

yarikoptic:
some recent change brought a new ramification of the locking issue on python3... heh

bpoldrack:
This is weird. Now it passes that test on NFS, too, but does NOT raise the `SkipTest`.

bpoldrack:
Okay, lock-issue happened again, everything else seems fine. Thus merging.

PR #1583, opened 2017-06-15
WIP: RF: lazy loading *Repo.repo

bpoldrack:
- don't instantiate `git.Repo` if we don't need it. Thus, making creation of *`Repo` instances cheaper
- fix *`Repo.is_valid_repo` to respect .git files
- make *`Repo.get_remotes()` not using `git.Repo` by default (needed in `AnnexRepo` constructor and therefore this change is required to not intantiate `git.Repo` immediatly

### DIFF
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -1170,6 +1170,12 @@ class GitRepo(RepoInterface):
         remotes : list of str
           List of names of the remotes
         """
+
+        # TODO: read directly from config and spare instantiation of gitpy.Repo
+        # since we need it in AnnexRepo constructor. Furthermore gitpy does it
+        # the same way and the use of a Repo instance seems to have no reason
+        # other than a nice object oriented look.
+
         if with_refs_only:
             # older versions of GitPython might not tolerate remotes without
             # any references at all, so we need to catch
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -1171,11 +1171,8 @@ class GitRepo(RepoInterface):
           List of names of the remotes
         """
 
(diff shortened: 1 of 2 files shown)


datalad/support/gitrepo.py
```diff
@@ -697,7 +710,7 @@ def __eq__(self, obj):
     @classmethod
     def is_valid_repo(cls, path):
         """Returns if a given path points to a git repository"""
-        return exists(opj(path, '.git', 'objects'))
+        return exists(opj(path, '.git'))
```
mih: Is this still good enough? We used `.git/objects` to avoid matching just anything that has a `.git` subdirectory. Any reason to change this?

bpoldrack: The most important reason is, that a repository with .git file was not accepted as valid by this method, since there is `.git/objects` in that case. We could read the file and construct the path to `objects` of course - would be a little bit more expensive.

Good enough - I think so. For now it doesn't break anything and: We also used the constructor of GitPython's `Repo` to determine whether a repo is valid. And this one does just this: Looking for a `.git`

bpoldrack: If we need a more extensive check of the validity of a repo, I'd vote for having two methods.
This one usually is used to find out, whether we should view a directory as a repository.
This should be cheap and not too sophisticated, I think.

If we say that just a `.git` isn't sufficient, one could also argue, that `.git/objects` isn't, since the sheer existence doesn't mean its content is valid ...

mih: There is no certainty, of course, but the likelihood is different -- more bits in the tested information ;-)

Anyways, good enough.

PR #1584, opened 2017-06-15
ENH: tag some tests as slow/usecase/integration and skip them in generic runs

### LINKED ISSUE #1580: proposal: tag and move into a dedicated run "usecase" tests
some usecase tests are pulling us down and we avoid incorporating them until coming up with a minimal example

I wonder if we should by default run non-usecase (usually faster) tests by default but have a matrix entry to run usecase ones separately?  this should help to overcome overly long durations

yarikoptic:
fixes #1580

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ env:
     - DATALAD_LOG_CMD_ENV=GIT_SSH_COMMAND
     - TESTS_TO_PERFORM=
     - NOSE_OPTS=-s
+    - NOSE_SELECTION_OP="not "   # so it would be "not (integration or usecase)"
     # Special settings/helper for combined coverage from special remotes execution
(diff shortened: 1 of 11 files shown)


yarikoptic:
whoohoo -- seems to "work" but changes in coverage are interesting:
- apparently body within 
```python
    def _git_custom_command(self, *args, **kwargs):
        if self.GIT_DIRECT_MODE_PROXY:
```
was tested only by one of the slow/... tests ... any clue which/why?  I thought it would be tested at least by something in test_annexrepo.py

I think it should be ok to annotate more of slowish tests (citing first 30 below) to get those additional runs balanced with typical runs and thus hopefully notably shortening duration of a typical run.

mih:
This is great!

PR #1586, opened 2017-06-16
BF: A non-installed subdataset should be considered 'unavailable'

mih:


yarikoptic:
I have no clue what is going on/for, but I guess it is all good ;)

PR #1589, opened 2017-06-17
`metadata` command 

mih:
On the way to 1st class preferred content management we need a frontend for git annex' metadata (tagging). Here is comes. While being add it I also implemented native support for dataset-level metadata:

Changes:

- [x] make `AnnexRepo` getter/setter generators
- [x] draft of a highlevel command to channel all of git-annex' file based metadata support
- [x] uniform support for dataset-level metadata via the same interface and semantics as for file-level metadata
- [x] dataset-level metadata is stored in JSON format (optimized for diffability) directly in Git
- [x] support for key definitions: while this is not the same as linked data, it represents a minimalistic compromise between precise semantics, and compact representation with minimal performance impact
- [x] somewhat useful custom result renderer for metadata
- [x] usual multi-dataset handling

More details are in the docstring of the command.

Possible TODOs (not necessarily in this PR):

- [ ] add `--import-from` option that uses a selected metadata parser (from the existing set that we have) to extract metadata from a supported format and imports it into our own dataset-level metadata. This should be easy to implement by building a corresponding value for the `init` arg of this command from the parser results, hence only amending existing metadata. Actually the edit strategy could be selected as `init`, `add`, or `reset`.

### DIFF
--- a/datalad/distribution/siblings.py
+++ b/datalad/distribution/siblings.py
@@ -522,17 +522,18 @@ def _configure_remote(
                         message='cannot configure as a common data source, '
                                 'URL protocol is not http or https',
                         **result_props)
-            for prop, var in (('wanted', annex_wanted),
-                              ('required', annex_required),
-                              ('group', annex_group)):
-                if var:
-                    ds.repo.set_preferred_content(prop, var, name)
-            if annex_groupwanted:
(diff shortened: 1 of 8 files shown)


mih:
Merging this matured PR before it gets rancid...

PR #1591, opened 2017-06-19
No API function with an underscore in its name has a proper docstring in the Python API

mih:
This is confounded with those functions not (yet) using the `@build_doc` decorator...


mih:
Will merge this now. I have a more comprehensive change in a sec....

Issue #1542, opened 2017-06-06
we are not guarding against duplicate (positional + kwarg) specification of API calls


yarikoptic:
I invoked 
```
subds.save('sub', message='saving sub', super_datasets=True)
```
but mistake thinking that we have consistent "files go as the first argument", but apparently the signature is

```python
    def __call__(message=None, files=None,  ...
```

so first one is the 'message' and that 'sub' was taken as the message while ignoring kward `message`. So it feels that there is absent some check somewhere to behave similar to python whenever there is a duplicate definition for a kwarg first through positional and then by the name

```
$> cat /tmp/2.py
def f(x=1, y=None):
  print(x)
f(1, x=2)

$> python /tmp/2.py 
Traceback (most recent call last):
  File "/tmp/2.py", line 3, in <module>
    f(1, x=2)
TypeError: f() got multiple values for keyword argument 'x'
```

PR #1592, opened 2017-06-19
Use @build_doc decorator for all commands

mih:
This fixes a number of issues

- [x] documentation for commands that specified 'Returns' was absent from the HTML docs, due to a formatting error
- [x] defaults for common args where no validated against their constraints
- [x] code duplication for arg docstring rendering removed


yarikoptic:
feels like https://github.com/datalad/datalad/issues/1542 could be related to these changes may be? ;)

mih:
Hmm, I dont see how. It is not changing the signature.

PR #1593, opened 2017-06-20
Some (standalone) parts of #1544

mih:
Attempt to slim PR #1544 down to be able to locate cause of segfault.


mih:
OK, so these were the good parts. I will merge this now, and then rebase #1544 onto master.

PR #1598, opened 2017-06-21
Limit recursive install on a per-subdataset basis

mih:
Docs are in the diff.

Data dependency datasets in the studyforrest collection want to marked up with this, before they come on board.

### DIFF
--- a/datalad/distribution/subdatasets.py
+++ b/datalad/distribution/subdatasets.py
@@ -33,6 +33,7 @@ from datalad.support.gitrepo import InvalidGitRepositoryError
 from datalad.support.exceptions import CommandError
 from datalad.interface.common_opts import recursion_flag
 from datalad.interface.common_opts import recursion_limit
+from datalad.distribution.dataset import Dataset
 from datalad.distribution.dataset import require_dataset
 from datalad.cmd import GitRunner
 from datalad.support.gitrepo import GitRepo
@@ -345,6 +346,9 @@ def _get_submodules(dspath, fulfilled, recursive, recursion_limit,
                     val)
                 # also add to the info we just read above
                 sm['gitmodule_{}'.format(prop)] = val
+            Dataset(dspath).add(
+                '.gitmodules', to_git=True,
+                message='[DATALAD] modified subdataset properties')
 
         #common = commonprefix((with_pathsep(subds), with_pathsep(path)))
         #if common.endswith(sep) and common == with_pathsep(subds):
--- a/datalad/distribution/subdatasets.py
+++ b/datalad/distribution/subdatasets.py
@@ -162,11 +162,22 @@ class Subdatasets(Interface):
     "gitmodule_<label>"
         Any additional configuration property on record.
 
-    Performance note: Requesting `bottomup` reporting order, or a particular
-    numerical `recursion_limit` implies an internal switch to an alternative
-    query implementation for recursive query that is more flexible, but also
-    notably slower (performs one call to Git per dataset versus a single call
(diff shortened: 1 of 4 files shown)


datalad/distribution/subdatasets.py
```diff
@@ -161,11 +162,22 @@ class Subdatasets(Interface):
...
+    dataset versus a single call for all combined).
+
+    The following properties for subdatasets are recognized by DataLad
+    (without the 'gitmodule_' prefix that is used in the query results):
+
+    "datalad-recursiveinstall"
+        If set to 'skip', the respective subdataset is skipped when DataLad
+        is recursively installing its superdataset. However, the subdataset
+        remains installable when explicitly requested, and no other features
+        are impaired.
```
mih: If that is an actual use case it would need a switch.

mih:
WOW, replicated the `publish` segfault!!! With completely different diff. I guess we exceeded the amount of possible changes....

datalad/distribution/subdatasets.py
yarikoptic: I wonder if we better just record dataset uuid and have a configurable per install strategy on what to do with datasets which were already installed elsewhere (above current dataset). With all the new ways of handling args, could we also pass such information inside?
My point is that I am not sure if it is the right place to decide... Sure thing we could add more  switches and knobs as we discover more

mih:
So you are saying that anyone that installs /// should receive a list of
uuids that they should supply to prevent 7 levels of redundant datasets?

Maybe not...

mih:
Just to clarify: without a switch for a maintainer to indicate that a subdataset is not for install we will have a problem once studyforrest is part of ///

Simply because some subdatasets will never be available, hence `install -r ///` always fail from that day onwards.

yarikoptic:
No. When you install top level you know uuids of datasets at that level. When you install a subdataset recursively, you can know what datasets available at the levels above

mih:
It does not help in the situation that I outlined, such dataset will never be installed and always cause failure.

yarikoptic:
Could you please point to that "outline"? ;-)

mih:
5cm up: https://github.com/datalad/datalad/pull/1598#issuecomment-310141516

yarikoptic:
Ah, so for some internal subdatasets, not to be shared (yet) etc. Yeah, for those needs explicit marker. I thought it was too be used also for marking the ones included in multiple places within the bigger collection

mih:
In studyforrest I use that same marker for both. The is simply no point in having the installed on `install -r`, but not because some variant of some dataset might have popped up somewhere to a magically "correct" order. These datasets are inputs to their parents. The only ever need to be there to recompute the parent. If I want that I use datalad to grab the files right from the non-installed subdatasets. The included tests documents that this functionality is not impaired.

mih:
That being said, the marker could have a different name. For example, it would be nice to be able to say "uninstall all input datasets --recursive"

mih:
So there you have it. This PR has virtually no new code that actually runs, yet it still segfaults. My point: this cause is not in the PRs, but it is already in master and we are sampling random variation that make it finally go KABOOM.

I am out of ideas. In any case, stopping to merge PRs because of this segfault seems counterproductive.

https://travis-ci.org/datalad/datalad/jobs/245645529

yarikoptic:
I will look at segfault as soon as I find Ethernet Port/internet for the laptop

yarikoptic:
So now we could install study Forrest? :-)

mih:
No, functionality is not in master...

PR #1602, opened 2017-06-22
NF: AnnexRepo.get_tags()

mih:


datalad/crawler/nodes/annex.py
```diff
@@ -1327,7 +1327,7 @@ def _finalize(data):
                         tag_ = last_version
                     # TODO: config.tag.sign
                     stats_str = "\n\n" + total_stats.as_str(mode='full')
-                    tags = self.repo.repo.tags
+                    tags = [t['name'] for t in self.repo.get_tags()]
```
yarikoptic: Getting the names of the tags seems to be too common to repeat as a list comprehension... Should get_tags get an argument to just return names?

mih: Done!

datalad/support/gitrepo.py
```diff
@@ -1999,6 +1999,36 @@ def tag(self, tag):
             '', ['git', 'tag', str(tag)]
         )
 
+    def get_tags(self, names_only=False):
```
yarikoptic: note to ourselves -- we might later want to standardize such parameter.  Eg. for `AnnexRepo.whereis` some time ago I added `output` option (return is a keywork) which could be from ` {'descriptions', 'uuids', 'full'}`...

yarikoptic:
restarted one python3 build on travis -- locking fights back... really need to look into it at some point :-/  now seems to happen only when our custom archives remote runs in the background, so it is better but not solved yet

datalad/support/gitrepo.py
mih: Yeah, yeah....communists....everything has to look the same.... I like it. Pushed.

PR #1605, opened 2017-06-23
RF: `save` not using GitPython API directly

mih:


mih:
This is a nobrainer. Will merge.

PR #1607, opened 2017-06-23
Close opened files and PIPEs

yarikoptic:
Saw python3-debug complaining that there are non-closed files, and found a few spots where we possibly didn't close files properly and could potentially leak/exhaust resources.


mih:
(fingers crossed!)

Issue #1612, opened 2017-06-24
`create` should initialize datasets with a default README file


mih:
This file should identify a repository as a datalad one, and should give an uninitiated reader some info on datalad and how to make use of the repo, as a dataset. This is related to #1462 

Some inspiration for default README content could be taken from: https://github.com/psychoinformatics-de/studyforrest-data-phase2/blob/master/README.rst

yarikoptic:
I don't think we should create by default... Should be a part of templating

mih:
This is no possible by running the `add_readme` plugin via `--run-after` on `create`.

Issue #1462, opened 2017-04-10
Missing feature: Dataset templates


mih:
We have the need to (repeatedly) create datasets with a specific structure and partially uniform content. One identified case is dataset metadata (#1457) that needs to be provided in some form. However, the need for pre-configuration goes beyond that. For example, I tend to have a `code/` folder in each dataset, and I always want to have all files in this directory to be committed to git directly and not annexed. Likewise, I tend to use one and the same licence for all datasets that I produce, and my name also hasn't changed since birth.

I think we should have the ability to create datasets from a pre-configured template. Such a template could be a tarball with a pre-crafted directory structure that has pre-populated metadata files, but also `.gitattribute` files in all the right places, etc. The mechanism should be format agnostic (just files and directories). However, it might be useful to have a dedicated place in the tarball the enables further modification at the time of creation. This could be implemented as a config file at `.datalad/create_tmpl.cfg` (or whatever) inside the tarball that has a general section, and (optionally, if more tuning is needed) one section per each to be modified file. The settings in each section are key value mappings of `PLACEHOLDER = TARGETVALUE` where each target value can be overwritten giving an option for the corresponding placeholder to `create`. For example: `create mydataset --override @@LICENSE@@ CC0`, where "@@LICENSE@@" is the placeholder and "CC0" the replacement for the preconfiguration in the template.

Such a mechanism would facilitate creation of structured datasets in any form. It would also shift the paradigm (that we use in our present tests, for example) from a rather procedural creation process (first a file, then two more) to a more goal-oriented approach: "gimme a BIDS skeleton", "gimme a CRCNS dataset shell". It would also generalize the approach in #1457 to scale beyond a single metadata file. The solution for #1457 could then be used as a default "interactive" template override that enables people to be guided to all files and all setting relevant to a dataset, instead of long option chains (some multi-line values will be hard to stringify and could be easily copied into an editor).

mih:
Any thoughts on this @yarikoptic ?

yarikoptic:
In general I like the idea... Yet to read few more times for better diggestion

yarikoptic:
that is what was nagging me:  such templates couldn't be "composed" together and thus would be not as flexible as they could be made if created on the fly given a set of specifications.  E.g. a particular case you brought up 
- "`code/` folder in each dataset, and I always want to have all files in this directory to be committed to git directly and not annexed".  
[...]

PR #1619, opened 2017-06-25
NF: `plugin` command

### LINKED ISSUE #1039: export --help-type "complains" instead of providing list of options
I guess it was intended (the same on 0.3.1 and last commit which modified datalad/export f2e48aaeb9730deec5f06d27e743fc5f282e169f , so probably never worked) to output summary of options for that backend but it just exits with error code

``` bash
(git)hopa:~datalad/datalad[enh-save]git
$> datalad export --help-type
usage: datalad export [--version] [-h] [-l LEVEL] [-p {condor}] [-d DATASET]
                      [--help-type] [-o OUTPUT]
                      {tarball}
datalad export: error: too few arguments
(venv-tests) 2 14040 ->2
```

mih:
Provide complete plugin interface (more general than `export` ever was). The main purpose is to be able to quickly craft little helpers as simple Python functions and run them as standalone commands, without all the overhead of crafting a full-scale command with a nifty API.

Sufficient docs should be inside. This is a precondition for a resolution of:  #1612 #1462 

Here is a list of existing issues that could be addressed by a plugin: https://github.com/datalad/datalad/labels/do-as-plugin

- [x] Replaces the old `export` command (fixes #1039)
- [x] Old tarball exporter is now a plugin `export_tarball`.
- [x] New `add_readme` plugin.
- [x] allow for giving plugin arguments multiple times to build input argument lists
- [x] more or less comprehensive tests for the general plugin mechanics
- [x] minimal WTF plugin
- [x] demo of a `--with-plugin` option that has the potential to simplify our API a bit
- [x] RF basic decorators `build_doc` and `eval_results`
- [x] common option (as a pair `--run-before`, `--run-after`) to run plugins with any command. With this model we could give users the ability to do `autoadd_untracked`, `autosave_dirty`, `fail_if_dirty` if they want to roll this way, without us having to decide what is best for them, and without having to explode the API to cover all possible cases.
- [x] allow for configuration of plugins that shall always run pre/post of a particular command

### DIFF
--- a/datalad/distribution/create.py
+++ b/datalad/distribution/create.py
@@ -27,6 +27,7 @@ from datalad.interface.common_opts import annex_init_opts
 from datalad.interface.common_opts import location_description
 from datalad.interface.common_opts import nosave_opt
 from datalad.interface.common_opts import shared_access_opt
+from datalad.interface.common_opts import with_plugin_opt
 from datalad.support.constraints import EnsureStr
 from datalad.support.constraints import EnsureNone
 from datalad.support.constraints import EnsureKeyChoice
@@ -112,12 +113,14 @@ class Create(Interface):
             doc="""enforce creation of a dataset in a non-empty directory""",
             action='store_true'),
         description=location_description,
+        # TODO could move into cfg_annex plugin
         no_annex=Parameter(
             args=("--no-annex",),
             doc="""if set, a plain Git repository will be created without any
             annex""",
             action='store_true'),
         save=nosave_opt,
+        # TODO could move into cfg_annex plugin
         annex_version=Parameter(
             args=("--annex-version",),
             doc="""select a particular annex repository version. The
@@ -125,6 +128,7 @@ class Create(Interface):
             version. This should be left untouched, unless you know what
             you are doing""",
             constraints=EnsureDType(int) | EnsureNone()),
+        # TODO could move into cfg_annex plugin
         annex_backend=Parameter(
             args=("--annex-backend",),
             constraints=EnsureStr() | EnsureNone(),
@@ -135,6 +139,7 @@ class Create(Interface):
             of datasets across platforms (especially those with limited
             path lengths)""",
             nargs=1),
+        # TODO could move into cfg_metadata plugin
         native_metadata_type=Parameter(
             args=('--native-metadata-type',),
             metavar='LABEL',
@@ -143,7 +148,9 @@ class Create(Interface):
             doc="""Metadata type label. Must match the name of the respective
             parser implementation in Datalad (e.g. "bids").[CMD:  This option
             can be given multiple times CMD]"""),
+        # TODO could move into cfg_access/permissions plugin
         shared_access=shared_access_opt,
+        with_plugin=with_plugin_opt,
         git_opts=git_opts,
         annex_opts=annex_opts,
         annex_init_opts=annex_init_opts,
@@ -164,6 +171,7 @@ class Create(Interface):
             native_metadata_type=None,
             shared_access=None,
             git_opts=None,
+            with_plugin=None,
             annex_opts=None,
             annex_init_opts=None):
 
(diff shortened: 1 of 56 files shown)


mih:
@yarikoptic @bpoldrack Any idea why OSX would yield different paths (see failing test)?

datalad/plugin/dlplugin_add_readme.py
```diff
@@ -0,0 +1,79 @@
...
+#PLUGINSYNOPSIS: add a README file to a dataset
+"""
+
+"""
+
+__docformat__ = 'restructuredtext'
+
+
+# PLUGIN API
+def datalad_plugin(dataset, filename='README.rst', existing='skip'):
```
yarikoptic: I wondered if we really need two levels of different "identification prefixes" -- "dlplugin_" for the filename and "datalad_" for the function?  imho should be enough to have just "plugin" here

datalad/plugin/dlplugin_wtf.py
```diff
@@ -0,0 +1,84 @@
...
+__docformat__ = 'restructuredtext'
+
+
+# PLUGIN API
+def datalad_plugin(dataset=None):
+    """Generate a report about the DataLad installation and configuration
+
+    IMPORTANT: Sharing this report with untrusted parties (e.g. on the web)
+    should be done with care, as it may include identifying information, and/or
+    credentials or access tokens.
```
yarikoptic: so would be lovely to have an option to share just what should be most relevant for debugging -- versions of externals, and may be dataset id

mih: It already filters out cfg items under 'user....' and anything that has 'token' in it. Adding more modes would be easy, but not priority in this PR.

datalad/plugin/dlplugin_add_readme.py
mih: True -- somewhat. Will go for 'dlplugin' in all cases. I don't think going for generic symbol names is a benefit or avoid any issues.

mih:
I'll take the liberty to shorten the test dump by, say 35 pages, and only leave the end ;-)

yarikoptic:
output -- yeah, was quick paste which I then shortened right away :-P

datalad/distribution/create.py
```diff
@@ -333,6 +341,17 @@ def __call__(
         path.update({'status': 'ok'})
         yield path
 
+        for pluginspec in with_plugin if with_plugin else []:
```
mih: done

mih:
@yarikoptic PY3 issue should be gone now, would still appreciate thoughts on the OSX path failure.

mih:
In order to do these tests properly, we would have to RF the tests to always run on a clean config. Reason, now any user can significantly alter the behavior of any command via plugins. For example, my own local config now add a custom to description to my datasets. This is additional content that plain datalad does not expect, hence tests fail hard under my account.

mih:
From @yarikoptic 

> OSX plugin PR issue -- symlinks shoot back (not sure yet how doesn't fail on travis): 
(Pdb) os.path.realpath('/var/folders/90/vkz4dwlx0ss72djd8hywgppc0000gp/T/datalad_temp_tree_test_tarballSRCw0g/datalad_0447f6e6-5c53-11e7-9286-00233295b230.tar.gz')
'/private/var/folders/90/vkz4dwlx0ss72djd8hywgppc0000gp/T/datalad_temp_tree_test_tarballSRCw0g/datalad_0447f6e6-5c53-11e7-9286-00233295b230.tar.gz'
(Pdb) p results
[{'status': 'ok', 'action': 'export_tarball', 'path': '/private/var/folders/90/vkz4dwlx0ss72djd8hywgppc0000gp/T/datalad_temp_tree_test_tarballSRCw0g/datalad_0447f6e6-5c53-11e7-9286-00233295b230.tar.gz', 'type': 'file', 'refds': '/var/folders/90/vkz4dwlx0ss72djd8hywgppc0000gp/T/datalad_temp_tree_test_tarballSRCw0g/ds'}]

> (Pdb) p kwargs
{'path': '/var/folders/90/vkz4dwlx0ss72djd8hywgppc0000gp/T/datalad_temp_tree_test_tarballSRCw0g/datalad_0447f6e6-5c53-11e7-9286-00233295b230.tar.gz'}
(venv-ci)bash-3.2$ ls -l /var
lrwxr-xr-x@ 1 root  wheel  11 Dec 18  2014 /var -> private/var

mih:
Trying to workaround the path issue on OSX -- will merge this baby once successful.

Issue #1618, opened 2017-06-24
does not provide a clue that we just can't install here


djarecka:
```shell

(datalad) Witalis:tmp dorota$ cd /
(datalad) Witalis:/ dorota$ datalad install ///
[INFO   ] Cloning dataset from 'http://datasets.datalad.org/' (trying 2 location candidate(s)) to '/datasets.datalad.org' 
[ERROR  ] Failed to clone data from any candidate source URL: ['http://datasets.datalad.org/', 'http://datasets.datalad.org/.git'] [install(/datasets.datalad.org)] 
install(error): /datasets.datalad.org (dataset) [Failed to clone data from any candidate source URL: ['http://datasets.datalad.org/', 'http://datasets.datalad.org/.git']]
```

mih:
Thanks for the report. This is related to, but different from #1563

mih:
With #1624 it looks like this now:

````
mih@meiner / % datalad install ///
[INFO   ] Cloning dataset from 'http://datasets.datalad.org/' (trying 2 location candidate(s)) to '/datasets.datalad.org' 
[ERROR  ] could not create work tree dir '/datasets.datalad.org': Permission denied [install(/datasets.datalad.org)] 
````

PR #1624, opened 2017-07-03
BF: Proper clone failure report due to insufficient permissions

mih:
Fixes:

- [x] #1618 
- [x] Broken argspec of `clone` cmdline parser

### DIFF
--- a/datalad/distribution/tests/test_clone.py
+++ b/datalad/distribution/tests/test_clone.py
@@ -45,6 +45,7 @@ from datalad.tests.utils import ok_clean_git
 from datalad.tests.utils import serve_path_via_http
 from datalad.tests.utils import use_cassette
 from datalad.tests.utils import skip_if_no_network
+from datalad.tests.utils import skip_if_on_windows
 
 from ..dataset import Dataset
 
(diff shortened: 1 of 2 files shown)


datalad/distribution/tests/test_clone.py
```diff
@@ -309,3 +310,13 @@ def test_clone_isnt_a_smartass(origin_path, path):
     assert clonedsub.path.startswith(path)
     # no subdataset relation
     eq_(cloned.subdatasets(), [])
+
+
+@skip_if_on_windows
+def test_clone_report_permission_issue():
+    with chpwd('/'):
```
yarikoptic: please do not!  noone forbids running datalad or its tests as root, which we do during package building.  This obviously would fail and leave junk laying around.  IMHO regular `with_tempfile(mkdir=True)` + mkdir + chmod call within and then chpwd into it would be sufficient BUT still would require skipping the test if ran as super-user -- there user/git has too much power 
```shell
$> mkdir -p /tmp/testperm/d; chmod a-w /tmp/testperm/d; cd /tmp/testperm/d; datalad install ///
[INFO   ] Cloning dataset from 'http://datasets.datalad.org/' (trying 2 location candidate(s)) to '/tmp/testperm/d/datasets.datalad.org' 
[ERROR  ] Failed to clone data from any candidate source URL: ['http://datasets.datalad.org/', 'http://datasets.datalad.org/.git'] [install(/tmp/testperm/d/datasets.datalad.org)] 
install(error): /tmp/testperm/d/datasets.datalad.org (dataset) [Failed to clone data from any candidate source URL: ['http://datasets.datalad.org/', 'http://datasets.datalad.org/.git']]
```
and

PR #1625, opened 2017-07-03
NF: Plugin that can convert BIDS metadata into ISATAB format optimized for submission to Scientific Data

mih:
This has been imported from https://github.com/mih/BIDS2ISATab which itself is a fork of https://github.com/INCF/BIDS2ISATab. It is functional, but not complete yet.

Changes
- [x] convert from standalone implementation into a datalad plugin
- [x] make use of datalad's metadata extractors, instead of "hand-parsing" files (https://github.com/INCF/pybids/issues/85 is a showstopper for now)
   - majority of the old code is concerned with assembling and merging metadata from various sources -- this is already done by datalad. The only thing left todo is to sort and group metadata by file type, and to format a nice table
- [x] assay tables can be generated in the order desired by SciData
- [x] support for multiple sources for a single assay table field (e.g. auto-amend BIDS from NIfTI header metadata)
- [x] reinject ontology info for parameters and parameter values, source info is in place, but respective table columns have to be (re)created
- [x] populate study and investigator info more completely, possibly using other metadata sources supported by datalad (including its own native metadata), absolutely straightforward for the author list
- [x] some remaining issue with the term ref columns in the assay tables, some seem to get misplaced, check phase2 output for demo

This is now working as intended. Future changes will have to be:

- stop relying on present `participants.tsv` file and use info from aggregated metadata -- requires conversion on metadata structure that nobody can afford right now
- implement regression test on (multiple) real datasets -- will become possible (and cheap) once our own datasets carry new aggregated metadata

### DIFF
--- a/datalad/metadata/definitions.py
+++ b/datalad/metadata/definitions.py
@@ -105,7 +105,7 @@ common_defs = {
         'descr': 'name of a resource'},
     "shortdescription": {
         'def': "doap:shortdesc",
-        'descr': 'sort description or summary or title of a resource'},
+        'descr': 'short description or summary or title of a resource'},
     # wondering why there is no title, eh?
     # MIH: we have name and short description
     #      adding title seems superfluous
--- a/datalad/metadata/metadata.py
+++ b/datalad/metadata/metadata.py
@@ -288,7 +288,7 @@ def _get_containingds_from_agginfo(info, rpath):
     return dspath
 
 
-def _query_aggregated_metadata(reporton, ds, aps, merge_mode, recursive=False,
+def query_aggregated_metadata(reporton, ds, aps, merge_mode, recursive=False,
                                **kwargs):
     """Query the aggregated metadata in a dataset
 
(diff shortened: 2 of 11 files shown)


yarikoptic:
Shouldn't the meat of this PR just get to eg pybids and plugin just use it?

mih:
AFAIK this code predates pybids. Eventually some aspects could be RFed to use it, I agree. However, my plan is actually to make more use of datalads metadata facilities, and less of raw BIDS parsing in this plugin. pybids should consequently play a bigger role on our BIDS parser and not so much in here. We'll see how that goes.

mih:
FTR: in master we are now sitting on top of pybids and have access to metadata from all files, plus stuff from `participants.tsv` which pybids won't give one ATM. Time has come to RF the plugin...

mih:
https://github.com/INCF/pybids/issues/86 bring this PR to a screeching halt for now.

chrisfilo:
What's a datalad plugin? Any chance of removing the redundancy and storing common code in https://github.com/INCF/BIDS2ISATab?

mih:
@chrisfilo A datalad plugin is like a datalad command

This PR started out from a fork of BIDS2ISATAB (see toplevel comment). It would certainly be possible to push back some of the changes to BIDS2ISATAB. However, the main point of this plugin is that we can generate ISATAB straight from datalads own representation of a dataset's metadata -- without the need to have the dataset itself around. So I am not sure whether there will be enough code left that is in common/redundant. But we'll see when it is done.

PR #1626, opened 2017-07-03
Intensified and cross-platform handling of colored terminal output

mih:
Please see the two commits for details.

### DIFF
--- a/datalad/__init__.py
+++ b/datalad/__init__.py
@@ -35,6 +35,20 @@ if not on_windows:
 else:
     ssh_manager = None
 
+try:
+    # this will fix the rendering of ANSI escape sequences
+    # for colored terminal output on windows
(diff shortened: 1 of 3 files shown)


datalad/__init__.py
```diff
@@ -35,6 +35,20 @@
...
+    # for colored terminal output on windows
+    # it will do nothing on any other platform, hence it
+    # is safe to call unconditionally
+    import colorama
+    colorama.init()
+    atexit.register(colorama.deinit)
+except ImportError:
+    if on_windows:
+        lgr.warning(
+            "'colorama' Python module missing, terminal output may look garbled")
```
yarikoptic: It is just an assumption... Import could fail for other reason (bug in module etc)... That is why I usually prefer just to include exception detail in the message

mih:
Good point, will do.

PR #1627, opened 2017-07-04
ENH(BF): close PIPEs when done (python-dbg complaints)

yarikoptic:
- [ ] - apparently wasn't finished properly yet...


yarikoptic:
not sure when and if still needed, so closing

PR #1628, opened 2017-07-04
BF: do not try to clone in / and skip if su

yarikoptic:
addressing my whining in https://github.com/datalad/datalad/pull/1624/files#r125349246


mih:
Thanks! Restarted PY3 test run. Will merge when succeeded.

Issue #1601, opened 2017-06-22
Should meta-data be unified


yarikoptic:
Related to #1589 
I wondered if we should make meta-data handling hierarchical, similar to the configuration (`where` currently supporting `datalad`, `local`, `global`) with  inheriting, so possible `where` for defining meta could be
- `dataset` -- `.datalad/meta` thus using the schema we already have) 
- `directory` -- not supported at all ATM, but we could have .datalad/meta there similar to dataset level) 
- `file` ( git-annex)

while unifying/extending our scheme (http://docs.datalad.org/en/latest/metadata.html#json-ld-meta-data-format)?

mih:
I think supporting hierarchical would be useful, but at the same time also tricky. I would be able to set a "tag" for all files in a subtree. At the same time, I would not be able to use such a tag to move data via annex.

I find it difficult to map triples onto annex' meta data capabilties. Hence in #1589 I went for unification with annex (except for key definitions). I think this works well. Provides a meaningful, yet minimal metadata system that complements "bigger", but separate metadata sets.

PR #1630, opened 2017-07-04
Metadata facility update

### LINKED ISSUE #1793: `git-annex metadata` silently ignores untracked files or files directly in Git
No output from which to infer that it didn't do anything, nor a non-zero exit code.

```sh
mih@meiner /tmp/demo (git)-[master] % git st
On branch master
mih@meiner /tmp/demo (git)-[master] % echo "134" > probe
mih@meiner /tmp/demo (git)-[master] % git annex metadata -s tag+=mike probe
mih@meiner /tmp/demo (git)-[master] % git st
On branch master
Untracked files:

	probe

nothing added to commit but untracked files present (use "git add" to track)
mih@meiner /tmp/demo (git)-[master] % git annex metadata probe
mih@meiner /tmp/demo (git)-[master] % git add probe
mih@meiner /tmp/demo (git)-[master] % git commit -m add
[master 3fd1a73] add
 1 file changed, 1 insertion(+)
 create mode 100644 probe
mih@meiner /tmp/demo (git)-[master] % git annex metadata probe -s tag+=mike
mih@meiner /tmp/demo (git)-[master] % git annex metadata probe
```

@joeyh Do you consider this a bug in git-annex?

If not, we need to adjust our metadata manipulation code to check that we got a response from git-annex for each file we asked to manipulate.

mih:
This is an attempt to adopt some ideas from huge PRs #1010, #1074, and #1077 on metadata that never made it into master. However, this time things will hopefully turn out a bit simpler....and they did.

Changes

- [x] unify JSON-LD context with a simple metadata key defintion list (related to #1601)
- [x] `metadata --show-keys` (with human-readable description)
- [x] `metadata` complains when setting undefined keys (`--permit-undefined-keys` is provided as an annoying bypass). This should get users to converge on the keys that we pre-define, as otherwise they are promted to define their own in datasets.
- [x] get info on a datasets native metadata types
- [x] go away from guessing metadata types. Either a dataset informs datalad that it conforms to a standard, or it doesn't. It is a maintainer's choice to advertise (some) native metadata. This does not affect any of the legacy metadata processing code (yet).
- [x] be able to query metadata keys from native metadata (if it exists) and transparently amend custom datalad metadata with it -- while the custom one is preferred, in case alternative source have something to report. This is essentially doing metadata homogenization at the moment of per-dataset query -- without the needs for JSON-LD graphs. Obviously, this is limited to metadata understood be our parser, hence requires human creativity (but that is a good thing, and it only needs to happen once when a parser is written).
- [x] support three meta data merge modes: init, add, reset (see dedicated test), same semantics as for metadata manipulation
- [x] new-style metadata aggregation that minimized the number of moving parts across (sub)datasets -- no JSON-LD for aggregation anymore. Theoretical capability of an arbitrary number of metadata sources.
- [x] metadata query transparently grabs aggregated metadata from the closest dataset if a path is unavailable -- yields identical, homogenized metadata results -- compared to a direct query.
- [x] fast and powerful `search` reimplementation based on https://pypi.python.org/pypi/Whoosh (fucking amazing), on datalad.org we can even use the generated search index for "search-as-you-type" style operations...
- [x] change agg object filename to something that avoid collision when the same dataset appears multiple times as a subdataset
- [x] pray for enlightenment on how to adjust a broken test, where the mock test of the mocked results is mocking real hard, it hurts (poor michael has no clue)....
- [x] fix #1793 -- do not silently ignore files in git or untracked ones
- [x] native meta data parsers can now auto-obtain required annexed files
- [x] make it possible to annex aggregated metadata objects. it is actually possible already, but it doesn't happen because of our preconfiguration
- [x] implement `metadata` support for lookup of file-based metadata -- only few lines missing, `search` can already do it
- [x] reevaluate the native metadata parser setup, now that the `metadata` is in place, it could be done in a leaner fashion -- presently the things feels bolted on -- this is related to the question what layer of code gets to decide whether files containing metadata are to be obtained #1796
   - fix up aggregation of aggregated metadata, likely needs a gimme everything-you-know flag for `metadata (also see #1796)
   - parsers should stop trying to get core metadata files that are not known to a dataset
- [x] adjust search to benefit from the new behavior of `metadata` to already look up all relevant (and also aggregated) metadata. In particular, implement a recursive mode for `_query_aggregated_metadata()` to be used for generating the search index purely from aggregated metadata, and then run `metadata()` as before on the paths of all hits, to get the most up-to-date metadata report.
- [x] put parentds field into search index documents to be able to quickly pull out all content for a specific dataset
- [x] check what is wrong with CRCNS metadata extraction, shows unspecific error message
```
2017-09-28 11:35:30,130 [INFO   ] Aggregate metadata for dataset /home/mih/fullcopy/crcns/hc-2 
2017-09-28 11:35:30,513 [ERROR  ] Failed to get dataset metadata (datacite): [Errno 2] No such file or directory: '/home/mih/fullcopy/crcns/hc-2/.datalad/meta.datacite.xml' [datacite.py:get_dataset_metadata:108] 
2017-09-28 11:35:30,634 [ERROR  ] Metadata extraction failed (see previous error message) [aggregate_metadata(/home/mih/fullcopy/crcns/hc-2)] 
aggregate_metadata(error): /home/mih/fullcopy/crcns/hc-2 [Metadata extraction failed (see previous error message)]
2017-09-28 11:40:45,172 [INFO   ] Aggregate metadata for dataset /home/mih/fullcopy/crcns/zipser-1 
2017-09-28 11:40:45,387 [ERROR  ] Failed to get dataset metadata (datacite): [Errno 2] No such file or directory: '/home/mih/fullcopy/crcns/zipser-1/.datalad/meta.datacite.xml' [datacite.py:get_dataset_metadata:108] 
2017-09-28 11:40:45,515 [ERROR  ] Metadata extraction failed (see previous error message) [aggregate_metadata(/home/mih/fullcopy/crcns/zipser-1)] 
aggregate_metadata(error): /home/mih/fullcopy/crcns/zipser-1 [Metadata extraction failed (see previous error message)]
```
- [x] aggregation of aggregated metadata seems to loose information (unpushed test available), needs fixing
- [x] deal with metadata key conflict when multiple regexes match a filename (likely use existing merge_mode logic as when blending git-annex metadata) -- or we are just saying that conflicts are a sign of a stupid metadata representation anyways....
- [x] aggregation must add info to dataset-global metadata on which keys (and optionally unique values) are available in content metadata
- [x] RF search index creation into a 2-pass procedure: 1. fast(er) over just datasets in order to assess what keys have to be supported in the search index, and 2. (slower) overall all datasets and files to actually build the index. Rational: all keys need to be known before a complete schema can be constructed, and any dataset can define its own keys.
- [x] rename `ontology_id` to `vocabulary_id`
- [x] make it possible to actually search for content summary values. Info is already aggregated, but not yet accessible in the search index.
- [x] EXIF metadata parser to turn datalad into a photo database
- [x] Audio/ID3 metadata parser to turn datalad into a music database
- [x] DICOM metadata parser that extracts metadata keys that are unique across all images in a series
- now that we can pull out various flavors of metadata, we have to make things more flexible:
  - [x] add generic config item to limit the fields of extracted metadata by size.  For example, EXIF includes thumbnails as metadata, ...
  - [x] blacklist fields by name (regex, mayby)
  - [x] aggregate content metadata, but have switch to not actuallt store it, but only use it to generate unique metadata key/value pairs across all files of a dataset
  - [x] allow for turning off git-annex metadata extraction for aggregation. This can be crucial for datasets with tons of files to enable somewhat fast operation
         this is now possible by adding `datalad.metadata.aggregate-content-datalad-core = no` to a dataset's config. Analog switched are tested for any activated parser, so one could selectively disable dataset-global or content-metadata for aggregation
- [x] automatic invalidation of generated search indices, based on comparing the commit sha of aggregated metadata between now and generation time
- [X] do fast local term resolution to give access to vocabulary definitions if any exist. Demo:
```
% datalad search --show-keys
@id (unique identifier of an entity)
comment<albumartist> (http://purl.obolibrary.org/obo/NCIT_C25393)
comment<conductor> (http://purl.obolibrary.org/obo/NCIT_C25393)
comment<date> (http://purl.obolibrary.org/obo/NCIT_C25393)
comment<discnumber> (http://purl.obolibrary.org/obo/NCIT_C25393)
comment<encodedby> (http://purl.obolibrary.org/obo/NCIT_C25393)
comment<tracknumber> (http://purl.obolibrary.org/obo/NCIT_C25393)
dcterms:rights (http://purl.org/dc/terms/rights)
duration(s) {'unit (http://purl.obolibrary.org/obo/UO_0000000)': 'http://purl.obolibrary.org/obo/UO_0000010', '@id': 'https://www.w3.org/TR/owl-time/#Duration'}
format (http://purl.org/dc/elements/1.1/format)
music:Composer (http://purl.org/ontology/mo/Composer)
music:Genre (http://purl.org/ontology/mo/Genre)
music:album (http://purl.org/ontology/mo/album)
music:artist (http://purl.org/ontology/mo/artist)
music:channels (http://purl.org/ontology/mo/channels)
music:sample_rate (http://purl.org/ontology/mo/sample_rate)
name (http://schema.org/name)
parentds (path of the datasets that contains an entity)
path (path name of an entity relative to the searched base dataset)
type {'@id': 'http://schema.org/type', 'description': 'type or category of a resource (e.g. file, dataset)'}
```
- [x] bring all tests back in alignment with new desired behavior
- [x] adjust docs (but only after @yarikoptic gave his complete and unconditional blessing)
- [x] extend `metadata` to report subdatasets for which metadata is present in a dataset
- [x] decide whether to switch from JSON object storage to JSON streams for storage of aggregated metadata
- [x] datalad needs to be aware which metadata version it can handle, and datasets need to report which versions of metadata they contain (@bpoldrack doing that via filename change)
- [x] make it possible to store aggregated metadata in annexed files
- [x] make it possible to store aggregated metadata in compressed binary files
- [x] possibly depart from per-regex storage
      - the only metadata format where this would be helpful is BIDS, but at the same time none of the BIDS tools supports reporting by filename regex, hence a complicated machinery would be necessary to reconstruct this
      - alternatively, and possibly more productively we could use a standard compression algorithm

Compression:

```py
# massively redundant metadata for 1M files
d={k:k for k in range(10)}
s=[dict(d, path='longfuckingfilename{:030}'.format(i)) for i in xrange(1000000)]
json.dump(s, open('big.json', 'w'))
```
```
mih@meiner /tmp % gzip -9 -c big.json > big.gz
mih@meiner /tmp % xz -9 -c big.json > big.xz
% ls -lh big*
-rw-rw-r-- 1 mih mih 2,8M Okt 19 18:04 big.gz
-rw-rw-r-- 1 mih mih 137M Okt 19 18:04 big.json
-rw-rw-r-- 1 mih mih  83K Okt 19 18:04 big.xz
```

There is no point in implementing a custom metadata reduction algorithm for BIDS.

lzma (xz) support is built into Python 3.3+ and a backport is available for Python 2.

- [x] fix issue of unsaved dataset old aggregate metadata file deletion when reaggregating with a metadata change #1912 
- [x] pybids-based metadata parser doing a full report for all files, plus our custom enhancements from participants.tsv


datalad/metadata/definitions.py
```diff
@@ -0,0 +1,42 @@
...
+#   copyright and license terms.
+#
+# ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+"""Metadata (key) definitions"""
+
+
+# for maximum compatibility with git-annex' metadata setup, _keys_ in this
+# dictionary should be all lower-case, and be limited to alphanumerics, plus
+# '_', '-', and '.' -- except for JSON-LD keywords (which start with '@' and
+# will be ignore in the context of git-annex metadata
```
yarikoptic: Just spelling - ignored

datalad/metadata/definitions.py
```diff
@@ -0,0 +1,42 @@
...
+    "@vocab": "http://schema.org/",
+    "doap": "http://usefulinc.com/ns/doap#",
+    "author": "schema:author",
+    "audience": "doap:audience",
+    "citation": "schema:citation",
+    "contributors": "schema:contributor",
+    "description": "schema:description",
+    "homepage": "doap:homepage",
+    "issuetracker": "doap:bug-database",
+    # TODO this is 'tag'
```
yarikoptic: Do you mean keywords? Then no - since values could come with other names in this list

mih: I don't get this. I mean that an entry 'tag' -> "schema:keywords" is missing. Tags are keywords, I think. No?

mih: http://schema.org/keywords

datalad/metadata/definitions.py
mih: fixed

datalad/metadata/definitions.py
```diff
@@ -0,0 +1,42 @@
...
+    "@vocab": "http://schema.org/",
+    "doap": "http://usefulinc.com/ns/doap#",
+    "author": "schema:author",
+    "audience": "doap:audience",
+    "citation": "schema:citation",
+    "contributors": "schema:contributor",
+    "description": "schema:description",
+    "homepage": "doap:homepage",
+    "issuetracker": "doap:bug-database",
+    "tag": "schema:keywords",
```
yarikoptic: Just for my own sake - why bother with tag which would complicate somewhat mapping ?

datalad/metadata/metadata.py
```diff
@@ -330,22 +366,73 @@ def __call__(
...
+            known_keys = set(common_key_defs.keys()).union(set(defs.keys()))
+            key_error = False
+            for cat in (init, add, reset) if not permit_undefined_keys else []:
+                for k in cat if cat else []:
+                    if k not in known_keys:
+                        yield get_status_dict(
+                            status='error',
+                            ds=ds,
+                            message=(
+                                "undefined key '%s', check spelling or use --define-key",
```
yarikoptic: Should we point to our bug tracker to suggest new ones?

datalad/metadata/metadata.py
```diff
@@ -357,7 +444,7 @@ def __call__(
                     db[k] = v
                 for k, v in add.items():
                     db[k] = sorted(unique(
-                        db.get(k, []) + v))
+                        db.get(k, []) + list(v)))
```
yarikoptic: Couldn't v be a string, this coffee working without failure but adding bunch of characters? (Blind question - do not see the rest of the coffee ATM) there is `assure_list`

mih: indeed, assure_list would be safer. Not sure about the "coffee"  -- I'll stick with beer.

datalad/metadata/metadata.py
mih: Great idea!

datalad/metadata/definitions.py
mih: `tag` is what is used by git-annex -- so that is a given.

datalad/metadata/metadata.py
yarikoptic: Coffee is a coffee word for beer... Especially on 4th of July

yarikoptic: Doh... Bloody phone... Code word

mih:
Tests started segfaulting again. This time seemingly unlrelated to any `publish` functionality:

````
datalad.metadata.tests.test_manipulation.test_basic_filemeta ... /home/travis/.travis/job_stages: line 54:  9422 Segmentation fault      (core dumped) PATH=$PWD/tools/coverage-bin:$PATH $NOSE_WRAPPER `which nosetests` $NOSE_OPTS -v -A "$NOSE_SELECTION_OP(integration or usecase or slow)" --with-doctest --doctest-tests --with-cov --cover-package datalad --logging-level=INFO $TESTS_TO_PERFORM
The command "PATH=$PWD/tools/coverage-bin:$PATH $NOSE_WRAPPER `which nosetests` $NOSE_OPTS -v -A "$NOSE_SELECTION_OP(integration or usecase or slow)" --with-doctest --doctest-tests --with-cov --cover-package datalad --logging-level=INFO $TESTS_TO_PERFORM" exited with 139.
````

File: datalad/crawler/pipelines/tests/test_openfmri.py [hunk 6]
```diff
@@ -265,15 +265,15 @@ def test_openfmri_pipeline1(ind, topurl, outd, clonedir):
     eq_(len(commits_l['incoming']), 3)
     eq_(len(commits['incoming-processed']), 6)
     eq_(len(commits_l['incoming-processed']), 4)  # because original merge has only 1 parent - incoming
-    eq_(len(commits['master']), 12)  # all commits out there -- dataset init, crawler init + 3*(incoming, processed, meta data aggregation, merge)
-    eq_(len(commits_l['master']), 6)
+    eq_(len(commits['master']), 11)  # all commits out there -- dataset init, crawler init + 3*(incoming, processed, merge)
+    eq_(len(commits_l['master']), 5)
```
mih: the aggregate commit is gone, a dataset isn't changed anymore

datalad/interface/base.py
```diff
@@ -389,7 +389,7 @@ def call_from_parser(cls, args):
                 'AddArchiveContent', 'AggregateMetaData',
                 'CrawlInit', 'Crawl', 'CreateSiblingGithub',
                 'CreateTestDataset', 'DownloadURL', 'Export', 'Ls', 'Move',
-                'Publish', 'SSHRun', 'Search'):
+                'Publish', 'SSHRun'):
```
mih: it is not gone, it is just no longer a special case for API generation.

datalad/metadata/aggregate.py
```diff
@@ -11,32 +11,146 @@
...
+def _adj2subbranches(base, adj):
+    # given a set of parent-child mapping, compute a mapping of each parent
+    # to all its (grand)children of any depth level
+    branches = dict(adj)
+    # from bottom up
+    for ds in sorted(adj, reverse=True):
+        subbranch = []
+        for sub in branches[ds]:
+            subbranch.append(sub)
+            subbranch.extend(branches.get(sub, []))
```
mih: I hope too ;-)

datalad/metadata/aggregate.py
```diff
@@ -11,32 +11,146 @@
...
+        branches[ds] = subbranch
+    return branches
```
mih: It refers to the branches of the dataset tree. Honestly I don't care about this name, it is use in this place in less than a handful places -- a random string is fine with me.

datalad/metadata/definitions.py
```diff
@@ -0,0 +1,93 @@
...
+common_key_defs = {
+    "@vocab": {
+        'def': "http://schema.org/",
+        'descr': 'base vocabulary'},
+    "doap": {
+        'def': "http://usefulinc.com/ns/doap#",
+        'descr': 'ontology for the description of a project'},
+    "author": {
+        'def': "schema:author",
+        'descr': 'author of some resource'},
```
mih: We talked about this before, I believe that hand-written metadata will be rare, but I don't object examples -- give them to me and I'll put them in.

datalad/metadata/metadata.py
```diff
@@ -27,22 +28,68 @@
...
+
+
+def get_metadata_type(ds, guess=False):
+    """Return the metadata type(s)/scheme(s) of a dataset
+
+    Parameters
+    ----------
+    ds : Dataset
+      Dataset instance to be inspected
+    guess : bool
```
mih: Guessing is expensive, as it will happen all the time, not just on aggregate. I really really cannot see a case where the creator of a dataset does not know what kind of metadata it contains, or where a guess would yield a more accurate result. If anything, guessing should be removed completely.

datalad/metadata/metadata.py
```diff
@@ -97,8 +144,209 @@ def _get_key(k):
     return k
 
 
+#
+# common manipulator functions
+#
```
mih: #timewillcome

datalad/metadata/metadata.py
```diff
@@ -97,8 +144,209 @@ def _get_key(k):
...
+def _init(db, spec):
+    for k, v in spec.items() if spec else []:
+        if k not in db:
+            db[k] = v
```
mih: ;-) I made a mental note about your mental note...

datalad/metadata/metadata.py
```diff
@@ -97,8 +144,209 @@ def _get_key(k):
...
+        try:
+            pmod = import_module('.{}'.format(nativetype),
+                                 package=parsers.__package__)
+        except ImportError as e:
+            lgr.warning(
+                "Failed to import metadata parser for '%s', "
+                "broken dataset configuration (%s)? "
+                "This type of native metadata will be ignored: %s",
+                nativetype, ds, exc_str(e))
+            continue
```
mih: I agree

datalad/metadata/metadata.py
```diff
@@ -97,8 +144,209 @@ def _get_key(k):
...
+            native_meta = pmod.MetadataParser(ds).get_global_metadata()
+        except Exception as e:
+            lgr.error('Failed to get native metadata ({}): {}'.format(
+                nativetype, exc_str(e)))
+            continue
+        if not native_meta:
+            continue
+        if not isinstance(native_meta, dict):
+            lgr.error(
+                "Metadata parser '%s' yielded something other than a dictionary "
```
mih: Wouldn't hurt, indeed.

datalad/metadata/metadata.py
yarikoptic: "creator" is usually a crawler.  Less manual interaction I would have to have -- the better.  In some datasets collections (e.g. crcns) type of metadata is pretty much "hard coded" so I could set it up.  For others, even with openfmri, some datasets might still be in old openfmri layout, some in BIDS.  I could assume that any new dataset comes with BIDS layout but I would not count on it.  My point is that indeed in many cases I could specify it, but it would be nice to "guess" otherwise.

yarikoptic: moreover whatever dataset was type X metadata, at some point might change to Y type and hardcoded type might not longer be "valid".  sure thing we could detect that dataset is no longer of that type, and crash to change manually.

yarikoptic:
As I have outlined in https://github.com/datalad/datalad/issues/921  I have already used quite a bit a `distribution-restrictions` meta data variable in some datasets (raiders, attention, ...) to mark non-defaced anatomicals.  I also introduced it to heudiconv for `sourcedata/` etc

mih:
@yarikoptic thanks for the push. All errors seem legit now. I will sort them all out. But I think I need feedback/help with the mocked test that is failing. The respective error is in the top-level comment.

mih:
Errr?

mih:
I respectfully decline your offer....

datalad/metadata/metadata.py
mih: Fixed.

datalad/metadata/aggregate.py
```diff
@@ -49,16 +163,21 @@ class AggregateMetaData(Interface):
...
-    _no_eval_results = True
-
     _params_ = dict(
         dataset=Parameter(
             args=("-d", "--dataset"),
-            doc="""specify the dataset to perform the install operation on.  If
-            no dataset is given, an attempt is made to identify the dataset
-            based on the current working directory and/or the `path` given""",
+            doc="""topmost dataset metadata will be aggregated into. All dataset
+            betwween this dataset and any given path will received updated
```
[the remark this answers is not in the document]
mih: Fixed

datalad/metadata/metadata.py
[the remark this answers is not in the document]
mih: What about this: Instead of guessing the metadata format for each and every metadata access, why not make the crawler guess, and re-guess each time it actually sees a dataset change.

datalad/metadata/metadata.py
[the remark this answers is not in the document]
mih: Fixed.

datalad/metadata/aggregate.py
```diff
@@ -12,31 +12,434 @@
...
+        # NOTE: this will also ignore datalad's native dataset-global metadata
+        # rationale: the metadata still describes the dataset content, so
+        # even if it changes, the description changes, but not the content
+        # it is describing -> ref commit should be unaffected
+        list(exclude_from_metadata) + subds_relpaths)
+
+    if not relevant_paths:
+        return None
+
+    try:
```
bpoldrack: I guess the entire `try ... except` should be a method of `GitRepo` (`get_last_commit_from_files()` or sth)

datalad/metadata/aggregate.py
```diff
@@ -45,187 +448,193 @@ class AggregateMetaData(Interface):
...
+
+        # at this point we have dumped all aggregated metadata into object files
+        # somewhere, we know what needs saving, but having saved anything, and
+        # we know about the states of all aggregated dataset in the DB
+        # what remains to do is to update all dataset, so they have there own copy
+        # of aggregated metadata and update their respective aggregate.json with
+        # info on what states we just aggregated from
+
+        # first, let's figure out what dataset need updating at all
+        # get adjencency info of the dataset tree spanning the base to all leave dataset
```
bpoldrack: Typo: leave vs. leaf

datalad/interface/common_cfg.py
```diff
@@ -207,4 +207,9 @@
                'text': 'Specifies the repository version for git-annex to be used by default'}),
         'type': EnsureInt(),
     },
+    'datalad.metadata.nativetype': {
+        'ui': ('question', {
+               'title': 'Native dataset metadata scheme',
+               'text': 'Set this label to engage a particular metadata extraction parser'}),
```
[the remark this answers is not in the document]
mih: There is no way to solve this with one sentence other then "read the source Luke", at least ATM.

datalad/interface/common_opts.py
```diff
@@ -231,6 +231,23 @@
...
+    metavar='MODE',
+    doc="""merge procedure to use when a dataset provides
+    native metadata in some format. Such a dataset has to
+    indicate the type of native metadata via its
+    configuration setting ``datalad.metadata.nativetype``.
+    Multiple different types of metadata are supported. Merging
+    is performed in the order in which they are configured.
+    Custom DataLad metadata always takes precedence over
+    native metadata. Merge procedure modes are semantically
+    identical to the corresponding manipulation arguments.
```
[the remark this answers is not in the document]
mih: Please make a suggestion.

datalad/metadata/aggregate.py
```diff
@@ -12,31 +12,434 @@
...
+    Info on the metadata objects is placed into a DB dict under the
+    absolute path of the dataset whose metadata was aggregated.
+
+    Parameters
+    ----------
+    agginto_ds : Dataset
+    aggfrom_ds : Dataset
+    db : dict
+    merge_native : str
+      Merge mode.
```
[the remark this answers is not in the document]
mih: and some actual description

datalad/metadata/aggregate.py
mih: Make an issue and I'll assign yourself ;-)

datalad/metadata/aggregate.py
mih: Fixed.

datalad/metadata/aggregate.py
```diff
@@ -12,31 +12,434 @@
...
+from datalad.metadata.metadata import _get_containingds_from_agginfo
+from datalad.distribution.dataset import datasetmethod, EnsureDataset, require_dataset
+from datalad.support.param import Parameter
+from datalad.support.constraints import EnsureStr
+from datalad.support.constraints import EnsureNone
 from datalad.support.exceptions import CommandError
-from ..log import lgr
-from . import get_metadata, metadata_filename, metadata_basepath, is_implicit_metadata
+from datalad.support.annexrepo import AnnexRepo
+from datalad.log import lgr
```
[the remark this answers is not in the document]
mih: Not sure where this came from. Anyways, fixed, thanks.

mih:
Note, in 0980afe I said: "entire diff is discarded", that should read "entire diff of search.py is discarded"

datalad/metadata/aggregate.py
bpoldrack: Created an issue. And assigned myself, of course ;-)

datalad/interface/common_opts.py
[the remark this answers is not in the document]
bpoldrack: Well, am I right to assume that the arguments you are talking about are those of `datalad metadata`?

datalad/metadata/metadata.py
```diff
@@ -264,163 +733,231 @@ def __call__(
...
+                    elif ap.get('type', None) == 'dataset' and not ap['path'] == ds_path:
+                        # this is an available subdataset, will be processed in another
+                        # iteration
+                        continue
+                    else:
+                        (query_ds if DIRECT_QUERY else query_agg).append(ap)
+                # report directly available metadata
+                if query_ds:
+                    pass
+                   # # TODO RF to use code form `aggregate`
```
bpoldrack: How is `pass` an alternative to using code from aggregate? ;-)

From reading I've no idea what is to be done here. If you're sure to not forget about this piece - I'm fine with it though.

datalad/metadata/metadata.py
```diff
@@ -430,47 +967,94 @@ def __call__(
...
+                        ds_paths,
+                        reset=reset,
+                        add=add,
+                        init=init,
+                        remove=remove,
+                        purge=purge,
+                        # we always go recursive
+                        # XXX is that a good thing? But how to otherwise distinuish
+                        # this kind of recursive from the one across datasets in
+                        # the API?
```
bpoldrack: May be it could be distinguished by renaming the parameter for `AnnexRepo.set_metadata` to `traverse`?

bpoldrack:
Don't care for the failures of those builds (`DATALAD_REPO_VERSION=`)::

https://travis-ci.org/datalad/datalad/jobs/276569623
https://travis-ci.org/datalad/datalad/jobs/276569627

I'm looking into it. The failing annex calls make clear, what it is about, but I'm wondering why this didn't happen when I introduced those builds. However, I'll care for that today.

datalad/metadata/metadata.py
mih: It is not about the name, it is about the inability to distinguish recursion from traversal when the respective boundaries are invisible.

datalad/metadata/metadata.py
mih: This code can never be reached. I am keeping it for now, until I know that the approach works in full.

datalad/metadata/metadata.py
```diff
@@ -97,8 +162,391 @@ def _get_key(k):
...
+                # the specific match within the containing dataset
+                # normpath() because containing_ds could be `curdir`
+                path=normpath(opj(ds.path, containing_ds, fpath)),
+                # we can only match files
+                type='file',
+                metadata=metadata,
+                **kwargs)
+
+            # loop over all records, check if path matched and merge properties
+            for pathexp in contentmeta:
```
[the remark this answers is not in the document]
mih: True, this is one of quite a number of optimization opportunities. However, this is not among the things that decide between slow and fast in the current implementation, hence TODO -- like all the others.

datalad/metadata/metadata.py
```diff
@@ -97,8 +162,391 @@ def _get_key(k):
...
+    else:
+        # not a direct hit, hence we find the closest
+        # containing subdataset (if there is any)
+        containing_ds = sorted(
+            [subds for subds in sorted(info)
+             if rpath.startswith(_with_sep(subds))],
+            # TODO os.sep might not be OK on windows,
+            # depending on where it was aggregated, ensure uniform UNIX
+            # storage
+            key=lambda x: x.count(os.sep), reverse=True)
```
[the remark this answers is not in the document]
mih: All paths in this code are coming from path annotation -- there should be no mixture.

datalad/interface/common_opts.py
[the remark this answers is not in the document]
mih: Indeed, this was written when this option was unique to `metadata()`

datalad/metadata/metadata.py
bpoldrack: Good point. I missed that.

datalad/metadata/metadata.py
bpoldrack: Ok.

datalad/metadata/metadata.py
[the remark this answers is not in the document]
bpoldrack: Sure. But may be mark it ;-)

datalad/metadata/metadata.py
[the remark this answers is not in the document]
bpoldrack: Well, I do see that, but then - this is kind of an implicit assumption about path annotation (and about `info`). If code changes overtime an implication like this one is hard to discover, I guess.

mih:
We are inching ever closer to a usable state, please take a look and give your feeback.

bpoldrack:
Note: I just pushed c55aae282d78de3b44fe1e83d4173ef5f46b9b0f

We can now specify a search for keys, which include colons, like:

`datalad search dc:creator:poldrack type:file`

@datalad/developers : Please check that commit for possible flaws, you python regex pros. :-)
Note, that the original regex, that I simply replace with that call, is: `r"(?P<text>\w+|[*]):"`

bpoldrack:
Please note, that the failure in https://travis-ci.org/datalad/datalad/jobs/284569763 is marked as `known_failure_v6` already in PR #1870

mih:
"allowed failure" test run .21 is passing. I guess we should make it "not allowed to fail again".

mih:
@bpoldrack thanks for the note, merged master and cherry-picked that change.

bpoldrack:
> "allowed failure" test run .21 is passing. I guess we should make it "not allowed to fail again".

Nope. Some of the current failures in those build don't ALWAYS happen. For different reasons this is true for both of "probing" ones ATM.

mih:
Just a note, quick glance at the recent changes gives me a bad feeling. I cannot give details now, hoping to get back to this later this month.

bpoldrack:
Okay, so @yarikoptic: No pushes for now - let's get to a mergeable state. If you have fixes to provide, may be make them a PR against this one, so I can see @mih approved them. Otherwise I fear we never get to merge this beast.
However, it would be nice if you can trigger the buildbots. :-)

bpoldrack:
@mih: Set it "on hold", so we know to not merge yet. Whenever you get back just remove it.

datalad/metadata/search.py
```diff
@@ -181,15 +181,14 @@ def _get_search_schema(ds):
...
+                #set.add(', '.join(i for i in v) if isinstance(v, (tuple, list)) else v)
                 # and perform the mapping to the current one in here
                 count = 0
                 uk = k
                 while uk in definitions:
                     count += 1
                     uk = '{}_{}'.format(k, count)
-                ds_defs[k] = uk
-                k = uk
+                ds_defs[k] = k = uk
```
mih: TBH: I don't like this kind of diff. It is functionally irrelevant and requires more thinking when reading (are you really sure you know the order of evaluation in compound code like this all the time?).

datalad/metadata/search.py
```diff
@@ -181,15 +181,14 @@ def _get_search_schema(ds):
                 # might be different when some aggregated metadata was
                 # generated with an old version of datalad
                 # in this case we should actually load the old vocabulary
-                set.add(', '.join(i for i in v) if isinstance(v, (tuple, list)) else v)
+                #set.add(', '.join(i for i in v) if isinstance(v, (tuple, list)) else v)
```
mih: Not sure what state of my brain can be held responsible here, but I think this should be `v = (.. if ... else v)` to avoid non-string/non-dict values in key definitions later on.

mih:
@bpoldrack @yarikoptic 

First test on real-world data. This is our DICOM parser RF'ed to yield the identical metadata for all files in an image series (about 50 tags). In this dataset are 36k files across 25 image series. Bottom line: 98k XZ-compressed from 72MB of massively redundant raw JSON for 36k files. This is pretty close to the simulation results.

```
mih@meiner /tmp/study_20140724_095117357 (git)-[master] % datalad aggregate-metadata
[INFO   ] Aggregate metadata for dataset /tmp/study_20140724_095117357 
[INFO   ] Attempting to extract DICOM metadata from 36401 files 
[INFO   ] Update aggregate metadata in dataset at: /tmp/study_20140724_095117357 
aggregate_metadata(ok): /tmp/study_20140724_095117357 (dataset)                                          
[INFO   ] Attempting to save 6 files/datasets 
save(ok): /tmp/study_20140724_095117357 (dataset)
action summary:
  aggregate_metadata (ok: 1)
  save (ok: 1)
datalad aggregate-metadata  437,00s user 5,58s system 91% cpu 8:02,10 total
mih@meiner /tmp/study_20140724_095117357 (git)-[master] % ll .datalad/metadata/objects/3e5
total 112K
-rw-rw-r-- 1 mih mih  98K Okt 22 20:48 cn-2c811493893ca23c9ffddf4988137
-rw-rw-r-- 1 mih mih 8,8K Okt 22 20:47 ds-2c811493893ca23c9ffddf4988137
mih@meiner /tmp/study_20140724_095117357 (git)-[master] % xz -d -c .datalad/metadata/objects/3e5/cn-2c811493893ca23c9ffddf4988137 | wc -l
36401
% xz -d -c .datalad/metadata/objects/3e5/cn-2c811493893ca23c9ffddf4988137 > d.dummy               
mih@meiner /tmp/study_20140724_095117357 (git)-[master] % ll d.dummy 
-rw-rw-r-- 1 mih mih 72M Okt 22 20:52 d.dummy
```

yarikoptic:
Great! How long did it take to load it for indexing?

mih:
@yarikoptic Not so fast with the young horses...

mih:
@yarikoptic @bpoldrack all TODOs are dealt with. The last remaining bit is to decide whether we want to apply XZ compression to dataset-global metadata aggregates too? It is now the case that it could be much larger in file size than the actual content metadata. With automagicIO functional, there is very little benefit of having these files in Git -- apart from the ability to use git-diff, which is really more of a debugging feature than something users would be doing a lot.

bpoldrack:
Note, that the remaining PY3 failure should be solved in PR #1917 .

Re dataset-global metadata: 
I'd go for compression of that one, too.

From my point of view the only thing to consider when having all of that in annex is whether we wnat to have an option for `install` to auto-get metadata related files right away, so you can say "I want those and I want to be able to access metadata while not having a connection to auto-get it later on".
But that doesn't even need to be part of this PR.

mih:
I don't think that pulling metadata by default is needed. But once we can test datalad's behavior without interfering with user config, we should have a config option that causes this behavior.

datalad/metadata/parsers/bids.py
```diff
@@ -145,9 +168,9 @@ def yield_participant_info(fname):
                 val = row[k]
                 if hk is None:
                     hk = 'comment<{}>'.format(normk)
-                elif hk == 'dlp_bids:sex':
-                    val = sex_label_map.get(row[k].lower(), None)
+                if hk == 'comment<sex>':
```
yarikoptic: btw, why those fields from participant are "comment"? could they be at least somewhat more specific, e.g. `comment<participant:sex>` in case if we are doomed to use a `comment` one

datalad/metadata/parsers/bids.py
```diff
@@ -102,27 +98,54 @@ def _get_dataset_metadata(self):
...
+        # now go over all files in the dataset and query pybids for its take
+        # on each of them
+        for f in self.paths:
+            md = {}
+            try:
+                md.update(
+                    {'bids:{}'.format(k): v
+                     for k, v in bids.get_metadata(opj(self.ds.path, f)).items()})
+            except ValueError:
+                lgr.debug('no BIDS metadata for %s in %s', f, self.ds)
```
yarikoptic: please include actual exception into the message -- `exc_str` would be of help.  

Also it feels like there could be many reason BIDSLayout could emit some ValueError, which wouldn't really mean that there is no BIDS metadata there.

datalad/metadata/parsers/bids.py
mih: Comment is anything that is nowhere defined. The content of these files is pretty much completely unregulated, hence we have no idea what label means what. We can add any prefix though -- it is just a comment.

datalad/metadata/parsers/bids.py
mih: Sadly, pybids is not very specific wrt to this error.

datalad/metadata/parsers/bids.py
```diff
@@ -8,59 +8,45 @@
...
 
-# BIDS parser metadata definitions (dlp_bids:)
 vocabulary = {
-    "dlp_bids": {
-        'def': 'TODO',
-        'descr': 'Vocabulary of the DataLad BIDS metadata parser',
-        'type': vocabulary_id},
     # characteristics (metadata keys)
-    "dlp_bids:age(years)": {
+    "bids:age(years)": {
```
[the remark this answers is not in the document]
mih: There is no real vocabulary, but we claim that BIDS specifies all the keys that are used in its JSON files. The 'bids:' prefix is defined in the file you were looking at too.

datalad/crawler/nodes/annex.py
```diff
@@ -1326,7 +1326,7 @@ def _finalize(data):
 
             if aggregate:
                 from datalad.api import aggregate_metadata
-                aggregate_metadata(dataset=self.repo.path, guess_native_type=True)
+                aggregate_metadata(dataset='^', path=self.repo.path)
```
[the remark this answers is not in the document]
mih: I do not understand the question. All datasets in between are populated. There is no change in behavior that I am aware of. This merely adjusts the call to use this symbol instead of having to track paths -- which the crawler is not capable of.

mih:
For all of openfmri, using pybids to pull out file-based metadata, plus our own handling of `participants.tsv`. Indexing is in serial fashion due to #1927 
```
datalad aggregate-metadata -r  1068,76s user 54,22s system 96% cpu 19:28,43 total
datalad search female --reindex  176,52s user 37,28s system 100% cpu 3:31,83 total
[INFO   ] Search index contains 96926 documents 
[INFO   ] Query completed in 0.116760015488 sec. Reporting max. 20 top matches. 
% du -sh .git/datalad/search_index
54M	.git/datalad/search_index
% du -sh .datalad/metadata
1,3M	.datalad/metadata
% tree .datalad/metadata |wc -l
243
% tree .datalad/metadata | head -n5
.datalad/metadata
├── aggregate_v1.json
└── objects
    ├── 01
    │   ├── cn-35bf2e17af5b545f8b505badb0705a
```

mih:
Hate me, or love me...

PR #1643, opened 2017-07-17
BF(workaround): use patched wrapt disabling its extensions

yarikoptic:
We wouldn't be able to release like that (we could patch debian packages for wrapt but still -- that is not a solution) but at least meanwhile we could bring tests under control until https://github.com/GrahamDumpleton/wrapt/issues/98 is resolved

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -178,7 +178,7 @@ script:
   # Verify that setup.py build doesn't puke
   - python setup.py build
   # Run tests
-  - PATH=$PWD/tools/coverage-bin:$PATH $NOSE_WRAPPER `which nosetests` $NOSE_OPTS -v -A "$NOSE_SELECTION_OP(integration or usecase or slow)" --with-doctest --doctest-tests --with-cov --cover-package datalad --logging-level=INFO $TESTS_TO_PERFORM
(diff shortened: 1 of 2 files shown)


mih:
travis doesnt like it somehow

yarikoptic:
I guess -- outdated pip not supporting that syntax for url with a branch
I thought though -- if we agree to go with such a workaround, I could decorate import and disable import of those extensions while importing wrapt so we could use with a stock wrapt without my patch...

pity that upstream is silent ATM :-/

mih:
what impact would disabling that extension have on which aspects? I guess there is a reason why it exists. In any case, if we can somehow disable it for or test runs (only, for now). I am in favor, pretty much regardless of the impact.

yarikoptic:
effect probably would be only in slight performance impact. in our case wrapt'ed interfaces are not the bottlneck so I do not expect actually any notable effect on that.
originally I think extension is developed to speed things up and make it usable for functions where speed would matter... otherwise dunno

ok -- I will try to update pip and see if we could make it install that way

yarikoptic:
ok -- tests are passing!  I will merge this one into master for now and then master into one of the segfaulting PRs to see how it helped

PR #1645, opened 2017-07-17
DOC: little cleaning of gettingstarted.rst

chrhaeusler:
Hey wizards,

I've read the doc a little and and as a first attempt revised the gettingstarted.rst.

What I have done:
- added the command to install git-annex; the official debian repositories offer an outdated version so I consider it worthwhile to mention it and how to get a current version.
- The usual case in the doc is to give a mysterious line of code and then explaining it. It would mimic the regular train of thought better to first make clear a problem and its solution in plain English and _then_ providing the line of code followed by detailed technical description. In this case, the reader gets primed on what he will read as code and it avoids the reader jumping back and forth between code and explanation.

If required, I can revise the whole doc similarly.

Winter is coming...
Chris

### DIFF
--- a/docs/source/gettingstarted.rst
+++ b/docs/source/gettingstarted.rst
@@ -12,19 +12,19 @@ Installation
 
 Datalad is a Python package and can be installed via pip_, which is the
 preferred method unless system packages are available for the target platform
-(see below)::
+(see below). To automatically install datalad and all its software dependencies
+type::
 
   pip install datalad
 
 .. _pip: https://pip.pypa.io
 
-This will automatically install all software dependencies necessary to provide
-core functionality. Several additional installation schemes are supported
-(e.g., ``publish``, ``metadata``, ``tests``, ``crawl``)::
+Several additional installation schemes are supported (``[SCHEME]`` can be e.g.
+``publish``, ``metadata``, ``tests`` or ``crawl``)::
 
-  pip install datalad[SCHEME]
-
-where ``SCHEME`` can be any supported scheme, such as the ones listed above.
+  pip install datalad [SCHEME]
+  
+.. cool, but why should I (or a first-time reader) even bother about the schemes?
 
 In addition, it is necessary to have a working installation of git-annex_,
 which is not set up automatically at this point.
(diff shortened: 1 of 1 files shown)


docs/source/gettingstarted.rst
```diff
@@ -39,6 +39,11 @@ is to enable the NeuroDebian_ repository, and to install datalad as a system
 package::
 
   sudo apt-get install datalad
+  
+A current version of git-annex (as also provided by the NeuroDebian_ 
+repository) can be installed by typing::
+
+  sudo apt-get install git-annex
```
yarikoptic: yeah -- we need up to date one... should be `git-annex-standalone` from neurodebian

yarikoptic: actually hold on -- whenever you 'apt-get install datalad' -- it should install needed version of annex automagically... if not -- then it is a bug in packaging which we should fix but I think it should be alright, isn't it?

chrhaeusler: At first, I used "pip install datalad" as recommended in the doc and consequently did not add the neurodebian repository before. 
Then, when installing git-annex ("sudo apt-get install git-annex") without having added the neurodebian repository I ran into an outdated version provided by the official debian repository. So I added the Neurodebian repository and did "sudo apt-get install git-annex-standalone" (and not just "sudo apt-get install git-annex"; my bad!)

mih:
Thanks for the PR. I support the change to "America first, code second" (although I have to confess that I personally prefer it the other way round -- but I am not the primary target audience).

docs/source/gettingstarted.rst
```diff
@@ -12,19 +12,19 @@ Installation
...
 
-This will automatically install all software dependencies necessary to provide
-core functionality. Several additional installation schemes are supported
-(e.g., ``publish``, ``metadata``, ``tests``, ``crawl``)::
+Several additional installation schemes are supported (``[SCHEME]`` can be e.g.
+``publish``, ``metadata``, ``tests`` or ``crawl``)::
 
-  pip install datalad[SCHEME]
-
-where ``SCHEME`` can be any supported scheme, such as the ones listed above.
```
mih: I would even say that this "expert info", and we should have a clearly separate "I am new here, and I am not a python expert" intro section. Not intermingling everything. But this can be done later.

docs/source/gettingstarted.rst
mih: This mixes instruction for installation on Debian, with the general installation instructions. `apt-get install datalad` is completely sufficient. This bug is that you had the impression that `pip install` is a good way to start on Debian -- an impression we should not make (at least in this way). I would probably not start with general instructions special ones afterwards, but reverse this: First: install on Debian, install on ..., and then if we got nothing for your platform -> `pip install`

PR #1649, opened 2017-07-19
DOC: Basic Principles -> API Principles revised

chrhaeusler:
Hey,

I edited the API principles'  first part. There seems to be no big difference but I think the subtle changes make reading a lot easier for users being naive to the command line, git etc.

...gonna wrap my head around the "Commands install vs get" section which did not intuitively make sense to me the first time reading it. 

P.S.: I added explicit examples for using the https and ssh protocol but did not provide any real address (``datalad install ###example####``) due to sheer ignorance (e.g. available datasets)

### DIFF
--- a/docs/source/basics.rst
+++ b/docs/source/basics.rst
@@ -75,10 +75,11 @@ Here are some common examples in command line notation:
     install canonical superdataset (alone, without subdatasets) in a
     `datasets.datalad.org/` subdirectory under the current directory
 ``datalad install -r ///openfmri``
-    install openfmri superdataset with a collection of all datasets available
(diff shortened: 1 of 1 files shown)


yarikoptic:
note that tests have failed for incorrect slightly formatting

```
Warning, treated as error:

/home/travis/build/datalad/datalad/docs/source/basics.rst:67: (WARNING/2) Title underline too short.

```

mih:
Before merging the `###example###` placeholder should become an actual URL. This would be a candidate:

`https://github.com/psychoinformatics-de/studyforrest-data-phase2.git`

PR #1652, opened 2017-07-21
ENH: create --text-to-git to establish .gitattributes so that text file to git

yarikoptic:
It became a frequent practice for me to create such a .gitattributes file nearly for many datasets
I was creating locally for practice etc.  Even though there is `no_annex` plugin, it is not really visible (now it will be in `create --help`) and ATM lacks even option for such mode. 
And once again -- I think extensibility of plugins is great but most common operations should have clear and concise options available

### DIFF
--- a/datalad/distribution/add.py
+++ b/datalad/distribution/add.py
@@ -141,6 +141,16 @@ class Add(Interface):
             as it inflates dataset sizes and impacts flexibility of data
             transport. If not specified - it will be up to git-annex to
             decide, possibly on .gitattributes options."""),
+        to_annex=Parameter(
+            args=("--to-annex",),
+            action='store_false',
+            dest='to_git',
+            doc="""flag whether to force adding data to Annex, instead of
+        git.  It might be that .gitattributes instructs for a file to be
+        added to git, but for some particular files it is desired to be
+        added to annex (e.g. sensitive files etc).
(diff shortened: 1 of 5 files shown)


mih:
I dont mind the option of create. But I do not like the further complication of the AnnexRepo constructor. Its semantics are confusing enough. Why not add the same logic to the no_annex (or even a new) plugin and just call it in create?

mih:
Another issue: we have "togit" and "noannex". I was thinking for a while to standardize on saying that something will not go into the annex. Reason: annexed content will also be registered in Git. In pro knows that, but it is not at all obvious to newbies. I find it more precise to say no_annex, dont_annex, ... I dont care which particular phrase, but I do care about the direction.

yarikoptic:
Agree about noannex, so `--text-no-annex`

I can move it out from AnnexRepo if that would help ;-)

yarikoptic:
Refactored to "--text-no-annex" and moved to `create`.  Also added `--to-annex` (for consistency with current `--to-git`) which in Python interface wouldn't need to be a dedicated option -- value is stored to `to_git`, which defaults to None but could be True or False with the corresponding behavior (was already implemented even that way ;) ).  Seems to work

yarikoptic:
world will be a better place when this will be inplace, so I will merge for now ;)

PR #1662, opened 2017-07-26
ENH: allow for "recursive" flag for remove (needed while crawling s3 where prefix is a directory)

yarikoptic:
ran into while crawling openneuro


yarikoptic:
I will owe @mih even more lines of ununittested code

PR #1663, opened 2017-07-26
WiP: Borrow pieces from NumPy to run the tests without needing entire numpy for that purpose

yarikoptic:
I have failed to pick out exactly how to invoke bloody nose.run so it works also outside of the code base, so decided to just copy those pieces into our code base and trim connections to NumPy so we could just still use their runner for now.
Needed so we could test 'as installed' datalad without needing numpy

might abandon it altogether so do not merge

note that anyways - `datalad test` doesn't work if datalad is built using nuitka for execution as a standalone (`--exe` mode) since nose would not be able to find datalad module


yarikoptic:
ran already in deployments where could not run tests since numpy wasn't there... will merge for now

Issue #1666, opened 2017-07-26
prep dataset for nih workshop


yarikoptic:
1. ds000114 (latest version)
2. we need derivatives/fmriprep  as obtained from https://openneuro.org/datasets/ds000108/versions/00001  (disregard 108 not 114, @chrisfilo swears it is the one)
   - should be added using URL backend since ATM zip balls aren't reproducible
   - complication also that annex fails to verify their presense http://git-annex.branchable.com/bugs/fails_to_verify_presence_via_http_while_wget_fetches_it_just_fine/?updated
3. add/extract (zip) under derivatives/fmriprep subdataset
4. duplicate to become available from http://datasets.datalad.org/?dir=/workshops/nih-2017
5. @satra would add ANTS transforms (if not already under fmriprep outputs)

chrisfilo:
We'll fixt that URL.

satra:
@yarikoptic - the ants transforms are available in fmriprep outputs

yarikoptic:
@chrisfilo FWIW
https://openneuro.org/datasets/ds000108  gives me now "You are not authorized to view this dataset" and trying to get to https://openneuro.org/datasets/ds000108/versions/00001 keeps "Loading" spinwheel rolling forever.
Thought that you have fixed it already, but https://openneuro.org/datasets/ds000114/versions/00001 leads to "Dataset not found"  and attempt to go to https://openneuro.org/datasets/ds000114/ bounces back to the main page of openneuro.org

In that DropBox folder @satra shared url to, tried to access one of the .html files in the fmriprep -- but got "400 Error"

chrisfilo:
Thanks for reporting @yarikoptic - looking into this.

Not sure which URL @satra shared with you.

satra:
@chrisfilo - it's a dropbox url with the data so that we can have individual URLs for each file.

chrisfilo:
https://openneuro.org/datasets/ds000108/versions/00001 should work now.
Still looking into other issues.

chrisfilo:
Actually I don't think you are authorized to see https://openneuro.org/
datasets/ds000108 - why did you expect to access it? Is there a link to
https://openneuro.org/datasets/ds000108 anywhere on openneuro.org?

yarikoptic:
[...]

PR #1672, opened 2017-07-29
sketch of a script/doc to describe how datasets for nih workshop to be made

yarikoptic:
a possible one for #1666  but may be we would end up using @satra 's ;)


yarikoptic:
not worth it atm as is

Issue #1674, opened 2017-07-30
web ui incorrectly lists subdatasets in subdirs at the top level


yarikoptic:
see e.g. http://datasets.datalad.org/?dir=/workshops/nih-2017/ds000114 where 'freesurfer' and 'fmriprep' should be under derivatives.  Reason is incorrect assumption(s) in the code where first traversal of the dataset file system is done ignoring submodules, and metadata for each directory is stored, but then submodules processed in addition separately (within `ds_traverse`) and their entries extend root directory metadata structure (`nodes` field)

yarikoptic:
looking at it ATM it seems that the issue was already resolved

PR #1675, opened 2017-07-30
Handle subdatasets in subdirectories correctly in the web browser listing frontend

yarikoptic:
Attempt to fix up #1674 ... coming out a bit more than just a bugfix, since to get better clue on what is going on in that code had to do some RFing

But in that light I think there are still missing points, since blind attempt to regenerate meta-data for
http://datasets.datalad.org/?dir=/workshops/nih-2017/ds000114 didn't fully worked out

But it worked correctly while having old meta-data inplace ;)  so I guess it is just missing some additional bit.

- [x] fix it up: 
     - atm seems to traverse subdatasets directories with only -a flag
     - for some reason assigns the same tags from superdataset into subdatasets: http://datasets.datalad.org/?dir=/workshops/nih-2017/ds000114/derivatives  (may be it is just a web ui bug? didn't check yet metadata itself) -- I guess this one was fixed since I am not observing it
- [x] reenable disabled checks
- [x] verify correct operation on ds000114 in typical invocation in the hook but also may be in -r mode (which so

The code there is messy and needs a huge RF and might even at some point be melded with new metadata handling since lots in common, but not now.

I have rerendered web metadata for http://datasets.datalad.org/?dir=/labs/haxby/attention and http://datasets.datalad.org/?dir=/workshops/nih-2017 so they look proper.  For that updated datalad on the webserver from 0.7.0-95-g551efff2 to this PR 0.9.1-291-ge61f0841


yarikoptic:
forgot about this one and thought it was already merged... doing that now

PR #1684, opened 2017-08-03
BFs for minor issues: access removed repo, unicode symbols in ls output

yarikoptic:


yarikoptic:
I will merge tomorrow unless someone says otherwise or does it before I do

mih:
User info message is not covered in tests....

mih:
It seems this is now causing master breakage -- strange.

yarikoptic:
oh - missed that it cause some chaos... sorry... will fix it tomorrow

PR #1686, opened 2017-08-04
BF: make multiple invocations for get/copy if too many files listed

### LINKED ISSUE #1683: on OSX get with large number of (resolved to be ffetched) files could fail
as reported  as an act of the datalad-hate:

```shell
~/Documents/datasets.datalad.org $ datalad get -r openfmri/ds000030
[INFO   ] Obtaining <Dataset path=/Users/xxx/Documents/datasets.datalad.org/openfmri/ds000030>  recursively 
[INFO   ] Getting 1 items of dataset <Dataset path=/Users/xxx/Documents/datasets.datalad.org/openfmri/ds000030> ...
[INFO   ] Actually getting 9112 files
[ERROR] [Errno 7] Argument list too long [subprocess.py:_execute_child:1335] (OSError)
```

apparently on OSX limits on cmdline length are smaller which ruins our approach to requesting filenames here by specifying them in the command line.  The best way would be to convert this (and also in copy_to (where it is not implemented yet: http://git-annex.branchable.com/todo/--batch_for_copy/?updated) to use --batch'ed execution ;-)

yarikoptic:
Ideally we should just use --batch mode for get, but it is N/A for copy ATM, so decided to implement this "do in groups of files" for get/copy

Closes #1683

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -52,6 +52,8 @@ from datalad.utils import on_windows
 from datalad.utils import swallow_logs
 from datalad.utils import assure_list
 from datalad.utils import _path_
+from datalad.utils import generate_chunks
+from datalad.utils import CMD_MAX_ARG
 from datalad.cmd import GitRunner
 
 # imports from same module:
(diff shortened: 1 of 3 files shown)


yarikoptic:
since life should be better with this than without, merging

PR #1690, opened 2017-08-07
Travis: Make Travis work on dev-master again

bpoldrack:
Trying to figure out, what needs to be done ...

Update:
Still waiting for some Travis jobs to complete, but main issue (segfault) seems to be gone. What's still failing is the nipype example. @mih, do we already have a solution for that one? Can't recall ...


mih:
Really? OSX and nd90 still segfault.

bpoldrack:
Re OSX nd90:
Don't see them easily from home. But for Travis it's gone as oppossed to the first commit. However, the buildbots' setup isn't changed, so it's likely it's still there. They would need to also disable the wrapt extensions, which in addition to that env var needs @yarikoptic 's fork of wrapt.

yarikoptic:
Heh, buildbots do not have that workaround for wrapt yet

bpoldrack:
@yarikoptic : Can you make that happen for the buildbots?
@mih: Any remembrance of the nipype issue?

mih:
I think I remeber that I had no idea what was wrong.

bpoldrack:
Okay, so I try looking into it.

bpoldrack:
Update: Nipype example is failing to publish recursively in the end, due to a lot of subdatasets reported as 'impossible' since they are apparently absent. Need to check whether or not they should be there. May be it's just that we fail on 'impossible' where we should ignore it in that case.

bpoldrack:
Re Nipype example:
See #1691 .

bpoldrack:
Since the main issue (segfault) is solved for Travis and apparently for OSX (Thanks @yarikoptic ), I vote for merge. The nipype failure is a topic of its own, I think.

yarikoptic:
hm, on the first try I thought that the issue wasn't fixed on osx although I adjusted config... good that it passes -- I restarted nd90 now, should pass too I hope

bpoldrack:
nd90 passed - thanks!

But now I notice there's a PY3 issue in addition to the nipype example issue. Looking into it ...

bpoldrack:
Re PY3: Stupid me just misinterpreted Travis' output. Tests are passing in PY3, too, except for the nipype example.

bpoldrack:
So, it's all fine - leaving #1691 to @mih for now and merging ...

Issue #1691, opened 2017-08-08
Nipype example failing


bpoldrack:
As mentioned in PR #1690, I looked into the failure of the nipype example.
What is failing is the `datalad publish -r` command, called from within the modified superdataset from `datasets.datalad.org`. We added a new subdataset and want to publish the entire beast. The issue here is, that we didn't install all the other subdatasets of `///`. Actually it's working (meaning: it publishes what is  expected), but returns non-zero, since the recursive publishing of the superdataset yields `impossible` for all the not installed subdatasets.
We have no nice and obvious way to avoid that. `datalad publish` itself provides a `--missing=skip` option, but this one is for missing siblings to publish to, not for locally absent subdatasets. I guess we need an additional option here. Do you agree, @datalad/developers ?

An ugly approach to make the test pass would be to use `--on-failure=ignore`, but this is bad for a test, since we can't specify the failure we want to ignore. Furthermore it doesn't work in this case.

mih:
I think it should not even attempt to publish something that does not exist locally. it would be more or less incomprehensible to error on something that is invisible to a user. `publish -r` should only recursive into installed datasets. I consider this a bug the refactored `publish` that now uses path annotation. So it should be fixed in `dev-master`.

mih:
@bpoldrack can you isolate this issue and push a (failing) test, please? I'll take care of the actual fix.

bpoldrack:
Okay, will do. But one thing: It's not entirely invisible to the user. The mount points of those subdatasets are present.

mih:
Mountpoints are present by default (but not guaranteed) as empty dirs, no immediate indication that they are anything but an empty dir in the windows explorer ;-)

PR #1694, opened 2017-08-08
TST: Add a test for publishing recursively with absent subdatasets

bpoldrack:
Demonstrating #1691, and the fix.


bpoldrack:
LGTM

Let's wait for the tests.

PR #1706, opened 2017-08-11
ENH+TST: Save 50% of the time I spend on debugging test failures

mih:


mih:
Again, errors visible so far are due to breakage in master.

bpoldrack:
Nice! Thank you!

PR #1708, opened 2017-08-11
BF: install's default result filter to respect Dataset instances

### LINKED ISSUE #1701: Providing a dataset instance as `source` for `install` yields `None` and no error
Just a note for now, discovered when debugging merge conflicts. @bpoldrack can you investigate this, please?

bpoldrack:
Closes #1701

### DIFF
--- a/datalad/distribution/tests/test_install.py
+++ b/datalad/distribution/tests/test_install.py
@@ -46,6 +46,7 @@ from datalad.tests.utils import assert_false
 from datalad.tests.utils import ok_file_has_content
 from datalad.tests.utils import assert_not_in
 from datalad.tests.utils import assert_raises
+from datalad.tests.utils import assert_is_instance
 from datalad.tests.utils import assert_result_count
 from datalad.tests.utils import assert_status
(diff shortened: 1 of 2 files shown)


datalad/distribution/tests/test_install.py
```diff
@@ -270,6 +271,20 @@ def test_install_dataset_from_just_source(url, path):
...
+    origin = Dataset(src)
+    clone = install(source=origin, path=dst)
+
+    assert_is_instance(clone, Dataset)
+    ok_startswith(clone.path, dst)
+    ok_(clone.is_installed())
+    ok_(GitRepo.is_valid_repo(clone.path))
+    ok_clean_git(clone.path, annex=None)
+    assert_in('INFO.txt', clone.repo.get_indexed_files())
+
```
mih: With those tests we can REALLY be sure! Lines 280 + 284-5 should be more than sufficient IMHO ;-)

bpoldrack: Well, it should just result in the very same thing as in the test above. ;-)

PR #1716, opened 2017-08-14
Small bits around `uninstall`

mih:


mih:
All green, except for this NFS error which now seems to be uniformly happening everywhere, hence not particular to the present changes:

````
Creating config file /etc/exports with new version
Creating config file /etc/default/nfs-kernel-server with new version
 * Not starting NFS kernel daemon: no exports.
Processing triggers for libc-bin (2.19-0ubuntu6.13) ...
Processing triggers for ureadahead (0.100.0-16) ...
exportfs: localhost:/tmp/nfsmount_: Function not implemented
mount.nfs: Connection timed out
The command "if [[ "${TMPDIR:-}" =~ .*/nfsmount ]]; then echo "mkdir $TMPDIR"; mkdir -p "$TMPDIR" "${TMPDIR}_"; sudo apt-get install -y nfs-kernel-server; sudo exportfs -o rw localhost:/tmp/nfsmount_; sudo mount -t nfs localhost:/tmp/nfsmount_ /tmp/nfsmount; fi" failed and exited with 32 during .
Your build has been stopped.
````

File: datalad/distribution/dataset.py [hunk 1]
```diff
@@ -309,6 +310,10 @@ def get_superdataset(self, datalad_only=False, topmost=False):
           id), or (if False, which is default) - any git repository
         topmost : bool, optional
           Return the topmost super-dataset. Might then be the current one.
+        registered_only : bool, optional
+          Test whether any discovered superdataset actually contains the
+          dataset in question as a registered subdataset (as opposed to
+          just being located in a subdirectory without a formal relationship).
```
yarikoptic: I think I might not mind making true the default.

mih: Yes, that is up for debate: #1717

Issue #1699, opened 2017-08-09
datalad save ideally shouldn't explore entire hierarchy and rely on "git status" output may be?


yarikoptic:
save -r seems to traverse full hierarchy, which is slower than just to query git status first for what paths might be needed saving.

initially encountered when had only one dataset known to be modified but datalad was oldish so decided to test current master and 
```shell
$> datalad save -m "Added initial subdataset with sherlock for MIND-2017" -r
save(notneeded): /mnt/btrfs/datasets/datalad/crawl/workshops/nipype-2017/ds000114 (dataset)
save(notneeded): /mnt/btrfs/datasets/datalad/crawl/workshops/nipype-2017 (dataset)
save(notneeded): /mnt/btrfs/datasets/datalad/crawl/workshops/nih-2017/ds000114/derivatives/freesurfer (dataset)
save(notneeded): /mnt/btrfs/datasets/datalad/crawl/workshops/nih-2017/ds000114/derivatives/fmriprep (dataset)
save(notneeded): /mnt/btrfs/datasets/datalad/crawl/workshops/nih-2017/ds000114 (dataset)
save(notneeded): /mnt/btrfs/datasets/datalad/crawl/workshops/nih-2017 (dataset)
save(notneeded): /mnt/btrfs/datasets/datalad/crawl/workshops/mind-2017/sherlock (dataset)
save(notneeded): /mnt/btrfs/datasets/datalad/crawl/workshops/mind-2017 (dataset)
save(notneeded): /mnt/btrfs/datasets/datalad/crawl/workshops (dataset)
action summary:
  save (notneeded: 9)
(dev)2 14833.....................................:Wed 09 Aug 2017 04:38:01 PM EDT:.
(git)smaug:/mnt/btrfs/datasets/datalad/crawl/workshops[master]git
$> git status
On branch master
nothing to commit, working directory clean
(dev)2 14834.....................................:Wed 09 Aug 2017 04:38:32 PM EDT:.
(git)smaug:/mnt/btrfs/datasets/datalad/crawl/workshops[master]git
$> datalad --version
datalad 0.8.0.dev1
```

so -- it does get through the entire sub-hierarchy one by one, although could just see that all is clean according to fast "git status"

mih:
Just an idea, without much thinking: It should be possible to amend the AnnotatePaths call in `save` with a `since=HEAD` argument. This should have the desired effect automatically. Internally, this will call `diff` which figures out the paths that were changed.

[...]

PR #1719, opened 2017-08-16
Misc RF+ENH for `save`

mih:
- [x] demo test and fix for #1699 
- [x] rename `files` argument of `save` to `path` for consistency with any other command


bpoldrack:
LGTM

Since the only error in Travis is the failing setup for the NFS build - merging.

Issue #1650, opened 2017-07-21
[wishlist] run/rerun tandem


yarikoptic:
Just discovered https://dataversioncontrol.com/ from the blog post https://medium.com/towards-data-science/how-to-version-control-your-machine-learning-task-cad74dce44c4  and thought that it is a nice feature to have indeed and should be very easy to implement!

1. `run cmd args` would 
   - verify that we are in a clean state
   - go through args and check if any point to file under annex control, and if that one is not present locally -- download it
   - execute the `cmd args`
   - `datalad -m "RUN <cmd> <args>" save` all changes

2. `rerun` (repro in DVC) probably with options for either to

 - `--steps=[all,N]` to rerun all commands or just last N,
- `--use-current-state` -- to use current state (thus using recent code version, I think it is what DVC does) or by default checkout previous states to check  if external changes impact our results

and with that it would

  - traverse git history (so we do not need to store anything else) to establish which files were output (thus changed) in any 'RUN' commit, and which were input, so we could build DAG on the fly, collect commands needed to be reran
  - execute commands, while either checking out those commits in case of no `--use-current-state` or just replacing input files with symlinks to their keys as discovered by traversing the history

Even without `rerun` I think it could be quite neat!  paired with the power of `git reset [--hard]` and `git commit --amend` I think I might use it pretty much for anything I would run in that repository ;) And then rerun could come handy as well.  There could be additional options

- `--branch=name` to rerun in a specific branch, may be even cherry picking (or even rerunning?) commits which do not have anything to do with this file.  This way we could later compare, similar to testkraut, at which step diversion happened etc while reproducing a full history

yarikoptic:
Implementation details to keep in mind:
- storing command in the commit message should use e.g. `repr` on the list of arguments to guarantee being able to get back to it in case of spaces in the filenames etc

mih:
I think the `rerun` case might come with lots of corner cases. However, the `run` case is straightforward, and coincidentally exactly the way that I use to construct my datasets. I would give this a go, if nobody beats me to it.

mih:
Thought: it can be tough to figure out which data files a command will need from the command line. We could go testkraut-style and which the actual syscalls to intercept anything that aims for something underneath a dataset. But that also comes with its own problem.

That being said, even without the auto-download aspects (for which we could eventually have multiple methods) it would still be useful. In my original datasets I typically have all data available anyways, hence for these cases I'd loose none of the benefits.

yarikoptic:
[...]

PR #1722, opened 2017-08-16
Towards a resolution of #1650 (run a command and capture its impact)

mih:
- [x] basic implementation of `run` with auto-commit, and recording of command, PWD, and exit code
- [x] basic tests, 100% coverage
- [x] waiting for #1733 

This PR does not implement support for command that do in-place modification of dataset files. This is waiting for a resolution of #1424.

Ping #1650

### DIFF
--- a/datalad/interface/run.py
+++ b/datalad/interface/run.py
@@ -18,8 +18,6 @@ import re
 from argparse import REMAINDER
 from os.path import join as opj
 from os.path import curdir
-from os.path import pardir
-from os.path import relpath
 from os.path import normpath
 
 from datalad.cmd import Runner
@@ -31,19 +29,14 @@ from datalad.interface.results import get_status_dict
 from datalad.interface.common_opts import save_message_opt
 
 from datalad.support.constraints import EnsureNone
-from datalad.support.constraints import EnsureStr
-from datalad.support.constraints import EnsureChoice
 from datalad.support.exceptions import CommandError
-from datalad.support.exceptions import InsufficientArgumentsError
 from datalad.support.param import Parameter
 
-from datalad.distribution.dataset import Dataset
 from datalad.distribution.dataset import require_dataset
 from datalad.distribution.dataset import EnsureDataset
 from datalad.distribution.dataset import datasetmethod
 
 from datalad.utils import get_dataset_root
-from datalad.utils import getpwd
 from datalad.tests.utils import ok_clean_git
 
 lgr = logging.getLogger('datalad.interface.run')
--- a/datalad/interface/run.py
+++ b/datalad/interface/run.py
@@ -1,0 +1,176 @@
+# emacs: -*- mode: python; py-indent-offset: 4; tab-width: 4; indent-tabs-mode: nil -*-
+# ex: set sts=4 ts=4 sw=4 noet:
+# ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+#
+#   See COPYING file distributed along with the datalad package for the
+#   copyright and license terms.
+#
+# ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+"""Run log arbitrary comands and track how they modify a dataset"""
+
+__docformat__ = 'restructuredtext'
+
+
+import logging
+import json
+
+from argparse import REMAINDER
+from os.path import join as opj
+from os.path import curdir
+from os.path import pardir
+from os.path import relpath
+
+from datalad.cmd import Runner
+
+from datalad.interface.base import Interface
+from datalad.interface.utils import eval_results
+from datalad.interface.base import build_doc
+from datalad.interface.results import get_status_dict
+
+from datalad.support.constraints import EnsureNone
+from datalad.support.constraints import EnsureStr
+from datalad.support.constraints import EnsureChoice
+from datalad.support.exceptions import CommandError
+from datalad.support.exceptions import InsufficientArgumentsError
+from datalad.support.param import Parameter
+
+from datalad.distribution.dataset import Dataset
+from datalad.distribution.dataset import require_dataset
+from datalad.distribution.dataset import EnsureDataset
+from datalad.distribution.dataset import datasetmethod
+
+from datalad.utils import get_dataset_root
+from datalad.utils import getpwd
+from datalad.tests.utils import ok_clean_git
+
+lgr = logging.getLogger('datalad.interface.run')
+
+
+@build_doc
+class Run(Interface):
+    """Run and explain
+    """
+    _params_ = dict(
+        cmd=Parameter(
+            args=("cmd",),
+            nargs=REMAINDER,
+            metavar='SHELL COMMAND',
+            doc="command for execution"),
+        dataset=Parameter(
+            args=("-d", "--dataset"),
+            doc="""specify the dataset to query.  If
+            no dataset is given, an attempt is made to identify the dataset
+            based on the input and/or the current working directory""",
+            constraints=EnsureDataset() | EnsureNone()),
+        # TODO
+        # --list-commands
+        #   go through the history and report any recorded command. this info
+        #   could be used to unlock the associated output files for a rerun
+        # --rerun #
+        #   rerun command # from the list of recorded commands using the info/cmd
+        #   on record
+        # --message
+        #   prepend default message and replace default short summary
+    )
+
+    # TODO running a command outside a dataset should be supported, but it will prevent
+    # a reliable rerun, and we do not want to record absolute CWD paths...
+
+    @staticmethod
+    @datasetmethod(name='run')
+    @eval_results
+    def __call__(
+            # it is optional, because `rerun` can get a recorded one
+            cmd=None,
+            dataset=None):
+        if not dataset:
+            # act on the whole dataset if nothing else was specified
+            dataset = get_dataset_root(curdir)
+        ds = require_dataset(
+            dataset, check_installed=True,
+            purpose='tracking outcomes of a command')
+        # not needed ATM
+        #refds_path = ds.path
+        lgr.debug('tracking command output underneath %s', ds)
+        try:
+            # base assumption is that the animal smells superb
+            ok_clean_git(ds.path)
+        except AssertionError:
+            yield get_status_dict(
+                'run',
+                ds=ds,
+                status='impossible',
+                message='unsaved modifications present, cannot detect changes by command')
+            return
+
+        if not cmd:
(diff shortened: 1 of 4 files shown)


mih:
Rerun demo

````
# command that generates a random number on each execution
% datalad run 'od -vAn -N4 -tu4 < /dev/urandom > some_random'
add(ok): some_random (file)                                                   
save(ok): /tmp/mike (dataset)
action summary:
  add (ok: 1)
  save (ok: 1)
% datalad run --rerun ; cat some_random
unlock(ok): some_random (file)
add(ok): some_random (file)                                                   
save(ok): /tmp/mike (dataset)
action summary:
  add (ok: 1)
  save (ok: 1)
  unlock (ok: 1)
 4079372669
% datalad run --rerun ; cat some_random
unlock(ok): some_random (file)
add(ok): some_random (file)                                                   
save(ok): /tmp/mike (dataset)
action summary:
  add (ok: 1)
  save (ok: 1)
  unlock (ok: 1)
  146996588
````

**Comment by mih** [comment]
From my POV this is done.

File: datalad/interface/run.py [hunk 1]
```diff
@@ -0,0 +1,268 @@
...
+                # to be able to do it recursively
+                raise CommandError(code=cmd_exitcode)
+
+        lgr.info("== Command exit (modification check follows) =====")
+
+        # ammend commit message with `run` info:
+        # - pwd if inside the dataset
+        # - the command itself
+        # - exit code of the command
+        run_info = {
```
yarikoptic: might be worth making it an ordered dict to guarantee the consistent order of the fields

datalad/interface/run.py
```diff
@@ -0,0 +1,268 @@
...
+            if rel_pwd.startswith(pardir):
+                lgr.warning('Process working directory not inside the dataset, command run may not be reproducible.')
+                rel_pwd = None
+
+        # anticipate quoted compound shell commands
+        cmd = cmd[0] if isinstance(cmd, list) and len(cmd) == 1 else cmd
+
+        # TODO do our best to guess which files to unlock based on the command string
+        #      in many cases this will be impossible (but see --rerun). however,
+        #      generating new data (common case) will be just fine already
```
yarikoptic: mental note to myself -- apparently `sed -i` operates on a copy and then just replaces original one, so on my simple attempts it just worked without explicit unlocking

datalad/interface/run.py
```diff
@@ -0,0 +1,268 @@
...
+    The executed command, the process working directory (PWD; if it is inside
+    the given dataset), and the command's exit code will be recorded in the
+    commit message.
+    """
+    _params_ = dict(
+        cmd=Parameter(
+            args=("cmd",),
+            nargs=REMAINDER,
+            metavar='SHELL COMMAND',
+            doc="command for execution"),
```
yarikoptic: eventually (now?) we might want to add `--automagicio` (or smth like that) mode which would just run the command (ATM would work only for python scripts) through out automagicIO support (`python -m datalad`).  Sure thing would be applicable only when the command/script is a python script, but imho would improve utility.  Eventually #1739 it should automagically unlock depending on the operation.  And may be even further down the line could run any process while monitoring syscalls....

datalad/interface/run.py
```diff
@@ -0,0 +1,268 @@
...
+        dataset=Parameter(
+            args=("-d", "--dataset"),
+            doc="""specify the dataset to query.  If
```
yarikoptic: `to query`? probably should be adjusted on what would it mean for `run` command if dataset is provided.  Feels like If I specify dataset, make sense to run command in that dataset top directory

datalad/interface/run.py
```diff
@@ -0,0 +1,268 @@
...
+                if r.get('type', None) == 'file' and \
+                        r.get('state', None) in ('added', 'modified'):
+                    r.pop('status', None)
+                    to_unlock.append(r)
+            if to_unlock:
+                for r in ds.unlock(to_unlock, return_type='generator', result_xfm=None):
+                    yield r
+        else:
+            # not a rerun, figure out where we are running
+            pwd = getpwd()
```
yarikoptic: re pwd, see my comment above on use/role of the `dataset` argument.

datalad/interface/run.py
```diff
@@ -0,0 +1,268 @@
...
+        try:
+            # base assumption is that the animal smells superb
+            ok_clean_git(ds.path)
+        except AssertionError:
+            yield get_status_dict(
+                'run',
+                ds=ds,
+                status='impossible',
+                message='unsaved modifications present, cannot detect changes by command')
+            return
```
yarikoptic: ATM would not be able to run whenever I have some files unlocked (e.g. to be modified):
```shell
(git-annex)hopa:/tmp/ds[master]git
$> datalad unlock 123                                 
unlock(ok): 123 (file)
$> datalad run "sed  -e 's,9,1,g' 123 | sponge > 123"
run(impossible): /tmp/ds (dataset) [unsaved modifications present, cannot detect changes by command]
```
```shell
$> git status
On branch master

	typechange: 123
```
making it not usable for any scenario where command should modify/(often as well overwrite if write is done without removing previous one first) an already existing file

datalad/interface/run.py
mih: Yes, I want that. Not yet sure how to approach this the best way.

datalad/interface/run.py
mih: copy/paste error. `dataset` arg is critical for `rerun`. I chose not to auto-use the top of the dataset to run commands, because sometimes it might make sense to run them in a subdir (which would be recorded and used on reruns).

datalad/interface/run.py
```diff
@@ -0,0 +1,268 @@
...
+            args=("-d", "--dataset"),
+            doc="""specify the dataset to query.  If
+            no dataset is given, an attempt is made to identify the dataset
+            based on the input and/or the current working directory""",
+            constraints=EnsureDataset() | EnsureNone()),
+        message=save_message_opt,
+        rerun=Parameter(
+            args=('--rerun',),
+            action='store_true',
+            doc="""re-run the command recorded in the last saved change (if any).
```
mih: TBH, I limited the flexibility for now, because it turned out that there was enough to fix to make it run like this ;-) In general, 100% ACK.

datalad/interface/run.py
mih: This is behavior intentional and desired. I want to be sure that the recorded changes are due to the command run, and nothing else. But I can modified unlocked files, without Git reporting this change:

````
mih@meiner /tmp/mike (git)-[master] % git st
On branch master
mih@meiner /tmp/mike (git)-[master] % touch mike
mih@meiner /tmp/mike (git)-[master] % datalad add mike
add(ok): /tmp/mike/mike (file):00, ?B/s]                                                                 
save(ok): /tmp/mike (dataset)
action summary:
  add (ok: 1)
  save (ok: 1)
mih@meiner /tmp/mike (git)-[master] % datalad unlock mike
unlock(ok): mike (file)
mih@meiner /tmp/mike (git)-[master] % git st
On branch master

	typechange: mike

mih@meiner /tmp/mike (git)-[master] % echo "this" > mike
mih@meiner /tmp/mike (git)-[master] % git st
On branch master

	typechange: mike

````

On reruns files get unlocked automatically.

datalad/interface/run.py
mih: See me reply above.

datalad/interface/run.py
mih: True.

datalad/interface/run.py
mih: Actually: The point is really about the issue of getting a command to run in the right place when using the Python API. So what about this: When the dataset arg comes in as a dataset instance, we are running in the dataset root. If it comes in as a string, we go with CWD.

datalad/interface/run.py
mih: Will do via key-sort on `dumps`

datalad/interface/run.py
yarikoptic: So you are saying that "run" cannot be used with commands causing in-place modification of files and that is intentional.  Since we aren't really to track the provenance, I feel that this just limits applicability,e.g. I had some defaced anatomicals but decided to use **another** command this time to deface them.  Command will need to override a file.  I would need to explicitly remove it first, commit and then use "datalad run"... or may be I want to "datalad run" a command which incrementally grows a file -- wouldn't be possible at all... I do not see why we should limit "intentionally" against such cases.

> But I can modified unlocked files, without Git reporting this change:
rright.  but as always -- not sure if we could avoid all possibilities to shoot yourself in a foot.

I feel that if we do not somehow automagically unlock, we should `--allow-dirty`... if really keen on protocolling it, could even add a field to the log json to state which files were dirty when running the command

datalad/interface/run.py
yarikoptic: such distinctions would not fit into my memory... why not just to stay consistent with the rest of the API, which currently says that e.g. all paths, if `-d` provided, are from the top of that dataset.  So it only makes it all consistent to use for `run` to dictate to run the command from the top of the dataset.  Also would make it easier to e.g. sweep through all datasets which have consistent layout (e.g. BIDS) to cause some modification in all of them (e.g. `for ds in ds*; do datalad run -d $ds do_something_to_bids_dataset path_inside; done`)

datalad/interface/run.py
yarikoptic: but how it would be different, besides pretty much just changing which commit to analyze?

datalad/interface/run.py
mih: No, I am not saying that. During the last hangout we agreed that we need an explicit argument to indicate which files need unlocking in this case, as it is otherwise impossible to discover this reliably, as you pointed out. This functionality is simply not there yet. The reason for that is that in-place modification requires #1424 to be solved, in order for a re-run of such a scenario to make any sense.

datalad/interface/run.py
mih: So be it. Less code, OK with me.

datalad/interface/run.py
mih: Turn that question into an answer, and we would be one step further ;-)

yarikoptic: ah, so it is difficult ATM to get the commit message for a commit given some committish?  just that you have done all the diff, it would be fresh'ier in your mind.  But something like just running `git show HEAD | sed -n -e '/Do not change lines below/,/Do not change lines above/p'` could pretty much do it (replace HEAD with desired committish)

mih:
Requested changes are pushed. Automagic IO and in-place modification support can come via separate incremental PRs when the pre-conditions are fulfilled. I will merge when the tests go green.

mih:
Bingo!

PR #1724, opened 2017-08-17
Faster `add`, more reliable and capable `diff`

mih:


datalad/distribution/add.py
```diff
@@ -404,7 +437,10 @@ def __call__(
                     # in the dataset
                     ap = {k: v for k, v in res.items() if k != 'status'}
                     ap['staged'] = True
-                    to_save.append({k: v for k, v in res.items() if k != 'status'})
+                    # strip any status and state info (e.g. save will refuse to save
+                    # stuff that is marked state='untracked'
```
bpoldrack: To be honest, I don't entirely get it. If it's marked 'untracked' why not change this, now that it was added, but pretend we have no idea of the state at all?

datalad/distribution/create.py
```diff
@@ -261,26 +261,26 @@ def __call__(
...
                 return
-            except StopIteration:
-                # all good
-                pass
+
+        # TODO here we need a further test that if force=True, we need to look if
+        # there is a superdataset (regardless of whether we want to create a
+        # subdataset or not), and if that superdataset tracks anything within
+        # this directory -- if so, we need to stop right here and whine, because
+        # the result of creating a repo here will produce an undesired mess
```
bpoldrack: I think I recall, that it was possible to move anything tracked by a potential superdataset into that newly created subdataset. If the user says "force" shouldn't that be the desired result instead of failing?

datalad/interface/diff.py
```diff
@@ -65,12 +66,61 @@ def _translate_type(mode, ap, prop):
...
     elif stat.S_ISDIR(mode):
-        # not sure if this can happen at all
+        # not sure if this can happen at all, at least not in the tests...
         ap[prop] = 'directory'
     else:
         ap[prop] = 'file'
 
 
+def _get_untracked_content(dspath, report_untracked, paths=None):
+    cmd = ['git', '--work-tree=.', 'status', '--porcelain',
```
bpoldrack: I guess this call should be made available via *Repo.status(). Objections? Otherwise I'd make it an issue.

datalad/distribution/add.py
mih: We could do that. But at the moment there would be little gain, if any.

datalad/distribution/create.py
mih: How was that possible? It would mean we have a `move` implementation somewhere -- I would want to use that!!!

datalad/interface/diff.py
mih: I don't see why not. Please do!

datalad/distribution/add.py
bpoldrack: I agree. To me it's just about consistency. It might be hard to find that spot, if that unknown state later turns out to be troublesome.

datalad/distribution/create.py
bpoldrack: I think I did it at some point manually (not using `datalad`). I'll try to figure it out again/find some traces of what I did back then. But may be I'm wrong. I'll report back about that.

PR #1730, opened 2017-08-18
DOC: Demo of a scripted and auto-narrated screencast/asciicast

mih:
For now the asciicast part is most functional. With the setup described in the commit message, one can run this:

`tools/rec_asciinema docs/casts/seamless_nested_repos.sh`

Four solid asciicasts are included: #1744, #1745, #1746, #1748

in the root of the devtree, and get an asciicast JSON that is ready to go. A demo is here: https://asciinema.org/a/134579 (source is this:  docs/casts/seamless_nested_repos.sh)

There is also a conversion tool `tools/cast2script` that turns cast scripts into normal shell scripts with comments that can be executed with `set -e` to verify that everything intended to run actually runs.


mih:
I am not sure how the travis failure can be explained, given that this PR contain non-executed code only.

mih:
Failure was intermittent, apparently. Will merge once tests are complete.

PR #1735, opened 2017-08-21
Update DataLad logo; small win: 4K -> 1K

aqw:


yarikoptic:
I thought that project logo change had conditions attached! ;-)  now there would be 3 logos circling around (official, different one in docs, and this one)

PR #1741, opened 2017-08-21
BF: try to specify nfs mount in /etc/exportfs before installing NFS kernel server

yarikoptic:
The failure started to happen I believe due to precise migrating from precise to trusty.  Actually quite cool that the rest remained working as it should

This one is somewhat of a blind try


yarikoptic:
That seems has fixed it. Enjoy

PR #1750, opened 2017-08-23
ENH/BF: do not "reset" largefiles setting in cmdline while crawling if already defined in .gitattributes

yarikoptic:
Also adds a new function to GitRepo API to get "default" git attributes, by running `git check-attr --auto .`

With this crawler default pipelines should respect those settings set eg by `datalad create --text-no-annex`


mih:
In my condition, I will not be very useful as a reviewer, but I conclude that is has enough lines in the diff to be worthy! Signing off for today...

PR #1754, opened 2017-08-24
Datalad -> DataLad

aqw:
This /only/ changes strings and leaves the classname DataladAnnexCustomRemote untouched.

This does not take care of 'datalad' -> 'DataLad', as that will be much more
difficult to sift through (between all the commands and URLs).


aqw:
@mih Cool; thanks.

PR #1759, opened 2017-08-24
BF: Don't escape '$' for scripts generated by tools/cast2script

aqw:


mih:
This is wrong. In the generated script for the reproducible analysis the resulting script in `code/` will be broken, if this PR is merged. It'll look like this:

```sh
% cat code/brain_extraction.sh
# enable FSL
. /etc/fsl/5.0/fsl.sh

# obtain all inputs
datalad get 
# perform brain extraction
count=1
for nifti in ; do
  subdir="sub-000"
  mkdir -p 
  echo "Processing "
  bet  /anat -m
  count=1
done
```

Note the `cat << EOT` bit in the asciicast.

aqw:
@mih Ahh, that's my bad. I should have looked closer.

Closing the PR and will re-fix on the website.

---Alex

PR #1760, opened 2017-08-25
ENH: screencasts - generate .cmd.md file to be cut/pasted into description + possible to modify terminal size for "big" recording

### LINKED ISSUE #1755: Metadata search demo with current system
Would be great to have a quick asciicast that shows how to search for data on a freshly installed system. I know @yarikoptic has had a few snappy demo. Please post them! Thanks.

yarikoptic:
Here is an example of asciinema with the description containing the commands with links to the manpages help:  https://asciinema.org/a/vhCxdfvqqnNGDk6xQFdxwvOQh

now https://asciinema.org/~DataLad contains two asciinemas generated at two different resolutions. @aqw  do you think it would be possible easily to switch between two different asciinemas depending on desktop-vs-mobile rendering? ideally making it possible to choose one or another

edit: also now includes a basic search cast, rendered here: https://asciinema.org/a/134920 and which I will add in a second to datalad.org repo (Closes #1755)

### DIFF
--- a/docs/casts/basic_search.sh
+++ b/docs/casts/basic_search.sh
@@ -1,0 +1,33 @@
+full_title="Demo of basic datasets meta-data search using DataLad"
+#run "set -eu  # Fail early if any error happens"
+
+say "DataLad allows to aggregate dataset-level meta-data, i.e. data describing the dataset (description, authors, etc), from a variety of formats (see http://docs.datalad.org/en/latest/metadata.html for more info)."
+say 'In this example we will start from a point where someone who has not used datalad before decided to find datasets which related to "raiders" (after "Raiders of the Lost Ark" movie) and neuroimaging.'
+say 'As you will see below, upon the very first invocation of "datalad search" command, DataLad will need first to acquire aggregated meta-data for our collection of datasets available at http://datasets.datalad.org and for that it will install that top level super-dataset (a pure git repository) under ~/datalad:'
+( sleep 4; type yes; key Return; ) &
+run "datalad search raiders neuroimaging"
+
+say '"search" searches within current dataset (unless -d option is used), and if it is outside of any it would offer to search within the ~/datalad we have just installed'
+( sleep 4; type yes; key Return; ) &
+run "datalad search raiders neuroimaging"
+
+say "To avoid interactive question, you can specify to search within that dataset by using -d /// . And now let's specialize the search to not only list the dataset location, but also report the fields where match was found.  This time let's search for datasets with Haxby among the authors"
+run "datalad search -d /// -s author -R haxby"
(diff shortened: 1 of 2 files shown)


aqw:
@yarikoptic Just to make sure I'm understanding this correctly: are you planning to upload these 'casts to asciinema.org as well?

If so, I don't think that's a good idea. Having the same content in multiple places is not needed and only inflates the risk of having out-of-sync, out-of-date demos.

Without a strong, demonstrated need to the contrary, I think it is best to have one canonical source for these demos.

---Alex

mih:
Small comment on that (I have uploaded asciicasts targeting datalad.org to asciinema.org, too).

The issue with maintaining an "official" presence of datalad on that site is that we cannot keep the casts up-to-date. We could only replace a cast (when a new version comes out) by deleting the old one and posting a new one. This would invalidate the old one on asciinema.org and wherever it was shared. For datalad.org we aim at re-recording casts whenever necessary to reflect the current version. Linking/promoting/preferring the casts on our page will be more up-to-date and more stable, hence we should not establish a second "official" presence, IMHO. Or, at the very least, make very clear that the content there could be any level of outdatedness.

That being said, I plan to continue to upload casts on datalad to asciinema -- for quick one-off publications with no desire to keep them up-to-date over time. I will use my personal account for that.

And now I have to apologies for continuing a topic that is very off-topic given the title of the issue.

aqw:
@mih Sorry for dragging us off-topic.

@yarikoptic In response to your question, it is certainly possible to flip between the two sizes, however, there are real implications with that. And I don't think it's worth going down that rabbit-hole.

---Alex

yarikoptic:
re presence: indeed we wouldn't be able to maintain consistent with the website presence since can't replace yet in place -- but it is on their TODO, so eventually we could keep them.   In my perception outdated materials, either shared in personal or centralized account are outdated.  It is easier though to find/adjust in a single account than multiple (especially when having no access to it).  Meanwhile I will just add a statement to those under `~DataLad`:  

  You could find this and other DataLad up-to-date casts at http://www.datalad.org/features.html

Also, on the website we do not have yet have the lovely "just commands + timing + help script" representation.  IMHO it could be really handy for people to quickly familiarize with used commands. If only we could provide hyperlinks to each option, that would be just awesome (it should be possible todo, but probably not now)

re sizes:  yeah, I am converging that my approach (generating two casts) is too cumbersome.  In the majority of the cases (may be all ATM), we will be just fine by scaling "at run time" -- at least for asciinema website it is possible to provide desired size, so we could scale up any "small" one up, e.g. https://asciinema.org/a/134921?cols=120&rows=36 .  Only the pre-wrapped comments text would stay old way, and anything tty-width-based (progressbars).  But I think it is Ok.  So it would be really nice if on our website we could have an option to scale those up one way (implicit desktop-vs-mobile) or another (some icons with custom sizes)

PR #1766, opened 2017-08-26
BF: Proper data transfer handling for publication dependencies

### LINKED ISSUE #1763: Publication dependency handling broken
Now tested after data was copied for the main target, whereas dependencies should have received it first.

mih:
- [x] fixes #1763 by shifting existing code blocks around and adding a loop across dependency remotes for `_publish_data`
- [x] consolidate all data transfer functionality in `_publish_data` to allow for simple loops across remotes.
- [x] dedicated test for #1763
- [x] asciicast for "publish to GitHub" https://asciinema.org/a/135078
- [x] minor RF and test polish


mih:
Quick looks at test failure suggests that the fix is trivial....

PR #1770, opened 2017-08-30
NF: cast_live helper to guide interactively through casts

yarikoptic:
This helper will 
- start a new xterm in a geometry/resolution which should be ok for HD video casts on e.g. youtube
- "say" in the original terminal 
- wait before typing or executing any command in the original terminal
So it is jumping between two terminals (original with "Say"ed instructions) for you and the one which you could be possibly recording (e.g. with obs)

This way original cast could be used to guide interactive session while still allowing to interact in that terminal

### DIFF
--- a/tools/cast_bash.rc
+++ b/tools/cast_bash.rc
@@ -59,4 +59,4 @@ git config --global core.pager cat
 
 
 # cleanup at the end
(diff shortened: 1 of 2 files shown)


yarikoptic:
since no objections and proven to be in a functional and useful state, merging.  eventually we might want to absorb common "xterm settings" etc which we use now in two cast* scripts into a common "library"

PR #1771, opened 2017-08-30
DO NOT MERGE RF: (travis) Install wrapt straight from git master, disable a workaround

yarikoptic:
This is just to enable testing of suggested fixes etc, e.g. https://github.com/GrahamDumpleton/wrapt/issues/98#issuecomment-325963440

- [ ] ATM it is made to fail, and then when replicated, will attempt a patched version


yarikoptic:
I will close it for now, might be reopened later when we again hit wrapt issues

yarikoptic:
reincarnated this PR to test wrapt since Debian claims it was fixed:
https://salsa.debian.org/openstack-team/python/python-wrapt/commit/cb4d98747244879eb4544df4c741209624df17cd
but our issue is still open: https://github.com/GrahamDumpleton/wrapt/issues/98

yarikoptic:
bleh -- Debian changelog was misleading!  It is just that new version has our workaround not really a fix!  but let's test nevertheless - may be it somehow went away

yarikoptic:
may be some time it would come handy... but as time shown:
- unlikely to see proper fix in wrapt
- seems to not fail in real world applications even without no-extensions workaround which we use during testing

PR #1772, opened 2017-08-31
BF: Report missing SSH on debug level only

bpoldrack:


bpoldrack:
Well, `codecov/patch` target is pointless here, from my point of view.

PR #1774, opened 2017-08-31
ENH: Don't fail on fetch if a (special) remote doesn't provide an URL

bpoldrack:
Annex special remotes may have no URL. Don't fail when trying to fetch from such remotes. Just do nothing and log on debug level.

### DIFF
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -1484,10 +1484,16 @@ class GitRepo(RepoInterface):
 
         fi_list = []
         for rm in remotes_to_fetch:
-            fetch_url = \
-                rm.config_reader.get('fetchurl'
-                                     if rm.config_reader.has_option('fetchurl')
-                                     else 'url')
+            from six.moves.configparser import NoOptionError
+            try:
+                fetch_url = \
+                    rm.config_reader.get('fetchurl'
+                                         if rm.config_reader.has_option('fetchurl')
+                                         else 'url')
+            except NoOptionError:
+                lgr.debug("Remote %s has no URL", rm)
+                return []
+
             if is_ssh(fetch_url):
                 ssh_manager.get_connection(fetch_url).open()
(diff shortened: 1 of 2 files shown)


datalad/support/gitrepo.py
```diff
@@ -1484,10 +1484,16 @@ def fetch(self, remote=None, refspec=None, progress=None, all_=False,
...
-            fetch_url = \
-                rm.config_reader.get('fetchurl'
-                                     if rm.config_reader.has_option('fetchurl')
-                                     else 'url')
+            from six.moves.configparser import NoOptionError
+            try:
+                fetch_url = \
+                    rm.config_reader.get('fetchurl'
+                                         if rm.config_reader.has_option('fetchurl')
+                                         else 'url')
```
mih: Isn't `config_reader` an instance of our own `ConfigManager`? If so, you could just pass a default argument to avoid having to catch an exception.

datalad/support/gitrepo.py
```diff
@@ -1484,10 +1484,16 @@ def fetch(self, remote=None, refspec=None, progress=None, all_=False,
...
+            except NoOptionError:
+                lgr.debug("Remote %s has no URl", rm)
```
mih: lower case 'l'

datalad/support/gitrepo.py
bpoldrack: It's not. This still is a `GitConfigParser` from GitPython (property of `git.Remote` "rm").
But should actually work nonetheless and furthermore `ConfigManager` should be used instead, I guess.
So, yes - agree.

bpoldrack: Actually, this is wrong ;-)
Just noticed, that we both misread that line. It's getting a different variable via that `else`, not a different value of the same variable. So it's not about default value.

mih:
Gracias @bpoldrack !

PR #1789, opened 2017-09-04
Change testing for direct mode and V6 to exclusion lists

### LINKED ISSUE #1562: Use exclusion list to disable select tests from direct mode testing
ATM we test a little bit of code for compliance with direct mode, using hand-picked tests. IMHO this should change. There is quite a chunk of usable functionality in direct mode, but from the test perspective it doesn't look like it. I think we should switch to a negative list, disabling only test that fail (possibly for a known reason). The current approach will or will not allow for the status quo to get worse -- we just won't know -- hence it isn't suitable for making progress towards direct mode compatibility.

bpoldrack:
Closes #1562 

Approach changed to use decorators `skip_direct_mode` and `skip_v6` for tests currently failing with those builds. In addition there's a comment `# FIXME` for all of them.

- [x] create initial lists to represent status quo
- [x] mark concerned tests with decorators which are raising `SkipTest`
- [x] properly set up Travis to make use of the above for DM as well as V6

### DIFF
--- a/datalad/crawler/nodes/tests/test_annex.py
+++ b/datalad/crawler/nodes/tests/test_annex.py
@@ -7,6 +7,7 @@
 #
 # ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
 
+from datalad.tests.utils import skip_direct_mode
 from os import listdir
 from os.path import join as opj, exists, lexists, basename
 from collections import OrderedDict
@@ -49,6 +50,7 @@ def test_annexificator_no_git_if_dirty(outdir):
 
 @with_tempfile(mkdir=True)
 @with_tempfile()
+@skip_direct_mode
 def test_initiate_dataset(path, path2):
     dataset_path = opj(path, 'test')
     datas = list(initiate_dataset('template', 'testdataset', path=dataset_path)())
@@ -166,6 +168,7 @@ def _test_annex_file(mode, topdir, topurl, outdir):
     assert_equal(output[0]['datalad_stats'], ActivityStats(files=1, add_git=1))
 
 
+@skip_direct_mode
 def test_annex_file():
     for mode in ('full', 'fast', 'relaxed',):
         yield _test_annex_file, mode
@@ -213,6 +216,7 @@ def _test_add_archive_content_tar(direct, repo_path):
         assert_false(annex.repo.dirty)
 
 
+@skip_direct_mode
 def test_add_archive_content_tar():
     for direct in (True, False):
         yield _test_add_archive_content_tar, direct
@@ -222,6 +226,7 @@ def test_add_archive_content_tar():
 @with_tempfile(mkdir=True)
 @with_tree(tree={'file': 'load'})
 @serve_path_via_http
+@skip_direct_mode
 def test_add_dir_file(repo_path, p, topurl):
     # test whenever file becomes a directory and then back a file.  Should all work!
     annex = Annexificator(path=repo_path, auto_finalize=False)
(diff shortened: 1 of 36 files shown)


datalad/crawler/pipelines/tests/test_balsa.py
```diff
@@ -124,6 +126,8 @@ def test_smoke_pipelines():
 @serve_path_via_http
 @with_tempfile
 @with_tempfile
+@skip_direct_mode  #FIXME
```
yarikoptic: note that if the entire module of tests failes (probably for issues not to be fixed in that module per se), might be worth just skip it at the module level, e.g. how it is done already for scrapy
```python
try:
    import scrapy
except ImportError:
    raise SkipTest("Needs scrapy")
```

bpoldrack: The entire point is to do it on a per-test level, so we are able to easily ensure we don't worsen the status quo and we can fix that stuff less complex steps. Have a look at #1562.

datalad/crawler/nodes/tests/test_annex.py
```diff
@@ -49,6 +49,7 @@ def test_annexificator_no_git_if_dirty(outdir):
 
 @with_tempfile(mkdir=True)
 @with_tempfile()
+@skip_direct_mode
 def test_initiate_dataset(path, path2):
```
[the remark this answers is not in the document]
bpoldrack: What do you mean by "even"? ;-)
Come on - it's currently 91 out of 764. Admittedly it's not single digit yet, but it's not THAT bad, given that we didn't monitor anything but `datalad/tests` and `datalad/support`.

bpoldrack:
Re "more concise":
Nope. See the comment above and the issue to be closed by this PR. You miss the point, I guess. ;-)

bpoldrack:
If currently running tests are passing, this is ready to merge from my point of view.

yarikoptic:
Ok, per test it is
Seems need a bit more work, some tests still fail.
in the future would be a way to quickly enable all those tests to see which might have been fixed for?

datalad/crawler/pipelines/tests/test_balsa.py
mih: Yes, please keep in mind that these are steps towards not having anything fail. We need a quick way to say if there is anything worth fixing at the level of and when touching individual tests. Please keep individual test labeling.

mih:
@bpoldrack I do not follow the reasoning on NOT having an external switch to disable the decorators. My personal workflow is to have a pre-crafted way that I execute nosetests when working on the tests. I usually run a subset of the tests that are related to some functionality. The issues underlying the test failures are most likely not related to functionality that is explicitly tested in one or a few dedicated tests, but are more general violations of direct-mode or v6 assumption. Hence the proper way to run the tests is to run many of them at once. Your proposal would require to hand-edit individual tests, of the subset that I want to run. I don't see anyone of us actually doing that. A config switch would be much more suitable.

mih:
Maybe a central switch `datalad.tests.dont_skip_known_failures` would be appropriate.

bpoldrack:
Done. Just vice versa. There is `datalad.tests.skipknownfailures` now, which defaults to 'yes'.

mih:
Works for me.

yarikoptic:
Morning idea - add a switch/run to the matrix, where all those marked are expected to fail, so decorate would run them, catch exception, and if no exception (ie passes) then it fails. This way we could see right away what/when got fixed, and remove decorator

mih:
Ah, just saw @yarikoptic 's message. That would be a nice addition.

Issue #1400, opened 2017-03-20
Dataset class should get some basic documentation


yarikoptic:
currently completely empty!  http://docs.datalad.org/en/latest/generated/datalad.api.html#dataset

mih:
Not only that, but it also doesn't appear anywhere in the rendered docs anymore. Not even at the page linked above.

PR #1791, opened 2017-09-04
DOC: Minimal docs on Dataset, rendered in docs now (fixes gh-1400)

mih:


mih:
Hmm, so if ones add a comment line this counts as a miss in coverage...?

PR #1797, opened 2017-09-06
RF: AnnexRepo.add

bpoldrack:
Reduce unnecessary trials for git-add --dry-run in direct mode

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -1357,7 +1357,7 @@ class AnnexRepo(GitRepo, RepoInterface):
 
     @normalize_paths
     def add(self, files, git=None, backend=None, options=None, commit=False,
-            msg=None, dry_run=False,
+            msg=None,
             jobs=None,
             git_options=None, annex_options=None, _datalad_msg=False,
             update=False):
@@ -1376,9 +1376,6 @@ class AnnexRepo(GitRepo, RepoInterface):
           the list of files that were added, is created by default.
         backend:
         options:
-        dry_run : bool, optional
-          Calls git add with --dry-run -N --ignore-missing, to just output list
-          of files to be added
         update: bool
           --update option for git-add. From git's manpage:
            Update the index just where it already has an entry matching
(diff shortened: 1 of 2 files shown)


datalad/support/annexrepo.py
```diff
@@ -1413,95 +1410,107 @@ def add(self, files, git=None, backend=None, options=None, commit=False,
...
+                            "We can't use --dry-run when there are submodules in "
+                            "direct mode, because the internal call to git status "
+                            "fails. To be resolved by using (Dataset's) status "
+                            "instead of a git-add --dry-run altogether.")
+                        # fake the return for now
+                        return self._process_git_get_output(
+                            linesep.join(["'{}'".format(f) for f in files]))
+                    else:
+                        # unexpected failure
+                        raise e
```
yarikoptic: this entire exception handling is not even smoke tested according to codecov report

datalad/support/annexrepo.py
```diff
@@ -1357,7 +1357,7 @@ def _get_expected_files(self, files, expr):
 
     @normalize_paths
     def add(self, files, git=None, backend=None, options=None, commit=False,
-            msg=None, dry_run=False,
```
yarikoptic: FTR `dry_run` used to be used by crawler, but no longer ATM, so I don't think any code is using it any longer

bpoldrack: Yes. According to a git-grep and the tests, there's no code using it directly. Only the recursive calls to `AnnexRepo.add` were left. And in those cases that call isn't needed.

datalad/support/annexrepo.py
bpoldrack: Yes. And it was covered before.

The entire point of the RF'ing is to not go there in any "regular" case. Which is why that doesn't happen anymore in the tests. So in fact the coverage report confirms the RF'ing worked. ;-)

yarikoptic: 1. if not used -- should be removed???
2. I still see `git_options=` so some evil soul could even pass --dry-run?

bpoldrack: 1. Well, it still CAN happen. As noted in the comments this will be the case when we are in a non-direct-mode annex but there's a submodule in direct mode, which then is touched by that dry-run call. That would lead to the error. But this means, that some user intentionally switched a part of the hierarchy to direct mode. So, this isn't a "regular" case, where direct mode is enforced either due to a FS considered "crippled" by annex or because it was set to be the default for all annexes.

2. Yes. Why not? If someone thinks he is smarter than us and wants to pass certain options - he shall have the exception he deserves. ;-)
What are you suggesting? To not allow for that option to be passed or to catch the resulting exception the same way (it would affect the call in line 1488)

yarikoptic: I am just suggesting to have some test which covers this piece of code.  Since it could be triggered (1.) -- I guess you know how ;)  or could actually be done more efficiently with little mocking of GitRepo.add to trigger those exceptions which you expect to be thrown.

bpoldrack:
@yarikoptic : If tests are passing can be merged.

yarikoptic:
failed in v6, but not sure if not a sporadic failure... restarted
FTR

https://travis-ci.org/datalad/datalad/jobs/273377764
```

FAIL: datalad.support.tests.test_annexrepo.test_annex_ssh

  File "/home/travis/virtualenv/python3.4.6/lib/python3.4/site-packages/nose/case.py", line 198, in runTest

    self.test(*self.arg)

  File "/home/travis/build/datalad/datalad/datalad/tests/utils.py", line 848, in newfunc

    return func(*args, **kwargs)

  File "/home/travis/build/datalad/datalad/datalad/tests/utils.py", line 593, in newfunc

  File "/home/travis/build/datalad/datalad/datalad/tests/utils.py", line 763, in newfunc

    t(*(arg + (uri,)), **kw)

  File "/home/travis/build/datalad/datalad/datalad/tests/utils.py", line 763, in newfunc

    t(*(arg + (uri,)), **kw)

  File "/home/travis/build/datalad/datalad/datalad/support/tests/test_annexrepo.py", line 1110, in test_annex_ssh

    ok_(exists(socket_2))

AssertionError: None
```

PR #1804, opened 2017-09-07
BF: detect git repo root by presence of .git alone

### LINKED ISSUE #1803: get_dataset_root  fails to determine the root in workdir "clones"
wanted to make use of run 'in development', but couldn't (see below)...  First I thought that it is because of its anal nature to even detect untracked files which are ignored by .gitignore **but** then I looked at the report message -- apparently it ignored current "workdir" git repo and goes for its parent with .mrconfig setup:

```shell
(git)hopa:~datalad/datalad-master[bf-subdatasets-int]git
$> datalad run git-sedi 'subm 2' 2
run(impossible): /home/yoh/proj/datalad (dataset) [unsaved modifications present, cannot detect changes by command]
```

I think allowing for datalad commands to be usable even on workdir or plain git repositories, where no datasets etc stuff is needed, would be of advantage.

yarikoptic:
That is how it is done in other places throughout the code (we should
provide a helper), so no reason to make it special here and thuse
detect correctly root within submodules and workdirs which would have
just .git file or a symlink

Closes #1803

### DIFF
--- a/datalad/utils.py
+++ b/datalad/utils.py
@@ -1199,7 +1199,7 @@ def get_dataset_root(path):
     as the input argument. If no associated dataset exists, or the
     input path doesn't exist, None is returned.
     """
-    suffix = os.sep + opj('.git', 'objects')
+    suffix = os.sep + '.git'
     if not isdir(path):
         path = dirname(path)
     apath = abspath(path)


yarikoptic:
not sure why neither travis nor buildbot runs!

PR #1801, opened 2017-09-07
BF: allow submodules names be numbers and "bool" values

yarikoptic:
More work might be needed to replace other uses of config.get_value
where we do not want automatic casting to be done, but I looked around and didn't find obvious problematic places -- where we used it (also not easy to figure out since we have our own config.get_value) it seems to point to values which couldn't be that degenerate

PR #1805, opened 2017-09-08
BF+ENH: allow for submodules to have numeric or bool names

yarikoptic:
sits on top of #1801 

to test even more thoroughly.
- [x] requires a fix in GitPython (https://github.com/gitpython-developers/GitPython/issues/613). 
- [x] Fix is still pending upstream and thus pip/debian packages do not provide a fixed version yet.  Merged upstream, but needs more testing/release (done - release 2.1.6)


yarikoptic:
ok, this one seems to be now ok with the most recent release of GitPython and we should merge it (don't forget to update your installation of GitPython -- now available through neurodebians etc)

Issue #1817, opened 2017-09-10
Use '--' path separator in GitRepo._git_custom_command()?


mih:
This code https://github.com/datalad/datalad/blob/master/datalad/support/gitrepo.py#L1366

essentially appends given files to a command, but it does not inject a '--' separator. Shouldn't that be done in order to avoid name clashing with arguments and options?

yarikoptic:
Generally I see why not, but
- I remember seeing adding -- in the code manually, probably passed to this function
- so -- could be specific outside and this function shouldn't add any of there is -- before files

bpoldrack:
I'm not sure what approach you are in favor of, @yarikoptic.

I guess, Michael is right.

yarikoptic:
I didn't say that Michael is wrong. I think approach is sound too. I just made a note for what to look after while implementing it

PR #1820, opened 2017-09-11
RF: add '--' automatically within GitRepo_git_custom_command 

bpoldrack:
Just quickly addressing #1817

### DIFF
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -1364,15 +1364,16 @@ class GitRepo(RepoInterface):
         stdout, stderr
         """
 
+        # ensure cmd_str becomes a well-formed list:
         if isinstance(cmd_str, string_types):
-            if files and not cmd_str.strip().endwith(" --"):
+            if files and not cmd_str.strip().endswith(" --"):
                 cmd_str += " --"
-            cmd = shlex.split(cmd_str + " " + " ".join(files),
-                              posix=not on_windows)
+            cmd_str = shlex.split(cmd_str, posix=not on_windows)
         else:
             if files and cmd_str[-1] != '--':
                 cmd_str.append('--')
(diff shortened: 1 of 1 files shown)


datalad/support/gitrepo.py
```diff
@@ -1363,9 +1363,17 @@ def _git_custom_command(self, files, cmd_str,
...
-            if isinstance(cmd_str, string_types) \
-            else cmd_str + files
+
+        if isinstance(cmd_str, string_types):
+            if files and not cmd_str.strip().endwith(" --"):
+                cmd_str += " --"
+            cmd = shlex.split(cmd_str + " " + " ".join(files),
+                              posix=not on_windows)
+        else:
+            if files and cmd_str[-1] != '--':
```
yarikoptic: I would have
-Taken out appending files our of condition (it is a list, we know)
- composed command list with files
 - kept this check dedented, and checked not only the last one but if any has `--`

This would allow to catch when anyone specified `--` anywhere

bpoldrack: Okay. I agree.

datalad/support/gitrepo.py
```diff
@@ -1363,9 +1363,18 @@ def _git_custom_command(self, files, cmd_str,
...
+        # ensure cmd_str becomes a well-formed list:
+        if isinstance(cmd_str, string_types):
+            if files and not cmd_str.strip().endswith(" --"):
+                cmd_str += " --"
+            cmd_str = shlex.split(cmd_str, posix=not on_windows)
+        else:
+            if files and cmd_str[-1] != '--':
+                cmd_str.append('--')
+
+        cmd = cmd_str + files
```
yarikoptic: Not exactly / fully what I meant, but ok, better than before I hope ;-)

PR #1824, opened 2017-09-11
ENH: remove fixups

### LINKED ISSUE #1821: remove spits outs an error (but complets the act)
```shell
(git-annex)hopa:/tmp/ds123-clone[master]git-annex
$> datalad remove 123
drop(ok): /tmp/ds123-clone/123 (file)
remove(ok): 123
[ERROR  ] 'NoneType' object has no attribute 'endswith' [posixpath.py:join:70] (AttributeError) 
(dev) 2 11128 ->1.....................................:Mon 11 Sep 2017 03:47:09 PM EDT:.
(git-annex)hopa:/tmp/ds123-clone[master]git-annex
$> ls -l 123
ls: cannot access '123': No such file or directory
```

edit 1: actually doesn't complete the act -- fails before saving... fixing it

### LINKED ISSUE #1822: we need --nosave to remove as well
as discovered while playing with #1821-- for consistent (with `add`) API / expected behaviour

yarikoptic:
- ENH: Added  --nosave and --message to  remove cmd   Closes #1822
- BF: make remove not fail if operated without a specified dataset  Closes #1821

### DIFF
--- a/datalad/distribution/remove.py
+++ b/datalad/distribution/remove.py
@@ -30,6 +30,8 @@ from datalad.interface.annotate_paths import annotated2content_by_ds
 from datalad.interface.base import Interface
 from datalad.interface.common_opts import if_dirty_opt
 from datalad.interface.common_opts import recursion_flag
+from datalad.interface.common_opts import nosave_opt
+from datalad.interface.common_opts import save_message_opt
 from datalad.interface.utils import path_is_under
 from datalad.interface.utils import eval_results
 from datalad.interface.base import build_doc
@@ -83,6 +85,8 @@ class Remove(Interface):
             constraints=EnsureStr() | EnsureNone()),
         recursive=recursion_flag,
         check=check_argument,
+        save=nosave_opt,
+        message=save_message_opt,
         if_dirty=if_dirty_opt,
     )
 
@@ -94,6 +98,8 @@ class Remove(Interface):
             dataset=None,
             recursive=False,
             check=True,
+            save=True,
+            message=None,
             if_dirty='save-before'):
         res_kwargs = dict(action='remove', logger=lgr)
         if not dataset and not path:
@@ -270,6 +276,9 @@ class Remove(Interface):
         if not to_save:
             # nothing left to do, potentially all errored before
             return
+        if not save:
+            lgr.debug('Not calling `save` as instructed')
+            return
 
         for res in Save.__call__(
                 # TODO compose hand-selected annotated paths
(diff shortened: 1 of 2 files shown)


yarikoptic:
Any concerns @datalad/developers ? Or let's get it merged

datalad/distribution/remove.py
```diff
@@ -270,15 +276,18 @@ def __call__(
...
                 # TODO compose hand-selected annotated paths
                 path=to_save,
                 # we might have removed the reference dataset by now, recheck
-                dataset=refds_path if GitRepo.is_valid_repo(refds_path) else None,
-                # TODO allow for custom message
-                #message=message if message else '[DATALAD] removed content',
-                message='[DATALAD] removed content',
+                dataset=refds_path
+                        if (refds_path and GitRepo.is_valid_repo(refds_path))
+                        else None,
```
bpoldrack: To solve it that way doesn't really make sense to me. Judging from #1821 the determination of the reference datasets fails for some reason and needs to be fixed. It's not about having none. You can't `datalad remove` anything from a non-repository, right?

datalad/distribution/remove.py
```diff
@@ -94,6 +98,8 @@ def __call__(
             dataset=None,
             recursive=False,
             check=True,
+            save=True,
+            message=None,
```
bpoldrack: I think it's inconsistent to have `save`/`nosave` (and consequently `message`) in `remove` but not in `add`.
Those are not meant to be commands that leave some staged things behind. ATM, I feel this undermines the API concept.

datalad/distribution/remove.py
yarikoptic: well... I am not strong in all this logic... looking at original code though it even allowed refds_path point to some invalid repo as far as I see it ;)  so checking if it is defined at all, sounds legit to me.  moreover we could fix up instead is_valid_repo to not puke upon none?

yarikoptic: BTW, I did add a test for this usecase, so if you see a better/proper fix, you could check it out quickly.

datalad/distribution/remove.py
yarikoptic: #1829 could probably the venue to discuss this further.  Also ref https://github.com/datalad/datalad/issues/1822#issuecomment-329766900 
meanwhile this one fixes an issue of remove crashing and save/nosave is already present in other command(s), so imho it is consistent in general. indeed `add` should get the same powers, and vise versa.  Also, in principle, all those which save should acquire all the powers of save (e.g. `-S`)

bpoldrack: Actually, `add` has it already. I was wrong about that.

datalad/distribution/remove.py
bpoldrack: Either I have misread the diff or sth else was merged in the meantime. Whatever is the case - what I was unhappy about isn't valid (anymore).

PR #1841, opened 2017-09-18
BF: Don't extract an empty string from datalad.repo.version

### LINKED ISSUE #1835: Something broke Travis matrix runs with empty annex version value
Since we never merge PRs which break things, I wonder what upgrade did that

bpoldrack:
Fixes Travis build using `DATALAD_REPO_VERSION=`.

We need a more general solution as stated in the comment (see diff), but a quickfix might be more important right now to not obfuscate other failures.

edit Closes #1835

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -197,6 +197,13 @@ class AnnexRepo(GitRepo, RepoInterface):
 
         if version is None:
             version = self.config.get("datalad.repo.version", None)
+            # we might get an empty string here
+            # TODO: if we use obtain() instead, we get an error complaining
+            # '' cannot be converted to int (via Constraint as defined for
+            # "datalad.repo.version" in common_cfg
+            # => Allow conversion to result in None?
+            if not version:
+                version = None
 
         if fix_it:
             self._init(version=version, description=description)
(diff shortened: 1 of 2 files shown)


mih:
Why do we need a more general solution? This one seems fairly general.

bpoldrack:
Because the same principle applies for other config items, that are set via an env variable.
So we might run in the same issue again. We can prevent this from happening by making sure `ConfigManager` (or may be the `Constraint`s themselves) is able to report `None` for a config item that contains an empty string.

mih:
Right. However, I am not confident that we are treating `''` and `None` as being the same thing throughout our use of config settings. So this would need a more thorough check when approached.

bpoldrack:
I agree. Which is why I wanted to have the urgent quickfix rather than spending time on figuring out a perfect solution right now. I'll have an issue pointing here.

datalad/support/annexrepo.py
```diff
@@ -197,6 +197,13 @@ def __init__(self, path, url=None, runner=None,
 
         if version is None:
             version = self.config.get("datalad.repo.version", None)
+            # we might get an empty string here
+            # TODO: if we use obtain() instead, we get an error complaining
+            # '' cannot be converted to int (via Constraint as defined for
+            # "datalad.repo.version" in common_cfg
+            # => Allow conversion to result in None?
+            if not version:
+                version = None
```
yarikoptic: Agree with @mih, empty is empty, not None generally. But for version I guess it would be ok

bpoldrack: Yeah - see #1842

yarikoptic:
ok, tests do not fail, so it is an improvement, will merge ;)

PR #1845, opened 2017-09-19
BF: towards fixing v6 (and may be direct mode)

### LINKED ISSUE #1826: Strange testrepo setup
In the test repo `basic_annex` (which others depend on), we have a file in git and a file in annex.
The annexed one is added via `addurl`, pointing to the file already in git! (`file:///tmp/wherever/..`)

This has weird effects in V6, leading to a lot of tests to fail. I don't fully understand what's going on in V6 yet, but apart from the need to reproduce this and probably consult Joey: Is there a specific reason to have that strange construct of the same content being in git and annex, @yarikoptic ?
Otherwise I'd just change that to be a file of different content.
And even if we need to have that for some reason, I'd not consider that setup "basic" and have a separate test repo for that.

bpoldrack:
- [x] set up test repos differently to not artificially complicate things for v6 builds (at least partially done):  annexed file in test repo now has different content to not collide with a file under git
- [x] ~fix remote testrepo on github, which is invalid now (see above)~
Can't be done. We can not have a simple annex on github. More complex setup with additional storage elsewhere doesn't serve the needs for a basic testrepo. To be done separatly; see #1826 
- [x] new decorators to be able to distinguish whether we want to skip a test in v6/DM build because it needs fixing or because it doesn't make sense in that mode
- [x] decorators to allow to either skip or not skip known failures
- [x] decorators to allow to probe whether or not a known failure is still failing as suggested in PR #1789
- [x] actually make use of those decorators
- [x] add Travis builds to use the functionality
- [x] don't use empty env vars to define Travis' matrix
- [x] add defaults to `datalad.repo.direct` and `datalad.repo.version`
- [x] make sure the defaults above don't conflict with annex-enforced direct mode
- [x] ~make known failure decorators work with generator tests (apparently only first `yield` is considered or sth like this)~
Won't easily work that way. Probably need to decorate the failing `yield`'s instead as already done in `datalad/interface/tests/test_crawl_init.py`.
- [x] improve and enhance implementation of `datalad unlock`, the errors of which came to attention after solving #1854; Note: Not to be enitrely done in this PR, due to #1860

**Summary regarding changes to known failure decorators:**
There are several decorators in `tests/utils.py` now:
 - `skip_v6` and `skip_direct_mode` for skipping the decorated tests in those builds, since those tests might make no sense (like testing direct mode switches in v6 build, which isn't possible)
 - `known_failure`, `known_failure_v6` and `known_failure_direct_mode` to mark tests, that are are known to fail (always/in v6 build/in DM build) and need some fixing

Behavior of the latter can be tuned via `datalad.tests.knownfailures.skip` (default: True) and `datalad.tests.knownfailures.probe` (default: False). Skipping takes precedence over probing. If not skipped and not probed those tests will yield the actual failure. If not skipped but probed those tests will pass if they still fail and vice versa. This is used in the new Travis builds to see, whether some of them were (may be implicitly) fixed but still marked as known to fail.

**New Travis builds**
There are two new builds probing the known failures in v6 and direct mode. That way we automatically get a hint on what might have been (implicitly) fixed. However, those builds are allowed to fail, since a failure means, that a test actually passed but was marked "known to fail". So it's no harm to master and shouldn't prevent anything from being merged. Furthermore there are some weird tests, that don't always fail, which in return might lead to a failing probing. This hopefully will become better once some basic issues (like the testrepo mess) are fixed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -126,7 +126,6 @@ matrix:
     - DATALAD_REPO_VERSION=6
     - DATALAD_TESTS_KNOWNFAILURES_SKIP=no
     - DATALAD_TESTS_KNOWNFAILURES_PROBE=yes
-
   - python: 2.7
     # test whether known direct mode failures still fail
     env:
@@ -141,6 +140,18 @@ matrix:
     env:
     - TMPDIR="/tmp/nfsmount"
     - _DATALAD_NONPR_ONLY=1
+  # test whether known v6 failures still fail
+  - env:
+    - DATALAD_TESTS_SSH=1
+    - DATALAD_REPO_VERSION=6
+    - DATALAD_TESTS_KNOWNFAILURES_SKIP=no
+    - DATALAD_TESTS_KNOWNFAILURES_PROBE=yes
+  # test whether known direct mode failures still fail
+  - env:
+    - DATALAD_TESTS_SSH=1
+    - DATALAD_REPO_DIRECT=yes
+    - DATALAD_TESTS_KNOWNFAILURES_SKIP=no
+    - DATALAD_TESTS_KNOWNFAILURES_PROBE=yes
 
 # Causes complete laptop or travis instance crash atm, but survives in a docker
 # need to figure it out (looks like some PID explosion)
--- a/.travis.yml
+++ b/.travis.yml
@@ -119,6 +119,17 @@ matrix:
     env:
     - TMPDIR="/tmp/nfsmount"
     - _DATALAD_NONPR_ONLY=1
+  - python: 2.7
+    # test whether known v6 failures still fail
+    - DATALAD_REPO_VERSION=6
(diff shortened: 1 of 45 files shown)


bpoldrack:
If tests are passing, I'll consider this done for now.
Further work on testrepos (see #1826) in next PR.
Further improvement on `unlock` depends on #1860.

bpoldrack:
@yarikoptic : Something is wrong with those buildbots. They timed out. If that was caused by that PR (accidentally waiting for input or sth) it should have happened in OSX and on Travis, too. But didn't. Can you have a look at this?

mih:
LGTM, thanks!

yarikoptic:
FWIW -- buildbots are fine now... not sure if I would have time for proper review but we already skimmed through it during last hangout so merging

PR #1851, opened 2017-09-21
ENH: for exc_str include repr in case no str, provided

yarikoptic:
A little enhancement also added to #1630 to get some sense of what is going on


yarikoptic:
will merge and checkout if our fixes to GitPython helped ;)

PR #1857, opened 2017-09-25
BF: support EXPORTSUPPORTED

yarikoptic:
- [x] test to not blow on unsupported options/commands


yarikoptic:
Fresh git-annex-standalone is available within neurodebian's -devel repository if you would like to try ;)  ideally I should provide a small test for that logic of "new" and "not supported" options/commands... and then we should get it merged... I think test is important here so will do it first

yarikoptic:
I think it is lovely

PR #1858, opened 2017-09-25
BF: report impossible if target repo does not exist during publish

yarikoptic:
Quick&Dirty fix:
In my case it was a submodule which we do not even probably want to publish -- its url points to a github repository (but may be we would?), and it was not locally installed (I believe):

mridefacer in http://datasets.datalad.org/?dir=/labs/gobbini/famface/data  (also subject to not yet finished fix in #1675

ideally we need to decide on what we are doing with "such" submodules
- if their url is some remote, but they are present locally -- should we publish them? (probably so)
   IIRC: `git clone` first tries to find those submodules based on supermodule url, not the url in .submodules. This way also more guarantee that this published clone would have what is needed/referenced

### DIFF
--- a/datalad/distribution/publish.py
+++ b/datalad/distribution/publish.py
@@ -461,10 +461,10 @@ def _get_remote_info(ds_path, ds_remote_info, to, missing):
     """Returns None if desired info was obtained, or a tuple (status, message)
     if not"""
     ds = Dataset(ds_path)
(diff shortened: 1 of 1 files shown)


yarikoptic:
it will probably need a better fix, where we could annotate that this submodule is not for publishing... but for now should suffice imho (if only there was a test ;) )

PR #1867, opened 2017-09-30
Create CODE_OF_CONDUCT.md

mih:


yarikoptic:
It must have been a very strong margarita!  
I don't mind, but also do not see the point really at this point in time, but since it is all fashionable to become this very official, let's do it.

mih:
Still on it -- "community check" on Github pointed out its absence -- now we are top-shelf.

PR #1868, opened 2017-09-30
Create issue_template.md

mih:
git-annex inspired


bpoldrack:
@yarikoptic : Please have a look at those two failures in Travis. This shouldn't happen any more with upstream GitPython, right?

mih:
Ok, test failure has got nothing to do with PR content and the template is what we are using on the webpage anyways...merging.

PR #1870, opened 2017-10-01
RF: dont swallow logs in actual datalad code

### LINKED ISSUE #1753: No logging above DEBUG from low-level Runner
This is still an issue. Unless someone stops me, I will force the change and have any Runner invocation log with DEBUG and no more. Otherwise we have conflict between low and high-level error handling. Previously, we had only low-level handling, but now we have a configurable high-level result management, and this should be used whenever possible.

### LINKED ISSUE #1865: Performance issue of save() with many files
Haven't investigated it, but `datalad add` on a directory with 40k files is very snappy until the `save` stage (single dataset), where an insane waiting period is interrupting the process. CPU is going full tilt, so the machine is clearly pondering some ideas.

I suspect that something in the "sorting" phase of `save` is implemented sub-optimally. We are still calling stuff twice in order to get stats for a progress bar (IIRC), and to be compatible with direct mode. 

Update: After 45min it finally crashed:

```
[ERROR  ] Failed to start ['git', '-c', 'receive.autogc=0', '-c', 'gc.auto=0', 'add', '--dry-run', '-N', '--ignore-missing', '--verbose', '--', <snip 40k files>
[Errno 7] Argument list too long [subprocess.py:_execute_child:1024]
[ERROR  ] add: [Errno 7] Argument list too long 
[ERROR  ] [Errno 7] Argument list too long [subprocess.py:_execute_child:1024] (OSError) 
datalad add study_20140724_095117357  2680,36s user 29,57s system 99% cpu 45:24,90 total
```

The add stage took less than a minute (I think), which is OK for 40k files with about 1GB total.

After the crash evrything was nicely staged, and a simple `git commit` was sufficient to finish the process in:

```
git commit -m save  4,21s user 1,24s system 69% cpu 7,836 total
```

datalad had already staged all files within that initial minute, so `save` spend 40+ minutes only to blow up eventually.

Conclusion: Our approach to pass a long list of files to git is fundamentally broken in such scenario. I know that we have to do this, because we want to be able to work with any flavor of dirty repositories, but clearly we need workarounds to still be able to work with lots of files. And most importantly we should be aware that we are paying a substantial price for this flexibility -- at least with the current implementation.

bpoldrack:
This pull request fixes #1865 and closes #1753

This pull request proposes
- to use `swallog_logs()` for testing purposes only
- to lower the log level used in several places
- new context manager disable_logger for suppressing possible output in critical places

### Changes
- [x] remove uses of `swallow_logs` outside test code (done except for crawler code)
- [x] figure out and fix consequences (proper use of `Runner`'s parameters) Done as far as I'm aware
- [x] don't pass `--debug` to git-annex when log level is `logging.DEBUG` but only when it's 8 or lower

### DIFF
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -76,7 +76,7 @@ class Runner(object):
              Protocol object to write to.
         log_outputs : bool, optional
              Switch to instruct either outputs should be logged or not.  If not
-             set (default), config 'datalad.log outputs' would be consulted
+             set (default), config 'datalad.log.outputs' would be consulted
         """
 
         self.cwd = cwd
@@ -194,8 +194,8 @@ class Runner(object):
     def _log_err(self, line, expected=False):
         if line and self.log_outputs:
             self.log("stderr| " + line.rstrip('\n'),
-                     level={True: logging.DEBUG,
-                            False: logging.ERROR}[expected])
+                     level={True: 9,
+                            False: 11}[expected])
 
     def _get_output_online(self, proc, log_stdout, log_stderr,
                            expect_stderr=False, expect_fail=False):
@@ -320,17 +320,17 @@ class Runner(object):
 
         expect_stderr: bool, optional
             Normally, having stderr output is a signal of a problem and thus it
-            gets logged at ERROR level.  But some utilities, e.g. wget, use
+            gets logged at level 11.  But some utilities, e.g. wget, use
             stderr for their progress output.  Whenever such output is expected,
-            set it to True and output will be logged at DEBUG level unless
+            set it to True and output will be logged at level 9 unless
             exit status is non-0 (in non-online mode only, in online -- would
-            log at DEBUG)
+            log at 9)
 
         expect_fail: bool, optional
             Normally, if command exits with non-0 status, it is considered an
-            ERROR and logged accordingly.  But if the call intended for checking
-            routine, such alarming message should not be logged as ERROR, thus
-            it will be logged at DEBUG level.
+            error and logged at level 11 (above DEBUG). But if the call intended
+            for checking routine, such messages are usually not needed, thus
+            it will be logged at level 9.
 
         cwd : string, optional
             Directory under which run the command (passed to Popen)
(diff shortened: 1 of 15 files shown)


bpoldrack:
@yarikoptic: Here the buildbots are experiencing the same time out as in #1845. So it's not about the PRs.

bpoldrack:
@yarikoptic : I leave the use of `swallow_logs` in `interface/crawl.py` and `crawler/dbs/files.py` to you to decide. The purpose of the former seems to be to have a special logfile, which is fine, I guess and the latter seems to just swallow. So, you might want to consider using the brand new `disable_logger` instead.

yarikoptic:
Great, thanks @bpoldrack. I hope to have a peek at this PR later today

bpoldrack:
Judging from your push, you looked into it. So - merge?

datalad/cmd.py
```diff
@@ -194,8 +194,8 @@ def _log_out(self, line):
     def _log_err(self, line, expected=False):
         if line and self.log_outputs:
             self.log("stderr| " + line.rstrip('\n'),
-                     level={True: logging.DEBUG,
-                            False: logging.ERROR}[expected])
+                     level={True: 9,
+                            False: 11}[expected])
```
yarikoptic: Why to log errors not at error level?

bpoldrack: To avoid shocking users if we missed one, that is actually treated in higher level code. Those logs are important for developers only, I think and we can still have a run at log level 11 to include them without having all the DEBUG output. See #1753 . That's the compromise nobody responded to. ;-)
Despite all that: It's not necessarily an actual error, but just an unexpected output on stderr or an unexpected non-zero return. Even the latter is an error only from the perspective of the command we called - doesn't mean it's an error from datalad's point of view or the user's. The only thing it does mean, is that WE either were not aware that could happen or we failed to properly avoid the ERROR level when we catch that "error" anyway. Either way: It's a hint for datalad developers, nobody else.

bpoldrack: And what's the relation to custom remotes here?

yarikoptic: was reviewing on the phone, mixed up the files I guess ;)

PR #1881, opened 2017-10-06
BF: AnnotatePaths

### LINKED ISSUE #1880: add doesn't error out if asked to add some bogus path
#### What is the problem?
```shell
$> datalad create buga; datalad add -d buga buga/nonexisting
[INFO   ] Creating a new annex repo at /tmp/buga 
create(ok): /tmp/buga (dataset)                                                                                 
add(notneeded): /tmp/buga (dataset)
2 13993.....................................:Fri 06 Oct 2017 07:48:29 AM EDT:.
hopa:/tmp
$> echo $?
0
```

I just woke up

bpoldrack:
This pull request fixes #1880.

To be more specific:
If 'modified' was set, AnnotatePaths failed to report on unavailable paths.

Please have a look @mih . You might want to say something about the approach and the comments I left in the code.

### DIFF
--- a/datalad/interface/annotate_paths.py
+++ b/datalad/interface/annotate_paths.py
@@ -531,7 +531,7 @@ class AnnotatePaths(Interface):
         requested_paths = assure_list(path)
 
         if modified is not None:
-            # modification detection wwould silently kill all nondataset paths
+            # modification detection would silently kill all nondataset paths
             # but we have to complain about them, hence doing it here
             if requested_paths and refds_path:
                 for r in requested_paths:
@@ -546,8 +546,20 @@ class AnnotatePaths(Interface):
                         **dict(res_kwargs, **path_props))
                     res['status'] = nondataset_path_status
                     res['message'] = 'path not associated with reference dataset'
-                    reported_paths[path] = res
+                    reported_paths[r] = res
                     yield res
+
+            # preserve non-existing paths to be silently killed by modification
+            # detection and append them to requested_paths again after detection.
+            # TODO: This might be melted in with treatment of non dataset paths
+            # above. Re-appending those paths seems to be better than yielding
+            # directly to avoid code duplication, since both cases later on are
+            # dealt with again.
+            preserved_paths = []
+            if requested_paths:
+                [preserved_paths.append(r)
+                 for r in requested_paths
+                 if not lexists(r['path'] if isinstance(r, dict) else r)]
 
             # replace the requested paths by those paths that were actually
             # modified underneath or at a requested location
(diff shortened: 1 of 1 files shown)


datalad/interface/annotate_paths.py
```diff
@@ -546,9 +546,25 @@ def __call__(
                         **dict(res_kwargs, **path_props))
                     res['status'] = nondataset_path_status
                     res['message'] = 'path not associated with reference dataset'
+                    # TODO: That `path` in following line seems to be incorrect;
+                    # we are operating on `r` or `p` respectively. Guess it
+                    # should be `r`.
+                    # Figure out, whether or not to report on the resolved one:
```
mih: Agree. But also cannot tell immediately what is the correct fix.

datalad/interface/annotate_paths.py
```diff
@@ -546,9 +546,25 @@ def __call__(
...
+            # TODO: This might be melted in with treatment of non dataset paths
+            # above. Re-appending those paths seems to be better than yielding
+            # directly to avoid code duplication, since both cases later on are
+            # dealt with again.
+            preserved_paths = []
+            if requested_paths:
+                for r in requested_paths:
+                    if not lexists(r['path'] if isinstance(r, dict) else r):
+                        preserved_paths.append(r)
+
```
mih: I aggree with the approach, but this should be a simple list comprehension, no?

datalad/interface/annotate_paths.py
```diff
@@ -560,6 +576,10 @@ def __call__(
                 report_untracked='all' if force_untracked_discovery else 'no',
                 recursion_limit=recursion_limit)
 
+            from itertools import chain
+            # re-append the preserved paths:
+            requested_paths = chain(requested_paths, iter(preserved_paths))
+
```
mih: Aren't these two plain (non-nested) lists? If so, why not just `requested_paths.extend(preserved_paths)`?

datalad/interface/annotate_paths.py
bpoldrack: Right. Will change.

datalad/interface/annotate_paths.py
bpoldrack: Nope. After modification detection `requested_paths` is a generator.

datalad/interface/annotate_paths.py
bpoldrack: So I'll go for the unresolved one for now, since we probably want to report on the original input. If that turns out to be wrong, we'll notice, I guess. :-)

datalad/interface/annotate_paths.py
mih: I knew there would be something. ACK.

bpoldrack:
I guess those buildbots have some trouble again, @yarikoptic ?

yarikoptic:
Crap sorry... I did restart firewall, but then did restart buildbots, thought it is all ok, will check

mih:
@bpoldrack can you trigger the buildbot tests please?

bpoldrack:
Just amended empty to trigger them, but ...
What the hell? Now a lot of shit fails to clone from datasets.datalad.org.
Travis passed before, so I doubt there's anything wrong with that PR.
@yarikoptic : Help, please.

bpoldrack:
FTR: Happens in other branches, too.

yarikoptic:
Fwiw the datasets.datalad.org was fixed, so now it should be genuine failures

bpoldrack:
Nope. The only failure in Travis is a "connection broken" when trying to download PyGithub ;-)

PR #1889, opened 2017-10-09
Nuitka build and bundle of DataLad

### LINKED ISSUE #1688: Provide nuitka-based binary packages
@yarikoptic made a nuitka-based binary build work. thanks!

Based on that we could provide self-contained binary packages in RPM, tar (etc) format to give users an easy way to deploy on *NIX-based infrastructure. Some should be possible for Windows -- maybe even with an MSI installer, or whatever is the method of today on windows.

yarikoptic:
This pull request fixes #1688

TODOs
- [ ] move to build under the eldest debian we could afford (squeeze) to provide max compatibility with other older linuxes
- [ ] support datalad plugins (at least for invocation we could just blindly import the submodule)
- [ ] support patool plugins
- [ ] support datalad test to work (there is also strange non-recognition of -s option)
- [ ] try//provide MacOS bundle

Eventually
- consider additional pieces to be bundled (additional annex special remotes, e.g. rclone-based one)
- fully automate build/upload via buildbot
- try/provide Windows bundle

Sample bundles
- linux: http://onerussian.com/tmp/datalad-linux_0.9.1+git67-g1c55ab3.tgz (built on jessie, so might not work on e.g. CentOS 6.9)


tools/nuitka/build_nuitka
```diff
@@ -0,0 +1,80 @@
...
+# This is a trial error to overcome a problem when using --recurse-all and then
+# inspect.isfunction starting to puke on majority of our interfaces
+
+# --recurse-stdlib  leads to some "fancy" errors
+
+#nuitka --exe --python-flag=no_site --recurse-directory=datalad --recurse-to={datalad{,.distribution,.support,.interface},wrapt,six,git,gitdb,argcomplete,iso8601,requests,keyring,appdirs,msgpack,smmap,tqdm,pyld,colorama,docopt,patoolib,urllib3,chardet,_patool_configdata,pkg_resources,backports.ssl_match_hostname,scrapy,cssselect,cssselect.xpath,lxml,lxml.html,w3lib.url,w3lib.http,w3lib.html,w3lib{,.encoding,.util},simplejson,nibabel,humanize,yaml,httpretty,github,jsmin,h5py,bs4,mock,nose,numpy} --recurse-not-to={boto,gtk,pyglet,tornado,matplotlib,twisted,numpy,pandas,vcr,sphinx,zmq,PIL,IPython,OpenGL} `which datalad` 2>&1 | tee $0.log
+
+set -eu
+
+os=linux
```
bpoldrack: `os` needs to be a parameter to the script, no?

tools/nuitka/build_nuitka
```diff
@@ -0,0 +1,80 @@
...
+python: $python_version
+datalad: $version
+host: `hostname`
+EOF
+
+
+# shouldn't exist yet, thus no -p
+mkdir datalad.dist/bin
+cp -p $thisdir/datalad datalad.dist/bin
+rsync -a /usr/lib/git-annex.${os} datalad.dist/
```
bpoldrack: That `rsync` call is os-specific. Not sure ATM whether rsync is available for Windows at all. In any case that path isn't the one used on windows. I'll look up what's the correct one in case of windows ...

yarikoptic: I guess that is why worth keeping "os" hardcoded for now (re above comment) until we make it work at least on Linux ;)

tools/nuitka/build_nuitka
kayhayen: The --exe is the default, you can remoe it.

Issue #1890, opened 2017-10-09
Test must not use existing user configuration


mih:
Even with our present capabilities we can substantially alter the behavior of datalad, to a point where tests wont pass anymore.

We need to change our setup such that tests run on a clean config.

mih:
No longer an issue with #2014

PR #1899, opened 2017-10-13
WIP: DON'T MERGE: Ultimate test setup

### LINKED ISSUE #1826: Strange testrepo setup

bpoldrack:
You may look, but don't touch!

- [x] Closes #1826
- [ ] find solution for #1890 
- [x] Implement `TestRepo` base class
- [x] Implement basic `Item` subclasses
- [ ] Implement `TestRepo`'s, `ItemRepo`'s and `ItemFile`'s `assert_intact` method (mostly done)
- [ ] resemble all existing testrepos (basically done) by defining corresponding subclasses 
- [ ] Implement new `with_testrepos`
- [ ] Implement transition helper to let tests use new repos without adapting them
- [x] Have an extensive check on definitions of subclasses of `TestRepo` with meaningful messages to ease doing so for everybody
- [ ] Provide new test helpers to easily use new powers within tests (see "TODO: - use the power!" in `testrepos.utils.py`
- [ ] Fully adapt at least some actual tests (probably `GitRepo`/`AnnexRepo`) to show off
- [x] delay implementation of probably required `ItemCommand`s and some properties until they are actually required. Do this while adapting tests and creating additional `TestRepos`. It's important to make use of what's already available ASAP.
- [x] Lazy creation (currently we pre-create ALL of them although the test might need just a single one of them)

Provide new concept for running unit tests in different setups.

**short overview of the goals**
New `TestRepo` subclasses are capable of setting up pretty much everything you can think of, including untracked files, independent repos (or even an entire forrest of hierarchies at once), staged and unstaged modifications, ... everything. This used to be not possible, since we relied on them being cloneable repositories.
They are delivered by `with_testrepos` as an object that contains all information about the SHOULD-BE state of the entire setup. Tests now can base their assertions on properties of the `TestRepo`. Thereby it's easy to write tests that can actually run on anything being thrown at them. They don't need (hard-)coded assumptions about the tesrepos that were known to the developer when writing the test. And they also don't need to read properties from FS, which basically means to make the IS state the SHOULD-BE state.
They provide full automatic integrity testing. No need for a test developer to each time think about, what could possibly have changed and come up with a complete set of assertions to make sure nothing has changed.
This also means there's now the possibility to just define intended changes and automatically run a full set of assertions making sure that those changes and those changes ONLY actually happened. Again: No need by the developer to always come up with the necessary assertions to make that sure.
If used via new `with_testrepos`decorator it is ensured that testrepos can't be left messed up by tests and still being used by other tests, too. However, if the test requests a "read-only" instance of a `TestRepo` they will get a persistent one, that isn't always recreated and used by others, too. Those persistent ones will check themselves after such a test (and raise of course if check failed) and will be automatically recreated from scratch for the next test that wants to use them.
Finally, the definition of new test setups is pretty easy, I think. To help with that even further there are supposed to be extensive checks on the definitions, generating pretty precise messages, that are telling what's wrong with the definition.

Implicit feature: There is basically no need for additional unit tests on `TestRepo` and the subclasses of `Item`. They implicitly test themselves as soon as they are used in a subclass of `TestRepo`. This subclass just needs to be instantiated. The "object consistency" part of `assert_intact` (which is autmatically and recursively called for all item of a `TestRepo`) does that. Whatever unit test you can come up with that actually contains assertions that are not tested anyway, implies that either those assertions should be added in a general way in the appropriate `assert_intact` method or it's about a not yet tested setup, meaning you should simply define a new `TestRepo` instead. The latter will automatically test the code in `testrepo` module against that situation and furthermore this also leads to us testing DATALAD against that untested situation!


datalad/tests/testrepos/items.py
```diff
@@ -0,0 +1,1396 @@
...
+        # TODO: if not self.is_untracked: git knows it
+        # TODO: if self.annexed: annex finds it and provides a key
+        # TODO: compare commits
+        # TODO: - state? This might need annex-proxy and ignore-submodules, but
+        #         since we have a certain file to specify it might work
+        #       - see corresponding note in ItemRepo
+
+
+@auto_repr
+class ItemInfoFile(ItemFile):
```
yarikoptic: invaluable concern:  use "Item" as a suffix (if at all) not as a prefix, to stay inline with other places (AnnexRepo, etc)

datalad/tests/testrepos/items.py
```diff
@@ -0,0 +1,1396 @@
...
+            # If that happens, that ItemCommand probably should be adapted to
+            # also remove self._src.
+            assert(self.remotes)
+
+        assert_is_instance(self.branches, set)
+        [assert_is_instance(b, string_types) for b in self.branches]
+
+        if self.branches:
+            assert self.commits
+            # TODO: Not necessarily vice versa? Could be just detached HEAD, I guess
```
yarikoptic: make sure that it also asserts on the states of the branches

although git-annex branch might be tricky, since many git-annex operations just adjust the timestamp within the log of the git-annex although no content being dropped etc. E.g. 'annex fsck' would change git-annex branch.

datalad/tests/testrepos/items.py
yarikoptic: here is an example for attrs
```python
>>> import attr
>>> @attr.s
... class C(object):
...     x = attr.ib(default=42)
...     y = attr.ib(default=attr.Factory(list))
```

so we could do just that `from testrepos import items`, and use `items.File`, or if really valuing our bytes, `from testrepos import items as it;  it.File`

yarikoptic:
the failing test in the env with lots of debugging etc
```

ERROR: datalad.tests.testrepos.tests.test_repos.test_BasicMixed_instantiation

  File "/home/travis/virtualenv/python2.7.13/lib/python2.7/site-packages/nose/case.py", line 197, in runTest

    self.test(*self.arg)

  File "/home/travis/build/datalad/datalad/datalad/tests/utils.py", line 593, in newfunc

  File "/home/travis/build/datalad/datalad/datalad/tests/testrepos/tests/test_repos.py", line 108, in test_BasicMixed_instantiation

    tr = BasicMixed(path)

  File "/home/travis/build/datalad/datalad/datalad/tests/testrepos/repos.py", line 343, in __init__

    self.assert_intact()

  File "/home/travis/build/datalad/datalad/datalad/tests/testrepos/repos.py", line 396, in assert_intact

    [item.assert_intact() for item in self._roots]

  File "/home/travis/build/datalad/datalad/datalad/tests/testrepos/items.py", line 592, in assert_intact

    for line in out.splitlines()]

IndexError: list index out of range
```

bpoldrack:
Re failing test:
It's about enabling datalad special remote. With that being set, the output parsing for `git remote -v` (from within `assert_intact`) fails, since it is unexpectedly just the word "datalad". Probably need to parse `.git/config` instead to account for special remotes.

bpoldrack:
Don't see myself finishing this anytime soon. Thus closing.

PR #1907, opened 2017-10-19
ENH(TST): test on travis (allow to fail) against -devel version of annex

yarikoptic:
This pull request proposes


yarikoptic:
ok, that additional matrix is passing right now (so my #1908 is indeed specific to my system).  I will merge.

PR #1911, opened 2017-10-24
TST: DON'T MERGE

bpoldrack:


bpoldrack:
Okay. This is proof the `pyld` failures are happening in master and have nothing to do with PR #1899 at all.
@yarikoptic , @mih : Still no idea where this is coming from. If nobody happens to be more enlightened, let's just get #1630 ready for merge before anything else and be done. :-)

PR #1912, opened 2017-10-24
`remove` does not commit the removal of a 2nd file in a sequence

mih:
#### What is the problem?
`datalad remove a b` will remove a and b from the file system, but will only commit the removal of a and not b -- left as staged for deletion.


bpoldrack:
Okay, not yet sure how to fix it, but I found the bug in `remove.py`

bpoldrack:
FTR: Apart from all the pyld/jsonld failures, the test for this issue still fails on Travis while it's passing for me locally.

bpoldrack:
Closing in favor of #1914

PR #1913, opened 2017-10-24
WIP: BF: module lzma isn't available

bpoldrack:


bpoldrack:
Okay, apparently that works. Failures are about `jsonld` and datacite once again.
@mih : You might want to just cherry-pick dee0b20b61c9243aaf0256fc7c7b9f0234517247

PR #1917, opened 2017-10-24
ENH: make AutomagicIO proxy also io.open and lzma.LZMAFile

### LINKED ISSUE #1884: How to use AutomagicIO within DataLad itself?
#### What is the problem?
DataLad might need to obtain file content for its internal processing (think: aggregated metadata). It would be great, if we could use automagicIO internally too, and not having to get()/unlock() things "by hand".

Any pointer on where/how to start with this?

Side-benefit: Creates a "dog-fooding" situation that should make this feature more robust over time.

yarikoptic:
This pull request fixes #1884 for those particular use cases (needs 1 more commit in that branch, will push together with this one): lzma.LZMAFile and io.open

### Changes
- [x] This change is complete

### DIFF
--- a/datalad/tests/test_auto.py
+++ b/datalad/tests/test_auto.py
@@ -9,6 +9,7 @@
 """Test proxying of core IO operations
 """
 
+import io
 import os
 from os.path import join as opj, dirname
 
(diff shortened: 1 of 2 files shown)


yarikoptic:
the issues are due to other failures, will need to have master merged after #1918 is merged

mih:
Works in #1630. Thanks!

datalad/tests/test_auto.py
```diff
@@ -172,12 +173,43 @@ def generate_dat(f):
...
             eq_(f.read(), "123")
 
     yield _test_proxying_open, generate_dat, verify_dat
 
 
+def test_proxying_io_open_regular():
+
+    def generate_dat(f):
+        with io.open(f, "w", encoding='utf-8') as f:
+            f.write(u"123")
```
bpoldrack: The `f.write(u"123")` is what is failing in PY3 build.
`write` expects sth, that implements [buffer interface](https://docs.python.org/3/c-api/buffer.html).
Since it's an ASCII string anyway `b"123"` should work in this case, I think.

bpoldrack:
Well, that should work, too, of course. :-)

yarikoptic:
for some reason I thought that `b""` isn't in 2.7... yeah, may be should have just gone for it. so, since no objections, will merge

mih:
FTR: Tests under PY3 are now flooded with:

```
[WARNING] Failed proxying open with (3, 'rb', -1), {}: lstat: illegal type for path parameter [posixpath.py:lexists:171] 
[WARNING] Failed proxying open with (5, 'rb', -1), {}: lstat: illegal type for path parameter [posixpath.py:lexists:171] 
[WARNING] Failed proxying open with (3, 'rb', -1), {}: lstat: illegal type for path parameter [posixpath.py:lexists:171] 
```

but there is no functional impairment. Haven't looked into it, but it smells like we are doing something wrong.

PR #1920, opened 2017-10-25
BF: Thanks to @davidlehn pyld works again with version >= 0.8.2

yarikoptic:
ref: https://github.com/digitalbazaar/pyld/issues/69#issuecomment-339193989

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -85,7 +85,7 @@ requires = {
     ],
     'metadata': [
         'simplejson',
-        'pyld<0.8,>=0.8.2',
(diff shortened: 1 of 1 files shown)


mih:
Seems it doesn't find a suitable candidate

bpoldrack:
Sure thing it doesn't. Since we it's looking for the wrong thing now: 
> `Could not find a version that satisfies the requirement pyld<0.8,>=0.8.2`

Whatever syntax for specifiying alternative version restrictions might be supported - that's not the one. :-)

yarikoptic:
I thought I have tried and it worked... oh well -- removed any versioned requirement now. must work ;)

PR #1926, opened 2017-10-27
ENH: crash fi git config is not yet setup with name/email

yarikoptic:
Otherwise we could get dozens of tests failing due to incorrect
operations etc, so better to crash as early as needed, e.g.
while instantiating the first GitRepo

I managed to run "datalad test" in a fresh'ish docker image - ended up with FAILED (SKIP=66, errors=60, failures=20) ;-)


bpoldrack:
LGTM
Failure seems unrelated. I think I saw this happen in PR #1630, too. However, since this appears to be about special remotes would be nice if you could have a look at it.

PR #1928, opened 2017-10-28
Some minor improvements of the plugins

mih:
### Changes
- [x] `export_tarball` has docs
- [x] `export_tarball can generate tarballs from incomplete datasets
- [x] show signature of any plugin, even when there are no real docs
- [x] `export_tarball` -> `export_archive` for TAR and ZIP archives, both compressed or uncompressed

Note that this is a prerequisite for #1942

### DIFF
--- a/datalad/plugin/export_archive.py
+++ b/datalad/plugin/export_archive.py
@@ -6,21 +6,26 @@
 #   copyright and license terms.
 #
 # ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
-"""export a dataset as a TAR archive"""
+"""export a dataset as a TAR/ZIP archive"""
 
 __docformat__ = 'restructuredtext'
 
 
 # PLUGIN API
-def dlplugin(dataset, filename=None, on_failure='error'):
-    """Export the content of a dataset as a TAR archive.
+def dlplugin(dataset, filename=None, archivetype='tar', compression='gz',
+             on_failure='error'):
+    """Export the content of a dataset as a TAR/ZIP archive.
 
     Parameters
     ----------
     filename : str, optional
       File name of the generated TAR archive. If no file name is given
       the archive will be generated in the current directory and will
-      be named: datalad_<dataset_uuid>.tar.gz.
+      be named: datalad_<dataset_uuid>.(tar.*|zip).
+    archivetype : {'tar', 'zip'}
+      Type of archive to generate.
+    compression : {'', 'gz', 'bz2')
+      Compression method to use. 'bz2' is not supported for ZIP archives.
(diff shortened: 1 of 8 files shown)


datalad/plugin/export_archive.py
```diff
@@ -0,0 +1,126 @@
...
+                add_method(
+                    fpath,
+                    arcname=aname,
+                    **(tar_args if archivetype == 'tar' else {}))
+            except OSError as e:
+                if on_failure in('ignore', 'continue'):
+                    (lgr.warning if on_failure == 'continue' else lgr.debug)(
+                        'Skipped %s: %s',
+                        fpath, exc_str(e))
+
```
yarikoptic: So is that the way to generate archive from incomplete? On error continue?

If this plugin is used as a plugin for another action, might be impossible to provide specific control for this behavior without allowing all other errors also continue

Why not to make a dedicated option?

mih: I don't understand. There is a dedicated option for it. It specifically handles errors for individual files being added to the archive. it is distinct from any general error handler, to avoid this kind of confusion. So what do you want to be different?

yarikoptic: sorry, my bad -- I have missed that `on_error` is an option to the plugin and assumed that it was that global setting.
BTW -- would be nice if any test exercised this .
It seems that if I mistype `on_failure` value - it would be `ignore'.

datalad/plugin/export_archive.py
```diff
@@ -0,0 +1,126 @@
...
+                filename, 'w',
+                zipfile.ZIP_STORED if not compression else zipfile.ZIP_DEFLATED) \
+            as archive:
+        add_method = archive.add if archivetype == 'tar' else archive.write
+        repo_files = sorted(repo.get_indexed_files())
+        if isinstance(repo, AnnexRepo):
+            annexed = repo.is_under_annex(
+                repo_files, allow_quick=True, batch=True)
+        else:
+            annexed = [False] * len(repo_files)
```
yarikoptic: would be nice to have it tested on some non-annex repo just to make sure it doesn't blow somewhere with assumption of being under annex

mih:
I have a small fix -- will push when back online.

mih:
@bpoldrack Can you have a quick look at this one please. Direct mode is giving me trouble. I thought my use of `file_has_content()` would solve this, but I was wrong, it seems. Thanks!

yarikoptic:
#1942 is based on this branch... rebasing here brings chaos to the world

mih:
Sorry about the rebase, juggling to many branches that build off each other here. Pushed a fix that will hopefully get this merged.

mih:
Travis is happy.

PR #1929, opened 2017-10-29
BF: specify if to request pyliblzma solely on python version

yarikoptic:
do not rely on current state of local affairs of having lzma to define it as a
requirement or not.  I had e.g. one installed globally.

Might fix remaining issues on linux buildbots with running tests.

On OSX need to figure out how to install liblzma-dev


yarikoptic:
I think this it the way to go if we are to stick with pyliblzma, so merging ;)  #1934 will provide an additional detail

PR #1931, opened 2017-10-29
Various smallish metadata improvements

### LINKED ISSUE #1828: bids meta: if README is not a correct utf-8, dumped into json description is not human readable
even though it was (re)loaded with errors="replace", so it should be legit'ish, then in json entire thing is stored prefixing each character with  a "unicode designator" `\u0000`. See e.g.
http://datasets.datalad.org/openfmri/ds000222/.datalad/meta/meta.json

I have tried a few sacred rinsing procedures but have failed to cast away the deamons

mih:
This pull request fixes #1828 

### Changes
- [x] Guess encoding of file
- [x] Better logging of metadata wrangling
- [x] never remove internal metadata fields due to size constraints
- [x] much higher default metadata size level -- people should lower it, once they notice a problem, not having no metadata until they realize they have stripped it.

### DIFF
--- a/datalad/interface/common_cfg.py
+++ b/datalad/interface/common_cfg.py
@@ -221,7 +221,7 @@ definitions = {
         'ui': ('question', {
                'title': 'Maximum metadata field size',
                'text': 'Metadata fields exceeding this size (in bytes/chars) are excluded from metadata extractio'}),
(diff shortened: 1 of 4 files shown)


mih:
OK, this is just small bait ...

PR #1932, opened 2017-10-29
BF(TST): swallow outputs while ds.run for the case of nose without -s

yarikoptic:
Recent tests of master were failing in the "heavy debug" environment, triggered by the fact that we use nose without `-s` there.  I do not remember how we addressed this in other places, but clearly here we seems to need a "proper" stdout with fileno, so I just wrapped with `swallow_outputs`.  Is there is a better way @bpoldrack?

### DIFF
--- a/datalad/interface/tests/test_run.py
+++ b/datalad/interface/tests/test_run.py
@@ -32,6 +32,7 @@ from datalad.tests.utils import assert_status
 from datalad.tests.utils import assert_result_count
 from datalad.tests.utils import assert_in
 from datalad.tests.utils import swallow_logs
+from datalad.tests.utils import swallow_outputs
 from datalad.tests.utils import skip_if_on_windows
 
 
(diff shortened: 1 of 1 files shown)


yarikoptic:
quick review/merge would be appreciated so we reduce "failed" noise from the CIs

bpoldrack:
IIRC, we have a dedicated decorator for this issue.

bpoldrack:
It's in `tests/utils.py` and it is called `ignore_nose_capturing_stdout`.
However, this just ensures the test doesn't formally fail. But: Since we just catch the exception therein, the test isn't entirely executed (until it raises only).
So, if we want to actually run those tests in the build without  `-s`, `swallow_outputs` is the better way.
Still - would be nice to integrate with existing `ignore_nose_capturing_stdout` (RF to change its implementation and name) and let it be a context manager as well as a decorator. So, we could use it the same way you did here but WITH knowing what the hell this is about when looking at the code later on ;)

PR #1922, opened 2017-10-30
AutomagicIO issues under PY3

mih:
FTR: Tests under PY3 are now flooded with:

```
[WARNING] Failed proxying open with (3, 'rb', -1), {}: lstat: illegal type for path parameter [posixpath.py:lexists:171] 
[WARNING] Failed proxying open with (5, 'rb', -1), {}: lstat: illegal type for path parameter [posixpath.py:lexists:171] 
[WARNING] Failed proxying open with (3, 'rb', -1), {}: lstat: illegal type for path parameter [posixpath.py:lexists:171] 
```
but there is no functional impairment. Haven't looked into it, but it smells like we are doing something wrong.

### DIFF
--- a/datalad/auto.py
+++ b/datalad/auto.py
@@ -134,7 +134,10 @@ class AutomagicIO(object):
                         lgr.debug("No name/file was given, avoiding proxying")
                         raise _EarlyExit
                     file = kwargs.get(filearg)
-
+                if isinstance(file, int):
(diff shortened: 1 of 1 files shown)


yarikoptic:
Damn... I thought buildbots came online... Will check tomorrow. Looks LGTM

PR #1949, opened 2017-10-31
Fix bytes object path issue with pandas #1948

### LINKED ISSUE #1948: Automagic - Path as binary object - pandas
#### What is the problem?
Opening files with pandas using AutomagicIO fails. This is because pandas (perhaps due to to using CPython?) returns a bytes object for the file name, so statements such as `'annex/objects' in realpath(filepath)` will return an error, because filepath is not a string.

Open a file that Datalad would have to fetch using `pandas.read_csv`

#### What version of DataLad are you using?
Latest version of Datalad on Ubuntu in Docker.

adelavega:
This pull request fixes #1948

This pull request proposes converting path names to strings prior to checking if 'annex/objects` is in path.

### DIFF
--- a/datalad/auto.py
+++ b/datalad/auto.py
@@ -180,7 +180,7 @@ class AutomagicIO(object):
         # For now, as long as it is a symlink pointing to under .git/annex
         if exists(path):
             return True
-        return lexists(path) and 'annex/objects' in realpath(path)
+        return lexists(path) and 'annex/objects' in str(realpath(path))
 
     def _dataset_auto_get(self, filepath):
         """Verify that filepath is under annex, and if so and not present - get it"""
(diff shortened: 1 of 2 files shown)


yarikoptic:
Thanks @adelavega !  I wonder though how your changes broke that single usecase for python3.x on travis...  I wonder if `datalad.utils.safe_unicode` would be a `safer` choice here.  Meanwhile restarted one python3 build to see if that wasn't just network glitch/limiting

Also it revealed a little python3 bug of us still accessing Exception.message in few spots -- I will fix for that shortly in a dedicated PR

mih:
Tests pass now

adelavega:
Awesome, thanks!

PR #1947, opened 2017-10-31
Have metadata parser for generic image files

mih:
Basic stuff like

- size
- dpi
- format
- bits per pixel
- color mode


yarikoptic:
thought to push a commit with a tiny-kinda-fix (primarily just to trigger buildbots) but apparently have no permission -- could you please allow @mih?

yarikoptic:
ah whatever -- will fix straight in master

PR #1961, opened 2017-11-12
BF: crawler needed to use now "options" instead of "git_options" apparently

yarikoptic:
`git_options` are ignored, `options` are respected... duplication and absent clarity in the interface.  Just patching up and adding a test


yarikoptic:
#1964 includes the fix for that lzma issue, unrelated to this PR, so will just merge

PR #1970, opened 2017-11-15
BF: do not blow if percent-progress is not provided by annex

yarikoptic:
Also just reuse one if already known from previous logic

The issue was that git annex does not report in json the percent progress
if key did not have size known (eg for a fast url or youtube video).
Next git-annex version (to be tested) should report one from the file size


yarikoptic:
thought also to improve it, but git annex already got needed fix so will leave it as is for now

PR #1990, opened 2017-11-24
DOC: Some love for the plugin docs

mih:


docs/source/customization.rst
```diff
@@ -83,13 +134,20 @@ be a string.
...
 of definitions must be done within the body of the function.
 
+Documention
+-----------
+
 The doc string of the plugin function is displayed when the plugin
 documentation is requested. The first line in a plugin file that starts
 with triple double-quotes will be used as the plugin short description
-(this will typically be the docstring of the module file). This short
+(this will typically be the doc string of the module file). This short
```
yarikoptic: I think it was ok to keep docstring used as a word -- we use it that way in other dozens of hits of git grep

Issue #1983, opened 2017-11-23
compressed metadata files better have a corresponding compression extension (e.g. .xz)


yarikoptic:
#### What is the problem?
Having an extension helps when using lesspipe etc helpers for viewing etc those files.  Also, in the future happen we switch to even more efficient one (e.g. piped piper) we could easily differentiate without guessing the type etc

PR #1992, opened 2017-11-24
Give XZ compressed metadata objects an .xz extension (fixes gh-1983)

mih:


datalad/metadata/aggregate.py
```diff
@@ -227,6 +227,8 @@ def _extract_metadata(agginto_ds, aggfrom_ds, db, merge_native, to_save):
             continue
         # only write to disk if there is something
         objrelpath = _get_obj_location(objid, label)
+        if store is json_py.dump2xzstream:
```
yarikoptic: feels like we might want to have the choice as a config option eventually,  but I guess for now should suffice.

yarikoptic:
related tests passed in travis, merging

datalad/metadata/aggregate.py
yarikoptic: if only there was a test though which would help to spot whenever regression might happen

PR #1998, opened 2017-11-27
WIP Add to matrix (alongside with upstream GitPython) to test against snapshot of annex (in master only)

yarikoptic:
if passes -- then I will revert last commit and be ready for merge


yarikoptic:
should be great

PR #2007, opened 2017-11-30
DOC: clarify that archive must be already under annex for add-archive-content

### LINKED ISSUE #1997: datalad add-archive-content: better documentation
#### What is the problem?

1. I have no clue on how to use `datalad add-archive-content`, and the error message/documentation could be more informative.
~~2. It fails to add the content of the archive under git annex.~~ (dumb me I had `.gitignore` set to ignore the files that were in the tar)

I have a tarball, which is untracked. The tarball is the only file that is untracked. Naturally, I would do `datalad add-archive-content mytarball.tgz` (as the documentation suggests). I would expect this to do two things: 1. add `mytarball.tgz` to annex; 2. extract `mytarball.tgz` and add the content to annex. 

Instead, I get this error

```terminal
mvdoc@euclid ~/exp/hauntedhouse_mne/singularity (master*) $ datalad add-archive-content hauntedhouse.img.tgz
[ERROR  ] You better commit all the changes and untracked files first [add_archive_content.py:__call__:235] (RuntimeError)
```
Which doesn't really suggest that I should add the tarball to annex, but I tried to do it anyway:
`datalad add mytarball.tgz && datalad add-archive-content mytarball.tgz`, and that did the trick.

```terminal
mvdoc@euclid ~/exp/hauntedhouse_mne/singularity (master) $ datalad --version
datalad 0.9.1
```

#### Have you had any success using DataLad before? (Sometimes we get tired of reading bug reports all day and a lil' positive end note does wonders)

I can't live in a world without datalad <3

mvdoc:
Also add more informative error message if archive is not under annex yet

This pull request fixes #1997 

### Changes
- [x] This change is complete
- [ ] This is in progress

### DIFF
--- a/datalad/interface/add_archive_content.py
+++ b/datalad/interface/add_archive_content.py
@@ -62,7 +62,7 @@ _KEY_OPT_NOTE = "Note that it will be of no effect if %s is given" % _KEY_OPT
 class AddArchiveContent(Interface):
     """Add content of an archive under git annex control.
 
-    This results in the files within archive (which should be under annex
+    This results in the files within archive (which must be already under annex
     control itself) added under annex referencing original archive via
     custom special remotes mechanism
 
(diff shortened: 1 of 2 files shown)


bpoldrack:
Thanks, @mvdoc!

PR #2010, opened 2017-11-30
ENH/BF: do not do anything in SSHManager upon its construction. Also guard around permission changes

### LINKED ISSUE #2009: sshconnector always tries to do something right upon import
#### What is the problem?
as recently "improved" travis setup, which now runs wtf upon completion shows after 'sudo'ed run of tests:  https://travis-ci.org/datalad/datalad/jobs/309577052
```shell
$ datalad plugin wtf
  File "/home/travis/virtualenv/python2.7.13/bin/datalad", line 7, in <module>
  File "/home/travis/build/datalad/datalad/datalad/__init__.py", line 34, in <module>
    ssh_manager = SSHManager()
  File "/home/travis/build/datalad/datalad/datalad/support/sshconnector.py", line 359, in __init__
    for p in listdir(self.socket_dir)
  File "/home/travis/build/datalad/datalad/datalad/support/sshconnector.py", line 374, in socket_dir
    chmod(self._socket_dir, 0o700)
OSError: [Errno 1] Operation not permitted: '/home/travis/.cache/datalad/sockets'
```
SSHManager, even upon import already is doing some house keeping stuff. Since many, if not majority of datalad commands do not need ssh for their operation, it would be nice if SSHManager delayed any of those actions all the way until it actually needs to do something.

yarikoptic:
This pull request fixes #2009 
Should fix travis build in 0.9.x (and then master) branches

### Changes
- [x] This change is complete

### DIFF
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -522,6 +522,10 @@ class GitRepo(RepoInterface):
                 msg="RF: url passed to init()"
             )
 
+        # So that we "share" control paths with git/git-annex
+        if ssh_manager:
+            ssh_manager.assure_initialized()
+
         if not GitRepo._config_checked:
             check_git_configured()
             GitRepo._config_checked = True
--- a/datalad/support/sshconnector.py
+++ b/datalad/support/sshconnector.py
@@ -356,40 +356,46 @@ class SSHManager(object):
         # handling of socket_dir, so we do not define them here explicitly
         # to an empty list to fail if logic is violated
         self._prev_connections = None
+        # and no explicit initialization in the constructor
+        # self.assure_initialized()
 
     @property
     def socket_dir(self):
         """Return socket_dir, and if was not defined before,
         and also pick up all previous connections (if any)
         """
-        if self._socket_dir is None:
-
-            from ..config import ConfigManager
-            from os import chmod
-            cfg = ConfigManager()
-            self._socket_dir = opj(cfg.obtain('datalad.locations.cache'),
-                                   'sockets')
-            assure_dir(self._socket_dir)
-            try:
-                chmod(self._socket_dir, 0o700)
-            except OSError as exc:
-                lgr.warning(
-                    "Failed to (re)set permissions on the %s. "
-                    "Most likely future communications would be impaired or fail. "
-                    "Original exception: %s",
-                    self._socket_dir, exc_str(exc)
-                )
-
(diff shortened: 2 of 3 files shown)


bpoldrack:
This might turn out to be troublesome.
We don't open all the SSH connections we care for by means of actual datalad code. Instead, git-annex can open some and we still want reuse those and consider them to be opened by our process (or it's childs). If this happens before "we" requested anything explicitly, those connections now would be considered to not "belong" to us. We might still be able to reuse them, but we wouldn't close them - they'd just timeout at some point. Might be sufficient but I'm not quite sure about security implications.

mih:
Not sure if their is a counter example, but if we do not open something, we should not be in the business of closing it -- unless we are the reason why whoever opened it could not close it.

bpoldrack:
So, I think it'd be better to consider any already existing connections right at import. Instead may be just catch the case, where we can't access them and may be have warning or sth. stating that we can't ensure connection sharing properly, since we have no permission. That's the actual failure and we don't solve it by delay. If in that case a non- sudo'ed call like this `datalad wtf` would actually request a connection we'd run right into the same failure.

bpoldrack:
Re Michael:
Yes, the thing is: In that hypthetic example we caused annex to open the connection and we told it to keep it open so we can reuse it. Then we SHOULD care for closing, but we can't now.

mih:
I cannot follow that argument. Why would
```
import datalad
print(datalad.__version__)
```
would want to deal with any network issue?

Likewise, issuing any warning that something network-related isn't possible when no network is needed is not compatible with humans. Ideally, such messages should appear when relevant, and not "remember when you imported datalad yesterday, I told you that this wouldn't work" ;-)

bpoldrack:
Well, we could delay the actual message by just storing the reason we couldn't detect anything and issue a warning only if a connection was actually requested.

But what we cannot do is to not detect "connections, that existed before" later on. The result of that detection is simply invalid. Except we reimplement annex to figure out beforehand whether or not it will decide to open an SSH connection.

bpoldrack:
Also:

> Why would would want to deal with any network issue?

sounds worse than it is. We don't care whether or not there network issues at that point. The only thing that is about is checking whether or not there are open connections already at the place that we want to use for our sockets. That's it.

yarikoptic:
@bpoldrack , I have done what we talked about ;) so ready for merge imho

PR #2012, opened 2017-12-01
ENH: automagicio - proxy also isfile

### LINKED ISSUE #2011: automagicio: also decorate os.path.{exists,file}
#### What is the problem?
some apps (such as fsleyes) sense first for the files to be there since they sweep through the extensions etc.  They do not use lexists to actually test if there is smth like what they need, so automagicio is of no effect for them.  I think we could/should get the files upon sensing as well.

yarikoptic:
This pull request fixes #2011 

### Changes
- [x] This change is complete

quite neat:
```shell
$> python -m datalad /usr/bin/fsleyes ~/datalad/openfmri/ds000002/sub-05/anat/sub-05_inplaneT2.nii.gz
[INFO   ] Activating DataLad's AutoMagicIO 
[INFO   ] Running code of /usr/bin/fsleyes 
[INFO   ] File /home/yoh/datalad/openfmri/ds000002/sub-05/anat/sub-05_inplaneT2.nii.gz has no content -- retrieving 
19:23:24: Debug: ScreenToClient cannot work when toplevel window is not shown                                   
19:23:24: Debug: ScreenToClient cannot work when toplevel window is not shown                                   
19:23:24: Debug: ScreenToClient cannot work when toplevel window is not shown
...
python -m datalad /usr/bin/fsleyes   6.40s user 1.62s system 71% cpu 11.146 total
```

### DIFF
--- a/datalad/auto.py
+++ b/datalad/auto.py
@@ -77,6 +77,7 @@ class AutomagicIO(object):
         self._builtin_open = __builtin__.open
         self._io_open = io.open
         self._builtin_exists = os.path.exists
+        self._builtin_isfile = os.path.isfile
         if h5py:
             self._h5py_File = h5py.File
         else:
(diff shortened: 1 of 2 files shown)


bpoldrack:
Makes sense to me. Once AutomagicIO is enabled it'd be inconsistent to be able to open file, but being reported that it's not a file or doesn't exist.

PR #2021, opened 2017-12-12
Some kind of changes that possibly make sense in master

mih:
I don't think this development model is good for anything. There is no point in having a master branch when it is neither a development nor a release branch.

I said it before, I say it again: let's drop master and go back to 0.9 until the magic date in the future when this thing happens that will be revealed to us when the time is ripe.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,6 +36,7 @@ matrix:
     # By default no logs will be output. This one is to test with log output at INFO level
     env:
     - _DATALAD_UPSTREAM_GITPYTHON=1
+    - _DATALAD_UPSTREAM_GITANNEX=1
     - _DATALAD_NONPR_ONLY=1
     # Just so we test if we did not screw up running under nose without -s as well
     - NOSE_OPTS=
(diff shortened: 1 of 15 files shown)


yarikoptic:
master is the development!  That is where the great things happen! ;)
0.9.x is the "stabilization"/bugfix branch etc. so we could have less great, but working state. 
Overall, I do not feel so much attached to the names -- they are just pointers.  I agreed in the past for master to become the "development" and have 0.9.x to absorb the fixes.  I can agree now to go back.  But I do not see a point in such dances because they do not solve anything IMHO.
We could make 0.9.x to be the default one though on github so any new PR would go by default against it since those are more frequent.

I truly hope that some time soon I will update datasets.datalad.org so it is compatible with both (unles there would be a notable impact somehow on .git/objects or whatnot in the superdatset) and then use master more.  But as during the last hangout I have already whined (i.e. expressed myself) -- I do not see it yet easy to use (either due to bugs, or rich but more complex interface to search, or just things taking forever etc).  And even if i switch to use master, I think we will need to keep absorbing fixes into 0.9.x since that is what people install ATM (unless they apparently listened to you promoting to go with the master).  We simple can't release a new version which doesn't do what previous did.

mih:
OK, so what is the alternative to this "dance" that puts me in the position to keep developing in master, but still be able to run tests independent of my user config?

yarikoptic:
didn't you just acquired those super-powers by this merge?

PR #2026, opened 2017-12-14
Unicode issue in error reporting, crashes datalad

mih:
#### What is the problem?

% datalad create-sibling -s mddatasrc XXX:/media/data/datasets/aggdemo --as-common-datasrc XXX --target-url http://XXX/aggdemo/.git     
[INFO   ] Connecting ... 
[INFO   ] Considering to create a target dataset XXX/scratch/aggdemo at /media/data/datasets/aggdemo of XXX 
Failed to run ['ssh', '-o', 'ControlPath="XXX/.cache/datalad/sockets/e43b2233"', 'XXX', 'export "PATH=/usr/bin:$PATH"; mkdir -p /media/data/datasets/aggdemo'] under None. Exit code=1. out= err=mkdir: cannot create directory ‘/media/data/datasets/aggdemo’: Permission denied
  File "/usr/bin/datalad", line 8, in <module>
    main()
  File "/usr/lib/python2.7/dist-packages/datalad/cmdline/main.py", line 381, in main
    if hasattr(exc.stderr, 'encode')  \
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 31: ordinal not in range(128)

% datalad --version
datalad 0.9.1

### DIFF
--- a/datalad/cmdline/main.py
+++ b/datalad/cmdline/main.py
@@ -377,13 +377,9 @@ def main(args=None):
                 if exc.msg:
                     os.write(2, exc.msg.encode() if isinstance(exc.msg, text_type) else exc.msg)
                 if exc.stdout:
-                    os.write(1, exc.stdout.encode()) \
-                        if hasattr(exc.stdout, 'encode')  \
-                        else os.write(1, exc.stdout)
+                    os.write(1, exc.stdout.encode() if isinstance(exc.stdout, text_type) else exc.stdout)
                 if exc.stderr:
-                    os.write(2, exc.stderr.encode()) \
-                        if hasattr(exc.stderr, 'encode')  \
-                        else os.write(2, exc.stderr)
+                    os.write(2, exc.stderr.encode() if isinstance(exc.stderr, text_type) else exc.stderr)
                 # We must not exit with 0 code if any exception got here but
                 # had no code defined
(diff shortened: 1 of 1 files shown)


datalad/cmdline/main.py
```diff
@@ -377,13 +377,9 @@ def main(args=None):
                 if exc.msg:
                     os.write(2, exc.msg.encode() if isinstance(exc.msg, text_type) else exc.msg)
                 if exc.stdout:
-                    os.write(1, exc.stdout.encode()) \
-                        if hasattr(exc.stdout, 'encode')  \
-                        else os.write(1, exc.stdout)
+                    os.write(1, exc.stdout.encode() if isinstance(exc.stdout, text_type) else exc.stdout)
```
yarikoptic: Hm... Could you describe for my education how it was failing before and now doesn't? (Since there is no test)

mih: The message is in the top-level comment. This is all I got.

mih: And obviously after the change I could successfully run that command. Sorry, this is just a quick fix to get it working, no time for further forensics now.

mih:
OK, so even if this change would not fix a bug, standardizing on a single approach to encoding is a good idea. Will merge.

PR #2033, opened 2017-12-15
BF: pyld is no longer needed

mih:


mih:
Failure on nd90 buildbot is unrelated, but has the same error as I had before locally, in a different context. I think this is actually masking another issue.

```
WARNING: Failed to access ds666, not attempting to crawl S3: 'SilentConsoleLog' object has no attribute 'question' [__init__.py:__getattribute__:70]
```

PR #2035, opened 2017-12-16
require GitPython >= 2.1.8 to assure compatibility with git 2.15

yarikoptic:
2.1.8 is compatible with fresh git 2.15.  I just uploaded 2.1.8 debian packages and it was released on pypi already.   For consistency we better just require this version

### Changes
- [x] This change is complete


yarikoptic:
Hm, codecov report makes little sense... I might resubmit PR, may be it picked the master instead of 0.9.

PR #2037, opened 2017-12-16
BF: Do not send files without content to metadata parsers

mih:
Parsers may not be prepared to deal with this situtation and crash,
taking down the entire aggregation effort.

Some parsers may be able to infer some metadata from filenames, but in
general there is no point in feeding no-content files to parsers.

In general to-be-aggregated datasets should have content.

### DIFF
--- a/after1.txt
+++ b/after1.txt
@@ -1,0 +1,8 @@
+# About this dataset
+
+## General information
+
+This is a DataLad dataset.
+
+For more information on DataLad and on how to work with its datasets,
+see the DataLad documentation at: http://docs.datalad.org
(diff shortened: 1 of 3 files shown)


yarikoptic:
seems needs some code fixing since many tests start to crash

mih:
I cannot replicate the failure of the travis build with the symlink TMPDIR locally. No idea, looks suspicious. Any idea @yarikoptic 

And the two other failures are in direct mode, but the respective test is flagged as "@known_failure_direct_mode  and @known_failure_v6" I don't understand why it is even running. Any idea @bpoldrack?

yarikoptic:
2 @mih: symlink one is in the test which fails once in a while in other scenarios and differently as well from time to time... I reran that matrix run, if this time passes -- should be ok.  dunno about other failures

PR #2046, opened 2017-12-20
ENH: use http://datasets-tests.datalad.org for /// while running tests

### LINKED ISSUE #2044: Stop using /// in tests
#### What is the problem?
it makes it impossible to figure out popularity of datasets, and of the /// altogether since tests are also ran in CI, which goes through a wide range of IPs, thus making it virtually impossible to figure out if access to a dataset is real or a "test".  E.g.

```bash
> zcat datasets.datalad.org-access-comb.log.2.gz | sed -n -e '/"git\/[12345]/s,^\([0-9\.]*\) .*"GET \(.*\)\.git/.*,\1 \2,gp'  | sort | uniq | awk '{print $2;}' | sort | uniq -c  | sort -n  | grep '^ *[0-9][0-9]'
     12 /workshops/nih-2017/ds000114/derivatives/fmriprep/
     12 /workshops/nih-2017/ds000114/derivatives/freesurfer/
     15 /workshops/nih-2017/ds000114/
     24 /dbic/QA/
     46 /labs/tarr/
     46 /labs/tarr/face_place/
     46 /nonexisting/
     48 /labs/
    168 /openfmri/ds000114/
    270 /
    276 /crcns/
```
made me think that `face_place` is very popular, just to find that we are using it in `test_remove_recursive_2` :-(

I think we should
1. change the tests to use some other, dedicated collection (could be a snapshot of current/some datasets.datalad.org)
2. add the test of the /// into buildbot set of tests though, so we do verify healthy state of ///

yarikoptic:
Fixes #2044

### Changes
- [x] This change is complete

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,7 @@ env:
     - NOSE_SELECTION_OP="not "   # so it would be "not (integration or usecase)"
     # Special settings/helper for combined coverage from special remotes execution
     - COVERAGE=coverage
+    - DATALAD_DATASETS_TOPURL=http://datasets-tests.datalad.org
   matrix:
     - DATALAD_REPO_DIRECT=yes
     - DATALAD_REPO_VERSION=6
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -356,7 +356,10 @@ Various hints for developers
 ### Useful Environment Variables
 Refer datalad/config.py for information on how to add these environment variables to the config file and their naming convention
 
-- *DATALAD_LOG_LEVEL*: 
+- *DATALAD_DATASETS_TOPURL*:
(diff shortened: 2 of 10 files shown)


.travis.yml
```diff
@@ -20,6 +20,7 @@ env:
     - NOSE_SELECTION_OP="not "   # so it would be "not (integration or usecase)"
     # Special settings/helper for combined coverage from special remotes execution
     - COVERAGE=coverage
+    - DATALAD_DATASETS_TOPURL=http://datasets-tests.datalad.org
```
yarikoptic: so that examples runs etc take it in effect

yarikoptic:
I think coverage is just lost its senses (according to #2047) so merging

PR #2047, opened 2017-12-20
DO NOT MERGE empty commit just to test effect on coverage

yarikoptic:
Other PRs strangely show strange decreases in coverage -- this PR atm introduces no changes intentionally 
e.g. looking at https://codecov.io/gh/datalad/datalad/pull/2046/src/datalad/tests/utils.py#L427 -- codecov states that the line was covered before but not now.  But I do not see any test which (used to) use recently `ok_annex_get` with a list of files and not just one (previous condition, which is covered)

or https://codecov.io/gh/datalad/datalad/pull/2046/src/datalad/tests/utils.py#L500  -- how `__init__` declaration line could go from red to green while body from green to red.

I have also dropped a question to codecov support chat, was told that they will come back to it tomorrow

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,6 @@ matrix:
     env:
     - _DATALAD_UPSTREAM_GITPYTHON=1
     - _DATALAD_UPSTREAM_GITANNEX=1
-    - _DATALAD_NONPR_ONLY=1
     # Just so we test if we did not screw up running under nose without -s as well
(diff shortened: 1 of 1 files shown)


yarikoptic:
so here we go... I really doubt that it is coming due to difference in having some other runs in the master -- but I will enable all of them now and see if anything changes

yarikoptic:
yeah, seems something weird with coverage... heh

Issue #2041, opened 2017-12-17
attemp to kill datalad process results in breeding git <defunct> processes


yarikoptic:
#### What is the problem?
As reported in #2036 aggregate-metadata got stuck, decided to Ctrl-C it but it keeps "resisting" -- doesn't quit:

```
Exception ignored in: <bound method AnnexRepo.__del__ of <AnnexRepo path=/etc (<class 'datalad.support.annexrepo.AnnexRepo'>)>>
Traceback (most recent call last):
  File "/home/yoh/proj/datalad/datalad-master/datalad/support/annexrepo.py", line 323, in __del__



^C^CException ignored in: <bound method GitRepo.__del__ of <GitRepo path=/home/yoh/proj/datalad/datalad-master (<class 'datalad.support.gitrepo.GitRepo'>)>>
Traceback (most recent call last):
  File "/home/yoh/proj/datalad/datalad-master/datalad/support/gitrepo.py", line 732, in __del__
    if hasattr(self, 'repo') and exists(self.path) \
  File "/home/yoh/proj/datalad/datalad-master/datalad/support/gitrepo.py", line 618, in repo
    self._repo = self.cmd_call_wrapper(Repo, self.path)
  File "/home/yoh/proj/datalad/datalad-master/datalad/cmd.py", line 134, in __call__
    return self.call(cmd, *args, **kwargs)
  File "/home/yoh/proj/datalad/datalad-master/datalad/cmd.py", line 492, in call
    return f(*args, **kwargs)
  File "/home/yoh/proj/datalad/datalad-master/datalad/support/gitrepo.py", line 387, in Repo
    return gitpy.Repo(*args, **kwargs)
  File "/home/yoh/proj/datalad/datalad-master/venvs/dev3/lib/python3.5/site-packages/git/repo/base.py", line 165, in __init__
    self._bare = self.config_reader("repository").getboolean('core', 'bare')
  File "/home/yoh/proj/datalad/datalad-master/venvs/dev3/lib/python3.5/site-packages/git/config.py", line 78, in assure_data_present
    self.read()
  File "/home/yoh/proj/datalad/datalad-master/venvs/dev3/lib/python3.5/site-packages/git/config.py", line 404, in read
    with open(file_path, 'rb') as fp:
  File "/home/yoh/proj/datalad/datalad-master/datalad/auto.py", line 165, in _proxy_open
    *args, **kwargs)
[...]

Issue #1988, opened 2017-11-23
Wrong error message due to automagic IO


mih:
```
$ datalad search --reindex
[INFO   ] Building search index 
[INFO   ] Scanning for metadata keys 
[INFO   ] AutomagicIO: retrieving file content of .datalad/metadata/objects/99/cn-0c4616eeead2d84b6763179f755bc0                                                                   
[ERROR  ] [Errno 2] No such file or directory: u'.datalad/metadata/objects/99/cn-0c4616eeead2d84b6763179f755bc0' [auto.py:_proxy_open_name_mode:160] (IOError) 

$ git annex get .datalad/metadata/objects/99/cn-0c4616eeead2d84b6763179f755bc0
get .datalad/metadata/objects/99/cn-0c4616eeead2d84b6763179f755bc0 (not available) 
  Try making some of these repositories available:
        6a2fd65b-95fc-4037-9850-26ed1d8fd868 -- xxx@xxx:/users/.../3Tdaten
failed
failed
git-annex: get: 1 failed
```

The file handle is there, but it cannot get the content.

yarikoptic:
well, I would not say it is "wrong", it is just insufficient ;)
Adding an additional error message but still allowing to proceed and fail how the corresponding `open` call would fail (that message came from it):
```shell
$> datalad search --reindex a
[INFO   ] Rebuilding search index 
[INFO   ] Scanning for metadata keys 
[INFO   ] AutomagicIO: retrieving file content of /tmp/ds000001-3/.datalad/metadata/objects/87/ds-1d35795a1ba0410f271672d1f11d7c 

[ERROR  ] Failed to retrieve .datalad/metadata/objects/87/ds-1d35795a1ba0410f271672d1f11d7c: Try making some of these repositories available:
	2b0836f4-b362-403c-87d5-8631d45b97fb -- yoh@hopa:/tmp/ds000001
[ERROR  ] [Errno 2] No such file or directory: '/tmp/ds000001-3/.datalad/metadata/objects/87/ds-1d35795a1ba0410f271672d1f11d7c' [auto.py:_proxy_open_name_mode:193] (IOError) 
```

PR #2050, opened 2017-12-21
A variety of tuneups to AutomagicIO to avoid redoing the same queries in 'search'

yarikoptic:
It might fix the "stalling" I have reported somewhere (probably was closed since "happened on smaug")
It should make AutomagicIO of lesser burden on 'search' since before now it would query for the same files over and over again , and look for git repo for every directory etc. So now

- gets a mode to consider each path only once
- caches repos for each directory considered
- skips paths containing `/.git/` since also not expecting anything to be under annex control there
- assumes that there could be no subdataset under .datalad (for now) which helps to avoid
  looking up for .git for every bloody objects/XX directory (funny part is that then it was all in vein in my case since repo is anyways is git, not annex)
- a bit improved and harmonized logging there
- potentially, at least partially addresses #2041 (thanks @mih for digging the # out)

Disclaimer: some of the fixes will make also into 0.9.x branch later on

### DIFF
--- a/datalad/auto.py
+++ b/datalad/auto.py
@@ -62,8 +62,10 @@ except Exception as exc:
 # TODO: RF to reduce code duplication among cases, also RF tests for the same reason
 
 class _EarlyExit(Exception):
-    """Helper to early escape try/except logic in wrappde open"""
-    pass
+    """Helper to early escape try/except logic in wrapped open"""
+    def __init__(self, msg, *args):
+        self.msg = msg
+        self.args = args
 
 
 class AutomagicIO(object):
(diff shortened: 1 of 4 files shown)


mih:
The issue you reported this is #2041.

datalad/auto.py
```diff
@@ -132,13 +151,22 @@ def _proxy_open_name_mode(self, origname, origfunc, *args, **kwargs):
...
+                if self._paths_cache is not None:
+                    filefull = file if isabs(file) else os.path.abspath(file)
+                    if filefull in self._paths_cache:
+                        raise _EarlyExit("considered before")
+                    else:
+                        self._paths_cache.add(filefull)
+
+                # TODO: Windows-proof?
+                if '/.git/' in file:
+                    raise _EarlyExit("we ignore paths under .git/")
```
mih: I guess that would not achieve the desired effect on windows. Why not `os.path.sep`?

mih:
Thanks!

mih:
Any chance to look into a solution for #1988 in this context?

yarikoptic:
ok - both aspects addressed now:
- more informative message
- a bit more friendliness to the land of Windows

yarikoptic:
the travis failure is unrelated, saw it before on and off, trying to catch locally but so far no luck.  So, from my perspective this one is ready @mih unless you spot more issues

PR #2048, opened 2017-12-22
datalad clean should remove search index

mih:


yarikoptic:
Good for starters ;-) issues info should (also) be tested after search command to assure that code actually creates that directory (now manually populated in the test)

mih:
I put that onto the golden todo list. Will merge for now and feel bad about having failed....

PR #2066, opened 2018-01-09
NF: decompress_file fixes possibly corrupted tree

bpoldrack:
This pull request fixes ...
... an issue I experienced with Windows-created tarballs, which lack the execution bit of directories.

This pull request proposes ...
... to just set it. Especially in the case of temporary extraction, there is no easy other way.

### Changes
- [x] This change is complete
- [ ] This is in progress


datalad/support/archives.py
```diff
@@ -142,6 +142,23 @@ def decompress_file(archive, dir_, leading_directories='strip'):
         if cmo.err:
             lgr.debug("patool gave stderr:\n%s" % cmo.err)
 
+    # Note: (ben) Experienced issue, where extracted tarball
+    # lacked execution bit of directories, leading to not being
```
yarikoptic: do you have an example?  could there be  a tiny unittest for such a case?

bpoldrack: Well, I have huge dicom tarballs that have that issue. But I don't know how to create them on purpose, since I only consume them.
The only way to test would be to mock that extracting result and see that the added code is actually setting the X bit. But then: What we really need to test for would be sideeffects and this is basically done by not breaking other tests, no?

bpoldrack:
FTR: Failures on Travis are all the same and apparently unrelated:

```
ERROR: datalad.crawler.pipelines.tests.test_crcns.test_get_metadata
    self.test(*self.arg)
  File "/home/travis/build/datalad/datalad/datalad/tests/utils.py", line 841, in newfunc
    return func(*args, **kwargs)
  File "/home/travis/build/datalad/datalad/datalad/crawler/pipelines/tests/test_crcns.py", line 29, in test_get_metadata
    all_meta = get_metadata()
  File "/home/travis/build/datalad/datalad/datalad/crawler/pipelines/crcns.py", line 72, in get_metadata
    rj = fetch_datacite_metadata()
  File "/home/travis/build/datalad/datalad/datalad/crawler/pipelines/crcns.py", line 44, in fetch_datacite_metadata
    text = get_cached_url_content(arx, name='crcns', maxage=1)
  File "/home/travis/build/datalad/datalad/datalad/support/network.py", line 936, in get_cached_url_content
    doc = fetcher(url)
  File "/home/travis/build/datalad/datalad/datalad/downloaders/providers.py", line 321, in fetch
    return self.get_provider(url).get_downloader(url).fetch(url, *args, **kwargs)
  File "/home/travis/build/datalad/datalad/datalad/downloaders/base.py", line 472, in fetch
    out = self.access(self._fetch, url, **kwargs)
  File "/home/travis/build/datalad/datalad/datalad/downloaders/base.py", line 188, in access
    "You would need to configure data provider authentication using TODO " % url)
AccessDeniedError: Access to http://search.datacite.org/api?q=datacentre_symbol:cdl.ucbcrcns&fl=doi,minted,updated,xml&fq=has_metadata:true&fq=is_active:true&rows=1000&start=0&sort=updated+asc&wt=json was denied but we don't know about this data provider. You would need to configure data provider authentication using TODO 
```

yarikoptic:
Heh, since it is denied and some did pass, I wonder if it is done kind of throttling they introduced. Ok, will deal with it if becomes frequent event

datalad/support/archives.py
yarikoptic: Correct. I just thought that may be there is no mystery how this tarball came about. My wild guess it's that it was done programmatically, and then some bits, like TarInfo.mode if was done in python, were not set properly 😉

bpoldrack: FWIW: It's done on Windows, so this mystery could possibly involve all kinds of shit. ;-)

yarikoptic: Sounds like a time to think about adding appveyor setup to run at least subset of tests of windows (e.g. eliminating those which we know to fail with direct mode etc).  #2083 .  Then I guess we could have observed/tested that behavior

Issue #2089, opened 2018-01-19
One test failure and one error with git 2.16.0


yarikoptic:
https://lwn.net/Articles/744749/

New release, has many changes relating to submodules handling. Would be good to check if no new issues or incompatibilities, and report here... Reporting with #2090 , on 0.9.x branch I observe:
```shell
$> DATALAD_TESTS_SSH=1 PATH=/home/yoh/proj/misc/git/INSTALL-2.16.1/bin:$PATH DATALAD_USE_DEFAULT_GIT=true python -m nose -s -v datalad.distribution.tests.test_uninstall:{test_drop_nocrash_absent_subds,test_uninstall_without_super}
datalad.distribution.tests.test_uninstall.test_drop_nocrash_absent_subds ... ERROR
datalad.distribution.tests.test_uninstall.test_uninstall_without_super ... FAIL
Versions: appdirs=1.4.3 boto=2.44.0 cmd:annex=6.20171211+gitgba511f4de-1~ndall+1 cmd:git=2.16.1 cmd:system-git=2.16.1 cmd:system-ssh=7.6p1 git=2.1.8 gitdb=2.0.2 humanize=0.5.1 iso8601=0.1.11 msgpack=0.4.8 requests=2.18.4 scrapy=1.4.0 six=1.11.0 tqdm=4.19.5 wrapt=1.9.0

======================================================================
ERROR: datalad.distribution.tests.test_uninstall.test_drop_nocrash_absent_subds
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/yoh/proj/datalad/datalad/datalad/tests/utils.py", line 627, in newfunc
    return t(*(arg + (filename,)), **kw)
  File "/home/yoh/proj/datalad/datalad/datalad/distribution/tests/test_uninstall.py", line 522, in test_drop_nocrash_absent_subds
    assert_status('notneeded', drop('.', recursive=True))
  File "/home/yoh/proj/datalad/datalad/datalad/interface/utils.py", line 437, in eval_func
    return return_func(generator_func)(*args, **kwargs)
  File "/home/yoh/proj/datalad/datalad/datalad/interface/utils.py", line 425, in return_func
    results = list(results)
  File "/home/yoh/proj/datalad/datalad/datalad/interface/utils.py", line 382, in generator_func
[...]

PR #2090, opened 2018-01-23
DATALAD_USE_DEFAULT_GIT to override detection and use default git in the PATH

yarikoptic:
Introduced to answer the question in #2089 

### Changes
- [x] This change is complete


yarikoptic:
also hindered by datacite :-/

yarikoptic:
otherwise works, will merge ;)

PR #2096, opened 2018-01-24
ENH(webui): autofocus on searchbox when entering a page

yarikoptic:
Allows for quick(er) navigation of the site

For now I just quickly manually patched website's .js to achieve that so you could try

the only cons I am aware of would be for vimperator and alike users -- you would need to press Esc first to make shortcuts of those extensions work instead of just being in edit mode. But imho it is a small price to pay ;)

### Changes
- [x] This change is complete


kyleam:
That merge was forced by @yarikoptic

PR #2087, opened 2018-01-25
Degrade git-annex metadata from having the last word to a mere metadata source among many

mih:
#### What is the problem?

git-annex has simplistic metadata capabilities in contrast to other metadata sources
  - constraints of key name composition
  - stored directly in git, hence bloats the repo if things change or are heavy
  - only considers annexed files

However, the current `master` git-annex metadata is queried as a special metadata source that is used to override/fix metadata aggregated from other sources (with the above limitations). Moreover, this actually brings @yarikoptic to tears when he aggregates the main superdataset, as git-annex is asked to reports its knowledge about any annexed files in any dataset, regardless of whether there was any change -- simply because all datasets are locally present. The same happens on index generation: when a dataset is locally present, git annex gets re-querried to check if there have been any metadata changes since the last aggregation -- this makes search index generation seemingly very slow when subdatasets are locally present.

Given all the above, I propose to treat git-annex metadata as just another metadata source, implement a parser for it (pretty much done already in the code base), and aggregate it like anything else. When metadata is requested, simply report locally aggregated metadata, do not query git annex again.

This will achieve two things:
- makes things substantially faster when using/querying metadata (apart from aggregation)
- does not impose git-annex' limits on keys and what it can store metadata for on other parts of datalad.

Implementing a layer capable of amending any arbitrary metadata can be done in the future, once the need arises.


mih:
Note that this PR is for documentation purposes only. It goes into a branch that will accumulate a substantial pile of changes.

PR #2108, opened 2018-01-26
rf-grammar-check

overlake333:
Fixing grammar and spelling errors

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -122,7 +122,7 @@ class AnnexRepo(GitRepo, RepoInterface):
         Parameters
         ----------
         path: str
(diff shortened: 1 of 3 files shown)


datalad/support/annexrepo.py
```diff
@@ -138,7 +138,7 @@ def __init__(self, path, url=None, runner=None,
           that are already annexed nor will it automatically migrate files,
           hat are 'getted' afterwards.
         create: bool, optional
-          Create and initializes an annex repository at path, in case
+          Creates and initializes an annex repository at path, in case
```
yarikoptic: it is a flag (type bool) to change behavior of the function.  So it is better to use imperative form as a continuation of the sentence "If you set this flag, this function will ...", so a better fix would have been to fix the second verb.

Issue #2088, opened 2018-01-19
Find a new (dedicated) home for our metadata parsers


mih:
We might still carry some kind of adapter, but we should use common implementations from a place where they are more exposed.
@jbpoline might generate an idea where such place could be.

mih:
Please see #2326 for a related RF.

mih:
Metadata parsers can now be provided by any extension via the setuptools `entry_points` mechanism. See https://github.com/datalad/datalad-neuroimaging/blob/master/setup.py for an example of an extension setup that provides multiple additional extractors.

Issue #2086, opened 2018-01-18
RFC: "Aggregate" original metadata representations up the hierarchy


yarikoptic:
#### What is the problem?

There is no problem ;)  Just an idea which came up from discussion the NIDM folks and possibly useful to discuss/keep in mind:

- datalad metadata parsing ATM concentrates on extraction and aggregation (across datasets) key/value metadata pairs
- ATM we are using whoosh, but as reiterated recently in https://github.com/datalad/datalad/issues/2082#issuecomment-358249883, there could be other applications built on top of aggregated metadata, e.g. a full SPARQL query engine
- PROV/NIDM provides *richer* metadata beyond key-value pairs.  

So, to facilitate the best utility for the aggregated metadata to be used by various applications, either our parsers/aggregation should be as expressive as any possible metadata representation, and then possibly just duplicate what is provided in the original (to be parsed) `.json` file, or may be we could just "aggregate" and propagate them "as-is" all the way up, so they could be used (or ignored) by the application (while constructing the index).

From other perspective, all our metadata parsers now harmonize metadata found in the dataset into jsonld for `ds-` and a list of key/value pairs for each file within `cn-` as the "metadata of the dataset".  Then whoosh index is aware of how to build search index from it.  Parsers are nohow participate in construction of the index.  But we could also just propagate other possible metadata descriptors (e.g., `nidm-` file with NIDM metadata representation) and extend parsers' API to implement "software specific" handling during index construction.  Then for whoosh it would extract key value pairs from `nidm-` portion.  for sparql engine, just return already consumable by it representation.

CC @satra @jbpoline @dbkeator @gkiar

mih:
This is already in the works.

mih:
To further clarify: We already need the ability to aggregate arbitrarily nested metadata for other reasons. Supporting NIDM and others is a special case of that AFAICS, hence this will come, and is actually already in progress. Also, AFAICS the only piece of code affected (or limiting right now) is the merging of values from multiple parser -- the parsers themselves can stay as they are (conceptually). I already have an alternative implementation that works with deeply nested values, based on the principle "if anything in such a structure is different from another one, the entire nested structure is considered as a different/unique value". This works nicely, so far.

[...]

PR #2045, opened 2017-12-20
RF: Metadata with arbitrarily nested structures

mih:
This change requires a pybids that is yet to be released after the merge of https://github.com/INCF/pybids/pull/94 I have adjusted the requirements to install pybids from its master branch in this PR. Ideally this would be reverted after a pybids release, before this PR is merged. So this is a wonderful opportunity to create yet another monster PR....

This pull request fixes #1957

Thanks to @tyarkoni!

PR #2095, opened 2018-01-28
Change representation of aggregated metadata

mih:
#### What is the problem?

There is a sudden interest in metadata that could change the pace of development and we better get the Datalad internal metadata representation right now.... Here is a dump of a brain storming with @bpoldrack 

There are a bunch of issues with the current state in master, examples:
- #2086 handling of individual metadata sources as "blobs"
- #2087 using git-annex capabilities is least common denominator
- #2088 factoring out the parser logic
- ...

We want our internal representation to be as agnostic as possible, while maintaining the integrity of aggregated metadata sources as much as possible. ATM we "merge" metadata from multiple sources into a single set. We support multiple strategies, but overall the handling is rather clumsy. Merging actually causes order issues and could have unforeseen implications when we start using many metadata sources with partially overlapping term.

That being said, in reality one parser produces mostly keys with a common prefix and merges only affect the most basic common keys (name, author, ...).

Proposal:

- we stop merging during aggregation, but instead aggregate per metadata source (a NIDM metadata set stays as-is, we are still handling it as part of the general metadata structure, not as a dedicated file)
- we generate unique values also per metadata source (for dataset discovery)
- we store metadata from a single source under its own key in a dedicated dictionary, hence something like this becomes the standard (keys are names of metadata sources)
```
metadata = {
  'bids': {...},
  'nifti': {...},
  'nidm': {...}
}
```
- keys (and also values) produced by a metadata source will not be modified in any way (no prefixing of any kind)
- appropriate flattening of the metadata structures (now done by an incremental merging strategy during aggregation), becomes a responsibility of a search/query application _using_ the aggregated metadata. It will be its responsibility to select desired metadata sources, flatten/prefix keys for a tabular search index, ...
   - at this layer we can also support mappings between terms that may be developed by efforts like interlex, e.g. if somebody tells us that `bids:resolution` is the same as `nifti:resolution`, or maybe tells us how to transform one to become the other (units, ....)
- now that metadata sources stay independent, they can each have their own "context" (in JSON-LD speak), i.e. a (separate) set of term definitions. We probably want to create a dedicated space in the dataset's global metadata to store a uniform context/definition set per metadata source. But ultimately it is up to the metadata sources to use/define this -- datalad just provides the technical means.
- keeping metadata sources separate will also enable straightforward incremental aggregation of individual metadata sources
- we likely want to implement a layer that allows for amending/fixing aggregated metadata to enable modification without having to change the actual dataset content, but that is an orthogonal issue to this one
- git-annex will become just-another-metadata-source, instead of being the top-layer #2087 

`datalad from_the_future`

@yarikoptic @jbpoline @dbkeator @gkiar


datalad/metadata/parsers/tests/test_nidm.py
```diff
@@ -0,0 +1,63 @@
...
+    assert_status('ok', res)
+    res = ds.metadata(reporton='all')
+    # ATM we do not forsee file-based metadata to come back from NIDM
+    assert_result_count(res, 1)
+    # show full structure of the assembled metadata from demo content
+    assert_result_count(
+        res, 1,
+        metadata={
+            "@context": {
+                "@vocab": "http://docs.datalad.org/schema_v2.0.json"
```
mih: In the base datalad vocabulary (this is built from this source: https://github.com/datalad/datalad/blob/master/datalad/metadata/definitions.py) we would have to identify the `nidm` namespace.

datalad/metadata/parsers/tests/test_nidm.py
```diff
@@ -0,0 +1,63 @@
...
+                        "@id": "time:Duration"
+                    },
+                    "myvocabprefix": {
+                        "@id": "http://purl.org/ontology/mydefinition",
+                        "description": "I am a vocabulary",
+                        "type": "http://purl.org/dc/dcam/VocabularyEncodingScheme"
+                    }
+                },
+                "mydurationkey": 0.6
+            }
```
mih: Anything in the dict under the `nidm` key is completely left to NIDM. Datalad will carry any content and will reproduce it on query, even when the dataset is not locally present, but only the metadata that was aggregated from it.

mih:
Next step is to absorb the remaining pieces from #2045

mih:
It could well be that the change in metadata layout is already complete at this stage. However, that won't be clear until an application (like `search`) is built on top. ATM we have no tests whatsoever to tell that (I know that search is broken at this point). Will continue in next PR for clarity.

PR #2111, opened 2018-01-28
Minimal RF of old search code for new class layout

mih:
Blast from the past: here is the old `search` code, minimally changed to not blow up completely when used with new metadata (as of #2095). I do not plan to invest more time into this, but interested parties could ;-)

Biggest issues in this context is that `///` would need to get new style aggregate metadata before any of this could be fully implemented.

Old tests are included for convenience, but no attempt was made to make them work. Only manual playing around with http://psydata.ovgu.de/aggdemo/demo/ (a toy sibling of `///`).

@yarikoptic only the last commit is relevant, the rest is #2095


mih:
Closing this one right away, but it is on the record now.

yarikoptic:
THANKS!  I promise prepare the /// again with aggregated metadata new style (using master or some particular PR?) in few days so we could test on a target use case

mih:
From my POV it makes no sense to use master at this point anymore.  #2112 is relevant now, which I hope to have finished some time this week.

yarikoptic:
So should I wait for #2112 to come to completion before reaggregating for testing?

mih:
FTR: #2115 has a much better implementation of a search in the spirit of the old command. Moreover, it does not require a completely different implementation, but more or less just a different (default) configuration. From my POV this PR should not be continued.

PR #2112, opened 2018-01-29
Supersedes  #2045 (arbitrarily nested metadata) built on top of the new aggregated metadata layout introduced with #2095

mih:
This PR will also RF search to make the changes complete from a use perspective.


mih:
@bpoldrack Last commit contains broken search queries. Could you please have a look at the plugin config? Thanks!

mih:
Last commit should bring this PR into a merge-able state. Will do when tests pass. At that point #2111 could be rebased onto datalad:enh-metadata.

PR #2115, opened 2018-02-02
This PR has the accumulated metadata changes that need to get into master

mih:


mih:
Buildbots cannot succeed, because they lack a good pybids version.

yarikoptic:
Master testing already broken due to pybids and forces await new release of it to realize. Difference is huge and I bet it just makes master even better, so I will merge it as is

yarikoptic:
It "needed the best merge!" and I've delivered!

PR #2119, opened 2018-02-06
Store 'offline' output (typically stderr) into a file to avoid lockdowns

### LINKED ISSUE #2116: datalad add  would get stuck if there is only errors (wrong permissions)
#### What is the problem?

The same from the series of us not handling situations with lots of stderr output, here is the commands to reproduce with force ;)

```shell
cd /tmp/; sudo rm -rf /tmp/test && datalad create /tmp/test ; cd /tmp/test;
for f in longfilenametogetlotsoferrorsandevenlonger{1..1000}; do echo $f>$f; done
datalad add *
```
and it will be stuck since git annex spits lots of 
```
  longfilenametogetlotsoferrorsandevenlonger98: setFileMode: permission denied (Operation not permitted)
  longfilenametogetlotsoferrorsandevenlonger980: setFileMode: permission denied (Operation not permitted)
```

yarikoptic:
This pull request fixes #2116 

This pull request proposes
- [x] store output marked as 'offline' into a file and then load from file whenever "done"
- [x] report at least a portion of stderr if git annex --json exited abnormally (e.g. as if permissions didn't allow `add`).  Ideally I guess all that code should become a generator, but not sure how feasible it is anyways since stderr isn't necessarily "coupled" with stdin so we wouldn't be able to generally assign specifically stderr to particular files.  Hence: http://git-annex.branchable.com/todo/include_msg_with_possible_reason_why_command___40__e.g._add__41___failed_into_--json_output/
- [x] more tests revealed an issue on OSX on not having processed lines which are still available upon process completes (.poll returns None).  So it demanded heavy code refactoring.  I am not sure if it became anyhow easier to follow but at least it became a bit more modular.  Also removed some conditions checking since seems to work without ;-)  

overall -- it came out much larger than what I hoped/expected/initially did.  But hopefully would be resilient enough to not introduce regressions

This might result in negative impact on performance while calling those batch commands (since would require creation of a file) but imho better slower but more reliable ;)

Unless immediate objections, I would like to merge it asap.  We could fix up/refactor later if needed, but "stuck" operation of datalad without any error msg isn't fun

### Changes
- [x] This change is complete

### DIFF
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -347,7 +347,7 @@ class Runner(object):
                 raise RuntimeError("must not get here")
             return line
         # it was output already directly but for code to work, return ""
-        return ""
+        return binary_type()
 
     def run(self, cmd, log_stdout=True, log_stderr=True, log_online=False,
             expect_stderr=False, expect_fail=False,
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -298,12 +298,11 @@ class Runner(object):
             if log_stderr_:
                 stderr += self._process_one_line(*stderr_args)
 
-        if log_stdout in {'offline'} or log_stderr in {'offline'}:
-            lgr.log(4, "Issuing proc.communicate() since one of the targets "
-                       "is 'offline'")
-            stdout_, stderr_ = proc.communicate()
-            stdout += self._process_remaining_output(outputstream, stdout_, *stdout_args)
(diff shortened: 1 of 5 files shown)


yarikoptic:
Review would be appreciated, but if none comes I would prefer to merge it to finally cook up a bugfix release

yarikoptic:
although this PR touches on the holy grounds all the tests seems to pass, so I will merge it so we could test it "in production"

PR #2125, opened 2018-02-07
BF: avoid sending ERROR to annex if we can still continue

### LINKED ISSUE #2124: mishandling of git-annex external special remote protocol unsupported request
To use new special remote protocol messages like the one added in
http://git-annex.branchable.com/todo/INFO_message_for_custom_special_remotes/ , external special remotes need a way to determine if git-annex supports the command. http://git-annex.branchable.com/design/external_special_remote_protocol has a plan to make git-annex send a new message after VERSION indicating its capabilities.

On auditing all external special remotes to see if they will work well with such a new message, I found all did, except for the one in datalad, which does this:

            method = getattr(self, "req_%s" % req, None)
            if not method:
                self.error("We have no support for %s request, part of %s response"
                           % (req, l))
                self.send("UNSUPPORTED-REQUEST")
                continue

As far as I can tell, this probably crashes. It should simply send back the UNSUPPORTED-REQUEST and let git-annex deal with it.

yarikoptic:
This pull request fixes #2124 

### Changes
- [x] This change is complete

### DIFF
--- a/datalad/customremotes/archives.py
+++ b/datalad/customremotes/archives.py
@@ -361,7 +361,10 @@ class ArchiveAnnexCustomRemote(AnnexCustomRemote):
                 self.debug("Failed to fetch {akey} containing {key}: {exc_}".format(**locals()))
                 continue
 
-        self.error("Failed to fetch any archive containing {key}. Tried: {akeys_tried}".format(**locals()))
+        raise RuntimeError(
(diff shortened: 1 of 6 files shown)


yarikoptic:
ok, coverage increased and hopefully I did not break any behavior... merging

PR #2126, opened 2018-02-09
ENH: Initial use/test ride of pyout -- almost works!

yarikoptic:
pyout 0.2 is ~~not yet~~ **released** ~~-- install from git repo~~

TODOs
- [x] fixup/extend pyout
- [ ] refactor ls code to not list one dataset (recursively or not) at a time but so all the entries processed in the same loop
- [ ] fixup tests


mih:

PR #2127, opened 2018-02-09
"Old" and new search in one command

mih:
This PR RFs the `search` command to be able to handle multiple search modes. Currently supported are:

- `autofield`: this is what used to be the new search (two-pass, autodiscover schema, files and datasets)
- `default`: this is some some ways the old search, in some ways not

The new default behavior is to works on datasets only (but it has the ability to do files as well, unlike the old command). All metadata is currently coerced into a single string using a default whoosh text analyzer. Unit tests document what kind of searches work.

Additional metadata homogenization could be supported by introducing custom mappings (e.g. for neuro* datasets, etc.). This would be necessary to support the previously available 'search by field' in the old command. The current aggregated metadata has no homogenized fields, hence such thing is impossible without a known mapping.

Performance: I can generate a search index for a datasets that contains aggregated metadata from two subdatasets with 36k files each in under a second.

Multiple concurrent search indices for the respective search modes are supported too.

This is the end of the road folks... @datalad/developers 

As before, the buildbots cannot succeed, because the lack a suitable pybids version for the tests.

### DIFF
--- a/datalad/metadata/search.py
+++ b/datalad/metadata/search.py
@@ -382,6 +382,7 @@ def _get_search_index(index_dir, label, ds, force_reindex, get_schema, meta2doc,
             old_ds_rpath = admin['path']
 
         doc.update({k: assure_unicode(v) for k, v in admin.items()})
+        lgr.debug("Adding document to search index: {}".format(doc))
         # inject into index
         idx.add_document(**doc)
         idx_size += 1
--- a/datalad/metadata/search.py
+++ b/datalad/metadata/search.py
@@ -36,7 +36,6 @@ from datalad.support.param import Parameter
 from datalad.support.constraints import EnsureNone
 from datalad.support.constraints import EnsureInt
 from datalad.metadata.metadata import query_aggregated_metadata
-from datalad.metadata.metadata import MetadataDict
 
 from datalad.consts import LOCAL_CENTRAL_PATH
 from datalad.consts import SEARCH_INDEX_DOTGITDIR
@@ -104,7 +103,7 @@ def _listdict2dictlist(lst):
         if len(v)}
 
 
-def _meta2index_dict(meta, val2str=True):
+def _meta2autofield_dict(meta, val2str=True):
     """Takes care of dtype conversion into unicode, potential key mappings
     and concatenation of sequence-type fields into CSV strings
     """
@@ -158,7 +157,7 @@ def _meta2index_dict(meta, val2str=True):
     }
 
 
-def _get_search_schema(ds):
+def _get_schema_autofield(ds):
     from whoosh import fields as wf
     from whoosh.analysis import StandardAnalyzer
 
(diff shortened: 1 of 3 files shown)


datalad/metadata/search.py
```diff
@@ -199,14 +204,86 @@ def _get_search_schema(ds):
...
 
 
-def _get_search_index(index_dir, ds, force_reindex):
+def _get_parser_autofield(idx_obj):
+    from whoosh import qparser as qparse
+
+    parser = qparse.MultifieldParser(
+        idx_obj.schema.names(),
+        idx_obj.schema)
+    # XXX: plugin is broken in Debian's whoosh 2.7.0-2, but already fixed
```
yarikoptic: Should there be a warning at least if broken whoosh is detected or shops we require future whoosh?

datalad/metadata/search.py
```diff
@@ -517,8 +600,32 @@ def __call__(query=None,
...
+
         idx_obj = _get_search_index(
-            index_dir, ds, force_reindex)
+            index_dir,
+            mode,
+            ds,
+            force_reindex,
+            get_schema_fx,
+            meta2doc_fx,
+            documenttype)
```
yarikoptic: 1c: Feels like it could have been earlier to navigate if above two types of searches were done via two classes and this generic one in a common method

datalad/metadata/search.py
mih: Maybe, if we end up using that plugin at all. At present, you would have to explicitly request the 'autofield' search index in order to have this code run.

datalad/metadata/search.py
mih: I didn't get it. What is earlier to navigate and for whom?

yarikoptic: Just navigate the code. Now it is a collection of functions grouped by the common prefix, which are passed around. Just felt that it would be easier to follow and use if it was classes

mih: I agree that this could be optimized. But we should first figure out whether we want this at all. From my perspective this is search v4 and the last three didn't make it into a release. Not the time for optimization, it seems.

yarikoptic:
Re buildbots: is new release of pybids is expected to fix buildbots, or then add git url to requirements.txt to not waste buildbots cycles

yarikoptic:
And I guess they don't know that we need a release... Someone had to talk
Buildbots use regular requirements.txt https://github.com/datalad/buildbot/blob/master/master/master.cfg#L182 but now I wonder if we should change that to devel

mih:
A few PRs back we were in touch re a release. I told Tal that there is no rush. We would need a release when we want to release this ....

mih:
From my POV, it would make sense to use -devel.txt. Would make it more similar to travis and I can imagine deps that we would not want to impose on a regular user, but would be useful for tests.

yarikoptic:
yeah, makes sense -- adjusted/restarted/pushed.  So whenever new changes come in should install using -devel

PR #2129, opened 2018-02-10
ENH: Alias --output-format with -o

mih:
This one is frequently used and quite long to type.

Please merge @datalad/developers

### DIFF
--- a/datalad/cmdline/main.py
+++ b/datalad/cmdline/main.py
@@ -110,7 +110,7 @@ def setup_parser(
         read from a file, but is potentially overridden itself by configuration
         variables in the process environment.""")
     parser.add_argument(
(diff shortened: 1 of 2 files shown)


yarikoptic:
It may be -f? Don't we have a way to redirect kids to a file, so may be eventually -o would make more sense for that.

mih:
True, let's use `-f`. Will also piggy-back another related change.

yarikoptic:
seems to be in need of fixing

mih:
Fixed. Hopefully.

yarikoptic:
For osx fix needs better if 0.9.x into master, otherwise good to go, but can't do in the phone due to above osx failure

Issue #2131, opened 2018-02-10
annex cannot handle filenames that start with a minus


mih:
#### What is the problem?

```
% datalad add .
add(ok): /tmp/lit/-Pals.wcover.pdf (file)                                                                                    
...
Failed to run ['git', '-c', 'receive.autogc=0', '-c', 'gc.auto=0', 'annex', 'add', '--json', u'-Pals.wcover.pdf', ..., '.'] under '/tmp/lit'. Exit code=1. out= err=Invalid option `-Pals.wcover.pdf'
```

Errors with annex help output displayed and exit code 1. Leave all files staged, nothing is commited. A subsequent `save` (no args) achieves the desired result.

Maybe we need to escape the file names that look like options when passing them to annex?

A subsequent `aggregate-metadata` command fails with an analog error of `annex metadata`.

Escaping the minus character seems to do the trick.

yarikoptic:
We probably just need to add `--` before the list of the filenames

yarikoptic:
Was fixed in #2134 which was merged into master in #2173

PR #2134, opened 2018-02-10
BF: Tolerate files starting with -

yarikoptic:
This pull request reveals the spread of #2131 and may be then will attempt at a few fixes.  "Sing along" - contributions will be open

This pull request proposes someone to
- [x] RF to provide `files` after `--` in a call to `git annex` commands
- (unlikely here) make `get_most_obscure_filename` into a generator to produce few of those lovely ones.  We cannot have just one pointing to all the issues, although I think spaces in the middle are nearly always as bad as in the beginning (trailing ones are trickier)
- [  x?  ] as a quicker resolution to above - make prefix to obscure filename controllable from env, so we could have a matrix run with `-` as a prefix. **I already have it but do not want to push until I make sure that all of the issues are caught first -- tests are still running although buildbots are already happy**

### Changes
- [x] This change is complete for whatever it wanted to show

### DIFF
--- a/datalad/interface/tests/test_add_archive_content.py
+++ b/datalad/interface/tests/test_add_archive_content.py
@@ -272,7 +272,7 @@ def test_add_archive_content(path_orig, url, repo_path):
     ok_archives_caches(repo.path, 0, persistent=False)
 
     repo.drop(opj('1', '1 f.txt'))  # should be all kosher
-    repo.drop(key_1tar, options=['--key'])  # is available from the URL -- should be kosher
+    repo.drop(key_1tar, key=False)  # is available from the URL -- should be kosher
     repo.get(opj('1', '1 f.txt'))  # that what managed to not work
 
     # TODO: check if persistent archive is there for the 1.tar.gz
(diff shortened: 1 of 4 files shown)


mih:
Seems the removed import was needed but some bits. They better import it from its real place.

yarikoptic:
I guess your meant to comment on  #2138. I will look into it... That removed import could get back but it is not used in that file, so likely just was leaked into another import, which could be fixed

datalad/interface/tests/test_add_archive_content.py
```diff
@@ -272,15 +272,15 @@ def d2_basic_checks():
     ok_archives_caches(repo.path, 0, persistent=False)
 
     repo.drop(opj('1', '1 f.txt'))  # should be all kosher
-    repo.drop(key_1tar, options=['--key'])  # is available from the URL -- should be kosher
+    repo.drop(key_1tar, key=False)  # is available from the URL -- should be kosher
```
yarikoptic: reviewing your own code is always so useful -- not sure how it worked out here ;-)

mih:
Will merge. Can we have this in master too?

yarikoptic:
eh, need to open another PR then first where I will enable back leading space obscure name by default -- I think we need it. then will merge that and then will sure thing will push it for master

PR #2036, opened 2018-02-12
Some kind of progress bar for index construction?

mih:
#### What is the problem?
Search index construction might take a while... E.g. I have started one yesterday evening (5pm) and it is still going (after 18h, edit 1: I am pretty sure it is stuck -- it is 28th hour now) on a reasonably fast machine (smaug ;) ).  I have no clue either it is not just stuck, or when it should be complete, all I see is
```shell
(dev3) 2 11290.....................................:Fri 15 Dec 2017 05:01:52 PM EST:.
(git)smaug:/tmp/crawl[master]
$> datalad search Haxby type:dataset
[INFO   ] Building search index
[INFO   ] Scanning for metadata keys
```
I wondered, if it goes sequentially through datasets and datasets list is initially constructed (full or just top level) -- would be nice to have some kind of a progress bar with % or at least some indication of activity/progress.

### DIFF
--- a/datalad/metadata/metadata.py
+++ b/datalad/metadata/metadata.py
@@ -43,7 +43,6 @@ import datalad.support.ansi_colors as ac
 from datalad.support.json_py import load as jsonload
 from datalad.support.json_py import load_xzstream
 from datalad.interface.common_opts import recursion_flag
-from datalad.interface.common_opts import recursion_limit
 from datalad.interface.common_opts import reporton_opt
 from datalad.distribution.dataset import Dataset
 from datalad.distribution.dataset import EnsureDataset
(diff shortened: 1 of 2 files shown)


mih:
@yarikoptic Attached diff does not add a progress bar, but it places the necessary (?) info in the right place. I have not yet implemented a progress bar, so it would take me some time that I don't have for this ATM. But even this minimal change should mitigate the issue to some degree -- with little runtime cost.

```
% datalad search --reindex
[INFO   ] Rebuilding search index 
[INFO   ] Processing metadata records for 5 datasets 
[INFO   ] Added 1 document on dataset . 
[INFO   ] Added 1 document on dataset crcns-vim-2 
[INFO   ] Added 1 document on dataset openfmri 
[INFO   ] Added 1 document on dataset openfmri/ds000001 
[INFO   ] Added 1 document on dataset openfmri/ds000002 
[INFO   ] Search index contains 5 documents 
```

mih:
I will merge this now, because I need the new bits in metadata.

PR #2149, opened 2018-02-12
NF: Plugin `extract_metadata`

mih:
Allow for quick invocation of any of datalad's extractors. (DOCs inside)

Thanks to @yarikoptic for the activation energy to write this.

- [x] tests

### DIFF
--- a/datalad/metadata/metadata.py
+++ b/datalad/metadata/metadata.py
@@ -376,7 +376,6 @@ def _get_metadata(ds, types, global_meta=None, content_meta=None, paths=None):
     ----------
     ds : Dataset
     types : list
-    mode : {'init', 'add', 'reset'}
     """
     errored = False
     dsmeta = MetadataDict()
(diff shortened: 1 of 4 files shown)


yarikoptic:
FWIW
- after merging in #2151 worked on a tricky (with unicode in participants) ds000201
- needing to specify a file as a keyworkd argument (`file=`) in comparison to just a path is really inconvenient since shell would not complete or even "fix" it by removing the `file=` prefix.  can't plugins get positional argument(s)?

mih:
@yarikoptic Thanks for trying!

re positional args: The tricky part is to find a strategy to distinguish multiple positionals from one positional that shall receive multiple values. The present implementation was sufficient so far. If we can find a better way I am all for it.

mih:
Whatever works. For me this has a very low priority. I do not have a real use case for any of that. I also think that development should go into the direction of unifying plugins and command, and not towards yet another markup language used for a minor bit of functionality in datalad.

PR #2151, opened 2018-02-12
Various bugfixes addressing some corner cases while parsing datasets.datalad.org with new extractors

### LINKED ISSUE #2147: open_r_encdetect -- 1kB might not be enough!
#### What is the problem?
`open_r_encdetect` helper promises to guess the encoding of the file by first reading first 1kB and guessing the type from it, and then reading using `io.open` while providing encoding

IMHO it is principally incorrect -- unicode character might happen to come anywhere in the text file (happens for 53) even in the last position in the text.  

overall `io` module afaik is more of a plumbing than user-oriented.  I do not see why we should not just read the entire file in bytes, and then guess encoding on the entirety of the content.  I thought even to some degree that is what stock `open` is intended to do and does in this particular use case just fine:
```
*(Pdb) print type(open(README_fname).read())
<type 'unicode'>
```
so what problem this helper was intended to solve?

yarikoptic:
This pull request fixes #2147

It seems that we need a bit more robust operation while reading those files which might contain unicode.

Ref: https://github.com/datalad/datasets/issues/17

### Changes
- [x] This change is complete.

Please have a look @datalad/developers even if it is still WiP

### DIFF
--- a/datalad/metadata/aggregate.py
+++ b/datalad/metadata/aggregate.py
@@ -400,16 +400,31 @@ def _update_ds_agginfo(refds_path, ds_path, subds_paths, agginfo_db, to_save):
                      for ai in ds_agginfos.values()
                      for k in location_keys
                      if k in ai)
-    objs2remove = objlocs_was.difference(objlocs_is)
     # TODO do we need to (double?) check if all object files exist?
-    #objs2add = [o for o in objlocs_is if exists(opj(ds_path, o))]
+    # objs2add = [o for o in objlocs_is if exists(opj(ds_path, o))]
     objs2add = objlocs_is
+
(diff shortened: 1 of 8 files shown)


yarikoptic:
eh -- magic didn't fully work.  On travis seems to fail to load unicode participants, on buildbots -- decodes sh.t left and right even when it shouldn't. heh

mih:
LGTM. Let's wait for the tests.

yarikoptic:
There was something funny with the network late yesterday evening. So I have pushed an irrelevant but a good commit (pylintification of the file I had opened already ;))

mih:
Not sure why nd90 would produce a commit difference in the openfmri pipeline for this PR.

Issue #2156, opened 2018-02-13
Make generation of unique metadata properties optional


mih:
#### What is the problem?

Some metadata extractors already generate much nicer dataset-global metadata (e.g. `dicom` generates metadatasets per image series). Those tend to be more useful. We should have a per-extractor config setting that allows dataset authors to turn the auto-generation off.

Issue #2146, opened 2018-02-12
Add a list of extractors to the "Aggregate metadata for dataset ..." message


yarikoptic:
#### What is the problem?
no problem but it would be very handy -- this way I could possibly catch datasets without any/inappropriate extractors configured etc

Something like
```
Aggregate metadata for dataset /mnt/btrfs/datasets-meta6-1-redo1/datalad/crawl/openfmri/ds000053: bids, nifti1
```
should suffice, or they could even go before the path since path is so long and not rel to curdir

mih:
#2157 will bring this:

```
% datalad aggregate-metadata -r
[INFO   ] Aggregate metadata for dataset /tmp/lit 
[INFO   ] Engage metadata extractors: ['datalad_core', 'xmp'] 
[INFO   ] Update aggregate metadata in dataset at: /tmp/lit 
aggregate_metadata(ok): /tmp/lit (dataset)                                                                                   
```
Doing it in one log message would require substantial RF to make info available in one place -- seems to have no benefit.

yarikoptic:
are engaged metadata extractors yielded back somehow in that record (if I change output to json*)?

mih:
They do not necessarily yield something. You wanted a list of things that run, to be able to spot if something is missing. If you want to see what was produced you can look at the keys in the metadata structure. There will be one for each extractor that produced something, and where the result did not get filtered as a consequence of one of your other settings.

yarikoptic:
I meant aggregate-metadata command, I assumed it is a generator as everything in your flashy new API

mih:
Ah, in the result record of `aggregate-metadata`. No, there are not in there. Might indeed be a useful bit of info. I will have to look into this.

mih:
What about this?

```
% datalad --dbg -f json_pp aggregate-metadata
[INFO   ] Aggregate metadata for dataset /tmp/lit 
[INFO   ] Engage metadata extractors: ['datalad_core', 'xmp'] 
[INFO   ] Update aggregate metadata in dataset at: /tmp/lit 
{                                                                                                       
  "action": "aggregate_metadata", 
  "content_info": "/tmp/lit/.datalad/metadata/objects/0a/cn-0a893378f6e3a8bbe7ab30a0a10c81.xz", 
  "dataset_info": "/tmp/lit/.datalad/metadata/objects/0a/ds-0a893378f6e3a8bbe7ab30a0a10c81", 
  "extractors": [
    "datalad_core", 
    "xmp"
[...]

PR #2157, opened 2018-02-13
Various small improvements to metadata aggregation

mih:
- [x] `metadata --get-aggregates` reports metadata layout version
- [x] Make generation of unique metadata properties optional #2156
- [x] Typo in variable definition
- [x] log list of active extractors #2146

### DIFF
--- a/datalad/metadata/aggregate.py
+++ b/datalad/metadata/aggregate.py
@@ -26,6 +26,7 @@ from os.path import curdir
 from hashlib import md5
 import shutil
 
+import datalad
 from datalad.interface.annotate_paths import AnnotatePaths
 from datalad.interface.base import Interface
 from datalad.interface.utils import eval_results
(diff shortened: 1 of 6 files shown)


datalad/metadata/aggregate.py
```diff
@@ -698,12 +705,14 @@ def __call__(
...
                 to_save)
             # update complete
-            yield get_status_dict(
+            res = get_status_dict(
                 status='ok',
                 action='aggregate_metadata',
                 path=parentds_path,
                 type='dataset',
                 logger=lgr)
+            res.update(agginfo_db.get(parentds_path, {}))
```
yarikoptic: FWIW we have `datalad.utils.updated` so could have been
```python
yield updated(
    get_status_dict(
￼                  status='ok',
￼                  action='aggregate_metadata',
￼                  path=parentds_path,
￼                  type='dataset',
￼                  logger=lgr),
    agginfo_db.get(parentds_path, {})
)
```
not sure if makes it anyhow better here (would be more difficult to debug if smth happens) but here you go

yarikoptic:
Finally tests completed

PR #2172, opened 2018-02-14
Bf spaced url failure

### LINKED ISSUE #2097: DataLad crawl fails on a url with a space
#### What is the problem? 

@yarikoptic gave me an exercise to crawl data given a url link to a website and crawler fails. 

``` shell
(ENV) Taylors-MacBook-Pro:demo taylor$ datalad crawl-init --save --template=simple_with_archives 'url=http://myweb.fsu.edu/bgomez/research.html' 'a_href_match_=.*\.zip$'
[INFO   ] Creating a pipeline to crawl data files from http://myweb.fsu.edu/bgomez/research.html 
[INFO   ] Not adding annex.largefiles=exclude=README* and exclude=LICENSE* to git annex calls because already defined to be (not(mimetype=text/*)) 
(ENV) Taylors-MacBook-Pro:demo taylor$ datalad crawl                                                                                    
[INFO   ] Loading pipeline specification from ./.datalad/crawl/crawl.cfg 
[INFO   ] Creating a pipeline to crawl data files from http://myweb.fsu.edu/bgomez/research.html 
[INFO   ] Not adding annex.largefiles=exclude=README* and exclude=LICENSE* to git annex calls because already defined to be (not(mimetype=text/*)) 
[INFO   ] Running pipeline [<function Annexificator.switch_branch.<locals>.switch_branch at 0x10c891b70>, [[<datalad.crawler.nodes.crawl_url.crawl_url object at 0x10c8889b0>, a_href_match(query='.*\\.zip$'), <datalad.crawler.nodes.annex.Annexificator object at 0x10a917438>]], <function Annexificator.switch_branch.<locals>.switch_branch at 0x10c891bf8>, [<function Annexificator.merge_branch.<locals>.merge_branch at 0x10c891c80>, [find_files(dirs=False, fail_if_none=True, regex='\\.(zip|tgz|tar(\\..+)?)$', topdir='.'), <function Annexificator.add_archive_content.<locals>._add_archive_content at 0x10c891d08>]], <function Annexificator.switch_branch.<locals>.switch_branch at 0x10c891d90>, <function Annexificator.merge_branch.<locals>.merge_branch at 0x10c891e18>, <function Annexificator.finalize.<locals>._finalize at 0x10c891ea0>] 
[INFO   ] Found branch non-dirty -- nothing was committed 
[INFO   ] Checking out an existing branch incoming 
[INFO   ] Fetching 'http://myweb.fsu.edu/bgomez/research.html' 
[INFO   ] Need to download 1.8 MB from http://myweb.fsu.edu/bgomez/HansfordGomez_APSR_2010.zip. No progress indication will be reported 
[INFO   ] Need to download 5.1 MB from http://myweb.fsu.edu/bgomez/Weather_Public_File.zip. No progress indication will be reported 
[INFO   ] Need to download 127.3 kB from http://myweb.fsu.edu/bgomez/GomezWilson_2006_JOP_Replication Material.zip. No progress indication will be reported 
[WARNING] Caught AnnexBatchCommandError: command 'addurl'
Error, annex reported failure for addurl: {'command': 'addurl', 'success': False, 'file': None} [annexrepo.py:add_url_to_file:2042] on trial #1. Sleeping 3.000000 and retrying 
[WARNING] Caught AnnexBatchCommandError: command 'addurl'
Error, annex reported failure for addurl: {'command': 'addurl', 'success': False, 'file': None} [annexrepo.py:add_url_to_file:2042] on trial #2. Sleeping 9.000000 and retrying 
[WARNING] Caught AnnexBatchCommandError: command 'addurl'
Error, annex reported failure for addurl: {'command': 'addurl', 'success': False, 'file': None} [annexrepo.py:add_url_to_file:2042] on trial #3. Sleeping 27.000000 and retrying 
[WARNING] Caught AnnexBatchCommandError: command 'addurl'
Error, annex reported failure for addurl: {'command': 'addurl', 'success': False, 'file': None} [annexrepo.py:add_url_to_file:2042] on trial #4. Sleeping 81.000000 and retrying 
[WARNING] Caught AnnexBatchCommandError: command 'addurl'
Error, annex reported failure for addurl: {'command': 'addurl', 'success': False, 'file': None} [annexrepo.py:add_url_to_file:2042] on trial #5. Sleeping 243.000000 and retrying 
```
Version 0.9.1

overlake333:
This pull request fixes #2097 

This pull request proposes replace spaces in url with correct syntax

### Changes
- [ ] This change is complete

### DIFF
--- a/datalad/crawler/nodes/misc.py
+++ b/datalad/crawler/nodes/misc.py
@@ -27,6 +27,7 @@ from ..pipeline import PIPELINE_TYPES
 from ...utils import auto_repr
 from ...utils import find_files as _find_files
 from six.moves.urllib.parse import quote
+from datalad.support.network import URL
 
 from logging import getLogger
 from nose.tools import eq_, assert_raises
@@ -598,5 +599,7 @@ def fix_url(data):
     """
 
     for key, value in data.items():
-         data[key] = quote(data[key], safe = ':/')
+        while data[key][0] == ' ':
+            data[key] = data[key][1:len(data[key])]
+        data[key] = URL(data[key])
     yield data
--- a/datalad/crawler/nodes/misc.py
+++ b/datalad/crawler/nodes/misc.py
@@ -26,6 +26,7 @@ from ..pipeline import xrun_pipeline
 from ..pipeline import PIPELINE_TYPES
 from ...utils import auto_repr
 from ...utils import find_files as _find_files
+from six.moves.urllib.parse import quote
 
 from logging import getLogger
 from nose.tools import eq_, assert_raises
@@ -597,5 +598,5 @@ def fix_url(data):
     """
 
     for key, value in data.items():
-        data[key] = data[key].replace(" ", "%20")
+         data[key] = quote(data[key], safe = ':/')
     yield data
--- a/datalad/crawler/nodes/misc.py
+++ b/datalad/crawler/nodes/misc.py
@@ -595,8 +595,7 @@ def fix_url(data):
 
     - replace spaces with %20
     """
-    url = data['url']
-    print(url)
-    if ' ' in url:
-        data['url'].replace(" ", "%20")
+
+    for key, value in data.items():
+        data[key] = data[key].replace(" ", "%20")
     yield data
--- a/datalad/crawler/nodes/misc.py
+++ b/datalad/crawler/nodes/misc.py
@@ -26,7 +26,6 @@ from ..pipeline import xrun_pipeline
 from ..pipeline import PIPELINE_TYPES
 from ...utils import auto_repr
 from ...utils import find_files as _find_files
-from six.moves.urllib.parse import quote
 from datalad.support.network import URL
 
 from logging import getLogger
@@ -592,14 +591,13 @@ class debug(object):
             pdb.set_trace()
 
 
-def fix_url(data):
+def fix_url(data, keys=['url']):
     """Given data, get value within 'url' key and fix up so it is legit url
 
     - replace spaces with %20
     """
-
-    for key, value in data.items():
-        while data[key][0] == ' ':
-            data[key] = data[key][1:len(data[key])]
-        data[key] = URL(data[key])
+    data = data.copy()
+    for key in keys:
+        if key in data: # catches key error if dictionary does not contain key
+            data[key] = URL(data[key]).as_str()
     yield data
--- a/datalad/crawler/nodes/misc.py
+++ b/datalad/crawler/nodes/misc.py
@@ -589,3 +589,14 @@ class debug(object):
             lgr.info("Ran node %s which yielded %d times", node, n)
             pdb.set_trace()
 
+
+def fix_url(data):
+    """Given data, get value within 'url' key and fix up so it is legit url
+
+    - replace spaces with %20
+    """
+    url = data['url']
+    print(url)
+    if ' ' in url:
+        data['url'].replace(" ", "%20")
+    yield data
--- a/datalad/crawler/nodes/tests/test_misc.py
+++ b/datalad/crawler/nodes/tests/test_misc.py
@@ -50,7 +50,6 @@ import logging
 from mock import patch
 from nose.tools import eq_, assert_raises
 from nose import SkipTest
-import six.moves.builtins as __builtin__
 
 
 # TODO: redo on a local example
@@ -477,8 +476,10 @@ def test_debug():
 
 
 def test_fix_url():
-    test_url = 'http://myweb.fsu.edu/bgomez/AJPS_2001_Erratum.zip'
-    # urlTest = 'http://myweb.fsu.edu/bgomez/GomezWilson_2006_JOP_Replication Material.zip'
-    eq_(list(fix_url({'url': test_url, 'url': test_url + ' ', 'url': test_url + '  ',
-                      'url': ' ' + test_url, 'url': '  ' + test_url})),
-        [{'url': test_url, 'url': test_url, 'url': test_url, 'url': test_url, 'url': test_url}])
+    eq_(list(fix_url({'url': "http://site/u r"})), [{'url': "http://site/u%20r"}])
+    eq_(list(fix_url({'url': "http://site/ur "})), [{'url': "http://site/ur%20"}]) # trailing spaces are still spaces
+    eq_(list(fix_url({'html': "http://site/u r", 'url':"http://site/ur "})),
+        [{'html': "http://site/u r", 'url': "http://site/ur%20"}])
+    eq_(list(fix_url({'html': "http://site/u r", 'url': "http://site/ur "}, keys=['url', 'html'])),
+        [{'html': "http://site/u%20r", 'url': "http://site/ur%20"}])
+    eq_(list(fix_url({'html': "http://site/u r"})), [{'html': "http://site/u r"}]) # no urls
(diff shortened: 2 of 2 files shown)


datalad/crawler/nodes/misc.py
```diff
@@ -589,3 +589,13 @@ def __call__(self, data):
             lgr.info("Ran node %s which yielded %d times", node, n)
             pdb.set_trace()
 
+
+def fix_url(data):
+    """Given data, get value within 'url' key and fix up so it is legit url
+
+    - replace spaces with %20
```
bpoldrack: May I propose to just use `urllib.quote()`?

bpoldrack: Additional hint: Since we want to be compatible with PY2 as well as PY3 you'd need to use
```python
from six.moves.urllib.parse import quote
```

datalad/crawler/nodes/tests/test_misc.py
```diff
@@ -471,4 +473,11 @@ def node(d):
...
     yield _test_debug, "About to run"
     yield _test_debug, "Ran node .* which yielded 1 times", ('after',)
-    yield _test_debug, "Ran node .* which yielded 0 times", ('empty',)
\ No newline at end of file
+    yield _test_debug, "Ran node .* which yielded 0 times", ('empty',)
+
+
+def test_fix_url():
+    # TODO: actual test
+    print("Test")
```
bpoldrack: In case that `print` was meant to be a marker for you to spot your test in nose's output, you can run that test only via `nosetests -s -v datalad/crawler/nodes/tests/test_misc.py:test_fix_url`.
If the purpose was different - just ignore me. ;-)

datalad/crawler/nodes/misc.py
yarikoptic: unfortunately you can't really... even if we do split into parts, to fix the "path", URL might already contain legit quoted symbols, and double quoting them would ruin them... this one is really just to fix up wrong urls like some ppl do mistakes

datalad/crawler/nodes/misc.py
```diff
@@ -589,3 +590,13 @@ def __call__(self, data):
...
+    """
+
+    for key, value in data.items():
+         data[key] = quote(data[key], safe = ':/')
```
yarikoptic: - if I am correct in above, quote wouldn't work for us, but I could be wrong and @bpoldrack might provide better detail
- we do not want to "fix" all the values (note that you haven't even used `value` here) -- just the url field.  You could add an optional keyword argument to list the keys to check/fix (default to `['url']`)

datalad/crawler/nodes/tests/test_misc.py
```diff
@@ -471,4 +473,9 @@ def node(d):
...
     yield _test_debug, "About to run"
     yield _test_debug, "Ran node .* which yielded 1 times", ('after',)
-    yield _test_debug, "Ran node .* which yielded 0 times", ('empty',)
\ No newline at end of file
+    yield _test_debug, "Ran node .* which yielded 0 times", ('empty',)
+
+
+def test_fix_url():
+    eq_(list(fix_url({'url': 'http://myweb.fsu.edu/bgomez/GomezWilson_2006_JOP_Replication Material.zip'})),
+        [{'url': 'http://myweb.fsu.edu/bgomez/GomezWilson_2006_JOP_Replication%20Material.zip'}])
```
yarikoptic: come up with few more examples for some "corner" cases even though if you know that they must work correctly. e.g. what about the space at the end of url?

datalad/crawler/nodes/misc.py
bpoldrack: > URL might already contain legit quoted symbols, and double quoting them would ruin them

Ah - my fault. Wasn't aware that would be possible at this point. `urllib.quote` isn't that smart indeed. However, why not fixing more possible mistakes than just spaces here? To guard against double quoting, we could `quote(unquote(url))`.

yarikoptic: `quote(unquote())` indeed might work, except for extreme abusers (i.e. having really meant `%20` to be `%20` ;)) but since we have no such usecases, probably would be the best way.  Care should be taken though about first splitting url into components and only fixing up the `path` since:
```ipython
In [19]: quote(unquote('http://example.com/wi$ th?blah=bu#123'), safe=':/')
Out[19]: 'http://example.com/wi%24%20th%3Fblah%3Dbu%23123'
```
so we should use our `URL` class, which is apparently does the full dance internally already so no manual quote call is needed
```ipython
In [42]: from datalad.support.network import URL

In [43]: url = 'http://example.com/wi$ th?blah=bu#123'

In [44]: url_obj = URL(url)

In [45]: url_obj.as_str()
Out[45]: 'http://example.com/wi%24%20th?blah=bu#123'
```

@overlake333 please fix up according to the above recipe and add tests for those more interesting cases ;)

datalad/crawler/nodes/misc.py
```diff
@@ -589,3 +591,15 @@ def __call__(self, data):
...
             pdb.set_trace()
 
+
+def fix_url(data):
+    """Given data, get value within 'url' key and fix up so it is legit url
+
+    - replace spaces with %20
+    """
+
+    for key, value in data.items():
```
yarikoptic: I left a comment before -- please review (it should have come in email as well):  we do not want to process all `data` values, only `"url"` (for now).  To make it more generic, add a keyword argument to this function (`keys=["url"]) and add tests where you change it so "it does the right thing"

datalad/crawler/nodes/misc.py
```diff
@@ -589,3 +591,15 @@ def __call__(self, data):
...
+        while data[key][0] == ' ':
+            data[key] = data[key][1:len(data[key])]
```
yarikoptic: why is this necessary?

datalad/crawler/nodes/tests/test_misc.py
```diff
@@ -471,4 +473,12 @@ def node(d):
...
\ No newline at end of file
+    yield _test_debug, "Ran node .* which yielded 0 times", ('empty',)
+
+
+def test_fix_url():
+    test_url = 'http://myweb.fsu.edu/bgomez/AJPS_2001_Erratum.zip'
+    # urlTest = 'http://myweb.fsu.edu/bgomez/GomezWilson_2006_JOP_Replication Material.zip'
+    eq_(list(fix_url({'url': test_url, 'url': test_url + ' ', 'url': test_url + '  ',
+                      'url': ' ' + test_url, 'url': '  ' + test_url})),
+        [{'url': test_url, 'url': test_url, 'url': test_url, 'url': test_url, 'url': test_url}])
```
yarikoptic: this is a dictionary, so 
```{'url': test_url, 'url': test_url, 'url': test_url, 'url': test_url, 'url': test_url}``` 
is identical to
```{'url': test_url}```
;)  make it into separate tests, e.g.

```python
eq_(list(fix_url({'url': "http://site/u r"})) == [{'url': "http://site/u%20r"}])
eq_(list(fix_url({'url': "http://site/ur "})) == [{'url': "http://site/ur%20"}])  # trailing spaces are still spaces
```

datalad/crawler/nodes/tests/test_misc.py
```diff
@@ -49,6 +50,7 @@
 from mock import patch
 from nose.tools import eq_, assert_raises
 from nose import SkipTest
+import six.moves.builtins as __builtin__
```
yarikoptic: why is it needed here?

PR #2174, opened 2018-02-15
RF: do not write down index every time in __del__

yarikoptic:
- causes all the ugly messages in read-only access
- not sure if needed any longer - hopefully all needed flushing was
  fixed in GitPython long ago

I want to see if any tests fail for us now after this change

### Changes
- [x] This change is complete

### DIFF
--- a/datalad/distribution/tests/test_clone.py
+++ b/datalad/distribution/tests/test_clone.py
@@ -215,6 +215,7 @@ def test_clone_isnot_recursive(src, path_nr, path_r):
         {'subm 1', '2'})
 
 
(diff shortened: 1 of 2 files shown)


bpoldrack:
I think we don't use GitPython's in-memory index to write anything anymore. So we shouldn't need it indeed.

But what's going on the PY2 V6 build? Looks like the entire setup is wrong and it doesn't find our special remote (the excutable).

yarikoptic:
ok, lets hope for survival (one failing run is unrelated -- jsonld, gone in master)

PR #2176, opened 2018-02-15
BF: None recursion-limits matter!!!

### LINKED ISSUE #2171: diff --recursive seems to limit to 1 level by default!
#### What is the problem?

I assumed that `--recursive` should by default have no limit, that is how other commands function AFAIK.  Seems to be not the case with `diff` and nothing in its docstring about that if that is intended

```shell
(git)smaug:/mnt/btrfs/datasets/datalad/crawl/dicoms/dartmouth-phantoms[master]git
$> datalad diff --recursive                   
   modified(dataset): bids_test4-20161014
   modified(dataset): bids_test4-20161014/phantom-1
   modified(dataset): bids_test4-20161014/phantom-2
   modified(dataset): bids_test4-20161014/phantom-3
   modified(dataset): bids_test4-20161014/phantom-4
   modified(dataset): bids_test4-20161014/phantom-5
   modified(dataset): bids_test4-20161014/phantom-6
   modified(dataset): bids_test4-20161014/phantom-7
   modified(dataset): bids_test4-20161014/phantom-8
   modified(dataset): bids_test4-20161014/phantom-9

$> datalad diff --recursive --recursion-limit=3
   modified(dataset): bids_test4-20161014
   modified(dataset): bids_test4-20161014/phantom-1
   modified(dataset): bids_test4-20161014/phantom-2
   modified(dataset): bids_test4-20161014/phantom-3
   modified(dataset): bids_test4-20161014/phantom-4
   modified(dataset): bids_test4-20161014/phantom-5
   modified(dataset): bids_test4-20161014/phantom-6
   modified(dataset): bids_test4-20161014/phantom-7
   modified(dataset): bids_test4-20161014/phantom-8
   modified(dataset): bids_test4-20161014/phantom-9
      modified(file): bids_test4-20161014/phantom-1/.datalad/config
      modified(file): bids_test4-20161014/phantom-2/.datalad/config
      modified(file): bids_test4-20161014/phantom-3/.datalad/config
      modified(file): bids_test4-20161014/phantom-4/.datalad/config
      modified(file): bids_test4-20161014/phantom-5/.datalad/config
      modified(file): bids_test4-20161014/phantom-6/.datalad/config
      modified(file): bids_test4-20161014/phantom-7/.datalad/config
      modified(file): bids_test4-20161014/phantom-8/.datalad/config
      modified(file): bids_test4-20161014/phantom-9/.datalad/config
```
```shell
$> datalad diff --help | grep -3 recursi  
Usage: datalad diff [-h] [-d DATASET] [--revision [REVISION EXPRESSION]]
                    [--staged] [--ignore-subdatasets IGNORE_SUBDATASETS]
                    [--report-untracked REPORT_UNTRACKED] [-r]
                    [--recursion-limit LEVELS]
                    [PATH [PATH ...]]

Report changes of dataset components.
--
                        work tree of subdatasets, only changes to the
                        revisions stored in the parent dataset are shown.
                        Using "all" hides all changes to subdatasets. Note,
                        even with "all" recursive execution will still report
                        other changes in any existing subdataset, only the
                        subdataset record in a parent dataset is not
                        evaluated. Constraints: value must be one of ('none',
--
                        in fully untracked directories. Constraints: value
                        must be one of ('no', 'normal', 'all') [Default:
                        'normal']
  -r, --recursive       if set, recurse into potential subdataset. [Default:
                        False]
  --recursion-limit LEVELS
                        limit recursion into subdataset to the given number of
                        levels. Constraints: value must be convertible to type
                        'int' [Default: None]
```
bleeding edge master 0.9.1-783-g4f58c02c

yarikoptic:
This pull request fixes #2171 
So it was/is(may be there is other spots) for special treatment of recursion-limit being None.  I did not analyze code well enough to see how I could reproduce or why it didn't reproduce -- someone with a better clue should come up with a test so the issue does not reappear -- any takers?

### Changes
- [x] This change is complete
- [ ] Add test which would fail before and pass now

### DIFF
--- a/datalad/distribution/subdatasets.py
+++ b/datalad/distribution/subdatasets.py
@@ -268,7 +268,7 @@ class Subdatasets(Interface):
                         k)
         try:
             if not (bottomup or contains or set_property or delete_property or \
-                    (recursive and recursion_limit is not None)):
+                    recursive):
                 # FAST IMPLEMENTATION FOR THE STRAIGHTFORWARD CASE
                 # as fast as possible (just a single call to Git)
                 # need to track current parent


datalad/distribution/subdatasets.py
```diff
@@ -268,7 +268,7 @@ def __call__(
                         k)
         try:
             if not (bottomup or contains or set_property or delete_property or \
-                    (recursive and recursion_limit is not None)):
+                    recursive):
                 # FAST IMPLEMENTATION FOR THE STRAIGHTFORWARD CASE
```
mih: This will have undesirable sideeffects. With this change every subdataset will get a separate call to `git diff`, instead of a single call to `git diff` for the entire tree. This is a substantial slowdown, as `git diff` will now evaluate the state of each submodule twice. Seperate calls are only necessary to support behavior that `git diff` cannot handle directly, such as recursion limits, reverse order, and path matches.

yarikoptic: this logic is far from my comprehension so unlikely I would be able to deliver the "correct" fix here.

yarikoptic:
closing since we figured it was wrong

PR #2177, opened 2018-02-15
ENH: `search`

mih:
- [x]  has modes encapsulated in different classes now. This gives a standard API for alternative searchers to implement.
- [x] "egrep" search mode with expression matching on a stringified metadata dump, slow but works without an intermediate metadata transformation into a search index
- [x] default search mode is configurable (by dataset if necessary) and defaults to 'egrep'


mih:
I will merge this state for now in order to get all the bug fixes and this one together.

PR #2178, opened 2018-02-15
Add HOWTO on Michael's way "issue -> PR" into CONTRIBUTING.md

yarikoptic:
#### What is the problem?
There is magical ways @mih knows, but I do not remember.  Please add a sentence or two to CONTRIBUTING.md on that

1. Become @yarikoptic
2. just let time pass by

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -179,16 +179,38 @@ provided by [Sphinx].
 Additional Hints
 ----------------
 
-1. For merge commits to have more informative description, add to your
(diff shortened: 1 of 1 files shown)


yarikoptic:
it did it! even figured out the "correct" branch ;-)

mih:
@kyleam Fair point. Additionally, the cost incurred for anything more than a one-issue-one-quick-fix makes it less attractive than it may seem. I'll try to stop doing it.

yarikoptic:
actually I will just remove that section altogether and push directly to master, no ci

PR #2188, opened 2018-02-16
BF of #2182

### LINKED ISSUE #2182: `aggregate-metadata` cannot aggregate metadata from absent datasets
#### What is the problem?

If a subdataset carries metadata on an absent subsubdataset, `aggregate-metadata` should be able to pick it up. And it is actually designed to do so.

However, the helper function `discover_dataset_trace_to_targets()` ignores absent datasets, hence this isn't actually happening.

I confirmed that the input to that function is proper.

Specifically, that helper relies on the condition of `isdir(basepath)` to be valid for any candidate dataset.

mih:
- [x] fixed #2182
- [x] if pre-aggregated metadata in the closest available subdataset is annex'ed, it may need to be obtained, before it can be included in a superdataset

### DIFF
--- a/datalad/metadata/aggregate.py
+++ b/datalad/metadata/aggregate.py
@@ -443,6 +443,8 @@ def _update_ds_agginfo(refds_path, ds_path, subds_paths, agginfo_db, to_save):
             to_save(dict(path=ds_path, type='dataset', staged=True))
 
     # must copy object files to local target destination
+    # make sure those objects are present
+    ds.get([f for f, t in objs2copy], result_renderer='disabled')
     for copy_from, copy_to in objs2copy:
         if copy_to == copy_from:
             continue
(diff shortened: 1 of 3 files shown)


mih:
nd90 failure is the frequent misbehavior of the openfmri pipeline test on this machine.

datalad/metadata/aggregate.py
```diff
@@ -443,6 +443,8 @@ def _update_ds_agginfo(refds_path, ds_path, subds_paths, agginfo_db, to_save):
             to_save(dict(path=ds_path, type='dataset', staged=True))
 
     # must copy object files to local target destination
+    # make sure those objects are present
+    ds.get([f for f, t in objs2copy], result_renderer='disabled')
```
yarikoptic: was that a fancy way to say `list(objs2copy)`?

mih:
?

mih:
Ah, maybe you overlooked that these are to/from tuples. I only need the from part.

PR #2190, opened 2018-02-17
ENH(TST): a humble attempt to eliminate coverage diff

yarikoptic:
As codecov reports for all current PRs show we have some code which is covered in non-PR runs, but not covered in PRs, so by default we have negative bias in coverage for PRs.  I thought to mediate it but probably ended up actually making coverage worse since removed some unused and stale code which was covered

Also RFed import of the extractors into a helper function so we could have all of it code covered and possibly reused elsewhere later on

Let's see what happens ;-)

### Changes
- [x] This change is complete


yarikoptic:
thanks! it will not be complete until red boxes exist on codecov report ;)

yarikoptic:
something is flaky in openfmri test ;-/

yarikoptic:
woohoo -- all green (besides the buildbot - openfmri, unrelated), marking the checkmark

PR #2196, opened 2018-02-18
RF: use existing path_startswith instead of repeating pattern of .startswith(_with_sep

yarikoptic:
Following commands used
```shell
 git grep -l '\.startswith(_with_sep(' | xargs sed -i -e 's|\(\<[^ ]*\)\.startswith(_with_sep(\([^)]*\)))|path_startswith(\1, \2)|g'
git-sedi -e 's|\(\<[^ ]*\)\.startswith(_with_sep(\([^)]*\)))|path_startswith(\1, \2)|g' 
```
should be done on master as well after this merged -- there is more uses
### Changes
- [x] This change is complete

### DIFF
--- a/datalad/distribution/get.py
+++ b/datalad/distribution/get.py
@@ -53,6 +53,7 @@ from datalad.dochelpers import single_or_plural
 from datalad.utils import get_dataset_root
 from datalad.utils import with_pathsep as _with_sep
 from datalad.utils import unique
+from datalad.utils import path_startswith
 
 from .dataset import Dataset
 from .dataset import EnsureDataset
@@ -300,7 +301,7 @@ def _recursive_install_subds_underneath(ds, recursion_limit, reckless, start=Non
                 "subdataset %s is configured to be skipped on recursive installation",
                 sub['path'])
             continue
-        if start is not None and not subds.path.startswith(_with_sep(start)):
+        if start is not None and not path_startswith(subds.path, start):
             # this one we can ignore, not underneath the start path
             continue
         if sub['state'] != 'absent':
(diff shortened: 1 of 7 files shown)


mih:
Maybe it would be useful to change the name of that function to suggest that it does the "with_sep" part too? Maybe `isunderdirectory`?

yarikoptic:
`isunderdirectory` is hard to parse visually ... `is_under_dir` may be?

and there is already a more sophisticated  `path_is_under` https://github.com/datalad/datalad/blob/d031c24bf5a816e79c4da4bdd1006fed84492b8f/datalad/interface/utils.py#L134 
I don't think we would like to "merge" them into one, but their names should be consistent and worth moving into the same `utils`.  But note that order of arguments is opposite, and it is a tricky aspect here since I keep forgetting what it should be , so name should tell it (`arg1 functioname arg2`). May be `is_under_dir` and `is_under_dirs` while swapping in the 2nd one, and demanding first to be explicit `None` when used for curdir ?

mih:
Any of that sounds good to me. I was just surprised that we have such helper and still widespread non-use.

yarikoptic:
I think we could have reduced wide-spread if we used `from datalad import utils; utils.<whatever>` -- then at least I would have resorted to completion to see what methods we have there already.  Might be worth also looking at unifying names so it is easier to find the needed  function, e.g. if all path related functions started with `path_`. But then may be we should move them into `.support.path` and then import/use  that one (`path.join`, `path.is_under_dir`, `path_is_under_dirs`, etc)?

mih:
Good enough for now. As always, would be good to see such things in master, so they could get some usage...

PR #2202, opened 2018-02-19
BF(TST): do not reuse the same providers in the vcr-taped test

yarikoptic:
The problem (will probably appear in other cases as well, that vcr
tapes gets "bound" in the previous test to the providers we get to reuse
in the next test, which then screws it up when both tests are ran.
I am afraid there are other places where this would reveal itself and more
systematic solution should come

This pull request fixes those reports where this test_s3.py was failing mumbling about not being able to override a cassette

### Changes
- [x] This change is complete


yarikoptic:
heh, we do not test it on travis, so coverage is impaired

PR #2204, opened 2018-02-19
Merges and adjusts for use of path_startswith from 0.9.x

yarikoptic:
### Changes
- [x] This change is complete

### DIFF
--- a/datalad/interface/annotate_paths.py
+++ b/datalad/interface/annotate_paths.py
@@ -245,7 +245,7 @@ def get_modified_subpaths(aps, refds, revision, recursion_limit=None,
                 ap.update(m)
                 yield ap
                 break
-            if path_startswith(m['path'], ap['path']):
+            if path_is_subpath(m['path'], ap['path']):
                 # a modified path is underneath this AP
                 # yield the modified one instead
                 yield m
--- a/datalad/interface/annotate_paths.py
+++ b/datalad/interface/annotate_paths.py
@@ -44,6 +44,7 @@ from datalad.distribution.dataset import datasetmethod
 from datalad.utils import get_dataset_root
 from datalad.utils import with_pathsep as _with_sep
 from datalad.utils import path_startswith
+from datalad.utils import path_is_subpath
 from datalad.utils import assure_list
 
 from datalad.consts import PRE_INIT_COMMIT_SHA
@@ -155,7 +156,7 @@ def yield_recursive(ds, path, action, recursion_limit):
         # this check is not the same as subdatasets --contains=path
         # because we want all subdataset below a path, not just the
         # containing one
-        if path_startswith(subd_res['path'], path):
+        if path_is_subpath(subd_res['path'], path):
             # this subdatasets is underneath the search path
             # be careful to not overwrite anything, in case
             # this subdataset has been processed before
--- a/datalad/interface/diff.py
+++ b/datalad/interface/diff.py
@@ -116,7 +116,8 @@ def _get_untracked_content(dspath, report_untracked, paths=None):
             # strip state marker
             line[3:])
         norm_apath = normpath(apath)
-        if paths and not any([norm_apath == p or path_startswith(apath, p) for p in paths]):
+        if paths and not any(norm_apath == p or path_startswith(apath, p)
+                             for p in paths):
             # we got a whitelist for paths, don't report any other
             continue
         ap = dict(
--- a/datalad/interface/results.py
+++ b/datalad/interface/results.py
@@ -22,7 +22,7 @@ from os.path import abspath
 from os.path import normpath
 from datalad.utils import assure_list
 from datalad.utils import with_pathsep as _with_sep
-from datalad.utils import path_startswith
+from datalad.utils import path_is_subpath
 
 from datalad.distribution.dataset import Dataset
 
@@ -339,7 +339,7 @@ def results_from_annex_noinfo(ds, requested_paths, respath_by_status, dir_fail_m
             # do we have any failures in a subdir of the requested dir?
             failure_results = [
                 fp for fp in respath_by_status.get('failure', [])
-                if path_startswith(fp, p)]
+                if path_is_subpath(fp, p)]
             if failure_results:
                 # we were not able to process all requested_paths, let's label
                 # this 'impossible' to get a warning-type report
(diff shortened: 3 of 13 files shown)


yarikoptic:
fell into the trap of possibly comparing for it to be the same path... so created one more helper. names are still inconsistent but at least it should be clearer to figure out intended semantic!

yarikoptic:
@mih I will need your help on this one.  Because of not clear semantics in previous compositions for checks of paths, I did screw something up within metadata aggregation.  I have tried but can't figure it out.  Could you please have a look?

datalad/utils.py
```diff
@@ -1101,10 +1101,41 @@ def get_path_prefix(path, pwd=None):
         return path
 
 
+def _get_normalized_paths(path, prefix):
+    if isabs(path) != isabs(prefix):
+        raise ValueError("Bot paths must either be absolute or relative. "
+                         "Got %r and %r" % (path, prefix))
```
yarikoptic: I wonder if any test starts to fail with this... pointing to a potential problem in the code since we should have then resolved paths to full before comparing for commonprefix

datalad/interface/annotate_paths.py
```diff
@@ -243,7 +245,7 @@ def get_modified_subpaths(aps, refds, revision, recursion_limit=None,
                 ap.update(m)
                 yield ap
                 break
-            if m['path'].startswith(_with_sep(ap['path'])):
+            if path_startswith(m['path'], ap['path']):
```
bpoldrack: Not sure whether this actually is a bugfix, but here the outcome is different from before, if both paths are equal.

datalad/interface/diff.py
```diff
@@ -115,7 +116,8 @@ def _get_untracked_content(dspath, report_untracked, paths=None):
             # strip state marker
             line[3:])
         norm_apath = normpath(apath)
-        if paths and not any([norm_apath == p or apath.startswith(_with_sep(p)) for p in paths]):
+        if paths and not any(norm_apath == p or path_startswith(apath, p)
+                             for p in paths):
```
bpoldrack: Same here, I think.

datalad/interface/results.py
```diff
@@ -337,7 +339,7 @@ def results_from_annex_noinfo(ds, requested_paths, respath_by_status, dir_fail_m
             # do we have any failures in a subdir of the requested dir?
             failure_results = [
                 fp for fp in respath_by_status.get('failure', [])
-                if fp.startswith(_with_sep(p))]
+                if path_startswith(fp, p)]
```
bpoldrack: ... and here

datalad/interface/results.py
```diff
@@ -351,7 +353,7 @@ def results_from_annex_noinfo(ds, requested_paths, respath_by_status, dir_fail_m
                 # otherwise cool, but how cool?
                 success_results = [
                     fp for fp in respath_by_status.get('success', [])
-                    if fp.startswith(_with_sep(p))]
+                    if path_startswith(fp, p)]
```
bpoldrack: one more ...

datalad/interface/utils.py
```diff
@@ -222,7 +224,7 @@ def discover_dataset_trace_to_targets(basepath, targetpaths, current_trace,
             # leading dir basepath
             # filter targets matching this downward path
             downward_targets = set(
-                t for t in targetpaths if t == p or t.startswith(_with_sep(p)))
+                t for t in targetpaths if path_startswith(t, p))
```
bpoldrack: Hehe, almost fall for it. That one's correct. ;-)

datalad/metadata/metadata.py
```diff
@@ -124,7 +124,7 @@ def _load_xz_json_stream(fpath, cache=None):
 
 def _get_metadatarelevant_paths(ds, subds_relpaths):
     return (f for f in ds.repo.get_files()
-            if not any(f.startswith(_with_sep(ex)) or
+            if not any(path_startswith(f, ex) or
                        f == ex
```
bpoldrack: That `or f == ex` is not needed anymore.

datalad/metadata/metadata.py
```diff
@@ -152,7 +152,7 @@ def _get_containingds_from_agginfo(info, rpath):
         # containing subdataset (if there is any)
         containing_ds = sorted(
             [subds for subds in sorted(info)
-             if rpath.startswith(_with_sep(subds))],
+             if path_startswith(rpath, subds)],
```
bpoldrack: Again: Subpath.

datalad/metadata/metadata.py
```diff
@@ -234,7 +234,7 @@ def query_aggregated_metadata(reporton, ds, aps, recursive=False,
                 matching_subds = [(sub, sub) for sub in sorted(agginfos)
                                   # we already have the base dataset
                                   if (rpath == curdir and sub != curdir) or
-                                  sub.startswith(_with_sep(rpath))]
+                                  path_startswith(sub, rpath)]
```
bpoldrack: At least, formally here again. Although I'm not sure, whether equality is implicitly excluded anyway

datalad/metadata/metadata.py
```diff
@@ -303,7 +303,7 @@ def _query_aggregated_metadata_singlepath(
     for fpath in [f for f in contentmeta.keys()
                   if rparentpath == curdir or
                   f == rparentpath or
-                  f.startswith(_with_sep(rparentpath))]:
+                  path_startswith(f, rparentpath)]:
```
bpoldrack: Correct, since `f == rparentpath` was checked in addition. Therefore it can be removed.

datalad/interface/diff.py
bpoldrack: Actually not true. that normpath confused me. Here it is okay, since equality is tested anyway.
The question remains though, why to use normpath for equality testing, but not for startswith? That doesn't make sense to me.
Smells like a misunderstanding of what normpath does.

yarikoptic:
FTR: decreased coverage in crcns related code seems simply because datacite started to respond to our queries, which apparently uses some internal API, so failed access seemed to no longer happen...

datalad/utils.py
yarikoptic: ok, according to he coverage -- not covered which is good, i.e. was not hit, but we need coverage for it, so will do now

PR #2193, opened 2018-02-20
datalad_core _get_content_metadata - shouldn't it provide paths for all (annexed and not) files?

mih:
#### What is the problem?
initially ran into within https://github.com/datalad/datalad/pull/2190
https://travis-ci.org/datalad/datalad/jobs/342624112#L3010

apparently a basic test starts to fail in direct mode that for some reason in that repository `git annex metadata` does not return empty entries  for the annexed files queried.
But looking at the code I have also realized that because git annex metadata doesn't return any metadata for non-annexed files, no entries will be returned for them either.  But whenever repo is just a pure GitRepo we do return empty entries for each file.
So there is this discrepancy in behavior it seems

For now I will just mark it as failing in direct in #2190 and proceed

### DIFF
--- a/datalad/metadata/extractors/datalad_core.py
+++ b/datalad/metadata/extractors/datalad_core.py
@@ -53,13 +53,14 @@ class MetadataExtractor(BaseMetadataExtractor):
         if not isinstance(self.ds.repo, AnnexRepo):
             for p in self.paths:
                 # this extractor does give a response for ANY file as it serves
-                # an an indicator of file presence (i.e. a file list) in the
+                # as an indicator of file presence (i.e. a file list) in the
                 # content metadata, even if we know nothing but the filename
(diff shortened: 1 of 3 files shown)


mih:
OK, I will retract that patch. Doing it this way (in low-level code) would mean that I would have to emulate git annex behavior for arbitrary commands. Doing it one level up (in the extractor), I know much better what kind of path arguments will be coming in.

mih:
I won't be able to work on this. And I cannot turn this back into an issue, and it would be wrong to close it. @bpoldrack do you want to take a look at a low-level solution?

yarikoptic:
@bpoldrack - any thoughts on this one which could help us out?

mih:
This is now done properly, with proper semantic markup in -revolution.

Issue #2006, opened 2017-11-29
Implement `datalad serve -d <path>`


mih:
We are almost there.

Largely string based arguments to all API functions. Return values are essentialy JSON (if requested).

We only need a few extra lines (with e.g. flask) to be able to make dataset query/manipulation accessible via a web-request API.

Once done, we can make sure that our coming extensions of web stuff still go through the same underlying API, and help further standardizing and homogenizing our structures.

Having another interface without an attached console should also help to bring more clarity to the handling of messaging and outputs.

NB: flask can be nicely integrated with proper webservers via WSGI, while also coming with a simple built-in devserver.

mih:
Such a server instance should be able to fulfill OAI-PMH requests: 
www.openarchives.org/OAI/openarchivesprotocol.html

mih:
We now have https://github.com/datalad/datalad-webapp

PR #2212, opened 2018-02-21
NF: Framework for webapps (tech demo)

mih:
This picks up on the idea in gh-2006, but takes it into a slightly different direction that is more manageable and useful IMHO.

The basic idea is to have a bunch of "webapps" that can be plugged together and build a (custom) big/joint webapp composed of all selected pieces. Webapp components can come from datalad itself, but also from system or user content (idea and mechanism resemble a plugin).

The whole thing is built on cherrypy (yes, not flask), I like classes. It can handle everything I could think of. I included pointers to run it on daemon mode, on privileged ports (commented out ATM), serving static files, session management, ....

A demo webapp is included that does very little, but shows how we can relay our internal JSON-like results to the outside world, provide and eat forms, and serve static files.

To try it out you can run:

`datalad webapp -d <somedatasetwithmetadata> --app metadata`

then visit http://localhost:8080 to play with it. E.g. query metadata: http://localhost:8080/m?path=.

Alternatively run:

`datalad webapp -d <somedatasetwithmetadata> --app metadata /mountpoint`

to "mount" the webapp under http://localhost:8080/mountpoint (all methods and static files will move logically under the path). The included static file would then be found under http://localhost:8080/mountpoint/this.css. There is no mandatory "static" or something like this in the path to give complete flexibility, so a webapp could also simple serve the full content of a dataset.

There is a bit of dynamic config being done in the `webapp` command, but each app can also be configured using a proper config file (one example is also included).


bpoldrack:
Nice and works for me. Will build an additional app for this to gain deeper insight.

Two quick notes FTR:

- dependency appears to be cherrypy3
- we should catch errors and not expose internal paths (and possibly other details), if things go wrong:

screenshot from 2018-02-22 09-21-52

mih:
CVEs for cherrypy (the base layer): https://www.cvedetails.com/vendor/4097/Cherrypy.html

mih:
Re deps: yes, it is cherrypy3, but this is more of an API version. I used cherrypy 8.9, current is 14, I think. Looking through the history I saw that last relevant API breakage happened between 3.0 and 3.1 -- I thought that was a nice feature.

mih:
Changed the default config now. An HTTP response would now look like this by default:

```
% curl -I http://localhost:8080
HTTP/1.1 200 OK
Content-Length: 283
X-Xss-Protection: 1; mode=block
Content-Security-Policy: default-src='self'
Server: CherryPy/8.9.1
Date: Thu, 22 Feb 2018 13:30:22 GMT
X-Frame-Options: DENY
Content-Type: text/html;charset=utf-8
Set-Cookie: session_id=404f5339c2b5319904ac1f048750704492953e6c; expires=Thu, 22 Feb 2018 14:30:22 GMT; httponly; Path=/; secure
```

But I cannot guarantee that you get the same session ID ;-)

mih:
> we should catch errors and not expose internal paths (and possibly other details), if things go wrong

This is only happening when the server runs in debug mode. I will add a config demo that shows how to trigger the desired behavior. At this stage a production mode has no purpose ;-)

mih:
New plugin framework is now in place. No need to keep this open as a PR.

Issue #2085, opened 2018-01-17
 Elevate the plugin interface to behave like full-scale commands


yarikoptic:
#### What is the problem?
originally brought up in #2082  etc
Separate issue from metadata handling/search, so filing separately.

I personally would love if plugins would be exposed and behave as regular commands with a regular/conventional `--help` etc.  The aspect to keep in mind though is the start up time for `datalad` cmdline .  Although a "minor" thing, any increase in starting a command could have notable effects on duration of the tests battery etc due to all the sandwiching with git-annex calls etc

mih:
Q: Where is the difference between a command and a plugin if we do that?

I think there is none. I think this issue should be about making the existing commands more flexible and breaking this monolithic beast of commands into more manageable pieces. Other than generating the list of available commands for `datalad --help` there is no need to import all commands at startup (and we can replace that by some other way to get a one line synopsis for a command). And if that is no longer done, there is also no start-up cost for adding more commands with potentially heavy deps -- may they be plugins or commands. One step further: plugins can be abandoned altogether and be replaced by proper docs on how to write a command.

The current `plugin` command can become a factory function that promotes any Python function into a full-scale command class.

mih:
I played with this a little and looked at the amount of code that needs to run to get a proper command implementation hooked into the python or the cmdline API. Bottom line, there is not much (after I discovered that #2214 is possible). SO what I outlined above is doable: when can kill the `plugin` command and make any plugin a real command, we just need to add the ability to search a few dedicated places for commands, instead of hard-coding an exhaustive list.

Q remains: how to we deal with the lookup in order to avoid unnecessary import/startup delays. For the cmdline interface, it should be relatively easy to load a shim on startup that has just the name and a one-line synopsis on a command. However, I don't have a good idea on how to achieve a similar effect for the python api. The latter is used throughout the entire code base, meaning that any command using anything from the API, causes all of the API to be built -- which makes any/most optimizations done at the cmdline end more or less pointless.

Any ideas?

mih:
Ok, 40min no response -- clearly nobody cares....

So here is the master plan:

1. `api` -> `coreapi`
   This will then be pruned of any command that is not super-important (or a dependency of a super-important command). This new coreapi will then be loaded the same way as we do now (hardcoded list of imports)
2. new `api`
[...]

PR #2214, opened 2018-02-22
RF: API generation got much simpler, remove cruft

mih:
Stumbled upon this while looking into #2085

Despite that code being seemingly unused, its removal causes my local test runs to segfault :(


mih:
Those 42 lines did not run anywhere for at least a year. I'll merge this now, as it is objectively the right thing to do.

PR #2218, opened 2018-02-23
OPT: cmdline startup faster (more cruft removed)

mih:
This should have no visible differences for actual cmdline use, but
the startup time for individual commands should be noticeably reduced
Here is an example:

Pre:
```
In [2]: %timeit !datalad subdatasets
subdataset(ok): .asv (dataset)
subdataset(ok): .asv (dataset)
subdataset(ok): .asv (dataset)
subdataset(ok): .asv (dataset)
1 loop, best of 3: 317 ms per loop
```
Post:
```
In [3]: %timeit !datalad subdatasets
subdataset(ok): .asv (dataset)
subdataset(ok): .asv (dataset)
subdataset(ok): .asv (dataset)
subdataset(ok): .asv (dataset)
1 loop, best of 3: 223 ms per loop
```

### DIFF
--- a/datalad/interface/ls.py
+++ b/datalad/interface/ls.py
@@ -37,6 +37,8 @@ from ..support import ansi_colors
 from ..support.constraints import EnsureStr, EnsureNone
 from ..distribution.dataset import Dataset
 
+from datalad.distribution.subdatasets import Subdatasets
+
 from datalad.support.annexrepo import AnnexRepo
 from datalad.support.annexrepo import GitRepo
 
--- a/datalad/api.py
+++ b/datalad/api.py
@@ -1,0 +1,13 @@
+# emacs: -*- mode: python; py-indent-offset: 4; tab-width: 4; indent-tabs-mode: nil -*-
+# ex: set sts=4 ts=4 sw=4 noet:
+# ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+#
+#   See COPYING file distributed along with the datalad package for the
+#   copyright and license terms.
+#
+# ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+"""Python DataLad API exposing user-oriented commands (also available via CLI)"""
+
+from datalad.coreapi import *
+
+# TODO load plugin commands
--- a/datalad/cmdline/helpers.py
+++ b/datalad/cmdline/helpers.py
@@ -168,19 +168,6 @@ queue
         os.unlink(f.name)
 
 
-class RegexpType(object):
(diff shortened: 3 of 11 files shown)


bpoldrack:
Apart from sphinx failing - LGTM

mih:
@bpoldrack @yarikoptic Please be aware that this PR changes one important thing. So far datalad use via cmdline or Python API happened after the full list of commands got imported. With this PR this is no longeer the case for cmdline use. You can see a related failure at: https://travis-ci.org/datalad/datalad/jobs/345132077#L3278

```
I: running docs/examples/3rdparty_analysis_workflow.sh
[INFO   ] Creating a new annex repo at /tmp/datalad_demo_bob.2JRe/myanalysis 
[ERROR  ] 'Dataset' object has no attribute 'add' [create.py:__call__:368] (AttributeError) 
I: done running docs/examples/3rdparty_analysis_workflow.sh: failed
```

Nothing in create.py seems to cause the import of the `add` command. So we really seem to become a little leaner, but things like this might pop up in various places now, and need to get fixed with a proper corresponding import.

yarikoptic:
we have similar effects in pymvpa with `@datasetmethod` bound methods (heh heh -- even the name is the same ;-)). I always burn myself when trying to do regular pythonic imports instead of the `mvpa2.suite` abomination.  really not "user friendly".
May be that is where we should introduce some "explicit listing", i.e. to add an explicit list of what modules must be imported for a specific attribute so if not yet bound, the submodule gets imported and the corresponding method bound, with a test verifying that we have listed all?

datalad/interface/ls.py
```diff
@@ -37,6 +37,8 @@
 from ..support.constraints import EnsureStr, EnsureNone
 from ..distribution.dataset import Dataset
 
+from datalad.distribution.subdatasets import Subdatasets
```
yarikoptic: This approach is errorprone since I or someone else might remove those imports since `Add` is used nowhere in this code and it requires really deep knowledge to know that such imports are really needed.  In some spots you did leave a comment to preclude removal,  but I am just saying that it is very suboptimal

yarikoptic:
Not sure if @mih is working to make it even better and no label or description checkbox to signal that it is ready to be merged.  I do not mind it being merged due to OPT it introduces but I am afraid it would burn us some time(s) without a "proper" solution.

mih:
I don't think I will come up with a proper solution. The thing you are complaining about is only possible as the tests import from datalad.api. Once this stops, and it should, removing such import breaks the tests. The status quo, btw, provides absolutely no indication whatsoever that exactly these imports are already critical (and have to be made *in another file). Hence I don't see this quite as negative.

Issue #2210, opened 2018-02-21
Aggregate metadata (dicom) stumbles upon some unicode issue without traceback


bpoldrack:
#### What is the problem?
As reported by @cni-md, `aggregate-metadata` fails on a DICOM dataset with an unicode related issue. Reproduced it with more detailed logging. Sadly, we don't get any useful traceback by the logs - it throws up all the way to main.py before it gets logged.

```
[INFO   ] datalad:8>main:359>interface.base:432>utils:408,475>aggregate:690,198>metadata:411  Engage metadata extractors: ['datalad_core', 'dicom']
[INFO   ] datalad:8>main:359>interface.base:432>utils:408,475>aggregate:690,198>metadata:437>dicom:61  Attempting to extract DICOM metadata from 11396 files
[DEBUG  ] datalad:8>main:359>interface.base:432>utils:408,475>aggregate:690,236>dataset:437>utils:322  Determined class of decorated function: <class 'datalad.interface.unlock.Unlock'>
[DEBUG  ] ...>utils:475>aggregate:690,236>dataset:437>utils:463,451,408,475>unlock:91>utils:322  Determined class of decorated function: <class 'datalad.interface.annotate_paths.AnnotatePaths'>
[DEBUG  ] ...>interface.base:432>utils:408,475>aggregate:690,236>dataset:437>utils:463,451,408,505  not controlled by annex, nothing to unlock [unlock(/home/bpoldrack/scratch/test2/utf8-session/dicoms/.datalad/metadata/objects/d8/ds-c87983928f4268a24922883d65be8c)]
[ERROR  ] datalad:8>main:389  'utf8' codec can't decode byte 0xb0 in position 0: invalid start byte [utf_8.py:decode:16] (UnicodeDecodeError)
```

However, I ran it via python API (PY2) and got:

```
[INFO   ] Attempting to extract DICOM metadata from 11396 files
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python2.7/dist-packages/datalad/distribution/dataset.py", line 437, in apply_func
    return f(**kwargs)
  File "/usr/local/lib/python2.7/dist-packages/datalad/interface/utils.py", line 463, in eval_func
    return return_func(generator_func)(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/datalad/interface/utils.py", line 451, in return_func
    results = list(results)
[...]

PR #2225, opened 2018-02-25
Ignore PSgxxxxx DICOM files during metadata extraction

bpoldrack:
This pull request addresses #2210 (partially)

This pull request proposes to ignore DICOM files with a name starting with `PSg`, since those files (apparently created by some QC routine on Philips scanners) seem to not carry any metadata relevant for describing the image series.

This PR was supported by funds from the 
German federal state of Saxony-Anhalt and the European Regional Development 
Fund (ERDF), Project: Center for Behavioral Brain Sciences, Imaging Platform
### Changes
- [X] This change is complete

### DIFF
--- a/datalad/metadata/extractors/dicom.py
+++ b/datalad/metadata/extractors/dicom.py
@@ -10,7 +10,7 @@
 from __future__ import absolute_import
 
 from six import string_types
-from os.path import join as opj
+from os.path import join as opj, basename
 import logging
 lgr = logging.getLogger('datalad.metadata.extractors.dicom')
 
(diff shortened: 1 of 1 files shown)


bpoldrack:
Once we are more confident what they are about and how they look like, that rule should be more specific - agreed. I'm not even sure, they are necessarily Philips only. But we might want to confirm they don't carry anything relevant. After all the trouble they cause ATM isn't happening when they are read in, but when writing the extracted metadata.

Anyways, for now it's a workaround that should do.

PR #2238, opened 2018-02-27
ENH: Aggregated metadata query now utilized annotated path types

mih:


yarikoptic:
I bet it is great. Hard to say since there is no dedicated test to show what this all about.

mih:
It is true that there is not dedicated test (yet). This will come with a general extension of the tests in the context of the "useful" queries that we talked about. For now the most relevant aspect is that there is *no* change in behavior, across any of the search modes.

PR #2240, opened 2018-02-27
implemented fix URL method

overlake333:
This pull request fixes #
Implements fix url method to ensure correct URL format.
This pull request proposes

### Changes
- [x] This change is complete

### DIFF
--- a/datalad/crawler/pipelines/simple_with_archives.py
+++ b/datalad/crawler/pipelines/simple_with_archives.py
@@ -11,6 +11,7 @@
 # Import necessary nodes
 import os
 from ..nodes.crawl_url import crawl_url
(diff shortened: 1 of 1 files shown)


datalad/crawler/nodes/matches.py
```diff
@@ -176,7 +176,6 @@ def _select_and_extract(self, selector, query, data):
 
 class a_href_match(AExtractorMatch):
     """Helper to simplify matching based on URL while also extracting various tags from URL while at it
-
```
yarikoptic: be careful to not modify (without reason) code unrelated to your changes

here we are following numpy docstrings convention (linked in CONTRIBUTING.md) and an empty line is mandatory

PR #2242, opened 2018-02-28
NF: Singularity recipe for a full datalad installation from current master

mih:
The image behaves like the main datalad entrypoint.

%test could be added...

- [x] This change is complete

### DIFF
--- a/tools/Singularity-devmaster
+++ b/tools/Singularity-devmaster
@@ -25,9 +25,16 @@ From:debian:latest
     pip3 install --system master.zip[full]
     rm -f master.zip
 
+    # until the next pybids release grab it from master
+    pip3 install https://github.com/INCF/pybids/archive/master.zip
+
     # clean up
     rm -rf /tmp/nd-config*
     apt-get clean
+
+
+%environments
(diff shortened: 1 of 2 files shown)


mih:
FTR: Startup time of datalad from within the container is ~700ms on my system. Not stellar, but not too bad either.

tools/Singularity-devmaster
```diff
@@ -0,0 +1,33 @@
+Bootstrap:docker
+From:debian:latest
```
yarikoptic: You could just use NeuroDebian: latest and should manual config for datalad

mih: Looking at that image I had the impression that it would be much larger. Am I wrong?

I was actually aiming to switch to the slimified debian base image.

yarikoptic: You are basing on docker. Our docker images are as small as Debian ones, well almost
It is the singularity image is big some should be usable right away

mih:
I canceled travis, no point. I will merge this now to set up singularity hub. We can improve iteratively from there on.

PR #1528, opened 2017-05-23
drop python 3.4 and aim for 3.5 and 3.6

yarikoptic:
since gods say that 3.6.1 is the one!

PR #2248, opened 2018-03-01
Drop --doctest-tests from nose args

mih:
As per analysis in #1528. Let's see if we loose anything with this
change.


mih:
This looks like no meaningful change. I will merge this now to get the full test run on master.

kyleam:
> Hm so we didn't doctested the code for a while? Should be fixed then

I think that flag is limited to doctests in the test modules:

```
--doctest-tests
      Also look for doctests in test modules. Note that classes, methods and functions should have either doctests or non-doctest tests, not both. [NOSE_DOCTEST_TESTS]
```

`--with-doctest` is still present.

In https://github.com/datalad/datalad/pull/1528#issuecomment-369502410, @mih mentions that he didn't find any doctests within the tests.

PR #2253, opened 2018-03-05
Switch to PY3 for all exotic flavors on travis

mih:
#### What is the problem?
https://pythonclock.org

Majority of all testing is done exclusively for PY2. I don't see a reason why that should be the case, other than history.


mih:
Looking good. One (spurious?) failure -- I restarted the job.

mih:
Houston, we a re go for launch!

PR #2267, opened 2018-03-05
Locking at the repo level for datalad-archives

yarikoptic:
This pull request fixes #???

I thought that if I make our datalad-archives refuse to prepare the 2nd instance, git-annex would just use the single instance of the custom special remote while downloading files, but it is not the case:
http://git-annex.branchable.com/bugs/howto_guarantee_a_single_instance_of_a_special_remote__63__/

so this one is just to see what else would break.

Ideally we should just lock "per file/operation", i.e. if I need to get a file - we lock for that operation and then other instances would just wait for that.  Similar for "extraction" and possibly "Cleanup". This way they could all co-exist in parallel while waiting for another instance to take care about getting the file, or may be meanwhile requesting another archive (if such rare case happens) and extracting it to serve individual files "in parallel" to git-annex

### Changes
- [ ] This change is complete

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,24 @@ This is a high level and scarce summary of the changes between releases.
 We would recommend to consult log of the 
 [DataLad git repository](http://github.com/datalad/datalad) for more details.
 
+
+## 0.9.3 (??? ??, 2018) -- if ever be released
+
+bet we will fix some bugs and we might need to kick it out before 0.10.
+
+### Major refactoring and deprecations
+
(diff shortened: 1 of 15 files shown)


yarikoptic:
I was late (forgot) to push 0.9.x branch so this one lists all those changes as well, although now it is pushed. hopefully they would go away when I push next commits if any

yarikoptic:
wow -- no tests failed.... that is interesting.  Well, I will close it since I think #2268 will supersede this one to provide more targeted resolution for now

PR #2268, opened 2018-03-06
ENH/BF: --jobs got auto + fix with locking datalad-archives issue

yarikoptic:
This pull request fixes the problem of inability of our datalad-archives remote to work whenever git-annex get has -JN where N>1:  it would start N instances of git-annex-remote-datalad-archives which will start competing for calling `git annex get` for the target tarball key, and extracting it without realizing that there is another instance already running.

This pull request proposes to 
- lock within datalad-archives while requesting the archive to be fetched by annex, and while extracting.  
- by default do parallel downloads (starting with 3 up to the # of CPUs but not more than 8)

There is one issue remaining to be resolved on git-annex side but I think it would manifest only with tons of small files extracted from archives.

### Itemlist
- [x] This change is complete
- [ ] make it not fail at all! ;): the issue is in annex imho: http://git-annex.branchable.com/bugs/get_-JX__58____transfer_already_in_progress_..._for_some_files/ so out of our hands. we can only mitigate it by adding delays but that would be really suboptimal. The issue was fixed up for on git-annex side.
- [x] may be make it better looking (e.g. a content manager while leaving actual callable to be called within its body)
- [x] Unittest(s)

Now just want to see what fails


yarikoptic:
New pyld 1.0.1 brings devastation to 0.9.x search https://github.com/digitalbazaar/pyld/issues/78

yarikoptic:
grr... despite all the niceties and tests, the newer version actually doesn't work in real life -- smells like locking is not in effect while multiple custom remotes are asking for a download of the same key... grrr

yarikoptic:
ok, I do not think that cause for errors is on our side: I see no requests from annex logged on either side for the failing file.  git-annex just reports 

transfer already in progress, or unable to take transfer lock\nUnable to access these remotes: datalad-archives

I will test bleeding edge git-annex since joey I believe did some relevant fixups recently.

yarikoptic:
@datalad/developers the beast is ready for your feedback/merge

datalad/support/locking.py
```diff
@@ -0,0 +1,102 @@
...
+import os
+
+from contextlib import contextmanager
+
+import logging
+lgr = logging.getLogger('datalad.locking')
+
+
+def _get(entry):
+    """A helper to ge the value, be it a callable or callable with args, or value
```
kyleam: s/ge/get/

datalad/support/locking.py
```diff
@@ -0,0 +1,102 @@
...
+    check,
+    lock_path,
+    operation=None,
+    blocking=True,
+    **kwargs
+):
+    """A context manager to establish a lock conditionally on a check
+    function result
+
+    It is intended to be used to lock for a specific file and/or operation, e.g.
```
kyleam: s/for// or s/to lock/as a lock/ ?

datalad/support/locking.py
```diff
@@ -0,0 +1,102 @@
...
+        return
+    # acquire blocking lock
+    lock_filename = _get(lock_path)
+
+    lock_filename += '.'
+    if operation:
+        lock_filename += operation + '-'
+    lock_filename += 'lck'
+
+    # could use with!
```
kyleam: Is the comment just to indicate that the `try-finally` is being used with `contextmanager`?  I find the "could use" weird: Can't `lock_if_check_fails` only be used as a context manager now?

datalad/support/locking.py
```diff
@@ -0,0 +1,102 @@
...
+    for `annex get`ing a file or extracting an archive, so only one process
+    would be performing such an operation.
+
+    If verification of the check fails, it tries to acquire the lock, but if
+    that fails on the first try, it will rerun check before proceeding to func
+
+    checker and lock_path_prefix could be a value, or callable, or
+    a tuple composing callable and its args
+
+    Unfortunately yoh did not find any way (in Python 2) to have context manager
```
kyleam: s/context/a &/

datalad/support/locking.py
```diff
@@ -0,0 +1,102 @@
...
+    which just skips the entire block if some condition is met (in python3 there
+    ExitStack which could potentially be used.  So we would need still to check
```
kyleam: s/there/& is/

datalad/support/locking.py
```diff
@@ -0,0 +1,102 @@
...
+    in the block body context manager return value is not None.
```
kyleam: s/body/& if the/ ?

yarikoptic:
thank you @kyleam for the review.  I have adjusted the description of the PR to state the problem first. will address your comments now.

datalad/support/locking.py
yarikoptic: I meant about the `fasteners.InterProcessLock` itself ;)  but for now just will remove this comment

datalad/support/locking.py
yarikoptic: didn't know about `&` -- thanks!

yarikoptic:
filter out the commit changing demanded outdated version of pyld -- 1.0.3 should have the fix(es) we need, so tests should pass with it now

yarikoptic:
@datalad/developers any feedback/action would be welcome or I will proceed merging it.

Issue #918, opened 2016-09-26
`datalad configure` command to initialize/modify basic settings


mih:
Looks through a list of key `datalad.*` config variables, reports the default them and asks for a potential update.

This could be useful to make it more transparent how things are set up (initially), e.g. where is the cache ... Or to change a user's configuration in one go.

Implementation is essentially a loop around `cfg.obtain()` once #915 is merged.

yarikoptic:
BTW, just FTR, there is a number of possible config options which I haven't made config calls for but was going to at some point to make configurable
```shell
$> git grep 'TODO: config'
datalad/cmd.py:            # TODO: config cmd.protocol = null
datalad/cmd.py:                # TODO: config cmd.protocol_prefix = protocol
datalad/crawler/nodes/annex.py:            # TODO: config crawl.subdataset.existing = skip|raise|replace|crawl|adjust
datalad/crawler/nodes/annex.py:            # TODO: config crawl.subdataset.crawl_new = false|true
datalad/crawler/nodes/annex.py:                    # TODO: config.tag.sign
datalad/crawler/nodes/annex.py:                        # TODO: config.tag.allow_override
datalad/crawler/nodes/crawl_url.py:                # TODO: config  -- crawl.failed='skip' should be a config option, for now always skipping
datalad/crawler/nodes/s3.py:            # TODO: config crawl.crawl_s3.versiondb.saveaftereach=True
datalad/customremotes/base.py:        self._contentlocations = DictCache(size_limit=100)  # TODO: config ?
datalad/dochelpers.py:        # TODO: config logging.exceptions.traceback_levels = 1
datalad/interface/add_archive_content.py:                    if not exists(link_path):  # TODO: config  addarchive.symlink-broken='skip'
datalad/interface/crawl.py:                # TODO: config.crawl.failure = full-reset | last-good-master
datalad/log.py:        # TODO: config log.timestamp=True
datalad/log.py:                           # TODO: config log.name, pid
datalad/support/status.py:        # TODO: config crawl.mtime_delta

```

mih:
Thanks for the list. I started in #922.

mih:
After some more thinking: I no longer believe this to be a good idea -- at least at this point. We have no solid understanding of the amount and dependency structure of config items, yet. Until we do such a command can only be a crude hack that is barely more than a `git config` wrapper.

I will instead try to implement support for auto-generated documentation on the available config items.

mih:
Done #939

Issue #2270, opened 2018-03-06
ConfigManager.set() not behaving as documented


mih:
```
       force: bool
          if set, replaces all occurrences of `var` by a single one with the
          given `value`. Otherwise raise if multiple entries for `var` exist
          already
```

I don't see this happening. Instead, it just replaces all of them with the new value.

bpoldrack:
What exactly don't you see happening?

mih:
With `force=False` it `replaces all occurrences of `var` by a single one with the given `value` -- no exception is being raised `if multiple entries for `var` exist already`

bpoldrack:
I get this:

```
CommandError: CommandError: command '['git', 'config', '--local', 'annex.some', 'different']' failed with exitcode 5
Failed to run ['git', 'config', '--local', 'annex.some', 'different'] under '/tmp/test'. Exit code=5. out= err=warning: annex.some has multiple values
error: cannot overwrite multiple values with a single value
```

What git version are you using?

mih:
% git --version
git version 2.16.1

bpoldrack:
Hm. Okay, may be that changed.

% git --version
git version 2.13.0.rc1.294.g07d810a77f

bpoldrack:
Can you verify it's about git by showing what the following is resulting in for you?

```
/tmp % mkdir test
/tmp % cd test 
/tmp/test % git init
Initialized empty Git repository in /tmp/test/.git/
/tmp/test % git config --add test.some first
/tmp/test % git config --add test.some second
/tmp/test % cat .git/config
[core]
	repositoryformatversion = 0
	filemode = true
	bare = false
	logallrefupdates = true
[test]
	some = first
	some = second
/tmp/test % git config --local test.some new
warning: test.some has multiple values
error: cannot overwrite multiple values with a single value
       Use a regexp, --add or --replace-all to change test.some.
```

mih:
I get the same thing that you get. Dunno what this is about. Ignore it for now, it makes no sense to have two people spending time on this.

PR #2271, opened 2018-03-06
BF: `configure` command (fixes gh-918)

mih:
First crude implementation of a command to configure DataLad and
datasets, interactively.

This is not meant as a replacement for the ConfigManager, nor as a
direct cmdline frontend for it. I believe `git config` works much better
for this case, and supports way more corner cases.

OTOH I think we do need something that bootstraps an initial
configuration for a fresh installation, and maybe for a new dataset
(until with have that templating mechanism).

There are no tests, and I won't write them, because I am not (yet)
confident that this a viable path to go down. In any case, the actual
interaction with the rest of the codebase is minimal (ui, and cfg).
And it already exposed a bug #2270.

This change is complete.


yarikoptic:
Just a comment on the description (haven't looked inside yet): I do still feel that we do need `datalad config` and that `git config` is too generic -- we need to be able to provide a list of config options we are aware about, what metadata extractors are available, etc. to make it explicit what config is modified (global/repo/dataset) without relying on user to know where each should go to.

mih:
After the recent changes to make an initial configuration optional, or at least force it, we could probably get away without this command. Closing.

yarikoptic:
well, I think we might like to have it regardless of optionality of the initial configuration.  E.g. present all known config options and let users possibly configure those they are interested in.

yarikoptic:
I think we should reincarnate this one.  I have also pushed 0d5044831d2c81c3c7182cb2dc0d72a1677caf5f  to call `configure` if mode is interactive.  The reason is that for new to git users all that "configuration" complaint is still a dark forest -- is it a github configuration? is that a repository configuration?

yarikoptic:
Re

>  I believe git config works much better for this case, and supports way more corner cases.

I would somewhat disagree.  We do need `config` functionality with --recursive semantics, with clear global/local/dataset separation of where config should go to instead of hoping that user would magically know all the places he could point `git config` to

mih:
I think we can close this again.

PR #2278, opened 2018-03-07
`Dataset.get_superdataset()` should be able to verify subdatasets

mih:
ATM this function return any dataset above a given dataset -- with `topmost=True` all the way up. However, despite the name, it does not actually check whether a "superdataset" actually knows about the original dataset as being a true subdataset (might be totally unrelated, or branch was switched, or previous version checked out, ...). As this function is a key player when saving dataset hierarchies, this behavior can lead to undesired outcomes, i.e. when a non-subdataset suddenly become one, just because somewhere upstairs another dataset is hanging around.

I think `get_superdataset()` should crosscheck by default, and get a switch to not do that when it is OK/desired.

In case you are wondering: yes, we do have a test that ensures that this "broken" behavior is maintained ;-)

### DIFF
--- a/datalad/crawler/nodes/annex.py
+++ b/datalad/crawler/nodes/annex.py
@@ -157,7 +157,9 @@ class initiate_dataset(object):
         )
         if self.add_to_super:
             # place hack from 'add-to-super' times here
-            sds = ds.get_superdataset()
+            # MIH: tests indicate that this wants to discover any dataset above
+            # not just true superdatasets
+            sds = ds.get_superdataset(registered_only=False)
             if sds is not None:
                 lgr.debug("Adding %s as a subdataset to %s", ds, sds)
                 sds.add(ds.path, save=False)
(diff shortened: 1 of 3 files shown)


mih:
This is ready.

mih:
Under high logging this fails:

```
ERROR: datalad.support.tests.test_annexrepo.test_is_available(True, False)
  File "/home/travis/virtualenv/python3.5.5/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/build/datalad/datalad/datalad/tests/utils.py", line 836, in newfunc
    t(*(arg + (uri,)), **kw)
  File "/home/travis/build/datalad/datalad/datalad/support/tests/test_annexrepo.py", line 1354, in test_is_available
    assert_false(is_available("boguskey", key=True))
  File "/home/travis/build/datalad/datalad/datalad/support/gitrepo.py", line 297, in newfunc
    for f in files_new
  File "/home/travis/build/datalad/datalad/datalad/support/gitrepo.py", line 297, in <listcomp>
    for f in files_new
  File "/home/travis/build/datalad/datalad/datalad/support/annexrepo.py", line 2845, in is_available
    "Received output %r from annex, whenever expect 0 or 1" % out
ValueError: Received output None from annex, whenever expect 0 or 1
```

I cannot see how this is related to this PR.

PR #2279, opened 2018-03-07
TST/BF: Adapt assertions on error message to new annex version

bpoldrack:
This pull request fixes test assertions, that check the error message of `git-annex-drop`, which changed recently.

Please note, that #2277 for example already suffers from that in a not-allowed-to-fail-build.

### Changes
- [ ] This change is complete

### DIFF
--- a/datalad/distribution/tests/test_uninstall.py
+++ b/datalad/distribution/tests/test_uninstall.py
@@ -35,6 +35,7 @@ from datalad.tests.utils import assert_raises
 from datalad.tests.utils import assert_status
 from datalad.tests.utils import assert_in
 from datalad.tests.utils import assert_result_count
+from datalad.tests.utils import assert_result_values_cond
 from datalad.tests.utils import ok_file_under_git
 from datalad.tests.utils import ok_clean_git
 from datalad.tests.utils import with_tempfile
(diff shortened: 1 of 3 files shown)


bpoldrack:
Didn't instantly work, so figuring out what's different on Travis ...

datalad/distribution/tests/test_uninstall.py
```diff
@@ -402,8 +412,21 @@ def test_kill(path):
...
+    assert_result_values_cond(
+            [err_result], 'message',
+            lambda x: "(Use --force to override this check, "
+                      "or adjust numcopies.)" in x
+    )
+    assert_result_values_cond(
+        [err_result], 'message',
+        lambda x: "configured minimum number of copies not found" in x or
+        "Could only verify the existence of 0 out of 1 necessary copies" in x
+    )
```
yarikoptic: seems to be a repeating pattern.  Shouldn't we have a dedicated helper
`assert_result_not_enough_copies(err_result, 'message')`

datalad/distribution/tests/test_uninstall.py
```diff
@@ -402,8 +412,21 @@ def test_kill(path):
...
+        status='error', path=testfile)
+    # Following two assertions on message are relying on the actual error.
+    # We have a second result with status 'impossible' for the ds, that we need
+    # to filter out for those assertions:
+    err_result = [r for r in res if r['status'] == 'error'][0]
+    assert_result_values_cond(
+            [err_result], 'message',
+            lambda x: "(Use --force to override this check, "
+                      "or adjust numcopies.)" in x
+    )
```
yarikoptic: the same here (as below):
`assert_result_require_numcopies_or_force(err_result, 'message')`
may be even making `message` to be the default field

datalad/distribution/tests/test_uninstall.py
bpoldrack: That's quite a specific helper. Actually, I think we should unify on lower level, so the call to drop would raise an `InsufficientNumCopiesError`, that could be detected independently on a version specific output. Then the result can have a stable message, that we can test for herein as before.

Issue #763, opened 2016-09-07
[wishlist] use `git describe` as version provider for the dataset (I guess during aggregation)


yarikoptic:
some datasets are more or less nicely versioned by "upstream", e.g.

``` shell
$> datalad ls ds*
ds000001  [annex]  master  2.0.0+4 2016-09-03/20:02:21  OK
ds000002  [annex]  master  1.0.0-3-g0a9f3b6 2016-09-03/20:03:21  OK
ds000003  [annex]  master  1.1.0+4 2016-09-03/20:03:37  OK
ds000005  [annex]  master  2.0.0+3 2016-09-03/20:03:57  OK
ds000006  [annex]  master  2.0.0+2 2016-09-03/20:09:44  OK
ds000007  [annex]  master  2.0.0+3 2016-09-03/20:10:44  OK
ds000008  [annex]  master  2.0.0+2 2016-09-03/20:11:08  OK
ds000009  [annex]  master  2.0.1+1 2016-09-03/20:12:31  OK
```

that version description comes from git repository itself and `+revision` depicts "crawled" revision (I am still tuning things up).  So it would be nice if for such datasets I could state to use `git describe` output for the version

mih:
Just to be sure: Something from git-describe should end up in the aggregated meta data, correct? Just a plain `git describe` call, or maybe we should have a first config option that informs us what parameters are to be used?

yarikoptic:
On September 8, 2016 7:04:10 AM EDT, Michael Hanke <notifications@github.com> wrote:
>Just to be sure: Something from git-describe should end up in the
>aggregated meta data, correct? Just a plain `git describe` call, or
>maybe we should have a first config option that informs us what
>parameters are to be used?
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/datalad/datalad/issues/763#issuecomment-245564530

I think more than that - I would like not to go and edit every dataset meta to specify new version every time it updates - that is done atm via standard git tags mechanism. So I would like to indicate that for specific dataset to use output of git describe as a version. I could of cause adjust crawler to duplicate that information in the tag and in meta, but I don't like duplication

yarikoptic:
@mih may be a compromise could be the output of e.g.
```shell
> git describe --match '[v0-9][0-9.]*' --always --long
2.0.1-0-g7ec9ffd
```
so yes -- we would enforce some kind of semantic versioning, and if none tag existed:
```shell
[...]

PR #2289, opened 2018-03-08
Pimp the datalad_core extractor to expose more of datalad's internal/implicit metadata

mih:
This is also the way that I see info like #763 get incorporated.

### DIFF
--- a/datalad/metadata/aggregate.py
+++ b/datalad/metadata/aggregate.py
@@ -204,6 +204,11 @@ def _extract_metadata(agginto_ds, aggfrom_ds, db, to_save):
         content_meta=None,
         paths=relevant_paths)
 
+    # inject the info which commmit we are describing into the core metadata
+    # this is done here in order to avoid feeding it all the way down
+    coremeta = dsmeta.get('datalad_core', {})
+    coremeta['version'] = refcommit
+    dsmeta['datalad_core'] = coremeta
     # shorten to MD5sum
     objid = md5(objid.encode()).hexdigest()
 
--- a/datalad/metadata/aggregate.py
+++ b/datalad/metadata/aggregate.py
@@ -192,7 +192,7 @@ def _extract_metadata(agginto_ds, aggfrom_ds, db, to_save):
     # if there is any chance for metadata
     # obtain metadata for dataset and content
     relevant_paths = sorted(_get_metadatarelevant_paths(aggfrom_ds, subds_relpaths))
-    nativetypes = ['datalad_core'] + assure_list(get_metadata_type(aggfrom_ds))
+    nativetypes = ['datalad_core', 'annex'] + assure_list(get_metadata_type(aggfrom_ds))
     agginfo['extractors'] = nativetypes
     agginfo['datalad_version'] = datalad.__version__
     dsmeta, contentmeta, errored = _get_metadata(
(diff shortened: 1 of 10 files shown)


datalad/metadata/extractors/datalad_core.py
```diff
@@ -41,6 +42,28 @@ def _get_dataset_metadata(self):
             obj['@context'] = obj['definition']
             del obj['definition']
         obj['@id'] = self.ds.id
+        subdsinfo = [{
+            # this version would change anytime we aggregate metadata, let's not
+            # do this for now
+            #'version': sds['revision'],
+            'type': sds['type'],
+            'name': sds['gitmodule_name'],
```
yarikoptic: would this be the place to eventually be able to ingest some "relationship" between datasets as [FAIR I3](https://www.go-fair.org/fair-principles/i3-metadata-include-qualified-references-metadata/) -- created a dedicated #2290

datalad/metadata/extractors/tests/test_nidm.py
```diff
@@ -37,6 +37,8 @@ def test_nidm(path):
     res = ds.metadata(reporton='datasets')
     # ATM we do not forsee file-based metadata to come back from NIDM
     assert_result_count(res, 1)
+    # kill version info
+    res[0]['metadata']['datalad_core'].pop('version')
```
yarikoptic: and that is where other types of "versions" should be popped out I guess

datalad/metadata/search.py
```diff
@@ -135,7 +135,7 @@ def _deep_kv(basekey, dct):
                 # temp files during index staging, and trips over absent "directories"
                 # TODO maybe even kill parentheses
                 # TODO actually, it might be better to have an explicit whitelist
-                k.replace(os.sep, '_').replace(' ', '_').replace('-', '_').replace('.', '_').replace(':', '-')
+                k.lstrip('@').replace(os.sep, '_').replace(' ', '_').replace('-', '_').replace('.', '_').replace(':', '-')
```
yarikoptic: it grows and might grow more... IMHO we should get a helper to make it more readable, e.g. the one in https://stackoverflow.com/a/2484247 which would have called `replace_many` or alike
```python
substitutions = [
    ('<', '&lt;'),
    ('>', '&gt;'),
    ...]

for search, replacement in substitutions:
    string = string.replace(search, replacement)
```
since apparently regular `.replace` even if chained still wins in speed over more sophisticated but explicit replacement solutions in that post

datalad/metadata/aggregate.py
```diff
@@ -203,6 +203,11 @@ def _extract_metadata(agginto_ds, aggfrom_ds, db, to_save):
         content_meta=None,
         paths=relevant_paths)
 
+    # inject the info which commmit we are describing into the core metadata
+    # this is done here in order to avoid feeding it all the way down
+    coremeta = dsmeta.get('datalad_core', {})
+    coremeta['version'] = refcommit
```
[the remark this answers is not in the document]
mih: Fine with me.

datalad/metadata/extractors/datalad_core.py
mih: I think that IS the information we have on the downward dataset relationships. Anything in adding would require the presence of a subdataset, which we can neither require nor guarantee.

datalad/metadata/search.py
mih: I would not oppose such move.

datalad/metadata/aggregate.py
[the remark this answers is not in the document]
mih: I want to get this done today. Can you please point out the desired what to get a proper version for us? Do we have a helper that deals with all the special cases (no tags, ....)?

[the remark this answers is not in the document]
yarikoptic: We could start with `aggfrom_ds.repo.describe` which ATM does not handle any special cases.  If there is no describe == no version, imho it is ok, just life.  `refcommit` would stay there for unique (but not comparable etc) identifier of the version

PR #2297, opened 2018-03-13
ENH: -c and -s|--sensitive=(all|some) for wtf, uniform(er) formatting, fix for __version_full__, more externals versions

yarikoptic:
I generally cannot use 'datalad wtf' since it dumps the entirety of my bloated `git config`.  That is why I introduced `--no-sensitive` to avoid exposing largely irrelevant git config and dataset metadata.
Also with -c now could be just copied to clipboard to be pasted into the browser/report

### Changes
- [x] This change is complete


datalad/plugin/wtf.py
```diff
@@ -35,12 +37,23 @@ class WTF(Interface):
             no dataset is given, an attempt is made to identify the dataset
             based on the current working directory.""",
             constraints=EnsureDataset() | EnsureNone()),
+        no_sensitive=Parameter(
+            args=("-ns", "--no-sensitive",),
```
kyleam: Not necessarily a reason not to use `-ns`, but grepping around, it looks like this would be the first spot in datalad where the short option consists of more than one letter.

datalad/support/external_versions.py
```diff
@@ -251,5 +255,14 @@ def dumps(self, indent=None, preamble="Versions:", query=False):
             out += " " + ' '.join(items)
         return out
 
+    def require(self, name, min_version=None, msg=""):
```
kyleam: This new method looks around and sees that all its neighbors have beautiful docstrings.  It stares out at the night sky and wonders, "Why do I exist? What's my purpose in all of this?"

datalad/plugin/wtf.py
yarikoptic: that is true!  any recommendations on what should it be otherwise?

datalad/support/external_versions.py
yarikoptic: oh, I made it into a bastard... sorry -- will fix now

datalad/plugin/wtf.py
kyleam: No good ones.  I don't like `-n` because it makes me think "dry run".  Since you didn't use it in the first place, I'm guessing you also don't like it, perhaps for the same reason.  If you renamed the long option to `--exclude-sensitive`, I think `-x` would work.  The long option would be longer, but I guess that's not a big deal when there is a short option.

yarikoptic: ok, will do exclude (although I do like shorter --no)

mih:
The level of poetry in this PR brings me joy!

One suggestions: I cannot think of a reason why "no-sensitive" should not be default. The previous default behavior was exactly that, only less information was filtered. This should also solve the `-ns` situation.

yarikoptic:
makes sense, we could make it `--sensitive=(some|all)` where in `some` it would add sections but still would filter out those which you marked as sensitive and `all` would list all values.  We could also issue a warning if `-c` and `--sensitive`.

yarikoptic:
@mih - what are your thoughts or this minor non-issue, so I could RF it to our agreement?

mih:
My 2ct: full filter and no filter is enough choice from my POV. Given that it is an explicit flag no warning is needed.

yarikoptic:
merged master so we could may be see WTF causing the occasional failure of datalad.metadata.tests.test_aggregation.test_aggregate_query

yarikoptic:
failures aren't our fault.... @datalad/developers - any feedback/desires or should we just merge?

kyleam:
`-s={all,some}` sounds fine to me.

yarikoptic:
ok, I have got one blessing, so merging

PR #2299, opened 2018-03-14
RF: do not raise if git is not configured -- just issue a warning

yarikoptic:
This pull request fixes #2272 

This pull request proposes to not fail if there is no git configuration -- just issue a warning.

The rationale is that git became even more "robust" in assigning some author name/email, and annex does some tricks too...  but if we always demand configuration -  that just complicates the simplest consumer usage.

### Changes
- [x] This change is complete

Please have a look @datalad/developers

Issue #1718, opened 2017-08-15
--annonimize-date 


yarikoptic:
git annex now allows to specify the date for to be used while recording time stamp(s) within git-annex branch files.  It allows now to fake all the dates stored in the git/git-annex repo:

```shell
$> da=40; d0="Thu Jan 1 00:00:$da 1970 +0000"; rm -rf /tmp/test; GIT_ANNEX_VECTOR_CLOCK=$da GIT_AUTHOR_DATE="$d0" GIT_COMMITTER_DATE="$d0" datalad create /tmp/test
[INFO   ] Creating a new annex repo at /tmp/test
create(ok): /tmp/test (dataset)                                                                                                                         
2 10381.....................................:Tue 15 Aug 2017 01:33:12 PM EDT:.
hopa:/tmp
$> git -C /tmp/test log -p git-annex
commit 5af2c2a04319a362417250aef269002e25c1d356 (git-annex)
Author: Yaroslav Halchenko <debian@onerussian.com>
Date:   Thu Jan 1 00:00:40 1970 +0000

    update

diff --git a/uuid.log b/uuid.log
new file mode 100644
index 0000000..748db51
--- /dev/null
+++ b/uuid.log
@@ -0,0 +1 @@
+97b0f01c-d7eb-44ed-813f-7c0723fd8a90 yoh@hopa:/tmp/test timestamp=40s

commit 0096b601c565ac1efd2535363880ebe7c8f278e1
Author: Yaroslav Halchenko <debian@onerussian.com>
Date:   Thu Jan 1 00:00:40 1970 +0000
```

We are going to use it within the scope of heudiconv, but I wondered if we should provide interface to this functionality within datalad itself.

- pros: might come handy in other use cases where dates should be obfuscated
- cons: REALLY easy to shot yourself in the foot, so it would warrant at least some checks (that previous commit in git and git-annex has date(s) before or == etc)

ljchang:
This seems like it could be useful functionality to have built into datalad at some point and not fully relying on heudiconv.  I'm imagining a scenario where someone wants to publicly publish a dataset and wants to ensure the data are fully anonymized.  This might include scrubbing names and scrambling dates.  They might not have used heudiconv at the outset and also not realize that the data creation dates might be tied to identifying subject information.

yarikoptic:
The strategy should probably be
- for initial commit - take some date in the past
- take the latest date from
  - the last commit within the branch and git-annex branch in the dataset
[...]

PR #2300, opened 2018-03-14
RF/BF: get_commit_date to return authored date by default, handle detached head as well

yarikoptic:
As pyout showed some times we were getting None for the date and it was
happening in detached head because we would just return None then since
there is no branch.

Later this function could be extended to print date of any commit, not just
the last one in the branch etc.

By default now returns "authored", not "committed" (thus rename) date since
it is the one of interest typically -- if date was altered, better be the
one specified

- [x] This change is complete

### DIFF
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -999,7 +999,7 @@ class GitRepo(RepoInterface):
         return DATALAD_PREFIX if not msg else "%s %s" % (DATALAD_PREFIX, msg)
 
     def commit(self, msg=None, options=None, _datalad_msg=False, careless=True,
-               files=None):
+               files=None, date=None):
         """Commit changes to git.
 
         Parameters
@@ -1016,12 +1016,16 @@ class GitRepo(RepoInterface):
           if True, don't care
         files: list of str, optional
           path(s) to commit
+        date: str, optional
+          Date in one of the formats git understands
         """
 
         self.precommit()
 
         if _datalad_msg:
             msg = self._get_prefixed_commit_msg(msg)
+
+        options = options or []
 
         if not msg:
             if options:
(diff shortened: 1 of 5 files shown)


datalad/support/gitrepo.py
```diff
@@ -999,7 +999,7 @@ def _get_prefixed_commit_msg(msg):
         return DATALAD_PREFIX if not msg else "%s %s" % (DATALAD_PREFIX, msg)
 
     def commit(self, msg=None, options=None, _datalad_msg=False, careless=True,
-               files=None):
+               files=None, date=None):
```
kyleam: Are you adding this argument just for testing?  If so, why not use the environment variables `GIT_AUTHOR_DATE` and `GIT_COMMITTER_DATE`?

yarikoptic: well, why should I use environment variables if command line has a dedicated option?

kyleam: Because you're adding an option that you otherwise have no use case for.  Also, this option only allows you to change the author, whereas but with the variables you could test both cases.

yarikoptic: could you fix it up to your liking? ;)  I was just trying to mimic cmdline, which ppl would be more familiar.
But indeed, having said that, we might want to keep in mind (not to be a part of this PR, but changes here should not preclude and only facilitate, thus env vars might be better) https://github.com/datalad/datalad/issues/1718 which relates and yet TODO.  interested to help with that one as well @kyleam -- it is needed for reproin ATM

kyleam: > could you fix it up to your liking? ;) I was just trying to mimic cmdline, which ppl would be more familiar.

It's fine as is, of course.  I was just curious.

yarikoptic:
I have CPed fixes for unrelated test failing here in travis within #2299 so should be ok to merge otherwise ;)

yarikoptic:
while noone is watching

PR #2304, opened 2018-03-15
OPT: cache (lru_cache) calls to guess_ri_cls

yarikoptic:
Since that information is persistent (type or parsing would not change through time)
and very frequently queried - having it cached would be benefitial and also avoids
logs clutter.

Instead of using our little explicit DictCache I have used lru_cache which is
built-in since 3.2, so I have added an (untested) backport from the web.  Seems to work.
Our mighty tests will show

### Changes
- [x] This change is complete


yarikoptic:
I am not intending to provide exhaustive testing of the backported code... otherwise -- seems (not explicitly unittested) to work, and locally I observed much less cluttered logs, so a win.  If no objections -- will merge later today @datalad/developers

PR #2305, opened 2018-03-15
BF 2276: pass path_src into commit as well (so we commit renamed files removal as well)

### LINKED ISSUE #2276: datalad run  misses original files which were renamed and leaves them deleted
```shell
$> datalad create /tmp/testrunrename                           
[INFO   ] Creating a new annex repo at /tmp/testrunrename 
create(ok): /tmp/testrunrename (dataset)                                                                                                                                                                                                      

$> cd /tmp/testrunrename

$> echo 123 > 123; datalad add 123
add(ok): /tmp/testrunrename/123 (file)                                                                                                                                                                                                        
save(ok): /tmp/testrunrename (dataset)                                                                                                                                                                                                        
action summary:
  add (ok: 1)
  save (ok: 1)

$> datalad run "git mv 123 124"
add(notneeded): 124 (file) [already included in the dataset]
save(ok): /tmp/testrunrename (dataset)
action summary:
  add (notneeded: 1)
  save (ok: 1)

$> git status
On branch master
Changes to be committed:

	deleted:    123

```
0.9.2

yarikoptic:
This pull request fixes #2276

This pull request proposes to provide `path_src` portions of the annotated paths while doing the commit in `save`.  Also while at it in aa0143b14982438392e98eb24880fc312d13f113  I did something along the same lines -- return path_src as well in annotated2content_by_ds  if only paths are requested.  I am not sure even if that line of logic gets triggered by any test/use-case.  Running all the tests will show. Otherwise it should be ready for review etc

### Changes
- [x] This change is complete

### DIFF
--- a/datalad/interface/annotate_paths.py
+++ b/datalad/interface/annotate_paths.py
@@ -10,6 +10,7 @@
 
 __docformat__ = 'restructuredtext'
 
+from collections import defaultdict
 
 import logging
 import textwrap
@@ -68,7 +69,7 @@ def annotated2ds_props(annotated):
     return props
 
 
-def annotated2content_by_ds(annotated, refds_path, path_only=False):
+def annotated2content_by_ds(annotated, refds_path):
     """Helper to convert annotated paths into an old-style content_by_ds dict
 
     Only items with an `status` property value not equal to 'ok', 'notneeded',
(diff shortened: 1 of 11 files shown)


datalad/interface/annotate_paths.py
```diff
@@ -124,21 +125,28 @@ def annotated2content_by_ds(annotated, refds_path, path_only=False):
...
-                toappendto.append(r['path'] if path_only else r)
+                toappendto.append(r_path if path_only else r)
                 content_by_ds[parentds] = toappendto
         else:
             # files and dirs
             # common case, something with a parentds
             toappendto = content_by_ds.get(parentds, [])
-            toappendto.append(r['path'] if path_only else r)
+            if path_only:
+                r_path_src = r.get('path_src')
```
yarikoptic: so it seems that nowhere in the code path_only=True is used
```shell
$> git grep 'path_only=[^F]'
```
I would be happy to remove this option/logic if @mih agrees to such lobotomization ;-)

mih: This option was needed for compat with legacy code. If it is no longer needed, I would be happy to see it go.

yarikoptic: cool, then I will do it in this PR so we do not forget. thanks

yarikoptic:
ok, RFed a bit, all green, merging

PR #2309, opened 2018-03-15
Enh merge 0.9.x into master

yarikoptic:
- locking -- in master we could enable "auto" as a default for "--jobs", and boost demanded git annex version to tomorrow's ;-)
- caching or RIs
Altogether would be interesting to see if any benefit to benchmarks

I think I did the right things while resolving the conflicts, review would be appreciated

### Changes
- [x] This change is complete


yarikoptic:
buildbot failures should be bogus,  but let's just keep this afloat and I will merge more stuff in as 0.9.3 gets released

yarikoptic:
ok, merging since no objections

Issue #2301, opened 2018-03-14
test_basic_aggregate: json.decoder.JSONDecodeError: Extra data: line 1 column 166 (char 165)


yarikoptic:
#### What is the problem?
Happened (or may be regularly) in that heavy debug mode matrix run (only):
https://travis-ci.org/datalad/datalad/jobs/353137426

```
======================================================================
ERROR: datalad.metadata.tests.test_aggregation.test_basic_aggregate
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/virtualenv/python3.5.5/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/build/datalad/datalad/datalad/tests/utils.py", line 494, in newfunc
    return t(*(arg + (d,)), **kw)
  File "/home/travis/build/datalad/datalad/datalad/tests/utils.py", line 905, in newfunc
    return func(*args, **kwargs)
  File "/home/travis/build/datalad/datalad/datalad/tests/utils.py", line 1057, in newfunc
    return func(*args, **kwargs)
  File "/home/travis/build/datalad/datalad/datalad/metadata/tests/test_aggregation.py", line 67, in test_basic_aggregate
    base.aggregate_metadata(recursive=True)
  File "/home/travis/virtualenv/python3.5.5/lib/python3.5/site-packages/wrapt/wrappers.py", line 562, in __call__
    args, kwargs)
  File "/home/travis/build/datalad/datalad/datalad/distribution/dataset.py", line 436, in apply_func
    return f(**kwargs)
  File "/home/travis/virtualenv/python3.5.5/lib/python3.5/site-packages/wrapt/wrappers.py", line 523, in __call__
    args, kwargs)
  File "/home/travis/build/datalad/datalad/datalad/interface/utils.py", line 470, in eval_func
    return return_func(generator_func)(*args, **kwargs)
  File "/home/travis/virtualenv/python3.5.5/lib/python3.5/site-packages/wrapt/wrappers.py", line 523, in __call__
    args, kwargs)
  File "/home/travis/build/datalad/datalad/datalad/interface/utils.py", line 458, in return_func
    results = list(results)
  File "/home/travis/build/datalad/datalad/datalad/interface/utils.py", line 414, in generator_func
    result_renderer, result_xfm, _result_filter, **_kwargs):
  File "/home/travis/build/datalad/datalad/datalad/interface/utils.py", line 482, in _process_results
[...]

PR #2314, opened 2018-03-16
RF: No special casing of pybids needed anymore, but versioned dep

mih:


mih:
Apparently, this change is wrong in some way. Pybids does not get pulled in during singularity image build...

mih:
Travis failure is #2301

PR #2325, opened 2018-03-19
NOT A PR, PLEASE LEAVE UNTOUCHED

mih:
Just to get an opinion from the build bots, Please ignore.


mih:
All failures are exactly as expected. The only "surprise" is the lzma handling in json_py.py, which is only necessary for metadata handling. The rest works just fine.

PR #2331, opened 2018-03-22
NF: Minimal NIDM extractor (blackbox in and out)

mih:


datalad/metadata/extractors/tests/test_nidm.py
```diff
@@ -37,27 +50,9 @@ def test_nidm(path):
     res = ds.metadata(reporton='datasets')
     # ATM we do not forsee file-based metadata to come back from NIDM
     assert_result_count(res, 1)
-    # show full structure of the assembled metadata from demo content
```
yarikoptic: master had some changes in here which causes a conflict now - I do not exactly know what would be the best resolution of the conflict there without possibly weakening the test

mih:
Now at https://github.com/datalad/datalad-neuroimaging/pull/3

PR #2349, opened 2018-03-27
Metadata unique value sort broken for PY3

mih:
The reason is an unexpected behavior change of the built-in `sorted()` for built-in types. Demo:

```
% python -c "sorted({('5.1.1', '5.1.1.0'), '0.4.11'})" && echo "happy"
happy
% python3 -c "sorted({('5.1.1', '5.1.1.0'), '0.4.11'})" && echo "happy"
TypeError: '<' not supported between instances of 'tuple' and 'str'
```

I do not have an idea how to fix that.


mih:
Travis failure is an old friend ("extra data")

Issue #2326, opened 2018-03-20
RF "0.10" metadata handling into a separately installable unit


mih:
This will implement an idea we had a while back (cannot find the issue ATM).

Advantages for the core tool: 
- shrink the code base by about 7k lines (that is more than 15% of the entire Python code base)
- massive reduction in the dependencies (including a reduced focus on neuroimaging as a field of application)
- faster testing

Advantages for the metadata functionality:
- separate release cycle at a different pace
- muuuch faster testing
- compatibility with multiple versions of datalad

The metadata package will be a regular python package that declares an install-dependency on datalad (>=0.10), as it will need the newly added ability for dynamically loading full command implementations. This can then also serve a role model for other extensions (crawler is a candidate), tailored DICOM/BIDS support (bids2scidata) is another. Naturally such extensions can also depend on each other.

Inclusion into DataLad proper will work similar to "plugins". However, extensions can provide entire suits of commands, hence will be able to declare "section names" for proper listing in `datalad --help`. Extensions will be able to shadow commands provided by datalad core, to enable development of customized extensions that can later on easily move into datalad core without requiring API changes, and user-end modifications.

Given the 9 month development that make up the diff between 0.9 and 0.10 a selective rebase is probably not feasibly as means to get it out of the mainline.
```
% git log --oneline 0.9.3..master |wc -l
821
```
However, it is quite easily possible to remove all the code at once (I test drove that in #2325), and replace the respective matching files with the code in 0.9.x. There will be some effort necessary to revert the (minimal) adjustments made to the crawler, but other than that it should be pretty much copy&paste, as there are no further contact point with the metadata handling code and the rest. The only thing we need to look out for is the LZMA support handling in json_py. While this could stay in, datalad core doesn't use it in any way and it should not be a dependency. The metadata tool will then have to use another helper in its own code base.

If you have a better idea on how to wrangle the Git history to get this out in a clean fashion, please chime in.

I assigned this to the 0.10 release milestone, as we must sort this out prior release.

yarikoptic:
Even after having thought about it I am still having a split opinion.  Could we discuss it (before action happens) when we see each other in a few hours instead of torturing the keyboards?

mih:
We can now have extension modules (#2353), but also decided to make this part of the core.

PR #2353, opened 2018-03-28
NF: Add generic support for independent extension modules

mih:
In the spirit of #2326 this enables independent packages to provide
DataLad command (suites), simply by defining the required info in a
structure reachable via a setuptools entrypoint.

This following repo implements a rather minimal extension module that
demonstrates this. The idea is that people could fork this, if they
want to implement an extension.

https://github.com/datalad/datalad-module-template

If this gets installed, DataLad will offer an additional "hello..."
command in a dedicated (and labeled) suite.

Not sure how we want to test this, but we could start installing this
extension in the standard travis builds and see that it does what it
should.

### DIFF
--- a/datalad/api.py
+++ b/datalad/api.py
@@ -25,7 +25,26 @@ def _load_plugins():
         globals()[camel.sub('\\1_\\2', pi.__name__).lower()] = pi.__call__
 
 
+def _generate_modules_api():
+    """Auto detect all available modules and generate an API from them
+    """
+    from importlib import import_module
+    from pkg_resources import iter_entry_points
+    from .interface.base import get_api_name
+
+    for entry_point in iter_entry_points('datalad.modules'):
+        grp_descr, interfaces = entry_point.load()
+        for intfspec in interfaces:
+            # turn the interface spec into an instance
+            mod = import_module(intfspec[0])
+            intf = getattr(mod, intfspec[1])
+            api_name = get_api_name(intfspec)
+            globals()[api_name] = intf.__call__
+
+
+_generate_modules_api()
 _load_plugins()
 
 # Be nice and clean up the namespace properly
 del _load_plugins
+del _generate_modules_api
--- a/datalad/cmdline/main.py
+++ b/datalad/cmdline/main.py
@@ -24,6 +24,8 @@ import os
 
 from six import text_type
 
+from pkg_resources import iter_entry_points
+
 import datalad
 
 from datalad.cmdline import helpers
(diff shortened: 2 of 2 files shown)


datalad/api.py
```diff
@@ -25,7 +25,26 @@ def _load_plugins():
...
 
 
+def _generate_modules_api():
+    """Auto detect all available modules and generate an API from them
+    """
+    from importlib import import_module
+    from pkg_resources import iter_entry_points
+    from .interface.base import get_api_name
+
+    for entry_point in iter_entry_points('datalad.modules'):
```
yarikoptic: why should it be a separate namespace/special?  i.e. shouldn't we then define our internal functionality as modules?  This mechanism is using "Interface", so may be this one should just be "datalad.interface"?

mih: Sure, but maybe one step after the other. Who knows how/why this breaks. Also interface is already heavily overloaded. Primary objective here is to allows for 3rd-party extensions, not for RF'ing datalad itself. If we find this to be working fine, we can adopt it in general. but please remember that _you_ remove the use of setuptools entrypoints for the main cmdline entrypoint, because it was sooo slow.

yarikoptic: yeap, I remember that I did it.

datalad/api.py
```diff
@@ -25,7 +25,26 @@ def _load_plugins():
...
+        grp_descr, interfaces = entry_point.load()
+        for intfspec in interfaces:
+            # turn the interface spec into an instance
+            mod = import_module(intfspec[0])
```
yarikoptic: let's wrap in try/except handling of each interface here so we do not crash if some rogue module is installed but just issue a warning

datalad/api.py
```diff
@@ -25,7 +25,26 @@ def _load_plugins():
...
+            intf = getattr(mod, intfspec[1])
+            api_name = get_api_name(intfspec)
+            globals()[api_name] = intf.__call__
```
yarikoptic: I wondered if we would want to check if it would not overload something previously loaded, and issue a warning (or at least debug message) so we know that multiple  beasts compete for the same spot

datalad/api.py
yarikoptic: and please add a debug msg.  It is is hard to troubleshoot when no log messages are provided

mih: Done

mih: Done

datalad/api.py
mih: Done

PR #2367, opened 2018-03-30
Make --show-keys more useful - report stats

yarikoptic:
I bet there is a better/proper way but I could not think about one... I thought that may be `metadata` call on the superdataset could just report those aggregated unique values but I was not smart enough to figure out if that was possible.

This pull request proposes to make --show-keys more versatile and report statistics on the key values since they are known at no additional cost while listing them.  My original goal was to check what unique values for dicom fields we collected and in how many datasets.  While still waiting (for an hour now) for output of (a more informative, since provides pairs)
```shell

$> datalad -f '{metadata[dicom][Manufacturer]}: {metadata[dicom][ManufacturerModelName]}  Version(s):{metadata[dicom][SoftwareVersions]}' --report-type file search --max-nresult=0 --mode=autofield 'dicom.Manufacturer:*' -f > /tmp/allhits
```
I achieved following with the suggested here changes:
```shell
$> datalad --dbg search --show-keys=full | grep -A2 dicom.Manufa
dicom.Manufacturer
 in  16 datasets
 has 3 unique values: u'SIEMENS', u'GE MEDICAL SYSTEMS', u'Philips Medical Systems'
dicom.ManufacturerModelName
 in  16 datasets
 has 7 unique values: u'Prisma', u'SIGNA HDx', u'DISCOVERY MR750', u'Achieva', u'TrioTim', u'Prisma_fit', u'Intera'

$> datalad --dbg search --show-keys=short | grep -A2 dicom.Manufa    
dicom.Manufacturer
 in  16 datasets
 has 3 unique values: u'SIEMENS', u'GE MEDICAL SYSTEMS', u'Philips Medical Systems'
dicom.ManufacturerModelName
 in  16 datasets
 has 7 unique values: u'Prisma', u'SIGNA HDx', u'DISCOVERY MR750', u'Achieva', u'TrioTim', u'Prisma_f ....

$> datalad --dbg search --show-keys=name | grep -A2 dicom.Manufa     
dicom.Manufacturer
dicom.ManufacturerModelName
dicom.MedicalAlerts
dicom.Modality
```

Feel free to scrape it especially if above results could be obtained in some different, smoother way

### Changes
- [x] This change is complete

### DIFF
--- a/datalad/metadata/search.py
+++ b/datalad/metadata/search.py
@@ -41,7 +41,7 @@ from datalad.support.constraints import EnsureBool
 
 from datalad.consts import LOCAL_CENTRAL_PATH
 from datalad.consts import SEARCH_INDEX_DOTGITDIR
-from datalad.utils import assure_list
+from datalad.utils import assure_list, assure_iter
 from datalad.utils import assure_unicode
 from datalad.support.exceptions import NoDatasetArgumentFound
 from datalad.ui import ui
@@ -231,8 +231,8 @@ class _Search(object):
     def __call__(self, query, max_nresults=None):
         raise NotImplementedError
 
-    def show_keys(self):
-        raise NotImplementedError
+    def show_keys(self, *args):
+        raise NotImplementedError(args)
 
     def get_query(self, query):
         raise NotImplementedError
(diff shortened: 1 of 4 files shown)


mih:
FTR: I think this should eventually be RF'ed to yield this info in result records and enhance the renderer to achieve the same output.

yarikoptic:
The cons of yield approach is always doing computation which night not be needed. Relates to our brief discussion on the way to Montreal airport on lazy loading of full metadata records

PR #2370, opened 2018-04-01
NF: `test` can pick up extension test suites and run them too

mih:
Without requiring a user to keep track of what extensions are installed
on a particular system.

@bpoldrack This implements your idea. The neuroimaging extensions already registers its module as tests.


mih:
I cannot see why the buildbots are that unhappy.

PR #2344, opened 2018-04-02
`extract-metadata` command does not need to use `--file` anymore

mih:
Could now be a simple positional arg, as the interface is not more powerful.


mih:
Test failures are old friends.

Issue #2342, opened 2018-03-25
Think about when and when not to produce a list of unique metadata values


mih:
Use case: #2339 

Each file can report multiple URLs (likely all unique), hence a list of all unique URLs across all files in a dataset can be quite large.

mih:
Closing duplicate

PR #2350, opened 2018-04-04
Need some clever way of limiting the output of uniquification of metadata values

mih:
The can grow really large. Example: DICOM slice timestamps aggregated over a 2h session (24 per seconds...).

For the `dicom` extractor is makes sense to disable uniquification completely, because it produces nice `Series` metadata, but we need something for the general case.

Also, unique value keys is what `search --mode autofield` bases the index build on during the first pass, hence we have to keep those for such use cases (what keys do you know).

Apparently already a concern in #2342

### DIFF
--- a/datalad/interface/test.py
+++ b/datalad/interface/test.py
@@ -15,6 +15,7 @@ __docformat__ = 'restructuredtext'
 import datalad
 from .base import Interface
 from datalad.interface.base import build_doc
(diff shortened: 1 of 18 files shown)


mih:
See https://github.com/datalad/datalad-neuroimaging/commit/c149dc823ac259ca64b356e542dbe4c2353fd577 for an impact assessment.

PR #2377, opened 2018-04-05
Change test setup to run *installed* datalad

mih:
It hit me in the past, and I still find it suboptimal, hence this proposal:

Let's change the travis setup to run tests on the installed datalad. ATM we run the tests on the result of `setup.py build` and only after the tests ran, we try to install.

At the same time, we offer a `test` command that is designed to pick up the installed tests.

With the new extension setup things become even weirder, as they need to be installed in order for the entrypoints to have any effect.

If there is any benefit to testing the non-installed datalad over the installed one, please let me know. Otherwise, I will push this change.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ env:
     - BOTO_CONFIG=/tmp/nowhere
     - DATALAD_TESTS_SSH=1
     - DATALAD_LOG_CMD_ENV=GIT_SSH_COMMAND
-    - TESTS_TO_PERFORM=
+    - TESTS_TO_PERFORM=datalad
     - NOSE_OPTS=-s
     - NOSE_SELECTION_OP="not "   # so it would be "not (integration or usecase)"
     # Special settings/helper for combined coverage from special remotes execution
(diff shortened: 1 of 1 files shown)


mih:
I made a first change in this direction. Feel free to jump in with your ideas, and push straight into this PR.

.travis.yml
```diff
@@ -224,8 +224,8 @@ install:
   - npm install grunt-contrib-qunit
 
 script:
-  # Verify that setup.py build doesn't puke
-  - python setup.py build
+  # Test installation system-wide
+  - sudo pip install .
   # Run tests
   - WRAPT_DISABLE_EXTENSIONS=1 http_proxy=
```
yarikoptic: Are you sure it is enough to not cd somewhere away from the source base? I was under impression that nose would pickup datalad directory right there

mih: We should do that, but I wanted to go one at a time. So far it seems nothing broke with just shifting the install to an earlier step.

mih:
Good to go it seems.

yarikoptic:
Osx issue is clearly unrelated to this PR Anand I haven't seen such a failure before
```
datalad.tests.test_cmd.test_runner_heavy_output(True,)
  File "/Users/datalad/buildslave/datalad-pr-dl-osx-64/build/venv-ci/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/datalad/buildslave/datalad-pr-dl-osx-64/build/datalad/tests/utils.py", line 1290, in newfunc
    func(*args, **kwargs)
  File "/Users/datalad/buildslave/datalad-pr-dl-osx-64/build/datalad/tests/test_cmd.py", line 221, in check_runner_heavy_output
    assert not ret[0], "all messages went into `logged`"
AssertionError: all messages went into `logged`
```

Issue #2378, opened 2018-04-04
Make storing aggregated metadata in leaf datasets optional


mih:
Sometimes leaf datasets are just temporarily needed or not meant to be modified. ATM metadata aggregation will always alter a leaf dataset it ran extractors on. This should be optional.

PR #2387, opened 2018-04-06
 NF: Switch to select type of metadata aggregation update (fixes gh-2378)

mih:
Mode `all` what we used to do: extract from all datasets and update in all datasets throughout the entire tree.

Now there is a mode `target` that still extracts from all datasets, but only updates the aggregated metadata in the top-level target/base/ref dataset.

The `target` mode is a more suitable default IMHO, as otherwise any user gets subdatasets modified whenever they had no metadata pre-aggregated. However, in general one likely does not want to create a diff to upstream just because we want to get metadata aggregated in a local superdataset.

Due to the change in default, all previous usage of `aggregate_metadata()` has the mode spelled out now.

### DIFF
--- a/datalad/crawler/nodes/annex.py
+++ b/datalad/crawler/nodes/annex.py
@@ -1316,7 +1316,7 @@ class Annexificator(object):
 
             if aggregate:
                 from datalad.api import aggregate_metadata
-                aggregate_metadata(dataset='^', path=self.repo.path)
+                aggregate_metadata(dataset='^', path=self.repo.path, update_mode='all')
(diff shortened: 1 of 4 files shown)


yarikoptic:
I am yet to comprehend what is going on here... ;-)  But in the big one I would really need to retain something like `--since=` to avoid any reaggregation among datasets which didn't change.  So far I didn't check it anyhow

mih:
--since is unaffected. This only determines where results of an aggregation are stored, not where aggregation happens.

Issue #1962, opened 2017-11-12
datalad run   runs the command in the top directory of the dataset


yarikoptic:
#### What is the problem?

must run it in the subdirectory where invoked.

#### What steps will reproduce the problem?

```shell
hopa:/tmp
$> datalad install ///openfmri/ds000001 
[INFO   ] Cloning http://datasets.datalad.org/openfmri/ds000001 to '/tmp/ds000001' 
^CERROR: 
Interrupted by user while doing magic: KeyboardInterrupt() [subprocess.py:_communicate_with_poll:1156]
(dev) 2 5561 ->1.....................................:Sun 12 Nov 2017 11:41:30 AM EST:.
hopa:/tmp
$> export http_proxy=http://127.0.0.1:3128 https_proxy=http://127.0.0.1:3128 ftp_proxy=http://127.0.0.1:3128
(dev) 2 5562.....................................:Sun 12 Nov 2017 11:41:32 AM EST:.
hopa:/tmp
$> datalad install ///openfmri/ds000001                                                                     
[INFO   ] Cloning http://datasets.datalad.org/openfmri/ds000001 to '/tmp/ds000001' 
[INFO   ] access to dataset sibling "datalad" not auto-enabled, enable with:
| 		datalad siblings -d "/tmp/ds000001" enable -s datalad 
install(ok): /tmp/ds000001 (dataset)
(dev) 2 5563.....................................:Sun 12 Nov 2017 11:41:37 AM EST:.
hopa:/tmp
$> cd ds000001/sub-01/
anat/  func/
(dev) 2 5564.....................................:Sun 12 Nov 2017 11:41:44 AM EST:.
(git-annex)hopa:/tmp/ds000001[master]sub-01
$> datalad run "echo 123 > 123"        
[INFO   ] == Command start (output follows) ===== 
[INFO   ] == Command exit (modification check follows) ===== 
add(ok): 123 (file)                                                                                             
save(ok): /tmp/ds000001 (dataset)
action summary:
  add (ok: 1)
  save (ok: 1)
(dev) 2 5565.....................................:Sun 12 Nov 2017 11:41:48 AM EST:.
(git-annex)hopa:/tmp/ds000001[master]sub-01
$> ls
anat/  func/
(dev) 2 5566.....................................:Sun 12 Nov 2017 11:41:49 AM EST:.
(git-annex)hopa:/tmp/ds000001[master]sub-01
$> ls ..
[...]

PR #1963, opened 2018-04-06
unlock reported path is incorrect if operating on ds within subdir

mih:
#### What is the problem?

was afraid that `add` etc have the same problem, but seems to be only unlock:
```shell
(git-annex)hopa:~/datalad/openfmri/ds000001[master]sub-01
$> datalad add -d .. sub-01/anat/sub-01_T1w.nii.gz           
add(ok): sub-01/anat/sub-01_T1w.nii.gz (file)                                                                         
save(notneeded): /home/yoh/datalad/openfmri/ds000001 (dataset)
action summary:
  add (ok: 1)
  save (notneeded: 1)

$> datalad unlock -d .. sub-01/anat/sub-01_T1w.nii.gz
unlock(ok): sub-01/sub-01/anat/sub-01_T1w.nii.gz (file)
```
as you can see  -- there is sub-01/sub-01

found while fixing for #1962 , and was confused by reported unlock -- so replicated independently

### DIFF
--- a/datalad/api.py
+++ b/datalad/api.py
@@ -25,8 +25,8 @@ def _load_plugins():
         globals()[camel.sub('\\1_\\2', pi.__name__).lower()] = pi.__call__
 
 
-def _generate_modules_api():
-    """Auto detect all available modules and generate an API from them
+def _generate_extension_api():
+    """Auto detect all available extensions and generate an API from them
     """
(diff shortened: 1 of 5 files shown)


mih:
No dedicated test, instead #2392.

PR #2400, opened 2018-04-09
BF: pass newline straight into _process_one_line to avoid joining empty new lines

yarikoptic:
for handled entries

Revealed by good old OSX that a recent fix in fe62e46955f5a428bcf72b28a0861194a59183b0
introduced breeding trailing \n in the output which should have already been handled
by the output processor. See e.g. http://smaug.datalad.org:8020/builders/datalad-pr-dl-osx-64/builds/3979/steps/nosetests/logs/stdio

The problem was that even if processor returns None, i.e. the line was already handled
we would have returned a newline to outside where we would have joined them with \n.
Fix, is to pass the suffix (used only in the case of handling those manually split lines)
to add to the end of the line if there was one remaining.

It should fix OSX buildbot issue

### DIFF
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -326,13 +326,13 @@ class Runner(object):
                 # resolving a once in a while failing test #2185
                 if isinstance(out_, text_type):
                     out_ = out_.encode('utf-8')
-                out += linesep_bytes.join(
-                    self._process_one_line(*pargs, line=line)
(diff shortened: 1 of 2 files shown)


yarikoptic:
I am not sure what is up with coverage... the only guess is that something in master manages to test more :-/
otherwise it is good and should resolve the osx buildbot failure, so merging

PR #2386, opened 2018-04-09
Access denied handling in downloader is broken

yarikoptic:
#### What is the problem?

Check this:
https://github.com/datalad/datalad/blob/master/datalad/downloaders/base.py#L173

It tries to reraise an exception, but the code is not part of any exception handling.


yarikoptic:
osx failures should have been addressed in master, otherwise it is good to go and even have tests coverage for that logic (besides an exception throw which shouldn't even be reachable)

PR #2406, opened 2018-04-10
Gets confused by symbolic links in the path

yarikoptic:
#### What is the problem?
```shell
(datalad) Taylors-MacBook-Pro-4:tmp taylor$  datalad create --text-no-annex demo
[INFO   ] Creating a new annex repo at /tmp/demo 
create(ok): /tmp/demo (dataset)                                                                                      
(datalad) Taylors-MacBook-Pro-4:tmp taylor$ cd demo/
(datalad) Taylors-MacBook-Pro-4:demo taylor$ datalad -l debug crawl-init --save --template=simple_with_archives url=http://example.com
...
[DEBUG  ] Assigning credentials into 18 providers 
[INFO   ] Not adding annex.largefiles=exclude=README* and exclude=LICENSE* to git annex calls because already defined to be (not(mimetype=text/*)) 
[DEBUG  ] Creating crawler configuration for template simple_with_archives under . 
                                                                                                                     [DEBUG  ] Determined class of decorated function: <class 'datalad.interface.save.Save'> 
[DEBUG  ] Determined class of decorated function: <class 'datalad.interface.annotate_paths.AnnotatePaths'> 
[WARNING] path not part of the reference dataset at /tmp/demo [save(/private/tmp/demo/.datalad/crawl/crawl.cfg)] 
[DEBUG  ] could not perform all requested actions: Command did not complete successfully [{'status': 'impossible', 'orig_request': './.datalad/crawl/crawl.cfg', 'raw_input': True, 'message': ('path not part of the reference dataset at %s', '/tmp/demo'), 'parentds': '/private/tmp/demo', 'action': 'save', 'path': '/private/tmp/demo/.datalad/crawl/crawl.cfg', 'type': 'file', 'refds': '/tmp/demo'}] [utils.py:generator_func:446] 
(datalad) Taylors-MacBook-Pro-4:demo taylor$ echo $TMPDIR
/var/folders/_r/lnsx28mx03v58m6p1vfml2b00000gn/T/
(datalad) Taylors-MacBook-Pro-4:demo taylor$ ls -ld /tmp
lrwxr-xr-x@ 1 root  wheel  11 Mar 22 22:03 /tmp -> private/tmp
(datalad) Taylors-MacBook-Pro-4:demo taylor$ pwd
/tmp/demo
(datalad) Taylors-MacBook-Pro-4:demo taylor$ python -c 'import os.path; print(os.path.realpath("/var/folders/_r/lnsx28mx03v58m6p1vfml2b00000gn/T"))'
/private/var/folders/_r/lnsx28mx03v58m6p1vfml2b00000gn/T
(datalad) Taylors-MacBook-Pro-4:demo taylor$ ls -ld /var
lrwxr-xr-x@ 1 root  wheel  11 Mar 22 22:03 /var -> private/var
(datalad) Taylors-MacBook-Pro-4:demo taylor$ datalad --version
datalad 0.9.3.dev1004
```

buy a Mac, do the above

Master branch 0.9.3-1004-g7635f467 

it seems to be quite a regular setup on OSX 10.13.3 (Sierra)

### DIFF
--- a/datalad/utils.py
+++ b/datalad/utils.py
@@ -1115,7 +1115,7 @@ def get_path_prefix(path, pwd=None):
     assumed
     """
     pwd = pwd or getpwd()
-    path = dlabspath()
+    path = dlabspath(path)
     path_ = with_pathsep(path)
     pwd_ = with_pathsep(pwd)
     common = commonprefix((path_, pwd_))
--- a/datalad/utils.py
+++ b/datalad/utils.py
@@ -1088,6 +1088,20 @@ class chpwd(object):
             self.__class__(self._prev_pwd, logsuffix="(coming back)")
 
 
+def dlabspath(path, norm=False):
+    """Symlinks-in-the-cwd aware abspath
+
+    os.path.abspath relies on os.getcwd() with would not know about symlinks
+    in the path
+    TODO: we might want to norm=True by default to match behavior of os
+    .path.abspath?
+    """
+    if not isabs(path):
+        # if not absolute -- relative to pwd
+        path = opj(getpwd(), path)
+    return normpath(path) if norm else path
+
+
 def with_pathsep(path):
     """Little helper to guarantee that path ends with /"""
     return path + sep if not path.endswith(sep) else path
@@ -1101,9 +1115,7 @@ def get_path_prefix(path, pwd=None):
     assumed
     """
     pwd = pwd or getpwd()
-    if not isabs(path):
-        # if not absolute -- relative to pwd
-        path = opj(getpwd(), path)
+    path = dlabspath()
     path_ = with_pathsep(path)
     pwd_ = with_pathsep(pwd)
     common = commonprefix((path_, pwd_))
--- a/datalad/utils.py
+++ b/datalad/utils.py
@@ -1091,10 +1091,11 @@ class chpwd(object):
 def dlabspath(path, norm=False):
     """Symlinks-in-the-cwd aware abspath
 
-    os.path.abspath relies on os.getcwd() with would not know about symlinks
+    os.path.abspath relies on os.getcwd() which would not know about symlinks
     in the path
-    TODO: we might want to norm=True by default to match behavior of os
-    .path.abspath?
+
+    TODO: we might want to norm=True by default to match behavior of
+    os .path.abspath?
     """
     if not isabs(path):
         # if not absolute -- relative to pwd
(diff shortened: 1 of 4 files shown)


yarikoptic:
added a unittest which should fail throughout all runs everywhere

kyleam:
Looks like it has a v6-specific failure:

```
FAIL: datalad.interface.tests.test_save.test_symlinked_relpath
  File "/home/travis/virtualenv/python3.5.5/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/build/datalad/datalad/datalad/tests/utils.py", line 664, in newfunc
  File "/home/travis/build/datalad/datalad/datalad/interface/tests/test_save.py", line 365, in test_symlinked_relpath
    ok_clean_git(dspath)
  File "/home/travis/build/datalad/datalad/datalad/tests/utils.py", line 285, in ok_clean_git
    eq_(head_diffs, [])
AssertionError: ['later\n=======================================================\nlhs: 100644 | bca257e554753db146185be2a7bd27ba97cf4765\nrhs: 100644 | 01d82652a31fd656a07e40d93369d3dffcc8003a', 'mike1\n=======================================================\nlhs: 100644 | ed570d9e2ed0ddba8c0bf50aefe5e184a8132a01\nrhs: 100644 | 463c3575d1cc551913e480b97843963c93954a37'] != []
```

yarikoptic:
Yikes, if  basic ok_clean_git doesn't work in v6... My hopes are ruined. Joking aside it is probably that issue if inability to retain the file under git, not annex. For this test it doesn't matter really, so I will make it all just a simple add (in an hour or so, unless you beat me to it) without forcing git on them. Then should work in v6 as well I guess

datalad/utils.py
```diff
@@ -1088,6 +1088,20 @@ def __exit__(self, exc_type, exc_val, exc_tb):
             self.__class__(self._prev_pwd, logsuffix="(coming back)")
 
 
+def dlabspath(path, norm=False):
+    """Symlinks-in-the-cwd aware abspath
+
+    os.path.abspath relies on os.getcwd() with would not know about symlinks
```
kyleam: s/with/which/ (tried to just push the fixup, but permission was denied.  perhaps something to do with the issue->pr conversion?)

yarikoptic: done, thanks!

indeed, it might be that with issue->PR by default it makes it unmarked... something to check how to change ;-)

yarikoptic:
@kyleam  re v6 -- kept it committing to git, but took a different approach on how to skip ;-)  decorators are quite neat indeed

yarikoptic:
fails only due to #4207, so ok to merge imho

yarikoptic:
ok, let's merge

Issue #2408, opened 2018-04-10
git config  calls are again abundant


yarikoptic:
#### What is the problem?

While testing the introduced lean'ish test for #2406 I was surprised with the 'slowness' of it.  Looked at what is called and apparently it seems that `git config` calls are again quite abundant imho without a reason (duplicates etc while nothing is really changing in the repo):
```shell
$> DATALAD_LOG_TIMESTAMP=1 DATALAD_LOG_LEVEL=1 python -m nose -s -v datalad/interface/tests/test_save.py:test_symlinked_relpath 2>&1 | tools/dtime | grep git.*config     
   21 2018-04-10 14:10:32,194 [Level 9] Running: ['git', 'config', 'user.name'] 
    0 2018-04-10 14:10:32,215 [Level 8] Finished running ['git', 'config', 'user.name'] with status 0 
   22 2018-04-10 14:10:32,215 [Level 9] Running: ['git', 'config', 'user.email'] 
    1 2018-04-10 14:10:32,237 [Level 8] Finished running ['git', 'config', 'user.email'] with status 0 
   18 2018-04-10 14:10:32,256 [Level 9] Running: ['git', 'config', '-z', '-l', '--show-origin'] 
    2 2018-04-10 14:10:32,274 [Level 8] Finished running ['git', 'config', '-z', '-l', '--show-origin'] with status 0 
   21 2018-04-10 14:10:32,277 [Level 9] Running: ['git', 'config', 'user.name'] 
    0 2018-04-10 14:10:32,298 [Level 8] Finished running ['git', 'config', 'user.name'] with status 0 
   17 2018-04-10 14:10:32,298 [Level 9] Running: ['git', 'config', 'user.email'] 
    1 2018-04-10 14:10:32,315 [Level 8] Finished running ['git', 'config', 'user.email'] with status 0 
   17 2018-04-10 14:10:32,393 [Level 9] Running: ['git', 'config', '-z', '-l', '--show-origin'] 
    0 2018-04-10 14:10:32,410 [Level 8] Finished running ['git', 'config', '-z', '-l', '--show-origin'] with status 0 
   18 2018-04-10 14:10:32,410 [Level 9] Running: ['git', 'config', '-z', '-l', '--show-origin'] 
    1 2018-04-10 14:10:32,428 [Level 8] Finished running ['git', 'config', '-z', '-l', '--show-origin'] with status 0 
[...]

PR #2410, opened 2018-04-10
TEMP: proof of concept -- delay reloading of config until anything is actually interested in it

yarikoptic:
inspired by #2408 
sits on top of #2406 

ATM there is way too many reloads of the config happening, just to make sure we are all good and kosher... this puts a toll.

This one is just a proof of concept to also see what breaks ;)


yarikoptic:
probably disregard it for now -- if we want to do all the dances turning direct/indirect and crippled fs back and forth using external processes (not through our .config) -- I guess it was doing "the right thing" while already relying on mtimes.  IT is not entirely clear why mtimes checks failed I guess (since it did run the same cmd for the same repo without external changes AFAIK).  `delayed_reload` also so far didn't show much benefit since I guess we do not have sweeps of configuration changes (when we do we explicitly say to not reload until the end)

mih:
Ok, so let's close this unmerged for now.

yarikoptic:
Want to check what tests fail currently with this changes

yarikoptic:
It was a temp, and I didn't pursue further

PR #2413, opened 2018-04-11
DOC: document how to use asv

### LINKED ISSUE #1790: Put benchmark run into Makefile?
I have to confess that I don't know how to run the benchmarks on my system. Any chance that this knowledge could make it into the Makefile? Thanks.

yarikoptic:
This pull request fixes #1790 (although without putting anything into Makefile since not clear what should it actually run there)

Also added hefty hyperreferences within our "file tree". We should populate with even more pointers where relevant.

### Changes
- [ ] This change is complete
- [ ] may be the entire Benchmarking section should be moved under `benchmarks/README.md` and just linked within CONTRIBUTING.md

@kyleam - could you please have a look and see if it is sufficient for others to start using asv.


mih:
Related: asv wants to run Python 3.4 -- shouldn't we converge on the versions that we are using in the tests? 3.5 for sure, possibly 3.6?

mih:
I pushed the .asv. Does it take additional action to make *my* results show up on http://datalad.github.io/datalad/#summarylist?sort=0&dir=asc ?

CONTRIBUTING.md
```diff
@@ -356,6 +357,96 @@ splits/joins making it easier to maintain code following the PEP8
 recommendations.  NeuroDebian provides `pycharm-community-sloppy`
 package to ease pycharm installation even further.
 
+### Benchmarking
+
+We use [asv] to benchmark some core DataLad functionality.
+The benchmarks suite is located under [benchmarks/](./benchmarks), and
+periodically we publish results of running benchmarks on a dedicated host
+to http://datalad.github.io/datalad/ .  Those results are collected
```
yarikoptic: @mih - here is the pointer to our dashboard.  but may be we should add a badge (always green ;-)) to the main README and here pointing to the dashboard

yarikoptic:
@mih that whatever you pushed is not "Results" -- you just pushed not needed changes to benchmark.json (those paths etc) and a record for your machine and otherwise that directory is empty.  i guess you didn't run `asv run` with all the virtualenv setups etc, or didn't `datalad add` later produced results files.
but once again --not sure if that is something we should all do

yarikoptic:
re 3.4 -> 3.5 -- good point. I guess I will change that and rerun on drogon,

yarikoptic:
eh, there were no 3.5 on jessie in the environment I am using there, would need to build a new singularity container based on stretch later on. for now I will even consider this done and merge -- we could move stuff later and improve. smth better than nothing ;-)

PR #2412, opened 2018-04-11
wtf should report extensions and other meaningful entrypoints

mih:


datalad/plugin/wtf.py
```diff
@@ -76,14 +76,24 @@ def __call__(dataset=None, sensitive=None, clipboard=None):
...
         import json
 
+        extractors={}
+        for ep in iter_entry_points('datalad.metadata.extractors'):
+            status = 'BROKEN'
+            try:
+                ep.load()
+                status = 'OK'
+            except:
+                pass
```
yarikoptic: would have been a good spot to use `exc_str` imho to not leave just noninformative BROKEN

datalad/plugin/wtf.py
```diff
@@ -187,7 +201,8 @@ def _t2s(t):
                      if ds_meta else '[no metadata]'
             ),
             externals=external_versions.dumps(preamble=None, indent='', query=True),
-            metaextractors='\n'.join(p for p in dir(metaextractors) if not p.startswith('_')),
+            extensions='\n'.join(ep.name for ep in iter_entry_points('datalad.extensions')),
+            metaextractors='\n'.join('{}: {}'.format(k, v) for k, v in extractors.items()),
```
yarikoptic: use `format_dict(extractors)` for uniform appearance/handling?

datalad/plugin/wtf.py
mih: Done

datalad/plugin/wtf.py
mih: Done

mih:
Still a massive timeout issue in one test run: https://travis-ci.org/datalad/datalad/jobs/365441895

Which cannot possibly be related to this change.

Issue #2405, opened 2018-04-10
FOI: impact of changes in master on startup time of datalad commands


yarikoptic:
#### What is the problem?
 x3-5 time slower for the same command(s):

# install --help
```shell

$> for dld in datalad{,-master,-neuroimaging}; do dl=$dld/venvs/dev/bin/datalad; echo $dl; time $dl install --help | wc -l; done                                                                                           
datalad/venvs/dev/bin/datalad
78
$dl install --help  0.25s user 0.05s system 104% cpu 0.297 total
wc -l  0.00s user 0.00s system 0% cpu 0.297 total
datalad-master/venvs/dev/bin/datalad
78
$dl install --help  0.61s user 0.09s system 100% cpu 0.698 total
wc -l  0.00s user 0.00s system 0% cpu 0.698 total
datalad-neuroimaging/venvs/dev/bin/datalad
78
$dl install --help  0.83s user 0.12s system 99% cpu 0.942 total
wc -l  0.00s user 0.00s system 0% cpu 0.942 total
```

difference washes out a bit while using a real command but still there and might be quite noticeable:

```shell
$> for dld in datalad{,-master,-neuroimaging}; do dl=$dld/venvs/dev/bin/datalad; echo $dl; rm -rf /tmp/123; time $dl install /// /tmp/123 1>/dev/null 2>&1 ; done
datalad/venvs/dev/bin/datalad
$dl install /// /tmp/123 > /dev/null 2>&1  0.31s user 0.09s system 102% cpu 0.390 total
datalad-master/venvs/dev/bin/datalad
$dl install /// /tmp/123 > /dev/null 2>&1  0.50s user 0.12s system 94% cpu 0.651 total
datalad-neuroimaging/venvs/dev/bin/datalad
$dl install /// /tmp/123 > /dev/null 2>&1  0.54s user 0.13s system 76% cpu 0.873 total
```

And for the mighty overall help:

```shell
$> for dld in datalad{,-master,-neuroimaging}; do dl=$dld/venvs/dev/bin/datalad; echo $dl; time $dl --help-np | wc -l; done                                                                                                
datalad/venvs/dev/bin/datalad
167
$dl --help-np  0.26s user 0.06s system 104% cpu 0.314 total
wc -l  0.00s user 0.00s system 0% cpu 0.314 total
datalad-master/venvs/dev/bin/datalad
175
$dl --help-np  0.72s user 0.11s system 100% cpu 0.825 total
[...]

PR #2414, opened 2018-04-12
OPT: delay import of mock, which is heavy (300ms)

yarikoptic:
Identified observing http://datalad.github.io/datalad/#core.startup.time_help_np

This pull request helps #2405

Actually on local runs of the `asv run -E existing -b core.startup.time_help_np` I didn't spot any improvement. So may be it is not there ;-)  But I know that we did move away import of mock into the code blocks everywhere but in tests, so should be the right thing to do

Ideally we should stop mocking within codebase anyways and find better ways ;-)

### Changes
- [x] This change is complete

### DIFF
--- a/datalad/plugin/addurls.py
+++ b/datalad/plugin/addurls.py
@@ -17,8 +17,6 @@ import os
 import re
 import string
 
-from mock import patch
-
 from six import string_types
 from six.moves.urllib.parse import urlparse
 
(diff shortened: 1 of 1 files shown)


kyleam:
>  But I know that we did move away import of mock into the code blocks everywhere but in tests, so should be the right thing to do

OK, sure.  Sorry about that.

> Ideally we should stop mocking within codebase anyways and find better ways

In this case, it's just used as an alternative to `try...finally` or creating context manager (and at [your request][0]).  I'd be fine switching to either of those.

[0]: https://github.com/datalad/datalad/pull/2352#discussion_r177703538

yarikoptic:
well -- that deep it wouldn't matter now I guess. but overall I feel that to just avoid that heavy import of mock (might be worth researching why) we should indeed restrain from using it in the main code base at all

PR #1597, opened 2018-04-12
`add()` will put .gitignore, .gitattributes, etc in annex unless explicitly instructed not to

mih:
This is no case in which this would be useful. We must prevent it.

### DIFF
--- a/datalad/distribution/create.py
+++ b/datalad/distribution/create.py
@@ -314,6 +314,9 @@ class Create(Interface):
             yield path
             return
 
+        # stuff that we create and want to have tracked with git (not annex)
+        add_to_git = []
+
         if no_annex:
             lgr.info("Creating a new git repo at %s", tbds.path)
             GitRepo(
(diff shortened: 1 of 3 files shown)


mih:
Now that the cards are on the table I think it is pretty clear that this is a conceptual problem with more than one technical solution. Main questions is:

When a user says `datalad add .gitignore`, do we want to put `.gitignore` in the annex -- knowing that it produces a most likely undesired situation, or do we not want to do that?

I haven't investigated the issue in in the V6 context where our current concept might be problematic. When saying something like `datalad -d . add .` we generally perform a diff internally and resolve `.` into the actually "modified/new" pieces. This might lead to a changed `.gitsomething` to get listed explicitly in the resulting call to git annex....

@yarikoptic @bpoldrack @kyleam

yarikoptic:
May be we should prepopulate .gitattributes of new datasets with all desired default "exclusions" (e.g. `.git*`) from annex?

mih:
Like a dataset configuration template? ;-)

But yeah, good and practical approach IMHO.

PR #2043, opened 2018-04-17
`save` update, fix bug, test intended behavior, improve docs

mih:
#### What is the problem?
well -- we have `add` to add stuff, which also `saves`.  git allows for a convenient "commit ." which saves all changes in the current sub directory, so I thought that it would only make sense to call "datalad save -m msg ."... to my surprise it added all kinds of stuff which it should have not!

```shell
hopa:~
$> datalad create /tmp/repro
[INFO   ] Creating a new annex repo at /tmp/repro 
create(ok): /tmp/repro (dataset)                                                 
(dev3) 2 13324.....................................:Mon 18 Dec 2017 09:47:55 PM EST:.
hopa:~
$> cd /tmp/repro
(dev3) 2 13325.....................................:Mon 18 Dec 2017 09:47:56 PM EST:.
(git-annex)hopa:/tmp/repro[master]
$> touch 1 2 3
(dev3) 2 13326.....................................:Mon 18 Dec 2017 09:48:03 PM EST:.
(git-annex)hopa:/tmp/repro[master]
$> datalad add -m added\ 1 1
add(ok): /tmp/repro/1 (file)00:00, ?B/s]                                         
save(ok): /tmp/repro (dataset)
action summary:
  add (ok: 1)
  save (ok: 1)
(dev3) 2 13327.....................................:Mon 18 Dec 2017 09:48:16 PM EST:.
(git-annex)hopa:/tmp/repro[master]
$> rm 1; echo 2>1

(dev3) 2 13328.....................................:Mon 18 Dec 2017 09:48:28 PM EST:.
(git-annex)hopa:/tmp/repro[master]
$> ls -l
total 0
-rw------- 1 yoh yoh 0 Dec 18 21:48 1
-rw------- 1 yoh yoh 0 Dec 18 21:48 2
-rw------- 1 yoh yoh 0 Dec 18 21:48 3
(dev3) 2 13329.....................................:Mon 18 Dec 2017 09:48:31 PM EST:.
(git-annex)hopa:/tmp/repro[master]
$> datalad save -m "123" .
save(ok): /tmp/repro (dataset)                                                   
(dev3) 2 13330.....................................:Mon 18 Dec 2017 09:48:40 PM EST:.
(git-annex)hopa:/tmp/repro[master]
$> git show --stat
commit 58562e29e3f3469f5e8b7b3f176dafe8044d6ad0 (HEAD -> master)
Date:   Mon Dec 18 21:48:40 2017 -0500

    123

 2 | 1 +
 3 | 1 +
(dev3) 2 13331.....................................:Mon 18 Dec 2017 09:48:43 PM EST:.
(git-annex)hopa:/tmp/repro[master]
$> ls -l 
total 12
lrwxrwxrwx 1 yoh yoh 108 Dec 18 21:48 1 -> .git/annex/objects/2W/kW/MD5E-s0--d41d8cd98f00b204e9800998ecf8427e/MD5E-s0--d41d8cd98f00b204e9800998ecf8427e
lrwxrwxrwx 1 yoh yoh 108 Dec 18 21:48 2 -> .git/annex/objects/2W/kW/MD5E-s0--d41d8cd98f00b204e9800998ecf8427e/MD5E-s0--d41d8cd98f00b204e9800998ecf8427e
lrwxrwxrwx 1 yoh yoh 108 Dec 18 21:48 3 -> .git/annex/objects/2W/kW/MD5E-s0--d41d8cd98f00b204e9800998ecf8427e/MD5E-s0--d41d8cd98f00b204e9800998ecf8427e
```
current master: 0.9.1-521-gf16f652d

### DIFF
--- a/datalad/interface/save.py
+++ b/datalad/interface/save.py
@@ -107,10 +107,12 @@ def save_dataset(
                    lexists(ap['path'])]
 
     if to_gitadd or save_entire_ds:
+        lgr.debug('Adding files straight to Git at %s: %s', ds, to_gitadd)
         ds.repo.add(to_gitadd, git=True, commit=False,
                     # this makes sure that pending submodule updates are added too
                     update=save_entire_ds)
     if to_annexadd:
+        lgr.debug('Adding files to annex at %s: %s', ds, to_annexadd)
         ds.repo.add(to_annexadd, commit=False)
 
     _datalad_msg = False
@@ -217,7 +219,9 @@ class Save(Interface):
 
         if not dataset and not path:
             # we got nothing at all -> save what is staged in the repo in "this" directory?
-            path = abspath(curdir)
+            # make sure we don't treat this as a user-provided '.' argument
+            path = [{'path': abspath(curdir), 'raw_input': False}]
+
         refds_path = Interface.get_refds_path(dataset)
 
         if message and message_file:
--- a/datalad/interface/save.py
+++ b/datalad/interface/save.py
@@ -148,25 +148,33 @@ def save_dataset(
 class Save(Interface):
     """Save the current state of a dataset
 
-    Saving the state of a dataset records all changes that have been made
-    to it. This change record is annotated with a user-provided description.
+    Saving the state of a dataset records changes that have been made to it.
+    This change record is annotated with a user-provided description.
     Optionally, an additional tag, such as a version, can be assigned to the
-    saved state. Such tag enables straightforward retrieval of past versions
-    at a later point in time.
+    saved state. Such tag enables straightforward retrieval of past versions at
+    a later point in time.
 
-    || PYTHON >>
-    Returns
-    -------
-    commit or None
-      `None` if nothing was saved, the resulting commit otherwise.
-    << PYTHON ||
+    Examples:
+
+      Save any content underneath the current directory, without altering
+      any potential subdataset (use --recursive for that)::
+
+        % datalad save .
+
+      Save any modification of known dataset content, but leave untracked
+      files (e.g. temporary files) untouched::
+
+        % dataset save -d <path_to_dataset>
+
+      Tag the most recent saved state of a dataset::
+
+        % dataset save -d <path_to_dataset> --version-tag bestyet
     """
 
     _params_ = dict(
         dataset=Parameter(
             args=("-d", "--dataset"),
-            doc=""""specify the dataset to save. If a dataset is given, but
-            no `files`, the entire dataset will be saved.""",
+            doc=""""specify the dataset to save""",
             constraints=EnsureDataset() | EnsureNone()),
         path=Parameter(
             args=("path",),
@@ -181,6 +189,8 @@ class Save(Interface):
             doc="""take the commit message from this file. This flag is
             mutually exclusive with -m.""",
             constraints=EnsureStr() | EnsureNone()),
+        # switch not functional from cmdline: default True, action=store_true
+        # TODO remove from API? all_updated=False is not used anywhere in the codebase
         all_updated=Parameter(
             args=("-u", "--all-updated"),
             doc="""if no explicit paths are given, save changes of all known
--- a/datalad/interface/save.py
+++ b/datalad/interface/save.py
@@ -181,11 +181,6 @@ class Save(Interface):
             doc="""take the commit message from this file. This flag is
             mutually exclusive with -m.""",
             constraints=EnsureStr() | EnsureNone()),
-        all_changes=Parameter(
-            args=("-a", "--all-changes"),
-            doc="""save all changes (even to not yet added files) of all components
-            in datasets that contain any of the given paths [DEPRECATED!].""",
-            action="store_true"),
         all_updated=Parameter(
             args=("-u", "--all-updated"),
             doc="""if no explicit paths are given, save changes of all known
@@ -205,18 +200,10 @@ class Save(Interface):
     @datasetmethod(name='save')
     @eval_results
     def __call__(message=None, path=None, dataset=None,
-                 all_updated=True, all_changes=None, version_tag=None,
+                 all_updated=True, version_tag=None,
                  recursive=False, recursion_limit=None, super_datasets=False,
                  message_file=None
                  ):
-        if all_changes is not None:
-            from datalad.support.exceptions import DeprecatedError
-            raise DeprecatedError(
-                new="all_updated option where fits and/or datalad add",
-                version="0.5.0",
-                msg="RF: all_changes option passed to the save"
-            )
-
         if not dataset and not path:
             # we got nothing at all -> save what is staged in the repo in "this" directory?
             # make sure we don't treat this as a user-provided '.' argument
--- a/datalad/interface/tests/test_save.py
+++ b/datalad/interface/tests/test_save.py
@@ -422,6 +422,8 @@ def test_bf1886(path):
     '2': '',
     '3': ''})
 def test_gh2043p1(path):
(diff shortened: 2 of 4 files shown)


mih:
I pushed a fixed for the second part of the problem. Now it needs a test that also covers the original case.

mih:
@bpoldrack is there any way to have a bit of test code to say "this should be untracked, this should not" without having to care what mode a repo is in? If I leave out 'staged' it fails in v5, if I put it in it fails in direct mode

https://travis-ci.org/datalad/datalad/jobs/367594201

I guess I go with the fashion and disable any test for direct mode...

bpoldrack:
It's supposed to work the way, you did it. What seems buggy to me actually is the failing in V5. Looking into it ...

mih:
I will now detach from this issue, as I do not feel confident deciding what should be happening. The core issue is that `datalad save .` commits untracked content. The OP thinks this is not OK, presumably because a `git commit .` wouldn't do it.  However, `save` is not just `commit`, as a quick demo (modified from above) shows -- the key is the last call:

```
(datalad3-dev) mih@meiner /tmp % datalad create /tmp/repro
[INFO   ] Creating a new annex repo at /tmp/repro 
create(ok): /tmp/repro (dataset)                                                             
(datalad3-dev) mih@meiner /tmp % cd /tmp/repro
(datalad3-dev) mih@meiner /tmp/repro (git)-[master] % touch 1 2 3
(datalad3-dev) mih@meiner /tmp/repro (git)-[master] % datalad add -m added\ 1 1
add(ok): /tmp/repro/1 (file)              
save(ok): /tmp/repro (dataset)
action summary:
  add (ok: 1)
  save (ok: 1)
(datalad3-dev) mih@meiner /tmp/repro (git)-[master] % rm 1; echo 2>1



(datalad3-dev) mih@meiner /tmp/repro (git)-[master] % git st
On branch master

        typechange: 1

Untracked files:

        2
        3

(datalad3-dev) mih@meiner /tmp/repro (git)-[master] % git commit .
git-annex: Cannot make a partial commit with unlocked annexed files. You should `git annex add` the files you want to commit, and then run git commit.
```

Hence we have to call `add`, but given the input of `path='.'` we cannot simply call `add('.')` this is the current behavior reported as a bug. In order to emulate `git commit` as much as possible, we would have to `diff` against HEAD and `add` anything that it reports (possibly discovering a slew of corner cases, and maybe even the reason why annex makes `git commit .` behave this way).

At the same time we have

```
  -u, --all-updated     if no explicit paths are given, save changes of all
                        known components in a datasets. [Default: True]
  -a, --all-changes     save all changes (even to not yet added files) of all
                        components in datasets that contain any of the given
                        paths [DEPRECATED!]. [Default: None]
```

(confusing docs of `-u` wrt to the default BTW). Presumably, `-a` maps to `path='.'` plus no `diff` constraint (current behavior). Not sure what `all_updated=False, path='.'` should do? Maybe it is the same as `-a`, but if so, why even support it when we have `add('.')`. ATM it cannot even be used from the cmdline -- possibly an indication that is should be removed altogether.

@bpoldrack @yarikoptic @kyleam What do you think?

bpoldrack:
Re direct mode:
It's actually not about "this should be untracked, this should not", but about "this should be staged", which is an invalid concept in direct mode.
If the assertion really is about what's untracked, just don't specify what is supposed to be staged, but only what should (still) be untracked. So, leaving it out, would be the way to go, but: we have comparable states for direct/indirect mode only AFTER commit. Before that, they are simply conceptionally different.

mih:
> But there is another way. You can use ok_file_under_git with annexed=True to confirm it was added, regardless of mode.

So maybe that is what `ok_clean_git()` should do for `staged=` in direct mode? Because it make little sense to be to project this logic into any test that uses `ok_clean_git()` in this way. Which would also mean that we should find a better name for `staged` that captures both states in V5 and direct mode. Maybe `known_to_git`?

bpoldrack:
Re second to last:
I said AFTER commit. You call it on the uncommitted/unsaved thing. ;-)

Re last comment:
Agree. We can do it that way and we then should also agree to not have arguments for `ok_clean_git`, that are called "head_modified", "index_modified" as I said long ago. ;-)
May be just "known" (to the repo) to not confuse it with helpers/parameters that use "to_git" to distinguish things known to git and known to annex.

mih:
> I said AFTER commit. You call it on the uncommitted/unsaved thing.
They call it testing the test pre-condition ;-) So you are basically saying, don't test the starting point. I don't think that counts as a solution...

kyleam:
Re: `datalad save .` expectations, mine match what @bpoldrack describes in his last comment: save everything, including untracked.  Perhaps that's because I've never used `git commit .`.

mih:
Thanks for your input. I do see these points, too.

However, I feel like this is too muddy on a conceptual basis. Two things:

1. For the purpose of the 0.10 release, I think the current behavior should be documented in a dedicated test (the currently broken one with minimal adjustment would be fine), and in the docs of `save` -- with no further change to the actual behavior of `save` (and #2426 needs fixing, of course).

2. For the next iteration, I'd like to do an assessment of the respective roles of `add` and `save`. Their functionality is massively overlapping with surprisingly different code paths. User-facing differences are: `add` has `--to-git`, while `save` has `--version-tag`, but that is about it.

My hunch is that we should merge the two into a single `save` command that largely behaves like the current `add(save=True)`, but with the remaining features of `save`, namely: `--version-tag` and `--all-updated=True`. Power-users always have `git add` and `git annex add` at their disposal to achieve everything that `save` might be to blunt of an instrument for -- although I cannot yet see what would be missing...

mih:
Conclusion now implemented as code. Let's see what the tests say...

bpoldrack:
> My hunch is that we should merge the two into a single save command that largely behaves like the current add(save=True), but with the remaining features of save, namely: --version-tag and --all-updated=True. Power-users always have git add and git annex add at their disposal to achieve everything that save might be to blunt of an instrument for

Sounds reasonable to me.

Re V6:
I think the failure actually is "correct", since in V6 the test results in something different. This is because in V6 the state of the file (unlocked/locked) is committed as well, and the test doesn't lock the file again. The assertion passes, if you call `git annex lock 1` before the call to `save`.

bpoldrack:
Since this test is supposed to test the behaviour of `save`, I think in this case it's really the test itself, that needs to consider V6 and in this case additionally lock the file. It shouldn't neither be `save` that behaves differently here nor `ok_clean_git`.

mih:
> The assertion passes, if you call git annex lock 1 before the call to save.

Which conveniently skips over the fact that we do not even have a function in the API that does that ;-)

Anyways, I found in the file a rather convenient paradigm to deal with this: adhoc functional call decorators:

```
known_failure_v6(ok_clean_git)(ds.path, untracked=['2', '3'])
```

I'll try to use this more often, instead of disabling entire tests.

bpoldrack:
> Which conveniently skips over the fact that we do not even have a function in the API that does that ;-)

:-) I know, it's available on `AnnexRepo` level, but not on `Dataset` level.

However, skipping just that assertions might do for now.

bpoldrack:
Ah - hold on. This "pattern" was used in generic tests (which yield test functions to be discovered by nose). And in those tests it's fine. But: The decorator raises `SkipTest`, which I think means, that whatever comes after that call within your test, will not be executed.

bpoldrack:
I'd suggest to change the decorator, so it can optionally be used the way, you want to use it (Just passing and not considering a potential failure in the decorated function).
Agree to that approach?

PR #2083, opened 2018-04-22
provide basic appveyor setup

mih:
Summary after a long time:

This should be merged ASAP, because we are going backwards on windows and we are not aware of it.

All this does (besides few fixes) is do a basic installation on windows and run `wtf` to see if anything works.

Future changes should slowly introduce stuff that out to work to the
    tests

----

and now the stuff from January....

#### What is the problem?
With all the annotations of tests which fail in direct and v6 modes, may be it is feasible now to approach having testing done on windows as well, and I think appveyor would be a good free service for that purpose ;)  It could help to catch issues "spawning" from windows origins, such as the one reported in https://github.com/datalad/datalad/pull/2066

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,6 +8,10 @@ environment:
       MINICONDA: C:\Miniconda35
       DATALAD_TESTS_SSH: 0
 
+cache:
+  # cache the pip cache
+  - C:\Users\appveyor\AppData\Local\pip\Cache -> appveyor.yml
+
 init:
   - "ECHO %PYTHON% %PYTHON_VERSION% %PYTHON_ARCH% %MINICONDA%"
   # this will display login RDP info for the build VM, but if the build VM should block
@@ -30,7 +34,7 @@ install:
   - python -c "import sys; print(sys.path)"
   # when something works we can go full
   #- pip install -e ".[full]"
-  - pip install "."
+  - pip install ".[tests]"
   # fixup
   # ATM datalad does not pull in colorama, which is needed for color output
   # on windows
@@ -47,8 +51,10 @@ test_script:
   - git annex version
   # first sign of life
   - datalad wtf
-  # and now this...
-  #- python -m nose -s -v datalad.distribution.tests.test_create
+  # and now this... [keep appending tests that should work!!]
+  # this is not good!
+  # - python -m nose -s -v datalad.plugin
+  - python -m nose -s -v datalad.cmdline datalad.tests.test_api datalad.tests.test_base datalad.ui
 
 #after_test:
 #  - bash <(curl -s https://codecov.io/bash)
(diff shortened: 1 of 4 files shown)


mih:
https://www.appveyor.com/docs/build-environment/ looks like most essential bits other than Python packages are already available, incl. Git. This pretty much leaves git-annex itself to be deployed.

mih:
RDP login works fine with a line like this (server and passwd are random). We could configure a passwd via the appveyor project settings.

`% xfreerdp /v:74.205.54.20:34383 /u:appveyor /p:'EL2v(#BxB3L*x-#'`

So far I failed to figure out how to install git-annex in an unsupervised way -- although the NullSoft Installer should, in principle, support "silent installs".

Installing manually via the Nullsoft installer + RDP makes it complain that it doesn't find an installed Git -- although Git is installed and functional.

mih:
One approach to installing git-annex might be to bypass the entire installation procedure and decompile the install exe with 7zip (preinstalled) and just move the pieces where they need to go...

mih:
We can now install datalad and git-annex. Running some tests is next.

mih:
I think the basic infrastructure is in place now. Login for debug is easy and reliable.

Now we need some tests that actually run under windows....

mih:
@joeyh Is the annex test suite expected to pass on windows, given the technical limitation wrt *nix? If yes, could you please have a quick look at the results of `git annex test` on appveyor (windows):

```
102 out of 165 tests failed (175.29s)
(Failures above could be due to a bug in git-annex, or an incompatibility
```

Here is a full log: https://ci.appveyor.com/project/mih/datalad/build/1.0.11

Installing git-annex in this environment is a bit of a hack (see appveyor.yml diff in this PR), I basically extract the installer into the Git install dir -- but it seems to work, at least superficially.

joeyh:
Michael Hanke wrote:
> @joeyh Is the annex test suite expected to pass on windows, given the technical
> limitation wrt *nix? If yes, could you please have a quick look at the results
> of git annex test on appveyor (windows):

The git-annex test suite often fails on windows, for unknown or
occasionally known reasons.

-- 
see shy jo

mih:
Now stuff is running:

```
Ran 869 tests in 838.946s
FAILED (SKIP=95, errors=387, failures=20)
```

mih:
This is all that was asked for: "provide basic appveyor setup"

Ready to merge.

mih:
```
Ran 872 tests in 485.865s
FAILED (SKIP=98, errors=380, failures=20)
```

mih:
@yarikoptic @kyleam UNless you have strong objections, this should be merged ASAP.

yarikoptic:
+1 -- we just need to wait for travis to finish to make sure no changes break it

mih:
@yarikoptic  Travis is not cooperative right now -- Ubuntu downloads fail...

mih:
Direct mode check broken due to #2523 . Rerun when merged.

PR #2437, opened 2018-04-22
RF: crawler into its own extension

mih:
Notes:
- [x] crawler extension with the code that is removed here https://github.com/datalad/datalad-crawler

some tests under crawler/ tests pieces that are not limited to crawler functionality, e.g. downloaders -- they would have to move and stay close to the actual implementations of that functionality. But see #2480


mih:
@yarikoptic Extension is ready. RC1 release is on pypi (you are owner). Docs are built automatically. Can be released any time.

mih:
I added a link to the docs from the main doc site. Planning to merge this later today.

PR #2446, opened 2018-04-25
ENH: Let `run` spit out the original full-detail error

mih:


mih:
Thanks!

I agree with your assessment of the negative impact on visual beauty -- let's fix that in the exception rendering...

kyleam:
> let's fix that in the exception rendering...

Yeah, here's probably where to do that:

https://github.com/datalad/datalad/blob/36767377e30a94152fb2e1fb8871345b4492d8d0/datalad/cmd.py#L518-L519

I'll look at it at some point, but you're of course welcome to beat me to it.

PR #2463, opened 2018-04-30
coverage for (all) plugins code is reported at 0%

yarikoptic:
#### What is the problem?
Initially (or not for someone else) was spotted in #2382 that even though tests run/fail, coverage is reported at 0%:
```
datalad/plugin/__init__.py                                   27      7    74%
datalad/plugin/add_readme.py                                 61     61     0%
datalad/plugin/addurls.py                                   332    332     0%
datalad/plugin/check_dates.py                                72     72     0%
datalad/plugin/export_archive.py                             67     67     0%
datalad/plugin/export_to_figshare.py                        137    137     0%
datalad/plugin/extract_metadata.py                           38     38     0%
datalad/plugin/no_annex.py                                   49     49     0%
datalad/plugin/wtf.py                                        94     94     0%
```
Testing with coverage locally though reports correctly.  Wild guesses about possible causes:
- ~~Out of codebase testing somehow picks up plugin files within original (under git) sources? (should be easy to test locally)~~
- ~~some outdated coverage etc which doesn't work nicely with our way to import/load plugins~~
- Our (my) fault ;)

### DIFF
--- a/datalad/utils.py
+++ b/datalad/utils.py
@@ -1529,12 +1529,12 @@ def read_csv_lines(fname, dialect=None, readahead=16384, **kwargs):
                 yield dict(zip(header, row_unicode))
 
 
-def import_modules(mods, pkg, msg="Failed to import {module}", log=lgr.debug):
+def import_modules(modnames, pkg, msg="Failed to import {module}", log=lgr.debug):
     """Helper to import a list of modules without failing if N/A
 
     Parameters
     ----------
-    mods: list of str
+    modnames: list of str
       List of module names to import
     pkg: str
       Package under which to import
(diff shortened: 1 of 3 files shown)


yarikoptic:
pushed a fix which might work, pushed so tests run and we see if anything breaks.  In our tests -- do we have a test to check for external plugins load?

datalad/utils.py
```diff
@@ -1546,56 +1546,66 @@ def import_modules(mods, pkg, msg="Failed to import {module}", log=lgr.debug):
...
-    for mod in mods:
+    mods_loaded = []
+    for modname in modnames:
         try:
-            _globals[mod] = import_module(
-                '.{}'.format(mod),
+            _globals[modname] = mod = import_module(
+                '.{}'.format(modname),
                 pkg)
+            mods_loaded.append(mod)
```
yarikoptic: ha ha -- apparently `import_modules` is not used anywhere and the test only tests for "bogus" case thus this line is never reached.  I will test it a bit better while preparing the test just so we have it available happen we need it

datalad/utils.py
```diff
@@ -1546,56 +1546,66 @@ def import_modules(mods, pkg, msg="Failed to import {module}", log=lgr.debug):
...
-            datalad_subpath = "datalad.plugin"  # e.g.
-            mod = __import__(datalad_subpath + '.%s' % modname,
-                             fromlist=[datalad_subpath])
+        if relmodpath:
+            from importlib import import_module
+            mod = import_module(relmodpath, pkg.__name__)
         else:
-            mod = __import__(modname, level=0)
-        return mod
+            dirname_ = dirname(modpath)
```
yarikoptic: hm, no code (smoke)tests loading by the path without specifying pkg :-/

yarikoptic:
buildbots do not pick it up somehow but travis is all green and the whooping 1% boost in coverage!, so needs best merge

mih:
Nice. LGTM. Tests are still running...

Issue #1778, opened 2017-08-31
move casts helpers into a separate repo/project


yarikoptic:
since could be of utility to others, and we could benefit from wider adoption/contributions

proposed names from top of the head
- autocast (bad, collides with many other projects)
- CastLad ;-)

mih:
And also the cast scripts itself ( not in the same repo, of course), but we now have website, docs of multiple extensions + core, all doing different thing. Maybe we should have a datalad/demos repo as a common place for all those.

mih:
@adswa What do you think about this? You are the only one I now, who is (or was) actively using this over the past years.

adswa:
I like them! I use the cast live tool in a modified version for the course repo - its cool, even though it involves a bit of fiddling sometimes.
A definite benefit of using a separate repo would be increased findability and more space for proper documentation for them. They certainly have value beyond datalad, and detaching them from datalad make them more attractive. Given that they aren't used for generating anything the docs (as far as I know), I'd vote for moving them into their own repo.

adswa:
If someone with permissions to do so creates a new repository under datalad for this purpose I could migrate the cast helpers.

mih:
I created https://github.com/datalad/screencaster

adswa:
Thanks. I have migrated all ``tools/cast*`` files into this new repo and preserved their Git history: https://github.com/datalad/screencaster. Their documentation is quite sparse, if someone wants to improve the README, please go for it. I declared MIT license.

PR #2476, opened 2018-05-05
ENH: adds python cast example

miykael:
I've created a python `cast` example, based on the [`nipype_tutorial` showcase](https://miykael.github.io/nipype_tutorial/notebooks/introduction_showcase.html). Perhaps this is useful for some?

I'm not sure if the `rm -rf ~/demo` command in [cast_ipython.rc](https://github.com/miykael/datalad/blob/1d14d4b2057c7bdf76eaecf6f7412f0458cab029/tools/cast_ipython.rc#L16) is a horrible abomination and perhaps should be changed.

### DIFF
--- a/docs/casts/nipype_tutorial_intro.sh
+++ b/docs/casts/nipype_tutorial_intro.sh
@@ -1,0 +1,83 @@
+say "Import nipype building blocks"
+show "Import nipype building blocks"
+run "from nipype import Node, Workflow"
+
+say "Import relevant interfaces"
+show "Import relevant interfaces"
+run "from nipype.interfaces.fsl import SliceTimer, MCFLIRT, Smooth"
+
+say "Create SliceTime correction node"
+show "Create SliceTime correction node"
+run "slicetimer = Node(SliceTimer(index_dir=False,
+                             interleaved=True,
+                             time_repetition=2.5),
+                  name='slicetimer')
+"
+
+say "Create Motion correction node"
+show "Create Motion correction node"
+run "mcflirt = Node(MCFLIRT(mean_vol=True,
+                       save_plots=True),
+               name='mcflirt')
+"
+
+show "Create Smoothing node"
+run "smooth = Node(Smooth(fwhm=4), name='smooth')"
+
+show "Create Workflow"
+run "preproc01 = Workflow(name='preproc01', base_dir='.')"
+
+show "Connect nodes within the workflow"
+run "preproc01.connect([(slicetimer, mcflirt, [('slice_time_corrected_file', 'in_file')]),
+                   (mcflirt, smooth, [('out_file', 'in_file')])])
+"
+
+show "Create a visualization of the workflow"
+run "preproc01.write_graph(graph2use='orig')"
+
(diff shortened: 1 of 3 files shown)


tools/cast_ipython.rc
```diff
@@ -0,0 +1,16 @@
+# This file contains ipython configuration variables to be used for generating
+# asciinema demos to guarantee consistent appearance.
+
+# make a fake temporary home dir and go into it
+SCREENCAST_HOME=~/demo
```
yarikoptic: use
```bash
SCREENCAST_HOME=${SCREENCAST_HOME:-/demo}
```
as we have in cast_bash.rc?  may be even just source `cast_bash.rc` to get also git configured etc?

tools/cast_ipython.rc
```diff
@@ -0,0 +1,16 @@
...
+    mkdir -p ${SCREENCAST_HOME} || {
+        echo "FAILED to create $SCREENCAST_HOME" >&2
+        exit 1;  # we need demo directory!
+    }
+fi
+cd $SCREENCAST_HOME
+ipython
+
+# cleanup at the end
+trap "cd ; rm -rf ~/demo > /dev/null 2>&1" EXIT
```
yarikoptic: it should be SCREENCAST_HOME not just ~/home.  cast_bash.rc already does it

tools/cast_live_python
```diff
@@ -0,0 +1,115 @@
+#!/bin/bash
+#
+set -u -e
+
+# reduce confusion with xdotool
+setxkbmap us
+
+test ! -e $1 && echo "input file does not exist" && exit 1
+title="$(echo $(basename $1) | sed -e 's/.sh$//')"
+bashrc_file="$(dirname $0)/cast_ipython.rc"
```
yarikoptic: so overall this script seems to be the `cast_live` without much of changes besides a different .rc file and executing the ipython command.  So I feel that 
- the `cast_live_ipython` should merely define a few env variables (`CAST_SHELL=ipython`)
- there might be no need for cast_ipython.rc, ipython could be executed within `cast_live`: `if [ ! -z "$CAST_SHELL" ] ; then $CAST_SHELL; fi` or smth like that .  If it would really be needed (e.g. to setup ipython config) then it could just source `cast_bash.rc` and just complement with additional needed stuff

What other customization need to be done?  could probably also be channeled via environment variables.
The problem with having two largely duplicate scripts is difficulty in maintaining them.

docs/casts/nipype_tutorial_intro.sh
```diff
@@ -0,0 +1,83 @@
+say "Import nipype building blocks"
```
yarikoptic: eh, someone really need to take care about #1778 and then this script should be contributed to nipype -- there is no reason for it to live hidden in datalad codebase

yarikoptic:
let's indeed close for now but feel welcome to chime in more on it if want to take over "casting production" ;)

miykael:
Thanks, will do so! Anyhow, it's an amazing tool that makes teaching so much better!

PR #2490, opened 2018-05-14
BF: handle DownloadErrors as well as CommandErrors while trying to enable special remote

yarikoptic:
Necessary for compatibility with git-annex 6.20180510+gitg500f7ea78-1~ndall+1
which changed error message so we identify it as AccessFailedError

I am not sure if that is the correct solution here but the entire logic is not
clear for me -- i.e. why we would allow it to happen and just issue an info
message instead of failing.  If it is to catch cases where it is a git
repo on the other end -- there should be some analysis of the stderr


yarikoptic:
I think that life will not be worse with this fix, so I will just merge it for now.  But eventually I think someone needs to review this logic of handling errors at this point

PR #2492, opened 2018-05-15
RF: do not issue warnings when progressbar regresses

yarikoptic:
The reason is still unknown but it is not worth bombarding users with those
developer-oriented warnings.  We will just log them at DEBUG level
for ourselves

I thought there was an issue mentioning it but can't find

Also emailed @joeyh asking if such cases (regress in reported progress) is legit and could happen for an ok reason somehow


kyleam:
Makes sense.  The travis failure is due to the known crawler failure.

yarikoptic:
I thought crawler is gone... Will check later

PR #2491, opened 2018-05-15
entry_points in extensions -- not in effect whatsoever? (tests do not run, not available within datalad namespace)

mih:
#### What is the problem?
I am not sure how to run tests for "everything in datalad"... e.g. in below you could see that I am installing datalad-crawler (`datalad crawl` command does become available) and expected that I could just test `datalad` module and automagically tests for `datalad.crawl` would also be executed... but nope:
```shell
$> git clone git://github.com/datalad/datalad-crawler                            
$> cd datalad-crawler 
$> virtualenv --system-site-packages venvs/dev                  
$> source venvs/dev/bin/activate
$> pip install -e .                 
$> which datalad                                                         
/tmp/datalad-crawler/venvs/dev/bin/datalad
$> python -m nose -s -v datalad 2>&1 | grep crawl
# nothing output of relevance to tests within datalad_crawler
```
also now looking into .travis.yml -- seems that the only way is to just invoke them by naming the extension module (i.e. datalad_crawl). 

I saw that extensions (e.g. datalad-crawler) sets up entry_points for setuptools, e.g.:
```python
    entry_points = {
        # 'datalad.extensions' is THE entrypoint inspected by the datalad API builders
        'datalad.extensions': [
            # the label in front of '=' is the command suite label
            # the entrypoint can point to any symbol of any name, as long it is
            # valid datalad interface specification (see demo in this extension)
            'crawler=datalad_crawler:command_suite',
        ],
        'datalad.tests': [
            'crawler=datalad_crawler',
        ],
    },
```
but it seems that none of that in effect since e.g.
```shell
$> python -c 'import datalad.extensions as dl; print(dl.crawler)' 
ImportError: No module named extensions
```
So what is the point behind all those entry_points records then or what am I doing wrong?

### DIFF
--- a/datalad/interface/test.py
+++ b/datalad/interface/test.py
@@ -18,6 +18,9 @@ from datalad.interface.base import build_doc
 from datalad.utils import assure_list
 from ..support.param import Parameter
 
+import logging
+lgr = logging.getLogger('datalad.interface.test')
+
 
 @build_doc
 class Test(Interface):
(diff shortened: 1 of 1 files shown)


mih:
@yarikoptic Please check the `test` command implementation for the code that discovers installed extension tests. I also attached a commit here that actually makes this work from the cmdline (too strict test).

Maye there is a more elegant way to do this, but I am not aware of any. The individual packages are (and should be) self-contained, so we cannot install actual modules into the datalad namespace.

mih:
Travis failure is know crawler issue.

mih:
Let's get this fix in.

PR #2495, opened 2018-05-16
OPT: delay import of pkg_resources until use, adjust log messages to differ

yarikoptic:
loading extensions is duplicating in two places with identical log messages making it difficult to localize that code while tracing the log msgs

delaying import of `pkg_resources` is heavy and should get closer to when actually needed. Eventually we might manage to avoid loading extensions if not needed (command is decided)

### DIFF
--- a/datalad/api.py
+++ b/datalad/api.py
@@ -38,8 +38,13 @@ def _generate_extension_api():
 
     for entry_point in iter_entry_points('datalad.extensions'):
         try:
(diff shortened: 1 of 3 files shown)


yarikoptic:
bloody vim -- I will fix it.  Thank you @kyleam !!!

mih:
Thanks for doing this. Always wanted to get to it, never managed.

PR #2500, opened 2018-05-17
`aggregate-metadata` avoids extraction if already done for the present state

mih:
Instead shuffles around aggregated files much more flexibly.

This also does some RF to improve and consolidate file content bits, and other stuff.

Yeah, not very descriptive, but if it works it is better -- trust me.


datalad/metadata/aggregate.py
```diff
@@ -186,16 +193,130 @@ def _extract_metadata(agginto_ds, aggfrom_ds, db, to_save):
...
         return False
 
-    # if there is any chance for metadata
-    # obtain metadata for dataset and content
+    # shorten to MD5sum
+    objid = md5(objid.encode()).hexdigest()
+
+    # assemble info on the metadata extraction and storage
+    #               label  type      targetds    storage method
+    metasources = {'ds': {'type': 'dataset', 'targetds': agginto_ds, 'dumper': json_py.dump}}
```
mih: Now might be the time to introduce a flag to make this also use compressed+annex by default.

datalad/metadata/aggregate.py
```diff
@@ -186,16 +193,130 @@ def _extract_metadata(agginto_ds, aggfrom_ds, db, to_save):
...
+        # info on identity is sufficient
+        if op.lexists(
+            op.join(
+                d.path,
+                dirname(agginfo_relpath),
+                _get_obj_location(objid, s)))
+    } if not force_extraction else False
+
+    if not metafound:
+        lgr.info('FULLEXTRACT')
```
mih: Kill this.

datalad/metadata/aggregate.py
```diff
@@ -186,16 +193,130 @@ def _extract_metadata(agginto_ds, aggfrom_ds, db, to_save):
...
+            metasources,
+            refcommit,
+            subds_relpaths)
+
+    # we did not actually run an extraction, so we need to
+    # assemble an aggregation record from the existing pieces
+    # that we found
+    # simple case: the target dataset has all the records already:
+    if all(d is agginto_ds for s, d in metafound.items()):
+        lgr.error('UNCHANGED')
```
mih: Kill this.

datalad/metadata/aggregate.py
```diff
@@ -186,16 +193,130 @@ def _extract_metadata(agginto_ds, aggfrom_ds, db, to_save):
...
+        # no change, use old record from the target dataset
+        db[aggfrom_ds.path] = old_agginfo
+        # no error
+        return False
+    else:
+        lgr.info('MVFORMSOURCE')
```
mih: Kill this.

datalad/metadata/aggregate.py
```diff
@@ -186,16 +193,130 @@ def _extract_metadata(agginto_ds, aggfrom_ds, db, to_save):
...
+
+        # actually copy dump files
+        for objrelpath in objrelpaths.values():
+            objpath = op.join(agginto_ds.path, objrelpath)
+            objdir = dirname(objpath)
+            if not exists(objdir):
+                makedirs(objdir)
+            # XXX TODO once we have a command that can copy/move files
+            # from one dataset to another including file availability
+            # info, this should be used here
```
mih: This would be lovely to have. And should not be that hard.

shutil.copy() + annex info reassignment + annex metadata duplication + done

or?

datalad/metadata/aggregate.py
```diff
@@ -213,36 +339,16 @@ def _extract_metadata(agginto_ds, aggfrom_ds, db, to_save):
...
-    for label, mtype, meta, dest, store in metasources:
-        if not meta:
+    for label, props in metasources.items():
+        dest = props['targetds']
+        if not meta[label]:
             continue
         # only write to disk if there is something
         objrelpath = _get_obj_location(objid, label)
-        if store is json_py.dump2xzstream:
+        if props['dumper'] is json_py.dump2xzstream:
```
mih: move this logic into the storage method config (up)?

datalad/metadata/tests/test_aggregation.py
```diff
@@ -340,8 +340,8 @@ def test_partial_aggregation(path):
     # from-scratch aggregation kills datasets that where not listed
     ds.aggregate_metadata(path='sub1', incremental=False)
     res = ds.metadata(get_aggregates=True)
-    assert_result_count(res, 2)
-    assert_result_count(res, 0, path=sub2.path)
+    assert_result_count(res, 3)
+    assert_result_count(res, 1, path=sub2.path)
```
mih: This is a change in behavior. Not sure if good or bad. In general, I don't see any sense in throwing away metadata records because a dataset isn't present locally.

mih:
Even travis cannot replicate the failure sometimes. There is something funny going on here.

datalad/metadata/aggregate.py
```diff
@@ -150,11 +151,14 @@ def _extract_metadata(agginto_ds, aggfrom_ds, db, to_save):
...
-    # 2, our own dataset-global metadata
-    dsmetafile = opj(aggfrom_ds.path, '.datalad', 'metadata', 'dataset.json')
-    if exists(dsmetafile):
-        objid += md5(open(dsmetafile, 'r').read().encode()).hexdigest()
+    # 2, our own dataset-global metadata and the dataset config
+    for tfile in (
+            opj(aggfrom_ds.path, '.datalad', 'metadata', 'dataset.json'),
+            opj(aggfrom_ds.path, '.datalad', 'config')):
+        if exists(tfile):
+            objid += md5(open(tfile, 'r').read().encode()).hexdigest()
```
yarikoptic: is there a need to have both digests concatenated within the objid (longish filename, more to commit to git) or it could be a either hexdigest on a collection of digests (easier), or just hexdigest of concatenated content?

we already have  datalad.utils:md5sum
```
datalad/utils.py-
datalad/utils.py-def md5sum(filename):
datalad/utils.py-    with open(filename, 'rb') as f:
datalad/utils.py:        return hashlib.md5(f.read()).hexdigest()
```

mih: If you skip a few more lines down, I'll see that it finally get's condensed into a single short MD5sum.

mih:
OK, let's run with it and see how it behaves in the wild.

PR #2497, opened 2018-05-18
datacite extractor requires file at .datalad/meta.datacite.xml

mih:
I think this should be underneath `.datalad/metadata/`, or even better, it should search for a file of this name in the list of files it gets to process.


mih:
@yarikoptic said OK.

PR #2511, opened 2018-05-21
BF(TST): do not test full path in annex description

yarikoptic:
since might be shortnened if under $HOME


yarikoptic:
I think this one is nobrainer thus merging

PR #2526, opened 2018-05-22
BF: Don't fail too early when parsing cmdline arguments

### LINKED ISSUE #2524: `run` cmdline API broken
Demo:

```
% datalad run dummy -l
usage: datalad [-l LEVEL] [--pbs-runner {condor}] [-C PATH] [--version]
               [--dbg] [--idbg] [-c KEY=VALUE]
               [-f {default,json,json_pp,tailored,'<template>'}]
               [--report-status {success,failure,ok,notneeded,impossible,error}]
               [--report-type {dataset,file}]
               [--on-failure {ignore,continue,stop}] [--cmd]
               command [command-opts]
```

It should fail on execution, because dummy is not a command. But instead it trips over `-l` and thinks it a datalad option.

I wonder if the recent OPT that hand parses the args to guess commands broke this. I have such calls in my shell history, which backs up my memory of this working very recently.

Also this pattern is used in the heudiconv call here: https://github.com/mih/ohbm2018-training/blob/master/fsl_glm_w_amazing_datalad.sh (which means it ran using some version of datalad 1-2 days ago).

FYI: @bpoldrack

bpoldrack:
This pull request fixes #2524

This pull request proposes to throw away (too) early failing when parsing cmdline arguments.

### Changes
- [ ] This change is complete

### DIFF
--- a/datalad/cmdline/main.py
+++ b/datalad/cmdline/main.py
@@ -205,9 +205,8 @@ def setup_parser(
                   parsed_args, unparsed_args)
     except Exception as exc:
         lgr.debug("Early parsing failed with %s", exc_str(exc))
-        fail_handler(parser)
         need_single_subparser = False
-
+        unparsed_args = cmdlineargs[1:]  # referenced before assignment otherwise
 
     interface_groups = get_interface_groups()
     # TODO: see if we could retain "generator" for plugins


bpoldrack:
I think so, just wanna see the tests.

mih:
0% coverage! Achievement unlocked.

datalad/cmdline/main.py
```diff
@@ -205,9 +205,8 @@ def setup_parser(
                   parsed_args, unparsed_args)
     except Exception as exc:
         lgr.debug("Early parsing failed with %s", exc_str(exc))
-        fail_handler(parser)
         need_single_subparser = False
-
+        unparsed_args = cmdlineargs[1:]  # referenced before assignment otherwise
```
yarikoptic: it is pity there is no unittest which demonstrates the problem and the fix... I am afraid I would break it again at some point in the future.

Academically, it is not clear to me why `parser` went all the way to that `-l` since it didn't even know about `run`, so I expected it to fail at that point already, leaving `-l` etc in unparsed_args.

bpoldrack: Well, academically it is not clear to me, why you use a private method of a class from a foreign package, while not being clear about what exactly it does. :-P ;-)

Issue #2518, opened 2018-05-21
Disable batched annex on windows? (also on inability to rmtree etc due to locked files)


mih:
__del__ handling there is a common failure reason (windows is very attached to open files). It would be useful, if it could be turned off to see whether underlying functionality works, albeit slower
 Is that possible?

yarikoptic:
IIRC there might be some commands where we ATM do not have/use `--json` output for non-batched versions... IIRC @kyleam even had to fix some recently for `run` functionality.  But there might be more

kyleam:
> IIRC there might be some commands where we ATM do not have/use --json output for non-batched versions

Quickly grepping/scanning, I didn't spot any other offenders aside from the one fixed by 0181a70e0 (ENH: annexrepo: Make add_url_to_file always return json output, 2018-04-16).

mih:
I am not following. Does this mean there is a switch, or that soemthing could be RF'ed in some way, but there is no switch?

mih:
;-) Thanks for the translation!

mih:
It seems to me that we should have a switch that turns on/off batched processing, and it should default to off on windows.

yarikoptic:
I thought to suggest the same.  I am afraid though that many operations will literally "crawl" (and not `datalad crawl`) with a per-file invocation of git-annex.  May be it would be worth exploring the initial problem (original description is "too concise" IMHO to deduce anything) and see if it could be addressed without removing the use of batched processes?  I think that git-annex itself uses batched versions of some git commands on windows as well, and I guess it works ok, isn't it @joeyh?

yarikoptic:
I wonder if at least some of those aren't real issues of us (or GitPython) not closing some files before trying to annihilate them/directories.  E.g. on linux this one works "just fine" (too many python opens since was playing around, file could be generated outside):

```shell
$> rm -rf /tmp/test; mkdir /tmp/test; python -c 'f=open("/tmp/test/1", "w"); f.write("1"); f.close(); f2=open("/tmp/test/1"); import shutil; shutil.rmtree("/tmp/test/"); print("READ: %s" % f2.read())' 
READ: 1
```
and I guess should fail on Windows?

mih:
Probably, Q is just how would change the code? Not remove?

yarikoptic:
I can only reply with another Q: who would change/remove what code?

[...]

Issue #2287, opened 2018-03-08
Robustify tests wrt to PY2 vs PY3


mih:
Below is a test that fails in our singularity image, because there is no PY2. Would be could to not require that. @kyleam ?

```
======================================================================
ERROR: datalad.interface.tests.test_run.test_rerun_subdir
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/local/lib/python3.5/dist-packages/datalad/interface/run.py", line 173, in run_command
    expect_fail=True,
  File "/usr/local/lib/python3.5/dist-packages/datalad/cmd.py", line 520, in run
    raise CommandError(str(cmd), msg, status, out[0], out[1])
datalad.support.exceptions.CommandError: CommandError: command 'python -c 'open("test.dat", "wb").close()'' failed with exitcode 127
Failed to run 'python -c \'open("test.dat", "wb").close()\'' under '/tmp/datalad_temp_test_rerun_subdir7r6913gm/subdir'. Exit code=127. out= err=

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.5/dist-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/usr/local/lib/python3.5/dist-packages/datalad/tests/utils.py", line 1290, in newfunc
    func(*args, **kwargs)
  File "/usr/local/lib/python3.5/dist-packages/datalad/tests/utils.py", line 664, in newfunc
    return t(*(arg + (filename,)), **kw)
  File "/usr/local/lib/python3.5/dist-packages/datalad/interface/tests/test_run.py", line 408, in test_rerun_subdir
    run("python -c 'open(\"test.dat\", \"wb\").close()'")
  File "/usr/local/lib/python3.5/dist-packages/wrapt/wrappers.py", line 523, in __call__
    args, kwargs)
  File "/usr/local/lib/python3.5/dist-packages/datalad/interface/utils.py", line 466, in eval_func
    return return_func(generator_func)(*args, **kwargs)
  File "/usr/local/lib/python3.5/dist-packages/wrapt/wrappers.py", line 523, in __call__
    args, kwargs)
  File "/usr/local/lib/python3.5/dist-packages/datalad/interface/utils.py", line 454, in return_func
    results = list(results)
  File "/usr/local/lib/python3.5/dist-packages/datalad/interface/utils.py", line 411, in generator_func
    result_renderer, result_xfm, _result_filter, **_kwargs):
[...]

PR #2542, opened 2018-05-23
NF: DATALAD_ASSERT_NO_OPEN_FILES

yarikoptic:
You can start playing with it using smth like
```shell
$> DATALAD_ASSERT_NO_OPEN_FILES=1 DATALAD_LOG_LEVEL=ERROR python -m nose -s -v datalad/support/tests/test_annexrepo.py                                                                              datalad.support.tests.test_annexrepo.test_AnnexRepo_addurl_batched_and_set_metadata ... ok
datalad.support.tests.test_annexrepo.test_AnnexRepo_addurl_to_file_batched ... SKIP: TODO: more, e.g. add with a custom backend
datalad.support.tests.test_annexrepo.test_AnnexRepo_backend_option ... [ERROR  ] Open files under /home/yoh/.tmp/datalad_temp_tree_test_AnnexRepo_backend_optionfOrqyh: {'/home/yoh/.tmp/datalad_temp_tree_test_AnnexRepo_backend_optionfOrqyh': 29666} 
ok
datalad.support.tests.test_annexrepo.test_AnnexRepo_web...
```

should help to figure out some of the #2518 problems on windows. It might also be of value to resolve those properly for operation on NFS (to avoid all the .nfsXXXXX files breeding)

TODOs:
- [ ] extend record with information about those processes which keep the file/directory open.  I guess some batched process or alike, but which exactly?


yarikoptic:
c224fe8c8362355ea81e5fda2a92520b9c40848f might resolve some (if not many) issues on windows build BUT it might be very suboptimal and should be done differently: 
- we might not want to cleanse GitPython's repo upon every precommit, since it seemed to work fine across commits before ;-)

mih:
Just found this: https://stackoverflow.com/questions/11114492/check-if-a-file-is-not-open-not-used-by-other-process-in-python

Looks like a platform independent way to get the list of open files.

yarikoptic:
Thanks @mih, implemented via psutils... test seems to run faster - if only someone made use of it ;-)

yarikoptic:
I really don't know what conflicts travis is whining about -- origin/master was just merged :-/ may be will get it tomorrow

yarikoptic:
github was whining because it was "Rebase and merge" strategy selected instead of "Merge pull request". Thanks @kyleam for spare eyes

datalad/support/gitrepo.py
```diff
@@ -54,6 +54,7 @@
 from datalad.utils import getpwd
 from datalad.utils import updated
 from datalad.utils import posix_relpath
+from datalad.utils import assert_no_open_files
```
mih: why the import?

datalad/support/archives.py
```diff
@@ -28,7 +28,7 @@
 import random
 
 from .locking import lock_if_check_fails
-from ..utils import any_re_search, assure_bytes
+from ..utils import any_re_search, assure_bytes, rmdir
```
mih: Multiple imports per line are a pain wrt code maintenance IMHO

yarikoptic: I actually started to go the round of "multiple imports per import statement", but I guess not in full...  What do you think about the convention alike
```python
from ..utils import (
    any_re_search,
    assure_bytes,
    rmdir,
)
```
which allows to minimize the diff, and also minimize needed changes while tuning from which module to import ?

datalad/support/gitrepo.py
yarikoptic: I guess left over from me trying to stick it int some "privates" (well -- the `__del__`) of GitRepo.  Now I forgot why I removed it from there, so will keep it at it and remove the import, thanks

datalad/tests/test_utils.py
```diff
@@ -1094,8 +1094,17 @@ def test_get_open_files(p):
 
     assert not get_open_files(subd)
     # if we start a process within that directory, should get informed
-    from subprocess import Popen
-    proc = Popen(['cat'], cwd=subd)
+    from subprocess import Popen, PIPE
+    from time import time
+    t0 = time()
+    proc = Popen(['python', '-c',
```
kyleam: This would fail in the same scenario reported [here](https://github.com/datalad/datalad/issues/2287#issue-303452753) (i.e., no "python", just "python3").

yarikoptic: Eh... Thanks. Then we should use sys.executable may be here and in the procedures magical extension handling... I will fix here

yarikoptic:
uff... fails are the known ones... merging

PR #2548, opened 2018-05-24
 BF: Do not fall for symlinks when looking for subdatasets

mih:


bpoldrack:
Do we want `is_valid_repo` to still claim a symlink is a valid repo? Otherwise this should go directly into `GitRepo.is_valid_repo`, no?

mih:
I think last time this came up as a bug we said "no" -- dunno recall why. Now it is a bug again (different location). I am not sure what the implications would be if we make this change, and I have no time for an in-depth investigation right now.

bpoldrack:
Ok. Agree - fixing `add` ASAP is more important.

PR #2552, opened 2018-05-25
BF: instruct .gitattributes to add itself to git not annex

yarikoptic:
otherwise if I just copy, and datalad add it - it would go into annex

### DIFF
--- a/datalad/distribution/create.py
+++ b/datalad/distribution/create.py
@@ -384,6 +384,7 @@ class Create(Interface):
             # TODO this will need adjusting, when annex'ed aggregate metadata
             # comes around
             gitattr.write('config annex.largefiles=nothing\n')
+            gitattr.write('.gitattributes annex.largefiles=nothing\n')
             gitattr.write('metadata/aggregate* annex.largefiles=nothing\n')
             gitattr.write('metadata/objects/** annex.largefiles=({})\n'.format(
                 cfg.obtain('datalad.metadata.create-aggregate-annex-limit')))


mih:
Is that true, despite #1597 merged for some time?

yarikoptic:
1597 deals with it in create by forcefully ading to git... In my case I copied that .gitattributes fromanother dataset and didn't use --to-git in add to force it, assuming that datalad knows about not adding such files to annex. Test would need to replicate such copying I guess... Or if fix is to deal with addition of .gitattributes anywhere, then just add it in a subdirectory

mih:
Your interpretation of #1597 is incorrect, it also adds the same config. But the real issue is that you are adding .gitattributes on `create` for another directory. But what about `.gitignore`, that is what we have `.git*` in the dir above.

But I do not see why we should treat this particular location different from any other. So what about
changing the toplevel .gitattributes to contain

```
**/.gitattributes ...
```

yarikoptic:
well, or main issue is use of '--include-dotfiles'  for `annex add`... how much would we need to fix if we add explicit option (`include_dotfiles`) into our API as well, and make it default to False?

mih:
TBH I dont like it. It only makes sense for annex repos, but it sounds like .gitignore for dotfiles. Also that all happens deep inside, so it would be "fun" to make it work.

It is a three char diff to #1597 to fix this issue.

yarikoptic:
Sorry, I am still missing how #1597 is relevant while I am taking about add and shady existing datasets

mih:
Indeed. I am also not following. Your diff is for create... Maybe talk later?

mih:
With #2557 merged this is now redundant.

PR #2557, opened 2018-05-26
BF+TST: Alternative fix to gh-2552 (with test)

mih:
Presently unresolved discussion is https://github.com/datalad/datalad/pull/2552


mih:
Let me leap ahead and claim that this is better than #2552 (and fully covers that case too).

Issue #2578, opened 2018-05-28
`run` should automatically convert inputs/outputs to relpaths


mih:
Seems to work fine with glob expressions as well, hence I see no reason not to do this kind of argument sanitization unconditionally for anything that points inside a dataset. If it doesn't point inside, giving it to --inputs or --outputs makes no sense to start with.

kyleam:
Yes, sounds fine to me.

PR #2587, opened 2018-05-30
ENH: `run` stores input/output as relative paths (fixes gh-2578)

mih:


datalad/interface/run.py
```diff
@@ -302,10 +303,11 @@ def run_command(cmd, dataset=None, inputs=None, outputs=None, expand=None,
         'cmd': cmd,
         'exit': cmd_exitcode if cmd_exitcode is not None else 0,
         'chain': rerun_info["chain"] if rerun_info else [],
-        'inputs': inputs,
+        'inputs': [relpath(p, start=ds.path) if isabs(p) else p for p in inputs],
```
kyleam: Shouldn't start be `pwd` rather than `ds.path`?

mih: Is the `rerun` `get` call resolved via `pwd`, or via `ds.path` ( I guess the latter, if you are stating this)?

kyleam: > Is the rerun get call resolved via pwd, or via ds.path

Eh, via ds.path, which means there's a bug because globbing is done relative to `pwd`. If you glob within a subdirectory, it will be detected as an input but the the get call will say it's not there.  So, your code is in line with get/unlock/remove, and my glob code is buggy.

kyleam: Thinking about this more, I find storing the inputs relative to `pwd` to be more intuitive, especially once we add the inputs/outputs placeholders to the command.  I can fix the subdir glob bug in a separate PR.

Issue #2576, opened 2018-05-28
Option to store run records in annexed sidecar files


mih:
Here is a scenario that might always lead to leakage of private information into a run record (TL;DR: identifying information somewhere in the path names): https://github.com/psychoinformatics-de/datalad-hirni/issues/16

As a general solution to the issue of sensitive information in file names (in subdatasets, that could be detached and stay private) with respect to leakage in run records, it would be nice to be able to GPG-encrypt run records.

There is already a place for the necessary config in git: `user.signingkey=...` so we could use that too.

`re-run` would then need to decrypt on-access.

@kyleam @yarikoptic

mih:
Excellent idea. I had thought of this already in another context, but didnt made the connection that it would also help here.

I would very much be in favor of such a development.

kyleam:
I too like the idea of being able to point to a file, and it's nice to see that it'd address this issue too.

---

Only slightly related: I've been thinking about whether using the `git notes` mechanism to store run information would be something worth exploring.  This is essentially a file pointing to a commit rather than a commit pointing to a file, but it shares some features with the file pointer: (1) you can modify the run information without changing the commit (could be seen as both good and bad, I guess) and (2) you could share the repo without sharing the run information, though it's at the level of the entire run history rather than per-file.  One benefit of the notes over files is that you could still see the information in the git log.  A major downside is that the user would have to fetch a non-standard refspec in order to  work with the notes ref.

mih:
The notes option keeps coming back -- I still lack the intuition to make any useful step in this direction.

My plan would be to put the runinfo sidecar files into their own folder `.datalad/runinfo` (adjustable via config setting). The filenames would be some hash of the content (no real additional info, when the exact same command runs 10k times). Files would use JSON stream format (easy to concat without having to parse). Commit messages would replace runinfo dict with just the filehash, to be able to relocate the storage inside the dataset and still be able to find it. Actual sidecar storage can be triggered via a config flag (use case: always do this), or via cmdline option(use case: just this one).

Please stop me if that sounds wrong.

kyleam:
> The notes option keeps coming back -- I still lack the intuition to make any useful step in this direction.

Yeah, I don't think all things considered that it's a direction worth going.  It just has enough there to make it interesting to consider.

> My plan would be to put the runinfo [...]
>
> Please stop me if that sounds wrong.

That all sounds in line with my current thinking.

PR #2588, opened 2018-05-30
ENH: `run` can store records in sidecar files

mih:
This is working towards #2576, but

- [x] `rerun` needs to learn how to work with this too
- [x] tests
- [x] documentation on added config switches

### DIFF
--- a/datalad/interface/rerun.py
+++ b/datalad/interface/rerun.py
@@ -34,6 +34,7 @@ from datalad.consts import PRE_INIT_COMMIT_SHA
 from datalad.support.constraints import EnsureNone, EnsureStr
 from datalad.support.gitrepo import GitCommandError
 from datalad.support.param import Parameter
+from datalad.support.json_py import load_stream
 
 from datalad.distribution.dataset import require_dataset
 from datalad.distribution.dataset import EnsureDataset
@@ -244,7 +245,7 @@ def _revs_as_results(dset, revs):
         res = get_status_dict("run", ds=dset, commit=rev)
         full_msg = dset.repo.repo.git.show(rev, "--format=%B", "--no-patch")
         try:
-            msg, info = get_run_info(full_msg)
+            msg, info = get_run_info(dset, full_msg)
         except ValueError as exc:
             # Recast the error so the message includes the revision.
             raise ValueError(
@@ -459,7 +460,7 @@ def _get_script_handler(script, since, revision):
     return fn
 
 
-def get_run_info(message):
+def get_run_info(dset, message):
     """Extract run information from `message`
 
     Parameters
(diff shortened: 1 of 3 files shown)


datalad/interface/run.py
```diff
@@ -113,6 +116,13 @@ class Run(Interface):
             commit message.""",
             constraints=EnsureNone() | EnsureChoice("inputs", "outputs", "both")),
         message=save_message_opt,
+        sidecar=Parameter(
+            args=('--sidecar',),
+            metavar="yes|no",
+            doc="""If given, a record with information on the command execution
+            is not placed into the commit message, but into a separate record
+            file instead.""",
+            constraints=EnsureNone() | EnsureBool()),
```
kyleam: Should be `EnsureChoice` instead of `EnsureBool`, right?

mih: No, that is not needed. If given a non-bool type, it will honor the following mapping:

```
    False: '0', 'no', 'off', 'disable', 'false'
    True: '1', 'yes', 'on', 'enable', 'true'
 ```

File: datalad/interface/rerun.py [hunk 2]
```diff
@@ -491,6 +492,19 @@ def get_run_info(message):
...
             '%s' % exc_str(e)
         )
+    if not isinstance(runinfo, (list, dict)):
+        # this is a run record ID -> load the beast
+        record_dir = dset.config.get(
+            'datalad.run.record-directory',
+            default=op.join('.datalad', 'runinfo'))
+        record_path = op.join(dset.path, record_dir, runinfo)
+        if not op.lexists(record_path):
+            # too harsh IMHO, but same harshness as few lines further down
```
kyleam: The ValueError is caught and eventually converted to a result with an error status.  Do you think it should be "impossible" and skipped instead?

mih: Ah, that sounds good. No, I think it is an error, I just wouldn't want it to crash. But if it yields an 'error' result eventually I am happy.

datalad/interface/run.py
```diff
@@ -443,7 +474,7 @@ def run_command(cmd, dataset=None, inputs=None, outputs=None, expand=None,
 """
     msg = msg.format(
         message if message is not None else _format_cmd_shorty(cmd),
-        json.dumps(run_info, indent=1, sort_keys=True, ensure_ascii=False))
+        '"{}"'.format(record_id) if sidecar else record)
```
yarikoptic: What about instead of "hard-coding" for a single possible sidecar file, we right away extend with the support for (possibly) multiple by extending current run record with additional field (e.g.`"sidecars"`) which would be a dict ({"backend": "filename"}) or a list (just["filenames"])?  Then those sidecar'ed files would be loaded to update current record, which could be entirely empty (besides `"sidecars"` field) or even partially filled in?

This might allow in the future, when we might decide on supporting various "runners" for them to "link" their "sidecar" report files to the run.  E.g. (hypothesizing)  `timeit` runner to the time records on how long it took, `niceman` could point to the environment description file, some `pbs-runner` to stdout/stderr files for the job(s), ...

mih: Sounds good. But it would have to be a dict under 'sidecars' as `rerun` needs to be able to load this info again and has to put it somewhere in the structure where we know how to find it again.

Thinking this further, we should probably take out the `diff` reporting in `rerun` and move it to such a complementary runner, such that a user could decide to store a full recursive diff alongside the initial run record in a sidecar file.

I like that.

mih: Gave it one more thought and changed my mind again: I prefer to keep it the way it is (only the hash in the commit message). The functionality you proposed can be achieved using an orthogonal approach. The run record can always have a 'sidecars' field in the way you described it &mdash; whether or not it is stored in the commit message. Stuff that works with the run records (like `rerun --report`) can then pull out the info from the additional sidecar files, **after** the core record is loaded from either message or file.

Otherwise we would always load a record from a message then come up with conditions when to fully or partially replace this record from one of the sidecar files. I think the "all or nothing (and everything else later)" approach is cleaner and leaner to implement without imposing limitations on further extensions.

PR #2590, opened 2018-05-30
BF: remove stale dump_graph

yarikoptic:


kyleam:
The two failed jobs seem spurious.  I've restarted.

yarikoptic:
There is no flatten_metadata_graph even which it conditionally imports - I think all leftovers from previous metadata handling. Afaik there is no graphs anymore ATM

kyleam:
I see.  Thanks for giving more details.

yarikoptic:
Then I will merge, we can revert if it was any useful ;-)

PR #2556, opened 2018-05-25
WIP: Fixing RI issues on windows

bpoldrack:
This pull request fixes #2545 

This pull request proposes

### Changes
- [ ] This change is complete

Please have a look @datalad/developers

Issue #2545, opened 2018-05-24
RI handling fails on Windows


mih:
It should have 

```
assert_raises(ValueError, clone, 'Zoidberg', path='ssh://mars:Zoidberg')
```

But did:

```
======================================================================
380ERROR: datalad.distribution.tests.test_clone.test_invalid_args
381----------------------------------------------------------------------
382Traceback (most recent call last):
383  File "C:\Miniconda35\envs\test-environment\lib\site-packages\nose\case.py", line 198, in runTest
384    self.test(*self.arg)
385  File "C:\Miniconda35\envs\test-environment\lib\site-packages\datalad\tests\utils.py", line 669, in newfunc
386    return t(*(arg + (filename,)), **kw)
387  File "C:\Miniconda35\envs\test-environment\lib\contextlib.py", line 77, in __exit__
388    self.gen.throw(type, value, traceback)
389  File "C:\Miniconda35\envs\test-environment\lib\site-packages\datalad\utils.py", line 1307, in make_tempfile
390    yield filename
391  File "C:\Miniconda35\envs\test-environment\lib\site-packages\datalad\tests\utils.py", line 669, in newfunc
392    return t(*(arg + (filename,)), **kw)
393  File "C:\Miniconda35\envs\test-environment\lib\site-packages\datalad\tests\utils.py", line 669, in newfunc
394    return t(*(arg + (filename,)), **kw)
395  File "C:\Miniconda35\envs\test-environment\lib\contextlib.py", line 77, in __exit__
396    self.gen.throw(type, value, traceback)
397  File "C:\Miniconda35\envs\test-environment\lib\site-packages\datalad\utils.py", line 1307, in make_tempfile
398    yield filename
399  File "C:\Miniconda35\envs\test-environment\lib\site-packages\datalad\tests\utils.py", line 669, in newfunc
400    return t(*(arg + (filename,)), **kw)
401  File "C:\Miniconda35\envs\test-environment\lib\site-packages\datalad\tests\utils.py", line 669, in newfunc
402    return t(*(arg + (filename,)), **kw)
403  File "C:\Miniconda35\envs\test-environment\lib\contextlib.py", line 77, in __exit__
404    self.gen.throw(type, value, traceback)
405  File "C:\Miniconda35\envs\test-environment\lib\site-packages\datalad\utils.py", line 1307, in make_tempfile
406    yield filename
[...]

PR #2601, opened 2018-06-05
Undesired submodule url

mih:
`datalad install -d . -s ../raw sourcedata --reckless`

leads to

```diff
--- /dev/null
+++ b/.gitmodules
@@ -0,0 +1,3 @@
+[submodule "sourcedata"]
+       path = sourcedata
+       url = /home/data/psyinf/scratch/multires3t/bids/../raw
```

### DIFF
--- a/datalad/distribution/add.py
+++ b/datalad/distribution/add.py
@@ -374,6 +374,10 @@ class Add(Interface):
                 # discovered remote configuration
                 remote, branch = subds.repo.get_tracking_branch()
                 subds_url = subds.repo.get_remote_url(remote) if remote else None
+                if subds_url and op.isabs(subds_url):
+                    # plain isabs() should be OK to determine a local platform
+                    # absolute path. No URL would start with a slash or a drive
+                    subds_url = op.relpath(subds_url, start=ds_path)
                 # Register the repository in the repo tree as a submodule
                 try:
                     ds.repo.add_submodule(subds_relpath, url=subds_url, name=None)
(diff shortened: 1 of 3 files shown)


mih:
Haha, test shows how fragile things are (Windows):

```
AssertionError: './sources' != 'C:\\Users\\appveyor\\AppData\\Local\\Temp\\1\\datalad_temp_fw7rzeoz\\ds\\..\\origin'
```

so on windows it is a relative URL, just not the correct one. I assume some RI issue has prevent it to detect the original relative path as a local path...

yarikoptic:
I think we are doomed to provide "custom handling" at that line you @mih pointed out (https://github.com/datalad/datalad/blob/master/datalad/distribution/add.py#L376).  I wonder why `git` makes it full path, I guess to avoid may be ambiguity either it is path relative to the `worktree` or to `worktree/.git`. And .git/config is not shared, so there is only little crime to it.  In our case we are adding to `.gitmodules`, which can be shared, and where full PATHs IMHO make little sense. Note that we already use "relative paths" while pointing to subdatasets.  So I think it would be ok to do analysis that if "URL" is actually a PATH, then relative path should be used?

mih:
So in order to move forward we need to sort out #2556 and with it #2545

yarikoptic:
#2556 and #2545 seems to be closed now -- should this be revisited @mih?

mih:
Closing due to inactivity. Underlying issue is #3538

Issue #2608, opened 2018-06-04
manpages generation needs some TLC


yarikoptic:
# lintian whines

 (taking sample one)
```
$> lintian --verbose ../datalad_0.10.0~rc4+git167-g88f3ea8f-1_all.deb | grep datalad-add-ar 
W: datalad: manpage-section-mismatch usr/share/man/man1/datalad-add-archive-content.1.gz:1 1 != add-archive-content
W: datalad: manpage-has-bad-whatis-entry usr/share/man/man1/datalad-add-archive-content.1.gz
```

## .TH
comapre
```
$> zcat ./usr/share/man/man1/datalad-add-archive-content.1.gz | head -n 1
.TH datalad add-archive-content 1 2018\-06\-04 "datalad add-archive-content 0.10.0.rc4"
```
to
```
$> zcat /usr/share/man/man1/git-add.1.gz| grep -A2 TH                    
.TH "GIT\-ADD" "1" "04/02/2018" "Git 2\&.17\&.0" "Git Manual"
```
so I guess should bring `"datalad add-archive-content"` within `""` and  tune up last fields

# WHATIS 

seems to be missing completely?
```shell
$> zcat /usr/share/man/man1/git-add.1.gz| grep -A2 NAME  
.SH "NAME"
git-add \- Add file contents to the index
.SH "SYNOPSIS"
1 27623 [1].....................................:Mon 04 Jun 2018 07:38:09 PM EDT:.
(git)hopa:~datalad[master]build-area/XXX
$> zcat ./usr/share/man/man1/datalad-add-archive-content.1.gz | grep -A2 NAME
\fBdatalad add\-archive\-content\fR [-h] [--annex\~ANNEX] [--add-archive-leading-dir] [--strip-leading-dirs] [--leading-dirs-depth\~LEADING_DIRS_DEPTH] [--leading-dirs-consider\~LEADING_DIRS_CONSIDER] [--use-current-dir] [-d] [--key] [-e\~EXCLUDE] [-r\~RENAME] [--existing {fail,overwrite,archive-suffix,numeric-suffix}] [-o\~ANNEX_OPTIONS] [--copy] [--no-commit] [--allow-dirty] [--stats\~STATS] [--drop-after] [--delete-after] archive


--
\fB-r\fR \fI\s-1RENAME\s0\fR, \fB--rename\fR \fI\s-1RENAME\s0\fR
[...]

PR #2609, opened 2018-06-05
DOC: Better manpages (fixes gh-2608)

mih:
- proper NAME section (uses first line of description, worked for all
commands but the main datalad entrypoint, needed update)

### DIFF
--- a/datalad/interface/run_procedure.py
+++ b/datalad/interface/run_procedure.py
@@ -121,11 +121,13 @@ class RunProcedure(Interface):
 
     1. 'datalad.locations.dataset-procedures'
 
-    2. 'datalad.locations.user-procedures' (on a *nix system this defaults
-       to: '$HOME/.config/datalad/procedures')
+    2. 'datalad.locations.user-procedures' (determined by
+        appdirs.user_config_dir; defaults to '$HOME/.config/datalad/procedures'
+        on GNU/Linux systems)
 
-    3. 'datalad.locations.system-procedures' (on a *unix system this
-       defaults to '/etc/xdg/datalad/procedures')
+    3. 'datalad.locations.system-procedures' (determined by
+        appdirs.site_config_dir; defaults to '/etc/xdg/datalad/procedures' on
+        GNU/Linux systems)
 
     and subsequently in the 'resources/procedures/' directories of any
     installed extension, and, lastly, of the DataLad installation itself.
--- a/datalad/interface/run_procedure.py
+++ b/datalad/interface/run_procedure.py
@@ -119,15 +119,13 @@ class RunProcedure(Interface):
 
     Directories identified by the configuration settings
 
-    1. 'datalad.locations.dataset-procedures'
-
-    2. 'datalad.locations.user-procedures' (determined by
-        appdirs.user_config_dir; defaults to '$HOME/.config/datalad/procedures'
-        on GNU/Linux systems)
-
-    3. 'datalad.locations.system-procedures' (determined by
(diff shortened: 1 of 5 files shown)


datalad/interface/run_procedure.py
```diff
@@ -97,29 +97,94 @@ def _guess_exec(script_file):
...
+    DataLad will look for procedures in the following locations and order:
+
+    Directories identified by the configuration settings
+
+    1. 'datalad.locations.dataset-procedures'
+
+    2. 'datalad.locations.user-procedures' (on a *nix system this defaults
+       to: '$HOME/.config/datalad/procedures')
+
+    3. 'datalad.locations.system-procedures' (on a *unix system this
```
kyleam: s/*unix/*nix/, though is the broader sentence correct?  I don't think on macOS this would default to anything in "/etc/".

mih: No idea, but if you bring it up, probably not. Kill entirely or say Linux? Feel free to push into the PR, BTW.

yarikoptic:
all travis runs aren't run for some reason... but here is output from buildbot:
```shell
FAIL: datalad.cmdline.tests.test_formatters.test_manpage_formatter
    self.test(*self.arg)
  File "/home/buildbot/datalad-pr-docker-dl-nd16_04/build/datalad/cmdline/tests/test_formatters.py", line 102, in test_manpage_formatter
    assert_in('.SH {0}'.format(section), mp)
AssertionError: '.SH DESCRIPTION' not found in '.TH "create-test-dataset" "1" "2018\\-06\\-05" ""create-test-dataset None""\n.SH NAME\n\\fBcreate-test-dataset\\fR \\- create test (meta-)dataset.\n.SH SYNOPSIS\n.nh\n.HP\n\\fBcreate\\-test\\-dataset\\fR [-h] [--spec\\~SPEC] [--seed\\~SEED] path\n\n\n.hy\n.SH OPTIONS\n.TP\npath                  path/name where to create (if specified, must not\n                        exist). Constraints: value must be a string [Default:\n                        None]\n\n.TP\n-h, --help, --help-np\n                        show this help message. --help-np forcefully disables\n                        the use of a pager for displaying the help message\n.TP\n--spec SPEC           spec for hierarchy, defined as a min-max (min could be\n                        omitted to assume 0) defining how many (random number\n                        from min to max) of sub-datasets to generate at any\n                        given level of the hierarchy. Each level separated\n                        from each other with /. Example: 1-3/-2 would generate\n                        from 1 to 3 subdatasets at the top level, and up to\n                        two within those at the 2nd level . Constraints: value\n                        must be a string [Default: None]\n.TP\n--seed SEED           seed for rng. Constraints: value must be convertible\n                        to type \'int\' [Default: None]\n.SH MYTEST\n uniquedummystring'
```

mih:
Fixed that already, but latter doc build failure is more confusing, tried one fix. we'll see.

yarikoptic:
Thanks.
BTW I am late to the game but `procedures` screams `hooks` at me... Just saying

mih:
@yarikoptic Does that mean you saw them described in the docs as useful for implementing hooks? Or that you dont like this fact?

yarikoptic:
more of an additional entry to my mental mapping: siblings -> remotes, procedures -> hooks ... ;-)

yarikoptic:
could someone remind me how some folks/PRs have fancy double split (travis-ci/pr /push) which I guess requires twice the time to complete?

mih:
I think this is because docs is also a branch on the master repo for testing rtd builds.

PR #2612, opened 2018-06-05
ENH: ad-hoc Bearer token authentication/session support for LORIS

yarikoptic:
TODOs
- [x] either just to switch to use proper oauth support library or implement chaining manually similarly to how was done for NDA
- [x] tests

### DIFF
--- a/datalad/downloaders/configs/loris.cfg
+++ b/datalad/downloaders/configs/loris.cfg
@@ -1,0 +1,11 @@
+[provider:loris-demo]
+url_re = https?://demo.loris.ca/api/v0.0.2/.*
+credential = loris-demo
+authentication_type = bearer_token
+bearer_token_failure_re = "User not authenticated"}$
+
+
+[credential:loris-demo]
+url = https://demo.loris.ca
+type = token
+
--- a/datalad/downloaders/configs/loris.cfg
+++ b/datalad/downloaders/configs/loris.cfg
@@ -1,11 +1,10 @@
 [provider:loris-demo]
 url_re = https?://demo.loris.ca/api/v0.0.2/.*
 credential = loris-demo
-authentication_type = bearer_token
-bearer_token_failure_re = "User not authenticated"}$
-
+authentication_type = loris-token
+loris-token_failure_re = "User not authenticated"}$
 
 [credential:loris-demo]
-url = https://demo.loris.ca
-type = token
+url = https://demo.loris.ca/api/v0.0.2/login
+type = loris-token
 
--- a/datalad/downloaders/http.py
+++ b/datalad/downloaders/http.py
@@ -297,7 +297,7 @@ class HTTPDigestAuthAuthenticator(HTTPRequestsAuthenticator):
 
 @auto_repr
 class HTTPBearerTokenAuthenticator(HTTPRequestsAuthenticator):
-    """Authenticate via HTTP digest authentication
+    """Authenticate via HTTP Authorization header
     """
     def __init__(self, **kwargs):
         # so we have __init__ solely for a custom docstring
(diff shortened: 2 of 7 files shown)


datalad/downloaders/configs/loris.cfg
```diff
@@ -0,0 +1,11 @@
+[provider:loris-demo]
+url_re = https?://demo.loris.ca/api/v0.0.2/.*
```
driusan: This would cover the most LORIS instances by default if it were `https://.*\.loris\.ca/api/.*` (I don't think it can be made more generic than that and needs #2622 to handle arbitrary LORIS installs.)

datalad/downloaders/http.py
```diff
@@ -293,6 +295,25 @@ def _post_credential(self, *args, **kwargs):
         raise NotImplementedError("not yet functioning, see https://github.com/kennethreitz/requests/issues/2934")
 
 
+@auto_repr
+class HTTPBearerTokenAuthenticator(HTTPRequestsAuthenticator):
+    """Authenticate via HTTP digest authentication
```
driusan: "HTTP authorization header", not "digest authentication"

datalad/downloaders/configs/loris.cfg
yarikoptic: well -- this one is `loris-demo` (not just `loris`), credential should be specific for server/deployment.  If there would be "realone.loris.ca" most likely it would need different credentials, right?
but indeed we could generalize it to all api versions  I guess ;-)

driusan: yes, that's true, each subdomain would have a different token. Is there any concept of "These all use the same configuration, but the individual credentials differ" in configuring providers?

yarikoptic: Not atm

driusan:
I added some tests and rebased this branch onto master. If someone could review, that'd be nice.

I'm not very convinced that they're good tests, and I'm sure there must be a better way to get a variable out of a closure in Python but these at least pass locally.

datalad/downloaders/http.py
```diff
@@ -293,6 +295,25 @@ def _post_credential(self, *args, **kwargs):
...
+    """Authenticate via HTTP Authorization header
+    """
+    def __init__(self, **kwargs):
+        # so we have __init__ solely for a custom docstring
+        super(HTTPBearerTokenAuthenticator, self).__init__(**kwargs)
+
+    def _post_credential(self, credentials, post_url, session):
+        # we do not need to post anything, just inject token into the session
+        session.headers['Authorization'] = "Bearer %s" % credentials['token']
+        # TODO: actually access some location to verify that token is correct
```
yarikoptic: @driusan I wondered if there is a way to verify that the token is valid... is there?

datalad/support/third/loris_token_generator.py
```diff
@@ -0,0 +1,31 @@
...
+# ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### #
+import sys
+import json
+
+if sys.version_info[0] == 2:
+    import urllib2 as urllib_request
+else:
+    from urllib import request as urllib_request
+
+class LORISTokenGenerator(object):
```
yarikoptic: A short docstring describing its purpose would be most welcomed ;-)

yarikoptic:
OSX failure is known: https://github.com/datalad/datalad/issues/2619  yet to get to it.

datalad/downloaders/http.py
driusan: I don't think Authorization headers have any meaning other than what the web server receiving it gives to them, so I don't think a generic HTTPBearerTokenAuthenticator could validate the token other than by trying to make a (real) superfluous request and making sure it doesn't receive a 401 response.

datalad/support/third/loris_token_generator.py
driusan: done

yarikoptic:
Let's merge it and see later if anything to be adjusted

PR #2613, opened 2018-06-05
BF: Escape dots in URL regular expressions

driusan:
This pull request fixes unescaped dots in some regular expressions for URLs.


yarikoptic:
thanks @driusan

PR #2615, opened 2018-06-05
ENH: make create-sibling work with local paths as well

### LINKED ISSUE #1680: create-sibling doesn't support local paths
and there is no ssh there

```shell
datalad create-sibling /mnt/sub-001-sibling
[ERROR  ] Unsupported SSH URL: '/mnt/sub-001-sibling', use ssh://host/path or host:path syntax [create_sibling.py:__call__:477] (ValueError) 
```

yarikoptic:
This pull request fixes #1680

@mvdoc since you have reminded me about this, I will leave it to you to finish up as a little hacckathon (or a flight to it) project

This pull request proposes to treat "ssh" as just a regular "shell" thus allowing for creating a local remote sibling with all the gadgets (web ui, hooks etc), since in principle ssh is just a shell it is 

TODOs:
- [ ] check if it actually works
- [ ] decide on either API could/should be adjusted.  Now many options still screams SSH
- [ ] at least some basic tests

### DIFF
--- a/datalad/distribution/create_sibling.py
+++ b/datalad/distribution/create_sibling.py
@@ -45,26 +45,47 @@ from datalad.support.constraints import EnsureStr, EnsureNone, EnsureBool
 from datalad.support.constraints import EnsureChoice
 from datalad.support.exceptions import InsufficientArgumentsError
 from datalad.support.exceptions import MissingExternalDependency
-from datalad.support.network import RI
+from datalad.support.network import RI, PathRI
 from datalad.support.network import is_ssh
 from datalad.support.sshconnector import sh_quote
 from datalad.support.param import Parameter
+from datalad.support.external_versions import external_versions
 from datalad.utils import make_tempfile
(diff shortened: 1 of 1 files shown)


mih:
Whoever is approaching this please read my plea:

Let's suck all the needed "actions" into the RunnerAdaptor class. ATM is already has `copy()`, but if you scan the code, you'll see that is sends actual shell commands to the "shell". That assumes that the other side has a shell. However, that stands in the way of making this works with a windows target.

Ok, I hear you whine, but think about this:

If we could not just talk to a shell, but to an "API", we could dissolve `create_sibling_github` -- and how cool would that be!?

yarikoptic:
yes, API would be nice to pursue, and dissolving github one would be great.  BUT could be done after as a  bigger refactoring. This one just shows that it should be quite easy to make it work at least for some  more common in datalad OSs ;-)

mih:
Closing due to inactivity.

PR #1995, opened 2017-11-25
Add missing dependency on 'duecredit'

kskyten:
The `aggregate-metadata` command did not work for me until I installed the `duecredit` package. Hopefully this is the right place to put the dependency in.

### Changes
- [x] This change is complete
- [ ] This is in progress

Please have a look @datalad/developers

Issue #2550, opened 2018-05-24
Failing (randomly) test(s?): Ref 'a7b0fde' did not resolve to an object


yarikoptic:
#### What is the problem?

I think we all observed this but I found no issue for it -- it just keeps coming and I just restart travis build for it to go away typically... but we should pin it down.  This time it is 
https://travis-ci.org/datalad/datalad/jobs/383314155
```
  File "/home/travis/build/datalad/datalad/datalad/distribution/tests/test_update.py", line 294, in test_update_volatile_subds
    assert_result_count(ds.update(merge=True, recursive=True), 1, type='dataset')
```
and I would not be surprised if #2542 and other "open files" issues would resolve it eventually ;-)

edit: speaking of a devil, another fresh one:
https://travis-ci.org/datalad/datalad/jobs/383312708
```
  File "/home/travis/build/datalad/datalad/datalad/distribution/tests/test_create_sibling.py", line 491, in test_replace_and_relative_sshpath
    published4 = ds.publish(to='localhost', transfer_data='all')
...
  File "/home/travis/build/datalad/datalad/datalad/distribution/publish.py", line 415, in _publish_dataset
    ds.repo.fetch(remote=remote)
  File "/home/travis/build/datalad/datalad/datalad/support/gitrepo.py", line 1707, in fetch
    fi_list += rm.fetch(refspec=refspec, progress=progress, **kwargs)
```

bpoldrack:
You linked Travis builds that are passing ...

yarikoptic:
another one https://travis-ci.org/datalad/datalad/jobs/388571693 (will not restart since know that "bogus"):
```
ERROR: datalad.distribution.tests.test_update.test_reobtain_data
...
  File "/home/travis/build/datalad/datalad/datalad/distribution/update.py", line 169, in __call__
    prune=True)  # prune to not accumulate a mess over time
  File "/home/travis/build/datalad/datalad/datalad/support/gitrepo.py", line 1711, in fetch
    fi_list += rm.fetch(refspec=refspec, progress=progress, **kwargs)
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/git/remote.py", line 789, in fetch
    res = self._get_fetch_info_from_stderr(proc, progress)
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/git/remote.py", line 706, in _get_fetch_info_from_stderr
[...]

PR #2618, opened 2018-06-06
BF: no duecredit should be needed

yarikoptic:
dependency was added in https://github.com/datalad/datalad/pull/1995 without us actually checking why it is needed... and AFAIK it shouldn't be (yet at least).  conda doesn't have it yet, so complicates packaging


yarikoptic:
failing travis is https://github.com/datalad/datalad/issues/2550

PR #2622, opened 2018-06-06
Add support for loading provider configs from other locations

driusan:
This adds support for loading config for providers from places
other than those configs that are distributed with the datalad
distribution. In particular, it adds support for:

 - dataset/.datalad/providers/
 - $HOME/.config/datalad/providers/
 - $HOME/.datalad/providers/
 - /etc/datalad/providers/

I'm not sure if this is the most idiomatic way to do this in the datalad codebase or if there's a more centralized location, but it seems to work.

(I also cleaned up the English in the function comment and clarified it slightly.)

### DIFF
--- a/datalad/downloaders/providers.py
+++ b/datalad/downloaders/providers.py
@@ -211,6 +211,7 @@ class Providers(object):
             # System config
             if dirs.site_config_dir is not None:
                 files.extend(glob(pathjoin(dirs.site_config_dir, "providers", "*.cfg")))
+
             # User config
             if dirs.user_config_dir is not None:
                 files.extend(glob(pathjoin(dirs.user_config_dir, 'providers', '*.cfg')))
--- a/datalad/downloaders/providers.py
+++ b/datalad/downloaders/providers.py
@@ -30,6 +30,8 @@ from ..utils import assure_list_from_str
 from ..utils import auto_repr
 from ..utils import get_dataset_root
 
+from ..interface.common_cfg import dirs
+
 lgr = getLogger('datalad.downloaders.providers')
 
 # dict to bind authentication_type's to authenticator classes
@@ -46,7 +48,6 @@ AUTHENTICATION_TYPES = {
 }
 
 from .credentials import CREDENTIAL_TYPES
-
 
 def resolve_url_to_name(d, url):
     """Given a directory (e.g. of SiteInformation._items or Credential._items)
@@ -180,9 +181,8 @@ class Providers(object):
 
         - datalad/downloaders/configs/*.cfg files provided by the codebase
         - current dataset .datalad/providers/
-        - Linux style user's config directory:  $XDG_CONFIG_HOME/datalad/providers or ~/.config/datalad/providers/*.cfg by default
-        - Unix style user's home directory: ~/.datalad/providers/*.cfg)
-        - system-wide datalad installation/config: /etc/datalad/providers/*.cfg
+        - User's home directory directory (ie ~/.config/datalad/providers/*.cfg)
+        - system-wide datalad installation/config (ie /etc/datalad/providers/*.cfg)
 
         For sample configs files see datalad/downloaders/configs/providers.cfg
 
@@ -199,21 +199,15 @@ class Providers(object):
         if files is None:
             # Config files from the datalad dist
             files = glob(pathjoin(dirname(abspath(__file__)), 'configs', '*.cfg'))
-            if 'HOME' in os.environ:
-                # Linux style config
-                if 'XDG_CONFIG_HOME' in os.environ:
-                    files.extend(glob(pathjoin(os.environ['XDG_CONFIG_HOME'], 'datalad', 'providers', '*.cfg')))
-                else:
-                    files.extend(glob(pathjoin(os.environ['HOME'], '.config', 'datalad', 'providers', '*.cfg')))
-                # Real Unix style config
-                files.extend(glob(pathjoin(os.environ['HOME'], '.datalad', 'providers', '*.cfg')))
+            # User config
+            files.extend(glob(pathjoin(dirs.user_config_dir, 'providers', '*.cfg')))
 
-                # Dataset config
-                dsroot = get_dataset_root("")
-                if dsroot is not None:
-                    files.extend(glob(pathjoin(dsroot, '.datalad', 'providers', '*.cfg')))
-            # System wide config
-            files.extend(glob("/etc/datalad/providers/*.cfg"))
+            # Dataset config
+            dsroot = get_dataset_root("")
+            if dsroot is not None:
+                files.extend(glob(pathjoin(dsroot, '.datalad', 'providers', '*.cfg')))
+            # System config
+            files.extend(glob(pathjoin(dirs.site_config_dir, "providers", "*.cfg")))
         config.read(files)
 
         # We need first to load Providers and credentials
--- a/datalad/downloaders/providers.py
+++ b/datalad/downloaders/providers.py
@@ -18,6 +18,7 @@ from six import iteritems
 import re
 from os.path import dirname, abspath, join as pathjoin
 from six.moves.urllib.parse import urlparse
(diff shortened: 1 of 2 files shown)


mih:
Quick feedback: We are using the `appdirs` package for that (also works on other platforms. Grep for `user_config_dir` and `site_config_dir` for examples.

driusan:
I updated it to use AppDirs, but I can't really test the `/etc/datalad` and `~/.config/datalad` variations since I'm on a Mac and AppDirs thinks my site and user config directories are `/Library/Application Support/datalad` and `/Users/driusan/Library/Application Support/datalad` respectively (though I suppose that suggests that it's working as intended..)

(I'm also not really sure why my code coverage is being reported as plummeting for files I didn't touch..)

yarikoptic:
let's wait for the fresh travis build and see if coverage becomes sane again ;)

datalad/downloaders/providers.py
```diff
@@ -173,28 +177,37 @@ def __iter__(self):
 
     @classmethod
     def from_config_files(cls, files=None, reload=False):
```
yarikoptic: note that ideally there should be a dedicated test which would verify that it finds present files and loads then in desired (here it is not actually, system is added last) order:  datalad -> dataset -> system -> user (I think this would be the best, from hardcoded, to shipped, to custom local); so that settings in e.g. dataset overload the one in datalad. 
 Would be nice to test in user directory overload as well, but HOME might be left not modified to custom one for  some test runs... not sure if appdirs has "memory" but you could create a temp directory and `patch.dict('os.environ', {'HOME': path})` to mock for that specific test 
makes sense or I talk jibberish?

driusan: I'll see about adding some tests, but I'm not sure how to fake the system config files in a testing context or how to make the user config test work in a cross platform way that wouldn't end up just using the same AppDir config (implying that it would be a somewhat tautologically and not doing much useful testing.)

Also, shouldn't the precedence be datalad dist -> system -> user -> dataset? (in any case you're right about the fact that system taking precedence is definitely wrong..)

yarikoptic: may be you are right.  my thinking was -- what if dataset is old and provided settings are no longer valid and need adjustment... well, dataset could be fixed up I guess.
@mih blind me can't find a clear description in the docs about our order in the `config` handling... I just remember that env variable overrides anything, I guess that it is system -> user -> dataset ->  dataset local (.git/config) -> env , right?

driusan: Nevermind, I just realized I can fake it by patching dirs.user_config_dir and dirs.site_config_dir for the test...

kyleam: > blind me can't find a clear description in the docs about our order in the config handling... I just remember that env variable overrides anything, I guess that it is system -> user -> dataset -> dataset local (.git/config) -> env , right?

Not a specific outline of the order, but the [configuration docs](http://docs.datalad.org/en/docs/config.html) claim that "global and local settings on a particular machine always override configuration shipped with a dataset", so I'd guess your guess isn't right (or at least isn't supposed to be).

driusan: That documentation seems to specifically be referring to the git configuration settings (where that makes some sense, since those settings are intended for user customization of settings). As an outsider, it makes less sense to me for datalad downloader configuration, because it means the system (if you call that "global") settings will always override anything thing set in a data set, and there would be no way to customize things on a per dataset level on any system where a system administrator has set something up.

But I can go with that if that's the intended design of datalad.

mih: @driusan Nothing is set in stone. We are running with this design, because it prevents arbitrary datasets from overriding system or user config (where the latter takes precedence over the former). This include identity information and credentials.

However, if we identify a scenario where this prevents sensible behavior, we will simply stop and reassess the situation -- no biggie.

driusan: @mih So dist -> dataset -> system -> user?

yarikoptic: thanks for the lively discussion ;-)  yes, I could have been wrong.  yes, lets proceed with `dist -> dataset -> system -> user`
Cheers!

yarikoptic:
FWIW @driusan - coverage is sane again ;-)

datalad/downloaders/tests/test_providers.py
```diff
@@ -85,4 +91,77 @@ def test_get_downloader_class():
...
+
+        datalad defaults < dataset defaults < system defaults < user defaults
+    """
+
+    # Setup. This would be nice to have a context manager for, but
+    # "with tempdir.TemporaryDirectory" wouldn't work on Python 2.7..
+
+    # Create a fake system directory, user directory, and dataset directory
+    sysdir = mkdtemp("sys", "dataladtest", None)
+    os.mkdir(sysdir + "/providers/", 0o700)
```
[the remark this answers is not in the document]
yarikoptic: but in general, we prefer to use `@with_tempfile(mkdir=True)` decorator(s) to create temporary directories -- that allows to keep them when needed even if test fails

datalad/downloaders/tests/test_providers.py
```diff
@@ -85,4 +91,77 @@ def test_get_downloader_class():
...
+    os.makedirs(dsdir+ "/.datalad/providers/", 0o700)
+    # datalad looks for the .git directory to find the root of the
+    # dataset, not the .datalad directory.
+    os.makedirs(dsdir+ "/.git", 0o700)
+
+    with open(dsdir+"/.datalad/providers/atest.cfg", "w") as f:
+        f.write("""[provider:dscrcns]
+url_re = https?://crcns\.org/.*
+authentication_type = none
+        """)
```
yarikoptic: instead of `@with_tempfile` you could, I think, at once use `@with_tree` helper where you could say what hierarchy to create (with content for the files) in  temp directories... eg for this one

```python
@with_tree(tree={
  '.datalad': {'providers': {'atest.cfg':"""\
[provider:syscrcns]
....
"""}})
...
def test_...(dsdir, ...):
```

datalad/downloaders/tests/test_providers.py
```diff
@@ -85,4 +91,77 @@ def test_get_downloader_class():
...
+    # Test the default, this is an arbitrary provider used from another
+    # test
+    providers = Providers.from_config_files(reload=True)
+    provider = providers.get_provider('https://crcns.org/data....')
+    assert_equal(provider.name, 'crcns')
+
+    # Test that the dataset provider overrides the datalad
+    # default
+    owd = os.getcwd()
+    os.chdir(dsdir)
```
yarikoptic: Just FYI: avoid direct `os.chdir`at all costs in datalad... use `datalad.utils.chpwd` if really really needed.

Here, use `with chpwd(dsdir):` context manager

datalad/downloaders/tests/test_providers.py
```diff
@@ -85,4 +91,77 @@ def test_get_downloader_class():
...
+
+    # Test that the user defaults take precedence over the system
+    # defaults
+    with patch.multiple("appdirs.AppDirs", site_config_dir=sysdir, user_config_dir=userdir):
+        providers = Providers.from_config_files(reload=True)
+        provider = providers.get_provider('https://crcns.org/data....')
+        assert_equal(provider.name, 'usercrcns')
+
+    # Cleanup
+    os.chdir(owd)
```
yarikoptic: see above on chdir, chpwd, etc, so even below would not be needed as well ;-)

datalad/downloaders/tests/test_providers.py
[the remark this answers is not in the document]
mih: Quick reminder that we feel like we are in a period of transitioning from importing individual functions from `os.path` to importing `import os.path as op` and use them through this prefix. ;-)

driusan:
I think I've done all the requested changes, let me know if I missed anything..

mih:
LGTM. @yarikoptic objections?

PR #2625, opened 2018-06-07
Mutagen and httpretty fixups

yarikoptic:
will be a quick merge if fixes and doesn't break more ;-)


mih:
FTR: mutagen seems pretty moody in general, so not sure if worth chasing versions. But I don't mind.

PR #2627, opened 2018-06-07
minor OPT: metadata + logging during (commented out)

yarikoptic:
- use iteritems not .items so stays generator for both PY2 and PY3.  I do not think that it would result in the major memory relief, but still worth using it this way
- while going through the code came to realization that all extractors manually repeat the progress logging, although whenever they are generators we could handle it centrally, thus added commented out log_progress pieces.  log messages aren't actually logged at all, I've tried to resolve it with https://github.com/yarikoptic/datalad/compare/opt-metadata...yarikoptic:opt-metadata-log?expand=1  but then it doesn't look good anyways in case of nested progresses (embedded one would print final message thus ruining the progressbar for the above), so better to leave as is

This one should be safe, so next aggregation attempt will go with this one

### DIFF
--- a/datalad/metadata/metadata.py
+++ b/datalad/metadata/metadata.py
@@ -566,29 +566,36 @@ def _get_metadata(ds, types, global_meta=None, content_meta=None, paths=None):
 
         unique_cm = {}
         extractor_unique_exclude = getattr(extractor_cls, "_unique_exclude", set())
-        log_progress(
-            lgr.debug,
-            'metadataextractors_loc',
-            'Metadata extraction per location for %s', mtype,
-            # contentmeta_t is a generator... so no cound is known
-            # total=len(contentmeta_t or []),
-            label='Metadata extraction per location',
-            unit=' locations',
-        )
+        # TODO: ATM neuroimaging extractors all provide their own internal
+        #  log_progress but if they are all generators, we could provide generic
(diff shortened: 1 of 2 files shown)


datalad/metadata/metadata.py
```diff
@@ -561,9 +566,36 @@ def _get_metadata(ds, types, global_meta=None, content_meta=None, paths=None):
 
         unique_cm = {}
         extractor_unique_exclude = getattr(extractor_cls, "_unique_exclude", set())
+        # TODO: ATM neuroimaging extractors all provide their own internal
+        #  log_progress but if they are all generators, we could provide generic
+        #  handling of the progress here.  Note also that log message is actually
+        #  seems to be ignored and not used, only the label ;-)
```
kyleam: The message are shown when `is_interactive()` is false.

```python
import logging

lgr = logging.getLogger("datalad.test")
log_progress(lgr.debug, "id", "message, start", label="label", unit=" unit")
log_progress(lgr.debug, "id", "message, update", update=1, increment=True)
log_progress(lgr.debug, "id", "message, done")
```

```
% DATALAD_LOG_LEVEL=debug python scratch.py | cat
[DEBUG] message, start
[DEBUG] message, update
[DEBUG] message, done
```

yarikoptic: I was blind, thanks!

PR #2629, opened 2018-06-07
BF(TST): do not require url in the error message if proxy is the reason

yarikoptic:
This pull request should fix FTBFS on 14.04


yarikoptic:
FWIW, build passed on 14.04 including py3.whoohoo

yarikoptic:
this is a trivial fix, passed testing while building debian package... so I will merge it and cancel travis build

PR #2644, opened 2018-06-10
BF: terminal size could be reported 0, thus causing the breakage due to -5 below in the code

yarikoptic:
See e.g. https://circleci.com/gh/conda-forge/datalad-feedstock/6

### DIFF
--- a/datalad/auto.py
+++ b/datalad/auto.py
@@ -54,6 +54,9 @@ except Exception as exc:
 
 lzma = None
 try:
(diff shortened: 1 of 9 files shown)


yarikoptic:
actually probably if exact 0 width is reported, we should make it 80 anyways. I will adjust the patch

yarikoptic:
travis passed... coverage dived so heavily --  yet to look if "legit"

yarikoptic:
ok , coverage is better but still `-0.5%`, not sure why exactly

PR #2642, opened 2018-06-11
run-procedure without argument just ERRORs out

mih:
#### What is the problem?
```
$> datalad run-procedure
[ERROR  ] list index out of range [run_procedure.py:__call__:200] (IndexError)
```
should instead spit out `--help` synopsis like any other misspecified command does


mih:
Ah, should not have attached the commit with the fix to @yarikoptic issue. Sorry.

Issue #2546, opened 2018-05-24
We need a cross-platform convention for paths


mih:
We have path issues on non-POSIX systems (i.e. Windows). See e.g. #2528 #2531 

One source of problems is how GitPython handles paths on Windows.  For example, our `GitRepo.get_indexed_files()` returns the keys in GitPython's index (which are relative paths within the repository), but they are always POSIX paths.

That conflicts with our apparent assumption that we could simply turn them into absolute paths with `os.path` functions on any platform. Soooo, what shall we do?

1. Always return any path in POSIX conventions
    - shifts a substantial amount of platform dependent code to become a user's responsibility
    - we would have to test/convert any incoming path specification to POSIX

2. Always return platform dependent paths
    - most gentle to users
    - we would have to convert any path specification that needs to become POSIX to get properly tested (for equality, etc.), or needs to get stored in Git (e.g. gitmodules, etc.) ourselves.

I think the responsible choice is (2), while (1) is sending the message "do not use this on non-POSIX platforms" (in which case I would argue that we should declare any effort to get anything to run on windows as futile and discouraged).

mih:
@yarikoptic @kyleam Thoughts?

yarikoptic:
Throughout the code we use is.path.join, which votes 2 afaik

mih:
OK, so decision would be to consider any POSIX path coming out of the low-level API to be a bug (assuming this is the level we want to take care of this, which seems to make sense to me).

bpoldrack:
Yes, except it's supposed to return a (part of) URL, which might confusingly be called `path`. In that case it should still return a "POSIX path" but it needs to be called `url` instead. Just saying, since the implicit assumption that a path can be treated that way is still widespread.

yarikoptic:
yes, making it clear(er) where a "path" component is the "urlpath" (or "posixpath") would be useful!

mih:
So that is settled.

mih:
I am reopening this: https://bytes.com/topic/python/answers/19793-convert-between-windows-style-paths-posix-style-paths

It looks like we are doing it wrong, and NT supports POSIX paths, just not at the command prompt. Hence we might want to rethink our judgement.

yarikoptic:
[...]

PR #2653, opened 2018-06-12
TST: Verify how much path flexibility Python has across platforms

mih:
Any target platform of ours (including Windows) can deal with paths that have forward-slashes as `os.path.sep`. That includes absolute paths with drive letters on Windows. They just have to use forward-slashes consistently and not mix and match back- and forward-slashes.

This  is just a test to verify that. But now we should move towards a consistent path handling. IMHO that requires two things:

1. stop using `os.path`, but use `posixpath` exclusively.
2. establish a layer (plus helper function) that converts any incoming path to a posix-like path, regardless of platform.

Here is a thing you are probably wondering about:

```python
>>> import posixpath
>>> p='C:/users/mike'
>>> posixpath.relpath(p, 'C:/users')
'mike'
>>> posixpath.relpath(p, 'C:/')
'users/mike'
```

### DIFF
--- a/datalad/tests/test_base.py
+++ b/datalad/tests/test_base.py
@@ -8,6 +8,7 @@
 # ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
 
 import os
+import os.path as op
 
 from .utils import (
     chpwd,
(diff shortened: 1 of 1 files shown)


bpoldrack:
That's really interesting. 

I'm not sure whether this is what you imagined, but I think we already have that required layer. It's called `PathRI`. It has an attribute `posixpath`. I think we should exclusively use `PathRI`(and the other `RI`s) instances and operate on its `posixpath` attribute. Then we can use python's `posixpath` module exclusively for any path operation. Those `RI`s might need some enhancement and/or fixing, but from my point of view that's the way to go.

kyleam:
> IMHO that requires two things:
>
> 1. stop using `os.path`, but use `posixpath` exclusively.
> 2. establish a layer (plus helper function) that converts any incoming path to a posix-like path, regardless of platform.

This involves the direction of (1) [here] (i.e. "always return any path in POSIX conventions"), right?  If so, does the path flexibility shown in this pr mean the "shifts a substantial amount of platform dependent code to become a user's responsibility"  under (1) isn't valid?

[here]: https://github.com/datalad/datalad/issues/2546#issue-326003288

mih:
> does the path flexibility shown in this pr mean the "shifts a substantial amount of platform dependent code to become a user's responsibility" under (1) isn't valid?

I think everything state there remains valid. My point here is that our own 'internal' representation choice is not POSIX vs native, but can be a more practical compromise: os.path.sep == '/' on all platforms,  but the rest stays native (drive letters, etc).

That would mean we have to perform this kind of normalization on the way in, and only this more minimal conversion on the way out.

I am still unclear on where exactly a good place for "on the way out" would be. Likely we could limit this to the cmdline interface. Using forward slash paths within Python on Windows works just fine. Only when we want to interact with other shell functionality it might become an issue on windows.

Qs are:

- do we care about shell scripting on windows?
- if so, what exactly are the issues we are facing (what shell cannot handle which kind of path style)

bpoldrack:
> I am still unclear on where exactly a good place for "on the way out" would be.

Any output, I guess. It would be weird to have `C:\some\path' as an user input and then report back something about `C:/some/path`. And since output might be generated by subprocesses, we should call them with native paths. At least if they were direct user inputs.

PR #2662, opened 2018-06-15
Change version number so that datalad-crawler can be installed

driusan:
v0.10.0.dev1 < v0.10.0.rc3, which is the minimum required version for datalad-crawler.

I'm not entirely sure that this is the right way to fix it, but at least it lets me
install datalad-crawler against the master branch.

### Changes
- [x] This change is complete

### DIFF
--- a/datalad/version.py
+++ b/datalad/version.py
@@ -13,7 +13,7 @@ import sys
 from os.path import lexists, dirname, join as opj, curdir
 
 # Hard coded version, to be done by release process
-__version__ = '0.10.0.dev1'
+__version__ = '0.10.1.dev1'
 __full_version__ = __version__
 
 # NOTE: might cause problems with "python setup.py develop" deployments


yarikoptic:
who knows what pypi/pip uses for version comparisons?
now I wonder what should we do in the future for the releases.  I was certain that `.dev` suffix is something what distutils comprehends but apparently I was wrong: http://epydoc.sourceforge.net/stdlib/distutils.version.StrictVersion-class.html , StrictVersion knows only about `a` and `b` suffixes for alpha/beta pre-releases

mih:
This page knows:

https://setuptools.readthedocs.io/en/latest/setuptools.html#specifying-your-project-s-version

`.dev` like `.rc` compares to less than the version string above.

yarikoptic:
I guess we should adjust our external_versions setup to use this one from pkg_resources instead distutils... This all is dark corners of python IMHO

PR #2663, opened 2018-06-15
BF: Handle 401 errors from LORIS tokens

driusan:
When a request is made with an incorrect username/password,
urllib.urlopen raises an HTTPError exception. Since datalad
is looking for AccessDeniedError exceptions in order to decide
whether to re-prompt for new credentials, it doesn't get caught
and there's no easy way to enter the correct credentials if
you don't get it right the first time.

This converts the implicit exception in LORIS token generator
to an AccessDeniedError.

### DIFF
--- a/datalad/support/third/loris_token_generator.py
+++ b/datalad/support/third/loris_token_generator.py
@@ -9,13 +9,8 @@
 import sys
 import json
 
-
-if sys.version_info[0] == 2:
-    import urllib2 as urllib_request
-    from urllib2 import HTTPError
-else:
-    from urllib import request as urllib_request
-    from urllib import HTTPError
+from six.moves.urllib.request import Request, urlopen
+from six.moves.urllib.error import HTTPError
 
 from datalad.downloaders.base import AccessDeniedError, AccessFailedError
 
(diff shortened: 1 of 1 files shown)


datalad/support/third/loris_token_generator.py
```diff
@@ -9,10 +9,11 @@
 import sys
 import json
 
-if sys.version_info[0] == 2:
-    import urllib2 as urllib_request
-else:
-    from urllib import request as urllib_request
+import six.moves.urllib
```
yarikoptic: from six.moves import urllib
?
Since using urllib below

driusan: I think I've finally got it working with both Python 2 and 3. It's `six.moves.urllib.request` that I needed, not `six.moves.urllib`

yarikoptic:
Just that matrix run finally became too long again... Not directly related to this one... I will have a look at it all when we finally get some stable ground under our feet (not airports etc)

yarikoptic:
Please merge or rebase on master -- in #2670 I've made that run a bit faster, and now it shouldn't fail with up to date git annex ;-)

driusan:
At least the Travis part is passing now..

PR #2670, opened 2018-06-26
ENH/BF: compatibility with annex of today (0626)

yarikoptic:
Which restricts which URLs could be accessed by default, removing localhosts and file:/// which we rely on in tests.  This PR proposes to relax this for any AnnexRepo during testing.  I think we should be secure in general since we 
- do not install/clone non-trusted repos from online (only some of our own, which we hopefully could trust)
- do not publish any of the test repos online, so even if any info gets leaked locally, shouldn't get online
- we do not use and I do not enable WORM backends or relaxed urls in the tests.

### Changes
- [ ] Verify that somehow doesn't break anything
- [ ] upload patched git-annex-standalone and rerun
- [ ] This change is complete


yarikoptic:
I am ready to strangle those pkg_requires and entry points until they stop! ;)  I think the issue why entry points do not load is as warned at the [pip install stage](http://smaug.datalad.org:8020/builders/datalad-pr-docker-dl-nd16_04/builds/2715/steps/pip%20install%20-e%20./logs/stdio): 
```
pygithub 1.40 has requirement requests>=2.14.0, but you'll have requests 2.9.1 which is incompatible.
```
but I do not get it why then it didn't upgrade the `requests` module???

yarikoptic:
oh, I saw that failure but forgot to make a note, with git 2.18 we get
```
FAIL: datalad.tests.test_config.test_something
    self.test(*self.arg)
  File "/home/travis/build/datalad/datalad/datalad/tests/utils.py", line 417, in newfunc
    return t(*(arg + (d,)), **kw)
  File "/home/travis/build/datalad/datalad/datalad/tests/utils.py", line 587, in newfunc
  File "/home/travis/build/datalad/datalad/datalad/tests/test_config.py", line 178, in test_something
    ok_file_has_content(global_gitconfig, '[datalad "unittest"]', strip=True)
  File "/home/travis/build/datalad/datalad/datalad/tests/utils.py", line 400, in ok_file_has_content
    assert_equal(content, content_, **kwargs)
AssertionError: '[datalad "unittest"]' != ''
```
I will have a look now

PR #2672, opened 2018-06-27
BF: move LOG_TRACEBACK=collide testing into leaner matrix

yarikoptic:
There is really no need to test it on the heavy debug output one,
and I think it is one of the reasons for that matrix run to be so slow,
now reaching critical 50min and failing testing as e.g. in #2663
So moved it into another and hope that testing would speed up for that run
enough to become usable again

note: it is probable that testing would fail on travis ATM due to https://github.com/datalad/datalad/pull/2670#issuecomment-400677339  but I will merge it as far as time reduces and no other immediate failures


yarikoptic:
ah, since it would need fixes in #2670 to do any reasonable testing, I will just meld it with that one

PR #2678, opened 2018-06-28
RF(TEMP): make "tailored" to be the "default" format -- just to see what breaks

yarikoptic:


yarikoptic:
closing -- nothing broke

mih:
Actually closing.

PR #2687, opened 2018-07-03
DOC: information on support of native git-annex metadata

yarikoptic:
I think we need to elaborate even more on what differs datalad metadata from git-annex metadata, but at least for now it would list pros/cons for git-annex and provide some usecases where people might want to use native one

### DIFF
--- a/docs/source/metadata.rst
+++ b/docs/source/metadata.rst
@@ -9,7 +9,7 @@ Overview
 DataLad has built-in, modular, and extensible support for metadata in various
 formats. Metadata is extracted from a dataset and its content by one or more
 extractors that have to be enabled in a dataset's configuration. Extractors
-yield metadata in a JSON-LD-like structure that can be arbitrarily complex and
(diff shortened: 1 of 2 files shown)


docs/source/metadata.rst
```diff
@@ -9,7 +9,7 @@ Overview
 DataLad has built-in, modular, and extensible support for metadata in various
 formats. Metadata is extracted from a dataset and its content by one or more
 extractors that have to be enabled in a dataset's configuration. Extractors
-yield metadata in a JSON-LD-like structure that can be arbitrarily complex and
+yield metadata in a JSON-LD_-like structure that can be arbitrarily complex and
```
kyleam: Presumably you didn't intend to introduce an underscore here?

kyleam: Doh, got it. Thanks

yarikoptic:
Thank you @kyleam , will merge now

PR #2697, opened 2018-07-12
Removed two stale _broken files

yarikoptic:
Life is different now and not as simple.  Looking back at them I like the idea of
easily specifying common "harmonized" metadata, so may be some "procedure" could be
crafted for that purpose at some point.

Were found by searching for collections.datalad.org which is I am about to retire

Clearly shouldn't break anything, but submitting as a PR for visibility of this action. Will merge when tests pass


kyleam:
3 failed runs are from the known gitdb ref issue and a timeout.

yarikoptic:
thanks @kyleam ... kept whining on gitpython about the badref issue - we need a workaround (possibly just to disable gc at those calls)

PR #2708, opened 2018-07-21
ENH+BF: S3 and downloaders: support anonymous access + interface to add new providers

yarikoptic:
- [x] Basic and Digest HTTP Auth should work now
- [x]  In interactive session would now allow to establish a new provider configuration for which would get saved into a user directory
- [x] unittests for `process_www_authenticate` etc
- [x] Test at least for the major portions of the UI to enter the new provider/credential
- [ ] may be would include or just be part of integration test above to test Basic and Digest auth on https://jigsaw.w3.org/HTTP/Basic and https://jigsaw.w3.org/HTTP/Digest
- [x] may be more RFing for credentials (allow to make new one persistent etc)
so marking WIP for now

### DIFF
--- a/datalad/downloaders/base.py
+++ b/datalad/downloaders/base.py
@@ -30,6 +30,7 @@ from ..ui import ui
 from ..utils import auto_repr
 from ..dochelpers import exc_str
 from .credentials import CREDENTIAL_TYPES
+from ..support.exceptions import *
 
 from logging import getLogger
 lgr = getLogger('datalad.downloaders')
@@ -139,9 +140,13 @@ class BaseDownloader(object):
                 # assume success if no puke etc
                 break
             except AccessDeniedError as e:
-                lgr.debug("Access was denied: %s", exc_str(e))
+                if isinstance(e, AnonymousAccessDeniedError):
+                    access_denied = "Anonymous"
+                else:
+                    access_denied = "Authenticated"
+                lgr.debug("%s access was denied: %s", access_denied, exc_str(e))
                 exc_info = sys.exc_info()
-                access_denied = True
+
             except IncompleteDownloadError as e:
                 exc_info = sys.exc_info()
                 incomplete_attempt += 1
@@ -155,6 +160,12 @@ class BaseDownloader(object):
                 raise
 
             if access_denied:  # moved logic outside of except for clarity
+                # TODO: what if it was anonimous attempt without authentication,
+                #     so it is not "requires_authentication" but rather
+                #     "supports_authentication"?  We should not report below in
+                # _get_new_credential that authentication has failed then since there
+                # were no authentication.  We might need a custom exception to
+                # be caught above about that
                 if needs_authentication:
                     # so we knew it needs authentication
                     if used_old_session:
@@ -182,25 +193,32 @@ class BaseDownloader(object):
                                 raise AccessDeniedError(
                                     "Interface is not interactive and we were denied access."
                                     " Report to datalad developers")
-                        self._get_new_credentials(url)
+                        self._enter_credentials(
+                            url, denied_msg=access_denied, new=not self.credential)
                         allow_old_session = False
                         continue
                 else:  # None or False
                     if needs_authentication is False:
-                        # those urls must or should NOT require authentication but we got denied
-                        raise DownloadError("Failed to download from %s, which must be available without "
-                                            "authentication but access was denied" % url)
-                    else:
+                        # those urls must or should NOT require authentication
+                        # but we got denied
+                        raise DownloadError(
+                            "Failed to download from %s, which must be available without "
+                            "authentication but access was denied" % url)
+                    else:  # yoh: not sure if we ever get there since do not see
+                        # how could be None or any other non-False bool(False)
                         assert(needs_authentication is None)
-                        # So we didn't know if authentication necessary, and it seems to be necessary, so
-                        # Let's ask the user to setup authentication mechanism for this website
+                        # So we didn't know if authentication necessary, and it
+                        # seems to be necessary, so Let's ask the user to setup
+                        # authentication mechanism for this website
+                        # TODO: self._get_new_credentials(url)  but what do we do about
+                        # need_authentication
                         raise AccessDeniedError(
                             "Access to %s was denied but we don't know about this data provider. "
                             "You would need to configure data provider authentication using TODO " % url)
 
         return result
 
-    def _get_new_credentials(self, url):
+    def _enter_credentials(self, url, denied_msg, new=True):
         """Use when authentication fails to set new credentials for url
 
         Raises
@@ -208,8 +226,9 @@ class BaseDownloader(object):
         DownloadError
           If either no known credentials type, or user refuses to update
         """
-        title = "Authentication to access {url} has failed.".format(url=url)
-        if not self.credential:
+        title = "{msg} access to {url} has failed.".format(
+            msg=denied_msg, url=url)
+        if new:
             # No credential was known, we need to create an
             # appropriate one
             if not self.authenticator:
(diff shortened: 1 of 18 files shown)


datalad/downloaders/base.py
```diff
@@ -75,11 +75,12 @@ def __init__(self, credential=None, authenticator=None):
         authenticator: Authenticator, optional
           Authenticator to use for authentication.
         """
+        # Allow for anonymous connection if no credential is provided
```
kyleam: Is this comment misplaced?  The code immediately below is unchanged and doesn't check `credential`.  Perhaps it should be above the `if not credential` line below?

datalad/ui/base.py
```diff
@@ -31,10 +31,21 @@ def question(self, text,
         pass
 
     def yesno(self, *args, **kwargs):
+        # Provide some default sugaring
+        default = kwargs.pop('default', None)
+        if default is not None:
+            if default in {True}:
```
kyleam: If you're going to have only one value, why not `is True` here and `is False` below?

datalad/ui/tests/test_base.py
```diff
@@ -85,3 +88,6 @@ def nothing(x, k=1):
     @with_testsui(responses='a')
     def ask():
         assert_equal(ui.question('what is a?'), 'a')
+
+
+    ask()
```
kyleam: AFAICS, this fix and the cosmetic changes above are unrelated to yesno default change, which isn't actually covered in the tests.

datalad/downloaders/base.py
```diff
@@ -185,40 +194,67 @@ def access(self, method, url, allow_old_session=True, **kwargs):
                         if not ui.is_interactive:
                             lgr.error(
                                 "Interface is non interactive, so we are "
-                                "reraising: %s" % exc_str(e))
+                                "re-raising: %s" % exc_str(e))
```
kyleam: I think reraising was fine.

datalad/downloaders/base.py
```diff
@@ -228,42 +264,56 @@ def _enter_credentials(self, url, denied_msg, new=True):
...
+                                        " No authenticator is known, cannot "
+                                        "set "
+                                        "any credential")
+                # we could try to just
+                credential_type = self.authenticator.DEFAULT_CREDENTIAL_TYPE
+                if not credential_type:
+                    raise DownloadError(
+                        title +
+                        " %s does not have a default credential type, "
+                        "cannot authenticate"
```
kyleam: Why put "any credential" on a new line?

datalad/downloaders/providers.py
```diff
@@ -337,6 +367,119 @@ def get_provider(self, url, only_nondefault=False):
...
+        if url:
+            ri = RI(url)
+            for f in ('hostname', 'name'):
+                try:
+                    # might need sanitarization
+                    name = str(getattr(ri, f))
+                except AttributeError:
+                    pass
+        known_providers_by_name = {p.name: p for p in self._providers}
+        PROVIDERS_USER_DIR = self._get_providers_dirs()['user']
```
kyleam: This is local to this function.  Why the allcaps?

datalad/downloaders/providers.py
```diff
@@ -337,6 +367,119 @@ def get_provider(self, url, only_nondefault=False):
...
+                title="New provider name",
+                text="Unique name to identify 'provider' for %s" % url,
+                default=name
+            )
+            filename = pathjoin(PROVIDERS_USER_DIR, '%s.cfg' % name)
+            if name in known_providers_by_name:
+                if ui.yesno(
+                    title="Known provider %s" % name,
+                    text="Provider with name %s already known. Do you want to "
+                         "use it for this session? If 'No' - enter another name"
```
kyleam: This could be misinterpreted as "To answer 'NO', enter another name here".

datalad/downloaders/providers.py
```diff
@@ -337,6 +367,119 @@ def get_provider(self, url, only_nondefault=False):
...
+            elif path.lexists(filename):
+                ui.error(
+                    "File %s already exists, choose another name" % filename)
+            else:
+                break
+
+        url_re = re.escape(url)
+        while True:
+            url_re = ui.question(
+                title="New provider regular_expression",
```
kyleam: Drop the underscore?

yarikoptic: indeed. thanks!

datalad/downloaders/providers.py
yarikoptic: indeed...  I better move this trailing sentence to avoid confusion - I think it will be ok without it

datalad/downloaders/providers.py
yarikoptic: I finally found CAPSLOCK useful! ;-)  I guess leftover from when I hoped to have those just a process-wide constants... will lowercase

datalad/downloaders/base.py
yarikoptic: what `"any credential"`? ah -- way above! will fix and compress here a bit. thanks

datalad/downloaders/base.py
yarikoptic: ok, will change throughout the codebase... i.e. in 1 more place (a comment ;-))

datalad/downloaders/base.py
yarikoptic: I guess it was overall like a TODO comment reminder... I just removed it now

datalad/ui/base.py
kyleam: I find it less readable because I'm not used to seeing it.  As for the logic, I don't consider changing `is True` to `in {True, X}` if the need arises to be a maintenance burden.  And it seems you could use the same logic for pretty much any `is ATOM`, or any `== ATOM` for that matter.  Anyway, my comment was a "just curious" one as well.

datalad/ui/tests/test_base.py
kyleam: > This one just spotted

That's fine, but it'd be easy to put that in a separate commit.  When reviewing, I find it confusing when unrelated bits are tossed in (including cosmetic fixups that just add noise to the diffs).

datalad/downloaders/http.py
```diff
@@ -65,6 +65,17 @@
 __docformat__ = 'restructuredtext'
 
 
+def process_www_authenticate(v):
```
yarikoptic: Could you please add a short docstring to describe the purpose of the function and why output should be a list? (that is should provide value for the `supported_types`)

datalad/downloaders/tests/test_providers.py
```diff
@@ -146,3 +151,78 @@ def test_Providers_from_config__files(sysdir, userdir, dsdir):
...
+            providers.enter_new(url, auth_types=["http_basic_auth"])
+        auth_types()
+        ok_exists(op.join(providers_dir, "foo2.cfg"))
+
+        @with_testsui(responses=["foo3", "doesn't match", url_re, auth_type,
+                                 creds, "yes"])
+        def nonmatching_url():
+            providers.enter_new(url, auth_types=["http_basic_auth"])
+        nonmatching_url()
+        ok_exists(op.join(providers_dir, "foo3.cfg"))
```
yarikoptic: Lovely!
I guess the next one would be a similar one for `_enter_credentials`?

datalad/downloaders/http.py
kyleam: Uhh you wrote that code.  I don't know its purpose or why its output should be a list, but, sure, I could guess and add a docstring.  In the time it took you to "review" the function you wrote, you could have wrote that docstring.

datalad/downloaders/tests/test_providers.py
kyleam: I was hoping I had already fulfilled your request to help with your PR.

datalad/downloaders/http.py
yarikoptic: ah, damn me -- should've added a docstring indeed, will do

datalad/downloaders/tests/test_providers.py
yarikoptic: you did help! ok,I could continue whenever I get a chance, but not sure when it would be :-(

kyleam: I wasn't saying I wouldn't help more, just stating what I had hoped :)

I'll look into filling out the tests more, starting with `_enter_credentials`.

yarikoptic: thanks in advance! ;-)

yarikoptic:
@kyleam pushed two commits to "enable" access to the s3 buckets with period in their name (apparently needs special handling and now there is openneuro.org bucket)

yarikoptic:
Thank you very much @kyleam

yarikoptic:
Looks great. Let's see how it holds to real life testing

Issue #2721, opened 2018-07-31
wishlist: Disable repeated entry of "hidden" values


yarikoptic:
#### What is the problem?
ATM (since f7e0536685806b2de1c95216cc7eb9b2e26bc4f0) regardless of the purpose (e.g. to define new value or just provide a value for an existing one) hidden entries asked to be entered twice.

IMHO this is neither necessary nor typical UI -- I do not remember any other UI which would ask to enter previously defined password twice just to make sure user entered it correctly.  Such double-entry is typically done on entry/when defining new passwords.  I think we should allow entry once, and if entered incorrectly, authentication would fail. and we should offer to reenter.  And I believe that is how it behaved before and would be after
#2708 
Thoughts @mih since you are the one who introduced it -- what was your use case which shown the need in this verification?

PS that code will be RFed slightly in the upcoming PR but I will retain the double entry as a default for now

yarikoptic-gitmate:
[GitMate.io](https://gitmate.io) thinks possibly related issues are https://github.com/datalad/datalad/issues/1176 (Hidden Exception problem?), https://github.com/datalad/datalad/issues/749 ([wishlist] foreach command), https://github.com/datalad/datalad/issues/748 ([wishlist] save-state and "undo"), https://github.com/datalad/datalad/issues/1650 ([wishlist] run/rerun tandem), and https://github.com/datalad/datalad/issues/466 (wishlist:  publish --failed={raise,skip}).

mih:
Is this still a thing? I cannot recall having entered a password twice, but I rarely deal with this case anyways. Do you have a demo by any chance?

PR #2722, opened 2018-07-31
Improve support of IPython

### LINKED ISSUE #2719: IPython notebook interface is not considered interactive, trick with using password entry is not working
#### What is the problem?
We need (or even must asap) to provide a custom IPython notebook (web frontend) UI and tune up `is_interactive()` which ATM considers ipython env to be non interactive:

gkrellshoot_07-31-18_115527

IIRC the "tricky"/"nasty" part is that AFAIK there is no way to figure out if you are in the ipython web frontend or console... I will check it all out

related. 
- [ ] https://github.com/datalad/datalad/issues/1541 -- tqdm progressbar needs special "mode". Should be easy if we know for sure we are in web ipython/jupyter frontend
- [ ] https://github.com/datalad/datalad/issues/2063 -- only somewhat related to provide automagicIO handling

yarikoptic:
This pull request fixes #2719 
- [x] sets custom progress bars in the ipython session
- it will not display any progress bars if ran under `ipython console` (still would work under regular `ipython` terminal) and we have no way to differentiate.
- [x] partially addresses #2721  to not ask twice at least the hidden ones with choices given

### DIFF
--- a/datalad/ui/__init__.py
+++ b/datalad/ui/__init__.py
@@ -20,12 +20,25 @@ lgr.log(5, "Starting importing ui")
 from .dialog import (
     ConsoleLog,
     DialogUI,
-    IPythonWebUI,
+    IPythonUI,
     UnderAnnexUI,
     UnderTestsUI,
 )
 from .dialog import SilentConsoleLog
-from ..utils import is_interactive
+from ..utils import (
+    is_interactive,
+    get_ipython_shell,
+)
+
+KNOWN_BACKENDS = {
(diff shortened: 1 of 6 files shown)


yarikoptic:
@datalad/developers   if you could try on your normal (terminal) and ipythons (notebook/terminal) and say if it works -- would be much appreciated

try your secure or insecure inputs and gets of some data ;)

yarikoptic:
I think it is ok and worth merging ;-)

PR #2724, opened 2018-08-01
A variety of enhancements and fixes for metadata search

yarikoptic:
See individual commits please for the summaries

Please have a look @datalad/developers

PR #2725, opened 2018-08-01
ENH: report absent fields reported as N/A in output instead of not rendering the record at all

yarikoptic:
Since this one might controversial and still needs tests, filing it separately from #2724:
- [x] If result record is missing a value (absent field, or subfield in metadata), do not crash rendering it, but just print `N/A` for that value.  I wonder though if there should be config option to turn that behavior off since might be handy to see original crash at times
- [x] at least a basic test for the functionality

- shortened access syntax item moved the other into a separate issue https://github.com/datalad/datalad/issues/2760

### DIFF
--- a/datalad/interface/base.py
+++ b/datalad/interface/base.py
@@ -326,6 +326,11 @@ class nagen(object):
     def __init__(self, missing=NA_STRING):
         self.missing = missing
 
+    def __repr__(self):
+        cls = self.__class__.__name__
+        args = str(self.missing) if self.missing != NA_STRING else ''
+        return '%s(%s)' % (cls, args)
+
     def __str__(self):
         return self.missing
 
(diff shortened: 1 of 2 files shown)


yarikoptic:
@mih so what do you think?  or should we come up with a bit extended support for that?  I was thinking about making dedicated output format modes, so we have `format` and may be `format-NA`  with a dedicated new option field (`-F`?) so I could say `-f format -F '{...}'` to allow for current case where it just fails to render if a field is missing a value.

mih:
Let's talk about a few more general aspects first.

yarikoptic:
ok, will add a test or two and if no objects would vote to merge it without overthinking some "general aspects"

yarikoptic:
If no strong objections are expressed, I will merge it tomorrow - I am loosing control over all PRs staged across a variety of projects etc

PR #2731, opened 2018-08-03
BF: Internal cmd call, leave error handling to outer layer

mih:


yarikoptic:
if you say that that is how it should be, I guess it should be that way

Issue #2337, opened 2018-03-23
Field-selective value matching for egrep search


mih:
@akeshavan Want to find all 't1w' files across all datasets, and then report the URL we have on record for that.

The use case is to get this information exported once, hence it is pretty useless to build a search index. Moreover, the field value is a proper tag, hence no chunking or stemming is needed or helpful. Consequently, the egrep mode is the best way to implement this.

mih:
@yarikoptic any chance you could give this a try?

yarikoptic:
may be... can't promise though... I guess we still need to drill down that idea we had about yet another cmdline option to specify "--query-format"?  ATM we have per mode specific, but I guess it could be a number of them ("key-value", "plain") with defaults varying per each mode?

mih:
Yes, that sounds good.

mih:
Reassigning to a later release -- everyone is too busy...

mih:
It seems to me that https://pypi.org/project/pyparsing/ could provide a straightforward way to implement a minimal query language for `egrep` search, primarily to address the usecase of this issue.

Issue #2639, opened 2018-06-09
Limiting to 20 hits before filtering by type is "suboptimal"


yarikoptic:
#### What is the problem?
Didn't analyze in detail but just now got:
```shell
$> datalad --report-type file search --mode egrep MRI 
[INFO   ] Reached the limit of 20 top matches, there could be more which were not reported. 
action summary:
  search (ok: 20)
```
which is confusing/suboptimal at least to me-a-newbie-user

`egrep` doesn't have "syntax" to specify what kind of an entry to consider (file vs dataset), so the only way to filter is by report-type.  But that one

I am not sure if anything could be done given current architecture, but decided to check... may be if there were no "built-in" limit for egrep and other dummy search backends which would just keep yielding than theoretically limiting could be done (if needed at all) at the "process results" stage?

mih:
I think your analysis is correct. In the `egrep` case it would make sense to not have any result limiter.

PR #2735, opened 2018-08-03
NF: Field-selective search in egrep mode (fixes gh-2337)

mih:
This PR implements a number of changes for the `egrep` default search mode that should make it substantially more useful, without putting too much strain on novice users.

- [x] by default queries are case-sensitive now. An example documents how to change that in a query on a case-by-case basis.
- [x] search can be limited to specific keys, syntax described and documented in examples
- [x] multiple queries (list of expressions) are evaluated using AND to determine whether something is a hit
- [x] a single multi-field query (e.g. `pa*:findme`) is a hit, when any matching field matches the query
- [x] all matching key/value combinations across all (multi-field) queries are reported in the `query_matched` result field.
- [x] mode-specific search hit number limit: egrep has none now #2639

### DIFF
--- a/datalad/metadata/search.py
+++ b/datalad/metadata/search.py
@@ -420,6 +420,10 @@ class _WhooshSearch(_Search):
         self.idx_obj = idx_obj
 
     def __call__(self, query, max_nresults=None, force_reindex=False, full_record=False):
+        if max_nresults is None:
+            # mode default
+            max_nresults = 20
+
         with self.idx_obj.searcher() as searcher:
             wquery = self.get_query(query)
 
@@ -622,6 +626,9 @@ class _EGrepSearch(_Search):
     # --consider_ucn - search through unique content properties of the dataset
     #    which might be more computationally demanding
     def __call__(self, query, max_nresults=None, consider_ucn=False, full_record=True):
+        if max_nresults is None:
+            # no limit by default
+            max_nresults = 0
         query = self.get_query(query)
 
         nhits = 0
@@ -965,9 +972,10 @@ class Search(Interface):
         max_nresults=Parameter(
             args=("--max-nresults",),
             doc="""maxmimum number of search results to report. Setting this
-            to 0 will report all search matches, and make searching substantially
-            slower on large metadata sets.""",
-            constraints=EnsureInt()),
+            to 0 will report all search matches. Depending on the mode this
+            can search substantially slower. If not specified, a
+            mode-specific default setting will be used.""",
+            constraints=EnsureInt() | EnsureNone()),
         mode=Parameter(
             args=("--mode",),
             choices=('egrep', 'textblob', 'autofield'),
@@ -1012,7 +1020,7 @@ class Search(Interface):
     def __call__(query=None,
                  dataset=None,
                  force_reindex=False,
-                 max_nresults=20,
+                 max_nresults=None,
                  mode=None,
                  full_record=False,
                  show_keys=None,
--- a/datalad/metadata/search.py
+++ b/datalad/metadata/search.py
@@ -759,7 +759,7 @@ class _EGrepSearch(_Search):
 
     def get_query(self, query):
         query = assure_list(query)
-        simple_fieldspec = re.compile(r"(?P<field>\S+?):(?P<query>.*)")
+        simple_fieldspec = re.compile(r"(?P<field>\S*?):(?P<query>.*)")
         quoted_fieldspec = re.compile(r"'(?P<field>[^']+?)':(?P<query>.*)")
         query = [
             simple_fieldspec.match(q) or
@@ -771,6 +771,11 @@ class _EGrepSearch(_Search):
             {k: re.compile(v) for k, v in q.groupdict().items()}
             if hasattr(q, 'groupdict') else re.compile(q)
             for q in query]
+        # turn "empty" field specs into simple queries
+        # this is used to forcibly disable field-based search
+        # e.g. when searching for a value
+        query = [q['query'] if isinstance(q, dict) and q['field'].pattern == '' else q
+                 for q in query]
         return query
 
 
--- a/datalad/metadata/search.py
+++ b/datalad/metadata/search.py
@@ -849,7 +849,7 @@ class Search(Interface):
     *Mode: egrep/egrepcs*
 
     These search modes are largely ignorant of the metadata structure, and
-    simply performs matching of a search pattern against a flat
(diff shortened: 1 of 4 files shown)


datalad/metadata/tests/test_search.py
```diff
@@ -239,49 +244,68 @@ def test_within_ds_file_search(path):
...
+        # 'audio.format', 'mime:audio/mp3'),
+        # but this one works
+        ('autofield',
+         "'mime audio mp3'",
+         opj('stim', 'stim1.mp3'),
+         {'audio.format': 'mp3'}),
+        # TODO extend with more complex queries to test whoosh
+        # query language configuration
     ):
+        print('KKKKKKKKKKKKKKKKKK', query, mode)
```
yarikoptic: leftover I guess

datalad/metadata/search.py
```diff
@@ -743,10 +758,20 @@ def __init__(self):
             ))
 
     def get_query(self, query):
-        # cmdline args might come in as a list
-        if isinstance(query, list):
-            query = u' '.join(query)
-        return query.lower()
+        query = assure_list(query)
+        simple_fieldspec = re.compile(r"(?P<field>\S+?):(?P<query>.*)")
+        quoted_fieldspec = re.compile(r"'(?P<field>[^']+?)':(?P<query>.*)")
```
yarikoptic: I guess this should be adjusted to allow for saying explicitly that `:` is a part of "direct" (just `q`), non-fieldspec query.  I guess we could allow for it to be escaped `\`, so above ones could then become instead of `\S+?` a `\S*?[^\]` , and we would then need to "unscape" `\:` ?  but then what if we are looking for `\:`? (I guess would need a lookbehind that previous one is not `\` and then we convert `\\` to `\`
?

mih: I though about it, and I find a leading colon to be more natural to indicate ALL FIELDS. I.e. `:term: with colon` will search for `term: with colon`. and `:this` is identical to `this`. If one needs to search for something starting with a colon, this will do: `::term`

yarikoptic: Awesome

mih:
> This is a change in behavior right? I thought it was incencitive for the case in the basic search for years, it an I wrong?

Yes. The change is case-sensitive by default. Rational: It is easy to turn into case-insensitive, but not the other way round. In the former implementation there wasn't even a way to conduct a case sensitive search -- which is quite handy to be able to do.

yarikoptic:
Could you give me a realistic use case where case really matters?

mih:
Sure. Any kind of exact label where case encodes meaning. Another one are names: a person with name Mike might be easier to distinguish from a mike. Same logic why the cmdline tool grep has case-insensitive match as an *option*.

But the point is really what is the value of being unable to perform case sensitive only vs. both. Any realistic query has way more fundamental issues than casing. In general, with unknown metadata structure and type, one has little idea what to search for at all.

yarikoptic:
> Any kind of exact label where case encodes meaning, or broadens the namespace (e.g. a hash based ID that is generated like those in a shortened URL).

most likely case doesn't matter there as far as I see it

>  Another one are names: a person with name Mike might be easier to distinguish from a mike. 

And with many other names, if they aren't as "unique" as Mike that they could be a part of the bigger word you would need to anchor them around anyways, thus making query more complex and thus possibly adding an option to make it also case-sensitive.
Some times even acronyms might be cased interestingly different: Unicef vs proper UNICEF or AOL and Aol.

> But the point is really what is the value of being unable to perform case sensitive only vs. both.

Yes -- it is useful to be able to restrict casing, but IMHO it shouldn't come at a cost of making case-sensitive to be the default behavior for our default search mode!

> Same logic why the cmdline tool grep has case-insensitive match as an option.

I could argue that here we should aim not for `grep` [**] level of user-friendliness for the default/simplest search utility.  Rather more I see a default search datalad to be of usability level of "google", "apt", "emacs" [*] (;-)), where by default casing doesn't matter.   And in the majority of the usecases I used, `search` case didn't matter. When I am searching for "alzheimer" or "spin-echo"  or "neuroimaging" I do not care what was the casing, either it was a first word in the title (and thus for sure upper-cased) or just in a list of keywords (who knows - may be they are all upper-cased there).  If I was to look for NeuroImage as the journal, I would indeed want case sensitive search (isn't it a nice example for emacs [*]? ;-))

So overall, I see case-sensitivity as a useful feature to have (either as an additional option, or parameter in query, or automagic like emacs does when upper-case letter is present in the query), but I would strongly vote for maintaining current behavior of default search being case insensitive since it covers (much) more of use-cases over the ones where case-sensitivity is actually needed.

If we were to breed cmdline options I would have voted for having `--case-sensitive=auto,off,on` with `auto` providing emacs-style case sensitivity.   Would be handy for the majority of use-cases.

[*] in emacs search is case insensitive if there is no upper-case letter in the search query.  Kinda interesting usability twist.
[**] we aren't really using or implementing grep here AFAIK so mode name is already just that -- a name which shouldn't really impose grep's default behavior upon us

mih:
OK, so I'll make a new mode that implements the emacs behavior. I am fine with making that default. But I will keep the present one around, as I think being able to perform a case-sensitive match is a must.

kyleam:
> But I will keep the present one around, as I think being able to perform a case-sensitive match is a must.

Yes, I like the Emacs behavior, but I agree the user should be able to be able to perform a case-sensitive lowercase match (and Emacs of course lets you do that too by pressing <kbd>M-c</kbd>).

mih:
FTR (random dataset), new emacs-like search mode:

```
(datalad3-dev) mih@meiner /tmp/public (git)-[master] % datalad -c datalad.search.index-egrep-documenttype=all search T1 |tail -n3
search(ok): /tmp/public/sub-21/anat/sub-21_acq-r30_mod-T1w_defacemask.nii.gz (file)
action summary:
  search (ok: 56)
(datalad3-dev) mih@meiner /tmp/public (git)-[master] % datalad -c datalad.search.index-egrep-documenttype=all search t1 |tail -n3
search(ok): /tmp/public/sub-21/func/sub-21_task-orientation_acq-r30_run-10_bold.nii.gz (file)
action summary:
  search (ok: 336)
```

The difference between 336 and 56 are false positives with regard to the intent of the query.

mih:
Thx @kyleam

mih:
Seems everyone is overjoyed with the update. Setting merge counter to 8h.

mih:
Alrighty!

PR #2738, opened 2018-08-04
ENH progressbars for addurl

yarikoptic:
This pull request fixes https://github.com/datalad/datalad-container/issues/49

- It got even messier in there, and ideally requires some fixups on git-annex side: http://git-annex.branchable.com/todo/provide___39__file__39___in_--json-progress_record_for_addurl/
- I also disabled Total progress bar if we expect only one .  hopefully I didn't ruin something great ;-)

I guess would require manual testdriving since it is UI which we might not test very thoroughly automatically.  But on my attempts seems to be quite kosher.

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -1331,9 +1331,6 @@ class AnnexRepo(GitRepo, RepoInterface):
         if len(fetch_files) != len(files):
             lgr.debug("Actually getting %d files", len(fetch_files))
 
-        # options  might be the '--key' which should go last
-        options = ['--json-progress'] + options
-
         # TODO: provide more meaningful message (possibly aggregating 'note'
         #  from annex failed ones
         # TODO: reproduce DK's bug on OSX, and either switch to
@@ -1348,6 +1345,7 @@ class AnnexRepo(GitRepo, RepoInterface):
             # TODO: eventually make use of --batch mode
             jobs=jobs,
             expected_entries=expected_downloads,
+            progress=True,
             **kwargs
         )
         results_list = list(results)
@@ -2057,6 +2055,7 @@ class AnnexRepo(GitRepo, RepoInterface):
             out_json = self._run_annex_command_json(
                 'addurl',
                 opts=options + [files_opt] + [url],
+                progress=True,
                 log_online=True, log_stderr=False,
                 **kwargs
             )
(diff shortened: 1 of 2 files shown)


datalad/support/annexrepo.py
```diff
@@ -2057,7 +2057,10 @@ def add_url_to_file(self, file_, url, options=None, backend=None,
             out_json = self._run_annex_command_json(
                 'addurl',
                 opts=options + [files_opt] + [url],
-                log_online=True, log_stderr=False,
+                progress=True,
+                expected_entries={file_: None},
+                #log_online=True,
+                #log_stderr=False,
```
kyleam: Why leave commented out code?

datalad/support/annexrepo.py
```diff
@@ -2279,6 +2283,8 @@ def _run_annex_command_json(self, command,
         expected_entries : dict, optional
           If provided `filename/key: size` dictionary, will be used to create
           ProcessAnnexProgressIndicators to display progress
+        progress: bool, optional
+          Either request/handle --json-progress
```
kyleam: s/Either/Whether to/

datalad/support/annexrepo.py
```diff
@@ -3650,8 +3676,8 @@ def __call__(self, line):
 
         # so we have a progress indicator, let's deal with it
         action = j['action']
-        download_item = action.get('file') or action.get('key')
-        download_id = (action['command'], action['key'])
+        download_item = action.get('key') or action.get('file')
```
kyleam: "download_item" seems like a weird name at the level of `ProcessAnnexProgressIndicators`.  Regardless of where this is used at the moment, couldn't the item correspond to  an upload or a local copy or any other action on a file/key?

yarikoptic: indeed!  I guess it i more of an `action_item`, will change now. thanks

datalad/support/annexrepo.py
yarikoptic: dunno... dirty yarik. will remove

yarikoptic:
I've restarted a travis build (hopefully uploaded fresh git-annex resolves the issue) since failure looked unrelated.  I am afraid I did break something with this PR since my checks I had to add hint that something is behaving differently now.  But I am afraid that due to this "interactive" nature of the element here, we would only discover by trying it out.
I will merge tomorrow if no new problems emerge

PR #2759, opened 2018-08-09
Accumulator for metadata handling fixes

mih:


datalad/metadata/metadata.py
```diff
@@ -666,8 +666,17 @@ def _get_metadata(ds, types, global_meta=None, content_meta=None, paths=None):
             # we also want to have each unique value set always come
             # in a top-level list, so we known if some unique value
             # was a list, os opposed to a list of unique values
+
+            def _ensure_serializable(val):
+                if isinstance(val, ReadOnlyDict):
+                    return {k: _ensure_serializable(v) for k, v in iteritems(val)}
```
yarikoptic: Should you use may be val.__class__ here to retain original class which could be derived?

datalad/metadata/search.py
```diff
@@ -107,6 +107,7 @@ def _meta2autofield_dict(meta, val2str=True, schema=None, consider_ucn=True):
                     # tailored data
                     continue
                 srcmeta[uk] = _listdict2dictlist(umeta[uk]) if umeta[uk] is not None else None
+            meta[src] = srcmeta
```
yarikoptic: Please review/contribute/merge to https://github.com/datalad/datalad/pull/2724 which has an alternative fix for this which wouldn't conflict but duplicate the fix

datalad/metadata/metadata.py
mih: I don't think that is necessary.

yarikoptic: It wouldn't hurt and would make code more scalable happen we come up with derivative classes for ReadOnlyDict (e.g. ReadOnlyOrderedDict) and it would be hard(er) to find all the cases like this.  Let me patch for that and see if that breaks anything (anyways need to merge with master ;-) )

yarikoptic: or "screw it" - indeed probably not necessary ;)

yarikoptic:
looks green enough, merging

Issue #147, opened 2015-04-30
WindowsError: [Error 32] The process cannot access the file because it is being used by another process: 'C:\\tmp\\datalad_temp_outputs_od4enw'


yarikoptic:
```
WindowsError: [Error 32] The process cannot access the file because it is being used by another process: 'C:\\tmp\\datalad_temp_outputs_od4enw'
```
full trace
```
======================================================================
ERROR: datalad.tests.test_annexrepo.test_AnnexRepo_backend_option
----------------------------------------------------------------------
Traceback (most recent call last):
  File "c:\python27\lib\site-packages\nose\case.py", line 197, in runTest
    self.test(*self.arg)
  File "c:\buildslave\datalad-pr-virtualbox-dl-win7-64\build\datalad\tests\utils.py", line 196, in newfunc
    t(*(arg + (d,)), **kw)
  File "c:\buildslave\datalad-pr-virtualbox-dl-win7-64\build\datalad\tests\utils.py", line 250, in newfunc
    func(*(arg + (path, url,)), **kw)
  File "c:\buildslave\datalad-pr-virtualbox-dl-win7-64\build\datalad\tests\test_annexrepo.py", line 346, in test_AnnexRepo_backend_option
    ar.annex_addurls([url +'faraway'], backend='SHA1')
  File "C:\Python27\Lib\contextlib.py", line 24, in __exit__
    self.gen.next()
  File "c:\buildslave\datalad-pr-virtualbox-dl-win7-64\build\datalad\utils.py", line 275, in swallow_outputs
    adapter.cleanup()
  File "c:\buildslave\datalad-pr-virtualbox-dl-win7-64\build\datalad\utils.py", line 246, in cleanup
    rmtemp(self._out.name)
  File "c:\buildslave\datalad-pr-virtualbox-dl-win7-64\build\datalad\utils.py", line 135, in rmtemp
    os.unlink(f)
WindowsError: [Error 32] The process cannot access the file because it is being used by another process: 'C:\\tmp\\datalad_temp_outputs_od4enw'
```
probably we need a workaround which would just try to remove a few times with some delay -- it feels like a race condition somewhere :-/

yarikoptic:
I think it is a race condition with the gc which is to remove the handler instance we created -- "del"ing it explicitly prior unlink might help, looping within rmtemp could beneficial anyways

bpoldrack:
[...]

PR #2770, opened 2018-08-14
BFing direct mode:  "git commit file(s)" etc

yarikoptic:
Disclaimer - sits on top of #2724 which I intend to merge if tests pass

This pull request fixes # ... TODO ... there might be a few
- [x] (nearly all) test decorators now carry identical to their name attributes for nose, so you can easily run only a subset of those decorated tests.  Some are decorated further, e.g. `direct_mode` would point to tests somehow related to direct mode - either explicitly testing in direct mode, known to fail, or ignored. etc
- [x] `git annex` invocations do not get ` -c core.bare=False --work-tree=.` since they shouldn't/mustn't!
- [x] in direct mode:  for `git commit file(s)` under direct mode, we resort to `git commit` while first analyzing `file(s)` given:
   - if any is not staged - we puke
   - if some staged aren't among `file(s)` - we prepare a new index file, where we reset those which aren't to be committed
   - **gotcha/shortcoming** if there is a modified file, which is staged but in some previous shape (so there are newer changes in worktree), they would not get committed.  We could have analyzed index regarding either staged version is annexified or git - and then do `git add` or `git annex add` accordingly, but not sure if worth the hassle ATM.  At least this fix/behavior would get us those 95% of correct functionality in direct mode ;)
    - This current "index" trick workaround could later be superseded by git supporting `git commit --cached file(s)` invocation.  but that wouldn't eliminate the need for the analysis if we decide to do regarding files which aren't yet staged (thus might needing `git annex add`)
- [x] redecorate `@skip_direct_mode` into `@known_failure_direct_mode`
- [x] undecorate tests which no longer fail in direct mode (there should be a few)
- [ ] possibly a dedicated test for `git commit file(s)` (`datalad save file(s)`) behavior in the case of some files being staged and some not... but might not do it if coverage would indicate that all possible lines are tested now

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -2702,6 +2702,9 @@ class AnnexRepo(GitRepo, RepoInterface):
                careless=True, files=None, proxy=False):
         self.precommit()
 
+        if files:
+            files = assure_list(files)
+
         # Note: `proxy` is for explicitly enforcing the use of git-annex-proxy
         #       in direct mode. This is needed in very special cases, which
         #       might go away once we figured out a better way. In any case, it
@@ -2727,7 +2730,6 @@ class AnnexRepo(GitRepo, RepoInterface):
                 # used to fail, because absolute paths are used. Using annex
                 # proxy this leads to an error (path outside repository)
                 if files:
-                    files = assure_list(files)
                     if options is None:
                         options = []
                     for i in range(len(files)):
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -2777,8 +2777,26 @@ class AnnexRepo(GitRepo, RepoInterface):
             #       git-annex-proxy, which is needed for --update option for
             #       example.
 
+            if files:
+                # Raise FileNotInRepositoryError if `files` aren't tracked.
+                try:
+                    super(AnnexRepo, self).commit(
+                        "dryrun", options=["--dry-run", "--no-status"],
+                        files=files)
+                except CommandError as e:
+                    if not (self.is_direct_mode() and
+                            AnnexRepo._is_annex_work_tree_message(e.stderr)):
+                        raise
+                    # The git call may fail with
+                    #   fatal: this operation must be run in a work tree
+                    #   fatal: 'git status --porcelain=2' failed in submodule ...
+                    # But that error message doesn't matter for the purpose of
+                    # the "file is tracked" check. It won't be shown if there
+                    # is a pathspec error.
+                    lgr.debug("Ignoring commit --dry-run failure")
             try:
                 alt_index_file = None
+
                 direct_mode = self.is_direct_mode()
                 # we might need to avoid explicit paths
                 files_to_commit = None if direct_mode else files
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -56,6 +56,7 @@ from datalad.utils import _path_
 from datalad.utils import generate_chunks
 from datalad.utils import CMD_MAX_ARG
 from datalad.utils import assure_unicode, assure_bytes
+from datalad.utils import make_tempfile
 from datalad.support.json_py import loads as json_loads
 from datalad.cmd import GitRunner
 
@@ -198,6 +199,14 @@ class AnnexRepo(GitRepo, RepoInterface):
             else:
                 raise e
 
+        # Below while setting for direct mode workaround, we change
+        # _GIT_COMMON_OPTIONS.
+        # But we should not pass workarounds such as --worktree=.
+        # -c core.bare=False to git annex commands, so for their
+        # invocation we will keep and use pristine version of the
+        # common options
+        self._ANNEX_GIT_COMMON_OPTIONS = self._GIT_COMMON_OPTIONS[:]
+
         # check for possible SSH URLs of the remotes in order to set up
         # shared connections:
         for r in self.get_remotes():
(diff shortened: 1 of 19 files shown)


datalad/support/annexrepo.py
```diff
@@ -2770,10 +2776,63 @@ def commit(self, msg=None, options=None, _datalad_msg=False,
...
+                    # here even if they are not changed at all and could be ignored
+                    #files_changed_notstaged = files_notstaged.difference(changed_files[False])
+                    # if files_changed_notstaged:
+                    #     # To be safe(r), let's just blow for now!
+                    #     # We could be smart(er) and do all the git or annex add
+                    #     # but let's see if we run into those cases first
+                    #     import pdb; pdb.set_trace()
+                    #     raise RuntimeError(
+                    #         "To commit files in direct mode, please first git "
+                    #         "or git-annex add them first!  Files: %s"
```
kyleam: In case this commented code gets added back: Two "firsts".

datalad/support/annexrepo.py
```diff
@@ -2770,10 +2776,63 @@ def commit(self, msg=None, options=None, _datalad_msg=False,
...
+                    #         % ', '.join(files_changed_notstaged)
+                    #     )
+                    # Files which were staged but not among files
+                    staged_not_to_commit = changed_files[True].difference(files_set)
+                    if staged_not_to_commit:
+                        # Need an alternative index_file
+                        with make_tempfile() as index_file:
+                            alt_index_file = index_file
+                            # Reset the files we are not to be committed
```
kyleam: I can't parse this comment.

datalad/support/annexrepo.py
```diff
@@ -2770,10 +2776,63 @@ def commit(self, msg=None, options=None, _datalad_msg=False,
...
+                            self._git_custom_command(
+                                list(staged_not_to_commit),
+                                ['git', 'reset'],
+                                index_file=alt_index_file)
```
kyleam: Huh I expected a write-tree/read-tree call somewhere, or a copy on the current index.  AFAICS what you end up with is an index that only has the files in `staged_not_to_commit` (at HEAD's version).  IOW, you have all files in the working tree except for those in `staged_not_to_commit` staged for deletion, and modifications in `staged_not_to_commit` as unstaged changes.  Once you go back to .git/index, you'd have the modifications from `staged_not_to_commit` staged, but you'd also have "A"s for all the other files in the working tree.

datalad/support/gitrepo.py
```diff
@@ -1575,6 +1579,9 @@ def _git_custom_command(self, files, cmd_str,
         if check_fake_dates and self.fake_dates_enabled:
             env = self.add_fake_dates(env)
 
+        if index_file:
+            env['GIT_INDEX_FILE'] = index_file
```
kyleam: `env` may be `None` here.

kyleam: Fixed in 7ca7c1023f.

datalad/support/annexrepo.py
kyleam: As of cc0bcd663e73, this copies the current index.

kyleam:
As of ccd071b8d, the test failures are restricted to the direct mode runs.

kyleam:
I haven't clicked through every remaining Travis failure, but I think they're all triggered by the Sphinx warning from #2774.

kyleam:
> would you mind pushing an ammended or a new commit to trigger "rebuild"?

done

yarikoptic:
whoohoo -- travis seems to be ok! awesome. It had unrelated error for the crawler tests - restarted.  appveyor though isn't happy for some tests, including now the dreaded old "The process cannot access the file because it is being used by another process" (e.g. #147) and `test_create` tests failing with "Repo unexpectly dirty "

datalad/support/gitrepo.py
```diff
@@ -2307,17 +2314,43 @@ def count_objects(self):
                                     if len(item.split(': ')) == 2]}
         return count
 
+    def get_changed_files(self, staged=False, filter='', index_file=None):
+        """Return a dictionary with files and their status code
```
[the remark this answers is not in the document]
yarikoptic: re submodules: hm...  I was just trying to generlize/centralize the call from previous commands in `get_missing_files` and `get_deleted_files`.  Did I change their behavior somehow?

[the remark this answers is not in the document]
kyleam: > Did I change their behavior somehow?

No.

kyleam:
> test_create tests failing with "Repo unexpectly dirty "

All right, after some unpleasant setup and debugging on smaug's windows vm, I think I know what's failing in these tests.  The files to add are coming in like `\\full\\path\\to\\repo\\.datalad\\config`, and then `_normalize_path` converts them to something like `.datalad\\config`.  All's fine in the Windows world at this point, I guess.  The problem arises from when we check this list against the files returned by `repo.get_changes_files`, which look like `.datalad/config`.  So any path that is not in the root of the repo doesn't get committed.

Obviously, we'll need to harmonize those sets, but I'll have to review the Windows path discussions to figure out who should change.

yarikoptic:
Oh, so Windows build of git annex is without mine support @joeyh? Then we might need to skip this test on Windows. Ideally (not here I guess) we should add somehow/somewhere (externals or AnnexRepo attributes) information about supported by annex optical features and then skip dynamically

yarikoptic:
heh, tests fail since Germans did massive RF to avoid `import *` via the flood of `from datalad.tests.utils import ...` lines in ad8ce546672e1ccf1ce04a03cccd98556228679a and that somehow only now got to master, and now test doesn't get its `opj` and should use `op.join`.  I will merge master into, and make it be used but also will use `datalad.support.path` not `os.path` since that is what we should use generally now AFAIK for workarounds for unicodes etc

kyleam:
> and that somehow only now got to master

It was just that all the old code was already adjusted for the rename.  I added an `opj` in this series, and the merge of course didn't take care of that

> I will merge master into

Sorry, pushed around the same time as your message came in.

> and make it be used but also will use datalad.support.path not os.path since that is what we should use generally now AFAIK for workarounds for unicodes etc

I just used os.path, so feel free to fix it up how you'd like.  But there is **a lot** of datalad code that uses os.path.

PR #2712, opened 2018-07-24
BF: workaround for BadName - precommit + fetch again if fails

yarikoptic:
This pull request hopefully fixes #2550

I guess we might want to rerun tests battery a few times to see if failure doesn't creep in.  Otherwise I could not find a reliable trigger for it

PR #2777, opened 2018-08-23
WIP: Removing some "failing in v6" tests

yarikoptic:
Should pass on that run (temporarily enabled) where we use devel annex and now test on v6

### DIFF
--- a/datalad/distribution/tests/test_clone.py
+++ b/datalad/distribution/tests/test_clone.py
@@ -219,7 +219,6 @@ def test_clone_isnot_recursive(src, path_nr, path_r):
 
 
 @slow  # 23.1478s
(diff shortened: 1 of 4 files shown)


yarikoptic:
cool, so new git-annex doesn't break anything, I will upload it to neurodebian proper now and strip this PR of unneeded last commit, and merge when tests pass again.
After that we would still have 32 `@known_failure_v6` worth looking at the underlying reason and let @joeyh know

kyleam:
The BadName reappears ([again](https://github.com/datalad/datalad/pull/2712#issuecomment-410383369)): https://travis-ci.org/datalad/datalad/jobs/419745297

yarikoptic:
Passes now -- merging!

Issue #2787, opened 2018-08-30
cannot "datalad install" within subsubdir submodule (?)


yarikoptic:
```shell
hopa:/tmp
$> datalad install http://github.com/MRtrix3/mrtrix3
[INFO   ] Cloning http://github.com/MRtrix3/mrtrix3 to '/tmp/mrtrix3' 
install(ok): /tmp/mrtrix3 (dataset)

$> cd mrtrix3/testing/data

$> datalad install .
[INFO   ] Cloning http://github.com/MRtrix3/mrtrix3/testing/data to '/tmp/mrtrix3/testing/data' 
[WARNING] Something went wrong while installing testing/data: [Errno 2] No such file or directory [base.py:clone_from:979] 
[ERROR  ] Installation of subdatasets /tmp/mrtrix3/testing/data failed with exception: InstallFailedError: 
Failed to install dataset from any of: ['http://github.com/MRtrix3/mrtrix3/testing/data', 'https://github.com/MRtrix3/test_data'] [get.py:_install_subds_from_flexible_source:182] [install(/tmp/mrtrix3/testing/data)] 
[ERROR  ] [Errno 2] No such file or directory [posixpath.py:abspath:364] (OSError)
```
works ok with stock submodule update ran within root of the repo:
```
$> cd ../../

$> git submodule update --init
Submodule 'testing/data' (https://github.com/MRtrix3/test_data) registered for path 'testing/data'
Cloning into '/tmp/mrtrix3/testing/data'...
Submodule path 'testing/data': checked out '6e6196af03e27948154f1f11285aa11d35af4a36'

```

datalad 0.10.2-207-g6b829527

yarikoptic-gitmate:
[GitMate.io](https://gitmate.io) thinks possibly related issues are https://github.com/datalad/datalad/issues/558 (datalad install -d . subds didn't really register it as a submodule to current git), https://github.com/datalad/datalad/issues/405 (datalad install of a subdataset into a dataset fails since provides None to git submodule call), https://github.com/datalad/datalad/issues/2551 (datalad diff would crash on an uninstalled submodule), https://github.com/datalad/datalad/issues/884 (datalad install   no longer can take multiple args), and https://github.com/datalad/datalad/issues/2211 (datalad install doesn't like gvfs).

yarikoptic:
also note that INFO message is misleading when I try to install from under `data/`:
```shell
$> datalad install data 
[...]

PR #2788, opened 2018-08-30
ENH+BF: Do not fail reporting if curdir is gone, do not rmdir the directory upon failed clone attempt

yarikoptic:
Addresses #2787

### DIFF
--- a/datalad/support/path.py
+++ b/datalad/support/path.py
@@ -12,9 +12,10 @@ One of the reasons is also to robustify operation with unicode filenames
 """
 
 # TODO: RF and move all paths related functions from datalad.utils in here
+import os
+import os.path as op
 
 from functools import wraps
-import os.path as op
 from ..utils import (
     assure_bytes,
     getpwd,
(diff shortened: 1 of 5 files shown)


datalad/support/path.py
```diff
@@ -48,3 +51,18 @@ def wrapped(*args, **kwargs):
...
+    """A helper which would not fail if p is relative and we are in non-existing directory
+
+    It will rely on getpwd, which would rely on $PWD env variable to report
+    the path.  Desired for improved resilience during e.g. reporting as in
+    https://github.com/datalad/datalad/issues/2787
+    """
+    try:
+        return abspath(p)
+    except OSError as exc:
+        if "No such" in str(exc) and not isabs(p):
```
kyleam: Eh I dislike relying on matching a particular string in the exception message, though I know there are already lots of case where we do it in the codebase.  In this case, couldn't you `getpwd()` and raise unless it *doesn't* exist and `not isabs(p)`?

yarikoptic: Ok, will do shortly

yarikoptic: Although it would make reraising original exception more cumbersome iirc

yarikoptic:
ok, approved is approved! ;)

PR #2790, opened 2018-09-04
ENH: appveyor - add a run for v6

yarikoptic:


yarikoptic:
woohoo - given that it took twice longer than default run, I guess the v6 mode was enabled, and nothing (new) has failed -- SUCCESS!! So I will merge it and then submit a new TEMP one enabling the entire battery just to see how we are doing ;-)

PR #2791, opened 2018-09-04
ENH: run all tests on appveyor

yarikoptic:
Doesn't hurt to try to assess where are we standing.
Unlikely to pass, thus unlikely to be merged ;-)


yarikoptic:
ran out of time, should get away from this Windows "goodness".  I will submit a separate PR with few improvements, and the main hurdle now I think is due to the "[WinError 6] The handle is invalid" spit out from GitPython's code where it tries to Popen:  https://github.com/gitpython-developers/GitPython/blob/master/git/cmd.py#L724 , and probably (just a hinch) it relates to
`close_fds=is_posix, # unsupported on windows`
argument to it where probably previous `git` processes linger around. 
Reminder that now we have `DATALAD_ASSERT_NO_OPEN_FILES` env variable, which if you set to `pdb` would jump into interactive session to troubleshoot on what is going on.
Possibly related issue against GitPython is https://github.com/gitpython-developers/GitPython/issues/722 but suggested resolution doesn't work for us (at least in those tests which I've tried).

yarikoptic:
"one more time"!

yarikoptic:
current state: `FAILED (SKIP=260, errors=63, failures=39`. Majority (if not all) of obvious "open files" should be resolved, and #2806 delivers those to master.

yarikoptic:
need to restart this one whenever a new build of git-annex for windows comes out.  We need to add more tests into the appveyor run, at least for v6 mode of operation where now hopefully more would pass.

yarikoptic:
might eventually come back

PR #2800, opened 2018-09-07
Again test -devel version of git-annex

yarikoptic:


yarikoptic:
there is still a test which reports v6 issue being fixed while running on travis, but can't reproduce locally. otherwise all kosher, closing for now

PR #2806, opened 2018-09-08
BFs and RF (Windows): more precommits etc to avoid operation on held up files/directories

yarikoptic:
This is absorbed from #2791 which keeps running all the tests under windows.

Notable change is that we should start using `@known_failure_windows` for the tests which potentially should work on Windows.


yarikoptic:
I resolved conflict locally and still have 
`FAILED (SKIP=94, errors=6, failures=3)`
for the `DATALAD_REPO_DIRECT=yes DATALAD_TESTS_SSH=1 DATALAD_ASSERT_NO_OPEN_FILES=assert ` run to figure out, and then I will push the update here (and incorporate into #2791 )

yarikoptic:
Thank you @kyleam ! I have restared one travis run which failed at npm install stage.. will merge whenever it is all green

PR #2815, opened 2018-09-12
BF: remove trailing period while matching a mesage from git

yarikoptic:
with git 2.19 there is no trailing period in "did not match any file(s)"


yarikoptic:
I've nailed it! ;)

PR #2817, opened 2018-09-13
WIP: test -devel version of git-annex 

yarikoptic:
testing annex version which should resolve ssh access to systems with older git-annexes


yarikoptic:
Seems to be kosher, only that upstream annex build locale issue, so I will upload the beast later today

yarikoptic:
uploaded and buildbots updated

yarikoptic:
ok, fresh git-annex which resolves problem with ssh access to content on servers with older git-annex seems to be operating correctly.  uploading to neurodebian proper

PR #2840, opened 2018-09-17
BF: annex_info - be robust to "unknown" size

### LINKED ISSUE #2839: annexrepo.py:repo_info: invalid literal for int() with base 10: 'unknown'
#### What is the problem?

```terminal
datalad create-sibling --recursive --existing reconfigure -s hydra-backup hydra:/backup/users/contematto/thesis

[INFO   ] Connecting ...
[INFO   ] Considering to create a target dataset /idata/DBIC/castello/thesis at /backup/users/contematto/thesis of hydra
[INFO   ] Target path /backup/users/contematto/thesis already exists. And it fails to rmdir (rmdir: failed to remove ���/backup/users/contematto/thesis���: Directory not empty). Will only reconfigure
[INFO   ] Updating dataset '/idata/DBIC/castello/thesis' ...
[ERROR  ] invalid literal for int() with base 10: 'unknown' [annexrepo.py:repo_info:2591] (ValueError)
```

It might be because of the following `available local disk space: unknown`:

```terminal
(datalad) [castello@discovery thesis]$ git annex info
repository mode: indirect
trusted repositories: 0
semitrusted repositories: 6
   < ... >
untrusted repositories: 0
transfers in progress: none
available local disk space: unknown
local annex keys: 0
local annex size: 0 bytes
annexed files in working tree: 0
size of annexed files in working tree: 0 bytes
bloom filter size:
  bloomfilter capacity too large to represent; falling back to sane value
32 mebibytes (0% full)
backend usage:
```

```
(datalad) [castello@discovery thesis]$ datalad --version
datalad 0.10.3.1

(datalad) [castello@discovery thesis]$ git annex version
git-annex version: 6.20180626-gdf91a5cff
```

yarikoptic:
Reported in https://github.com/datalad/datalad/issues/2839
Local disk space might be "unknown" so we must not blindly
convert it to int.  I wondered if we should keep original value
("unknown") but because in other place we already get it as None
by default, I thought that it would be better just to log at debug
and assign a None

This pull request fixes #2839

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -2588,7 +2588,12 @@ class AnnexRepo(GitRepo, RepoInterface):
         info = json_records[0]
         for k in info:
             if k.endswith(' size') or k.endswith(' disk space') or k.startswith('size of '):
-                info[k] = int(info[k].split()[0])
+                size = info[k].split()[0]
+                if size.isdigit():
+                    info[k] = int(size)
+                else:
+                    lgr.debug("Size %r reported to be %s, setting to None", k, size)
+                    info[k] = None
         assert(info.pop('success'))
         assert(info.pop('command') == 'info')
         return info  # just as is for now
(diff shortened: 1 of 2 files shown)


kyleam:
Set up of one test failed.  Restarted.

PR #2707, opened 2018-07-17
BFs (workarounds) for Windows: untaint DataLadRIs, use getcwd for PWD

yarikoptic:
This pull request fixes #2643 , fixed #2700 

Ah, also while troubleshooting that git-annex compatibility issue, added to WTF output the maximal filename length in current path (yes - it just creates a bunch of files until it can't, might take awhile if on the floppy)

PR #2847, opened 2018-09-19
BF: Crash on import with missing colorama (on win)

mih:


mih:
```
C:\Users\Pussy Galore\datalad>python
Python 3.6.6 (v3.6.6:4cf1f54eb7, Jun 27 2018, 02:47:15) [MSC v.1900 32 bit (Intel)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import datalad
  File "<stdin>", line 1, in <module>
  File "C:\Users\Pussy Galore\datalad\datalad\__init__.py", line 69, in <module>
    exc_str(e))
NameError: name 'e' is not defined
```

datalad/__init__.py
```diff
@@ -22,6 +22,11 @@
     import random
     random.seed(_seed)
 
+from .config import ConfigManager
+cfg = ConfigManager()
+
+from .log import lgr
+
```
yarikoptic: I think there might be a side effect of colorama not overloading sys.stdout then if lgr imported before.

In general, order of imports in `__init__.py` is "fragile" so I would advice to not mess with it unless really needed, and otherwise delay until actual used.  e.g. here you could leave cfg where it was, and add additional from .log import lgr right before your lgr.warning before .

datalad/__init__.py
```diff
@@ -49,25 +61,12 @@
...
 ssh_manager = SSHManager()
 atexit.register(ssh_manager.close, allow_fail=False)
-
-if on_windows and colorama is None:
-    from datalad.dochelpers import exc_str
-
-    lgr.warning(
-        "'colorama' Python module missing, terminal output may look garbled ["
-        "%s]",
-        exc_str(e))
```
yarikoptic: so `e` variable was eliminated from `locals()` even though above `except ... as e` was triggered?

kyleam:
the fine product of [self-dictatorship-while-germans-rest](https://github.com/datalad/datalad/pull/2707#issuecomment-406696789)

datalad/__init__.py
kyleam: That pattern doesn't work in Python 3.  Try this in py2 and py3:

```python
try:
    raise ValueError("bbbbbb")
except ValueError as e:
    pass
raise e
```

mih:
Changed my attitude to this issue. Why should we even worry about colorama, it is a tiny dependency. If things are so fragile, let's not warn and deal with absence, but ignore the error and declare a proper dependency.

setup.py
```diff
@@ -72,7 +73,9 @@ def findsome(subdir, extensions):
         'patool>=1.7',
         'six>=1.8.0',
         'wrapt',
-    ] + pbar_requires,
+    ] +
+    pbar_requires +
+    (['colorama'] if os.name == 'nt' else []),
```
kyleam: I believe you could also do 'colorama;platform_system=="Windows"', which I wasn't aware of until recently.  Not sure if there is any reason to prefer one or the other.

yarikoptic: We use
```
platform_system = platform.system().lower()
on_windows = platform_system == 'windows'
```
so into better to be consistent either way

yarikoptic:
I do so I guess I will now

PR #2864, opened 2018-09-21
BF: account for possible decorations around filenames while chunking cmdline

### LINKED ISSUE #2863: metadata handling manages to hit filename length limit
gkrellshoot_09-21-18_154105

yarikoptic:
Discovered on Windows, see https://github.com/datalad/datalad/issues/2863 ,
where error message was suggesting just getting over the maximum filename
size, but it was not it.  Accounting with at least +3 resolved the issue.
Here we do more to account for all the rest to be even safer

Fixes #2863

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -2328,7 +2328,16 @@ class AnnexRepo(GitRepo, RepoInterface):
             else:
                 files = assure_list(files)
                 maxl = max(map(len, files))
-                chunk_size = CMD_MAX_ARG // maxl
+                chunk_size = max(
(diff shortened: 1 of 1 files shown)


datalad/support/annexrepo.py
```diff
@@ -2328,7 +2328,16 @@ def _run_annex_command_json(self, command,
             else:
                 files = assure_list(files)
                 maxl = max(map(len, files))
-                chunk_size = CMD_MAX_ARG // maxl
+                chunk_size = max(
```
kyleam: I wonder why we only do this chunking at the `_run_annex_command_json` level.  Wouldn't the same limits apply to `_run_annex_command`?

PR #2870, opened 2018-09-24
A number of fixes from a quality time with Windows

yarikoptic:
- do not misdetect Windows paths for sshri
- enable `@with_testrepo` while on Windows and in default v6 mode for annex repos


bpoldrack:
Re ssh ri:
Shouldn't it be `is_windows_path` (called in `network.py` line 324), that is to be fixed for this issue?

yarikoptic:
it is not a "classical" windows  path, it is the abomination from those shells on Windows:

```
$> python -c 'from datalad.support.network import is_windows_path; print(is_windows_path("/C:/f/s1"))'
False

$> python -c 'from datalad.support.network import is_windows_path; print(is_windows_path("C:/f/s1"))' 
True
```
because of them we already have also special treatment of `///` already since leading `/` would be stripped by them for their own conventions

bpoldrack:
Sure. But the logical flow here should be to consider it a windows path and therefore return 'file', right? I just think that figuring it's a path after having `is_windows_path` return `False` leads to really hard to understand code. If we need to add things here for we discover more flaws of RI detection this is gonna get messy.

yarikoptic:
FWIW, on a sample run on our virtualbox that test (ran in isolation) didn't stall...

yarikoptic:
wanted to restart appveyor, didn't find a better way but just to rewrite last commit. force pushed

yarikoptic:
consistent "datalad.distribution.tests.test_dataset.test_is_installed ... " stall on appveyor -- will need to enjoy their interactive session to figure it out

yarikoptic:
eh, appveyor is still cursed with the stalling issue in test_dataset, and only in `DATALAD_REPO_VERSION=6` mode... needs to be debugged

yarikoptic:
one of the core problems with testing on Windows seems to be a regression of [not being able to addurl file:/// urls](http://git-annex.branchable.com/bugs/Unable_to_addurl_file__58____47____47____47___on_Windows/?updated) which we use in our `with_testrepos`

mih:
Closing due to inactivity.

Issue #2755, opened 2018-08-09
Convention for generator methods in *Repo classes


mih:
The `*Repo` classes are not really intended to be the user-facing interface, but historically they offer some convenience functionality (e.g. path normalization and type conversion). Some if this turned into more of a problem than an aid. For example, when results are forcibly gathered into a list, thereby preventing any benefits of implementing top-level commands as generator.

In #2749 I started working on rectifying this, but rewriting the low-level methods to work as generators. I have added `add_()` as the generator for `add()`, while keeping the old one around. The motivation for the latter was to keep the tests/old code working.

However, the Q is, whether we should use this as a common convention to ease use even for methods that are only available as generators now (e.g. `set_metadata()`).

Please tell me what you think.

yarikoptic-gitmate:
[GitMate.io](https://gitmate.io) thinks possibly related issues are https://github.com/datalad/datalad/issues/299 (__repr__ for *Repo classes (and others)), https://github.com/datalad/datalad/issues/2488 (Even with #2417 cooked *Repo.add() still cannot become generators), https://github.com/datalad/datalad/issues/2546 (We need a cross-platform convention for paths), https://github.com/datalad/datalad/issues/1323 (get_subdatasets  should become a generator), and https://github.com/datalad/datalad/issues/1200 (Better exception(s) for paths outside of a repo).

driusan:
I recently (yesterday) learned `set_metadata()` is a generator while writing datalad/datalad-crawler#13. As an external developer, I don't think having an `_` convention would have helped since I wouldn't have known the convention. A clear statement in the doc block (rather than having to trace through the code to find a yield statement) would have been more useful for me.

kyleam:
> A clear statement in the doc block (rather than having to trace through the code to find a yield statement) would have been more useful for me.

Isn't this already the case?

https://github.com/datalad/datalad/blob/d0d332c9b5e17cd6b97841697d188aacd55f870c/datalad/support/annexrepo.py#L3232-L3235

kyleam:
> However, the Q is, whether we should use this as a common convention to ease use even for methods that are only available as generators now (e.g. set_metadata()).

I vote no.  Like @driusan, I'm satisfied as long as its mentioned in the docstring.

mih:
Let me clarify: Yes, it should be in the docstring (like in the example above). However, I was more thinking along the lines of (always) providing a companion function that works like a generate (with `_` suffix), but keep the non-generator that we already have in general.

[...]

Issue #2015, opened 2017-12-08
Super slow addition of an existing subdaset within a top of "rich" hierarchy


yarikoptic:
#### What is the problem?
may be it is echo of #1467 , may be already fixed in master (I was using 0.9.x at this point).

```shell
(dev)2 11161.....................................:Fri 08 Dec 2017 12:05:17 AM EST:.
(git)smaug:/mnt/btrfs/datasets/datalad/crawl[master]git
$> git status --ignore-submodules=dirty
On branch master
Your branch is ahead of 'datalad-public/master' by 1 commit.
  (use "git push" to publish your local commits)

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

        modified:   abide2 (new commits)
        modified:   indi (new commits)

Untracked files:
  (use "git add <file>..." to include in what will be committed)

        abide-TODOs
        abide/
        hbn-TODOs
        hbn/
        princeton-dataspace/

no changes added to commit (use "git add" and/or "git commit -a")
(dev)2 11162.....................................:Fri 08 Dec 2017 12:05:19 AM EST:.
(git)smaug:/mnt/btrfs/datasets/datalad/crawl[master]git
$> datalad add -d . -m "adding abide dataset" abide

add(ok): abide (dataset) [added new subdataset]
add(notneeded): abide (dataset) [nothing to add from /mnt/btrfs/datasets/datalad/crawl/abide]
add(notneeded): .gitmodules (file) [already included in the dataset]
save(ok): /mnt/btrfs/datasets/datalad/crawl (dataset)
action summary:
  add (notneeded: 2, ok: 1)
  save (ok: 1)
datalad add -d . -m "adding abide dataset" abide  4.65s user 25.38s system 8% cpu 6:00.18 total
(dev)2 11163.....................................:Fri 08 Dec 2017 12:11:39 AM EST:.
```
so it took 6 minutes (!) for a simple *add a new submodule*.  I bet it was due to the fact that again full recursive listing of submodules was invoked to decide on the destiny of the path. 

To reproduce -- just go to one of those snapshots of the entire collection and try to add new pre-populated datasets.

#### What version of DataLad are you using (run `datalad --version`)? On what operating system (consider running `datalad plugin wtf`)?
0.9.1-137-g104c7864  in a bugfix branch on top of 0.9.x

mih:
[...]

Issue #1874, opened 2017-10-05
a simple create is way too slow if invoked within a "rich" hierarchy


yarikoptic:
#### What is the problem?

whatever should take a second top, takes a minute (or even two).  Thus for 1000 datasets, I will need to wait about 16-32  hours just to get them created :-(

#### What steps will reproduce the problem?
take some good hierarchy of directories/files (in my case a tree inbox/year/date/accession/study with bunch of DICOMs) and create a hierarchy of datasets at each of those levels, using `datalad create -d inbox $PWD/inbox/year/date/accession/study` (or at any level above) command.  I guess for sorting the paths or smth datalad does 'git diff' or 'status' which takes awhile... or may be something else. dunno yet

Here is a complete script
```bash
#!/bin/bash

set -eu

# helper since we blow if there is already a dataset there
function createds() {
 p="$1"
 [ -e "$p/.datalad" ] || \
   datalad create -d inbox --force --no-annex $PWD/$p
}

datalad create --force --no-annex inbox;
for yds in inbox/*; do   # year
  createds $yds;
  for mds in $yds/*; do   # month
      createds $mds;
      for dds in $mds/*; do   # day
         createds $dds;
         for ads in $dds/*; do   # accession
             [ -e "$ads/.datalad" ] || datalad create -d inbox --force  $PWD/$ads;
             # finally add data
             datalad add -d $ads -m "adding all the data" $PWD/$ads/*;
             # tune up configuration per @mih recommendation
             git -C $ads config --file .datalad/config --replace-all metadata.nativetype dicom
             git -C $ads config --file .datalad/config --replace-all metadata.aggregate-content-datalad-core no
             git -C $ads config --file .datalad/config --replace-all metadata.store-aggregate-content no
             # ATM can't do create with enh-metadata option, so we will do in two stages
             # datalad aggregate-metadata -d $ads
         done;
      done
   done
done
```
#### What version of DataLad are you using (run `datalad --version`)? On what operating system (consider running `datalad plugin wtf`)?

0.9.1 atm

edit 1:
yeap:
```
2017-10-05 13:56:58,437 [DEBUG  ] Committing via direct call of git: ['git', 'commit', '-m', '[DATALAD] new dataset']
[...]

PR #2872, opened 2018-09-25
Removing obstacles for generator behavior in low-level code

### LINKED ISSUE #2488: Even with #2417 cooked *Repo.add() still cannot become generators
because the `normalize_paths()`decorator requires that any return value has `__len__`.

It looks as if the decorator needs to go away completely.

This is the follow-up to #2417.

@bpoldrack please share your thoughts on how important this bit still is with the current concept. It would be very helpful to have all these low-level function not being blockers to high-level generators (which is the fundamental paradigm that we are using). I count 24 functions were this is used -- exclusively limited to `annex/gitrepo.py`.

mih:
This pull request shall eventually fix #2488

Current status of RF:
- [x] low level code and its own tests
- [ ]  usage of low-level code in high-level tests

Goal:
- Remove any magic file path handling, including:
  - magic conversion of list return values into single items when input path was a single item
  - resolution of paths relative to the curdir
- for now keep the ability to supply a single path as input, when a list is actually expected further in, but eventually this should be removed, to clearly see what code is capable/optimized for processing multiple inputs vs a single one

Rational:
All look-ups and conversions should be done by top-level code, once.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,6 @@ env:
     - COVERAGE=coverage
     - DATALAD_DATASETS_TOPURL=http://datasets-tests.datalad.org
   matrix:
-    - DATALAD_REPO_DIRECT=yes
 ##1    - DATALAD_REPO_VERSION=6
     - DATALAD_REPO_VERSION=5
 
@@ -153,14 +152,6 @@ matrix:
 #    - DATALAD_TESTS_KNOWNFAILURES_PROBE=yes
 #    - DATALAD_TESTS_SSH=1
 #    - _DL_CRON=1
-  - python: 3.5
-    # test whether known direct mode failures still fail
-    env:
-    - DATALAD_REPO_DIRECT=yes
-    - DATALAD_TESTS_KNOWNFAILURES_SKIP=no
-    - DATALAD_TESTS_KNOWNFAILURES_PROBE=yes
-    - DATALAD_TESTS_SSH=1
-    - _DL_CRON=1
 
   - python: 3.5
     # test with extension datalad-crawler
@@ -196,13 +187,6 @@ matrix:
 #    - DATALAD_TESTS_KNOWNFAILURES_PROBE=yes
 #    - DATALAD_TESTS_SSH=1
 #    - _DL_CRON=1
-  # test whether known direct mode failures still fail
-  - env:
-    - DATALAD_REPO_DIRECT=yes
-    - DATALAD_TESTS_KNOWNFAILURES_SKIP=no
-    - DATALAD_TESTS_KNOWNFAILURES_PROBE=yes
-    - DATALAD_TESTS_SSH=1
-    - _DL_CRON=1
 
 # Causes complete laptop or travis instance crash atm, but survives in a docker
 # need to figure it out (looks like some PID explosion)
--- a/datalad/auto.py
+++ b/datalad/auto.py
@@ -279,11 +279,11 @@ class AutomagicIO(object):
         try:
             # might fail.  TODO: troubleshoot when it does e.g.
             # datalad/tests/test_auto.py:test_proxying_open_testrepobased
-            under_annex = annex.is_under_annex(filepath, batch=True)
+            under_annex = annex.is_under_annex(filepath, batch=True)[filepath]
         except:  # MIH: really? what if MemoryError
             under_annex = None
         # either it has content
-        if (under_annex or under_annex is None) and not annex.file_has_content(filepath):
+        if (under_annex or under_annex is None) and not annex.file_has_content(filepath)[0]:
             lgr.info("AutomagicIO: retrieving file content of %s", filepath)
             out = annex.get(filepath)
             if not out.get('success', False):
--- a/datalad/customremotes/tests/test_archives.py
+++ b/datalad/customremotes/tests/test_archives.py
@@ -78,15 +78,15 @@ def test_basic_scenario(direct, d, d2):
     annex.add_url_to_file(fn_extracted, file_url, ['--relaxed'])
     annex.drop(fn_extracted)
 
-    list_of_remotes = annex.whereis(fn_extracted, output='descriptions')
+    list_of_remotes = annex.whereis(fn_extracted, output='descriptions')[0]
     in_('[%s]' % ARCHIVES_SPECIAL_REMOTE, list_of_remotes)
 
-    assert_false(annex.file_has_content(fn_extracted))
+    assert_false(annex.file_has_content(fn_extracted)[0])
     annex.get(fn_extracted)
-    assert_true(annex.file_has_content(fn_extracted))
+    assert_true(annex.file_has_content(fn_extracted)[0])
 
     annex.rm_url(fn_extracted, file_url)
-    assert_false(annex.drop(fn_extracted)['success'])
+    assert_false(annex.drop(fn_extracted)[0]['success'])
 
     annex.add_url_to_file(fn_extracted, file_url)
     annex.drop(fn_extracted)
@@ -100,12 +100,12 @@ def test_basic_scenario(direct, d, d2):
     # we still need to enable manually atm that special remote for archives
     # cloned_annex.enable_remote('annexed-archives')
 
-    assert_false(cloned_annex.file_has_content(fn_archive))
-    assert_false(cloned_annex.file_has_content(fn_extracted))
+    assert_false(cloned_annex.file_has_content(fn_archive)[0])
+    assert_false(cloned_annex.file_has_content(fn_extracted)[0])
     cloned_annex.get(fn_extracted)
-    assert_true(cloned_annex.file_has_content(fn_extracted))
+    assert_true(cloned_annex.file_has_content(fn_extracted)[0])
     # as a result it would also fetch tarball
-    assert_true(cloned_annex.file_has_content(fn_archive))
+    assert_true(cloned_annex.file_has_content(fn_archive)[0])
 
     # Check if protocol was collected
     if os.environ.get('DATALAD_TESTS_PROTOCOLREMOTE'):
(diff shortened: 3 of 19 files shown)


mih:
`FAILED (SKIP=52, errors=9, failures=25)`

mih:
image

mih:
`FAILED (SKIP=42, errors=7, failures=4)`

mih:
I could not finish this in time, and need to move on. It should largely be ready, but tests point to a few more easter eggs. Feel free to pick up or abandon.

yarikoptic:
> Rational:
> All look-ups and conversions should be done by top-level code, once.

`<devil-advocate>`
I would not call this a rationale, but rather an excuse for ripping that decorator out.  I wonder if we may be first should decide on "co-existence" of generators and regular methods (see https://github.com/datalad/datalad/issues/2755#issuecomment-411806205) or overview what methods should become generators? then this decorator might still then be used for "method" versions of the functions.   If no co-existence is planned, then I overall wonder if the race for making all those methods into generators would be more of a benefit or a pain?:
- **debugging** I find it rather hard to debug the code whenever things blow somewhere "upstairs" on a value derived from a value obtained from a generator down in the core -- at times you can't even know what was that value, which generator, and where down in the generator code it was produced.  Happened to me multiple times... IIRC I even asked around either there is some way to "pdb" through the "stack" of "current generators" and there is no way ATM. Some "workarounds" like 
- **coding convenience** if they become generators requiring lists, it wouldn't even be a question of using an ugly `func([x])[0]` instead of `func(x)` where x is a single item and you know it -- indexing can't apply to generators.  We would need smth like `next(func([x]))`, but that would loose checking if that value you get is the only value you expected and nothing more is coming, so we would need another helper like `theonly(func([x]))`.
- **performance** there definitely will be some impact, the question is how big (see below on deciding on what "level" -- dataset, file, file block (e.g. during download/upload progress) we would like to have generators?

For within repository operations which have long run time (e.g. `get`) we at least have progress reporting from git-annex and git (happen #2876 gets reviewed/merged).  Moreover, if we start reporting all those `(get) success` messages right after we get the file, it might interfere with progress bars reporting progress on the other files; although may be it all would just work out magically and progress bars would just keep scrolling at the bottom of that flood of `get` messages... dunno

That is why I wonder -- if that is really worth it?  For what methods, or to say "entities" returned of our *Repo in particular?  I am not yet sure if "files" is that level where we would want it... if it is a file, should we then also yield progress report may be for each update on `getting` them?  Although sounds kinda cool since then top level could decide how to report progress in a central fashion -- that would mean that every thing which might be calling `AnnexRepo.get` would need how to handle those records or skip them... bleh...  (see above "coding convenience")

May be I just do not see the actual rationale and goal here. If making core into generators is to make behavior of datalad more dynamic/fluid to the end user (using datalad via high level API) is the goal here, I would have appreciated if the issues touching upon exploration of the datasets hierarchy became more "fluid" instead, e.g. https://github.com/datalad/datalad/issues/1874 https://github.com/datalad/datalad/issues/2015.  Flooding my screen "in real time" with per-file "ok" status when `get`-ing them has little value for me... waiting for minutes for `datalad add -d . submodule` is precious.
`</devil-advocate>`

mih:
I do not think it is useful to discuss this within this PR, as it vastly exceeds its scope, but rather in #2755.

It is however important to realize that this PR neither turns any method into a generator, nor does it plan to. It merely remove a magical decision making step around low-level code that makes it a priori impossible to eventually pass a generator **in** and have any of the affected functions act on it.

Logging is completely orthogonal AFAICS to the scope of this PR. Nothing prevents logging from being optimized at the top-level. Not fundamental change should be needed, it just needs doing.

Multiple concurrent progress bars are already a reality and should work just fine via the current logging setup. The metadata aggregation code uses those.

Improvements on processing of dataset hierarchies is long planned, but waiting for direct mode to be retired. Whatever direction this will take, a path magic normalization step that gathers results into lists that then need to be taken apart again to be inspected and fused with essential property information provides no useful functionality IMHO, but make "fluid" behavior impossible (not just harder).

That being said, I am not married to this PR. The point was to make incremental changes. However, I'd personally prefer a more radical approach. I hope to have low-level `status` and `diff` done by mid next week. Based on those I plan to approach a substantial RF of top-level code that should largely avoid all existing query functions in the *Repo classes. But even in this case the presence of this decorator in essential helpers like `add`, `drop`, etc. is a major obstacles that I cannot work around without a complete base class alternative.

yarikoptic:
> I do not think it is useful to discuss this within this PR, as it vastly exceeds its scope, but rather in #2755.

My understanding of #2755 was that we have converted on ["providing a companion function that works like a generator (with `_` suffix), but keep the non-generator that we already have in general"](https://github.com/datalad/datalad/issues/2755#issuecomment-411806205) . This PR though going far beyond that by breaking API.   Given that there are already projects which depend on DataLad and use its API, breaking it would be a disservice to all of us.

> It is however important to realize that this PR neither turns any method into a generator, nor does it plan to. It merely remove a magical decision making step around low-level code that makes it a priori impossible to eventually pass a generator **in** and have any of the affected functions act on it.

_magical decision_ - was I from hogwarts, I would have said "there is nothing wrong with reliable and deterministic magic".  This decorator proven to be such overall.  Overall, because I e.g. also dislike it going from a list of a single item to just a single item (instead of keeping it as a list)...

_impossible to eventually pass a generator **in** and have any of the affected functions act on it_ - why? [Here is just a proof of concept](https://github.com/datalad/datalad/compare/master...yarikoptic:enh-normalize-generator) changing that decorator so if a generator provided for the input `files`, it would return a generator if underlying function is returning a generator.  As an example (dunno if broke anything) that prototype adjusts `AnnexRepo.get_file_key` to become a generator.  Here is what we are getting:
```shell
(git-annex)hopa:~/datalad/openfmri/ds000001/sub-01/anat[master]git-annex
$> DATALAD_LOG_LEVEL=2 python -c 'import sys; from datalad.support.annexrepo import AnnexRepo; r = AnnexRepo("../../"); func=r.get_file_key; [sys.stdout.write(k+"\n") for k in func((f for f in ["./sub-01_T1w.nii.gz", "./sub-01_inplaneT2.nii.gz"]))]'
...
[Level 5] Sending 'sub-01/anat/sub-01_T1w.nii.gz\n' to batched annex BatchedAnnex(annex_cmd=['lookupkey'], annex_options=[], git_options=<<['-c', 'receive.autogc...>>, output_proc=<function>, path=<<'/home/yoh/datalad/ope...>>) 
[Level 5] Done sending. 
[Level 5] Received output: u'MD5E-s5663237--4608ffbd6b78ce3a325eb338fa556589.nii.gz' 
MD5E-s5663237--4608ffbd6b78ce3a325eb338fa556589.nii.gz
[Level 5] Sending 'sub-01/anat/sub-01_inplaneT2.nii.gz\n' to batched annex BatchedAnnex(annex_cmd=['lookupkey'], annex_options=[], git_options=<<['-c', 'receive.autogc...>>, output_proc=<function>, path=<<'/home/yoh/datalad/ope...>>) 
[Level 5] Done sending. 
[Level 5] Received output: u'MD5E-s669578--0017a7174b9fdebeb1e57f36027bfb96.nii.gz' 
MD5E-s669578--0017a7174b9fdebeb1e57f36027bfb96.nii.gz
```

so we are getting generated values out as soon as we get them from underlying batched annex call.  And if we pass a list:
```shell
$> DATALAD_LOG_LEVEL=2 python -c 'import sys; from datalad.support.annexrepo import AnnexRepo; r = AnnexRepo("../../"); func=r.get_file_key; [sys.stdout.write(k+"\n") for k in func(["./sub-01_T1w.nii.gz", "./sub-01_inplaneT2.nii.gz"])]'  
...
[Level 5] Command: ['git', '-c', 'receive.autogc=0', '-c', 'gc.auto=0', 'annex', 'lookupkey', '--batch'] 
[Level 5] Sending 'sub-01/anat/sub-01_T1w.nii.gz\n' to batched annex BatchedAnnex(annex_cmd=['lookupkey'], annex_options=[], git_options=<<['-c', 'receive.autogc...>>, output_proc=<function>, path=<<'/home/yoh/datalad/ope...>>) 
[Level 5] Done sending. 
[Level 5] Received output: u'MD5E-s5663237--4608ffbd6b78ce3a325eb338fa556589.nii.gz' 
[Level 5] Sending 'sub-01/anat/sub-01_inplaneT2.nii.gz\n' to batched annex BatchedAnnex(annex_cmd=['lookupkey'], annex_options=[], git_options=<<['-c', 'receive.autogc...>>, output_proc=<function>, path=<<'/home/yoh/datalad/ope...>>) 
[Level 5] Done sending. 
[Level 5] Received output: u'MD5E-s669578--0017a7174b9fdebeb1e57f36027bfb96.nii.gz' 
MD5E-s5663237--4608ffbd6b78ce3a325eb338fa556589.nii.gz
MD5E-s669578--0017a7174b9fdebeb1e57f36027bfb96.nii.gz
```
they arrive together.  Obviously we could enhance that decorator with an argument to force returning a generator if underlying function returns a generator and even if input `files` argument was not a generator.   This way we could even take advantage of having this decorator for providing new desired "generator" behavior while allowing for compatibility with previous API behavior.

> Logging is completely orthogonal AFAICS to the scope of this PR. Nothing prevents logging from being optimized at the top-level. Not fundamental change should be needed, it just needs doing.
> 
> Multiple concurrent progress bars are already a reality and should work just fine via the current logging setup. The metadata aggregation code uses those.

We do not use progress bars interleaved with other output AFAIK.  If we are to start doing that, then for proper progress bar outputs [we would need to channel all other output through tqdm.write](https://github.com/tqdm/tqdm#writing-messages).  Impossible? no. More complicated? yes


> Improvements on processing of dataset hierarchies is long planned, but waiting for direct mode to be retired. Whatever direction this will take, a path magic normalization step that gathers results into lists that then need to be taken apart again to be inspected and fused with essential property information provides no useful functionality IMHO, but make "fluid" behavior impossible (not just harder).
> 
> That being said, I am not married to this PR. The point was to make incremental changes. However, I'd personally prefer a more radical approach. I hope to have low-level `status` and `diff` done by mid next week. Based on those I plan to approach a substantial RF of top-level code that should largely avoid all existing query functions in the *Repo classes. But even in this case the presence of this decorator in essential helpers like `add`, `drop`, etc. is a major obstacles that I cannot work around without a complete base class alternative.

Thank you for all this awesome work shaping up DataLad into a wonderful beast!  But once again, I think it is time now to keep in mind downstream users while doing any refactoring. Even what seemed benign change of `set_metadata` from a function to a generator, as you saw already, had notable detremental effects (e.g, https://github.com/datalad/datalad/issues/2755#issuecomment-411770040, https://github.com/nipy/heudiconv/commit/dddbd5f151a5a4e240cd766ab2465bf1d3c10f39).  Such refactorings are especially "nasty" for such `set_` methods where we typically do not inspect output and just assume that they are "setting" some values.  Whenever they become generators, they start to do nothing unless explicitly iterated.  Not quite Pythonic I would say.

mih:
OK

PR #2873, opened 2018-09-25
TST: Force direct mode tests out of travis runs

mih:
With the recent developments those are just bad for the planet. Also human costs when doing stuff like #2872 is substantial.


yarikoptic:
I would agree to disagree (imho not time yet) and do not see how direct mode testing relates to #2872 linked here

mih:
That means that I cannot invest effort in any of these activities until direct mode is officially retired (given that it is broken for any real world use case already). This includes any work on add/save.

PR #2876, opened 2018-09-26
ENH(UI): interface our progress bars into GitPython for clone/push/pull

### LINKED ISSUE #1564: some progress indication for "install" of a dataset
ATM it might be silent for minutes without giving any kind of vital signs to the user.  in the background, git reports its progress for objects fetching/unpacking... so if we could at least get some 'spinner' to show that we are still working on it -- would IMHO be useful.

yarikoptic:
This pull request fixes #1564

This pull request proposes to provide GitPython with our progress reporting for clone/push/pull.  May be there is other command where it could be provided - didn't find with `git grep`

### Changes
- [ ] This change is complete

### DIFF
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -40,6 +40,7 @@ from six import string_types
 from six import add_metaclass
 from functools import wraps
 import git as gitpy
+from git import RemoteProgress
 from gitdb.exc import BadName
 from git.exc import GitCommandError
 from git.exc import NoSuchPathError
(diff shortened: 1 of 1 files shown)


kyleam:
> I think either git or GitPython doesn't flush some buffers somewhere so typically in my cases progress bar just blinks on the screen rapidly just screwing display while leaving a blank line. But please try it out - would it show you the progress?

Hmm, I tried `datalad clone https://github.com/datalad/datalad`, and the progress bars look nice on my end -- no blinking.

yarikoptic:
Which is great! Thanks . Haven't tried GitHub yet

yarikoptic:
ok -- the reason is a good old reliance on parsing stderr from `git`.  By enabling progress reporting, we make GitPython to parse stderr output from git, so if something goes kaboom, we end up without stderr being recorded as part of the exception's .stderr.  I do feel that it is an issue within GitPython since it overall could still somehow record the entirety of stderr and assign it to the exception in such cases (yet to investigate/report) BUT in this particular case we should might better just explicitly check if that path could be created/modified even before invoking `clone` if we want to provide such a custom error message.

yarikoptic:
heh heh, with all the RFing and bugs workarounding seems have managed to kill progress bar reporting somehow in 9afc60ac2e216776cfd0cbb9b93ddde2e7ea83fb

yarikoptic:
ha ha . I bet it is this fix which is needed ;)
```
     def __enter__(self):
-        return self.close()
+        return self
```

yarikoptic:
indeed, in python3 it looks much better! Also lacks that ugly empty line after progress bar is done, which happens to me with python2

yarikoptic:
coverage is reduced due to code which hopefully would never be triggered but is there to provide robust handling of those cases without overall blow up on a progress bar error

yarikoptic:
py2 issue, as relating to GitPython, was backlinked above:  https://github.com/gitpython-developers/GitPython/issues/799 .  I do not think it is something we could/should "fix" on our end.

kyleam:
> py2 issue, as relating to GitPython, was backlinked above: gitpython-developers/GitPython#799 . I do not think it is something we could/should "fix" on our end.

The main point of my review was: "it would have been nice to see some mention/discussion of this in the commit messages or code comments so that it's clear that it is a known issue to someone investigating it later."

PR #2882, opened 2018-09-28
[WIP] New query tools in *Repo

mih:
Not worth looking at ATM. But eventually will offer a more convenient way to get info on repository content, without having to invent queries and data structures for each scenario.

### Problem being addressed

ATM we have low-level helpers that can query individual aspects of dataset content properties (`is_under_annex`, `file_has_content`, ...). Moroever, we never had a general purpose `status()` for Git repositories. Both aspects combined lead to a particular way of processing: We limits the paths that need attention in a particular case to the smallest possible subset through a series of queries, and then act upon them (path annotation approach etc...). Queries tend to be expensive (frequently lead to a dedicated git/annex call per aspect or sometimes even per path). This has the potential to be slow.

### Core idea

Implement a minimum number of helpers that can extract a maximum number of information from git/annex with the least number of git calls. The theme being to get as much in one go as possible that has no significant penalties in terms of file system access. It should become possible to get a tailored view on the state of a *single* repository without having to invent approaches and data structures for each use case. In this concept, one would err on the side of getting too much information, rather then too little. This would them allow for RFs of commands where tracking state information across various levels is crucial (i.e. `save`) and the present implementation do a good job of confusing users and authors ;-)

In terms of v6 compatibility, emphasis is put on only using Git/annex API calls that work in all scenarios and perform test of filesystem representation only for a few select aspects that generalize well.

### Major changes
- [x] `GitRepo.get_content_info()` as a frontend for `ls-files/tree` to get type, shasum and path of anything known to git, or detectable by its change-detection mechanism (incl. untracked content)
- [x] `AnnexRepo.get_content_annexinfo()` as a frontend for `find/findref` to get basic annex properties for anything known to annex
- [x] `GitRepo.status()` to report the difference between `HEAD` and the worktree. Possible states can be: 'added', 'untracked', 'clean', 'deleted', 'modified'
- [x] `AnnexRepo.annexstatus()` to amend `GitRepo.status()` with info on local content availability of annex'ed files.
- [x] RF'ed `subdatasets` to remove custom Git call and replace with `GitRepo.get_content_info()`
- [x] RF'ed `diff` to remove custom Git call to detect untracked files and replace with `GitRepo.status()`

### Open issues

- [ ] make annex content availability accessible without a call to `status()` for enhanced speed
- [ ] `ls` computes file sizes with dedicated code, but those could be taken from Repo.annexstatus()
- [ ] NFS server tests fail, because it seems files aren't dropped as fast as they should be, e.g. https://travis-ci.org/datalad/datalad/jobs/435532533
```
FAIL: datalad.support.tests.test_fileinfo.test_get_content_info
    self.test(*self.arg)
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/datalad/tests/utils.py", line 591, in newfunc
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/datalad/support/tests/test_fileinfo.py", line 226, in test_get_content_info
    assert_equal(annexstatus[p]['has_content'], 'dropped' not in s)
AssertionError: True != False

  File "/opt/python/3.5.6/lib/python3.5/multiprocessing/util.py", line 262, in _run_finalizers
    finalizer()
  File "/opt/python/3.5.6/lib/python3.5/multiprocessing/util.py", line 186, in __call__
    res = self._callback(*self._args, **self._kwargs)
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/shutil.py", line 480, in rmtree
    _rmtree_safe_fd(fd, path, onerror)
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/shutil.py", line 438, in _rmtree_safe_fd
    onerror(os.unlink, fullname, sys.exc_info())
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/shutil.py", line 436, in _rmtree_safe_fd
    os.unlink(name, dir_fd=topfd)
OSError: [Errno 16] Device or resource busy: '.nfs00000000000411f700000001'
```

### Notes

- status functions are not called `get_*status()` yet, because there is `AnnexRepo.get_status()` that does one aspect of the proposed functionality in a different way. It should likely be removed, but at present `GitRepo.status()` does not (and does not want to) detect subdataset modification, as it requires dataset traversal and is better done in top-level code IMHO.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,6 @@ env:
     - COVERAGE=coverage
     - DATALAD_DATASETS_TOPURL=http://datasets-tests.datalad.org
   matrix:
-    - DATALAD_REPO_DIRECT=yes
 ##1    - DATALAD_REPO_VERSION=6
     - DATALAD_REPO_VERSION=5
 
@@ -153,14 +152,6 @@ matrix:
 #    - DATALAD_TESTS_KNOWNFAILURES_PROBE=yes
 #    - DATALAD_TESTS_SSH=1
 #    - _DL_CRON=1
-  - python: 3.5
-    # test whether known direct mode failures still fail
-    env:
-    - DATALAD_REPO_DIRECT=yes
-    - DATALAD_TESTS_KNOWNFAILURES_SKIP=no
-    - DATALAD_TESTS_KNOWNFAILURES_PROBE=yes
-    - DATALAD_TESTS_SSH=1
-    - _DL_CRON=1
 
   - python: 3.5
     # test with extension datalad-crawler
@@ -196,13 +187,6 @@ matrix:
 #    - DATALAD_TESTS_KNOWNFAILURES_PROBE=yes
 #    - DATALAD_TESTS_SSH=1
 #    - _DL_CRON=1
-  # test whether known direct mode failures still fail
-  - env:
-    - DATALAD_REPO_DIRECT=yes
-    - DATALAD_TESTS_KNOWNFAILURES_SKIP=no
-    - DATALAD_TESTS_KNOWNFAILURES_PROBE=yes
-    - DATALAD_TESTS_SSH=1
-    - _DL_CRON=1
 
 # Causes complete laptop or travis instance crash atm, but survives in a docker
 # need to figure it out (looks like some PID explosion)
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,6 +7,7 @@ environment:
       PYTHON_ARCH: "32"
       MINICONDA: C:\Miniconda35
       DATALAD_TESTS_SSH: 1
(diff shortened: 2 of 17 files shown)


mih:
Will come back later. This hit the wall with windows path handling in datalad. But as the whole point of this is having an query tools that work regardless of platform-specific, there is no point in pursuing this at this point.

FTR: the main obstacles is that path snippets are fused to absolute paths in many places. The data paths tend to come from the input arguments, usually in platform dependent form, but pretty much all of git's reporting is in POSIX conventions. Any `op.join()` of those is invalid. Any universal platform dependent handling is incompatible with stuff that needs to go into Git's helper files (config, .gitmodules), and our own data structures.

Ping #2546

PR #2886, opened 2018-10-01
RF: Move get_git_dir() into GitRepo

mih:
Previously in datalad.distribution.utils, but such base functionality
should no be scattered around to prevent needless reinvention.

Kept the function static to avoid any cost of Repo instantiation,
given that this is the main usage pattern ATM.

### DIFF
--- a/datalad/distribution/dataset.py
+++ b/datalad/distribution/dataset.py
@@ -10,7 +10,6 @@
 """
 
 import logging
-from os.path import abspath
 from os.path import curdir
 from os.path import exists
 from os.path import join as opj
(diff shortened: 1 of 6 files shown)


mih:
@kyleam Made it static to match `is_valid_repo()`

I guess I will change the content of this helper to make the call you suggested. No memory of the origin of the current implementation. I guess we were not aware at that point.

thx much!

mih:
Changed my mind. This would require a dedicated call to Git, and I am not confident that it is worth it, given we have this implementation already. I opened an issue about it regardless of this concern: #2888

kyleam:
> Made it static to match is_valid_repo()

I see.  Not surprisingly, I also would prefer `is_valid_repo` to be a top-level function in `gitrepo`.  And even if it's contained within the class, I don't understand why it's a class method rather than a static method.  But at any rate, that's both minor and outside the scope of this PR.

> Changed my mind. This would require a dedicated call to Git, and I am not confident that it is worth it, given we have this implementation already.

Sure.  My comment was a genuine sidenote, not something I was trying to get tacked on to this PR.

PR #2891, opened 2018-10-03
BF: Don't crash on no reported percentage

bpoldrack:
This pull request fixes a failure discovered in https://github.com/datalad/datalad-container. Test was failing in https://travis-ci.org/datalad/datalad-container/jobs/436153807, because `annex-addurl` didn't report progress in the JSON report. Therefore `ProcessAnnexProgressIndicators` set the reported percentage to default to an empty string, which was then casted to a `float`, yielding a `ValueError`.

Independently on why exactly `addurl` doesn't report it (seems to just take a while for the first report to contain a field "percent-progress"), WE are setting the default to an empty string and therefore have to be able to deal with that value in the function we pass it to.

This pull request proposes to just catch the `ValueError` and return `None` in `AnnexRepo._get_size_from_perc_complete()`. 

### Changes
- [x] This change is complete

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -3665,7 +3665,7 @@ def _get_size_from_perc_complete(count, perc):
     try:
         portion = (float(perc) / 100.)
     except ValueError:
-        return None
+        portion = None
     return int(math.ceil(int(count) / portion)) \
         if portion else 0
 
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -2112,8 +2112,8 @@ class AnnexRepo(GitRepo, RepoInterface):
                     % str(out_json))
             # Make the output's structure match bcmd's.
             out_json = out_json[0]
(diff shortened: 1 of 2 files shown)


datalad/support/annexrepo.py
```diff
@@ -3661,7 +3661,11 @@ def close(self, return_stderr=False):
 
 def _get_size_from_perc_complete(count, perc):
     """A helper to get full size if know % and corresponding size"""
-    portion = (float(perc) / 100.)
+
+    try:
+        portion = (float(perc) / 100.)
+    except ValueError:
+        return None
```
yarikoptic: ATM it should be ok to introduce `None` as possible output here (at the use point it is `or`-ed with 0 after all), but I guess it would have been safer to just return 0 (although `None` is more semantically correct).
I am just afraid that if above code is changed under assumption that this one returns an `int`, we could be in trouble without realizing it (since there is no dedicated "integration" test for this use case). 
So - I am ok with it as is, but ideally - should we just return 0 here (or just `portion = None` and leave return to logic below)

Issue #1651, opened 2017-07-21
add fails to add unlocked file which now would need to go to git if was added anew


yarikoptic:
so there is a change of .gitattributes interim:
```
yoh@hopa:~> datalad create /tmp/testunlock
[INFO   ] Creating a new annex repo at /tmp/testunlock 
create(ok): /tmp/testunlock (dataset)                                                                           

yoh@hopa:~> cd /tmp/testunlock

yoh@hopa:/tmp/testunlock> echo 123 > 123

yoh@hopa:/tmp/testunlock> datalad add 123
add(ok): /tmp/testunlock/123 (file)                                                                             
save(ok): /tmp/testunlock (dataset)
action summary:
  add (ok: 1)
  save (ok: 1)

yoh@hopa:/tmp/testunlock> datalad unlock 123
unlock(ok): 123 (file)

yoh@hopa:/tmp/testunlock> datalad add 123
add(ok): /tmp/testunlock/123 (file)                                                                             
save(notneeded): /tmp/testunlock (dataset)
action summary:
  add (ok: 1)
  save (notneeded: 1)

yoh@hopa:/tmp/testunlock> datalad unlock 123
unlock(ok): 123 (file)

yoh@hopa:/tmp/testunlock> cat .gitattributes
* annex.backend=MD5E

*yoh@hopa:/tmp/testunlock> echo '* annex.largefiles=(not(mimetype=text/*' >> .gitattributes

yoh@hopa:/tmp/testunlock> datalad save -m 'all text files go to git' .gitattributes
save(ok): /tmp/testunlock (dataset)                                                                             

yoh@hopa:/tmp/testunlock> datalad unlock 123

yoh@hopa:/tmp/testunlock> datalad add 123
add(ok): /tmp/testunlock/123 (file) [non-large file; adding content to git repository]                          
add(ok): /tmp/testunlock/123 (file) [non-large file; adding content to git repository]
Failed to run ['git', '-c', 'receive.autogc=0', '-c', 'gc.auto=0', 'commit', '-m', '[DATALAD] added content', u'123'] under '/tmp/testunlock'. Exit code=1. out= err=git-annex: Cannot make a partial commit with unlocked annexed files. You should `git annex add` the files you want to commit, and then run git commit.
git-annex: Cannot make a partial commit with unlocked annexed files. You should `git annex add` the files you want to commit, and then run git commit.

yoh@hopa:/tmp/testunlock> datalad add 123
[...]

PR #2898, opened 2018-10-05
WiP: `datalad.coreapi.open`

### LINKED ISSUE #2890: "edit" or even "open" construct
This discussion would not be new, previous ones:
- #378 (Brainstorming for "git annex unlock" use) - boiled down to use unlock/lock pair
- #453 (initial title is irrelevant) - has [joey's comment](https://github.com/datalad/datalad/issues/453#issuecomment-220850060) on `git annex adjust --unlock`

I still think that in particular with Python API it would be a big convenience to the users to be able to manipulate file content in basic operations without doing all the manual get/unlock/lock/... combination.  E.g.  something like

## 'w' mode:

```python
with ds.open("filename", "w") as f:
   f.write("new content replacing the old one even if it was not present")
```
which would do 
- `rm filename`
- `open("filename", "w")`
- ... (context) 
- `git annex add "filename"`

## "a" (`b` or not)

it would 
- `get`
- `unlock`
- `open("filename", "a")`;
- ...
- `git annex add ...`

## 'r'

- `get`
- `open`

in `w` and `a` mode if possible/needing hiding any v5 vs v6 peculiarities (i.e. locking it back so real symlink is committed and not git reference file).

What do you think?

yarikoptic:
This pull request fixes #2890

In current state it is more of a tech preview (although not that "revolutionary")

### Changes
- [x] Decide if `open` is a good name (overloads built-in)
- [ ] Address numerous TODOs left around in the code and tests
    - [x] Docstrings
    - [ ] Decide if any custom handling of exception is needed, or actually we should just avoid doing any post_close altogether if exception happened
    - [ ] Upon `w`, transfer file permissions (executability, group writeability) and ownership
    - [ ] Extend testing
    - [ ] Custom handling of "unlocked" files in V6
- [ ] Decide on the destiny of https://github.com/datalad/datalad/issues/1651 although largely independent of this issue which shouldn't care (but initial test was failing due to #1651)

### DIFF
--- a/datalad/interface/open.py
+++ b/datalad/interface/open.py
@@ -10,7 +10,7 @@
 
 __docformat__ = 'restructuredtext'
 
-
+import os
 import logging
 from contextlib import contextmanager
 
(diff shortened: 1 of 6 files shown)


datalad/interface/open.py
```diff
@@ -0,0 +1,177 @@
...
+import logging
+
+from datalad.distribution.dataset import resolve_path
+from datalad.distribution.dataset import datasetmethod
+from datalad.utils import assure_list
+
+
+lgr = logging.getLogger('datalad.interface.open')
+
+_builtin_open = open
```
kyleam: I wonder if we should use `io.open` and make this open support the full set of parameters.  One downside of that is that it'd be slower for Python 2, which I would obviously weigh different than you since you still primarily use py2.

yarikoptic: I thought that `io.open` is considered an "internal" implementation and `open` generally recommended instead... I could be wrong - will check later

kyleam: > I thought that io.open is considered an "internal" implementation and open generally recommended instead...

I'm not aware of that and didn't find anything that reflects the in the Python docs.  My understanding is that `io.open` is useful on py2 for py3-compatibility.  On py3, `io.open` is simply an alias for `open`.

mih: That is my understanding too.

yarikoptic: ok, let's proxy `io.open` then

mih:
Closing due to inactivity.

PR #2904, opened 2018-10-08
RF: Centralize submodule addition logic in GitRepo

### LINKED ISSUE #2909: GitRepo.add_submodule() produced unsupported/unexpected .git setup
Changes in #2904 exposed the fact that from `GitRepo.add_submodule()` was used outside the context
of top-level `install`, it yields a `.git` setup located in `../.git/modules/sub` (i.e. in the superdataset).

We have `_fixup_submodule_dotgit_setup()` which is used to assert `install` results, but does not actually fixup anything anymore, and was not used in `add_submodule()` until #2904.

It should probably be RF to actually fix things up again, in `add_submodule()`.

mih:
Essential verification check was only available in `add`

- [x] fix #2909
- [x] adjust Repo tests to handle the newly reported issues during submodule addition (prev. silent or different)

### DIFF
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -58,6 +58,7 @@ from datalad.utils import on_windows
 from datalad.utils import getpwd
 from datalad.utils import updated
 from datalad.utils import posix_relpath
+from datalad.utils import assure_dir
 from ..utils import assure_unicode
 
 # imports from same module:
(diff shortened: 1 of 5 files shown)


datalad/support/gitrepo.py
```diff
@@ -2208,6 +2208,21 @@ def add_submodule(self, path, name=None, url=None, branch=None):
...
             cmd += ['-b', branch]
         if url is None:
+            # repo must already exist locally
+            subm = GitRepo(op.join(self.path, path), create=False, init=False)
+            # check that it has a commit, and refuse
+            # to operate on it otherwise, or we would get a bastard
+            # submodule that cripples git operations
+            if not subm.get_hexsha():
+                raise InvalidGitRepositoryError(
+                    'cannot add subdataset %s with no commits', subm)
```
kyleam: This message needs to be formatted.

mih: Thx.

PR #2905, opened 2018-10-08
ENH+BF: run-procedure

bpoldrack:
Fixes and additions to `run-procedure`:

- `run-procedure` required the `-d` parameter to discover procedures within a dataset even if called from within the dataset. That's not intuitive and I changed that. When called from within a dataset it will use it (except when passing in sth else via `-d` of course).
- BF: files that are neither executable nor end with `.py` or `.sh` but are in a configured look up location for procedures caused the command to crash. Fixed that.
- procedures can now recursively be discovered in subdatasets as well. The uppermost has highest priority
- added two configs for procedures: `datalad.procedures.NAME.call-format` and `datalad.procedures.NAME.help`. The former replaces (if given) the default, which is based on discovered file type. This allows not only for changing order or even execute different scripts than just python and bash, but also for usage of `run` substitutions.
The latter is queried by `datalad run-procedure --help-proc NAME`, which yields the value of this config as `message`. So we can provide description, usage info etc. on a procedure.
- proper 'impossible' result when procedure not found (instead of `ValueError`)

### DIFF
--- a/datalad/interface/run_procedure.py
+++ b/datalad/interface/run_procedure.py
@@ -264,8 +264,9 @@ class RunProcedure(Interface):
             for m, cmd_tmpl, cmd_help in _get_procedure_implementation('*', ds=ds):
                 if m in reported:
                     continue
+                cmd_type, cmd_tmpl_g = _guess_exec(m)
                 if not cmd_tmpl:
-                    cmd_type, cmd_tmpl = _guess_exec(m)
+                    cmd_tmpl = cmd_tmpl_g
                 if cmd_type is None and cmd_tmpl is None:
                     # doesn't seem like a match
                     continue
@@ -295,15 +296,25 @@ class RunProcedure(Interface):
             procedure_file, cmd_tmpl, cmd_help = \
                 next(_get_procedure_implementation(name, ds=ds))
         except StopIteration:
-            # TODO error result
-            raise ValueError("Cannot find procedure with name '%s'", name)
+            res = get_status_dict(
+                    action='run_procedure',
+                    # TODO: Default renderer requires a key "path" to exist.
+                    # Doesn't make a lot of sense in this case
+                    path=name,
+                    logger=lgr,
+                    refds=ds.path if ds else None,
+                    status='impossible',
+                    message="Cannot find procedure with name '%s'" % name)
+            yield res
+            return
 
+        cmd_type, cmd_tmpl_g = _guess_exec(procedure_file)
         if not cmd_tmpl:
-            cmd_type, cmd_tmpl = _guess_exec(procedure_file)
-            if cmd_tmpl is None:
-                raise ValueError("No idea how to execute procedure %s. "
-                                 "Missing 'execute' permissions?",
-                                 procedure_file)
+            cmd_tmpl = cmd_tmpl_g
+        if not cmd_tmpl:
+            raise ValueError("No idea how to execute procedure %s. "
+                             "Missing 'execute' permissions?",
+                             procedure_file)
 
         if help_proc:
             res = get_status_dict(
@@ -313,6 +324,7 @@ class RunProcedure(Interface):
                     logger=lgr,
                     refds=ds.path if ds else None,
                     status='ok',
+                    procedure_type=cmd_type,
                     procedure_callfmt=cmd_tmpl,
                     message=cmd_help)
             yield res
--- a/datalad/interface/run_procedure.py
+++ b/datalad/interface/run_procedure.py
@@ -32,11 +32,13 @@ from datalad.support.constraints import EnsureNone
 from datalad.support.param import Parameter
 from datalad.distribution.dataset import datasetmethod
 from datalad.support.exceptions import InsufficientArgumentsError
+from datalad.support.exceptions import NoDatasetArgumentFound
 
 from datalad.utils import assure_list
 
 # bound dataset methods
 from datalad.interface.run import Run
+from datalad.distribution.subdatasets import Subdatasets
 
 lgr = logging.getLogger('datalad.interface.run_procedures')
 
@@ -61,6 +63,11 @@ def _get_procedure_implementation(name='*', ds=None):
             # TODO `get` dirs if necessary
             for m in _get_file_match(op.join(ds.path, dir), name):
                 yield m
+        # 1.1. check subdatasets recursively
+        # for subds in ds.subdatasets(return_type='generator',
+        #                             result_xfm='datasets'):
+        #     yield _get_procedure_implementation(name=name, ds=subds)
+
     # 2. check system and user account for procedure
     for loc in (cfg.obtain('datalad.locations.user-procedures'),
                 cfg.obtain('datalad.locations.system-procedures')):
@@ -97,7 +104,7 @@ def _guess_exec(script_file):
     elif script_file.endswith('.py'):
         return (u'python_script', u'python "{script}" "{ds}" {args}')
     else:
-        return None
+        return None, None
 
 
 @build_doc
@@ -204,9 +211,12 @@ class RunProcedure(Interface):
         if not spec and not discover:
             raise InsufficientArgumentsError('requires at least a procedure name')
 
-        ds = require_dataset(
-            dataset, check_installed=False,
-            purpose='run a procedure') if dataset else None
+        try:
+            ds = require_dataset(
+                dataset, check_installed=False,
+                purpose='run a procedure')
+        except NoDatasetArgumentFound:
+            ds = None
 
         if discover:
             reported = set()
@@ -214,6 +224,9 @@ class RunProcedure(Interface):
                 if m in reported:
                     continue
                 cmd_type, cmd_tmpl = _guess_exec(m)
+                if cmd_type is None and cmd_tmpl is None:
+                    # doesn't seem like a match
+                    continue
                 res = get_status_dict(
                     action='run_procedure',
                     path=m,
--- a/datalad/interface/run_procedure.py
+++ b/datalad/interface/run_procedure.py
@@ -129,15 +129,43 @@ def _get_procedure_implementation(name='*', ds=None):
 
 
 def _guess_exec(script_file):
+
+    state = None
+    try:
+        is_exec = os.stat(script_file).st_mode & stat.S_IEXEC
+    except OSError as e:
+        from errno import ENOENT
+        if e.errno == ENOENT:
+            # path does not exist or is a broken symlink
+            state = 'absent'
+            if op.islink(script_file):
+                # broken symlink;
+                # we can't figure whether it's executable:
+                is_exec = False
+                # apart from that proceed, since we can still tell in case of
+                # .py or .sh
+            else:
+                # does not exist; there's nothing to detect at all
+                return {'type': None, 'template': None, 'state': None}
+        else:
+            from six import reraise
+            reraise(e)
+
     # TODO check for exec permission and rely on interpreter
-    if os.stat(script_file).st_mode & stat.S_IEXEC:
-        return ('executable', u'"{script}" "{ds}" {args}')
+    if is_exec:
+        return {'type': u'executable',
+                'template': u'"{script}" "{ds}" {args}',
+                'state': state}
     elif script_file.endswith('.sh'):
-        return (u'bash_script', u'bash "{script}" "{ds}" {args}')
+        return {'type': u'bash_script',
+                'template': u'bash "{script}" "{ds}" {args}',
+                'state': state}
     elif script_file.endswith('.py'):
-        return (u'python_script', u'python "{script}" "{ds}" {args}')
+        return {'type': u'python_script',
+                'template': u'python "{script}" "{ds}" {args}',
+                'state': state}
     else:
-        return None, None
+        return {'type': None, 'template': None, 'state': None}
 
 
 @build_doc
@@ -294,12 +322,15 @@ class RunProcedure(Interface):
             for m, cmd_tmpl, cmd_help in _get_procedure_implementation('*', ds=ds):
                 if m in reported:
                     continue
-                cmd_type, cmd_tmpl_g = _guess_exec(m)
-                if not cmd_tmpl:
-                    cmd_tmpl = cmd_tmpl_g
-                if cmd_type is None and cmd_tmpl is None:
+                ex = _guess_exec(m)
+                # configured template (call-format string) takes precedence:
+                if cmd_tmpl:
+                    ex['template'] = cmd_tmpl
+                if ex['type'] is None and ex['template'] is None:
                     # doesn't seem like a match
                     continue
+                message = ex['type'] if ex['type'] else 'unknown type'
+                message += ' (missing)' if ex['state'] == 'absent' else ''
                 res = get_status_dict(
                     action='run_procedure',
                     path=m,
@@ -307,9 +338,10 @@ class RunProcedure(Interface):
                     logger=lgr,
                     refds=ds.path if ds else None,
                     status='ok',
-                    procedure_type=cmd_type,
-                    procedure_callfmt=cmd_tmpl,
-                    message=cmd_type)
+                    state=ex['state'],
+                    procedure_type=ex['type'],
+                    procedure_callfmt=ex['template'],
+                    message=message)
                 reported.add(m)
                 yield res
             return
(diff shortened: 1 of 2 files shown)


yarikoptic:
Description is quite non descriptive...

bpoldrack:
There's more to come. I'll add a description.

bpoldrack:
FTR: For some reason the creation of the test dataset (all three tests) fails to correctly commit if on windows *and* in direct mode. Works in direct mode on linux and in v6 on windows though. Changed the skipping accordingly.

@kyleam : There's one thing I'm currently not sure about. Should we expose `inputs` and `outputs` for `run-procedure` and pass it on to `run` or do you see any reason to not do that? While a procedure can do some (un-)locking and getting from within, I wonder whether we should allow to use the mechanism that is already there.

@mih: Anything in mind we need, that I may have missed?

bpoldrack:
@yarikoptic :
Any idea what this is about? https://travis-ci.org/datalad/datalad/jobs/440130217#L3025

kyleam:
> @kyleam : There's one thing I'm currently not sure about. Should we expose inputs and outputs for run-procedure and pass it on to run or do you see any reason to not do that? While a procedure can do some (un-)locking and getting from within, I wonder whether we should allow to use the mechanism that is already there.

How would the inputs and outputs of a procedure be specified?

bpoldrack:
> How would the inputs and outputs of a procedure be specified?

Just the same way. I thought about providing those parameters for `run-procedure` as well and just pass them on to `run`. This would lack a possibility to define them via that call format string, but at least allow to use them with the actual call to `run-procedure` and possibly use `{{inputs}}` and `{{outputs}}` in the format string. But I may miss something here.

kyleam:
> Just the same way. I thought about providing those parameters for `run-procedure` as well and just pass them on to `run`.

OK.

> This would lack a possibility to define them via that call format string,

Yeah, that's one reason why I'm wondering what specific mechanism you have in mind.  Also, there's the interaction with `proc_{pre,post}`.

> but at least allow to use them with the actual call to `run-procedure` and possibly use `{{inputs}}` and `{{outputs}}` in the format string. But I may miss something here.

It's quite possible I'm the one missing something :]  Here's the main thing I'm confused about: I think of procedures as these sort of canned routines that the caller doesn't need to know much about aside from their name and general purpose.  Ignoring `proc_{pre,post}`, it seems like `run_procedure` callers would need to have an understanding of a procedure's innards in order to specify the inputs and outputs.

bpoldrack:
> Also, there's the interaction with proc_{pre,post}. 

> it seems like run_procedure callers would need to have an understanding of a procedure's innards in order to specify the inputs and outputs.

Actually, those are two good arguments to not make things more confusing without an actual necessity. Initially I thought: Well, if it doesn't really cost something, why hiding it away? But then this might be confusing and not perceived as an additional option but the standard way to go (which it shouldn't be). And it might indeed lead to a mess with `proc_{pre, post}`, where we can't easily decide about a paradigm on what to record in what commit under what circumstances.
Long story short: Thinking about it again, I guess we should not do that as long as we don't have a case where it's actually needed.

yarikoptic:
re travis failure
```
reading sources... [100%] usecases/simple_provenance_tracking                   
Warning, treated as error:
/home/travis/build/datalad/datalad/datalad/interface/run_procedure.py:docstring of datalad.api.run_procedure:66:Unexpected indentation.
make: *** [html] Error 2
make: Leaving directory `/home/travis/build/datalad/datalad/docs'
The command "if [ "${DATALAD_LOG_LEVEL:-}" != 2 -a -z "${BUILD_DATALAD_EXTENSION:-}" ]; then PYTHONPATH=$PWD $NOSE_WRAPPER make -C docs html doctest; fi" exited with 2.
```

bpoldrack:
Yes, I noticed. I'm amending like hell to figure what fx$!% whitespace crashes the beast. ;-)

datalad/interface/tests/test_run_procedure.py
```diff
@@ -86,3 +126,144 @@ def test_procedure_discovery():
...
+    ds = Dataset(path).create(force=True)
+    # TODO: this procedure would leave a clean dataset, but `run` cannot handle dirty
+    # input yet, so manual for now
+    # V6FACT: this leaves the file staged, but not committed
+    ds.add('code', to_git=True)
+    # V6FACT: even this leaves it staged
+    ds.add('.')
+    # V6FACT: but this finally commits it
+    ds.save()
+    # TODO remove above two lines
```
kyleam: Again, I know this is just copied, and I don't think it needs to be addressed in this PR, but based on quick testing, v6 mode now appears to behave as expected without these lines (presumably either due to an update in annex or a fix on our end).

datalad/interface/run_procedure.py
```diff
@@ -130,6 +170,14 @@ class RunProcedure(Interface):
     and subsequently in the 'resources/procedures/' directories of any
     installed extension, and, lastly, of the DataLad installation itself.
 
+    Please note, that a dataset that defines
```
kyleam: s/,//

datalad/interface/run_procedure.py
```diff
@@ -47,26 +49,64 @@ def _get_file_match(dir, name='*'):
...
+        return v[-1], h
+    else:
+        return v, h
 
 
 def _get_procedure_implementation(name='*', ds=None):
     ds = ds if isinstance(ds, Dataset) else Dataset(ds) if ds else None
-    # 1. check dataset for procedure
+
+    # 1. check system and user account for procedure
```
kyleam: I don't have a strong preference on what the precedence should be, but since you're changing the original order, it would have been nice if the commit message gave some rational for this.

And I think the order in "Directories identified by the configuration settings" of the help is supposed to follow the order, so that should be updated.

datalad/interface/run_procedure.py
```diff
@@ -47,26 +49,64 @@ def _get_file_match(dir, name='*'):
...
-                yield m
-    # 2. check system and user account for procedure
-    for loc in (cfg.obtain('datalad.locations.user-procedures'),
-                cfg.obtain('datalad.locations.system-procedures')):
-        for dir in assure_list(loc):
-            for m in _get_file_match(dir, name):
-                yield m
+            for m, n in _get_file_match(op.join(ds.path, dir), name):
+                yield (m,) + _get_proc_config(n, ds=ds)
+        # 1.1. check subdatasets recursively
```
kyleam: s/1.1/2.1/ ?

datalad/interface/run_procedure.py
```diff
@@ -151,6 +199,25 @@ class RunProcedure(Interface):
     of taking at least one positional argument (the absolute path to the
     dataset they shall operate on).
 
+    For further customization there are two configuration settings per procedure
+    available:
+
+    - 'datalad.procedures.<NAME>.call-format'
+      fully customizable format string to determine how to execute procedure
+      NAME (see also datalad-run).
+      It currently requires to include the following placeholders:
```
kyleam: s/currently requires to/must/

Though I suppose only "{script}" is strictly necessary.  I imagine all scripts would want the dataset, but `args` may often be empty and a callfmt without `{args}` would work fine.

datalad/interface/run_procedure.py
```diff
@@ -151,6 +199,25 @@ class RunProcedure(Interface):
     of taking at least one positional argument (the absolute path to the
     dataset they shall operate on).
 
+    For further customization there are two configuration settings per procedure
+    available:
+
+    - 'datalad.procedures.<NAME>.call-format'
+      fully customizable format string to determine how to execute procedure
+      NAME (see also datalad-run).
```
kyleam: I'm not sure we should refer to `run` here because I don't imagine we want to encourage passing through placeholders like "{{pwd}}" in the call format.  The view the fact that `run-procedure` uses `run` underneath as an implementation detail.

datalad/interface/run_procedure.py
```diff
@@ -192,6 +259,14 @@ class RunProcedure(Interface):
             doc="""if given, all configured paths are searched for procedures
             and one result record per discovered procedure is yielded, but
             no procedure is executed"""),
+        help_proc=Parameter(
+            args=('--help-proc',),
+            action='store_true',
+            doc="""if given, get a help message for procedure NAME from config
+            setting datalad.procedures.NAME.help. [CMD: Note, that this
```
kyleam: The CMD note is true for other options too and isn't worth mentioning IMO.

datalad/interface/run_procedure.py
kyleam: Never mind.  I see that that one of the tests explicitly deals with passing through substitution placeholders, so I will update my view that using `run` is an implementation detail :]

datalad/interface/run_procedure.py
```diff
@@ -47,26 +49,64 @@ def _get_file_match(dir, name='*'):
...
+      if there's nothing configured
+    """
+    # figure what ConfigManager to ask
+    cm = cfg if ds is None else ds.config
+    v = cm.get('datalad.procedures.{}.call-format'.format(name), None)
+    h = cm.get('datalad.procedures.{}.help'.format(name), None)
+    if isinstance(v, tuple):
+        # If multiple values are defined (at the same level), ConfigManager
+        # returns a tuple. Go with the last one.
+        # TODO: Should we issue a warning or fail entirely?
```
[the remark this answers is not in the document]
bpoldrack: I don't think always returning a single value will be the solution. `ConfigManager` is explicitly made to be able to deal with multiple definitions of the same config item at the same level. But we probably should somehow be able to tell `ConfigManager` that for a particular item we only accept one value. However, I agree: This exceeds this PR.

datalad/interface/run_procedure.py
```diff
@@ -200,20 +275,31 @@ class RunProcedure(Interface):
...
-            for m in _get_procedure_implementation('*', ds=ds):
+            for m, cmd_tmpl, cmd_help in _get_procedure_implementation('*', ds=ds):
                 if m in reported:
                     continue
-                cmd_type, cmd_tmpl = _guess_exec(m)
+                cmd_type, cmd_tmpl_g = _guess_exec(m)
+                if not cmd_tmpl:
+                    cmd_tmpl = cmd_tmpl_g
+                if cmd_type is None and cmd_tmpl is None:
+                    # doesn't seem like a match
```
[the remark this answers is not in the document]
bpoldrack: Agree.

datalad/interface/tests/test_run_procedure.py
bpoldrack: Well, I should probably try to remove it within this PR. It will hang around there for quite some while otherwise, I guess.

datalad/interface/run_procedure.py
bpoldrack: You're right, of course.

datalad/interface/run_procedure.py
bpoldrack: Right. Thanks for spotting.

datalad/interface/run_procedure.py
bpoldrack: Actually this requires an addition. While `{args}` isn't actually necessary, I think it would currently crash if it isn't in the format string, but one gives additional args when calling `run-procedure`. This should yield a proper error instead.

datalad/interface/run_procedure.py
bpoldrack: :)
Yes. And this passing through of substitutions is a main point of the toolbox-usecase that led to coming up with this PR.

datalad/interface/run_procedure.py
bpoldrack: Yes and no. While it's true for `-d` and `--discover` as well, `-d` probably is hardly ever used and it's kind of intuitive to use `--discover` only without a `spec` argument (and therefore not run into the issue). `run-procedure XXX --help-proc` however looks perfectly reasonable but might lead to a completely unexpected and non-telling failure (or even not a failure at all depending on what `XXX` does with that argument). So, if anything we should mention it for the options, too. It might not be that obvious at a first glance that the procedures name will eat up everything following, although it should be clear once you actually think about it.

datalad/interface/tests/test_run_procedure.py
kyleam: > It will hang around there for quite some while otherwise, I guess.

I've noted it, so I would look into it once this PR lands, but of course I don't mind you looking into it now.

datalad/interface/run_procedure.py
kyleam: `.format` works fine if a placeholder isn't present, so I suspected it would be OK, and I tested it locally and it worked fine, but perhaps something was wonky with my test.

datalad/interface/run_procedure.py
kyleam: I don't understand that, but I take your word for it.

datalad/interface/run_procedure.py
kyleam: None of that convinces me it is worth mentioning, but of course not thinking it's worth mentioning also means I don't particularly care if it *is* mentioned :]

datalad/interface/run_procedure.py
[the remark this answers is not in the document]
kyleam: > I don't think always returning a single value will be the solution

Yes, you're right, things aren't so simple and some callers do expect multiple values.

> But we probably should somehow be able to tell ConfigManager that for a particular item we only accept one value.

Sounds good.  Based on frequency of use by our current callers, I wonder if "single value" should be the default and we should have a separate method for multiple items.

Anyway, I do have a question related to this PR:  Your comment says "If multiple values are defined (at the same level), ConfigManager returns a tuple".  I'm confused about what the "level" is here.   I don't think it is related to config location (global, .git/config, .datalad/config,...) because settings don't need to be at the same level to be represented as a tuple:

```
$ git config --global datalad.testconfig false                                                  
$ git config datalad.testconfig false                                                          
$ python -c "import datalad.api as dl; print(dl.Dataset('.').config.get('datalad.testconfig'))"
('false', 'false')                 
```

So the only way I can think of to interpret "at the same level" as meaning "this level in the dataset hierarchy".

P.S. I like this example because a naive caller would get say `if config.get(...)` and unexpectedly get something that evaluates to true even though both the global and local setting is false.  `getbool` also fails to handle it:

```
$ python -c "import datalad.api as dl; print(dl.Dataset('.').config.getbool('datalad', 'testconfig'))"
  File "/home/kyle/src/python/datalad/datalad/config.py", line 473, in getbool
    return anything2bool(val)
  File "/home/kyle/src/python/datalad/datalad/config.py", line 113, in anything2bool
    % repr(val))
TypeError: Got value ('false', 'false') which could not be interpreted as a boolean
```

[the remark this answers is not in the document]
bpoldrack: > I don't think it is related to config location (global, .git/config, .datalad/config,...)

It is related. What I mean by "level" is the hierarchy of things we look at to find the (or a) procedure. In your case it's both "user level": First goes into `~/.gitconfig`, second into `.git/config`. Other levels would be "system", "dataset" (`.datalad/config`, which comes with the dataset in opposition to `.git/config`), datalad itself and last any datalad extension.

Re your PS:
That's a good catch and definitely sth to fix!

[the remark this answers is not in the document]
bpoldrack: Just for clarification: Those are the levels to be searched for procedures by `run-procedure` and also the levels of consideration by `ConfigManager` when figuring what to return. And this again is related to the change of order you noticed. They are consistent now. (But I still should write down the rational for that particular order)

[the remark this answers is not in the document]
kyleam: Thanks for clarifying. I think your mapping between "procedure levels" and "config levels" breaks down because AFAICS `ConfigManager` (as a getter) only distinguishes two levels: dataset (".datalad/config") versus "everything else git config will report".

If I have a setting in `/etc/gitconfig`, `~/.gitconfig`, and `$REPO/.git/config`, `ConfigManager` reports

```shell
$ python -c "import datalad.api as dl; print(dl.Dataset('.').config.get('datalad.test'))"
('etc-config', 'home-config', 'repo-config')
```

By procedure level standards, that includes both "system" and "user" levels.

bpoldrack:
Re broken symlinks:
@mih: Please have a look at https://github.com/datalad/datalad/pull/2905/commits/1ff10fefa1f30a634a351044d7814b03ce4c3c47

datalad/interface/run_procedure.py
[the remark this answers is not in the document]
bpoldrack: True, but this reporting by `git-config` still happens in a particular order (more general to more specific). So, this is the reason I went with the last entry in case of multiple values returned.

I'll try to articulate this idea more clearly in comment/docstring

bpoldrack:
@kyleam : I hope changes of today clarify things sufficiently. Otherwise don't hesitate to further complain. ;-)

bpoldrack:
Thanks for having a closer look, @kyleam!

PR #2907, opened 2018-10-08
Move source package lintian overrides to debian/source.

jelmer:
Move source package lintian overrides to debian/source.


yarikoptic:
according to the `man lintian` seems to be the new (?) preferred location, thus merging, thanks @jelmer

Issue #2910, opened 2018-10-10
RFC: our reliance on PWD env var backfires...


yarikoptic:
#### What is the problem?
If users use datalad as a python module, and they, or some other process (jupyter?), switches to another workdir (obviously using `os.chdir`, not our `chpwd`) then any datalad command operating on relative filenames would fail with confusing errors or just would do the wrong thing... causes confusion, frustration, etc. Definitely we should mitigate it somehow.

Background: Use of env var PWD was introduced with the sole purpose of being able to operate on (compare etc) full paths while some directory in the path is a symlink and then `os.getcwd()` would return dereferenced path. It might be that one of the initial use-cases was a "subdataset" which is just symlinked from another location? because otherwise I think we could have just used/report full dereferenced path.  Overall - not sure now what problem we were/are solving besides reporting the full path user expects to see whenever in his shell he is within such a symlinked directory.

And now I wonder what should we do in respect to the original problem that we are not really in full control of current directory.  I see possibilities

- in `getpwd` verify that envvar PWD and `os.getpwd()` point to the same directory (after realpath'ing it), and if not - return `os.getpwd()` (with some warning?).
- completely eliminate reliance on PWD and use `getpwd()`.  That might ruin that use-case which might no longer exist, and would cause a somewhat confusing reported paths not matching what user expects (e.g. on OSX IIRC users' homes might be symlinked etc)

mih:
Had a long brain storming today re relativ paths. My conclusion was that we cannot really work with them, but have to transform them as soon as they come in, for various reasons. Not sure yet exactly what the conditions are, but I started looking into it.   re reporting: I do not yet see an elegant solution that could report paths in the same style as the input. One issue are the various reference directories we are using: Repos use the realpath (flyweight), datasets use the normpath. Paths that are exchanged between the two change their reference, ... It is a complciated beast.

mih:
I think this issue has served its purpose. Current datalad has a pretty consistent handling of paths (absolute an relative) now.

PR #2914, opened 2018-10-11
BF(workaround): switch to use os.getcwd if something odd detected with PWD env var

yarikoptic:
This pull request might help to address  #2910 but in an odd ad-hoc ugly fashion.  See description in patch/code

If we decide to adopt this at least as a temporary stop-gap measure, then  TODOs
- [x] tests

Testing locally with smth like
```shell
$> python -c 'import os; os.chdir("/tmp"); from datalad.utils import getpwd; print(getpwd(), os.getcwd())'
[WARNING] realpath of PWD=/home/yoh/proj/datalad/datalad-master is /home/yoh/proj/datalad/datalad-master whenever os.getcwd()=/tmp. From now on will be returning os.getcwd(). Directory symlinks in the paths will be resolved 
('/tmp', '/tmp')
```

### DIFF
--- a/datalad/utils.py
+++ b/datalad/utils.py
@@ -1272,7 +1272,16 @@ def getpwd():
     if _pwd_mode == 'cwd':
         return os.getcwd()
     elif _pwd_mode == 'PWD':
-        cwd = os.getcwd()
+        try:
+            cwd = os.getcwd()
+        except OSError as exc:
+            if "o such file" in str(exc):
+                # directory was removed but we promised to be robust and
+                # still report the path we might know since we are still in PWD
+                # mode
+                cwd = None
+            else:
+                raise
         try:
             pwd = os.environ['PWD']
             pwd_real = op.realpath(pwd)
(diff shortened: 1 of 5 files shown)


yarikoptic:
all the failures are known suspects and I see no reason to torture CI by merging master. Any feedback @datalad/developers ?

bpoldrack:
No objections by me. 

Except for a  poor gut feeling about the entire beast. But this is neither about this PR nor do I have a better idea. ;-)

datalad/utils.py
```diff
@@ -1216,25 +1216,99 @@ def updated(d, update):
...
+    shell and such functions as `os.chdir` and `os.getcwd` nohow use or modify
+    it, thus `os.getcwd()` returns path with links dereferenced.
+
+    While returning current working directory based on PWD env variable we
+    verify that the directory is the same as `os.getcwd()` after resolving all
+    symlinks.  If that verification fails, we fall back to always use
+    `os.getcwd()`.
+
+    Initial decision to either use PWD env variable or os.getcwd() is done upon
+    the first call of this function.
```
kyleam: The workaround is ugly, but this docstring does a nice job of explaining why it exists.

yarikoptic:
yeah, it is quite ugly but I also see no other way yet to handle it, and I do know that I ran into this issue (of us simply relying on wrong PWD value) while working in ipython notebook... so better this than nothing imho. merging, thanks!

PR #2915, opened 2018-10-12
Test neurodebian -devel fresh release of annex

yarikoptic:


yarikoptic:
ok, no harm is done, so will upload to neurodebian proper

Issue #2917, opened 2018-10-12
`wtf` should still work when git-annex is not installed


mih:
ATM it crashes: https://travis-ci.org/datalad/datalad-extension-template/jobs/440472721

PR #2918, opened 2018-10-12
BF: `wtf` should not crash when no annex is around (fixes gh-2917)

mih:
Given the uncontroversial nature of the PR, I will merge this as soon as all tests turn green.


mih:
The lines that are not covered by the tests, are actually covered by the extension template tests.

PR #2932, opened 2018-10-17
ENH: Make core reporting ready for paths as pathlib objects

mih:
Their str() is the path in it native representation.

This also adjusts test helpers that previously showed a relatively
uninformative message ("not serializable") where str() might
give some useful indication what is actually contained..


mih:
Buildbots seem broken -- failed within seconds after opening this PR.

yarikoptic:
I do not have strong understanding of the situation but sounds good enough, so ...

mih:
Arrg, thanks for the fix. Either new glasses or a bigger font seem necessary...

PR #2944, opened 2018-10-24
RF: Do not consider a symlink to the root of a dataset as the root of a dataset

mih:
I don't feel strongly about merging this PR, but in my exploration of symlink behavior I found this one problematic in -core. It is worked around in some places, and this PR merely want to see what would break.

Specifically one would see a change in behavior when looking to `add` a symlink that is pointing to the root of a subdirectory. With this change the parent of the symlink is the root, without this change it would be the root of the dataset that the symlink is pointing too. In the latter case (the present behavior) such a symlink is essentially transparent to datalad -- which is suboptimal IMHO.

### DIFF
--- a/datalad/utils.py
+++ b/datalad/utils.py
@@ -1546,10 +1546,17 @@ def get_dataset_root(path):
     The root path is returned in the same absolute or relative form
     as the input argument. If no associated dataset exists, or the
     input path doesn't exist, None is returned.
+
+    If `path` is a symlink or something other than a directory, the root dataset
+    containing its parent directory will be reported.  If none can be found, and
+    a symlink at `path` is pointing to a dataset, `path` itself will be
+    reported as the root.
     """
     suffix = '.git'
-    if not isdir(path):
-        path = dirname(path)
+    altered = None
+    if op.islink(path) or not op.isdir(path):
+        altered = path
+        path = op.dirname(path)
     apath = abspath(path)
     # while we can still go up
     while psplit(apath)[1]:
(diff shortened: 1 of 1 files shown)


bpoldrack:
Agree.

mih:
Very interesting that there is no failure with this change. The single buildbot failure is a fluke (certs issue). If we can have this merged, I do not need to fork this function in revolution.

datalad/utils.py
```diff
@@ -1546,10 +1546,17 @@ def get_dataset_root(path):
...
+    reported as the root.
     """
     suffix = '.git'
-    if not isdir(path):
-        path = dirname(path)
+    altered = None
+    if op.islink(path) or not op.isdir(path):
+        altered = path
+        path = op.dirname(path)
     apath = abspath(path)
```
yarikoptic: may be it worked even before because `abspath` doesn't dereference symlinks anyways:
```shell
(git-annex)hopa:~/datalad[master]
$> ls -ld dataset_description.json
lrwxrwxrwx 1 yoh yoh 42 Oct 26 11:02 dataset_description.json -> openfmri/ds000001/dataset_description.json

$> readlink -f dataset_description.json
/home/yoh/datalad/openfmri/ds000001/dataset_description.json

$> python -c 'from os.path import abspath; print(abspath("dataset_description.json"))'                                      
/home/yoh/datalad/dataset_description.json
```

that is why it is handy to have unittests which actually demonstrate the problem before "fixing"  it?

mih:
I'll take that as a "rather not merge".

yarikoptic:
That wasn't said... What inspired this change?  is there a usecase where it was providing incorrect result somehow?

mih:
Nothing beyond what is in the description. Let's not worry about it, I am running with a fork already and the changes are increasingly intertwined.

yarikoptic:
oh, there is finally a fork (branch) of the datalad-revolution?  or you call that extension a fork?

mih:
Neither of the two. As stated above, I have a fork of that function. I need that function but I need it two follow a specific concept of symlink handling, rather than what -core does (which I cannot verbalize).

PR #2946, opened 2018-10-24
BF: Reevaluation of Dataset properties

### LINKED ISSUE #2942: Two unwanted changes within .datalad/config and datalad/.gitattributes when rerunning datalad create --force
#### What is the problem?
Subsequent changes to an existing datalad dataset via datalad create --force 

1. change the dataset id
2. unnecessarily add the line ` 'metadata/aggregate* annex.largefile=nothing ` in .gitattributes again

`datalad create foo`
`datalad create --force --text-no-annex foo`

`tig` shows:  

`diff --git a/.datalad/config b/.datalad/config`
`index a8b1392..a530b30 100644`
`--- a/.datalad/config`
`+++ b/.datalad/config`
`@@ -1,2 +1,2 @@`
` [datalad "dataset"]`
`-       id = 65f3c046-d6fc-11e8-9873-40a3cc1bd9a6`
`+       id = 6f267582-d6fc-11e8-bc04-40a3cc1bd9a6`
`diff --git a/.gitattributes b/.gitattributes`
`index c3aaefe..57c0a11 100644`
`--- a/.gitattributes`
`+++ b/.gitattributes`
`@@ -1,3 +1,5 @@`

` * annex.backend=MD5E`
`+**/.git* annex.largefiles=nothing`
`+* annex.largefiles=(not(mimetype=text/*))`
` **/.git* annex.largefiles=nothing`
`\ No newline at end of file`

datalad 0.10.3.1.dev166

@yarikoptic please add anything that I stated unclear or wrong

bpoldrack:
Reevaluate properties when physically underlying repo changes, ceases to exist or comes into existence.
Note, that Dataset is supposed to represent a path reference to a repository that may not even exist yet, whereas *Repo classes are supposed to represent the actual repository. This leads to the Repos being flyweights wrt to realpath and Dataset being flyweight wrt the unresolved path!

There's a sideeffect to be mentioned: Previously it was possible to read/write `.datalad/config` via `Dataset.config` even if there was no repository (yet) at all. This is not the case anymore and I think this makes sense. Ultimately the previous behavior suggests, that you can store config at "dataset-level" (Parameter `where='dataset') while there is no dataset. This is misleading and I don't see a reason for it. However, a `Dataset` instance without an underlying physical repository still provides access to user and system level config, of course.

Also:
Closes #2942 

### Changes
- [x] reevaluate `Dataset.repo`
- [x] reevaluate `Dataset.config`
- [x] reevaluate `Dataset.id`
- [x] This change is complete

### DIFF
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -1687,7 +1687,7 @@ class GitRepo(RepoInterface):
                             shell=None, expect_fail=False,
                             check_fake_dates=False,
                             index_file=None,
-                            update_tree=False):
+                            updates_tree=False):
         """Allows for calling arbitrary commands.
 
         Helper for developing purposes, i.e. to quickly implement git commands
@@ -1699,7 +1699,7 @@ class GitRepo(RepoInterface):
         files: list of files
         cmd_str: str or list
           arbitrary command str. `files` is appended to that string.
-        update_tree: bool
+        updates_tree: bool
           whether or not command updates the working tree. If True, triggers
           necessary reevaluations like self.config.reload()
 
@@ -1751,7 +1751,7 @@ class GitRepo(RepoInterface):
                                      paths=ignored.groups()[0].splitlines())
             raise
 
-        if update_tree:
+        if updates_tree:
             lgr.debug("Reloading config due to supposed working tree update")
             self.config.reload()
 
@@ -2125,7 +2125,7 @@ class GitRepo(RepoInterface):
             cmd += options
         cmd += [str(name)]
 
-        self._git_custom_command('', cmd, expect_stderr=True, update_tree=True)
+        self._git_custom_command('', cmd, expect_stderr=True, updates_tree=True)
 
     # TODO: Before implementing annex merge, find usages and check for a needed
     # change to call super().merge


datalad/distribution/dataset.py
```diff
@@ -218,8 +219,10 @@ def repo(self):
                     lgr.log(5,
                             "Oops -- guess on repo type was wrong?: %s",
                             exc_str(exc))
-                    pass
+                    self._repo = None
```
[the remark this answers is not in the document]
bpoldrack: > eh, and we really can't do regular lazy evaluation here?

We never did. Run that same command on master and watch the exact same thing happening.

datalad/distribution/dataset.py
```diff
@@ -236,16 +239,19 @@ def id(self):
...
         str
           This is either a stored UUID, or `None`.
         """
-        if self._id is None:
-            # if we have one on record, stick to it!
-            self.config.reload()
-            self._id = self.config.get('datalad.dataset.id', None)
-        return self._id
+
+        return self.config.get('datalad.dataset.id', None)
```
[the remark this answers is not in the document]
bpoldrack: I don't see a reason to reload when we access `Dataset.id`. The important part is to ask the correct `ConfigManager` instance instead of assuming that it never changes. In fact the `reload()` before was quite pointless, since the only time it was called was the first access to `id` which is by no means a significant or even valid point in time regarding the properties of the dataset. Try this on master:

```
$> python -c 'from datalad.api import Dataset; ds = Dataset("/tmp/notexistent"); print(ds.id)'
  File "/home/ben/Development/datalad/datalad/distribution/dataset.py", line 246, in id
    self.config.reload()
  File "/home/ben/Development/datalad/datalad/distribution/dataset.py", line 261, in config
    self._cfg = ConfigManager(dataset=self, dataset_only=False)
  File "/home/ben/Development/datalad/datalad/config.py", line 204, in __init__
    self.reload(force=True)
  File "/home/ben/Development/datalad/datalad/config.py", line 258, in reload
    stdout, stderr = self._run(run_args, log_stderr=True)
  File "/home/ben/Development/datalad/datalad/config.py", line 526, in _run
    out = self._runner.run(['git', 'config'] + args, **kwargs)
  File "/home/ben/Development/datalad/datalad/cmd.py", line 673, in run
    cmd, env=self.get_git_environ_adjusted(env), *args, **kwargs)
  File "/home/ben/Development/datalad/datalad/cmd.py", line 487, in run
    stdin=stdin)
  File "/usr/lib/python3.5/subprocess.py", line 676, in __init__
    restore_signals, start_new_session)
  File "/usr/lib/python3.5/subprocess.py", line 1282, in _execute_child
    raise child_exception_type(errno_num, err_msg)
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/notexistent'
```
And even if we would fix for not crashing it would need to report `None` as its id. Then call `ds.create()` and `ds.id` again with old implementation of `id`. It would still be `None`, which is wrong. See also: #2948

datalad/distribution/dataset.py
```diff
@@ -255,12 +261,20 @@ def config(self):
...
+            # dataset level, but only at user/system level
+            # However, if this was the case before as well, we don't want a new
+            # instance of ConfigManager
+            if self._cfg_bound in (True, None):
+                self._cfg = ConfigManager(dataset=None, dataset_only=False)
+                self._cfg_bound = False
+
+        else:
+            self._cfg = self.repo.config
+            self._cfg_bound = True
```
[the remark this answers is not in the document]
bpoldrack: Yes, but it's just a pointer assignment. The only thing really not lazy here is `repo` (and yes -that's a performance issue). But this is not to fix within `config`. Here we need to ensure we are pointing to the correct `ConfigManager` instance. If `repo` changes, we get an entirely inconsistent `Dataset.config` otherwise.

datalad/distribution/dataset.py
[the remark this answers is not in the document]
bpoldrack: Re 1 and 2: Agree.

datalad/distribution/tests/test_dataset.py
```diff
@@ -323,3 +323,107 @@ def test_Dataset_flyweight(path1, path2):
...
+    assert (third_config is not forth_config)
+    assert (ds.id is not None)
+    assert (ds.id != first_id)
+    second_id = ds.id
+
+    # While os.symlink does work on windows (since vista), os.path.realpath
+    # doesn't resolve such symlinks. This has all kinds of implications.
+    # Hopefully this can be dealt with, when we switch to using pathlib
+    # (see datalad-revolution).
+    if not on_windows:
```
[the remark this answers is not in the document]
bpoldrack: Agree.

datalad/distribution/dataset.py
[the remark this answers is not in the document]
bpoldrack: Also note, that `self.config.get('datalad.dataset.id', None)` does NOT mean to query from FS. `ConfigManager` caches. This might look more expensive than it is. The actual cost comes indirectly via `repo` (triggered by `config`).

datalad/distribution/dataset.py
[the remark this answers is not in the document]
bpoldrack: This also means, that this:

> (so previous value of ._repo gets entirely forgotten/overwritten)

is not correct. It would be the case if we set it to `None` before. But since we don't, the line
```python
self._repo = cls(self._path, create=False, **kw)
```
will return the exact same object we started with, if nothing changed.

bpoldrack:
However, I'm looking into making it cheaper. We actually do check validity of the repo several times, where once was sufficient. We do it explicitly in `Dataset.repo`, then the flyweight engine does it and in case the actual constructors of `*Repo` are called, it is done again, making it way costlier than it needs to be. I'll try to address at least the redundancy of `Dataset.repo` and flyweight mechanics within this PR.

bpoldrack:
FTR: I'm close to a significant performance improvement. Still have an issue with catching the case, that there was a valid repo before and is now, but wasn't in between (recreation). Dataset fails to reevaluate `repo` in this case and points to the wrong object (which has implications => wrong `ConfigManager` used). Problem is: Did not figure out yet, why this even worked before! ;-)

bpoldrack:
Okay. That's it for now, I guess. Comparison to current master:

```
% asv compare 84f44bc882f01f4e3a4c28bff92a6cb7e6f1fc69 884492538644c43b24cbe2792ccd7c028ab18033 --machine tree     

All benchmarks:

    before     after       ratio
  [84f44bc8] [88449253]
      1.11s      1.09s      0.99  api.supers.time_createadd
      1.08s      1.07s      0.99  api.supers.time_createadd_to_dataset
      3.16s      3.17s      1.00  api.supers.time_installr
   127.22ms   126.12ms      0.99  api.supers.time_ls
      1.12s      1.12s      1.00  api.supers.time_ls_recursive
      1.27s      1.27s      1.00  api.supers.time_ls_recursive_long_all
      1.80s      1.81s      1.01  api.supers.time_remove
   351.43ms   360.70ms      1.03  api.testds.time_create_test_dataset1
      2.36s      2.35s      1.00  api.testds.time_create_test_dataset2x2
     2.04ms     2.05ms      1.01  core.runner.time_echo
     2.36ms     2.36ms      1.00  core.runner.time_echo_gitrunner
   508.62ms   506.12ms      1.00  core.startup.time_help_np
   101.02ms   100.66ms      1.00  core.startup.time_import
   456.64ms   446.75ms      0.98  core.startup.time_import_api
```
Please note two things:
- This is all about the speed of `Dataset.repo` evaluation. If you consider, that we now do it significantly more often, due to the implicit access to `repo` via `config`, not getting slower overall is quite something.
- There's more that can be done about it, but it should involve the entire call path from `Dataset.repo` to Flyweight to `AnnexRepo.__init__` and `GitRepo.__init__` (plus probably some RF'ing for `is_valid_repo`). So, that's a topic for a different PR, I think.

yarikoptic:
Cool, I will have a look
I guess the idea to bind .repo in functions to a local variable then still might be quite helpful, or you are planning to work through the stack to see if anything could be done (in a separate pr)?

bpoldrack:
> I guess the idea to bind .repo in functions to a local variable then still might be quite helpful, or you are planning to work through the stack to see if anything could be done (in a separate pr)?

I do plan to do so, since I started looking into it anyway.
However, this can only lead to further speed up of the validation. Using local variables in functions will remain a good idea. May be I should add that to the docstring of `repo`.

datalad/distribution/tests/test_dataset.py
```diff
@@ -323,3 +323,112 @@ def test_Dataset_flyweight(path1, path2):
...
+
+
+@with_tempfile
+def test_property_reevaluation(repo1):
+
+    from os.path import lexists
+    from datalad.tests.utils import ok_clean_git
+
+    ds = Dataset(repo1)
+    assert (ds.repo is None)
```
kyleam: pytest leakage?  Shouldn't these asserts (and their unnecessary parentheses :] ) be converted to the nose assert helpers to get more informative errors?

datalad/distribution/tests/test_dataset.py
```diff
@@ -323,3 +323,112 @@ def test_Dataset_flyweight(path1, path2):
...
+    assert (ds.config is not None)
+    first_config = ds.config
+    assert (ds._cfg_bound is False)
+    assert (ds.id is None)
+
+    ds.create()
+    ok_clean_git(repo1)
+    # after creation, we have `repo` and `config` was reevaluated to point
```
kyleam: A comma would help reader's parse this: "have `repo`**,** and"

bpoldrack:
FTR: nd14_04 failure seems to be the usual one, but nd16_04 failure is

```======================================================================
ERROR: datalad.support.tests.test_annexrepo.test_AnnexRepo_addurl_to_file_batched
    self.test(*self.arg)
  File "/home/buildbot/datalad-pr-docker-dl-nd16_04/build/datalad/tests/utils.py", line 418, in newfunc
    return t(*(arg + (d,)), **kw)
  File "/home/buildbot/datalad-pr-docker-dl-nd16_04/build/datalad/tests/utils.py", line 530, in newfunc
  File "/home/buildbot/datalad-pr-docker-dl-nd16_04/build/datalad/tests/utils.py", line 591, in newfunc
  File "/home/buildbot/datalad-pr-docker-dl-nd16_04/build/datalad/support/tests/test_annexrepo.py", line 1036, in test_AnnexRepo_addurl_to_file_batched
    ar.add_url_to_file(testfile, testurl, batch=True)
  File "/home/buildbot/datalad-pr-docker-dl-nd16_04/build/datalad/support/gitrepo.py", line 228, in newfunc
    return func(self, file_new, *args, **kwargs)
  File "/home/buildbot/datalad-pr-docker-dl-nd16_04/build/datalad/support/annexrepo.py", line 2147, in add_url_to_file
    % (url, str(out_json)))
AnnexBatchCommandError: AnnexBatchCommandError: command 'addurl'
Error, annex reported failure for addurl (url='http://127.0.0.1:40389/about.txt'): {'command': 'addurl', 'file': 'about.txt', 'success': False}

----------------------------------------------------------------------======================================================================
ERROR: datalad.support.tests.test_annexrepo.test_AnnexRepo_addurl_to_file_batched
    self.test(*self.arg)
  File "/home/buildbot/datalad-pr-docker-dl-nd16_04/build/datalad/tests/utils.py", line 418, in newfunc
    return t(*(arg + (d,)), **kw)
  File "/home/buildbot/datalad-pr-docker-dl-nd16_04/build/datalad/tests/utils.py", line 530, in newfunc
  File "/home/buildbot/datalad-pr-docker-dl-nd16_04/build/datalad/tests/utils.py", line 591, in newfunc
  File "/home/buildbot/datalad-pr-docker-dl-nd16_04/build/datalad/support/tests/test_annexrepo.py", line 1036, in test_AnnexRepo_addurl_to_file_batched
    ar.add_url_to_file(testfile, testurl, batch=True)
  File "/home/buildbot/datalad-pr-docker-dl-nd16_04/build/datalad/support/gitrepo.py", line 228, in newfunc
    return func(self, file_new, *args, **kwargs)
  File "/home/buildbot/datalad-pr-docker-dl-nd16_04/build/datalad/support/annexrepo.py", line 2147, in add_url_to_file
    % (url, str(out_json)))
AnnexBatchCommandError: AnnexBatchCommandError: command 'addurl'
Error, annex reported failure for addurl (url='http://127.0.0.1:40389/about.txt'): {'command': 'addurl', 'file': 'about.txt', 'success': False}
```

ATM I have no idea how this could possibly be related.

yarikoptic:
I will restart 16.04 (when get to work) just to make sure that it was a fluke

yarikoptic:
ok, [rerun](http://smaug.datalad.org:8020/builders/datalad-pr-docker-dl-nd16_04/builds/3056/steps/nosetests/logs/stdio) showed that the test failure was a fluke (now fails with the keyring issue, which is fixed in master -- buildbot doesn't do master merge before running)

datalad/distribution/dataset.py
```diff
@@ -187,22 +189,61 @@ def path(self):
...
+        # (or none). But it's still faster for the vast majority of cases.
+        #
+        # TODO: Dig deeper into it and melt with new instance guessing. This
+        # should also involve to reduce redundancy of testing such things from
+        # within Flyweight.__call__, AnnexRepo.__init__ and GitRepo.__init__!
+        #
+        # Also note, that this could be forged into a single big condition, but
+        # that is hard to read and we should be well aware of the actual
+        # criteria here:
+        if self._repo is not None and realpath(self.path) == self._repo.path:
```
yarikoptic: shouldn't it be `_repo.realpath` then? wouldn't it make sense to store `.realpath` for the dataset as well to avoid realpathing upon each call?

datalad/distribution/dataset.py
```diff
@@ -187,22 +189,61 @@ def path(self):
...
+            # we got a repo and path references still match
+            if isinstance(self._repo, AnnexRepo):
+                # it's supposed to be an annex
+                if self._repo is AnnexRepo._unique_instances.get(
+                        self._repo.path, None) and \
+                        AnnexRepo.is_valid_repo(self._repo.path,
+                                                allow_noninitialized=True):
```
yarikoptic: for what usecases this check (`is_valid_repo`) is needed?  if it is already AnnexRepo, I guess such a check was done before and should remain valid, or should it not?

datalad/distribution/dataset.py
```diff
@@ -187,22 +189,61 @@ def path(self):
...
+                    # it's still the object registered as flyweight and it's a
+                    # valid annex repo
+                    return self._repo
+            elif isinstance(self._repo, GitRepo):
+                # it's supposed to be a plain git
+                if self._repo is GitRepo._unique_instances.get(
+                        self._repo.path, None) and \
+                        GitRepo.is_valid_repo(self._repo.path) and not \
```
yarikoptic: the same here -- when could it already be `GitRepo` but no longer a valid git repo?

datalad/distribution/dataset.py
```diff
@@ -187,22 +189,61 @@ def path(self):
...
+                        self._repo.is_with_annex():
+                    # it's still the object registered as flyweight, it's a
+                    # valid git repo and it hasn't turned into an annex
+                    return self._repo
```
yarikoptic: in all of the above, if we did not care about auto Git->Annex upgrade (which is irrelevant if it is already an annex), wouldn't it be sufficient to store in Dataset  `._repo.inode` when assigning new `._repo` (e.g. as `._repo_inode`) and comparing it for a given path upon invocation?  should probably be faster than all of the checks above...?

datalad/interface/tests/test_run.py
```diff
@@ -352,6 +352,7 @@ def test_rerun_just_one_commit(path):
     # in a repo" case.
     ds.repo.checkout("orph", options=["--orphan"])
     ds.repo.repo.git.reset("--hard")
+    ds.repo.config.reload()
```
yarikoptic: should all the smarty checks detect the necessity for config reload somehow?  how did it work before without? (worth adding a comment why needed may be?)

datalad/support/exceptions.py
```diff
@@ -261,6 +261,11 @@ def __init__(self, id_, msg=None):
         self.msg = msg
 
 
+class InvalidAnnexRepositoryError(RuntimeError):
```
yarikoptic: I wonder if might make sense to get `InvalidRepositoryError` to be a parent class for `InvalidAnnex...` and `InvalidGit...` ?

datalad/support/gitrepo.py
```diff
@@ -2117,6 +2117,8 @@ def checkout(self, name, options=None):
         cmd += [str(name)]
 
         self._git_custom_command('', cmd, expect_stderr=True)
+        # Note: force=True shouldn't be needed, since mtime changes on checkout:
+        self.config.reload()
```
yarikoptic: If `.config` is not detecting the change (I thought it does!?) I wonder if to be safe we should just do it for every `_git_custom_command`?  or at least also in those other which adjust current worktree: `merge`, `cherry_pick`, `pull`, etc?

datalad/distribution/dataset.py
bpoldrack: Re 1: _repo.path is realpath already. The actual constructor is only called by flyweight with already resolved path. `GitRepo.realpath` is not needed at all and can be removed, I think.

Re 2: Nope. If `Dataset.path` is a symlink, it could change during lifetime of the Dataset instance. See the tests I added in `test_dataset.py`. If we only resolve it once and store it, we would fail to notice that change.

datalad/distribution/dataset.py
bpoldrack: Remove/uninstall. Or something went really wrong and you killed .git. The point is: Just because something was there, it doesn't mean it's there forever.

datalad/distribution/dataset.py
bpoldrack: > in all of the above, if we did not care about auto Git->Annex upgrade

We did. Line 235: `self._repo.is_with_annex()`.

Re inode: That might be a way to clarify some things faster, yes. But still not all of it. If you uninstall a subdataset, the directory the Repo instance is pointing to, remains. But it's not a valid repo anymore. And: Remember - I want to do more about the performance of it. Just not in this PR.

datalad/support/exceptions.py
bpoldrack: Agree.

datalad/interface/tests/test_run.py
bpoldrack: It worked by accident. ;-)

Generally yes. Certain *Repo methods should know themselves to reload config. But here GitPython's reset is called, which is why the reload needs to be called in addition. What should be done is: Have a `GitRepo.reset`, that would call reload at the end and use this one here.

datalad/support/gitrepo.py
bpoldrack: It does detect based on mtime. But the detection happens within `reload` (if not called with `force=True`). Where else?

Yes, other methods, that might change `.datalad/config` should also call `reload`.

datalad/distribution/dataset.py
mih: I am not following this close enough to make a meaningful contribution to the discussion, but `inode` sounds like it won't work on Windows. While I am working on avoiding GitPy code in -revolution already, introducing it this deep in the machine would make it substantially harder.

datalad/support/gitrepo.py
yarikoptic: So, shouldn't this be done here?  and if based on mtime I would just call in each `_git_custom_command` to avoid duplication etc

datalad/distribution/dataset.py
yarikoptic: well, harddrive might disappear too!  Should we then constantly sample for the `/` to still be there? my point is that all these operations cost time, and since if "something went really wrong and you killed .git" you are then screwed anyways, it would be nice to avoid wasting that time

bpoldrack: > if "something went really wrong and you killed .git" you are then screwed anyways

Sure, but you still want to fail reasonably in that case instead of pretending everything was in order, don't you? And apart from that, as I pointed out before, even a simple uninstall would cause `self.repo` to be invalid if we wouldn't update that property. Have a look at the tests I added.
Plus: As far as `self.repo` is concerned, we do it in master as well, but even MORE expensive! This is faster than before! (And yes, it could be even faster and I'll do it) The entire point of this PR is, that we did it only for `repo` and thus decoupling `config` and `id`, since we didn't do it for them. In master we simply end up in an inconsistent state.
I can't see how this discussion would prevent merging. If we want to do it differently altogether (and change the already existing tests that would fail ATM if this was done differently), we can have that discussion, but then we need to come up with (and agree on) a solution that at the very least results in a consistent state.

In case I somewhat misread your objection and it's really just about figuring out, what exact conditions really need checking: Yes, there's room for improvement. I just really want to do it in a dedicated PR, since this involves tuning of what is done in flyweight and in the *Repo constructors and in the `is_valid_repo` methods (particularly in case of annex). I'm happy to discuss the details of that, since we do redundant things indeed. But I need this bug fixed first and further optimize the validation second.

datalad/support/gitrepo.py
bpoldrack: Well, it could be done within `_git_custom_command` but that would be way more expensive, than figuring out which methods really need it. What's the point of checking mtimes of config files after calling `git ls-files`?

yarikoptic: I think we should make it more declarative/centralized because may be other things would need to be done in such cases besides reloading config.  I would have added `updates_tree=False` kwarg to the `_git_custom_command`, and then added the logic to call this reload in there conditioned on that argument.  Necessary methods would then specify `updates_tree=True` where applicable.

bpoldrack: Ok. Will do.

bpoldrack:
FTR: appveyor fails in direct mode in `test_datasets_datalad_org`. This appears to be unrelated.

datalad/support/gitrepo.py
```diff
@@ -1697,7 +1698,10 @@ def _git_custom_command(self, files, cmd_str,
         ----------
         files: list of files
         cmd_str: str or list
-            arbitrary command str. `files` is appended to that string.
+          arbitrary command str. `files` is appended to that string.
+        update_tree: bool
```
yarikoptic: `update_tree` is imho somewhat misleading suggesting to do something ("update the tree"), if anything let it be `updates_tree`, I will quickly push that change and also add its use in other relevant commands

yarikoptic:
pushed 6f256eae8e0ffeb4f65ef50e1242c4871bd33aea which introduces use of that `updates_tree` to other relevant command calls

yarikoptic:
here is the overall impact I see in comparison with the master
```shell
$> asv compare -m hopa 34d4360e4b984e5f04fed183338e7a2bdf5f9163 6f256eae8e0ffeb4f65ef50e1242c4871bd33aea
                                                               
All benchmarks:

     [34d4360e]       [6f256eae]
            1.71s            1.86s     1.09  api.supers.time_createadd
+           1.70s            1.97s     1.16  api.supers.time_createadd_to_dataset
            4.89s            4.59s     0.94  api.supers.time_installr
          145±4ms          153±5ms     1.05  api.supers.time_ls
            1.52s            1.54s     1.01  api.supers.time_ls_recursive
            1.83s            1.77s     0.97  api.supers.time_ls_recursive_long_all
            2.44s            2.46s     1.01  api.supers.time_remove
+         498±3ms          565±8ms     1.13  api.testds.time_create_test_dataset1
+           3.22s            4.23s     1.32  api.testds.time_create_test_dataset2x2
       2.41±0.2ms      1.87±0.03ms    ~0.78  core.runner.time_echo
-      2.92±0.3ms      2.05±0.04ms     0.70  core.runner.time_echo_gitrunner
+         461±8ms          530±0ms     1.15  core.startup.time_help_np
+        82.6±2ms          105±3ms     1.27  core.startup.time_import
+         410±2ms        462±0.8ms     1.13  core.startup.time_import_api
```
not sure yet why import/help have notable impact - should analyze, may be some rogue import needs to be delayed or some structure not created?

mih:
I am concernd by the update trees thing and how it is used. I fear that I config reload is much slower than one would hope in some cases (nas drive, etc). And now we are doing it unconditionally all over the place. Even to support esotheric scenarios like .datalad being. submodule!?

Those config reloads are only concerned with .datalad/config. Why not test for this specifically, instead of bruteforcing config reloads that parse much more than this file?

bpoldrack:
> here is the overall impact I see in comparison with the master

Well, that's why I didn't want it to be default and was stating that we need to carefully consider what methods actually need it. This is just another topic that should be addressed on its own from my point of view. There are additional reloads that should be reconsidered, if we do it that way.
I really don't know why this PR has to grow to resolve everything it shines a light on.

I'd just go back to https://github.com/datalad/datalad/pull/2946/commits/7fda144547a374f71e491f529782e88da62c33b6 and get that merged. Speeding up `Dataset.repo` and ironing out config reloads are separate issues, which should be addressed separatly.

If we can't agree on that, I need to leave it and bring this into revolution instead. I need that fix, the PR actually is about.

yarikoptic:
Sure, bring back or remove triggering for "isoteric" situations. I felt that many others not mentioned originally were quite warranted but that is ok, we can just wait for bug reports to add them back

PR #2950, opened 2018-10-26
ENH: add annex key to the extracted datalad-core metadata record

yarikoptic:
This is to possibly facilitate additional search functionality such as requested in
  https://neurostars.org/t/return-hash-and-url-for-every-nifti-file-in-datalad-super-dataset/2733
in case if a known checksum matches the backend used by git annex

I bet it is done better in the rev- version of the universe.  Here I chose to provide an ugly hack to avoid secondary call to lookup key for a file.

Should break some tests, possibly in datalad-neuroimaging since output record is extended now.

- [x] may be it should be just a `key` or `annex_key` since I think `-` is not used in the metadata fields

### DIFF
--- a/datalad/metadata/extractors/datalad_core.py
+++ b/datalad/metadata/extractors/datalad_core.py
@@ -126,7 +126,7 @@ class MetadataExtractor(BaseMetadataExtractor):
             # just first one.  `add_key` was added to avoid separate call to
             # get_file_key which would need to be ran separately
             if whereis:
-                meta['annex-key'] = list(whereis.values())[0]['key']
+                meta['annex_key'] = list(whereis.values())[0]['key']
             yield (file, meta)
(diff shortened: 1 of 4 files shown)


yarikoptic:
@mih @bpoldrack @kyleam what do you think? should `annex_key` be a part of `datalad_core` or `annex` (just a `key` there), or may be even at the top level alongside with `path`, `type`?

mih:
I would prefer to have annex stuff done by the annex extractor.

yarikoptic:
ok. while at it - do you see any other information from annex wortwhile extracting? we already do metadata, and urls (as part of the `datalad_core`). Could be the list of clones (remotes) where file is present (too variable imho).

yarikoptic:
superseded by #2952

PR #2952, opened 2018-10-26
ENH: add annex key to the extracted "annex" metadata record

yarikoptic:
- Replaces #2950 with more native placement of key record
- adjusts `AnnexRepo.get_file_key` with explicit `batch=None` option to trigger batch mode of operation

cons: with this change (and as test fixups show) now all annexed files will get some metadata record within "annex" section.  Time will show if any notable impact on size/performance

### DIFF
--- a/datalad/metadata/extractors/annex.py
+++ b/datalad/metadata/extractors/annex.py
@@ -36,7 +36,8 @@ class MetadataExtractor(BaseMetadataExtractor):
             label='Annex metadata extraction',
             unit=' Files',
         )
-        if not isinstance(self.ds.repo, AnnexRepo):
+        repo = self.ds.repo   # OPT: .repo could be relatively expensive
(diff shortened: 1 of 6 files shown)


yarikoptic:
FWIW @chrisfilo, with this RF search would look more like
```shell
$ datalad_ -f '{path}: {metadata[annex][key]} {metadata[datalad_core][url]}' -c datalad.search.index-egrep-documenttype=all search path:.*\.nii.gz annex.key:MD5E.* datalad_core.url:.
```
to limit to only the files having any url and being under MD5E backend (although even MD5 would suffice if someone uses such one ;-) i.e. without extensions within keyname)

yarikoptic:
@mih any comments/recommendations/merge?

yarikoptic:
well, since the government wants it, how could we refuse? ;-)  no objects were stated, so must be the best piece of code out there in the wild!

PR #2956, opened 2018-10-28
TST: -revolution extension test run

mih:
For travis, to not have things randomly break.


mih:
Looks good from my POV

mih:
All good, this should be uncontroversial.

PR #2957, opened 2018-10-29
RF: repo creation into a dedicated helper

mih:
Eases customization in derived classes.


mih:
It seems the test failures are due to a credential issue that is totally unrelated.

PR #2983, opened 2018-11-07
BF+TST: do not pass "dataset" kwarg into a bound .get call

### LINKED ISSUE #2982: install recursive into existing dataset
`install --recursive` is broken. It works fine if just calling `datalad install -r -s whatever to/whereever`.
But it doesn't if you called with an additional `-d some`. First level is installed, then it fails with

```
[ERROR  ] get() got an unexpected keyword argument dataset [dataset.py:apply_func:486] (TypeError)
```

I didn't look deeply into it yet, but this suggests, that `get` is called with `dataset=...` while already being bound to a dataset instance (or sth similar). I guess, we never really considered this use case and when calling `get` there's a confusion of what's the reference dataset to operate on.

FTR: Of course the same thing can still be achieved. `datalad install .d . -s toplevel` and only afterwards calling `datalad install toplevel/sub` works fine.

yarikoptic:
Closes #2982

### DIFF
--- a/datalad/distribution/install.py
+++ b/datalad/distribution/install.py
@@ -363,7 +363,7 @@ class Install(Interface):
         # Now, recursive calls:
         if recursive or get_data:
             # dataset argument must not be passed inside since we use bound .get
-            # It is ok to do "inplace" as long as we still return right right
+            # It is ok to do "inplace" as long as we still return right
             # after the loop ends
             common_kwargs.pop('dataset', '')
             for r in destination_dataset.get(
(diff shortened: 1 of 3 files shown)


datalad/distribution/install.py
```diff
@@ -362,6 +362,10 @@ def __call__(
 
         # Now, recursive calls:
         if recursive or get_data:
+            # dataset argument must not be passed inside since we use bound .get
+            # It is ok to do "inplace" as long as we still return right right
```
kyleam:
```suggestion
            # It is ok to do "inplace" as long as we still return right
```

yarikoptic:
merged with typo fix applied, thanks @kyleam !

bpoldrack:
Looks good indeed. Thanks!

PR #2993, opened 2018-11-12
ENH: run-procedure

bpoldrack:
### Changes
- [x] yield procedure names and help messages on `--discover` and `--help-proc`
- [x] custom result renderer to make use of it

### DIFF
--- a/datalad/interface/run_procedure.py
+++ b/datalad/interface/run_procedure.py
@@ -35,6 +35,7 @@ from datalad.support.exceptions import InsufficientArgumentsError
 from datalad.support.exceptions import NoDatasetArgumentFound
 
 from datalad.utils import assure_list
+import datalad.support.ansi_colors as ac
 
 # bound dataset methods
 from datalad.interface.run import Run
@@ -325,6 +326,8 @@ class RunProcedure(Interface):
         )
     )
 
+    result_renderer = 'tailored'
+
     @staticmethod
     @datasetmethod(name='run_procedure')
     @eval_results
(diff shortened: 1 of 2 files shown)


datalad/interface/run_procedure.py
```diff
@@ -456,3 +464,47 @@ def __call__(
...
+        from datalad.interface.utils import default_result_renderer
+
+        if res['status'] != 'ok':
+            # logging complained about this already
+            return
+
+        if 'procedure' not in res.get('action', ''):
+            # it's not our business
+            default_result_renderer(res)
+            return
```
kyleam: This bit made me pause.  The previous 'status' guard ignores not "ok" results because logging has already complained about it.  But then here, we're deciding that a result doesn't concern us and using the default renderer.  That suggests that a non-ok result above could also not concern us, yet we're assuming it was logged.

Should this guard be move above the previous 'status' one?

bpoldrack: This might be true. I'm still a bit confused myself about the entire rendering/logging/yielding implications. 
But: As far as I'm aware, non-ok is always logged, so even if the results comes from an internal command call, it would be logged. That's not a matter of `run-procedure`'s implementation, except that it only applies if `run-procedure` actually yields those results from internal calls. So, throwing away non-ok first, seems right to me including results that are not 'procedure` results.

bpoldrack: As an example:

```
% datalad run-procedure --help-proc some
[WARNING] Cannot find procedure with name 'some' [run_procedure(some)] 
```

If you look into the code, you'll find, that `run_procedure` doesn't log that on its own. It's just yielding a result with that message.

kyleam: > But: As far as I'm aware, non-ok is always logged, so even if the results comes from an internal command call, it would be logged.

Indeed.  Sorry for the confusion.

>  So, throwing away non-ok first, seems right to me including results that are not 'procedure` results.

Agreed.

PR #2994, opened 2018-11-13
ENH: added initial .zenodo.json, and its mentioning in PR template and CONTRIBUTING

yarikoptic:
For format of the file: http://developers.zenodo.org/

So far I have failed to
- find DOI for BMBF (in grants)
- clear description on either we could put spectific grant # after :: in grants
  entries, or just expand those entries with some other fields to point to
  the specific grant(s0

[ci: skip]

Dear @datalad/contributors  and @datalad/developers - please add/adjust/review your entries in .zenodo.json and push your changes into this PR

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -4,5 +4,6 @@ This pull request proposes
 
 ### Changes
 - [ ] This change is complete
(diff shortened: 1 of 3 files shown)


bpoldrack:
We should probably do that for all the extensions as well, I guess. And it might be about time I get myself an orcid ;-)

PR #2995, opened 2018-11-14
ENH: initial Singularity recipe to provide an image to build git-annex for bisection

yarikoptic:
Should provide a build environment to deploy on some other box (e.g. CentOS) so we could build and bisect git-annex

TODOs:
- [ ] supplement with the bisection script, which would mimic `git bisect` in API but would use this singularity image to `make standalone install DESTDIR=somewhere`  first and make that built version available for the bisection script run

"inspiration": http://git-annex.branchable.com/bugs/multiple_ssh_prompts__44___and_thread_blocked_indefinitely_in_an___63____63____63___transaction/#comment-89f62985af5dd49438a578851e4d541f

Attn @joeyh

### DIFF
--- a/tools/Singularity.git-annex-dev
+++ b/tools/Singularity.git-annex-dev
@@ -1,0 +1,25 @@
+Bootstrap:docker
+From:neurodebian:buster
+
+%post
+    echo "Configuring the environment"
+    apt-get -y update
+
+    # setup the container sources themselves
+    apt-get -y install eatmydata vim less devscripts
+
+    sed -i -e 's,^deb \(.*\),deb \1\ndeb-src \1,g' /etc/apt/sources.list /etc/apt/*.d/*.list \
+     && apt-get -y update \
+     && eatmydata apt-get -y build-dep git git-annex git-annex-standalone
+
+    # we need a UTF locale for DataLad to work properly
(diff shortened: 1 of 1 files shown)


tools/bisect-git-annex
```diff
@@ -0,0 +1,32 @@
...
+shift
+
+if [ ! -e "$SINGULARITY_IMAGE" ]; then
+    echo "I: trying to fetch singularity image with build env"
+    singularity pull --name "$SINGULARITY_IMAGE" shub://datalad/datalad:git-annex-dev
+fi
+
+SINGULARITY_IMAGE=$(readlink -f "$SINGULARITY_IMAGE")
+annex_gitver=$(git describe)
+logfile=$PWD/.git/logs/git-annex-build-$annex_gitver.log
```
kyleam: Probably doesn't matter in practice, but it seems inappropriate to store the log in git's reflog directory.  Why not use a temporary path?

yarikoptic: things get lost in temporary path since then you need to have a temporary filename etc.
but indeed, I thought this `logs/` it is more generic than reflogs -- just curious what else could there be other than HEAD?
anyways -- will now RF to use `mktemp -p '' git-annex-build-${annex_gitver}-XXXXXX.log` instead, testing now and will push.
With that I will consider this PR done - it already served its purpose ;-)

PR #3001, opened 2018-11-20
BF: Assume 50% safety margin for max cmdline length

yarikoptic:
Hit it again with current master ... not sure if this would have resolved it but it should ;)

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -2399,7 +2399,6 @@ class AnnexRepo(GitRepo, RepoInterface):
                      - len(command)
                      - sum((len(x) + 3) for x in annex_options)
                      - 4   # for '--' below
-                     - 50  # for safety since we are to add more options
                      ) // (maxl + 3)  # +3 for possible quotes and a space
                 )
                 file_chunks = generate_chunks(files, chunk_size)
--- a/datalad/utils.py
+++ b/datalad/utils.py
@@ -94,7 +94,15 @@ except Exception as exc:
     lgr.debug(
         "Failed to query SC_ARG_MAX sysconf, will use hardcoded value: %s",
         exc)
-lgr.debug("Maximal length of cmdline string: %d", CMD_MAX_ARG)
+# Even with all careful computations we do, due to necessity to account for
+# environment and what not, we still could not figure out "exact" way to
+# estimate it, but it was shown that 300k safety margin on linux was sufficient.
+# https://github.com/datalad/datalad/pull/2977#issuecomment-436264710
+# 300k is ~15%, so to be safe, we will just use up to 80% of the length
+CMD_MAX_ARG = int(0.8 * CMD_MAX_ARG)
+lgr.debug(
+    "Maximal length of cmdline string (adjusted for safety margin): %d",
+    CMD_MAX_ARG)
 
 #
 # Little helpers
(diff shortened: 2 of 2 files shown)


yarikoptic:
> I think this change is fine as is, particularly assuming it fixes real limits hit, but does it really make sense for the safety margin to scale with CMD_MAX_ARG?

As it was discovered on linux it should be 300k which is greater than the CMD_MAX_ARG itself on OSX or Windows ;)  that is why I decided to go for `%`

yarikoptic:
pushed a rewrite of my silly untested quick change -- this time must work ;-)

yarikoptic:
My reason is greed!
Annex probably generally doesn't invoke very long commands - probably passes . and not all the files from under current directory. 10k partitioning would result in unnecessarily large number of separate commands to run in our case

kyleam:
> 10k partitioning would result in unnecessarily large number of separate commands to run in our case

OK, I'll trust your gut that this is true in practice.

kyleam:
@kyleam
>  But if ~300k padding is required for a command on linux, isn't that command almost certainly going to fail elsewhere?

Hmm, there's still this.  You didn't answer, so I'm (perhaps incorrectly) assuming I'm not mistaken.  If so, it seems like it'd be better to drop to just drop to a very conservative absolute value for the maximum length if not on linux or `SC_ARG_MAX` is lower than a particular value.  Then, you wouldn't take the hit you're worried about when on linux in the usual conditions, but there's an increased chance that commands will succeed in the other scenarios.

yarikoptic:
> But if ~300k padding is required for a command on linux, isn't that command almost certainly going to fail elsewhere?

I am not sure what you mean here.
"elsewhere" where?  within git annex call itself? (that one has its own ways -- batch calls to git etc).
Since we (or at least I) haven't figured out why those 300k are needed (even after I accounted for env etc), this is all a heuristic.  Could be more concervative (fixed number, which might become more liberal for some OSs) or less (fixed '%')... if you like it more conservative '%' then which one -- 70? 60? ...   Going all the way down to 0.5% sounds a bit of an overkill for my liking.

kyleam:
> > But if ~300k padding is required for a command on linux, isn't that command almost certainly going to fail elsewhere?
> 
> I am not sure what you mean here.
> "elsewhere" where?

"elsewhere" was in direct contrast to the "linux" at the beginning of the sentence, so on other OSs.

yarikoptic:
ah, ok... I am not sure why it would "certainly" fail, I have no prior experiences of that kind ;-)

but if you want to bargain, what if we take  100% - 10240/32767 =~ 70% margin (i.e. have `0.3` multiplier) and still stay friends? ;)  This way at least we wouldn't go all the way to 99.5% margin on linux ;-)

kyleam:
> ah, ok... I am not sure why it would "certainly" fail, I have no prior experiences of that kind ;-)

"certainly" was bad word choice on my part.  I was just trying to reason through it.

> This way at least we wouldn't go all the way to 99.5% margin on linux ;-)

I'm confused by what that means given that above I said:

> If so, it seems like it'd be better to just drop to a very conservative absolute value for the maximum length if not on linux or `SC_ARG_MAX` is lower than a particular value. Then, you wouldn't take the hit you're worried about when on linux in the usual conditions, but there's an increased chance that commands will succeed in the other scenarios.

Anyway, I'm unconvinced a fixed scaling is a useful thing to do for low CMD_MAX_ARG values, but that's likely me being obtuse.  Proceed without me :]

yarikoptic:
it might be that more complex OS based decision making would be more robust indeed... ok, you made me paranoid, I will go for 50% margin and be done with it for now.
Ideally/eventually any of those ad-hoc ones would fail (unicode, etc) unless we figure out exact way to estimate it reliably while still accounting for possible limits being hit by tools we would call, or add some progressive splitting if we hit "maximal length reached"...

Issue #2841, opened 2018-09-18
subdataset's children record is not aggregated into superdataset's


yarikoptic:
#### What is the problem?
I am trying to aggregate metadata for a freshly added physionet.  When I am attempting to aggregate physinet subdataset's metadata it doesn't include physionet's subdataset's (`eegmmidb`) metadata:

```shell
(git)smaug:/mnt/btrfs/datasets/datalad/crawl[master]
$> datalad aggregate-metadata --update-mode all --incremental physionet
[INFO   ] Aggregate metadata for dataset /mnt/btrfs/datasets/datalad/crawl/physionet 
[INFO   ] Aggregate metadata for dataset /mnt/btrfs/datasets/datalad/crawl 
[INFO   ] Update aggregate metadata in dataset at: /mnt/btrfs/datasets/datalad/crawl/physionet 
aggregate_metadata(ok): /mnt/btrfs/datasets/datalad/crawl/physionet (dataset)
[INFO   ] Update aggregate metadata in dataset at: /mnt/btrfs/datasets/datalad/crawl 
aggregate_metadata(ok): /mnt/btrfs/datasets/datalad/crawl (dataset)
[INFO   ] Attempting to save 1058 files/datasets 
save(notneeded): /mnt/btrfs/datasets/datalad/crawl/physionet (dataset)
save(notneeded): /mnt/btrfs/datasets/datalad/crawl (dataset)
action summary:
  aggregate_metadata (ok: 2)
  save (notneeded: 2)
datalad aggregate-metadata --update-mode all --incremental physionet  15.82s user 32.13s system 130% cpu 36.641 total
(dev) 1 12262.....................................:Mon 17 Sep 2018 10:08:58 PM EDT:.
(git)smaug:/mnt/btrfs/datasets/datalad/crawl[master]
$> cat physionet/.datalad/metadata/aggregate_v1.json 
{
".":
{
"content_info":
"objects/99/cn-07e4520ee3f835ced0b044bb9baf2f.xz",
"datalad_version":
"0.10.2.dev243",
"dataset_info":
"objects/99/ds-07e4520ee3f835ced0b044bb9baf2f",
"extractors":
[
"datalad_core",
"annex"
],
"id":
"42dbd700-ba9c-11e8-86a9-002590f97d84",
"refcommit":
"91cbd9eba120d21c7fbfafe6ae8543dcf4f38b74"
},
"eegmmidb":
{
"content_info":
[...]

PR #3002, opened 2018-11-21
BF(RF?): do not aggregate dataset itself if paths for aggregation provided

yarikoptic:
Otherwise it causes negative side effect in --recursive mode that it would start
recursing all subdatasets.

This issue was mentioned in
https://github.com/datalad/datalad/issues/2841

Aggregation of current dataset could always be forced by providing '.' or just not providing paths.
If this PR causes some ideological breakage, it is not visible since no tests seems to get broken.


yarikoptic:
reverted one commit back -- will submit the other changes as a separate PR

yarikoptic:
@mih - what do you think?

yarikoptic:
ok, at least @mih doesn't have objections.  I bet his life will also get much better so I will merge

PR #3003, opened 2018-11-21
Test -devel version of git-annex

yarikoptic:
Replaces https://github.com/datalad/datalad/pull/2915

New version should fix parallel ssh access versions etc

If doesnt break anything, I would encourage you to try it out as well , and I will upload it to neurodebian asap


yarikoptic:
ok, looks good to me

PR #3008, opened 2018-11-23
RF: Allow for customization of change recording in `run`

mih:
This is admittedly a bit clunky. An alternative way of doing this would be to RF the `add()` call into the main command class, and have a helper method deal with it. In this scenario, it could simply be overwritten in a derived class, instead of having to introduce another argument. What would you prefer @kyleam?

Why is this needed? ATM `run` performs a hard-coded `add(recursive)`, and this adds the flexibility to tailor this behavior (non-recursive add, or more than just adding).

### DIFF
--- a/datalad/interface/run.py
+++ b/datalad/interface/run.py
@@ -529,11 +529,21 @@ def _execute_command(command, pwd, expected_exit=None):
     return cmd_exitcode or 0, exc
 
 
+def _save_outputs(ds, to_save, msg):
+    """Helper to save results after command execution is completed"""
+    return ds.add(
+        to_save,
+        recursive=True,
+        message=msg,
+        return_type='generator')
+
+
 def run_command(cmd, dataset=None, inputs=None, outputs=None, expand=None,
                 explicit=False, message=None, sidecar=None,
                 extra_info=None,
                 rerun_info=None, rerun_outputs=None,
-                inject=False):
+                inject=False,
+                saver=_save_outputs):
     """Run `cmd` in `dataset` and record the results.
 
     `Run.__call__` is a simple wrapper over this function. Aside from backward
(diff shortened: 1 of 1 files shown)


datalad/interface/run.py
```diff
@@ -559,6 +569,10 @@ def run_command(cmd, dataset=None, inputs=None, outputs=None, expand=None,
         preparation and command execution. In this mode, the caller is
         responsible for ensuring that the state of the working tree is
         appropriate for recording the command's results.
+    saver : callable, optional
+        Must take a dataset instance, a list of paths to save, and a
+        message string as arguments and must record any changes done
+        to any content matching an entry in the path list.
```
yarikoptic: Current one returns a generator of results records... May be with documenting here on what is expected from saver to be returned?

mih:
@kyleam I now think that this is not the best approach. The reason is that to actually by able to call a custom saver, I would have to replace the body of Run.__call__ as well, in order to get to specify this parameter. Possible of course, but implies that the more (unrelated) logic must be duplicated.

kyleam:
> The reason is that to actually by able to call a custom saver, I would have to replace the body of Run.**call** as well, in order to get to specify this parameter

I don't understand this comment.  I assumed the caller would be using `run_command`.

>  implies that the more (unrelated) logic must be duplicated.

Aside from the `run_command` call, most of `Run.__call__` is backward-compatibility kludges that I don't think a `run_command` caller should have to worry about.  

Perhaps I'm not understanding the specifics of how you plan to use `saver` though.

datalad/interface/run.py
mih: Added

mih:
@kyleam You convinced me. I added the missing aspect to the docs and left everything else as-is. Will merge now.

Issue #3030, opened 2018-11-29
`run` broken on windows


mih:
`run` is currently broken on windows. That doesn't really surprise anyone, but I think it is worth making it run.

So here is the (growing) list of things that don't work:

- [ ] crashes due to bytes-representation of message. See here for a demo https://ci.appveyor.com/project/mih/datalad-revolution/builds/20644761
    there is an explicit `assure_bytes()` call in the code that is the culprit here. Not sure what the reason is/was.
- [ ] test_run_failure uses sh cmd:  `echo x$(cat grows) > grows && false`  it looks as if a simple failing command would do too, but for some reason the aspect of repeated invocation is tested(?) Is that needed?

kyleam:
>  `echo x$(cat grows) > grows && false`  it looks as if a simple failing command would do too, but for some reason the aspect of repeated invocation is tested(?) Is that needed?

It could be something else, provided that it produces an output file so that there's something to be saved and reran.

mih:
@kyleam Is this still an issue? Should not, right?

kyleam:
The two specific issues mentioned should be resolved and, as of gh-3251, tested.  I'm sure there are still Windows-compatibility issues for `run` (there definitely are for `rerun`), but with gh-3251 -core should match -revolution's compatibility.

PR #3031, opened 2018-11-29
RF: Do not force byte-encoding of commit message

mih:
This fails on windows, see #3030

This PR may be accumulating more changes.

### DIFF
--- a/datalad/interface/run.py
+++ b/datalad/interface/run.py
@@ -713,14 +713,13 @@ def run_command(cmd, dataset=None, inputs=None, outputs=None, expand=None,
     msg = msg.format(
         message if message is not None else _format_cmd_shorty(cmd),
         '"{}"'.format(record_id) if use_sidecar else record)
-    msg = assure_bytes(msg)
 
     outputs_to_save = outputs.expand(full=True) if explicit else '.'
     if not rerun_info and cmd_exitcode:
         if outputs_to_save:
             msg_path = relpath(opj(ds.repo.path, ds.repo.get_git_dir(ds.repo),
                                    "COMMIT_EDITMSG"))
-            with open(msg_path, "wb") as ofh:
+            with open(msg_path, "w") as ofh:
                 ofh.write(msg)
             lgr.info("The command had a non-zero exit code. "
                      "If this is expected, you can save the changes with "


kyleam:
> RF: Do not force byte-encoding of commit message

Note that we've done so in one form or another since 163cf27f1 (DOC+ENH: Improve docs, ensure sorted command properties, 2017-08-21).  I wonder if you added it there in response to any particular issue.

datalad/interface/run.py
```diff
@@ -713,14 +713,13 @@ def run_command(cmd, dataset=None, inputs=None, outputs=None, expand=None,
     msg = msg.format(
         message if message is not None else _format_cmd_shorty(cmd),
         '"{}"'.format(record_id) if use_sidecar else record)
-    msg = assure_bytes(msg)
```
yarikoptic: what about introducing a dedicated test (or enabling some existing one in this code base which currently would fail due to this bug?) for appveyor to avoid a wild goose chase?

mih:
> > RF: Do not force byte-encoding of commit message
> 
> Note that we've done so in one form or another since [163cf27](https://github.com/datalad/datalad/commit/163cf27f15317b6f2f27f15d811702f402608b5a) (DOC+ENH: Improve docs, ensure sorted command properties, 2017-08-21). I wonder if you added it there in response to any particular issue.

Yeah, could not recall any reason, and found no comment -- hence the PR. There is likely more...

datalad/interface/run.py
mih: Not sure what goose is being changed. I don't see what test could be written to make `run` work on windows without investing in a range of windows bug fixes first.

Let me know if you oppose such changes and I'll tackle this by an RF that allows me to replace this from outside.

yarikoptic: Please don't blackmail, I do not oppose the change, I just want to avoid a blind fix. A possible RF to do is to take this saving functionality and move into a dedicated method to be called here, and tested independently without requiring the whole run to work properly

mih: Did you see the failure log lonked in #3030? It fails due to a mixture of strong and byte args in the command invocation. I don't see how I could test that, or what the advantage would be if testing that the command would then fail for other reasons.

kyleam:
> Ok, tests fail, but those failures seem unrelated. I conclude that none of the present test check any behavior that may have caused this forced encoding to be done.

To further support this: There is one test that uses `OBSCURE_FILENAME` (yes, there should be more).  The obscure name is used for an input, but the inputs aren't expanded into the commit message.  Locally, I've changed it so that it expands:

```diff
diff --git a/datalad/interface/tests/test_run.py b/datalad/interface/tests/test_run.py
index d7110b525..bda046b21 100644
--- a/datalad/interface/tests/test_run.py
+++ b/datalad/interface/tests/test_run.py
@@ -951,7 +951,7 @@ def test_inputs_quotes_needed(path):
     # spaces ...
     cmd_str = "{} -c \"{}\" {{inputs}} {{outputs[0]}}".format(
         sys.executable, cmd)
-    ds.run(cmd_str, inputs=["*.t*"], outputs=["out0"])
+    ds.run(cmd_str, inputs=["*.t*"], outputs=["out0"], expand="inputs")
     expected = u"!".join(
         list(sorted([OBSCURE_FILENAME + u".t", "bar.txt", "foo blah.txt"])) +
         ["out0"])
```

On top of your changes, the run still passes and the log message looks fine.

Issue #2976, opened 2018-11-05
test_datasets_datalad_org fails on appveyor


yarikoptic:
#### What is the problem?

Since https://github.com/datalad/datalad/pull/2941 which introduced this test we got the failure, although it did pass then (https://ci.appveyor.com/project/mih/datalad/builds/19892298).  Seems to be related to introduction of v7 (started to fail [here](https://ci.appveyor.com/project/mih/datalad/builds/19963437/job/gpudc71cnjh44yye) with annex 7. and preceding green [run](https://ci.appveyor.com/project/mih/datalad/builds/19920699/job/87fwsw40qs2v0gkm) had 6) but https://github.com/datalad/datalad/pull/2975 unfortunately still exhibits it in its current shape

<details>
<summary>extract from appveyor</summary>

```
======================================================================
FAIL: datalad.distribution.tests.test_install.test_datasets_datalad_org('',)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Miniconda35\envs\test-environment\lib\site-packages\nose\case.py", line 198, in runTest
    self.test(*self.arg)
  File "C:\Miniconda35\envs\test-environment\lib\site-packages\datalad\tests\utils.py", line 818, in newfunc
    return func(*args, **kwargs)
  File "C:\Miniconda35\envs\test-environment\lib\site-packages\datalad\tests\utils.py", line 592, in newfunc
    return t(*(arg + (filename,)), **kw)
  File "C:\Miniconda35\envs\test-environment\lib\site-packages\datalad\distribution\tests\test_install.py", line 931, in check_datasets_datalad_org
    status='ok')
  File "C:\Miniconda35\envs\test-environment\lib\site-packages\datalad\tests\utils.py", line 1208, in assert_result_count
    dumps(results, indent=1, default=lambda x: str(x))))
AssertionError: Got 0 instead of 1 expected results matching {'status': 'ok'}. Inspected 1 record(s):
[
 {
  "status": "notneeded",
  "type": "file",
  "path": "C:\\Users\\appveyor\\AppData\\Local\\Temp\\1\\datalad_temp_4udmjxa7\\001-anat-scout_ses-{date}\\000001.dcm",
  "action": "get",
  "refds": "C:\\Users\\appveyor\\AppData\\Local\\Temp\\1\\datalad_temp_4udmjxa7",
  "message": "already present"
 }
]
======================================================================
[...]

PR #3036, opened 2018-11-30
WiP: annihilation of direct mode support for the sake of the revolution

yarikoptic:
A can remove so good

Removed:
- `direct=` option from AnnexRepo interface
- `AnnexRepo._submodules_dirty_direct_mode` - not even sure if was used
- `AnnexRepo._git_custom_command` and `AnnexRepo.proxy` - no longer needed

RFed:
- `set_direct_mode` -> `_set_direct_mode` which just calls `git annex direct`.  To be used internally for testing

Kept for now (but might go as well):
- `is_direct_mode()`

Behavior:
- encountering `DATALAD_REPO_DIRECT=` env var (well -- datalad config) set to True causes `DirectModeNoLongerSupportedError` (but happens only after possibly repo got created)
- encountering AnnexRepo in direct mode also causes `DirectModeNoLongerSupportedError` with a different message for reason .
- Both `DirectModeNoLongerSupportedError` provide instruction to upgrade corresponding repo using direct git-annex upgrade -- we might want to provide auto-upgrade convenience or not?

- [x] adjust AnnexRepo class to remove any custom direct mode handling (WiP) while announcing absent direct mode handling
- [x] update minimal annex requirement to `7` for more uniform handling and remove other related craft
- [x] reintroduce `@known_failure_direct_mode` as a shim which just merely issues a deprecation warning when executed. This way we could keep extensions compatible with new datalad core and allow them to deprecate usage of this decorator, without requiring them all to release a new version.  That should prevent [breakage of travis](https://travis-ci.org/datalad/datalad/jobs/462619117) while testing datalad-crawler
- [x] make sure tests pass

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,6 @@ env:
     - COVERAGE=coverage
     - DATALAD_DATASETS_TOPURL=http://datasets-tests.datalad.org
   matrix:
-    - DATALAD_REPO_DIRECT=yes
 ##1    - DATALAD_REPO_VERSION=6
     - DATALAD_REPO_VERSION=5
 
@@ -159,14 +158,6 @@ matrix:
 #    - DATALAD_TESTS_KNOWNFAILURES_PROBE=yes
 #    - DATALAD_TESTS_SSH=1
 #    - _DL_CRON=1
-  - python: 3.5
-    # test whether known direct mode failures still fail
-    env:
-    - DATALAD_REPO_DIRECT=yes
-    - DATALAD_TESTS_KNOWNFAILURES_SKIP=no
-    - DATALAD_TESTS_KNOWNFAILURES_PROBE=yes
-    - DATALAD_TESTS_SSH=1
-    - _DL_CRON=1
 
   - python: 3.5
     # test with extension datalad-crawler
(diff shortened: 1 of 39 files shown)


yarikoptic:
FTR, failures on windows are probably due to removed `@skip_direct_mode` or `@known_failure_direct_mode` which is no longer direct mode I guess... some look just like path transformation issues etc.  But I guess for consistency/irrelevance to this PR in particular I will just skip them on windows for now
```
$> grep FAIL: log.txt
[00:08:51] FAIL: datalad.distribution.tests.test_create.test_create_sub
[00:08:51] FAIL: datalad.distribution.tests.test_update.test_newthings_coming_down
[00:08:51] FAIL: datalad.distribution.tests.test_update.test_reobtain_data
[00:08:51] FAIL: datalad.distribution.tests.test_update.test_update_volatile_subds
[00:08:51] FAIL: datalad.distribution.tests.test_install.test_install_consistent_state
[00:08:51] FAIL: datalad.distribution.tests.test_install.test_failed_install_multiple
```

kyleam:
> but we can now drop one of the appveyor runs, right? Before annex 7.20181031, it would have been the direct mode run, but now it is just a repeat v7 run, I think.

Answering my own question: Although the first appveyor run should be v7, it's not a straight repeat of the second.  Even with the current tip of this PR, there is some yet-to-be-figured-out difference leading to #2976.

yarikoptic:
thanks for pushing one more fix!  and yeap -- I thought that may be that test is for some reason skipped on the "good run" but nope:
```
datalad.distribution.tests.test_install.test_datasets_datalad_org('',) ... ok
datalad.distribution.tests.test_install.test_datasets_datalad_org('/.git',) ... ok
```

yarikoptic:
was autoclosed when I removed develop to make it rf-revolution1 as prommised

yarikoptic:
according to https://codecov.io/gh/datalad/datalad/pull/3036# drop in coverage just 0.07% and codecov web ui fails to work for me ATM to investigate :-(  so I will just claim success and merge into the target rf-revolution1 branch

PR #3047, opened 2018-12-03
TEMP: troubleshoot problematic appveyor run

yarikoptic:
Ref #2976


yarikoptic:
the observations without explicit DATALAD_REPO_VERSION=6, and without datalad altogether -- cloned repor becomes version 5 (according to .git/config) repo in `adjusted/master(unlocked)` branch upon `git annex info` command causing it to "transition"

image

@joeyh -- should't it jump to v7 right away??? (at least that was my impression from what to expect from 7.XXXXXX annex)

observation #2 (irrelevant to this troubleshooting issue I hope):  even pure git is not satisfied with this filesystem:
image

yarikoptic:
the reason was figured out, no need to keep open

PR #3050, opened 2018-12-03
ENH(TST): add a test run on a crippled FS to disentangle from appveyor Windows specific fails

yarikoptic:
We might be too far from being ready for it, then I would either close it or move it to allowed failures on CRON

But it allows me to replicaate locally https://github.com/datalad/datalad/issues/2976 so I could test fresh git annex now

- [ ] complete initial run and see how many/which fail
- [ ] enable TEMPorarily fresh git-annex build (available now from debian-devel) and see how things look then
- [ ] decide how to proceed with adopting or not this ENH(TST)


yarikoptic:
with new one (7.20181121+git58-gbc4aa3f0e-1~ndall+1 ) slightly better [(FAILED (SKIP=54, errors=56, failures=42))](https://travis-ci.org/datalad/datalad/jobs/463147444#L5668)

yarikoptic:
well, it was a total disaster, good to have for reference, but not worth keeping it afloat

PR #3061, opened 2018-12-05
ENH: Add the ability for custom result summary rendering

mih:
Now each class could provide its own idea of how results can be summarized, and override the default action counter that is sometimes rather uninformative.

Instrumental in addressing situations like https://github.com/datalad/datalad-revolution/issues/56

Let's see if it breaks anything existing. Demo PR: https://github.com/datalad/datalad-revolution/pull/57

### DIFF
--- a/datalad/interface/utils.py
+++ b/datalad/interface/utils.py
@@ -414,6 +414,12 @@ def eval_results(func):
                             **_kwargs):
                         yield r
 
+            # if a custom summary is to be provided, collect the results
+            # of the command execution
+            results = []
+            do_custom_result_summary = result_renderer == 'tailored' \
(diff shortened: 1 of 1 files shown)


datalad/interface/utils.py
```diff
@@ -414,13 +414,22 @@ def generator_func(*_args, **_kwargs):
                             **_kwargs):
                         yield r
 
+            # if a custom summary is to be provided, collect the results
+            # of the command execution
+            results = []
+            do_custom_result_summary = result_renderer == 'tailored' \
+                and hasattr(_func_class, 'custom_result_summary_renderer')
```
yarikoptic: if checking for `tailored` why not `tailored_results_summary_renderer`?

PR #3062, opened 2018-12-05
RF: [DATALAD RUNCMD] -> [DATALAD]

yarikoptic:
Because
- it takes precious space of the summary line for no immediate need
- none other datalad command (save, add, etc) embeds the command into
  commit summary message.  IMHO no need for run to be special and
  just makes parsing of the log more difficult in oneline mode
  due to misalignment of messages

This is largely to sense for consensus one way or another

TODOs (if decided to proceed forward with it)
- [ ] a unittest that we still can detect/parse previously recorded run records


kyleam:
> - it takes precious space of the summary line for no immediate need

It certainly takes up more space.  I suppose whether it's useful depends
on the user and the context.  I've found the "RUNCMD" (or rather the
distinct prefix) helpful for visually distinguishing run commits while
scanning a log.  That's from the standpoint of debugging and developing
run/rerun, and development convenience shouldn't be a reason for
retaining it.  But it does make me wonder if some users rely on the
prefix to do the same sort of scanning.

In terms of distinguishing run commits, I also like that I can do `git
--grep=RUNCMD ...`.  I can get similar information by grepping "Do not
change lines below" or `rerun --report ...`, but it is nice to have a
straightforward, easy-to-remember fragment to use with `git log --grep`.

> - none other datalad command (save, add, etc) embeds the command into
>   commit summary message.

True.

> IMHO no need for run to be special

The one way I view run as special is that it's about a user-specified,
arbitrary command, so I could see a point in distinguishing it from more
restricted DATALAD operations.

But that same point could be used to argue that the prefix shouldn't
take up so much space available for the user's message (or the command,
if none is supplied).  Of course, distinguishing run commits from other
datalad commits doesn't have to take up more space.  We could use a
prefix like "[RUN]" or "[DL RUN]".

>   just makes parsing of the log more difficult in oneline mode
>   due to misalignment of messages

Along the lines of what I mentioned above, I find that useful for
distinguishing run commits.  (Interestingly, misalignment doesn't stop
you from preferring the "BF:" and "ENH:" prefixes that I dislike for
non-misalignment reasons :)

Those are counter-arguments against using "[DATALAD]" from my
perspective, but I'm not strongly opposed to dropping the " RUNCMD" bit
and am sympathetic to the argument that "[DATALAD RUNCMD]" takes up too
much space.

bpoldrack:
As far as the summary line is concerned I agree. But it's not entirely useless, as you can see in rerun (where you changed the pattern to look for). Now we would consider everything a run record that has those "Do not change lines"-remark.
But those lines are only pointing to a run record by chance. Somewhere in there (may be just next line) should be an explicit mark, so that way of recording things can be reused with a changed mark by extensions (or sth core comes up with in the future) without breaking the detection by `rerun`.

kyleam:
@bpoldrack:
> But those lines are only pointing to a run record by chance.

I do agree with this as well.  We shouldn't rely on the "Do not change lines" block being run-specific.

mih:
Going one step back: Why is there a '[DATALAD]' in the first place? I think the idea was to indicate "automatic" commits with (in general) poor commit messages. The fact that DataLad did them is not as important as that no thought went into the commit message. The situation is different for run records. The summary line contains a shortened command, and the RUNCMD is the important piece for decoding the commit message. So if anything is to be kept, it should be RUNCMD (or a similar indicator). I would echo the importance of enabling *humans* to scan the git log.

Maybe just `[CMD]` or `[PROV]`?

kyleam:
> Maybe just `[CMD]` or `[PROV]`?

I'd prefer `[CMD]` or, along the lines of my suggestion above, `[RUN]`.

And I think dropping "DATALAD" and using a short prefix is a good option: frees up summary line space but also keeps it distinct.

bpoldrack:
> I think the idea was to indicate "automatic" commits with (in general) poor commit messages. The fact that DataLad did them is not as important as that no thought went into the commit message.

Agree.That's been the idea. And while I'm not strongly opposing to shorten it to `[DL]` or sth, I don't really see a point. `[DATALAD]` is very easy (and quick!) to parse for humans and therefore to step over or to focus on in the log. And since it is about being somewhat "stupid" with the actual message I don't think there is a need to free up space. If datalad itself has to say a lot about the commit, it's likely to be about a list of details that should go into the body. A really meaningful summary that would need the space is likely to come from a user anyway and therefore doesn't need the prefix at all. If a datalad commit runs out of space for the summary line, I'd guess that summary line isn't "good". So I tend to prefer `[DATALAD]` and force ourselves to make better messages.

>> Maybe just [CMD] or [PROV]?

> I'd prefer [CMD] or, along the lines of my suggestion above, [RUN].

I'd go with `[RUN]`. Simply because it refers to what the marker really is about: It's an "encoded" message by `run`. If at some point we want to have additional and (slightly) different records by `run-procedure` or something else yet to be invented, that also runs some flavor of a "command", `[CMD]` could easily end up being a disturbance of some kind. And I think the same goes for `[PROV]`.

mih:
Closing due to inactivity. #3370 is in place to track the issue.

PR #3098, opened 2018-12-28
ENH+BF(TST): create-sibling --group option

yarikoptic:
1.
    ENH: create-sibling --group option
    
    When needing to create a group shared remote sibling, we need to be able
    to specify which group actually should it be (if directories setup does not
    enforce it).

2.
    BF+RF: @skip_ decorators must not be used on generator tests
    
    Apparently some tests were simply not ran because decorated to be skipped,
    but our decorators do not work nicely with those generatorfunctions tests,
    so we must use them via their functional counterparts.
    
    Ideally those @skip_ helpers should be refactored since there is a repeating
    pattern and only the check is different.  But I kept it "minimal" for now,
    and made them usable as functions within tests in such cases.
    
    There might now be test failures detected where they are still not used
    appropriately

### DIFF
--- a/datalad/distribution/create_sibling.py
+++ b/datalad/distribution/create_sibling.py
@@ -204,10 +204,10 @@ def _create_dataset_sibling(
             delayed_super, name, ssh)
 
     if group:
-        # Either it existed before or a new directory for a repo, set its
-        # group to a desired one if was provided
+        # Either repository existed before or a new directory was created for it,
+        # set its group to a desired one if was provided with the same chgrp
         ssh("chgrp -R {} {}".format(
-            group,
+            sh_quote(group),
             sh_quote(remoteds_path)))
     # don't (re-)initialize dataset if existing == reconfigure
     if not only_reconfigure:
--- a/datalad/distribution/create_sibling.py
+++ b/datalad/distribution/create_sibling.py
@@ -389,8 +389,9 @@ class CreateSibling(Interface):
         group=Parameter(
             args=("--group",),
             metavar="GROUP",
-            doc="""Filesystem group for the repository. Important in particular
-            when --shared=group""",
+            doc="""Filesystem group for the repository. Specifying the group is
+            particularly important when [CMD: --shared=group CMD][PY:
+            shared="group" PY]""",
             constraints=EnsureStr() | EnsureNone()
         ),
         ui=Parameter(
--- a/datalad/distribution/create_sibling.py
+++ b/datalad/distribution/create_sibling.py
@@ -207,7 +207,7 @@ def _create_dataset_sibling(
         # Either repository existed before or a new directory was created for it,
         # set its group to a desired one if was provided with the same chgrp
         ssh("chgrp -R {} {}".format(
-            sh_quote(group),
+            sh_quote(text_type(group)),
             sh_quote(remoteds_path)))
     # don't (re-)initialize dataset if existing == reconfigure
     if not only_reconfigure:
--- a/datalad/distribution/create_sibling.py
+++ b/datalad/distribution/create_sibling.py
@@ -70,6 +70,7 @@ def _create_dataset_sibling(
         target_pushurl,
         existing,
         shared,
+        group,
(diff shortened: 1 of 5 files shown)


yarikoptic:
this test (was not run for awhile) fails from travis but not locally
```
ERROR: datalad.downloaders.tests.test_http.test_download_ftp('ftp://ftp.gnu.org/README', None, 'This is ftp.gnu.org')
    self.test(*self.arg)
  File "/home/travis/build/datalad/datalad/datalad/tests/utils.py", line 615, in newfunc
  File "/home/travis/build/datalad/datalad/datalad/downloaders/tests/test_http.py", line 251, in check_download_external_url
    downloaded_path_ = downloader.download(url, path=d, size=s, overwrite=True)
  File "/home/travis/build/datalad/datalad/datalad/downloaders/base.py", line 468, in download
    return self.access(self._download, url, path=path, **kwargs)
  File "/home/travis/build/datalad/datalad/datalad/downloaders/base.py", line 145, in access
    result = method(url, **kwargs)
  File "/home/travis/build/datalad/datalad/datalad/downloaders/base.py", line 369, in _download
    downloader_session = self.get_downloader_session(url)
  File "/home/travis/build/datalad/datalad/datalad/downloaders/http.py", line 508, in get_downloader_session
    check_response_status(response, session=self._session)
  File "/home/travis/build/datalad/datalad/datalad/downloaders/http.py", line 112, in check_response_status
    raise AccessFailedError(err_msg)
AccessFailedError: Access to ftp://ftp.gnu.org/README has failed: status code 503
```
anyone knows some other good ftp resource we could test with?

datalad/distribution/create_sibling.py
```diff
@@ -202,6 +203,12 @@ def _create_dataset_sibling(
         shared = CreateSibling._get_ds_remote_shared_setting(
             delayed_super, name, ssh)
 
+    if group:
+        # Either it existed before or a new directory for a repo, set its
+        # group to a desired one if was provided
+        ssh("chgrp -R {} {}".format(
+            group,
```
kyleam: While not advisable or something we're likely to encounter, it seems like group can contain characters that would need to be shell escaped (that was based on quickly googling, so I may be wrong here).  Should we use `sh_quote` here to?

datalad/distribution/create_sibling.py
```diff
@@ -202,6 +203,12 @@ def _create_dataset_sibling(
         shared = CreateSibling._get_ds_remote_shared_setting(
             delayed_super, name, ssh)
 
+    if group:
+        # Either it existed before or a new directory for a repo, set its
+        # group to a desired one if was provided
```
kyleam: I'm having trouble parsing this.

yarikoptic: I just wanted to say that the same command (chgrp) will be used regardless either it (directory/repository) already existed or we just created a new directory for it

datalad/distribution/create_sibling.py
yarikoptic: I guess it shouldn't hurt, will add now

datalad/distribution/create_sibling.py
yarikoptic: will try to make it more english since anyways need to push

datalad/downloaders/tests/test_http.py
```diff
@@ -311,10 +311,16 @@ def test_download_ftp():
         import requests_ftp
     except ImportError:
         raise SkipTest("need requests_ftp")  # TODO - make it not ad-hoc
-    yield check_download_external_url, \
-          "ftp://ftp.gnu.org/README", \
-          None, \
-          "This is ftp.gnu.org"
+    # Started to throw 504 when on travis
+    try:
+        yield check_download_external_url, \
```
kyleam: I'd guess that the tests are still failing because you're trying to catch the exception around the `yield` rather than the actual invocation.  Given that nothing else is yielded in this test, can you call `check_download_external_url` directly instead?

PR #3101, opened 2019-01-04
[WIP] ENH: Allow Git to talk to the caller in Repo.commit()

mih:
- [x] let stdout through. Without this change, DataLad blocks any attempt to get Git to launch an editor to tailor commit messages.
- [x] RF handling of commit messages too (blocks more `git commit` features like message templates)
- [ ] catch error when `git commit` fails due to an unedited commit message (template) and issue a proper error instead of crashing with debug output

It turns out that the last point, and, even worse, acting on any error is tricky (or impossible) -- see below and #3103

### DIFF
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -52,6 +52,7 @@ from datalad.cmd import GitRunner
 from datalad.consts import GIT_SSH_COMMAND
 from datalad.dochelpers import exc_str
 from datalad.config import ConfigManager
+from datalad.ui import ui
 from datalad.utils import assure_list
 from datalad.utils import optional_args
 from datalad.utils import on_windows
@@ -1265,16 +1266,39 @@ class GitRepo(RepoInterface):
         self.precommit()
 
         if _datalad_msg:
+            # will not just prefix, but also return just the prefix
+            # as the sole message -- may or may not be a good thing
             msg = self._get_prefixed_commit_msg(msg)
 
         options = options or []
 
-        if not msg:
-            if options:
-                if "--allow-empty-message" not in options:
-                        options.append("--allow-empty-message")
-            else:
-                options = ["--allow-empty-message"]
+        # if _datalad_msg was set, there is always a message at this
+        # point, so internal calls setting this flag, because they have no
+        # better clue what is happening are safe now
+        # for any other call, not supplying a message can be meaningful
+        # depending on the options given to `commit`, e.g.
+        # --amend will re-use a message, only if none if supplied
+        # --template will insert a template, only if no message is given
+        # of course not supplying a message could also be a mistake
+        # look for --no-edit as an explicit flag to let empty messages
+        # through
+        # if not message was supplied, but the session is non-interactive
+        # there is no way to acquire a message, hence also enable
+        # pass-through, as we don't want to fail for procedural issues
+        # that deep in the stack -- better fix this stuff in the caller
+        if (not ui.is_interactive or '--no-edit' in options) \
+                and "--allow-empty-message" not in options:
+            allow_interactive = False
+            options.append("--allow-empty-message")
+        else:
+            allow_interactive = True
+
+        # likewise, if edit was specifically requested, do not force-pass
+        # an empty message
+        if msg:
+            options.extend(('-m', msg))
+        elif '--edit' not in options:
+            options.extend(('-m', ''))
 
         if date:
             options += ["--date", date]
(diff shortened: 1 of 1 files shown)


datalad/utils.py
```diff
@@ -1539,7 +1539,7 @@ def make_tempfile(content=None, wrapped=None, **tkwargs):
 
     filename = {False: tempfile.mktemp,
                 True: tempfile.mkdtemp}[mkdir](**tkwargs_)
-    filename = realpath(filename)
+    #filename = realpath(filename)
```
yarikoptic: might be worth checking why it was added - there was a reason I guess for realpath, if not -- remove and leave a comment

datalad/support/gitrepo.py
```diff
@@ -1296,14 +1320,16 @@ def commit(self, msg=None, options=None, _datalad_msg=False, careless=True,
...
 
         try:
             self._git_custom_command(files, cmd,
-                                     expect_stderr=True, expect_fail=True,
+                                     log_stdout=not allow_interactive,
+                                     expect_stderr=True,
+                                     expect_fail=True,
                                      check_fake_dates=True,
                                      index_file=index_file)
         except CommandError as e:
```
yarikoptic:
```
add(ok): /tmp/datalad_demo_bob.0wlc/myanalysis/code (directory)
[ERROR  ] argument of type 'NoneType' is not iterable [gitrepo.py:commit:1336] (TypeError) 
I: done running docs/examples/3rdparty_analysis_workflow.sh: failed
The command "if [ ${DATALAD_TESTS_SSH:-0} != 1 -o -n "${BUILD_DATALAD_EXTENSION:-}" ] || echo "${TMPDIR:-}" | grep -q ' '; then echo "Testing examples is disabled by configuration"; else $NOSE_WRAPPER ../tools/testing/run_doc_examples; fi" exited with 1.
```
With this change stdout could be passed through and would be `None` here.

datalad/utils.py
mih: I will take this out (was committed unintentionally). I carry this change locally, as it makes it possible for the test battery to discover symlink issue. But I have no juice to figure them all out in -core (all that code is replaced in -revolution.

datalad/support/gitrepo.py
mih: Not sure what you mean. Will research it in the next iteration.

mih:
> Main concerns

Oh la la, I hope the (future?) minor concerns will be manageable too ;-)

> as the test failure shows needs a test for such scenario.

Everything needs a test. Almost all test failures, however, are already long present (and real) issues of symlink handling failure in -core.

> I think we need then (here or separate) provide support for config (default to auto) to override logic in is_interactive, so ppl on CIs etc could force mode to be non-interactive

This is a worthwhile mechanism to have, regardless of this change. As commented elsewhere, "non-interactive" (as in "terminal attached") is actually not a good condition to test for. When invoked in a GUI environment, the editor would not need to show up in a terminal, and not have to communicate through `stdout`. However, we don't have this kind of concept in DataLad, yet (AFAIK), and I don't plan to introduce it in this PR.

For a CI context this change should not be an issue (AFAICS right now).

I will add "WIP" to the subject to indicate that this PR does not need to be looked at yet (or is the issue tag the preferred method?).

datalad/utils.py
mih: Force-pushed removal.

datalad/support/gitrepo.py
yarikoptic: if log_stdout is False, then e.stdout is None, and thus this crash

That is the kind of "CI failures" I was talking about where previous we might just commit with empty commit message when after this particular change we would end up either crashing like here or hanging asking "CI users" to enter commit message

mih: Hmm, I think the fact that it crashes is exactly the desired behavior, because the code (and myself), are unaware of these inter-dependencies. The changes aim to not affect CI behavior in any way.

mih: Looked into it, the whole thing seems flawed. If I let stdout through to the caller, error messages are not properly extracted, and things like `Repo.commit(careless=True)` are not functional (and that is a critical issue for us). So indeed, and as feared, this is not a CI issue, but a more global problem. I could not yet determine, whether a particular combination of Runner options would make it possible to act on errors and have stdout be connected to the calling terminal. The numerical error codes seem unspecific (e.g. 'nothing to commit' has `1`, like all the others).

If there is no solution to this, we can scrap the whole thing and stay with `[DATALAD] Recorded changes` :(

mih:
Closing due to inactivity. Before revisiting this, `Runner` needs to see a substantial RF (possibly into more targeted subclasses).

PR #3104, opened 2019-01-05
Bunch of minor cleanup's

mih:
Goes along with https://github.com/datalad/datalad-crawler/pull/30

This is done, but the crawler PR resulted in a complete nervous breakdown of its tests. Looks unrelated, but hard to tell.

### DIFF
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -702,13 +702,3 @@ def link_file_load(src, dst, dry_run=False):
         shutil.copystat(src_realpath, dst)
     else:
         lgr.log(2, "Hardlinking finished")
-
-
-def get_runner(*args, **kwargs):
-    # needs local import, because the ConfigManager itself needs the runner
-    from . import cfg
-    # TODO:  this is all crawl specific -- should be moved away
-    if cfg.obtain('datalad.crawl.dryrun', default=False):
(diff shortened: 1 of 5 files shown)


mih:
OK, moving ahead with this one.

datalad/cmd.py
```diff
@@ -674,41 +683,3 @@ def run(self, cmd, env=None, *args, **kwargs):
...
-        os.link(src_realpath, dst)
-    except AttributeError as e:
-        lgr.warn("Linking of %s failed (%s), copying file" % (src, e))
-        shutil.copyfile(src_realpath, dst)
-        shutil.copystat(src_realpath, dst)
-    else:
-        lgr.log(2, "Hardlinking finished")
-
-
-def get_runner(*args, **kwargs):
```
yarikoptic: FWIW, pruning/moving API which is used by extensions without lengthy deprecation cycle brings pain to the developers and users. Primary reason - absent ability to declare Breaks for pypi and conda. So now, when we release datalad, we must release crawler, but even then who had older crawler extension installed, and updates only datalad would be screwed... Could be even at the level of just incoming datalad commands. I truly "enjoyed" similar situation with keyring and keyrings.alt doing similar act.

Issue #2926, opened 2018-10-16
Bring revolution home?


yarikoptic:
#### What is the problem?
What is your plan @mih about the https://github.com/datalad/datalad-revolution ?  To keep it as a separate extension for eternity or bring pieces (or entirety) into datalad codebase for perspective future possibly API breaking release?  
As I expressed myself, reviewing those changes is not easy and thus largely is not done. As I stated before, IMHO having them even in `master` branch (aiming e.g. for 0.11 release), while possibly still carrying out bugfixing in 0.10.x branch would be IMHO more preferable long term.

edits to outline steps to see this issue resolved:
- [x] establish `rf-revolution1` branch to contain initial changes
- [x] remove direct mode (#3036) - merge into `rf-revolution1`. WiP
    - [ ] go through leftover '# TODO: RM DIRECT' to see if anything else could/should be removed/refactored
    - [ ] add dedicated tests for handling situations when a repository in direct mode is encountered (directly by AnnexRepo or as a submodule) and assure a consistent/informative blow up
- Workflow for changes propagation
   - [ ] (`rf-revolution1-3rd` branch) bring in revolution code base into the core under 3rd/datalad_revolution (TODO: tools/injest_revolution_code to be reused to suck in the changes)
   - [ ] (`rf-revolution1` branch, to keep merging `rf-revolution1-3rd`) `git mv 3rd/datalad_revolution/datalad_revolution datalad/revolution` (verify that tests pass, needs relative imports)
   - [ ] move out pieces from `datalad/revolution/`  around the core, where needed prefixing with `rev_`
        - [ ] move our `*repo.py` to `core_*repo.py`, while having revolution ones replace ours, while importing the `core_` ones
- [ ] establish rel-0.11 branch to absorb relevant fixes (if possible - submit fixes against that branch when applicable)
- [ ] finalize #3108 and merge into rf-revolution1
- [ ] merge `rf-revolution1` into `master`
- [ ] extend core *Repo classes with the API from revolution
- [ ] investigate mechanism to by default disable "experimental" revolution API within core datalad (config setting `datalad.experimental.revolution` to be enabled by extensions before importing API etc?)
- [ ] **compare intentions/usecases/convergence for the API [google doc API Use-cases convergence: 1st Revolution](https://docs.google.com/spreadsheets/d/1bi3AQK5yMKwTA_6FN1oCPnIGASy9VmN8LbARr15XMKE/edit?usp=sharing)**
- [ ] ...
[...]

PR #3106, opened 2019-01-06
BF: GitRepo.commit to work correctly on directory paths (Fixes #3087)

### LINKED ISSUE #3087: save fails when passed a directory
`save` doesn't work when given a directory as of 5ae188c89 (BF: possibly make use of temporary index every time (not only direct mode) committing a list of files, 2018-11-23, gh-3009).

```
% mkdir sdir && echo foo >sdir/foo

% datalad save sdir
CommandError: command '['git', '-c', 'receive.autogc=0', '-c', 'gc.auto=0', 'commit', '-m', '[DATALAD] Recorded existing changes']' failed with exitcode 1
Failed to run ['git', '-c', 'receive.autogc=0', '-c', 'gc.auto=0', 'commit', '-m', '[DATALAD] Recorded existing changes'] under '/tmp/tmp.DatFpoHPdN-dl'. Exit code=1. out=On branch master
Untracked files:
        sdir/

nothing added to commit but untracked files present
 err=
On branch master
Untracked files:
        sdir/

nothing added to commit but untracked files present

% git status
On branch master
Changes to be committed:
        new file:   sdir/foo
```

<details>

<summary>bisect script</summary>

```sh
#!/bin/sh

set -eux

tmpdir=$(mktemp -d --suffix=-savefail)
datalad create $tmpdir
cd $tmpdir

mkdir -p s
echo foo >s/a
datalad save s

test -z "$(git diff-index --cached HEAD -- s/a)" || exit 1
```

</summary>

yarikoptic:
This pull request fixes #3087 through comparing to the `git diff` for the files provided to the commit call, so should behave properly in case of directories etc.

### Changes
- [x] Needs fixing since breaks the tests
- Make checks optional for the `rev-` functionality since it cares not about staged or not and about direct mode, so I guess a plain `commit(files)` should suffice in that case - probably should better be done in master upon merge

### DIFF
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -11,7 +11,7 @@
 For further information on GitPython see http://gitpython.readthedocs.org/
 
 """
-
+from itertools import chain
 import logging
 import re
 import shlex
@@ -61,8 +61,7 @@ from datalad.utils import getpwd
 from datalad.utils import updated
 from datalad.utils import posix_relpath
 from datalad.utils import assure_dir
-from datalad.utils import CMD_MAX_ARG
-from datalad.utils import generate_chunks
+from datalad.utils import generate_file_chunks
 from ..utils import assure_unicode
 
 # imports from same module:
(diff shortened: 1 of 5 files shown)


mih:
Please make the new behavior optional and (ideally) off by default. The -revolution code stack performs this kind of normalization very early on (it is needed in other places too). Doing it unconditionally this late in the game would have no benefit and slow things down in its context.

If this is not desirable, I can overwrite `commit()` in `RevolutionGitRepo`, but I wanted to mention it.

It might be worth investigating whether the `ls-files` call done here should be modified to honor `.gitignore` setttings too. You could have a look at the specific call done in `RevolutionGitRepo.get_content_info()` -- it took a few cycles to arrive at this particular setup.

yarikoptic:
Thanks for the info/pointers!
Unless someone gets time before then, I really hope to push on #2926 in a week or two so may be this one would even be superseded ;)

datalad/support/gitrepo.py
```diff
@@ -2624,8 +2609,23 @@ def get_changed_files(self, staged=False, diff_filter='', index_file=None):
             opts.append('--diff-filter=%s' % diff_filter)
         if index_file:
             kwargs['env'] = {'GIT_INDEX_FILE': index_file}
-        return [normpath(f)  # Call normpath to convert separators on Windows.
-                for f in self.repo.git.diff(*opts, **kwargs).split('\0') if f]
+        if files:
```
yarikoptic: surprised that test didn't fail (locally) -- I intended to make it `if files is not None` here and I thought I added explicit test (may be died in intermediate RF), TODO

yarikoptic:
ok, @kyleam - should be ready for review.
I've done some limited benchmarking -- no notable change in comparison to 0.11.x ;)

datalad/support/gitrepo.py
yarikoptic: done now in db0a844c691d6848ca9627e567e658e9f3722594

PR #3108, opened 2019-01-07
Update rf-revolution1

yarikoptic:
This pull request fixes #
- [x] Merge master (conflict)
- [ ] Bring in fresh pieces from revolution (after tests pass for merge)


yarikoptic:
seems needing minor fixup for revolution piece. Feels like a repeating pattern, so may be worth copying this test into revolution to avoid it there to start with @mih 
```
FAIL: datalad.tests.test_api.test_consistent_order_of_args(<class 'datalad.revolution.revdiff.RevDiff'>, set(['path']))
    self.test(*self.arg)
  File "/home/travis/build/datalad/datalad/datalad/tests/test_api.py", line 59, in _test_consistent_order_of_args
    eq_(set(args[:len(spec_posargs)]), spec_posargs)
AssertionError: set(['fr']) != set(['path'])
```

mih:
Arrgh, this one. TBH I always have trouble to understand why the condition the test considers desirable is actually desirable. Why is 'path' more important then the revision for a diff call? I hardly ever use a path constrain in a diff. I almost always give a revision.

I feel this test forces a congruency between Python and cmdline API that is not really helpful.

yarikoptic:
It just forces that positional arguments (i.e. `path` here, without `--` prefix to depict options) come first in the Python signature.  Here you have `path` and `--fr`. IMHO generally makes total sense so if you call python method without naming the argument you get the same as you call it without naming an option on cmdline.  Otherwise - confusion, bugs, etc would hunt you/us/users down

PR #3109, opened 2019-01-07
RF: place all extracted metadata under git-annex

yarikoptic:
We have discussed it before, and decided to place small files directly under git.
But I was thinking about it and converging on placing all metadata under annex
control, so there is a better control over availability: we might need to remove
leaked sensitive metadata.

What do you think?

Not sure yet if any unittest would be effected


mih:
I am in favor of this change, as it only changes the default.

yarikoptic:
well, if we decide to go this way I guess I will adjust all  .gitattributes across http://datasets.datalad.org

may be to assess the effect, I will create a fresh snapshot for datasets-dev. so we could see how it effects anything

yarikoptic:
done: http://datasets-dev.datalad.org now has all metadata under annex (it is apparently under dummy git backend, so no progress report and slow initial clone)
so it requires 1272 files to be fetched for searching at the files level -- only ~10 sec on fast mit eduroam.  Give it a try I guess for feasibility check

PR #3115, opened 2019-01-10
ENH: Allow to replace .gitattributes entirely

bpoldrack:
ATM `GitRepo.set_gitattributes` only can add lines to `.gitattributes`. Introducing a `mode` parameter to allow for rewriting the entire thing.

### Changes
- [x] This change is complete

### DIFF
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -2599,8 +2599,8 @@ class GitRepo(RepoInterface):
     def set_gitattributes(self, attrs, attrfile='.gitattributes', mode='a'):
         """Set gitattributes
 
-        By default appends additional lines to .gitattributes. Note, that later
-        lines in .gitattributes overrule earlier ones, which may or may not be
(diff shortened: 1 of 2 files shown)


mih:
Worth stating that just adding a line will have different effects depending on the existing content. Only the replacement of the entire set of attributes give (more-or-less) full control over what is in effect.

datalad/support/gitrepo.py
```diff
@@ -2596,9 +2596,14 @@ def get_gitattributes(self, path, index_only=False):
             attr = []
         return attributes
 
-    def set_gitattributes(self, attrs, attrfile='.gitattributes'):
+    def set_gitattributes(self, attrs, attrfile='.gitattributes', mode='a'):
         """Set gitattributes
 
+        By default appends additional lines to .gitattributes. Note, that later
```
kyleam: nit: would be better to say `attrfile`.  Otherwise looks good to me.

PR #3124, opened 2019-01-16
ENH: datalad --help   uses no man(page)

### LINKED ISSUE #3118: man datalad (invoked by datalad --help) does not list commands
#### What is the problem?

`--help-np` dynamically generates a list of commands available directly from datalad or provided by one of its extensions.  But running `datalad --help` brings up the manpage for datalad which lacks any list of commands. Moreover, we simply cannot rely on manpage in this case, since it would be pregenerated once, whenever the list of commands would change depending on the list of extensions.
I think we are doomed to disable fancy "jump to manpage" for `datalad --help`.  What do you think @datalad/developers?

yarikoptic:
This pull request fixes #3118


mih:
neuroimaging failure is unrelated.

PR #3136, opened 2019-02-01
ENH: Report actual process handle, not just PID

mih:
gives:

psutil.Process(pid=1401, name='tracker-extract', started='2018-12-18 12:22:04')

instead of

1401


yarikoptic:
ha -- cool idea.. but I guess all the tests using it should adjust:
```
$> git grep get_open_files
datalad/tests/test_utils.py:from ..utils import get_open_files
datalad/tests/test_utils.py:def test_get_open_files(p):
datalad/tests/test_utils.py:    eq_(get_open_files(p), {})
datalad/tests/test_utils.py:        eq_(get_open_files(p, log_open=40), {op.realpath(f1): os.getpid()})
datalad/tests/test_utils.py:    assert not get_open_files(subd)
datalad/tests/test_utils.py:    eq_(get_open_files(p), {op.realpath(subd): proc.pid})
datalad/tests/test_utils.py:    eq_(get_open_files(subd), {op.realpath(subd): proc.pid})
datalad/tests/test_utils.py:    assert not get_open_files(subd)
datalad/utils.py:def get_open_files(path, log_open=False):
datalad/utils.py:        files = get_open_files(path, log_open=40)
```

yarikoptic:
probably storing `proc` would have been the best, and just filter out for needed information in the tests? (although I wonder why I didn't do it - may be there was a reason...?)

mih:
Reverted to just `proc`, and fixed tests and docs.

PR #3137, opened 2019-02-01
BF: Prevent issue with dangling batch processes

mih:
It might be more useful to execute `precommit()` on a higher level, maybe at the end of aggregation...


yarikoptic:
cool, merging...
or may be more pythonic would be to provide something like `with repo.batched` context handler which would close them at the end?

PR #3153, opened 2019-02-08
TEMP: test using -devel annex from neurodebian

yarikoptic:
should be 7.20190129+git78-g3fa6be1fe-1~ndall+1 version


yarikoptic:
ha! fully green travis, enjoy!  I will upload now this flagman version of git annex to neurodebian proper

PR #3156, opened 2019-02-10
BF/RF: automatically import needed interface submodules if Dataset.datasetmethod is not bound yet

### LINKED ISSUE #3155: Delayed bindings are not good -- we should come up with systematic solution
even if it would involve some hardcoding, at least for the core datalad functionality.  Now even our users keep discovering missing imports: https://github.com/templateflow/templateflow/issues/11 showing how fragile it is

yarikoptic:
This pull request fixes #3155

### DIFF
--- a/datalad/interface/base.py
+++ b/datalad/interface/base.py
@@ -100,6 +100,22 @@ def get_interface_groups(include_plugins=False):
     return grps
 
 
+def get_interface_name(intfspec, domain='python'):
+    """Given interface spec, return the func/method name"""
+    assert domain == 'python', "no other ATM supported for this function"
+
+    if len(intfspec) >= 4:
+        name = intfspec[3]
+    else:
+        # there might be just 2 entries - module, class
+        if len(intfspec) <= 2:
+            name = intfspec[1].lower()
+        else:
+            name = intfspec[2]
+        name = name.replace('-', '_')
+    return name
+
+
 def get_cmd_summaries(descriptions, groups, width=79):
     """Return summaries for the commands in `groups`.
 
--- a/datalad/interface/base.py
+++ b/datalad/interface/base.py
@@ -100,22 +100,6 @@ def get_interface_groups(include_plugins=False):
     return grps
 
 
-def get_interface_name(intfspec, domain='python'):
-    """Given interface spec, return the func/method name"""
-    assert domain == 'python', "no other ATM supported for this function"
-
-    if len(intfspec) >= 4:
-        name = intfspec[3]
-    else:
-        # there might be just 2 entries - module, class
-        if len(intfspec) <= 2:
-            name = intfspec[1].lower()
-        else:
-            name = intfspec[2]
-        name = name.replace('-', '_')
-    return name
-
-
 def get_cmd_summaries(descriptions, groups, width=79):
     """Return summaries for the commands in `groups`.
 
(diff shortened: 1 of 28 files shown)


yarikoptic:
Note: travis must be all green from now on. Someone yet to get hands dirty in Windows

datalad/interface/base.py
```diff
@@ -100,6 +100,22 @@ def get_interface_groups(include_plugins=False):
     return grps
 
 
+def get_interface_name(intfspec, domain='python'):
```
kyleam: Doesn't `get_api_name` exist to handle "interface spec -> python api name" conversions?

yarikoptic:
Than you Kyle!! Hard to answer per each on the phone, overall you are probably right on all points, please remove my function if there is already one to use - I must have missed it. Regarding ineffective test, I have made judgement based on the debug messages and they were there, so not sure how it wasn't in effect yet

yarikoptic:
re test not in effect -- I only can guess that I've removed those api imports but then while preparing the "clean commit" reverted those changes when found that they are needed... oh well -- thanks for catching! RFing now

yarikoptic:
As far as I care - it is ready to be merged. If anything, might just want some more manual testing

yarikoptic:
actually now I think we should also add handling of plugins there for consistent behavior with .api imports

kyleam:
> (doing it now)

OK.  Just in case you didn't stumble upon `load_interface`:


```diff
diff --git a/datalad/distribution/dataset.py b/datalad/distribution/dataset.py
index 208f47bb4..8cae0d19c 100644
--- a/datalad/distribution/dataset.py
+++ b/datalad/distribution/dataset.py
@@ -179,16 +179,14 @@ def __getattr__(self, attr):
         # provided to @datasetmethod and what is defined in interfaces
         if not attr.startswith('_'):  # do not even consider those
             from datalad.interface.base import (
-                get_interface_groups, get_api_name
+                get_interface_groups, get_api_name, load_interface
             )
-            for _, _, interfaces in get_interface_groups():
+            for _, _, interfaces in get_interface_groups(include_plugins=True):
                 for intfspec in interfaces:
                     # lgr.log(5, "Considering interface %s", intfspec)
                     name = get_api_name(intfspec)
                     if attr == name:
-                        from importlib import import_module
-                        # turn the interface spec into an instance
-                        import_module(intfspec[0], package='datalad')
+                        load_interface(intfspec)
                         # Now it must be bound
                         meth = getattr(self, attr, None)
                         if meth:
```

yarikoptic:
pushed, but I am not certain about 60b57d88e6dc4cdba72c884ba4d71d622f546361 -- may be worth keeping it on the 6db6f6c4f746ba3c1f163f05e0d90a170547ca03 to be safe (do go through all without early breaks) -- since that OPT unlikely to save much

yarikoptic:
uff -- thanks for all the findings - didn't run into `load_interface` indeed ... grr... will revert the last one and will go with exhaustive loop and load_interface

kyleam:
Failed run is due to a connectivity issue.  Restarted.

yarikoptic:
Ok, travis is green, all is good.  Let's give @mih another day possibly to express his opinion and then merge.

yarikoptic:
Let's hope for the best

PR #3164, opened 2019-02-14
BF: set annex.merge-annex-branches=false for invocations not requiring updated remote information

yarikoptic:
git-annex will try to merge remote's git-annex branches pretty much for any command which could potentially benefit from updated availability information present in those updated remote git-annex
branches.  But if that fails, the whole call fails, even if what we are asking does not require updated remote information.  Original whining/analysis+response from @joeyh is a year old and is at https://git-annex.branchable.com/bugs/impossible_to_perform___34__read-only__34___git_annex_info_without_write_permissions/

Use cases where we ran into it already:
-  `datalad ls -L`  of a dataset owned by someone else
-  `datalad get files`   whenever files load is already there but partition
      is not writeable (singularity image, see https://github.com/poldracklab/fmriprep/issues/1500#issuecomment-463384533

So with this change we set `-c annex.merge-annex-branches=false` for those
git annex invocations where (I think, you check!) we do not really need any git-annex
branch being merged since we care only about local stuff.

This commit comes without a test since I have failed to simulate the
failing scenario without sudo chown someonelse -R testrepo . We could
probably do that on Travis, but I thought first to check if there
could be better ideas .  If not, and tests pass - I guess we better merge soonish.

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -2655,8 +2655,10 @@ class AnnexRepo(GitRepo, RepoInterface):
 
         options = ['--bytes', '--fast'] if fast else ['--bytes']
 
+        # note - shouldn't use merge_annex_branches=False because updated
+        # info/description about remote UUIDs might be needed
         json_records = list(self._run_annex_command_json(
-            'info', opts=options, merge_annex_branches=False))
+            'info', opts=options))
         assert(len(json_records) == 1)
 
         # TODO: we need to abstract/centralize conversion from annex fields
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -1023,6 +1023,7 @@ class AnnexRepo(GitRepo, RepoInterface):
                            git_options=None, annex_options=None,
                            backend=None, jobs=None,
                            files=None,
+                           merge_annex_branches=True,
                            **kwargs):
         """Helper to run actual git-annex calls
 
@@ -1045,6 +1046,10 @@ class AnnexRepo(GitRepo, RepoInterface):
             If command passes list of files. If list is too long
             (by number of files or overall size) it will be split, and multiple
             command invocations will follow
+        merge_annex_branches: bool, optional
+            If False, annex.merge-annex-branches=false config will be set for
+            git-annex call.  Useful for operations which are not intended to
+            benefit from updating information about remote git-annexes
         **kwargs
             these are passed as additional kwargs to datalad.cmd.Runner.run()
 
@@ -1063,6 +1068,9 @@ class AnnexRepo(GitRepo, RepoInterface):
 
         if not self.always_commit:
             git_options += ['-c', 'annex.alwayscommit=false']
+
+        if not merge_annex_branches:
+            git_options += ['-c', 'annex.merge-annex-branches=false']
 
         if git_options:
             cmd_list = ['git'] + git_options + ['annex']
@@ -1354,7 +1362,9 @@ class AnnexRepo(GitRepo, RepoInterface):
 
         if not key:
             expected_downloads, fetch_files = self._get_expected_files(
-                files, ['--not', '--in', 'here'])
+                files, ['--not', '--in', 'here'],
+                merge_annex_branches=False  # interested only in local info
+            )
         else:
             fetch_files = files
             assert len(files) == 1, "When key=True only a single file be provided"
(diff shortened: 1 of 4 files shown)


datalad/support/annexrepo.py
```diff
@@ -1023,6 +1023,7 @@ def _run_annex_command(self, annex_cmd,
                            git_options=None, annex_options=None,
                            backend=None, jobs=None,
                            files=None,
+                           merge_annex_branches=False,
```
yarikoptic: oh boy -- that should have been True (I was messing around with name...)

effigies:
> without sudo chown someonelse -R testrepo

`chmod -R -w testrepo`?

yarikoptic:
> > without sudo chown someonelse -R testrepo
> 
> `chmod -R a-w testrepo`?

my git/annex smokes something stronger -- couldn't scare it away this way ;-)  can you?

effigies:
`mount --bind -o ro`?

yarikoptic:
If sudo is there and can be ran, I can just chown root the directory

yarikoptic:
Could probably done via basic fuse layer... Also too involved

datalad/support/annexrepo.py
```diff
@@ -2636,7 +2655,10 @@ def repo_info(self, fast=False):
 
         options = ['--bytes', '--fast'] if fast else ['--bytes']
 
-        json_records = list(self._run_annex_command_json('info', opts=options))
+        # note - shouldn't use merge_annex_branches=False because updated
+        # info/description about remote UUIDs might be needed
+        json_records = list(self._run_annex_command_json(
```
kyleam: OK, but this does mean that the `datalad ls -L` usecase isn't resolved by this PR, right?

yarikoptic: I thought it was even after my fix but I could be wrong... The test will show. Thanks for catching this!

yarikoptic:
you were correct about `repo_info` @kyleam  -- I will never trust myself again (until I do prepare a test for the change ;)).  Now comes with a test via sudo - let's see if it doesn't get skipped on Travis. Works locally as well, as long as I pre-seed sudo session ;-)  evil evil evil - but I found no other way

yarikoptic:
ha -- current failure is interesting in that it is a regression in git-annex while working in direct mode -- absent with 7.20181205+git27-g21eaaac6e-1~ndall+1 but present with current 7.20190129+git78-g3fa6be1fe-1~ndall+1  - `git pull` freaks out. (see below @joeyh if of interest).  Since we are to abandon direct mode, I guess I will not care and just replace that part of the test (it is not actually that much relevant there)

```
   File "/home/travis/build/datalad/datalad/datalad/support/tests/test_annexrepo.py", line 2621, in test_ro_operations
    repo2.pull()
  File "/home/travis/build/datalad/datalad/datalad/support/gitrepo.py", line 1996, in pull
    **kwargs
  File "/home/travis/build/datalad/datalad/datalad/support/gitrepo.py", line 1956, in _call_gitpy_with_progress
    ret = callable(**git_kwargs)
  File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/git/remote.py", line 808, in pull
    res = self._get_fetch_info_from_stderr(proc, progress)
  File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/git/remote.py", line 675, in _get_fetch_info_from_stderr
    proc.wait(stderr=stderr_text)
  File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/git/cmd.py", line 415, in wait
    raise GitCommandError(self.args, status, errstr)
GitCommandError: Cmd('/usr/lib/git-annex.linux/git') failed due to: exit code(1)
  cmdline: /usr/lib/git-annex.linux/git -c receive.autogc=0 -c gc.auto=0 -c core.bare=False --work-tree=. pull --progress -v origin refs/heads/master
  stderr: 'fatal: Couldn't find remote ref refs/heads/master
fatal: the remote end hung up unexpectedly'
```

yarikoptic:
didn't we all miss our lovely buildbots? ;-)

yarikoptic:
Heh... NFS causes troubles obviously - will need to either skip of fails to chown or adjust NFS Mont options to allow passing root

yarikoptic:
finally travis is green! and I've checked -- it does run the test on travis:
```
datalad.support.tests.test_annexrepo.test_ro_operations ... /tmp/datalad_temp_tree_test_ro_operationsdvy538wc/clone   [annex]  master  ✗ 2019-02-15/17:45:30  ✓  5 Bytes/5 Bytes
ok
```
so all good -- merging

Issue #3130, opened 2019-01-28
appveyor became unhappy (recent annex release?)


yarikoptic:
initially ran into in https://github.com/datalad/datalad/pull/3129

I think it is due to new annex cmd:annex: 7.20190122-gc94b3a389  since a few days prior that tests were all green.
failure
```
======================================================================
FAIL: datalad.distribution.tests.test_install.test_datasets_datalad_org('',)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Miniconda35\envs\test-environment\lib\site-packages\nose\case.py", line 198, in runTest
    self.test(*self.arg)
  File "C:\Miniconda35\envs\test-environment\lib\site-packages\datalad\tests\utils.py", line 849, in newfunc
    return func(*args, **kwargs)
  File "C:\Miniconda35\envs\test-environment\lib\site-packages\datalad\tests\utils.py", line 615, in newfunc
    return t(*(arg + (filename,)), **kw)
  File "C:\Miniconda35\envs\test-environment\lib\site-packages\datalad\distribution\tests\test_install.py", line 945, in check_datasets_datalad_org
    status='ok')
  File "C:\Miniconda35\envs\test-environment\lib\site-packages\datalad\tests\utils.py", line 1281, in assert_result_count
    dumps(results, indent=1, default=lambda x: str(x))))
AssertionError: Got 0 instead of 1 expected results matching {'status': 'ok'}. Inspected 1 record(s):
[
 {
  "status": "notneeded",
  "path": "C:\\Users\\appveyor\\AppData\\Local\\Temp\\1\\datalad_temp_81qc9hhh\\001-anat-scout_ses-{date}\\000001.dcm",
  "type": "file",
  "refds": "C:\\Users\\appveyor\\AppData\\Local\\Temp\\1\\datalad_temp_81qc9hhh",
  "action": "get",
  "message": "already present"
 }
]
======================================================================
FAIL: datalad.distribution.tests.test_install.test_datasets_datalad_org('/.git',)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Miniconda35\envs\test-environment\lib\site-packages\nose\case.py", line 198, in runTest
    self.test(*self.arg)
  File "C:\Miniconda35\envs\test-environment\lib\site-packages\datalad\tests\utils.py", line 849, in newfunc
    return func(*args, **kwargs)
[...]

PR #3166, opened 2019-02-15
BF: Stop running that failing test on appveyor - needs fix in git-annex

yarikoptic:
This makes all the appveyor runs a useless red mark - hiding away possibly new
bugs we would introduce.  So, I think it is better to disable that test on
appveyor for now

This is a temporary workaround for #3130

### DIFF
--- a/datalad/distribution/tests/test_install.py
+++ b/datalad/distribution/tests/test_install.py
@@ -9,7 +9,10 @@
 
 """
 
-from datalad.tests.utils import known_failure_direct_mode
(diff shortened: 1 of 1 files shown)


yarikoptic:
Appveyor is green, I am ok to merge unless objections/improvements to come

yarikoptic:
Doh, missed your reply... Ok, will do more selective approach

yarikoptic:
ok, appveyor is still green - merge merge merge!

Issue #3175, opened 2019-02-22
patool/datalad.Runner/gzip indigestion - stalls on a sample tiny file


yarikoptic:
Troubleshooting stalling run of a test for the crawler https://github.com/datalad/datalad-crawler/pull/37 , which came about when I recently introduced support for crawling .gz files.

My exploration, first trying on trusty env to just realize the difference in execution with local:

- installing `p7zip-full` "mitigates" the issue. patool then uses `7z` and things work lovely
- with either regular gzip or pigz we would get it stalling at us running them for patool 
- we default to log stdout and stderr in the Runner
- ... more to come ... just wanted to protocol it

yarikoptic:
additional aspect is inconsistency in patool's/archivers behavior for extracting a file, which is in our case is relevant if file is annexed and its filename is actually the annex key: https://github.com/wummel/patool/issues/67

kyleam:
I think this is OK to close. Please re-open if I'm mistaken.

PR #3176, opened 2019-02-23
BF: fix execution of patool commands

yarikoptic:
It is an incomplete fix, which would still cause test failure in the crawler if no p7zip is installed in crawler due to  https://github.com/datalad/datalad/issues/3175 , but wanted to see if anything breaks here before proceeding further

### DIFF
--- a/datalad/support/archives.py
+++ b/datalad/support/archives.py
@@ -81,7 +81,19 @@ def _patool_run(cmd, verbosity=0, **kwargs):
     # use our runner
     try:
         # kwargs_ = kwargs[:];         kwargs_['shell'] = True
-        _runner.run(cmd, **kwargs)
+        # Any debug/progress output could be spit out to stderr so let's
+        # "expect" it.
+        #
+        if isinstance(cmd, (list, tuple)) and kwargs.get('shell'):
+            # patool (as far as I see it) takes care about quoting args
+            cmd = ' '.join(cmd)
+        out, err = _runner.run(cmd,
+                    #log_stdout='offline',
+                    #log_stderr='offline',
+                    #expect_stderr=True,
+                    #stdin=open('/dev/null'),
+                    **kwargs)
+        lgr.debug("Finished running for patool. stdout=%s, stderr=%s", out, err)
         return 0
     except CommandError as e:
         return e.code
--- a/datalad/support/archives.py
+++ b/datalad/support/archives.py
@@ -18,9 +18,14 @@ assert(external_versions["patoolib"] >= "1.7")
 
 import os
 import tempfile
-from .path import join as opj, exists, abspath, isabs, normpath, relpath, pardir, isdir
-from .path import sep as opsep
-from .path import realpath
+from .exceptions import MissingExternalDependency
+from .path import (
+    basename,
+    join as opj,
+    exists, abspath, isabs, normpath, relpath, pardir, isdir,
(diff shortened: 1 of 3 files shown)


yarikoptic:
grr... Not sure which way to proceed here. So

- so far support for `.gz` relied pretty much on the uniform implementation across all archives, where we just extract an archive under a directory and look at the files extracted
- so it was inline with how p7zip handles all of them including `.gz` and `p7zip` does use stored in the .gz header original filename
- not sure yet what is the behavior if no filename was stored in the header

I thought that may be we could provide our custom gunzip functionality which would respect that filename and thus would mitigate discrepancy, but  

- `gzip/gunzip` itself does not care about possibly stored original filename in the gzip header (unless given explicitly `--name` option)
- Python's `gzip` library explicitly reads and discards (does not even store anywhere, bleh) possibly present in the header filename: https://github.com/python/cpython/blob/master/Lib/gzip.py#L427

so I do not see an easy way  to figure out that filename in the header.  But we can't easily switch to "just use original filename stripped of .gz" since 
- we might be operating already on a annex key (so no filename pointing to it)
- would require implementing .gz-special handling of `datalad-archive` urls.  It would actually be the only proper solution probably since that stored filename is not really something to rely on (later on could be extended to other .bz2, .xz etc; I don't think xz even stores the original filename). But once again -- currently `datalad-archive` relies on the handling for those as for any other archive. So we had urls like `dl+archive:MD5E-s24--5590a3cbf10c77d61fb4058b741760d9.gz#path=123&size=0` not `dl+archive:MD5E-s24--5590a3cbf10c77d61fb4058b741760d9.gz#path=MD5E-s24--5590a3cbf10c77d61fb4058b741760d9&size=0` or even more adequate here just `dl+archive:MD5E-s24--5590a3cbf10c77d61fb4058b741760d9.gz#size=0` (without any path within)
- so if we were to provide "proper" treatment/url we might need to provide compatibility layer just in case there is some datalad dataset with pointers to previously .gz'ed files.  I.e. if there is a `path` for any of those .gz, we pretty much ignore it and verify that only a single file is extracted and take it.

yarikoptic:
ok -- for now just decided to blow in no `7z` is installed.

yarikoptic:
oops -- forgot about our bright future, changed base to be 0.11.x

kyleam:
Thanks for summarizing the results of all your digging.

> ok -- for now just decided to blow in no `7z` is installed.

Sounds sensible.

yarikoptic:
with @kyleam 's blessing will merge for now

datalad/support/archives.py
```diff
@@ -81,7 +86,19 @@ def _patool_run(cmd, verbosity=0, **kwargs):
     # use our runner
     try:
         # kwargs_ = kwargs[:];         kwargs_['shell'] = True
-        _runner.run(cmd, **kwargs)
+        # Any debug/progress output could be spit out to stderr so let's
+        # "expect" it.
+        #
+        if isinstance(cmd, (list, tuple)) and kwargs.get('shell'):
+            # patool (as far as I see it) takes care about quoting args
+            cmd = ' '.join(cmd)
```
yarikoptic: Also we might want to consider moving this into the runner itself. I didn't spot any place where we use shell and pass the list, so it shouldn't change any other behavior but might prevent us from hitting it again

Issue #3185, opened 2019-03-01
duplicate warnings while aggregate-metadata on a dataset without any


yarikoptic:
based on snippet I was checking out in https://github.com/datalad/datalad/pull/1995#issuecomment-346950824:

```shell
$> bash .git/trash/gh-1995      
remove(ok): /tmp/test-1995 (dataset)
[INFO   ] Creating a new annex repo at /tmp/test-1995 
create(ok): /tmp/test-1995 (dataset)
add(ok): /tmp/test-1995/avatar.png (file)
save(ok): /tmp/test-1995 (dataset)
action summary:
  add (ok: 1)
  save (ok: 1)
[WARNING] Found no aggregated metadata info file /tmp/test-1995/.datalad/metadata/aggregate_v1.json. You will likely need to either update the dataset from its original location or reaggregate metadata locally. 
[WARNING] Found no aggregated metadata info file /tmp/test-1995/.datalad/metadata/aggregate_v1.json. You will likely need to either update the dataset from its original location or reaggregate metadata locally. 
[INFO   ] Aggregate metadata for dataset /tmp/test-1995 
                                                                                                     [INFO   ] Update aggregate metadata in dataset at: /tmp/test-1995                                    
[WARNING] Found no aggregated metadata info file /tmp/test-1995/.datalad/metadata/aggregate_v1.json. You will likely need to either update the dataset from its original location or reaggregate metadata locally. 
[WARNING] Found no aggregated metadata info file /tmp/test-1995/.datalad/metadata/aggregate_v1.json. You will likely need to either update the dataset from its original location or reaggregate metadata locally. 
[WARNING] Found no aggregated metadata info file /tmp/test-1995/.datalad/metadata/aggregate_v1.json. You will likely need to either update the dataset from its original location or reaggregate metadata locally. 
aggregate_metadata(ok): /tmp/test-1995 (dataset)
[INFO   ] Attempting to save 6 files/datasets 
save(ok): /tmp/test-1995 (dataset)
action summary:
  aggregate_metadata (ok: 1)
  save (ok: 1)
bash .git/trash/gh-1995  4.06s user 1.35s system 96% cpu 5.612 total
```
the snippet:
```shell
$> cat .git/trash/gh-1995
#!/bin/bash

ds=/tmp/test-1995
[ -e "$ds" ] && datalad remove --nocheck $ds
datalad create $ds
cd $ds
[...]

PR #3186, opened 2019-03-02
BF: Add switch to avoid pointless warnings (fixes gh-3185)

mih:
This is an incomplete attempt to tackle the actual problem (the depth of the code stack this warning is issued at, such that it is out of touch with the calling context). But it should mitigate the issue for two
superficial scenarios.

### DIFF
--- a/datalad/metadata/aggregate.py
+++ b/datalad/metadata/aggregate.py
@@ -756,7 +756,8 @@ def _update_ds_agginfo(refds_path, ds_path, subds_paths, incremental, agginfo_db
 
 def _store_agginfo_db(ds, db):
     # base path in which aggregate.json and objects is located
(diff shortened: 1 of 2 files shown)


mih:
Hmm, I need it in master. I am out of touch with what is happening in other branches.

kyleam:
> Hmm, I need it in master. I am out of touch with what is happening in other branches.

After master dropped direct mode, 0.11.x was spun off as a maintenance branch.  It's got a few bug fixes that aren't in master right now, but that's just because a periodic merge into master hasn't happen.

Because your PR is a bugfix and applies cleanly to 0.11.x, I'll move it there.  Once tests pass, I'll merge the PR into 0.11.x and than 0.11.x into master.

PR #3193, opened 2019-03-03
Prepare for first preview release of 0.12 series

mih:
I see no point of adding tons of changelog entries for individual
pre-releases (not even a real RC yet). Instead, I started an 0.12.0 entry
so we can fill it up with what is to come.

I would have preferred to call this 0.12.0pre1, but setuptools insists
on "normalizing" the version to 0.12.0rc1 -- so we better use that right
away.

Once this gets a GO, I will merge it, push a tag (I don't see the need to pollute github releases, so I'll skip that), upload to pypi, point -revolution to use it as a dependency, and be done.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,21 +10,24 @@ We would recommend to consult log of the
 [DataLad git repository](http://github.com/datalad/datalad) for more details.
 
 
-## 0.11.4 (??? ??, 2019) -- will be better than ever
-
-bet we will fix some bugs and make a world even a better place.
+## 0.12.0rc1 (Mar 03, 2019) -- to boldly go ...
 
 ### Major refactoring and deprecations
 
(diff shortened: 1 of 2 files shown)


mih:
Hmm, I was too eager. The changelog was broken for a while already it seems...

mih:
So a tests prevents having changelog entries that are incomplete. So I guess we will have a ton of RC changelog entries from now on...

mih:
Round four in trying to bump the version string. I will not wait till the test have failed again, but continue on this epic journey on another day.

yarikoptic:
> So a tests prevents having changelog entries that are incomplete. So I guess we will have a ton of RC changelog entries from now on...

or adjust the test (which one?) to allow for rcs toward the release? although I would not mind good summaries per rc to actually understand the difference between them

PR #3201, opened 2019-03-06
ENH: Uniformize API and capabilities of JSON dump and dump2stream

mih:
This PR makes it a bit easier to read/write compressed JSON dumps, by softening the association stream==compressed, dump==uncompressed. Nothing amazing though.

Also a few more "tests".

Required for https://github.com/datalad/datalad-revolution/pull/84

### DIFF
--- a/datalad/support/json_py.py
+++ b/datalad/support/json_py.py
@@ -50,21 +50,28 @@ from ..log import lgr
 from ..dochelpers import exc_str
 
 
-def dump(obj, fname):
+def dump(obj, fname, compressed=False):
+
+    _open = LZMAFile if compressed else io.open
+
     indir = dirname(fname)
     if not exists(indir):
         makedirs(indir)
     if lexists(fname):
         os.unlink(fname)
-    with io.open(fname, 'wb') as f:
-        return dump2fileobj(obj, f)
+    with _open(fname, 'wb') as f:
+        return dump2fileobj(
+            obj,
+            f,
+            **(compressed_json_dump_kwargs if compressed else json_dump_kwargs)
+        )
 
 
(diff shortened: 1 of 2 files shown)


datalad/support/json_py.py
```diff
@@ -50,21 +50,28 @@
 from ..dochelpers import exc_str
 
 
-def dump(obj, fname):
+def dump(obj, fname, compressed=False):
+
+    _open = LZMAFile if compressed else open
```
yarikoptic: Wasn't io.open instead of playing open used for some Unicode compatibility on py2? Ie I just wonder if there was a reason not warranting this change

mih: Ah indeed, PY2... Adjusted....

datalad/support/json_py.py
```diff
@@ -50,21 +50,28 @@
 from ..dochelpers import exc_str
 
 
-def dump(obj, fname):
+def dump(obj, fname, compressed=False):
```
yarikoptic: somewhat has mentioned docstrings today ;-) is that why it is still a "Draft"?

mih: Yes, there is likely more to come. This PR could grow alongside the demands resulting from the rewrite of metadata aggregation happening in -revolution right now. I made it a draft, because there is no need to merge this piecewise -- and also because I really like this ability to have real draft PRs (watch what someone is working on) instead of the various ways in which we tag real PRs to make them pretend drafts ;-)

mih:
Test failure is the crawler's direct mode issue and unrelated.

mih:
Thx.

PR #3205, opened 2019-03-06
BF: Fix LORIS token authenticator type with Python3

driusan:
The HTTPResponse object returned by urlopen under Python3 is not
compatible with json.load(!) because read() returns bytes, while json.load
is a thin wrapper around json.loads (which expects a str). This stores
the response as a str in a temporary variable and converts it to a string,
before calling json.loads directly to prevent a type error when used under
Python3 while trying to authenticate against a LORIS instance.

### DIFF
--- a/datalad/support/third/loris_token_generator.py
+++ b/datalad/support/third/loris_token_generator.py
@@ -39,6 +39,7 @@ class LORISTokenGenerator(object):
         except HTTPError:
             raise AccessDeniedError("Could not authenticate into LORIS")
 
-        data = json.load(response)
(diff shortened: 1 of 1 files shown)


driusan:
@kyleam I'm not really familiar with the function, but I think it should be using it (properly?) now.

kyleam:
Thanks, looks good to me.  We've recently gained a maintenance branch, so I've moved this PR on top of that.  (We should document this in the pull request template.)

kyleam:
The test failure is a known one: datalad/datalad-revolution#92

PR #3207, opened 2019-03-06
BF: Fixing for the needed post-release dance with Changelog

yarikoptic:
Post release commit must introduce a placeholder for the next Changelog entry
and adjust CONTRIBUTING.md to match the after next one so it could be simply
copy pasted next time

This pull should fix currently failing in master `test_version.py`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,23 @@ We would recommend to consult log of the
 [DataLad git repository](http://github.com/datalad/datalad) for more details.
 
 
+## 0.12.0rc2 (??? ??, 2019) -- will be better than ever
+
+bet we will fix some bugs and make a world even a better place.
+
+### Major refactoring and deprecations
+
+- hopefully none
+
+### Fixes
+
+?
+
+### Enhancements and new features
+
+?
+
+
 ## 0.12.0rc1 (Mar 03, 2019) -- to boldly go ...
 
 ### Major refactoring and deprecations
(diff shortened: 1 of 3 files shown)


mih:
What problem is solved by having a propective changelog entry with a speculative version number? And in addition a template in another file that looks just like any another in the changelog, but must have a particular even more speculative version number for the release after the next release?

This appears over-regulated to me.

yarikoptic:
The idea was and remains:

I think it would be great if we add a habit of adding changelog entries for every PR we submit, e.g. when we feel that no more changes needed -- composing them after in bulk is a burden noone likes to endure AFAIK.
For that to happen we need to have a changelog record to add them, and sparse enough to minimize chance of conflicts.  So that is where the initial template comes from.

Re the one in CONTRIBUTING - that is the only one which has all sections and really easy to use - just cut/paste into the CHANGELOG instead of possibly "copy previous one over as a new one, fix up version, remove dates, remove existing entries, add needed sections".

So, may be a bit regulated indeed, but was a really minor pain for some gain.

mih:
That sound very sensible to me, but I believe that none of the advantages would be lost by not having to envision version numbers and just keep them at ??? in all places other than version.py.

yarikoptic:
Note that the comment at that assertion says that it is testing for the scenario where we forget to tag the release. I think it was triggered by a real case

yarikoptic:
either ware to change current procedure(s) I guess is for a separate discussion, so to bring master back into green zone - merging

PR #3210, opened 2019-03-07
Do not decode fetched content when retrieving status

driusan:
When annexificator is used in a datalad-crawler pipeline, it attempts
to call get_status on the URL to see if the file has changed. get_status
attempts to use _fetch, which attempts to decode the content under Python3.

This is invalid for content that isn't utf-8, and get_status was only
trying to verify that it wasn't receiving a login or failure page or similar.
_fetch, confusingly, converts the UnicodeFormatError into a generic DownloadError
(despite the fact that the download succeeded, but conversion to a string failed.)

This teaches _fetch a "decode" parameter that toggles whether or not it should
even attempt to decode. In the case of get_status, the content is discarded,
so the decode serves no purpose as far as I can tell.

NB. fetch (without the _)'s documentation claims that it doesn't decode, but
it calls _fetch (which currently always decodes.) This can new parameter can
eventually be used to fix that, but currently trying to fix "fetch" results in
a json.loads error from a mysterious part in the code if fetch is updated to
not decode as documented.

### DIFF
--- a/datalad/downloaders/base.py
+++ b/datalad/downloaders/base.py
@@ -487,7 +487,7 @@ class BaseDownloader(object):
             atexit.register(self._cache.close)
         return self._cache
 
-    def _fetch(self, url, cache=None, size=None, allow_redirects=True):
+    def _fetch(self, url, cache=None, size=None, allow_redirects=True, decode=True):
         """Fetch content from a url into a file.
 
         Very similar to _download but lacks any "file" management and decodes
@@ -541,7 +541,7 @@ class BaseDownloader(object):
             downloaded_size = len(content)
 
             # now that we know size based on encoded content, let's decode into string type
-            if PY3 and isinstance(content, binary_type):
+            if PY3 and isinstance(content, binary_type) and decode:
                 content = content.decode()
             # downloaded_size = os.stat(temp_filepath).st_size
 
@@ -615,7 +615,7 @@ class BaseDownloader(object):
             and self.authenticator.failure_re \
             else 0
 
-        _, headers = self._fetch(url, cache=False, size=download_size)
+        _, headers = self._fetch(url, cache=False, size=download_size, decode=False)
 
         # extract from headers information to depict the status of the url
         status = self.get_status_from_headers(headers)


driusan:
This is trying to fix a crawler that I'm writing here: https://github.com/driusan/datalad-crawler/blob/enh-loriscrawler/datalad_crawler/pipelines/loris.py which only gets triggered under Python3.. but if my understanding of the code is right, it should be triggered with any datalad-crawler pipeline where a generator directly yields a URL to an annexificator node and the url contains binary data.

kyleam:
> So it might be worth removing that code to see if anything breaks

Let's see what happens with https://travis-ci.org/datalad/datalad/builds/503692960

kyleam:
OK, there is one failing test when we drop the py3-specific decoding:

```
FAIL: datalad.downloaders.tests.test_http.test_HTTPDownloader_basic
    self.test(*self.arg)
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/datalad/tests/utils.py", line 442, in newfunc
    return t(*(arg + (d,)), **kw)
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/datalad/tests/utils.py", line 554, in newfunc
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/datalad/downloaders/tests/test_http.py", line 121, in test_HTTPDownloader_basic
    assert_equal(downloader.fetch(furl), 'abc')
AssertionError: b'abc' != 'abc'
```

IMO the test is wrong because on Python 3 it's asserting a unicode result despite what is documented.

core's [crawler test run][0] isn't too informative because only a small subset of the tests are executed (I'm guessing some issue with how the nose selector is specified; haven't looked).  But if I run py3 crawler tests locally, I see one failure:

```
ERROR: datalad_crawler.pipelines.tests.test_crcns.test_get_metadata
  File "/home/kyle/src/python/venvs/datalad-crawler/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/kyle/src/python/datalad/datalad/tests/utils.py", line 849, in newfunc
    return func(*args, **kwargs)
  File "/home/kyle/src/python/datalad-crawler/datalad_crawler/pipelines/tests/test_crcns.py", line 29, in test_get_metadata
    all_meta = get_metadata()
  File "/home/kyle/src/python/datalad-crawler/datalad_crawler/pipelines/crcns.py", line 69, in get_metadata
    rj = fetch_datacite_metadata()
  File "/home/kyle/src/python/datalad-crawler/datalad_crawler/pipelines/crcns.py", line 42, in fetch_datacite_metadata
    return json.loads(text)
  File "/usr/lib/python3.5/json/__init__.py", line 312, in loads
    s.__class__.__name__))
TypeError: the JSON object must be str, not 'bytes'
```

We should fix this caller in crcns, but so that we don't have compatibility issues between the core and the extension, I think we should also update `get_cached_url_content`, whose only caller is crawler's crcns, to return unicode.

[0]: https://travis-ci.org/datalad/datalad/jobs/503692979

So, I'll plan to move things in this direction:

- [ ] drop py3 decoding in `_fetch`
- [ ] make crawler's crncns decode for `json.loads`
- [ ] make `get_cached_url_content` return unicode

yarikoptic:
Some tests for crawler might not run because require credentials (for aws, or crcns).  I will share crcns ones to use later on privately just in case

yarikoptic:
I think appveyor failure is unrelated.

driusan:
@kyleam Yes, those are the exact errors I'm trying to work around. I can't catch the UnicodeDecodeError in the caller because fetch converts it to a DownloadError, I can't catch DownloadError in the caller because it could be a real download error, and I didn't want to have fetch not convert the  UnicodeDecodeError to a DownloadError because I wasn't sure how many places depend on the current behaviour, so I just added a parameter to avoid it.

I'm surprised only one test is failing when you remove it. When I tried to pass Decode=False from fetch (so that it works as documented) I was getting similar type errors to what you're seeing in datalad_crawler.pipelines.tests.test_crcns.test_get_metadata from pretty much everywhere.

kyleam:
> I'm surprised only one test is failing when you remove it. When I tried to pass Decode=False from fetch (so that it works as documented) I was getting similar type errors to what you're seeing in datalad_crawler.pipelines.tests.test_crcns.test_get_metadata from pretty much everywhere.

Eh good to know.  I might be doing something silly on my end.

driusan:
It's the same json.loads error that I was seeing, so it's possible that there is only one root cause of it and I was just seeing it from my own crawler when I tried

kyleam:
OK, given that this is a targeted fix to an internal helper function and that the commit message explains the outstanding issues with `fetch`, I'd say we go ahead with it.  I'll have a look at fixing `fetch` (and hopefully #3212 will make the crawler run more informative regarding the effects of the change).

PR #3184, opened 2019-03-01
ENH: duecredit support for citing datalad itself AND datasets it opens

yarikoptic:
This pull request fixes #2065 - those eyes didn't let me sleep! I will blame @tyarkoni for anything to come.
- includes support for actually DueCrediting any dataset we "instantiate"
- ATM just ad-hoc extraction of metadata fields from `bids` and `datacite` (for CRCNS) records. Proper solution will need integration with extractors or some other harmonization which we do not have ATM
- Relies on changes yet to be merged/released in DueCredit: https://github.com/duecredit/duecredit/pull/144
   - basic support for free text citations (so we could at least point to those datasets etc)
   - access to `due.active`

With those changes and this PR, we are getting smth like:
```shell
> rm .duecredit.p; DUECREDIT_ENABLE=1 datalad ls ~/datalad/openfmri/ds000241 /home/yoh/datalad/crcns/aa-1                                                                        
/home/yoh/datalad/openfmri/ds000241   [annex]  master  1.0.0+1-2-g6bb6937 2018-06-08/07:44:21  ✓
/home/yoh/datalad/crcns/aa-1   [annex]  master  ✗ 2018-06-08/10:23:47  ✓

DueCredit Report:
- DataLad - Data management and distribution platform / datalad (v 0.11.3.dev22) [1]
  - Single-unit recordings from two auditory areas in male zebra finches / datalad:77271f9e (v 1.0) [2]
  - AK6: Animal Kingdom; 6 Species / datalad:f7ccf78c (v 1.0.0+1-2-g6bb6937) [3]

1 package cited
0 modules cited
2 functions cited

References
----------

[1] Visconti di Oleggio Castello, M. et al., 2019. datalad/datalad 0.11.3.
[2] Theunissen, F.E. et al., 2009. Single-unit recordings from two auditory areas in male zebra finches.
[3] Please cite: 'Connolly, A. C., J. S. Guntupalli, J. Gors, M. Hanke, Y. O. Halchenko, Y.-C. Wu, H. Abdi, and J. V. Haxby. The Representation of Biological Classes in the Human Brain. Journal of Neuroscience 32.8 (2012): 2608-618.' and include the following message: 'This data was obtained from the OpenfMRI database. Its accession number is ds000241'
```

```shell
$> duecredit summary --format bibtex              
@misc{https://doi.org/10.5281/zenodo.808846,
  doi = {10.5281/zenodo.808846},
  url = {https://zenodo.org/record/808846},
[...]

PR #3215, opened 2019-03-11
BF: Fix PWD to match provided cwd in Runner

yarikoptic:
We do take care about tuning up PWD while chdir etc.  But when we run
external process and set its cwd, we left PWD unchanged if was provided
in (likely) inherited and tuned environment.  This fix would change
PWD to match provided cwd, if it is known to the environment and cwd
is set.

I was not bold enough to always get an env (if not provided) and change it
there, but may be that is what we will need to end up doing at some
point.

This should resolve the failing test in
https://github.com/datalad/datalad/pull/3184
caused by underlying datalad process, which is started from a temp directory
while we provide it with a now non-matching PWD:
https://travis-ci.org/datalad/datalad/jobs/501525590#L4325

### DIFF
--- a/datalad/tests/test_cmd.py
+++ b/datalad/tests/test_cmd.py
@@ -291,7 +291,11 @@ def test_runner_fix_PWD(path):
     env = os.environ.copy()
     env['PWD'] = orig_cwd = os.getcwd()
     runner = Runner(cwd=path, env=env)
-    out, err = runner.run(u"echo $PWD", cwd=path)
+    out, err = runner.run(
+        [sys.executable, '-c', 'import os; print(os.environ["PWD"])'],
+        cwd=path,
+        shell=False
+    )
     eq_(err, '')
     eq_(out.rstrip(os.linesep), path)  # was fixed up to point to point to cwd's path
     eq_(env['PWD'], orig_cwd)  # no side-effect
--- a/datalad/tests/test_cmd.py
+++ b/datalad/tests/test_cmd.py
@@ -285,6 +285,18 @@ def test_runner_failure_unicode(path):
         runner.run(u"β-command-doesnt-exist", cwd=path)
 
 
+@skip_if_on_windows  # likely would fail
+@with_tempfile(mkdir=True)
+def test_runner_fix_PWD(path):
+    env = os.environ.copy()
+    env['PWD'] = orig_cwd = os.getcwd()
+    runner = Runner(cwd=path, env=env)
+    out, err = runner.run(u"echo $PWD", cwd=path)
+    eq_(err, '')
+    eq_(out.rstrip(os.linesep), path)  # was fixed up to point to point to cwd's path
+    eq_(env['PWD'], orig_cwd)  # no side-effect
+
+
 @with_tempfile(mkdir=True)
 def test_git_path(dir_):
     from ..support.gitrepo import GitRepo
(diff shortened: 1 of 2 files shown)


yarikoptic:
it got me thinking -- may be we should check and reset PWD upon start of datalad right away but then without a warning

datalad/tests/test_cmd.py
```diff
@@ -285,6 +285,18 @@ def test_runner_failure_unicode(path):
         runner.run(u"β-command-doesnt-exist", cwd=path)
 
 
+@skip_if_on_windows  # likely would fail
+@with_tempfile(mkdir=True)
+def test_runner_fix_PWD(path):
```
kyleam: This test passes for when I apply it, without the cmd.py changes, on top of this PR's base.

```
% git rev-parse HEAD
5d7014599a1476355df986d167f4b14689c1ff0c

% python -m nose -vs datalad/tests/test_cmd.py:test_runner_fix_PWD
datalad.tests.test_cmd.test_runner_fix_PWD ... ok
Versions: appdirs=1.4.3 boto=2.49.0 cmd:annex=7.20190219 cmd:git=2.21.0 cmd:system-git=2.21.0 cmd:system-ssh=7.4p1 exifread=2.1.2 git=2.1.11 gitdb=2.0.5 humanize=0.5.1 iso8601=0.1.12 keyring=18.0.0 keyrings.alt=3.1.1 msgpack=0.6.1 mutagen=1.42.0 requests=2.21.0 six=1.12.0 wrapt=1.11.1
Obscure filename: str=b' "\';a&b&c\xce\x94\xd0\x99\xd7\xa7\xd9\x85\xe0\xb9\x97\xe3\x81\x82 `| ' repr=' "\';a&b&cΔЙקم๗あ `| '
Encodings: default='utf-8' filesystem='utf-8' locale.prefered='UTF-8'
Environment: LANG='en_US.UTF-8' PYTHONPATH='/home/kyle/.local/lib/python' GIT_SSL_CAINFO='/home/kyle/.guix-profile/etc/ssl/certs/ca-certificates.crt' GIT_EXEC_PATH='/home/kyle/.guix-profile/libexec/git-core' PATH='/home/kyle/src/python/venvs/datalad/bin:/home/kyle/src/python/venvs/datalad//bin:/home/kyle/.guix-profile/bin:/home/kyle/.config/guix/current/bin:/home/kyle/.cabal/bin:/usr/lib/ccache/bin:/home/kyle/.local/bin:/usr/local/bin:/usr/bin:/bin:/usr/local/games:/usr/games'

Ran 1 test in 0.091s

OK
```

yarikoptic: strange -- will check

yarikoptic:
changed base to 0.11.x

datalad/tests/test_cmd.py
yarikoptic: ha -- it is (99% confidence guess) because test uses "shell=True" so sure thing that shell fixes it up.  So the fixup is needed only if shell is not True. Thanks for catching this!!!

datalad/tests/test_cmd.py
```diff
@@ -285,6 +285,22 @@ def test_runner_failure_unicode(path):
...
+    env = os.environ.copy()
+    env['PWD'] = orig_cwd = os.getcwd()
+    runner = Runner(cwd=path, env=env)
+    out, err = runner.run(
+        ['python', '-c', 'import os; print(os.environ["PWD"])'],
```
kyleam: Elsewhere we've avoided assuming `python` is available (see #2287).  How about using `sys.executable` instead?

yarikoptic: right! that was the intention - will fix up now, thanks!

PR #3220, opened 2019-03-12
BF: specify correctly dependency on pylibzma as prior python 3.3

yarikoptic:
0.12.0rc1 was sdist uploaded while building package using python3. That
caused pyliblzma not being listed altogether as dependency, and thus no
longer autoinstalled.  The reason was this ad-hoc specification.  It might
have been done by me this way because of elderly Debians but it is just not
right in the Python world, so we need to have it fixed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -57,14 +57,6 @@ on_windows = platform_system == 'windows'
 if dist[0] == 'debian' and dist[1].split('.', 1)[0] == '7':
     keyring_requires = ['keyring<8.0']
 
-# lzma is included in python since 3.3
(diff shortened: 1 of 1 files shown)


bpoldrack:
>  The reason was this ad-hoc specification. It might
have been done by me this way

I think that's actually my fault. So, thanks.

yarikoptic:
> > The reason was this ad-hoc specification. It might
> > have been done by me this way
> 
> I think that's actually my fault. So, thanks.

we both can be blamed, me in 99cf56666cd61d2490965dca6a8c64cab37a3336, so no worries ;-)

PR #3221, opened 2019-03-12
BF: Report on deleted files

mih:
Bug was originally reported at https://github.com/datalad/datalad-revolution/issues/94
but has its origin in the GitRepo method.

### DIFF
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -3337,7 +3337,16 @@ class GitRepo(RepoInterface):
                     # we would always see individual files
                     recursive=False):
                 # TODO normalize result
-                yield r
+                yield get_status_dict(
+                    action='delete',
+                    refds=self.pathobj,
+                    # TODO make remove() report the type
+                    # for now it claims to report on files only
+                    type='file',
+                    path=(self.pathobj / ut.PurePosixPath(r)),
+                    # make remove() report on failures too
+                    status='ok',
+                    logger=lgr)
 
         # TODO this additonal query should not be, base on status as given
         # if anyhow possible, however, when paths are given, status may


mih:
Coverage should be there, but clearly it was lacking in evaluation. The todo in the diff points to another potential issue. I'll try to improve the tests quickly.

mih:
Ok, crawler's direct mode issue is the only failure. Going in...

datalad/support/gitrepo.py
```diff
@@ -3337,7 +3337,16 @@ def save_(self, message=None, paths=None, _status=None, **kwargs):
                     # we would always see individual files
                     recursive=False):
                 # TODO normalize result
-                yield r
+                yield get_status_dict(
+                    action='delete',
+                    refds=self.pathobj,
+                    # TODO make remove() report the type
+                    # for now it claims to report on files only
```
yarikoptic: I would have  that assumption coded in, so whenever it fails we would know where to fix. Feels strange that original `r` is completely ignored (besides taking part in `path`).

mih: It is not clear to me what is requested.

mih:
I am not planning on writing a duplicate test. If this is a problem, this PR needs to join the other code that is waiting for the revolution commands, as it is no more or less special.

yarikoptic:
Ok, gimme some hours and I will do both for you if you aren't willing.
The test will not be duplicate since no test failed before in the core before this change

kyleam:
> The test will not be duplicate since no test failed before in the core before this change

There are already tests being added on -revolution's side (see the references in this thread), which is where the [rest of the tests](https://github.com/datalad/datalad/pull/3183#issuecomment-468755440) live for these methods that came over from the recent -revolution merge.  It isn't ideal, but it is temporary.

yarikoptic:
> > The test will not be duplicate since no test failed before in the core before this change
> 
> There are already tests being added on -revolution's side (see the references in this thread), which is where the [rest of the tests](https://github.com/datalad/datalad/pull/3183#issuecomment-468755440) live for these methods that came over from the recent -revolution merge. It isn't ideal, but it is temporary.

ah, that is what I just realized that there is no tests for these methods here in the core yet... ok - no test then here.  I will just add the assertion FWIW

yarikoptic:
and re assertion - I was also under wrong impression that `r` is the "record" not just "result" which is a filename.  Nothing to assert/check there I guess. Sorry for the noise

mih:
Thx.

PR #3226, opened 2019-03-15
Changes from master applicable to 0.11.x

yarikoptic:
* b0d383eb - ENH(TST): cherry picked test from a BF in master
* 27b9905d - ENH: Use common checksuming code, reduce memory load


yarikoptic:
I do not think there is anything to discuss here and it is all green, so proceeding forward

PR #3227, opened 2019-03-15
TEMP: test using -devel annex from neurodebian

yarikoptic:
should be 7.20190219+git191-g2d6a364d4-1~ndall+1 now
- [x] verify if it still works with direct mode


yarikoptic:
so, seems to be all good and I do not see any changes in its behavior (tests pass) in relationship to direct mode (attn: @mih)
I will now test if master is ok with it , and then upload this new snapshot of annex if all good

yarikoptic:
Travis seems to not care about changed base (doesn't it test the merged state not the PR branch state?), so will explicitly rebase it on master

yarikoptic:
ok, all green

yarikoptic:
will test a new one

yarikoptic:
so, seems to be ok.

PR #3228, opened 2019-03-15
Rel 0.11.4 planing

yarikoptic:
- [x] CHANGELOG (contributions/assistance is welcome; otherwise will do tomorow or so)
- [x] [PRs intended for 0.11.4](https://github.com/datalad/datalad/pulls?q=is%3Aopen+is%3Apr+milestone%3A"Release+0.11.4")
- [PRs "hoped" for  0.11.x and might still appear in the release](https://github.com/datalad/datalad/pulls?utf8=✓&q=is%3Aopen+is%3Apr+milestone%3A"0.11.x"+)
     - [x] https://github.com/datalad/datalad/pull/3184 - DueCredit -- not sure since might be destructive, no feedback received
     - [ ] https://github.com/datalad/datalad/pull/3180 - GitHub authentication -- would love to have but needs test(s|ing)

PR #3237, opened 2019-03-18
Prep for 0.12.0rc2 release -- as -revolution dependency

mih:


CHANGELOG.md
```diff
@@ -10,21 +10,28 @@ We would recommend to consult log of the
...
 
-### Major refactoring and deprecations
+- `GitRepo.dirty` does not report on nested empty directories ([#3196]).
 
-- hopefully none
+- LORIS token authenticator now works with Python 3.
 
-### Fixes
+- Introduce an explicit switch to enable decoding of fetched content
+  in `BaseDownloader`
```
kyleam: The LORIS and BaseDownloader entries will already be covered by 0.11.4 (#3228), so I don't think we need them here.

mih: OK, thx for the info. Will kill them later, don't want to go through another test run, but rather proceed with the revolution partner release. Hope that is fine with all.

kyleam: Works for me.  Thanks.

Issue #3055, opened 2018-12-04
Superds fails to find subds metadata whenever subds doesn't have it aggregated and is installed


yarikoptic:
For upcoming `///openneuro` I am aggregating metadata into that superdataset from subdatasets (https://github.com/datalad/datalad-crawler/pull/28/files#diff-8e8fc59a503a8bdc5f90e33e16d020b7R146), which seems to work lovely.  But then I have discovered that I cannot query metadata within subdataset whenever it is installed:
```shell
$> datalad ls ds000008 
ds000008   [annex]  master  ✗ 2018-07-13/21:41:18  ✓

$> datalad -f json_pp metadata --reporton all ds000008/sub-01/anat/sub-01_inplaneT2.nii.gz
[WARNING] Found no aggregated metadata info file /mnt/btrfs/scrap/datalad/openneuro-samples/ds000008/.datalad/metadata/aggregate_v1.json. Found following info files, which might have been generated with newer version(s) of datalad: .datalad/metadata/aggregate_v1.json. You will likely need to either update the dataset from its original location,, upgrade datalad or reaggregate metadata locally. 
[WARNING] Dataset at . contains no aggregated metadata on this path [metadata(/mnt/btrfs/scrap/datalad/openneuro-samples/ds000008/sub-01/anat/sub-01_inplaneT2.nii.gz)] 
{
  "action": "metadata", 
  "path": "/mnt/btrfs/scrap/datalad/openneuro-samples/ds000008/sub-01/anat/sub-01_inplaneT2.nii.gz", 
  "status": "impossible", 
  "type": "file"
}
```

whenever it works fine as soon as I uninstall it
```shell
$> datalad uninstall ds000008 
uninstall(ok): /mnt/btrfs/scrap/datalad/openneuro-samples/ds000008 (dataset)
action summary:
  drop (notneeded: 1)
  uninstall (ok: 1)

$> datalad -f json_pp metadata --reporton all ds000008/sub-01/anat/sub-01_inplaneT2.nii.gz | head -n 10
{
  "action": "metadata", 
  "dsid": "ada6363e-8706-11e8-b2f9-0242ac12001e", 
  "metadata": {
    "@context": {
      "@vocab": "http://docs.datalad.org/schema_v2.0.json"
    }, 
    "annex": {
      "key": "MD5E-s592833--8d37e298f8afe55cfd04a19870399198.nii.gz"
    }, 
...
```

mih:
[...]

PR #3240, opened 2019-03-19
Bf metadata aggregate absent

yarikoptic:
I hoped to find a quick resolution to https://github.com/datalad/datalad/issues/3055 but just got to the conclusion to step away.  The first commit is actually worthwhile on its own, so I will submit it as a separate PR.

In this PR I have tried to iron out a few use cases where metadata might be absent (one log message was misleading a bit), but it boiled again down to the assumption of having a "context dataset" for operation, which in the case of no `-d` specified, is absent.  I've decided to kick it out into public just in case, feel free to close.

### DIFF
--- a/datalad/metadata/tests/test_base.py
+++ b/datalad/metadata/tests/test_base.py
@@ -270,3 +270,34 @@ def test_get_containingds_from_agginfo():
     # will not tollerate mix'n'match
     assert_raises(ValueError, _get_containingds_from_agginfo, {'match': {}}, op.abspath(down))
     assert_raises(ValueError, _get_containingds_from_agginfo, {op.abspath('match'): {}}, down)
+
+
+@with_tempfile(mkdir=True)
+def test_get_metadata_fromparentds_file(path):
+    # https://github.com/datalad/datalad/issues/3055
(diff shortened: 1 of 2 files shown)


datalad/metadata/tests/test_base.py
```diff
@@ -270,3 +270,65 @@ def test_get_containingds_from_agginfo():
...
+    # without causing extraction again in sub1/sub2 which lacks metadata ATM
+    ds.aggregate_metadata(update_mode='target', recursive=True)
+
+    # common kwargs for metadata call
+    mkwargs = dict(return_type='item-or-list')
+    # will be ok since querying using `ds.`, i.e. `-d`
+    out_ds = ds.metadata('sub1/sub2', reporton='datasets', **mkwargs)
+    out_file = ds.metadata('sub1/sub2/somefile', **mkwargs)
+    # assert that no aggregation has happened
+    ok_(not op.exists(op.join(sub2.path, '.datalad', 'metadata')))
```
yarikoptic: this test is insufficient!  
it might still re-extract metadata, and just do not save it in that sub1/sub2.  It should wrap aggregate_metadata with the mock and disallow re-extraction in `sub2`.

mih:

Issue #3230, opened 2019-03-15
Should we disambiguate semantic of having `-d PATH` specification?


yarikoptic:
With revolution upon/within us, I wondered if we could simplify the semantic of having `-d` specification of the dataset in the commands.  At the moment I see it mixing-in at least 2 possible semantics:

### `#1`. How to treat relative paths relative to the top of the specified dataset

The rationale for this is the consistency with Python use of all those as methods of a dataset instance.

<details>
<summary>peculiarity of `-d .`</summary>

Note that `-d` should point to the top directory of the dataset, so `-d .` does not mean "this dataset", and thus all of the following commands fail, even if we prepend relative path with `./` (which I apparently incorrectly) thought we coded for to describe that it is relative to cwd, not top of the dataset:

```shell
(git-annex)hopa:/tmp/superds[master]
$> git status
On branch master
nothing to commit, working tree clean

$> mkdir subdir

$> cd subdir

$> touch 123

$> datalad add -d . 123
[WARNING] path not part of the reference dataset at /tmp/superds/subdir [add(/tmp/superds/subdir/123)] 
add(impossible): 123 (file) [path not part of the reference dataset at /tmp/superds/subdir]

$> datalad add -d . ./123
[WARNING] path not part of the reference dataset at /tmp/superds/subdir [add(/tmp/superds/subdir/123)] 
add(impossible): 123 (file) [path not part of the reference dataset at /tmp/superds/subdir]

$> datalad rev-save -d . ./123   
[ERROR  ] No installed dataset found at /tmp/superds/subdir. [dataset.py:require_dataset:612] (ValueError)
```
</details>

But then many commands also add additional semantic beyond "how to treat relative paths":

### `#2`. either to consider (only new?) paths to be a part of the dataset

So `install -d . URL` adds installed dataset, and `install URL` doesn't. 
Similarly `create` and even `rev-create` (https://github.com/datalad/datalad-revolution/issues/97) would not add new subdataset.
I no longer remember what was (if any) the rationale for not adding installed or created subdatasets (without -d) right away.  Note that with aforementioned `-d .` gotcha it is actually quite inconvenient to add subdatasets in deep subdirectories within subdatasets.

Because of this confusion I didn't even reply right away to @kyleam suggestion for a fix in 
https://github.com/datalad/datalad-container/issues/72#issuecomment-472925070 since I am not sure what effect that would have, e.g. may be then making behavior different in the top directory of the dataset from how it would behave in subdirectory.

[...]

PR #3211, opened 2019-03-08
ENH: prevent users to create a repo in a subdir with files known to git above

yarikoptic:
This pull request should prevent such problematic situations as it was revealed in https://github.com/datalad/datalad/issues/3068#issuecomment-470549719 from occurring. 

This situation was probably also the origin for the situation which inspired https://github.com/datalad/datalad/issues/3139, although in that report I have come up with a minimal use case (empty git subrepo) which might have actually be a different problem.

Merge into master would conflict, and it might (or not) be wiser to just use `get_content_info` which is provided in there instead of a custom call to `ls-files`

PR #3242, opened 2019-03-19
ENH: -d^. to point to the top of the current dataset

yarikoptic:
It is often desired to reference current dataset. E.g. to add an existing
dataset as subdataset to the current dataset somewhere not at the top of the
dataset directories tree. Especially in quite nested hierarchies it
becomes a burden to figure out where is the boundary of the dataset to specify
it in -d option.  -d .  would point to the (not installed, and possibly illegit
see #3211) subdataset in current directory, whenever actual need is to specify
smth like -d ../../..  .

With -d^. it becomes possible to say "refer to the current dataset I am in".

The only gotcha I am aware of is that shell might like to use "^" as
a quick substitution for previous command, that is why I specified it
as -d^., i.e. without space - then seems to work as intended -- running
"datalad add -d^. ./subds"  would add subds in current subdirectory to the
dataset without requiring full path in -d.

Depending on the outcome of #3230 discussion, this change might not be really
needed, e.g. if default context for operations (including on subdatasets) would
migrate to be of "current dataset".  Otherwise, if -d would be needed to "bind"
the context to current dataset - at least with this helper it would be easy to
unambigously reference current dataset regardless of the position in the
directories hierarchy.

If you have a better suggestion instead of `^.` - please suggest.  My choice was because we have `^` to point to supermost dataset, thus use it to point to the (very) top. So it kinda made sense to me to point to the top of dataset curdir belongs to with `^.`, and have it on the left because that target path is to the left of curdir (although `.^` probably would have been better for use in shell due to aforementioned gotcha)

### Instructions

- [x] **Delete these instructions**. :-)

### DIFF
--- a/datalad/distribution/dataset.py
+++ b/datalad/distribution/dataset.py
@@ -111,6 +111,9 @@ class Dataset(object, metaclass=Flyweight):
             # might have its ideas on what to do with ^, so better use as -d^
             path_ = Dataset(get_dataset_root(curdir)).get_superdataset(
                 topmost=True).path
+        elif path == '^.':
+            # get the dataset containing current directory
+            path_ = get_dataset_root(curdir)
         elif path == '///':
             # TODO: logic/UI on installing a default dataset could move here
             # from search?
--- a/datalad/distribution/tests/test_dataset.py
+++ b/datalad/distribution/tests/test_dataset.py
@@ -203,6 +203,23 @@ def test_subdatasets(path):
 
     # TODO actual submodule checkout is still there
 
+    # Test ^. (the dataset for curdir) shortcut
+    # At the top should point to the top
+    with chpwd(ds.path):
+        dstop = Dataset('^.')
+        eq_(dstop, ds)
+
+    # and still does within subdir
+    os.mkdir(opj(ds.path, 'subdir'))
+    with chpwd(opj(ds.path, 'subdir')):
+        dstop = Dataset('^.')
+        eq_(dstop, ds)
+
+    # within submodule will point to submodule
+    with chpwd(subsubds.path):
+        dstop = Dataset('^.')
+        eq_(dstop, subsubds)
+
 
 @with_tempfile(mkdir=True)
 def test_hat_dataset_more(path):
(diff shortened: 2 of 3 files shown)


mih:
FTR: I think we should be reluctant to add general meaning to argument values that are only needed to shoehorn particular behavior into the existing machinery. This issue is a flavor of "the shell is to dumb". In Python code this should be rather unnecessary (one would use proper Dataset instances). Maybe it is worth implementing such argument transformation in the initial cmdline API layer, and not further complicate path interpretation globally.

Given that the example uses `add`, I would prefer to discuss this in the context of `rev-save` which doesn't have many of add's issues and also consider my proposal in https://github.com/datalad/datalad/issues/3230#issuecomment-473661218 for any command to stop behaving differently with or without `-d` given -- which would make the entire issue go away, without adding more decision making logic.

yarikoptic:
> Maybe it is worth implementing such argument transformation in the initial cmdline API layer

It could be done there but I do not see a real need to break concurrency between python and cmdline APIs for this.
Re #3230 - indeed, if we make commands behave similar with and without `-d` such shortcut might not be needed (as I noted in the PR description), so I am ok for this PR to linger around until we have a clear(er) picture.

kyleam:
It makes sense to see how things settle, but 

> > Maybe it is worth implementing such argument transformation in the initial cmdline API layer
> 
> It could be done there but I do not see a real need to break concurrency between python and cmdline APIs for this.

... from the standpoint of someone inspecting the code, I like that in this patch `^.` is handled in the same location as `///`.

driusan:
I think the usage of "^" is likely to cause some confusion, since it already has meaning to git and the usage here is context sensitive. With this change places referring to datasets it means one thing, while places specifying git revisions it means something completely unrelated.

mih:
Ok, I am with @kyleam now. Good to go.

It would be very good, though, to have these symbols and behavior documented somewhere. Thanks in advance!

mih:
What shall we do with this one?

mih:
Ping!

mih:
Shall I change the base and merge this one into master?

yarikoptic:
Without agreeing on smth like 1. of https://github.com/datalad/datalad/issues/3905 I guess we should indeed have something like this. I have rebased and added minimal doc mentioning.  Note: Since `--dataset` is command specific, duplicated across all the commands, we do not mention `///` or `^` anywere "in the API docs", so the same remains for `^.`

kyleam:
@yarikoptic Thanks for the updates.

PR #3243, opened 2019-03-19
I am not strong enough for this...

mih:
A release is the PR to prepare it, and the PR to fix the tests that are
automatically broken in any PR by any release until a second PR comes
along to fix the master branch. Fixing means pasting in a snippet that
gets deleted later, and to envision not only the next version, but also
 the one that maybe follows it. And a machine will then check, if you guessed
 correctly -- after a few hours of deliberation.... As a human, I feel raped.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,21 @@ This is a high level and scarce summary of the changes between releases.
 We would recommend to consult log of the 
 [DataLad git repository](http://github.com/datalad/datalad) for more details.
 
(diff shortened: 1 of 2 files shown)


yarikoptic:
There is a soviet times anecdote I will tell when I see you.
I usually introduce the change you did here in a PR via a direct push to master with `[ci: skip]` to save some cycles.
I am ok with whatever would make your life easier, I just will blame you whenever I again forget to tag or boost the version.

PR #3247, opened 2019-03-19
RF: Suck in and use  RF'ed helper function from -revolution

mih:
This is used also in the RF metadata code.

This is to be filled up with tests demonstrating the status quo re -revolution commands and https://github.com/datalad/datalad/issues/3230

### DIFF
--- a/datalad/core/local/status.py
+++ b/datalad/core/local/status.py
@@ -13,7 +13,6 @@ __docformat__ = 'restructuredtext'
 
 import logging
 import os
-import os.path as op
 from six import (
     iteritems,
     text_type,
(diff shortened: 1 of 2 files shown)


yarikoptic:
note that Draft PRs are not tested by Travis (https://travis-ci.community/t/draft-pull-requests-not-being-built/2434/9)

mih:
I will close this for now. The current implementation of this function is at https://github.com/datalad/datalad-metalad/blob/df3f28d234107e9f1d4196de572d7be3f2643b0b/datalad_metalad/utils.py#L18

A number of issues have been fixed in the meantime, but it certainly won't hurt to prolong the maturation period a bit more.

PR #3249, opened 2019-03-19
BF: do not cause a side-effect on kwargs in @eval_results

yarikoptic:
Ran into it while creating a test where multiple kwargs should have been
consistently passed to subsequent invocations of metadata (as a function or
a method).  But pre-created kwargs dictionary was augmented by calling
metadata, which (as any side-effect) was quite confusing.


yarikoptic:
good, no code seemed to rely on the side-effect, thus merging. Thanks @kyleam  for review

PR #3250, opened 2019-03-19
Rename old constant, and define/use a few new ones to avoid hardcoded copies of the paths

yarikoptic:
Rationale is to make them defined in a single place and be able to quickly (git grep) identify code which relies on them.  Otherwise it is quite hard since string quotes and spacing (separate lines) could vary etc.  Moreover e.g. DATASET_METADATA_FILE I believe was defined separately in multiple points.
    
I have not touched the tests somewhat on purpose, e.g. to make consts somewhat more constant ;)

### DIFF
--- a/datalad/auto.py
+++ b/datalad/auto.py
@@ -30,7 +30,7 @@ from .utils import getpwd
 from .dochelpers import exc_str
 from .support.annexrepo import AnnexRepo
 from .cmdline.helpers import get_repo_instance
-from .consts import HANDLE_META_DIR
+from .consts import DATALAD_DOTDIR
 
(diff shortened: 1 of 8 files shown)


mih:
Wouldn't it make sense to rename `HANDLE_META_DIR` as part of this change?

Everyone who knew what a handle was has a beard now. And the `META` part of misleading, if not factually wrong.

yarikoptic:
;-) I thought the same. 
I would retain a "deprecation" alias in consts for a bit but internally we should indeed rename/use a new one.  `DATALAD_DOTDIR` ?

PR #3255, opened 2019-03-21
Add Python information to "datalad wtf" command

driusan:
datalad wtf did not include any information about which
version of Python is being used. This adds a section to
include the python version number and python implementation
being used to run datalad to aid in debugging.

### DIFF
--- a/datalad/plugin/wtf.py
+++ b/datalad/plugin/wtf.py
@@ -128,7 +128,7 @@ def _describe_environment():
 def _describe_python():
     import platform
     return {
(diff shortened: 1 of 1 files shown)


yarikoptic:
Who said "elephant"? ;-) Thanks @driusan . FWIW, here is what I see
```
## python 
  - implementation: CPython
  - python: 2.7.14+
```
may be change "python" item to "version"? ;)

driusan:
Sure, I changed the key to "version"

PR #3260, opened 2019-03-22
RF: reincarnate benchmarking of non-rev create/add.

yarikoptic:
rev- versions will now be benchmarked too if available, but by a separate dedicated benchmark.
On revisions where there is no `rev_` commands (0.11.x) we will still be able to benchmark


yarikoptic:
Since probably noone but me cares about benchmarks, and I have one more (more interesting) PR coming, merging

PR #3263, opened 2019-03-23
BM: new benchmark based on make_studyforrest_mockup

yarikoptic:
It is also possible to profile and produce profile figure right away with

    pip install gprof2dot
    asv profile -v -o profile.gprof usecases.study_forrest.time_make_studyforrest_mockup
    gprof2dot -f pstats profile.gprof | dot -Tpng -o profile.png && xdg-open profile.png

Here is the results of running this benchmark on my elderly laptop on 0.11.x and here:
```
$> asv compare -m hopa 0.11.x HEAD        

All benchmarks:

     [930f464b]       [2ec70199]
     <0.11.x>         <bm-forrest>
...
        50.3±0.3s        43.7±0.4s    ~0.87  usecases.study_forrest.time_make_studyforrest_mockup [hopa/virtualenv-py2.7]
          49.1±1s        43.0±0.2s    ~0.88  usecases.study_forrest.time_make_studyforrest_mockup [hopa/virtualenv-py3.4]

### DIFF
--- a/benchmarks/usecases.py
+++ b/benchmarks/usecases.py
@@ -31,15 +31,11 @@ class study_forrest(SuprocBenchmarks):
     timeout = 180  # especially with profiling might take longer than default 60s
 
     def setup(self):
-        sys.stderr.write("IN SETUP!")
         self.path = tempfile.mkdtemp(**get_tempfile_kwargs({}, prefix='bm_forrest'))
 
     def teardown(self):
-        sys.stderr.write("IN TEARDOWN!")
         if osp.exists(self.path):
-            sys.stderr.write("REMOVING")
             rmtree(self.path)
-            sys.stderr.write("REMOVED")
 
     def time_make_studyforrest_mockup(self):
         path = self.path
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: benchmarks/usecases.py [hunk 1]
```diff
@@ -0,0 +1,102 @@
...
+    def setup(self):
+        sys.stderr.write("IN SETUP!")
+        self.path = tempfile.mkdtemp(**get_tempfile_kwargs({}, prefix='bm_forrest'))
+
+    def teardown(self):
+        sys.stderr.write("IN TEARDOWN!")
+        if osp.exists(self.path):
+            sys.stderr.write("REMOVING")
+            rmtree(self.path)
+            sys.stderr.write("REMOVED")
```
yarikoptic: oops, forgot that I was troubleshooting those -- will first make sure that it actually works, something wasn't entirely "kosher" IIRC

yarikoptic: removed

mih:
Cool!

PR #3268, opened 2019-03-24
BF: check for stdin/out/err to not be closed before checking for .isatty

### LINKED ISSUE #3267: Need to guard is_interactive better
While running benchmarks I noted that `time_ls*` ones were failing.
Running
```shell
asv run -j 4 --machine drogon-neurodebian10.20190322.1 -q -e -v -b api.supers.time_ls
```
helped to identify the reason:
```
                 File "/home/yoh/proj/datalad/datalad/venvs/benchmark-10.20190322.1/local/lib/python2.7/site-packages/asv/benchmark.py", line 569, in benchmark_timing
                   timing = timer.timeit(number)
                 File "/usr/lib/python2.7/timeit.py", line 202, in timeit
                   timing = self.inner(it, self.timer)
                 File "/usr/lib/python2.7/timeit.py", line 100, in inner
                   _func()
                 File "/home/yoh/proj/datalad/datalad/venvs/benchmark-10.20190322.1/local/lib/python2.7/site-packages/asv/benchmark.py", line 474, in <lambda>
                   func = lambda: self.func(*param)
                 File "/home/yoh/proj/datalad/datalad/benchmarks/api.py", line 126, in time_ls
                   ls(self.ds.path)
                 File "/home/yoh/proj/datalad/datalad/.asv/env/1b5862776063636a1bd6a1931ef20cce/local/lib/python2.7/site-packages/datalad/interface/ls.py", line 149, in __call__
                   return _ls_json(loc, json=json, **kw) if json else _ls_dataset(loc, **kw)
                 File "/home/yoh/proj/datalad/datalad/.asv/env/1b5862776063636a1bd6a1931ef20cce/local/lib/python2.7/site-packages/datalad/interface/ls.py", line 484, in _ls_dataset
                   path_fmt = u"{ds.path!U:<%d}" % (maxpath + (11 if is_interactive() else 0))  # + to accommodate ansi codes
                 File "/home/yoh/proj/datalad/datalad/.asv/env/1b5862776063636a1bd6a1931ef20cce/local/lib/python2.7/site-packages/datalad/utils.py", line 210, in is_interactive
                   return sys.stdin.isatty() and sys.stdout.isatty() and sys.stderr.isatty()
               ValueError: I/O operation on closed file
```
according to http://datalad.github.io/datalad/#api.supers.time_ls , the last commit which was successfully benchmarked was https://github.com/datalad/datalad/commit/3bcc58fa in Aug 8 2018... but that doesn't give me an immediate idea on what has happened since code for the is_interactive seemed to have no changes since then.

Will check in more detail later

yarikoptic:
Closes #3267 

TODOs:
- [x] test
- [x] try to figure out what change in our behavior lead to the benchmark to start failing after Aug 8 2018. See #3267

### DIFF
--- a/datalad/utils.py
+++ b/datalad/utils.py
@@ -203,15 +203,22 @@ def auto_repr(cls):
     return cls
 
 
+def _is_stream_tty(stream):
+    try:
+        # TODO: check on windows if hasattr check would work correctly and
+        # add value:
+        return stream.isatty()
+    except ValueError as exc:
+        # Who knows why it is a ValueError, but let's try to be specific
+        # If there is a problem with I/O - non-interactive, otherwise reraise
+        if "I/O" in str(exc):
+            return False
+        raise
+
+
 def is_interactive():
     """Return True if all in/outs are tty"""
-    # TODO: check on windows if hasattr check would work correctly and add value:
-    #
-    return (
-        (not sys.stdin.closed and sys.stdin.isatty()) and
-        (not sys.stdout.closed and sys.stdout.isatty()) and
-        (not sys.stderr.closed and sys.stderr.isatty())
-    )
+    return all(_is_stream_tty(s) for s in (sys.stdin, sys.stdout, sys.stderr))
 
 
 def get_ipython_shell():
--- a/datalad/utils.py
+++ b/datalad/utils.py
@@ -217,7 +217,13 @@ def _is_stream_tty(stream):
 
 
 def is_interactive():
-    """Return True if all in/outs are tty"""
+    """Return True if all in/outs are open and tty.
+
+    Note that in a somewhat abnormal case where e.g. stdin is explicitly
+    closed, and any operation on it would raise a
+    `ValueError("I/O operation on closed file")` exception, this function
+    would just return False, since the session cannot be used interactively.
+    """
     return all(_is_stream_tty(s) for s in (sys.stdin, sys.stdout, sys.stderr))
 
 
--- a/datalad/utils.py
+++ b/datalad/utils.py
@@ -207,7 +207,11 @@ def is_interactive():
     """Return True if all in/outs are tty"""
     # TODO: check on windows if hasattr check would work correctly and add value:
     #
-    return sys.stdin.isatty() and sys.stdout.isatty() and sys.stderr.isatty()
+    return (
+        (not sys.stdin.closed and sys.stdin.isatty()) and
+        (not sys.stdout.closed and sys.stdout.isatty()) and
+        (not sys.stderr.closed and sys.stderr.isatty())
+    )
 
 
 def get_ipython_shell():
--- a/datalad/tests/test_utils.py
+++ b/datalad/tests/test_utils.py
@@ -34,6 +34,7 @@ from ..utils import getpwd, chpwd
 from ..utils import get_path_prefix
 from ..utils import auto_repr
 from ..utils import find_files
+from ..utils import is_interactive
 from ..utils import line_profile
 from ..utils import not_supported_on_windows
 from ..utils import file_basename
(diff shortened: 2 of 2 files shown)


datalad/utils.py
```diff
@@ -207,7 +207,11 @@ def is_interactive():
     """Return True if all in/outs are tty"""
     # TODO: check on windows if hasattr check would work correctly and add value:
     #
-    return sys.stdin.isatty() and sys.stdout.isatty() and sys.stderr.isatty()
+    return (
+        (not sys.stdin.closed and sys.stdin.isatty()) and
+        (not sys.stdout.closed and sys.stdout.isatty()) and
+        (not sys.stderr.closed and sys.stderr.isatty())
```
yarikoptic: oh hoh, `.closed` itself pukes if stdout is closed
```shell
$> python2 /tmp/1.py 
  File "/tmp/1.py", line 3, in <module>
    print(sys.stdout.closed)
ValueError: I/O operation on closed file

$> python3 /tmp/1.py 
  File "/tmp/1.py", line 3, in <module>
    print(sys.stdout.closed)
ValueError: I/O operation on closed file.

$> cat /tmp/1.py    
import sys; 
sys.stdout.close(); 
print(sys.stdout.closed)
```
so I think there is no way around try/except :-/

yarikoptic:
legit concern, but 
- we do not close streams, so I do not think we should guard downstream "too much".  If anything relevant is closed, I expect code to crash at the appropriate point.
- how probable is the situation? ATM we have only a single use-case - running benchmarks under asv.  Note that no other errors emerge AFAIK -- just for `is_interactive`, which gets fixed here (and there would be follow up PR as well to monkey patch benchmarks setup)

PS just mentioned that I managed to submit it against master, although I think it is well worth 0.11.x. will rebase

kyleam:
> * we do not close streams, so I do not think we should guard downstream "too much".  If anything relevant is closed, I expect code to crash at the appropriate point.

Agreed, my suggestion wasn't that we guard downstream.  I think we should assume that the streams are open and let an error be raised---including in `is_interactive()`.  This patch is making `is_interactive` report "false" when the answer is really "neither/invalid question" (which the exception describes nicely).

>  * how probable is the situation? ATM we have only a single use-case - running benchmarks under asv.  Note that no other errors emerge AFAIK -- just for `is_interactive`, which gets fixed here (and there would be follow up PR as well to monkey patch benchmarks setup)

I'd think not likely at all.  Again, I'd say we should assume that std{in,out,err} are open.  Which makes me think that this should be fix this upstream (why does asv result in stdin being closed?)---or at least make a more local, well-documented kludge outside of `is_interactive()`.

But I won't push back more than that because I don't think it matters all that much in practice.

yarikoptic:
> Which makes me think that this should be fixed upstream (why does asv result in stdin being closed?)---or at least make a more local, well-documented kludge outside of `is_interactive()`.

yeah, I was on a fence to either do the archaeological study on when it was introduced to asv... my argumentation in support of "it is ok" was that I think it should not be guaranteed that `stdin` is open.   My wild guess is that it was intentional on their end to guarantee that no input should somehow be expected. That is why I also think thought it is ok for `is_interactive` to just return `False` in such a case.  Hence

> This patch is making `is_interactive` report "false" when the answer is really "neither/invalid question" (which the exception describes nicely).

could be argued or clarified in documentation - IMHO the main purpose is to detect when environment is interactive and thus all streams are open and tty.  If something is not satisfying that -- it is just `False`. I do not see any advantage for `is_interactive` throwing an exception when environment clearly is not interactive.  
While at it I even wondered if we shouldn't check if `stderr` is a `tty`. I think it should be ok to redirect stderr to a file while stdin/out interacting with the user... but I guess it is a separate issue/PR

Overall:
- let's check with upstream: https://github.com/airspeed-velocity/asv/issues/802
- I will push a slight tune up to `is_interactive` docstring describing its "resilience".

yarikoptic:
appveyor fail - unrelated https://github.com/datalad/datalad/issues/3287
travis -- probably due to my data transfer from the server its connection was more flaky. restarted those

kyleam:
> appveyor fail - unrelated #3287

The error from 3287 doesn't appear in [the failure here][0].  The [only failure][1] I spot in that run is

```
Invoke-WebRequest : 503 Service Temporarily Unavailable
nginx/1.13.12
[...]
```

Looks unrelated in any case.

[0]: https://ci.appveyor.com/project/mih/datalad/builds/23549086/job/6m6jpt8pi637061i
[1]: https://ci.appveyor.com/project/mih/datalad/builds/23549086/job/6m6jpt8pi637061i#L1997

PR #3271, opened 2019-03-26
Make local doc builds less painful

mih:
Now that docs are being printed, the should become a lot better, a lot faster.


mih:
@kyleam This makes it possible to build the docs with a system installed sphinx, but datalad being installed in a dev-virtualenv only. It should not get in the way in other scenarios, and readthedocs is checked to be OK with it.

CHANGELOG.md
```diff
@@ -29,7 +29,7 @@ bet we will fix some bugs and make a world even a better place.
 
 ### Fixes
 
-- `GitRepo.dirty` does not report on nested empty directories ([#3196]).
+- `GitRepo.dirty` does not report on nested empty directories ([#3196][]).
```
yarikoptic: just so I know - why is it (explicit empty `[]`) needed?

We would need to have `tools/link_issues_CHANGELOG` adjusted for it as well then I believe

yarikoptic: commit message unfortunately is not informative.  AFAIK it was a valid markdown, e.g. according to examples at https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#links

mih: It may be valid, but then it is not supported by pandoc. On conversion to RST all that gets stripped.

mih:
Good to go, I think.

yarikoptic:
AFAIK it was not good to go since `tools/link_issues_CHANGELOG` was not tuned up for the changes.  Sending a PR

PR #3274, opened 2019-03-27
RF: Move the availability check for git-annex all the way to the end

mih:
Rational: We otherwise refuse to do anything with repos that could have an annex, but for operations like `status` this is not needed. Doing a late check will enable (more) datalad use on systems that do not have
git-annex.

@effigies wondered if it was possible to get information on a dataset without git-annex. This would be step one. It might be useful in this context to replace `AnnexRepo.get_content_annexinfo()` to not call `git-annex find/findref`, but to inspect the keys in a dataset on its own. This would make `status --annex <mode>` work without git-annex for any query mode where we would want to support this. It would also be straightforward to support a fallback that is only used when git-annex is actually not around.


effigies:
🎉

PR #3296, opened 2019-04-05
BF: for compatibility with elderly git 2.11.0 check for known paths from parent directory

### LINKED ISSUE #3295: older git (2.11.0) ls-files does not list cwd which belongs subdataset
While testing git-annex build on jessie (for maximum portability across kernels) using "stack" distribution of ghc and libraries, ran into our safe-guard test failing:
```
FAIL: datalad.distribution.tests.test_dataset.test_is_installed
    self.test(*self.arg)
  File "/home/yoh/proj/datalad/datalad/datalad/tests/utils.py", line 791, in newfunc
    t(*(arg + (uri,)), **kw)
  File "/home/yoh/proj/datalad/datalad/datalad/tests/utils.py", line 615, in newfunc
  File "/home/yoh/proj/datalad/datalad/datalad/distribution/tests/test_dataset.py", line 111, in test_is_installed
    subds.create()
AssertionError: PathKnownToRepositoryError not raised
```

which is due to
```
(git-annex)hopa:~/.tmp/datalad_temp_test_is_installedd9cRgg[master]subm 1
$> /usr/lib/git-annex.linux/git ls-files       
$> /usr/lib/git-annex.linux/git --version
git version 2.11.0
````
as compared to
```shell
$> git ls-files                                
./
$> git --version
git version 2.20.1
```
so it seems we need to check also for minimal git version compatibility :-/  

master version (0.12.0rc2-128-gf9ff64ea) seems to be ok on that test, but now I will run the full battery to see if any other effect

yarikoptic:
(if such exists and is a different one)

Closes #3295

### DIFF
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -729,13 +729,17 @@ class GitRepo(RepoInterface):
         self._fake_dates_enabled = None
 
     def _create_empty_repo(self, path, **kwargs):
-        if op.lexists(path):
+        # We need to check from parent directory for elderly Git (e.g. 2.11.0)
+        # https://github.com/datalad/datalad/issues/3295
+        path = op.abspath(path)
+        parpath = op.dirname(path)
+        if op.lexists(path) and op.lexists(parpath) and parpath != path:
             # Verify that we are not trying to initialize a new git repository
             # under a directory some files of which are already tracked by git
             # use case: https://github.com/datalad/datalad/issues/3068
             try:
                 stdout, _ = self._git_custom_command(
-                    None, ['git', 'ls-files'], cwd=path, expect_fail=True
+                    [path], ['git', 'ls-files'], cwd=parpath, expect_fail=True
                 )
                 # The 2nd check to verify that the files exctually exist is for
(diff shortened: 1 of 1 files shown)


kyleam:
From my POV the added complication isn't worth it.  If an older git is found, we should just leave a debug message and skip the safety check.  (I'd by happy to track down the specific git version that introduced the ls-files change, if you want to go this direction.)

---

There are a lot of test failures.  At least one source of failure appears to be that the `path` argument to the git call can get converted into "." in `normalize_paths`:

https://github.com/datalad/datalad/blob/7728939f2ac04aaf23d13d1ff91d86fa71931939/datalad/support/gitrepo.py#L289

kyleam:
FWIW this ls-files change came with v2.14.0, specifically cbca060e10
(ls-files: prevent prune_cache from overeagerly pruning submodules,
2017-05-11).

yarikoptic:
ok, what about https://github.com/datalad/datalad/pull/3347 ?

kyleam:
> ok, what about #3347 ?

LGTM

PR #3298, opened 2019-04-07
RF: Make AnnexRepo.set_metadata() work without iteration

mih:
Analog to function like add/add_ the normal version now just does things, whereas the _ version returns a generator that needs to be iterator over to actually perform anything. The prior behavior was
a constant source of surprise and anger.

Related to #2755


mih:
I suspect this is non-controversial.

PR #3299, opened 2019-04-07
ENH: Report 'bytesize' from plain Git, whenever ls-tree runs anyways

mih:
It is cheap and useful.

### DIFF
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -2872,10 +2872,12 @@ class GitRepo(RepoInterface):
             else:
                 raise ValueError(
                     'unknown value for `untracked`: %s', untracked)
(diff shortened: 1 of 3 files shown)


datalad/support/gitrepo.py
```diff
@@ -2930,6 +2930,8 @@ def get_content_info(self, paths=None, ref=None, untracked='all'):
                     # symlink-nature is a technicality that is dependent
                     # on the particular mode annex is in
                     inf['type'] = 'file'
+                if ref and inf['type'] == 'file':
+                    inf['bytesize'] = int(props.group(4))
```
yarikoptic: Why not to use named groups in regular expression itself?

mih: Good point! RF'ed. Looks much nicer in code, indeed. Thx.

Issue #3300, opened 2019-04-08
GitRepo.get_content_info() performance issue


mih:
In https://github.com/datalad/datalad/blob/cc3a9e6dd56eaeb0f1c9c14dc1a3bc3d453fe881/datalad/support/gitrepo.py#L2918 we homogenize the report of `symlinks` in a dataset repository to `type='file'` if we can determine that that symlink is an annex pointer.

It turns out that this has a real performance hit. In a repo with 33k files doing this increases the runtime of a `status()` from 4s to 12s.

However, simply disabling this will impact the convenience-level felt in high-level code, as this was done to homogenize cross-platform/mode feel when dealing with annex'ed files.

Using pathlib would immediately gain 2s speedup (but see comment in code re PY3.5).

Switch to `os.readlink()` based test seems to be the obvious improvement: runtime ~5s

PR #3301, opened 2019-04-08
ENH: Speed-up GitRepo.get_content_info() (fixes gh-3300)

mih:
#3300

### DIFF
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -2920,19 +2920,9 @@ class GitRepo(RepoInterface):
                 if inf['type'] == 'symlink' and \
                         '.git/annex/objects' in \
                         ut.Path(
-                            op.realpath(op.join(
-                                # this is unicode
-                                self.path,
-                                # this has to become unicode on older Pythons
-                                # it doesn't only look ugly, it is ugly
-                                # and probably wrong
-                                unicode(str(path), 'utf-8')
-                                if PY2 else str(path)))).as_posix():
-                    # ugly thing above could be just
-                    #  (self.pathobj / path).resolve().as_posix()
-                    # but PY3.5 does not support resolve(strict=False)
-
-                    # report locked annexed files as file, their
+                            os.readlink(text_type(self.pathobj / path))
+                        ).as_posix():
+                    # report annex symlink pointers as file, their
                     # symlink-nature is a technicality that is dependent
                     # on the particular mode annex is in
                     inf['type'] = 'file'
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -2879,6 +2879,7 @@ class GitRepo(RepoInterface):
             props_re = re.compile(
                 r'(?P<type>[0-9]+) ([a-z]*) (?P<sha>[^ ]*) [\s]*(?P<size>[0-9-]+)\t(?P<fname>.*)$')
 
+        lgr.debug('Query repo: %s', cmd)
         try:
             stdout, stderr = self._git_custom_command(
                 # specifically always ask for a full report and
@@ -2899,6 +2900,7 @@ class GitRepo(RepoInterface):
             if "fatal: Not a valid object name" in str(exc):
                 raise ValueError("Git reference '{}' invalid".format(ref))
             raise
+        lgr.debug('Done query repo: %s', cmd)
 
         for line in stdout.split('\0'):
             if not line:
@@ -2963,6 +2965,7 @@ class GitRepo(RepoInterface):
                     else 'directory' if path.is_dir() else 'file'
             info[path] = inf
 
+        lgr.debug('Done %s.get_content_info(...)', self)
         return info
 
     def status(self, paths=None, untracked='all', ignore_submodules='no'):


yarikoptic:
Note to myself/other: this PR also has RFing of .support.annexrepo.BatchedAnnex into a more generic cmd.BatchedAnnex

mih:
> Note to myself/other: this PR also has RFing of .support.annexrepo.BatchedAnnex into a more generic cmd.BatchedAnnex

Note that this RF takes out the notion of "annex". However, travis seems unhappy with it (stalled tests). SO which one is the other PR that may have done this properly already?

yarikoptic:
I don't think we had any PR for generic BatchCommand.  We did rely though on some batched runs of git done by GitPython.

mih:
Damn, I thought there is an easy way out.... :(

yarikoptic:
well, may be there is - don't they stall for you locally? seems to do for me:
```shell
(git)hopa:~datalad/datalad[enh-statusperf]git
$> python -m nose -v -s datalad/support/tests/test_*repo.py 
datalad.support.tests.test_annexrepo.test_AnnexRepo_addurl_batched_and_set_metadata ... ok
datalad.support.tests.test_annexrepo.test_AnnexRepo_addurl_to_file_batched ... 

```
so it is just the matter to figure out what you have changed ;-)

yarikoptic:
FWIW here is how it hangs
```
[DEBUG  ] Initiating a new process for BatchedCommand(cmd=['git', 'cat-file', '--batch'], output_proc=<function>, path=None) 
[Level 5] Command: ['git', 'cat-file', '--batch'] 
[Level 5] Sending 'HEAD:about.txt\n' to batched command BatchedCommand(cmd=['git', 'cat-file', '--batch'], output_proc=<function>, path=None) 
[Level 5] Done sending. 

```

yarikoptic:
to be exact on 
```
(Pdb) p entry
'HEAD:about.txt\n'
(Pdb) l
790  	            process = self._process  # _check_process might have restarted it
791  	            process.stdin.write(assure_bytes(entry) if PY2 else entry)
792  	            process.stdin.flush()
793  	            lgr.log(5, "Done sending.")
794  	            import pdb; pdb.set_trace()
795  ->	            still_alive, stderr = self._check_process(restart=False)
796  	            # TODO: we might want to handle still_alive, e.g. to allow for
797  	            #       a number of restarts/resends, but it should be per command
798  	            #       since for some we cannot just resend the same query. But if
799  	            #       it is just a "get"er - we could resend it few times
800  	            # We are expecting a single line output
(Pdb) n
> /home/yoh/proj/datalad/datalad/datalad/cmd.py(803)__call__()
-> if not process.stdout.closed else None
(Pdb) n
```
so something in how `git .. --batch` works probably differs from how annex does

mih:
Thx much! I think I know what is happening.

datalad/cmd.py
```diff
@@ -692,3 +693,151 @@ def run(self, cmd, env=None, *args, **kwargs):
...
+
+
+def readline_rstripped(stdout):
+    #return iter(stdout.readline, b'').next().rstrip()
+    return stdout.readline().rstrip()
+
+
+@auto_repr
+class BatchedCommand(object):
+    """Container for an annex process which would allow for persistent communication
```
kyleam: This copied over code has some references to "annex".  I'll push a clean-up commit after the current test run completes.

datalad/cmd.py
```diff
@@ -692,3 +693,151 @@ def run(self, cmd, env=None, *args, **kwargs):
...
+            process = self._process  # _check_process might have restarted it
+            process.stdin.write(assure_bytes(entry) if PY2 else entry)
+            process.stdin.flush()
+            lgr.log(5, "Done sending.")
+            still_alive, stderr = self._check_process(restart=False)
+            # TODO: we might want to handle still_alive, e.g. to allow for
+            #       a number of restarts/resends, but it should be per command
+            #       since for some we cannot just resend the same query. But if
+            #       it is just a "get"er - we could resend it few times
+            # We are expecting a single line output
```
kyleam: I think this was true for annex batch commands, but it's not true for batch commands in general.  I'll push a clean-up commit that clarifies that the default `output_proc` assumes a single line of output.

datalad/support/gitrepo.py
```diff
@@ -2899,6 +2898,25 @@ def get_content_info(self, paths=None, ref=None, untracked='all'):
...
+
+        if ref:
+            def _read_symlink_target_from_catfile(lines):
+                # it is always the second line, all checks done upfront
+                header = lines.readline()
+                if header.rstrip().endswith('missing'):
+                    # something we do not know about, should not happen
+                    # in real use, but guard against to avoid stalling
+                    return ''
+                return lines.readline().rstrip()
```
kyleam: This function is being called with only with targets where `inf['type'] == 'symlink'`, so we know that we can expect two lines (unlike the general cat-file --batch output).  OK.

datalad/cmd.py
mih: Thx.

datalad/cmd.py
mih: Cool!

datalad/support/gitrepo.py
mih: Yes, correct.

mih:
Please feel free to take this PR over. Tests now run and reveal a bunch of uncovered cases. If you have any idea what they are and how to fix it, please plow ahead!

kyleam:
> If you have any idea what they are

Aside from the encoding issues, the failing tests seem to come down to the faster `readlink` approach being stricter than the previous `realpath` approach.  Our ls-files call is reporting that there is a symlink, but if it changed in the working tree (either deleted or a regular, unlocked file), `readlink` errors.

> and how to fix it

Hmm, I'm not coming up with anything that wouldn't just put us close to where we started with `realpath`.

mih:
> Hmm, I'm not coming up with anything that wouldn't just put us close to where we started with `realpath`.

Thx for the assessment. Will think about it and see what can be done.

kyleam:
> > Hmm, I'm not coming up with anything that wouldn't just put us close to where we started with `realpath`.
> 
> Thx for the assessment. Will think about it and see what can be done.

I've adjusted the code to call `readlink` and, if it errors, fall back to `realpath`.

datalad/support/gitrepo.py
```diff
@@ -2963,6 +2991,11 @@ def get_content_info(self, paths=None, ref=None, untracked='all'):
                     else 'directory' if path.is_dir() else 'file'
             info[path] = inf
 
+        if ref:
+            # cancel batch process
+            _get_link_target.close()
```
yarikoptic: Shouldn't be in the finally of try to encompass it? An alternative would be to keep those batched processes known to the instance to be reused across calls. Then have them closed in precommit. That is what AnnexRepo is doing at least

mih:
> I've adjusted the code to call readlink and, if it errors, fall back to realpath
Thx. This seems to be the optimal choice.

mih:
@kyleam Addressed @yarikoptic advise. Test are running.

datalad/support/gitrepo.py
mih: try/finally is done. We could think about tracking these processes once we have at least one more use case.

Issue #3303, opened 2019-04-08
`rev-save()` should record dataset.id of add subdatasets in .gitmodules


mih:
This (or something equivalent) would be necessary to be able to report `hasPart` metadata properties with proper identifiers. Otherwise we would depend on subdatasets being present upon metadata extraction, which can be a big thing to ask for.

PR #3304, opened 2019-04-08
ENH: Record a dataset's ID in a superdataset on `add/save` (fixes gh-3303)

mih:
Otherwise we have incomplete identity information on all dataset components (i.e. we only have the state of a subdataset, but not its persistent ID). This is an issue during metadata extraction, where we would require subdatasets to be around in order to be able to give a full report.

We already ensure that create(force=True) does not alter a dataset's ID, so getting out of sync would require manual fiddling with DataLad's internal config. Sure be safe...

### DIFF
--- a/datalad/core/local/tests/test_create.py
+++ b/datalad/core/local/tests/test_create.py
@@ -164,7 +164,7 @@ def test_create_sub(path):
     ds.rev_create()
 
     # 1. create sub and add to super:
-    subds = ds.rev_create(op.join("some", "what", "deeper")
+    subds = ds.rev_create(op.join("some", "what", "deeper"))
     ok_(isinstance(subds, Dataset))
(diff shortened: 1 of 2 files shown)


yarikoptic:
Should be ok... but the code which starts relying on it (probably should later add check to install/clone) should assure to not crash if there is no ID recorded.

mih:
Well, tests are going nuts in a few places. Apparently `datalad/interface/tests/test_ls_webui.py` tests on scenarios where `.gitmodules` is an annex file. Not yet sure whether this makes any sense or why it is happening.

mih:
I think I should move this into `rev_save()`. It should still catch all relevant cases, but not run counter to more simple-minded command assumptions

mih:
Finally!

PR #3310, opened 2019-04-09
BMs: proper cleaning, extended a list of benchmarks

yarikoptic:
- [x] One of the problems with benchmarks was them filling up the TMPDIR, this should be fixed now so we have a chance to complete their full run.

- [x] This also includes #3268 style monkey patching of existing code so we could run under fresh asv, at least until there is some closure with https://github.com/airspeed-velocity/asv/issues/802 on asv closing streams.

- [x] Added more of rudimentary benchmarks based on that sample layout of repositories I was using.

This adjusted setup is now running on drogon, hopefully we will see some results in a few days


kyleam:
I don't know much about asv, but this sounds good.  Thanks!

mih:
Likewise, thx.

yarikoptic:
ok, no objections, so merging as is for now ;-)

PR #3319, opened 2019-04-12
ENH(BM)+OPT: benchmarks to measure overhead from our Runner(s), no sleeping in .poll loop

yarikoptic:
Those track_overhead report percents over the duration via a simple
os.system call where all outputs go to /dev/null.  Here is a result from running
on my laptop:

    $> DATALAD_LOG_LEVEL=2 asv run -j 4 -b core.runner
    · Creating environments
    · Discovering benchmarks
    ·· Uninstalling from virtualenv-py2.7
    ·· Installing 6f9ac93e <master> into virtualenv-py2.7.
    · Running 28 total benchmarks (2 commits * 2 environments * 7 benchmarks)
    [  0.00%] · For DataLad commit 6f9ac93e <master>:
    [  0.00%] ·· Building for virtualenv-py2.7, virtualenv-py3.7..
    [  0.00%] ·· Benchmarking virtualenv-py2.7
    [  1.79%] ··· Running (core.runner.time_echo--)..
    [ 14.29%] ··· core.runner.time_echo                                                           2.26±0.1ms
    [ 16.07%] ··· core.runner.time_echo_gitrunner                                                2.65±0.08ms
    [ 17.86%] ··· core.runner.track_overhead_100ms                                                      4.28
    [ 19.64%] ··· core.runner.track_overhead_echo                                                       98.6
    [ 21.43%] ··· core.runner.track_overhead_heavyout                                                    2.8
    [ 23.21%] ··· core.runner.track_overhead_heavyout_online_process                                  146.17
    [ 25.00%] ··· core.runner.track_overhead_heavyout_online_through                                   39.01
    [ 25.00%] ··· asv: forkserver:
    [ 25.00%] ··· [Level 5] Exiting

    [ 25.00%] ·· Benchmarking virtualenv-py3.7
    [ 26.79%] ··· Running (core.runner.time_echo--)..
    [ 39.29%] ··· core.runner.time_echo                                                           1.92±0.1ms
    [ 41.07%] ··· core.runner.time_echo_gitrunner                                                2.17±0.05ms
    [ 42.86%] ··· core.runner.track_overhead_100ms                                                      3.38
    [ 44.64%] ··· core.runner.track_overhead_echo                                                      83.94
    [ 46.43%] ··· core.runner.track_overhead_heavyout                                                   3.25
    [ 48.21%] ··· core.runner.track_overhead_heavyout_online_process                                  210.61
    [ 50.00%] ··· core.runner.track_overhead_heavyout_online_through                                   21.88
[...]

PR #2478, opened 2018-05-08
datalad run causes 100% CPU utilization by datalad 

yarikoptic:
#### What is the problem?
Just spotted that for me even running a dummy `datalad run sleep 100` causes datalad to become extremely busy with something... what is it doing? probably some side effect of how subprocess is used?

```
[DEBUG  ] tracking command output underneath <Dataset path=/tmp/QA> 
[Level 9] Running: ['git', '-c', 'receive.autogc=0', '-c', 'gc.auto=0', 'annex', 'status', '--debug', '-c', 'remote.origin.annex-ssh-options=-o ControlMaster=auto -S /home/yoh/.cache/datalad/sockets/23d1f58a', '--json', '--ignore-submodules=all', '--'] 
[Level 8] Finished running ['git', '-c', 'receive.autogc=0', '-c', 'gc.auto=0', 'annex', 'status', '--debug', '-c', 'remote.origin.annex-ssh-options=-o ControlMaster=auto -S /home/yoh/.cache/datalad/sockets/23d1f58a', '--json', '--ignore-submodules=all', '--'] with status 0 
[Level 8] Using existing Git repository at /tmp/QA 
[INFO   ] == Command start (output follows) ===== 
[Level 9] Running: ['sleep', '100']
```

PR #3322, opened 2019-04-12
OPT: do not bother sleeping when not output was received in the Runner processing of output

yarikoptic:
As "discussed" in #3319 
It could help quite a bit for any runs of `git-annex CMD --json` (not batched)

### DIFF
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -317,8 +317,8 @@ class Runner(object):
             if log_stderr_:
                 stderr_ = self._process_one_line(*stderr_args)
                 stderr += stderr_
-            if not (stdout_ or stderr_):
-                # no output was generated, so sleep a tiny bit
+            if stdout_ is None and stderr_ is None:
+                # no output was really produced, so sleep a tiny bit
                 time.sleep(0.001)
 
         # Handle possible remaining output


kyleam:
So the sleep came about from e3e8665e3 (BF: sleep a bit if there were no output in log_online mode to avoid 100% load, 2018-05-08), which was a response to #2478.  Won't merging this bring back that issue?

yarikoptic:
you are right @kyleam - thanks for looking it up.  Getting to the bottom of it. will report back shortly

yarikoptic:
ok, mystery is not resolved but mitigated by fixing up the check in cfb1a73e58e59f3c505ba8ce5d41f8a6770fa815

FTR I also came up with abomination of an adaptable sleep but I do not think it provided any notable advantage, so decided against it but here is the sketch of a patch (with some debug printouts etc):
```patch
diff --git a/.asv b/.asv
--- a/.asv
+++ b/.asv
@@ -1 +1 @@
-Subproject commit 9bdd343b290d86bb71874d7d17e7acc24d074052
+Subproject commit 9bdd343b290d86bb71874d7d17e7acc24d074052-dirty
diff --git a/datalad/cmd.py b/datalad/cmd.py
index 2e4cf837..30c14126 100644
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -305,12 +305,13 @@ class Runner(object):
                 expect_stderr or expect_fail
         )
 
+        nempty = 0
         while proc.poll() is None:
             # see for a possibly useful approach to processing output
             # in another thread http://codereview.stackexchange.com/a/17959
             # current problem is that if there is no output on stderr
             # it stalls
-            # Monitor if anything was output
+            # Monitor if anything was output and if nothing, sleep a bit
             stdout_, stderr_ = None, None
             if log_stdout_:
                 stdout_ = self._process_one_line(*stdout_args)
@@ -318,6 +319,18 @@ class Runner(object):
             if log_stderr_:
                 stderr_ = self._process_one_line(*stderr_args)
                 stderr += stderr_
+            if stdout_ is None and stderr_ is None:
+            #if not (stdout_ is not None or stderr_ is not None):
+            #if not (stdout_ or stderr_):
+                #print("HERE: %r %r" % (stdout_, stderr_))
+                # no output was generated, so sleep a tiny bit
+                time.sleep(min(0.001, 2**nempty * 0.000001))
+                nempty = min(20, nempty + 1)  # 2**20 ~= 1e6 - more than enough
+                # time.sleep(0.001)
+                pass
+            else:
+                #print("HERE EMPTY")
+                nempty = 0
 
         # Handle possible remaining output
         stdout_, stderr_ = proc.communicate()
```

kyleam:
> ok, mystery is not resolved but mitigated by fixing up the check in [cfb1a73](https://github.com/datalad/datalad/commit/cfb1a73e58e59f3c505ba8ce5d41f8a6770fa815)

Nice, thank you.

One thing I wonder is why we even need to call `_get_output_online` when the `run(..., log_online=True, log_stdout=False, log_stderr=False)`.  Can't we let that go through to `.communicate()`?  Something like

```diff
diff --git a/datalad/cmd.py b/datalad/cmd.py
index d86ea8787..ee941f79b 100644
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -514,7 +514,7 @@ def run(self, cmd, log_stdout=True, log_stderr=True, log_online=False,
                     self.protocol.end_section(prot_id, prot_exc)

             try:
-                if log_online:
+                if log_online and (log_stdout or log_stderr):
                     out = self._get_output_online(proc,
                                                   log_stdout, log_stderr,
                                                   outputstream, errstream,
```

yarikoptic:
one of the cases which this would change behavior of (don't know yet if with any positive or negative effects) is within sshrun
```
  /home/yoh/proj/datalad/datalad/venvs/dev/bin/datalad(8)<module>()
-> main()
  /home/yoh/proj/datalad/datalad/datalad/cmdline/main.py(499)main()
-> ret = cmdlineargs.func(cmdlineargs)
  /home/yoh/proj/datalad/datalad/datalad/interface/base.py(626)call_from_parser()
-> ret = cls.__call__(**kwargs)
  /home/yoh/proj/datalad/datalad/datalad/support/sshrun.py(89)__call__()
-> out, err = ssh(cmd, stdin=stdin_, log_output=False)
  /home/yoh/proj/datalad/datalad/datalad/support/sshconnector.py(162)__call__()
-> **kwargs)
> /home/yoh/proj/datalad/datalad/datalad/cmd.py(521)run()
-> out = self._get_output_online(proc,
```
and the logic there in SSHConnection is 
```
        kwargs = dict(
            log_stdout=log_output, log_stderr=log_output,
            log_online=not log_output
        )
```
so with such a change we would never hit that `_get_output_online` when called from it.  Adding this additional condition starts to break the tests such as `datalad/support/tests/test_gitrepo.py:test_GitRepo_ssh_fetch` ... with all the mess of the runner options it is hard to figure out exactly why, so I would prefer to stick to minimal change for now, just fixing the issue.

kyleam:
@yarikoptic:

> one of the cases which this would change behavior of (don't know yet if with any positive or negative effects) is within sshrun [...]

OK, thanks for looking into it.

> Adding this additional condition starts to break the tests such as `datalad/support/tests/test_gitrepo.py:test_GitRepo_ssh_fetch` ... with all the mess of the runner options it is hard to figure out exactly why, so I would prefer to stick to minimal change for now, just fixing the issue.

Sure, sounds fine to me.  I really was "just wondering", not objecting
to the proposed change here.

PR #3326, opened 2019-04-13
DOC: Auto-set metavar for parameters with value choices

mih:
Old output:

`--eval-subdataset-state EVAL_SUBDATASET_STATE`

New output:

`--eval-subdataset-state {no|commit|full}`

This improves the docs of a range of commands, and addresses a comment by @kyleam in https://github.com/datalad/datalad/pull/3324

### DIFF
--- a/datalad/distribution/create_sibling.py
+++ b/datalad/distribution/create_sibling.py
@@ -380,7 +380,7 @@ class CreateSibling(Interface):
         inherit=inherit_opt,
         shared=Parameter(
             args=("--shared",),
-            metavar='false|true|umask|group|all|world|everybody|0xxx',
+            metavar='{false|true|umask|group|all|world|everybody|0xxx}',
             doc="""if given, configures the access permissions on the server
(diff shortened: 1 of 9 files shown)


mih:
@kyleam Lovely, thx much!

kyleam:
> Just an observation (probably not directly related to changes proposed
> here)

Yeah, I'd say it's certainly orthogonal.

> - so we have None, "none", and "" used across different commands options to represent "nothing", might be something to unify?

Hmm I'm not sure.  None only works if it is being used as the default
value because it isn't accessible to Python callers.  I think it'd be
too inconvenient and disruptive for all of the spots that look like
__call__(..., param=None) to become __call__(..., param="none"), so it
seems like we need both None and at least one string variant that means
"none".  But I don't think we should collapse all empty string values to
"none" because "none" isn't always a good description (e.g., rerun's
--onto=).

On top of that, this is all user-facing, so we'd need deprecation
cycles.  At first glance I don't think any effort to unify would be
worth it.

PR #3338, opened 2019-04-15
TST(BK): demonstrate the regression from 0.11.4-5-g78e00dcd - remote submodule is not "initialized

yarikoptic:
Recently 0.11.4-5-g78e00dcd removed initialization of git submodules on remote
within .git/hooks since it was indiscriminate to the cases where such
initialization was not even desired.  Initially such initialization as was introduced
because our submodule listing was using "git submodule" directly and now we do
parsing ourselves and probably because of that probably it all works even
without proper "git submodule update --init  newsubmodule" to make it properly
registered within .git/config.

This test is primarily to demonstrate the effect and seek decision on either it should
be addressed or not.  The main problematic use case I see which might mandate
action is discovery of the web ui not being populated properly, especially whenever
remote host has somewhat datalad DataLad

Only when submitting this PR I realized that this test/changes sit straight on top of 0.11.4, so before all the changes to post-update hook, and it still fails!   so something is not really working out even without disabling the post-update hook and/or removing `git submodule update --init` from it.

The question is largely, should "datalad" care to establish proper/clean/properly initialized hierarchy of submodules on the remote?  I would say "yes".  But I am not sure what would be the cleanest way in terms of where in DataLad.

### DIFF
--- a/datalad/distribution/tests/test_publish.py
+++ b/datalad/distribution/tests/test_publish.py
@@ -674,3 +674,43 @@ def test_gh1763(src, target1, target2):
         assert_in(
             'probe1',
             Dataset(target).repo.get_annexed_files(with_content_only=True))
+
+
+@skip_if_on_windows  # create_sibling incompatible with win servers
+@skip_ssh
+@with_tempfile(mkdir=True)
+@with_tempfile()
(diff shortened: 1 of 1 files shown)


mih:

Issue #3342, opened 2019-04-17
`status` impairs usability in sizeable repos


yarikoptic:
e.g. git status (0.3sec) vs datalad status (45.8sec):

```shell
fmriprep001$ time git status                                                                                 
On branch master
Changes to be committed:
  (use "git reset HEAD <file>..." to unstage)

        new file:   fmriprep/dataset_description.json
        new file:   fmriprep/logs/CITATION.md

Untracked files:
  (use "git add <file>..." to include in what will be committed)

        freesurfer/fsaverage5/surf/.xdebug_tksurfer


real    0m0.253s
user    0m0.184s
sys     0m0.360s
```
NB - those "new file"s are somehow uncommitted after `datalad add` call on all of the files -- yet to possibly figure out reproducible demo
```shell
fmriprep001$ time datalad status                                                                             |
untracked: freesurfer/fsaverage5/surf/.xdebug_tksurfer (file)
    added: fmriprep/dataset_description.json (file)
    added: fmriprep/logs/CITATION.md (file)

real    0m45.858s
user    0m37.752s
sys     0m11.092s
```
so what additional functionality/advantages of `datalad status` demand x100 run time and either it is avoidable?

this dataset is fmriprep output on a dataset with 253 subjects, so there is ~95k files in it (git counted ~200k objects), majority is annexed, no submodules, 

After `git gc --prune=now`, `datalad status` took less -- 34 sec, `git ls-files` takes 0.108 sec, `git ls-tree master -r` also takes whooping 0.1 sec. `datalad save -m '...'` saved those staged files also within just a second or so,  `datalad rev-save` took 45 sec (and did save that `.xdebug_tksurfer` file). 
edit 1: Even when specifying explicitly paths for just a few files to be saved to `rev-save` in a repo with 66k files, requires then a handful of seconds to do a simple commit.

datalad 0.12.0rc2.dev268 (0.12.0rc2-268-ge79d4ea8)  python3

yarikoptic:
as for submodules
<details>
<summary>datalad subdatasets is nicely fast</summary>

```shell
WM-R03$ time datalad subdatasets -r --recursion-limit 1                                                                     
subdataset(ok): BIOS (dataset)
subdataset(ok): BPA2 (dataset)
subdataset(ok): COBY (dataset)
subdataset(ok): DIAMOND (dataset)
subdataset(ok): LAMS (dataset)
subdataset(ok): logs (dataset)
action summary:
  subdataset (ok: 6)

[...]

PR #3343, opened 2019-04-17
NF: GitRepo._diffstatus() internal helper for global state evaluation (fixes gh-3342)

mih:
This is used inside GitRepo.diffstatus() to implement an early-stopping strategy that yields a state label for a subdataset at the earliest sign of a modification. #3342 

Performance info:

Prep:
```
 import datalad.api as dl
 ds = dl.rev_create('deep')
 path = ds.pathobj  
 for i in range(20):       
      path = path / 'down'
      ds.rev_create(path)
```

followed by a

```
% touch deep/down/down/here
```

on present master
```
>>> timeit dl.Dataset('deep').status()
1 loop, best of 3: 1.16 s per loop
```

this PR
```
>>> timeit dl.Dataset('deep').status()
1 loop, best of 3: 198 ms per loop
```


mih:
Appveyor actually passed already...

PR #3347, opened 2019-04-22
ENH: issue a warning (using warnings.warn) if git is too old for us to check

### LINKED ISSUE #3295: older git (2.11.0) ls-files does not list cwd which belongs subdataset

yarikoptic:
for known paths from parent directory.

Decided to (start) using warnings.warn (instead of logging.warn) because

- we need to have it issued only once, not upon every init (why to flood?)

- we are expecting action on "client" side -- upgrade, so that is the recommended way
  to do it according to https://docs.python.org/2/howto/logging.html#when-to-use-logging

Closes #3295
Replaces #3296

### DIFF
--- a/datalad/support/exceptions.py
+++ b/datalad/support/exceptions.py
@@ -351,3 +351,19 @@ class CrawlerError(Exception):
 
 class PipelineNotSpecifiedError(CrawlerError):
     pass
+
+
+#
+# Warnings
+#
+
+class DataLadWarning(Warning):
+    pass
+
+
+# We have an exception OutdatedExternalDependency, but it is intended for
+# an instance being raised.  `warnings` module requires a class to be provided
+# as a category, so here is a dedicated Warning class
+class OutdatedExternalDependencyWarning(DataLadWarning):
+    """Warning "category" to use to report about outdated"""
+    pass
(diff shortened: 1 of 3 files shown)


kyleam:
Test failure was due to connectivity issue in setup.  Restarted.

PR #3350, opened 2019-04-24
More useful AnnexRepo.whereis()

mih:
https://github.com/datalad/datalad-revolution/pull/84 needs efficient batched whereis queries. The current `AnnexRepo.whereis()` is not suitable for this. It aims to be universal and fails at it. Given that `annex whereis` does not support all queries in batched mode, I decided that it would be best to have a dedicated implementation for path queries.

In contrast to the previous whereis() this one has:

- a simpler API (no options at all)
- cannot query for a key (not possible in batched mode)
- cannot do recursive queries (not possible in batched mode)
- states what file a response belongs too

I comes in two flavors: single-path queries, and generator for multi-path queries.

I also include an RF for `BatchedCommand` that adds a `yield_()` method which bypasses the collect-everything-into-a-list-first approach, which torpedos generator-style usage.

I will adjust the tests, once it is clear to me that this solution is optimal for the metadata use case.

### DIFF
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -778,8 +778,12 @@ class BatchedCommand(object):
         if not input_multiple:
             cmds = [cmds]
 
-        output = [o for o in self.yield_(cmds)]
-        return output if input_multiple else output[0]
+        output = list(self.yield_(cmds))
+        if input_multiple:
+            return output
+        else:
+            assert len(output) == 1, "There should be a single output item"
+            return output[0]
 
     def yield_(self, cmds):
         for entry in cmds:
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -778,8 +778,10 @@ class BatchedCommand(object):
         if not input_multiple:
             cmds = [cmds]
 
-        output = []
+        output = [o for o in self.yield_(cmds)]
+        return output if input_multiple else output[0]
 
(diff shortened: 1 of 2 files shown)


yarikoptic:
pushed a tiny tune up in 3d02ce477cbab4ae6890038e5f5711f3b5ce8020

- ideally we shouldn't breed functions, and probably original `whereis` could have been adjusted for the desired behavior, and may be stripped of some functionality if it is not really used
- it is not clear to me how output differs since in general seems to not work yet
```shell
(git-annex)hopa:~/datalad/openfmri/ds000001[master]sub-01/anat
$> python -c 'from datalad.support.annexrepo import AnnexRepo; r=AnnexRepo("../../"); print(len(r.whereis_file(["sub-01_inplaneT2.nii.gz"])))' 
  File "/home/yoh/proj/datalad/datalad-master/datalad/support/annexrepo.py", line 2222, in whereis_file
    return list(self.whereis_file_([path]))[0]
  File "/home/yoh/proj/datalad/datalad-master/datalad/support/annexrepo.py", line 2259, in whereis_file_
    status='ok' if r.get('success', False) else 'error',
AttributeError: 'list' object has no attribute 'get'
```

mih:
Ok, will leave present implementation as is, and implement dedicated support in metadata extractor.

Issue #3264, opened 2019-03-23
Command hooks


mih:
There is a dilemma between the desire to have a compact API and some customizability of command behavior. Example: `rev-create` does not have  `--text-no-annex`, but it can come handy to be able to get such a setup conveniently.

I believe a "hook" mechanism is one way to have this. Here is an example how a command invocation could look like that could be supported to run a dedicated configuration procedure
after a dataset got created:

  `datalad -c datalad.hooks.create-done=cfg_text2git rev-create`

This uses our tooling to have a temporary config setting, but of course such a setting could also become part of a dataset/user/system configuration.

Using this mechanism would be nice, because we already have a need to improve config override handling for better control of Git/Git-annex behavior within low-level commands (#3246, and various failed attempts along the lines of #3204, #2851, #2818, #905). There we face the exact same issue: desire to fine-tune, without the willingness or ability to explode the API to be able to pass through tons of special-purpose options.

TODOs if we want to move forward with this:

- agree on a nomenclature of such config keys. Proposal: as above, with last key segment being `command`-`custom slot name` (where each command describes its slots in its docs)
- agree on a syntax of such config values (plain procedure names, or stringified Python list of  arguments for run-procedure, or ...

ATM we do have `--proc-pre`, `--proc-post` arguments at the cmdline API level, but those do not scale and are hard to use for a specific customization within a larger command. I think they should be removed.

Eventually, we should have the ability to define criteria to detect specific results that pass through our main "event-loop", and to trigger specific actions/hooks. However, until we have that, I see no problem in implementing the execution of an initial set of hooks (like the example above) within individual command implementations. Importantly, those should merely define the position of a hook within the processing flow of a command, but not the nature of an intervention. So in the above example we simply have a loop over `ds.config.get('datalad.hooks.create-done')` where `run-procedure` is called for each value.

Possible use-cases to inform development
- populate git-annex metadata for added files (e.g. annotate with `distribution-restrictions` key=value pairs).  ATM [there is no way AFAIK](http://git-annex.branchable.com/git-annex-metadata/#comment-fde59930f108af0fff842f5e25351e93) to do so via `.gitattributes`

yarikoptic:
[...]

PR #3353, opened 2019-04-24
ENH (sugaring): -c | --cfg-proc option for rev-create to specify cfg_ procedure(s) to run

yarikoptic:
ATM there is no easy way to specify procedures to run on a freshly created dataset. But in my case it is a common use case to quickly create datasets where I know what type they should be - should not
place text files to git, or might be BIDS datasets, etc.  So it would be nice to be able to succinctly specify procedure(s) to be ran on the new dataset as to configure it.

Alternative implementation could be adding  --proc-post to the rev-create command itself. Pros then would be:
  - consistent with --proc-post for datalad command itself
  - could then run any procedure, not only cfg_ one

Cons:
  - would be notably more long/tedius to type - i.e.

        rev-create --proc-post cfg_text2git

    vs

        rev-create -c text2git

I also added a `lgr.info` call within `run_procedure` to signal what procedure is actually about to be ran.  Although might be considered a bit oxymoronic in the case of direct `run_procedure` invocation, makes sense everywhere else procedures could be invoked.

So now it would look like
```shell
$> datalad rev-create -c text2git -f /tmp/newds2        
[INFO   ] Creating a new annex repo at /tmp/newds2 
create(ok): /tmp/newds2 (dataset)                                                              
```

With something like this I think I wouldn't miss any other option of current `create` too much happen we just decide to replace `create` with `rev-create`.

TODOs:
- [ ] Document special purpose/treatment of `cfg_` procedures
- [ ] Decide on `cfg_` vs `setup_`?


datalad/interface/run_procedure.py
```diff
@@ -451,6 +451,7 @@ def __call__(
         lgr.debug('Attempt to run procedure {} as: {}'.format(
             name,
             cmd))
+        lgr.info("Running procedure %s", name)
```
mih: Maybe "upgrade" the debug message on the line before? Two log messages with almost identical content seem unnecessary.

kyleam:
Should we decide the fate of command hooks (#3264) before going forward with this?

datalad/interface/run_procedure.py
yarikoptic: Indeed! Good catch - I was blind.  Making it 
```python
        lgr.info('Running procedure %s', name)
        lgr.debug('Procedure command is %r', cmd)
```
this way we can see gory details of the command for debugging purposes (like apparently there is some trailing space ATM, does not matter but might come critical, who knows).

BTW looking at the 
```python
        lgr.debug('Attempt to run procedure {} as: {}'.format(
            name,
            cmd))
```
it is imho better to delay any string formatting until (and if) actually needed, that is why I use `%` form, and would do
```python
        lgr.debug('Attempt to run procedure %s as: %s', name, cmd)
```
I didn't time the effects, but assume that it should spare a cycle or two ;-)

yarikoptic:
> Should we decide the fate of command hooks (#3264) before going forward with this?

as [I have expressed myself](https://github.com/datalad/datalad/issues/3264#issuecomment-475870474) hooks specification doesn't allow for quick/concise option to achieve trivial dataset customization upon `create`.  Although might be more generic and useful (although besides this `create` use case I do not have immediate others yet), it would probably largely be complimentary. May be even this sugaring functionality could be refactored later to (ab)use hooks for some reason instead of invoking `run_procedure` directly.

Overall, I do not think there is need to decide on hooks before starting to provide this convenience `-c` option.  Or am I missing something?

yarikoptic:
My concern with this PR is necessity to agree on the `cfg_` prefix for this purpose and us already having at least two - `cfg_` and `setup_` which IMHO are the same thing here semantically:

```shell
$> datalad run-procedure --discover                                        
setup_bids_dataset (datalad_hirni/resources/procedures/setup_bids_dataset.py) [python_script]
setup_hirni_dataset (datalad_hirni/resources/procedures/setup_hirni_dataset.py) [python_script]
cfg_metadatatypes (../datalad-master/datalad/resources/procedures/cfg_metadatatypes.py) [python_script]
cfg_text2git (../datalad-master/datalad/resources/procedures/cfg_text2git.py) [python_script]
setup_yoda_dataset (../datalad-master/datalad/resources/procedures/setup_yoda_dataset.py) [python_script]
```
Do you think we could rename `setup_` to `cfg_` or there is some principled difference?

kyleam:
[quoting out of order]

@yarikoptic:

> Overall, I do not think there is need to decide on hooks before starting to provide this convenience `-c` option.  Or am I missing something?

I think you're right that this shouldn't be held up by command hooks.
If we end up adding hooks (and specifically a `create-done` one), this
should be easy enough to map onto that without any API change.

> as [I have expressed myself](https://github.com/datalad/datalad/issues/3264#issuecomment-475870474) hooks specification doesn't allow for quick/concise option to achieve trivial dataset customization upon `create`.

My question wasn't implying that the syntax there was an alternative to
sugaring.

> Although might be more generic and useful (although besides this
> `create` use case I do not have immediate others yet), it would
> probably largely be complimentary.

AFAICT it calls `run_procedure` where the `create-done` hook would,
without being hooked up to the configuration and being restricted to
procedures "cfg_*".  So...

> May be even this sugaring functionality could be refactored later to (ab)use hooks for some reason instead of invoking `run_procedure` directly.

I'd guess that, if we do add hooks, we should refactor this to work with
the `run_procedure` loop for the `create-done` spot.

mih:
I am fine with `cfg_` and adjusted accordingly: https://github.com/datalad/datalad-neuroimaging/pull/60

mih:
> TODO: Document special purpose/treatment of cfg_ procedures

What do you envision here, it seem done already?

datalad/interface/run_procedure.py
mih: I pushed such a modification.

mih:
@yarikoptic Pushed a small test too

yarikoptic:
Thank you @mih for all the tune ups!  I am ok to merge it as is or we could may be discuss the issue about possible need of allowing for parametrization which I initially brough up in https://github.com/datalad/datalad-neuroimaging/pull/60#issuecomment-487283125 and may be we better decide upon right away to not require changing API later on:

... if we figure out how to "parametrize" additional actions/parametrizations scalably/flexibly.  E.g. here `bids_derivative` is probably just a flavor of `bids`, so we could may be better encode/interface parameters of the procedures? i.e. I want `bids` which supports `derivatives` specification and may be some other specifics which might be orthogonal to being bids raw or derivative (e.g. multisession vs single session) and thus ideally be in a single command and not elevating one aspect (raw-vs-derivative, or single-vs-multisession) into procedure name.
In the case of `-c` we probably then should follow (for consistency) what you have already for --proc-pre/post: `--proc-pre <PROCEDURE NAME> [ARGS ...]`: `-c bids derivative` or `-c bids flavor=derivative`, although I dislike those open ended options and preferred some encoding within the name (`-c bids[derivative]` or `-c bids[flavor=derivative]`).

What do you think?

mih:
I think this is good for now. Any parameterization syntax should be discussed and implemented at a more global scope, as `run_procedure()`, the hooks, and storage of such specs in the config would all benefit from this.

yarikoptic:
Fwiw run_procedure already has one in place AFAIK, that is what I thought to follow. Similar treatment is established by crawl-init

PR #3355, opened 2019-04-24
DOC: initiate 4git_users document

yarikoptic:
@jbpoline has mentioned that it would have been useful to dismiss some of the
"black box" magic behind DataLad, so I took a stub a initiating such a document.
Just want to seek feedback/contributions first ;-)

### DIFF
--- a/docs/source/4git_users.rst
+++ b/docs/source/4git_users.rst
@@ -1,0 +1,108 @@
+DataLad For git/git-annex users
+*******************************
+
+Overview
+========
+
+DataLad could be considered a "thin layer" on top of ``git`` and ``git-annex``
+with the primary goal to provide a unified and simplified interface to
+manipulate git/git-annex repositories.  DataLad does neither stores any additional
+"state" information nor provides additional to `.git/objects` and
+`.git/annex/objects` "content store". So, at large, you can work with DataLad
+datasets as with any other git/git-annex repository, or you can use DataLad to work
+with regular git/git-annex repositories.  DataLad extensively relies on standard
+``git submodule`` to manage collections/hierarchies of the repositories
+(called :term:`superdataset` in DataLad terms). So you could even use
+`git submodule` commands directly.
+
+Having said all that DataLad adds and uses some additional information to git/git-annex
+repositories:
+
+- DataLad uses plain `git config` for managing configuration/settings, but in addition
+  to regular configuration file locations (such as `.git/config`) it also considers
+  `.datalad/config` which is also kept under git.
+- The main difference of DataLad :term:`dataset` from a regular git/git-annex
+  repository is that each dataset having a unique identifier (UUID) assigned and committed
+  (by `datalad create`) within `.datalad/config` (``datalad.dataset.id``).
+- DataLad's meta-data "facilities" store extracted and aggregated metadata
+  under `.datalad/metadata`.
+
+
+Commands
+========
+
+Primary difference(s) from regular git/git-annex operation is that DataLad
+
+- can operate on paths which belong to some repository other than the current
+  one. Both git and git-annex require first to ``cd`` to the target repository.
+  DataLad first figures out what repository given paths belong and then works
+  within that repository. `-d` option could be used to instruct DataLad to work
+  within specific repository.
+
+As DataLad is just a "thin layer", most of the actual "useful" work is done via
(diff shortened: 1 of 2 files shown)


docs/source/4git_users.rst
```diff
@@ -0,0 +1,108 @@
...
+
+- DataLad uses plain `git config` for managing configuration/settings, but in addition
+  to regular configuration file locations (such as `.git/config`) it also considers
+  `.datalad/config` which is also kept under git.
+- The main difference of DataLad :term:`dataset` from a regular git/git-annex
+  repository is that each dataset having a unique identifier (UUID) assigned and committed
+  (by `datalad create`) within `.datalad/config` (``datalad.dataset.id``).
+- DataLad's meta-data "facilities" store extracted and aggregated metadata
+  under `.datalad/metadata`.
+
```
mih: We do have issues when working with non-dataset repositories. Even with old datalad datasets that lack the default .gitattributes config. Stuff ends up being annexed, sometimes unintentionally and annoyingly.

docs/source/4git_users.rst
```diff
@@ -0,0 +1,108 @@
...
+~~~~~~
+
+``datalad create DS``::
+
+    mkdir DS && \
+     cd DS && \
+     git init && \
+     git annex init && \
+     mkdir .datalad && \
+     echo '* annex.backend=MD5E' > .gitattributes && \
```
mih: This is not the case for `rev-create`, more attributes are being set:

```
.
├── .datalad
│   ├── config
│   └── .gitattributes
└── .gitattributes
```

docs/source/4git_users.rst
```diff
@@ -0,0 +1,108 @@
...
+     git add .gitattributes && \
+     git commit -m '[DATALAD] Set default backend for all files to be MD5E' && \
+     git config -f .datalad/config --add datalad.dataset.id `uuid`
+     git commit -m '[DATALAD] New dataset'
```
mih: `rev-create` thankfully makes one commit only and saves a lot of runtime with it ;-)

mih:
I am not sure how far you can get with listing equivalent git/annex commands at this level of detail (and how we want to keep this information up-to-date). I would limit this to a very high-level view. `create` could be:

```
git init
git annex init
<add default configruation>
git commit
```

docs/source/4git_users.rst
yarikoptic: that is worth a note indeed, something like

DataLad `create` adds "safe-guards" in the configuration to streamline working with git-annex, e.g. to ensure that git configuration files do not get added to the annex etc.

docs/source/4git_users.rst
yarikoptic: might be worth expanding, or not... As mentioned in the opening, this document is not to provide an exhaustive list of commands we run, but more of the "high level" schematics on what we do.

docs/source/4git_users.rst
```diff
@@ -0,0 +1,103 @@
...
+
+At its core, DataLad can be considered a thin layer on top of ``git`` and
+``git-annex`` that tries to unify and simplify working with repositories.
+DataLad does not store additional state information or use a content store
+outside of ``.git/objects`` and ``.git/annex/objects``.  You can work with
+DataLad datasets as with any other git/git-annex repository, or you can use
+DataLad to work with regular git/git-annex repositories.  DataLad extensively
+relies on git's builtin ``submodule`` mechanism to manage a hierarchy of
+repositories (:term:`superdataset` in DataLad terms), so you could even use
+``git submodule`` directly.
```
kyleam: I think somewhere we should explain our use of non-absorbed submodules because anyone familiar with submodules outside of DataLad might be surprised by this, but I doubt this is the place to do that.

mih:
Closing due to inactivity.

adswa:
(FTR: I will steal some of this for the handbook FAQ where applicable -- will commit with author being @yarikoptic )

yarikoptic:
Please steal as much as possible, no need for authorship transfer ;)  On the larger note -- would indeed be really informative in the handbook to have a common thread of "Direct Git/git-annex Analog"/"Under the hood" set of brief explanations for DataLad commands on what they do.  Might better be in a separate appendix and referred to from relevant parts in the text.

PR #3364, opened 2019-04-30
ENH: AnnexRepo.get_metadata(batch=True) 

mih:
This is a fragile enterprise due to the limitations of `BatchedCommand()`. When queried for a non-annexed file, git annex will error not via JSON response, but through stderr

```
{"file": "."}
git-annex: not an annexed file: .
```

which ATM is not monitored at all. I assume due to the general sadness of subprocess under PY2.

Docstrings warn about this, and there is no `batch=None` automagic because of this.

Other than that, it works.

This is needed to achieve a sensible performance of the `annex` extractor in -revolution (the behavior of the one in -core is pointless anyways), where all these limitation do not matter.

Waiting here: https://github.com/datalad/datalad-revolution/compare/enh-annex

### DIFF
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -770,10 +770,6 @@ class BatchedCommand(object):
         str or list
           Output received from process.  list in case if cmds was a list
         """
-        # TODO: add checks -- may be process died off and needs to be reinitiated
-        if not self._process:
-            self._initialize()
-
         input_multiple = isinstance(cmds, list)
         if not input_multiple:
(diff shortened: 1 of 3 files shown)


yarikoptic:
Two things to fix:
- our `BatchedCommand` should not hang whenever the output is all in stderr
- `git-annex metadata --json ` call should report json records with errors (@yarikoptic thinks we had similar case for some other annex commands, may be `add` or alike before)

Issue #3361, opened 2019-04-29
we might need to reconsider approach for text2git


yarikoptic:
Attn @joeyh 
ATM we are using 
```
* annex.largefiles=(not(mimetype=text/*))
```
but as we have recently discovered ([git-annex bug report](http://git-annex.branchable.com/bugs/manages_to_incorrectly_add_to_annex_instead_of_git_based_on___34__mimetype__34___-_we_cannot_figure_it_out_why/#comment-e0fec0da5dec5241db65324dcba2d3be)), it has stopped working for .json files since `file/libmagic` acquired additional powers to annotate them as `application/json`.

# Possible "proper" solutions

[Discussion on the `file` mailing list](https://mailman.astron.com/pipermail/file/2019-April/000106.html) brought following,
- We might want to consider using instead `encoding` not `type`. So we could then say:
```
* annex.largefiles=(mimeencoding=binary)
```
which is kinda neat IMHO!  But that would require adding support for that to git-annex
- We could (potentially, since there is [a bug in file/libmagic forbidding it](https://bugs.astron.com/view.php?id=77)) use `-k, --keep-going` flag and get all entries for each file to be matched.  But that would also require changes to git-annex and possibly more substantial since probably it was not coded to associate multiple mime types per file

There is also `-e json` which could be added to `file` call to exclude detection of json, but if anything (again, changed in annex) it would be a dirty hack/workaround, nothing really to rely upon.

# Temporary workaround
So I guess, in the short run, we should just add explicit 
```
*.json annex.largefiles=nothing
```
or alike (limit by size in addition).

Any other ideas?

joeyh:
I like the mimeencoding=binary idea and have implemented it in
git-annex.

-- 
see shy jo

yarikoptic:
Awesome, thank you @joeyh

yarikoptic:
With git-annex-standalone=7.20190322+git133-g59922f1f4-1~ndall+1 in `debian-devel` of neurodebian `mimeencoding` is indeed available and seems to work for this use case:

first replicating prior erroneous behavior:
```shell
$> datalad create --text-no-annex /tmp/testds1
[INFO   ] Creating a new annex repo at /tmp/testds1 
create(ok): /tmp/testds1 (dataset)

$> cd /tmp/testds1

$> f=file.json; echo '{"a": 1}' >| $f; datalad add $f; ls -ld $f 
add(ok): /tmp/testds1/file.json (file)
[...]

PR #3373, opened 2019-05-01
Temp test devel annex with mimeencoding support

yarikoptic:
This one should provide mimeencoding support for better detection of text files, see https://github.com/datalad/datalad/issues/3361#issuecomment-488349192


yarikoptic:
hm, apparently travis isn't running on Draft PRs! Apparently it is "by design" but debated now: https://travis-ci.community/t/draft-pull-requests-not-being-built/2434/9 
So as far as I see it Draft PRs are really for drafting something and not even expecting it to be tested. Moving this PR into an Open one but with do not merge label

kyleam:
@yarikoptic:

> hm, apparently travis isn't running on Draft PRs!

Yep, in fact we've already discussed it :]

https://github.com/datalad/datalad/pull/3208#issuecomment-470866713

yarikoptic:
Oh, missed that lively discussion, thanks

yarikoptic:
so - tests pass (so no bad side effects), coverage differences make no sense to me

PR #3376, opened 2019-05-02
RF: Use a dedicated new exception InvalidGitReferenceError

### LINKED ISSUE #3375: Use dedicated exception to report invalid git reference
As pointed out here: https://github.com/datalad/datalad/pull/3366/files#r280067311

ATM `GitRepo.get_content_info()` raises a `ValueError` after string-matching a message in `stderr`. I don't know how to do it differently, but certainly the need for another string operation in high-level functions can be avoided with a dedicated exception.

We need to look for such users and RF.

yarikoptic:
Closes #3375 
So my comment was nothing really "tricky" but may be I have missed the point.

### DIFF
--- a/datalad/core/local/diff.py
+++ b/datalad/core/local/diff.py
@@ -47,6 +47,9 @@ from datalad.core.local.status import (
     Status,
     _common_diffstatus_params,
 )
+from datalad.support.exceptions import (
+    InvalidGitReferenceError,
+)
 
 lgr = logging.getLogger('datalad.core.local.diff')
 
@@ -251,15 +254,12 @@ def _diff_ds(ds, fr, to, constant_refs, recursion_level, origpaths, untracked,
             untracked=untracked,
             eval_file_type=eval_file_type,
             _cache=cache)
-    except ValueError as e:
-        msg_tmpl = "reference '{}' invalid"
-        # not looking for a debug repr of the exception, just the message
-        estr = text_type(e)
-        if msg_tmpl.format(fr) in estr or msg_tmpl.format(to) in estr:
+    except InvalidGitReferenceError as e:
+        if e.ref in (fr, to):
             yield dict(
                 path=ds.path,
                 status='impossible',
-                message=estr,
+                message=text_type(e),
             )
             return
         raise
--- a/datalad/core/local/diff.py
+++ b/datalad/core/local/diff.py
@@ -20,6 +20,7 @@ from six import (
 from collections import OrderedDict
 from datalad.utils import (
     assure_list,
+    assure_unicode,
 )
 from datalad.interface.base import (
     Interface,
@@ -118,8 +119,8 @@ class Diff(Interface):
         for r in _diff_cmd(
                 ds=ds,
                 dataset=dataset,
-                fr=fr,
-                to=to,
+                fr=assure_unicode(fr),
+                to=assure_unicode(to),
                 constant_refs=False,
                 path=path,
                 annex=annex,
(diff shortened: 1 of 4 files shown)


datalad/core/local/diff.py
```diff
@@ -251,15 +254,12 @@ def _diff_ds(ds, fr, to, constant_refs, recursion_level, origpaths, untracked,
             untracked=untracked,
             eval_file_type=eval_file_type,
             _cache=cache)
-    except ValueError as e:
-        msg_tmpl = "reference '{}' invalid"
-        # not looking for a debug repr of the exception, just the message
-        estr = text_type(e)
-        if msg_tmpl.format(fr) in estr or msg_tmpl.format(to) in estr:
+    except InvalidGitReferenceError as e:
+        if e.ref in (fr, to):
```
yarikoptic: not even sure if this check is strictly needed here, but I guess would not hurt?

mih: I'd prefer not to have it, because we don't know what future `get_content_info()` implementations will do.

yarikoptic:
@mih -- pushed 2b6b5f4a9e216094d7924c22466d9b5621b1e591 to remove that check

datalad/support/gitrepo.py
```diff
@@ -2907,7 +2908,7 @@ def get_content_info(self, paths=None, ref=None, untracked='all',
                 expect_fail=True)
         except CommandError as exc:
             if "fatal: Not a valid object name" in str(exc):
```
kyleam: Not introduced by this PR, but this could easily fail on py2:

```
$ datalad diff -f HEAD~ -t β-doesnt-exist
[ERROR  ] 'ascii' codec can't encode character u'\u03b2' in position 387: ordinal not in range(128) [gitrepo.py:get_content_info:2910] (UnicodeEncodeError)
```

If we were to fix that:

```diff
diff --git a/datalad/support/gitrepo.py b/datalad/support/gitrepo.py
index e9df21970..ff9ba6477 100644
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -2907,7 +2907,7 @@ def get_content_info(self, paths=None, ref=None, untracked='all',
                 # we don't want it to scream on stdout
                 expect_fail=True)
         except CommandError as exc:
-            if "fatal: Not a valid object name" in str(exc):
+            if "fatal: Not a valid object name" in text_type(exc):
                 raise InvalidGitReferenceError(ref)
             raise
         lgr.debug('Done query repo: %s', cmd)
```

then we'd get an error in code that this PR touched (though a similar error would be there before this PR):

```
$ datalad diff -f HEAD~ -t β-doesnt-exist
[ERROR  ] 'ascii' codec can't decode byte 0xce in position 15: ordinal not in range(128) [diff.py:_diff_ds:261] (UnicodeDecodeError) 
```

To complete the fix, we could (1) ensure to/fr are unicode when they first come in and (2) make `InvalidGitReferenceError.__str__` return unicode on py2 (ideally it'd use six.python_2_unicode_compatible or something like it, but we don't currently do that for any of our `__str__` methods).

yarikoptic: Nice find. So - should we deal with it here or after? Could you post suggested changes?

kyleam:
> So - should we deal with it here or after?

I suspect there are lots of similar issues in spots where we use our custom `__str__` methods, but we might as well deal with this particular one now.

> Could you post suggested changes?

Pushed.

yarikoptic:
Thank you @kyleam!

PR #3388, opened 2019-05-07
Prep for 0.12.0rc3 release

mih:
Final changelog touch-up after #3387, plus version bump, and RST changelog update.


mih:
Will merge now, and proceed with a PyPi only release.

PR #3391, opened 2019-05-07
RF: no need for import X as X - just import X

yarikoptic:
Spotted in rev_create -> create PR


kyleam:
Thanks for catching those!

PR #3397, opened 2019-05-08
RF: versioned depends on sphinx within setup.py itself

yarikoptic:
Ideally there should be no "manual" installations within CI setups, and such versioned dependencies are due to be specified in setup.py -- similar but a bit more extensive PR is coming to datalad-neuroimaging


yarikoptic:
travis is happy, everyone else is quiet and happy, merging!

PR #3400, opened 2019-05-11
RF: Various little annoyances in SSH support code

mih:
- [x] stop use of `/dev/null`
- [x] stop unconditional use of POSIX path conventions


mih:
Thx @kyleam !

PR #3402, opened 2019-05-13
On-demand SSHConnection.get/put()

### LINKED ISSUE #3401: SSHConnection.copy() is actually copy_to_remote()
It will only use the connection to copy things to the remote host, and not from it to the local host. Given the nature of the matter (there is an existing connection to a single remote host in SSHConnection) it makes sense to me to add a sibling method that does the same as `copy()` (prefix the remote paths with the connection host), but for the source files. This would also need a rename of `copy()` to something like `copy_to_remote()`, so that the new method becomes `copy_from_remote()`.

### LINKED ISSUE #3403: SSHConnection.copy() does not open a shared connection
The docs promissed that connections are opened on demand. However, that is not the case. I control master is given, but if none exists it will go on without starting on.

mih:
- [x] methods for file transfer from AND to a remote host (fixes #3401)
- [x] add test once API has gotten positive feedback
- [x] fixes #3403 (on demand connection)
- [x] allows for disabling remote annex bundle detection -- e.g. useful when it is known that a connection is not used to run Git commands. Include switch in connection hash so that there can be two connections to the same system/user, one with and one without a modified PATH
- [x] add a tiny test that should also run on windows, but doesn't, hence disabled for now (comment inside)
- [x] add a test for `get()`

### DIFF
--- a/datalad/support/sshconnector.py
+++ b/datalad/support/sshconnector.py
@@ -33,6 +33,7 @@ from datalad.dochelpers import exc_str
 from datalad.utils import (
     auto_repr,
     Path,
+    assure_list,
 )
 from datalad.cmd import Runner
 
@@ -263,7 +264,8 @@ class SSHConnection(object):
                 # not a "normal" SSH error
                 raise e
 
-    def copy(self, source, destination, recursive=False, preserve_attrs=False):
+    def copy_to_remote(self, source, destination,
+                       recursive=False, preserve_attrs=False):
         """Copies source file/folder to destination on the remote.
 
         Parameters
@@ -286,8 +288,7 @@ class SSHConnection(object):
         scp_cmd = ["scp"] + scp_options
 
         # add source filepath(s) to scp command
-        scp_cmd += source if isinstance(source, list) \
-            else [source]
+        scp_cmd += assure_list(source)
 
         # add destination path
         scp_cmd += ['%s:"%s"' % (self.sshri.hostname, destination)]
--- a/datalad/support/sshconnector.py
+++ b/datalad/support/sshconnector.py
@@ -512,7 +512,9 @@ class SSHManager(object):
         if ctrl_path in self._connections:
             return self._connections[ctrl_path]
         else:
-            c = SSHConnection(ctrl_path, sshri, identity_file=identity_file)
+            c = SSHConnection(
+                ctrl_path, sshri, identity_file=identity_file,
+                use_remote_annex_bundle=use_remote_annex_bundle)
             self._connections[ctrl_path] = c
             return c
 
--- a/datalad/support/sshconnector.py
+++ b/datalad/support/sshconnector.py
@@ -125,7 +125,7 @@ class SSHConnection(object):
         """
 
         # TODO:  do not do all those checks for every invocation!!
-        # TODO: check for annex location once, check for open socket once
+        # TODO: check for open socket once
         #       and provide roll back if fails to run and was not explicitly
         #       checked first
         # make sure we have an open connection, will test if action is needed
--- a/datalad/support/sshconnector.py
+++ b/datalad/support/sshconnector.py
@@ -301,6 +301,35 @@ class SSHConnection(object):
         scp_cmd += ['%s:"%s"' % (self.sshri.hostname, destination)]
         return self.runner.run(scp_cmd)
 
+    def copy_from_remote(self, source, destination,
+                         recursive=False, preserve_attrs=False):
+        """Copies source file/folder from remote to a local destination.
+
+        Parameters
+        ----------
+        source : str or list
+          file/folder path(s) to copy from the remote host
+        destination : str
+          file/folder path to copy to on the local host
+        recursive : bool
+          flag to enable recursive copying of given sources
+        preserve_attrs : bool
+          preserve modification times, access times, and modes from the
+          original file
+
+        Returns
+        -------
+        str
+          stdout, stderr of the copy operation.
+        """
+        scp_cmd = self._get_scp_command_spec(recursive, preserve_attrs)
+        # add source filepath(s) to scp command, prefixed with the remote host
+        scp_cmd += ['%s:"%s"' % (self.sshri.hostname, s)
+                    for s in assure_list(source)]
+        # add destination path
+        scp_cmd += [destination]
+        return self.runner.run(scp_cmd)
+
     def get_annex_installdir(self):
         key = 'installdir:annex'
         if key in self._remote_props:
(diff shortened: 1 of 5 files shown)


datalad/support/sshconnector.py
```diff
@@ -263,36 +264,72 @@ def close(self):
...
+          original file
+
+        Returns
+        -------
+        str
+          stdout, stderr of the copy operation.
+        """
+        scp_cmd = self._get_scp_command_spec(recursive, preserve_attrs)
+        # add source filepath(s) to scp command, prefixed with the remote host
+        scp_cmd += ['%s:"%s"' % (self.sshri.hostname, s)
```
kyleam: sidenote: I know this is just the same as what already existed in the copy_to variant, but I'm surprised that scp would be ok with the unconditional quoting of the file name (e.g., I think it'd fail if the file name contained `"`).

mih:
> slight preference for a fabric-inspired get()/put()

Sounds good to me too.

datalad/support/sshconnector.py
mih: Indeed. Looks like another bug.

mih: Nope, file with quotes in the name works, and is tested!

kyleam: All right.  Thanks for checking.

mih:
I rebased on master and disabled to windows test (leaving a comment). This is a whole different side issue that I would rather not approach in this PR.

mih:
Pushed a small smoke test for `get()`. This is now ready and I will merge once the tests are through. And proceed to tag rc4

datalad/support/tests/test_sshconnector.py
```diff
@@ -212,10 +213,16 @@ def test_ssh_copy(sourcedir, sourcefile1, sourcefile2):
         with open(targetpath, 'r') as fp:
             eq_(content, fp.read())
 
+    # and now a quick smoke test for get
+    togetfile = Path(targetdir) / '2get.txt'
+    togetfile.write_text(text_type('something'))
+    ssh.get(opj(remote_url, togetfile), sourcedir)
+    (Path(sourcedir) / '2get.txt').exists()
```
kyleam: Missing assertion.

datalad/support/tests/test_sshconnector.py
```diff
@@ -212,10 +213,16 @@ def test_ssh_copy(sourcedir, sourcefile1, sourcefile2):
         with open(targetpath, 'r') as fp:
             eq_(content, fp.read())
 
+    # and now a quick smoke test for get
+    togetfile = Path(targetdir) / '2get.txt'
+    togetfile.write_text(text_type('something'))
```
kyleam: or `u'something'` :)

datalad/support/tests/test_sshconnector.py
```diff
@@ -264,3 +272,21 @@ def test_ssh_git_props():
...
+# local and "remote" needs more research
+@skip_if_on_windows
+@skip_ssh
+@with_tempfile(mkdir=True)
+def test_bundle_invariance(path):
+    remote_url = 'ssh://localhost'
+    manager = SSHManager()
+    testfile = Path(path) / 'dummy'
+    for flag in (True, False):
+        assert(not testfile.exists())
```
kyleam: nitpick: Unnecessary parentheses (same below as well), though I think `assert_false` would be more consistent with other parts of the codebase.

datalad/support/tests/test_sshconnector.py
mih: Jeez....done. Thx.

datalad/support/tests/test_sshconnector.py
mih: yeah... will keep this one.

datalad/support/tests/test_sshconnector.py
mih: True, changed to `assert_false`

mih:
I have no idea why this fails on travis:

```
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/datalad/support/sshconnector.py", line 353, in get
    return self.runner.run(scp_cmd)
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/datalad/cmd.py", line 553, in run
    raise CommandError(text_type(cmd), msg, status, out[0], out[1])
datalad.support.exceptions.CommandError: CommandError: command '['scp', '-o', 'ControlPath="/home/travis/.cache/datalad/sockets/17e65fd7"', '-P', '22', 'localhost:"/tmp/datalad_temp_test_ssh_copyfywot1ey.c opy/2get.txt"', '/tmp/datalad_temp_test_ssh_copyfywot1ey']' failed with exitcode 1
Failed to run ['scp', '-o', 'ControlPath="/home/travis/.cache/datalad/sockets/17e65fd7"', '-P', '22', 'localhost:"/tmp/datalad_temp_test_ssh_copyfywot1ey.c opy/2get.txt"', '/tmp/datalad_temp_test_ssh_copyfywot1ey'] under None. Exit code=1. out= err=protocol error: filename does not match request
```

Works fine locally. Giving up for today.

kyleam:
> I have no idea why this fails on travis:

It looks like this is due to a recent change in openssh.  See <https://stackoverflow.com/questions/54598689/scp-fails-with-protocol-error-filename-does-not-match-request>.

Debian has applied this fix to older versions, so at least stretch has it despite not having V_8_0_P1.  

This test fails for me locally, but I can get it to pass with the `-T` flag added in the commit referenced by the link above.  We can't unconditionally add that to the scp options, though, because older versions won't have it.  And we can't check the openssh version because it may have been applied as a security fix.  I suppose we could try to catch the failure and add `-T` as an option, but for now I'd say we should just not test get() with an obscure filename and accept that our get() may fail for weird names (leaving a comment).

yarikoptic:
What if we just make adding (or not) -T configurable via some config option (it even more generic DATALAD_SCP_EXTRA_OPTIONS, empty by default), so if failure happens there is a possible workaround at least?

mih:
Or we use rsync right away. But for now, I think I will go with the approach to not test on complicated paths.

mih:
I an the end I went for a different approach to quoting filenames. I disliked the idea of using `-T`, or even recommending/suggesting that people would need to do, as it would have been required for filenames with spaces already -- and those are not THAT uncommon.

datalad/support/sshconnector.py
```diff
@@ -294,6 +294,11 @@ def _get_scp_command_spec(self, recursive, preserve_attrs):
     def put(self, source, destination, recursive=False, preserve_attrs=False):
         """Copies source file/folder to destination on the remote.
 
+        Note: this method perform escaping of filenames to an extend that
+        moderately weird ones should work (spaces, quotes, pipes, other
+        characters with special shell meaning), but more complicated cases
+        might require appropriate external preprocessing of filenames.
+
```
kyleam: My understanding is that the upstream change only concerns remote => local, so making this change for put() isn't strictly necessary.  Though for consistency perhaps it still makes sense for put() and get() to have consistent file name washing.

PR #3405, opened 2019-05-14
TEMP: check if wrapt devel fixed segfault

yarikoptic:
there is some fresh info that they might be gone
https://github.com/GrahamDumpleton/wrapt/issues/98#issuecomment-492283655


yarikoptic:
so neither fails with release nor with devel -- I will remove `WRAPT_DISABLE_EXTENSIONS=1` in the followup PR

PR #3406, opened 2019-05-14
ENH: Use 3.7 (not -dev) and add a single run for 3.8-dev

yarikoptic:
If passes, up to discussion either we care enough to test for it already

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,8 @@ python:
   - 2.7
   - 3.5
 ##1  - 3.6
+##1  - 3.7
+##1  - 3.8-dev
 
 cache:
   - apt
@@ -35,7 +37,14 @@ env:
 matrix:
   include:
   # Additional custom ones
-  - python: 3.7-dev
+  - python: 3.8-dev
+    # Single run for Python 3.8
+    env:
+    # Run all tests in a single whoop here
+    # We cannot have empty -A selector, so the one which always will be fulfilled
+    - NOSE_SELECTION=
+    - NOSE_SELECTION_OP=not
+  - python: 3.7
     # Single run for Python 3.7
     env:
     # Run all tests in a single whoop here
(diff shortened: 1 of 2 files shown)


kyleam:
> Use 3.7 (not -dev) 

I think this will fail.  At least it did when I tried in gh-3339:

https://travis-ci.org/datalad/datalad/jobs/520854708

> and add a single run for 3.8-dev. If passes, up to discussion either we care enough to test for it already

I vote not to unless we drop another build somewhere else.

kyleam:
I wrote:

> > Use 3.7 (not -dev)
> 
> I think this will fail. At least it did when I tried in [gh-3339](https://github.com/datalad/datalad/pull/3339):
> 
> https://travis-ci.org/datalad/datalad/jobs/520854708

Indeed, still appears to be the case:

https://travis-ci.org/datalad/datalad/jobs/532444091

kyleam:
3.8-dev build failed in a similar way.  I think we'd need to use xenial to make 3.7 and 3.8-dev available:

https://docs.travis-ci.com/user/languages/python/#python-37-and-higher

yarikoptic:
thank you @kyleam - will do set it to use xenial for those two

kyleam:
> thank you @kyleam - will do set it to use xenial for those two

Thanks, that worked, but the 3.8 build didn't run because we use platform.dist() in setup.py, which was removed in 3.8.  I pushed a temporary commit that comments that out so that we hopefully we can actually see the tests run.

https://travis-ci.org/datalad/datalad/jobs/532829023

kyleam:
OK, here's the 3.8 run: https://travis-ci.org/datalad/datalad/jobs/532901906

Looks like we have [two errors][0] related to wtf trying to use the removed platform.dist().  There are also a bunch of regexp-related [SyntaxWarnings][1] that we should fix up.

[0]: https://travis-ci.org/datalad/datalad/jobs/532901906#L1851
[1]: https://travis-ci.org/datalad/datalad/jobs/532901906#L1891

PR #3407, opened 2019-05-14
NF: Add check for whether color is enabled

### LINKED ISSUE #3404: Request: Respect NO_COLOR environment variable
#### What is the problem?

There is no obvious way to disable ANSI colors in the terminal outputs. There is an effort to have CLI tools uniformly respect a `NO_COLOR` environment variable, which seems like a good place to start. See https://no-color.org/.

The context this comes up in is in [this example notebook](https://fitlins.readthedocs.io/en/latest/examples/ds003_sample_analysis.html#install-data-and-fitlins), where Sphinx does not properly render ANSI colors. It would be useful to simply disable them at the datalad end, rather than play around on the parsing end.

[snipped other headers]

effigies:
Checks for interactive terminal, `datalad.ui.color` configuration and `NO_COLOR` environment variable, in that order of precedence.

Closes #3404.

### DIFF
--- a/datalad/interface/ls.py
+++ b/datalad/interface/ls.py
@@ -366,16 +366,6 @@ class FsModel(AnnexModel):
 
 
 class LsFormatter(string.Formatter):
-    # condition by interactive
-    if is_interactive():
-        BLUE = ansi_colors.COLOR_SEQ % ansi_colors.BLUE
-        RED = ansi_colors.COLOR_SEQ % ansi_colors.RED
-        GREEN = ansi_colors.COLOR_SEQ % ansi_colors.GREEN
-        RESET = ansi_colors.RESET_SEQ
-        DATASET = ansi_colors.COLOR_SEQ % ansi_colors.UNDERLINE
-    else:
-        BLUE = RED = GREEN = RESET = DATASET = u""
-
     # TODO: we might want to just ignore and force utf8 while explicitly .encode()'ing output!
     # unicode versions which look better but which blow during tests etc
     # Those might be reset by the constructor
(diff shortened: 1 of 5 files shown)


effigies:
Let me know if this seems like the right approach. I can add tests if you agree.

datalad/support/ansi_colors.py
```diff
@@ -39,9 +41,34 @@
...
+    Color is only enabled if the terminal is interactive.
+    If the datalad.ui.color configuration setting is 'on' or 'off', then
+    respect that.
+    If the datalad.ui.color setting is 'auto' (default), then color is
+    enabled unless the environment variable NO_COLOR is defined.
+    """
+    if not ui.is_interactive:
+        return False
+
+    UIC = cfg.get_value('datalad', 'ui.color', 'auto')
```
yarikoptic: Thanks!
Ideally I think we should add an item (at the end, along with `datalad.ui.progressbar`) to datalad/interface/common_cfg.py and use `cfg.obtain('datalad.ui.color')` here which would then pick up default from the common cfg definition.

effigies:
Cool. Moving to a config option that is required to have a value made the checks much shorter. Testing is going to involve some manipulation of the config, and my reading is that setting values stores them on disk. Do you have helper functions for creating an environment with a disposable config?

datalad/interface/ls.py
```diff
@@ -415,15 +407,22 @@ def convert_field(self, value, conversion):
...
-                return u"%s%s%s" % (self.RED, self.NONE, self.RESET)
+                return ansi_colors.color_word(self.NONE, ansi_colors.RED)
             return value
         elif conversion in {'B', 'R', 'U'}:
-            return u"%s%s%s" % ({'B': self.BLUE, 'R': self.RED, 'U': self.DATASET}[conversion], value, self.RESET)
+            return ansi_colors.color_word(
+                value,
+                {'B': ansi_colors.BLUE,
+                 'R': ansi_colors.RED,
+                 'U': self.DATASET}[conversion])
```
effigies: `self.DATASET` is equivalent to `ansi_colors.DATASET`, though given that the code is `'U'`, perhaps `ansi_colors.UNDERLINE` makes more sense than the indirection.

kyleam:
> Testing is going to involve some manipulation of the config, and my reading is that setting values stores them on disk. Do you have helper functions for creating an environment with a disposable config?

There's a patch_config helper you can use to override config values for tests (that doesn't involve touching the disk).  For testing NO_COLOR, you could patch os.environ.  There are examples of both in the codebase.

datalad/interface/ls.py
kyleam: > `self.DATASET` is equivalent to `ansi_colors.DATASET`,

Ah good catch.  I didn't even realize ansi_colors.DATASET existed (though it is already used elsewhere in ls.py).

> though given that the code is `'U'`, perhaps `ansi_colors.UNDERLINE` makes more sense than the indirection.

True, though I don't care enough one way or the other to change the original decision to use DATASET.

yarikoptic:
> Let me know if this seems like the right approach

looks good to me

effigies:
I added unit tests for the other utilities while I was at it, though now I see they were already being thoroughly exercised. Ah well. I can remove them if they seem excessive.

PR #3410, opened 2019-05-15
RF: disable wrapt workaround

yarikoptic:
According to https://github.com/datalad/datalad/pull/3405
we likely no longer hit it.  If we hit it again - it would also be useful
to know, possibly to test devel branch of wrapt (currently 1.11.1-4-g03e90ab)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -295,7 +295,7 @@ script:
   - mkdir -p __testhome__
   - cd __testhome__
   # Run tests
-  - WRAPT_DISABLE_EXTENSIONS=1 http_proxy=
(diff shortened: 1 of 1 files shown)


yarikoptic:
coverage is off and smells like nothing of direct mode was tested
ok, I think (a bit wild) that the oddity with coverage is because 
- I initially posted against `master` branch
- travis picked up .travis.yml as if merged with master where no DIRECT mode testing
- I changed the base, but runs in the matrix already were as of master

but altogether -- seems like our current state + recent wrapt avoid segfaults... I will merge to possibly prove us otherwise

PR #3423, opened 2019-05-16
BF: stop relying on setup.py runtime environment using platform.dist

### LINKED ISSUE #3417: Avoid platform.dist (removed in py3.8)
We use platform.dist() in setup.py and wtf.py (see gh-3406).  The recommendation is to use https://pypi.org/project/distro/ instead, but that of course doesn't work well for setup.py.

yarikoptic:
- colorama Windows specificity is specified via `platform_system=="Windows"`
- always demand `keyring >= 8.0`

And also simplify dependencies where there is no alternative used now:

- tqdm

If older systems with too old setuptools (debian jessie) fail to install from sources,
they might require patching away platform_system=="Windows" and python_version < "3.3"
which is already the case anyways, but while at it TODO (after dinner):
- ~~add conditioning to check for too old setuptools (not sure what version is min to support them yet) being used and avoid adding those qualifiers then. This would avoid patching e.g. for NeuroDebian backports~~ Tried, could succeed, but then it would still might require patching for liblzma so decided not to bother for now.  Note, that setuptools throws SystemExit (not just some kind of an Exception), so it had to be caught, both `install_requires` and values within `extras_require` tuned up

Closes #3417

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,9 +6,6 @@
 #
 # ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
 
-import sys
-import platform
-import os
 from os.path import dirname
 from os.path import join as opj
 from os.path import sep as pathsep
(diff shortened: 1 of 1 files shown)


kyleam:
Thanks, the `requires` changes look good to me.  It would have been nice if the commit message mentioned that wheezy EOL info that you mentioned in gh-3417.

PR #3426, opened 2019-05-16
RF: Drop use of `add` in `addurls(save=False)`

mih:
One step closer to a complete removal of this broken code.

Please give feedback on test bit below. Thx.

https://youtu.be/dmxCA03DJ04?t=90

### DIFF
--- a/datalad/plugin/addurls.py
+++ b/datalad/plugin/addurls.py
@@ -16,7 +16,10 @@ import os
 import re
 import string
 
-from six import string_types
+from six import (
+    string_types,
+    text_type,
+)
 from six.moves.urllib.parse import urlparse
 
 from datalad.dochelpers import exc_str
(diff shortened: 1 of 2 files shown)


datalad/plugin/tests/test_addurls.py
```diff
@@ -379,12 +379,16 @@ def get_annex_commit_counts():
...
+            # MIH: Disabled this test when dropping the use of the API `add`
+            # command and switching to repo.add() calls. Unclear why it was
+            # tested, given that the test was already conditional
+            #
+            # # Ignore this check if we're faking dates because that disables
+            # # batch mode.
+            # if not os.environ.get('DATALAD_FAKE__DATES'):
+            #     # We should have two new commits on the git-annex: one for the
+            #     # added urls and one for the added metadata.
+            #     eq_(n_annex_commits + 2, get_annex_commit_counts())
```
mih: Would be good to understand the intentions of this test.

kyleam: > MIH: Disabled this test when dropping the use of the API `add`
> command and switching to repo.add() calls. Unclear why it was
> tested, given that the test was already conditional

This test was introduced by 9faa07ba8 (ENH: addurls: Reduce the number of metadata commits, 2018-03-26), which in my biased opinion gives a pretty good explanation of what behavior this test is guarding against (i.e. we don't want a commit for each file, because that could be a lot of commits).

I'm not sure I understand the conditional comment.  That's just relevant for faking dates: 4d7ada267 (RF: annexrepo: Disable batching when fake dates are enabled, 2018-04-13).

Anyway, I'll push a fixup to your commit that should make tests pass.

kyleam:
The failures in the latest run are the beloved `test__version__` failures.  (I'll adjust master directly.)

mih:
Thx again @kyleam !

Issue #3192, opened 2019-03-03
Restructure source tree


mih:
I believe we agree that the source tree is a bit out of sync with the internal (dependency) structure of the code. I would like to propose a restructuring as follows (lumping together stuff that is scattered across -core and the -revolution extension for now):

- [ ] move core dataset and repo classes into `dataset/` (or rather `core/dataset/`)
  - [ ] it might make sense to split Git/AnnexRepo into a core and an extended interface also (via inheritance), the number of rarely used methods is substantial ATM
- [ ] `core/` -- in the spirit of #2765, introduce `core/` for a minimal set of commands that serve as the foundation for others and receive special scrutiny regard API composition and changes.
  - [ ]  `local/` -- essential commands for working with a single local dataset(-hierarchy)
    - [ ] `status`
    - [ ] `diff`
    - [ ] `save`
    - [ ] `create`
    - [ ] `run`
    - [ ] `metadata` (anything that is concerned with *accessing* the internal metadata representation)
    - [ ] `extract-metadata` (coming via -revolution, stable interface to execute extractors)
  - [ ] `distributed/` -- key commands needed for communication with other dataset instances
    - [ ]  `clone`
    - [ ]  `get`
    - [ ]  `update`
    - [ ]  `publish`
    - [ ]  `siblings`

In addition, there is a second set of categories that are identical by name, but are treated more like applications instead of foundation with the goal of having them move faster (or at all) in order to expose features and/or adapt to changes in the software/service ecosystem
- [ ] `local/`
  - [ ] `rerun`
  - [ ] `run-procedure`
  - [ ] `add*`
  - [ ] `download-url`
  - [ ] `ls`
  - [ ] `subdatasets`
    - [ ] `aggregate-metadata` (not a core command, there could be multiple ways of aggregation, as long as the result is a data structure that the core `metadata` command can access)
    - [ ] `search`
- [ ] `distributed/`
    - [ ]  `install` (placed here and not in core, because we have plans for complex features that could be difficult to keep stable long-term, hence this could be seen as an "application" that installs dataset via all kinds of ways)
    - [ ]  `drop`
    - [ ] `create-sibling*`
    - [ ] `export-*`

Other (more "leafy") commands could either be injected in these categories, or populate to additional ones, to possibly better indicate (and enforce?) their (in)dependence on/from others:
- remove
- uninstall
- clean

Given the target directories do not conflict with anything, we could approach such a transitions slowly and keep import shims in place for any length of time to ease the pain.

[...]

PR #3429, opened 2019-05-17
`distribution.subdatasets` -> `local.subdatasets`

mih:
This PR implements the transition of the `subdatasets` command to the "local" API section #3192 . It aligns arguments, behavior, and result structure with the rest "core.local" API.

Python API breakage is caused by the necessary introduction of positional `PATH` arguments (like the other commands have) in conjunction with the requirement for uniform argument order.

- [x] multi path matching with `--contains`
- [x] supports path-constrained queries like `status/diff/save`
- [x] when invoked on the cmdline with no args, reports only subdatasets underneath `curdir`
- [x] result content now aligned with `status/diff/save`, e.g. `revision` -> `gitshasum` (but `revision` is kept as a duplicate for compatibility)
- [x] no custom parser for subdatasets anymore
- [x] no custom parser for `.gitmodules` anymore
- [x] no GitPy functionality used anymore
- [x] move to `local/subdatasets.py`


datalad/distribution/subdatasets.py
```diff
@@ -56,67 +66,68 @@
...
+    db, _ = _parse_gitconfig_dump(out, {}, None, True)
     mods = {}
-    for sec in parser.sections():
-        try:
-            modpath = parser.get(sec, 'path')
-        except Exception:
-            lgr.debug("Failed to get '%s.path', skipping section", sec)
+    for k, v in iteritems(db):
+        if not k.startswith('submodule.'):
+            # we don't know what this is
```
yarikoptic: I think it would be useful to leave some debug message at least if something unexpected discovered.  could as well be a warning since AFAIK and according to `man gitmodules` there should be no other sections there.

datalad/distribution/subdatasets.py
```diff
@@ -56,67 +66,68 @@
...
             continue
-        if not modpath or not sec.startswith('submodule '):
+        k_l = k.split('.')
```
yarikoptic: suggestion actually to replace 4 lines (can't be done in web ui AFAIK)
```suggestion
        # module name could contain '.' - need to split from the left to get "submodule" (not used)
        # and then from the right to get field
        mod_name, field = k.split('.', 1)[1].rsplit('.', 1)
        mod = mods.get(mod_name, {})
        mod[mod_name] = field
```
I find it difficult to read all the `[X]`s.

<details>
<summary>git is ok with '.' in submodule names so middle parts I guess constitute the mod_name, not just [1] </summary>

```shell
$> datalad create -d . my.subds
[INFO   ] Creating a new annex repo at /tmp/testds/my.subds 
create(ok): my.subds (dataset)                                                                                         
action summary:
  add (ok: 2)
  create (ok: 1)
  save (ok: 1)
(dev) 1 13959 [1].....................................:Fri 17 May 2019 04:31:28 PM EDT:.
(git-annex)hopa:/tmp/testds[master]git-annex
$> cat .gitmodules 
[submodule "my.subds"]
	path = my.subds
	url = ./my.subds
	datalad-id = bf34882e-78e2-11e9-a650-8019340ce7f2
(dev) 1 13960 [1].....................................:Fri 17 May 2019 04:31:32 PM EDT:.
(git-annex)hopa:/tmp/testds[master]git-annex
$> 'git' 'config' '-z' '-l' '--file' '.gitmodules' 
submodule.my.subds.path
(dev) 1 13961 [1].....................................:Fri 17 May 2019 04:31:54 PM EDT:.
(git-annex)hopa:/tmp/testds[master]git-annex
$> git submodule
 6e0a46306893a150df47f7de130fbe74931723ca my.subds (heads/master)
```
</details>

datalad/distribution/tests/test_subdataset.py
```diff
@@ -166,6 +166,15 @@ def test_get_subdatasets(path):
...
                        result_xfm='paths'),
         [])
+    eq_(subdatasets(
+        ds, recursive=True,
+        contains=[target_sub, 'sub dataset1/2'],
+        result_xfm='relpaths'), [
+        'sub dataset1',
+        'sub dataset1/2',
+        'sub dataset1/sub sub dataset1',
+        'sub dataset1/sub sub dataset1/subm 1',
```
yarikoptic: please add a test with `.`s in the modname

datalad/distribution/subdatasets.py
mih: Done

datalad/distribution/subdatasets.py
mih: Done in a different way.

datalad/distribution/tests/test_subdataset.py
mih: Done.

mih:
This is now ready for review.

datalad/local/subdatasets.py
```diff
@@ -0,0 +1,438 @@
...
+            recursive=False,
+            recursion_limit=None,
+            contains=None,
+            bottomup=False,
+            set_property=None,
+            delete_property=None):
+        # no constraints given -> query subdatasets under curdir
+        if path is None and dataset is None:
+            path = os.curdir
+        paths = [rev_resolve_path(p, dataset) for p in assure_list(path)] \
```
mih: Could be simplified once #3435 is merged.

mih:
Remaining test failure is the already-known neuroimaging extension test run failure, due to a missing config procedure.

datalad/distribution/subdatasets.py
```diff
@@ -6,364 +6,8 @@
...
-                    refds_path):
-                yield r
-        if bottomup and \
-                (fulfilled is None or
-                 GitRepo.is_valid_repo(sm['path']) == fulfilled):
-            yield subdsres
-    if parser is not None:
-        # release parser lock manually, auto-cleanup is not reliable in PY3
-        parser.release()
+# Import command class to ease 3rd-party transitions
```
kyleam: Should we issue a warning (probably with warnings.warn() rather than logging.warn()) when this module is imported?

datalad/local/subdatasets.py
```diff
@@ -0,0 +1,438 @@
...
+            dataset=None,
+            fulfilled=None,
+            recursive=False,
+            recursion_limit=None,
+            contains=None,
+            bottomup=False,
+            set_property=None,
+            delete_property=None):
+        # no constraints given -> query subdatasets under curdir
+        if path is None and dataset is None:
```
kyleam: From the command line, `path` won't ever come in as None.

This means that the test

```python
with chpwd(text_type(ds.pathobj / 'subdir')):
    # imitate cmdline invocation w/ no dataset argument
    # -> curdir limits the query, when no info is given
    eq_(subdatasets(dataset=None,
                    path=None,
                    recursive=True,
                    result_xfm='paths'),
        [text_type(ds.pathobj / dots)]
    )
```

doesn't actually match the command-line behavior:

```
% datalad subdatasets -r
subdataset(ok): sub dataset1 (dataset)
subdataset(ok): sub dataset1/2 (dataset)
subdataset(ok): sub dataset1/sub sub dataset1 (dataset)
subdataset(ok): sub dataset1/sub sub dataset1/2 (dataset)
subdataset(ok): sub dataset1/sub sub dataset1/subm 1 (dataset)
subdataset(ok): sub dataset1/subm 1 (dataset)
subdataset(ok): subdir/.lots.of.dots. (dataset)
action summary:
  subdataset (ok: 7)
```

Perhaps this should be `if not path ...` instead.

datalad/local/subdatasets.py
```diff
@@ -0,0 +1,438 @@
...
+                            ]
+                    )
+                except CommandError as e:
+                    yield get_status_dict(
+                        'subdataset',
+                        status='impossible',
+                        message=(
+                            "Deleting subdataset property '%s' failed for "
+                            "subdataset '%s', possibly did "
+                            "not exist",
```
kyleam: Given that you bind the exception to `e` above, did you intend to include `e` in the message like you do for the replace exception below?

yarikoptic:
I did upload fresh -neuroimaging with the fix, and restarted one sample run which was failing -- https://travis-ci.org/datalad/datalad/builds/534685969?utm_source=github_status&utm_medium=notification  -- all green now back then a few commits back before ci skips

datalad/distribution/subdatasets.py
mih: I don't think it is needed. Even in our own code base we did not import it directly.

datalad/local/subdatasets.py
mih: Indeed. Fixed.

datalad/distribution/subdatasets.py
kyleam: OK.

datalad/local/subdatasets.py
mih: I did not. The exception is rather verbose, but the only meaningful bit is exit code 5, which is a bit pointless and decoded in the actual message. I removed `as e`. Thx.

mih:
All green. @yarikoptic didnt update his review, but I think I successfully addressed his requests.

Issue #3282, opened 2019-03-29
Code/semantics homogenization of dataset/path argument resolution


mih:
This is connected to various issues (#3230, #2589, #3242), but is also focused on the flow with which command should process and resolve arguments. We should have a standard, documented way (if possible) to reduce the heterogeneity across commands that contributes to the issues listed above.

Here is a pattern proposal (not a description of the status quo):

### Dataset resolution

Determine which dataset to operate on (if needed). All relevant commands get a `dataset` argument, and we can use `require_dataset()` to get a `ds` Dataset instance. It is important to maintain the pristine `dataset` argument value for path resolution (below), or in case further commands need to be called, so that argument semantics stay constant.
Besides normal dataset path specification like absolute/relative paths, this **could** support special symbols like `^` and/or `^^` to indicate the root of the dataset immediately containing CWD, or the topmost dataset respectively.

### Path argument resolution

The goal is to turn all arguments into absolute paths. The challenge is to make the semantics sensible for both cmdline and Python API.
In the cmdline the CWD is an important and highly visible anchor. Interpretation of nifty things like `--dataset ^^ --path here/below` is relatively clear (the topmost dataset, and $PWD/here/below).

The same things in Python with bound methods: `topmost.cmd('here/below')` is not at all saying the same (rather path relative to the dataset).

We also support `cmd(dataset='^^', path='here/below')`, which could mean any of the those interpretations from my POV.

Proposal:

1. if the `dataset` argument (not the resolved `ds`) is a `Dataset` instance: any relative path argument is interpreted relative to that dataset (this covers the bound-method case).
2. If the `dataset` argument is anything else (`None`, or any path specification, including magic symbols): any relative path is interpreted relative to CWD
3. There is no difference in semantics between "explicit" relative paths (`./here/below`) and relative paths (`here/below`)

yarikoptic:
interoperable agreement achieved despite the lack of the magical pepper liquid.

mih:
Awesome!

Issue #2589, opened 2018-05-30
Dataset methods don't resolve "./path" to current dataset


kyleam:
I expected "./path" to be resolved when passed as a path to a dataset method.  "./" is unnecessary, but it's nice in terms of being able to `os.path.join(dirname, path)`, where dirname may be `os.path.curdir`.

But this fails.  As an example:

```python
#!/usr/bin/env python3

import os.path as op
import tempfile
from datalad.api import Dataset

repo = tempfile.mkdtemp()
ds = Dataset(repo).create()

with open(op.join(repo, "foo"), "w") as fh:
    fh.write("ok")
ds.add("./foo")
```

```bash
% /tmp/scratch.py
[INFO   ] Creating a new annex repo at /tmp/tmpk5_0ksad
[WARNING] path not associated with reference dataset
[WARNING] path not associated with any dataset [add(/tmp/foo)]
Traceback (most recent call last):
  File "/tmp/scratch.py", line 15, in <module>
    ds.add("./foo")
  File "/home/kyle/src/python/datalad/datalad/distribution/dataset.py", line 444, in apply_func
    return f(**kwargs)
  File "/home/kyle/src/python/datalad/datalad/interface/utils.py", line 470, in eval_func
    return return_func(generator_func)(*args, **kwargs)
  File "/home/kyle/src/python/datalad/datalad/interface/utils.py", line 458, in return_func
    results = list(results)
  File "/home/kyle/src/python/datalad/datalad/interface/utils.py", line 446, in generator_func
    msg="Command did not complete successfully")
datalad.support.exceptions.IncompleteResultsError: Command did not complete successfully [{'message': 'path not associated with reference dataset', 'status': 'impossible', 'action': 'add', 'refds': '/tmp/tmpk5_0ksad'}, {'status': 'impossible', 'raw_input': True, 'action': 'add', 'orig_request': './foo', 'message': 'path not associated with any dataset', 'state': 'absent', 'path': '/tmp/foo', 'refds': '/tmp/tmpk5_0ksad'}]
```

The path is resolved to the current directory rather than the dataset directory.

yarikoptic:
I believe that it is a feature... If dataset is known and relative path, but not starting with curdir or pardir, then from top of the dataset, otherwise from current directory.

mih:
Yes, it is a feature.

kyleam:
Eh, a feature that makes "this" different than "./this"? IMO there shouldn't be a magical distinction between the two.

[...]

PR #3435, opened 2019-05-20
New "standard" path resolution logic

mih:
Talked about it multiple times: here it is.

Any absolute `path` given to a command is used as such. Any relative `path`is interpreted as being
relative to the CWD, **unless** an also provided `dataset` argument is a `Dataset` instance (not just a 
path to or inside a dataset).

I briefly looked into implementing this for all commands, but that seems to be a bottomless pit -- to expensive for me and for now. Instead, I implemented it for `rev_resolve_path()` that is used by all commands that already comply with the structure proposed in #3192.

- [x] implement logic change
- [x] adjust tests
- [x] add ability to process multiple paths without re-localization a dataset each and every time (immediately benefits https://github.com/datalad/datalad/pull/3429)
- [x] turn `EnsureDataset` into a NOOP with documentation purpose only. This constraint class auto instantiates Dataset objects based on command line arguments (paths), essentially disabling the ability to detect and act on the key distinction of the decision logic above.
- [x] @kyleam please double-check https://github.com/datalad/datalad/pull/3435/commits/971620ec8ddae9fe31f193e83a60a88debe40ad3
   km: Looks fine (and outputs come in as absolute paths, so string/instance for dataset would be the same anyway).

### DIFF
--- a/datalad/core/local/tests/test_save.py
+++ b/datalad/core/local/tests/test_save.py
@@ -222,7 +222,7 @@ def test_symlinked_relpath(path):
     with chpwd(op.join(dspath, 'd')):
         save(dataset=ds.path,
              message="committing",
-             path=op.join(op.curdir, "mike2"))
+             path="mike2")
 
         later = op.join(op.pardir, "later")
         ds.repo.add(later, git=True)
--- a/datalad/core/local/tests/test_save.py
+++ b/datalad/core/local/tests/test_save.py
@@ -216,14 +216,17 @@ def test_symlinked_relpath(path):
         ds.repo.add("mike1", git=True)
         ds.save(message="committing", path="./mike1")
 
-    # Let's also do in subdirectory
+    # Let's also do in subdirectory as CWD, check that relative path
+    # given to a plain command (not dataset method) are treated as
+    # relative to CWD
     with chpwd(op.join(dspath, 'd')):
-        ds.save(
-            message="committing", path=op.join(op.curdir, "mike2"))
+        save(dataset=ds.path,
+             message="committing",
+             path=op.join(op.curdir, "mike2"))
 
         later = op.join(op.pardir, "later")
         ds.repo.add(later, git=True)
-        ds.save(message="committing", path=later)
+        save(dataset=ds.path, message="committing", path=later)
 
     assert_repo_status(dspath)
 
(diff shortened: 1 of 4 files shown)


datalad/core/local/tests/test_save.py
```diff
@@ -216,14 +216,17 @@ def test_symlinked_relpath(path):
...
-    # Let's also do in subdirectory
+    # Let's also do in subdirectory as CWD, check that relative path
+    # given to a plain command (not dataset method) are treated as
+    # relative to CWD
     with chpwd(op.join(dspath, 'd')):
-        ds.save(
-            message="committing", path=op.join(op.curdir, "mike2"))
+        save(dataset=ds.path,
+             message="committing",
+             path=op.join(op.curdir, "mike2"))
```
bpoldrack: Shouldn't `path` be just "mike2" here to prove the point?

mih: historically, yes. will change.

mih:
All tests pass -- minus the neuroimaging procedure issue that is not specific to this PR.

yarikoptic:
> Any absolute `path` given to a command is used as such. Any relative `path`is interpreted as being
> relative to the CWD, **unless** an also provided `dataset` argument is a `Dataset` instance (not just a
> path to or inside a dataset).

Just to make sure I/we get it right.  It means, that in the command line it will always be relative to the CWD, so for any operation on relative paths users would need to `cd` to that directory, correct?  If so, in my experience it is an ok compromise -- I don't remember relying on relative paths with specification of the dataset (as the ways to achive `cd` effect) much since paths completion is not there then.

> I briefly looked into implementing this for all commands, but that seems to be a bottomless pit -- to expensive for me and for now.

I was hoping that it should be doable due to all the centralization (AnnotatePaths?) of paths handling... I think we really wouldn't want to treat paths differently or is that already the case?

datalad/core/local/tests/test_save.py
mih: Done

mih:
> Just to make sure I/we get it right. It means, that in the command line it will always be relative to the CWD, so for any operation on relative paths users would need to `cd` to that directory, correct?

Not sure I can follow. If you have a relative path at the cmdline, it is typically relative to CWD (where else would you have it from?). So there is no CDing needed in general. If you have a path relative to any other reference then yes, as there is no way we could know which one that would be. We used to make the distinction between `./here` and `here` the former being relative to CWD and latter relative to a given datasets, but only if there was any, relative to CWD otherwise -- but we decided that this is overcomplicated and prone to failure.

> I was hoping that it should be doable due to all the centralization (AnnotatePaths?) of paths handling... I think we really wouldn't want to treat paths differently or is that already the case?

I cannot comprehensively describe the status quo. I have no plans to make *any* changes to annotate paths other then a final `git rm`. If this has to be done in one go, it needs to wait to all commands are reimplemented without it.

kyleam:
For posterity:  Related to #3282 and #2589.

mih:
Let's go for incremental improvement now, instead of ultimate awesomeness later.

PR #3439, opened 2019-05-20
Use "distro" for .distribution for python >= 3.8

### LINKED ISSUE #3431: Avoid platform.dist (removed in py3.8) in wtf
This is split off from gh-3417.  gh-3423 fixed the setup.py part, so this is a dedicated issue for wtf's use of platform.dist().

yarikoptic:
Closes #3431 

Decided to depend on distro for 3.8 and beyond for now so we have a complete (albeit slow) fix

TODOs
- [ ] remove TEMP enabling tests on travis for 3.8-dev

### DIFF
--- a/datalad/plugin/wtf.py
+++ b/datalad/plugin/wtf.py
@@ -108,12 +108,33 @@ def _describe_annex():
 def _describe_system():
     import platform as pl
     from datalad import get_encoding_info
+
+    if hasattr(pl, 'dist'):
+        dist = pl.dist()
+    else:
+        # Python 3.8 removed .dist but recommended "distro" is slow, so we
+        # try it only if needed
+        try:
+            import distro
+            dist = distro.linux_distribution(full_distribution_name=False)
+        except ImportError:
+            lgr.info(
+                "Please install 'distro' package to obtain distribution information"
+            )
+            dist = tuple()
+        except Exception as exc:
+            lgr.warning(
+                "No distribution information will be provided since 'distro' "
+                "fails to import/run: %s", exc_str(exc)
+            )
+            dist = tuple()
+
     return {
         'type': os.name,
         'name': pl.system(),
         'release': pl.release(),
         'version': pl.version(),
-        'distribution': ' '.join([_t2s(pl.dist()),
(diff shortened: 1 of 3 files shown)


datalad/plugin/wtf.py
```diff
@@ -104,16 +104,27 @@ def _describe_annex():
...
+
+    if hasattr(pl, 'dist'):
+        dist = pl.dist()
+    else:
+        # Python 3.8 removed .dist but recommended "distro" is slow, so we
+        # try it only if needed
+        try:
+            import distro
+            dist = distro.linux_distribution(full_distribution_name=False)
+        except Exception as exc:
```
kyleam: nitpick 1: I'd prefer to narrow the catch to an import error.
nitpick 2: No need to bind `exc`.

Also, should we log a warning here?

yarikoptic:
f.ck, wrong base, will rebase ... actually I will meld it with another quick ENH for WTF.  Unless there will be major complaint, let's keep in the same PR

kyleam:
> f.ck, wrong base, will rebase ...

I changed the base on the PR.  You don't need to rebase locally.

kyleam:
But you do need to fix up the Travis run for 3.8 to use xenial.  (I was about to push, but I won't in case you're making changes on your end.)

yarikoptic:
oh, I just got confused with those commits listed, couldn't make sense... thanks!

need also to fix up travis to specify correct dist IIRC.

datalad/plugin/wtf.py
yarikoptic: I smile when I see some diagnostic command to provide information about a problem blows itself up due to some other issue...  So I would prefer to capture `Exception` here as well, since I do not trust (yet) `distro` to not blow.  So now doing dedicated captures... 
I again pushed before getting rev-diff quickly, sorry... I will get there

kyleam: What you went with looks fine to me.  Thanks.

kyleam:
Merged (dropping the temporary commit).

yarikoptic:
Thank you!

PR #3440, opened 2019-05-20
ENH: wtf -S to specify which sections to query/display (by default -- all)

yarikoptic:
`-S dependencies`   is probably one of the most useful, but unfortunately it is not including concise versioning of datalad and datalad extensions, which are all heavy sections...

### DIFF
--- a/datalad/plugin/wtf.py
+++ b/datalad/plugin/wtf.py
@@ -368,7 +368,7 @@ class WTF(Interface):
                 'pyperclip', msg="It is needed to be able to use clipboard")
             import pyperclip
             report = _render_report(res)
-            pyperclip.copy(report)
+            pyperclip.copy(assure_bytes(report))
             ui.message("WTF information of length %s copied to clipboard"
                        % len(report))
         yield res
--- a/datalad/plugin/wtf.py
+++ b/datalad/plugin/wtf.py
@@ -300,7 +300,7 @@ class WTF(Interface):
             metavar="SECTION",
             constraints=EnsureChoice(*sorted(SECTION_CALLABLES)) | EnsureNone(),
             doc="""section to include.  If not set, all sections.
-            [CMD: This option can be given multiple times. CMD]."""),
+            [CMD: This option can be given multiple times. CMD]"""),
         decor=Parameter(
             args=("-D", "--decor"),
             constraints=EnsureChoice('html_details') | EnsureNone(),
--- a/datalad/plugin/wtf.py
+++ b/datalad/plugin/wtf.py
@@ -14,6 +14,7 @@ import logging
 import os
 import os.path as op
 from functools import partial
+from collections import OrderedDict
 
 from datalad.interface.base import Interface
 from datalad.interface.base import build_doc
@@ -237,6 +238,13 @@ def _describe_dataset(ds, sensitive):
         return {"invalid": exc_str(e)}
 
 
+def _describe_location(res):
+    return {
+        'path': res['path'],
+        'type': res['type'],
+    }
+
+
 # Actuall callables for WTF. If None -- should be bound later since depend on
 # the context
 SECTION_CALLABLES = {
@@ -245,6 +253,7 @@ SECTION_CALLABLES = {
     'system': _describe_system,
     'environment': _describe_environment,
     'configuration': None,
+    'location': None,
     'extensions': _describe_extensions,
     'metadata_extractors': _describe_metadata_extractors,
     'dependencies': _describe_dependencies,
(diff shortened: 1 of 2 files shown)


yarikoptic:
thanks @kyleam , did that with a few other minor tune ups (eventually I will learn how to use `git range-diff` ;)) (I expected `git range-diff 'HEAD^..HEAD@{2}' 'HEAD^..HEAD@{0}'` to be it but it wasn't ;))

kyleam:
> eventually I will learn how to use git range-diff ;)) (I expected git range-diff 'HEAD^..HEAD@{2}' 'HEAD^..HEAD@{0}' to be it but it wasn't ;))

:]

Something like the above could work, but of course it's completely dependent on the state of your HEAD reflog and that could easily not line up.   If you did it pre-push (and the root was the same for both), it could be as simple as `git range-diff yarikoptic/enh-wtf-S...enh-wtf-S` (that's three dots).

yarikoptic:
grr... couldn't stop. Please re-review again.  Now my primary use case will be 

    datalad wtf -S datalad -S extensions -S dependencies -D html_details -c

to just then paste e.g. here

<details><summary>DataLad 0.11.4.dev107 WTF (datalad, extensions, dependencies)</summary>

# WTF
## datalad 
  - full_version: 0.11.4.dev107-g03da-dirty
  - version: 0.11.4.dev107
## extensions 
## dependencies 
  - tqdm: 4.28.1
  - git: git
  - wrapt: 1.11.1
  - boto: 2.44.0
  - keyrings.alt: 3.1
  - six: 1.12.0
  - keyring: 16.0.0
  - cmd:git: 2.20.1
  - iso8601: 0.1.11
  - mutagen: 1.40.0
  - exifread: 2.1.2
  - gitdb: 2.0.3
  - msgpack: 0.5.6
  - cmd:annex: 7.20190322+git133-g59922f1f4-1~ndall+1
  - requests: 2.21.0
  - cmd:system-ssh: 7.9p1
  - cmd:system-git: 2.21.0.593.g511ec345e18
  - humanize: 0.5.1
  - appdirs: 1.4.3
  - scrapy: 1.5.1
</details>
        
unless tests fail -- I am done with it.  Feel free to tune up to your liking or drop last commits ;-)

kyleam:
As I said on the call, I'm in favor of the `action="append"` change.  I didn't look closely at the updates, but adding an option for the details-formatted output is fine with me.

datalad/plugin/wtf.py
```diff
@@ -262,6 +293,20 @@ class WTF(Interface):
...
+            dest='sections',
+            metavar="SECTION",
+            constraints=EnsureChoice(*sorted(SECTION_CALLABLES)) | EnsureNone(),
+            doc="""section to include.  If not set, all sections. This option
+            can be given more than once to include multiple sections."""),
+        decor=Parameter(
+            args=("-D", "--decor"),
+            constraints=EnsureChoice('html_details') | EnsureNone(),
+            doc="""decoration around the rendering to facilitate embedding into
+            issues etc, e.g. use 'html_details_md' for posting collapsable entry
```
kyleam: 'html_details_md' doesn't match the above value.

datalad/plugin/wtf.py
```diff
@@ -262,6 +293,20 @@ class WTF(Interface):
...
             information (credentials, names, etc.).  If 'some', the fields
             which are known to be sensitive will still be masked out"""),
+        sections=Parameter(
+            args=("-S", "--section"),
+            action='append',
+            dest='sections',
+            metavar="SECTION",
+            constraints=EnsureChoice(*sorted(SECTION_CALLABLES)) | EnsureNone(),
+            doc="""section to include.  If not set, all sections. This option
+            can be given more than once to include multiple sections."""),
```
kyleam: This would benefit from `[CMD]` and `[PY]` markers.

datalad/plugin/wtf.py
```diff
@@ -297,35 +342,42 @@ def __call__(dataset=None, sensitive=None, clipboard=None):
...
+        for s in sections:
+            infos[s] = section_callables[s]()
 
         if clipboard:
             external_versions.check(
                 'pyperclip', msg="It is needed to be able to use clipboard")
             import pyperclip
             report = _render_report(res)
-            pyperclip.copy(report)
+            pyperclip.copy(assure_bytes(report))
```
kyleam: Our tastes regarding collapsing unrelated changes into commits differ, so I try not to remark upon it too much, but this really should get its own commit!

datalad/plugin/wtf.py
yarikoptic: thanks!

datalad/plugin/wtf.py
yarikoptic: did for `[CMD` using similar phrasing as found in others

datalad/plugin/wtf.py
yarikoptic: did that too ;-)

PR #3458, opened 2019-05-24
Enh bring gc back

### LINKED ISSUE #3457: Bring back git gc
ATM we explicitly disable git's gc during calls to git/annex since GitPython's state might get outdated and we did experience problems in the past.  But that often, largely after a sweep of git annex actions, leaves **huge** .git/objects behind and then mandates either running some `git` commands manually to get gc triggered or an explicit run to `gc`.  Example after a one large `datalad save` call:
```shell
$ du -scm .git/objects
1978    .git/objects
1978    total

$ git gc --prune=now
Counting objects: 524428, done.
Delta compression using up to 40 threads.
Compressing objects: 100% (520901/520901), done.
Writing objects: 100% (524428/524428), done.
Total 524428 (delta 50448), reused 18895 (delta 1701)
Removing duplicate objects: 100% (256/256), done.
Checking connectivity: 524428, done.

$ du -scm .git/objects
71      .git/objects
71      total
```

Especially if GitPython would be gone, or may be even without it (could have been fixed up in GitPython to react upon gc?) we should consider dropping disabling of automatic gc.  I will submit a test PR to see if things still get crashing (unable to find objects etc) ;-)

yarikoptic:
Closes #3457 if we decide to go for it

### DIFF
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -403,7 +403,7 @@ def Repo(*args, **kwargs):
     # TODO: This probably doesn't work as intended (or at least not as
     #       consistently as intended). gitpy.Repo could be instantiated by
     #       classmethods Repo.init or Repo.clone_from. In these cases 'odbt'
-    #       would be needed as a paramter to these methods instead of the
+    #       would be needed as a parameter to these methods instead of the
     #       constructor.
     if 'odbt' not in kwargs:
         kwargs['odbt'] = default_git_odbt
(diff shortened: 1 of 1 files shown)


kyleam:
Ah, but doing that globally is not a good idea because we don't know which dataset(s) to run it in.

yarikoptic:
So only neuroimaging expected failure.. hm... Will rerun it a few times later when after neuroimaging

yarikoptic:
appveyor failure might be "legit"... restarted

yarikoptic:
I have reran travis a few times and it seems to show no bad signs... should we merge?

kyleam:
> should we merge?

I think so.  Easy enough to revert if the issues pop up again.

yarikoptic:
ok, let's see what would happen ;)

PR #3517, opened 2019-07-08
Temp test devel annex

yarikoptic:
should be the fixed up bleeding edge 7.20190708+git9-gfa3524b95-1~ndall+1


kyleam:
Looking good.

yarikoptic:
But what is going on with Travis reporting? May be we need to relink or something?

kyleam:
> But what is going on with Travis reporting? May be we need to relink or something?

I see Travis listed in the checks above, but I did notice it was missing from #3516.  Is that what you're referring to or are you not seeing it on this PR either?

yarikoptic:
I could have sworn that it was not here either, but now I see it reporting failure (-containers, known). So all good -- closing and uploading to neurodebian proper

PR #3518, opened 2019-07-09
UX: Don't log stderr as warning when expect_stderr was True

### LINKED ISSUE #3515: Stop unconditional warning on stderr/failure in run_annex_command_json
I am running it like this:

```
repo._run_annex_command_json(
        cmd,
        files=paths,
        git_options=git_options,
        opts=args,
        expect_stderr=True,
        expect_fail=True,
    )
```

but still:

```
[WARNING] Running fsck resulted in stderr output:   ** Based on the location log, one.txt
  ** was expected to be present, but its content is missing.
  ** Based on the location log, subdir/two
  ** was expected to be present, but its content is missing.
git-annex: fsck: 2 failed
```

I think it should inspect `kwargs` to test for `expect_fail=True` and use `lgr.debug()` instead of `lgr.warning()` in such a case.

bpoldrack:
Fixes #3515

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -2124,7 +2124,9 @@ class AnnexRepo(GitRepo, RepoInterface):
             if e.stderr:
                 # else just warn about present errors
                 shorten = lambda x: x[:1000] + '...' if len(x) > 1000 else x
-                lgr.warning(
+
+                _log = lgr.debug if kwargs.get('expect_stderr', False) else lgr.warning
+                _log(
                     "Running %s resulted in stderr output: %s",
                     command, shorten(e.stderr)
                 )
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -2125,7 +2125,7 @@ class AnnexRepo(GitRepo, RepoInterface):
                 # else just warn about present errors
                 shorten = lambda x: x[:1000] + '...' if len(x) > 1000 else x
 
-                _log = lgr.debug if kwargs.get('expect_stderr', False) else lgr.warning
+                _log = lgr.debug if kwargs.get('expect_fail', False) else lgr.warning
                 _log(
                     "Running %s resulted in stderr output: %s",
                     command, shorten(e.stderr)
--- a/datalad/support/tests/test_annexrepo.py
+++ b/datalad/support/tests/test_annexrepo.py
@@ -2254,7 +2254,3 @@ def test_annex_cmd_expect_fail(path):
         repo._run_annex_command_json('add', ['non-existing'], expect_fail=False)
         # message shows up at WARNING level
         assert_re_in(r".*\[WARNING\][^[]*git-annex: add: 1 failed", cml.out, flags=DOTALL)
-    with swallow_logs(logging.DEBUG) as cml:
-        repo._run_annex_command_json('add', ['non-existing'], expect_fail=None)
-        # message shows up at WARNING level
-        assert_re_in(r".*\[WARNING\][^[]*git-annex: add: 1 failed", cml.out, flags=DOTALL)
--- a/datalad/support/tests/test_annexrepo.py
+++ b/datalad/support/tests/test_annexrepo.py
@@ -2235,3 +2235,26 @@ def test_ro_operations(path):
 
     # just check that all is good again
     repo2.repo_info()
+
+
+@with_tempfile
+def test_annex_cmd_expect_stderr(path):
+    # test, that log message about stderr is logged on debug level if expect_stderr was set to True,
+    # warning level else
+
+    from re import DOTALL
+
+    repo = AnnexRepo(path)
+
+    with swallow_logs(logging.DEBUG) as cml:
+        repo._run_annex_command_json('add', ['non-existing'], expect_stderr=True)
+        # message shows up at DEBUG level:
+        assert_re_in(r".*\[DEBUG\][^[]*git-annex: add: 1 failed", cml.out, flags=DOTALL)
+    with swallow_logs(logging.DEBUG) as cml:
+        repo._run_annex_command_json('add', ['non-existing'], expect_stderr=False)
+        # message shows up at WARNING level
+        assert_re_in(r".*\[WARNING\][^[]*git-annex: add: 1 failed", cml.out, flags=DOTALL)
+    with swallow_logs(logging.DEBUG) as cml:
+        repo._run_annex_command_json('add', ['non-existing'], expect_stderr=None)
+        # message shows up at WARNING level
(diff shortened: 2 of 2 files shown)


datalad/support/annexrepo.py
```diff
@@ -2124,7 +2124,9 @@ def _run_annex_command_json(self, command,
             if e.stderr:
                 # else just warn about present errors
                 shorten = lambda x: x[:1000] + '...' if len(x) > 1000 else x
-                lgr.warning(
+
+                _log = lgr.debug if kwargs.get('expect_stderr', False) else lgr.warning
```
kyleam: In deciding to log at 9 or 11, `Runner` looks at `expect_fail` rather than `expect_stderr` in the case of a non-zero exit status.  Given that we're processing a non-zero exit (we're under `except CommandError`), should we look at `expect_fail` rather than `expect_stderr`?

datalad/support/tests/test_annexrepo.py
```diff
@@ -2235,3 +2235,26 @@ def test_ro_operations(path):
...
+    # warning level else
+
+    from re import DOTALL
+
+    repo = AnnexRepo(path)
+
+    with swallow_logs(logging.DEBUG) as cml:
+        repo._run_annex_command_json('add', ['non-existing'], expect_stderr=True)
+        # message shows up at DEBUG level:
+        assert_re_in(r".*\[DEBUG\][^[]*git-annex: add: 1 failed", cml.out, flags=DOTALL)
```
kyleam: Fine as is, but the helper cml.assert_logged() could also be used here, which I find slightly more readable but always forget about myself.

datalad/support/annexrepo.py
bpoldrack: You're absolutely right. Will change.

datalad/support/tests/test_annexrepo.py
bpoldrack: It would be nice if it could be used, but it can't ATM. Requires a rewrite of assert_logged, which isn't flexible enough for that. It will prepend the message I'm looking for with a fixed prefix of for the level. But this prefix doesn't account for other things in the captured log before that. It would assume the captured log actually starts with `[DEBUG]` for example. But that doesn't work if there are other things logged at higher level before. In short: I tried, but gave up on it, since I was already spending more time than I wanted. Only tackled this one, because I thought: "Well, one-liner. Just spend those 5 minutes." Always turns out to be wrong ;-)

datalad/support/annexrepo.py
bpoldrack: Although - it is somewhat confusing how exactly we want to have it. `expect_stderr` and `expect_fail` are different, independent parameters. So, what should be the meaning of combining them? If I declare to expect a non-zero exit, but not expecting something on stderr shouldn't we still have it logged as a warning? I guess, we are not really consistent in the interpretation of those.
However, I agree and would like to go with the simple approach: really consider `expect_stderr` only for zero exit.

bpoldrack:
FTR: Failure in Travis is about singularity hub being down.

datalad/support/tests/test_annexrepo.py
kyleam: > It would be nice if it could be used, but it can't ATM. Requires a rewrite of assert_logged, which isn't flexible enough for that. It will prepend the message I'm looking for with a fixed prefix of for the level. But this prefix doesn't account for other things in the captured log before that. It would assume the captured log actually starts with `[DEBUG]` for example. But that doesn't work if there are other things logged at higher level before.

`match=False` should cover that:

```diff
diff --git a/datalad/support/tests/test_annexrepo.py b/datalad/support/tests/test_annexrepo.py
index ddacc5ef8..99bd2b10a 100644
--- a/datalad/support/tests/test_annexrepo.py
+++ b/datalad/support/tests/test_annexrepo.py
@@ -2249,8 +2249,10 @@ def test_annex_cmd_expect_fail(path):
     with swallow_logs(logging.DEBUG) as cml:
         repo._run_annex_command_json('add', ['non-existing'], expect_fail=True)
         # message shows up at DEBUG level:
-        assert_re_in(r".*\[DEBUG\][^[]*git-annex: add: 1 failed", cml.out, flags=DOTALL)
+        cml.assert_logged(r"[^[]*git-annex: add: 1 failed",
+                          level="DEBUG", match=False)
     with swallow_logs(logging.DEBUG) as cml:
         repo._run_annex_command_json('add', ['non-existing'], expect_fail=False)
         # message shows up at WARNING level
-        assert_re_in(r".*\[WARNING\][^[]*git-annex: add: 1 failed", cml.out, flags=DOTALL)
+        cml.assert_logged(r"[^[]*git-annex: add: 1 failed",
+                          level="WARNING", match=False)
```

> In short: I tried, but gave up on it, since I was already spending more time than I wanted. Only tackled this one, because I thought: "Well, one-liner. Just spend those 5 minutes." Always turns out to be wrong ;-)

A familiar feeling :)

datalad/support/annexrepo.py
kyleam: > Although - it is somewhat confusing how exactly we want to have it. `expect_stderr` and `expect_fail` are different, independent parameters. So, what should be the meaning of combining them? If I declare to expect a non-zero exit, but not expecting something on stderr shouldn't we still have it logged as a warning?

Yeah, I find the interaction between `expect_stderr` and `expect_fail` confusing.  I can find spots in cmd.py where it clearly looks at `expect_fail` if the status is non-zero and `expect_stderr` otherwise, but I don't think that's necessarily the behavior I'd expect from their docstring descriptions.  I'd guess we could collapse them into one parameter and retain all the cases we actually care about.

Anyway, for the case of this change, where we are in a non-zero exit, I think whether we look at `expect_stderr` or `expect_fail` isn't a big deal because I'd think `expect_fail == expect_stderr` would be the dominant case.

yarikoptic:
Thanks!

kyleam:
> yarikoptic added the pick up to 0.11.x label

FWIW I don't think these sorts of non-bugfix tweaks are worth cherry picking from master to 0.11.x.

datalad/support/tests/test_annexrepo.py
bpoldrack: Well, that seems true. Still don't like assert_logged much, since it fails to properly check a log entry at a time (which is why `r"[^[]*` is there), but I'll honor you pointing that out by using it :-)

yarikoptic:
Sure thing we have more sever divergences in UX/API between the too, but I thought that given its size might be worth. But ok, will remove the label

kyleam:
> but I thought that given its size might be worth

If you end up deciding it is, please discard my two cents and pick away :]

PR #3525, opened 2019-07-11
ENH: fail early if some metadata extractor is not available

yarikoptic:
Otherwise might wait for half an hour just to blow

+ inform about all unavailable, not just the first one


yarikoptic:
any objections/recommendations/contributions @kyleam and @mih?

PR #3526, opened 2019-07-12
Disable -containers test run until Singularity-Hub is back

mih:
n.t.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -187,14 +187,15 @@ matrix:
     - NOSE_SELECTION_OP=not
     - NOSE_SELECTION=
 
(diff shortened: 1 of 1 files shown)


mih:
Forget about this. I will instead skip the relevant tests in -container

kyleam:
> Forget about this. I will instead skip the relevant tests in -container [datalad/datalad-container#92](https://github.com/datalad/datalad-container/pull/92)

That doesn't (yet) help the -container tests here because the released version of -container is used.  I've marked the run as an allowed failure on Travis, so it will no longer cause the whole build to be red.

PR #3527, opened 2019-07-12
PY3: safeguard __del__ in Annex and Git Repo to not breed useless exceptions

yarikoptic:
as e.g. reported in https://github.com/datalad/datalad-crawler/issues/47

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -356,7 +356,7 @@ class AnnexRepo(GitRepo, RepoInterface):
             are no longer bound"""
             try:
                 return lgr.debug(exc_str(e))
-            except AttributeError:
+            except (AttributeError, NameError):
                 return
 
         try:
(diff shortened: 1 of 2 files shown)


datalad/support/annexrepo.py
```diff
@@ -350,6 +350,15 @@ def set_default_backend(self, backend, persistent=True, commit=True):
             self.config.set('annex.backends', backend, where='local')
 
     def __del__(self):
+
+        def safe__del__debug(e):
+            """We might be too late in the game and either .debug or exc_str
+            are no longer bound"""
```
kyleam: Only an AttributeError is caught below.  If exc_str() was no longer bound, wouldn't that be a NameError?

yarikoptic: good question... may be?  let me add it as well just in case ;-) thanks!

yarikoptic: although I do not think local binds disappear first... at least I can't recall if I ran into such before.  But shouldn't hurt - pushed 59d7d48a553ed8e2851a37ae0af839e28330d528

kyleam: OK, thanks.  If exc_str is not a known problem, of course it'd be fine to just remove the false information from the docstring.  But not a big deal either way.

yarikoptic: ah, I see !  thanks for spotting that -- I forgot that I wrote it ;)

Issue #3520, opened 2019-07-10
`status --annex basic` -> `status --annex`


mih:
It just feels needlessly complicated to have to state an explicit mode for the commandline `--annex` option. It should just do `--annex basic` implicitly, if just `--annex` is given.

Issue #3519, opened 2019-07-09
`datalad status --annex basic` fails with URL-type keys


mih:
```
(datalad3-dev) mih@meiner /tmp/machinelearning-books (git)-[master] % dl status --annex all
[ERROR  ] invalid literal for int() with base 10: 'unknown' [annexrepo.py:get_content_annexinfo:3141] (ValueError) 
(datalad3-dev) 1 mih@meiner /tmp/machinelearning-books (git)-[master] % dl --dbg status --annex all
Traceback (most recent call last):
  File "/home/mih/env/datalad3-dev/bin/datalad", line 8, in <module>
    main()
  File "/home/mih/hacking/datalad/git/datalad/cmdline/main.py", line 505, in main
    ret = cmdlineargs.func(cmdlineargs)
  File "/home/mih/hacking/datalad/git/datalad/interface/base.py", line 645, in call_from_parser
    ret = list(ret)
  File "/home/mih/hacking/datalad/git/datalad/interface/utils.py", line 428, in generator_func
    result_renderer, result_xfm, _result_filter, **_kwargs):
  File "/home/mih/hacking/datalad/git/datalad/interface/utils.py", line 520, in _process_results
    for res in results:
  File "/home/mih/hacking/datalad/git/datalad/core/local/status.py", line 370, in __call__
    content_info_cache):
  File "/home/mih/hacking/datalad/git/datalad/core/local/status.py", line 124, in _yield_status
    ref=None)
  File "/home/mih/hacking/datalad/git/datalad/support/annexrepo.py", line 3141, in get_content_annexinfo
    rec['bytesize'] = int(rec['bytesize'])
ValueError: invalid literal for int() with base 10: 'unknown'

> /home/mih/hacking/datalad/git/datalad/support/annexrepo.py(3141)get_content_annexinfo()
-> rec['bytesize'] = int(rec['bytesize'])
(Pdb) rec
[...]

PR #3534, opened 2019-07-16
Fix minor status issues 

mih:
- [x] `status --annex basic` -> `status --annex` #3520
- [x] `datalad status --annex basic` fails with no-size URL-type keys #3519


mih:
Thx @kyleam !

PR #3535, opened 2019-07-16
ENH(TST): code to produce s3://datalad-test1-manydirs-versioned/ for testing

yarikoptic:
should be safe to merge ;)


yarikoptic:
Bucket is there now, so will merge for provenance

PR #3541, opened 2019-07-17
Temp test devel annex

yarikoptic:
should be 7.20190708+git36-g32d526164-1 fixing an inability to `get` in detached HEAD (we don't have tests for that AFAIK,  PR is coming) http://git-annex.branchable.com/bugs/get_fails_to_place_v7_unlocked_file_content_into_the_file_tree_in_v7_in_repo_with_detached_HEAD/


yarikoptic:
cool, all green, uploading

PR #3544, opened 2019-07-18
TST: Add travis run that compares PR benchmarks against current master

### LINKED ISSUE #3539: Include benchmark comparison in dedicated travis run
In the light of #3537 I wonder whether it wouldn't be sensible to have a dedicated travis run that performs whether is needed to enables @yarikoptic and @kyleam to compare benchmark performance. It should be sufficient to compare HEAD of a PR against whatever state of master, or even (in addition) the last release tag. If that is possible, this increasingly relevant information would be readily available for each PR -- whether or not one would expect a performance impact. It might additionally provide a great source of motivation to improve the validity and reliability of benchmark implementations.

It would probably be important to run pre and post assessment of performance within the same travis run and machine, to avoid unnecessary cross-machine differences.

mih:
Just for my personal pleasure (fixes #3539). Example: https://travis-ci.org/datalad/datalad/jobs/560509048#L1246

- appveyor intentionally not ran

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -299,23 +299,7 @@ install:
   - if [ "${BUILD_DATALAD_EXTENSION:-}"x = "datalad-container"x ]; then travis_retry sudo eatmydata apt-get install singularity-container; fi
 
 script:
-  - if [ ! -z "${_DL_BENCHMARK:-}" ]; then
-        pip install asv;
-        asv machine --yes ;
-        pip install -e .;
-        sed -i -e 's,"branches,//branches,g' -e 's,"pythons,//pythons,g' -e 's,.asv/,travisasv/,g' asv.conf.json ;
-        git tag '__bench_target__';
-        git rev-parse HEAD __bench_target__;
-        asv run -E existing --set-commit-hash $(git rev-parse __bench_target__);
-        git reset --hard;
-        git checkout origin/master;
-        git rev-parse HEAD __bench_target__;
-        pip install -e .;
-        sed -i -e 's,"branches,//branches,g' -e 's,"pythons,//pythons,g' -e 's,.asv/,travisasv/,g' asv.conf.json ;
-        asv run -E existing --set-commit-hash $(git rev-parse origin/master);
-        asv compare origin/master __bench_target__;
-        exit 0;
-    fi
+  - if [ ! -z "${_DL_BENCHMARK:-}" ]; then tools/ci/benchmark-travis-pr.sh; exit $?; fi
   # Test installation system-wide
   - sudo pip install .
   - if [ ! -z "${BUILD_DATALAD_EXTENSION:-}" ]; then pip install "$BUILD_DATALAD_EXTENSION"; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -201,6 +201,10 @@ matrix:
     - NOSE_SELECTION_OP=not
     - NOSE_SELECTION=
 
+  # Run benchmark comparison against current master
+  - python: 3.6
+    env: _DL_BENCHMARK=1
+
 # Requires current master (0.12.x) so cannot be tested against 0.11.x
 #
 #  - python: 3.5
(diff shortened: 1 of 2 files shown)


yarikoptic:
It would have been awesome if we could do it, but I am not sure if time shared Travis instances would provide any stable measurement. If there was time to bring up turned down buildbot, I could provide get a dedicated server for that

mih:
@yarikoptic That is not much of a concern here. It runs the entire comparison within one machine, on both versions. Check the link above.

As far as I can see right now, those estimates are not much different from my own laptop, and some are *really* variable.

.travis.yml
```diff
@@ -285,6 +288,23 @@ install:
...
+        git tag '__bench_target__';
+        git rev-parse HEAD __bench_target__;
+        asv run -E existing --set-commit-hash $(git rev-parse __bench_target__);
+        git reset --hard;
+        git checkout origin/master;
+        git rev-parse HEAD __bench_target__;
+        pip install -e .;
+        sed -i -e 's,"branches,//branches,g' -e 's,"pythons,//pythons,g' -e 's,.asv/,travisasv/,g' asv.conf.json ;
+        asv run -E existing --set-commit-hash $(git rev-parse origin/master);
+        asv compare origin/master __bench_target__;
```
yarikoptic: There should be some env variable which would show the merge target, or if Travis tests merged state, then I guess it would be HEAD^2 or something like that? That would allow to benchmark 0.11.x PRs as well

yarikoptic:
I understand that it is the same virtual box both times. But through time of benchmarking that physical box could vary in how many other jobs, and how io intensive they are, which might effect benchmarks

But oh well, it is better than nothing for sure

.travis.yml
yarikoptic: Also may be this whole code could go into a script, which could then get used even locally or across cis?

mih: re variables: dunno about those. But it would be great to test against whatever the base is, indeed.

re into-a-script: I didn't do that, because that script would be in a branch, and possibly change between `HEAD` and base. Of course this could all be mitigated by explicitly checking things out (`cat-file ...`) from `HEAD`, but I had no such ambitions for now.

kyleam: > There should be some env variable which would show the merge target, or if Travis tests merged state, then I guess it would be HEAD^2 or something like that? That would allow to benchmark 0.11.x PRs as well

GitHub's 'refs/pull/*/merge' ref is the result of the PR merged into the mainline, so HEAD^1 would be either origin/master or origin/0.11.x.   Though I'd find defining the target branch an environment variable cleaner because then the same logic would work for non-PR branches.

Even if we don't bother making this work for 0.11.x in this PR, I think we should skip the benchmarks if the run is for the master branch where `__bench_target__` would be master.

kyleam: > re variables: dunno about those.

I _think_ the suggestion is to have a variable in .travis.yml (say `_DL_MERGE_TARGET`) that points to "origin/master" in `master:.travis.yml` and to "origin/0.11.x" in `0.11.x:.travis.yml`.  (It could also default to "origin/master" when not defined.)

yarikoptic: Re branch variable from https://docs.travis-ci.com/user/environment-variables/#convenience-variables

TRAVIS_BRANCH:
- for push builds, or builds not triggered by a pull request, this is the name of the branch.
- for builds triggered by a pull request this is the name of the branch targeted by the pull request.
- for builds triggered by a tag, this is the same as the name of the tag (TRAVIS_TAG).

yarikoptic: Which also reminds that this run is relevant only for PRs and should exit cleanly in regular branch build

yarikoptic: Regarding the script - I don't see how anything is different from just coding it in .travis.yml which could also differ between branches. But it will be easier to check/sync across branches if it was in a script

kyleam: > Re branch variable from https://docs.travis-ci.com/user/environment-variables/#convenience-variables

The variables you list, along with TRAVIS_PULL_REQUEST, are fine for detection.  But as I said, "I'd find defining the target [merge] branch as an environment variable cleaner because then the same logic would work for non-PR branches."

> Which also reminds that this run is relevant only for PRs and should exit cleanly in regular branch build

AFAICT this is the [same issue I mentioned above](https://github.com/datalad/datalad/pull/3544#discussion_r304982395), but you propose to exclude all non-PR runs rather than just run for  master or 0.11.x.

mih: It seems there is much that I am/was not aware of. I'll leave it here for now, and try to come back to it later. In the meantime I can satisfy my desires by plopping this temporarily onto relevant PRs.

kyleam: >  I'll leave it here for now, and try to come back to it later. 

I can take it over.

> In the meantime I can satisfy my desires by plopping this temporarily onto relevant PRs.

Beware that the current form isn't doing the intended comparison AFAICT.  As I was reworking this, I realized that `asv run` reports to be benchmarking the same commit (compare [here][0] to [here][1]).  The recipe in this PR expects that HEAD is benchmarked if "branches" isn't configured, but, based on the comment in asv.conf.json and local testing, it looks like it benchmarks master if no branches are specified.

[0]: https://travis-ci.org/datalad/datalad/jobs/560509048#L1116
[1]: https://travis-ci.org/datalad/datalad/jobs/560509048#L1205

yarikoptic:
Fwiw quick Google to how others might use it lead to

https://github.com/scipy/scipy/blob/master/.travis.yml#L215
```
  - if [ "${ASV_CHECK}" == "1" ]; then (cd benchmarks && python -masv check -E existing); fi
```

yarikoptic:
Which is just to check validity, boring ;-)

kyleam:
>  The recipe in this PR expects that HEAD is benchmarked if "branches" isn't configured, but, based on the comment in asv.conf.json and local testing, it looks like it benchmarks master if no branches are specified.

Resolved with the push of 0b06b180b: <https://travis-ci.org/datalad/datalad/jobs/561080423>

That push also restricts the run to PR branches and takes HEAD^1 as the base.  I've pushed a final tweak to the output.  Once that run completes, I'd say this is good to go.

mih:
Thanks @kyleam !

PR #3546, opened 2019-07-19
BF: AnnexRepo to deal with INFO response from annex special remotes

bpoldrack:
`AnnexRepo`, in particular `_run_annex_command_json`  didn't handle an INFO response from a special remote (see https://git-annex.branchable.com/design/external_special_remote_protocol/#index5h2).
Fixed that.

FTR: I won't implement an entire special remote just to fake some test herein. However, it will at least be implicitly tested by https://github.com/datalad/git-annex-ria-remote.
Meaning: I think it's fine if I didn't break anything ;-)

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -2160,6 +2160,25 @@ class AnnexRepo(GitRepo, RepoInterface):
                         "Received no json output for --json command, only:\n{}"
                         .format("  ".join(others)))
                 raise exc
+
+        # A special remote might send a message via "info". This is supposed to be printed by annex but in case of
+        # `--json` is returned by annex as "{'info': '<message>'}".
+        # See https://git-annex.branchable.com/design/external_special_remote_protocol/#index5h2
+        #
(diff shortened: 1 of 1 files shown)


datalad/support/annexrepo.py
```diff
@@ -2160,6 +2160,25 @@ def _run_annex_command_json(self, command,
...
+        # other than spitting it out.
+
+        to_remove = []
+        for obj in json_objects:
+            if len(obj.keys()) == 1 and obj['info']:
+                lgr.info(obj['info'])
+                to_remove.append(obj)
+
+        for obj in to_remove:
+            json_objects.remove(obj)
```
yarikoptic: Code will be shorter if you just compose that list to return in the loop... Not sure even if remove would work correctly here eh in case of multiple identical info messages

yarikoptic:
looks LGTM.  Re tests, if interested to dig a bit -- `datalad/customremotes/tests/test_base.py` is the one which could code generic logic tests for special remotes

PR #3547, opened 2019-07-19
BF: check special remotes of type git by uuid to (not) inform if they are not enabled

yarikoptic:
Matching by the name is not sufficient, since when we just clone such
repository it would get an arbitrary (by default "origin") name which
would not match the one git-annex calls it by.  Checking by annex UUID
seems to be the best way to proceed

FOI: test takes 1.7sec on my laptop (on network close to the datasets.datalad.org)

### DIFF
--- a/datalad/distribution/utils.py
+++ b/datalad/distribution/utils.py
@@ -19,13 +19,14 @@ import posixpath
 
 from six.moves.urllib.parse import unquote as urlunquote
 
+from ..dochelpers import single_or_plural
 from datalad.support.annexrepo import GitRepo
 from datalad.support.annexrepo import AnnexRepo
 from datalad.support.network import DataLadRI
 from datalad.support.network import URL
 from datalad.support.network import RI
 from datalad.support.network import PathRI
-from datalad.utils import knows_annex
+from datalad.utils import knows_annex, assure_bool
 
 
 lgr = logging.getLogger('datalad.distribution.utils')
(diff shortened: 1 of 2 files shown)


yarikoptic:
THANK YOU @kyleam for the review and catching the regression.  I should have thought about it ;)

> Why don't we just check the value of the "autoenable" key?

But to check it for what remote exactly, i.e. how (if not using uuid) we can associate any special remote with a known to git remote?

may be upon clone and seeing an annex we should run `annex info` which would contact those remotes and assign uuid:

```shell
/tmp > datalad clone http://github.com/ReproNim/containers && cd containers
[INFO   ] Cloning http://github.com/ReproNim/containers [1 other candidates] into '/tmp/containers' 
[INFO   ]   Remote origin not usable by git-annex; setting annex-ignore                                         
[INFO   ] access to dataset sibling "datasets.datalad.org" not auto-enabled, enable with:
| 		datalad siblings -d "/tmp/containers" enable -s datasets.datalad.org 
install(ok): /tmp/containers (dataset)

/tmp/containers > git config remote.datasets.datalad.org.annex-uuid                   

/tmp/containers > git annex info | grep datasets.datalad.org
 	71c620b5-997f-4849-bb30-c42dbb48a51e -- [datasets.datalad.org]

/tmp/containers > git config remote.datasets.datalad.org.annex-uuid
71c620b5-997f-4849-bb30-c42dbb48a51e
```

?

kyleam:
@yarikoptic:

>> Why don't we just check the value of the "autoenable" key?
>
> But to check it for what remote exactly, i.e. how (if not using uuid) we can associate any special remote with a known to git remote?

Hmm, I might be getting confused, but I'll try to elaborate in case you're missing my suggestion.

The current behavior is "for each remote in git-annex:remote.log, try to find the remote in .git/config based on $CRITERIA and, if it isn't found, give the user instructions for enabling it".  Your PR is changing $CRITERIA for `type=git`, but I'm asking why not get rid of the .git/config inspection altogether, warning based on the autoenable key for each item in get_special_remotes().

yarikoptic:
> I'm asking why not get rid of the .git/config inspection altogether, warning based on the autoenable key for each item in get_special_remotes().

ah, got it -- but would not that *assume* (without checking) that the remote was enabled?

kyleam:
> ah, got it -- but would not that _assume_ (without checking) that the remote was enabled?

Yes, but as I said above, I don't see how the check really helps:

> Why don't we just check the value of the "autoenable" key? That might miss cases where `git annex init` failed to enable the remote, but presumably the `datalad sibling` snippet we give for enabling will fail for the same reason, so dropping the instructions isn't really hurting the situation.

yarikoptic:
observation: our "install" is not really causing git-annex to autoenable any of those special remotes.  They do get added to git configuration as remotes, but annex doesn't even try to reach them to establish their annex UUID.

<details>
<summary>annex UUID fetching and  "autoenable" happens later (or with the initial version of the this PR here due to `annex info`) by e.g. invoking either `git annex info` or `git annex get` (! was our discussion topic).</summary>

```shell
/tmp > datalad install https://github.com/ReproNim/containers
[INFO   ] Cloning https://github.com/ReproNim/containers [1 other candidates] into '/tmp/containers' 
[INFO   ]   Remote origin not usable by git-annex; setting annex-ignore                                                        
install(ok): /tmp/containers (dataset)

/tmp > git -C containers config remote.datasets.datalad.org.annex-uuid  

/tmp > git -C containers annex get images/bids/bids-validator--1.2.3.sing
get images/bids/bids-validator--1.2.3.sing (from datasets.datalad.org...) 
4%    1.3 MiB        1018 KiB/s 35s^C

/tmp > git -C containers config remote.datasets.datalad.org.annex-uuid   
71c620b5-997f-4849-bb30-c42dbb48a51e
```
</details>

It somewhat also relates to our discussion on "either should we enable special remotes on a remote while publishing?".

I think to stay more consistent with git-annex and also avoid spitting out some warnings and failures whenever it is not yet relevant (may be I don't need to `get` any files in the installed dataset?), I will indeed RF this PR to avoid `info` call, and just provide analysis of special remotes based on `autoenable`.

But we might need to improve reporting of failures of autoneabling remotes from git-annex (see #3564) later in the game (`info` or `get` or ... whatever triggers it calls)

yarikoptic:
hm, while coding for it realized that the situation is more fun.  git special remote which has no autoenable might also be the one we just cloned! ;)  so I will need to do some check by UUID (without calling `info` - just from config) to prune those out

kyleam:
> observation: our "install" is not really causing git-annex to autoenable any of those special remotes. They do get added to git configuration as remotes, but annex doesn't even try to reach them to establish their annex UUID.
> annex UUID fetching and "autoenable" happens later (or with the initial version of the this PR here due to `annex info`) by e.g. invoking either `git annex info` or `git annex get` (we had some discussion on this topic recently).

It's not accurate to say `install` isn't causing git-annex to autoenable the special remotes.  It is autoenabling them in the same way as `git annex init` does, which is why `get` can find the files on the special remote.

AFAICT all the remaining information you provide about a UUID not showing up in the config is restating what I said in my initial comment:

> The problem comes from assuming that the `git annex init` call (triggered by `_handle_possible_annex_dataset`) configures the UUID when enabling the remote. That isn't (at least always) the case:
> 
> ```
> # in the last cloned repo from above
> $ cat .git/config | tail -3
> [remote "datasets.datalad.org"]
>         url = http://datasets.datalad.org/repronim/containers/.git
>         fetch = +refs/heads/*:refs/remotes/datasets.datalad.org/*
> ```
> 
> It isn't until later, say getting a file, that the UUID shows up:

datalad/distribution/utils.py
```diff
@@ -152,30 +153,57 @@ def _handle_possible_annex_dataset(dataset, reckless, description=None):
...
-                sr_name,
-                dataset.path,
-                sr_name)
+            if uuid not in remote_uuids:
+                srs[sr_autoenable].append(sr_name)
+
+    if srs[True]:
+        lgr.debug(
+            "configuration for %s %s added because of autoenable,"
+            " but no UUIDs for them yet known for dataset %s",
```
kyleam: I don't see much value in this because it seems to be the norm, but it's of course not a big deal given that it's logged at the debug level.

yarikoptic: All Debug messages have only hypothetical value of helping to troubleshoot when something goes wrong. Given the tricky nature of all this, might come handy IMHO

PR #3555, opened 2019-07-24
Fix `siblings` crash with missing --name

### LINKED ISSUE #3553: `datalad siblings add --url <path>` crashes
```
% dl siblings add --url ~/hacking/datalad/git
[ERROR  ] 'PathRI' object has no attribute 'hostname' [network.py:__getattribute__:547] (AttributeError) 
```

we should guard against that in any case, and possibly generate a default name from the path to provide the same convenience as when an actual URL is given.

mih:
Fixes #3553 

Plus bonus reformatting of imports.

### DIFF
--- a/datalad/distribution/siblings.py
+++ b/datalad/distribution/siblings.py
@@ -14,8 +14,7 @@ __docformat__ = 'restructuredtext'
 import logging
 
 import os
-from os.path import basename
-from os.path import relpath
+import os.path as op
 
 from six.moves.urllib.parse import urlparse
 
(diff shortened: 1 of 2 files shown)


datalad/distribution/siblings.py
```diff
@@ -123,10 +136,10 @@ class Siblings(Interface):
         name=Parameter(
             args=('-s', '--name',),
             metavar='NAME',
-            doc="""name of the sibling. For sibling removal this option is
-            mandatory, otherwise the hostname part of a given URL is used as a
-            default. This option can be used to limit 'query' to a specific
-            sibling.""",
+            doc="""name of the sibling. For sibling removal and addition
+            with path "URLs" this option is mandatory, otherwise the hostname
```
kyleam: The reader would have to infer whether 'with path "URLs" applies to just addition or both addition and removal.  You could avoid this ambiguity if you move 'addition with path "URLs"' before "sibling removal".

yarikoptic: Dear @kyleam -- could you please do what needs to be done (rebase on 0.11.x, adjust docstring here) so we could have this one merged ?

kyleam: Done

<details>
<summary>range-diff</summary>

```
1:  3b708139b = 1:  d9321778e RF: Modernize imports
2:  474415ba9 ! 2:  b186b1427 BF+TST: Don't crash with missing arg (fixes gh-3553)
    @@ -13,8 +13,8 @@
     -            mandatory, otherwise the hostname part of a given URL is used as a
     -            default. This option can be used to limit 'query' to a specific
     -            sibling.""",
    -+            doc="""name of the sibling. For sibling removal and addition
    -+            with path "URLs" this option is mandatory, otherwise the hostname
    ++            doc="""name of the sibling. For addition with path "URLs" and
    ++            sibling removal this option is mandatory, otherwise the hostname
     +            part of a given URL is used as a default. This option can be used
     +            to limit 'query' to a specific sibling.""",
```

</details>

Issue #3382, opened 2019-05-03
Replace some "pragma: no cover" with "pragma: more cover"


yarikoptic:
original discussion: https://github.com/datalad/datalad/pull/3366#discussion_r280824123

"no cover" means (to coverage.py) that the code is not even intended to be covered, so it is excluded from the total '%' count.

"more cover" is a mental note that even if code is covered, it is known to be covered insufficiently

PR #3568, opened 2019-07-30
BF: replace "no cover" with "more cover" for cases where a dedicated testing was not implemented

yarikoptic:
See https://github.com/datalad/datalad/pull/3366#discussion_r280289780

Closes: #3382

### DIFF
--- a/datalad/core/local/create.py
+++ b/datalad/core/local/create.py
@@ -445,7 +445,7 @@ class Create(Interface):
 
 
     @staticmethod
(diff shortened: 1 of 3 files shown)


yarikoptic:
hm, no codecov report at all this time -- when it most matters ;-)

kyleam:
> hm, no codecov report at all this time -- when it most matters ;-)

Doh.  

The "no cover" vs "more cover" distinction sounds fine to me.

Issue #3567, opened 2019-07-29
seems we need to optimize get (may be others) results rendering (?) to avoid a heavy loop of resolving subdatasets


yarikoptic:
Manifests on 0.11.x (log messages are different but feels the same) and master (0.12.0rc4-267-g5e410c26).  
`datalad run` with a good number of (globbed) inputs (>1000) from subdatasets resulted in an hour or so of waiting to get pass the message `Making sure inputs are available (this may take some time)`. I interrupted it since it was not clear if it would complete in any (un)reasonable time.

With the "generators all the way down" I thought that backtrace is useless and that the actual "useful action" is somewhere else, but I think what I have observed (shown below) is indeed just a "results rendering" loop!

```
...
  /home/XXX/proj/datalad/datalad-master/datalad/interface/utils.py(520)_process_results()
-> for res in results:
  /home/XXX/proj/datalad/datalad-master/datalad/distribution/get.py(470)__call__()
-> on_failure='ignore'):
  /home/XXX/proj/datalad/datalad-master/datalad/interface/utils.py(428)generator_func()
-> result_renderer, result_xfm, _result_filter, **_kwargs):
  /home/XXX/proj/datalad/datalad-master/datalad/interface/utils.py(520)_process_results()
-> for res in results:
  /home/XXX/proj/datalad/datalad-master/datalad/interface/annotate_paths.py(679)__call__()
-> result_xfm=None, result_filter=None, return_type='list')
  /data/R03/WM-R03/venvs/master/lib/python3.5/site-packages/wrapt/wrappers.py(603)__call__()
-> args, kwargs)
  /home/XXX/proj/datalad/datalad-master/datalad/distribution/dataset.py(525)apply_func()
-> return f(**kwargs)
  /data/R03/WM-R03/venvs/master/lib/python3.5/site-packages/wrapt/wrappers.py(564)__call__()
-> args, kwargs)
  /home/XXX/proj/datalad/datalad-master/datalad/interface/utils.py(491)eval_func()
-> return return_func(generator_func)(*args, **kwargs)
  /data/R03/WM-R03/venvs/master/lib/python3.5/site-packages/wrapt/wrappers.py(564)__call__()
-> args, kwargs)
  /home/XXX/proj/datalad/datalad-master/datalad/interface/utils.py(479)return_func()
-> results = list(results)
  /home/XXX/proj/datalad/datalad-master/datalad/interface/utils.py(428)generator_func()
[...]

PR #3570, opened 2019-07-30
OPT: maintain cache of discovered subdatasets in the loop of annotate_paths

yarikoptic:
Otherwise there is a rediscovery of subdatasets for EVERY path provided to annotate_paths even when they all belong to the same parent dataset.

Closes: #3567 

TODOs:
- [x] verify that nothing is broken at least according to the tests.  As is mentioned in the code comment, because of its "generator" nature, if annotate_path is called from within some loop altering the list of submodules, cache could get invalidated.  If there is such a usage case we should probably add some kind of sensing of mtime for .gitmodules of the parent or add an explicit argument to not retain this cache, but then things will get slow again whenever not needed
- [x] see and report if any notable effect on benchmarks
- [x] verify that it helps in a real use case which "inspired" this fix
- (not here) we really should add a benchmark suite operating on a set of paths within a bigger dataset across a variety of commands

### DIFF
--- a/datalad/interface/annotate_paths.py
+++ b/datalad/interface/annotate_paths.py
@@ -563,6 +563,16 @@ class AnnotatePaths(Interface):
             # re-append the preserved paths:
             requested_paths = chain(requested_paths, iter(preserved_paths))
 
+        # Possibly to be used "cache" of known subdatasets per each parent
+        # to avoid re-querying subdatasets per each path.  The assumption here
+        # is that the list of sub-datasets for a given parent should not change
+        # through the execution of this loop, which (hypothetically) could be
+        # incorrect while annotating paths for some commands.
+        # TODO: verify this assumption and possibly add an argument to turn
+        #  caching off if/when needed, or provide some other way to invalidate
+        #  it
+        subdss_cache = {}
+
         # do not loop over unique(), this could be a list of dicts
         # we avoid duplicates manually below via `reported_paths`
         for path in requested_paths:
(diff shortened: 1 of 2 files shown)


yarikoptic:
<details>
<summary>benchmarks on travis show no significant effect of any kind and all suggest a bit of slowness... kheh kheh ;-)
</summary> 

```shell
     [c411d6a2]       [58c23cd8]
     <bm/merge-target>       <bm/pr>   
       1.25±0.01s       1.26±0.02s     1.00  api.supers.time_createadd
       1.22±0.01s       1.24±0.01s     1.02  api.supers.time_createadd_to_dataset
       3.32±0.04s       3.51±0.06s     1.06  api.supers.time_installr
          136±1ms          148±8ms     1.09  api.supers.time_ls
       1.23±0.01s       1.28±0.06s     1.04  api.supers.time_ls_recursive
       1.34±0.02s       1.40±0.02s     1.04  api.supers.time_ls_recursive_long_all
       2.01±0.02s       2.06±0.01s     1.03  api.supers.time_remove
          398±4ms          412±3ms     1.03  api.testds.time_create_test_dataset1
       2.78±0.03s       2.95±0.03s     1.06  api.testds.time_create_test_dataset2x2
      2.04±0.03ms      2.13±0.03ms     1.05  core.runner.time_echo
      2.84±0.05ms      2.94±0.02ms     1.04  core.runner.time_echo_gitrunner
         655±10ms          688±3ms     1.05  core.startup.time_help_np
        156±0.8ms          160±1ms     1.02  core.startup.time_import
          595±4ms        621±0.9ms     1.04  core.startup.time_import_api
```
</details>

yarikoptic:
real life use case, 
- 382,940 files (via `git ls-tree -r`) in the repo with 249,691 `local annex keys`
- 1,830 *feat sub directories pointing to 368,292 symlinks (annexed) and 12,810 regular files  (in git)
- `.git/objects` is quite sizeable -- 59,322 MB (may be I should `git gc` after the last `datalad save` run)
- no sub-modules (but this repo is a 3rd level submodule)
- all data is already here (and didn't go anywhere ;-))

here is timings with eyeball averaging 3 runs:

- pure `git annex get` -- 3m30s
- pure `'git' '-c' 'annex.merge-annex-branches=false' 'annex' 'find' '--json' '--not' '--in' 'here' '--' ` which we use in `datalad get` to see what we actually need to get (to get idea on total size before running `annex get`) -- 11m43s (I guess we should check with @joeyh on how to get a more performant querying to get closer to `get` performance)
- patched (this PR) `datalad get` -- 11m48s  ( so we are bound by `annex find` above)
- master `datalad get` -- 13m13s (so not that bad, but >10% time wasted)

yarikoptic:
so, checked with @joeyh, and with using `.` instead of `here` we would get boost

- pure 'git' '-c' 'annex.merge-annex-branches=false' 'annex' 'find' '--json' '--not' '--in' '.' '--' 4m

PR #3574 will introduce that change.

yarikoptic:
ok, since there is a boost, and no side-effects were detected so far, merging

PR #3573, opened 2019-07-31
ENH merge 0.11.x (post 0.11.6 release)

yarikoptic:


yarikoptic:
travis had some difficulties with neurodebian (I didn't even know that it is i386 and not amd64), restart of one job seems has helped, so I restarted all

PR #3574, opened 2019-08-01
OPT: use find --in ., not --in here to gain 3x speed up on some cases

yarikoptic:
Originally found while timing effects of a PR:
  https://github.com/datalad/datalad/pull/3570#issuecomment-517084987
Similar change is now done also on git-annex level in
  d1a0c7b16fcd41f74def394a55c32f67340670b7
but to not wait for its new release, adjusting our
find invocations

TODOs
- [x] verify nothing is broken
- [x] no effect on our existing benchmarks?
- [x] time up that use case in #3570

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -1383,7 +1383,7 @@ class AnnexRepo(GitRepo, RepoInterface):
 
         if not key:
             expected_downloads, fetch_files = self._get_expected_files(
-                files, ['--not', '--in', 'here'],
+                files, ['--not', '--in', '.'],
                 merge_annex_branches=False  # interested only in local info
             )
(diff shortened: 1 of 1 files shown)


yarikoptic:
FWIW, locally `$> asv continuous 0.11.x HEAD` resulted in "BENCHMARKS NOT SIGNIFICANTLY CHANGED", so no useful benchmark for it there is here (at least in 0.11.x)

yarikoptic:
can't say anything about windows ATM, but otherwise -- no side effects found, performance boost confirmed on real case, and no negative effect on our benchmarks (including on [travis](https://travis-ci.org/datalad/datalad/jobs/566489113#L1213)).  So -- merging!

Issue #3575, opened 2019-08-01
OPT: Something slows down datalad (run) within a dirty source code tree...


yarikoptic:
Using `datalad run --explicit` left me in suspense for too long when I ran it within my main development repo:
```shell
~datalad/datalad > time DATALAD_LOG_TIMESTAMP=1 datalad run --explicit --input benchmarks echo nothing todo 2>&1 | tools/dtime 
 2126 2019-08-01 10:47:11,163 [INFO] Making sure inputs are available (this may take some time) 
    - nothing todo
    3 2019-08-01 10:47:13,289 [INFO] == Command start (output follows) ===== 
----- 2019-08-01 10:47:13,292 [INFO] == Command exit (modification check follows) ===== 
DATALAD_LOG_TIMESTAMP=1 datalad run --explicit --input benchmarks echo nothin  2.70s user 0.20s system 100% cpu 2.890 total

~datalad/datalad > datalad --version
datalad 0.11.6.dev3
```
so it took almost 3 sec to "do nothing".  It doesn't happen in a clean repo:
```shell
~datalad/datalad-clone > time DATALAD_LOG_TIMESTAMP=1 datalad run --explicit --input benchmarks echo nothing todo 2>&1 | tools/dtime
  231 2019-08-01 10:48:34,657 [INFO] Making sure inputs are available (this may take some time) 
    - nothing todo
    3 2019-08-01 10:48:34,888 [INFO] == Command start (output follows) ===== 
----- 2019-08-01 10:48:34,891 [INFO] == Command exit (modification check follows) ===== 
DATALAD_LOG_TIMESTAMP=1 datalad run --explicit --input benchmarks echo nothin  0.65s user 0.11s system 101% cpu 0.755 total
tools/dtime  0.02s user 0.00s system 2% cpu 0.760 total
```
and I do not think it is just a matter of asserting dirtiness, since 

<details>
<summary>if I run it with high level of debug</summary> 

```shell
~datalad/datalad > time DATALAD_LOG_TIMESTAMP=1 datalad -l 2 run --explicit --input benchmarks echo nothing todo 2>&1 | tools/dtime        
    0 2019-08-01 10:50:04,153 [DEBUG] Command line args 1st pass. Parsed: Namespace() Unparsed: ['run', '--explicit', '--input', 'benchmarks', 'echo', 'nothing', 'todo'] 
    0 2019-08-01 10:50:04,153 [DEBUG] Discovering plugins 
[...]

PR #3576, opened 2019-08-01
OPT+ENH: determine config to query for procedure once, add more "log(2," to eval_func

yarikoptic:
Original OPT issue: https://github.com/datalad/datalad/issues/3575 . This PR addresses only one aspect, and doesn't touch upon relatively slow `.config` and `.is_installed` which are yet to be figured out and probably should be properly "lazied" at the dataset level since  we are doing this over and over for every decorated function in the same dataset!

Those added debug statements seems to have no cost but could help determining unexpected
slow downs

For me it is cutting down time for a dummy `datalad run` from 2.9 to 2.2 sec (so ~20% boost! ;))
```
~datalad/datalad > time DATALAD_LOG_TIMESTAMP=1 datalad run --explicit --input benchmarks echo nothing todo 2>&1 | tools/dtime 
 1280 2019-08-01 11:40:50,695 [INFO] Making sure inputs are available (this may take some time) 
    - nothing todo
    3 2019-08-01 11:40:51,975 [INFO] == Command start (output follows) ===== 
----- 2019-08-01 11:40:51,978 [INFO] == Command exit (modification check follows) ===== 
DATALAD_LOG_TIMESTAMP=1 datalad run --explicit --input benchmarks echo nothin  2.02s user 0.21s system 100% cpu 2.221 total
tools/dtime  0.01s user 0.01s system 0% cpu 2.225 total
```
and I do not expect negative  effect on other usecases.

### DIFF
--- a/datalad/interface/utils.py
+++ b/datalad/interface/utils.py
@@ -382,7 +382,7 @@ def eval_results(func):
             if spec is None:
                 return None, proc_cfg
             elif not isinstance(spec, tuple):
-                spec = [spec], proc_cfg
+                spec = [spec]
             return [shlex.split(s) for s in spec], proc_cfg
 
         # query cfg for defaults
--- a/datalad/interface/utils.py
+++ b/datalad/interface/utils.py
@@ -297,6 +297,7 @@ def eval_results(func):
 
     @wrapt.decorator
     def eval_func(wrapped, instance, args, kwargs):
+        lgr.log(2, "Entered eval_func for %s", func)
         # for result filters and pre/post procedures
         # we need to produce a dict with argname/argvalue pairs for all args
         # incl. defaults and args given as positionals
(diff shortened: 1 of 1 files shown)


datalad/interface/utils.py
```diff
@@ -363,33 +364,39 @@ def eval_func(wrapped, instance, args, kwargs):
                 def _result_filter(res):
                     return result_filter(res, **allkwargs)
 
-        def _get_procedure_specs(param_key=None, cfg_key=None, ds=None):
+        def _get_procedure_specs(param_key=None, cfg_key=None, ds=None, proc_cfg=None):
```
kyleam: The additional parameter is fine by me, though I have a slight preference for just defining `proc_cfg` above `_get_procedure_specs`.

yarikoptic: that is how I did it first, but then it should be not a pure variable since you can't rebind it inside, but some kind of a container (I had list).  Just looks ugly etc, so decided to make it explicit

yarikoptic:
no travis report here, but it is [there](https://travis-ci.org/datalad/datalad/builds/566539731) and I seems screwed up
```
1114======================================================================
1115ERROR: datalad.distribution.tests.test_create.test_create_withprocedure
1116----------------------------------------------------------------------
1117Traceback (most recent call last):
1118  File "/home/travis/virtualenv/python2.7.15/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
1119    self.test(*self.arg)
1120  File "/home/travis/virtualenv/python2.7.15/lib/python2.7/site-packages/datalad/tests/utils.py", line 615, in newfunc
1121    return t(*(arg + (filename,)), **kw)
1122  File "/home/travis/virtualenv/python2.7.15/lib/python2.7/site-packages/datalad/distribution/tests/test_create.py", line 320, in test_create_withprocedure
1123    proc_post=[['cfg_metadatatypes', 'xmp', 'datacite']])
1124  File "/home/travis/virtualenv/python2.7.15/lib/python2.7/site-packages/datalad/interface/utils.py", line 399, in eval_func
1125    proc_cfg=proc_cfg)
1126ValueError: need more than 1 value to unpack
```
so will fix it now

kyleam:
> so will fix it now

I've already fixed it in fe9bff6f8.

kyleam:
> > so will fix it now
> 
> I've already fixed it in [fe9bff6](https://github.com/datalad/datalad/commit/fe9bff6f8467c941b10d11ba4e15f94527a53b23).

Sorry, ignore me.  I fixed another error that the initial commit caused in test_run_procedure, but it looks like this is a different source.

yarikoptic:
then I guess there more!  I should have indeed went the "outside variable" way to make it less fragile :-/
```shell
$> git show                                        
commit 4601749db64e9d0e45b81b68f8086f9feb422d4c (HEAD -> opt-doitonce)
Date:   Fri Aug 2 09:42:04 2019 -0400

    BF: eval_func: One more fix to return value

diff --git a/datalad/interface/utils.py b/datalad/interface/utils.py
index ea510f598..0a169aaae 100644
--- a/datalad/interface/utils.py
+++ b/datalad/interface/utils.py
@@ -369,7 +369,7 @@ def eval_results(func):
             spec = common_params.get(param_key, None)
             if spec is not None:
                 # this is already a list of lists
-                return spec
+                return spec, proc_cfg
 
             if not proc_cfg:
                 # .is_installed and .config can be costly, so ensure we do
changes on filesystem:                                                                                                                                                                                                            
 .asv | 0
(dev) 1 15626.....................................:Fri 02 Aug 2019 09:50:14 AM EDT:.
(git)hopa:~datalad/datalad[opt-doitonce]git
$> git range-diff gh-yarikoptic/opt-doitonce...HEAD
-:  --------- > 1:  4601749db BF: eval_func: One more fix to return value

```
why range-diff doesn't show the diff?  here is no rewrite but a new commit

pushing

kyleam:
> why range-diff doesn't show the diff? here is no rewrite but a new commit

range-diff won't show the diff for added or dropped commits, only those for matched patches, which means it's not particularly useful (or needed) when you're not rewriting a series.

kyleam:
> ```shell
> commit 4601749db64e9d0e45b81b68f8086f9feb422d4c (HEAD -> opt-doitonce)
> Author: Yaroslav Halchenko <debian@onerussian.com>
> Date:   Fri Aug 2 09:42:04 2019 -0400
> 
>     BF: eval_func: One more fix to return value
> 
> diff --git a/datalad/interface/utils.py b/datalad/interface/utils.py
> index ea510f598..0a169aaae 100644
> --- a/datalad/interface/utils.py
> +++ b/datalad/interface/utils.py
> @@ -369,7 +369,7 @@ def eval_results(func):
>              spec = common_params.get(param_key, None)
>              if spec is not None:
>                  # this is already a list of lists
> -                return spec
> +                return spec, proc_cfg
> ```

Looks good.  I missed that in my fixup commit.

PR #3587, opened 2019-08-05
BF: pass "paths" as "files" argument to _run...command for proper chunking

### LINKED ISSUE #3585: datalad run is hurt by "Argument list too long"
Just reporting for now that we are aware, yet to dig in what lead to it:
```
...
/WM-R03/analyses/1st-level/sub-000768_ses-01_run-03_task-efnback.feat/stats/zstat5.nii.gz', '/data/R03/WM-R03/an
efnback.feat/stats/zstat6.nii.gz', '/data/R03/WM-R03/analyses/1st-level/sub-000768_ses-01_run-03_task-efnback.fe
yses/1st-level/sub-000768_ses-01_run-03_task-efnback.feat/stats/zstat8.nii.gz', '/data/R03/WM-R03/analyses/1st-l
/stats/zstat9.nii.gz']'': [Errno 7] Argument list too long [subprocess.py:_execute_child:1282]
2019-08-03 12:29:40,491 [ERROR  ] [Errno 7] Argument list too long [subprocess.py:_execute_child:1282] (OSError)

$ datalad --version
datalad 0.12.0rc4.dev311
```
smells like an unlock call, but I could be wrong -- rerunning with `--dbg` (if time allows to wait)

yarikoptic:
Otherwise long lists of full paths quickly cause "too long" crash.

Closes #3585 

TODOs
- [x] verify nothing is broken in tests
- [x] see if helps in real-life

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -3181,15 +3181,19 @@ class AnnexRepo(GitRepo, RepoInterface):
         # it takes care of git-annex reporting on any known key, regardless
         # of whether or not it actually (did) exist in the local annex
         opts = ['--copies', '0']
+        files = None
         if ref:
             cmd = 'findref'
             opts.append(ref)
         else:
             cmd = 'find'
             # stringify any pathobjs
-            opts.extend([text_type(p) for p in paths]
-                        if paths else ['--include', '*'])
-        for j in self._run_annex_command_json(cmd, opts=opts):
+            if paths:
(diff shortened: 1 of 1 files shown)


yarikoptic:
for real-life -- I think it helped (didn't crash as fast), but there is something else which makes datalad very busy for very long, so I had to interrupt and rerun dumping logs into a file to possibly see where things are "busy".

yarikoptic:
it did help in RL, nothing seems to be broken, merging

PR #3076, opened 2018-12-07
WiP RF: bring revolution home

yarikoptic:
This pull request fixes #2926 

There is a long road ahead, that issue should provide details on what to be done.

1. This branch should receive periodic merges of `master` branch.  It will be merged into `master` whenever all checkboxes are marked "done".
2. If new changes from revolution to be introduced, bring them in into rf-revolution1-3rd branch first, using `tools/injest-revolution` or better `datalad rerun -m "Injest XXX from revolution" _run_revolution_injest`, where `_run_revolution_injest` is a tag available in this repo pointing to initial injestion invocation (see also #3074 about `--explicit` which you might need)
3. This (rf-revolution1) branch should always be ready to be merged into master, so not for adding broken new features to it.  It is recommended to bring new changes to this branch via a PR (i.e. to be merged into `rf-revolution1` branch)

- [x] no direct mode support
- [x] absorbed datalad-revolution as `datalad.revolution`
- [x] datalad.revolution.tests pass
    - [x] on appveyor
    - [x] on travis: remaining issue seems to be solely reporting of paths from repos that are accessed through a symlink. At first glance everything is actually happening as it should, but the reported `path` has the true location of a file, not the linked path. Needs a careful look what should happen in each of the four situations.
- [x] bring in replacement for `diff` from https://github.com/datalad/datalad-revolution/pull/58
- [ ] RF -core to use -revolution pieces and decide on how "joint" API should look like - checkboxes in #2926

Issue #2160, opened 2018-02-13
datalad diff --content


yarikoptic:
#### What is the problem?
`datalad diff` is great to review changes across hierarchy of datasets.  But I frequently miss ability to do diff of the content.  If, for starters, it could do it for at least text files (according to mime) -- that would already be super useful.

I have played with diff drivers for git/annex while comparing results on AFNI test battery run -- it is "cumbersome" to say the least.  Us providing a helping hand while dealing with annexed files could be of huge value here.

yarikoptic:
just randomly ran into what might come handy: https://github.com/juhakivekas/multidiff  (FWIW in Python as well)

mih:
From the perspective of `diff` it would be easy to add such a feature. It is practically just an alternative "result_renderer", as the results already contain all info. Here is an example of a file that I `git mv`'ed in a subdataset:

```
{
  "action": "diff", 
  "mode": 33188, 
  "mode_src": 33188, 
  "parentds": "/tmp/aggdemo/openfmri", 
  "path": "/tmp/aggdemo/openfmri/mike", 
  "path_src": "/tmp/aggdemo/openfmri/datapackage.json", 
  "perc_similarity": 91.0, 
  "refds": "/tmp/aggdemo", 
  "revision": "0690366164b8aa515b60fd647c58ffe63e396a3b", 
  "revision_src": "62dc679772dd12479249985d865977e76dcaf126", 
  "state": "renamed", 
  "status": "ok", 
  "type": "file", 
  "type_src": "file"
}
```

You can see that all revision info and path (changes) are represented. However, there is a whole new level of complexity being added by all the combinations of old/new version of annexed files present/absent.

FTR: The API for result renderers is

```
def custom_result_renderer(res, **kwargs):
```

where `res` is a result status dict (as shown above), and `kwargs` are all the args that the original command received (so it would get to see a `--content` flag too). The renderer is called once for each item that comes out of the generator, and a content diff renderen could be produced when all conditions are met: has to be a file (before and after), content is locally available, etc.

I am afraid that we will also have to add a few metric tons of formating and filtering options (see `git diff` for the attractor state ;-)

mih:
Same as #2812

mih:
[...]

PR #3600, opened 2019-08-13
RF:  more of common logistics into setup_support

yarikoptic:
I thought to provide some face lifting to our template (https://github.com/datalad/datalad-extension-template/issues/4) before starting to use it, but noted that real-life cases (extensions) diverge from the template quite a bit and have lots of duplication in setup.py.  I would advocate for us to absorb all common functionality within setup_support.py which we already copy around the extensions, and provide it within the template.

Well, ideally we should probably package/release it as `datalad-setup-support` or alike, but I am not sure if I am ready for yet another tiny packaging endeavor, so will be ok with plain copying it around the extensions for now.

If we agree that this is the way to go, then I will
- RF and see if anything missing the extensions
  - [ ] neuroimaging
  - [ ] container
  - [ ] crawler
  - [ ] template
  - ... not familiar with the others, so will leave it to respective developers/maintainers


mih:
Should we drop our own version handling and use versioneer?

yarikoptic:
> Should we drop our own version handling and use versioneer?

probably. I have no experience with it yet to make an informed decision and probably could be done as independent subsequent PR since this one is primarily moving code around, but FWIW nibabel is going that way too: https://github.com/nipy/nibabel/pull/786

setup_support.py
```diff
@@ -11,24 +11,27 @@
...
 import formatters as fmt
 
 
 def _path_rel2file(p):
     return opj(dirname(__file__), p)
 
 
-def get_version():
+def get_version(name):
     """Load version of datalad from version.py without entailing any imports
```
kyleam: docstring should be updated to drop hard-coded "datalad".

yarikoptic: adjusted and pushed

yarikoptic:
ok, introduced some breakage:
```
updating environment: 80 added, 0 changed, 0 removed
reading sources... [100%] usecases/simple_provenance_tracking                   
Warning, treated as error:
/home/travis/build/datalad/datalad/docs/source/cmdline.rst:13:toctree contains reference to nonexisting document u'generated/man/datalad'
Makefile:55: recipe for target 'html' failed
make: *** [html] Error 2
make: Leaving directory '/home/travis/build/datalad/datalad/docs'
```
will look into it later

kyleam:
> ok, introduced some breakage:

The problem comes from a bit [above](https://travis-ci.org/datalad/datalad/jobs/571561768#L1355):

```
  File "/home/travis/build/datalad/datalad/setup_support.py", line 111, in run
    format = cls(cmdname, ext_sections=sections, version=get_version())
```

So the `get_version` call in `BuildManPage` needs to be updated for the new `name` argument.

https://github.com/datalad/datalad/blob/40dac183bfae581dad72b05686be698a065c2271/setup_support.py#L111

yarikoptic:
d'oh -- thanks!  pushed the workaround in 652cd4d9803a6af2b0d0ce9703ddb71bd3461247

yarikoptic:
so, with the background failures now for both appveyor and travis (3.x) not easy to say if there is no side effects but it might be that we caught all the gotchas, thanks @kyleam for the review.
I guess I could proceed to try it (copy setup_support.py over) in the extensions (when get a moment) to see if anything is missing and/or needs to be changed

setup_support.py
```diff
@@ -100,7 +108,7 @@ def run(self):
                 cmdname = "{0}{1}".format(
                     'datalad ' if cmdname != 'datalad' else '',
                     cmdname)
-                format = cls(cmdname, ext_sections=sections, version=get_version())
+                format = cls(cmdname, ext_sections=sections, version=get_version(appname))
```
kyleam: Wouldn't this result in the manpages for extensions having the wrong version (datalad's rather than the extension's)?

yarikoptic: good catch!
- apparently we do not yet ship manpages for extensions' commands (I see none in datalad-container debian package at least, and `datalad` package itself doesn't ship the ones for extensions)
- I don't even know if they build fine in extensions ;)
- even if we did we stick to hardcoded "datalad" in those copies of setup_support.py
- indeed should be fixed properly here and better to provide the version of extension here than of datalad.
- ... I will look into checking/adding support for it later (not today) unless someone beats me to it

kyleam: > * I don't even know if they build fine in extensions ;)

Based on the [-container rtd page][0] and on running `python setup.py build_manpage` locally, I think they build fine.

[0]: https://datalad-container.readthedocs.io/en/latest/generated/man/datalad-containers-add.html

> * even if we did we stick to hardcoded "datalad" in those copies of setup_support.py

We stick to a hard-coded `appname` of "datalad" in `BuildManPage`, but `get_version` does appear to be adjusted for each extension that has setup_support.py:

https://github.com/datalad/datalad-container/blob/8cf9f9b250d58695a38b3335f59fe667566ff375/setup_support.py#L28-L34

So the current state is that an extension uses the extension's version in `BuildManPage`.  Running `python setup.py build_manpage` for -container generated a manpage having the expected footer:

```
datalad-containers-add 0.5.0                  2019-08-14                     datalad-containers-add(1)
```

Looking more closely at `get_version`, I actually think passing "datalad" to `get_version` would fail for extensions because `get_version` would expect "./datalad/version.py" to exist:

https://github.com/datalad/datalad/blob/652cd4d9803a6af2b0d0ce9703ddb71bd3461247/setup_support.py#L39-L40

setup_support.py
```diff
@@ -361,3 +369,78 @@ def _provide_lean_script_contents(
...
     return setup_kwargs
+
+
+def get_long_description_from_README():
+    """Read README.md, convert to .rst using pypandoc
+
+    If pypandoc is not available or fails - just output original .md
+    """
+    # PyPI doesn't render markdown yet. Workaround for a sane appearance
+    # https://github.com/pypa/pypi-legacy/issues/148#issuecomment-227757822
```
effigies: PyPI does render markdown these days and setuptools will alert it to do so [as of 38.6.0](https://setuptools.readthedocs.io/en/latest/setuptools.html#metadata), so you could do something like:

```Python
ret = {}
if setuptools.__version__ > '38.6.0':  # Probably want a more correct check than this
    ret['long_description'] = open(README).read()
    ret['long_description_content_type'] = 'text/markdown'
    return ret

# Convert or fall-back

return {'long_description': long_description}
```

If it's even worth bothering.

yarikoptic: Sweet, thanks! I will incorporate it

setup_support.py
yarikoptic: Thanks for the analysis!

yarikoptic: ok, pushed somewhat ugly (not worse than IMHO some other `datalad.extensions` handling points) but seems to be working workaround for all of this.  Now datalad-container does get their manpages generated properly
```shell
(git)hopa:~datalad/datalad-container[debian]git
$> rm -rf build/man && python setup.py build_manpage && for f in ./build/man/*.*; do head -n 1 $f; done
running build_manpage
.TH datalad containers-add 1 2019\-08\-14 "datalad containers-add 0.5.0"
.TH datalad containers-list 1 2019\-08\-14 "datalad containers-list 0.5.0"
.TH datalad containers-remove 1 2019\-08\-14 "datalad containers-remove 0.5.0"
.TH datalad containers-run 1 2019\-08\-14 "datalad containers-run 0.5.0"
```
with its setup.py being
```python
#!/usr/bin/env python

from setup_support import (
    BuildManPage,
    BuildRSTExamplesFromScripts,
    datalad_setup,
)


cmdclass = {
    'build_manpage': BuildManPage,
    'build_examples': BuildRSTExamplesFromScripts,
}


datalad_setup(
    # basic project properties can be set arbitrarily
    name="datalad_container",
    description="DataLad extension package for working with containerized environments",
    # datalad command suite specs from here
    install_requires=[
        'datalad>=0.11.5',
        'requests>=1.2',  # to talk to Singularity-hub
    ],
    extras_require={
        'devel-docs': [
            # used for converting README.md -> .rst for long_description
            'pypandoc',
            # Documentation
            'sphinx',
            'sphinx-rtd-theme',
        ]},
    cmdclass=cmdclass,
    entry_points = {
        # 'datalad.extensions' is THE entrypoint inspected by the datalad API builders
        'datalad.extensions': [
            # the label in front of '=' is the command suite label
            # the entrypoint can point to any symbol of any name, as long it is
            # valid datalad interface specification (see demo in this extension)
            'container=datalad_container:command_suite',
        ],
        'datalad.tests': [
            'container=datalad_container',
        ],
    },
)
```
which seems to be quite minimalistic

setup_support.py
yarikoptic: done

setup_support.py
yarikoptic: done

kyleam: I can confirm that your latest changes add the correct version to the manpage.  Thanks!

I did notice that the manpage looks a bit off.  A diff with your changes as the new version:

```diff
--- out.0	2019-08-15 10:08:07.731791152 -0400
+++ out.1	2019-08-15 10:08:19.400217070 -0400
@@ -1,7 +1,7 @@
-datalad-containers-list(1)              General Commands Manual             datalad-containers-list(1)
+datalad(containers-list)             datalad containers-list 0.5.0            datalad(containers-list)
 
 SYNOPSIS
-       datalad-containers-list [-h] [-d DATASET] [-r]
+       datalad containers-list [-h] [-d DATASET] [-r]
 
 DESCRIPTION
        List containers known to a dataset
@@ -22,4 +22,4 @@
 AUTHORS
         datalad is developed by The DataLad Team and Contributors <team@datalad.org>.
 
-datalad-containers-list 0.5.0                 2019-08-15                    datalad-containers-list(1)
+2019-08-15                                         1                          datalad(containers-list)
```

The synopsis change looks like an improvement because a hyphenated `datalad-containers-list` isn't valid, but the other lines suggest that `container-list` is taken as the section.

Perhaps we should update `_mk_title` to do something like this?

```diff
diff --git a/formatters.py b/formatters.py
index 3a997db..1b93e52 100644
--- a/formatters.py
+++ b/formatters.py
@@ -69,8 +69,9 @@ def _mk_synopsis(self, parser):
         return usage
 
     def _mk_title(self, prog):
-        name_version = "\"{0} {1}\"".format(prog, self._version)
-        return '.TH {0} {1} {2} {3}\n'.format(prog, self._section,
+        prog_hyph = "-".join(prog.split())
+        name_version = "\"{0} {1}\"".format(prog_hyph, self._version)
+        return '.TH {0} {1} {2} {3}\n'.format(prog_hyph, self._section,
                                               self._today, name_version)
 
     def _make_name(self, parser):
```

Diffing with the output before this PR, that gives

```diff
--- out.0	2019-08-15 10:08:07.731791152 -0400
+++ out.2	2019-08-15 10:08:26.224466367 -0400
@@ -1,7 +1,7 @@
 datalad-containers-list(1)              General Commands Manual             datalad-containers-list(1)
 
 SYNOPSIS
-       datalad-containers-list [-h] [-d DATASET] [-r]
+       datalad containers-list [-h] [-d DATASET] [-r]
 
 DESCRIPTION
        List containers known to a dataset
```

setup_support.py
```diff
@@ -361,3 +411,100 @@ def _provide_lean_script_contents(
...
+      for newer setuptools which support uploading of markdown as is.
+    """
+    # PyPI used to not render markdown. Workaround for a sane appearance
+    # https://github.com/pypa/pypi-legacy/issues/148#issuecomment-227757822
+    # is still in place for older setuptools
+
+    README = opj(dirname(__file__), 'README.md')
+
+    ret = {}
+    if LooseVersion(setuptools.__version__) > '38.6.0':
```
effigies: Woops. Sorry about the sloppy suggestion.

```suggestion
    if LooseVersion(setuptools.__version__) >= '38.6.0':
```

yarikoptic: Thanks -- I just amended the last commit, will force push later

setup_support.py
yarikoptic: `<rant>`
oh crap - there is also formatters.py which is copied and changed around - I have not even looked into that one yet, and it does have difference to version (probably just outdated) in extensions ... 
It does feel that we are just reaching the point of making `datalad-support` (or what other name?) package now since situation is really far from optimal :-/  BUT we actually cannot since setuptools doesn't have a way to declare "build dependencies" . IIRC it is some kind of a new builder (can't find ATM) has it, and there is https://www.python.org/dev/peps/pep-0518/ and https://github.com/pypa/pipenv/issues/3651 .  But it feels that even if we make such a package we end up just copying it around.  At least may be it would get proper versioning information to check were it is outdated etc.
`</rant>`

yarikoptic: > Diffing with the output before this PR, that gives
> 
> ```diff
> --- out.0	2019-08-15 10:08:07.731791152 -0400
> +++ out.2	2019-08-15 10:08:26.224466367 -0400
> @@ -1,7 +1,7 @@
>  datalad-containers-list(1)              General Commands Manual            
> ```

I am confused a bit -- so it reverts back to suboptimal?

If I symlink formatters.py from datalad core as well (no changes to it there), then things seems to be rendered ok for me:

```shell
(git)hopa:~datalad/datalad-container[debian]git
$> rm -rf build/man && python setup.py build_manpage && for f in ./build/man/*.*; do man $f | head -n 8; done 
running build_manpage
datalad containers-add(1)                                            datalad                                            datalad containers-add(1)

NAME
       datalad containers-add - add a container to a dataset

SYNOPSIS
       datalad containers-add [-h] [-u URL] [-d DATASET] [--call-fmt FORMAT] [-i IMAGE] [--update] NAME

datalad containers-list(1)                                           datalad                                           datalad containers-list(1)

NAME
       datalad containers-list - list containers known to a dataset

SYNOPSIS
       datalad containers-list [-h] [-d DATASET] [-r]

datalad containers-remove(1)                                         datalad                                         datalad containers-remove(1)

NAME
       datalad containers-remove - remove a known container from a dataset

SYNOPSIS
       datalad containers-remove [-h] [-d DATASET] [-i] NAME

datalad containers-run(1)                                            datalad                                            datalad containers-run(1)

NAME
       datalad containers-run - drop-in replacement of 'run' to perform containerized command execution

SYNOPSIS
       datalad containers-run [-h] [-n NAME] [-d DATASET] [-i PATH] [-o PATH] [-m MESSAGE] [--expand WHICH] [--explicit] [--sidecar yes|no] ...
```

yarikoptic:
ok -- I moved both setup_support (as setup.py) and formatters under `datalad_build_support/` which I made into a package (added `__init__.py`).  Seems to work to build datalad itself and datalad-container (if entire directory just symlinked, I guess would work if copied). 
With and while at that, may be it is indeed to add versioneer.py  to just be over with this mess ;)  might look into it in upcoming days unless someone beats me to it.

setup_support.py
kyleam: > I am confused a bit -- so it reverts back to suboptimal?

The main point I was trying to make is that, with your setup_support.py change and my suggested tweak to `_mk_title`, running `python setup.py build_manpage` from the -container repo produces the same rendered manpage output, aside from the more correct synopsis line.  Without the `_mk_title` tweak, there is the weirdness with "containers-list" being formatted as the section.

> If I symlink formatters.py from datalad core as well (no changes to it there), then things seems to be rendered ok for me:

Good catch!  formatters.py from datalad and -container have indeed diverged (woohoo for copying files :):

```
# -container repo
% git describe
0.5.0
% git diff --no-index --stat formatters.py ../datalad/formatters.py
 formatters.py => ../datalad/formatters.py | 18 +++++++++++-------
 1 file changed, 11 insertions(+), 7 deletions(-)
```

I haven't inspected the differences, but using datalad core's version of formatters.py from the -container repo, I don't see the "containter-list as section" issue.  So, as your comment suggests, all seems good there!

(With the datalad master version of formatter.py, I'm not seeing the version in the footer.  I'm assuming that is an unintended change,   but the behavior is shared by both datalad master and this PR, so I don't think we should worry about it in this PR.)

yarikoptic: > (With the datalad master version of formatter.py, I'm not seeing the version in the footer. I'm assuming that is an unintended change, but the behavior is shared by both datalad master and this PR, so I don't think we should worry about it in this PR.)

NICE CATCH!  reproduced.  I would just fix here -- will look later at what is the diff -- I bet we broke it with those formatters changes in the core which aren't in the extensions' copy

kyleam:
> I moved both setup_support (as setup.py) and formatters under `datalad_build_support/`

Nice.  This move would also make using `git-subtree` possible (along the lines of what I outline in [this comment][0]), which in my view would be a big improvement over copying the files and avoids the complication of trying to work in a build dependency.

[0]: https://github.com/datalad/datalad/pull/3076#issuecomment-445371086

yarikoptic:
oh, `git-subtree`! indeed.  But may be it is even better to make it into a submodule?  then it would get its own versioned tags/releases/etc (which is what subtree wouldn't give if I get it right)?

yarikoptic:
cons of having a submodule
- anyone developing would need to run `git submodule update --init datalad_build_support` or alike
- investigating diff across branches if there is any difference could be trickier -- but may be it would just push us to improve `datalad diff` finally (e.g. #2160)

yarikoptic:
BTW, I don't think versioneer can leave under `datalad_build_support/` - seems need to go along side with `.git/` and `setup.py` (doc calls those "siblings" ;) ).

yarikoptic:
<details>
<summary>pushed the fix (e0bd96234bd7008c6d4c7d23f87d9661e688cdce) for version being gone  -- was double quoted</summary> 

```shell
(git)hopa:~datalad/datalad[rf-more_setup_support]git
$> rm -rf build/man && python setup.py build_manpage && for f in ./build/man/*.*; do man $f | tail -n 2; done 
running build_manpage                                                                         
 
datalad 0.11.6                                    2019-08-15                                       datalad(1)

datalad add 0.11.6                                2019-08-15                                   datalad add(1)

datalad add-archive-content 0.11.6                2019-08-15                   datalad add-archive-content(1)

datalad add-readme 0.11.6                         2019-08-15                            datalad add-readme(1)

datalad addurls 0.11.6                            2019-08-15                               datalad addurls(1)
troff: <standard input>:35: warning: macro 'datalad.metadata.nativetype'' not defined (possibly missing space after 'da')
troff: <standard input>:49: warning: macro 'datalad_core',' not defined (possibly missing space after 'da')

datalad aggregate-metadata 0.11.6                 2019-08-15                    datalad aggregate-metadata(1)

datalad annotate-paths 0.11.6                     2019-08-15                        datalad annotate-paths(1)

datalad check-dates 0.11.6                        2019-08-15                           datalad check-dates(1)

datalad clean 0.11.6                              2019-08-15                                 datalad clean(1)

datalad clone 0.11.6                              2019-08-15                                 datalad clone(1)

datalad create 0.11.6                             2019-08-15                                datalad create(1)

```
</details>

kyleam:
> oh, `git-subtree`! indeed. But may be it is even better to make it into a submodule?

I strongly prefer -subtree here.

> then it would get its own versioned tags/releases/etc (which is what subtree wouldn't give if I get it right)?

You'd split the datalad_build_support/ into a separate repo that would have tags/releases/etc.  datalad core and extensions would pull that repo into a subtree.  When we are interested in updating the subtree from the upstream repo we can do so, but otherwise we can work with the tree as we would any other subdirectory in the extension repo.

> cons of having a submodule
> 
> * anyone developing would need to run `git submodule update --init datalad_build_support` or alike

Yes, that's a big one.  For a directory as light as datalad_build_support/, I don't see any issue with the git-subtree approach of duplicating the tree objects in datalad core as well as extension repos.

> * investigating diff across branches if there is any difference could be trickier -- but may be it would just push us to improve `datalad diff` finally (e.g. #2160)

Also not an issue with subtrees :]

yarikoptic:
Ok, subtree it is then

yarikoptic:
Let's merge and proceed from there.
Eventually (when we probably stop carrying about easy backports) we might beefup our build setup following @effigies steps and Michael's adoption in https://github.com/datalad/git-annex-ria-remote

PR #3620, opened 2019-08-25
Debian packaging update

mih:
Given the draft stage of the prev. PR lots of conflicts accumulated. It would be good to have this change merged. It makes the control file compatible with `cme` and minimizes the diff to its output.

- python-six no longer has versioned dep (was only relevant for jessie)
- standard order
- whitespace standardization


mih:
Offline approval from @yarikoptic

PR #3640, opened 2019-09-01
TST: Reenable testrepo setup for windows tests

mih:
With the testrepo setup presumably usable on windows now, reenable the related tests to see how much fixing is TODO.


yarikoptic:
> With the testrepo setup presumably usable on windows now,

what has fixed it?
here is my elderly attempt 
https://github.com/datalad/datalad/pull/2870/files#diff-b7612e7046f9ade40f3a28b003758e32
which (#2870) we closed as stale since I  have not finalized it (IIRC more tests started to break)

mih:
Ah, misunderstood you then. So lets let in finish and get an updated assessment.

PR #3659, opened 2019-09-10
Attempt to establish benchmarks across a set of different types of "heavy" datasets

yarikoptic:
A semi-failed attempt to introduce benchmarking of common operations across a collection of datasets.
With this approach I thought to take advantage of multiple inheritance and mix in different datasets and states while then mix-ing in also corresponding benchmarks.

It is ugly and probably not the way to go because

- `setup_cache` taking care about creating a tarball with a dataset instance is identified by asv simply by location in the code -- leads to fragile need to replicate `setup_cache` in each subclass: https://github.com/airspeed-velocity/asv/issues/880
- we cannot make asv  to ignore a to be mixed in (like `TypicalOps` here) class defining benchmarks, so it would try running them and fail

Since then I discovered that there are ways to "parametrize" benchmarks: https://asv.readthedocs.io/en/stable/writing_benchmarks.html#parameterized-benchmarks .  So similarly to what we do in `@with_testrepos` I guess we just need to establish some "lazy" benchmark datasets generator, and then tests classes would be parametrized with the name of the dataset to use, and ask `self.datasets[name]` to get it in the `setup`, possibly eliminating the need for `setup_cache`.  We would just need to track them and remove `atexit`.  Really close to what we already do in `@with_testrepos`

So submitting this PR as a draft and not intended to be merged

### DIFF
--- a/benchmarks/api.py
+++ b/benchmarks/api.py
@@ -7,6 +7,7 @@
 # ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
 """Benchmarks of the datalad.api functionality"""
 
+from glob import glob
 from os.path import join as opj
 
 try:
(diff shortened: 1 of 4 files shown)


yarikoptic:
let's close for now

PR #3667, opened 2019-09-13
don't annex empty files after cfg_text2git,  fixes #3663

### LINKED ISSUE #3663: FYI: empty files are annexed; inconsistent mimetype reporting
I'm just noting a behavior that confused me:

Let's say I am creating an empty file In a dataset configured with ``text2git`` with ``touch somefile`` (no extension, no content).
The ``mimetype`` command reports this file as ``text/plain``:

```sh
╰─➤ datalad create -c text2git blubb
[INFO   ] Creating a new annex repo at /tmp/blubb 
create(ok): /tmp/blubb (dataset)
╭─adina@muninn /tmp
╰─➤ cd blubb; touch somefile
╭─adina@muninn /tmp/blubb on master
╭─adina@muninn /tmp/blubb on master
╰─➤ mimetype somefile
somefile: text/plain

```

A subsequent ``datalad save`` annexes the file:

```
╭─adina@muninn /tmp/blubb on master
╰─➤ datalad save -m "blabla"
add(ok): somefile (file)
save(ok): . (dataset)
action summary:
  add (ok: 1)
  save (ok: 1)
╭─adina@muninn /tmp/blubb on master
╰─➤ ls -l
total 4
lrwxrwxrwx 1 adina adina 108 Sep 12 10:30 somefile -> .git/annex/objects/2W/kW/MD5E-s0--d41d8cd98f00b204e9800998ecf8427e/MD5E-s0--d41d8cd98f00b204e9800998ecf8427e
```

The inconsistency/my confusion arise from the fact that ``.gitattributes`` is configured to not regard any text files as ``largefiles`` by ``text2git``

```
* annex.backend=MD5E
**/.git* annex.largefiles=nothing
* annex.largefiles=(not(mimetype=text/*))
```

The rules for largefiles in the docs of Git-annex help to clarify what I was missing:

> mimetype=glob
> Looks up the MIME type of a file, and checks if the glob matches it.
> For example, "mimetype=text/*" will match many varieties of text files, including "text/plain", but  also "text/x-shellscript", "text/x-makefile", etc.
> The MIME types are the same that are displayed by running file --mime-type
> This is only available to use when git-annex was built with the MagicMime build flag.

If I actually run ``file --mime-type somefile`` it is not reported as a text file anymore:

```
╰─➤ file --mime-type somefile
somefile: inode/symlink
```

That's just as an FYI. However, is it actually useful/intended to annex files with a size of 0?

adswa:
As noted in #3663, in a dataset configured with ``cfg_text2git``, empty files are being annexed. This PR adds the additional rule to not regard empty files (mimetype ``inode/x-empty``) as a largefile to the ``.gitattributes`` configuration ``cfg_text2git`` provides.
Additionally, it adds a small test for this.

It's WIP because I want to see whether any test fail as @kyleam noted in #3663:

> I suppose one tricky part about testing cfg_text2git is that the behavior depends on git-annex being built with the MagicMime flag. But assuming MagicMime support should be fine because neurodebian's git-annex has it and it looks like conda's does now too. If it ends up being a problem down the road, we can skip the test based on the git annex version output.

### DIFF
--- a/datalad/distribution/create.py
+++ b/datalad/distribution/create.py
@@ -343,9 +343,9 @@ class Create(Interface):
                 attrs = tbrepo.get_gitattributes('.')
                 # some basic protection against useless duplication
                 # on rerun with --force
-                if not attrs.get('.', {}).get('annex.largefiles', None) == '(not(mimetype=text/*))':
+                if not attrs.get('.', {}).get('annex.largefiles', None) == '(not((mimetype=text/*)or(smallerthan=0.001kb)))':
                     tbrepo.set_gitattributes([
-                        ('*', {'annex.largefiles': '(not(mimetype=text/*))'})])
+                        ('*', {'annex.largefiles': '(not((mimetype=text/*)or(smallerthan=0.001kb)))'})])
                     add_to_git.append('.gitattributes')
 
         if native_metadata_type is not None:
--- a/datalad/distribution/create.py
+++ b/datalad/distribution/create.py
@@ -343,9 +343,9 @@ class Create(Interface):
                 attrs = tbrepo.get_gitattributes('.')
                 # some basic protection against useless duplication
                 # on rerun with --force
-                if not attrs.get('.', {}).get('annex.largefiles', None) == '(not((mimetype=text/*)or(smallerthan=0.001kb)))':
+                if not attrs.get('.', {}).get('annex.largefiles', None) == '(not(mimetype=text/*)and(largerthan=0))':
                     tbrepo.set_gitattributes([
-                        ('*', {'annex.largefiles': '(not((mimetype=text/*)or(smallerthan=0.001kb)))'})])
+                        ('*', {'annex.largefiles': '(not(mimetype=text/*)and(largerthan=0))'})])
                     add_to_git.append('.gitattributes')
 
         if native_metadata_type is not None:
--- a/datalad/interface/tests/test_run_procedure.py
+++ b/datalad/interface/tests/test_run_procedure.py
@@ -34,6 +34,7 @@ from datalad.tests.utils import OBSCURE_FILENAME
 from datalad.tests.utils import on_windows
 from datalad.tests.utils import known_failure_direct_mode
 from datalad.tests.utils import known_failure_windows
+from datalad.tests.utils import skip_if_on_windows
 from datalad.distribution.dataset import Dataset
 from datalad.support.exceptions import (
     CommandError,
@@ -347,3 +348,18 @@ def test_quoting(path):
             "datalad run-procedure just2args \"with ' sing\" 'with \" doub'")
         with assert_raises(CommandError):
             runner.run("datalad run-procedure just2args 'still-one arg'")
+
+@skip_if_on_windows
+@with_tempfile
+def test_text2git_empty(path):
+    """
+    Tests that empty files are not annexed in a ds configured with text2git.
+    """
+    ds = Dataset(path).create(force=True)
+    ds.run_procedure('cfg_text2git')
+    ok_clean_git(ds.path)
+    # create an empty file, no extension
+    open(op.join(path, 'emptyfile'), 'a').close()
+    ds.save(message="add empty file")
+    # check that it's not annexed
+    assert_false(ds.repo.is_under_annex("emptyfile"))
--- a/datalad/resources/procedures/cfg_text2git.py
+++ b/datalad/resources/procedures/cfg_text2git.py
@@ -10,7 +10,7 @@ ds = require_dataset(
     check_installed=True,
     purpose='configuration')
 
-annex_largefiles = '(not(mimetype=text/*))'
+annex_largefiles = '(not((mimetype=text/*)or(smallerthan=0.001kb)))'
 attrs = ds.repo.get_gitattributes('*')
 if not attrs.get('*', {}).get(
         'annex.largefiles', None) == annex_largefiles:
--- a/datalad/resources/procedures/cfg_text2git.py
+++ b/datalad/resources/procedures/cfg_text2git.py
@@ -10,7 +10,7 @@ ds = require_dataset(
     check_installed=True,
     purpose='configuration')
 
-annex_largefiles = '(not((mimetype=text/*)or(smallerthan=0.001kb)))'
+annex_largefiles = '(not(mimetype=text/*)and(largerthan=0))'
 attrs = ds.repo.get_gitattributes('*')
 if not attrs.get('*', {}).get(
         'annex.largefiles', None) == annex_largefiles:
--- a/datalad/distribution/tests/test_create.py
+++ b/datalad/distribution/tests/test_create.py
@@ -334,7 +334,7 @@ def test_create_text_no_annex(path):
     import re
     ok_file_has_content(
         _path_(path, '.gitattributes'),
-        content='\* annex\.largefiles=\(not\(mimetype=text/\*\)\)',
+        content='\* annex\.largefiles=\(not\(\(mimetype=text/\*\)or\(smallerthan=0.001kb\)\)\)',
         re_=True,
         match=False,
         flags=re.MULTILINE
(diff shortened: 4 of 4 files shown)


adswa:
Travis is happy, but in Appveyor it fails with

```
bad annex.largefiles configuration: Parse failure: "mimetype" not supported; not built with MagicMime support Parse failure: "mimetype" not supported; not built with MagicMime support
```

and the initial ``datalad wtf`` shows that ``MagicMime`` is not part of the build flags:
```
[...]
## git-annex 
  - build flags: 
    - Assistant
    - Webapp
    - Pairing
    - S3
    - WebDAV
    - TorrentParser
    - Feeds
    - Testsuite
  - dependency versions: 
    [...]
```

I guess I should decorate the test to skip it on Windows (``@known_failure_windows``) or is there a decorator specifically for Git-annex build flags? (or something else I should do?)

mih:
skip_on_windows is the right one (if i spelled that right). we know the windows build doesnt have it

kyleam:
> skip_on_windows is the right one (if i spelled that right)

Missing an if: skip_if_on_windows

datalad/interface/tests/test_run_procedure.py
```diff
@@ -339,3 +340,18 @@ def test_quoting(path):
...
+@with_tempfile
+def test_text2git_empty(path):
+    """
+    Tests that empty files are not annexed in a ds configured with text2git.
+    """
+    ds = Dataset(path).create(force=True)
+    ds.run_procedure('cfg_text2git')
+    ok_clean_git(ds.path)
+    # create an empty file, no extension
+    open(op.join(path, 'emptyfile'), 'a').close()
```
kyleam: Hmm, the 'a' rather than 'w' made me pause.  We're in a new empty directory though, so either way is fine.

Just a note: `@with_tree` is nice for pre-populating the test directory with files and contents.  In this case, I don't using it really makes the test any more readable (and would make the `ok_clean_git` call more complicated).

kyleam:
The AppVeyor failure is [distribution.tests.test_update.test_multiway_merge][0], so almost certainly unrelated.

Based on looking at the last passing build, it's probably due to a switch of git-annex [from][1] 7.20190731-gc131c0f0c [to][2] 7.20190912-g05bc37910.  I triggered an appveyor run in a scratch branch and will open an issue if the failure shows up there as well.

[0]: https://ci.appveyor.com/project/mih/datalad/builds/27398021/job/o5ae734n4sxubf7f#L1386
[1]: https://ci.appveyor.com/project/mih/datalad/builds/27389604/job/mdymsm41isx4mt3d?fullLog=true#L523
[2]: https://ci.appveyor.com/project/mih/datalad/builds/27398021/job/o5ae734n4sxubf7f#L513

datalad/resources/procedures/cfg_text2git.py
```diff
@@ -10,7 +10,7 @@
     check_installed=True,
     purpose='configuration')
 
-annex_largefiles = '(not(mimetype=text/*))'
+annex_largefiles = '(not((mimetype=text/*)or(mimetype=inode/x-empty)))'
```
yarikoptic: Why bother with mimetype if it could be directly based on the file size? We did see mimetype detection getting new ideas/bugs so why not to KISS?

kyleam:
> yarikoptic added the pick up to 0.11.x 

FWIW my view is that, if it's not a bug fix, it doesn't need to be on top of 0.11.x.  But if you plan to cherry pick it there anyway, I have a strong preference that you instead just request that the series be rebased onto 0.11.x.

datalad/resources/procedures/cfg_text2git.py
kyleam: > Why bother with mimetype if it could be directly based on the file size?

I'm guessing just because it is what flowed out of the discussion in gh-3663.  I'm fine going with the size condition.

yarikoptic: This comment confused me into thinking that the size comparison was considered and I missed it while glancing over that discussion. I reviewed it again and found no explicit consideration. Did I miss it or this ambigouse comment wasn't intended to be informative? ;-)

yarikoptic:
Even if it is not a straight bug fix, it would lead to difference in datasets behavior between datasets. In 0.11.x we obtain the same behavior with a dedicated command line option to `create`. I don't remember ATM if we had this procedure. If we did, indeed better to rebase, @adswa could you please check?

datalad/resources/procedures/cfg_text2git.py
kyleam: > Did I miss it or this ambigouse comment wasn't intended to be informative? ;-)

By saying it "flowed out of the discussion" I intended to express that we reached that patch in a way that was dependent on the course of the discussion.  Roughly "an empty file isn't caught by mimetype=text/*" -> "oh, file --mime-type reports inode/x-empty" -> "ok, we could add that condition".  IOW, nobody (and particularly me who suggested the patch) stepped back and considered doing it without mimetype.

In general, given that human communication is lossy, I'd appreciate if you'd be more charitable and---rather than suggest that I was trying to be uninformative---assume that the information wasn't conveyed because it was poorly expressed or misread or some combination of both.

adswa:
Yes, will check, thanks for the suggestion. Am traveling ATM, but will get to it in the next 24h.

adswa:
If I understand it correctly, in ``0.11.x``, there is both the ``cfg_text2git`` procedure, and a dedicated ``--text-no-annex`` flag for ``create``:

https://github.com/datalad/datalad/blob/5926b241a90fad6276375071dd1e0910e6bad574/datalad/distribution/create.py#L342-L349

I've for now changed the rule to file size, but I'm not sure whether I should rebase onto ``0.11.x`` (and if so, whether I should add the file size rule to the ``--text-no-annex`` conditional?). Could you let me know? Also, please let me know whether I should squish all of these commits together. thanks!

yarikoptic:
> I've for now changed the rule to file size, but I'm not sure whether I should rebase onto `0.11.x` (and if so, whether I should add the file size rule to the `--text-no-annex` conditional?). Could you let me know? 

yes, please!

> Also, please let me know whether I should squish all of these commits together.

up to you.  If rebased -- I don't care since it would just all get merged and no cherry picking would be needed (where a single commit would be preferable)

datalad/resources/procedures/cfg_text2git.py
yarikoptic: > In general, given that human communication is lossy, I'd appreciate if you'd be more charitable and---rather than suggest that I was trying to be uninformative---assume that the information wasn't conveyed because it was poorly expressed or misread or some combination of both.

agree 100%
yeap -- let's us all be more charitable! ;)

adswa:
I tried my best to rebase -- could anyone check whether I succeeded?

kyleam:
> I tried my best to rebase -- could anyone check whether I succeeded?

You did.

mih:
I just found https://github.com/datalad/datalad/issues/3361 which points into a different direction for more robust operation across file types.

datalad/distribution/create.py
```diff
@@ -343,9 +343,9 @@ def __call__(
                 attrs = tbrepo.get_gitattributes('.')
                 # some basic protection against useless duplication
                 # on rerun with --force
-                if not attrs.get('.', {}).get('annex.largefiles', None) == '(not(mimetype=text/*))':
+                if not attrs.get('.', {}).get('annex.largefiles', None) == '(not((mimetype=text/*)or(smallerthan=0.001kb)))':
                     tbrepo.set_gitattributes([
-                        ('*', {'annex.largefiles': '(not(mimetype=text/*))'})])
+                        ('*', {'annex.largefiles': '(not((mimetype=text/*)or(smallerthan=0.001kb)))'})])
```
yarikoptic: IMHO this simpler version would be better.

```
(not(mimetype=text/*) and (largerthan=0))
```
I will push the tune up

PR #3677, opened 2019-09-15
DOC: PyPi classifier proposal

mih:
This adds an opinionated selection of PyPi classifiers that I am happy to debate and alter as you see fit.

This change touches upon an issue related to the upcoming removal of PY2 support that was brought up in https://github.com/datalad/datalad/pull/3629#issuecomment-527197881

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -156,6 +156,30 @@ entry_points.update({
     ]})
 setup_kwargs['entry_points'] = entry_points
 
+classifiers = [
+    'Development Status :: 5 - Production/Stable',
+    'Environment :: Console',
+    'Intended Audience :: Developers',
+    'Intended Audience :: Education',
+    'Intended Audience :: End Users/Desktop',
+    'Intended Audience :: Science/Research',
+    'License :: DFSG approved',
+    'License :: OSI Approved :: MIT License',
+    'Natural Language :: English',
+    'Operating System :: POSIX',
+    'Programming Language :: Python :: 3 :: Only',
+    'Programming Language :: Unix Shell',
+    'Topic :: Communications :: File Sharing',
+    'Topic :: Education',
+    'Topic :: Internet',
+    'Topic :: Other/Nonlisted Topic',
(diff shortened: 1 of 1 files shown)


mih:
Thanks @kyleam for the feedback and for adjusting the PY2 PR. Given that the appveyor failure is unrelated, I will merge this now.

I'll add a note to #3629 in a second.

setup.py
```diff
@@ -156,6 +156,30 @@
...
+    'Development Status :: 5 - Production/Stable',
+    'Environment :: Console',
+    'Intended Audience :: Developers',
+    'Intended Audience :: Education',
+    'Intended Audience :: End Users/Desktop',
+    'Intended Audience :: Science/Research',
+    'License :: DFSG approved',
+    'License :: OSI Approved :: MIT License',
+    'Natural Language :: English',
+    'Operating System :: POSIX',
```
yarikoptic: I wonder if that is anyhow used by PyPi/pip to decide/limit installations on only POSIX compliant systems... If we list it I think we should then add `MacOS :: MacOS X` and `Microsoft :: Windows` at least. and may be even `Unix`?

mih:
MacOSX is a POSIX OS. I intentionally chose that to say "in principle agnostic, but...". I did not want to include/advertise Windows, because we are not there yet.

setup.py
mih: MacOSX is a POSIX OS. I intentionally chose that to say "in principle agnostic, but...". I did not want to include/advertise Windows, because we are not there yet.

yarikoptic: I just looked at the hierarchy in the side bar on https://pypi.org/search/?c=Operating+System+%3A%3A+POSIX+%3A%3A+GNU+Hurd where OSX has its own section. I am ok to no Ad for Windows (yet)

mih: Indeed. Searching for MacOS does not give any hits for packages that are 'POSIX' alone. So we need to add that. Thx for checking.

PR #3678, opened 2019-09-15
DOC: Orthogonalize docs with handbook content

mih:
Concept: All docs remaining in here are technical and developer-focused. The handbook has acquired all previously existing content and use case descriptions, and is now a complete and more comprehensive replacement for any user/usage-focused information.

This commit strips away anything that is duplicate, while keeping executable code that can still serve a purpose in the test battery.

In a future pass through the docs, and after the structure and content of the handbook has matured a bit more, content here can be further refined and more (and more useful) cross-references can be added.

Potential TODOs (please give feedback):

- [ ] the content of chapter [basic principles](http://docs.datalad.org/en/latest/basics.html) is essentially contained in the handbook already, hence could be removed too (as it is not developer-focused). The only two things missing are:
  - [ ] info on ID types (this should become a dedicated chapter on the general metadata/identification approach that should be approached in the metalad-context (note to self @mih)
  - [ ] "API principles", which has only few principles, and is mostly an assorted collection of information otherwise -- which might be useful to place on a cheat sheet https://github.com/datalad-handbook/book/issues/139 (ping @adswa for awareness)


mih:
OK, let's go with this state for now.

PR #3680, opened 2019-09-16
ENH+TST: cfg_yoda will commit selectively and fail on dirty

### LINKED ISSUE #3666: `cfg_yoda` saves any untracked content
Found while composing https://github.com/datalad-handbook/book/pull/158#discussion_r324167794

The procedure should selectively save the pieces it touched.

In general I wonder, whether it is a good idea to have `run-procedure` operate on dirty datasets (by default). It makes it impossible to tell which changes are really coming from the procedure, unless procedure authors (which could be anyone) are mindful of this setup.

mih:
Before it would save the entire state of a dataset. Fixes #3666

Adjust and extend tests to cover the new behavior.

### DIFF
--- a/datalad/resources/procedures/cfg_yoda.py
+++ b/datalad/resources/procedures/cfg_yoda.py
@@ -15,6 +15,30 @@ ds = require_dataset(
     check_installed=True,
     purpose='YODA dataset setup')
 
+to_modify = [
+    ds.pathobj / 'code' / 'README.md',
+    ds.pathobj / 'code' / '.gitattributes',
+    ds.pathobj / 'README.md',
+    ds.pathobj / 'CHANGELOG.md',
+    ds.pathobj / '.gitattributes',
+]
+
+dirty = [
+    s for s in ds.status(
+        to_modify,
+        result_renderer='disabled',
+        return_type='generator',
+    )
+    if s['state'] != 'clean'
+]
+
+if dirty:
+    raise RuntimeError(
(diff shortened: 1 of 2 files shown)


mih:
Unreleated appveyor failure.

datalad/resources/procedures/cfg_yoda.py
```diff
@@ -15,6 +15,30 @@
...
 
+to_modify = [
+    ds.pathobj / 'code' / 'README.md',
+    ds.pathobj / 'code' / '.gitattributes',
+    ds.pathobj / 'README.md',
+    ds.pathobj / 'CHANGELOG.md',
+    ds.pathobj / '.gitattributes',
+]
+
+dirty = [
```
yarikoptic: here it probably doesn't matter either way, but IMHO better to use generators more to gain some efficiency.  E.g. making here it into a `()`, and below condition into `if any(dirty):`

mih: Hmm, but wouldn't the generator be consumed by `any()` and leave me no info to put in the exception?

mih:
OK, let's move.

datalad/resources/procedures/cfg_yoda.py
yarikoptic: ah, right. sorry I have missed that it was used below!  Sorry for the noise, disregard original comment ;)

Issue #3637, opened 2019-08-31
ssh calls on win10 broken


mih:
```
(base) C:\Users\mih>datalad -l 1 sshrun localhost "uname -a"
[DEBUG] Command line args 1st pass. Parsed: Namespace() Unparsed: ['sshrun', 'localhost', 'uname -a']
[DEBUG] Discovering plugins
[Level 5] Importing module datalad.support.sshrun
[DEBUG] Building doc for <class 'datalad.support.sshrun.SSHRun'>
[Level 5] Finished setup_parser
[DEBUG] Parsing known args among ['C:\\Users\\mih\\Miniconda3\\Scripts\\datalad', '-l', '1', 'sshrun', 'localhost', 'uname -a']
[Level 9] Running: ['git', 'config', '-z', '-l', '--show-origin']
[Level 8] Finished running ['git', 'config', '-z', '-l', '--show-origin'] with status 0
[DEBUG] sshrun invoked: login='localhost', cmd='uname -a', port=None, options=None, ipv4=False, ipv6=False, no_stdin=False
[Level 5] Parsed ri ssh://localhost into fields {'path': '', 'scheme': 'ssh', 'username': '', 'password': '', 'hostname': 'localhost', 'port': '', 'query': '', 'fragment': ''}
[Level 5] Detected url ri
[Level 9] Running: ['git', 'version']
[Level 8] Finished running ['git', 'version'] with status 0
[Level 9] Running: ['git', 'config', '-z', '-l', '--show-origin']
[Level 8] Finished running ['git', 'config', '-z', '-l', '--show-origin'] with status 0
[Level 5] Found 1 previous connections
[Level 9] Running: ['ssh', '-o', 'ControlPath="C:\\Users\\mih\\AppData\\Local\\datalad.org\\datalad\\Cache\\sockets\\9558c181"', 'localhost', "sh -e -c 'dirname $(readlink -f $(which git-annex-shell))'"]
[Level 9] Failed to run ['ssh', '-o', 'ControlPath="C:\\Users\\mih\\AppData\\Local\\datalad.org\\datalad\\Cache\\sockets\\9558c181"', 'localhost', "sh -e -c 'dirname $(readlink -f $(which git-annex-shell))'"] under None. Exit code=255. out= err=command-line line 0: Missing argument.
|
[...]

Issue #2575, opened 2018-05-27
SSH Connection handling completely broken on Windows


mih:
It cannot open any connection. The commands that it build are fine after #2574 is fixed, but nothing works. The same command ran manually on the cmdline outside datalad works fine (even with the same paths), so it seems to be a genuine datalad issue.

mih:
2.5 years later, trying to clone a dataset via SSH on windows (arguably a possible first point of contact with a user). The source dataset was created by `create-sibling` on a Debian server. Clone is confirmed to work with `git` on windows, and `datalad` on a different Debian box. Both times using `master` of today.

```
(base) C:\Users\mih>datalad clone mih@somehost.de:dummyds somerepo
mih@somehost.de's password: .00/1.00 [00:00<?, ? Candidate locations/s]
mm_receive_fd: no message header
mux_master_process_new_session: failed to receive fd 0 from slave
mm_receive_fd: no message header
mux_master_process_new_session: failed to receive fd 0 from slave
[ERROR] Failed to clone from any candidate source URL. Encountered errors per each url were:
| - mih@somehost.de:dummyds
  CommandError: '"git" "clone" "--progress" "mih@somehost.de:dummyds" "C:\Users\mih\somerepo"' failed with exitcode 128 [err: 'Cloning into 'C:\Users\mih\somerepo'...
mux_client_request_session: read from master failed: Connection reset by peer
Permission denied, please try again.
Permission denied, please try again.
Received disconnect from 134.94.10.217 port 22:2: Too many authentication failures
Disconnected from 134.94.10.217 port 22
CommandError: '"ssh" "-o" "ControlPath=C:\Users\mih\AppData\Local\datalad.org\datalad\Cache\sockets\24590fdb" "mih@somehost.de" "git-upload-pack 'dummyds'"' failed with exitcode 255
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.'] [install(C:\Users\mih\somerepo)]
install(error): C:\Users\mih\somerepo (dataset) [Failed to clone from any candidate source URL. Encountered errors per each url were:
- mih@somehost.de:dummyds
  CommandError: '"git" "clone" "--progress" "mih@somehost.de:dummyds" "C:\Users\mih\somerepo"' failed with exitcode 128 [err: 'Cloning into 'C:\Users\mih\somerepo'...
mux_client_request_session: read from master failed: Connection reset by peer
Permission denied, please try again.
Permission denied, please try again.
[...]

Issue #2574, opened 2018-05-27
SSH ControlMaster path handling broken on Windows


mih:
This seems to be a core issue underlying many test failures. It forces POSIX path concatenation, instead of os.path.join.

mih:
Still an issue, here the output (git bash) windows 7:

```
 MINGW32 /h/studyforrest-data-aligned (adjusted/master(unlocked))
$ datalad create-sibling this.example.com:dummyrepo
[INFO] Connecting ...
command-line line 0: Missing argument.
[WARNING] Failed to run cmd ['ssh', '-fN', '-o', 'ControlMaster=auto', '-o', 'ControlPersist=15m', '-o', 'ControlPath="C:\\Users\\mihanke\\AppData\\Local\\datalad.org\\datalad\\Cache\\sockets/18f493ff"', 'this.example.com']. Exit code=255
| stdout: None
| stderr: None
[ERROR] Cannot open SSH connection to this.example.com [sshconnector.py:__call__:129] (RuntimeError)
```

mih:
But a switch to switch to `op.join()` doesn't fix the issue. It still fails when executed by datalad. However, when I take the command that is now reported failing, and run it in the git-bash myself, it does what it should.

yarikoptic:
Ran into the same while trying to make `@with_testrepos` work on windows:
![gkrellshoot_09-24-18_145921](https://user-images.githubusercontent.com/39889/45972774-a083d380-c00a-11e8-8354-8e61c8af22c5.png)
for me it stalls the entire process, can't even interrupt in that conda shell... pain

yarikoptic:
ha ha -- but in my case overall it is another issue -- "Detected ssh ri" -- so it just incorrectly determines it to be ssh url which shouldn't be the case here ;-)

bpoldrack:
> but in my case overall it is another issue -- "Detected ssh ri"

It's not another issue, but an additional one. The path considered an SSH RI is not the socket path. That one still fails the same way (not existing because of wrong concatenation).
However, that's interesting. I remember the detection was supposed to first check for a windows path and only then for SSH. We were aware of this potential confusion at some point.

mih:
The release of 0.12 with a fix is imminent (#3993) closing.

PR #3688, opened 2019-09-19
BF: Do not quote path in cmd arg list (fixes gh-3637)

mih:
This quoting should be unnecessary (and counter-productive), because the
command arg list is not evaluated by a shell.

Given that SSH on windows is more broken than that, there is no real
test for this on windows yet, but I manually confirmed that this works.

Also fixes

- [x] #2575
- [x] #2574

### DIFF
--- a/datalad/support/sshconnector.py
+++ b/datalad/support/sshconnector.py
@@ -33,6 +33,7 @@ from datalad.utils import (
     auto_repr,
     Path,
     assure_list,
+    on_windows,
 )
 from datalad.cmd import Runner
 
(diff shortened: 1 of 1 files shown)


bpoldrack:
This probably means, that we have to check `ctrl_path` before to properly escape at least spaces, I think.

mih:
Why? This goes straight into `Popen(shell=False)`.

bpoldrack:
```python
>>> subprocess.Popen(['ssh', '-o' 'ControlPath=/tmp/some space/path', 'ben@localhost'], shell=False)
<subprocess.Popen object at 0x7facb1c5ec10>
>>> command-line line 0: garbage at end of line; "space/path".
```

mih:
Hmm, looking further the story is rather complicated:

> On Windows, if args is a sequence, it will be converted to a string in a manner described in Converting an argument sequence to a string on Windows. This is because the underlying CreateProcess() operates on strings.

https://docs.python.org/3.7/library/subprocess.html#converting-argument-sequence

mih:
Ok, pushed a platform-specific approach.

Issue #3679, opened 2019-09-16
Fully rip out LZMA compat code after PY2 was dropped


mih:
Following https://github.com/datalad/datalad/pull/3629 all of `datalad.support.lzma` could go (only a few lines left anyway).

PR #3691, opened 2019-09-19
RF: Rip out LZMA compat code (fixes gh-3679)

mih:
No longer needed with PY3.5+ being supported only.


mih:
For https://github.com/datalad/git-annex-ria-remote we considered it, and ended up deciding that a straight-up `ssh` client is the more robust choice.

Issue #3672, opened 2019-09-13
sphinx converts -- to a single long –


yarikoptic:
so even if people find it is possible to do they get incorrect instruction on how to:

![image](https://user-images.githubusercontent.com/39889/64881032-b3caed00-d627-11e9-9287-2aba39e5061f.png)

mih:
AFAICS this is only an issue for the RST manpages. I have implemented a fix.

PR #3692, opened 2019-09-19
BF: Ensure double-dashes in RST manpages show up as such (fixes gh-3672)

mih:
This workaround makes this possible without having to wrap the entire
statement in 

```
``...``
```
(which looks weird), and without having to turn
off smart quotes in general (which looks bad elsewhere).


mih:
OK, failures are unrelated, and the fix is trivial....

PR #3694, opened 2019-09-19
ENH: exit with dedicated 99 exit code if installed annex is newer than -devel

yarikoptic:
I found this change lingering in the bf-annex-devel-run branch (e.g. used in #3466 ) intended to test
development version of annex from debian-devel of neurodebian.  It was to solve
the ambiguity of tools/ci/prep-travis-devel-annex.sh exiting with non-0 for other
reasons and thus us just skipping the run (marking it green)


yarikoptic:
correct... will fix that now

Issue #3451, opened 2019-05-23
`datalad -c <SOME>` needs more sanity checking


mih:
I wanted to do

`datalad create -c bids some`

but did

`datalad -c bids create some`

which led to 

```
[INFO   ] Creating a new annex repo at /home/homeGlobal/mih/some 
[ERROR  ] 'NoneType' object has no attribute 'group' [config.py:<listcomp>:438] (AttributeError) 
```

a rather incomprehensible error. I think we should validate the value given to `-c` and fail immediately, if it does not contain `=`.

PR #3695, opened 2019-09-20
ENH: Config override spec sanity checks (fixes gh-3451)

mih:
To prevent us from shooting ourselves in the foot. Without
this change, we could have specified section-less variables
that our own ConfigManager cannot handle, and don't seem to
be supported by Git either.


mih:
Second appveyor run times out -- unlikely to be related to this change.

PR #3701, opened 2019-09-22
ENH: Expose status() file type inspection mode switch

mih:
This can make an essential performance difference in datasets with a large number of files.

Demo:

```
% git annex info | grep '^annexed'
annexed files in working tree: 20054

% time datalad status -t eval
datalad status -t eval  79.15s user 1.51s system 98% cpu 1:21.53 total

% time datalad status -t raw
datalad status -t raw  2.13s user 0.64s system 100% cpu 2.765 total
```

Ping https://github.com/datalad/datalad/issues/3342 @yarikoptic

### DIFF
--- a/datalad/core/local/status.py
+++ b/datalad/core/local/status.py
@@ -105,7 +105,7 @@ STATE_COLOR_MAP = {
 
 
 def _yield_status(ds, paths, annexinfo, untracked, recursion_limit, queried,
-                  eval_submodule_state, cache):
+                  eval_submodule_state, eval_filetype, cache):
     # take the datase that went in first
     repo_path = ds.repo.pathobj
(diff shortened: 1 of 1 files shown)


yarikoptic:
Hard to say for this PR along where it would help since raw isn't used anywhere yet in the code, is it?

mih:
The now exposed internal switch should be in use. However, there were complaints about status itself being slow when called directly. This is for that kind of use case.

This is not to say that a careful inspection would bring up more cases, where we should use this internally as well.

mih:
Ok, no objections. Even if this helps with nothing, but figuring out what impact file type checking would have in a given scenario, it is helpful IMHO -- merging.

Issue #3653, opened 2019-09-06
flaky test test_run_inputs_outputs 


yarikoptic:
ran into while building 0.11.7 package for neurodebian 16.04:

```
======================================================================
ERROR: datalad.interface.tests.test_run.test_run_inputs_outputs
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/build/datalad-0.11.7/.pybuild/pythonX.Y_2.7/build/datalad/tests/utils.py", line 442, in newfunc
    return t(*(arg + (d,)), **kw)
  File "/build/datalad-0.11.7/.pybuild/pythonX.Y_2.7/build/datalad/tests/utils.py", line 615, in newfunc
    return t(*(arg + (filename,)), **kw)
  File "/build/datalad-0.11.7/.pybuild/pythonX.Y_2.7/build/datalad/interface/tests/test_run.py", line 770, in test_run_inputs_outputs
    ds.run("touch dummy{}".format(idx), inputs=inputs_arg)
  File "/usr/lib/python2.7/dist-packages/wrapt/wrappers.py", line 557, in __call__
    args, kwargs)
  File "/build/datalad-0.11.7/.pybuild/pythonX.Y_2.7/build/datalad/distribution/dataset.py", line 527, in apply_func
    return f(**kwargs)
  File "/usr/lib/python2.7/dist-packages/wrapt/wrappers.py", line 518, in __call__
    args, kwargs)
  File "/build/datalad-0.11.7/.pybuild/pythonX.Y_2.7/build/datalad/interface/utils.py", line 499, in eval_func
    return return_func(generator_func)(*args, **kwargs)
  File "/usr/lib/python2.7/dist-packages/wrapt/wrappers.py", line 518, in __call__
    args, kwargs)
  File "/build/datalad-0.11.7/.pybuild/pythonX.Y_2.7/build/datalad/interface/utils.py", line 487, in return_func
    results = list(results)
  File "/build/datalad-0.11.7/.pybuild/pythonX.Y_2.7/build/datalad/interface/utils.py", line 474, in generator_func
    msg="Command did not complete successfully")
[...]

PR #3705, opened 2019-09-24
NF: GitRepo.for_each_ref_()

### LINKED ISSUE #3703: Reimplement branch querying helpers around `git for-each-ref`
Relevant for #2970 

Concrete targets are:

- `get_active_branch()`
- `get_branches()`
- `get_remote_branches()`

mih:
Work towards gh-3703

- [x] `for_each_ref()` helper
- [x] actually fix #3703
- [x] figure out why `repodates` test fails
       km: 477e23a85 fixes the test_repodates.py:test_check_dates failure.

It seems that the present code doesn't conveniently handle things like annotated tags that point to an actual commit we are interested in, but actually have a separate `objectname` (or SHA1).

### DIFF
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -1270,13 +1270,27 @@ class GitRepo(RepoInterface, metaclass=Flyweight):
                       pattern=None, points_at=None, sort=None, count=None):
         """Wrapper for `git for-each-ref`
 
+        Please see manual page git-for-each-ref(1) for a complete overview
+        of its functionality. Only a subset of it is supported by this
+        wrapper.
+
         Parameters
         ----------
-        fields : list or str
-        pattern : list or str
-        points_at : str
-        sort : list or str
-        count : int
+        fields : iterable or str
+          Used to compose a NULL-delimited specification for for-each-ref's
+          --format option. The default field list reflects the standard
+          behavior of for-each-ref when not --format option is given
+        pattern : list or str, optional
+          If provided, report only refs that match at least one of the given
+          patterns.
+        points_at : str, optional
+          Only list refs which points at the given object.
+        sort : list or str, optional
+          Field name(s) to sort-by. If multiple fields are given, the last one
+          becomes the primary key. Prefix any field name with '-' to sort in
+          descending order.
+        count : int, optional
+          Stop iteration after the given number of matches.
 
         Yields
         ------
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -1729,8 +1729,8 @@ class GitRepo(RepoInterface, metaclass=Flyweight):
         """
 
         return [
-            b['refname'][11:]  # strip 'refs/heads/'
-            for b in self.for_each_ref_(fields='refname', pattern='refs/heads')
+            b['refname:strip=2']
+            for b in self.for_each_ref_(fields='refname:strip=2', pattern='refs/heads')
         ]
 
     def get_remote_branches(self):
@@ -1747,8 +1747,8 @@ class GitRepo(RepoInterface, metaclass=Flyweight):
         # currently this is done in collection
 
         return [
-            b['refname'][13:]  # strip 'refs/remotes/'
-            for b in self.for_each_ref_(fields='refname', pattern='refs/remotes')
+            b['refname:strip=2']
+            for b in self.for_each_ref_(fields='refname:strip=2', pattern='refs/remotes')
         ]
 
     def get_remotes(self, with_urls_only=False):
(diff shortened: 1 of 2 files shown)


datalad/support/gitrepo.py
```diff
@@ -1266,6 +1266,60 @@ def _get_prefixed_commit_msg(msg):
         DATALAD_PREFIX = "[DATALAD]"
         return DATALAD_PREFIX if not msg else "%s %s" % (DATALAD_PREFIX, msg)
 
+    def for_each_ref_(self, fields=('objectname', 'objecttype', 'refname'),
+                      pattern=None, points_at=None, sort=None, count=None):
+        """Wrapper for `git for-each-ref`
+
+        Parameters
+        ----------
+        fields : list or str
```
kyleam: For most these parameters, I'm okay with not having a description because it is just a direct map to the `git for-each-ref` option.  But I think the description for `fields` should mention that it is used to construct the `--format` value.

Also, I suspect you're using a tuple as the default to avoid the bad practice/gotcha of using a mutable list as the default value (quickly scanning, I don't think a list would actually be problematic here, though it should still be avoided).  I'm fine with that (and would be fine with the more standard "None default, set list" as well), but I think if the default value is a tuple, the type shouldn't be documented as "list or str".  Perhaps "str or iterable of str"?

datalad/support/gitrepo.py
```diff
@@ -1678,21 +1746,10 @@ def get_remote_branches(self):
...
-                    # this will happen with git annex special remotes
-                    pass
-                else:
-                    raise e
-        return remote_branches
-        # return [branch.strip() for branch in
-        #         self.repo.git.branch(r=True).splitlines()]
+        return [
+            b['refname'][13:]  # strip 'refs/remotes/'
+            for b in self.for_each_ref_(fields='refname', pattern='refs/remotes')
```
kyleam: I was going to remark that, if a "<remote>/HEAD" symbolic ref exists, this will return it, which we probably don't want.  But it turns out that the gitpy variant does the same thing.  OK.

datalad/support/gitrepo.py
```diff
@@ -2661,16 +2718,15 @@ def get_tags(self, output=None):
...
-                'name': t.name,
-                'hexsha': t.commit.hexsha
-             }
-            for t in tag_objs
+            dict(
+                name=t['refname:lstrip=2'],
+                hexsha=t['object'] if t['object'] else t['objectname'],
+            )
+            for t in self.for_each_ref_(
+                fields=['refname:lstrip=2', 'objectname', 'object'],
```
kyleam: Ah, neat, I didn't realize fields for `object` (and other header values) existed.  So this is equivalent to dereferencing `objectname`: `['refname:lstrip=2', 'objectname', '*objectname']`.

For sorting, `taggerdate` isn't ideal because it only applies to annotated tags.  `creatordate` is better here.

I was a bit surprised to see you use ":lstrip" here rather than splicing (so `[10:]`) like you do elsewhere.  I think either is fine, though if we're going to use one consistently, I have a weak preference for using ":lstrip", or actually ":strip".  ":strip" has been around since Git v2.7.1, while the synonym ":lstrip" has only been around since v2.13.0.  While I personally don't think we should worry about supporting v2.13.0, we might as well use the older spelling.  (We really should specify a minimum Git version so that we can make clear decisions in these cases.)

My other related thought was whether you should use `--short` for symbolic-ref and `:short` for for-each-ref to let git handle the truncation, keeping what is needed for the ref to be unambiguous (e.g., refs/heads/foo -> heads/foo and refs/tags/foo -> tags/foo).  But on second thought, there are at least two reasons not to qualify ambiguous refs: (1) the gitpy versions you are replacing do not and (2) get_active_branch(), get_{remote_,}branches}(), and get_tags() are by definition limited to a particular namespace, so we do not need to qualify that the ref is ambiguous in some other namespace.

datalad/support/gitrepo.py
mih: I will add a description to all of them.

I used a tuple for those reasons, I used it over None to make it obvious from the signature what dict keys one can expect. I will fix the description.

datalad/support/gitrepo.py
mih: Ha, I didn't know that you can dereference objectname! ;-)

I now uniformly do `:strip=2`. Thx for illustrating the problem space.

I agree re `--short`, and left it as-is.

mih:
@kyleam I think I have addressed all points you have raised. Thx for the review!

kyleam:
test_get_tags is failing in the run that overrides the git author/date variables: https://travis-ci.org/datalad/datalad/jobs/589416798#L1138

I'll look into it and fix it.

We're seeing red frequently due to gh-3653, which makes it hard to spot the failures triggered by the PR.  I'm still not able to trigger 3653 locally and don't understand why the failure has emerged, but I'll open a PR later today proposing a workaround for it.

PR #3712, opened 2019-09-25
RF: Blind attempt to relay status() performance boost possibility on to save()

mih:
According to my current logic and assessment, the `status()` run done in
top-level `save()` does not need to inspect filetypes. The tests will show
if that is true.

If true, performance gains of the same magnitude as shown in
https://github.com/datalad/datalad/pull/3701 for `status()` also become
available for `save()`.

### DIFF
--- a/datalad/core/local/save.py
+++ b/datalad/core/local/save.py
@@ -192,6 +192,7 @@ class Save(Interface):
                 dataset=dataset,
                 path=path,
                 untracked=untracked_mode,
+                report_filetype=False,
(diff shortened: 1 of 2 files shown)


mih:
Hmm, strange timeouts on travis. I cannot reproduce that locally and travis no longer tells which particular test it is happening in.

mih:
OK, whatever it would break, our tests don't see it. Neither does my imagination. So let's do it and wait for the eventual bug report.

PR #3718, opened 2019-09-26
Enable sshrun tests on appveyor

mih:


mih:
This lead to strange timeouts... no idea.

mih:
Did not make progress on this one, and could not come back. No need to keep open.

Issue #3633, opened 2019-08-31
datalad.core test failures on real win10/conda


mih:
The appveyor environment is supercharged in many ways. Here is a glimpse of reality starting from https://github.com/datalad-handbook/book/issues/132:

```
> python -m nose datalad.core
======================================================================
ERROR: datalad.core.local.tests.test_create.test_create_with_obscure_name({'path': ' ;abcΔЙקم๗ あ'},)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Users\mih\Miniconda3\lib\site-packages\nose\case.py", line 198, in runTest
    self.test(*self.arg)
  File "c:\users\mih\code\datalad\datalad\tests\utils.py", line 607, in newfunc
    return t(*(arg + (filename,)), **kw)
  File "c:\users\mih\code\datalad\datalad\core\local\tests\test_create.py", line 470, in check_create_obscure
    ds = create(result_renderer="default", **create_kwargs)
  File "C:\Users\mih\Miniconda3\lib\site-packages\wrapt\wrappers.py", line 564, in __call__
    args, kwargs)
  File "c:\users\mih\code\datalad\datalad\interface\utils.py", line 499, in eval_func
    return return_func(generator_func)(*args, **kwargs)
  File "C:\Users\mih\Miniconda3\lib\site-packages\wrapt\wrappers.py", line 564, in __call__
    args, kwargs)
  File "c:\users\mih\code\datalad\datalad\interface\utils.py", line 487, in return_func
    results = list(results)
  File "c:\users\mih\code\datalad\datalad\interface\utils.py", line 435, in generator_func
    result_renderer, result_xfm, _result_filter, **_kwargs):
  File "c:\users\mih\code\datalad\datalad\interface\utils.py", line 595, in _process_results
    default_result_renderer(res)
  File "c:\users\mih\code\datalad\datalad\interface\utils.py", line 526, in default_result_renderer
    if 'message' in res else ''))
  File "c:\users\mih\code\datalad\datalad\ui\dialog.py", line 75, in message
    self.out.write(msg)
  File "C:\Users\mih\Miniconda3\lib\encodings\cp1252.py", line 19, in encode
    return codecs.charmap_encode(input,self.errors,encoding_table)[0]
UnicodeEncodeError: 'charmap' codec can't encode characters in position 71-76: character maps to <undefined>

======================================================================
ERROR: datalad.core.local.tests.test_diff.test_path_diff
----------------------------------------------------------------------
[...]

PR #3721, opened 2019-09-26
Tst win

mih:
With these changes #3633 is fixed.


mih:
Travis failure is related to https://github.com/datalad/datalad/pull/3716

I suspect that disabling the unicode infusion in the tests can e considered controversial, but it would help me quite a bit in structuring the salvaging operation on windows. And given that I am presently the only one working on it, I'll take the liberty to make this decision -- which is to be reverted, as soon as we can see the silver lining ...

PR #3727, opened 2019-09-27
NF: a prototype for ext special remote to store files in git/objects of a remote

yarikoptic:
I have been blurbing about a remote which would (ab)use regular `.git/objects` and refs of a remote to store annexed context.  Nobody got excited, but I kept thinking about it because  often we annex context which is not necessarily "too large" to push it to github.  It might be that I need to share files which
- need to be encrypted if shared publicly;
- I know that later on might not need on every clone (thus shouldn't just go to git)
- are sizeable but not huge, but I know that committing them to git is a crime, but they would fit nicely to github
- I might need to make unavailable later on without rewriting entire history
- I might need to keep only the most recent version available publicly/from that repository

So, welcome the prototype (in bash) for now of `git-annex-remote-gitobjects` , so if you place it in your PATH (and make executable), you should be able to

    datalad install -gs https://github.com/yarikoptic/testds.git

and get 2 annexed files from that GitHub repo.

What I need to figure out is  how to RF CHECKPRESENT  to avoid fetching the object pointed by the ref.  If there is no way, I have ideas for a workaround... but hopefully git gurus (@kyleam and @joeyh) would tell me HOWTO

And then RF to store not a remote name but remoteurl as a parameter of the remote (or both?)

[ci skip]


mih:
Sweet!

kyleam:
> What I need to figure out is how to RF CHECKPRESENT to avoid fetching the object pointed by the ref.

How about `git ls-remote --exit-code <remote> <ref>`?

kyleam:
> And then RF to store not a remote name but remoteurl as a parameter of the remote (or both?)

Using just the url seems like the way to go to me because you don't have control over what cloners name their remote, and git will accept the full url as the remote.

sandbox/git-annex-remote-gitobjects
```diff
@@ -0,0 +1,168 @@
...
+	fi
+}
+
+# This has to come first, to get the protocol started.
+echo VERSION 1
+
+# quick and dirty way to figure out root, dunno any better ATM but set
+# an alias
+git config --local --add alias.root '!pwd'
+dotgit=$(git root)/.git
```
kyleam: `git rev-parse --git-dir` or `git rev-parse --absolute-git-dir`?  (applies to another spot below too)

sandbox/git-annex-remote-gitobjects
```diff
@@ -0,0 +1,168 @@
...
+		;;
+		*)
+			RET=""
+		;;
+	esac
+}
+
+cleanup () {
+	# remove reference and object locally
+	rm -f "$dotgit/$ref"
```
kyleam: `git update-ref -d "$ref"` or `git update-ref -d "$ref" "$h"`?

yarikoptic:
> Using just the url seems like the way to go to me because you don't have control over what cloners name their remote, and git will accept the full url as the remote.

yep yep... it was initially just a quick way to work it out to see if it would work at all ;)

sandbox/git-annex-remote-gitobjects
yarikoptic: sweet -- thank you!!

yarikoptic:
> How about `git ls-remote --exit-code <remote> <ref>`?

Thanks!!! Seems to work wonderfully. committing

yarikoptic:
<details>
<summary>
note: something "interesting" is happening if I enable `encryption=shared`. Then we could get different git blob somehow
</summary> 

```shell
$> git -c gc.auto=0 push github refs/gitobjects/GPGHMACSHA1--a91a9ecb6216a31574c9512646bfc8164bc27881:refs/gitobjects/GPGHMACSHA1--a91a9ecb6216a31574c9512646bfc8164bc27881
To github.com:yarikoptic/testds.git
 ! [rejected]        refs/gitobjects/GPGHMACSHA1--a91a9ecb6216a31574c9512646bfc8164bc27881 -> refs/gitobjects/GPGHMACSHA1--a91a9ecb6216a31574c9512646bfc8164bc27881 (needs force)
error: failed to push some refs to 'git@github.com:yarikoptic/testds.git'
hint: You cannot update a remote ref that points at a non-commit object,
hint: or update a remote ref to make it point at a non-commit object,
hint: without using the '--force' option.
1 31082 ->1.....................................:Fri 27 Sep 2019 11:05:48 AM EDT:.
(git-annex)hopa:/tmp/testds[master]git-annex
$> git ls-remote --exit-code "githubobjects" "refs/gitobjects/GPGHMACSHA1--a91a9ecb6216a31574c9512646bfc8164bc27881"
fatal: 'githubobjects' does not appear to be a git repository                                                       

and the repository exists.
1 31083 ->128.....................................:Fri 27 Sep 2019 11:07:37 AM EDT:.
(git-annex)hopa:/tmp/testds[master]git-annex
$> cat .git/config 
[core]
	repositoryformatversion = 0
	filemode = true
	bare = false
	logallrefupdates = true
[annex]
	uuid = dfb93d03-27b7-4e22-8c4f-bf76ade05801
	version = 7
	backends = MD5E
[filter "annex"]
	smudge = git-annex smudge -- %f
	clean = git-annex smudge --clean -- %f
[remote "github"]
	annex-ignore = true
	url = https://github.com/yarikoptic/testds.git
	fetch = +refs/heads/*:refs/remotes/github/*
[remote "githubobjects"]
	annex-externaltype = gitobjects
	annex-uuid = 2b31fadd-572f-4fe3-a3eb-3baf00159e84
	annex-cost = 200.0
	annex-availability = GloballyAvailable
1 31084.....................................:Fri 27 Sep 2019 11:07:45 AM EDT:.
(git-annex)hopa:/tmp/testds[master]git-annex
$> git ls-remote --exit-code "github" "refs/gitobjects/GPGHMACSHA1--a91a9ecb6216a31574c9512646bfc8164bc27881"       
38cabb4d34a0b0d57eb25ca80d58516a519921ab	refs/gitobjects/GPGHMACSHA1--a91a9ecb6216a31574c9512646bfc8164bc27881
1 31085.....................................:Fri 27 Sep 2019 11:07:57 AM EDT:.
(git-annex)hopa:/tmp/testds[master]git-annex
$> git -c gc.auto=0 show-ref -s "refs/gitobjects/GPGHMACSHA1--a91a9ecb6216a31574c9512646bfc8164bc27881" 
a4a15725fdd7ef68cb699f24b52820b6da08090b
```
</details>

will redo without encryption

sandbox/git-annex-remote-gitobjects
kyleam: But perhaps you can get away with not creating a local ref at all:  `git push <url> "$h":"$ref"` and `git fetch <url> "$h"`.

joeyh:
This seems very similar to the bup special remote, in particular you
have one git reference per annexed object. (Which can be a scalability
problem.) 

I suppose one benefit of this over bup is that to get bup to store
content on a remote server seems needs shell access to the server
to run bup there.

(The git-lfs special remote also covers a somewhat similar use case.)

Re encryption, two encrypted objects with the same name can have
different content since gpg output has timestamps and maybe other things
that make it not reproducible. This is not normally a problem because if
git-annex sees the remote has the object it won't re-encrypt the content
again. Of course if CHECKPRESENT returns bad information, git-annex
could re-encrypt and re-send an object.

I suppose that could also happen in a situation where there are two
independent writers.. I have not thought about that before.

-- 
see shy jo

yarikoptic:
<details>
<summary>Note: .git/objects seems still to grow. e.g. upon drop grew up, but they do get cleaned up and might be git-annex or our datalad use of gc.auto=0</summary> 

```shell
datalad drop *  7.60s user 2.07s system 18% cpu 52.169 total
(dev3) 1 31182.....................................:Fri 27 Sep 2019 11:53:45 AM EDT:.
(git-annex)hopa:/tmp/testds-clone[master]git-annex
$> du -scmb .git/*
111898	.git/annex
4096	.git/branches
606	.git/config
73	.git/description
167	.git/FETCH_HEAD
23	.git/HEAD
23297	.git/hooks
11110	.git/index
4689	.git/info
28252	.git/logs
961189	.git/objects
41	.git/ORIG_HEAD
306	.git/packed-refs
33005	.git/refs
1178752	total
(dev3) 1 31185.....................................:Fri 27 Sep 2019 11:57:54 AM EDT:.
(git-annex)hopa:/tmp/testds-clone[master]git-annex
$> git gc --prune=now
Enumerating objects: 1614, done.
Counting objects: 100% (1614/1614), done.
Delta compression using up to 4 threads
Compressing objects: 100% (875/875), done.
Writing objects: 100% (1614/1614), done.
Total 1614 (delta 387), reused 1133 (delta 268)
(dev3) 1 31186.....................................:Fri 27 Sep 2019 11:58:04 AM EDT:.
(git-annex)hopa:/tmp/testds-clone[master]git-annex
$> du -scmb .git/*   
111898	.git/annex
4096	.git/branches
606	.git/config
73	.git/description
167	.git/FETCH_HEAD
23	.git/HEAD
23297	.git/hooks
11110	.git/index
4755	.git/info
28255	.git/logs
173351	.git/objects
41	.git/ORIG_HEAD
372	.git/packed-refs
32800	.git/refs
390844	total

```
</details>

sandbox/git-annex-remote-gitobjects
yarikoptic: oh... I could do that? ;)  I will give it a shot later on! thanks

yarikoptic:
FYI: http://github.com/yarikoptic/testds now contains a new version of the test dataset with many more files of a variety (all are not huge).  The problem with it is that I called it 'github' remote, and haven't tuned up for storing url not remote name yet.  So after install/clone you need `git remote rename origin github` to get it working.
I pushed all my changes and will not touch it for a bit (other things todo) happen someone decide to give it tune ups.

sandbox/git-annex-remote-gitobjects
```diff
@@ -37,7 +37,7 @@ setconfig () {
 
 # Sets ref to the reference to use to store a key.
 calcref () {
-	ref="refs/annexed/$1"
+	ref="refs/gitobjects/$1"
```
kyleam: I see the value of using the name of the special remote, but I think it'd still for good for there to be "annex" in the name.  Length shouldn't matter here, so what about "refs/annex-gitobjects" or "refs/gitobjects-annexed"?

yarikoptic: Yes, I thought about that too but went for shorter for no good reason. Probably refs/annex-gitobjects would be best

kyleam:
> Note: .git/objects seems still to grow. e.g. upon drop grew up, but they do get cleaned up and might be git-annex or our datalad use of gc.auto=0

By default, the loose objects you're creating locally with `git hash-object ...`  will stick around until `gc` is executed and the object is older than two weeks (the `gc.pruneExpire` default).

yarikoptic:
> > Note: .git/objects seems still to grow. e.g. upon drop grew up, but they do get cleaned up and might be git-annex or our datalad use of gc.auto=0
> 
> By default, the loose objects you're creating locally with `git hash-object ...` will stick around until `gc` is executed and the object is older than two weeks (the `gc.pruneExpire` default).

I thought I was removing generated object in cleanup? Or there are more generated?

kyleam:
> I thought I was removing generated object in cleanup? Or there are more generated?

Sorry for the noise.  Overlooked that part in my skimming.

kyleam:
FWIW with the test script below I don't see any unexpected objects being left behind.  The two dangling blobs are unrelated and all the new objects are expected objects in the git-annex branch.

<details>
<summary>script</summary>

```
#!/bin/sh
set -eu

cd $(mktemp -d --tmpdir dl-XXXXXXX)
datalad create a >/dev/null 2>&1
datalad install -s a b >/dev/null 2>&1
cd b
git annex initremote go \
    remote=origin type=external externaltype=gitobjects encryption=none >/dev/null 2>&1

printf one >one
git annex add one >/dev/null 2>&1
git commit -mone -q

find .git/objects -type f | sort >gobjects-pre

git annex move --to go one >/dev/null 2>&1

git annex whereis one
echo

find .git/objects -type f | sort >gobjects-after

content_id=$(printf one | git hash-object --stdin)

git rev-parse --quiet --verify $content_id"^{blob}" || \
    printf "content blob %s not present\n\n" $content_id

git fsck
echo

diff -u gobjects-pre gobjects-after
```

```
whereis one (1 copy) 
  	2652c5eb-c144-4774-a888-7dbf239a52e9 -- [go]
ok

content blob 43dd47ea691c90a5fa7827892c70241913351963 not present

dangling blob 9c7d504362403baf2ce42a2d7769462d0c516d5b
dangling blob e69de29bb2d1d6434b8b29ae775ad8c2e48c5391

--- gobjects-pre	2019-09-27 16:19:51.147726143 -0400
+++ gobjects-after	2019-09-27 16:19:51.403728803 -0400
@@ -1,4 +1,7 @@
 .git/objects/06/77f4e94a60898d1a1b19f0a233bdf11c8cb8bc
+.git/objects/1d/9208921c9f2e074d5abf455137782e7e490038
+.git/objects/21/66e3214dc696cfb53d4251449858ac0cab7449
+.git/objects/23/cd03ce6ae758a724c2966edc486b582454c679
 .git/objects/24/b69eebf02312e6748678e4120fd582808437c9
 .git/objects/25/7d5436e00f063b8499bfa876f7f8e1271190e3
 .git/objects/38/602b99198a81d7149e0891bd06669c44e70d5e
@@ -21,6 +24,7 @@
 .git/objects/9e/1db3e5e57730903bf03d67d7c7bc47ec0753f1
 .git/objects/9e/57fe02e69c7675f52e68624b7cae1405897996
 .git/objects/a5/de79690a41625b9411f02a49f205577e4044bb
+.git/objects/b2/56b4372a61d8358118c592b0f53f0fde0537d9
 .git/objects/b5/40820107ca5718dc0f196a08edf3729239bfef
 .git/objects/c3/aaefef9a2470b31ba9213350046ff7cde75737
 .git/objects/c7/2c456d9e983ce1efaaca2d86a5630f40cabdfa
@@ -28,3 +32,4 @@
 .git/objects/df/b87fb3e346c1fbc45d06da9aa8f9b13efd9d40
 .git/objects/e3/e5d71d9a5cddd7abc19e17d5b9a339820cebbb
 .git/objects/e6/9de29bb2d1d6434b8b29ae775ad8c2e48c5391
+.git/objects/fe/7cd080b3a8417e404eda127f343a7099114ad5
```

</details>

yarikoptic:
> > Using just the url seems like the way to go to me because you don't have control over what cloners name their remote, and git will accept the full url as the remote.
> 
> yep yep... it was initially just a quick way to work it out to see if it would work at all ;)

I think it should follow closely the `type=git` special remote on that and have `location` (and `name` since it is mandatory)

yarikoptic:
> FWIW with the test script below I don't see any unexpected objects being left behind. The two dangling blobs are unrelated and all the new objects are expected objects in the git-annex branch.

thank you ! for looking into it.  Then I think it was a side-effect of datalad using gc.auto=0 for all (?) `git` invocations.  FWIW -- it was primarily done to not confuse GitPython IIRC (just saying).

kyleam:
> Then I think it was a side-effect of datalad using gc.auto=0 for all (?) `git` invocations. FWIW -- it was primarily done to not confuse GitPython IIRC (just saying).

I'm not following your logic, but either way datalad no longer sets that as of 203aa652a (RF: remove  receive.autogc=0 and gc.auto=0 options to git calls, 2019-05-24).

yarikoptic:
Hm... Then it must have been git annex!  nice to have so many possible blame targets ;-)

yarikoptic:
Thank you @joeyh for the feedback!

> This seems very similar to the bup special remote, in particular you have one git reference per annexed object. (Which can be a scalability problem.)

well - if I get it right it would be scalability concern on remote (e.g. github side). But overall the main purposes for this remote are not really intended neither for storing lots of files, nor for storing lots of heavy content.  So I wouldn't worry too much about scalability here

> Re encryption,  ... Of course if CHECKPRESENT returns bad information, git-annex
could re-encrypt and re-send an object.

indeed may be I should check that `CHECKPRESENT` is functioning reliable enough.

One more question:  I do not see any `CLEAN` directive.  I see possibility of the remote

joeyh:
Yaroslav Halchenko wrote:
> Thank you @joeyh for the feedback!
> 
>     This seems very similar to the bup special remote, in particular you have
>     one git reference per annexed object. (Which can be a scalability problem.)
> 
> well - if I get it right it would be scalability concern on remote (e.g. github
> side).

Git's scalability problems with large numbers of references are mostly
at the protocol level AIUI, unless they've improved the protocol to not
list them all on every fetch since the last time I looked.

> One more question: I do not see any CLEAN directive. I see possibility of the
> remote

Seems this question got truncated..

-- 
see shy jo

yarikoptic:
Indeed would be nice to check for fetch. 

Question - sorry about that. Ended up posting it to https://git-annex.branchable.com/design/external_special_remote_protocol/#comment-67c52cfaba1dd8b7428806ee48e7a095 and forgot to remove preamble here

yarikoptic:
One of the recent use-cases where I have thought that it would have been really convenient if such special remote was built-in in annex was: me trying git-annex assistant again on my phone for sync'ing TODOs in org-mode format (with encryption, I could have emailed myself a key; and just published that repo to github).
@joeyh - any chance you might implement something like this special remote natively within git-annex?
Another use case was https://github.com/dandi-datasets/nwb_test_data -- ATM it is quite small but it felt wrong committing any data files directly into git.  I guess I will pay the price for doing that later.

yarikoptic:
> Or you could have used git-annex's git-lfs support to store the annex on github.

oh, I have not realized that github is providing any git-lfs storage for free. Apparently it is: "Every account using Git Large File Storage receives 1 GB of free storage and 1 GB a month of free bandwidth.".  For notes syncing that should be enough!  (I will pester on that topic separately on git-annex site, since unrelated here per se)

> Storing blobs in a git repo feels perhaps a bit more foot-shooty than would be wise to expose in the assistant gui.

well, if it would be marked as untrusted repository, I do not see any harm which would come unexpected.  But it would open interoperability with git hosting providers which have no git-lfs support.

But that is ok.  At least for now, if I get git-lfs working smoothlish (and closing the eyes on inability to delete any data from the special remote), it could provide a solution

yarikoptic:
would anyone mind if I merge this one so we have a copy in the `sandbox/`?

kyleam:
> would anyone mind if I merge this one so we have a copy in the `sandbox/`?

I'm not sure I see the benefit of this being in datalad's main working tree, but if that's your preference I don't mind.  We don't currently have a sandbox directory, so it'd be nice to document somewhere (perhaps just a README within that directory) what the purpose of that directory is in the context of datalad.

yarikoptic:
@kyleam - thought either to just close it or "merge"... decided that it might be worth keeping around. Added description for `sandbox/` to CONTRIBUTING.md

sandbox/git-annex-remote-gitobjects
yarikoptic: FWIW -- it was done
```
gitdir=$(git rev-parse --git-dir)
```

yarikoptic:
hm, `annex testremote` gets stuck on `removeKey when not present... fixed up -- will push and retag whenever `annex testremote` passes

yarikoptic:
pushed, canceling CI runs.  The commit message of 11a265c4182b08172e77889d508704a5319a5362 has testremote output and also my "summary":

    testremote took almost 2h to complete!!! So this thing is quite ineffecient,
    but interestingly eyeballing top from time to time showed that nothing really
    is busy -- neither git or git-annex.  And to that attests zsh reported (at the
    bottom) times:
    
        git annex testremote go  51.37s user 13.49s system 1% cpu 1:45:27.44 total
    
    so there was only a minute of user + system time consumed, while total run was
    1h 45 min.  So, it seems that git-annex, git and interaction between the two
    might be the explanation for datalad often not being that busy -- something else
    (absent flushes?) hinders efficient interaction with (or within) git or git-annex.

yarikoptic:
documentation was extended to describe `sandbox`, no other objections were expressed, merging

PR #3729, opened 2019-09-27
Github workflow to run benchmarks against master

mih:
I canceled all other CIs

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,37 @@
+name: Benchmarks
+
+on: [push]
+
+jobs:
+  vs-master:
+
+    runs-on: ubuntu-latest
+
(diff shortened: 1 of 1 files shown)


yarikoptic:
Hm, when I click on green vs-master on the phone it just reloads this page, so can't see actual logs etc

yarikoptic:
In desktop mode guess to Checks tab, which is not present in Mobile view.

mih:
Good to go then?

yarikoptic:
Sure!

yarikoptic:
But ideally we should add "Diff benchmarks" with just `asv compare` call after "Run benchmarks" so when expanded it just shows the diff.  Now you need to scroll for a while to skip all the setup/running

PR #3730, opened 2019-09-27
Github CI based Windows tests

mih:
In comparions to appveyor setup:

- more modern OS (win2019 server)
- PY37 not PY35 (ie. appveyor is no longer a blocker for making PY36 minimum dep)
- runs more tests
- no SSH setup (yet), but SSH pretty much not functional anyways
- no unicode related tests are running, because the whole encoding issue is shadowing other bugs (encoding can later be tackled in with a proper PY3 approach)
- each failing tests is individually marked as such (need to go through them one by one and want to see it in the code)

The checks won't run until one of these PR has made it into master. But here is the outcome for this branch shown in my fork: https://github.com/mih/datalad/pull/32/checks

### DIFF
--- a/.github/workflows/test_win2019.yml
+++ b/.github/workflows/test_win2019.yml
@@ -1,0 +1,64 @@
+name: Win2019 tests
+
+on: [push]
+
+jobs:
+  test:
+
+    runs-on: windows-2019
+
+    strategy:
+      fail-fast: false
+      matrix:
+        module: [
+            datalad.core,
+            datalad.customremotes,
+            datalad.distributed,
+            datalad.distribution,
+            datalad.downloaders,
+            datalad.interface,
+            datalad.local,
+            datalad.metadata,
+            datalad.plugin,
+            datalad.support,
+            datalad.tests,
+            datalad.ui datalad.cmdline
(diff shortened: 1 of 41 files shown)


.github/workflows/test_win2019.yml
```diff
@@ -0,0 +1,59 @@
...
+      run: |
+        datalad wtf
+        dir
+    - name: ${{ matrix.module }} tests
+      run: |
+        mkdir -p __testhome__
+        cd __testhome__
+        python -m nose -s -v --with-cov --cover-package datalad ${{ matrix.module }}
+    # coverage report is not functional because codecov refuses to accept the
+    # report
```
yarikoptic: eh heh -- that is sad.
From all the annotation of tests I was not sure if any of those was passing on appveyor but fails on github ci?

mih: I would be surprised if that would be the case, but I cannot claim that with much confidence. In any case, this is not meant to replace appveyor in the short term. My experience is that windows systems are highly flexible.Many things that work on appveyor fail on my win10 test box. This just gives us a better grasp on real-world variability.

PR #3734, opened 2019-09-30
Workflow to build docs in a dedicated CI run

mih:
Canceled all CI runs except for the new one to save on carbon footprint.

### DIFF
--- a/.github/workflows/docbuild.yml
+++ b/.github/workflows/docbuild.yml
@@ -1,0 +1,27 @@
+name: Docs
+
+on: [push]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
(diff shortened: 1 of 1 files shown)


mih:
I think this is uncontroversial, so I will merge it.

Follow-up question is: do we want to strip the doc builds from travis for a bit of speed-up? I do not care whether the docs build under all possible conditions. But I do care about the total runtime of the tests.

kyleam:
> do we want to strip the doc builds from travis for a bit of speed-up?

Sounds good to me.

yarikoptic:
yep, sounds good to me too!  There are benefits from building docs under various possible scenarios IMHO but the build/test run time out-weights it! thanks!

Issue #3714, opened 2019-09-25
Investigate fate of AnnexRepo.add()


mih:
After a future removal of the obsolete old `save()` and `add()` `AnnexRepo.commit()` is completely bypassed by high-level API code. However, there are a bunch of tests specifically for this function. It needs to be investigated inhowfar those tests test something that is relevant for the `save()` code paths, and/or what parts of `save()` can or need to move into `AnnexRepo.commit()` to remain valid and useful.

Same issue for `AnnexRepo.add()`. It is not used by `save()`. Instead there is `AnnexRepo._save_add()`.

mih:
General feeling during initial discussion in conference call was that the Repo classes are internal API. Hence those methods could be removed, when no longer needed internally.

The situation is complicated by the fact that old code (e.g. the crawler) still uses the Repo API as the primary API. So in practice we will have to carry this burden for a while -- knowing that bitrot will accumulate in them.

yarikoptic:
> ... Repo classes are internal API.

As long as they are exposed as `Dataset.repo`, they are not.  
We have made a similar mistake to expose GitPython's `.repo` as `*Repo.repo` and now e.g. our prominent openneuro downstream uses them: https://github.com/OpenNeuroOrg/datalad-service/search?q=.repo&unscoped_q=.repo&type=Code (at times even `.repo.repo`). So pruning should be done with care or at least keeping openneuro folks aware (as was done in the past).

> The situation is complicated by the fact that old code (e.g. the crawler) still uses the Repo API as the primary API.

yeah, DataLad's grand-father crawler needs to be RFed to start using higher level API. Possibly we could avoid full "compatibility" RFing, but rather only RF main used crawling pipelines to use new interface/implementation.

kyleam:
@yarikoptic:
>  our prominent openneuro downstream uses them: https://github.com/OpenNeuroOrg/datalad-service/search?q=.repo&unscoped_q=.repo&type=Code (at times even `.repo.repo`). So pruning should be done with care or at least keeping openneuro folks aware (as was done in the past).

No need to look too far into the past.  The same discussion that spurred the comment you're replying to led the following discussion on their side:

* https://github.com/OpenNeuroOrg/datalad-service/pull/114#issuecomment-540262800

  A comment where you say, "We are thinking to overhaul those classes which we primarily see as the internal interface."

* https://github.com/OpenNeuroOrg/datalad-service/issues/116

mih:
[...]

Issue #3368, opened 2019-05-01
Retire `annotate_paths()`


mih:
RF the following remaining users:

- [x]  `add` -- to be retired via `rev_save`
- [x]  `create`
- [x]  `create_sibling` #5011
- [x]  `drop` #5012
- [x]  `get`
- [x]  `publish` -- moving to deprecated with https://github.com/datalad/datalad-deprecated/pull/28
- [x]  `remove` -- fixed with #6111
- [x]  `uninstall`
- [x]  `update`
- [x]  `clean`
- [x]  `diff` -- `rev_diff` replacement ready
- [x]  `save`
- [x] `unlock` (gh-3459)
- [x]  `aggregate` -- workaround implemented in #6128 to-be-retired via new metadata stack from -revolution
- [x] `metadata` -- workaround implemented in #6128 to-be-retired via new metadata stack from -revolution (@mih tried to RF it to use `status()` instead, but this is so intertwined with the assumptions and behavior of `annotate_paths()` that it would end up being a rewrite -- a rewrite that is already ongoing from a different angle by @christian-monch )

PR #3735, opened 2019-09-30
Removal of obsolete add/save

mih:
This is not intended to be merged any time soon, but is critical for #3368 and will likely involve a relatively elaborate RF (due to issues like #3714). Therefore I think it is best to put up a draft now, and let it simmer till one cannot see the potatoes anymore ;-)

- [x] core tests pass
- [x] benchmarks cleaned up
- [ ] -neuroimaging tests pass
- [ ] -container tests pass


datalad/distribution/tests/test_add.py
```diff
@@ -1,477 +0,0 @@
...
-    with open(attrfile, 'a') as f:
-        f.write('# just a comment\n')
-    ds.add('.')
-    ok_clean_git(ds.path)
-    ok_file_under_git(ds.path, op.relpath(attrfile, start=ds.path), annexed=False)
-
-
-# Failed to run ['git', '--work-tree=.', 'diff', '--raw', '-z', '--ignore-submodules=none', '--abbrev=40', 'HEAD', '--'] This operation must be run in a work tree
-@with_tempfile(mkdir=True)
-def test_gh1597(path):
```
yarikoptic: I think at least some tests need just to migrate into using new save to guarantee that we don't breed regressions etc

mih: All relevant tests have been ported to the new implementation last year already.

mih:
I will close this. A lot has happened, but still unforeseeable when we can pull this off.

PR #3737, opened 2019-10-01
TST: Let actions run on 'pull_request' events

mih:
Running them on 'push' will cause actions to run in the repo
that has been pushed into. However, in almost all cases we
want to see the result in the base repo.

Moreover, I feel better to not kick of users of computing
for each push that doesn't happen in a PR.


mih:
Actions were triggered as desired. No need to wait for a full test run.Canceled.

PR #3739, opened 2019-10-01
Further cleanup in Repo classes

mih:
Mostly code homogenization with patterns with use elsewhere.


yarikoptic:
Such changes would breed conflicts for having 0.11.x . I know that your don't care about it but we still maintain and release from it and 0.12 isn't released yet. So may be it would be better to postpone such harmonization efforts until 0.11.x is left to rest?

yarikoptic:
Or first harmonize 0.11.x code and then merge and compete in master please

datalad/support/annexrepo.py
```diff
@@ -544,22 +536,6 @@ def _check_git_annex_version(cls):
             raise OutdatedExternalDependency(ver_present=ver, **exc_kwargs)
         cls.git_annex_version = ver
 
-    @classmethod
-    def check_direct_mode_support(cls):
```
yarikoptic: I feel that it was left for a purpose of us being able to at least detect mode support and datalad spitting out DirectModeNoLongerSupportedError in case of running into a such repo. That was the test you are removing about. We no longer want/need that ? How should datalad behave if stepping into a direct mode repo with some older supported git annex?

mih: I believe the actual checker `is_direct_mode()` is untouched and functional.

mih:
I am fine not to merge this. I tried to get `annotate_paths()` out, but I have reached a point where I am faced with wrappers around wrappers for path handling that carry complexities that stem from direct mode support, plus an army of helpers for batch mode queries of properties that are now often provided by high-level code to start with. It might be better to just regrow the Repo classes from a minimal set of functionality, and leave the old ones to rot.

I somehow need to break the spell.

datalad/support/annexrepo.py
kyleam: > I feel that it was left for a purpose of us being able to at least detect mode support and datalad spitting out DirectModeNoLongerSupportedError in case of running into a such repo.

As @mih mentioned, aborting with DirectModeNoLongerSupportedError still works and is tested.

There are two parts to test_direct_cfg:

  1) check that trying to initialize an annex repo in direct mode reports that DataLad no longer supports it
  2) check that we can call `_set_direct_mode(True)` and get into a direct mode repo despite DataLad not supporting it

That second part fails if git-annex itself doesn't support direct mode.  That's why check_direct_mode_support was restored in af94ddc36 (ENH: annexrepo: Add check_direct_mode_support method, again, 2019-09-06).

>  That was the test you are removing about. We no longer want/need that ? How should datalad behave if stepping into a direct mode repo with some older supported git annex?

As mentioned above, the part of the test being removed check that we can force direct mode with `_set_direct_mode(True)`.  I don't think that is a behavior that we need to test (or even necessarily should), especially give the latest git-annex has removed direct mode entirely.  So I'm fine with deleting part 2 of test_direct_cfg and pruning the then unused check_direct_mode_support.

Issue #3661, opened 2019-09-11
`uninstall --recursive *` does not recurse into subdatasets


mih:
Here is what I see:

```
% datalad uninstall -r --nocheck *
uninstall(impossible): /tmp/alot/00 (directory) [can only uninstall datasets (consider the `drop` command)]
uninstall(impossible): /tmp/alot/01 (directory) [can only uninstall datasets (consider the `drop` command)]
...
```

The structure of the dataset is this (the uuid-like directories are the actual subdatasets)

```
.
├── 00
│   ├── 0ea832de-bec0-45a9-989f-c0376394d6e6
│   ├── 3790b8c2-1beb-494c-a5a5-189ccb9737b6
│   ├── 4255e451-d5d4-4829-8053-8c2e55c9a647
│   ├── f339565c-345f-42ce-a7ed-f83e1733a237
│   └── f617fc62-0b65-4471-97f4-f8d006f929a2
├── 01
│   ├── 1c9a767d-0b4c-4214-a8b3-403666191251
│   ├── 545dd1fc-b816-4725-a11f-1aafbb71a6bf
│   ├── 67d925e2-1bd9-4ec3-9e04-be15aa5f49e1
│   ├── 884cb78d-02ee-4762-86fc-4b02f96e0e7f
│   ├── 8b773663-eb54-4c7e-bfc2-06ddfb83c6d8
│   ├── ad1461b2-40d8-4983-bbb0-8d5b125045a7
│   ├── d92e1877-f2f2-403d-ac81-dd747c34a234
│   └── f673af13-1fe8-44c3-aa52-e382299c59e4

```

I somewhat expected the above command to uninstall any subdatasets that match the given path, everything underneath the given path (because `--recursive`).

Question is: what should we do, if we call `uninstall` with a path to a directory that has two subdatasets underneath, but **no** recursive is given?

yarikoptic:
Without -r shouldn't uninstall them I guess unless path points direct it to them

mih:
Q is: what does recursion mean in this case. I can never uninstall a top-level subdataset without recursively uninstalling all potential subdatasets. So in the best case, this is just a flag that, if absent, prevents any operation in the case of subsubdatasets.

bpoldrack:
Hm. Absence of `--recursive` could cause the command to still try to uninstall datasets further down, but not passing `--nocheck`, `--if-dirty`.
However, I agree - it's a weird UI indeed.
[...]

Issue #3655, opened 2019-09-07
Performance observations with large subdataset numbers


mih:
I created 1000 datasets (not particular content). That takes about 500ms per dataset on my laptop. I used this code (yeah, lazy, I know)

```python
 [dl.create('/'.join((str(uuid.uuid4())[:2], str(uuid.uuid4())))) for i in range(1000)]  
```

Adding them to a superdataset worked without issues. However the results have the wrong type (`file`).

A superdataset with 1000 empty subdatasets weighs 315 MB on the filesystem.

```
% datalad save 
...
add(ok): ff/b143ff42-85c4-4d6f-bcc7-c4d9c6fd50cc (file)
add(ok): ff/e4287631-3e94-4183-9a68-e7f84d87c99d (file)
add(ok): .gitmodules (file)
save(ok): . (dataset)
action summary:
  add (ok: 1001)
  save (ok: 1)
datalad save  132.18s user 59.01s system 100% cpu 3:11.04 total
```

Subdataset reporting seems to be speedy and working fine.

```
% datalad subdatasets
...
subdataset(ok): ff/e4287631-3e94-4183-9a68-e7f84d87c99d (dataset)
action summary:
  subdataset (ok: 1000)
datalad subdatasets  0.72s user 0.18s system 96% cpu 0.933 total
```

A save attempt on the (clean) superdataset is a bit sluggish though

```
% time dl save
datalad save  35.30s user 18.30s system 100% cpu 53.371 total
```

Uninstall is a problem:

```
% datalad uninstall -r --nocheck */*
uninstall(ok): /tmp/alot/ff/e4287631-3e94-4183-9a68-e7f84d87c99d (dataset)
uninstall(ok): /tmp/alot/ff/b143ff42-85c4-4d6f-bcc7-c4d9c6fd50cc (dataset)
...
action summary:
  uninstall (ok: 1000)
datalad uninstall -r --nocheck */*  301.58s user 131.57s system 99% cpu 7:15.19 total
```

The superdataset's `.git` weighs ~25MB, and it has

```
% git count-objects 
1002 objects, 23436 kilobytes
```

roughly 25KB per subdataset.

yarikoptic:
Would be nice to have it as a benchmark if we don't have any. May be just use 100 datasets

mih:
Ping @loj @bpoldrack this is the relevant starting point when we are going for 30k+ subdatasets.

loj:
[...]

Issue #2967, opened 2018-11-01
Make uninstall uninstall top-level datasets


yarikoptic:
I think we had that discussion at some point but I do not think we have an explicit issue.  I think that API should become a bit more "symmetric".  We already can `drop` what we `get`, `remove` what we `add`, and IMHO we should be able to `uninstall` what we `install`.  But that is not the case ATM with top level datasets:

```shell
$> datalad install https://github.com/OpenNeuroDatasets/ds001499
[INFO   ] Cloning https://github.com/OpenNeuroDatasets/ds001499 [1 other candidates] into '/mnt/btrfs/scrap/tmp/openneuro/ds001499' 
install(ok): /mnt/btrfs/scrap/tmp/openneuro/ds001499 (dataset)

$> datalad uninstall ds001499 
uninstall(error): /mnt/btrfs/scrap/tmp/openneuro/ds001499 (dataset) [will not uninstall top-level dataset (consider `remove` command)]
```
and I cannot recall the original motivation for having it so (IIRC it was very much intentional).

mih:
The reason was to reflect the impact of the action. When one uninstall a top-level dataset it is actually gone, and cannot be brought back from what is left after the operation. The is different for any of the other "undo" examples that you gave above.

yarikoptic:
Well, I can bring back any dataset which I installed from somewhere else no problem (like in the above example). And I could have a subdataset which was locally created with content still available somewhere (eg s3) which if I uninstall I loose. Altogether I am not sure if availability of the installation url within git should be the factor deciding between uninstall and remove. I could even argue that removing something which is not a part of a dataset is inconsistent with above pairing of commands.

Git annex verifies presence of files to be dropped on remotes, I wonder if we just should verify availability of the datasets on the remotes they have and if no remotes have it, then refuse to uninstall it, as a factor to decide when to refuse to uninstall (unless --force)

I could have

Issue #1142, opened 2016-11-26
uninstall must check whether a dataset has a non-local remote and all changes are pushed


mih:
Probably for all branches.

Copied from merged #1130 TODO notes:

- there is no check if a subdatasets only has a local URL (e.g. ./anything), which implies that uninstall actually equals removal with no way back (no change from status quo)
- there is no check that the relevant branched are pushed to a remote before an uninstall (no change from status quo)

Also see #1144 for notes under what condition these checks need to be made.

PR #3741, opened 2019-10-01
RF: uninstall

mih:
- [x] no annotate_paths() anymore (gh-3368)
- [x] now uses standard subdatasets() for path-constraint recursion
      (fixes gh-3661)
- [x] now uses standard subdatasets() for fast generator-style
      processing of large numbers of subdatasets (fixes gh-3655)

      was:
      datalad uninstall -r --nocheck */*  301.58s user 131.57s system 99% cpu 7:15.19 total

      is:
      datalad uninstall -r --nocheck */*  111.21s user 48.07s system 105% cpu 2:31.61 total

### DIFF
--- a/datalad/distribution/remove.py
+++ b/datalad/distribution/remove.py
@@ -197,10 +197,7 @@ class Remove(Interface):
                         # this is for subdatasets of the to-be-removed dataset
                         # we want to simply uninstall them in a regular manner
                         for r in Uninstall.__call__(
-                                # use annotate path as input, but pass a copy because
-                                # we cannot rely on it being unaltered by reannotation
-                                # TODO maybe adjust annotate_path to do that
-                                [ap.copy()],
+                                ap['path'],
                                 dataset=refds_path, recursive=recursive, check=check,
                                 if_dirty=if_dirty, result_xfm=None, result_filter=None,
                                 on_failure='ignore'):
--- a/datalad/distribution/tests/test_uninstall.py
+++ b/datalad/distribution/tests/test_uninstall.py
@@ -61,6 +61,11 @@ def test_safetynet(path):
                 assert_status(
                     ('error', 'impossible'),
                     uninstall(path=target, on_failure='ignore'))
+    sub = ds.create('sub')
+    subsub = sub.create('subsub')
+    for p in (sub.path, subsub.path):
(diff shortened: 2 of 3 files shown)


mih:
> For the old install, you'd get an error saying you need to provide a dataset or path. If you try datalad uninstall -d., you'll get an error saying "refusing to uninstall current or parent directory". If you go outside of that dataset, a datalad -d subds or datalad subds will wipe it. I think that current behavior is bad, but I think allowing a bare datalad uninstall makes the situation worse.

I felt the same pain. But I don't have a good solution (maybe @yarikoptic has?). At the very least, it would make sense to have the same rules apply to an argument-less `uninstall` call. So when inside a subdataset, it refuses to uninstall anything that is at or above the CWD.

yarikoptic:
Ideally, similar to annex drop, we should check if current head commit(s) available from any remote, and refuse (without --nocheck) to uninstall if something is missing a copy from a remote location.
Although bare uninstall is indeed scary, with above check IMHO it should be ok. And it would be a symmetric behavior to install.

mih:
> Ideally, similar to annex drop, we should check if current head commit(s) (all branches) available from any remote, and refuse (without --nocheck) to uninstall if something is missing a copy from a remote location.

Ping https://github.com/datalad/datalad/issues/1142

> Although bare uninstall is indeed scary, with above check IMHO it should be ok. And it would be a symmetric behavior to install.

I don't think it is that simple, or indeed symmetric. Once the local repo is gone, there is no guarantee to bring it back with a single corresponding `install()` invocation (e.g. multiple remotes, local configuration, all branches pushed but not to a single homogeneous remote, etc.). The problem is only symmetric for the simplistic case of a repo that came from a single location, changes have been pushed back and uninstall happens. While this is already an arguably simplistic case, even here it is not really symmetric, as there is no data structure that would enable an argument-less `install()` to accompany the argument-less `uninstall()`.

This relates to a previous discussion on the difference of `uninstall` and `remove`: https://github.com/datalad/datalad/issues/2967

While all these things will need to be addressed eventually, I will go for the approach mentioned at https://github.com/datalad/datalad/pull/3741#issuecomment-538281318

mih:
Looking good, going in.

Issue #3742, opened 2019-10-01
FOI: install/get() behavior change when stripping annotate_paths()


mih:
When RFing commands to use `subdatasets()` or `status/diff()` and friends instead of `annotate_paths()` there is always one big behavior change implied: The former only support a single root dataset, whereas the latter could operate on arbitrary collections of path scattered across multiple disjoint dataset hierarchies. This is true for all commands listed in #3368, and has lead to said changes for all completed commands already.

Same will be the case for `install/get()` when fixing #3469 by RFing `annotate_paths()` out.

mih:
Closing FOI

Issue #3469, opened 2019-06-03
`install()` misinterprets relative path


mih:
We decided that in the cmdline, relative is always relative to PWD. This is not what happens in `install`

```
(datalad3-dev) mih@meiner /tmp % datalad install -d bids -s raw bids/sourcedata
[INFO   ] Cloning raw into '/tmp/bids/bids/bids/sourcedata' 
install(ok): bids/bids/sourcedata (dataset)                                                  
action summary:
  add (ok: 2)
  install (ok: 1)
  save (ok: 1)
(datalad3-dev) mih@meiner /tmp % datalad subdatasets -d bids
subdataset(ok): bids/bids/sourcedata (dataset)
```

So while the relative path given as source (`-s`) is treated as being relative to PWD, the target path is treated as being relative to the specified dataset path.

Even like this it doesn't work (pointless and uncalled for `bids/` subdirectory):

```
(datalad3-dev) mih@meiner /tmp % datalad install -d bids -s raw sourcedata
[INFO   ] Cloning raw into '/tmp/bids/bids/sourcedata' 
install(ok): bids/sourcedata (dataset)                                                       
action summary:
  add (ok: 2)
  install (ok: 1)
  save (ok: 1)
(datalad3-dev) mih@meiner /tmp % datalad -f json_pp subdatasets -d bids
{
  "gitmodule_name": "bids/sourcedata",
  "gitmodule_url": "/tmp/raw",
  "parentds": "/tmp/bids",
  "path": "/tmp/bids/bids/sourcedata",
  "refds": "/tmp/bids",
  "type": "dataset"
}
```

Only this gives the desired result:

```
% datalad -C bids install -d . -s ../raw sourcedata
[INFO   ] Cloning ../raw into '/tmp/bids/sourcedata' 
install(ok): sourcedata (dataset)                                                            
action summary:
  add (ok: 2)
  install (ok: 1)
  save (ok: 1)
(datalad3-dev) mih@meiner /tmp % datalad -f json_pp subdatasets -d bids
{
  "gitmodule_name": "sourcedata",
  "gitmodule_url": "/tmp/bids/../raw",
  "parentds": "/tmp/bids",
  "path": "/tmp/bids/sourcedata",
  "refds": "/tmp/bids",
  "type": "dataset"
}
```

yarikoptic:
well, was that behavior changed for any "old style" (`clone`, `publish`, etc) command?

yarikoptic:
[...]

Issue #3551, opened 2019-07-22
run: seems to incorrectly handle relative paths if ran within subdirectory


yarikoptic:
There could be more oddities or just that my eyes are sore already (couldn't reliably reproduce for a bit and then it "worked").  Calling datalad run with output relative path being somewhere `../` causes two kinds of (wrong) behavior when output path points to a subdataset:

1. immediate subdataset
```
/tmp/topds > cd /tmp; sudo rm -rf topds; datalad create /tmp/topds; cd /tmp/topds/; mkdir sub; touch sub/input; datalad save -m cleanup; datalad create -d . output; cd sub; datalad run --explicit --input ../sub/input --output ../output md5sum '{inputs}'                        
[INFO   ] Creating a new annex repo at /tmp/topds 
create(ok): /tmp/topds (dataset)                                                                                
add(ok): sub/input (file)
save(ok): . (dataset)
action summary:
  add (ok: 1)
  save (ok: 1)
[INFO   ] Creating a new annex repo at /tmp/topds/output 
create(ok): output (dataset)                                                                                    
action summary:
  add (ok: 2)
  create (ok: 1)
  save (ok: 1)
[INFO   ] Making sure inputs are available (this may take some time) 
[WARNING] path not associated with any dataset [get(/tmp)] 
get(impossible): .. (directory) [path not associated with any dataset]
[INFO   ] == Command start (output follows) ===== 
d41d8cd98f00b204e9800998ecf8427e  input
[INFO   ] == Command exit (modification check follows) ===== 
action summary:
  get (impossible: 1, notneeded: 1)

```
so seems to work but see the WARNING above -- `..` seems to be taken from the top of the dataset?

2. when output points to sub-subdataset, it just crashes:

```shell
/tmp/topds > cd /tmp; sudo rm -rf topds; datalad create /tmp/topds; cd /tmp/topds/; mkdir sub; touch sub/input; datalad save -m cleanup; datalad create -d . -c text2git output; datalad create -d output/sub; cd sub; datalad run --explicit --input ../sub/input --output ../output/sub bash -c 'md5sum {inputs} > {outputs}/checksum' 
[INFO   ] Creating a new annex repo at /tmp/topds 
create(ok): /tmp/topds (dataset)                                                                                
add(ok): sub/input (file)
save(ok): . (dataset)
action summary:
  add (ok: 1)
  save (ok: 1)
[INFO   ] Creating a new annex repo at /tmp/topds/output 
[...]

PR #3746, opened 2019-10-02
RF: get() without annotate_paths()

mih:
This tries to be a gentle RF that doesn't change (much) of the original
behavior (see test diff). However, a change in-line with #3742 is
unavoidable (see removed test).

This is bringing path semantics in line with status/subdatasets, and is
a precondition to fixing #3469

Ping #3368

### DIFF
--- a/datalad/core/local/run.py
+++ b/datalad/core/local/run.py
@@ -285,7 +285,10 @@ def prepare_inputs(dset, inputs, extra_inputs=None):
         for res in _install_and_reglob(dset, gp):
             yield res
         for res in dset.get(gp.expand(full=True), on_failure="ignore"):
-            if res.get("state") == "absent":
+            if res.get("action") == "get" and \
+                    res.get("status") == "impossible" and \
+                    res.get("message") == "path does not exist":
+                # MIH why just a warning if given inputs are not valid?
                 lgr.warning("Input does not exist: %s", res["path"])
             else:
                 yield res
--- a/datalad/distribution/get.py
+++ b/datalad/distribution/get.py
@@ -602,8 +602,11 @@ class Get(Interface):
                 dataset=dataset,
                 # always come from the top to get sensible generator behavior
                 bottomup=False,
-                recursive=recursive,
-                recursion_limit=recursion_limit,
+                # when paths are given, they will constrain the recursion
+                # automatically, and we need to enable recursion so we can
+                # location path in subdatasets several levels down
+                recursive=True if path else recursive,
+                recursion_limit=None if path else recursion_limit,
                 return_type='generator',
                 on_failure='ignore'):
             if sdsres.get('type', None) != 'dataset':
--- a/datalad/distribution/get.py
+++ b/datalad/distribution/get.py
@@ -384,6 +384,7 @@ def _install_targetpath(
             path=ds.path,
             status='notneeded',
             contains=[target_path],
+            refds=refds_path,
         )
     else:
         # we don't have it yet. is it in a subdataset?
@@ -401,6 +402,7 @@ def _install_targetpath(
                 action='get',
                 path=str(target_path),
                 status='impossible',
+                refds=refds_path,
                 message='path does not exist',
             )
             return
--- a/datalad/distribution/get.py
+++ b/datalad/distribution/get.py
@@ -290,7 +290,7 @@ def _install_necessary_subdatasets(
         # report installation, whether it helped or not
         yield get_status_dict(
             'install', ds=sd, status='ok', logger=lgr, refds=refds_path,
-            message=("Installed subdataset in order to get %s", path))
+            message=("Installed subdataset in order to get %s", str(path)))
 
         # now check whether the just installed subds brought us any closer to
         # the target path
--- a/datalad/distribution/get.py
+++ b/datalad/distribution/get.py
@@ -12,58 +12,67 @@
 
 import logging
 
-from os.path import join as opj
-from os.path import relpath
+import os.path as op
 
 from datalad.interface.base import Interface
-from datalad.interface.annotate_paths import AnnotatePaths
-from datalad.interface.annotate_paths import annotated2content_by_ds
 from datalad.interface.utils import eval_results
 from datalad.interface.base import build_doc
-from datalad.interface.results import get_status_dict
-from datalad.interface.results import results_from_paths
-from datalad.interface.results import annexjson2result
-from datalad.interface.results import count_results
-from datalad.interface.results import success_status_map
-from datalad.interface.results import results_from_annex_noinfo
-from datalad.interface.common_opts import recursion_flag
-# from datalad.interface.common_opts import git_opts
-# from datalad.interface.common_opts import annex_opts
-# from datalad.interface.common_opts import annex_get_opts
-from datalad.interface.common_opts import location_description
-from datalad.interface.common_opts import jobs_opt
-from datalad.interface.common_opts import reckless_opt
-from datalad.interface.common_opts import verbose
-from datalad.support.constraints import EnsureInt
-from datalad.support.constraints import EnsureChoice
-from datalad.support.constraints import EnsureStr
-from datalad.support.constraints import EnsureNone
+from datalad.interface.results import (
+    get_status_dict,
+    results_from_paths,
+    annexjson2result,
+    count_results,
+    success_status_map,
+    results_from_annex_noinfo,
+)
+from datalad.interface.common_opts import (
+    recursion_flag,
+    location_description,
+    jobs_opt,
+    reckless_opt,
+)
+from datalad.interface.results import is_ok_dataset
+from datalad.support.constraints import (
+    EnsureInt,
+    EnsureChoice,
+    EnsureStr,
+    EnsureNone,
+)
 from datalad.support.param import Parameter
 from datalad.support.annexrepo import AnnexRepo
 from datalad.support.gitrepo import (
     GitRepo,
     _fixup_submodule_dotgit_setup,
 )
-from datalad.support.exceptions import InsufficientArgumentsError
-from datalad.support.exceptions import InstallFailedError
-from datalad.support.exceptions import IncompleteResultsError
-from datalad.support.network import URL
-from datalad.support.network import RI
-from datalad.support.network import urlquote
-from datalad.dochelpers import exc_str
-from datalad.dochelpers import single_or_plural
-from datalad.utils import get_dataset_root
-from datalad.utils import with_pathsep as _with_sep
-from datalad.utils import unique
-from datalad.utils import path_startswith
-from datalad.utils import path_is_subpath
+from datalad.support.exceptions import (
+    InsufficientArgumentsError,
+    InstallFailedError,
+    IncompleteResultsError,
+)
+from datalad.support.network import (
+    URL,
+    RI,
+    urlquote,
+)
+from datalad.dochelpers import (
+    exc_str,
+    single_or_plural,
+)
+from datalad.utils import (
+    unique,
+    Path,
+)
 
-from .dataset import Dataset
-from .dataset import EnsureDataset
-from .dataset import datasetmethod
-from .clone import Clone
-from .utils import _get_flexible_source_candidates
-from .utils import _get_tracking_source
+from datalad.local.subdatasets import Subdatasets
+
+from datalad.distribution.dataset import (
+    Dataset,
+    EnsureDataset,
+    datasetmethod,
+    require_dataset,
+)
+from datalad.distribution.clone import Clone
+from datalad.distribution.utils import _get_flexible_source_candidates
 
 __docformat__ = 'restructuredtext'
 
@@ -145,7 +154,7 @@ def _install_subds_from_flexible_source(
         ds, sm_path, sm_url)
 
     # prevent inevitable exception from `clone`
-    dest_path = opj(ds.path, sm_path)
+    dest_path = op.join(ds.path, sm_path)
     clone_urls = [src for src in clone_urls if src != dest_path]
 
     if not clone_urls:
@@ -247,15 +256,11 @@ def _install_necessary_subdatasets(
     path: str
     reckless: bool
     """
-
-    def subds_result_filter(res):
-        return res.get('status') == 'ok' and res.get('type') == 'dataset'
-
     # figuring out what dataset to start with, --contains limits --recursive
     # to visit only subdataset on the trajectory to the target path
     subds_trail = ds.subdatasets(contains=path, recursive=True,
                                  on_failure="ignore",
-                                 result_filter=subds_result_filter)
+                                 result_filter=is_ok_dataset)
     if not subds_trail:
         # there is not a single known subdataset (installed or not)
         # for this path -- job done
@@ -269,7 +274,7 @@ def _install_necessary_subdatasets(
         try:
             sd = _install_subds_from_flexible_source(
                 Dataset(cur_subds['parentds']),
-                relpath(cur_subds['path'], start=cur_subds['parentds']),
+                op.relpath(cur_subds['path'], start=cur_subds['parentds']),
                 cur_subds['gitmodule_url'],
                 reckless,
                 description=description)
@@ -291,7 +296,7 @@ def _install_necessary_subdatasets(
         # the target path
         subds_trail = sd.subdatasets(contains=path, recursive=False,
                                      on_failure='ignore',
-                                     result_filter=subds_result_filter)
+                                     result_filter=is_ok_dataset)
         if not subds_trail:
             # no (newly available) subdataset get's us any closer
             return
@@ -305,16 +310,16 @@ def _recursive_install_subds_underneath(ds, recursion_limit, reckless, start=Non
         return
     # install using helper that give some flexibility regarding where to
     # get the module from
+
     for sub in ds.subdatasets(
-            return_type='generator', result_renderer='disabled'):
+            path=start,
+            return_type='generator',
+            result_renderer='disabled'):
         subds = Dataset(sub['path'])
         if sub.get('gitmodule_datalad-recursiveinstall', '') == 'skip':
             lgr.debug(
                 "subdataset %s is configured to be skipped on recursive installation",
                 sub['path'])
-            continue
-        if start is not None and not path_is_subpath(subds.path, start):
-            # this one we can ignore, not underneath the start path
             continue
         if sub.get('state', None) != 'absent':
             # dataset was already found to exist
@@ -328,7 +333,7 @@ def _recursive_install_subds_underneath(ds, recursion_limit, reckless, start=Non
             try:
                 subds = _install_subds_from_flexible_source(
                     ds,
-                    relpath(sub['path'], start=ds.path),
+                    op.relpath(sub['path'], start=ds.path),
                     sub['gitmodule_url'],
                     reckless,
                     description=description)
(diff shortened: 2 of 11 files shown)


yarikoptic:
Removed test is `test_autoresolve_multiple_datasets`. So it will not be possible to `get sub*/anat` if every subject is a separate sub dataset?

mih:
> Removed test is `test_autoresolve_multiple_datasets`. So it will not be possible to `get sub*/anat` if every subject is a separate sub dataset?

It is possible and tested, as long as it is running in a dataset that is parent to all.

datalad/distribution/get.py
```diff
@@ -247,15 +256,11 @@ def _install_necessary_subdatasets(
...
-
-    def subds_result_filter(res):
-        return res.get('status') == 'ok' and res.get('type') == 'dataset'
-
     # figuring out what dataset to start with, --contains limits --recursive
     # to visit only subdataset on the trajectory to the target path
     subds_trail = ds.subdatasets(contains=path, recursive=True,
                                  on_failure="ignore",
-                                 result_filter=subds_result_filter)
+                                 result_filter=is_ok_dataset)
```
kyleam: When adding f634bea13 (TST+BF: Update subdatasets(contains=...) calls for previous commit, 2019-10-01), I thought a function like `is_ok_dataset` might exist, but failed to find it.  Thanks for cleaning this up.

yarikoptic:
> > Removed test is `test_autoresolve_multiple_datasets`. So it will not be possible to `get sub*/anat` if every subject is a separate sub dataset?
> 
> It is possible and tested, as long as it is running in a dataset that is parent to all.

## Gotcha 1:  considers current path (repository) even though abs path(s) point to another dataset

- does not happen with current master (0.12.0rc5-175-g13ee49884)
- affects `install` as well

<details>
<summary>example 1 - running within some other unrelated repo</summary> 

```shell
/home/yoh/proj/datalad/datalad-master > datalad get ~/datalad/openfmri/ds*/sub-01/anat
[ERROR  ] /home/yoh/proj/datalad/datalad-master/.git [fun.py:is_git_dir:47] (WorkTreeRepositoryUnsupported) 

exit:1 /home/yoh/proj/datalad/datalad-master > git co master
Switched to branch 'master'
Your branch is behind 'origin/master' by 73 commits, and can be fast-forwarded.
  (use "git pull" to update your local branch)
changes on filesystem:                                                                                                                      
 .asv | 0

/home/yoh/proj/datalad/datalad-master > datalad get ~/datalad/openfmri/ds*/sub-01/anat
Total:   0%|                                                                                                    | 0.00/6.38M [00:00<?, ?B/s]ERROR:                                                                                                                                      
Interrupted by user while doing magic: KeyboardInterrupt() [cmd.py:_process_one_line:354]

/home/yoh/proj/datalad/datalad-master > datalad install ~/datalad/openfmri/ds*/sub-01/anat
[ERROR  ] /home/yoh/proj/datalad/datalad-master/.git [fun.py:is_git_dir:47] (WorkTreeRepositoryUnsupported) 
```
</details>

<details>
<summary>example 2 -- running from a pure `/tmp` -- NoDatasetArgumentFound</summary> 

```shell
/tmp > datalad get ~/datalad/openfmri/ds*/sub-01/anat
[ERROR  ] No dataset found [dataset.py:require_dataset:594] (NoDatasetArgumentFound) 
usage: datalad get [-h] [-s LABEL] [-d PATH] [-r] [-R LEVELS] [-n]
                   [-D DESCRIPTION] [--reckless] [-J NJOBS]
                   [PATH [PATH ...]]

```
</details>

I think it should not consider/be affected by CWD (could be not "available" for many other reasons) whenever full paths are provided.

## Gotcha 2: it does not `get` if superdataset is specified as the dataset

<details>
<summary>even if I do specify top (superdataset) dataset (`~/datalad`) </summary> 

```shell
exit:1 /tmp > datalad get -d ~/datalad ~/datalad/openfmri/ds*/sub-01/anat 
action summary:
  get (notneeded: 7)

/tmp > du -scmL ~/datalad/openfmri/ds*/sub-01/anat                
1	/home/yoh/datalad/openfmri/ds000001/sub-01/anat
4	/home/yoh/datalad/openfmri/ds000002/sub-01/anat
du: cannot access '/home/yoh/datalad/openfmri/ds000003/sub-01/anat/sub-01_T1w.nii.gz'
du: cannot access '/home/yoh/datalad/openfmri/ds000003/sub-01/anat/sub-01_inplaneT2.nii.gz'
0	/home/yoh/datalad/openfmri/ds000003/sub-01/anat
du: cannot access '/home/yoh/datalad/openfmri/ds000011/sub-01/anat/sub-01_T1w.nii.gz'
du: cannot access '/home/yoh/datalad/openfmri/ds000011/sub-01/anat/sub-01_inplaneT2.nii.gz'
0	/home/yoh/datalad/openfmri/ds000011/sub-01/anat
du: cannot access '/home/yoh/datalad/openfmri/ds000109/sub-01/anat/sub-01_T1w.nii.gz'
0	/home/yoh/datalad/openfmri/ds000109/sub-01/anat
du: cannot access '/home/yoh/datalad/openfmri/ds000216/sub-01/anat/sub-01_T1w.nii.gz'
0	/home/yoh/datalad/openfmri/ds000216/sub-01/anat
du: cannot access '/home/yoh/datalad/openfmri/ds000241/sub-01/anat/sub-01_T1w.nii.gz'
0	/home/yoh/datalad/openfmri/ds000241/sub-01/anat
4	total
```
</details>

<details>
<summary>seems to work ok in master</summary> 

```shell
/tmp > datalad get -d ~/datalad ~/datalad/openfmri/ds*/sub-01/anat
[INFO   ] To obtain some keys we need to fetch an archive of size 412.8 MB                                                                  
                                                                                                                                           ^C[WARNING] Still have 1 active progress bars when stopping                                                      | 0.00/5.71M [00:00<?, ?B/s]
ERROR:                                                                                                                                      
Interrupted by user while doing magic: KeyboardInterrupt() [cmd.py:_process_one_line:354]
```
</details>

<details>
<summary>works if I point directly to the subdataset containing those from which to get (`~/datalad/openfmri`) </summary> 

```shell
/tmp > datalad get -d ~/datalad/openfmri ~/datalad/openfmri/ds*/sub-01/anat
get(ok): ds000011/sub-01/anat/sub-01_inplaneT2.nii.gz (file) [from web...]                                                                  
get(ok): ds000011/sub-01/anat/sub-01_T1w.nii.gz (file) [from web...]                                                                        
get(ok): ds000011/sub-01/anat (directory)                                                                       | 0.00/3.54M [00:00<?, ?B/s]
action summary:
  get (notneeded: 6, ok: 3)
```
</details>

<details>
<summary>works with `-r` </summary> 

```shell
/tmp > datalad get -d ~/datalad -r ~/datalad/openfmri/ds*/sub-01/anat
[INFO   ] Installing <Dataset path=/home/yoh/datalad/openfmri> underneath /home/yoh/datalad/openfmri/ds000001/sub-01/anat recursively 
[INFO   ] Installing <Dataset path=/home/yoh/datalad/openfmri> underneath /home/yoh/datalad/openfmri/ds000002/sub-01/anat recursively 
[INFO   ] Installing <Dataset path=/home/yoh/datalad/openfmri> underneath /home/yoh/datalad/openfmri/ds000003/sub-01/anat recursively 
[INFO   ] Installing <Dataset path=/home/yoh/datalad/openfmri> underneath /home/yoh/datalad/openfmri/ds000011/sub-01/anat recursively 
[INFO   ] Installing <Dataset path=/home/yoh/datalad/openfmri> underneath /home/yoh/datalad/openfmri/ds000109/sub-01/anat recursively 
[INFO   ] Installing <Dataset path=/home/yoh/datalad/openfmri> underneath /home/yoh/datalad/openfmri/ds000216/sub-01/anat recursively 
[INFO   ] Installing <Dataset path=/home/yoh/datalad/openfmri> underneath /home/yoh/datalad/openfmri/ds000241/sub-01/anat recursively 
[INFO   ] Installing <Dataset path=/home/yoh/datalad/openfmri/ds000001> underneath /home/yoh/datalad/openfmri/ds000001/sub-01/anat recursively 
[INFO   ] Installing <Dataset path=/home/yoh/datalad/openfmri/ds000002> underneath /home/yoh/datalad/openfmri/ds000002/sub-01/anat recursively 
[INFO   ] Installing <Dataset path=/home/yoh/datalad/openfmri/ds000003> underneath /home/yoh/datalad/openfmri/ds000003/sub-01/anat recursively 
[INFO   ] Installing <Dataset path=/home/yoh/datalad/openfmri/ds000011> underneath /home/yoh/datalad/openfmri/ds000011/sub-01/anat recursively 
[INFO   ] Installing <Dataset path=/home/yoh/datalad/openfmri/ds000109> underneath /home/yoh/datalad/openfmri/ds000109/sub-01/anat recursively 
[INFO   ] Installing <Dataset path=/home/yoh/datalad/openfmri/ds000216> underneath /home/yoh/datalad/openfmri/ds000216/sub-01/anat recursively 
[INFO   ] Installing <Dataset path=/home/yoh/datalad/openfmri/ds000241> underneath /home/yoh/datalad/openfmri/ds000241/sub-01/anat recursively 
get(ok): openfmri/ds000216/sub-01/anat/sub-01_T1w.nii.gz (file) [from web...]
get(ok): openfmri/ds000216/sub-01/anat (directory)
sub-01/anat/sub-01_T1w.nii.gz:  69%|██████████████████████████████████████████████▏                    | 4.56M/6.62M [00:01<00:00, 3.15MB/s]^C[WARNING] Still have 1 active progress bars when stopping 
ERROR:                                                                                                                                      
Interrupted by user while doing magic: KeyboardInterrupt() [cmd.py:_process_one_line:354]

```

Those `INFO` msgs happen also with master version so not specific to the PR
</details>

mih:
> I think it should not consider/be affected by CWD (could be not "available" for many other reasons) whenever full paths are provided.

This code is intentionally incapable of doing that. The dataset-unbound behavior is THE reason why nobody can or will fix annotate_paths().

mih:
@kyleam [This run test failure](https://travis-ci.org/datalad/datalad/jobs/592984928#L2584) is due to `get()` now following the present path interpretation conventions. Here is the test:

```
    # ... producing output file in specified dataset and passing output file as
    # relative to current directory
    with chpwd(ds1_subdir):
        out = op.join(ds0.path, "three")
        run("cd .> {}".format(out), dataset=ds0.path, explicit=True,
            outputs=[op.relpath(out, ds1_subdir)])
 ```

Leading to

```
> /home/mih/hacking/datalad/git/datalad/core/local/run.py(256)_install_and_reglob()
-> on_failure="ignore"):
(Pdb) dirs_new
['../../ds0']
(Pdb) l
251  
252         dirs, dirs_new = [], glob_dirs()
253         while dirs != dirs_new:
254             for res in dset.install(dirs_new,
255                                     result_xfm=None, return_type='generator',
256  ->                                 on_failure="ignore"):
257                 if res.get("state") == "absent":
258                     lgr.debug("Skipping install of non-existent path: %s",
259                               res["path"])
260                 else:
261                     yield res
```

So the incoming arg is given unmodified to `install` (and underneath `get`), but it was originally handed to an unbound `run`, while internally calling a bound `install, hence invalidating the semantics.

In general I believe that there are two major paradigms to path handling.

1) Resolve to absolute paths at the start of a command
2) Make at all internal dataset method calls unbound, using the original `dataset` argument verbatim

In `run`'s case (1) can be rather difficult or impossible (globs etc.), so I guess (2) is the way to go. This would require some RF'ing in order to get the `dataset` arg to the places it needs to be passed. What do you think?

**Comment by mih** [comment]
@yarikoptic re gotcha2: I cannot replicate this:

```
(datalad3-dev) mih@meiner /tmp % dl install -s /// ts
[INFO   ] Cloning http://datasets.datalad.org/ [1 other candidates] into '/tmp/ts' 
install(ok): /tmp/ts (dataset)
(datalad3-dev) mih@meiner /tmp % dl get -d /tmp/ts /tmp/ts/openfmri/ds000001/sub-01/anat
[INFO   ] Cloning http://datasets.datalad.org/openfmri/.git into '/tmp/ts/openfmri' 
install(ok): openfmri (dataset) [Installed subdataset in order to get /tmp/ts/openfmri/ds000001/sub-01/anat]
[INFO   ] Cloning http://datasets.datalad.org/openfmri/ds000001/.git into '/tmp/ts/openfmri/ds000001' 
install(ok): openfmri/ds000001 (dataset) [Installed subdataset in order to get /tmp/ts/openfmri/ds000001/sub-01/anat]
get(ok): openfmri/ds000001/sub-01/anat/sub-01_inplaneT2.nii.gz (file) [from web...]          
get(ok): openfmri/ds000001/sub-01/anat/sub-01_T1w.nii.gz (file) [from web...]                
get(ok): openfmri/ds000001/sub-01/anat (directory)        | 223k/5.66M [00:00<00:12, 447kB/s]
action summary:
  get (ok: 3)
  install (ok: 2)
```

also multi-paths are fine

```
(datalad3-dev) 1 mih@meiner /tmp % dl get -d /tmp/ts /tmp/ts/openfmri/ds00000{2,3}/sub-01/anat
[INFO   ] Cloning http://datasets.datalad.org/openfmri/ds000002/.git into '/tmp/ts/openfmri/ds000002' 
install(ok): openfmri/ds000002 (dataset) [Installed subdataset in order to get /tmp/ts/openfmri/ds000002/sub-01/anat]
[INFO   ] Cloning http://datasets.datalad.org/openfmri/ds000003/.git into '/tmp/ts/openfmri/ds000003' 
install(ok): openfmri/ds000003 (dataset) [Installed subdataset in order to get /tmp/ts/openfmri/ds000003/sub-01/anat]
get(ok): openfmri/ds000002/sub-01/anat/sub-01_T1w.nii.gz (file) [from web...]                
get(ok): openfmri/ds000002/sub-01/anat (directory)
get(ok): openfmri/ds000003/sub-01/anat/sub-01_inplaneT2.nii.gz (file) [from web...]          
get(ok): openfmri/ds000003/sub-01/anat/sub-01_T1w.nii.gz (file) [from web...]                
get(ok): openfmri/ds000003/sub-01/anat (directory)      | 1.41M/5.71M [00:00<00:03, 1.13MB/s]
action summary:
  get (ok: 5)
  install (ok: 2)
```

**Comment by yarikoptic** [comment]
> dl install -s /// ts

Either please use full "standard" name (`datalad`) or we should centrally deploy `datalad` as `dl`.  Trying to reproduce your snippets requires additional tuning, and I don't want to breed custom `dl` in bashrc (e.g. I never post use of my shortcuts `dg` `d+` etc)

**Comment by yarikoptic** [comment]
<details>
<summary>getting bitten by gotcha1 (if intended - not sure if I like that new intended behavior - breaks our original promise of being able to use datalad from outside of the target dataset, might be introducing side effects from taking CWD dataset configuration and applying it to operation on the target path/dataset, etc)</summary> 

```shell
$> sudo rm -rf /tmp/ts && datalad install -s /// /tmp/ts && datalad install /tmp/ts/openfmri/ds00000{2,3} && datalad get -d /tmp/ts /tmp/ts/openfmri/ds0000*/sub-01/anat
[INFO   ] Cloning http://datasets.datalad.org/ [1 other candidates] into '/tmp/ts' 
install(ok): /tmp/ts (dataset)
[ERROR  ] No dataset found [dataset.py:require_dataset:594] (NoDatasetArgumentFound) 
usage: datalad install [-h] [-s SOURCE] [-d DATASET] [-g] [-D DESCRIPTION]
                       [-r] [-R LEVELS] [--nosave] [--reckless] [-J NJOBS]
                       [PATH [PATH ...]]
```
</details>

in your replication attempt of gotcha2 you didn't replicate the situation -- I had already those subdatasets installed.  Here is a complete minimal example where I first install those subdatasets:

```shell
$> sudo rm -rf /tmp/ts && datalad install -s /// /tmp/ts && builtin cd /tmp/ts && datalad install openfmri/ds00000{2,3} && builtin cd /tmp && datalad get -d /tmp/ts /tmp/ts/openfmri/ds0000*/sub-01/anat       
[INFO   ] Cloning http://datasets.datalad.org/ [1 other candidates] into '/tmp/ts' 
install(ok): /tmp/ts (dataset)
[INFO   ] Cloning http://datasets.datalad.org/openfmri/.git into '/tmp/ts/openfmri' 
[INFO   ] Cloning http://datasets.datalad.org/openfmri/ds000002/.git into '/tmp/ts/openfmri/ds000002'                                       
install(ok): /tmp/ts/openfmri/ds000002 (dataset) [Installed subdataset in order to get /tmp/ts/openfmri/ds000002]                           
[INFO   ] Cloning http://datasets.datalad.org/openfmri/ds000003/.git into '/tmp/ts/openfmri/ds000003' 
install(ok): /tmp/ts/openfmri/ds000003 (dataset) [Installed subdataset in order to get /tmp/ts/openfmri/ds000003]                           
action summary:
  install (ok: 3)
datalad install openfmri/ds00000{2,3}  5.31s user 1.98s system 74% cpu 9.740 total
action summary:
  get (notneeded: 2)
```
and works with `openfmri` sub being specified:
```shell
$> datalad get -d /tmp/ts/openfmri /tmp/ts/openfmri/ds0000*/sub-01/anat 
get(ok): ds000002/sub-01/anat/sub-01_T1w.nii.gz (file) [from web...]                                                                        
get(ok): ds000002/sub-01/anat (directory)
get(ok): ds000003/sub-01/anat/sub-01_inplaneT2.nii.gz (file) [from web...]                                                                  
get(ok): ds000003/sub-01/anat/sub-01_T1w.nii.gz (file) [from web...]                                                                        
get(ok): ds000003/sub-01/anat (directory)  
```

mih:
> getting bitten by gotcha1

I don't mind `install()` doing all kinds of stunts. From my POV it is already an outlier in terms of behavior, so I don't mind to make it fit this (or other) use cases. Install is merely a loop around `clone/get()` so we can easily treat every single input argument as independent. Performance will be an issue, but I don't think performance is an issue for `install()` in general.

> re gotcha2: thanks for the pointer. Found it, and pushed a fix.

kyleam:
@mih re run test issues: These failures are related to gh-3551.  I'd suggest marking them as known failures so that this PR isn't blocked (this PR brings them to the forefront but doesn't introduce them), and I'll work on fixing them.

---

More details: I started looking at this a while ago.  One approach to the install issue is expand those return values.

Here's the patch I had:

<details>
<summary>patch</summary>

```
Subject: [PATCH] BF: run: Pass absolute paths to install

We do a series of globs and installs to make sure the required
datasets are installed.  During this procedure, we expand the globs
with relative paths (full=False).  This is incorrect in two ways.
First, we expect calling dirname() on these paths to give something
useful, but that can easily result in an empty string when the paths
are relative.  Second, we pass the relative to a dataset-bound install
call, which considers the paths as relative to the dataset, leading to
incorrectly specified paths if the call is occurring from a
subdirectory [*].

Expand the globs to absolute paths instead.

Re: #3551

[*] ... though these incorrect calls may still work.  See gh-3650.
---
 datalad/core/local/run.py             | 2 +-
 datalad/interface/tests/test_rerun.py | 3 ++-
 2 files changed, 3 insertions(+), 2 deletions(-)

diff --git a/datalad/core/local/run.py b/datalad/core/local/run.py
index e95473248..3d05b21ee 100644
--- a/datalad/core/local/run.py
+++ b/datalad/core/local/run.py
@@ -247,7 +247,7 @@ def _install_and_reglob(dset, gpaths):
     Generator with the results of the `install` calls.
     """
     def glob_dirs():
-        return list(map(op.dirname, gpaths.expand(refresh=True)))
+        return list(map(op.dirname, gpaths.expand(full=True, refresh=True)))
 
     dirs, dirs_new = [], glob_dirs()
     while dirs != dirs_new:
diff --git a/datalad/interface/tests/test_rerun.py b/datalad/interface/tests/test_rerun.py
index 7305f119c..0b089411c 100644
--- a/datalad/interface/tests/test_rerun.py
+++ b/datalad/interface/tests/test_rerun.py
@@ -593,7 +593,8 @@ def test_rerun_script(path):
                                         "b.txt": "b"}},
                         "s1_1": {"s2": {"c.dat": "c",
                                         "d.txt": "d"}},
-                        "ss": {"e.dat": "e"}}})
+                        "ss": {"e.dat": "e"}},
+                 "subdir": {"f.dat": "f"}})
 @with_tempfile(mkdir=True)
 def test_run_inputs_outputs(src, path):
     for subds in [("s0", "s1_0", "s2"),
-- 
2.23.0

```

</details>


The problem is that this exposes a core incompatibility between how `run`, from the very beginning, has handled calls from subdirectories and the kinds of absolute paths that `rev_resolve_path` will take:

https://github.com/datalad/datalad/blob/13ee49884779549c508de7a0eef6970e48077dc2/datalad/distribution/dataset.py#L612-L626

I'll revisit that issue today.

There's also a tangentially related issue that run's (or really the GlobbedPaths helper's) iterative globbing of subdatasets only works in a restricted set of cases.  That needs to be fixed (perhaps `_install_necessary_subdatasets` can help), but I'll punt on that for now unless fixing path handling issues above require it to be addressed.

datalad/core/local/run.py
```diff
@@ -285,7 +285,10 @@ def prepare_inputs(dset, inputs, extra_inputs=None):
         for res in _install_and_reglob(dset, gp):
             yield res
         for res in dset.get(gp.expand(full=True), on_failure="ignore"):
-            if res.get("state") == "absent":
+            if res.get("action") == "get" and \
+                    res.get("status") == "impossible" and \
+                    res.get("message") == "path does not exist":
+                # MIH why just a warning if given inputs are not valid?
```
kyleam: I don't spot any obvious conceptual reasons why we couldn't be stricter here, and I'd be fine with switching to an error.  However, quickly testing that out, we would need to rework things due to a non-obvious interaction with GlobbedPaths (shown by the resulting failure in `test_run_inputs_no_annex_repo`).  I'm going to open an issue about run's installation of missing subdatasets only working in a restricted set of cases (same issue mentioned elsewhere in this thread).  It looks like fixing that will require core change to the same part of GlobbedPaths, so I'll make sure to note that we should keep this "warning -> error" switch in mind.

kyleam:
> [This other failure](https://github.com/datalad/datalad/pull/3746/checks?check_run_id=246193575#step:8:374) seems to be related to glob expansion not working on windows. Do you consider this as "should work", given that this is windows?

I don't feel confident about saying anything should work on windows, but for this particular test the glob expansion isn't a hit on other systems either.  Because GlobbedPaths hangs onto the globs when there are no matches^,  get() receives a literal "*" as the path, and it gives back a result saying that the path doesn't exist.

With the gh-3747 changes, the call is equivalent to `get(dataset=ds.path, path=['*'])`, and that has the same failure:

https://github.com/datalad/datalad/pull/3747/checks?check_run_id=248506600#step:8:374

^ I'd need to review that code to recall why, but that's probably another aspect that needs to be reworked along with the changes mentioned [here](https://github.com/datalad/datalad/pull/3746#pullrequestreview-297713359).

kyleam:
I've rebased this to resolve conflicts with master, and I've marked test_run_inputs_no_annex_repo as a known windows failure.

<details>
<summary>range-diff</summary>

```
 1:  6c1ac2c6f =  1:  6060043e9 RF: get() without annotate_paths()
 2:  6158dd3b8 =  2:  35bb3fcaf BF: Adjust results to get install() result filter working
 3:  5b166d443 =  3:  096ea9aa3 BF: Not useful to send Path objects in messages
 4:  6c3ce76ba !  4:  e7242c343 TST: Adjust tests to match new calling conventions for get()
    @@ datalad/distribution/tests/test_uninstall.py: def test_remove_recursive_2(tdir):
     -        install('labs/tarr/face_place')
     +        with chpwd('labs'):
     +            install('tarr/face_place')
    -         remove('labs', recursive=True)
    +         remove(dataset='labs', recursive=True)
      
      
     
 5:  ab984551e =  5:  01029df1a RF: Adjust run for new get() error reporting
 6:  84ecda578 =  6:  2fe9d5d3a TST: No longer a pointless toplevel 'notneeded' result
 7:  89467b9c4 =  7:  276655687 BF: Fix up subdataset recursion logic in get()
 8:  c14e2acb1 =  8:  acfc6f145 TST: Mark known failures as suggested by @kyleam
 -:  --------- >  9:  a43da55f2 TST: Mark test_run_inputs_no_annex_repo as known Windows failure
```

</details>

kyleam:
The [remaining failure][0], datalad_container.tests.test_run.test_custom_call_fmt, is [resolved by][1] gh-3747.

[0]: https://travis-ci.org/datalad/datalad/jobs/594239385#L1180
[1]: https://travis-ci.org/datalad/datalad/jobs/594243964

mih:
Thx much @kyleam ! I'd be happy to close this one in favor of gh-3747

yarikoptic:
I have created a dedicated issue to discuss and possibly mitigate the removal of a feature with these changes: https://github.com/datalad/datalad/issues/3759  
I will proceed with merging this one and Kyle's #3747 (so we have clear separate merges), will do it locally

Issue #3748, opened 2019-10-05
`create()` does not yet obey new relpath semantics


mih:
```
(datalad3-dev) mih@meiner /tmp % datalad create top
[INFO   ] Creating a new annex repo at /tmp/top 
create(ok): /tmp/top (dataset)                                                               
(datalad3-dev) mih@meiner /tmp % datalad create -d top top/sub
[INFO   ] Creating a new annex repo at /tmp/top/top/sub 
create(ok): top/sub (dataset)                                                                
action summary:
  add (ok: 2)
  create (ok: 1)
  save (ok: 1)
```

It should have create `/top/sub` not `/top/top/sub`

PR #3749, opened 2019-10-05
BF: create() path semantics (fixes gh-3748)

mih:
Now with a more comprehensive test.


mih:
Given that the fix was trivial, I'll merge this now without a review.

Issue #3685, opened 2019-09-18
save provides no details for an error (in case of permission issues)


yarikoptic:
```shell
(git)smaug:/mnt/btrfs/datasets/datalad/crawl/dicoms/salo/MEEPI[master]6ET_MB4
$> datalad save -m 'initially downloaded from gdrive .zip' complex_multiecho_phantom_data.zip
add(error): 6ET_MB4/complex_multiecho_phantom_data.zip (file)
action summary:
  add (error: 1)
  save (notneeded: 1)
```
the actual reason could be unveiled only by running with `-l debug`:
```shell
...
[DEBUG  ] Determined 1 datasets for saving from input arguments 
[DEBUG  ] 1 path(s) to add to <AnnexRepo path=/mnt/btrfs/datasets/datalad/crawl/dicoms/salo/MEEPI (<class 'datalad.support.annexrepo.AnnexRepo'>)> {'6ET_MB4/complex_multiecho_phantom_data.zip': {'state': 'untracked', 'type': 'file'}} 
[DEBUG  ] Running add resulted in stderr output:   6ET_MB4/complex_multiecho_phantom_data.zip: setFileMode: permission denied (Operation not permitted)
git-annex: add: 1 failed
 
add(error): 6ET_MB4/complex_multiecho_phantom_data.zip (file)
```
initially detected with datalad 0.12.0rc4.dev330 and tested then with current master 0.12.0rc5.dev60.
I expected some `ERROR` level log msg or some other indication in the results renderer. But  json record contains no message or stderr from the original invocation:
```
$> datalad -f json_pp save -m 'initially downloaded from gdrive .zip' complex_multiecho_phantom_data.zip
{
  "action": "add",
  "key": null,
  "path": "/mnt/btrfs/datasets/datalad/crawl/dicoms/salo/MEEPI/6ET_MB4/complex_multiecho_phantom_data.zip",
  "refds": "/mnt/btrfs/datasets/datalad/crawl/dicoms/salo/MEEPI",
  "status": "error",
  "type": "file"
}
{
  "action": "save",
  "path": "/mnt/btrfs/datasets/datalad/crawl/dicoms/salo/MEEPI",
  "refds": "/mnt/btrfs/datasets/datalad/crawl/dicoms/salo/MEEPI",
  "status": "notneeded",
  "type": "dataset"
}
```
so I guess the proper fix would be to somehow channel error/stderr within this record so it could be rendered?

bpoldrack:
[...]

PR #3751, opened 2019-10-05
Capture error messages from git-annex's JSON output

mih:
Turns this:

```
% datalad save
add(error): that3 (file)
```

into

```
% datalad save
add(error): that3 (file) [  that3: setFileMode: permission denied (Operation not permitted)]
```

hence fixes gh-3685

Note that this does not fix the same issue in `add()`, because different code is used (see gh-3714).

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -3341,6 +3341,7 @@ class AnnexRepo(GitRepo, RepoInterface):
 
     def _save_add(self, files, git=None, git_opts=None):
         """Simple helper to add files in save()"""
+        from datalad.interface.results import get_status_dict
         # alter default behavior of git-annex by considering dotfiles
         # too
         # however, this helper is controlled by save() which itself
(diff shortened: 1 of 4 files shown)


datalad/support/annexrepo.py
```diff
@@ -2060,7 +2060,7 @@ def _run_annex_command_json(self, command,
                 log_online=True
             ))
         # TODO: refactor to account for possible --batch ones
-        annex_options = ['--json']
+        annex_options = ['--json', '--json-error-messages']
```
yarikoptic: We would need to check from which version of annex it was introduced (I still have vague memory asking for it so it might have been not too long ago) and possibly boost our minimal annex version requirement accordingly.

mih: I thought we are fine: https://git-annex.branchable.com/todo/include_msg_with_possible_reason_why_command___40__e.g._add__41___failed_into_--json_output/

mih:
I failed to come up with a test that isnt insanely complex across test setups, otherwise I would have added one. If you consider this a showstopper, please contribute.

yarikoptic:
Test is Not a show stopper but very nice to have. Grep for sudo lead me to https://github.com/datalad/datalad/blob/b950066dbcea8a8f7a90ba2fbef31cb44a73f17f/datalad/support/tests/test_annexrepo.py#L2231
I could look later when I get to the laptop which line or two to add

mih:
Test is in. Test passes. Merging....

PR #3756, opened 2019-10-07
ENH: publish - choose a remote with a url, ask interactively to clarify if multiple

yarikoptic:
I think we should use capabilities of the interactive UI more, whenever we detect running in interactive mode.  E.g. so far I have not heard much of questions and WTF statements about e.g. export-to-figshare which is highly interactive.  But users do come back inquiring about various warnings and errors we spit out when cmdline specification is "not good enough".
So in this PR I am trying to resolve a specific complaint/question about unclear error `publish` reports whenever multiple remotes/siblings could be used for publishing.

- in the case where there is no tracking remote yet specified, I switched from considering remotes with `push` spec to remotes with a `url`.  Main rationale -- remotes established by `create-sibling` have no `push` refspec.

In the long run we should even consider **all** remotes, since publish could be intended to publish to special remotes (so there would be no tracking branch established etc).

- I also listed candidate remotes in the warning message (in case of non-interactive interface)

- ui import is delayed, since otherwise not used much in the publish. But I am ok to move it out

TODOs:
- [ ] targetted unittest
- [ ] reincarnate test for inability to publish when multiple remotes present, may be test added above could do that

### DIFF
--- a/datalad/distribution/publish.py
+++ b/datalad/distribution/publish.py
@@ -472,12 +472,20 @@ def _get_remote_info(ds_path, ds_remote_info, to, missing):
         track_remote, track_refspec = ds.repo.get_tracking_branch()
         if not track_remote:
             # no tracking remote configured, but let try one more
-            # if we only have one remote, and it has a push target
-            # configured that is "good enough" for us
+            # if we only have one remote, and it has a url
+            # configured that is "good enough" for us. If multiple - we will
+            # clarify in interactive mode
             cand_remotes = [r for r in ds.repo.get_remotes()
-                            if 'remote.{}.push'.format(r) in ds.config]
+                            if 'remote.{}.url'.format(r) in ds.config]
             if len(cand_remotes) > 1:
-                lgr.warning('Target sibling ambiguous, please specific via --to')
+                from datalad.ui import ui
+                if ui.is_interactive:
+                    track_remote = _ui_ask_track_remote(ds, cand_remotes)
+                else:
+                    lgr.warning(
+                        'Target sibling ambiguous (candidates: {}), please '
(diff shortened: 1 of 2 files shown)


datalad/distribution/publish.py
```diff
@@ -472,12 +472,20 @@ def _get_remote_info(ds_path, ds_remote_info, to, missing):
...
             if len(cand_remotes) > 1:
-                lgr.warning('Target sibling ambiguous, please specific via --to')
+                from datalad.ui import ui
+                if ui.is_interactive:
+                    track_remote = _ui_ask_track_remote(ds, cand_remotes)
+                else:
+                    lgr.warning(
+                        'Target sibling ambiguous (candidates: {}), please '
+                        'specify via --to'.format(', '.join(cand_remotes))
+                    )
```
kyleam: nit: Please don't use `.format` for log messages.

For readability, I find it nicer when a list with an unknown number of items comes at the end of the message.

yarikoptic:
Thank you @kyleam! I agree on all points... will try to fix it up soonish. I think we should indeed favor the single remote with push to maintain the behavior resulting in success, but allow to choose among all with a URL (or somehow only push) defined.

yarikoptic:
I found no record of which use case lead me to this PR :-/  Re-thinking, I would say we better be strict and not try to push some random remotes with a URL... interactive aspect -- I remain of the opinion that we should use interactivity more but as @kyleam pointed out it should indeed be more clearly "prescribed".  Majority of the commands ATM indeed are assumed to be non-interactive (besides password prompts may be) with only some relying on interaction (e.g., `export2figshare`), and introducing interactivity option should indeed be "optional".
As such as I will just close this PR for now.

PR #3761, opened 2019-10-10
ENH: create: assure cfg_proc is list

adswa:
This uses ``assure_list`` to ensure that ``cfg_proc`` is a list. Else, something like this happens when using the Python API and ``cfp_proc='<proc_name>'``

```python
create(path='/tmp/t8', cfg_proc='text2git')                                                     
[INFO   ] Creating a new annex repo at /tmp/t8 
[WARNING] Cannot find procedure with name 'cfg_t' [run_procedure(cfg_t)]                                
```

### DIFF
--- a/datalad/core/local/create.py
+++ b/datalad/core/local/create.py
@@ -37,7 +37,10 @@ from datalad.support.constraints import (
     EnsureKeyChoice,
 )
 from datalad.support.param import Parameter
-from datalad.utils import getpwd
+from datalad.utils import (
+    getpwd,
+    assure_list,
+)
 
 from datalad.distribution.dataset import (
     Dataset,
@@ -205,6 +208,9 @@ class Create(Interface):
 
         # we know that we need to create a dataset at `path`
         assert(path is not None)
+
+        # assure cfg_proc is a list (relevant if used via Python API)
+        cfg_proc = assure_list(cfg_proc)
 
         # prep for yield
         res = dict(action='create', path=str(path),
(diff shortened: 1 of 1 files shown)


adswa:
Appveyor fails, but it does so because
"Build execution time has reached the maximum allowed time for your plan (60 minutes)."

datalad/core/local/create.py
```diff
@@ -206,6 +209,9 @@ def __call__(
         # we know that we need to create a dataset at `path`
         assert(path is not None)
 
+        # assure cfg_proc is a list (relevant if used via Python API)
+        cfg_proc = assure_list(cfg_proc)
```
yarikoptic: I guess in the long run we should implement generic `assure_list` for any arg which has `action='append',`

yarikoptic: although, it should not be just `assure_list` but rather allow having default `None` to discriminate between "nothing was specified" and "empty list was specified". 
Then change below removing `or []` would need to be reverted.

kyleam: Yeah, automatically handling `action='append'` arguments seems like a good idea.

If backwards-compatibility weren't a concern, I'd prefer to get rid of all the `atom -> list` argument conversion that happens when using datalad's python api.  But that ship has sailed (and I don't think others share my view).

datalad/core/local/create.py
```diff
@@ -206,6 +209,9 @@ def __call__(
         # we know that we need to create a dataset at `path`
         assert(path is not None)
 
+        # assure cfg_proc is a list (relevant if used via Python API)
```
kyleam: Oh no, the misnamed function has infected your comment :]  You're not trying to ease this list's mind or give it confidence about something.  You're forcibly casting it to a list.  Ensure!

PR #3763, opened 2019-10-10
TEMP: test devel annex

yarikoptic:
tests master with 7.20191009-1~ndall+1

PR #3762, opened 2019-10-10
TEMP (0.11.x): run build against debian-devel release of git-annex

yarikoptic:
should be the the fresh 7.20191009-1~ndall+1

PR #3765, opened 2019-10-10
BF: reflect in our test the change in behavior (fix) in annex 7.20191009

yarikoptic:
detected in #3763 and #3762 test runs

git annex changelog says
```
+  * Fix bug in handling of annex.largefiles that use largerthan/smallerthan.
+    When adding a modified file, it incorrectly used the file size of the
+    old version of the file, not the current size.
```


yarikoptic:
great, I will adjust changelog merge locally, merge into master, and push into master directly to avoid all the CI consumption.  I will also release 0.11.8 asap so we do not hold new git-annex package a hostage

Issue #3771, opened 2019-10-11
simplify access to benchmarks results (for PRs) + establish "failure" indication


yarikoptic:
# Simplify

I will make tools/ci/benchmark-travis-pr.sh accept parameters for what to do -- `run` or `compare` and split then invocation in the action workflow into two, so we could easily jump to compare

# Failure

I think that benchmarks run should fail whenever ratio exceeds pre-specified amount (default is 1.1 as used to group benchmark results).  I do not immediately see any option in `asv compare` to indicate presence of regressions via exit code, so we might just use some grep etc.

edit wishlist item for asv: https://github.com/airspeed-velocity/asv/issues/884

mih:
I think this is a reality for quite a while now.

PR #3773, opened 2019-10-11
ENH: CI benchmark runner

yarikoptic:
aims to address #3771 
- [x] splits into three dedicated actions (setup, run, compare)
- [ ] remove eval (see comment below)
- [ ] exits with non-0 in case of notable hit
- [ ] remove TEMP commit


mih:
Note that some of the benchmarks are pretty fragile ('heavy_output' etc.), and have shown significant performance loss and gains (sometimes 35%). Those should not lead to benchmark failure, otherwise it becomes noise and will be ignored like appveyor.

yarikoptic:
> Note that some of the benchmarks are pretty fragile ('heavy_output' etc.), and have shown significant performance loss and gains (sometimes 35%). Those should not lead to benchmark failure, otherwise it becomes noise and will be ignored like appveyor.

yeap, those should either be ignored (easier) or fixed up to become more robust (harder)

tools/ci/benchmark-travis-pr.sh
```diff
@@ -23,20 +26,35 @@ run_asv () {
     pip install -e .
     git show --no-patch --format="%H (%s)"
     configure_asv
-    asv run -E existing --set-commit-hash $(git rev-parse HEAD)
+    asv run -E existing --set-commit-hash "$(git rev-parse HEAD)"
```
kyleam: Quotes aren't needed in this case.  (If you grep through git's source code you'll see that they usually do not surround `$(git rev-parse HEAD)` with quotes.)  But if it eases your mind or you like the consistency, it of course doesn't hurt.

tools/ci/benchmark-travis-pr.sh
```diff
@@ -23,20 +26,35 @@ run_asv () {
...
-git checkout --force refs/bm/merge-target
-run_asv
+compare () {
+    asv compare refs/bm/merge-target refs/bm/pr
+}
 
-asv compare refs/bm/merge-target refs/bm/pr
+case "$action" in
+    '') setup && run && compare;;
+    *) eval "$action";;
```
kyleam: Please don't `eval` arbitrary input.  There are only a few valid cases, and I don't see any problem with having an arm for each one.

yarikoptic:
hm, `compare` step still runs them, although they are presumably ran in a previous step ... not sure yet WTF

tools/ci/benchmark-travis-pr.sh
yarikoptic: I succumb to pleasures of shellcheck. If there is no harm - I just obey

tools/ci/benchmark-travis-pr.sh
yarikoptic: Yeah... Better be safe than sorry, will redo with a dedicated arm for all three and the eval ;-)

tools/ci/benchmark-travis-pr.sh
kyleam: Fair enough.

yarikoptic:
eh, still think this would be worthwhile, just need to get to it some time

mih:
Closing due to inactivity.

PR #3780, opened 2019-10-14
Light RF of `clone()`

mih:
- [x] use correct helper for current relpath semantics
- [x] move internal helpers into `clone.py` (also move tests)
- [x] apply current formatting patterns


mih:
Thx @kyleam !

PR #3782, opened 2019-10-14
ENH: Do not hide intermediate results in `get()` helper

mih:
Prompted by me enjoying `datalad.log.result-level=debug`, but our tests verifying that an otherwise inaccessible result left its trace in a log message. With this change there is an actual result being propagated upward that can still yield a log message -- but more importantly can be acted upon.

- [x] RF to make processing result based, instead of automatically generated dataset instances
- [x] move control on how to deal with `clone()` errors inside `get()` to the caller (where it should be).
- [x] remove left-over internal assertion to harvest speed-up
   ```
       7.86±0.02s       6.50±0.02s    ~0.83  api.supers.time_installr
   ```


datalad/distribution/get.py
```diff
@@ -158,14 +154,20 @@ def _install_subds_from_flexible_source(
...
-        raise InstallFailedError(
-            msg="Have got no candidates to install subdataset {} from".format(
-                sm_path))
+        # yield error
+        yield get_status_dict(
+            action='install',
+            ds=ds,
+            status='error',
+            message=(
+                "Have got no candidates to install subdataset {} from.",
```
kyleam: `s/{}/%s/`

datalad/distribution/install.py
```diff
@@ -234,7 +236,7 @@ def __call__(
                     # should be necessary here, all done by code further
                     # down that deals with an install from an actuall `source`
                     # any necessary fixes should go there too!
-                    # TODO generator: possibly adjust refds
+                    r.update(dict(refds=refds_path))
```
kyleam: `r['refds'] = refds_path` would be simpler and faster:

```
% python -m timeit -s "d = {}" "d['k'] = 'v'"
10000000 loops, best of 5: 27.1 nsec per loop
% python -m timeit -s "d = {}" "d.update(dict(k='v'))"
1000000 loops, best of 5: 252 nsec per loop
```

mih: Done

datalad/distribution/get.py
mih: Arrg...Done

mih:
@kyleam ha, thx. Force-pushed an update

PR #3475, opened 2019-06-09
BF: More precise test for valid repo paths (fixes gh-3473)

mih:
Additionally check that a .git directory is not empty

- [x] fixes #3473

PR #3785, opened 2019-10-14
PERF/RF: Dataset.repo and friends

bpoldrack:
This is a long due follow up to PR #2946.

### DIFF
--- a/datalad/distribution/dataset.py
+++ b/datalad/distribution/dataset.py
@@ -263,7 +263,16 @@ class Dataset(object, metaclass=Flyweight):
             # we got a repo and path references still match
             if isinstance(self._repo, AnnexRepo):
                 # it's supposed to be an annex
-                # TODO: Checking _unique_instances might be superfluous here
+                # Note, that checking object equivalence with flyweight dict is much cheaper, than
+                # letting flyweight do the checks by calling AnnexRepo(self._repo.path). However, if we can conclude
+                # equivalence from validity it doesn't need to be there in addition.
+                # TODO: This could be just a call to self._repo._flyweight_invalid, if it wasn't for the issue with
+                #       initializing annex via parameter to AnnexRepo.__init__ (see AnnexRepo._flyweight_invalid())
+                #       That way we would do the same thing without instantiating a new one. allow_noninitialized=True
+                #       is the crucial barrier here.
+                #       Also: the call to _flyweight_invalid would be done in order to reduce code duplication.
+                #       is_valid_annex would do, but this needs to be consistent with AnnexRepo's flyweight
+                #       implementation.
                 if self._repo is AnnexRepo._unique_instances.get(
                         self._repo.path, None) and self._repo.is_valid_annex(allow_noninitialized=True):
                     # it's still the object registered as flyweight and it's a
--- a/datalad/distribution/dataset.py
+++ b/datalad/distribution/dataset.py
@@ -287,7 +287,7 @@ class Dataset(object, metaclass=Flyweight):
         # actually new instance. This is unnecessarily costly.
         valid = False
         for cls, ckw, kw in (
-                # Non-initialized is okay. We want to figure the correct instance to represent waht's there - that's it.
+                # Non-initialized is okay. We want to figure the correct instance to represent what's there - that's it.
                 (AnnexRepo, {'allow_noninitialized': True}, {'init': False}),
                 (GitRepo, {}, {})
         ):
--- a/datalad/distribution/dataset.py
+++ b/datalad/distribution/dataset.py
@@ -263,18 +263,17 @@ class Dataset(object, metaclass=Flyweight):
             # we got a repo and path references still match
             if isinstance(self._repo, AnnexRepo):
                 # it's supposed to be an annex
+                # TODO: Checking _unique_instances might be superfluous here
                 if self._repo is AnnexRepo._unique_instances.get(
-                        self._repo.path, None) and \
-                        AnnexRepo.is_valid_repo(self._repo.path,
-                                                allow_noninitialized=True):
+                        self._repo.path, None) and self._repo.is_valid_annex(allow_noninitialized=True):
                     # it's still the object registered as flyweight and it's a
                     # valid annex repo
                     return self._repo
             elif isinstance(self._repo, GitRepo):
                 # it's supposed to be a plain git
+                # TODO: Checking _unique_instances might be superfluous here
                 if self._repo is GitRepo._unique_instances.get(
-                        self._repo.path, None) and \
-                        GitRepo.is_valid_repo(self._repo.path) and not \
+                        self._repo.path, None) and self._repo.is_valid_git() and not \
                         self._repo.is_with_annex():
                     # it's still the object registered as flyweight, it's a
                     # valid git repo and it hasn't turned into an annex
@@ -288,8 +287,7 @@ class Dataset(object, metaclass=Flyweight):
         # actually new instance. This is unnecessarily costly.
         valid = False
         for cls, ckw, kw in (
-                # TODO: Do we really want to allow_noninitialized=True here?
-                # And if so, leave a proper comment!
+                # Non-initialized is okay. We want to figure the correct instance to represent waht's there - that's it.
                 (AnnexRepo, {'allow_noninitialized': True}, {'init': False}),
                 (GitRepo, {}, {})
         ):
--- a/datalad/distribution/dataset.py
+++ b/datalad/distribution/dataset.py
@@ -12,12 +12,14 @@
 import logging
 import os
 import os.path as op
-from os.path import curdir
-from os.path import exists
-from os.path import join as opj
-from os.path import normpath, isabs
-from os.path import pardir
-from os.path import realpath
+from os.path import (
+    curdir,
+    exists,
+    join as opj,
+    normpath, isabs,
+    pardir,
+    realpath,
+)
 from weakref import WeakValueDictionary
 import wrapt
 
@@ -30,21 +32,24 @@ from datalad.support.constraints import Constraint
 from datalad.support.due import due
 from datalad.support.due_utils import duecredit_dataset
 from datalad.support.exceptions import NoDatasetArgumentFound
-from datalad.support.external_versions import external_versions
-from datalad.support.gitrepo import GitRepo
-from datalad.support.gitrepo import InvalidGitRepositoryError
-from datalad.support.gitrepo import NoSuchPathError
+from datalad.support.gitrepo import (
+    GitRepo,
+    InvalidGitRepositoryError,
+    NoSuchPathError
+)
 from datalad.support.repo import Flyweight
 from datalad.support.network import RI
 from datalad.support.exceptions import InvalidAnnexRepositoryError
 
 import datalad.utils as ut
-from datalad.utils import getpwd
-from datalad.utils import optional_args
-from datalad.utils import get_dataset_root
-from datalad.utils import Path
-from datalad.utils import PurePath
-from datalad.utils import assure_list
+from datalad.utils import (
+    getpwd,
+    optional_args,
+    get_dataset_root,
+    Path,
+    PurePath,
+    assure_list,
+)
 
 
 lgr = logging.getLogger('datalad.dataset')
--- a/datalad/distribution/dataset.py
+++ b/datalad/distribution/dataset.py
@@ -148,20 +148,7 @@ class Dataset(object, metaclass=Flyweight):
           Path to the dataset location. This location may or may not exist
           yet.
         """
-        # TODO: lazy but persistent pathobj!
-        # In [16]: %timeit a = '.'
-        # 8.09 ns ± 0.0628 ns per loop (mean ± std. dev. of 7 runs, 100000000 loops each)
-        #
-        # In [17]: %timeit a = pathlib.Path('.')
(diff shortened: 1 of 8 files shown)


datalad/support/gitrepo.py
```diff
@@ -931,18 +935,48 @@ def __eq__(self, obj):
...
+
         # the aim here is to have this test as cheap as possible, because
         # it is performed a lot
         # recognize two things as good-enough indicators of a present
-        # repo: 1) a non-empty .git directory (#3473) and 2) a pointer
-        # file or symlink
-        return path.exists() and (
-            not path.is_dir() or \
-            any(path.iterdir()))
+        # repo: 1) a non-empty .git directory (#3473)
```
bpoldrack: @mih, @kyleam : Please double-check the change to `is_valid_repo`. In PR #3475 you changed it before and wrote a test for creation (the second half of which might now fail). As I test for not just "not empty", I'm not quite clear how `create` should behave in the test situation and how the validity of the repo actually changes the outcome.

kyleam: The test was just documenting the limits of the validity check.  Now that you do more, I see no issue with updating the test to account for the new behavior.

mih:
Hmm, outside of the fragile benchmarks this seems to cause slow-down, not speed-up: https://github.com/datalad/datalad/pull/3785/checks?check_run_id=259182139#step:7:227

How come?

datalad/support/gitrepo.py
bpoldrack: Okay, thanks!

bpoldrack:
@mih : Not sure yet, but first I need to check what's broken ;-)
The decisive functions themselves are faster. In any case not a lot of code is making use of cheaper testing at this state.

bpoldrack:
Re performance, @mih: Guess, it's due to wrong and different call paths. Something is broken with the detection of a repo's type and in particular we fail to initialize annexes (and even removing some?), which might cause completely different things to happen. So - no basis for comparison until the fix.

bpoldrack:
FTR: That benchmarking is ... well ... better than nothing (may be), but:

```
$ asv continuous HEAD~1 HEAD
... [snip] ...
     [90091789]       [cab057d0]
     <perf-ds-repo~1>       <perf-ds-repo>
-           -1.07            -15.6    14.58  core.runner.track_overhead_heavyout_online_process [tree/virtualenv-py3.7]
+      1.78±0.2ms       2.46±0.4ms     1.38  core.runner.time_echo [tree/virtualenv-py3.7]
+           35.94            43.92     1.22  core.runner.track_overhead_echo [tree/virtualenv-py3.7]
+             3.7             4.13     1.12  core.runner.track_overhead_100ms [tree/virtualenv-py3.7]
-          -19.94           -18.26     0.92  core.runner.track_overhead_heavyout [tree/virtualenv-py3.7]
-           21.15            -2.82    -0.13  core.runner.track_overhead_heavyout_online_through [tree/virtualenv-py3.7]

SOME BENCHMARKS HAVE CHANGED SIGNIFICANTLY.
PERFORMANCE DECREASED.
$ git diff pdiff --git a/datalad/distribution/dataset.py b/datalad/distribution/dataset.py
index 846f62ae2..4a06d7942 100644
--- a/datalad/distribution/dataset.py
+++ b/datalad/distribution/dataset.py
@@ -293,7 +293,16 @@ class Dataset(object, metaclass=Flyweight):
             # we got a repo and path references still match
             if isinstance(self._repo, AnnexRepo):
                 # it's supposed to be an annex
-                # TODO: Checking _unique_instances might be superfluous here
+                # Note, that checking object equivalence with flyweight dict is much cheaper, than
+                # letting flyweight do the checks by calling AnnexRepo(self._repo.path). However, if we can conclude
+                # equivalence from validity it doesn't need to be there in addition.
+                # TODO: This could be just a call to self._repo._flyweight_invalid, if it wasn't for the issue with
+                #       initializing annex via parameter to AnnexRepo.__init__ (see AnnexRepo._flyweight_invalid())
+                #       That way we would do the same thing without instantiating a new one. allow_noninitialized=True
+                #       is the crucial barrier here.
+                #       Also: the call to _flyweight_invalid would be done in order to reduce code duplication.
+                #       is_valid_annex would do, but this needs to be consistent with AnnexRepo's flyweight
+                #       implementation.
```

datalad/distribution/dataset.py
```diff
@@ -169,6 +178,20 @@ def __init__(self, path):
...
+        # In [17]: %timeit a = pathlib.Path('.')
+        # 1.86 µs ± 29.6 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
+
+        # Also: str from Path is faster than vice versa:
+        #  In [21]: %timeit str(path)
+        # 146 ns ± 1.33 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
+        #
+        # In [22]: %timeit pathlib.Path('.')
+        # 1.78 µs ± 19.7 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
+
```
mih: These comments related to what is happening in `Dataset.pathobj`

datalad/distribution/dataset.py
```diff
@@ -321,6 +349,8 @@ def repo(self):
             # at DEBUG level and if necessary "complaint upstairs"
             lgr.log(5, "Failed to detect a valid repo at %s", self.path)
         elif due.active:
+            # TODO: Figure out, when exactly this is needed. Don't think it makes sense to do this for every dataset,
+            #       no matter what => we want .repo to be as cheap as it gets.
             # Makes sense only on installed dataset - @never_fail'ed
             duecredit_dataset(self)
```
mih: This seems to be expensive indeed:

```
>>> ds=dl.Dataset('.')
>>> timeit duecredit_dataset(ds)                                                                                         
[WARNING] Found no aggregated metadata info file /home/mih/hacking/datalad/git/.datalad/metadata/aggregate_v1.json. You will likely need to either update the dataset from its original location or reaggregate metadata locally. 
...
98.7 ms ± 3.35 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

# for a dataset with actual metadata
>>> ds=dl.Dataset('/tmp/studyforrest-data-phase2/')
timeit duecredit_dataset(ds)                                                                                         
>>> 141 ms ± 8.48 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

However, I think this is only a concern when running "inside" duecredit and has no effect under normal operations. Is that correct @yarikoptic ?

datalad/support/annexrepo.py
```diff
@@ -180,31 +184,64 @@ def __init__(self, path, url=None, runner=None,
...
         fix_it = False
+        do_init = False
         try:
             super(AnnexRepo, self).__init__(
                 path, url, runner=runner,
                 create=create, create_sanity_checks=create_sanity_checks,
                 repo=repo, git_opts=git_opts, fake_dates=fake_dates)
         except GitCommandError as e:
+            # XXX Ben: This seems outdated, since cloning shouldn't happen in GitRepo() anymore.
+            #          Needs double-check whether this is true and whether it needs to be moved to the clone routine.
```
mih: It would be great to get this removed, if that is confirmed.

datalad/support/repo.py
```diff
@@ -88,26 +88,30 @@ def _flyweight_id_from_args(cls, *args, **kwargs):
...
+    #       ID of the requested instance
+    #
+    #     Returns
+    #     -------
+    #     bool
+    #       whether to consider an existing instance with that ID invalid and
+    #       therefore create a new instance. Default implementation always returns
+    #       False.
+    #     """
+    #     return False
```
mih: Is there value in keeping this around?

bpoldrack: Only in order to base a rephrased doc on it regarding the actual method requirement, that is the replacement for that classmethod. :-)

datalad/support/annexrepo.py
bpoldrack: I'm on it. Moved error handling to clone routine. Running tests locally ...

mih:
Appveyor failure is the usual stalling.

mih:
A recent merge has introduced a few conflicts. Can you fix the up please @bpoldrack ? Thx.

kyleam:
> A recent merge has introduced a few conflicts. Can you fix the up please @bpoldrack ? Thx.

I've pushed a rebased branch.  I thought that'd just be due to the `call_git` changes, but it turns out that there were a few other conflicts.

<details>
<summary>range-diff</summary>

```
 1:  628ee67d2 =  1:  0ed6dcc27 PERF/RF: _flyweight_invalid and is_valid_*
 2:  a4be78401 =  2:  bd00445bf BF: Use Path object correctly
 3:  3a1a40523 =  3:  39afa41ce PERF: Use the bound validity methods within Dataset.repo
 4:  23b8c4d97 !  4:  bd6f82f4b BF: Fix annex-init issue
    @@ Commit message
         Fix later, when we know this is likely the only place where AnnexRepo()
         is called like it was a direct call to AnnexRepo.__init__()
     
    - ## datalad/distribution/dataset.py ##
    -@@ datalad/distribution/dataset.py: def repo(self):
    -         # actually new instance. This is unnecessarily costly.
    -         valid = False
    -         for cls, ckw, kw in (
    --                # Non-initialized is okay. We want to figure the correct instance to represent waht's there - that's it.
    -+                # Non-initialized is okay. We want to figure the correct instance to represent what's there - that's it.
    -                 (AnnexRepo, {'allow_noninitialized': True}, {'init': False}),
    -                 (GitRepo, {}, {})
    -         ):
    -
    - ## datalad/distribution/utils.py ##
    -@@ datalad/distribution/utils.py: def _handle_possible_annex_dataset(dataset, reckless, description=None):
    + ## datalad/distribution/clone.py ##
    +@@ datalad/distribution/clone.py: def _handle_possible_annex_dataset(dataset, reckless, description=None):
          lgr.debug("Initializing annex repo at %s", dataset.path)
          # XXX this is rather convoluted, init does init, but cannot
          # set a description without `create=True`
    @@ datalad/distribution/utils.py: def _handle_possible_annex_dataset(dataset, reckl
          # so do manually see #1403
          if description:
     
    + ## datalad/distribution/dataset.py ##
    +@@ datalad/distribution/dataset.py: def repo(self):
    +         # actually new instance. This is unnecessarily costly.
    +         valid = False
    +         for cls, ckw, kw in (
    +-                # Non-initialized is okay. We want to figure the correct instance to represent waht's there - that's it.
    ++                # Non-initialized is okay. We want to figure the correct instance to represent what's there - that's it.
    +                 (AnnexRepo, {'allow_noninitialized': True}, {'init': False}),
    +                 (GitRepo, {}, {})
    +         ):
    +
      ## datalad/support/annexrepo.py ##
     @@ datalad/support/annexrepo.py: class AnnexRepo(GitRepo, RepoInterface):
          _unique_instances = WeakValueDictionary()
 5:  f1b4414b8 =  5:  33e7d1863 TST: Adapt test_create.py:test_empty_git_upstairs
 6:  db7c27a2a =  6:  939171cfd PERF/BF: Undo stupid mistake
 7:  900917895 =  7:  a7de75772 RF: Fail early in *Repo constructors
 8:  322b96d27 =  8:  464ecb281 TEMP: Add a note
 9:  fa8816562 =  9:  50de44f6e PERF: Resolve .git in GitRepo.__init__
10:  bd10677a4 = 10:  54cd07e09 PERF: Don't invalidate AnnexRepo instances whenever they aren't initialized
11:  b68d79307 = 11:  f5c9ebb38 Dataset.pathobj lazy but persistent
12:  1de32a26d = 12:  61e3074fc Remove superfluous GitRepo.realpath
13:  c51137ef4 ! 13:  2da56eb0b Cleanup imports
    @@ datalad/distribution/dataset.py
     -from os.path import normpath, isabs
     -from os.path import pardir
     -from os.path import realpath
    --from os.path import relpath
     +from os.path import (
     +    curdir,
     +    exists,
    @@ datalad/distribution/dataset.py
      
      import datalad.utils as ut
     -from datalad.utils import getpwd
    --from datalad.utils import optional_args, expandpath, is_explicit_path
    +-from datalad.utils import optional_args
     -from datalad.utils import get_dataset_root
    --from datalad.utils import dlabspath
     -from datalad.utils import Path
     -from datalad.utils import PurePath
     -from datalad.utils import assure_list
     +from datalad.utils import (
     +    getpwd,
     +    optional_args,
    -+    expandpath,
    -+    is_explicit_path,
     +    get_dataset_root,
    -+    dlabspath,
     +    Path,
     +    PurePath,
     +    assure_list,
    @@ datalad/support/gitrepo.py
      
      # shortcuts
      _curdirsep = curdir + sep
    -@@ datalad/support/gitrepo.py: def _run_command_files_split(
    -                 err += err_
    -         return out, err
    +@@ datalad/support/gitrepo.py: def call_git_success(self, args, files=None, expect_stderr=False):
    +             return False
    +         return True
      
    --
     -# TODO: --------------------------------------------------------------------
     -
          def add_remote(self, name, url, options=None):
14:  5747defb1 ! 14:  b2a0913b4 Leave a note on to-be-avoided get_git_dir
    @@ datalad/support/gitrepo.py: def get_git_dir(repo):
     +        returns an absolute path.
     +
     +
    -         Parameter
    -         ---------
    +         Parameters
    +         ----------
              repo: path or Repo instance
15:  2dc5fd9a3 = 15:  4f2ee8c66 RF: Move clone error handling to where it belongs
```

</details>

mih:
Thx @bpoldrack for the initiative and @kyleam for fixing it up. Let's put it in and continue improving from here.

Issue #3496, opened 2019-06-27
Maintain an API deprecation table


mih:
ATM we put this info in the changelog, and issue messages at runtime. I just saw that sphinx maintains a table. I found that to be very convenient for porting some code that was written for an old version of sphinx. Should we start doing that too?

http://www.sphinx-doc.org/en/master/extdev/deprecated.html

kyleam:
Sounds fine to me.

In addition to helping users upgrade, it could help us review the forgotten bits that we should be pruning.

I don't think it's much of a maintenance burden once the initial table is put together, so my main concern is that it would become stale.  But I think if we include the table link in the changelog's "Major refactoring and deprecations" section, it will serve as a reminder to update the table.

I notice that sphinx's table has a "(will be) Removed" column.  I'm not sure we should bother trying to project in which particular version we will remove something rather than just adding the version once that thing is actually removed (i.e., drop the "(will be) " part).

yarikoptic:
we might want to start using  https://deprecation.readthedocs.io/en/latest/  . Could be useful for others (iirc @tyarkoni and/or @effigies wanted similar for  pybids)

effigies:
That Sphinx table is nice. I've also started maintaining [one](https://github.com/nipy/nibabel/blob/master/nibabel/tests/test_removalschedule.py) in nibabel tests so that tests start complaining when you forget.

mih:
> we might want to start using https://deprecation.readthedocs.io/en/latest/ . Could be useful for others (iirc @tyarkoni and/or @effigies wanted similar for pybids)

That looks nice! Maybe we can get a table view out of it, too?

yarikoptic:
I have adjusted title and added `corpse-in-basement`, although it is not about being the corpse, but rather corpse preventer ;-)  I think also that we should really get this done!

yarikoptic:
I think that using a library/helpers to automate detection etc of deprecations like the above mentioned library or nibabel's https://github.com/nipy/nibabel/blob/debf173903b949f21ff57369f99388768263a7bc/nibabel/deprecator.py (I have not compared both in terms of features parity yet) would be great in addition to the table.  What is your opinion everyone?

PR #3786, opened 2019-10-14
RF: 'assure_*()' -> 'ensure_*()', fixes #3779

### LINKED ISSUE #3779: RF `assure_*()` -> `ensure_*()`
For reasons of using that language more like a few others do https://github.com/datalad/datalad/pull/3761#discussion_r333591437 ;-)

- [x] `assure_list()`
- [x] `assure_bool()`

adswa:
This refactors every helper in ``datalad/utils.py`` that is named ``assure_*`` to ``ensure_*`` :smile: 

For backward compatibility I have done this (example):

```python

def assure_list(s, copy=False, iterate=True):
    lgr.warning("assure_list() has been renamed to ensure_list().")
    return ensure_list(s, copy, iterate)

```

I won't be disappointed if you don't want this PR, feel free to disregard and  keep ``assure_*`` :smile:

Fixes #3779

### DIFF
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -34,8 +34,8 @@ from .support.protocol import (
 from .utils import (
     on_windows,
     get_tempfile_kwargs,
-    assure_unicode,
-    assure_bytes,
+    ensure_unicode,
+    ensure_bytes,
     unlink,
     auto_repr,
 )
@@ -356,13 +356,13 @@ class Runner(object):
             lgr.log(3, "Processing provided line")
         if line and log_is_callable:
             # Let it be processed
-            line = log_(assure_unicode(line))
+            line = log_(ensure_unicode(line))
             if line is not None:
                 # we are working with binary type here
-                line = assure_bytes(line)
+                line = ensure_bytes(line)
         if line:
             if out_type == 'stdout':
-                self._log_out(assure_unicode(line))
+                self._log_out(ensure_unicode(line))
             elif out_type == 'stderr':
                 self._log_err(line.decode('utf-8'),
                               expected)
@@ -689,7 +689,7 @@ class GitRunner(Runner):
             cmd, env=self.get_git_environ_adjusted(env), *args, **kwargs)
         # All communication here will be returned as unicode
         # TODO: do that instead within the super's run!
-        return assure_unicode(out), assure_unicode(err)
+        return ensure_unicode(out), ensure_unicode(err)
 
 
 def readline_rstripped(stdout):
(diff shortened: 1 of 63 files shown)


yarikoptic:
coolio. My concern is that because it touches too much code, it would make merging in 0.11.x, while we still have it, hard(er).   I would have said "let's introduce it to 0.11.x first", but then @kyleam would rightfully raise a concern that it is not a bug fix so shouldn't get there.  So let's see what his opinion would be

As for 

```python
def assure_list(s, copy=False, iterate=True):
    lgr.warning("assure_list() has been renamed to ensure_list().")
    return ensure_list(s, copy, iterate)
```

I think we better use standard warnings with DeprecationWarning mechanism here, with a comment stating that they would be completely removed e.g. in 0.13.

Related:
- https://github.com/datalad/datalad/issues/3496 -- maintain list of deprecations
- I thought we had an wishlist issue to elaborate a helper for deprecation support . something along the lines of https://nipy.org/nibabel/reference/nibabel.deprecator.html?highlight=deprecation#nibabel.deprecator.Deprecator  but I can't find it... if we don't have such an issue -- we should get one (orthogonal to this PR)

kyleam:
@adswa Thanks for the PR!

@yarikoptic:
> coolio. My concern is that because it touches too much code, it would make merging in 0.11.x, while we still have it, hard(er).

I share that concern.

> I would have said "let's introduce it to 0.11.x first", but then @kyleam would rightfully raise a concern that it is not a bug fix so shouldn't get there. So let's see what his opinion would be

Hmm, at first I agreed that using 0.11.x as the base would be easier in terms of conflicts.  I bet the initial merge of 0.11.x into master would have a lot of conflicts, but once that's out of the way you're done.

But thinking about it more, I don't know.  0.11.x isn't seeing a lot of movement, so dealing with the small conflict occasionally might be better than dealing the initial wave of conflicts you'd see if you used 0.11.x as the base.  On top of that, using 0.11.x as the base means you'd need to do a series for 0.11.x and then follow up with a master-specific series.

In either case, I must admit that, despite being the instigator of this change, I don't think dealing with the conflicts is worth making this tree-wide change _right now_.

My suggestion would be this:

  * Keep the utils.py changes that rename all the functions.
  * Add a block of assignments to utils.py that maps each of the new functions to the old name.

    Any new code is master is free to use the new names, but there is no rush to purge the `assure_` variants tree-wide.  Because we'd still be using the `assure_` variants in our own code base, don't bother yet with deprecation warnings.

   * Once we're no longer releasing from 0.11.x, tack on `DeprecationWarning`s to the `assure_` variants and make the tree-wide substitution.

If others don't think that's a good way to go though, I'm okay dealing with the conflicts.  Even if they're many in number, the resolution is straightforward.

mih:
#3909 implements a smaller step in the same direction. Closing this one.

PR #3788, opened 2019-10-14
RF: Delete some cruft and reformat

mih:
Just that


mih:
Classic appveyor stalling...

Issue #3151, opened 2019-02-08
datalad COMMAND --help does not load pager/manpage version


yarikoptic:
I guess I broke it in 107cd6748509d543561b4ea1737abf7d0b4353db intended to resolve #3118 -- with 0.11.2 debian packages, `datalad install --help` doesn't fire up the `man datalad-install`

mih:
Nope, the reason was a change in the manpage content.

PR #3798, opened 2019-10-18
BF: Adjust version extraction from manpage for proper test

mih:
Due to an earlier change the version is now located in a different position in the header.

Fixes gh-3151


mih:
This is objectively correct, merging.

Issue #1811, opened 2017-09-08
publish crashed with sequence item 1: expected string, NoneType found [publish.py:has_diff:90] (TypeError)


yarikoptic:
```shell
% datalad publish --to=datalad-public --missing=inherit -r
[INFO   ] Publishing <Dataset path=/mnt/btrfs/datasets/datalad/crawl/workshops/nipype-2017/ds000114> data to datalad-public 
[INFO   ] Actually copying 8 files 
publish(notneeded): workshops/nipype-2017/ds000114 (dataset)                                                                                            
publish(notneeded): workshops/nipype-2017 (dataset)
...  lots of notneeded entries, although I thought that by default --since last push should have been in effect!
publish(notneeded): crcns/aa-1 (dataset)
publish(notneeded): crcns (dataset)
[ERROR  ] sequence item 1: expected string, NoneType found [publish.py:has_diff:90] (TypeError) 
datalad publish --to=datalad-public --missing=inherit -r  166.99s user 55.13s system 61% cpu 6:03.21 total
```

datalad `0.8.1-210-g869f97e8` ... will retry now with --dbg to get a better glimpse (if reproducible)

yarikoptic:
here is the reason -- detached head (shouldn't happen but we shouldn't crash I guess)
```shell
publish(notneeded): crcns (dataset)
2017-09-08 13:09:04,700 [DEBUG  ] Using existing Git repository at /mnt/btrfs/datasets/datalad/crawl/corr/Resources 
2017-09-08 13:09:04,700 [DEBUG  ] detached HEAD in <AnnexRepo path=/mnt/btrfs/datasets/datalad/crawl/corr/Resources (<class 'datalad.support.annexrepo.AnnexRepo'>)> 
Traceback (most recent call last):
  File "/home/yoh/proj/datalad/datalad-master/venvs/dev/bin/datalad", line 8, in <module>
    main()
  File "/home/yoh/proj/datalad/datalad-master/datalad/cmdline/main.py", line 348, in main
    ret = cmdlineargs.func(cmdlineargs)
  File "/home/yoh/proj/datalad/datalad-master/datalad/interface/base.py", line 424, in call_from_parser
    ret = list(ret)
  File "/home/yoh/proj/datalad/datalad-master/datalad/interface/utils.py", line 382, in generator_func
    result_renderer, result_xfm, _result_filter, **_kwargs):
[...]

PR #3804, opened 2019-10-18
BF: Detect a detached HEAD that prevents publication of HEAD

mih:
Yield 'impossible' result in this case.

Fixes gh-1811

- [x] Import test from gh-1811

Now looks like this:

```
mih@meiner /tmp/clone (git)-[9995911...] % datalad publish --to origin
publish(impossible): . (dataset) [Cannot determine remote branch name from HEAD]
```


mih:
Thx @yarikoptic

PR #3794, opened 2019-10-18
BF: explicitely terminate the Popen process if exception (e.g. KeyboardInterrupt)

yarikoptic:
Otherwise it was observed (#3613) that batched parallel `git annex get` instances might still remain
running consuming bandwidth and resources.

I have added try/except around .terminate() call itself, although it might be
an overkill since in normal scenarios I do not expect that to fail.

I have tested locally by interrupting `datalad get` invocations, and it seems to do the trick!

Closes #3613

TODOs
- [ ] test  ~~well, could be done, even if only for smoke testing that code logic, but wasn't in the best traditions of the fast progress~~ I think I will not be lazy, making a test

PR #3805, opened 2019-10-19
TEMP: bogus

yarikoptic:
just to check  if the same stall happens like in https://github.com/datalad/datalad/pull/3794


yarikoptic:
So all good here, heh heh

PR #3806, opened 2019-10-19
Start documenting design patterns

mih:
This is by no means comprehensive, but it is a start. It is also no competition for CONTRIBUTING.md -- there is no overlap.

Please feel encouraged to contribute the patterns that you are following in a dedicated PR. For now I only tried to write up the two aspects that we recently discussed and reconfirmed in https://github.com/datalad/datalad/pull/3791


yarikoptic:
Great!
I will wait for posts on paths handling and results records - will come useful to me ;-)

Issue #3750, opened 2019-10-05
check_git_configured() performs inadequate check


mih:
This function is called whenever the first repo instance is created, i.e. for any commandline execution once. It invokes `git config` twice to check for `user.name` and `user.email`.

I have two issues with that.

1. it is inadequate for figuring out what it wants to see (git configured to make proper commits with meaningful IDs), because it only checks the config, but ignores all other possibilities.

```
      user.email
           Your email address to be recorded in any newly created commits. Can be
           overridden by the GIT_AUTHOR_EMAIL, GIT_COMMITTER_EMAIL, and EMAIL environment
           variables. See git-commit-tree(1).

       user.name
           Your full name to be recorded in any newly created commits. Can be overridden
           by the GIT_AUTHOR_NAME and GIT_COMMITTER_NAME environment variables. See git-
           commit-tree(1).
```

2. Paying the price of two `git config` calls before there even is a repo seems inappropriate. It may be better to perform such a check whenever the first git config is loaded into the config manager, instead of handling this in `GitRepo`. Both `Repo` and `Dataset` have `ConfigManager` instances that engage `git config anyways.

yarikoptic:
I am ok with delaying the check, but not sure how it would help 1, which is probably a separate issue (can't check ATM if git config call itself doesn't consult those env cars for sure)

PR #3807, opened 2019-10-19
BF: Move Git identity check into ConfigManager and make more valid

mih:
Also considers standard environment variables now.

Importantly, the check is no down without dedicated extra calls to git
config, saving two calls (that were pretty much always performed).

Fixes gh-3750

### DIFF
--- a/datalad/config.py
+++ b/datalad/config.py
@@ -25,6 +25,9 @@ from os.path import getmtime
 from os.path import abspath
 from time import time
 
+import logging
+lgr = logging.getLogger('datalad.config')
+
 cfg_kv_regex = re.compile(r'(^.*)\n(.*)$', flags=re.MULTILINE)
 cfg_section_regex = re.compile(r'(.*)\.[^.]+')
 cfg_sectionoption_regex = re.compile(r'(.*)\.([^.]+)')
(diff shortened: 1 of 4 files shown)


datalad/config.py
```diff
@@ -275,6 +281,19 @@ def reload(self, force=False):
...
+                    ('user.name', ('GIT_AUTHOR_NAME', 'GIT_COMMITTER_NAME')),
+                    ('user.email', ('GIT_AUTHOR_EMAIL', 'GIT_COMMITTER_EMAIL'))):
+                if cfg not in self._store \
+                        and not any(e in os.environ for e in envs):
+                    lgr.warning(
+                        "It is highly recommended to configure Git before using "
+                        "DataLad. Set both 'user.name' and 'user.email' "
+                        "configuration variables."
+                    )
+            ConfigManager._checked_git_identity = True
```
kyleam: `s/ConfigManager/self/` feels a bit more natural to me, but it's of course fine as is.

mih:
Didnt realize we have a test run that sets identity variables. Will fix up

datalad/config.py
mih: True, but I my personal preference is pointing the other way ;-)

mih:
Thx @kyleam for the review. All good now.

Issue #3810, opened 2019-10-20
Remove GitRepo._get_files_history()


mih:
Only used in a single spot in `get()` and broken on windows: https://github.com/datalad/datalad/pull/3809/checks?check_run_id=267125068#step:8:270

Ping #2970

PR #3811, opened 2019-10-20
RF: Remove obsolete GitPy-based GitRepo._get_files_history()

mih:
Fixes gh-3810


mih:
This should be uncontroversial.

PR #3812, opened 2019-10-20
RF: Remove unused and untested GitRepo.get_file_content()

mih:
This sits on top of gh-3811


mih:
OK, let's get rid of it. I cannot see any reason to keep it in.

Issue #3796, opened 2019-10-18
RF: Merge/standardize `rev_get_dataset_root()` and `get_dataset_root()`


mih:
The difference is https://github.com/datalad/datalad/pull/2944, which was initially rejected/closed but ultimately made it in through the revolution code merge.

PR #3815, opened 2019-10-20
RF: Standardize on rev_get_dataset_root() -> get_dataset_root()

mih:
A `rev_get_dataset_root` symbol remains in place to allow for peaceful transition of code that was already using it.

Fixes gh-3796


mih:
Seems I did not break anything enough to become obvious.

PR #3817, opened 2019-10-20
ENH: Record the active branch of a subdataset in the parent

mih:
This brings us closer to gh-1424

- [x] do not record the name of adjusted branches: https://github.com/datalad/datalad/pull/3817/checks?check_run_id=267367157#step:8:280
- [x] figure out why `git annex sync` blows up on windows (was https://github.com/datalad/datalad/issues/3824 and is fixed)
- [x] proposed handling involves a commit, which invalidates a bunch of test expectations.

### DIFF
--- a/datalad/core/local/tests/test_diff.py
+++ b/datalad/core/local/tests/test_diff.py
@@ -68,15 +68,23 @@ def test_repo_diff(path, norepo):
     ds = Dataset(path).create()
     assert_repo_status(ds.path)
     assert_raises(ValueError, ds.repo.diff, fr='WTF', to='MIKE')
+
+    if ds.repo.is_managed_branch():
+        fr_base = "master"
+        to = "master"
+    else:
+        fr_base = "HEAD"
+        to = None
+
     # no diff
-    eq_(ds.repo.diff('HEAD', None), {})
+    eq_(ds.repo.diff(fr_base, to), {})
     # bogus path makes no difference
-    eq_(ds.repo.diff('HEAD', None, paths=['THIS']), {})
+    eq_(ds.repo.diff(fr_base, to, paths=['THIS']), {})
     # let's introduce a known change
     create_tree(ds.path, {'new': 'empty'})
     ds.save(to_git=True)
     assert_repo_status(ds.path)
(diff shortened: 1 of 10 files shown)


mih:
Windows test failure is somewhat related to #3818

mih:
Rebased on master and gepushed.

mih:
Remaining test failures are due to the adjusted branch carrying an additional commit after an `annex sync` despite `--no-commit`. This is a problem for any test that verifies the nature of "the last commit". In particular `diff` and `run` tests are affected.

Here is what a dataset like this feels like:

image

Unsure about the best way to proceed. Given that the branch info would be useful to have, we could

- adjust all tests to make them aware of this situation
- not run these tests

kyleam:
> Given that the branch info would be useful to have, we could
> 
> * adjust all tests to make them aware of this situation (preferred)
> * not run these tests, with adjusted branches

I agree that the latter is the way to go.  I'll look at adjusting at least some of those tests today.

kyleam:
I've pushed a fix for the easier cases.  (I'll update the check list once I verify that the pass.)  I think the remaining ones start to touch on deeper issues related to our code base not being designed to work with adjusted branches.  Moreover, I view the explicit sync call in `save` as just exposing existing issues with our handling, rather than creating new issues.  It probably makes sense to mark these as known failures and deal with them when we try to approach these larger issues (e.g., gh-3818).

mih:
Thanks @kyleam for your assessment! I'll try implementing this missing pieces.

datalad/core/local/tests/test_diff.py
```diff
@@ -84,7 +84,7 @@ def test_repo_diff(path, norepo):
     create_tree(ds.path, {'new': 'empty'})
     ds.save(to_git=True)
     assert_repo_status(ds.path)
-    eq_(ds.repo.diff(fr=fr_base + '~1', to=to),
+    eq_(ds.repo.diff(fr=fr_base + '~1', to=fr_base),
```
kyleam: Sorry about that.  Thanks for fixing my mistake.

mih:
I am hopeful that this could be it!

mih:
THIS IS IT!!

Issue #3827, opened 2019-10-23
Add ability to generate submodule clone URLs based on submodule properties


mih:
ATM DataLad already supports the generation of additional candidate clone URLs besides the one hardcoded in `.gitmodules` https://github.com/datalad/datalad/blob/004daa2216c8451a6ad84eec527123c1481242c9/datalad/distribution/get.py#L80

It can be instrumental to not have to record a URL, or at least be able to overwrite a URL record without having to actually modify a dataset -- especially when dealing with large numbers of datasets. Having a configurable mechanism would also aid longevity, as data**set** availability can be made more robust.

This somewhat relates to https://github.com/datalad/datalad/issues/947 where a global dataset catalog could be queried for the location of a particular dataset that is needed as a subdataset, even if the included submodule URL is no longer valid.

As of https://github.com/datalad/datalad/commit/16dc4937595542ebb2b8d26a1220eb93e5828bba we include the DataLad dataset ID in the submodule record. Consequently, while a submodule record of a locally created dataset hierarchy can be rather useless looking like this:

```
% cat .gitmodules
[submodule "sub"]
        path = sub
        url = ./sub
        branch = master
        datalad-id = 740aa162-f3d5-11e9-9625-f0d5bf7b5561
```

The dataset ID offers the possibility for further investigation.

### Proposal

I want to enhance the candidate URL generation with a configuration-based templating mechanism that can use any known submodule property to format a URL. Let's say I have a tree of bare repos on a server that are structured based on their datalad ID. I could have:

```
datalad.get.subdataset_url_candidate = 'store.example.com:/data/{id:.3}/{id[3]}{id[4]}{id[5]}{id[6]}{id[7]}{id[8]}{id[9]}{id[10]}{id[11]}{id[12]}{id[13]}{id[14]}{id[15]}{id[16]}{id[17]}{id[18]}{id[19]}{id[20]}{id[21]}{id[22]}{id[23]}{id[24]}{id[25]}{id[26]}{id[27]}{id[28]}{id[29]}{id[30]}{id[31]}{id[32]}{id[33]}{id[34]}{id[35]}
```

which resolves to `store.example.com:/data/740/aa162-f3d5-11e9-9625-f0d5bf7b5561` when sent through `templ.format(**submod_props)` (datalad id, broken into its first three bytes and the rest; admittely obscure, but what counts is that it is possible).

The same mechanism can later be used to enable query of a particular redirector, e.g.

```
[...]

PR #3828, opened 2019-10-23
More flexible clone URL generation for submodules

mih:
This fixes gh-3827, while also tweaking a little bit on the sides -- this is rather old code.

This is technically ready. Once agreed on, there will be a use case description in the handbook on feature.

### DIFF
--- a/datalad/distribution/get.py
+++ b/datalad/distribution/get.py
@@ -112,7 +112,7 @@ def _get_flexible_source_candidates_for_submodule(ds, sm_path, sm_url=None):
     # if we have a remote, let's check the location of that remote
     # for the presence of the desired submodule
     try:
-        last_commit = ds_repo.get_last_commit_hash(sm_path)
+        last_commit = ds_repo.get_last_commit_hexsha(sm_path)
         # ideally should also give preference to the remotes which have
         # the same branch checked out I guess
         candidate_remotes += list(_get_remotes_having_commit(ds_repo, last_commit))
--- a/datalad/distribution/get.py
+++ b/datalad/distribution/get.py
@@ -128,7 +128,11 @@ def _get_flexible_source_candidates_for_submodule(ds, sm):
 
     Returns
     -------
-    list
+    list of tuples
+      Where each tuples consists of a name and a URL. Names are not unique
+      and either derived from the name of the respective remote, template
+      configuration variable, or 'origin' for the candidate URL that was
+      obtained from the .gitmodule record.
     """
     # short cuts
     ds_repo = ds.repo
@@ -174,19 +178,23 @@ def _get_flexible_source_candidates_for_submodule(ds, sm):
                 sm_path_url = sm_path
 
             clone_urls.extend(
-                _get_flexible_source_candidates(
+                (remote, url)
+                for url in _get_flexible_source_candidates(
                     # alternate suffixes are tested by `clone` anyways
-                    sm_path_url, remote_url, alternate_suffix=False))
+                    sm_path_url, remote_url, alternate_suffix=False)
+            )
 
             # attempt: provided (configured?) submodule URL
             # TODO: consider supporting DataLadRI here?  or would confuse
             #  git and we wouldn't want that (i.e. not allow pure git clone
             #  --recursive)
             if sm_url:
-                clone_urls += _get_flexible_source_candidates(
-                    sm_url,
-                    remote_url,
-                    alternate_suffix=False
+                clone_urls.extend(
+                    (remote, url)
+                    for url in _get_flexible_source_candidates(
+                        sm_url,
+                        remote_url,
+                        alternate_suffix=False)
                 )
 
         for name, tmpl in [(c[40:], ds_repo.config[c])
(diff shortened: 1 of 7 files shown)


datalad/support/gitrepo.py
```diff
@@ -1308,6 +1309,9 @@ def for_each_ref_(self, fields=('objectname', 'objecttype', 'refname'),
           descending order.
         count : int, optional
           Stop iteration after the given number of matches.
+        contains : str, optional
+          Only list refs which contain the specified commit (HEAD if not
+          specified).
```
kyleam: If not specified, it's `None`, and `--contains` isn't added.  That's not the same as restricting the results to `--contains=HEAD`.

datalad/distribution/get.py
```diff
@@ -77,36 +77,99 @@
...
+    remote_branches = [
+        b['refname:strip=2']
+        for b in repo.for_each_ref_(
+            fields='refname:strip=2',
+            pattern='refs/remotes',
+            contains=commit_hexsha)]
+    return [
+        remote
+        for remote in repo.get_remotes(with_urls_only=with_urls_only)
+        if any(rb.startswith(remote + '/') for rb in remote_branches)
```
kyleam: I'm not sure I see the point of using `get_remote(with_urls_only=...)` here, especially given this is only used in one spot.  Wouldn't `for_each_ref_`s output (with an additional layer stripped) be sufficient?

datalad/support/gitrepo.py
```diff
@@ -1577,18 +1583,18 @@ def get_hexsha(self, commitish=None, short=False):
...
         """Return the hash of the last commit the modified any of the given
         paths"""
         try:
-            stdout, stderr = self._git_custom_command(
-                files,
-                ['git', 'log', '-n', '1', '--pretty=format:%H'],
-                expect_fail=True)
-            commit = stdout.strip()
-            return commit
+            commit = self.call_git(
```
kyleam: Given that this is a call within GitRepo, I think we might as well use `self._git_custom_command` to avoid an additional function call and a couple of additional attribute lookups.

Also, should this be called with `expect_fail=True` since it has the expected unborn branch failure?

datalad/support/gitrepo.py
```diff
@@ -1577,18 +1583,18 @@ def get_hexsha(self, commitish=None, short=False):
...
+                ['rev-list', '-n1', 'HEAD'],
+                files=files,
+            )
+            commit = commit.strip()
+            return commit if commit else None
         except CommandError as e:
-            if 'does not have any commits' in e.stderr:
+            if 'unknown revision or path not in the working tree' in e.stderr:
```
kyleam: From my POV it'd be best to avoid brittle error message matching if we can.  Given we're in the error handling arm, another git call seems acceptable, so this could be replaced by `if self.get_hexsha() is None`.  (`get_hexsha()` relies on `format_commit` which itself does error message matching, but at least we'd avoid introducing another spot with a new message.)

datalad/distribution/get.py
```diff
@@ -115,41 +178,67 @@ def _get_flexible_source_candidates_for_submodule(ds, sm_path, sm_url=None):
...
+                           for c in ds_repo.config.keys()
+                           if c.startswith(
+                               'datalad.get.subdataset-source-candidate-')]:
+            url = tmpl.format(**sm_candidate_props)
+            # we don't want "flexible_source_candidates" here, this is
+            # configuration that can be made arbitrarily precise from the
+            # outside. Additional guesswork can only make it slower
+            clone_urls.append((name, url))
+
+    # CANDIDATE: the actual configured gitmodule URL
```
kyleam: I find this new wording more confusing than what it replaced.  In Git's eyes, relative paths have the following resolution: (1) relative to the URL of the tracking branch, (2) if there is not a tracking branch, relative to the URL of "origin", or (3) if there is not a configured remote, relative to the working directory.  So the new description is only accurately describing what Git considers to be the configured URL in cases that fall through to no. 3.

(As a sidenote that's not directly related to these changes, it looks like `_get_flexible_source_candidates_for_submodule` doesn't handle case no. 2, so there's a scenario where `git submodule` can clone something while `datalad install` can't.)

datalad/distribution/get.py
```diff
@@ -115,41 +178,67 @@ def _get_flexible_source_candidates_for_submodule(ds, sm_path, sm_url=None):
...
     if sm_url:
-        clone_urls += _get_flexible_source_candidates(
-            sm_url,
-            ds.path,
-            alternate_suffix=False)
+        clone_urls.extend(
+            ('origin', url)
```
kyleam: It sounds like these names aren't used for anything at this point, so it doesn't really matter, but perhaps "local" would be a better description?

datalad/support/gitrepo.py
mih: Fixed, by removing that part from the docs (copied from Git).

datalad/distribution/get.py
mih: This is just keeping what the keep did before. I tried to get rid of it, but couldn't figure out why things would break. I'll leave it for another time.

datalad/support/gitrepo.py
mih: `expect_fail` is fixed.

Re not using `call_git()`: I'd prefer to move away from the old internal helper. My goal is to see what functionality we actually need, and ultimately have runners that are optimized for what they need to do. If we keep call the all-in-one method, the cognitive complexity of the problem stays where it is -- which is too high to do anything about it. Otherwise someone would have started with it already IMHO.

datalad/support/gitrepo.py
mih: Done.

datalad/distribution/get.py
mih: Funny enough `local` is what I had. Now it is local again ;-)

datalad/distribution/get.py
kyleam: > This is just keeping what the keep did before.

Yes...

> I tried to get rid of it, but couldn't figure out why things would break.

ah, all right, so my guess that it isn't need is wrong.  Thanks.

datalad/support/gitrepo.py
kyleam: OK.  I don't know how likely it is that using `call_git` instead of `_git_custom_command` really helps in that regard, but it doesn't cost much.

mih:
@kyleam Love'ya!

PR #3831, opened 2019-10-24
RF: Move tracking branch update functionality into GitRepo.update_submodule()

mih:
This implements a long-standing TODO. It was originally part of https://github.com/datalad/datalad/pull/3828 but I moved it out, because it may need more work.

### DIFF
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -2732,6 +2732,15 @@ class GitRepo(RepoInterface, metaclass=Flyweight):
           since would result in not so annex-friendly .git symlinks/references
           instead of full featured .git/ directories in the submodules
         """
+        if GitRepo.is_valid_repo(self.pathobj / path):
+            subrepo = GitRepo(self.pathobj / path, create=False)
+            subbranch = subrepo.get_active_branch() if subrepo else None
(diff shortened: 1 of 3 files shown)


datalad/support/gitrepo.py
```diff
@@ -2760,6 +2760,43 @@ def update_submodule(self, path, mode='checkout', init=False):
...
         self._git_custom_command(path, cmd)
+
+        if not init:
+            return
+
+        # track branch originally cloned
+        branch = self.get_active_branch()
+        branch_hexsha = self.get_hexsha(branch)
+        updated_branch = self.get_active_branch()
+        if branch and not updated_branch:
```
bpoldrack: This is not plausible to me. In the original code block there was a submodule-update in between the two calls to `get_active_branch`. But here - How would `branch` and `updated_branch` ever differ from each other?

mih: Arrg, thx for spotting that. Pretty much the entire change was non-sensicle -- pushed an update now.

mih:
This is ready, I think.

bpoldrack:
Looks right to me.

mih:
Thanks for the test @kyleam!

Issue #3813, opened 2019-10-20
Remove GitRepo.get_branch_commits()


mih:
This is a GitPy-based method that is largely used is a test helper (gh-2970). Only usage outside tests is:

- `publish()`: only as a complicated with to compare the identity of two commits (i.e. simple to replace with a hexsha comparison)
- `GitRepo.get_commit_date()`: `format_commit()` (gone with gh-3814)

mih:
While not removed, the method is now a better citizen.

Issue #2970, opened 2018-11-02
GitPython use in GitRepo API


mih:
This is just a protocol of my exploration. This is no implications re -core, but might be useful in the GitPython RF context.

### GitPython users

Focusing on small helpers, not the big ones like `fetch`

Checked means gone or no longer GitPy-based at the moment:

-  [x] `get_indexed_files()` and `get_files()` (the former seems obsolete, the latter last used in metadata)
-  [x] `get_commit_date()` gh-3814
- [x] `get_active_branch()`
- [x] `get_branches()`
- [x] `get_remote_branches()`
- [x] `get_file_content()` (not used, and not tested) gh-3812
- [x] `_get_files_history()` gh-3811
- [x] `_call_gitpy_with_progress()` gh-4087
- [x] `get_branch_commits()` -- this seems to be mostly used as a test helper gh-3834
- [x] `is_dirty()`
- [x] `untracked_files()`
- [x] `get_submodules()` -- only to check if HEAD exists
- [x] `get_changed_files()` #4169 
- [x] `get_missing_files()`
- [x] `get_deleted_files()`

PR #3834, opened 2019-10-24
Non-GitPy GitRepo.get_branch_commits()

mih:
This is sitting on top of gh-3833. And aims to be another step towards #2970 by fixing #3813

Sibling PR in the crawler is https://github.com/datalad/datalad-crawler/pull/66

### DIFF
--- a/datalad/distribution/publish.py
+++ b/datalad/distribution/publish.py
@@ -330,7 +330,7 @@ def _publish_dataset(ds, remote, refspec, paths, annex_copy_options, force=False
     # changes
     if not diff and is_annex_repo:
         try:
-            git_annex_commit = next(ds.repo.get_branch_commits('git-annex'))
+            git_annex_commit = next(ds.repo.get_branch_commits_('git-annex'))
         except StopIteration:
             git_annex_commit = None
         #diff = _get_remote_diff(ds, [], git_annex_commit, remote, 'git-annex')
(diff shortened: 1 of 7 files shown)


bpoldrack:
LGTM

mih:
I prepared an adjustment PR for the crawler https://github.com/datalad/datalad-crawler/pull/66

I would merge this when the test have gone through.

datalad/tests/utils.py
```diff
@@ -1827,6 +1827,37 @@ def newfunc(*args, **kwargs):
     return newfunc
 
 
+def get_branch_commits(repo, branch=None, limit=None, stop=None):
+    """Return commit hexshas for a branch
```
yarikoptic: My main question is - why not to keep it as a method (well, rename to `get_branch_commits_`) of GitRepo and just remove `value` parameter, thus always yielding `hexsha`s?  
or we do not foresee any need for such functionality for anything but the tests?  I already see one usecase RFed above in https://github.com/datalad/datalad/pull/3834/files#diff-18445826672fc8f047d0b7d56512986cR327 for direct call to `git` (IMHO suboptimal to breed those)

mih: The idea here, and in other PRs before, is to get rid all these tiny wrapper around simple git calls and leave them to the return value focused wrappers we have now. I was under the impression that this direction had received general support. I dont see why this particular method should be treated differently.

mih: I have now implemented what you suggested @yarikoptic and adjusted the companion PR in the crawler too.

mih:
I have dropped all changes that removed this method from the class and turned into a standalone function.
Remaining changes are:

- rename to match generator-method convention
- stripping the flexibility of the method down to what is needed in core and crawler, and reimplement it with `rev-list`
- adjust usage due to those changes

mih:
As with other PRs the windows github ci failure is unrelated. Crawler failure is addressed in companion PR.

PR #3835, opened 2019-10-24
TST: Enable crawler extension tests

mih:
Unclear why they were not enabled already.

### DIFF
--- a/.github/workflows/test_extensions.yml
+++ b/.github/workflows/test_extensions.yml
@@ -1,4 +1,4 @@
-name: Extension
+name: Extensions
 
 on: [pull_request]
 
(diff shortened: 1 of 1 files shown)


kyleam:
> Unclear why they were not enabled already.

There are a few compatibility fixes that haven't yet made there way into a -crawler release, namely dcbff7a (MNT: Update save() call for compatibility with new save, 2019-07-02) and ebdcd73 (MNT: Adjust imports for compatibility with developmental DataLad, 2019-08-27).

kyleam:
Sadly it seems more recent incompatibilities have crept in: https://github.com/datalad/datalad/pull/3835/checks?check_run_id=281715869#step:9:128

yarikoptic:
ok, datalad-crawler was released, tests were restarted -- seems to be green now.  While everyone is asleep anyways, I will push a little tune up to split a step in the action into two to be able easily identify the line with the installation/version of an extension

PR #3841, opened 2019-10-29
RF: Refine what Get yields

### LINKED ISSUE #3836: get broke reporting on 'notneeded' subdatasets in 0.12.0rc6
Installing an already existing subdataset used to yield a "notneeded" result (as `get` does). With 0.12.0rc6 there's no result at all (which as a result breaks datalad-neuroimaging test helper for example dicom datasets:

```
❱ datalad create some
[INFO   ] Creating a new annex repo at /tmp/some
create(ok): /tmp/some (dataset)                                                                                              
(hirni-test-rc6) bpoldrack@brainbfast in /tmp
❱ cd some
(hirni-test-rc6) bpoldrack@brainbfast in /tmp/some on git:master
❱ datalad create -d . nested
[INFO   ] Creating a new annex repo at /tmp/some/nested
create(ok): nested (dataset)                                                                                                 
action summary:
  add (ok: 2)
  create (ok: 1)
  save (ok: 1)
(hirni-test-rc6) bpoldrack@brainbfast in /tmp/some on git:master
❱ datalad status
(hirni-test-rc6) bpoldrack@brainbfast in /tmp/some on git:master
❱ datalad -f json_pp get nested
{
  "action": "get",
  "path": "/tmp/some/nested",
  "refds": "/tmp/some",
  "status": "notneeded",
  "type": "directory"
}
(hirni-test-rc6) bpoldrack@brainbfast in /tmp/some on git:master
❱ datalad -f json_pp install nested
(hirni-test-rc6) bpoldrack@brainbfast in /tmp/some on git:master
❱
```

From the diff of `install` itself the change seems to be not in there, which suggests it's actually `get`. However, `get` does yield the correct result, so I'm not yet clear, where it goes wrong. Does that ring a bell, @mih?

bpoldrack:
Closes #3836

Not entirely sure yet, what we really want.
Top-level dataset result can only ever be `notneeded` (since otherwise we couldn't have successfully called `get` in the first place) and therefore seems superfluous. But, if we explicitly say `get .` it's somewhat strange for the results to follow a different pattern, in that the explicitly passed dataset is not reported on at all. Plus: Not sure yet, how excluding the top-level dataset would work in the recursive case.
Additional aspect: While "top-level" is quite clear in case of (interactive) CLI usage, what would make sense becomes quickly vague in the context of the python API.
Finally, from within `install` perspective might change again. It makes a lot of sense to return the (top-level) dataset, including "notneeded".
So, do I correct the tests to not expect no result for the top-level dataset or do I try to differentiate more special cases on what to yield under what circumstances? What do you think, @datalad/developers ?

### DIFF
--- a/datalad/distribution/get.py
+++ b/datalad/distribution/get.py
@@ -592,14 +592,15 @@ def _yield_dataset_result(ds_record, result, in_paths):
     # we want to yield dataset results only once (not on every input path) and
     # in addition "notneeded" results are relevant only, if those paths were
     # actually requested:
-    if result['path'] not in ds_record and \
-            (result.get('status', None) != 'notneeded' or result.get('path', None) in in_paths):
+    result_tuple = (result['path'], result.get('status', None))
+    if result_tuple not in ds_record and \
+            (result.get('status', None) != 'notneeded' or result['path'] in in_paths):
         yield result
         # record that we have a dataset-type result for that path already
         # to avoid following up with an additional directory-type result
         # via annex-get or just another notneeded dataset-type result as
         # we deal with another input path
-        ds_record.append(result['path'])
+        ds_record.append(result_tuple)
 
 
 @build_doc
@@ -793,7 +794,7 @@ class Get(Interface):
                     jobs):
 
                 if res.get('status', None) == 'notneeded' and \
-                        res.get('path', None) in yielded_datasets:
+                        res['path'] in [p for p, s in yielded_datasets]:
                     # don't yield notneeded results about paths that we already reported on.
                     continue
                 yield res
--- a/datalad/distribution/get.py
+++ b/datalad/distribution/get.py
@@ -672,10 +672,10 @@ class Get(Interface):
 
         path = assure_list(path)
         # resolve input paths against dataset in order to compare absolute paths in results
-        # to the given paths later on
-        # TODO: Not yet clear whether to use dataset or refds here. Why is dataset used in the call to Subdatasets?
+        # to the given paths later on:
         input_paths = [str(resolve_path(p, dataset)) for p in path]
         content_by_ds = {}
+        # keep a record of what datasets we reported on already:
         yielded_dataset_paths = []
         # use subdatasets() to discover any relevant content that is not
         # already present in the root dataset (refds)
--- a/datalad/distribution/get.py
+++ b/datalad/distribution/get.py
@@ -577,6 +577,31 @@ def _get_targetpaths(ds, content, refds_path, source, jobs):
         yield r
 
 
+def _update_content_dict(cont_ds, result):
+    # fish out the datasets that 'contains' a targetpath
+    # and store them for later
+    if result.get('status', None) in ('ok', 'notneeded') and \
+            'contains' in result.keys():
+        dsrec = cont_ds.get(result['path'], set())
+        dsrec.update(result['contains'])
+        cont_ds[result['path']] = dsrec
+
+
+def _yield_dataset_result(ds_record, result, in_paths):
+
+    # we want to yield dataset results only once (not on every input path) and
+    # in addition "notneeded" results are relevant only, if those paths were
+    # actually requested:
+    if result['path'] not in ds_record and \
+            (result.get('status', None) != 'notneeded' or result.get('path', None) in in_paths):
+        yield result
+        # record that we have a dataset-type result for that path already
+        # to avoid following up with an additional directory-type result
+        # via annex-get or just another notneeded dataset-type result as
+        # we deal with another input path
+        ds_record.append(result['path'])
+
+
 @build_doc
 class Get(Interface):
     """Get any dataset content (files/directories/subdatasets).
@@ -676,7 +701,7 @@ class Get(Interface):
         input_paths = [str(resolve_path(p, dataset)) for p in path]
         content_by_ds = {}
         # keep a record of what datasets we reported on already:
-        yielded_dataset_paths = []
+        yielded_datasets = []
         # use subdatasets() to discover any relevant content that is not
         # already present in the root dataset (refds)
         for sdsres in Subdatasets.__call__(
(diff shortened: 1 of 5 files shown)


bpoldrack:
@mih Guess, you might wanna have a look at this.

datalad/distribution/get.py
```diff
@@ -668,7 +670,13 @@ def __call__(
         refds = require_dataset(
             dataset, check_installed=True, purpose='get content')
 
+        path = assure_list(path)
+        # resolve input paths against dataset in order to compare absolute paths in results
+        # to the given paths later on
+        # TODO: Not yet clear whether to use dataset or refds here. Why is dataset used in the call to Subdatasets?
```
mih: It always has to be `dataset` never a dataset instance, because path arguments always have to be evaluated in the context of the given dataset (if any), not a discovered dataset.

bpoldrack: Right. That makes sense.

kyleam:
> So, do I correct the tests to not expect no result for the top-level dataset or do I try to differentiate more special cases on what to yield under what circumstances?

Adjusting the tests to accommodate a result for the top-level datatset seems like the better option to me, though I'm not confident I have a good understanding of this issue.

mih:
I am also not fully comprehending the implications. `get .` in the (empty) subdir of a subdataset mountpoint should yield a result for the top-level(?) dataset that will materialize as `.`, and it won't be `notneeded`.

bpoldrack:
>   get . in the (empty) subdir of a subdataset mountpoint should yield a result for the top-level(?) dataset that will materialize as ., and it won't be notneeded.

"top-level" is what it will operate on. In that case the superdataset, since there's nothing else yet (and this is "notneeded").

But now say you call `get .` on a dataset that's actually there in order to get the file content. My change will currently yield a notneeded result for that dataset, since `get` discovers it's not needed and it didn't report on it yet and its path is explicitly passed into the call.
However, whether or not that is intuitive and desired isn't objectively clear - so wanted to know whether anybody is aware of implications.

bpoldrack:
And now, that I wrote it - there are implications as one of the failing tests for install shows (`test_install_recursive_repeat`):

Yielding that notneeded dataset result would require to change user code, if `get` was called with `result_xfm='datasets'`, since the returned datasets don't know the status of the results they are based on. Now, in other places this is desired (like datalad-neuroimaging's `get_dicom_dataset`, which lead to the issue) to have and in places like that install test is isn't (or: wasn't).

bpoldrack:
Trying to be a bit more clear.
This is the above mentioned install test, that fails:

```
    # install again, now with data and recursive, but recursion_limit 1:
    result = get(path, dataset=path, recursive=True, recursion_limit=1,
                 result_xfm='datasets')
    # top-level dataset was not reobtained
    assert_not_in(top_ds, result)
```
`top_ds` is already installed at `path` at that point.
To me, it's plausible to yield top_ds (or the respective `notneeded` result), since we called get and passed that very path into it. However, obviously that wasn't desired before.

bpoldrack:
Adapted the two failing `install` tests. Have a look at `test_install.py` in the diff.

datalad/distribution/get.py
```diff
@@ -720,11 +728,19 @@ def __call__(
...
-                            # all those messages on not having installed anything
-                            # are a bit pointless
-                            # "notneeded" for annex get comes below
+
+                        # we want to yield dataset results only once (not on every input path) and
+                        # in addition "notneeded" results are relevant only, if those paths were
+                        # actually requested:
+                        if res['path'] not in yielded_dataset_paths and \
+                                (res.get('status', None) != 'notneeded' or res.get('path', None) in input_paths):
                             yield res
```
yarikoptic: Since I have little knowledge of possible 'status' values, not sure if my comment has value -- if there is more than a single other than 'notneeded' value is possible here -- wouldn't we then yield only one (first) `res` and just swallow another with possibly different `"status"`? does it matter? if does -- then `yielded_dataset_paths` should memoize not only paths but `(path, status)` pairs.

datalad/distribution/get.py
```diff
@@ -754,11 +770,18 @@ def __call__(
...
+                    # in addition "notneeded" results are relevant only, if those paths were
+                    # actually requested:
+                    if res['path'] not in yielded_dataset_paths and \
+                            (res.get('status', None) != 'notneeded' or res.get('path', None) in input_paths):
                         yield res
+                        # record that we have a dataset-type result for that path already
+                        # to avoid following up with an additional directory-type result
+                        # via annex-get or just another notneeded dataset-type result as
+                        # we deal with another input path
+                        yielded_dataset_paths.append(res['path'])
```
yarikoptic: oi -- dejavu... I think I saw this code before... duplication, suggests needed RF of logic or coming up with a helper so just to have `if should_dataset_be_yielded(res): yield res` which would in turn memoize what needed in that `yielded_dataset_paths` or alike within `should_dataset_be_yielded`

bpoldrack: Agree. Duplication was there before and for now just tried to figure that out while being minimally invasive. If we agree on what's the solution - this should be RF'ed, yes.

datalad/distribution/get.py
bpoldrack: Not sure, whether I really get what you mean. We would yield only the first dataset result for a particular path - yes. I don't think it really matters. However, theoretically a dataset we just installed successfully could be possibly be screwed in a way that we neither would discover it to be "notneeded" nor be able to install again (yielding "ok"), but fail with "impossible" or "error". All that within a single run of that command.
Possible - so kind of agree. But also: Not really testable. Just saying. We can't produce such a situation without destroying the dataset(s) we are operating on in a parallel job. Even if we had that, timing would be relevant.

bpoldrack:
How about that, @yarikoptic ?

bpoldrack:
Closing in favor of #3898

Issue #3759, opened 2019-10-09
Bring back ability to `install` and `get` on arbitrary path


yarikoptic:
#### What is the problem?

Originally introduced in #3746 and built on top in #3747 was RFing to remove use of AnnotatePaths.  It resulted in removing a possibility to operate on paths to be `install`ed or `get`ed, when CWD is outside of their (single) super dataset.

I don't think it is not productive for me to stop progress in those PRs by demanding retaining backward feature parity.  That is why I decided to move discussion into this separate issue.

# Use cases which were made impossible with those changes

All boil down to the original promise that, unlike git/git-annex, datalad could operate seamlessly on a path while residing outside of the dataset itself.  It provides a convenience for scripting etc.  Introduced changes make following use cases, which I used from time to time, not possible

* ```datalad -f '{path}' search -d /// haxby | xargs datalad install```
would need to become e.g. 
   - ```cd ~/datalad && datalad -f '{path}' search haxby | xargs datalad install```
* ```datalad install ///openneuro/ds000001 && datalad get ds000001/sub-01/anat```
would need to become e.g. 
  - ```datalad install ///openneuro/ds000001 && cd ds000001 && datalad get sub-01/anat```
  - ```datalad install ///openneuro/ds000001 && datalad -C ds000001 get sub-01/anat```
* pretty much for any "let me just get that file in my other collection of datasets" I would not be able to simply provide a path to it while keeping working in my project. With our new treatment of relative paths I would likely need to duplicate the path or really `cd` there (could be via `-C` through) to perform that operation
* ```datalad install ///openneuro/ds00000{1,2,3} && datalad get ds00000*/sub-01/anat```
would need to
  - establish top level super dataset first
    - ```datalad create throwaway && cd throwaway && datalad install -d . ///openneuro/ds00000{1,2,3} && datalad get ds00000*/sub-01/anat```
    - or some ugly looping

Looking at the diff I think it should be feasible to bring back some part of the annotate_paths notion (just to deduce a set of top level superdatasets given a list of paths), and thus bring back prior functionality.  I am yet to decide either it is really needed or we could get away with it.

If we are to bring it back (in a new implementation), here is a note that a test verifying this feature was removed and would be nice to bring it back: https://github.com/datalad/datalad/pull/3747/files#diff-8e9c3c169ed7a22baba947b11ad9f402L400

mih:
[...]

PR #3846, opened 2019-10-30
[WIP] ENH/BF: deduct refds from path argument

bpoldrack:
Trying to attack #3759 here.
As proposed in #3759, trying to find a reference dataset based on passed in paths if other (and cheaper) methods fail.

Tried to solve that from within `require_dataset` first, but this doesn't easily work out.

### DIFF
--- a/datalad/distribution/install.py
+++ b/datalad/distribution/install.py
@@ -27,7 +27,10 @@ from datalad.interface.results import (
     YieldDatasets,
     is_result_matching_pathsource_argument,
 )
-from datalad.interface.utils import eval_results
+from datalad.interface.utils import (
+    eval_results,
+    dataset_from_args,
+)
 from datalad.interface.base import build_doc
 from datalad.support.constraints import (
     EnsureNone,
@@ -202,19 +205,12 @@ class Install(Interface):
         # did we explicitly get a dataset to install into?
         # if we got a dataset, path will be resolved against it.
         # Otherwise path will be resolved first.
+        dataset = dataset_from_args(dataset, path)
         ds = None
         if dataset is not None:
             ds = require_dataset(dataset, check_installed=True,
                                  purpose='installation')
             common_kwargs['dataset'] = dataset
-        else:
-            from datalad.interface.utils import common_root_ds
-            parent_ds = common_root_ds(resolve_path(path))
-            lgr.debug("common_root: %s", parent_ds)
-            if parent_ds:
-                ds = require_dataset(parent_ds, check_installed=True,
-                                     purpose='installation')
-                common_kwargs['dataset'] = parent_ds
         # pre-compute for results below
         refds_path = Interface.get_refds_path(ds)
 
--- a/datalad/distribution/install.py
+++ b/datalad/distribution/install.py
@@ -207,6 +207,14 @@ class Install(Interface):
             ds = require_dataset(dataset, check_installed=True,
                                  purpose='installation')
             common_kwargs['dataset'] = dataset
+        else:
+            from datalad.interface.utils import common_root_ds
+            parent_ds = common_root_ds(resolve_path(path))
+            lgr.debug("common_root: %s", parent_ds)
+            if parent_ds:
+                ds = require_dataset(parent_ds, check_installed=True,
+                                     purpose='installation')
+                common_kwargs['dataset'] = parent_ds
         # pre-compute for results below
         refds_path = Interface.get_refds_path(ds)
 
--- a/datalad/interface/utils.py
+++ b/datalad/interface/utils.py
@@ -107,31 +107,18 @@ def handle_dirty_dataset(ds, mode, msg=None):
         raise ValueError("unknown if-dirty mode '{}'".format(mode))
 
 
-def common_root_ds(paths):
-    """Return a common root dataset for a set of absolute paths
+def dataset_from_args(dataset=None, path=None):
 
-    Note, that in opposition to get_tree_roots, this will figure a *common* root, not a set of roots.
-    In addition, this currently does actually resolve the paths rather than performing pure path operations.
-    """
-
-    from collections import Counter
-    from datalad.utils import assure_list
+    from datalad.support.path import common_prefix
+    from datalad.distribution.dataset import resolve_path
     from datalad.distribution.dataset import get_dataset_root
 
-    paths = assure_list(paths)
-    paths = [p.resolve() if p.exists() else p for p in paths]
-    counter = Counter()
-    # Note, we need the stringified paths for sorting as well as to pass the common_parent into get_dataset_root
-    counter.update([str(p) for p in paths])
-    for p in paths:
-        [counter.update(str(par)) for par in p.parents]
-    try:
-        common_parent = sorted((x for x, count in counter.items() if count >= len(paths)),
-                               key=lambda x: len(x))[-1]
-    except LookupError:
-        common_parent = None
-
-    return get_dataset_root(common_parent) if common_parent else None
+    if dataset == 'auto':
+        # deduct a common root dataset from given paths
+        # (None otherwise)
+        return get_dataset_root(str(common_prefix(resolve_path(path))))
+    else:
+        return dataset
 
 
 def get_tree_roots(paths):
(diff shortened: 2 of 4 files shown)


datalad/interface/utils.py
```diff
@@ -107,6 +107,33 @@ def handle_dirty_dataset(ds, mode, msg=None):
...
 
 
+def common_root_ds(paths):
+    """Return a common root dataset for a set of absolute paths
+
+    Note, that in opposition to get_tree_roots, this will figure a *common* root, not a set of roots.
+    In addition, this currently does actually resolve the paths rather than performing pure path operations.
+    """
+
+    from collections import Counter
```
yarikoptic: if coveralls is correct, this function is not tested at all -- could be a subject to a dedicated unittest (and a benchmark)

datalad/interface/utils.py
```diff
@@ -107,6 +107,33 @@ def handle_dirty_dataset(ds, mode, msg=None):
...
+    from datalad.utils import assure_list
+    from datalad.distribution.dataset import get_dataset_root
+
+    paths = assure_list(paths)
+    paths = [p.resolve() if p.exists() else p for p in paths]
+    counter = Counter()
+    # Note, we need the stringified paths for sorting as well as to pass the common_parent into get_dataset_root
+    counter.update([str(p) for p in paths])
+    for p in paths:
+        [counter.update(str(par)) for par in p.parents]
```
yarikoptic: this might end up quite inefficient, would be `len(paths) * len(parents)`  but what is needed to just "shrink down" common prefix, pretty much taking parents of the first path, then retaining the matching tail of that list with the next path, and break early if list gets empty (no common_parent), or then taking the first (longest) element

yarikoptic: and [get_parents](https://github.com/datalad/datalad/blob/master/datalad/support/path.py#L119) sounds like something we might be able to use it here? e.g. given paths with parents of the first path, and some post-analysis.... not yet sure, didn't try but i feel it was solving a similar problem and avoided `len(paths) * len(parents)` runtime

datalad/distribution/install.py
```diff
@@ -207,6 +207,14 @@ def __call__(
             ds = require_dataset(dataset, check_installed=True,
                                  purpose='installation')
             common_kwargs['dataset'] = dataset
+        else:
```
yarikoptic: according to codecov coloring this codepath is not tested

datalad/interface/utils.py
bpoldrack: Re 1: Yep. Thought of that, too.

Re 2: Thanks for pointing to it. Will check. And even if I can't make use of it, that common parent detection should probably move there (from current interface/utils.py)

datalad/distribution/install.py
bpoldrack: Yes. Test needs to follow. Just wasn't clear what really is the thing we want. What I pushed here is already the third iteration of my exploration yesterday ;-)

datalad/interface/utils.py
bpoldrack: However, what's probably more important re performance is, whether or not to resolve those paths in order to find a common dataset they belong to. If we don't, we'll miss out on some valid cases not recognizing they belong to the same thing. But then - to be completely correct we would actually need to discover actual repositories bottom up to account for all the fancy things you can do with symlinks + `.git`, etc.
That'd be really expensive though.

Not resolving and do plain path operation not checking FS at all would be fast and still cover a lot of cases, I think. Might be acceptable.

bpoldrack:
@datalad/developers :
Apart from some RF'ing, testing and performance, there's one question that needs to be answered.
That approach now says: If there's no `--dataset` specified, we check whether the passed paths would suggest a common dataset to operate on.
But as mentioned above: The failing test `datalad.distribution.tests.test_add.test_add_subdataset` does try to install something beneath a dataset without adding it as a subdataset. This logic relies on "if I don't specify a dataset `install` will not operate on any". This is currently indistinguishable from "I didn't specify a dataset, but I want you to discover it from the given path".
I don't see a solution to that ambiguity. Either we don't solve #3759 (or approach it very differently) or we break that use case, making it pretty much impossible to install something without adding to super.
If we deduct a dataset from given paths, we break part of the meaning of providing or not providing a `--dataset` argument.

The more I look into it, the more I think we shouldn't really "bring it back". What we can do independently is to allow `install` to just loop over several path arguments.

mih:
I agree with the assessment. I think:

- the targeted behavior should be enabled by a dedicated dataset argument. We already have `^` and this behavior could be conditioned upon a similarly special symbol. Otherwise we run into the problem described above, but we might also have recurring costs whenever a command calls another command inside, where argument preprocessing might need to be redone. ATM it looks like that if it isn't `None` nothing would need to be done. But I am not confident that this is true across all commands (we fell into that trap before, and the only things that has worked out so far, is to pass arguments down verbatim). 
- based on personal communication with @bpoldrack I agree that this implementation should eventually move into a decorator for command `__call__` methods. But this needs further investigation regarding the validity of this approach.
- I would prefer that the entire diff to `get` in this PR is a single line that call the argument preprocessing helper with a signature `fx(dataset, path)`, where both are verbatim argument values as given to the command, and the function returns the new value of `dataset`. The function should return `dataset` as is whenever it is not the special symbol that triggers the target behavior. The function should should not be inside a dedicated conditional in `get`, but any conditional handling should be done inside the function to give us a single place to alter the behavior of any command adopting this behavior.

bpoldrack:
> the targeted behavior should be enabled by a dedicated dataset argument. We already have ^ and this behavior could be conditioned upon a similarly special symbol. 

Used `'auto'` for now, but I guess any word is a bad idea, since it could easily be an actual path. However, had no idea for conflict-free symbol yet. Suggestions?

yarikoptic:
> But as mentioned above: The failing test `datalad.distribution.tests.test_add.test_add_subdataset` does try to install something beneath a dataset without adding it as a subdataset. This logic relies on "if I don't specify a dataset `install` will not operate on any". This is currently indistinguishable from "I didn't specify a dataset, but I want you to discover it from the given path".

First of all thank you @bpoldrack for this PR trying to address the #3759!  With our behavioral difference of having or not `-d` (`dataset` argument in Python) the situation is tricky indeed in how to address it "generally".  

<details>
<summary>Here is a dump of my thoughts which might or not be of some value ;)  but it is largely for myself trying to get some better understanding of the code.
</summary> 

The general pattern we have is
```
> git grep -l -e 'require_dataset(' | xargs grep -2 -e __call__ -e 'require_dataset('
...
datalad/core/local/status.py-    @datasetmethod(name='status')
datalad/core/local/status.py-    @eval_results
datalad/core/local/status.py:    def __call__(
datalad/core/local/status.py-            path=None,
datalad/core/local/status.py-            dataset=None,
--
datalad/core/local/status.py-        # come with evidence that speed does not suffer, and complexity stays
datalad/core/local/status.py-        # on a manageable level
datalad/core/local/status.py:        ds = require_dataset(
datalad/core/local/status.py-            dataset, check_installed=True, purpose='status reporting')
datalad/core/local/status.py-
```
so there is `@datasetmethod`, `@eval_results`, `require_dataset` working in tandem (often that `ds` is referred as `refds` in the code, as well as in the results records). Unfortunately logic is not uniform, many places do something like
```shell
        ds = require_dataset(
            dataset, check_installed=True, purpose='cloning') \
            if dataset is not None else dataset
```
(or alike with additional conditioning) so functions do not even resolve `ds` if dataset is not given, and then in the code having `ds` not None instructs the difference in behavior.  

My point is that if any of those two decorators would simply "resolve" dataset from the paths, it would change behavior within `__call__`.  I do not see any way to mitigate it besides some ugliness:

- The easiest "solution" could have been to just `cd` to the topmost common path for the provided paths, if they are outside of the current dataset and then invoke the operation.  **cons**: `cd`.
- in one of those decorators we do not just pass a dataset string or instance, we wrap it into some `PathDeducedDataset` wrapper, which then would be "understood" by `require_dataset` as "oh, this is the special case, I should not enable the logic for having `-d` specified".  But since in many places conditioning like `if dataset is None` is happening outside -- we would need to somehow unify that part as well, so it would also become clearer either we are "engaging" the logic of having `-d` specified or not, and returned `ds` is merely just a reference point to perform operations on.

</details>

mih:
Closing due to inactivity.

PR #3851, opened 2019-10-31
ENH: swallow exceptions in __del__ if lgr or os.fdopen already unbound

yarikoptic:
E.g. was happening during

	$> rm 100206_3T_Diffusion_unproc.json; datalad download-url s3://hcp-openaccess/HCP_1200/100206/.xdlm/100206_3T_Diffusion_unproc.json
	[INFO   ] Downloading 's3://hcp-openaccess/HCP_1200/100206/.xdlm/100206_3T_Diffusion_unproc.json' into '/tmp/testds'
	[INFO   ] S3 session: Connecting to the bucket hcp-openaccess with authentication
	download_url(ok): /tmp/testds/100206_3T_Diffusion_unproc.json (file)
	add(ok): 100206_3T_Diffusion_unproc.json (file)
	[WARNING] Registering /tmp/testds/100206_3T_Diffusion_unproc.json with s3://hcp-openaccess/HCP_1200/100206/.xdlm/100206_3T_Diffusion_unproc.json failed: AnnexBatchCommandError: command 'addurl'
	Error, annex reported failure for addurl (url='s3://hcp-openaccess/HCP_1200/100206/.xdlm/100206_3T_Diffusion_unproc.json'): {'command': 'addurl', 'success': False, 'file': '100206_3T_Diffusion_unproc.json'} [annexrepo.py:add_url_to_file:2261]
	action summary:
	  add (ok: 1)
	  download_url (ok: 1)
	  save (notneeded: 1)
	Exception ignored in: <function BatchedAnnexes.__del__ at 0x7f45502c3840>
	  File "/home/yoh/proj/datalad/datalad/datalad/support/annexrepo.py", line 3646, in __del__
	  File "/home/yoh/proj/datalad/datalad/datalad/support/annexrepo.py", line 3643, in close
	  File "/home/yoh/proj/datalad/datalad/datalad/support/annexrepo.py", line 3815, in close
	  File "/usr/lib/python3.7/logging/__init__.py", line 1370, in debug
	  File "/usr/lib/python3.7/logging/__init__.py", line 1626, in isEnabledFor
	TypeError: 'NoneType' object is not callable
	Exception ignored in: <function BatchedAnnex.__del__ at 0x7f45502c3c80>
	  File "/home/yoh/proj/datalad/datalad/datalad/support/annexrepo.py", line 3795, in __del__
	  File "/home/yoh/proj/datalad/datalad/datalad/support/annexrepo.py", line 3815, in close
	  File "/usr/lib/python3.7/logging/__init__.py", line 1370, in debug
	  File "/usr/lib/python3.7/logging/__init__.py", line 1626, in isEnabledFor
	TypeError: 'NoneType' object is not callable

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -3590,9 +3590,25 @@ class AnnexRepo(GitRepo, RepoInterface):
             out,
             re.MULTILINE | re.DOTALL | re.IGNORECASE)
 
+
+class SafeDelCloseMixin(object):
+    """A helper class to use where __del__ would call .close() which might
+    fail if "too late in GC game"
+    """
+    def __del__(self):
+        try:
+            self.close()
+        except TypeError:
+            if os.fdopen is None or lgr.debug is None:
+                # if we are late in the game and things already gc'ed in py3,
+                # it is Ok
+                return
+            raise
+
+
 # TODO: Why was this commented out?
 # @auto_repr
-class BatchedAnnexes(dict):
+class BatchedAnnexes(SafeDelCloseMixin, dict):
     """Class to contain the registry of active batch'ed instances of annex for
     a repository
     """
@@ -3642,9 +3658,6 @@ class BatchedAnnexes(dict):
         for p in self.values():
             p.close()
 
-    def __del__(self):
-        self.close()
-
 
 def readline_rstripped(stdout):
     #return iter(stdout.readline, b'').next().rstrip()
@@ -3675,7 +3688,7 @@ def readline_json(stdout):
 
 
 @auto_repr
-class BatchedAnnex(object):
+class BatchedAnnex(SafeDelCloseMixin):
     """Container for an annex process which would allow for persistent communication
     """
 
@@ -3790,9 +3803,6 @@ class BatchedAnnex(object):
             output.append(stdout)
 
         return output if input_multiple else output[0]
-
-    def __del__(self):
-        self.close()
 
     def close(self, return_stderr=False):
         """Close communication and wait for process to terminate


datalad/support/annexrepo.py
```diff
@@ -3590,9 +3590,25 @@ def _is_annex_work_tree_message(out):
...
 
+
+class SafeDelCloseMixin(object):
+    """A helper class to use where __del__ would call .close() which might
+    fail if "too late in GC game"
+    """
+    def __del__(self):
+        try:
+            self.close()
+        except TypeError as exc:
```
kyleam: nit: No need for `as exc`.

datalad/support/annexrepo.py
```diff
@@ -3590,9 +3590,25 @@ def _is_annex_work_tree_message(out):
...
+                # if we are late in the game and things already gc'ed in py3,
+                # it is Ok
+                return
+            raise
+
+
 # TODO: Why was this commented out?
 # @auto_repr
-class BatchedAnnexes(dict):
+class BatchedAnnexes(dict, SafeDelCloseMixin):
```
kyleam: It doesn't matter in this particular case, but the base class `dict` should be the right-most argument.

yarikoptic: thanks for the comments -- amended an force pushed!  Need to learn why `dict` should be the rightmost

kyleam: Oh no, hopefully on further reflection you don't disagree with my statement then :]

I'm making the claim that `SafeDelCloseMixin` should have higher precedence in the resolution chain, even if by design it is not intended define any methods that overlap with `dict`.

```py
class M():
    def blah(self):
        return "M"


class A():
    def blah(self):
        return "A"


class B0(A, M):
    pass


class B1(M, A):
    pass


print(B0.__mro__)  # (<class '__main__.B0'>, <class '__main__.A'>, <class '__main__.M'>, <class 'object'>)
print(B0().blah()) # A
print(B1.__mro__)  # (<class '__main__.B1'>, <class '__main__.M'>, <class '__main__.A'>, <class 'object'>)
print(B1().blah()) # M
```

yarikoptic: ;) so let's keep as is now ;) but I guess that is why I typically saw mixins last, since they would then get lower precedence, BUT still probably would be invoked within MRO deeper (typically we call super first while inheriting) even when multiple subclasses (re)define the same method

kyleam:
The [appveyor failure][0] was the codecov failure that we've been seeing occasionally.

[0]: https://ci.appveyor.com/project/mih/datalad/builds/28520008

PR #3863, opened 2019-11-12
BF: Perform result rendering prior error handling

mih:
Otherwise we might not be able to act on error results (depending on 'on-failure' setting), and also will not get them to a result renderer.

This is a surprisingly old problem.

### DIFF
--- a/datalad/interface/utils.py
+++ b/datalad/interface/utils.py
@@ -550,23 +550,7 @@ def _process_results(
                 res_lgr(msg, *msgargs)
             else:
                 res_lgr(msg)
-        ## error handling
-        # looks for error status, and report at the end via
-        # an exception
-        if on_failure in ('continue', 'stop') \
-                and res['status'] in ('impossible', 'error'):
-            incomplete_results.append(res)
-            if on_failure == 'stop':
-                # first fail -> that's it
-                # raise will happen after the loop
-                break
-        if result_filter:
-            try:
-                if not result_filter(res):
-                    raise ValueError('excluded by filter')
-            except ValueError as e:
-                lgr.debug('not reporting result (%s)', exc_str(e))
-                continue
+
         ## output rendering
         # TODO RF this in a simple callable that gets passed into this function
         if result_renderer is None or result_renderer == 'disabled':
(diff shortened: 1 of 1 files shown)


mih:
Test failures are unrelated.

kyleam:
> Otherwise we might not be able to act on error results (depending on 'on-failure' setting),

Could you provide a snippet or test to demonstrate a scenario this describes?  Is "act on" restricted to callers that are inspecting the rendered output?

> and also will not get them to a result renderer.

Before this change, this snippet

```
cd $(mktemp -d --tmpdir dl-XXXXXXX)
datalad create ds >/dev/null 2>&1
datalad -f json_pp --on-failure=stop status -d ds not-in-ds
```

produces

```
[ERROR  ] path not underneath this dataset [status(/tmp/dl-kiUevIC/not-in-ds)] 
```

After it produces

```
[ERROR  ] path not underneath this dataset [status(/tmp/dl-axRCRFa/not-in-ds)] 
{
  "action": "status",
  "path": "/tmp/dl-axRCRFa/not-in-ds",
  "refds": "/tmp/dl-axRCRFa/ds",
  "status": "error"
}
```
Makes sense.

mih:
@kyleam thanks for answering this question. This illustrates the core of the matter. I actually stumbled upon this issue, because I have result logging disabled locally, and ran a `create` in a dirty directory, which came back with no error message and no dataset. I also had the impression that there was (is?) a chance that it came back with exit code zero under some conditions, but I could not figure out how it happened (maybe I was mistaken).

kyleam:
> @kyleam thanks for answering this question. This illustrates the core of the matter.

Oh, I actually wasn't intending to answer my own question :)  With that snippet, I was just trying to work through the other part of the commit message, the quoted bit right above.

The main thing I wasn't sure about and was trying to clarify with the question is whether "not being able to act on" is about a caller not getting the rendered result on stdout, or whether the ordering here somehow affects Python callers working with the returned results (I don't think that's the case).

> I actually stumbled upon this issue, because I have result logging disabled locally, and ran a `create` in a dirty directory, which came back with no error message and no dataset.

Got it.

The change in this PR looks right, but I think for `create` more is needed (on top of an otherwise vanilla datalad config) to make `create` show the result because of `create`s default result filter:

```
$ git config datalad.log.result-level debug
$ datalad --on-failure=stop create  # existing repo; nothing shown
$ datalad --on-failure=stop --report-status=error create
create(error): /tmp/dlc-Zs1BWVL (dataset) [will not create a dataset in a non-empty directory, use `force` option to ignore]
```

Anyway, assuming nothing that I wrote above reveals a misunderstanding on my part, this code movement looks correct to me.

mih:
Arrgh, I missed the filter... thx! Will have a closer look.

mih:
Changes here are broken. They will actually cause dataset to not give a non-zero exit code on create error. Not yet sure how exactly.

kyleam:
> Changes here are broken. They will actually cause dataset to not give a non-zero exit code on create error. Not yet sure how exactly.

It comes down to `create`s result filter again.  Because of the result filter, `_process_results` will abort early:

https://github.com/datalad/datalad/blob/15deff69673276e68cff6462465cfab08d505d28/datalad/interface/utils.py#L553-L560

And that means it never appends to `incomplete_results`:

https://github.com/datalad/datalad/blob/15deff69673276e68cff6462465cfab08d505d28/datalad/interface/utils.py#L592

mih:
I have force-pushed on update that maintains the sequence of error handling and result filtering in this code move. This should address the remaining issue. Moreover, I have edited the commit message to no longer talk about a vague "act on":

```
Otherwise a user might not become aware of the specific nature of error
results (depending on 'on-failure' setting), because we will not get them
to a result renderer.
```

mih:
Windows test failure looks familiar and unrelated.

PR #3867, opened 2019-11-13
TST(BK): demonstrates that current git-annex-standalone fails to clone from https:// urls

yarikoptic:
So it fails with `7.20191017+git2-g7b13db551-1~ndall+1` available from neurodebian devel and passes with `7.20190819+git2-g908476a9b-1~ndall+1` available from neurodebian proper. So I expect tests to pass here and I will merge it then. TODO will be checkmarked "post-merge" if I fix it later

TODO
- [ ] provide fixed git-annex-standalone package


yarikoptic:
FWIW built a fresh snapshot (7.20191106+git38-g5899aaffd-1...) -- fails the test too.... so need to dig (hoped that issue self heals)

Issue #3869, opened 2019-11-15
Behavior with large N datasets


mih:
Demo dataset with 100k tiny files. A size like this is often considered the upper sensible limit of numbers of files.  Each file is placed in a 4th-level subdirectory, because I find that is a frequent average depth in the datasets that I work with.

```sh
$ for i in $(seq 100000); do s=$(uuid | tr '-' '/'); mkdir -p ${s:5:19} && echo $s > ${s:5}; done
# [7min, total size <4MB]
```

Create a dataset

```sh
/tmp/big % time datalad create -f
[INFO   ] Creating a new annex repo at /tmp/big 
create(ok): /tmp/big (dataset)                                                               
datalad create -f  1.03s user 0.52s system 66% cpu 2.327 total
```

```sh
/tmp/big (git)-[master] % time datalad save
Total (100000 ok out of 100000): 100%|██████████████████| 3.70M/3.70M [12:50<00:00, 12.0kB/s]
<snip 99999 lines>
add(ok): ffe/07be/11ea/ba1c/1fdd75885bdc (file)
save(ok): . (dataset)
action summary:
  add (ok: 100000)
  save (ok: 1)
datalad save  781.46s user 189.75s system 109% cpu 14:46.91 total
```

First, it works! Second, every ~30k files, my machine took a break (desktop sluggish). It seems inside `git-annex-add` numerous `git ls-files` processes (among others) are started. I am attaching a `pstree` of a moment in such a phase that has ~100 such processes running.

[ptree.txt](https://github.com/datalad/datalad/files/3852232/ptree.txt)

NB: while we are calling `git-annex-add` in batches with long lists of filenames already, `git-annex` itself further splits to-be-processed files into (smaller) batches.

The result renderer output is completely useless (100k `add(ok)` lines), nobody knows if there was an important message on top.

It took 2x longer to save the files than to create them on the file system -- not too bad IMHO.

```sh
% time datalad status
datalad status  60.22s user 13.44s system 61% cpu 2:00.10 total
% time datalad status -t raw
datalad status -t raw  6.09s user 1.45s system 100% cpu 7.487 total

% time datalad save
datalad save  8.90s user 1.56s system 99% cpu 10.485 total
```

Sensible IMHO. With a hot cache `git status` does

```
% time git status
On branch master
nothing to commit, working tree clean
git status  0.42s user 0.94s system 127% cpu 1.064 total
```

and here some truths:

```
/tmp/big (git)-[master] % find . |wc -l
506736
[...]

PR #3880, opened 2019-11-21
ENH: Speed-up initial status() run in unlock()

mih:
it should not be necessary determine the actual file type for unlock's internal need. So all those cycles can be saved.

Discovered in the context of https://github.com/datalad/datalad/issues/3869


mih:
AppVeyor has familiar time out issues. Will merge this now, the change is trivial and impactful.

Issue #3882, opened 2019-11-21
ENH: Commit message for subdataset installation?


adswa:
When writing a basic, exemplary YODA-compliant data analysis ([here](http://handbook.datalad.org/en/latest/basics/101-130-yodaproject.html)), I'm making a point of how easily comprehensible a YODA project log can be:

```
Let’s take a look at the history of the midterm_project analysis dataset:

$ git log --oneline
c73efce [DATALAD RUNCMD] analyze iris data with classification analysis
789e6b4 add script for kNN classification and plotting
9466bd3 [DATALAD] Recorded changes
ab62c74 Apply YODA dataset setup
8d3c887 [DATALAD] new dataset

“Wow, this is so clean an intuitive!” you congratulate yourself.
```

What is not easily comprehensible from the log however is ``9466bd3 [DATALAD] Recorded changes`` -- this is the installation of a subdataset.

What I would like to do is attaching a commit message when installing a subdataset:

```
datalad install -d . -s https://github.com/datalad-handbook/iris_data.git input/ -m "install input data"
```

Can you see any adverse effect this could have? I've quickly implemented this in a PR I'll submit shortly (at least locally the tests don't complain) to show what I mean codewise.

kyleam:
Sounds like a good idea to me, and in general I'm in favor of anything that enables specifying more meaningful messages.  (To really encourage/facilitate good commit messages I think we'll eventually need to support something like datalad/datalad-revolution#71.)

> Can you see any adverse effect this could have?

Nothing comes to mind.

yarikoptic:
Great to see this coming, thank you @adswa !  Larger scope/discussion for this issue is also at https://github.com/datalad/datalad/issues/3316

PR #3883, opened 2019-11-21
ENH: add message parameter to install

adswa:
This PR tries to do what I outline in #3882  and allows to provide a (optional) commit message for the installation of a subdataset.

TODO
- [ ] add tests should the general idea of this PR be sensible

### DIFF
--- a/datalad/distribution/clone.py
+++ b/datalad/distribution/clone.py
@@ -22,6 +22,7 @@ from datalad.interface.results import get_status_dict
 from datalad.interface.common_opts import (
     location_description,
     reckless_opt,
+    save_message_opt,
 )
 from datalad.support.gitrepo import (
     GitRepo,
@@ -123,6 +124,7 @@ class Clone(Interface):
             doc="""Alternative sources to be tried if a dataset cannot
             be obtained from the main `source`""",
             constraints=EnsureStr() | EnsureNone()),
+        message=save_message_opt,
     )
 
     @staticmethod
@@ -134,7 +136,8 @@ class Clone(Interface):
             dataset=None,
             description=None,
             reckless=False,
-            alt_sources=None):
+            alt_sources=None,
+            message=None):
 
         # did we explicitly get a dataset to install into?
         # if we got a dataset, path will be resolved against it.
@@ -325,6 +328,7 @@ class Clone(Interface):
             # -> make submodule
             for r in ds.save(
                     dest_path,
+                    message = message,
                     return_type='generator',
                     result_filter=None,
                     result_xfm=None,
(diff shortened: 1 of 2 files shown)


adswa:
I suspected that it would be part of a way bigger issue, and that the changes I did in here are rather tackling specific symptoms than the general issue. Your view involves concepts/functions that (to me ;-) ) are *very* complex and wide-ranging, so I'm afraid I can't really engage in a discussion/grasp the overall scope, but take #3882 and this PR as an affirmation that I'd appreciate the possibility to attach more commit messages. :) I don't want to stifle any discussion by closing this, but please feel free to close this, anyone!

mih:
@adswa Not so fast! I just expressed my view on this issue. And we have always sported a diversity of views in this project. Let's see what the conclusion will be.

yarikoptic:
> we have to make it for `create` as well.

I would say "it would be nice to" instead of "have to"

> expose --message, what about -F/--message-file?

Absence of `-F` shouldn't hold `-m` a hostage.  But ideally indeed it should be provided alongside with adding `-m` since `save` has it already.  So there should be a helper to handle those flags uniformly across commands which would absorb

```python
        if message and message_file:
            raise ValueError("Both a message and message file were specified")

        if message_file:
            with open(message_file, "rb") as mfh:
                message = assure_unicode(mfh.read())
```
from current save.

> what we are really missing is to expose Git's own behavior when messages are needed/helpful #3101

I don't think it is a panacea here.  It might be confusing to the user to get messages for sub-operations of multi-commit operations datalad might do.  
And it is IMHO largely orthogonal to being able to specify message from the command line (I use `-m` more than the editor way, which I keep for longer commit messages). 

> * we should consider changing `create` and `install` to not modify a superdataset, because it multiplexes a secondary functionality onto `-d/--dataset` which is hard to comprehend and keep in mind (cannot find the issue with that discussion, sorry)

on one hand I am somewhat agreeing that it would contribute to the purity of DataLad, but then see myself cursing you later for now requiring to run two commands, and often just duplicating the same long paths to re-specify the same `-d pathtodataset pathtosubdataset`.  It would also would be a very notable breakage of behavior and would require any user/downstream to adjust their workflows.  So, not sure if such pains are really worth the purity.
  
NB `install` still has `--nosave` which would pretty much do what you intend here, but `create` lacks it.

So, overall, I would vote to proceed forward with this one ;)

mih:
Fate and further progress depends on resolution of #3896

mih:
Conclusion of yesterday's call is to go ahead with such a change, but do it in a coordinated fashion that leads to any command that modifies a dataset to get a `--message` option, and just this one option for any message related functionality. There was a general tendency to postpone any such changes to after a now imminent 0.12 release.

yarikoptic:
I would have voted to re-open this PR:  No coordinated fashion happened, we are past 0.14.0, I wouldn't keep it as hostage of #3896 (whatever happens for `save` if does, could be adopted for `install`), and `install` was the first one I stumbled upon again just now while trying to provide a message.

PR #3885, opened 2019-11-21
Patched unicode issue for python2

soichih:
This fixes the following bug (too minor to open a github issue?)

$ cd ~/datalad/openneuro
$ datalad search -f json --show-keys short

...
bids.CogAtlasID
 in  17 datasets
 has 15 unique values: u' ', u'trm_4c8a84f20dde2', u'http://www.cognitiveatlas.org/task/id/trm_4c8a83 ....
[ERROR  ] 'ascii' codec can't encode character u'\u200b' in position 15: ordinal not in range(128) [search.py:show_keys:804] (UnicodeEncodeError)

### DIFF
--- a/datalad/metadata/search.py
+++ b/datalad/metadata/search.py
@@ -800,7 +800,7 @@ class _EGrepCSSearch(_Search):
                     "Unknown value for stats. Know full and short")
 
             print(
-                '{k}\n in  {stat.ndatasets} datasets\n has {stat.uvals_str}'.format(
+                u'{k}\n in  {stat.ndatasets} datasets\n has {stat.uvals_str}'.format(
                 k=k, stat=stat
             ))
         # After #2156 datasets may not necessarily carry all


mih:
Thanks @soichih ! The PR goes against master, but master has no support for PY2 anymore. Let's check with @yarikoptic and @kyleam how this should be treated re 0.11

yarikoptic:
@soichih which version (`datalad --version`) of datalad do you use?  if you could reproduce on top of 0.11.x branch, then it would be great if you rebase this PR against it.  `master` is indeed looking into the future and has no python2 support any longer

kyleam:
@yarikoptic 
> @soichih which version (`datalad --version`) of datalad do you use?  if you could reproduce on top of 0.11.x branch, then it would be great if you rebase this PR against it.
  
I can trigger it with py2/0.11.x.  I think these sorts of py2 issues are widespread on 0.11.x, but we might as well bring in this particular fix.  I've changed the base for this PR to 0.11.x and will merge if the tests pass.

yarikoptic:
Thank you @soichih and @kyleam

PR #3862, opened 2019-11-12
ENH: Improve readability of short help

mih:
- sort commands alphabetically (there is no point in maintaining a
  semantic structure within a single line)
- wrap to 70 chars line length to match option display
- introduce spaces between commands to improve readability

Was:
```
Usage: datalad [-l LEVEL] [--pbs-runner {condor}] [-C PATH] [--version]
               [--dbg] [--idbg] [-c KEY=VALUE]
               [-f {default,json,json_pp,tailored,'<template>'}]
               [--report-status {success,failure,ok,notneeded,impossible,error}]
               [--report-type {dataset,file}]
               [--on-failure {ignore,continue,stop}]
               [--proc-pre <PROCEDURE NAME> [ARGS ...]]
               [--proc-post <PROCEDURE NAME> [ARGS ...]] [--cmd] [-h]
               {create,install,get,add,publish,uninstall,drop,remove,update,create-sibling,create-sibling-github,create-sibling-gitlab,unlock,save,search,metadata,aggregate-metadata,extract-metadata,containers-list,containers-remove,containers-add,containers-run,crawl,crawl-init,meta-dump,meta-extract,meta-aggregate,ria-export-archive,wtf,test,ls,clean,add-archive-content,download-url,run,rerun,run-procedure,addurls,export-archive,check-dates,export-to-figshare,no-annex,add-readme,annotate-paths,clone,create-test-dataset,status,diff,siblings,sshrun,subdatasets}
               ...
```

Is:

```
Usage: datalad [-l LEVEL] [--pbs-runner {condor}] [-C PATH] [--version]
               [--dbg] [--idbg] [-c KEY=VALUE]
               [-f {default,json,json_pp,tailored,'<template>'}]
               [--report-status {success,failure,ok,notneeded,impossible,error}]
               [--report-type {dataset,file}]
               [--on-failure {ignore,continue,stop}]
               [--proc-pre <PROCEDURE NAME> [ARGS ...]]
               [--proc-post <PROCEDURE NAME> [ARGS ...]] [--cmd] [-h]
               {add, add-archive-content, add-readme, addurls, aggregate-metadata,
                annotate-paths, check-dates, clean, clone,
                containers-add, containers-list, containers-remove,
                containers-run, crawl, crawl-init, create,
                create-sibling, create-sibling-github,
                create-sibling-gitlab, create-test-dataset, diff,
                download-url, drop, export-archive,
                export-to-figshare, extract-metadata, get, install,
[...]

PR #3886, opened 2019-11-22
BF: All subcommand shorthelps crashed

mih:
Logic did not consider the situation of no subcommand list being
present in the help string.

I will merge this quickly!

Fucked up by myself in https://github.com/datalad/datalad/pull/3862


bpoldrack:
Looks correct to me.

mih:
That is what @kyleam said in https://github.com/datalad/datalad/pull/3862 ;-)

PR #3888, opened 2019-11-22
ENH: wtf() reports dataset ID

mih:
Dunno why we didn't have this since the start. But being able to
identify a dataset without having to fall back to Git (hoping that we
never move that file):

  git config --file .datalad/config datalad.dataset.id

seems like a good thing:

  datalad -f '{infos[dataset][id]}' wtf -S dataset


yarikoptic:
I might remove it later and nobody would notice :-P merging, appveyor stuck for a known reason, restarted

PR #3893, opened 2019-11-25
ENH(working Draft): with explicitly empty save message, invoke editor to provide one

yarikoptic:
Alternative to (closed without merge for now) #3101 to allow user to specify commit message.
 
In contrast to #3101, I do not think it is actually desired to try to make regular git
commit editor work, since we would not want it to pop up for every subdataset
we might be operating on in a recursive invocation.  It would also be inconsistent
with a single `-m` we provide for recursive `save` operation. With all the recursion
we better provide our own statistics and information about what is about
to happen.  In this initial commit I am not harvesting those stats, since
it would require further code RFing, and possibly some side-effect OPT to not
bother collecting and/or carrying around all the "clean" state information
about all the files in all datasets (but on a quick attempt it caused some tests expecting 'ok'
for even clean subdatasets)

To invoke editor upon save, just specify -m .  In the long run, we could
make it behave like git, i.e. upon not only an explicitly empty, but `None`
as well.

For now I have used `subprocess.call` directly without bothering with `Runner` to not inherit any possible complications. Moreover, interface to edit a file should be provided by UIs, so then some gui/web ui had a chance to provide needed interface.

### DIFF
--- a/datalad/core/local/save.py
+++ b/datalad/core/local/save.py
@@ -267,6 +267,18 @@ class Save(Interface):
                 for p, props in paths_by_ds.pop(pdspath).items()}
             start_commit = pds_repo.get_hexsha()
             if not all(p['state'] == 'clean' for p in pds_status.values()):
+                if message == '':
+                    # ask the user
+                    # TODO: RF to pre-do once the analysis of the mighty
+                    #   paths_by_ds_orig which has everything to be saved or not
+                    #  to provide useful information about which datasets
+                    #  and what changes to them (stats at least) would be saved,
+                    #  to incorporate that into the comment of the commit message
+                    #  to be edited by the user
+                    message = get_commit_message(
+                        repo=pds_repo,
+                        recursive=recursive
+                    )
                 for res in pds_repo.save_(
                         message=message,
                         # make sure to have the `path` arg be None, as we want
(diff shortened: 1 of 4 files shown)


bpoldrack:
>   I do not think it is actually desired to try to make regular git
commit editor work, since we would not want it to pop up for every subdataset
we might be operating on in a recursive invocation.

I think this could be solved by an option to pass or not pass the message to subdatasets in the recursive case (default: Nope.) We could then get the first message as committed by git from git and pass it onto the other calls.

However, this:
> Moreover, interface to edit a file should be provided by UIs, so then some gui/web ui had a chance to provide needed interface.

seems like a good point to me for being in control of the editing (and the stats).

So, overall looks like a reasonable approach to me.

yarikoptic:
> We could then get the first message as committed by git from git and pass it onto the other calls.

Comments from git about what to be committed in that editing session (will be the deepest dataset) will have no reflection of what to be committed in other (upper) datasets.

datalad/core/local/save.py
```diff
@@ -322,3 +334,40 @@ def __call__(path=None, message=None, dataset=None,
                     status='error',
                     message=('cannot tag this version: %s', e.stderr.strip()))
                 yield dsres
+
+
+def get_commit_message(repo, recursive):
+    import os
+    from ...utils import edit_text_in_file
+    msg_file = repo.pathobj / '.git' / 'COMMIT_EDITMSG'
```
yarikoptic: probably we shouldn't mess with files used by git itslf, and name this one `SAVE_EDITMSG`

yarikoptic:
ok, no strong objections were voiced, I will try to plow it forward a bit soon so it could come out from a Draft mode

mih:
6+ month of inactivity, closing.

PR #3897, opened 2019-11-28
BF: auto - if empty list is returned by get, do not try to treat it as dict record

yarikoptic:
Ran into while trying to do `python -m datalad `which heudiconv` ...` . For some reason it was querying `/.git/` and `get` would return `[]` and then kaboom.  May be will craft a test for it later


yarikoptic:
appveyor failure is I guess something unrelated, changes are minimal and obvious, so will just merge

PR #3898, opened 2019-11-29
BF: Reintroduce 'notneeded' results for existing datasets in get()

mih:
However, instead of reverting to the behavior prior 0.12rc6 (blind duplicate reporting of dataset results for subdataset install and annex-get of its root directory, this patch compromises and actually inspects if there already was a report on a dataset, and only silences follow-up duplicate reports. This impacts the result composition of large-ish operations (recursive install of a dataset with data), but AFAICS it should have now actual impact on logic and actionability of the results.

This fixes gh-3836

Alternative to https://github.com/datalad/datalad/pull/3841 @bpoldrack

### DIFF
--- a/datalad/distribution/get.py
+++ b/datalad/distribution/get.py
@@ -725,15 +725,16 @@ class Get(Interface):
                         reckless,
                         refds_path,
                         description):
+                    known_ds = res['path'] in content_by_ds
                     if res.get('status', None) in ('ok', 'notneeded') and \
                             'contains' in res:
                         dsrec = content_by_ds.get(res['path'], set())
                         dsrec.update(res['contains'])
                         content_by_ds[res['path']] = dsrec
-                    if res.get('status', None) != 'notneeded':
-                        # all those messages on not having installed anything
-                        # are a bit pointless
-                        # "notneeded" for annex get comes below
+                    # prevent double-reporting of datasets that have been
+                    # installed by explorative installation to get to target
+                    # paths, prior in this loop
+                    if res.get('status', None) != 'notneeded' or not known_ds:
                         yield res
 
         if not get_data:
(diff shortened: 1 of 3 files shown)


bpoldrack:
I'm struggling a bit to reload what I was doing in #3841, but I think one difference is where you capture `known_ds`. I did the same thing with `yielded_datasets` but outside that loop. I guess that was because `get` can get several paths pointing into to-be-installed subdatasets and therefore can discover the same subdataset several times. With not keeping the info from the processing of previous paths, that `known_ds` might lack datasets we already installed (and yielded a result for).

bpoldrack:
Otherwise, that `_update_content_dict` in my PR is simply there for reducing code duplication. It doesn't actually change anything. We just do what's in there in two spots.
Remaining difference is, that I also kept the actual status of already yielded dataset-results and TBH I'm not entirely sure anymore why I considered that necessary. I guess it also was about dealing with several paths that lead to multiple detections of the same subdatasets. I don't remember how I thought that might still lead to incorrectly not-yielded results.

mih:
> With not keeping the info from the processing of previous paths, that known_ds might lack datasets we already installed (and yielded a result for)

`content_by_ds` should provide that kind of persistent knowledge base.

mih:
@bpoldrack So what shall we do about this one?

mih:
Check fate of https://github.com/datalad/datalad/issues/3759 before merge.

bpoldrack:
@mih: We should go with your solution. The basic idea is the very same and fail to recall why exactly I ended up being more detailed in what exact results with what status I treat. Parts of it was due to a discussion with @yarikoptic , but that only arised because I already did something similar. Other than that there's only an unimportant minor RF that isn't needed to fix anything.
So - yours is leaner, solves the issue and everything else should be done only, if know why ;-)
I'll close my PR.

Issue #3818, opened 2019-10-21
Subdataset handling in v7 adjusted branch?


mih:
Consumption seems to be fine already:

Preparation on a good file system
```
/tmp % datalad create top
/tmp % datalad create -d top top/sub
/tmp % echo 123 > top/sub/file
/tmp % datalad save -d top -r
```

Consumption on a crippled FS

```
% datalad install -s /tmp/top -r crippled 
[INFO   ] Cloning /tmp/top into '/tmp/crippled' 
[INFO   ]   Detected a filesystem without fifo support. 
[INFO   ]   Disabling ssh connection caching. 
[INFO   ]   Detected a crippled filesystem. 
[INFO   ]   Entering an adjusted branch where files are unlocked as this filesystem does not support locked files. 
[INFO   ] Switched to branch 'adjusted/master(unlocked)' 
install(ok): /tmp/crippled (dataset)
[INFO   ] Installing <Dataset path=/tmp/crippled> underneath /tmp/crippled recursively 
[INFO   ] Cloning /tmp/top/sub into '/tmp/crippled/sub' 
[INFO   ]   Detected a filesystem without fifo support. 
[INFO   ]   Disabling ssh connection caching. 
[INFO   ]   Detected a crippled filesystem. 
[INFO   ]   Entering an adjusted branch where files are unlocked as this filesystem does not support locked files. 
[INFO   ] Switched to branch 'adjusted/master(unlocked)' 
[INFO   ] Submodule HEAD got detached. Resetting branch adjusted/master(unlocked) to point to 61de7f3c. Original location was 3e574196 
action summary:
  get (ok: 1)
  install (ok: 1)

/tmp % datalad status -d crippled
#clean
/tmp % git -C crippled status
On branch adjusted/master(unlocked)
nothing to commit, working tree clean

```

Modified subdataset in order to save a new subdataset state in the parent:

```
/tmp % echo 234 > crippled/sub/file2
/tmp % datalad save -d crippled -r
add(ok): file2 (file)
save(ok): sub (dataset)
add(ok): sub (file)
save(ok): . (dataset)
action summary:
  add (ok: 2)
  save (ok: 2)

/tmp % git -C crippled status
On branch adjusted/master(unlocked)
nothing to commit, working tree clean
% datalad status -d crippled -r
# clean
```

The result looks clean, but we commit the state in the adjusted branch in the parent:

```
/tmp % datalad -f json_pp status -d crippled crippled/sub | grep '"gitshasum"'
[...]

PR #3903, opened 2019-12-06
NF: Result hooks

mih:
This is aims to be a more flexible alternative to --proc-pre/post and its proposed successor of command hooks (#3264).

The key idea is that we have DataLad's results that all pass through the main event loop, and we can define ad-hoc hooks that run custom actions whenever a matching result is observed.

Key differences to what we already have:

- can act more than once per command execution
- no "pre" action anymore
- informed by the actual result itself
- runs dataset procedures, but also any proper datalad command

To define a hook, two config variables need to be set:

- `datalad.result-hook.<name>.match`
- `datalad.result-hook.<name>.proc`

where `<name>` is any Git config compatible identifier.

`match` contains a JSON-encoded dict that is used to match a result against in order to test whether the respective hook should run. It can contain any number of keys. For each key it is tested, if the value matches the one in the result, if all match the hook is executed. In addition to `==` tests, `in`, `not in`, and `!=` tests are supported. The operation can be given by wrapping the test value into a list, the first item is the operation label 'eq', 'neq', 'in', 'nin' -- the second value is the test value (set). Example:

```
{"type": ["in", ["file", "directory"]], "action": "get", "status": "notneeded"}
```

`proc` is the specification of what the hook execution comprises. Any datalad command is suitable (which includes `run_procedure`). The value is a string, where the first word is the name of the datalad command to run (in Python notation). The remainder of the string is a JSON-encoded dict with keyword arguments for the command execution. Unlike `match` string substitution is supported. Any key from a matching result can be used to trigger a substitution with the respective value in the result dict. In addition a `dsarg` key is supported that is expanded with the `dataset` argument that was giving to the command that the `eval_func` decorator belongs to and is processing the results.
Because of the string substitution using Python's `format()`, curly braces have to be protected. Hence an example setting could look like:

```
unlock {{"dataset": "{dsarg}", "path": "{path}"}}
```

or

```
run {{"cmd": "touch {path}_annoyed", "dataset": "{dsarg}", "explicit": true}}
```

Hook evaluation obviously slows processing, especially given the location in the code path (`eval_func`). The code is trying to minimize this impact. However, the lookup of potential hooks in the config represents an unconditional additional cost.

However, I consider this an extremely powerful mechanism that can be used to achieve custom setups without having to add features to the implementation of particular commands. So in summary I think this is worth the cost. 

For more info, please see the test inside.

### Benchmarks:

Our standard benchmarks show no impact (not a surprise, not much happening in them). So I ran tests that generate a lot of results (saving a dataset with 10k tiny files):

```
# setup
for i in $(seq 10000); do s=$(uuid | tr '-' '/'); mkdir -p ${s:5:19} && echo $s > ${s:5}; done;

# no hooks defined, this PR
datalad save  27.97s user 11.27s system 101% cpu 38.800 total

# no hooks defined, master 6031944e8a7770ec59389b6d7ec02d122cfbcbc3
datalad save  28.72s user 11.56s system 84% cpu 47.818 total
```

Looking forward to your feedback @datalad/developers 

TODO:

- [x] string substitution with windows paths leads to invalid JSON. No idea how to deal with that yet
- [x] investigate and potentially adjust timing of installation success results for datasets. It seems as if further processing is being perform on the dataset (content) after it is yielded (see disabled test inside). Now addressed in https://github.com/datalad/datalad/pull/3906 (merged into this PR too)

### DIFF
--- a/datalad/core/local/resulthooks.py
+++ b/datalad/core/local/resulthooks.py
@@ -36,17 +36,32 @@ def get_jsonhooks_from_config(cfg):
     for h in cfg.keys():
         if not (h.startswith('datalad.result-hook.') and h.endswith('.match-json')):
             continue
-        call = cfg.get('{}.call-json'.format(h[:-11]), None)
+        hook_basevar = h[:-11]
+        hook_name = hook_basevar[20:]
+        call = cfg.get('{}.call-json'.format(hook_basevar), None)
         if not call:
             lgr.warning(
                 'Incomplete result hook configuration %s in %s' % (
-                    h[:-11], cfg))
+                    hook_basevar, cfg))
             continue
-        sep = call.index(' ')
-        hooks[h[20:-11]] = dict(
-            cmd=call[:sep],
-            args=call[sep + 1:],
-            match=json.loads(cfg.get(h)),
+        # split command from any args
+        call = call.split(maxsplit=1)
+        # get the match specification in JSON format
+        try:
+            match = json.loads(cfg.get(h))
+        except Exception as e:
+            from datalad.dochelpers import exc_str
+            lgr.warning(
+                'Invalid match specification in %s: %s [%s], '
+                'hook will be skipped',
+                h, cfg.get(h), exc_str(e))
+            continue
+
+        hooks[hook_name] = dict(
+            cmd=call[0],
+            # support no-arg calls too
+            args=call[1] if len(call) > 1 else '{{}}',
+            match=match,
         )
     return hooks
 
@@ -169,7 +184,16 @@ def run_jsonhook(hook, spec, res, dsarg=None):
         # skip any present logger that we only carry for internal purposes
         **{k: enc(str(v)).strip('"') for k, v in res.items() if k != 'logger'})
     # now load
-    args = json.loads(args)
+    try:
+        args = json.loads(args)
+    except Exception as e:
+        from datalad.dochelpers import exc_str
+        lgr.warning(
+            'Invalid argument specification for hook %s '
+            '(after parameter substitutions): %s [%s], '
+            'hook will be skipped',
+            hook, args, exc_str(e))
+        return
     # only debug level, the hook can issue its own results and communicate
     # through them
     lgr.debug('Running hook %s: %s%s', hook, cmd_name, args)
--- a/datalad/core/local/resulthooks.py
+++ b/datalad/core/local/resulthooks.py
@@ -19,6 +19,19 @@ lgr = logging.getLogger('datalad.core.local.hooks')
 
 
 def get_hooks_from_config(cfg):
+    """Parse out hook definitions given a ConfigManager instance
+
+    Returns
+    -------
+    dict
+      where keys are hook names/labels, and each value is a dict with
+      three keys: 'cmd' contains the name of the to-be-executed DataLad
+      command; 'args' has a JSON-encoded string with a dict of keyword
+      arguments for the command (format()-language based placeholders
+      can be present; 'match' hold a JSON-encoded string representing
+      a dict with key/value pairs that need to match a result in order
+      for a hook to be triggered.
+    """
     hooks = {}
     for h in (k for k in cfg.keys()
               if k.startswith('datalad.result-hook.')
@@ -39,6 +52,37 @@ def get_hooks_from_config(cfg):
 
 
 def match_hook2result(hook, res, match):
+    """Evaluate a hook's result match definition against a concrete result
+
+    A match definition is a dict that can contain any number of keys. For each
+    key it is tested, if the value matches the one in a given result.
+    If all present key/value pairs match, the hook is executed. In addition to
+    ``==`` tests, ``in``, ``not in``, and ``!=`` tests are supported. The
+    test operation can be given by wrapping the test value into a list, the
+    first item is the operation label 'eq', 'neq', 'in', 'nin'; the second value
+    is the test value (set). Example::
+
+        {
+          "type": ["in", ["file", "directory"]],
+          "action": "get",
+          "status": "notneeded"
+        }
+
+    Parameters
+    ----------
+    hook : str
+      Name of the hook
+    res : dict
+      Result dictionary
+    match : dict
+      Match definition (see above for details).
+
+    Returns
+    -------
+    bool
+      True if the given result matches the hook's match definition, or
+      False otherwise.
+    """
     for k, v in match.items():
         # do not test 'k not in res', because we could have a match that
         # wants to make sure that a particular value is not present, and
@@ -68,7 +112,35 @@ def match_hook2result(hook, res, match):
     return True
 
 
-def run_hook(hook, spec, r, dataset_arg):
+def run_hook(hook, spec, res, dsarg=None):
+    """Execute a hook on a given result
+
+    A hook definition's 'proc' specification may contain placeholders that
+    will be expanded using matching values in the given result record. In
+    addition to keys in the result an '{dsarg}' placeholder is supported.
+    The characters '{' and '}' in the 'proc' specification that are not part
+    of format() placeholders have to be escaped as '{{' and '}}'. Example
+    'proc' specification to execute the DataLad ``unlock`` command::
+
+        unlock {{"dataset": "{dsarg}", "path": "{path}"}}
+
+    Parameters
+    ----------
+    hook : str
+      Name of the hook
+    spec : dict
+      Hook definition as returned by `get_hooks_from_config()`
+    res : dict
+      Result records that was found to match the hook definition.
+    dsarg : Dataset or str or None, optional
+      Value to substitute a {dsarg} placeholder in a hook 'proc' specification
+      with. Non-string values are automatically converted.
+
+    Yields
+    ------
+    dict
+      Any result yielded by the command executed as hook.
+    """
     import datalad.api as dl
     cmd_name = spec['cmd']
     if not hasattr(dl, cmd_name):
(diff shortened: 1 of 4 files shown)


datalad/core/local/resulthooks.py
```diff
@@ -0,0 +1,174 @@
...
+      True if the given result matches the hook's match definition, or
+      False otherwise.
+    """
+    for k, v in match.items():
+        # do not test 'k not in res', because we could have a match that
+        # wants to make sure that a particular value is not present, and
+        # not having the key would be OK in that case
+
+        # in case the target value is an actual list, an explicit action 'eq'
+        # must be given
```
bpoldrack: This needs to be mentioned in the docstring.

bpoldrack:
Generally looks very nice indeed.

I'm struggling a bit ATM to get my head around security implications and what part of it might need addressing where. In principle we have a similar issue with procedures anyway, since you can "catch them" by updating a dataset. And we can't completely avoid the issue without loosing substantial functionality.
However, if we allow code execution via a committed config, we kinda make it worse. So, this is not an objection to that PR and approach (I do think we need that), but I think this raises level of we-need-to-address-this. At least we should give a defense to users, like ignoring `.datalad/config` wrt to procedures and hooks.

mih:
We already allow this kind of execution based on config with proc-pre/post for individual commands. So in that respect this PR doesnt add a new threat.

bpoldrack:
Yes, agree. Just raises the priority we need to think about this with.

datalad/core/local/resulthooks.py
mih: True

mih:
Maybe we should afford a new ConfigManager instance that is not reading out `.datalad/config` within `eval_func`.

However, I would do that in a separate PR, as it affects any and all types of hooks.

mih:
I proposed #3907 to address the security concerns.

datalad/core/local/resulthooks.py
mih: Done

mih:
Cool, thanks. When #3907 is merged, I'll fix the resulting conflicts with this one, and proceed to merge, once the tests have passed again.

kyleam:
> Unless @kyleam comes to a different conclusion ;-)

If you'd like to wait, I'll find time to review this today, but it's of course fine if you want to proceed with the merge.

Without having looked at any detail into this, I'll just say that I'm happy that @bpoldrack brought up the security concerns, and that that's already led to a PR.  I should have thought of those concerns when `datalad.<name>.proc-{pre,post}` stuff was introduced.  Even worse, I was thinking about git hooks/config and security somewhat recently when reading [this git thread][0] and it did not occur to me that---given that untracked .git/{config,hooks} raise security concerns---tracked values of `.proc-{pre,post}` (and probably other stuff related to procedures) should be re-evaluated carefully.

[0]: https://lore.kernel.org/git/20191116011125.GG22855@google.com/T/#u

mih:
I'll be more then happy to wait. Thanks!

mih:
Working on resolving the conflict...

datalad/core/local/resulthooks.py
```diff
@@ -0,0 +1,177 @@
...
+"""Interface utility functions
+
+"""
+
+__docformat__ = 'restructuredtext'
+
+import logging
+import json
+
+lgr = logging.getLogger('datalad.core.local.hooks')
```
kyleam: Quickly grepping this code base, I wasn't able to spot anywhere else that the logger doesn't follow the module name, so `s/hooks/result&/`?

datalad/core/local/resulthooks.py
```diff
@@ -0,0 +1,177 @@
...
+      command; 'args' has a JSON-encoded string with a dict of keyword
+      arguments for the command (format()-language based placeholders
+      can be present); 'match' holds a JSON-encoded string representing
+      a dict with key/value pairs that need to match a result in order
+      for a hook to be triggered.
+    """
+    hooks = {}
+    for h in (k for k in cfg.keys()
+              if k.startswith('datalad.result-hook.')
+              and k.endswith('.match')):
```
kyleam: A matter of taste, but I stumbled reading this more than think I would have if instead it were structured as `for h in cfg.keys()` with the condition from the list comprehension moved to the body.

datalad/core/local/resulthooks.py
```diff
@@ -0,0 +1,177 @@
...
+        proc = cfg.get('{}.proc'.format(h[:-6]), None)
```
kyleam: Hmm, so `proc` is for "procedure".  While that could be a fairly general term, we already have procedures, so I think that's a bit confusing here given it's about calling any dataset method, not just `run_procedure`.  How about `s/.proc/.call/` ?

datalad/core/local/resulthooks.py
```diff
@@ -0,0 +1,177 @@
...
+        if not proc:
+            lgr.warning(
+                'Incomplete result hook configuration %s in %s' % (
+                    h[:-6], cfg))
+            continue
+        sep = proc.index(' ')
```
kyleam: Would there be a downside here of being more lenient and splitting on the first whitespace (`proc.split(maxsplit=1)`)?  Solely from a readability point, I'd find it nice to get rid of some of the indexing below.

---

Given that the `proc` and `match` values are coming from the user, it be good to show a bit more helpful message if they aren't valid.  For example, the `.index()` call above would fail with

    [ERROR  ] substring not found [resulthooks.py:get_hooks_from_config:45] (ValueError) 

if the value didn't have a space.  The `json.loads()` is another case (and perhaps more likely when users are trying to put valid json into a git config value). And misformatted arguments would make it through to `run_hook` before causing issues, so that's yet another spot.

Given that these hooks are secondary to the main command, perhaps these failures should give a warning and continue on with the rest of result processing.

datalad/core/local/resulthooks.py
mih: Forgotten leftover, will fix. Thx!

datalad/core/local/resulthooks.py
mih: Will do, but go one step further. See below...

mih:
> I think this is a clever approach. In addition to being flexible, the implementation is pleasing in terms of being a mostly independent layer on top of the current result handling.

Thx!
 
> On the other hand, it doesn't seem particularly pleasant to work with from a user-perspective. Beyond the somewhat tedious task of formatting the `json` values for `match` and `proc`, this would require a decent investment from the user to think about how they should match results. It also makes me wonder how consistent or useful of a surface our result records provide for latching onto. And makes me worried that hooks will start matching very specific things about the results, and we're going to fear causing breakage even when changing minor details about the results.

I acknowledge all these concerns. But I also see the increased focus on result composition and placement as a chance to treat them more rigorously and thoughtfully, e.g. https://github.com/datalad/datalad/pull/3906

Re usability: I simply could not yet come up with something more convenient. But what about tweaking things right away to enable better forward compatibility:

```
.match -> .match-json
.proc -> call-json (taking above comment into account)
```

This gives us the freedom to implement support for different/simpler approaches in the future. I guess it is reasonable to assume that any approach would need to specify a criterion and a call specification.

> At any rate, given that this is non-intrusive and seems like it could be quite powerful, I'm for going forward with it and seeing how it plays.

Cool, thx!

Perhaps a word on the concrete underlying motivation to implement this now, and like this. The use case is the installation of subdatasets and file content getting in a YODA dataset scenario. Say we install a YODA dataset as a throwaway clone in an HPC environment. We do not care about data safety much, but we do care about performance. Hence we want to turn on as much performance tuning as possible (reckless-install,...). Additionally, we cannot have symlinks for files, because the particular analysis code cannot handle them. At the same time, there are real-world constraints that we need to deal with (unlock blows storage demands up, annex.thin needs v7, v7 adjusted branches have conceptual issues for YODA https://github.com/datalad/datalad/issues/3818).

With this feature implemented, I can configure the root dataset clone to arbitrarily tune any newly installed subdataset however I see fit, independent of the particular analysis. I can also make `get` unlock any file that is actually being obtained, as opposed to all files upfront, or putting the entire (sub)dataset into unlocked mode. And I can do all of that without actually altering any committed content, hence I am not overfitting the datasets themselves to a particular purpose or environment.

datalad/core/local/resulthooks.py
mih: Done.

datalad/core/local/resulthooks.py
mih: Motivation was speed. However, the two approaches are practically indistinguishable for any reasonable size of config, and differ only marginally for huge dicts. Will simplify to better match the human condition ;-)

datalad/core/local/resulthooks.py
mih: > Would there be a downside here of being more lenient and splitting on the first whitespace (`proc.split(maxsplit=1)`)? Solely from a readability point, I'd find it nice to get rid of some of the indexing below.

Oh yes! Thanks for reminding me that we are in PY3 land now!! Done.

I also added the ability to have argument-less call specifications, and a smoke test for them.

> Given that the `proc` and `match` values are coming from the user, it be good to show a bit more helpful message if they aren't valid. For example, the `.index()` call above would fail with
> 
> ```
> [ERROR  ] substring not found [resulthooks.py:get_hooks_from_config:45] (ValueError) 
> ```
> 
> if the value didn't have a space. The `json.loads()` is another case (and perhaps more likely when users are trying to put valid json into a git config value). And misformatted arguments would make it through to `run_hook` before causing issues, so that's yet another spot.
> 
> Given that these hooks are secondary to the main command, perhaps these failures should give a warning and continue on with the rest of result processing.

A broken match spec would now yield such warning:

```
WARNING: Invalid match specification in datalad.result-hook.annoy.match-json: {"type":["in", ["file"]],"action":"get",status":"notneeded"} [Expecting property name enclosed in double quotes: line 1 column 41 (char 40) [decoder.py:raw_decode:353]], hook will be skipped
```

Likewise, a broken call spec given something like this:

```
WARNING: Invalid argument specification for hook annoy (after parameter substitutions): {"cmd":"touch" /tmp/datalad_temp_test_basicsnscytin3/file1_annoyed","dataset":"/tmp/datalad_temp_test_basicsnscytin3","explicit":true} [Expecting ',' delimiter: line 1 column 16 (char 15) [decoder.py:raw_decode:353]], hook will be skipped
```

mih:
OK, I'll be bold and merged this in a few min.

adswa:
Thanks for the comprehensive comments in this PR! I am adding a handbook section on this new feature [here](https://github.com/datalad-handbook/book/pull/304).

PR #3906, opened 2019-12-07
BF: Make sure to not issue an install-ok result, before the repo state is final

mih:
The underlying issue was discovered in https://github.com/datalad/datalad/pull/3903


mih:
Rebased on master after merge of #3904. Could not replicate any of the windows test failures on a win10 box with Git 2.23.

mih:
Cool, thanks!

PR #3907, opened 2019-12-09
ConfigManager can be limited to local config; prevent remote changes of hooks

mih:
The issue was raised in https://github.com/datalad/datalad/pull/3903#issuecomment-562979754 but is in fact a long-standing issue in DataLad. Hence this comes as a separate PR, and https://github.com/datalad/datalad/pull/3903 will need to be adjusted, once this is merged.

We may want to have an override that re-enables this feature -- but I think we should wait if an actual use case materializes. Otherwise not supporting it seems like the better way.

### DIFF
--- a/datalad/config.py
+++ b/datalad/config.py
@@ -159,23 +159,29 @@ class ConfigManager(object):
       If provided, all `git config` calls are executed in this dataset's
       directory. Moreover, any modifications are, by default, directed to
       this dataset's configuration file (which will be created on demand)
-    dataset_only : bool
-      If True, configuration items are only read from a datasets persistent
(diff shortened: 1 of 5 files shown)


bpoldrack:
>   We may want to have an override that re-enables this feature -- but I think we should wait if an actual use case materializes. Otherwise not supporting it seems like the better way.

Agree. We also may want to have a more fine-grained restriction that cares about execution related configs only. But again: Having this as is in this PR should give us a basis to discover what else we might need to consider. Let's have that and see what emerges from that.

mih:
Thanks I will merge this, once travis is done.

mih:
Perfect, thanks to both of you!

Issue #2849, opened 2018-09-19
`download-url` should record the download url in annex


mih:
```
% datalad create this
% cd this
% datalad download-url https://cdnjs.cloudflare.com/ajax/libs/open-iconic/1.1.1/font/css/open-iconic-bootstrap.min.css
% git annex whereis open-iconic-bootstrap.min.css
whereis open-iconic-bootstrap.min.css (1 copy) 
        e6d48224-ed10-4327-8f2d-0259a65d7574 -- mih@meiner:/tmp/this [here]
ok
```

For comparision

```
% git annex addurl https://cdnjs.cloudflare.com/ajax/libs/open-iconic/1.1.1/font/css/open-iconic-bootstrap.min.css --file test
addurl https://cdnjs.cloudflare.com/ajax/libs/open-iconic/1.1.1/font/css/open-iconic-bootstrap.min.css 
(to test) ok
(recording state in git...)
% git annex whereis test
whereis test (2 copies) 
        00000000-0000-0000-0000-000000000001 -- web
        e6d48224-ed10-4327-8f2d-0259a65d7574 -- mih@meiner:/tmp/this [here]

  web: https://cdnjs.cloudflare.com/ajax/libs/open-iconic/1.1.1/font/css/open-iconic-bootstrap.min.css
ok
```

kyleam:
Hmm, odd.  That is the intention:

https://github.com/datalad/datalad/blob/4763acc30ddcece5db71062a1f37369ea9d42db9/datalad/interface/download_url.py#L166-L172

Perhaps the `is_under_annex` call above is getting confused.

mih:
I found some files for which it worked, no idea what the pattern is.

kyleam:
When I try the original command, I also see the warning 

```
[WARNING] Registering /tmp/this/open-iconic-bootstrap.min.css with https://cdnjs.cloudflare.com/ajax/libs/open-iconic/1.1.1/font/css/open-iconic-bootstrap.min.css failed: AnnexBatchCommandError: command 'addurl'
Error, annex reported failure for addurl (url='https://cdnjs.cloudflare.com/ajax/libs/open-iconic/1.1.1/font/css/open-iconic-bootstrap.min.css'): {'success': False, 'file': 'open-iconic-bootstrap.min.css', 'command': 'addurl'} [annexrepo.py:add_url_to_file:2120] `
```

I'm not sure why that is failing even though the direct `git annex addurl` call doesn't.

[...]

PR #3911, opened 2019-12-10
Fix download-url breakage reported in #2849

mih:
This is the least-cost path that made it work for me. Here is the compact use case:

```
datalad remove -d bf2849 --nocheck ; datalad create bf2849; datalad -C bf2849 download-url https://cdnjs.cloudflare.com/ajax/libs/open-iconic/1.1.1/font/css/open-iconic-bootstrap.min.css
```

The dataset deletion is critical for the demo, as otherwise git-annex seems to use prior knowledge about the key and doesn't fail on second attempt, even with a `git reset --hard HEAD~1`.

Fixes gh-2849

### DIFF
--- a/datalad/interface/download_url.py
+++ b/datalad/interface/download_url.py
@@ -220,7 +220,9 @@ URLs:
                                 url,
                                 # avoid batch mode for single files
                                 # https://github.com/datalad/datalad/issues/2849
-                                batch=len(annex_paths) > 1)
+                                batch=len(annex_paths) > 1,
+                                # bypass URL size check, we already have the file
+                                options=['--relaxed'])
                         except AnnexBatchCommandError as exc:
                             lgr.warning("Registering %s with %s failed: %s",
                                         path, url, exc_str(exc))
--- a/datalad/interface/download_url.py
+++ b/datalad/interface/download_url.py
@@ -215,7 +215,12 @@ URLs:
                         try:
                             # The file is already present. This is just to
                             # register the URL.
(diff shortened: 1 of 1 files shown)


mih:
re URL: Yes, this URL doesn't come with size info. It implied incompatibility with `download-url` though -- which I think is a bit strict.

kyleam:
> re URL: Yes, this URL doesn't come with size info. It implied incompatibility with `download-url` though

I'm not sure I understand the last sentence.  Is it saying that it is known that calling `git annex addurl` with an existing file will fail if the header doesn't have size information?

> which I think is a bit strict.

Sure, I wasn't objecting to this change.  I was trying to understand why that URL failed.  If you understood that it was due to a lack of size in the header, it'd be good to see that mentioned in the commit message.

mih:
> > re URL: Yes, this URL doesn't come with size info. It implied incompatibility with `download-url` though
> 
> I'm not sure I understand the last sentence. Is it saying that it is known that calling `git annex addurl` with an existing file will fail if the header doesn't have size information?

Yes.

> > which I think is a bit strict.
> 
> Sure, I wasn't objecting to this change. I was trying to understand why that URL failed. If you understood that it was due to a lack of size in the header, it'd be good to see that mentioned in the commit message.

Just re-read the my message. I agree that it is not clear from the text. Sorry, will fix.

mih:
Message is amended and force-pushed. No other code change, hence merging.

Issue #3233, opened 2019-03-16
Make (rev-)create take multiple PATHs arguments


yarikoptic:
ATM both `create` and `rev-create` operate on a single argument, so it is impossible in a single command to create multiple (sub)datasets, although I do not see any technical or semantical reason against it.  Sure thing 
`Dataset("newpath").create()`
 would create just a single dataset, but 
`create(["newds1", "newds2"])`
or even 
`Dataset("existingds").create(["newds1", "newds2"])`
should be perfectly clear in terms of semantic and probably implementation.

mih:
I am skeptical:

### Technical issues:

- the `description` argument *might* need the ability to get different values for different datasets
- lumping together operations on multiple datasets is ugly from the Git history perspective already, and will remain ugly until https://github.com/datalad/datalad-revolution/pull/71 has a solution, which will allow for the introduction of proper commit message handling when working with dataset hierarchies (only tangentially related here)

### Conceptual

- which problem is being solved, other than "it could be done" -- nothing against that, but giving tons of stuff that is rather urgent, it would inform us about the priority of this issue

Other than that I don't see an issue, but *please* let us have ONE `create` that needs to be adjusted first, not two.

yarikoptic:
> * the `description` argument _might_ need the ability to get different values for different datasets

if anything to differ, sure thing it would need multiple calls. I personally just resort to default description since I find it more useful than however I could potentially describe it for annex.

> * lumping together operations on multiple datasets is ugly from the Git history perspective already

I would disagree. One of the DataLad's goals is to blur the lines between datasets, i.e. often to  be able to use them as we use directories. And we do make commits affecting multiple directories.
I might want to create a collection of subdatasets preparing for multi subject/session study, so I ideally would like to do `datalad create -m "Creating datasets for my 10 initial subjects" sub-00{1..10}`.  I might want to create a default layout `datalad create code data images` to keep each component separately.  I really do not see why not to allow it if it could be easily done.



> * which problem is being solved, other than "it could be done" -- nothing against that, but giving tons of stuff that is rather urgent, it would inform us about the priority of this issue

+1 on that



> _please_ let us have ONE `create` that needs to be adjusted first, not two.

+10 on that ;-)

PR #3918, opened 2019-12-11
Towards a partial resolution of #3759

mih:
This has a long way to go. I am still in the phase of understanding why `install` works at all ;-)

- [x] be able to handle Path objects with PathRI
- [x] resolve issue that `install` reported results out-of-order wrt a given argument list

ref: #3759

### DIFF
--- a/datalad/distribution/tests/test_install.py
+++ b/datalad/distribution/tests/test_install.py
@@ -674,6 +674,11 @@ def test_install_skip_list_arguments(src, path, path_outside):
     assert_result_count(
         result, 1, status='impossible', message="path does not exist",
         path=opj(ds.path, 'not_existing'))
+    # TODO unclear to me why this error should be like this
+    # we are asking install to pull from multiple sources, this one doesn't
+    # work, but not because it is not "associated" with a dataset, but
+    # because it doesn't exist. If it would, we would expect it to
+    # install the source as a subdataset, wouldn't we?
     assert_result_count(
         result, 1, status='error',
         message=("path not associated with dataset %s", ds),
--- a/datalad/distribution/install.py
+++ b/datalad/distribution/install.py
@@ -39,7 +39,10 @@ from datalad.support.network import (
     RI,
     PathRI,
 )
-from datalad.utils import assure_list
+from datalad.utils import (
+    assure_list,
+    Path,
+)
 from datalad.dochelpers import exc_str
 
 from datalad.distribution.dataset import (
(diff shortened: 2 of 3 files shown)


datalad/distribution/tests/test_install.py
```diff
@@ -674,6 +674,11 @@ def test_install_skip_list_arguments(src, path, path_outside):
...
         result, 1, status='impossible', message="path does not exist",
         path=opj(ds.path, 'not_existing'))
+    # TODO unclear to me why this error should be like this
+    # we are asking install to pull from multiple sources, this one doesn't
+    # work, but not because it is not "associated" with a dataset, but
+    # because it doesn't exist. If it would, we would expect it to
+    # install the source as a subdataset, wouldn't we?
     assert_result_count(
         result, 1, status='error',
         message=("path not associated with dataset %s", ds),
```
mih: Do you agree with this assessment?

bpoldrack: I don't think so (or I misread that comment).
We are not calling `install -d . path_outside`, so we could conclude that `path_outside` is in fact a source. Rather we say explicitly that this the `path` to install. If it is not associated with a dataset and we didn't give a source, then we cannot deal with that, because it's not in a dataset (where we then could conclude a source from).

bpoldrack: Then again - I'm not quite clear what you mean by "we are asking install to pull from multiple sources".
How so? The install call doesn't say anything about sources. It's pretty much a get for a list of things, isn't it?

mih: > We are not calling `install -d . path_outside`, so we could conclude that `path_outside` is in fact a source.

Yes, we do `ds.install(...)`

> Rather we say explicitly that this the `path` to install. If it is not associated with a dataset and we didn't give a source, then we cannot deal with that, because it's not in a dataset (where we then could conclude a source from).

That is not have `install` works. It aims to know what you want. If you check the original issue, such multisource calls are explicitly desired. And unless `--source` is given, any path arg could be a source.

mih: IOW how is this different from:

```
datalad install -d . ///openfmri/ds00000{1,2} existing_subds
```

?

bpoldrack: Man, that's screwed even more than I remember. ;-)

But still. That would mean, that if it existed we would want to treat it as a source and install from there. However, if that path doesn't point to *some* dataset, what would you install then from there? It would still fail, because that thing is nothing we can install, since it's not associated with a dataset. Or what else do I miss?

mih: Yes, it should fail, just not with that error. Instead it should say

```
{
  "action": "install",
  "path": "/tmp/datalad_temp_test_install_skip_list_argumentscn7klenf/datalad_temp_test_install_skip_list_argumentsc8_dxvo2",
  "type": "dataset",
  "refds": "/tmp/datalad_temp_test_install_skip_list_argumentscn7klenf",
  "status": "error",
  "message": [
   "Failed to clone from all attempted sources: %s",
   [
    "/tmp/datalad_temp_test_install_skip_list_argumentsc8_dxvo2"
   ]
  ],
  "source_url": "/tmp/datalad_temp_test_install_skip_list_argumentsc8_dxvo2"
 },
```

which is what happens now with this PR.

It is just that I cannot fully figure out what is desired behavior, hence I appreciate the feedback.

bpoldrack: I see. I read the comment as "it should fail with something like `path doesn't exist`".
That new error is better - I agree.

yarikoptic: "Failed to clone from all attempted sources:"
nope.  
Current behavior (for better or for worse) is that whenever install gets a path as an argument, it is not taken as a source, but as a destination. Again the hybrid nature of datalad as "distribution"+VCS, so we can do e.g. `datalad install subds*` or `datalad ... search | xargs datalad install` etc.
 
If destination is already installed -- all good, IIRC we just would verify (where applicable) if it is the right one (source/version), etc.
Here because there  `-d` (some time in the past at 6957e9ab4835422d361356851ad4ee109acb524a or before) it was made forbidden to install paths outside of the explicitly specified dataset, and hence the message.  It is reall not that it cannot clone **from** that location (regardless either it exists or not).

mih: Oh wow. That is a revelation for me. And indeed:

```
(datalad3-dev) 127 mih@meiner /tmp % datalad create 11
[INFO   ] Creating a new annex repo at /tmp/11 
create(ok): /tmp/11 (dataset)
(datalad3-dev) mih@meiner /tmp % cd 11
(datalad3-dev) mih@meiner /tmp/11 (git)-[master] % datalad install -d . ~/forrest/collection/phase2
[WARNING] path not part of the reference dataset at /tmp/11 [get(/home/mih/forrest/collection/phase2)] 
get(impossible): ../../home/mih/forrest/collection/phase2 (dataset) [path not part of the reference dataset at /tmp/11]
(datalad3-dev) 1 mih@meiner /tmp/11 (git)-[master] % datalad --version
datalad 0.11.8
```

I am now at a complete loss of how this shall work. If a path given to `install` is always a destination, it can only ever be a location in an existing dataset. So it is really just a `get` that figures out the dataset to operate on.

mih: FTR: This means that `datalad -f '{path}' search -d /// haxby | xargs datalad install` (which I found a useful pattern to get matching datasets for a desired analysis is actually not applicable for a YODA scenario. I thought it would install all matching dataset in PWD, such that I could `save` them in a YODA dataset as inputs (or use `install -d .` right away). But it doesn't do that. I installs them scattered in their respective hierarchy locations under `~/datalad...`. Which is not helpful for setting up a YODA dataset. Going one by one via `install -s ///something dest` is the only working approach, which `clone` does as well.

bpoldrack: Shouldn't that be `datalad -f '{path}' search -d /// haxby | xargs datalad install -s` then? Emphasize on the `-s`.

yarikoptic: those are not sources -- those are destinations, so they should not be listed in `-s` unless you would like to take those found (but they might not even be present, so would not even work then) and install in another location.  I never wanted/needed to do that yet, and it would have not worked since `datalad install -s SRC [DEST]` can take only a single SRC at the moment.

bpoldrack: > unless you would like to take those found (but they might not even be present
Yes. I think that is what Michael had in mind and referred to as the YODA scenario. While that wouldn't work ATM, this could be the way to make it work (allow a list of sources as well). That way, with `-s` it would do, what Michael wants and without it would do, what you apparently had in mind when pointing out that usecase originally.

yarikoptic: I see.
So we have a new scenario which we have not originally envisioned or even encountered before.  I think that adding a new option `--sources` with `nargs="+"` to do `install DESTPREFIX --sources PATH1 [PATH2]` to complement `--source SRC [DEST]` would be the least ground shaking way to achieve it without breaking any compatibility.  Alternatively `--source` could be "broken" (`--source SRC DEST` would no longer work as desired; would require reordering but even then work differently since DEST would not be DESTPREFIX) to become `nargs="+"` and then `[DEST]` becoming the `[DESTPREFIX]`.  

So, I guess it was not yet even a single use case so far requiring such "multiple sources" functionality since we detected it only this late in the game, not sure if worth breaking API just to fulfill it or breeding new commands (as suggested in #3922).  
Related: we might also like to re-consider https://github.com/datalad/datalad/issues/3233 then (which would not even require any breakage ;-) )

mih:
Closing. Useful bit now in https://github.com/datalad/datalad/pull/3933

Issue #3915, opened 2019-12-10
WTF: make order of items consistent/deterministic


yarikoptic:
Comparing dumps from two appveyor runs for https://ci.appveyor.com/project/mih/datalad/builds/29438131
shows that things differ in order of items. compare
https://ci.appveyor.com/project/mih/datalad/builds/29438131/job/ytg87aoxo867wnw4?fullLog=true#L468
to
https://ci.appveyor.com/project/mih/datalad/builds/29438131/job/j5iln58t32ehlupt?fullLog=true#L468

The easiest would be just to sort those entries/keys before output

PR #3927, opened 2019-12-15
BF: Sort keys in wtf() result renderer (fixes gh-3915)

mih:
This fix conflicts with a prior test condition (now flipped). Please check.

I consider a defined top-level (non-alphanum) sorting a needless complication.


mih:
Appveyor stalls, like it often does.

yarikoptic:
> I consider a defined top-level (non-alphanum) sorting a needless complication.

yeah, and the sorted version not too far from original anyways, but easier to find sections when sorted. cool, thanks! merging

Issue #3870, opened 2019-11-15
0.12.0rc6.dev92 doesn't report OutOfSpaceError (0.11.8.dev16 does)


yarikoptic:
The original issue of reporting out of space https://github.com/datalad/datalad/issues/2938 was fixed by https://github.com/datalad/datalad/pull/2958 a year ago.

```
(dev3) yoh@ip-172-31-37-151:~/allen-neuropixels-try1$ datalad --version
datalad 0.11.8.dev16
(dev3) yoh@ip-172-31-37-151:~/allen-neuropixels-try1$ datalad get probe_848037572_lfp.nwb
OutOfSpaceError: command 'annex get' needs 1.4 GB more
```

but with

```
(dev3) yoh@ip-172-31-37-151:~/allen-neuropixels-try1$ datalad --version
datalad 0.12.0rc6.dev92
(dev3) yoh@ip-172-31-37-151:~/allen-neuropixels-try1$ datalad get probe_848037572_lfp.nwb
[WARNING] Running get resulted in stderr output: git-annex: get: 1 failed
 
[ERROR  ] Unable to access these remotes: web; Try making some of these repositories available:;        00000000-0000-0000-0000-000000000001 -- web; ; (Note that these git remotes have annex-ignore set: origin) [get(/home/yoh/allen-neuropixels-try1/probe_848037572_lfp.nwb)] 
get(error): probe_848037572_lfp.nwb (file) [Unable to access these remotes: web; Try making some of these repositories available:;      00000000-0000-0000-0000-000000000001 -- web; ; (Note that these git remotes have annex-ignore set: origin)]
```

```
(dev3) yoh@ip-172-31-37-151:~/allen-neuropixels-try1$ git annex version
git-annex version: 7.20190819+git2-g908476a9b-1~ndall+1
```

kyleam:
My shot in the dark at the culprit: 3fd47a079 (BF: annex --json -> annex --json --json-error-messages, 2019-10-05)

yarikoptic:
and unfortunately the only test we have is the mock'ed up test of reporting:
```
@with_tempfile(mkdir=True)
def test_AnnexRepo_get_outofspace(annex_path):
    ar = AnnexRepo(annex_path, create=True)

    def raise_cmderror(*args, **kwargs):
        raise CommandError(
            cmd="whatever",
            stderr="junk around not enough free space, need 905.6 MB more and after"
        )

[...]

Issue #3930, opened 2019-12-16
datalad.interface.results.annexjson2result() needs tests


mih:
This is a key function and the central place for translating annex JSON reports into DataLad results. It has no dedicated unit tests. We need to have some.

mih:
Oh well, many things need tests. And this one too, but no one has cared over the past two years, and likely no one will start doing so, because of this issue being open (ignoring the up-to-date coverage reports that are readily available).

PR #3931, opened 2019-12-16
BF: annexjson2result() did not report 'error-messages' (fixes gh-3870)

mih:
Such information is now extracted from the annex report, and used
preferentially as 'message' when competing with a simultaneously
present 'note'. While this may hide some information, the omitted
parts are arguably less important and obfuscate any error. Without
this omission, a 'disk full' error may look like:
```
   get(error): sub-01/ses-movie/func/sub-01_ses-movie_task-movie_run-1_bold.nii.gz (file) [  not enough free space, need 31.04 MB more (use --force to override this check or adjust annex.diskreserve); Unable to access these remotes: mddatasrc; Try making some of these repositories available:;     2f51364e-815a-4ad9-b380-617c682b1680 -- [mddatasrc];    40809380-c618-4801-bb8d-81320474c9da -- studyforrest phase2 original source repository]
```

Also: https://github.com/datalad/datalad/issues/3930

### DIFF
--- a/datalad/interface/results.py
+++ b/datalad/interface/results.py
@@ -212,6 +212,7 @@ def annexjson2result(d, ds, **kwargs):
       Passes as-is to `get_status_dict`. Must not contain `refds`.
     """
     lgr.debug('received JSON result from annex: %s', d)
+    messages = []
     res = get_status_dict(**kwargs)
     res['status'] = 'ok' if d.get('success', False) is True else 'error'
     # we cannot rely on any of these to be available as the feed from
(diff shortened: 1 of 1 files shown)


mih:
Appveyor failed to submit coverage in 1/2 runs.

yarikoptic:
oh, now it is in the 1st run, for me it was in the 2nd, so indeed it is just in 1/2 of runs, weird

mih:
It tends to be random.

datalad/interface/results.py
```diff
@@ -227,13 +228,18 @@ def annexjson2result(d, ds, **kwargs):
         res['metadata'] = {k: v[0] if isinstance(v, list) and len(v) == 1 else v
                            for k, v in d['fields'].items()
                            if not k.endswith('lastchanged')}
-    # avoid meaningless standard messages
-    if 'note' in d:
+    if d.get('error-messages', None):
+        messages.extend(d['error-messages'])
+    # avoid meaningless standard messages, and collision with actual error
+    # messages
+    elif 'note' in d:
```
yarikoptic: I do agree that errors are more important, but I wonder if `note` should also be retained? 
i.e. `elif` to be changed back to `if`?

mih: That is the example I posted. I think it makes it worse.

yarikoptic: and there is no way to keep error message "separate" so it could e.g. get a distinctive color (red) while rendering the result?

mih: We don't have that yet, but I recorded the need https://github.com/datalad/datalad/issues/3939

mih:
Minus a [future enhancement](https://github.com/datalad/datalad/issues/3939) this seems to be ready, and I will merge it to get the issue resolved for now.

PR #3935, opened 2019-12-16
BF: lgr - use .setLevel() instead of .level =

### LINKED ISSUE #3545: swallow_logs() doesn't appear to swallow output on py3.7
My system Python recently went from 3.6 to 3.7, and I noticed that some tests got a lot noisier, though I can't be sure the two things actually coincided.  Comparing the [3.6] and [3.7] runs on Travis seem to confirm that there's something wrong with swallow_logs() on 3.7.

[3.6]: https://travis-ci.org/datalad/datalad/jobs/560427245#L1096
[3.7]: https://travis-ci.org/datalad/datalad/jobs/560427244#L1098

yarikoptic:
Apparently .level is not even a property and I guess starting from
3.7 treatment has changed, or broke.  May be smth like (scheduled for 3.9
with backport requests for 3.8) https://github.com/python/cpython/pull/16325
was intended to fix it, I did not investigate deep enough -- just fixed
the issue we started to experience (Closes #3545)

After merged into 0.11.x, will send PR against master

### DIFF
--- a/datalad/utils.py
+++ b/datalad/utils.py
@@ -1290,7 +1290,8 @@ def swallow_logs(new_level=None, file_=None, name='datalad'):
         # TODO: if file_ and there was an exception -- most probably worth logging it?
         # although ideally it should be the next log outside added to that file_ ... oh well
     finally:
-        lgr.handlers, lgr.level = old_handlers, old_level
+        lgr.handlers = old_handlers
+        lgr.setLevel(old_level)
         adapter.cleanup()
 
 


kyleam:
>  just fixed the issue we started to experience (Closes #3545)

Running a test that I know was problematic earlier, I can confirm that this quiets it.  Thanks.

kyleam:
> After merged into 0.11.x, will send PR against master

FWIW my opinion is that the extra PR is unnecessary.  A local merge and push is fine unless that merge is particularly involved (and even in those cases, I think pushing to a scratch branch to make sure the tests passes is usually sufficient).

yarikoptic:
Ack

yarikoptic:
FTR: bug is interesting in that it manifested itself only after 2nd use of swallow_logs

kyleam:
> FTR: bug is interesting in that it manifested itself only after 2nd use of swallow_logs

Weird.  I remember that when I looked into it a bit when reporting, I couldn't figure out why the issue didn't seem to affect some tests.  That probably explains it.

kyleam:
[Before][0] and [after][1].  Much better.  Thanks @yarikoptic.

[0]: https://travis-ci.org/datalad/datalad/jobs/625682102#L1085
[1]: https://travis-ci.org/datalad/datalad/jobs/625736272#L1097

yarikoptic:
coolio, pushed master with 0.11.x (just this PR) merged

Issue #2935, opened 2018-10-19
a way to limit search output with a single (or specified # of) hit per dataset


yarikoptic:
#### What is the problem?
Quite a few times I wanted to find datasets which satisfy my search query, and see what (file(s)) actually satisfied it.  E.g. now I wanted to see datasets which carry files with `_acq-ap` in their names (for suggesting an enhancement to BIDS) I have used
```
datalad -f dataset -c datalad.search.index-egrep-documenttype=all search path:.*_acq-ap.*
```
which keeps dumping lots of hits per each dataset, wasting time and screen space.
Limiting output to `-f '{parentds}'` and piping to `uniq` provides a list of datasets but wastes time and cannot be used with `{path}` to also see "sample hit(s)".

I am not sure yet if there is any other similar usecase to have such a feature implemented at a "base" level (and either it would be easy to do so while interrupting underlying command generator before it exhausts itself), but definitely makes sense for the search command.

What do you think?

yarikoptic:
@mih - your input would be greatly appreciated. I keep running into this use case, and thinking about either adding an additional option or extending existing
```
  --max-nresults MAX_NRESULTS
                        maxmimum number of search results to report. Setting
                        this to 0 will report all search matches. Depending on
                        the mode this can search substantially slower. If not
                        specified, a mode-specific default setting will be
                        used. Constraints: value must be convertible to type
                        'int' [Default: None]
```

to be able to say smth like `--max-nresults parentds=2` which would limit to 2 outputs per each `parentds`

yarikoptic:
a user use case: https://neurostars.org/t/datalad-search-openneuro/18386

mih:
This issues refers to functionality that has been removed in https://github.com/datalad/datalad/pull/7014

PR #3948, opened 2019-12-19
NF: search - --mode pyeval which evaluates python expression on provided metadata

yarikoptic:
This mode is motivated by question on OpenNeuro
https://neurostars.org/t/fast-tr-open-task-fmri-datasets/5601
on how to find datasets/files with sub-second TR.

Unfortunately

- egrep mode is too crude since relies on dealing with string
  representation of values, so would be inconvenient to generate regular
  expression and less efficient than simply comparing numerical values.

- whoosh based implementations relying on constructing index could work if
  custom index was first created. But we would need to implement
  interface for specifying index, and initial invocation would be quite slow
  since first it would need to establish index and then run the query.
  For one time invocation might be too much.  For any complicated query it
  would also require learning woosho syntax first.

With pyeval we would enable an arbitrarily complex queries which could be
expressed as a python expression.  That would even allow for use of types'
methods (e.g. .lower() or even .format() on string values) and operations
(+, -, etc) making it quite flexible and versatile.  Many users (if not most)
would be able to easily construct such queries.

Sample invocation:

```shell
$> datalad -f '{path}: {metadata[bids][RepetitionTime]}' -c datalad.search.index-pyeval-documenttype=all search -d /// --mode pyeval "(bids['type'] == 'bold') and (bids['RepetitionTime'] < 1)"
/home/yoh/datalad/dbic/QA/sourcedata/sub-emmet/ses-20180521/func/sub-emmet_ses-20180521_task-rest_acq-3mm_bold.dicom.tgz: 0.426
/home/yoh/datalad/dbic/QA/sourcedata/sub-emmet/ses-20180531/func/sub-emmet_ses-20180531_task-rest_acq-3mm_bold.dicom.tgz: 0.426
/home/yoh/datalad/dbic/QA/sub-emmet/ses-20180521/func/sub-emmet_ses-20180521_task-rest_acq-3mm_bold.nii.gz: 0.426
/home/yoh/datalad/dbic/QA/sub-emmet/ses-20180531/func/sub-emmet_ses-20180531_task-rest_acq-3mm_bold.nii.gz: 0.426
/home/yoh/datalad/dbic/QA/sub-qa/ses-20161128/func/sub-qa_ses-20161128_task-rest_acq-p2Xs4X35mm_bold.nii.gz: 0.525
/home/yoh/datalad/dbic/QA/sub-qa/ses-20161205/func/sub-qa_ses-20161205_task-rest_acq-p2Xs4X35mm_bold.nii.gz: 0.525
...
```

TODOs (here or later elsewhere):

- [ ] Common base for `egrep` and `pyeval`.  This crude prototype was coded by
  taking `egrep` implementation.  Both are naive search by looping through
  provided to them records.  I think we might want
  RF to provide a common "looping" and possibly generalized treatment for
  keys reporting etc.  That would allow for centralized optimization (e.g.
  parallel execution). etc
- [x] see if there is a way to OPT that `eval` on the same expression over and over again (probably just use [compile](https://www.geeksforgeeks.org/python-compile-function/))
- [ ] should we open the door to `exec` ? (multiple statements could be taken as parts of the `matcher` function body intended to return `bool`) .
- [ ] RF search (at least of `egrep` and `pyeval`) so we could limit by number of hits per dataset (#2935).
  - [ ] generalize to allow to "group" within a dataset based on some additional keys (e.g., could be a `subject` field for BIDS)
  - [ ] groupping for query ("dataset + subject") and "reporting" ("dataset") could be different... 
  - [ ] flag `--report-missing`, makes sense only if groupping enabled, and would report groups where not a single match happened.  Use case: "find datasets (or subjects within a dataset) missing some particular file type"
- [ ] with aforementioned RF to make it dataset/files nested loop - we could provide `matcher` with both metadata for the record and for the parentds (if a file), making it possible to make joint inquiry based on metadata of the dataset and the file (we do not have any other "levels" or "grouppings" such as subject ATM, but that might change ;-))
- [ ] docs

Notes:

<details>
<summary> On a prototypical run, seems to be about twice faster than egrep version</summary> 

```shell
$> time datalad -f '{path}: {metadata[nifti1][dim]}  {metadata[bids][RepetitionTime]}' -c datalad.search.index-egrep-documenttype=all search -d ~/datalad/openneuro --mode egrep bids.type:bold 'bids.RepetitionTime:1(\.0*)*$' 2>&1 | nl | tail -n 3
   660	/home/yoh/datalad/openneuro/ds001720/sub-238/ses-1/func/sub-238_ses-1_task-rest_acq-EPI_bold.nii.gz: [4, 90, 12, 70, 450, 1, 1, 1]  1
   661	/home/yoh/datalad/openneuro/ds001720/sub-239/ses-1/func/sub-239_ses-1_task-rest_acq-EPI_bold.nii.gz: [4, 90, 12, 70, 450, 1, 1, 1]  1
   662	/home/yoh/datalad/openneuro/ds001720/sub-240/ses-1/func/sub-240_ses-1_task-rest_acq-EPI_bold.nii.gz: [4, 90, 12, 70, 450, 1, 1, 1]  1
datalad -f '{path}: {metadata[nifti1][dim]}  {metadata[bids][RepetitionTime]}  5.86s user 0.24s system 100% cpu 6.073 total
nl  0.01s user 0.00s system 0% cpu 6.073 total
tail -n 3  0.00s user 0.00s system 0% cpu 6.073 total
(dev3) 1 12899.....................................:Thu 19 Dec 2019 09:24:06 AM EST:.
lena:/tmp
$> time datalad -f '{path}: {metadata[nifti1][dim]}  {metadata[bids][RepetitionTime]}' -c datalad.search.index-pyeval-documenttype=all search -d ~/datalad/openneuro --mode pyeval "(bids['type'] == 'bold') and (bids['RepetitionTime'] == 1)" 2>&1 | nl | tail -n 3
   660	/home/yoh/datalad/openneuro/ds001720/sub-238/ses-1/func/sub-238_ses-1_task-rest_acq-EPI_bold.nii.gz: [4, 90, 12, 70, 450, 1, 1, 1]  1
   661	/home/yoh/datalad/openneuro/ds001720/sub-239/ses-1/func/sub-239_ses-1_task-rest_acq-EPI_bold.nii.gz: [4, 90, 12, 70, 450, 1, 1, 1]  1
   662	/home/yoh/datalad/openneuro/ds001720/sub-240/ses-1/func/sub-240_ses-1_task-rest_acq-EPI_bold.nii.gz: [4, 90, 12, 70, 450, 1, 1, 1]  1
datalad -f '{path}: {metadata[nifti1][dim]}  {metadata[bids][RepetitionTime]}  2.62s user 0.19s system 100% cpu 2.787 total
nl  0.01s user 0.00s system 0% cpu 2.787 total
tail -n 3  0.00s user 0.00s system 0% cpu 2.787 total
```
</details>


yarikoptic:
although I even dream of it now, never managed to finish yet, so no need to keep it floating around -- the branch is there

PR #3956, opened 2019-12-21
BF: use mimeencoding not mimetype for text2git

### LINKED ISSUE #3361: we might need to reconsider approach for text2git

yarikoptic:
Also extends the test to test some structured text files to guarantee that there would be no change in behavior (from git-annex or libmagic changes).

I was a bit not sure if it is worth introducing in 0.11.x since it would demand minimal git annex boost.  But since I use ATM both versions interchangeably and older annex would not support (crash) mimeencoding - I have decided that it is worth adding it to 0.11.x first.

I also was not sure if we shouldn't detect "old" line in `.gitattributes` and warn the user and may be even suggest (in interactive sessions) to adjust that gory statement.  But if we do so, we would need to mark somewhere to not pester user about it again in that dataset if they say "nah"... so - pain. I guess we are just doomed to look forward.

Closes #3361

### DIFF
--- a/datalad/distribution/create.py
+++ b/datalad/distribution/create.py
@@ -343,9 +343,9 @@ class Create(Interface):
                 attrs = tbrepo.get_gitattributes('.')
                 # some basic protection against useless duplication
                 # on rerun with --force
-                if not attrs.get('.', {}).get('annex.largefiles', None) == '(not(mimetype=text/*)and(largerthan=0))':
+                if not attrs.get('.', {}).get('annex.largefiles', None) == '((mimeencoding=binary)and(largerthan=0))':
                     tbrepo.set_gitattributes([
-                        ('*', {'annex.largefiles': '(not(mimetype=text/*)and(largerthan=0))'})])
+                        ('*', {'annex.largefiles': '((mimeencoding=binary)and(largerthan=0))'})])
                     add_to_git.append('.gitattributes')
 
         if native_metadata_type is not None:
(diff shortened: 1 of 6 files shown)


yarikoptic:
> LGTM.
> 
> Sadly still no solution for the windows build of git-annex :(
> 
> ```
> git-annex: bad annex.largefiles configuration: Parse failure: "mimeencoding" not supported; not built with MagicMime support
> ```

I forgot already why there is no such a build.  So, I just asked -- https://git-annex.branchable.com/todo/provide_windows_build_with_MagicMime/?updated

Issue #3619, opened 2019-08-25
run-procedure --discover report strange location


mih:
I am in a subdir inside a dataset
```
% datalad run-procedure --discover
cfg_text2git (../git/datalad/resources/procedures/cfg_text2git.py) [python_script]
cfg_yoda (../git/datalad/resources/procedures/cfg_yoda.py) [python_script]
cfg_metadatatypes (../git/datalad/resources/procedures/cfg_metadatatypes.py) [python_script]
```

The relative paths do not  point to the scripts. Possibly the paths are reported relative to the dataset I am in, but they should be relative to the subdirectory IMHO.

bpoldrack:
Minor comment: My guess is they are relative to datalad installation root dir.

PR #3959, opened 2019-12-21
BF: Always report absolute location of discovered procedures

mih:
Before it would report relpaths if operating on a dataset, leading to
the issue at hand.

I considered relpaths for dataset internal procedures, and abspaths for
others. However, when not invoked at the root of a dataset these would
again be somewhat confusing. So full abspath it is.

Fixes gh-3619


mih:
Thx!

PR #3963, opened 2019-12-22
Refurbish eval_func()

### LINKED ISSUE #3908: Remove --proc-pre/post (replacement: result hooks)
This new feature is proposed in https://github.com/datalad/datalad/pull/3903

I propose to remove its predecessors, because their are (from my POV) completely unused, untested, and less capable (only dataset procedures, not any DataLad command).

### LINKED ISSUE #3962: result_xfm specification disables result hooks
The reason is that a `result_xfm` is applied at the innermost result loop, and result hooks are matched against a result after it has been transformed. This confirms my suspicion from https://github.com/datalad/datalad/issues/3953

mih:
This became a little more elaborate than desired, but I think it is useful nevertheless.

- [x] Added a test that demos how `result_xfm` broke result hooks https://github.com/datalad/datalad/issues/3962
- [x] Ripped out proc-pre/post feature based on feedback in https://github.com/datalad/datalad/issues/3908 in order to get a clean fix for #3962 (which fixes #3908)
- [x] Actually fix #3962 
- [x] RF the decorator code, created helper functions (even if used just once) and removed duplicate variable to improve the subjective readability of the code
- [x] Result renderers now get to see all command args. Before it was dependent on how exactly a command was called: `cmd(path='this')` vs `cmd('this')` -- only the former case used to make it to the result renderer `kwargs`.

### DIFF
--- a/datalad/interface/utils.py
+++ b/datalad/interface/utils.py
@@ -353,20 +353,6 @@ def eval_results(func):
                 def _result_filter(res):
                     return result_filter(res, **allkwargs)
 
-        def _get_procedure_specs(param_key=None, cfg_key=None, ds=None, proc_cfg=None):
-            lgr.log(2, "Getting procedure specs for %s of %s", param_key, ds)
-            spec = common_params.get(param_key, None)
-            if spec is not None:
-                # this is already a list of lists
-                return spec
-
-            spec = proc_cfg.get(cfg_key, None)
-            if spec is None:
-                return None
-            elif not isinstance(spec, tuple):
-                spec = [spec]
-            return [shlex.split(s) for s in spec]
-
         # query cfg for defaults
         cmdline_name = cls2cmdlinename(_func_class)
         dataset_arg = allkwargs.get('dataset', None)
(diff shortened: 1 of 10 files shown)


mih:
I'll merge this later today, unless someone stops me.

datalad/interface/utils.py
```diff
@@ -309,70 +320,49 @@ class (if any).
...
+
+        # determine the command class associated with `wrapped`
+        wrapped_class = get_wrapped_class(wrapped)
 
         # retrieve common options from kwargs, and fall back on the command
         # class attributes, or general defaults if needed
         kwargs = kwargs.copy()  # we will pop, which might cause side-effect
         common_params = {
             p_name: kwargs.pop(
+                # go way a any explicitly given default
```
kyleam: `s/go way a/go with/`?

datalad/interface/utils.py
```diff
@@ -453,31 +423,34 @@ def generator_func(*_args, **_kwargs):
...
                         for hr in run_jsonhook(hook, spec, r, dataset_arg):
-                            yield hr
-                yield r
+                            # apply same logic as for main results, otherwise
+                            # any filters would only tackle the primary results
+                            # and a mixture of return values could happen
+                            if not keep_result(hr, result_filter, **allkwargs):
+                                continue
+                            hr = xfm_result(hr, result_xfm)
+                            # rational for conditional is a few lines down
```
kyleam: `s/rational/rationale`

datalad/interface/utils.py
mih: Thx! done.

datalad/interface/utils.py
mih: Done.

mih:
Thx @kyleam !

I have fixed the typos. Your assessment re documentation quality is, unfortunately, correct. Initially this was supposed to fix the hooks implementation, but then it got messy. :(

kyleam:
@mih
> I'll merge this later today, unless someone stops me.

No one has yelled "stop!" :), so I'll merge this in now.  Thanks @mih.

mih:
Thx @kyleam

PR #3966, opened 2019-12-27
Transition `clone()` to core API

mih:
This is implementing a resultion of https://github.com/datalad/datalad/issues/3957

- [x] RF to make clone(`alt_sources`) parameter obsolete (new internal helper for `get()` and `clone()`)
- [x] Optimize candidate URL handling (only generate once, do proper progress logging, no adhoc)
- [x] Equalize return value behavior with `install()`
- [x] Turn `--reckless` into a mode-style option (backward-compatible)
- [x] Fix #3958 
- [x] Mitigate #3969
- [x] Fix https://github.com/datalad/datalad/issues/3971
- [x] promote `_clone_dataset()` to `clone_dataset()` for use by extensions and other expert Python users
- [x] configure origin of source as a remote https://github.com/datalad/datalad/issues/3926
- [x] go beyond #3926 and discover all traceable local dataset siblings as remotes
- [x] move to `datalad.core.distributed` following https://github.com/datalad/datalad/issues/3192
- [x] add compatibility shim analog to `datalad/interface/run.py`
- [x] investigate https://github.com/datalad/datalad/issues/1146
- [x] investigate https://github.com/datalad/datalad/issues/1786
- [x] investigate https://github.com/datalad/datalad/issues/1956
- [x] beautify relpath submodule url/path. fixes #3538
- [x] investigate #1956
- [x] get testrepos working, so that all tests could run on windows (via #3975)

PR #3975, opened 2019-12-31
Enabled testrepo-based tests on Windows

mih:
...as much as feels feasible right now. In any case, this is a huge step forward in terms of critical test coverage on windows. We are now in the position to investigate individual tests, instead of facing global failure.

- [x] cross-platform implementation of get_local_file_url()
- [x] basic testrepo setup is running on windows
- [x] all new test failures due to testrepos being enabled on windows are marked "known" and annotated with a traceback log link
- [x] addurl is not working on appveyor, fix or disable testrepos on appveyor (leaning towards the latter)

There is still an open issue with `get_local_file_ur l()` https://github.com/datalad/datalad/issues/3977, but given that this is needed for #3966 it will have to come in a separate PR.

### DIFF
--- a/datalad/support/network.py
+++ b/datalad/support/network.py
@@ -900,8 +900,6 @@ def is_ssh(ri):
         or (isinstance(_ri, URL) and _ri.scheme == 'ssh')
 
 
-#### windows workaround ###
-# TODO: There should be a better way
 def get_local_file_url(fname):
     """Return OS specific URL pointing to a local file
 
--- a/datalad/support/network.py
+++ b/datalad/support/network.py
@@ -39,6 +39,7 @@ from datalad.dochelpers import exc_str
 from datalad.utils import (
     on_windows,
     PurePath,
+    Path,
 )
 from datalad.utils import assure_dir, assure_bytes, assure_unicode, map_items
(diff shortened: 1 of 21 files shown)


datalad/support/network.py
```diff
@@ -900,22 +901,24 @@ def is_ssh(ri):
         or (isinstance(_ri, URL) and _ri.scheme == 'ssh')
 
 
-def get_local_file_url(fname):
+def get_local_file_url(path):
```
kyleam: Grepping around the extensions, I don't see any calls to `get_local_file_url` that use `fname` as a keyword argument, so changing the parameter name should be fine.  But at the same time, any subjective benefit of the name change doesn't seem to be worth making a breaking change.

If we do stick with name change, the docstring needs to be updated.

mih: You are right, I will revert it.

mih: Done

mih:
Given that this PR makes a number of additional tests run, with relatively minimal impact on behavior on other platforms I'll merge this one now. A follow-up PR will (have to) come with #3977, so potential fixups can be incorporated there.

Issue #3972, opened 2019-12-29
siblings('add', url=<windows path>) yields non-standard path in config


mih:
Here is how it is supposed to look (clone from local path on windows)
```
[remote "origin"]
        url = C:/Users/mih/code/datalad
```

However, giving a windows path to `siblings()` as URL, causes that path to go verbatim into the config, not in frontslash from.

PR #3976, opened 2020-01-01
BF: Store path URLs for remotes in POSIX (fixes gh-3972)

mih:
with a test.


mih:
This should be uncontroversial. Will merge.

Issue #3639, opened 2019-09-01
Optimized gitrepo cloning broken on win10


mih:
Originally reported in #3635. Making it a separate issue, because the test will be disabled on win in order to make progress on more important issues.

This issue is legit, inodes seem to be reported properly, but do not match.

```
======================================================================
FAIL: datalad.support.tests.test_gitrepo.test_optimized_cloning
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Users\mih\Miniconda3\lib\site-packages\nose\case.py", line 198, in runTest
    self.test(*self.arg)
  File "c:\users\mih\code\datalad\datalad\tests\utils.py", line 607, in newfunc
    return t(*(arg + (filename,)), **kw)
  File "c:\users\mih\code\datalad\datalad\support\tests\test_gitrepo.py", line 1161, in test_optimized_cloning
    eq_(origin_inodes, clone_inodes, msg='with src={}'.format(src))
AssertionError: with src=file:///C%3A/Users/mih/AppData/Local/Temp/datalad_temp_xfuka682/origin
```

mih:
This is not just a windows issue, but affects crippled FS as well.

```
======================================================================
FAIL: datalad.support.tests.test_gitrepo.test_optimized_cloning
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/mih/env/datalad3-dev/lib/python3.7/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/mih/hacking/datalad/git/datalad/tests/utils.py", line 187, in newfunc
    return func(*args, **kwargs)
  File "/home/mih/hacking/datalad/git/datalad/tests/utils.py", line 715, in newfunc
    return t(*(arg + (filename,)), **kw)
  File "/home/mih/hacking/datalad/git/datalad/support/tests/test_gitrepo.py", line 1152, in test_optimized_cloning
    eq_(origin_inodes, clone_inodes, msg='with src={}'.format(src))
AssertionError: with src=/media/mih/Samsung_T5/tmp/datalad_temp_test_optimized_cloningby6pyqhk/origin
```

Revealed via gh-4708

yarikoptic:
FWIW, on a verge of 2022 we still have that test disabled on windows 
```
(git)lena:~datalad/datalad-maint[maint]git
$> git grep -4 test_optimized_cloning
datalad/support/tests/test_gitrepo.py-# this is simply broken on win, but less important
[...]

PR #3979, opened 2020-01-02
BF: Implement workaround for file:// URL compatibility on Windows

mih:
The new 'compatibility' argument of get_local_file_url() is relatively
painless and the workaround implementation can be removed (and the
argument rendered into a no-op) if there should be a compatibility
fix at the git-annex end.

Fixes gh-3639

### DIFF
--- a/datalad/distribution/clone.py
+++ b/datalad/distribution/clone.py
@@ -217,7 +217,8 @@ class Clone(Interface):
                 # this is where it was actually installed from
                 track_name, track_url = _get_tracking_source(destination_dataset)
                 if track_url in guessed_sources or \
-                        get_local_file_url(track_url) in guessed_sources:
+                        get_local_file_url(
+                            track_url, compatibility='git') in guessed_sources:
                     yield get_status_dict(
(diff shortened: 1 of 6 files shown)


mih:
Ha, so even if we generate proper Git-compatible file:// URLs on Windows this doesn't work, because GitRepo.clone() throws them away and takes a wrong `localpath`. Here is the relevant code:

```
       # try to get a local path from `url`:
        try:
            url = url_ri.localpath
            url_ri = RI(url)
        except ValueError:
            pass
```

It is unclear to me why a `file://` URL cannot be passed as-is. Tracing the evolution of a URL through the code is a bit funny. There is also this step in `clone.py`:

```
        # Possibly do conversion from source into a git-friendly url
        # luckily GitRepo will undo any fancy file:/// url to make use of Git's
        # optimization for local clones....
        source_url = source
        source_ = _get_git_url_from_source(source)
 ```

which converts from str to RI to str, which is later converted back into RI.

mih:
Thx @kyleam

PR #3981, opened 2020-01-02
RF: minor linting for consistent code etc (just found uncomitted)

yarikoptic:


datalad/core/local/status.py
```diff
@@ -306,8 +305,8 @@ def __call__(
         # To the next white knight that comes in to re-implement `status` as a
         # special case of `diff`. There is one fundamental difference between
         # the two commands: `status` can always use the worktree as evident on
-        # disk as a contraint (e.g. to figure out which subdataset a path is in)
-        # `diff` cannot do that (everything need to be handled based on a
+        # disk as a contsraint (e.g. to figure out which subdataset a path is
```
mih: I has all the letters now ;-)

Issue #3973, opened 2019-12-29
datalad update -r fails with "AttributeError() [dataset.py:_flyweight_id_from_args:123]"


soichih:
I am running `datalad update -r .` on previously installed datalad superset root directory. It ran for a while and eventually ran into the following error message.

```
update(ok): /mnt/datalad/datasets.datalad.org/workshops/repronim-sfn-2017/challenge (dataset)
[ERROR  ] AttributeError() [dataset.py:_flyweight_id_from_args:123] (AttributeError) 
```

I am using running datalad 0.11.8 on Ubuntu 18.04

yarikoptic:
Haven't seen such errors (from flyweight) for a while... Given that update without merge is just a git fetch - strange. Could you try with current master?

soichih:
I just ran this again and it finished successfully this time. I didn't update any client or the environment..  Has anything changed on datasets.datalad.org end that fixed it?

yarikoptic:
hm... I did re "publish" earlier today to add more datasets.  It should have not been related, but oh well -- bugs are tricky some times ;)  Did you try multiple times originally? if it just crashed once and then you didn't try second time -- I would say that the update I did today is unlikely to relate. 
FWIW, I have created a "manual" (guaranteed to not disappear) snapshot of the entirety of `/srv` on the server, so in principle we could try to reproduce, if the issue was reproducible originally (and if you somehow also got a snapshot of original issue, or remember the state from which you originally installed it).
BTW, in future, if you get into such a crash, rerun with `datalad --dbg` so if it crashes, you end up in `pdb` :  in this particular case 

- code could be improved to provide at least some information in the exception message, and currently it is just
```python
        if path is None:
            raise AttributeError
```
and could have been
```python
        if path is None:
            raise AttributeError("path must not be None. args: %s, kwargs: %s" % (args, kwargs))
```
(I am not even sure why it is AttributeError and not e.g. ValueError, not familiar with this code path).
and then investigate traceback (`bt`) and actual values which lead to this crashing scenario

soichih:
I ran 3, 4 times on both Dev and prod instances and I was seeing the same problem. Thank you for looking into this!

yarikoptic:
> I ran 3, 4 times on both Dev and prod instances and I was seeing the same problem. Thank you for looking into this!

[...]

PR #3986, opened 2020-01-03
RF: Attribute -> Value Error + provide an informative message for the exception

yarikoptic:
See https://github.com/datalad/datalad/issues/3973 .   I am not really sure why it was
intended to be an AttributeError, but I feel that ValueError is a better fit here.
lgr.debug message might be of help in some cases to identify the reason which lead
to crash, but because args and kwargs are modified by the code above, not sure really
if it would be very informative.  but I do not think that it would somehow negatively
effect the performance so I added it

Looking at the code I see that we have duplicated functionality across `_flyweight_id_from_args` implementations.  I will look into unifying it in a separate PR (on top)

### DIFF
--- a/datalad/distribution/dataset.py
+++ b/datalad/distribution/dataset.py
@@ -120,7 +120,8 @@ class Dataset(object):
             raise TypeError("__init__() requires argument `path`")
 
         if path is None:
-            raise AttributeError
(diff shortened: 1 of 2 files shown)


yarikoptic:
any objections/recommendations, or good to merge?

PR #3987, opened 2020-01-04
RF: bring common (?) functionality under PathBasedFlyweight

yarikoptic:
We had to almost identical copies of the code for `_flyweight_id_from_args`.
That can breed bugs having fixes in one copy but not another.  I have decided
to go superclass subclassing path so we could also "modularize" the logic for
handling aliases (ATM only for the datasets).
    
I have taken implementation from  Dataset, which did not explicitly
"realpath" the path.  Not sure yet if such difference in behavior (Dataset vs *Repo)
was intended/desired, tests might show

- [x] make sure no notable hit on benchmarks
- [x] decide on that `realpath` handling
- Sits on top of #3986 
- TODO: For the merge in master, conflict resolution should add handling of yet another piece of seems duplicate path handling/conversion:
```diff
+        # mirror what is happening in __init__
+        if isinstance(path, ut.PurePath):
+            path = str(path)
```

### DIFF
--- a/datalad/distribution/dataset.py
+++ b/datalad/distribution/dataset.py
@@ -36,7 +36,7 @@ from datalad.support.external_versions import external_versions
 from datalad.support.gitrepo import GitRepo
 from datalad.support.gitrepo import InvalidGitRepositoryError
 from datalad.support.gitrepo import NoSuchPathError
-from datalad.support.repo import Flyweight
+from datalad.support.repo import PathBasedFlyweight
 from datalad.support.network import RI
 from datalad.support.exceptions import InvalidAnnexRepositoryError
 
@@ -79,7 +79,7 @@ def resolve_path(path, ds=None):
     return normpath(opj(top_path, path))
 
 
-@add_metaclass(Flyweight)
+@add_metaclass(PathBasedFlyweight)
 class Dataset(object):
     """Representation of a DataLad dataset/repository
 
@@ -104,26 +104,8 @@ class Dataset(object):
     _unique_instances = WeakValueDictionary()
 
     @classmethod
-    def _flyweight_id_from_args(cls, *args, **kwargs):
-
-        if args:
-            # to a certain degree we need to simulate an actual call to __init__
-            # and make sure, passed arguments are fitting:
-            # TODO: Figure out, whether there is a cleaner way to do this in a
-            # generic fashion
-            assert('path' not in kwargs)
-            path = args[0]
-            args = args[1:]
-        elif 'path' in kwargs:
-            path = kwargs.pop('path')
-        else:
-            raise TypeError("__init__() requires argument `path`")
-
-        if path is None:
-            lgr.debug("path is None. args: %s, kwargs: %s", args, kwargs)
-            raise ValueError("path must not be None")
-
-        # Custom handling for few special abbreviations
+    def _flyweight_preproc_path(cls, path):
+        """Custom handling for few special abbreviations for datasets"""
         path_ = path
         if path == '^':
             # get the topmost dataset from current location. Note that 'zsh'
@@ -135,20 +117,7 @@ class Dataset(object):
             path_ = cfg.obtain('datalad.locations.default-dataset')
         if path != path_:
             lgr.debug("Resolved dataset alias %r to path %r", path, path_)
-
-        # Sanity check for argument `path`:
-        # raise if we cannot deal with `path` at all or
-        # if it is not a local thing:
-        path_ = RI(path_).localpath
-
-        # we want an absolute path, but no resolved symlinks
-        if not isabs(path_):
-            path_ = opj(getpwd(), path_)
-
-        # use canonical paths only:
-        path_ = normpath(path_)
-        kwargs['path'] = path_
-        return path_, args, kwargs
+        return path_
     # End Flyweight
 
     def __init__(self, path):
--- a/datalad/distribution/dataset.py
+++ b/datalad/distribution/dataset.py
@@ -39,6 +39,7 @@ from datalad.support.gitrepo import NoSuchPathError
 from datalad.support.repo import PathBasedFlyweight
 from datalad.support.network import RI
 from datalad.support.exceptions import InvalidAnnexRepositoryError
+from datalad.support import path as op
 
 from datalad.utils import assure_unicode
 from datalad.utils import getpwd
@@ -118,6 +119,15 @@ class Dataset(object):
         if path != path_:
             lgr.debug("Resolved dataset alias %r to path %r", path, path_)
         return path_
+
+    @classmethod
+    def _flyweight_postproc_path(cls, path):
+        # we want an absolute path, but no resolved symlinks
+        if not op.isabs(path):
+            path = op.join(op.getpwd(), path)
+
+        # use canonical paths only:
+        return op.normpath(path)
     # End Flyweight
 
     def __init__(self, path):
--- a/datalad/distribution/dataset.py
+++ b/datalad/distribution/dataset.py
@@ -120,7 +120,8 @@ class Dataset(object):
             raise TypeError("__init__() requires argument `path`")
 
         if path is None:
-            raise AttributeError
+            lgr.debug("path is None. args: %s, kwargs: %s", args, kwargs)
+            raise ValueError("path must not be None")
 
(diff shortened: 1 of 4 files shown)


bpoldrack:
>  I have taken implementation from Dataset, which did not explicitly
"realpath" the path. Not sure yet if such difference in behavior (Dataset vs *Repo)
was intended/desired, tests might show

Intended!

bpoldrack:
The `*Repo` classes are supposed and required to be tied to the real path obviously.
`Dataset` however, is intentionally allowed to point to a symlink or a not (yet) existing path. 
 
  - this is consistent with `Dataset` not actually being a `*Repo`, but having an attribute pointing to one
  - it allows to account for having different (sub-)datasets that actually are referencing the very same repo on FS via symlinks, while they themselves are actually being different submodules at different mount points, etc.
  - instantiation of `Dataset` is cheap that way
  - we can have `Dataset`s that don't yet exist. ( think `Dataset(''some/where').create()`)

yarikoptic:
Gotcha, thanks! I will improve to account for such desired difference. As long as we don't have parallel datasets operations we should be good I guess

bpoldrack:
>  As long as we don't have parallel datasets operations we should be good I guess

To an extend we are good even then. Those Datasets would still point to the very same *Repo instance, so that we don't run into inconsistent objects that would need some kind of syncing. While this is by no means sufficient to have real parallelization, at least it gives us the one spot to deal with it by design.

yarikoptic:
> To an extend we are good even then. Those Datasets would still point to the very same *Repo instance, so that we don't run into inconsistent objects that would need some kind of syncing.

yes, but neither our higher level operations are atomic nor even Flywheight code itself . And that is where things could go awry (e.g., one "thread" adds a file, second one commits with irrelevant message after adding some other files etc.; multiple "identical" flyweight instances appear etc).  But yeah -- it is nothing to worry about ATM ;)

yarikoptic:
I have pushed 3f4f9535bed5947e71f8c9a02f4cf42a951201b3 to provide separate logic for "postprocessing" of paths **but** overloading of that `@classmethod` didn't work, and I cannot figure out (may be to blame `@add_metaclass`?) why?
```
$> python -c 'from datalad.api import Dataset; sub1=Dataset("./sub1"); sub2=Dataset("./sub2"); print(sub1); print(sub2)'
> /home/yoh/proj/datalad/datalad/datalad/support/repo.py(213)_flyweight_id_from_args()
-> path_ = cls._flyweight_preproc_path(path)
(Pdb) p cls
<class 'datalad.support.repo.PathBasedFlyweight'>
(Pdb) up
> /home/yoh/proj/datalad/datalad/datalad/support/repo.py(142)__call__()
-> id_, new_args, new_kwargs = cls._flyweight_id_from_args(*args, **kwargs)
(Pdb) p cls
<class 'datalad.distribution.dataset.Dataset'>
(Pdb) p cls._flyweight_postproc_path
<bound method Dataset._flyweight_postproc_path of <class 'datalad.distribution.dataset.Dataset'>>
```
so even though within `repo.py(142)__call__()` `cls` is `Dataset`, and it does have its own `_flyweight_postproc_path`, within `cls._flyweight_id_from_args` call `cls` is generic `PathBasedFlyweight`.

On a simple (non-metaclass) example things work just as they should:
```
$> cat /tmp/ppp.py
class Base(object):
    @classmethod
    def _over(cls):
        print("Base")

    @classmethod
    def call(cls):
        cls._over()

class Der1(Base):
    pass

class Der2(Base):
    @classmethod
    def _over(cls):
        print("Over")

Der1.call()  # Base
Der2.call()  # Over?

$> python /tmp/ppp.py
Base
Over
```

what am I missing?

bpoldrack:
>  (may be to blame @add_metaclass?)
> what am I missing?

Yes, I think, what you're missing is that it's a metaclass. Or rather what a metaclass is. A metaclass is a class for classes. 
You're declaring `classmethods` in `PathBasedFlyweight`. What is that supposed to do? Have a look at `Flyweight` itself. It does not define `_flyweight_id_from_args` as a classmethod, while you're doing that explicitly in `PathBasedFlyweight`. Why is that? Because a method of a metaclass *is* a classmethod of the classes using that metaclass, which is why the override in `GitRepo` for example is using a `@classmethod`. Because `_flyweight_id_from_args` is a method not of its superclass but its metaclass.

So, I think having those methods not declared `@classmethods` within `PathBasedFlyweight`, but only the overrides within `Dataset` should work. An (only quick) check confirms:

Removed `@classmethod` from `PathBasedFlyweight._flyweight_id_from_args` results in:

```
$> python -c 'from datalad.api import Dataset; sub1=Dataset("./sub1"); sub2=Dataset("./sub2"); print(sub1); print(sub2)'
> /home/ben/work/hacking/datalad/datalad/support/repo.py(218)_flyweight_id_from_args()
-> localpath = RI(path_).localpath
(Pdb) p cls
<class 'datalad.distribution.dataset.Dataset'>
(Pdb) up
> /home/ben/work/hacking/datalad/datalad/support/repo.py(142)__call__()
-> id_, new_args, new_kwargs = cls._flyweight_id_from_args(*args, **kwargs)
(Pdb) p cls
<class 'datalad.distribution.dataset.Dataset'>
```

yarikoptic:
right!  Thank you @bpoldrack - I forgot that `PathBasedFlyweight` is a metaclass. 
<details>
<summary> Removed those:  </summary>
```
$> git show
commit 2af2b9c3cc578e15c46fb8dd519cfe9bd140375c (HEAD -> enh-flyweight-dedup)
Date:   Tue Jan 7 09:31:06 2020 -0500

    BF: removing @classmethod from methods of PathBasedFlyweight (which is a metaclass)

diff --git a/datalad/support/repo.py b/datalad/support/repo.py
index d30a187d1..0e4d794ad 100644
--- a/datalad/support/repo.py
+++ b/datalad/support/repo.py
@@ -168,7 +168,6 @@ class Flyweight(type):
 
 class PathBasedFlyweight(Flyweight):
 
-    @classmethod
     def _flyweight_preproc_path(cls, path):
         """perform any desired path preprocessing (e.g., aliases)
 
@@ -176,7 +175,6 @@ class PathBasedFlyweight(Flyweight):
         """
         return path
 
-    @classmethod
     def _flyweight_postproc_path(cls, path):
         """perform any desired path post-processing (e.g., dereferencing etc)
 
@@ -188,7 +186,6 @@ class PathBasedFlyweight(Flyweight):
         # physical repository at a time
         return op.realpath(path)
 
-    @classmethod
     def _flyweight_id_from_args(cls, *args, **kwargs):
 
         if args:
```
</details>
will test locally and push then

yarikoptic:
ok - tests pass (appveyor issue is codecov submission). unfortunately I see no codecov (should be green ;), and I see no immediate benchmark hits: https://travis-ci.org/datalad/datalad/jobs/633807180?utm_medium=notification&utm_source=github_status
```
All benchmarks:
     [7f08f584]       [dbf4f692]
     <bm/merge-target>       <bm/pr>   
         976±20ms       1.01±0.01s     1.03  api.supers.time_createadd
       1.03±0.05s         997±10ms     0.96  api.supers.time_createadd_to_dataset
       2.62±0.08s       2.68±0.07s     1.02  api.supers.time_installr
          112±3ms        112±0.6ms     1.00  api.supers.time_ls
         983±40ms        1000±10ms     1.02  api.supers.time_ls_recursive
       1.12±0.06s       1.12±0.01s     1.00  api.supers.time_ls_recursive_long_all
       1.50±0.01s       1.59±0.01s     1.06  api.supers.time_remove
         304±10ms          312±3ms     1.02  api.testds.time_create_test_dataset1
       2.20±0.03s       2.30±0.02s     1.04  api.testds.time_create_test_dataset2x2
      1.68±0.05ms      1.67±0.02ms     0.99  core.runner.time_echo
      2.24±0.02ms      2.24±0.05ms     1.00  core.runner.time_echo_gitrunner
          512±3ms         520±10ms     1.02  core.startup.time_help_np
          123±2ms          122±3ms     0.99  core.startup.time_import
          459±3ms         474±10ms     1.03  core.startup.time_import_api
```
so should be good to go

bpoldrack:
LGTM.

Do you also take care of getting it into master?

PR #2972, opened 2018-11-02
BF(TST): realpath path before figuring out which repo it belongs to

yarikoptic:
Surprisingly, I ran into it only within https://github.com/datalad/datalad/pull/2898
despite all our testing with symlinked `TMPDIR`s on travis.  This time I have added
a dedicated unittest to trigger that failure.

The rationale for using realpath there is because GitRepo already realpath"ing
the path to be able to return the singleton Repo for the same path

This pull request fixes #2961

Issue #2961, opened 2018-10-30
realpath'ing Repo path -vs- ok_file_under_git


yarikoptic:
Originally ran into within https://github.com/datalad/datalad/pull/2898 ([failed travis run](https://travis-ci.org/datalad/datalad/jobs/442291459)), not sure how it didn't float up before (at least for me):  
in [GitRepo._flyweight_id_from_args](https://github.com/datalad/datalad/blob/master/datalad/support/gitrepo.py#L576) we realpath'ing every path so any repository's path is the realpath (although we do have `.realpath` for that purpose).  But in other spots (like `ok_file_under_git`) we seems to not do that at causes incorrect assessment.  Here are examples in both DataLad proper and the -revolution one:

```shell
$> python -c 'from datalad.support.gitrepo import GitRepo; p="/var/tmp/sym link/datalad_temp_tree_wrappedVlKRSJ/ds-clone"; print(GitRepo(p)); from datalad.tests.utils import ok_file_under_git; ok_file_under_git(p + "/in-annex", annexed=True)' 
<GitRepo path=/tmp/datalad_temp_tree_wrappedVlKRSJ/ds-clone (<class 'datalad.support.gitrepo.GitRepo'>)>
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/yoh/proj/datalad/datalad-master/datalad/tests/utils.py", line 226, in ok_file_under_git
    assert_in(file_repo_path, repo.get_indexed_files())  # file is known to Git
  File "/usr/lib/python2.7/unittest/case.py", line 804, in assertIn
    self.fail(self._formatMessage(msg, standardMsg))
  File "/usr/lib/python2.7/unittest/case.py", line 410, in fail
    raise self.failureException(msg)
AssertionError: '../../../var/tmp/sym link/datalad_temp_tree_wrappedVlKRSJ/ds-clone/in-annex' not found in [u'.gitattributes', u'.datalad/.gitattributes', u'in-annex', u'.datalad/config', u'in-git']
(dev) 1 14949 ->1.....................................:Mon 29 Oct 2018 08:35:53 PM EDT:.
(git)hopa:~datalad/datalad-revolution[master]git
[...]

PR #3991, opened 2020-01-09
Merge 0.11.x:  mimeencoding, flyweight code RF

yarikoptic:
Conflicts were not fun.  but seems to test selected tests I ran locally, so I guess I didn't ruin it.

### DIFF
--- a/datalad/distribution/dataset.py
+++ b/datalad/distribution/dataset.py
@@ -98,7 +98,8 @@ class Dataset(object, metaclass=Flyweight):
             raise TypeError("__init__() requires argument `path`")
 
         if path is None:
-            raise AttributeError
+            lgr.debug("path is None. args: %s, kwargs: %s", args, kwargs)
+            raise ValueError("path must not be None")
 
         # mirror what is happening in __init__
         if isinstance(path, ut.PurePath):
--- a/datalad/distribution/dataset.py
+++ b/datalad/distribution/dataset.py
@@ -10,7 +10,6 @@
 """
 
 import logging
-import os.path as op
 from os.path import (
     curdir,
     exists,
@@ -36,9 +35,10 @@ from datalad.support.gitrepo import (
     InvalidGitRepositoryError,
     NoSuchPathError
 )
-from datalad.support.repo import Flyweight
+from datalad.support.repo import PathBasedFlyweight
 from datalad.support.network import RI
 from datalad.support.exceptions import InvalidAnnexRepositoryError
+from datalad.support import path as op
 
 import datalad.utils as ut
 from datalad.utils import (
@@ -58,7 +58,7 @@ lgr = logging.getLogger('datalad.dataset')
 lgr.log(5, "Importing dataset")
 
 
-class Dataset(object, metaclass=Flyweight):
+class Dataset(object, metaclass=PathBasedFlyweight):
     """Representation of a DataLad dataset/repository
 
     This is the core data type of DataLad: a representation of a dataset.
@@ -82,30 +82,8 @@ class Dataset(object, metaclass=Flyweight):
     _unique_instances = WeakValueDictionary()
 
     @classmethod
-    def _flyweight_id_from_args(cls, *args, **kwargs):
-
-        if args:
-            # to a certain degree we need to simulate an actual call to __init__
-            # and make sure, passed arguments are fitting:
-            # TODO: Figure out, whether there is a cleaner way to do this in a
-            # generic fashion
-            assert('path' not in kwargs)
-            path = args[0]
-            args = args[1:]
-        elif 'path' in kwargs:
-            path = kwargs.pop('path')
-        else:
-            raise TypeError("__init__() requires argument `path`")
-
-        if path is None:
-            lgr.debug("path is None. args: %s, kwargs: %s", args, kwargs)
-            raise ValueError("path must not be None")
-
-        # mirror what is happening in __init__
-        if isinstance(path, ut.PurePath):
-            path = str(path)
-
-        # Custom handling for few special abbreviations
+    def _flyweight_preproc_path(cls, path):
+        """Custom handling for few special abbreviations for datasets"""
         path_ = path
         if path == '^':
             # get the topmost dataset from current location. Note that 'zsh'
@@ -121,20 +99,16 @@ class Dataset(object, metaclass=Flyweight):
             path_ = cfg.obtain('datalad.locations.default-dataset')
         if path != path_:
             lgr.debug("Resolved dataset alias %r to path %r", path, path_)
+        return path_
 
-        # Sanity check for argument `path`:
-        # raise if we cannot deal with `path` at all or
-        # if it is not a local thing:
-        path_ = RI(path_).localpath
-
+    @classmethod
+    def _flyweight_postproc_path(cls, path):
         # we want an absolute path, but no resolved symlinks
-        if not isabs(path_):
-            path_ = opj(getpwd(), path_)
+        if not op.isabs(path):
+            path = op.join(op.getpwd(), path)
 
         # use canonical paths only:
-        path_ = normpath(path_)
-        kwargs['path'] = path_
-        return path_, args, kwargs
+        return op.normpath(path)
 
     def _flyweight_invalid(self):
         """Invalidation of Flyweight instance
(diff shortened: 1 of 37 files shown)


kyleam:
> Conflicts were not fun.

Thanks for handling it.  I bet I'll relive your fun as I review it.

yarikoptic:
well -- the ultimate diff here is quite easier to digest than 3way diff of merge conflicts which were there ;)
FWIW: not sure what's up with coverage

datalad/distribution/dataset.py
```diff
@@ -121,29 +100,25 @@ def _flyweight_id_from_args(cls, *args, **kwargs):
...
-        kwargs['path'] = path_
-        return path_, args, kwargs
+        return op.normpath(path)
 
     def _flyweight_invalid(self):
         """Invalidation of Flyweight instance
 
-        Dataset doesn't need to be invalidated during its lifetime at all. Instead the underlying *Repo instances are.
+        Dataset doesn't need to be invalidated during its lifetime at all.
+        Instead the underlying *Repo instances are.
```
kyleam: When dealing with merge conflicts, please avoid this sort of unrelated formatting.  It introduces changes that aren't shared by any side, which are typically the changes of most interest, but here they're just noise.

datalad/distribution/dataset.py
```diff
@@ -36,9 +36,10 @@
     InvalidGitRepositoryError,
     NoSuchPathError
 )
-from datalad.support.repo import Flyweight
+from datalad.support.repo import PathBasedFlyweight
 from datalad.support.network import RI
 from datalad.support.exceptions import InvalidAnnexRepositoryError
+from datalad.support import path as op
```
kyleam: This is introducing a duplicate import of `op`.

datalad/interface/tests/test_run_procedure.py
```diff
@@ -304,17 +305,52 @@ def test_quoting(path):
...
+    'emptyline': '\n',  # libmagic: "binary" "application/octet-stream"
+})
+def test_text2git(path):
+    # Test if files being correctly annexed in a ds configured with text2git.
+    TEXT_FILES = ('JSON', 'YAML', 'MARKDOWN', 'empty')
+    BINARY_FILES = ('0blob', 'emptyline')
+
     ds = Dataset(path).create(force=True)
     ds.run_procedure('cfg_text2git')
+
```
kyleam: This space is another change that isn't from either side.

datalad/interface/tests/test_run_procedure.py
```diff
@@ -304,17 +305,52 @@ def test_quoting(path):
...
-    ds.save(message="add empty file")
-    # check that it's not annexed
-    assert_false(ds.repo.is_under_annex("emptyfile"))
+
+    # check that text files are not annexed
+    for f in TEXT_FILES:
+        assert_false(ds.repo.is_under_annex(f))
+    # and trivial binaries - annexed
+    for f in BINARY_FILES:
+        assert_true(ds.repo.is_under_annex(f))
```
kyleam: Hmm, da27e86c4 (part of gh-3956) dropped the `assert_false(ds.repo.is_under_annex("emptyfile"))` above for these `ok_file_under_git` calls.  Why aren't you keeping that change here?

datalad/distribution/dataset.py
yarikoptic: not 100% certain I did it consciously ... generally I am trying to avoid such changes while working on merges.  do you think it is worth rewriting for this one?

datalad/interface/tests/test_run_procedure.py
yarikoptic: this whole test was pretty much rewritten from being a test for only an empty file, to a test which tests various types (text and "binary") of files -- from `test_text2git_empty` to `test_text2git`. So `diff` here could be a bit misleading - I would not strive to minimize diff here, nor re-adding `assert_false(ds.repo.is_under_annex("emptyfile"))` would add any additional testing

datalad/interface/tests/test_run_procedure.py
yarikoptic: see comment below

datalad/distribution/dataset.py
kyleam: That'd be my personal preference, but given what I perceive to be the preference of this project in general, I'd say "no, not if this were the only issue".  But if you agree that the `ok_file_under_git` point below should be addressed anyway, then I think it'd be good to resolve these minor issues in the same rewrite.

datalad/distribution/dataset.py
yarikoptic: rright -- removing `import os.path as op` above. pushed 37e9a9e05599b271cc5743f3cb0155babcf9c197

datalad/interface/tests/test_run_procedure.py
kyleam: Let me try again.  Here is the resolution that you pushed:

```
+     ds.save(path=TEXT_FILES + BINARY_FILES, message="added all files")
 -    ok_clean_git(ds.path)
 +    assert_repo_status(ds.path)
-     # create an empty file, no extension
-     open(op.join(path, 'emptyfile'), 'a').close()
-     ds.save(message="add empty file")
-     # check that it's not annexed
-     assert_false(ds.repo.is_under_annex("emptyfile"))
+ 
+     # check that text files are not annexed
+     for f in TEXT_FILES:
 -        ok_file_under_git(path, f, annexed=False)
++        assert_false(ds.repo.is_under_annex(f))
+     # and trivial binaries - annexed
+     for f in BINARY_FILES:
 -        ok_file_under_git(path, f, annexed=True)
++        assert_true(ds.repo.is_under_annex(f))
```

Note that you're introducing two changes (the `++` liines) that come from neither side.  Unless there is a master-specific reason for why doing that, you should not.

datalad/interface/tests/test_run_procedure.py
kyleam: I don't see how the comment below is relevant.  Even if the test was rewritten on 0.11.x's side (specifically da27e86), the resolution shouldn't introduce a space that wasn't in 0.11.x or in master.

datalad/interface/tests/test_run_procedure.py
yarikoptic: hm, IIRC I was just replacing "0.11.x" style (`ok_file_under_git(path, f, annexed=False)`) from one side to "master style" (`assert_false(ds.repo.is_under_annex(f))`).  What should have I done differently?

datalad/interface/tests/test_run_procedure.py
yarikoptic: oh, gotcha. True!  Let's see if we agree on the below and then I could rewrite that merge to make it clearer (to avoid possible future conflicts)... or you are most welcome to do so and force push here!

kyleam: > you are most welcome to do so and force push here!

OK, I'll do that.

datalad/interface/tests/test_run_procedure.py
kyleam: > I was just replacing "0.11.x" style (`ok_file_under_git(path, f, annexed=False)`) from one side to "master style" (`assert_false(ds.repo.is_under_annex(f))`)

Thanks, knowing your intention here helps.

Given that the `assert_false(ds.repo.is_under_annex(f))` line was shared with 0.11.x, and you just did the reverse substitution in 0.11.x's da27e86, in my view the commit message should explain why we should prefer one form on master and the other on 0.11.x, ideally with a pointer to where we've discussed `ok_file_under_git` being problematic.  The main thing I can find is gh-2961, but it looks like that was resolved by your gh-2972.

Even more ideally, if `ok_file_under_git` is still an issue and should be avoided on master, we'd also avoid it on 0.11.x, because (1) it seems likely that the same issues would be relevant for 0.11.x and (2) it introduces needless differences when the master variant will work fine on 0.11.x.

I don't think it is worth a PR to 0.11.x at this point, so I'll go with adding something to the commit message to the merge that I force-push here.

yarikoptic: thanks for the analysis! I just saw difference, so assumed that it would be the right way to do in master without checking compatibility with 0.11.x .  Please do as you see it fit! Thanks in advance!

kyleam:
AppVeyor failure is the familiar "Invoke-WebRequest : Unable to connect" failure.  Will merge.

PR #3996, opened 2020-01-09
BF: update -- do not save if no --merge, use save instead of Dataset(refds_path).add

### LINKED ISSUE #3097: datalad update -r causes _flyweight_id_from_args crash on path being None when '.' path is provided
Details on how to get there are yet to be discovered, but original issue is here: https://github.com/templateflow/templateflow/issues/5 

code in question is Dataset's:

```
    @classmethod
    def _flyweight_id_from_args(cls, *args, **kwargs):

        if args:
            # to a certain degree we need to simulate an actual call to __init__
            # and make sure, passed arguments are fitting:
            # TODO: Figure out, whether there is a cleaner way to do this in a
            # generic fashion
            assert('path' not in kwargs)
            path = args[0]
            args = args[1:]
        elif 'path' in kwargs:
            path = kwargs.pop('path')
        else:
            raise TypeError("__init__() requires argument `path`")

        if path is None:
            raise AttributeError
```

yarikoptic:
`refds_path` could be empty if dataset is not specified but path (to subdataset) is provided

Note that it would change existing behavior -- apparently we could have been saving some prior changes if there were some before running `update`.  I would consider that also to be a bug fix

Fixes #3097 
ref: https://github.com/templateflow/templateflow/issues/5

TODOs (after merge)
- merge/adjust for `master` way of things

### DIFF
--- a/datalad/distribution/tests/test_update.py
+++ b/datalad/distribution/tests/test_update.py
@@ -104,6 +104,19 @@ def test_update_simple(origin, src_path, dst_path):
     dest.repo.get_file_key("update.txt")  # raises if unknown
     eq_([False], dest.repo.file_has_content(["update.txt"]))
 
+    # test that update doesn't crash if we specify only a single path (submod) to
+    # operate on
+    with chpwd(dest.path):
+        assert_result_count(
+            update(path=['subm 1'], recursive=True), 1,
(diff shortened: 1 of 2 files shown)


bpoldrack:
LGTM

PR #3998, opened 2020-01-09
DOC: Automatically add current year to copyright notice

adswa:
I noticed that the copyright notice on [docs.datalad.org](http://docs.datalad.org/en/latest/index.html) says 2016-2019. This change automatically adds the current year.

While I did this, I noticed an unused import (shlex), and removed it as well.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -15,6 +15,7 @@
 import sys
 import os
 import shlex
+import datetime
 from os.path import join as opj, exists
 from os import pardir
 from glob import glob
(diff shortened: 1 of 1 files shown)


docs/source/conf.py
```diff
@@ -95,8 +95,9 @@ def setup(sphinx):
 master_doc = 'index'
 
 # General information about the project.
+current_year = datetime.datetime.now().year
 project = u'DataLad'
-copyright = u'2016-2019, DataLad team'
+copyright = (u'2016-{}, DataLad team').format(current_year)
```
yarikoptic: god bless Python-based "configuration" files! ;)

Issue #3997, opened 2020-01-09
diff behavior is dependent on checkout branch


mih:
Demo

```
(git)-[incoming-processed] % datalad diff -f incoming~1 -t incoming
  deleted: .ukbbatch (file)
  deleted: 5874415_20249_2_0.zip (file)
  deleted: 5874415_20227_2_0.zip (file)
(git)-[incoming-processed] % git checkout incoming
Switched to branch 'incoming'
(git)-[incoming] % datalad diff -f incoming~1 -t incoming
  deleted: 5874415_20227_2_0.zip (file)
```

Only the latter result is factually correct. It seems the pattern is whenever the worktree doesn't match the queries branch, results are wrong.

The reason is an inappropriate test in GitRepo._diffstatus() that uses that worktree as a reference, even when the `to` state to compare against is not None (i.e. the worktree).

This is a likely source of a range of diff failures on windows, were, due to the presence of adjusted branches, we are always diffing a branch that is not checked out when using names like `master` as the state ID.

mih:
Fix was merged.

PR #3999, opened 2020-01-09
BF: gh-3997, never use the worktree as reference when diffing against a commit

mih:
Story in gh-3997

### DIFF
--- a/datalad/core/local/tests/test_diff.py
+++ b/datalad/core/local/tests/test_diff.py
@@ -481,3 +481,31 @@ def test_diff_nonexistent_ref_unicode(path):
         1,
         path=ds.path,
         status="impossible")
+
+
+# https://github.com/datalad/datalad/issues/3997
+@with_tempfile(mkdir=True)
+def test_no_worktree_impact_false_deletions(path):
+    ds = Dataset(path).create()
+    # create a branch that has no new content
+    ds.repo.call_git(['checkout', '-b', 'test'])
+    # place to successive commits with file additions into the master branch
+    ds.repo.call_git(['checkout', 'master'])
+    (ds.pathobj / 'identical').write_text('should be')
+    ds.save()
+    (ds.pathobj / 'new').write_text('yes')
+    ds.save()
+    # now perform a diff for the last commit, there is one file that remained
+    # identifical
+    ds.repo.call_git(['checkout', 'test'])
+    res = ds.diff(fr='master~1', to='master')
+    # under no circumstances can there be any reports on deleted files
+    # because we never deleted anything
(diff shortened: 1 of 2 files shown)


datalad/core/local/tests/test_diff.py
```diff
@@ -481,3 +481,31 @@ def test_diff_nonexistent_ref_unicode(path):
...
+    # now perform a diff for the last commit, there is one file that remained
+    # identifical
+    ds.repo.call_git(['checkout', 'test'])
+    res = ds.diff(fr='master~1', to='master')
+    # under no circumstances can there be any reports on deleted files
+    # because we never deleted anything
+    assert_result_count(res, 0, state='deleted')
+    # the identical file must be reported clean
+    assert_result_count(
+        ds.diff(fr='master~1', to='master'),
```
kyleam: Nit: You could reuse `res` from above.

mih:
This must go into 0.12, hence quick feedback is appreciated.

datalad/support/gitrepo.py
```diff
@@ -3546,8 +3546,8 @@ def _get_cache_key(label, paths, ref, untracked=None):
                 if to_state_r['type'] != 'dataset':
                     # no change in git record, and no change on disk
                     props = dict(
-                        state='clean' if f.exists() or \
-                              f.is_symlink() else 'deleted',
+                        state='clean' if to is not None or (f.exists() or \
+                              f.is_symlink()) else 'deleted',
```
kyleam: Okay, so at this point we know that the reported object ids for this file are identical in the `to` and `from` records.  If `to` is None, we're comparing to the working tree and a deleted file will still have an identical id, so we need to check whether the file is gone before declaring it clean.  This working tree check is irrelevant and wrong if `to` is a ref.

Makes sense.

datalad/core/local/tests/test_diff.py
mih: A true... spedd accuracy trade-off ;-)

datalad/support/gitrepo.py
mih: Thx, that is indeed the reasoning. I will take the liberty and place it into a comment ;-)

kyleam: Ah, no need!   I was just working through it.

kyleam: But I see you already did it, which is also ok :)

mih: Done nevertheless!

datalad/core/local/tests/test_diff.py
mih: Done!

PR #4002, opened 2020-01-09
BF: do not add [Default: ...] for --help if it is a flag (action="store_...")

### LINKED ISSUE #3203: download-url --nosave documentation contradicts itself
The `--nosave` option documentation at http://docs.datalad.org/en/latest/generated/man/datalad-download-url.html reads:

```
by default all modifications to a dataset are immediately saved. Giving this option will disable this behavior. [Default: True]
```

The English text and the parenthesis contradict each other.

yarikoptic:
Closes #3203 

Or there could be a flag without `action="store_..."`, or some `store_` wouldn't be a flag?

I  found no other conveniently available unittesting setup to test this change.  Any pointers or assistance would be appreciated

### DIFF
--- a/datalad/interface/base.py
+++ b/datalad/interface/base.py
@@ -574,7 +574,9 @@ class Interface(object):
                 # showing the Default: (likely boolean).
                 #   See https://github.com/datalad/datalad/issues/3203
                 if not parser_kwargs.get('action', '').startswith('store_'):
-                    help += " [Default: %r]" % (defaults[defaults_idx],)
+                    # [Default: None] also makes little sense for cmdline
+                    if defaults[defaults_idx] is not None:
+                        help += " [Default: %r]" % (defaults[defaults_idx],)
             # create the parameter, using the constraint instance for type
             # conversion
             parser.add_argument(*parser_args, help=help,
--- a/datalad/interface/base.py
+++ b/datalad/interface/base.py
@@ -570,7 +570,11 @@ class Interface(object):
                     cdoc = cdoc[1:-1]
                 help += '  Constraints: %s' % cdoc
             if defaults_idx >= 0:
-                help += " [Default: %r]" % (defaults[defaults_idx],)
+                # if it is a flag, in commandline it makes little sense to show
+                # showing the Default: (likely boolean).
+                #   See https://github.com/datalad/datalad/issues/3203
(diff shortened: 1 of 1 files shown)


yarikoptic:
while at it expanded cleansing in eed4ddea15d8edc32d30bf238828ec6408cacc35

    ENH: also do not include misguiding [Default: None]
    
    "is None a string which is considered by default?"  would be a logical
    question of any cmdline user who is not aware that DataLad is written in
    Python

datalad/interface/base.py
```diff
@@ -570,7 +570,13 @@ def setup_parser(cls, parser):
                     cdoc = cdoc[1:-1]
                 help += '  Constraints: %s' % cdoc
             if defaults_idx >= 0:
-                help += " [Default: %r]" % (defaults[defaults_idx],)
+                # if it is a flag, in commandline it makes little sense to show
+                # showing the Default: (likely boolean).
+                #   See https://github.com/datalad/datalad/issues/3203
+                if not parser_kwargs.get('action', '').startswith('store_'):
+                    # [Default: None] also makes little sense for cmdline
```
bpoldrack: The argument for `[Default: None]` being confusing for command line holds true for non-flag options as well or am I missing something?

bpoldrack: BTW: Same argument is true for `True` and `False` in general. It's not (necessarily) referring to those strings.

yarikoptic: hm, ... what if there is a `choice` allowing for bool and some strings, and default is a bool?  Note: This block is executed only for non `store_`, so all flags are excluded already by here, but IMHO it shouldn't be generalized solely based on a default value

yarikoptic: but now I am getting a better clue on what you meant ;)  indeed it is unlikely we have `bool` specified as a value (e.g. as `True`) in command line... we do though have `assure_bool` etc handling which was/is used by crawler, but I am not sure we have anything like that for command line options.  I would though probably not dive into deep analysis at this point -- could be improved later if we run into such odd places.

yarikoptic:
Good point Kyle. I by default left fixes statements to pr. Should change my habits.
If no further comments/objections, let's merge?

yarikoptic:
ok, no objections were expressed, so let's do it

PR #4004, opened 2020-01-10
BF: Maintain path arg semantics when passing from update-> subdatasets

mih:
It is sufficient to pass a dataset arg as-in to achieve that.


mih:
Fix is trivial, time is pressing, dedicated test passes: merge without review.

Issue #2566, opened 2018-05-27
`find_files` broken on Windows


mih:
```
[00:15:30] ======================================================================
[00:15:30] FAIL: datalad.tests.test_utils.test_find_files
[00:15:30] ----------------------------------------------------------------------
[00:15:30] Traceback (most recent call last):
[00:15:30]   File "C:\Miniconda35\envs\test-environment\lib\site-packages\nose\case.py", line 198, in runTest
[00:15:30]     self.test(*self.arg)
[00:15:30]   File "C:\Miniconda35\envs\test-environment\lib\site-packages\datalad\tests\test_utils.py", line 482, in test_find_files
[00:15:30]     assert_in(opj(tests_dir, 'test_utils.py'), files3)
[00:15:30] AssertionError: 'C:\\Miniconda35\\envs\\test-environment\\lib\\site-packages\\datalad\\tests\\test_utils.py' not found in []
[00:15:30] 
[00:15:30] ======================================================================
[00:15:30] FAIL: datalad.tests.test_utils.test_find_files_exclude_vcs
[00:15:30] ----------------------------------------------------------------------
[00:15:30] Traceback (most recent call last):
[00:15:30]   File "C:\Miniconda35\envs\test-environment\lib\site-packages\nose\case.py", line 198, in runTest
[00:15:30]     self.test(*self.arg)
[00:15:30]   File "C:\Miniconda35\envs\test-environment\lib\site-packages\datalad\tests\utils.py", line 499, in newfunc
[00:15:30]     return t(*(arg + (d,)), **kw)
[00:15:30]   File "C:\Miniconda35\envs\test-environment\lib\site-packages\datalad\tests\test_utils.py", line 500, in test_find_files_exclude_vcs
[00:15:30]     assert_equal({basename(f) for f in files}, {'d1', 'git'})
[00:15:30] AssertionError: Items in the first set but not the second:
[00:15:30] '.git'
[00:15:30] '1'
```

mih:
Still the case today.

PR #4013, opened 2020-01-11
BF: Properly handle directory separators on Windows (fixes gh-2566)

mih:


datalad/tests/test_utils.py
```diff
@@ -1139,6 +1140,8 @@ def f(j):
 
 @with_tempfile(mkdir=True)
 def test_dlabspath(path):
+    if not has_symlink_capability():
```
yarikoptic: Ideally we should just make it into a decorator to stay consistent?

mih: I have mixed feelings re decorators. They definitely shouldn't be the only method. I have RF'ed so many tests recently that could run almost in full under all conditions, but had a single conditional that would have needed scenario-specific treatment, but only a decorator was available, hence I had to disable the full test...

Issue #2109, opened 2018-01-27
install a specific version of a dataset


yarikoptic:
#### What is the problem?
ATM we do not have AFAIK datalad way to install a specific version of a dataset. Although whenver it is a git-annex repo in indirect or v6 mode it should be trivial, direct mode makes it not so trivial, and with side-effects (just tried git annex proxy call and was left with bunch  of empty directories for the previous version)

I though I had somewhere an issue for this but could not find.
It is a common convention to use `@` in url to point to a specific commitish, but I wonder if we should make it more explicit via url option to be handled on the client side (e.g., `///path/repo#version=commitish` and correspondingly with a regular url)

related: #1671 - recursive checkout/merge

yarikoptic:
anyone has ideas/recommendations, or should I just kick out initial version as a PR and we would decide from there?

mih:
re `@` vs a url parameter: I think the key question is what we would want to encode that is equally important as the version (because I think the `@` makes for a rather intuitive URI.

In the worst case, we could treat the entire thing as the URL and still parametrize it, e.g.

`///openfmri/ds000113b@v1.2?unlock=all`

bpoldrack:
+1 for `@`

I think giving a commitish isn't really a parameter, that might control what to do with something, but rather part of specifying the thing it is about. Plus, it is an existing convention.

yarikoptic:
> @

I still have a filling that we should avoid it since it is "reserved" and used for specific purpose(s).  It is used as a delimeter of user from address (in email), but it is not even used as a delimiter for the "end of path" (https://tools.ietf.org/html/rfc3986), so
```
In [13]: urlparse("http://datalad.org/sub1@1.0.0#get")
Out[13]: ParseResult(scheme='http', netloc='datalad.org', path='/sub1@1.0.0', params='', query='', fragment='get')
```
we leak into path, and more fun:

```
In [11]: URL("http://datalad.org@1.0.0#get")
Out[11]: URL(fragment='get', hostname='1.0.0', scheme='http', username='datalad.org')
```
- we made it completely obscure (comprehends as a user at 1.0.0 domain).

> ?unlock=all

`?` is usually for "server (or local JS code talking to that remote server) side" options.  Not the case here -- it is also not about really identifying some "query" (what it is for: https://tools.ietf.org/html/rfc3986#section-3.4).

[...]

PR #4022, opened 2020-01-13
Towards supporting access to RIA stores in clone()

mih:
- [x] generalize existing URL normalization function to a helper to decode more it less arbitrary specifications to git-compatible clone url plus any number of additional properties
- [x] add support for `ria+<protocol>://` URIs
- [x] populated `store.datalad.org` with a test dataset using https://github.com/datalad/git-annex-ria-remote/pull/38. Requires the client-side setup described below and then:
   - `datalad create test`
   - `datalad -d test create-sibling-ria storedataladorg`

   Result now sitting at `http://store.datalad.org/76b/6ca66-36b1-11ea-a2e6-f0d5bf7b5561`
- [x] make `datalad clone ria+http://store.datalad.org#76b6ca66-36b1-11ea-a2e6-f0d5bf7b5561` work
   - for now I had to manually run `git update-server-info` on the server, would a smart http setup on the server help @yarikoptic ?
- [x] make `datalad clone ria+ssh://store.datalad.org#76b6ca66-36b1-11ea-a2e6-f0d5bf7b5561` work
  - only works with additional client-side configuration (see below), because the actual location on the server cannot be inferred from the identifier. This is not a problem, because only people with write access will need to have this, and they already need that for `ria-remote`.
- [x] add tests against `store.datalad.org`
   - [x] clone from http
   - [x] verify required post-clone configuration is in place
- [x] enhance the clone source specification decoding to also suggest a destination path.Often more clever names will be possible then just the basename derived from the final git-clone url
- [x] describe `ria+http|ssh://` URIs in docs
- [x] check that `ria+http` logic can deal with URLs were the store needs a path in addition to the hostname, .e.g. `example.com/mystore#dsid`

### DIFF
--- a/datalad/core/distributed/clone.py
+++ b/datalad/core/distributed/clone.py
@@ -734,20 +734,6 @@ def decode_source_spec(spec, cfg=None):
         props['giturl'] = source_ri.as_git_url()
     elif isinstance(source_ri, URL) and source_ri.scheme.startswith('ria+'):
         # parse a RIA URI
-        # check for store related configuration for SSH access
-        if source_ri.scheme == 'ria+ssh':
-            # we support the scenario were hostname is just a label for a config
-            # not an actual hostname
-            hostname = cfg.get(
-                'annex.ria-remote.{}.ssh-host'.format(source_ri.hostname),
-                source_ri.hostname)
-            basepath = cfg.get(
-                'annex.ria-remote.{}.base-path'.format(source_ri.hostname),
-                '')
-        else:
-            hostname = source_ri.hostname
-            basepath = ''
-
         dsid, version = source_ri.fragment.split('@', maxsplit=1) \
             if '@' in source_ri.fragment else (source_ri.fragment, None)
         uuid_regex = r'[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}'
@@ -760,11 +746,9 @@ def decode_source_spec(spec, cfg=None):
         # strip the custom protocol and go with standard one
         source_ri.scheme = source_ri.scheme[4:]
         # take any existing path, and add trace to dataset within the store
-        source_ri.path = '{urlpath}{urldelim}{basepath}{basedelim}{trace}'.format(
+        source_ri.path = '{urlpath}{urldelim}{trace}'.format(
             urlpath=source_ri.path if source_ri.path else '',
             urldelim='' if not source_ri.path or source_ri.path.endswith('/') else '/',
-            basepath=basepath,
-            basedelim='' if not basepath or basepath.endswith('/') else '/',
             trace='{}/{}'.format(dsid[:3], dsid[3:]),
         )
         props.update(
--- a/datalad/core/distributed/clone.py
+++ b/datalad/core/distributed/clone.py
@@ -214,7 +214,7 @@ class Clone(Interface):
             action='install',
             logger=lgr,
             refds=refds_path,
-            source_url=source['giturl'])
+            source_url=source['source'])
 
         try:
             # this will implicitly cause pathlib to run a bunch of checks
--- a/datalad/core/distributed/clone.py
+++ b/datalad/core/distributed/clone.py
@@ -675,6 +675,9 @@ def decode_source_spec(spec):
         # parse a RIA URI
         dsid, version = source_ri.fragment.split('@', maxsplit=1) \
             if '@' in source_ri.fragment else (source_ri.fragment, None)
+        uuid_regex = r'[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}'
+        if not re.match(uuid_regex, dsid):
+            raise ValueError('RIA URI does not contain a valid dataset ID: {}'.format(spec))
         props.update(
             type='ria',
             giturl='{}://{}{}{}/{}'.format(
--- a/datalad/core/distributed/clone.py
+++ b/datalad/core/distributed/clone.py
@@ -16,7 +16,6 @@ from os.path import expanduser
 from collections import OrderedDict
 from urllib.parse import unquote as urlunquote
 
-from datalad import cfg as dlcfg
 from datalad.interface.base import Interface
 from datalad.interface.utils import eval_results
 from datalad.interface.base import build_doc
@@ -194,7 +193,7 @@ class Clone(Interface):
         # Possibly do conversion from source into a git-friendly url
         # luckily GitRepo will undo any fancy file:/// url to make use of Git's
         # optimization for local clones....
-        source_ = decode_source_spec(source)
+        source_ = decode_source_spec(source, cfg=None if ds is None else ds.config)
         lgr.debug("Resolved clone source from '%s' to '%s'",
                   source, source_)
         source = source_
@@ -639,13 +638,17 @@ def _get_installationpath_from_url(url):
     return path
 
 
-def decode_source_spec(spec):
+def decode_source_spec(spec, cfg=None):
     """Decode information from a clone source specification
 
     Parameters
     ----------
     spec : str
       Any supported clone source specification
+    cfg : ConfigManager, optional
+      Configuration will be queried from the instance (i.e. from a particular
+      dataset). If None is given, the global DataLad configuration will be
+      queried.
 
     Returns
     -------
@@ -658,6 +661,8 @@ def decode_source_spec(spec):
       (None else); 'default_destpath' a relative path that that can be used as
       a clone destination.
     """
+    if cfg is None:
+        from datalad import cfg
     # standard property dict composition
     props = dict(
         source=spec,
@@ -676,10 +681,10 @@ def decode_source_spec(spec):
         # parse a RIA URI
         # check for store related configuration for SSH access
         if source_ri.scheme == 'ria+ssh':
-            hostname = dlcfg.get(
+            hostname = cfg.get(
                 'annex.ria-remote.{}.ssh-host'.format(source_ri.hostname),
                 source_ri.hostname)
-            basepath = dlcfg.get(
+            basepath = cfg.get(
                 'annex.ria-remote.{}.base-path'.format(source_ri.hostname),
                 '')
         else:
--- a/datalad/core/distributed/clone.py
+++ b/datalad/core/distributed/clone.py
@@ -190,20 +190,18 @@ class Clone(Interface):
         if path is not None:
             path = resolve_path(path, dataset)
 
-        # Possibly do conversion from source into a git-friendly url
-        # luckily GitRepo will undo any fancy file:/// url to make use of Git's
-        # optimization for local clones....
-        source_ = decode_source_spec(source, cfg=None if ds is None else ds.config)
-        lgr.debug("Resolved clone source from '%s' to '%s'",
-                  source, source_)
-        source = source_
-
         # derive target from source:
         if path is None:
             # we got nothing but a source. do something similar to git clone
             # and derive the path from the source and continue
             # since this is a relative `path`, resolve it:
-            path = resolve_path(source['default_destpath'], dataset)
+            # we are not going to reuse the decoded URL, as this is done for
+            # all source candidates in clone_dataset(), we just use to determine
+            # a destination path here in order to perform a bunch of additional
+            # checks that shall not pollute the helper function
+            source_ = decode_source_spec(
+                source, cfg=None if ds is None else ds.config)
+            path = resolve_path(source_['default_destpath'], dataset)
             lgr.debug("Determined clone target path from source")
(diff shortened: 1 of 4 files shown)


datalad/core/distributed/clone.py
```diff
@@ -636,3 +638,76 @@ def _get_installationpath_from_url(url):
...
+                source_ri.hostname)
+            basepath = dlcfg.get(
+                'annex.ria-remote.{}.base-path'.format(source_ri.hostname),
+                '')
+        else:
+            hostname = source_ri.hostname
+            basepath = ''
+
+        dsid, version = source_ri.fragment.split('@', maxsplit=1) \
+            if '@' in source_ri.fragment else (source_ri.fragment, None)
```
yarikoptic: my reservations for use of `@` remain: https://github.com/datalad/datalad/issues/2109#issuecomment-573678798  Delaying coming up with generally applicable "schema" for URLs would only bring pains (or even prevent) down the road.

bpoldrack: But here the `@` is used withn `fragment`, so _after_ `#`.

yarikoptic: rright, and then for other types of URLs we would need to place it into a fragment explicitly as `http://host/path#@version` even when there is nothing else to go into fragment. 
It might work - we would need anyways to generally urlencode all values for the keys in fragment, so `@` or `;` in e.g. email address or even a path (if present among keys in the fragment) shouldn't end up seen in URL, confusing our parser.
Also having fragment for ria as `#datasetid` makes it probably ria specific -- so we would need per url backend handling or just in general decide that non-keyed element in fragment is the `dataset id`.  So, general "schema" would then be
`#[datasetid][;key=value]*[@version]`
although probably `datasetid` wouldn't be handled by majority of "url backends"?

yarikoptic: note though, that IIRC `@` is not officially an allowed character within fragment
https://tools.ietf.org/html/rfc3986#section-3.5 :
```
      fragment    = *( pchar / "/" / "?" )
```
but above it says:
```
A fragment identifier component is indicated by the presence of a
number sign ("#") character and terminated by the end of the URI.
```
so it seems that splitting it out from a url is unambiguous and having `@` in it shouldn't confuse any system.

datalad/core/distributed/clone.py
```diff
@@ -636,3 +638,76 @@ def _get_installationpath_from_url(url):
...
+    # scenario switch, each case must set 'giturl' at the very minimum
+    if isinstance(source_ri, DataLadRI):
+        # we have got our DataLadRI as the source, so expand it
+        props['type'] = 'dataladri'
+        props['giturl'] = source_ri.as_git_url()
+    elif isinstance(source_ri, URL) and source_ri.scheme.startswith('ria+'):
+        # parse a RIA URI
+        # check for store related configuration for SSH access
+        if source_ri.scheme == 'ria+ssh':
+            hostname = dlcfg.get(
```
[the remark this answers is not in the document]
mih: True, will address. Thx!

mih:
I concur with the general sentiment here that this is not a (or the) general solution to all URL setups, but it should be a meaningful setup for this particular use case.

I am adding a bunch of additional TODOs to the top comment.

datalad/core/distributed/clone.py
mih: Cool, thanks for the detailed assessment!

datalad/core/distributed/clone.py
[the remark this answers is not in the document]
mih: Done

datalad/core/distributed/tests/test_clone.py
```diff
@@ -499,3 +506,64 @@ def test_relative_submodule_url(path):
...
+    assert_raises(ValueError, decode_source_spec, 'ria+http://example.com#123')
+
+
+@skip_if_no_network
+@with_tempfile()
+def test_ria_http(path):
+    # can we clone from the store w/o any dedicated config
+    ds = clone('ria+http://store.datalad.org#{}'.format(datalad_store_testds_id), path)
+    ok_(ds.is_installed())
+    eq_(ds.id, datalad_store_testds_id)
```
bpoldrack: - special remote needs to be autoenabled
- origin needs annex-ignore
- ideally we should probably set publish-depends for origin
- test availability via whereis?

bpoldrack: Re availability: May be test, that fsck doesn't spit out some (fixed loaction log ...)?

bpoldrack: - `get` needs work. 
- May be also include some annexed content in archive?

mih: Thx!

I will not test `get`. A get would require `ria-remote`. And we can test a get in the ria-remote tests for the main store. The test dataset already contains an annexed file.

If we decide to suck ria-remote into datalad-core, adding a test would be trivial, and can be done

bpoldrack: > The test dataset already contains an annexed file.

Yes, but we should prob. put an archive in there as well. However, we can do that within git-annex-ria-remote tests for now - agreed.

mih: That would require SSH access to the main store. Not a good idea, IMHO

mih: I have added a new test that goes through all the steps for a dataset with a subdataset and ensures a proper setup for both

mih:
From my POV this is reasonably complete, and ready for review/merge. It enables full and convenient access to RIA stores for any clone-related functionality.

A next step (in a separeate PR) would be to decide on and implement https://github.com/datalad/datalad/issues/4024

Once done, `get()`-related functionality should be tested for RIA clones, too, as suggested by @bpoldrack  above. The test included here can easily be extended to do that.

bpoldrack:
I think, this approach lacks one aspect of the current setup in `git-annex-ria-remote` with its `ria-post-intall` procedure: The configuration to setup subdatasets in the same reckless/ephemeral mode as the superdataset. Since this is not an information from within the URL but a parameter to `Clone` itself, we would still need such a procedure (which might be connected to the result hooks by `postclonecfg_annexdataset` now).

mih:
True, but I see this as a something different that only has marginally to do with RIA. It is a reckless mode for any local annex that uses hashdir mixed. Hence, I refer to handle that separately to avoid overfitting, and to keep things orthogonal.

bpoldrack:
>  Hence, I prefer to handle that separately to avoid overfitting, and to keep things orthogonal.

Yes, makes sense. Just want to make sure, we are aware.
Otherwise this LGTM.

datalad/core/distributed/clone.py
```diff
@@ -636,3 +691,101 @@ def _get_installationpath_from_url(url):
...
+            giturl=str(source_ri),
+            version=version,
+            default_destpath=dsid,
+        )
+    else:
+        # let's assume that anything else is a URI that Git can handle
+        # TODO are there any advantages of going through a URL-reconversion
+        # by the RI instance, instead of using the original input right away?
+        # it feels like we could only accumulate bugs. Keeping how it was,
+        # for now...
```
kyleam: I agree that it'd be better to use the original value.  This is particularly important because git supports flexible sources via `gitremote-helper`.

By the way, in the context of the call's discussion today on whether "ria" (or whatever name) should be in the scheme part or in the fragment, in my view `gitremote-helper` provides two related points in favor of putting it in the scheme: (1) to be consistent with what git does (`<transport>://<address>` or `<transport>::<address>`) and (2) it keeps open the possibility of eventually implementing remote helpers for `ria+` URLs so that git commands can work directly with them.

datalad/core/distributed/clone.py
```diff
@@ -636,3 +691,101 @@ def _get_installationpath_from_url(url):
...
+        # parse a RIA URI
+        # check for store related configuration for SSH access
+        if source_ri.scheme == 'ria+ssh':
+            # we support the scenario were hostname is just a label for a config
+            # not an actual hostname
+            hostname = cfg.get(
+                'annex.ria-remote.{}.ssh-host'.format(source_ri.hostname),
+                source_ri.hostname)
+            basepath = cfg.get(
+                'annex.ria-remote.{}.base-path'.format(source_ri.hostname),
```
kyleam: If I understand the role of these options, I think `url.<base>.insteadOf` could also be used to handle the URL rewrites, in which case pointing users to a more general existing solution might be better than adding more options on our side.

bpoldrack: The original intend for those options is about the special remote itself, which doesn't pull its configuration from `git-annex:remote.log` but from those, allowing to reconfigure the mapping of an (autoenabled) special remote to the actual store being used by it depending on config. This was intended to easily switch the store depending on where a clone is (infrastructure reasons) based on a system config the user doesn't have to care about.

However, as the RIA approach evolves, we do indeed want to switch to something like you propose here, since we now can have all required info in that kind of URL. But this needs to be in sync with git-annex-ria-remote itself, so I'd postpone that change wrt this PR.

mih: I like the `insteadOf` idea. The consequence for this PR would be to simply remove the config query, and let lower-level code handle any conversion. I will do just that.

mih: Done.

datalad/core/distributed/clone.py
yarikoptic: Thank you @kyleam for pointing to `gitremote-helper` -- I should have remembered about it since I must have looked at it in the past.
My non-executive summary:
- I am Ok with going with this PR as is ATM

but in the long run (although `datalad::ria::` instead of `ria+` could be a short one) I think
- we should RF to provide `datalad` git remote helper and advise to use `datalad::url` or at least to adhere to such schema ourselves to start. **pros**: even for local paths it would become sensible `datalad::some/possibly/relative/path` (what would it be now in this PR?  haven't seen an example, could it would be smth like `ria+some/possibly/relative/path` if to stay consistent). That would make it clear to the user who sees `datalad::` prefix that it relates to datalad, similarly to how `hg::URL` gives immediate information to me that remote URL I am giving to `git clone` is `hg` repository. (please no grandma examples, use prototypical datalad users)
- layout could be specified as explicit option in the fragment, e.g. `#layout=uuid2` (and describe in docs what uuid2 is and that ria uses such layout scheme (hardcoded)) e.g. `datalad::path#id=UUID;layout=uuid2` or `datalad::path#layout=dataset-tarballs` (for whenever we implement export of the datasets with tarball for `.git` itself), but   
- I think that we might want to altogether avoid hard coding assumptions of the layout on remote end into URL, be it a `ria` (abbreviation which has nothing to do with layout per se) or anything else; neither in prefix of transport nor in URL fragment like I was proposing.  I think, we might want it to be "queried" by datalad from that remote (e.g. some config option within `.datalad/config` on the remote end) similarly how to git-annex queries UUID etc.  It would though require providing support for underlying transport mechanisms to requests such a file (or just entire super/top dataset to some temporary location if not present), but **pros**: it would open possibilities for migrating layout/etc on remote end without requiring adjustment to the URLs or code, and avoid possible guesswork then otherwise needed.  But ATM we even do guesswork for either there is `/.git` or not, so I guess it would be a long shot.
- may alternatively in the short run we could/should "chain" hardcoded `ria` as `datalad::ria::URL` but I am not found of it as much as of `ria+` because then no alternative tool (e.g. browser rendering top JS page of the URL, such as http://store.datalad.org) to datalad would be given a chance.

bpoldrack: > ria (abbreviation which has nothing to do with layout per se) 

Yes and no. You can have that layout without a RIA special remote and current implementation here does that. But: The other way around: RIA would produce such a layout. So, if you know your store was created by it, you wouldn't need know about the layout but just that it's a RIA(-created) store.
Another aspect of it (not currently implemented here): That layout RIA special remote is created is versioned and stored in that store's base-path. So, if we know it's a RIA store and the special remote makes it into core, we can query the special remote for the correct layout without changing that URL. RIA remote's own implementation does exactly that. To do it here would require to import it, of course.


>  I think, we might want it to be "queried" by datalad from that remote (e.g. some config option within .datalad/config on the remote end) similarly how to git-annex queries UUID etc.

I actually wanted to make a very similar point as an argument for using the scheme rather than fragment: Currently we can have a label for a store, that just determines the base of the dataset tree as an actual URL. But: If it's not a regular URL we can enhance that to adress a distrubuted store. That is, several of such stores the user would know about as one thing and let datalad query an index to figure out which store to use for a given dataset id.

mih:
This is done from my POV. Given that we are all somewhat OK with this, I will merge once the tests pass -- unless someone stops me.

It is clear that this is not the end of road, but it seems to me that this PR is not blocking any future developments. Moreover, I consider it a good thing (TM) that we use a special "ria" label for all of this, and do not claim any definitive "datalad" namespace. Whenever I have done such a power-move in the past, I quickly came to regret it.

PR #4026, opened 2020-01-15
BF: Fix clone relpath

bpoldrack:
Based on @kyleam 's #4025 (Couldn't push to your branch)

Don't actually revert, but (hopefully) fix the issue, by manipulating not the originally given URL, but the one git-clone created. Keeping the revert commit in for its added (and a fixed) test.

### DIFF
--- a/datalad/core/distributed/tests/test_clone.py
+++ b/datalad/core/distributed/tests/test_clone.py
@@ -487,7 +487,6 @@ def test_cfg_originorigin(path):
 
 
 # test fix for gh-2601/gh-3538
-@known_failure
 @with_tempfile()
 def test_relative_submodule_url(path):
     Dataset(op.join(path, 'origin')).create()
--- a/datalad/core/distributed/tests/test_clone.py
+++ b/datalad/core/distributed/tests/test_clone.py
@@ -500,9 +500,27 @@ def test_relative_submodule_url(path):
     ds_cloned.repo.fetch()
 
     subinfo = ds.subdatasets(return_type='item-or-list')
+
+    from datalad.support.network import RI
     eq_(subinfo['gitmodule_url'],
         # must be a relative URL, not platform-specific relpath!
-        '../../origin')
+        '../../origin',
+        msg="XXXXXXXXXXX:\n"
+            "submodule pathobj: %s\n"
+            "submodule path: %s\n"
+            "super pathobj: %s\n"
+            "super path: %s\n"
+            "\n"
+            "original ds path: %s\n"
+            "RI(path).localpath: %s\n"
+            "realpath(path): %s\n" %
+            (str(Dataset(op.join(path, 'ds', 'sources')).repo.pathobj),
+             Dataset(op.join(path, 'ds', 'sources')).repo.path,
+             str(ds.repo.pathobj),
+             ds.repo.path,
+             op.join(path, 'origin'),
+             RI(op.join(path, 'origin')).localpath,
+             op.realpath(RI(op.join(path, 'origin')).localpath)))
 
 
 @with_tree(tree={"subdir": {}})
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -934,6 +934,24 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
                 gr.fsck()
             else:
                 lgr.warning("Experienced issues while cloning: %s", exc_str(fix_annex))
+        # make sure that Git doesn't mangle relative path specification into
+        # mildly obscure absolute paths
+        # https://github.com/datalad/datalad/issues/3538
+        if isinstance(url_ri, PathRI):
+            path = Path(url)
+            if not path.is_absolute():
+                # get git-created path
+                git_url = gr.config.get('remote.origin.url')
+                # Note: Not sure, whether there are circumstances where this is relative already
+                if op.isabs(git_url):
+                    # ... and make it a relative one
+                    # Note: Using os.path here, since pathlib's relative_to isn't what you'd expect
+                    git_url = op.relpath(git_url, gr.path)
+                path = Path(git_url)
+                # always in POSIX even on windows
+                path = path.as_posix()
+                gr.config.set('remote.origin.url', path,
+                              where='local', force=True)
         return gr
 
     def __del__(self):
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -944,7 +944,7 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
                 # get git-created path
                 git_url = gr.config.get('remote.origin.url')
                 # Note: Not sure, whether there are circumstances where this is relative already
-                if op.isabs(git_url):
+                if posixpath.isabs(git_url):
                     # ... and make it a relative one
                     # Note: Using posixpath here, since pathlib's relative_to isn't what you'd expect
                     git_url_rel = posixpath.relpath(git_url, gr.pathobj.as_posix())
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -937,6 +937,7 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
         # make sure that Git doesn't mangle relative path specification into
         # mildly obscure absolute paths
         # https://github.com/datalad/datalad/issues/3538
+        # Note, that POSIX is required even on windows, since it's an URL!
         if isinstance(url_ri, PathRI):
             path = Path(url)
             if not path.is_absolute():
@@ -945,12 +946,9 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
                 # Note: Not sure, whether there are circumstances where this is relative already
                 if op.isabs(git_url):
                     # ... and make it a relative one
-                    # Note: Using os.path here, since pathlib's relative_to isn't what you'd expect
-                    git_url = op.relpath(git_url, gr.path)
-                path = Path(git_url)
-                # always in POSIX even on windows
-                path = path.as_posix()
-                gr.config.set('remote.origin.url', path,
+                    # Note: Using posixpath here, since pathlib's relative_to isn't what you'd expect
+                    git_url = posixpath.relpath(git_url, gr.pathobj.as_posix())
+                gr.config.set('remote.origin.url', git_url,
                               where='local', force=True)
         return gr
 
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -74,6 +74,7 @@ from datalad.dochelpers import exc_str
 import datalad.utils as ut
 from datalad.utils import (
     Path,
+    PurePath,
     PurePosixPath,
     assure_list,
     optional_args,
@@ -937,17 +938,18 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
         # make sure that Git doesn't mangle relative path specification into
         # mildly obscure absolute paths
         # https://github.com/datalad/datalad/issues/3538
-        # Note, that git reports POSIX paths even on windows. It's an URL after all. Don't mix in native paths.
+        # Note, that git reports POSIX paths ( '/' instead of `\`) even on windows. It's an URL after all. Don't mix in
+        # native paths.
         if isinstance(url_ri, PathRI):
             url_path = Path(url)
             if not url_path.is_absolute():
                 # get git-created path
-                git_url = gr.config.get('remote.origin.url')
+                git_url = PurePath(gr.config.get('remote.origin.url'))
                 # Note: Not sure, whether there are circumstances where this is relative already
-                if posixpath.isabs(git_url):
+                if git_url.is_absolute():
                     # ... and make it a relative one
                     # Note: Using posixpath here, since pathlib's relative_to isn't what you'd expect
-                    git_url_rel = posixpath.relpath(git_url, gr.pathobj.as_posix())
+                    git_url_rel = posixpath.relpath(git_url.as_posix(), gr.pathobj.as_posix())
                     lgr.debug("Remap origin url from %s to %s", git_url, git_url_rel)
                     gr.config.set('remote.origin.url', git_url_rel, where='local', force=True)
         return gr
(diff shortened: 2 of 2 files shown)


bpoldrack:
FTR: Codecov doesn't make sense really.

datalad/support/gitrepo.py
```diff
@@ -940,10 +940,16 @@ def clone(cls, url, path, *args, **kwargs):
         if isinstance(url_ri, PathRI):
             path = Path(url)
             if not path.is_absolute():
+                # get git-created path
+                git_url = gr.config.get('remote.origin.url')
+                # Note: Not sure, whether there are circumstances where this is relative already
+                if op.isabs(git_url):
+                    # ... and make it a relative one
+                    # Note: Using os.path here, since pathlib's relative_to isn't what you'd expect
+                    git_url = op.relpath(git_url, gr.path)
```
mih: This will not work in general on windows. `git_url` will always be posix, `gr.path` will always be native. So we should like use

`posixpath.relpath(git_url, gr.pathobj.as_posix())`

datalad/support/gitrepo.py
```diff
@@ -940,10 +940,16 @@ def clone(cls, url, path, *args, **kwargs):
...
+                path = Path(git_url)
                 # always in POSIX even on windows
                 path = path.as_posix()
```
mih: If my comment above is implemented, this conversion to and from Path is superfluous.

datalad/support/gitrepo.py
bpoldrack: True. Thx for catching.

bpoldrack:
FTR: Current appveyor failure is about a failed request for codecov. Tests themselves seem to have passed.

mih:
Verdict in conference call was to go with #4025 to stop undesired behavior quickly, and think this through some more.

kyleam:
> (Couldn't push to your branch)

Sorry about that.  AFAIK I haven't changed anything one my side.  Perhaps this is a change on github's end because I just saw a failure when tried to push to your branch for this pr:

```
1 git … push -v bpoldrack refs/heads/revert-clone-relpath\:refs/heads/revert-clone-relpath
Pushing to git@github.com:bpoldrack/datalad.git
Writing objects: 100% (7/7), 760 bytes | 253.00 KiB/s, done.
Total 7 (delta 4), reused 0 (delta 0)
remote: Resolving deltas: 100% (4/4), completed with 4 local objects.        
To github.com:bpoldrack/datalad.git
 ! [remote rejected]     revert-clone-relpath -> revert-clone-relpath (permission denied)
error: failed to push some refs to 'git@github.com:bpoldrack/datalad.git'
```

<details>
<summary>patch</summary>

```diff
From 3c3f7ad95cf6a209c80f07a2339b2c0de4c216d4 Mon Sep 17 00:00:00 2001
From: Kyle Meyer <kyle@kyleam.com>
Date: Wed, 15 Jan 2020 10:55:53 -0500
Subject: [PATCH] TST: clone: Reenable test_relative_submodule_url

This was marked as a known failure in e8c5d7069 (BF: clone: Revert
incorrect relative path adjustment to URLs, gh-4025), but the previous
two commits should make it pass.
---
 datalad/core/distributed/tests/test_clone.py | 1 -
 1 file changed, 1 deletion(-)

diff --git a/datalad/core/distributed/tests/test_clone.py b/datalad/core/distributed/tests/test_clone.py
index 9f86469ad..404bffd5e 100644
--- a/datalad/core/distributed/tests/test_clone.py
+++ b/datalad/core/distributed/tests/test_clone.py
@@ -487,7 +487,6 @@ def test_cfg_originorigin(path):
 
 
 # test fix for gh-2601/gh-3538
-@known_failure
 @with_tempfile()
 def test_relative_submodule_url(path):
     Dataset(op.join(path, 'origin')).create()
-- 
2.24.1

```

</details>

datalad/core/distributed/tests/test_clone.py
```diff
@@ -487,15 +487,37 @@ def test_cfg_originorigin(path):
 
 
 # test fix for gh-2601/gh-3538
+@known_failure
```
kyleam: With your changes, this `@known_failure` should be dropped.  (I tried to push that change, but like you I'm [seeing permission issues][0] when trying to push to other people's branches.)

[0]: https://github.com/datalad/datalad/pull/4026#issuecomment-574729391

datalad/support/gitrepo.py
```diff
@@ -937,14 +937,18 @@ def clone(cls, url, path, *args, **kwargs):
         # make sure that Git doesn't mangle relative path specification into
         # mildly obscure absolute paths
         # https://github.com/datalad/datalad/issues/3538
+        # Note, that POSIX is required even on windows, since it's an URL!
```
kyleam: [ minor question about this comment, mostly for my own curiosity/understanding ]

Is it true that posix is required because git sees this as a URL?  Or is it just because that this is how git would represent windows paths regardless?  I have very limited understanding of how git handles windows paths, but from bits like

https://github.com/datalad/datalad/blob/a21026635fa58d1b488139b04090504233f710f5/datalad/support/gitrepo.py#L3166-L3170

my impression was that git represented windows paths as posix paths underneath, converting at some more outer layer.

datalad/support/gitrepo.py
```diff
@@ -934,6 +934,24 @@ def clone(cls, url, path, *args, **kwargs):
                 gr.fsck()
             else:
                 lgr.warning("Experienced issues while cloning: %s", exc_str(fix_annex))
+        # make sure that Git doesn't mangle relative path specification into
+        # mildly obscure absolute paths
+        # https://github.com/datalad/datalad/issues/3538
+        if isinstance(url_ri, PathRI):
+            path = Path(url)
```
kyleam: I know this is inherited from 02e2b4c46 (BF: Avoid relpath mangling for submodule url configuration, 2020-01-10), but I'm not a fan of using `path` as the variable name.  The two main parameters for this method are `url` and `path`, so I find it confusing to redefine `path` with a value that is derived from `url`.  Perhaps `url_path`?

datalad/support/gitrepo.py
```diff
@@ -937,14 +937,18 @@ def clone(cls, url, path, *args, **kwargs):
...
-                path = path.as_posix()
-                if not path.startswith(op.pardir + '/'):
-                    path = posixpath.join(op.curdir, path)
-                gr.config.set('remote.origin.url', path,
+                # get git-created path
+                git_url = gr.config.get('remote.origin.url')
+                # Note: Not sure, whether there are circumstances where this is relative already
+                if op.isabs(git_url):
+                    # ... and make it a relative one
+                    # Note: Using posixpath here, since pathlib's relative_to isn't what you'd expect
```
kyleam: As I mentioned on the call, I'm a bit worried that pathlib is avoiding relpath's more extensive behavior for some edge case where it can behave incorrectly.  But I can't come up with what that would be, and looking at pathlib's docs and source doesn't reveal any hints, so that worry is probably unfounded.

datalad/support/gitrepo.py
```diff
@@ -937,14 +937,18 @@ def clone(cls, url, path, *args, **kwargs):
...
+                    git_url = posixpath.relpath(git_url, gr.pathobj.as_posix())
```
kyleam: I think it'd be worth having a debug level message here reporting the remapping.

datalad/support/gitrepo.py
```diff
@@ -937,14 +937,18 @@ def clone(cls, url, path, *args, **kwargs):
...
+                gr.config.set('remote.origin.url', git_url,
```
kyleam: If the `op.isabs(git_url)` condition is false, the url hasn't changed from the configured one, so we might as well avoid writing it out again with `.set` (i.e. move this call under the condition).

datalad/support/gitrepo.py
bpoldrack: I should prob. reword that comment. AFAIK it's not technically "required" (I think can deal with actual system paths), but the standard way of git to represent it. So, we need to expect posix path being reported by git and we want to write posix path to be consistent. Effectively that makes it a "requirement" *for us* from my point of view. Otherwise we tend to shoot our feet by mixing.

datalad/core/distributed/tests/test_clone.py
bpoldrack: True, indeed. Thx.

datalad/support/gitrepo.py
bpoldrack: Agree. `url_path` is fine.

datalad/support/gitrepo.py
bpoldrack: Well, there has to be a reason for pathlib to provide such limited functionality with `relative_to`. But I can't come up with a scenario where `os.path.relpath` wouldn't work here either. May be there's a way to break it, when the path to the repository (our starting point for the relative path) would contain symlinks. But that doesn't apply here, since `GitRepo.path` is realpath'd already.

datalad/support/gitrepo.py
bpoldrack: Yes. Will do.

datalad/support/gitrepo.py
bpoldrack: True.

bpoldrack:
Thanks much, @kyleam !
I completely agree and pushed respective changes.

kyleam:
Thanks for the updates.  I see that as of 109fd6c5c the new test still [fails][0] on the windows run.  Once that's figured out, I think this is good to go.

[0]: https://github.com/datalad/datalad/pull/4026/checks?check_run_id=393032329#step:8:669

bpoldrack:
@kyleam: Yes, saw that. Not completely clear to me yet, what's going on. But I have an idea ...

bpoldrack:
FTR: test in question passes on a native windows box:

```
(base) C:\Users\mih\code\datalad>python -m nose -s -v datalad.core.distributed.tests.test_clone:test_relative_submodule_url
datalad.core.distributed.tests.test_clone.test_relative_submodule_url ...

XXXXXXXXXXX: submodule pathobj: C:/Users/mih/AppData/Local/Temp/datalad_temp_kbgqj8l1/ds/sources
ok
Versions: appdirs=1.4.3 boto=2.49.0 cmd:annex=7.20191107-g8ea269ef7 cmd:bundled-git=UNKNOWN cmd:git=2.23.0.windows.1 cmd:system-git=2.23.0.windows.1 cmd:system-ssh=8.0p1 git=3.0.2 gitdb=2.0.5 humanize=0.5.1 iso8601=0.1.12 keyring=19.1.0 keyrings.alt=3.1.1 msgpack=0.6.1 requests=2.22.0 tqdm=4.32.1 wrapt=1.11.2
Obscure filename: str=b' ;abc' repr=' ;abc'
Encodings: default='utf-8' filesystem='utf-8' locale.prefered='cp1252'
Environment: PATH='C:\\Users\\mih\\Miniconda3;C:\\Users\\mih\\Miniconda3\\Library\\mingw-w64\\bin;C:\\Users\\mih\\Miniconda3\\Library\\usr\\bin;C:\\Users\\mih\\Miniconda3\\Library\\bin;C:\\Users\\mih\\Miniconda3\\Scripts;C:\\Users\\mih\\Miniconda3\\bin;C:\\Users\\mih\\Miniconda3\\condabin;C:\\Program Files (x86)\\Intel\\Intel(R) Management Engine Components\\iCLS;C:\\Program Files\\Intel\\Intel(R) Management Engine Components\\iCLS;C:\\WINDOWS\\system32;C:\\WINDOWS;C:\\WINDOWS\\System32\\Wbem;C:\\WINDOWS\\System32\\WindowsPowerShell\\v1.0;C:\\WINDOWS\\System32\\OpenSSH;C:\\Program Files (x86)\\Intel\\Intel(R) Management Engine Components\\DAL;C:\\Program Files\\Intel\\Intel(R) Management Engine Components\\DAL;C:\\Program Files (x86)\\Intel\\Intel(R) Management Engine Components\\IPT;C:\\Program Files\\Intel\\Intel(R) Management Engine Components\\IPT;C:\\Program Files\\TortoiseGit\\bin;C:\\Program Files (x86)\\Intel\\Intel(R) Management Engine Components\\iCLS;C:\\Program Files\\Intel\\Intel(R) Management Engine Components\\iCLS;C:\\WINDOWS\\system32;C:\\WINDOWS;C:\\WINDOWS\\System32\\Wbem;C:\\WINDOWS\\System32\\WindowsPowerShell\\v1.0;C:\\WINDOWS\\System32\\OpenSSH;C:\\Program Files (x86)\\Intel\\Intel(R) Management Engine Components\\DAL;C:\\Program Files\\Intel\\Intel(R) Management Engine Components\\DAL;C:\\Program Files (x86)\\Intel\\Intel(R) Management Engine Components\\IPT;C:\\Program Files\\Intel\\Intel(R) Management Engine Components\\IPT;C:\\Program Files\\TortoiseGit\\bin;C:\\Users\\mih\\AppData\\Local\\Microsoft\\WindowsApps;C:\\Users\\mih\\Miniconda3\\Library\\mingw64\\bin;C:\\Program Files (x86)\\Midnight Commander;C:\\Program Files (x86)\\7-Zip;.'

Ran 1 test in 14.796s

OK
```

Trying to figure what's different with that Github Action thingy.

bpoldrack:
Ok. At least I found the problem.
In the environment of that Github Action, `GitRepo.pathobj.as_posix()` reports `C:/Users/RUNNER~1/AppData/Local/Temp/datalad_temp__ey2dkku/ds/sources` as the submodules root dir, whereas the path we are trying to make relative is represented with the long name `runneradmin` instead of `RUNNER~1`. Therefore `relpath` has to go up to `C:/Users`. While the resulting relative path is technically correct and a fetch still works, this is not an actually useful relative path.

So, we need to find out how `GitRepo` can be taught to use the long names (which should work under any remotely recent windows).

mih:
> So, we need to find out how `GitRepo` can be taught to use the long names (which should work under any remotely recent windows). This might well solve other issues as well.

https://github.com/datalad/datalad/pull/4054#issuecomment-576150443

mih:
@bpoldrack This PR is still open. It feels as if that need not be the case. Can you please offer your assessment. Thx!

mih:
6+ month of inactivity, no response from OP, closing.

Issue #4035, opened 2020-01-17
Make clone() accept arbitrary git-clone options


mih:
We do it for create(), so it would make sense to do it for the other command that can materialize datasets.

Given that we may, at some point, want to be able to clone a dataset from a source that git-clone doesnt support, it seems to be better to use a dedicated option, rather than an argparse.REMAINDER- like setup.

It would br necessary to add a bunch of tests that make sure we operate properly without having to interpret a particular option combination. That means we should inspect the resulting repo, instead of making decisions upfront.

First and foremost, this would be instrumental in starting to support bare datasets.

mih:
Verdict in call: have a dedicated option to collection additional option to pass to git-clone.

mih:
Closing in favor of near-duplicate #5586

PR #4036, opened 2020-01-17
NF: Support cloning of specific repository versions (fixes gh-2109)

mih:
`GitRepo.clone()` is enhanced to accept arbitrary additional options
that are passed on to `git-clone`. We use the `--branch` option
(that, despite its name, can handle any relevant version identifier)
to let `git-clone` do all the work.

`datalad-clone` is now set up to acknowledge a request for a particular
version, for any kind of source URL where `decode_source_spec()`
yields a non-None `version` property.

At the moment this is only the case for `ria+http|ssh://` URL, but
future additions only need to alter/enhance `decode_source_spec()`
to extend this functionality for other types.

This change also sets the stage for gh-4035

The user experience of this new feature is negatively impacted by #4038.

### DIFF
--- a/datalad/core/distributed/clone.py
+++ b/datalad/core/distributed/clone.py
@@ -385,16 +385,25 @@ def clone_dataset(
     )
     error_msgs = OrderedDict()  # accumulate all error messages formatted per each url
     for cand in candidate_sources:
+        log_progress(
+            lgr.info,
+            'cloneds',
+            'Attempting to clone from %s to %s', cand['giturl'], dest_path,
+            update=1,
+            increment=True)
+
+        clone_opts = {}
+
+        if cand.get('version', None):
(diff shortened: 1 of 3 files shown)


yarikoptic:
At least according to the man git-clone, there is no promise for git clone to be able to checkout specific hexsha.

--branch <name>
Instead of pointing the newly created HEAD to the branch pointed to by the cloned repository’s HEAD, point to <name> branch instead. In a non-bare repository, this is the branch that will be checked out. --branch can also take tags and detaches the HEAD at that commit in the resulting repository.

But I guess underneath it just uses `checkout` those could consume anything it consumes. 

Either way - great stuff, thanks!

mih:
FTR: We test this approach with a tag (also not promised to work), as a likely non-branch approach to versioning.

Issue #4030, opened 2020-01-16
clone should record its "reckless" setting to "get" (clone) subdatasets in the same mode


mih:
n.t.

PR #4037, opened 2020-01-17
NF: Clone of subdatasets uses reckless mode of superdataset (fixes gh-4030)

mih:
If `--reckless` is given, the label is stored verbatim under config
variable `datalad.clone.reckless`. `clone()` consults this variable
whenever `--reckless` is None.

As a consequence of this change, subdatasets will inherit any reckless
mode.

Obviously, this also makes it possible to provide global configuration
to force a particular reckless mode for any cloned dataset.

### DIFF
--- a/datalad/core/distributed/clone.py
+++ b/datalad/core/distributed/clone.py
@@ -174,11 +174,6 @@ class Clone(Interface):
             dataset=None,
             description=None,
             reckless=None):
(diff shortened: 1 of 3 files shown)


mih:
Test failure might be due to external service unavailability.

mih:
FTR: I will merge this on Monday, unless objections are expressed.

Issue #4019, opened 2020-01-12
Replace patool with direct cmdline calls?


mih:
I wonder if there was a specific motivation to use patool for archive file management? Given issues like #3769 and #2555, and now #4018 I am thinking about reimplementing `compress_file` and `decompress_file` with direct calls to `7z` and `tar`. I think it would be relatively easy to do, and cut away an external dependency (even patool needs the underlying tools to be installed, it seems).

@yarikoptic

yarikoptic:
the motivation was/is to not re-implement what was already implemented, and for many more archives than just .7z and .tar{,.xz,.gz}.  
We did have `.gz` and `.zip` at least as well.  I think I had some .rar and may be few others which I had "handled" with datalad at some point.

> I am thinking about reimplementing compress_file and decompress_file with direct calls to 7z and tar

Well, I guess nobody would stop you... (or you could fix patool to benefit not only datalad but other users of patool as well).  If you are to reimplement - just make it configurable (to be able to force use of patool) and use patool for other types of archives support for which you wouldn't reimplement.

mih:
Point taken, but I am still not seeing the benefit of patool for our purposes. All of what you write above is possible with straight-up 7z cmdine alone. It supports all the formats you mentioned, including `tar.?` (so I wouldn't even bother with TAR). It might well be that the entire code to achieve the same results through 7z is less than the compatibility dance we are currently doing to pass information to patool.

yarikoptic:
one of the issues I had with patool (and 7z for that matter) is handling large files: https://github.com/wummel/patool/issues/49

mih:
I checked my recent work re archive file size. The largest archive I built has 1.9T, I can reliably extract arbitrary content from it:

```
$ ls -lh archive.7z
-rw-r--r-- 1 me users 1.9T Nov 16 22:54 archive.7z
$ 7z e archive.7z physio.json

7-Zip (a) [64] 16.02 : Copyright (c) 1999-2016 Igor Pavlov : 2016-05-21
p7zip Version 16.02 (locale=en_US.UTF-8,Utf16=on,HugeFiles=on,64 bits,32 CPUs Intel(R) Xeon(R) Gold 6142 CPU @ 2.60GHz (50654),ASM,AES-NI)

Scanning the drive for archives:
1 file, 2063868614059 bytes (1923 GiB)

Extracting archive: archive.7z
--
Path = archive.7z
Type = 7z
Physical Size = 2063868614059
Headers Size = 391804
[...]

PR #4039, opened 2020-01-17
RF: Move all patool code into a dedicated code file

mih:
This shall make providing an alternative but API compatible
implementation easier to place and test alongside of this existing
one.

Towards #4019

### DIFF
--- a/datalad/support/archive_utils_patool.py
+++ b/datalad/support/archive_utils_patool.py
@@ -1,0 +1,236 @@
+# emacs: -*- mode: python; py-indent-offset: 4; tab-width: 4; indent-tabs-mode: nil -*-
+# ex: set sts=4 ts=4 sw=4 noet:
+# ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+#
+#   See COPYING file distributed along with the datalad package for the
+#   copyright and license terms.
+#
+# ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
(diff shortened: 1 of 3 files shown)


mih:
I will ignore the coverage report. Moving untested code doesn't make things substantially worse.

datalad/tests/test_archives.py
```diff
@@ -25,8 +25,8 @@
     compress_files,
     decompress_file,
     ExtractedArchive,
-    unixify_path,
 )
+from datalad.support.archive_utils_patool import unixify_path
```
yarikoptic: Why to introduce absolute import when the rest (at least below) are relative?

datalad/support/archive_utils_patool.py
```diff
@@ -0,0 +1,213 @@
...
+
+import os
+from .exceptions import MissingExternalDependency
+from .path import (
+    basename,
+    join as opj,
+    exists,
+)
+
+from datalad.utils import (
```
yarikoptic: While at it make into a relative import? Note that there are more imports from ..utils below

mih:
Thx for the review!

re imports: Everything is switched to absolute. If you check the follow-up PR, you will see that no relative imports are left. If you are concerned about this intermediate state, we can leave this hanging here, or even close it till #4041 is done. I just wanted to disentangle to move from the subsequent changes. Your call, I am fine either way.

yarikoptic:
I am confused on where and why everything changed to absolute, since I thought ideally we should use relative, but ok, I guess we could RF later.

PR #4041, opened 2020-01-17
7z-based (optional) replacement for patool

mih:
Sitting on top of #4039 

This is a complete draft implementation of a 7z-based file compression/decompression that should be as capable as the patool-based one.

TODO

- [x] confirm runs on windows (main objective here)
- [x] expand tests for additional archive formats and compressors (prev. tests were fairly minimal)
  - [x] support 7z archives
  - [x] support xz and tar.xz (required to pull in `xz-utils` on travis
  - [x] run entirety of `test_archives.py` on all platforms
- [x] implement fall-back to patool on non-windows platforms
  - [x] requestable via boolean `datalad.runtime.use-patool`config switch
  - [x] transparent fall-back when `7z` is not found/supported (as requested by @yarikoptic) although now we need close inspection of what was actually executed whenever something archive-related fails.
- [x] `wtf` reports `7z` version
- [x] RF if #4050 should land before this or vice versa

Further related TODOs (for subsequent PRs) surfaced by this PR:

- #4048 -> #4050
- add `p7zip` as dependency/recommends to Debian package

### DIFF
--- a/datalad/support/archives.py
+++ b/datalad/support/archives.py
@@ -44,7 +44,9 @@ from datalad.utils import (
     get_tempfile_kwargs,
 )
 from datalad import cfg
-if cfg.get('datalad.runtime.use-patool', False):
+from datalad.config import anything2bool
+if cfg.obtain('datalad.runtime.use-patool', default=False,
+              valtype=anything2bool):
     from datalad.support.archive_utils_patool import (
         decompress_file as _decompress_file,
         # other code expects this to be here
--- a/datalad/support/archives.py
+++ b/datalad/support/archives.py
@@ -43,11 +43,19 @@ from datalad.utils import (
     rmtree,
     get_tempfile_kwargs,
 )
-from datalad.support.archive_utils_patool import (
-    decompress_file as _decompress_file,
-    # other code expects this to be here
-    compress_files
-)
+from datalad import cfg
+if cfg.get('datalad.runtime.use-patool', False):
+    from datalad.support.archive_utils_patool import (
+        decompress_file as _decompress_file,
+        # other code expects this to be here
+        compress_files
+    )
+else:
+    from datalad.support.archive_utils_7z import (
+        decompress_file as _decompress_file,
+        # other code expects this to be here
+        compress_files
+    )
 
 lgr = logging.getLogger('datalad.support.archives')
 
--- a/datalad/support/archives.py
+++ b/datalad/support/archives.py
@@ -32,6 +32,7 @@ from datalad.support.path import (
 )
 
 from datalad.support.locking import lock_if_check_fails
+from datalad.support.external_versions import external_versions
 from datalad.consts import ARCHIVES_TEMP_DIR
 from datalad.utils import (
     any_re_search,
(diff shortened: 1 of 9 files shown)


mih:
Grrrr, this was supposed to be in draft mode. I see no way to achieve that post-factum.

datalad/support/archives.py
```diff
@@ -11,131 +11,53 @@
...
-        if drive:
-            # last one must be :
-            assert(drive[-1] == ":")
-            return '/%s%s' % (drive[:-1], path_)
-        else:
-            return path_
-    else:
-        return path
+from datalad import cfg
+if cfg.get('datalad.runtime.use-patool', False):
```
yarikoptic: I would have also checked if `external_versions["cmd:7z"]` since if it not available (we cannot force its installation via setup.py but it is available) better to announce that (e.g. even a lgr.warning) and fall to patool which might choose another (7z is the default for many types) way to handle those files.  IMHO it would be better to just crash with "7z command is not found" message

datalad/support/archives.py
```diff
@@ -314,9 +155,9 @@ def __init__(self, toppath=None, persistent=False):
                 self._made_path = True
                 os.makedirs(path)
                 lgr.debug("Cache initialized")
-            except:
+            except Exception as e:
                 lgr.error("Failed to initialize cached under %s" % path)
-                raise
+                raise e
```
yarikoptic: Since except is merely for logging, I would say that
```suggestion
                raise
```
would remain a preferred way to simply reraise exception, since otherwise it would use this line as the source of exception while also mentioning original one (in python3 only):
```
/tmp > cat /tmp/1.py
try:
    raise ValueError("raising on line 2")
except Exception as e:
    raise e

/tmp > python3 /tmp/1.py
  File "/tmp/1.py", line 4, in <module>
    raise e
  File "/tmp/1.py", line 2, in <module>
    raise ValueError("raising on line 2")
ValueError: raising on line 2
```
imho just adds confusion and probably would somewhat interfere with our logging of exceptions (would point to this re-raised line)

yarikoptic:
> * expand tests for additional archive formats and compressors

would also be nice to
- [ ] make one of the travis runs use alternative (non default) archives handler (i.e. patool instead of 7z) by providing config var in the env.

mih:
Holy cow, this is a rabbit hole... I am so deep in `add_archive_content.py` and I don't even know why...

datalad/support/archives.py
mih: I will add that test.

datalad/support/archives.py
mih: Will do.

mih:
> Automagically fallback to patool if 7z command is not available.

I am reluctant to do this. We are pretty much in the same situation with patool. We can pull patool in via dependency, but we cannot pull in its dependencies (tar, zip, ...). Moreover, the reason why I am doing this to begin with, is that this code is broken, and it should be removed, or fixed. Hence, I went for a dedicated switch.

datalad/support/archives.py
mih: Done

datalad/support/archives.py
mih: Done

yarikoptic:
> We are pretty much in the same situation with patool. We can pull patool in via dependency, but we cannot pull in its dependencies (tar, zip, ...).

patool has fallbacks from 7z to e.g. zip, jar, etc, so it could work on systems without 7z.   You might fix some bugs (on windows) but would introduce datalad crashes on others (without 7z, which is generally not installed by default, whenever zip etc are more known and more likely to be installed).  so I would strongly appreciate (could buy a bottle) if we could avoid such crashes with a cost of a single `if` condition.

mih:
> I would strongly appreciate (could buy a bottle) if we could avoid such crashes with a cost of a single `if` condition.

ATM this is complicated by #4047

mih:
I am planning to merge this relatively quickly: On Monday, or whenever the test pass (whichever happens last). It is the foundation for more work on windows, which will go into different directions, hence should not be added here.

mih:
Jeez, why would an NFS mount make a difference?

Issue #4034, opened 2020-01-16
Need sphinx/docs solution to reference handbook content


mih:
Instead of bloating the command docstrings it would often be helpful to add pointers to handbook content for background info or details.

If I want to do this now, I'd have to add something like this to a docstring

```
    .. seealso::

      http://handbook.datalad.org/en/latest/usecases/datastorage_for_institutions.html
        More information on Remote Indexed Archive (RIA) stores
 ```

It renders nicely in both manpage and HTML docs, but it still ugly, because it will unconditionally reference the latest version of the handbook, and not necessarily the version that corresponds to the running datalad.

@adswa has started maintaining release branches matching `MAJOR.MINOR` versions of DataLad in the handbook. So we have this URL available now:

http://handbook.datalad.org/en/0.12/usecases/datastorage_for_institutions.html

Now we could introduce a directive that enables us to place more readable handbook references, that point to the relevant version of the documentation.

adswa:
I've started to look into defining custom roles that allow to reference the handbook, similar to the built-in ``:manpage:`` role of Sphinx. For the ``:manpage:`` directive, one defines the base url in ``conf.py``, e.g., ``manpages_url = 'http://docs.datalad.org/generated/man/{page}.html'``, and can subsequently link to the docs with ``:manpage:`datalad-create` ``.

From what the internet tells me, defining custom roles (e.g., a ``:handbook:`` role one can use with ``:handbook:`<version><section>` ``) seems to require writing a sphinx extension. Here is a tutorial that describes how a role similar to what we want is written:
https://doughellmann.com/blog/2010/05/09/defining-custom-roles-in-sphinx/

mih:
I am on it -- half-working already.

PR #4046, opened 2020-01-18
NF: Intersphinx mapping to link to handbook content (fixes gh-4034)

mih:
Demo link is included. All rendered references are version to match the
current "major" release.
    
Looks like this: `--help` output and `help(clone)` in Python
    
```
  handbook:3-001 (http://handbook.datalad.org/symbols)
    More information on Remote Indexed Archive (RIA) stores
```
    
The given URL points to a dedicated index section in the handbook, where
we can maintain any number of references that point to arbitrary
locations. The name 'symbols' is used, because this is how sphinx lists
numeric references. The ?-???? scheme aims to mimic manpage sections
(where we can come up with some categorization, if needed, and
incremental integer IDs).

- 1: commands
- 2: concepts
- 3: use cases

Ping @adswa for awareness (I have pushed a corresponding change to the
handbook; directly, no PR).

### DIFF
--- a/datalad/core/distributed/clone.py
+++ b/datalad/core/distributed/clone.py
@@ -99,7 +99,7 @@ class Clone(Interface):
 
     .. seealso::
 
-      http://handbook.datalad.org/en/latest/usecases/datastorage_for_institutions.html
+      :ref:`handbook:3-001`
         More information on Remote Indexed Archive (RIA) stores
     """
     # by default ignore everything but install results
--- a/datalad/interface/base.py
+++ b/datalad/interface/base.py
@@ -15,6 +15,7 @@ __docformat__ = 'restructuredtext'
 import logging
 lgr = logging.getLogger('datalad.interface.base')
 
+import os
 import sys
 import re
 import textwrap
@@ -239,6 +240,19 @@ def alter_interface_docs_for_api(docs):
         lambda match: match.group(1),
         docs,
         flags=re.MULTILINE | re.DOTALL)
+    if 'DATALAD_SPHINX_RUN' not in os.environ:
+        # remove :role:`...` RST markup for cmdline docs
+        docs = re.sub(
+            r':\S+:`[^`]*`[\\]*',
+            lambda match: ':'.join(match.group(0).split(':')[2:]).strip('`\\'),
+            docs,
+            flags=re.MULTILINE | re.DOTALL)
+        # make the handbook doc references more accessible
+        # the URL is a redirect configured at readthedocs
+        docs = re.sub(
+            r'(handbook:[0-9]-[0-9]*)',
+            '\\1 (http://handbook.datalad.org/symbols)',
+            docs)
     docs = re.sub(
         r'\|\| REFLOW \>\>\n(.*?)\<\< REFLOW \|\|',
         lambda match: textwrap.fill(match.group(1)),
@@ -282,6 +296,12 @@ def alter_interface_docs_for_cmdline(docs):
         lambda match: ':'.join(match.group(0).split(':')[2:]).strip('`\\'),
         docs,
         flags=re.MULTILINE | re.DOTALL)
+    # make the handbook doc references more accessible
+    # the URL is a redirect configured at readthedocs
+    docs = re.sub(
+        r'(handbook:[0-9]-[0-9]*)',
+        '\\1 (http://handbook.datalad.org/symbols)',
+        docs)
     # remove None constraint. In general, `None` on the cmdline means don't
     # give option at all, but specifying `None` explicitly is practically
     # impossible
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -3,14 +3,9 @@
 
 # You can set these variables from the command line.
 SPHINXOPTS    = -W
-SPHINXBUILD   = sphinx-build
+SPHINXBUILD   = DATALAD_SPHINX_RUN=1 sphinx-build
 PAPER         =
 BUILDDIR      = build
-
-# User-friendly check for sphinx-build
-ifeq ($(shell which $(SPHINXBUILD) >/dev/null 2>&1; echo $$?), 1)
-$(error The '$(SPHINXBUILD)' command was not found. Make sure you have Sphinx installed, then set the SPHINXBUILD environment variable to point to the full path of the '$(SPHINXBUILD)' executable. Alternatively you can add the directory with the executable to your PATH. If you don't have Sphinx installed, grab it from http://sphinx-doc.org/)
-endif
 
 # Internal variables.
 PAPEROPT_a4     = -D latex_paper_size=a4
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -336,4 +336,12 @@ texinfo_documents = [
 
 
 # Example configuration for intersphinx: refer to the Python standard library.
-intersphinx_mapping = {'https://docs.python.org/': None}
+intersphinx_mapping = {
+    'handbook': (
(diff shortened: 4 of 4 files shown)


adswa:
It might be me or my setup, but when I build the docs locally on this commit, it only references the handbook correctly in the Python module reference of clone, but not in the commandline reference:

<details><summary>Python API works </summary>

image

</details>
<details><summary>cmdline reference doesn't</summary>

image

</details>

I'm not sure why, or what I might be doing wrong

kyleam:
> It might be me or my setup [...]

FWIW I was just in the process of writing the the same thing.

mih:
That is kinda how it works. Only rendered refs look nice automatically. Like with any other ref, we can tweak the raw text version (which what is in docstring and manpage).

```:ref:`better text <handook:usecase-scalable-datastore` ```

I have another implementation that uses a custom link role, but that has the same issue, because it can only show what is in the docstring. We could extend the docstring preprocessing to discover such refs and somehow make them better, but I had no good idea yet.

**Comment by adswa** [comment]
I think I found the culprits:

https://github.com/datalad/datalad/blob/fca00c076d45326920989e9246449d67a2d9ac01/datalad/interface/base.py#L280-L284

https://github.com/datalad/datalad/blob/fca00c076d45326920989e9246449d67a2d9ac01/datalad/interface/base.py#L293-L298

https://github.com/datalad/datalad/blob/fca00c076d45326920989e9246449d67a2d9ac01/datalad/interface/base.py#L304-L309

But without them, the commandline help contains rst markup:
<details>

```
datalad clone --help
Usage: datalad clone [-h] [-d DATASET] [-D DESCRIPTION] [--reckless [{auto}]]
                     SOURCE [PATH]

Obtain a dataset (copy) from a URL or local directory

The purpose of this command is to obtain a new clone (copy) of a dataset
and place it into a not-yet-existing or empty directory. As such `clone`
provides a strict subset of the functionality offered by `install`. Only a
single dataset can be obtained, and immediate recursive installation of
subdatasets is not supported. However, once a (super)dataset is installed
via `clone`, any content, including subdatasets can be obtained by a
subsequent `get` command.

Primary differences over a direct `git clone` call are 1) the automatic
initialization of a dataset annex (pure Git repositories are equally
supported); 2) automatic registration of the newly obtained dataset as a
subdataset (submodule), if a parent dataset is specified; and 3) support
for additional resource identifiers (DataLad resource identifiers as used
on datasets.datalad.org, and RIA store URLs as used for store.datalad.org;
see examples); and 4) automatic configurable generation of alternative
access URL for common cases (such as appending '.git' to the URL in case
the accessing the base URL failed).

.. note::

   # this is a test

.. seealso::

  :ref:`handbook:usecase-scalable-datastore`
    More information on Remote Indexed Archive (RIA) stores
```
</details>

**Comment by adswa** [comment]
oh sorry, I didn't see your earlier comment...

**Comment by kyleam** [comment]
Stepping back, I'm negative on this change in general because it makes the local `--help` and `pydoc` output less useful.  I would prefer a working link that points to latest version rather the source-matched version (i.e. the problem described in gh-4034) that leads to an obscure reference in several of the representations.

**Comment by mih** [comment]
I concur with the general criticism, and I am fine to not merge this PR.

That being said: The status quo (full URL) is rather fragile, in addition to being bulky. Here is, for example, a URL referencing a section in the handbook:

`http://handbook.datalad.org/en/0.12/usecases/datastorage_for_institutions.html#the-data-store-as-a-git-annex-ria-remote`

120 chars that are not only dependent on the specific source file name, but also on the particular wording of the headline. Given the state of the handbook (fast development, frequent structure optimizations), these will break in *any* representation of the documentation. Only the intersphinx approach gives us a compensation layer, as we only have to maintain to link anchors in the handbook sources, of which we can have any number, and the cost of making them stable "forever" is very low.

I have pushed an alternative solution based on `extlinks` that would lessen the pain of URL size, but remains fragile.

The last thing I can think about is to use index-markup to identify points of interested for linking from the docstring. Those would show up at http://handbook.datalad.org/en/0.12/genindex.htm so we could still have a version pages, but would leave the final navigation step to users coming from the docstring/cmdline help. We cannot use the intersphinx resolution approach for the docstrings, because it would be way too slow.

Personally, without any such helper, I would not start adding handbook links to the docstrings, because they would be to painful to maintain.

ATM the following approach seems the most viable:

```
:ref:`DataLad handbook usecase 13 <handbook:usecase-scalable-datastore>`
```

Where `13` is some kind of identifier or numbering that is discoverable by a human via the table of contents. ATM the handbook does that at the chapter level for the "Basics" already. I would be fine to
have the docstring references be approximate/coarse, when the HTML and PDF links for be up-to-date and precise.

**Comment by adswa** [comment]
FTR: I have begun to restructure the handbook toctree and numbered the use cases. This should make easy navigation based on a numerical identifier possible, and improves the index page of the handbook regardless of this proposed feature. Personally, I agree that the full URL is both ugly and fragile. What you refer to as "the most viable approach" seems very sensible and workable to me, compared with everything else, and I believe that at least for some functionality such as the RIA store URLS it is beneficial to point to handbook content.

**Comment by kyleam** [comment]
Perhaps I'm being obtuse or not understanding what's being proposed, but I don't think I'd look at

    :ref:`DataLad handbook usecase 13 <handbook:usecase-scalable-datastore>`

and have much more sense what to do with it than

    :ref:`handbook:usecase-scalable-datastore`

At any rate, please don't waste time on coming up with a solution to appease one dissenting opinion.  We weigh the tradeoffs differently, but given that you two are the ones working on the handbook, your assessment matters more here.  (And I don't think most datalad users share my fondness for local help/documentation.)

I do think you should try to not restructure the handbook too often.  You want people to link to it, and no one likes broken URLs.

**Comment by mih** [comment]
> Perhaps I'm being obtuse or not understanding what's being proposed, but I don't think I'd look at
> 
> ```
> :ref:`DataLad handbook usecase 13 <handbook:usecase-scalable-datastore>`
> ```
> 
> and have much more sense what to do with it than
> 
> ```
> :ref:`handbook:usecase-scalable-datastore`
> ```

True in many ways. For the machine it is the same info, for a human reader I was hoping it would be enough, hence I proposed the latter initially, but it wasn't well received. The former is turned into

```
DataLad handbook usecase 13
```

which I would argue is damn close to something like `Ezekiel 17`, and as such proven to be decipherable for more than a thousand years ;-)

> I do think you should try to not restructure the handbook too often. You want people to link to it, and no one likes broken URLs.

Same here. It will stop immediately, once an optimum is reached ;-)

kyleam:
All right, I'm ok with being dissenting opinion here, but not ok with it being blocking.  I say let's reset to the initial commit (5a1ca180f) and merge.

mih:
Dont worry, I am just distracted. You are not blocking anything.

mih:
FTR: @adswa is investigating how to best implement "medieval navigation" for the handbook. Once that is figured out, we can reset this to 5a1ca18 -- possibly implement a docstring rendering improvement -- and start filling the docs up with links.

mih:
Ok, reworked the beast with @adswa Should be functional now. Top-comment edited to reflect current approach.

mih:
Unless there are strong objections, I will merge this tomorrow morning. This is mostly relevant for handbook content and @adswa was cool with it. The few links we will have in the docstrings are easy to change in structure and content later on.

kyleam:
> Unless there are strong objections, I will merge this tomorrow morning.

From my end, no new objections to the approach, and no objection to merging despite my objections.

mih:
Thx @kyleam !

PR #4050, opened 2020-01-19
Replace shlex.quote() with quote_cmdlinearg()

mih:
This enables more appropriate command-line argument quoting on
windows, and will retain the previous behavior on any other platform.

This could potentially improve SSH support on Windows substantially,
because shlex.quote() was often inappropriately and unconditionally
used in associated code.

- [x] also replace `maybe_shlex_quote()`


mih:
Going in. Will RF #4041 to make use of it.

Issue #4049, opened 2020-01-19
Use of shlex.split() likely causing issues on Windows


mih:
Used in

- `run_procedure()`
- `Runner.run()`
- `AddArchiveContent`
- `AnnexRepo._annex_custom_command()`
- `AnnexRepo.copy_to()`
- `GitRepo._git_custom_command()`
- `sshrun()`
- `test_cmd.py`

Solution candidate:
https://stackoverflow.com/questions/33560364/python-windows-parsing-command-lines-with-shlex/35900070#35900070

However, each use of split() should be investigated, because splitting for the purpose of `Popen` is unnecessary and would be  immediately undone by it.

PR #4053, opened 2020-01-20
NF: split_cmdline() platform-appropriate shlex.split()

mih:
shlex.split() is kept for any non-Windows platform, as custom
implementation for Windows was added. All use of shlex.split()
was replaced by the new helper.

Fixes gh-4049

### DIFF
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -16,7 +16,6 @@ import subprocess
 import sys
 import logging
 import os
-import shlex
 import atexit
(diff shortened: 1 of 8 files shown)


mih:
I will merge this now. The change is no functional impact on non-Windows platforms, but the pieces on Windows that are touched are still entangled with too many other problems so putting follow-ups in here would bloat the scope. This is merely a setup for me to continue enabling tests on windows.

PR #4054, opened 2020-01-20
BF: Make all known_failure GitRepo tests run on windows

mih:
This includes a fix to GitRepo.get_toppath() which used POSIX-convention
paths on windows, but was tested to produce native paths. It now
yields native paths.

This PR sits on top of #4053


mih:
Test failure is due to the fact that the github CI windows setup reports short paths (ala `C:\SHORT~1\some`

PR #4058, opened 2020-01-20
RM: Wipe out obsolete save() from pre 0.12 times

mih:
All tests had been ported ages ago.


mih:
Argh, I got too excited. I found no way to make `add()` work without it...

PR #4064, opened 2020-01-21
NF: Helper to support Git-style URL rewriting

mih:
With this helper (and its use in the relevant places), we can make
DataLad acknowledge URL rewrite instructions by supporting Git's own
mechanism: `url.<base>.insteadof`

With this move we should be able to avoid custom configuration for
the purpose of mapping locations (like needed for the RIA support
code). The included test contains the desired mappings.

As suggested by @kyleam this should set us up better for eventually
supporting custom protocols like `ria+ssh://` in a proper Git remote
helper.

### DIFF
--- a/datalad/config.py
+++ b/datalad/config.py
@@ -722,3 +722,49 @@ class ConfigManager(object):
 
         # use unset all as it is simpler for now
         self._run(['--unset-all', var], where=where, reload=reload)
+
+
+def rewrite_url(cfg, url):
+    """Any matching 'url.<base>.insteadOf' configuration is applied
+
+    Any URL that starts with such a configuration will be rewritten
+    to start, instead, with <base>. When more than one insteadOf
+    strings match a given URL, the longest match is used.
+
+    Parameters
+    ----------
+    cfg : ConfigManager or dict
+      dict-like with configuration variable name/value-pairs.
+    url : str
+      URL to be rewritten, if matching configuration is found.
+
+    Returns
+    -------
+    str
+      Rewritten or unmodified URL.
+    """
+    insteadof = {
+        # only leave the base url
+        k[4:-10]: v
+        for k, v in cfg.items()
+        if k.startswith('url.') and k.endswith('.insteadof')
+    }
+    prev_url = None
+    while url != prev_url:
+        prev_url = url
+        # all config that applies
+        matches = {k: len(v) for k, v in insteadof.items()
+                   if url.startswith(v)}
+        # find longest match, like Git does
+        if matches:
+            rewrite_base, match_len = sorted(
+                matches.items(),
+                key=lambda x: x[1],
+                reverse=True,
+            )[0]
+            url = '{}{}'.format(rewrite_base, url[match_len:])
+    return url
+
+
+# for convenience, bind to class too
+ConfigManager.rewrite_url = rewrite_url
--- a/datalad/config.py
+++ b/datalad/config.py
@@ -749,21 +749,32 @@ def rewrite_url(cfg, url):
         for k, v in cfg.items()
         if k.startswith('url.') and k.endswith('.insteadof')
     }
+
     # all config that applies
     matches = {
-        key: len(v)
+        key: v
         for key, val in insteadof.items()
         for v in (val if isinstance(val, tuple) else (val,))
         if url.startswith(v)
     }
     # find longest match, like Git does
     if matches:
-        rewrite_base, match_len = sorted(
+        rewrite_base, match = sorted(
             matches.items(),
-            key=lambda x: x[1],
+            key=lambda x: len(x[1]),
             reverse=True,
         )[0]
-        url = '{}{}'.format(rewrite_base, url[match_len:])
+        if sum(match == v for v in matches.values()) > 1:
+            lgr.warning(
+                "Ignoring URL rewrite configuration for '%s', "
+                "multiple conflicting definitions exists: %s",
+                match,
+                ['url.{}.insteadof'.format(k)
+                 for k, v in matches.items()
+                 if v == match]
+            )
+        else:
+            url = '{}{}'.format(rewrite_base, url[len(match):])
     return url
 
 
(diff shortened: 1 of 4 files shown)


datalad/config.py
```diff
@@ -722,3 +722,49 @@ def unset(self, var, where='dataset', reload=True):
...
+        # only leave the base url
+        k[4:-10]: v
+        for k, v in cfg.items()
+        if k.startswith('url.') and k.endswith('.insteadof')
+    }
+    prev_url = None
+    while url != prev_url:
+        prev_url = url
+        # all config that applies
+        matches = {k: len(v) for k, v in insteadof.items()
```
bpoldrack: I do see one problem here: You match all configs (well, the longest). However, what if I have a `--global` config that is long and a `--local` overwrite which is shorter? I think this should use `ConfigManager`'s get, check for a returned tuple and consider order therein.

mih: That is not how gitconfig works:

```
% git config --show-origin -l | grep 'github.*insteadof'
file:/home/mih/.gitconfig       url.git@github.com:.insteadof=github:
% git config --add 'url.git@github.com:.insteadof' 'mike:'
% git config --show-origin -l | grep 'github.*insteadof'                                                      
file:/home/mih/.gitconfig       url.git@github.com:.insteadof=github:
file:.git/config        url.git@github.com:.insteadof=mike:
% git config --get 'url.git@github.com:.insteadof'
mike:
```

Local overwrites global, it does not append.

bpoldrack: Actually, my suggestion is wrong. The usual tuple considers the keys, not the values. So, `get` doesn't work. Still, I think same replacements should overwrite each other. So, assuming `items` yielding in correct order (least significant first), that `insteadof` dict, needs to be based on the values, I guess.

mih: I think I don't get it yet. What is it that this function does or doesn't that git itself does when rewriting URLs. The point here is to be as clever (no more, no less) than Git.

bpoldrack: > The point here is to be as clever (no more, no less) than Git.
I absolutely agree.
I was talking about using the same "label" with different replacements (basically ria usecase). So, by definition the matches don't differ in length. And I thought that git itself would overwrite them, if specified at different levels. However, that doesn't seem to be the case.

```
/tmp/some$ git config --add --global url."example.com:".insteadOf ex:
/tmp/some$ git config --add --local url."other.com:".insteadOf ex:
/tmp/some$ git remote add test ex:/some/where
/tmp/some$ git remote -v
test    example.com:/some/where (fetch)
test    example.com:/some/where (push)
```
That's quite unexpected to me.

bpoldrack: It might also well be, that the two of user interpret differently what length is considered by git. 

> When more than one insteadOf strings match a given URL, the longest match is used.

In my example I'd consider `ex:` the thing that needs to be matched and whose length is relevant. If it's actually the length of the replacement it's a different story. But then we need to have better wording than git itself.

bpoldrack: Last for now: Tried to somehow apply local config for that `ex:` label above. So, length of replacement doesn't matter, length of label does (as it should judging by my understanding of the docs). However, at equal length apparently `--global` takes precedence over `--local`, which I find weird.
Not sure, whether we really account for that here.

mih: The length of the match is relevant, not the length of the replacement (see `len(v)` where `v` is the value of the config).

yarikoptic:
Just FTR for myself (and others):
- git does rewrites as part of the preparing configuration record for an existing remote
  - `git clone` first initiates repo, then adds remote, and then operates on it
- so it is unlikely to be able to query `git` solely to get a url rewritten (without creating a throw away git repo, adding remote, and querying its config)

mih:
Verdict in call was to warn whenever we discover two `url..insteadof` settings with the same value.

datalad/config.py
kyleam: It looks like te discussion ended up being about unexpected behavior that is a consequence of git's decision to use the replacement as the key (discussed [here][0]).  But I think Ben's point about the value potentially being a tuple still needs to be considered to be safe.  Consider for example the unlikely event that the user has repeat values for the same key:

    $ cd $(mktemp -d --tmpdir ga-XXXXXXX)
    $ git init
    $ git config url.'bar:'.insteadof foo:
    $ git config --add url.'bar:'.insteadof foo:
    $ git remote add test bar:abc
    $ git remote -v
    test	bar:abc (fetch)
    test	bar:abc (push)

Here's what `rewrite_url` would currently return.

    $  python -c 'from datalad import config; cfg = config.ConfigManager(); print(config.rewrite_url(cfg, "foo:abc"))'
    bar:o:abc

That's because the value is a tuple and the length is calculated as 2:

    $ python -c 'from datalad import cfg; print(cfg.get("url.bar:.insteadof"))'
    ('foo:', 'foo:')

[0]: https://public-inbox.org/git/20100802153720.GA28481@fishbowl.rw.madduck.net/T/#u

datalad/config.py
```diff
@@ -722,3 +722,49 @@ def unset(self, var, where='dataset', reload=True):
...
+      Rewritten or unmodified URL.
+    """
+    insteadof = {
+        # only leave the base url
+        k[4:-10]: v
+        for k, v in cfg.items()
+        if k.startswith('url.') and k.endswith('.insteadof')
+    }
+    prev_url = None
+    while url != prev_url:
```
kyleam: Hmm, git itself doesn't recursively resolve the `insteadof` values.  Is there a concrete use case for this?

mih: I found it to be an awesome thing to do, but now I cannot think of an actual need for it. Will remove (but keep the memory of having done it close to my heart).

datalad/config.py
mih: Fixed.

datalad/config.py
mih: Done.

mih:
This should now address all corner-cases, except for Git's own weaknesses. I think this is ready to merge, and we should start using it.

Issue #4066, opened 2020-01-21
RF `create_sibling` with IO support from `ria_remote`


mih:
ATM `create_sibling` has custom code to perform a bunch of operations via SSH that are equivalents of:
- `exist()`
- `rmtree()`
- `chmod`
- `chgrp`
- `rm`
- `mkdir()`
- `git update-server-info`
- `git config`
- enable post-update hook
- `annex init`

All those commands are issued one-by-one and are relatively slow as a result (my target machine needs 600ms to give me a new shell through an existing socket). RIARemote has IO classes that can do (some of) these things much faster, but ATM they are interwoven with the remote code. It would be good to RF them, move them into -core, and use them in `create-sibling` too. Once we move the actual RIARemote implementation, this would be the second user.

mih:
This issue is describing one of many aspects that require a general overhaul of the `create-sibling` command. This overhaul effort is being tracked by the meta-issue #6149. I am closing this issue here, after having added it to that TODO list.

PR #4068, opened 2020-01-22
NF: Command abstractions

mih:
ATM This is more thinking out loud than a concrete implementation. The idea
is this:

Let's collect basic operation into classes that can take care of doing
the right thing under different conditions in order to achieve this.
At the moment, we have the Runner to help with that, but the actual
nature of the operations done are scattered throughout the code.

This limits us, and, for example, is something that contributes to the
fact that something like `create_sibling()` can only work with
POSIX-like system, via SSH connections. Moving the local and remote
operation it performs into helper classes that can be specialized for any
target scenario, would help simplify the implementation of that command,
while simultaneously make it more flexible, e.g. it could discover
that the target sibling is at a local path, and internally pick a local
operations helper and achieve the desired result without having to go
through SSH. #4066

Likewise, any, now hardcoded, command execution could be centrally
switched to a variant that is optimized for the local platform
(look for `LocalOperations` in the diff).

Another angle for mid-term improvements would be that less code talks
directly to a Runner. Having specialized methods for particular
operations would enable us to optimize the execution to meet our needs
better.

I envision no limit to the scope of supported operations. For example,
something like `enable_post_update_hook(path)` would be perfectly
in-scope. But even `edit_textfile_interactively(path)` is easily within
the range of possibilities. It would be important, though, that the
classes remain cheap to instantiate.

It is also fine, if not all operations are implemented for all potential
target scenarios immediately. However, the API of an operation would
need to be carefully crafted to avoid overfitting any particular
environment or use case.

We should also consider a similar approach for interfacing with Git
commands, in particular those that report paths -- which are reported in
POSIX conventions, but were/are inappropriately treated as native paths,
leading to failure on Windows. However, it is less clear to me how to
consolidate this with our current approach (ie.  GitRepo.call_git...()).

In conclusion, I'd argue that having such an abstraction layer would
benefit old code by making it more flexible with little effect, and aid
the design of new code that could be written with the fact in mind that
low-level operations are taken care off in a platform-appropriate
fashion, but also that they have to be implemented in a
platform-appropriate fashion, and where this can be done.

[skip ci]

### DIFF
--- a/datalad/support/operations/operations_abstract.py
+++ b/datalad/support/operations/operations_abstract.py
@@ -190,3 +190,25 @@ class RemoteOperationsBase(OperationsBase):
         if not path.is_absolute() and self._remote_cwd:
             path = self._remote_cwd / path
         return path
+
+    def get(self, source, destination, recursive=False, preserve_attrs=False):
+        # TODO: Doc. Not quite sure yet. For now signature is copied from
+        #       SSHConnection. Trying to use it delegating to SSHConnection from
+        #       within RemoteSSHShellOperations and differently from within
+        #       RemotePersistentSSHShellOperations. Let's see how that works.
+        #
+        #       Eventually we might even want get/put to be "copy" instead,
+        #       that can be performed locally as well. Would most consistent
+        #       with base idea about those command abstractions.
+        raise NotImplementedError
+
+    def put(self, source, destination, recursive=False, preserve_attrs=False):
+        # TODO: Doc. Not quite sure yet. For now signature is copied from
+        #       SSHConnection. Trying to use it delegating to SSHConnection from
+        #       within RemoteSSHShellOperations and differently from within
+        #       RemotePersistentSSHShellOperations. Let's see how that works.
+        #
+        #       Eventually we might even want get/put to be "copy" instead,
+        #       that can be performed locally as well. Would most consistent
+        #       with base idea about those command abstractions.
+        raise NotImplementedError
--- a/datalad/support/operations/operations_abstract.py
+++ b/datalad/support/operations/operations_abstract.py
@@ -32,7 +32,7 @@ class OperationsBase(object):
     def _ensure_absolute(self, path):
         """Internal helper to make a path absolute
 
-        If relative path is given, interprete it as relative to initialized CWD
+        If relative path is given, interpret it as relative to initialized CWD
         if any, to PWD otherwise.
 
         Parameters
@@ -117,6 +117,7 @@ class OperationsBase(object):
         # Need a unified behavior across subclasses
         raise NotImplementedError
 
+    # TODO: What do we want to do with symlinks? Follow? Make an option?
     def change_permissions(self, path, mode, recursive=False):
         """Change the permissions for a path
 
@@ -131,6 +132,7 @@ class OperationsBase(object):
         """
         raise NotImplementedError
 
+    # TODO: What do we want to do with symlinks? Follow? Make an option?
     def change_group(self, path, label, recursive=False):
         """Change the group ownership for a path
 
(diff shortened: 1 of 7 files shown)


bpoldrack:
I think having this layer is a desirable first step towards "better" execution (re all the ideas and desires around building new Runner(s) ). I'd like to mention that this idea arose from having a very similar approach in git-annex-ria-remote, that can operate locally as well as via a remote SSH shell using this very mechanism.

>  In conclusion, I'd argue that having such an abstraction layer would
benefit old code by making it more flexible with little effect, and aid
the design of new code that could be written with the fact in mind that
low-level operations are taken care off in a platform-appropriate
fashion, but also that they have to be implemented in a
platform-appropriate fashion, and where this can be done.

I very much agree.

What's somewhat missing (not actually conflicting with the concept) is a clear way of communication. Interface for signatures is fine, but we should also define what to expect from those functions. When are what kind of exceptions to be expected, how to get see stdout/stderr (if any). Higher level code should not be aware which subclass it is currently using, of course (that's the point), therefore there needs to be a uniform way of dealing with return values, errors, logs, progress(?).

kyleam:
The high-level idea sounds like a good direction to go, though as in most cases I think the details (e.g., things @bpoldrack already mentioned) will be what's tricky and require more thought.

> However, the API of an operation would
> need to be carefully crafted to avoid overfitting any particular
> environment or use case.

That seems exceptionally hard, particularly when there is no limit in scope and the use cases come in at different times.

I think we should make it very clear that nothing in these classes should be considered stable for use outside of DataLad.

mih:
> I think we should make it very clear that nothing in these classes should be considered stable for use outside of DataLad.

Amen! I should have written that on top and in **BOLD**.

bpoldrack:
> I think we should make it very clear that nothing in these classes should be considered stable for use outside of DataLad.

Agree. Reconsidering my thoughts above, it probably should be limited in scope in that there's a clear statement that this meant to be a very low level layer. No output magic whatsoever, just plain execution helper. Anything more sophisticated needs to be a caller of this layer.
My point here is, to have a clear guideline what can go in there and how. So, my thoughts about results, progress, whatever might simply not apply, making this easier but also defining some kind of a scope limit.

mih:
I think it will be hard to come up with a general setup that works for all helpers uniformly. Some will only need the ability to crash-report via exception, some will need to return output, who knows what else.

I think if we can properly wrap a single operation into such a method, make it do whatever makes sense, the primary concern is to have its input/output mapping be uniform (enough) across platforms.

bpoldrack:
Agree. Just wonder whether we can actually articulate a line that shouldn't be crossed within those classes to avoid growing it all too magical again (see Runner).

bpoldrack:
FTR: Working on pushing this forward. Will include basic operations for remote execution and respective tests. In addition will try to replace ria-remote's own implementation with this one.

yarikoptic:
I think this PR is missing "interconnect" with #1680 and https://github.com/datalad/datalad/issues/1680#issuecomment-584219222

datalad/support/operations/operations_abstract.py
```diff
@@ -0,0 +1,192 @@
...
+        path : Path
+          Path to test.
+
+        Returns
+        -------
+        bool
+          True, if the path exists (even in case of a broken symlink), False
+          otherwise.
+        """
+        raise NotImplementedError
```
yarikoptic: `listdir` or alike? then could be used in `create_sibling` safeguard: https://github.com/datalad/datalad/pull/4147/files#diff-865501de3883a6b49b32476bdc7b3380R305

bpoldrack: Yes, agree.

bpoldrack:
Update: Rebased on back on it.

mih:
OK, no point in pretending someone would be working on this.

PR #4070, opened 2020-01-22
BF: Depend on Gitpython >= 2.1.12

bpoldrack:


bpoldrack:
@kyleam : Changed commit message to include my comment above. Will cancel new CI runs, since nothing changed besides that commit message.

Issue #4071, opened 2020-01-22
ConfigManager(ds) reads unexpected configuration(from PWD)!


mih:
Just stumbled over this while debugging(I hope it has nothing to do with the changeset I am working on).

When PWD is in the root of a dataset, but `ConfigManager(ds=<someotherdataset>)` the config file in `PWD/.git/config` seems to be read nevertheless.

This is rather unfortunate, and could be the source of many confusing bugs.

mih:
```
        if dataset is not None:
            # make sure we run the git config calls in the dataset
            # to pick up the right config files
            run_kwargs['cwd'] = dataset.path
 ```

suggests otherwise.

**Comment by mih** [comment]
But

```
(datalad3-dev) mih@meiner /tmp % TMPDIR="/var/tmp/sym link" python -m nose -v -s datalad.core.distributed.tests.test_clone:test_ria_http --pdb
...
(Pdb) cloneds.config.keys()
dict_keys(['user.name', 'user.email', 'datalad.datasets.topurl', 'datalad.log.level'])
(Pdb) cloneds.path
'/var/tmp/sym link/datalad_temp_tree_test_ria_http9a9nlirq/clone/subdir/subds'
```

and 

```
(datalad3-dev) mih@meiner ~/hacking/datalad/git (git)-[enh-clonesubdatasets] % TMPDIR="/var/tmp/sym link" python -m nose -v -s datalad.core.distributed.tests.test_clone:test_ria_http --pdb
(Pdb) cloneds.config.keys()
dict_keys(['user.name', 'user.email', 'core.repositoryformatversion', 'core.filemode', 'core.bare', 'core.logallrefupdates', 'hub.upstream', 'hub.forkremote', 'hub.pullbase', 'remote.gh-mine.url', 'remote.gh-mine.fetch', 'remote.gh-mine.annex-ignore', 'branch.master.remote', 'branch.master.merge', 'remote.gh-master.url', 'remote.gh-master.fetch', 'remote.gh-master.annex-ignore', 'remote.gh-yarik.url', <snip>)
(Pdb) cloneds.path
'/var/tmp/sym link/datalad_temp_tree_test_ria_http3owtffgd/clone/subdir/subds'
```

**Comment by mih** [comment]
Holy shit, this is real:

```
        if line.startswith('file:'):
            # origin line
            fileset.add(abspath(line[5:]))
            continue
```

[...]

PR #4072, opened 2020-01-22
Fix ConfigManager file tracking

mih:
This is a fix for #4071, in particular:

- [x] track `.git/config` from a given dataset, not something that just happens to be found in `PWD`
- [x] prevent invalid paramterization of `ConfigManager()`
- [x] enfore CWD for `git-config` calls to be a dataset's root dir, whenever it exists
- [x] prevent accidental interaction with a dataset at PWD, when a `ConfigManager` was parameterized with a dataset instance that has no equivalent on disk
- [x] quite a few additional tests

### DIFF
--- a/datalad/config.py
+++ b/datalad/config.py
@@ -14,7 +14,10 @@ from datalad.consts import (
     DATASET_CONFIG_FILE,
     DATALAD_DOTDIR,
 )
-from datalad.cmd import GitRunner
+from datalad.cmd import (
+    GitRunner,
+    CommandError,
+)
 from datalad.dochelpers import exc_str
 from distutils.version import LooseVersion
 
@@ -26,6 +29,7 @@ from os.path import (
     getmtime,
     abspath,
     isabs,
+    expanduser,
 )
 from time import time
 
@@ -195,6 +199,11 @@ class ConfigManager(object):
                 raise ValueError('Refuse to combine legacy dataset_only flag, with '
                                  'source setting')
             source = 'dataset'
+        if dataset and source in ('local', 'dataset-local') \
+                and not dataset.is_installed():
+            raise ValueError(
+                'Cannot manage local configuration of an absent dataset'
+            )
         # store in a simple dict
         # no subclassing, because we want to be largely read-only, and implement
         # config writing separately
@@ -213,7 +222,7 @@ class ConfigManager(object):
         if overrides is not None:
             self.overrides.update(overrides)
         if dataset is None:
-            if source == 'dataset':
+            if source in ('dataset', 'dataset-local'):
                 raise ValueError(
                     'ConfigManager configured to read dataset only, '
                     'but no dataset given')
@@ -229,10 +238,6 @@ class ConfigManager(object):
         # any "facilitated" leakage -- just disable logging of outputs for
         # this runner
         run_kwargs = dict(log_outputs=False)
-        if dataset is not None:
-            # make sure we run the git config calls in the dataset
-            # to pick up the right config files
-            run_kwargs['cwd'] = dataset.path
         self._runner = GitRunner(**run_kwargs)
         try:
             self._gitconfig_has_showorgin = \
@@ -299,10 +304,33 @@ class ConfigManager(object):
 
         if self._src_mode == 'dataset-local':
             run_args.append('--local')
-        stdout, stderr = self._run(run_args, log_stderr=True)
-        self._store, self._cfgfiles = _parse_gitconfig_dump(
-            stdout, self._store, self._cfgfiles, replace=True,
-            cwd=self._runner.cwd)
+
+        dspath_exists = self._dataset_path and exists(self._dataset_path)
+        if not dspath_exists:
+            # if the dataset is not here (yet), we prevent reading from
+            # CWD in case this happens to be another dataset
+            run_args.append(
+                '--global'
+                # gitconfig would fail with --global if it is not here
+                if exists(opj(expanduser('~'), '.gitconfig'))
+                else '--system')
+
+        try:
+            stdout, stderr = self._run(
+                run_args,
+                log_stderr=True,
+            )
+            self._store, self._cfgfiles = _parse_gitconfig_dump(
+                stdout, self._store, self._cfgfiles, replace=True,
+                cwd=self._dataset_path if dspath_exists else None)
+        except CommandError as e:
+            if 'unable to read config file' in e.stderr:
+                lgr.debug(
+                    'Cannot read non-dataset config, likely there is none: %s',
+                    exc_str(e),
+                )
+            else:
+                raise
 
         # always monitor the dataset cfg location, we know where it is in all cases
         if self._dataset_cfgfname:
@@ -587,7 +615,18 @@ class ConfigManager(object):
         """
         if where:
             args = self._get_location_args(where) + args
-        out = self._runner.run(['git', 'config'] + args, **kwargs)
+        dspath_exists = self._dataset_path and exists(self._dataset_path)
+        if where in ('local', 'dataset') and not dspath_exists:
+            raise ValueError(
+                'Dataset does not exist at {}, cannot write to {} '
+                'configuration.'.format(
+                    self._dataset_path,
+                    where,
+                ))
+        out = self._runner.run(
+            ['git', 'config'] + args,
+            cwd=self._dataset_path if dspath_exists else None,
+            **kwargs)
         if reload:
             self.reload()
         return out
--- a/datalad/config.py
+++ b/datalad/config.py
@@ -172,19 +172,21 @@ class ConfigManager(object):
       Legacy option, do not use.
     overrides : dict, optional
       Variable overrides, see general class documentation for details.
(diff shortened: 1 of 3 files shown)


datalad/config.py
```diff
@@ -81,7 +82,10 @@ def _parse_gitconfig_dump(dump, store, fileset, replace):
             continue
         if line.startswith('file:'):
             # origin line
-            fileset.add(abspath(line[5:]))
+            fname = line[5:]
+            if not isabs(fname):
+                fname = opj(cwd, fname) if cwd else abspath(fname)
+            fileset.add(fname)
```
kyleam: Fix looks good.

As far as I can see, the main consequence of the faulty expansion here is that the mtime check with `.reload(force=False)` could be wrong.

kyleam:
> This is chasing #4071, but is not yet the real thing

I tried for a bit to find a scenario that triggered what's described/shown in gh-4071.  Here is the only thing I could come up with:

```python
import tempfile
import datalad.api as dl
from datalad.utils import chpwd

ds = dl.Dataset(tempfile.mkdtemp(prefix="dl-a-")).create()
ds.config.set("a.b.c", "d", where="local")

with chpwd(ds.path):
    ds_b = dl.Dataset(tempfile.mkdtemp(prefix="dl-b-")) # repo doesn't yet exist
    print(ds_b.config.get("a.b.c"))                     # => d
    ds_b.create()                                       # now exists
    print(ds_b.config.get("a.b.c"))                     # => None
```

I haven't actually stepped through this, but here is what I think is happening.  When the repo isn't yet created, `Dataset.config` generated uses `ConfigManager(dataset=None, dataset_only=False)`.  That means the `git config` dump is from the current directory (i.e., `ds.path` above), so its local configuration is included.

To fix this, I think we need a way to call `ConfigManager` that means "only include git's --global and --system configuration".  I suspect the easiest way to do that would be to extend the `source` parameter and then based on that value decide whether to instruct `_parse_gitconfig_dump` to ignore all relative paths.

And I think we should also use this new mode when instantiating `datalad.cfg`.

mih:
Thanks for the analysis! That brings some sanity. I agree re source more addition, and I have actually implemented one already, and will add the one you suggested too.

datalad/config.py
mih: Correct.

mih:
Ok, I think this is it ... 24 f?PKWDWing hours later... I would appreciate reviews and any performance fix you can come up with (but maybe later...primary focus is correct operation...this time).

thx!

datalad/config.py
```diff
@@ -283,16 +294,43 @@ def reload(self, force=False):
...
-        self._store, self._cfgfiles = _parse_gitconfig_dump(
-            stdout, self._store, self._cfgfiles, replace=True)
+        if self._src_mode == 'dataset-local':
+            run_args.append('--local')
+
+        dspath_exists = self._dataset_path and exists(self._dataset_path)
+        if not dspath_exists:
+            # if the dataset is not here (yet), we prevent reading from
+            # CWD in case this happens to be another dataset
+            run_args.append(
```
bpoldrack: Shouldn't that happen only if `self._cfgfiles` do indeed include global/system configs?
Since docstring reads "Reload all configuration items from the configured sources". Now, those configured sources aren't actually configured yet when `reload ` is called first by `__init__`. So, may be store `__init__`'s `source` and consider it here?

mih:
I would be grateful if someone with vcrpy knowledge could look that this test failures:
```======================================================================
FAIL: datalad.distribution.tests.test_create_github.test_integration1_datalad_tester_org
  File "/home/mih/env/datalad3-dev/lib/python3.7/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/mih/hacking/datalad/git/datalad/tests/utils.py", line 141, in newfunc
    return func(*args, **kwargs)
  File "/home/mih/env/datalad3-dev/lib/python3.7/site-packages/vcr/cassette.py", line 99, in __call__
    return type(self)(self.cls, args_getter)._execute_function(function, args, kwargs)
  File "/home/mih/env/datalad3-dev/lib/python3.7/site-packages/vcr/cassette.py", line 113, in _execute_function
    return self._handle_function(fn=handle_function)
  File "/home/mih/env/datalad3-dev/lib/python3.7/site-packages/vcr/cassette.py", line 137, in _handle_function
    return fn(cassette)
  File "/home/mih/env/datalad3-dev/lib/python3.7/site-packages/vcr/cassette.py", line 106, in handle_function
    return function(*args, **kwargs)
  File "/home/mih/hacking/datalad/git/datalad/tests/utils.py", line 1498, in newfunc
    ret = t(*args, **kwargs)
  File "/home/mih/hacking/datalad/git/datalad/distribution/tests/test_create_github.py", line 156, in test_integration1_datalad_tester_org
    oauthtokens=['fake1', 'fake2']
  File "/home/mih/hacking/datalad/git/datalad/tests/utils.py", line 684, in newfunc
    return t(*(args + (keyring,)), **kwargs)
  File "/home/mih/hacking/datalad/git/datalad/tests/utils.py", line 732, in newfunc
  File "/home/mih/hacking/datalad/git/datalad/distribution/tests/test_create_github.py", line 183, in check_integration1
    res = ds.create_sibling_github(repo_name, **kwargs)
  File "/home/mih/hacking/datalad/git/datalad/distribution/dataset.py", line 500, in apply_func
    return f(**kwargs)
  File "/home/mih/hacking/datalad/git/datalad/distribution/create_sibling_github.py", line 200, in __call__
    existing, access_protocol, dryrun)
  File "/home/mih/hacking/datalad/git/datalad/support/github_.py", line 298, in _make_github_repos
    github_organization):
  File "/home/mih/hacking/datalad/git/datalad/support/github_.py", line 272, in _gen_github_entity
    for ses, cred in _gen_github_ses(github_login, github_passwd):
  File "/home/mih/hacking/datalad/git/datalad/support/github_.py", line 197, in _gen_github_ses
    text="Do you want to try (re)entering GitHub credentials?"
  File "/home/mih/hacking/datalad/git/datalad/ui/base.py", line 41, in yesno
    response = self.question(*args, choices=['yes', 'no'], **kwargs).rstrip('\n')
  File "/home/mih/hacking/datalad/git/datalad/ui/dialog.py", line 361, in question
    "We are asked for a response whenever none is left to give"
AssertionError: We are asked for a response whenever none is left to give


```

datalad/config.py
bpoldrack: Another thing: `dspath_exists`:
I wonder whether this is sufficient. The question is whether at that location that particular dataset exists, isn't it? Existence of the path doesn't imply that there's some config file to read from or even an actual dataset at that place. Currently not quite sure about the implications, but looking a few lines down, there's `325     cwd=self._dataset_path if dspath_exists else None)` for ` _parse_gitconfig_dump`, while the `_run` call before doesn't get that passed based on the same criteria. Which means, that `_parse_gitconfig_dump` does the file path completion based on the assumption, that the call ran in that CWD, but that's not clear to me.

kyleam:
Hmm so this series went in a pretty different direction then [where I thought it would][0] last night.  I need to take a closer look to really understand why, but my first impression is that we should try to avoid an approach that does things like

```python
            run_args.append(
                '--global'
                # gitconfig would fail with --global if it is not here
                if exists(opj(expanduser('~'), '.gitconfig'))
                else '--system')
```

That assumes too much about where git gets its values from.  For example, perhaps the user keeps their configuration in `$XDG_CONFIG_HOME/git/config`.  I suspect this is why on my system, with 9e35ee00b (TST: Adjust tests for new, more correct behavior) checked out, I see:

```
$ git status                                    
fatal: not a git repository (or any of the parent directories): .git
$ python -c "import datalad"                    
It is highly recommended to configure Git before using DataLad. Set both 'user.name' and 'user.email' configuration variables.
It is highly recommended to configure Git before using DataLad. Set both 'user.name' and 'user.email' configuration variables.
```

[0]: https://github.com/datalad/datalad/pull/4072#issuecomment-577503919

mih:
> we should try to avoid an approach that does things like

The only alternative that I see would be to try to actually make the call, catch the error, hope it is related to `--global` and try again with `--system`. This would involve substantial AFAICS. Without `--global` it will shamelessly read from PWD.

kyleam:
> The only alternative that I see would be to try to actually make the call, catch the error, hope it is related to `--global` and try again with `--system`.

I agree with you that that's not a good alternative.  On top of the reasons you mentioned, in the case I'm pointing out, calling `--system` would still either fail (because the corresponding file doesn't exist) or incorrectly limit the set of "non-repository" configuration values that are considered.  (The second issue, just in the other direction, is true for the use of `--global`.)

Let me try to come up with an alternative approach (in patch form).  I can think a couple of things worth exploring, none of them being the relative path filtering idea I mentioned last night, which is flawed because there are scenarios where repository config values (`$GITDIR/config` but also `$GITDIR/worktrees/$WTREE/config.worktree`) will be reported as absolute paths.

mih:
This has been replaced by a superior approach #4078

PR #4073, opened 2020-01-23
BF: Make sure source candidate configs for submodules are inherited

mih:
[ this is sitting on top of #4072 and hence inherits its issues]

This makes it possible to have large collections of datasets put in
some dataset store, without any of them carrying knowledge about their
hosting location, even if they have super-sub-dataset relationships
among themselves. The only requirement to make any of these datasets
accessible, is a single top-level superdataset with a source candidate
configuration that is point to the store. Everything else will be
handle with on-clone configuration in `.git/config` of each subdataset.

### DIFF
--- a/datalad/core/distributed/clone.py
+++ b/datalad/core/distributed/clone.py
@@ -456,6 +456,7 @@ def clone_dataset(
         if cand['type'] == 'ria':
             yield from postclonecfg_ria(destds, cand)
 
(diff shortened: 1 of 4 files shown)


mih:
Rebased on #4078 will merge when tests are happy.

kyleam:
Very nice (and a pleasant series to read).  Thanks.

Issue #3766, opened 2019-10-10
OPT: up to 1.7 times slower subdatasets discovery due to #3705 RF


yarikoptic:
I have pushed a fresh run of benchmarks. 
 The topmost hist on [regressions](http://datalad.github.io/datalad/#/regressions) is api.supers.time_subdatasets with x3.57 factor.  It points to recently merged #3705, and reconfirmed by the travis benchmarks run for it back then (we should find a way to communicate it back to the github PR status!): https://travis-ci.org/datalad/datalad/jobs/589416808#L1253 seems to be affecting all subdatasets-related operations .

mih:
Thanks for catching that. Below is the profile graph. Clearly the removed caching of this information by gitpy has an impact. However, I see potential for improvements without having to (re-)introduce caching. For example `is_with_annex()` should make one call to `for_each_ref()` instead of going through convienence helpers that call it twice.

![image](https://user-images.githubusercontent.com/136479/66630328-44cfbc80-ec04-11e9-95d3-0bac405e9d7b.png)

@bpoldrack also mentioned that other optimizations are possible.

yarikoptic:
With current state of master, it is still quite impacted:
```shell
drogon:~/proj/datalad/datalad
*$> asv compare --machine drogon-neurodebian10.20190322.1  0.11.8-1-gd2101409d 0.12.0rc5-262-g4a7947bc9 | grep -v failed
grep: warning: GREP_OPTIONS is deprecated; please use an alias or script

All benchmarks:

       before           after         ratio
     [d2101409]       [4a7947bc]
     <0.11.x>         <master>  
-        505~10ms          387~3ms     0.77  api.supers.time_createadd [drogon-neurodebian10.20190322.1/virtualenv-py3.7]
         490~10ms          497~7ms     1.01  api.supers.time_createadd_to_dataset [drogon-neurodebian10.20190322.1/virtualenv-py3.7]
+      1.27~0.01s       1.63~0.02s     1.28  api.supers.time_installr [drogon-neurodebian10.20190322.1/virtualenv-py3.7]
         72.3~3ms         58.7~2ms    ~0.81  api.supers.time_ls [drogon-neurodebian10.20190322.1/virtualenv-py3.7]
+         407~4ms          513~7ms     1.26  api.supers.time_ls_recursive [drogon-neurodebian10.20190322.1/virtualenv-py3.7]
[...]

Issue #3333, opened 2019-04-14
Have a critical look at the (internal) use of Repo.methods


mih:
Here is an initial list of methods that gather information that is also provided by `GitRepo.get_content_info()` or `AnnexRepo.get_content_annexinfo()`, or `GitRepo.status()` (in rare cases). Each of them individually is typically a little bit faster then the two above, however, they only get a single piece of info, while the others get all of them combined. So whenever more than one of the below is needed, an RF to using the new methods would translate into a speed-up.

AnnexRepo:
- `get_file_key()`
- `find()`
- `file_has_content()`
- `is_under_annex()`
- `info()`
- `get_annexed_files()`
- `get_contentlocation()`
- `get_file_backend()`

GitRepo:
- get_indexed_files()` [GitPy]
- get_files()` [GitPy]
- is_submodule_modified() `
- get_changed_files()`
- `get_missing_files()`
- `get_deleted_files()`

mih:
https://github.com/datalad/datalad/pull/4169 removed `get_changed_files()`

PR #4077, opened 2020-01-24
OPT(dirty attempt): memoize is_with_annex value

yarikoptic:
I had spotted that identical for-each-ref was ran twice during a simple
datalad status in a regular git repository... investigation shown that
multple code paths from dataset.py lead to the same is_with_annex.
Each call was taking about 45 msec on my speedy laptop.  Oh well, not that
much, but trying to run  datalad rerun of a single commit, resulted in
over 20 of such calls.  That is already a second of wasted time for a single rerun.

Proposed ATM solution is not full.  I am not sure yet what solution we
want to see though (memoization and resetting upon update/add_remote/etc),
or some more universal check (anything in git has changed), or just reimplement using GitPython ;)  But just wanted to see what breaks if anything, and either there would be a notable impact
on benchmarks.

Most likely similar "let's just do it again" invocations happen for other properties of a repository now as well.

### DIFF
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -641,6 +641,7 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
         elif self.dot_git.is_symlink():
             self.dot_git = self.dot_git.resolve()
 
+        self._is_with_annex = None  # unknown yet
         self._valid_git_test_path = self.dot_git / 'HEAD'
         _valid_repo = self.is_valid_git()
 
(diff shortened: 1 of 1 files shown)


bpoldrack:
Hm. You would then need to consider all cases where this could possibly change. That's at least any `git-annex-init` call, anything that fetches a remote. Plus: Any usage of datalad python API where such an operation is performed not going through datalad will break. Then there's `git-annex-uninit` as well.

So, I'd not make that lazy at the low level, but only within higher level code. So, if `status` checked that already it can possibly avoid to check it again.

mih:
To me this is a specific variant of https://github.com/datalad/datalad/issues/3333

I agree with @bpoldrack that there are limits to what can be done at the lowest level. In general, we have been moving further and further away from "clever" processing in low-level code for this reason (although the dreaded `normalize_paths` is still making my life miserable).

IMHO what we should be doing is consequently reduce/remove code that assembles this kind of information bit-by-bit and have "clever" higher-level code that actually can "know" when and for how long assumptions can be valid.

In this particular case, maybe some use of `is_with_annex` can be replaced by `is_valid_annex`, if that is faster -- and a grasp the subtle difference in the name correctly ;-)

Likely, a key contributor to the overuse of this method are access patterns like:

```
for stuff in otherstuff:
    ds.repo.something()
```

which is expensive in many ways, and better be:

```
repo = ds.repo
for stuff in otherstuff:
    repo.something()
```

mih:
FTR (just stumbled over this related piece of info): https://github.com/datalad/datalad/issues/3766#issuecomment-540937940

yarikoptic:
> IMHO what we should be doing is consequently reduce/remove code that assembles this kind of information bit-by-bit and have "clever" higher-level code that actually can "know" when and for how long assumptions can be valid.

not really a pragmatic/feasible solution to achieve in general - relevant code could be spread across methods etc, passing such "clever" state across explicitly is not feasible.

but oh well -- his draft PR is also not a solution so will close it

mih:
> > IMHO what we should be doing is consequently reduce/remove code that assembles this kind of information bit-by-bit and have "clever" higher-level code that actually can "know" when and for how long assumptions can be valid.
> 
> not really a pragmatic/feasible solution to achieve in general - relevant code could be spread across methods etc, passing such "clever" state across explicitly is not feasible.

It pains me a bit to see the work that I have been doing over the past 18 months being declared as infeasible. The fact that it has happened nevertheless invalidates the assessment as "not practical", though.

yarikoptic:
> It pains me a bit to see the work that I have been doing over the past 18 months being declared as infeasible. The fact that it has happened nevertheless invalidates the assessment as "not practical", though.

don't expand the scope -- I was commenting about solely about relying on "clever" coding to mitigate underlying implementation "wits", not about "the work ... [of] the past 18 months", unless it was spent exactly in that -- finding all `.repo` uses and replacing them with local bindings.

PR #4080, opened 2020-01-24
WitlessRunner() to replace gitpy.Repo.clone_from() with plain `git-clone` call, but keep progress reporting

mih:
- [x] actually replace GitPython in `clone()`
- [x] implement a leaner runner that is capable of pulling out progress info from a `git clone` run
- [x] implement a progress info extractor for `git-clone` based on GitPython's approach, but simplified
- [x] bring back progress reporting for `GitRepo.clone()
- [x] document `WitlessRunner`
- [x] find answer to test WTF
- [x] show that this is sufficient for fetch/pull/push #4087 (concrete usage needs more work, but no structural limitation were identified, which was the aim within the context of this PR)
- [x] basic tests
- [x] decide whether to capture or kill all process output, instead of relaying it to the parent by default -- decided: better to relay then to hide by default.
- [x] works just as well on windows

### DIFF
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -98,7 +98,125 @@ def _cleanup_output(stream, std):
         std.close()
 
 
-class Runner(object):
+class LeanRunner(object):
+    __slots__ = ['cwd', 'env', '_communicate_period']
+
+    def __init__(self, cwd=None, env=None, communicate_period=0.1):
+        self._communicate_period = communicate_period
+        self.env = env.copy() if env else None
+        self.cwd = cwd
+        if cwd and env is not None:
+            # if CWD was provided, we must not make it conflict with
+            # a potential PWD setting
+            self.env['PWD'] = cwd
+
+    def run(self, cmd, proc_stdout=None, proc_stderr=None, stdin=None):
+        proc_out = (proc_stdout, proc_stderr)
+        if all(p is None for p in proc_out):
+            proc_out = None
+        try:
+            process = subprocess.Popen(
+                cmd,
+                # from PY37 onwards
+                #capture_output=proc_output,
+                stdout=subprocess.PIPE if proc_out else None,
+                stderr=subprocess.PIPE if proc_out else None,
+                shell=False,
+                cwd=self.cwd,
+                env=self.env,
+                stdin=stdin,
+                # intermediate reports are never decoded anyways
+                # from PY37 onwards
+                text=False,
+                universal_newlines=False,
+            )
+        except Exception as e:
+            lgr.log(11, "Failed to start %r%r: %s" %
+                    (cmd,
+                     (" under %r" % self.cwd) if self.cwd else '',
+                     exc_str(e)))
+            raise
+
+        try:
+            out = [b'', b'']
+            last_idx = [0, 0]
+            while process.poll() is None:
+                try:
+                    # get a chunk of output for the specific period
+                    # of time
+                    pout = process.communicate(
+                        timeout=self._communicate_period,
+                    )
+                except subprocess.TimeoutExpired as poll:
+                    # this will always be the full report so far,
+                    # not just an output increment
+                    pout = (poll.stdout, poll.stderr)
+
+                if proc_out is not None:
+                    for i, (o, proc) in enumerate(zip(pout, proc_out)):
+                        if proc_out is None:
+                            out[i] += o
+                            # no index update needed, can never change
+                            continue
+                        # current length of output
+                        len_o = len(o) if o else 0
+                        # anything new in the output?
+                        if len_o - last_idx[i] > 0:
+                            # engage output processed
+                            processed, unprocessed_len = proc(o[last_idx[i]:])
+                            # only keep the processed output
+                            out[i] += processed
+                            # record point to which we processed the output
+                            # subtract number of bytes reported as unprocessed
+                            last_idx[i] = len_o - unprocessed_len
+
+            # obtain exit code
+            status = process.poll()
+
+            # decode bytes to string
+            out = tuple(o.decode('utf-8') if o else '' for o in pout)
+
+            if status not in [0, None]:
+                msg = "Failed to run %r%s." % (
+                    cmd,
+                    (" under %r" % self.cwd) if self.cwd else '',
+                )
+                raise CommandError(
+                    cmd=str(cmd),
+                    msg=msg,
+                    code=status,
+                    stdout=out[0],
+                    stderr=out[1],
+                )
+            else:
+                lgr.log(8, "Finished running %r with status %s", (cmd, status))
+
+        except CommandError:
+            # do not bother with reacting to "regular" CommandError
+            # exceptions.  Somehow if we also terminate here for them
+            # some processes elsewhere might stall:
+            # see https://github.com/datalad/datalad/pull/3794
+            raise
+
+        except BaseException as exc:
+            exc_info = sys.exc_info()
+            # KeyboardInterrupt is subclass of BaseException
+            lgr.debug("Terminating process for %s upon exception: %s",
+                      cmd, exc_str(exc))
+            try:
+                # there are still possible (although unlikely) cases when
+                # we fail to interrupt but we
+                # should not crash if we fail to terminate the process
+                proc.terminate()
+            except BaseException as exc2:
+                lgr.warning("Failed to terminate process for %s: %s",
+                            cmd, exc_str(exc2))
+            raise exc_info[1]
+
+        return out
+
+
+class Runner(LeanRunner):
     """Provides a wrapper for calling functions and commands.
 
     An object of this class provides a methods that calls shell commands or
@@ -131,9 +249,8 @@ class Runner(object):
              Switch to instruct whether outputs should be logged or not.  If not
              set (default), config 'datalad.log.outputs' would be consulted
         """
+        super(Runner, self).__init__(cwd=cwd, env=env)
 
-        self.cwd = cwd
-        self.env = env
         if protocol is None:
             # TODO: config cmd.protocol = null
             protocol_str = os.environ.get('DATALAD_CMD_PROTOCOL', 'null')
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -99,10 +99,10 @@ def _cleanup_output(stream, std):
 
 
 class LeanRunner(object):
-    __slots__ = ['cwd', 'env', '_communicate_period']
+    __slots__ = ['cwd', 'env', '_poll_period']
 
-    def __init__(self, cwd=None, env=None, communicate_period=0.1):
-        self._communicate_period = communicate_period
+    def __init__(self, cwd=None, env=None, poll_period=0.1):
+        self._poll_period = poll_period
         self.env = env.copy() if env else None
         self.cwd = cwd
         if cwd and env is not None:
@@ -127,7 +127,7 @@ class LeanRunner(object):
                 stdin=stdin,
                 # intermediate reports are never decoded anyways
                 # from PY37 onwards
-                text=False,
+                #text=False,
                 universal_newlines=False,
             )
         except Exception as e:
@@ -145,7 +145,7 @@ class LeanRunner(object):
                     # get a chunk of output for the specific period
                     # of time
                     pout = process.communicate(
-                        timeout=self._communicate_period,
+                        timeout=self._poll_period,
                     )
                 except subprocess.TimeoutExpired as poll:
                     # this will always be the full report so far,
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -144,7 +144,7 @@ class WitlessRunner(object):
             self.env['PWD'] = self.cwd
 
     def run(self, cmd, proc_stdout=None, proc_stderr=None, stdin=None,
-            poll_period=0.1):
+            poll_latency=0.1):
         """Execute a command and communicate with it.
 
         Parameters
@@ -171,7 +171,9 @@ class WitlessRunner(object):
         poll_latency : float, optional
           Shortest interval at which the running process is queried for
           output (in seconds). Any potential output processing will increase
-          the effective interval.
+          the effective interval. When the effective polling frequency is
+          too low to keep the output buffers below their maximum size,
+          a process will deadlock.
 
         Returns
         -------
@@ -226,8 +228,7 @@ class WitlessRunner(object):
             while process.poll() is None or keep_going:
                 # one last read?
                 keep_going = process.returncode is None
-                # get a chunk of output for the specific period
-                # of time
+                # get a chunk of output
                 pout = [
                     # read whatever is available, should not block
                     o.read1(-1) if p else None
@@ -249,7 +250,7 @@ class WitlessRunner(object):
                     buffer = unprocessed[i] + o if unprocessed[i] else o
                     # engage output processor
                     unprocessed[i] = _handle_output(buffer, out[i], proc)
-                time.sleep(poll_period)
+                time.sleep(poll_latency)
 
             # obtain exit code
             status = process.poll()
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -236,7 +236,11 @@ class WitlessRunner(object):
                 # get a chunk of output
                 pout = [
                     # read whatever is available, should not block
-                    o.read1(-1) if p else None
+                    # From P37 onwards we could say "I don't care how much you can give me"
+                    #o.read1(-1) if p else None
+                    # but instead we say, give me what you have, and if you have nothing,
+                    # make the shortest possible blocking read
+                    o.read1(1) if p else None
                     for o, p in zip(
                         (process.stdout, process.stderr),
                         proc_out)
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -313,7 +313,7 @@ class WitlessRunner(object):
         return out
 
 
-class Runner(WitlessRunner):
+class Runner(object):
     """Provides a wrapper for calling functions and commands.
 
     An object of this class provides a methods that calls shell commands or
@@ -346,8 +346,9 @@ class Runner(WitlessRunner):
              Switch to instruct whether outputs should be logged or not.  If not
              set (default), config 'datalad.log.outputs' would be consulted
         """
-        super(Runner, self).__init__(cwd=cwd, env=env)
 
+        self.cwd = cwd
+        self.env = env
         if protocol is None:
             # TODO: config cmd.protocol = null
             protocol_str = os.environ.get('DATALAD_CMD_PROTOCOL', 'null')
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -98,6 +98,21 @@ def _cleanup_output(stream, std):
         std.close()
 
 
+def kill_output(output):
+    """Helper for WitlessRunner to swallow all output and neither
+    relay it to the parent process's stdout, not provide it as
+    return value of WitlessRunner.run().
+    """
+    return b'', 0
+
+
+def capture_output(output):
+    """Helper for WitlessRunner to capture all output and
+    provide it as return value of WitlessRunner.run().
+    """
+    return output, 0
+
+
 class WitlessRunner(object):
     """Minimal Runner with support for online command output processing
 
@@ -142,10 +157,15 @@ class WitlessRunner(object):
           that it is simply the name of the program, no complex shell
           commands are supported.
         proc_stdout : callable, optional
-          If given, all stdout is passed as byte-string to this callable,
-          in the chunks it was received by polling the processing. The
-          callable may transform it in any way, its output (byte-string)
-          is concatenated and provided as stdout return value.
+          By defaultno stdout is captured, but relayed to the parent
+          process's stdout. If given, all stdout is passed as byte-string
+          to this callable, in the chunks it was received by polling the
+          processing. The callable may transform it in any way, its output
+          (byte-string) is concatenated and provided as stdout return value.
+          The helper functions 'kill_output', and 'capture_output' are
+          provided to either swallow all output (and not relay it to the
+          parent), or to capture all output and provide it as return
+          value.
         proc_stderr : callable, optional
           Like proc_stdout, but for stderr.
         stdin : byte stream, optional
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -194,6 +194,8 @@ class WitlessRunner(object):
           raised which provides the command (cmd), stdout, stderr,
           exit code (status), and a message identifying the failed
           command, as properties.
+        FileNotFoundError
+          When a given executable does not exist.
         """
         proc_out = (proc_stdout, proc_stderr)
         try:
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -155,10 +155,15 @@ class WitlessRunner(object):
           commands are supported.
         proc_stdout : callable, optional
           By default no stdout is captured, but relayed to the parent
-          process's stdout. If given, all stdout is passed as a byte-string
-          to this callable, in the chunks it was received by polling the
-          processing. The callable may transform it in any way, its output
-          (byte-string) is concatenated and provided as stdout return value.
+          process's stdout. If given, all stdout is sequentially passed
+          as a byte-string to this callable, in the chunks it was received
+          by polling the process (see `poll_latency`).
+          The callable may transform it in any way. It must
+          return a byte-string of the transformed output, and an integer
+          with the number of bytes at the end of the original output
+          that were left unprocessed (or 0, if the entire output was
+          considered). The returned byte-strings are concatenated and
+          provided as stdout return value.
           The helper functions 'kill_output' and 'capture_output' are
           provided to either swallow all output (and not relay it to the
           parent) or to capture all output and provide it as the return
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -120,9 +120,9 @@ class WitlessRunner(object):
     functionality. Derived classes should be used for additional
     specializations and convenience features.
     """
-    __slots__ = ['cwd', 'env', '_poll_period']
+    __slots__ = ['cwd', 'env']
 
-    def __init__(self, cwd=None, env=None, poll_period=0.1):
+    def __init__(self, cwd=None, env=None):
         """
         Parameters
         ----------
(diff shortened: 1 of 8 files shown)


mih:
Peculiar test failures:

`datalad.distribution.tests.test_install:test_install_subds_from_another_remote`
`datalad.distribution.tests.test_install:test_install_subds_with_space`
```
[ERROR  ] Failed to clone from any candidate source URL. Encountered errors per each url were:
| - localhost:/tmp/datalad_temp_test_install_subds_with_space9hgjc1_9
  CommandError: command '['git', 'clone', '--progress', 'localhost:/tmp/datalad_temp_test_install_subds_with_space9hgjc1_9', '/tmp/datalad_temp_test_install_subds_with_spaceyjzaxonl']' failed with exitcode 128
Failed to run ['git', 'clone', '--progress', 'localhost:/tmp/datalad_temp_test_install_subds_with_space9hgjc1_9', '/tmp/datalad_temp_test_install_subds_with_spaceyjzaxonl'].
stdout=
stderr=Cloning into '/tmp/datalad_temp_test_install_subds_with_spaceyjzaxonl'...
datalad sshrun: 1: datalad: not found

and the repository exists.
```

Don't even know why it would attempt to run `ssrun`...

mih:
> Don't even know why it would attempt to run `ssrun`... But I also see that locally.

Because we place `ssrun` as `GIT_SSH_COMMAND` in the environment -- and that makes a lot of sense. However, it turns out that `Runner(env=)` was effectively incremental, and inheriting the original environment. For `LeanRunner()` I thought this would limit its flexibility, hence in the `clone()` case we need to inherit explicitly when we aim to patch/ammend it with the SSH helper.

mih:
The log monster travis run now exceeds maximum log length. Not sure how to mitigate without going against the purpose of its existence.

mih:
This is all good now from my POV.

datalad/support/gitrepo.py
```diff
@@ -416,6 +418,208 @@ def wrapped(repo, *args, **kwargs):
     return wrapped
 
 
+class GitProgress(object):
+    """Reduced variant of GitPython's RemoteProgress class
```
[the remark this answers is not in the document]
mih: Sure, will do.

mih:
> Sweet! tested locally -- seems to work.

It also passes the test!

> * besides "testing in action" with clone, there is no single dedicated test to LeanRunner's functionality. And because of that - [there is no coverage for some exception handling in the new code](https://codecov.io/gh/datalad/datalad/compare/5cb174b7079e6c3574782405263a7c6df8bf94a0...c02ad497bfd3bed279889f38718cbb3375ee9297/diff#D2-170) and the case where neither stdout/stderr handling is provided.  So "surprises" might resurface later .

They surface all the time, hence #4087 

> It would be great to at least get the basic case of run without handling + the keyboardinterrupt (could patch mock subprocess.Popen to return an object which would raise KeyboardInterrupt upon `.poll()`)

I will look into porting some of the tests, once I have abandoned the plan to replace `Runner()` entirely with this code and use all of its tests instead .... hold my beer....

> * it calls itself Lean, but no performance comparison against regular Runner... may be it would be Bulkier but speedier (or not) -- it could be super useful to know as to guide further adoption of LeanRunner as a possible replacement for a Runner

I will rename to `LessCodeAndLessFeaturesRunner` for clarity ;-)

datalad/support/gitrepo.py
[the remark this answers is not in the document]
mih: Done.

mih:
I have renamed the class to something that can no longer be mistaken for any ambition.

mih:
A set of basic tests is now included.

datalad/cmd.py
```diff
@@ -98,7 +98,208 @@ def _cleanup_output(stream, std):
...
+        self._poll_period = poll_period
+        self.env = env.copy() if env else None
+        # stringify to support Path instances on PY35
+        self.cwd = str(cwd) if cwd is not None else None
+        if self.cwd and env is not None:
+            # if CWD was provided, we must not make it conflict with
+            # a potential PWD setting
+            self.env['PWD'] = self.cwd
+
+    def run(self, cmd, proc_stdout=None, proc_stderr=None, stdin=None):
```
yarikoptic: note that "original" runner has `log_stdout` and `log_stderr` (for what here `proc_stdout`, `proc_stderr`).

mih: With different semantics though, hence I preferred to use another name.

datalad/cmd.py
```diff
@@ -98,7 +98,208 @@ def _cleanup_output(stream, std):
...
+        proc_stdout : callable, optional
+          By defaultno stdout is captured, but relayed to the parent
+          process's stdout. If given, all stdout is passed as byte-string
+          to this callable, in the chunks it was received by polling the
+          processing. The callable may transform it in any way, its output
+          (byte-string) is concatenated and provided as stdout return value.
+          The helper functions 'kill_output', and 'capture_output' are
+          provided to either swallow all output (and not relay it to the
+          parent), or to capture all output and provide it as return
+          value.
```
yarikoptic: needs to expand on returned value to be a tuple of the `(processed, unprocessed_len)`  with the description of what `unprocessed_len` is (I guess the length of the trailing portion of the input value provided to the callable, which is yet to be processed.  but could be also the trailing portion within returned value which is yet to be processed)

mih: True, will do. Thx.

yarikoptic:
FWIW in https://github.com/yarikoptic/datalad/commit/8f0bb0abfe2990841a15f2fcd22563cabb745994 I have tried to establish some "shim" for this Runner to be used instead of the old Runner in the benchmarks... but I cannot comprehend Python's error on claiming that string is not callable -- may be you guys with a fresh eye could spot a problem (e.g. naive attempt to use a closure on `v` variable reused in the loop...?)

yarikoptic:
probably disregard -- most likely indeed that "closure" + lambda thing... using minimal decorator function seems has resolved it in https://github.com/yarikoptic/datalad/commit/b50dc1ae246cee870259b26d6c9251df28e2c066

kyleam:
Hmm, with vanilla datalad configuration, I'm not seeing any progress bars when I clone.  I've tried with datalad/artwork as well as the larger git.git repo (https://github.com/git/git).  Here's how it looks with datalad/artwork:

clone

datalad/cmd.py
```diff
@@ -98,7 +98,208 @@ def _cleanup_output(stream, std):
...
+          Unicode string with the cumulative standard output and error
+          of the process.
+
+        Raises
+        ------
+        CommandError
+          On execution failure (non-zero exit code) this exception is
+          raised which provides the command (cmd), stdout, stderr,
+          exit code (status), and a message identifying the failed
+          command, as properties.
```
kyleam: It will raise a FileNotFoundError if the program doesn't exist.  Do you think, since we're reworking things, we should wrap that as a CommandError?

datalad/cmd.py
```diff
@@ -98,7 +98,208 @@ def _cleanup_output(stream, std):
...
+        if all(p is None for p in proc_out):
+            proc_out = None
+        try:
+            lgr.log(8, "Start running %r", cmd)
+            process = subprocess.Popen(
+                cmd,
+                # from PY37 onwards
+                #capture_output=proc_output,
+                stdout=subprocess.PIPE if proc_out else None,
+                stderr=subprocess.PIPE if proc_out else None,
```
kyleam: I'm confused about why, if one stream is being captured, that means the other stream should (i.e. about why it's `if proc_out` instead of `if proc_{stdout,stderr}`).  Assuming there's a good reason for this, the interaction should be documented for the proc_stdout/proc_stderr parameters.

datalad/cmd.py
```diff
@@ -98,7 +98,208 @@ def _cleanup_output(stream, std):
...
+                    # of time
+                    pout = process.communicate(
+                        timeout=self._poll_period,
+                    )
+                    # we know the process ended at this point
+                    keep_going = False
+                except subprocess.TimeoutExpired as poll:
+                    # this will always be the full report so far,
+                    # not just an output increment
+                    pout = (poll.stdout, poll.stderr)
```
kyleam: On my end, I don't see output stored on timeouts:

```python
import subprocess as sp

proc = sp.Popen(["sh", "-c", "echo a && sleep 3 && echo b"],
                stdout=sp.PIPE, stderr=sp.PIPE)
try:
    proc.communicate(timeout=1)
except sp.TimeoutExpired as exc:
    print(exc.stdout)

print(proc.communicate())
```

```
None
(b'a\nb\n', b'')
```

This is probably the reason why I don't see progress bars when I clone.  Puzzled.

datalad/cmd.py
```diff
@@ -98,7 +98,208 @@ def _cleanup_output(stream, std):
...
+                            # no index update needed, can never change
+                            continue
+                        # current length of output
+                        len_o = len(o) if o else 0
+                        # anything new in the output?
+                        if len_o - last_idx[i] > 0:
+                            # engage output processed
+                            processed, unprocessed_len = proc(o[last_idx[i]:])
+                            # only keep the processed output
+                            out[i] += processed
```
kyleam: For lots of output, I think it would be better performance-wise to append to a list and join at the end rather than use string concatenation.  (Comment applies above too,)

datalad/cmd.py
```diff
@@ -98,7 +98,208 @@ def _cleanup_output(stream, std):
...
+                            # record point to which we processed the output
+                            # subtract number of bytes reported as unprocessed
+                            last_idx[i] = len_o - unprocessed_len
+
+            # obtain exit code
+            status = process.poll()
+
+            # decode bytes to string
+            out = tuple(o.decode('utf-8') if o else '' for o in out)
```
kyleam: I suppose 'utf-8' is consistent with what we do in `ensure_unicode` (before we fall back to all the guesswork), though I think it'd be better to use `locale.getpreferredencoding(False)`.

datalad/cmd.py
```diff
@@ -98,7 +98,208 @@ def _cleanup_output(stream, std):
...
+                )
+            else:
+                lgr.log(8, "Finished running %r with status %s", cmd, status)
+
+        except CommandError:
+            # do not bother with reacting to "regular" CommandError
+            # exceptions.  Somehow if we also terminate here for them
+            # some processes elsewhere might stall:
+            # see https://github.com/datalad/datalad/pull/3794
+            raise
```
kyleam: I know this is copy-and-paste, but why catch at all here?

datalad/cmd.py
```diff
@@ -98,7 +98,208 @@ def _cleanup_output(stream, std):
...
+                process.terminate()
+            except BaseException as exc2:
+                lgr.warning("Failed to terminate process for %s: %s",
+                            cmd, exc_str(exc2))
+            raise exc_info[1]
+
+        return out
+
+
+class Runner(WitlessRunner):
```
kyleam: Why make Runner() inherit from WitlessRunner()?

datalad/support/gitrepo.py
```diff
@@ -416,6 +418,213 @@ def wrapped(repo, *args, **kwargs):
...
+          All input in its original form, except for lines that
+          were identified as recognized progress reports.
+        """
+        keep_lines = []
+        for line in byts.splitlines(keepends=True):
+            if not self._parse_progress_line(line):
+                keep_lines.append(line)
+        # the zero indicated that no data remained unprocessed at the
+        # end of the input
+        # TODO reevaluate whether it is useful to keep this feature
```
kyleam: Considering you said this PR was ready to go in your view, I guess you've decided it was useful, and this comment should be removed and the docstring(s) updated (which I think Yarik mentioned).

I didn't spot a helper that returned something other than 0.  Why is the feature useful?

datalad/support/gitrepo.py
```diff
@@ -881,24 +1091,17 @@ def clone(cls, url, path, *args, clone_options=None, **kwargs):
...
-                        # non-GitPy based implementation easier. Do conversion
-                        # here
-                        multi_options=to_options(**clone_options) if clone_options else None,
-                        odbt=default_git_odbt,
-                        progress=git_progress
-                    )
-                # Note/TODO: signature for clone from:
-                # (url, to_path, progress=None, env=None, **kwargs)
 
+                # TODO bring back progress reporting
```
kyleam: stale to-do?

datalad/support/gitrepo.py
```diff
@@ -881,24 +1091,17 @@ def clone(cls, url, path, *args, clone_options=None, **kwargs):
...
+                with GitProgress() as progress:
+                    WitlessRunner(env=env).run(
```
kyleam: Hmm, looking at the previous code, I guess this keeps the same `env`.  I'm a bit confused because I would think that it should pass the same environment that `GitRunner.run` does (i.e., one adjusted with `get_git_environ_adjusted`).

mih:
@kyleam Thanks for the review. It'll take me a while to get back to address this. For the moment, I am a bit puzzled by the behavior difference you are seeing:

output

mih:
Real-life testing surfaced another issue:

```
% dl --dbg get -n pics/slides 
Clone attempt:   0%|                                                           | 0.00/2.00 [00:00<?, ? Candidate locations/s]Traceback (most recent call last):
  File "/home/mih/env/datalad3-dev/bin/datalad", line 8, in <module>
    main()
  File "/home/mih/hacking/datalad/git/datalad/cmdline/main.py", line 507, in main
    ret = cmdlineargs.func(cmdlineargs)
  File "/home/mih/hacking/datalad/git/datalad/interface/base.py", line 727, in call_from_parser
    ret = list(ret)
  File "/home/mih/hacking/datalad/git/datalad/interface/utils.py", line 411, in generator_func
    allkwargs):
  File "/home/mih/hacking/datalad/git/datalad/interface/utils.py", line 527, in _process_results
    for res in results:
  File "/home/mih/hacking/datalad/git/datalad/distribution/get.py", line 751, in __call__
    description):
  File "/home/mih/hacking/datalad/git/datalad/distribution/get.py", line 456, in _install_targetpath
    ds, target_path, reckless, refds_path, description=description):
  File "/home/mih/hacking/datalad/git/datalad/distribution/get.py", line 338, in _install_necessary_subdatasets
    description=description):
  File "/home/mih/hacking/datalad/git/datalad/distribution/get.py", line 260, in _install_subds_from_flexible_source
    **kwargs):
  File "/home/mih/hacking/datalad/git/datalad/core/distributed/clone.py", line 413, in clone_dataset
    create=True)
  File "/home/mih/hacking/datalad/git/datalad/support/gitrepo.py", line 1100, in clone
    proc_stderr=progress,
  File "/home/mih/hacking/datalad/git/datalad/cmd.py", line 297, in run
    raise exc_info[1]
  File "/home/mih/hacking/datalad/git/datalad/cmd.py", line 228, in run
    timeout=self._poll_period,
  File "/usr/lib/python3.7/subprocess.py", line 964, in communicate
    stdout, stderr = self._communicate(input, endtime, timeout)
  File "/usr/lib/python3.7/subprocess.py", line 1701, in _communicate
    selector.register(self.stdout, selectors.EVENT_READ)
  File "/usr/lib/python3.7/selectors.py", line 352, in register
    key = super().register(fileobj, events, data)
  File "/usr/lib/python3.7/selectors.py", line 238, in register
    key = SelectorKey(fileobj, self._fileobj_lookup(fileobj), events, data)
  File "/usr/lib/python3.7/selectors.py", line 225, in _fileobj_lookup
    return _fileobj_to_fd(fileobj)
  File "/usr/lib/python3.7/selectors.py", line 40, in _fileobj_to_fd
    "{!r}".format(fileobj)) from None
ValueError: Invalid file object: <_io.BufferedReader name=4>

> /usr/lib/python3.7/selectors.py(40)_fileobj_to_fd()
-> "{!r}".format(fileobj)) from None

```

datalad/cmd.py
kyleam: Looking at subprocess.py, it doesn't seems like the `TimeoutExpired` that would be raised by `.communicate()` would fill in `stdout` and `stderr` until 580d2782f7 (bpo-37424: Avoid a hang in subprocess.run timeout output capture [...], 2019-09-11).  Before that commit, `subprocess.run` looks to be the only spot that populates `TimeoutExpired`s `stdout` and `stderr`.

v3.9.0a1 is the first tag that contains that commit.  That fits with what I'm seeing (I have 3.7.3), but doesn't make sense with the output that you and Yarik are seeing because I doubt you both are running a development Python version.

mih: That is weird. I have Python 3.7.3 and the tests (that also verify the the increments with which the output is processed pass on my machine. However, it makes much more sense now that the same tests failed on the various CIs. But how is my machine special?

datalad/cmd.py
mih: True, will do.

datalad/cmd.py
mih: I cannot come up with a reason (anymore), will change.

datalad/cmd.py
mih: I was torn, and ultimately decided it would be more useful if those are distinguished. It will be a rather common exception, because any invocation with just-a-string is likely to raise it. However, I am open to alternative opinions (and in either case the docs need to be adjusted).

datalad/cmd.py
```diff
@@ -98,7 +98,208 @@ def _cleanup_output(stream, std):
...
+        ----------
+        cwd : path-like, optional
+          If given, commands are executed with this path as PWD,
+          the PWD of the parent process is used otherwise.
+        env : dict, optional
+          Environment to be passed to subprocess.Popen(). If `cwd`
+          was given, 'PWD' in the environment is set to its value.
+          This must be a complete environment definition, no values
+          from the current environment will be inherited.
+        poll_period : float, optional
```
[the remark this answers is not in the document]
mih: True, will do.

datalad/support/gitrepo.py
mih: yes, thanks.

datalad/cmd.py
kyleam: Yeah, I'm on the fence, and just the doc adjustment is okay with me.

datalad/cmd.py
kyleam: > I have Python 3.7.3 and the tests (that also verify the the increments with which the output is processed pass on my machine

All of the test_witless_runner.py tests, and in particular `test_runner_incomplete_capture`, pass on my machine as well.  But those tests are not a reliable indication that `stdout` and `stderr` of `TimeoutExpired` are being populated: the tests pass for me even when I apply the following patch.

```diff
diff --git a/datalad/cmd.py b/datalad/cmd.py
index 74003b6bb..8893d883b 100644
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -230,6 +230,8 @@ def run(self, cmd, proc_stdout=None, proc_stderr=None, stdin=None):
                     # we know the process ended at this point
                     keep_going = False
                 except subprocess.TimeoutExpired as poll:
+                    assert poll.stdout is None
+                    assert poll.stderr is None
                     # this will always be the full report so far,
                     # not just an output increment
                     pout = (poll.stdout, poll.stderr)
```

kyleam: >  I have Python 3.7.3 [...] But how is my machine special?

I'm having no luck coming up with plausible things that could be going on here.  Hoping that maybe the issue is with my machine and that I'm misreading subprocess.el or the cpython git history, I tried in a `debian:latest` docker container with neurodebian sources and Python 3.7.3.  The behavior I see there matches what I report for my local machine: the snippet above shows `None` for `TimeoutExpired.stdout` and `datalad clone https://github.com/datalad/datalad` doesn't show progress (matching the gif I posted).

datalad/cmd.py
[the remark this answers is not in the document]
mih: done

datalad/cmd.py
mih: Because it also handles `self.cwd` and `self.env` for both. If there is a reason to duplicate that, and could be convinced to revert this.

mih:
I have now reimplemented the approach to go fully manual and not use `communicate()`. Let's see...

datalad/support/gitrepo.py
mih: done

datalad/support/gitrepo.py
mih: It is useful whenever some output processor can only meaningfully process a sufficiently large chunk of data. For example, something expects JSON lines output, but for some reason only gets passed 1.5 lines (a full one and another line with an incomplete record -- who knows why, maybe a buffer setting in an external tool). With this feature an output processor can reject some input, and get it passed again later on. This avoids having to implement state-tracking in processors and might keep them simpler.

datalad/cmd.py
mih: Done

datalad/cmd.py
mih: Done

datalad/cmd.py
mih: Done

datalad/cmd.py
kyleam: IMO that's not a reason to inherit from a base class when there are large api conflicts between the parent and child.

mih:
> ValueError: Invalid file object: <_io.BufferedReader name=4>

This crash is no longer happening with the current implementation. I also verified that manual input reaches SSH and a clone succeeds.

datalad/cmd.py
mih: OK, will copy&paste the code.

mih:
This might be working now :)

datalad/support/gitrepo.py
mih: True. I added an appropriate TODO to remove this from GitRepo entirely, and for now converted this one case.

datalad/cmd.py
mih: Reverted all changes to `Runner`

datalad/support/gitrepo.py
mih: Removed in full in my local branch of #4087

datalad/cmd.py
```diff
@@ -98,6 +98,221 @@ def _cleanup_output(stream, std):
...
+            raise
+
+        try:
+            out = [[], []]
+            unprocessed = [None, None]
+            # make sure to run this loop at least once, even if the
+            # process is already dead
+            keep_going = True
+
+            def _handle_output(u, o, p):
```
kyleam: `p` isn't used in the function.  Instead `proc` is taken from the outer scope.

kyleam:
Guh, last night I [posted][0] to the wrong pr:

> I haven't had a chance to review the updates, but I can confirm that the progress bars seem to be functioning nicely on my end now. Thanks.

[0]: https://github.com/datalad/datalad/pull/4087#issuecomment-582738185

datalad/cmd.py
```diff
@@ -98,6 +98,221 @@ def _cleanup_output(stream, std):
...
+                processed, unprocessed_len = proc(u)
+                if processed:
+                    o.append(processed)
+                return u[-(unprocessed_len):] if unprocessed_len else None
+
+            def _read_stream(stream):
+                #if sys.version_info.major == 3 and sys.version_info.minor < 7:
+                nbytes_avail = len(stream.peek())
+                return stream.read(nbytes_avail) if nbytes_avail > 0 else None
+                # o.read1(-1)
```
kyleam: It looks like the negative `size` support added in Python's ccb2c0e310 (Implement optional BufferedReader, BytesIO read1() argument, 2016-10-20) is straightforward:

```diff
diff --git a/Lib/_pyio.py b/Lib/_pyio.py
index d0947f06d5..569527bc37 100644
--- a/Lib/_pyio.py
+++ b/Lib/_pyio.py
@@ -1073,12 +1073,12 @@ def _peek_unlocked(self, n=0):
                 self._read_pos = 0
         return self._read_buf[self._read_pos:]
 
-    def read1(self, size):
+    def read1(self, size=-1):
         """Reads up to size bytes, with at most one read() system call."""
         # Returns up to size bytes.  If at least one byte is buffered, we
         # only return buffered bytes.  Otherwise, we do one raw read.
         if size < 0:
-            raise ValueError("number of bytes to read must be positive")
+            size = self.buffer_size
         if size == 0:
             return b""
         with self._read_lock:
```

So, to get the same behavior before 3.7, can't you use `.buffer_size` here rather than the `peek()` call?

datalad/support/gitrepo.py
```diff
@@ -416,6 +418,213 @@ def wrapped(repo, *args, **kwargs):
...
+
+        Returns
+        -------
+        bytes
+          All input in its original form, except for lines that
+          were identified as recognized progress reports.
+        """
+        keep_lines = []
+        for line in byts.splitlines(keepends=True):
+            if not self._parse_progress_line(line):
```
kyleam: Hmm, so we're potentially dealing with a partial line here.  For example, on one run of clone I see a line `r` (which is returned as non-progress output) followed by a line "emote: Counting objects:  86% [...]" (also returned as non-progress output).  If we're just losing a counting line here and there, that isn't too concerning, I suppose.  Each segment of a partial error/warning/fatal line is unlikely to have a progress match, so both partial lines should end up in `keep_lines`.  `keep_lines` will be joined back together, so downstream code will still be able to detect, say, a "warning: " line.

So I'm not sure how much of a problem this is.  But I think it at least deserves a comment.

datalad/cmd.py
kyleam: This comment still applies to spots in latest revision (5a758c303) that use `.decode('utf-8')`.

datalad/cmd.py
mih: :man_facepalming: 

Fixed.

datalad/cmd.py
mih: Thanks for the reminder, done now.

datalad/cmd.py
mih: The specific comment by @yarikoptic was https://github.com/datalad/datalad/pull/3794#issuecomment-544213473

I don't know any additional details. We can take it out and see if it has any effect.

kyleam: Doh, I misread the code and didn't realize that it was avoiding going into the `except BaseException` clause below it.  Sorry for the noise.

datalad/cmd.py
mih: I have adjusted the comment. I came to realize that `read(-1)` will still make a blocking read (when the buffer happens to be empty), which can lead to process deadlock, when the other output channel is flooded with data in the meantime.

datalad/support/gitrepo.py
kyleam: Thanks for the explanation.   Makes sense and it's fine by me, though it does feel a bit premature to add it when no current callers use it.

datalad/support/gitrepo.py
mih: I added a comment.

datalad/cmd.py
mih: Took it out.

datalad/cmd.py
mih: Added now.

mih:
Thx @kyleam Will merge and continue in #4087

PR #4081, opened 2020-01-24
BF/ENH: get_size_from_key to match specification

bpoldrack:
Suck in implementation from git-annex-ria-remote.
Former implementation couldn't correctly deal with chunks for example.
See https://git-annex.branchable.com/internals/key_format/ for specification.

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -620,9 +620,9 @@ class AnnexRepo(GitRepo, RepoInterface):
         # don't lookup the dict for the same things several times;
         # Is there a faster (and more compact) way of doing this? Note, that
         # locals() can't be updated.
-        s = parsed['s'] if 's' in parsed.keys() else None
-        S = parsed['S'] if 'S' in parsed.keys() else None
-        C = parsed['C'] if 'C' in parsed.keys() else None
+        s = parsed.get('s')
+        S = parsed.get('S')
+        C = parsed.get('C')
 
         if S is None and C is None:
             return s  # also okay if s is None as well -> no size to report
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -594,40 +594,48 @@ class AnnexRepo(GitRepo, RepoInterface):
 
     @staticmethod
     def get_size_from_key(key):
-        """A little helper to obtain size encoded in a key"""
+        """A little helper to obtain size encoded in a key
+
+        Returns
+        -------
+        int or None
+          size of the file or None if either no size is encoded in the key or
+          key was None itself
+
+        Raises
+        ------
+        ValueError
+          if key is considered invalid (at least its size-related part)
+        """
         if not key:
             return None
-
-        # TODO: this method can be more compact. we don't need particularly
-        #       elaborated error distinction
 
         # see: https://git-annex.branchable.com/internals/key_format/
         key_parts = key.split('--')
         key_fields = key_parts[0].split('-')
+        parsed = {field[0]: int(field[1:]) if field[1:].isdigit() else None
+                  for field in key_fields[1:]
+                  if field[0] in "sSC"}
 
-        s = S = C = None
+        # don't lookup the dict for the same things several times;
+        # Is there a faster (and more compact) way of doing this? Note, that
+        # locals() can't be updated.
+        s = parsed['s'] if 's' in parsed.keys() else None
+        S = parsed['S'] if 'S' in parsed.keys() else None
+        C = parsed['C'] if 'C' in parsed.keys() else None
 
-        for field in key_fields[1:]:  # first one has to be backend -> ignore
-            if field.startswith('s'):
-                # size of the annexed file content:
-                s = int(field[1:]) if field[1:].isdigit() else None
-            elif field.startswith('S'):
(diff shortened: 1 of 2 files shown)


datalad/support/annexrepo.py
```diff
@@ -597,12 +597,38 @@ def get_size_from_key(key):
         """A little helper to obtain size encoded in a key"""
         if not key:
             return None
-        try:
-            size_str = key.split('-', 2)[1].lstrip('s')
-        except IndexError:
-            # has no 2nd field in the key
+
+        # TODO: this method can be more compact. we don't need particularly
+        #       elaborated error distinction
```
kyleam: I'm not sure what this is referring to.  I spot only one exception that's raised.

bpoldrack: TBH, I don't remember. Might be a leftover from an earlier version. Will have a closer look when I write the tests for it and delete, if I can't decode that comment again (otherwise clarify or resolve it, of course) ;-)

datalad/support/annexrepo.py
```diff
@@ -597,12 +597,38 @@ def get_size_from_key(key):
...
+        for field in key_fields[1:]:  # first one has to be backend -> ignore
+            if field.startswith('s'):
+                # size of the annexed file content:
+                s = int(field[1:]) if field[1:].isdigit() else None
+            elif field.startswith('S'):
+                # we have a chunk and that's the chunksize:
+                S = int(field[1:]) if field[1:].isdigit() else None
+            elif field.startswith('C'):
+                # we have a chunk, this is it's number:
+                C = int(field[1:]) if field[1:].isdigit() else None
```
yarikoptic: am I missing some difference or all 3 lines are identical and just assign to different variables? if there is no difference, then I would have used something like 
```python
if field and field[0] in 'sSC':
    value = int(field[1:]) if field[1:].isdigit() else None
    if field[0] == 's':
        s = value
    elif ...
```
to address code duplication.  or might  have even got brave for something like

```python
if field and field[0] in 'sSC':
    locals().update({field[0]:  int(field[1:]) if field[1:].isdigit() else None})
```
;-)

datalad/support/annexrepo.py
```diff
@@ -597,12 +597,38 @@ def get_size_from_key(key):
...
+        if s is None:
             return None
-        return int(size_str) if size_str.isdigit() else None
+        elif S is None and C is None:
+            return s
+        elif S and C:
+            if C <= int(s / S):
+                return S
+            else:
+                return s % S
```
yarikoptic: It is a nice little function to unittest ;)

datalad/support/annexrepo.py
bpoldrack: Right! And that's exactly what the mysterious comment was about. The case distinction is needed only if we want to be more specific in "parsing" errors. Since we don't (I think) in can be "compacted" this way. Doing it now.

datalad/support/annexrepo.py
bpoldrack: See below. @yarikoptic kind of found what this was about. ;-)

datalad/support/annexrepo.py
bpoldrack: FTR: `locals()` can't be updated. So, changed to a dict comprehension but updating local variables "manually".

datalad/support/annexrepo.py
bpoldrack: Done.

datalad/support/annexrepo.py
yarikoptic: > FTR: `locals()` can't be updated.

d'oh. indeed. why did I think they could?

<details>
<summary>demo</summary> 

```shell
$> cat /tmp/11.py

def func():
    c = 1
    locals().update({'c': 2})
    print("c=%s" % c)

func()

$> python /tmp/11.py
c=1

```
</details>

datalad/support/annexrepo.py
```diff
@@ -594,40 +594,48 @@ def check_repository_versions(cls):
...
-        elif S is None and C is None:
-            return s
+        parsed = {field[0]: int(field[1:]) if field[1:].isdigit() else None
+                  for field in key_fields[1:]
+                  if field[0] in "sSC"}
+
+        # don't lookup the dict for the same things several times;
+        # Is there a faster (and more compact) way of doing this? Note, that
+        # locals() can't be updated.
+        s = parsed['s'] if 's' in parsed.keys() else None
```
kyleam: `parsed.get('s')`?

---

I know this was in response to a review comment from Yarik, but I must admit that I don't find the newer version to be an improvement in terms of readability.  Perhaps a compromise between this and the previous form would be an internal function that encapsulates `int(field[1:]) if field[1:].isdigit() else None` while keeping the expanded `ifelif` branches.

Anyway, it's a minor thing either way, and I'm fine with how it is.  (I'm just glad you didn't use `locals()` :] )

bpoldrack: > `parsed.get('s')`

WTH? Why didn't I think of it? :-D

bpoldrack: Changed to `get`, but otherwise will leave as is. Kind of agree with the readability aspect, but it's not too bad and TBH I don't really like additional function calls for that.

PR #4089, opened 2020-01-26
BF: Adjust test for clone default return value setup

mih:
Otherwise `assert_status()`runs on a dataset instance.

Unclear why this hasn't been spotted by the tests before...

### DIFF
--- a/datalad/core/distributed/tests/test_clone.py
+++ b/datalad/core/distributed/tests/test_clone.py
@@ -401,7 +401,7 @@ def test_autoenabled_remote_msg(path):
 def test_clone_autoenable_msg_handles_sameas(repo, clone_path):
     ds = Dataset(repo.path)
     with swallow_logs(new_level=logging.INFO) as cml:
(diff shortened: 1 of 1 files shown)


datalad/core/distributed/tests/test_clone.py
```diff
@@ -401,7 +401,7 @@ def test_autoenabled_remote_msg(path):
 def test_clone_autoenable_msg_handles_sameas(repo, clone_path):
     ds = Dataset(repo.path)
     with swallow_logs(new_level=logging.INFO) as cml:
-        res = clone(ds, clone_path)
+        res = clone(ds, clone_path, result_xfm=None, return_type='list')
```
kyleam: Eh, thanks for catching and fixing this.  I'm still scratching my head over why this wasn't failing on Travis.

Locally, standing up a container as the ssh target for `datalad-test`, I do see the failure without your fix.  Dunno.

PR #4091, opened 2020-01-27
DOC: examples for more commands

adswa:
- I've added a few examples to `update`, `uninstall`, `remove`, `download-url`, `diff`, and `run-procedure`. 
- For `drop`, I removed duplicate examples that were still present in the docstring. 
- For `rerun`, I transferred examples from the docstring into the `_examples_` list and added Python translations when possible. 

I deemed one of the examples of `rerun` as not very suitable for the Python API help:
```
Re-execute all previous commands and compare the old and new results:
% # on master branch
% datalad rerun --branch=verify --since=
% # now on verify branch
% datalad diff --revision=master..
% git log --oneline --left-right --cherry-pick master...
```
Therefore, I modified `build_example` to only include an example in the help text of an API if a code snippet exists for this API, and did not translate it to a Python example.

### DIFF
--- a/datalad/core/local/diff.py
+++ b/datalad/core/local/diff.py
@@ -98,18 +98,18 @@ class Diff(Interface):
     )
 
     _examples_ = [
-        dict(text="Show unsaved changes in the current dataset",
-             code_py="diff(dataset='.')",
+        dict(text="Show unsaved changes in a dataset",
+             code_py="diff()",
              code_cmd="datalad diff"),
         dict(text="Compare a previous dataset state identified by shasum "
                   "against current worktree",
-             code_py="diff(dataset='.', fr='SHASUM')",
+             code_py="diff(fr='SHASUM')",
              code_cmd="datalad diff --from <SHASUM>"),
         dict(text="Compare two branches against each other",
              code_py="diff(fr='branch1', to='branch2')",
              code_cmd="datalad diff --from branch1 --to branch2"),
         dict(text="Show unsaved changes in the dataset and potential subdatasets",
-             code_py="diff(recursive=True, dataset='.')",
+             code_py="diff(recursive=True)",
              code_cmd="datalad diff --recursive"),
         dict(text="Show unsaved changes made to a particular file",
              code_py="diff(path='path/to/file')",
--- a/datalad/core/local/diff.py
+++ b/datalad/core/local/diff.py
@@ -98,14 +98,14 @@ class Diff(Interface):
     )
 
     _examples_ = [
-        dict(text="""Show unsaved changes in the current dataset""",
+        dict(text="Show unsaved changes in the current dataset",
              code_py="diff(dataset='.')",
              code_cmd="datalad diff"),
-        dict(text="""Compare a previous dataset state identified by shasum
-                     against current worktree""",
+        dict(text="Compare a previous dataset state identified by shasum "
+                  "against current worktree",
              code_py="diff(dataset='.', fr='SHASUM')",
              code_cmd="datalad diff --from <SHASUM>"),
-        dict(text="""Compare two branches against each other""",
+        dict(text="Compare two branches against each other",
              code_py="diff(fr='branch1', to='branch2')",
              code_cmd="datalad diff --from branch1 --to branch2"),
         dict(text="Show unsaved changes in the dataset and potential subdatasets",
--- a/datalad/core/local/diff.py
+++ b/datalad/core/local/diff.py
@@ -97,6 +97,25 @@ class Diff(Interface):
             constraints=EnsureStr() | EnsureNone()),
     )
 
+    _examples_ = [
+        dict(text="""Show unsaved changes in the current dataset""",
+             code_py="diff(dataset='.')",
+             code_cmd="datalad diff"),
+        dict(text="""Compare a previous dataset state identified by shasum
+                     against current worktree""",
+             code_py="diff(dataset='.', fr='SHASUM')",
+             code_cmd="datalad diff --from <SHASUM>"),
+        dict(text="""Compare two branches against each other""",
+             code_py="diff(fr='branch1', to='branch2')",
+             code_cmd="datalad diff --from branch1 --to branch2"),
+        dict(text="Show unsaved changes in the dataset and potential subdatasets",
+             code_py="diff(recursive=True, dataset='.')",
+             code_cmd="datalad diff --recursive"),
+        dict(text="Show unsaved changes made to a particular file",
+             code_py="diff(path='path/to/file')",
+             code_cmd="datalad diff <path/to/file>"),
+    ]
+
     @staticmethod
     @datasetmethod(name='diff')
     @eval_results
--- a/datalad/distribution/remove.py
+++ b/datalad/distribution/remove.py
@@ -62,12 +62,6 @@ class Remove(Interface):
     file content is dropped. As these checks could lead to slow operation
     (network latencies, etc), they can be disabled.
 
-    Examples:
-
-      Permanently remove a subdataset from a dataset and wipe out the subdataset
-      association too::
-
-        ~/some/dataset$ datalad remove somesubdataset1
     """
     _action = 'remove'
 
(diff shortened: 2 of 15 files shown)


mih:
Test failure is unrelated and familiar.

datalad/core/local/diff.py
```diff
@@ -97,6 +97,25 @@ class Diff(Interface):
             constraints=EnsureStr() | EnsureNone()),
     )
 
+    _examples_ = [
+        dict(text="""Show unsaved changes in the current dataset""",
+             code_py="diff(dataset='.')",
+             code_cmd="datalad diff"),
+        dict(text="""Compare a previous dataset state identified by shasum
```
kyleam: The triple-quote,  multiliine string here and the other instances throughout these commits aren't rendered correctly.  For example, here is what I see in the output of `datalad diff --help`:

```
Compare a previous dataset state identified by shasum
Against current worktree::

   % datalad diff --from <SHASUM>
```

datalad/distribution/remove.py
```diff
@@ -86,6 +80,21 @@ class Remove(Interface):
         if_dirty=if_dirty_opt,
     )
 
+    _examples_ = [
+        dict(text="""Permanently remove a subdataset from a dataset and wipe out
+                     the subdataset association too""",
+             code_py="remove(path='path/to/subds')",
+             code_cmd="datalad remove <path/to/subds>"),
+        dict(text="Permanently remove a dataset and all subdatasets",
+             code_py="remove('path/to/dataset/', recursive=True)",
```
kyleam: I think in other places you use `dataset=` explicitly.  It might be worth doing that consistently.

datalad/distribution/remove.py
```diff
@@ -86,6 +80,21 @@ class Remove(Interface):
...
+             code_cmd="datalad remove -d <path/to/dataset/> --recursive"),
+        dict(text="""Permanently remove a dataset and all subdatasets even if there
+                  are less than the configured minimum amount of (remote) sources
```
kyleam: s/less/fewer/
s/amount/number/

datalad/distribution/uninstall.py
```diff
@@ -145,6 +140,18 @@ class Uninstall(Interface):
         if_dirty=if_dirty_opt,
     )
 
+    _examples_ = [
+        dict(text="Uninstall a subdataset (undo installation)",
+             code_py="uninstall(path='path/to/subds')",
+             code_cmd="datalad uninstall <path/to/subds>"),
+        dict(text="Uninstall a subdataset and all potential further subdatasets",
```
kyleam: Drop "further"?

datalad/distribution/uninstall.py
```diff
@@ -145,6 +140,18 @@ class Uninstall(Interface):
...
+             code_py="uninstall(path='path/to/subds', recursive=True)",
+             code_cmd="datalad uninstall -r <path/to/subds>"),
+        dict(text="Skip checks that assure a minimal number of (remote) sources",
```
kyleam: s/assure/ensure/

datalad/interface/download_url.py
```diff
@@ -91,6 +91,16 @@ class DownloadURL(Interface):
         message=save_message_opt
     )
 
+    _examples_ = [
+        dict(text="Download files from an http and S3 URL",
+             code_py="download_url(urls=['http://example.com/file.dat', 's3://bucket/file2.dat'])",
+             code_cmd="datalad download-url http://example.com/file.dat s3://bucket/file2.dat"),
+        dict(text="Downlaod a file to a path and provide a commit message",
```
kyleam: s/Downlaod/Download/

datalad/core/local/diff.py
mih: This smells like a bug in the general docstring formatting code, I didn't spot anything specific to examples....

kyleam: The upper-casing is certainly a bug, and perhaps not stripping the newline between "shasum" and "against" is.  While it's outside this PR's scope to fix that, I think it should avoid a form that introduces lots of these misrendered cases.  Using single quotes, for example, would fix it.

datalad/interface/rerun.py
```diff
@@ -178,6 +154,28 @@ class Rerun(Interface):
...
+             code_py="rerun(onto='', since='HEAD~5')",
+             code_cmd="datalad rerun --onto= --since=HEAD~5"),
+        dict(text="""Re-execute all previous commands and compare the old and
+                     new results""",
+             code_py=" ",
+             code_cmd="""# on master branch
+                      % datalad rerun --branch=verify --since=
+                      % # now on verify branch
+                      % datalad diff --revision=master..
+                      % git log --oneline --left-right --cherry-pick master..."""),
```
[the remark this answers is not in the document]
adswa: Thanks for catching that!

datalad/core/local/diff.py
adswa: Thanks, I will use single quotes.

datalad/distribution/uninstall.py
adswa: :woman_facepalming: thanks. I never had a problem with assure vs ensure until #3761

datalad/interface/rerun.py
```diff
@@ -178,6 +154,28 @@ class Rerun(Interface):
...
+        dict(text="Re-execute any commands in the last five commits",
+             code_py="rerun(since='HEAD~5')",
+             code_cmd="datalad rerun --since=HEAD~5"),
+        dict(text="""Do the same as above, but re-execute the commands on top of
+                     HEAD~5 in a detached state""",
+             code_py="rerun(onto='', since='HEAD~5')",
+             code_cmd="datalad rerun --onto= --since=HEAD~5"),
+        dict(text="""Re-execute all previous commands and compare the old and
+                     new results""",
+             code_py=" ",
```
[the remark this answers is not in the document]
adswa: thanks!

datalad/distribution/remove.py
adswa: An hour ago I thought "oh yeah, why not do for all?" and now I'm buried under notes which command resolves paths as relative to the dataset argument and which not. :smile:

I'm considering writing all of the examples as being dataset methods to avoid this confusion. Do you have thoughts on this?

mih: My take would be: dont worry about the old stuff. If it behaves strange, let fix it first and then document it more beautifully.

adswa: Okay, unless someone else has a different opinion, I won't use the `dataset=` argument for remove at this point. All "core" commands either have it specified or have examples with dataset methods, and we can revisit other commands at a later point when there is more consistency in their behavior.

kyleam: > An hour ago I thought "oh yeah, why not do for all?" and now I'm buried under notes which command resolves paths as relative to the dataset argument and which doesn't.

Those notes would be valuable because we should fix these.

---

Sorry, my comment led you in a different direction than I was trying to go.  I was just wondering whether you want to use `dataset='path/to/dataset/'`, thinking 'path/to/dataset/' was being passed a positional argument for `dataset`.  Looking at it again, I was wrong in taking 'path/to/dataset/' as the `dataset` argument.  It is `path`.  But then doesn't that mean the code_py and code_cmd counterparts don't match??

```python
             code_py="remove('path/to/dataset/', recursive=True)",
             code_cmd="datalad remove -d <path/to/dataset/> --recursive"),
```

Your Python example is passing 'path/to/dataset/' as `path`, while the command-line example is passing it as `dataset`.

Ignoring that issue, my original question would become "Should you put "path='path/to/dataset/'" for consistency with, say, the example right above it?".  And perhaps in general it's worth taking the stance that our examples should use keyword arguments for clarity.

kyleam: > Okay, unless someone else has a different opinion, I won't use the `dataset=` argument for remove at this point. All "core" commands either have it specified or have examples with dataset methods, and we can revisit other commands at a later point when there is more consistency in their behavior.

At least for the examples you're adding in this PR, I'd recommend consistently using unbound command calls for the Python examples, specifying the `dataset=` argument (as a string) when the command-line variant has  `-d/--dataset` specified.  Even for commands that don't follow the new path semantics, this should get you matching behavior between the command-line and Python variants.

datalad/core/local/diff.py
kyleam: > The upper-casing is certainly a bug, 

That's fixed with gh-4094.

> and perhaps not stripping the newline between "shasum" and "against" is

@adswa  Given the `text` field of the example is free form, do you think we should do something like this?

```diff
diff --git a/datalad/interface/base.py b/datalad/interface/base.py
index 15fe8850c..9f6798785 100644
--- a/datalad/interface/base.py
+++ b/datalad/interface/base.py
@@ -402,7 +402,7 @@ def build_example(example, api='python'):
     if code_field not in example:
         # only show an example if it exist for the API
         return ''
-    description = dedent_docstring(example.get('text'))
+    description = textwrap.fill(example.get('text'))
     # this indent the code snippet to get it properly rendered as code
     # we are not using textwrap.fill(), because it would not acknowledge
     # any meaningful structure/formatting of code snippets. Instead, we
```

datalad/distribution/remove.py
adswa: Thanks a lot for clarifying, I misunderstood. I agree to using keyword arguments for clarity and matching arguments in Python and CMD examples. I'll push an update to the examples in this PR.

As a more general note to think about, some examples use dataset methods (e.g., run and save - if I recall correctly because these were preexisting examples in the docstrings), and it may be useful to unify whether examples show functions or methods at some point.

> Those notes would be valuable because we should fix these.

Agreed, I will go the remaining commands and file and issue with my observations.

datalad/core/local/diff.py
adswa: It makes sense to me, yes.

PR #4093, opened 2020-01-27
OPT: is_with_annex - sample if there is .git/refs/heads/git-annex to avoid calling git

yarikoptic:
Partially (only for git-annex repos) provides alternative solution to #4077, at least for already the repos with `git-annex` branch initialized.

I also mentally excercized an idea to establish some helper, which in this case would be

`@memoize_based_on_mtimes(paths=[`.git/refs`], recursive=True, dirs=True, files=False)` 

so it would quickly (I hope) check if any mtime was modified on any subdir under `.git/refs`, so if there is no changes from previous time, it means no new/updated references, and we could take previous value.  But yet to see if it would provide any benefit.

### DIFF
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -1079,6 +1079,11 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
     def is_with_annex(self):
         """Report if GitRepo (assumed) has (remotes with) a git-annex branch
         """
+        # OPT: To avoid "lengthy" invocation of git, first sample ref for
+        # `git-annex` branch
+        if op.exists(self.dot_git / "refs" / "heads" / "git-annex"):
+            return True
+        # It still could be present in the remote ones
         return any(
             b['refname:strip=2'] == 'git-annex' or b['refname:strip=2'].endswith('/git-annex')
             for b in self.for_each_ref_(fields='refname:strip=2', pattern=['refs/heads', 'refs/remotes'])
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -1081,7 +1081,7 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
         """
         # OPT: To avoid "lengthy" invocation of git, first sample ref for
         # `git-annex` branch
-        if op.exists(self.dot_git / "refs" / "heads" / "git-annex"):
+        if (self.dot_git / "refs" / "heads" / "git-annex").exists():
             return True
         # It still could be present in the remote ones
         return any(
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -1081,6 +1081,8 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
         """
         # OPT: To avoid "lengthy" invocation of git, first sample ref for
         # `git-annex` branch
+        if (self.pathobj / ".noannex").exists():
+            return False
         if (self.dot_git / "refs" / "heads" / "git-annex").exists():
             return True
         # It still could be present in the remote ones


datalad/support/gitrepo.py
```diff
@@ -1079,6 +1079,11 @@ def config(self):
     def is_with_annex(self):
         """Report if GitRepo (assumed) has (remotes with) a git-annex branch
         """
+        # OPT: To avoid "lengthy" invocation of git, first sample ref for
+        # `git-annex` branch
+        if op.exists(self.dot_git / "refs" / "heads" / "git-annex"):
```
yarikoptic: I expect that `dot_git` would be "correct" for bare repos ;)

kyleam: I don't know if the wink means you're being sarcastic, but just in case: it wouldn't work for bare repos (though it wouldn't do any harm either):

https://github.com/datalad/datalad/blob/1aebd29c563828e31121b2dc5d6b6da0110fb8b1/datalad/support/gitrepo.py#L628

It also wouldn't catch a packed git-annex ref (e.g., after a `git gc` run), though again that doesn't do any harm since we'd just fall back to the for-each.

Looks fine to me.

datalad/support/gitrepo.py
```diff
@@ -1081,7 +1081,7 @@ def is_with_annex(self):
         """
         # OPT: To avoid "lengthy" invocation of git, first sample ref for
         # `git-annex` branch
-        if op.exists(self.dot_git / "refs" / "heads" / "git-annex"):
+        if (self.dot_git / "refs" / "heads" / "git-annex").exists():
```
yarikoptic: d'oh, right, thanks!

datalad/support/gitrepo.py
yarikoptic: > I don't know if the wink means you're being sarcastic, but just in case: it wouldn't work for bare repos (though it wouldn't do any harm either):
> 
> https://github.com/datalad/datalad/blob/1aebd29c563828e31121b2dc5d6b6da0110fb8b1/datalad/support/gitrepo.py#L628

I assumed that for the bare repos cases, the `.dot_git` value would be adjusted... to just point to the `.pathobj`. Was that too naive to assume?

packed refs: forgot about them.  Indeed an ideal solution here should also check `$GIT_DIR/packed-refs`.  Do you think it is worthwhile adding to this PR?

yarikoptic:
hm according to [CI's benchmarking](https://github.com/datalad/datalad/pull/4093/checks?check_run_id=411739952#step:7:221) - no improvement :-/  on a quick check -- this "solution would be a "red herring".  `GitRepo.is_with_annex`  seems to execute (in case of `status` or `run` at least) only for git repositories. If it is known to be annex already, it is not even ran, so this condition is unlikely to trigger but in some limited set of cases, so probably not worth even going this route...

While testing it I realized that we do populate `.noannex` when `create(no_annex=True)`, so I pushed ecd8fe8fbee49e4ba8a52a5a5c74334156863b4f.  that one is of positive effect on create, but I am still not sure if any of this OPTs are worth it.

datalad/support/gitrepo.py
kyleam: > I assumed that for the bare repos cases, the `.dot_git` value would be adjusted... to just point to the `.pathobj`. Was that too naive to assume?

I don't spot the adjustment anywhere, but perhaps I'm missing it.

> packed refs: forgot about them. Indeed an ideal solution here should also check `$GIT_DIR/packed-refs`. Do you think it is worthwhile adding to this PR?

No, I don't think we should try to optimize this in this PR (or any other, really) by inspecting packed-refs.  I was just pointing out a case when this optimization wouldn't be in effect.

kyleam:
> [...] If it is known to be annex already, it is not even ran, so this condition is unlikely to trigger but in some limited set of cases, so probably not worth even going this route...

Thanks for the deeper analysis.

> While testing it I realized that we do populate `.noannex` when `create(no_annex=True)`, so I pushed [ecd8fe8](https://github.com/datalad/datalad/commit/ecd8fe8fbee49e4ba8a52a5a5c74334156863b4f). that one is of positive effect on create, but I am still not sure if any of this OPTs are worth it.

I lean towards not including the checks, but they're simple enough that it's okay with me if you feel strongly about adding them.

mih:
> check if any mtime was modified on any subdir under `.git/refs`, so if there is no changes from previous time, it means no new/updated references, and we could take previous value. But yet to see if it would provide any benefit.

We would have to be extra clever with such an approach. There are monstrous pitfalls re timestamp precision across filesystems, and Git is doing quite a bit of work (especially on Windows, AFAIK). I am not confident that bypassing Git is a good idea. Not benefiting from an optimization under some circumstances would be a minor thing. But failing to act, because of optimization would obviously be undesirable.

yarikoptic:
I will leave it to others to decide either this OPT is worth it.  It is just a band-aid, probably through the years of use but all datalad users would not make a world a greener place ;)  So, feel free to close with or without merge.

kyleam:
> I will leave it to others to decide either this OPT is worth it. [...]

All right, I'll go with my opinion expressed above then.  Of course, if others chime in to support merging, we can reopen.

Issue #3528, opened 2019-07-15
Expand `clone --reckless` with modes for a symlinked annex or a plain worktree


mih:
The purpose is to streamline the re-use of resources that already exist locally, and to minimize impact on file system performance or quotas.

Background: https://github.com/datalad/datalad/issues/3521

IMHO we should

- [ ] implement the scenario of a `.git/annex` directory that is symlinked to a source repository as an immediate solution
- [ ] re-evaluate our own support for worktrees, and support this scenario (too) if and when we can handle it (see https://github.com/datalad/datalad/issues/753)

kyleam:
> IMHO we should
> 
>     * [ ]  implement the scenario of a `.git/annex` directory that is symlinked to a source repository as an immediate solution

Makes sense.

>     * [ ]  re-evaluate our own support for worktrees, and support this scenario (too) if and when we can handle it (see #753)

I agree it'd be nice to support worktrees (ideally just by removing gitpython use rather than adding a kludge), but this comment is about expanding `install --reckless` with something for worktrees, right?  That seems like a weird fit to me, given that `git worktree ...` is designd to be called from within the main, _local_ repo.

mih:
> install --reckless with something for worktrees, right? That seems like a weird fit to me

I was thinking along the lines of first supporting worktrees, and then using that feature in `install` to represent a subdataset as a worktree of a local repository that is located outside the dataset directory tree. But this is not fully cooked in my brain yet.

mih:
FTR: https://github.com/datalad/git-annex-ria-remote/pull/31 contains a dataset procedure that covers most of what is asked for in here.

kyleam:
Now that gh-4172 pruned the last remnants of gitpython (thanks @mih), we have the ability to finally see where we stand with worktrees.  (That is, we no longer get to point to gh-753 and use gitpython as an excuse for not being compatible with worktrees. :]) I've done a bit of experimenting today.  I'm not at the point where I think I can do a proper write-up, but a few high-level comments:

  * For plain git repo with no submodules: Basic things seem to work.  I think it'd only take a bit of targeted poking to get a failure though, as worktrees of course use a gitdir file and spots in datalad hard code the ".git/" (gh-3529).

  * For annex repos with no submodules: To be compatible with worktrees, git-annex replaces the gitdir file with a `.git -> <path to main wtree>/.git/worktrees/wtree` link and 
[...]

Issue #3521, opened 2019-07-10
Feature-request: IO-limiting mode for `get` operations


mih:
## Scenario

There is a dataset with a few (<30) files that are each huge (each is several dozen GB of https://www.awaresystems.be/imaging/tiff/bigtiff.html). The "normal" procedure to capture the (comparatively) small results of an analysis of these data is to

- create a new dataset for the results
- install the bigtiff dataset as a subdataset
- use `datalad run` with appropriate `--input` args to execute the analysis

## Problem

This procedure creates a substantial IO bottleneck. Even if all IO is happening locally on a machine (shared file system), about 1 *TB* of data needs to be copied -- which takes time, a sizable chunk of additional storage, and causes noticeable churn on the filesystem. In this scenario, this copy has a lifetime of only a few minutes (the run time of the analysis). Afterwards, it will be thrown away.

## Potential solutions

We already have `install --reckless` that was motivated by a similar desire (fast read-only access, previous discussion https://github.com/datalad/datalad/issues/1):

```
  --reckless      Set up the dataset to be able to obtain content in the
                        cheapest/fastest possible way, even if this poses a
                        potential risk to data integrity (e.g. hardlink files
                        from a local clone of the dataset). Use with care, and
                        limit to "read-only" use cases. With this flag the
                        installed dataset will be marked as untrusted.
                        [Default: False]
```

which ultimately sets `annex.hardlink=true` and `untrust`s that clone. This can help if the data is on the same filesystem, but is helpless if not. In the concrete case that motivates this issue, we are on HPC infrastructure, and the input data lives on a dedicated filesystem (GPFS) that is providing relatively static data to an army of compute nodes.

What about populating `.git/annex/objects` of the temporary input dataset clone with **symlinks** to the keys in `.git/annex/objects` of the source dataset on the other file system? This would constrain storage impact to the number of inodes, but have little to no cost re size and bandwidth.
Or even have `.git/annex/objects` itself be a straight-up symlink to the entire annex tree of the source dataset? In that case, we could even save a (possibly substantial) number of required inodes on the FS (not much of an issue in this concrete scenario, but a general concern with many HPC setups). Of course, this is only meaningful for read-only operations (and writes could be prevented by appropriate permission constraints on the source dataset directories).

@bpoldrack will shortly report on an exploration on how git-annex reacts to such a setup. 

@joeyh it would be great to get your thoughts on this approach for a cheap, temporary, read-only access.

bpoldrack:
[...]

PR #4099, opened 2020-01-28
NF: clone: --reckless='ephemeral'

bpoldrack:
Introduce reckless mode 'ephemeral' for throw-away clones, symlinking their .git/annex to origin's .git/annex.
In addition declare 'here' dead to not propagate availability info from it when publishing captured changes back.

### DIFF
--- a/datalad/core/distributed/clone.py
+++ b/datalad/core/distributed/clone.py
@@ -64,6 +64,9 @@ from datalad.distribution.dataset import (
 from datalad.distribution.utils import (
     _get_flexible_source_candidates,
 )
+from datalad.tests.utils import (
+    has_symlink_capability
+)
 
 __docformat__ = 'restructuredtext'
 
@@ -301,7 +304,7 @@ def clone_dataset(
       Any suitable clone source specifications (paths, URLs)
     destds : Dataset
       Dataset instance for the clone destination
-    reckless : {None, 'auto'}, optional
+    reckless : {None, 'auto', 'ephemeral'}, optional
       Mode switch to put cloned dataset into throw-away configurations, i.e.
       sacrifice data safety for performance or resource footprint.
     description : str, optional
@@ -550,7 +553,7 @@ def postclonecfg_annexdataset(ds, reckless, description=None):
         return
 
     # init annex when traces of a remote annex can be detected
-    if reckless:
+    if reckless == 'auto':
         lgr.debug(
             "Instruct annex to hardlink content in %s from local "
             "sources, if possible (reckless)", ds.path)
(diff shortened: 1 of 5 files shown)


bpoldrack:
Currently not able to grasp what's going on in the windows test. Also it fails differently on our windows testing box. Any thoughts, @mih?

datalad/core/distributed/clone.py
```diff
@@ -571,16 +586,64 @@ def postclonecfg_annexdataset(ds, reckless, description=None):
...
     repo = AnnexRepo(ds.path, init=True)
     if not repo.is_initialized() or description:
         repo._init(description=description)
-    if reckless:
+    if reckless == 'auto':
         repo._run_annex_command('untrust', annex_options=['here'])
 
+    elif reckless == 'ephemeral':
+        # with ephemeral we declare 'here' as 'dead' right away, whenever
+        # we symlink origins annex. Because we want annex to copy to
```
kyleam: s/origins/origin's/

datalad/core/distributed/clone.py
```diff
@@ -571,16 +586,64 @@ def postclonecfg_annexdataset(ds, reckless, description=None):
...
+        # the ria remote to get availability info correct for an eventual
+        # git-push into the store
```
kyleam: I'm having a hard time parsing/understanding from "Because [...]" to here.  If it is meant to convey what is said in the commit message (re: not propagating back), I think that phrasing is clearer.

datalad/core/distributed/clone.py
```diff
@@ -571,16 +586,64 @@ def postclonecfg_annexdataset(ds, reckless, description=None):
...
+        # % git annex find . --not --in here
+        # % git annex find . --in here
+        # d1
+
+        # we don't want annex copy-to origin
+        ds.config.set(
+            'remote.origin.annex-ignore', 'true',
+            where='local')
+
+        ds.repo._run_annex_command('dead', annex_options=['here'])
```
kyleam: `ds.repo.set_remote_dead('here')`?

datalad/core/distributed/clone.py
```diff
@@ -571,16 +586,64 @@ def postclonecfg_annexdataset(ds, reckless, description=None):
...
+
+        if check_symlink_capability(ds.repo.dot_git / 'dl_link_test',
+                                    ds.repo.dot_git / 'dl_target_test'):
+            # symlink the annex to avoid needless copies in an emphemeral clone
+            annex_dir = ds.repo.dot_git / 'annex'
+            origin_annex_url = ds.config.get("remote.origin.url", None)
+            if origin_annex_url:
+                try:
+                    # deal with file:// scheme URLs as well as plain paths
+                    # if origin isn't local, we have nothing to do
```
kyleam: comment nit-pick: I'm frequently tripped up by comments in this project that drop a period and treat a new line as a sentence separator.  As a reader, I would prefer to not have to guess from the context.  Like here, I'd guess that it should be "paths. If", but I'm not sure.

datalad/core/distributed/clone.py
```diff
@@ -571,16 +586,64 @@ def postclonecfg_annexdataset(ds, reckless, description=None):
...
+                    origin_git_path = Path(PathRI(origin_annex_url).localpath)
+                    if origin_git_path.name != '.git':
+                        origin_git_path /= '.git'
+                except Exception:
```
kyleam: Can a more specific exception be caught here?

datalad/core/distributed/clone.py
```diff
@@ -571,16 +586,64 @@ def postclonecfg_annexdataset(ds, reckless, description=None):
...
+                    # TODO: What level? + note, that annex-dead is independ
```
kyleam: s/independ/independent/?  I'm not sure what the note part of the to-do is about.

datalad/core/distributed/clone.py
```diff
@@ -571,16 +586,64 @@ def postclonecfg_annexdataset(ds, reckless, description=None):
...
+                    pass
+            if origin_git_path:
+                # TODO make sure that we do not delete any unique data
+                rmtree(str(annex_dir)) \
+                    if not annex_dir.is_symlink() else annex_dir.unlink()
+                annex_dir.symlink_to(origin_git_path / 'annex',
+                                     target_is_directory=True)
+        else:
+            # TODO: What level? + note, that annex-dead is independ
+            lgr.warning("Unable to create symlinks.")
```
kyleam: I think something about "reckless=ephemeral" should be added to this message.

datalad/core/distributed/tests/test_clone.py
```diff
@@ -763,3 +765,72 @@ def test_ria_http_storedataladorg(path):
     ok_(ds.is_installed())
     eq_(ds.id, datalad_store_testds_id)
 
+
+@skip_if_on_windows  # see gh-
```
kyleam: s/gh-/&4131/

bpoldrack:
Thanks, @kyleam - will address those comments.

>  Playing around with this a bit, I noticed that I could trigger a failure if I did anything in the original repository (the one the symlink points to). Have you hit into any issues on your end?

I can reproduce that, yes. However, I don't think we can do much about that, other than maybe enhancing the doc for it (and possibly complain to Joey). Annex complains about the object for the file you saved in origin, since in `clone1` it doesn't know about it. IMO this shouldn't necessarily lead to failure to commit on annex' end, since an unknown object in the object tree doesn't really hurt (particularly not the new object that we tried to commit).

So, in my (current) view, we should add to the doc, that a symlinked annex comes with some potential issues (it's called `reckless` after all). A `save` should prob. be prepended by an `update` ( `git fetch` actually is sufficient), which solves the case you set up. The test case would still fail on publishing back, since it's then not fast-forward, of course.

The things we possibly could  do, I think, are:

1. catch that error on `save`, detect whether `.git/annex` is symlinked and suggest to update first via error message OR

2. save `ephemeral` mode in repo (we prob. should anyway, since parameters to clone should be stored as the default for subdatasets - just noticed, that I forgot to make sure that's true for `reckless` as well) and based on that config have an automatic fetch prior to a save.

What do you think @mih ?

mih:
@kyleam Thanks for the review!

@bpoldrack your reasoning seems sound!

kyleam:
@bpoldrack  Thanks for the pushed updates.  Those look good to me.

kyleam:
For posterity:
  * original discussion around this idea: gh-3521
  * related issue that is partially fulfilled by this pr: gh-3528

bpoldrack:
Changed storing the `reckless` mode accordingly. And added to the doc.
From my point of view this is it for this PR. If you guys agree, @kyleam , @mih , I would make an issue regarding dealing with detecting the potential issue from within `save` and have a separate PR for that.

kyleam:
Sounds good to me.  Thanks @bpoldrack.

datalad/core/distributed/clone.py
```diff
@@ -652,6 +645,16 @@ def postclonecfg_annexdataset(ds, reckless, description=None):
...
 
+    if reckless:
+        # we successfully dealt with reckless here.
+        # store the reckless setting in the dataset to make it
+        # known to later clones of subdatasets via get()
+        ds.config.set(
+            'datalad.clone.reckless', reckless,
+            where='local',
+            # delay reload until all config IO is done
+            reload=False)
```
kyleam: This lack of a reload is what's behind the new test failures.  In the deleted call above, the very next line has `reload=True`.

bpoldrack: Thx!

Issue #3916, opened 2019-12-11
"datalad status" no terminal output for clean datasets


dorianps:
#### What is the problem?
`datalad status` shows no response on dataset without changes.

#### What steps will reproduce the problem?
`datalad save ; datalad status`

#### What version of DataLad are you using (run `datalad --version`)? On what operating system (consider running `datalad wtf`)?
`datalad 0.12.0rc6`

#### Is there anything else that would be useful to know in this context?
This is no bug. If there is nothing to stage or commit, `datalad status` does not return anything, no message at all. This is a bit confusing, it would make sense to actually return something on the terminal, maybe "nothing to save" etc, similar to what `git status` returns when there is nothing to commit.

#### Have you had any success using DataLad before? (to assess your expertise/prior luck.  We would welcome your testimonial additions to https://github.com/datalad/datalad/wiki/Testimonials as well)
N/A

mih:
I am not convinced that rendering a message is needed or helpful. Being silent is a common behavior IMHO.

```
mih@meiner ~ % mkdir empty
mih@meiner ~ % ls empty
mih@meiner ~ %
```
It doesn't say "there is nothing here" ;-)

There are also other reasons for running `status`, besides an intention to `save` something. Also,  it is not entirely correct that there is no "return value". In fact status reports verbosely. You can enable that via something like: `datalad -f json_pp status`. It is just the behavior of the default result renderer to be silent, when everything is clean.

I hope that sheds some light on the situation.

dorianps:
Mine was a suggestion, of course. But from a development perspective I can hardly think of any tool that implies the work is done by not printing anything on screen. In addition, the `ls` command return the prompt instantly if there are no files, while `datalad status` can sit for 10 minutes and then return nothing. The first is a linux command with a single well defined purpose, whereas `datalad status` is a full algorithm of checks on git, git-annex, symlinks, etc. It just looks weird that the user request to know the status of a dataset gives no response at all, even more so since the underlying tools produce a response for the same `status` call.

I checked `datalad -f json_pp status` and, honestly, I don't think is of any use for routine use. It just prints a huge list of information that is not informative for humans. One would need to check by eye if the two commits match. Even trying to filter the output with ` datalad -f json_pp status | grep status" prints a huge list that brings no information.

Just my view on interfacing tools with a user. Maybe it would be helpful to add a flag on the level of information, i.e., say something at the end, or show some verbosity on what steps are being performed during a, say 10 minute long, run.

[...]

PR #4106, opened 2020-01-30
BF: Don't be silent in default renderer when everything is clean

mih:
Follow up on a report by @dorianps considering no output confusing. `status`
now follows `git status` in wording.

Fixes gh-3916

```
(datalad3-dev) mih@meiner ~/hacking/datalad/git (git)-[bf-3916] % datalad status
```

### DIFF
--- a/datalad/core/local/status.py
+++ b/datalad/core/local/status.py
@@ -473,3 +473,6 @@ class Status(Interface):
                         len(annexed),
                         single_or_plural('file', 'files', len(annexed)),
                         total_size))
+        if all(r.get('state', None) == 'clean' for r in results):
+            from datalad.ui import ui
+            ui.message("nothing to save, working tree clean")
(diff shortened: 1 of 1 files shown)


mih:
Appveyor failure is failure to upload coverage.

datalad/core/local/status.py
```diff
@@ -473,3 +473,6 @@ def custom_result_summary_renderer(results):  # pragma: more cover
                         len(annexed),
                         single_or_plural('file', 'files', len(annexed)),
                         total_size))
+        if all(r.get('state', None) == 'clean' for r in results):
```
kyleam: Is it valid to assume that every result coming through is `action='status'`?  It looks like code just upstream does not make this assumption.

mih: Good Q. Will investigate and add a filter, if needed. Thx!

mih: I have added the check. Even if it is not necessary now, we run the risk of silently breaking it in the future, if it is no ensured.

PR #4108, opened 2020-01-30
Update screen cast for 0.12

mih:
Old version was made with 0.8

### DIFF
--- a/docs/casts/seamless_nested_repos.sh
+++ b/docs/casts/seamless_nested_repos.sh
@@ -1,43 +1,43 @@
-say "DataLad provides seamless management of nested Git repositories..."
+say "DataLad makes a tree of nested Git repositories feel like a big monorepo..."
 
-say "Let's create a dataset"
+say "Let's create a root dataset"
 run "datalad create demo"
 run "cd demo"
-say "A DataLad dataset is just a Git repo with some initial configuration"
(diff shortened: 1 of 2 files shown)


mih:
Neuroimaging extension failure is unrelated (like change in pandas).

docs/casts/seamless_nested_repos.sh
```diff
@@ -1,43 +1,43 @@
-say "DataLad provides seamless management of nested Git repositories..."
+say "DataLad makes a tree of nested Git repositories feel like a big monorepo..."
```
kyleam: Ha, first time I've seen that tagline.  I'm not sure what I think of it, though my guess is that monorepo is not really a term that'd carry much weight with most of DataLad's target audience.

mih: I used it, because it is used on https://www.datalad.org/for/git-users

kyleam: > I used it, because it is used on https://www.datalad.org/for/git-users

The top of `content/asciicast/seamless_nested_repos.sh` in the `datalad/datalad.org` repo tells me that that file is generated from `docs/casts/seamless_nested_repos.sh`, which I think refers to the file from this PR.  So doesn't your above quote reduce to "I used it here, because I used it here"?

Anyway, my comment was really just a comment, not an objection to using that phrasing.

PR #4112, opened 2020-01-31
DOC: Cast to compare git vs datalad for a common submodule workflow

mih:
It would be instrumental to get feedback on whether I am missing a more elegant approach for Git. Thx!

Recorded asciicast at https://raw.githubusercontent.com/datalad/talk-fosdem-2020/master/asciicast/datalad_convenience_vs_git.json

### DIFF
--- a/docs/casts/datalad_convenience_vs_git.sh
+++ b/docs/casts/datalad_convenience_vs_git.sh
@@ -1,0 +1,109 @@
+say "Fast forward through some prior work..."
+
+say "Let's assume a student developed an algorithm, using Git."
+run "mkdir code; git -C code init; touch code/work; git -C code add work ; git -C code commit -m 'MSc thesis by student A done'"
+
+say "An another student collected some data, tracked with Git-annex."
+run "mkdir data; git -C data init; git -C data annex init; echo 'DATA!' > data/work; git -C data annex add work ; git -C data commit -m 'Data collection by student B done'"
+
+say "A postdoc performed an analysis with the new algorithm on that data, results tracked with Git-annex."
+
+run "mkdir analysis; git -C analysis init && git -C analysis annex init"
+
+say "Git submodules are perfect for versioned tracking of dependencies, code or data."
+run "git -C analysis submodule add ../code"
+run "git -C analysis submodule add ../data"
+run "git -C analysis/data annex init"
+run "git -C analysis commit -m 'Add dependencies'"
+
+run "touch analysis/work; git -C analysis add work && git -C analysis commit -m 'Analysis by postdoc done'"
+
+say "This happened in the past... Now the students left and the postdocs is on vacation. Time for the PI to write up the paper..."
+
+say "The paper is nothing but a new project that depends on the analysis of the study it will describe."
+run "mkdir paper"
+run "cd paper"
+run "git init"
+run "git submodule add ../analysis study"
+say "Need to remember what the postdoc said: If a repo uses git-annex, need to init it. Does it?"
+run "git -C study branch -a | grep git-annex"
+say "Ok, seems to be"
+run "git -C study annex init"
+run "git commit -m 'Add analysis'"
+
+say "Let Git assemble the entire working tree"
+run "git submodule update --init --recursive"
+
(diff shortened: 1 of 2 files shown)


docs/casts/datalad_convenience_vs_git.sh
```diff
@@ -0,0 +1,109 @@
+say "Fast forward through some prior work..."
+
+say "Let's assume a student developed an algorithm, using Git."
+run "mkdir code; git -C code init; touch code/work; git -C code add work ; git -C code commit -m 'MSc thesis by student A done'"
```
yarikoptic: To be realistic, drop use of -C

mih: That is only the setup. The key bits are the "PI" work.

docs/casts/datalad_convenience_vs_git.sh
```diff
@@ -0,0 +1,109 @@
...
+run "datalad clone --dataset . ../analysis study"
+say "Requesting a particular file automatically obtains all needed subdatasets"
+run "datalad get study/code/work"
+
+say "Apply the code fix, and have it be detected"
+run "echo 'fix' >> study/code/work"
+run "datalad status"
+
+say "Unlike Git, DataLad makes nested datasets feel like a monorepo"
+run "datalad status --recursive --untracked all"
```
kyleam: `--untracked all` isn't relevant for this example and could dropped so there is less the viewer has to process.

docs/casts/datalad_convenience_vs_git.sh
```diff
@@ -0,0 +1,109 @@
...
+run "git submodule update --init --recursive"
+
+say "Arrgh, there is a bug in the code, and the postdoc isn't here to work the fix in. But hey, Git is all distributed -- let's apply it right here, so we can move on with the science..."
+run "echo 'fix' >> study/code/work"
+
+say "Quickly commit the fix, so it can be pushed upstream later..."
+run "git add study/code/work"
+
+say "Erm... Needs to happen in the repository that actually contains the file...Need to look up the boundaries..."
+run "git submodule status --recursive"
```
kyleam: With  a mostly clean hierarchy, `git diff` also makes it fairly easy to spot the repo with the changed file:

```
$ git diff --submodule=diff
Submodule study contains modified content
Submodule code contains modified content
diff --git a/study/code/work b/study/code/work
index e69de29..a609090 100644
--- a/study/code/work
+++ b/study/code/work
@@ -0,0 +1 @@
+fix
```

docs/casts/datalad_convenience_vs_git.sh
kyleam: [ a not relevant quip ]

> To be realistic,

It amused me to see this phrase near a snippet where the "algorithm" was an empty file.

docs/casts/datalad_convenience_vs_git.sh
```diff
@@ -0,0 +1,109 @@
...
+
+say "Ok, let's do this..."
+run "cd study/code"
+run "git add work && git commit -m 'Fix'"
+
+say "Damn, the postdoc didn't specify a branch when adding the submodule. Need to fix, or it will be difficult to push."
```
kyleam: This is wrong, I think.  The branch recorded in .gitmodules doesn't change whether Git will get into a detached state.  It is used to identify the branch from the remote that should be used to update the submodule when `submodule update --remote [...]` is called.

docs/casts/datalad_convenience_vs_git.sh
mih: Thanks! That is much more elegant.

docs/casts/datalad_convenience_vs_git.sh
mih: ok, will reword.

docs/casts/datalad_convenience_vs_git.sh
mih: Done

docs/casts/datalad_convenience_vs_git.sh
mih: Done

docs/casts/datalad_convenience_vs_git.sh
mih: Done

Issue #3103, opened 2019-01-05
`Runner` incapable of "logging" while letting executed command prompt for input 


mih:
Use case needed here: https://github.com/datalad/datalad-revolution/pull/71

Ideally `git commit` should be able to exhibit its normal behavior, i.e. ask for a commit message when none was provided (by launching an editor), even when executed through DataLad's runner. However, when we run `git commit`, we also need to be able to act on a variety of error conditions, hence we need to capture stdout/err. ATM runner can capture the output streams XOR let them through. In shell lingo, we can do

`git commit` and `git commit > log`, but for the desired behavior we need `git commit | tee log`

Looking through the Runner code it is not immediately clear to me where/if such feature could be added, but here is a sketch of the concept: https://shallowsky.com/blog/programming/python-tee.html

yarikoptic:
AFAIK 
- runner supports commands with prompts e.g. asking for credentials. But that is usually done via directly talking to /dev/tty, so it is presented to the user even from within datalad sandwiched within git annex remote call, which is done by datalad itself. We use getpass.getpass for that. see dialog.py
- `| tee log` can also be done 
```
 log_stdout: bool or callable or 'online' or 'offline'
```
so if it is a callable -- you would get line as soon as it is received. That is what we use for e.g. capturing progress reports from git-annex

I could be wrong, but in the context of `git commit`, to display an entire editor, we probably need to surrender any opportunity to receive stdout, and use smth like `os.system` or `pty.spawn` instead of popen..., e.g.
```shell
$> python -c 'import os; os.system("vim 2>/tmp/2")'            
$> ls -ld /tmp/2
-rw------- 1 yoh yoh 0 Jan  5 17:54 /tmp/2
```
but again -- I could be wrong and missing what is needed

mih:
It is perfectly OK to use `vim` through Runner.

```python
from datalad.cmd import Runner
Runner().run('vim', log_stdout=False, log_stderr=False)
```

OTOH using a callable to intercept and pass on anything that comes through the pipe does not work.

```python
Runner().run('vim', log_stdout=lambda x: sys.stdout.write(x), log_stderr=lambda x: sys.sterr.write(x))
```

The reason for that is still unclear to me. I suspect that it never gets to see "a line".

bpoldrack:
FWIW:
Just tried and it seems to me, that - at least for this use case - it works if only stderr is logged. Meaning

```python
from datalad.cmd import Runner
[...]

Issue #3236, opened 2019-03-17
Improve (automated) commit messages


yarikoptic:
ATM automated commit messages ("Added content", "Recorded changes", ...) are useless.
Some (if not majority) of the commands, e.g. `*create` do not even allow to have commit message specified.  I wonder if we could improve upon that.

In many cases the intention/result of the command is clear. e.g. if I `create .../subds`, I think it would be better to commit "Added subdataset subds" and "Recorded changes caused by "Added subdataset subds'".  If there is multiple objects of action, they could be quantified (`X files, Y datasets`, etc).  And I think it should be possible if we collect a list of actions which were taken by the command before `*save` was invoked.

<details>
<summary>I do something like that in the (good, khe khe) and (definetely) old crawler:</summary>

```
    [DATALAD] Added files from extracted archives
    
    Files processed: 6202
     skipped: 4
     +annex: 6168
    Branches merged: incoming->incoming-processed

commit c6259378478bbcc634b3295b0c710886b9eecb01 (origin/incoming, synced/incoming, incoming)
Author: Yaroslav Halchenko <debian@onerussian.com>
Date:   Tue May 16 23:20:05 2017 -0400

    [DATALAD] Updated git/annex from a remote location
    
    URLs processed: 34
    Files processed: 35
     skipped: 34
     +git: 1
```
</details>

Related: 
- https://github.com/datalad/datalad-revolution/pull/71 - for non-automated/user input messages

mih:
I do not believe that adding information to commit messages that can be readily extracted from the rest of the commit itself actually improves the commit messages. What we are lacking is the ability to say "why", not "what" or "how". I think we need https://github.com/datalad/datalad-revolution/pull/71

And for that one, we need an improved Runner that actually permits us from behaving like `git commit` in asking for human input.

yarikoptic:
For "why", indeed the user is the only source of such information.  But "why" usually is described in the extended description, not in the summary of the commit.

But ATM even "what" and "how" - which is what typically summarized in the summary line of the commit message - we do have an idea.  "what" is primarily instructed by the top level invoked command.

yarikoptic:
and some messages are now a complete lie in case of `add`. `save` and `rev-save`'s messages at least do not lie, although as meaningful as having no message (sure thing there were some changes which were recorded -- we do not allow empty commits)

```shell
$> git show
commit 3648ea509bc3d6a90cc52b4ef7899a0bd256d0ee (HEAD -> master)
[...]

Issue #3914, opened 2019-12-10
Saving with (many) staged changes can produce multiple commits


loj:
#### What is the problem?

I used `datalad run` to execute a shell script to convert an mri dataset to a BIDS structure (mostly through moving/renaming files) and it resulted in 6 commit messages.

* all commits had the same run record, but the files that were contained/executed upon in each commit were different
* I did not declare any `inputs` or `outputs` in the `datalad run` call
* the shell script operated on ~22,000+ files.

My assumption is that this is due to the number of files that I'm operating on. Curious on what you guys think.

#### What version of DataLad are you using (run `datalad --version`)? On what operating system (consider running `datalad wtf`)?

```
❱ datalad wtf
# WTF
## configuration <SENSITIVE, report disabled by configuration>
## datalad 
  - version: 0.12.0rc6
  - full_version: 0.12.0rc6
## dataset 
  - path: /data/project/rehab_camcan/cc700_mri
  - repo: AnnexRepo
  - metadata: <SENSITIVE, report disabled by configuration>
## dependencies 
  - tqdm: 4.28.1
  - cmd:annex: 7.20190819+git2-g908476a9b-1~ndall+1
  - cmd:git: 2.20.1
  - cmd:bundled-git: 2.20.1
  - cmd:system-git: 2.24.0
  - cmd:system-ssh: 7.9p1
  - appdirs: 1.4.3
  - boto: 2.44.0
  - exifread: 2.1.2
  - git: 2.1.11
  - gitdb: 2.0.5
  - humanize: 0.5.1
  - iso8601: 0.1.11
  - keyring: 17.1.1
  - keyrings.alt: 3.1.1
  - msgpack: 0.5.6
  - mutagen: 1.40.0
  - requests: 2.21.0
  - wrapt: 1.10.11
## environment 
  - LC_CTYPE: en_US.UTF-8
  - PATH: /home/loj/.dotfiles/bin:/usr/local/sbin:/usr/local/bin:/usr/bin:/bin:/usr/games
  - LANGUAGE: en_US.UTF-8
  - LANG: en_US.UTF-8
  - LC_ALL: en_US.UTF-8
  - GIT_PYTHON_GIT_EXECUTABLE: /usr/lib/git-annex.linux/git
## extensions 
  - ria: 
    - load_error: None
    - description: Helper for the remote indexed archive (RIA) special remote
    - module: ria_remote
    - version: 0.6+6.ge755961
    - entrypoints: 
[...]

PR #4121, opened 2020-02-07
ENH: Enable interactive commit message entry

mih:
The lack of the ability to specify commit messages interactively
(like Git does by default) is a major usability issue when detailed
commit messages are desired (also see gh-3236).

This resurrects gh-3101 using a different and more comprehensive
approach, based on the new WitlessRunner.

Ultimately, this would be a basis for a move towards a more
comprehensive message handling
(see https://github.com/datalad/datalad-revolution/pull/71).

Change summary:
- a config switch `datalad.save.no-message` is introduced to decides
  how to behave, when a commit is attempted, but no message is provided
  - 'generic': maintain previous behavior and use a generic message
  - 'allow': make a commit without a message
  - 'interactive' : Let Git figure it out, using its default behavior
- the tests are set up to use 'generic' to maintain non-interactive
  behavior.
- `GitRepo.commit()` is rewritten to use `WitlessRunner` while
  maintaining API and behavior for all previously supported modes of
  operations.
- An additional `commit --dry-run` is used to preempt errors, but only
  in interactive mode, when no message was given -- when performance
  is a non-issue anyway.

fixes gh-3103

TODO

- [ ] splitting of long file lists
- [ ] splitting of long file lists might yield multiple commits, where messages would need to be typed
       interactively. Consider switching to the model:
    - make any number of commits with an empty message
    - squash commit series into a single commit
    - amend with message

### DIFF
--- a/datalad/__init__.py
+++ b/datalad/__init__.py
@@ -108,6 +108,8 @@ def setup_package():
     _test_states['DATASETS_TOPURL_ENV'] = os.environ.get('DATALAD_DATASETS_TOPURL', None)
     _test_states['DATASETS_TOPURL'] = consts.DATASETS_TOPURL
     os.environ['DATALAD_DATASETS_TOPURL'] = consts.DATASETS_TOPURL = 'http://datasets-tests.datalad.org/'
+    # ensure non-interactive behavior of save when no message was given
+    os.environ['DATALAD_SAVE_NO__MESSAGE'] = 'generic'
 
     # To overcome pybuild overriding HOME but us possibly wanting our
     # own HOME where we pre-setup git for testing (name, email)
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -300,13 +300,20 @@ class WitlessRunner(object):
         if protocol is None:
             # by default let all subprocess stream pass through
             protocol = NoCapture
-        # include the subprocess manager in the asyncio event loop
+        # start a new event loop, which we will close again further down
+        # if this is not done events like this will occur
+        #   BlockingIOError: [Errno 11] Resource temporarily unavailable
+        #   Exception ignored when trying to write to the signal wakeup fd:
+        # It is unclear to me why it happens when reusing an event looped
+        # that it stopped from time to time, but starting fresh and doing
+        # a full termination seems to address the issue
         if sys.platform == "win32":
             # use special event loop that supports subprocesses on windows
             event_loop = asyncio.ProactorEventLoop()
-            asyncio.set_event_loop(event_loop)
         else:
-            event_loop = asyncio.get_event_loop()
+            event_loop = asyncio.SelectorEventLoop()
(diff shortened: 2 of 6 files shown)


adswa:
I haven't looked at the code, but I just tried it and this is really cool! Thanks!

kyleam:
>  splitting of long file lists might yield multiple commits, where messages would need to be typed
> interactively. Consider switching to the model:
>
>    * make any number of commits with an empty message
>    * squash commit series into a single commit
>    * amend with message

[ related to gh-3914 ]

As of Git 2.25, git-commit has `--pathspec-from-file`, so I think we should make use of that when it's available.  Perhaps we should add a kludge for older gits, but I think I'd be okay with the letting this case resolve itself with time.

yarikoptic:
Cool -- it works for me too!
Note though that it causes multiple commit dialogues whenever operating recursively, and that is nohow discussed. Such behavior could be a desired feature or a curse.  It would definitely be a surprise and might trigger a variety of failed operations on "automated setups".
I have tried on `create-test-dataset` command, which also ATM doesn't even have `-m` option.  Contribution attempts such as https://github.com/datalad/datalad/issues/3882 to add `-m` to other commands were frozen until some ultimate decision is (yet to be made) in https://github.com/datalad/datalad/issues/3896 . 
There is also a Draft https://github.com/datalad/datalad/pull/3893 to explicitly trigger specification of the commit message for all (sub)datasets to be saved via `-m ""` .  IMHO it fits majority of use-cases better (than per sub-dataset commit message), but also could be suboptimal in case of some "multi-commit" operations (besides the ones for too long list of files argument). 

So I think the default behavior of automated commit messages should be maintained and ability to specify per-commit message should be made optional (not default as it seems to be now judging from the attempt with `create-test-dataset`). And then may be we should get back to #3896 to decide on how to handle all those cases consistently and with good visibility/usability in user interface (config variables are cool but IMHO aren't "the way")

yarikoptic:
> Note though that it causes multiple commit dialogues whenever operating recursively

I do not think it would be easily possible, but I think it would have been cool if we could to add to the templated commit message in comments some switch to allow users tell "apply this commit message to all subsequent commits for the originating command", e.g. smth like
```

## DataLad commit options.  Modify to adjust behavior of subsequent commits
#  inherit-message: False
#
```

mih:
@yarikoptic Thanks for your thought. I pretty much agree in full. Most of that was already implemented in https://github.com/datalad/datalad-revolution/pull/71 (incl templating). it just died due to running limitations. This time we have a better chance ;-)

However, I don't think it will make 0.12.3 (too much to do), but it will be 0.12.

mih:
There is still much to achieve until interactive messages become a reality. Too much for me to reach the goal without accumulating too much drag. I will therefore close this PR, and reopen a new one, with pretty much the same code, but downscaled goals.

PR #4122, opened 2020-02-07
RF: Turn GitRepo non-method into standalone function

mih:
- rename to `run_gitcommand_on_file_list_chunks`
- move to `cmd.py` next to the runners it can work with
- avoid intermediate str concatenation
- add debug message
- more detailed docstring


mih:
Thanks for the fixes @kyleam

PR #4123, opened 2020-02-08
RF: Make runner code (subjectively) more readable

mih:
Prev. state used very similar variable names, and had (documented)
issues of scope confusion. This is an attempt to make things a bit
clearer.


mih:
Thanks @kyleam ! I keep the suggestion in mind. I have the feeling that I need to go back to this soonish...

PR #4124, opened 2020-02-08
NF: create-sibling-ria command

mih:
This is the code that was used to push the HCP datasets to store.datalad.org -- written by @bpoldrack (see https://github.com/datalad/git-annex-ria-remote/pull/38). It is one of the last pieces of RIA functionality to move into -core (the actual special remote implementation being the very last one). It is being proposed as a separate PR, because it can function without it (HCP doesn't use the ria-special remote), and to keep the complexity of the PR more manageable.

Feedback is very much welcome! Thanks!

Related:

- #4068 consolidation on common operations with `create_sibling`
- UX issues:
  - https://github.com/datalad/git-annex-ria-remote/issues/44
  - https://github.com/datalad/git-annex-ria-remote/issues/45

Update by @bpoldrack :

From my point of view, this is ready. Above mentioned UX issues are addressed and test failures fixed (includes a fix for #4151). A long the way I doscovered #4152 and #4153. However, those are neither fixed herein nor are fixes for that actually required for this PR.

Tests currently depend on installing `https://github.com/datalad/git-annex-ria-remote@transition-to-core`, which removes the conflicting implementation of `create-sibling-ria` therein. While this seems somewhat ugly, it should be a quite short lived approach. Rational: We are moving the special remote to core. I'm currently working on PR #4068 to integrate execution via a persistent remote shell. When this is done, git-annex-ria-remote can go into core completely. If we had an actual dependency on the `ria_remote` package via `setup.py`, this would lead to people installing it, when shortly afterwards it will become at least unnecessary, if not conflicting to have it installed, which is why I consider this approach superior.

Edit by @bpoldrack :

Plans changed. Since there's some time pressure and a new idea re special remote configuration ( https://github.com/datalad/datalad/pull/4124#issuecomment-587487164 ), updated plan for ria-remote's move to core:
  - move ria-remote as is with remote execution factored out for easier RF later on.
  - once it's merged:
      - this PR (`create-sibling-ria`) can remove Travis dependency it properly reference/import within core and be merged
      - then special remote's `initremote` can be adapted to store the annex UUID in the bare repo's config
      - and `clone` then needs to be aware to make use of it

Note, that integration with #4068 is thereby postponed.

So, 3 PRs:
  - the main move to core
  - this very PR with adaptions
  - change `RIARemote.initremote()` and datalad-clone

### DIFF
--- a/datalad/distributed/create_sibling_ria.py
+++ b/datalad/distributed/create_sibling_ria.py
@@ -19,8 +19,8 @@ from datalad.interface.common_opts import (
     recursion_limit
 )
 from datalad.interface.base import (
-    Interface,
     build_doc,
+    Interface,
 )
 from datalad.interface.results import (
     get_status_dict,
@@ -28,14 +28,14 @@ from datalad.interface.results import (
 from datalad.interface.utils import eval_results
 from datalad.support.param import Parameter
 from datalad.support.constraints import (
+    EnsureBool,
+    EnsureChoice,
     EnsureNone,
     EnsureStr,
-    EnsureBool,
-    EnsureChoice
 )
 from datalad.distribution.dataset import (
-    EnsureDataset,
     datasetmethod,
+    EnsureDataset,
     require_dataset,
 )
 from datalad.utils import (
--- a/datalad/distributed/create_sibling_ria.py
+++ b/datalad/distributed/create_sibling_ria.py
@@ -1,0 +1,599 @@
+# emacs: -*- mode: python; py-indent-offset: 4; tab-width: 4; indent-tabs-mode: nil -*-
+# ex: set sts=4 ts=4 sw=4 noet:
+# ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+#
+#   See COPYING file distributed along with the datalad package for the
+#   copyright and license terms.
+#
+# ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+"""Create a sibling in a RIA store"""
+
+__docformat__ = 'restructuredtext'
+
+
+import logging
+import subprocess
+
+from datalad.interface.common_opts import (
+    recursion_flag,
+    recursion_limit
+)
+from datalad.interface.base import (
+    Interface,
+    build_doc,
+)
+from datalad.interface.results import (
+    get_status_dict,
+)
+from datalad.interface.utils import eval_results
+from datalad.support.param import Parameter
+from datalad.support.constraints import (
+    EnsureNone,
+    EnsureStr,
+    EnsureBool,
+    EnsureChoice
+)
+from datalad.distribution.dataset import (
+    EnsureDataset,
+    datasetmethod,
+    require_dataset,
+)
+from datalad.utils import (
+    Path,
+    quote_cmdlinearg,
+    rmtree,
+)
+from datalad.support.exceptions import (
+    CommandError
+)
+from datalad.support.gitrepo import (
+    GitRepo
+)
+from datalad.core.distributed.clone import (
+    decode_source_spec
+)
+from datalad.log import log_progress
+from datalad.distributed.ria_utils import (
+    get_layout_locations,
+    verify_ria_url,
+)
+
+lgr = logging.getLogger('datalad.distributed.create_sibling_ria')
+
+
+@build_doc
+class CreateSiblingRia(Interface):
+    """Creates a sibling to a dataset in a RIA store
+
+    This creates a representation of a dataset in a ria-remote compliant
+    storage location. For access to it two siblings are configured for the
+    dataset by default. A "regular" one and a RIA remote (git-annex
+    special remote).  Furthermore, the former is configured to have a
+    publication dependency on the latter. If not given a default name for
+    the RIA remote is derived from the sibling's name by appending "-ria".
+
+    The store's base path currently is expected to either:
+
+      - not yet exist or
+      - be empty or
+      - have a valid `ria-layout-version` file and an `error_logs` directory.
+
+    In the first two cases, said file and directory are created by this
+    command. Alternatively you can manually create the third case, of course.
+    Please note, that `ria-layout-version` needs to contain a line stating the
+    version (currently '1') and optionally enable error logging (append '|l' in
+    that case). Currently, this line MUST end with a newline!
+
+    Error logging will create files in the `error_log` directory whenever the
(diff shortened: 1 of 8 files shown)


mih:
@bpoldrack can you have a look at the test failures please, it seems we are missing a few of their idiosyncracies in datalad-core (which should preferably by removed). Thx!

bpoldrack:
Added `trust-level` option. @mih: You might want to have a quick look at https://github.com/datalad/datalad/pull/4124/commits/ff095ff24f720d6ede79033e402cace11d115d2c

mih:
@bpoldrack  That last patch tripped the tests

bpoldrack:
Note, that https://github.com/datalad/git-annex-ria-remote/issues/44 and https://github.com/datalad/git-annex-ria-remote/issues/45 are addressed here.

Re test: That's strange. `unexpected keyword argument` should happen locally as well if legit. But doesn't. Hmm.

bpoldrack:
FTR: ATM I have no idea waht's going on on Travis. Can't reproduce and looks like it has nothing to do with my last commit.

However, meanwhile fixing an issue wrt `--existing`. Therefor removed merge label for now.

bpoldrack:
Finally able to reproduce issue on Travis. Reason: Happens only on the merge commit with master. At least twofold:
1. #4151 
2. Somehow a `url...insteadOf` config gets lost when cloning. No matter whether I store it `global` or use `override` with `datalad.cfg`.

mih:
I have lost track of all the moving pieces here. Could you please leave a note in the top comment on the status of this PR. It looks ready function-wise, but there is still the test dependency issue on that branch in ria-remote. What is the general plan?

bpoldrack:
@mih: Updated top-level comment.

mih:
Idea:

- when `create-sibling-ria` has call `initremote` we record the UUID of the special remote in the `config` of the bare repo in the store
- when we `datalad clone` from a store and the special remote is not functional, we check that store config, and `enableremote` the special remote with the matching UUID to use the store URL and access protocol that has just led to a successful clone

mih:
```
% datalad create-sibling-ria -s datastore THISISIT
[INFO   ] create siblings 'datastore' and 'datastore-ria' ... 
[INFO   ] Fetching updates for <Dataset path=/tmp/test> 
CommandError: command '['git', 'fetch', '--verbose', '--progress', '--prune', '--recurse-submodules=no', 'datastore']' failed with exitcode 128
Failed to run ['git', 'fetch', '--verbose', '--progress', '--prune', '--recurse-submodules=no', 'datastore'] at '/tmp/test'.
stdout=
stderr=fatal: unable to find remote helper for 'ria+ssh'

fatal: unable to find remote helper for 'ria+ssh'

% git config -l |grep THISIS
url.ria+ssh://mih@XXX.YYY.de/tmp/tmpstore.insteadof=THISISIT
```

The failure doesn't happen with a URL that doesn't have a user specified. It is likely a URL parsing issue.

bpoldrack:
> The failure doesn't happen with a URL that doesn't have a user specified. It is likely a URL parsing issue.

Addressed in https://github.com/datalad/datalad/pull/4124/commits/b4533eabee17416f8770adf08ee96438bcc0e264

bpoldrack:
As for the annex UUID in bare repo's config: Will move ria-remote to core first and only then implement it in its `initremote`. This leads to updated plans - see toplevel.

datalad/distributed/create_sibling_ria.py
```diff
@@ -0,0 +1,586 @@
...
+from datalad.interface.results import (
+    get_status_dict,
+)
+from datalad.interface.utils import eval_results
+from datalad.support.param import Parameter
+from datalad.support.constraints import (
+    EnsureNone,
+    EnsureStr,
+    EnsureBool,
+    EnsureChoice
```
yarikoptic: Ideally import lines should be sorted within the tuple. (Here and elsewhere in the diff)

datalad/distributed/create_sibling_ria.py
```diff
@@ -0,0 +1,586 @@
...
+    <dataset id>.<annex uuid of the remote>.log showing 'who' ran into this
+    issue with what dataset. Since this logging can potentially leak personal
+    data (like local file paths for example) it can be disabled from the client
+    side via "annex.ria-remote.<RIAREMOTE>.ignore-remote-config".
+
+    Todo
+    ----
+    Where to put the description of a RIA store (see below)?
+
+    The targeted layout of such a store is a tree of datasets, starting at the
```
yarikoptic: Make it "tree of directories containing datasets at the lowest level", since otherwise suggests hierarchy of Datasets

datalad/distributed/create_sibling_ria.py
```diff
@@ -0,0 +1,586 @@
...
+            `git init --shared` and are described in its documentation.""",
+            constraints=EnsureStr() | EnsureBool() | EnsureNone()),
+        group=Parameter(
+            args=("--group",),
+            metavar="GROUP",
+            doc="""Filesystem group for the repository. Specifying the group is
+            crucial when [CMD: --shared=group CMD][PY: shared="group" PY]""",
+            constraints=EnsureStr() | EnsureNone()),
+        ria_remote=Parameter(
+            args=("--no-ria-remote",),
```
yarikoptic: This name is confusing given the have of the command. May be make it `--no-ria-archive` if that what it is?

datalad/distributed/create_sibling_ria.py
```diff
@@ -0,0 +1,586 @@
...
+            dest='ria_remote',
+            doc="""Whether to establish remote indexed archive (RIA) capabilties
```
yarikoptic: Check how we phrased other `--no-something` options, since currently description with "if enabled" is not clear

datalad/distributed/create_sibling_ria.py
bpoldrack: Potential confusion - true. But it's not about an archive either. It's about whether or not to have a special remote, since it can be useful to have that ID based tree of bare repos without any annex object.

bpoldrack:
@yarikoptic : Addressed your comments (and pushed) except for the one answered. Don't see a better name for `--no-ria-remote` ATM.

bpoldrack:
FTR: Windows failure is in `datalad.tests.test_utils.test_get_open_files` and looks like usual path resolution issue with short paths. No idea how this would be related to this PR.

datalad/distributed/create_sibling_ria.py
mih: I agree with @bpoldrack

PR #4103, opened 2020-01-29
RF: GitRepo.clone clone_options -> options

yarikoptic:
Many other methods in GitRepo have ability to pass options to git.
All of them, but add* (overloaded by git-annex, separate storry to
try/tell), seems to have "options" kwarg (some are on the 2nd line, not shown),
not command_options, since association to the command is obvious.

    $> git grep 'def .*options=' datalad/support/*repo.py
    datalad/support/annexrepo.py:    def get_toppath(cls, path, follow_up=True, git_options=None):
    datalad/support/annexrepo.py:    def add_remote(self, name, url, options=None):
    datalad/support/annexrepo.py:    def get(self, files, remote=None, options=None, jobs=None, key=False):
    datalad/support/annexrepo.py:    def add(self, files, git=None, backend=None, options=None, jobs=None,
    datalad/support/annexrepo.py:    def add_(self, files, git=None, backend=None, options=None, jobs=None,
    datalad/support/annexrepo.py:    def lock(self, files, options=None):
    datalad/support/annexrepo.py:    def unlock(self, files, options=None):
    datalad/support/annexrepo.py:    def adjust(self, options=None):
    datalad/support/annexrepo.py:    def unannex(self, files, options=None):
    datalad/support/annexrepo.py:    def add_url_to_file(self, file_, url, options=None, backend=None,
    datalad/support/annexrepo.py:    def add_urls(self, urls, options=None, backend=None, cwd=None,
    datalad/support/annexrepo.py:    def drop(self, files, options=None, key=False, jobs=None):
    datalad/support/annexrepo.py:    def drop_key(self, keys, options=None, batch=False):
    datalad/support/annexrepo.py:    def whereis(self, files, output='uuids', key=False, options=None, batch=False):
    datalad/support/annexrepo.py:    def commit(self, msg=None, options=None, _datalad_msg=False,
    datalad/support/annexrepo.py:    def copy_to(self, files, remote, options=None, jobs=None):
    datalad/support/annexrepo.py:    def __init__(self, batch_size=0, git_options=None):
    datalad/support/annexrepo.py:    def __init__(self, annex_cmd, git_options=None, annex_options=None, path=None,
    datalad/support/gitrepo.py:def kwargs_to_options(func, split_single_char_options=True,
    datalad/support/gitrepo.py:    def get_toppath(cls, path, follow_up=True, git_options=None):
    datalad/support/gitrepo.py:    def add(self, files, git=True, git_options=None, update=False):
[...]

PR #4125, opened 2020-02-09
RF: WitlessRunner reimplemented with asyncio 

mih:
Previous implementation had deadlock issues that I could not resolve.
This code is inspired by the highly informative

https://pymotw.com/3/asyncio/subprocesses.html

Reimplementations for the previous output processors are provided too.
They are now using "protocols" that are natively provided by asyncio,
and offer a conceptually similar, but extremely superior and more
flexible approach to the problem.

A no longer relevant test case was removed.

TODO:
- [x] work into #4121 and verify it solves the deadlock issue
- [x] work into #4103 and verify things keep working

### DIFF
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -300,13 +300,20 @@ class WitlessRunner(object):
         if protocol is None:
             # by default let all subprocess stream pass through
             protocol = NoCapture
-        # include the subprocess manager in the asyncio event loop
+        # start a new event loop, which we will close again further down
+        # if this is not done events like this will occur
+        #   BlockingIOError: [Errno 11] Resource temporarily unavailable
+        #   Exception ignored when trying to write to the signal wakeup fd:
+        # It is unclear to me why it happens when reusing an event looped
+        # that it stopped from time to time, but starting fresh and doing
+        # a full termination seems to address the issue
         if sys.platform == "win32":
             # use special event loop that supports subprocesses on windows
             event_loop = asyncio.ProactorEventLoop()
-            asyncio.set_event_loop(event_loop)
         else:
-            event_loop = asyncio.get_event_loop()
+            event_loop = asyncio.SelectorEventLoop()
+        asyncio.set_event_loop(event_loop)
+        # include the subprocess manager in the asyncio event loop
         return_code, results = event_loop.run_until_complete(
             run_async_cmd(
                 event_loop,
@@ -317,6 +324,9 @@ class WitlessRunner(object):
                 env=self.env,
             )
         )
+        # terminate the event loop, cannot be undone, hence we start a fresh
+        # one each time (see BlockingIOError notes above)
+        event_loop.close()
         # when we are here the process finished, take output from bytes to string
         output = tuple(
             o.decode(getpreferredencoding(do_setlocale=False))
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -157,13 +157,16 @@ async def run_async_cmd(loop, cmd, protocol, stdin, **kwargs):
         stderr=asyncio.subprocess.PIPE if protocol.proc_err else None,
         **kwargs
     )
+    transport = None
     try:
         lgr.debug('Launching process %s', cmd)
         transport, protocol = await proc
         lgr.debug('Waiting for process %i to complete', transport.get_pid())
         await cmd_done
     finally:
-        transport.close()
+        # protect against a crash whe launching the process
+        if transport:
+            transport.close()
 
     return cmd_done.result()
 
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -234,8 +234,9 @@ class WitlessProtocol(asyncio.SubprocessProtocol):
         lgr.debug('Process %i started', self.pid)
 
     def pipe_data_received(self, fd, data):
-        if lgr.isEnabledFor(logging.DEBUG):
-            lgr.debug(
+        if lgr.isEnabledFor(5):
+            lgr.log(
+                5,
                 'Read %i bytes from %i[%s]',
                 len(data), self.pid, self.FD_NAMES[fd])
         # store received output if stream was to be captured
@@ -286,9 +287,11 @@ class KillOutput(WitlessProtocol):
     proc_err = True
 
     def pipe_data_received(self, fd, data):
-        if lgr.isEnabledFor(logging.DEBUG):
-            lgr.debug('Discarded %i bytes from %i[%s]',
-                      len(data), self.pid, self.FD_NAMES[fd])
+        if lgr.isEnabledFor(5):
+            lgr.log(
+                5,
+                'Discarded %i bytes from %i[%s]',
+                len(data), self.pid, self.FD_NAMES[fd])
 
 
 class WitlessRunner(object):
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -20,6 +20,8 @@ import atexit
 import functools
 import tempfile
 from locale import getpreferredencoding
+import asyncio
+from collections import namedtuple
 
 from collections import OrderedDict
 from .support import path as op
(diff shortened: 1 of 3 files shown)


mih:
Travis shows this (and I can also see that locally):

```
...
datalad.core.local.tests.test_create.test_create_relpath_semantics('parentds', None, 'relpath') ... Exception ignored when trying to write to the signal wakeup fd:
BlockingIOError: [Errno 11] Resource temporarily unavailable
Exception ignored when trying to write to the signal wakeup fd:
...
```

https://bugs.python.org/issue21595

kyleam:
> Exception ignored when trying to write to the signal wakeup fd:
> BlockingIOError: [Errno 11] Resource temporarily unavailable
> Exception ignored when trying to write to the signal wakeup fd:
> ```

Yeah, I was seeing a similar issue on my asyncio attempt at gh-3660:

https://github.com/datalad/datalad/blob/f79b5f2db8871e945635aa9bf52557b425994b0a/runner-asyncio-notes.org#L2-L16

mih:
It is mysterious. Common reason for this error are:

- event loop not actually running (confirmed not the case)
- to many simultaneous subprocesses (confirmed max 1)

On my local machine the errors start in tests where no async code is involved. I read that a stopped event loop could "accumulate" wakeup events. I saw that they way I wrote it, has the eventloop stopped whenever there is no runner run ongoing (made sense to me). However, if async requires an event loop wakeup from time to time, it would explain the delayed onset of the issue.

mih:
With async debugging on and `python -Wdefault -m nose -v -s datalad.support.tests.test_gitrepo` I get:

```
/home/mih/env/datalad3-dev/lib/python3.7/site-packages/msgpack/exceptions.py:10: ResourceWarning: unclosed <socket.socket fd=6, family=AddressFamily.AF_UNIX, type=SocketKind.SOCK_STREAM, proto=0, raddr=/run/user/1000/bus>
  class BufferFull(UnpackException):
ResourceWarning: Enable tracemalloc to get the object allocation traceback
/home/mih/env/datalad3-dev/lib/python3.7/site-packages/msgpack/exceptions.py:10: ResourceWarning: unclosed <socket.socket fd=7, family=AddressFamily.AF_UNIX, type=SocketKind.SOCK_STREAM, proto=0, raddr=/run/user/1000/bus>
  class BufferFull(UnpackException):
ResourceWarning: Enable tracemalloc to get the object allocation traceback
/home/mih/env/datalad3-dev/lib/python3.7/site-packages/msgpack/exceptions.py:10: ResourceWarning: unclosed <socket.socket fd=9, family=AddressFamily.AF_UNIX, type=SocketKind.SOCK_STREAM, proto=0, raddr=/run/user/1000/bus>
  class BufferFull(UnpackException):
ResourceWarning: Enable tracemalloc to get the object allocation traceback
/home/mih/env/datalad3-dev/lib/python3.7/site-packages/msgpack/exceptions.py:10: ResourceWarning: unclosed <socket.socket fd=11, family=AddressFamily.AF_UNIX, type=SocketKind.SOCK_STREAM, proto=0, raddr=/run/user/1000/bus>
  class BufferFull(UnpackException):
ResourceWarning: Enable tracemalloc to get the object allocation traceback
...
sys:1: ResourceWarning: unclosed file <_io.BufferedReader name=3>
sys:1: ResourceWarning: unclosed file <_io.BufferedReader name=5>
```

Nothing of relevance, it seems.

mih:
I have pushed a seemingly working workaround for the issue. I use a fresh event loop for each `run()` invocation.

datalad/cmd.py
```diff
@@ -141,27 +143,106 @@ def run_gitcommand_on_file_list_chunks(func, cmd, files, *args, **kwargs):
...
+    finally:
+        # protect against a crash whe launching the process
+        if transport:
+            transport.close()
+
+    return cmd_done.result()
+
+
+class WitlessProtocol(asyncio.SubprocessProtocol):
+
```
kyleam: Given this is still in draft mode, you're probably planning on updating this and `run_async_cmd` with docstrings.   I think it'd be good to mention the link you put in the pr description (https://pymotw.com/3/asyncio/subprocesses.html), since this appears to follow that code pretty closely.

datalad/cmd.py
```diff
@@ -141,27 +143,106 @@ def run_gitcommand_on_file_list_chunks(func, cmd, files, *args, **kwargs):
...
+        self.pid = None
+        super().__init__()
 
+    def connection_made(self, transport):
+        self.transport = transport
+        self.pid = transport.get_pid()
+        lgr.debug('Process %i started', self.pid)
+
+    def pipe_data_received(self, fd, data):
+        lgr.debug('Read %i bytes from %i[%s]', len(data), self.pid, self.FD_NAMES[fd])
```
kyleam: I think it's probably reasonable to assume this could be called many times and that `data` could be somewhat large.  In that case, it probably makes sense to guard this `.debug` call with `logger.isEnabledFor`.  (The same applies to `KillOutput.pipe_data_received`.)

datalad/cmd.py
mih: Yes. And will do.

datalad/cmd.py
mih: True!

mih:
> I have only a superficial understanding of asyncio protocols, but this look pretty appealing from an implementation standpoint. Though, as with other times when I've looked at asyncio stuff, I feel like there are lots of potential ways to accomplish similar things, and I don't have a good grasp of alternatives. (Plus, my opinion shouldn't count for much here because in [gh-4080](https://github.com/datalad/datalad/pull/4080) I blindly took your word for it that your use of .peek() didn't block :] )

:blush: 

Yeah, let's hope it works out better this time.. I am a bit tired of writing replacement runners, and also ashamed to need to ask for your review each and every time....

datalad/cmd.py
mih: Done.

datalad/cmd.py
mih: Done.

kyleam:
>  Even if we don't do proper benchmarks, though, it would be nice to run benchmarks/scripts/heavyout.py through WitlessRunner and Runner to get a sense on where things stand. (Based on what [I was seeing in gh-3660](https://github.com/datalad/datalad/pull/3660#issuecomment-531008561), it might be worth varying the values a bit.) I can have a look at that today.


Here's a script that compares the two.

<details>
<summary>script</summary>

```sh
#!/bin/sh

set -eu

cd "$(mktemp -d --tmpdir dl-runner-XXXXXXX)"
echo $PWD

write_cmd () {
    nlines="${1:-100}"
    nchars="${2:-100}"

    echo
    echo "Making command with $nlines lines per stream, $nchars characters per line"
    cat >heavy.py <<-EOF
	import sys
	
	if __name__ == "__main__":
	    x = 'A' * ($nchars - 1) + '\n'
	    for i in range($nlines):
	        s = "%d" % i + x
	        sys.stdout.write(s)
	        sys.stderr.write(s)
	EOF
}

timeit () {
    echo "Runner()"
    python -m timeit -s 'from datalad import cmd; runner = cmd.Runner()' \
           'runner.run(["python", "heavy.py"], log_stdout=True, log_stderr="offline")'
    echo "WitlessRunner()"
    python -m timeit -s 'from datalad import cmd; runner = cmd.WitlessRunner()' \
           'runner.run(["python", "heavy.py"], protocol=cmd.StdOutErrCapture)'
}

time_with_nlines () {
    write_cmd $1
    timeit
}

time_with_nlines 1
time_with_nlines 10
time_with_nlines 100
time_with_nlines 1000
time_with_nlines 10000
time_with_nlines 100000
time_with_nlines 1000000
```

</details>

That has the following output on my machine:

<details>
<summary>output</summary>

```
/tmp/dl-runner-vS9TCOs

Making command with 1 lines per stream, 100 characters per line
Runner()
20 loops, best of 5: 12.8 msec per loop
WitlessRunner()
20 loops, best of 5: 12.9 msec per loop

Making command with 10 lines per stream, 100 characters per line
Runner()
20 loops, best of 5: 12.8 msec per loop
WitlessRunner()
20 loops, best of 5: 12.8 msec per loop

Making command with 100 lines per stream, 100 characters per line
Runner()
20 loops, best of 5: 12.9 msec per loop
WitlessRunner()
20 loops, best of 5: 12.9 msec per loop

Making command with 1000 lines per stream, 100 characters per line
Runner()
20 loops, best of 5: 13.9 msec per loop
WitlessRunner()
20 loops, best of 5: 14 msec per loop

Making command with 10000 lines per stream, 100 characters per line
Runner()
10 loops, best of 5: 21 msec per loop
WitlessRunner()
10 loops, best of 5: 22.4 msec per loop

Making command with 100000 lines per stream, 100 characters per line
Runner()
5 loops, best of 5: 92.1 msec per loop
WitlessRunner()
2 loops, best of 5: 99.1 msec per loop

Making command with 1000000 lines per stream, 100 characters per line
Runner()
1 loop, best of 5: 783 msec per loop
WitlessRunner()
1 loop, best of 5: 862 msec per loop
```

</details>

yarikoptic:
FWIW: If the last two measurements didn't diverge, I would have said "too consistent, there must have been something wrong in how comparison was done" :-)

mih:
@kyleam  My laptop yields very similar results that are structurally identical.

datalad/cmd.py
```diff
@@ -141,27 +143,159 @@ def run_gitcommand_on_file_list_chunks(func, cmd, files, *args, **kwargs):
...
+        super().__init__()
+
+    def connection_made(self, transport):
+        self.transport = transport
+        self.pid = transport.get_pid()
+        lgr.debug('Process %i started', self.pid)
+
+    def pipe_data_received(self, fd, data):
+        if lgr.isEnabledFor(logging.DEBUG):
+            lgr.debug(
```
jbpoline: @yarikoptic:  `lgr.debug` seems to be too high level for such type of messages. could you move it under `lgr.log(5, ` or even below?

mih: Done @jbpoline

mih:
OK, one last test run -- will merge when done.

PR #4127, opened 2020-02-10
OPT/RF: AnnexRepo.commit - remove invocation to commit --dry-run + add check for unknown files

### LINKED ISSUE #4126: AnnexRepo.commit() investigate performance of test for untracked files
Here is the code:

https://github.com/datalad/datalad/blob/7afff524f5b5ad36f1c2f204a7965407f7ac1055/datalad/support/annexrepo.py#L2569-L2572

This uses a dry-run commit to seemingly test for untracked files given to commit. This can be a fairly expensive test to make.

It is unclear to me why it is done, because a potential exception is not handled, hence will blow up in the same way that it would have with a real commit.

If it needs to be done, we should investigate whether a call to `GitRepo.get_content_info()` is able to perform this test faster.

yarikoptic:
Closes #4126

This PR is to provide OPT/RF aiming to not change behavior. Subsequent issue would be to point to desired further RF.

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -81,6 +81,7 @@ from .external_versions import external_versions
 from .exceptions import (
     CommandNotAvailableError,
     CommandError,
+    FileNotInRepositoryError,
     FileNotInAnnexError,
     FileInGitError,
     AnnexBatchCommandError,
@@ -2566,11 +2567,6 @@ class AnnexRepo(GitRepo, RepoInterface):
         if files:
             files = assure_list(files)
 
-            # Raise FileNotInRepositoryError if `files` aren't tracked.
-            super(AnnexRepo, self).commit(
-                "dryrun", options=["--dry-run", "--no-status"],
-                files=files)
-
         alt_index_file = None
         try:
             # we might need to avoid explicit paths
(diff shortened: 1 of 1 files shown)


kyleam:
I said:

> * [...] this is not a heavily used part of code because the `save` command does not go through it.

Grepping through the codebase, I'm not sure there is any command (except the obsolete `interface.save`) that calls `AnnexRepo.commit` with a values for `files`.

yarikoptic:
do you remember if there is a reason to keep obsolete `interface.save` around?  may be we should just kill both AnnexRepo.commit and `interface.save`?

kyleam:
> do you remember if there is a reason to keep obsolete `interface.save` around?

The obsolete `add` still depends on `interface.save`.  The "remove add/interface.save" waters were tested a bit with gh-3735.

Aside from some compatibility-window for third-party code that uses `add`, I think the main reason is just compatibility with extensions.   Now that a non-rc 0.12 is out, we could have releases in all the extensions that rework them as needed and bump their minimal datalad dependency to 0.12.

mih:
FTR: I would be delighted to get rid of all of this code!!! And its tests! (and yes, I did port all that still make sense)

mih:
Old `save` and `add` are gone now.

yarikoptic:
superseded by #4168 which removed that `.commit`

PR #4135, opened 2020-02-11
ENH/RF: change how we map credentials to env variables

yarikoptic:
This changes might break somebody"s workflow who somehow discovered this feature.
We might want to consult old version too

With this change we could also request parts for the composite credential,
such as loris token, where they have index like :1, which we would sanitize now.

Here is how the one for loris test would look like: DATALAD_KEYRING_loris_demo_1__token

TODO:
- [ ] add basic documentation about credentials (in the docs) and how they could be specified via env vars 
- [ ] add a test if we don't have any (if nothing breaks ATM)
- [ ] check interaction with config and `wtf` to make sure we don't leak those

### DIFF
--- a/datalad/support/keyring_.py
+++ b/datalad/support/keyring_.py
@@ -42,10 +42,16 @@ class Keyring(object):
     def _get_service_name(cls, name):
         return "datalad-%s" % str(name)
 
+    @staticmethod
+    def _get_envvar_name(name, field):
+        """Return env variable which would be checked first for a credential"""
+        return ('DATALAD_KEYRING_%s__%s' % (name, field))\
+            .replace('-', '_').replace(':', '_')
+
     # proxy few methods of interest explicitly, to be rebound to the module's
     def get(self, name, field):
         # consult environment, might be provided there and should take precedence
-        env_var = ('DATALAD_%s_%s' % (name, field)).replace('-', '_')
+        env_var = self._get_envvar_name(name, field)
         if env_var in os.environ:
             return os.environ[env_var]
         return self._keyring.get_password(self._get_service_name(name), field)


yarikoptic:
Any feedback/objections/recommendations @datalad/developers before I proceed to finalize this PR?

kyleam:
> Any feedback/objections/recommendations before I proceed to finalize this PR?

Yes.  I'm too dense to connect the description and diff to the problem this addressing.

mih:
This sounds relevant to us, but I am with @kyleam in being uncertain what and why this is done.

yarikoptic:
Thank you for asking!  Now it is a bit vague, and I think it might need to be improved further, i.e. not to replace with '_' but something else... or we could just let it be unless someone sees a possible way to cause collision or error in parsing if we decide to parse those env vars to separate name from field. (e.g. some `crazy--name` would cause adding `__` which is used as a separator between name and field).

I think the reason was vaguely described in "With this change we could also request parts for the composite credential, such as loris token, where they have index like :1, which we would sanitize now." .  
So first of all this change adds replacing `:` with `_` making it possible to accept credentials via env vars with `:` in the name somewhere (would not be possible otherwise).  
Then it changed to separate the two entities with `__` instead of `_` so there is less probability of collision between a name and a field in case we decide to parse it (I do not think we do ATM, we just match to the target env var).

I hope that makes more sense now

datalad/support/keyring_.py
```diff
@@ -42,10 +42,16 @@ def _keyring(self):
     def _get_service_name(cls, name):
         return "datalad-%s" % str(name)
 
+    @staticmethod
+    def _get_envvar_name(name, field):
+        """Return env variable which would be checked first for a credential"""
+        return ('DATALAD_KEYRING_%s__%s' % (name, field))\
```
yarikoptic, line 48:
```suggestion
        return ('DATALAD_KEYRING_%s_%s' % (name, field)) \
```

mih:
FTR: We are actively using the previous setup for providing credentials to datalad via env variables. It would be good to not have it break suddenly, but enable a transition.

mih:
Closing this after 6+ months of inactivity.

Issue #4090, opened 2020-01-27
migrate to pytest or nose2


yarikoptic:
#### What is the problem?

According to @effigies issue https://github.com/nipy/nibabel/issues/864 we might soon experience fallout of nose.  So, sooner or later we will need to migrate from it to pytest (or some other helper/runner if there is any, but unlikely pure unittest2 or alike). 

# pytest

I guess it will be our choice.  I personally have love/hate relationship with pytest.  IIRC major obstacles will be
- parametric tests.   pytest dropped support for nose-style generator-based parametric tests awhile back, so we will need to convert any parametric test like that to be implemented "pytest-way"
- decorators-vs-fixtures. Some of our decorators pass positional args into the tests. I would need to recall details on what we encountered in reproman, but I believe pytest would consider positional args as fixtures. So we would need to convert them to keyword args to avoid that or in some places convert our decorators into proper pytest fixtures.
- as we start using more of straight `assert` calls, instead of our (nose) helpers like `assert_smth`, we might need to provide pytest with ways to to be totally "pytest-ified" (later) we might like to provide our custom assertion failure handlers: https://docs.pytest.org/en/latest/assert.html#defining-your-own-explanation-for-failed-assertions  

# other alternatives mentioned by others

- nose2
   - cons: seems already not actively developed (last release 0.10.0 (2020-01-27) is almost 2 yo), on my attempts I did not see an easy pattern to migrate over to it
- https://packages.debian.org/python3-testrepository 
- https://packages.debian.org/python3-stestr

effigies:
There is also nose2. I haven't really looked into it deeply.

yarikoptic:
Indeed... Last time I had looked at it was years ago and it gave me feeling I had after trying to migrate from cfengine to cfengine2... But nose2 is indeed there, seems to be slowly but going forward, should provide easier transition than to pytest (haven't spotted any allow stoppers in description of differences from nose).

djarecka:
@yarikoptic - let me know if you will need help. once `nibabel` is done, I can help with `datalad`

yarikoptic:
Than you @djarecka ! I guess for us it might be worth first trying out nose2. And first starting with centralizing all testing support imports, so we don't import from nose but from datalad.tests.utils. that should simplify future switch

djarecka:
@yarikoptic - ok, I know nothing about `nose2`, so don't have opinion.

[...]

PR #4136, opened 2020-02-12
RF: centralize nose imports

yarikoptic:
I have left only nose imports in package level fixture (btw, seems nose2 might not even support those according to docs), and 3rd party externals we included to facilitate `datalad test` command:
```
$> git grep nose | grep import | grep -v support/third                                           
datalad/__init__.py:    from datalad.support.third.nosetester import NoseTester
datalad/__init__.py:    from nose.ext import dtcompat
datalad/__init__.py:    from nose.plugins import capture, multiprocess, plugintest
datalad/cmdline/tests/test_main.py:    # - python -c "import nose; nose.main()" -s -v datalad/cmdline/tests/test_main.py would result in:
datalad/tests/test_tests_utils.py:from .utils import ignore_nose_capturing_stdout
datalad/tests/utils.py:from nose.plugins.attrib import attr
datalad/tests/utils.py:from nose.tools import (
datalad/tests/utils.py:from nose.tools import assert_set_equal
datalad/tests/utils.py:from nose.tools import assert_is_instance
datalad/tests/utils.py:from nose import SkipTest
datalad/tests/utils.py:        from nose import SkipTest
datalad/tests/utils.py:from nose.plugins.attrib import attr
tools/checkpwd.py:from nose.plugins.base import Plugin
tools/checkpwd.py:from nose.util import src, tolist
tools/checkpwd.py:    import nose
```
Ref: #4090


mih:
The rest of the package uses absolute imports (at least the code that has been touched in the last two years). I think we should use that style here too.

yarikoptic:
> The rest of the package uses absolute imports (at least the code that has been touched in the last two years). I think we should use that style here too.
> 
> https://stackoverflow.com/a/4209771

yeap, comments there clarify that PEP8 no longer says that etc.  
I find relative imports less confusing IMHO since they **explicitly** state which package (i.e., **this** package) to import from, not some other "datalad" possibly available somewhere.  They also make those modules portable/borrowable easily into other projects, since then they could be just copied over, and all imports would still work as long as hierarchy preserved.   They particularly make lots of sense in `test_` modules, since typically you test modules defined in `..`, and breeding long import paths to just get "upstairs" is tedious.  
Having said that, I am ok to change them to absolute if that is desired to stay consistent.

mih:
> I find relative imports less confusing IMHO since they **explicitly** state which package (i.e., **this** package) to import from, not some other "datalad" possibly available somewhere.

Yeah, I am aware of this rational, but could never get behind it. For me the RF variant it optimizes for just doesn't seem to happen. Moreover, I do not consider relative imports ala `from .utils` anyhow more obvious than `from datalad...utils`, especially with:

```
% find . -name utils.py
./datalad/interface/utils.py
./datalad/distribution/utils.py
./datalad/ui/utils.py
./datalad/support/tests/utils.py
./datalad/tests/utils.py
./datalad/downloaders/tests/utils.py
./datalad/utils.py
```

I know that you did `....tests.utils` for this purpose, but the general point still holds IMHO. And lastly:

> They particularly make lots of sense in test_ modules, since typically you test modules defined in `..`

True, but in this PR the imports are not the *subject* of the tests, but are generic utilities that are always imported from the same location, regardless of the particular test. Hence when the file moves, these imports would need to be adjusted.

Anyways: yes, I'd prefer to use absolute imports for consistency ;-)

yarikoptic:
done

mih:
I will shortly make a follow-up push -- if we edit these many location we should perform more normalization then just this.

yarikoptic:
Minimal edit as is now might should conflicts which would come with more normalization

mih:
Pushed. Also fixed few spots that were missed before.

effigies:
Re: relative/absolute imports, the text discouraging relative imports was declared outdated by Guido in 2010: https://mail.python.org/pipermail/python-dev/2010-October/104476.html

yarikoptic:
> Pushed. Also fixed few spots that were missed before.

whatever script your are using for normalization, it has bugs. Will push fixes shortly

yarikoptic:
I have joined @mih on the crusade to maximize the chance of possible conflicts with master in the other staged PRs by introducing here changes unrelated to the original topic of the PR and will be pushing more of "normalization" (which BTW should also include sorting of the entries in the import "tuples") shortly.  Rationale: if we do it, we better do it once.

mih:
Merging. Any intermediate merge is likely going to introduce conflicts.

PR #4140, opened 2020-02-13
BF: export-to-figshare - make default title include #id@version, ask for the title in UI

### LINKED ISSUE #4138: export-to-figshare can fail with 400 response code from figshare
Originally reported by @dnkennedy in not so public communication.

I have at first failed to reproduce, to only run into another issue https://github.com/datalad/datalad/issues/4137 .  So there seems to be something specific to the dataset to be exported:

```shell
/tmp > datalad create ds && cd ds && datalad create -d . subds                                                                   
[INFO   ] Creating a new annex repo at /tmp/ds 
create(ok): /tmp/ds (dataset)                                                                                                                                                                                                      
[INFO   ] Creating a new annex repo at /tmp/ds/subds 
add(ok): subds (file)                                                                                                                                                                                                              
add(ok): .gitmodules (file)
save(ok): . (dataset)
create(ok): subds (dataset)
action summary:
  add (ok: 2)
  create (ok: 1)
  save (ok: 1)

/tmp/ds > datalad export-to-figshare                             
[INFO   ] Exporting current tree as an archive under /tmp/ds since figshare does not support directories 
[INFO   ] Uploading /tmp/ds/datalad_ab192df4-4dee-11ea-ac2c-956cf9ab033f.zip to figshare 
Article
Would you like to create a new article to upload to?  If not - we will list existing articles (choices: yes, no): yes

[ERROR  ] Error status Got return code 400 for <function post at 0x7f721e081830>(https://api.figshare.com/v2/account/articles. [export_to_figshare.py:__call__:63] (RuntimeError) 
```

yarikoptic:
Decided not to expose title specification in API since the whole experience is interactive, but I guess it could be added.

Closes #4138

### DIFF
--- a/datalad/plugin/export_to_figshare.py
+++ b/datalad/plugin/export_to_figshare.py
@@ -14,6 +14,9 @@ from datalad.utils import unlink
 from datalad.interface.base import Interface
 from datalad.interface.base import build_doc
 
+import logging
+lgr = logging.getLogger('datalad.export_to_figshare')
+
 
 class FigshareRESTLaison(object):
     """A little helper to provide minimal interface to interact with Figshare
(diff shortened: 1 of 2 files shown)


yarikoptic:
ok, earned an approval, thank you @kyleam

PR #4147, opened 2020-02-13
ENH: safe-guard --existing-replace for cases of populated directories

### LINKED ISSUE #4145: `create-sibling --existing replace` is dangerous
It will perform a `rm -rf` on the remote end. No check if it would wipe out the last existing copy of a file, or whether there even is a dataset sibling.

I propose the complete removal of this functionality. It is totally unexpected that a command that is called `create...` destroys something remote:

```
/tmp % mkdir precious
/tmp % echo "all my secret keys" > precious/keys
/tmp % datalad create reckless
/tmp % cd reckless
/tmp/reckless (git)-[master] % datalad create-sibling localhost:/tmp/precious
[ERROR  ] Target path /tmp/precious already exists. And it fails to rmdir (rmdir: failed to remove '/tmp/precious': Directory not empty). [create_sibling.py:_create_dataset_sibling:161] (RuntimeError) 

# death
/tmp/reckless (git)-[master] % datalad create-sibling localhost:/tmp/precious --existing replace
create_sibling(ok): . (dataset)
% ls -l /tmp/precious
total 0

```

yarikoptic:
A safe guard to Close #4145 

If remote directory contains some files/directories, we would ask user whenever session is interactive or just raise an exception if not interactive

edit 1: I have been thinking about it and I maintain the opinion that `--existing=replace` is already an explicit indication from the user of the intent to replace (remove, recreate) a possibly existing remote dataset.  Adding the interactive question or raising an exception (if not interactive) goes against the explicitly expressed desire to "replace".  So, what if we do that (question or exception) only if remote location does not already have `.git` and `.datalad/config`?  This way we would safe guard against some possible erroneous invocations, but would allow for automated (no interaction) and **explicitly** specified behavior. 

TODOs (if agreed that it is the way to go)
- [x] add tests
- ? add config option (or could be mode "force-replace"?) to be able to "proceed" instead of raising an exception in non-interactive mode or asking in interactive

### DIFF
--- a/datalad/distribution/create_sibling.py
+++ b/datalad/distribution/create_sibling.py
@@ -305,16 +305,21 @@ def _create_dataset_sibling(
 def _ls_remote_path(ssh, path):
     try:
         # yoh tried ls on mac
-        out, err = ssh("ls -a1 {}".format(sh_quote(path)))
+        out, err = ssh("ls -A1 {}".format(sh_quote(path)))
+        if err:
+            # we might even want to raise an exception, but since it was
+            # not raised, let's just log a warning
+            lgr.warning(
+                "There was some output to stderr while running ls on %s via ssh: %s",
+                path, err
+            )
     except CommandError as e:
         if "No such file or directory" in e.stderr and \
                 path in e.stderr:
             return None
         else:
             raise  # It's an unexpected failure here
-    return [
-        o for o in out.split(os.linesep) if o not in ('.', '..', '')
-    ]
+    return [l for l in out.split(os.linesep) if l]
 
 
 @build_doc
--- a/datalad/distribution/create_sibling.py
+++ b/datalad/distribution/create_sibling.py
@@ -19,6 +19,9 @@ from os.path import join as opj, relpath, normpath, dirname, curdir
 
 import datalad
 from datalad import ssh_manager
+
+from datalad.ui import ui
+
 from datalad.cmd import CommandError
 from datalad.consts import WEB_HTML_DIR, WEB_META_LOG
 from datalad.consts import TIMESTAMP_FMT
@@ -164,8 +167,6 @@ def _create_dataset_sibling(
                                  + " It is %sa git repository and has %d files/dirs." % (
                                      "" if has_git else "not ", len(path_children)
                                  )
-                    from datalad.ui import ui
-
                     if ui.is_interactive:
                         remove = ui.yesno(
                             "Do you really want to remove it?",
--- a/datalad/distribution/create_sibling.py
+++ b/datalad/distribution/create_sibling.py
@@ -15,7 +15,13 @@ from distutils.version import LooseVersion
 from glob import glob
 import logging
 import os
-from os.path import join as opj, relpath, normpath, dirname, curdir
+from os.path import (
+    curdir,
+    dirname,
+    join as opj,
+    normpath,
+    relpath,
+)
 
 import datalad
 from datalad import ssh_manager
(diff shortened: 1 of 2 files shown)


datalad/distribution/create_sibling.py
```diff
@@ -287,6 +302,21 @@ def _create_dataset_sibling(
     return remoteds_path
 
 
+def _ls_remote_path(ssh, path):
+    try:
+        # yoh tried ls on mac
```
yarikoptic: and a dd-wrt router -- both had `ls -a1`

datalad/distribution/create_sibling.py
```diff
@@ -287,6 +302,21 @@ def _create_dataset_sibling(
     return remoteds_path
 
 
+def _ls_remote_path(ssh, path):
```
yarikoptic: candidate for commands abstractions: https://github.com/datalad/datalad/pull/4068/files#r379009046

bpoldrack: True. Will integrate (I'm on it anyway)

datalad/distribution/create_sibling.py
```diff
@@ -287,6 +302,21 @@ def _create_dataset_sibling(
...
+    try:
+        # yoh tried ls on mac
+        out, err = ssh("ls -a1 {}".format(sh_quote(path)))
```
bpoldrack: Why not just `-A`, eliminating the need to filter out `.` and `..` ourselves?

datalad/distribution/create_sibling.py
```diff
@@ -163,6 +157,27 @@ def _create_dataset_sibling(
                 lgr.info(_msg + " Skipping")
                 return
             elif existing == 'replace':
+                if path_children:
+                    has_git = '.git' in path_children
+                    _msg_stats = _msg \
```
bpoldrack: I'd prefer a message that states the actual path it is about to remove rather than calling it "it". Particularly with urls (including replacements `insteadOf`) it might not be clear what's the thing we resolved it to.

datalad/distribution/create_sibling.py
yarikoptic: good idea! will do

datalad/distribution/create_sibling.py
yarikoptic: I changed it, but then realized -- path is provided within `_msg`, so no need to duplicate, thus "It" is to stay ;)

bpoldrack: Ah, true. Sorry.

yarikoptic:
ok, pushed c55d702cbacdad7805069a74efb0388f4c802fdb where I extend doc for `--existing` to describe possible dangers and replacing exception in non-interactive mode with a warning.

This option is indeed as dangerous as `rm`. Comments in the test (not made by me BTW) equate it to "force"ful operation, which it is.  It should be used with care, as `rm`.  For me in `zsh` (might be my settings), `zsh` would request confirmation for multi files removal if interactive, and would do it silently in a "batch" mode:
```shell
$> mkdir -p dir; touch dir/{1,2,3}; rm dir/*
zsh: sure you want to delete all 3 files in /tmp/dir [yn]? y
$> zsh -c 'mkdir -p dir; touch dir/{1,2,3}; rm dir/*'
$>
```
Given that so far nobody shot themselves in the foot, dangers would be properly documented/described,  $HOME removal so far has not happened I think this should suffice, and breeding double-force option would not really provide an additional safety blanket.  Once again -- the `--existing=replace` shouldn't be used blindly.

yarikoptic:
Pushed a fix and a test for interactive mode. AFAIK this is a viable candidate for the merge.

yarikoptic:
@mih, this PR awaits your feedback.

datalad/distribution/create_sibling.py
```diff
@@ -163,6 +157,28 @@ def _create_dataset_sibling(
...
                 return
             elif existing == 'replace':
+                remove = False
+                if path_children:
+                    has_git = '.git' in path_children
+                    _msg_stats = _msg \
+                                 + " It is %sa git repository and has %d files/dirs." % (
+                                     "" if has_git else "not ", len(path_children)
+                                 )
+                    from datalad.ui import ui
```
kyleam: It's fine to leave this delayed import, but it doesn't actually help in this case:

```sh
#!/bin/sh

python <<\EOF | grep datalad | grep ui
import sys
import datalad.distribution.create_sibling
for k in sorted(sys.modules.keys()):
    print(k)
EOF
```

```
datalad.ui
datalad.ui.base
datalad.ui.dialog
```

datalad/distribution/create_sibling.py
```diff
@@ -163,6 +157,28 @@ def _create_dataset_sibling(
...
+
+                    if ui.is_interactive:
+                        remove = ui.yesno(
+                            "Do you really want to remove it?",
+                            title=_msg_stats,
+                            default=False
+                        )
+                    else:
+                        remove = True
```
kyleam: Hmm, this is my main concern with this PR.  To me it seems like it'd be better to just make the replace-functionality interactive-only.

datalad/distribution/create_sibling.py
yarikoptic: ok, will move up

datalad/distribution/create_sibling.py
yarikoptic: ok, will do. better be safe than sorry they say ;-)

mih:
Cool! Works for me, thx!

PR #4156, opened 2020-02-16
RF: GitRepo.commit() based on WitlessRunner

mih:
Previous (positive) API and behavior are maintained, except for the lack
of 'protocol' support which is only available in the traditional runner.

In addition, when multiple calls to git-commit need to be made, --amend
is now used from the second call onward to avoid leaving multiple
commits behind for technical reasons, when conceptually there is only
as single changeset. Fixes gh-3914

This is a trimmed down changeset from a previous attempt to enable
interactive commit messages. Compare to
aa86a8d0df2af95399418f771f592aedbb848df1 to recover the discarded diff.

Replaces #4121 for the time being...

- [x] waiting for #4158

### DIFF
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -63,6 +63,7 @@ from datalad.cmd import (
     GitRunner,
     BatchedCommand,
     run_gitcommand_on_file_list_chunks,
+    StdOutErrCapture,
 )
 from datalad.config import (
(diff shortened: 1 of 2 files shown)


mih:
Re test failures: 
- Appveyor is coverage submission
- github ci windows is some incompatibility with shortpaths in the ancient pre 0.12 save/add code

PR #4161, opened 2020-02-18
BF: Reload datalad.cfg for test environment

### LINKED ISSUE #4153: temp. $HOME in test setup conflicts with datalad.cfg
Stumbeld upon this in PR #4124 :

We shoot our feet with the fake $HOME we set up for tests:
```
(Pdb) cfg.__dict__
{'_store': {'user.name': 'Benjamin Poldrack', 'user.email': 'benjaminpoldrack@gmail.com', 'url.ria+file:///tmp/datalad_temp__test_create_storele6bc8ij.insteadof': 'ria+ssh://test-store:', 'datalad.datasets.topurl': 'http://datasets-tests.datalad.org/', 'datalad.log.level': '100'}, '_cfgfiles': {'/home/ben/.gitconfig'}, '_cfgmtimes': {'/home/ben/.gitconfig': 1581683659.070651}, '_dataset_path': None, '_dataset_cfgfname': None, '_repo_cfgfname': None, '_config_cmd': ['git', '--git-dir=', 'config'], 'overrides': {}, '_src_mode': 'any', '_runner': <datalad.cmd.GitRunner object at 0x7f86b2795520>, '_gitconfig_has_showorgin': True}
```

Note, that it reads `/home/ben/.gitconfig`.
But:

```
(Pdb) os.environ['HOME']
'/tmp/datalad_temp_ft_s2hz0'
```

Since `datalad.cfg` is instantiated on import, it doesn't consider what `setup_package` does to $HOME and won't reconsider what global config to read from. This leads to inconsistencies in test configs.

I'd suggest to

```python
global cfg
cfg.reload(force=True)
```
in `setup_package` after changing home dir. What do you think, @datalad/developers ? Anything I miss here?

bpoldrack:
We change /home/ben from within setup_package, therefore we
need to consider the correct global git config file. This
used to be wrong, since datalad.cfg is instantiated on import
hence before setup_package is executed.
(Closes #4153)

### DIFF
--- a/datalad/__init__.py
+++ b/datalad/__init__.py
@@ -130,6 +130,11 @@ def setup_package():
 """)
         _TEMP_PATHS_GENERATED.append(new_home)
 
+    # Re-load ConfigManager, since otherwise it won't consider global config
+    # from new $HOME (see gh-4153
+    global cfg
+    cfg.reload(force=True)
+
     # To overcome pybuild by default defining http{,s}_proxy we would need
     # to define them to e.g. empty value so it wouldn't bother touching them.
     # But then haskell libraries do not digest empty value nicely, so we just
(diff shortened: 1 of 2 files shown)


datalad/__init__.py
```diff
@@ -130,6 +130,11 @@ def setup_package():
 """)
         _TEMP_PATHS_GENERATED.append(new_home)
 
+    # Re-load ConfigManager, since otherwise it won't consider global config
+    # from new $HOME (see gh-4153
+    global cfg
```
mih: Why `global`? There is only a method call, no assignment.

bpoldrack: Just being explicit. Doesn't do any harm, does it?

yarikoptic: I would also vote to remove it until we need it -- other variable(s) in the function do not use `global`

bpoldrack: Done.

Issue #4149, opened 2020-02-14
Improve Gdrive remote (any rclone-supported type) UX


mih:
At the moment this is already doable, but one has to reach relatively deep into the guts: http://handbook.datalad.org/en/latest/basics/101-138-sharethirdparty.html#setting-up-3rd-party-services-to-host-your-data

Initial concept of an approach

- extend `siblings` to also deal with special remote configuration, by adding a `--type` parameter that is `git-remote` by default to maintain current behavior
- by able to drive an interactive `rclone config` and use the result to automatically compose the matching `git annex initremote` command.

adswa:
Further notes:

I found this piece of information in [git-annex-remote-gdrive](https://github.com/Lykos153/git-annex-remote-googledrive#google-drive-api-lockdown). Not sure how that could affect the API keys we generated.

> Google has started to lockdown their Google Drive API in order to enhance security controls for the user. Developers are urged to "move to a per-file user consent model, allowing users to more precisely determine what files an app is allowed to access". Unfortunately they do not provide a way for a user to allow access to a specific folder, so git-annex-remote-googledrive still needs access to the entire Drive in order to function properly. This makes it necessary to get it verified by Google. Until the application is approved (IF it is approved), the OAuth consent screen will show a warning (#31) which the user needs to accept in order to proceed.

> It is not yet clear what will happen in case the application is not approved. The warning screen might be all. But it's also possible that git-annex-remote-googledrive is banned from accessing Google Drive in the beginning of 2020. If you want to prepare for this, it might be a good idea to look for a different cloud service. However, it seems that rclone got approved, so you'll be able to switch to git-annex-remote-rclone in case git-annex-remote-googledrive is banned. To do this, follow the steps described in its README, then type git annex enableremote <remote_name> externaltype=rclone rclone_layout=nodir. This will not work for export-remotes, however, as git-annex-remote-rclone doesn't support them.

yarikoptic:
+100 on that!
Note also, that recently (7.20200202) git-annex gained ability to list options for the special remotes:

```
$> git annex initremote myrsync type=rsync --whatelse
shellescape
	avoid usual shell escaping (not recommended)
	(yes or no)
rsyncurl
	(required) url or hostname:/directory for rsync to use
chunk
	size of chunks (eg, 1MiB)
encryption
	how to encrypt data stored in the special remote
	(hybrid or none or pubkey or shared or sharedpubkey)
embedcreds
	embed credentials into git repository
	(yes or no)
[...]

PR #4162, opened 2020-02-18
TMP/NF: Draft a wrapper around rclone

adswa:
ping @mih.
Here is what I came up with so far for #4149:

- I gave ``siblings`` a ``-t/--type`` argument. The default is ``git-remote``, but at the moment it also takes ``drive`` or ``any``
- If anything but ``git-remote`` is specified, an rclone-based special remote is configured.
    - If ``drive`` is given, a special remote to Google drive is configured (with no further arguments, the default configuration is used, but the optional arguments ``apikey`` and ``apisecret`` can be given to specify ``client_id`` and ``client_secret``. 
   - If ``--type any`` is given  or if ``-i/--interactive`` is specified, the interactive ``rclone configure`` command is started.
- After configuration, the special remote is initialized.
   - Without additional arguments, the following ``git annex init`` call is used: ``git annex init <remote-name> encryption=none externaltype='rclone' chunk=50MiB type=external prefix=datalad/<ds-ID> rclone_layout=lower target=<remote-name>``
   - Alternatively, ``encryption``, ``layout``, ``chunk``, ``prefix`` can also be given as arguments to the command line call

Currently, I can do something like this to configure a special remote called "mygdrive" as an rclone based Gdrive special remote with my Google API key:

```
datalad siblings add --type 'drive' -s mygdrive --apikey '774309559379-sapf1isni06pfnoahes23ldsm08jaa8qx.apps.googleusercontent.com' --apisecret '9l7iR5a2u2lHHocfNs862fPx'
```

It would be cool if you could check whether the general approach makes sense. I'm opening it as a draft because there is still a lot missing:

- I don't know how to write tests for an interactive process
- I don't know yet how to do result handling/reporting in the end. After successful command completion I would also like to provide an example command as help on how to configure the special remote as a publication dependency
- It's inconvenient that when started interactively, ``-s/--name`` given to the command and the name given in the interactive configuration need to match. Else, it fails with a not very helpful error
- This currently only does the configuration for the ``publishing`` part of sharing a dataset, but it doesn't take care of the "receiving end", i.e. configuration a special remote and enabling it for user who want to obtain the dataset contents from a clone
- Documentation

### DIFF
--- a/datalad/distribution/siblings.py
+++ b/datalad/distribution/siblings.py
@@ -18,6 +18,10 @@ import os.path as op
 
 from urllib.parse import urlparse
 
+from datalad.cmd import (
+    WitlessRunner,
+)
+
 from datalad.interface.base import Interface
 from datalad.interface.utils import eval_results
 from datalad.interface.base import build_doc
@@ -82,6 +86,10 @@ def _mangle_urls(url, ds_name):
     if not url:
         return url
     return url.replace("%NAME", ds_name.replace("/", "-"))
+
+
+#global list of supported special remotes
+special_remotes = ['drive', 'any']
 
 
 @build_doc
@@ -190,7 +198,60 @@ class Siblings(Interface):
             doc="""Whether to query all information about the annex configurations
             of siblings. Can be disabled if speed is a concern"""),
         recursive=recursion_flag,
-        recursion_limit=recursion_limit)
+        recursion_limit=recursion_limit,
+        remotetype=Parameter(
+            args=('-t', '--type',),
+            dest='remotetype',
+            doc="""Configure a remote of your choice. Default is 'git-remote'.
+                Alternatively, configure special remotes to third party services
+                via rclone. Note: requires rclone and git-annex-remote-rclone.
+                Currently available: 'drive' (Google drive), 'any' (get an
+                interactive prompt)"""),
+        encryption=Parameter(
+            args=('--encryption',),
+            dest='encryption',
+            doc="""If a special remote to a third party service (--type/-t) should
+                be inititialized, which encryption should be used? Default: none.
+                See https://git-annex.branchable.com/encryption/ for more info.
+                """),
+        interactive=Parameter(
+            args=('-i, --interactive',),
+            dest='interactive',
+            action="store_true",
+            doc="""If given, configure rclone special remote interactively.
+                """),
+        chunk=Parameter(
+            args=('--chunk',),
+            dest='chunk',
+            doc="""If a special remote to a third party service (--type/-t) should
+                be inititialized, which chunksize should be used? Default: 50MiB.
+                """),
+        prefix = Parameter(
+            args=('--prefix',),
+            dest='prefix',
+            doc="""If a special remote to a third party service (--type/-t) should
+                be inititialized, which prefix should be used in the directory
+                tree name at the third party remote? Default: dataset ID
+                """),
+        layout=Parameter(
+            args=('--layout',),
+            dest='layout',
+            doc="""If a special remote to a third party service (--type/-t) should
+            be inititialized, which layout should be used in the directory
+            tree name at the third party remote? Default: lower.
+            """),
+        apikey=Parameter(
+            args=('--apikey',),
+            dest='apikey',
+            doc="""Optional API key for third party service if a special remote
+            is created. Default: None"""),
+        apisecret=Parameter(
+            args=('--apisecret',),
+            dest='apisecret',
+            doc="""Optional API secret for a third party service if a special
+            remote is created. Default: None.
+        """),
+    )
 
     @staticmethod
     @datasetmethod(name='siblings')
@@ -214,7 +275,16 @@ class Siblings(Interface):
             inherit=False,
             get_annex_info=True,
             recursive=False,
-            recursion_limit=None):
+            recursion_limit=None,
+            remotetype='git-remote',
+            interactive=False,
+            encryption='none',
+            chunk='50MiB',
+            prefix=None,
+            layout='lower',
+            apikey=None,
+            apisecret=None,
+    ):
 
         # TODO: Detect malformed URL and fail?
         # XXX possibly fail if fetch is False and as_common_datasrc
@@ -243,6 +313,28 @@ class Siblings(Interface):
 
         res_kwargs = dict(refds=refds_path, logger=lgr)
 
+        # setup for special remote creation and initialization
+        if remotetype != 'git-remote':
+            if not prefix:
+                prefix = 'datalad/' + str(dataset.repo.config.get('datalad.dataset.id', None))
+        # dict with options for inititialization of a special remote
+        init_kwargs = dict(encryption=encryption,
+                           externaltype='rclone',
+                           chunk=chunk,
+                           type='external',
+                           prefix=prefix,
+                           rclone_layout=layout,
+                           )
+        # dict with options for configuration (creation) of an rclone special
+        # remote
+        if remotetype == 'any':
+            print("meeep")
+            # if no specifically implemented special remote is given, just
+            # get into the interactive session later
+            interactive = True
+        config_kwargs = dict(interactive=interactive,
+                             API_key=apikey,
+                             API_secret=apisecret)
         ds_name = op.basename(dataset.path)
 
         # do not form single list of datasets (with recursion results) to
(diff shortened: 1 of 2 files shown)


mih:
I have implemented https://github.com/datalad/git-remote-rclone so that we need not limit us to use rclone remotes as special remotes for annex, but can simultaneously also utilize them as normal Git remotes. The remote helper already puts its data in a dedicated directory per repository, hence it is straightforward to place an annex in it too.

So if we have the name of an rclone config, we can `initremote` an annex, and also `remote add` the same location.

mih:
So, I have been thinking about this for a couple of days now. I am not sure if I have settled on an opinion yet, but here is a dump of my thought and an interim conclusion.

- things are complicated and rather heterogeneous across rclone's supported remotes 
- we likely do not want a specific setup command for each remote type, but between dropbox and gdrive alone the procedures are different enough to warrant dedicated support to make things convenient.
- while a gdrive or a dropbox "already exist", and could conceptually be just "added as a sibling", in practice that would be rather inconvenient for anything more complex than a single dataset with no subdatasets
- efficient sharing of multiple disjoint collections from a single storage account (a common scenario IMHO) requires the concept of a "store" (in the sense that we use for the RIA code. Each store (for multiple datasets) represents a unit that is shared (i.e. identical access permissions). Hence we need a command that can setup/fill multiple stores conveniently, and does so with maximum homogeneity across services.
- much logic is shared with the RIA store concept (tandem of git-remote and annex special remote; need to reconfigure a special remote setup based on how a store looks like after cloning from it, i.e. identify the associated special remote setup via UUID and reconfigure it automatically)

### Proposal

- take this changeset and use it also as a starting point of a `create-sibling-rclone` command
- the command should be similar to `create-sibling-ria`, in particular use an `rclone://<remotecfg>/path-to-store-within-remote` type URL as its main argument
- when it encounters an unknown remote label (check against `rclone listremotes`), it should start `rclone config create` with sensible defaults for a given remote `--type`, and the main `rclone://` URL argument. This is superior to always directing people to the fully interactive rclone config runs, because we can make sensible decisions for defaults. This will not prevent any customizations, users can always configure rclone however they see fit, before running `create-sibling-rclone`.
- the setup customizations that `clone` performs for `ria+http|ssh://` sources also need to be matched for rclone (sort out which special remote belongs to the source, and reenable it with a matching configuration: prefix, rclone config)

### the fate of `siblings()`

ATM we have three ways in which a dataset becomes aware of a sibling

- it is cloned from it (`clone`)
- it has new siblings created for it (`create-sibling*`)
- it has an existing sibling registered (`siblings add`)

The first two are quite flexible and sometimes even convenient. The latter is a rather low-level tool that has need been designed with this kind of flexibility in mind, and primarily focuses on Git remotes.
It would be best, IMHO, if we come up with a superior version of `siblings` (along the lines of this draft PR) that we can use in both `clone` and `create-sibling` to perform whatever configuration is necessary for a particular remote type. API-wise it doesn't look that bad even now IMHO. However, we should strive to not have stuff like `--apikey` and `--api-secret` and leave that to special purpose commands,or the native tools (like rclone).

My impulse idea would be to modularize `siblings()` further and RF all functionality into dedicated modules for each remote-type (git, rclone, ria, ...), each module would provide functions for the supported actions: 'query', 'add', 'remove', 'configure', 'enable' (although I think that configure could also do what enable does).

yarikoptic:
**Proposal** sounds good to me.  But as for the "fate", I would like to bring up following points:

### Layouts

I feel that any "unification" attempt could also benefit from adding an explicit option for a layout of dataset(s) on remote end.  We so far (conceptuatlly) have

- `hierarchical` -   the default one for `create-sibling -r` (used e.g. for datasets.datalad.org)
- `flattenned` - supported by `create-sibling -r` with `%RELNAME` placeholder where "any slashes replaced by dashes" (I think I might have used for some datasets published on github, can't recall)
- `ria-store` - supported by `create-sibling-ria` (used for http://store.datalad.org and in Germany ;)). Note: [as I have mentioned before](https://github.com/datalad/datalad/pull/4022#discussion_r367438449) "ria" in the name has nothing to do with what "ria" abbreviation stands for, so name might like to be changed

`hierarhical` and `flattenned` are conceptually different from `ria-store` in that they reuse local dataset path **name**, while `ria-remote` - dataset **UUID**.  I think it might be worthwhile also considering 

- `flat-ria-store` - which is ria but without "caching levels", so using dataset UUIDs in a flat list.  Could facilitate establishing "data stores" on infrastructure without support of multiple levels in the hierarchy (e.g., github, which could be legit usecase with LFS annex special remote).

There is also an implied feature of setups - `bare` vs `non-bare`.  `ria-store` ATM implies `bare` (but they could be non-bare in the ria hierarchy), while `hierarchical` implies `non-bare`.

Sure thing not every type of a remote would support all possible layouts, but I feel that having them explicitly listed as supported for any given type of remote should allow not only to make `create-sibling` more flexible, and clear on what to expect as a result, but also make code less "remote specific".  In summary - I think layout might become an option for "configure" interface.

#### Shared logic

>  much logic is shared with the RIA store concept (tandem of git-remote and annex special remote...)

as pointed out,  the "tandem" seems to be not really "ria" specific, but might relate to the "hierarchy" aspect.

### Credentials

> --apikey and --api-secret and leave that to special purpose commands,or the native tools (like rclone).

on one hand, leaving this to native tools should be most robust to datalad developers, but might be inconvenient to the datalad users, especially if we manage to provide those tools with all necessary details (even if for some common use cases) without making users interact with them.  Then, it might be that using (and improving if necessary) datalad credentials handling and "proxying" it to underlying users would provide most consistent interface (to the users).

mih:
> Sure thing not every type of a remote would support all possible layouts, but I feel that having them explicitly listed as supported for any given type of remote should allow not only to make create-sibling more flexible.

I was talking about `siblings` not `create-sibling*`. The latter are easy, because they can implement whatever is possible for a given remote. The tricky use case is to assign/enable matching siblings (not just remotes, but special remotes too) for a local dataset hierarchy that already exists (i.e. no `clone`), and where the siblings also already exist (i.e. no `create-sibling`).

Having a bunch of possible layouts centrally supported sounds doable. Intuitively, I would implement a function that gives back the relative location of a dataset in a "store" (which could also be underneath a root dataset) given a submodule record (which nowadays has all we need: path, name, UUID, branch).

The Q for me is though, if `siblings` can do all this, why have `create-sibling*` commands at all?

> Then, it might be that using (and improving if necessary) datalad credentials handling and "proxying" it to underlying users would provide most consistent interface (to the users).

Hard to argue against that. But how would that actually look like? Every argument adds to the complexity of a command, and combinatorial interpretation issues arise quickly. We already have:

- `--apikey`
- `--apisecret`
- `--user`(in various flavors)
- `--password`
- `--github-organization`

We presently defer to SSH, `gitlab` module, `git hub` token management, and soon `rclone` to actually handle authentication in a platform-appropriate fashion. I don't even have an approximate idea how we would "proxy" to any and all of them in a way that doesn't suck. A related aspect is the frequency with which authorization (obtain/enter) vs authentication (use credentials) occurs. I rarely do the former, and the latter happens in the background all the time. Making both (somewhat invisible) aspects a prominent part of the API seems suboptimal.

nicolocin:
Hi - our lab is looking into using `datalad` to manage data on our HPC, part of which synced to a shared Google Drive using a `rclone` singularity container.  Is it possible to use `datalad` to link files on the cluster and Google Drive?  We have `git-annex` and `datalad` on the cluster but not in the container.

Also is `https://github.com/DanielDent/git-annex-remote-rclone` still the one we should use?  It's the one used in the tutorial [here](http://handbook.datalad.org/en/latest/basics/101-138-sharethirdparty.html)

kyleam:
@nicolocin

[ sorry for the slow response ]

> Hi - our lab is looking into using `datalad` to manage data on our HPC, part of which synced to a shared Google Drive using a `rclone` singularity container. Is it possible to use `datalad` to link files on the cluster and Google Drive? We have `git-annex` and `datalad` on the cluster but not in the container.

I don't know for sure, but it sounds like that should be possible.  I think the main thing is to get git-annex-remote-rclone working with the rclone singularity container.  Based on looking at the install instructions for git-annex-remote-rclone, it seems to rely on rclone being in `PATH`, so the container would probably need to be included as an executable with an appropriate runscript.

If the above ends up being problematic, rclone appears to be available via conda, so that might be an alternative to the singularity container if conda is available on the cluster.

Once the special remote is functioning with git-annex, datalad should work with it fine.

> Also is `https://github.com/DanielDent/git-annex-remote-rclone` still the one we should use? It's the one used in the tutorial [here](http://handbook.datalad.org/en/latest/basics/101-138-sharethirdparty.html)

Yes.  This draft PR/conversation touches a few different topics, but as far as annex data is concerned, I don't think there is any proposal to move away from using the git-annex-remote-rclone special remote underneath.

I haven't used rclone/git-annex-remote-rclone, so hopefully others with direct experience will point out anything above that's incorrect.  Also, feel free to open a dedicated issue if you run into problems.

mih:
Last push >6 month ago. Closing do to inactivity.

PR #4158, opened 2020-02-17
Removal of obsolete add/save (2nd attempt)

mih:
This revitalizes https://github.com/datalad/datalad/pull/3735 but this time with the goal of getting it merged soon. It is critical for #3368 and will likely involve a relatively elaborate RF.

- [x] core tests pass
- [x] benchmarks cleaned up
- [x] -neuroimaging tests pass https://github.com/datalad/datalad-neuroimaging/pull/80
- [x] -container tests pass https://github.com/datalad/datalad-container/pull/116
- [x] -crawler tests pass https://github.com/datalad/datalad-crawler/pull/70

Tests will go green with #4163 

Related:

- #3714
- #4156

PR #4163, opened 2020-02-19
TST: Make GitHub CI test master of extensions and not releases.

mih:
Not every change in an unreleased master here, should require an extension release.

### DIFF
--- a/.github/workflows/test_extensions.yml
+++ b/.github/workflows/test_extensions.yml
@@ -11,10 +11,10 @@ jobs:
       fail-fast: false
       matrix:
         extension: [
-            datalad_neuroimaging,
-            datalad_container,
-            datalad_metalad,
-            datalad_crawler,
+            datalad-neuroimaging,
+            datalad-container,
+            datalad-metalad,
+            datalad-crawler,
         ]
     steps:
     - name: Set up system
@@ -39,15 +39,14 @@ jobs:
         pip install -r requirements.txt
     - name: Install ${{ matrix.extension }} extension
       run: |
-        pip install ${{ matrix.extension }}
-      if: matrix.extension != 'datalad_crawler'
-    - name: Install ${{ matrix.extension }} extension
+        pip install https://github.com/datalad/${{ matrix.extension }}/archive/master.zip
+    - name: Install additional dependencies
       run: |
-        pip install ${{ matrix.extension }}[devel]
-      if: matrix.extension == 'datalad_crawler'
+        pip install mock
+      if: matrix.extension == 'datalad-crawler'
     - name: Install Singularity
       run: sudo eatmydata apt-get install singularity-container
-      if: matrix.extension == 'datalad_container'
+      if: matrix.extension == 'datalad-container'
     - name: WTF!?
       run: |
         datalad wtf
@@ -55,4 +54,4 @@ jobs:
       run: |
         mkdir -p __testhome__
         cd __testhome__
-        python -m nose -s -v --with-cov --cover-package datalad ${{ matrix.extension }}
+        python -m nose -s -v --with-cov --cover-package datalad $(echo ${{ matrix.extension }} | tr '-' '_')


mih:
Windows test failure is a fluke, there was no code change.

kyleam:
> Not every change in an unreleased master here, should require an extension release.

The problem is that now we don't have a test run to tell us that a PR in core is breaking compatibility with a released extension.  That's an important thing to know because, in order for the extension to work with the next release, we need to either

  * add a compatibility kludge and make a release in the extension
  * bump the required version to what will be the next datalad release with the breaking feature and then synchronize the releases of core and the extension.  (Due to the timings and incompatibilities, this one is more painful.)

mih:
> The problem is that now we don't have a test run to tell us that a PR in core is breaking compatibility with a released extension.

Is that true, though? A PR that breaks an extension will have a broken extension test run (on its master). As a response, a change in that extension needs to be made. That change should prompt an eventual release. So the information is not lost, but more transient when only looking at datalad-core. We can of course have both tests, but that is only needed, if we encourage people to combine released versions of extensions with unreleased versions of datalad-core. If we only supported combinations of released versions, we just need to go through the extensions (trigger their own tests in their repos, immediately prior a release of core).

kyleam:
> > The problem is that now we don't have a test run to tell us that a PR in core is breaking compatibility with a released extension.
> 
> Is that true, though? A PR that breaks an extension will have a broken extension test run (on its master).

As far as I know, PRs for extensions run against the released version of datalad.  So, we'll eventually see breakage once datalad is released, but then, at a potentially much later time, someone has to figure out what change in core broke the functionality.  And during this time, users of the extension that update datalad see breakage in the wild.

So, yes, I think it's effectively true that now we won't be able to detect that a PR in core breaks functionality until it has already made it into a release and is broken in the wild.

We could set up a run for the extensions against core's master.  That makes us catch the issue sooner, but in my view it's still too removed from the PR.  If an incompatible change is being made in a PR, it is good to know _before_ it is merged and deal with it then.  (Not to mention that seeing that a PR is breaking an extension should be an indicator to think more closely about whether third-party packages will be affected.)  Then the PR has the choice of

  (1) adding a compatibility kludge to core before the merge of the PR, at some point updating the extension to not need the kludge and releasing the extension, and then later removing the kludge in core

  (2) dealing with it in the extension and disregarding the failing extension run in core for the time being.

None of that's particularly pleasant, but I think it's an inherent to having separate packages and should be dealt with at the time of the PR.

> As a response, a change in that extension needs to be made. That change should prompt an eventual release. So the information is not lost, but more transient when only looking at datalad-core
> [...] we just need to go through the extensions (trigger their own tests in their repos, immediately prior a release of core).

As I said above, the direct connection to the PR is lost.  It then needs to be dug out after the fact rather than dealt with as the PR is under consideration.

mih:
Sorry to be a pain @kyleam, but I must be missing something, because your rational makes perfect sense, but doesn't seem to apply to the matter as I see it. I made this PR in response to #4158 -- which broke extensions. There is a linked PR in each extension that fixes it. Each of these PRs got merged, reducing the TODO for each extension to an eventual release. In this particular case, the timing of these releases only has the constraint that it must come prior to the next datalad-core release. This PR makes the extension test not fail for the time between now and an extension release. I think this is good, because it saves the devs that do not care about a particular extension to double-check every single time whether any breakage is part of a set of accumulated known breakages, or a new one. With this PR merged, any breakage reported in a PR is likely caused by that PR (minus the unfortunate tests that behave unreliable).

> As far as I know, PRs for extensions run against the released version of datalad. So, we'll eventually see breakage once datalad is released, but then, at a potentially much later time, someone has to figure out what change in core broke the functionality. And during this time, users of the extension that update datalad see breakage in the wild.

So this premise doesn't seem to be valid to me: In #4158 we saw the breakage, and we would still see it with this PR. The fixes clearly identify the source of the breakage. And any users that require the latest versions (or even unreleased code), can update to a consistent set of packages and should be fine.

Where do I have it wrong?

kyleam:
Reading over this again, I find my argument to be a mix of confused and unconvincing.  While it's true that we now don't have a test run against the released version, the testing against the extension's master should catch most the incompatibilities with the released extension, and I don't think it's worth worrying about the small subset of potential incompatibilities in code that happened to change between the last extension release and its master.

The remaining concern is that an extension may not end up being released in step with a core release.  We're relying on the release cutter on datalad's end to remember to do that because AFAICS that wouldn't be caught by any tests.  If we forget, it's just a matter of making a release (not "digging after the fact" as I was saying above) because we know the extension's master is in a passing state with the latest core release, but with the current setup it'd take a user (hopefully just one of us) hitting into this to realize the mistake.

> Where do I have it wrong?

Paying me any mind :]

yarikoptic:
> While it's true that we now don't have a test run against the released version, the testing against the extension's master should catch most the incompatibilities with the released extension, and I don't think it's worth worrying about the small subset of potential incompatibilities in code that happened to change between the last extension release and its master.

I would say that such logic is "developer-oriented", and it is Ok for "development CI". But I think we should (re)establish per-extension (or some centralized) CI runs which would provide status of extension releases "readiness", so that at least before the release of core we could take care about releasing them if necessary so users do not run into incompatibilities among released versions of datalad and extensions.  I will look into establishing one in https://github.com/datalad/datalad-extensions 
  
Such breaking releases should also be accompanied with explicit "Breaks" statements in Debian packaging (i.e. "datalad" package for upcoming release would declare `Breaks: datalad-container (<< 0.5.1)`. Unfortunately neither pypi/distuils nor conda AFAIK has any similar way to declare such breakages ATM.

kyleam:
>  I will look into establishing one in https://github.com/datalad/datalad-extensions

Great, thanks.

PR #4164, opened 2020-02-19
ENH: Move git-annex-ria-remote to datalad core as git-annex-ora-remote

bpoldrack:
Integrate RIA special remote and `ria-export-archive` command in datalad core.

- [x] move entire code base of https://github.com/datalad/git-annex-ria-remote to datalad core
- [ ] adapt some tests/utils to make use of what we have already
- [x] add test for binary data
- [x] change special remote config to be based on bare repo's config rather than actual user configs (idea appeared in context of PR #4124); this will need additional adapting `clone`, but in another PR.
- [ ] have a progressbar for transfer via special remote (done for get over SSH - note, that it's not actually implementing a progress *bar*, but reporting progress to annex, of course. It's a special remote.) More of that will have to wait until we're done with more essential parts.

Edit/Update:

New Plan. We need the special remote in master urgently on our side. As long as we don't break anything in core, a merge will only lead to make it easier to deal with RIA. Some of the things I still need to do require proper RF'ing first. So, strongly vote for merge - I'll open follow-up PRs instantly, that can be used for review etc. There need to be at least the following things to address in next PRs:

- RF to move the creation of the remote end (may include a separation of store creation from dataset creation)
- Find a spot for docs and update them (last step of moving into core - git-annex-ria-remote repo can be archived/deleted/ignored after that)
- Rework tests. Partially outdated and otherwise not nicely structured ATM.
- Move remote execution to the command abstraction classes

### DIFF
--- a/datalad/customremotes/ria_remote.py
+++ b/datalad/customremotes/ria_remote.py
@@ -1,0 +1,923 @@
+from annexremote import SpecialRemote
+from annexremote import RemoteError
+
+from pathlib import (
+    Path,
+)
+import shutil
+from shlex import quote as sh_quote
+import subprocess
+import logging
+from functools import wraps
+from datalad.distributed.ria_utils import (
+    get_layout_locations,
+    verify_ria_url,
+)
+
+lgr = logging.getLogger('ria_remote')
+
+# TODO
+# - make archive check optional
+
+
+def _get_gitcfg(gitdir, key, cfgargs=None, regex=False):
+    cmd = [
+        'git',
+        '--git-dir', gitdir,
+        'config',
+    ]
+    if cfgargs:
+        cmd += cfgargs
+
+    cmd += ['--get-regexp'] if regex else ['--get']
+    cmd += [key]
+    try:
+        return subprocess.check_output(
+            cmd,
+            # yield text
+            universal_newlines=True).strip()
+    except Exception:
+        lgr.debug(
+            "Failed to obtain config '%s' at %s",
+            key, gitdir,
+        )
+        return None
+
+
+def _get_datalad_id(gitdir):
+    """Attempt to determine a DataLad dataset ID for a given repo
+
+    Returns
+    -------
+    str or None
+      None in case no ID was found
+    """
+    dsid = _get_gitcfg(
+        gitdir, 'datalad.dataset.id', ['--blob', ':.datalad/config']
+    )
+    if dsid is None:
+        lgr.debug(
+            "Cannot determine a DataLad ID for repository: %s",
+            gitdir,
+        )
+    else:
+        dsid = dsid.strip()
+    return dsid
+
+
+class RemoteCommandFailedError(Exception):
+    pass
+
+
+class RIARemoteError(RemoteError):
+
+    def __init__(self, msg):
+        super().__init__(msg.replace('\n', '\\n'))
+
+
+class IOBase(object):
+    """Abstract class with the desired API for local/remote operations"""
+    def mkdir(self, path):
+        raise NotImplementedError
+
+    def put(self, src, dst):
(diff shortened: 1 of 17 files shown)


datalad/customremotes/ria_remote.py
```diff
@@ -0,0 +1,933 @@
...
+            # TODO: Theoretically terminate() can raise if not successful. How to deal with that?
+            self.shell.terminate()
+        self.sshmanager.close()
+
+    def _append_end_markers(self, cmd):
+        """Append end markers to remote command"""
+
+        return cmd + ' && echo "{}" || echo "{}"\n'.format(self.REMOTE_CMD_OK, self.REMOTE_CMD_FAIL)
+
+    def _get_download_size_from_key(self, key):
```
mih: How is this different from `AnnexRepo.get_size_from_key()`? It only says "doesn't fit", but now why or how. If annexrepo import is too expensive, it might be worth filing an issue that there is code duplication happening and that we should RF accordingly. `AnnexRepo.get_size_from_key()` is a static method, but could easily move to `datalad.utils` as a standalone function.

datalad/customremotes/ria_remote.py
```diff
@@ -0,0 +1,933 @@
...
+        if mode == 'w':
+            mode = ">"
+        elif mode == 'a':
+            mode = ">>"
+        else:
+            raise ValueError("Unknown mode '{}'".format(mode))
+        if not content.endswith('\n'):
+            content += '\n'
+
+        cmd = "echo \"{}\" {} {}".format(content, mode, str(file_path))
```
mih: Wouldn't this add newlines? `printf` might be a better choice.

datalad/customremotes/ria_remote.py
bpoldrack: This is outdated. I already had a PR fixing `AnnexRepo`'s method in preparation of the RF.

datalad/customremotes/ria_remote.py
bpoldrack: Added `-n` in PR #4203 already. Don't want to invest thoughts right now, since it is to be replaced by proper command abstraction in PR #4068 anyways.

bpoldrack:
> I left a few in-line comments. My main concern is that many pieces do not come with docstring, which will make a later (already known to be needed RF) harder than necessary.

Generally I agree, but in this case I'm on the follow-ups already. "Later" actually is right now, so it's not a real concern for me. Already working on updating and enhancing docs.

datalad/customremotes/ria_remote.py
mih: I think it is objective wrong here, and `echo -n` isn't well supported across platforms.

datalad/customremotes/ria_remote.py
mih: Does that mean the function is outdated and can be removed, or the comment is outdated and should be removed?

mih:
#4258 implements a related reduction in the overuse of the 'ria' label for different concepts and functionality.

mih:
I tried running the tests, but they fail due to the `--include-dotfiles` files issue. I tried again after rebasing this branch onto current master, but several tests still fail, including ones in `ria-basics`.

mih:
Root of the test failures seems to be that `name = self.annex.getconfig('name')` returns `None`. Not sure how that is possible.

mih:
I started to consolidate this changeset with #4258

datalad/customremotes/ria_remote.py
mih: Implemented in #4260

mih:
I will close this now, a replacement that also contains the changes of this PR is at #4260

PR #4168, opened 2020-02-20
CLN: Remove obsolete AnnexRepo.commit()

mih:
The GitRepo implementation should be capable of doing what is needed in
the rest of the codebase.

Removing it helps with gh-2970 and makes gh-4127 obsolete by fixing
gh-4126 too.


mih:
I do not see this Travis error locally:
```
FAIL: datalad.support.tests.test_annexrepo.test_commit_annex_commit_changed(True,)
1185----------------------------------------------------------------------
1186Traceback (most recent call last):
1187  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
1188    self.test(*self.arg)
1189  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/datalad/tests/utils.py", line 586, in newfunc
1190    return t(*(arg + (d,)), **kw)
1191  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/datalad/support/tests/test_annexrepo.py", line 2169, in check_commit_annex_commit_changed
1192    , untracked=['untracked']
1193  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/datalad/tests/utils.py", line 363, in ok_clean_git
1194    eq_(sorted(index_modified_), sorted(index_modified))
1195AssertionError: ['tobechanged-annex', 'tobechanged-git', 'willgetshort'] != ['tobechanged-annex', 'tobechanged-git']

```

kyleam:
> Instead of digging up the reasons, I have first RF'ed the test to use `Repo.save()` instead of commit -- which is what all top-level "save" code goes through.

The reason is in [this comment][0] in gh-4126, which points to 1a609d4 (BF: annexrepo: Avoid typechange when using temporary index, 2019-04-30) and 205b2ca (TST: rev-save: Demonstrate partial commit failure, 2019-04-30).

It fails for the reason that `AnnexRepo.commit` continued to use the temporary index after direct mode was gone.   The new `save` partially deals with this, but 205b2cac4 (TST: rev-save: Demonstrate partial commit failure, 2019-04-30) shows where `interface.save/AnnexRepo.commit` would pass where the new approach wouldn't.  Though none of this will be relevant once the minimum git-annex version is a version that auto-upgrades to v7+ repos.

[0]: https://github.com/datalad/datalad/issues/4126#issuecomment-584205783

mih:
Thanks for the clarification and the review @kyleam !

Issue #2879, opened 2018-09-27
Brain dump on GitPython removal


mih:
I am aware that this idea is unpopular, but I am yet to be convinced that carrying around a repository instance in an otherwise entirely stateless setup is adding something valuable enough to be worth the cost. However, this is not to debate this point but rather to summarize what I know about the situation.

### Usage patterns

[ km: Updated for #2902 and #3530 ]

Direct usage of the GitPython API in non-Repo class code is most problematic. Here are those in Datalad-core that I am aware of (`repo` is the GitPython repository object):

- [ ] repo.remotes[remote].refs (create-sibling, publish)
- [ ] repo.repo.commit() (publish)
- [ ] repo.remotes[remote].refs[remote_branch_name] (publish)
- [ ] repo.close() (ls)
- [X] repo.git.show() (gone with #2902)
- [ ] repo.git.merge_base() (GitRepo)
- [X] repo.git.rev_parse() (all but two gone with #2902)
- [X] repo.git_dir (gone with #2902)
- [ ] repo.git (GitRepo, AnnexRepo)
- [X] repo.git.for_each_ref()
- [X] repo.git.log()
- [X] repo.git.ls_tree()
- [X] repo.git.cat_file

and additionally in the associated tests

- [X] repo.git.rev_list()
- [X] repo.head.commit.message
- [x] repo.commit().hexsha
- [X] repo.git.reset()
- [ ] repo.submodule() (GitRepo tests)

### Data structure/object use

Except for PushInfo and Blob these are pretty harmless, as they either do not need a Repo instance, or are more or less static objects.

- [x] git.GitConfigParser
- [x] git.remote.PushInfo
- [x] git.objects.blob.Blob
- [ ] git.exc.InvalidGitRepositoryError
- [ ] gitdb.exc.BadName
- [ ] git.exc.GitCommandError
- [ ] git.exc.NoSuchPathError

bpoldrack:
Actually the only thing unpopular thing is making it happen, I guess. ;-)

Just to clarify: Those usage patterns are what you found outside of or including our `*Repo` classes?

yarikoptic:
In my view removal is indeed just a call to reimplement what we already use from it. Could be done? Sure. Is it worth the effort to redo the wheel? Don't think so.
Absorbing all interaction with it within our Repo classes and not leaking it outside of them - that is worthwhile... is it a priority? I don't think so, but if someone does that part - I won't mind

kyleam:
[...]

PR #4170, opened 2020-02-20
RF: Reimplement GitRepo.merge_base() without GitPython

mih:
This furthers gh-2879

### DIFF
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -1977,16 +1977,18 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
             commitishes = commitishes + [self.get_active_branch()]
 
         try:
-            bases = self.repo.merge_base(*commitishes)
-        except GitCommandError as exc:
-            if "fatal: Not a valid object name" in str(exc):
-                return None
-            raise
+            bases = self.call_git_oneline(['merge-base'] + commitishes)
+        except CommandError as exc:
+            if "fatal: Not a valid object name" not in exc.stderr and (
+                    exc.stdout or exc.stderr):
+                # we don't know what this is, and we have something to tell about it
+                # then log it. Without output, it could also just be "no merge base"
+                lgr.debug(exc_str(exc))
+            return None
 
         if not bases:
             return None
-        assert(len(bases) == 1)  # we do not do 'all' yet
-        return bases[0].hexsha
+        return bases
 
     def is_ancestor(self, reva, revb):
         """Is `reva` an ancestor of `revb`?
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -1977,18 +1977,16 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
             commitishes = commitishes + [self.get_active_branch()]
 
         try:
-            bases = self.call_git_oneline(['merge-base'] + commitishes)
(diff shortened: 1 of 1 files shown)


mih:
Appveyor failure is only coverage submission.

datalad/support/gitrepo.py
```diff
@@ -1977,16 +1977,18 @@ def get_merge_base(self, commitishes):
             commitishes = commitishes + [self.get_active_branch()]
 
         try:
-            bases = self.repo.merge_base(*commitishes)
-        except GitCommandError as exc:
-            if "fatal: Not a valid object name" in str(exc):
-                return None
-            raise
+            bases = self.call_git_oneline(['merge-base'] + commitishes)
```
kyleam: `bases` made sense given the gitpython return value, but I think for clarity (because we don't support `--all`) the rewrite should use `base`.

datalad/support/gitrepo.py
```diff
@@ -1977,16 +1977,18 @@ def get_merge_base(self, commitishes):
...
+        except CommandError as exc:
+            if "fatal: Not a valid object name" not in exc.stderr and (
+                    exc.stdout or exc.stderr):
+                # we don't know what this is, and we have something to tell about it
+                # then log it. Without output, it could also just be "no merge base"
+                lgr.debug(exc_str(exc))
+            return None
```
kyleam: There's a change in behavior here, switching to logging unknown exceptions rather than raising them.  My preference would be to raise the error rather than logging in this situation.  Perhaps something like this:

```diff
diff --git a/datalad/support/gitrepo.py b/datalad/support/gitrepo.py
index 30f08f652..642bfb511 100644
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -1979,12 +1979,12 @@ def get_merge_base(self, commitishes):
         try:
             bases = self.call_git_oneline(['merge-base'] + commitishes)
         except CommandError as exc:
-            if "fatal: Not a valid object name" not in exc.stderr and (
-                    exc.stdout or exc.stderr):
-                # we don't know what this is, and we have something to tell about it
-                # then log it. Without output, it could also just be "no merge base"
-                lgr.debug(exc_str(exc))
-            return None
+            if exc.code == 1 and not (exc.stdout or exc.stderr):
+                # No merge base was found (unrelated commits).
+                return None
+            if "fatal: Not a valid object name" in exc.stderr:
+                return None
+            raise
 
         if not bases:
             return None
```

I don't have a strong objection to leaving it as is, though.

datalad/support/gitrepo.py
```diff
@@ -1977,16 +1977,18 @@ def get_merge_base(self, commitishes):
...
 
         if not bases:
```
kyleam: This condition isn't relevant any more, is it?

datalad/support/gitrepo.py
mih: Will do.

datalad/support/gitrepo.py
mih: I like this more than what I did, will change. Thx!

datalad/support/gitrepo.py
mih: True, will remove.

mih:
A pushed the changes you pointed out @kyleam

kyleam:
> A pushed the changes you pointed out @kyleam

Thanks, LGTM

PR #4173, opened 2020-02-20
RF: Discontinue use of GitCommandError

mih:
Was only ever used to fake a GitPython error, but was nowhere acted
upon, specifically. It is now replaced with ValueError, as it is done
in other places when inappropriate arguments are given (here a non-empty
directory).

Also discontinue the practice of logging a warning about a situation
that will unconditionally raise an exception on the next line. If that
exception is caught anywhere, it is the responsibility of the caller to
decide if and how to log this event.

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -73,7 +73,6 @@ from .gitrepo import (
     _normalize_path,
     normalize_path,
     normalize_paths,
-    GitCommandError,
     to_options
 )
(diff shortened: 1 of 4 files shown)


mih:
Windows failure is the unreliable too many files open issue.

bpoldrack:
LGTM. I think it was only ever acted upon back when we used to clone via `*Repo` constructor, which is long gone.

PR #4175, opened 2020-02-20
RF: Use ConfigManager instead of GitPython config classes

mih:
Also strips an unused function with such code.


mih:
Windows problems are related to the familiar (two many open files issue), not caused by this PR.

Issue #4180, opened 2020-02-21
Approach `get()` parallelization


mih:
After #4172 we can approach #450. Here is what I know so far:

This is a basic sketch of an implementation:

```
# go with joblib, it is simple
from joblib import Parallel, delayed

# at the core this is what our functions do, yield dicts
def install_ds(path):
    yield dict(
        path=path,
    )

# joblib doesn't work with generators as jobs
# we need to unwind them
def unwind(fx, *args, **kwargs):
    return list(fx(*args, **kwargs))

paths = [1, 2, 3, 4, 5]

# use a "central" context manager to control the workers,
# avoid over-subscription of CPU and re-use workers
with Parallel(n_jobs=2) as parallel:
    res = parallel(
        delayed(unwind)(install_ds, p)
        for p in paths
    )
# inspect results and act on them
print('RES', res)
```

Issues
- [ ] `Parallel(n_jobs=2, prefer='threads')`  core-dumps for me, hence we would be limited to multiprocessing (might not be an issue)
- [ ] jobs should not "reach up" into a superdataset to avoid concurrency issues, but sadly this is not how we wrote it. RF required.

yarikoptic:
Probably joblib isn't the way to go generally since can't return a generator yet. https://github.com/joblib/joblib/pull/588

Get (title says it, but example goes with install) is probably the last of interest to parallelize since it does it already at annex level. Clone/install,    publish, uninstall, remove those are much more valuable targets IMHO.

mih:
> Get (title says it, but example goes with install) is probably the last of interest to parallelize since it does it already at annex level. Clone/install, publish, uninstall, remove those are much more valuable targets.

`clone` cannot to more than one, `install` is a front-end for `get`, `uninstall/remove` are stone-age code that I will not touch before they have been fully overhauled.

mih:
I am closing this issue. We have concrete (stalled) PRs as a better reference. Latest is https://github.com/datalad/datalad/pull/4434

PR #4181, opened 2020-02-21
BF: Current/target location of a submodule is not a sensible clone source

mih:
Subdataset clone code already filters it out, because of that, but it should never be reported, and the test falsely expected it.

This is simplifying the work on #4180 which will also remove the double-filter again.


mih:
Thx @kyleam !

Issue #4190, opened 2020-02-25
Investigate PermissionError: [WinError 32] (pending annex --batched process)


mih:
This has been showing up in the windows GitHub CI tests for a while

```
Traceback (most recent call last):
  File "C:\hostedtoolcache\windows\Python\3.7.6\x64\lib\site-packages\nose\case.py", line 198, in runTest
    self.test(*self.arg)
  File "C:\hostedtoolcache\windows\Python\3.7.6\x64\lib\site-packages\datalad\tests\utils.py", line 589, in newfunc
    rmtemp(d)
  File "C:\hostedtoolcache\windows\Python\3.7.6\x64\lib\site-packages\datalad\utils.py", line 521, in rmtemp
    rmtree(f, *args, **kwargs)
  File "C:\hostedtoolcache\windows\Python\3.7.6\x64\lib\site-packages\datalad\utils.py", line 434, in rmtree
    _rmtree(path, *args, **kwargs)
  File "C:\hostedtoolcache\windows\Python\3.7.6\x64\lib\site-packages\datalad\utils.py", line 1882, in wrapped
    return f(*args, **kwargs)
  File "C:\hostedtoolcache\windows\Python\3.7.6\x64\lib\site-packages\datalad\utils.py", line 1922, in _rmtree
    return shutil.rmtree(*args, **kwargs)
  File "C:\hostedtoolcache\windows\Python\3.7.6\x64\lib\shutil.py", line 516, in rmtree
    return _rmtree_unsafe(path, onerror)
  File "C:\hostedtoolcache\windows\Python\3.7.6\x64\lib\shutil.py", line 404, in _rmtree_unsafe
    onerror(os.rmdir, path, sys.exc_info())
  File "C:\hostedtoolcache\windows\Python\3.7.6\x64\lib\shutil.py", line 402, in _rmtree_unsafe
    os.rmdir(path)
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\datalad_temp_tree_2d6un7dp'
```

I can reproduce it on a real windows box, and on my linux laptop -- not every time, but frequently. There is a `git-annex` process that is holding on to a directory with a dataset.

```
% DATALAD_ASSERT_NO_OPEN_FILES=pdb TMPDIR="/var/tmp/sym link" python -m nose -v -s datalad.support.tests.test_annexrepo:test_find_batch_equivalence
[...]

PR #4191, opened 2020-02-25
BF: Prevent test cleanup breakage on Windows

mih:
`AnnexRepo` does not manage to cleanup pending `annex --batch` processes
in time for the test cleanup code to wipe out the respective repository.

This only happens in this one test, that explicitly starts a batched
`find`. Hence I decided to make the symptom go away, instead of figuring
out how `AnnexRepo` could always make sure this can never happen.
The code in `AnnexRepo.__del__` aiming to achieve that does not seem to
be sufficient (anymore).

Fixes gh-4190


mih:
Test failure is #4192, and the crawler failure is already fixed in the extension.

mih:
I am merging this now, as a potential inappropriateness of the fix would have minimal impact on the operation. Please reopen #4190 if further work is needed.

PR #4194, opened 2020-02-25
BF+ENH+OPT: @memoize, correct config target for logging outputs

yarikoptic:
Since CI is too busy, decided to just bundle in a single PR.  If needed -- could split into two or three if needed (ideally should have added a test for the `DATALAD_LOG_OUTPUTS` env var actually having an effect).

See individual commits for possibly more info ;-)

### DIFF
--- a/datalad/config.py
+++ b/datalad/config.py
@@ -9,6 +9,7 @@
 """
 """
 
+from functools import lru_cache
 import datalad
 from datalad.consts import (
     DATASET_CONFIG_FILE,
@@ -54,6 +55,7 @@ _where_reload_doc = """
 
 # we cannot import external_versions here, as the cfg comes before anything
 # and we would have circular imports
+@lru_cache()
 def get_git_version(runner=None):
     """Return version of available git"""
     runner = runner or GitRunner()
--- a/datalad/config.py
+++ b/datalad/config.py
@@ -54,8 +54,9 @@ _where_reload_doc = """
 
 # we cannot import external_versions here, as the cfg comes before anything
 # and we would have circular imports
-def get_git_version(runner):
+def get_git_version(runner=None):
     """Return version of available git"""
+    runner = runner or GitRunner()
     return runner.run('git version'.split())[0].split()[2]
 
 
@@ -241,7 +242,7 @@ class ConfigManager(object):
         self._runner = GitRunner(**run_kwargs)
         try:
             self._gitconfig_has_showorgin = \
-                LooseVersion(get_git_version(self._runner)) >= '2.8.0'
+                LooseVersion(get_git_version()) >= '2.8.0'
         except Exception:
             # no git something else broken, assume git is present anyway
             # to not delay this, but assume it is old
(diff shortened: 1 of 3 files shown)


datalad/utils.py
```diff
@@ -1081,6 +1081,31 @@ def wrapped_func(*args, **kwargs):
         return wrapped_func
 
 
+@optional_args
+def memoize(f):
```
kyleam: Why not use `@functools.lru_cache` rather than adding custom code?

datalad/config.py
```diff
@@ -54,6 +55,7 @@
 
 # we cannot import external_versions here, as the cfg comes before anything
 # and we would have circular imports
+@memoize
 def get_git_version(runner):
```
kyleam: Hmm, I wonder why config.py should be the home of `get_git_version`, which is then used by external_versions.py, rather than external_versions.py defining `get_git_version`.  Then config.py could go through the standard `external_versions` interface, and the caching the `git version` result would already be taken care of.

datalad/utils.py
yarikoptic: ha ha -- because I forgot about that one!  will abandon this bastard child here and RF to use it

datalad/config.py
yarikoptic: why? hm, per comment above the function that we cannot import `external_versions` here.
or am I missing smth.
Related: as I have noted (in the commit msg?) this version check is primarily to check against now elderly git 2.8.  As of 7.20191218~26  git-annex debian pkg demands git >= 2.22 (to avoid memory leaks), but in principle it is ok with git 2.1. So it seems that we are doomed to carry this check for just a bit longer.

kyleam: > why? hm, per comment above the function that we cannot import `external_versions` here.
> or am I missing smth

No, I am.  I somehow managed to not see the comment in the context lines of the diff.  Sorry for the sloppy review.

kyleam:
I've switched this to use `lru_cache`.  Note that the previous version with the custom decorator wasn't actually effective because each `ConfigManager` instance passed a new `GitRunner`, so there were no common arguments to cache.  To give some numbers: when I ran `core/local/tests/test_save.py` off the old base of this PR (6fe00418ee), there were 241 calls to `get_git_version`.  With your change in d76b4bee80, there were 239.  The current push gets it down to 2.

yarikoptic:
Heh heh, I guess I checked/saw those 2 saved ones on a simple invocation and got satisfied ;-) thank you!

kyleam:
Windows builds should be back to a green state, so I'll do a no-op rewrite to trigger the tests and make sure no new windows issues were masked.

kyleam:
> I'll do a no-op rewrite to trigger the tests and make sure no new windows issues were masked

Checks out.

My update of this branch to use `lru_cache` hasn't sat for very long, but the idea is the same as the initial iteration of the PR, which didn't receive any objections over a few days, so I'll go ahead with the merge.

PR #4196, opened 2020-02-25
RF: Decomplexify configure_origins()

mih:
Previous code looked at two datasets simultaneously per iteration.
This could result in making a fetch attempt to a non-local repository,
because it would no check whether the second dataset was actually local.

The new code focuses on a single dataset at a time, and all checks are
performed on each one.

### DIFF
--- a/datalad/core/distributed/clone.py
+++ b/datalad/core/distributed/clone.py
@@ -727,38 +727,58 @@ def configure_origins(cfgds, probeds, label=None):
     cfgds : Dataset
       Dataset to receive the remote configurations
     probeds : Dataset
-      Dataset to start looking for 'origin' remotes. May be identical with `cfgds`.
+      Dataset to start looking for 'origin' remotes. May be identical with
+      `cfgds`.
     label : int, optional
       Each discovered 'origin' will be configured as a remote under the name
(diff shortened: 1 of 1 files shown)


kyleam:
> Previous code looked at two datasets simultaneously per iteration.
> This could result in making a fetch attempt to a non-local repository,
> because it would no check whether the second dataset was actually local.

Oy, good catch.

Rewrite looks good to me.

datalad/core/distributed/clone.py
```diff
@@ -727,38 +727,58 @@ def configure_origins(cfgds, probeds, label=None):
...
+        return
+
+    # no need to reconfigure original/direct origin again
+    if cfgds != probeds:
+        # prevent duplicates
+        known_remote_urls = set(
+            cfgds.config.get(r + '.url', None)
+            for r in cfgds.config.sections()
+            if r.startswith('remote.')
+        )
```
kyleam: Hmm, what's the scenario in which these duplicates would arise?  Is it when, in a chain of origins---say "cfgds -> origin A -> origin B -> origin C"---origin C's "origin" remote points back to B?

mih: Yes, that is the scenario

mih:
Thx @kyleam  for the review!

PR #4199, opened 2020-02-26
BF: LFS special remote has "url" in .git/config

yarikoptic:
Without this fix, initial publish would fail:

	/tmp > cd test-github-lfs2

	/tmp/test-github-lfs2 > git annex initremote github-lfs type=git-lfs url=https://github.com/yarikoptic/test-github-lfs2 encryption=none embedcreds=no
	initremote github-lfs ok
	(recording state in git...)

	/tmp/test-github-lfs2 > datalad create-sibling-github --publish-depends github-lfs test-github-lfs2
	[WARNING] Authentication failed using a token.
	[INFO   ] Configure additional publication dependency on "github-lfs"
	.: github(-) [https://github.com/yarikoptic/test-github-lfs2.git (git)]
	'https://github.com/yarikoptic/test-github-lfs2.git' configured as sibling 'github' for <Dataset path=/tmp/test-github-lfs2>

	/tmp/test-github-lfs2 > echo content > file.dat

	/tmp/test-github-lfs2 > datalad save -m content
	add(ok): file.dat (file)
	save(ok): . (dataset)
	action summary:
	  add (ok: 1)
	  save (ok: 1)

	/tmp/test-github-lfs2 > datalad publish --to=github
	CommandError: command '['git', 'fetch', '--verbose', '--progress', 'github-lfs']' failed with exitcode 128
	Failed to run ['git', 'fetch', '--verbose', '--progress', 'github-lfs'] at '/tmp/test-github-lfs2'.
	stdout=
	stderr=fatal: Couldn't find remote ref HEAD

	fatal: Couldn't find remote ref HEAD

Alternative/complimentary solution is coming in a subsequent PR.  I think this one is Ok, the other one might prove to be more robust but more ad-hoc

PR #4200, opened 2020-02-26
BF: allow for fetch to fail due to HEAD not yet been available on a remote for annex

yarikoptic:
This is an alternative solution to https://github.com/datalad/datalad/pull/4199
which is lfs specific.  In this solution which would probably behave "correctly" for any
special remote which has URL and for which we never managed to fetch HEAD.

### DIFF
--- a/datalad/distribution/publish.py
+++ b/datalad/distribution/publish.py
@@ -33,7 +33,10 @@ from datalad.support.constraints import EnsureChoice
 from datalad.support.constraints import EnsureNone
 from datalad.support.annexrepo import AnnexRepo
 from datalad.support.sshconnector import sh_quote
-from datalad.support.exceptions import InsufficientArgumentsError
+from datalad.support.exceptions import (
+    CommandError,
+    InsufficientArgumentsError,
+)
 from datalad.support.network import URL, RI, SSHRI, is_ssh
 
 from datalad.utils import assure_list
@@ -420,8 +423,18 @@ def _publish_dataset(ds, remote, refspec, paths, annex_copy_options, force=False
         # even if we already fetched above we need to do it again
         if is_annex_repo:
             lgr.debug("Obtain remote annex info from '%s'", remote)
-            ds.repo.fetch(remote=remote)
-            ds.repo.merge_annex(remote)
+            try:
+                ds.repo.fetch(remote=remote)
+                ds.repo.merge_annex(remote)
+            except CommandError as exc:
+                # We might be publishing for the first time to this
+                # possibly special (e.g. LFS) annex special remote,
+                # and no HEAD was known
+                if ("Couldn't find remote ref HEAD" in exc.stderr) and \
+                   ('%s/HEAD' % remote not in ds.repo.get_remote_branches()):
+                    pass
+                else:
+                    raise
 
         # Note: git's push.default is 'matching', which doesn't work for first
         # time publication (a branch, that doesn't exist on remote yet)
--- a/datalad/distribution/publish.py
+++ b/datalad/distribution/publish.py
@@ -34,7 +34,6 @@ from datalad.support.constraints import EnsureNone
 from datalad.support.annexrepo import AnnexRepo
 from datalad.support.sshconnector import sh_quote
 from datalad.support.exceptions import (
-    CommandError,
     InsufficientArgumentsError,
 )
 from datalad.support.network import URL, RI, SSHRI, is_ssh
@@ -262,6 +261,15 @@ def _check_and_update_remote_server_info(ds, remote):
     return False
 
 
+def _maybe_fetch(repo, remote):
+    if repo.config.get("remote.{}.fetch".format(remote)):
+        repo.fetch(remote=remote)
+    else:
+        # Fetching would lead to "Couldn't find remote ref HEAD" if no
+        # branch" error.  See gh-4199 for an example.
+        lgr.warning("Remote %s has no configured refspec", remote)
+
+
 def _publish_dataset(ds, remote, refspec, paths, annex_copy_options, force=False, jobs=None,
                      transfer_data='auto', **kwargs):
     remote_branch_name = _get_remote_branch(ds, refspec)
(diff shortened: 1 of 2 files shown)


datalad/distribution/publish.py
```diff
@@ -426,8 +429,18 @@ def _publish_dataset(ds, remote, refspec, paths, annex_copy_options, force=False
...
-            ds.repo.merge_annex(remote)
+            try:
+                ds.repo.fetch(remote=remote)
+                ds.repo.merge_annex(remote)
+            except CommandError as exc:
+                # We might be publishing for the first time to this
+                # possibly special (e.g. LFS) annex special remote,
+                # and no HEAD was known
+                if ("Couldn't find remote ref HEAD" in exc.stderr) and \
+                   ('%s/HEAD' % remote not in ds.repo.get_remote_branches()):
```
kyleam: Failing to find HEAD on the remote and the existence of the local `refs/remotes/<remote>/HEAD` are distinct things.  If a user doesn't like recording a default branch for the remote, they can `git remote set-head <remote> --delete`, and it won't be created again.

While it may seem unlikely that the remote was successfully fetched from, then locally the remote's HEAD was removed, and then for a later fetch the remote HEAD couldn't be found, I don't see a benefit of making this assumption.  Would just checking whether any refs are under `refs/remotes/<remote>` be a reliable enough indicator that we shouldn't raise the error in this case?  Something like the (untested)

```diff
diff --git a/datalad/distribution/publish.py b/datalad/distribution/publish.py
index ed2ad6cec8..5aa6250790 100644
--- a/datalad/distribution/publish.py
+++ b/datalad/distribution/publish.py
@@ -436,10 +436,7 @@ def _publish_dataset(ds, remote, refspec, paths, annex_copy_options, force=False
                 # We might be publishing for the first time to this
                 # possibly special (e.g. LFS) annex special remote,
                 # and no HEAD was known
-                if ("Couldn't find remote ref HEAD" in exc.stderr) and \
-                   ('%s/HEAD' % remote not in ds.repo.get_remote_branches()):
-                    pass
-                else:
+                if next(ds.repo.for_each_ref_(pattern="refs/remotes/" + remote):
                     raise
 
         # Note: git's push.default is 'matching', which doesn't work for first
```

---

As a tangent, I wonder if a non-zero exit code for an empty repository is specific to the hosting service or git version.  With git 2.25.1, if I add an empty gitlab repo or regular ssh target as a remote and fetch, the fetch call doesn't fail; it just doesn't bring anything down.

yarikoptic: sounds reasonable, I will add the change, but will delegate to `get_remote_branches` use of `refs/remotes/` ;-)
FWIW -- system git I have is 2.25, while the one bundled with git annex is (and probably was) 2.24.  So, in your case if you run the snippet in #4199 (sorry I did not make it totally cut/pasteable) -- it works on master (or ideally 0.12.2) version?

kyleam: >  but will delegate to `get_remote_branches` use of `refs/remotes/` ;-)

Okay.  The advantage of using `refs/remotes/<remote>` is that we have potentially less output to process, but, given that we're working within an exception that should not be triggered often, it's unlikely that it matters.

kyleam: >  So, in your case if you run the snippet in #4199 (sorry I did not make it totally cut/pasteable) -- it works on master (or ideally 0.12.2) version?

Sorry, I haven't taken the time to put datalad into the mix.  I'm just talking about when I create an empty repo on gitlab or an plain ssh server, add that as a remote, and `git fetch` it.  I tried with an empty repo on github, and `git fetch` (via ssh and https) exits with status of 0.

Based on the description of the problem, I think that's the state things are in when this call to `fetch` in `publish` fails, but perhaps it's some other in between state, and that is the reason for the difference.

kyleam: Stepping through fetch.c, the essential piece to trigger "fatal: Couldn't find remote ref HEAD" is that the remote does not have `remote.*.fetch` set.

yarikoptic: Ha. So may be we shouldn't even fetch for such remotes?

kyleam: > So may be we shouldn't even fetch for such remotes?

Given both `clone` and `remote add` will populate `.fetch` values, I think that'd be a reasonable approach, though I don't mind sticking with the current one either.

kyleam: > So may be we shouldn't even fetch for such remotes?

I've updated the PR to go with this approach.

<details>
<summary>range-diff</summary>

```
1:  19854724e1 = 1:  ac1871c47e BF: allow for fetch to fail due to HEAD not yet been available on a remote for annex
-:  ---------- > 2:  676d360692 BF: publish: Don't fetch if remote.*.fetch is not set
```

</details>

kyleam:
AppVeyor failure is the familiar `Invoke-WebRequest`issue.

datalad/distribution/tests/test_publish.py
```diff
@@ -713,3 +716,18 @@ def test_gh1811(srcpath, clonepath):
...
+
+    path = Path(path)
+    GitRunner(cwd=str(path)).run(["git", "init", "--bare", "empty-remote"])
+    ds = Dataset(path / "ds").create()
+    ds.repo.add_remote("origin", str(ds.pathobj.parent / "empty-remote"))
+    # Mimic a situation that can happen with an LFS remote. See gh-4199.
+    ds.repo.config.unset("remote.origin.fetch", where="local")
+    (ds.repo.pathobj / "foo").write_text("a")
+    ds.save()
+    ds.publish(to="origin")
```
yarikoptic: I wonder if it is worth to explicitly note that it is just a smoke test.  it could have been an explicit mock for .fetch verifying that it was not called

kyleam: > I wonder if it is worth to explicitly note that it is just a smoke test.

I don't find "smoke test" a useful description.  It's specific purpose it to make sure that the fetches do not fail in this situation.  If you revert the changes from this PR, you see the failure that the changes here are trying to prevent: the "fatal: Couldn't find remote ref HEAD" if no branch" error on master or the GitPython failure on maint.

> it could have been an explicit mock for .fetch verifying that it was not called

Ew

kyleam:
> Ideally though it should have mimicked original problem better - there same source repo is used as both regular and special remote (eg directory), with dependence for publishing. Then test could check that things got published correctly.

In my opinion the added test covers the core issue well.  But please feel free to take back over your PR and make whatever changes you'd prefer.

yarikoptic:
> But I am ok with it as is since it fixes the issue

Thank you Kyle for taking care about this pr!

PR #4203, opened 2020-02-26
RF: RIA remote

bpoldrack:
First of the announced follow up PR sitting on top of PR #4164 .

This includes:

 - [x] move creation routines out of the special remote itself, leading to `initremote` and `prepare` to expect an existing and accessible storage location. Furthermore split store from dataset representation in that regard, meaning two creation routines (one for a store and its layout and for a dataset within a store and its layout)
 - [x] rewrite tests
 - [x] New concept of storing rewritten URL in special remote config and its uuid in the bare repo's config in a store. `clone` gets a fallback solution in `postclone_ria` in case we cloned from a RIA store and `annex-init` didn't enable an ORA remote. We would now look for an ORA remote UUID in origin's config and try to enable this one using the `ria+` URL we cloned from. This is useful for example when the special remote URL used to publish to the store isn't valid from the point of view of a consumer. Note, that this currently supports `file:` and `ssh:` URLS. Meaningful `http:`support depends on actual support for HTTP in the special remote itself.
 - [x] renaming of the special remote and related configs, etc. from RIA to ORA. Includes command renaming `ria-export-archive` to `ora-export-archive`
- [x] exclude `@turtle` tests in AppVeyor, too. Not just in Travis.

There is still some RF'ing to do and issues like #4436 and #4437 to address/investigate. However, this should be based on the remaining RF'ing which depends on PR #4068. Plan is to build a remote execution abstraction using the approach of `SSHRemoteIO` in the special remote but with the async based new runner.


mih:
Needs to start from #4260 now

bpoldrack:
Update: rebased on master and hopefully fixed the fallout. `test_ria_basics` should now pass again.
Proceeding to rewrite/restructure the remaining tests to account for change that requires creation of the store location beforehand now.

datalad/core/distributed/clone.py
```diff
@@ -585,15 +585,125 @@ def postclonecfg_ria(ds, props):
                    if s.get('annex-externaltype', None) == 'ora'
     ]
     if not ria_remotes:
-        lgr.debug("Found no RIA special remote")
+        # nothing autoenabled. Let's check origin's config for
+        # datalad.ora-remote.uuid as stored by create-sibling-ria and enable try
+        # enabling that one.
+        lgr.debug("Found no autoenabled ORA special remote. Trying to look it"
+                  "up in source config ...")
```
mih, line 592: One of the lines needs a space.

datalad/core/distributed/clone.py
```diff
@@ -585,15 +585,125 @@ def postclonecfg_ria(ds, props):
...
+
+        # First figure whether we cloned via SSH, HTTP or local path and then
+        # get that config file the same way:
+
+        config_content = None
+        scheme = props['source'].split(':')[0][4:]
```
mih, line 598: This assumes the `props['source']` is something like `ria+...://` which is true at this point (with the current code), but it could be relatively painful to decode this fact back from this line in a future RF. Maybe add a comment? Or strip `len('ria+')`?

datalad/core/distributed/clone.py
```diff
@@ -585,15 +585,125 @@ def postclonecfg_ria(ds, props):
...
+            # except requests.RequestException as e:
+            #     lgr.debug("Failed to get config file from source:\n%s",
+            #               exc_str(e))
+
+            # config_content still None should lead to skip everything special
+            # remote related. Instead enable origin to deliver annex keys via
+            # HTTP:
+            lgr.debug("Trying to enable remote origin for git-annex")
+            ds.repo.enable_remote('origin')
+            ds.repo.fsck(remote='origin', fast=True)
```
mih, line 622: I assume the last line is needed, because git annex isn't aware of any availability at "origin", because the data was copied through the special remote. If that is correct, then I think we should not do this. The result would be that annex believe that two copies of the file exist, where there is only one -- thereby invalidating any minimum redundancy requirements.

datalad/core/distributed/clone.py
```diff
@@ -585,15 +585,125 @@ def postclonecfg_ria(ds, props):
...
+            # changed with command abstractions). So we need to get that part to
+            # have a valid path to origin's config file:
+            from datalad.support.network import URL
+            from datalad.utils import PurePosixPath
+            cfg_path = PurePosixPath(URL(props['giturl']).path) / 'config'
+            op = SSHRemoteIO(props['giturl'])
+            try:
+                # TODO: double-check path
+                response = op.read_file(cfg_path)
+                config_content = response
```
mih, line 638: why keep `response`?

datalad/core/distributed/clone.py
```diff
@@ -585,15 +585,125 @@ def postclonecfg_ria(ds, props):
...
+            except RIARemoteError as e:
+                lgr.debug("Failed to get config file from source:\n%s",
+                          exc_str(e))
```
mih, line 641: Just a side-comment: I don't know why we have `exc_str()` (compares to having a `__str__()` in our exceptions), but there is a cost for each *not shown* debug message for using it -- in contrast to just passing `e` and let stringification happen later, if actually needed.

datalad/core/distributed/clone.py
```diff
@@ -585,15 +585,125 @@ def postclonecfg_ria(ds, props):
...
+            op = LocalIO()
+            from datalad.support.network import URL
+            cfg_path = Path(URL(props['giturl']).localpath) / 'config'
+            try:
+                # TODO: double-check path
+                response = op.read_file(cfg_path)
+                config_content = response
+            except RIARemoteError as e:
+                lgr.debug("Failed to get config file from source:\n%s",
+                          exc_str(e))
```
mih, line 656: Maybe I missed it, but is the `ssh` conditional completely identical to the `file` conditional? Can this duplication be avoided?

datalad/core/distributed/clone.py
```diff
@@ -585,15 +585,125 @@ def postclonecfg_ria(ds, props):
...
+        else:
+            lgr.debug("Unknown URL-Scheme in %s. Can handle HTTP, SSH or FILE."
+                      "", props['source'])
```
mih, line 659: What is the `""` for?

datalad/core/distributed/clone.py
```diff
@@ -585,15 +585,125 @@ def postclonecfg_ria(ds, props):
...
+                try:
+                    result = runner.run(
+                        ['git', 'config', '-f', cfg_file,
+                         'datalad.ora-remote.uuid'],
+                        protocol=StdOutCapture
+                    )
+                    org_uuid = result['stdout'].strip()
+                except CommandError:
+                    # doesn't contain what we are looking for
+                    lgr.debug("Found nothing in source config.")
```
mih, line 682: Maybe `Found no UUID for ORA special remote at 'origin'`?

datalad/core/distributed/clone.py
```diff
@@ -585,15 +585,125 @@ def postclonecfg_ria(ds, props):
...
+            srs = ds.repo.get_special_remotes()
+
+            # TODO: - Double-check autoenable value and only do this when true?
+            #       - What if still fails? -> Annex shouldn't change config in
+            #         that case
+
+            new_url = props['source'].split('#')[0]  # we only need the store
+            ds.repo.enable_remote(srs[org_uuid]['name'],
+                                  options=['url={}'.format(new_url)]
+                                  )
```
mih, line 697: Should protect against `org_uuid` not being a key in `srs`. Or is this impossible? IOW can it be that the store has a remote on file, but the special remote config in the dataset is not there? I think it can happen (git-annex branch not pushed/uptodate, maybe even special remote config removed).

datalad/core/distributed/clone.py
```diff
@@ -585,15 +585,125 @@ def postclonecfg_ria(ds, props):
...
+
     elif len(ria_remotes) == 1:
         yield from ds.siblings('configure',
```
mih, line 700: All autoenabling happens in a conditional branch that prevents this line from being reached, so "discovered" special remotes will not be set up as publication dependencies.

datalad/core/distributed/clone.py
bpoldrack: Agree. `lstrip("ria+")` might be most explicit.

mih: Careful with using `*strip()`: 'aaaaa'.lstrip('a')

Better stick to your pattern, but with `len()`

datalad/core/distributed/clone.py
bpoldrack: I'm struggling with the need for this as well. There are several possible cases: We could have store with no special remote at all, in which case `fsck` should be superfluous. So, even in case of HTTP we should prob. check origin's config and not call `fsck` if there's no special remote uuid to be found.
In case a special remote was indeed used for upload to the store, I think the `fsck` is needed, since otherwise there's no way to get any content via origin. But you're right - minimum redundancy would be torpedoed. I did not yet play with `sameas`. May be that would be a solution. So, if we decide to try to use origin we also declare it `sameas` the special remote.

datalad/core/distributed/clone.py
bpoldrack: This can go indeed. Separation was an intermediate need.

datalad/core/distributed/clone.py
bpoldrack: All that's different is the IO class to be used (which is to be replaced by the command abstraction PR). So, should be unified with the command abstraction RF. Determination of the actual path to access is also slightly different ATM, but should become identical via the new command abstractions.

datalad/core/distributed/clone.py
bpoldrack: hehe. Prob. used a `format` call before. In that case I often use `"".format(...` in a new line.

datalad/core/distributed/clone.py
bpoldrack: Yes, better.

datalad/core/distributed/clone.py
bpoldrack: Yes, seems possible to happen.

datalad/core/distributed/clone.py
bpoldrack: Thx! Missed that.

datalad/core/distributed/clone.py
bpoldrack: `exc_str` is about getting its original location from traceback. TBH I forgot when exactly this is needed and when it's superfluous.

datalad/core/distributed/clone.py
bpoldrack: Yeah - not concerned with "ria+ria+ria+" URLs actually, but still simply faster to with slicing.

datalad/core/distributed/clone.py
bpoldrack: No idea why this isn't marked outdated. But this is done.

datalad/core/distributed/clone.py
mih: I think, if there nothing equivalent to `sameas` then the approach must be dropped.

datalad/core/distributed/clone.py
mih: It would also strip `air+` or `ira+` or `ai`. OMG it removes any AI...

datalad/core/distributed/clone.py
bpoldrack: Should be fine to finish polishing that with PR #4068

datalad/core/distributed/clone.py
bpoldrack: > OMG it removes any AI...

Sounds like a feature ;-)

kyleam:
@bpoldrack Just a heads up: An old run for this on appveyor is approaching the 1 hr limit (another from the same build already hit it), and there are several queued builds for old iterations of this PR.  I'm going to prune those so that appveyor isn't clogged up.

datalad/core/distributed/clone.py
bpoldrack: Played a bit with passing `sameas` and `uuid` parameters, but no success so far. So - will drop HTTP for now and do a proper implementation in ORA remote. But won't blow up this PR even more.

bpoldrack:
@kyleam : Yes, feel free to cancel those any time. I think this PR is "feature complete" now, so will try to tackle failing tests (like crippled FS) and figure what's still running forever and not marked `turtle` yet only now.

mih:
@bpoldrack this should be a draft PR, I think.

datalad/core/distributed/clone.py
yarikoptic: It did come super useful to me many times when I had to dig out how the heck an exception came about to be logged (especially is swallowed upstairs somewhere).  We have `DATALAD_EXC_STR_TBLIMIT` env variable to instruct to include such information.

As for performance -- if you are hitting exceptions, I think a few `ms` spent on collecting a bit more data would not really be what causes bad UX ;)

yarikoptic: <details>
<summary>Did some basic timing of exc_str. Takes ~2ms on a rudimentary use case. IMHO it is worth a "wait".  2ms means that you would wait for 1 sec if you manage to throw and catch 500 exceptions.  Did not time how long logging itself would be taking in that case etc</summary> 

```shell
$> DATALAD_EXC_STR_TBLIMIT=10 python ../trash/bench_exc_str.py
str:     0.000009 = my exception
exc_str: 0.001914 = my exception [bench_exc_str.py:<module>:12,bench_exc_str.py:f1:9,bench_exc_str.py:f2:8,bench_exc_str.py:f3:7]
(dev3) 1 22499.....................................:Tue 28 Apr 2020 12:56:58 PM EDT:.
$> python ../trash/bench_exc_str.py        
str:     0.000008 = my exception
exc_str: 0.001510 = my exception [bench_exc_str.py:f3:7]
(dev3) 1 22500.....................................:Tue 28 Apr 2020 12:57:03 PM EDT:.
$> cat ../trash/bench_exc_str.py
import timeit
from datalad.dochelpers import exc_str

def f1():
    def f2():
        def f3():
            raise IndexError("my exception")
        f3()
    f2()

try:
    f1()
except Exception as e:
    g = {'e': e, 'exc_str': exc_str}
    print("str:     %f = %s" % (timeit.timeit("str(e)", globals=g, number=100), str(e)))
    print("exc_str: %f = %s" % (timeit.timeit("exc_str(e)", globals=g, number=100), exc_str(e)))

```
</details>

bpoldrack:
Note, that failure in CrippledFS github workflow is in `test_create_sibling_gitlab`. No Idea how this could be related.
AppVeyor currently hangs (#4436). I don't think this is reasonably addressable within this PR. If we agree on that, I'll skip the test on windows and address it in context of rewriting SSH remote execution based on command abstractions and async.

appveyor.yml
```diff
@@ -7,6 +7,8 @@ environment:
       PYTHON_ARCH: "32"
       MINICONDA: C:\Miniconda35
       DATALAD_TESTS_SSH: 1
+      DATALAD_LOG_LEVEL: 5
+      DATALAD_LOG_TRACEBACK: "collide"
```
mih, line 11: Why those? Travis doesn't do that.

datalad/distributed/create_sibling_ria.py
```diff
@@ -52,9 +56,11 @@
     decode_source_spec
 )
 from datalad.log import log_progress
-from datalad.distributed.ria_utils import (
+from datalad.customremotes.ria_utils import (
```
mih, line 59: I don't like this move. We should stick to the plan to slowly move code into the new package structure. https://github.com/datalad/datalad/issues/3192

datalad/distributed/ora_export_archive.py
```diff
@@ -45,23 +45,23 @@
...
 )
 
-lgr = logging.getLogger('datalad.customremotes.export_archive')
+lgr = logging.getLogger('datalad.customremotes.ora_export_archive')
 
 
 @build_doc
-class RIAExportArchive(Interface):
-    """Export an archive of a local annex object store for the RIA remote.
+class ORAExportArchive(Interface):
```
mih, line 52: I wonder, if we rename, should we rename to `export_archive_ora` or at least `export_ora_archive`?

datalad/distributed/create_sibling_ria.py
bpoldrack: Yeah, I'm not entirely clear about how to deal with the actual special remote. #3192 is mostly concerned with "commands". In #4164 you moved parts into `customremotes` and parts into 
`distributed`. I had the impression that the actual special remote should be in `customremotes` while related commands are supposed to go into `distributed`.

appveyor.yml
bpoldrack: Left over. Will remove alongside skipping the hanging test on windows (if we agree on that).

datalad/distributed/ora_export_archive.py
bpoldrack: Agree. To me `export_archive_ora` feels better. We are exporting *for* use by ORA, not from something used by ORA.

bpoldrack: Done.

datalad/distributed/create_sibling_ria.py
mih: I can live with that. Thx for the explanation.

mih:
@bpoldrack I have no issues with disabling Windows tests and fixing the bugs later. Usage on Windows is still largely hypothetical, and we need to make a release.

datalad/core/distributed/clone.py
```diff
@@ -582,20 +582,136 @@ def postclonecfg_ria(ds, props):
     # and was enabled (there could be RIA stores that actually only have repos)
     # make this function be a generator
     ria_remotes = [s for s in ds.siblings('query', result_renderer='disabled')
-                   if s.get('annex-externaltype', None) == 'ora'
-    ]
+                   if s.get('annex-externaltype', None) == 'ora']
     if not ria_remotes:
```
mih, line 586: I know that this isn't introduced here, but shouldn't the entire lookup operation be conditions upon the actual presence of _any_ ora special remote configuration, i.e. look into `AnnexRepo.get_special_remote()`. That would be cheap, but all the remote talking is relatively expensive. For example, the HCP datasets live in a RIA store but none has an ORA remote configured. It already takes 5+ hours to clone them. This would make it even longer.

kyleam:
Sorry, I've canceled the remaining AppVeyor builds for this PR.  Those one hour hangs are really clogging up the queued jobs given AppVeyor runs one job at a time.  If it's just one or a few tests hanging, perhaps decorating the tests with `@nose.timed(<some reasonable value>)` would be a good way to debug these.

datalad/core/distributed/clone.py
bpoldrack: Agree. But to be solved separately, right?

mih:
Trying this code I see this:
```
dl clone ria+ssh://bulk1.htc.inm7.de/ds/cat_12.5/srv#f2e9193e-8401-11ea-99ab-a0369f287950 myclone
[INFO   ] RIA store doesn't exist at /ds/cat_12.5/srv                                        
[INFO   ] Configure additional publication dependency on "inm7-storage"                      
install(ok): /tmp/myclone (dataset)
action summary:
  configure-sibling (ok: 1)
  install (ok: 1)
```

The store does exist, cloning and getting work fine. The message is wrong, or misleading.

mih:
Performance anecdote: Rsyncing a dataset store location vs getting a full dataset (7k files, 23GB total, all files via ora-special remote) is ~70% faster. I am not sure which pieces in the (obviously more complicated and more capable) stack contribute to the slowdown, but it might be worth investigation at some point.

mih:
This is broken for `GitRepo`:

```
% datalad create --no-annex
% datalad create-sibling-ria -d . -s inm7 --trust-level trust ria+file:///ds/cat_12.5/srv
[INFO   ] create siblings 'inm7' and 'inm7-storage' ... 
[ERROR  ] 'GitRepo' object has no attribute 'init_remote' [create_sibling_ria.py:_create_sibling_ria:491] (AttributeError) 
```

It should not attempt to enable a storage sibling and then crash for a non-annex dataset.

mih:
Code seems to have issues with plain `GitRepo`:

```
% rm -r /ds/cat_12.5/srv/653/8a7fe-8ac7-11ea-83ba-a0369f287950 
rm: cannot remove '/ds/cat_12.5/srv/653/8a7fe-8ac7-11ea-83ba-a0369f287950': No such file or directory
% dl create-sibling-ria -d . -s inm7 ria+file:///ds/cat_12.5/srv --no-storage-sibling
[INFO   ] create sibling 'inm7' ... 
[ERROR  ] [Errno 17] File exists: '/ds/cat_12.5/srv/653/8a7fe-8ac7-11ea-83ba-a0369f287950' [pathlib.py:mkdir:1251] (FileExistsError)
```

bpoldrack:
Plain git issues addressed in https://github.com/datalad/datalad/pull/4203/commits/18e45e49586ae54b2ef484094296734be031fb09

bpoldrack:
For now skipping ORA tests completely. #4469
No time to investigate ATM.

bpoldrack:
While above mentioned call worked for @mih as well, the actual issue was still not solved.

Issue was that the `[INFO   ] RIA store doesn't exist at /ds/cat_12.5/srv ` is a factually correct but irrelevant failure. It was checking that path locally rather than at the store's machine via SSH. This is because the special remote URL committed in the dataset is a `ria+file:` URL, while we cloned on a different machine via `ria+ssh:`. Therefore autoenabling during `git-annex-init` failed (correctly!) and new feature in `datalad-clone` fix the situation by using the URL we cloned from to reconfigure the special remote.
Addressed in https://github.com/datalad/datalad/pull/4203/commits/58016f8f9ab2683ec08184d4f495a306b6613bb5

datalad/core/distributed/clone.py
mih: I can fix that in master right now. Conflicts will then have to be resolved in here. I am OK with either approach.

mih: No, I take that back. The problem doesn't exist in master. It is introduced here.

mih: I pushed a fix in here.

bpoldrack: I thought replacing the `siblings` call right away and collect info on all special remotes (enabled or not) in one call. Anyways - your commit solves the bigger issue. Agree.

datalad/core/distributed/clone.py
```diff
@@ -581,21 +581,141 @@ def postclonecfg_ria(ds, props):
...
-                               result_filter=None,
-                               result_renderer='disabled')
-    else:
-        lgr.warning("Found multiple RIA remotes. Couldn't decide which "
-                    "publishing to origin should depend on: %s",
-                    [r['name'] for r in ria_remotes])
+    ora_remotes = [s for s in ds.siblings('query', result_renderer='disabled')
+                   if s.get('annex-externaltype', None) == 'ora']
+    if not ora_remotes and any(
+            r.get('externaltype', None) == 'ora'
```
yarikoptic, line 587: although I keep doing it as well (but is trying to break the habit), but note that `, None` is not necessary. That is the point of `.get` vs `__getitem__` that it would return `None` if key is not present - there is a default `None`

datalad/core/distributed/clone.py
```diff
@@ -581,21 +581,141 @@ def postclonecfg_ria(ds, props):
...
+                lgr.debug("Failed to get config file from source:\n%s",
+                          e)
+        else:
+            lgr.debug("Unknown URL-Scheme in %s. Can currently handle SSH or "
+                      "FILE scheme URLs.", props['source'])
+
+        # 3. And read it
+        org_uuid = None
+        if config_content:
+            from datalad.utils import make_tempfile
```
yarikoptic, line 655: no reason to delay imports from core modules such as `datalad.utils` (and probably such as `.support.network`) all the way to the point of use.  Those modules would get imported anyways somewhere else in the code and should  not be taxing on performance.

datalad/core/distributed/clone.py
```diff
@@ -581,21 +581,141 @@ def postclonecfg_ria(ds, props):
...
+                except CommandError as e:
+                    lgr.debug("Failed to reconfigure ORA special remote: %s",
+                              e)
+            else:
+                lgr.debug("Unknown ORA special remote uuid at 'origin': %s",
+                          org_uuid)
+    if ora_remotes:
+        if len(ora_remotes) == 1:
+            yield from ds.siblings('configure',
+                                   name='origin',
```
yarikoptic, line 711: FWIW it might be worth defining a const like `ORA_REMOTE_NAME = 'origin'` and using through the code instead of hardcoding 'origin' everywhere.  Might help later on if multiple remotes providing ORA would be supported etc.

datalad/core/distributed/clone.py
```diff
@@ -581,21 +581,141 @@ def postclonecfg_ria(ds, props):
...
+        # no ORA remote autoenabled, but configuration know about at least one.
+        # Let's check origin's config for datalad.ora-remote.uuid as stored by
+        # create-sibling-ria and enable try enabling that one.
+        lgr.debug("Found no autoenabled ORA special remote. Trying to look it "
+                  "up in source config ...")
+
+        # First figure whether we cloned via SSH, HTTP or local path and then
+        # get that config file the same way:
+        config_content = None
+        scheme = props['giturl'].split(':')[0]
```
yarikoptic, line 598: since we are all about micro-optimizations these days,
```suggestion
        scheme = props['giturl'].split(':', 1)[0]
```

datalad/core/distributed/clone.py
```diff
@@ -581,21 +581,141 @@ def postclonecfg_ria(ds, props):
...
+
+        elif scheme == 'ssh':
+
+            # TODO: switch the following to proper command abstraction:
+            from datalad.distributed.ora_remote import SSHRemoteIO, RIARemoteError
+            # SSHRemoteIO ignores the path part ATM. No remote CWD! (To be
+            # changed with command abstractions). So we need to get that part to
+            # have a valid path to origin's config file:
+            from datalad.support.network import URL
+            from datalad.utils import PurePosixPath
```
yarikoptic, line 627: move imports to the top of the file, see another similar comment

datalad/core/distributed/clone.py
```diff
@@ -581,21 +581,141 @@ def postclonecfg_ria(ds, props):
...
+            cfg_path = PurePosixPath(URL(props['giturl']).path) / 'config'
```
yarikoptic, line 628: it might be good to program a bit more defensively here and try/except for KeyError in case `giturl` was not defined, and spit out a specific meaningful exception describing what is missing `giturl` which is required

datalad/core/distributed/clone.py
```diff
@@ -581,21 +581,141 @@ def postclonecfg_ria(ds, props):
...
+
+            # TODO: switch the following to proper command abstraction:
+            from datalad.distributed.ora_remote import LocalIO, RIARemoteError
+            op = LocalIO()
+            from datalad.support.network import URL
+            cfg_path = Path(URL(props['giturl']).localpath) / 'config'
+            try:
+                config_content = op.read_file(cfg_path)
+            except (RIARemoteError, OSError) as e:
+                lgr.debug("Failed to get config file from source:\n%s",
```
yarikoptic, line 646: why here specifically to bother with `\n` to separate out the exception but not in other places?

datalad/core/distributed/clone.py
```diff
@@ -581,21 +581,141 @@ def postclonecfg_ria(ds, props):
...
+
+        # 3. And read it
+        org_uuid = None
+        if config_content:
+            from datalad.utils import make_tempfile
+            from datalad.cmd import (
+                GitRunner,
+                WitlessRunner,
+                StdOutCapture
+            )
```
yarikoptic, line 660: I don't think delaying these imports also would make sense, since we are guaranteed to import those early to run `git config` etc anyways

datalad/core/distributed/clone.py
```diff
@@ -581,21 +581,141 @@ def postclonecfg_ria(ds, props):
...
+                    result = runner.run(
+                        ['git', 'config', '-f', cfg_file,
+                         'datalad.ora-remote.uuid'],
+                        protocol=StdOutCapture
+                    )
+                    org_uuid = result['stdout'].strip()
+                except CommandError:
+                    # doesn't contain what we are looking for
+                    lgr.debug("Found no UUID for ORA special remote at "
+                              "'origin'")
```
yarikoptic, line 674: worth providing details of the `CommandError` exception here in debug message.  Could crash for a variety of reasons (formatting, etc) and hiding it behind a too generic message would require more digging to figure out what went wrong

datalad/core/distributed/clone.py
```diff
@@ -581,21 +581,141 @@ def postclonecfg_ria(ds, props):
...
+    if ora_remotes:
+        if len(ora_remotes) == 1:
+            yield from ds.siblings('configure',
+                                   name='origin',
+                                   publish_depends=ora_remotes[0]['name'],
+                                   result_filter=None,
+                                   result_renderer='disabled')
+        else:
+            lgr.warning("Found multiple ORA remotes. Couldn't decide which "
+                        "publishing to origin should depend on: %s",
```
yarikoptic, line 717: and what should the user do when is warning displayed - should it may be suggest something ? (remove one?)

datalad/customremotes/ria_utils.py
```diff
@@ -0,0 +1,158 @@
...
+"""Helper for RIA stores
+
+"""
+
+
+class UnknownLayoutVersion(Exception):
+    pass
+
+
+known_versions_objt = ['1', '2']
```
yarikoptic, line 18: what is `_objt`.  having comments describing those constants (are they? if so - make `tuple` ?) would help new to the code developers to orient themselves

datalad/customremotes/ria_utils.py
```diff
@@ -0,0 +1,158 @@
...
+      the directory with archive files for the dataset, and the
+      annex object directory are return in that order.
+    """
+    if version == 1:
+        dsgit_dir = base_path / dsid[:3] / dsid[3:]
+        archive_dir = dsgit_dir / 'archives'
+        dsobj_dir = dsgit_dir / 'annex' / 'objects'
+        return dsgit_dir, archive_dir, dsobj_dir
+    else:
+        raise ValueError("Unknown layout version: {}".format(version))
```
yarikoptic, line 48: include information on which are known? (known_versions_dst? known_versions_objt?)

datalad/customremotes/ria_utils.py
```diff
@@ -0,0 +1,158 @@
...
+    ------
+    ValueError
+
+    Returns
+    -------
+    tuple
+      (host, base-path, rewritten url)
+    """
+    from datalad.config import rewrite_url
+    from datalad.support.network import URL
```
yarikoptic, line 78: no need to delay the imports... since so prevalent -- is there a shift in coding style to delay all imports until point of use we should all discuss/adopt???

datalad/customremotes/ria_utils.py
```diff
@@ -0,0 +1,158 @@
...
+    url = rewrite_url(cfg, url)
+    url_ri = URL(url)
+    if not url_ri.scheme.startswith('ria+'):
+        raise ValueError("Missing ria+ prefix in final URL: %s" % url)
+    if url_ri.fragment:
+        raise ValueError(
+            "Unexpected fragment in RIA-store URL: %s" % url_ri.fragment)
+    protocol = url_ri.scheme[4:]
+    if protocol not in ['ssh', 'file']:
+        raise ValueError("Unsupported protocol: %s" % protocol)
```
yarikoptic, line 92: include the list of supported ones (define a const with them?)

datalad/customremotes/ria_utils.py
```diff
@@ -0,0 +1,158 @@
...
+
+    return url_ri.hostname if protocol == 'ssh' else None, url_ri.path, url
+
+
+def create_store(io, base_path, version):
```
yarikoptic, line 97: Docstring would be of value here since it is not private method, and `io` is not even easy to figure out without knowing details of the codebase around here.

datalad/customremotes/ria_utils.py
```diff
@@ -0,0 +1,158 @@
...
+
+    # At store level the only version we know as of now is 1.
+    if version not in known_versions_dst:
+        raise UnknownLayoutVersion("store layout version unknown: {}"
```
yarikoptic, line 101: and which ones are "known"? BTW, is known here (see my comment about adding a comment) is really "known" or "supported"?

datalad/customremotes/ria_utils.py
```diff
@@ -0,0 +1,158 @@
...
+    #       Requires proper IO (command abstraction class).
+    #       But: Likely unnecessary. For now, check for version conflict only.
+    if io.exists(version_file):
+        existing_version = io.read_file(version_file).split('|')[0].strip()
+        if existing_version != version.split('|')[0]:
+            # We have an already existing location with a conflicting version on
+            # record.
+            # Note, that a config flag after pipe symbol is fine.
+            raise ValueError("Conflicting version found at target: {}"
+                             "".format(existing_version))
```
yarikoptic, line 116: I think it would be as fine with
```suggestion
                             .format(existing_version))
```

datalad/customremotes/ria_utils.py
```diff
@@ -0,0 +1,158 @@
...
+            # already exists, recorded version fits - nothing to do
+            return
+
+    # Note, that the following does create the base-path dir as well, since
+    # mkdir has parents=True:
+    io.mkdir(error_logs)
+    io.write_file(version_file, version)
+
+
+def create_ds_in_store(io, base_path, dsid, obj_version, store_version):
```
yarikoptic, line 127: docstring

datalad/distributed/tests/test_ria_basics.py
```diff
@@ -206,9 +359,133 @@ def test_version_check(path, objtree):
...
+def test_gitannex_ssh():
+    _test_gitannex('datalad-test')
+
+
+@slow
+def test_gitannex_local():
+    _test_gitannex(None)
+
+
+@skip_if_on_windows  # see gh-4469
```
yarikoptic, line 435: should it ideally be `@known_failure_windows` (here and others)?

datalad/core/distributed/clone.py
bpoldrack: What would be a useful description here? It's the URL we just cloned from. If it's not in `props` we have a bug in `clone_dataset`. I wouldn't know a better message for a user. They can't do anything about it. And all we need to know to fix it is that KeyError.

datalad/customremotes/ria_utils.py
bpoldrack: Added comments to the constants, but made the `tuple` suggestion a TODO only. This would require a bit more RF'ing that shouldn't delay merge right now, I think. But: Certainly useful approach. However, may be better elsewhere (define versions as `(label, description)` and reference here by `label` only) - not sure yet.

bpoldrack:
@yarikoptic : 
See https://github.com/datalad/datalad/pull/4203#discussion_r419226947
Apart from that addressed your comments in https://github.com/datalad/datalad/pull/4203/commits/2d3edaae007f90ed219d3ebd41b28da518f1d0a1

PR #4211, opened 2020-02-26
NF: Parametrized Protocols with non-stdout/err return values for WitlessRunner

mih:
Changes:
- parameterized protocols are instrumental for passing handles to a protocol that can be used for specific communication via callbacks, or access to information that changes while the protocol is operating.
- move output stream decoding into Protocol class
- make WitlessRunner more ignorant of the actual nature of the Protocol return value, and allow for alternative implementations that are still compatible with WitlessRunner.

  The Protocol result is now always a dict that is expected to have keys 'code', 'stdout', 'stderr', matching the semantics of the CommandError exception.

- enhance CommandError to carry arbitrary additional results that may be reported by a protocol (e.g. stdout->parsed JSON), to avoid having to keep duplicates of the raw output strings

- taken together WitlessRunner can now drive Protocols that produce arbitrary additional results. Awareness of these additional results only needs to exist in "client-code", generic error handling needs no further customization.

### DIFF
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -249,14 +249,30 @@ class WitlessProtocol(asyncio.SubprocessProtocol):
             self.buffer[fd - 1].extend(data)
 
     def process_exited(self):
+        """Prepares the final result to be returned to the runner
+
+        Note for derived classes overwriting this method:
+
+        The result set for the `done` future must be a dict with keys
+        that do not unintentionally conflict with the API of
+        CommandError, as the result dict is passed to this exception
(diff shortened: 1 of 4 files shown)


mih:
Thx @kyleam ! Will update once original use case is confirmed possible with this feature.

mih:
@kyleam I adjusted the docs and made a few more changes to generalize protocol handling. There is also a small (unrelated) fix that I noticed to be needed when playing with the code.

mih:
Failure is known open files issue.

PR #4206, opened 2020-02-26
datalad.core.distributed.push

mih:
Approach is similar as before:

- do not attempt to build a 1:1 replacement of `publish`
- see what makes sense for the scope of a "plumbing" command
- focus on correct operation, not feature-wealth
- keep parallelization in mind
- hide as little as shallow as possible

Note that the tests here are affected (probabilisticly) by #4279 

Status quo:

- [X] basic functionality working across all platforms
- [x] support for managed branches
- [x] investigate speed-impact of `diff` not having `--report-filetype raw` capabilities ATM. If this is an issue, RF `_diff_cmd()` into a proper public function that we can use. it already exposes this parameter.
- [x] support for recursive operation
- [x] >90% test coverage
- [x] path handling uniform with `status` and `diff`. Fixes gh-1726 and fixes gh-4098
- [x] dedciated `action=publish` result for each pushed branch. Fixes gh-2000
- [x] dedicated error result when remote branch name cannot be determined. Fixes gh-2855
- [x] sync with remote annex branch after data transfer and before a push of the local git-annex branch to the remote. Fixes gh-2636
- [x] always publish the main branch first, and the git-annex branch last. Fixes gh-2001
- [x] no longer use `AnnexRepo.copy_to()`, but a dedicated async call with a protocol class implementing
annex JSON comminucation. Results are directly reported after normalization without intermediate inspection and decision making. Fixes gh-3412
- [x] if no branches are configured for push, the active branch (or rather its corresponding branch) is pushed by default. Fixes gh-1275
- [x] if branches are configured to be pushed by default, such configuration is honored. Fixes gh-4006
- [x] passes `--jobs` on to `git annex copy`. Fixes gh-3732
- [x] can push to an empty repository, regardless of without it is just a GitRepo, a (non-bare) AnnexRepo, or even a pseudo-empty AnnexRepo in adjusted-unlock mode (were initially an unrelated commit exists to mark the state prior to the created (empty) adjusted branch). Fixes gh-4074
- [x] recognized and supports non-annex datasets. Fixes gh-4082
- [x] reports an `impossible` result when content should be transferred, but isn't present locally. Fixes gh-3424
- [x] supports 'force' modes `datatransfer` and `nodatatransfer` (among other) that replace the clunky `--transfer-data` parameter of `publish`. Fixes gh-3414
- [x] rebase should #4228 be altered
- [x] rebase when #4229 is merged
- [x] PR cleanup

PR #4228, opened 2020-03-02
NF: WitlessProtocol for `annex ... --json` commands

mih:
that returns parsed JSON. No attempt is made (yet) to replace `_run_annex_command_json()`

`--json-progress` type messages are acted upon immediately (like in GitProgress), and are subsequently excluded from the normal JSON record reporting.

Split from #4206


mih:
Test failure is windows open file issue.

PR #4229, opened 2020-03-02
ENH: Expose diff_dataset() and add reporting_order parameter.

mih:
This make a faster diff with `eval_file_type=True` accessible to
internal users (like `push()`).

`reporting_order` enables more flexible applications. While the past and
current default `depth-first` is most suitable for human-readable
reporting, the alternative `breadth-first` enables straightforward
consolidation of all diff reports for a dataset (incl. all submodule
records) into an uninterrupted series of results. This simplies client
code that needs to investigate a complete changes set of a dataset, and
aims to yield its conclusions immediately, in order to kick of further
processing (generator-style) -- useful for effective parallelization.

Split from #4206 

TODO:

- [ ] add compatibility kludge for metalad

### DIFF
--- a/datalad/core/local/diff.py
+++ b/datalad/core/local/diff.py
@@ -277,7 +277,7 @@ def diff_dataset(
             origpaths=None if not path else OrderedDict(path),
             untracked=untracked,
             annexinfo=annex,
-            eval_file_type=True,
+            eval_file_type=eval_file_type,
             cache=content_info_cache,
             order=reporting_order):
         res.update(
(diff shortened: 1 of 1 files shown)


kyleam:
> This make a faster diff with `eval_file_type=True` accessible to
> internal users (like `push()`).

Makes sense, though there's a bug in the `eval_file_type` handling within `_diff_cmd` (now `diff_dataset`): `_diff_ds` is always called with `eval_file_type=True` regardless of what's passed in for `eval_file_type`.

kyleam:
> `reporting_order` enables more flexible applications [...]

LGTM

PR #4232, opened 2020-03-03
[WIP] testing pytest version of the tests

djarecka:
### Instructions

Please go through the following checklist before submitting the PR:

- [ ] Provide an overview of the changes you're making and explain why you're proposing them.

started converting tests to pytest version, see #4090 

- [ ] Include `Fixes #NNN` or `Closes #NNN` somewhere in the description if this PR addresses an existing issue.

- [ ] If this PR is not complete, select "Create Draft Pull Request" in the pull request button's menu.

  Consider using a task list (e.g., `- [ ] add tests ...`) to indicate remaining to-do items.

- [ ] If you would like to list yourself as a DataLad contributor and your name is not mentioned please modify .zenodo.json file.

- [ ] **Delete these instructions**. :-)

Thanks for contributing!

### DIFF
--- a/datalad/tests/utils_pytest.py
+++ b/datalad/tests/utils_pytest.py
@@ -1,0 +1,121 @@
+# emacs: -*- mode: python; py-indent-offset: 4; tab-width: 4; indent-tabs-mode: nil; coding: utf-8 -*-
+# ex: set sts=4 ts=4 sw=4 noet:
+# ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+#
+#   See COPYING file distributed along with the datalad package for the
+#   copyright and license terms.
+#
+# ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+"""Miscellaneous utilities to assist with testing"""
+
+import glob
+import gzip
+import inspect
+import shutil
+import stat
+from json import dumps
+import os
+import re
+import tempfile
+import platform
+import multiprocessing
+import logging
+import random
+import socket
+import textwrap
+import warnings
+from fnmatch import fnmatch
+import time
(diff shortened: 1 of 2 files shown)


datalad/tests/utils_pytest.py
```diff
@@ -0,0 +1,122 @@
...
+    ARCHIVES_TEMP_DIR,
+)
+
+import pytest
+
+
+# dj: if content is None, tmpdir.join("afile.txt") could be used (?)
+# this is a simplified version of with_tempfile
+@pytest.fixture(scope="function")
+def with_tempfile_pyt(tmpdir):
```
yarikoptic, line 102: I think, if I was eager to move over to pytest, I would not jump into replacing decorators with fixtures, but first made current decorators and their use "pytest friendly". 
E.g. converting our test args (which are passed from those decorators) into kwargs, so pytest doesn't consider them to be its magical fixture, which it would then fail to find.
then I think tests codebase in many cases (e.g. where we don't use generators tests) could stay both nose and pytest compatible, and only then, in one sweep (if we decide to move over to pytest) we could address generator tests etc - those which we cannot keep both nose and pytest "friendly".

yarikoptic:
I will close this one for now.  IMHO migration to `pytest` would be a herculean undertaking for a number of reasons (use of generators, heavy reliance on custom decorators interfering with fixtures, etc.; spread out effects through datalad extensions which reuse `datalad.tests.utils` and also use nose overall), thus nobody from core developers even expressed their intent to even attempt it.  If @djarecka or anyone else would like to continue on this endeavor, I would be happy to re-open this PR.

djarecka:
@yarikoptic - yes, i don't think I have time and energy to do it on my own. `datalad` is indeed not the easiest case for the migration ;-)

Issue #4233, opened 2020-03-03
datalad.tests.test_utils.test_get_open_files not sensible on Windows


mih:
We have been seeing this for a while now.

```
======================================================================
ERROR: datalad.tests.test_utils.test_get_open_files
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\hostedtoolcache\windows\Python\3.7.6\x64\lib\site-packages\nose\case.py", line 198, in runTest
    self.test(*self.arg)
  File "C:\hostedtoolcache\windows\Python\3.7.6\x64\lib\site-packages\datalad\tests\utils.py", line 586, in newfunc
    return t(*(arg + (d,)), **kw)
  File "C:\hostedtoolcache\windows\Python\3.7.6\x64\lib\site-packages\datalad\tests\test_utils.py", line 1206, in test_get_open_files
    eq_(get_open_files(subd)[op.realpath(subd)].pid, proc.pid)
KeyError: 'C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\datalad_temp_tree_xu_lia6j\\d'
```

Prior investigation in #4165 

This is due to the test assumptions being broken, not actual breakage in `get_open_file()`. On windows this function is extremely slow, but the test assumes that is could complete within 10s. To be on the save side, we would need to budget well over a minute on windows.

PR #4234, opened 2020-03-03
TST: Relax get_open_files() test on Windows due to performance issues.

mih:
Rational in the diff. We would need to budget ~1min per call to
get_open_files. This isn't worth it, given that this is only an
occasionally used test helper.

The test is now cut short on windows, only performing the timing
insensitive parts. It still runs for almost 3min...

Fixes gh-4233

### DIFF
--- a/datalad/tests/test_utils.py
+++ b/datalad/tests/test_utils.py
@@ -1190,6 +1190,16 @@ def test_get_open_files(p):
             os.getpid())
 
     assert not get_open_files(subd)
+
+    if on_windows:
+        # the remainder of the test assume a certain performance.
+        # however, on windows get_open_files() can be very slow
+        # (e.g. the first invocation in this test (above) can easily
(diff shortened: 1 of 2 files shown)


mih:
Test failure is due to the recently introduced https://github.com/datalad/datalad/pull/4228 -- which passed this morning, but does not seem to be reliable enough.

yarikoptic:
FWIW: https://stackoverflow.com/questions/45026542/why-is-psutil-process-iter-running-so-slowly  

"I just restarted the Windows 10 machine I was having this issue on, and now the code completes in less than 1 second. Go figure!"  But I guess it could be to lower number of processes after reboot. 
but also https://github.com/giampaolo/psutil/issues/1061 (closed without any resolution)

mih:
Github CI machines should be fresh, but presumably show the same behavior.

PR #4235, opened 2020-03-03
ENH: Extend rsync-like diff(path='subds/') to path constraints within the subds

mih:
When called with a trailing slash in a path argument matching a subdataset,
`diff()` will report on the content, not only on the subdataset record in
its superdataset -- even with `recursive=False`.

While this is not universally loved, it makes sense to not limit this
behavior to a single trailing slash.

When `subds/` reports on the entire subdataset, `subds/subdir` will now
only report on the content within `subdir` of `subds`.

Split from #4206


mih:
Test failure will be fixed with https://github.com/datalad/datalad/pull/4237

Issue #4236, opened 2020-03-03
AnnexJsonProtocol breaks when multiple records are read at once


mih:
I introduced that in a merge this morning, and now it fails occasionally in different individual travis runs, seemingly at random.

```
======================================================================
12221FAIL: datalad.support.tests.test_annexrepo.test_annexjson_protocol
12222----------------------------------------------------------------------
12223Traceback (most recent call last):
12224  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
12225    self.test(*self.arg)
12226  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/datalad/tests/utils.py", line 586, in newfunc
12227    return t(*(arg + (d,)), **kw)
12228  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/datalad/support/tests/test_annexrepo.py", line 2142, in test_annexjson_protocol
12229    eq_(len(orig_j), 2)
12230AssertionError: 0 != 2
```

I managed to replicate this locally. Seems to fail ~1/150 times on my machine.

When the issue happens the runner reports a successful `git annex find` return with

```
{'stdout': '', 'stderr': '', 'stdout_json': []}
```

My guess is that JSON decoding fails, and indeed:

```
b'{"bytesize":"8","mtime":"unknown","keyname":"dab741b6289e7dccc1ed42330cae1accc2b755ce8079c2cd5d4b5366c9f769a6","backend":"SHA256E","key":"SHA256E-s8--dab741b6289e7dccc1ed42330cae1accc2b755ce8079c2cd5d4b5366c9f769a6","humansize":"8 B","hashdirmixed":"F9/xf/","file":"dir1/file2","hashdirlower":"e94/f35/"}\n{"bytesize":"8","mtime":"unknown","keyname":"d0b425e00e15a0d36b9b361f02bab63563aed6cb4665083905386c55d5b679fa","backend":"SHA256E","key":"SHA256E-s8--d0b425e00e15a0d36b9b361f02bab63563aed6cb4665083905386c55d5b679fa","humansize":"8 B","hashdirmixed":"zP/JF/","file":"file1","hashdirlower":"bef/add/"}\n'
```

When it reports more than one record at once, it dies.

PR #4237, opened 2020-03-03
BF: Protect against multiple JSON records being read at once

mih:
Fixes gh-4236

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -3524,15 +3524,19 @@ class AnnexJsonProtocol(WitlessProtocol):
         # this is where the JSON records come in
         # json_loads() is already logging any error, which is OK, because
         # under no circumstances we would expect broken JSON
-        try:
-            j = json_loads(data)
-        except Exception:
-            # TODO turn this into an error result, or put the exception
-            # onto the result future -- needs more thought
-            if data.strip():
-                # do not complain on empty lines
-                lgr.error('Received undecodable JSON output: %s', data)
-            return
+        for line in data.splitlines():
+            try:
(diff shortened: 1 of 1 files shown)


datalad/support/annexrepo.py
```diff
@@ -3524,15 +3524,20 @@ def pipe_data_received(self, fd, data):
...
-                lgr.error('Received undecodable JSON output: %s', data)
-            return
+        for line in data.splitlines():
+            try:
+                j = json_loads(line)
+            except Exception:
+                # TODO turn this into an error result, or put the exception
+                # onto the result future -- needs more thought
+                if line.strip():
+                    print("FUCKINGDECODE", data)
```
yarikoptic: leftover

mih:
Those are rather curious CI failures. Even more so, as they are Windows only.

kyleam:
> Those are rather curious CI failures. Even more so, as they are Windows only.

Comparing several PR that now show these failures to recent PRs that don't, the main difference I'm spotting is that the failing ones seem to be using Git 2.25.1, but all the recent passing ones I've been able to find use 2.25.*0*.  Seems pretty unlikely that that would matter, though.

kyleam:
> the main difference I'm spotting is that the failing ones seem to be using Git 2.25.1

I tried tweaking some of the config variables specific to Git for Windows, in particular `core.fscache` based on it being mentioned in 2.25.1.windows.1 release notes.  That didn't resolve these failures.

mih:
Given that this PR fixes a real issue that affects real systems I will merge this now and worry about the GitHub Windows CI oddities later/separately. Thx @kyleam for looking into this!

PR #4239, opened 2020-03-03
BF: Allow to explicitly specify host to aws-s3 resource

yarikoptic:
See first commit (cc7dbd5ac0f2722a6056e1ce4149d535d1bb080a ATM) for details

### DIFF
--- a/datalad/downloaders/s3.py
+++ b/datalad/downloaders/s3.py
@@ -46,10 +46,21 @@ class S3Authenticator(Authenticator):
     allows_anonymous = True
     DEFAULT_CREDENTIAL_TYPE = 'aws-s3'
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args, host=None, **kwargs):
+        """
+
+        Parameters
+        ----------
+        host: str, optional
+          In some cases it is necessary to provide host to connect to. Passed
+          to boto.connect_s3
+        """
         super(S3Authenticator, self).__init__(*args, **kwargs)
         self.connection = None
         self.bucket = None
+        self._conn_kwargs = {}
+        if host:
+            self._conn_kwargs['host'] = host
 
     def authenticate(self, bucket_name, credential, cache=True):
         """Authenticates to the specified bucket using provided credentials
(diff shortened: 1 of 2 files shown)


yarikoptic:
poor windowses

datalad/downloaders/s3.py
```diff
@@ -46,10 +46,21 @@ class S3Authenticator(Authenticator):
     allows_anonymous = True
     DEFAULT_CREDENTIAL_TYPE = 'aws-s3'
 
-    def __init__(self, *args, **kwargs):
-        super(S3Authenticator, self).__init__(*args, **kwargs)
+    def __init__(self, host=None, **kwargs):
```
kyleam: Why do the unrelated drop of `*args` here?  While it probably doesn't matter, I don't see any reason to make that backwards-incompatible change.

datalad/downloaders/tests/test_s3.py
```diff
@@ -120,3 +130,31 @@ def deny_access(*args, **kwargs):
     with assert_raises(DownloadError):
         with patch.object(downloader, '_download', deny_access):
             downloader.download("doesn't matter")
+
+
+@with_tempfile
+def test_boto_host_specification(tempfile):
+    # This test relies on a yoh-specific set of credentials to access
```
kyleam: While this of course is useful on your end, my preference would be to not bring this into the main tree.

yarikoptic: where/how to keep it so I don't forget and test whenever I do change anything in related code?

datalad/downloaders/s3.py
yarikoptic: fair enough -- will RF to use `kwargs.pop`. In the longer run -- the whole thing could be RFed to become more explicit/consistent.

kyleam: > will RF to use `kwargs.pop`

All right, though keeping `*args` around doesn't require doing that.

yarikoptic: ah, right -- python3 !

datalad/downloaders/tests/test_s3.py
kyleam: If having an appropriately named local branch isn't enough, you could leave a one-line comment in this file as a note for yourself.  Anyway, just my two cents.  Feel free to disregard.

datalad/downloaders/s3.py
yarikoptic: ok, this one is done

PR #4243, opened 2020-03-05
RF: Dedicated method to perform a local annex sync

mih:
Per request in https://github.com/datalad/datalad/pull/4206#discussion_r387201686

To consolidate corresponding branches and sync 'git-annex' branches of
one or more given, or all remotes. Does not leave behind 'synced/...'
branches, unless they existed prior calling the method.

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -3503,13 +3503,14 @@ class AnnexRepo(GitRepo, RepoInterface):
         """
         branch = self.get_active_branch()
         corresponding_branch = self.get_corresponding_branch(branch)
-        if branch == corresponding_branch:
-            # nothing to sync
-            return
+
+        have_corresponding_branch = branch != corresponding_branch
 
         lgr.debug(
-            "Git-annex adjusted branch detected, sync'ing %s",
-            corresponding_branch)
+            "Sync local 'git-annex' branch%s%s%s.",
+            ", and corresponding '" if have_corresponding_branch else '',
+            corresponding_branch if have_corresponding_branch else '',
+            "' branch")
 
         synced_branch = 'synced/{}'.format(branch)
         had_synced_branch = synced_branch in self.get_branches()
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -1843,19 +1843,8 @@ class AnnexRepo(GitRepo, RepoInterface):
                 raise e
         self.config.reload()
 
-    def merge_annex(self, remote=None):
-        """Merge git-annex branch
-
-        Merely calls `sync` with the appropriate arguments.
-
-        Parameters
-        ----------
-        remote: str, optional
-          Name of a remote to be "merged".
-        """
-        self.sync(
-            remotes=remote, push=False, pull=False, commit=False, content=False,
-            all=False)
+    def merge_annex(self, remote=None):  # do not use anymore, use localsync()
+        self.localsync(remote)
 
     def sync(self, remotes=None, push=True, pull=True, commit=True,
              content=False, all=False, fast=False):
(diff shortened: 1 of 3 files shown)


datalad/support/annexrepo.py
```diff
@@ -3476,24 +3476,43 @@ def _save_post(self, message, status, partial_commit):
...
-        adjusted_match = ADJUSTED_BRANCH_EXPR.match(branch if branch else '')
-        if adjusted_match:
-            orig_branch = adjusted_match.group('name')
-            synced_branch = 'synced/{}'.format(orig_branch)
-            had_synced_branch = synced_branch in self.get_branches()
-            lgr.debug(
-                "Git-annex adjusted branch detected, sync'ing %s",
-                orig_branch)
+        corresponding_branch = self.get_corresponding_branch(branch)
+        if branch != corresponding_branch:
```
kyleam: Shouldn't this be `==`?

mih: Yes, and another conditional was also reversed :man_facepalming: now fixed.

datalad/support/annexrepo.py
```diff
@@ -3476,24 +3465,56 @@ def _save_post(self, message, status, partial_commit):
         # first do standard GitRepo business
         super(AnnexRepo, self)._save_post(
             message, status, partial_commit)
-        # check if we have to deal with adjusted branch
+        # then sync potential managed branches
+        self.localsync()
```
kyleam: The "is corresponding branch" conditional is now removed from `localsync`, so shouldn't `_save_post` do it to avoid unnecessary `sync` calls?

datalad/distribution/update.py
```diff
@@ -394,9 +394,7 @@ def _plain_merge(repo, _, target, merge_opts=None):
 
 def _annex_plain_merge(repo, _, target, merge_opts=None):
     yield from _plain_merge(repo, _, target, merge_opts=merge_opts)
-    # Note: Avoid repo.merge_annex() so we don't needlessly create synced/
-    # branches.
-    repo.call_git(["annex", "merge"])
+    repo.localsync()
```
kyleam: Hmm, I still prefer `annex merge` because all this is trying to do is trigger a merge of the `git-annex` branches.  (Which means I didn't leave a very good comment because it only gave a superficial reason for avoing the `git annex sync`-based `AnnexRepo.merge_annex()`.)

mih: OK, will revert.

mih: Done

datalad/support/annexrepo.py
mih: Makes sense. Although I spent a day trying graps all the combinations, and I am still not done. So....we'll see....

mih: I will introduce a `managed_only=False` parameter that allows for triggering the test from the outside.

kyleam: Sounds good.  And it looks like being able to avoid `annex sync` is necessary to prevent merge conflict failures like the one now happening in test_rerun.  Here's a reduced script that will trigger such a conflict without datalad:

<details>
<summary>conflict demo</summary>

```sh
#!/bin/sh

set -eux

cd "$(mktemp -d --tmpdir gx-XXXXXXX)"
git init
git annex init

echo foo >foo
git annex add foo
git commit -mfoo

git annex unlock foo
echo more >>foo
git annex add foo
git commit -m"foo more"

git annex sync --no-pull --no-push --no-commit --no-resolvemerge

git reset --hard HEAD~1
git annex unlock foo
echo different more >>foo
git annex add foo
git commit -m"foo different more"

git annex sync --no-pull --no-push --no-commit --no-resolvemerge
```

</details>

mih: I pushed that change. Need to detach now till some time next week.

mih:
> LGTM (and sorry my suggestion in [gh-4206](https://github.com/datalad/datalad/pull/4206) led to so much work :/)

I think this is now in a better place, so keep those coming! ;-)

Issue #3824, opened 2019-10-22
`git annex sync` crash with submodule on windows


mih:
@joeyh It looks as if it doesn't anticipate a submodule being present. 

Demo:

```
(base) C:\Users\mih>datalad create demo
[INFO] Creating a new annex repo at C:\Users\mih\demo
[INFO]   Detected a filesystem without fifo support.
[INFO]   Disabling ssh connection caching.
[INFO]   Detected a crippled filesystem.
[INFO]   Entering an adjusted branch where files are unlocked as this filesystem does not support locked files.
[INFO] Switched to branch 'adjusted/master(unlocked)'
create(ok): C:\Users\mih\demo (dataset)

(base) C:\Users\mih>cd demo

(base) C:\Users\mih\demo>datalad create -d . sub
[INFO] Creating a new annex repo at C:\Users\mih\demo\sub
[INFO]   Detected a filesystem without fifo support.
[INFO]   Disabling ssh connection caching.
[INFO]   Detected a crippled filesystem.
[INFO]   Entering an adjusted branch where files are unlocked as this filesystem does not support locked files.
[INFO] Switched to branch 'adjusted/master(unlocked)'
create(ok): sub (dataset)
action summary:
  add (ok: 2)
  create (ok: 1)
  save (ok: 1)

(base) C:\Users\mih\demo>git status
On branch adjusted/master(unlocked)
nothing to commit, working tree clean

(base) C:\Users\mih\demo>git annex sync
commit
On branch adjusted/master(unlocked)
nothing to commit, working tree clean
ok
fatal: entry 'sub' object type (blob) doesn't match mode type (commit)
git-annex: user error (git ["--git-dir=.git","--work-tree=.","--literal-pathspecs","mktree","--batch","-z"] exited 128)

```

joeyh:
Michael Hanke wrote:
> fatal: entry 'sub' object type (blob) doesn't match mode type (commit)
> git-annex: user error (git ["--git-dir=.git","--work-tree=.","--literal-pathspecs","mktree","--batch","-z"] exited 128)

I found a bug report about that:

https://git-annex.branchable.com/bugs/git-annex-sync_sometimes_fails_in_submodule_in_V6_adjusted_branch/

Please followup there, if you have any way to help me reproduce it.

To my releif, I don't think this is specific to windows..

-- 
see shy jo

mih:
Thanks @joeyh

joeyh:
[...]

PR #4245, opened 2020-03-06
TST: Add Github CI workflow to run core tests on a crippled FS

mih:
This aims to disentangle detection of issues with Windows and issues
with adjusted branch operation. There is no expectation that a run
would initially succeed... even on Linux.

Appveyor and Travis runs will be canceled.

TODO

- [ ] VFAT or NTFS?

This seems to be a great success. In 8min we get to see that a large chunk of tests don't work out:

```
Ran 117 tests in 481.873s
FAILED (SKIP=6, errors=43, failures=5)
```

The most popular error is

```
'git annex sync --no-commit --no-push --no-pull' failed with exitcode 1 under /crippledfs/datalad_temp_check_create_path_semantics1p3hq_fs [err: 'fatal: entry '<imagine subdataset path here>' object type (blob) doesn't match mode type (commit)
```

which I feel has been reported before.

### DIFF
--- a/.github/workflows/test_crippled.yml
+++ b/.github/workflows/test_crippled.yml
@@ -1,0 +1,48 @@
+name: CrippledFS
+
+on: [pull_request]
+
+jobs:
+  test:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Set up system
+      shell: bash
+      run: |
+        bash <(wget -q -O- http://neuro.debian.net/_files/neurodebian-travis.sh)
+        sudo apt-get update -qq
+        sudo apt-get install eatmydata
+        sudo eatmydata apt-get install git-annex-standalone dosfstools
+        # 500 MB VFAT FS in a box
+        sudo dd if=/dev/zero of=/crippledfs.img count=500 bs=1M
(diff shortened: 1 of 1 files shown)


kyleam:
> This aims to disentangle detection of issues with Windows and issues
> with adjusted branch operation

Thanks.  I often try to figure this out locally (by either using Yarik's handy `tools/eval_under_testloopfs` script or by setting ` annex.crippledfilesystem=true` temporarily in my global config), so I'd find this additional run helpful.

> which I feel has been reported before.

An error with the same message was reported in gh-3824.  It seems like that particular case was fixed on git-annex's side.

kyleam:
> An error with the same message was reported in [gh-3824](https://github.com/datalad/datalad/issues/3824). It seems like that particular case was fixed on git-annex's side.

Or rather, the git-annex in neurodebian that is being pulled in for this run doesn't yet have that fix.  The fix was first released with 7.20191024, while this run is using 7.20190819+git2-g908476a9b.

mih:
If only Github would offer a Debian unstable runner, we could be tracking git-annex more closely. I have started to enjoy the speed of the non-standalone git-annex.

PR #4250, opened 2020-03-06
ENH: run github workflows on push as well

yarikoptic:
I was confused to not see any "master" builds in the github actions.  That
precluded to see what last state of master was ok on windows.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -1,6 +1,6 @@
 name: Benchmarks
 
-on: [pull_request]
+on: [push, pull_request]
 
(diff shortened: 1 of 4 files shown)


yarikoptic:
I will cancel travis runs since unrelated

mih:
Should we be concerned about using even more of the little free time we have for near-duplicate test runs?

yarikoptic:
I don't know.  Looking at https://github.com/organizations/datalad/settings/billing  - they seems to account only for Private repos ATM.  We never hit the limit AFAIK.  When we do - we could start worrying/mitigating (external workers).

PR #4252, opened 2020-03-06
A more sane post-clone setup for adjusted-branch repos

### LINKED ISSUE #4227: `clone()` configures adjusted branch as 'upstream'
Cloning a dataset that has an adjusted master, yields this configuration in the clone:

```
branch.adjusted/master(unlocked).remote=origin
branch.adjusted/master(unlocked).merge=refs/heads/adjusted/master(unlocked)
```

This will make any `git push` push something that is being rebased all the time back to `origin`. This is likely to fail, and be rather confusing, as git will recommend conflict resolution approaches that are all wrong (actually needs `annex sync`). With this config removed, things will go back to 

```
% git push --dry-run --porcelain      
fatal: The current branch adjusted/master(unlocked) has no upstream branch.
To push the current branch and set the remote as upstream, use

    git push --set-upstream origin adjusted/master(unlocked)
```

which seems better. Confusingly, however, the adjusted branch will still be pushed to any other remote:

```
% git push --dry-run --porcelain target
To /media/mih/Samsung_T5/tmp/datalad_temp_test_push_recursiveu5n5sxxi
*       refs/heads/adjusted/master(unlocked):refs/heads/adjusted/master(unlocked)       [new branch]
Done
```

Datalad should actively prevent that.

mih:
Summary: A clone of a dataset in adjusted mode no longer ends up being in adjusted mode too, by default -- unless the target filesystem actually imposes this constraint.

fixes #4227

### DIFF
--- a/datalad/core/distributed/tests/test_clone.py
+++ b/datalad/core/distributed/tests/test_clone.py
@@ -190,6 +190,14 @@ def test_clone_simple_local(src, path):
         uuid_before = ds.repo.uuid
         eq_(ds.repo.get_description(), 'mydummy')
 
+    # make sure we do not have any adjusted branch configured
+    # for push or as destination
+    for v in ('branch.adjusted/master(unlocked).remote',
+              'branch.adjusted/master(unlocked).merge'):
+        assert_not_in(v, ds.config)
+    eq_(ds.config.get('branch.master.remote'), 'origin')
+    eq_(ds.config.get('branch.master.merge'), 'refs/heads/master')
+
     # installing it again, shouldn't matter:
     res = clone(src, path, result_xfm=None, return_type='list')
     assert_result_values_equal(res, 'source_url', [src])
--- a/datalad/distribution/tests/test_update.py
+++ b/datalad/distribution/tests/test_update.py
@@ -206,8 +206,8 @@ def test_update_fetch_all(src, remote_1, remote_2):
     rmt2 = AnnexRepo.clone(src, remote_2)
 
     ds = Dataset(src)
-    ds.siblings('add', name="sibling_1", url=remote_1)
-    ds.siblings('add', name="sibling_2", url=remote_2)
+    ds.siblings('add', name="sibling_1", url=remote_1, result_renderer=None)
+    ds.siblings('add', name="sibling_2", url=remote_2, result_renderer=None)
 
     # modify the remotes:
     with open(opj(remote_1, "first.txt"), "w") as f:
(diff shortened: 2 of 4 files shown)


kyleam:
The `test_update` failure makes sense.  The failing test specifically puts the source repository into an adjusted state and then of course expects that the clone is also in an adjusted state.  With this PR, the clone no longer is.  So, okay, modify that test to also put the clone in an adjusted state (given that that is the state we're interested in testing).  Something like

```diff
diff --git a/datalad/distribution/tests/test_update.py b/datalad/distribution/tests/test_update.py
index 2b7a5a1a56..2b8fe90266 100644
--- a/datalad/distribution/tests/test_update.py
+++ b/datalad/distribution/tests/test_update.py
@@ -721,6 +721,10 @@ def check_merge_follow_parentds_subdataset_detached(on_adjusted, path):
     ds_clone = install(source=ds_src.path, path=path / "clone",
                        recursive=True, result_xfm="datasets")
     ds_clone_s1 = Dataset(ds_clone.pathobj / "s0" / "s1")
+    if on_adjusted:
+        _adjust(ds_clone.repo)
+        _adjust(ds_clone_s1.repo)
+        ds_clone.save(recursive=True)
 
     ds_src_s1.repo.checkout("master^0")
     (ds_src_s1.pathobj / "foo").write_text("foo content")
```

Then that second adjust call fails because `.repo` for the second-level subdataset is `None` on crippled file systems.  `.repo` is `None` because this change deletes the local adjusted branch, but then a new adjusted branch is created.  That gets into a confused state where the clone has the first-level submodule registered at a commit that lacks the second-level submodule, so the `.update_submodule()/git submodule update` call puts the hierarchy into a state where the second-level submodule doesn't exist.  Some of the above might not be an entirely accurate description, but the overall effect is that `install` does not register the expected dataset hierarchy with this change.

If you were to not delete the initial local branch

```diff
diff --git a/datalad/support/gitrepo.py b/datalad/support/gitrepo.py
index ea3d92e0c2..d20bb711ba 100644
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -1165,7 +1165,6 @@ def clone(cls, url, path, *args, clone_options=None, **kwargs):
                 '-t',
                 remote_corresponding_branch])
             # remove local adjusted branch
-            gr.call_git(['branch', '-d', active_branch])
 
         if fix_annex:
             # cheap check whether we deal with an AnnexRepo - we can't check the class of `gr` itself, since we then
```

then the install now works (at least in this particular scenario), but the `sync` call by `update` fails with

```
datalad.support.exceptions.CommandError: CommandError: 'git-annex sync -c annex.dotfiles=true --no-commit --no-content --no-push --pull origin' failed with exitcode 1 under /tmp/datalad_temp_check_merge_follow_parentds_subdataset_detachedbp4g_gki/clone [out: 'pull origin 
failed'] [err: 'git-annex: sync: 1 failed']
```

So, somehow I guess the checkout/configuration from this PR affects the state that `sync` relies on.

I don't think this `update` test itself is problematic, though.  I think it is just one of the few places that explicitly checks the adjusted branch behavior, giving a glimpse of the adjusted branch/submodule brokenness.

mih:
Thx @kyleam ! I think your analysis was pointing to the right issue, but the fix was missing to put the intermediate "s0" subdataset into adjusted mode as well. It now passes for me locally.

mih:
The remaining test failure indicates that this change cannot be made independent of #4275. I will merge the two PRs.

PR #4257, opened 2020-03-07
ENH: Complain like Git if to-be-removed sibling is unknown

### LINKED ISSUE #4249: Unlike Git, Datalad doesn't complain when removing an unknown sibling
I'm not sure if its known or on purpose, but I found that datalad stays silent when I attempt to remove a sibling that does not exist - Git on the other hand talks to me when I remove a non-existent remote.

```sh
(handbook) ╭─adina@muninn /tmp/ds on master
╰─➤ datalad siblings remove -s nonexistent
(handbook) ╭─adina@muninn /tmp/ds on master
╰─➤ git remote remove nonexistent
fatal: No such remote: 'nonexistent'
```

Personally, I would find it helpful if DataLad says something similar to Git, but that's just a subjective opinion :)

adswa:
fixes #4249

### DIFF
--- a/datalad/distribution/siblings.py
+++ b/datalad/distribution/siblings.py
@@ -301,6 +301,11 @@ class Siblings(Interface):
     @staticmethod
     def custom_result_renderer(res, **kwargs):
         from datalad.ui import ui
+        # should we attempt to remove an unknown sibling, complain like Git does
+        if res['status'] == 'notneeded' and res['action'] == 'remove-sibling':
+            ui.message(
+                'Warning: No sibling "{name}" in dataset {path}'.format(**res))
+            return
         if res['status'] != 'ok' or not res.get('action', '').endswith('-sibling') :
             # logging complained about this already
             return
--- a/datalad/distribution/siblings.py
+++ b/datalad/distribution/siblings.py
@@ -74,6 +74,8 @@ from .dataset import (
     EnsureDataset,
     datasetmethod,
 )
+import datalad.support.ansi_colors as ac
+
 
 lgr = logging.getLogger('datalad.distribution.siblings')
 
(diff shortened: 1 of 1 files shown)


adswa:
As far as I can see, the failing tests don't seem to be related to the change in this PR.

datalad/distribution/siblings.py
```diff
@@ -301,6 +301,10 @@ def __call__(
     @staticmethod
     def custom_result_renderer(res, **kwargs):
         from datalad.ui import ui
+        # should we attempt to remove an unknown sibling, complain like Git does
+        if res['status'] == 'notneeded' and res['action'] == 'remove-sibling':
+            ui.message('fatal: No such remote: {name}'.format(**dict(res)))
```
kyleam: Thanks, this looks like an improvement to me.  I have the following suggestions, some nit-picks:

  * You could just use `**res` instead of passing `res` to `dict`.

  * While I think we should complain in the spirit of `git remote rm`, I think it'd be better to not complain _exactly_ like it.  In particular, I'd say we should drop "fatal: " and replace "remote" with "sibling".

  * The message should include some indication of the path so that recursive operation gives a clearer message:

    ```
    % datalad subdatasets
    subdataset(ok): subds (dataset)
    % datalad --on-failure=continue siblings -r remove -s nonexistent
    fatal: No such remote: nonexistent
    fatal: No such remote: nonexistent
    ```

kyleam:
> As far as I can see, the failing tests don't seem to be related to the change in this PR...

Yep, this is unlikely to cause issues given that functionally this is just adding a `ui.message` call.  I've gone through the failures and those all look like ones we're seeing elsewhere.

datalad/distribution/siblings.py
adswa: Thanks for these comments, I think I have addressed them now.

adswa:
uhhh,cool, color!

mih:
Travis failure is unrelated and there is no chance that the type of change I made would fail in this test run only.

PR #4258, opened 2020-03-07
RF: Reduce use of "ria" label

mih:
Remove entirely when not necessary. If necessary 'ria' refers to the
concept of a store, whether the datasets are addressable by their
dataset ID.

A further renaming will happen as part of gh-4164, where 'ria-remote'
(the special remote implementation) is replaced by 'ora-remote'.

Together, this changes disambiguate the term 'ria'.

This change also removes TODOs from the documentation.


mih:
Typo fixed. Thx @adswa

Issue #4259, opened 2020-03-07
git-annex no longer provides remote name with GETCONFIG name


mih:
In the ria special remote implementation we are using AnnexRemote with

```
name = self.annex.getconfig('name')
```
to issue a `GETCONFIG name` request to to git-annex.

This used to give the `name` of the special remote as given to `git annex initremote`, and is confirmed to work as of git-annex 7.20191230.

With git-annex 8.20200226 the response to this request (while `git annex initremote` is driving the special remote) is empty. This breaks our special remote, as it needs the name to look up additional configuration.

@joeyh was this an intentional change? I could not locate related topic on the git-annex changelog other than the introduction of LISTCONFIGS -- which should not break prior behavior, as far as I can see.

joeyh:
This change was part of the remote config parsing. It now separates
the configs used by the special remote with ones used by git-annex
itself.

I had not considered that they may look at their name or other configs
like that. Will need to revisit that change. Could you please file a
bug report?

-- 
see shy jo

mih:
Done, thx!

https://git-annex.branchable.com/bugs/git-annex_no_longer_provides_remote_name_with_GETCONFIG_name

mih:
Fixed, thx @joeyh !

https://git-annex.branchable.com/bugs/git-annex_no_longer_provides_remote_name_with_GETCONFIG_name/#comment-12a9c26a4f22d0e792058ed48f6c2e52

aqw:
@bpoldrack pointed me in the direction of this issue (and looks similar to #4509). If it's the wrong one, please let me know, and I will deflect all blame to him. ;-)

I seem to have re-encountered this bug with version `8.20201127`

```
mkdir watch_it_burn
cd watch_it_burn
datalad create .
datalad create -d . www
cd www
datalad create-sibling-ria -r --name inm7 --storage-name "inm7-ssh" "ria+ssh://debian_inm7.ds.inm7.de/ds/debian_inm7/srv"
git annex initremote inm7-ro url="ria+https://debian_inm7.ds.inm7.de" autoenable=true encryption=none externaltype=ora type=external --sameas inm7-ssh
```

(I don't pass `archive-id` to the `initremote` to avoid a second bug, which @bpoldrack is fixing as we speak.)

<details><summary>datalad wtf</summary>
<p>

```
❱ datalad wtf
# WTF
## configuration <SENSITIVE, report disabled by configuration>
## datalad 
  - full_version: 0.13.7
[...]

PR #4260, opened 2020-03-07
ENH: Move git-annex-ria-remote to datalad core as git-annex-ora-remote (2nd attempt)

mih:
This PR sits on top of #4258, and incorporates and replaces #4164.

Major changes in addition to those two base PRs are:

- [x] Adopt desired package layout
- [x] Tests and `ria_export_archive()` work on crippled FS
- [x] Tests are sped up
- [x] File headers and clean imports
- [x] Anticipate #4259 and fail, instead of proceeding without configuration


mih:
OK, I will merge this now. I realize that this is a bit forced, but having this hanging in a PR complicates related efforts substantially. The code should be organized in a way that its impact on existing functionality should be minimal. You can rightfully blame any fallout on me, and I promise to take care of it.

PR #4264, opened 2020-03-09
TST: Pin git-annex version to 7.20191107 for windows CI runs

mih:
Version was picked, because it happened to have been installed on @mih debug win10 box.

As expressed elsewhere, datalad+git-annex on windows is moving to fast as a target for me to ever catch up. I'd rather have some things working reasonably well with a particular annex version that nothing with the latest version. Also, the noise contributed by the windows tests is hurting, because they make devs more likely to ignore actual failures caused by regressions.

Will cancel Travis run.

### DIFF
--- a/.github/workflows/test_win2019.yml
+++ b/.github/workflows/test_win2019.yml
@@ -31,7 +31,10 @@ jobs:
         python-version: 3.7
     - name: Set up git-annex
       run: |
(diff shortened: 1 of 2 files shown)


yarikoptic:
cool. after we regain stable grounds, I will add testing against "default install" git-annex for windows into `datalad-extensions` to be done on cron, so we could have an idea without explicit PRs

PR #4266, opened 2020-03-09
TST: Make removal of test temp artifacts more robust

mih:
This should mitigate the effect of not yet finished git-annex batch
processes. No idea if 10s is an effective time range, but we got to
start somewhere.

Update: even with 10s across 5 attempts it is still happening.

### DIFF
--- a/datalad/utils.py
+++ b/datalad/utils.py
@@ -542,11 +542,29 @@ def rmtemp(f, *args, **kwargs):
             lgr.debug("Path %s does not exist, so can't be removed" % f)
             return
         lgr.log(5, "Removing temp file: %s" % f)
-        # Can also be a directory
-        if os.path.isdir(f):
-            rmtree(f, *args, **kwargs)
-        else:
-            unlink(f)
+        # we have issues with not-yet closed git-annex processes.
+        # but a timing issue and a conceptual test failure are two different
+        # things, hence spend a few attempts and a couple of seconds to
+        # let an issue resolve itself
+        removed = False
+        last_exc = None
+        for attempt in range(5):
+            try:
+                # Can also be a directory
+                if os.path.isdir(f):
+                    rmtree(f, *args, **kwargs)
+                else:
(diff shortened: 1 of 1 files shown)


datalad/utils.py
```diff
@@ -542,11 +542,29 @@ def rmtemp(f, *args, **kwargs):
...
+                removed = True
+                break
+            except PermissionError as e:
+                lgr.warn(
+                    'Attempt %i to remove temp location failed: %s',
+                    attempt + 1, e)
+                last_exc = e
+                time.sleep(2)
+        if not removed:
+            raise last_exc
```
yarikoptic: note -- we have `try_multiple` and `@try_multiple_dec` , and I thought that we already use them:
```
datalad/utils.py:@try_multiple_dec
datalad/utils.py-def unlink(f):
datalad/utils.py-    """'Robust' unlink.  Would try multiple times
--
datalad/utils.py:@try_multiple_dec
datalad/utils.py-def _rmtree(*args, **kwargs):
datalad/utils.py-    """Just a helper to decorate shutil.rmtree.
```
so you just need to use "our" `_rmtree` and `unlink` here

mih: But it is using `rmtree` from `datalad.utils`, which is our `rmtree`, and it does use `_rmtree`. But it "just" doesn't work ;-)

yarikoptic:
Oops... Somehow managed to kill the comment thread we had. Overall point was - we already try me times using try_multiple, trying even more on top send to be off no help, so pr could just be closed?

mih:
I am afraid that this is a valid conclusion. So we are left with adding explicit close(). calls to all tests that utilize AnnexRepo instances directly.

yarikoptic:
Since we should have access to all *Repo instances due to flyweight pattern, if we could make nose call some ".close_all()" method upon completion of each test (didn't find a quick way to accomplish that in nose) we could avoid hand tuning each test

PR #4268, opened 2020-03-09
More crucial testing on Windows

mih:


mih:
Good on Windows, no worse elsewhere. Going in.

PR #4273, opened 2020-03-10
NF: Factory function to get Repo instances from a path

mih:
And RF Dataset.repo to use it.

This is needed for work in GitRepo, where type-specific prcoessing of
submodules is needed.


mih:
This is a relatively simple, but useful change. Will merge now.

PR #4274, opened 2020-03-10
RF: AnnexRepo.get_corresponding_branch() -> None without a managed branch

mih:
When dealing with managed branches, one frequently has to switch
behavior depending on their presence. With this change the return value
of `get_corresponding_branch()` carries the information whether there is
such a branch or not, and immediately allows decision making based on
it.

Previously, it required an additional call to `is_managed_branch()`
(that is duplicated in `get_corresponding_branch()` in order to determine
this condition.

This cost reduction is minimal, but the code complexity goes down.
The previous behavior is easily achieved via

    `get_corresponding_branch(branch) or branch`

and seem to communicate more direct what is intended.

A `GitRepo.get_corresponding_branch()` that always returns `None` has been
added to further simplify decision making and avoid type-based
switching.

### DIFF
--- a/datalad/distribution/tests/test_update.py
+++ b/datalad/distribution/tests/test_update.py
@@ -756,8 +756,7 @@ def check_merge_follow_parentds_subdataset_detached(on_adjusted, path):
     # the explicit revision fetch fails.
     (ds_src_s1.pathobj / "bar").write_text("bar content")
     ds_src.save(recursive=True)
-    ds_src_s1.repo.checkout(
-        ds_src_s1.repo.get_corresponding_branch("master"))
+    ds_src_s1.repo.checkout('master')
     # This is the default, but just in case:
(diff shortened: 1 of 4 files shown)


datalad/distribution/tests/test_update.py
```diff
@@ -756,8 +756,7 @@ def check_merge_follow_parentds_subdataset_detached(on_adjusted, path):
     # the explicit revision fetch fails.
     (ds_src_s1.pathobj / "bar").write_text("bar content")
     ds_src.save(recursive=True)
-    ds_src_s1.repo.checkout(
-        ds_src_s1.repo.get_corresponding_branch("master"))
+    ds_src_s1.repo.checkout('master')
```
kyleam: Eh, that was silly of me.  I guess in some earlier iteration that was a plain `.get_corresponding_branch()`, and I didn't drop the call when I switched to explicitly using "master", or something :x

mih:
Thanks @kyleam

PR #4275, opened 2020-03-10
ENH: Record state of a submodule's corresponding branch in the parent

mih:
Otherwise any subdataset in adjusted mode will have a state recorded
that belongs to a branch that is continuously rebased.

Also adjust DataLad's own diff-functionality to detect this intentional
'modified' state and report it as 'clean'.

This is a major step toward a first serious attempt to support nested
dataset in adjusted mode

Sitting on top of #4273 #4274


datalad/support/gitrepo.py
```diff
@@ -3705,19 +3742,40 @@ def _get_cache_key(label, paths, ref, untracked=None):
...
+                # 2) run an unconditional localsync() and therefor
+                #    enforce a clean diff between adjusted and
+                #    corresponding branch.
+                # Going with (2) due to implementation simplicity, but
+                # without having tested performance differences on complex
+                # datasets
+                lgr.debug(
+                    'Sync corresponding branch %s at %s prior repository '
+                    'state evaluation', subm_cbranch, f)
+                subrepo.localsync(managed_only=True)
```
mih: Can we be cheaper than this, especially if we just had sync'ed before? Should we make an attempt to count the commits on the adjusted branch, and if there is only one not go for a sync. This would only miss cases, where someone amends commits on the adjusted branch, but that would be problematic behavior anyways, so maybe that is good enough. OTOH, I dunno how clever git-annex is in figuring out that nothing needs to be done....

datalad/support/gitrepo.py
```diff
@@ -2825,10 +2832,26 @@ def add_submodule(self, path, name=None, url=None, branch=None):
                 url = path
         cmd += [url, Path(path).as_posix()]
         self.call_git(cmd)
+
+        if subm is None:
+            # in case it did not happen above
+            subm = repo_from_path(self.pathobj / path)
```
kyleam: Why not just replace `subm = None` with this (and remove the assignment from under `if url is None`)?

datalad/support/gitrepo.py
```diff
@@ -3933,6 +3991,10 @@ def save_(self, message=None, paths=None, _status=None, **kwargs):
                         str(cand_sm.relative_to(self.pathobj)),
                         url=None,
                         name=None,
+                        # MIH: this looks strange, why would we want to check
+                        # out a branch that matches the name of a branch
+                        # in the parentds, we don't even know if there is
+                        # such a branch...
```
kyleam: You'll have to ask yourself that: 2fda83e2ca (ENH: Record the active branch of a subdataset in the parent, 2019-10-20).  Based on that subject and the fact that `self.get_active_branch()` is evaluated for each `cand_sm`, I'd guess you were intending to specify the currently checked out branch in `cand_sm`, not the parent.

datalad/support/gitrepo.py
```diff
@@ -3975,6 +4037,10 @@ def save_(self, message=None, paths=None, _status=None, **kwargs):
                         self,
                         to_stage_submodules,
                         git_opts=None):
+                    # _save_add doesn't know better, but we know that
+                    # we only passed submodule records
+                    r['type'] = 'dataset'
```
kyleam: I see you're taking the public API exposed via result hooks seriously :]

datalad/support/gitrepo.py
```diff
@@ -4011,6 +4077,19 @@ def save_(self, message=None, paths=None, _status=None, **kwargs):
                                 else tuple())}):
                 yield r
 
+        # and lastly fixup subdataset states
+        for sub in [f for f, props in status.items()
```
kyleam: Oy, so I think we're looping over all tracked items in the repo again.  Can this be guard by an "on adjusted branch" condition?  Also, conceptually this seems a better fit for `AnnexRepo._save_post`.

datalad/core/local/tests/test_run.py
```diff
@@ -252,7 +252,7 @@ def test_run_from_subds_gh3551(path):
     assert_repo_status(ds.path)
     subds = Dataset(op.join(ds.path, subds_path))
     ok_exists(op.join(subds.path, "f"))
-    if not on_windows:  # FIXME
+    if ds.repo.is_managed_branch():  # FIXME
```
kyleam: Assuming this is indeed an adjusted branch thing, not a windows thing, shouldn't the `not` stay?

datalad/support/gitrepo.py
```diff
@@ -3644,16 +3668,24 @@ def _get_cache_key(label, paths, ref, untracked=None):
...
                     # but had subsequent modifications done to it that are
                     # unstaged. Such file would presently show up as 'added'
                     # ATM I think this is OK, but worth stating...
-                    state='modified' if f.exists() or \
-                    f.is_symlink() else 'deleted',
+                    # re `None`: if we face a dataset that git reports to
+                    # have a changed state, we cannot rely on this judgement
+                    # with subdatasets that are potentially in adjusted mode.
+                    # Setting the `state` to None will queue this record for
+                    # further processing below.
```
kyleam: In general I find this and related functions hard to grok, so your comments are very much appreciated.

datalad/support/gitrepo.py
```diff
@@ -2825,10 +2832,26 @@ def add_submodule(self, path, name=None, url=None, branch=None):
...
+        if subm_cbranch:
+            lgr.debug(
+                'Found submodule in git-annex adjusted mode, recording state '
+                'of corresponding branch %s instead',
+                subm_cbranch)
+            self.call_git([
+                'update-index', '--add', '--replace', '--cacheinfo',
+                '160000',
+                subm.get_hexsha(subm_cbranch),
+                path])
```
kyleam: Wow, whoever came update with the invocation is a very clever person :]

datalad/support/gitrepo.py
kyleam: I'd vote for sticking with the simpler `sync` call for now.

> OTOH, I dunno how clever git-annex is in figuring out that nothing needs to be done....

Quickly testing suggests that annex will _not_ sync back the changes if you amend the initial adjusted branch commit and have no other commits.

PR #4276, opened 2020-03-10
TST: Run crippled FS tests with development version of git-annex

mih:


mih:
Almost there:

```
FAILED (SKIP=8, errors=1, failures=2)
```

PR #4265, opened 2020-03-09
TEMP: (maint) - test devel annex 8.20200226+git16-ge156a2b74-1~ndall+1

yarikoptic:
Now that #4262 is merged, we should observe no stalling, and this hopefully would highlight possible left-over (in `maint`) issues

PR #4277, opened 2020-03-10
BF: resolve stalls in Runner while trying to access stdout/err from exited process

yarikoptic:
Got to it while troubleshooting https://github.com/datalad/datalad/pull/4265 but for unclear to me reason, I see no bloody effect on the actual run (even annex  `--debug` output is not printed): https://travis-ci.org/datalad/datalad/builds/660779184?utm_medium=notification&utm_source=github_status

Want to test across all tests now.

### DIFF
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -19,6 +19,7 @@ import os
 import atexit
 import functools
 import tempfile
+from select import select
 
 from collections import OrderedDict
 from .support import path as op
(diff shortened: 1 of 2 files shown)


yarikoptic:
d'oh -- now spotted that I didn't add `log_online` there and 2nd invocation was not RF'ed -- will do later, first want to see the effect of proposed changes

yarikoptic:
Just spotted a comment in the docstring of `select`:
```
    *** IMPORTANT NOTICE ***
    On Windows, only sockets are supported; on Unix, all file
    descriptors can be used.
```
and tests on windows confirmed that :-/:
```
  File "C:\hostedtoolcache\windows\Python\3.7.6\x64\lib\site-packages\datalad\cmd.py", line 121, in select_read
    r, w, e = select([f], [], [], timeout)
OSError: [WinError 10038] An operation was attempted on something that is not a socket
```

but also it seems to not do the right thing:
```
FAIL: datalad.tests.test_cmd.test_runner_remaining_output_two_pipes
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/datalad/tests/utils.py", line 559, in newfunc
    return t(*(arg + (d,)), **kw)
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/datalad/tests/test_cmd.py", line 367, in test_runner_remaining_output_two_pipes
    log_stderr=log_stderr)[0])
AssertionError: 'count: 0\nsize: 0\nin-pack: 0\npacks: 0\nsize-pack: 0\nprune-packable: 0\ngarbage: 0\nsize-garbage: 0\n' != 'count: 0\nsize: 0\nin-pack: 0\n'
```
and even `proc.communicate` works completely differently if I provide `timeout`... so whatever this attempt was -- it is not a solution :-/

PR #4283, opened 2020-03-12
TST: Update annex version for windows tests

mih:
When they say "jump!", we ask "how high?" ;-)


mih:
Looking better on crippled FS:

```
FAILED (SKIP=10, errors=2, failures=1)
```

PR #4285, opened 2020-03-12
ENH: extend message for a NoDatasetArgumentFound in require_dataset

yarikoptic:
This exception is now (since 0.12) is frequently shown to users due to breakage
of behavior as discussed in https://github.com/datalad/datalad/issues/3759 .
As it is worded minimally, it provides no instructions to the users and just
brings frustration.  This commit extends message to provide instructions on how
to mitigate the error they start to observe

### DIFF
--- a/datalad/core/local/tests/test_diff.py
+++ b/datalad/core/local/tests/test_diff.py
@@ -16,7 +16,7 @@ import os
 import os.path as op
 from datalad.support.external_versions import external_versions
 from datalad.support.exceptions import (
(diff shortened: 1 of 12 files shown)


mih:
Pushed to additionsto the original change. Cummulative effect is:

```
/tmp % datalad status
[ERROR  ] No dataset found at '/tmp'.  Specify a dataset to work with by providing its path via the `dataset` option, or change the current working directory to be in a dataset. [dataset.py:require_dataset:570] (NoDatasetFound) 
```

PR #4288, opened 2020-03-12
BF(TST,workaround): skip test which checks that git-annex creates ssh socket

yarikoptic:
Situation is not yet clear either it is purely our Runner (which did
reveal possible situation where it would hang, see
https://github.com/datalad/datalad/pull/4265) or there is something
really new worth considering in how git-annex process "behaves".

This commit is intended for 0.12.3 (maint branch).
When merging to master, this commit changes should first be RFed to
see if using GitWitlessRunner (instead of the old one as here)
mitigates the issue.

### DIFF
--- a/datalad/support/tests/test_annexrepo.py
+++ b/datalad/support/tests/test_annexrepo.py
@@ -1157,6 +1157,13 @@ def test_annex_ssh(repo_path, remote_1_path, remote_2_path):
     else:
         ok_(not exists(socket_1))
 
+    # TODO: figure it out, and possibly remove while merginging into
+    # master if WitlessRunner performs fine
+    if external_versions['cmd:annex'] >= '8.20200226':
+        # This is not necessarily the version where it started to hang
+        # See https://github.com/datalad/datalad/pull/4265 for more info
+        raise SkipTest("Version of git-annex might cause us to stall.")
+
     from datalad import lgr
     # remote interaction causes socket to be created:
     try:


kyleam:
> When merging to master, this commit changes should first be RFed to
> see if using GitWitlessRunner (instead of the old one as here)
> mitigates the issue.

I've pushed [a scratch branch][0] with that merge.  Here's [the Travis run][1]. 

[0]: https://github.com/datalad/datalad/compare/master...scratch/check-master-hanging-annex
[1]: https://travis-ci.org/github/datalad/datalad/builds/661702000

yarikoptic:
good that I spotted the comment -- was about to do the same.  I would have also removed that `swallow_logs` before that call in the tests, and comments about stalling ;-)

<details>
<summary>my diff:</summary> 

```shell
$> git diff
index 2e9b6a4a2..9fdfa1247 100644
@@ -1165,24 +1165,15 @@ def test_annex_ssh(repo_path, remote_1_path, remote_2_path):
     else:
         ok_(not exists(socket_1))
 
-    from datalad import lgr
     # remote interaction causes socket to be created:
     try:
-        # Note: For some reason, it hangs if log_stdout/err True
-        # TODO: Figure out what's going on
-        #  yoh: I think it is because of what is "TODOed" within cmd.py --
-        #       trying to log/obtain both through PIPE could lead to lock
-        #       downs.
-        # here we use our swallow_logs to overcome a problem of running under
-        # nosetests without -s, when nose then tries to swallow stdout by
-        # mocking it with StringIO, which is not fully compatible with Popen
-        # which needs its .fileno()
-        with swallow_outputs():
-            ar._run_annex_command('sync',
-                                  expect_stderr=True,
-                                  log_stdout=False,
-                                  log_stderr=False,
-                                  expect_fail=True)
+        ar._run_annex_command('sync',
+                              expect_stderr=True,
+                              log_stdout=False,
+                              log_stderr=False,
+                              expect_fail=True,
+                              runner='gitwitless',
+                              )
     # sync should return exit code 1, since it can not merge
     # doesn't matter for the purpose of this test
     except CommandError as e:
@@ -1206,12 +1197,13 @@ def test_annex_ssh(repo_path, remote_1_path, remote_2_path):
 
     # sync with the new remote:
     try:
-        with swallow_outputs():
-            ar._run_annex_command('sync', annex_options=['ssh-remote-2'],
-                                  expect_stderr=True,
-                                  log_stdout=False,
-                                  log_stderr=False,
-                                  expect_fail=True)
+        ar._run_annex_command('sync', annex_options=['ssh-remote-2'],
+                              expect_stderr=True,
+                              log_stdout=False,
+                              log_stderr=False,
+                              expect_fail=True,
+                              runner='gitwitless',
+                              )
```
</details>
Let's wait for what travis says!  Thanks @kyleam

kyleam:
> Let's wait for what travis says!

It's hanging: https://travis-ci.org/github/datalad/datalad/jobs/661702001#L1869

Also, another run showed a good number of failures in tests that were added in master: https://travis-ci.org/github/datalad/datalad/jobs/661702003

I can trigger at least some of them locally, so I might try to dig into them tomorrow.

>I would have also removed that `swallow_logs` before that call in the tests, and comments about stalling ;-)

Hmm, that made me think I missed comments that were added by this PR/workaround, but those are pretty old:

```
1719faafc6c (2017-04-07 17:05:41 -0400 1168)     from datalad import lgr
9df66cf7b6e (2016-05-26 00:43:57 +0200 1169)     # remote interaction causes socket to be created:
9df66cf7b6e (2016-05-26 00:43:57 +0200 1170)     try:
9df66cf7b6e (2016-05-26 00:43:57 +0200 1171)         # Note: For some reason, it hangs if log_stdout/err True
9df66cf7b6e (2016-05-26 00:43:57 +0200 1172)         # TODO: Figure out what's going on
7c074229804 (2017-03-10 14:03:24 -0500 1173)         #  yoh: I think it is because of what is "TODOed" within cmd.py --
7c074229804 (2017-03-10 14:03:24 -0500 1174)         #       trying to log/obtain both through PIPE could lead to lock
7c074229804 (2017-03-10 14:03:24 -0500 1175)         #       downs.
7c074229804 (2017-03-10 14:03:24 -0500 1176)         # here we use our swallow_logs to overcome a problem of running under
7c074229804 (2017-03-10 14:03:24 -0500 1177)         # nosetests without -s, when nose then tries to swallow stdout by
7c074229804 (2017-03-10 14:03:24 -0500 1178)         # mocking it with StringIO, which is not fully compatible with Popen
7c074229804 (2017-03-10 14:03:24 -0500 1179)         # which needs its .fileno()
7c074229804 (2017-03-10 14:03:24 -0500 1180)         with swallow_outputs():
```

Following up with a clean-up commit along the lines of your patch would make sense.  The unused import can of course be removed (though it's not made unused by this PR or the merge), and the `log_stdout/err=True` bit should no longer be true with the witless runner, but that also means that the call should probably no longer set those to parameters to False.  I'm not sure about the `swallow_outputs`/nose-without-dash-s bit though.  Isn't that still be relevant?

Anyway, given that it looks like we still need the workaround on master, all of this probably doesn't matter.

yarikoptic:
> It's hanging: https://travis-ci.org/github/datalad/datalad/jobs/661702001#L1869

heh, so no WitlessRunner to the rescue... and that somewhat reassures my suspicion that it might as well be that the process (git annex or ssh it invokes to provide a new ssh socket) is hanging.  I guess at some point in time we should backdoor travis again and have a closer look.  Otherwise - I am out of ideas on this one.

So, I will merge one in, and then it would make its due course into master ;) optimally with your changes - witless runner at least should provide us with more assurance that it is not our process outputs handling.

kyleam:
> optimally with your changes - witless runner at least should provide us with more assurance that it is not our process outputs handling.

Sorry, I decided to not put those commits on top.  We can of course do that now, ideally as a PR.  However, given that we're skipping this anyway with newer git-annex releases, I don't see much value in making the code diverge, and it seems better if done by a future investigation into the stalling.

kyleam:
> Also, another run showed a good number of failures in tests that were added in master: https://travis-ci.org/github/datalad/datalad/jobs/661702003
> 
> I can trigger at least some of them locally, so I might try to dig into them tomorrow.

For posterity: Those failures on master are due to the GETCONFIG  name issue described in gh-4259 and fixed on git-annex's side with 6a9147192 (GETCONFIG name fix, 2020-03-09), so there's nothing to be done here.

Issue #1974, opened 2017-11-20
Datalad save to annex files that were already in git


mvdoc:
#### What is the problem?

Whenever a file (which was under git) gets modified in a subdataset, and gets added to the parent dataset with `datalad add -d . subdataset`, the file gets committed to annex instead of git.

#### What steps will reproduce the problem?

```terminal
contematto@talete /tmp $ datalad create test_dataset
[INFO   ] Creating a new annex repo at /tmp/test_dataset
create(ok): /tmp/test_dataset (dataset)

contematto@talete /tmp $ cd test_dataset

contematto@talete /tmp/test_dataset (master) $ datalad create subdataset
[INFO   ] Creating a new annex repo at /tmp/test_dataset/subdataset
create(ok): /tmp/test_dataset/subdataset (dataset)

contematto@talete /tmp/test_dataset (master*) $ datalad add -d . subdataset
add(ok): subdataset (dataset) [added new subdataset]
add(notneeded): subdataset (dataset) [nothing to add from /tmp/test_dataset/subdataset]
add(notneeded): .gitmodules (file) [already included in the dataset]
save(ok): /tmp/test_dataset (dataset)
action summary:
  add (notneeded: 2, ok: 1)
  save (ok: 1)

contematto@talete /tmp/test_dataset (master) $ cd subdataset

contematto@talete /tmp/test_dataset/subdataset (master) $ touch myfileingit.txt

contematto@talete /tmp/test_dataset/subdataset (master*) $ git add myfileingit.txt

contematto@talete /tmp/test_dataset/subdataset (master*) $ git commit -m "Added file"
[master 1c48f6d] Added file
 1 file changed, 0 insertions(+), 0 deletions(-)
 create mode 100644 myfileingit.txt

contematto@talete /tmp/test_dataset/subdataset (master) $ ls
myfileingit.txt

contematto@talete /tmp/test_dataset/subdataset (master) $ cd ..

contematto@talete /tmp/test_dataset (master*) $ ls
subdataset

contematto@talete /tmp/test_dataset (master*) $ datalad add -d . subdataset
add(notneeded): subdataset/myfileingit.txt (file) [already included in the dataset]
add(notneeded): subdataset (dataset) [already known subdataset]
save(notneeded): /tmp/test_dataset/subdataset (dataset)
save(ok): /tmp/test_dataset (dataset)
action summary:
  add (notneeded: 2)
  save (notneeded: 1, ok: 1)

[...]

PR #4290, opened 2020-03-13
DOC: Rephrase save(to_git=) docs to indicate lack of choice persistence

mih:
Closes gh-1974

CI canceled, only doc change.

### DIFF
--- a/datalad/core/local/save.py
+++ b/datalad/core/local/save.py
@@ -130,14 +130,15 @@ class Save(Interface):
             args=("--to-git",),
             action='store_true',
             doc="""flag whether to add data directly to Git, instead of
-            tracking data identity only.  Usually this is not desired,
-            as it inflates dataset sizes and impacts flexibility of data
-            transport. If not specified - it will be up to git-annex to
-            decide, possibly on .gitattributes options. Use this flag
-            with a simultaneous selection of paths to save. In general,
-            it is better to pre-configure a dataset to track particular paths,
+            tracking data identity only.  Use with caution, there is no
+            guarantee that a file put directly into Git like this will
+            not be annexed in a subsequent save operation.
+            If not specified, it will be up to git-annex to decide how
+            a file is tracked, based on a dataset's configuration
+            to track track particular paths,
             file types, or file sizes with either Git or git-annex.
-            See https://git-annex.branchable.com/tips/largefiles/"""),
+            (see https://git-annex.branchable.com/tips/largefiles).
+            """),
     )
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


datalad/core/local/save.py
```diff
@@ -130,14 +130,15 @@ class Save(Interface):
...
-            transport. If not specified - it will be up to git-annex to
-            decide, possibly on .gitattributes options. Use this flag
-            with a simultaneous selection of paths to save. In general,
-            it is better to pre-configure a dataset to track particular paths,
+            tracking data identity only.  Use with caution, there is no
+            guarantee that a file put directly into Git like this will
+            not be annexed in a subsequent save operation.
+            If not specified, it will be up to git-annex to decide how
+            a file is tracked, based on a dataset's configuration
+            to track track particular paths,
```
adswa: There is a double "track" here:

```suggestion
            to track particular paths,
```

mih:
Thx @adswa !

datalad/core/local/save.py
```diff
@@ -130,14 +130,15 @@ class Save(Interface):
...
             doc="""flag whether to add data directly to Git, instead of
-            tracking data identity only.  Usually this is not desired,
-            as it inflates dataset sizes and impacts flexibility of data
-            transport. If not specified - it will be up to git-annex to
-            decide, possibly on .gitattributes options. Use this flag
-            with a simultaneous selection of paths to save. In general,
-            it is better to pre-configure a dataset to track particular paths,
+            tracking data identity only.  Use with caution, there is no
+            guarantee that a file put directly into Git like this will
+            not be annexed in a subsequent save operation.
```
kyleam: I personally find this unclear.  It sounds like it could be expected behavior `save()` for a save operation to throw _unmodified_ git files into annex.

Guess you didn't like my suggestion in gh-1974, but I think that avoids the issue.

mih: I think given the behavior shown in gh-1974 we have to discourage this option as much as possible. I did not consider the possibility to interpret my changes in the way you point out. But just the change you suggested did not emphasize the potential pitfalls enough for my taste. Personally, I think if the new wording discourages this kind of use more than necessary, it is still a good thing.

Ultimately, I think we have to make such a call add matching `.gitattribute` settings.

kyleam: > Personally, I think if the new wording discourages this kind of use more than necessary, it is still a good thing.

I don't think wording things in a way that could lead to a misunderstanding is a good thing, even if it leads to a particular behavior that we see as desirable.  Anyway, we don't seem to agree on this, it doesn't particularly matter, and it is already merged, so I'll drop it.

PR #4291, opened 2020-03-13
TST: Attempt to avoid short path reporting in GitHub Windows CI

mih:
Shot in the dark.


mih:
Didn't help in full, but removes some abiguity.

PR #4296, opened 2020-03-14
BF: resolve() does not imply absolute() on Windows for non-existing paths

mih:
This might be an issue in other places too!

The behavior of `resolve()` conflicts with its [documentation](https://docs.python.org/3/library/pathlib.html#pathlib.Path.resolve) that claims:

> Make the path absolute, resolving any symlinks.


mih:
It had the desired effect.

PR #4308, opened 2020-03-16
BF: downloaders -  lock (interprocess) initiation of the session or re-authentication

### LINKED ISSUE #4284: avoid multiple login prompts by git-annex-remote-datalad
ATM, an initial attempt to get (by default parallelized with git-annex) content provided via our special remote results in multitude "parallel" login prompts bringing chaos etc. E.g.

```shell
$> datalad install https://github.com/conpdatasets/preventad-open
[INFO   ]   Remote origin not usable by git-annex; setting annex-ignore                                          
install(ok): /tmp/preventad-open (dataset)                                                                       
datalad install https://github.com/conpdatasets/preventad-open  10.86s user 2.27s system 57% cpu 22.764 total

$> datalad -C preventad-open get  1004359                
Total:   0%|                                                                          | 0.00/631M [00:00<?, ?B/s]You need to authenticate with 'loris-prevent-ad' credentials. https://openpreventad.loris.ca/api/v0.0.3-dev/login provides information on how to gain access
user: You need to authenticate with 'loris-prevent-ad' credentials. https://openpreventad.loris.ca/api/v0.0.3-dev/login provides information on how to gain access
user: You need to authenticate with 'loris-prevent-ad' credentials. https://openpreventad.loris.ca/api/v0.0.3-dev/login provides information on how to gain access
user: You need to authenticate with 'loris-prevent-ad' credentials. https://openpreventad.loris.ca/api/v0.0.3-dev/login provides information on how to gain access
user: You need to authenticate with 'loris-prevent-ad' credentials. https://openpreventad.loris.ca/api/v0.0.3-dev/login provides information on how to gain access
user: You need to authenticate with 'loris-prevent-ad' credentials. https://openpreventad.loris.ca/api/v0.0.3-dev/login provides information on how to gain access
user: You need to authenticate with 'loris-prevent-ad' credentials. https://openpreventad.loris.ca/api/v0.0.3-dev/login provides information on how to gain access
user: You need to authenticate with 'loris-prevent-ad' credentials. https://openpreventad.loris.ca/api/v0.0.3-dev/login provides information on how to gain access
user: 

```

I think it could be easily avoided by locking, similarly to how we already do in git-annex-remote-datalad-archives special remote to avoid "parallel" refetching of the same tarball.  Probably locking could be done at the level of the Credential class level itself for any interaction with it, but we would need to check that a typical operation is "atomic".

yarikoptic:
This way parallel downloaders/special-remotes (e.g. as launched by git-annex) should not present multiple prompts for logins.  Only one would handle it, and then the others would just follow (since credentials would then be known).

Closes #4284

### DIFF
--- a/datalad/downloaders/base.py
+++ b/datalad/downloaders/base.py
@@ -131,6 +131,7 @@ class BaseDownloader(object, metaclass=ABCMeta):
             needs_authentication = self.credential
 
         attempt, incomplete_attempt = 0, 0
+        result = None
         while True:
             attempt += 1
             if attempt > 20:
@@ -139,8 +140,8 @@ class BaseDownloader(object, metaclass=ABCMeta):
             exc_info = None
             msg_types = ''
             supported_auth_types = []
+            used_old_session = False
             try:
-                used_old_session = False
                 used_old_session = self._establish_session(url, allow_old=allow_old_session)
                 if not allow_old_session:
                     assert(not used_old_session)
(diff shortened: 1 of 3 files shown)


datalad/downloaders/base.py
```diff
@@ -104,6 +108,11 @@ def __init__(self, credential=None, authenticator=None):
         self.credential = credential
         self.authenticator = authenticator
         self._cache = None  # for fetches, not downloads
+        self._lock = InterProcessLock(
+            op.join(
+                cfg.obtain('datalad.locations.cache'),
+                'locks',
+                'downloader-auth.lck'))
```
kyleam: Based on the `InterProcessLock` documentation, I didn't see any guarantees that this directory is created if it doesn't exist.  However, looking at the `fasteners` source code (and based on the passing tests), it does seem to be the case.  Okay.

yarikoptic: yeah -- I figured it "trial&error" way and didn't want to explicitly call `mkdirs` here since lock might not be needed (downloader not actually used) etc.  So decided to stick on fasteners to provide it

yarikoptic:
> For reviewers that aren't very familiar with the mocking in test_http (and didn't just spend time digging into it to try to understand a test failure), a small description of the problem would have been nice to see. From what I can gather, though, this looks fine to me.

ok, I will adjust commit description locally, merge into maint and push to close this PR.

PR #4319, opened 2020-03-18
ENH(UX): provide a hint on a deprecated add command

yarikoptic:
Now message would look like:

    datalad: Unknown command 'add'.  See 'datalad --help'.

    Did you mean any of these?
            addurls
    Hint: Command 'add' was deprecated in favor of 'save' command.

thus giving enough information to not dig through CHANGELOG.md to
proceed

### DIFF
--- a/datalad/cmdline/main.py
+++ b/datalad/cmdline/main.py
@@ -244,7 +244,10 @@ def setup_parser(
 
         if unparsed_arg not in known_commands:
             # check if might be coming from known extensions
-            from ..interface import _known_extension_commands
+            from ..interface import (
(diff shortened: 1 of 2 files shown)


adswa:
I like this. :)

yarikoptic:
linked travis run is the wrong one (I first by mistake pushed to origin, then removed this branch, and that run fails to find it), correct one seems to be https://travis-ci.org/github/datalad/datalad/builds/664109499?utm_medium=notification&utm_source=github_status  and it is green.
So - merging. Thank you for the review/feedback @adswa and @kyleam

PR #4320, opened 2020-03-18
RF: make non-interactive sessions use QuietConsoleLog (with LogProgressBar)

yarikoptic:
Such mode already was used during tests runs (only).

We would gain LOG (at INFO level) messages reporting upon completion of a
process with a progress (bar) reporting.  That would include uses of log_progress.

This also leads to removal of custom interactive vs non-interactive difference
in LoggerHelper.

Example output:

	$> rm -rf datalad && datalad clone https://github.com/datalad/datalad.git  | cat
	[INFO]  Enumerating done in 0.000221014 sec
	[INFO]  Counting 153 Objects done in 0.0139375 sec at 10978 Objects/sec
	[INFO]  Compressing 94 Objects done in 0.0118935 sec at 7903 Objects/sec
	[INFO]  Receiving 69836 Objects done in 3 seconds at 19489 Objects/sec
	[INFO]  Resolving 54608 Deltas done in 0.610522 sec at 89445 Deltas/sec
	[INFO]  Clone attempt 2 Candidate locations done in 5 seconds at 0.398732 Candidate locations/sec
	install(ok): /tmp/datalad (dataset)

This PR (branch) is based on a state of master before #4294 merge, which provides an alternative solution with explicit filtering of the log, thus would conflict in this PR initial version.

Pros:

- would be consistent across logging (INFO vs DEBUG) levels

Cons:

- timing/speed reporting might complicate inclusion of such output snippets
  into a handbook since they would be varying from run to run

- [ ] if deemed worthwhile, should be meld with an alternative solution for empty log lines from `log_progress` in master  (#4294)

### DIFF
--- a/datalad/log.py
+++ b/datalad/log.py
@@ -457,13 +457,12 @@ class LoggerHelper(object):
         #  logging.Formatter('%(asctime)-15s %(levelname)-6s %(message)s'))
         self.lgr.addHandler(loghandler)
 
-        if is_interactive():
-            phandler = ProgressHandler()
-            # progress only when interactive
-            phandler.addFilter(OnlyProgressLog())
-            # no stream logs of progress messages when interactive
-            loghandler.addFilter(NoProgressLog())
-            self.lgr.addHandler(phandler)
+        phandler = ProgressHandler()
+        # progress only
+        phandler.addFilter(OnlyProgressLog())
+        # no stream logs of progress
+        loghandler.addFilter(NoProgressLog())
+        self.lgr.addHandler(phandler)
 
         self.set_level()  # set default logging level
         return self.lgr
--- a/datalad/ui/__init__.py
+++ b/datalad/ui/__init__.py
@@ -38,6 +38,7 @@ KNOWN_BACKENDS = {
     'annex': UnderAnnexUI,
     'tests': UnderTestsUI,
     'tests-noninteractive': QuietConsoleLog,
+    'quiet': QuietConsoleLog,
     'no-progress': SilentConsoleLog,
 }
 
@@ -81,7 +82,7 @@ class _UI_Switcher(object):
                 else:
                     backend = 'dialog'
             else:
-                backend = 'dialog' if is_interactive() else 'no-progress'
+                backend = 'dialog' if is_interactive() else 'quiet'
         self._ui = KNOWN_BACKENDS[backend]()
         lgr.debug("UI set to %s" % self._ui)
         self._backend = backend


kyleam:
I haven't looked at the code yet, and, with a pass of your commit message, I'm having a hard time orienting myself but ...

> datalad:master from yarikoptic:rf-quiet

It looks like you meant for this to be against maint.  In my view this sort of change should be against master.

yarikoptic:
I meant it to go against master, but could rebase against maint if we decide it is with it

kyleam:
> I meant it to go against master

Ah, well then please ignore my comment.  I assumed you were targeting maint but forgot to select it as the base when opening the PR because you say "[melding with alternative] in master" and because GitHub reported conflicts with master.

Looking at the actual branch, I see you chose a base a bit upstream of the current master, right before gh-4293.  As this deliberate and unusual choice of base is essential to assessing this PR and trying to understand the context and your rationale, it seems worth a mention in the PR's description.

yarikoptic:
> Looking at the actual branch, I see you chose a base a bit upstream of the current master, right before [gh-4293](https://github.com/datalad/datalad/pull/4293). As this deliberate and unusual choice of base is essential to assessing this PR and trying to understand the context and your rationale, it seems worth a mention in the PR's description.

FWIW - done.  I there will be no further objections I will proceed with it later today.  Also I have filed a complimentary #4325 on how we could improve LogProgressBar.

kyleam:
> I there will be no further objections I will proceed with it later today.

FWIW I haven't had a chance to take a look at this, so the lack of objections from me doesn't mean much.

datalad/ui/__init__.py
```diff
@@ -81,7 +82,7 @@ def set_backend(self, backend):
                 else:
                     backend = 'dialog'
             else:
-                backend = 'dialog' if is_interactive() else 'no-progress'
+                backend = 'dialog' if is_interactive() else 'quiet'
```
kyleam: In terms of non-interactive behavior, I agree that showing the translating a progress bar to a log message upon completion is better than silence.  The label "quiet" doesn't feel quite intuitive to me, but I guess it makes sense given the name `QuietConsoleLog`, and I don't have a better suggestion.

datalad/log.py
```diff
@@ -457,13 +457,12 @@ def filter(self, record):
...
-            phandler.addFilter(OnlyProgressLog())
-            # no stream logs of progress messages when interactive
-            loghandler.addFilter(NoProgressLog())
-            self.lgr.addHandler(phandler)
+        phandler = ProgressHandler()
+        # progress only
+        phandler.addFilter(OnlyProgressLog())
+        # no stream logs of progress
+        loghandler.addFilter(NoProgressLog())
+        self.lgr.addHandler(phandler)
```
kyleam: This is problematic.  It means that non-interactively individual `log_progress` calls aren't showing as log messages, but that's a feature at the heart of `log_progress`.

yarikoptic: I am missing it -- if we get a LogProgressBar, why non-interactive mode should have any special handling here/ in `log_progress`?

kyleam: > if we get a LogProgressBar, why non-interactive mode should have any special handling here/ in `log_progress`?

Because it's a fundamental change to how `log_progess` is intended to work.  It's designed to log _individual_ calls to `log_progress` when operating non-interactively.

As an example, take this script.

```python
import logging

lgr = logging.getLogger("datalad.foo")
nthings = 4

log_progress(lgr.info, "foo",
             "Checking %d things", nthings,
             label="Checking things", total=nthings, unit=" things")
for i in range(nthings):
    log_progress(lgr.info, "foo",
                 "Checking thing %d", i,
                 increment=True, update=1)
log_progress(lgr.info, "foo",
             "Finished checking %d things", nthings)
```

Before this PR, the output is

```
[INFO] Checking 4 things 
[INFO] Checking thing 0 
[INFO] Checking thing 1 
[INFO] Checking thing 2 
[INFO] Checking thing 3 
[INFO] Finished checking 4 things 
```
With this PR, the output is

```
[INFO]  Checking things 4 things done in 0.000181913 sec at 21988 things/sec 
```

This change is entirely disabling the non-interactive side of `log_progress`.  I don't think that's a good thing, but if that's the intent, then the `log_progress` functionality/wrapper should be substantially reworked and reduced if not removed.  And the commit messages should very clearly state this and argue for this change.

yarikoptic:
"Conflicts" in logic were not resolved, and there seems to be some (although without metrics etc) INFO log lines now
```
$> rm -rf datalad && datalad clone https://github.com/datalad/datalad.git  | cat
[INFO] Cloning dataset to Dataset(/tmp/datalad) 
[INFO] Attempting to clone from https://github.com/datalad/datalad.git to /tmp/datalad 
[INFO] Start enumerating objects 
[INFO] Start receiving objects 
[INFO] Start resolving deltas 
[INFO] Completed clone attempts for Dataset(/tmp/datalad) 
install(ok): /tmp/datalad (dataset)

$> datalad --version
datalad 0.13.3.dev3
```
so I will just close this one for now

Issue #4207, opened 2020-02-26
create(no_annex=False) -> create(annex=True)


mih:
My brain hurts with the double-negatives. We should flip the option (deprecation cycle and all of course....)

yarikoptic:
Wouldn't it make it inconsistent with cmdline, this setup causing head aches but of a different kind?

mih:
No, it is easy to have --no-annex simultaneously. We do it elsewhere too.

PR #4321, opened 2020-03-19
create(no_annex=False) -> create(annex=True)

mih:
This avoids double-negative thinking when using the Python API, and fixes gh-4207

There is no change wrt the command-line API.

A warning is emitted if, and only if, using the Python API `no_annex=True` is given in conflict with the value of `annex`.

In order to make this a smooth transition (but possibly with further future benefits), it is now possible to have Python-only arguments for our commands.

This can be achieved by setting the `args` of a `Parameter` instance to be an empty tuple. The value `None` continues to indicate that a default argument name should be derived from the
parameter name. Example:

```
some_param=Parameter(args=tuple(), ...)
```

### DIFF
--- a/datalad/core/local/create.py
+++ b/datalad/core/local/create.py
@@ -158,10 +158,16 @@ class Create(Interface):
             action='store_true'),
         description=location_description,
         no_annex=Parameter(
-            args=("--no-annex",),
-            doc="""if set, a plain Git repository will be created without any
-            annex""",
+            # hide this from the cmdline parser, replaced by `annex`
+            args=tuple(),
+            doc="""this option is deprecated, use `annex` instead""",
             action='store_true'),
+        annex=Parameter(
+            args=("--no-annex",),
+            dest='annex',
+            doc="""if [CMD: set CMD][PY: disabled PY], a plain Git repository
+            will be created without any annex""",
+            action='store_false'),
         # TODO seems to only cause a config flag to be set, this could be done
         # in a procedure
         fake_dates=Parameter(
@@ -192,9 +198,23 @@ class Create(Interface):
             description=None,
             dataset=None,
             no_annex=False,
+            annex=True,
             fake_dates=False,
             cfg_proc=None
     ):
+        # TODO: introduced with 0.13, remove with 0.14
+        if annex != (not no_annex) and no_annex:
+            # the two mirror options do not agree and the deprecated one is
+            # not at default value
+            lgr.warning("datalad-create's `no_annex` option is deprecated "
+                        "and will be removed in a future release, "
+                        "use the reversed-sign `annex` option instead.")
+            # honor the old option for now
+            annex = not no_annex
+
+        # we only perform negative tests below
+        no_annex = not annex
+
         if dataset:
             if isinstance(dataset, Dataset):
                 ds = dataset
--- a/datalad/core/local/create.py
+++ b/datalad/core/local/create.py
@@ -14,9 +14,9 @@ import os
 import logging
 import random
 import uuid
+import warnings
 from argparse import (
     REMAINDER,
-    ONE_OR_MORE,
 )
 
 from os import listdir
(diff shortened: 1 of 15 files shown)


datalad/core/local/create.py
```diff
@@ -192,9 +198,23 @@ def __call__(
...
             fake_dates=False,
             cfg_proc=None
     ):
+        # TODO: introduced with 0.13, remove with 0.14
+        if annex != (not no_annex) and no_annex:
+            # the two mirror options do not agree and the deprecated one is
+            # not at default value
+            lgr.warning("datalad-create's `no_annex` option is deprecated "
+                        "and will be removed in a future release, "
+                        "use the reversed-sign `annex` option instead.")
```
kyleam: I think `warnings.warn(..., DeprecationWarning)` is preferable here for a few reasons, including consistency within our codebase, allowing developers to control `DeprecationWarnings` in the same way as they do in other spots, and avoiding a flood of warnings by default.

datalad/core/local/create.py
```diff
@@ -192,9 +198,23 @@ def __call__(
             description=None,
             dataset=None,
             no_annex=False,
+            annex=True,
             fake_dates=False,
             cfg_proc=None
     ):
+        # TODO: introduced with 0.13, remove with 0.14
+        if annex != (not no_annex) and no_annex:
```
kyleam: Okay, so the treatment of these potentially conflicting parameters is the meat of this PR in terms of logic.  Restricting the call to boolean values (or assuming all values are cast to booleans), here are the cases.

```
{'annex': True, 'no_annex': False}
[INFO   ] Creating a new annex repo at /tmp/dl-uszzhl4w 
```

This one is unproblematic in terms of the type of repo produced.  The caller may be explicitly passing `no_annex=False` and won't see a warning.

```
{'annex': False, 'no_annex': False}
[INFO   ] Creating a new git repo at /tmp/dl-ifdhlpe6 
```

The caller is probably using the new `annex=False` while keeping `no_annex` at its default value.  The fact that the non-default value for the newer parameter is honored (i.e. a git repo is created) seems reasonable.  While unlikely, it is possible that the caller is also explicitly passing `no_annex=False` and won't see a warning.

```
{'annex': False, 'no_annex': True}
[INFO   ] Creating a new git repo at /tmp/dl-8p1k5_71 
```

The caller is specifying a non-default value for both `annex` and `no_annex`.  This is unproblematic because they produce the same repo typo, but ideally the caller would see a warning if using `no_annex` (even if this scenario is unlikely).

```
{'annex': True, 'no_annex': True}
[WARNING] datalad-create's `no_annex` option is deprecated and will be removed in a future release, use the reversed-sign `annex` option instead. 
[INFO   ] Creating a new git repo at /tmp/dl-sy5pumc2 
```

The caller is specifying `no_annex`, and probably not explicitly giving a value for `annex`.   Based on the likely scenario, we honor the `no_annex=True` setting and create a git repo.

---

So, I think there's room there for confusion, even if the most likely scenarios are handled reasonably.  But we can pretty easily handle this unambiguously.  Set the default for `no_annex` to a value that we know the user did not pass in.  If `no_annex` isn't that value, honor `no_annex` and give the warning.  Something like:

```diff
diff --git a/datalad/core/local/create.py b/datalad/core/local/create.py
index cc3cd3442e..ba5c7da0b0 100644
--- a/datalad/core/local/create.py
+++ b/datalad/core/local/create.py
@@ -62,6 +62,10 @@
 lgr = logging.getLogger('datalad.core.local.create')
 
 
+class _NoAnnexDefault(object):
+    pass
+
+
 @build_doc
 class Create(Interface):
     """Create a new dataset from scratch.
@@ -197,13 +201,13 @@ def __call__(
             force=False,
             description=None,
             dataset=None,
-            no_annex=False,
+            no_annex=_NoAnnexDefault,
             annex=True,
             fake_dates=False,
             cfg_proc=None
     ):
         # TODO: introduced with 0.13, remove with 0.14
-        if annex != (not no_annex) and no_annex:
+        if no_annex is not _NoAnnexDefault:
             # the two mirror options do not agree and the deprecated one is
             # not at default value
             lgr.warning("datalad-create's `no_annex` option is deprecated "
```

mih: Makes sense. Thx for the inspiration!

datalad/core/local/create.py
mih: Will do!

Issue #1146, opened 2016-11-30
datalad install --shared


yarikoptic:
(should also be in `create`): Could be just `group` to setup full group shared (rw) repository, or may be even in addition `group-ro` which would inherit group settings and assure that stuff is group readable.

the setting specified during install should be stored within `.git/config` so it could be inherited while installing sub-datasets later

bpoldrack:
Note: within `Install` `git_clone_opts` are ignored. They need to be passed to the clone helper function and then down to `GitRepo.clone`, which has `**kwargs` for this. Implementation of a fix is just about the correct passing of options like this.

mih:
Given https://github.com/datalad/datalad/issues/1893 it might not be useful to have this at all. Alternatively, we could offer it as a "reckless" mode. Multiple reckless-flavors are now possible with https://github.com/datalad/datalad/pull/3966

yarikoptic:
well, not "not useful" but just a bit "dangerous" ;)
But it might (still) be needed for any centralized collaborative work (if there is no alternative permissions handling framework like of gitlab etc is there to do it).  It is not really about "reckless"ness - those are pretty much orthogonal if I see it right

Issue #1893, opened 2017-10-10
any --shared setting is futile for annex -- annexed makes files writeable!


yarikoptic:
```shell
$> sudo rm -rf tttt; mkdir tttt; ( cd tttt; git init --shared .; git annex init ; echo 123 >123; git annex add 123; git commit -m 'added 123'; ls -lL 123; ) 
Initialized empty shared Git repository in /tmp/tttt/.git/
init  ok
(recording state in git...)
add 123 ok
(recording state in git...)
[master (root-commit) b137ea6] added 123
 1 file changed, 1 insertion(+)
 create mode 120000 123
-rw-rw---- 1 yoh yoh 4 Oct 10 12:59 123

$> echo 11 >> tttt/123
# no error
```

yet to file an issue with annex. tried with old annex 6.20170101-1+b1 and new one 6.20171003-g14ffdd779

ref: http://git-annex.branchable.com/bugs/--shared_setting_of_git_causes_annex__39__ed_files_to_be_writeable__33__/?updated

yarikoptic:
I thought to mark this issue as FOI and have it closed, but then changed my mind.
I do still consider this "Feature" to be a bug or a limitation in design/implementation in git-annex.  Arguments in the original git-annex thread made analogies to "indirect mode" (now gone), and argued for not introducing another locking mechanism without major repository version change (of which we had a number by now).  So, if there is an interested DataLad person in using shared repositories, might be good time to follow up on that [thread](https://git-annex.branchable.com/bugs/shared_setting_of_git_causes_annex__39__ed_files_to_be_writeable__33__/).

PR #4324, opened 2020-03-19
 NF: clone/install --reckless shared-{group|all|...}

mih:
Makes it possible to set up datasets for collaborative write access
on install. This is implemented as a reckless mode, because it bypasses
the standard annex write protection feature (see gh-1893).

- rewrote the `reckless` option docs
- added example to the clone docs
- add test of permission setup validity

This fixes gh-1146 although it doesn't implement a dedicated `--shared`
option. And it mitigates gh-1893, by providing a dedicated, documented mode
that marks the effected datasets as untrusted.

### DIFF
--- a/datalad/core/distributed/clone.py
+++ b/datalad/core/distributed/clone.py
@@ -10,7 +10,6 @@
 
 
 import logging
-import os
 import re
 from os.path import expanduser
 from collections import OrderedDict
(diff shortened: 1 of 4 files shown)


yarikoptic:
hm, interesting -- although I was the one to whine, I have not considered "shared" to be a "reckless" kind. Also, since it is directing `git` operations it is partially orthogonal to currently available "ephemeral" and "hardlink" options. E.g. I could establish a shared git repository with ephemeral storage. Moreover if ephemeral source is itself shared, this would actually be desired (so install might want to "propagate" the "shared" effect of source).
Altogether I still wonder if it shouldn't be a separate option.

mih:
@yarikoptic I considered these scenarios and found them unlikely. Until there is evidence, I'd prefer to keep things tightly localized and not introduce a dedicated option. If combinatorics of reckless modes become necessary, we can non-invasively enable multiple modes to be passed and deal with that internally.

PS: Consider closing all issue that you don't want me to fix ;-P

yarikoptic:
> we can non-invasively enable multiple modes to be passed and deal with that internally.

I was just thinking to avoid possible future API breakage when such combinations would become  desired. I have avoided using `shared` for now, but situation might change ;-)
 
Having a "non-reckless" `shared` repository should be possible, #1893 is documenting/relating to a bug in git-annex about this, and until it is fixed (in git-annex), I would prefer it to stay open (thus adjusted description here to not close it).
But I am ok to proceed with this -- it makes things possible and explicitly documents the dangers, if anything we could deal with it later and introduce a dedicated shared when time is ripe ;-)

mih:
I'm cool with keeping #1893 open.

Just FTR: It would not cause any API breakage to start accepting multiple reckless modes in a single call (list in Python, multiple specification of `--reckless` on the cmdline) -- in the future.

Issue #4032, opened 2020-01-16
Double-dash options get escaped in manpages


mih:
![image](https://user-images.githubusercontent.com/136479/72535701-46e78e00-3879-11ea-9851-4bdf021eaa88.png)

mih:
This is a side-effect of avoiding typographic m-dashes in the HTML-rendered manpages.

PR #4332, opened 2020-03-24
BF: Avoid spurious backslashes in manpage double-dash option listings

mih:
They were introduced to avoid rendering of typographic m-dashes instead
of double-dashes in the HTML version. This change limits this escaping
to the HTML version.

Fixes gh-4032


yarikoptic:
shouldn't this be against `maint`?
```
$> git grep 'for p in parts)'
_datalad_build_support/formatters.py:            return ', '.join(p.replace('--', '-\\\\-') for p in parts)

$> git describe
0.12.4-9-gcc7e62771
```

kyleam:
In my view a cosmetic fix like this is fine to be master only.

mih:
> In my view a cosmetic fix like this is fine to be master only.

I agree, but also wouldn't object if someone picks it up in `maint`. I admit that the choice of `master` wasn't a conscious one.

PR #4341, opened 2020-03-24
OPT: use locally bound variables instead of dataset.repo and .config

yarikoptic:
Also includes the helper decorator in the first commit (`@collect_method_callstats`) -- see that commit for an example use.

Let's see what benchmarks run would say ;-)


yarikoptic:
hm, I do not see travis report?! and there is no notable difference on benchmarks.. I had high hopes for this one
```
30.5±0.08s       29.9±0.08s     0.98  usecases.study_forrest.time_make_studyforrest_mockup
```
but `-2%` is within "error margin"

PR #4342, opened 2020-03-25
BF: set LC_ALL=C while running git

### LINKED ISSUE #4340: Override language?
On the call today, I said that our matching of Git output that is marked for translation would cause errors for environments where English isn't configured as the default language.  Yarik mentioned that he thinks we override the language in datalad's environment to avoid this problem.  I might be missing something, but that doesn't seem to be the case.  For example, here's the failing test from gh-4328:

```
$ git describe                              
0.12.4-554-ge49e39d458

$ export LANG=bg_BG.UTF-8

$ python -m nose datalad/support/tests/test_gitrepo.py:test_GitRepo_gitignore
E
ERROR: datalad.support.tests.test_gitrepo.test_GitRepo_gitignore
[...]
    raise exc
datalad.support.exceptions.CommandError: CommandError: 'git -c annex.largefiles=nothing add --verbose -- ignore.me' failed with exitcode 1 under /tmp/datalad_temp_tree_test_GitRepo_gitignore9_0hvt58 [err: 'Следните пътища ще бъдат игнорирани според някой от файловете „.gitignore“:
ignore.me
подсказка: Ползвайте опцията „-f“, ако наистина искате да ги добавите.
подсказка: За да изключите това съобщение, изпълнете:
подсказка: 
подсказка:     git config advice.addIgnoredFile false']
[...]
```

yarikoptic:
That should ensure that error messages are hints are in English so
we could still parse them and react accordingly.

Closes #4340

TODO:
- [DID NOT WORK] test - should we set LANG in some of the travis runs?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,6 +40,9 @@ matrix:
     # We cannot have empty -A selector, so the one which always will be fulfilled
     - NOSE_SELECTION=
     - NOSE_SELECTION_OP=not
+    # To test https://github.com/datalad/datalad/pull/4342 fix.
+    # From our testing in that PR seems to have no effect, but kept around since should not hurt.
+    - LC_ALL=ru_RU.UTF-8
   - python: 3.6
     # Single run for Python 3.6
     env:
@@ -56,6 +59,9 @@ matrix:
     env:
     - DATALAD_REPO_VERSION=6
     - NOSE_SELECTION_OP=""
+    # To test https://github.com/datalad/datalad/pull/4342 fix in case of no "not" for NOSE.
+    # From our testing in that PR seems to have no effect, but kept around since should not hurt.
+    - LANG=bg_BG.UTF-8
   - python: 3.5
     # Run slow etc tests under a single tricky scenario
     env:
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -704,6 +704,11 @@ class GitRunner(Runner):
             git_env['GIT_SSH_COMMAND'] = GIT_SSH_COMMAND
             git_env['GIT_SSH_VARIANT'] = 'ssh'
 
+        # We are parsing error messages and hints. For those to work more
+        # reliably we are doomed to sacrifice i18n effort of git, and enforce
+        # consistent language of the messages
+        git_env['LC_ALL'] = 'C'
+
         return git_env
 
     def run(self, cmd, env=None, *args, **kwargs):
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -707,7 +707,10 @@ class GitRunner(Runner):
         # We are parsing error messages and hints. For those to work more
         # reliably we are doomed to sacrifice i18n effort of git, and enforce
         # consistent language of the messages
-        git_env['LC_ALL'] = 'C'
+        git_env['LC_MESSAGES'] = 'C'
+		# But since LC_ALL takes precedence, over LC_MESSAGES, we cannot
+		# "leak" that one inside, and are doomed to pop it
+        git_env.pop('LC_ALL', None)
 
         return git_env
 


kyleam:
I said:

> [...] but I'm also fine with this as is.

Oh, but it looks like the tests are not ([example][0]).

[0]: https://travis-ci.org/github/datalad/datalad/jobs/666583054#L1108

yarikoptic:
"interesting".  Either it just fails early enough into the test, or it is specially `git annex init` ( or likely - any git annex command ) which can't operate under a path with unicode characters in it if `LC_ALL=C`.  Something starts nagging in the back of my head about some old discussion with joeyh about locales&git-annex which might be relevant ;-)

I thought that we might want to just overload `LC_MESSAGES` but then (which seems to work ok and takes precedence over LANG), if `LC_ALL` is set, we are back to the original issue.  So, pushed e2d177b55b840e01d07c52412498bc5a57f4746f which sets LC_MESSAGES and pops LC_ALL ;) let's see how that one works out (seems ok on a quick local testing)

kyleam:
Nearly all the builds were failing in the setup phase (pypi download issue, I think).  On restart, the extensions can't seem to check out the right state, but Travis and AppVeyor got to the actual tests and are running now.

yarikoptic:
hoorray -- travis and appveyor are green!  Good base -- now I will try adding LC_ALL and LANG (for the sake of it) into one of the travis runs, and see how things look after that.

.travis.yml
```diff
@@ -40,6 +40,8 @@ matrix:
     # We cannot have empty -A selector, so the one which always will be fulfilled
     - NOSE_SELECTION=
     - NOSE_SELECTION_OP=not
+    # To test https://github.com/datalad/datalad/pull/4342 fix
+    - LC_ALL=ru_RU.UTF-8
```
kyleam: Wouldn't you have to generate these locales?

yarikoptic: good question on either they are in effect.  Let me do one more run with a TEMP commit reverting the changes to code -- then things better fail

kyleam: perfect, thanks

kyleam: > Let me do one more run with a TEMP commit reverting the changes to code -- then things better fail

Looking green.

yarikoptic: :-( so indeed need to install/generate them I guess. will try now

.travis.yml
```diff
@@ -74,6 +74,10 @@ before_install:
   - if [ ! -z "${_DL_UPSTREAM_GIT:-}" ]; then
       sudo tools/ci/install-latest-git.sh || { [ $? -eq 100 ] && exit 0; } || exit 1;
     fi
+  - sudo eatmydata apt-get install -y locales
+  - sudo sed -e 's,^# \(ru_RU.UTF-8\|bg_BG.UTF-8\),\1,g' /etc/locale.gen
```
kyleam: One run on the most recent push passed (and the other is still running).  Missing an `-i`?

yarikoptic: Damn it, .. thank you!

yarikoptic:
```suggestion
  - sudo sed -i -e 's,^# \(ru_RU.UTF-8\|bg_BG.UTF-8\),\1,g' /etc/locale.gen 
```

kyleam:
The build still wasn't showing the expected failure.  I though that might be due to the use of the bundled git.  In 216204b I set `DATALAD_USE_DEFAULT_GIT=1` but the build [still passes][0], and even a plain `git status` call [is still in English][1].  Stumped.

[0]: https://travis-ci.org/github/datalad/datalad/builds/667706479
[1]: https://travis-ci.org/github/datalad/datalad/jobs/667706480#L732

yarikoptic:
bundled -- indeed, it comes with its own locales... but I am using git-annex-standalone locally as well and can trigger it.
I am ok to give up on trying to make testing "work" ;) I would have probably left those two LC_ and LANG in just in case they would trigger at some point.

kyleam:
> I am ok to give up on trying to make testing "work" ;)

I am too (not surprisingly, since initially I said I'm okay without a test here).

> I would have probably left those two LC_ and LANG in just in case they would trigger at some point.

I'm a bit confused by this comment.  It makes it sound like I pushed an update that removed them.

Anyway, I personally would prefer not to leave such cruft in the final state, but doing so is fine by me assuming there's an accompanying comment.

yarikoptic:
> > I would have probably left those two LC_ and LANG in just in case they would trigger at some point.
> 
> I'm a bit confused by this comment. It makes it sound like I pushed an update that removed them.

nah, I meant to say that I would have taken 48c49fced6f1b6e87d7ee8f59e33558404fbe5d5 as the state to merge

> Anyway, I personally would prefer not to leave such cruft in the final state, but doing so is fine by me assuming there's an accompanying comment.

comment in the commit or in .travis.yml?

kyleam:
> comment in the commit or in .travis.yml?

My preference would be a comment in .travis.yml for visibility and to prevent anyone needing to dig to find out that these didn't have an effect from the start.

kyleam:
> nah, I meant to say that I would have taken [48c49fc](https://github.com/datalad/datalad/commit/48c49fced6f1b6e87d7ee8f59e33558404fbe5d5) as the state to merge

"would have" suggests that something different has already been done.

yarikoptic:
ok, done -- if still green, we could merge

PR #4354, opened 2020-03-27
BF+ENH: search --show-keys -- no incorrect "unhashable", use repr

yarikoptic:
+ENH: to be able to limit which keys to report

If you ran smth like `datalad search -d /// --show-keys short` you could see a long list of keys, where you would see smth like
```
annex.age
 in  1 datasets
 has 1 unique values: 'unhashable 1688 out of 1690 entries'
```
which would not only be uninformative but also incorrect.  It is 1688 entries which were "hashable" and the rest not.  But also you wouldn't see a single value.  With first 2 commits (1 commit is the first attempt, superseded by 2nd commit, decided not to squash them) we would use `repr` for all the values so there is no "unhashable" results reported.  With the 3rd commit I added an ENH to be able to limit which keys to report --  a really small/easy addition but quite powerful, so now I can get full result of interest without additional `grep` or alike:

```shell
$> datalad search -d /// --show-keys short '^annex\.age$'     
annex.age
 in  1 datasets
 has 1690 unique values: '13.25'; '19.64'; '26.78'; '29.9685'; '32.63'; '34.28884326'; '8.93'; '9.4'; ....
```
As you can't see here -- output formatting changed a bit as well -- now uses ';' instead of ','.

### DIFF
--- a/datalad/metadata/search.py
+++ b/datalad/metadata/search.py
@@ -252,7 +252,15 @@ class _Search(object):
     def __call__(self, query, max_nresults=None):
         raise NotImplementedError
 
-    def show_keys(self, *args):
+    @classmethod
+    def _key_matches(cls, k, regexes):
+        """Return which regex the key matches
+        """
+        for regex in regexes:
+            if re.search(regex, k):
+                return regex
+
+    def show_keys(self, *args, **kwargs):
         raise NotImplementedError(args)
 
     def get_query(self, query):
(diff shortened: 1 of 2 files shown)


yarikoptic:
since it is something I would like to use (and in general - use this functionality), and since nobody else objects and/or cares, I will merge it tomorrow.

yarikoptic:
thank you @kyleam !  I am split between maint-vs-master, since ideally I should have just made a BF for maint, but since it is a minor enhancement to rarely used user-interfaced (not machine readable) information, I will just proceed with maint.

PR #4358, opened 2020-03-29
RF: Remove .asv as a submodule

yarikoptic:
I think it kept causing more pain than gain with commits of changed .asv to git etc.
.asv/ already in .gitignore.  I am yet not sure how/if it would impair my benchmarking
setup - will deal with it later

Positioning against `maint` so we do not get into issues while switching between maint and master if one has it and another one doesn't (although I guess git might handle it just fine)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule ".asv"]
-	path = .asv
-	url = https://github.com/datalad/.asv.git


yarikoptic:
not yet sure if in the short term it wouldn't cause pains but I do think it is the right thing to do in the long term ;) Thank you @kyleam

PR #4368, opened 2020-04-01
BF: clone doesn't always annex-init

### LINKED ISSUE #4367: clone fails to annex-init due to discovery of "chained" origins
I have repo with autoenabled special remote and clone it. Gets enabled, I can obtain data - all good. Now I clone it from that clone to a third location. Special remote isn't autoenabled, while `git annex init` will do it. Nevertheless I have a git-annex branch and a filled `.git/annex` before that `annex-init` call.

What's happening is, that we discover from within `clone`, that there is that original `origin` which we add to the clone of the clone as well as `origin-2` and try to enable it to figure its UUID. This `git-annex enableremote` call is done right before we check whether we need to `git-annex-init` the fresh clone and now it looks already initialized to us, since we find a local git-annex branch and everything. Therefore we don't call `git-annex-init`.

Trying to fix this by simply change the order of things. A possible need for `annex-init` should be considered before anything else, I think.

bpoldrack:
Fixes #4367

### DIFF
--- a/datalad/core/distributed/tests/test_clone.py
+++ b/datalad/core/distributed/tests/test_clone.py
@@ -505,8 +505,13 @@ def test_cfg_originorigin(path):
     # Clone another level, this time with a relative path. Drop content from
     # lev2 so that origin is the only place that the file is available from.
     clone_lev2.drop("file1.txt")
-    with chpwd(path):
+    with chpwd(path), swallow_logs(new_level=9) as cml:
         clone_lev3 = clone('clone_lev2', 'clone_lev3')
+        # we called git-annex-init; see gh-4367:
+        cml.assert_logged(msg=r"[^[]*Running: \[('git', 'annex'|'git-annex'), "
+                              r"'init'",
+                          match=False,
+                          level='Level 9')
     assert_result_count(
         clone_lev3.get('file1.txt', on_failure='ignore'),
         1,
--- a/datalad/core/distributed/clone.py
+++ b/datalad/core/distributed/clone.py
@@ -564,10 +564,6 @@ def postclonecfg_annexdataset(ds, reckless, description=None):
         ds.config.set(
             'annex.hardlink', 'true', where='local', reload=True)
 
-    # we have just cloned the repo, so it has 'origin', configure any
-    # reachable origin of origins
-    yield from configure_origins(ds, ds)
-
     lgr.debug("Initializing annex repo at %s", ds.path)
     # Note, that we cannot enforce annex-init via AnnexRepo().
     # If such an instance already exists, its __init__ will not be executed.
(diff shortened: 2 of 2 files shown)


yarikoptic:
ideally should have come with a unittest which would have failed without the fix. Otherwise - we might get into the same situation later after some code RFing ;)

kyleam:
[ I can answer for @bpoldrack :) ]

> ideally should have come with a unittest which would have failed without the fix. Otherwise - we might get into the same situation later after some code RFing ;)

In addition to the "[WIP]" label, his commit message says

```
TODO: This is to be ammended with a test. Just pushing
to see whether the fix breaks anything
```

bpoldrack:
Thanks, @kyleam . :-D

Here we go ... force-pushed.

bpoldrack:
And before you point me to it, @yarikoptic : Yes, I'm aware of the `assert_logged` method for those log swallowing context managers. But for some reason doesn't work for me with regex. Ran into it several times, I believe, but I I have no time or juice to fix that ATM. It doesn't even match the level correctly because of double-escaping the brackets.

kyleam:
re: assert_logged, at least one discussion I think you're is with me [here][0].  I promise not to mention it :]

[0]: https://github.com/datalad/datalad/pull/3518#discussion_r301685701

bpoldrack:
Damn it. Now I have to try at least. Gimme a second ;-)

bpoldrack:
Okay, okay. That approach worked. Still don't know, what needs to go into the docstring for `assert_logged` for me to figure that solution again in the future.

datalad/core/distributed/tests/test_clone.py
```diff
@@ -505,8 +506,13 @@ def test_cfg_originorigin(path):
     # Clone another level, this time with a relative path. Drop content from
     # lev2 so that origin is the only place that the file is available from.
     clone_lev2.drop("file1.txt")
-    with chpwd(path):
+    with chpwd(path), swallow_logs(new_level=9) as cml:
         clone_lev3 = clone('clone_lev2', 'clone_lev3')
+        # we called git-annex-init; see gh-4367:
+        cml.assert_logged(msg=r"[^[]*Running: \[('git', 'annex'|'git-annex'), "
```
kyleam, line 512: Hmm, so this is a little removed from testing the actual behavioral issue in gh-4367, but, given this rearrangement fixes the issue, it should be good enough to prevent a regression.

datalad/core/distributed/tests/test_clone.py
```diff
@@ -51,6 +51,7 @@
     ok_file_has_content,
     assert_not_in,
     assert_raises,
+    assert_re_in,
```
kyleam, line 54: This import is no longer needed with the `assert_logged` switch.  (Sorry, for unintentionally starting all this trouble way back in [3518][0]!  I really don't have a strong preference between using `assert_logged` or `assert_re_in`.)

Given the current testing run is well underway, I'd say it'd be fine to drop this import once the tests complete and then do a merge without the full test cycle.

[0]: https://github.com/datalad/datalad/pull/3518#discussion_r301685701

bpoldrack: Thx!

kyleam:
Looks like this interacts badly with master:

https://travis-ci.org/github/datalad/datalad/jobs/669818777#L1158

I'll try to take a look later today.

PR #4389, opened 2020-04-09
DOC: Custom Favicon to docs.datalad.org

adswa:
Following up on https://github.com/datalad-handbook/book/issues/407, here is a PR that would replace the readthedocs favicon with a custom one (the same as the  [DataLad](https://www.datalad.org) website), if you want.


adswa:
(I have cancelled everything I was able to cancel. Docs currently fail because of the recent sphinx update. I can retrigger the test after  #4386)

kyleam:
> I can retrigger the test after #4386

I tried just hitting the rerun button, and the docbuild failed in the setup stage.  Rebased to latest master (with sphinx compatibility fix) to try to get into a good state.

Issue #4404, opened 2020-04-15
`--jobs=auto` should consider more than just CPU availability


mih:
@adswa just encountered an issue of a failing `datalad get` with defaults, seemingly due to the number of file descriptors hitting the hard limit (machine has 32 cores).

For me this is another reason (besides https://github.com/datalad/datalad/issues/4141) to dislike the default behavior.

adswa:
It would be cool to not encounter this error, I think its really hard to make sense of from a user perspective. 

<details><summary>Here is how that looked like for me for the first get</summary>

```
datalad get .
                                                                                [INFO   ] To obtain some keys we need to fetch an archive of size 825.7 MB 
[INFO   ] To obtain some keys we need to fetch an archive of size 459.0 MB 
                                                                                [INFO   ] To obtain some keys we need to fetch an archive of size 31.2 MB 
                                                                                [INFO   ] To obtain some keys we need to fetch an archive of size 46.7 MB 
                                                                                [INFO   ] To obtain some keys we need to fetch an archive of size 31.2 MB 
                                                                                [INFO   ] To obtain some keys we need to fetch an archive of size 46.7 MB 
                                                                                [INFO   ] To obtain some keys we need to fetch an archive of size 31.2 MB 
Total (295 ok, 605 failed out of 863):  49%|▍| 2.01G/4.14G [15:09<01:45, 20.2MB/[WARNING] Running get resulted in stderr output: [INFO] To obtain some keys we need to fetch an archive of size 459.0 MB 
[INFO] To obtain some keys we need to fetch an archive of size 825.7 MB 
[...]

Issue #4141, opened 2020-02-13
datalad get (with default settings is an aggressive citizen)


mih:
I did a `datalad get .` in a dataset with a few hundred files (~2GB in total) on my laptop. I then had to watch the glorious progress bars, and could enjoy the speedy download -- but I also had no choice, because the machine was way too busy to do anything else. Not sure if "excuse-me-I-am-downloading" is a good default ;-)

![image](https://user-images.githubusercontent.com/136479/74410536-95b52300-4e39-11ea-9cbd-29e4c5cf90d5.png)

there are actually more git annex processes running than are visible in the screenshot.

PR #4409, opened 2020-04-17
RF: Make default number of annex jobs configurable and 1 by default

mih:
The previous default behavior `(max(8, min(3, nCPUs)))` could cause
issues (crashes, system stalls) due to the large number of child
processes spawned and file descriptors opened.

This change introduces a new config switch
`datalad.runtime.max-annex-jobs` that can be used to tailor
parallelization behavior, and effectively replaces the previously
hard-coded maximum value of `8`. Moreover, it drops the default to `1`
to disable parallel annex jobs by default to ensure a potentially
slower, but more robust behavior out of the box.

The config switch is specific to annex jobs in order to leave room for
an anticipated parallelization of some functionality in DataLad itself.
This will likely need separate configuration to be able to handle the
multiplicative impact of parallelization on two levels.

Implementation detail: The number of jobs is now lazily evaluated on
first use. While it only takes ~100ns to get the CPU count, adding a
config query would make this longer (still microsecond range). However,
the configuration scope would then be limited to the user-level, and
could not be tailored to individual datasets, which may be desired for
datasets with a huge number of files.

I left the previously defined `N_AUTO_JOBS` symbol in place to avoid
breaking consumer code (I am not aware of any). But it now also reflects
the changed default behavior.

Fixes gh-4404
Fixes gh-4141

PR is targeting `maint` based on this comment: https://github.com/datalad/datalad/issues/4404#issuecomment-614043733

### DIFF
--- a/datalad/interface/common_cfg.py
+++ b/datalad/interface/common_cfg.py
@@ -309,8 +309,8 @@ definitions = {
     },
     'datalad.runtime.max-annex-jobs': {
         'ui': ('question', {
-               'title': 'Maximum number of git-annex jobs to request',
-               'text': 'Set this value to enable parallel annex jobs that may speed up certain operations (e.g. get file content). The effective number of jobs will not exceed the number of available CPU cores.'}),
+               'title': 'Maximum number of git-annex jobs to request when "jobs" option set to "auto" (default)',
+               'text': 'Set this value to enable parallel annex jobs that may speed up certain operations (e.g. get file content). The effective number of jobs will not exceed the number of available CPU cores (or 3 if there is less than 3 cores).'}),
         'type': EnsureInt(),
         'default': 1,
     },
--- a/datalad/interface/common_cfg.py
+++ b/datalad/interface/common_cfg.py
@@ -307,6 +307,13 @@ definitions = {
                'text': 'Git-annex large files expression (see https://git-annex.branchable.com/tips/largefiles; given expression will be wrapped in parentheses)'}),
         'default': 'anything',
     },
+    'datalad.runtime.max-annex-jobs': {
+        'ui': ('question', {
+               'title': 'Maximum number of git-annex jobs to request',
+               'text': 'Set this value to enable parallel annex jobs that may speed up certain operations (e.g. get file content). The effective number of jobs will not exceed the number of available CPU cores.'}),
+        'type': EnsureInt(),
+        'default': 1,
+    },
     'datalad.runtime.raiseonerror': {
         'ui': ('question', {
                'title': 'Error behavior',
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -98,8 +98,9 @@ from .exceptions import (
 
 lgr = logging.getLogger('datalad.annex')
 
-# Limit to # of CPUs and up to 8, but at least 3 to start with
-N_AUTO_JOBS = min(8, max(3, cpu_count()))
+# TODO Constant is no longer used, but left defined to avoid breakage in
+# dependent code. Remove in 0.14 release.
+N_AUTO_JOBS = 1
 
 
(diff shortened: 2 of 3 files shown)


bpoldrack:
LGTM

kyleam:
Untested, but looks good to me as well.

datalad/interface/common_cfg.py
```diff
@@ -307,6 +307,13 @@
                'text': 'Git-annex large files expression (see https://git-annex.branchable.com/tips/largefiles; given expression will be wrapped in parentheses)'}),
         'default': 'anything',
     },
+    'datalad.runtime.max-annex-jobs': {
```
yarikoptic, line 310: I think it should be `datalad.annex.jobs` because it is annex specific setting, and pretty much we just pass it into annex call, similarly to the `retry`, whenever `runtime` is more about our own behavior:
```
$> git grep datalad.annex datalad/interface/common_cfg.py
datalad/interface/common_cfg.py:    'datalad.annex.retry': {

$> git grep datalad.runtime datalad/interface/common_cfg.py
datalad/interface/common_cfg.py:    'datalad.runtime.raiseonerror': {
datalad/interface/common_cfg.py:    'datalad.runtime.report-status': {
```

datalad/support/annexrepo.py
```diff
@@ -2144,7 +2148,14 @@ def _run_annex_command_json(self, command,
             annex_options += ['--json-progress']
 
         if jobs == 'auto':
-            jobs = N_AUTO_JOBS
+            # Limit to # of CPUs (but at least 3 to start with)
+            # and also an additional config constraint (by default 1
+            # due to https://github.com/datalad/datalad/issues/4404)
+            jobs = self._n_auto_jobs or min(
+                self.config.obtain('datalad.runtime.max-annex-jobs'),
+                max(3, cpu_count()))
```
yarikoptic, line 2156: why bother limiting here more from what user requested? I could have 1 cpu but download in parallel could still be beneficial (it is not CPU bound job).  I would just pass the value of the configuration here.
Although indeed it kinda deminishes the notion of any "auto" here -- it pretty much becomes "config", it is worth adjusting documentation where jobs is mentioned that it wouldn't be "auto" but rather consult config variable, or better change from "auto" to "config", and then here leave it all as it was (i.e. for "auto" - do previous behavior).  So by default it would consult config - would become 1.

yarikoptic: This way, later, if we could come up with some true "auto" we could change back while leaving possibility to still use "config" setting.

mih:
I am not confident that I understand your intentions. My intentions were to change the previous behavior as little as possible, while getting the issues fixed (hence I target `maint`). You seem to suggest that more fundamental changes should be made (new config option, direct pass-through).

I disagree with

> I think it should be datalad.annex.jobs because it is annex specific setting, and pretty much we just pass it into annex call, similarly to the retry, whenever runtime is more about our own behavior

Datalad is doing "auto" not git-annex, and we are applying additional decision-making before coming up with a modified value to pass on, hence my choice of the variable name.

Overall, I think the current state of the PR is not really in conflict with your thoughts, but rather is the BF part of them, that could go into `maint/`. Feel free to take this PR over.

yarikoptic:
Ok, if we are to pursue "minimal change" approach here, I just pushed two commits adjusting description of the newly added configuration item (to make it more factually correct) and `--jobs` parameter.  But my point was: I do not see much value in doing all the `min(,. max())` dance if we now would be simply providing an explicit value, and making it more explicit - "--jobs argument to pass to annex invocation" (and thus renaming this configuration option) would have made things clear(er) not only for the sake of `maint` but for the future release as well.

mih:
@yarikoptic Thx for the edits! I am OK with changing the default to an explicitly configured number in a second step. Will merge this for now.

Issue #1253, opened 2017-02-03
Why is it important to be notified when URL quoting happens?


mih:
From my logs:

````
[WARNING] Parsed version of URL 'http://kumo.ovgu.de/%7Emih/demoannex' differs from original 'http://kumo.ovgu.de/~mih/demoannex' 
````

Do we need that? With a warning? Can we go to defcon 2?

mih:
Please let me know whether this can be silenced @bpoldrack @yarikoptic

yarikoptic:
may be move it indeed to DEBUG level for now... theoretically str of parsed should always be identical to original but in this case (explicit `~` vs encoded) - not sure, thus I guess it is safe.  but there could be spots where such warnings would hint to omitted  portion of the url... well -- we could rely on tests of cause ;-)

PR #4414, opened 2020-04-17
RF: Demote log message and make generation conditional

mih:
Pretty much any recursive operation will yield log messages like:

```
Parsed version of PathRI '/tmp/crcns/vim-2' differs from original PosixPath('/tmp/crcns/vim-2')
```

at the `debug` level. Not only are they largely uninformative, they are
also inconsequential.

Here I make the processing that leads to that log message conditional on
that log level being actually active. This is a frequently used code
path, so it makes sense to pay a bit more attention to performance.

Additionally, due to the nature of the test and message, I demote the
log level from `debug` to `5`, to match other inconsequential log messages
in the RI code.

This cuts down the logging caused by, for example,
`subdatasets(recursive=True)` dramatically, and improves comprehension
what is actually going on.

### DIFF
--- a/datalad/support/network.py
+++ b/datalad/support/network.py
@@ -443,13 +443,13 @@ class RI(object):
         self._set_from_fields(**fields)
 
         # If was initialized from a string representation
-        if self._str is not None:
+        if lgr.isEnabledFor(5) and self._str is not None:
             # well -- some ris might not unparse identically back
             # strictly speaking, but let's assume they do
             ri_ = self.as_str()
             if ri != ri_:
-                lgr.debug("Parsed version of %s %r differs from original %r",
-                          self.__class__.__name__, ri_, ri)
+                lgr.log(5, "Parsed version of %s %r differs from original %r",
+                        self.__class__.__name__, ri_, ri)
 
     @classmethod
     def _get_blank_fields(cls, **fields):
--- a/datalad/support/tests/test_network.py
+++ b/datalad/support/tests/test_network.py
@@ -150,7 +150,7 @@ def test_url_eq():
 
 def _check_ri(ri, cls, exact_str=True, localpath=None, **fields):
     """just a helper to carry out few checks on urls"""
-    with swallow_logs(new_level=logging.DEBUG) as cml:
+    with swallow_logs(new_level=5) as cml:
         ri_ = cls(**fields)
         murl = RI(ri)
         eq_(murl.__class__, cls)  # not just a subclass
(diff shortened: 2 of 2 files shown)


yarikoptic:
This log message is "inconsequential" until it becomes very consequential whenever our URL parsing/recomposition results in a url leading somewhere else than expected! Overall lowering it would just possibly bury indication of the problem too deep to easily be discoverable.
In this particular case, it just feels that PathRI should just become aware (before I believe it was all string-centered) about `PosixPath` and such particular messages then would not be logged AT ALL.  So IMHO this PR is approaching it from a wrong angle.

kyleam:
@yarikoptic Good points.

> it just feels that PathRI should just become aware (before I believe it was all string-centered) about `PosixPath` and such particular messages then would not be logged AT ALL.

Perhaps this would be sufficient (in terms of the underlying purpose of the check; it is sufficient to prune the log messages that this PR is about)?  Or do you think we should check for a path object explicitly?

```diff
diff --git a/datalad/support/network.py b/datalad/support/network.py
index b8ab935e58..135d82b9fd 100644
--- a/datalad/support/network.py
+++ b/datalad/support/network.py
@@ -447,7 +447,7 @@ def __init__(self, ri=None, **fields):
             # well -- some ris might not unparse identically back
             # strictly speaking, but let's assume they do
             ri_ = self.as_str()
-            if ri != ri_:
+            if str(ri) != ri_:
                 lgr.debug("Parsed version of %s %r differs from original %r",
                           self.__class__.__name__, ri_, ri)
 
```

mih:
My point was that whether or not a back-conversion worked, the only consequence was a log message. That costs cycles that were unconditionally spent.

kyleam:
> This log message is "inconsequential" until it becomes very consequential whenever our URL parsing/recomposition results in a url leading somewhere else than expected!

On the other hand, when you phrase it like that, it seems like this should be at a logging level that would be shown by default, not DEBUG.

yarikoptic:
> My point was that whether or not a back-conversion worked, the only consequence was a log message. That costs cycles that were unconditionally spent.

indeed, all that code could be avoided to be ran if not to be logged.   Might be worth it, but I would probably call it "over optimization" since it is very unlikely to make any measurable difference.  I don't mind it though ;)  Also the issue is orthogonal from demotion below DEBUG.

> On the other hand, when you phrase it like that, it seems like this should be at a logging level that would be shown by default, not DEBUG.

it was a warning, then demoted to DEBUG to similarly "fix" #1253. Here @mih strikes again trying to bring it further down ;-) But situation there was trickier -- here indeed simpler solution might be possible:

> Perhaps this would be sufficient (in terms of the underlying purpose of the check; it is sufficient to prune the log messages that this PR is about)? Or do you think we should check for a path object explicitly?

docstring for `ri` Parameter still says it should be a string  and the value is used only passed into `_str_to_fields` ;)  so may be we should just cast it to string at the entry and that should probably do it.

kyleam:
> > Perhaps this would be sufficient (in terms of the underlying purpose of the check; it is sufficient to prune the log messages that this PR is about)? Or do you think we should check for a path object explicitly?
> 
> docstring for `ri` Parameter still says it should be a string

Sure, that would be something to be updated in a proper patch.

> and the value is used only passed into `_str_to_fields` ;)  so may be we should just cast it to string at the entry and that should probably do it.

Also fine.  It was a quick, minimal-diff patch for the purpose of asking whether you thought the blanket `str` casting was sufficient.

mih:
Closing in favor of #4423

PR #4420, opened 2020-04-17
RF: <Dataset path=...> -> <Dataset ...>

mih:
__repr__ of a dataset always uses the 'path=' prefix to the path, but a
dataset is also always and only identified by a path. Hence the prefix
is superfluous, and only takes space. On any platform an absolute path
is easy to recognize as such.

This change removes the 'path=' prefix.

### DIFF
--- a/datalad/distribution/dataset.py
+++ b/datalad/distribution/dataset.py
@@ -48,6 +48,7 @@ from datalad.utils import (
     Path,
     PurePath,
     assure_list,
+    quote_cmdlinearg,
 )
 
 
@@ -148,7 +149,7 @@ class Dataset(object, metaclass=PathBasedFlyweight):
         return self._pathobj
 
     def __repr__(self):
-        return "<Dataset path=%s>" % self.path
+        return 'Dataset({})'.format(quote_cmdlinearg(self.path))
 
     def __eq__(self, other):
         if not hasattr(other, 'pathobj'):


yarikoptic:
YES, PLEASE! ;)  I always hated it but was too shy to suggest stripping it away -- one pros will be that you could just double click on the path to select it in the terminal, where now it would also grab `path=` prefix

kyleam:
> one pros will be that you could just double click on the path to select it in the terminal

I'm looking forward to that too :)  In the debugger, I often type `<dataset>.path` so that I can get something I can double-click on.

kyleam:
I'd be happy to see this change land as is, but even better to my eyes would be a format like `Dataset('/path/to/ds')`, which I probably prefer just because it is a more common format for `__repr__`s.

mih:
So what about this then (to really get to a C&P-able state):

```py
>>> print('Dataset({})'.format(quote_cmdlinearg("/home/mih/this is fun/dataset")))       
Dataset('/home/mih/this is fun/dataset')

>>> print('Dataset({})'.format(quote_cmdlinearg("/home/mih/that's it folks/dataset")))   
Dataset('/home/mih/that'"'"'s it folks/dataset')
```

kyleam:
Looks nice to me.

yarikoptic:
Yes!

mih:
Force-pushed this change.

Issue #4425, opened 2020-04-19
[WARNING] realpath of PWD=... is ... whenever os.getcwd()=...


mih:
https://github.com/datalad-handbook/book/pull/438 will add a statement to the handbook why this warning is emitted, and why people should train themselves to ignore it.

I think this should not be a warning, but a debug message. There is nothing any user could or should be doing about it AFAICS.

Issue #3442, opened 2019-05-20
local to subdatasets paths in action status report are misleading


yarikoptic:
if before (0.11.x) it was suboptimal due to full paths:
```
(git-annex)hopa:/tmp/testds-remove[master]git-annex
$> echo 123 >> subds/subds12/subds/ingit 
(dev) 1 14595 [2].....................................:Mon 20 May 2019 06:38:30 PM EDT:.
(git-annex)hopa:/tmp/testds-remove[master]git-annex
$> datalad save -r -d . subds/subds12/subds/ingit
save(ok): /tmp/testds-remove/subds/subds12/subds (dataset)
save(ok): /tmp/testds-remove/subds/subds12 (dataset)
save(ok): /tmp/testds-remove/subds (dataset)
save(ok): /tmp/testds-remove (dataset)
action summary:
  save (ok: 4)
```
Now (0.12.0rc4.dev5) it is easier on eyes but without consideration of the "context" from which the command was called, reported paths within subdatasets could be confusing, wrong, or misleading:

```
(git-annex)hopa:/tmp/testds-remove[master]git-annex
$> echo 123 >> subds/subds12/subds/ingit
(dev) 1 14599 [2].....................................:Mon 20 May 2019 06:39:20 PM EDT:.
(git-annex)hopa:/tmp/testds-remove[master]git-annex
$> datalad save -r -d . subds/subds12/subds/ingit
add(ok): ingit (file)
save(ok): subds/subds12/subds (dataset)
add(ok): subds (file)
save(ok): subds/subds12 (dataset)
add(ok): subds12 (file)
save(ok): subds (dataset)
add(ok): subds (file)
save(ok): . (dataset)
action summary:
  add (ok: 4)
  save (ok: 4)
```
not to mention that I am not sure what "add(ok): subds (file)" means since I do not have `subds` file in this example.

mih:
Thank you for reporting this issue!

However, at this point I am closing it at a time of ~600 open issues, in an attempt to regain control over the issue tracker. This particular issue was posted more than 6 months ago, has not received a single response, and does not describe an immediate software defect. *Closing this issue does not imply that it is not relevant*. If any reader objects the closing of this issue, please feel more than free to reopen it with an update on how it is relevant for the current state of the project, and which concrete actions to address it are ongoing or planned.

PR #4426, opened 2020-04-19
RF: Demote symlinked CWD related log message to debug-level

mih:
This is the message:

`[WARNING] realpath of PWD=</some/path/to/symlinked/file> is </resolved/path/to/symlinked/file> whenever os.getcwd()=<path/to/dataset>`

Questions about it have motivated a dedicated "please ignore it" PR
in the handbook: https://github.com/datalad-handbook/book/pull/438

I think this message is pointless (but still unconditionally shown) if
everything works. When something doesn't work, it is likely caused by
a bug in DataLad. Hence there is no need to bother users with this
warning that they can or should do nothing about.

Fixes gh-4425

### DIFF
--- a/datalad/tests/test_utils.py
+++ b/datalad/tests/test_utils.py
@@ -377,7 +377,7 @@ def test_getpwd_change_mode(tdir):
     # The evil plain chdir call
     os.chdir(tdir)
     # Just testing the logic of switching to cwd mode and issuing a warning
-    with swallow_logs(new_level=logging.WARNING) as cml:
+    with swallow_logs(new_level=logging.DEBUG) as cml:
         pwd = getpwd()
         eq_(pwd, str(Path(pwd).resolve()))  # might have symlinks, thus realpath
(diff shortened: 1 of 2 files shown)


kyleam:
I think demoting this would be good, but perhaps I'm not thinking of some nasty case that makes the noise/user confusion worth it in the other cases.

In gh-2910, @yarikoptic wrote

> in getpwd verify that envvar PWD and os.getpwd() point to the same directory (after realpath'ing it), and if not - return os.getpwd() (with some warning?).

The last bit suggests he was on the fence about the warning at the time.  Not sure if he ended up with a clear preference after working on the PR.

yarikoptic:
also it was added when we largely returned absolute paths, now we more often return relative (too relative IMHO, see #3442) so indeed there is less sources of confusion we tried to avoid by issuing this warning.  So FWIW I am ok to lower it to DEBUG thus let's go

PR #4430, opened 2020-04-21
NF: `copy_file()` command

### LINKED ISSUE #600: Merge data availability info from detached datasets
Example

```
% git clone https://github.com/psychoinformatics-de/studyforrest-data-phase2
% cd studyforrest-data-phase2
% git annex init
% git annex whereis sub-01/ses-localizer/func/sub-01_ses-localizer_task-objectcategories_run-1_bold.nii.gz

  Failed to get annex.uuid configuration of repository origin
whereis sub-01/ses-localizer/func/sub-01_ses-localizer_task-objectcategories_run-1_bold.nii.gz (2 copies) 
    2f51364e-815a-4ad9-b380-617c682b1680 -- [mddatasrc]
    40809380-c618-4801-bb8d-81320474c9da -- studyforrest phase2 original source repository
ok
```

Now datalad also has this dataset, albeit with a Git history that is not based on the original repo, but an independently generated dataset based on openfmri tarballs. To get this as an additional data source, we can do this:

```
% git remote add datalad http://datasets.datalad.org/openfmri/ds000113d/.git
% git fetch datalad
% git annex merge
% git annex whereis sub-01/ses-localizer/func/sub-01_ses-localizer_task-objectcategories_run-1_bold.nii.gz
  Failed to get annex.uuid configuration of repository origin
whereis sub-01/ses-localizer/func/sub-01_ses-localizer_task-objectcategories_run-1_bold.nii.gz (4 copies) 
    2f51364e-815a-4ad9-b380-617c682b1680 -- [mddatasrc]
    40809380-c618-4801-bb8d-81320474c9da -- studyforrest phase2 original source repository
    590d0da0-e791-4848-a09e-a47c70939758 -- datalad-archives
    c1be0852-f8e7-4b2b-8809-14ce4dd170f9 -- yoh@smaug:/mnt/btrfs/datasets/datalad/crawl/openfmri/ds000113d
ok
```

This is a seemingly clean way to add new data sources without having to interfere with the appearance/history of a dataset. Afterwards the remote could safely be removed again (if desired) without loosing data availability info.

### LINKED ISSUE #3681: [wishlist] copy/move command to support copying annex availability/metadata across repos
ATM it is not trivial to copy an annexed file(s) from one dataset to another, while trying to preserve information about its availability via URLs (from web and other remotes).    Use cases are

- copying a subset of files from a large repo into a smaller / more targetted one
- moving files from one dataset into another (e.g. withing a hierarchy of datasets)

`move` is just a special case which would entail running `remove` after `copy`, although still desiring that `--no-drop` for remove for the majority of the use cases I see (https://github.com/datalad/datalad/issues/2673)

Following actions should be done when copying files from one repo to another:

- annex level metadata should be copied as well
- register those urls from the original repository in the target repository for annexed file
- If some URLs are handled via special remotes, we should also init/enable those remotes in the target repo.  If a remote with such a name already exists, we could verify if parameters are the same, and if not -- warning/crash/ignore (config var) and/or inform the parameters to the user

### http(s)
Usecase from @satra is to get/use a few files from freesurfer repo.  It is not pragmatic to just add that repo as a remote so annex could fetch files from it since that repo's .git/objects is already >600MB.  That usecase is a bit trickier:  files do not have URLs associated with them, but the remote is available via http (https://surfer.nmr.mgh.harvard.edu/pub/dist/freesurfer/repo/freesurfer.git) and the actual load is somewhere there under [annex/objects](https://surfer.nmr.mgh.harvard.edu/pub/dist/freesurfer/repo/freesurfer.git/annex/objects/) but because it is a bare repo special remote (ref: [annex docs](http://git-annex.branchable.com/internals/hashing/)), annex/objects cache paths use different layout from the one on a clone on a regular clone. 
```bash
> git annex examinekey --format='${hashdirlower}' SHA256E-s200057528--a4da07a728f5cfac9c20e600e0c2ca0417ebd4e92f89aca1faddb11793b12c7a.tar.gz
958/132/% 
```
<details>
<summary>so `examinekey` could be used to deduce the cache dir and compose the url :</summary> 

```shell
$> wget 'https://surfer.nmr.mgh.harvard.edu//pub/dist/freesurfer/repo/annex.git//annex/objects/958/132/SHA256E-s200057528--a4da07a728f5cfac9c20e600e0c2ca0417ebd4e92f89aca1faddb11793b12c7a.tar.gz/SHA256E-s200057528--a4da07a728f5cfac9c20e600e0c2ca0417ebd4e92f89aca1faddb11793b12c7a.tar.gz'
--2019-09-16 09:45:36--  https://surfer.nmr.mgh.harvard.edu//pub/dist/freesurfer/repo/annex.git//annex/objects/958/132/SHA256E-s200057528--a4da07a728f5cfac9c20e600e0c2ca0417ebd4e92f89aca1faddb11793b12c7a.tar.gz/SHA256E-s200057528--a4da07a728f5cfac9c20e600e0c2ca0417ebd4e92f89aca1faddb11793b12c7a.tar.gz
Resolving surfer.nmr.mgh.harvard.edu (surfer.nmr.mgh.harvard.edu)... 132.183.240.105
Connecting to surfer.nmr.mgh.harvard.edu (surfer.nmr.mgh.harvard.edu)|132.183.240.105|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 200057528 (191M) [application/x-gzip]
Saving to: ‘SHA256E-s200057528--a4da07a728f5cfac9c20e600e0c2ca0417ebd4e92f89aca1faddb11793b12c7a.tar.gz’

    SHA256E-s200057528--a4da07a728f5cfac9   4%[==>   
```
</details>

may be there is even more straightforward way to obtain such URLs directly from git-annex? (attn: @joeyh)  i.e. for every file to be `get`'ed to just report the urls from which it could be obtained (possibly including ssh, see below)

### ssh

In some cases (probably primarily for private repos without data being shared publicly), the only remotes carrying data would be clones over ssh. So instead of `http*` urls, `ssh://` urls should be created. git-annex ATM does not natively support them, but there is a TODO: https://git-annex.branchable.com/todo/support_ssh__58____47____47___or_sftp__58____47____47___urls_via___34__built-in__34___ssh_support/  so in principle/future it should be feasible.

mih:
This command is an approach to fix #3681 and fix #600. However, instead of lumping things together and having git-annex sort things out, this is a more targeted attempt at picking individual files, looking for their availability info, and transfer/enable necessary special remote configuration.

This can only be considered a first step. Many (speed) optimizations are possible, various features could be added (e.g. arbitrary special remote support, for now only `datalad` remote). However, it is reasonably complete and functional to a degree that it is capable of filtering a dataset like https://github.com/datalad-datasets/human-connectome-project-openaccess

##### Acknowledgements
This work was, in part, supported by the European Union's Horizon 2020 Research and Innovation Programme under Grant Agreement no. 785907 (HBP SGA2).

### DIFF
--- a/datalad/local/copy_file.py
+++ b/datalad/local/copy_file.py
@@ -53,16 +53,17 @@ class CopyFile(Interface):
             # not really needed on the cmdline, but for PY to resolve relative
             # paths
             args=("-d", "--dataset"),
-            doc="""target dataset to save copied files into. This may be a
-            superdataset containing the actual destination dataset. In this case,
-            any changes will be save up to this dataset.[PY: This dataset is also
-            the reference for any relative paths. PY].""",
+            doc="""root dataset to save after copy operations are completed.
+            All destination paths must be within this dataset, or its
+            subdatsets. [PY: This dataset is also the reference for any relative
+            paths. PY] If no dataset is given, dataset modifications will be
+            left unsaved.""",
             constraints=EnsureDataset() | EnsureNone()),
         path=Parameter(
             args=("path",),
             metavar='PATH',
             doc="""paths to copy (and possibly a target path to copy to).""",
-            nargs='+',
+            nargs='*',
             constraints=EnsureStr() | EnsureNone()),
         recursive=Parameter(
             args=("--recursive", "-r",),
@@ -71,12 +72,12 @@ class CopyFile(Interface):
         target_dir=Parameter(
             args=('--target-directory', '-t'),
             metavar='DIRECTORY',
-            doc="""copy all PATH arguments into DIRECTORY""",
+            doc="""copy all source files into this DIRECTORY""",
             constraints=EnsureStr() | EnsureNone()),
         specs_from=Parameter(
             args=('--specs-from',),
             metavar='SOURCE',
-            doc="""read list of source and destination path names from a given
+            doc="""read list of source (and destination) path names from a given
             file, or stdin (with '-'). Each line defines either a source
             path, or a source/destination path pair (separated by a null byte
             character).[PY:  Alternatively, a list of 2-tuples with
--- a/datalad/local/copy_file.py
+++ b/datalad/local/copy_file.py
@@ -96,9 +96,11 @@ class CopyFile(Interface):
         target_dir=Parameter(
             args=('--target-dir', '-t'),
             metavar='DIRECTORY',
-            doc="""copy all source files into this DIRECTORY. This value
-            is overridden by any explicit destination path provided via
-            [CMD: --specs-from CMD][PY: 'specs_from' PY].""",
+            doc="""copy all source files into this DIRECTORY. This value is
+            overridden by any explicit destination path provided via [CMD:
+            --specs-from CMD][PY: 'specs_from' PY]. When not given, this
+            defaults to the path of the dataset specified via [CMD: --dataset
+            CMD][PY: 'dataset' PY].""",
             constraints=EnsureStr() | EnsureNone()),
         specs_from=Parameter(
             args=('--specs-from',),
--- a/datalad/local/copy_file.py
+++ b/datalad/local/copy_file.py
@@ -73,7 +73,9 @@ class CopyFile(Interface):
         target_dir=Parameter(
             args=('--target-dir', '-t'),
             metavar='DIRECTORY',
-            doc="""copy all source files into this DIRECTORY""",
+            doc="""copy all source files into this DIRECTORY. This value
+            is overwritten by any explicit destination path provided via
+            [CMD: --specs-from CMD][PY: 'specs_from' PY].""",
             constraints=EnsureStr() | EnsureNone()),
         specs_from=Parameter(
             args=('--specs-from',),
@@ -405,7 +407,7 @@ def _copy_file(src, dest, cache):
     if finfo.get('has_content', True):
         dest_key = dest_repo.call_git_oneline(['annex', 'calckey', str_src])
     else:
-        lgr.warning(
+        lgr.debug(
             'File content not available, forced to reuse previous annex key: %s',
             str_src)
         dest_key = src_key
--- a/datalad/local/copy_file.py
+++ b/datalad/local/copy_file.py
@@ -165,7 +165,9 @@ class CopyFile(Interface):
         try:
             for src_path, dest_path in _yield_specs(specs_from):
                 src_path = Path(src_path)
-                dest_path = None if dest_path is None else Path(dest_path)
+                dest_path = None \
+                if dest_path is None \
+                else resolve_path(dest_path, dataset)
                 lgr.debug('Processing copy specification: %s -> %s',
                           src_path, dest_path)
                 if not recursive and src_path.is_dir():
--- a/datalad/local/copy_file.py
+++ b/datalad/local/copy_file.py
@@ -498,7 +498,7 @@ def _replace_file(str_src, dest, str_dest, follow_symlinks):
         dest.unlink()
     else:
         dest.parent.mkdir(exist_ok=True, parents=True)
-    copyfile(str_src, str_dest, follow_symlinks=False)
+    copyfile(str_src, str_dest, follow_symlinks=follow_symlinks)
 
 
 def _extract_special_remote_info(repo):
--- a/datalad/local/copy_file.py
+++ b/datalad/local/copy_file.py
@@ -147,6 +147,9 @@ class CopyFile(Interface):
                 # in all other cases we have a plain source list
                 specs_from = paths
 
+        if not specs_from:
+            raise ValueError("Neither `paths` nor `specs_from` given.")
+
         res_kwargs = dict(
             action='copy',
             logger=lgr,
--- a/datalad/local/copy_file.py
+++ b/datalad/local/copy_file.py
@@ -357,6 +357,10 @@ def _yield_src_dest_filepaths(src, dest, src_base=None, target_dir=None):
     src, dest
       Path instances
     """
+    if src.name == '.git':
+        # we never want to copy the git repo internals into another repo
+        # this would break the target git in unforseeable ways
+        return
     if src.is_dir():
         # special case: not yet a file to copy
         if src_base is None:
--- a/datalad/local/copy_file.py
+++ b/datalad/local/copy_file.py
@@ -401,6 +401,12 @@ def _copy_file(src, dest, cache):
             str_src)
         dest_key = src_key
 
+    if op.lexists(str_dest):
+        # if the target already exists, we remove it first, because we want to
+        # modify this path (potentially pointing to a new key), rather than
+        # failing next on 'fromkey', due to a key mismatch.
+        # this is more compatible with the nature of 'cp'
+        dest.unlink()
     dest_repo._run_annex_command_json(
         'fromkey',
         # we use force, because in all likelihood there is no content for this key
--- a/datalad/local/copy_file.py
+++ b/datalad/local/copy_file.py
@@ -286,6 +286,9 @@ def _yield_src_dest_filepaths(src, dest, src_base=None, target_dir=None):
             yield from _yield_src_dest_filepaths(p, dest, src_base, target_dir)
         return
 
+    if target_dir is None and not dest:
+        raise ValueError("Neither target_dir nor dest specified")
+
     if not dest:
         # no explicit destination given, build one from src and target_dir
         # reflect src hierarchy if dest is a directory, otherwise
--- a/datalad/local/copy_file.py
+++ b/datalad/local/copy_file.py
@@ -251,6 +251,15 @@ class CopyFile(Interface):
                     )
                     continue
 
+                if dest_path and dest_path.name == '.git' or src_path.name == '.git':
+                    yield dict(
+                        path=str(src_path),
+                        status='impossible',
+                        message="refuse to place '.git' into destination dataset",
+                        **res_kwargs
+                    )
+                    continue
+
                 if not (dest_path or target_dir):
                     yield dict(
                         path=str(src_path),
@@ -357,11 +366,12 @@ def _yield_src_dest_filepaths(src, dest, src_base=None, target_dir=None):
     src, dest
       Path instances
     """
-    if src.name == '.git':
-        # we never want to copy the git repo internals into another repo
-        # this would break the target git in unforseeable ways
-        return
     if src.is_dir():
+        # we only get here, when recursion is desired
+        if src.name == '.git':
+            # we never want to copy the git repo internals into another repo
+            # this would break the target git in unforseeable ways
+            return
         # special case: not yet a file to copy
         if src_base is None:
             # TODO maybe an unconditional .parent isn't a good idea,
--- a/datalad/local/copy_file.py
+++ b/datalad/local/copy_file.py
@@ -74,7 +74,7 @@ class CopyFile(Interface):
             args=('--target-dir', '-t'),
             metavar='DIRECTORY',
             doc="""copy all source files into this DIRECTORY. This value
-            is overwritten by any explicit destination path provided via
+            is overridden by any explicit destination path provided via
             [CMD: --specs-from CMD][PY: 'specs_from' PY].""",
             constraints=EnsureStr() | EnsureNone()),
         specs_from=Parameter(
@@ -84,7 +84,7 @@ class CopyFile(Interface):
             file, or stdin (with '-'). Each line defines either a source
             path, or a source/destination path pair (separated by a null byte
             character).[PY:  Alternatively, a list of 2-tuples with
-            source/destination pairs can be given provided. PY]."""),
+            source/destination pairs can be given. PY]"""),
         message=save_message_opt,
     )
 
--- a/datalad/local/copy_file.py
+++ b/datalad/local/copy_file.py
@@ -58,8 +58,9 @@ class CopyFile(Interface):
     content is available locally, by just copying the required metadata on
     content identity and availability.
 
-    Note: URLs for only built-in 'web' and external 'datalad' special remotes
-    are copied ATM.
+    .. note::
+      At the moment, only URLs for the special remotes 'web' (git-annex built-in)
+      and 'datalad' are recognized and transferred.
 
     || REFLOW >>
     The interface is modeled after the POSIX 'cp' command, but with one
--- a/datalad/local/copy_file.py
+++ b/datalad/local/copy_file.py
@@ -152,11 +152,10 @@ class CopyFile(Interface):
             datalad copy-file -d path/to/myds -t path/to/myds --specs-from specfile"""
         ),
         dict(
-            text="Read a specification from stdin: Pipe the output of a find "
-                 "command into the copy-file command",
-            code_py="",
+            text="Read a specification from stdin and pipe the output of a find"
+                 " command into the copy-file command",
             code_cmd="""\
-            find <expr> | datalad copy-file -d myds -t path/to/myds --specs-from -"""
+            find <expr> | datalad copy-file -d path/to/myds -t path/to/myds --specs-from -"""
         )
     ]
 
--- a/datalad/local/copy_file.py
+++ b/datalad/local/copy_file.py
@@ -47,7 +47,33 @@ lgr = logging.getLogger('datalad.local.copy_file')
 
 @build_doc
 class CopyFile(Interface):
-    """
+    """Copy files and their availability metadata from one dataset to another.
+
+    The difference to a system copy command is that here additional content
+    availability information, such as registered URLs, are also being copied to
+    the target dataset. Moreover, potentially required git-annex special remote
+    configurations are detected in a source dataset and are applied to a target
+    dataset in an analogous fashion. It is possible to copy a file for which no
+    content is available locally, by just copying the required metadata on
+    content identity and availability.
+
+    || REFLOW >>
+    Interface and behavior are modeled after the POSIX 'cp' command, with two
+    main differences. 1) Via [CMD: --specs-from CMD][PY: `specs_from` PY] it is
+    possible to flexibly input source-destination path pairs. 2) Saving changes
+    in a target dataset is conditional upon a specification of
+    [CMD: --dataset CMD][PY: `dataset`, or calling this command as a `Dataset`
+    method PY].
+    << REFLOW ||
+
+    || REFLOW >>
+    This command can copy files out of and into a hierarchy of nested datasets.
+    Unlike with other DataLad command, the [CMD: --recursive CMD][PY: `recursive`
+    PY] switch does not enable recursion into subdirectories, but is analogous
+    to the POSIX 'cp' command switch and enables subdirectory recursion, regardless
+    of dataset boundaries. It is not necessary to enable recursion in order to
+    save changes made to nested target subdatasets.
+    << REFLOW ||
     """
     _params_ = dict(
         dataset=Parameter(
@@ -56,8 +82,7 @@ class CopyFile(Interface):
             args=("-d", "--dataset"),
             doc="""root dataset to save after copy operations are completed.
             All destination paths must be within this dataset, or its
-            subdatsets. [PY: This dataset is also the reference for any relative
-            paths. PY] If no dataset is given, dataset modifications will be
+            subdatasets. If no dataset is given, dataset modifications will be
             left unsaved.""",
             constraints=EnsureDataset() | EnsureNone()),
         path=Parameter(
--- a/datalad/local/copy_file.py
+++ b/datalad/local/copy_file.py
@@ -23,6 +23,7 @@ from datalad.support.constraints import (
     EnsureStr,
     EnsureNone,
 )
+from datalad.interface.common_opts import save_message_opt
 from datalad.support.param import Parameter
 from datalad.distribution.dataset import (
     Dataset,
@@ -82,6 +83,7 @@ class CopyFile(Interface):
             path, or a source/destination path pair (separated by a null byte
             character).[PY:  Alternatively, a list of 2-tuples with
             source/destination pairs can be given provided. PY]."""),
+        message=save_message_opt,
     )
 
     @staticmethod
@@ -91,9 +93,9 @@ class CopyFile(Interface):
             path=None,
             dataset=None,
             recursive=False,
-            # TODO needs message
             target_dir=None,
-            specs_from=None):
+            specs_from=None,
+            message=None):
         # Concept
         #
         # Loosely model after the POSIX cp command
@@ -219,6 +221,7 @@ class CopyFile(Interface):
             path=to_save,
             # we provide an explicit file list
             recursive=False,
+            message=message,
         )
 
 
--- a/datalad/local/copy_file.py
+++ b/datalad/local/copy_file.py
@@ -285,7 +285,11 @@ def _yield_src_dest_filepaths(src, dest, src_base=None, target_dir=None):
     if not dest:
         # no explicit destination given, build one from src and target_dir
         # reflect src hierarchy if dest is a directory, otherwise
-        dest = target_dir / (src.relative_to(src_base) if src_base else src.name)
+        if src.is_absolute():
+            dest = target_dir / (src.relative_to(src_base) if src_base else src.name)
+        else:
+            dest = target_dir / src
+
     yield src, dest
 
 
--- a/datalad/local/copy_file.py
+++ b/datalad/local/copy_file.py
@@ -412,49 +412,16 @@ def _copy_file(src, dest, cache):
         return
 
     # at this point we are copying an annexed file into an annex repo
-    src_key = finfo['key']
-
-    # make an attempt to compute a key in the target repo, this will hopefully
-    # become more relevant once "salted backends" are possible
-    # https://github.com/datalad/datalad/issues/3357 that could prevent
-    # information leakage across datasets
-    if finfo.get('has_content', True):
-        dest_key = dest_repo.call_git_oneline(['annex', 'calckey', str_src])
+    if dest_repo.is_managed_branch():
+        res = _place_filekey_managed(
+            finfo, str_src, dest, str_dest, dest_repo_rec)
     else:
-        lgr.debug(
-            'File content not available, forced to reuse previous annex key: %s',
-            str_src)
-        dest_key = src_key
-
-    if op.lexists(str_dest):
-        # if the target already exists, we remove it first, because we want to
-        # modify this path (potentially pointing to a new key), rather than
-        # failing next on 'fromkey', due to a key mismatch.
-        # this is more compatible with the nature of 'cp'
-        dest.unlink()
-    dest_repo._run_annex_command_json(
-        'fromkey',
-        # we use force, because in all likelihood there is no content for this key
-        # yet
-        opts=[dest_key, str_dest, '--force'],
-    )
-    if 'objloc' in finfo:
-        # we have the chance to place the actual content into the target annex
-        # put in a tmp location, git-annex will move from there
-        tmploc = dest_repo_rec.get('tmp', None)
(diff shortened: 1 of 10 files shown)


mih:
This is far from being "done", but a bunch of tests showing the intended behavior are in place. If anyone could give it a test ride and report how it feels that would be very useful! Thx.

kyleam:
> If anyone could give it a test ride and report how it feels that would be very useful!

The first thing I tried was

```sh
datalad create a
echo one >a/one
datalad -C a save
datalad create b
datalad copyfile a/one b/two
tree --charset=ascii a b | colrm 50
```

```
[...]
a
`-- one -> .git/annex/objects/zZ/KF/MD5E-s4--5bbf
b
`-- one -> .git/annex/objects/zZ/KF/MD5E-s4--5bbf

0 directories, 2 files
```

There might be good implementation or design reasons for that not being able to specify a target with a different name, but perhaps that being the first thing I tried suggests that it'd be worth a warning or documentation that the non-directory part of the target will be lopped off.

adswa:
> If anyone could give it a test ride and report how it feels that would be very useful!

I only played around with it a bit and did not look into the code, but here are the things that caught my attention:

**If specifying different target names is not possible, maybe warn/report better in accidental attempts of copying identically named files?**

After I saw @kyleam's comment, I wondered what would happen if I copied a file into a directory with a file of the same name but different contents

```
datalad create a
echo one >a/one
datalad -C a save
datalad create b
echo somethingelse > b/one
datalad -C b save
datalad copyfile a/one b/two
```

It would be cool if the warning could state something along the lines of "File ``b/one`` already exist, and is not identical - can't copy". Its a bit hard to find the command that is run in the output, and the ``Failed to clean up temporary directory`` along with the MD5E keys look so much like an internal error that it is hard to infer a mistake on the user side

```
create(ok): /tmp/a (dataset)
add(ok): one (file)
save(ok): . (dataset)
action summary:
  add (ok: 1)
  save (ok: 1)
[INFO   ] Creating a new annex repo at /tmp/b 
create(ok): /tmp/b (dataset)
add(ok): one (file)
save(ok): . (dataset)
action summary:
  add (ok: 1)
  save (ok: 1)
[WARNING] Running fromkey resulted in stderr output: git-annex: fromkey: 1 failed
 
[WARNING] Failed to clean up temporary directory: [Errno 39] Directory not empty: '/tmp/b/.git/tmp/datalad-copy' 
CommandError: 'git-annex reinject -c annex.dotfiles=true -c annex.retry=3 /tmp/b/.git/tmp/datalad-copy/MD5E-s4--5bbf5a52328e7439ae6e719dfe712200 /tmp/b/one' failed with exitcode 1 under /tmp/b
reinject /tmp/b/.git/tmp/datalad-copy/MD5E-s4--5bbf5a52328e7439ae6e719dfe712200 
failed
git-annex: /tmp/b/.git/tmp/datalad-copy/MD5E-s4--5bbf5a52328e7439ae6e719dfe712200 does not have expected content of /tmp/b/one
git-annex: reinject: 1 failed
``` 

--------------------------------------------------------------------------------------------------------------------------------------

#### Copying things into datasets without an annex -- maybe warn?

By chance, the very first thing I did was to copy an annexed file (with retrieved content) into a dataset with no annex (a few HCP subjects files into the top-level ``human-connectome-project-openacess`` dataset). Afterwards I end up with a broken symlink (understandable), but I wondered whether this info message is accurate (in particular: ``Copying non-annexed file or copy`` - it sounds like the json file would not be annexed, but it is)

```
─adina@muninn /tmp/hcp on master
╰─➤ datalad copyfile HCP1200/103111/.xdlm/103111_3T_rfMRI_REST1_fixextended.json .
[INFO   ] Copying non-annexed file or copy into non-annex dataset: /tmp/hcp/HCP1200/103111/.xdlm/103111_3T_rfMRI_REST1_fixextended.json -> <GitRepo path=/tmp/hcp (<class 'datalad.support.gitrepo.GitRepo'>)> 
copy(ok): /tmp/hcp/HCP1200/103111/.xdlm/103111_3T_rfMRI_REST1_fixextended.json
add(ok): 103111_3T_rfMRI_REST1_fixextended.json (file)
save(ok): . (dataset)
action summary:
  add (ok: 1)
  copy (ok: 1)
  save (ok: 1)
```

Would it be maybe good to warn when it is detected that annexed contents are copied into datasets without an annex? I'm thinking of someone copying a dataset with large data into a plain Git repository (e.g, maybe one for their code) and being not able to retrieve any data afterwards.

Also: When I `copyfile` an annexed file I didn't `get` yet into a non-annex dataset it fails, but the warning is misleading. To me, the warning would suggest that I would need to get content first, which is futile in this case - even if I did, copying it into a dataset without an annex will turn it into a broken symlink.
 
```
datalad copyfile HCP1200/103010/.xdlm/103010_3T_rfMRI_REST1_fixextended.json .
[INFO   ] Copying non-annexed file or copy into non-annex dataset: /tmp/hcp/HCP1200/103010/.xdlm/103010_3T_rfMRI_REST1_fixextended.json -> <GitRepo path=/tmp/hcp (<class 'datalad.support.gitrepo.GitRepo'>)> 
copy(impossible): /tmp/hcp/HCP1200/103010/.xdlm/103010_3T_rfMRI_REST1_fixextended.json [file has no content available]
```

----------------------------------------------------------------------------------------------------------------------------------------
**Behavior of** ``--recursive``:

I was confused that recursive copy did not preserve the directory structure. 

```
datalad create -c yoda yodads
echo "my script" > yodads/code/script
datalad -C yodads save
datalad create noyodads      
datalad copyfile -r yodads/* noyodads

```
```
tree yodads noyodads
yodads
├── CHANGELOG.md
├── code
│   ├── README.md
│   └── script
└── README.md
noyodads
├── CHANGELOG.md
├── README.md
└── script
```

(I was also a bit confused that ``--recursive`` in the case of ``copyfile`` is about directories, when all other ``--recursive`` operations are about subdatasets, but on second thought, it matches the behavior of ``cp``, and its written in the help text)

-----------------------------------------------------------------------------------------------------------------------------------------

**Something I anticipate to be confusing, but can't see a way around**

If I have a dataset with configurations that prevent files from being annexed, copying these files into a new dataset will annex them. This is behavior makes sense as long someone knows about files in annex versus in Git, but for anyone using the command naively it may be confusing:

```
datalad create anotherds
datalad copyfile yodads/code/script anotherds
tree anotherds
anotherds
└── script -> .git/annex/objects/vz/z1/MD5E-s10--4af6de6a8afa5377c382b539d8479406/MD5E-s10--4af6de6a8afa5377c382b539d8479406
```

I don't think this can be easily prevented, but maybe its something to keep in mind.

mih:
Thx @kyleam @adswa that is very helpful. It is actually pointing to the necessity to turn thing a bit inside out internally. Will report once done.

yarikoptic:
minor comment: I would have made it `copy-file` (`copy_file` in Python) to be more inline with the rest of the datalad API separating action and object (e.g. `create-sibling`, `download-url`, etc)

mih:
> There might be good implementation or design reasons for not being able to specify a target with a different name

There weren't ;-)

Works now (857fd8d71dbd904c22953586d25047a9b27e61e1), added to tests, too.

mih:
@adswa 
> If specifying different target names is not possible, maybe warn/report better in accidental attempts of copying identically named files?

I went another way. `cp` would overwrite the original file, and now this command does that too.  In combination with the rectified behavior of being able to perform renaming copies, I think this addresses this aspect.

mih:
> minor comment: I would have made it `copy-file` (`copy_file` in Python) to be more inline with the rest of the datalad API separating action and object (e.g. `create-sibling`, `download-url`, etc)

@yarikoptic  This is implemented now.

mih:
> By chance, the very first thing I did was to copy an annexed file (with retrieved content) into a dataset with no annex (a few HCP subjects files into the top-level `human-connectome-project-openacess` dataset). Afterwards I end up with a broken symlink (understandable)

This is not intentional. It should have copied the content, not the symlink. I will investigate.

> Would it be maybe good to warn when it is detected that annexed contents are copied into datasets without an annex? I'm thinking of someone copying a dataset with large data into a plain Git repository (e.g, maybe one for their code) and being not able to retrieve any data afterwards.

I have not thought about it this way so far, but I think that is a fair point: If a file was annexed in its original location, it probably was for a reason. Hence might might make sense to fail....

This might hint at the necessity to introduce a `--force` argument with some modes, one of them being to commit originally annexed files into git at the destination.

> Also: When I `copyfile` an annexed file I didn't `get` yet into a non-annex dataset it fails, but the warning is misleading. To me, the warning would suggest that I would need to get content first, which is futile in this case - even if I did, copying it into a dataset without an annex will turn it into a broken symlink.

The warning in this case says what I intended, the file should get annexed.

mih:
> Behavior of --recursive

This is fixed now, and the prev behavior was unintentional:

```
% tree yodads noyodads
yodads
├── CHANGELOG.md
├── code
│   ├── README.md
│   └── script
└── README.md
noyodads
├── CHANGELOG.md
├── code
│   ├── README.md
│   └── script
└── README.md
```

mih:
> Something I anticipate to be confusing, but can't see a way around

> If I have a dataset with configurations that prevent files from being annexed, copying these files into a new dataset will annex them. This is behavior makes sense as long someone knows about files in annex versus in Git, but for anyone using the command naively it may be confusing

I have to investigate that. I would really really like to avoid having to anticipate/inspect what *would* happen on `save`, but maybe there is a way to catch common scenarios and work them int o a `force` mode.

mih:
@adswa I think I have found and fixed all immediate bugs you discovered. There is no `--force` yet, and the warning strategy stays the same for now. But If you could have another look, and see if the present behavior is an improvment, I'd be glad. Thx!

kyleam:
I pushed some commits, most of them related to the failing test.  Please revert/rework if any of them if needed.

mih:
@kyleam  wonderful, thx!

mih:
I hope the last push will make the cripples FS tests pass. The only thing missing from my POV is a docstring.

mih:
I can tune the history, if desired. I would like to maintain the contributions by individual people, though.

datalad/local/copy_file.py
```diff
@@ -0,0 +1,666 @@
...
+            text="Copy files using a path and optionally target specification "
+                 "from a file",
+            code_py="""\
+            copy_file(dataset='path/to/myds', target_dir='path/to/myds',
+                      specs_from='path/to/specfile')""",
+            code_cmd="""\
+            datalad copy-file -d path/to/myds -t path/to/myds --specs-from specfile"""
+        ),
+        dict(
+            text="Read a specification from stdin: Pipe the output of a find "
```
kyleam, line 155: This colon reads oddly to me.  Perhaps just s/: Pipe/and pipe/

datalad/local/copy_file.py
```diff
@@ -0,0 +1,666 @@
...
+                 "command into the copy-file command",
+            code_py="",
+            code_cmd="""\
+            find <expr> | datalad copy-file -d myds -t path/to/myds --specs-from -"""
```
kyleam, line 159: Hmm, shouldn't the dataset path match between `-d` and `-t`?

adswa: oh true, thanks for catching that!

datalad/local/copy_file.py
adswa: done, thx

datalad/local/copy_file.py
```diff
@@ -0,0 +1,666 @@
...
+            code_py="""\
+            copy_file(dataset='path/to/myds', target_dir='path/to/myds',
+                      specs_from='path/to/specfile')""",
+            code_cmd="""\
+            datalad copy-file -d path/to/myds -t path/to/myds --specs-from specfile"""
+        ),
+        dict(
+            text="Read a specification from stdin: Pipe the output of a find "
+                 "command into the copy-file command",
+            code_py="",
```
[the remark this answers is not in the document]
adswa: uh, yes! how could I miss this... thanks!

datalad/local/copy_file.py
```diff
@@ -0,0 +1,665 @@
...
+    content identity and availability.
+
+    || REFLOW >>
+    Interface and behavior are modeled after the POSIX 'cp' command, with two
+    main differences. 1) Via [CMD: --specs-from CMD][PY: `specs_from` PY] it is
+    possible to flexibly input source-destination path pairs. 2) Saving changes
+    in a target dataset is conditional upon a specification of
+    [CMD: --dataset CMD][PY: `dataset`, or calling this command as a `Dataset`
+    method PY].
+    << REFLOW ||
```
kyleam, line 67: The added docstring overall looks good to me, but I don't like this section.  I think "interface *and behavior*" is a bit too broad for the reader to know what you have in mind when you're listing these differences, and how they, er, differ from the differences the last paragraph just mentioned.  Perhaps something like

    The interface is modeled after the POSIX 'cp' command, but with
    one additional way to specify what to copy where: [CMD:
    --specs-from CMD][PY: `specs_from` PY] allows the caller to
    flexibly input source-destination path pairs.

That mostly just drops the `--dataset` bit.  I think it was included because it's the other copy-file argument that cp doesn't have.  But that's obvious enough that I don't think it's worth an explicit mention, especially because doing so would mean that when `copy-file` gains a new option, this part would need to be updated.

datalad/local/copy_file.py
```diff
@@ -0,0 +1,665 @@
...
+
+            if not specs_from:
+                # in all other cases we have a plain source list
+                specs_from = paths
+
+        if not specs_from:
+            raise ValueError("Neither `paths` nor `specs_from` given.")
+
+        res_kwargs = dict(
+            action='copy',
```
kyleam, line 227: I think this should be "copy_file" to (1) be consistent with the module name and (2) to avoid a conflict with `push`, which already uses a "copy" action for something else.

mih: Done.

datalad/local/copy_file.py
mih: I can follow your logic. Applied!

kyleam:
I was hoping to see more people chime in on the interface.

From my perspective, it looks nice and the simple things I tried behaved as I expected.  The one thing I notice from the examples is that common use cases would require repeating the dataset path as both the target directory and dataset.  There is a comment asking whether the target directory, if not given, should be set to the dataset.  That'd remove the need to repeat in many cases, but then I guess there's an interaction between options that makes the interface conceptually more complicated.  @mih, is that your main concern with making the target directory default to --dataset?

datalad/local/tests/test_copy_file.py
```diff
@@ -339,4 +339,4 @@ def test_copy_file_prevent_dotgit_placement(srcpath, destpath):
             [sub.pathobj / '.git' / 'config',
              dest.pathobj / 'some', '.git'], on_failure='ignore'),
         status='impossible',
-        action='copy')
+        action='copy_file')
```
kyleam, line 342: Just a heads up: when I tried to test this simple update locally, this test errors for me with

```
ERROR: datalad.local.tests.test_copy_file.test_copy_file_prevent_dotgit_placement
  File "/home/kyle/src/python/venvs/datalad/lib/python3.7/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/kyle/src/python/datalad/datalad/tests/utils.py", line 691, in newfunc
  File "/home/kyle/src/python/datalad/datalad/tests/utils.py", line 691, in newfunc
  File "/home/kyle/src/python/datalad/datalad/local/tests/test_copy_file.py", line 340, in test_copy_file_prevent_dotgit_placement
    dest.pathobj / 'some', '.git'], on_failure='ignore'),
  File "/home/kyle/src/python/datalad/datalad/distribution/dataset.py", line 499, in apply_func
    return f(**kwargs)
  File "/home/kyle/src/python/datalad/datalad/interface/utils.py", line 494, in eval_func
  File "/home/kyle/src/python/datalad/datalad/interface/utils.py", line 482, in return_func
    results = list(results)
  File "/home/kyle/src/python/datalad/datalad/interface/utils.py", line 413, in generator_func
    allkwargs):
  File "/home/kyle/src/python/datalad/datalad/interface/utils.py", line 552, in _process_results
    for res in results:
  File "/home/kyle/src/python/datalad/datalad/local/copy_file.py", line 286, in __call__
    for res in _copy_file(src_file, dest_file, cache=repo_cache):
  File "/home/kyle/src/python/datalad/datalad/local/copy_file.py", line 470, in _copy_file
    eval_file_type=True,
KeyError: 'dictionary is empty'
```

My first thought is that it's due to the (newer) git-annex version that I have locally, but I haven't yet looked into it.

mih:
FTR: @adswa has built a "structural images only" dataset from the full HCP dataset collection like so:

```
datalad run -m "..." --output '[0-9]*' --explicit \
  "find .hcp/HCP1200 -maxdepth 3 -name 'T*w_acpc_dc.nii.gz' \
    -o -name 'T*w_acpc_dc_restore.nii.gz' -o -name 'T1wDividedByT2w.nii.gz' \
    -type l | sed -e 's#\(.hcp/HCP1200\)\(.*\)\$#\1\2\x00.\2#' \
    | datalad copy-file -r --specs-from -"
```
Running in a fresh dataset that had the HCP superdataset mounted under `.hcp/`. The whole thing took ~90min for `5548 annex'd files (364.8 GB recorded total size)`. As expected, this isn't fast, but it is also not intolerably slow for a one-time operation.

Few observations:

- `--explicit` was necessary to make things practical. We used a symlink to point to the actual location of the HCP dataset hierarchy, as we did not want to invest into hours of cloning.
- specifying `--input` would have been nice, but wasn't possible, because `run` would have started downloading file content, which isn't needed for this operation.

Other than that, it led to a decent run record, and the desired result: a single 9.2M dataset that can be installed in seconds and captures everything that is necessary for a certain range of analyses, and in itself captures which HCP version it was built from, without requiring any significant duplication of storage resources.

We will now repeat this for different subsamples, with varying complexity.

mih:
> The one thing I notice from the examples is that common use cases would require repeating the dataset path as both the target directory and dataset. There is a comment asking whether the target directory, if not given, should be set to the dataset. That'd remove the need to repeat in many cases, but then I guess there's an interaction between options that makes the interface conceptually more complicated. @mih, is that your main concern with making the target directory default to --dataset?

I share this view. I merely wasn't done thinking yet, hence the comment. ATM I am leaning towards setting `target_dir` to `ds.pathobj` if no specific `target_dir` was given. We definitely need the ability to express "no save" (I posted a concrete example for a combination with `datalad run` in https://github.com/datalad/datalad/pull/4430#issuecomment-620382755, where this is instrumental. However, it should not prevent us from deriving `target_dir` from `dataset` (if the latter is given, but the former is not).

mih:
The current implementation leads to a large number of commit in the git-annex branch. 11k in the case of the HCP subset mentioned above. I need to find a way to squash these.

mih:
This is good to go from my POV. Actual usage patterns will show where to improve in the future.

kyleam:
> The current implementation leads to a large number of commit in the git-annex branch

Can these be avoided in the first place by `copy_file` setting `always_commit=False`?

kyleam:
> This is good to go from my POV.

Mine too.  @adswa, is the interface/behavior in a place you're happy with?

datalad/local/tests/test_copy_file.py
```diff
@@ -0,0 +1,339 @@
...
+# ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+#
+#   See COPYING file distributed along with the datalad package for the
+#   copyright and license terms.
+#
+# ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+"""Test copy_file command"""
+
+
+from os.path import (
```
yarikoptic, line 11: BTW, in many places we already moved to use `from datalad.support import path as op` which is our "centralized" shims for `os.path` and then `op.join` which isn't too long ;)  ATM it seems there is still lots of `opj` (~700) but `op.join` is catching up (~500).  Such centralization of imports / shims could allow to tune / fix/ workaround in case of need.

datalad/local/copy_file.py
```diff
@@ -0,0 +1,672 @@
...
+                repo = repo_rec['repo']
+                if not repo or repo.pathobj in done:
+                    continue
+                tmp = repo_rec.get('tmp', None)
+                if tmp:
+                    try:
+                        tmp.rmdir()
+                    except OSError as e:
+                        lgr.warning(
+                            'Failed to clean up temporary directory: %s', e)
```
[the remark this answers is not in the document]
kyleam: Based on this [comment][0], I think @mih is intentionally avoiding it.  You guys might want to hash that out to avoid inconsistency and unnecessary churn in the code base.

[0]: https://github.com/datalad/datalad/pull/4203#discussion_r410191253

yarikoptic:
> > The current implementation leads to a large number of commit in the git-annex branch
> 
> Can these be avoided in the first place by `copy_file` setting `always_commit=False`?

oh, I misunderstood @mih's comment on the call about "manual squash" of those -- I thought it was "manual" (in that the direct call to `git`) but automated... FWIW: 
- in my little trial `git annex forget` attempt seemed to do the squashing
- I left a [TODO comment](https://github.com/datalad/datalad/pull/4430/files#diff-ae21f3c66c9d2695154bfa09844bbda5R573) that ideally we should interface `registerurl` at AnnexRepo level and add `batch` option which it supports.  IIRC that should also lead to smaller number of commits to git-annex branch

datalad/local/copy_file.py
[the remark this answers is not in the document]
yarikoptic: thanks, left a comment there.  Here it is even higher level than debug -- a `warning` -- so I am ok to spend those few `ms` to possibly provide more diagnostic information if so desired.  I will also later today propose a PR to make it a "non-issue" and probably to demonstrate that it indeed was not any performance issue to start with.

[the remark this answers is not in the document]
yarikoptic: reread the code -- I was wrong in that I assumed that if no limit explicitly provided we would not include any traceback location information thus we could simply avoid it altogether if DATALAD_EXC_STR_TBLIMIT variable is not set.  But we are defaulting to limit of 1 (thus showing the file:lineno where exception was raised).   I think it is valuable even if in many cases we could match location where it was thrown by `git grep`ing the message (until it starts coming from python standard libraries etc).  So I will keep it as is and shortly push a docstring change (with no ci) to clarify the parameters etc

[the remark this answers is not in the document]
yarikoptic: did in did in 11932b3f4ae60fb4ee7798731a669d6e1598481e

adswa:
> Mine too. @adswa, is the interface/behavior in a place you're happy with?

Its much harder to judge whether the interface feels weird if I'm more involved ;-) But yes, I am. Especially now that the duplication of the target dir/ds is removed it feels much more intuitive to use than when I initially played around with it.

From the past few days when I have been using the command productively there is only one thing that slowed me down/I found confusing - but I can see why it is, and hope that the handbook can eliminate confusion/be a good demonstration of how it can be done: When reading sources and destinations from stdin with ``--specs-from``, the source and target path need to be separated by a Null byte. I had never heard of a null byte before, I spent a few hours wrangling `sed` to give me one, and its not straightforward to see if one succeeded producing one.

datalad/local/tests/test_copy_file.py
mih: I hope that in most places we have moved away from using either model, and use `from datalad.utils import Path`.

mih:
I have mixed feelings about the contributions to this PR. Some are great. Thx for those!

But some are style changes. I don't think that we have a style-guide, and I would appreciate tolerance. It pains me that some people use editors that remove the final newline of a file, but I can live with that -- please be tolerant to my kinks too.

Some commits mix aspects, e.g. subjective reformatting plus a necessary change.

I would like to revert 48adce4909282084131917ccda37ba85eda826c7 -- it is unclear why that change had to be made. Moreover the new implementation is objectively less robust.

re (some part of) 780aefe650e71287a4df5a9575c7f4cbd0fdb9c1: I intentionally did not use this method, precisely because it is a warning. The normal str(e) gives a perfectly fine output and pinpoints the problem, the new one is needlessly wordy.

Sidenote, checked if I can turn off the wordyness and still use the function, but the docstring of `exc_str()` is ;-)

```
    """Enhanced str for exceptions.  Should include original location

    Parameters
    ----------
    Exception to
    """
 ```

I find 929102aa3a7b894959259284b614a0cf6a01af68 misleading (despite the message saying that it reduces confusion). The function doesn't necessarily copy "with_annex_info" -- that depends on circumstances. Moreover, there is no other copy that it could be confused with.

I think 8e5909800f5f0bed80407865c8dd7e20bed11f19 is a style matter that I don't like. This code will see massive RF -- if the rest of datalad is any indication. This type of code condensing is premature and makes RF harder. I have taken so many of these apart. This code is not optimized to any degree, optimizing the error messaging is too early IMHO.

File: datalad/local/copy_file.py [hunk 9]
```diff
@@ -0,0 +1,672 @@
...
+        # things up
+        repo_cache = {}
+        # which paths to pass on to save
+        to_save = []
+        try:
+            for src_path, dest_path in _yield_specs(specs_from):
+                src_path = Path(src_path)
+                dest_path = None \
+                if dest_path is None \
+                else resolve_path(dest_path, dataset)
```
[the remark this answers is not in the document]
mih: I do that frequently, I don't find it confusing.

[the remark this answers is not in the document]
yarikoptic: Frequency of you doing it does not make it less confusing to others and PEP8:
```
$> flake8 /tmp/1.py
/tmp/1.py:2:1: E122 continuation line missing indentation or outdented
/tmp/1.py:3:1: E122 continuation line missing indentation or outdented

$> cat /tmp/1.py
bu = 1 \
if True \
else 2
```

datalad/local/tests/test_copy_file.py
yarikoptic: yes, I should use it more often; but comment was for non `pathlib` ways, which is what was used here.

yarikoptic:
We could discuss more in the next hangout - meanwhile feel free to revert anything you don't like, I will not fight for it.

Briefly my rationale: In many places would start wasting more time trying to parse that code again if I ever get to it. As it is more than one person to care about that code in the longer run, I decided to contribute.

Re styling, you can see https://github.com/datalad/datalad/blob/master/CONTRIBUTING.md#linting summarized as "[PEP8 is a] guidelines for your contributions".  I found some code hard to parse due to excessive verbosity/duplication, non-standard indentation, comparison to some long strings which I hope have correct number of 0s in all of their manifestations.

**NB** Now I remembered that we have AnnexRepo.WEB_UUID... submitted RF #4460 .

Re warning: might not be informative to the user, but I found it very useful on many occasions to 1. see that there is a stack in principle, raise the variable `DATALAD_EXC_STR_TBLIMIT`, 2. rerun and see how did I actually get to that particular exception came from (otherwise that warning could come from anywhere).  The majority of log lines showing str of the exception use `exc_str` so I thought it would be valuable even if only for consistency.

Once again, the summary -- tune to your liking, and I will try to refrain from further direct changes to this PR.

mih:
I decided to rebase on master in order to strip

- 929102aa3a7b894959259284b614a0cf6a01af68 (function rename)
- 48adce4909282084131917ccda37ba85eda826c7 (approximate special remote filtering)

rather than adding explicit commits that revert these.

I left the `exc_str()` call in. I still very much dislike the messing setup presently used by DataLad. It is targeting developers, optimizing for debugging. It confuses users with wordy, hard to parse, duplicate messages. But this single message won't make a dent.  I filed a general issue re `exc_str()`. https://github.com/datalad/datalad/issues/4478

mih:
I force-pushed the rebase and added a final documentation tuning. I will merge, once the tests are completed.

PR #4182, opened 2020-02-21
Failed attempt to parallelize subdataset installation in `get()`

mih:
This sits on top of #4172 and only the last commit is the one that puts joblib into the picture. It should work as committed, but fail once the number of jobs is set to `>1`. Maybe someone can spot the issue if it is recoverable. 

The end of the traceback is

```
  File "/home/mih/hacking/datalad/git/datalad/distribution/get.py", line 448, in _recursive_install_subds_underneath
    for sub, url_info, urls in to_install
  File "/home/mih/env/datalad3-dev/lib/python3.7/site-packages/joblib/parallel.py", line 1017, in __call__
    self.retrieve()
  File "/home/mih/env/datalad3-dev/lib/python3.7/site-packages/joblib/parallel.py", line 909, in retrieve
    self._output.extend(job.get(timeout=self.timeout))
  File "/home/mih/env/datalad3-dev/lib/python3.7/site-packages/joblib/_parallel_backends.py", line 562, in wrap_future_result
    return future.result(timeout=timeout)
  File "/usr/lib/python3.7/concurrent/futures/_base.py", line 435, in result
    return self.__get_result()
  File "/usr/lib/python3.7/concurrent/futures/_base.py", line 384, in __get_result
    raise self._exception
joblib.externals.loky.process_executor.BrokenProcessPool: A task has failed to un-serialize. Please ensure that the arguments of the function are all picklable.
```

PR #4434, opened 2020-04-21
NF(TEMP): reincarnated #4182 with parallel via subprocess

yarikoptic:
This is #4182 with one more commit (6a6425f38e5ab6d70caa46c5571858c5ed83c142):

    TEMP: use ProcessPoolExecutor to parallelize get
    
    I have tried with purely uncommented code -- no core dump, but it seemed
    to do no parallelization whatsoever.  So decided to switch to processpool.
    Had to avoid returning a generator (thus ugly TEMP clone_dataset_list).  But
    overall -- seemed to work on a simple invocation
    
       datalad get -n -r -d ~/datalad/crcns/
    
    and I saw no core dumps

### DIFF
--- a/datalad/core/distributed/clone.py
+++ b/datalad/core/distributed/clone.py
@@ -324,6 +324,9 @@ def clone_dataset(
     dict
       DataLad result records
     """
+    if not isinstance(destds, Dataset):
+        destds = Dataset(destds)
+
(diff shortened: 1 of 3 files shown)


yarikoptic:
didn't bring up interest, and could be found (hopefully) if/when needed

PR #4439, opened 2020-04-22
ENH+RF: use repr, not quote_cmdlinearg for repr, all *Repos need a single __repr__ implementation

yarikoptic:
Continuing #4435 et al (since merged already)

With quote_cmdlinearg we would get inconsistent reprs -- quotes will be used
only if needed, thus it would be less of a true repr and imho it is suboptimal
to have such a jumpy appearance. e.g. in master we would have

	$> python -c 'from datalad.api import Dataset; ds = Dataset("/tmp/v e\" ры"); print(ds, ds.repo)'
	Dataset('/tmp/v e" ры') AnnexRepo('/tmp/v e" ры')
	$> python -c 'from datalad.api import Dataset; ds = Dataset("///"); print(ds, ds.repo)'
	Dataset(/home/yoh/datalad) AnnexRepo(/home/yoh/datalad)

with the changes here it would be more consistent imho:

	$> python -c 'from datalad.api import Dataset; ds = Dataset("/tmp/v e\" ры"); print(ds, ds.repo)'
	Dataset('/tmp/v e" ры') AnnexRepo('/tmp/v e" ры')
	$> python -c 'from datalad.api import Dataset; ds = Dataset("///"); print(ds, ds.repo)'
	Dataset('/home/yoh/datalad') AnnexRepo('/home/yoh/datalad')

Also it is really not needed to duplicate code in __repr__s of *Repos.
Better imho to rely on inheritance and knowing the class name.

### DIFF
--- a/datalad/distribution/dataset.py
+++ b/datalad/distribution/dataset.py
@@ -149,7 +149,7 @@ class Dataset(object, metaclass=PathBasedFlyweight):
         return self._pathobj
 
     def __repr__(self):
-        return 'Dataset({})'.format(quote_cmdlinearg(self.path))
+        return 'Dataset({!r})'.format(self.path)
 
     def __eq__(self, other):
         if not hasattr(other, 'pathobj'):
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -974,9 +974,6 @@ class AnnexRepo(GitRepo, RepoInterface):
             srs[sr_id] = sr_info
         return srs
 
-    def __repr__(self):
-        return 'AnnexRepo({})'.format(quote_cmdlinearg(self.path))
-
     def _run_annex_command(self, annex_cmd,
                            git_options=None, annex_options=None,
                            backend=None, jobs=None,
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -1145,7 +1145,7 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
         self._cfg = None
 
     def __repr__(self):
-        return 'GitRepo({})'.format(quote_cmdlinearg(self.path))
+        return '{}({!r})'.format(self.__class__.__name__, self.path)
 
     def __eq__(self, obj):
         """Decides whether or not two instances of this class are equal.


mih:
Can we then maybe use `__str__` to get something that works in the cmdline for C&P?

yarikoptic:
> Can we then maybe use `__str__` to get something that works in the cmdline for C&P? The difficult case that I found was this
> 
> [#4420 (comment)](https://github.com/datalad/datalad/pull/4420#issuecomment-615591243)
> 
> Maybe this solution does it better.

why repr wouldn't be good enough for cmdline?

```shell
$> python -c 'from datalad.api import Dataset; from glob import glob; ds = Dataset(glob("/tmp/that*/dataset")[0]); print(ds, ds.repo)'
Dataset("/tmp/that's it folks/dataset") AnnexRepo("/tmp/that's it folks/dataset")
$> ls -l "/tmp/that's it folks/dataset"
total 0
```
ok - may be in completely obnoxious cases it wouldn't :
```
$> datalad create "/tmp/that's it \"folks/dataset"
[INFO   ] Creating a new annex repo at /tmp/that's it "folks/dataset 
[INFO   ] Scanning for unlocked files (this may take some time) 
create(ok): /tmp/that's it "folks/dataset (dataset)

$> python -c 'from datalad.api import Dataset; from glob import glob; ds = Dataset(glob("/tmp/that*/dataset")[0]); print(ds, ds.repo)'
Dataset('/tmp/that\'s it "folks/dataset') AnnexRepo('/tmp/that\'s it "folks/dataset')

$> ls -l '/tmp/that\'s it "folks/dataset'
dquote> 
```
but this is imho probably <0.01% of the cases (at least I had none)

kyleam:
> but this is imho probably <0.01% of the cases (at least I had none)

While that is likely be true for regular use, in the context of debugging/pasting from tests, we have datasets that have our obscure file name in them.  Wanting to be able to paste those in my terminal during debugging is my selfish reason for preferring the command-line bias of gh-4420 and gh-4435.  For me, that outweighs any concern for the "jumping"/inconsistency that you mention.

yarikoptic:
> While that is likely be true for regular use, in the context of debugging/pasting from tests, we have datasets that have our obscure file name in them. Wanting to be able to paste those in my terminal during debugging is my selfish reason for preferring the command-line bias of [gh-4420](https://github.com/datalad/datalad/pull/4420) and [gh-4435](https://github.com/datalad/datalad/pull/4435). For me, that outweighs any concern for the "jumping"/inconsistency that you mention.

Well, if debugging happens in Python, so pasting into python shell/code, you would like to have `repr` representation here for copy/pasteability.  So probably there is no solution which would work for both Python and shell modes all the time anyways. In terminal you could probably complete by the beginning you paste in case of the most obscure ones.

kyleam:
> Well, if debugging happens in Python [...]

If it weren't clear from my comment, I was very much talking about _my_ reason for preferring the command-line biased output.  In my workflow, I copy and paste paths from the debugger _to_ my terminal frequently.

mih:
Given the disagreement (I also prefer the "jumpy" behavior), what about my earlier question to use `__str__`?

yarikoptic:
> Given the disagreement (I also prefer the "jumpy" behavior), what about my earlier question to use `__str__`?

do you mean just to use `{}` instead of `{!r}` -- then it will be usable neither in bash nor in Python (thus wouldn't be any string `repr` of an object):
```
$> python -c 'from datalad.api import Dataset; from glob import glob; ds = Dataset(glob("/tmp/that*/dataset")[0]); print(ds, ds.repo)'
Dataset(/tmp/that's it "folks/dataset) AnnexRepo(/tmp/that's it "folks/dataset)
```

mih:
No, I meant to define a `__str__()` in all classes that need not aim to provide a true `repr` and could still perform as the prev PR desired.

yarikoptic:
Not sure on utility of such distinction, but FWIW implemented it, I also RFed more to avoid any code duplication across classes, since the last commit ;) so would look like
```shell
$> python -c 'from datalad.api import Dataset; from glob import glob; ds = Dataset(glob("/tmp/*\"*/dataset")[0]); print(ds, repr(ds)); print(ds.repo, repr(ds.repo))'
Dataset('/tmp/that'"'"'s it "folks/dataset') Dataset('/tmp/that\'s it "folks/dataset')
AnnexRepo('/tmp/that'"'"'s it "folks/dataset') AnnexRepo('/tmp/that\'s it "folks/dataset')

$> python -c 'from datalad.api import Dataset; from glob import glob; ds = Dataset(glob("/tmp/testds")[0]); print(ds, repr(ds)); print(ds.repo, repr(ds.repo))'     
Dataset(/tmp/testds) Dataset('/tmp/testds')
AnnexRepo(/tmp/testds) AnnexRepo('/tmp/testds')
```
especially lovely it would be for us to figure out while looking at some message if it is a repr (thus - do not bother pasting into terminal) or str.

datalad/support/repo.py
```diff
@@ -249,3 +249,32 @@ class RepoInterface(object):
 
     # Test!
     pass
+
+
+def path_based_str_repo(cls):
+    """A helper decorator for a class to define str and repo based on its .path
```
kyleam, line 255: s/repo/repr/ in several places.

Other minor issues I noticed with the text on a quick scan: first sentence does not parse, "cut" is probably a confusing word to use, double space between "idea" and "is".

datalad/support/repo.py
```diff
@@ -249,3 +249,32 @@ class RepoInterface(object):
...
+
+    For the rationale/discussion on why to bother distinguishing the two is
+    in https://github.com/datalad/datalad/pull/4439 . The idea  is that
+    `__str__` should provide cut/pasteable to shell representation of the path,
+    with all necessary escapes for characters shell might care about.
+    `__repr__` lives to its promise to provide string representation consumable
```
kyleam, line 261: Hrm, probably not worth changing, but I think "promise" is a bit strong, based on my understanding of `repr` conventions and anything I see in the [official docs](https://docs.python.org/3/reference/datamodel.html?highlight=iadd#object.__repr__).

datalad/support/repo.py
yarikoptic: Cool, thanks for picking those up!... I will tune up those if we decide that is a way to go (or welcome to push fixup to make it to your liking)

yarikoptic:
BTW, can `.path` change through the life time of the instance of any of those?  if not -- we could memoize it - thus possibly making this frequently invoked functions faster.  it will not really solve the question why datalad is slow, but hey -- not wasted cycles as @mih said ;)

yarikoptic:
FWIW, pushing a commit which would add memoization (without checking if .path changed) -- saves about 80% of time on subsequent invocations ;) but has no notable impact on any of the benchmarks

datalad/support/repo.py
yarikoptic: yeah, could have just removed "lives to its promise" I guess. would do if further changes to happen.

yarikoptic: oh spotted that I haven't addressed comment above, so will push `no ci` commit with those now

datalad/support/repo.py
```diff
@@ -249,3 +249,41 @@ class RepoInterface(object):
 
     # Test!
     pass
+
+
+def path_based_str_repo(cls):
```
kyleam, line 254: There is still a repo->repr typo here.

kyleam: Pushed  a fixup.

<details>
<summary>range-diff</summary>

```
1:  1f84216366 ! 1:  0c61b58457 RF: @path_based_str_repo defining __str__ and __repr__
    @@ Metadata
     
      ## Commit message ##
    -    RF: @path_based_str_repo defining __str__ and __repr__
    +    RF: @path_based_str_repr defining __str__ and __repr__
    +
    +    Modified-by: Kyle Meyer <kyle@kyleam.com>
    +      Fix repo->repr typo.
     
      ## datalad/distribution/dataset.py ##
     @@
      from datalad.support.exceptions import (
          NoDatasetFound,
      )
    -+from datalad.support.repo import path_based_str_repo
    ++from datalad.support.repo import path_based_str_repr
      from datalad.support.gitrepo import (
          GitRepo,
      )
    @@ datalad/distribution/dataset.py
      lgr.log(5, "Importing dataset")
      
      
    -+@path_based_str_repo
    ++@path_based_str_repr
      class Dataset(object, metaclass=PathBasedFlyweight):
          """Representation of a DataLad dataset/repository
      
    @@ datalad/support/gitrepo.py
          PathBasedFlyweight,
     -    RepoInterface
     +    RepoInterface,
    -+    path_based_str_repo,
    ++    path_based_str_repr,
      )
      
      # shortcuts
    @@ datalad/support/gitrepo.py: def _from_line(cls, line):
      Submodule = namedtuple("Submodule", ["name", "path", "url"])
      
      
    -+@path_based_str_repo
    ++@path_based_str_repr
      class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
          """Representation of a git repository
      
    @@ datalad/support/repo.py: class RepoInterface(object):
          pass
     +
     +
    -+def path_based_str_repo(cls):
    ++def path_based_str_repr(cls):
     +    """A helper decorator for a class to define str and repr based on its .path
     +
     +    For the rationale/discussion on why to bother distinguishing the two is
2:  60a5f94ff7 ! 2:  911da5c728 OPT: memoize __str__ and __repr__ of @path_based_str_repo"ed classes
    @@ Metadata
     
      ## Commit message ##
    -    OPT: memoize __str__ and __repr__ of @path_based_str_repo"ed classes
    +    OPT: memoize __str__ and __repr__ of @path_based_str_repr"ed classes
     
         no check is done that .path was not changed -- so assumption is that it does
         not change.  Shaves off about 80% time on subsequent invocations
     
      ## datalad/support/repo.py ##
    -@@ datalad/support/repo.py: def path_based_str_repo(cls):
    +@@ datalad/support/repo.py: def path_based_str_repr(cls):
          # 62 ns ± 0.345 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
          # and similarly 58ns vs 97ns for %r vs !r
          def __str__(self):
```

</details>

kyleam:
I feel okay leaving the CI run to the merge given that the fixup was a find/replace and that light testing locally didn't reveal any issues.

datalad/support/repo.py
yarikoptic: heh heh -- I just fixed in the docstring and missed that function itself had it. Thank you for fixing it!

PR #4441, opened 2020-04-23
ENH: ExternalVersions.add - for extensions etc to add custom checkers

yarikoptic:
Use case: datalad-neuroimaging started to fail tests.  Preliminary
check seems to point toward newer dcm2niix.

It would be nice if extensions, where matters, could check for their
externals and possibly even register them within our "central"
ExternalVersions checker.  This is what `add` effectively does.  Also
for regular python modules, it would pretty much just allow to add
them to "INTERESTING" list - displayed by dumps.

I have positioned it against maint for now since it would allow extensions to use this feature sooner than later, and it is largely a minor enhancement which should not have negative side-effects (I hope ;)).  But if there are objections, I am ok to change it to go to master.

### DIFF
--- a/datalad/plugin/wtf.py
+++ b/datalad/plugin/wtf.py
@@ -14,6 +14,7 @@ import logging
 import os
 import os.path as op
 from functools import partial
+from itertools import chain
 from collections import OrderedDict
 
 
@@ -221,7 +222,7 @@ def _describe_metadata_extractors():
 def _describe_dependencies():
     # query all it got
     [external_versions[k]
-     for k in tuple(external_versions.CUSTOM) + external_versions.INTERESTING]
+     for k in chain(external_versions.CUSTOM, external_versions.INTERESTING)]
 
     return {
         k: str(external_versions[k]) for k in external_versions.keys()
--- a/datalad/plugin/wtf.py
+++ b/datalad/plugin/wtf.py
@@ -220,12 +220,8 @@ def _describe_metadata_extractors():
 
 
 def _describe_dependencies():
-    # query all it got
-    [external_versions[k]
-     for k in chain(external_versions.CUSTOM, external_versions.INTERESTING)]
-
     return {
-        k: str(external_versions[k]) for k in external_versions.keys()
+        k: str(external_versions[k]) for k in external_versions.keys(query=True)
     }
 
 
--- a/datalad/support/external_versions.py
+++ b/datalad/support/external_versions.py
@@ -13,6 +13,7 @@ import os.path as op
 from os import linesep
 
 from distutils.version import LooseVersion
+from itertools import chain
 
 from datalad.dochelpers import exc_str
 from datalad.log import lgr
@@ -133,7 +134,7 @@ class ExternalVersions(object):
 
     UNKNOWN = UnknownVersion()
 
-    CUSTOM = {
+    _CUSTOM = {
         'cmd:annex': _get_annex_version,
         'cmd:git': _get_git_version,
         'cmd:bundled-git': _get_bundled_git_version,
@@ -141,7 +142,7 @@ class ExternalVersions(object):
         'cmd:system-ssh': _get_system_ssh_version,
         'cmd:7z': _get_system_7z_version,
     }
-    INTERESTING = (
+    _INTERESTING = (
         'appdirs',
         'boto',
         'exifread',
@@ -162,6 +163,8 @@ class ExternalVersions(object):
 
     def __init__(self):
         self._versions = {}
+        self.CUSTOM = self._CUSTOM.copy()
+        self.INTERESTING = list(self._INTERESTING)  # make mutable for `add`
 
     @classmethod
     def _deduce_version(klass, value):
@@ -252,6 +255,44 @@ class ExternalVersions(object):
     def __contains__(self, item):
         return bool(self[item])
 
+    def add(self, name, func=None, interesting=True):
+        """Add a version checker
+
+        This method allows third-party libraries to define additional checks.
+        It will not add `name` if already exists.  If `name` exists and `func`
+        is different - it will override with a new `func`.
+
+        Parameters
+        ----------
+        name: str
+          Name of the check (usually a name of the Python module, or an
+          external command prefixed with "cmd:")
+        func: callable, optional
+          Function to be called to obtain version information. This should be
+          defined when checking the version of something that is not a Python
+          module or when this class's method for determining the version of a
+          Python module isn't sufficient.
+        interesting: bool, optional
+          Makes sense only for entries without func - whether it is considered
+          "interesting" and should be included in output of dumps
+           (with query=True).
+        """
+        if func:
+            func_existing = self.CUSTOM.get(name, None)
+            was_known = False
+            if func_existing and func_existing is not func:
+                lgr.debug(
+                    "Adding a new custom version checker %s for %s, "
+                    "old one: %s", func, name, func_existing)
+                was_known = name in self._versions
+            self.CUSTOM[name] = func
+            if was_known:
+                # pop and query it again right away to possibly replace with a new value
+                self._versions.pop(name)
+                _ = self[name]
+        if interesting and name not in self.INTERESTING:
+            self.INTERESTING.append(name)
+
     @property
     def versions(self):
         """Return dictionary (copy) of versions"""
@@ -272,7 +313,7 @@ class ExternalVersions(object):
           get those which weren't queried for yet
         """
         if query:
-            [self[k] for k in tuple(self.CUSTOM) + self.INTERESTING]
+            [self[k] for k in chain(self.CUSTOM, self.INTERESTING)]
         if indent and (indent is True):
             indent = ' '
         items = ["%s=%s" % (k, self._versions[k]) for k in sorted(self._versions)]
--- a/datalad/support/external_versions.py
+++ b/datalad/support/external_versions.py
@@ -248,8 +248,17 @@ class ExternalVersions(object):
 
         return self._versions.get(modname, self.UNKNOWN)
 
-    def keys(self):
-        """Return names of the known modules"""
+    def keys(self, query=False):
+        """Return names of the known modules
+
+        Parameters
+        ----------
+        query: bool, optional
+          If True, we will first query all CUSTOM and INTERESTING entries
+          to make sure we have them known.
+        """
+        if query:
+            [self[k] for k in chain(self.CUSTOM, self.INTERESTING)]
         return self._versions.keys()
 
     def __contains__(self, item):
@@ -309,11 +318,9 @@ class ExternalVersions(object):
           To query for versions of all "registered" custom externals, so to
           get those which weren't queried for yet
         """
-        if query:
-            [self[k] for k in chain(self.CUSTOM, self.INTERESTING)]
         if indent and (indent is True):
             indent = ' '
-        items = ["%s=%s" % (k, self._versions[k]) for k in sorted(self._versions)]
+        items = ["%s=%s" % (k, self._versions[k]) for k in sorted(self.keys(query=query))]
         out = "%s" % preamble if preamble else ''
         if indent is not None:
             if preamble:
--- a/datalad/support/external_versions.py
+++ b/datalad/support/external_versions.py
@@ -255,12 +255,13 @@ class ExternalVersions(object):
     def __contains__(self, item):
         return bool(self[item])
 
-    def add(self, name, func=None, interesting=True):
+    def add(self, name, func=None):
         """Add a version checker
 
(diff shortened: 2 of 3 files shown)


datalad/plugin/wtf.py
```diff
@@ -219,7 +220,7 @@ def _describe_metadata_extractors():
 def _describe_dependencies():
     # query all it got
     [external_versions[k]
-     for k in tuple(external_versions.CUSTOM) + external_versions.INTERESTING]
+     for k in chain(external_versions.CUSTOM, external_versions.INTERESTING)]
```
kyleam, line 223: So this and the other `chain` call are needed because INTERESTING is now a list.  An alternative would have been `s/tuple/list/` in both spots.  Okay.

datalad/support/external_versions.py
```diff
@@ -252,6 +255,43 @@ def keys(self):
     def __contains__(self, item):
         return bool(self[item])
 
+    def add(self, name, func=None, interesting=True):
+        """Add a version checker
+
+        Could be used e.g. by extensions to add important to them custom checks.
```
kyleam, line 261: This took me a few attempts to parse, particularly the "to them" part.  Perhaps "This method allows third-party libraries to define additional checks"?

datalad/support/external_versions.py
```diff
@@ -252,6 +255,43 @@ def keys(self):
...
+        If will not add `name` if already exists.  If it exists and `func` is
```
kyleam, line 262: s/If will/It will/

s/If it/If `name`/ for clarity

datalad/support/external_versions.py
```diff
@@ -252,6 +255,43 @@ def keys(self):
...
+
+        Parameters
+        ----------
+        name: str
+          Name of the check (usually a name of the Python module, or an
+          external command prefixed with "cmd:")
+        func: callable, optional
+          Function to be called to obtain version information, if it is not
+          a Python module following standard ways to expose version via
+          `__version__`
```
kyleam, line 273: This description could be read as saying that `func` won't be called if `name` is not a Python module with `__version__`.  Perhaps "Function to be called to obtain version information.  This should be defined when checking the version of something that is not a Python module or when this class's method for determining the version of a Python module isn't sufficient."

datalad/support/external_versions.py
```diff
@@ -252,6 +255,43 @@ def keys(self):
...
+        interesting: bool, optional
+          Makes sense only for entries without func - either it is considered
+          "interesting" and should be included in output of dumps
```
kyleam, line 276: Is this needed?  Would anyone use this method with something they thought wasn't interesting?

datalad/support/external_versions.py
```diff
@@ -252,6 +255,43 @@ def keys(self):
...
+        ----------
+        name: str
+          Name of the check (usually a name of the Python module, or an
+          external command prefixed with "cmd:")
+        func: callable, optional
+          Function to be called to obtain version information, if it is not
+          a Python module following standard ways to expose version via
+          `__version__`
+        interesting: bool, optional
+          Makes sense only for entries without func - either it is considered
```
kyleam, line 275: s/either/whether/

datalad/support/external_versions.py
```diff
@@ -252,6 +255,43 @@ def keys(self):
...
+          "interesting" and should be included in output of dumps
+           (with query=True).
+        """
+        if func:
+            func_existing = self.CUSTOM.get(name, None)
+            was_known = False
+            if func_existing and func_existing is not func:
```
kyleam, line 282: Is there much value in supporting overrides?  What about simply raising an exception on a name conflict?

datalad/support/tests/test_external_versions.py
```diff
@@ -25,12 +25,17 @@
...
+    assert_equal,
+    assert_false,
+    assert_greater,
+    assert_greater_equal,
+    assert_in,
+    assert_not_in,
+    assert_raises,
+    assert_true,
 )
+from unittest.mock import patch
```
kyleam, line 38: Aside from trying to introduce conflicts :), why move this `patch` import?  My first guess was that you were sorting somehow, but that doesn't seem to be the case because `nose.SkipTest` import is below.

datalad/support/tests/test_external_versions.py
```diff
@@ -254,4 +259,21 @@ def test_check():
...
-        ev.check('datalad', min_version="10000000")  # we will never get there!
\ No newline at end of file
+        ev.check('datalad', min_version="10000000")  # we will never get there!
+
+
+def test_add():
+    ev = ExternalVersions()
+    ev.add('custom1', lambda: "0.1.0")
+    assert_in("custom1=0.1.0", ev.dumps(query=True))
+    assert "numpy" not in ev.INTERESTING   # we do not have it by default yet
```
kyleam, line 269: `assert_not_in`?

adswa:
> > It would be nice if extensions, where matters, could check for their
> > externals and possibly even register them within our "central"
> > ExternalVersions checker.
> 
> Nice.
> 
> re: master vs maint: Not surprisingly I'd vote for this to go to master.

@kyleam very important piece of unrelated information :upside_down_face: :
https://meet.jit.si/DataLad

datalad/support/external_versions.py
yarikoptic: what if definition for the same added by two  different extensions/modules but with different implementations?  Exception would cause the entire import to fail, due to possibly unrelated 3rd party extension.  So I thought that better to let it through here

datalad/plugin/wtf.py
yarikoptic: yes. I decided to not rely on strict typing but rather on generic programming - that we just want to go through both regardless what they are

datalad/support/external_versions.py
yarikoptic: well, that is a bigger question I guess. Before now, all the `INTERESTING` was just an attribute which nobody care about. Here I just thought to make sure to describe what effect it would have to have something "interesting" (that only then `dumps` would take care to query it if it was not queried).

datalad/support/tests/test_external_versions.py
yarikoptic: ;-) yeah, who knows.  I will RF as where it belongs in master version ;)

yarikoptic:
Thanks @kyleam , I have pushed an update against master.

datalad/plugin/wtf.py
kyleam: To avoid any confusion: These sorts of comments are "thinking aloud" review comments.  I personally like seeing such comments from reviewers, both for my own patches and others.  They are not some sort of indirect way of me saying "you should have done X instead".

datalad/support/external_versions.py
kyleam: Do you anticipate any caller of this method will want to add something and call `interesting=False`?  If not, I'd say this is unnecessary cruft.  I don't think it should exist just for the purpose of documenting `.INTERESTING`, which can be documented in more usual ways.

datalad/support/external_versions.py
kyleam: All right.

datalad/support/external_versions.py
yarikoptic: hm... ok -- you think I should just get rid of this option, and add to `INTERESTING` whenever there is no `func` provided?

kyleam: > you think I should just get rid of this option, and add to `INTERESTING` whenever there is no `func` provided?

yes

yarikoptic: ok, did in 83d010ecb361ff51658c93a0ef990f74ac8e894e

datalad/support/external_versions.py
```diff
@@ -252,6 +255,41 @@ def keys(self):
...
+                    "Adding a new custom version checker %s for %s, "
+                    "old one: %s", func, name, func_existing)
+                was_known = name in self._versions
+            self.CUSTOM[name] = func
+            if was_known:
+                # pop and query it again right away to possibly replace with a new value
+                self._versions.pop(name)
+                _ = self[name]
+        if name not in self.INTERESTING:
+            self.INTERESTING.append(name)
```
kyleam, line 291: Shouldn't this happen only when `func` isn't specified?

yarikoptic: I had it first as `elif` but then decided against it since in effect it would be the same since all of those are used only to populate the known versions.  But since indeed we are not adding callable checks to INTERESTING in general, I will change it now to `elif`.

yarikoptic: hm, actually @mih did consider `cmd:7z` so special that now we do have it also in INTERESTING.  I think at some point we should RF it so to make INTERESTING handling independent of `CUSTOM` i.e. CUSTOM not being queried by default in `dumps`, but then here argument `interesting` would come back to make it explicit.
So, for now -- I actually think it (`if` vs `elif`) doesn't matter one way or another here , so will leave it as is to not stress CI for no good reason, unless we decide to also fix up all this INTERESTING handling and add explicitly all the `cmd:git` etc into `INTERESTING` as well.

kyleam: It doesn't matter functionally, but it's confusing with the current code.  wtf and .dumps are chaining together CUSTOM and INTERESTING, so this will lead to unnecessary repeats.  Of course these don't matter performance-wise, but it's confusing, and, given it can be avoided with an `elif`, I don't see why we shouldn't.

re: cmd:7z: I'd say that's a mistake, at least with the way the current code is written.

yarikoptic: ok, what do you think about (me) extend this PR separating our CUSTOM from INTERESTING (so dumps queries only INTERESTING) and returning `interesting` option here?

kyleam: I guess I'm not so bothered by the current behavior of treating CUSTOM as "interesting" by definition and not requiring them to be part of INTERESTING.  Why else would someone add a custom function if they didn't consider it interesting?  And again, with the `.add` method, I don't see when someone would go through the trouble of adding a version checker if they didn't consider it interesting.

If things are reworked, I'd consider a more useful change to be a method/property that does the chaining of INTERESTING and CUSTOM underneath.  Then wtf/.dumps and others could call it that rather than worry about which attributes contain a list of interesting things.

So, I don't know, if you feel strongly that what you're proposing is an improvement, go for it, but I'm slightly negative on the change (as well as coupling it with this PR).  I'm also okay with no change to this PR at all, though I'd prefer to see the `elif` here.

yarikoptic: Ok, will do `elif` and since changes to come:
yes -- I thought to add such a method as well -- will do right here.

yarikoptic: FWIW, avoided a new method, see d9aaae807d279d95d93f2f022541ecb2cecabd35 pushed

datalad/support/tests/test_external_versions.py
```diff
@@ -262,3 +263,18 @@ def test_check():
...
+
+
+def test_add():
+    ev = ExternalVersions()
+    ev.add('custom1', lambda: "0.1.0")
+    assert_in("custom1=0.1.0", ev.dumps(query=True))
+    assert_not_in("numpy", ev.INTERESTING)  # we do not have it by default yet
+    assert_not_in("numpy=", ev.dumps(query=True))
+    ev.add('numpy')
+    assert_in("numpy=", ev.dumps(query=True))
```
kyleam, line 275: Fails on appveyor: https://ci.appveyor.com/project/mih/datalad/builds/32511871/job/ddbp8p80apy7i34r#L1038

I can trigger the same failure locally.  I think the issue is that `__getitem__` has an early return of None on an import error:

https://github.com/datalad/datalad/blob/d9aaae807d279d95d93f2f022541ecb2cecabd35/datalad/support/external_versions.py#L220-L244

So `._version.keys()` will not have "numpy" present if numpy isn't installed in the environment.  I think this test was assuming that the output would include "numpy=UNKNOWN" in this case.  This feels like a logic bug in external_versions (exposed, not added, by this PR).

yarikoptic: THANK YOU for the analysis!  I first thought that indeed code is wrong but then decided that  code is correct as is: We do have a notion of "UNKNOWN" version but that is when the external **is** present but we do not know its version.  If external is not present, then `__getitem__` returns None, as to signal that it is really None, i.e. no any kind of version of the external is available.
I have pushed a fix to the test in d4dcc2c4db2b7c7b9ed9081fa65c0b0984cc12a4 . Hopefully this would make this what should have been a trivial PR finally ready ;-)

kyleam:
CrippledFS failure in last run looks unrelated: https://github.com/datalad/datalad/pull/4441/checks?check_run_id=628312224

Issue #4390, opened 2020-04-09
No overall progress indication for `get -n`


mih:
I am installing the full https://github.com/datalad-datasets/human-connectome-project-openaccess. Given that I know that a few thousand repos need to be cloned I know that it will take a while, but there is no overall progress reporting that would indicate how long approximately.

This is likely hard to do in a meaningful way (subdatasets are only discovered along the way), but I think it would be better to update the total whenever new information arrives, rather than no having a total at all.

To be clear, there is progress reporting, but only within each clone attempt, not across all clones.

mih:
Thank you for reporting this issue!

However, at this point I am closing it at a time of ~600 open issues, in an attempt to regain control over the issue tracker. This particular issue was posted more than 6 months ago, has not received a single response, and does not describe an immediate software defect. *Closing this issue does not imply that it is not relevant*. If any reader objects the closing of this issue, please feel more than free to reopen it with an update on how it is relevant for the current state of the project, and which concrete actions to address it are ongoing or planned.

PR #4444, opened 2020-04-27
ENH: Draft of an improved progress experience

mih:
Results of an exploration with @adswa

- give `subdatasets()` a progress bar, because it gives `get()` a
  progress bar (and many other functionality that iterates over
  datasets). Similarly, such progress logging can be put into
  status/diff.

- update the subdataset total, as new information comes in. It is
  prohibitive expensive to precompute this information. This is the
  majority of time spend in this command

- clear the bars before each ui.message() (this include result renderer
  output) to keep them nicely at the bottom of the screen, instead of
  butchering them with our own output

#### Issues

- any INFO log message ruins any progress bar display (all git/annex
  output is reported at the info level: Scanning for unlocked files,
  etc)

- we cannot prevent that using logger config, because we log progress
  also at the INFO level. It would be better to position a dedicated
  PROGRESS level between INFO and WARNING

Fixes gh-4390

### DIFF
--- a/datalad/local/subdatasets.py
+++ b/datalad/local/subdatasets.py
@@ -46,6 +46,7 @@ from datalad.distribution.dataset import (
     datasetmethod,
     resolve_path,
 )
+from datalad.log import log_progress
 
 lgr = logging.getLogger('datalad.local.subdatasets')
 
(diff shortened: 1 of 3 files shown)


yarikoptic:
> * any INFO log message ruins any progress bar display (all git/annex
>   output is reported at the info level: Scanning for unlocked files,
>   etc)

possible solution to go along with 

> - clear the bars before each ui.message()

would be to use `ui` (and its `message`) as actual log handler, which is ATM (unless we dump to a file) is 
```python
        if logtarget.lower() in ('stdout', 'stderr'):
            loghandler = logging.StreamHandler(getattr(sys, logtarget.lower()))
```

I have not looked what methods StreamHandler expects from the stream (probably .write, .flush).  Since all the logic on either display or not is done before passing content to the handler, on typical log level of INFO I would not expect any negative effect on performance with even active progress bars getting cleared once in a while.

mih:
The key part has been merged with #4555. The rest isn't cooked enough yet. No need to pollute the long PR list further.

PR #4447, opened 2020-04-27
Some RFings and few added type hints

yarikoptic:
Well, I just thought to run `mypy` on `datalad.cmd` and tried to address some of the issues it has detected.  It is by no means an intensive attempt to introduce type hints into our code, but I do not see why we shouldn't start doing that especially when typing is clear and we do not do some duck typing etc.

### DIFF
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -27,6 +27,8 @@ from collections import (
     OrderedDict,
 )
 
+from typing import Optional, Callable
+
 from .consts import GIT_SSH_COMMAND
 from .dochelpers import (
     borrowdoc,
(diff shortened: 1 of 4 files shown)


yarikoptic:
oh, adding import of typing adds ~50msec ... might not be worth for these few cases and would need to be put into back burner for now since no benefit and wasted time.  Let's just review and then I strip the `ENH(TYP)` to be possibly added later

yarikoptic:
ok, nobody got excited.  I will send the shortened PR with RFs

yarikoptic:
#4457 came with those RFs, closing this PR

PR #4453, opened 2020-04-28
RF: make push stop pretending being publish in its result records/messages

yarikoptic:
I was trying to compare behavior of `publish` and `push` and was confused (thought that I still had `publish` command somewhere) by the produced output claiming that `publish (ok)`.

May be I have missed/forgot the intention for `push` to eventually replace `publish` command, where such "pretense" of  it being `publish` would make sense, but then may be it should happen for 0.13.0 release instead of introducing a new command to just later rename it back into `publish`?  If not -- then it better gets its own name in the messages (and that  is what this PR attempted to do -- didn't check where else `push` already used and expected to look like `publish`).

### DIFF
--- a/datalad/core/distributed/push.py
+++ b/datalad/core/distributed/push.py
@@ -253,7 +253,7 @@ class Push(Interface):
             yield dict(
                 res_kwargs,
                 status='notneeded',
-                message='Given constraints did not match any changes to publish',
+                message='Given constraints did not match any changes to push',
                 type='dataset',
                 path=ds.path,
             )
@@ -330,7 +330,7 @@ def _datasets_since_(dataset, since, paths, recursive, recursion_limit):
             if any(arg == p for arg in paths) and \
                     not GitRepo.is_valid_repo(res['path']):
                 raise ValueError(
-                    'Cannot publish subdataset, not present: {}'.format(res['path']))
+                    'Cannot push subdataset, not present: {}'.format(res['path']))
 
         if parentds != cur_ds:
             if ds_res:
--- a/datalad/core/distributed/tests/test_push.py
+++ b/datalad/core/distributed/tests/test_push.py
@@ -131,7 +131,7 @@ def check_push(annex, src_path, dst_path):
     assert_result_count(res, 2 if annex else 1)
     assert_in_results(
         res,
-        action='publish', status='ok', target='target',
+        action='push', status='ok', target='target',
         refspec='refs/heads/master:refs/heads/master',
         operations=['new-branch'])
 
@@ -164,7 +164,7 @@ def check_push(annex, src_path, dst_path):
     res = src.push(to='target', since="HEAD~2", jobs=2)
     assert_in_results(
         res,
-        action='publish', status='ok', target='target',
+        action='push', status='ok', target='target',
         refspec='refs/heads/master:refs/heads/master',
         # we get to see what happened
         operations=['fast-forward'])
@@ -201,14 +201,14 @@ def check_push(annex, src_path, dst_path):
         assert_result_count(res, 1)
         assert_in_results(
             res,
-            action='publish', status='error', target='target',
+            action='push', status='error', target='target',
             refspec='refs/heads/master:refs/heads/master',
             operations=['rejected', 'error'])
         # push with force=True works:
         res = src.push(to='target', since='HEAD~2', force='gitpush')
         assert_in_results(
             res,
-            action='publish', status='ok', target='target',
+            action='push', status='ok', target='target',
             refspec='refs/heads/master:refs/heads/master',
             operations=['forced-update'])
         eq_(list(target.get_branch_commits_("master")),
@@ -252,7 +252,7 @@ def test_push_recursive(
     assert_in_results(
         res, path=top.path, type='dataset',
         refspec='refs/heads/master:refs/heads/master',
-        operations=['new-branch'], action='publish', status='ok',
+        operations=['new-branch'], action='push', status='ok',
         target='target')
     for d in (sub, subsub, subnoannex):
         assert_in_results(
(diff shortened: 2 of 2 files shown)


yarikoptic:
woohoo - no other tests were hurt.

<details>
<summary>Travis failure in a single run is unrelated and for some reason I see no "Restart" button</summary> 

```shell
ERROR: datalad.core.distributed.tests.test_clone.test_clone_dataset_from_just_source
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/datalad/tests/utils.py", line 1018, in newfunc
    t(*(arg + (uri,)), **kw)
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/datalad/tests/utils.py", line 691, in newfunc
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/datalad/core/distributed/tests/test_clone.py", line 204, in test_clone_dataset_from_just_source
    ds = clone(url, result_xfm='datasets', return_type='item-or-list')
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/datalad/interface/utils.py", line 494, in eval_func
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/datalad/interface/utils.py", line 482, in return_func
    results = list(results)
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/datalad/interface/utils.py", line 469, in generator_func
datalad.support.exceptions.IncompleteResultsError: Command did not complete successfully [{'action': 'install', 'path': '/tmp/datalad_temp_test_clone_dataset_from_just_sourcefwe25ba9/testrepo--basic--r1', 'type': 'dataset', 'status': 'error', 'message': ('Failed to clone from any candidate source URL. Encountered errors per each url were:\n- %s', "git://github.com/datalad/testrepo--basic--r1\n  CommandError: 'git clone --progress git://github.com/datalad/testrepo--basic--r1 /tmp/datalad_temp_test_clone_dataset_from_just_sourcefwe25ba9/testrepo--basic--r1' failed with exitcode 128 [err: 'Cloning into '/tmp/datalad_temp_test_clone_dataset_from_just_sourcefwe25ba9/testrepo--basic--r1'...\nfatal: unable to connect to github.com:\ngithub.com[0: 140.82.112.3]: errno=Connection timed out']"), 'source_url': 'git://github.com/datalad/testrepo--basic--r1'}]
```
</details>

yarikoptic:
ok, lets agree in #4461 one way or another and close this PR if push becomes publish.

yarikoptic:
Based on my [summary of now closed #4461](https://github.com/datalad/datalad/issues/4461#issuecomment-625378805) this PR remains relevant - `push` is not intended to become `publish` thus I do not see why it should keep pretending.

mih:
I am fine with your decision to kill publish.

kyleam:
@yarikoptic says

> Based on my [summary of now closed #4461](https://github.com/datalad/datalad/issues/4461#issuecomment-625378805) this PR remains relevant - `push` is not intended to become `publish` thus I do not see why it should keep pretending.

What @mih said above does not depend on `push` being renamed to `publish`:

> Both commands achieve the same result and the timing of the reported results is trying to imitate `publish`. Anyone using result hooks will appreciate to not have to reconfigure everything when switching from `publish` to `push`.
> 
> We used the same approach for `clone()/install()` and it did not surface as a problem so far.

---

In my view, there is nothing conceptually problematic about `push` using `action="publish"`.

mih:
And it uses more than one action label in its reports.

yarikoptic:
> And it uses more than one action label in its reports.

aren't those actions correspond to the commands we have (or used to have)?

mih:
> > And it uses more than one action label in its reports.
> 
> aren't those actions correspond to the commands we have (or used to have)?

In addition to the counter-examples given above, there is also `save()` that reports `add` results (among others). The answer seems to be "no".

yarikoptic:
> > > And it uses more than one action label in its reports.
> > 
> > 
> > aren't those actions correspond to the commands we have (or used to have)?
> 
> In addition to the counter-examples given above, there is also `save()` that reports `add` results (among others). The answer seems to be "no".

We used to have `add` command, so the answer is still "yes".  Also there is underlying `git add` or `git annex add`.  With `publish` gone, there would be nothing for that `publish` to relate to, especially for a new user who is not burdened with knowledge of previous interfaces.

mih:
> We used to have `add` command, so the answer is still "yes". Also there is underlying `git add` or `git annex add`. With `publish` gone, there would be nothing for that `publish` to relate to, especially for a new user who is not burdened with knowledge of previous interfaces.

I cannot follow the logic: For historical reasons it is ok for `save` to emit `add` results, and it is ok for `clone` to emit `install` results, but *not* ok for `push` to emit `publish` results.

In any case, the amount of time that went into this discussion already doesn't compare well to the level with which I care about this issue. Please feel free to rename however you see fit, and leave guidelines that would avoid such a discussion in the future.

yarikoptic:
Updated. Following discussion during weekly meeting - we do have ambiguity between "reported commands" and "intended actions" (what hooks would need to hook up).  We might want to indeed make separation clear(er).

Another relevant which wasn't mentioned is configuration -- we have `publish_depends` config setting.

mih:
What is the trajectory for this PR now, after the release of 0.13?

yarikoptic:
I guess nobody gave enough damn, and it is just waiting for you to have it closed.  Let's just wait on user feedback in issues ;)

yarikoptic:
yeap, we should have came up with a word "pushpublish" and declared it "done".  I am ok with pretend games

Issue #4424, opened 2020-04-18
Support convenience aliases in RIA stores


mih:
It would be nice, if some datasets in a RIA store could be addressable via non-UUID convenience aliases. Those could be symlinks on the file system, but the resolving code would also need a change to accept them.

Use case: All 4.5k HCP subdatasets live in a store. It would be nice to not have to post the superdataset on another platform, only to give it a non-UUID URL. Doing the initial clone from the same store also get's all the subdataset resolving features readily configured, such that no tuning of the superdataset for a specific publication location would be needed.

bpoldrack:
Nice to have, but not completely clear to me how to approach it. We would need to accept sth like `ria+http://mystore.org/base#<some label>`. So we would need to recognize whether or not or how to resolve such a URL. Generally the URL resolution should prob. not require a look up at the remote end as long as this is avoidable.
Therefore might be easiest to just use a different URL form to distinguish from ID based resolution. Different marker than `#` or an additional one. Say `ria+http://mystore.org/base##<label>`. This could then resolve to `http://mystore.org/base/label` which needs to point to such a symlink. Note, that we can't just use a not-to-be resolved URL, while simultanously making sure to recognize it's a RIA store and not just pointing to the base path of it. This approach would also allow to have different ways to resolve those later on based on the store itself (lookup resolution rule or sth).

mih:
`##` is one way to deal with it. Another would be to simply check if the suffix is a UUID or not (already done at some point). Anything that is not, could be taken literally.

bpoldrack:
Sure. That's simpler. Just potentially less flexible re future additions in how to interprete what comes after `#` or how to deal with a label.
However, I'd consider this pretty much the same approach.

What would be really different is some kind of lookup at the remote end (dedicated file w/ lookup table for labels?), so that we actually resolve to the real ID on our end. In a way that's more "beautiful", because it sticks to the ID based approach. But it would mean to not be able to resolve it without a connection to the store.

Wonder whether there's a clever third way.

bpoldrack:
One more argument for a lookup like thing: Inodes. We can have thousands of labels in a table.

mih:
I thought about this for a while now. I could not come up with a reason why we would *want* to "actually resolve to the real ID on our end" **before** a clone. Anything prior to a clone doesn't need a UUID, and anything after a clone has trivial access to it. Do you have something concrete in mind?

[...]

PR #4459, opened 2020-04-29
ENH: Support dataset aliases in RIA stores

mih:
`clone()` can now handle URLs of the format

   `ria+<protocol>://<storelocation>#~<aliasname>`

in addition to those containing UUIDs (examples inside).

A future addition should likely equip `create-sibling-ria()` with the ability
to establish those aliases too. But for now it is not clear to me how to
do that best (in particular how to perform properly on reconfigure).
Given time-pressure this needs to wait for another time.

Fixes gh-4424

### DIFF
--- a/datalad/core/distributed/clone.py
+++ b/datalad/core/distributed/clone.py
@@ -141,9 +141,9 @@ class Clone(Interface):
         dict(text="Install the main superdataset from datasets.datalad.org",
              code_py="clone(source='///')",
              code_cmd="datalad clone ///"),
-        dict(text="Install a dataset identified by its ID from store.datalad.org",
-             code_py="clone(source='ria+http://store.datalad.org#76b6ca66-36b1-11ea-a2e6-f0d5bf7b5561')",
-             code_cmd="datalad clone ria+http://store.datalad.org#76b6ca66-36b1-11ea-a2e6-f0d5bf7b5561"),
+        dict(text="Install a dataset identified by a literal alias from store.datalad.org",
+             code_py="clone(source='ria+http://store.datalad.org#~hcp-openaccess')",
+             code_cmd="datalad clone ria+http://store.datalad.org#~hcp-openaccess"),
         dict(
             text="Install a dataset in a specific version as identified by a"
                  "branch or tag name from store.datalad.org",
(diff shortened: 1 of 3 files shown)


bpoldrack:
LGTM. Somehow I like the tilde.

datalad/core/distributed/clone.py
```diff
@@ -926,8 +926,16 @@ def decode_source_spec(spec, cfg=None):
         dsid, version = source_ri.fragment.split('@', maxsplit=1) \
             if '@' in source_ri.fragment else (source_ri.fragment, None)
         uuid_regex = r'[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}'
-        if not re.match(uuid_regex, dsid):
-            raise ValueError('RIA URI does not contain a valid dataset ID: {}'.format(spec))
+        if re.match(uuid_regex, dsid):
+            trace = '{}/{}'.format(dsid[:3], dsid[3:])
+            default_destpath = dsid
+        elif dsid.startswith('~'):
+            trace = 'alias/{}'.format(dsid[1:])
```
yarikoptic, line 933: since this is more of an internal purpose, why not to have it as `.alias/` or even `.ria/aliases/` or alike?  then you could still have an `alias` dataset (happen ria allow it at some point)

yarikoptic:
sorry if I am missing something.  
Why bother with explicit `~` handling and not just rely on symlink based URL like `ria+<protocol>://<storelocation>/<aliasname>`? where that aliasname is a symlink?  I just symlinked `abide` as `abide-link` and `git clone http://datasets.datalad.org/abide-link/.git` worked just fine having symlinks forwarded.

mih:
All ria URLs require explicit handling. None are meant to expose the underlying store organization. That organization is versioned and may change. URLs must not change in that case.

datalad/core/distributed/clone.py
mih: I think it is fine to not have further obfuscation. Having a dataset stored in a directory named like this would represent a fundamental concept change. I doubt that we would keep the name or URL scheme at that point.

yarikoptic:
> That organization is versioned and may change.

haven't spotted any reflection of any possible versioning of the "organization" in the code ATM... but it is ok with me if you say that no direct symlinks would work.

kyleam:
This looks nice implementation-wise, and I don't have any issue with using `~` in the URL or `alias/` in the store, but I'm confused by this statement:

> All ria URLs require explicit handling. None are meant to expose the underlying store organization.

It seems to me that there is freedom to restructure only in terms of future cloners.  Wouldn't it break all existing clones that already point directly to the resolved location?  An example clone from a ria test:

```
$ git remote -v
origin	http://127.0.0.1:43805/850/9d199-8a4f-11ea-9dee-28c63f920482 (fetch)
origin	http://127.0.0.1:43805/850/9d199-8a4f-11ea-9dee-28c63f920482 (push)
```

mih:
> This looks nice implementation-wise, and I don't have any issue with using `~` in the URL or `alias/` in the store, but I'm confused by this statement:
> 
> > All ria URLs require explicit handling. None are meant to expose the underlying store organization.
> 
> It seems to me that there is freedom to restructure only in terms of future cloners. Wouldn't it break all existing clones that already point directly to the resolved location? An example clone from a ria test:
> 
> ```
> $ git remote -v
> origin	http://127.0.0.1:43805/850/9d199-8a4f-11ea-9dee-28c63f920482 (fetch)
> origin	http://127.0.0.1:43805/850/9d199-8a4f-11ea-9dee-28c63f920482 (push)
> ```

Yes, correct. The explicitness is there, after the ria URLs get resolved. That is no change from before. I think, if persistence like that is desired/required than it should either be achieved at the server side, or with a dedicated remote helper

re @yarikoptic You are correct that there is no trace of versioning in the code, because we are still at version `1`, but the stores are versioned: http://store.datalad.org/ria-layout-version

mih:
Given the carefully positive feedback, I will merge this now. thx!

Issue #4478, opened 2020-05-02
Make exc_str() silent by default


mih:
We are trying to use this function to make it easier to debug exceptions. However, debugging isn't always necessary and certainly not a desire of most users. The function has (at least) this kind of impact on (error) messages:

```
w/o:  [Errno 39] Directory not empty: 'dir'
with: [Errno 39] Directory not empty: 'dir' [pathlib.py:rmdir:1302]
```

If we want to use this function unconditionally, and we acknowledge that, in the general case, we cannot decide whether an exception is indeed an indication of an error (in the original context of the example above, it was not #4430), then this function should be silent by default and not add additional information.

Output that is unconditionally tuned for debugging is only suitable for code that is so fragile that it isn't really expected to work INMO. This may be true from some pieces, but certainly should not be true for DataLad as a whole.

The above example is actually not representative of the real impact. In many cases, DataLad error messages flood the terminal with tons of info already, and the additions by `exc_str()` just add the icing to the cake.

mih:
I looked into this now, and I think I see a possibility to change the concept of `exc_str()` to make it easier to work with (and faster).

My issue with this function is that it calls `str(exc)`, unconditionally and at the time of log message composition (not when it is eventually emitted, if at all). It is unforeseeable how much that costs, because it depends on the exception itself. Moreover, it moves all decision making on how the most appropriate formatting looks like into the log message composition, where it is cumbersome to make things conditional on whether a message will actually be seen.

I think we should stop turning exceptions into strings with amended information, and instead amend the exception objects, while leaving the formatting to the `Logger`, where it already happens, and at the latest possible stage.

Proposal:

- stop using `exc_str()` and replace with `exc_add_tb()`
- `exc_add_tb()` adds an attribute `datalad_tb_entries` to an exception object that matches the filtered selection `entries[-limit:]` that is presently used for str formatting in `exc_str()`.
- the actual exception str formatting moves into `datalad.log.ColorFormatter.format()` that receives the log message args as a tuple (`record.args`), that can be easily transformed by looping over all args and formatting the ones that are of type `Exception` and have our custom traceback information. The rest is left untouched.

IMHO this is a better fit for the design of `logging`. The user-facing behavior is unchanged, information collection is now longer conflated with information display.

bpoldrack:
I like that idea.

yarikoptic:
[...]

PR #4479, opened 2020-05-02
RF(OPT): make exc_str into a class which would do needed only upon __str__

yarikoptic:
- also queries env variable DATALAD_EXC_STR_TBLIMIT once in the runtime at
  the class level.  Unlikely would be desired to be changed in the running
  process
- now str() evaluation of exc_str must happen within "except" block and
  should not be masked by some other exception which might happen between
  creating an instance and `__str__` being called

Altogether I am not really sure if there would be any measurable impact from such optimization in real life.

ref: #4478 (the "performance" aspect of it, not making it silent here)

### DIFF
--- a/datalad/dochelpers.py
+++ b/datalad/dochelpers.py
@@ -314,52 +314,64 @@ def borrowkwargs(cls=None, methodname=None, exclude=None):
 
 
 # TODO: make limit respect config/environment parameter
-def exc_str(exc=None, limit=None):
+class exc_str(object):
     """Enhanced str for exceptions.  Should include original location
-
-    Parameters
-    ----------
-    exc: Exception, optional
-      If not provided, information about the last exception caught by except
-      clause (as provided by `sys.exc_info`) will be used.
-    limit: int, optional
-      Number of levels in the traceback stack from the point where exception
-      was raised to include. If `None`, environment variable
-      DATALAD_EXC_STR_TBLIMIT is consulted.
-
-    Returns
-    -------
-    str
-      String representation of the exception with traceback information
-      appended.
     """
-    out = str(exc)
-    if limit is None:
-        # TODO: config logging.exceptions.traceback_levels = 1
-        limit = int(os.environ.get('DATALAD_EXC_STR_TBLIMIT', '1'))
-    try:
-        exctype, value, tb = sys.exc_info()
-        if not exc:
-            exc = value
-            out = str(exc)
-        if not out:
-            out = repr(exc)
-        # verify that it seems to be the exception we were passed
-        #assert(isinstance(exc, exctype))
-        if exc:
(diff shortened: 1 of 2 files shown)


datalad/dochelpers.py
```diff
@@ -314,52 +314,64 @@ def _borrowkwargs(method):
...
                 )
-            )
-    except:  # MIH: TypeError?
-        return out  # To the best of our abilities
-    finally:
-        # As the bible teaches us:
-        # https://docs.python.org/2/library/sys.html#sys.exc_info
-        del tb
-    return out
+        except IOError:  # MIH: TypeError?
```
mih, line 371: The addition of the exception type addresses my comment, I can be remove. Thx!

datalad/dochelpers.py
```diff
@@ -314,52 +314,64 @@ def _borrowkwargs(method):
...
-                    for x in entries[-limit:]
+    def __str__(self):
+        exc = self._exc
+        out = str(exc)
+        limit = self._limit
+        if limit == 0:
+            return out
+
+        try:
+            # TODO: think if we would still be getting correct exception
```
mih, line 351: AFAICS we should. Unless in the time between the desired exception and the log message formatting another exception occurred. Such an exception would then happen in generic log handling code, and better be fixed to begin with.

datalad/dochelpers.py
yarikoptic: Oh, that was unintentionally committed I think - I was testing something. Will need to remove. The point of swallowing all was to not cause main process fail of something trips in this minor helper invocation

datalad/dochelpers.py
kyleam: I agree, this doesn't look problematic to me.

mih:
@yarikoptic What is your plan for this PR? It looks like the failing tests are all related to functionality that inspects the error messages -- which would need updating. But otherwise, do you plan to go  ahead with this change? Or do you need help?

yarikoptic:
I don't quite like it any more and think it is not really worth it.  It was a public facing function returning a string, possibly used in extensions (I see some failing) and who knows were else.  Now it would become not a string (as those failures hinting on that some code managed to assume it). IMHO doing this type change is is not worth the benefit.  
As I had mentioned before -- I do not think this one would make **any** dent in our sluggish performance. E.g. with #4526 I expect to save myself 100x more time than I would have saved with `exc_str` being gone completely.
And if someone decides to look into replacing it, most likely we should check out [passing original exception](https://docs.python.org/3/library/logging.html#logging.Logger.debug) via `exc_info` and see if we could format it more sanely (by default), and then just migrate to use it instead of `exc_str` which we deprecate "properly" then if decide to remove entirely.

PR #4480, opened 2020-05-04
DOC: Fixed documentation builder to allow for overriding opts in eval…

### LINKED ISSUE #2762: Documentation - dataset.siblings() does not have result_renderer to None by default
#### What is the problem?
In the [documentation](http://docs.datalad.org/en/latest/generated/datalad.api.siblings.html), it is said that "sibling information is rendered as one line per sibling following this scheme". However, next to the `result_renderer` parameter, it is said to be `None` by default. However, it is required to manually set it to `None` in order to hide the siblings rendering.
#### What version of DataLad are you using?
Datalad: datalad 0.10.1
OS: Ubuntu 18.04.1 LTS

andycon:
…_defaults. Closes #2762

Edited the decorator function build_doc in interface/base.py, so that when a class __call__ function specifies another default value other than that in eval_defaults/eval_params from common_opts, the specified default properly overwrites the generic default value, and repacked the default value in the doc string.


datalad/interface/base.py
```diff
@@ -510,6 +510,16 @@ class defining a datalad command
 
     # build standard doc and insert eval_doc
     spec = getattr(cls, '_params_', dict())
+
+    # update defaults for any opts in spec that can override eval_defaults
+    overriders = dict()
+    for opt in spec.keys():
+        if opt in eval_params.keys():
```
yarikoptic, line 517: no need for an explicit `.keys()` in either of above two lines or below

datalad/interface/base.py
```diff
@@ -510,6 +510,16 @@ class defining a datalad command
...
+            overriders[opt] = spec[opt]
+    for opt in overriders.keys():
+        xx = spec.pop(opt)
```
yarikoptic, line 520: since do get `.pop`ed at the end anyways, `.pop` could happen at the point of assigning to `overriders`. And entire code block would be more concise and descriptive as a dict comprehension, such as 
```python
overriders = {
  opt: spec.pop(opt)
  for opt in spec
  if opt in eval_params
}
```

yarikoptic:
datalad-container failure (in running `wtf`!) -- I wonder if it could be related since I do not remember it seeing failing otherwise recently:
```
Run datalad wtf
[ERROR] value is not one of ('ignore', 'continue', 'stop') [constraints.py:__call__:289] (ValueError) 
##[error]Process completed with exit code 1.
```
that is getting deep into the weeds ;-/

kyleam:
> datalad-container failure (in running `wtf`!) -- I wonder if it could be related since I do not remember it seeing failing otherwise recently:

It is related.  I have a review coming.

datalad/interface/base.py
```diff
@@ -510,6 +510,16 @@ class defining a datalad command
...
+    eval_defaults.update(overriders)
```
kyleam, line 521: So, it seems like you added `result_renderer` to `Siblings._params_` to get access to it here.  But that's not necessary, because it's already an attribute on the class.  This means you can take care of all the overrides, without adding each override to the class's `_params_`.  Perhaps something like this:

```diff
diff --git a/datalad/interface/base.py b/datalad/interface/base.py
index 9c47d029a..618dc0282 100644
--- a/datalad/interface/base.py
+++ b/datalad/interface/base.py
@@ -511,14 +511,10 @@ class defining a datalad command
     # build standard doc and insert eval_doc
     spec = getattr(cls, '_params_', dict())
 
-    # update defaults for any opts in spec that can override eval_defaults
-    overriders = dict()
-    for opt in spec.keys():
-        if opt in eval_params.keys():
-            overriders[opt] = spec[opt]
-    for opt in overriders.keys():
-        xx = spec.pop(opt)
-    eval_defaults.update(overriders)
+    if hasattr(cls, '_no_eval_results'):
+        add_args = None
+    else:
+        add_args = {k: getattr(cls, k, v) for k, v in eval_defaults.items()}
 
     # get docs for eval_results parameters:
     spec.update(eval_params)
@@ -527,7 +523,7 @@ class defining a datalad command
         cls.__call__, spec,
         prefix=alter_interface_docs_for_api(cls_doc),
         suffix=alter_interface_docs_for_api(call_doc),
-        add_args=eval_defaults if not hasattr(cls, '_no_eval_results') else None
+        add_args=add_args,
     )
 
     # return original
```

Re: the datalad-container failure: I confirmed locally that your change triggers it (and the proposed diff above does not), though I haven't yet debugged it to understand what's going on.

andycon:
From that, I take away that the problem is that, whenever a class overrides the default for something in eval_params, the Python documentation shows an incorrect default.

kyleam:
Andy: Thanks for the updates.  Why did you delete your branch and close this PR?

andycon:
I did not mean to do that. I am still learning how to use the wed interface
on GitHub.
I will send a new PR now.

ac

kyleam:
> I did not mean to do that. I am still learning how to use the wed interface on GitHub. I will send a new PR now.

Okay, no worries.  Either way works, but you may just be able to restore your branch and reopen this PR (via github buttons).

andycon:
ok. I will try that.
ac

yarikoptic:
> Those look sensible to me, so this seems good to go from my point of view,

hm... 
```
[Default: <function is_result_matching_pathsource_argument at 0x7f5698338598>]
```
and
```
[Default:  constraint:(action:{create} and status:{ok, notneeded})]
```
seems to beg for some tune up?

kyleam:
> seems to beg for some tune up?

I wouldn't mind if someone cared to improve the presentation, but surely those are preferable to showing an incorrect default?

yarikoptic:
That's true. And might be worth it's dedicated discussion. So I will let Andy decide either he is willing to add a unit test here to ensure (!) correct behavior going forward for this particular change. I think it would be very valuable.

andycon:
I'm happy to write a unit test for this. The test should check all instances in the project where decorator @build_doc is applied and then ensure (!) that the values in the doc string match the actual values of the class attributes. This would only check attributes included in eval_defaults.

Does that sound right?

kyleam:
> The test should check all instances in the project where decorator @build_doc is applied and then ensure (!) that the values in the doc string match the actual values of the class attributes. [...]
> Does that sound right?

I think that sounds correct, though more ambitious than at least I had in mind.  In my view checking the `.__call_.__doc__` for a couple of commands that we know don't use the defaults would be sufficient.

yarikoptic:
yeah, would be great to have but too ambitious indeed.  I would have created a new ad-hoc interface command for the sake of testing with known properties, and tested that one -- a "true" unit-test ;-)  Actually we even have one already in https://github.com/datalad/datalad/blob/master/datalad/interface/tests/test_utils.py#L204 so you could just tune it up changing it so it would fail test without the introduced here fix, and then passing it with the fix

yarikoptic:
Hi @andycon .  Please let us know your plan with this one -- test or no test? ;-) Cheers

kyleam:
@andycon Just a heads up: I'm about to bring in gh-4482, which will cause a conflict here.  If you have any questions about resolving it, let me know.

datalad/interface/tests/test_utils.py
```diff
@@ -255,6 +258,10 @@ def test_eval_results_plus_build_doc():
     assert_in("Parameters", doc1)
     assert_in("It's a number", doc1)
 
+    # docstring shows correct overide values of defaults in eval_params
```
yarikoptic, line 261:
```suggestion
    # docstring shows correct override values of defaults in eval_params
```
I will add this typo fix and merge/push directly.

Thanks @andycon !

PR #4481, opened 2020-05-04
BF: fix handling of environment variables that contain whitespace

christian-monch:
This PR encloses the shell-substitution for PATH and PYTHONPATH in the test-code target of the Makefile in double-quotes. That prevents errors where the original PATH or PYTHONPATH environment variables contain whitespaces.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ clean:
 
 bin:
 	mkdir -p $@
-	PYTHONPATH=bin:$(PYTHONPATH) $(PYTHON) setup.py develop --install-dir $@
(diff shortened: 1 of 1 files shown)


yarikoptic:
ideally should be positioned over `maint` branch

christian-monch:
> Thanks.
> 
> Shouldn't another spot (a couple of lines above) receive the same treatment?

Thanks, you are right. Fixed in the latest commit.

kyleam:
> ideally should be positioned over `maint` branch

I'll change the base (though given the nature of the changes and the timing, I don't think it matters much in this case).

kyleam:
I've canceled the tests because none of them are relevant for this change.

PR #4492, opened 2020-05-07
ENH(TST): Run travis against oldest supported annex

yarikoptic:
Unfortunately ATM we do not have in snapshots-neuro the minimal version matching the one we claim to support (7.20190503): http://snapshot-neuro.debian.net/package/git-annex/, so I went for the next one available from two months after.

TODOs:
- [x] verify that all tests pass.  If not -- we need to either skip some tests or boost our minimal version or both ;) (not sure if worth fixing anything in the actual code)
- [ ] ideally there should be a unittest which would verify that if that env var is set, that used git-annex is the one we expect so we do not miss something breaking that "installed minimal version" testing 
- [ ] Remove TEMP commit so we run only one matrix run on cron (actually may be I should just add another cron entry which would just set that variable explicitly in its settings so all matrix runs run?)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -122,6 +122,14 @@ matrix:
     # Just so we test if we did not screw up running under nose without -s as well
     - NOSE_OPTS=
     - _DL_CRON=1
+  # run with minimal supported git-annex version as defined in AnnexRepo.GIT_ANNEX_MIN_VERSION
+  # TODO: ATM we do not have that minimal version as a Debian package in
+  # snapshots! When we do match - add a unit-test to guarantee that if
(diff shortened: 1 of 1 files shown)


yarikoptic:
wow -- tests passed (metalad environment preparation bugs out for some apt-key reason... unrelated). WTF does report ` cmd:annex: 7.20190708+git9-gfa3524b95-1~ndall+1` so correct one is used. That is great! 
I think I will scrape the 2nd TODO for now, and will now remove the TEMP commit and will just directly push into `maint` to avoid a needless waste of CI.

Issue #4499, opened 2020-05-07
tests stall (again!) with recent git-annex


yarikoptic:
There was a fight for a while trying to figure out what is going on:

- e.g. the https://github.com/datalad/datalad/pull/4265
- annex issue: https://git-annex.branchable.com/bugs/enableremote_stuck_with_a_recentish_git-annex/

which was "mitigated" by adding a skip of the test which checks that git-annex creates a socket if we do not pass configuration option to use the one we create: https://github.com/datalad/datalad/pull/4288 . And I thought that we are "all good"ish at least for now.  But then @kyleam today has mentioned that our cron jobs are hanging! And indeed looking at maint or master -- we have hangs eg https://travis-ci.org/github/datalad/datalad/jobs/683567878
```shell
$ http_proxy= PATH=$PWD/../tools/coverage-bin:$PATH $NOSE_WRAPPER `which nosetests` $NOSE_OPTS -A "$NOSE_SELECTION_OP($NOSE_SELECTION)" --with-doctest --with-cov --cover-package datalad --logging-level=INFO $TESTS_TO_PERFORM
...............
........................................................S........S......................................................
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received
The build has been terminated
```

so saga continues -- we can't upload recent annex to neurodebian until this issue again is mitigated in tests (skips), fixed up (if we do not start ssh socket where we could/should) or properly troubleshooted. heh heh

meanwhile I will build/upload bleeding edge git-annex release to -devel of neurodebian

yarikoptic:
maint branch testing.
From: https://travis-ci.org/github/datalad/datalad/jobs/684702932 
```
datalad.distribution.tests.test_install.test_install_subds_from_another_remote ... .: clone2(?) [ssh://localhost/tmp/datalad_temp_test_install_subds_from_another_remoteoaxxo88n/clone2 (git)]
```
From: https://travis-ci.org/github/datalad/datalad/jobs/684702934
```
datalad.distribution.tests.test_install.test_install_subds_with_space ... 
```

yarikoptic:
a pointer to prior troubleshooting which ran into `enableremote`: https://github.com/datalad/datalad/pull/4116#issuecomment-582791175

yarikoptic:
[...]

PR #4502, opened 2020-05-08
TEMP: (maint) - test devel annex 

yarikoptic:
This is a reincarnated #4265 in which we decided to just skip the test leading to hanging, but it seems new places emerged (#4499).  So this is a sweep for travis with -s -v flags for nose to at least see first in what test it hangs


yarikoptic:
```
too old, life is much better now
                       -- optimist
```

Issue #4501, opened 2020-05-08
`push` doesn't immediately emit `copy` results


mih:
Follow-up to #4500. The reason is

```py
        res = GitWitlessRunner(
            cwd=ds.path,
        ).run(
            cmd,
            # TODO report how many in total, and give global progress too
            protocol=AnnexJsonProtocol,
            stdin=file_list)
        for c in ('stdout', 'stderr'):
            if res[c]:
                lgr.debug('Received unexpected %s from `annex copy`: %s',
                          c, res[c])
        for j in res['stdout_json']:
            yield annexjson2result(j, ds, type='file', **res_kwargs)
 ```

It completes the entire `run()` and only then get's to emit the results. Combined with #4500 this can lead to half a day of silence.

Issue #4483, opened 2020-05-05
push: does not respect wanted git-annex wanted expression


yarikoptic:
As discovered during meeting, following script:
```
#!/bin/bash

datalad --version
set -eu
export PS4='> '
set -x
cd "$(mktemp -d ${TMPDIR:-/tmp}/dl-XXXXXXX)"

datalad create src

(
cd src
# now modify subds without saving updated state in superdataset
echo "precious content" > file
datalad save -m "added a file"
datalad create-sibling --annex-wanted nothing  -r -s target ../target
datalad push -d . --to target
)
ls -ld target/file
cat target/file
```

produces

```
> ls -ld target/file
lrwxrwxrwx 1 yoh yoh 110 May  5 10:51 target/file -> .git/annex/objects/63/PP/MD5E-s17--94887fe09c10ca3340603e9a7ec9efdf/MD5E-s17--94887fe09c10ca3340603e9a7ec9efdf
> cat target/file
precious content
```

so "nothing" was not respected

yarikoptic:
FWIW FTR, `git annex copy --auto PATHs` just restricts `--auto` operation to keys found under `PATHs` which would be in agreement with `datalad push` description for having `PATH`s specified.

yarikoptic:
note: make sure to add `--auto` only if wanted is set, since otherwise it would not cause any transfer.

kyleam:
> note: make sure to add `--auto` only if wanted is set

numcopies configuration is also considered.

PR #4503, opened 2020-05-08
ENH: Overall progressbar for AnnexJsonProtocol

mih:
Until now it would only create individual progress bars for items that
git-annex would report progress about. This change adds an additional
global progress bar that counts the number of results, if a total number
of expected results was provided.

This fixes gh-4501

Or rather, it fixes the underlying desire to see something happen, it
won't change how and when results are emitted.

### DIFF
--- a/datalad/core/distributed/push.py
+++ b/datalad/core/distributed/push.py
@@ -719,6 +719,7 @@ def _push_data(ds, target, content, force, jobs, res_kwargs):
     # XXX must not be a SpooledTemporaryFile -- dunno why, but doesn't work
     # otherwise
     with TemporaryFile() as file_list:
+        nfiles = 0
         for c in to_transfer:
             if not c.get('has_content', False):
                 # warned about above, now just skip
@@ -726,9 +727,16 @@ def _push_data(ds, target, content, force, jobs, res_kwargs):
             file_list.write(
                 bytes(Path(c['path']).relative_to(ds.pathobj)))
             file_list.write(b'\0')
+            nfiles += 1
 
         # rewind stdin buffer
         file_list.seek(0)
+
+        # tailor the progress protocol with the total number of files
+        # to be transferred
+        class TailoredPushAnnexJsonProtocol(AnnexJsonProtocol):
+            total_res_count = nfiles
+
         # and go
         # TODO try-except and yield what was captured before the crash
         #res = GitWitlessRunner(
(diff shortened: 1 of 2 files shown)


mih:
Thx @kyleam

yarikoptic:
Thanks @kyleam for the script!

I know that some progress indication is better than no progress indication, but `get`, `publish` provide progress in amount of data transferred. Here it is just a number of files which could be very jumpy and not indicative of actual ETA if sizes of files have inconsistent variance in size... it seems that `push` doesn't use `AnnexRepo.copy_to` (used e.g. by `publish`) which would have done progress indication AFAIK relaying progress from git-annex -- was some problem solved by not using this already existing interface in `push`?

Also I don't get it why `push` here pushes any data -- there is no `annex wanted` set for the remote, no paths specified... shouldn't it just do `git push` and not transfer any data?

mih:
> I know that some progress indication is better than no progress indication, but `get`, `publish` provide progress in amount of data transferred.

The same is happening here, you likely don't see it, because you are moving data locally, and it is too quick. Push to an SSH server, and you'll see that type of progress.

> Here it is just a number of files which could be very jumpy and not indicative of actual ETA if sizes of files have inconsistent variance in size...

That is true. A smoother progress reporting would require to re-stat all files that git-annex reports on to increment a byte counter properly. I personally consider that a worthwhile addition (for somebody else). As stated earlier, I usually don't see progress reports at all, because I have to parallelize across datasets.

> it seems that `push` doesn't use `AnnexRepo.copy_to` (used e.g. by `publish`) which would have done progress indication AFAIK relaying progress from git-annex -- was some problem solved by not using this already existing interface in `push`?

I don't use it, because it is duplicating the collection of information that I already have, and it is too inflexible (and uses the `normalize_paths` decorator, which is a red flag for me). Progress reporting in `push` is available regardless.

> Also I don't get it why `push` here pushes any data -- there is no `annex wanted` set for the remote, no paths specified... shouldn't it just do `git push` and not transfer any data?

I think this should be a separate issue.

mih:
I can now confirm that the progress bar does its job in the case where a particular special remote implementation provides no within-file progress reporting at all.

kyleam:
> Also I don't get it why `push` here pushes any data -- there is no `annex wanted` set for the remote, no paths specified... shouldn't it just do `git push` and not transfer any data?

Note the call to `annex-untrust`.  As I mentioned over at gh-4483, `--auto` also considers the numcopies configuration.

mih:
Ok, in the interest of progress, let's merge this and approve upon in in the future.

PR #4513, opened 2020-05-12
(Re)Introduce typing

yarikoptic:
This is a reincarnated #4447 with only type checking and other changes already merged into `master`.

If we agree that it should be a "centralized" effort, that we do it for all code we change, then it might be worth it.

Should not be merged before 0.13 release

### DIFF
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -27,6 +27,8 @@ from collections import (
     OrderedDict,
 )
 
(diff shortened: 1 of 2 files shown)


mih:
If this is not intended to be merged, can it be a draft PR, please. I keep opening it and checking its status.

yarikoptic:
in case we decide to go forward we know where to find it

Issue #4493, opened 2020-05-07
Deprecate `publish` command


mih:
Documentation of `publish` must be adjusted to announce the plan to remove `publish in the future (https://github.com/datalad/datalad/pull/4453#issuecomment-625379803 https://github.com/datalad/datalad/issues/4461#issuecomment-625378805). Removal could happen in 0.14 (at the earliest).

PR #4515, opened 2020-05-13
DOC: Announce deprecation of `publish`

mih:
Fixes gh-4493

### DIFF
--- a/datalad/distribution/publish.py
+++ b/datalad/distribution/publish.py
@@ -573,7 +573,8 @@ class Publish(Interface):
 
     ..note::
       This command is deprecated. It will be removed from DataLad eventually,
-      but no earlier than with the 0.14 release. Use the `push` command instead.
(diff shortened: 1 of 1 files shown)


datalad/distribution/publish.py
```diff
@@ -571,6 +571,10 @@ def _get_remote_diff(repo, current_commit, remote, remote_branch_name):
 class Publish(Interface):
     """Publish a dataset to a known :term:`sibling`.
 
+    ..note::
+      This command is deprecated. It will be removed from DataLad eventually,
+      but no earlier than with the 0.14 release. Use the `push` command instead.
```
yarikoptic, line 576:
```suggestion
      but no earlier than with the 0.14 release. Use the `push` and `create-sibling` 
      (if you used `--missing` flag) commands instead.
```

mih:
Thx @yarikoptic

Issue #4506, opened 2020-05-10
`push` fails too late on unknown target remote


mih:
`datalad push --to absurd4982438r`

Will only fail after the full `diff` routine is performed. Given that it is doomed to fail, we should fail immediately.

PR #4517, opened 2020-05-13
BF: Fail early on push to unknown target.

mih:
```
% datalad push --to mike
[ERROR  ] Unknown push target 'mike'. No targets configured in dataset. [push.py:__call__:200] (ValueError)
% datalad push --to mike
[ERROR  ] Unknown push target 'mike'. Known targets: 'localtmp'. [push.py:__call__:200] (ValueError)
```

Fixes gh-4506

### DIFF
--- a/datalad/core/distributed/push.py
+++ b/datalad/core/distributed/push.py
@@ -188,16 +188,31 @@ class Push(Interface):
             dataset, check_installed=True, purpose='pushing')
         ds_repo = ds.repo
 
-        if to and to not in ds_repo.get_remotes(
-                exclude_special_remotes=False):
+        res_kwargs = dict(
+            action='publish',
+            refds=ds.path,
+            logger=lgr,
+        )
+
+        get_remote_kwargs = {'exclude_special_remotes': False} \
+            if isinstance(ds_repo, AnnexRepo) else {}
+        if to and to not in ds_repo.get_remotes(**get_remote_kwargs):
             # get again for proper error:
-            sr = ds_repo.get_remotes(exclude_special_remotes=False)
-            raise ValueError(
-                "Unknown push target '{}'. {}".format(
+            sr = ds_repo.get_remotes(**get_remote_kwargs)
+            # yield an error result instead of raising a ValueError,
+            # to enable the use case of pushing to a target that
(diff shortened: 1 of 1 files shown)


datalad/core/distributed/push.py
```diff
@@ -188,6 +188,16 @@ def __call__(
...
+        if to and to not in ds_repo.get_remotes(
+                exclude_special_remotes=False):
+            # get again for proper error:
+            sr = ds_repo.get_remotes(exclude_special_remotes=False)
+            raise ValueError(
+                "Unknown push target '{}'. {}".format(
+                    to,
+                    'Known targets: {}.'.format(', '.join(repr(s) for s in sr))
+                    if sr
+                    else 'No targets configured in dataset.'))
```
kyleam, line 200: Failing earlier is of course good, but should this emit a `{"status": "error", ....}` result, as it did before?

mih: It would be nice to have a rule set to answer that question. I feel like I meander all the time. Here are the two thoughts that I use ATM:

- it happens on the dataset `push` is call on. A broken request, nothing can, was, or should be done: `ValueError`, because no action was performed 
- in the recursive case, a subdataset may not have a sibling defined, but publishing the dataset was attempted, and it failed: `error` result, because need to be able to continue (if desired)

kyleam: > It would be nice to have a rule set to answer that question. I feel like I meander all the time

Yeah, I agree that'd be nice.  It don't think I have anything like a clear rule set in my head, though I'd guess I lean towards results when I'm unsure (though looking at code might show that's not actually the case :).

> Here are the two thoughts that I use ATM: [...]

Thanks, that's helpful.  Even with your edited "What would not be [...]" point, I think raising a `ValueError` here does seem sensible then.

mih:
Travis failure is failure to submit coverage.

datalad/core/distributed/push.py
mih: It does, but you convinced me otherwise before ;-) The ability to force `push` through this kind of error could be useful. I'll keep it with an error result.

PR #4528, opened 2020-05-14
RM: remove 3rdparty_analysis_workflow.sh and testing of it

yarikoptic:
With all the greatness of the handbook this single example is probably
no longer pertinent.  Also today was unlucky day where it triggered
failures on travis (#4512) when tested due seems some intermittent connectivity
issues.  Removing its testing should shave off some 30sec or so from Travis
CI runs


yarikoptic:
Hm, I thought that somehow we effected coverage in http.py, but coverage change to datalad/downloaders/tests/test_http.py really makes no sense to me.. heh heh, I guess could be ignored again

PR #4529, opened 2020-05-14
BF: introduce GitTransportRI - a basic implementation only

### LINKED ISSUE #4489: git-annex's "gcrypt::" prefix in special remote url value causes datalad to "fail"
... since we treat that URL as an ssh url, and try to establish our socket to that host.

Background story: I tried `gcrypt::` prefix since I want to have data compressed on the special remote, and the only way to achieve it ATM AFAIK is simply to enable encryption which would compress keys as a "side effect". But git-annex (I am using 8.20200309+git133-gc29943f40-1~ndall+1 ATM) decides to be overprotective and says

```shell
> git annex initremote github-lfs type=git-lfs url=https://github.com/yarikoptic/test-github-lfs encryption=shared
initremote github-lfs (encryption setup) (encryption key stored in git repository) 
git-annex: Encryption is enabled for this remote, but only the files that git-annex stores on it would be encrypted;  anything that git push sends to it would not be encrypted. Recommend prefixing the url with "gcrypt::" to also encrypt git pushes. (Use --force if you want to use this likely insecure configuration.)
```

if I just use `--force` -- `autoenable` in effect is not working since it does the same check at the later `init` point in the clone... so I decided to try `gcrypt::` prefix.

<details>
<summary>Here is a full script I was cooking up to establish heudiconv-testdata with some comments of exploration</summary> 

```shell
#!/bin/bash
  
# Prepare a datalad dataset for heudiconv testdata, with data from
# 

datalad --version
git annex version

export PS4='> '
set -x
set -eu

n=heudiconv-testdata
d=heudiconv-dcm_physio

[ ! -e "$n" ]

# let's fetch the repo with data

[ -e "$d" ] || git clone -b dcm_physio https://github.com/cbinyu/heudiconv.git "$d"

datalad create -c text2git "$n"
(
cd "$n"
datalad create-sibling-github --name origin "$n"

# Initial push setting master to be pushed there by default
git push -u origin master

# Configure git-annex LFS special remote which should be autoenabled
#
# --force to avoid, since we do not really care about encrypting git itself here
# > git annex initremote github-lfs type=git-lfs url=https://github.com/yarikoptic/test-github-lfs encryption=shared
# initremote github-lfs (encryption setup) (encryption key stored in git repository) 
# git-annex: Encryption is enabled for this remote, but only the files that git-annex stores on it would be encrypted;  anything that git push sends to it would not be encrypted. Recommend prefixing the url with "gcrypt::" to also encrypt git pushes. (Use --force if you want to use this likely insecure configuration.)
# 
# git annex initremote --force github-lfs type=git-lfs url=https://github.com/yarikoptic/"$n" encryption=shared autoenable=true
#
# but if we --force it here, then it would still fight us back on autoenable - causing the whole process to crash
# so we will try to follow recommendatation and do gcrypt:: thing
git annex initremote github-lfs type=git-lfs url=gcrypt::https://github.com/yarikoptic/"$n" encryption=shared autoenable=true

# Tell that LFS should contain all data, unless we explicitly mark it with some
# distribution-restrictions metadata
git annex wanted github-lfs "not metadata=distribution-restrictions=*"

# Configure locally that publishing to origin should also first copy
# the data
git config --add remote.origin.datalad-publish-depends github-lfs

# Copy and save the data
# Let's create some structure
mkdir -p physio
datalad run -m "Injesting original data from @cbinyu" cp -r ../$d/heudiconv/tests/data/samplePhysio physio/sample1

# Publish -- since we setup everything, nothing else to be said ;)
datalad publish
)
```
</details>

which at the end stalls for a while and eventually times out with:

```
[INFO   ] Publishing to configured dependency: 'github-lfs' 
[INFO   ] Publishing <Dataset path=/home/yoh/picts/mris/heudiconv-testdata> data to github-lfs 
[INFO   ] Publishing <Dataset path=/home/yoh/picts/mris/heudiconv-testdata> to github-lfs 
[WARNING] Remote github-lfs has no configured refspec 


ssh: connect to host gcrypt port 22: Connection timed out
ConnectionOpenFailedError: command '['ssh', '-fN', '-o', 'ControlMaster=auto', '-o', 'ControlPersist=15m', '-o', 'ControlPath=/home/yoh/.cache/datalad/sockets/e4ebca10', 'gcrypt']' failed with exitcode 255
Failed to open SSH connection (could not start ControlMaster process)
bash prep-heudiconv-testdata  7.60s user 2.54s system 20% cpu 50.585 total
```

and when I looked into `ps` hierarchy I saw:
```
yoh      3630715  0.0  0.0   6728  2460 pts/46   S+   17:20   0:00  |   |       \_ bash prep-heudiconv-testdata
yoh      3631627  1.1  0.1 111064 33072 pts/46   S+   17:20   0:00  |   |           \_ /home/yoh/proj/datalad/datalad-maint/venvs/dev3.8/bin/python /home/yoh/proj/datalad/datalad-maint/venvs/dev3.8/bin/datalad publish
yoh      3632278  0.0  0.0  11328  6056 pts/46   S+   17:20   0:00  |   |               \_ ssh -fN -o ControlMaster=auto -o ControlPersist=15m -o ControlPath=/home/yoh/.cache/datalad/sockets/e4ebca10 gcrypt
```
which explained why it was "waiting".
Clearly our RI is not aware of such "url" form:
```
$> python -c 'from datalad.support.network import RI; print(repr(RI("gcrypt::https://github.com/yarikoptic/heudiconv-testdata")))'
SSHRI(hostname='gcrypt', path=':https://github.com/yarikoptic/heudiconv-testdata')
```
although standard `urlparse` is ok with it placing everything after first `:` into path component:

```shell
$> python -c 'from urllib.parse import urlparse; print(repr(urlparse("gcrypt::https://github.com/yarikoptic/heudiconv-testdata")))'
ParseResult(scheme='gcrypt', netloc='', path=':https://github.com/yarikoptic/heudiconv-testdata', params='', query='', fragment='')
```

Looking at https://en.wikipedia.org/wiki/URL I think it might be all legit (didn't check full grammar).  Not sure why joey didn't just go for `+` composition (like we now with `ria+URL`) so `scheme` remains "legit" without this confusing path with `:` etc.  Moreover,  IIRC `::` is also used by rsync to signal using rsync protocol, e.g.
```
$> rsync neuro.debian.net::neurodebian
drwxr-sr-x             78 2014/02/03 21:59:37 .
-rw-r--r--             11 2020/05/06 17:10:01 .timestamp
...
```
so if we were to auto-detect that one -- we would have had even more fun.

Since overall it seems to be a "legitish" URL, I am afraid we would need to do something about it on datalad side, but I am not yet sure what should be the most sensible generic way (we might need a new type of an RI, like AnnexRI which would carry this prefix + original URL in `.url`).  I just wish (didn't file any issue yet for @joeyh) that git-annex was not pretty much twisting my arm to use this `gcrypt::` prefix which I do not really care about, or had it as a special remote option instead of wrapping it into a url.

yarikoptic:
this should be sufficient to address https://github.com/datalad/datalad/issues/4489
to avoid treating such RIs as ssh addresses.  Current implementation of
GitTransportRI simply splits apart transport and underlying RI attributes.
Since directed toward maint I have refrained from doing any code refactoring, such as
in RI().__init__ and in the tests check() to avoid using "ri" (and other) arguments
which might collide with fields provided in **kwargs.  May be in some next sweep
those all could become prefixed (or suffixed) with  _ to avoid such collisions.

We could also make this GitTransportRI immediately turn underlying RI into an
instance (ATM it is just a string), and also expose .path as all other RIs
apparently do.  But again, for the sake of the quick fix for now to not cause
hard to decypher errors, I decided to postpone that.

Note: I have not attempted to actually use it in some sample case fully since
e.g. our create-sibling-github has no clue about transport::, would need to
override url it provides for the remote and may be more...

Closes #4489

### DIFF
--- a/datalad/support/network.py
+++ b/datalad/support/network.py
@@ -829,7 +829,7 @@ class GitTransportRI(RI, RegexBasedURLMixin):
     # Due to poor design, `ri` argument already present in various
     # places intermixed with **kwargs treatment. So we will use RI
     # here instead of ri.
-    _REGEX = re.compile(r'(?P<transport>[^:/@]+)::(?P<RI>.*)$')
+    _REGEX = re.compile(r'(?P<transport>[A-Za-z0-9][A-Za-z0-9+.-]*)::(?P<RI>.*)$')
 
     def as_str(self):
         return '{self.transport}::{self.RI}'.format(self=self)
--- a/datalad/support/network.py
+++ b/datalad/support/network.py
@@ -320,7 +320,8 @@ def _guess_ri_cls(ri):
         'url': URL,
         'ssh':  SSHRI,
         'file': PathRI,
-        'datalad': DataLadRI
+        'datalad': DataLadRI,
+        'git-transport': GitTransportRI,
     }
     if isinstance(ri, PurePath):
         lgr.log(5, "Detected file ri")
@@ -339,8 +340,12 @@ def _guess_ri_cls(ri):
     # file:///path should stay file:
     if fields['scheme'] and fields['scheme'] not in {'file'} \
             and not fields['hostname']:
+        # transport::URL-or-path
+        if fields['path'].startswith(':'):  # there was ::
+            lgr.log(5, "Assuming git transport style ri and returning")
+            type_ = 'git-transport'
         # dl+archive:... or just for ssh   hostname:path/p1
-        if '+' not in fields['scheme']:
+        elif '+' not in fields['scheme']:
             type_ = 'ssh'
             lgr.log(5, "Assuming ssh style ri, adjusted: %s" % (fields,))
 
@@ -810,6 +815,24 @@ class DataLadRI(RI, RegexBasedURLMixin):
         if self.remote:
             raise NotImplementedError("not supported ATM to reference additional remotes")
         return "{}{}".format(consts.DATASETS_TOPURL, urlquote(self.path))
+
+
+class GitTransportRI(RI, RegexBasedURLMixin):
+    """RI for some other RI with git transport prefix"""
+
+    # TODO: check how crticial to "inherit" RI._FIELDS asking to provide path
+    _FIELDS = RI._FIELDS + (
+        'transport',
+        'RI',
+    )
+
+    # Due to poor design, `ri` argument already present in various
+    # places intermixed with **kwargs treatment. So we will use RI
+    # here instead of ri.
+    _REGEX = re.compile(r'(?P<transport>[^:/@]+)::(?P<RI>.*)$')
+
+    def as_str(self):
+        return '{self.transport}::{self.RI}'.format(self=self)
 
 
 def _split_colon(s, maxsplit=1):
--- a/datalad/support/tests/test_network.py
+++ b/datalad/support/tests/test_network.py
@@ -41,23 +41,24 @@ from datalad.tests.utils import (
 )
 
 from datalad.support.network import (
-    same_website,
-    dlurljoin,
-    get_tld,
-    get_url_straight_filename,
-    get_response_disposition_filename,
-    parse_url_opts,
+    DataLadRI,
+    GitTransportRI,
+    PathRI,
     RI,
     SSHRI,
-    PathRI,
-    DataLadRI,
     URL,
     _split_colon,
-    is_url,
-    is_datalad_compat_ri,
+    dlurljoin,
     get_local_file_url,
+    get_response_disposition_filename,
+    get_tld,
+    get_url_straight_filename,
+    is_datalad_compat_ri,
     is_ssh,
+    is_url,
     iso8601_to_epoch,
+    parse_url_opts,
+    same_website,
 )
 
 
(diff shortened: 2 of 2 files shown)


datalad/support/network.py
```diff
@@ -812,6 +817,24 @@ def as_git_url(self):
...
+    # TODO: check how crticial to "inherit" RI._FIELDS asking to provide path
+    _FIELDS = RI._FIELDS + (
+        'transport',
+        'RI',
+    )
+
+    # Due to poor design, `ri` argument already present in various
+    # places intermixed with **kwargs treatment. So we will use RI
+    # here instead of ri.
+    _REGEX = re.compile(r'(?P<transport>[^:/@]+)::(?P<RI>.*)$')
```
kyleam, line 832: I was curious what characters Git permits for transport.  As far as I can tell, it goes through `is_urlschemechar`, which carries this comment:

```
/*
 * The set of valid URL schemes, as per STD66 (RFC3986) is
 * '[A-Za-z][A-Za-z0-9+.-]*'. But use slightly looser check
 * of '[A-Za-z0-9][A-Za-z0-9+.-]*' because earlier version
 * of check used '[A-Za-z0-9]+' so not to break any remote
 * helpers.
 */
```

So, I think we could make this regexp stricter, though I doubt it matters much in practice.

yarikoptic: Thanks for digging it out, I looked only into a manpage -- I will adjust the regex to be `[A-Za-z0-9][A-Za-z0-9+.-]*`.

yarikoptic:
```suggestion
    _REGEX = re.compile(r'(?P<transport>[A-Za-z0-9][A-Za-z0-9+.-]*)::(?P<RI>.*)$')
```

PR #4540, opened 2020-05-15
BF: if content comes as bytes, ensure that our regex is bytes as well

yarikoptic:


yarikoptic:
awesome analysis @kyleam -- thanks!   I think that is exactly what triggered this one as well as other ones I was about to submit a PR for.  I think I will just join two PRs into a new one, with separate commits though (the other ones are fixes to tests), with description in a commit message about the switch from string to bytes you pointed out to.

cmadjar:
@yarikoptic This fixed the issue I was having. Thank you so much!!

PR #4544, opened 2020-05-16
ENH: Avoid needless calls with diff(fr=None, annex=!None)

mih:
`diff` treats `fr=None` as a diff-from-scratch. However, when recursing
into subdatasets it breaks this paradigm by translating this into
`fr=PRE_INIT_COMMIT_SHA`. This can lead to a needless call to
`AnnexRepo.get_content_annexinfo()`. This changes rectifies that by
using `None` also for subdatasets.

Moreover, `push` uses `diff(fr=since)`, which can also be `None`. This
change also avoid an equivalent needless call for a top-level dataset.

### DIFF
--- a/datalad/core/local/diff.py
+++ b/datalad/core/local/diff.py
@@ -322,14 +322,14 @@ def _diff_ds(ds, fr, to, constant_refs, recursion_level, origpaths, untracked,
         return
 
     if annexinfo and hasattr(repo, 'get_content_annexinfo'):
-        # this will ammend `status`
+        # this will ammend `diff_state`
         repo.get_content_annexinfo(
             paths=paths.keys() if paths is not None else paths,
             init=diff_state,
             eval_availability=annexinfo in ('availability', 'all'),
             ref=to)
         # if `fr` is None, we compare against a preinit state, and
-        # nothing needs to be done
+        # a get_content_annexinfo on that state doesn't get us anything new
         if fr and fr != to:
             repo.get_content_annexinfo(
                 paths=paths.keys() if paths is not None else paths,
--- a/datalad/core/local/diff.py
+++ b/datalad/core/local/diff.py
@@ -38,7 +38,6 @@ from datalad.support.constraints import (
     EnsureStr,
 )
 from datalad.support.param import Parameter
-from datalad.consts import PRE_INIT_COMMIT_SHA
 
 from datalad.core.local.status import (
     Status,
@@ -329,7 +328,9 @@ def _diff_ds(ds, fr, to, constant_refs, recursion_level, origpaths, untracked,
             init=diff_state,
             eval_availability=annexinfo in ('availability', 'all'),
             ref=to)
-        if fr != to:
+        # if `fr` is None, we compare against a preinit state, and
+        # nothing needs to be done
+        if fr and fr != to:
             repo.get_content_annexinfo(
                 paths=paths.keys() if paths is not None else paths,
                 init=diff_state,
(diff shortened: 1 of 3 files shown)


datalad/core/local/diff.py
```diff
@@ -329,7 +328,9 @@ def _diff_ds(ds, fr, to, constant_refs, recursion_level, origpaths, untracked,
             init=diff_state,
             eval_availability=annexinfo in ('availability', 'all'),
             ref=to)
-        if fr != to:
+        # if `fr` is None, we compare against a preinit state, and
+        # nothing needs to be done
```
bpoldrack, line 332: Might be worth being a bit more explicit why "nothing needs to be done" (we just got all we need to know, since current state of affairs *is* the diff to preinit). Took me a while to realize - `diff` is just not the easiest thing to have a quick look at.
That is if I'm not mistaken, of course. If I am wrong then it's even more important ;-)

mih: I am not sure I understand. "preinit" is before the repository existed. That state is "nothing" and it is identical across all datasets. So in order to figure everything out about this state, nothing needs to be done. I don't know how to change that comment to address your question. Feel free to push a better comment.

yarikoptic:
Sorry for being anal and stating the obvious: without a test checking for the desired behavior, whatever is enhanced can get unenhanced in the next round of RFing or BFing and we would not even mention.

mih:
> Sorry for being anal and stating the obvious: without a test checking for the desired behavior, whatever is enhanced can get unenhanced in the next round of RFing or BFing and we would not even mention.

That is a universal truth. I nevertheless see no point in testing (with unittests) that a particular change still improves the performance. I don't even know how to do that. But I am sure that I have a bunch of issue to work on before I want to learn about it ;-)

yarikoptic:
>  I don't even know how to do that

If you point me to a test which exercises this specific case, I can try: mock diff and see that it isn't called with parameters we expect it to not be called with

mih:
I cannot follow. This code change has the only purpose to achieve the *identical* result, *faster*. So if you want to test the effectiveness of the change, you need to test performance difference.

yarikoptic:
> I cannot follow. This code change has the only purpose to achieve the _identical_ result, _faster_. So if you want to test the effectiveness of the change, you need to test performance difference.

Title of the PR says "ENH: Avoid needless calls with diff(fr=None, annex=!None)", so by "avoid" I assumed the change would cause some `diff` invocations should no longer happen whenever they were happening before.  And that is what I thought to test.

kyleam:
> Title of the PR says "ENH: Avoid needless calls with diff(fr=None, annex=!None)", so by "avoid" I assumed the change would cause some `diff` invocations should no longer happen whenever they were happening before.

A needless call to `get_content_annexinfo` is avoided.  So, you could mock that and assert that it's called N times rather than N+1.  I personally don't like the idea of testing these internal details as a proxy for performance.  However, you seem to have strong preference here, so I'd say let's add the test and then we can drop it later if it becomes a headache.

kyleam:
> However, you seem to have strong preference here, so I'd say let's add the test and then we can drop it later if it becomes a headache.

I've pushed a test that I hope covers what you had in mind.

mih:
If anyhow possible, please make it a standalone test. We already know that push isn't yet doing what it should. There is near certainty that internals will change. We also know that performance isn't good. Whatever this test will achieve, it is not likely that it will last till 0.14. It would be good to not have to disentangle it from an unrelated test.

kyleam:
@mih A few minutes ago I pushed a test in 6fbf52b.  It's small and separate, so I don't think it should be too problematic.

mih:
I saw it just now. Thx @kyleam that works for me.

PR #4545, opened 2020-05-16
ENH: Delay query of `get_content_annex_info()` in `push()`

mih:
Previously this was done in the initial `diff()`. This is often not
needed (e.g. `--force no-datatransfer`), and it is also too expensive,
because diff will obtain annex info twice (for 'from' and 'to' state of
the diff, while `push()` only cares about the 'to' state).

Now it is done only once, and at the latest possible stage.

The change includes a TODO note with further optimization possibilities
that need further investigation.

### DIFF
--- a/datalad/core/distributed/push.py
+++ b/datalad/core/distributed/push.py
@@ -286,8 +286,11 @@ def _datasets_since_(dataset, since, paths, recursive, recursion_limit):
             path=paths,
             # we need to know what is around locally to be able
             # to report something that should have been pushed
-            # but could not, because we don't have a copy
-            annex='availability',
+            # but could not, because we don't have a copy.
+            # however, getting this info here is needlessly
+            # expensive, we will do it at the latest possible stage
+            # in _push_data()
+            annex=None,
             recursive=recursive,
             recursion_limit=recursion_limit,
             # make it as fast as possible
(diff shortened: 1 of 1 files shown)


datalad/core/distributed/push.py
```diff
@@ -691,11 +687,34 @@ def _push_data(ds, target, content, force, jobs, res_kwargs):
...
+    # has in store
+    ds_repo = ds.repo
+    # paths must be recoded to a dataset REPO root (in case of a symlinked
+    # location
+    annex_info_init = \
+        {ds_repo.pathobj / Path(c['path']).relative_to(ds.pathobj): c
+         for c in content} if ds.pathobj != ds_repo.pathobj else \
+        {Path(c['path']): c for c in content}
+    content = ds.repo.get_content_annexinfo(
+        # paths are taken from `content`
```
bpoldrack, line 701: Shouldn't that read "paths are taken from `annex_info_init`"? A bit confusing since `content` is what we assign the result to and it's not an input argument.
Note, that I take this as a comment on `paths=None`. If that is meant to be a remark on what we need the result for it should go outside the `get_content_annexinfo` call, I think.

mih: `annex_info_init` it built from `content`, I pushed a change to clarify that. Thx.

mih:
I have canceled all tests triggered by the change in comment wording

kyleam:
Oh, perhaps my commit on top broke the crawler run.  I'll retrigger it, but if it still fails, I will drop the commit.

https://github.com/datalad/datalad/pull/4545/checks?check_run_id=685618462#step:8:197

kyleam:
> I'll retrigger it, but if it still fails, I will drop the commit.

Well, I don't see how that change would cause the failure, but github workflows isn't getting passed the setup when I try to retrigger it.  That change wasn't important, so I'll drop it and merge, canceling the tests since they passed 86fa157)

Issue #4508, opened 2020-05-10
`push` warns to early and too dramatically


mih:
Case:

```
datalad push --to inm7 -r
...
[WARNING] Slated for transport, but no content present [copy(....nii.gz)] 
...
```
The local dataset doesn't have the content, but the remote has it already. There are now three angles:

#### UX

All these files should be tested, whether the remote has the content already, and if so, the logging should happen on DEBUG instead.

#### Performance

No further testing should be done. No path was given, if content isn't there locally, it can never be copied. That is "obvious" and not worth a warning even.

#### Trust

I might be under the impression that after this command, the remote is guaranteed to have a content that is *tracked* (not just available). Without a warning, I can easily maintain this false impression.

mih:
https://github.com/datalad/datalad/issues/4530 is another case where this is massively annoying. 60k+ warning log messages after

`datalad push --to origin`

kyleam:
> There are now three angles:

Reading over these, I have a hard time landing on any angle other than the second.

mih:
Thx for the feedback @kyleam ! This is also my current attitude.

PR #4546, opened 2020-05-17
ENH: Give insight in push progress with a step-based progress bar

mih:
It makes it more obvious what is happening, and also exposes the
processing of publication targets.

This could also be a model for other commands, where presently
`lgr.info()` calls serve this purpose -- which annoyingly interfere
with progressbars and result rendering.

In combination with #4544 and #4545 it looks like this:

image

PR #4547, opened 2020-05-17
ENH: Only evaluate and warn about content availability with explicit paths

mih:
This implements the conclusion in the issue discussion of gh-4508 and thereby fixes gh-4508.

This PR sits on top of #4545 (essential) and #4546 (to limit unavoidable conflicts).

In all cases with no `path=` argument given to `push()` this change avoids an expensive content availability check that can make the difference between minutes and hours of processing large N datasets.

The relevant commit is https://github.com/datalad/datalad/commit/ef04b1e50d8418eb001329d19c578ef2040e9df3

### DIFF
--- a/datalad/core/distributed/push.py
+++ b/datalad/core/distributed/push.py
@@ -246,7 +246,8 @@ class Push(Interface):
             lgr.debug('Attempt push of Dataset at %s', dspath)
             pbars = {}
             yield from _push(
-                dspath, dsrecords, to, force, jobs, res_kwargs.copy(), pbars)
+                dspath, dsrecords, to, force, jobs, res_kwargs.copy(), pbars,
+                got_path_arg=True if path else False)
             # take down progress bars for this dataset
             for i, ds in pbars.items():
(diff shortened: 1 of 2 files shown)


yarikoptic:
IMHO it is just a bandaid which conflates two semantics as I outlined in https://github.com/datalad/datalad/issues/4551#issuecomment-630404940

mih:
The main purpose of the PR is performance. On sizable repos it can shave of an hour of pre-upload runtime. It would be good to stick that bandaid somewhere, until a superior solution is implemented.

yarikoptic:
Sounds good. Unfortunately GitHub mumbles something about conflicts now

kyleam:
> Unfortunately GitHub mumbles something about conflicts now

I've merged master (conflict was handled by automatic resolution) and pushed .

kyleam:
> I've merged master (conflict was handled by automatic resolution) and pushed .

CrippledFS failed one way and, then on rerun, another.  It passed before the merge (00c78b1) to resolve a conflict that involved only a comment.  So I'll assume it's unrelated and merge this.

PR #4552, opened 2020-05-19
BF: use '' as default (not None) while checking for a key not to be a test key

yarikoptic:
I just saw it in https://github.com/datalad/datalad-extensions/runs/665491656?check_suite_focus=true
which I believe failed for other reasons (but may be not), that there was a traceback
to this line.  And indeed above we might ask for a 'key' or a 'file', so hypothetically
there might be no 'key' field.  I have not really analyzed situation on how we have not
hit it before and how it really emerges in that failure while testing against
bleeding edge git annex

### DIFF
--- a/datalad/customremotes/tests/test_datalad.py
+++ b/datalad/customremotes/tests/test_datalad.py
@@ -31,18 +31,30 @@ def check_basic_scenario(url, d):
     # TODO skip if no boto or no credentials
     get_test_providers(url) # so to skip if unknown creds
 
-    # git-annex got a fix where it stopped replacing - in the middle of the filename
-    filename = '3versions%sallversioned.txt' % (
-        '_' if external_versions['cmd:annex'] < '8.20200501+git53-gcabbc91b1' else '-'
-    )
-
     # Let's try to add some file which we should have access to
     with swallow_outputs() as cmo:
         annex.add_urls([url])
         annex.commit("committing")
-        whereis1 = annex.whereis(filename, output='full')
-        eq_(len(whereis1), 2)  # here and datalad
-        annex.drop(filename)
+
+    # git-annex got a fix where it stopped replacing - in the middle of the filename
+    # Let's cater to the developers who might have some intermediate version and not
+    # easy to compare -- we will just check that only one file there is an that it
+    # matches what we expect when outside of the development versions range:
+    filenames = glob.glob(op.join(d, '3versions[-_]allversioned.txt'))
+    assert_equal(len(filenames), 1)
+    filename = op.basename(filenames[0])
+    if external_versions['cmd:annex'] < '8.20200501':
+        assert_in('_', filename)
+    # Date after the fix in 8.20200501-53-gcabbc91b1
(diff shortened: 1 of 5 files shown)


yarikoptic:
oh, on the most recent run https://github.com/datalad/datalad-extensions/runs/687454738?check_suite_focus=true it is the only error which causes the failure!  I will try to check out what is that record now...

kyleam:
I can trigger it locally.  I'll try to bisect it.

yarikoptic:
no need -- I recalled what it is about!

yarikoptic:
I bet (didn't check explicitly) it is due to [8.20200501-55-g39d7e6dd2](https://git.kitenet.net/index.cgi/git-annex.git/commit/?id=39d7e6dd2a28dad3cda5f987f4da3458ff4703ee)
  where git-annex stopped "sanitizing" `-` in the middle of the filenames for urls it downloads from the web.

kyleam:
> I bet (didn't check explicitly) it is due to [8.20200501-55-g39d7e6dd2](https://git.kitenet.net/index.cgi/git-annex.git/commit/?id=39d7e6dd2a28dad3cda5f987f4da3458ff4703ee)

FWIW the bisect pointed to the related cabbc91b1 (addurl, importfeed: Allow '-' in filenames, as long as it's not the first character, 2020-05-11).

yarikoptic:
oh, thanks!  I just pushed, but will rewrite and force push with the correct exact version! ;)

yarikoptic:
done... also "polluted" this PR with more commits stripping away some ancient annex versions checks where version is below the min version we currently support. It all must be green! ;)

datalad/customremotes/tests/test_datalad.py
```diff
@@ -30,14 +31,19 @@ def check_basic_scenario(url, d):
     # TODO skip if no boto or no credentials
     get_test_providers(url) # so to skip if unknown creds
 
+    # git-annex got a fix where it stopped replacing - in the middle of the filename
+    filename = '3versions%sallversioned.txt' % (
+        '_' if external_versions['cmd:annex'] < '8.20200501+git53-gcabbc91b1' else '-'
```
kyleam, line 36: This is fine, though I think it'd be even better to check whether a file with one of the two names exists after the `add_urls` call.  The current approach means this test would fail for me with a local git-annex build after 8.20200501 but before that commit, where cmd:annex would be something like `8.20200502-g[xdigit]+`

kyleam:
>  with more commits stripping away some ancient annex versions checks where version is below the min version we currently support

Thanks.

datalad/customremotes/tests/test_datalad.py
yarikoptic: Heh.. I thought to use glob first... Ok, will adjust later while also checking that expected name is popping out for versions outside of the development range

yarikoptic: Since I do want to have at least this single test to check for expected annex behavior

PR #4553, opened 2020-05-19
RF: throw dedicated RuntimeError with information about question in case of SilentConsoleLog

yarikoptic:
It is hard to impossible to troubleshoot failures of datalad in automated
deployments where we want to ask some question for some reason, but UI simply
does not support that since it is rightfully silent/non-interactive.

See e.g. https://github.com/templateflow/templateflow/issues/50 .

With this PR we would still fail but with a little more informative RuntimeError,
instead of obscure AttributeError stating what was actually asked.  We will not
disclose possible present other fields other than question and title if it was a
question asking for secret information ("hidden").  AFAIK in our use of the dialogues
we do not disclose any sensitive information in title or the question but I could be
proven wrong.

I did position this one on top of maint since might benefit people using released
version in production.  Delaying till 0.13.0 would require them to migrate first to
just be able to troubleshoot

### DIFF
--- a/datalad/ui/dialog.py
+++ b/datalad/ui/dialog.py
@@ -131,7 +131,7 @@ class SilentConsoleLog(ConsoleLog):
             if kwargs_str:
                 msg += " Additional arguments: %s" % kwargs_str
         else:
-            msg += " Additional arguments are not shown due to 'hidden' is set."
+            msg += " Additional arguments are not shown because 'hidden' is set."
         raise RuntimeError(msg)
 
 
--- a/datalad/ui/dialog.py
+++ b/datalad/ui/dialog.py
@@ -27,6 +27,7 @@ import getpass
 # from unittest.mock import patch
 from collections import deque
 from copy import copy
+from itertools import chain
 
 from ..utils import auto_repr
 from ..utils import on_windows
@@ -117,6 +118,21 @@ class SilentConsoleLog(ConsoleLog):
     def get_progressbar(self, *args, **kwargs):
         from .progressbars import SilentProgressBar
         return SilentProgressBar(*args, **kwargs)
+
+    def question(self, text, title=None, **kwargs):
+        msg = "A non-interactive silent UI was asked for a response to a question: %s." % text
+        if title is not None:
+            msg += ' Title: %s.' % title
+        if not kwargs.get('hidden'):
+            kwargs_str = ', '.join(
+                ('%s=%r' % (k, v)
+                for k, v in kwargs.items()
+                if v is not None))
+            if kwargs_str:
+                msg += " Additional arguments: %s" % kwargs_str
+        else:
+            msg += " Additional arguments are not shown due to 'hidden' is set."
+        raise RuntimeError(msg)
 
 
 @auto_repr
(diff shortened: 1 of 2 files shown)


datalad/ui/dialog.py
```diff
@@ -118,6 +119,21 @@ def get_progressbar(self, *args, **kwargs):
...
+            msg += ' Title: %s.' % title
+        if not kwargs.get('hidden'):
+            kwargs_str = ', '.join(
+                ('%s=%r' % (k, v)
+                for k, v in kwargs.items()
+                if v is not None))
+            if kwargs_str:
+                msg += " Additional arguments: %s" % kwargs_str
+        else:
+            msg += " Additional arguments are not shown due to 'hidden' is set."
```
effigies, line 134:
```suggestion
            msg += " Additional arguments are not shown because 'hidden' is set."
```

yarikoptic:
> Also, would newlines be better separators than spaces between question, title and args? It's probably parseable as-is, but I'm wondering if longer questions or more arguments might not get too cluttered.

I don't know why really, but for some reason I am avoiding adding multiline formatting into exception messages... I wonder if there is some reason for such behavior of mine since those messages are indeed just for a user to see and I could have gone "wild" in formatting them to actually make them readable. hm...

yarikoptic:
Probably for such cases I better come up with some dedicated (probably mix-in) exception class which would get args, kwargs args and then render them nicely in its `str`. Then it could be reused for a number of use-cases

effigies:
Well, don't hold this up on my hypothetical aesthetics. If it turns out to be unreadable in practice, that's what the "New pull request" button is for.

yarikoptic:
Oh well, while liberal Germans are vacationing, the anal Russian might be looking for the ultimate perfection ;-)

datalad/ui/dialog.py
```diff
@@ -27,6 +27,7 @@
 # from unittest.mock import patch
 from collections import deque
 from copy import copy
+from itertools import chain
```
kyleam, line 30: This import is used.  Otherwise looks good to me.

kyleam:
> I did position this one on top of maint since might benefit people using released
> version in production. Delaying till 0.13.0 would require them to migrate first to
> just be able to troubleshoot

I just merged this to maint, but it looks like you chose master as the base for the PR.  Sorry I didn't spot it, but based on the above I'm pretty sure you intended to choose maint.  Rather than closing this or trying to change the base now, perhaps let's just wait for maint to hit master, so this shows up as merged.

datalad/ui/dialog.py
kyleam: Pruned after merge.

kyleam:
@effigies:
> Well, don't hold this up on my hypothetical aesthetics. If it turns out to be unreadable in practice, that's what the "New pull request" button is for.

@yarikoptic:
> Oh well, while liberal Germans are vacationing, the anal Russian might be looking for the ultimate perfection ;-)

Oy, I'm sorry, I overlooked that @yarikoptic was intending additional improvements for this series.

effigies:
The enterprising American disrupts the process and makes history.

yarikoptic:
;-) LOL, indeed! so I will just succumb to the progress and will avoid being anal -- I think it is + and + ;)

PR #4555, opened 2020-05-20
BF: Prevent ugly interaction or result and progress reporting

mih:
This is an excerpt from gh-4444 that solely addresses the suboptimal interaction of result reporting and progress reporting that is presently evident in `push()` usage.

The idea is to clear any progressbar(s) before rendering a result with `ui.message()` and afterward refresh the progressbar(s). With this approach results will appear above any active progress bar in the terminal and do not get half-swallowed by a progress update.

https://github.com/datalad/datalad/pull/4444#issuecomment-620285637 outlines an additional improvement that addresses the analog issue for the interaction of log messages and progress reporting, also pointed out by (but unaddressed in) gh-4444. However, that solution can be seen as an incremental step that should not delay the fix contained herein.

@kyleam this patch should address the visual glitches that you have observed.

### DIFF
--- a/datalad/log.py
+++ b/datalad/log.py
@@ -201,6 +201,16 @@ class ProgressHandler(logging.Handler):
 
     def emit(self, record):
         from datalad.ui import ui
+        maint = getattr(record, 'dlm_progress_maint', None)
+        if maint == 'clear':
+            # remove the progress bar
+            for pb in self.pbars.values():
+                pb.clear()
+            return
+        elif maint == 'refresh':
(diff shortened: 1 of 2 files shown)


kyleam:
> @kyleam this patch should address the visual glitches that you have observed.

Confirmed.  Thanks!

yarikoptic:
Looks good to me too.
Filed https://github.com/datalad/datalad/issues/4556 someone might to tackle at some point in the future for a more generic solution to the issue.

PR #4561, opened 2020-05-21
ENH: HTTP for ORA special remote

### LINKED ISSUE #4410: ORA remote over HTTP
In https://github.com/datalad/datalad/pull/4203/commits/e5cba2b8a797e31e3591006f611c945cc075aba6 I started implementing a feature for `clone`, that would try to reconfigure an ORA remote with the `ria+` URL we cloned from if autoenabling failed for the URL stored with the special remote. For proof-of-concept I started with HTTP to see that and how we would retrieve the required info from the store. (Commit will be amended soon with the real thing ;-) )

However, while I'm doing the actual implementation and proper test for SSH and FILE URLs, HTTP actually is a special case. A special remote is a local thing, while access to an archive at the remote end over HTTP would require something to run at the store's end for extraction and delivery. ORA special remote would only work over HTTP if it is not using an archive. However, just enabling the git remote is an alternative for consumption only as the structure of annex/objects depends on RIA layout
version. There's currently dirhashmixed and dirhashlower while the repo itself is a bare repo. It can deliver both kinds of annex keys (simultaneously), but it would write dirhashlower regardless of any ria-layout-version.
So, currently we can't enable anything really, since no HTTP is implemented in ORA remote yet and we risk messing up the layout via pushes by enabling the git remote (given annex can write to it at all, of course). We could of course double-check ria-layout-version the same way we do check config and enable git remote whenever layout indicates that we are using dirhashlower, but that's half a solution only.

Conclusion: We probably need a proper HTTP implementation in ORA remote. Meanwhile we can try to just enable the git remote if autoenable failed and we cloned via HTTP in order to retrieve keys not "hidden" in an archive. (How do we double check, that we can't write?) Agree, @mih?
Question remains, whether we want some server command to be ran on the store's end for archive access over HTTP.

bpoldrack:
Implements HTTP support for the ORA special remote (read-only for now).

Closes #4410

In opposition to local and SSH, everything is supposed to happen server-side. ORA sends a request to e location of a key. This location is based on resolved RIA URL to the dataset and assumes dirhashmixed with its `annex/objects` tree. Remapping based on a different layout and archive access needs to be done by the server (That is what PR #4572 is for).

### DIFF
--- a/datalad/core/distributed/clone.py
+++ b/datalad/core/distributed/clone.py
@@ -11,6 +11,7 @@
 
 import logging
 import re
+import requests
 from os.path import expanduser
 from collections import OrderedDict
 from urllib.parse import unquote as urlunquote
@@ -612,23 +613,15 @@ def postclonecfg_ria(ds, props):
         scheme = props['giturl'].split(':', 1)[0]
         if scheme == 'http':
 
-            # Note: Following outcommented code was proof how to check via HTTP.
-            #       However, for now we only want to annexenable origin in case
-            #       of HTTP, not the special remote. Hence it's uuid isn't
-            #       actually required. (see gh-4410)
-
-            # import requests
-            # try:
-            #     response = requests.get("{}{}config".format(
-            #         props['giturl'],
-            #         '/' if not props['giturl'].endswith('/') else '')
-            #     )
-            #     config_content = response.text
-            # except requests.RequestException as e:
-            #     lgr.debug("Failed to get config file from source:\n%s",
-            #               exc_str(e))
-
-            lgr.debug("No ORA reconfiguration for HTTP implemented yet")
+            try:
+                response = requests.get("{}{}config".format(
+                    props['giturl'],
+                    '/' if not props['giturl'].endswith('/') else '')
+                )
+                config_content = response.text
+            except requests.RequestException as e:
+                lgr.debug("Failed to get config file from source:\n%s",
+                          exc_str(e))
 
         elif scheme == 'ssh':
             # TODO: switch the following to proper command abstraction:
@@ -653,12 +646,16 @@ def postclonecfg_ria(ds, props):
                 lgr.debug("Failed to get config file from source: %s",
                           exc_str(e))
         else:
-            lgr.debug("Unknown URL-Scheme in %s. Can currently handle SSH or "
+            lgr.debug("Unknown URL-Scheme in %s. Can handle SSH, HTTP or "
                       "FILE scheme URLs.", props['source'])
 
         # 3. And read it
         org_uuid = None
         if config_content:
+            # TODO: We might be able to spare the saving to a file.
+            #       "git config -f -" is not explicitly documented but happens
+            #       to work and would read from stdin. Make sure we know this
+            #       works for required git versions and on all platforms.
             with make_tempfile(content=config_content) as cfg_file:
                 runner = WitlessRunner(env=GitRunner.get_git_environ_adjusted())
                 try:
@@ -689,7 +686,7 @@ def postclonecfg_ria(ds, props):
                 try:
                     repo.enable_remote(srs[org_uuid]['name'],
                                        options=['url={}'.format(new_url)]
-                                      )
+                                       )
                     lgr.info("Reconfigured %s for %s",
                              srs[org_uuid]['name'], new_url)
                     # update ora_remotes for considering publication dependency
--- a/datalad/distributed/ora_remote.py
+++ b/datalad/distributed/ora_remote.py
@@ -4,7 +4,9 @@ from annexremote import ProtocolError
 
 from pathlib import (
     Path,
+    PurePosixPath
 )
+import requests
 import shutil
 from shlex import quote as sh_quote
 import subprocess
@@ -372,7 +374,8 @@ class SSHRemoteIO(IOBase):
 
     def get(self, src, dst, progress_cb):
 
-        # Note, that as we are in blocking mode, we can't easily fail on the actual get (that is 'cat').
+        # Note, that as we are in blocking mode, we can't easily fail on the
+        # actual get (that is 'cat').
         # Therefore check beforehand.
         if not self.exists(src):
             raise RIARemoteError("annex object {src} does not exist.".format(src=src))
@@ -509,6 +512,52 @@ class SSHRemoteIO(IOBase):
             self._run(cmd, check=True)
         except RemoteCommandFailedError:
             raise RIARemoteError("Could not write to {}".format(str(file_path)))
+
+
+class HTTPRemoteIO(object):
+    # !!!!
+    # This is not actually an IO class like SSHRemoteIO and LocalIO and needs
+    # respective RF'ing of special remote implementation eventually.
+    # We want ORA over HTTP, but with a server side CGI to talk to in order to
+    # reduce the number of requests. Implementing this as such an IO class would
+    # mean to have separate requests for all server side executions, which is
+    # what we do not want. As a consequence RIARemote class implementation needs
+    # to treat HTTP as a special case until refactoring to a design that fits
+    # both approaches.
+
+    # NOTE: For now read-only. Not sure yet whether an IO class is the right
+    # approach.
+
+    def __init__(self, ria_url, id):
+        assert ria_url.startswith("ria+http")
+        self.base_url = ria_url[4:]
+        if self.base_url[-1] == '/':
+            self.base_url = self.base_url[:-1]
+
+        self.base_url += "/" + id[:3] + '/' + id[3:]
+
+    def checkpresent(self, key_path):
+        # Note, that we need the path with hash dirs, since we don't have access
+        # to annexremote.dirhash from within IO classes
+
+        url = self.base_url + "/annex/objects/" + str(key_path)
+        response = requests.head(url)
+        return response.status_code == 200
+
+    def get(self, key_path, filename, progress_cb):
+        # Note, that we need the path with hash dirs, since we don't have access
+        # to annexremote.dirhash from within IO classes
+
+        url = self.base_url + "/annex/objects/" + str(key_path)
+        response = requests.get(url, stream=True)
+
+        with open(filename, 'wb') as dst_file:
+            bytes_received = 0
+            for chunk in response.iter_content(chunk_size=1024,
+                                               decode_unicode=False):
+                dst_file.write(chunk)
+                bytes_received += len(chunk)
+                progress_cb(bytes_received)
 
 
 def handle_errors(func):
@@ -831,7 +880,9 @@ class RIARemote(SpecialRemote):
             if not self.archive_id:
                 # fall back on the UUID for the annex remote
                 self.archive_id = self.annex.getuuid()
-        self.get_store()
+
+        if not self.ria_store_url.startswith("ria+http"):
+            self.get_store()
 
         self.annex.setconfig('archive-id', self.archive_id)
         # make sure, we store the potentially rewritten URL
@@ -839,7 +890,7 @@ class RIARemote(SpecialRemote):
 
     def _local_io(self):
         """Are we doing local operations?"""
-        # let's not make this decision dependent on the existance
+        # let's not make this decision dependent on the existence
         # of a directory the matches the name of the configured
         # store tree base dir. Such a match could be pure
         # coincidence. Instead, let's do remote whenever there
@@ -874,12 +925,16 @@ class RIARemote(SpecialRemote):
             raise RIARemoteError("Remote is treated as read-only. "
                                  "Set 'ora-remote.<name>.force-write=true' to "
                                  "overrule this.")
+        if isinstance(self.io, HTTPRemoteIO):
+            raise RIARemoteError("Write access via HTTP not implemented")
 
     @property
     def io(self):
         if not self._io:
             if self._local_io():
                 self._io = LocalIO()
+            elif self.ria_store_url.startswith("ria+http"):
+                self._io = HTTPRemoteIO(self.ria_store_url, self.archive_id)
             elif self.storage_host:
                 self._io = SSHRemoteIO(self.storage_host)
                 from atexit import register
@@ -896,7 +951,9 @@ class RIARemote(SpecialRemote):
         gitdir = self.annex.getgitdir()
         self.uuid = self.annex.getuuid()
         self._verify_config(gitdir)
-        self.get_store()
+
+        if not isinstance(self.io, HTTPRemoteIO):
+            self.get_store()
 
         # report active special remote configuration
         self.info = {
@@ -942,6 +999,13 @@ class RIARemote(SpecialRemote):
 
     @handle_errors
     def transfer_retrieve(self, key, filename):
+
+        if isinstance(self.io, HTTPRemoteIO):
+            self.io.get(PurePosixPath(self.annex.dirhash(key)) / key / key,
+                        filename,
+                        self.annex.progress)
+            return
+
         dsobj_dir, archive_path, key_path = self._get_obj_location(key)
         abs_key_path = dsobj_dir / key_path
         # sadly we have no idea what type of source gave checkpresent->true
@@ -959,6 +1023,11 @@ class RIARemote(SpecialRemote):
 
     @handle_errors
     def checkpresent(self, key):
+
+        if isinstance(self.io, HTTPRemoteIO):
+            return self.io.checkpresent(
+                PurePosixPath(self.annex.dirhash(key)) / key / key)
+
         dsobj_dir, archive_path, key_path = self._get_obj_location(key)
         abs_key_path = dsobj_dir / key_path
         if self.io.exists(abs_key_path):
@@ -999,6 +1068,13 @@ class RIARemote(SpecialRemote):
 
     @handle_errors
     def whereis(self, key):
+
+        if isinstance(self.io, HTTPRemoteIO):
+            # display the URL for a request
+            # TODO: method of HTTPRemoteIO
+            return self.ria_store_url[4:] + "/annex/objects" + \
+                   self.annex.dirhash(key) + "/" + key + "/" + key
+
         dsobj_dir, archive_path, key_path = self._get_obj_location(key)
         return str(key_path) if self._local_io() \
             else '{}: {}:{}'.format(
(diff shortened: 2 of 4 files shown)


datalad/distributed/ora_remote.py
```diff
@@ -510,6 +513,52 @@ def write_file(self, file_path, content, mode='w'):
...
+    def get(self, key_path, filename, progress_cb):
+        # Note, that we need the path with hash dirs, since we don't have access
+        # to annexremote.dirhash from within IO classes
+
+        url = self.base_url + "/annex/objects/" + str(key_path)
+        response = requests.get(url, stream=True)
+
+        with open(filename, 'wb') as dst_file:
+            bytes_received = 0
+            for chunk in response.iter_content(chunk_size=1024,
```
mih, line 555: How is this `chunk_size` value determined?

mih:
I did a few tests on a medium-complexity dataset (~40k files, ~20GB, so loads of small files). Somewhat beefy server that exhibited little CPU load.
Reconfiguration of the source datasets special remote setup worked for all protocols:

### RIA store, apache server, and clone target were all on the same machine. `get(jobs=1)` :

#### ria+http (no CGI deployed)
- Progress bar reported: ~20MB/s
- `datalad get .  267.26s user 79.93s system 34% cpu 16:39.66 total`

#### ria+ssh
- Progress bar reported: ~80MB/s
- `datalad get .  213.99s user 61.64s system 83% cpu 5:32.08 total`

#### ria+file
- Progress bar reported: ~80MB/s
- `datalad get .  215.33s user 62.15s system 80% cpu 5:45.12 total`

datalad/distributed/ora_remote.py
bpoldrack: Not at all. And the same is true for SSH. Wonder how to determine what's a reasonable size. And I guess we should make it configurable.

mih: Reading 1kb at a time seems very little. It wouldn't even fill up a single block on the file system.

mih:
@bpoldrack would it make sense to split this PR into its logical pieces. It seems that if the client-side should be able to operate without the CGI script, the CGI script could be developed and evaluated separately.

datalad/distributed/ora_remote.py
bpoldrack: Absolutely agree. Any suggestion as to how to determine the size?

mih: ATM none other than to try a few and measure the impact, if there is any.

bpoldrack:
> @bpoldrack would it make sense to split this PR into its logical pieces. It seems that if the client-side should be able to operate without the CGI script, the CGI script could be developed and evaluated separately.

Yes. Special remote and clone adaption for auto-reconfigure can be separated from CGI, if we want to stick to "the client-side should be able to operate without the CGI script".

mih:
I think it makes sense to keep this flexibility. It simplifies deployment requirements quite a bit.

bpoldrack:
>  I think it makes sense to keep this flexibility. It simplifies deployment requirements quite a bit.

I do see the point. It allows for easy deployment in some cases. Complicates matters server-side with respect to performance for archive access and range access to big files, though. Solvable, but a bit trickier. The latter may not matter much yet, the former is more likely to (and ultimately running into the same problem, I think). Via SSH we extract from archive and stream the output to the client while extracting. We can't do that with a single GET request, I think. (I might miss something, though - happy to be corrected).

Just want to make sure, we are aware of what we are running into with this.

mih:
Isn't this solely in the domain of the webserver? The CGI script just writes the data to stdout. How much and how exactly the server reads it and how it sends it out isn't up to the CGI script. Am I wrong?

mih:
@bpoldrack Can you share the apache config you are using to ease my testing efforts? Thx!

bpoldrack:
> @bpoldrack Can you share the apache config you are using to ease my testing efforts? Thx!

```
❱ cat /etc/apache2/conf-available/ora-cgi.conf 
<IfModule mod_alias.c>
    <IfModule mod_cgi.c>
        Define ENABLE_USR_LIB_CGI_BIN
    </IfModule>

    <IfModule mod_cgid.c>
        Define ENABLE_USR_LIB_CGI_BIN
    </IfModule>

    <IfDefine ENABLE_USR_LIB_CGI_BIN>
        SetEnv GIT_PROJECT_ROOT /var/www/html
        SetEnv GIT_HTTP_EXPORT_ALL
        ScriptAliasMatch "(?x)^/(.*/(HEAD | \
            info/refs | \
            objects/(info/[^/]+ | \
                [0-9a-f]{2}/[0-9a-f]{38} | \
                pack/pack-[0-9a-f]{40}\.(pack|idx)) | \
                git-(upload|receive)-pack))$" /usr/lib/git-core/git-http-backend/$1

        ScriptAliasMatch "^(.*)/annex/objects/(.*)" "/usr/lib/cgi-bin/ora_cgi.py"

        <Directory "/usr/lib/git-core">
            Options +ExecCgi -MultiViews +SymLinksIfOwnerMatch
            AllowOverride none
            Order allow,deny
            Allow from all
            Require all granted
        </Directory>
    </IfDefine>
</IfModule>
```

bpoldrack:
> How much and how exactly the server reads it and how it sends it out isn't up to the CGI script. Am I wrong?

I think you are. As far as I'm aware the server is supposed to wait for CGI to finish. Which makes sense, since if it returns non-zero, the server would need to repond with 500 status. Therefore status can't be determined beforehand but it's a *header* field ...

bpoldrack:
For ranged access of non-archived files, it would be nice to just let the webserver do it's thing instead and just redirect to a URI, the server would serve w/o the CGI. But this doesn't easily work either, since with the requests we use, the "natural" URI for this might be the exact same request that triggered the CGI in the first place.

datalad/distributed/ora_remote.py
bpoldrack: Just tried to use `os.stat('.').st_blksize` to set the buffer to "preferred block size". 
Saw ~40MB/s w/o that change and ~140MB/s w/ that change. So - plausible and worthy starting point.

bpoldrack: Suspected potential there (and for SSH at the respective point), but a bit surprised by the effect size ...

mih:
> > How much and how exactly the server reads it and how it sends it out isn't up to the CGI script. Am I wrong?
> 
> I think you are. As far as I'm aware the server is supposed to wait for CGI to finish. Which makes sense, since if it returns non-zero, the server would need to repond with 500 status. Therefore status can't be determined beforehand but it's a _header_ field ...

I somewhat doubt that. It would mean that the largest response can never be more than what RAM is available. Many IP cams have CGI interfaces to stream live video. There is no way that they kick off a new process for every couple of frames.

datalad/distributed/ora_remote.py
bpoldrack: For me that size is 4KB. If I go to 16KB I got another factor 2. Suggests, that the effect actually has little to do with block size (which is kind of a relief, since my estimation, that it shouldn't do much for *that* reason wasn't that stupid after all ;-) ).

bpoldrack: Increasing further keeps getting faster, although the gain per additional buffer size slows down. Difference between 64KB and 80KB seems marginal. With 64KB I'm down to 4.9s for downloading 1.8GB in 32 files from local Apache, while it's been ~66s with current 1KB. (That's `datalad get` not `git-annex-get`)

bpoldrack: So, I'd suggest to make that a config (same for HTTP and SSH or separate?), defaulting to 64KB for now. With that config we can then have easy benchmarks for our setups, giving us a better idea. Agree @mih?

mih: My remark re blocksize wasn't meant to imply causation, just to say that it is very little. Making it configurable sound sensible. Providing protocol-specific tuning seems like overkill at this point.

mih: It seems rather than playing with the chunk size, we can leave it to the professionals: https://stackoverflow.com/a/39217788/1194609

mih:
Can you swap out your CGI script, but use the same setup, and use a script that keep spitting out random numbers and never exists. If you point curl to this script, would it ever download anything?

datalad/distributed/ora_remote.py
bpoldrack: Tried that. In my local setup that is still 10% slower + leaves us with no progress reporting.

bpoldrack:
> Can you swap out your CGI script, but use the same setup, and use a script that keep spitting out random numbers and never exists. If you point curl to this script, would it ever download anything?

It does indeed spit out the CGI's output forever (as does wget). So, you're right in general. Although looking at that with wireshark, there seems to simply be no HTTP response at all (in that sense, I'm right too). It's all TCP only. 

So, still not entirely clear to me how exactly to use that correctly - need to play a bit more. However, that seems to provide the answer I couldn't see. Hence, the GET request is fine as is and that's the important part. Thx!

bpoldrack:
> ria+http (no CGI deployed)
Reconfiguration of special remote failed.

@mih: Is that still failed as "clone didn't do it automatically" or did you try to manually run `git annex enableremote SPECIALREMOTENAME url=ria+http://whereever/that/is` and *that* failed?

mih:
I did not attempt a manual reconfigure yet

bpoldrack:
>  So, still not entirely clear to me how exactly to use that correctly - need to play a bit more.

Did that. Easy. Interesting. I thought that kind of thing would involve either a bunch of requests or a protocol switch. Past time to learn about it ;-)
Will separate out CGI in its own PR and implement streaming for archive access.

mih:
> > ria+http (no CGI deployed)
> > Reconfiguration of special remote failed.
> 
> @mih: Is that still failed as in "clone didn't do it automatically" or did you try to manually run `git annex enableremote SPECIALREMOTENAME url=ria+http://whereever/that/is` and _that_ failed?

I now attempted to manually enable the remote:

```
(datalad) mih@data1 ~/super/ADNI_complete (git)-[master] % git annex enableremote inm7-storage url=ria+http://cat_12.5.ds.inm7.de 
enableremote inm7-storage 
git-annex: Non-absolute object tree base path configuration: .
failed
git-annex: enableremote: 1 failed
(datalad) 1 mih@data1 ~/super/ADNI_complete (git)-[master] % git annex enableremote inm7-storage url=ria+http://cat_12.5.ds.inm7.de/
enableremote inm7-storage ok
(recording state in git...)
```

It seems to require a trailing slash. It think that shouldn't be necessary.

I will update the report in https://github.com/datalad/datalad/pull/4561#issuecomment-633030137

datalad/distributed/ora_remote.py
```diff
@@ -267,6 +269,8 @@ def __init__(self, host):
                 break
         # TODO: Same for stderr?
 
+        self.buffer_size = buffer_size if buffer_size else 65536
```
mih, line 272: 64k constant is defined twice in this file. Would make sense to use a single symbol.

mih:
https://github.com/datalad/datalad/pull/4561/commits/701b50b645e5a6508e1aa9364a9f92ca69edf1e9 breaks clone:

```
(datalad) mih@data1 ~/super (git)-[master] % dl get -n ADNI_complete
[INFO   ] Configured RIA store not found at ria+file:///ds/cat_12.5/srv                      
[ERROR  ] __init__() missing 1 required positional argument: 'buffer_size' [clone.py:postclonecfg_ria:639] (TypeError) 
```

bpoldrack:
> It seems to require a trailing slash. It think that shouldn't be necessary.

Agree.  Will fix.
In case I'm not the only one who missed that so far, here is the issue (otherwise at least the message would have made more sense):
```
In [2]: Path()                                                                                                                         
Out[2]: PosixPath('.')
```

bpoldrack:
"required buffer_size" and "trailing slash" fixed.

bpoldrack:
Test Failures: There appears to be an issue with recent `annexremote`. Investigating.

mih:
> "required buffer_size" and "trailing slash" fixed.

Works now. Thx! Updated the stats above. HTTP still slowest across the board. SSH became faster with the buffer change.

mih:
Tried to see if we would gain anything by computing the dirhash in Python, instead of asking git-annex. Here is the hack -- but it isn't worth it.

```diff
diff --git a/datalad/distributed/ora_remote.py b/datalad/distributed/ora_remote.py
index 182eefd12..ad264c192 100644
--- a/datalad/distributed/ora_remote.py
+++ b/datalad/distributed/ora_remote.py
@@ -1,5 +1,8 @@
 from annexremote import SpecialRemote
 from annexremote import RemoteError
+import hashlib
+import struct
+import os.path as op
 
 from pathlib import (
     Path,
@@ -1013,7 +1016,16 @@ class RIARemote(SpecialRemote):
     def transfer_retrieve(self, key, filename):
 
         if isinstance(self.io, HTTPRemoteIO):
-            self.io.get(PurePosixPath(self.annex.dirhash(key)) / key / key,
+            hasher = hashlib.md5()
+            hasher.update(key.encode())
+            digest = hasher.digest()
+            first_word = struct.unpack('<I', digest[:4])[0]
+            nums = [
+                "0123456789zqjxkmvwgpfZQJXKMVWGPF"[first_word >> (6 * x) & 31]
+                for x in range(4)
+            ]
+            dirhash = "%s%s/%s%s" % (nums[1], nums[0], nums[3], nums[2])
+            self.io.get(op.join(dirhash, key, key),
                         filename,
                         self.annex.progress)
             return
```

bpoldrack:
@mih
> HTTP still slowest across the board. SSH became faster with the buffer change.

Considering your test dataset, that seems to be a per file overhead rather than an actual bandwith issue. Buffer change should only effect actual data transmission. With local webserver I saw a within-file bandwith reporting rising from 40MB/s to 600+MB/s with that change.
However, I don't currently see what we can do about per file overhead in HTTP client side. The overhead from ORA itself is actually much less than with the two other protocols.

bpoldrack:
> Tried to see if we would gain anything by computing the dirhash in Python, instead of asking git-annex. Here is the hack -- but it isn't worth it.

Yes, tried that in context of #4563, too and came to the same conclusion.

bpoldrack:
There currently is one Travis job failing:
https://travis-ci.org/github/datalad/datalad/jobs/691679185#L6221
While the failing test is part of this PR, the actual failure is about `download_url`. @kyleam, does this ring a bell?
```
datalad.support.exceptions.IncompleteResultsError: Command did not complete successfully [{'action': 'download_url', 'path': '/tmp/datalad_temp__test_binary_databm838ddi/dicomfile', 'message': "'QuietConsoleLog' object has no attribute 'yesno' [__init__.py:__getattribute__:101]", 'type': 'file', 'status': 'error'}]
```

bpoldrack:
Apart from that failure in the above post, I think this is ready. The performance issues, @mih reported, appear to be server-side. See https://github.com/datalad/datalad/issues/4579 for an analysis. This needs to be solved in PR #4572 and does not affect the special remote implementation. Except someone shows me wrong in #4579, of course.

kyleam:
> While the failing test is part of ~this PR~ the special remote tests, the actual failure is about `download_url`. @kyleam, does this ring a bell?

No, I don't recall seeing that `test_ria_basics.test_binary_data` failure.  I couldn't trigger the failure locally with either the latest git-annex or one built from the version in neurodebian, using the environment variables set up in that run.  I'd guess it's just transient, though maybe we should apply changes like 5ba071303 more widely to get better information here.

bpoldrack:
> I'd guess it's just transient, ...

Restarted and it passed this time. So, yeah ...

> though maybe we should apply changes like 5ba0713 more widely to get better information here

That would certainly be useful. Thanks for pointing to that commit!

mih:
FTR: Cummulativ impact of a fixes, server-side and in this code yield:

### ria+http (no CGI deployed)
`datalad get .  363.65s user 94.57s system 85% cpu 8:56.79 total`

up from

`datalad get . 404.39s user 100.34s system 49% cpu 16:56.93 total`

datalad/core/distributed/clone.py
```diff
@@ -653,12 +646,16 @@ def postclonecfg_ria(ds, props):
...
+            lgr.debug("Unknown URL-Scheme in %s. Can handle SSH, HTTP or "
                       "FILE scheme URLs.", props['source'])
 
         # 3. And read it
         org_uuid = None
         if config_content:
+            # TODO: We might be able to spare the saving to a file.
+            #       "git config -f -" is not explicitly documented but happens
+            #       to work and would read from stdin. Make sure we know this
+            #       works for required git versions and on all platforms.
```
kyleam, line 658: FWIW the stdin behavior is explicitly checked in Git's tests and has been around since v2.0.0, specifically 3caec73b55 (config: teach "git config --file -" to read from the standard input, 2014-02-19).

datalad/distributed/tests/test_ora_http.py
```diff
@@ -0,0 +1,134 @@
...
+               for uuid, cfg in ds.repo.get_special_remotes().items()]
+              )
+    assert_repo_status(ds.path)
+    # git-annex:remote.log should have:
+    #   - url
+    #   - common_init_opts
+    #   - archive_id (which equals ds id)
+    remote_log = ds.repo.call_git(['cat-file', 'blob', 'git-annex:remote.log'])
+    assert_in("url={}".format(url), remote_log)
+    [assert_in(c, remote_log) for c in common_init_opts]
```
kyleam, line 74: Not a proper review, but I read through and didn't spot any obvious issues.

This line, though... what is this world coming to? :)

bpoldrack: With this line the world is coming to ... comprehension?

bpoldrack: Unlikely, that this would go wrong for just a single config option while not actually failing during initremote. Which is why I don't believe it makes a difference. But if I can make you happy by a proper loop, I will! :)

kyleam: yeah, I guess "comprehension" must be among valid answers to that question :)

[ Hitting the resolve button just to make it clear that I wasn't requesting a change here. ]

PR #4572, opened 2020-05-25
ENH: ORA CGI for RIA over HTTP

bpoldrack:
Initial implementation of CGI script to provide ORA functionality
for a RIA store over HTTP. Read-only access for now.

This is meant in conjunction with PR #4561.
As discussed over there, trying to change this to use streaming for GET. Reason for this is, that we currently need to read an entire file (or worse - extract an entire file) before we build the response, which suggests to lead to an undesirable latency. However, while I got streaming to technically work, with my local apache setup I see a slowdown by factor 2 instead of a speed-up. Not yet clear to me, what this is about. I suspected that simply connecting the 7z-subprocess-stdout to sys.stdout might get us into trouble with buffering, but an alternative approach using Popen and introduce a buffer didn't improve anything so far (at least with buffer sizes from 60K - 10MB it didn't make any noteable difference).

### DIFF
--- a/datalad/resources/ora_cgi.py
+++ b/datalad/resources/ora_cgi.py
@@ -1,0 +1,236 @@
+#!/usr/bin/env python
+
+import os
+import hashlib
+import subprocess
+from pathlib import Path
+
+# for debugging: (consider enable(display=0, logdir=/log/path/)  later on)
+# import cgi
+#import cgitb
+#cgitb.enable()  # note: logdir="/tmp/logs" didn't work locally (why?)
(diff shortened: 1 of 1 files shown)


mih:
Demand for this feature is now high. Shall I take over?

mih:
FTR: There is no need, or much benefit from having this bloat the code base of -core further. Let's put this elsewhere.

bpoldrack:
Closing in favor of https://github.com/datalad/ria-tools

PR #4573, opened 2020-05-25
Fix ORA for recent annex/annexremote

bpoldrack:
`annexremote` changed two things in most recent 1.4.2:
 - special remote protocol's LISTCONFIGS is now supported and yields an actual response rather
   the former UNSUPPORTED REQUEST. If we don't declare configs we use w/ `initremote`, annex
   will now retrieve an empty list
 - `annexremote.Master.protocol` turned ~private~ protected

In addition git-annex 8 (used in github Action workflow) apparently now requires configs for `initremote` to be listed, if the response isn't UNSUPPORTED REQUEST. In combination with above change in `annexremote` this currently leads to all `initremote` calls for ORA to return non-zero "Unexpected parameter: url".

This PR fixes those failures (without changing dependencies).

### DIFF
--- a/datalad/distributed/ora_remote.py
+++ b/datalad/distributed/ora_remote.py
@@ -1,5 +1,6 @@
 from annexremote import SpecialRemote
 from annexremote import RemoteError
+from annexremote import ProtocolError
 
 from pathlib import (
     Path,
@@ -570,9 +571,6 @@ class RIARemote(SpecialRemote):
         self.store_base_path = None
         # by default we can read and write
         self.read_only = False
-        # to be figured out later, since annex.protocol.extensions is not yet
-        # accessible:
-        self.can_notify = None
         self.force_write = None
         self.uuid = None
         self.ignore_remote_config = None
@@ -759,7 +757,7 @@ class RIARemote(SpecialRemote):
 
         file_content = self.io.read_file(path).strip().split('|')
         if not (1 <= len(file_content) <= 2):
-            self._info("invalid version file {}".format(path))
+            self.message("invalid version file {}".format(path))
             return None
 
         remote_version = file_content[0]
@@ -846,23 +844,27 @@ class RIARemote(SpecialRemote):
         #return self.store_base_path.is_dir()
         return not self.storage_host
 
-    def _info(self, msg):
+    def debug(self, msg):
+        # Annex prints just the message, so prepend with
+        # a "DEBUG" on our own.
+        self.annex.debug("ORA-DEBUG: " + msg)
 
-        if self.can_notify:
+    def message(self, msg):
+        try:
             self.annex.info(msg)
-        # TODO: else: if we can't have an actual info message, at least have a
-        #       debug message.
-        #       This probably requires further refurbishment of datalad's
-        #       capability to deal with such aspects of the
-        #       special remote protocol when parsing annex' output.
+        except ProtocolError:
+            # INFO not supported by annex version.
+            # If we can't have an actual info message, at least have a
+            # debug message.
+            self.debug(msg)
 
     def _set_read_only(self, msg):
 
         if not self.force_write:
             self.read_only = True
-            self._info(msg)
+            self.message(msg)
         else:
-            self._info("Was instructed to force write")
+            self.message("Was instructed to force write")
 
     def _ensure_writeable(self):
         if self.read_only:
@@ -887,9 +889,6 @@ class RIARemote(SpecialRemote):
 
     @handle_errors
     def prepare(self):
-
-        # can we use self.annex.info() for sending user output to annex?
-        self.can_notify = "INFO" in self.annex.protocol.extensions
 
         gitdir = self.annex.getgitdir()
         self.uuid = self.annex.getuuid()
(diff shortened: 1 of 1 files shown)


datalad/distributed/ora_remote.py
```diff
@@ -846,23 +847,27 @@ def _local_io(self):
...
+    def message(self, msg):
+        try:
             self.annex.info(msg)
-        # TODO: else: if we can't have an actual info message, at least have a
-        #       debug message.
-        #       This probably requires further refurbishment of datalad's
-        #       capability to deal with such aspects of the
-        #       special remote protocol when parsing annex' output.
+        except ProtocolError:
+            # INFO not supported by annex version.
```
yarikoptic, line 859: IMHO it is better to just compare  current git annex version against known once and behave accordingly. That will later allow to prune such conditional  code when our minimal supported annex version goes ahead like did recently in one of prs.

bpoldrack: I don't agree. It's a matter of protocol extensions which git-annex is build with. And `annexremote` asks annex about supported extensions during PREPARE. So it's not exactly a version number this is about.

bpoldrack: It's explicitly part of the protocol to figure supported features rather than relying on some reported version number. Seems the cleanest way to me, to make use of that. We have that specific information available in `super` anyway.

yarikoptic: oh, my bad.  I thought that `INFO` is like `DEBUG` just was introduced as a part of the standard protocol, and not as an EXTENSION.  I do not think it would be really "annex build dependent" but indeed better to not just rely on the version.  I guess this should be ok in the short run.
In the long run -- I feel that AnnexRemote should provide public API to query for extensions. Could you please raise an issue on https://github.com/Lykos153/AnnexRemote/issues ?  It would also be useful to alert author that API is used by 3rd party (us) and ideally should not be too fluid ;)

bpoldrack: I can do that, but I don't really see the benefit. What's wrong with catching that exception instead of duplicating what `annexremote.Master.info` does? (https://github.com/Lykos153/AnnexRemote/blob/master/annexremote/annexremote.py#L1256)

kyleam:
Thanks, @bpoldrack.  00ac09fd9 does a nice job of explaining the interaction between git-annex and annexrepo.

For posterity, here's a link to one of the failing runs in datalad-extensions: https://github.com/datalad/datalad-extensions/runs/704820248

With git-annex 8.20200522 and datalad's current master (a3a154bcb), I tried a few of those tests locally.  They passed with the annexremote I had installed (1.3.1).  Upgrading annexremote to 1.4.2, they failed.  Checking out this PR returned them to a passing state.  So, all as expected.

bpoldrack:
Thanks for checking, @kyleam!

datalad/distributed/ora_remote.py
yarikoptic: re benefit: just catching a generic ProtocolError is a guess work.
API breakage on AnnexRemote side is that -- API breakage.  See above why letting developer know about it could be useful.

yarikoptic: >  It's just that I adapted our code to meet the paradigm annexremote is suggesting (use those methods and catch a ProtocolError)

oh, could you please point to such suggestion?

bpoldrack: 1. Misunderstanding, I think. Not opposing letting `annexremote`'s author know, that API breakage has consequences. ;)
My point is about making extensions public.

2. 
> just catching a generic ProtocolError is a guess work.
```
        Raises
        ----------
        ProtocolError
            If INFO is not available in this version of git-annex.
```
I don't know what's unspecific about that. There's nothing else that could raise a `ProtocolError` when we call this method.

The alternative you're proposing by requesting a public extensions query is to simply reimplement `annexremote.Master.info` ourselves (see link above). Why would we rather rely on the stability of that query than the stability of `info` raising that Error?
Plus: As mentioned in original post and commit message with that approach we would then also need to either raise dependency version for `annexremote` or check its version in addition. Whereas with the exception it just works with both versions and leaves to `annexremote` what it already implemented.

> oh, could you please point to such suggestion?

Well, that's just the way it's implemented. I take code as a suggestion ;-)

kyleam: (my two cents) Catching a `ProtocolError` here looks like a good approach.

yarikoptic: ok, thank you @bpoldrack for the details and bearing with me.  Let's hope that in the future this generic (and base of `UnsupportedRequest` which IMHO would have been a better choice) ProtocolError will also not be spitted out for anything else while issuing an info message.  ATM it is unlikely indeed we would hit any other functionality while issuing an INFO msg.

Re API breakage "ping" -- please make friends or I will! ;)

bpoldrack: >  Re API breakage "ping" -- please make friends or I will! ;)

Nooo - what a mean threat! :-D

Done.

Issue #4563, opened 2020-05-21
ORA performance


bpoldrack:
------------- IGNORE POST UNTIL THIS LINE DISAPPEARS -------------

Just securing some snippets to be put into analysis later

```
(datalad-dev) ben@tree in /tmp/100k-files on git:master
❱ time datalad push --to store
copy(ok): 1 (file) [to store-storage...]                                                                                                                                                                                        
copy(ok): 10 (file) [to store-storage...]                                                                                                                                                                                       
copy(ok): 100 (file) [to store-storage...]                                                                                                                                                                                      
copy(ok): 1000 (file) [to store-storage...]                                                                                                                                                                                     
copy(ok): 10000 (file) [to store-storage...]                                                                                                                                                                                    
copy(ok): 10001 (file) [to store-storage...]                                                                                                                                                                                    
copy(ok): 10002 (file) [to store-storage...]                                                                                                                                                                                    
copy(ok): 10003 (file) [to store-storage...]                                                                                                                                                                                    
copy(ok): 10004 (file) [to store-storage...]                                                                                                                                                                                    
copy(ok): 10005 (file) [to store-storage...]                                                                                                                                                                                    
  [99989 similar messages have been suppressed]                                                                                                                                                                                 
publish(ok): . (dataset) [refs/heads/master->store:refs/heads/master [new branch]]                                                                                                                                              
publish(ok): . (dataset) [refs/heads/git-annex->store:refs/heads/git-annex [new branch]]                                                                                                                                        
datalad push --to store  445.34s user 114.71s system 113% cpu 8:14.90 total            
```                                      

```
git annex copy . --to store-storage  280.97s user 87.55s system 109% cpu 5:35.11 total
```

mih:
FTR: It would be good to compare the times after push has started transferring data. In the code that is current in master three things are happening before that (and produce an overhead):

- diff to figure out what has changed in *git*
- non-annex branch push
- annex find to gather information on the annex'd files that are in the changeset

bpoldrack:
--- RESERVE POST TO THE SAME AS IN ORIGINAL POST FOR DOWNLOAD ----

bpoldrack:
Re more detailed measurement of `push`:
Yes, agree and will do while I'm at it. But I'd prefer to put that in another issue. Since ORA needs RF'ing (long planned as well as a consequence of supporting HTTP), it want to focus on figuring want needs to be considered for the ORA implementation in addition to what I already know.

[...]

PR #4574, opened 2020-05-25
PERF: Reduce per-file overhead by ORA remote

bpoldrack:
Trying to address #4563

This should reduce the per-file overhead by avoiding to query annex for DIRHASH twice during `git annex copy --to` and have that effect with a local RIA store as well as over SSH.

### DIFF
--- a/datalad/distributed/ora_remote.py
+++ b/datalad/distributed/ora_remote.py
@@ -587,6 +587,10 @@ class RIARemote(SpecialRemote):
         self.remote_obj_dir = None
         self._io = None  # lazy
 
+        # cache obj_locations:
+        self._last_archive_path = None
+        self._last_keypath = (None, None)
+
     def verify_store(self):
         """Check whether the store exists and reports a layout version we
         know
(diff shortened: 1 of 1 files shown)


bpoldrack:
For local FS and 10k files there appears to be no notable difference, while profile and protocol monitoring suggest, that the change works as intended. This makes little sense to me. Didn't expect huge impact, but zero? Wondered whether tuples are particularly slow and compared, but this isn't it either. 

master:
```
git annex copy 1*.txt --to perf-store-storage  58.82s user 11.83s system 144% cpu 48.994 total
git annex copy 1*.txt --to perf-store-storage  59.14s user 11.73s system 142% cpu 49.616 total
git annex copy 1*.txt --to perf-store-storage  59.28s user 12.39s system 142% cpu 50.459 total
```

https://github.com/datalad/datalad/pull/4574/commits/155b0165bea14ad78f2ceaeb7c674aea3a3c537a:
```
git annex copy 1*.txt --to perf-store-storage  59.43s user 12.05s system 142% cpu 50.198 total
git annex copy 1*.txt --to perf-store-storage  60.42s user 11.31s system 149% cpu 47.974 total
git annex copy 1*.txt --to perf-store-storage  58.36s user 11.59s system 146% cpu 47.783 total
```

`dict('key': .., 'path': ..)` instead of `tuple(key, path)` for caching:
```
git annex copy 1*.txt --to perf-store-storage  58.79s user 11.65s system 146% cpu 48.108 total
git annex copy 1*.txt --to perf-store-storage  60.44s user 11.85s system 150% cpu 48.040 total
git annex copy 1*.txt --to perf-store-storage  59.50s user 11.97s system 147% cpu 48.456 total
```

So, try again with 100k files (same as in #4563):

master:
```
git annex copy . --to perf-store-storage  369.26s user 121.57s system 108% cpu 7:32.89 total
git annex copy . --to perf-store-storage  336.37s user 111.84s system 105% cpu 7:06.60 total
```

https://github.com/datalad/datalad/pull/4574/commits/155b0165bea14ad78f2ceaeb7c674aea3a3c537a:
```
git annex copy . --to perf-store-storage  289.65s user 73.47s system 111% cpu 5:25.92 total
git annex copy . --to perf-store-storage  295.42s user 72.73s system 111% cpu 5:30.45 total
```

`dict('key': .., 'path': ..)` instead of `tuple(key, path)` for caching:
```
git annex copy . --to perf-store-storage  338.43s user 104.63s system 109% cpu 6:42.80 total
git annex copy . --to perf-store-storage  310.55s user 98.92s system 110% cpu 6:09.93 total
```

Not sure at all. Measurement is way off what it was in #4563 even on master. So, something is
quite unreliable here. However, at least in the 100k case, the difference between master and https://github.com/datalad/datalad/pull/4574/commits/155b0165bea14ad78f2ceaeb7c674aea3a3c537a seems sufficient to not be explained by overall variance of the measurement, thereby justifying the change ;-)
According to the profile in #4563 I would expect an improvement of about 5-6% (given the time we spend in `get_obj_locations` and `dirhash` respectively. Comparing best master run to worst run with this change seems too big an effect (100k files), while in the 10k case there appears to be none at all. Furthermore variance within the same code running is easily within the effect size of improvements we might want to measure. I have no idea ATM, but we need a better way to measure this kind of thing.

mih:
Accidentally reviewed. Spotted the "draft" too late. Sorry.

bpoldrack:
That's fine. Rebased and wanted to see tests before removing "draft" - that's all.

PR #4578, opened 2020-05-26
ENH: un-deprecate publish

yarikoptic:
per weekly hangout discussion.

individual commit msgs might provide more reasoning

### DIFF
--- a/datalad/distribution/publish.py
+++ b/datalad/distribution/publish.py
@@ -571,11 +571,6 @@ def _get_remote_diff(repo, current_commit, remote, remote_branch_name):
 class Publish(Interface):
     """Publish a dataset to a known :term:`sibling`.
 
-    ..note::
-      This command is deprecated. It will be removed from DataLad eventually,
-      but no earlier than with the 0.14 release. Use the `push` and `create-sibling` 
-      (if you used `--missing` flag) commands instead.
-
     This makes the last saved state of a dataset available to a sibling
     or special remote data store of a dataset. Any target sibling must already
     exist and be known to the dataset.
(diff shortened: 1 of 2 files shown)


kyleam:
> remove (duplicate) deprecation notices

What point does "(duplicate)" here serve aside from suggesting you think it was an error and wanting to point that out?  One of them is in the rc2 section, which clearly states its to be pruned.  The other was an entry for the actual 0.13.0 release.   So they both were intended.

yarikoptic:
no point other than in the current state it was two of them (in either to be pruned or not).

datalad/distribution/publish.py
```diff
@@ -600,6 +595,12 @@ class Publish(Interface):
       to publish a dataset. Publication targets are either configured remote
       Git repositories, or git-annex special remotes (if they support data
       upload).
+
+    ..note::
+      A new (since 0.13.0) `push` command provides an alternative interface to
+      publish to siblings. Critical differences from `publish` behavior are: `push`
+      by default copies annexed data, and it does not have `--missing` flag so you
+      will use `create-sibling`.
```
kyleam, line 603: From a comprehension standpoint, I struggled with the "you will use `create-sibling`" part.  Perhaps "need to" after "will" would make things clearer.  Otherwise, it reads fine to me and (AFAIU) is in line with what was decided on the call.

Here's a fuller and more subjective rewrite, in case you think any parts of it are an improvement:

```
      The `push` command (new in 0.13.0) provides an alternative interface.
      Critical differences are that `push` transfers annexed data by default
      and does not handle sibling creation (i.e. it does not have a `--missing`
      option).
```

mih: I like!

yarikoptic: thanks, will rewrite with this wording

yarikoptic:
updated, should be good to go AFAIK

Issue #4551, opened 2020-05-18
`push` default behavior is to push all the data -- IMHO undesirable in many cases and unsafe


yarikoptic:
`publish` by default copied no data to the remote unless specific files were pointed to.

`push` seems to try to publish all the data right away.  

<details>
<summary>Script adjusted from the one posted in https://github.com/datalad/datalad/issues/4550#issuecomment-630268894 but without wanted set</summary> 

```shell
$> cat ./check-push-content-nowanted.sh
#!/bin/bash

export PS4='> '
set -eu
cd "$(mktemp -d ${TMPDIR:-/tmp}/dl-XXXXXXX)"
set -x

datalad create src

(
cd src
# Produce and annotate some data
echo -e '\x00' > data.0
echo -e '\x01' > secure.1
echo -e '\x02' > secure.2
datalad save -m "added some files"

# Dropping a file to mimic a case of simply not having it locally (thus not to be "pushed")
git annex drop --force secure.2

# Annotate sensitive content, actual value "verysecure" does not matter in this example
git annex metadata --set distribution-restrictions=verysecure secure.1 secure.2

datalad save -m "added a file"
datalad create-sibling -r -s target ../target
datalad push --to target
)

(
cd target
git annex list
)
```
</details>

which results in
```shell
> datalad push --to target
publish(ok): . (dataset) [refs/heads/master->target:refs/heads/master [new branch]]                            
[WARNING] Slated for transport, but no content present [copy(/home/yoh/.tmp/dl-npppKEt/src/secure.2)] 
copy(impossible): secure.2 (symlink) [Slated for transport, but no content present]
copy(ok): data.0 (file) [to target...]                                                                         
copy(ok): secure.1 (file) [to target...]
publish(ok): . (dataset) [refs/heads/git-annex->target:refs/heads/git-annex 369c340..a4ffe63]                
```

I see that in some cases such default behavior would be desired, and warnings about some files be impossible to transfer are due. But that is IMHO an undesirable default behavior:

- quite often people do not have all the files in the repository (containers etc), and want to publish what they have -- complaining about files they do not have be just wrong and distracting (could be many more from what they actually desire to share/publish)
- I might not want to publish stuff which is already available elsewhere (e.g. containers)
- only a matter of time until pushing of all data by default would leak some sensitive data into public space 
- I typically do not push all the data by default. I rely on `wanted` set and `--auto` (see further discussion on --auto in #4550) 

[...]

Issue #4550, opened 2020-05-17
When should `push` (NOT) add `--auto` to `git-annex copy`?


mih:
In https://github.com/datalad/datalad/pull/4541#issuecomment-629589141 I proposed that `--auto` should always be used, unless there is counter-evidence. This is a departure from what `publish` does (add `--auto` if `wanted` is set; https://github.com/datalad/datalad/pull/4541#issuecomment-629479069), and what `push` currently does (`not to_transfer and force not in ('all', 'datatransfer')`, which is problematic #4483, and doesn't make much sense in general).

@kyleam pointed out that `numcopies=1` has the potential to shoot us in the foot, if we do say `--auto` by default (https://github.com/datalad/datalad/pull/4541#issuecomment-629689667). `numcopies=1` is actually the default, and worse, it could be set on a per-file level. For this reason, it is not advisable to adopt the behavior of `publish`.

It felt attractive to me to say that a precondition for adding `--auto` internally is the absence of a `path` argument (a way of saying "you go figure it out"), but that would imply that any subselection of subdatasets to publish would not be possible. For my workflows this has negligible impact (if any), because hierarchies are typically shallow and homogeneous (subdatasets are either technical means of splitting a bigger monorepo, or all subdatasets are "data dependencies"/input). However, @yarikoptic's workflows might be completely different and may involve deep heterogeneous dataset hierarchies.

The more I think about it, the more attractive https://github.com/datalad/datalad/pull/4541 becomes, which made adding `--auto` explicit (ignoring the implementation details for now).

I am also still pondering the idea set add `--auto` unconditionally, and combine it with `numcopies=2` for the runtime of `push`. If one calls `push` they clearly want to transfer information. However, that may cause undesired interactions with tailored wanted expressions where file content might be copied without a desire to have that done. But this part is still too fuzzy to me.

In any case the test included in #4541 should be salvaged (thx @kyleam )

yarikoptic:
I (or someone else) yet to double check interaction between numcopies and wanted - if I set wanted to exclude some files, would they still be copied over if numcopies>1? That would undermine effort to avoid distribution of sensitive data

yarikoptic:
I was not sure if I should create another issue or it is just manifestation of absence of `--auto` treatment which is discussed here, but 

<details>
<summary>here is a test script (ATM wouldn't even complete running since `push` would somewhat rightfully exit with non-0 ATM due to wrong IMHO behavior)</summary> 

```shell
[...]

Issue #4357, opened 2020-03-29
Dataset-provided configuration for `publish/push`


mih:
Re https://neurostars.org/t/using-bitbucket-with-datalad/6226/6

What would setting `remote.origin.datalad-publish-depends=<some special remote>` in `.datalad/config` do? The special remote could come preconfigured and autoenabled with a dataset. If `datalad` honors such configuration, it would always push to that remote first.

mih:
This thought does not require living in an open issue.

PR #4581, opened 2020-05-28
Finishing touches to `push` for 0.13

mih:
Rename push(force='all') -> push(force='pushall')  to better forward-compatibility.

This also adds the implementation of the conclusion of a discussion on this topic. Follow-up to gh-4541.

Add --auto to `git annex copy` if `datalad.push.copy-auto-if-wanted` is set, and `--force datatransfer` isn't given.

Fixes gh-4550: --auto is never added, unless requested via config

Fixes gh-4551 and fixes gh-4483: now obeys distribution-restrictions in a "default"  `datalad push --to target` if configuration says so (not the default (yet?))

TODO:

- [x] Is this secure enough (see gh-4357). A dataset could set this flag in its own config (.datalad/config). Are we confident that it would never override a user config (add test)?
- [x] update docstring to say that be default "all content" will be pushed, and explain under which conditions exactly (tracking remote, etc.)

<details>
<summary>Script to play with the changes</summary>

```sh

#!/bin/bash

export PS4='> '
set -eu
set -x

datalad create src

(
cd src
# Produce and annotate some data
echo -e '\x00' > data.0
echo -e '\x01' > secure.1
echo -e '\x02' > secure.2
datalad save -m "added some files"

# Dropping a file to mimic a case of simply not having it locally (thus not to be "pushed")
git annex drop --force secure.2

# Annotate sensitive content, actual value "verysecure" does not matter in this example
git annex metadata --set distribution-restrictions=verysecure secure.1 secure.2

datalad save -m "added a file"
datalad create-sibling --annex-wanted "not metadata=distribution-restrictions=*" -r -s target ../target
#datalad -c datalad.push.copy-auto-if-wanted=true push --to target
datalad push --to target

)

(
cd target
git annex list
)
```

</details>

### DIFF
--- a/datalad/core/distributed/push.py
+++ b/datalad/core/distributed/push.py
@@ -16,7 +16,6 @@ import logging
 from tempfile import TemporaryFile
 
 from datalad.cmd import GitWitlessRunner
-from datalad.config import anything2bool
 from datalad.interface.base import (
     Interface,
     build_doc,
@@ -774,8 +773,7 @@ def _push_data(ds, target, content, force, jobs, res_kwargs,
         cmd.extend(['--jobs', str(jobs)])
 
     if force not in ('pushall', 'datatransfer') and ds_repo.config.obtain(
-            'datalad.push.copy-auto-if-wanted', default=False,
-            valtype=anything2bool):
+            'datalad.push.copy-auto-if-wanted'):
         if ds_repo.get_preferred_content('wanted', target):
             lgr.debug("Invoking copy --auto")
             cmd.append('--auto')
--- a/datalad/core/distributed/push.py
+++ b/datalad/core/distributed/push.py
@@ -64,9 +64,9 @@ lgr = logging.getLogger('datalad.core.distributed.push')
 class Push(Interface):
     """Push a dataset to a known :term:`sibling`.
 
-    This makes the last saved state of a dataset available to a sibling
-    or special remote data store of a dataset. Any target sibling must already
-    exist and be known to the dataset.
+    This makes a saved state of a dataset available to a sibling or special
+    remote data store of a dataset. Any target sibling must already exist and
+    be known to the dataset.
 
     Optionally, it is possible to limit a push to change sets relative
     to a particular point in the version history of a dataset (e.g. a release
--- a/datalad/core/distributed/push.py
+++ b/datalad/core/distributed/push.py
@@ -125,9 +125,9 @@ class Push(Interface):
             making: use --force with git-push ('gitpush'); do not use --fast
             with git-annex copy ('datatransfer'); do not attempt to copy
             annex'ed file content ('no-datatransfer'); combine force modes
-            'gitpush' and 'datatransfer' ('all').""",
+            'gitpush' and 'datatransfer' ('pushall').""",
             constraints=EnsureChoice(
-                'all', 'gitpush', 'no-datatransfer', 'datatransfer', None)),
+                'pushall', 'gitpush', 'no-datatransfer', 'datatransfer', None)),
         recursive=recursion_flag,
         recursion_limit=recursion_limit,
         jobs=jobs_opt,
@@ -650,7 +650,7 @@ def _push_refspecs(repo, target, refspecs, force, res_kwargs):
         push_res.extend(repo.push(
             remote=target,
             refspec=refspec,
-            git_options=['--force'] if force in ('all', 'gitpush') else None,
+            git_options=['--force'] if force in ('pushall', 'gitpush') else None,
         ))
     # TODO maybe compress into a single message whenever everything is
     # OK?
@@ -709,7 +709,7 @@ def _push_data(ds, target, content, force, jobs, res_kwargs,
             res_kwargs,
             action='copy',
             status='impossible'
-            if force in ('all', 'datatransfer')
+            if force in ('pushall', 'datatransfer')
             else 'notneeded',
             message=(
                 "Target '%s' does not appear to be an annex remote",
@@ -743,7 +743,7 @@ def _push_data(ds, target, content, force, jobs, res_kwargs,
         c
         for c in content.values()
         # by force
-        if ((force in ('all', 'datatransfer') or
+        if ((force in ('pushall', 'datatransfer') or
              # or by modification report
              c.get('state', None) not in ('clean', 'deleted'))
             # only consider annex'ed files
@@ -767,11 +767,11 @@ def _push_data(ds, target, content, force, jobs, res_kwargs,
     if jobs:
         cmd.extend(['--jobs', str(jobs)])
 
-    if not to_transfer and force not in ('all', 'datatransfer'):
+    if not to_transfer and force not in ('pushall', 'datatransfer'):
         lgr.debug("Invoking copy --auto")
         cmd.append('--auto')
 
-    if force not in ('all', 'datatransfer'):
+    if force not in ('pushall', 'datatransfer'):
         # if we force, we do not trust local knowledge and do the checks
         cmd.append('--fast')
 
--- a/datalad/core/distributed/push.py
+++ b/datalad/core/distributed/push.py
@@ -767,9 +767,11 @@ def _push_data(ds, target, content, force, jobs, res_kwargs,
     if jobs:
         cmd.extend(['--jobs', str(jobs)])
 
-    if not to_transfer and force not in ('pushall', 'datatransfer'):
-        lgr.debug("Invoking copy --auto")
-        cmd.append('--auto')
+    if force not in ('pushall', 'datatransfer') and ds_repo.config.get(
+            'datalad.push.copy-auto-if-wanted', False):
+        if ds_repo.get_preferred_content('wanted', target):
+            lgr.debug("Invoking copy --auto")
+            cmd.append('--auto')
 
     if force not in ('pushall', 'datatransfer'):
         # if we force, we do not trust local knowledge and do the checks
--- a/datalad/core/distributed/push.py
+++ b/datalad/core/distributed/push.py
@@ -69,15 +69,17 @@ class Push(Interface):
     remote data store of a dataset. Any target sibling must already exist and
     be known to the dataset.
 
+    || REFLOW >>
     By default, all files tracked in the last saved state (of the current
     branch) will be copied to the target location. Optionally, it is possible
     to limit a push to changes relative to a particular point in the version
     history of a dataset (e.g. a release tag) using the
     [CMD: --since CMD][PY: since PY] option in conjunction with the
-    specification of a reference dataset. Changes can also be evaluated
-    recursively, i.e. only those subdatasets are pushed where a change was
+    specification of a reference dataset. In recursive mode subdatasets will also be
+    evaluated, and only those subdatasets are pushed where a change was
     recorded that is reflected in the current state of the top-level reference
     dataset.
+    << REFLOW ||
 
     Which files are copied can be further tailored via the
     'datalad.push.copy-auto-if-wanted' configuration. If set, push will test
--- a/datalad/core/distributed/push.py
+++ b/datalad/core/distributed/push.py
@@ -78,6 +78,13 @@ class Push(Interface):
     recorded that is reflected in the current state of the top-level reference
     dataset.
 
+    Which files are copied can be further tailored via the
+    'datalad.push.copy-auto-if-wanted' configuration. If set, push will test
+    whether a git-annex "wanted" configuration is present for the target
+    location, and in this case instruct git-annex to obey this configuration
+    when deciding which files to consider for transfer (i.e. use the --auto
+    flag with git-annex copy).
+
     .. note::
       Power-user info: This command uses :command:`git push`, and :command:`git
       annex copy` to push a dataset. Publication targets are either configured
--- a/datalad/core/distributed/push.py
+++ b/datalad/core/distributed/push.py
@@ -68,25 +68,21 @@ class Push(Interface):
     remote data store of a dataset. Any target sibling must already exist and
     be known to the dataset.
 
-    Optionally, it is possible to limit a push to change sets relative
-    to a particular point in the version history of a dataset (e.g. a release
-    tag). By default, the state of the local dataset is evaluated against the
-    last known state of the target sibling. An actual push is only attempted
-    if there was a change compared to the reference state, in order to speed up
-    processing of large collections of datasets. Evaluation with respect to
-    a particular "historic" state is only supported in conjunction with a
-    specified reference dataset. Change sets are also evaluated recursively, i.e.
-    only those subdatasets are pushed where a change was recorded that is
-    reflected in the current state of the top-level reference dataset.
-    See "since" option for more information.
-
-    Only a push of saved changes is supported.
+    By default, all files tracked in the last saved state (of the current
+    branch) will be copied to the target location. Optionally, it is possible
+    to limit a push to changes relative to a particular point in the version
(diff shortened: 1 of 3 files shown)


yarikoptic:
> Add --auto to git annex copy if datalad.push.copy-auto-if-wanted is set, and --force datatransfer isn't given.

A door of combinatorics is slightly opened... I forgot - did we discuss many be the `--force autodatatransfer` as an approach to trigger it instead?

What about renaming --force to --mode?

mih:
>> Add --auto to git annex copy if datalad.push.copy-auto-if-wanted is set, and --force datatransfer isn't given.

> A door of combinatorics is slightly opened... I forgot - did we discuss many be the `--force autodatatransfer` as an approach to trigger it instead?

I debated that question too. I found that it is better and useful to be able to say `datalad push` and get what is "right" for a particular dataset, instead of having to remember to say `datalad push -f autodatatransfer` for some. At the same time it seems inappropriate to configure a fixed "force" mode. However, the only critical issue with a preconfigured "force" mode is that once configured, a cmdline option isn't capable of going back to "no force". And having a `--force no` for this purpose seems a bit silly.

> What about renaming --force to --mode?

I am not convinced that one is better than the other. While I agree that `--mode` fits here, I can easily see `--force` be a better fit across commands. We already have it for `create`, and I am not aware of a `--mode` for another command. So if a compact, predictable setup for options is a goal, I'd vote for `--force` to stay.

yarikoptic:
> I debated that question too. I found that it is better and useful to be able to say `datalad push` and get what is "right" for a particular dataset, instead of having to remember to say `datalad push -f autodatatransfer` for some

But then isn't such logic applicable for any "force" mode of operation? i.e. shouldn't it be just `datalad.push.force` configuration setting which would set the mode of force'ful operation?  (that would also avoid combinatorics between config setting and --force option).  And then it could/should be treated in a regular chain of preferences we have in general for options - BUT we IIRC indeed yet to establish it for "when it is also a regular --option" - would config take over cmdline or the other way around.

> I am not convinced that one is better than the other. While I agree that `--mode` fits here, I can easily see `--force` be a better fit across commands. We already have it for `create`, and I am not aware of a `--mode` for another command.

although the difference indeed not really well defined (between force and mode), I would argue that analogy with `create` is in favor of not using `--force` here.  In `create` it is really to force operation which otherwise can error out (similar to other tools use of `--force` e.g. `cp`, and even `git push` as well, and good old `datalad publish`).  It is also a flag in `create` (and `git push`, `datalad publish`, `cp` et al), while here it is a parameter.

mih:
re force v mode: I didn't want to present an analogy, I just pointed out that we already have `--force`, but not `--mode`. When we have modes (e.g. `siblings`) the command implement that without a central mode switch. Hence the notion of `--mode` is completely new, and I am not convinced that it solves a problem better than `--force`.

All implemented "modes" of `--force` alter the behavior of the command. They are nothing but a bunch of mutually exclusive options. For each of them a force... interpretation is straightforward

- force gitpush (quite literally)
- force datatransfer (despite local knowledge that data is already present at the remote)
- force no-datatransfer (despite a local change indicating that it should happen)
- force pushall (just make it happen, I don't care)

`--auto` is different. Whether or not to add `--auto` isn't that exclusive as far as I can see. It is more like "if you happen to attempt a copy, do it like this".

Yes, we could add configuration layers to all options, but so far there wasn't a need. If we do it, there is nothing in here that hinders that AFAICS.

datalad/core/distributed/push.py
```diff
@@ -64,29 +65,32 @@
...
+    remote data store of a dataset. Any target sibling must already exist and
+    be known to the dataset.
+
+    By default, all files tracked in the last saved state (of the current
+    branch) will be copied to the target location. Optionally, it is possible
+    to limit a push to changes relative to a particular point in the version
+    history of a dataset (e.g. a release tag) using the
+    [CMD: --since CMD][PY: since PY] option in conjunction with the
+    specification of a reference dataset. Changes can also be evaluated
+    recursively, i.e. only those subdatasets are pushed where a change was
```
yarikoptic, line 78: a little unclear to me, since it was talking about "states" but then switched to "changes".  Do I understand it correctly that the last sentence means: in `-r` mode, `push` will not even consider recursing into any subdataset if the last commit in its parent dataset did not have that subdataset changed?

kyleam:
@yarikoptic
> A door of combinatorics is slightly opened... I forgot - did we discuss many be the `--force autodatatransfer` as an approach to trigger it instead?

I think it was discussed, in the sense that it's exactly what gh-4541 proposed IIUC.

@mih
> For each of them a force... interpretation is straightforward [...]

This presentation seems clear enough to me.  The value that is a bit of an oddball in my eyes is "no-datatransfer" because, unlike every other option, it feels like it's going in the opposite direction relative to the default.  But when taken with the initial description "force particular operations, overruling automatic decision making", it fits, I think.

> `--auto` is different. Whether or not to add `--auto` isn't that exclusive as far as I can see. It is more like "if you happen to attempt a copy, do it like this".

This makes more sense to me than trying to shoehorn `--auto` into `--force` (as in gh-4541)

> Is this secure enough (see [gh-4357](https://github.com/datalad/datalad/issues/4357)). A dataset could set this flag in its own config (.datalad/config). Are we confident that it would never override a user config (add test)?

Thanks for raising this issue and adding the test.  If we're worried about something coming in via `.datalad/config`, I guess a similar worry could apply to `numcopies` via the git-annex branch.  Given `push` is following `publish`s narrow preferred-content-only view of `--auto`, perhaps we should add an `--numcopies` when `--auto` is used.  But the same is true for `publish`, so I think this can be handled later.

datalad/core/distributed/push.py
```diff
@@ -767,11 +771,14 @@ def _push_data(ds, target, content, force, jobs, res_kwargs,
     if jobs:
         cmd.extend(['--jobs', str(jobs)])
 
-    if not to_transfer and force not in ('all', 'datatransfer'):
-        lgr.debug("Invoking copy --auto")
-        cmd.append('--auto')
+    if force not in ('pushall', 'datatransfer') and ds_repo.config.obtain(
```
kyleam, line 774: nit-pick: Why not use `getbool` here?

datalad/core/distributed/push.py
```diff
@@ -64,29 +65,32 @@
...
+    to limit a push to changes relative to a particular point in the version
+    history of a dataset (e.g. a release tag) using the
+    [CMD: --since CMD][PY: since PY] option in conjunction with the
+    specification of a reference dataset. Changes can also be evaluated
+    recursively, i.e. only those subdatasets are pushed where a change was
+    recorded that is reflected in the current state of the top-level reference
+    dataset.
+
+    Which files are copied can be further tailored via the
+    'datalad.push.copy-auto-if-wanted' configuration. If set, push will test
```
kyleam, line 83: Should an entry be added to common_cfg.py?

datalad/core/distributed/tests/test_push.py
```diff
@@ -572,3 +572,58 @@ def test_gh1811(srcpath, clonepath):
...
+    # check that wanted is obeyed, if instructed by configuration
+    src.config.set('datalad.push.copy-auto-if-wanted', 'true', where='local')
+    res = src.push(to='target')
+    assert_in_results(
+        res, action='copy', path=str(src.pathobj / 'data.0'), status='ok')
+    for p in ('secure.1', 'secure.2'):
+        assert_not_in_results(res, path=str(src.pathobj / p))
+    assert_status('notneeded', src.push(to='target'))
+
+    # check that dataset-config cannot overrule this
```
yarikoptic, line 610: from this comment I thought you would do something like

datalad/core/distributed/push.py
yarikoptic, line 78: Cannot do multiline change suggestion, so here is those two lines above modified
```
    specification of a reference dataset. In recursive mode subdatasets will also be
    evaluated, and only those subdatasets are pushed where a change was
```

yarikoptic:
TL;DR: To "do the right thing" `push` should just follow the `publish` behavior and just add `--auto` without requiring any additional config whenever `wanted` is set for the sibling.  That will play nicely with `--force datatransfer` (and other forces such as `gitpush`) which just would not add `--fast` to annex invocation and just do data transfer regardless current knowledge within git-annex branch.  Additional `--force no-autodatatransfer` ideally could be added to disable considering `wanted`.

Full grumpy Russian on friday version: I had just closed the #4551 (since I think default push behavior is to stay), and thought to open a new issue about autosetting the `datalad.push.copy-auto-if-wanted` by `create-sibling` and friends, but then noticed again this phrase of yours:

> I found that it is better and useful to be able to say `datalad push` and get what is "right" for a particular dataset...

with which I 100% agree!  But the "devil is in detail" and smart Joey foresaw it so really it should have been 

> I found that it is better and useful to be able to say `datalad push` and get what is "right" for a particular **sibling**.

In our use cases siblings correspond to the same dataset (but they do not have to) so really it becomes a combination of "dataset and sibling", but definitely not just dataset since the "right thing" could vary between siblings. And this situation become particularly interesting in case of also using `datalad-publish-depends` with multiple siblings (e.g. private and public... I thought you are getting to that land some time soon in metalad ;-)).

And that is where we could just can keep using git annex `wanted` setting which you set for **that particular sibling** to "do the right thing" when desired/necessary.  That would not be achievable with a suggested config setting which is not tuned per sibling.  Moreover with any additional config setting (e.g. `remote.<sibling>.datalad-push-auto` to go alongside with `datalad-publish-depends`) you would require "data migration" in existing datasets to establish the config setting,  adjust `create-sibling*` to set it if created with `wanted` (via option or by inheriting). 

To me it sounds like a really complex way to make it even possible to "do the right thing".

That is why I feel the need to strongly oppose introducing any new config setting to just to not use already available native "git annex" mechanism to allow those who want to "to do the right thing".   I do not remember why, besides some purity, `publish`-like behavior of treating `wanted` was not considered before, but I do not see immediate shortcomings.

My only other concern before I could even just try `push` on `///` is #4551 since I cannot afford "push all data by default", and I will run into cases where I will forget to set `wanted` to `nothing`.  But may be it is somewhat for the best - having an explicit `wanted` would be a good thing ... But `push` new behavior is a 180 degree turn in the paradigm on what "the right thing" by default is: to ensure that you have content published/pushed (push) vs to **intentionally** have content published/pushed (publish). In my cases I just got used to have my legal ass covered instead of assuring that it is all out there.  But may be I am just too easy to get paranoid, after all `git annex copy --to=remote` just copies whenever no paths are provided).  And now I think I recall a debate we had years back in `publish` birth about the default of `--transfer-data=none` and me arguing the other way around; so overall I think we are converging ;-)

  -- Yarik, (almost) Summer 2020

yarikoptic:
<details>
<summary>appveyor failure (from running post-update) is "legit" but I do not immediately see how it relates to this very test case (haven't looked at any appveyor failures recently):</summary> 

```shell
ERROR: datalad.core.distributed.tests.test_push.test_push_wanted
  File "C:\Miniconda35\envs\test-environment\lib\site-packages\nose\case.py", line 198, in runTest
    self.test(*self.arg)
  File "C:\Miniconda35\envs\test-environment\lib\site-packages\datalad\tests\utils.py", line 714, in newfunc
  File "C:\Miniconda35\envs\test-environment\lib\contextlib.py", line 77, in __exit__
    self.gen.throw(type, value, traceback)
  File "C:\Miniconda35\envs\test-environment\lib\site-packages\datalad\utils.py", line 1794, in make_tempfile
    yield filename
  File "C:\Miniconda35\envs\test-environment\lib\site-packages\datalad\tests\utils.py", line 714, in newfunc
  File "C:\Miniconda35\envs\test-environment\lib\site-packages\datalad\tests\utils.py", line 714, in newfunc
  File "C:\Miniconda35\envs\test-environment\lib\contextlib.py", line 77, in __exit__
    self.gen.throw(type, value, traceback)
  File "C:\Miniconda35\envs\test-environment\lib\site-packages\datalad\utils.py", line 1794, in make_tempfile
    yield filename
  File "C:\Miniconda35\envs\test-environment\lib\site-packages\datalad\tests\utils.py", line 714, in newfunc
  File "C:\Miniconda35\envs\test-environment\lib\site-packages\datalad\core\distributed\tests\test_push.py", line 599, in test_push_wanted
    name='target',
  File "C:\Miniconda35\envs\test-environment\lib\site-packages\datalad\distribution\dataset.py", line 498, in apply_func
    return f(**kwargs)
  File "C:\Miniconda35\envs\test-environment\lib\site-packages\datalad\interface\utils.py", line 494, in eval_func
  File "C:\Miniconda35\envs\test-environment\lib\site-packages\datalad\interface\utils.py", line 482, in return_func
    results = list(results)
  File "C:\Miniconda35\envs\test-environment\lib\site-packages\datalad\interface\utils.py", line 469, in generator_func
datalad.support.exceptions.IncompleteResultsError: Command did not complete successfully [{'message': ('failed to run post-update hook under remote path %s (%s)', 'C:\\Users\\appveyor\\AppData\\Local\\Temp\\1\\datalad_temp_3jb32evv', 'CommandError: \'"cd ""C:\\Users\\appveyor\\AppData\\Local\\Temp\\1\\datalad_temp_3jb32evv\\.git"" && ( [ -x hooks/post-update ] && hooks/post-update || : )"\' failed with exitcode 1 [err: \'\':\' is not recognized as an internal or external command,\r\noperable program or batch file.\'] [cmd.py:run:930]'), 'status': 'error', 'action': 'create_sibling', 'path': 'C:\\Users\\appveyor\\AppData\\Local\\Temp\\1\\datalad_temp_hh9cpvut', 'type': 'dataset', 'orig_request': 'C:\\Users\\appveyor\\AppData\\Local\\Temp\\1\\datalad_temp_hh9cpvut', 'refds': 'C:\\Users\\appveyor\\AppData\\Local\\Temp\\1\\datalad_temp_hh9cpvut', 'raw_input': True}]
```
</details>

mih:
Given that there still seems to be a need for further discussion, I'll just leave it here till the next call. It seems my take-away from the last call "we do not want what publish does with auto, but we want users to keep doing that with push" was inadequate. I suggest we have a *write-up* of the agreed solution to this series of issues as the first agenda item for the next call.

datalad/core/distributed/push.py
mih: I dislike the compatibility interfaces that require splitting the config key into section and option name.

datalad/core/distributed/push.py
mih: Will do.

datalad/core/distributed/push.py
mih: Will adopt this wording.

datalad/core/distributed/push.py
mih: Done in 5421cdc730ea58d25ed1dff9cacc0c2e5152e504 Thx for the motivation!

mih:
I made the changes we talked about today. This is ready to merge from my POV (pending tests).

mih:
I will merge this now. This PR fixes legitimate issues in the current `master`. There are various issues that question the concept of `push` in a fundamental fashion (#4483 #4602 #4603 #4604), it won't really make a difference whether the changes in here are added to master, when conceptual changes to the API will be made after a conclusion was found. But right now, it will improve the situation in master, as it stops an objectively faulty behavior with `--auto`.

With respect to #4605 (`wanted` vs `numcopies`) when my conclusion in https://github.com/datalad/datalad/issues/4605#issuecomment-640010274 turns out to be valid, this change won't affect it. If we decide to override `numcopies` in `publish/push` this PR also doesn't cause additional problems.

PR #4583, opened 2020-05-28
RM deprecated (promoted to .core.local.run) in 0.12 release interface.run and saver option

yarikoptic:
- datalad-container is expected to fail, PR there https://github.com/datalad/datalad-container/pull/120


yarikoptic:
agree in general (thus did not label it for 0.13.0 milestone although said "likely" in containers PR ;) ). Will mark for 0.14.0. 

> And in general there is no reason this needs to be removed right now.

My selfish rationale for removal of those deprecated interfaces - pycharm allows to quickly open an file by starting to type its name (like `run`).  Having multiple `run`, `diff`, etc across code base complicates navigation etc.  Since I think we all would agree that they should be removed, I did not want to delay submitting a PR since I would have just forgotten and it would lurk around for longer.

yarikoptic:
Logistics wise:  should we establish `next` or `0.14` branch to which we merge such changes? Otherwise we would need to keep such PRs open until we decide that master is seeing no new features and destined for the new "major" release. IMHO version based (0.14) is better but I am ok with `next`.

Issue #3942, opened 2019-12-17
"helpme" reports from CI


yarikoptic:
To not distract @vsoch from the main goal of https://github.com/datalad/datalad/issues/3649 (user oriented reporting), filing this idea extending on #3649 separately.  And it might not be worth it... 

I just ran into the use case, where "token based" reporting (with similar fingerprinting to #3649) could be useful - to automagically collect CI test errors (sample: https://github.com/datalad/datalad/issues/3647) (primary interest CI of main branches, not PRs)
- reports would point to build logs
- smart helper could have somehow figure out previous similar run for the branch which succeeded, and provide diff on code and WTF thus potentially identifying changes which lead to the crash 
- It would allow for earl(ier) detection of new issues or regressions (e.g. we have no clue when #3647 really came about and since it is a fluke, hard to tell)
- In our case it is still `nose` though so not sure how (in terms of longevity etc) it would be worthwhile to develop a plugin to collect those failures automagically (with back references to the specific jobs).

vsoch:
@yarikoptic just for FYI here - the implementation I'm doing will work either way with or without a token. If a token is found, it will use it. If not, then it will open the browser. So likely this wouldn't be too hard to do, we would just give the client a token before calling the same flow. I'm working on the dummy example now - I'm going to add a simpler version of the wtf plugin you have to helpme so that (by default) non-datalad folks can have some similar metrics. Will report back when I have something to look at! It's in the rseng organization I just created for small, open source research software engineering projects so I'll invite y'all too! :D

yarikoptic:
> If a token is found, it will use it. If not, then it will open the browser.

I think ideally at least there should be an option to demand specific way and disable altogether.  For user-visible executions, if session is interactive - we would want "browser" version (no token based - users should be the one to press "Submit issue"); for CI -- token based, if no token or disabled - we would just log message that no "helpme" submission since there is no token and/or session is not interactive.

vsoch:
I can add a variable to the client to enforce it (or not) with default to not enforce, and the calling function can then require that it's provided, and default will just work for the general user.

Issue #3649, opened 2019-09-03
Add feedback submission to cmdline crash handler


mih:
Both @yarikoptic and I had the impression that we do not get enough feedback on cases where datalad doesn't work as intended and crashes. One possibility to mitigate this situation would be to add an automated feedback submission ability to the cmdline interface's crash handler. It could format a document with:

- sys.argv
- traceback of the crash
- wtf output

Given that such a report could contain sensitive information, we should allow users to scan/edit such a report before it is sent (unless configured to be OK).

Maybe there is already some tool that would facilitate the server-side of things.

yarikoptic:
https://github.com/con/communitator was my quest for the tool which would take care about such specs. Helpme from Vanessa was the only one suggested

bpoldrack:
Might be worth considering to also include (and allow to edit of course) all relevant config files. May be even some structural info on the dataset the command ran in (like super-/subdatasets).

vsoch:
This would be very doable with helpme (sorry didn't see this when it was posted). For GitHub, it would require a user token, which is a likely barrier in that a user wouldn't readily want to generate and provide one.

Even if we set up a post to discourse (askci or neurostars or similar) with tag / topic of "datalad" we would still require the user to accept (on first issue) the OAuth screen.

Or a third option - a custom datalad submitter that sends data somewhere else?

yarikoptic:
> For GitHub, it would require a user token

oh... that reminded me about https://github.com/vsoch/helpme/issues/44  -- IMHO no token should be necessary, the full form submission I believe could be crafted via URL.  Then any user could submit a PR

yarikoptic:
> Even if we set up a post to discourse (askci or neurostars or similar) with tag / topic of "datalad" we would still require the user to accept (on first issue) the OAuth screen.

yes, neurostars would be the one, and I think it is ok for a user to accept OAuth screen if really needed (i.e. couldn't be done similarly via passing the whole body via url)

vsoch:
@yarikoptic that's definitely doable! The url would open and then (hopefully) the user would be logged in to submit.

Would it be okay with you to add a custom datalad helper for this purpose? The official github integration happens with the token (more progamatically) and I wouldn't want to fuss around with that.

Or are you more interested in submission to neurostars?

vsoch:
Going for a quick run, back in a bit!

vsoch:
[...]

PR #4586, opened 2020-05-28
Adding cmdline support submission with helpme

vsoch:
This pull request will address #3649 and #3942 , specifically adding the helpme client (used with GitHub) to support handling the "other errors not known" at the end of the except statements in the datalad/cmdline/main.py.  I developed this interactively and used a different error that I could trigger with:

```bash
datalad --cmd search
```

I wasn't sure what kind of real use case would lead to the last error, so I was only able to get there by commenting out the other exceptions. It would be good to have a few examples to trigger the error so I can test the final thing **and** add an example to the user docs. Interactively, the example issue is here https://github.com/datalad/datalad-helpme/issues/4. 

## Fixes
 - #3649 
 - #3942

## Changes
Changes include:

 - a reporting.rst documentation page to describe helpme and how it works with datalad
 - a submit_helpme function defined in datalad/cmdline/support.py, which doesn't trigger an error if helpme isn't installed (just skips over)
 - adding helpme to install under misc (note that the github and base of helpme only requires requests, which is already required by datalad)
 - an envar `DATALAD_HELPME_DISABLE` that the user can set to disable it.
 - instructions for exporting `HELPME_GITHUB_TOKEN` if they don't want the interactive submission.

Signed-off-by: vsoch <vsochat@stanford.edu

### DIFF
--- a/datalad/cmdline/helpers.py
+++ b/datalad/cmdline/helpers.py
@@ -255,6 +255,7 @@ def get_repo_instance(path=curdir, class_=None):
     else:
         raise RuntimeError("No datalad repository found in %s" % abspath_)
 
+
 from appdirs import AppDirs
 from os.path import join as opj
 
--- a/datalad/cmdline/helpers.py
+++ b/datalad/cmdline/helpers.py
@@ -255,7 +255,6 @@ def get_repo_instance(path=curdir, class_=None):
     else:
         raise RuntimeError("No datalad repository found in %s" % abspath_)
 
-
 from appdirs import AppDirs
 from os.path import join as opj
 
--- a/datalad/cmdline/main.py
+++ b/datalad/cmdline/main.py
@@ -560,15 +560,18 @@ def main(args=None):
                 # had no code defined
                 sys.exit(exc.code if exc.code is not None else 1)
             except Exception as exc:
-                identifier = generate_datalad_identifier(
-                    stack = traceback.extract_stack(), exc=exc,
-                )
-                submit_helpme(
-                    title = str(exc),
-                    tb = traceback.format_exc(),
-                    identifier = identifier,
-                    detail = " ".join(map(quote_cmdlinearg, sys.argv)),
-                )
+
+                # If the user requests to disable, or in testing environment don't submit
+                if os.environ.get("DATALAD_HELPME_DISABLE") is None:
+                    identifier = generate_datalad_identifier(
+                        stack = traceback.extract_stack(), exc=exc,
+                    )
+                    submit_helpme(
+                        title = str(exc),
+                        tb = traceback.format_exc(),
+                        identifier = identifier,
+                        detail = " ".join(map(quote_cmdlinearg, sys.argv)),
+                    )
                 lgr.error('%s (%s)' % (exc_str(exc), exc.__class__.__name__))
                 sys.exit(1)
     else:
--- a/datalad/cmdline/main.py
+++ b/datalad/cmdline/main.py
@@ -30,6 +30,7 @@ from datalad.support.exceptions import InsufficientArgumentsError
 from datalad.support.exceptions import IncompleteResultsError
 from datalad.support.exceptions import CommandError
 from .helpers import strip_arg_from_argv
+from .support import submit_helpme
 from ..utils import (
     assure_unicode,
     chpwd,
@@ -557,6 +558,8 @@ def main(args=None):
                 # had no code defined
                 sys.exit(exc.code if exc.code is not None else 1)
             except Exception as exc:
+                title = "DataLad Cmd Error: %s" % exc.__class__.__name__
+                submit_helpme(title = title, detail = exc_str(exc))
                 lgr.error('%s (%s)' % (exc_str(exc), exc.__class__.__name__))
                 sys.exit(1)
     else:
--- a/datalad/cmdline/main.py
+++ b/datalad/cmdline/main.py
@@ -560,9 +560,8 @@ def main(args=None):
                 # had no code defined
                 sys.exit(exc.code if exc.code is not None else 1)
             except Exception as exc:
-                title = "DataLad Cmd Error: %s" % exc.__class__.__name__
                 submit_helpme(
-                    title=title,
+                    title="[helpme] %s" % str(exc),
                     traceback=traceback.format_exc(),
                     detail=" ".join(map(quote_cmdlinearg, sys.argv))
                 )
--- a/datalad/cmdline/main.py
+++ b/datalad/cmdline/main.py
@@ -502,9 +502,6 @@ def main(args=None):
     # enable overrides
     datalad.cfg.reload(force=True)
 
-    import IPython
-    IPython.embed()
-
     if cmdlineargs.change_path is not None:
         from .common_args import change_path as change_path_opt
         for path in cmdlineargs.change_path:
--- a/datalad/cmdline/main.py
+++ b/datalad/cmdline/main.py
@@ -20,6 +20,7 @@ import argparse
 from collections import defaultdict
 import sys
 import textwrap
+import traceback
 import os
 
 
@@ -559,7 +560,11 @@ def main(args=None):
                 sys.exit(exc.code if exc.code is not None else 1)
             except Exception as exc:
                 title = "DataLad Cmd Error: %s" % exc.__class__.__name__
-                submit_helpme(title = title, detail = exc_str(exc))
+                submit_helpme(
+                    title = title, 
+                    traceback = traceback.format_exc(),
+                    detail = ' '.join(sys.argv)
+                )
                 lgr.error('%s (%s)' % (exc_str(exc), exc.__class__.__name__))
                 sys.exit(1)
     else:
--- a/datalad/cmdline/main.py
+++ b/datalad/cmdline/main.py
@@ -31,7 +31,7 @@ from datalad.support.exceptions import InsufficientArgumentsError
 from datalad.support.exceptions import IncompleteResultsError
 from datalad.support.exceptions import CommandError
 from .helpers import strip_arg_from_argv
-from .support import submit_helpme
+from .support import submit_helpme, generate_datalad_identifier
 from ..utils import (
     assure_unicode,
     chpwd,
@@ -502,6 +502,9 @@ def main(args=None):
     # enable overrides
     datalad.cfg.reload(force=True)
 
+    import IPython
+    IPython.embed()
+
     if cmdlineargs.change_path is not None:
         from .common_args import change_path as change_path_opt
         for path in cmdlineargs.change_path:
@@ -560,10 +563,14 @@ def main(args=None):
                 # had no code defined
                 sys.exit(exc.code if exc.code is not None else 1)
             except Exception as exc:
+                identifier = generate_datalad_identifier(
+                    stack = traceback.extract_stack(), exc=exc,
+                )
                 submit_helpme(
-                    title="[helpme] %s" % str(exc),
-                    traceback=traceback.format_exc(),
-                    detail=" ".join(map(quote_cmdlinearg, sys.argv))
+                    title = str(exc),
+                    tb = traceback.format_exc(),
+                    identifier = identifier,
+                    detail = " ".join(map(quote_cmdlinearg, sys.argv)),
                 )
                 lgr.error('%s (%s)' % (exc_str(exc), exc.__class__.__name__))
                 sys.exit(1)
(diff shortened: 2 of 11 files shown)


docs/source/reporting.rst
```diff
@@ -0,0 +1,68 @@
...
+
+And then the browser will open to the issue repository, where you can write more content
+into the issue. Using the standard template, useful information about your system 
+and libraries is shared so you don't need to do it. You can look at a sample_issue_
+for an example of what is shared.
+
+If you need a completely headless and non-interactive interaction, then you can
+simply export a ``HELPME_GITHUB_TOKEN`` to the environment and it will
+be discovered, and use the GitHub API to open the same issue. See the helpme_
+documentation for more details.
```
yarikoptic, line 53: Note to ourselves: we could probably integrate with our handling of GitHub token credentials

datalad/cmdline/helpers.py
```diff
@@ -255,7 +255,6 @@ def get_repo_instance(path=curdir, class_=None):
     else:
         raise RuntimeError("No datalad repository found in %s" % abspath_)
 
-
```
yarikoptic, line 258: Unrelated change ideally to be reverted (two empty lines too separate at this level I believe)

datalad/cmdline/support.py
```diff
@@ -0,0 +1,85 @@
...
+**WTF Output** :open_file_folder:
+%s""" % (
+                detail or "",
+                cmo.out,
+            )
+
+        # Submit the issue
+        issue = helper.run_headless(repo=repo, body=body, title=title)
+
+    except:
```
yarikoptic, line 84: Worth catching and logging at debug level this exception

datalad/cmdline/support.py
```diff
@@ -0,0 +1,85 @@
...
+        "configuration",
+        "location",
+        "extensions",
+        "metadata_extractors",
+        "dependencies",
+        "dataset",
+    ]
+
+    try:
+        from helpme.main import get_helper
```
yarikoptic, line 57: That is where to catch import error,  suggest to install it with an lgr.info message, and then return

datalad/cmdline/support.py
```diff
@@ -0,0 +1,85 @@
...
+import sys
+import os
+
+from logging import getLogger
+
+lgr = getLogger("datalad.cmdline")
+
+os.environ["MESSAGELEVEL"] = "QUIET"
+os.putenv("MESSAGELEVEL", "QUIET")
+
```
yarikoptic, line 25: Why is this process wide upon import change of this environment variables is needed? Ideally should be avoided, or if really needed - a comment why it is needed is needed

datalad/cmdline/support.py
```diff
@@ -0,0 +1,85 @@
...
+def submit_helpme(detail=None, title=None):
+    """Submit a request to the datalad-helpme repository at
+       https://github.com/datalad/datalad-helpme. If helpme isn't installed,
+       we skip this step. The basic submission includes the entire grab from
+       wtf (capturing system and library information) and an optional message.
+    """
+    title = title or default_title
+
+    # If the user requests to disable, or in testing environment don't submit
+    if os.environ.get("DATALAD_HELPME_DISABLE") is not None:
```
yarikoptic, line 39: Note more to ourselves: make it defined in https://github.com/datalad/datalad/blob/master/datalad/interface/common_cfg.py#L373 and use global cfg.getbool here, then it would allow setting it in global or user gitconfig

datalad/cmdline/support.py
```diff
@@ -0,0 +1,85 @@
...
+os.environ["MESSAGELEVEL"] = "QUIET"
+os.putenv("MESSAGELEVEL", "QUIET")
+
+repo = "datalad/datalad-helpme"
+default_title = "Test issue opened manually by helpme"
+
+
+def submit_helpme(detail=None, title=None):
+    """Submit a request to the datalad-helpme repository at
+       https://github.com/datalad/datalad-helpme. If helpme isn't installed,
```
yarikoptic, line 32: Very minor: use that repo variable here?

datalad/cmdline/support.py
```diff
@@ -0,0 +1,85 @@
...
+        "dataset",
+    ]
+
+    try:
+        from helpme.main import get_helper
+
+        helper = get_helper(name="github")
+
+        with swallow_outputs() as cmo:
+            wtf(decor="html_details", sections=sections)
```
yarikoptic, line 62: Note to myself: check if we could just return rendered value avoiding any swallowing

docs/source/reporting.rst
```diff
@@ -0,0 +1,68 @@
...
+separately:
+
+.. code-block:: console
+
+    $ pip install helpme[github]
+
+
+To submit to GitHub, HelpMe only requires requests, which is already required by
+Datalad, so the install above will not add any additional dependencies other
+than the helpme library itself. If you don't choose to
```
yarikoptic, line 27: If there is no additional dependencies if helpme, I would be ok to make it party of the default installation. Let's see what others say though

setup.py
```diff
@@ -52,6 +52,7 @@
         'PyGithub',          # nice to have
     ],
     'misc': [
+        'helpme[github]',    # support for helpme error submission
```
yarikoptic, line 55: Let's put current version of helpme as minimally required

datalad/cmdline/support.py
vsoch: Do you want some called of the function to be able to control the repo? I almost put it there but changed my mind because it seemed more like a global / non-changeable variable.

datalad/cmdline/support.py
vsoch: helpme has a message level controlled via environment, and that's how you disable warnings. if you don't put that there will be an ugly WARNING no GitHub token.... every time the helpme_submit function is used.

vsoch: I'll add a note.

vsoch:
Quick update - here is the issue submission with traceback, command, and identifier added https://github.com/datalad/datalad-helpme/issues/5. The "this issue is already open" functionality needs to be handled via a GitHub action.

datalad/cmdline/helpers.py
vsoch: Fixed locally.

datalad/cmdline/support.py
vsoch: Fixed locally.

datalad/cmdline/support.py
vsoch: Fixed locally.

datalad/cmdline/support.py
vsoch: I'll do the same as I did for title.

setup.py
vsoch: got it.

datalad/cmdline/support.py
yarikoptic: Then it should be patched only around helpme invocation . I can do that later. I would also recommend in helpme to prepend it with HELPME_ since it sounds to be specific to it

datalad/cmdline/support.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+def submit_helpme(title=None, traceback="", detail="", identifier=None, repo=None):
+    """Submit a request to the datalad-helpme repository at
+       https://github.com/datalad/datalad-helpme. If helpme isn't installed,
+       we skip this step. The basic submission includes the entire grab from
+       wtf (capturing system and library information) and an optional message.
+
+       Arguments:
+        - title (str)     : the title for the GitHub issue
```
yarikoptic, line 40: Just spotted it - will need to be adjusted to follow numpy docstring formatting as everywhere else in the code base, I can do that too later

datalad/cmdline/support.py
```diff
@@ -0,0 +1,105 @@
...
+        - traceback (str) : the full traceback
+        - identifier (str): the identifier string (will use traceback if not defined)
+        - repo (str)      : GitHub repo (<username>/<repo>) to submit to.
+    """
+    title = title or default_title
+    repo = repo or default_repo
+
+    # If no identifier defined, use traceback
+    if identifier is None:
+        identifier = traceback
```
yarikoptic, line 51: Is there a special meaning of an empty string identifier?

datalad/cmdline/support.py
vsoch: Yeah that was bothering me too :) I can do a quick release for that and I'll update the PR here.

datalad/cmdline/main.py
```diff
@@ -557,6 +559,12 @@ def main(args=None):
                 # had no code defined
                 sys.exit(exc.code if exc.code is not None else 1)
             except Exception as exc:
+                title = "DataLad Cmd Error: %s" % exc.__class__.__name__
+                submit_helpme(
+                    title = title, 
+                    traceback = traceback.format_exc(),
+                    detail = ' '.join(sys.argv)
```
yarikoptic: note 2 @vsoch : pushed minor tune up here (see e.g. deecbc0378588e066609e7442385763c77a0bf15)

datalad/cmdline/support.py
vsoch: okay pushing update shortly.

datalad/cmdline/support.py
vsoch: I can give this a shot at least.

datalad/cmdline/support.py
vsoch: We check `if identifier is not None` so it can't be an empty string.

vsoch:
@yarikoptic I made a helpme release that honors the environment namespace with `HELPME_MESSAGELEVEL` and moved the setting down into the function, also made an attempt at the docstring. https://github.com/datalad/datalad/pull/4586/commits/11551e783649bac7ecc08aafeeca1daf58576cc5

vsoch:
okay I'm peacing out for today - let me know next steps! I think it would be good to know some commands that I could do to trigger this "not understood error" so we can test helpme. I think we need to do:

 - test with GitHub token
 - test with GitHub token via GitHub action (so it can run with the CI)
 - test to open issue on datalad-helpme
 - test GitHub workflow for datalad-helpme https://github.com/datalad/datalad-helpme/pull/7

Peeeace! and Peas.

datalad/cmdline/main.py
```diff
@@ -557,6 +560,12 @@ def main(args=None):
                 # had no code defined
                 sys.exit(exc.code if exc.code is not None else 1)
             except Exception as exc:
+                title = "DataLad Cmd Error: %s" % exc.__class__.__name__
+                submit_helpme(
```
yarikoptic, line 564: Myself: need to check if actually current UI  is interactive (UI while running tests is not, so that would avoid setting envvar in Cis). We might actually make env var such that it could force helpme in noon interactive sessions to assist filling issues in such cases

yarikoptic: And ideally in the longer run would be great if helpme could use our ui interface to ask questions etc

datalad/cmdline/support.py
yarikoptic: Checked, can't do ATM but we look into making it possible

datalad/cmdline/main.py
```diff
@@ -557,6 +560,12 @@ def main(args=None):
                 # had no code defined
                 sys.exit(exc.code if exc.code is not None else 1)
             except Exception as exc:
+                title = "DataLad Cmd Error: %s" % exc.__class__.__name__
```
yarikoptic, line 563: I think we can strip this coming prefix (no unique info), add a new label eg helpme or autofiled, and include instead limited in length str(exc) .

vsoch: Fixed locally.

datalad/cmdline/main.py
vsoch: Definitely! We can handle this in another scoped PR / issue.

vsoch:
@yarikoptic I've updated helpme to allow using a custom generated identifier, and figured out how to extract a list of (non datalad) and datalad functions, and updated here to generate hashes for each, along with the exception name. See the dummy issue at https://github.com/datalad/datalad-helpme/issues/9. Once we have decided that this identifier / format looks okay I'll update the github workflow PR on the datalad-helpme respository to handle parsing it.

mih:
Closing this after 6+ months of inactivity. I have placed an agenda item for next week's developer call to discuss a future of this feature. It certainly is not desirable to have it sit in an open PR for months.

PR #4591, opened 2020-05-30
ENH: (+minor DOC fix) pretty print IncompleteResultsError

yarikoptic:
Someone recently had asked me why we are not having "multiline" exception renderings... and I forgot what actually was the reason for us (may be if only for logging) why we leaned toward one liners.

I think the change is trivial enough to be submitted against maint.

    ENH: Use pprint.pformat for IncompleteResultsError str
    
    Those are are now more rare but when they come, they become hard to parse visually,
    see e.g. one in https://github.com/datalad/datalad/pull/4581#issuecomment-636239700 .
    With pprint they become much more human friendly/informative even though might occupy
    considerable amount of terminal output space.
    
    E.g. here is the exception from that issue pprinted
    
            ERROR: test.test
              File "/usr/lib/python3/dist-packages/nose/case.py", line 197, in runTest
                    self.test(*self.arg)
              File "/tmp/testds/test.py", line 5, in test
                    raise IncompleteResultsError(
            datalad.support.exceptions.IncompleteResultsError: None [{'action': 'create_sibling',
              'message': ('failed to run post-update hook under remote path %s (%s)',
                                      'C:\\Users\\appveyor\\AppData\\Local\\Temp\\1\\datalad_temp_3jb32evv',
                                      'CommandError: \'"cd '
                                      '""C:\\Users\\appveyor\\AppData\\Local\\Temp\\1\\datalad_temp_3jb32evv\\.git"" '
                                      '&& ( [ -x hooks/post-update ] && hooks/post-update || : )"\' '
                                      "failed with exitcode 1 [err: '':' is not recognized as an "
                                      'internal or external command,\r\n'
                                      "operable program or batch file.'] [cmd.py:run:930]"),
              'orig_request': 'C:\\Users\\appveyor\\AppData\\Local\\Temp\\1\\datalad_temp_hh9cpvut',
              'path': 'C:\\Users\\appveyor\\AppData\\Local\\Temp\\1\\datalad_temp_hh9cpvut',
              'raw_input': True,
              'refds': 'C:\\Users\\appveyor\\AppData\\Local\\Temp\\1\\datalad_temp_hh9cpvut',
              'status': 'error',
              'type': 'dataset'}]
    
    it has not become the prettiest thing, BUT it became much more approachable

### DIFF
--- a/datalad/support/exceptions.py
+++ b/datalad/support/exceptions.py
@@ -320,8 +320,10 @@ class IncompleteResultsError(RuntimeError):
     # such results have been yielded already at the time this exception is
     # raised, little point in collecting them just for the sake of a possible
     # exception
-    # MIH: AnnexRepo is the last remaining user of this functionality, in a
-    # single context
+    # MIH+YOH: AnnexRepo.copy_to and @eval_results are the last
+    # remaining user of this functionality.
+    # General use (as in AnnexRepo) of it discouraged but use in @eval_results
+    # is warranted
     def __init__(self, results=None, failed=None, msg=None):
         super(IncompleteResultsError, self).__init__(msg)
         self.results = results
--- a/datalad/support/exceptions.py
+++ b/datalad/support/exceptions.py
@@ -11,7 +11,7 @@
 
 import re
 from os import linesep
-
+from pprint import pformat
 
 
 class CommandError(RuntimeError):
@@ -332,7 +332,7 @@ class IncompleteResultsError(RuntimeError):
     def __str__(self):
         super_str = super(IncompleteResultsError, self).__str__()
         return "{} {}" \
-               "".format(super_str, self.failed)
+               "".format(super_str, pformat(self.failed))
 
 class InstallFailedError(CommandError):
     """Generic exception to raise whenever `install` command fails"""
--- a/datalad/support/exceptions.py
+++ b/datalad/support/exceptions.py
@@ -331,8 +331,14 @@ class IncompleteResultsError(RuntimeError):
 
     def __str__(self):
         super_str = super(IncompleteResultsError, self).__str__()
-        return "{} {}" \
-               "".format(super_str, pformat(self.failed))
+        return "{}{}{}".format(
+            super_str,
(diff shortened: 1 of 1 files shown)


yarikoptic:
I know that it is very minor and not urgent etc, so feel free to demote from 0.13.0

yarikoptic:
and only now mentioned that it is actually even against `maint`... could go in either I guess.  Just would appreciate feedback ;-)

datalad/support/exceptions.py
```diff
@@ -330,7 +332,7 @@ def __init__(self, results=None, failed=None, msg=None):
     def __str__(self):
         super_str = super(IncompleteResultsError, self).__str__()
         return "{} {}" \
```
bpoldrack, line 334: May be then also make it `"{}{}{}".format(super_str, os.linesep, pformat(self.failed))` at least?
Still separating both by a space seems a bit weird if it's not a one-line output anymore. In your example output I find the `None [ ...` completely unclear wrt what it's trying to tell me.

yarikoptic: Re None - note that I am only changing formatting of failed, and that None is the absent message given by super. I am only trying to make exception message more human parsable, not absolve all of its sins ;-)

bpoldrack: Yes, I'm aware. What I'm saying is, that if we generate mutliline output anyway, that it's weird to have that one separated by a space only, suggesting some connection to what follows in that line.

yarikoptic: Yeap yeap - I am putting my pants on already and will go downstairs to my laptop to tune it all up. I agree, it would be useful to provide better context of we are to improve it, but then just a new line probably wouldn't suffice

yarikoptic:
ok, pushed my tune up. See e1951ce3ab6e9b8f7b9aaea1b1482e4e95ce22b5 for an example of rendering etc

bpoldrack:
Tune up LGTM. Fail to see why Travis didn't pick it up, though.

Issue #2054, opened 2017-12-22
Consts should likely not define dir paths including '.git'


mih:
#### What is the problem?

In `consts.py` we define stuff like this:

```
DATALAD_GIT_DIR = join('.git', 'datalad')
ANNEX_TEMP_DIR = join('.git', 'annex', 'tmp')
```

I think this is inappropriate, because `.git` can easily be a file. That is why we have `get_git_dir(path)` that will report the actual location of `.git/`, possibly by reading it from a .git file. Such constants should probably just identify the name inside the `.git/` directory.

yarikoptic:
Good find
But do we support at all working in those repos with .git being a file? For submodules we assure it is not so the second one should be ok since iirc otherwise annex wouldn't work at all in Independence mode...
But probably indeed better to remove dependency on such assumptions where possible

yarikoptic:
marked it as a target for 1.0, but note that `.consts` aren't a part of the core (at least not yet to any part of it) ;)

mih:
Today, the only place of `DATALAD_GIT_DIR` usage is other consts:

```
datalad/consts.py:ARCHIVES_TEMP_DIR = join(DATALAD_GIT_DIR, 'tmp', 'archives')
datalad/consts.py:WEB_META_LOG = join(DATALAD_GIT_DIR, 'logs')
datalad/consts.py:WEB_META_DIR = join(DATALAD_GIT_DIR, 'metadata')
```

where `WEB_META_LOG` is only used in `create_sibling` (in a remote context, where a const declaration might be OK).

and where `WEB_META_DIR` is not used in the core package at all.

`ARCHIVES_TEMP_DIR` is used by the `clean` command and `datalad/support/archives.py`, and could be easily replaced with a proper `.dotgit` usage. Same goes for the `ANNEX_TEMP_DIR` const.

mih:
I am linking this in #1095 -- it is pretty much the same thing. No need to keep them both around. Consts that are not that, should not be in consts.py

PR #4597, opened 2020-05-30
RF: Make .get_git_dir to use .dot_git logic, prepare for possible deprecation of .get_git_dir

yarikoptic:
#2054 made me look into this code path.

Possible changes to consider

- make `._get_dot_git` public so it could be used instead of `.get_git_dir`, thus not requiring an instance if that seems needed; but I think it better be delayed until we know that such functionality is really desired

For now I just position it against master (thus for 0.13) but if we feel that better be delayed, I will reassign milestone and finally produce `next` branch ;)

### DIFF
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -1182,8 +1182,27 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
         )
 
     @staticmethod
-    def _get_dot_git(pathobj, ok_missing=False, relative=False):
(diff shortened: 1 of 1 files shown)


yarikoptic:
pushed perspective fix in e0d4a253dbbde3bd482f0b2d4eaf5660ee475e92 .

yarikoptic:
> so feel free to merge this.

thanks! I am ok though to wait for 0.13 to get out first and to not disturb the "real core" functionality in the rc stage -- just don't want to break some existing functionality.

PR #4608, opened 2020-06-05
ENH: dataset cache for tests

bpoldrack:
Inspired by discussion in https://github.com/datalad/datalad-extensions/pull/12 and internal chat.

This allows to use remote datasets in tests via a cache, in order to reduce time and network traffic.
The goal is to replace the original remote repository (specified by a URL) by a local clone in a caching location. From the point of view of a test, this is kind of transparent - the cached dataset simply replaces the actually remote one. It introduces a config variable `datalad.tests.cache` to specify the caching location or disable caching entirely. Default location is `<user's cache dir>/datalad/tests`, which would result in datasets used by tests that way being persistent across test *runs*, not only across tests! This should be convenient for ourselves and irrelevant in CI builds, where the only thing interesting is to be persistent across tests (entire environment is fresh every time after all).

This PR brings two test decorators:

1. `cached_dataset`:
    This delivers a `Dataset` to the test, that is a temporary clone of the dataset in the cache (or of the actual URL if caching was disabled). That clone intentionally is *not* persistent - each test gets its own clone. The difference is that not every test using it, needs to clone (and get content) via network again, making it much cheaper (plus reducing chance to be rejected/blacklisted by flooding some services during our CI runs). In addition this decorator allows to specify paths to get and a version to checkout in that clone (of a clone).

2. `cached_url`:
    In opposition to above, this decorator merely delivers a URL, not a clone, to a test. If caching is disabled it merely passes on the specified URL, otherwise it delivers a file-scheme URL to the instance in cache. If the test merely needs something to clone from rather than to actually operate on, this is the way to go.

Note, that there are TODOs in the code wrt to enhancing this concept. If everything apart from that is fine with the rest of @datalad/developers , I'd rather postpone that and get the current version into 0.13.0 in order to solve the immediate usecase with datalad-hirni, which therefore needs to depend on a datalad version, that actually has those decorators.
Also: Those TODOs raise some questions to explore.

### DIFF
--- a/datalad/tests/test_utils_cached_dataset.py
+++ b/datalad/tests/test_utils_cached_dataset.py
@@ -1,0 +1,302 @@
+"""Testing cached test dataset utils"""
+
+from datalad.distribution.dataset import Dataset
+from datalad.support.annexrepo import AnnexRepo
+from datalad.support.gitrepo import GitRepo
+from datalad.tests.utils_cached_dataset import (
+    get_cached_dataset,
+    cached_dataset,
+    cached_url,
+    url2filename
+)
+from datalad.utils import (
+    ensure_list,
+    opj,
+    Path,
+)
+from datalad.tests.utils import (
+    assert_equal,
+    assert_false,
+    assert_in,
+    assert_is,
+    assert_is_instance,
+    assert_not_in,
+    assert_not_equal,
+    assert_raises,
+    assert_result_count,
+    assert_true,
+    with_tempfile
+)
+from unittest.mock import patch
+
+
+CACHE_PATCH_STR = "datalad.tests.utils_cached_dataset.DATALAD_TESTS_CACHE"
+CLONE_PATCH_STR = "datalad.tests.utils_cached_dataset.Clone.__call__"
+
+
+@with_tempfile(mkdir=True)
+def test_get_cached_dataset(cache_dir):
+
+    # patch DATALAD_TESTS_CACHE to not use the actual cache with
+    # the test testing that very cache.
+    cache_dir = Path(cache_dir)
+
+    # store file-based values for testrepo-minimalds for readability:
+    annexed_file = opj('inannex', 'animated.gif')
+    annexed_file_key = "MD5E-s144625--4c458c62b7ac8ec8e19c8ff14b2e34ad.gif"
+
+    with patch(CACHE_PATCH_STR, new=cache_dir):
+
+        # tuples to test (url, version, keys, class):
+        test_cases = [
+
+            # a simple testrepo
+            ("https://github.com/datalad/testrepo--minimalds",
+             "541cf855d13c2a338ff2803d4488daf0035e568f",
+             None,
+             AnnexRepo),
+            # Same repo, but request paths to be present. This should work
+            # with a subsequent call, although the first one did not already
+            # request any:
+            ("https://github.com/datalad/testrepo--minimalds",
+             "9dd8b56cc706ab56185f2ceb75fbe9de9b606724",
+             annexed_file_key,
+             AnnexRepo),
+            # Same repo again, but invalid version
+            ("https://github.com/datalad/testrepo--minimalds",
+             "nonexistent",
+             "irrelevantkey",  # invalid version; don't even try to get the key
+             AnnexRepo),
+            # same thing with different name should be treated as a new thing:
+            ("https://github.com/datalad/testrepo--minimalds",
+             "git-annex",
+             None,
+             AnnexRepo),
+            # try a plain git repo to make sure we can deal with that:
+            # Note, that we first need a test case w/o a `key` parameter to not
+            # blow up the test when Clone is patched, resulting in a MagicMock
+            # instead of a Dataset instance within get_cached_dataset. In the
+            # second case it's already cached then, so the patched Clone is
+            # never executed.
+            ("https://github.com/datalad/datalad.org",
+             None,
+             None,
+             GitRepo),
+            ("https://github.com/datalad/datalad.org",
+             "gh-pages",
+             "ignored-key",  # it's a git repo; don't even try to get a key
+             GitRepo),
+
+        ]
+        for url, version, keys, cls in test_cases:
+            target = cache_dir / url2filename(url)
+
+            # assuming it doesn't exist yet - patched cache dir!
+            in_cache_before = target.exists()
+            with patch(CLONE_PATCH_STR) as exec_clone:
+                try:
+                    ds = get_cached_dataset(url, version, keys)
+                    invalid_version = False
+                except AssertionError:
+                    # should happen only if `version` wasn't found. Implies
+                    # that the dataset exists in cache (although not returned
+                    # due to exception)
+                    assert_true(version)
+                    assert_false(Dataset(target).repo.commit_exists(version))
+                    # mark for later assertions (most of them should still hold
+                    # true)
+                    invalid_version = True
+
+            assert_equal(exec_clone.call_count, 0 if in_cache_before else 1)
+
+            # Patch prevents actual execution. Now do it for real. Note, that
+            # this might be necessary for content retrieval even if dataset was
+            # in cache before.
+            try:
+                ds = get_cached_dataset(url, version, keys)
+            except AssertionError:
+                # see previous call
+                assert_true(invalid_version)
+
+            assert_is_instance(ds, Dataset)
+            assert_true(ds.is_installed())
+            assert_equal(target, ds.pathobj)
+            assert_is_instance(ds.repo, cls)
+
+            if keys and not invalid_version and \
+                    AnnexRepo.is_valid_repo(ds.path):
+                # Note: it's not supposed to get that content if passed
+                # `version` wasn't available. get_cached_dataset would then
+                # raise before and not download anything only to raise
+                # afterwards.
+                here = ds.config.get("annex.uuid")
+                where = ds.repo.whereis(ensure_list(keys), key=True)
+                assert_true(all(here in remotes for remotes in where))
+
+            # version check. Note, that all `get_cached_dataset` is supposed to
+            # do, is verifying, that specified version exists - NOT check it
+            # out"
+            if version and not invalid_version:
+                assert_true(ds.repo.commit_exists(version))
+
+            # re-execution
+            with patch(CLONE_PATCH_STR) as exec_clone:
+                try:
+                    ds2 = get_cached_dataset(url, version, keys)
+                except AssertionError:
+                    assert_true(invalid_version)
+            exec_clone.assert_not_called()
+            # returns the same Dataset as before:
+            assert_is(ds, ds2)
+
+
+@with_tempfile(mkdir=True)
+def test_cached_dataset(cache_dir):
+
+    # patch DATALAD_TESTS_CACHE to not use the actual cache with
+    # the test testing that very cache.
+    cache_dir = Path(cache_dir)
+    ds_url = "https://github.com/datalad/testrepo--minimalds"
+    name_in_cache = url2filename(ds_url)
+    annexed_file = Path("inannex") / "animated.gif"
+
+    with patch(CACHE_PATCH_STR, new=cache_dir):
+
+        @cached_dataset(url=ds_url)
+        def decorated_test1(ds):
+            # we get a Dataset instance
+            assert_is_instance(ds, Dataset)
+            # it's a clone in a temp. location, not within the cache
+            assert_not_in(cache_dir, ds.pathobj.parents)
+            assert_result_count(ds.siblings(), 1, type="sibling",
+                                name="origin",
+                                url=str(cache_dir / name_in_cache))
+            here = ds.config.get("annex.uuid")
+            origin = ds.config.get("remote.origin.annex-uuid")
+            where = ds.repo.whereis(str(annexed_file))
+            assert_not_in(here, where)
+            assert_not_in(origin, where)
+
+            return ds.pathobj, ds.repo.pathobj
+
+        @cached_dataset(url=ds_url, paths=str(annexed_file))
+        def decorated_test2(ds):
+            # we get a Dataset instance
+            assert_is_instance(ds, Dataset)
+            # it's a clone in a temp. location, not within the cache
+            assert_not_in(cache_dir, ds.pathobj.parents)
+            assert_result_count(ds.siblings(), 1, type="sibling",
+                                name="origin",
+                                url=str(cache_dir / name_in_cache))
+            here = ds.config.get("annex.uuid")
+            origin = ds.config.get("remote.origin.annex-uuid")
+            where = ds.repo.whereis(str(annexed_file))
+            assert_in(here, where)
+            assert_in(origin, where)
+
+            return ds.pathobj, ds.repo.pathobj
+
+        @cached_dataset(url=ds_url)
+        def decorated_test3(ds):
+            # we get a Dataset instance
+            assert_is_instance(ds, Dataset)
+            # it's a clone in a temp. location, not within the cache
+            assert_not_in(cache_dir, ds.pathobj.parents)
+            assert_result_count(ds.siblings(), 1, type="sibling",
+                                name="origin",
+                                url=str(cache_dir / name_in_cache))
+            # origin is the same cached dataset, that got this content in
+            # decorated_test2 before. Should still be there. But "here" we
+            # didn't request it
+            here = ds.config.get("annex.uuid")
+            origin = ds.config.get("remote.origin.annex-uuid")
+            where = ds.repo.whereis(str(annexed_file))
+            assert_not_in(here, where)
+            assert_in(origin, where)
+
+            return ds.pathobj, ds.repo.pathobj
+
+        @cached_dataset(url=ds_url,
+                        version="541cf855d13c2a338ff2803d4488daf0035e568f")
+        def decorated_test4(ds):
+            # we get a Dataset instance
+            assert_is_instance(ds, Dataset)
+            # it's a clone in a temp. location, not within the cache
+            assert_not_in(cache_dir, ds.pathobj.parents)
+            assert_result_count(ds.siblings(), 1, type="sibling",
+                                name="origin",
+                                url=str(cache_dir / name_in_cache))
+            # origin is the same cached dataset, that got this content in
+            # decorated_test2 before. Should still be there. But "here" we
+            # didn't request it
+            here = ds.config.get("annex.uuid")
+            origin = ds.config.get("remote.origin.annex-uuid")
+            where = ds.repo.whereis(str(annexed_file))
+            assert_not_in(here, where)
+            assert_in(origin, where)
+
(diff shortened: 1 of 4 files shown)


datalad/tests/fixtures.py
```diff
@@ -0,0 +1,73 @@
+"""utilities to assist with setting up tests"""
+
+from datalad.api import clone
```
yarikoptic, line 3: Add direct import trio the line below, API is heavyish

datalad/tests/fixtures.py
```diff
@@ -0,0 +1,73 @@
...
+def get_cached_dataset(url, dataset_name=None, version=None):
+    """ Helper to get a cached clone from url
+
+    Intended for use from within `datalad_dataset` decorator.
+    Clones `url` into user's cache under datalad/tests/`name`. If such a clone
+    already exists, don't clone but return the existing one. So, it's supposed
+    to cache the original source in order to reduce time and traffic for tests,
+    by letting subsequent requests clone from a local location directly.
+
+    If it's an annex get the content, too.
```
yarikoptic, line 23: I would make it an option. May be make it a list of paths to get?
 Some tests might need just some files.
Add a note that dataset might or not already contain data for annexed files

datalad/tests/fixtures.py
```diff
@@ -0,0 +1,73 @@
...
+
+    Note
+    ----
+    `version` not yet implemented. Intention: verify that `version` can be
```
yarikoptic, line 27: Most likely then version should be included in the cache path, since different tests might want different versions... Alternative is to checkout a version for the test, but it would prohibit running in parallel ever

datalad/tests/fixtures.py
```diff
@@ -0,0 +1,73 @@
...
+
+    # TODO: a (env var) switch to use tmp location rather than user's cache?
+    #       would be: persistent across tests vs persistent across test runs.
+
+    if not dataset_name:
+        dataset_name = decode_source_spec(url)['default_destpath']
+
+    ds = Dataset(DATALAD_TESTS_CACHE / dataset_name)
+    if not ds.is_installed():
+        ds = clone(url, ds.pathobj)
```
yarikoptic, line 54: ds.clone(url)?

datalad/tests/fixtures.py
```diff
@@ -0,0 +1,73 @@
...
+def datalad_dataset(f, url=None, version=None, name=None):
+
+    # TODO: env var to enable/disable caching? Or melt with location switch
+    #       in get_cached_dataset?
+
+    @better_wraps(f)
+    @with_tempfile
+    def newfunc(*arg, **kw):
+        ds = get_cached_dataset(url, name, version=version)
+        clone_ds = clone(ds.pathobj, arg[-1])
```
yarikoptic, line 70: In many cases clone wouldn't be needed, option? That is why I thought to make get also optional in the original cached one

yarikoptic: May be also could be reckless? Pass clone kwargs?

bpoldrack: I think it should always be a clone. We shouldn't actually operate on datasets that are used across tests. And it wouldn't be a cache, if we deliver it directly. Getting a cached dataset shouldn't have any side effects compared to getting it directly from that URL (which would give you - a clone of the original one). To me the entire point is to be equivalent to cloning from that URL to a temp. location from the point of view of a test. Except that it becomes cheaper across tests and test runs in terms of network traffic and time needed. Cache serves just as a drop-in replacement for the original URL. And with the suggested env var it could then be disabled and clone directly from that URL (or use a temp. cache instead of user's). Therefore env var makes no difference to the test at all, but only to its performance.

datalad/tests/fixtures.py
bpoldrack: See comment below about necessity of `@datalad_dataset` always delivering a clone. I'd prefer to checkout a particular version only in that clone, that is delivered to the test, while the cached dataset itself only replaces the original source.

datalad/tests/fixtures.py
bpoldrack: Agreed.

datalad/tests/fixtures.py
bpoldrack: Yes. Will go for a list of paths, I guess.

datalad/tests/fixtures.py
bpoldrack: Shouldn't that clone *into* ds instead? That's not what I want.

datalad/tests/fixtures.py
bpoldrack: Re clone kwargs:
Guess that's useful, although I start being afraid of having to write tests for all those features and their combinations.

bpoldrack: Thinking about it a bit more:
May be it should be two decorators (don't want to much features in one thing):

1. as described above: replaces the remote one specified by `url` by a cached one as the origin and delivers a clone thereof. If caching is disabled, delivers a direct clone of the original remote.

2. Doesn't deliver a clone, but a URL to be used in the test. original URL if caching is disabled and path to a cached dataset (or `file:` scheme URL) if enabled.

What do you think about this, @yarikoptic ?

bpoldrack: Done. `@cached_dataset` + `@cached_url` instead of `@datalad_dataset`.

datalad/tests/fixtures.py
```diff
@@ -0,0 +1,155 @@
+"""utilities to assist with setting up tests"""
```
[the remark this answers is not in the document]
bpoldrack: Sure, I'm fine with that. Just went for your original proposal and didn't want to grow `.tests.utils` even further, given that there are several functions to deal with the concept.

bpoldrack:
> If such a dataset is used as input, I don't see the need for a clone of a clone with get - it just would double disc consumption and keep regretting data locally over and over again

Well, depends on what the test is going to do with the dataset. Generally I'd go for the safer option. Blowing up all tests simultaneously makes it harder to figure out what's going on. We've had those issues before. But I agree - as a test knows how dangerous it would be it can now decide to use either of those two decorators. With the second one no per-test clone is enforced. However, if we want to be able to decide whether or not caching is enabled, we cannot possibly provide a cached dataset directly, since then the test itself needs to be aware of the cache and its status. Kinda contradicts the definition of a cache.

datalad/tests/fixtures.py
```diff
@@ -0,0 +1,155 @@
...
+    if not dataset_name:
+        dataset_name = decode_source_spec(url)['default_destpath']
+
+    ds = Dataset(DATALAD_TESTS_CACHE / dataset_name)
+    if not ds.is_installed():
+        ds = Clone()(url, ds.pathobj)
+    if version:
+        # check whether version is available
+        assert ds.repo.commit_exists(version)
+    # TODO: What about an outdated dataset in cache? How do we properly update
```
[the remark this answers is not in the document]
bpoldrack: I think an option isn't it. That's again because it is supposed to work as cache and thereby needs to be transparent to the test. If the cache was skipped, no update would ever be required. If we now need to specify an option on how to update the usage becomes messy, I think.
I'd rather do something like:
If the dataset exists in cache, fetch and just proceed if there was nothing to fetch. If there was something however, we would need to decide what to do with that (merge isn't necessarily it) or - simplest solution - throw cache away in that case and get a fresh one.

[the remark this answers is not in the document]
yarikoptic: Good point about transparency! Then update only if version isn't found locally or explicitly requested overall?
We also will benefit from an explicit trigger to reset or ignore cached version (relates to the "fresh one") - I had to do that in dandi as well (https://github.com/dandi/dandi-cli/blob/master/dandi/support/cache.py#L38) although for a different type of cache (see https://github.com/con/pyfscacher/issues/1 for a higher level description/discussion, forgot to ping you there). I can see how it would be useful here as well

yarikoptic:
I don't think it really contradicts. Just that test should announce if it needs a clone, eg for the purpose of guarantees of absent content or for performing changes to it. We much have that also in "good" old with_test_repos was well already. (I think this new helper might provide the core functionality for RF of that elderly beast was well later on ;-))

datalad/tests/fixtures.py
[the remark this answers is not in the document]
bpoldrack: Will have a look. But quick remark:

> Then update only if version isn't found locally or explicitly requested overall?

Insufficient. "version" could be a branch. A test would "mean" the branch as available from `url`. But that one could have been updated. We would "find" `version` in an outdated dataset in cache.

bpoldrack:
>  Just that test should announce if it needs a clone, eg for the purpose of guarantees of absent content or for performing changes to it.

Don't think so. If we would deliver the cached dataset itself, it's not transparent anymore and the test relies on something being present in the cache. It shouldn't. Requesting a persistent-across-tests dataset to be locally available is very different thing than a cache. Of course we could build that almost the same way, but I would at least separate that kind of a request from what is implemented here. So - very similar implementation (different config) and a different decorator. I can do that, but:
1. I don't have an immediate use case
2. I generally don't like to encourage and suggest that kind of test writing. Sure - it's "wasteful" if all the tests work just right, but the point of testing is to ensure it's working right, not to assume it ;)

yarikoptic:
BTW! We could make original clone read only (after get competes, and rw for getting before get)! So the tests which do not request clone and try to modify it anyhow would fail - a nice test on itself IMHO

yarikoptic:
Anyways, that is just my 1c, and even without original clone reuse it would be useful, so it is ok to nevermind me ;-)

bpoldrack:
> BTW! We could make original clone read only (after get competes, and rw for getting before get)! So the tests which do not request clone and try to modify it anyhow would fail - a nice test on itself IMHO
> Anyways, that is just my 1c, and even without original clone reuse it would be useful, so it is ok to nevermind me ;-)

If combined with the separation I mentioned above (have dedicated decorator requesting a persistent clone rather than an actual transparent cache), I kinda like that idea.

datalad/tests/fixtures.py
```diff
@@ -0,0 +1,155 @@
...
+    ensure_list,
+    optional_args,
+)
+from datalad.tests.utils import with_tempfile
+
+
+DATALAD_TESTS_CACHE = cfg.obtain("datalad.tests.cache")
+
+
+def get_cached_dataset(url, dataset_name=None, version=None, paths=None):
```
yarikoptic, line 20: BTW, I was thinking about this - where to give etc. Since I expect external tools also make use of it, and cache location to be the same, default location and path would likely not avoid conflicts (different urls into the same location). Since we don't know dataset uuId from url either - the best would be to use the same approach as git annex addurl does and create target folder name from entire url. For a test it should matter, and would likely to avoid confusion etc. Would look quite alright for all /// datasets I guess

bpoldrack: Not sure what you mean by "external tools". However, default clone names might be insufficient - I agree. addurl approach sounds good. But then I will eliminate `dataset_name` altogether and ensure the name does always reflect the URL.

datalad/tests/fixtures.py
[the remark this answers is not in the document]
yarikoptic: I guess it was my bad to forget about what is out there already, but also in nose fixtures notion is more vague than in pytest where IIRC they need to be in fixtures.py abs that is what misguided me

datalad/tests/fixtures.py
yarikoptic: Re external: so we could use it eg in heudiconv by importing from datalad . I don't see an immediate need for name to stay, so removal sounds good to me

bpoldrack: Done. Simply replaced `/,:,?` by `_` for now. Thought about hashing the URL instead, but this would become more complicated if we add version to the URL itself. Guess we need to see how this evolves. Will need more sophisticated URL handling as we make use of those tools for `ria+` URLs anyway.

bpoldrack:
Assuming that tests still pass, I think this is ready to go into 0.13.0 as is.
If we agree on this, I'll make an issue to collect next pieces as discussed here with @yarikoptic and outlined in the code comments. This would include to add very similar decorators for providing persistent read-only clones (in opposition to cache, this would not be a transparent thing for the tests). Furthermore cache could be more efficient with `--bare` or `--mirror` clones, etc. Which in turn requires such option(s) for `datalad-clone`. Recursion is yet another topic to address. Generalization for other uses such as in heudiconv or even some kind of datalad-cache feature needs some evolution and experience in using it, I guess.
However, this is already something very useful, I think, it solves the immediate use-case and most of the possible changes and additions are either independent or transparent for the decorated tests.

Agreed, @datalad/developers ?

P.S.: If so, I should prob. squash before merge.

datalad/tests/utils_cached_dataset.py
```diff
@@ -0,0 +1,233 @@
...
+    @better_wraps(f)
+    @with_tempfile
+    def newfunc(*arg, **kw):
+
+        if DATALAD_TESTS_CACHE:
+            # Note: We can't pass keys based on `paths` parameter to
+            # get_cached_dataset yet, since translation to keys depends on a
+            # worktree. We'll have the worktree of `version` only after cloning.
+            ds = get_cached_dataset(url, version=version)
+            clone_ds = Clone()(ds.pathobj, arg[-1])
```
yarikoptic, line 160: BTW -- to be more "transparent cache" instance, shouldn't then it 
- clone so that local cached "origin" is named differently than "origin" remote (e.g. `local-cached`)
- add "origin" remote which would be the original URL?
the only possible side-effect I see could be -- it would need to fetch that "origin" remote so might prevent testing "offline" but so far I think the entire approach is for "online" testing

yarikoptic: also probably that would undermine benefits for some tests use cases which might want to request data specifically from "origin".

bpoldrack: >  add "origin" remote which would be the original URL?
the only possible side-effect I see could be -- it would need to fetch that "origin" remote so might prevent testing "offline" but so far I think the entire approach is for "online" testing

Quite the opposite! With that approach a test would need to be aware of the cache in order to make any use of it - that's the exact opposite of transparent. With the implementation here, the test would always talk to "origin" - not knowing whether or not it's cached. That is the point! Also: I can't disable caching without side effect on the tests, if they would explicitly talk to a cache remote.

datalad/tests/utils_cached_dataset.py
```diff
@@ -0,0 +1,233 @@
...
+    ds = Dataset(DATALAD_TESTS_CACHE / url2filename(url))
+
+    if not ds.is_installed():
+        ds = Clone()(url, ds.pathobj)
+
+    # When/How to update a dataset in cache? If version is a commit SHA and we
+    # have it, there's no need for an update. Otherwise it gets tricky, because
+    # this is a cache, not a checkout a test would operate on. It needs to
+    # behave as if it was the thing at `url` from the point of view of the test
+    # using it (cloning/getting content from here). We would nee to update all
```
kyleam, line 93: typo: need

datalad/tests/utils_cached_dataset.py
```diff
@@ -0,0 +1,233 @@
...
+                # cache is enabled; we need to make sure it has the desired
+                # content, so clone_ds can get it from there. However, we got
+                # `paths` and potentially a `version` they refer to. We can't
+                # assume the same (or any) worktree in cache. Hence we need to
+                # translate to keys.
+                keys = clone_ds.repo.get_file_key(paths)
+                ds.repo.get(keys, key=True)
+                clone_ds.repo.fsck(remote='origin', fast=True)
+
+            clone_ds.get(ensure_list(paths))
```
kyleam, line 178: Doesn't `get` already handle the string case?

datalad/tests/test_utils_cached_dataset.py
```diff
@@ -0,0 +1,320 @@
...
+             AnnexRepo),
+            # same thing with different name should be treated as a new thing:
+            ("https://github.com/datalad/testrepo--minimalds",
+             "git-annex",
+             None,
+             AnnexRepo),
+            # try a plain git repo to make sure we can deal with that:
+            # Note, that we need it twice to not blow up the test when Clone
+            # is patched, resulting in a MagicMock instead of a Dataset instance
+            # within get_cached_dataset.
```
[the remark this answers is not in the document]
bpoldrack: A test case with that URL is needed twice, if I want to test `version` and `keys` parameter with a git repo, too. Will change that comment.

datalad/tests/test_utils_cached_dataset.py
```diff
@@ -0,0 +1,320 @@
...
+                    invalid_version = True
+
+            if not in_cache_before:
+                # clone was called
+                # Note: assert_called was only introduced in 3.6, while
+                # assert_not_called exists in 3.5 already. As we test for python
+                # 3.5, we need to solve it differently. At the same time the
+                # test isn't about the precise call as required by
+                # assert_called_with.
+                # -> just negate assert_not_called
```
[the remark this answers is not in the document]
bpoldrack: True. Thanks.

datalad/tests/test_utils_cached_dataset.py
```diff
@@ -0,0 +1,320 @@
...
+                # raise before and not download anything only to raise
+                # afterwards.
+                # Note2 / TODO: We currently have no method to check whether a
+                # particular is locally present. Strangely enough even
+                # annex-checkpresentkey doesn't allow to specify "here" as the
+                # remote to check. Therefore being somewhat inconsistent with
+                # the `test_cases` approach here, assuming we know the path to
+                # the key. At least make that assumption explicit, so addition
+                # of a test case w/ different key will lead to failure here.
+                # However, there should be a better solution!
```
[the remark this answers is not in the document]
bpoldrack: Right. That's what I do in the other tests. Sometimes my brain does weird things. ;-)

datalad/tests/utils_cached_dataset.py
```diff
@@ -0,0 +1,233 @@
...
+
+DATALAD_TESTS_CACHE = cfg.obtain("datalad.tests.cache")
+
+
+def url2filename(url):
+    """generate file/directory name from a URL"""
+
+    # TODO: Not really important for now, but there should be a more
+    #       sophisticated approach to replace. May be just everything that
+    #       isn't alphanumeric? Or simply hash the URL?
```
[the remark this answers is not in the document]
bpoldrack: It can be a big deal to have an additional clone. The immediate use case, causing this entire PR, is about a dataset with several container images. Getting that content is what I need a cache for.

But I do like a hash as well. Given that I don't want a clone per version, I'd prob. need to parse the URL and hash it w/o any parameters.

datalad/tests/utils_cached_dataset.py
bpoldrack: True. That seems superfluous.

bpoldrack:
Okay, per today's call: will squash, have a proper issue for further development and merge ...

yarikoptic:
Per discussion, squash, and greens of CI - merging! Thanks @bpoldrack!

Issue #4613, opened 2020-06-08
RIA subdataset install "broken" by early success


mih:
Demo:

```
% datalad clone ria+http://rs.ds.inm7.de#~super all_rs_data
% cd all_rs_data
% datalad get eNKI/sub-A00008399
[INFO   ] Remote origin not usable by git-annex; setting annex-ignore                        
[INFO   ] Configured RIA store not found at ria+file:///ds/rs/srv                            
install(ok): /tmp/all_rs_data/eNKI (dataset) [Installed subdataset in order to get /tmp/all_rs_data/eNKI/sub-A00008399]
Total (0 ok, 4 failed out of 4):   0%|                            | 0.00/207M [00:00<?, ?B/s][WARNING] Running get resulted in stderr output: git-annex: get: 4 failed
 
                                                                                             [ERROR  ] not available; Try making some of these repositories available:;      019b3420-fc32-4984-93c6-b79a705511a2 -- mih@bulk1:/ds/rs/incoming/eNKI;      38cb587d-c7ec-4228-a8a2-a1abec89331a -- inm7-storage; ; (Note that these git remotes have annex-ignore set: origin) [get(/tmp/all_rs_data/eNKI/sub-A00008399/ses-DS2/Confounds_sub-A00008399_ses-DS2.mat)] 
...
```

The ora special remote config doesn't work `ria+file:///ds/rs/srv` so a reconfiguration should have been triggered, but wasn't. The reason is this:

```
% git grep -A 3 'submodule.*eNKI' -- .gitmodules |cat
.gitmodules:[submodule "eNKI"]
.gitmodules-    path = eNKI
.gitmodules-    url = http://rs.ds.inm7.de/23e/45ab0-a588-11ea-8ccf-a0369f287950
.gitmodules-    datalad-id = 23e45ab0-a588-11ea-8ccf-a0369f287950
```
The URL recorded in `.gitmodules` is actually functional and is tried before the configured RIA lookup:

```
% git config datalad.get.subdataset-source-candidate-origin
ria+http://rs.ds.inm7.de#{id}
```

Not sure which way would be best here. The problem is caused by the facts that I used

```
datalad clone -d^ ria+http://rs.ds.inm7.de#23e45ab0-a588-11ea-8ccf-a0369f287950 eNKI
```

when adding the subdataset originally -- which get's resolved to the working http URL above. I could simply not do that and use a local clone path. or we change the lookup order to prefer datalad config sources over `.gitmodules`.

[...]

PR #4619, opened 2020-06-09
ENH: Give clone candidates a priority

mih:
This is achieved by prefixing their existing label with three digits and sorting them before processing them.

This implementation makes no change to the previously established order, it just makes that order explicit in the labels.

The default setup uses the priorities in the second half of the spectrum:

- 500 for remote URL + submodule path
- 600 for the configured submodule URL
- 700 for any unprioritized 'datalad.get.subdataset-source-candidate' config
- 900 for the local subdataset path

The first half of the spectrum and the spaces inbetween are available for re-prioritizing clone attempts. The idea is to be able to configure

  `datalad.get.subdataset-source-candidate-000takemefirst = http://some`

and have this be processed first.

I have confirmed manually that this solution enables successful clones on first attempt, and substantially speeds up dataset installation. Likewise using '999label' it is possible to configure a fallback source candidate.

Towards a resolution of gh-4613

TODO:
- [x] adjust `clone` to use this feature
- [x] adjust documentation
- [x] add test
- ~~potentially prevent "flexibilizing" of explicitly configured URLs (e.g. `/.git` appending)~~

### DIFF
--- a/datalad/core/distributed/clone.py
+++ b/datalad/core/distributed/clone.py
@@ -584,7 +584,7 @@ def postclonecfg_ria(ds, props):
     ds.config.set(
         # we use the label 'origin' for this candidate in order to not have to
         # generate a complicated name from the actual source specification.
-        # we pick a priority of 20 to sort it before datalad's default candidates
+        # we pick a priority of 200 to sort it before datalad's default candidates
         # for non-RIA URLs, because they prioritize hierarchical layouts that
         # cannot be found in a RIA store
         'datalad.get.subdataset-source-candidate-200origin',
--- a/datalad/core/distributed/clone.py
+++ b/datalad/core/distributed/clone.py
@@ -583,8 +583,11 @@ def postclonecfg_ria(ds, props):
     # get probe this RIA store when obtaining subdatasets
     ds.config.set(
         # we use the label 'origin' for this candidate in order to not have to
-        # generate a complicated name from the actual source specification
-        'datalad.get.subdataset-source-candidate-origin',
+        # generate a complicated name from the actual source specification.
+        # we pick a priority of 20 to sort it before datalad's default candidates
+        # for non-RIA URLs, because they prioritize hierarchical layouts that
+        # cannot be found in a RIA store
+        'datalad.get.subdataset-source-candidate-20origin',
         # use the entire original URL, up to the fragment + plus dataset ID
         # placeholder, this should make things work with any store setup we
         # support (paths, ports, ...)
--- a/datalad/core/distributed/clone.py
+++ b/datalad/core/distributed/clone.py
@@ -584,7 +584,7 @@ def postclonecfg_ria(ds, props):
     ds.config.set(
         # we use the label 'origin' for this candidate in order to not have to
         # generate a complicated name from the actual source specification.
-        # we pick a priority of 200 to sort it before datalad's default candidates
+        # we pick a cost of 200 to sort it before datalad's default candidates
         # for non-RIA URLs, because they prioritize hierarchical layouts that
         # cannot be found in a RIA store
         'datalad.get.subdataset-source-candidate-200origin',
--- a/datalad/core/distributed/clone.py
+++ b/datalad/core/distributed/clone.py
@@ -587,7 +587,7 @@ def postclonecfg_ria(ds, props):
         # we pick a priority of 20 to sort it before datalad's default candidates
         # for non-RIA URLs, because they prioritize hierarchical layouts that
         # cannot be found in a RIA store
-        'datalad.get.subdataset-source-candidate-20origin',
+        'datalad.get.subdataset-source-candidate-200origin',
         # use the entire original URL, up to the fragment + plus dataset ID
         # placeholder, this should make things work with any store setup we
         # support (paths, ports, ...)
--- a/datalad/distribution/get.py
+++ b/datalad/distribution/get.py
@@ -100,22 +100,22 @@ def _get_remotes_having_commit(repo, commit_hexsha, with_urls_only=True):
 def _get_flexible_source_candidates_for_submodule(ds, sm):
     """Assemble candidate locations from where to clone a submodule
 
-    The following locations candidates are considered. For each candidate
-    a priority is given in parenthesis, lower values indicate higher priority:
+    The following locations candidates are considered. For each candidate a
+    cost is given in parenthesis, lower values indicate higher cost:
 
     - URL of any configured superdataset remote that is known to have the
       desired submodule commit, with the submodule path appended to it.
-      There can be more than one candidate (priority 500).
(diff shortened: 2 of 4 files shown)


mih:
Crippled failure, not yet replicatable locally...
```
FAIL: datalad.core.distributed.tests.test_push.test_push(True,)
  File "/opt/hostedtoolcache/Python/3.5.9/x64/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/opt/hostedtoolcache/Python/3.5.9/x64/lib/python3.5/site-packages/datalad/tests/utils.py", line 714, in newfunc
  File "/opt/hostedtoolcache/Python/3.5.9/x64/lib/python3.5/site-packages/datalad/tests/utils.py", line 714, in newfunc
  File "/opt/hostedtoolcache/Python/3.5.9/x64/lib/python3.5/site-packages/datalad/core/distributed/tests/test_push.py", line 131, in check_push
    assert_result_count(res, 2 if annex else 1)
  File "/opt/hostedtoolcache/Python/3.5.9/x64/lib/python3.5/site-packages/datalad/tests/utils.py", line 1305, in assert_result_count
    _format_res(results)))
AssertionError: Got 3 instead of 2 expected results matching
  {}
...
```

mih:
I will not add the ability to disable automatically appending `/.git` to all relevant URLs. It is unclear to me when that is possible without impairing legitimate use cases. The sorting already enable to get in front of the list of candidate and provide a URL that works as the first one to be tried.

datalad/core/distributed/clone.py
```diff
@@ -583,8 +583,11 @@ def postclonecfg_ria(ds, props):
     # get probe this RIA store when obtaining subdatasets
     ds.config.set(
         # we use the label 'origin' for this candidate in order to not have to
-        # generate a complicated name from the actual source specification
-        'datalad.get.subdataset-source-candidate-origin',
+        # generate a complicated name from the actual source specification.
+        # we pick a priority of 20 to sort it before datalad's default candidates
+        # for non-RIA URLs, because they prioritize hierarchical layouts that
+        # cannot be found in a RIA store
+        'datalad.get.subdataset-source-candidate-20origin',
```
yarikoptic, line 590: I wondered if it might make sense to separate out priority from the variable: i.e. introduce e.g.`datalad.get.subdataset-source-candidate-origin-priority` which if not set would get `20`.  That would 
- make existing clones work without any changes
- make it more explicit
- internal implementation could do whatever it does now (I would have preferred to have priority an explicit variable/tuple element to avoid all the splitting etc)  and could be refactored later if desired/needed

mih: If a second variable is introduce, one would also need to make sure that they stay in sync across all sources we read config from. I do not know how to achieve that, and how to test whether I have achieved that. It is perfectly possible that `datalad.get.subdataset-source-candidate-origin-priority` comes from user config and `datalad.get.subdataset-source-candidate-origin` does not.

I am not much concerned about existing clones -- I have all of them in front of me.

mih: I addressed the topic of the messy internal implementation in 05e4d8f600cac4fcb999c88e89deaaa184e1c1b9 and it also switched to a three-digit priority setup.

datalad/distribution/get.py
```diff
@@ -191,27 +192,35 @@ def _get_flexible_source_candidates_for_submodule(ds, sm):
...
-                           for c in ds_repo.config.keys()
-                           if c.startswith(
-                               'datalad.get.subdataset-source-candidate-')]:
-            url = tmpl.format(**sm_candidate_props)
-            # we don't want "flexible_source_candidates" here, this is
-            # configuration that can be made arbitrarily precise from the
-            # outside. Additional guesswork can only make it slower
-            clone_urls.append((name, url))
+    prio_candidate_expr = re.compile('[0-9][0-9].*')
+    for name, tmpl in [(c[40:], ds_repo.config[c])
```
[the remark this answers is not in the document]
mih: Fixed in e0798e9c680d7a56c04bd09f5059b6360829414a

datalad/core/distributed/clone.py
yarikoptic: > It is perfectly possible that datalad.get.subdataset-source-candidate-origin-priority comes from user config and datalad.get.subdataset-source-candidate-origin does not.

Wouldn't that be a feature - I could boost or penalize priority of all ria stores with one config setting?

yarikoptic: Besides I would follow annex and call it cost and not priority (ambiguous - is higher number a higher priority? More obvious with cost, and is taking the one with lower cost first).

yarikoptic: And in current approach it would be easy only to raise priority (lower cost) but not to lower it since the original variable would still be available somewhere with higher cost?

mih: > Wouldn't that be a feature - I could boost or penalize priority of all ria stores with one config setting?

That may be, but a feature that I do not comprehend enough to be the one implementing it, and there wasn't a use case for it either.

> Besides I would follow annex and call it cost and not priority (ambiguous - is higher number a higher priority? More obvious with cost, and is taking the one with lower cost first).

I can do this, if needed (but it may not be enough, see below).

> And in current approach it would be easy only to raise priority (lower cost) but not to lower it since the original variable would still be available somewhere with higher cost?

There is no concept of raising or lowering, there are just candidates with (default) priorities.

I suspect that you want to aim for a bigger solution. If you insist, I can obscure the fact that the sorting can be used to achieve a prioritizing and just talk about sorting. In this case the space would be open for the feature that you have in mind. I just want to get a handle on gh-4613, have that quickly, and then I will detach from this topic.

yarikoptic: > There is no concept of raising or lowering, there are just candidates with (default) priorities.

ah -- then there should be no need for any config changes? just RF code (if still needed) so priority for RIA remotes is higher and be done?

> I suspect that you want to aim for a bigger solution. If you insist, ...

oh no -- I am not insisting or even suggesting to implement anything bigger solution here.  I just want this solution not require changes to user visible parts (e.g. config) later if we decide to go with some global user visible approach.

mih: If an existing unprioritized candidate configuration is found, it automatically gets priority 700 such that identical behavior is maintained. This doesn't fix gh-4613, but I see no reason for doing such intervention in repos where this wasn't a problem so far (and I believe there are just a handful). All new clones will get the new behavior.

mih:
Thanks for the reviews and for the approval @kyleam 

I took the liberty to still rename `priority` to `cost` to resolve interpretation issue pointed out in https://github.com/datalad/datalad/pull/4619#discussion_r439326884

Will merge when green.

Issue #4604, opened 2020-06-03
push: yet another round on the interface, but with a new twist


yarikoptic:
Decided to not pollute #4581 or any other prior discussion thus filed separately

Current `--force` options are already non orthogonal (impossible to use `no-datatransfer` and `datatransfer` together).  I think it would be great to reserve `--force` for tune ups to specific git or annex actions (like `gitpush` already and possibly `no-annexfast`  #4603)  while making a dedicated option to define the data transfer mode, and just change the default to follow @mih's desire for the new "right thing" behavior (#4551) to copy annexed data by default.  Moreover, while going through such RFing - we gain a chance to align it with git-annex a bit more.

What if we add the same option as `publish` has -- `--transfer-data` with following possible values:

- (default) "anything" -- transfer any file not available on remote
- (previous `publish` behavior) "nothing" -- transfer nothing
- "auto" -- pass `--auto` to `annex copy` to satisfy "wanted" expression for the remote if any, and otherwise transfer nothing (note: it would also respect a possible `numcopies` setting).

"anything" and "nothing" from above - are wanted expressions of annex.  Further documentation could point to `--annex-wanted` of `create-siblings` (and may be `siblings` if exposed there) to hint on how it could be tuned to automate.

`--fast` will be used, and could be removed via `--force` option.

I do  like an idea of configurable default `push` data transfer behavior discussed in #4581 .

Later we could expand the list if any additional specific desire, but specific for data transfer, not fusing it into the list of `forces`.

**Alternative** to `--transfer-data MODE` is to align with current `get` and its `-n|--no-data`, and also introduce `-a|--auto-data`.

edit1: "annex-auto" -> "auto", and added `get` inspired alternative

mih:
I am mentally dead right now, so only briefly: the reason that we do not have transter-data anymore is the cumulative effect of all issues fixed by push.

Also there is no desire for all force modes to be orthogonal, they are no more than mutually exclusive.

yarikoptic:
I am not suggesting to adopt transferdata implementation from `publish`.  I merely suggest to RF interface - implementation should stay the same (just use `data_transfer` option instead of force modes).  Would a PR help to illustrate? (could try on weekend or so)

bpoldrack:
[...]

PR #4620, opened 2020-06-09
ENH: specify `push` --data modes; keep `--force` to only force git push and checkdatapresent

### LINKED ISSUE #4602: push: --force=onlywanteddatatransfer
I still wonder how I could use new place in my workflow without major changes to existing Datasets. Could such (or is there a better name?) force mode be considered, which would mimic current publish behavior (no data transfer unless there is a wanted setting, and then --auto) in case of no paths provided?

### LINKED ISSUE #4603: push: --force=datatransfer -> --force=no-annexfast
#### What is the problem?

Absence of `--fast` does not actually force any data transfer if content is already available on the remote.  

<details>
<summary>It just causes annex to check explicitly remote if it contains the key</summary> 

```shell
$> time git annex copy --to neurosmaug --force freesurfer-extlibs-sloppy_201* --debug
[2020-06-03 11:49:06.135276985] read: git ["--git-dir=.git","--work-tree=.","--literal-pathspecs","show-ref","git-annex"]
[2020-06-03 11:49:06.15099345] process done ExitSuccess
[2020-06-03 11:49:06.151612786] read: git ["--git-dir=.git","--work-tree=.","--literal-pathspecs","show-ref","--hash","refs/heads/git-annex"]
[2020-06-03 11:49:06.16968624] process done ExitSuccess
[2020-06-03 11:49:06.171218456] read: git ["--git-dir=.git","--work-tree=.","--literal-pathspecs","log","refs/heads/git-annex..91f750b383c1bae79d831a204a942b82d42f966d","--pretty=%H","-n1"]
[2020-06-03 11:49:06.185886773] process done ExitSuccess
[2020-06-03 11:49:06.186056453] read: git ["--git-dir=.git","--work-tree=.","--literal-pathspecs","log","refs/heads/git-annex..03d545ae03f864c862e5e0846ad1ddfbb95ca5b7","--pretty=%H","-n1"]
[2020-06-03 11:49:06.205074385] process done ExitSuccess
[2020-06-03 11:49:06.239273514] chat: git ["--git-dir=.git","--work-tree=.","--literal-pathspecs","cat-file","--batch"]
[2020-06-03 11:49:06.241931433] chat: git ["--git-dir=.git","--work-tree=.","--literal-pathspecs","cat-file","--batch-check=%(objectname) %(objecttype) %(objectsize)"]
[2020-06-03 11:49:06.266777301] read: git ["--git-dir=.git","--work-tree=.","--literal-pathspecs","symbolic-ref","-q","HEAD"]
[2020-06-03 11:49:06.287075158] process done ExitSuccess
[2020-06-03 11:49:06.287590555] read: git ["--git-dir=.git","--work-tree=.","--literal-pathspecs","show-ref","refs/heads/master"]
[2020-06-03 11:49:06.304955359] process done ExitSuccess
[2020-06-03 11:49:06.306021498] read: git ["--git-dir=.git","--work-tree=.","--literal-pathspecs","ls-files","--cached","-z","--","freesurfer-extlibs-sloppy_20160908.orig.tar.gz","freesurfer-extlibs-sloppy_20180917.orig.tar.gz"]
copy freesurfer-extlibs-sloppy_20160908.orig.tar.gz [2020-06-03 11:49:06.323629099] chat: ssh ["neurosmaug","-S",".git/annex/ssh/neurosmaug","-o","ControlMaster=auto","-o","ControlPersist=yes","-o","ControlMaster=auto","-S","/home/yoh/.config/datalad/neurosmaug","-T","git-annex-shell 'p2pstdio' '/~/deb/tarballs' '--debug' 'b8e253ac-6612-11e0-bab6-b3cc258b7ba2' --uuid 9b49e014-65ed-11e0-af7a-f7428466115a"]
[2020-06-03 11:49:06.421561299] [ssh connection 1] [ThreadId 4] P2P < AUTH-SUCCESS 9b49e014-65ed-11e0-af7a-f7428466115a
[2020-06-03 11:49:06.421979074] [ssh connection 1] [ThreadId 4] P2P > VERSION 1
[2020-06-03 11:49:06.445732611] [ssh connection 1] [ThreadId 4] P2P < VERSION 1
[2020-06-03 11:49:06.446496622] [ssh connection 1] [ThreadId 4] P2P > CHECKPRESENT SHA256E-s362575955--0a322a937d5ed1314d331cbf6bf760e4bce2b0cc59e642d25b8ea5b97fbd8e3d.tar.gz
[2020-06-03 11:49:06.432505233] [ThreadId 4] P2P < VERSION 1[2020-06-03 11:49:06.432708744] [ThreadId 4] P2P > VERSION 1[2020-06-03 11:49:06.455380867] [ThreadId 4] P2P < CHECKPRESENT SHA256E-s362575955--0a322a937d5ed1314d331cbf6bf760e4bce2b0cc59e642d25b8ea5b97fbd8e3d.tar.gz[2020-06-03 11:49:06.468464896] [ssh connection 1] [ThreadId 4] P2P < SUCCESS
[2020-06-03 11:49:06.456677252] [ThreadId 4] P2P > SUCCESSok
copy freesurfer-extlibs-sloppy_20180917.orig.tar.gz [2020-06-03 11:49:06.472954707] [ssh connection 1] [ThreadId 4] P2P > CHECKPRESENT SHA256E-s335050843--4d97a6b51ac2dbc9b4a3e64700aebae954e250033fc394f104c5079ef2c29c13.tar.gz
[2020-06-03 11:49:06.476913577] [ThreadId 4] P2P < CHECKPRESENT SHA256E-s335050843--4d97a6b51ac2dbc9b4a3e64700aebae954e250033fc394f104c5079ef2c29c13.tar.gz[2020-06-03 11:49:06.477221749] [ThreadId 4] P2P > SUCCESS[2020-06-03 11:49:06.49077453] [ssh connection 1] [ThreadId 4] P2P < SUCCESS
ok
[2020-06-03 11:49:06.499921703] process done ExitSuccess
[2020-06-03 11:49:06.503645896] process done ExitSuccess
git annex copy --to neurosmaug --force freesurfer-extlibs-sloppy_201* --debug  0.18s user 0.17s system 70% cpu 0.503 total
```
</details>

As such, name of the "force" is IMHO misleading even if described in the doc that it is actually not that but just not passing `--fast` to git-annex copy call.  I think `no-annexfast` would align better with another force of `gitpush`.  Note that `annex copy` also has `--force` option although I am not certain about its purpose/effects.

yarikoptic:
This is an implementation for #4604 .

Notes:
- it is not harmonized with `publish` or `get` (see #4604 for notes on that)
- from Python programming side perspective I do dislike that we cannot say `data=True` or `False`. `data='nothing'` is not worse IMHO though than `force=['no-datatransfer']` though.
-  `--force=checkdatapresent`  (resolves #4603) in b20ed1379da540621e48a1515191a9ea32ae258f
- `--force=all` instead of `--force=pushall` in 65aae48ef0cee708a7e0fee89f076b152b64eef2 since IMHO forces should be orthogonal and could all be combined
- Closes #4602 which was an alternative proposed solution (`--force=onlywanteddatatransfer`)
- Removed `datalad.push.copy-auto-if-wanted` config adding a mode `auto-if-wanted` (1be419f0483dde6db939b2263bdbcf3d355dafb5) and making it to be default mode of operation.  It would cause transfer ("anything") if no "wanted" is not set, thus fulfilling the "do the right thing by default".
- Uses `--data` (see #4622)

See individual commits for more information.  I had tried to make them atomic enough so we could drop some if so desired.

### DIFF
--- a/datalad/core/distributed/push.py
+++ b/datalad/core/distributed/push.py
@@ -123,16 +123,23 @@ class Push(Interface):
             data or changes for those paths are considered for a push.""",
             nargs='*',
             constraints=EnsureStr() | EnsureNone()),
+        transfer_data=Parameter(
+            args=("--transfer-data",),
+            doc="""what to do with data (annex'ed) data. 'anything' would cause
+            transfer of all annexed content, 'nothing' would avoid call to
+            `git annex copy` altogether. 'auto' would use 'git annex copy' with
+            '--auto' thus transferring only data which would satisfy "wanted"
+            or "numcopies" settings for the remote.""",
+            constraints=EnsureChoice('anything', 'nothing', 'auto')),
         force=Parameter(
             # multi-mode option https://github.com/datalad/datalad/issues/3414
             args=("-f", "--force",),
             doc="""force particular operations, overruling automatic decision
             making: use --force with git-push ('gitpush'); do not use --fast
-            with git-annex copy ('datatransfer'); do not attempt to copy
-            annex'ed file content ('no-datatransfer'); combine force modes
+            with git-annex copy ('datatransfer'); combine force modes
             'gitpush' and 'datatransfer' ('pushall').""",
             constraints=EnsureChoice(
-                'pushall', 'gitpush', 'no-datatransfer', 'datatransfer', None)),
+                'pushall', 'gitpush', 'datatransfer', None)),
         recursive=recursion_flag,
         recursion_limit=recursion_limit,
         jobs=jobs_opt,
@@ -181,6 +188,7 @@ class Push(Interface):
             dataset=None,
             to=None,
             since=None,
+            transfer_data='anything',
             force=None,
             recursive=False,
             recursion_limit=None,
@@ -249,7 +257,7 @@ class Push(Interface):
             lgr.debug('Attempt push of Dataset at %s', dspath)
             pbars = {}
             yield from _push(
-                dspath, dsrecords, to, force, jobs, res_kwargs.copy(), pbars,
+                dspath, dsrecords, to, transfer_data, force, jobs, res_kwargs.copy(), pbars,
                 got_path_arg=True if path else False)
             # take down progress bars for this dataset
             for i, ds in pbars.items():
@@ -366,7 +374,7 @@ def _datasets_since_(dataset, since, paths, recursive, recursion_limit):
         yield (cur_ds, ds_res)
 
 
-def _push(dspath, content, target, force, jobs, res_kwargs, pbars,
+def _push(dspath, content, target, transfer_data, force, jobs, res_kwargs, pbars,
           done_fetch=None, got_path_arg=False):
     if not done_fetch:
         done_fetch = set()
@@ -526,6 +534,7 @@ def _push(dspath, content, target, force, jobs, res_kwargs, pbars,
             content,
             # to this particular dependency
             r,
+            transfer_data,
             force,
             jobs,
             res_kwargs.copy(),
@@ -566,7 +575,7 @@ def _push(dspath, content, target, force, jobs, res_kwargs, pbars,
     if not is_annex_repo:
         return
 
-    if force == 'no-datatransfer':
+    if transfer_data == "nothing":
         lgr.debug("Data transfer to '%s' disabled by argument", target)
         return
 
@@ -578,6 +587,7 @@ def _push(dspath, content, target, force, jobs, res_kwargs, pbars,
         ds,
         target,
         content,
+        transfer_data,
         force,
         jobs,
         res_kwargs.copy(),
@@ -696,7 +706,7 @@ def _push_refspecs(repo, target, refspecs, force, res_kwargs):
         )
 
 
-def _push_data(ds, target, content, force, jobs, res_kwargs,
+def _push_data(ds, target, content, transfer_data, force, jobs, res_kwargs,
                got_path_arg=False):
     if ds.config.getbool('remote.{}'.format(target), 'annex-ignore', False):
         lgr.debug(
@@ -772,11 +782,15 @@ def _push_data(ds, target, content, force, jobs, res_kwargs,
     if jobs:
         cmd.extend(['--jobs', str(jobs)])
 
-    if force not in ('pushall', 'datatransfer') and ds_repo.config.obtain(
-            'datalad.push.copy-auto-if-wanted'):
-        if ds_repo.get_preferred_content('wanted', target):
-            lgr.debug("Invoking copy --auto")
-            cmd.append('--auto')
+    # Since we got here - we already have some  transfer_data != "nothing"
+    if (transfer_data == 'auto') or \
+        (
+            force not in ('pushall', 'datatransfer') and
+            ds_repo.config.obtain('datalad.push.copy-auto-if-wanted') and
+            ds_repo.get_preferred_content('wanted', target)
+        ):
+        lgr.debug("Invoking copy --auto")
+        cmd.append('--auto')
 
     if force not in ('pushall', 'datatransfer'):
         # if we force, we do not trust local knowledge and do the checks
--- a/datalad/core/distributed/push.py
+++ b/datalad/core/distributed/push.py
@@ -134,10 +134,10 @@ class Push(Interface):
         force=Parameter(
             # multi-mode option https://github.com/datalad/datalad/issues/3414
             args=("-f", "--force",),
-            doc="""force particular operations, overruling automatic decision
-            making: use --force with git-push ('gitpush'); do not use --fast
-            with git-annex copy ('checkdatapresent'); combine force modes
-            'gitpush' and 'checkdatapresent' ('all').""",
+            doc="""force particular operations, possibly overruling safety
+            protections or optimizations: use --force with git-push ('gitpush');
+            do not use --fast with git-annex copy ('checkdatapresent');
+            combine all force modes ('all').""",
             constraints=EnsureChoice(
                 'all', 'gitpush', 'checkdatapresent', None)),
         recursive=recursion_flag,
--- a/datalad/core/distributed/push.py
+++ b/datalad/core/distributed/push.py
@@ -80,13 +80,6 @@ class Push(Interface):
     dataset.
     << REFLOW ||
 
-    Which files are copied can be further tailored via the
-    'datalad.push.copy-auto-if-wanted' configuration. If set, push will test
-    whether a git-annex "wanted" configuration is present for the target
-    location, and in this case instruct git-annex to obey this configuration
-    when deciding which files to consider for transfer (i.e. use the --auto
-    flag with git-annex copy).
-
     .. note::
       Power-user info: This command uses :command:`git push`, and :command:`git
       annex copy` to push a dataset. Publication targets are either configured
@@ -129,8 +122,12 @@ class Push(Interface):
             transfer of all annexed content, 'nothing' would avoid call to
             `git annex copy` altogether. 'auto' would use 'git annex copy' with
             '--auto' thus transferring only data which would satisfy "wanted"
-            or "numcopies" settings for the remote.""",
-            constraints=EnsureChoice('anything', 'nothing', 'auto')),
+            or "numcopies" settings for the remote (thus "nothing" otherwise).
+            'auto-if-wanted' would enable '--auto' mode only if there is a 
+            "wanted" setting for the remote, and transfer 'anything' otherwise.
+            Note: 'anything' and 'nothing' are "wanted" expressions understood
+            by git-annex.""",
+            constraints=EnsureChoice('anything', 'nothing', 'auto', 'auto-if-wanted')),
         force=Parameter(
             # multi-mode option https://github.com/datalad/datalad/issues/3414
             args=("-f", "--force",),
@@ -188,7 +185,7 @@ class Push(Interface):
             dataset=None,
             to=None,
             since=None,
-            transfer_data='anything',
+            transfer_data='auto-if-wanted',
             force=None,
             recursive=False,
             recursion_limit=None,
@@ -787,8 +784,7 @@ def _push_data(ds, target, content, transfer_data, force, jobs, res_kwargs,
     # Since we got here - we already have some  transfer_data != "nothing"
     if (transfer_data == 'auto') or \
         (
-            force not in ('all', 'checkdatapresent') and
-            ds_repo.config.obtain('datalad.push.copy-auto-if-wanted') and
+            (transfer_data == 'auto-if-wanted') and
             ds_repo.get_preferred_content('wanted', target)
         ):
         lgr.debug("Invoking copy --auto")
--- a/datalad/core/distributed/push.py
+++ b/datalad/core/distributed/push.py
@@ -118,16 +118,16 @@ class Push(Interface):
             constraints=EnsureStr() | EnsureNone()),
         data=Parameter(
             args=("--data",),
-            doc="""what to do with data (annex'ed) data. 'anything' would cause
+            doc="""what to do with (annex'ed) data. 'anything' would cause
             transfer of all annexed content, 'nothing' would avoid call to
             `git annex copy` altogether. 'auto' would use 'git annex copy' with
             '--auto' thus transferring only data which would satisfy "wanted"
             or "numcopies" settings for the remote (thus "nothing" otherwise).
             'auto-if-wanted' would enable '--auto' mode only if there is a 
             "wanted" setting for the remote, and transfer 'anything' otherwise.
-            Note: 'anything' and 'nothing' are "wanted" expressions understood
-            by git-annex.""",
-            constraints=EnsureChoice('anything', 'nothing', 'auto', 'auto-if-wanted')),
+            """,
+            constraints=EnsureChoice(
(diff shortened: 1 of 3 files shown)


mih:
No in-depth review yet, just a bunch of comments after looking at the general changes.

> --force=all instead of --force=pushall in 65aae48 since IMHO forces should be orthogonal and could all be combined

We have discussed that before and came to the conclusion that it *should not* be `all`, because it puts a strong limitation on future changes. Namely that any addition is indeed orthogonal. However, there is no requirement for that. A mode is something that is mutually exclusive to other modes, not that the behavior is fully orthogonal (whatever that even means in detail). What does orthogonality bring as a benefit? Original change was https://github.com/datalad/datalad/pull/4581/commits/e9727d14ce11c7bcb7c6a089d7de19e20d4e0e90

I started looking for a `push` invocation in my workflows that does *not* require `--transfer-data anything` now (or rather again, because that was one motivation to keep this separate and different from `publish`). So far I have not found a single case. I think a command that is meant to transfer things elsewhere should do that by default. I have not manually verified all use of `push` shown in the handbook, but it looks as if they also all need `--transfer-data anything`.

datalad/core/distributed/push.py
```diff
@@ -123,16 +116,27 @@ class Push(Interface):
...
+            args=("--transfer-data",),
+            doc="""what to do with data (annex'ed) data. 'anything' would cause
+            transfer of all annexed content, 'nothing' would avoid call to
+            `git annex copy` altogether. 'auto' would use 'git annex copy' with
+            '--auto' thus transferring only data which would satisfy "wanted"
+            or "numcopies" settings for the remote (thus "nothing" otherwise).
+            'auto-if-wanted' would enable '--auto' mode only if there is a 
+            "wanted" setting for the remote, and transfer 'anything' otherwise.
+            Note: 'anything' and 'nothing' are "wanted" expressions understood
+            by git-annex.""",
```
mih, line 129: What purpose has the "note"? These values are not actually passed on to git-annex, hence it seems the note merely points out an inconsequential similarity.

bpoldrack:
> I started looking for a push invocation in my workflows that does not require --transfer-data anything now (or rather again, because that was one motivation to keep this separate and different from publish). So far I have not found a single case. 

To clarify: Did you try that in this description:

> 'auto-if-wanted' would enable '--auto' mode only if there is a "wanted" setting for the remote, and transfer 'anything' otherwise.

turns out to be incorrect or did you miss it?

datalad/core/distributed/push.py
yarikoptic: It avoids a possible user questions:
- why not just 'all' and 'nothing'? (or any other pair of words)
- why not just 'auto', 'none', 'all' as `publish` has it?

I did not want to unnecessarily expand it here, but  overall docstring above or other docs could be extended to state e.g. that "instead of specifying explicitly `--transfer-data=nothing` to `push`, you can configure the specific remote via `git annex wanted REMOTE nothing` to make no data pushed to that particular remote by default."

yarikoptic:
> > --force=all instead of --force=pushall in [65aae48](https://github.com/datalad/datalad/commit/65aae48ef0cee708a7e0fee89f076b152b64eef2) since IMHO forces should be orthogonal and could all be combined
> 
> We have discussed that before and came to the conclusion that it _should not_ be `all`, because it puts a strong limitation on future changes. Namely that any addition is indeed orthogonal. However, there is no requirement for that. A mode is something that is mutually exclusive to other modes, not that the behavior is fully orthogonal (whatever that even means in detail). What does orthogonality bring as a benefit? Original change was [e9727d1](https://github.com/datalad/datalad/commit/e9727d14ce11c7bcb7c6a089d7de19e20d4e0e90)

Unfortunately I think discussion was not in the issues, but in our weekly meeting. `all` made no sense given non-orthogonal "forces" to start with.  The ultimate and the only combination which made sense was "datatransfer" + "gitpush" which was indeed "bundled" into "pushall" in e9727d14ce11c7bcb7c6a089d7de19e20d4e0e90.  Indeed we had a concern that if yet another mode would be introduced - it would make "all" even less clear.  As now `--force` "flags" are separate from modes, and we could force them upon any action where they might make sense, e.g. `--transfer-data=auto --force=all` to signal that I am using "auto" mode but will force push git branch and do "checkdatapresent".  `--transfer-data=auto --force=pushall` makes it a bit more ambiguous IMHO with "push" in the force name -- "am I now forcing some additional 'push'?" . Overall -- I do not see the point in "push" prefix there after this RF.  But if you feel strongly - we could just revert that change, I hope to never use it one way or another due to it incorporating `git push --force` ;-)

yarikoptic:
> To clarify: Did you try that and this description:
> 
> > 'auto-if-wanted' would enable '--auto' mode only if there is a "wanted" setting for the remote, and transfer 'anything' otherwise.
> 
> turns out to be incorrect or did you miss it?

Yes @bpoldrack observation is correct: default behavior of `push` was not changed in cases where `wanted` was not setup.  It will "do the right thing", and unless some `--force` was used/mentioned, nothing would need to be changed in the handbook.  I only found following uses:
```
docs/basics/101-141-push.rst:  in some cases. Therefore, ``--force datatransfer`` is a slower, but more fail-safe
docs/basics/101-141-push.rst:the ``-f/--force`` option. In the latter case, specifying ``-f no-datatransfer``
```
needing tune up.

datalad/core/distributed/push.py
yarikoptic: but also it probably asks for RFing `publish` to use those values, which could be done for 0.13, possibly just introducing equivalences of `none` == `nothing`, and `all` == `anything`; while issuing deprecation warnings whenever someone uses "old style" values. With `publish` RFing later on (I could attempt it on top of this PR to verify if my assumptions are correct) to use `push` - it would probably be just a matter of changing default from `auto-if-wanted` to `auto` and be done (for the `push` aspect altogether.  Will need to RF also just to call `create-sibling` first ;))

bpoldrack:
Re `pushall` vs `all`:
Agree with @mih here. `--force` should still be "modes" (and it is w/ this PR!) rather than flags and we might need to introduce other force modes in the future leading to `all` not making sense.
For clarity re modes vs flags: Flags would allow for arbitrary combinations. Modes "encode" particular combinations, that you have to choose from. In the state of this PR there isn't much of a difference, because ATM all combinations are available, but it is conceptually different, and it makes a difference for future additions.

mih:
I went through the code again and besides renaming things and adding an additional option, I cannot see which use case is enabled by this PR. I thought that is the key here, that `push` cannot do a particular thing and it is being added here. However, it looks as if it effectively turns the default of `datalad.push.copy-auto-if-wanted` from false to true. Which would invalidate some of the discussions we had, and ultimately led to https://github.com/datalad/datalad/pull/4615

I would appreciate if my note from https://github.com/datalad/datalad/issues/4602#issuecomment-640001236 could be elaborate on in this new context. How and why is this different from 

  `datalad datalad.push.copy-auto-if-wanted=true push`

I have a a few more data moving operations scheduled for next week, and will be able to get some practical experience with this code at that point.

yarikoptic:
There were no auto mode before (which wouldn't transfer unless wanted set), only auto-if-wanted (which would transfer if wanted if set).

bpoldrack:
> I thought that we agreed with @bpoldrack on the value of having them separated from other "--force modes"

Yes, agree. Just wanted some clarity in language. ;-)
Because I also agree with @mih, in that the API by design should be concerned with "modes", that only require to be mutually exclusive. Only that now there are two such options for two independent aspects of `push` behavior. That some of the currently available modes are not only mutually exclusive but also orthogonal in functionality is a different story. The API design should not require that. That's why I agree with that split, but no further.

yarikoptic:
Added a short 7dd99fd271cb30211e8a01f55ca6092c3aec3500 making it into `--data`, will adjust top description, but ok if we decide to not go with it.

datalad/core/distributed/push.py
```diff
@@ -123,16 +116,27 @@ class Push(Interface):
             data or changes for those paths are considered for a push.""",
             nargs='*',
             constraints=EnsureStr() | EnsureNone()),
+        data=Parameter(
+            args=("--data",),
+            doc="""what to do with data (annex'ed) data. 'anything' would cause
```
kyleam, line 121: typo: repeated "data"

datalad/core/distributed/push.py
```diff
@@ -772,13 +782,16 @@ def _push_data(ds, target, content, force, jobs, res_kwargs,
...
-            lgr.debug("Invoking copy --auto")
-            cmd.append('--auto')
+    # Since we got here - we already have some  data != "nothing"
+    if (data == 'auto') or \
+        (
+            (data == 'auto-if-wanted') and
+            ds_repo.get_preferred_content('wanted', target)
+        ):
+        lgr.debug("Invoking copy --auto")
+        cmd.append('--auto')
```
kyleam, line 792: Okay, so ignoring the obvious addition of the `--data` flag, the core change is here.  We switch the decision to pass `--auto` from

```
you didn't explicitly ask to transfer data and you set this config
```

to


```
you didn't explicitly ask to transfer data and you've configured wanted for the remote
```

As far as I can tell, that'd achieve both the behavior that @yarikoptic is after (respect "wanted") and @mih is after (just transfer the data).  It comes with an assumption that there aren't callers that set wanted yet would still prefer `push` to not use `--auto`; sounds pretty safe.

It swaps a `config.obtain` call for a `.get_preferred_content` call per dataset, which seems acceptable given any data transfer should swamp that expense.

datalad/core/distributed/push.py
```diff
@@ -123,16 +116,27 @@ class Push(Interface):
             data or changes for those paths are considered for a push.""",
             nargs='*',
             constraints=EnsureStr() | EnsureNone()),
+        data=Parameter(
+            args=("--data",),
+            doc="""what to do with data (annex'ed) data. 'anything' would cause
+            transfer of all annexed content, 'nothing' would avoid call to
+            `git annex copy` altogether. 'auto' would use 'git annex copy' with
```
kyleam, line 123: Should we drop 'auto' until someone requests it?  It seems no one complained about `publish` not honoring numcopies, so perhaps 'auto-if-wanted' is sufficient.

datalad/core/distributed/push.py
kyleam: I also felt this note was out of place.

datalad/core/distributed/push.py
mih: If we do that, what would remain is identical to `datalad datalad.push.copy-auto-if-wanted=true push` -- according to https://github.com/datalad/datalad/pull/4620#issuecomment-641833047

I started using this code in daily operations, will report.

datalad/core/distributed/push.py
yarikoptic:
```suggestion
            doc="""what to do with (annex'ed) data. 'anything' would cause
```

datalad/core/distributed/push.py
yarikoptic: `auto` mode is to be able (not yet tested) to make `publish` to use `push` since it is it's mode of operation - to not copy data by default, unless wanted is set, and no paths are explicitly specified

datalad/core/distributed/push.py
yarikoptic: ok, I will remove it then.

mih:
So, I have now used it on a few datasets and it didn't break my workflows.

I still feel that this move negates/neglects one of the core aspects of our previous discussions, namely that adding `--auto` whenever there is `wanted`, but it seems there is no way around that.

So feel free to go ahead.

However, it would be good to get tests for the additional functionality that was added. In particular, because we are lacking an actual use case for it, on which it could be evaluated.

yarikoptic:
> However, it would be good to get tests for the additional functionality that was added. In particular, because we are lacking an actual use case for it, on which it could be evaluated.

good point.  Will do.  Also probably would be valuable to test with `paths` being provided to restrict effect of `auto` and subdatasets.

kyleam:
@mih:
> However, it would be good to get tests for the additional functionality that was added. In particular, because we are lacking an actual use case for it, on which it could be evaluated.

Not really a use case, but I've pushed a commit that tests `data='auto'` (basically the test from gh-4541).

@yarikoptic:
>  Also probably would be valuable to test with `paths` being provided to restrict effect of `auto` and subdatasets.

I wasn't able to figure out what the intended effect is here and in general I'm now a bit turned around about what the path argument's role is.

kyleam:
>  Also probably would be valuable to test with `paths` being provided to restrict effect of `auto` and subdatasets.

I've tried to address that with 20d80cf85.

I said:
> I wasn't able to figure out what the intended effect is here and in general I'm now a bit turned around about what the path argument's role is.

My confusion was based on doing something stupid in the setup of an earlier iteration of the test added in 20d80cf85.  I think what's now in that test captures the intended effect.

yarikoptic:
Thank you @kylam -- tests look great to me!  crippledfs failure - filed separate https://github.com/datalad/datalad/issues/4644 (search for the test name didn't point to any existing issue)

yarikoptic:
coverage boost of  (+39.83%)  is especially impressive ;)

yarikoptic:
altogether -- I would just wait for @mih 's final blessing or TODOs ;)

PR #4625, opened 2020-06-10
Make build support code usable by extensions

mih:
- install code as a package
- hack things up to filter out non-extension commands

Not beautiful, but works. Starting point is admittedly also not a beauty.

### DIFF
--- a/_datalad_build_support/setup.py
+++ b/_datalad_build_support/setup.py
@@ -18,7 +18,7 @@ from distutils.errors import DistutilsOptionError
 from distutils.version import LooseVersion
 from genericpath import exists
 from os import linesep, makedirs
-from os.path import dirname, join as opj, sep as pathsep, splitext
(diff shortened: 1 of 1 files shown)


mih:
Now that I am playing with it more, I think this should become a real standalone package. Not all the code in `_datalad_build_support/` built for, or even meant to be used elsewhere.

kyleam:
> I think this should become a real standalone package.

That aligns with the sentiment from gh-3600, though @yarikoptic [mentioned][0] a technical issue related to declaring build dependencies.

[0]: https://github.com/datalad/datalad/pull/3600#discussion_r314364353

mih:
Thanks for the pointer! Will investigate subtree/module...

mih:
This is now replaced by https://github.com/datalad/datalad-extension-template/pull/6

Issue #4626, opened 2020-06-11
Clarify `push --since` docstring


mih:
https://github.com/datalad/datalad/pull/4617#discussion_r437790933

mih:
This is the docstring

> specifies commit-ish (tag, shasum, etc.) from which to look for changes to decide whether pushing is necessary. If '^' is given, the last state of the current branch at the sibling is taken as a starting point.

The linked comment asked for "starting point for what?", but this is clearly defined in the sentence before: "from which to look for changes"

I am closing this issue, as from my POV there is nothing to do here, and given the years that passed since, this viewpoint seems to be shared.

PR #4627, opened 2020-06-12
DOC: push - adjusted description for --since to better describe its purpose

yarikoptic:
Closes gh-4626

Small font: neither eithers nor whethers were hurt as part of this linguistic exercise.

### DIFF
--- a/datalad/core/distributed/push.py
+++ b/datalad/core/distributed/push.py
@@ -113,9 +113,10 @@ class Push(Interface):
             args=("--since",),
             constraints=EnsureStr() | EnsureNone(),
             doc="""specifies commit-ish (tag, shasum, etc.) from which to look for
-            changes to decide whether pushing is necessary.
-            If '^' is given, the last state of the current branch at the sibling
-            is taken as a starting point."""),
+            changes to decide whether pushing of data or recursion into subdatasets
+            (in recursive mode) is necessary.
+            If '^' is given, changes are looked up from the commit of the current
+            branch at the sibling."""),
         path=Parameter(
             args=("path",),
             metavar='PATH',


datalad/core/distributed/push.py
```diff
@@ -113,9 +113,10 @@ class Push(Interface):
             args=("--since",),
             constraints=EnsureStr() | EnsureNone(),
             doc="""specifies commit-ish (tag, shasum, etc.) from which to look for
-            changes to decide whether pushing is necessary.
-            If '^' is given, the last state of the current branch at the sibling
-            is taken as a starting point."""),
+            changes to decide whether pushing of data or recursion into subdatasets
```
kyleam, line 116: Perhaps "transferring data" rather than "pushing of data"?  My thought is that "transfer" would require fewer brain cycles from the reader to connect up with annex data given the option names (force's current `{no-,}datatransfer`), though that of course would carry less weight with the `--transfer-data -> --data` rename in gh-4620.

Then again there's probably no way around that the caller needing to realize that "data" in the context of `push` (and probably lots of other spots in datalad) means "annex data", so there's probably not much point in the making this change.

datalad/core/distributed/push.py
```diff
@@ -113,9 +113,10 @@ class Push(Interface):
...
+            (in recursive mode) is necessary.
+            If '^' is given, changes are looked up from the commit of the current
+            branch at the sibling."""),
```
kyleam, line 119: Hmm, this makes the situation less clear to me.  With the previous version, I would have assumed it was talking about the remote-tracking branch.  With this version, it sounds like the state on the remote may actually be considered.  Perhaps this could be clarified with a "last known state" somewhere in there.

yarikoptic:
it seems my wording was not better to any aspect of it, so let's just close

Issue #4585, opened 2020-05-28
Let's establish a roadmap toward 1.0 release?


yarikoptic:
I think it would benefit us if we kick out 1.0 some time soon (e.g. this year) with some guarantees (so truly follow semver upon those components API changes) of API.  As touched upon in recent call, I would say we could restrict guarantees to the `datalad.core` and `datalad.coreapi` (and only at the top level, i.e. not propagating to `.repo` etc which comes from `.support`). 
ATM I think `coreapi` is too "bloated"

```shell
$> python -c 'import datalad.coreapi as dlcore; print([x for x in dir(dlcore) if not x.startswith("_")])'
['Dataset', 'add_archive_content', 'aggregate_metadata', 'annotate_paths', 'clean', 'clone', 'copy_file', 'create', 'create_sibling', 'create_sibling_github', 'create_sibling_gitlab', 'create_sibling_ria', 'create_test_dataset', 'diff', 'download_url', 'drop', 'export_archive_ora', 'extract_metadata', 'get', 'install', 'ls', 'metadata', 'publish', 'push', 'remove', 'rerun', 'run', 'run_procedure', 'save', 'search', 'siblings', 'sshrun', 'status', 'subdatasets', 'test', 'uninstall', 'unlock', 'update', 'wtf']
```
e.g. all the metadata related functionality (`search` included) should not be there since I expect it to be sooner than later superseded by metalad (approach). Also depending on `publish/push` interactions etc, `publish` might be gone (or not), and similarly `install`.  I still think that we by 1.0 might come up with a solution to avoid breeding `_flavor` commands (i.e. of `create_sibling` family). 

By then we also establish a policy that no new modules/interfaces are added directly  to `.core` while being developed, unless they are released in some MAJOR.MINOR release for a while; and only then moved to `.core` to announce that indeed their API stabilized.

To orchestrate 1.0 release effort I have initiated ["Release 1.0"](https://github.com/datalad/datalad/projects/11#card-39084742) project, while milestones could still be used to annotate any specific issue/pr for "minor" `0.` releases they would actually appear in.

mih:
I am skeptical re "this year", but I think having a roadmap would be instrumental. Thank you for the initiative!

[...]

Issue #4238, opened 2020-03-03
Release management


mih:
Here is what I would consider a sensible release management strategy that is suitable for the workforce we have and is compatible with my personal development pace. Please let me know what you think @yarikoptic @kyleam @bpoldrack @adswa 

For clarity, I am using the example of an upcoming 0.13.0 release for everything below.

1. When the time is ripe a `maint/0.13` branch is split from `master`. From this new branch a 0.13.0 release will be made eventually. This event imposes no constraints on further development in `master`. The timing of this event is either *subjective* (enough new stuff in `master`), or *objective* (a certain time time has passed since the last release). Only bug fixes are allowed into `maint/0.13`, ever. They must not be mixed with API changes, beautifications, or minor additions. It is important to maintain this rule to keep development focused on `master` and the "pressure" to release high.

2. All developers now devote a reasonable fraction of their time to manual (i.e. additional) testing of the `maint/0.13` code. This period is kept reasonably short (e.g. one week, but no more than two) to ensure that the testing effort goes towards quality control of an impending release (otherwise efforts will diffuse and commitment vanish). Newly discovered issues are reported and tagged with a `rc-0.13.0` (release-critical) milestone whenever they reflect regressions with respect to the last release. A initial issue "update changelog for 0.13" is filed, and tagged accordingly. This milestone is **not** populated with existing issues (except for those with severity *critical*, or regressions with respect to the last release) that would also be nice to get fixed. They could have been fixed before the start of the 0.13 release process, but were not, which is enough evidence that a fix is not considered valuable enough to invest, or is not possible with the given resources. This is not to say that these issues should not get fixed, just that the lack of the fix does not impact a 0.13.0 release.

3. All fixes merged via PR into `maint/0.13`. When new issues are discovered the `rc-0.13.0` milestone is continued to be populated, according to the criteria listed in (2). It may be necessary to untag originally release-critical issue. Reasons could be systemic problems that are not manageable at the moment, or developers conclude that the relative importance of an individual fix does not outweigh the need for a release.

4. Whenever the `rc-0.13.0` milestone is empty after the test period in (2), `0.13.0`is tagged in `maint/0.13` and released immediately. There is no need for further deliberation. It will take as long as it takes to fix the release critical bugs.

5. After a `0.13.0` release, the `maint/0.13` branch is kept and continued using the rules in (1). Strictly for fixes only. Further maintenance releases can be made at any time.

[...]

PR #4628, opened 2020-06-12
CONTRIBUTING.md: tune up to section formatting + release etc workflow draft

yarikoptic:
If needed - can split into two PRs (or just merge first commit into master directly) -- first commit is kinda obvious.  
2nd one is aiming to establish some formalization of the workflow so we could reach 1.0 (#4585)  and is largely based on current workflow and ideas in (closed) #4238. It might not yet be complete, and not as detailed as #4238 but I think it would be valuable to agree on these basic items first before expanding/formalizing it further.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -351,7 +351,7 @@ below passes ref names instead of commit IDs, which requires asv v0.3
 or later.)
 
 ```shell
-> asv compare -m hopa 0.9.x master
+> asv compare -m hopa maint master
 
 All benchmarks:
 
@@ -377,9 +377,9 @@ All benchmarks:
 [asv continuous] could be used to first run benchmarks for the to-be-tested
 commits and then provide stats:
 
-- `asv continuous 0.9.x master` - would run and compare 0.9.x and master branches
-- `asv continuous HEAD` - would compare HEAD against HEAD^
-- `asv continuous master HEAD` - would compare HEAD against state of master
+- `asv continuous maint master` - would run and compare `maint` and `master` branches
+- `asv continuous HEAD` - would compare `HEAD` against `HEAD^`
+- `asv continuous master HEAD` - would compare `HEAD` against state of master
 - [TODO: contineous -E existing](https://github.com/airspeed-velocity/asv/issues/338#issuecomment-380520022)
 
 Notes:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -548,6 +548,32 @@ Refer datalad/config.py for information on how to add these environment variable
 
 # Release(s) workflow
 
+## Branches
+
+- `master`: changes toward the next `MAJOR.MINOR.0` release (e.g., `0.13` as of time of writing).
+  It is also used for release candidates (with `rcX` suffix)
+- `maint`: bug fixes and minor enhancements for the latest released `MAJOR.MINOR.PATCH`
+ (e.g., `0.12.7` as of time of writing)
+- `MAJOR.MINOR`: generally not used, unless some bugfix release with a crticial bugfix is needed.
+
+## Workflow
+
+- upon release of `MAJOR.MINOR.0`, `maint` branch need to be fast-forwarded to that release
+- bug fixes to functionality released within the `maint` branch, should (HIGHLY RECOMMENDED) be
+  submitted against `maint ` branch
+- cherry-picking fixes from `master` into `maint` is allowed where needed
+- `master` branch accepts PRs with new functionality
+- `master` branch merges `maint` as frequently as possible/needed 
+
+## Code flow
+
+- new functionality is initially introduced outside of the `datalad.core.` namespace
+- new functionality should mature for at least a single full `MAJOR.MINOR` release until considered
+  for migration to `datalad.core.`
+  - convenience shim imports remain in-place in old locations for at least a single `MAJOR.MINOR` cycle
+  - they should be accompanied with deprecation warnings to trigger downstream code fixups
+  - our internal code should get adjusted immediately upon migration and testing should fail if was not
+
 ## Helpers
 
 [Makefile](./Makefile) provides a number of useful `make` targets:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -553,7 +553,7 @@ Refer datalad/config.py for information on how to add these environment variable
 - `master`: changes toward the next `MAJOR.MINOR.0` release.
   Release candidates (tagged with an `rcX` suffix) are cut from this branch
 - `maint`: bug fixes for the latest released `MAJOR.MINOR.PATCH`
-- `MAJOR.MINOR`: generally not used, unless some bug fix release with a crticial bug fix is needed.
+- `maint-MAJOR.MINOR`: generally not used, unless some bug fix release with a critical bug fix is needed.
 
 ## Workflow
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -550,10 +550,9 @@ Refer datalad/config.py for information on how to add these environment variable
 
 ## Branches
 
-- `master`: changes toward the next `MAJOR.MINOR.0` release (e.g., `0.13` as of time of writing).
-  It is also used for release candidates (with `rcX` suffix)
-- `maint`: bug fixes and minor enhancements for the latest released `MAJOR.MINOR.PATCH`
- (e.g., `0.12.7` as of time of writing)
+- `master`: changes toward the next `MAJOR.MINOR.0` release.
+  Release candidates (tagged with an rcX suffix) are cut from this branch
+- `maint`: bug fixes for the latest released `MAJOR.MINOR.PATCH`
 - `MAJOR.MINOR`: generally not used, unless some bugfix release with a crticial bugfix is needed.
 
 ## Workflow
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,10 +1,8 @@
-Contributing to DataLad
-=======================
+# Contributing to DataLad
 
 [gh-datalad]: http://github.com/datalad/datalad
 
-Files organization
-------------------
+## Files organization
 
 - [datalad/](./datalad) is the main Python module where major development is happening,
   with major submodules being:
@@ -37,8 +35,7 @@ Files organization
   benchmarking of DataLad.  Implemented in any most appropriate language
   (Python, bash, etc.)
 
-How to contribute
------------------
+## How to contribute
 
 The preferred way to contribute to the DataLad code base is
 to fork the [main repository][gh-datalad] on GitHub.  Here
@@ -120,8 +117,7 @@ we outline the workflow used by the developers:
 (If any of the above seems like magic to you, then look up the
 [Git documentation](http://git-scm.com/documentation) on the web.)
 
-Development environment
------------------------
+## Development environment
 
 We support Python 3 only (>= 3.5).
 
@@ -155,8 +151,7 @@ and you will need to install recent git-annex using appropriate for your
 OS means (for Debian/Ubuntu, once again, just use NeuroDebian).
 
 
-Documentation
--------------
+## Documentation
 
 ### Docstrings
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -535,6 +535,32 @@ Refer datalad/config.py for information on how to add these environment variable
 
 # Release(s) workflow
 
+## Branches
+
+- `master`: changes toward the next `MAJOR.MINOR.0` release (e.g., `0.13` as of time of writing).
```
kyleam, line 540: I think the "(e.g., ...)" bit can safely be dropped (same for the instance below).  If it stays, this example tag should have a matching format by appending a ".0".

CONTRIBUTING.md
```diff
@@ -543,10 +533,47 @@ Refer datalad/config.py for information on how to add these environment variable
 - *DATALAD_ALLOW_FAIL*:
   Instructs `@never_fail` decorator to allow to fail, e.g. to ease debugging.
 
-# Changelog section
+# Release(s) workflow
+
+## Branches
+
+- `master`: changes toward the next `MAJOR.MINOR.0` release (e.g., `0.13` as of time of writing).
+  It is also used for release candidates (with `rcX` suffix)
```
kyleam, line 541: Perhaps "Release candidates (tagged with an `rcX` suffix) are cut from this branch".

CONTRIBUTING.md
```diff
@@ -543,10 +533,47 @@ Refer datalad/config.py for information on how to add these environment variable
...
+- `maint`: bug fixes and minor enhancements for the latest released `MAJOR.MINOR.PATCH`
```
kyleam, line 542: I'd prefer to drop the "minor enhancements" part.  Of course, they can be argued on a case by case base, but I don't see any reason to suggest they should go to maint.

CONTRIBUTING.md
```diff
@@ -543,10 +533,47 @@ Refer datalad/config.py for information on how to add these environment variable
...
+ (e.g., `0.12.7` as of time of writing)
+- `MAJOR.MINOR`: generally not used, unless some bugfix release with a crticial bugfix is needed.
```
kyleam, line 544: If we were to need to do this, I'd vote for a "maint-" in front.

CONTRIBUTING.md
```diff
@@ -543,10 +533,47 @@ Refer datalad/config.py for information on how to add these environment variable
...
+
+## Workflow
+
+- upon release of `MAJOR.MINOR.0`, `maint` branch need to be fast-forwarded to that release
+- bug fixes to functionality released within the `maint` branch, should (HIGHLY RECOMMENDED) be
+  submitted against `maint ` branch
+- cherry-picking fixes from `master` into `maint` is allowed where needed
+- `master` branch accepts PRs with new functionality
+- `master` branch merges `maint` as frequently as possible/needed
```
kyleam, line 553: "as possible/needed" seems to be presenting two conflicting options.  FWIW my handling and preference for it is more towards the "needed".

CONTRIBUTING.md
```diff
@@ -543,10 +533,47 @@ Refer datalad/config.py for information on how to add these environment variable
...
+- `master`: changes toward the next `MAJOR.MINOR.0` release (e.g., `0.13` as of time of writing).
+  It is also used for release candidates (with `rcX` suffix)
+- `maint`: bug fixes and minor enhancements for the latest released `MAJOR.MINOR.PATCH`
+ (e.g., `0.12.7` as of time of writing)
+- `MAJOR.MINOR`: generally not used, unless some bugfix release with a crticial bugfix is needed.
+
+## Workflow
+
+- upon release of `MAJOR.MINOR.0`, `maint` branch need to be fast-forwarded to that release
+- bug fixes to functionality released within the `maint` branch, should (HIGHLY RECOMMENDED) be
```
kyleam, line 549: "should" alone seems sufficient.

CONTRIBUTING.md
```diff
@@ -543,10 +533,47 @@ Refer datalad/config.py for information on how to add these environment variable
...
+  submitted against `maint ` branch
+- cherry-picking fixes from `master` into `maint` is allowed where needed
+- `master` branch accepts PRs with new functionality
+- `master` branch merges `maint` as frequently as possible/needed 
+
+## Code flow
```
kyleam, line 555: This seems like it should certainly be in a separate commit, but probably a separate PR, since I don't recall seeing a proposal that matches this.

CONTRIBUTING.md
```diff
@@ -543,10 +533,47 @@ Refer datalad/config.py for information on how to add these environment variable
...
+
+## Branches
+
+- `master`: changes toward the next `MAJOR.MINOR.0` release (e.g., `0.13` as of time of writing).
+  It is also used for release candidates (with `rcX` suffix)
+- `maint`: bug fixes and minor enhancements for the latest released `MAJOR.MINOR.PATCH`
+ (e.g., `0.12.7` as of time of writing)
+- `MAJOR.MINOR`: generally not used, unless some bugfix release with a crticial bugfix is needed.
+
+## Workflow
```
kyleam, line 546: I think it's fine to have this section, but I think more importantly we should resurrect and tweak the text we had in .github/PULL_REQUEST_TEMPLATE.md before 82e9d6a57 (MNT: Drop "Choose a base" item from PR template checklist, 2020-01-16).

CONTRIBUTING.md
kyleam, line 542: You're inconsistent across these entries with your spelling of "bug fix".

CONTRIBUTING.md
```diff
@@ -543,10 +533,47 @@ Refer datalad/config.py for information on how to add these environment variable
...
+- `master` branch merges `maint` as frequently as possible/needed 
+
+## Code flow
+
+- new functionality is initially introduced outside of the `datalad.core.` namespace
+- new functionality should mature for at least a single full `MAJOR.MINOR` release until considered
+  for migration to `datalad.core.`
+  - convenience shim imports remain in-place in old locations for at least a single `MAJOR.MINOR` cycle
+  - they should be accompanied with deprecation warnings to trigger downstream code fixups
+  - our internal code should get adjusted immediately upon migration and testing should fail if was not
```
kyleam, line 562: AFAIK we don't currently have a build that fails if there are internal deprecation warnings.  Within the context of a CONTRIBUTING, I think "testing should fail" will be unclear (is it describing something the contributor needs to actively do? something they should expect will happen with the existing setup? ...).  And writing that makes me think that maybe CONTRIBUTING isn't the best place for this section.

Stepping back, I think we might want to reshape CONTRIBUTING with the goal of making it a bit more useful to first-time contributors, as they are likely the ones that sit down and read it.  I think it'd make sense to find another home for the more detailed reference material that is currently in CONTRIBUTING.

CONTRIBUTING.md
yarikoptic: On one hand - sounds good. On the other 
- we already have such branches for past releases (well - they could be renamed I guess to follow)
- I thought to keep possibility open for the bright future where we might have a more thorough road map, and decide to accept/review prs for future releases, like already happening, without accumulating them unmerged until release. (For that though we would've need to abandon master/maint for clarity). But hey - if such dire need arises, we indeed could rename branches again.

So, I guess I am ok either way. But it reminded me that if we follow current idea - I should add an explicit statement that changes toward release one after the one to come should aim master but will not be merged until release comes, and should be assigned to the corresponding next milestone

CONTRIBUTING.md
yarikoptic: Ok. I could strip it off and postpone (if it feels needing a separate heavy discussion) a pr with it until we settle on previous items.

CONTRIBUTING.md
yarikoptic: Yes, I felt the same somewhat and thought to place it somewhere in docs. But then I also hate searching through docs for the right document and thought that the trail if this document would be appropriate. Also it is somewhat difficult to make a clear separation of what is useful for new vs old contributors. But indeed release management is probably clearly the one which could be elsewhere

CONTRIBUTING.md
yarikoptic: Oi, fresh related topic: https://www.google.com/amp/s/www.zdnet.com/google-amp/article/github-to-replace-master-with-alternative-term-to-avoid-slavery-references/

yarikoptic:
Thank you @kyleam for the review! I will incorporate suggested obvious fixes in short time.
I would have really appreciated if you had used suggestion mode for them (where possible) - that would've allowed for immediate review and acceptance even while on the phone, and having them marked resolved automagically.

CONTRIBUTING.md
kyleam: > * we already have such branches for past releases (well - they could be renamed I guess to follow)

I think ideally the old, inactive branches we have (0.5.x, 0.9.x, 0.11.x) should just be deleted.  If we decide for some reason that we need to make a release off of them, we can branch off of the tag.

> * I thought to keep possibility open for the bright future where we might have a more thorough road map, and decide to accept/review prs for future releases, like already happening, without accumulating them unmerged until release.

If I understand your suggestion, it's something I'd be opposed to, but that's a separate proposal and discussion.

However, re "like already happening": We (in particular me) botched this rc.  I didn't realize you were opposed to releasing `push` in its current form.  We shouldn't have entered the rc phase when things were not settled internally.  It makes what should be a short period drag on indefinitely.

yarikoptic: > However, re "like already happening": 

I had in mind PRs like #4597 which are to be  "scheduled" for the next release

<details>
<summary>I think the `push` part is not really related to this PR thus collapsed here</summary>

> We (in particular me) botched this rc. I didn't realize you were opposed to releasing push in its current form.

Hm... you can prove me wrong, but I thought that I never actively "opposed" - that is why the only issue in 0.13.0 milestone ATM is https://github.com/datalad/datalad/pull/4453 but I will remove it from any milestone since I do not think we have reached or will reach any agreement on that any time soon and it is not really a critical issue.  I did express my dislikes though, and stated that we would likely need to break API if we release `push` in current form.  So I think it was just a common sense and delaying the release while we hadn't agreed on the API.

</details>

kyleam: > I had in mind PRs like #4597 which are to be "scheduled" for the next release

Yes.  My point is that it's being held up by an rc that should not have taken as long as it has.  And the reason it is taking so long is we entered the rc period without realizing there disagreements about `push` in the form it was in at rc1.

yarikoptic:
Dear @kyleam , could you please tune it up to your liking/comments?  I have ran out of time :-(

CONTRIBUTING.md
yarikoptic: was dropped

CONTRIBUTING.md
yarikoptic: I would still vote for no `maint-` prefix for the reasons stated above. So we need one more vote from other @datalad/developers to break the tie.

CONTRIBUTING.md
yarikoptic: This is for `Release Management` section, so not sure how PR template relates to that.

CONTRIBUTING.md
yarikoptic: or -- on a 2nd though -- "whatever!  + we do have `maint` branch used actively , so ok -- let's go with `maint-` prefix". adjusting

CONTRIBUTING.md
yarikoptic: removed in aca2430ddcde1f68bc34e87ab4c7cbd1bf111b6e -- might resurrect it as a follow up PR for a separate discussion

yarikoptic:
From my PoV it is ready for re-review/merge, please chime in @datalad/developers

CONTRIBUTING.md
```diff
@@ -556,10 +546,37 @@ Refer datalad/config.py for information on how to add these environment variable
...
+- `master`: changes toward the next `MAJOR.MINOR.0` release.
+  Release candidates (tagged with an `rcX` suffix) are cut from this branch
+- `maint`: bug fixes for the latest released `MAJOR.MINOR.PATCH`
+- `maint-MAJOR.MINOR`: generally not used, unless some bug fix release with a critical bug fix is needed.
+
+## Workflow
+
+- upon release of `MAJOR.MINOR.0`, `maint` branch need to be fast-forwarded to that release
+- bug fixes to functionality released within the `maint` branch, should be
+  submitted against `maint ` branch
```
adswa, line 562: I found typos

```suggestion
- upon release of `MAJOR.MINOR.0`, `maint` branch needs to be fast-forwarded to that release
- bug fixes to functionality released within the `maint` branch should be
  submitted against `maint` branch
```

PR #4652, opened 2020-06-23
BF: skip running testremote with recent not yet fixed up git-annex

yarikoptic:
See the bugreport

  https://git-annex.branchable.com/bugs/testremote_breeds_way_too_many_instances_of_the_externals_remote/?updated

for more information.

### DIFF
--- a/datalad/distributed/tests/test_ria_basics.py
+++ b/datalad/distributed/tests/test_ria_basics.py
@@ -422,6 +422,13 @@ def _test_gitannex(host, store, dspath):
     init_opts = common_init_opts + ['url={}'.format(store_url)]
     ds.repo.init_remote('store', options=init_opts)
 
+    from datalad.support.external_versions import external_versions
+    if '8.20200330' < external_versions['cmd:annex'] < '8.20200624':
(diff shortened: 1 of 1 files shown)


datalad/distributed/tests/test_ria_basics.py
```diff
@@ -422,6 +422,13 @@ def _test_gitannex(host, store, dspath):
     init_opts = common_init_opts + ['url={}'.format(store_url)]
     ds.repo.init_remote('store', options=init_opts)
 
+    from datalad.support.external_versions import external_versions
+    if '8.20200330' < external_versions['cmd:annex'] < '8.20200624':
+        # https://git-annex.branchable.com/bugs/testremote_breeds_way_too_many_instances_of_the_externals_remote/?updated
```
kyleam, line 427: Looks good to me.  Just for the record: Joey said it happened around 9fa940569c7941ad794637017ae5296e657c7f8d.  The first release to contain that commit is 8.20200501.

yarikoptic: rright - but if someone runs a snapshot (e.g. I had snapshot builds for neurodebian -devel), it would be based on 8.20200330 release.  So I decided to cover more than less.

Issue #4653, opened 2020-06-23
ENH?: make test_utils_cached_dataset tests use local (not networked) dataset


yarikoptic:
ATM relies on reuse of https://github.com/datalad/testrepo--minimalds and thus relies on stable network connection to it etc. We have  `@serve_path_via_http()`, so it might be better to just (re)create such dataset in a global fixture and then serve it.

Meanwhile, will mark all those tests with `@skip_if_no_network`

mih:
Thank you for reporting this issue!

However, at this point I am closing it at a time of ~600 open issues, in an attempt to regain control over the issue tracker. This particular issue was posted more than 6 months ago, has not received a single response, and does not describe an immediate software defect. *Closing this issue does not imply that it is not relevant*. If any reader objects the closing of this issue, please feel more than free to reopen it with an update on how it is relevant for the current state of the project, and which concrete actions to address it are ongoing or planned.

Issue #4621, opened 2020-06-09
Something is flaky in crippled FS workflow


bpoldrack:
Stumbled upon this in PR #4608 :

Crippled FS workflow passed. After squashing the PR and pushing again it suddenly failed, because a `ds.config.get("annex.uuid")` returned `None`. After restarting the job it passed again. I have no idea ATM, but want to take a note here. However, this also lead me to discover, that some tests are using it like `assert_equal(uuid_before, uuid_after)`, which wouldn't raise in such a case.

mih:
I have seen multiple spurious failures over the past days, never the same one, though. :(

bpoldrack:
"Fixed" the spots I could find where we wouldn't notice that we got no `uuid`. I think this is the only thing coming out of this issue.
The actual thing can't be addressed for now w/o more information and reproducibility.

PR #4654, opened 2020-06-23
BF: annotate tests in test_utils_cached_dataset with @skip_if_no_network

yarikoptic:
Since they all rely on access to a github repository ATM.
See https://github.com/datalad/datalad/issues/4653 for a possible RF to avoid that


bpoldrack:
Hm. Crippled FS failure looks unrelated, but then: How does it not happen in PR #4652 ?

kyleam:
> Crippled FS failure looks unrelated, but then: How does it not happen in PR #4652 ?

And on rerun, a different failure, so I guess count these towards the spurious failures mentioned at gh-4621.

PR #4666, opened 2020-06-27
ENH: Let default result renderer tolerate non-path results

mih:
Example use case: credential query
(datalad/datalad-osf#105)

There is nothing on the file system that such credentials are (or have
to be) associated with.

Before this change, the default result handler would crash on results
without a `path` key. After this change, it merely doesn't report a
path at all. Spacing is adjusted be

   `<action>(<status>): [<message>]`

in this case.

### DIFF
--- a/datalad/interface/utils.py
+++ b/datalad/interface/utils.py
@@ -498,9 +498,9 @@ def eval_results(func):
 
 def default_result_renderer(res):
     if res.get('status', None) != 'notneeded':
-        path = res['path']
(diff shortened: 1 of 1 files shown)


mih:
Appveyor failure is coverage submission death.

bpoldrack:
Thx! Was about to do the same thing for this was annoying me elsewhere, too.

mih:
Thx!

PR #4673, opened 2020-07-04
BF: Do not fail result rendering with no `path` but `refds`

mih:
The employed `relpath` call cannot handle it.

Also remove the condition that triggered the specific observed crash
that prompted this change -- push was not reporting the dataset path
in question for one of its errors.


mih:
Thx @kyleam !

PR #4674, opened 2020-07-04
BF: Do not hide `git-push` crashes

mih:
Previous setup assumed that any `git-push` error is an orderly
report that could be forensically analyzed. However, in particular
with custom remote helpers, `git-push` can also just crash, and
there is no guarantee that we can inspect anything.

This is a minimal change that makes sure the `CommandError` bubbles
up, when we know that something is fundamentally not right, because
there was an error, but the expected error message channel is empty.

In all likelihood that is still far from comprehensive, but still better
than before.


mih:
Thx @kyleam !

PR #4675, opened 2020-07-04
OPT: Avoid unnecessary `push --dryrun`

mih:
This change adds the required conditionals to re-use the results
of an identical `push --dry-run` that figures out which refs to push
-- that always did already happen, when the remote itself was discovered
via a dry-run.


mih:
Thx @kyleam !

PR #4682, opened 2020-07-07
ENH(UX): explicitly disallow empty string for --since in push

yarikoptic:
I had forgotten that it got a new magical `^` to operate since the last published state, so ran it (`datalad push --data auto --since= --to=datalad-public -r`) and was confused with all the progress bars (about to transfer 140G etc... disappeared quickly though... some separate issue probably to demonstrate/file) and only then remembered:

```
  --since SINCE         specifies commit-ish (tag, shasum, etc.) from which to
                        look for changes to decide whether pushing is
                        necessary. If '^' is given, the last state of the
                        current branch at the sibling is taken as a starting
                        point. Constraints: value must be a string

```

but I wonder if we should explicitly check if value is not an empty string (not `None`) and fail

Positioned against `maint` since an empty string doesn't match documented allowed values ;-)

### DIFF
--- a/datalad/core/distributed/push.py
+++ b/datalad/core/distributed/push.py
@@ -190,6 +190,10 @@ class Push(Interface):
             recursive=False,
             recursion_limit=None,
             jobs=None):
+        # push uses '^' to annotate the previous pushed committish, and None for default
+        # behavior. '' was/is (to be deprecated) used in `publish`. Alert user about the mistake
+        if since == '':
+            raise ValueError("'since' should point to commitish or use '^'.")
         # we resolve here, because we need to perform inspection on what was given
         # as an input argument further down
         paths = [resolve_path(p, dataset) for p in assure_list(path)]
--- a/datalad/core/distributed/tests/test_push.py
+++ b/datalad/core/distributed/tests/test_push.py
@@ -75,6 +75,12 @@ def test_invalid_call(origin, tdir):
         ValueError,
         ds.push, to='target', since='09320957509720437523')
 
+    # If a publish() user accidentally passes since='', which push() spells as
+    # since='^', the call is aborted.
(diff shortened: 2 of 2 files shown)


kyleam:
In my view this is a pure bug fix.

I think ideally the check you're adding would be done a bit downstream:

https://github.com/datalad/datalad/blob/9744e8ae625f34c72cdd099351cc566598ba37b2/datalad/core/distributed/push.py#L227-L236

We could add a separate arm for the empty string, but IMO a better option is to `s/elif since/elif since is not None/` and let the empty string be handled like all other invalid values.

yarikoptic:
Sounds good, I will move or feel welcome to do so, I am unlikely to get to it today

kyleam:
> Sounds good, I will move or feel welcome to do so, I am unlikely to get to it today

Done.  Of course feel free to drop any of the changes I pushed.

datalad/core/distributed/push.py
```diff
@@ -235,7 +231,7 @@ def __call__(
                 lgr.info(
                     "No tracked remote for active branch, "
                     "detection of last pushed state not in effect.")
-        elif since:
+        elif since is not None:
             # will blow with ValueError if unusable
             ds_repo.get_hexsha(since)
```
yarikoptic, line 236: indeed it does guard but I think we could improve UX with a more informative message above instead of
```shell
$> datalad push --since=''   
[ERROR  ] Unknown commit identifier:  [gitrepo.py:format_commit:1721] (ValueError) 
```

kyleam: I quite like that uniform treatment, but I thought you might not.  As I said, feel free to drop what you don't want and add whatever extra code you'd like.

PR #4683, opened 2020-07-07
RF: make publish understand --since=^ and deprecate --since=

yarikoptic:
Positioned against `maint` since 
- previous code should still work, we will just announce (well, DeprecationWarnings are not shown! by default, heh) deprecation
- it would improve my UX while switching from `publish` to `push` (although #4682 would also help) so I could just use `--since=^` in both

Who knows, may be in 0.15 the entire `publish` would be gone...

### DIFF
--- a/datalad/distribution/publish.py
+++ b/datalad/distribution/publish.py
@@ -623,11 +623,11 @@ class Publish(Interface):
         since=Parameter(
             args=("--since",),
             constraints=EnsureStr() | EnsureNone(),
-            doc="""When publishing dataset(s), specifies commit (treeish, tag, etc)
-            from which to look for changes
-            to decide whether updated publishing is necessary for this and which children.
-            If empty argument is provided, then we would take from the previously 
-            published to that remote/sibling state (for the current branch)"""),
+            doc="""specifies commit-ish (tag, shasum, etc.) from which to look for
+            changes to decide whether pushing is necessary.
+            If '^' is given, the last state of the current branch at the sibling
+            is taken as a starting point. An empty string ('') for the same effect is
+            still supported but will be removed in 0.15.0).."""),
         # since: commit => .gitmodules diff to head => submodules to publish
         missing=missing_sibling_opt,
         path=Parameter(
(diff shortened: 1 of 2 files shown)


kyleam:
I have no objections to making `publish` understand '^' for consistency.  As for applying it to maint, all right, though I think we should avoid using "it doesn't break" (should be true for nearly all code for master too) and "improves things for me" as reasons to make non-bugfix changes to maint.

However, I really don't think we should introduce deprecations on maint unless we have a very strong reason to.  In this case, I think publish should just learn to treat '^' as another spelling of ''.  Continuing to support the latter has virtually no maintenance cost.

> Who knows, may be in 0.15 the entire `publish` would be gone...

It shouldn't be.  We haven't deprecated it yet, so that'd be at least 0.16.

yarikoptic:
Re no maintenance cost - there is UX cost due to more than one way to do the same, confusion etc. I wish I just did it for 0.13, where it would have been right in place.. heh heh... Also iirc now there is datalad diff which might need the same (didn't check)

kyleam:
> there is UX cost due to more than one way to do the same, confusion etc.

Yes, ideally an interface wouldn't have multiple ways to say the same thing.  But in this case the tradeoff is that it allows consistency between `publish` and `push` without making users or us have to go through a deprecation sequence.

yarikoptic:
>  But in this case the tradeoff is that it allows consistency between publish and push without making users or us have to go through a deprecation sequence.

it wouldn't be "consistent" unless we also allow empty string for `push`... the sooner we encourage (and really weakly since that deprecation warning wouldn't be shown by default) users to adhere to consistent interface, the better IMHO.

kyleam:
> > But in this case the tradeoff is that it allows consistency between publish and push without making users or us have to go through a deprecation sequence.
> 
> it wouldn't be "consistent" unless we also allow empty string for `push`...

To reword my statement in a less sloppy, context-dependent way: "it allows users to consistently say --since=^ across `publish` and `push`".

> the sooner we encourage (and really weakly since that deprecation warning wouldn't be shown by default) users to adhere to consistent interface, the better IMHO.

I disagree.  I'll leave it at that, since I'd just repeat my statements from above.

yarikoptic:
@bpoldrack @mih @adswa , any opinion? Deprecate or not --since= in publish?

bpoldrack:
I find @kyleam's argument convincing.

yarikoptic:
Removed deprecation.

Issue #3988, opened 2020-01-05
Establish core `*Repo` classes


mih:
Following the plan from https://github.com/datalad/datalad/issues/3192 I want to approach the move of the `Dataset` and `*Repo` classes under `datalad/core` in the source tree.

Having done a fair amount of digging through the code layers over the past few months, I am playing with an idea already mentioned in #3192:

> move core dataset and repo classes into dataset/ (or rather core/dataset/). it might make sense to split Git/AnnexRepo into a core and an extended interface also (via inheritance), the number of rarely used methods is substantial ATM

While from a technical angle this may cause more trouble then benefits, I believe the potential for strategic development is substantial. I am thinking about issues like #3714 that are both hard to see and hard to tackle, because everything is so intertwined.

So what about the following plan:

- establish new base classes for `GitRepo`, `AnnexRepo`, `Dataset`, each with the name prefix `Core`, and place them, and only these new classes under `datalad/core`. The original classes will then derive from the new bases while keeping there API 100% intact.
- adjust all code under `datalad/core` to exclusively use the new classes, and populate them with any functionality that is actually required by code in `datalad/core`. When moving code to the base classes, make sure strategic development goals are met (GitPy removal, `normalize_paths` removal, result communication rather than error handling in low-level code, etc.)

This is likely to become a lengthy process (slow transition of code from super into the base class), but it should be a process that minimizes the impact on users and functionality, while still providing the wiggle room necessary to tackle a number of long-standing issues.

What do you think @datalad/developers ?

bpoldrack:
I think, this is generally the right way to go. However, as you pointed out in the second point, it should involve a major review of those classes, definition of the goals, etc.
My issue here is, that this might well lead to a redesign that would simply not have the ´*Repo` classes. I'm not yet sure about this, but it seems to me, that - while being intuitive - subclassing `GitRepo` for `AnnexRepo` isn't particularly useful.

I'm not clear whether start transitioning to core that way is helping that process or making it harder.

yarikoptic:
a minor comment on one aspect:

> * establish new base classes for `GitRepo`, `AnnexRepo`, `Dataset`, each with the name prefix `Core`, and place them, and only these new classes under `datalad/core`. The original classes will then derive from the new bases while keeping there API 100% intact.

[...]

Issue #4589, opened 2020-05-29
Decide on destiny of .repo (and *Repos)


yarikoptic:
Before we discuss details I see two major possible options to try to decide on for 1.0 under assumption that `.core` will be providing a stable API:

- :+1: make `Dataset` s `.repo` protected (`._repo`) to hide the corpse and do not worry much about *Repos
- :smile: keep `.repo` "public" but make it use only `core.**repo.py` classes
- :-1: I don't care

I am not considering following option(s) as viable:
- Document that `.repo` is supposed to be considered "protected" and not used outside of the DataLad code base

Please vote. If there is another high level approach I have missed, please add

mih:
Related re core Repo classes: https://github.com/datalad/datalad/issues/3988

yarikoptic:
well, decision was kinda made ;-)

PR #4689, opened 2020-07-11
Core Repo/Dataset classes

mih:
Merely the first steps towards a resolution of gh-3988 in the spirit of gh-3192, hopefully and eventually leading to a sliming/modernization of the public API (gh-4589).

There is not much happening here, other than moving code around and establishing the classes. The plan is to lower the threshold to transition some carefully filtered functionality in the core API. At the same time the main Repo classes will be used primarily for a long time still. But it should become progressively easier to write code that only uses the core API.

Major pieces:

- core repo classes `GitRepo` and `AnnexRepo` (i.e. identically named as suggested in https://github.com/datalad/datalad/issues/3988#issuecomment-571241784)
- `call_git*()` methods have moved and no longer rely on an execution helper that invokes `normalize_paths()`
- transition a minimal additional set of functionality to make the flyweight pattern work with the core API only
- not moving code that directly uses `wrapt`

Fixes gh-3988

I am currently pondering whether we might also want to established corresponding locations for core `utils`, `exceptions` and `support` with the analog goal of a rejuvenation/slimming of the code base. But later and in a separate effort.


mih:
This is now about as far as I'd go for a meaningful first step.

datalad/core/dataset/dataset.py
```diff
@@ -0,0 +1,267 @@
...
+    def close(self):
+        """Perform operations which would close any possible process using this Dataset
+        """
+        repo = self._repo
+        self._repo = None
+        if repo:
+            # might take care about lingering batched processes etc
+            del repo
+
+    @property
```
bpoldrack, line 163: I wonder whether we want to put `path` into core. We prob. should still aim to switch to `pathlib` entirely. If `path` remains in core, we would need to break it later or not get rid of `os.path`. How about renaming to `_std_path` or something like that as a compromise to allow for this move w/o the need to do all the pathlib related work first while making it (at least conventionally) clear, that this is nothing an user should rely on?

datalad/core/dataset/gitrepo.py
```diff
@@ -0,0 +1,519 @@
...
+    def __init__(self, path):
+        """Creates representation of git repository at `path`.
+
+        Parameters
+        ----------
+        path: str
+          path to the git repository; In case it's not an absolute path,
+          it's relative to PWD
+        """
+        self.path = path
```
bpoldrack, line 89: See above. `path` related stuff would need to involve all three classes, of course.

datalad/core/dataset/gitrepo.py
```diff
@@ -0,0 +1,531 @@
...
+        for line in out.splitlines():
+            props = line.split('\0')
+            if len(fields) != len(props):
+                raise RuntimeError(
+                    'expected fields {} from git-for-each-ref, but got: {}'.format(
+                        fields, props))
+            yield dict(zip(fields, props))
+
+    def _call_git(self, args, files=None, expect_stderr=False, expect_fail=False,
+                  honor_fake_dates=True, cwd=None, env=None, index_file=None):
```
kyleam, line 288: Is `honor_fake_dates=True` worth adding at this point?  Unlike the `check_fake_dates` counterpart in `_git_custom_command`, this doesn't look to be used anywhere.

The comments at <https://github.com/datalad/datalad/pull/3791#discussion_r336722413> might be relevant, depending on what the motivation for this parameter is.

datalad/core/dataset/gitrepo.py
```diff
@@ -0,0 +1,531 @@
...
+          rather than as part of `args` is that the call will be split into
+          multiple calls to avoid exceeding the maximum command line length.
+        expect_stderr : bool, optional
+          Standard error is expected and should not be elevated above the DEBUG
+          level.
+        expect_fail : bool, optional
+          A non-zero exit is expected and should not be elevated above the
+          DEBUG level.
+        honor_fake_dates : bool, optional
+          Whether to actually fake dates for Git operations, if enable for a
```
kyleam, line 308: s/enable/&d/

datalad/core/dataset/gitrepo.py
```diff
@@ -0,0 +1,531 @@
...
+        CommandError if the call exits with a non-zero status.
+        """
+        cmd = ['git'] + self._GIT_COMMON_OPTIONS + args
+
+        if honor_fake_dates and self.fake_dates_enabled:
+            env = self.add_fake_dates(env)
+
+        if index_file:
+            env = (env if env is not None else environ).copy()
+            env['GIT_INDEX_FILE'] = index_file
```
kyleam, line 327: AFAICS `index_file` is unused.  Should `_call_git` drop the parameter?

datalad/core/dataset/gitrepo.py
```diff
@@ -0,0 +1,531 @@
...
+        variables for git and git-annex set.
+        """
+        env = (env if env is not None else environ).copy()
+        if 'DATALAD_FAKE_DATE' in env:
+            # there is already a fake date setup
+            return env
+        # prevent infinite recursion, should this function be called again
+        # inside the next call.
+        environ['DATALAD_FAKE_DATE'] = '1'
+        last_date = list(self.for_each_ref_(
```
kyleam, line 494: Note mostly to myself: Okay, so this wasn't problem before because the `for_each_ref_` call led to a `_git_custom_command` call with `check_fake_dates=False`.

datalad/core/dataset/gitrepo.py
```diff
@@ -0,0 +1,531 @@
...
+            fields='committerdate:raw',
+            count=1,
+            pattern='refs/heads',
+            sort="-committerdate",
+        ))
+        del environ['DATALAD_FAKE_DATE']
```
kyleam, line 500: In case a Python caller happens to catch an exception (most likely unintentionally) that is raised by `for_each_ref_`, it might be worth guarding this "DATALAD_FAKE_DATE" set/delete with a try-finally.

datalad/core/dataset/gitrepo.py
```diff
@@ -0,0 +1,531 @@
...
+
+        lgr.debug("Setting date to %s",
+                  time.strftime("%a %d %b %Y %H:%M:%S +0000",
+                                time.gmtime(seconds_new)))
+
+        env["GIT_AUTHOR_DATE"] = date
+        env["GIT_COMMITTER_DATE"] = date
+        env["GIT_ANNEX_VECTOR_CLOCK"] = str(seconds_new)
+        # leave a marker to prevent duplicate processing
+        env["DATALAD_FAKE_DATE"] = date
```
kyleam, line 520: Hmm, do you know a scenario where this marker is used?  Testing with the patch below, which should fail if this value makes it to the other side, all `datalad.core` and `datalad.support.tests.test_gitrepo` tests passed on my end when running with `export DATALAD_FAKE__DATES=1`.

```diff
diff --git a/datalad/core/dataset/gitrepo.py b/datalad/core/dataset/gitrepo.py
index 8780ab713..c67133793 100644
--- a/datalad/core/dataset/gitrepo.py
+++ b/datalad/core/dataset/gitrepo.py
@@ -486,6 +486,8 @@ def add_fake_dates(self, env):
         """
         env = (env if env is not None else environ).copy()
         if 'DATALAD_FAKE_DATE' in env:
+            if env['DATALAD_FAKE_DATE'] != '1':
+                assert 0
             # there is already a fake date setup
             return env
         # prevent infinite recursion, should this function be called again
```

If the marker does have an effect in some cases, I'm concerned that it could break the intended behavior of ticking the faked date on each commit-producing operation, but perhaps I'm thinking about it incorrectly.

datalad/support/gitrepo.py
```diff
@@ -1467,137 +1317,12 @@ def _get_prefixed_commit_msg(msg):
...
-            last_date = last_date[0]['committerdate:raw'].split()[0]
-            seconds = int(last_date)
-        else:
-            seconds = self.config.obtain("datalad.fake-dates-start")
-        seconds_new = seconds + 1
-        date = "@{} +0000".format(seconds_new)
-
-        lgr.debug("Setting date to %s",
-                  time.strftime("%a %d %b %Y %H:%M:%S +0000",
-                                time.gmtime(seconds_new)))
```
kyleam, line 1593: This is the last use of `time` in support.gitrepo.

datalad/core/dataset/gitrepo.py
mih: I think you are right. I will (try to) remove it.

datalad/core/dataset/gitrepo.py
mih: It should. I will try.

datalad/core/dataset/gitrepo.py
mih: Correct.

datalad/core/dataset/gitrepo.py
mih: It seems that is a left-over from an attempt of mine to fix something. I will look into removing it.

datalad/support/gitrepo.py
mih: Thx!

datalad/core/dataset/dataset.py
mih: I think you are right, I will try to remove it.

datalad/core/dataset/gitrepo.py
mih: Would go away with the removal of the parameter.

mih:
I pushed a series of commits that address all raised points. I will evaluate these changes in the context of #4698 now.

datalad/core/dataset/gitrepo.py
mih: Done

datalad/core/dataset/gitrepo.py
mih: Done

datalad/core/dataset/gitrepo.py
mih: Done

datalad/core/dataset/gitrepo.py
mih: Done

datalad/core/dataset/gitrepo.py
mih: So this is needed. Or rather it is the way I chose to prevent infinite recursion (the next few lines internally lead to a `call_git()` call that leads to `add_fake_dates()` being called again.

datalad/support/gitrepo.py
mih: Done

datalad/core/dataset/gitrepo.py
kyleam: Could you provide me an example of how to trigger it?  As I mentioned, at least with the iteration of the PR I was looking at, I could not, and I'm worried it's an unintended change in behavior.

datalad/core/dataset/dataset.py
mih: Done

datalad/core/dataset/gitrepo.py
mih: I am sorry. I was looking at the diff in the comment, and not at the line you commented on. That lines is removed, and was not needed. My statement referred to https://github.com/datalad/datalad/pull/4689/files#diff-bd096260933e4f20f182a597bd90fe7dR477 being needed. If you remove it, it will trigger an infinite recursion. But that was not what motivated your original comment. Sorry for the confusion.

kyleam: Ah, okay great.  Yep, that one I understand.  Thanks.

bpoldrack:
Seems you stripped `path` from `GitRepo` but not from `Dataset`. Oversight or is there a reason for it?

mih:
Rebased on master after an eternity has passed.

mih:
This will need to be attempted at a later point again. Too much has changed.

Issue #4667, opened 2020-06-28
`push` inefficient due to multiple `git-push` calls


mih:
`push` of an annex repo will perform to separate `git push` calls. Once for all normal branches, and afterwards a dedicated push of the `git-annex` branch. This is probably good for detecting conflicts. Comments state:

```
# push the main branches of interest first, but not yet (necessarily)
# the git-annex branch. We want to push first in order to hit any
# conflicts or unknown history before we move data. Otherwise out
# decision making done above (--since ...) might have been
# inappropriate.
```

However, whenever a `git push` is not as efficient as with a real git remote (e.g. with https://github.com/datalad/git-remote-rclone, and also git-remote-osf that uses a similar approach), each push can be an expensive maneuver.

We should investigate whether two `push` calls are really necessary.

mih:
I could not convince myself that there is any benefit from splitting the branch pushes in the way we are doing them. However, fixing that would not be enough, because ATM `GitRepo.push()` can only handle the push of a single ref. Optimizing that is a precondition for making `datalad push` faster.

mih:
Turns out that `GitRepo.push()` can do multiple refspecs at once. The documentation just made it sound it would not.

PR #4692, opened 2020-07-12
OPT: Avoid multiple calls to `git-push` with multiple refspecs

mih:
As noted in gh-4667, `GitRepo.push()` can actually do all at once.
It just claimed not to be able to. Fixed the docs, and adjusted
the call to the method in the `push()` command.


mih:
I am not insisting on a merge into `maint`, hence #4693 has it too, but I thought it would be trivial enough to propose it.

mih:
Thx @yarikoptic @kyleam

PR #4693, opened 2020-07-12
OPT: Consolidate on a single `git-push` call in `push()`

mih:
Instead of splitting between regular branches and the git-annex branch, in order to improve efficiency -- in particular with expensive operations in custom git-remote-helper implementations.

Fixes gh-4667

This PR includes #4692 which is a related, but more focused optimization that was proposed for inclusion into `maint`.


mih:
Awesome! Thx @kyleam !

Issue #4670, opened 2020-07-02
AppVeyor codecov submission failing regular


kyleam:
In the past, codecov submission errors on AppVeyor were common but didn't happen every time.  The submission is now failing regularly and in a style I don't recall seeing until recently (a lot of the error message is one-character per line).  Here's the latest example from gh-4669:

https://ci.appveyor.com/project/mih/datalad/builds/33876628/job/5g5p2qmsckpupkcw#L1923

kyleam:
Fixed by gh-5223 (thanks @yarikoptic).

PR #4694, opened 2020-07-12
TST: Disable coverage submission on appveyor

mih:
ATM it fails unconditionally, and obscures the reporting of actual test
failures. Until this can be investigated there is no point in keeping it
fail.


yarikoptic:
sounds good to me. Bye bye 0.34% of coverage!

kyleam:
Thanks @mih.

re: gh-4670 (for posterity and to avoid someone looking at that issue and concluding it fixed itself)

PR #4696, opened 2020-07-13
RF/OPT: just assign None to linux_distribution*, and False to on_debian_wheezy

yarikoptic:
Those are too expensive to assign a proper value on recent Pythons where
we have to use distro (see https://github.com/datalad/datalad/issues/4690).

on_debian_wheezy was not used anywere in this code base and cannot be True these days
anyways since wheezy EOLed 2 years back, and we do not even provide backports of datalad
for it from neurodebian.

The other variables were used only for a better git-annex installation
suggestion message -- we might want to get away from that and just refer to
installation instructions anyways, but or now I just kept it around.

Even though these changes can possibly considered a "breakage", I think the
actual breakage of anything is very unlikely since code should not really rely in any
critical functionality on what distribution and its release it is running on.

I had kept variables with None values to really avoid any breakage, and so we could in
0.14 announce their proper deprecation cycle and remove them later on

### DIFF
--- a/datalad/utils.py
+++ b/datalad/utils.py
@@ -34,7 +34,10 @@ import wrapt
 
 from copy import copy as shallow_copy
 from contextlib import contextmanager
-from functools import wraps
+from functools import (
+    lru_cache,
+    wraps,
+)
 from time import sleep
 import inspect
 from itertools import tee
@@ -77,6 +80,8 @@ on_msys_tainted_paths = on_windows \
                         and os.environ.get('MSYSTEM', '')[:4] in ('MSYS', 'MING')
 
 
+# Takes ~200msec, so should not be called at import time
+@lru_cache()  # output should not change through life time of datalad process
 def get_linux_distribution():
     """Compatibility wrapper for {platform,distro}.linux_distribution().
     """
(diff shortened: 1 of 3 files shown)


datalad/utils.py
```diff
@@ -91,16 +96,12 @@ def get_linux_distribution():
...
-    on_debian_wheezy = on_linux \
-                       and linux_distribution_name == 'debian' \
-                       and linux_distribution_release.startswith('7.')
-except:  # pragma: no cover
-    # MIH: IndexError?
-    on_debian_wheezy = False
-    linux_distribution_name = linux_distribution_release = None
+# TODO: deprecated, remove in 0.15
+# Wheezy EOLed 2 years ago, no new datalad builds from neurodebian
+on_debian_wheezy = False
```
bpoldrack, line 101: Hm. I don't know where this is actually used. `git grep` says nowhere, but I have no clue about building debian package. However, if it is used anywhere, it's a bit weird to claim `on_debian_wheezy` is `False`. Either the name is suggesting the wrong thing ("Are we on wheezy?") or it should not report a non-factual thing. And if it's not used anywhere in the build process, then just remove it entirely?

yarikoptic: It is "factual" (with 99.9% confidence) since it **very** unlikely anyone would run it on EOLed wheezy. We aren't even compatible with its python3 if it had any(next one, oldoldstable jessie had 3.4).
I didn't want to just remove it since  if some (elderly) code in some extension or 3rd party code DOES use it for some reason, it would fail -- not something we would like to cause by updates in `maint` branch.  We will remove it in some next release though (as comment says).

bpoldrack: True. I didn't notice it's against `maint`.
I'm fine with deprecating and later remove, of course. Although w/o any deprecation warning I'd be surprised if anyone actually using it notices the deprecation. So, in my estimation the removal will be as surprising as it would be in the next minor release. Anyways - don't consider this a blocker.

kyleam: In my view this variable, as well as `linux_distribution_name` and `linux_distribution_release`, are obscure and minor enough that just a mention in the 0.14.0 changelog would be fine.  

Anyway, I think we all agree that the specifics of how we handle this is unlikely to matter, and delaying/caching `get_linux_distribution` is valuable.  Will merge.

PR #4698, opened 2020-07-13
GitWitless-based call_git() and wipe out _git_custom_command()

mih:
This PR sits on top of #4689 and takes things a bit further:

- Use `GitWitlessRunner` for `call_git*()` methods
- Replace `_git_custom_command()` with `call_git*()` calls

In principle this is ready, but it shows some disturbing randomness in its behavior. The source for that is likely already present in 0.13, but now we are using this relatively new code much much more frequently for all kinds of tiny calls.

At present I have no idea where to start looking.

<details>
<summary>Diff to #4689 </summary>

```diff
diff --git a/datalad/core/dataset/gitrepo.py b/datalad/core/dataset/gitrepo.py
index 8780ab713..e37cf42dd 100644
--- a/datalad/core/dataset/gitrepo.py
+++ b/datalad/core/dataset/gitrepo.py
@@ -34,8 +34,9 @@ from datalad.support.exceptions import (
     InvalidGitRepositoryError,
 )
 from datalad.cmd import (
-    GitRunner,
+    GitWitlessRunner,
     run_gitcommand_on_file_list_chunks,
+    StdOutErrCapture,
 )
 
 lgr = logging.getLogger('datalad.core.dataset.gitrepo')
@@ -100,7 +101,7 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
         # ['-c', 'receive.autogc=0', '-c', 'gc.auto=0']
         self._GIT_COMMON_OPTIONS = []
 
-        self._cmd_call_wrapper = GitRunner(cwd=path)
+        self._git_runner = GitWitlessRunner(cwd=path)
 
         # Set by fake_dates_enabled to cache config value across this instance.
         self._fake_dates_enabled = None
@@ -285,7 +286,7 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
             yield dict(zip(fields, props))
 
     def _call_git(self, args, files=None, expect_stderr=False, expect_fail=False,
-                  honor_fake_dates=True, cwd=None, env=None, index_file=None):
+                  honor_fake_dates=True, index_file=None):
         """Allows for calling arbitrary commands.
 
         Internal helper to the call_git*() methods.
@@ -317,30 +318,35 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
         ------
         CommandError if the call exits with a non-zero status.
         """
+        runner = self._git_runner
+        stderr_log_level = {True: 5, False: 11}[expect_stderr]
+
[...]

PR #4699, opened 2020-07-13
RF: Use WitlessRunner for config manager

mih:
Includes 109ac04 from #4703 that exposed a critical flaw in `GitWitlessRunner` and a limitation in `WitlessRunner`

### DIFF
--- a/datalad/support/external_versions.py
+++ b/datalad/support/external_versions.py
@@ -42,24 +42,31 @@ class UnknownVersion:
 #
 # Custom handlers
 #
-from datalad.cmd import Runner
-from datalad.cmd import GitRunner
+from datalad.cmd import (
+    WitlessRunner,
+    GitWitlessRunner,
+    StdOutErrCapture,
+)
 from datalad.support.exceptions import (
     MissingExternalDependency,
     OutdatedExternalDependency,
 )
-_runner = Runner()
-_git_runner = GitRunner()
+_runner = WitlessRunner()
+_git_runner = GitWitlessRunner()
 
 
 def _get_annex_version():
     """Return version of available git-annex"""
     try:
-        return _runner.run('git annex version --raw'.split())[0]
+        return _runner.run(
+            'git annex version --raw'.split(),
+            protocol=StdOutErrCapture)['stdout']
     except CommandError:
         # fall back on method that could work with older installations
-        out, err = _runner.run(['git', 'annex', 'version'])
-        return out.split('\n')[0].split(':')[1].strip()
+        out = _runner.run(
+            ['git', 'annex', 'version'],
+            protocol=StdOutErrCapture)
+        return out['stdout'].split('\n')[0].split(':')[1].strip()
 
 
 def _get_git_version():
@@ -81,7 +88,9 @@ def _get_bundled_git_version():
     """
     path = _git_runner._get_bundled_path()
     if path:
-        out = _runner.run([op.join(path, "git"), "version"])[0]
+        out = _runner.run(
+            [op.join(path, "git"), "version"],
+            protocol=StdOutErrCapture)['stdout']
         # format: git version 2.22.0
         return out.split()[2]
 
@@ -92,30 +101,31 @@ def _get_system_ssh_version():
     Annex prior 20170302 was using bundled version, but now would use system one
     if installed
     """
-    out, err = _runner.run('ssh -V'.split(),
-                           expect_fail=True, expect_stderr=True)
+    out = _runner.run(
+        'ssh -V'.split(),
+        protocol=StdOutErrCapture)
     # apparently spits out to err but I wouldn't trust it blindly
-    if err.startswith('OpenSSH'):
-        out = err
-    assert out.startswith('OpenSSH')  # that is the only one we care about atm
-    return out.split(' ', 1)[0].rstrip(',.').split('_')[1]
+    stdout = out['stdout']
+    if out['stderr'].startswith('OpenSSH'):
+        stdout = out['stderr']
+    assert stdout.startswith('OpenSSH')  # that is the only one we care about atm
+    return stdout.split(' ', 1)[0].rstrip(',.').split('_')[1]
 
 
 def _get_system_7z_version():
     """Return version of 7-Zip"""
-    out, err = _runner.run(
-        ['7z'], expect_fail=True, expect_stderr=True,
-    )
+    out = _runner.run(
+        ['7z'],
+        protocol=StdOutErrCapture)
     # reporting in variable order across platforms
     # Linux: 7-Zip [64] 16.02
     # Windows: 7-Zip 19.00 (x86)
-    pieces = out.strip().split(':', maxsplit=1)[0].strip().split()
+    pieces = out['stdout'].strip().split(':', maxsplit=1)[0].strip().split()
     for p in pieces:
         # the one with the dot is the version
         if '.' in p:
             return p
-    lgr.debug("Could not determine version of 7z from stdout. "
-              "stdout: %s, stderr: %s", out, err)
+    lgr.debug("Could not determine version of 7z from stdout. %s", out)
 
 
 class ExternalVersions(object):
--- a/datalad/support/external_versions.py
+++ b/datalad/support/external_versions.py
@@ -66,7 +66,7 @@ def _get_annex_version():
         out = _runner.run(
             ['git', 'annex', 'version'],
             protocol=StdOutErrCapture)
-        return out['stdout'].split('\n')[0].split(':')[1].strip()
+        return out['stdout'].splitlines()[0].split(':')[1].strip()
 
 
 def _get_git_version():
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -249,8 +249,16 @@ class WitlessProtocol(asyncio.SubprocessProtocol):
     proc_out = None
     proc_err = None
 
-    def __init__(self, done_future):
-        # future promise to be fulfilled when process exits
+    def __init__(self, done_future, encoding=None):
+        """
+        Parameters
+        ----------
+        done_future : asyncio.Future
+          Future promise to be fulfilled when process exits.
+        encoding : str
+          Encoding to be used for process output bytes decoding. By default,
+          the preferred system encoding is guessed.
+        """
         self.done = done_future
         # capture output in bytearrays while the process is running
         Streams = namedtuple('Streams', ['out', 'err'])
@@ -260,7 +268,7 @@ class WitlessProtocol(asyncio.SubprocessProtocol):
         )
         self.pid = None
         super().__init__()
-        self.encoding = getpreferredencoding(do_setlocale=False)
+        self.encoding = encoding or getpreferredencoding(do_setlocale=False)
 
         self._log_outputs = False
         if lgr.isEnabledFor(5):
--- a/datalad/config.py
+++ b/datalad/config.py
@@ -242,8 +242,9 @@ class ConfigManager(object):
             run_kwargs['cwd'] = dataset.path
         self._runner = GitWitlessRunner(**run_kwargs)
         try:
+            # reuse the runner we already have for the version check
             self._gitconfig_has_showorgin = \
-                LooseVersion(get_git_version()) >= '2.8.0'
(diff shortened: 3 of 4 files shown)


yarikoptic:
Do you expect any positive or negative effect on run time?  config reading is in the core of all operations, that is why I wonder.  According to benchmarks run on github actions -- no effect either way pretty much.

mih:
I have no strong expectations either way. I have two motivations: 1. It would be nice to have all code use a single runner. 2. I want to find the suspected bug reflected in https://github.com/datalad/datalad/pull/4698 by exercising this runner extremely often, on something very simple.

mih:
I have long suspected it, but never saw it that clearly. Here is running `datalad.tests.test_config.test_bare` as of https://github.com/datalad/datalad/pull/4699/commits/7ab8955f6b7a072be91303bff2e51fa06dec944f, once individually, and once as part of the entire test module. The test is altered to print some key info:

```diff
diff --git a/datalad/tests/test_config.py b/datalad/tests/test_config.py
index 7a078cef0..6b55905cc 100644
--- a/datalad/tests/test_config.py
+++ b/datalad/tests/test_config.py
@@ -520,6 +520,9 @@ def test_bare(path):
     # can we handle a bare repo?
     gr = GitRepo(path, create=True, bare=True)
     # any sensible (and also our CI) test environment(s) should have this
+    print('CFG', gr.config._store)
+    print('RUNNERCWD', gr.config._runner.cwd)
+    print('REPOPATH', gr.path)
     assert_in('user.name', gr.config)
     # not set something that wasn't there
     obscure_key = 'sec.reallyobscurename!@@.key'
```

Standalone output (test passes):

```
CFG {'user.name': 'DataLad Tester', 'user.email': 'test@example.com', 'core.repositoryformatversion': '0', 'core.filemode': 'true', 'core.bare': 'true', 'datalad.datasets.topurl': 'http://datasets-tests.datalad.org/', 'datalad.log.level': '100'}
RUNNERCWD /tmp/datalad_temp_test_barexej2qt6c
REPOPATH /tmp/datalad_temp_test_barexej2qt6c
```

Moduletest output (test fails):

```
CFG {'user.name': 'DataLad Tester', 'user.email': 'test@example.com', 'datalad.unittest.youcan': 'removeme', 'core.repositoryformatversion': '0', 'core.filemode': 'true', 'core.bare': 'true', 'datalad.datasets.topurl': 'http://datasets-tests.datalad.org/', 'datalad.log.level': '100', 'datalad.crazy.cfg': 'impossibletoguess', 'datalad.crazy.override': 'fromenv'}
RUNNERCWD /tmp/datalad_temp_test_barexf3mick9
REPOPATH /tmp/datalad_temp_test_barexf3mick9
```

what we are seeing here is the state of a ConfigManager instance that doesn't even exist in the context of that specific test. It has residuals from various other tests (e.g. `test_from_env`, `test_something`).

I believe that this has at least partly to do with the runner (`WitlessRunner` in this case) keeping a copy of the environment that was present when it was initialized. It is unclear to me how exactly configuration can leak into other config manager instances, and whether this is due to some change in this PR. However, I have seen this kind of magic change repeatedly in other contexts, and I am afraid it was already present in some form, somewhere, before this PR.

bpoldrack:
I suspect it as nothing to do with the  runner itself, but with our flyweights and `with_tempfile`. `*Repo` and `ConfigManager` should be new instances on a new path and we assume this throughout the test battery. However, `with_tempfile` does clean up after a test, which means, that the very same path is available for another tempfile creation later on. If those decorators therefore manage to create the same temporary repository location in several tests, the respective `*Repo`/`ConfigManager` instance might still exist and when it is accessed, re-evaluation would figure: Thing is still valid, since there's "still" a repo at that path.

Just a theory by now, but it seems more likely to me ATM.

mih:
Push the fix from #4703 -- works locally.

bpoldrack:
> Push the fix from #4703 -- works locally.

I'm not convinced. It's plausible that this fix mitigates the issue, of course (and that fix might be desirable either way). But why does it have an effect in case of - say `datalad.tests.test_config.test_bare`? It should be a brandnew instance in that test. The fact that it isn't is the underlying issue from my POV.

Do you want to try, whether I'm right or should I dig into it independently?

mih:
@bpoldrack Independent is better. I do not debate that there could be more issues (in all likelyhood there are), but #4703 definitely fixes *one* of them.

bpoldrack:
Ok. Tested and it looks like I'm wrong. Ran the entire test battery w/ a patch to record the paths `with_tempfile` is delivering. Slightly more than 100k different ones, not a single double usage of a such a path. Since it should be rather random this isn't exactly proving, but it seems unlikely that I was on the right track.

However, still not clear why it's the same instance, so need to keep digging.

bpoldrack:
I'm trying to reproduce the issues you pointed to in previous post about how `datalad.tests.test_config.test_bare` w/o the env fix for WitlessRunner in order to find the underlying problems I suspect to be there.
However, looking into it this:

> what we are seeing here is the state of a ConfigManager instance that doesn't even exist in the context of that specific test. It has residuals from various other tests (e.g. test_from_env, test_something).

appears to be an incorrect assessment, I think. Picking up those configs from `test_from_env`, `test_something` actually is correct behavior of `ConfigManager`. It's those tests that do the wrong thing. They don't temporarily patch the environment for their own purpose but change the environment of the running process. They should actually use some `with patch ...` instead of assigning to `os.environ`.

mih:
> appears to be an incorrect assessment, I think. Picking up those configs from `test_from_env`, `test_something` actually is correct behavior of `ConfigManager`. It's those tests that do the wrong thing. They don't temporarily patch the environment for their own purpose but change the environment of the running process. They should actually use some `with patch ...` instead of assigning to `os.environ`.

Sounds reasonable ... and orthogonal to this PR. Do you want to open a separate one?

bpoldrack:
> Sounds reasonable ... and orthogonal to this PR. Do you want to open a separate one?

#4707

datalad/support/external_versions.py
```diff
@@ -42,24 +42,31 @@ def __cmp__(self, other):
...
+            'git annex version --raw'.split(),
+            protocol=StdOutErrCapture)['stdout']
     except CommandError:
         # fall back on method that could work with older installations
-        out, err = _runner.run(['git', 'annex', 'version'])
-        return out.split('\n')[0].split(':')[1].strip()
+        out = _runner.run(
+            ['git', 'annex', 'version'],
+            protocol=StdOutErrCapture)
+        return out['stdout'].split('\n')[0].split(':')[1].strip()
```
bpoldrack, line 69: `splitlines` rather than `split('\n')`?

mih: I didnt want to make changes I wasnt sure about. It should do the same, but the original author didnt use it, and it wasnt a problem so far.

bpoldrack: Well, on windows the result of that split should end with `\r` now. Which in my estimation is a not yet unveiled issue rather than a non-existing one. ;-)
Anyways - not a blocker.

mih: Fair point!

mih:
Rebased after merge of #4835

mih:
All good on linux now! Wonderful!

Windows test failure point to the fact that `WitlessRunner` operates with system encoding by default, whereas the old Runner forces `utf8`. I will prepare a change that makes this flexible in `WitlessRunner` and configure it to be always utf8 when interacting with `git-config`

yarikoptic:
Note: benchmarks seems to suggest up to 20% penalty

mih:
> Note: benchmarks seems to suggest up to 20% penalty

I have learned to largely ignore the benchmarks. However, I also see a 5% slowdown (stddev=3%) in the the config manager test runtime.

mih:
Merged present master to incorporate recent ConfigManager optimizations.

datalad/support/external_versions.py
mih: Done

mih:
As they will likely be of interest, here are the benchmarks:

```
2020-09-01T09:49:09.7917807Z All benchmarks:
2020-09-01T09:49:09.7918442Z 
2020-09-01T09:49:09.7918827Z        before           after         ratio
2020-09-01T09:49:09.7919523Z      [7b416b40]       [72bf70ca]
2020-09-01T09:49:09.7920374Z      <bm/merge-target>       <bm/pr>   
2020-09-01T09:49:09.7921006Z        1.30±0.01s       1.32±0.02s     1.01  api.supers.time_createadd
2020-09-01T09:49:09.7921765Z        1.27±0.01s       1.28±0.01s     1.01  api.supers.time_createadd_to_dataset
2020-09-01T09:49:09.7922146Z            failed           failed      n/a  api.supers.time_diff
2020-09-01T09:49:09.7922604Z            failed           failed      n/a  api.supers.time_diff_recursive
2020-09-01T09:49:09.7923147Z        6.40±0.01s       6.43±0.02s     1.00  api.supers.time_installr
2020-09-01T09:49:09.7923701Z           223±3ms          226±3ms     1.01  api.supers.time_ls
2020-09-01T09:49:09.7924257Z        2.14±0.02s       2.17±0.01s     1.01  api.supers.time_ls_recursive
2020-09-01T09:49:09.7924831Z        2.34±0.02s       2.35±0.03s     1.00  api.supers.time_ls_recursive_long_all
2020-09-01T09:49:09.7925223Z            failed           failed      n/a  api.supers.time_remove
2020-09-01T09:49:09.7925736Z        1.05±0.02s       1.07±0.02s     1.02  api.supers.time_status
2020-09-01T09:49:09.7926305Z        1.80±0.01s       1.82±0.03s     1.01  api.supers.time_status_recursive
2020-09-01T09:49:09.7927032Z          97.2±2ms         98.7±3ms     1.02  api.supers.time_subdatasets
2020-09-01T09:49:09.7927941Z           520±3ms         537±20ms     1.03  api.supers.time_subdatasets_recursive
2020-09-01T09:49:09.7928537Z           100±2ms          100±5ms     1.00  api.supers.time_subdatasets_recursive_first
2020-09-01T09:49:09.7929126Z        4.01±0.03s       4.09±0.05s     1.02  api.supers.time_uninstall
2020-09-01T09:49:09.7929709Z           640±9ms          641±4ms     1.00  api.testds.time_create_test_dataset1
2020-09-01T09:49:09.7930447Z        3.92±0.02s       3.88±0.01s     0.99  api.testds.time_create_test_dataset2x2
2020-09-01T09:49:09.7931257Z       2.15±0.04ms       2.11±0.2ms     0.98  core.runner.time_echo
2020-09-01T09:49:09.7931913Z       2.31±0.08ms       2.39±0.2ms     1.04  core.runner.time_echo_gitrunner
2020-09-01T09:49:09.7932684Z          795±20ms          782±7ms     0.98  core.startup.time_help_np
2020-09-01T09:49:09.7933263Z           141±3ms          138±4ms     0.98  core.startup.time_import
2020-09-01T09:49:09.7933844Z           525±8ms          527±5ms     1.00  core.startup.time_import_api
2020-09-01T09:49:09.7934590Z          32.8±1ms       34.3±0.9ms     1.05  repo.gitrepo.time_get_content_info
2020-09-01T09:49:09.7935171Z        7.34±0.7ms       7.05±0.8ms     0.96  support.path.get_parent_paths.time_allsubmods_toplevel
2020-09-01T09:49:09.7935795Z        6.25±0.3ms       6.11±0.3ms     0.98  support.path.get_parent_paths.time_allsubmods_toplevel_only
2020-09-01T09:49:09.7936399Z           295±8ns         333±40ns    ~1.13  support.path.get_parent_paths.time_no_submods
2020-09-01T09:49:09.7937084Z        5.48±0.2ms       5.31±0.3ms     0.97  support.path.get_parent_paths.time_one_submod_subdir
2020-09-01T09:49:09.7937688Z        5.45±0.3ms       5.43±0.2ms     1.00  support.path.get_parent_paths.time_one_submod_toplevel
2020-09-01T09:49:09.7938281Z        31.8±0.09s        32.4±0.2s     1.02  usecases.study_forrest.time_make_studyforrest_mockup
```

Looks same to me.

bpoldrack (approved):
Can't see anything wrong with this one.

mih:
Wonderful, thx for the feedback!

Issue #4702, opened 2020-07-14
GitWitlessRunner maintains copy of (outdated) environment


mih:
In contrast to `GitRunner` the environment is adjusted on at the time of `__init__()`, and not at `run()`. This causes any later environment modification to have no effect on any command execution. And leads to substantial confusion (at least in my brain, #4699).

This can have sever implications, in particular, on long-running instances, as evident when switching the ConfigManager to use `GitWitlessRunner` in #4699

PR #4703, opened 2020-07-14
BF: Do not maintain an environment copy in GitWitlessRunner

mih:
Instead, adjust the current environment, as needed, at the time of
each `run()` invocation -- like `GitRunner()` does it.

This avoids silently ignoring changes in the process environment
after a `GitWitlessRunner` instance was created.

Fixes gh-4702

I will try this out in the context of #4699, which is an ideal testbed, due to the long-running `ConfigManager` instance.


mih:
Thx!

PR #4708, opened 2020-07-15
TST: Honor TMPDIR if DATALAD_TESTS_TEMP_DIR is not defined

mih:
Rational: We are using TMPDIR in quite a few places to direct where
tests do IO (to have pathnames with spaces, symlinks, different
filesystems, etc.). However, setting TMPDIR did not affect
paths created by our own test helpers. For example:

```
TMPDIR=/tmp/dltest DATALAD_TESTS_TEMP_KEEP=1 python -m nose -s -v datalad
[INFO   ] Keeping temp file: /tmp/datalad_temp_get_most_obscure_supported_name4v4gmbdq
[INFO   ] Keeping temp file: /tmp/datalad_temp_e3h1pw_e
...
```

This change rectifies this (and with it might reveal other issues):

```
TMPDIR=/tmp/dltest DATALAD_TESTS_TEMP_KEEP=1 python -m nose -s -v datalad
[INFO   ] Keeping temp file: /tmp/dltest/datalad_temp_get_most_obscure_supported_namej5wa186s
[INFO   ] Keeping temp file: /tmp/dltest/datalad_temp_pf8hschp
```

### DIFF
--- a/datalad/utils.py
+++ b/datalad/utils.py
@@ -1068,7 +1068,10 @@ def get_tempfile_kwargs(tkwargs=None, prefix="", wrapped=None):
             ([prefix] if prefix else []) +
             ([''] if (on_windows or not wrapped) else [wrapped.__name__]))
 
-    directory = os.environ.get('DATALAD_TESTS_TEMP_DIR')
+    directory = os.environ.get(
+        'DATALAD_TESTS_TEMP_DIR',
+        os.environ.get('TMPDIR')
+    )
     if directory and 'dir' not in tkwargs_:
         tkwargs_['dir'] = directory
 


mih:
Holy cow! Now stuff is happening when putting TMPDIR on an exfat drive. Waiting for the test repos to build is a feat on its own. Then SSH breakage, platform assumption broken...

bpoldrack:
While this is an obvious improvement, I think it actually should set `TMPDIR` as the default for that config and have that code `datalad.cfg.obtain("datalad.tests.temp.dir") instead.

mih:
> While this is an obvious improvement, I think it actually should set `TMPDIR` as the default for that config and have that code `datalad.cfg.obtain("datalad.tests.temp.dir") instead.

I am not sure what you mean by "set TMDIR as default" -- I thought I did that. Maybe you can suggest a patch instead of a verbal description.

Re config query: I will not even try that. Any code that is involved in testing the config better not use it.

mih:
What is scary is the sever breakage at the repo and utility level (see issues linked above). What is nice is the fact that most of `core` works just fine.

bpoldrack:
Merge is fine with me - I'm working on a follow-up PR, since there's more to it. This actually is the wrong function to respect DATALAD_TESTS_TEMP_DIR in the first place (which isn't about this PR, of course).

mih:
One more observation: Because we rely on `get_most_obscure_supported_name()` in various places to figure out which corner cases of filename composition we can expect to work, we are suffering from the fact that before this PR that function would use a test path that is not under TMPDIR, hence not necessarily in the same crippled FS that the tests intent to run in. Which invalidates the entire test.

bpoldrack:
> Because we rely on get_most_obscure_supported_name() in various places to figure out which corner cases of filename composition we can expect to work, we are suffering from the fact that before this PR that function would use a test path that is not under TMPDIR, hence not necessarily in the same crippled FS that the tests intent to run in. Which invalidates the entire test.

This will be fixed by my PR, since `get_most_obscure_supported_name() ` relies on `with_tempfile`, which I fix.

yarikoptic:
I am confused a bit since I expected that without `DATALAD_TESTS_TEMP_DIR` set we would just rely on default behavior of `tempfile` which would consult `TMPDIR`... and it does!  the gotcha is that if that path does not exist - it would fall back to `/tmp`:

```
$> TMPDIR=/tmp/buga python -c 'import tempfile; print(tempfile.mktemp())'
/tmp/tmpe3qz6xnq

$> mkdir /tmp/buga; TMPDIR=/tmp/buga python -c 'import tempfile; print(tempfile.mktemp())'
/tmp/buga/tmp_fei_iwm
```

yarikoptic:
having said that, I don't think there should be some negative side effect from defaulting `DATALAD_TESTS_TEMP_DIR` to `TMPDIR`, although someone could check why it was not done by default (if there was some reasoning) in the history where it was introduced.

kyleam:
@yarikoptic:
> I am confused a bit since I expected that without `DATALAD_TESTS_TEMP_DIR` set we would just rely on default behavior of `tempfile` which would consult `TMPDIR`... and it does! the gotcha is that if that path does not exist - it would fall back to `/tmp`:

Interesting.  Thanks.

> although someone could check why it was not done by default (if there was some reasoning) in the history where it was introduced.

I didn't spot anything in the related pr/commit.  (It was originally named `DATALAD_TESTS_TEMPDIR`.)

Okay, that looks like two "merge is fine by me"s.  In it goes.

Thanks @mih.

Issue #4726, opened 2020-07-15
ssh connector broken on crippled FS


mih:
Happened with gh-4708 running on an exFAT SSD.

```
======================================================================
ERROR: datalad.support.tests.test_sshconnector.test_ssh_copy
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/mih/env/datalad3-dev/lib/python3.7/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/mih/hacking/datalad/git/datalad/tests/utils.py", line 187, in newfunc
    return func(*args, **kwargs)
  File "/home/mih/hacking/datalad/git/datalad/tests/utils.py", line 262, in newfunc
    return func(*args, **kwargs)
  File "/home/mih/hacking/datalad/git/datalad/tests/utils.py", line 715, in newfunc
    return t(*(arg + (filename,)), **kw)
  File "/home/mih/hacking/datalad/git/datalad/tests/utils.py", line 715, in newfunc
    return t(*(arg + (filename,)), **kw)
  File "/home/mih/hacking/datalad/git/datalad/tests/utils.py", line 715, in newfunc
    return t(*(arg + (filename,)), **kw)
  File "/home/mih/hacking/datalad/git/datalad/support/tests/test_sshconnector.py", line 219, in test_ssh_copy
    togetfile.write_text(str('something'))
  File "/usr/lib/python3.7/pathlib.py", line 1225, in write_text
    with self.open(mode='w', encoding=encoding, errors=errors) as f:
  File "/usr/lib/python3.7/pathlib.py", line 1193, in open
    opener=self._opener)
  File "/usr/lib/python3.7/pathlib.py", line 1046, in _opener
    return self._accessor.open(self, flags, mode)
FileNotFoundError: [Errno 2] No such file or directory: '/media/mih/Samsung_T5/tmp/datalad_temp_test_ssh_copy0ertf7ow.c opy/2|g>e"t.t&x;t'

======================================================================
ERROR: datalad.support.tests.test_sshconnector.test_ssh_open_close
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/mih/env/datalad3-dev/lib/python3.7/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/mih/hacking/datalad/git/datalad/tests/utils.py", line 187, in newfunc
    return func(*args, **kwargs)
[...]

Issue #4728, opened 2020-07-15
What filenames are (il)legal?


mih:
Digging into test failures on crippled filesystems I found that we still use random collections of characters as test filenames in some places. We cannot expect them to work on windows.

Here is a helpful link: https://kb.acronis.com/content/39790

Summary: no `/ ? < > \ : * | "` and any character you can type with the Ctrl key, also no `^` on FAT and no space or `.` at the end of the filename

yarikoptic:
And we take the most obscure over which file system supports: https://github.com/datalad/datalad/blob/master/datalad/tests/utils.py#L1444

bpoldrack:
We didn't, @yarikoptic . See PR #4733.

yarikoptic:
Was may be the actual reason - absence of that `TMPDIR` directory? see
https://github.com/datalad/datalad/pull/4708#issuecomment-658795652

bpoldrack:
Not only, but that's certainly part of it. There's more we did wrong and for the particular issue here, there's hardcoded obscurity in addtion as Michael shows in #4730

mih:
I think this can go on record now.

PR #4730, opened 2020-07-15
TST+BF: SSH test assumptions invalid on windows/*FAT

mih:
Tested filename not supported by filesystem (see gh-4728)

Test also failed due to the inability to use SSH control master
based connections, but the actual functionality works.

Arguably, SSH tooling should be able to handle that, and this particular
test should not test whether some internal connection mechanism worked.

Fixes gh-4726

### DIFF
--- a/datalad/support/tests/test_sshconnector.py
+++ b/datalad/support/tests/test_sshconnector.py
@@ -74,7 +74,7 @@ def test_ssh_get_connection():
 
 @skip_if_on_windows
 @skip_ssh
-@with_tempfile(suffix=' "`suffix:;& ',  # get_most_obscure_supported_name(),
+@with_tempfile(suffix=get_most_obscure_supported_name(),
                content="1")
 def test_ssh_open_close(tfile1):
 
@@ -84,7 +84,6 @@ def test_ssh_open_close(tfile1):
                get_connection_hash('localhost', bundled=True))
     # TODO: facilitate the test when it didn't exist
     existed_before = exists(path)
-    print("%s existed: %s" % (path, existed_before))
 
     c1 = manager.get_connection('ssh://localhost')
     c1.open()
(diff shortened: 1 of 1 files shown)


yarikoptic:
Crap, I can't find it... I have a memory of seeing git annex refusing to establish socket on the file system which doesn't support them, and providing an environment (or config?) variable to set alternative path... There is also `git config annex.sshcaching false`

yarikoptic:
Oops, I think I mixed up - above comment is unrelated to this PR.

mih:
I can replicate the travis failure locally:

```
datalad.support.exceptions.CommandError: CommandError: 'scp -o ControlPath=/home/mih/.cache/datalad/sockets/7e2df38e -P 22 'localhost:/tmp/datalad_temp_test_ssh_copyq6gwyx4h.c\ opy/\ \"\'"'"'\;a\&b\&cΔЙקم๗あ\ `\|\ 2' /tmp/datalad_temp_test_ssh_copyq6gwyx4h' failed with exitcode 1 [err: 'zsh:1: unmatched `']
```

Quoting is indeed broken.

mih:
Interestingly, the command giving to the runner seems to be fine:

```
> /home/mih/hacking/datalad/git/datalad/support/sshconnector.py(394)get()
-> return self.runner.run(scp_cmd)
(Pdb) scp_cmd
['scp', '-o', 'ControlPath=/home/mih/.cache/datalad/sockets/7e2df38e', '-P', '22', 'localhost:/tmp/datalad_temp_test_ssh_copyoo2yn6jj.c\\ opy/\\ \\"\\\'\\;a\\&b\\&cΔЙקم๗あ\\ `\\|\\ 2', '/tmp/datalad_temp_test_ssh_copyoo2yn6jj']
```

Also confusion, we is `zsh` (for me) and `bash` on travis reporting the error. The runner is invoked with `shell=None` and docs say:

```
836             shell: bool, optional
837                 Run command in a shell.  If not specified, then it runs in a shell
838                 only if command is specified as a string (not a list)
```
and we pass a list.

mih:
Ah, I get it now! This is and error that comes in from the remote! At which point the command obviously was (wrongly) stringified.

mih:
So my original change managed to trip `scp` itself. So I force-pushed a change that slightly simplified the most obscure name to not do that, but still remains pretty obscure.

yarikoptic:
@mih shouldn't this to go over to `maint` so we do not trip bugfix releases testing on windows/FAT?

Issue #4723, opened 2020-07-15
test_get_most_obscure_supported_name fails on crippled FS


mih:
Happened with gh-4708 running on an exFAT SSD.

```
======================================================================
FAIL: datalad.tests.test_tests_utils.test_get_most_obscure_supported_name
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/mih/env/datalad3-dev/lib/python3.7/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/mih/hacking/datalad/git/datalad/tests/test_tests_utils.py", line 223, in test_get_most_obscure_supported_name
    eq_(n, OBSCURE_PREFIX + OBSCURE_FILENAMES[1])
AssertionError: ' ;abcΔЙקم๗あ ' != ' "\';a&b&cΔЙקم๗あ `| '
```

This issue is that the test uses the platform to determine which results there should be, but the trigger is the filesystem not the platform.

```
if platform.system() in ('Linux', 'Darwin'):
```

PR #4731, opened 2020-07-15
BF: Test inferred filesystem properties from platform name

mih:
I'd argue that the test remains about as useful as it was before.

Fixes gh-4723


bpoldrack:
FTR: Still requires #4733 to work correctly if filesystem for tests differs (via TMPDIR / DATALAD_TESTS_TEMP_DIR), since `get_most_obscure_filename` would test the wrong location otherwise.

Issue #4719, opened 2020-07-15
test_tests_utils.test_ok_symlink_helper fails on crippled FS


mih:
Happened with gh-4708 running on an exFAT SSD.

```
======================================================================
ERROR: datalad.tests.test_tests_utils.test_ok_symlink_helpers
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/mih/env/datalad3-dev/lib/python3.7/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/mih/hacking/datalad/git/datalad/tests/utils.py", line 715, in newfunc
    return t(*(arg + (filename,)), **kw)
  File "/home/mih/hacking/datalad/git/datalad/tests/test_tests_utils.py", line 265, in test_ok_symlink_helpers
    os.symlink(tmpfile, tmpfile_symlink)
OSError: [Errno 38] Function not implemented: '/media/mih/Samsung_T5/tmp/datalad_temp_test_ok_symlink_helpersmxmq279t' -> '/media/mih/Samsung_T5/tmp/datalad_temp_test_ok_symlink_helpersmxmq279t_symlink'

```

Issue #4718, opened 2020-07-15
test_tests_utils.test_ok_file_under_git_symlinks fails on crippled FS


mih:
Happened with gh-4708 running on an exFAT SSD.

```
======================================================================
ERROR: datalad.tests.test_tests_utils.test_ok_file_under_git_symlinks
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/mih/env/datalad3-dev/lib/python3.7/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/mih/hacking/datalad/git/datalad/tests/utils.py", line 187, in newfunc
    return func(*args, **kwargs)
  File "/home/mih/hacking/datalad/git/datalad/tests/utils.py", line 542, in newfunc
    return t(*(arg + (d,)), **kw)
  File "/home/mih/hacking/datalad/git/datalad/tests/test_tests_utils.py", line 663, in test_ok_file_under_git_symlinks
    os.symlink(path, lpath)
OSError: [Errno 38] Function not implemented: '/media/mih/Samsung_T5/tmp/datalad_temp_tree_test_ok_file_under_git_symlinksze9pfucy' -> '/media/mih/Samsung_T5/tmp/datalad_temp_tree_test_ok_file_under_git_symlinksze9pfucy-symlink'

```

Issue #4715, opened 2020-07-15
test_utils.test_getpwd_symlink fails on crippled FS


mih:
Happened with gh-4708 running on an exFAT SSD.

```
======================================================================
ERROR: datalad.tests.test_utils.test_getpwd_symlink
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/mih/env/datalad3-dev/lib/python3.7/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/mih/hacking/datalad/git/datalad/tests/utils.py", line 187, in newfunc
    return func(*args, **kwargs)
  File "/home/mih/hacking/datalad/git/datalad/tests/utils.py", line 715, in newfunc
    return t(*(arg + (filename,)), **kw)
  File "/home/mih/hacking/datalad/git/datalad/tests/utils.py", line 1174, in newfunc
    raise exc_info[1]
  File "/home/mih/hacking/datalad/git/datalad/tests/utils.py", line 1146, in newfunc
    ret = func(*args, **kwargs)
  File "/home/mih/hacking/datalad/git/datalad/tests/test_utils.py", line 395, in test_getpwd_symlink
    os.symlink('.', sdir)
OSError: [Errno 38] Function not implemented: '.' -> '/media/mih/Samsung_T5/tmp/datalad_temp_test_getpwd_symlink4jgv08gp/s1'

```

PR #4732, opened 2020-07-15
TST: Robustify tests that rely on symlinks

mih:
Use actual failure to create on as the indicator that it isn't possible.
Inference based on platform name, or `os.symlink` availability isn't
reliable enough.

Fixes gh-4719
Fixes gh-4718
Fixes gh-4715

### DIFF
--- a/datalad/tests/test_tests_utils.py
+++ b/datalad/tests/test_tests_utils.py
@@ -254,15 +254,15 @@ def test_keeptemp_via_env_variable():
 @with_tempfile
 def test_ok_symlink_helpers(tmpfile):
 
-    if on_windows:  # pragma: no cover
-        raise SkipTest("no sylmlinks on windows")
-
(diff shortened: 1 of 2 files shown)


bpoldrack:
What about `has_symlink_capability`?

bpoldrack:
I'd fix `has_symlink_capability` to honor `datalad.tests.temp.dir` and then use this instead. I can do that in #4733

mih:
> I'd fix `has_symlink_capability` to honor `datalad.tests.temp.dir` and then use this instead. I can do that in #4733

That should be done too. But it would not be better or more precise, IMHO. This one checks and reports exactly what is the case, when it is the case.

bpoldrack:
Better only in terms of code duplication and somewhat inconsistent approaches across tests. However, this can be merged and I'd then unify in #4733 if that helps you making progress.

PR #4733, opened 2020-07-15
Properly account for TMPDIR and datalad.tests.temp.dir

bpoldrack:
Sits on top of #4708 but could also just replace it.

There's probably more similar issues + I want to see the full test runs, since I suspect this issue also lead to not doing the right thing with several test runs (like setting TMPDIR to "some/sym link/"). Therefore: Don't merge yet.


mih:
It looks as if it is not actually sitting on top of #4708

bpoldrack:
> It looks as if it is not actually sitting on top of #4708

Technically it does. But it changes what 4708 is doing again. Just anticipating that you might want to merge that one before as it already allows to dig into other issues. So - simply avoiding a conflict.

bpoldrack:
Not sure, whether this should better be a PR against `maint`. Fixes are not "user-facing", but would affect the testing of additional fixes for `maint`.

yarikoptic:
@bpoldrack - could you please rebase it on top of `maint`? I would have preferred to see it in `maint` also to avoid possible merge conflicts etc since changes went quite broad.

mih:
I reworked this PR against `maint`. Fingers crossed.

mih:
Hmm, it seems switching the base has confused the CI machinery.

kyleam:
Okay, this has been marked as ready for review for a while and no objections were raised, and the rebase looks good to me.  Will merge.

Issue #4714, opened 2020-07-15
test_AnnexRepo_on_uninited_annex fails on crippled FS


mih:
Happened with gh-4708 running on an exFAT SSD.

```
======================================================================
FAIL: datalad.support.tests.test_annexrepo.test_AnnexRepo_on_uninited_annex
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/mih/env/datalad3-dev/lib/python3.7/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/mih/hacking/datalad/git/datalad/tests/utils.py", line 1042, in newfunc
    t(*(arg + (uri,)), **kw)
  File "/home/mih/hacking/datalad/git/datalad/tests/utils.py", line 715, in newfunc
    return t(*(arg + (filename,)), **kw)
  File "/home/mih/hacking/datalad/git/datalad/support/tests/test_annexrepo.py", line 757, in test_AnnexRepo_on_uninited_annex
    ok_(annex.file_has_content('test-annex.dat'))
AssertionError: None

```

mih:
So this seems to be breakage of `AnnexRepo.file_has_content()` which would explain a whole slew of error reports on crippled FS. Here is what I see on the same repo, right after `pdb` dropped into the test failure:

```
(Pdb) annex.path
'/media/mih/Samsung_T5/tmp/datalad_temp_test_AnnexRepo_on_uninited_annexwyp29d18'
(Pdb) annex.file_has_content('test-annex.dat')
False
```

but also

```
mih@meiner ...t_AnnexRepo_on_uninited_annexwyp29d18 (git)-[adjusted/master(unlocked)] % git annex find --not --in here
mih@meiner ...t_AnnexRepo_on_uninited_annexwyp29d18 (git)-[adjusted/master(unlocked)] % git annex find --in here                                                                    
test-annex.dat
mih@meiner ...t_AnnexRepo_on_uninited_annexwyp29d18 (git)-[adjusted/master(unlocked)] % cat test-annex.dat 
content to be annex-addurl'd%                                                                
```

which seems to be the desired content. Hence `git-annex` itself is correct in reporting the file to be available, and DataLad is wrong.

mih:
The reason is that `AnnexRepo.file_has_content()` is ultimately performing its "quick" check which does

```
filepath.exists() and (filepath.is_symlink() and opj('annex', 'objects') in os.readlink(str(filepath))
```

[...]

PR #4736, opened 2020-07-16
RF+BF+TST: Discontinue "quick" annex (availability) checks

mih:
These quick alternative checks yield invalid result on crippled FS.
Moreover, they are not used in non-test code.

Consequently, their removal should improve validity of tests without
impacting the performance of user-facing code.

Fixes gh-4714

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -1737,27 +1737,24 @@ class AnnexRepo(GitRepo, RepoInterface):
 
         return objects
 
-    def _check_files(self, fn, quick_fn, files, allow_quick, batch):
+    def _check_files(self, fn, files, batch):
         # Helper that isolates the common logic in `file_has_content` and
         # `is_under_annex`. `fn` is the annex command used to do the check, and
         # `quick_fn` is the non-annex variant.
         pointers = self.supports_unlocked_pointers
-        if pointers or batch or not allow_quick:
-            # We're only concerned about modified files in V6+ mode. In V5
-            # `find` returns an empty string for unlocked files.
-            modified = [
-                f for f in self.call_git_items_(
-                    ['diff', '--name-only', '-z'], sep='\0')
-                if f] if pointers else []
-            annex_res = fn(files, normalize_paths=False, batch=batch)
-            return [bool(annex_res.get(f) and
-                         not (pointers and normpath(f) in modified))
-                    for f in files]
-        else:  # ad-hoc check which should be faster than call into annex
-            return [quick_fn(f) for f in files]
+        # We're only concerned about modified files in V6+ mode. In V5
+        # `find` returns an empty string for unlocked files.
+        modified = [
+            f for f in self.call_git_items_(
+                ['diff', '--name-only', '-z'], sep='\0')
+            if f] if pointers else []
+        annex_res = fn(files, normalize_paths=False, batch=batch)
+        return [bool(annex_res.get(f) and
+                     not (pointers and normpath(f) in modified))
+                for f in files]
 
     @normalize_paths
-    def file_has_content(self, files, allow_quick=True, batch=False):
+    def file_has_content(self, files, allow_quick=False, batch=False):
         """Check whether files have their content present under annex.
 
         Parameters
@@ -1765,8 +1762,7 @@ class AnnexRepo(GitRepo, RepoInterface):
         files: list of str
             file(s) to check for being actually present.
         allow_quick: bool, optional
-            allow quick check, based on having a symlink into .git/annex/objects.
-            Works only in non-direct mode (TODO: thin mode)
+            This is no longer supported.
 
         Returns
         -------
@@ -1774,28 +1770,10 @@ class AnnexRepo(GitRepo, RepoInterface):
             For each input file states whether file has content locally
         """
         # TODO: Also provide option to look for key instead of path
-
-        def quick_check(filename):
-            filepath = self.pathobj / filename
-            # MIH wonders why it cannot just return exists(filepath)...
-            # this test must not use op.path.exists(), because automagic
-            # IO patches it to return True for any symlink pointing into
-            # the annex
-            return filepath.exists() and (
-                filepath.is_symlink()
-                # TODO: checks for being not outside of this repository
-                # Note: ben removed '.git/' from '.git/annex/objects',
-                # since it is not true for submodules, whose '.git' is a
-                # symlink and being resolved to some
-                # '.git/modules/.../annex/objects'
-                and opj('annex', 'objects') in os.readlink(str(filepath))  # realpath OK
-            )
-
-        return self._check_files(self.find, quick_check,
-                                 files, allow_quick, batch)
+        return self._check_files(self.find, files, batch)
 
     @normalize_paths
-    def is_under_annex(self, files, allow_quick=True, batch=False):
+    def is_under_annex(self, files, allow_quick=False, batch=False):
         """Check whether files are under annex control
 
         Parameters
@@ -1803,8 +1781,7 @@ class AnnexRepo(GitRepo, RepoInterface):
         files: list of str
             file(s) to check for being under annex
         allow_quick: bool, optional
-            allow quick check, based on having a symlink into .git/annex/objects.
-            Works only in non-direct mode (TODO: thin mode)
+            This is no longer supported.
 
         Returns
         -------
@@ -1826,20 +1803,7 @@ class AnnexRepo(GitRepo, RepoInterface):
             return self.info([f for f in files if not isdir(f)],
                              fast=True, **kwargs)
 
-        def quick_check(filename):
-            filepath = opj(self.path, filename)
-            # todo checks for being not outside of this repository
-            # Note: ben removed '.git/' from '.git/annex/objects',
-            # since it is not true for submodules, whose '.git' is a
-            # symlink and being resolved to some
-            # '.git/modules/.../annex/objects'
-            return (
-                islink(filepath)
-                and opj('annex', 'objects') in os.readlink(filepath)  # realpath OK
-            )
-
-        return self._check_files(check, quick_check,
-                                 files, allow_quick, batch)
+        return self._check_files(check, files, batch)
 
     def init_remote(self, name, options):
         """Creates a new special remote
--- a/datalad/support/tests/test_annexrepo.py
+++ b/datalad/support/tests/test_annexrepo.py
@@ -753,9 +753,8 @@ def test_AnnexRepo_on_uninited_annex(origin, path):
     annex = AnnexRepo(path, create=False, init=False)  # so we can initialize without
     # and still can get our things
     assert_false(annex.file_has_content('test-annex.dat'))
-    with swallow_outputs():
-        annex.get('test-annex.dat')
-        ok_(annex.file_has_content('test-annex.dat'))
+    annex.get('test-annex.dat')
+    ok_(annex.file_has_content('test-annex.dat'))
 
 
 @assert_cwd_unchanged


bpoldrack:
Just FTR: This not only affects crippled FS. annex can commit the locked/unlocked state of a file for a long time now. Symlink assumptions are long overdue to be removed.

kyleam:
AFAICT the entire discussion around this is confused.  `allow_quick=True` should only be in effect for v5 repos (if not, there's a bug elsewhere).  Crippled FS should be upgraded past v5 (if not, there's a bug elsewhere).  Plus, this is only relevant for a narrow range of git-annex versions: those above our minimum version (7.20190503) that do not auto-upgrade to a repository version above 5 (below 7.20190912), so any mention of failures should include what annex version is there.

Given the restricted git-annex range above, I have no objections to removing the `quick=True` handling.  It came from a time where v5 was the default.  But I can't see how the explanations so far are valid descriptions of the situation.

kyleam:
>  `allow_quick=True` should only be in effect for v5 repos (if not, there's a bug elsewhere)

It looks like the elsewhere is the assumption that `supports_unlocked_pointers` returning False is a valid indication that a repo is v5. It's not because of the `KeyError -> False` treatment in `supports_unlocked_pointers`:

https://github.com/datalad/datalad/blob/849801bb1a784e9f208329b9431ef81a20c75a62/datalad/support/annexrepo.py#L1185-L1190

There's very little point in keeping `allow_quick=True` around (our next git-annex version bump will likely take us over any git-annex version the defaults to v5), so I think we should go ahead with this PR.

mih:
@kyleam Then there are more bugs. I was debugging today and saw it using the quick checks with:

```
mih@meiner /media/mih/Samsung_T5/tmp/some % git annex version |head -n1
git-annex version: 8.20200330
mih@meiner /media/mih/Samsung_T5/tmp/some % git --version
git version 2.27.0
mih@meiner /media/mih/Samsung_T5/tmp/some % git init
Initialized empty Git repository in /media/mih/Samsung_T5/tmp/some/.git/
mih@meiner /media/mih/Samsung_T5/tmp/some (git)-[master] % git annex init
init  
  Detected a filesystem without fifo support.

  Disabling ssh connection caching.

  Detected a crippled filesystem.
(scanning for unlocked files...)

  Entering an adjusted branch where files are unlocked as this filesystem does not support locked files.

Switched to branch 'adjusted/master(unlocked)'
ok
(recording state in git...)
mih@meiner /media/mih/Samsung_T5/tmp/some (git)-[adjusted/master(unlocked)] % git annex version  | tail -n3
supported repository versions: 8
upgrade supported from repository versions: 0 1 2 3 4 5 6 7
local repository version: 8
```

bpoldrack:
FTR: Even for v5 it wasn't exactly valid, since there was direct mode at the time.

Re auto-upgrade: FWIW we still have the `datalad.repo.version` config, that would cause `annex-init` to explicitly called with the version. I think this prevents auto-upgrade and for testing purposes it's a good thing to have in my estimation.

kyleam:
> Re auto-upgrade: FWIW we still have the `datalad.repo.version` config, that would cause `annex-init` to explicitly called with the version. I think this prevents auto-upgrade and for testing purposes it's a good thing to have in my estimation.

No, it doesn't prevent auto-upgrade.

kyleam:
@mih

> @kyleam Then there are more bugs. I was debugging today and saw it using the quick checks with:

My point is that the discussion around this is completely missing that, if the repo support unlocked files (i.e. is v6 or above), then quick should never be used.  In the [comment][0]  following my initial one, I described where the faulty logic is that let's the quick path be taken even though it shouldn't on v6+.

[0]: https://github.com/datalad/datalad/pull/4736#issuecomment-659454894

kyleam:
@bpoldrack 
> FTR: Even for v5 it wasn't exactly valid, since there was direct mode at the time.

I haven't dug back to when we supported direct mode to see what the handling was like back then, but perhaps we didn't account for it.  Or perhaps we did and pruned it or otherwise reworked things with the removal of direct mode support.  Either way, at the moment, we of course have a code base that doesn't support direct mode but does support a git-annex version that defaults to v5.

yarikoptic:
had only a brief look: I expect such a change could have tremendous performance hit on a number scenarios, including run time of the tests.
Anyone timed it?

> ... Moreover, they are not used in non-test code.

default was to allow quick checks, so any of the uses below could get effected, in particular I expect major impact on `auto` (either it is popular/used - is a separate question):
```shell
$> git grep -A2 'file_has_content' | grep -v test_| grep '\.py[-:]'
datalad/auto.py:        if (under_annex or under_annex is None) and not annex.file_has_content(filepath):
datalad/auto.py-            lgr.info("AutomagicIO: retrieving file content of %s", filepath)
datalad/auto.py-            out = annex.get(filepath)
datalad/interface/add_archive_content.py:                # if annex.file_has_content(target_path):
datalad/interface/add_archive_content.py-                #     # if not --  it was added to git, if in annex, it is present and output is True
datalad/interface/add_archive_content.py-                #     annex.add_url_to_file(target_file, url, options=['--relaxed'], batch=True)
datalad/interface/ls.py:                    if self.repo.file_has_content(str(self._path)) \
datalad/interface/ls.py-                    else 0
datalad/interface/ls.py-            # else ask fs for node size (= ondisk_size)
datalad/metadata/aggregate.py:            if not repo.file_has_content(relpath):
datalad/metadata/aggregate.py-                present = False
datalad/metadata/aggregate.py-                backends.append(repo.get_key_backend(key))
datalad/metadata/metadata.py:        content_info = zip(paths, ds.repo.file_has_content(paths), ds.repo.is_under_annex(paths))
datalad/metadata/metadata.py-        paths = [p for p, c, a in content_info if not a or c]
datalad/metadata/metadata.py-        nocontent = len(fullpathlist) - len(paths)
datalad/plugin/export_archive.py:                has_content = repo.file_has_content(
datalad/plugin/export_archive.py-                    repo_files, allow_quick=True, batch=True)
datalad/plugin/export_archive.py-            else:
... removed its own definition
datalad/tests/utils.py:    has_content = ar.file_has_content(files)
datalad/tests/utils.py-    if isinstance(has_content, bool):
datalad/tests/utils.py-        ok_(has_content)
datalad/tests/utils.py:def ok_file_has_content(path, content, strip=False, re_=False,
datalad/tests/utils.py-                        decompress=False, **kwargs):
datalad/tests/utils.py-    """Verify that file exists and has expected content"""
```

I wonder if it shouldn't be just fixed (if not a symlink -- do "proper" check, if a symlink to `.git/annex/objects/` -- leave at current "quick" logic) ?

kyleam:
> had only a brief look: I expect such a change could have tremendous performance hit on a number scenarios, including run time of the tests.
> Anyone timed it?

While timing would be fine to see, any performance benefits would be restricted to git annex versions below 7.20190912.  Relevant for neurodebian and the conda we test with at the moment, but we can't stay stuck behind that forever, so it's just delaying the inevitable.

> I wonder if it shouldn't be just fixed

An alternative fix to the failure would be something like this (would need doc/comment adjustment):

```diff
diff --git a/datalad/support/annexrepo.py b/datalad/support/annexrepo.py
index c740500b9..31a292697 100644
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -1187,7 +1187,7 @@ def supports_unlocked_pointers(self):
         except KeyError:
             # If annex.version isn't set (e.g., an uninitialized repo), assume
             # that unlocked pointers aren't supported.
-            return False
+            return None
 
     # TODO: RM DIRECT  might be gone but pieces might be useful for establishing
     #       migration to v6+ mode and testing. For now is made protected to
@@ -1742,7 +1742,7 @@ def _check_files(self, fn, quick_fn, files, allow_quick, batch):
         # `is_under_annex`. `fn` is the annex command used to do the check, and
         # `quick_fn` is the non-annex variant.
         pointers = self.supports_unlocked_pointers
-        if pointers or batch or not allow_quick:
+        if pointers is not False or batch or not allow_quick:
             # We're only concerned about modified files in V6+ mode. In V5
             # `find` returns an empty string for unlocked files.
             modified = [
```

yarikoptic:
> ... any performance benefits would be restricted to git annex versions below 7.20190912.

sorry, could you elaborate on this?

kyleam:
> > ... any performance benefits would be restricted to git annex versions below 7.20190912.
> 
> sorry, could you elaborate on this?

Starting with that git-annex version, v5 repos are auto-upgraded to v7, in which case `supports_unlocked_pointers` is True (except when inside uninitialized annex repos, which is the logic bug responsible for the linked `test_AnnexRepo_on_uninited_annex` failure).  If `supports_unlocked_pointers` is True, the non-quick code path is taken regardless of whether `allow_quick` is True.

yarikoptic:
THANK YOU! 
`supports_unlocked_pointers` was introduced in 209897c7b21daf7c60ca180e35d8dc43c6b11418 of #2975.. I do not see any discussion of `pointers` unconditionally (supporting pointers does not make all symlinks go away) "masking away" `allow_quick` in the commit message or PR. So may be that PR already has contributed to some slow downs when using those recent git-annex versions, and may be we could regain some speed up if `allow_quick` was reintroduced (with necessary fixes) even in cases of annex with support for pointers, since even in v7 we still have symlinks in most typical IMHO use cases.

kyleam:
> `supports_unlocked_pointers` was introduced in [209897c](https://github.com/datalad/datalad/commit/209897c7b21daf7c60ca180e35d8dc43c6b11418) of #2975.. I do not see any discussion of `pointers` unconditionally (supporting pointers does not make all symlinks go away) "masking away" `allow_quick` in the commit message or PR.

The commit didn't change the behavior.  It is just introducing the property.  The behavior changed in 5dc2eb495 (BF: annexrepo.file_has_content: Fix check for unlocked files in V6, 2018-09-19).  It does discuss sending everything through the slower check:

```
Note: If we're in V6, we send everything through the slower annex
check, but we could still check locked files with the quick check,
which would probably improve performance.  Punt on this for now
though.
```

That aligns with what you propose next...

> So may be that PR already has contributed to some slow downs when using those recent git-annex versions, and may be we could regain some speed up if `allow_quick` was reintroduced (as patch suggested above + with necessary fixes detected on crippled FSs) even in cases of annex with support for pointers, since even in v7 we still have symlinks in most typical IMHO use cases.

... though if we want to go with the patch above, I don't see any reason to couple it to yet-to-materialize performance enhancements.

yarikoptic:
> The commit you point to didn't change the behavior. It is just introducing the property.

RRIGHT! I have missed that it just replaces already present `is_v6`. thanks!

bpoldrack:
> No, it doesn't prevent auto-upgrade.

Well, that seems to not be true for v5. We explicitly test it: https://github.com/datalad/datalad/blob/master/datalad/support/tests/test_annexrepo.py#L1274

kyleam:
@bpoldrack

> > No, it doesn't prevent auto-upgrade.
> 
> Well, that seems to not be true for v5. We explicitly test it: https://github.com/datalad/datalad/blob/master/datalad/support/tests/test_annexrepo.py#L1274

Auto-upgrade is about moving from an unsupported version to a supported one.  Note the condition on the test you point to:

https://github.com/datalad/datalad/blob/493add92fc213aa6eda848f957a281ce6c871f00/datalad/support/tests/test_annexrepo.py#L1273-L1278

And, as I mentioned above, v5 is no longer the default as of 7.20190912.  In that version, 5 is no longer in the supported version list:

```
git-annex version: 7.20190912-gab739242a
build flags: Assistant Webapp Pairing S3 WebDAV Inotify TorrentParser Feeds Testsuite
dependency versions: aws-0.21.1 bloomfilter-2.0.1.0 cryptonite-0.25 DAV-1.3.3 feed-1.0.1.0 ghc-8.6.5 http-client-0.5.14 persistent-sqlite-2.9.3 torrent-10000.1.1 uuid-1.3.13 yesod-1.6.0
key/value backends: SHA256E SHA256 SHA512E SHA512 SHA224E SHA224 SHA384E SHA384 SHA3_256E SHA3_256 SHA3_512E SHA3_512 SHA3_224E SHA3_224 SHA3_384E SHA3_384 SKEIN256E SKEIN256 SKEIN512E SKEIN512 BLAKE2B256E BLAKE2B256 BLAKE2B512E BLAKE2B512 BLAKE2B160E BLAKE2B160 BLAKE2B224E BLAKE2B224 BLAKE2B384E BLAKE2B384 BLAKE2BP512E BLAKE2BP512 BLAKE2S256E BLAKE2S256 BLAKE2S160E BLAKE2S160 BLAKE2S224E BLAKE2S224 BLAKE2SP256E BLAKE2SP256 BLAKE2SP224E BLAKE2SP224 SHA1E SHA1 MD5E MD5 WORM URL
remote types: git gcrypt p2p S3 bup directory rsync web bittorrent webdav adb tahoe glacier ddar git-lfs hook external
operating system: linux x86_64
supported repository versions: 7
upgrade supported from repository versions: 0 1 2 3 4 5 6
```

And here is what passing `--version=5` to git-annex-init does:

```
$ git init
Initialized empty Git repository in /tmp/gx-ALhWOUr/.git/
$ git annex init --version=5
init  (scanning for unlocked files...)
ok
(recording state in git...)
$ git config annex.version
6
```

Yes, it's odd/wrong that it landed on v6 when v7 is the only supported on.  That's due to an [annex bug][0] that was fixed a while ago.

[0]: https://git-annex.branchable.com/bugs/__39__init_--version__61__N__39___can_land_on_unsupported_N+1/

bpoldrack:
>  In that version, 5 is no longer in the supported version list:

Uh. I missed that. Thx!

mih:
If I am picking up the vibe correctly, this can be merged.

PR #4755, opened 2020-07-21
TEMP: try with new annex merged into conda-forge 1h ago

yarikoptic:


kyleam:
`conda install` isn't finding the new version (probably isn't available yet): https://travis-ci.org/github/datalad/datalad/jobs/710484358#L916

It looks like install-annex.sh needs to detect if the `conda install` fails.

kyleam:
Tried locally, and the new version is there now.  Will restart the Travis build.

yarikoptic:
Something really wrong with this conda build, yet to look in details

```
datalad.support.exceptions.CommandError: CommandError: 'git fetch --verbose --progress --prune --recurse-submodules=no target1' failed with exitcode 128 under /tmp/datalad_temp_test_gh1763rp4bl6a9 [err: 'Bad owner or permissions on /home/travis/.ssh/config
ConnectionOpenFailedError: 'ssh -fN -o ControlMaster=auto -o ControlPersist=15m -o ControlPath=/tmp/datalad_temp_op36k3iu/.cache/datalad/sockets/e9cb6b28 datalad-test' failed with exitcode 255 [Failed to open SSH connection (could not start ControlMaster process)]
and the repository exists.']
```
attn @notestaff

notestaff:
From the error messages, sounds like this could be due to transient errors -- try restarting?  ("ssh: Could not resolve hostname localhost: Name or service not known")
Also try re-rendering with the latest conda-smithy

notestaff:
P.S.  To see if the error is due to the latest git-annex build  on conda-forge, try requiring the previous version in meta.yaml

yarikoptic:
I have restarted the Travis build but it is unlikely it was something transient with static localhost.

yarikoptic:
As for previous version - will be next step, it default ATM is to test with 7.20190819

yarikoptic:
So resorting to previous didn't help really... We need to retry with current in conda

yarikoptic:
too old, needs a fresh start some time

PR #4756, opened 2020-07-21
TEMP: try with new annex 8.20200720.1 from neurodebian -devel

yarikoptic:


kyleam:
So, the good news is that there were no obvious SSH stalls.  But there are failures:

  * The 3.7 full run exceeds the max time limit of 50 minutes.

    https://travis-ci.org/github/datalad/datalad/jobs/710485444

  * The nfsmount job appears to have stalled, though that shouldn't be SSH related because it runs with `DATALAD_TESTS_SSH=1`.

    https://travis-ci.org/github/datalad/datalad/jobs/710485452

  * The "/var/tmp/sym link/" run had two failures.  I can trigger both locally and will look into them more.

    https://travis-ci.org/github/datalad/datalad/jobs/710485450#L7806

yarikoptic:
re the `sym link` (also has space), I wonder if it could relate to https://git-annex.branchable.com/bugs/cannot___40__or_how__63____41___to_pass_socket_path_with_a_space_in_its_path_via_annex-ssh-options/ ?

kyleam:
> re the `sym link` (also has space), I wonder if it could relate to

No, it doesn't.  I'll open a PR tomorrow.

yarikoptic:
I think it served its purpose, we will try a new one soonish

PR #4769, opened 2020-07-25
NF: create GitHub sibling as private repo

nicholsn:
### Overview
This PR adds a new `--private` flag to the `datalad create-sibling-github` command that result in a GitHub repo being created that is only viewable by members that have been given access.

### How it was tested
Created both public and private repositories successfully using the following

`datalad create-sibling-github --github-organization $TEST --access-protocol ssh --name $TEST --private  $TEST`

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -92,6 +92,11 @@
       "affiliation": "Psychoinformatics Lab, INM-7, Research Centre Juelich",
       "name": "Wagner, Adina S.",
       "orcid": "0000-0003-2917-3450"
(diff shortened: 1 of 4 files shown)


nicholsn:
@mih thanks! Updated the test to pass, but now coverage is "failing"... what do you suggest?

mih:
Thanks! IMHO the loss of coverage isn't too terrible in this case.

kyleam:
Looks good to me too.  Thanks @nicholsn.

Issue #4785, opened 2020-07-30
Switch dataset ID to a fully random identifier


mih:
ATM we are using `uuid1` (unless a specific random seed is present in the environment). While this is a random identifier, it also contains encoded information. In the case of `uuid1` this is the timestamp of creation and a machine MAC-address (which could be random on some machines).

Given that any identifying information can become problematic in some context, I'd propose we switch from `uuid1` to `uuid4` which is a fully random identifier.

https://en.wikipedia.org/wiki/Universally_unique_identifier#Versions

yarikoptic:
```shell
$> rm -rf testds; DATALAD_SEED=1 datalad create testds; grep id testds/.datalad/config testds/.git/config
[INFO   ] Creating a new annex repo at /tmp/testds 
create(ok): /tmp/testds (dataset)                                                                                                                                                                                                  
testds/.datalad/config:	id = ae2ab026-0b53-7e7c-7a65-f0407a6e75f5
testds/.git/config:	uuid = 5be87e09-8cb6-4b90-9e7e-25437df228d7

$> rm -rf testds; DATALAD_SEED=1 datalad create testds; grep id testds/.datalad/config testds/.git/config
[INFO   ] Creating a new annex repo at /tmp/testds 
create(ok): /tmp/testds (dataset)                                                                                                                                                                                                  
testds/.datalad/config:	id = ae2ab026-0b53-7e7c-7a65-f0407a6e75f5
testds/.git/config:	uuid = a2122058-e663-49ce-a077-2fb72373e8a7
```
and then on smaug
```
smaug% rm -rf testds; DATALAD_SEED=1 datalad create testds; grep id testds/.datalad/config testds/.git/config

[INFO   ] Creating a new annex repo at /tmp/testds 
create(ok): /tmp/testds (dataset)
testds/.datalad/config:	id = ae2ab026-0b53-7e7c-7a65-f0407a6e75f5
testds/.git/config:	uuid = 463d59b6-86e4-4e74-a6d4-03e51b99ef91
```
so dataset uuid seems to be entirely RNG driven (seeding it makes it reproducible). BUT annex instance uuid is not - and might be driven by mac+time (didn't look)

yarikoptic:
damn -- I should have read your message better -- removed my comment (that it is reproducible if RNG is seeded).

Issue #4773, opened 2020-07-27
test_AnnexRepo_commit failures on Travis


kyleam:
`test_AnnexRepo_commit` failed in two recent PRs (gh-4762 and gh-4768).  I think it's intermittent.

- https://travis-ci.org/github/datalad/datalad/jobs/712219198#L1996
- https://travis-ci.org/github/datalad/datalad/jobs/712340298#L1991

```
======================================================================
ERROR: datalad.support.tests.test_annexrepo.test_AnnexRepo_commit
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/datalad/tests/utils.py", line 1185, in newfunc
    raise exc_info[1]
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/datalad/tests/utils.py", line 1157, in newfunc
    ret = func(*args, **kwargs)
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/datalad/tests/utils.py", line 728, in newfunc
    return t(*(arg + (filename,)), **kw)
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/datalad/support/tests/test_annexrepo.py", line 786, in test_AnnexRepo_commit
    assert_raises(FileNotInRepositoryError, ds.commit, files="not-existing")
  File "/opt/python/3.7.1/lib/python3.7/unittest/case.py", line 743, in assertRaises
    return context.handle('assertRaises', args, kwargs)
  File "/opt/python/3.7.1/lib/python3.7/unittest/case.py", line 178, in handle
    callable_obj(*args, **kwargs)
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/datalad/support/gitrepo.py", line 1654, in commit
    stdin=None,
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/datalad/cmd.py", line 504, in run
    **results,
datalad.support.exceptions.CommandError: CommandError: 'git commit -m '[DATALAD] Recorded changes' -- not-existing' failed with exitcode 1 under /tmp/datalad_temp_test_AnnexRepo_commitmkv02v44
```

kyleam:
[...]

PR #4790, opened 2020-07-31
Use a fully random UUID as dataset ID

mih:
Until now the dataset ID encoded the time of dataset creation, and  a machine-identifying MAC-address (which may or may not be a reliable/persistent identifier).

DataLad neither used, nor advertised this encoded information in the ID. Moreover, having such information unconditionally encoded can counteract other features (e.g. "fake dates").

I'd argue that if identifying information of this kind is desired, it should be placed explicitly into a dataset. If UUIDs with encoded content are desired, the UUID of a dataset can be altered (right after creation) to match such needs.

Fixes gh-4785

This issue is related to https://github.com/datalad/datalad-metalad/issues/30 and impacts the suitability of an identifier for use in anonymous metadata on personal data.

This PR goes against `maint`, because it should have no functional impact, but there is value in getting this change out sooner.

### DIFF
--- a/datalad/core/local/create.py
+++ b/datalad/core/local/create.py
@@ -460,7 +460,7 @@ class Create(Interface):
         if _seed is None:
             # just the standard way
             # use a fully random identifier (i.e. UUID version 4)
-            uuid_id = uuid.uuid4().urn.split(':')[-1]
+            uuid_id = str(uuid.uuid4())
         else:
             # Let's generate preseeded ones
             uuid_id = str(uuid.UUID(int=random.getrandbits(128)))
--- a/datalad/core/local/create.py
+++ b/datalad/core/local/create.py
@@ -459,7 +459,8 @@ class Create(Interface):
 
         if _seed is None:
             # just the standard way
-            uuid_id = uuid.uuid1().urn.split(':')[-1]
+            # use a fully random identifier (i.e. UUID version 4)
+            uuid_id = uuid.uuid4().urn.split(':')[-1]
         else:
             # Let's generate preseeded ones
             uuid_id = str(uuid.UUID(int=random.getrandbits(128)))


datalad/core/local/create.py
```diff
@@ -459,7 +459,8 @@ def __call__(
 
         if _seed is None:
             # just the standard way
-            uuid_id = uuid.uuid1().urn.split(':')[-1]
+            # use a fully random identifier (i.e. UUID version 4)
+            uuid_id = uuid.uuid4().urn.split(':')[-1]
```
vsoch, line 463: Just curious, how is this different from `str(uuid.uuid4())`?

mih: The outcome looks identical. But I was struck by the sophistication of the code and felt attached. Now that you made me look:

```
3b8431e304 datalad/distribution/dataset.py (Michael Hanke      2016-09-01 15:38:47 +0200 154)                 self._id = uuid.uuid1().urn.split(':')[-1]
```

That was already in Datalad 0.3 ;-)

Pushed a change. thx!

yarikoptic:
Not of any specific advantage (at least I don't see it besides being able to say that some random uuId is a uuid of a datalad dataset), and probably would limit the space a bit, but I recalled that for datalad special remotes I have used uuid5 with our datalad namespace:

https://github.com/datalad/datalad/blob/master/datalad/customremotes/base.py#L663

mih:
Test failure is #4773

mih:
> Not of any specific advantage (at least I don't see it besides being able to say that some random uuId is a uuid of a datalad dataset), and probably would limit the space a bit, but I recalled that for datalad special remotes I have used uuid5 with our datalad namespace:
> 
> https://github.com/datalad/datalad/blob/master/datalad/customremotes/base.py#L663

I don't see how a UUID5 would fit our needs. AFAICS there is no "internal" randomness, hence we would need to generate random names to get any randomness in the UUID.

```
In [1]: uuid.uuid5(uuid.NAMESPACE_URL, 'http://datalad.org')
Out[1]: UUID('1a50c8dc-30dc-5327-8e59-450be59a00b7')

In [2]: uuid.uuid5(uuid.NAMESPACE_URL, 'http://datalad.org')
Out[2]: UUID('1a50c8dc-30dc-5327-8e59-450be59a00b7')
```

If the names have to be random, it doesn't matter if there is a constant part of it (eg. 'http://datalad.org'). Also it is based on the SHA1 of the name, hence decoding that part will not be possible

```
In [3]: uuid.uuid5(uuid.NAMESPACE_URL, 'http://datalad.org/a')
Out[3]: UUID('a14c0cef-27b1-50e0-9044-c8d0fe90650f')

In [4]: uuid.uuid5(uuid.NAMESPACE_URL, 'http://datalad.org/b')
Out[4]: UUID('9e0672b3-732b-5763-85e4-68462cb26385')
```

I think we can stick with `uuid4()`

datalad/core/local/create.py
```diff
@@ -459,7 +459,8 @@ def __call__(
 
         if _seed is None:
             # just the standard way
-            uuid_id = uuid.uuid1().urn.split(':')[-1]
+            # use a fully random identifier (i.e. UUID version 4)
+            uuid_id = str(uuid.uuid4())
         else:
             # Let's generate preseeded ones
             uuid_id = str(uuid.UUID(int=random.getrandbits(128)))
```
yarikoptic, line 466: why not to just always use this one (so no `if _seed` conditioning) then so we have a single implementation instead of conditioning here on `_seed`:
```
def uuid4():
    """Generate a random UUID."""
    return UUID(bytes=os.urandom(16), version=4)
```
so the only difference would be from stock uuid4 is that we would use  `random.getrandbits` and not  `os.urandom` ("Return a bytes object containing random bytes suitable for cryptographic use.").  Not sure if "cryptographic" aspect would really matter here...?

yarikoptic: may be just add an explicit `version=4` as well

mih: This is part of an effort towards a certified identifier setup. Being able to say "this is a UUID4" is better than "this should not be too different from a UUID4".

That being said, I am also OK with leaving things as they are and adding an explicit, separate mode where the properties of all identifiers are precisely defined.

yarikoptic: nah, with the above I was just trying to reduce number of "modes" of operation in the code, not to increase it... I am ok with this PR as is then.

mih:
Thx @yarikoptic

Issue #2880, opened 2018-09-27
datalad save could use parallelism more


notestaff:
I have a repo with many submodules.   datalad save takes a while.  It'd be faster if saving of submodules under independent dirs was done in parallel.

notestaff:
Also nested, though not that deeply (2-3 levels?)   Basically I'm trying to consolidate all my projects in one place.

So, I understand that if I have datasets A, A/B, A/B/C, A/D, A/D/E, then C has to be saved before B and E before D; but the A/B/* and A/D/* branches can be done in parallel.

notestaff:
" None of the code though is written with thread safety in mind AFAIK" -- at least for read-only operations like diff, this shouldn't be a concern?
git-annex itself is said to be thread-safe; on the other hand, this isn't currently tested ( https://git-annex.branchable.com/git-annex-test/#comment-7315b575dc4c77ac5abf7186423584ec )...
I tend to work on machines with at least 8 cores, often more, so parallelizing would certainly help.

mih:
This is already on my list. ATM I am working one level deeper, once this is done, I'll come up for this (ping #2428)

notestaff:
@mih thanks a lot!

mih:
Q on parallelization within datalad vs in git-annex:

Many relevant annex command can take a `--jobs` argument. In many cases we do not expose this in datalad ATM. If we are adding parallelization in datalad no, we have to think about avoiding oversubscription of resources.

A standard paradigm would be to use a pool of "workers", where the number of workers is limited by hardware specs and/or config. However, if we manage this pool in datalad, we would likely have no other choice but to limit git-annex to `--jobs 1`.

I wonder what is more advantageous: keep datalad single-threaded and (more opportunistically) use git-annexes features, or vice versa? Leaving it to git-annex certainly seems simpler. Any intuitions on how these two might compare re speed-up potential?

bpoldrack:
[...]

Issue #3538, opened 2019-07-17
Undesired submodule url


mih:
Originally at https://github.com/datalad/datalad/pull/2601

`datalad install -d . -s ../raw sourcedata --reckless`

leads to

```diff
--- /dev/null
+++ b/.gitmodules
@@ -0,0 +1,3 @@
+[submodule "sourcedata"]
+       path = sourcedata
+       url = /home/data/psyinf/scratch/multires3t/bids/../raw
```

kyleam:
Reverted in 0.12.1, but a new approach has been proposed in gh-4026.

PR #4793, opened 2020-08-02
OPT: Bypass `submodule add` when adding subdatasets

mih:
Instead go manual, avoiding various checks, because we already
know inside `save` that a particular directory is considered untracked
by Git.

This approach is substantially faster. Already with 50 subdatasets
the speed-up is 2x.

Quick'n'dirty benchmark:

```
% rm -rf * .git* .datalad; datalad create . && \
  ( for i in $(seq 50); do datalad create $i; done ) && \
  time datalad save && datalad status
```

Current master

```
datalad save  3.70s user 1.50s system 103% cpu 4.998 total
```

This PR

```
datalad save  1.68s user 1.02s system 108% cpu 2.499 total
```

It may well be that it needs more safety checks that would eat up that
margin. But in the context of
https://github.com/datalad/datalad/issues/2880#issuecomment-667041918
(save thousands of new subdatasets) even much smaller benefits would
still translate to substantial absolute speed-ups.

Thx again to @kyleam for digging out that `update-index` trick.

TODO:

- [x] consolidate config and `.gitmodules` manipulation into a single action for all added submodules at once
- [x] RF helpers into documented standalone functions with tests

### DIFF
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -4006,7 +4006,8 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
                           props.get('type', None) == 'directory']
         to_add_submodules = []
         if untracked_dirs:
-            to_add_submodules = [sm for sm, sm_props in
+            to_add_submodules = [
+                sm for sm, sm_props in
                 self.get_content_info(
                     untracked_dirs,
                     ref=None,
@@ -4015,30 +4016,11 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
                     untracked='all').items()
                 if sm_props.get('type', None) == 'directory']
             to_add_submodules = _prune_deeper_repos(to_add_submodules)
-            for cand_sm in to_add_submodules:
-                try:
-                    self._save_add_submodule(cand_sm)
-                except (CommandError, InvalidGitRepositoryError) as e:
-                    yield get_status_dict(
-                        action='add_submodule',
-                        ds=self,
-                        path=self.pathobj / ut.PurePosixPath(cand_sm),
-                        status='error',
-                        message=e.stderr if hasattr(e, 'stderr')
-                        else ('not a Git repository: %s', exc_str(e)),
-                        logger=lgr)
-                    continue
-                # This mirrors the result structure yielded for
-                # to_stage_submodules below.
-                yield get_status_dict(
-                    action='add',
-                    refds=self.pathobj,
-                    type='file',
-                    key=None,
-                    path=self.pathobj / ut.PurePosixPath(cand_sm),
-                    status='ok',
-                    logger=lgr)
-                added_submodule = True
+            if to_add_submodules:
+                for r in self._save_add_submodules(to_add_submodules):
+                    if r.get('status', None) == 'ok':
+                        added_submodule = True
+                    yield r
         if not need_partial_commit:
             # without a partial commit an AnnexRepo would ignore any submodule
             # path in its add helper, hence `git add` them explicitly
@@ -4060,6 +4042,8 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
                     yield r
 
         if added_submodule or vanished_subds:
+            # the config has changed too
+            self.config.reload()
             # need to include .gitmodules in what needs saving
             status[self.pathobj.joinpath('.gitmodules')] = dict(
                 type='file', state='modified')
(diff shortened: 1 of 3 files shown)


mih:
This works now. The only test failure is:

```
1649ERROR: datalad.distribution.tests.test_install.test_install_recursive
1650----------------------------------------------------------------------
1651Traceback (most recent call last):
1652  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/nose/case.py", line 134, in run
1653    self.runTest(result)
1654  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/nose/case.py", line 152, in runTest
1655    test(result)
1656AssertionError: Test was too slow (took 31.0307s, threshold was 30.0000s)
```

it is unclear to me how it can get slower (I have not yet observed a scenario where things got slower), but there are some optimization that can still be done.

mih:
Update: Consolidation on applying all `.gitmodules` and `.git/config` changes at once speeds things up by another 40%, leading to a 3x overall speed-up for 50 subdatasets.

datalad/support/gitrepo.py
```diff
@@ -4076,6 +4056,102 @@ def _save_add(self, files, git_opts=None):
...
+                    'update-index', '--add', '--replace', '--cacheinfo', '160000',
+                    i['commit'], i['rpath']
+                ])
+                gmprops = dict(path=i['rpath'], url=i['url'])
+                if i['id']:
+                    gmprops['datalad-id'] = i['id']
+                write_submodule_config(
+                    gmf, i['rpath'], gmprops)
+                write_submodule_config(
+                    gcf, i['rpath'], dict(active='true', url=i['url']))
```
kyleam, line 4140: With this switch to writing the `.git/config` entries on our own, there's a change in how the `url` is represented: before it was an absolute path and now it is relative.  I think that should be fine and for our purposes probably preferable.

datalad/support/gitrepo.py
```diff
@@ -4076,6 +4056,102 @@ def _save_add(self, files, git_opts=None):
...
+
+                # This mirrors the result structure yielded for
+                # to_stage_submodules below.
+                yield get_status_dict(
+                    action='add',
+                    refds=self.pathobj,
+                    # MIH: really? why not 'dataset', it seems "add" of an existing
+                    # subdataset (i.e. update) does that, and this one mimiks it
+                    # still, two wrongs don't make a right.
```
kyleam, line 4149: How to handle this was brought up explicitly at the time.

https://github.com/datalad/datalad/pull/3398#discussion_r282928154

Ideally I think we'd switch to using `type=dataset` for adding both modified and new datasets.  The main obstacle I see is that the result values are exposed via result hooks, so we need to worry about breakage.

In my view this comment doesn't add any value, though.

mih: I will replace this comment with a link to this statement

datalad/support/gitrepo.py
```diff
@@ -4076,6 +4056,102 @@ def _save_add(self, files, git_opts=None):
...
+            v = v.replace('\\', '\\\\')
+            if section:
+                # must not have additional unquoted quotes
+                v = v.replace('"', '\\"')
+            elif v and (v[0] in white or v[-1] in white):
+                # quoting the value due to leading/trailing whitespace
+                # prevent inner unquotes quotes
+                v = v.replace('"', '\\"')
+                v = '"{}"'.format(v)
+            return v
```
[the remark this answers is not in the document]
mih: Thx! Will change accordingly.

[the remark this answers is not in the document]
mih: This is done now.

datalad/support/gitrepo.py
mih: True. I had the same feeling of "should be better". Related: https://github.com/datalad/datalad/pull/4025#issue-362968783

mih:
Pushed update, rebased on current master

datalad/support/gitrepo.py
mih: In any case the issue from #3538 remains here, as something earlier makes the path absolute:

```
% dl create raw
% dl create bids
% cd bids
bids % datalad install -d . -s ../raw sourcedata --reckless
bids % cat .gitmodules
[submodule "sourcedata"]
        path = sourcedata
        url = /tmp/d1/bids/../raw
        datalad-id = ca96f420-2f90-4f9c-8c46-0e6213dcfd99
```

mih:
Related: #4853

bpoldrack (commented):
LGTM

mih:
Thx all!

Issue #4802, opened 2020-08-05
`create -d . --cfg-proc` is not looking for procedure in reference dataset


mih:
```
datalad -l debug create -c xnat_dataset -d . sub3
```

`.` is `/tmp/rp`

I expected it to look for `cfg_xnat_dataset` in `/tmp/rp/.datalad/procedures`, but it is not happening. Instead:

````
[DEBUG  ] Looking for procedure 'cfg_xnat_dataset' in '/home/mih/.config/datalad/procedures' 
[DEBUG  ] Looking for procedure 'cfg_xnat_dataset' in '/etc/xdg/datalad/procedures' 
[DEBUG  ] Looking for procedure 'cfg_xnat_dataset' in '/tmp/rp/sub3/.datalad/procedures' 
[DEBUG  ] Looking for procedure 'cfg_xnat_dataset' in '/home/mih/hacking/datalad/neuroimaging/datalad_neuroimaging/resources/procedures' 
[DEBUG  ] Looking for procedure 'cfg_xnat_dataset' in '/home/mih/hacking/datalad/git/datalad/resources/procedures' 
[DEBUG  ] Cannot find procedure with name 'cfg_xnat_dataset' [run_procedure(cfg_xnat_dataset)] 
````

My usecase is a particular configuration that needs to be applied to all subdatasets of a superdataset. It makes sense to me to put such a procedure in the root dataset. I also want to use it with `addurls --cfg-proc`. I have not investigated whether this usage pattern suffers from the same lookup issue.

mih:
Given that #4803 was rejected, here is a summary of the key problem:

I have a dataset that contains a config procedure. I want to make that procedure accessible to `addurls --cfg-proc` for the post-create configuration of subdatasets in a uniform way. In my use case I cannot rely on a specific user config (different users), nor machine config (different machines). All logic and configuration to update a superdataset (and all its subdatasets) should come from the dataset itself, which is only briefly cloned onto a machine for the purpose of updating from a set of generated URLs. Afterwards, the update is pushed and the local clone is destroyed -- and with it all local configuration. This rules out a configuration based on result hooks, as such configuration is not read from the persistent dataset config.

yarikoptic:
@mih please have a look at #5156 and give your blessing or "mneh, since ..."

mih:
The fix in #5156 is not exactly addressing the issue, but it is good enough for now and for me.

PR #4803, opened 2020-08-05
ENH: Look for dataset procedure in refds on `create`

mih:
Supports use cases where a particular configuration needs to be applied
to all subdatasets of a superdataset. It makes sense to put such a
procedure in the root dataset.

This change temporarily expands any existing search location
configuration to include any locations configured in a superdataset
(including the default location).

Fixes gh-4802

TODO:
- [ ] Add test after feedback on the general direction of the implementation

### DIFF
--- a/datalad/core/local/create.py
+++ b/datalad/core/local/create.py
@@ -39,8 +39,9 @@ from datalad.support.constraints import (
 from datalad.support.param import Parameter
 from datalad.utils import (
     getpwd,
-    assure_list,
+    ensure_list,
     get_dataset_root,
+    Path,
 )
 
 from datalad.distribution.dataset import (
@@ -268,7 +269,7 @@ class Create(Interface):
         assert(path is not None)
 
         # assure cfg_proc is a list (relevant if used via Python API)
-        cfg_proc = assure_list(cfg_proc)
+        cfg_proc = ensure_list(cfg_proc)
 
         # prep for yield
         res = dict(action='create', path=str(path),
@@ -499,8 +500,34 @@ class Create(Interface):
         )
 
         for cfg_proc_ in cfg_proc:
-            for r in tbds.run_procedure('cfg_' + cfg_proc_):
-                yield r
+            loc_var = 'datalad.locations.dataset-procedures'
+            proc_locs = None
+            if isinstance(refds, Dataset) and refds.path != tbds.path:
+                # we have a parent dataset, patch the config to temporarily also
+                # look into its procedure store, if there is any
+                # https://github.com/datalad/datalad/issues/4802
+                proc_locs = [
+                    l if Path(l).is_absolute() else str(refds.pathobj / l)
+                    for l in ensure_list(refds.config.obtain(loc_var))
+                ]
+            old_proc_locs = None
+            try:
+                if proc_locs:
+                    # save a possible existing override
+                    old_proc_locs = tbds.config.overrides.pop(loc_var, None)
+                    # place temp override, we do not need to worry about merging
+                    # them with prev existing ones. Any overrides must come from
+                    # a global source that would also have affected refds.config
+                    tbds.config.overrides[loc_var] = proc_locs
+                    tbds.config.reload(force=True)
+                for r in tbds.run_procedure('cfg_' + cfg_proc_):
+                    yield r
+            finally:
+                if old_proc_locs is None:
+                    tbds.config.overrides.pop(loc_var, None)
+                else:
+                    tbds.config.overrides[loc_var] = old_proc_locs
+                tbds.config.reload(force=True)
 
         # the next only makes sense if we saved the created dataset,
         # otherwise we have no committed state to be registered


kyleam:
The direction of the implementation looks fine to me.

The use case makes sense to me, and the only concern I can think of is that this special case might confuse some users into thinking the parent dataset's config applies more generally.  I don't think that's likely to be an issue though, especially if the documentation for `--cfg-proc` explains the special treatment.

yarikoptic:
Shouldn't any config related handling be implemented within config code (thus adding more files to watch for changes etc), not per specific command?

mih:
> Shouldn't any config related handling be implemented within config code (thus adding more files to watch for changes etc), not per specific command?

It is not like we are not using a `ConfigManager` (in fact, it uses two). what you seem to be asking for is to expand `ConfigManager` to track an arbitrary number of additional source files. This is possible, but it adds the necessity to define an order in which to consider these sources, to add one `git-config` call per additional file, in #4829 expand the internal representation to keep such configuration separate.

Altogether that is a substantial complication of code that needs to run extremely often vs this localized extraction of a single variable to be used as an override for a single call. If this turns out to be a common necessity, I'd say that we should consider such an RF. But this is the first time we ever encountered such a need.

In this particular case, such a solution would be particularly problematic, because we would be adding `.git/config` of another dataset to a dataset's `ConfigManager`.

yarikoptic:
FWIW: I have ran into the same gotcha with another command -- addurls, which uses 

```python
... ds.create(
                    spath,
                    result_xfm=None,
                    cfg_proc=cfg_proc,
                    return_type='generator')
```
FWIW, taking this PR (+merging master) resolved the original issue for me.

mih:
I will turn this into a draft until it makes sense for me to take another look.

yarikoptic:
BTW (forgot to formalize).  I think an alternative or complimentary solution could be -- we add a configuration variable to specify some additional location of procedures.  That would be more flexible, explicit, and not require this PR giving special treatment to procedures.

yarikoptic:
@mih do you still want to pursue this approach?

mih:
> BTW (forgot to formalize). I think an alternative or complimentary solution could be -- we add a configuration variable to specify some additional location of procedures. That would be more flexible, explicit, and not require this PR giving special treatment to procedures.

I am not sure if I understand your comment correctly. We do have a variable for that (`datalad.locations.dataset-procedures`) it is is being manipulated to add an additional location -- *automatically*. The point is that there was/is no way to indicate "look into the reference dataset too, whatever that is". There is no way for the ConfigManager to be aware of such caller context information. AFAICS it has to be provided by the calling command.

If that is all correct, it could be argued that this context information should come from the outermost caller (in #4802 this would be addurls). However, in the context of `create` being the outermost caller, this feature still makes sense -- so if implemented there, the addurls use case requires no further changes.

The only other way I can interpret your suggestion in the context of #4802 would be: If one thinks a dataset procedure needs to live in a dataset, it is solely the callers responsibility to set `datalad.locations.dataset-procedures` each time to include its momentary location.

Given that we both ran into the same problem (#4802 and https://github.com/datalad/datalad/pull/4803#issuecomment-689038268; which ultimately were both coming from `addurls`) and this change has worked for both of us, we could treat this as an indication that we want this. Please let me know either way.

If you decide that you prefer having to manually set the location for each call, I will replace this PR with a test that we can successfully override.

yarikoptic:
> We do have a variable for that (datalad.locations.dataset-procedures) it is is being manipulated to add an additional location -- automatically.

That one is a relative path within a corresponding dataset "acted upon".  IMHO should not be "overloaded". I was suggesting adding one more variable (e.g. `datalad.locations.extra-procedures`; ideally making it possible to be a list).  Then whenever you need procedure from the "super dataset" or somewhere else, you just explicitly specify that full path.
The question would be either it should be the first or the last to consider among locations?  If `extra` - sounds like the last (so may be name could be chosen better), but since it would be really the one to be specified by the user, IMHO it should be first to consider thus to allow not only to just find "extra" procedures but to overload the ones provided by the dataset without needing to modify it (e.g. a buggy one, or just outdated).

possible gotcha: as it is a config variable,  some subds's `.datalad/config` could overload it I guess :-/  I would resolve that by explicit documentation that generally it should not be "set" within dataset or .git/config.

If we decide that we would like generally to lookup procedures in `refds`, then internally it could specify/extend that variable for the duration of that specific command execution.

mih:
I dont see how any of these approaches can achieve the goal of having self-contained datasets that carry all information necessary for updating them. We need this for HCP, for XNAT, and anything else that needs to feed addurls --cfg-proc.

But OK, I guess we need to close this one and document that this is by design, and not a shortcoming.

yarikoptic:
> I dont see how any of these approaches can achieve the goal of having self-contained datasets that carry all information necessary for updating them

May be then do not discourage to add that extra config to .datalad/config in those super datasets. It gets loaded/applied, but more explicit and allows for embedding subdatasets providing those procedures

mih:
> > I dont see how any of these approaches can achieve the goal of having self-contained datasets that carry all information necessary for updating them
> 
> May be then do not discourage to add that extra config to .datalad/config in those super datasets. It gets loaded/applied, but more explicit and allows for embedding subdatasets providing those procedures

I cannot follow your statement. The fact that superdataset config is loaded but NOT applied in the context of creating a new subdataset is the fact that this PR wants to rectify in the context of procedure discovery. Maybe I am not aware of the feature you are referring to. What mechanism can I use to make `datalad run-procedure -d subds ...` consider the configuration of a superdataset that may be one, two, or many levels up a hierarchy, and have this information encoded in that superdataset, and not in the volatile execution environment?

yarikoptic:
> Maybe I am not aware of the feature you are referring to.

I am not referring to an existing feature,

> ... consider the configuration of a superdataset  ...

I keep suggesting to avoid hard-coding config-related semantic within a single command (that is what this PR does), and decouple configuration from implementation:
- add a new (hence -- not yet implemented, thus you cannot be aware) configuration item (e.g. `datalad.locations.extra-procedures`) which could be used to point to an additional location of procedures.  This would allow for "standardized" and more flexible discovery of procedures.  Probably only a global (not dataset specific) `cfg` should be consulted in `_get_procedure_implementation` to help avoid injection in malicious dataset configurations to run some bad code
- make `create` add an override (just via `patch_config`, currently used only for tests, but IMHO we need something like that at user level for command invocation overrides) for that configuration variable while invoking `tbds.run_procedure` to have that super-datasets `datalad.locations.dataset-procedures` (but as a full path) passed down to be picked up by `_get_procedure_implementation` downstream.

mih:
I don't see how this will work. I hope that someone can pick this issue up and provide an acceptable implementation. Given that we need to move on the underlying issue, I will give up and close this PR. We will have to hack up a local solution -- not sure how that would look like. But maybe just rebasing this patch each time an update is needed.

Issue #4801, opened 2020-08-05
GitRepo.get_hexsha() can be extremely slow (6min)


mih:
TL;DR: Should we switch from `git-show` to `git-log` for `GitRepo.format_commit()`. Or is there a way to further limit processing in `git-show` beyond `--no-patch`.

I discovered this via a `datalad status` call, which I thought would be hanging.

<details>
<summary>Here is the log level 1 output</summary>

```
% dl -l 1 status -e commit --untracked no -t raw
[DEBUG  ] Command line args 1st pass. Parsed: Namespace() Unparsed: ['status', '-e', 'commit', '--untracked', 'no', '-t', 'raw']                                                          [DEBUG  ] Discovering plugins                                                                [Level 5] Importing module datalad.core.local.status
[DEBUG  ] Building doc for <class 'datalad.core.local.status.Status'>                        [Level 5] Finished setup_parser                   
[DEBUG  ] Parsing known args among ['/usr/bin/datalad', '-l', '1', 'status', '-e', 'commit',
'--untracked', 'no', '-t', 'raw']                                      
[Level 5] Running: ['git', '--git-dir=', 'config', '-z', '-l', '--show-origin']             
[Level 8] Finished running ['git', '--git-dir=', 'config', '-z', '-l', '--show-origin'] with
status 0                                                                                     
[Level 2] Entered eval_func for <function Status.__call__ at 0x7f119481f378>                
[DEBUG  ] Determined class of decorated function: <class 'datalad.core.local.status.Status'>
[Level 2] Returning generator_func from eval_func for <class 'datalad.core.local.status.Statu
s'>
[Level 5] Parsed ri /data/project/enki/processed into fields {'path': '/data/project/enki/pro
cessed', 'scheme': '', 'username': '', 'password': '', 'hostname': '', 'port': '', 'query': '
', 'fragment': ''}
[Level 5] Detected file ri
[DEBUG  ] Resolved dataset for status reporting: /data/project/enki/processed
[Level 5] Detected <class 'datalad.support.annexrepo.AnnexRepo'> at /data/project/enki/processed
[Level 5] Running: ['git', '--git-dir=', 'config', '-z', '-l', '--show-origin']
[Level 8] Finished running ['git', '--git-dir=', 'config', '-z', '-l', '--show-origin'] with status 0
[Level 5] Found 1 previous connections
[...]

PR #4806, opened 2020-08-06
OPT: Faster approach to GitRepo.get_hexsha()

mih:
The previous `git-show` can be slow with more complex commits (e.g.
octopus merge commit, see
https://github.com/datalad/datalad/issues/4801)

Switch to `git-rev-parse` instead as suggested by @kyleam in
https://github.com/datalad/datalad/issues/4801#issuecomment-669480561

Behavior of `GitRepo.get_hexsha()` should stay constant for all
observed usage patterns. In particular raising a ValueError when
querying for a specific, but non existent commitish is critical
for subdataset handling.

Fixes gh-4801

I will tackle an analog change (using `git-log`)  in `GitRepo.format_commit()` in a separate PR.

### DIFF
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -1741,14 +1741,28 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
 
         Returns
         -------
-        str or, if there are not commits yet, None.
+        str or, if no commitish was given and there are no commits yet, None.
+
+        Raises
+        ------
+        ValueError
+          If a commitish was given, but no corresponding commit could be
+          determined.
         """
-        stdout = self.format_commit("%{}".format('h' if short else 'H'),
-                                    commitish)
-        if stdout is not None:
-            stdout = stdout.splitlines()
-            assert(len(stdout) == 1)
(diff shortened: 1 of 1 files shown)


datalad/support/gitrepo.py
```diff
@@ -1741,14 +1741,28 @@ def get_hexsha(self, commitish=None, short=False):
...
-        stdout = self.format_commit("%{}".format('h' if short else 'H'),
-                                    commitish)
-        if stdout is not None:
-            stdout = stdout.splitlines()
-            assert(len(stdout) == 1)
-            return stdout[0]
+        # use --quiet because the 'Needed a single revision' error message
+        # that is the result of running this in a repo with no commits
+        # isn't useful to report
+        cmd = ['rev-parse', '--verify', '{}^{{commit}}'.format(
```
kyleam, line 1755: The comment above mentions `--quiet`, but it's not passed here.

mih: Arrgh, thx, wil fix.

mih: Force-pushed an update

Issue #1455, opened 2017-04-06
add locking


yarikoptic:
now that we have singletons all around we should add at least trivial locking that those *Repo objects should lock (where have permissions) operations on those upon themselves until they die off. So lock probably should contain host, PID, and *Repo object ID, so if dies off without cleaning up the lock, we could always (try) to verify if it is safe for us to remove that lock and proceed with a fresh one.  who knows -- may be it wouldn't be that painful ;)

mih:
This could be extra motivation: https://travis-ci.org/datalad/datalad/jobs/229865267

````
Traceback (most recent call last):
  File "/home/travis/virtualenv/python3.4.4/lib/python3.4/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/build/datalad/datalad/datalad/tests/utils.py", line 592, in newfunc
    return t(*(arg + (filename,)), **kw)
  File "/home/travis/build/datalad/datalad/datalad/interface/tests/test_crawl_init.py", line 27, in _test_crawl_init
    crawl_init(args=args, template=template, template_func=template_func, save=save)
  File "/home/travis/build/datalad/datalad/datalad/interface/crawl_init.py", line 94, in __call__
    % (exc_str(exc), get_func_kwargs_doc(pipeline_func)))
RuntimeError: Running the pipeline function resulted in Lock at '/tmp/datalad_temp__test_crawl_initxp7ka5gp/.git/index.lock' could not be obtained [util.py:open:314].FYI this pipeline only takes the following args: ['dataset', 'versioned_urls', 'topurl', 'versions_overlay_level', 'leading_dirs_depth', 'prefix', 's3_prefix']
````

or I broke something...

yarikoptic:
I think such errors do pop up in different places from time to time... I am wondering if it is not some side effect of Singleton dying and somehow not invoking gitpython's deconstructor or smth which should have removed it... Or race condition when recently Singleton died but gitpyhton object not yet but then new class for that location is requested.
Can you reproduce it reliably?

mih:
No, only saw it once in travis (see link).

yarikoptic:
you restarted it, right? (I was about to do so to check as well if reproducible)

yarikoptic:
FTR we now have locking used while extracting archives and in our special remotes . It is not  repository locking which is the topic of this issue so I will keep it open

mih:
[...]

Issue #450, opened 2016-05-20
install --recursive could run jobs in parallel


yarikoptic:
created a new label `do-in-parallel` to collect use-cases which beg for parallelization, and seems to be trivial to be implemented

specific implementation approach yet to be discussed, thus tagged for conference agenda

``` shell
$> datalad install --recursive 1 -l info --source http://datasets.datalad.org/
2016-05-20 19:01:20,681 [INFO   ] Creating a new git repo at /mnt/datasets/datalad/test-crawled/datasets.datalad.org (install.py:634)
2016-05-20 19:01:20,715 [ERROR  ] 'git clone -v http://datasets.datalad.org/ /mnt/datasets/datalad/test-crawled/datasets.datalad.org' returned with exit code 128
| stderr: 'Cloning into '/mnt/datasets/datalad/test-crawled/datasets.datalad.org'...
| fatal: repository 'http://datasets.datalad.org/' not found
| ' (gitrepo.py:321)
2016-05-20 19:01:20,716 [INFO   ] Creating a new git repo at /mnt/datasets/datalad/test-crawled/datasets.datalad.org (install.py:634)
2016-05-20 19:01:20,840 [INFO   ] Creating a new git repo at /mnt/datasets/datalad/test-crawled/datasets.datalad.org/openfmri (install.py:634)
2016-05-20 19:01:20,986 [INFO   ] Creating a new git repo at /mnt/datasets/datalad/test-crawled/datasets.datalad.org/openfmri/ds000011 (install.py:634)
...
```

Since we are cloning them manually first, and then only 'registering' as a submodule under that path, it sounds like it should be trivial to implement in parallel -- those are independent repositories to clone.

yarikoptic:
related complication:  inability to serialize GitPython's Git -- https://github.com/gitpython-developers/GitPython/issues/504

mih:
Many things became simpler in this regard. However `install/get` still do not support this. With http://handbook.datalad.org/en/latest/basics/101-146-gists.html#parallelize-subdataset-processing the pressure to have built-in parallelization is much reduced.
However, that paradigm is not applicable to subdataset installation. I tried parallel installation of 42k subdatasets via:

```
datalad -f '{path}' subdatasets | xargs -n 1 -P10 datalad get -n
```

after I estimated a standard serial run `datalad get -n -r` to take somewhere between 25h and 40h. It finished after 18h with parallelization (not a stellar result). However, it left the dataset in a broken state due to random locking issues of this kind:

[...]

PR #4364, opened 2020-03-31
BF: .reload() config upon each access to repo.config

yarikoptic:
Fixes #4363  although I am not sure if that is the fix we actually want.

this difference probably has  bad effect on performance, although on smaug only got:
```
       before           after         ratio
     [2a1e96c1]       [efe2fe0c]
     <bf-config~1>       <bf-config>
+      1.66±0.07s       2.09±0.02s     1.25  api.supers.time_createadd_to_dataset
+      1.81±0.02s       2.14±0.01s     1.18  api.supers.time_createadd
-           48.28            36.48     0.76  core.runner.track_overhead_heavyout_online_through
+           -1.81              2.2    -1.22  core.runner.track_overhead_heavyout
```

I really think we should stop trying to be "reactive" to outside
config changes -- it hurts performance and probably cannot be
relied upon anyways.  ~~And when we know that the config changed outside, we explicitly reload -- I will submit a separate alternative PR to fix #4363 that way~~

Issue #4363, opened 2020-03-31
dataset.config fails to detect changed config file (seems to be master specific)


yarikoptic:
Origin: https://github.com/datalad/datalad-extensions/pull/7#issuecomment-606725942
```
(git)lena:~datalad/datalad-master[master]git
$> python -m nose -s -v datalad/metadata/tests/test_base.py:test_get_metadata_type
datalad.metadata.tests.test_base.test_get_metadata_type ... HERE0
HERE1:  /home/yoh/.tmp/datalad_temp_test_get_metadata_typet6b5ptfh
FAIL
Versions: appdirs=1.4.3 boto=2.49.0 cmd:7z=16.02 cmd:annex=7.20190819+git2-g908476a9b-1~ndall+1 cmd:bundled-git=2.20.1 cmd:git=2.20.1 cmd:system-git=2.25.1 cmd:system-ssh=8.1p1 exifread=2.1.2 git=3.0.7 gitdb=2.0.6 humanize=0.5.1 iso8601=0.1.12 keyring=18.0.1 keyrings.alt=3.2.0 msgpack=0.5.6 mutagen=1.40.0 requests=2.22.0 tqdm=4.30.0 wrapt=1.11.2
Obscure filename: str=b' "\';a&b&c\xce\x94\xd0\x99\xd7\xa7\xd9\x85\xe0\xb9\x97\xe3\x81\x82 `| ' repr=' "\';a&b&cΔЙקم๗あ `| '
Encodings: default='utf-8' filesystem='utf-8' locale.prefered='UTF-8'
Environment: PATH='/home/yoh/proj/datalad/datalad-master/venvs/dev3/bin:/home/yoh/gocode/bin:/home/yoh/gocode/bin:/home/yoh/bin:/home/yoh/.local/bin:/usr/local/bin:/usr/bin:/bin:/usr/local/games:/usr/games:/sbin:/usr/sbin:/usr/local/sbin' LANG='en_US.UTF-8' GIT_PAGER='less --no-init --quit-if-one-screen' GIT_PYTHON_GIT_EXECUTABLE='/usr/lib/git-annex.linux/git'

======================================================================
FAIL: datalad.metadata.tests.test_base.test_get_metadata_type
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
[...]

PR #4808, opened 2020-08-07
NF: Inter-process locking for configuration modifications

### LINKED ISSUE #4809: ConfigManager does not support `.git` pointer files
Having a `.git` file pointing to the location of the git-dir is a legit dataset setup. However, the `ConfigManager` insists on monitoring `<dsroot>/.git/config`. A file which may not exist.

In such a case the decision making of when to reload the configuration is impaired, as mtime assessment will be incomplete.

Moreover, the location of the `local` config file is configured once at the creation of the config manager, hence will never change for a repository/dataset instance, because we use singletons.

This is not an issue for configuration modification, because we call `git-config` which does the right thing. But any `ConfigManager.reload(force=False)` will not pick up changes to the local configuration.

Discovered looking at the error in #4808 (which uses a file placed next to `ConfigManager._repo_cfgfname`, which points to the wrong location for any submodule that has not yet been processed by `_fixup_submodule_dotgit_setup()`.

mih:
As described here:
https://github.com/datalad/datalad/issues/450#issuecomment-670365732
the manipulation of configuration is a collision point for multiple
processes interacting with a single dataset.

This change introduces an inter-process locking mechanism using the
`fasteners` package. Relevant bits from the docs:

- does not require any cleanup
- lock gets dropped automatically if the process crashes, even if
  `__exit__` is not executed

Locking behavior is logged at level 5 and looks like this:

```
[Level 5] Acquired file lock `b'/tmp/datalad_temp_test_obtaink1uf9sf9/.git/config_dataladlock'` after waiting 0.000s [1 attempts were required]
[Level 5] Unlocked and closed file lock open on `b'/tmp/datalad_temp_test_obtaink1uf9sf9/.git/config_dataladlock'`
```

The `fasteners` package is already a direct dependency. It seems to
provide the standard solution for this problem.

Doing locking only at config-write, instead of more global "dataset is
in use"-locks has the advantage of minimizing bottlenecks with external
parallelization approaches (i.e. via `xargs -P`). However, it is
obviously not sufficient in general.

One could debate whether the lock should also cover read operations to
avoid missing partial update done by process A when process B is
reloading a dataset configuration.

Related : #1455

Update: This PR now also fixes #4809

### DIFF
--- a/datalad/config.py
+++ b/datalad/config.py
@@ -22,13 +22,7 @@ from distutils.version import LooseVersion
 
 import re
 import os
-from os.path import (
-    join as opj,
-    exists,
-    getmtime,
-    abspath,
-    isabs,
-)
+from pathlib import Path
 from time import time
 
 import logging
@@ -86,9 +80,9 @@ def _parse_gitconfig_dump(dump, store, fileset, replace, cwd=None):
             continue
         if line.startswith('file:'):
             # origin line
-            fname = line[5:]
-            if not isabs(fname):
-                fname = opj(cwd, fname) if cwd else abspath(fname)
+            fname = Path(line[5:])
+            if not fname.is_absolute():
+                fname = Path(cwd) / fname if cwd else Path.cwd() / fname
             fileset.add(fname)
             continue
         if line.startswith('command line:'):
@@ -211,9 +205,8 @@ class ConfigManager(object):
         self._store = {}
         self._cfgfiles = set()
         self._cfgmtimes = None
-        self._dataset_path = None
-        self._dataset_cfgfname = None
-        self._repo_cfgfname = None
+        self._repo = None if dataset is None else \
+            dataset if hasattr(dataset, 'dot_git') else dataset.repo
         self._config_cmd = ['git', 'config']
         # public dict to store variables that always override any setting
         # read from a file
@@ -232,13 +225,7 @@ class ConfigManager(object):
             # when calling 'git config' to prevent a repository in the current
             # working directory from leaking configuration into the output.
             self._config_cmd = ['git', '--git-dir=', 'config']
-        else:
-            self._dataset_path = dataset.path
-            if source != 'local':
-                self._dataset_cfgfname = opj(self._dataset_path,
-                                             DATASET_CONFIG_FILE)
-            if source != 'dataset':
-                self._repo_cfgfname = opj(self._dataset_path, '.git', 'config')
+
         self._src_mode = source
         # Since configs could contain sensitive information, to prevent
         # any "facilitated" leakage -- just disable logging of outputs for
@@ -271,7 +258,7 @@ class ConfigManager(object):
             # we aren't forcing and we have read files before
             # check if any file we read from has changed
             current_time = time()
-            curmtimes = {c: getmtime(c) for c in self._cfgfiles if exists(c)}
+            curmtimes = {c: c.stat().st_mtime for c in self._cfgfiles if c.exists()}
             if all(curmtimes[c] == self._cfgmtimes.get(c) and
                    # protect against low-res mtimes (FAT32 has 2s, EXT3 has 1s!)
                    # if mtime age is less than worst resolution assume modified
@@ -295,10 +282,12 @@ class ConfigManager(object):
         if self._gitconfig_has_showorgin:
             run_args.append('--show-origin')
 
-        if self._dataset_cfgfname:
-            if exists(self._dataset_cfgfname):
+        dataset_cfgfile = None
+        if self._repo and self._src_mode != 'local':
+            dataset_cfgfile = self._repo.pathobj / DATASET_CONFIG_FILE
+            if dataset_cfgfile.exists():
                 stdout, stderr = self._run(
-                    run_args + ['--file', self._dataset_cfgfname],
+                    run_args + ['--file', str(dataset_cfgfile)],
                     log_stderr=True
                 )
                 # overwrite existing value, do not amend to get multi-line
@@ -320,10 +309,11 @@ class ConfigManager(object):
             cwd=self._runner.cwd)
 
         # always monitor the dataset cfg location, we know where it is in all cases
-        if self._dataset_cfgfname:
-            self._cfgfiles.add(self._dataset_cfgfname)
-            self._cfgfiles.add(self._repo_cfgfname)
-        self._cfgmtimes = {c: getmtime(c) for c in self._cfgfiles if exists(c)}
+        if dataset_cfgfile:
+            self._cfgfiles.add(dataset_cfgfile)
+        if self._src_mode != 'dataset' and self._repo:
+            self._cfgfiles.add(self._repo.dot_git / 'config')
+        self._cfgmtimes = {c: c.stat().st_mtime for c in self._cfgfiles if c.exists()}
 
         # superimpose overrides
         self._store.update(self.overrides)
@@ -610,9 +600,9 @@ class ConfigManager(object):
 
         # all other calls are modifications
         lockfile = None
-        if self._repo_cfgfname and ('--local' in args or '--file' in args):
+        if self._repo and ('--local' in args or '--file' in args):
             # modification of config in a dataset
-            lockfile = self._repo_cfgfname + '_dataladlock'
+            lockfile = self._repo.dot_git / 'config_dataladlock'
 
         # we are not protecting against concurrent modification of the
         # global config, because MIH cannot think of a use case and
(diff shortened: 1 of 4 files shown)


mih:
It looks as if the placement of the lockfile conflicts with some assumptions in `fasteners`

datalad/config.py
```diff
@@ -603,7 +604,25 @@ def _run(self, args, where=None, reload=False, **kwargs):
...
+
+        # all other calls are modifications
+        lockfile = None
+        if self._repo_cfgfname and ('--local' in args or '--file' in args):
+            # modification of config in a dataset
+            lockfile = self._repo_cfgfname + '_dataladlock'
+
+        # we are not protecting against concurrent modification of the
+        # global config, because MIH cannot think of a use case and
+        # it is unclear where a lockfile should be placed
```
yarikoptic, line 619: found on my laptop:
```
$> ls -l ~/.cache/datalad/locks 
total 0
-rw------- 1 yoh yoh 0 Mar 15 21:36 downloader-auth.lck
```

yarikoptic:
```
datalad/downloaders/base.py-        self._lock = InterProcessLock(
datalad/downloaders/base.py-            op.join(
datalad/downloaders/base.py-                cfg.obtain('datalad.locations.cache'),
datalad/downloaders/base.py-                'locks',
datalad/downloaders/base.py:                'downloader-auth.lck'))
```

mih: Thx!

(looks like a stale lock! ;-)

kyleam:
The last commit leads to a `test_resulthooks:test_basics` failure ([example][1]).  This bit is failing with `None != 'bids'`:

```
    clone.get('subds')
    clone_sub = Dataset(clone.pathobj / 'subds')
    eq_(clone_sub.config.get('datalad.metadata.nativetype'), 'bids')
```

I haven't figured out what aspect of the last commit is responsible for the change in behavior.  A `clone_sub.config.reload()` above the assertion is sufficient to get it to pass.

[1]: https://travis-ci.org/github/datalad/datalad/jobs/715898284#L1997

mih:
I can confirm that assessment. I looked at the files that are being tracked at the time of the test, and the list is complete. It seems that the restructuring caused a reload() to no longer happen. I am still unclear where that relaod should be placed structurally. It should not be just in the tests, I think.

A candidate place would be in the hook handling code, but that has other issues I started thinking about. ATM I see these two as more independent issues than on initial assessment.

yarikoptic:
might be just yet another manifestation of what was tried to be addressed semi-blindly with https://github.com/datalad/datalad/pull/4364 with original issue in https://github.com/datalad/datalad/issues/4363

kyleam:
> I am still unclear where that relaod should be placed structurally. It should not be just in the tests, I think.

Yep, I agree that it shouldn't be in the tests.  (My comment was just intended as debugging information.)

mih:
> might be just yet another manifestation of what was tried to be addressed semi-blindly with #4364 with original issue in #4363

It seems this is different. Here all config files are tracked properly and a `reload(force=False)` picks up the config just fine.

mih:
I looked at this in more detail now, and I think that `run_procedure()` should trigger a config reload.  Details in https://github.com/datalad/datalad/pull/4808/commits/9cccdda58c492214ceecb0744058306aebba07a9

kyleam:
> I looked at this in more detail now, and I think that `run_procedure()` should trigger a config reload.

That spot makes sense to me.  It still annoys me that I don't see what about the changes in the second commit result in a `reload` being dropped.  Oh well.

mih:
> > I looked at this in more detail now, and I think that `run_procedure()` should trigger a config reload.
> 
> That spot makes sense to me.  It still annoys me that I don't see what about the changes in the second commit result in a `reload` being dropped.  Oh well.

Yes, same here. I also couldn't figure it out on 2nd attempt. Out of ideas.

datalad/config.py
mih: Done now.

mih:
So the windows test failure is weird.

```
ERROR: datalad.interface.tests.test_ls_webui.test_fs_traverse
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\appveyor\\AppData\\Local\\Temp\\1\\datalad_temp_tree_iqli6jv7'
```

This directory seems to be the root dir of the main dataset used in the test that was prepared with `with_tree`. Now the test is essentially a deep dive into the `ls_webui` stack. However, I could not find a trace of "another process" being used explicitly. That leaves a batched annex as a likely option, and indeed:

```
{"info:annex_options:['--bytes']:json:True:path:C:\\Users\\mih\\AppData\\Local\\Temp\\datalad_temp_tree_77rgc_qf:git_options:['-c', 'annex.merge-annex-branches=false']:annex_cmd:info": BatchedAnnex(cmd=<<['git', '-c', ++105 chars++ch']>>, output_proc=<function>, path=<<'C:\\Users\\mi++48 chars++_qf'>>)}
```

However, a `del annex` doesn't stop it. Neither does an additional `del repo`.

Setting debug level to `1` doesn't reveal any additonal info other than rmtree trying to remove that dir for a bunch of seconds and then failing after 100 attempts.

Does anybody have a clue what is going on here (I assume somewhere some repo handle is still alive and that prevents `del` from being effective.

It might also make sense to disable that test on windows entirely. I doubt that `webui` code will ever be executed on windows -- if I interpret the name correctly.

mih:
So this looks good to me, but there is a stalled annex in the crawler tests that I have personally not seen before.

kyleam:
> So this looks good to me, but there is a stalled annex in the crawler tests that I have personally not seen before.

Looking through some of the recent builds, it seems plausible that it's somehow triggered by these changes because I see it for three of this PR's builds but didn't spot the stall in any other crawler builds from around the same time.

mih:
Yes, same here. The issue remains in #4829 so we are stuck for now.

mih:
https://github.com/mih/datalad/commit/fb2fac39c76d6c57bdbf6f2e964c3b1d694c3239 Makes it pass for me (github workflow).

datalad/config.py
```diff
@@ -21,13 +22,7 @@
...
 import re
 import os
-from os.path import (
-    join as opj,
-    exists,
-    getmtime,
-    abspath,
-    isabs,
-)
+from pathlib import Path
```
bpoldrack, line 25: General remark: It seems quite inconsistent to me, where we import `Path`from. Half the code base gets it from `pathlib` and the other half from `datalad.utils`.

datalad/config.py
```diff
@@ -85,9 +80,9 @@ def _parse_gitconfig_dump(dump, store, fileset, replace, cwd=None):
             continue
         if line.startswith('file:'):
             # origin line
-            fname = line[5:]
-            if not isabs(fname):
-                fname = opj(cwd, fname) if cwd else abspath(fname)
+            fname = Path(line[5:])
+            if not fname.is_absolute():
+                fname = Path(cwd) / fname if cwd else Path.cwd() / fname
```
bpoldrack, line 85: This seems incorrect to me. The dump comes from a `git config` call that is executed in `self._runner.cwd`, right? So, how would `Path.cwd()` be relevant? Shouldn't it be relative to CWD of the git-config call? This may usually be the same, but I think we tend to shoot our feet with this kind of assumption as soon as someone changes something.

bpoldrack (commented):
Left minor comments. Otherwise LGTM

datalad/config.py
mih: It is of value to keep the dependencies of this code to the rest of datalad at the minimum level. Unless there is a reason to enforce consistency, I'd rather not be consistent here.

datalad/config.py
kyleam: > Shouldn't it be relative to CWD of the git-config call?

That's the `if cwd` above.

mih: Just to clarify: Are you saying that the translation of the code to pathlib introduces a behavior change? Or are you saying what we used to do is wrong? In the latter case, I'd prefer to file that as an issue and deal with it in a dedicated PR.

bpoldrack: @mih: The latter.

@kyleam: Yes, but that's kind of the problem. Now we call git-config (possibly with a `cwd` argument or whatever logic) to then call a parsing helper, where we need to make sure to end up passing the same `cwd`. I think those two should be linked.

datalad/config.py
bpoldrack: The only reason is, that there's hardly a point in centralizing the import from pathlib in the first place, if we don't stick to it. Not a merge blocker though.

bpoldrack:
FTR: Re: mih@fb2fac3

Kind of an ugly "solution", but keeps the issue from blocking everything else - so I'm fine with that.

datalad/config.py
kyleam: What you just said is different than saying it's "incorrect".  You can audit the three call sites and see that they all explicitly pass `cwd`.  It's outside the scope of this PR, but I agree that tighter coupling would be better here or at the very least the probably-not-what-you-want "cwd=None => .cwd()" fallback could be dropped in favor of a mandatory argument.

datalad/config.py
mih: I believe it is OK to not be consistent for a reason. And the reason here is that ConfigManager is needed very early in DataLad and the more of DataLad it needs to import, the more difficult it becomes to make changes in the rest of DataLad.

datalad/config.py
mih: Ok, so let's keep things the way they are until a white knight comes and cleans out the cow shit ;-)

mih:
Given we have approval for #4829 I am closing this one. Thx for all input!

PR #4811, opened 2020-08-07
ENH(TST): remove from a setup_package a fixture needed for GitPython

yarikoptic:
Since no GitPython is used any longer, no need to tune env variables for it


mih:
Thx!

PR #4824, opened 2020-08-15
BF: fix authentication + download from S3 for NDA

yarikoptic:
- Uses new version of NDA authenticator to obtain S3 credentials for download
- Makes it possible to download from NDA bucket(s) which are quite "locked down"
  - default boto3 access to a bucket with verification would fail, so upon that specific error we would try to connect without validation
  - getting a key for which versionId is known is not possible while specifying that versionId.  And default boto behavior is to do so.  In this PR I change behavior to "kill" version_id from the key we are requesting if original url requested did not have versionId.

Although it is bug fix in nature, I am ok to reposition it against `master` if so deemed better.

### DIFF
--- a/datalad/downloaders/configs/nda.cfg
+++ b/datalad/downloaders/configs/nda.cfg
@@ -9,6 +9,25 @@ authentication_type = nda-s3
 url = https://ndar.nih.gov/access.html
 type = nda-s3
 
+# Development instance
+[provider:NDA-development]
+url_re = https://development.nimhda.org($|/.*)
+credential = NDA-development
+# or some kind of html_form???
+authentication_type = http_basic_auth
+
+[credential:NDA-development]
+type = nda-s3
+
+[provider:NDA-development-s3]
+url_re = s3://TODO?
+credential = NDA-development-s3
+authentication_type = nda-s3
+
+[credential:NDA-development-s3]
+type = nda-s3
+
+
 #
 # Configuration used by the crawler to fetch "interesting" selection of stuff
 # from miNDAR which has its own authentication scheme, and actually we might need
--- a/datalad/downloaders/configs/nda.cfg
+++ b/datalad/downloaders/configs/nda.cfg
@@ -9,25 +9,6 @@ authentication_type = nda-s3
 url = https://ndar.nih.gov/access.html
 type = nda-s3
 
-# Development instance
-[provider:NDA-development]
-url_re = https://development.nimhda.org($|/.*)
-credential = NDA-development
-# or some kind of html_form???
-authentication_type = http_basic_auth
-
-[credential:NDA-development]
-type = nda-s3
-
-[provider:NDA-development-s3]
-url_re = s3://TODO?
-credential = NDA-development-s3
-authentication_type = nda-s3
-
-[credential:NDA-development-s3]
-type = nda-s3
-
-
 #
 # Configuration used by the crawler to fetch "interesting" selection of stuff
 # from miNDAR which has its own authentication scheme, and actually we might need
--- a/datalad/downloaders/s3.py
+++ b/datalad/downloaders/s3.py
@@ -238,6 +238,20 @@ class S3Downloader(BaseDownloader):
         # Consult about filename
         url_filename = get_url_straight_filename(url)
 
+        if 'versionId' not in params and key.version_id:
+            # boto adds version_id to the request if it is known present.
+            # It is a good idea in general to avoid race between moment of retrieving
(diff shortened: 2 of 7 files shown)


yarikoptic:
<details>
<summary>failed travis debug run is unrelated test_ria_basics fail I think we saw</summary> 

```shell
ERROR: datalad.distributed.tests.test_ria_basics.test_binary_data('datalad-test',)
    self.test(*self.arg)
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/datalad/tests/utils.py", line 271, in _wrap_skip_ssh
    return func(*args, **kwargs)
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/datalad/tests/utils.py", line 731, in _wrap_with_tempfile
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/datalad/tests/utils.py", line 731, in _wrap_with_tempfile
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/datalad/distributed/tests/test_ria_basics.py", line 468, in _test_binary_data
    ds.download_url(url, path=file, message="Add DICOM file from github")
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/datalad/distribution/dataset.py", line 503, in apply_func
    return f(**kwargs)
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/datalad/interface/utils.py", line 494, in eval_func
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/datalad/interface/utils.py", line 482, in return_func
    results = list(results)
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/datalad/interface/utils.py", line 469, in generator_func
datalad.support.exceptions.IncompleteResultsError: Command did not complete successfully. 1 failed:
[{'action': 'download_url',
  'message': "'ConsoleLog' object has no attribute 'yesno' "
             '[__init__.py:__getattribute__:101]',
  'path': '/tmp/datalad_temp__test_binary_datajhq0ejj_/dicomfile',
  'status': 'error',
  'type': 'file'}]
```
</details>

datalad/downloaders/configs/nda.cfg
```diff
@@ -9,6 +9,25 @@ authentication_type = nda-s3
...
+url_re = https://development.nimhda.org($|/.*)
+credential = NDA-development
+# or some kind of html_form???
+authentication_type = http_basic_auth
+
+[credential:NDA-development]
+type = nda-s3
+
+[provider:NDA-development-s3]
+url_re = s3://TODO?
```
kyleam, line 23: Should this TODO remain?

datalad/downloaders/tests/test_s3.py
```diff
@@ -158,3 +158,12 @@ def test_boto_host_specification(tempfile):
     with swallow_outputs():
         providers.download(url_dandi1, path=tempfile)
     assert_equal(md5sum(tempfile), '97f4290b2d369816c052607923e372d4')
+
+
+def test_restricted_bucket_on_NDA():
```
kyleam, line 163: So, this only runs on your local machine, correct?

datalad/downloaders/s3.py
```diff
@@ -238,6 +238,20 @@ def get_downloader_session(self, url, **kwargs):
         # Consult about filename
         url_filename = get_url_straight_filename(url)
 
+        if 'versionId' not in params and key.version_id:
+            # boto adds version_id to the request if it is known present.
+            # It is a good idea in general to avoid race between moment of retrieving
+            # they key information and actual download.
```
kyleam, line 244: s/they/the/

datalad/downloaders/tests/test_s3.py
yarikoptic: yes unfortunately.  NDA access, and to this specific dataset is required and it is a not that easy procedure to get. And with their specific settings for the bucket - the value from the test is really in talking to NDA.
May be at some point we could setup a github workflow worker which would sit on "trusted" infrastructure and has access to the credentials, but then we might need to (as I think I did with buildbot in the past) to make it run tests only if PRs submitted by trusted github users or a dedicated label is added to signal that PR diff is good (no malicious code which would expose secrets) to be tested on

datalad/downloaders/configs/nda.cfg
yarikoptic: I think it is safe to remove all NDA-development* for now, will do, thanks

PR #4804, opened 2020-08-05
Temp try recent neurodebian devel annex (with custom backends support etc) [maint]

yarikoptic:

PR #4831, opened 2020-08-25
BF(TST): Define NOSE_OPTS as an array to start with

yarikoptic:
CPed from https://github.com/datalad/datalad/pull/4804


yarikoptic:
if no objections, I would merge this trivial change in a day

PR #4833, opened 2020-08-26
RF+ENH: eval_under_testloopfs - set also TMPDIR; make shellcheck happy

yarikoptic:
So this helper could be used for other tools, not only datalad.

Submitting against `maint` so it could be readily used when troubleshooting etc, and it is not used anywhere in our code AFAIK, so no possible harm should be done


yarikoptic:
d'oh -- I have said that I submit against `maint` but did against `master`.  Thanks @kyleam for approval.  I assume "maint" is ok with you?  (I would just merge/push locally to not waste CI cycles)

kyleam:
> I assume "maint" is ok with you?

Yes.  It's a change to a helper script in tools/.

PR #4837, opened 2020-08-27
Benchmarks: disable track overhead

yarikoptic:
They showed to be too unreliable.  although some did reveal significant regressions when looking historically:

http://datalad.github.io/datalad/#core.runner.track_overhead_heavyout_online_process

their values still jump in `[-15, 35]` range thus not reliable predictors.

Alternative could be just exclude them just while running on CI.  Let me know what you think

### DIFF
--- a/benchmarks/core.py
+++ b/benchmarks/core.py
@@ -108,24 +108,24 @@ class runner(SuprocBenchmarks):
         #overhead = round(min(overheads), 2)
         return overhead
 
-    def track_overhead_echo(self):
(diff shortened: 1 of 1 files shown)


mih:
These particular benchmarks never provided me with any usable signal. Given that I do not have historical records, I also have no ability to look at the outputs from a statistical angle. I would personally loose nothing if they were gone. Given that they produce CO2 each time they run, I'd vote for their removal.

PR #4829, opened 2020-08-24
Big overhaul of ConfigManager

mih:
This includes the changes from #4808 (I will rebase this PR, once it is merged).

Primarily, this PR fixes #4813. Along the road various RFs have led to optimization that reduce the number of calls to git-config and hopefully led to more readable code.

PR #4838, opened 2020-08-27
Address issue with correctly deleting *Repo instances

### LINKED ISSUE #4827: Need test helper to take down pending git-annex processes
We continue to have issues on windows with test cases that fail due to the `rmtree` call in the testbed take down code, because a git-annex process is still running. The latest installment is described here 
https://github.com/datalad/datalad/pull/4808#issuecomment-674381095 and an error log is  https://ci.appveyor.com/project/mih/datalad/builds/34672374/job/0og7edl8hw0nsbwl#L1073

In that particular case, our typical mitigation strategy (add `del repo` to the end of the test code) isn't working here -- presumably because multiple variables reference the same repo singleton. @bpoldrack had some spontaneous thought on the reasons elsewhere.

I think we would benefit from a test helper that reliably takes down all related processes.

bpoldrack:
Closes #4827

Opposed to what was suggested in the issue, what we need is not a helper, but have our clean up of *Repo instances (in particular `AnnexRepo` and its potential `BatchedAnnexes`) be more reliable.

- Switch from `__del__` methods to `weakref.finalize`-registered callbacks for Flyweight classes, which should be called more reliably than `__del__`. 
- Improve testing of those objects behavior regarding (re-)creation and deletion and proof they are properly destroyed once `gc` picks them up.

Note, that there's a failure on crippledFS (config-related) that is unrelated and happens in master, too!

### DIFF
--- a/datalad/distribution/tests/test_dataset.py
+++ b/datalad/distribution/tests/test_dataset.py
@@ -299,8 +299,6 @@ def test_require_dataset():
 @with_tempfile(mkdir=True)
 def test_dataset_id(path):
 
-    import gc
-
     ds = Dataset(path)
     assert_equal(ds.id, None)
     ds.create()
@@ -320,13 +318,11 @@ def test_dataset_id(path):
     # creating a new object for the same path
     # yields the same ID
     del ds
-    gc.collect()
     newds = Dataset(path)
     assert_equal(dsorigid, newds.id)
 
     # recreating the dataset does NOT change the id
     del newds
-    gc.collect()
 
     ds = Dataset(path)
     ds.create(annex=False, force=True)
@@ -334,7 +330,6 @@ def test_dataset_id(path):
 
     # even adding an annex doesn't
     del ds
-    gc.collect()
     ds = Dataset(path)
     ds.create(force=True)
     assert_equal(ds.id, dsorigid)
@@ -354,9 +349,15 @@ def test_dataset_id(path):
 def test_Dataset_flyweight(path1, path2):
 
     import gc
+    import sys
 
     ds1 = Dataset(path1)
     assert_is_instance(ds1, Dataset)
+    ds1.create()
+
+    # Don't create circular references or anything similar
+    assert_equal(1, sys.getrefcount(ds1) - 1)
+
     # instantiate again:
     ds2 = Dataset(path1)
     assert_is_instance(ds2, Dataset)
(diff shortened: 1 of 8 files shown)


bpoldrack:
Note, that there's PR #4849 as a follow-up, hopefully fixing issues with test tear-down due to still running batch processes.

datalad/distribution/tests/test_dataset.py
```diff
@@ -296,57 +298,51 @@ def test_require_dataset():
...
     assert_equal(ds.id.count('-'), 4)
     assert_equal(len(ds.id), 36)
+
+    # Ben: The following part of the test is concerned with creating new objects
+    #      and therefore used to reset the flyweight dict while keeping a ref to
+    #      the old object for comparison etc. This is ugly and in parts
+    #      retesting what is already tested in `test_Dataset_flyweight`. No need
+    #      for that. If we del the last ref to an instance and gc.collect(),
+    #      then we get a new instance on next request. This test should trust
+    #      the result of `test_Dataset_flyweight`.
```
mih, line 318: Should we strip the test down to cover only what is not tested in `test_Dataset_flyweight` then?

mih (commented):
This LGTM, and given the general helplessness of `__del__` is making me hopeful. I have little experience with all things flyweight, so I am not hitting "approve", but count me as a positive feedback.

datalad/distribution/tests/test_dataset.py
bpoldrack: I think that's what I did. No assertions about the objects anymore. Just dataset id.

bpoldrack:
Full stop. PR #4850 & Co. seem to fix the last remaining issue with respect to GC. Will try to prove it. If that's true, we can abandon the `gc.collect()` calls, which will make life much better.

bpoldrack:
Okay. With #4850 merged and a patch to `ConfigManager` in 0e1bc78 to remove circular references between the repos and their `config`, we no longer need a call to `gc.collect()` but have the *Repos destroyed immediately as the last reference is deleted! This is much better and deprecates PR #4849. Re-enabled the respective test that hung on windows for a proof.

yarikoptic:
Might be related to that crawler stalling and we might want to remove the workaround within https://github.com/datalad/datalad/pull/4829#pullrequestreview-476763261

PR #4841, opened 2020-08-28
RF: Replace trivial usage of _git_custom_command with call_git*()

mih:
This is recovering the first set of changes from the stale gh-4689

The main reason for not redoing #4689 is the lack of time to handle the complexity. So here is something simpler that doesn't reach the finish line, but is more comprehensible.


mih:
Cool, thx!

PR #4842, opened 2020-08-28
TST: Re-enable win test for issue replication

bpoldrack:
Trying to reproduce the windows issue in https://github.com/datalad/datalad/pull/4808#issuecomment-674381095 and see whether the switch to finalizers in PR #4838 does anything about it (calling `BatchedAnnex.close()` correctly).

No intention to merge. Will abort other CI runs.

### DIFF
--- a/datalad/distribution/tests/test_dataset.py
+++ b/datalad/distribution/tests/test_dataset.py
@@ -47,6 +47,7 @@ from datalad.tests.utils import (
     assert_is_not,
     assert_is_not_none,
     assert_not_equal,
+    assert_not_in,
     assert_raises,
     assert_repo_status,
     assert_result_count,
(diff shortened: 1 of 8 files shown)


bpoldrack:
FTR: Didn't pay attention to the crippledFS failure. 

I think I found the issue with the left-over batch process (https://github.com/datalad/datalad/pull/4808#issuecomment-674381095). The tear-down that leads to the actual failure is happening within `with_tree`'s `finally`-clause. The repo instances of the test function are simply not yet picked up by garbage collection, which is why PR #4838 alone doesn't fix it. So, I added an explicit `gc.collect()` to that `finally` and that fixes it. (ec051e8 now passes on appveyor)

Will go through the test decorators to find similar spots and come up with a dedicated PR.

Ping @mih, @kyleam, since you looked at the original issue, I think.

bpoldrack:
Close in favor of #4849.

PR #4847, opened 2020-08-31
RF: Remove last usage of GitRepo._git_custom_command from AnnexRepo

mih:
Follow-up to gh-4841 to eventually fulfill gh-4689

Diff looks complicated but it is basically moving the function body into
the `try` block to be able to use `call_git_items_()`


mih:
Extension tests fail at the "setup" stage. I restarted them a bunch of times. Either the neuroimaging one or the crawler one fail to start. But both have succeeded in more than one run (just not both at the same time).

bpoldrack (approved):
LGTM

mih:
Thx @bpoldrack merging this one too, as I do not anticipate objections.

PR #4848, opened 2020-08-31
RF: Test helpers no longer use GitRepo._git_custom_command

mih:
But use `call_git*()` methods instead.

Follow-up to gh-4841 to eventually fulfill gh-4689


bpoldrack (approved):
Seems correct.

mih:
There should be no controversy for this one.

PR #4849, opened 2020-08-31
BF: Safe guard rmtree

bpoldrack:
In issues like this: https://github.com/datalad/datalad/pull/4808#issuecomment-674381095
we encountered failures to tear-down tests due to hanging batch processes several times. AFAICT this is because garbage collection isn't that fast. Those processes are tied to an `AnnexRepo` instance, which isn't immediately destroyed if we leave the context of the last reference (like in a test decorator running the actual test within `try ... finally`). Therefore those processes aren't closed yet. Originally fixed the issue by adding `gc.collect()` into such `finally`-clause in PR #4842.
Looking for other spots that are likely to create a similar issue, I decided to go one level down and integrate that call into `rmtree`/`rmdir` helpers.

Re-enabled the windows test from above mentioned issue as proof.

Disclaimer: Some situations may require #4838 in addition for this change to take full effect.

### DIFF
--- a/datalad/interface/tests/test_ls_webui.py
+++ b/datalad/interface/tests/test_ls_webui.py
@@ -69,7 +69,7 @@ def test_ignored(topdir):
 
 
 # https://github.com/datalad/datalad/pull/4808#issuecomment-674381095
-@known_failure_windows
+#@known_failure_windows
 @with_tree(
     tree={'dir': {'.fgit': {'ab.txt': '123'},
                   'subdir': {'file1.txt': '124', 'file2.txt': '123'},
--- a/datalad/utils.py
+++ b/datalad/utils.py
@@ -445,10 +445,13 @@ def rmtree(path, chmod_files='auto', children_only=False, *args, **kwargs):
     # Give W permissions back only to directories, no need to bother with files
     if chmod_files == 'auto':
         chmod_files = on_windows
-    # TODO:  yoh thinks that if we could quickly check our Flyweight for
-    #        repos if any of them is under the path, and could call .precommit
-    #        on those to possibly stop batched processes etc, we did not have
-    #        to do it on case by case
+
+    # Make sure *Repo instances that used to point to `path` are
+    # actually destructed before trying to delete anything. This is particularly
+    # important regarding git-annex batch processes that might not have been
+    # closed yet otherwise.
+    gc.collect()
+
     # Check for open files
     assert_no_open_files(path)
 
@@ -468,6 +471,13 @@ def rmtree(path, chmod_files='auto', children_only=False, *args, **kwargs):
 
 def rmdir(path, *args, **kwargs):
     """os.rmdir with our optional checking for open files"""
+
+    # Make sure *Repo instances that used to point to `path` are
+    # actually destructed before trying to delete anything. This is particularly
+    # important regarding git-annex batch processes that might not have been
+    # closed yet otherwise.
+    gc.collect()
+
     assert_no_open_files(path)
     os.rmdir(path)
 


datalad/interface/tests/test_ls_webui.py
```diff
@@ -69,7 +69,7 @@ def test_ignored(topdir):
 
 
 # https://github.com/datalad/datalad/pull/4808#issuecomment-674381095
-@known_failure_windows
+#@known_failure_windows
```
mih, line 72: Remove (with comment above)?

mih (changes requested):
Cool, awesome find!

However, I have concerns regarding this particular approach to a fix. Both `rmtree` and `rmdir` are no test helpers, but generic utilities. This change will torpedo garbage collection for any call to `rmtree` or `rmdir`, a portion of those will happen in contexts where there is no dataset, no git-annex process to stop.

So if this is what needs to be done, I think we should have dedicated test helpers that do that, document the necessity for DataLad Python API users to do so, and possibly execute this (earlier) in the tear-down of the cmdline interface (before other cleanup code might run).

datalad/interface/tests/test_ls_webui.py
bpoldrack: True

bpoldrack:
What do you mean by "torpedoing" GC? It would be an unecessary call in those cases - granted. But performance hit shouldn't be heavy, I think. The other way around we'd need to keep that in mind all the time. That's what I tried to avoid, so we don't reintroduce the issue so easily.

There's actually another point: The reason we need that call is, that for some reason an instantiation of a `GitRepo` within a test generates six references to the object instead of one. I don't know why yet. But plain datalad import in IPython doing the same thing is just one. Why does that matter? Because if we would have only that one reference it would be collected immediatly by refcount-based garbage collection instead of requiring the generational garbage collection of `gc`. So, if I figure that out, we might be able to remove that call again, which is why I'm hesitating to put that all over the code and into the docs right now.

bpoldrack:
Looking into the above mentioned issue re refcount:

Seems we only create the need for `gc.collect()`, if we create a new repo in an already existing directory. In that case, function frames from the creation are somehow still alive after we returned from  `GitRepo.__init__()`. If we could get rid of this weird effect ...

mih:
> What do you mean by "torpedoing" GC? It would be an unecessary call in those cases - granted. But performance hit shouldn't be heavy, I think. The other way around we'd need to keep that in mind all the time. That's what I tried to avoid, so we don't reintroduce the issue so easily.

There is no way of telling that "It would be an unecessary call in those cases" -- it might not garbage collect repos, true, but other than that anything goes. My point is that we should consider actual datalad usage too. What evidence is there for "But performance hit shouldn't be heavy?" Anything that uses `rmtree` will trigger garbage collection now, outside of the datalad tests too. How heavy garbage collection will be completely depends on what a user is doing. In most analysis scripts datalad usage should be a footnote at most, but it will reach deep into the inner workings of the entire Python session with this change IMHO. I think side-effects are more or less unknown in a particular application.

bpoldrack:
Full stop. While reviewing PR #4850 I found it might solve the issue (see https://github.com/datalad/datalad/pull/4850#issuecomment-683766140). If that turns out to be true, this PR might be superfluous altogether, which would be much better.

datalad/utils.py
```diff
@@ -445,10 +445,13 @@ def rmtree(path, chmod_files='auto', children_only=False, *args, **kwargs):
...
-    # TODO:  yoh thinks that if we could quickly check our Flyweight for
-    #        repos if any of them is under the path, and could call .precommit
-    #        on those to possibly stop batched processes etc, we did not have
-    #        to do it on case by case
+
+    # Make sure *Repo instances that used to point to `path` are
+    # actually destructed before trying to delete anything. This is particularly
+    # important regarding git-annex batch processes that might not have been
+    # closed yet otherwise.
+    gc.collect()
```
yarikoptic, line 453: if we were to proceed with this (after checking the impact on performance through timing tests run and checking effect on our few benchmarks) -- might be worth consider addressing to be removed `TODO` comment - such check might be not that heavy (although might be shown to be heavier than just calling `gc.collect`)

bpoldrack: That check would be cheap - it's just looking up a key in a dict. There's no `precommit` to call anymore, though. What needs to be done is killing that object (it would be invalid after an `rmtree` anyway).

Anyways - this PR is discontinued, since we found a better solution.

bpoldrack:
That solution: https://github.com/datalad/datalad/pull/4838#issuecomment-684684173 renders this PR superfluous. Closing.

PR #4850, opened 2020-08-31
Remove GitRepo._git_custom_command usage from call_git() methods

mih:
This PR completes the puzzle (with #4847 and #4848) to get to where #4689 wanted to be (minus the transition to a CoreRepo/Dataset class set)

Given the severity of this change, it would be good to get additional comments (thx @bpoldrack already). This takes a previous key behavior (normalize_paths) out of the equation for a large number of git-calls (if not most).

We also have to decide what to do with `_git_custom_command()`. Its intended internal helper status would call for immediate removal, but defacto we have to accept that it may have been considered as a part of a more public API.

### DIFF
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -2161,7 +2161,7 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
                                      code=e.code, stdout=e.stdout,
                                      stderr=e.stderr,
                                      paths=ignored.groups()[0].splitlines())
-            lgr.log(5 if expect_fail else 11, e)
+            lgr.log(5 if expect_fail else 11, str(e))
             raise
 
         if err:
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -939,6 +939,7 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
         if kwargs:
             git_opts.update(kwargs)
 
+        self._git_runner = GitWitlessRunner(cwd=self.path)
         self.cmd_call_wrapper = runner or GitRunner(cwd=self.path)
         self._cfg = None
 
@@ -964,8 +965,10 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
             # under a directory some files of which are already tracked by git
             # use case: https://github.com/datalad/datalad/issues/3068
             try:
-                stdout, _ = self._git_custom_command(
-                    None, ['git', 'ls-files'], cwd=path, expect_fail=True
+                stdout, _ = self._call_git(
+                    ['-C', path, 'ls-files'],
+                    expect_fail=True,
+                    check_fake_dates=False,
                 )
                 if stdout:
                     raise PathKnownToRepositoryError(
@@ -977,26 +980,21 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
                 # assume that all is good -- we are not under any repo
                 pass
 
-        cmd = ['git', 'init']
+        cmd = ['-C', path, 'init']
         cmd.extend(kwargs.pop('_from_cmdline_', []))
         cmd.extend(to_options(**kwargs))
         lgr.debug(
             "Initialize empty Git repository at '%s'%s",
             path,
-            ' %s' % cmd[2:] if cmd[2:] else '')
+            ' %s' % cmd[3:] if cmd[3:] else '')
 
         try:
-            stdout, stderr = self._git_custom_command(
-                None,
+            stdout, stderr = self._call_git(
                 cmd,
-                cwd=path,
-                log_stderr=True,
-                log_stdout=True,
-                log_online=False,
-                expect_stderr=False,
-                shell=False,
                 # we don't want it to scream on stdout
-                expect_fail=True)
+                expect_fail=True,
+                # there is no commit, and none will be made
+                check_fake_dates=False)
         except CommandError as exc:
             lgr.error(exc_str(exc))
             raise
@@ -1373,13 +1371,14 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
 
         try:
             # without --verbose git 2.9.3  add does not return anything
-            add_out = self._git_custom_command(
-                files,
+            add_out = self._call_git(
                 # Set annex.largefiles to prevent storing files in
                 # annex with a v6+ annex repo.
-                ['git', '-c', 'annex.largefiles=nothing', 'add'] +
+                ['-c', 'annex.largefiles=nothing', 'add'] +
                 ensure_list(git_options) +
-                to_options(update=update) + ['--verbose']
+                to_options(update=update) + ['--verbose'],
+                files=files,
+                check_fake_dates=True,
             )
             # get all the entries
             for o in self._process_git_get_output(*add_out):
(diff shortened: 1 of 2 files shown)


datalad/support/gitrepo.py
```diff
@@ -2129,10 +2197,9 @@ def call_git(self, args, files=None,
         ------
         CommandError if the call exits with a non-zero status.
         """
-        out, _ = self._git_custom_command(files, ["git"] + args,
-                                          expect_stderr=expect_stderr,
-                                          expect_fail=expect_fail,
-                                          check_fake_dates=True)
+        out, _ = self._call_git(args, files,
+                                expect_stderr=expect_stderr,
+                                expect_fail=expect_fail)
```
bpoldrack, line 2202: Not clear to me, why `check_fake_dates` is gone here. True for other spots below, too.

bpoldrack (commented):
Confused by disappearing `check_fake_dates`, but otherwise LGTM.

edit: My fault.

datalad/support/gitrepo.py
mih: That is, because `_call_git()` always does it by default. And that, in turn, is because all usage of `_git_custom_command()`  by them had that setting. So this change is a no-change.

bpoldrack: Missed the default for `_call_git`, obv.

bpoldrack (approved):

bpoldrack:
Actually: More than approval. I'm happy!

Leaving the Runner behind fixed what I was debugging in context of #4838 and #4849: we used to have a traceback of a catched exception lingering in memory. This was referencing function frames we already left, leading to a refcount of 6 for a newly created `GitRepo` in an already existing directory (resulting in a `CommandError` from `git ls-files` in `create_empty_repo`.

Now, with this branch, this is not happening (I suspected some of our exception logging craziness in `Runner` to be the issue). Therefore we can probably drop the `gc.collect()` calls and rely and plain refcount-based garbage collection. Will update those PRs accordingly and try it.

So, I'm all for merging! ;-)

yarikoptic:
re

> We also have to decide what to do with _git_custom_command(). Its intended internal helper status would call for immediate removal, but defacto we have to accept that it may have been considered as a part of a more public API.

It is not a public method, in theory  it should be safe to remove it **but** elderly crawler does still use it :-/  so would need a round of crawler fixup/release

kyleam:
> It might be (in part or fully, didn't check) due to the workaround fix for WitlessRunner @kyleam recently introduced in #4835

Please, let's avoid blaming my workaround every time runner slowness is observed unless there's something to back it up.

mih:
> > It might be (in part or fully, didn't check) due to the workaround fix for WitlessRunner @kyleam recently introduced in #4835
> 
> Please, let's avoid blaming my workaround every time runner slowness is observed unless there's something to back it up.

;-)

If needed, I can provide a long list of location where we could become substantially faster. The only price would be incorrect behavior.

But jokes aside: We should be careful making yes/no decisions on intermediate steps for performance reasons. I think we all agree that the internals need to get simpler in order to be able to work on them (again) in a strategic fashion. The new runner was born out of that concrete need. Maintaining two runners is expensive, using the presence of the respective other one as an argument for no change isn't productive.

This PR is part of a longer project to sanitize the Repo API https://github.com/datalad/datalad/issues/3988

Performance is only one part of that picture. And the past months have shown that we have vast performance gains to harvest for operations that runs for days, not milliseconds (e.g. #4793)

mih:
> In general - I approve if it helps to mitigate some issues. Note though that it seems to come at a notable performance cost. According to benchmarks there is general (non significant) negative impact on run time across nearly all benchmarks with one benchmark even getting "significant"

Hmm, you attribute much more value to this kind of information than I do. I have to confess that I largely ignore these as invalid assessments of "performance" in any realworld task.

Let's consider these:

```
    2.16±0.2ms      2.27±0.04ms     1.05  core.runner.time_echo
   2.30±0.1ms      2.61±0.07ms    ~1.14  core.runner.time_echo_gitrunner
```

They continue to test the old Runner (hardcoded). The benchmarks indicate that this PR slows that code between 5 and 14%. 
Given the benchmark code below it is not obvious to me how this would work (if the estimates are reliable). If there is a link, this is worth investigating and would hint at an even more twisted setup that I am currently aware of.

```
class runner(SuprocBenchmarks):
    """Some rudimentary tests to see if there is no major slowdowns from Runner
    """
    def setup(self):
        self.runner = Runner()
        # older versions might not have it
        try:
            from datalad.cmd import GitRunner
            self.git_runner = GitRunner()
        except ImportError:
            pass

    def time_echo(self):
        self.runner.run("echo")

    def time_echo_gitrunner(self):
        self.git_runner.run("echo")
```

kyleam:
> Travis fail in the case of heavy logging seems to be legit -- logs is filled up (I didn't dig into fully)

Hopefully that's resolved with the latest push (445c2fd9a).

mih:
While @kyleam 's commit makes the benchmarks look good again (?!) https://github.com/datalad/datalad/pull/4851/checks?check_run_id=1051658582 suggests a slowdown of ~40% comparing `run(echo)` for both runners. If this can be trusted, we would have to conclude that there is a substantial per-invocation overhead for the witlessrunner (at the very least).

kyleam:
> https://github.com/datalad/datalad/pull/4851/checks?check_run_id=1051658582 suggests a slowdown of ~40% comparing `run(echo)` for both runners. If this can be trusted, we would have to conclude that there is a substantial per-invocation overhead for the witlessrunner (at the very least).

Locally I see a pretty consistent overhead for WitlessRunner.run():
```sh
# on 04e294b4a
python -m timeit -s 'from datalad import cmd; runner = cmd.Runner()' \
       'runner.run(["true"], log_stdout=False, log_stderr=False)'
# 200 loops, best of 5: 1.06 msec per loop
python -m timeit -s 'from datalad import cmd; runner = cmd.WitlessRunner()' \
       'runner.run(["true"])'
# 200 loops, best of 5: 1.92 msec per loop
```

yarikoptic:
FWIW, `time_echo` seems to be pretty consistent overall: http://datalad.github.io/datalad/#core.runner.time_echo?machine=drogon-neurodebian10.20190322.1&os=Linux%204.6.0-rc1%2B&python=3.7&branch=master .  It is the overhead one which were a wild chase, and "removed" in #4837 .

> While @kyleam 's commit makes the benchmarks look good again (?!)

hm, to me that only suggests that that overall benchmarking run I quoted was not to be trusted in terms of "assessing statistical significance" (no multiple comparisons correction?) since indeed the most recent one has no significant excursions although eyeballing https://github.com/datalad/datalad/pull/4851/checks?check_run_id=1051658582 suggests tendency to get a bit slower overall, but not as drastically, which is good.

Once again -- I approve and welcome this change, especially now that benchmarks rerun showed that it is not as drastic. I do welcome more correct over faster incorrect behavior as well ;-)  I just wanted to make a note that there is a performance cost -- something to keep in mind/annotated.

mih:
Ok, so there is overall positivity. Will merge then.

mih:
> > We also have to decide what to do with _git_custom_command(). Its intended internal helper status would call for immediate removal, but defacto we have to accept that it may have been considered as a part of a more public API.
> 
> It is not a public method, in theory it should be safe to remove it **but** elderly crawler does still use it :-/ so would need a round of crawler fixup/release

https://github.com/datalad/datalad-crawler/pull/82

yarikoptic:
FWIW, slow down indeed not drastic but consistently present.  Ran benchmarks locally overnight since 0.13.0:
image

that first bump after 0.13.3 is this PR, the last one - is merge of `maint` -- we might have brought smth :-/ will benchmark `maint` as well now

PR #4851, opened 2020-08-31
TST: Basic benchmarks for WitlessRunner and friends

mih:
They execute the same command as the old runner, hence the performance
estimates should be comparable.

This is complete shot in the dark, I have no local setup to run this.


yarikoptic:
I would have even got it within `maint` so we could see effects there when we fix up etc before next "major" release.
Otherwise: LGTM and seems to run
```
              n/a  0.0030749148749862343±0.00019      n/a  core.witlessrunner.time_echo
              n/a  0.0032512814999989814±9.3e-05      n/a  core.witlessrunner.time_echo_gitrunner
              n/a  0.0034958670000075167±7.7e-05      n/a  core.witlessrunner.time_echo_gitrunner_fullcapture
```
timing seems to be different from an existing runner benchmarks, I have not looked into why it could be though
```
      2.12±0.07ms      2.18±0.03ms     1.03  core.runner.time_echo
      2.38±0.06ms      2.53±0.06ms     1.06  core.runner.time_echo_gitrunner
```

yarikoptic:
I think it will be of value, so let's proceed

PR #4853, opened 2020-09-01
RF: Bypass `git-submodule update --init` when getting subdatasets

mih:
Doing it with config settings has two advantages.

- benefits from interprocess locking of datalad ConfigManager, such that
  we can "get" in parallel without running into locking issues.
  (`datalad -f '{path}' subdatasets | xargs -n 1 -P10 datalad get -n`)

- we can set the real source URL of the successful clone, whereas
  `GitRepo.update_submodule()` is what cause a guess to be made.

  For example, cloning a superdataset from

  git@github.com:datalad-datasets/human-connectome-project-openaccess.git

  would lead to invalid submodule URLs like:

  git@github.com:datalad-datasets/human-connectome-project-openaccess.git/HCP1200/100307

  whereas the actual clone came from

  http://store.datalad.org/8fd/bc694-2c2b-11ea-b467-0025904abcb0

This change has a high potential for missed corner-cases. It would be good to get this into master a long time before a planned release.

For example, we are resetting the existing master branch to the recorded commit. Before, we might have had to fix a detached state that was created by `git-submodule update`. I think this is a net win, but maybe we are missing out on other standardizations.

### DIFF
--- a/datalad/distribution/get.py
+++ b/datalad/distribution/get.py
@@ -309,9 +309,30 @@ def _install_subds_from_flexible_source(ds, sm, **kwargs):
                 res.get('path', None) == dest_path:
             _fixup_submodule_dotgit_setup(ds, sm_path)
 
-            # do fancy update
             lgr.debug("Update cloned subdataset {0} in parent".format(dest_path))
-            ds.repo.update_submodule(sm_path, init=True)
+            section_name = 'submodule.{}'.format(sm['gitmodule_name'])
+            # do not use `git-submodule update --init`, it would make calls
+            # to git-config which will not obey datalad inter-process locks for
+            # modifying .git/config
+            sub = GitRepo(res['path'])
+            # TODO we could simulatanously check out a configured branch
+            # and setup a tracking branch. This should be fine, as the implied
+            # config manipulation is local to the subdataset.
+            sub.call_git(['reset', '--hard', sm['gitshasum']])
+            ds.config.set(
+                '{}.active'.format(section_name),
+                'true',
+                reload=False, force=True, where='local',
+            )
+            ds.config.set(
+                '{}.url'.format(section_name),
+                # record the actual source URL of the successful clone
+                # and not a funky prediction based on the parent ds
+                # like ds.repo.update_submodule() would do (does not
+                # accept a URL)
+                res['source']['giturl'],
+                reload=True, force=True, where='local',
+            )
         yield res
 
     subds = Dataset(dest_path)
--- a/datalad/distribution/get.py
+++ b/datalad/distribution/get.py
@@ -323,7 +323,28 @@ def _install_subds_from_flexible_source(ds, sm, **kwargs):
             # if we are on a branch this hexsha will be the tip of that branch
             sub_orig_hexsha = sub.get_hexsha()
             # make sure we have the desired commit locally
-            sub.fetch(remote='origin', refspec=target_commit)
+            # expensive and possibly error-prone fetch conditional on cheap
+            # local check
+            if not sub.commit_exists(target_commit):
+                sub.fetch(remote='origin', refspec=target_commit)
+                # instead of inspecting the fetch results for possible ways
+                # with which it could failed to produced the desired result
+                # let's verify the presence of the commit directly, we are in
+                # expensive-land already anyways
+                if not sub.commit_exists(target_commit):
+                    res.update(
+                        status='error',
+                        message=(
+                            'Target commit %s does not exist in the clone, and '
+                            'a fetch that commit from origin failed',
+                            target_commit[:8]),
+                    )
+                    yield res
+                    # there is nothing we can do about this
+                    # MIH thinks that removing the clone is not needed, as a likely
+                    # next step will have to be a manual recovery intervention
+                    # and not another blind attempt
+                    break
             # checkout the desired commit
             sub.call_git(['checkout', target_commit])
             # did we detach?
--- a/datalad/distribution/get.py
+++ b/datalad/distribution/get.py
@@ -47,6 +47,7 @@ from datalad.support.gitrepo import (
     _fixup_submodule_dotgit_setup,
 )
 from datalad.support.exceptions import (
+    CommandError,
     InsufficientArgumentsError,
 )
 from datalad.support.network import (
@@ -326,7 +327,10 @@ def _install_subds_from_flexible_source(ds, sm, **kwargs):
             # expensive and possibly error-prone fetch conditional on cheap
             # local check
             if not sub.commit_exists(target_commit):
-                sub.fetch(remote='origin', refspec=target_commit)
+                try:
+                    sub.fetch(remote='origin', refspec=target_commit)
+                except CommandError:
+                    pass
                 # instead of inspecting the fetch results for possible ways
                 # with which it could failed to produced the desired result
                 # let's verify the presence of the commit directly, we are in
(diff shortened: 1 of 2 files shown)


bpoldrack (commented):
Not sure about any implications tbh, but going through the code it looks right.

datalad/distribution/get.py
```diff
@@ -309,9 +309,30 @@ def _install_subds_from_flexible_source(ds, sm, **kwargs):
...
-            ds.repo.update_submodule(sm_path, init=True)
+            section_name = 'submodule.{}'.format(sm['gitmodule_name'])
+            # do not use `git-submodule update --init`, it would make calls
+            # to git-config which will not obey datalad inter-process locks for
+            # modifying .git/config
+            sub = GitRepo(res['path'])
+            # TODO we could simulatanously check out a configured branch
+            # and setup a tracking branch. This should be fine, as the implied
+            # config manipulation is local to the subdataset.
+            sub.call_git(['reset', '--hard', sm['gitshasum']])
```
kyleam, line 321: Hmm, so this is a bit more aggressive than the logic in `update_submodule`.  We could end up resetting an unrelated branch.  What about checking out gitshasum and then using the same `get_merge_base` check from `update_submodule`?  (It looks like it could probably be pulled out into a helper fairly cleanly, but perhaps I'm missing some complication.)

Another potential problem here is that gitshasum might not have been pulled in with the clone.  `git submodule update` deals with this by following up with a direct fetch of the commit to try to retrieve it.  We could do the same, but at the very least I think we should translate the failed reset into an error result.

<details>
<summary>demo</summary>

Patch to make cloning a local repo behave like cloning a remote one.

```diff
diff --git a/datalad/core/distributed/clone.py b/datalad/core/distributed/clone.py
index 4b3dd898d..f11f99d6e 100644
--- a/datalad/core/distributed/clone.py
+++ b/datalad/core/distributed/clone.py
@@ -426,7 +426,7 @@ def clone_dataset(
             update=1,
             increment=True)

-        clone_opts = {}
+        clone_opts = {"no_local": True}

         if cand.get('version', None):
             clone_opts['branch'] = cand['version']
```

```sh
#!/bin/sh

set -eu

git init a
(
    cd a
    git commit --allow-empty -ma
    git init sub
    (
        cd sub
        git commit --allow-empty -msub
        git checkout master^0
        git commit --allow-empty -mdetached
        git config uploadpack.allowAnySHA1InWant true
    )
    git submodule add ./sub
    git commit -m'add sub'
    git -C sub checkout master
)

git clone a b
(
    cd b
    datalad get sub
)
```

</details>

mih:
> > * we can set the real source URL of the successful clone, whereas
> >   `GitRepo.update_submodule()` what cause a guess to be made.
> >   For example, cloning a superdataset from
> 
> That should be subdataset, right?

No, there first URL below is for the superdataset, but the differences reported next are for the "getted" subdatasets of it.

> If I'm thinking about this correctly, the main advantage of here, beyond .git/config being more informative to someone inspecting it, is that direct git callers will be able to update from the correct place. That sounds good. Or are there other places where this currently comes into play?

ATM I am not aware of any addition (side-)effects.

datalad/distribution/get.py
mih: Thanks for the comments and pointers. I pushed https://github.com/datalad/datalad/pull/4853/commits/96e7c8f11591574c0d880c524943838b11f5c2e3 that hopefully addresses them.

datalad/distribution/get.py
```diff
@@ -309,9 +309,75 @@ def _install_subds_from_flexible_source(ds, sm, **kwargs):
...
+            # modifying .git/config
+            sub = GitRepo(res['path'])
+            # record what branch we were on right after the clone
+            # TODO instead of the active branch, this should first consider
+            # a configured branch in the submodule record of the superdataset
+            sub_orig_branch = sub.get_active_branch()
+            # if we are on a branch this hexsha will be the tip of that branch
+            sub_orig_hexsha = sub.get_hexsha()
+            # make sure we have the desired commit locally
+            sub.fetch(remote='origin', refspec=target_commit)
```
kyleam, line 326: Thanks for the update.

I think we should guard this fetch with a `commit_exists` call because 1) in the common case that commit will exist and that local check is less expensive than the unnecessary fetching and 2) a direct fetch of a commit may fail depending on the remote's configuration

If the commit isn't present locally and the fetch fails, it seems like it'd be good to yield an error result so the operation can respect `--on-failure`.

mih: Done. Thx for the analysis. I fully agree.

kyleam: Thanks.  I've tweaked it a bit more to catch the CommandError and added a test.  Of course feel free to drop an of the changes.

kyleam (approved):

yarikoptic (changes requested):
I am not thrilled about re-implementing git functionality due to the stated "high potential for missed corner-cases".
Also this PR largely re-implements logic already provided within `update_submodule` on resetting the HEAD upon ending up with detached HEAD so we might endup needing to change both code places (if we remember about their existence) if we decide to change the logic.  IMHO that portion should have been extracted into a helper to be used by both implementations.

But ok -- let's proceed as is.  I left only 1 request to future proof the code a bit.

datalad/distribution/get.py
```diff
@@ -309,9 +310,99 @@ def _install_subds_from_flexible_source(ds, sm, **kwargs):
...
+                # expensive-land already anyways
+                if not sub.commit_exists(target_commit):
+                    res.update(
+                        status='error',
+                        message=(
+                            'Target commit %s does not exist in the clone, and '
+                            'a fetch that commit from origin failed',
+                            target_commit[:8]),
+                    )
+                    yield res
```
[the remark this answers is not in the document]
mih: Ah, no. But thanks for raising the issue. The fix, however, is to replace the `break` immediately below with a `continue`. Whether or not processing should stop here must be left to decide by the caller (`on_failure=stop` vs `on_failure=continue`). Given the last `break` is still inside the loop the second part of your concern should be addressed as well.

mih:
> Also this PR largely re-implements logic already provided within `update_submodule` on resetting the HEAD upon ending up with detached HEAD so we might endup needing to change both code places (if we remember about their existence) if we decide to change the logic. IMHO that portion should have been extracted into a helper to be used by both implementations.

I share the sentiment. Once this is merged, I will propose a PR to rip out large parts of the submodule functionality from `GitRepo`. At that point it will be a largely (or completely) unused part of the codebase that is less efficient to use. But we will have the chance to discuss that, when there is a concrete proposal.

datalad/distribution/get.py
mih: Thx very much, in particular for the test!

mih:
All green. Will merge.

datalad/distribution/get.py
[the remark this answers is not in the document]
yarikoptic: Wouldn't it yield res twice if the last loop iteration ends up here?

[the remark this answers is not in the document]
mih: A continue in the last iteration should still skip the rest of the loop body. How could it reach that second yield?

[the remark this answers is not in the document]
yarikoptic: .
It is too long if a spaghetti now though to see clearly what is happening. I think I miscounted indentation for line 406, and thought it to be outside of the loop body,  sorry for the noise

PR #4735, opened 2020-07-16
RF: ssh connector code uses WitlessRunner instead of Runner

mih:

PR #4860, opened 2020-09-02
DBG: SSH stalls

bpoldrack:


bpoldrack:
Notes:

- seems to happen in PY3.5 only - can't reproduce locally 

- dedicated build for `datalad.core.distributed.tests`, log level 2, PY3.5, `nose -s -v`
   https://travis-ci.org/github/datalad/datalad/jobs/723400702
   At the end postclonecfg test for ria is stalling while git-cloning.
   strange: missing debug output from witlles runner about starting the command, etc.

- Changed debug level to 8 and inserted `print` statements. Trying to figure where exactly it stalls.

bpoldrack:
After yet another iteration of adding and enabling output and narrowing it down, here is the current state of things:
As shown in https://travis-ci.org/github/datalad/datalad/builds/723505406 the hanging happens within `WitlessRunner`.
The actual call chain is `datalad clone` -> `WitlessRunner` -> `git clone` -> `datalad sshrun` -> `WitlessRunner` -> `ssh -O check`.
SSH spits out all expected output and due to current debugging we get to see it.
While in python 3.8 everything is fine and the runner returns, in python 3.5 it's hanging in `run_async_cmd` where it never finishes `await asyncio.ensure_future(transport._wait())`.

Does that ring a bell, @kyleam?

kyleam:
> Does that ring a bell, @kyleam?

No, though that's certainly not good, since it suggests the workaround for dropped output (which already made it into a release) may result in hanging on python 3.5 in some cases.

bpoldrack:
>  since it suggests the workaround for dropped output (which already made it into a release) may result in hanging on python 3.5 in some cases.

I doubt that. IIRC that very hanging happened before that workaround. But I will try whether removing and just waiting `cmd_done` would work in that particular case.

kyleam:
> I doubt that. IIRC that very hanging happened before that workaround. But I will try whether removing and just waiting `cmd_done` would work in that particular case.

Ah, that's a relief!  I thought you already narrowed it down to that statement.

bpoldrack:
>  I thought you already narrowed it down to that statement.

Only the sense that this is where it hangs. Not in terms of causation.
Original issue is PR #4735. And that's older than your fix.

bpoldrack:
Confirmed by 6fc6ed6, @kyleam. Removing that and awaiting cmd_done as before makes this the line we hang at. So - doesn't seem to be your fix. Something appears to be screwed in python 3.5. I wonder whether there's a bug in asyncio. Question would be: When exactly does it occur? This spot seems to be the only reliable case and it's not obvious what's special about it. We use that runner, we use SSH with that runner with the very same SSH commands elsewhere.
Wonder whether it has to do with "deep" process nesting.

bpoldrack:
Since we don't support python 3.5 anymore and it seems to be an issue with 3.5 only - closing.

PR #4864, opened 2020-09-03
NF: Minimalist snakemake wrapper

mih:
It injects a `datalad-get` call in front of every file property
inspection by snakemake, and thereby should ensure that the files in a
dataset that are relevant for a workflow are actually around for a valid
inspection.

This approach does not deal with the use case where a file can be
downloaded, but is actually cheaper to recompute.

To play with it, do something along the lines of

edit by @yarikoptic: added `-e` to `echo`

```
datalad create snakeoil
cd snakeoil
echo -e "\n**/Snakefile annex.largefiles=nothing" >> .gitattributes
echo ".snakemake" > .gitignore
git annex addurl --file some.txt http://example.com
cat << EOT > Snakefile
rule length:   
    input:
        "some.txt"
    output:
        "length.txt"
    shell:
        "wc < {input} > {output}"
EOT
datalad save
```

and now run

```
datalad snakemake -- -j1 length.txt
# save by hand, could later be done inside
datalad save
```

so far so good. Snakemake would have done all that on its own. But:

```
# now drop everything (or do a fresh clone)
datalad drop --nocheck .
# and redownload/compute with snakemake
datalad snakemake -- -j1 length.txt
```

### DIFF
--- a/datalad/interface/__init__.py
+++ b/datalad/interface/__init__.py
@@ -77,7 +77,9 @@ _group_misc = (
         ('datalad.interface.rerun', 'Rerun', 'rerun'),
         ('datalad.interface.run_procedure', 'RunProcedure', 'run-procedure'),
         ('datalad.distributed.export_archive_ora', 'ExportArchiveORA',
-         'export-archive-ora')
+         'export-archive-ora'),
+        ('datalad.local.snakemake', 'SnakeMake',
+         'snakemake'),
     ])
 
 _group_plumbing = (
--- a/datalad/local/snakemake.py
+++ b/datalad/local/snakemake.py
@@ -1,0 +1,112 @@
+# emacs: -*- mode: python; py-indent-offset: 4; tab-width: 4; indent-tabs-mode: nil -*-
+# ex: set sts=4 ts=4 sw=4 noet:
+# ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+#
+#   See COPYING file distributed along with the datalad package for the
+#   copyright and license terms.
+#
+# ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+"""Snakemake wrapper"""
+
+__docformat__ = 'restructuredtext'
+
+
+import logging
+from argparse import (
+    REMAINDER,
+)
+
+from datalad.interface.base import Interface
+from datalad.interface.utils import eval_results
+from datalad.interface.base import build_doc
+from datalad.support.constraints import (
+    EnsureNone,
+)
+from datalad.support.param import Parameter
+from datalad.distribution.dataset import (
+    require_dataset,
+)
+from datalad.utils import (
+    ensure_list,
+)
+
+from datalad.distribution.dataset import (
+    EnsureDataset,
+    datasetmethod,
+)
+
+lgr = logging.getLogger('datalad.local.snakemake')
+
+
+@build_doc
+class SnakeMake(Interface):
+    """
+    """
+    _params_ = dict(
+        dataset=Parameter(
+            # not really needed on the cmdline, but for PY to resolve relative
+            # paths
+            args=("-d", "--dataset"),
+            doc="""""",
+            constraints=EnsureDataset() | EnsureNone()),
+        smargs=Parameter(
+            args=("smargs",),
+            metavar='SNAKEMAKE ARGUMENTS',
+            nargs=REMAINDER,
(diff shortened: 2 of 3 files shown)


yarikoptic:
Sweet!
FWIW ```python3 -m datalad `which snakemake` -j1 length.txt``` also works although slower/with more noise I guess.

kyleam:
Interesting, thanks for sharing.

> It injects a `datalad-get` call in front of every file property
> inspection by snakemake, and thereby should ensure that the files in a
> dataset that are relevant for a workflow are actually around for a valid
> inspection.

Nice, I think that covers the main inconvenience when using snakemake with annex files.  The `get` call happens with `on_failure='ignore'`, so it doesn't really matter, but it probably makes sense to skip remote files (by looking at `.is_remote`).

Another inconvenience used to be locked output files.  That's no longer an issue as of v5.7.0, specifically ade653ee (Merged in protected_symlink_output, 2019-10-04).

Just for completeness, the only other inconvenience that comes to mind is that the "node is out of date?" check is based on mtimes.  This is a general issue with git and timestamp-based workflow managers [*].  It's not necessarily a huge deal, but it means that, even with pretty vanilla git usage, you can end up needing to re-execute more of the DAG than necessary (or end up doing some hacky, error-prone `touch -h`s).

I know you've thought a good about core vs extension, so I'm curious why you're leaning towards core here.  My initial reaction is that something like this feels better suited for an extension, in the same realm as -htcondor or -container.  And I'm waiting for all the new variants to start marching in: datalad-luigi, datalad-pydra, datalad-toil, datalad-airflow, ... :)

[*] Snakemake's FAQ mentions this, suggesting a shell command that could be use to set mtimes based on commit metadata.

yarikoptic:
> [*] Snakemake's FAQ mentions this, suggesting a shell command that could be use to set mtimes based on commit metadata.

I think it might be a nice feature (optional based on configuration) to add such functionality upon get/install/update to update mtime of the "touched" files. (more references to 3rd party projects providing that: https://stackoverflow.com/a/13284558)  It will not be 100% proof since it would quantize all mtimes into commit date, but IMHO it is not worth going further with actually storing additional metadata somewhere with mtimes per each file.

mih:
> Nice, I think that covers the main inconvenience when using snakemake with annex files. The `get` call happens with `on_failure='ignore'`, so it doesn't really matter, but it probably makes sense to skip remote files (by looking at `.is_remote`).

Good point, thx!

> Another inconvenience used to be locked output files. That's no longer an issue as of v5.7.0, specifically ade653ee (Merged in protected_symlink_output, 2019-10-04).

Yes, I was pleased to see that working already, even outside any datalad wrapping context.

> Just for completeness, the only other inconvenience that comes to mind is that the "node is out of date?" check is based on mtimes. This is a general issue with git and timestamp-based workflow managers [*]. It's not necessarily a huge deal, but it means that, even with pretty vanilla git usage, you can end up needing to re-execute more of the DAG than necessary (or end up doing some hacky, error-prone `touch -h`s).

Looking into overriding `_IOFile.mtime_local()` is still a TODO. Original behavior is

```py
os.lstat(self.file).st_mtime
```

but we could report the timestamp that git has on file. Or we could follow the path that @yarikoptic suggested and adjust mtimes in general. The former has the "advantage" of having minimal (or no) impact outside of snakemake.

> I know you've thought a good deal about core vs extension, so I'm curious why you're leaning towards core here. My initial reaction is that something like this feels better suited for an extension, in the same realm as -htcondor or -container. And I'm waiting for all the new variants to start marching in: datalad-luigi, datalad-pydra, datalad-toil, datalad-airflow, ... :)

I did and the (draft!) PR doesn't reflect my view ;-) I think this should be an extension. However, I didn't want to pay for the dev-overhead before I had an idea that things would work at all, and I had to make an actual datalad command to test the interfacing (which turned out to be 3x more complex that the actual snakemake-related aspects). And I ran out of time yesterday....

So, if this "works" (and the feedback so far seems to indicate that) this PR would be closed and come back as an extension.

mih:
The present implementation is available from https://github.com/mih/datalad-mihextras

yarikoptic:
- FTR: here is a version of that wrapper https://github.com/mih/datalad-mihextras/blob/afc44c91a3d06728ae53b14676bffd6338a08402/datalad_mihextras/snakemake.py
- I wonder ATM how it might play with `datalad-fuse` mounted partitions. Needless to say that solution would not work on HPC since unlikely FUSE is available but still - would be interesting to check

Issue #4859, opened 2020-09-02
Substantial overhead of `get` vs direct clone of a subdataset


mih:
Context is the UKB dataset with 42k subdatasets.

Cloning a particular subject dataset takes less than a second (baseline):

```
datalad clone http://ukb.ds.inm7.de/...   0.47s user 0.19s system 76% cpu 0.855 total
```

Obtaining the same dataset as a subdataset via the UKB superdataset (no subdatasets installed yet) takes more than 7x longer:

```
datalad get -n -d . sub-...  5.76s user 1.11s system 98% cpu 6.985 total
```

Clone source and target are identical, the difference in runtime is reliable.

mih:
With #4868 runtime drops from 7s to 4s:

```
datalad get -n -d . sub-3507513/  3.67s user 0.52s system 102% cpu 4.078 total
```

adswa:
I just tried this (just once, on a compute node on juseless) again, and I see 

```
time datalad get -n -d . sub-1000201   
[INFO   ] Scanning for unlocked files (this may take some time)                 
[INFO   ] Configured RIA store not found at ria+file:///ds/ukb/srv              
[INFO   ] Fetching 'http://ukb.ds.inm7.de/6fb/61dc8-d244-11ea-abd9-b4969146869c/config' 
[INFO   ] Reconfigured inm7-storage for ria+http://ukb.ds.inm7.de 
[INFO   ] Fetching 'http://ukb.ds.inm7.de/6fb/61dc8-d244-11ea-abd9-b4969146869c/config' 
[INFO   ] Configure additional publication dependency on "inm7-storage" 
install(ok): /home/adina/scratch/super2/sub-1000201 (dataset) [Installed subdataset in order to get /home/adina/scratch/super2/sub-1000201]
datalad get -n -d . sub-1000201  6.03s user 1.58s system 54% cpu 13.842 total
```

versus 

```
time datalad clone 'ria+http://ukb.ds.inm7.de#7bb480ba-d244-11ea-beb4-002590496000'
[INFO   ] Scanning for unlocked files (this may take some time)                 
[INFO   ] Configured RIA store not found at ria+file:///ds/ukb/srv              
[INFO   ] Fetching 'http://ukb.ds.inm7.de/7bb/480ba-d244-11ea-beb4-002590496000/config' 
[INFO   ] Reconfigured inm7-storage for ria+http://ukb.ds.inm7.de 
[...]

PR #4868, opened 2020-09-04
OPT: Speedup of per-subdataset "contains?" matching

mih:
Speeds up the rejection of non-containing subdatasets. For the UKB dataset with 42k subdatasets, this shaves of 1s of the previous total runtime of 2.5s.

Related to gh-4859

This also removes a backward compatibility kludge that we put in place over a year ago (`revision` instead of the present `gitshasum` property in the submodule records).

Edit: This also ups the plain reporting performance (same 42k subdatasets):

Before: `datalad subdatasets  9.49s user 3.41s system 57% cpu 22.284 total`
After: `datalad subdatasets  9.81s user 2.49s system 74% cpu 16.399 total`
and with a twist: `datalad subdatasets -d .  6.95s user 2.14s system 66% cpu 13.622 total` (see TODO)

TODO:

- [x] `GitRepo.get_submodules_()` calls `GitRepo.get_content_info()`. This might be made faster, by filtering the path-contraints with the `contains` parameter before calling any of this stack. Right now `contains` is considered last and all info is requested without any path constraint in the context of #4859 -> #5063 
- [x] Suprise differences with or without `--dataset`
    ```
    datalad subdatasets -r --contains sub-3507513  2.76s user 0.33s system 103% cpu 2.973 total                                           
    datalad subdatasets -d . -r --contains sub-3507513  1.36s user 0.21s system 104% cpu 1.507 total                                      
    ```
- [x] `update` is broken by the second commit, not clear why yet (likely it is still using the `revision` property)

### DIFF
--- a/datalad/local/subdatasets.py
+++ b/datalad/local/subdatasets.py
@@ -40,6 +40,7 @@ from datalad.utils import (
     ensure_list,
     getpwd,
     partition,
+    Path,
 )
 
 from datalad.distribution.dataset import (
@@ -224,8 +225,8 @@ class Subdatasets(Interface):
 
         # no constraints given -> query subdatasets under curdir
         if not paths and dataset is None:
-            cwd = getpwd()
-            paths = None if cwd == ds.path else [cwd]
+            cwd = Path(getpwd())
+            paths = None if cwd == ds.pathobj else [cwd]
 
         lgr.debug('Query subdatasets of %s', dataset)
         if paths is not None:
--- a/datalad/local/subdatasets.py
+++ b/datalad/local/subdatasets.py
@@ -73,20 +73,17 @@ def _parse_git_submodules(ds_pathobj, repo, paths):
             else:
                 # we had path contraints, but none matched this dataset
                 return
+    # can we use the reported as such, or do we need to recode wrt to the
+    # query context dataset?
+    recode_paths = ds_pathobj != repo.pathobj
     for props in repo.get_submodules_(paths=paths):
         path = props["path"]
         if props.get('type', None) != 'dataset':
             continue
-        if ds_pathobj != repo.pathobj:
+        if recode_paths:
             props['path'] = ds_pathobj / path.relative_to(repo.pathobj)
         else:
             props['path'] = path
-        if not path.exists() or not GitRepo.is_valid_repo(str(path)):
-            props['state'] = 'absent'
-        # TODO kill this after some time. We used to do custom things here
-        # and gitshasum was called revision. Be nice and duplicate for a bit
-        # wipe out when patience is gone
-        props['revision'] = props['gitshasum']
         yield props
 
 
@@ -302,6 +299,11 @@ def _get_submodules(ds, paths, fulfilled, recursive, recursion_limit,
                 # we are not looking for this subds, because it doesn't
                 # match the target path
                 continue
+        # the following used to be done by _parse_git_submodules()
+        # but is expensive and does not need to be done for submodules
+        # no matching `contains`
+        if not sm_path.exists() or not GitRepo.is_valid_repo(sm_path):
+            sm['state'] = 'absent'
         # do we just need this to recurse into subdatasets, or is this a
         # real results?
         to_report = paths is None \
--- a/datalad/local/subdatasets.py
+++ b/datalad/local/subdatasets.py
@@ -77,15 +77,11 @@ def _parse_git_submodules(ds_pathobj, repo, paths):
                 return
     # can we use the reported as such, or do we need to recode wrt to the
     # query context dataset?
-    recode_paths = ds_pathobj != repo.pathobj
-    for props in repo.get_submodules_(paths=paths):
-        path = props["path"]
-        if props.get('type', None) != 'dataset':
-            continue
-        if recode_paths:
-            props['path'] = ds_pathobj / path.relative_to(repo.pathobj)
-        else:
-            props['path'] = path
+    if ds_pathobj == repo.pathobj:
+        yield from repo.get_submodules_(paths=paths)
+    else:
+        for props in repo.get_submodules_(paths=paths):
+            props['path'] = ds_pathobj / paths['path'].relative_to(repo.pathobj)
         yield props
 
 
--- a/datalad/local/subdatasets.py
+++ b/datalad/local/subdatasets.py
@@ -284,13 +284,20 @@ def _get_submodules(ds, paths, fulfilled, recursive, recursion_limit,
     repo = ds.repo
     if not GitRepo.is_valid_repo(dspath):
         return
+    # expand all test cases for the contains test in the loop below
+    # leads to ~20% speedup per loop iteration of a non-match
+    expanded_contains = [
+        # convert to strings, as sm_path below will be a string and
+        # we do not want repeated type-conversion
+        [c] + list(c.parents)
+        for c in (contains if contains is not None else [])
+    ]
     # put in giant for-loop to be able to yield results before completion
     for sm in _parse_git_submodules(ds.pathobj, repo, paths):
-        contains_hits = []
+        sm_path = sm['path']
+        contains_hits = None
         if contains:
-            contains_hits = [
-                c for c in contains if sm['path'] == c or sm['path'] in c.parents
-            ]
+            contains_hits = [c[0] for c in expanded_contains if sm_path in c]
             if not contains_hits:
                 # we are not looking for this subds, because it doesn't
                 # match the target path
@@ -298,7 +305,7 @@ def _get_submodules(ds, paths, fulfilled, recursive, recursion_limit,
         # do we just need this to recurse into subdatasets, or is this a
         # real results?
         to_report = paths is None \
-            or any(p == sm['path'] or p in sm['path'].parents
+            or any(p == sm_path or p in sm_path.parents
                    for p in paths)
         if to_report and (set_property or delete_property):
             # first deletions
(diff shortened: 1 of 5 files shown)


yarikoptic:
FWIW -- no significant impact on existing benchmarks (if anything - slight tendency to slow down).  Would be nice to get a dedicated benchmark to demonstrate the effect.

mih:
> FWIW -- no significant impact on existing benchmarks (if anything - slight tendency to slow down). Would be nice to get a dedicated benchmark to demonstrate the effect.

Can you clarify what you want? From the times you see above you can decide on how big of a dataset you need to show an effect. For 1s you need 40k subdatasets, if you are satisfied with less, a few thousand should work. So let's say you want to see 100ms and do 4k subdatasets, you need to invest the ~3h to create such a dataset.

yarikoptic:
are you saying that there should be no notable (but thus also no negative) impact on a superdataset with e.g. 20 subdatasets?

Moreover we already have some benchmarks setup with some sizeable (although may be not big engouh) datasets, e.g. subclasses of https://github.com/datalad/datalad/blob/master/benchmarks/common.py#L104  - might be worth adding a  `time_contains` in one of those subclasses.

mih:
There should be no negative impact on small size datasets.

adswa:
Here is what I get (on NFS):

In master:

```
datalad subdatasets > somelog 11.98s user 1.08s system 100% cpu 13.023 total
datalad subdatasets -d . > somelog 10.05s user 1.15s system 100% cpu 11.179 total
datalad subdatasets > /dev/null  12.45s user 2.10s system 72% cpu 20.132 total   (???)
datalad subdatasets -d . > /dev/null  9.14s user 0.90s system 100% cpu 10.022 total
datalad -f '{path}' subdatasets > /dev/null  10.39s user 0.90s system 100% cpu 11.241 total
datalad -f '{path}' subdatasets -d . > /dev/null  7.19s user 0.91s system 100% cpu 8.078 total
```

On this branch:

```
SLOWER datalad subdatasets > somelog 11.06s user 1.87s system 76% cpu 16.922 total
FASTER datalad subdatasets -d . > somelog  8.62s user 0.87s system 100% cpu 9.416 total
FASTER datalad subdatasets > /dev/null   11.49s user 1.90s system 77% cpu 17.305 total
FASTER datalad subdatasets -d . > /dev/null   8.61s user 0.84s system 100% cpu 9.427 total
SLOWER datalad -f '{path}' subdatasets > /dev/null  9.09s user 1.95s system 73% cpu 15.015 total
FASTER datalad -f '{path}' subdatasets -d . > /dev/null 6.83s user 0.87s system 100% cpu 7.695 total
```

@mih's stats in comparison (on local drive):

```
datalad subdatasets > somelog  11.18s user 2.10s system 76% cpu 17.326 total
datalad subdatasets -d . > somelog  8.42s user 0.74s system 100% cpu 9.102 total
datalad subdatasets > /dev/null  11.00s user 1.17s system 101% cpu 12.001 total
datalad subdatasets -d . > /dev/null  8.69s user 2.04s system 69% cpu 15.361 total
datalad -f '{path}' subdatasets > /dev/null  8.87s user 1.01s system 101% cpu 9.714 total
datalad -f '{path}' subdatasets -d . > /dev/null  6.55s user 1.84s system 65% cpu 12.799 total
```

yarikoptic:
Wild guess re `-f '{path}'` effect - may be there is some `.repo` is involved in default full renderer?

mih:
No different conditional on `--dataset` anymore, with or without writing to a log:

```
datalad subdatasets > somelog  7.45s user 2.03s system 100% cpu 9.395 total
datalad subdatasets -d . > somelog  7.57s user 1.94s system 100% cpu 9.434 total
datalad subdatasets > /dev/null  7.48s user 1.98s system 100% cpu 9.396 total
datalad subdatasets -d . > /dev/null  7.44s user 2.10s system 100% cpu 9.477 total
```

Still a substantial impact of the result rendering:

```
datalad -f '{path}' subdatasets > /dev/null  6.19s user 2.10s system 100% cpu 8.219 total
```

But there is nothing `subdatasets`  specific about this, hence not subject of this PR.

mih:
Current benchmark runs seem to indicate no performance penalty for the small scale dataset we are using in the tests

https://github.com/datalad/datalad/pull/4868/checks?check_run_id=1286049657

datalad/local/subdatasets.py
```diff
@@ -284,21 +281,33 @@ def _get_submodules(ds, paths, fulfilled, recursive, recursion_limit,
     repo = ds.repo
     if not GitRepo.is_valid_repo(dspath):
         return
+    # expand all test cases for the contains test in the loop below
+    # leads to ~20% speedup per loop iteration of a non-match
+    expanded_contains = [
+        # convert to strings, as sm_path below will be a string and
+        # we do not want repeated type-conversion
```
kyleam, line 288: This comment isn't accurate.  `c` and `c.parents` aren't converted, and `sm_path` is a path object.

mih:
FTR: I tried to further improve the situation by using the `--contains` specification as additional query constraints, but that did not work -- made it slower. For posterity, here was my patch:

```diff
diff --git a/datalad/local/subdatasets.py b/datalad/local/subdatasets.py
index cc36a074f..5c0d63b9d 100644
--- a/datalad/local/subdatasets.py
+++ b/datalad/local/subdatasets.py
@@ -75,6 +75,14 @@ def _parse_git_submodules(ds_pathobj, repo, paths):
             else:
                 # we had path contraints, but none matched this dataset
                 return
+        else:
+            # filter out all the paths that point outside the repo. they would
+            # throw errors in GitRepo.get_submodules_(). such paths would
+            # make it here, because we mix top-level query path constrains with
+            # --contains specification in order to speed things up. If the latter
+            # are garbage, not reporting on them will be detected and impossible
+            # results are generated
+            paths = [p for p in paths if p.parts and p.parts[0] != os.pardir]
     # can we use the reported as such, or do we need to recode wrt to the
     # query context dataset?
     if ds_pathobj == repo.pathobj:
@@ -289,8 +297,12 @@ def _get_submodules(ds, paths, fulfilled, recursive, recursion_limit,
         [c] + list(c.parents)
         for c in (contains if contains is not None else [])
     ]
+    submodule_path_constraints = (paths or []) + contains if contains else paths
     # put in giant for-loop to be able to yield results before completion
-    for sm in _parse_git_submodules(ds.pathobj, repo, paths):
+    for sm in _parse_git_submodules(
+            ds.pathobj,
+            repo,
+            submodule_path_constraints):
         sm_path = sm['path']
         contains_hits = None
         if contains:
```

datalad/local/subdatasets.py
```diff
@@ -284,21 +281,33 @@ def _get_submodules(ds, paths, fulfilled, recursive, recursion_limit,
...
     # put in giant for-loop to be able to yield results before completion
     for sm in _parse_git_submodules(ds.pathobj, repo, paths):
-        contains_hits = []
+        sm_path = sm['path']
+        contains_hits = None
         if contains:
-            contains_hits = [
-                c for c in contains if sm['path'] == c or sm['path'] in c.parents
-            ]
+            contains_hits = [c[0] for c in expanded_contains if sm_path in c]
```
kyleam, line 297: Okay, so as far as I can see, the two potential sources of speed ups are that it 1) drops the repeated 'path' lookup in the `sm` dict and 2) uses list's `__contains__` over `_PathParents.__contains__` (which is `Sequence.__contains__`).

I doubt no. 1 would be measurable in this context, but I think it's still good to do.  So I'd guess the performance gains are coming from no 2.

<details>
<summary>timeit runs, .parents versus list __contains__</summary>

```sh
echo "--- .parents"
python -m timeit \
       -n 3 \
       -s 'from pathlib import Path' \
       -s 'import string' \
       'path = Path(*string.ascii_lowercase)' \
       'for _ in range(int(4e4)): "X" in path.parents'

echo "--- parents, no attribute access"
python -m timeit \
       -n 3 \
       -s 'from pathlib import Path' \
       -s 'import string' \
       -s 'path = Path(*string.ascii_lowercase)' \
       'parents = path.parents' \
       'for _ in range(int(4e4)): "X" in parents'

echo "--- parents, as list"
python -m timeit \
       -n 3 \
       -s 'from pathlib import Path' \
       -s 'import string' \
       -s 'path = Path(*string.ascii_lowercase)' \
       'parents = list(path.parents)' \
       'for _ in range(int(4e4)): "X" in parents'

# --- .parents
# 3 loops, best of 5: 1.31 sec per loop
# --- parents, no attribute access
# 3 loops, best of 5: 1.33 sec per loop
# --- parents, as list
# 3 loops, best of 5: 144 msec per loop
```

</details>

kyleam (commented):
Reading up to cbf79db8ed, these changes look good to me aside from my comment about a inaccurate comment.

datalad/local/subdatasets.py
```diff
@@ -284,21 +281,33 @@ def _get_submodules(ds, paths, fulfilled, recursive, recursion_limit,
...
             if not contains_hits:
                 # we are not looking for this subds, because it doesn't
                 # match the target path
                 continue
+        # the following used to be done by _parse_git_submodules()
+        # but is expensive and does not need to be done for submodules
+        # no matching `contains`
```
kyleam, line 304: s/no/not/

kyleam: Pushed a fixup for the failing tests, so I took care of this as well.

datalad/local/subdatasets.py
kyleam: When I was pushing fixups for the tests, I removed this comment.  If I'm mistaken, of course feel free to restore it.

mih: All good, thx!

datalad/local/subdatasets.py
mih: Thx!

PR #4869, opened 2020-09-04
ENH: Add separate  asv compare  step within Benchmarks

yarikoptic:
It always takes time and cursing to get to the actual comparison within
the long output of "Run benchmarks".  This is the simplest (and quick -
no benchmarks should be reran) way to get a dedicated action step which
would show the comparison.

PS I will cancel travis run since not relevant


yarikoptic:
so much better.  If no objections -- will merge within a day

PR #4872, opened 2020-09-05
Last step before {Git|Annex}Repo.add() removal

mih:
https://github.com/datalad/datalad/issues/3714 posed the question re the fate of these methods. This PR is supposed to show that they are obsolete in the context of datalad-core, and to explore what extension work is TODO. It looks as if only `metalad` and `crawler` need attention.

Before a merge, the first commit is likely to be dropped (see its message). The other commits will make it easier for a larger portion of the tests to run on windows. Further homogenization will hopefully ease the porting effort substantially.

Guidelines for porting other code (examples can be found in the diff of the test code in this PR):

- If possible/desirable use `Dataset`-level API, not `Repo`-level
- Replace immediate `add/commit` sequences with `Repo.save()` calls
- Replace explicit staging needs (should be very few) with `Repo._save_add()` helper

TODO

- [x] check failure of `datalad.customremotes.tests.test_archives.test_basic_scenario` (accidentally put variable name on quotes)
- [x] check failure of `datalad.support.tests.test_annexrepo.test_annex_add_no_dotfiles`
      - this is only happening with a v7 version of `git-annex-standalone`, it doesn't happen for v8 versions of standalone, nor does it happen for v8 versions of a normal `git-annex`
      - I forgot (again) why we are clinging to this old version for the tests, but given that this is a non-issue for any version that would get normally installed, I am not planning to invest time into an investigation.
- [x] windows failure of `datalad.support.tests.test_repodates.test_check_dates`


kyleam:
> check failure of `datalad.support.tests.test_annexrepo.test_annex_add_no_dotfiles`
>  - this is only happening with a v7 version of `git-annex-standalone`, it doesn't happen for v8 versions of standalone, nor does it happen for v8 versions of a normal `git-annex`

Note: The condition in e2e97e77a (TST: Avoid old v7 dotfile behavior being considered a problem, 2020-09-06) should be inverted: newer git-annex versions lack --include-dotfiles.  However, as I mention below, I think we should just drop the test.

Here is what's happening:

  * With the original `repo.add()` version of this test and a git-annex that skips dotfiles (before 8.20200226), the first `repo.add()` would _not_ add the dotfile, and the subsequent `repo.add('.', git=True")` would then add it to git (reflecting the underlying difference in dotfile treatment between git and git-annex versions before 8.20200226).  For git-annex versions after 8.20200226, this first `repo.add` call was skipped because it would add the file to the annex and mess up the next `add('.', git=True)` call.

  * With the new `_save_add()` version and a git-annex that skips dotfiles (before 8.20200226), the first `_save_add` does add the file to git-annex, unlike `add()`, because `_save_add` deliberately passes `--include-dotfiles` to get more consistent behavior between git and git-annex versions before 8.20200226.  This messes up the next `add('.', git=True)` call because the file is already tracked by git-annex.

  * With the new `_save_add()` version and a git-annex that does _not_ skip dotfiles (8.20200226 or later), the first `_save_add` call above is skipped (again, because newer git-annex version would add to the file to the annex, going against the original assumption of the test), so the test ends up working out fine.

In any case, I don't think it makes sense to keep this test around after the removal of `.add`.  (Or, assuming the first commit is dropped and a deprecated `add` stays around, this test could also be kept, sticking with `.add`).  It is testing a case where `save` intentionally diverged from `add` in order to make git-annex-add's behavior more consistent with git-add's (which is a direction that later git-annex versions ended up going anyway.) `test_save:test_save_dotfiles` already checks the dotfile behavior of the `_save_add` code path.

mih:
Thx @kyleam for the analysis. I rebased this PR on a recent master and pushed a change that implements your suggestion to maintain the problematic test as-is, until `add()` is actually removed.

Once the tests have passed. I will drop the first commit and annotate the methods to announce their corpse-in-basement status.

kyleam (commented):
Based on this...

> Before a merge, the first commit is likely to be dropped (see its message)

... and the corresponding PRs in -metalad and -crawler saying `add` is being deprecated, I think you intended to replace 60f8394193 (RF: Remove {Git|Annex}Repo.add()) with a commit that marks the methods as obsolete rather than removing them.

mih:
Yes, but there was still this travis issue that needs to be fixed, which I thought I had and now it is back.

yarikoptic:
> If possible/desirable use Dataset-level API, not Repo-level

That is what I am not sure about actually.  IMHO at least the tests of `*Repo` (and we should keep those) should use methods of `*Repo` classes and not some high level interface. 

The larger question would be -- what should interface of `Repo` classes look like? E.g. should they all be transformed to return `get_status_dict` records (ATM only `_save_add` does that) and thus possibly acquire the burden of `@eval_results` to be used as simple procedural methods; or should they still stay "closer" to git/git-annex with some "per function" native return values, while allowing for higher level api functionality to  use them efficiently.

Also in #5048 I point to the fact that `_save_add` also subject to the issue #5048 addressing in `add` -- ATM it would just let a call to it succeed without error even if there are some unsuccessful files (not added).  That boils down to either introduce `@eval_results` to every method of *Repo interfaces if we are to go after using `status_dict`s or RF so e.g. we have generator "versions" for all methods talking to annex via json, not fail right away but keep yielding until the error, etc;  (some of that discussion is also in #5048).  
So, overall this PR would not "fix" anything, but if #5048 is merged, it would introduce a regression. And IMHO we need to discuss/decide on the fate of the Repo classes APIs and behavior first.

mih:
OK, I have no interest to join the 6+ months inactivity club ;-)

I will bring this back later -- likely when there is a Repo interface in `core`.

PR #4882, opened 2020-09-06
RF: Remove AnnexRepo.get_toppath()

mih:
Had some special handling for direct mode, but now the GitRepo implementation should be sufficient.


kyleam (approved):
Nice catch.

mih:
Thx!

PR #4883, opened 2020-09-06
RF: Strip AnnexRepo._set_direct_mode() from class and make test helper

mih:
That is its only purpose. Any chance to reduce clutter in this monster class should be used.


kyleam (approved):
Sounds good.

`--color-moved` output looks as expected.

mih:
Thx @kyleam

Issue #4595, opened 2020-05-30
Strip normalize_paths()


mih:
Essentially, this functionality is from a time where the Repo classes had to be considered part of the primary API. But now there are many discrepancies between the actual API (commands) and these function.

A key issue of this approach is that its semantics are incompatible with generator-style implementations. A list of file names passed to a method will cause the all results to be assembled into a list before the method can communicate any results back to a caller (see `match_return_type` which is default-on). This causes unresponsiveness issues of various kinds and has led to alternative implementations of methods using this approach, and yielded several abandonned duplicates (see below).

Moreover, for each an every file passed to any such method (not once, but per each file!) a repositories root dir is fully resolved on the file system (this can be a slow operation, and when saving 10k files, it is done 10k times). If absolute paths are given, or paths of the style `./somepath` or `../somepath` (which is not a rare case), also the parent path of each file path is fully resolved. Taken together, this decorator performs between 1 and 2 `resolve()` calls per given file. This problem is compounded by the fact that pretty much all calling command implementations have already performed such tests and conversions.

Ping @bpoldrack if you recall a full list of the relevant aspects, please add here.

Usage is limited to
- datalad/support/annexrepo.py
- datalad/support/gitrepo.py
- datalad/support/tests/test_gitrepo.py

Outside the tests `GitRepo` uses it for

- [ ] `add()`: to be gone #4872 
- [ ] `remove()`
- [ ] `get_last_commit_hexsha()`: used in 3 locations, if adjusted, the decorator could be dropped
- [x] `_git_custom_command()`: to be gone #4850 

Usage in `AnnexRepo`:

- [ ] `get()`
- [ ] `add()`: to be gone #4872 
- [ ] `get_file_key()`: replace by get_file_props() see https://github.com/datalad/datalad/pull/6104#issuecomment-948715491
- [ ] `unlock()`
- [ ] `unannex()`
- [ ] `find()`
- [ ] `file_has_content()`: replace by get_file_props() see https://github.com/datalad/datalad/pull/6104#issuecomment-948715491 -- last user `metadata.py`, the rest are old tests.
- [ ] `is_under_annex()`: replace by get_file_props() see https://github.com/datalad/datalad/pull/6104#issuecomment-948715491 -- last users `download_url`, `add_archive_content`, and `metadata.py`
- [ ] `drop()`
- [ ] `whereis()`
- [ ] `info()`
- [ ] `remove()`
- [ ] `is_available()`
- [x] `_annex_custom_command()`
[...]

PR #4887, opened 2020-09-06
TMP: Explore what actually needs AnnexRepo.get() and what of it

mih:
Ping gh-#4595


yarikoptic:
should be kept open or closed @mih?

mih:
No need for this anymore. Sorry to have kept it open for so long.

Issue #4890, opened 2020-09-06
Use Repo.config.runner for Repo, instead of maintaining another one?


mih:
#4889 got me thinking: each repo instance has a runner, each repo instance has a config manager which also has a runner. Right now I cannot think of a reason why this is needed or beneficial.

Should we make `GitRepo._git_runner = GitRepo.config._runner`? Or rather `GitRepo.config._runner = GitRepo._git_runner`?

PR #4891, opened 2020-09-06
OPT: Re-use ConfigManager._runner for an associated Repo instance

mih:
I cannot think of a reason why the (previously two) would need to be
configured differently (which could have been the case before). Having
only one runner prevents this possibility -- for better or worse.

Fixes gh-4890


yarikoptic (approved):

mih:
Thx @yarikoptic !

Issue #4862, opened 2020-09-03
3 fresh test failures in master on a build against current git-annex snapshot


yarikoptic:
I guess the result of https://github.com/datalad/datalad/pull/4838

see https://github.com/datalad/datalad-extensions/runs/1064624383?check_suite_focus=true
<details>
<summary>FAIL: datalad.support.tests.test_gitrepo.test_GitRepo_flyweight</summary> 

```
======================================================================
FAIL: datalad.support.tests.test_gitrepo.test_GitRepo_flyweight
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/site-packages/datalad/tests/utils.py", line 731, in _wrap_with_tempfile
    return t(*(arg + (filename,)), **kw)
  File "/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/site-packages/datalad/tests/utils.py", line 731, in _wrap_with_tempfile
    return t(*(arg + (filename,)), **kw)
  File "/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/site-packages/datalad/support/tests/test_gitrepo.py", line 1180, in test_GitRepo_flyweight
    assert_equal(1, sys.getrefcount(repo1) - 1)
AssertionError: 1 != 6
-------------------- >> begin captured logging << --------------------
asyncio: DEBUG: Using selector: EpollSelector
asyncio: DEBUG: Using selector: EpollSelector
--------------------- >> end captured logging << ---------------------
```
</details>

<details>
<summary>FAIL: datalad.support.tests.test_annexrepo.test_AnnexRepo_flyweight</summary> 

```
======================================================================
FAIL: datalad.support.tests.test_annexrepo.test_AnnexRepo_flyweight
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/site-packages/datalad/tests/utils.py", line 731, in _wrap_with_tempfile
    return t(*(arg + (filename,)), **kw)
[...]

PR #4897, opened 2020-09-07
TST: Refcount is unreliable ATM

bpoldrack:
In light of (unresolved) issue #4862, it seems necessary to adapt the tests for now.
Since the issue is neither happening for me locally nor on Travis, I need you, @kyleam and @mih to confirm, that these tests are passing for you with this patch.

If so, I'll need to dig out #4849 in addition (and find a better solution for that one).

### DIFF
--- a/datalad/distribution/tests/test_dataset.py
+++ b/datalad/distribution/tests/test_dataset.py
@@ -353,10 +353,24 @@ def test_Dataset_flyweight(path1, path2):
 
     ds1 = Dataset(path1)
     assert_is_instance(ds1, Dataset)
-    ds1.create()
-
     # Don't create circular references or anything similar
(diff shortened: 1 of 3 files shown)


kyleam:
> Since the issue is neither happening for me locally nor on Travis, I need you, @kyleam and @mih to confirm, that these tests are passing for you with this patch.

With these changes, all three tests listed in that issue pass for me (python 3.7.3).  (As I mentioned in that issue, I wasn't able to trigger the `test_Dataset_flyweight` to begin with.)

mih:
I can confirm that all three tests pass for me with this PR. Thx

mih:
The singularity failure is unrelated and happens in other PRs too. I will merge this now, incremental improvements can come in subsequent PRs. Thx @bpoldrack

PR #4898, opened 2020-09-08
BF(TST): skip some tests which would fail to Encode a unicode (or just read a unicode file?) with python 3.6

yarikoptic:
fails detected while building for ubuntu 18.04 with python 3.6 but also replicate in conda env with `LANG=C`.  I think that later pythons are Ok. I might add more skips later on if we agree that such ugly workaround is ok

### DIFF
--- a/datalad/distribution/tests/test_create_sibling.py
+++ b/datalad/distribution/tests/test_create_sibling.py
@@ -13,6 +13,8 @@ import os
 from os import chmod
 import stat
 import re
+import sys
+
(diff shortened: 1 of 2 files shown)


yarikoptic:
I bet I will recall about this one some time next time I build a package... I have merged current maint for a face lift, and will look into having it merged unless there are objections voiced.

yarikoptic:
Get, and I did waste time redoing this fire Debian package... Merging

Issue #4893, opened 2020-09-06
install --reckless ephemeral creates invalid link to remote annex if source is a bare repo


dickscheid:
- please see my latest post below, problem now clearly understood but it needs a fix -  

#### What is the problem?

I want to clone a repo with some Tiff images as an input-subdataset, from a RIA store that sits on the same filesystem. Since the tiff images are very large, I want to do a reckless-ephemeral install that points to the original data blobs in the RIA store. This works well.

Since my tiff library  doesn't like the hardlinks, I want to unlock files first. If I run "datalad unlock file.tif" however, I receive an error:

```
CommandError: 'git-annex find -c annex.dotfiles=true -c annex.retry=3 --json --json-error-messages --copies 0 -- <my filename>.tif' failed with exitcode 1 under <my directory>
git-annex: .git/annex: createDirectory: already exists (File exists)
```

If I do the exact same procedure on the same system, but install the same subdataset from a non-RIA store, it works. 

If I do not use --reckless ephemeral, the "datalad get" + "datalad unlock" is much slower than the ephemeral variant.

#### What steps will reproduce the problem?

1. Create a repository A with some files.
2. Create a RIA sibling of A, A-RIA on the same filesystem.
3. Install a clone of A-RIA as a subdataset in a new repository on the same filesysetm, using --reckless ephemeral.
4. Try to `dataset unlock` a file - the error occurs.
5. Repeat 3, but use A as the origina (not A-RIA).
6. Try to `dataset unlock` a file - works.

#### What version of DataLad are you using (run `datalad --version`)? On what operating system (consider running `datalad wtf`)?

0.13.3

dickscheid:
Update on this issue: The problem with my tiff library were not hardlinks, but rather the way that `--reckless ephemeral` symlinks the annex. I will dig deeper into the problem, but the different behavior for unlocking described above ist still an open question to me.

dickscheid:
I think I figured the problem myself: 

The `datalad install --reckless ephemeral` models the new clone's `.git/annex` as a symlink to the origin's / remote's annex. When I clone from a RIA store, thsi remote is a bare repo. The install command however creates a symlink to the remote's `.git/annex`, which does not exist.

Suggested solution is that `datalad install --reckless ephemeral` should detect wether the source repo is a bare repo, and in that case craete the symbolic link to `annex/` instead `.git/annex`.

mih:
[...]

PR #4899, opened 2020-09-08
BF+TST: ephemeral clone to account for bare repo

bpoldrack:
Somehow we managed to ignore the intial usecase leading to the development of RIA stores and ephemeral clones! *facepalm*

This fixes `clone` to symlink an ephemeral clone's annex correctly, if `origin` is a bare repo.
Tests are enhanced in two ways:

1. Test merely the right link in `test_clone.py` as this is all `clone` itself is concerned with.
2. Add a new test file for usecases/workflows involving RIA stores and add the first one, which is an enhancement of what @dickscheid described in #4893 in that it writes to the unlocked file in the clone and propagates everything back to origin via the store.

Ping @mih

### DIFF
--- a/datalad/core/distributed/clone.py
+++ b/datalad/core/distributed/clone.py
@@ -799,7 +799,21 @@ def postclonecfg_annexdataset(ds, reckless, description=None):
                     # Deal with file:// scheme URLs as well as plain paths.
                     # If origin isn't local, we have nothing to do.
                     origin_git_path = Path(RI(origin_annex_url).localpath)
-                    if origin_git_path.name != '.git':
+
+                    # we are local; check for a bare repo first to not mess w/
+                    # the path
+                    # Note, that w/o support for bare repos in GitRepo we also
+                    # can't use ConfigManager ATM.
+                    from datalad.cmd import GitWitlessRunner, StdOutErrCapture
+                    gc_response = GitWitlessRunner(
+                        cwd=origin_git_path,
+                    ).run(['git', 'config', '--local', '--get', 'core.bare'],
+                          protocol=StdOutErrCapture,
+                          encoding='utf-8')
+                    if gc_response['stdout'].lower().strip() == 'true':
+                        # origin is a bare repo -> use path as is
+                        pass
+                    elif origin_git_path.name != '.git':
                         origin_git_path /= '.git'
                 except ValueError:
                     # Note, that accessing localpath on a non-local RI throws
--- a/datalad/core/distributed/clone.py
+++ b/datalad/core/distributed/clone.py
@@ -808,8 +808,7 @@ def postclonecfg_annexdataset(ds, reckless, description=None):
                     gc_response = GitWitlessRunner(
                         cwd=origin_git_path,
                     ).run(['git', 'config', '--local', '--get', 'core.bare'],
-                          protocol=StdOutErrCapture,
-                          encoding='utf-8')
+                          protocol=StdOutErrCapture)
                     if gc_response['stdout'].lower().strip() == 'true':
                         # origin is a bare repo -> use path as is
                         pass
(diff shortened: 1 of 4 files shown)


datalad/core/distributed/clone.py
```diff
@@ -799,7 +799,20 @@ def postclonecfg_annexdataset(ds, reckless, description=None):
                     # Deal with file:// scheme URLs as well as plain paths.
                     # If origin isn't local, we have nothing to do.
                     origin_git_path = Path(RI(origin_annex_url).localpath)
-                    if origin_git_path.name != '.git':
+
+                    # we are local; check for a bare repo first to not mess w/
+                    # the path
+                    # Note, that w/o support for bare repos in GitRepo we also
+                    # can't use ConfigManager ATM.
+                    from datalad.cmd import GitWitlessRunner, StdOutErrCapture
```
mih, line 807: That is not true, and tested to be such: https://github.com/datalad/datalad/blob/6cbb11a28d28fa1e1eb2eb5b40722dcb3873d9d1/datalad/tests/test_config.py#L524

datalad/core/distributed/tests/test_clone.py
```diff
@@ -850,6 +850,8 @@ def _test_ria_postclonecfg(url, dsid, clone_path):
     assert_result_count(res, 2)
 
 
+
+
```
mih, line 854: I guess those two could be removed again

datalad/interface/common_opts.py
```diff
@@ -151,7 +151,9 @@
     w/o git-annex being aware of it. In case of a change in origin you need to
     update the clone before you're able to save new content on your end.
     Alternative to 'auto' when hardlinks are not an option, or number of consumed
-    inodes needs to be minimized.
+    inodes needs to be minimized. Please note, that you might run into issues
+    when using this mode to link a bare (or any tuned annex) and a non-bare 
+    repository that way, since their annex object trees are organized differently.
```
mih, line 156: Maybe: Using this mode with bare repositories (outside of a RIA store) or tuned repositories is not supported, because their annex object trees are organized  differently.

It will not work. There is no uncertainty

mih (changes requested):
Thx, this will feel a painfull void. Left a few comment

datalad/core/distributed/clone.py
bpoldrack: It is true. `GitRepo` can only "handle" a bare repo, that it is actively creating. It can't be instantiated on an existing one. And `ConfigManager` requires a repo instance.

datalad/interface/common_opts.py
bpoldrack: Don't think so. First of all: It is supported in that `--reckless=ephemeral` will do it's thing. Second, there is uncertainty, since you can have a tuned repo, using dirhashlower. It will work in that case. Third: At some point I tested, that a bare repo can actually deal with dirhashmixed keys in its object tree. It just wouldn't put it in there itself. So, if you use an ephemeral clone to write to it you can also read those keys from another clone. But the linked ephemeral clone can't handle existing dirhashlower objects in the bare repo.
All depends on the setup. Can work or not.

bpoldrack: But will add "outside RIA store" as a hint.

datalad/core/distributed/clone.py
bpoldrack: Just looked into fixing `GitRepo`/`ConfigManager` for it, since it seemed to work for creation. Conclusion: It doesn't the test you pointed to is misleading. Doesn't actually read a bare repo's config. Instead it pretends to by first writing to it and then read out that value (into and from the wrong location, though).

I'll give it shot, but not within this PR.

bpoldrack: See PR #4911

bpoldrack:
Failure in datalad-container is known and unrelated. And I'm pretty damn sure the one in crippledFS is unrelated, too (added it to the list).

mih (approved):
Two exclamation marks?! I would still say that revealing the importance of dirhash_mixed or dirhash method equivalence is sufficient and appropriate, but this will do too.

PR #4900, opened 2020-09-08
RF: Make archive (de)compression use WitlessRunner

mih:
Another small step.

Went for a dedicated subshell to implement the pipe, rather than an exploration how to connect two runners. Not worse than what it was before.

### DIFF
--- a/datalad/support/archive_utils_7z.py
+++ b/datalad/support/archive_utils_7z.py
@@ -25,7 +25,10 @@ from datalad.utils import (
 import logging
 lgr = logging.getLogger('datalad.support.archive_utils_7z')
 
-from datalad.cmd import Runner
+from datalad.cmd import (
+    WitlessRunner as Runner,
+    StdOutErrCapture,
+)
 
 
 def _normalize_fname_suffixes(suffixes):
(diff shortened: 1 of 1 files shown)


datalad/support/archive_utils_7z.py
```diff
@@ -52,13 +55,14 @@ def decompress_file(archive, dir_):
...
-        cmd = '7z x {} -so | 7z x -si -ttar'.format(quote_cmdlinearg(archive))
+        cmd = [
+            'sh', '-c',
+            '7z x {} -so | 7z x -si -ttar'.format(quote_cmdlinearg(archive)),
+        ]
     else:
         # fire and forget
         cmd = ['7z', 'x', archive]
-    runner.run(cmd)
+    runner.run(cmd, protocol=StdOutErrCapture)
```
kyleam, line 65: Given that this function and the one below aren't returning the output, would it be better to use the `KillOutput` protocol?

mih: Oh. You are absolutely correct!

mih: Done

mih:
Thanks for the review @kyleam. I made the switch to `KillOutput` and will merge now. All test failures are unrelated.

bpoldrack:
Hm. `test_archives` just started failing in master in PY3.7: https://travis-ci.org/github/datalad/datalad/builds/725525494

I'm aware that it passed here, but since it only started *now*, seems related to this merge.

PR #4877, opened 2020-09-05
ENH: Add support of .tgz archives

mih:
I am not aware of any other common extensions that indicate compressed tarballs in a single suffix. There is now a dedicated function to normalize such suffixes to make them detectable with the existing approach -- without affecting the actual filename.

Fixes #4875

Issue #4875, opened 2020-09-05
BF: 7z based re-implementation of archives handling doesn't handle .tgz (and others) "fully" (only get .tar)


yarikoptic:
Extracts only .tar, thus requiring a 2nd step.

Original discovery: https://github.com/datalad-handbook/book/pull/569/files/639b8c1602542d9d4bc8975414e82967162c100c#diff-950b7583ea5bdc7e6f1ab0492c8b3eea

in particular https://github.com/datalad/datalad/blob/HEAD/datalad/support/archive_utils_7z.py#L43 : it doesn't know about `.tg?`.  and `7z x` doesn't care about "composition" -- one step extraction of such archives seems to be a feature asked for a decade: https://sourceforge.net/p/sevenzip/discussion/45797/thread/5eb1b8d5/?page=3  .

Meanwhile, the undesired behavior could be overcome with `-c datalad.runtime.use-patool=1` and that one (after patool step, so most likely unrelated to 7z vs patool) manages to trigger the workaround for stuck git annex batch processes!, heh heh (might be of advantage to be able to pin point underlying issue in such as case vs strange order of tests):

```shell
(git)smaug:/tmp/testds[master]data/raw
$> datalad -c datalad.runtime.use-patool=1 add-archive-content imagenette2-160.tgz
[INFO   ] Adding content of the archive imagenette2-160.tgz into annex AnnexRepo(/tmp/testds) 
[INFO   ] Finished adding imagenette2-160.tgz: Files processed: 13394, renamed: 13394, +annex: 13394 
[WARNING] Caught Command '['git', 'annex', 'addurl', '--with-files', '--json', '--json-error-messages', '--batch']' timed out after 3.0 seconds [subprocess.py:_wait:1616] on trial #1. Sleeping 1.000000 and retrying 
[WARNING] Caught Command '['git', 'annex', 'addurl', '--with-files', '--json', '--json-error-messages', '--batch']' timed out after 3.0 seconds [subprocess.py:_wait:1616] on trial #2. Sleeping 1.000000 and retrying 
datalad -c datalad.runtime.use-patool=1 add-archive-content   94.95s user 36.41s system 103% cpu 2:06.96 total

$> git annex version 2>&1 | head -n 1
git-annex version: 7.20190819+git2-g908476a9b-1~ndall+1

$> datalad --version                 
datalad 0.13.3.dev191

$> cd ~datalad/datalad-master 


(git)smaug:~datalad/datalad-master[master]
$> git describe
[...]

PR #4901, opened 2020-09-08
BF: check that files are extracted after normalizing suffixes

adswa:
I think this was a missing piece from #4877. With this change, the problem I described in https://github.com/datalad/datalad/issues/4875#issuecomment-688767763 is fixed.


adswa:
Oh, nevermind - I think I have to take this back, it doesn't fix anything. I will follow up with more observations on this in the issue.

PR #4909, opened 2020-09-08
RF(PY3): addurls - use "yield from"

yarikoptic:
Just `addurls` since I was there at the moment, but I guess there could be many other spots which could benefit from a similar RF

### DIFF
--- a/datalad/plugin/addurls.py
+++ b/datalad/plugin/addurls.py
@@ -426,8 +426,7 @@ def sort_paths(paths):
     def level_and_name(p):
         return p.count(os.path.sep), p
 
-    for path in sorted(paths, key=level_and_name):
-        yield path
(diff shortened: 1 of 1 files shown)


kyleam (commented):
Is this sort of churn really worth it?

yarikoptic:
not sure what you mean... it is a Python3 construct we have not used since were supporting python2.  
Looking forward I think we should prefer code to use it instead of explicit `for r in ...: yield r`.  Either it is "worth a PR", I don't see why not

kyleam:
Looking forward is fine.  We've already starting using this in new code.  But what does going through old code, especially submitting a PR for a module here and a module there, gain us?  It's just needless churn that takes up developer, reviewer, and CI time.

kyleam (approved):
meta-comments aside...

PR #4911, opened 2020-09-09
Towards (minimal) support for bare repos

bpoldrack:
This PR aims to add minimal support for bare repositories to `GitRepo`. We should now be able to properly create one (not just pass `--bare` to `git-init`, which is insufficient) and instantiate on an existing bare repo. Added `bare` property, so `GitRepo` is aware and we can enhance it to react to that knowledge from within its methods. That property does not only check the config but also does an additional consistency check on `pathobj` and `dot_git` (should be cheap enough, I think).

Please have a look, @datalad/developers

### DIFF
--- a/datalad/config.py
+++ b/datalad/config.py
@@ -232,10 +232,16 @@ class ConfigManager(object):
         self._src_mode = source
         run_kwargs = dict()
         if dataset is not None:
-            # make sure we run the git config calls in the dataset
-            # to pick up the right config files
-            run_kwargs['cwd'] = dataset.path
-        self._runner = GitWitlessRunner(**run_kwargs)
+            if hasattr(dataset, '_git_runner'):
+                self._runner = dataset._git_runner
+            elif dataset.repo:
+                self._runner = dataset.repo._git_runner
+            else:
+                # make sure we run the git config calls in the dataset
+                # to pick up the right config files
+                run_kwargs['cwd'] = dataset.path
+        else:
+            self._runner = GitWitlessRunner(**run_kwargs)
 
         self.reload(force=True)
 
--- a/datalad/config.py
+++ b/datalad/config.py
@@ -231,6 +231,7 @@ class ConfigManager(object):
 
         self._src_mode = source
         run_kwargs = dict()
+        self._runner = None
         if dataset is not None:
             if hasattr(dataset, '_git_runner'):
                 self._runner = dataset._git_runner
@@ -240,7 +241,7 @@ class ConfigManager(object):
                 # make sure we run the git config calls in the dataset
                 # to pick up the right config files
                 run_kwargs['cwd'] = dataset.path
-        else:
+        if self._runner is None:
             self._runner = GitWitlessRunner(**run_kwargs)
 
         self.reload(force=True)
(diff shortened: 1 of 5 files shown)


datalad/support/gitrepo.py
```diff
@@ -886,17 +886,29 @@ def __init__(self, path, url=None, runner=None, create=True,
...
         self.pathobj = ut.Path(self.path)
         self.dot_git = self._get_dot_git(self.pathobj, ok_missing=True)
+
+        # We'll need some awareness of whether or not this is a bare repo, if we
+        # are to support it beyond instantiation
+        # TODO: We should prob. double check config for core.bare=true! However,
+        #       at this point it would sabotage the lazy loading of
+        #       ConfigManager. May be `bare` property, that would check config +
+        #       path == dot_git for consistency? Once at it: To what extend can
+        #       this (or the location of .git in general) change during runtime?
```
bpoldrack, line 909: Just noted, that someone merged the destruction of lazy loading `ConfigManager` already ... Hmpf.

datalad/config.py
```diff
@@ -232,10 +232,16 @@ def __init__(self, dataset=None, overrides=None, source='any'):
...
-            run_kwargs['cwd'] = dataset.path
-        self._runner = GitWitlessRunner(**run_kwargs)
+            if hasattr(dataset, '_git_runner'):
+                self._runner = dataset._git_runner
+            elif dataset.repo:
+                self._runner = dataset.repo._git_runner
+            else:
+                # make sure we run the git config calls in the dataset
+                # to pick up the right config files
+                run_kwargs['cwd'] = dataset.path
```
kyleam, line 242: I think I'm missing something because this code seems to be covered and didn't error, but doesn't this arm leave `_runner` undefined?

kyleam (commented):
Thanks for working on this.

> Subject: [PATCH 2/7] BF+TST: ephemeral clone to account for bare repo

This patch is in master now (via gh-4899), and that series is presumably the source of the current conflicts.

datalad/config.py
bpoldrack: I think you're right, @kyleam. Thanks for spotting that.
It might not error, because this only happens, when we instantiate a `Dataset.config` for a dataset that wasn't created yet (has no `.repo`), so we probably never actually call anything on it in the tests.

bpoldrack: This arm should do the same thing as the outer `else`.

datalad/support/gitrepo.py
```diff
@@ -974,6 +976,20 @@ def __init__(self, path, url=None, runner=None, create=True,
...
 
+
+    @property
+    def bare(self):
+        if self.config.getbool("core", "bare") and \
+                self.pathobj == self.dot_git:
+            return True
+        elif not self.config.getbool("core", "bare") and \
+                not self.pathobj == self.dot_git:
+            return False
```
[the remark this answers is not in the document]
bpoldrack: Yes, in those checking spots anything that isn't cached would hurt us wrt performance.

mih:
@bpoldrack what is the state of this PR? The fix for #4893 is still urgent, but hasn't made it into the mainline yet, it seems. Can it be pulled from this PR?

kyleam:
> The fix for #4893 is still urgent, but hasn't made it into the mainline yet, it seems

Didn't that come with gh-4899?

kyleam (approved):

PR #4914, opened 2020-09-09
TEMP: try recent debian devel master

yarikoptic:
ATM only for nfs run -- to see if stalls


yarikoptic:
ha -- this one passes!  Let me redo a full run

PR #4915, opened 2020-09-09
TEMP: try recent debian devel master

yarikoptic:


yarikoptic:
damn, it did stall (this run was without `-s -v` to nose BTW, unlike one in #4914) for NFS.

Also https://github.com/datalad/datalad/issues/4916 -- the only failed in 3 matrix runs but didn't reproduce locally

Some ran out of time (50 min)

yarikoptic:
earth was flat back then

Issue #4919, opened 2020-09-13
ConfigManager.get() not aligned with `git config <var>` behavior


mih:
Problem: A dataset has two (identical) pushurls configured (for no good reason).

```
% cat /ds/hcp/super/HCP1200/108121/MNINonLinear/.git/config 
...
[remote "origin"]
        url = http://store.datalad.org/ceb/604be-2baf-11ea-ae67-002590496000
        fetch = +refs/heads/*:refs/remotes/origin/*
        annex-ignore = true
        pushurl = falkor.dartmouth.edu:/srv/store.datalad.org/www/ceb/604be-2baf-11ea-ae67-002590496000
        pushurl = falkor.dartmouth.edu:/srv/store.datalad.org/www/ceb/604be-2baf-11ea-ae67-002590496000
...
```

However, Git is unimpaired by that, because it reports the last values by default, whereas our `get()` always reports all values:

```
% git -C /ds/hcp/super/HCP1200/108121/MNINonLinear config remote.origin.pushurl 
falkor.dartmouth.edu:/srv/store.datalad.org/www/ceb/604be-2baf-11ea-ae67-002590496000
```

I cannot show that, because it crashes DataLad:

```
% python -c 'from datalad.api import Dataset; Dataset("/ds/hcp/super/HCP1200/108121/MNINonLinear").config'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/mih/code/datalad/datalad/distribution/dataset.py", line 343, in config
    repo = self.repo
  File "/home/mih/code/datalad/datalad/distribution/dataset.py", line 289, in repo
    self._repo = repo_from_path(self._path)
  File "/home/mih/code/datalad/datalad/core/local/repo.py", line 61, in repo_from_path
    repo = cls(path, create=False, **kw)
  File "/home/mih/code/datalad/datalad/support/repo.py", line 151, in __call__
    instance = type.__call__(cls, *new_args, **new_kwargs)
  File "/home/mih/code/datalad/datalad/support/annexrepo.py", line 260, in __init__
    self._set_shared_connection(r, url)
  File "/home/mih/code/datalad/datalad/support/annexrepo.py", line 428, in _set_shared_connection
    if is_ssh(url):
  File "/home/mih/code/datalad/datalad/support/network.py", line 923, in is_ssh
    _ri = RI(ri) if not isinstance(ri, RI) else ri
  File "/home/mih/code/datalad/datalad/support/network.py", line 419, in __new__
[...]

Issue #2628, opened 2018-06-07
config: Parser respresents multiple entries as tuples, which callers don't handle


kyleam:
If a value occurs multiple times in the git configuration, it's represented as a tuple.

https://github.com/datalad/datalad/blob/1bdc37a5ec2e8d63bba541e7c9c0778b4559fd89/datalad/config.py#L79-L85

AFAICS most, if not all, callers of `config.get` and friends do not handle this multi-value case. This means that I can't set a value in `.gitconfig` and then override it in the repo's `.git/config` (e.g., if I want to enable `datalad.run.record-sidecar` in general, but disable it in a specific repo).

---

```bash
#!/bin/sh
set -e

mkdir -p /tmp/dl/
export HOME=/tmp/dl/

git config --global user.name "blah"
git config --global user.email "blah@blah"

repo=/tmp/dl/config-test

test -e $repo && datalad remove --nocheck $repo

datalad create $repo

cd $repo
git config --global datalad.run.record-sidecar true
git config datalad.run.record-sidecar false

pycmd=$(tr "\n" " " <<-EOF
        from datalad.api import Dataset;
        d = Dataset('.');
        print(d.config.get('datalad.run.record-sidecar'));
        print(d.config.getbool('datalad', 'run.record-sidecar'))
EOF
)

python -c "$pycmd"
```

```
remove(ok): /tmp/dl/config-test (dataset)
[INFO   ] Creating a new annex repo at /tmp/dl/config-test
create(ok): /tmp/dl/config-test (dataset)
('true', 'false')
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/kyle/src/python/datalad/datalad/config.py", line 473, in getbool
    return anything2bool(val)
  File "/home/kyle/src/python/datalad/datalad/config.py", line 113, in anything2bool
    % repr(val))
TypeError: Got value ('true', 'false') which could not be interpreted as a boolean
```

PR #4924, opened 2020-09-16
RF: Change ConfigManager.get() behavior for multiple identical key specs

mih:
Before this change, the values of all specifications of an identical key
were returned as a tuple. Now only the last value is returned, and
`get_all=True` has to be set to get the previous behavior.

Rationale: That a single value is returned is a sensible default
expectation. Valid use cases for multiple values are few, and require
custom handling in consuming code. The previous behavior would require
protection against the multi-value-return possibility all throughout
the code base. This switch is disruptive, but makes overall operation
more robust.

Fixes gh-2628
Fixes gh-4919

### DIFF
--- a/datalad/config.py
+++ b/datalad/config.py
@@ -504,9 +504,27 @@ class ConfigManager(object):
         return self._merged_store.keys()
 
     # XXX should this be *args?
-    def get(self, key, default=None):
-        """D.get(k[,d]) -> D[k] if k in D, else d.  d defaults to None."""
-        return self._merged_store.get(key, default)
+    def get(self, key, default=None, get_all=False):
+        """D.get(k[,d]) -> D[k] if k in D, else d.  d defaults to None.
+
+        Parameters
+        ----------
+        default : optional
+          Value to return when key is not present. `None` by default.
+        get_all : bool, optional
+          If True, return all values of multiple identical configuration keys.
+          By default only the last specified value is returned.
+        """
+        try:
+            val = self[key]
+            if get_all or not isinstance(val, tuple):
+                return val
+            else:
+                return val[-1]
+        except KeyError:
+            # return as-is, default could be a tuple, hence do not subject to
+            # get_all processing
+            return default
 
     def get_from_source(self, source, key, default=None):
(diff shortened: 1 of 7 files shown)


kyleam (commented):
> Rationale: That a single value is returned is a sensible default
> expectation. Valid use cases for multiple values are few, and require
> custom handling in consuming code. The previous behavior would require
> protection against the multi-value-return possibility all throughout
> the code base. This switch is disruptive, but makes overall operation
> more robust.

That matches my view.  I think the number of potential bugs/errors this squashes outweighs any disruption.  And quickly checking the extensions, I didn't spot any callers that accounted for a possible tuple return value, though it'd be good to do a more thorough check.  (Perhaps the -neuroimaging test failure is related, though I didn't look closely yet.)

I've pushed some additional changes (most notably, making `getbool` go through `get`).  Of course feel free to drop anything you don't agree with.

The test failures on Travis seem to be in `test_create_github.test_integration1_datalad_tester_org`.  I'm not sure if that's related.

mih:
Thanks for the additional changes @kyleam I'd prefer to keep them all. I will try to have a look at the test failures later today.

kyleam:
> And indeed, we explicitly rely on multi-value returns for configured GitHub access tokens. Pushed a fix.

Thanks.  The only failure now is the unrelated -container issue.

@bpoldrack Since we've discussed the single/multi-value config behavior in the past: does this direction look okay to you?

bpoldrack:
> @bpoldrack Since we've discussed the single/multi-value config behavior in the past: does this direction look okay to you?

Yes, it does. In most cases only the "effective" config is relevant indeed and wherever we want to account for several we do know that this is spot where we need `get_all`. There is one piece still missing, I guess: We cannot have a "multi-config" that still allows to be overwritten by more specific config files (like: `~/.gitconfig` defines A and B, but I want to overwrite in `.git/config` by C and D). However, this doesn't become harder by this PR, so - perfectly fine with it.

mih:
Ok, seems we're on the some page. Thx much!

PR #4925, opened 2020-09-16
ENH: log up to 100 of last lines in stderr (if log outputs) for batched processe

yarikoptic:
I am chasing some bug in datalad/git-annex where batched addurls
eventually crashes with

	datalad.cmd     [DEBUG  ] Closing stdin of <subprocess.Popen object at 0x7fa14a7cb940> and waiting process to finish
	datalad.utils   [WARNING] Caught Command '['git', 'annex', 'addurl', '--fast', '--with-files', '--json', '--json-error-messages', '--batch']' timed out after 3.0 seconds [subprocess.py:_wait:1616] on trial #1. Sleeping 1.000000 and retrying
	datalad.utils   [WARNING] Caught Command '['git', 'annex', 'addurl', '--fast', '--with-files', '--json', '--json-error-messages', '--batch']' timed out after 3.0 seconds [subprocess.py:_wait:1616] on trial #2. Sleeping 1.000000 and retrying
	datalad.cmd     [WARNING] Batched process <subprocess.Popen object at 0x7fa14a7cb940> did not finish, abandoning it without killing it
	Traceback (most recent call last):
	  File "datalad-nda/scripts/datalad-nda", line 416, in <module>
		main()
	  File "/usr/lib/python3/dist-packages/click/core.py", line 764, in __call__
		return self.main(*args, **kwargs)
	  File "/usr/lib/python3/dist-packages/click/core.py", line 717, in main
		rv = self.invoke(ctx)
	  File "/usr/lib/python3/dist-packages/click/core.py", line 1137, in invoke
		return _process_result(sub_ctx.command.invoke(sub_ctx))
	  File "/usr/lib/python3/dist-packages/click/core.py", line 956, in invoke
		return ctx.invoke(self.callback, **ctx.params)
	  File "/usr/lib/python3/dist-packages/click/core.py", line 555, in invoke
		return callback(*args, **kwargs)
	  File "datalad-nda/scripts/datalad-nda", line 236, in add2datalad
		on_failure="stop",
	  File "/mnt/scrap/tmp/abcd/datalad/datalad/distribution/dataset.py", line 503, in apply_func
		return f(**kwargs)
	  File "/mnt/scrap/tmp/abcd/datalad/datalad/interface/utils.py", line 481, in eval_func
		return return_func(generator_func)(*args, **kwargs)
	  File "/mnt/scrap/tmp/abcd/datalad/datalad/interface/utils.py", line 469, in return_func
		results = list(results)
[...]

PR #4926, opened 2020-09-16
BF: custom remotes - replace os.linesep with an explicit \n in messages to annex

yarikoptic:
Otherwise it make annex puke since it does not expect multiple lines in a
single command message/content.

E.g. seeing (thanks to #4925 )

	datalad.cmd     [Level 5] stderr| datalad.customremotes [Level 4] Sending 'DEBUG Failed to check url s3://NDAR_Central_2/submission_23229/sub-XXX/ses-baselineYear1Arm1/fmap/sub-XXX_ses-baselineYear1Arm1_dir-AP_run-01_epi.nii.gz: Cannot connect to any (originally requested NDAR_Central_2) S3 bucket. Exception: S3ResponseError: 400 Bad Request\n [s3.py:_handle_exception:75]'
	datalad.cmd     [Level 5] stderr|   external special remote protocol error, unexpectedly received " [s3.py:_handle_exception:75]" (unable to parse command)
	datalad.cmd     [Level 5] stderr| datalad.customremotes [Level 4] Sending 'CHECKURL-FAILURE'
	datalad.cmd     [Level 5] stderr| datalad         [Level 5] Instantiating ssh manager
	datalad.cmd     [Level 5] stderr| datalad         [Level 5] Done importing main __init__
	datalad.cmd     [Level 5] stderr| datalad.customremotes [Level 5] Importing datalad.customremotes.main
	datalad.cmd     [Level 5] stderr| datalad         [Level 5] Requested to provide version for tqdm

since credentials timed out and downloader puked an exception with multiple
lines.  That \n is due to repr on the msg when we log, so in reality we were
trying to send multiline string

### DIFF
--- a/datalad/customremotes/base.py
+++ b/datalad/customremotes/base.py
@@ -322,6 +322,8 @@ class AnnexCustomRemote(object):
            arguments to be joined by a space and passed to git-annex
         """
         msg = " ".join(map(str, args))
+        # Sanitize since there must be no new lines
+        msg = msg.replace(os.linesep, r'\n')
         if not self._in_the_loop:
             lgr.debug("We are not yet in the loop, thus should not send to annex"
                       " anything.  Got: %s" % msg.encode())


kyleam (approved):

yarikoptic:
Thank you @kyleam ! Unfortunately I have not came up with an easy unittest to test this :-/  also real use case is running with #4927 merged in (which addresses underlying issue why I was getting newlines to begin with), so I hope that I will not trigger any of such cases as well.  So I will proceed with the merge.

PR #4928, opened 2020-09-17
ENH: retry individual addurl if addurl fails up to 3 times

yarikoptic:
Not yet sure if that is the best way to proceed.  But I feel that there is
some "fluke" which causes git annex to report a failed CHECKURL .
I could not make sense of available logs to see if it is on our end or not yet.

- [ ] decide either to bother with this approach at all


yarikoptic:
decided not to bother and better address underlying issue with #4931

PR #4931, opened 2020-09-18
ENH+RF: @try_multiple_dec_s3 to retry interaction with S3 upon "retriable" errors

yarikoptic:
I think this is an alternative (as a better solution to the underlying issue) to the https://github.com/datalad/datalad/pull/4928 -- I kept hitting a `400` error from S3 while `addurls` ABCD.  I believe in the course of DANDI project we had similar occasions whenever S3 would fail with 400 without apparent reason: well, there are many reasons why 400 could be returned - https://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html#ErrorCodeList and some of them are "legit", i.e. retry probably should not 'fix' that.  But I do not want to sift through all of them ATM.  IMHO it would be perfectly fine to retry a few times.

I have also added other codes (with comments in the code) which I think are legit to retry with sleep upon hitting.

Also `@try_multiple_dec` was enhanced and refactored a little to make it more useful generally.

I will now do a rerun of `addurls` with this PR applied so I could see if I finally manage to complete the runs of `addurls` without crash.

TODOs
- [x] figure out why not "effective" for my use case
- test the code by mocking `.get_bucket` to raise an exception on first try so forcing a retry - I did come up with a test but it is only an integration test which has extended run time... was not sure if mocked up test would be more pain than gain since it is really to test interaction with S3.  So decided to not pursue that.

### DIFF
--- a/datalad/downloaders/s3.py
+++ b/datalad/downloaders/s3.py
@@ -208,11 +208,14 @@ class S3Downloader(BaseDownloader):
         bucket_name = self._parse_url(url, bucket_only=True)
         if allow_old and self._bucket:
             if self._bucket.name == bucket_name:
-                lgr.debug(
-                    "S3 session: Reusing previous connection to bucket %s",
-                    bucket_name
-                )
-                return True  # we used old
+                if self.credential and self.credential.is_expired:
+                    lgr.debug("S3 session: credential expired")
+                else:
+                    lgr.debug(
+                        "S3 session: Reusing previous connection to bucket %s",
+                        bucket_name
+                    )
+                    return True  # we used old
             else:
                 lgr.warning("No support yet for multiple buckets per S3Downloader")
 
--- a/datalad/downloaders/s3.py
+++ b/datalad/downloaders/s3.py
@@ -17,7 +17,10 @@ from ..utils import (
     assure_dict_from_str,
     auto_repr,
 )
-from ..dochelpers import borrowkwargs
+from ..dochelpers import (
+    borrowkwargs,
+    exc_str,
+)
 from ..support.network import (
     get_url_straight_filename,
     iso8601_to_epoch,
@@ -27,7 +30,7 @@ from ..support.network import (
 from .base import Authenticator
 from .base import BaseDownloader, DownloaderSession
 from ..support.exceptions import (
-    DownloadError,
+    AccessDeniedError,
     TargetFileAbsent,
 )
 from ..support.s3 import (
@@ -237,6 +240,13 @@ class S3Downloader(BaseDownloader):
                 url_filepath, version_id=params.get('versionId', None)
             )
         except S3ResponseError as e:
+            # e.g. 400 Bad request could happen due to timed out key.
+            # Since likely things went bad if credential expired, just raise general
+            # AccessDeniedError. Logic upstream should retry
+            if self.credential and self.credential.is_expired:
+                raise AccessDeniedError(
+                    "Failed to get a key likely due to expired key: %s"
+                    % exc_str(e))
             raise TargetFileAbsent("S3 refused to provide the key for %s from url %s: %s"
                                 % (url_filepath, url, e))
         if key is None:
--- a/datalad/downloaders/s3.py
+++ b/datalad/downloaders/s3.py
@@ -30,7 +30,7 @@ from ..support.network import (
 from .base import Authenticator
 from .base import BaseDownloader, DownloaderSession
 from ..support.exceptions import (
-    AccessDeniedError,
+    AccessPermissionExpiredError,
     TargetFileAbsent,
 )
 from ..support.s3 import (
@@ -244,7 +244,7 @@ class S3Downloader(BaseDownloader):
             # Since likely things went bad if credential expired, just raise general
             # AccessDeniedError. Logic upstream should retry
             if self.credential and self.credential.is_expired:
-                raise AccessDeniedError(
+                raise AccessPermissionExpiredError(
                     "Failed to get a key likely due to expired key: %s"
                     % exc_str(e))
             raise TargetFileAbsent("S3 refused to provide the key for %s from url %s: %s"
--- a/datalad/downloaders/s3.py
+++ b/datalad/downloaders/s3.py
@@ -13,11 +13,16 @@ import re
 
 from urllib.parse import urlsplit, unquote as urlunquote
 
-from ..utils import auto_repr
-from ..utils import assure_dict_from_str
+from ..utils import (
+    assure_dict_from_str,
+    auto_repr,
+)
 from ..dochelpers import borrowkwargs
-from ..support.network import get_url_straight_filename
-from ..support.network import rfc2822_to_epoch, iso8601_to_epoch
+from ..support.network import (
+    get_url_straight_filename,
+    iso8601_to_epoch,
+    rfc2822_to_epoch,
+)
 
 from .base import Authenticator
 from .base import BaseDownloader, DownloaderSession
@@ -25,8 +30,13 @@ from ..support.exceptions import (
     DownloadError,
     TargetFileAbsent,
 )
-from ..support.s3 import boto, S3ResponseError, OrdinaryCallingFormat
-from ..support.s3 import get_bucket
+from ..support.s3 import (
+    OrdinaryCallingFormat,
+    S3ResponseError,
+    try_multiple_dec_s3,
+    boto,
+    get_bucket,
+)
 from ..support.status import FileStatus
 
 import logging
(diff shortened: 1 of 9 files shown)


datalad/utils.py
```diff
@@ -1983,28 +1995,31 @@ def try_multiple_dec(f, ntrials=None, duration=0.1, exceptions=None, increment_t
...
         ntrials = 100 if on_windows else 10
-
+    if logger is None:
+        def logger(*args, **kwargs):
+            return lgr.log(5, *args, **kwargs)
     assert increment_type in {None, 'exponential'}
 
     @wraps(f)
-    def wrapped(*args, **kwargs):
+    def _wrap_try_multiple_dec(*args, **kwargs):
```
yarikoptic, line 2004: note: while at it I have renamed this generic `wrapped` which generally should not be used for DX while observing the traceback etc

datalad/utils.py
```diff
@@ -1974,7 +1978,15 @@ def try_multiple_dec(f, ntrials=None, duration=0.1, exceptions=None, increment_t
     increment_type: {None, 'exponential'}
       Note that if it is exponential, duration should typically be > 1.0
       so it grows with higher power
-
+    exceptions: Exception or tuple of Exceptions, optional
+      Exception or a tuple of multiple exceptions, on which to retry
+    exceptions_filter: callable, optional
+      If provided, this unction will be called with a caught exception
```
kyleam, line 1984:
```suggestion
      If provided, this function will be called with a caught exception
```

kyleam (approved):
Looks good.

datalad/utils.py
yarikoptic: THANK YOU!

yarikoptic:
Update: somehow this change did not help me -- yet to figure out why (difficulty is all the sandwidching of datalad/git-annex/datalad-remote + inability to enter even epdb since it dumps to stdout tripping git annex).  so I let's wait to merge it until I see how to make it actually work for the target use case or why it didn't help when it should've

yarikoptic:
I think I figured it out (although I guess I should have figured it without additional logging ;)):  CHECKURL happens right after the token has expired but we are reusing previous session.  That is why we receive Bad Request.  But that deep in the code we do not check on validity of the credentials and thus just "keep knocking on now closed door".  Also interestingly (not yet sure why since we do swallow all --debug output from batched annex and with #4925 (which I use as well) show only the tail of the log and misses that point where restart happens.  Upon restart of the special remote it does check that credential is still valid and does re-mint it since it is expired.  Pushed some commits to address it and other discovered while at it gotchas -- it MUST work now! ;)

yarikoptic:
uff, it dosn't (at least not fully). Current stumbling point is within [this question](https://github.com/boto/boto3/issues/2598) to boto people and might have a simple resolution.

Nevertheless -- I have pushed more of IMHO relevant etc fixes/refactorings, and although I believe they should be ok, no longer sure if `maint` would remain appropriate target.   I have not even tried to rerun in real use case of mine (ABCD) with those changes yet since I think they would not help yet.  Ideally I should establish a small test which would exercise all the added logic etc.

yarikoptic:
grr, still not there... 
```
datalad.downloaders.credentials {2375010}[DEBUG] Credential NDA:1 will expire in 11.96h
datalad.s3      {2375010}[Level 5] Empty error_code in S3ResponseError: 400 Bad Request

datalad.s3      {2375010}[DEBUG] Caught S3ResponseError: 400 Bad Request
 [bucket.py:_get_key_internal:231] on trial #0. Sleeping 2.000000 and retrying
datalad.s3      {2375010}[Level 5] Empty error_code in S3ResponseError: 400 Bad Request

datalad.s3      {2375010}[DEBUG] Caught S3ResponseError: 400 Bad Request
 [bucket.py:_get_key_internal:231] on trial #1. Sleeping 4.000000 and retrying
datalad.s3      {2375010}[Level 5] Empty error_code in S3ResponseError: 400 Bad Request

datalad.s3      {2375010}[DEBUG] Caught S3ResponseError: 400 Bad Request
 [bucket.py:_get_key_internal:231] on trial #2. Sleeping 8.000000 and retrying
datalad.s3      {2375010}[Level 5] Empty error_code in S3ResponseError: 400 Bad Request

datalad.downloaders.credentials {2375010}[DEBUG] Credential NDA:1 will expire in 11.96h
datalad.customremotes {2375010}[DEBUG] Failed to check url s3://NDAR_Central_3/XXX: S3ResponseError: 400 Bad Request
|  [s3.py:get_downloader_session:262]
datalad.customremotes {2375010}[Level 4] Sending 'DEBUG Failed to check url s3://NDAR_Central_3/XXX: S3 refused to provide the key for XXX from url s3://NDAR_Central_3/XXX: S3ResponseError: 400 Bad Request\\n [s3.py:get_downloader_session:262]'
datalad.customremotes {2375010}[Level 4] Sending 'CHECKURL-FAILURE'
```

`11.96h` suggests that it might still be some zone issue -- too close to an even hour... I will check that first

datalad/downloaders/s3.py
```diff
@@ -198,18 +212,127 @@ def _establish_session(self, url, allow_old=True):
...
+            return self._bucket.get_key(key_name, version_id=version_id, headers=headers)
+        except S3ResponseError as e:
+            if e.status != 400:
+                raise  # we will not deal with those here
+            # e.g. 400 Bad request could happen due to timed out key.
+            # Since likely things went bad if credential expired, just raise general
+            # AccessDeniedError. Logic upstream should retry
+            self._check_credential()
+            try:
+                return self._get_key_via_get(key_name, version_id=version_id, headers=headers)
```
yarikoptic, line 255: crap -- I forgot to add a `lgr.debug` here to actually spot such cases when they happen.  Current run 'succeeded' but I do not know if it is thanks to this code or not :-/

yarikoptic:
uff -- CIs fail due to some python 3.8 specific code I blindly introduced:
```
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/datalad/downloaders/tests/test_s3.py", line 214
    yield check_get_key, *p
                         ^
SyntaxError: invalid syntax
```
so I will submit a fix up.
But meanwhile I would like to declare a victory.  In the logs of the process which passed over expiration time point (12h) I see

```
$> grep -a expired datalad-20200922-11.log
2020-09-22 20:00:23,041 datalad.downloaders.credentials {3368068}[DEBUG] Credential NDA:1 has expired -2.74h ago 
2020-09-23 08:00:21,381 datalad.downloaders.credentials {3368068}[DEBUG] Credential NDA:1 has expired -0.00h ago 
2020-09-23 08:00:21,382 datalad.s3      {3368068}[DEBUG] S3 session: credential expired 
2020-09-23 08:00:21,385 datalad.downloaders.credentials {3368068}[DEBUG] Credential NDA:1 has expired -0.00h ago 
2020-09-23 08:00:24,747 datalad.downloaders {3368204}[DEBUG] Authenticated access was denied: Used token to access S3 has expired: S3ResponseError: 400 Bad Request
<Error><Code>ExpiredToken</Code><Message>The provided token has expired.</Message><Token-0>SENSORED</Token-0><RequestId>41B5B32858B9CF06</RequestId><HostId>SENSORED</HostId></Error> [s3.py:_get_key_via_get:334] [s3.py:_check_S3ResponseError:99] 
| DEBUG:datalad.downloaders.credentials:Credential NDA:1 has expired -0.00h ago
| DEBUG:datalad.s3:S3 session: credential expired
| DEBUG:datalad.downloaders.credentials:Credential NDA:1 has expired -0.00h ago
```
and it is still going forward "strong".  So we did receive an error_code on expired token (I guess via GET request -- will add a log line for future), and handled it gracefully AFAIK.

kyleam:
> @yarikoptic yarikoptic requested a review from kyleam

Could you summarize what has changed since I last reviewed it?  From a glance, this has morphed into a much larger and different PR that will be hard for anyone except the person that was working on it at the time to make sense of.

kyleam (approved):
I read through each of the new commits and didn't spot any obvious issues.  I'm not very familiar with this area of the code base and don't have a good feel for the wider context, though.

yarikoptic:
Thank you @kyleam for the review.  Indeed PR grew up (un)fortunately. I am 90% confident (as the "most familiar" but still not really knowledgeable in this now old code) that there should only be positive outcomes from these changes, although indeed they touch on the "heart" of downloaders which are used quite a bit.  I have also ran locally datalad-crawler tests (which should excercise S3 code as well given that I have credentials locally) - seems to be all green and no side-effects.  So, overall, I hope that it just addresses issues of

- being able to detect and re-authenticate upon S3 token expiration
- retry download in various cases where S3 returns an error making download more robust
- correctly validate that no page with "login failed" message was returned as output by external website whenever target file is non utf8

I will let it age for a day or so, and if nobody else raises objections, will merge it.

yarikoptic:
Woohoo -- my addurls process is still going strong (well -- struggles with `git` since I did not split into subdatasets enough, so some are huge), but it has not crashed!  So let's proceed and test more in real-life after merge before any release.

PR #4940, opened 2020-09-21
prep-travis-forssh.sh: Check SSH readiness with nc instead of sleeping

jwodder:
This change causes `prep-travis-forssh.sh` to wait for the SSH containers to be ready by repeatedly checking whether connections can be made to their SSH ports.  This should be more robust than just sleeping for a fixed number of seconds.


kyleam:
Nice, thank you.  I'm going to rebase this to maint.

Issue #4944, opened 2020-09-22
Add a plugin-in infrastructure for specialized index-creators


christian-monch:
To support search, datalad creates an index from the metadata that is emitted by the extractors. This index is a set of key-value pairs, where the keys are dotted-name representations of the objects that the extractors return, for example: "datalad_core.id".

The conversion from metadata-object hierarchies to key-value pairs is done automatically. The automated conversion is necessarily generic and works well for a number of metadata objects, but will not always produce the desired results. For example, JSON-LD-like metadata objects are usually represented in unnecessary detail.

I suggest creating a registry for metadata-to-index conversion where metadata extractors can register a conversion method that will create the "desired" key-value pairs from their output. This conversion method could then be used for metadata objects emitted by the registered extractors. Indices for all other metadata objects would still be created by the automated object-index conversion.

PR #4945, opened 2020-09-22
Nf json ld search

christian-monch:
These changes modify the automated index creation from metadata objects to be able to create indices from JSON-LD objects, e.g. process @graph and @list. In addition, this PR adds a dedicated JSON-LD converter for studyminimeta-metadata objects (this converted should be moved to a different place later, cf. issue https://github.com/datalad/datalad/issues/4944)

The change will increase the number of key-value pairs that are searchable, i.e. the list that emitted by the command "datalad search --show-keys name" is longer.

In order to properly separate key-name spaces in woosh search, key names have been amended with an index notation. This allows separating the subtrees contained in a list-like metadata-objects. For example, the following keys will be printed by "datalad search --show-keys name", if the metalad_core extractor was executed:

```
 . . .
metalad_core.graph[0].agent.email
metalad_core.graph[0].agent.id
metalad_core.graph[0].agent.name
 . . .
``` 

A follow up to this commit would be to add a mechanism to register extractor-specific index creation and use this mechanism, instead of hard-coded branches, to perform extractor-specific index creation (see also issue: https://github.com/datalad/datalad/issues/4944)

### DIFF
--- a/datalad/metadata/tests/test_search.py
+++ b/datalad/metadata/tests/test_search.py
@@ -388,7 +388,7 @@ _mocked_studyminimeta_jsonld = {
         {
             "@id": "#study",
             "@type": "CreativeWork",
-            "name": "Eine Studie",
+            "name": "A small study",
             "abstract": "a short description of the study",
             "accountablePerson": "a@example.com",
             "keywords": [
@@ -424,7 +424,7 @@ _mocked_studyminimeta_jsonld = {
             "version": "a02312398324778972389472834",
             "name": "Datenddaslk",
             "url": "http://dlksdfs.comom.com",
-            "description": "Ein paar Daten, die ich mal gesammelt habe. Ein paar Daten, die ich mal gesammelt habe.",
+            "description": "Some data I collected once upon a time, spending hours and hours in dark places.",
             "keywords": [
                 "d_k1",
                 "d_k2",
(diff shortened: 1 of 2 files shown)


datalad/metadata/tests/test_search.py
```diff
@@ -358,3 +358,265 @@ def test_meta2autofield_dict():
...
+    # Just a test that we would obtain the value stored for that extractor
+    # instead of what unique values it already had (whatever that means)
+    eq_(
+        _meta2autofield_dict({"r": {"@list": ["a", "b", "c"]}}),
+        {'r.list[0]': 'a', 'r.list[1]': 'b', 'r.list[2]': 'c'}
+    )
+
+
+_mocked_studyminimeta_jsonld = {
+    "@context": {
```
yarikoptic, line 384: Since this structure already is here, could there be a test for actual search? Just to make sure that no side effects etc?

christian-monch: Good point. Adding tests for show-keys and proper search

christian-monch: @yarikoptic: tests added.

yarikoptic: Thanks!

yarikoptic:
- this PR sounds more like a new feature, has an extensive diff, so I wonder if `master` would be more appropriate IMHO
- is there any public (meta)dataset may be to try it out on a real use case?

christian-monch:
> * this PR sounds more like a new feature, has an extensive diff, so I wonder if `master` would be more appropriate IMHO
> * is there any public (super)dataset may be to try it out on a real use case?

I can close this PR and open a new PR to master. A candidate would be the archived-studies super-dataset, it is not public though.

In a new PR against master, I might first open a PR to address issue #4944, i.e. individual indexers (analog to extractors)

christian-monch:
Will close this PR, work is continued on master

PR #4947, opened 2020-09-22
Workflow for testing on macOS

### LINKED ISSUE #4942: Add github workflow to test on OSX
In the light of #4941 I think it would be very useful to get at least some (`datalad.support` and `datalad.core` at least or all which are not `slow` or `turtle`?) unittests being tested on OSX on github actions.  Then PRs addressing failures listed in #4941 could expand coverage to cover what is fixed up.

For now git-annex could be installed from the officially distributed .dmg, but eventually I hope that we could provide some builds straight from https://github.com/datalad/datalad-extensions/ .

jwodder:
Closes #4942.

### DIFF
--- a/.github/workflows/test_macos.yml
+++ b/.github/workflows/test_macos.yml
@@ -1,0 +1,41 @@
+name: Test on macOS
+
+on:
+    - pull_request
+    - push
+
+jobs:
+  test:
+    runs-on: macos-latest
+    steps:
+    - name: Set up system
+      shell: bash
+      run: |
+        brew install git-annex
+
+    - name: Set up environment
+      run: |
+        git config --global user.email "test@github.land"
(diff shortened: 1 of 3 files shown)


yarikoptic:
Great. Please add `known_failure_githubci_osx` similar to `known_failure_githubci_win` and decorate those two failing tests with it so we get green at least and proceed.  IMHO ideally this one should be positioned against `maint` branch (not `master`) since would be of benefit there.

jwodder:
@yarikoptic Done.

yarikoptic:
I think more green runs is good, so I will proceed with merge.
Could you please try to add codecov submission in a follow up PR? There was something not working for windows workflow https://github.com/datalad/datalad/blob/master/.github/workflows/test_win2019_disabled#L59 but may be it would work for osx?

Issue #4951, opened 2020-09-23
downloader -- unicode wows upon verification of the fetched content


yarikoptic:
we haven't seen such ones for a while, but here is one ;-)

```shell
$> datalad download-url https://portal.nersc.gov/project/crcns/download/ac-5/docs/data_analysis_instructions.txt
[INFO   ] Downloading 'https://portal.nersc.gov/project/crcns/download/ac-5/docs/data_analysis_instructions.txt' into '/tmp/' 
[ERROR  ] Failed to download https://portal.nersc.gov/project/crcns/download/ac-5/docs/data_analysis_instructions.txt into /tmp/data_analysis_instructions.txt: 'utf-8' codec can't decode byte 0xd4 in position 785: invalid conti
nuation byte [base.py:_download:452,base.py:_verify_download:370,codecs.py:decode:322] 
download_url(error): /tmp/ (file) ['utf-8' codec can't decode byte 0xd4 in position 785: invalid continuation byte [codecs.py:decode:322] [base.py:_download:474]]

$> datalad --version
datalad 0.13.3
```

PR #4952, opened 2020-09-23
downloader: handle non-utf8 files being downloaded

yarikoptic:
Some downloaders need to read-in the actual downloaded content to see if it may be a "login error" message since portals do not return a proper 403 etc.  If content is not utf8 - it was causing a puke (#4951) .

This is likely a tip of an iceberg and there could be more of places in datalad where we read some content assuming it to be utf-8, whenever it is not.  Introduced here `read_file` (I am ok to rename if desired, initially `read_file_ensure_unicode` - generalized into one to use without decoding as well) would be of help.  May be you know a better way? but I thought it would be nice to centralize all such logic on encoding discovery via `ensure_unicode` which we do use already quite a bit throughout the code.

edit 1 after more pushes: One functionality change/fix is for S3 downloader is to not try to decode content (when no file is provided) to stay consistent with http downloader which uses BytesIO and just returns its content.  It is up for outside logic on either to decode it ATM, although generally downloaders should respect content-type/encoding header fields and act appropriately (which AFIAK they don't do ATM) but that is outside of the scope of this BF PR.

### DIFF
--- a/datalad/downloaders/base.py
+++ b/datalad/downloaders/base.py
@@ -27,7 +27,6 @@ from ..ui import ui
 from ..utils import (
     auto_repr,
     ensure_unicode,
-    read_file_ensure_unicode,
     unlink,
 )
 from ..dochelpers import exc_str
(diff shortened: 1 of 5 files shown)


yarikoptic:
oh -- gotcha immediately discovered on another dataset...  if content is binary -- would kaboom... so more fixing needed.

yarikoptic:
Please review.  This is yet another downloaders fixup which got a bit "evolved".  Original description and commits messages hopefully provide enough reasoning behind changes.

kyleam (approved):
This looks okay to me, and as far as I can tell, it doesn't break not-so-distant changes in this area (in particular gh-3210).

yarikoptic:
Thank you! Let's proceed then!  Next release of `maint` would be either rock solid or ... the opposite ;)

Issue #4954, opened 2020-09-23
FOI: some timings for "external remote" interaction + idea for OPT of CHECKPRESENT


yarikoptic:
So I was mumbling during last meeting that may be that it is the interaction between annex and external remotes (via stdin/out) is what is slowing down some use cases (no high CPU or IO load neither on datalad nor git-annex process), and that may be git-annex/external remote are spending some time waiting on input etc.

So I

- cloned http://github.com/datalad-datasets/hcp_MNI_T1w which has some notable number of annexed files in the tree (~4k) available via s3
- used `DATALAD_TESTS_PROTOCOLREMOTE=1` feature which produced a bash script for replay of the interactions of our special remote with annex (stored under .git/bin/git-annex-remote-datalad)
- lobotomized it to just always immediately respond with `CHECKPRESENT-SUCCESS`

<details>
<summary>the script</summary> 

```shell
#!/bin/bash

set -e

# Gets a VALUE response and stores it in $RET
report () {
    # echo "$@" >&2
    :
}

recv () {
    read resp
    #resp=${resp%\n}
    target="$@"
    if [ "$resp" != "$target" ]; then
        report "! exp $target"
        report "  got $resp"
    else
        report "+ got $resp"
    fi
}

send () {
    echo "$@"
    # report "sent $@"
}

send VERSION 1
recv EXTENSIONS INFO ASYNC
send UNSUPPORTED-REQUEST
recv PREPARE
send PREPARE-SUCCESS
send DEBUG Encodings: filesystem utf-8, default utf-8

# from now on should be checkpresent sequence
# recv CHECKPRESENT MD5E-s70189716--45f4514f972325f481d5c09434c1c94a.nii.gz
while read CMD KEY; do
    if [ $CMD != CHECKPRESENT ]; then
        echo "Was expecting CHECKPRESENT, got $CMD"
        exit 1
    fi
    #send GETURLS MD5E-s70189716--45f4514f972325f481d5c09434c1c94a.nii.gz http:
    #recv VALUE 
    #send GETURLS MD5E-s70189716--45f4514f972325f481d5c09434c1c94a.nii.gz https:
    #recv VALUE 
    #send GETURLS MD5E-s70189716--45f4514f972325f481d5c09434c1c94a.nii.gz s3:
[...]

PR #4955, opened 2020-09-23
OPT: gen_URLS (to replace get_URLS) - query schemes in order of prior "successes"

yarikoptic:
Ref: #4954 

I do  not think it would be of tremendous performance boost (since most of the time is likely to be spent in interaction with external resources) but it can cut interactions between git-annex and `datalad` special remote significantly, and will pollute log less.

I think the change is quite straightforward and I would not mind to rebase it against `maint`, but I am also ok to keep it in master.

### DIFF
--- a/datalad/customremotes/base.py
+++ b/datalad/customremotes/base.py
@@ -17,6 +17,8 @@ import errno
 import os
 import sys
 
+from collections import Counter
+
 from ..support.path import exists, join as opj, dirname, lexists
 
 from urllib.parse import urlparse
@@ -257,6 +259,12 @@ class AnnexCustomRemote(object):
         # Delay introspection until the first instance gets born
         # could in principle be done once in the metaclass I guess
         self.__class__._introspect_req_signatures()
+
+        # OPT: a counter to increment upon successful encounter of the scheme
+        # (ATM only in gen_URLS but later could also be used in other requests).
+        # This would allow to consider schemes in order of decreasing success instead
+        # of arbitrary hardcoded order
+        self._scheme_hits = Counter({s: 0 for s in self.SUPPORTED_SCHEMES})
 
     @classmethod
     def _introspect_req_signatures(cls):
@@ -624,7 +632,7 @@ class AnnexCustomRemote(object):
 
         """
         nurls = 0
-        for scheme in self.SUPPORTED_SCHEMES:
+        for scheme, _ in self._scheme_hits.most_common():
             scheme_ = scheme + ":"
             self.send("GETURLS", key, scheme_)
             # we need to first to slurp in all for a given SCHEME
@@ -642,8 +650,12 @@ class AnnexCustomRemote(object):
                     scheme_urls.append(url[0])
                 else:
                     break
-            for url in scheme_urls:
-                yield url
+            if scheme_urls:
+                # note: generator would ceise to exist thus not asking
+                # for URLs for other schemes if this scheme is good enough
+                self._scheme_hits[scheme] += 1
+                for url in scheme_urls:
+                    yield url
 
         self.heavydebug("Got %d URL(s) for key %s", nurls, key)
 
--- a/datalad/customremotes/base.py
+++ b/datalad/customremotes/base.py
@@ -619,14 +619,17 @@ class AnnexCustomRemote(object):
         else:
             return val
 
-    def get_URLS(self, key):
-        """Gets URL(s) associated with a Key.
+    def gen_URLS(self, key):
+        """Yield URL(s) associated with a Key.
 
         """
-        urls = []
+        nurls = 0
         for scheme in self.SUPPORTED_SCHEMES:
             scheme_ = scheme + ":"
             self.send("GETURLS", key, scheme_)
+            # we need to first to slurp in all for a given SCHEME
+            # since annex would be expecting to send its final empty VALUE
+            scheme_urls = []
             while True:
                 url = self.read("VALUE", 1)
                 if not url or len(url) <= 1:
@@ -635,16 +638,22 @@ class AnnexCustomRemote(object):
                 url = url[1:]
                 if url:
                     assert(len(url) == 1)
-                    urls.append(url[0])
+                    nurls += 1
+                    scheme_urls.append(url[0])
                 else:
                     break
+            for url in scheme_urls:
+                yield url
 
-        self.heavydebug("Got %d URL(s) for key %s: %s", len(urls), key, urls)
+        self.heavydebug("Got %d URL(s) for key %s", nurls, key)
 
-        #if not urls:
-        #    raise ValueError("Did not get any URLs for %s which we support" % key)
+    def get_URLS(self, key):
+        """Gets URL(s) associated with a Key.
 
-        return urls
+        Use a generator gen_URLS where possible.
+        This one should be deprecated in 0.15.
+        """
+        return list(self.gen_URLS(key))
 
     def _get_key_path(self, key):
         """Return path to the KEY file
--- a/datalad/customremotes/datalad.py
+++ b/datalad/customremotes/datalad.py
@@ -38,7 +38,6 @@ class DataladAnnexCustomRemote(AnnexCustomRemote):
         # about.  So for a key we might get back multiple URLs and as a
         # heuristic let's use the most recently asked one
 
-        self._last_url = None  # for heuristic to choose among multiple URLs
         self._providers = Providers.from_config_files()
 
     #
@@ -144,11 +143,6 @@ class DataladAnnexCustomRemote(AnnexCustomRemote):
         # TODO: We might want that one to be a generator so we do not bother requesting
         # all possible urls at once from annex.
         urls = self.get_URLS(key)
-
-        if self._last_url in urls:
-            # place it first among candidates... some kind of a heuristic
-            urls.pop(self._last_url)
-            urls = [self._last_url] + urls
 
         # TODO: priorities etc depending on previous experience or settings
 
--- a/datalad/customremotes/datalad.py
+++ b/datalad/customremotes/datalad.py
@@ -92,7 +92,7 @@ class DataladAnnexCustomRemote(AnnexCustomRemote):
         """
         lgr.debug("VERIFYING key %s" % key)
         resp = None
-        for url in self.get_URLS(key):
+        for url in self.gen_URLS(key):
             # somewhat duplicate of CHECKURL
             try:
                 status = self._providers.get_status(url)
@@ -142,11 +142,12 @@ class DataladAnnexCustomRemote(AnnexCustomRemote):
 
         # TODO: We might want that one to be a generator so we do not bother requesting
         # all possible urls at once from annex.
-        urls = self.get_URLS(key)
+        urls = []
 
         # TODO: priorities etc depending on previous experience or settings
 
-        for url in urls:
+        for url in self.gen_URLS(key):
+            urls.append(url)
             try:
                 downloaded_path = self._providers.download(
                     url, path=path, overwrite=True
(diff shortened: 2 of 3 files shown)


datalad/customremotes/datalad.py
```diff
@@ -93,7 +92,7 @@ def req_CHECKPRESENT(self, key):
         """
         lgr.debug("VERIFYING key %s" % key)
```
yarikoptic, line 93: note: hm, coverage suggests that this piece of code is not covered... that is a bit surprising since `test_basic_scenario_local_url` does hit that code but in the external process of special remote.  I guess at some point we lost proper functioning of merging coverage from multiple reports (i.e. from the main nose process + special remotes which run underneath). And indeed, filed https://github.com/datalad/datalad/issues/4956

datalad/customremotes/base.py
```diff
@@ -635,16 +646,26 @@ def get_URLS(self, key):
...
                 if url:
                     assert(len(url) == 1)
-                    urls.append(url[0])
+                    nurls += 1
+                    scheme_urls.append(url[0])
                 else:
                     break
+            if scheme_urls:
+                # note: generator would ceise to exist thus not asking
+                # for URLs for other schemes if this scheme is good enough
```
kyleam, line 655: s/ceise/cease/

I don't understand this comment and how it relates to the commit where it was added (7da0b6909).

kyleam (commented):
> it can cut interactions between git-annex and datalad special remote significantly, and will pollute log less.

One thought I had when reading over these changes is whether it'd be better to call `GETURLS` with an empty prefix and then do post-processing of all the URLs on our side.

yarikoptic:
> One thought I had when reading over these changes is whether it'd be better to call `GETURLS` with an empty prefix and then do post-processing of all the URLs on our side.

yeah, I was thinking about that as well... in some cases it might be beneficial (all URLs are "datalad" urls) but in some might be the opposite -- bunch of urls we do not want to handle, but now would need to fetch and all those lines from annex first before we proceed.  With this reordering of supported protocols and asking for the one which we got responses for before, we make it more efficient as a whole (no URLs we would not care, immediate hit for the prefix we do care about).  So I decided to go with the sorting approach.

datalad/customremotes/base.py
yarikoptic: it somewhat relates to both changes of 1. making it a generator; 2. sorting so we hit first the one we had succeeded before with; and probably to 1. more than to the 2.  But since both commits are in the same PR I guess no need to move the comment, unless you insist.
As for more explanation re 1.: since it is a generator, if the first scheme succeeds to find URLs to which we e.g. perform successfully the CHECKPRESENT, then we would not even consider the next scheme/prefix since generator will not be asked for the "next" one.  If it was not a generator -- we go through all schemes regardless.

kyleam:
>  With this reordering of supported protocols and asking for the one which we got responses for before, we make it more efficient as a whole (no URLs we would not care, immediate hit for the prefix we do care about). So I decided to go with the sorting approach.

I don't know that it is "more efficient on the whole".  Do you expect it to be common to get a whole bunch of URLs we don't care about?  And even if you did, how does the processing weigh against eliminating additional interactions with annex?  Plus, wouldn't asking for all the URLs have the advantage of gen_URLS yielding URLs for all supported schemes?

At any rate, without actual timings it's hard to say anything.  Feel free of course to go with whatever approach you think is best.

datalad/customremotes/base.py
kyleam: > since it is a generator, if the first scheme succeeds to find URLs to which we e.g. perform successfully the CHECKPRESENT, then we would not even consider the next scheme/prefix since generator will not be asked for the "next" one. If it was not a generator -- we go through all schemes regardless.

If I understand correctly, perhaps something like this then: ""Note: Yield a list of URLs per scheme so that a caller can decide to stop when it gets a scheme that it considers good enough".

yarikoptic:
> I don't know that it is "more efficient on the whole". Do you expect it to be common to get a whole bunch of URLs we don't care about? 

it can happen. E.g. in the course of openneuro we had a mix of regular urls which could be handled internally by git-annex or by  `datalad`  + `datalad-archives` remote urls:
```
(git-annex)lena:~/datalad/openfmri/ds000001[master]sub-01/anat
$> git annex whereis sub-01_inplaneT2.nii.gz
whereis sub-01_inplaneT2.nii.gz (5 copies) 
  	00000000-0000-0000-0000-000000000001 -- web
   	899f0347-0888-48ef-91b6-bac213ca8cef -- [datalad-archives]
   	c8bd3d05-33d4-4b59-9d53-ca7efbdcdd13 -- yoh@smaug:/mnt/btrfs/datasets/datalad/crawl/openfmri/ds000001
   	ce96a357-a40c-4107-9039-77124f428aae -- [datalad]
   	e89f69c2-d802-415d-a819-b29ca80daefd -- [mydirectory]

  The following untrusted locations may also have copies:
  	766676d0-7be4-4b0b-a583-d115a5d18353 -- [mydirectory-exported2]
   	9f989208-cccf-471a-8b92-877dec8bb622 -- yoh@hopa:~/.annex-cache [cache]

  web: http://openneuro.s3.amazonaws.com/ds000001/ds000001_R1.1.0/uncompressed/sub001/anatomy/inplane001.nii.gz?versionId=ystKDnaPkdzSwzdRPZH0PtMMknZJCQV4
  web: http://openneuro.s3.amazonaws.com/ds000001/ds000001_R2.0.1/uncompressed/sub-01/anat/sub-01_inplaneT2.nii.gz?versionId=wVn1tHi1XBn7avQ9Q1oXuoqE2Splyc.X
  web: http://openneuro.s3.amazonaws.com/ds000001/ds000001_R2.0.2/uncompressed/sub-01/anat/sub-01_inplaneT2.nii.gz?versionId=RQFsRCUlj0X77.OTMJZ01Dcx5MRjfljQ
  web: http://openneuro.s3.amazonaws.com/ds000001/ds000001_R2.0.3/uncompressed/sub-01/anat/sub-01_inplaneT2.nii.gz?versionId=XdlgqNsFbXeHxzHsdZkR1CUNxozBcL8W
  web: http://openneuro.s3.amazonaws.com/ds000001/ds000001_R2.0.4/uncompressed/sub-01/anat/sub-01_inplaneT2.nii.gz?versionId=ZY7d3uaj43tD6oPwq8JCuRyEBD91OgCY

  datalad-archives: dl+archive:MD5E-s2415911708--63e02dea2a059517f25d1c2c9ae6c423.zip#path=ds000001_R2.0.2/sub-01/anat/sub-01_inplaneT2.nii.gz&size=669578
  datalad-archives: dl+archive:MD5E-s2415911710--5e509bac5a729d668f900d0440228402.zip#path=ds000001_R2.0.3/sub-01/anat/sub-01_inplaneT2.nii.gz&size=669578
  datalad-archives: dl+archive:MD5E-s2416078169--53fbcb667531f96e2532723e74ed218d.zip#path=ds001_R2.0.4/sub-01/anat/sub-01_inplaneT2.nii.gz&size=669578
  datalad-archives: dl+archive:MD5E-s2416581890--662e0713d0ce42bcdbadb8251b893b8a.tgz#path=ds001/sub-01/anat/sub-01_inplaneT2.nii.gz&size=669578
  datalad-archives: dl+archive:MD5E-s2426534406--aa96b8184370de7170ab46e1a643c3c0.0raw.tgz#path=ds001_R2.0.0/sub-01/anat/sub-01_inplaneT2.nii.gz&size=669578
  datalad-archives: dl+archive:MD5E-s2426534406--aa96b8184370de7170ab46e1a643c3c0.0raw.tgz/ds001_R2.0.0/sub-01/anat/sub-01_inplaneT2.nii.gz#size=669578
  datalad-archives: dl+archive:MD5E-s2527262329--bd3ea399057c529b37b09dcecec1ca60.0raw.tgz/ds001_R1.1.0/sub001/anatomy/inplane001.nii.gz#size=669578
```

Note that in case of `datalad-archives:` special remote it makes no sense to not specify the only prefix it supports.  So it becomes then - when to ask for all + filter (clearly cannot be more efficient) vs. ask for only specific prefixes.

Also without using a prefix and being able to stop generator without checking all schemes, there would be no benefit from having `gen_URLS` as generator since we would be doomed to first suck in all reponses from annex for a given prefix.

> Plus, wouldn't asking for all the URLs have the advantage of gen_URLS yielding URLs for all supported schemes?

it would yield them all as well in current implementation, although looping through schemas (in order of preference based on prior interactions).

> And even if you did, how does the processing weigh against eliminating additional interactions with annex?
> At any rate, without actual timings it's hard to say anything

Indeed, and for any realistic example it would be heavily masked by actual interactions with the resource (remote urls) making it hard-to-impossible to time.  BUT it already resulted in less pollution of the logs, so even for that it should be of benefit even if performance advantage is in question.

kyleam:
We're talking past each other, but it doesn't matter.  Please go with whatever you want.

yarikoptic:
ok, although I think I might be getting on what you are trying to hint on -- that every request to git-annex for GETURLS has an overhead of first getting all the urls and then filtering for a specified prefix (on git-annex side).  Then indeed, if the remote ends up querying more than a single prefix -- we would just multiply that overhead. Is that what you are talking about? The idea/hope here is that a single prefix hit would be sufficient.  

meanwhile, I wondered how much of penalty from doing multiple queries to annex. Here is an original script performance and then performance of modified one to also do those geturl requests from annex for one or more schemas

<details>
<summary>script from `~datalad/trash/speedyannex2`</summary> 

```shell
$> cat git-annex-remote-datalad
#!/bin/bash

set -e

schemes=( $SCHEMES )
# Gets a VALUE response and stores it in $RET
report () {
    # echo "$@" >&2
    :
}

recv () {
    read resp
    #resp=${resp%\n}
    target="$@"
    if [ "$resp" != "$target" ]; then
        report "! exp $target"
        report "  got $resp"
    else
        report "+ got $resp"
    fi
}

send () {
    echo "$@"
    # report "sent $@"
}

send VERSION 1
recv EXTENSIONS INFO ASYNC
send UNSUPPORTED-REQUEST
recv PREPARE
send PREPARE-SUCCESS
send DEBUG Encodings: filesystem utf-8, default utf-8

# from now on should be checkpresent sequence
# recv CHECKPRESENT MD5E-s70189716--45f4514f972325f481d5c09434c1c94a.nii.gz
while read CMD KEY; do
    if [ $CMD != CHECKPRESENT ]; then
        echo "Was expecting CHECKPRESENT, got $CMD"
        exit 1
    fi
    for scheme in ${schemes[*]}; do
        send GETURLS $KEY $scheme:
        url=bogusforuntil
        while [ ! -z "$url" ]; do
          read resp url
          test "$resp" = VALUE
        done
	done
    send CHECKPRESENT-SUCCESS $KEY
done

```
</details>

```shell
$> time PATH=~datalad/trash/speedyannex:$PATH git annex fsck --from datalad --fast >/dev/null 
PATH=~datalad/trash/speedyannex:$PATH git annex fsck --from datalad --fast >   11.18s user 2.51s system 141% cpu 9.690 total

$> time PATH=~datalad/trash/speedyannex2:$PATH git annex fsck --from datalad --fast >/dev/null
PATH=~datalad/trash/speedyannex2:$PATH git annex fsck --from datalad --fast >  11.75s user 2.74s system 141% cpu 10.241 total

$> time SCHEMES=s3 PATH=~datalad/trash/speedyannex2:$PATH git annex fsck --from datalad --fast >/dev/null
SCHEMES=s3 PATH=~datalad/trash/speedyannex2:$PATH git annex fsck --from   >   25.33s user 5.68s system 123% cpu 25.158 total

$> time SCHEMES="s3 http https ftp" PATH=~datalad/trash/speedyannex2:$PATH git annex fsck --from datalad --fast >/dev/null
SCHEMES="s3 http https ftp" PATH=~datalad/trash/speedyannex2:$PATH git annex   28.21s user 7.41s system 122% cpu 29.180 total
```
</details>
so it does add a little bit (4 seconds added to 25).  Once again, in a realistic case such impact would likely be not notable since it would take longer to actually query the resource.

yarikoptic:
Thank you Kyle. Fwiw, was using this in "production" since then and spotted no side effects

PR #4972, opened 2020-09-29
ENH: add shellcheck run on shell scripts which ATM are shellcheck kosher

yarikoptic:
I find `shellcheck` really handy!

I thought to make this workflow conditioned on changes to those files, but IMHO too much hassle for what it is worth since shellcheck run should be quite fast

### DIFF
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -15,7 +15,7 @@ jobs:
         sudo apt-get install shellcheck
     - name: Run shellcheck
       run: |
-      shellcheck \
+        shellcheck \
           tools/bisect-git-annex.scripts/bisect-git-annex-lock.sh \
           tools/ci/install-annex.sh \
           tools/ci/install-minimum-git.sh \
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -13,6 +13,7 @@ jobs:
       run: |
         sudo apt-get update -qq
         sudo apt-get install shellcheck
+    - uses: actions/checkout@v1
     - name: Run shellcheck
       run: |
         shellcheck \
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -16,6 +16,7 @@ jobs:
     - uses: actions/checkout@v1
     - name: Run shellcheck
       run: |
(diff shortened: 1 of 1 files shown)


yarikoptic:
grr, it didn't run... so I screwed up somewhere but I don't see where and there is no annotation of any kind from the system: https://github.com/yarikoptic/datalad/actions/runs/279238347   may be someone else could see the gotcha

kyleam:
> grr, it didn't run...

Should be fixed now.

yarikoptic:
woohoo - thank you @kyleam !

.github/workflows/shellcheck.yml
```diff
@@ -0,0 +1,24 @@
...
+        sudo apt-get install shellcheck
+    - uses: actions/checkout@v1
+    - name: Run shellcheck
+      run: |
+        shellcheck \
+          tools/bisect-git-annex.scripts/bisect-git-annex-lock.sh \
+          tools/ci/install-annex.sh \
+          tools/ci/install-minimum-git.sh \
+          tools/ci/prep-travis-devel-annex.sh \
+          tools/ci/prep-travis-forssh.sh
```
kyleam, line 24: > which ATM are shellcheck kosher

I think it's worth adding a code comment explaining that you're only testing the subset of scripts that pass shellcheck.

I was curious about what this looked like overall:

```console
$ git grep -l '^#!/.*sh$' tools/ | wc
     29      29     851
$ shellcheck $(git grep -l '^#!/.*sh$' tools/) | grep -c "In tools/"
144
```

.github/workflows/shellcheck.yml
```diff
@@ -0,0 +1,24 @@
...
+
+    steps:
+    - name: Set up system
+      shell: bash
+      run: |
+        sudo apt-get update -qq
+        sudo apt-get install shellcheck
+    - uses: actions/checkout@v1
+    - name: Run shellcheck
+      run: |
```
yarikoptic, line 18:
```suggestion
      run: |
        # I: running only on a subset of scripts which are shellcheck clean ATM
```

.github/workflows/shellcheck.yml
yarikoptic: added a comment, thanks.  Some scripts just need minor fixups to become shellcheck compliant, but I did not embark on that mission

Issue #4968, opened 2020-09-28
addurls: add --ifsubdataset-exists="warn" (or "error", or "")


yarikoptic:
ATM `addurls` would warn if a subdataset already exists.  In the case of using `addurls` to update an existing (super)dataset with hundreds/thousands of subdatasets such warning is not desired since floods the screen with not not pertinent warnings (I know that subdatasets already exist). 
`addurls` already has `-ifexists {overwrite|skip}` for individual files, so I named suggested option to be inline with that. 
Note though that we do have `--existing` in `create_sibling*` and `--missing` in `publish`. IMHO at some point we should harmonize how we name such options (with `if` or not, `exists` vs `existing` etc)

kyleam:
> In the case of using `addurls` to update an existing (super)dataset with hundreds/thousands of subdatasets such warning is not desired since floods the screen with not not pertinent warnings (I know that subdatasets already exist).

Yeah, I think this is one of probably at least a few places where it's clear that addurls wasn't designed with updates in mind.  Avoiding the situation you describe is worth another option in my view.

However, given that it seems using addurls for updates will be a common scneario and I don't think there's a risk of data loss (I'd need to check the handling of unlocked files), perhaps it was a mistake to warn here in the first place, and we should demote the log message to the debug level (or, probably better, yield a "notneeded" result).

What do you think?

yarikoptic:
I would have both logged (DEBUG) and yielded a record - they serve different purposes and might even go to different medium (I started to tend to direct logs into a file, especially for <= DEBUG) for future inspection.  Messages live their own live, can be swallowed, suppressed, combined ("X similar...").

kyleam:
> I would have both logged (DEBUG) and yielded a record - they serve different purposes

With the default `datalad.log.result-level` value, the message of notneeded results is logged at the debug level.  But I agree with your general point that they serve different purposes (e.g., gh-3953).

But, setting aside the specifics of how we drop the warning, you're okay with the idea of getting rid of the warning instead of adding an option like `--ifsubdataset-exists`?

Issue #4929, opened 2020-09-17
addurls: ignores on_failure="stop" setting due to use of  `with_result_progress`


yarikoptic:
My usecase: I have following invocation 
```python
        out = ds.addurls(
            submission_file, '{associated_file}', '{path}',
            exclude_autometa='*',
            fast=fast,
            save=False,
            ifexists="overwrite",
            cfg_proc='nda',
            on_failure="stop",
            # result_renderer='default'
        )
```
which I run on thousands of files.  I do want it to stop as soon as an error encountered since it would make no sense to proceed forward (I want to get from clean state to clean state to save) but after a few minutes I get a first failed reported in progress bar (not sure which one) and it still has hours to go.  The only way is to interrupt but I would not even know what has failed.

The reason is: `@with_result_progress` is "accumulating" all returned or yielded values into the list first and then returns it.  That is what completely prevents correct treatment of `on_failure` in `_process_results`.

I think ideally `@with_result_progress` should keep generator functions to be generator functions **but**, depending on the use case immediate yields could cause `_process_result` rendering of the results thus possibly interfering with the progress bars of `@with_result_progress`.  And `on_failure` is not passed into underlying functions (like `addurls` itself) so they cannot decide one way or another (fail or continue if error).  So I do not see an easy solution here yet.

apparently `@with_result_progress` is used only in addurls ATM within datalad-core:

```shell
$> git grep with_result_progress
datalad/log.py:def with_result_progress(fn, label="Total", unit=" Files"):
datalad/plugin/addurls.py:from datalad.log import log_progress, with_result_progress
datalad/plugin/addurls.py:@with_result_progress("Adding URLs")
datalad/plugin/addurls.py:@with_result_progress("Adding metadata")
```

kyleam:
Thanks for the analysis.  I definitely overlooked this issue when writing `with_result_progress`.

PR #4867, opened 2020-09-03
OPT: addurls - no add per file, mock.patch once per dataset

yarikoptic:
I have been running addurls with `-x '*'` to not bother with adding metadata for this datasets with a notable amount of files (will endup with 2M files across subdataset(s)).  In the "batch" it seems to be only 20k files, and actual adding urls completed quite fast.  Now I had been staring at slowly growing `Adding metadata (10240 skipped):...` so majority were skipped but not due to "no metadata" I think but because `.add` was not needed would be my guess (`with_result_progress` maps "notneeded" into "skipped" and does nohow filters by record).

PR is now a collection of small changes intended to optimize (although didn't test yet either there is a real advantage) `addurls` while running on large(r) number of files and not bothering to add metadata.

One change in behavior would be that I removed `repo.add` which now was ran for every file and then message was perculated up and then at the end in a typical use case `ds.save` would have been called on those files again.  Now, if `save=False`, then those files would remain "annexed" (converted to symlinks) but not added to git repository.  IMHO that is Ok since "staging does no longer exists" in datalad world, and `save=False` was IIRC introduced to facilitate use with `datalad run` which would call `save` anyways.

Let me know what you think, and I will probably test next batch of running addurls on this version to see if it feels any better.

PR #4973, opened 2020-09-29
BF+OPT: addurls - avoid resources (processes) exhaustion in case of having large number of subdatasets

yarikoptic:
Motivation: For ABCD, due to a vast number of subjects (`>1k`) and number of files per each (`>200`) it is desired to adhere to the same ideas as for HCP dataset and create a subdataset for each subject (at least in the `derivatives/*/*`) subfolders. Upon its analysis, `addurls` discovers which subdatasets,  creates (or warns about their existence, see #4968) them, and then proceeds with invoking `AnnexRepo.add_url_to_file` in a batched mode (which is good).  But then each such batched git-annex instantiates about 19 subprocesses (one of which is `git-remote-annex-datalad`) which leads to resources exhaustion, and at some point one of the batched processes doesn't start or perform as expected and the whole run collapses.

IMHO a "resolution" to this is quite straightforward -- group rows by their datasets, and issue `.precommit` when done with any particular to stop all the batched processes for that dataset.repo.  That is what is implemented in this PR.

I went a little further with OPT though after I noticed access to `.repo` per each row.  IIRC that operation is not "fast".  How much it is affecting overall runtime - in case of non `--fast`, any such penalty is probably negligible.  I do not know yet if there is any notable difference, but once again -- why to waste cycles.

Commit messages and comments in the code provide more argumentation and information. 
TODOs
- [x] see how it effects a real run on ABCD - verified that git/git-annex count stays close to "hard to spot" since processes are short lived (spotted that the datalad process itself is quite heavy - 4GB virtual/resident memory.  I guess all the datasets' instances, didn't investigate)
- [ ] decide on either to proceed fixing up progress indication here one way or another, or just move the entire BF against `master` with a more invasive RFing of addurls to avoid unnecessary duplicate actions across `add_urls` and `add_meta` subcalls
- [ ] may be add a unittest which would verify that no more than a single batched addurls exists any given point in time throughout addurls testing.

### DIFF
--- a/datalad/plugin/addurls.py
+++ b/datalad/plugin/addurls.py
@@ -505,7 +505,7 @@ def add_urls(rows, ifexists=None, options=None):
     """
     for row in rows:
         filename_abs = row["filename_abs"]
-        ds, filename = row["ds"], row["ds_filename"]
+        ds, repo, filename = row["ds"], row["repo"], row["ds_filename"]
         lgr.debug("Adding metadata to %s in %s", filename, ds.path)
 
         if os.path.exists(filename_abs) or os.path.islink(filename_abs):
@@ -523,7 +523,7 @@ def add_urls(rows, ifexists=None, options=None):
                 lgr.debug("File %s already exists", filename_abs)
 
         try:
-            out_json = ds.repo.add_url_to_file(filename, row["url"],
+            out_json = repo.add_url_to_file(filename, row["url"],
                                                batch=True, options=options)
         except AnnexBatchCommandError as exc:
             yield get_status_dict(action="addurls",
@@ -548,9 +548,13 @@ def add_meta(rows):
     from unittest.mock import patch
 
     for row in rows:
-        ds, filename = row["ds"], row["ds_filename"]
-        with patch.object(ds.repo, "always_commit", False):
-            res = ds.repo.add(filename)
+        ds, repo, filename = row["ds"], row["repo"], row["ds_filename"]
+        # TODO: in principle in current implementation add_meta can be called
+        # only with rows for the same dataset/repository, so we could have
+        # patch.object oustide of the loop. But ideally it is just that
+        # signature of add_files should change to accept repo (and ds?) arg
+        with patch.object(repo, "always_commit", False):
+            res = repo.add(filename)
             res_status = 'notneeded' if not res \
                 else 'ok' if res.get('success', False) \
                 else 'error'
@@ -565,7 +569,7 @@ def add_meta(rows):
             )
 
             lgr.debug("Adding metadata to %s in %s", filename, ds.path)
-            for a in ds.repo.set_metadata_(filename, add=row["meta_args"]):
+            for a in repo.set_metadata_(filename, add=row["meta_args"]):
                 res = annexjson2result(a, ds, type="file", logger=lgr)
                 # Don't show all added metadata for the file because that
                 # could quickly flood the output.
@@ -895,6 +899,7 @@ class Addurls(Interface):
                 ds_current = Dataset(os.path.join(ds.path, subpath))
             else:
                 ds_current = ds
+            repo = ds_current.repo  # OPT: relatively expensive, pass along
 
             for row in ds_rows:
                 # Add additional information that we'll need for various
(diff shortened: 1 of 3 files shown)


yarikoptic:
Oh well, observing real life performance brings me to the realization that there is `@with_result_progress` in use for internal `add_urls`.  So I need to RF to either

1. not use it and use log_progress within outside loop (which would then also address #4929 as well), or
2. bring groupping inside the add_urls... Actually I had already done a similar act for add_meta in https://github.com/datalad/datalad/pull/4867 which I had proposed as a pure OPT thus against `master`. And indeed it is just OPT since no batched `annex metadata` is yet implemented in datalad but I guess if I decide to go with 1., that change would become obsolete

@kyleam - what do you think?

yarikoptic:
eh, and overall -- in reality it doesn't help my usecase, since I still see a growing number of git-annex calls, so `.precommit` somehow ... d'oh, I did `ds.repo.precommit()` instead of `ds.repo.precommit()`, damn, will fix

yarikoptic:
ok, fixed that dumb bug with `ds`. For fun of it I think I should have also addressed #4929 in the decorator itself -- I do like its convenience, although we do not use it anywhere besides `addurls` .

I will just test either fix is effective on abcd, disregarding suboptimal progress reporting, and then we will decide on either to proceed with BFing `maint` further or just proceed by BF+RFing `master` addurls one way or another.  I do not oppose idea also to rebase a copy of #4867 against `maint` and doing BF+RFing right here ;)

kyleam:
The changes here look fine to me.  Proceed however you'd like.

datalad/plugin/addurls.py
```diff
@@ -520,7 +523,7 @@ def add_urls(rows, ifexists=None, options=None):
                 lgr.debug("File %s already exists", filename_abs)
 
         try:
-            out_json = ds.repo.add_url_to_file(filename, row["url"],
+            out_json = repo.add_url_to_file(filename, row["url"],
                                                batch=True, options=options)
```
kyleam, line 527:
```suggestion
                                            batch=True, options=options)
```

datalad/plugin/addurls.py
```diff
@@ -891,26 +884,60 @@ def __call__(dataset, urlfile, urlformat, filenameformat,
...
         files_to_add = set()
-        for r in add_urls(rows, ifexists=ifexists, options=annex_options):
-            if r["status"] == "ok":
-                files_to_add.add(r["path"])
-            yield r
+        # We need to group by dataset since otherwise we will initiate
+        # batched annex process per each subdataset, which might be infeasible
+        # in any use-case with a considerable number of subdatasets.
+        # TODO: logical spot for parallel run across (sub)datasets
+        for subpath, ds_rows in itertools.groupby(rows, itemgetter("subpath")):
```
yarikoptic, line 891: eh Yarik -- RTFM https://docs.python.org/3/library/itertools.html#itertools.groupby "Generally, the iterable needs to already be sorted on the same key function.".  That is why current overall slowness and progressbars virtually never being present -- just keeps adding only few files at a time... not good

kyleam: I changed the rows to be sorted in ba05a61535.  Please of course change/drop that if that's not the right fix.

yarikoptic: Thanks! unfortunately `itemgetter` is not be enough (not that I figured it out, the script has crashed ;)) since for top level dataset it would be `None` and sorted would fail, thus we either need a tuned up getter or store empty string not None

kyleam: Sorry, I should have considered that (or at least noted the failing tests).

> thus we either need a tuned up getter or store empty string not None

I've updated the key function in 9d8e58d.  Tests pass for me locally.

yarikoptic: THANK YOU! I will give it a shot

yarikoptic:
I don't like "per dataset" progress bar -- gives no feedback on how long to wait... need to RF that to make this considered for merging.

yarikoptic:
I think I might just fold RF into #5022 and parallelize (across datasets) altogether.  That might also reduce runtime memory consumption impact which is now the case (when there are thousands of subdatasets kept in memory)

yarikoptic:
I am considering it superseded by #5022 which already addressed what was tried to be addressed in this PR but more radically (thus against `master`)

PR #4974, opened 2020-09-30
install-annex.sh: Support installation on macOS

### LINKED ISSUE #4971: install-annex.sh: Add OSX installation for "autobuild", "snapshot", and "datalad-extensions-build" options
ATM https://github.com/datalad/datalad/blob/master/tools/ci/install-annex.sh is Linux specific, but now that we even have datalad-extensions built .dbg it could also support OSX.  I think it would be nice if it would internally (no dedicated new option) download appropriate file (.dmg for OSX) and install it accordingly.

Ideally `--help` could either be limited only to the options supported on current OS (but then a bit painful since some are Debian specific), or at least annotate, e.g.

```
    "autobuild  # Linux, OS X"
    "conda-forge [version]  # Linux"
    "conda-forge-last [version]  # Linux"
    "datalad-extensions-build  # Debian-based Linux, OS X"
    "deb-url URL  # Debian-based Linux"
    "neurodebian  # NeuroDebian-based Linux"
    "neurodebian-devel # NeuroDebian-based Linux"" 
    "snapshot   # Linux, OS X"
```

NB "snapshot" is somewhat a bad name.  It corresponds to "current" (release?) for both Linux and OSX

jwodder:
Closes #4971.

### DIFF
--- a/.github/workflows/test_macos.yml
+++ b/.github/workflows/test_macos.yml
@@ -23,7 +23,8 @@ jobs:
 
     - name: Install git-annex
       run: |
-        eval source tools/ci/install-annex.sh --adjust-bashrc ${{ matrix.install_scenario }}
+        eval source tools/ci/install-annex.sh ${{ matrix.install_scenario }}
+        echo "::set-env name=PATH::$PATH"
 
     - name: Set up Python 3.6
       uses: actions/setup-python@v1
--- a/.github/workflows/test_macos.yml
+++ b/.github/workflows/test_macos.yml
@@ -7,18 +7,22 @@ on:
 jobs:
   test:
     runs-on: macos-latest
+    strategy:
+      matrix:
+        install_scenario:
+          - brew
+          - snapshot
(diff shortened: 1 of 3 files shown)


yarikoptic:
awesome. To test that it works for OSX (we use it on travis for linux so there it is used/tested), could you please tune OSX github workflow to use it?  which reminds me that you just used brew, so may be "brew" install target could be added (OSX specific) and then a matrix on the `DL_ANNEX_INSTALL_SCENARIO` over "brew" and "snapshot" added to OSX workflow just to call the script as we do for travis/Linux.   That would expand our testing of most common installation scenarios on OSX and will test this script -- win/win

jwodder:
@yarikoptic Done.

yarikoptic:
GREAT, Thank you @jwodder!

FTR: shellcheck is happy. I don't see how crawler or appveyor could get affected, so I will proceed with merge

Issue #4981, opened 2020-10-04
Approach to streamline credential handling


mih:
Credential handling is a bit of dark magic in DataLad, although the foundations are very flexible and should be capable of dealing with any scenario that I can come up with. Previous plans to replace it have not led to any results (e.g. #1783). However, there are long-standing limitations like the lack of a modification API (#396), and priorities (#322).

The biggest issue right now (from my POV) is the specification of credentials in non-interactive environments without a credential store being easily available (HPC/HTC). The only practical approach is to supply them via ENV variables. How these must be named is undocumented and knowledge that is only discoverable by the initiated.

The last attempt to do something in that direction was #4135 (stalled).

I propose to adjust credential handling to:

- use `ConfigManager` instead of direct environment variable inspection (this will open up many standard ways to supply credentials that are all better documented then the present one) -- such a PR should replace #4135
- document a predictable credential (variable) naming scheme
- keep using python-keyring whenever the environment doesn't have what is needed
- address #396 by implementing a datalad command that can at least list, and delete credentials

yarikoptic:
I agree, we need to overhaul. Just a quick note for now: I use https://github.com/datalad/datalad/wiki/FAQ#how-to-configure-insecure-local-credentials-store on hpc etc

mih:
> * address #396 by implementing a datalad command that can at least list, and delete credentials

So this turns out to be a major issue. `keyring` has no support for listing known secrets. We would need to interface with individual backends for that. Simple for `SecretService`, more or less a reimplementation for `PlaintextKeyring` (to name just two).

Without being able to list secrets it is near-impossible to delete secrets (#3126), because their naming is completely hidden from users. In order to filter relevant secrets from `SecretStorage` I have to know that good candidates are ones with this property

`'application': 'Python keyring library'`

and have a `service` property, and it starts with `datalad-`. On my system that yields:

```
 'datalad-https://nmrxnat....de:443/xnat',
 'datalad-datalad-test-s3',
 'datalad-https://nmrxnat....de:443/xnats',
 'datalad-hcp-s3',
 'datalad-github',
 'datalad-crcns',
 'datalad-https://nmrxnat.....de:443/xnat',
 'datalad-nmrxnat.ime....de:443/xnat',
 'datalad-gitadshub',
 'datalad-hcp-s3',
 'datalad-nmrxnat.....de:443/xnat',
 'datalad-github',
 'datalad-datalad-test-s3',
[...]

PR #4982, opened 2020-10-04
ENH: Show info on credential setup in wtf()

mih:
Would look something like this (paths truncated):

```
 % datalad wtf -S credentials
 # WTF
 ## credentials
   - keyring:
     - active_backends:
       - SecretService Keyring
       - PlaintextKeyring with no encyption v.1.0 at ...keyring_pass.cfg
     - config_file: ....config/python_keyring/keyringrc.cfg
     - data_root: ....local/share/python_keyring
```

If got tired of looking up the code snippets in the keyring docs to
run when switching machines, and discovering the left-overs of previous
configuration attempts.

Related to gh-4981

### DIFF
--- a/datalad/plugin/wtf.py
+++ b/datalad/plugin/wtf.py
@@ -24,6 +24,7 @@ from datalad.utils import (
     ensure_unicode,
     getpwd,
     unlink,
+    Path,
 )
 from datalad.dochelpers import exc_str
 from datalad.support.external_versions import external_versions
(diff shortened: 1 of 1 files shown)


datalad/plugin/wtf.py
```diff
@@ -261,6 +262,34 @@ def _describe_location(res):
     }
 
 
+def _describe_credentials():
+    # might later add information on non-keyring credentials
+    props = {}
+    try:
+        import keyring
+        from keyring.util import platform_
+    except ImportError:
```
yarikoptic, line 271: since keyring is part of `install_requires`, I think ImportError for `import keyring` must not happen (silently).  Not sure about `keyring.util.platform_` since not sure if part of "public API".  In either case, I would have logged a warning or at least a DEBUG message so later we could figure out WTF there is no keyring information when expected to be

mih: Ah, good point! I removed the try-except handling. `platform_` is a very visible part of the API, I don't think it needs any more special handling.

Issue #3360, opened 2019-04-29
`cfg_text2git` not functional on Windows


mih:
This procedure is one of the central use cases behind https://github.com/datalad/datalad/pull/3353, but it is not working on windows, as the git-annex build lacks support of MIME types.

```
bad annex.largefile configuration: Parse failure: "mimetype" not supported; not built with MagicMime support
```

This looks as if any dataset configured in such way would not be operational on Windows.

yarikoptic:
@joeyh I think we have touched briefly on the possibility to empower windows build with libmagic, or was there some showstopper?

yarikoptic:
crap... I feel like we with @joeyh  had a brief "discussion" on why there is difficulty with libmagic for windows, but can't find a trace of it, damn. So I guess you @mih just drink "pure annex" for everything in your windows world?

mih:
I think we should update the procedure to act differently on windows, or provide a different procedure that acts differently. @adswa has outlined a possible approach in https://github.com/datalad-handbook/book/pull/532

adswa:
Windows requires file extensions, correct? Would it make sense to have ``text2git`` define a long list of text file file extensions as ``largefiles=nothing`` if we detect that we are on Windows?

mih:
I don't think we should use the same procedure: While a dataset might be created on linux (and `text2git` working), it will still fail later when used on windows. And when `text2git` does something different on windows initially, it would impair all linux users. A separate procedure would make the separation clear. We might also consider stopping to ship `text2git` entirely, because it makes datasets less portable.

yarikoptic:
I think the best investment of time would be to actually address the problem in the root of it, by providing libmagic build for windows in conda-forge, see https://github.com/conda-forge/file-feedstock/issues/3 (libmagic feedstock is archived but has no reference to migration over to file-feedstock, took time to find it).

yarikoptic:
I would invite you to join fun on https://github.com/nscaife/file-windows/pull/9  where I am trying to cross-build libmagic... if we succeed, I think it it might pave the way to build it on conda-forge

mih:
https://github.com/nscaife/file-windows/pull/9 was merged. Can you briefly summarize the state of the remaining effort, or link to a description? Thx!

mih:
@adswa Maybe we should start a parallel effort to avoid expensive workarounds like the one you had to describe in http://handbook.datalad.org/usecases/ml-analysis.html

[...]

PR #4983, opened 2020-10-05
Text2git: Largefiles = nothing rules for common textfiles if on Windows

adswa:
This adds a workaround for #3360. Its fairly naive, and I'm happy to hear suggestions how to approach it in a more clever way. But it works like this (tested on our windows machine).

Instead of applying the mimenecoding based configuration of text2git, this writes a bunch of largefiles = nothing rules for common textfile extensions if ``text2git`` is used on Windows.

Downsides:
- every extension would need to be specified
- Datasets aren't portable to Windows if created on a non-Windows machine (but this was true before this change, anyway)

Todo:
- does this need a test?
- are there other common text file extensions I have missed?

### DIFF
--- a/datalad/resources/procedures/cfg_text2git.py
+++ b/datalad/resources/procedures/cfg_text2git.py
@@ -2,6 +2,7 @@
 
 import sys
 import os.path as op
+import platform
 
 from datalad.distribution.dataset import require_dataset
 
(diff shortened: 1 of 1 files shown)


yarikoptic:
Since, as you mentioned, it matters not only on which platform dataset is created, but on which (a clone of) is used, I don't think we should have platform dependent .gitattributes creation. We would either need to see git annex built with magic support on windows (attn @joeyh ) or switch to use such extensive listing in general across platforms.

mih:
> or switch to use such extensive listing in general across platforms.

I am OK with that.

yarikoptic:
it would require "migration" for already existing datasets.  Extensive list would indeed work for majority of use cases, and even would be beneficial in some cases of huge text files (xml, json) used as a container for data (IIRC some files produced by fmriprep etc), but I think it is a suboptimal solution overall.
I will ping joey again about [existing issue](https://git-annex.branchable.com/todo/provide_windows_build_with_MagicMime/#comment-36e31d0ae3d55820482ce7e35badd70c) - may be he would look into building with magic support for windows.  Partial problem was that the build machine for windows is not managed by joey.  So, ideally, some public CI based setup should be done to facilitate.

mih:
I can only partially follow that reasoning. But ok. Until this happens, we can put the new approach into a new procedure `textfiles_in_git`, replace all references to `text2git` with the new one, delete the pages with workaround descriptions from the handbook. That buys us an infinite amount of time, avoids any migration needs, and makes it easier for new users. Overall win.

yarikoptic:
- a dedicated procedure sounds good to me (e.g. `cfg_text2git_explicit` or `cfg_text2gitE` or something else but probably with `text2git` as its prefix)
  - could it also gain an option for "conversion" from regular `text2git`?
- `cfg_text2git` could check if git-annex is built with magic support (e.g. `wtf` already parses those, it is MagicMime), and if not -- kaboom with an informative message that annex without libmagic support is used, either get your hands on git-annex with libmagic or use `cfg_text2gitE` procedure instead.

pros:
- It would make it explicit
- we would need to do nothing whenever annex built with magic support appears
- we alert when annex unexpectedly is built without (happened with some conda builds)

cons:
- not sure how "stable" `MagicMime` marker is (@joeyh ?) or might get renamed at some point

mih:
My argument was very much going into the exact opposite direction. I would like to avoid any kind of further complication of a fix to #3360. I would like to have something practical that avoids the workarounds that are forced on people now. I would like to not have to think about transitions back and forth, or implications of initiating platform capabilities vs. consuming platform capabilities.

We need to keep the cognitive load in the handbook low, hence we should strive to avoid complex conditionals as much as we can. Establishing mimetype-based configurations on windows has not worked out over the past 1.5 years. I think it is time to provide a functional solution.

Given that this proposal is semantically different from `text2git` and does not need to ever evolve back to `text2git`, it makes sense to name it differently, and not advertise it as an alternative to `text2git`. If anything, `text2git`, being the platform-constrained solution, is the limited alternative for some special cases, where people should think carefully, if they want to employ it.

If `textfiles_in_git` is too similar to `text2git` I propose `no_annex_for_textlike`. However, I doubt that there will be much confusion, once `text2git` is no longer mentioned in the handbook.

Introducing such a procedure has no negative impact on the future of `text2git` AFAICS. Consequently, such considerations should no longer delay a solution.

yarikoptic:
It seems you are arguing only for what name should it be, and otherwise we agree... given the nature of the "method" of a hardcoded selection of extensions, I think that fact should be reflected in the name. `textlike` and `textfiles` does not reflect that and too ambiguous to me. Long names are IMHO also suboptimal.  And even if we "don't advertise" it, it is an alternative -- there are multiple ways to decide "is this file likely to be a text file", and that is why having some symmetry in the names would IMHO be good instead of having related procedures named (and sorted) completely differently. I would have preferred something like `textext2git`.  Whatever the handbook uses is Ok with me (since I have little to no say in that), but I would have made it explicit that there are multiple ways to identify "text files", and list those (two) procedures, while saying that throughout the book you would use the `X` because of `Y`.
 
BTW extensions based decision making would be useful in cases like fmriprep output which has large number of large text (IIRC XML) files, so overall would be useful beyond "for Windows sake".

Re "long-term": I do believe that building mime library for Windows is possible, and might happen in the nearest future (could happen faster if people working with Windows gave it a try).  Meanwhile, @jwodder ATM is working on establishing build of git-annex within datalad-extensions (https://github.com/datalad/datalad-extensions/pull/40)  which would be the stepping stone towards that.  If we happen to get annex build with magic for windows, then whatever procedure you choose to be the "flagman" of the handbook, the decision would be not just merely "to make Windows happy" but because of a more solid argument.

yarikoptic:
As git-annex with magic for windows is coming (well -- it is at http://datasets.datalad.org/datalad/packages/windows/ if you want to try it), let's just close this one?  I would still like to see some `cfg_text2git_ext` or alike, since it might be preferable in some cases as I have mentioned, so may be not "close" but "RF"? ;)

adswa:
I assume this can be closed with https://github.com/datalad/datalad/issues/3360#issuecomment-721495368 as soon as we have documented how users can get their hands on a git-annex version with mimeencoding :-)

adswa:
I will close this in favor of an issue about the need for documentation on the git-annex version with MagicMime.

PR #4985, opened 2020-10-05
BF: get_versioned_url - support anonymous acces (no provider) and buckets with . in name

yarikoptic:


kyleam (approved):

yarikoptic:
and thank you @kyleam for the review

PR #4994, opened 2020-10-08
RF: ssh connector code uses WitlessRunner instead of Runner

mih:
And just that. Related, but orthogonal: #4996


mih:
This looks good, but moving patch into #4996 to have it all in a single spot.

PR #4996, opened 2020-10-08
Runner -- shot of mercy

mih:
All Runner code is now hanging on a thin thread. The remaining connections are:

- `GitRepo._run_git_custom_command` is still on life support but already brain dead
- `run` still uses `Runner` -> #5002
- [x] `create_sibling` has a fancy `Runner` adaptor class that adds a few functions that partially duplicate functionality in `SSHConnection` and want to become command abstractions (although that effort #4068 died a lonely death, @bpoldrack )

Significant changes:
- WitlessRunner now supports shell-based execution https://github.com/datalad/datalad/pull/4996/commits/df4f782bd8a30693cb5a28665ad4e41eb82c9907

### DIFF
--- a/datalad/core/local/tests/test_diff.py
+++ b/datalad/core/local/tests/test_diff.py
@@ -21,7 +21,10 @@ from datalad.support.exceptions import (
 )
 
 from datalad.consts import PRE_INIT_COMMIT_SHA
-from datalad.cmd import GitRunner
+from datalad.cmd import (
+    GitWitlessRunner,
+    StdOutCapture,
+)
 from datalad.utils import (
     Path,
     on_windows,
@@ -61,8 +64,10 @@ def test_magic_number():
     # commit
     # given the level of dark magic, we better test whether this stays
     # constant across Git versions (it should!)
-    out, err = GitRunner().run('cd ./ | git hash-object --stdin -t tree')
-    eq_(out.strip(), PRE_INIT_COMMIT_SHA)
+    out = GitWitlessRunner().run(
+        ['git', 'hash-object', '-t', 'tree', '/dev/null'],
+        protocol=StdOutCapture)
+    eq_(out['stdout'].strip(), PRE_INIT_COMMIT_SHA)
 
 
 @with_tempfile(mkdir=True)
--- a/datalad/core/local/tests/test_diff.py
+++ b/datalad/core/local/tests/test_diff.py
@@ -65,7 +65,7 @@ def test_magic_number():
     # given the level of dark magic, we better test whether this stays
     # constant across Git versions (it should!)
     out = GitWitlessRunner().run(
-        ['git', 'hash-object', '-t', 'tree', '/dev/null'],
+        'cd ./ | git hash-object --stdin -t tree',
         protocol=StdOutCapture)
     eq_(out['stdout'].strip(), PRE_INIT_COMMIT_SHA)
 
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -1319,12 +1319,9 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
             cmd.extend(git_options)
         cmd += ["rev-parse", "--show-toplevel"]
         try:
-            toppath, err = GitRunner().run(
-                cmd,
-                cwd=path,
-                log_stdout=True, log_stderr=True,
-                expect_fail=True, expect_stderr=True)
-            toppath = toppath.rstrip('\n\r')
+            out = GitWitlessRunner(cwd=path).run(
+                cmd, protocol=StdOutErrCapture)
+            toppath = out['stdout'].rstrip('\n\r')
         except CommandError:
             return None
         except OSError:
--- a/datalad/tests/test_utils.py
+++ b/datalad/tests/test_utils.py
@@ -111,6 +111,7 @@ from .utils import (
     ok_file_has_content,
     ok_generator,
     ok_startswith,
+    on_travis,
     probe_known_failure,
     skip_if,
     skip_if_no_module,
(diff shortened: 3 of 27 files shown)


datalad/core/local/run.py
```diff
@@ -454,24 +454,16 @@ def format_command(dset, command, **kwds):
...
-            log_online=True,
-            # not yet sure what we should do with the command output
-            # IMHO `run` itself should be very silent and let the command talk
-            log_stdout=False,
-            log_stderr=False,
-            expect_stderr=True,
-            expect_fail=True,
-            # TODO stdin
+            # command is always a string
+            ['sh', '-c', command],
```
yarikoptic, line 466: why not `/bin/sh`? (unlikely any silly soul gets `sh` into their PATH but you never know)... moreover subprocess in python uses `/bin/sh` according to docs:
```
   On POSIX with ``shell=True``, the shell defaults to :file:`/bin/sh`.  If
```

datalad/core/local/tests/test_diff.py
```diff
@@ -61,8 +64,10 @@ def test_magic_number():
     # commit
     # given the level of dark magic, we better test whether this stays
     # constant across Git versions (it should!)
-    out, err = GitRunner().run('cd ./ | git hash-object --stdin -t tree')
-    eq_(out.strip(), PRE_INIT_COMMIT_SHA)
+    out = GitWitlessRunner().run(
+        ['git', 'hash-object', '-t', 'tree', '/dev/null'],
```
yarikoptic, line 68: *nix only? or if there is a pipe `|` there is a guarantee to have `/dev/null`?

yarikoptic (changes requested):
only cursory review :
- requested `sh` -> `/bin/sh`
- FTR: ATM no negative (or positive) impact on benchmarks, so -- good

datalad/core/local/tests/test_diff.py
mih: Yes, this test did not and does not run on windows.

datalad/core/local/run.py
mih: Yes, that was the point here. `sh` should work fine on unix, but also works on windows. I will have to check whether `/bin/sh` would work too, or whether switching logic needs to be introduced.

In any case, this change would need to be done throughout the code base and not just in this PR.

mih: Confirmed that `/bin/sh` does **not** work on windows/conda

mih: I have implemented platform-shell execution support for WitlessRunner instead. https://github.com/datalad/datalad/pull/4996/commits/95f84a5dcd894b5ca05fea0bb72330be3c941af8

datalad/core/local/tests/test_diff.py
mih: Good catch. I copied this too blindly from the tests of the old `diff`, where the test isn't cross-platform. Fixed.

mih:
Ok, so this works in principle. Playing with the outcome, I think we could delete 2k lines of, now unused, code. However, the Runners are still not dead. The culprit after this PR is now `AnnexRepo._run_annex_command()` which has a range of complications and compatibility "hacks" hanging off of it.

kyleam (commented):
Thanks for doing this.  It looks like quite the slog.  Looking through (at f7af7ecdf0), I didn't spot any problematic conversions, and -core's tests look happy.

I scratched my head at the -container failure (and associated PR) for a bit.  If I understand it correctly, the reason that `swallow_outputs` works for `Runner` and not `WitlessRunner` is that `Runner` explicitly sets the `stdout` argument of `Popen` to `sys.stdout`, which allows the `swallow_outputs` override to be effective.

bpoldrack (approved):
LGTM - nothing that smells like a trap to me ;-)

yarikoptic:
Thanks @kyleam  for the analysis - I think that is indeed what is happening, and probably is a "feature". So https://github.com/datalad/datalad-container/pull/131 should be finalized, merged, and new datalad-container released (we could merge this before but then we would end up closing our eyes on the -container failures in master possibly missing new regressions)

datalad/support/gitrepo.py
```diff
@@ -1324,12 +1324,9 @@ def get_toppath(cls, path, follow_up=True, git_options=None):
             cmd.extend(git_options)
         cmd += ["rev-parse", "--show-toplevel"]
         try:
-            toppath, err = GitRunner().run(
-                cmd,
-                cwd=path,
-                log_stdout=True, log_stderr=True,
-                expect_fail=True, expect_stderr=True)
-            toppath = toppath.rstrip('\n\r')
+            out = GitWitlessRunner(cwd=path).run(
```
yarikoptic, line 1327: a note: coverage marked this as not-covered for me -- doesn't make sense since it would be the first line in `try:` and prior lines are covered.

kyleam:
> I think that is indeed what is happening

Just to lend more concrete support to my explanation:

<details>
<summary>demo</summary>

```python
import subprocess as sp
import sys

from datalad.cmd import Runner
from datalad.cmd import WitlessRunner
from datalad.utils import swallow_outputs

def old_runner():
    Runner().run(["pwd"], log_online=True,
                 log_stdout=False,
                 log_stderr=False)

def new_runner():
    WitlessRunner().run(["pwd"])

def popen():
    p = sp.Popen(["pwd"])
    p.communicate()

def popen_with_sys_stdout():
    p = sp.Popen(["pwd"], stdout=sys.stdout)
    p.communicate()

for fn in [old_runner, new_runner, popen, popen_with_sys_stdout]:
    print(f"===== {fn.__name__}")
    with swallow_outputs() as out:
        fn()
        x = out.out
    print("----- swallowed")
    print(x)
```

```
===== old_runner                  
----- swallowed
/tmp

===== new_runner
/tmp
----- swallowed

===== popen
/tmp
----- swallowed

===== popen_with_sys_stdout
----- swallowed
/tmp

```

</details>

mih:
Given that this PR isn't achieving its goal (total removal of Runner) anyways, I am OK with stripping the change to `run`.

datalad/tests/test_utils.py
```diff
@@ -1305,8 +1316,17 @@ def get_interactive(py_pre="", **run_kwargs):
...
+    # happen, also test the core protocols
+    # (we can only be interactive in a runner, if the test execution
+    # itself happens in an interactive environment)
+    for proto, interactive in ((NoCapture, is_interactive()),
+                               (KillOutput, False),
+                               (StdOutErrCapture, False),
+                               (GitProgress, False),
+                               (AnnexInitOutput, False),
+                               (AnnexJsonProtocol, False)):
+        eq_(get_interactive(protocol=proto), interactive)
```
mih, line 1329: I am not sure what is happening on travis with this test. https://travis-ci.org/github/datalad/datalad/jobs/734538506#L2302 says that while the test execution itself considers `is_interactive() == False`, a process executed by the test is `is_interactive() == True`. This seems wrong. I hand-verified that the logic works on my Debian system, and on Win10.

mih: Added a special case handling for travis.

kyleam (approved):

PR #5001, opened 2020-10-09
Win/BF: don't invoke 'git-annex' but 'git annex'

### LINKED ISSUE #4892: error with datalad install
It is the first time for me to use datalad to download data~ And here is the error info in cmd. 
E:\OPT\research\fMRI\test>datalad install https://github.com/OpenNeuroDatasets/ds002372.git
[ERROR] [WinError 2] The system cannot find the file specified [subprocess.py:_execute_child:1178] (FileNotFoundError)

adswa:
On a Windows machine, while ``git-annex`` and ``git annex`` invocations work fine from an Anaconda prompt, a "git-annex" command invoked via ``run_annex_command`` fails to find the git-annex executable.
When using 'git annex', though, commands seem to work just fine. This change fixes #4892. I lack an in-depth understanding of why that is, yet, but I also figured that this change doesn't do harm to anything else.

### DIFF
--- a/datalad/customremotes/archives.py
+++ b/datalad/customremotes/archives.py
@@ -424,7 +424,7 @@ class ArchiveAnnexCustomRemote(AnnexCustomRemote):
         def progress_indicators(l):
             self.info("PROGRESS-JSON: " + l.rstrip(os.linesep))
 
-        self.runner(["git-annex", "get",
+        self.runner(["git", "annex", "get",
                      "--json", "--json-progress",
                      "--key", akey
                      ],
--- a/datalad/customremotes/base.py
+++ b/datalad/customremotes/base.py
@@ -652,7 +652,7 @@ class AnnexCustomRemote(object):
         # TODO: should actually be implemented by AnnexRepo
         #       Command is available in annex >= 20140410
         (out, err) = \
-            self.runner(['git-annex', 'contentlocation', key], cwd=self.path)
+            self.runner(['git', 'annex', 'contentlocation', key], cwd=self.path)
         # TODO: it would exit with non-0 if key is not present locally.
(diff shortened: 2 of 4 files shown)


mih (approved):
I agree, this should do not harm. But I also lack insight on why it would make a difference.

kyleam (commented):
As you and @mih say, I don't see how this would be problematic, and there are places in the code (including a couple of non-test spots) that already use the separated variant that this PR switches to.

A few comments:

  * Grepping for `["']git-annex["']` is noisy, but I see at least a few other instances that might trigger the same issue (e.g., a `["git-annex", "get", ...]` in datalad/customremotes/archives.py).

  * Since it's not obvious what's going on here, it'd be nice to add a code comment in `_run_annex_command` that mentions the importantance of the separated form.

  * I think this could go to maint.

adswa:
Thanks much for your comments, Kyle! I think I have applied them, and I hope I caught all git-annex command calls.I rebased onto maint.

kyleam (commented):
Thanks for the updates.

I'm going to restart the failing macOS build to see if that failure goes away.  (If it does, I'll open a separate issue for the apparently intermittent failure.)

https://github.com/datalad/datalad/pull/5001/checks?check_run_id=1237765992

edit: restarted build passed

https://github.com/datalad/datalad/pull/5001/checks?check_run_id=1242140615

kyleam:
> (If it does, I'll open a separate issue for the apparently intermittent failure.)

Just to follow up on this: I was about to do that now, but I don't recall seeing it elsewhere, and it looks to be an internal git-annex locking issue, so I'll wait to see if it pops up again before opening an issue that nobody will be able to act on.

PR #5002, opened 2020-10-10
RF: Stop using old-Runner in `run`

mih:
Instead of switching to `subprocess.run()` (original plan), I still use a runner in order to retain the CommandError raise that might be important to consumers, and simultaneously avoid having to duplicate the CommandError construction logic.

This used to be part of #4996, but has been placed into a standalone PR due to the complications of https://github.com/datalad/datalad-container/pull/131 -- and caused by a no-log behavior of `run` introduced by this PR. Executed commands communicate directly with the terminal.

### DIFF
--- a/datalad/core/local/run.py
+++ b/datalad/core/local/run.py
@@ -454,24 +454,16 @@ def format_command(dset, command, **kwds):
 
 
 def _execute_command(command, pwd, expected_exit=None):
-    from datalad.cmd import Runner
+    from datalad.cmd import WitlessRunner
 
     exc = None
     cmd_exitcode = None
-    runner = Runner(cwd=pwd)
+    runner = WitlessRunner(cwd=pwd)
     try:
         lgr.info("== Command start (output follows) =====")
         runner.run(
-            command,
-            # immediate output
-            log_online=True,
-            # not yet sure what we should do with the command output
-            # IMHO `run` itself should be very silent and let the command talk
-            log_stdout=False,
-            log_stderr=False,
-            expect_stderr=True,
-            expect_fail=True,
-            # TODO stdin
+            # command is always a string
+            command
         )
     except CommandError as e:
         # strip our own info from the exception. The original command output


yarikoptic:
appveyor not happy:
```
ERROR: datalad.interface.tests.test_rerun.test_run_failure
  File "C:\Miniconda36\envs\test-environment\lib\site-packages\nose\case.py", line 198, in runTest
    self.test(*self.arg)
  File "C:\Miniconda36\envs\test-environment\lib\site-packages\datalad\tests\utils.py", line 731, in _wrap_with_tempfile
  File "C:\Miniconda36\envs\test-environment\lib\site-packages\datalad\interface\tests\test_rerun.py", line 316, in test_run_failure
    eq_('x \n' if on_windows else 'x\n', open(outfile).read())
FileNotFoundError: [Errno 2] No such file or directory: 'C:\\Users\\appveyor\\AppData\\Local\\Temp\\1\\datalad_temp_x3fwio0e\\sub\\grows'
```

kyleam:
There was a -container release, so this should be unblocked on that front.  I don't have an good guesses about the appveyor failure.

mih:
Pushed a rebase on current master to see what the new appveyor setup has to say now.

mih:
I can replicate the following failure on a real win10 box

```
FAIL: datalad.core.local.tests.test_run.test_run_path_semantics
  File "C:\Miniconda37\lib\site-packages\nose\case.py", line 198, in runTest
    self.test(*self.arg)
  File "C:\Miniconda37\lib\site-packages\datalad\tests\utils.py", line 755, in _wrap_with_tempfile
  File "C:\Miniconda37\lib\site-packages\datalad\core\local\tests\test_run.py", line 460, in test_run_path_semantics
    ok_exists(op.join(ds0.path, "two"))
  File "C:\Miniconda37\lib\site-packages\datalad\tests\utils.py", line 521, in ok_exists
    assert Path(path).exists(), 'path %s does not exist (or dangling symlink)' % path
AssertionError: path C:\Users\appveyor\AppData\Local\Temp\1\datalad_temp_dbip_eio\ds0\two does not exist (or dangling symlink)
```

instead of creating the file that the desired location, it places them into the CWD but all messed-up.

```
(base) C:\Users\mih\AppData\Local\Temp\datalad_temp_gzvr856m\ds1\s1>ls
C:UsersmihAppDataLocalTempdatalad_temp_gzvr856mds0two
```

It should have written to `C:\Users\mih\AppData\Local\Temp\datalad_temp_gzvr856m\ds0\two`, so somehow all the backslashed got mangled.

When I execute the command reported to run, manually in a windows shell (including the `sh -c`):

```
Async run ['sh', '-c', 'cd .> C:\\Users\\mih\\AppData\\Local\\Temp\\datalad_temp_gzvr856m\\ds0\\two']
```

I get the desired outcome. Puzzled....

mih:
I pushed a number of fixes that simplify the test setup (uniform use of POSIX commands, because we have them in a standard datalad installation on windows too).

However, there is a problem in `run()` or `WitlessRunner`, not sure. Whenever `.format()` is used with windows paths, the backslashes get mangled, leading to changes in target filenames that make the tests fail, although the actual functionality works.

I left `test_rerun:test_run_failure` in this state to have something to debug.

datalad/core/local/run.py
```diff
@@ -454,24 +454,16 @@ def format_command(dset, command, **kwds):
...
-            log_online=True,
-            # not yet sure what we should do with the command output
-            # IMHO `run` itself should be very silent and let the command talk
-            log_stdout=False,
-            log_stderr=False,
-            expect_stderr=True,
-            expect_fail=True,
-            # TODO stdin
+            # command is always a string
+            ['sh', '-c', command],
```
mih, line 466: This is a problem, and will not work on windows properly (or at least be fragile). While there is a `sh` it may behave totally different than the actual calling shell. For example, with a system-wide installed git, executed through an `sh.exe` process inside a normal `cmd.exe` terminal, it might not read/find the user's `.gitconfig` (thanks to @adswa for pointing that out).

Here is a demo:

With `sh -c`:

```
(base) C:\DLTMP\some>datalad run git config --show-origin -l
file:C:/Program Files/Git/etc/gitconfig diff.astextplain.textconv=astextplain
file:C:/Program Files/Git/etc/gitconfig filter.lfs.clean=git-lfs clean -- %f
file:C:/Program Files/Git/etc/gitconfig filter.lfs.smudge=git-lfs smudge -- %f
file:C:/Program Files/Git/etc/gitconfig filter.lfs.process=git-lfs filter-process
file:C:/Program Files/Git/etc/gitconfig filter.lfs.required=true
file:C:/Program Files/Git/etc/gitconfig http.sslbackend=openssl
file:C:/Program Files/Git/etc/gitconfig http.sslcainfo=C:/Program Files/Git/mingw64/ssl/certs/ca-bundle.crt
file:C:/Program Files/Git/etc/gitconfig core.autocrlf=true
file:C:/Program Files/Git/etc/gitconfig core.fscache=true
file:C:/Program Files/Git/etc/gitconfig core.symlinks=false
file:C:/Program Files/Git/etc/gitconfig pull.rebase=false
file:C:/Program Files/Git/etc/gitconfig credential.https://dev.azure.com.usehttppath=true
file:C:/Program Files/Git/etc/gitconfig credential.helper=manager-core
file:.git/config        core.repositoryformatversion=0
file:.git/config        core.filemode=false
file:.git/config        core.bare=false
file:.git/config        core.logallrefupdates=true
file:.git/config        core.symlinks=false
file:.git/config        core.ignorecase=true
file:.git/config        annex.uuid=7bc9391d-e0eb-4ce0-bb16-10ecc2cefdc9
file:.git/config        annex.sshcaching=false
file:.git/config        annex.crippledfilesystem=true
file:.git/config        annex.version=8
file:.git/config        annex.backends=MD5E
file:.git/config        filter.annex.smudge=git-annex smudge -- %f
file:.git/config        filter.annex.clean=git-annex smudge --clean -- %f
```

and without:

```
(base) C:\DLTMP\some>datalad run git config --show-origin -l
file:C:/Program Files/Git/etc/gitconfig diff.astextplain.textconv=astextplain
file:C:/Program Files/Git/etc/gitconfig filter.lfs.clean=git-lfs clean -- %f
file:C:/Program Files/Git/etc/gitconfig filter.lfs.smudge=git-lfs smudge -- %f
file:C:/Program Files/Git/etc/gitconfig filter.lfs.process=git-lfs filter-process
file:C:/Program Files/Git/etc/gitconfig filter.lfs.required=true
file:C:/Program Files/Git/etc/gitconfig http.sslbackend=openssl
file:C:/Program Files/Git/etc/gitconfig http.sslcainfo=C:/Program Files/Git/mingw64/ssl/certs/ca-bundle.crt
file:C:/Program Files/Git/etc/gitconfig core.autocrlf=true
file:C:/Program Files/Git/etc/gitconfig core.fscache=true
file:C:/Program Files/Git/etc/gitconfig core.symlinks=false
file:C:/Program Files/Git/etc/gitconfig pull.rebase=false
file:C:/Program Files/Git/etc/gitconfig credential.https://dev.azure.com.usehttppath=true
file:C:/Program Files/Git/etc/gitconfig credential.helper=manager-core
file:C:/Users/mih/.gitconfig    user.email=michael.hanke@gmail.com
file:C:/Users/mih/.gitconfig    user.name=Michael Hanke
file:C:/Users/mih/.gitconfig    annex.security.allowed-url-schemes=http https file local
file:C:/Users/mih/.gitconfig    annex.security.allowed-url-addresses=all
file:.git/config        core.repositoryformatversion=0
file:.git/config        core.filemode=false
file:.git/config        core.bare=false
file:.git/config        core.logallrefupdates=true
file:.git/config        core.symlinks=false
file:.git/config        core.ignorecase=true
file:.git/config        annex.uuid=7bc9391d-e0eb-4ce0-bb16-10ecc2cefdc9
file:.git/config        annex.sshcaching=false
file:.git/config        annex.crippledfilesystem=true
file:.git/config        annex.version=8
file:.git/config        annex.backends=MD5E
file:.git/config        filter.annex.smudge=git-annex smudge -- %f
file:.git/config        filter.annex.clean=git-annex smudge --clean -- %f
```

Any use of `sh -c` in the tests is likely to run into the same problem.

mih: Ha, and this might also explain the magic file name mangling that is happening on windows with this PR.

When we send a windows path to `sh -c` it will never work. The reason for that is that it will come in as python string `'C:\\something\\else'`, which will translate to the command `sh -c "C:\something\else"`, and `sh` will apply standard handling of `\` as an escape char, and kill all "path separators".

Overall all these quirks seem to suggest that we should only very carefully, probably never, try to apply UNIXy paradigm to windows. It might work for a few cases and some layers, but then it breaks.

mih:
It seems not calling `sh -c` makes all problems go away and also remove the need for any adjustments of tests. I pushed a minified patch.

mih:
Appveyor is happy now, and Travis is slow.

mih:
> > It seems not calling `sh -c` makes...
> 
> So, this PR's use of `sh -c ...` was just a holdover what would have been needed before [df4f782](https://github.com/datalad/datalad/commit/df4f782bd8a30693cb5a28665ad4e41eb82c9907) (NF: WitlessRunner support platform shell command execution, 2020-10-09), right?

Yes. A wasteful dead-end.

kyleam (approved):
Thanks @mih and @adswa.

Nice to see that the solution to the file name mangling end up being removing unnecessary code rather than adding an ugly kludge.

mih:
Thx @kyleam ! Happy end to a long story.

Issue #5004, opened 2020-10-10
Finally enable interactive commit message entry


mih:
There were previous efforts (e.g. #4121), but none made it into the mainline. However, the situation changed quite a bit since the last attempt. We are already making multiple calls to `git commit` in `GitRepo.commit()` and a simplistic approach would be to add another `git commit --amend` with a `NoCapture` protocol execution to enable interactive specification of a commit message, if so desired (i.e. configured). The approach could be similar/identifical to #4121, but there would be a `StdOutErrCapture` execution first (which is the status quo) in order to maintain the error/behavior detection we have in place, and one additional invocation of `git commit` at the end. Performance is a non-issue for interactive message edits anyways.

I added a test to #4996 to verify that interactive execution is viable via `WitlessRunner` (https://github.com/datalad/datalad/pull/4996/commits/17339db0d5b9bcaebe56d1584479dc065880d366)

PR #4212, opened 2020-02-27
BF(workaround): position --debug AFTER other annex_options

yarikoptic:
Issue filed within annex: https://git-annex.branchable.com/bugs/--debug_is_not_in_effect_if_precedes_-c/

verified that works, no dedicated test

PR #5005, opened 2020-10-10
ENH: Allow experimenting with interactive commit-message entry

mih:
Concept: Run a follow-up `git commit --amend` in a runner session that doesn't capture any output, hence can drive any editor properly. This approach allows us to maintain any error message inspection, because these commit calls are unchanged. The interactive edit is a pure add-one, that is only attempted if anyhow possible. It should also not fail in ways that it would not have failed in the previous commit call.

Based on the feedback in gh-4121, the default (non-interactive) behavior
is unchanged. Trigger the interactive behavior via:

`datalad -c datalad.save.no-message=interactive save`

Fixes gh-5004

The primary purpose of this change is to allow enabling in dev environments, in order to get practical experience with the interactive behavior in various scenarios (simple one-dataset save, up to complex hierarchical operations). The current concept on how to deal with these is datalad/datalad-revolution#71 (which should see practical validation, which is prepared with this change). In any other scenario or deployment, this should be a no-op.

### DIFF
--- a/datalad/interface/common_cfg.py
+++ b/datalad/interface/common_cfg.py
@@ -406,6 +406,15 @@ definitions = {
         'default': 'auto',
         'type': EnsureChoice('on', 'off', 'auto'),
     },
+    'datalad.save.no-message': {
+        'ui': ('question', {
+            'title': 'Commit message handling',
+            'text': 'When no commit message was provided: '
+                    'attempt to obtain one interactively (interactive); '
+                    'or use a generic commit message (generic)'}),
+        'default': 'generic',
+        'type': EnsureChoice('interactive', 'generic'),
+    },
     'datalad.install.inherit-local-origin': {
         'ui': ('question', {
             'title': 'Inherit local origin of dataset source',
--- a/datalad/interface/common_cfg.py
+++ b/datalad/interface/common_cfg.py
@@ -411,7 +411,9 @@ definitions = {
             'title': 'Commit message handling',
             'text': 'When no commit message was provided: '
                     'attempt to obtain one interactively (interactive); '
-                    'or use a generic commit message (generic)'}),
+                    'or use a generic commit message (generic). '
+                    'NOTE: The interactive option is experimental. The '
+                    'behavior may change in backwards-incompatible ways.'}),
         'default': 'generic',
         'type': EnsureChoice('interactive', 'generic'),
     },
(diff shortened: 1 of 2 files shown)


yarikoptic:
For a hierarchy of datasets, what would be the behavior?

mih:
> For a hierarchy of datasets, what would be the behavior?

The change is placed in `GitRepo.commit()`, hence it will apply to any commit, regardless of why it happens (e.g. recursive save) -- if the list of conditions are fulfilled (https://github.com/datalad/datalad/pull/5005/files#diff-d9648c7010ac950da0207d22bd299b8eR1729). The last aspect can vary between specific contexts, hence the question cannot be easily answered comprehensively.

Even with this new behavior enabled by config, ATM no interactive message entry would happen for a plain `create`. However, there would be an editor popping up for adding a subdataset (and thereby fixes #3882), regardless of whether it was cloned into a superdataset or created.

In any case, like #4212 this is just the prelude to introducing the templating from https://github.com/datalad/datalad-revolution/pull/71 (almost 2 years later).

I have used a patch like this for a while now in our data archiving workflow, after noticing that many steps needed to be followed up by `git commit --amend` to achieve an intelligible history.

Next steps following this PR would be to equip more commands with `--message` options following the original intent of #3316, but not its follow-up discussion. What remains then would be to scan the code re "internal" commit messages that are generated and given to `commit()` and replace them with external input, or turn them into templates.

Until this happens, the present change takes whatever automated message composition result and makes it editable.

mih:
FTR: While not visible by default, the Travis test execution is affected by this https://github.com/datalad/datalad/pull/4996#discussion_r502793713, and would be broken by this PR, if the new behavior was enabled by config. Here is a demo:
https://travis-ci.org/github/datalad/datalad/jobs/734568991

I have no idea yet, why test code on Travis thinks the execution is in an interactive environment.

mih:
Given that this is a no-op, unless explicitly activated, I plan to merge this later today. Please object if that is a problem.

kyleam:
> Given that this is a no-op, unless explicitly activated, I plan to merge this later today. Please object if that is a problem.

No objections from me.  

I think it'd be good for the documentation of `datalad.save.no-message` to say something along the lines of "This option is experimental. The behavior may change".

kyleam:
> documentation of `datalad.save.no-message` to say something along the lines of "This option is experimental. The behavior may change".

Based on discussion during the last call, it seems like @mih was on board with this idea.  I've pushed a commit doing that, but of course feel free to tweak/revert.

mih:
Thx much!


Ups, I had plans to merge this a week ago 🤔

mih:
OK, let's do this! Thx @kyleam

PR #5009, opened 2020-10-10
RM: some spotted direct mode (support was removed) related code

yarikoptic:


mih:
I restarted a failing Travis test ("too slow").

PR #5011, opened 2020-10-11
RF: Strip `AnnotatePaths` from `create_sibling()`

mih:
Use a 2-stage approach (subdatasets(recursive) + diff()) to avoid the
runtime cost of a full diff(recursive).

Ping gh-3368


kyleam (approved):
Stepping through with a few different scenarios, I didn't notice any issues, and the tests are happy.

mih:
Thx @kyleam

PR #5012, opened 2020-10-11
RF: Strip `AnnotatePaths` from `drop()`

mih:
Ping gh-3368

### DIFF
--- a/datalad/distribution/drop.py
+++ b/datalad/distribution/drop.py
@@ -22,11 +22,12 @@ from datalad.utils import ensure_list
 from datalad.support.param import Parameter
 from datalad.support.constraints import EnsureStr, EnsureNone
 from datalad.support.exceptions import InsufficientArgumentsError
-from datalad.support.gitrepo import GitRepo
-from datalad.distribution.dataset import Dataset, EnsureDataset, \
-    datasetmethod
-from datalad.interface.annotate_paths import AnnotatePaths
-from datalad.interface.annotate_paths import annotated2content_by_ds
+from datalad.distribution.dataset import (
+    Dataset,
+    EnsureDataset,
+    datasetmethod,
+    require_dataset,
+)
 from datalad.interface.base import Interface
 from datalad.interface.common_opts import if_dirty_opt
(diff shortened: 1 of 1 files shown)


datalad/distribution/drop.py
```diff
@@ -178,53 +180,56 @@ def __call__(
...
+
         if dataset and not path:
             # act on the whole dataset if nothing else was specified
             path = refds_path
-        to_drop = []
-        for ap in AnnotatePaths.__call__(
-                dataset=refds_path,
+        content_by_ds = {}
+        for st in Status.__call__(
+                # no not use `ds` to preserve path semantics
```
kyleam, line 201: s/no/do/ ?

datalad/distribution/drop.py
```diff
@@ -178,53 +180,56 @@ def __call__(
...
                 recursion_limit=recursion_limit,
-                action='drop',
-                # justification for status:
-                # content need not be dropped where there is none
-                unavailable_path_status='notneeded',
-                nondataset_path_status='error',
+                eval_subdataset_state='no',
+                report_filetype='raw',
                 return_type='generator',
-                on_failure='ignore'):
```
kyleam, line 196: Should `drop` behave the same as `save` here?  It passes `on_failure='ignore'` to `status` and then lets the outer layer decide how to handle non-ok results:

https://github.com/datalad/datalad/blob/a55baf68cbcd624b467b25757bb9689a438e15d4/datalad/core/local/save.py#L208-L216

kyleam (commented):
Nice.  

I think the `test_get_cached_dataset` failure in one of the Travis builds is unrelated.

https://travis-ci.org/github/datalad/datalad/jobs/734784386#L2249

datalad/distribution/drop.py
mih: Yes, thx, fixed!

datalad/distribution/drop.py
mih: Oh, indeed! Thanks for catching that.

Issue #4998, opened 2020-10-08
Address "non-finishing" batched processes properly


yarikoptic:
"Introduced" within https://github.com/datalad/datalad/pull/4829/files#diff-9f14d6f5afddd497b06677e855a05d9eR1333

- our "waiting time" might be insufficient in legit cases, so we should not just abandon git-annex processes and proceed (e.g. with a commit) since it could lead to inconsistent commits (yet to be seen, but I saw batched processes exiting peacefully after 1 or 2 loops)
- reported to @joeyh : https://git-annex.branchable.com/bugs/some_annex_addurl_--fast_--with-files_--json_--json-error-messages_--batch__do_not_quit/ . More investigation is needed, may be bisection of git-annex (yet to double check that older versions aren't the subject to this) and more careful analysis of changes in #4829 on what has introduced that effect
- the way I  reproduce it mimics original stall on CI
```shell
(dev3) test@smaug:~/proj/datalad/datalad$ git diff
diff --git a/datalad/cmd.py b/datalad/cmd.py
index d4d54eb46..c8a666e24 100644
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -1353,6 +1353,7 @@ class BatchedCommand(SafeDelCloseMixin):
                     timeout=3.0,
                 )
             except subprocess.TimeoutExpired:
+                import pdb; pdb.set_trace()
                 lgr.warning(
                     "Batched process %s did not finish, abandoning it without killing it",
                     process)
...
(dev3) test@smaug:~/proj/datalad/datalad-crawler/__testhome__$ PATH=~/git-annex/20201005-snapshot/ DATALAD_LOG_LEVEL=WARNING DATALAD_TESTS_TEMP_KEEP=1 python -m nose -s -v --with-cov --cover-package datalad ../datalad_crawler/dbs/tests/test_files.py ../datalad_crawler/nodes/tests/test_annex.py
```
and end up in pdb upon timeout

yarikoptic:
grrr this bug... as soon as became brave to bisect sh.t out of it -- cannot reproduce locally or on smaug in **exactly** the same env I prepared for joey and reproduced it reliably in the past.  May be it is the fact that more starlink satellites in the orbit, I am out of ideas.... But I know that if I remove this kludge, travis will stall!  I bet it will!

yarikoptic:
[...]

PR #5013, opened 2020-10-11
Revert "ENH: When closing batched processes avoid waiting forever"

yarikoptic:
This reverts commit 383819eafc77e604c9cea2977820dcd299ea2902.

That workaround, introduced in #4829, just hides away the problem, possibly
leading to data loss (e.g. batched process would not finish its processing by
then and we proceed to eg save or alike), and makes it hard to impossible to
spot/debug in real life cases where it manifests itself.

This is unlikely to fix the issue, but I want to see if it still
reproduces (I failed to reproduce it in the past 2 days).

More references:

- datalad issue https://github.com/datalad/datalad/issues/4998
- git-annex report https://git-annex.branchable.com/bugs/some_annex_addurl_--fast_--with-files_--json_--json-error-messages_--batch__do_not_quit/

I do not think of any reasonable "timeout" we could specify and be "correct", since in some cases (very large index) it could take awhile for e.g. `addurl --batch` to complete and I do not know how long exactly.  What we could do is to add some "activity indication", e.g. if it did not complete within X seconds, we bring up "progress bar like" information bar which would report PID of the batched child process, and may be its tree with CPU/IO  information. That would allow to ease troubleshooting, alert user about still ongoing (or stalled?) activity.  In interactive mode we could may be even add some shortcut (e.g. if Ctrl-C is pressed and thus we get an exception) to jump into pdb... just ideas

Conflicts:
	datalad/cmd.py - melded with added later logging of stderr

### DIFF
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -1338,27 +1338,34 @@ class BatchedCommand(SafeDelCloseMixin):
                 "Closing stdin of %s and waiting process to finish", process)
             process.stdin.close()
             process.stdout.close()
-            # try waiting for the annex process to finish 3 times for 3 sec
-            # with 1s pause in between
-            try:
-                try_multiple(
-                    # ntrials
-                    3,
-                    # exception to catch
-                    subprocess.TimeoutExpired,
-                    # base waiting period
-                    1.0,
-                    # function to run
-                    process.wait,
-                    timeout=3.0,
-                )
-            except subprocess.TimeoutExpired:
-                lgr.warning(
-                    "Batched process %s did not finish, abandoning it without killing it",
-                    process)
-            if process.returncode is not None:
-                lgr.debug("Process %s has finished", process)
+            from . import cfg
+            cfg_var = 'datalad.runtime.stalled-external'
+            cfg_val = cfg.obtain(cfg_var)
+            if cfg_val == 'wait':
+                process.wait()
+            elif cfg_val == 'abandon':
+                # try waiting for the annex process to finish 3 times for 3 sec
+                # with 1s pause in between
+                try:
+                    try_multiple(
+                        # ntrials
+                        3,
+                        # exception to catch
+                        subprocess.TimeoutExpired,
+                        # base waiting period
+                        1.0,
+                        # function to run
+                        process.wait,
+                        timeout=3.0,
+                    )
+                except subprocess.TimeoutExpired:
+                    lgr.warning(
+                        "Batched process %s did not finish, abandoning it without killing it",
+                        process)
+            else:
+                raise ValueError(f"Unexpected {cfg_var}={cfg_val!r}")
             self._process = None
+            lgr.debug("Process %s has finished", process)
 
         # It is hard to debug when something is going wrong. Hopefully logging stderr
         # if generally asked might be of help
--- a/datalad/interface/common_cfg.py
+++ b/datalad/interface/common_cfg.py
@@ -359,6 +359,15 @@ definitions = {
         'type': EnsureChoice('all', 'success', 'failure', 'ok', 'notneeded', 'impossible', 'error'),
         'default': None,
     },
+    'datalad.runtime.stalled-external': {
+        'ui': ('question', {
+            'title': 'Behavior for handing external processes',
+            'text': 'What to do with external processes if they do not finish in some minimal reasonable time. '
+                    'If "abandon", datalad would proceed without waiting for external process to exit. '
+                    'ATM applies only to batched git-annex processes. Should be changed with caution.'}),
+        'type': EnsureChoice('wait', 'abandon'),
+        'default': 'wait',
+    },
     'datalad.search.indexercachesize': {
         'ui': ('question', {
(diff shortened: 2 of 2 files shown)


mih:
I would much prefer not to revert this without implementing one of the suggested alternative mitigation strategies. We did not add this for no reason. It impacts many completely unrelated developments and forces a substantial number of PRs to also become digging operations for faulty batch process handling, and results in an overall slowing.

yarikoptic:
> I would much prefer not to revert this without implementing one of the suggested alternative mitigation strategies

Well
- I filed this PR with the hope also to reproduce original stall, but it did not reproduce :-(  As I have mentioned I have failed to reproduce it in the past days although I was 100% successful a week ago.  It is a mystery
- I do not remember any other viable mitigation strategy ATM, please suggest one 
- I also do not want to revert until we have some assurance that it is somehow "gone" 

> We did not add this for no reason. 

We added it for a to workaround a new stall in for which we had an unknown cause.  Or am I forgetting something?

> It impacts many completely unrelated developments and forces a substantial number of PRs to also become digging operations for faulty batch process handling, and results in an overall slowing.

As reported to [git-annex board](https://git-annex.branchable.com/bugs/some_annex_addurl_--fast_--with-files_--json_--json-error-messages_--batch__do_not_quit/) I did observe similar stalls with heavy in subdatasets `addurls` invocations.  Did you observe them in real life (beyond that original PR where it came about) recently? If so -- please share, may be it would shine the light on the underlying cause.
 
NB I have restarted testing of externals...

yarikoptic:
FTR: another thing which has changed is datalad itself.  I did reproduce on crawler tests by going with master back to #4829 merge point: be47e8d66d5d67ca405daf6628bb07fcc7c9bac5 .  Just for the fun of it I will see if I could bisect it to what caused in datalad for it to "go away" (I do not think really but at least as affecting our tests)

yarikoptic:
uff... automated bisection didn't work for me... merge of #4850  seems made it go away, and it seems in particular bc3117240865f1398290410751b67d82a839c4fd

yarikoptic:
Note: that fix was used by in my real-life example on ABCD.  I did see some processes reporting being stuck/abandoned.  BUT some were stuck only for a single iteration of `wait` and then did quit on their own -- they were just taking too long to finish.  SO I no longer think that I actually observed  a real life full stall since that "fix".

@mih - have you seen such messages in your recent past with some certainty that they were actual stalls and not that underlying annex process was taking awhile to complete?

yarikoptic:
As joey [pointed out to now "obvious" fact](https://git-annex.branchable.com/bugs/some_annex_addurl_--fast_--with-files_--json_--json-error-messages_--batch__do_not_quit/#comment-f12f224c91d9d2795e18975278f0be61): it is a separate forked `nose` process which is holding stdin of git-annex . So, I suspect, that if a test (in datalad or datalad extensions) does not close all batched processes by completion -- then there could be a stranded fork of the nose process, so even if we close them while running the next batched test -- we could get into such a situation.  In case of this particular test stall it was `git annex info` batched call while testing "DBs".  I do not think that DBs there should close anything explicitly since they are just passed an AnnexRepo instance, so nothing to fix there (although I have tested that adding `__del__` with `self.annex.precommit()` seems to resolve it as well.

FWIW, I have tried to add `--process-restartworker --processes=0 ` to `python -m nose ` invocation, but that did not help!

But note -- that `git-annex` call is a "read-only" operation!  So if we were to still provide safe guards with time out - I guess we could start to discriminate between read-only and write annexed batched processes.  But ideally we shouldn't probably.

As I think we "figured out" the cause (forked by `nose` process), and it seems to no longer manifest itself in tests (green here), and I have not seen reports from the "wild" on such stalls, I think it should be safe to merge this.  If it resurfaces again in the tests -- we know what to look for -- leaking across tests workers processes batched Annex processes.

yarikoptic:
I have missed the ping from #5013 ... Indeed we better figure out what is going there in windows land first

yarikoptic:
> I have missed the ping from #5015 ... Indeed we better figure out what is going there in windows land first

I am also considering RFing this PR if so seem desired to retain current functionality "windows only".  IMHO more of possible data loss for windows users is an acceptable measure ;-)  (so far it seems that stuck process is imho just a just a sign of another underlying problem, which would not necessarily go away on its own)

yarikoptic:
Ok, I have redone it, so now there is a configuration variable so we could get back to "wait and abandon" when needed to help a user.  Otherwise -- we should wait (by default).

yarikoptic:
insofar days spent debugging datalad+git-annex on windows  hint into real issue, where abandoning would just lead to breeding those poor bastard processes -- not a solution and thus windows is not a blocker. Configuration setting is implemented, so imho this one is ready to be merged.  Unless any objections voiced, I would merge it in a day or two (CI is green,  appveryor just complains that A != B, so no immediate "danger" ;-))

yarikoptic:
ok, "mitigation strategy" (via config) was implemented, everyone (and myself) had been warned, nobody objected, let's proceed while Germans are asleep

Issue #4967, opened 2020-09-25
save: Modifications remain after 'git mv' of annex file


kyleam:
This is a follow-up to gh-3686.  As @loj reported over there, there's an unresolved aspect of that issue.  If a symbolic link is moved with `git mv` and the depth of the link changes, `save` commits the staged changes as is (i.e., no adjustment in the link path) and then git-annex fixes up the path afterwards, leaving the dataset dirty and requiring a second `save` call.  Here's the failing test case linked to from that issue:

```diff
diff --git a/datalad/core/local/tests/test_save.py b/datalad/core/local/tests/test_save.py
index 08b706b857..7384ab0ee7 100644
--- a/datalad/core/local/tests/test_save.py
+++ b/datalad/core/local/tests/test_save.py
@@ -11,6 +11,7 @@
 import os
 import os.path as op
 
+from datalad.cmd import Runner
 from datalad.utils import (
     assure_list,
     Path,
@@ -834,3 +835,17 @@ def test_save_dotfiles():
     for git in [True, False, None]:
         for save_path in [None, "nodot-subdir"]:
             yield check_save_dotfiles, git, save_path
+
+
+@with_tree({"subdir": {"foo": "foocontent"}})
+def test_save_git_mv_fixup(path):
+    ds = Dataset(path).create(force=True)
+    ds.save()
+    assert_repo_status(ds.path)
+    ds.repo.call_git(["mv", op.join("subdir", "foo"), "foo"])
+    ds.save()
+    # Was link adjusted properly?  (gh-3686)
+    eq_(Runner(cwd=ds.path).run(["cat", "foo"])[0],
+        "foocontent")
+    # FIXME: But only the deletion was saved.
+    assert_repo_status(ds.path)
```

<details>
<summary>output</summary>

```
======================================================================
FAIL: datalad.core.local.tests.test_save.test_save_git_mv_fixup
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/kyle/src/python/venvs/datalad/lib/python3.7/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/kyle/src/python/datalad/datalad/tests/utils.py", line 554, in _wrap_with_tree
    return t(*(arg + (d,)), **kw)
  File "/home/kyle/src/python/datalad/datalad/core/local/tests/test_save.py", line 851, in test_save_git_mv_fixup
    assert_repo_status(ds.path)
[...]

PR #5016, opened 2020-10-12
BF: Treat git-mv'ed annexed files like untracked content

mih:
`git-mv` stages the move, and our handling of such a change causes
a potential fix-up of an annex symlink to happen outside and after
our change analysis. This change unstages such paths and
re-categorizes such a modification as an untracked addition. This
essentially removes the special case and fixes gh-4967

### DIFF
--- a/datalad/core/local/tests/test_save.py
+++ b/datalad/core/local/tests/test_save.py
@@ -905,5 +905,5 @@ def test_save_git_mv_fixup(path):
     eq_(WitlessRunner(cwd=ds.path).run(
         ["cat", "foo"], protocol=StdOutCapture)['stdout'],
         "foocontent")
-    # FIXME: But only the deletion was saved.
+    # all clean
     assert_repo_status(ds.path)
--- a/datalad/core/local/tests/test_save.py
+++ b/datalad/core/local/tests/test_save.py
@@ -52,7 +52,10 @@ from datalad.api import (
     install,
     save,
 )
-
+from datalad.cmd import (
+    StdOutCapture,
+    WitlessRunner,
+)
 
 tree_arg = dict(tree={'test.txt': 'some',
                       'test_annex.txt': 'some annex',
@@ -888,4 +891,19 @@ def test_save_diff_ignore_submodules_config(path):
                        where="local", reload=True)
     # Saving a subdataset doesn't fail when diff.ignoreSubmodules=all.
     ds.save()
+    assert_repo_status(ds.path)
+
+
+@with_tree({"subdir": {"foo": "foocontent"}})
+def test_save_git_mv_fixup(path):
+    ds = Dataset(path).create(force=True)
+    ds.save()
+    assert_repo_status(ds.path)
+    ds.repo.call_git(["mv", op.join("subdir", "foo"), "foo"])
+    ds.save()
+    # Was link adjusted properly?  (gh-3686)
+    eq_(WitlessRunner(cwd=ds.path).run(
+        ["cat", "foo"], protocol=StdOutCapture)['stdout'],
+        "foocontent")
+    # FIXME: But only the deletion was saved.
     assert_repo_status(ds.path)
--- a/datalad/core/local/tests/test_save.py
+++ b/datalad/core/local/tests/test_save.py
@@ -907,3 +907,23 @@ def test_save_git_mv_fixup(path):
         "foocontent")
     # all clean
     assert_repo_status(ds.path)
+
+
+@with_tempfile
+def test_save_staged_nonannex_goes_to_annex(path):
+    # Due to the `git mv` workaround above, staged paths are reset, so `datalad
+    # save` will send a file to the annex despite an explicit `git add` before.
+    ds = Dataset(path).create()
+    repo = ds.repo
+    foo = (ds.pathobj / "foo")
+    foo.write_text("content")
+    repo.call_git(["add", "foo"])
+    assert_repo_status(ds.path, added=[foo])
+
+    assert_not_in("backend",
+                  repo.get_content_annexinfo(paths=[foo])[foo])
+
+    ds.save()
+    assert_repo_status(ds.path)
+    assert_in("backend",
+              repo.get_content_annexinfo(paths=[foo])[foo])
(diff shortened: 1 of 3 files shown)


kyleam (commented):
> This change unstages such paths and
> re-categorizes such a modification as an untracked addition. This
> essentially removes the special case and fixes [gh-4967](https://github.com/datalad/datalad/issues/4967)

Okay, so with the path unstaged, it gets added by `git annex add`, which leads to the path being fixed up before the commit.  I haven't looked too closely, but I believe the test failures come from not being selective enough when picking out the paths to unstage (in particular, including subdatasets).

This PR's approach seems to fit fine with `save`s no-index worldview, and I don't see any real problems with it once the collection of paths to unstage is tightened.  The main alternative that I can think of is running `git annex fix` rather than unstaging.  Something like this:

```diff
diff --git a/datalad/support/gitrepo.py b/datalad/support/gitrepo.py
index 199d62d627..14e475579e 100644
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -3962,7 +3962,14 @@ def save_(self, message=None, paths=None, _status=None, **kwargs):
         # - commit (with all paths that have been touched, to bypass
         #   potential pre-staged bits)
 
-        need_partial_commit = True if self.get_staged_paths() else False
+        staged_paths = self.get_staged_paths()
+        need_partial_commit = bool(staged_paths)
+        if need_partial_commit and hasattr(self, "_run_annex_command"):
+            for sp in staged_paths:
+                st = status.get(self.pathobj / Path(sp), {})
+                if st.get("state") == "added" and st.get("type") == "symlink":
+                    self._run_annex_command("fix", runner="gitwitless")
+                    break
 
         # remove first, because removal of a subds would cause a
         # modification of .gitmodules to be added to the todo list
```

It's admittedly ugly that `GitRepo.save_` knows about and handles a git-annex thing.  We could restructure it to avoid that, though 1) `GitRepo.save_` already does a similar check for `annexstatus` and 2) the same could be said for the unstage approach of this PR (it's just more hidden away).

yarikoptic:
I guess needs a "come back", conflicts resolved, and rebased on `maint`?

mih:
Rebased on maint, conflicts resolved. Let's see where we stand.

mih:
This is what breaks with `maint`:

```
2895ERROR: datalad.interface.tests.test_rerun.test_rerun_explicit
2896----------------------------------------------------------------------
2897Traceback (most recent call last):
2898  File "/home/appveyor/dlvenv/lib/python3.8/site-packages/nose/case.py", line 198, in runTest
2899    self.test(*self.arg)
2900  File "/home/appveyor/dlvenv/lib/python3.8/site-packages/datalad/tests/utils.py", line 1972, in _wrap_skip_if_adjusted_branch
2901    return func(*args, **kwargs)
2902  File "/home/appveyor/dlvenv/lib/python3.8/site-packages/datalad/tests/utils.py", line 578, in _wrap_with_tree
2903    return t(*(arg + (d,)), **kw)
2904  File "/home/appveyor/dlvenv/lib/python3.8/site-packages/datalad/interface/tests/test_rerun.py", line 821, in test_rerun_explicit
2905    with open(op.join(ds.path, "to_modify"), "a") as ofh:
2906PermissionError: [Errno 13] Permission denied: '/home/appveyor/DLTMP/datalad_temp_tree_test_rerun_explicite9ktxlv4/to_modify'
```

kyleam:
I've changed the approach of `test_rerun_explicit` to sidestep the issue.  The problem is that a `git add <file>` followed by a `datalad save` will now send a file to the annex.  I've also added a dedicated test to demonstrate that behavior.

Above I said:

> This PR's approach seems to fit fine with `save`s no-index worldview, and I don't see any real problems with it once the collection of paths to unstage is tightened.

Thinking about it again, it seems to me that the behavior above is worse than the `git mv`-related issue it fixes.  Even if DataLad doesn't expose the concept of the index, it doesn't seem unlikely that users will use `git add` directly every now and then, and sending staged files to the annex is pretty surprising.

Perhaps the operation should at least be restricted to symlinks?  (I'm unclear why the subdataset failures I mentioned above went away with the rebase, but, if it's still there but not triggered by a test, I guess restricting the paths would avoid that issue too.)

The alternative I propsed above would need an update for AnnexRepo changes since then.  FWIW here it is against the current tip of this PR (52cc4b17ed):

<details>
<summary>diff</summary>

```diff
diff --git a/datalad/support/gitrepo.py b/datalad/support/gitrepo.py
index 1de63e18b3..666cb61b95 100644
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -4024,26 +4024,18 @@ def save_(self, message=None, paths=None, _status=None, **kwargs):
         # - commit (with all paths that have been touched, to bypass
         #   potential pre-staged bits)
 
-        need_partial_commit = False
+        staged_paths = self.get_staged_paths()
+        need_partial_commit = bool(staged_paths)
         # we need to look out for git-mv'ed annexed files. if annex
         # needs to fix up a symlink after `add`, we would loose track of
-        # it when the change was staged. look through all staged paths,
-        # and unstage them, if they would be added anyways, according
-        # to the given status info. That approach normalizes this special
-        # case to the addition of an untracked file.
+        # it when the change was staged.
         # see https://github.com/datalad/datalad/issues/4967
-        unstage = []
-        for sp in self.get_staged_paths():
-            qp = self.pathobj / Path(sp)
-            if status.get(qp, {}).get('state', None) == 'added':
-                unstage.append(sp)
-                status[qp]['state'] = 'untracked'
-            else:
-                need_partial_commit = True
-        if unstage:
-            # keep compat with older Git versions for now
-            #self.call_git(['restore', '--staged'], files=unstage)
-            self.call_git(['reset', 'HEAD'], files=unstage)
+        if need_partial_commit and hasattr(self, "call_annex"):
+            for sp in staged_paths:
+                st = status.get(self.pathobj / Path(sp), {})
+                if st.get("state") == "added" and st.get("type") == "symlink":
+                    self.call_annex(["fix"])
+                    break
 
         # remove first, because removal of a subds would cause a
         # modification of .gitmodules to be added to the todo list
```

</details>

mih:
I am closing this now, due to inactivity, and lack of time.

PR #5017, opened 2020-10-12
BF: always save run sidecar files 

bpoldrack:
Fix a bug in `datalad-run`, where sidecar files weren't saved, when outputs are specified w/ `--explicit`.
A caller of `run` can't know the to be saved path and shouldn't need to care for a builtin feature to be functional. Can't see an argument why this wouldn't be a bug.


kyleam (approved):
Thanks @bpoldrack.

PR #5020, opened 2020-10-12
RM: remove _license_info and use in customremotes/main.py

yarikoptic:
we stopped showing license for --version output for the main process,
but still were showing it for special remotes.  I see no reason for
the difference and maintaining another place where we would need to
tune up copyright years, so just removing it altogether


kyleam (approved):

mih (approved):
Indeed, thx!

Issue #5023, opened 2020-10-13
RIA clone HTTP403 handling wrong and doesn't support authenticated access


mih:
When cloning a dataset from a RIA store I see:

```
fatal: bad config line 1 in file /tmp/datalad_temp_drscop6t
```

When debug is turned on , I see the following log message:

```
[DEBUG  ] Found no UUID for ORA special remote at 'origin' (CommandError: 'git config -f /tmp/datalad_temp_drscop6t datalad.ora-remote.uuid' failed with exitcode 128 [cmd.py:run:517]) 
```

The conclusion seems wrong:

```
% git cat-file -p git-annex:remote.log
08350c08-13a8-4026-92d8-ec50772beed7 archive-id=5df8eb3a-95c5-11ea-b4b9-a0369f287950 autoenable=true encryption=none externaltype=ora name=inm7-storage type=external url=ria+file:///ds/archive/srv timestamp=1589552352.070973114s
```

But that may be caused by some kind of misconfiguration of the store. In any case, the "fatal" error reported for a file that won't be around for an investigation should likely not be presented like this.

Ping @bpoldrack

mih:
So this issue is this: There are actually two issues.

First: HTTP response handling is wrong (the pattern may also be the cause for #4980)

The broken config file, isn't broken. It was not obtained. The error handling at https://github.com/datalad/datalad/blob/961c42660e198c5bfa5c43665d9fcd70c4f80bd2/datalad/core/distributed/clone.py#L632-L640 seems to assume that any failure causes an exception, but in this case there is "only" a regular 403 response. The code doesn't notice that and proceeds happily.

Second: The URL that the repo was cloned from requires HTTP authentication. `git clone` consulted Git's credential manager and used the right credentials automatically. But the subsequently request by clone falsely assumes unauthorized access is possible. It should also support the same authorization flexibility that Git provides.

mih:
I think we should stop manual HTTP requests for downloads and centralize on a standard helper. The closest thing that I am aware of are the guts of `download_url()`, so something like this:

```
providers = Providers.from_config_files()
downloaded_path = providers.download(url, path=path, overwrite=overwrite)
```

I have yet to check whether this would be able to authenticate properly.

Issue #5024, opened 2020-10-13
ORA does not support authenticated HTTP(S) access


mih:
It performs a plain GET requests using the `requests` package and bypasses any internal helper. In contrast to #5023 this is not trivial to fix, because ORA reads chunks of bytes from a stream.

Alternatively, we should explore using an internal downloader that has its own progress reporting, and utilizes datalad's credential capabilties.

mih:
I tried something simple:

```diff
diff --git a/datalad/distributed/ora_remote.py b/datalad/distributed/ora_remote.py
index 7b8a3547c..14bd48053 100644
--- a/datalad/distributed/ora_remote.py
+++ b/datalad/distributed/ora_remote.py
@@ -556,15 +556,8 @@ class HTTPRemoteIO(object):
         # to annexremote.dirhash from within IO classes
 
         url = self.base_url + "/annex/objects/" + str(key_path)
-        response = requests.get(url, stream=True)
-
-        with open(filename, 'wb') as dst_file:
-            bytes_received = 0
-            for chunk in response.iter_content(chunk_size=self.buffer_size,
-                                               decode_unicode=False):
-                dst_file.write(chunk)
-                bytes_received += len(chunk)
-                progress_cb(bytes_received)
+        from datalad.utils import download_url
+        download_url(url, filename, overwrite=True)
```

This makes `git annex get` work

```
% git -C atj3 annex get readme.docx
get readme.docx (from inm7-storage...) 
[INFO] Downloading 'https://archive.ds.inm7.de/00e/0c610-0f27-41bb-9626-157ec8815bfe/annex/objects/xk/vw/MD5E-s12060--9a9019561007e616e328cd15e87863a3.docx/MD5E-s12060--9a9019561007e616e328cd15e87863a3.docx' into '.git/annex/tmp/MD5E-s12060--9a9019561007e616e328cd15e87863a3.docx' 
[INFO] http session: Authenticating into session for https://archive.ds.inm7.de/00e/0c610-0f27-41bb-9626-157ec8815bfe/annex/objects/xk/vw/MD5E-s12060--9a9019561007e616e328cd15e87863a3.docx/MD5E-s12060--9a9019561007e616e328cd15e87863a3.docx 
(checksum...) ok
(recording state in git...)
```

but strangely fails with `datalad get`

```
% datalad -C atj3 -l debug get readme.docx
[...]

PR #5025, opened 2020-10-13
BF: RIA clone HTTP403 handling and support authenticated access

mih:
This PR essentially removes a custom HTTP GET request handling and uses a new (but pretty trivial) download helper.

Fixes gh-5023

... although it could be made better by supporting git-credentials as a credential sources in DataLad. This, however, would only make internal changes and no have consequences for the proposed change in `clone`.

A significant annoyance is that low-level code insists on reporting INFO level log messages that will no report URL to be downloaded to short-lived temporary files with cryptic names. But at least it is functioning properly for now.

### DIFF
--- a/datalad/core/distributed/clone.py
+++ b/datalad/core/distributed/clone.py
@@ -45,9 +45,11 @@ from datalad.support.constraints import (
     EnsureStr,
     EnsureKeyChoice,
 )
+from datalad.support.exceptions import DownloadError
 from datalad.support.param import Parameter
 from datalad.support.network import (
     get_local_file_url,
+    download_url,
     URL,
     RI,
     DataLadRI,
@@ -630,15 +632,13 @@ def postclonecfg_ria(ds, props):
         scheme = props['giturl'].split(':', 1)[0]
         if scheme in ['http', 'https']:
             try:
-                response = requests.get("{}{}config".format(
-                    props['giturl'],
-                    '/' if not props['giturl'].endswith('/') else '')
-                )
-                config_content = response.text
-            except requests.RequestException as e:
+                config_content = download_url(
+                    "{}{}config".format(
+                        props['giturl'],
+                        '/' if not props['giturl'].endswith('/') else ''))
+            except DownloadError as e:
                 lgr.debug("Failed to get config file from source:\n%s",
                           exc_str(e))
-
         elif scheme == 'ssh':
             # TODO: switch the following to proper command abstraction:
             # SSHRemoteIO ignores the path part ATM. No remote CWD! (To be
@@ -662,8 +662,8 @@ def postclonecfg_ria(ds, props):
                 lgr.debug("Failed to get config file from source: %s",
                           exc_str(e))
         else:
-            lgr.debug("Unknown URL-Scheme in %s. Can handle SSH, HTTP or "
-                      "FILE scheme URLs.", props['source'])
+            lgr.debug("Unknown URL-Scheme %s in %s. Can handle SSH, HTTP or "
+                      "FILE scheme URLs.", scheme, props['source'])
 
         # 3. And read it
         org_uuid = None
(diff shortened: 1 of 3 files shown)


bpoldrack (approved):
I agree it could be better but provides a necessary step forward. Eventually we probably want to iron out the downloader/provider/authenticator stuff and use its parts in a more sophisticated way.

datalad/core/distributed/clone.py
```diff
@@ -630,12 +633,14 @@ def postclonecfg_ria(ds, props):
         scheme = props['giturl'].split(':', 1)[0]
         if scheme in ['http', 'https']:
             try:
-                response = requests.get("{}{}config".format(
-                    props['giturl'],
-                    '/' if not props['giturl'].endswith('/') else '')
-                )
-                config_content = response.text
-            except requests.RequestException as e:
+                with TemporaryDirectory() as _dest:
```
bpoldrack, line 636: @mih: Looking at the failure in CrippledFS, using `TemporaryDirectory` might be the issue. We have dedicated helpers for temp locations that deal with several env vars/configs to figure out where exactly to put it.

bpoldrack:
FTR: Looking into Travis failure. Actual error message implies an `is_interactive` issue, since it's trying to interactively ask a `yesno`, when we aren't interactive. Underlying, however, is an authentication issue it seems:

```
datalad.downloaders: Level 5: Calling out into <bound method BaseDownloader._download of HTTPDownloader(authenticator=None, credential=None)> for https://github.com/datalad/example-dicom-functional/blob/master/dicoms/MR.1.3.46.670589.11.38317.5.0.4476.2014042516042547586

datalad.downloaders: DEBUG: Authenticated access was denied: Access to https://github.com/datalad/example-dicom-functional/blob/master/dicoms/MR.1.3.46.670589.11.38317.5.0.4476.2014042516042547586 has failed: status code 429 [http.py:check_response_status:107]

```

Question is how we can end up there, when no authentication is required and `HTTPDownloader(authenticator=None, credential=None)` suggests that we don't try.

bpoldrack:
FTR: Tried to debug by pushing a commit with additional logging, but the issue didn't replicate. Of f***** course! :-/

datalad/core/distributed/clone.py
bpoldrack: Update: Just like the interactive/authentication issue below, the CrippledFS failure actually seem to depend on moon phases and the color of my socks. *grmpf*
However, we ran into inconsistent temp location issues before, so I still think it's better to get that directory from the helpers.

datalad/utils.py
```diff
@@ -2531,6 +2531,39 @@ def split_cmdline(s):
     return args
 
 
+def download_url(url, dest, overwrite=False):
```
yarikoptic, line 2534: if anywhere, better to be in `support.network` I guess?

datalad/core/distributed/clone.py
yarikoptic: you can just fetch content via our downloaders (used by `download_url`) directly without saving into a file. E.g.
```
$> python -c 'from datalad.downloaders.providers import Providers; url="http://datasets.datalad.org/.git/config"; print(Providers.from_config_files().fetch(url))' | head
[INFO] Fetching 'http://datasets.datalad.org/.git/config' 
[core]
	repositoryformatversion = 0
	filemode = true
	bare = false
	logallrefupdates = true
	sharedrepository = 2
[receive]
...
```
so there is `.fetch` to complement `.download`, and thus you do not need any temp files AFAIK

yarikoptic (changes requested):
could be fetched without tempfiles

datalad/utils.py
mih: FTR: I am interpreting this comment as "should be moved" and not as "should not exist", although both interpretations match the statement.

datalad/core/distributed/clone.py
mih: This is implemented now

datalad/utils.py
mih: Moved.

mih:
OK, reworked. Rebuilding #5024 on top of the current state.

PR #5032, opened 2020-10-13
RF/BF: report "impossible" upon get of a annexed file without git-annex branch

### LINKED ISSUE #4510: datalad get silently fails when git-annex branch does not exist
I am using datalad 0.12.6 on ubuntu 18.04.4 machine (python 3.6.9).

When I run datalad get, it silently fails if the repo doesn't have any git-annex branch. There are a few OpenNeuro repos with this condition (see https://github.com/OpenNeuroOrg/openneuro/issues/1608) 

```
$ datalad get sub-010002_ses-01_T2w.nii.gz
$ datalad get .
$ 
```

Is it possible to add a check in datalad get to maybe output some error message and return non-0 exit code so I can distinguish between success v.s. failure?

yarikoptic:
Solution here is incomplete (comment added) since it should consider git
link files in cases of unlocked (and edited here: and crippled fs) annexed files.  It would be more expensive AFAIK,
thus before even attempting to implement that, I have decided to provide the
fix which would work in 90% of the time first.

Specifically it would Closes #4510 if accepted ;)

Without the fix, asking get to get content for a broken symlink results in "ah
-- all is good -- you do not need it" behavior (similar to the slogan of the
local little town store "We do not have it - you do not need it!"), which is
IMHO plain incorrect:

	$> datalad -f json_pp get sub-01/anat/sub-01_inplaneT2.nii.gz
	{
	  "action": "get",
	  "path": "/tmp/ds000001/sub-01/anat/sub-01_inplaneT2.nii.gz",
	  "refds": "/tmp/ds000001",
	  "status": "notneeded",
	  "type": "file"
	}

with exit 0.  With the fix:

	$> datalad -f json_pp get sub-01/anat/sub-01_inplaneT2.nii.gz
	[WARNING] no dataset annex, content already present [get(/tmp/ds000001/sub-01/anat/sub-01_inplaneT2.nii.gz)]
	{
	  "action": "get",
	  "path": "/tmp/ds000001/sub-01/anat/sub-01_inplaneT2.nii.gz",
	  "refds": "/tmp/ds000001",
	  "status": "impossible",
	  "type": "file"
	}
	(dev3) 1 29415 ->1

exits with 1, and status "impossible" which is IMHO factually correct, and
WARNING message is actually informative.

### DIFF
--- a/datalad/distribution/get.py
+++ b/datalad/distribution/get.py
@@ -664,7 +664,7 @@ def _get_targetpaths(ds, content, refds_path, source, jobs):
     # needs to be an annex to get content
     if not isinstance(ds_repo, AnnexRepo):
         for r in results_from_paths(
-                content, status='notneeded',
+                content,
                 message="no dataset annex, content already present",
                 action='get',
                 type='file',
--- a/datalad/distribution/tests/test_get.py
+++ b/datalad/distribution/tests/test_get.py
@@ -25,6 +25,7 @@ from datalad.interface.results import only_matching_paths
 from datalad.distribution.get import _get_flexible_source_candidates_for_submodule
 from datalad.support.annexrepo import AnnexRepo
 from datalad.support.exceptions import (
+    IncompleteResultsError,
     InsufficientArgumentsError,
     RemoteNotAvailableError,
 )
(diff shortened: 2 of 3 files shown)


kyleam (commented):
If we decide to go this way (I still lean towards leaving it as is), I think the following things are worth considering:

* In addition to not accounting for unlocked files, this doesn't handle directories, including a top-level ".".  `datalad get .` is one of the examples given in the initial post of gh-4510, so I'm not sure if the partial fix here should be seen as addressing that issue.  Either way, it seems worth noting in the commit message or code.

* The warning message shown is "no dataset annex, content already present".  Saying that content is already present seems confusing given that it's being reported as an impossible result due to the file not existing.

* The "doesn't exist -> impossible" logic is added to `datalad.interface.results.results_from_paths`.  Even though `results_from_paths` is used only in `get`, I think `datalad.distribution.get._get_targetpaths`, rather than a helper removed from the `get` context, would be a more appropriate spot to add the logic.

* With this change, non-annex broken symlinks for a `GitRepo` object are reported as errors.  This is a case that's probably not too common or important, but it's inconsistent with how they would be reported for an `AnnexRepo` object.

PR #5026, opened 2020-10-13
BF: ORA authenticated HTTP access

mih:
This sits on top of #5025 (to be able to make use of the `download_url()` helper), and replaces a custom HTTP GET that is unaware of error response codes (#4980) and necessary credentials (#5024).

ATM this solution is only working for `git annex get`, but not `datalad get`. No idea why.
https://github.com/datalad/datalad/issues/5024#issuecomment-707541179

Fixes #4980
Fixes #5024

PR #5035, opened 2020-10-14
BF: Don't modify PATH if not necessary

### LINKED ISSUE #5030: datalad ignores virtualenvs
Stumbled upon this in context of #5024 :

In that case it's `datalad get` behaving differently than `git-annex-get` because it's calling `git-annex` in a different environment leading to a different implementation of a special remote being used.
Ultimately it boils down to the `Runner` prefixing `PATH` with `/usr/bin`. This is because `GitRunnerBase` looks for `git-annex` and then prepending `PATH` with its location. This does not only make sure we know what annex installation we are using. It also overrules anything in a virtualenv that is also installed system-wide.

bpoldrack:
PATH modification in Runner classes is meant to make sure we use
bundled w/ annex git executable (except when instructed otherwise
by DATALAD_USE_DEFAULT_GIT). However, don't mess with PATH when
what we find to be bundled is the same thing that is found with
unmodified PATH anyway. This is particularly important when the
"bundeled" location is /usr/bin and we are running in a virtual
environment, leading to that PATH overruling the venv.

Closes #5030 and by extension #5024  (when combined with #5026)

### DIFF
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -1095,7 +1095,7 @@ class GitRunnerBase(object):
             # we only need to consider bundled git if it's actually different
             # from default. (see issue #5030)
             alongside = op.lexists(bundled_git_path) and \
-                        bundled_git_path != find_executable('git')
+                        bundled_git_path != op.realpath(find_executable('git'))
 
         return annex_path if alongside else ''
 
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -1091,7 +1091,12 @@ class GitRunnerBase(object):
             alongside = False
         else:
             annex_path = op.dirname(op.realpath(annex_fpath))
-            alongside = op.lexists(op.join(annex_path, 'git'))
+            bundled_git_path = op.join(annex_path, 'git')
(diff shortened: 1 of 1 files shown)


datalad/cmd.py
```diff
@@ -1091,7 +1091,12 @@ def _get_bundled_path():
             alongside = False
         else:
             annex_path = op.dirname(op.realpath(annex_fpath))
-            alongside = op.lexists(op.join(annex_path, 'git'))
+            bundled_git_path = op.join(annex_path, 'git')
+            # we only need to consider bundled git if it's actually different
+            # from default. (see issue #5030)
+            alongside = op.lexists(bundled_git_path) and \
+                        bundled_git_path != find_executable('git')
```
kyleam, line 1098: Okay, ignoring the guards, that expands to 
```
join(dirname(realpath(find_executable('git-annex'))), 'git') == find_executable('git')
```

Is the `realpath` on the left but not right a concern?

kyleam (commented):
Thanks for the analysis.  Make sense to me.

kyleam (approved):

PR #5037, opened 2020-10-14
Use environment file instead of set-env

jwodder:
The set-env workflow command is deprecated in favor of environment files: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

### DIFF
--- a/.github/workflows/test_macos.yml
+++ b/.github/workflows/test_macos.yml
@@ -24,7 +24,7 @@ jobs:
     - name: Install git-annex
       run: |
         eval source tools/ci/install-annex.sh ${{ matrix.install_scenario }}
-        echo "::set-env name=PATH::$PATH"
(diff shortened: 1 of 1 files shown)


.github/workflows/test_macos.yml
```diff
@@ -24,7 +24,7 @@ jobs:
     - name: Install git-annex
       run: |
         eval source tools/ci/install-annex.sh ${{ matrix.install_scenario }}
-        echo "::set-env name=PATH::$PATH"
+        echo "PATH=$PATH" >> $GITHUB_ENV
```
yarikoptic, line 27:
```suggestion
        echo "PATH=$PATH" >> "$GITHUB_ENV"
```
just to be on a safer side ;)

yarikoptic:
Great, thank you!

PR #5040, opened 2020-10-15
BF: UI - .error also should go through .message to handle pbars

yarikoptic:
Otherwise might interfere with progress bars

I have not "observed" wrong behavior but change looks logical to me.


kyleam (approved):
Good catch.  Reviewing b7e57dfa2d, I don't see a reason `error` should't get the same treatment as `message`.

PR #5022, opened 2020-10-13
NF+OPT+ENH: a helper to parallelize operations (via threads)

yarikoptic:
We virtually never see datalad "busy".  I believe that even with the simplest parallelization via threads, thus simpler mechanisms to orchestrate, we can make datalad more efficient by working with multiple git/git-annex processes at once, e.g in cases of multiple datasets to work on

- the primary "motivational" use case is `addurls` which might need to create a large number of subdatasets.  IMHO the most logical (single) point of "optimization" would be the `create` command if it was made to accept multiple datasets (#3233), but since it is not the case, the last commit (ATM ca87b8ad292e17106918a76cd66e87ae9f34d299) does it within `addurls`
  - unfortunately it is not 100% working yet due to a freshly discovered #5021 (but works on a simpler example without nesting within added unittest for ConsumerProducer)
- also includes 39b03e7818074c47a38f08751ed2feaf19570509 RF/ENH which is IMHO generally useful, and if agreed, I can submit it in a separate PR.  It removes log filtering and makes `ProgressHandler` redirect to the default handler for non-progress log messages AFTER cleaning up all progress bars, and then refreshing them after that.  That allows to avoid interference between regular log and progress logs bars.  Even in DEBUG heavy output, progress bar is nicely lingering at the bottom
   - the only possible intererences with progressbars which would remain AFAIK are 1). direct unhandled output to stdout/stderr from `run` processes; ~~2). results records.  For 2), I think if we could add a "log handler" which would not format them at all, and then still use ProgressHandler logger - we could have those records nicely displayed without interfering with progress bars even when some processes are still ongoing.~~ result records already go through `ui.message` which stops all progress bars first and then shows again (I just submitted a minor #5040 which unlikely relates).  There is no thread safety measures there, so I think it *might* (have not analyzed) interfere with clear/refresh progress bars done by `ProgressHandler` (is logging thread safe?)

- an aspect I wanted to retain is to keep generators generators, be it a producer (e.g. listing of subdatasets) or consumer (some operation on those subdatasets), while also providing progress bars. With this implementation, we will get progress report as soon as producer starts producing entries we process by a consumer, ~~and only whenever producer exhausts itself, and thus we know total number -- we get a proper progressbar~~ as of 0ce386ea0c7ee12537e95e14c83ce7d07906efa0 progress bar appears right away but total might grow as well as we get more stuff to do.

<details>
<summary>demo 1: simple</summary> 

[...]

PR #5051, opened 2020-10-18
BF: push - do not call localsync unless annex

yarikoptic:
Discovered by chance while troubleshooting failing
https://github.com/datalad/datalad/pull/5022
which due to "out of order" completion of get swapped datasets between
noannex and annexed.  Was fund to debug (so added descriptive temp dirs suffixes) .... heh ;)

While at it spotted unused kwarg in `_push`.  Removed - no need to keep and must not break anything


mih (approved):
LGTM, thx!

PR #5052, opened 2020-10-19
ENH: added pragma to allow for no test coverage in case of ftp downloader on url which at times 503s

yarikoptic:
- [x] figure out what is the coverage diff
- [x] fix up (move/add tests or add pragma)


yarikoptic:
current coveralls red is because those lines are no longer covered (pragma) thus coverage reduced.  I think there is no better way, and it is minor, so I will just merge to not keep queue of PRs growing.

PR #5060, opened 2020-10-21
OPT: Rate limit result suppression message update

mih:
Updating it for each result is the cause of the slow-down reported in
https://github.com/datalad/datalad/pull/4868#issuecomment-688417401

With this change we only update at max 2Hz. For the original scenario
this results in a ~30% speed-up:

```
datalad subdatasets -d .  4.30s user 1.99s system 101% cpu 6.225 total
datalad subdatasets -d .  6.33s user 2.27s system 100% cpu 8.538 total
```

### DIFF
--- a/datalad/interface/utils.py
+++ b/datalad/interface/utils.py
@@ -17,6 +17,7 @@ import logging
 import wrapt
 import sys
 import re
+from time import time
 from os import curdir
 from os import pardir
 from os import listdir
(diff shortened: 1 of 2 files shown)


datalad/interface/utils.py
```diff
@@ -514,16 +515,24 @@ def default_result_renderer(res):
                 if res.get('message', None) else ''))
 
 
-def _display_suppressed_message(nsimilar, ndisplayed, final=False):
+def _display_suppressed_message(nsimilar, ndisplayed, last_ts, final=False):
```
kyleam, line 518: This direction sounds good to me, but I guess that the rate-limiting shouldn't be in effect when `final=True`?

datalad/interface/utils.py
```diff
@@ -514,16 +515,24 @@ def default_result_renderer(res):
...
-                           single_or_plural("message has",
-                                            "messages have",
-                                            n_suppressed, False)),
-                   cr="\n" if final else "\r")
+        ts = time()
+        # rate-limit update of suppression message, with a large number
+        # of fast-paced results updating for each one can result in more
+        # CPU load than the actual processing
+        # arbitrarily go for a 2Hz update frequency -- it "feels" good
+        if last_ts is None or (ts- last_ts > 0.5):
```
yarikoptic, line 527:
```suggestion
        if last_ts is None or (ts - last_ts > 0.5):
```
my invaluable contribution ;-)

datalad/interface/utils.py
mih: Good point! Adjusted, thx.

PR #5066, opened 2020-10-21
OPT: completely noop .debug and .log if loglevel is set to >= INFO

yarikoptic:
I do not expect major boost of performance, but there might be some since
now there would be no check of logging level upon each log.debug etc.
Yet to try to get some real life benchmarking going (in a zoom ATM, any timing
would be screwed)

TODOs
- [ ] verify that there is some notable benefit, if not -- not worth it

### DIFF
--- a/datalad/log.py
+++ b/datalad/log.py
@@ -10,8 +10,9 @@
 from functools import partial
 import logging
 import os
-import sys
 import platform
+import sys
+import types
 import logging.handlers
 
 from os.path import basename, dirname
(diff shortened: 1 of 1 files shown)


yarikoptic:
FWIW: benchmarks on github actions [show no change](https://github.com/datalad/datalad/pull/5066/checks?check_run_id=1288819447#step:8:1)

yarikoptic:
Let's assume it is not worth it

Issue #5068, opened 2020-10-22
AnnexRepo.get_file_key() broken on windows


mih:
This isn't obvious, because the tests are disabled on Windows. Associated functionality is:

- metadata aggregation and query
- annex copy-to
- annex file-backend query
- availability test
- archive special remote (this might explain some of the ill behavior on windows #5015)
- figshare export
- test helper `ok_file_under_git`

Numerous cases in the tests.

I propose to replace the implementation with a call to `AnnexRepo.get_content_annexinfo()`, and RF actual usage to see if individual file property queries (keys, backends, etc.) can be condensed into single "status" queries.

mih:
In today's mainline code, this method is no longer used anywhere outside its own tests, and it is deprecated. Although possibly still broken on windows, it does not matter anymore.

PR #5069, opened 2020-10-22
RF: AnnexRepo.get_file_key() as a compat-frontend for get_content_annexinfo()

mih:
The previous implementation did not work on windows, and the desired
information is readily provided by more commonly used methods.

Fixes gh-5068

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -2732,10 +2732,13 @@ class AnnexRepo(GitRepo, RepoInterface):
             For each file in input list indicates the used backend by a str
             like "SHA256E" or "MD5".
         """
-
         return [
-            self.get_key_backend(self.get_file_key(f))
-            for f in files
+            r.get('backend')
+            for r in self.get_content_annexinfo(
+                # we provide records for files to inspect. this bypasses
+                # any internal git-calls to acquire status info
+                init={self.pathobj / fn: {} for fn in files},
+                eval_availability=False).values()
         ]
 
     @property
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -1410,16 +1410,23 @@ class AnnexRepo(GitRepo, RepoInterface):
     def get_file_key(self, files, batch=None):
         """Get key of an annexed file.
 
+        This method is inefficient when called repeatedly for single files,
+        rather than once for a list of files.
+
+        When more annex file properties need to be queried in addition to the
+        file key, it is more efficient to use get_content_annexinfo(). This
+        method is a frontend for it, and selectively returns file keys only.
+
         Parameters
         ----------
         files: str or list
             file(s) to look up
         batch: None or bool, optional
-            If True, `lookupkey --batch` process will be used, which would
-            not crash even if provided file is not under annex (but directly
-            under git), but rather just return an empty string. If False,
-            invokes without --batch. If None, use batch mode if more than a
-            single file is provided.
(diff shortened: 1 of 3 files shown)


kyleam:
Test failures are due to three spots (two tests) needing `FileInGitError -> FileNotInAnnexError`.

mih:
Question is: should we achieve fully compatible behavior through additional inspection at a runtime cost? Or adjust the tests?

mih:
It seems `get_file_key()` is called in `is_available()` and might be replaced in a similar fashion.

mih:
Benchmarks report a 5-10% slowdown across the board. That is surprising. I fail to see this function being used in any core functionality...

https://github.com/datalad/datalad/pull/5069/checks?check_run_id=1292299793

So even if it is slower, the impact should be "not much".

kyleam:
> Benchmarks report a 5-10% slowdown across the board. That is surprising. I fail to see this function being used in any core functionality...
[...]
> So even if it is slower, the impact should be "not much".

Yeah, that doesn't seem plausible based on where it is used.  For the one-call case, `find --json FILE` doesn't appear to be much slower than `lookupkey FILE`.  For the many-separate-calls batch case, I think the difference could be quite large, though as far as I can see, `datalad.metadata.extractors.annex` is the only caller that uses `batch=True` like this.  (And of course none of that is relevant to the general benchmark slowdown.)

kyleam:
Keeping compatibility by raising `FileInGitError` would be nice (and the execution cost would only need to be paid when files aren't known to annex), but I think that's a minor issue compared to fixing this on Windows [*], and so no objections from me if you prefer to leave it as is.  I looked for current callers in -core that would need to be updated, and the only one I spot is `datalad.metadata.aggregate._the_same_across_datasets`.

Also, the docstring's description of `batch` should be updated to mention that it's an obsolete noop now.

[*] Just out of curiosity, presumably the lookupkey call itself was fine on Windows.  I guess it was the post-call handling on our side that was the issue?

mih:
It seems the initial benchmark results were a fluke, all normal now.

mih:
OK, so while I stand by these changes, I have found a windows box that still doesn't work. And probably, appveyor is also not completely fine, as I cannot find a trace of any annexrepo test actually being attempted (:roll_eyes: seriously, wtf!? we are not even anywhere close). I tried multiple git-annex versions, no change. I suspect it is still the preparation of the test repos that is severely fucked. Failure pattern looks like this:

```
AssertionError: 'SHA256E-s4--181210f8f9c779c26da1d9b2075bde0127302ee0e3fca38c9a83f5b1dd8e5d3b.dat' != 'SHA256E-s28--2795fb26981c5a687b9bf44930cc220029223f472cea0f0b17274f4473181e7b.dat'
```

where the former is a hard coded SHA to test against. The change in size suggests that we are looking at different content, and I suspect adjusted branches to be the cause.

mih:
Let's wait with this one until #5073 has landed and rerun the tests.

kyleam:
With gh-5080, `test_annexrepo` is executed on appveyor.

mih:
Rebased on master after merge of various test setup fixes on windows.

mih:
The key test is still not being attempted on appveyor, this time due to the broken testrepo setup (see #5093 for the state of things. spoiler: :vomiting_face: )

mih:
The test passes with #5303 https://ci.appveyor.com/project/mih/datalad/build/job/f7o54hjcfqfhs8e3#L644 without any required fix in the actual implementation -- just the broken test.

PR #5070, opened 2020-10-22
OPT+BF: annotate read_only git/annex invocations, thread lock read-write git/annex and git-config invocations

### LINKED ISSUE #5131: parallel install of subdatasets needs some thread locking while operating on superdataset config etc
```
datalad install ///openneuro && datalad install -r -J10 -d openneuro openneuro
```
leads to
```
|               datalad siblings -d "/mnt/scrap/tmp/openneuro/ds000121" enable -s s3-PRIVATE                                                                                                                                       
[WARNING] Received an exception CommandError: 'git config --local --replace-all submodule.ds000121.active true' failed with exitcode 255 under /mnt/scrap/tmp/openneuro [err: 'error: could not lock config file .git/config: File exists'] [cmd.py:run:514].                                                                                                                                                                                                         
| Canceling not-yet running jobs and waiting for completion of running.    
```
so most likely multiple threads are trying to adjust the status for the installed submodule within their superdataset. The most general solution of locking `.config` for any `set` operation.  Since it involves running external process, most likely performance impact from locking would be negligible in comparison but need to be assessed

edit: well, ideally even `get` should be locked to avoid race condition between `get` and `set` but imho it would be red herring since it would not resolve outside code which could lead to such a race which would involve the tandem of get and set, so locking then should be done somewhere there to ensure atomic operation of `set_if` or alike

yarikoptic:
old description which got it going:  While looking at py-spy stack from long running status I noticed that majority
of time is spent on git config call, triggered by checking for fake dates
config setting. AFAIK for read-only operations, which cannot produce a commit,
it might benefit to skip that check. And _call_git already has that kwarg,
but it is not proxied by "higher-level" helpers.  To avoid breeding workarounds
(like the one RFed within this commit) I decided that we might benefit from
exposing that option in higher-level helpers as well. Then "read-only" invocations could set it to False and we might avoid needless  config read in some cases (I have not done any timing on any prototypical use case which might benefit from this, but I would assume that status or diff might).

What is in the PR
- RF to make possible to make "lean" `call_git` read only invocations which would not even read config since not needed
- BF add thread locking non-read-only invocations of git/annex -- in `GitRepo.call_git` and `ConfigManager._run`.  That closes #5131 .  Now you can really animate your terminals with something like `datalad install -r -J50 /// ` ;-)

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -1621,8 +1621,7 @@ class AnnexRepo(GitRepo, RepoInterface):
         # reports a type change as modified.
         modified = [
             f for f in self.call_git_items_(
-                ['diff-files', '--name-only', '-z'], sep='\0',
-                read_only=True)
+                ['diff', '--name-only', '-z'], sep='\0')
             if f] if pointers else []
         annex_res = fn(files, normalize_paths=False, batch=batch)
         return [bool(annex_res.get(f) and
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -859,7 +859,8 @@ class AnnexRepo(GitRepo, RepoInterface):
         srs = {}
         try:
             for line in self.call_git_items_(
-                    ['cat-file', 'blob', 'git-annex:remote.log']):
+                    ['cat-file', 'blob', 'git-annex:remote.log'],
+                    check_fake_dates=False):
                 # be precise and split by spaces
                 fields = line.split(' ')
                 # special remote UUID
@@ -1620,7 +1621,8 @@ class AnnexRepo(GitRepo, RepoInterface):
         # reports a type change as modified.
         modified = [
             f for f in self.call_git_items_(
-                ['diff', '--name-only', '-z'], sep='\0')
+                ['diff', '--name-only', '-z'], sep='\0',
+                check_fake_dates=False)
             if f] if pointers else []
         annex_res = fn(files, normalize_paths=False, batch=batch)
         return [bool(annex_res.get(f) and
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -1621,7 +1621,7 @@ class AnnexRepo(GitRepo, RepoInterface):
         # reports a type change as modified.
         modified = [
             f for f in self.call_git_items_(
-                ['diff', '--name-only', '-z'], sep='\0',
+                ['diff-files', '--name-only', '-z'], sep='\0',
                 read_only=True)
             if f] if pointers else []
         annex_res = fn(files, normalize_paths=False, batch=batch)
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -860,7 +860,7 @@ class AnnexRepo(GitRepo, RepoInterface):
         try:
             for line in self.call_git_items_(
                     ['cat-file', 'blob', 'git-annex:remote.log'],
-                    check_fake_dates=False):
+                    read_only=True):
                 # be precise and split by spaces
                 fields = line.split(' ')
                 # special remote UUID
@@ -1622,7 +1622,7 @@ class AnnexRepo(GitRepo, RepoInterface):
         modified = [
             f for f in self.call_git_items_(
                 ['diff', '--name-only', '-z'], sep='\0',
-                check_fake_dates=False)
+                read_only=True)
             if f] if pointers else []
         annex_res = fn(files, normalize_paths=False, batch=batch)
         return [bool(annex_res.get(f) and
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -2209,11 +2209,12 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
           A non-zero exit is expected and should not be elevated above the
           DEBUG level.
         read_only : bool, optional
-          If set to True, we assume that command does not introduce changes to recorded
-          in git information (commits, heads, tags) and merely reads/outputs information
-          (note: git might still adjust caches and index).
-          If set to False, we would inspect if dates should be faked for dataset
-          commits and set up the Git environment to ensure that.
+          By setting this to True, the caller indicates that the command does
+          not write to the repository, which lets this function skip some
+          operations that are necessary only for commands the modify the
+          repository. Beware that even commands that are conceptually
+          read-only, such as `git-status` and `git-diff`, may refresh and write
+          the index.
 
         Returns
         -------
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -34,6 +34,7 @@ from os.path import (
 )
 
 import posixpath
+import threading
 from functools import wraps
 from weakref import (
     finalize,
@@ -880,6 +881,9 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
           C='/my/path'   => -C /my/path
 
         """
+        # A lock to prevent multiple threads performing write operations in parallel
+        self._write_lock = threading.Lock()
+
         if self.git_version is None:
             self._check_git_version()
 
@@ -2230,6 +2234,8 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
 
         out = err = None
         try:
+            if not read_only:
+                self._write_lock.acquire()
             out, err = run_gitcommand_on_file_list_chunks(
                 self._git_runner.run,
                 cmd,
@@ -2246,6 +2252,9 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
                                      paths=ignored.groups()[0].splitlines())
             lgr.log(5 if expect_fail else 11, str(e))
             raise
+        finally:
+            if not read_only:
+                self._write_lock.release()
 
         if err:
             for line in err.splitlines():
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -2209,11 +2209,11 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
           A non-zero exit is expected and should not be elevated above the
           DEBUG level.
         read_only : bool, optional
-          If False (default), do not assume that command does not introduce changes and
-          is merely out read/output information.
-          E.g., if not read_only, we would inspect if dates should be faked for dataset
-          commits and set up the Git environment to ensure that. Must be
-          disabled for repository initialization.
+          If set to True, we assume that command does not introduce changes to recorded
+          in git information (commits, heads, tags) and merely reads/outputs information
+          (note: git might still adjust caches and index).
+          If set to False, we would inspect if dates should be faked for dataset
+          commits and set up the Git environment to ensure that.
 
         Returns
         -------
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -1583,11 +1583,7 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
         if count:
             cmd.append('--count={:d}'.format(count))
 
-        # cannot use call_git_items_() which would perform fake_dates
-        # processing. Going one level deeper to avoid it, no date
-        # modification possible here
-        out, _ = self._call_git(cmd, check_fake_dates=False)
-        for line in out.splitlines():
+        for line in self.call_git_items_(cmd, check_fake_dates=False):
             props = line.split('\0')
             if len(fields) != len(props):
                 raise RuntimeError(
@@ -1810,7 +1806,8 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
         cmd.append('--')
         try:
             stdout = self.call_git(
-                cmd, expect_stderr=True, expect_fail=True)
+                cmd, expect_stderr=True, expect_fail=True,
+                check_fake_dates=False)
         except CommandError as e:
             if 'bad revision' in e.stderr:
                 raise ValueError("Unknown commit identifier: %s" % commitish)
@@ -1853,7 +1850,7 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
         if short:
             cmd.append('--short')
         try:
-            return self.call_git_oneline(cmd)
+            return self.call_git_oneline(cmd, check_fake_dates=False)
         except CommandError as e:
             if commitish is None:
                 return None
@@ -1868,6 +1865,7 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
                 ['rev-list', '-n1', 'HEAD'],
                 files=files,
                 expect_fail=True,
+                check_fake_dates=False,
             )
             commit = commit.strip()
             return commit if commit else None
@@ -1904,7 +1902,7 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
         cmd = ["log", "--format={}".format(fmt)]
         cmd.extend((options or []) + revrange + ["--"])
         try:
-            stdout = self.call_git(cmd, expect_fail=True)
+            stdout = self.call_git(cmd, expect_fail=True, check_fake_dates=False)
         except CommandError as e:
             if "does not have any commits" in e.stderr:
                 return []
@@ -1927,7 +1925,9 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
         # doesn't succeed if passed a full hexsha that is valid but doesn't
         # exist.
         return self.call_git_success(
-            ["rev-parse", "--verify", commitish + "^{commit}"])
+            ["rev-parse", "--verify", commitish + "^{commit}"],
+            check_fake_dates=False,
+        )
 
     def get_merge_base(self, commitishes):
         """Get a merge base hexsha
@@ -1953,7 +1953,8 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
             commitishes = commitishes + [self.get_active_branch()]
 
         try:
-            base = self.call_git_oneline(['merge-base'] + commitishes)
+            base = self.call_git_oneline(['merge-base'] + commitishes,
+                                         check_fake_dates=False)
         except CommandError as exc:
             if exc.code == 1 and not (exc.stdout or exc.stderr):
                 # No merge base was found (unrelated commits).
@@ -1977,7 +1978,8 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
         bool
         """
         return self.call_git_success(
-            ["merge-base", "--is-ancestor", reva, revb])
+            ["merge-base", "--is-ancestor", reva, revb],
+            check_fake_dates=False)
 
     def get_commit_date(self, branch=None, date='authored'):
         """Get the date stamp of the last commit (in a branch or head otherwise)
@@ -2012,7 +2014,8 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
           and the branch name otherwise.
         """
         try:
-            out = self.call_git(["symbolic-ref", "HEAD"], expect_fail=True)
+            out = self.call_git(["symbolic-ref", "HEAD"], expect_fail=True,
+                                check_fake_dates=False)
         except CommandError as e:
             if 'HEAD is not a symbolic ref' in e.stderr:
                 lgr.debug("detached HEAD in {0}".format(self))
@@ -2249,7 +2252,7 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
         return out, err
 
     def call_git(self, args, files=None,
-                 expect_stderr=False, expect_fail=False):
+                 expect_stderr=False, expect_fail=False, check_fake_dates=True):
         """Call git and return standard output.
 
         Parameters
@@ -2277,10 +2280,12 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
         """
         out, _ = self._call_git(args, files,
                                 expect_stderr=expect_stderr,
-                                expect_fail=expect_fail)
+                                expect_fail=expect_fail,
+                                check_fake_dates=check_fake_dates)
         return out
 
-    def call_git_items_(self, args, files=None, expect_stderr=False, sep=None):
+    def call_git_items_(self, args, files=None, expect_stderr=False, sep=None,
+                        check_fake_dates=True):
         """Call git, splitting output on `sep`.
 
         Parameters
@@ -2305,10 +2310,11 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
         ------
         CommandError if the call exits with a non-zero status.
         """
-        out, _ = self._call_git(args, files, expect_stderr=expect_stderr)
+        out, _ = self._call_git(args, files, expect_stderr=expect_stderr,
+                                check_fake_dates=check_fake_dates)
         yield from (out.split(sep) if sep else out.splitlines())
 
-    def call_git_oneline(self, args, files=None, expect_stderr=False):
+    def call_git_oneline(self, args, files=None, expect_stderr=False, check_fake_dates=True):
         """Call git for a single line of output.
 
         Parameters
@@ -2331,14 +2337,15 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
         AssertionError if there is more than one line of output.
         """
         lines = list(self.call_git_items_(args, files=files,
-                                          expect_stderr=expect_stderr))
+                                          expect_stderr=expect_stderr,
+                                          check_fake_dates=check_fake_dates))
         if len(lines) > 1:
             raise AssertionError(
                 "Expected {} to return single line, but it returned {}"
                 .format(["git"] + args, lines))
         return lines[0]
 
-    def call_git_success(self, args, files=None, expect_stderr=False):
+    def call_git_success(self, args, files=None, expect_stderr=False, check_fake_dates=True):
         """Call git and return true if the call exit code of 0.
 
         Parameters
@@ -2359,7 +2366,8 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
         """
         try:
             self._call_git(
-                args, files, expect_fail=True, expect_stderr=expect_stderr)
+                args, files, expect_fail=True, expect_stderr=expect_stderr,
+                check_fake_dates=check_fake_dates)
 
         except CommandError:
             return False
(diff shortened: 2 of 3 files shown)


kyleam:
> While looking at py-spy stack from long running status I noticed that majority
> of time is spent on git config call, triggered by checking for fake dates
> config setting.

Hmm, I'd need to look into this again, but that doesn't sound right (i.e. if that's happening, I think it's a bug).  For a repo instance, the value should be cached at `_fake_dates_enabled`.

> AFAIK for read-only operations, which cannot produce a commit,
> it might benefit to skip that check. And _call_git already has that kwarg,
> but it is not proxied by "higher-level" helpers.

Note that this was explicitly discussed when the original series was proposed: https://github.com/datalad/datalad/pull/3791#discussion_r335355781

kyleam:
> For a repo instance, the value should be cached at `_fake_dates_enabled`.

That appears to be working as expected, at least for a simple case.

<details>
<summary>test</summary>

```diff
diff --git a/datalad/support/gitrepo.py b/datalad/support/gitrepo.py
index eab25a48e8..d5af0c3733 100644
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -1603,6 +1603,7 @@ def fake_dates_enabled(self):
         """Is the repository configured to use fake dates?
         """
         if self._fake_dates_enabled is None:
+            print("--- config.getbool()")
             self._fake_dates_enabled = \
                 self.config.getbool('datalad', 'fake-dates', default=False)
         return self._fake_dates_enabled
```

```python
import tempfile
import datalad.api as dl

ds = dl.Dataset(tempfile.mkdtemp(prefix="dl-")).create()
ds.repo.call_git(["status"])
ds.repo.call_git(["status"])
```

```
[INFO   ] Creating a new annex repo at /tmp/dl-jwsbo0v0 
--- config.getbool()
[INFO   ] Scanning for unlocked files (this may take some time) 
```

</details>

yarikoptic:
> Hmm, I'd need to look into this again, but that doesn't sound right (i.e. if that's happening, I think it's a bug). For a repo instance, the value should be cached at `_fake_dates_enabled`.

yes, it is !!  But in my case it is LOTS (thousands) of small subdatasets.  So it does cache per dataset, but for a single dataset it doesn't need to do much ;) Actually, I was wrong about the status (or diff?): here is the flamegraph I got: http://www.onerussian.com/tmp/busy-datalad.svg  so it comes from `_save_add_submodules` in call to `get_hexsha` (read only, no need for fake date handling)
```
            subm_commit = subm.get_hexsha()
            if not subm_commit:
                yield get_status_dict(
                    action='add_submodule',
                    ds=self,
                    path=path,
                    status='error',
                    message=('cannot add subdataset %s with no commits', subm),
                    logger=lgr)
                continue
```
and then continues to the next one if not to be committed.  No need to consult config anywhere AFAIK, so no need to load all 4k of them if no submodule to be added.

yarikoptic:
hm, benchmarks reported
```
      7.75±0.04ms         9.30±1ms    ~1.20  support.path.get_parent_paths.time_allsubmods_toplevel
```
which is strange... I will restart to see if reproduces

yarikoptic:
took a longer py-spy trace: http://www.onerussian.com/tmp/busy-datalad-3.svg (for some reason no colors) -- different code path, just shows that: 1. caching of config worked and since it was first loaded, subsequent `fake_dates_enabled` does not trigger reload; 2. this PR will not be a "full solution" for that long running step since there is other long running steps (there is no logs at INFO or progressbar to tell them apart ATM).

mih:
Can we close this after #5076 was merged?

yarikoptic:
IMHO this one is largely orthogonal. Even though subdatasets use case was solved at higher level, and I don't have another affected use case, this pr changes make sense to me

kyleam:
Hmm, I'm on the fence about whether we should wait until having something concrete before deciding to expose this in each `call_git*` signature.  

If we do go with these changes, I think the commit message really should be updated to point to the previous discussion/decision on this and mention why it should be reversed (to account for the "many repos that wouldn't otherwise trigger a config read" usecase, IIUC).

yarikoptic:
yeap, adjusted commit message locally (not pushing for now) to

<details>
<summary>new commit msg</summary> 

    OPT: pass and use check_fake_dates kwargs through call call_git helpers
    
    While looking at py-spy stack from long running status I noticed that majority
    of time is spent on git config call, triggered by checking for fake dates
    config setting. AFAIK for read-only operations, which cannot produce a commit,
    it might benefit to skip that check. And _call_git already has that kwarg,
    but it is not proxied by "higher-level" helpers.  To avoid breeding workarounds
    (like the one RFed within this commit) I decided that we might benefit from
    exposing that option in higher-level helpers as well. Then "read-only" invocations
    could set it to False and we might avoid needless  config read in some cases
    (I have not done any timing on any prototypical use case which might benefit
    from this, but I would assume that status or diff might).
    
    We had an original discussion on either to expose check_fake_dates in higher
    level *Repo interfaces before:
    https://github.com/datalad/datalad/pull/3791#discussion_r335355781 and decided
    to not do it at that point.
    
    As for 'subdatasets' call, performance issue was now addressed with
    https://github.com/datalad/datalad/pull/5076 , so this change is not strictly
    necessary to optimize "subdatasets", but would still be generally benefitial
    for possible other invocations of *Repo methods across many instances without
    unnecessarily triggering loading of the config, typically in the case of
    read-only git operations (git-annex might need to merge git-annex branch, so
    typically annex calls should not disable fake dates).
</details>

but I wonder if any such RF to be done, should we may be introduce `read_only=False` instead? 
`check_fake_dates` is just current single (although nicely specific) operation, but it really boils down to either invoked `git` command is expected to be just "reading" or might "write" (anything which creates a commit -- `commit`, `merge`, etc) and thus then fake dates check should be invoked.

kyleam:
> I wonder if any such RF to be done, should we may be introduce `read_only=False` instead? [...]

Ah, good point.  I like that much better.

yarikoptic:
ok, redone it for `read_only=False` instead of `check_fake_dates=True` and then added locking on top for parallel execution in non-read_only mode in GitRepo and ConfigManager .  So it somewhat piled up two aspects - if desired can split the 2nd one in a follow up PR to keep our CIs warm.

yarikoptic:
heh, forgot that @mih just reported that crawler uses those older, now bad helpers,
```
  File "/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/site-packages/datalad_crawler/nodes/annex.py", line 909, in _commit
    check_fake_dates=True)
  File "/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/site-packages/datalad/support/gitrepo.py", line 329, in _wrap_normalize_paths
    result = func(self, files_new, *args, **kwargs)
TypeError: _git_custom_command() got an unexpected keyword argument 'check_fake_dates'
asyncio: DEBUG: Using selector: EpollSelector
asyncio: DEBUG: Using selector: EpollSelector
```
I guess will need to fix up/release there, but cannot since then it would need first datalad released, unless we do go through a deprecation cycle here  with a kludge for check_fake_dates... eh heh... sure thing I could provide signature based kludge in the crawler -- not sure what would be better ;-)

yarikoptic:
for the crawler -- picking up on @mih's https://github.com/datalad/datalad-crawler/pull/82 but unfortunately it caused infamous stalls of batched processes, which we thought are gone, and for which we have mitigation (abandon) only in master ATM... might need to bring it in also into `maint`.

yarikoptic:
fixedup crawler released, restarted workflows.

datalad/support/gitrepo.py
```diff
@@ -2111,7 +2118,7 @@ def _git_custom_command(self, files, cmd_str,
                             log_stdout=True, log_stderr=True, log_online=False,
                             expect_stderr=True, cwd=None, env=None,
                             shell=None, expect_fail=False,
-                            check_fake_dates=False,
+                            read_only=False,
                             index_file=None,
```
kyleam, line 2122: `_git_custom_command` is no longer used in the code base.  I think it'd be better to just leave it be.

kyleam:
With the lock changes, I've been unable to trigger the config issue from gh-5131.  As you say there, I also don't think using thread.Lock will add noticeable overhead.  So that part looks good to me.

datalad/support/gitrepo.py
yarikoptic: I am ok either way (since likely to be removed if not used I guess).  Let's see if any other changes needed and I will adjust

datalad/support/gitrepo.py
```diff
@@ -2201,8 +2208,10 @@ def _call_git(self, args, files=None, expect_stderr=False, expect_fail=False,
         expect_fail : bool, optional
           A non-zero exit is expected and should not be elevated above the
           DEBUG level.
-        check_fake_dates : bool, optional
-          If enabled (default), inspect if dates should be faked for dataset
+        read_only : bool, optional
+          If False (default), do not assume that command does not introduce changes and
```
kyleam, line 2212: I find the double negative hard to parse here.  When trying to come up with a rewrite, I stumbled on how to phrase it accurately.  I started to think about the larger issue with `read_only`: some commands in git are semantically read-only but actually modify the repo in ways that can clash with other running git commands.  For example, `status` and `diff` will refresh and write the index.  These writes are optimizations to avoid recomputing the same thing on the next git call.

In the case of `status`, the global `--no-optional-locks` option can be used to avoid these.  27344d6a6c (git: add --no-optional-locks option, 2017-09-27) makes it clear that this option doesn't cover (at least yet) all the commands that do such an "opportunistic" refresh.

So, while I think `read_only` (versus a more specific `check_fake_dates`) is appealing because it adds a single high-level knob that conceptually covers `check_fake_dates` (and now locking), my concern is that it's misleading because it doesn't reflect how git behaves underneath.

Anyway, perhaps not a reason to avoid adding `read_only`, but I think it's worth thinking about and making the docstring clearer.  Also, given the support for parallel operations now, our `status` calls in `.dirty` should use `--no-optional-locks`, and we should audit our `diff` calls to see if we can switch to a lower-level command (`diff-files`, ...) that doesn't refresh the index.  (But neither of those things need to be tied to this PR of course.)

As an unrelated note, this parameter should be documented in the other `call_git*` commands, perhaps with some fancy formatting like config's `_where_reload`, though I think nearby comments reminding future updaters to cover the other spots would do.

yarikoptic: yeah, I was not 100% happy with read_only as well, but other ones I thought about are either not generic enough (e.g, `no_commit`) or vague (e.g. `changes_state`).
Please suggest changes to the wording - I did think about making it also more specific, like "no commits etc", but just makes it longer for no good value.  We could add a disclaimer/note that "Note: even if 'read-only' commands could refresh index etc." may be, but again -- not 100% certain if such level of detail is really needed here

datalad/support/gitrepo.py
yarikoptic: ok, added reverting commit

datalad/support/gitrepo.py
yarikoptic: adjusted description a bit, hopefully it is better

datalad/support/gitrepo.py
```diff
@@ -2201,10 +2208,12 @@ def _call_git(self, args, files=None, expect_stderr=False, expect_fail=False,
...
-        check_fake_dates : bool, optional
-          If enabled (default), inspect if dates should be faked for dataset
-          commits and set up the Git environment to ensure that. Must be
-          disabled for repository initialization.
+        read_only : bool, optional
+          If set to True, we assume that command does not introduce changes to recorded
+          in git information (commits, heads, tags) and merely reads/outputs information
+          (note: git might still adjust caches and index).
+          If set to False, we would inspect if dates should be faked for dataset
+          commits and set up the Git environment to ensure that.
```
kyleam, line 2216: How about something like this?  I'm trying to make it clearer that `read_only=True` isn't appropriate for the cases where git does an opportunistic refresh.

```suggestion
          By setting this to True, the caller indicates that the command does
          not write to the repository, which lets this function skip some
          operations that are necessary only for commands the modify the
          repository. Beware that even commands that are conceptually
          read-only, such as `git-status` and `git-diff`, may refresh and write
          the index.
```

datalad/support/annexrepo.py
```diff
@@ -1620,7 +1621,8 @@ def _check_files(self, fn, files, batch):
         # reports a type change as modified.
         modified = [
             f for f in self.call_git_items_(
-                ['diff', '--name-only', '-z'], sep='\0')
+                ['diff', '--name-only', '-z'], sep='\0',
```
kyleam, line 1624: As mentioned in my other response, I think `git-diff` is an example of a command that may refresh the index and is not safe to mark as `read_only=True`, provided that we're using `read_only` to decide whether to take a lock for parallel operations.


```suggestion
                ['diff-files', '--name-only', '-z'], sep='\0',
```

datalad/support/gitrepo.py
kyleam: > I did think about making it also more specific, like "no commits etc", but just makes it longer for no good value.

Or it actually makes it incorrect.  The problem isn't that the parameter name is inaccurate; it's that the caller may think some commands are clearly read-only, but in fact they write to the repository.  In those cases, it wouldn't be appropriate to skip the locks needed for parallel operations.

yarikoptic:
Thank you @kyleam - applied adjusted wording (i like it), and switch from `diff`

kyleam (approved):
Thanks.  Looks good to me.

I think all of the `call_git*` docstrings should reference the parameter, perhaps just with `call_git` holding the actual text and the description for all of the other methods pointing to `call_git`s description.  But I can propose it in a follow-up PR (and either way it's not a big deal).

yarikoptic:
Restarted failed osx and appveyor - don't think the failures are related to pr but who knows

yarikoptic:
<details>
<summary>OSX failures persist so must be related since otherwise I think OSX runs are green.  grr...  I will restart once more (only OSX) and will try to look into it - this time recording a sample failed test for comparison</summary> 

```shell
FAIL: datalad.support.tests.test_annexrepo.test_AnnexRepo_file_has_content(False,)
  File "/Users/runner/hostedtoolcache/Python/3.6.12/x64/lib/python3.6/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/Users/runner/hostedtoolcache/Python/3.6.12/x64/lib/python3.6/site-packages/datalad/tests/utils.py", line 1090, in _wrap_with_testrepos
    t(*(arg + (uri,)), **kw)
  File "/Users/runner/hostedtoolcache/Python/3.6.12/x64/lib/python3.6/site-packages/datalad/tests/utils.py", line 755, in _wrap_with_tempfile
  File "/Users/runner/hostedtoolcache/Python/3.6.12/x64/lib/python3.6/site-packages/datalad/support/tests/test_annexrepo.py", line 319, in test_AnnexRepo_file_has_content
    [ar.supports_unlocked_pointers])
AssertionError: [False] != [True]
asyncio: DEBUG: Using selector: KqueueSelector
asyncio: DEBUG: Using selector: KqueueSelector
asyncio: DEBUG: Using selector: KqueueSelector
asyncio: DEBUG: Using selector: KqueueSelector
asyncio: DEBUG: Using selector: KqueueSelector
asyncio: DEBUG: Using selector: KqueueSelector
asyncio: DEBUG: Using selector: KqueueSelector
asyncio: DEBUG: Using selector: KqueueSelector
asyncio: DEBUG: Using selector: KqueueSelector
asyncio: DEBUG: Using selector: KqueueSelector
asyncio: DEBUG: Using selector: KqueueSelector
asyncio: DEBUG: Using selector: KqueueSelector
asyncio: DEBUG: Using selector: KqueueSelector
asyncio: DEBUG: Using selector: KqueueSelector
asyncio: DEBUG: Using selector: KqueueSelector
asyncio: DEBUG: Using selector: KqueueSelector
asyncio: DEBUG: Using selector: KqueueSelector
asyncio: DEBUG: Using selector: KqueueSelector
asyncio: DEBUG: Using selector: KqueueSelector
asyncio: DEBUG: Using selector: KqueueSelector
  File "/Users/runner/hostedtoolcache/Python/3.6.12/x64/lib/python3.6/site-packages/datalad/support/tests/test_annexrepo.py", line 354, in test_AnnexRepo_is_under_annex
    [ar.supports_unlocked_pointers])
AssertionError: [False] != [True]
asyncio: DEBUG: Using selector: KqueueSelector
asyncio: DEBUG: Using selector: KqueueSelector
asyncio: DEBUG: Using selector: KqueueSelector
asyncio: DEBUG: Using selector: KqueueSelector
asyncio: DEBUG: Using selector: KqueueSelector
asyncio: DEBUG: Using selector: KqueueSelector
asyncio: DEBUG: Using selector: KqueueSelector
asyncio: DEBUG: Using selector: KqueueSelector
asyncio: DEBUG: Using selector: KqueueSelector
asyncio: DEBUG: Using selector: KqueueSelector
asyncio: DEBUG: Using selector: KqueueSelector
asyncio: DEBUG: Using selector: KqueueSelector
asyncio: DEBUG: Using selector: KqueueSelector
asyncio: DEBUG: Using selector: KqueueSelector
asyncio: DEBUG: Using selector: KqueueSelector
asyncio: DEBUG: Using selector: KqueueSelector
asyncio: DEBUG: Using selector: KqueueSelector
asyncio: DEBUG: Using selector: KqueueSelector
asyncio: DEBUG: Using selector: KqueueSelector
asyncio: DEBUG: Using selector: KqueueSelector
asyncio: DEBUG: Using selector: KqueueSelector
asyncio: DEBUG: Using selector: KqueueSelector
```
</details>

yarikoptic:
FWIW, the same failures happen for me on OSX on our test OSX box when I run the same full selection of tests on master, but they do not happen if I run just a single test (e.g. `python -m nose -s -v --with-cov --cover-package datalad --cover-xml ../datalad/support/tests/test_annexrepo.py:test_AnnexRepo_file_has_content `).

yarikoptic:
it is "fun": apparently the important difference in either specifying by path or by module path, i.e. `python -m nose -s -v --with-cov --cover-package datalad --cover-xml datalad.support.tests.test_annexrepo:test_AnnexRepo_file_has_content` would fail whenever `python -m nose -s -v --with-cov --cover-package datalad --cover-xml ../datalad/support/tests/test_annexrepo.py:test_AnnexRepo_file_has_content` would be ok

yarikoptic:
grrr, f.cking bug ran away somehow.

<details>
<summary>here is some incomplete (forgot to copy the last line but I 99% certain it was True ;)) copy/paste I managed to catch into my clipboard history</summary> 

```shell
> /Users/yoh/proj/datalad/datalad/venvs/dev3/lib/python3.8/site-packages/datalad/support/tests/test_annexrepo.py(318)test_AnnexRepo_file_has_content()
-> eq_(ar.file_has_content(["test-annex.dat"], batch=batch),
(Pdb) p ar.file_has_content(["test-annex.dat"], batch=batch)
[False]
(Pdb) p ar.file_has_content(["test-annex.dat"], batch=batch)
[False]
(Pdb) p ar.file_has_content(["test-annex.dat"], batch=batch)
[False]
(Pdb) p batch
False
(Pdb) p ar.supports_unlocked_pointers
True
(Pdb) p ar.file_has_content(["test-annex.dat"], batch=batch)
[False]
(Pdb) p ar.file_has_content(["test-annex.dat"], batch=batch)
[False]
(Pdb) p ar
AnnexRepo('/private/var/folders/wr/944pwmkx3yg0r2sf1g3zpkrh0000gn/T/datalad_temp_test_AnnexRepo_file_has_content30u1oh06')
(Pdb) 
[1]+  Stopped                 DATALAD_TESTS_TEMP_KEEP=1 python -m nose -s -v --with-cov --cover-package datalad --pdb --pdb-failures --cover-xml datalad.support.tests.test_annexrepo:test_AnnexRepo_file_has_content
content30u1oh06evel) bash-3.2$ cd /private/var/folders/wr/944pwmkx3yg0r2sf1g3zpkrh0000gn/T/datalad_temp_test_AnnexRepo_file_has_ 
(dev3) (pyout-devel) bash-3.2$ git status
On branch master
Your branch is up to date with 'origin/master'.

Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
        typechange: test-annex.dat

(dev3) (pyout-devel) bash-3.2$ fg
DATALAD_TESTS_TEMP_KEEP=1 python -m nose -s -v --with-cov --cover-package datalad --pdb --pdb-failures --cover-xml datalad.support.tests.test_annexrepo:test_AnnexRepo_file_has_content (wd: ~/proj/datalad/datalad/__testhome__)
^[[A
p ar
AnnexRepo('/private/var/folders/wr/944pwmkx3yg0r2sf1g3zpkrh0000gn/T/datalad_temp_test_AnnexRepo_file_has_content30u1oh06')
(Pdb) p ar.file_has_content(["test-annex.dat"], batch=batch)

```
</details>

which suggests that somehow `git status` made things straight, i.e. so that `file_has_content(` started to report correct `True` after I ran `git status` outside.
My hypothesis is that with this OPT we managed to avoid some `git config` call which had similar "positive" effect and avoided some kind of a race condition somewhere... but not yet sure where/how 

<details>
<summary>note: appveyor failure is also about file_has_content - just in a different test</summary> 

```shell
FAIL: datalad.support.tests.test_annexrepo.test_AnnexRepo_web_remote
  File "C:\Miniconda37\lib\site-packages\nose\case.py", line 198, in runTest
    self.test(*self.arg)
  File "C:\Miniconda37\lib\site-packages\datalad\tests\utils.py", line 576, in _wrap_with_tree
    return t(*(arg + (d,)), **kw)
  File "C:\Miniconda37\lib\site-packages\datalad\tests\utils.py", line 688, in _wrap_serve_path_via_http
  File "C:\Miniconda37\lib\site-packages\datalad\tests\utils.py", line 755, in _wrap_with_tempfile
  File "C:\Miniconda37\lib\site-packages\datalad\support\tests\test_annexrepo.py", line 384, in test_AnnexRepo_web_remote
    ok_(ar.file_has_content(testfile))
AssertionError: None
asyncio: DEBUG: Using proactor: IocpProactor
```
</details>
may be it could be the one to track down instead...

yarikoptic:
on windows it boils down to the switch from `diff` to `diff-files` and here is the difference in the logged output
```
-datalad.cmd: DEBUG: Async run ['git', 'diff', '--name-only', '-z']^M
-datalad.cmd: DEBUG: Launching process ['git', 'diff', '--name-only', '-z']^M
-datalad.cmd: DEBUG: Process 19716 started^M
-datalad.cmd: DEBUG: Waiting for process 19716 to complete^M
-datalad.cmd: Level 5: Read 132 bytes from 19716[stderr]:^M
-datalad.cmd: Level 5: stderr| warning: LF will be replaced by CRLF in 127.0.0.1_about.txt.^M
-| The file will have its original line endings in your working directory^M
-|  ^M
-datalad.cmd: DEBUG: Process 19716 exited with return code 0^M
-datalad.cmd: Level 8: Finished running ['git', 'diff', '--name-only', '-z'] with status 0^M
-datalad.gitrepo: Level 11: stderr| warning: LF will be replaced by CRLF in 127.0.0.1_about.txt.^M
-datalad.gitrepo: Level 11: stderr| The file will have its original line endings in your working directory^M
+datalad.cmd: DEBUG: Async run ['git', 'diff-files', '--name-only', '-z']^M
+datalad.cmd: DEBUG: Launching process ['git', 'diff-files', '--name-only', '-z']^M
+datalad.cmd: DEBUG: Process 10040 started^M
+datalad.cmd: DEBUG: Waiting for process 10040 to complete^M
+datalad.cmd: Level 5: Read 20 bytes from 10040[stdout]:^M
+datalad.cmd: Level 5: stdout| 127.0.0.1_about.txt^@ ^M
+datalad.cmd: DEBUG: Process 10040 exited with return code 0^M
+datalad.cmd: Level 8: Finished running ['git', 'diff-files', '--name-only', '-z'] with status 0^M
``` 
and what confuses me is that int he case of `diff` (the one where test passes) there is no stdout logged at all, just all the warnings in the stderr (none of that in diff-files). any immediate clues @kyleam ?

kyleam:
> on windows it boils down to the switch from `diff` to `diff-files` and here is the difference in the logged output [...] any immediate clues @kyleam ?

No, but I can trigger it with `tools/eval_under_testloopfs`, so something with adjusted branches (maybe to due with filters).  I'll dig more.

kyleam:
Under `tools/eval_under_testloopfs`, here's the repo from `test_AnnexRepo_web_remote` immediately before the first `file_has_content` call:

```console
$ git diff-files -p
$ git diff-files
:100755 100755 6712b5d7d14789b99e55540b0dec07fe2fc0d421 0000000000000000000000000000000000000000 M	127.0.0.1_about.txt
$ git diff
$ git diff-files
```

So, we're switching to `diff-files` because it doesn't do a refresh of the index, but I suppose that that comes along that comes along with not triggering the clean filter.

In this case, I'd say it makes sense to switch this spot back `"diff", read_only=False`.

yarikoptic:
eh, ok, will do... since anything else (checking if we are in adjusted branch etc) would only take more time, thus ruining any OPT effect.  I would expect OSX also to turn green, but I still think it would only be hiding the issue - I think there is more to it there.

kyleam:
> since anything else (checking if we are in adjusted branch etc) would only take more time, thus ruining any OPT effect.

I don't think any of this is for this PR but...

Another option would be to switch to `ls-files --modified PATH`, which seems to correctly run the filter yet doesn't opportunistically refresh the index.  I don't think that's a step worth taking though, because in my view `file_has_content` should just be rewritten to use a combination of `diffstatus` (which calls `ls-files --modified`) and `get_content_annexinfo`.  Also, it seems like we should consider whether `get_content_annexinfo` ought to report that content is present if a file is modified, as it does now; if we decide it shouldn't, then we don't really need `file_has_content` and it can be reduced to a very thin (perhaps deprecated) wrapper around `get_content_annexinfo`.

And I think it's worth keeping in mind that avoiding the thread lock by disabling the opportunistic refresh of the index might not be a net gain because downstream the index update would need to be redone.  Perhaps tricky, but it'd bee good to measure in some way rather than assuming there's actually a benefit.

Unrelated to any of that: the fact that `ls-files --modified` and `diff-files` when combined with flags like `-p` and `-s` do seem to run the filter makes me think that `diff-files` not running the filter in some cases should be looked into more and reported upstream.

> I would expect OSX also to turn green, but I still think it would only be hiding the issue - I think there is more to it there.

I'm confused by this.  Based on you saying "OSX tests were failing even at 4d84e6e so before this switch to diff-files was introduced", I thought those were unrelated.

yarikoptic:
> > I would expect OSX also to turn green, but I still think it would only be hiding the issue - I think there is more to it there.
> 
> I'm confused by this. Based on you saying "OSX tests were failing even at [4d84e6e](https://github.com/datalad/datalad/commit/4d84e6ec063706a6c35e47cebf4b569d155c3e0c) so before this switch to diff-files was introduced", I thought those were unrelated.

but they were `read_only=True`.  My hypothesis [was](https://github.com/datalad/datalad/pull/5070#issuecomment-736907282) that some `git config` call we do in case of non-read_only does something "magically necessary" to avoid those failures.  And BTW indeed those OSX builds turned green!  Travis is still cranking, and crippledFS failures were closed eyes upon for awhile so would be too much effort to say if related ATM.

yarikoptic:
ok, let's proceed!

PR #5072, opened 2020-10-23
ENH(tool): downgrade-annex helper

yarikoptic:
Could have gone probably directly without PR but I do not to
disturb the force ;)

[ci skip]

### DIFF
--- a/tools/downgrade-annex
+++ b/tools/downgrade-annex
@@ -1,0 +1,77 @@
+#!/bin/bash
+
+set -eu
+
+export PS4='> '
+# set -x
+
+function info () {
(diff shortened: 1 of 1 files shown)


kyleam (commented):
Do you want to add this to the shellcheck ci job?  It bet it would have a recommendation for you.

yarikoptic:
> Do you want to add this to the shellcheck ci job? It bet it would have a recommendation for you.

but just one and thus I didn't add it to shellcheck, may be some time later

yarikoptic:
we have too many PRs, this might come handy even if not shellcheck-proof, merging

PR #5073, opened 2020-10-23
More windows testing

mih:
This is an effort to enable more testing on windows. In particular the aim is to have some testing working with a windows datalad installation that a user following the handbook would have -- rather than just within the somewhat twister appveyor setup.

All changes are cross-verified manually on a win10 box with an installation as described in https://github.com/datalad-handbook/book/pull/588

- Remove testrepo setup special case on windows https://github.com/datalad/datalad/pull/5069#issuecomment-715013343
- run all `datalad.tests` tests, mark individual ones as failing or skip them, if there is no hope -- this will help further breakage from creeping in #5074

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -83,7 +83,7 @@ test_script:
   # first sign of life
   - datalad wtf
   # and now this... [keep appending tests that should work!!]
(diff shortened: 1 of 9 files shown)


yarikoptic (approved):
FWIW, there once was a PR https://github.com/datalad/datalad/pull/2870/files which also addressed some windows fails with testrepos etc, but never was finished and thus closed.  Might have some commits to CP/adopt

Issue #5075, opened 2020-10-23
`subdatasets -r` listing: should be fast(er) 


yarikoptic:
haven't look in detail yet but IMHO `subdatasets` seems to be doing something `avoidable`.  if `subdatasets -r` supposed to list only "known" subdatasets, should it even "consult" `.config` of leaf datasets if there is no `.gitmodules` in the top directory of that leaf dataset? 

Here is the flame graph of running process where it was slowly listing leaf subdatasets: http://www.onerussian.com/tmp/busy-datalad-subdatasets-1.svg . So it seems (didn't look in detail) that .config is consulted by AnnexRepo in `__init__`, which I guess we instantiate to verify that subdataset is installed...  May be could be "avoided"?  (didn't look in detail yet)

Also,  If it is collecting all kinds of "useful information" which is otherwise disregarded at the end (e.g. whenever I just need to list subdatasets), we might look into making such "additional lookup" lazy (i.e. triggered only when some key is requested from the yielded record).

mih:
Can you please share which datalad version you were using, and whether it had #4868?

mih:
I have adjusted the title of the issue to make clear that this is about recursive processing.

yarikoptic:
FWIW, I think it had no #4868 merged , but I guess irrelevant since confirmed etc.  Thanks again!

Issue #4411, opened 2020-04-17
Essential tests are not attempted on appveyor


mih:
No idea. The call looks good:

```
python -m nose -s -v --with-cov --cover-package datalad datalad.core datalad.local datalad.distributed datalad.cmdline datalad.distribution datalad.interface datalad.support datalad.ui
```
it leads to
```
...
datalad.core.distributed.tests.test_push.test_push(False,) ... ok
datalad.core.distributed.tests.test_push.test_push(True,) ... ok
datalad.core.distributed.tests.test_push.test_force_datatransfer ... ok
...
```

but there is no evidence that `test_clone` was even touched: https://ci.appveyor.com/project/mih/datalad/builds/32243133/job/tdb62dq122n7rx6m (example)

mih:
Does anyone have any idea what is happening here, and why these tests are silently dropped. Given https://github.com/datalad/datalad/issues/2575 is is critical to test such a core piece, and it should be tests AFAICS, but isn't.

yarikoptic:
I would have tried a dedicated run with `datalad/core/distributed/tests/test_clone.py` only to see what it brings:  may be all those decorators `with_testrepos` etc one at a time make those to be ignored/not tested on windows? 
or - have you tried on a windows box?
although then I would have expected SkipTest, but may be some of those raises SkipTest at the level of creating a decorator thus skipping entire file?

mih:
So this issue is way more severe than I thought initially. If you check: https://ci.appveyor.com/project/mih/datalad/builds/35918345

I see no trace of

- `datalad.core.local.tests.test_run`
- `datalad.core.local.tests.test_save`
- `datalad.distributed.tests.test_clone`
- `datalad.support.tests.test_gitrepo`
- `datalad.support.tests.test_annexrepo`

(but other tests in these locations do run)

I stopped looking at that point, because if we are not testing those, we aren't testing anything.

I have no idea what is happening.

kyleam:
Based on a very non-systematic search, I tried to figure out when those dropped off.  It seems like they were around in gh-3976 but not gh-3975.  (gh-3976 was opened after gh-3975 but was not based on it.)  There are hits in the 3976 run but not the 3975 for all the cases above except test_annexrepo, but it looks like that [wasn't specified in the appveyor run][1] at the time.

3975: https://ci.appveyor.com/project/mih/datalad/builds/29850819/job/a5metsmn7c0a9oed
[...]

PR #5076, opened 2020-10-23
OPT: Delay Dataset.repo access until necessary

mih:
Results in substantial speedup for recursive processing, as datasets
without any submodules are rejected prior any Repo instantiation.

For a dataset with 42k subdatasets this means going from

```
datalad subdatasets -r  270.45s user 615.64s system 104% cpu 14:07.22 total
```

to

```
datalad subdatasets -r  5.82s user 0.69s system 101% cpu 6.422 total
```

Fixes gh-5075

### DIFF
--- a/datalad/local/subdatasets.py
+++ b/datalad/local/subdatasets.py
@@ -55,8 +55,9 @@ lgr = logging.getLogger('datalad.local.subdatasets')
 valid_key = re.compile(r'^[A-Za-z][-A-Za-z0-9]*$')
 
 
-def _parse_git_submodules(ds_pathobj, repo, paths):
+def _parse_git_submodules(ds, paths, cache):
     """All known ones with some properties"""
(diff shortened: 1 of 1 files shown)


yarikoptic:
Awesome! Thank you @mih! Now the problem is only that I need to figure out what to do with all that free time ;-)

mih:
> Awesome! Thank you @mih! Now the problem is only that I need to figure out what to do with all that free time ;-)

Hehe.... #4411 could take a creative thought :eyes:

yarikoptic:
FWIW, even our tiny benchmarks show the gain
```
-         560±7ms          442±4ms     0.79  api.supers.time_subdatasets_recursive
```
which is great, so we don't miss some pessimisation if any comes

yarikoptic (approved):
only cursory review of code is done -- but benefit is too big to not to just go forward ;)

Issue #4592, opened 2020-05-30
git 2.27.0~rc2-1 seems to break GitPython (2.26.2-1) ok


yarikoptic:
Tried to build 0.12.7 for debian and it failed with many failures like
```
======================================================================
ERROR: datalad.distribution.tests.test_update.test_update_git_smoke
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/build/datalad-0.12.7/.pybuild/cpython3_3.8_datalad/build/datalad/tests/utils.py", line 732, in newfunc
    return t(*(arg + (filename,)), **kw)
  File "/build/datalad-0.12.7/.pybuild/cpython3_3.8_datalad/build/datalad/tests/utils.py", line 732, in newfunc
    return t(*(arg + (filename,)), **kw)
  File "/build/datalad-0.12.7/.pybuild/cpython3_3.8_datalad/build/datalad/distribution/tests/test_update.py", line 181, in test_update_git_smoke
    target.update(recursive=True, merge=True), 1,
  File "/build/datalad-0.12.7/.pybuild/cpython3_3.8_datalad/build/datalad/distribution/dataset.py", line 500, in apply_func
    return f(**kwargs)
  File "/build/datalad-0.12.7/.pybuild/cpython3_3.8_datalad/build/datalad/interface/utils.py", line 492, in eval_func
    return return_func(generator_func)(*args, **kwargs)
  File "/build/datalad-0.12.7/.pybuild/cpython3_3.8_datalad/build/datalad/interface/utils.py", line 480, in return_func
    results = list(results)
  File "/build/datalad-0.12.7/.pybuild/cpython3_3.8_datalad/build/datalad/interface/utils.py", line 399, in generator_func
    for r in _process_results(
  File "/build/datalad-0.12.7/.pybuild/cpython3_3.8_datalad/build/datalad/interface/utils.py", line 527, in _process_results
    for res in results:
  File "/build/datalad-0.12.7/.pybuild/cpython3_3.8_datalad/build/datalad/distribution/update.py", line 174, in __call__
    for fr in _update_repo(ds, sibling_, reobtain_data):
  File "/build/datalad-0.12.7/.pybuild/cpython3_3.8_datalad/build/datalad/distribution/update.py", line 240, in _update_repo
    repo.pull(remote=remote, refspec=active_branch)
[...]

PR #5081, opened 2020-10-23
ENH:  addurls(..., drop_after=True) to drop content after adding

yarikoptic:
We have already similar option for add_archive_content and a few places in the
   crawler.  It is a very much needed feature while crawling/addurl-ing big
   datasets, which might not even fit to local harddrive -- we need to drop
   content as soon as it is annexed (well -- later, may be after we extract
   metadata).  For now I am just adding it as a bool, but we might later
   indeed specify when to drop it (right after added, or after dataset saved
   and possibly metadata extracted via some hook or explicit operation etc).
    
   This PR is based on #5022 changes since that is where addurls was RFed
   to accomodate parallel execution, and I would prefer to not mess resolving
   conflicts later on

First I added a `drop` after all files are added, but then decided to not be lazy and RF'ed to
use `drop_key` in batch mode right after the file is addurl'ed.  So it is two last commits which are relevant, the rest is #4592 
Yet to try out on real case.


yarikoptic:
Osx fail is unrelated, needs to wait for #5022 to be merged, meanwhile I will test drive it along

kyleam (approved):

yarikoptic:
eh, forgot about this PR to only discover that we have not yet merged it whenever I tried to use it  (from master) :-/ since pertinent changes were actually small and blessed by @kyleam back then, I will just proceed and merge without a fresh round of CI.  I will deal with consequences if somehow some failures would come about

Issue #5077, opened 2020-10-23
ria+ssh without git on the remote server


mikapfl:
#### What is the problem?

If I create a ria sibling on a server without git installed, it does not work.

I thought this should be possible because the handbook says:
> On the RIA store hosting infrastructure, only 7z is to be installed, if the archive feature is desired. Specifically, no Git, no git-annex, and no otherwise running daemons are necessary. If the RIA store is set up remotely, the server needs to be SSH-accessible.

#### What steps will reproduce the problem?

On a client:
```bash
$ datalad create test_remote
[INFO   ] Creating a new annex repo at /p/tmp/pflueger/dl-test/test_remote 
[INFO   ] Scanning for unlocked files (this may take some time) 
create(ok): /p/tmp/pflueger/dl-test/test_remote (dataset)
$ cd test_remote/
$ dd bs=1M count=1 if=/dev/random of=test_file
0+1 records in
0+1 records out
115 bytes copied, 0,00027092 s, 424 kB/s
$ datalad save -m 'asdf'
add(ok): /p/tmp/pflueger/dl-test/test_remote/test_file (file)
save(ok): /p/tmp/pflueger/dl-test/test_remote (dataset)
action summary:
  add (ok: 1)
  save (ok: 1)
$ datalad create-sibling-ria -s 'sria' ria+ssh://aix:/home/pflueger/sria
[INFO   ] create siblings 'sria' and 'sria-storage' ... 
CommandError: 'ssh -o ControlPath=/home/pflueger/.cache/datalad/sockets/891b01a8 aix 'cd /home/pflueger/sria/177/bda76-5d0c-4bbb-8d9b-ae8a6b366395 && git init --bare'' failed with exitcode 127
bash: git: command not found
```

#### What version of DataLad are you using (run `datalad --version`)? On what operating system (consider running `datalad wtf`)?

<details>
<summary>datalad wtf</summary>
```
# WTF
## configuration <SENSITIVE, report disabled by configuration>
## datalad 
  - full_version: 0.13.4
  - version: 0.13.4
## dataset 
  - id: 177bda76-5d0c-4bbb-8d9b-ae8a6b366395
  - metadata: <SENSITIVE, report disabled by configuration>
  - path: /p/tmp/pflueger/dl-test/test_remote
  - repo: AnnexRepo
## dependencies 
  - annexremote: 1.4.3
  - appdirs: 1.4.4
  - boto: 2.49.0
  - cmd:7z: 16.02
[...]

PR #5090, opened 2020-10-24
NF: create-sibling-ria --storage-sibling only

mih:
- add support for `storage_sibling='only'` which triggers the mode
  desired in gh-5077

- add an additional parameter `disable_storage__=None` which is
  used as a destination for the previous `--no-storage-sibling` cmdline
  option. Whenever it is used, as deprecation warning is issued

- introduce `--storage-sibling {on|off|only}` cmdline argument handling.
  This opens up straightforward handling of further storage sibling
  modes (different special remote types/configs) in the future -- in
  contrast to an alternative approach along the lines of a dedicated off
  switch for the repository sibling part.

From the perspective of the cmdline interface this deprecates
`--no-storage-sibling` in favor of `--storage-sibling`.

From the perspective of the Python API, an additional argument value of
`storage_sibling` is recognized.

Fixes gh-5077

TODO:

- [x] Obtain feedback from OP
- [x] Add test

### DIFF
--- a/datalad/distributed/create_sibling_ria.py
+++ b/datalad/distributed/create_sibling_ria.py
@@ -195,9 +195,9 @@ class CreateSiblingRia(Interface):
             sibling are created ([CMD: on CMD][PY: True|'on' PY]).
             Alternatively, creation of the storage sibling can be disabled
             ([CMD: off CMD][PY: False|'off' PY]), or a storage sibling
-            is created only and no Git sibling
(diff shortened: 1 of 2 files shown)


kyleam (approved):
I'm not that familiar with ria stuff, but this looks good to me.  I dropped git/git-annex/datalad from our datalad-ssh container to play around with this, and it worked very nicely.

Issue #5078, opened 2020-10-23
ENH: python -m datalad.install [COMPONENT[=version] [--adjust-bashrc] [-u|--upgrade] [-s|--scheme SCHEME]] [COMPONENT...]


yarikoptic:
We have https://github.com/datalad/datalad/blob/HEAD/tools/ci/install-annex.sh which came handy already on many occasions and used in our CI.  But it is a bash script, so CLI parsing is too basic, would not work on Windows natively, harder to unit-test etc. As its names says, it is only about `annex`.

For all means and purposes all systems we would care about would have `python` (be installed as part of some distribution,  or "natively"), so I think it would be beneficial to redo it in "pure" (no 3rd party libs outside of standard libraries) Python

First stage TODOs:
- [ ] redo in Python using only standard libs (so argparse, no click)
- [ ] place it as `datalad.install` submodule which would also support 
  - a script (thus `if __name__ == "__main__"`) mode
  - has a `__main__` with callout to the same as script mode `main`
- [ ] support only `git-annex` as `COMPONENT`

Use case: I could just `git clone git://github.com/datalad/datalad; python datalad/datalad/install.py git-annex` (adjust paths `/` to `\` if on windows), and have appropriate to the platform most recent (compatible) release of git-annex installed

Additional features to expand with:
- [ ] add `--upgrade` -- detect how `COMPONENT` is installed given environment (conda, native, apt) and upgrade accordingly
- [ ] add `miniconda` component: I keep copy pasting from the handbook... If with this helper I could automate -- that would be awesome!  (Having `--env NAME`  to mint a new/enter conda environment would also be a nice feature)
- [ ] add `datalad` component (for `--upgrade` would use `apt-get` or `pip` depending on how was installed)
- [ ] `--deps` (`conda install --only-deps`, `apt-get build-dep` -- not the same, but AFAIK closest we can do ATM)  -- install build dependencies for the component (if conda it is, `conda install --only-deps -c conda-forge datalad` to get myself all dependencies installed)
- [ ] support of [future](https://github.com/datalad/datalad-extensions/issues/46) "datalad-extensions-artifacts distribution"  (in particular for Windows ATM) 
- [ ] add `venv` component (may be, as an alternative to `conda`)

[...]

PR #5098, opened 2020-10-26
Convert install-annex.sh to Python

jwodder:
Step 1 of #5078

### DIFF
--- a/datalad/install.py
+++ b/datalad/install.py
@@ -313,7 +313,7 @@ class GitAnnexInstaller:
                 "-c",
                 "conda-forge",
                 "-y",
-                f"git-annex{version or ''}",
+                f"git-annex={version}" if version is not None else "git-annex",
             ],
             check=True,
         )
--- a/datalad/install.py
+++ b/datalad/install.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#!/usr/bin/env python3
 __python_requires__ = "~= 3.6"
 import argparse
 from contextlib import contextmanager
--- a/datalad/install.py
+++ b/datalad/install.py
@@ -3,6 +3,7 @@ __python_requires__ = "~= 3.6"
 import argparse
 from contextlib import contextmanager
 from glob import glob
+import logging
 import os
 import os.path
 from pathlib import Path
@@ -16,6 +17,8 @@ import tempfile
 DOWNLOAD_LATEST_ARTIFACT = (
     Path(__file__).parent / "tools" / "ci" / "download-latest-artifact"
 )
+
+log = logging.getLogger("datalad.install")
 
 
 def main():
@@ -170,9 +173,10 @@ class GitAnnexInstaller:
                 check=True,
             )
         else:
-            print(
-                f"I: devel version {devel_annex_version} is not newer than installed {current_annex_version}",
-                file=sys.stderr,
+            log.info(
+                "devel version %s is not newer than installed %s",
+                devel_annex_version,
+                current_annex_version,
             )
         self.post_install()
 
@@ -206,7 +210,7 @@ class GitAnnexInstaller:
     def _install_via_autobuild_or_snapshot_linux(self, subpath):
         tmpdir = tempfile.mkdtemp(prefix="ga-")
         self.annex_bin = os.path.join(tmpdir, "git-annex.linux")
-        print("I: downloading and extracting under {self.annex_bin}", file=sys.stderr)
+        log.info("downloading and extracting under %s", self.annex_bin)
         wget = subprocess.Popen(
             [
                 "wget",
@@ -253,9 +257,8 @@ class GitAnnexInstaller:
         tmpdir = tempfile.mkdtemp(prefix="ga-")
         self.annex_bin = os.path.join(tmpdir, "annex-bin")
         if shutil.which("git-annex") is not None:
-            print(
-                "W: git annex already installed.  In this case this setup has no sense",
-                file=sys.stderr,
+            log.warning(
+                "git annex already installed.  In this case this setup has no sense"
             )
             sys.exit(1)
         # We are interested only to get git-annex into our environment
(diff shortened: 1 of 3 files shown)


datalad/install.py
```diff
@@ -0,0 +1,396 @@
...
+            self._install_via_autobuild_or_snapshot_linux("linux/current")
+        elif systype == "Darwin":
+            self._install_via_autobuild_or_snapshot_macos("OSX/current/10.10_Yosemite")
+        else:
+            raise RuntimeError(f"E: Unsupported OS: {systype}")
+
+    def _install_via_autobuild_or_snapshot_linux(self, subpath):
+        tmpdir = tempfile.mkdtemp(prefix="ga-")
+        self.annex_bin = os.path.join(tmpdir, "git-annex.linux")
+        print("I: downloading and extracting under {self.annex_bin}", file=sys.stderr)
```
yarikoptic, line 209: Let's use logging, request a logger for `datalad.install` and if there is no handlers in that logger's "root"  (thus not imported from datalad) it would establish the simple default one (probably dumping to stderr).  This way it would stay nicely consistent with datalad logging when used from "inside" of datalad and not as a script, and would still log nicely if used as a script.

datalad/install.py
```diff
@@ -0,0 +1,396 @@
...
+        self._path = None
+        self.annex_bin = "/usr/bin"
+        self.adjust_bashrc = adjust_bashrc
+        self.env_write_file = Path(env_write_file)
+
+    def setpath(self, newpath):
+        self._path = newpath
+        if self.env_write_file is not None:
+            with self.env_write_file.open("a") as fp:
+                print(f"PATH={quote(newpath)}", file=fp)
```
yarikoptic, line 93: shouldn't we  `export PATH` ?

yarikoptic: also why just not to make this `addpath` and thus have here
```suggestion
                print(f"export PATH={quote(newpath)}:$PATH", file=fp)
```
so outside could `source` the file possibly from different environments (thus possibly having different PATH set).  E.g. I already used this installed with snapshots of git-annex within virtualenv and outside etc.

yarikoptic (changes requested):
Woohoo!  Thanks! I left some requests. 
But also, since tests would run anyways, in the next batch of pushed commits (datalad CI run takes awhile, so better be discrete with pushes ;)), just adjust all the CI scripts to use this one instead of prior bash version or manual invocation if present. That would immediately test it as well (I guess we might come up with some mocked up or even a few integration tests later on).

jwodder:
@yarikoptic I could only find two direct uses of `install-annex.sh` in the repo (one in the `test_macos.yml` workflow, the other in `.travis.yml`).  There's also `tools/ci/prep-travis-devel-annex.sh`, which is almost like the `neurodebian-devel` schema, except it fails if the devel version is not newer than the installed version, but that script isn't even used by anything in the repo.  Are there any other scripts you are aware of that need to be updated to use `install.py`?

datalad/install.py
jwodder: Unfortunately, the conda-forge-last schema needs to add its path directory to the *end* of `PATH`.  The comment from the original source reads:

> We are interested only to get git-annex into our environment  
> So as to not interfere with "system wide" Python etc, we will add  
> miniconda at the end of the path

yarikoptic: that is ok, add the option `last=False` and append to `$PATH` instead of prepending `if last`

yarikoptic:
probably only those which you already found ATM.

datalad/install.py
```diff
@@ -0,0 +1,399 @@
+#!/usr/bin/python3
```
yarikoptic, line 1:
```suggestion
#!/usr/bin/env python3
```
?

datalad/install.py
```diff
@@ -0,0 +1,399 @@
...
+            )
+            sys.exit(1)
+        # We are interested only to get git-annex into our environment
+        # So as to not interfere with "system wide" Python etc, we will add
+        # miniconda at the end of the path
+        self.setpath(f"{os.environ['PATH']}:{self.annex_bin}")
+        self._install_via_conda(version, tmpdir)
+
+    def _install_via_conda(self, version, tmpdir):
+        miniconda_script = "Miniconda3-latest-Linux-x86_64.sh"
```
yarikoptic, line 277: note: Linux specific ATM. Later will need to be RFed to support Windows and OSX

datalad/install.py
jwodder: Done.

yarikoptic (commented):
shouldn't shebang use `/usr/bin/env`?

datalad/install.py
jwodder: Done.

jwodder:
@yarikoptic Some of the Travis tests are failing because they're using Python 3.5, yet I wrote this script for 3.6.  Python 3.5 reached EOL at the end of September; are you planning to drop support from Datalad, or do I have to give up my precious f-strings?

datalad/install.py
jwodder: Done.

yarikoptic:
oh, sorry -- I missed that. I will change base to become `master` - we dropped 3.5 in there already, and hopefully 0.14.0 is not too far away, so let's look forward and not loose precious `f-strings`

yarikoptic:
travis isn't happy due to seems missing `=`:
```
subprocess.CalledProcessError: Command '['/tmp/ga-tsoa94ls/miniconda/bin/conda', 'install', '-q', '-c', 'conda-forge', '-y', 'git-annex7.20190819']' returned non-zero exit status 1.
```
in the last argument between git-annex and its version to install .  Didn't check what appveyor is complaining about.

jwodder:
@yarikoptic I fixed the `=` bit.  Appveyor was failing due to running on Python 3.5.  The change of base branch seems to have fixed that, though I also rebased my branch just to be sure.

yarikoptic:
Great, thank you @jwodder ! 

@datalad/developers - any objections on having this merged?

Possible TODOs I have spotted for later:
- make it possible to pass `args` to `main`/arg parser  so that `main` could be tested and used within Python
  - add some smoke or unit-tests for the code (possibly mocking out actual calls to installers where possible)
- check if any extensions etc make use of `remove tools/ci/install-annex.sh` in their CI setups, and if not -- remove it to avoid confusion etc

yarikoptic (approved):

kyleam:
> any objections on having this merged?

None from me.

Issue #5100, opened 2020-10-27
"Cannot run the event loop while another loop is running" if imported from within another async process


yarikoptic:
possibly relates to #5064 although manifests slightly differently (thus filed separately).

I am "exploring" pycrunch , see https://github.com/gleb-sevruk/pycrunch-engine/pull/29 -- it would be nice to be able to tell which parts of code any particular test is affecting, and thus when fixing a bug also run locally only tests possibly affected by that particular code path, and actually have those tests automagically reran.

But immediate fiasco is inability to import datalad's modules (that is how pycrunch discovers tests ATM, see related https://github.com/gleb-sevruk/pycrunch-engine/pull/29) from within pycrunch:
```
[1076264] 2020-10-27 13:00:17,905 - pycrunch.discovery.simple - WARNING - importing datalad.distributed.tests.test_ria_basics
[1076264] 2020-10-27 13:00:17,905 - pycrunch.discovery.simple - ERROR - Failed to load `datalad/distributed/tests/test_ria_basics.py`
[1076264] 2020-10-27 13:00:17,905 - pycrunch.discovery.simple - ERROR - importing datalad.distributed.tests.test_ria_basics failed with exception: Cannot run the event loop while another loop is running
Traceback (most recent call last):
  File "/home/yoh/proj/misc/pycrunch-engine/pycrunch/discovery/simple.py", line 104, in find_tests_in_folder
    module = importlib.import_module(module_name)
  File "/usr/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 961, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 961, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 961, in _find_and_load_unlocked
[...]

PR #5101, opened 2020-10-27
RF: make external versions return UnknownVersion which compares False

yarikoptic:
(unless the same) if external version cannot be deduced

Then such comparisons as  '0.1' < external_versions['blah'] < '0.2'
used in skip_if decorators would return False instead of just crashing.
This happens now e.g. due to
https://github.com/datalad/datalad/issues/5100

### DIFF
--- a/datalad/support/external_versions.py
+++ b/datalad/support/external_versions.py
@@ -38,6 +38,18 @@ class UnknownVersion:
             return 0
         raise TypeError("UNKNOWN version is not comparable")
 
+    def __eq__(self, other):
+        return other is self
+
+    def __ne__(self, other):
+        return other is not self
+
+    def __lt__(self, other):
+        return False
+
+    __gt__ = __lt__
+    # True only if the same, the rest -- False
+    __le__ = __ge__ = __eq__
 
 #
 # Custom handlers
@@ -238,7 +250,7 @@ class ExternalVersions(object):
                 except Exception as exc:
                     lgr.debug("Failed to deduce version of %s due to %s"
                               % (modname, exc_str(exc)))
-                    return None
+                    return UnknownVersion()
             else:
                 if module is None:
                     if modname not in sys.modules:
--- a/datalad/support/tests/test_external_versions.py
+++ b/datalad/support/tests/test_external_versions.py
@@ -36,8 +36,10 @@ from datalad.tests.utils import (
     assert_greater_equal,
     assert_in,
     assert_not_in,
+    assert_not_equal,
     assert_raises,
     assert_true,
+    assert_false,
 )
 from datalad.tests.utils import SkipTest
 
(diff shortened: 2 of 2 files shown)


kyleam (commented):
Hmm, I'm not sure what I think of this yet, but it has some rough edges (or at least is disturbing/shifting already rough edges).

> Subject: [PATCH] RF: make external versions return UnknownVersion which
>  compares False (unless the same) if external version cannot be deduced
> 
> Then such comparisons as  '0.1' < external_versions['blah'] < '0.2'
> used in skip_if decorators would return False instead of just crashing.

This looks like it only achieves the intended result for a particular case (the name is in `CUSTOM` and an exception is raised while determining the versing), while None is still returned in other scenarios, including the example in the message:

```
$ python -c "from datalad.support.external_versions import external_versions as ev; print(ev['blah'])"
None
```

Based on the above description, I'm guessing that's not intended.

Also, in the comments and docstring, there is discussion of the None return value, which should be updated if we move in this direction.

At least one caller that explicitly handles `None` is `.check`, and this PR changes its handling for missing custom packages:

```console
$ # on master
$ python -c "from datalad.support.external_versions import external_versions as ev; print(ev.check('cmd:7z'))"
  File "/home/kyle/src/python/datalad/datalad/support/external_versions.py", line 367, in check
    name, ver=min_version, msg=msg)
datalad.support.exceptions.MissingExternalDependency: cmd:7z is missing.
$ # on pr
$ python -c "from datalad.support.external_versions import external_versions as ev; print(ev.check('cmd:7z'))"
None
```

And while `__contains__` doesn't explicitly handle it, the new return value makes it report the wrong answer since `bool(UnknownVersion()) -> True`.

```
$ 7z
zsh: command not found: 7z
$ # on master
$ python -c "from datalad.support.external_versions import external_versions as ev; print('cmd:7z' in ev)"
False
$ # on pr
$ python -c "from datalad.support.external_versions import external_versions as ev; print('cmd:7z' in ev)"
True
```

One spot in the external_versions tests also expects None:

```
FAIL: datalad.support.tests.test_external_versions.test_custom_versions
  File "/home/kyle/src/python/venvs/datalad/lib/python3.7/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/kyle/src/python/datalad/datalad/support/tests/test_external_versions.py", line 164, in test_custom_versions
    assert_equal(ev['bogus'], None)
AssertionError: <datalad.support.external_versions.UnknownVersion object at 0x7f0d2ac81748> != None
```

yarikoptic:
Thank you @kyleam ! I again don't like it ;-) I think I better not reuse unknown version indeed but come up with a similar but different class AbsentExternal which would be used where None is now returned, OR may be leave it as is, but in such comparisons would need to ensure that left version isn't just a string but a Version - that might work too... Will check later

yarikoptic:
let's forget about this for now

PR #5104, opened 2020-10-28
RF: lower level for maint clear/refres log_progress from ConsoleLog

yarikoptic:
I am not sure if that would not caused some undesired side-effects -- may be
lgr.info was under implicit assumption that all other log_progress are at lgr.info
level (or below).  But the reason why I am looking it is that in
https://github.com/dandi/dandisets/blob/master/tools/backups2datalad.py
we use datalad and have "our own" logger.  In interactive run I do see
those messages in the log (not progress bars I think) intermixed with dedicated
logger of that script, e.g.

    2020-10-28T11:04:02-0400 [INFO    ] backups2datalad.py Asset /sub-Templeton/sub-Templeton_ses-Templeton-032415_ecephys.nwb metadata unchanged; not taking any further action
    2020-10-28T11:04:02-0400 [INFO    ] datalad.ui.dialog Clear progress bars
    2020-10-28T11:04:02-0400 [INFO    ] datalad.ui.dialog Refresh progress bars
    2020-10-28T11:04:02-0400 [INFO    ] datalad.ui.dialog Clear progress bars
    2020-10-28T11:04:02-0400 [INFO    ] datalad.ui.dialog Refresh progress bars
    2020-10-28T11:04:02-0400 [INFO    ] backups2datalad.py No changes made to repository; deleting logfile
    2020-10-28T11:04:02-0400 [INFO    ] backups2datalad.py Syncing Dandiset 000042
    2020-10-28T11:04:03-0400 [INFO    ] dandi Traversing remote dandisets (000042) recursively
    2020-10-28T11:04:03-0400 [INFO    ] backups2datalad.py Updating metadata file
    2020-10-28T11:04:03-0400 [INFO    ] datalad.ui.dialog Clear progress bars
    2020-10-28T11:04:03-0400 [INFO    ] datalad.ui.dialog Refresh progress bars
    2020-10-28T11:04:03-0400 [INFO    ] datalad.ui.dialog Clear progress bars
    2020-10-28T11:04:03-0400 [INFO    ] datalad.ui.dialog Refresh progress bars
    2020-10-28T11:04:03-0400 [INFO    ] backups2datalad.py No changes made to repository; deleting logfile

so at large they provide no relevant information and just pollute the display.
May be an alternative solution is to hardcode to log any "maint" messages at
the non-interactive level (set to 5)?

PS ATM positioned against master since I do not know if would cause side-effects

### DIFF
--- a/datalad/ui/dialog.py
+++ b/datalad/ui/dialog.py
@@ -72,12 +72,12 @@ class ConsoleLog(object):
 
     def message(self, msg, cr='\n'):
         from datalad.log import log_progress
-        log_progress(lgr.info, None, 'Clear progress bars', maint='clear',
+        log_progress(lgr.debug, None, 'Clear progress bars', maint='clear',
                      noninteractive_level=5)
         self.out.write(msg)
         if cr:
             self.out.write(cr)
-        log_progress(lgr.info, None, 'Refresh progress bars', maint='refresh',
+        log_progress(lgr.debug, None, 'Refresh progress bars', maint='refresh',
                      noninteractive_level=5)
 
     def error(self, error):


kyleam:
> I am not sure if that would not caused some undesired side-effects

Unfortunately it does.  With the default INFO level, it results in the progress bar not being cleared (i.e. it effectively reverts gh-4555).

<details>
<summary>demo</summary>

```python
import time
import logging

from datalad.ui import ui

lgr = logging.getLogger("datalad.foo")
nthings = 4

log_progress(lgr.info, "foo",
             "Checking %d things", nthings,
             label="Checking things", total=nthings, unit=" things")
for i in range(nthings):
    log_progress(lgr.info, "foo",
                 "Checking thing %d", i,
                 increment=True, update=1)
    if i == 2:
        ui.message("Two")
    time.sleep(1)

log_progress(lgr.info, "foo",
             "Finished checking %d things", nthings)
```

On master, the end result is:

```
Two
```

On this PR

```
Checking things:  75%|███████████████████▌      | 3.00/4.00 [00:02<00:00, 1.54 things/s]
Two
```

</details>


> we use datalad and have "our own" logger

Okay, so none of the setup by `LoggerHelper.get_initialized_logger` would be in place, including configuring `ProgressHandler` in the interactive case or `filter_noninteractive_progress` in the non-interactive.  While I agree that datalad's logging isn't following [best practices][1] when used as a library, I'm not seeing an easy solution with the current design.  For a non-ideal solution to the immediate problem, I think the script could add `datalad.log.filter_noninteractive_progress` as a filter.

[1]: https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library

yarikoptic:
> Okay, so none of the setup by `LoggerHelper.get_initialized_logger` would be in place, including configuring `ProgressHandler` in the interactive case or `filter_noninteractive_progress` in the non-interactive.

ah, right... so if outside python script logs while consuming some yielded from inside the datalad's `log_progress`'ed loop, there would be "side-effects".  Haven't thought about that, good point... needs investigation if we could/should amend base logger's handler to trigger our progressbars cleanup/refresh if there is a log message.  But that would nohow solve the issue if outside code just "prints", so would be a partial solution anyways (unless we also overload `print`?  too much), eh.

yarikoptic:
Just FTR I also observe bunch (one per subdataset) 'nothing to save, working tree clean' `ui.message` from `status`  (its custom renderer, probably invoked by `save`) which is not really desired, but that is yet another issue (to file/explore).

yarikoptic:
ok, as is this PR is of no value, and the most valuable is the @kyleam 's code snippet in https://github.com/datalad/datalad/pull/5104#issuecomment-718830958  so I will close it for now

PR #5109, opened 2020-10-30
Run all tests on macOS

jwodder:
This PR is mainly just to see what fails as a result of enabling all tests.


jwodder:
No longer needed.

Issue #5112, opened 2020-11-02
Windows network drives are not supported (repo realpath breaks every runner CWD, and more)


mih:
I have a machine that has the user home "mounted" as a network drive `Z:` that maps to `\\BNBDATALAD\Users\mi, where `BNBDATALAD` is the name of the machine and am working on (i.e. "localhost").

I can

```
Z:
cd TMP
md some
cd some
git init
git annex init
```

with no problems. However I cannot create a dataset with datalad:

```
(base) Z:\TMP>datalad create this
[INFO] Creating a new annex repo at Z:\TMP\this
[INFO] Detected a filesystem without fifo support.
[INFO] Disabling ssh connection caching.
[INFO] Detected a crippled filesystem.
[INFO] git-annex: .git\hooks\: openTempFile: does not exist (No such file or directory)
[INFO] git-annex: init: 1 failed
CommandError: '"git" "annex" "init" "-c" "annex.dotfiles=true"' failed with exitcode 1 under \\BNBDATALAD\Users\mih\TMP\this
```

This smells like a path handling bug. Possibly, git-annex is not executed in `Z:\TMP`.

More evidence in this regard: The runner is not instructed to work under `Z:`

```
(Pdb) p cwd
'\\\\BNBDATALAD\\Users\\mih\\TMP\\this'
```

and the reason for that is that the repo doesn't believe it is under `Z:`

```
(Pdb) p self.path
'\\\\BNBDATALAD\\Users\\mih\\TMP\\this3'
```

I suspect that is an effect of `realpath`'ing the repo location.... Jup:

```
(Pdb) realpath('Z:')
'\\\\BNBDATALAD\\Users\\mih\\TMP'
```

And such a path is not usable in `cmd.exe`:

```
(base) Z:\TMP>cd \\\\BNBDATALAD\\Users\\mih\\TMP
"\\\\BNBDATALAD\\Users\\mih\\TMP"
CMD unterstützt keine UNC-Pfade als aktuelles Verzeichnis.
```

(German for "hell, no!")

Conclusion: no `realpath`'ing on Windows, please.

PR #5113, opened 2020-11-02
BF: Prevent Repo to resolve network mount to UNC paths on windows

mih:
Fixes gh-5112

No idea how to programatically create a network mount on appveyor, hence fixing without a test for now. Left extensive note to prevent accidental removal.

### DIFF
--- a/datalad/support/repo.py
+++ b/datalad/support/repo.py
@@ -189,7 +189,15 @@ class PathBasedFlyweight(Flyweight):
         # physical repository at a time
         # do absolute() in addition to always get an absolute path
         # even with non-existing paths on windows
-        return str(ut.Path(path).resolve().absolute())
+        resolved = str(ut.Path(path).resolve().absolute())
+        if ut.on_windows and resolved.startswith('\\\\'):
+            # resolve() ended up converting a mounted network drive into a UNC path.
+            # such paths are not supoprted (e.g. as cmd.exe CWD), hence redo and take
+            # absolute path at face value. This has the consequence we cannot determine
+            # repo duplicates mounted on different drives, but this is no worse than
+            # on UNIX
+            return str(ut.Path(path).absolute())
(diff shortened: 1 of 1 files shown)


mih:
Appveyor timed out.

kyleam (approved):

kyleam:
> Appveyor timed out.

I guess this might start to be a somewhat regular annoyance after gh-5080 :/  Should we split tests across the runs?

yarikoptic:
or just switch from appveyor to github actions? although split might be easier ATM and in particular for `maint`

mih:
> or just switch from appveyor to github actions? although split might be easier ATM and in particular for `maint`

We have used github actions rather unsuccessfully in the past. Primary reason for that is that they are a bitch to debug and behave even more different from a real win10 box than appveyor. I.e. placement of files across drives https://github.com/datalad/datalad/issues/5116.

I am not saying that those are not worth fixing, but moving to github action doesn't yield passing tests per se.

mih:
Thx for the review @kyleam

yarikoptic:
FWIW

>  I.e. placement of files across drives #5116.

like it could happen in real life , so IMHO only a + . debugging is a b.tch indeed.  
But if we get setup consistent so could be reproduced locally - so far I think I was able to reproduce windows issues I saw on github actions in a local VM. 
FWIW, https://github.com/datalad/git-annex/pull/7/files (after git-annex stops to FTBFS) provides workflows for all OSes, and will be merged and later extended to run datalad tests on windows as well. As on Linux and OSX the target will be `        version: [master, maint, release]` for datalad.  It would be nice to workout a way to avoid code duplication and "absorb" datalad testing specific components of that workflow also here, or in general look how we could harmonize/centralize "test datalad" github actions between that repository and this one (here - against our "base" git-annex version, there - against freshly built).

PR #5119, opened 2020-11-03
BF: workarounds to tests on Windows

yarikoptic:
No real fixes since should be done either on git-annex level (misbehaving locking when path is too long) or us RFing tests to rely on capabilities of the TMPDIR instead of `if on_windows`, so might even start failing where was working before.

BUT with today's git-annex (soon to be built on CRON job, will upload to packages/ tomorrow) `datalad/customremotes/tests/test_archives.py` tests all pass for me on windows VM.

also I gift you the mighty `which.py` which you might love in a moment of despair.


mih (approved):
Cool, thx!

Issue #5097, opened 2020-10-26
unable to git clone /// on windows


yarikoptic:
Tried in git bash and conda (conda installed `git`), getting the same
```
(base) C:\tmp>git clone http://datasets.datalad.org/.git
Cloning into 'datasets.datalad.org'...
Fetching objects: 11320, done.
error: invalid path 'con'
fatal: unable to checkout working tree
warning: Clone succeeded, but checkout failed.
You can inspect what was checked out with 'git status'
and retry with 'git restore --source=HEAD :/'
```

anyone saw such one, has a clue on WTF, or why `con` is so special?

<details><summary>WTF within conda env</summary>

# WTF
## credentials 
  - keyring: 
    - active_backends: 
      - Windows WinVaultKeyring
      - Windows RegistryKeyring
      - EncryptedKeyring with not defined v.1.0 at C:\Users\DataLad\AppData\Local\Python Keyring\wincrypto_pass.cfg
      - PlaintextKeyring with no encyption v.1.0 at C:\Users\DataLad\AppData\Local\Python Keyring\keyring_pass.cfg
    - config_file: C:\Users\DataLad\.config\python_keyring\keyringrc.cfg
    - data_root: C:\Users\DataLad\AppData\Local\Python Keyring
## datalad 
  - full_version: 0.13.4.dev374-g635b6
  - version: 0.13.4.dev374
## dependencies 
  - annexremote: 1.2.1
  - appdirs: 1.4.4
  - boto: 2.49.0
  - cmd:7z: 19.00
  - cmd:annex: 8.20201008-g3be4731
  - cmd:bundled-git: UNKNOWN
  - cmd:git: 2.29.1.windows.1
  - cmd:system-git: 2.29.1.windows.1
  - cmd:system-ssh: 7.1p2
  - exifread: 2.1.2
  - humanize: 3.1.0
  - iso8601: 0.1.13
  - keyring: 21.4.0
  - keyrings.alt: 3.1
  - msgpack: 1.0.0
  - mutagen: 1.41.1
  - requests: 2.24.0
  - tqdm: 4.51.0
  - wrapt: 1.12.1
## environment 
[...]

PR #5122, opened 2020-11-03
TST: Disable tests involving the /// dataset on windows

mih:
It contains a directory `con/` which prevents cloning on windows:
https://github.com/datalad/datalad/issues/5097


yarikoptic:
I will be renaming `con/` into `con_/` today to make windows happy.  Also cloning of any subdataset, e.g. `///repronim` should not be effected afaik by `con` at the same level

yarikoptic:
related now: https://github.com/datalad/datalad/issues/5123

yarikoptic:
/// is installable (if typed as `////` if under msys2 bash;)) on windows

Issue #5107, opened 2020-10-29
Tracking branch logic makes platform homogeneity assumptions


mih:
Use case: Register subdatasets into a dataset in adjusted unlocked mode. If the source datasets is also in adjusted mode, A record like this is created:

```
[submodule "sub1"]
        path = sub1
        url = localhost:/media/mih/Samsung_T5/tmp/orig_sub
        datalad-id = 8d8cdc78-ab1b-43fd-ad21-b33cb6b77def
```

If the source dataset is not in unlocked mode, the record looks like this:

```
[submodule "sub2"]
        path = sub2
        url = ./sub2
        datalad-id = 6763e4b4-3271-4cac-888b-32768da39e7e
```

In both cases the source URL for the clone was `localhost:/somepath/orig_sub`. In the second case that URL does not make it into the submodule record.

This is related to all kinds of issue, proposed, stale and reverted PRs, but nothing really fits, and this issue is central to the failures on windows, hence opening yet another issue.

PR #5125, opened 2020-11-04
ENH: appveyor -- use newer git-annex + try to re-enable codecov submission

yarikoptic:
TODOs:

- [x] Switch to win10-like test environment
- [x] Switch to PY37. @mih thinks there is no point in supporting outdated environment at this point. However PY38 is not fully functional on the current appveyor build image. So PY37
- [x] split into multiple test runs
- [x] also execute known2fail test modules in an allowed-failure test run, so we can detect improvements more easily and move tests into proper test runs accordingly
- [x] fixup testing
- [x] remove TEMP removing all other CIs
- [x] figure out or giveup (again) on codecov submission (@mih: I removed the blind re-enabling again, for now, did not work out)

PR #5127, opened 2020-11-05
Avoid repo type assumptions to enable proper tracking branch detection (maint)

mih:
Force-treating a to-be-added submodule as a GitRepo prevented the
detection of a corresponding branch to the active branch, in order to
determine the true tracking branch. Which is necessary to discover
the effective clone URL from a repository (when it is in adjusted
mode).

This is an alternative to 901c8fd4eaf8070fa632f03d9428eceba3c0c97c but
reworked to be applicable to the 0.13 series.

Fixes gh-5107

### DIFF
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -100,6 +100,7 @@ from .repo import (
     RepoInterface,
     path_based_str_repr,
 )
+from datalad.core.local.repo import repo_from_path
 
 # shortcuts
 _curdirsep = curdir + sep
@@ -2837,7 +2838,7 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
             cmd += ['-b', branch]
         if url is None:
             # repo must already exist locally
-            subm = GitRepo(op.join(self.path, path), create=False, init=False)
+            subm = repo_from_path(op.join(self.path, path))
             # check that it has a commit, and refuse
             # to operate on it otherwise, or we would get a bastard
             # submodule that cripples git operations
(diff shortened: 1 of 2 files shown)


datalad/support/gitrepo.py
```diff
@@ -2837,7 +2838,7 @@ def add_submodule(self, path, name=None, url=None, branch=None):
             cmd += ['-b', branch]
         if url is None:
             # repo must already exist locally
-            subm = GitRepo(op.join(self.path, path), create=False, init=False)
+            subm = repo_from_path(op.join(self.path, path))
```
kyleam, line 2841: Stepping through this with the added tests and `eval_under_testloopfs`, this looks good to me.

datalad/support/gitrepo.py
```diff
@@ -3115,7 +3116,7 @@ def get_tracking_branch(self, branch=None, remote_only=False):
             (remote or None, refspec or None) of the tracking branch
         """
         if branch is None:
-            branch = self.get_active_branch()
+            branch = self.get_corresponding_branch() or self.get_active_branch()
```
kyleam, line 3119: Hmm, this change isn't needed for the test case, as `AnnexRepo.get_tracking_branch(..., branch=None, corresponding=True)` already handles it.  And if GitRepo is being used directly, `get_corresponding_branch` always returns None.

Is the idea with this change just that if `GitRepo.get_corresponding_branch` ever changed to do anything meaningful, this is one (of many) spots that should take it into account?

mih: Ah, this might be an artifact of the order in which I explored the situation. I will investigate and adjust accordingly. Thx!

mih:
Timeouts on appveyor again, but #5125 was merged and things should get better.

datalad/support/gitrepo.py
mih: Remove, thx for the catch!

PR #5139, opened 2020-11-10
Add "miniconda" install scheme, --batch and --path-miniconda options

jwodder:
Part of #5078.

### DIFF
--- a/datalad/install.py
+++ b/datalad/install.py
@@ -1,8 +1,6 @@
 #!/usr/bin/env python3
 __python_requires__ = "~= 3.6"
 import argparse
-from contextlib import contextmanager
-from glob import glob
 import logging
 import os
 import os.path
@@ -41,8 +39,12 @@ def main():
     schemata.add_parser("autobuild", help="Linux, macOS only")
     schemata.add_parser("brew", help="macOS only")
     scm_conda_forge = schemata.add_parser("conda-forge", help="Linux only")
+    scm_conda_forge.add_argument("-b", "--batch", action="store_true")
+    scm_conda_forge.add_argument("--path-miniconda")
     scm_conda_forge.add_argument("version", nargs="?")
     scm_conda_forge_last = schemata.add_parser("conda-forge-last", help="Linux only")
+    scm_conda_forge_last.add_argument("-b", "--batch", action="store_true")
+    scm_conda_forge_last.add_argument("--path-miniconda")
     scm_conda_forge_last.add_argument("version", nargs="?")
     schemata.add_parser("datalad-extensions-build", help="Linux, macOS only")
     scm_deb_url = schemata.add_parser("deb-url", help="Linux only")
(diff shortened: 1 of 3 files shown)


datalad/install.py
```diff
@@ -318,12 +312,37 @@ def _install_via_conda(self, version, tmpdir):
...
-                for fname in glob("../miniconda/bin/git-annex*"):
-                    os.symlink(fname, os.path.basename(fname))
+            annex_bin = Path(self.annex_bin)
+            annex_bin.mkdir(parents=True, exist_ok=True)
+            for p in Path(conda_bin).glob("git-annex*"):
+                (annex_bin / p.name).symlink_to(p.resolve())
         self.post_install()
 
+    def install_miniconda(self, miniconda_path, batch=False):
+        miniconda_script = "Miniconda3-latest-Linux-x86_64.sh"
```
yarikoptic, line 322: while at it, could you make it also at least for OSX for which there is also a https://repo.anaconda.com/miniconda/Miniconda3-latest-MacOSX-x86_64.sh which I hope behaves identically to the one for linux?

Ideally should have one for windows, but since it is a python script, it is somewhat of a chicken/egg issue so we could postpone (evil plan: I want later to try to build some executable version of this self-contained script)

datalad/install.py
```diff
@@ -318,12 +312,37 @@ def _install_via_conda(self, version, tmpdir):
...
+        log.info("Downloading and running miniconda installer")
+        with tempfile.TemporaryDirectory() as tmpdir:
+            script_path = os.path.join(tmpdir, miniconda_script)
+            subprocess.run(
+                [
+                    "wget",
```
yarikoptic, line 328: looking forward to glorious windows times etc, might be worth to redo this avoiding `wget` with pure python urllib.  Some linux boxes also might not have `wget`.

yarikoptic (changes requested):
seems to not work for me
```
$> python datalad/install.py miniconda -b --path-miniconda /tmp/mymini            
  File "datalad/install.py", line 407, in <module>
    main()
  File "datalad/install.py", line 67, in main
    installer = GitAnnexInstaller(
  File "datalad/install.py", line 106, in __init__
    self.env_write_file = Path(env_write_file)
  File "/usr/lib/python3.8/pathlib.py", line 1041, in __new__
    self = cls._from_parts(args, init=False)
  File "/usr/lib/python3.8/pathlib.py", line 682, in _from_parts
    drv, root, parts = self._parse_args(args)
  File "/usr/lib/python3.8/pathlib.py", line 666, in _parse_args
    a = os.fspath(a)
TypeError: expected str, bytes or os.PathLike object, not NoneType

```
might be a good time to initiate `datalad/tests/test_install.py` with some `@turtle` unittest which would do above.

datalad/install.py
jwodder: Done.

datalad/install.py
jwodder: Done.

jwodder:
@yarikoptic To be clear, you're suggesting a single test that does `subprocess.run([sys.executable, "datalad/install.py", "miniconda", ...], ...)` and then what?  Just checks that `/tmp/mymini` exists and is nonempty?  That it contains certain entries?

yarikoptic:
right -- that command + running e.g. `.../bin/conda create -n test -y ` and verify that `conda activate test` in the output?

jwodder:
@yarikoptic Test added.

yarikoptic:
FWIW, tested out on osx box via `python3 datalad/install.py miniconda --path-miniconda ~/miniconda3-20201113 -b` and it worked!

so let's consider this stage one, and proceed

Issue #3733, opened 2019-09-27
siblings configure crashes with non-informative message when unknown remote is specified


yarikoptic:
```shell
$> datalad siblings -s github --publish-depends githubobjects configure       
[ERROR  ] expected str, bytes or os.PathLike object, not NoneType [subprocess.py:_execute_child:1453] (TypeError) 

$> git remote
githubobjects
```
so apparently I do not have github remote yet

PR #5146, opened 2020-11-11
BF: Prevent a sibling URL from being added when there is none

mih:
Turns:

```
$> datalad siblings -s github --publish-depends githubobjects configure
[ERROR  ] expected str, bytes or os.PathLike object, not NoneType [subprocess.py:_execute_child:1453] (TypeError)
```

into

```
% datalad -f json siblings -s github --publish-depends githubobjects configure
{"action": "configure-sibling", "name": "github", "path": "/tmp/gintest", "refds": "/tmp/gintest", "status": "error", "type": "sibling"}
```

However, the default result renderer does not care about such error
results and produces no output. But at least it doesn't crash anymore.

With the default `datalad.log.result-level=info` there is at least an
INFO log message with the error.

However, in general I'd say we discontinue the custom result renderer of
`siblings()` -- it is cryptic without providing a benefit.

Fixes gh-3733

### DIFF
--- a/datalad/distribution/siblings.py
+++ b/datalad/distribution/siblings.py
@@ -421,7 +421,7 @@ def _configure_remote(
     if name != 'here':
         # do all configure steps that are not meaningful for the 'here' sibling
         # AKA the local repo
-        if name not in known_remotes:
+        if name not in known_remotes and url:
             # this remote is fresh: make it known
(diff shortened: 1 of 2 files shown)


kyleam (approved):

mih:
Thx @kyleam !

Issue #5003, opened 2020-10-10
Fully remove (Git)Runner


mih:
This tracks the remaining issues before this can happen:

- [x] #4996 
- [x] #5002 
- [x] `customremotes` code depends on them #5142 
- [x] `datalad/support/archive_utils_patool.py` injects a Runner-based execution helper into `patool` #5141 
- [x] https://github.com/datalad/datalad-container/pull/141
- [ ] `datalad/tests/test_cmd.py` can be removed in full, when the classes are gone
- [ ] `datalad/tests/test_protocols.py` 
- [ ] large parts of `datalad.cmd` contain the runner implementation and various helpers that can all be deleted
- [ ] "protocols" will no longer be a thing and can be stripped from
    -  `datalad/customremotes/tests/__init__.py`
    -  `datalad/customremotes/base.py` 
    -  `datalad/customremotes/tests/test_archives.py`
    -  `datalad.interface/common_cfg.py`
    -  `datalad/support/gitrepo.py`
    -  `datalad/support/{protocol.py,tests/test_protocol.py}`, (delete in full)
    -  `datalad/support/tests/test_gitrepo.py`
- [ ] `GitRepo._run_git_custom_command` must be removed
- [x] Fix up command execution helpers in AnnexRepo (this could be a bottomless pit), in particular `_run_annex_command_json()` relies on Runner logging features. #5163

yarikoptic:
FTR I do use protocoling of annex calls for troubleshooting datalad/customremotes/ from time to time. Really handy

mih:
> FTR I do use protocoling of annex interactions for troubleshooting datalad/customremotes/ from time to time. Really handy

Please let me know, if that has any implications on the removal of `Runner`, which implies that any support for that functionality vanishes. I do not plan to implement support for that in `WitlessRunner` and I did not consider adding such a feature selectively to the `AnnexRemote`-based reimplementation of the datalad and archives special remote.

PR #5152, opened 2020-11-12
Trying to remove (Git)Runner

mih:
This is more or less just to document the state of things. Lots TODO. Record is kept at #5003


mih:
No need to keep this open.

PR #5154, opened 2020-11-12
Added copy_file() function that detects support for --reflink

### LINKED ISSUE #2953: wishlist: "smart" copyfile which takes advantage of CoW filesystems
Initially instigated/discussed in https://github.com/datalad/datalad-container/pull/55#discussion_r228609020  , in some situations we might like to have our "own" `copyfile` which performs similarly to git-annex internal copy using `--reflink=auto` so the `copy` takes advantage of possibly supported by filesystem CoW.

in current codebase we seems to not rely on it much
```shell
$> git grep -p copyfile            
datalad/cmd.py=def link_file_load(src, dst, dry_run=False):
datalad/cmd.py:        shutil.copyfile(src_realpath, dst)
datalad/metadata/aggregate.py=def _dump_extracted_metadata(agginto_ds, aggfrom_ds, db, to_save, force_extraction):
datalad/metadata/aggregate.py:            shutil.copyfile(
datalad/support/tests/test_annexrepo.py=from os.path import exists
datalad/support/tests/test_annexrepo.py:from shutil import copyfile
datalad/support/tests/test_annexrepo.py=def test_AnnexRepo_addurl_to_file_batched(sitepath, siteurl, dst):
datalad/support/tests/test_annexrepo.py:    copyfile(opj(sitepath, 'about.txt'), opj(dst, testfile))
datalad/support/tests/test_annexrepo.py:    copyfile(opj(sitepath, 'about2.txt'), opj(dst, testfile2))
datalad/support/tests/test_annexrepo.py:    copyfile(opj(sitepath, 'about2_.txt'), opj(dst, testfile2_))
```
but in the context of datalad-container might come real handy

jwodder:
Closes #2953.


datalad/utils.py
```diff
@@ -2590,3 +2591,41 @@ def check_symlink_capability(path, target):
...
+    if _cp_supports_reflink:
+        symopts = ["-R", "-L"] if follow_symlinks else ["-R", "-P"]
+        subprocess.run(
+            ["cp", "-f", "-p", "--reflink=auto"]
+            + symopts
+            + ["--", str(src), str(dst)],
+            check=True,
+        )
+    else:
+        return shutil.copy2(src, dst, follow_symlinks=follow_symlinks)
```
yarikoptic, line 2631: haven't looked inside shutil.copy2 etc, is there any way/hope to get some alignment in exceptions which would be raised if e.g. permissions denied or dst directory does not exist?  
I do not think we should "parse" the output from `cp` call but ideally there should be at least some common base exception or smth like that and if different (which is likely) between two possible runs -- docstring adjusted to document that.

jwodder: No.  If `cp` fails, all the caller gets is a `CalledProcessError` with the exit code (which isn't standardized) and the output.

yarikoptic: and indeed there are dedicated exceptions by copy2:
```
> python -c 'import shutil; shutil.copyfile(".git", "/tmp/gggg")'
  File "/usr/lib/python3.9/shutil.py", line 264, in copyfile
    with open(src, 'rb') as fsrc, open(dst, 'wb') as fdst:
IsADirectoryError: [Errno 21] Is a directory: '.git'

$> python -c 'import shutil; shutil.copy2(".git/index", "/")' 
  File "/usr/lib/python3.9/shutil.py", line 435, in copy2
    copyfile(src, dst, follow_symlinks=follow_symlinks)
  File "/usr/lib/python3.9/shutil.py", line 264, in copyfile
    with open(src, 'rb') as fsrc, open(dst, 'wb') as fdst:
PermissionError: [Errno 13] Permission denied: '/index'
```
I am ok to sacrifice them .... but maybe others not? @datalad/developers ?

yarikoptic (changes requested):
-R should not be applied by default

need feedback from others ... decided not to ask to use WitlessRunner for such a simplistic use case but may be it is worthwhile considering as well

datalad/utils.py
```diff
@@ -2590,3 +2591,41 @@ def check_symlink_capability(path, target):
...
+            ["cp", "--help"],
+            stdout=subprocess.PIPE,
+            stderr=subprocess.STDOUT,
+            universal_newlines=True,
+        )
+        # Ignore command failures (e.g., if cp doesn't support --help), as the
+        # command will still likely output its usage info.
+        _cp_supports_reflink = "--reflink" in r.stdout
+    if _cp_supports_reflink:
+        symopts = ["-R", "-L"] if follow_symlinks else ["-R", "-P"]
```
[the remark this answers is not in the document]
jwodder: `-R` removed.

yarikoptic (approved):
As it would be of great positive effect for me on BTRFS, I am in favor of this PR. Let's wait a bit more to hear possible objections from other @datalad/developers , and if none - I will merge tomorrow

datalad/utils.py
```diff
@@ -2590,3 +2591,41 @@ def check_symlink_capability(path, target):
...
+
+    Parameters
+    ----------
+    src: str or os.PathLike
+      path to the file to copy
+    dst: str or os.PathLike
+      path that `src` should be copied to
+    """
+    global _cp_supports_reflink
+    if _cp_supports_reflink is None:
```
mih, line 2612: Shouldn't this be made conditional to POSIX systems? Windows is already a slow target, and there is no chance that it could come back with a positive result.

datalad/utils.py
mih: We are not monkey-patching things, so I am OK with that. But maybe add a `Raises` section to the docstring that states what would happen. DataLad consumers might expect a `CommandError` (which is not coming), and Python consumers might expect the normal type of exceptions (which will not happen either).

mih (changes requested):
I left two comments. Sorry for the late review.

datalad/utils.py
yarikoptic: wouldn't this then become yet another OS-specific, vs scenario (tool, filesystem, etc) specific decisions we are trying to avoid?

yarikoptic: e.g. some https://en.wikipedia.org/wiki/ReFS from microsoft seems to aim to support CoW...

mih: I doubt that MS will expose CoW through win32 version of a POSIX command. With this change, the first call for each process will either determine that `cp` is not a command (on any standard windows system), or will use a command that is not from the OS. In both cases there is likely not benefit, and the cost is unconditional -- on windows.

yarikoptic: DataLad uses `git` and `git-annex`. `cp` is shipped with `git` on Windows, and AFAIK is used at least by `git-annex`, so I expect POSIX `cp` to be available to every DataLad installation including Windows.  As for doubts and hopes of CoW support -- we would not know until we try, and can only see if it is not supported today, but situation might change tomorrow. so I still do not see the reason for limiting besides making sure that this function works correctly with Windows style paths.

mih: `cp` is not in the PATH of cmd.exe when Git is installed with defaults via the installer and git-annex too. Both work nevertheless.

I am not sure, why `not on_windows` is problematic in this very spot.

datalad/utils.py
jwodder: Docstring updated.

adswa:
(argh apoplogies! - I found this old PR, wondered why it got stale, wanted to restart the tests on current ``master``, and accidentally introduced a line thats removal was committed in this PR (the import in ``test_annexrepo.py``). Really sorry, my bad :( )

mih:
I am closing this PR due to inactivity for 6+ months.

Issue #5155, opened 2020-11-13
ENH: config.patch context manager


yarikoptic:
CLI has `datalad -c var=val ` which makes `var=val` in effect only for the duration of that `datalad` call.
In Python API AFAIK we do not have such a convenience, and only `patch_config` helper for tests.

I think a good design pattern could be to just move that `patch_config` to become a context manager within `Config` itself, so we could do

```python
from datalad import cfg
with cfg.patch({'var': val, ...}):
    ...
```

That would also avoid "touching" protected `cfg._merged_store` within `datalad.tests.utils` thus keeping access to protected attributes within Config

WDYT @datalad/developers ?

mih:
Makes sense. It may be necessary to actually patch `.overrides` and trigger a reload going in an out of the context. Otherwise the patch will not survive a reload that happens in the context.

mih:
The is an (alternative) proposal that would likely make internal use of the feature proposed here

- https://github.com/datalad/datalad/issues/5587

yarikoptic:
FTR, we have `tests.utils.patch_config` https://github.com/datalad/datalad/blob/HEAD/datalad/tests/utils.py#L2071 which pretty much provides this functionality but better be not used outside of tests.

mih:
https://github.com/datalad/datalad-next/pull/760 now has a context manager that can do this. It posts overrides to the Git-aware ENV, hence affects subprocesses also. Looks like this:

```py
    with manager.overrides(cfg_overrides):
        ...
```

where `cfg_overrides` is a `dict` with overrides and `manager` is the incoming config maanager based on https://github.com/datalad/datasalad/pull/47

PR #5156, opened 2020-11-13
ENH: datalad.locations.extra-procedures to provide extra location(s) for procedures

yarikoptic:
This could be useful to point to extra locations for procedures which might
reside in some subdataset, or an alternative to system/user-wide locations.
E.g. it provides a solution IMHO to https://github.com/datalad/datalad/issues/4802 (alternative to #4803) which avoids need for some new semantic of --dataset in "create" command, thus keeping it possible to do

  datalad -c datalad.locations.extra-procedures=$PWD/procedures create  -d . -c mine sub/

as well as

  datalad -c datalad.locations.extra-procedures=$PWD/procedures create  -d ^ -c mine sub/

if current dataset is a subdataset within hierarchy.  Or point that location to any other arbitrary location (within subdataset, directory, anywhere).
Some inconvenience for use through Python API, for which I filed
https://github.com/datalad/datalad/issues/5155

TODOs
- [x] follow up commit since I believe a docstring needs further unrelated tune up

### DIFF
--- a/datalad/interface/common_cfg.py
+++ b/datalad/interface/common_cfg.py
@@ -86,6 +86,12 @@ definitions = {
         'destination': 'global',
         'default': opj(dirs.user_config_dir, 'procedures'),
     },
+    'datalad.locations.extra-procedures': {
+        'ui': ('question', {
+            'title': 'Extra procedure directory',
+            'text': 'Where should datalad search for some additional procedures?'}),
+        'destination': 'global',
+    },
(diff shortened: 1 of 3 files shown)


mih:
I have not had the time to investigate this re my original use case. It might still take a considerable time for me to come back to this.

yarikoptic:
Not sure if anyone else is interested in this one, but since IMHO it does at large addresses #4802, I will merge it in a day or two unless some objections are voiced.

PR #5048, opened 2020-10-16
BF: make AnnexRepo.add() raise an IncompleteResultsException if any non-success

yarikoptic:
Initial awareness came from troubleshooting on Windows
https://github.com/datalad/datalad/issues/5015#issuecomment-709436065

Majority of .add (as opposed to .add_ generator) invocations are not
testing/using returned value and GitRepo.add would return only
success-ful results and raise some Exception otherwise.

Since AnnexRepo.add relies on add_ which is going through --json
mode of git-annex to provide both succeeded and not results,
add() just returned records with success: False and did not
raise any exception (even if GitAnnex exited with non-0).  That is not
"in-line" with GitRepo.add and also leads to us not detecting errors
when plain AnnexRepo.add  is used.

I have not figured out a better/less confusing way than to add additional kwarg
expect_success to _run_annex_command_json .  Generator versions then would set
it to False, while outside "procedural" ones, where not "manually" processing
results (e.g. like .copy_to does), should set it to True for the
"centralized" handling.

I have subsequent similar changes (also introducing matching `_` generator forms) to some other AnnexRepo interfaces (eg `get`, `drop`, ...) which IMHO should be done to unify behavior and avoid "errors disregard".  Although some of them should probably go against `maint`, not yet sure if should, or should go against `master`.  First I want to see if any side-effects from `add` "fixup".

<details>
<summary>current version of that further patch</summary> 

```diff
diff --git a/datalad/distribution/drop.py b/datalad/distribution/drop.py
index 64a7292c1..03adec72d 100644
--- a/datalad/distribution/drop.py
+++ b/datalad/distribution/drop.py
@@ -94,7 +94,7 @@ def _drop_files(ds, paths, check, noannex_iserror=False, **kwargs):
 
     opts = ['--force'] if not check else []
     respath_by_status = {}
-    for res in ds.repo.drop(paths, options=opts):
+    for res in ds.repo.drop_(paths, options=opts):
         res = annexjson2result(
             # annex reports are always about files
             res, ds, type='file', **kwargs)
diff --git a/datalad/distribution/get.py b/datalad/distribution/get.py
index 925b61919..60542cbb8 100644
--- a/datalad/distribution/get.py
+++ b/datalad/distribution/get.py
@@ -584,7 +584,9 @@ def _get_targetpaths(ds, content, refds_path, source, jobs):
     for res in ds_repo.get(
             content,
             options=['--from=%s' % source] if source else [],
-            jobs=jobs):
+            jobs=jobs,
[...]

Issue #3415, opened 2019-05-15
Unified specification of `--jobs`


mih:
This is closely related to https://github.com/datalad/datalad/issues/3246, but now also relevant in the `publish` rewrite https://github.com/datalad/datalad/projects/9

I don't have a good idea on how to approach this, but in both cases we complicate the high-level API only to be able to not act on this information at all, and merely pass it down to the low-level handlers.

bpoldrack:
Should be a datalad option not a command option from my point of view. And yes, this would require a unified approach on how exactly to act upon it. However, moving this option accordingly and just let the command then pass it onto low-level would still be an improvement and a step in the right direction from my point of view.

mih:
I think the pattern to implement support for individual commands is identified and is being applied on a case by case basis. The need to do so has been reduced with the introduction of the `foreach-dataset` command. Future work needs to explore in how far this paradigm (across dataset parallelization with `foreach-dataset` and within-dataset parallelization via individual `--jobs` parameters per command) is applicable and capable to a large-enough range of use cases. Depending on the results of such an exploration, concerted adjustment of the API needs to be planned.

For now, this issue cannot contribute to this effort any further.

PR #5163, opened 2020-11-16
NF: AnnexRepo.call_annex*()

### LINKED ISSUE #4406: Investigate call_git*(["annex", ...]) use
It occurred to me while working on gh-4405 that any place where we call git-annex via one of the `call_git*` methods, rather than going through `_run_annex_command`, is likely problematic because that won't set the options (inlcuding `annex.alwayscommit`) that `_run_annex_command` takes care of setting.  Quickly grepping, there are a few spots.

### LINKED ISSUE #5161: Refactor `run_annex*` API in AnnexRepo
In the context of #5003 and specifically #5152 the git-annex execution helper in `AnnexRepo` are probably the last major connection of old `Runner` functionality with the rest of the code base. There are

- `_run_simple_annex_command()`
- `_run_annex_command_json()`

that both sit on top of `_run_annex_command()`. The latter already has some support with witless-execution, but in particular `_run_annex_command_json` does not use it, and implements a different progress handling than `AnnexJsonProtocol`.

There are numerous ways in which these three methods are called. It is unclear to me what an RF strategy would be or should be. However, from a design perspective it would be nice to have `call_annex()` to parallel `call_git()`.

It would also make sense to me to have `call_annex_items_()` to do the work of reporting JSON records. I do not see why the fact that JSON is used to receive information from git-annex should be reflected in the API.

So roughly speaking, I could see the following analogs:

- `GitRepo._call_git`: `AnnexRepo._run_annex_command` -> `AnnexRepo._call_annex`
- `GitRepo.call_git`: `AnnexRepo._run_simple_annex_command` -> `AnnexRepo.call_annex`
- `GitRepo.call_git_items_`: `AnnexRepo._run_annex_command_json` -> `AnnexRepo.call_annex_items_`

Here is a comprehensive list of `_run_annex_command()` usage in the code base of datalad proper. First the two main wrappers:

```
...(*args, **kwargs) [by _run_simple_annex_command]
...(command, files=files,  annex_options=annex_options, **kwargs)
```

where `args` and `kwargs` are exposed, hence could be anything, but contain legacy `log_*` options in any call via `_run_annex_command_json()`.

Remaining usage:

```
...('addurl', annex_options=options + urls, backend=backend, log_online=True,log_stderr=False, cwd=cwd)
...('adjust', annex_options=options, runner="gitwitless")
...('checkpresentkey',annex_options=list(annex_input),expect_fail=True,runner="gitwitless")
...('contentlocation',annex_options=[key],expect_fail=True,runner="gitwitless")
...('describe', annex_options=[name, description])
...('enableremote',annex_options=[name] + ensure_list(options),runner="gitwitless",env=env)
...('find', files=[f],annex_options=["--print0"], expect_fail=True,merge_annex_branches=False,runner="gitwitless")
...('find', annex_options=args, merge_annex_branches=False,  runner="gitwitless")
...('init', runner="gitwitless", protocol=AnnexInitOutput, annex_options=opts)
...('initremote', annex_options=[as_common_datasrc, 'type=git', 'location={}'.format(url), 'autoenable=true'])
...('initremote', annex_options=[name] + options, runner="gitwitless")
...('lookupkey', files=[files], expect_fail=True,runner="gitwitless")
...('metadata', '.', '-g', 'lastchanged')
...('migrate',annex_options=files,backend=backend,runner="gitwitless")
...('registerurl', annex_options=[dest_key, url])
...('reinject', annex_options=[str(tmploc), str_dest])
...('rmurl', files=[file_, url], runner="gitwitless")
...('setpresentkey', annex_options=[dest_key, dest_rid, '1'])
...('sync', annex_options=args, runner="gitwitless")
...('unannex', annex_options=options, files=files, runner="gitwitless")
...('untrust', annex_options=['here'])
```

and in the tests:
```
...('dead', annex_options=['here'])
...('direct' if enable_direct_mode else 'indirect', expect_stderr=True, runner="gitwitless")
`..("initremote", annex_options=["special", "type=git", "autoenable=true", "location=" + clone_url])
...('log')
...('unannex', annex_options=[op.join(lvl2relpath, 'file')])
...('upgrade')
```

Now same for `_run_simple_annex_command()`:

```
...("group", annex_options=[remote or '.'])
...("groupwanted", annex_options=[remote or '.'])
...("required", annex_options=[remote or '.'])
...("wanted", annex_options=[remote or '.'])
```

and for `_run_annex_command_json()`:

```
...('add',opts=options,files=files,backend=backend,expect_fail=True,jobs=jobs,expected_entries=expected_additions,expect_stderr=True)
...('add',opts=options,files=list(files.keys()),backend=None,expect_fail=True,jobs=None,expected_entries=expected_additions,expect_stderr=True)
...('addurl',opts=options + [files_opt] + [url],progress=True,expected_entries={file_: None},**kwargs)
...('copy',opts=annex_options,files=files,jobs=jobs,expected_entries=expected_copys,progress=True)
...('drop',opts=options + [k],jobs=jobs)
...('drop',opts=options,files=files,jobs=jobs)
...('dropkey', opts=options, files=keys, expect_stderr=True)
...('find', opts=expr, files=files, merge_annex_branches=merge_annex_branches)
...('fromkey',opts=[dest_key, str_dest, '--force'], expect_fail=True,)
...('fsck',iles=paths,git_options=git_options,opts=args,expect_stderr=True,expect_fail=True,)
...('get',jobs=jobs,expected_entries=expected_downloads,progress=True,**kwargs)
...('info', opts=options, files=files, merge_annex_branches=False)
...('info', opts=options, merge_annex_branches=merge_annex_branches)
...('ls-files', opts=opts, files=files)
...('ls-tree', opts=opts, files=files)
...('metadata', opts=['--json'], files=files)
...('metadata',args, files=files)
...("unlock", files=files)
...('whereis', **kwargs)
```

and in tests:

```
...('add', files='gl\\orious BS')
...('add', ['non-existing'], expect_fail=True)
...('add', ['non-existing'], expect_fail=False)
...("info",log_stdout=log_stdout,log_stderr=log_stderr,runner="gitwitless")
```

### LINKED ISSUE #5167: AnnexRepo._run_annex_command() does not handle `jobs=auto`
But `AnnexRepo._run_annex_command_json()` does. I do not see why this should be limited to JSON-based communication.

### LINKED ISSUE #5169: AnnexRepo.whereis() does not always pass on options to git-annex
https://github.com/datalad/datalad/blob/58abf0db07ea6119fd0c520f2445a758ca1b1bf1/datalad/support/annexrepo.py#L2305-L2311

If I read this correctly, it is only happening in "key-mode".

Documentation says:

```
        options: list, optional
            Options to pass into git-annex call
```

hence providing no hint that this is intended.

### LINKED ISSUE #5172: Remove AnnexRepo._annex_custom_command()
This function is in addition to:

- `_run_annex_command()`
- `_run_simple_annex_command()`
- `_run_annex_command_json()`

which a planned to be removed: #5161 

It is basically a wrapper around `Runner.run()` (non-witless), plus fake date handling.

I don't think it has any purpose anymore, and should be removed.

mih:
Analog to GitRepo.call_git*() and a requirement for #5152 

Issues touched upon:

- Fixes #4406 
- Further towards #5003
- f3d28e8 changes error behavior in line with the proposal in #5048
- Fixes #5161 
- Fixes #5167
- Fixed #5169
- Fixes #5172

TODO

- [x] check that tests pass, mostly adjust for new behavior to raise whenever a command fails, and not hide errors in results #5048 
- [x] adjust `drop` command (and maybe helpers) to anticipate command error when calling `AnnexRepo.drop()`. ATM @mih thinks that calling `AnnexRepo._call_annex_records(['drop'])` is cleaner than going through the various hoops that `AnnexRepo.drop()` implies.
- [x] crosscheck with #4406
- [x] complete docstrings
- [x] investigate removal of `_annex_custom_command()` -> https://github.com/datalad/datalad-crawler/issues/84
- [x] remove `_annex_custom_command()`
- [x] re the removal of `_run_*annex_command*()` I can only see a single use outside -core:
    ```
    ./ukbiobank/datalad_ukbiobank/update.py:            for rec in repo._run_annex_command_json('drop', **drop_kwargs):
    ```
   for @mih it is OK to rip out immediately (i.e. keep the respective commits in this PR, and fix ukbiobank in parallel)
- [x] discuss a public partner API for `_call_annex_records()`
- [x] return mechanism for records delivered before command failure. Right now they end up in `CommandError.kwargs['stdout_json']`
- [x]  discuss which post-error recovery/annotation functionality to shift from `_call_annex_records()` into `_call_annex()`
- [x] should we anticipate a future generator capability of the `Runner` and make `call_annex_records()` a `call_annex_records_()` right away?


kyleam:
There are failures in the `DATALAD_LOG_LEVEL=2` build.  It looks like all of the failures might boil down to
`rm_url`s use of `'--', file_, url`.  With the log level <= 8, `--debug` is appended, which git-annex complains about because it's after `-- file url`.  (To my surprise, git-annex seems fine with `-c k=v` values there).

https://travis-ci.com/github/datalad/datalad/jobs/441808237

mih:
Yes, saw that. Thx! I wanted to be clever, but wasn't. Will revert to the prev. setup.

kyleam:
This series looks really nice.

I'm pretty wary about changing the behavior of things like `AnnexRepo.drop` and `AnnexRepo.fsck` without adding a compatibility kludge and warning about the upcoming change.  Going from the first behavior below to second is a big change:

```console
$ python -c "from datalad.support.annexrepo import AnnexRepo; ar = AnnexRepo('.'); from pprint import pprint; pprint(ar.fsck([]))" 
[{'command': 'fsck',
  'dead': [],
  'error-messages': ['  Only 1 of 3 trustworthy copies exist of one',
                     '  Back it up with git-annex copy.'],
  'file': 'one',
  'input': ['one'],
  'key': 'SHA256E-s4--2c8b08da5ce60398e1f19af0e5dccc744df274b826abe585eaba68c525434806',
  'note': 'checksum...',
  'success': False,
  'untrusted': []}]

```

```console
$ python -c "from datalad.support.annexrepo import AnnexRepo; ar = AnnexRepo('.'); from pprint import pprint; pprint(ar.fsck([]))" 
  File "/home/kyle/src/python/datalad/datalad/support/annexrepo.py", line 2708, in fsck
    git_options=git_options,
  File "/home/kyle/src/python/datalad/datalad/support/annexrepo.py", line 1073, in _call_annex_records
    raise e
  File "/home/kyle/src/python/datalad/datalad/support/annexrepo.py", line 999, in _call_annex_records
    **kwargs,
  File "/home/kyle/src/python/datalad/datalad/support/annexrepo.py", line 971, in _call_annex
    **kwargs)
  File "/home/kyle/src/python/datalad/datalad/cmd.py", line 518, in run
    **results,
datalad.support.exceptions.CommandError: CommandError: 'git annex fsck --json --json-error-messages -c annex.dotfiles=true -c annex.retry=3' failed with exitcode 1 under /tmp/po-OUCQPk6 [err: 'git-annex: fsck: 1 failed'] [info keys: stdout_json]
```

> discuss a public partner API for `_call_annex_records()`

Based on the added callers aside from `push` outside of annexrepo.py, perhaps we could get be with `cmd` and `files`, adding as needed later?

mih:
> I'm pretty wary about changing the behavior of things like `AnnexRepo.drop` and `AnnexRepo.fsck` without adding a compatibility kludge and warning about the upcoming change. 

Yes, I am uncertain about the best approach here, too. But if we take #5048 seriously and exercise it in full, all these methods must raise an exception. Few things that I know about this so far:

- the way that results up to a failure are provided is suboptimal (`CommandError.kwargs['stdout_json']`)
- we should adjust the code base to the new behavior in full, such that any compatibility kludges are not exercised by us

I have not settled on a good place to put such a kludge. I think most error handling that is still scattered around should move into `_call_annex()`, and it (or `_call_annex_records()`) should get a private switch that catches the `CommandError` to maintain previous behavior. We turn it on for 0.14, switch the default to off in 0.15, and remove the kludge in 0.16.

> > discuss a public partner API for `_call_annex_records()`
> 
> Based on the added callers aside from `push` outside of annexrepo.py, perhaps we could get be with `cmd` and `files`, adding as needed later?

Yes, that matches my POV. Thx for your thoughts!

kyleam:
>> I'm pretty wary about changing the behavior of things like `AnnexRepo.drop` and `AnnexRepo.fsck` without adding a compatibility kludge and warning about the upcoming change.
>
> Yes, I am uncertain about the best approach here, too. But if we take #5048 seriously and exercise it in full, all these methods must raise an exception.

If we were starting fresh, I would probably share that view, though I think always raising could be a bit awkard.  For example, it seems very likely that a caller of `whereis` is inspecting the result; making the caller go through an additional try-except dance seems unnecessary.  But perhaps it's worth it for consistency.

As for whether it's worth changing the behavior now, I don't know.

> Few things that I know about this so far:
>
> - the way that results up to a failure are provided is suboptimal (`CommandError.kwargs['stdout_json']`)

Hmm, I don't see a better alternative than attaching them to the exception, so the only improvement that comes ot mind is tweaking `CommandError`s constructor so that `stdout_json` is stored as an attribute.

> - we should adjust the code base to the new behavior in full, such that any compatibility kludges are not exercised by us
>
> I have not settled on a good place to put such a kludge.

Yeah, it's of course easy for me to say that we should, but how to go about it does seem tricky...

> I think most error handling that is still scattered around should move into `_call_annex()`, and it (or `_call_annex_records()`) should get a private switch that catches the `CommandError` to maintain previous behavior. We turn it on for 0.14, switch the default to off in 0.15, and remove the kludge in 0.16.

How's the switch flipped in this case?  Above you say core code shouldn't use it (and I agree, our code base shouldn't be emitting our own warnings), but if it's just a parameter on one of the `_call_annex*` methods, `fsck` would have to use the old code path in order retain its current behavior.

I haven't come up with any workable ideas that wouldn't be essentially adding a parameter to each top-level method that calls `_records` and phasing that through, perhaps all the way to a noop.  That'd be quite ugly, and require updates of all callers in our code base.

I think a real alternative is to say the consistency isn't worth it at this point (in terms of coming up with the approach or making third-party code update their scripts), unless its carried in separately with something like new core repo classes.

Anyway, that is of course all just my two cents.  Thanks for all of your work on this.

datalad/cmd.py
```diff
@@ -1184,6 +1184,72 @@ def _get_adjusted_env(self, env=None, cwd=None, copy=True):
             copy=False,
         )
 
+    def run_on_filelist_chunks(self, cmd, files, protocol=None,
+                               cwd=None, env=None, **kwargs):
```
kyleam, line 1188: Should we switch `_call_git` over to this method as well?  Then the only remaining use of `run_gitcommand_on_file_list_chunks` would be in the now unused `_git_custom_command`.

datalad/support/annexrepo.py
```diff
@@ -3018,8 +2901,8 @@ def _format_response(res):
                 files = [files]
             # anything else is assumed to be an iterable (e.g. a generator)
         if batch is False:
-            for res in self._run_annex_command_json(
-                    'metadata', opts=['--json'], files=files):
+            for res in self._call_annex_records(
+                    ['metadata', '--json'], files=files):
```
kyleam, line 2905: Doesn't functionally matter and is true for the old code too, but since it's being converted: this leads to git-annex being called with two `--json` flags.

mih: Done

datalad/cmd.py
mih: YES! Thx, done!

mih:
Trying to get back to this project. Pushed a rebase on present master and addressed two of @kyleam's more recent comments.

mih:
We discussed how the repo methods' error behavior should be (relative to the high-level API and considering established usage). However, we failed to document the consensus here or in the meeting notes. Does anyone remember?

kyleam:
I think the consensus was something along the lines of "most of these repo methods should raise exceptions and that they don't can be considered a bug; in the few methods where we think inspection of the output is the main goal (perhaps things like whereis and fsck [*]), specific methods can catch and retain the current behavior".

[*] I don't think there was agreement about whether these shouldn't raise.

kyleam:
Hopefully all of the remaining travis failures are addressed by the last push, and the -crawler failure should be covered by datalad/datalad-crawler#87.

mih:
I love you too!

mih:
Re potential `call_annex_records()` API: Cases that do not fit `(cmd, files=None)` are:

- `push`: needs custom protocol parameterization and `stdin`
- `get`: needs custom protocol parameterization
- `_get_expected_files()`: needs `merge_annex_branches` flag
- `add`: needs custom protocol parameterization and `jobs` parameter
- `addurl`: needs `progress` flag
- `drop`: needs `jobs`
- `info`: needs `merge_annex_branches`
- `fsck`: needs `git_options`
- `copy`: needs `jobs`, `progress`, and custom protocol parameterization

Given this assessment, I concur with @kyleam that going with `(cmd, files=None)` as a start is a good thing.

I would not touch `jobs` now, and until there is a plan for https://github.com/datalad/datalad/issues/3415

As a next step it could make sense to formalize the progress reportin request (`progress`) and the expected amount (`total_nbytes`) into a single parameter and start supporting that. But that is not cooked well enough right now IMHO.

mih:
@kyleam I rebase on current master to get coverage fixes.

mih:
Travis failure is due to the test run exceeding the max log output volume. But not as a consequence of a bug.

mih:
> should we anticipate a future generator capability of the Runner and make call_annex_records() a call_annex_records_() right away?

I'd say, there is no need to do that now and little cost to maintaining a non-generator variant, later and/or in-addition.

mih:
I believe this is ready.

kyleam (approved):
Great, thanks for the updates.  Reading through again (not as carefully), I only spotted two minor docstring/comments issues (push incoming).

---

> RF: Replace last _run_annex_command_json() call
> 
> Blind attempt, this is not tested on my machine, and seamingly untested
> in our CIs too.

Thanks.  This codepath is only triggered with fake dates.  (I should add a test case that patches `DATALAD_FAKE__DATES`.)  Anyway, your blind substitutions look good and pass on my machine under `DATALAD_FAKE__DATES=1`.

mih:
Alrighty. This took a long time, but I think the outcome is quite nice and pleasantly mirrors the `GitRepo` solution. Thanks for the feedback over the past months, and to @kyleam for his contributions specifically.

PR #5170, opened 2020-11-18
Make benchmark workflow fail if a benchmark slowed down too much

jwodder:
As requested [here](https://github.com/con/fscacher/pull/16#issuecomment-729761695).


yarikoptic (approved):
Great, thank you @jwodder 

1.2 is quite far from typical differences we see, so we might eventually like to lower it.  But if it does turn benchmarking red - would be really a bad sign

Issue #5181, opened 2020-11-20
install.py: implement more flexible/modular but also easy on user parsing schema


yarikoptic:
initial `install.py` (and shell prototype it was based on) was targeting 1 component (git-annex) and then miniconda was added as an individual component but overall parsing was not yet refactored, and now looks like

```
schema:
  Type of git-annex installation (default "conda-forge")

  {autobuild,brew,conda-forge,conda-forge-last,datalad-extensions-build,deb-url,neurodebian,neurodebian-devel,snapshot,miniconda}
```

It think we are at a good point to RF it along the lines proposed in https://github.com/datalad/datalad/issues/5078 (and somewhat inspired by neurodocker , check it out)

``` [--options-applied-to-all] COMPONENT1[=version] [--options] COMPONENT2[=version] [--options] ...```

or more specifically

``` [--options-applied-to-all] COMPONENT1[=version] [--options] [--schema X] [--schemaoptions] COMPONENT2[=version]...```

where current target components should be (adding `venv`, I think it is time -- could come handy for testing PRs etc)

- miniconda
- venv
- git-annex
- datalad

among `--options`, both datalad and git-annex should get `--schema` option (thus I listed it explicitly above), which would be "auto" by default, and otherwise allow for following "schemas":

- datalad
  - apt
  - conda
  - deb-url
  - neurodebian
  - neurodebian-devel
  - pip
- git-annex
  - apt
  - autobuild
  - brew
  - conda
  - ~~conda-last  # or let's drop it as not used ATM anyways and not to be sourced?~~
  - datalad/annex  # from datalad-extensions-build since we moved
  - deb-url
  - neurodebian
  - neurodebian-devel
  - snapshot

some of those (conda-forge, pip, apt) should use specific `[=VERSION]` if specified, the others should raise some exception.

we can decouple installation of git-annex from installation of miniconda, i.e. if it is `install.py git-annex --schema conda-forge` and prior `miniconda` component which would install miniconda and make `conda` available, and there is no `conda` in the PATH  -- it is ok to just exit with an exception if conda is not available.  Ideally it should be a dedicated common exception like `IncompatibleEnvrionmentError(ValueError) ` (similar exception for e.g trying neurodebian install on OSX etc). 

Installation of a component could lead to 
[...]

PR #5182, opened 2020-11-20
install.py: Add "datalad" schema for installing Miniconda with datalad

jwodder:
[As requested](https://github.com/datalad/datalad/issues/5078#issuecomment-730549628) in #5078


yarikoptic:
I think it is ok for what it is (already would be useful) but let's make it flexible (see #5181)

PR #5188, opened 2020-12-01
ENH: travis -- remove cron run with bad scrapy (of no concern here any longer)

yarikoptic:
Paid for 5 concurrent jobs plan, let's see if it works anyhow

PR #5190, opened 2020-12-01
Prevent cron-only Travis builds from starting on non-cron events

### LINKED ISSUE #5189: TRAVIS: make cron jobs not even start the VM
ATM we have
```yaml
before_install:
  - if [ ! -z "${_DL_CRON:-}" ]; then
      if [ ! ${TRAVIS_EVENT_TYPE} = "cron" ]; then
        echo "Exiting since should only be ran on a weekly basis by cron!";
        exit 0;
      fi
    fi
```
so it requires first for a machine to boot, waste of time.
Not yet sure exactly how but https://github.com/travis-ci/travis-ci/issues/7149#issuecomment-335825605 

```yaml
language: python
python:
  - "3.6"
matrix:
  include:
    - if: branch =~ ^deploy
      python: "2.7"
```
suggests that we could add conditional to those to not even be considered if some expression is not matched.

There is also "stages" feature https://blog.travis-ci.com/2017-09-12-build-stages-order-and-conditions so may be we could have some stages (tests and extra-tests) or alike, again where some are triggered only for cron jobs.  @jwodder please have a look either you see a clean way

jwodder:
Closes #5189.

Docs references:
* https://docs.travis-ci.com/user/conditional-builds-stages-jobs/
* https://docs.travis-ci.com/user/conditions-v1

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,7 +62,6 @@ matrix:
     # We cannot have empty -A selector, so the one which always will be fulfilled
     - NOSE_SELECTION=
     - NOSE_SELECTION_OP=not
-    - _DL_CRON=1
   - python: 3.5
     # Split runs for v6 since a single one is too long now
     env:
@@ -137,14 +136,12 @@ matrix:
     python: 3.5
     env:
     - _DL_ANNEX_INSTALL_SCENARIO="deb-url http://snapshot-neuro.debian.net/archive/neurodebian/20190710T050502Z/pool/main/g/git-annex/git-annex-standalone_7.20190708%2Bgit9-gfa3524b95-1~ndall%2B1_amd64.deb"
-    - _DL_CRON=1
   # Run with git's master branch rather the default one on the system.
   - if: type = cron
     python: 3.5
     env:
     - DATALAD_USE_DEFAULT_GIT=1
     - _DL_UPSTREAM_GIT=1
-    - _DL_CRON=1
   # Run with our reported minimum Git version.
   - if: type = cron
     python: 3.5
@@ -152,7 +149,6 @@ matrix:
     - DATALAD_USE_DEFAULT_GIT=1
     - _DL_MIN_GIT=1
     - PATH="$PWD/git-src/bin-wrappers/:$PATH"
-    - _DL_CRON=1
   - if: type = cron
     python: 3.5
     env:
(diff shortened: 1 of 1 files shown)


yarikoptic:
oh sweet!  thanks @jwodder !  verified that travis did run the other jobs (failed due to workaround was pending in #5188) please also 
- remove any mentioning/use of `_DL_CRON` then since it is of no use/purpose AFAIK with this PR changes
- rebase against `maint` -- I have just merged https://github.com/datalad/datalad/pull/5188 which should make travis green again

then when maint is green and ready, we will just merge it into master to propagate

jwodder:
@yarikoptic 

> remove any mentioning/use of _DL_CRON then since it is of no use/purpose AFAIK with this PR changes

Actually, it is still of use; the `after_success` step at the end of `.travis.yml` sends coverage off to Codecov only if the job is not a cron-only job, and `_DL_CRON` is needed to make that determination.

jwodder:
Rebased against maint.

yarikoptic:
> @yarikoptic
> 
> > remove any mentioning/use of _DL_CRON then since it is of no use/purpose AFAIK with this PR changes
> 
> Actually, it is still of use; the `after_success` step at the end of `.travis.yml` sends coverage off to Codecov only if the job is not a cron-only job, and `_DL_CRON` is needed to make that determination.

RF that coverage submission to be conditioned on `if [ ! ${TRAVIS_EVENT_TYPE} = "cron" ]; then` instead and remove `_DL_CRON`?

jwodder:
@yarikoptic That would be a behavior change: when all the jobs are run for a cron, none of them — not just the cron-only ones — would submit coverage.  Is that what you want?

yarikoptic:
> @yarikoptic That would be a behavior change: when all the jobs are run for a cron, none of them — not just the cron-only ones — would submit coverage. Is that what you want?

oh, you are right, I have missed that.  In general I think it should be ok if none of the cron-triggered jobs submits coverage report and we just rely on the reports submitted from regular runs on a (master or maint) branch.  I still think it is worth it since would simplify our convoluted .travis.yml even if tiny bit, and avoid duplicate specification needed for cron-only jobs.

jwodder:
@yarikoptic `_DL_CRON` removed.

kyleam:
Nice.  Thanks @jwodder.

PR #5117, opened 2020-11-02
Test _RunnerAdapter.put(recursive=True, preserve_attrs=True)

jwodder:
As instructed in #5108.

PR #5196, opened 2020-12-02
Windows: Experiment with quoting

adswa:
I gave up trying to understand how [(different things on) Windows interpret and escape quotes](https://stackoverflow.com/a/31413730), but  the error that surfaced in #5117, where the source and destination in a ``cp`` command got grouped into a single argument instead of two, is resolved by escaping a double quote with another double-quote (``""``) instead of two double-quotes (``"""``). Since Travis works again, I'm opening a PR just to see which things break with switching from triple double-quotes to double double-quotes.

### DIFF
--- a/datalad/core/distributed/tests/test_clone.py
+++ b/datalad/core/distributed/tests/test_clone.py
@@ -509,8 +509,10 @@ def test_installationpath_from_url():
 def test_expanduser(srcpath, destpath):
     src = Dataset(Path(srcpath) / 'src').create()
     dest = Dataset(Path(destpath) / 'dest').create()
-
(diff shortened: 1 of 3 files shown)


adswa:
I cherry-picked the fix to the failing test from #5199 and rebased to the current master (hoping that this will eradicate the travis failures)

adswa:
The quoting experiment wasn't successful, I introduced all kinds of other failures with it. Maybe @bpoldrack has a better idea how to handle this. I will close this PR, other relevant pieces have made it into other PRs and it at least led to a test-fixing spree for Appveyor :-)

PR #5199, opened 2020-12-02
BF/Win: Use windows own concept of HOME

adswa:
This should fix a test failure that surfaced in #5196 and in @mih's new appveyor setup builds.
On Windows, instead of using ``HOME``, its equivalent ``USERPROFILE`` should be patched. ([list of env vars under windows](https://docs.microsoft.com/en-us/windows/deployment/usmt/usmt-recognized-environment-variables)).
On my windows computer, this change makes the test pass.

PR #5201, opened 2020-12-03
TST/BF: enable overwriting downloads under Windows

adswa:
This PR aims to fix a test failure that surfaced when a re-download attempts to overwrite an existing file (https://github.com/datalad/datalad/pull/5199#issuecomment-737973537). On Windows systems, an ``os.rename()`` will always raise a FileExistsError
if the file exits, while on Unix systems the existing file is replaced silently if the user has permissions, see https://docs.python.org/3/library/os.html#os.rename.
``os.replace()`` appears to be a cross-platform compatible solution, but let's hope that this doesn't break something else - then I could just do it ``if on_windows``.

### DIFF
--- a/datalad/downloaders/base.py
+++ b/datalad/downloaders/base.py
@@ -475,7 +475,7 @@ class BaseDownloader(object, metaclass=ABCMeta):
                 os.utime(temp_filepath, (time.time(), status.mtime))
 
             # place successfully downloaded over the filepath
-            os.rename(temp_filepath, filepath)
+            os.replace(temp_filepath, filepath)
 
             if stats:
                 stats.downloaded += 1
--- a/datalad/downloaders/tests/test_http.py
+++ b/datalad/downloaders/tests/test_http.py
@@ -129,7 +129,6 @@ def test_process_www_authenticate():
                  [])
 
 
-@known_failure_githubci_win
 @with_tree(tree=[('file.dat', 'abc')])
 @serve_path_via_http
 def test_HTTPDownloader_basic(toppath, topurl):
@@ -808,7 +807,6 @@ def test_lorisadapter(d, keyring):
     assert_equal(content, "correct body")
 
 
-@known_failure_githubci_win
 @with_tree(tree=[('file.dat', 'abc')])
 @serve_path_via_http
 def test_download_url(toppath, topurl):


adswa:
mental note: remove the known_failure_githubciwin annotation of ``test_HTTPDownloader_Basic`` if this fixes the error.

adswa:
yay, the two tests in question (``test_HTTPDownloader_Basic`` and ``test_download_url``) now pass. I have removed the knownfailure annotation for the GH workflow and will see what happens

kyleam:
Thanks, @adswa.  I wasn't aware of `os.replace`.

> yay, the two tests in question (`test_HTTPDownloader_Basic` and `test_download_url`) now pass. I have removed the knownfailure annotation for the GH workflow and will see what happens

Hmm, do you mean by looking at the builds for this PR?  I think we haven't had a windows build for the github ci since gh-4303.

adswa:
> Hmm, do you mean by looking at the builds for this PR? I think we haven't had a windows build for the github ci since gh-4303.

oh, that's a misconception of mine then, sorry. I *thought* that this would be the CrippledFS workflow. Shall I drop that commit (4a09de1)?

kyleam:
> I _thought_ that this would be the CrippledFS workflow.

That's ubuntu with `$TMPDIR` pointing to a vfat fs (with the idea that that will reveal things like adjusted branch issues that aren't windows-specific).

> Shall I drop that commit ([4a09de1](https://github.com/datalad/datalad/commit/4a09de19d4bdb4e325b71c946d555ac81b2bb982))?

Given it [fixed][1] these test failures on your end, I'd say it's fine to keep.  If at some point a github ci windows build is re-enabled, we'll need assess where things stand anyway, so the worst case is that these two tests need to be marked up again.

[1]: https://github.com/datalad/datalad/pull/5199#issuecomment-737973537

kyleam:
Any objections to moving this to maint?

adswa:
> Any objections to moving this to maint?

Not at all. I have to admit, I too rarely think of this in advance. I can rebase shortly, will just finish a commit on another branch

kyleam:
@adswa I'm guessing the build failures are from a push followed by changing the base on github rather than the other way around.  Triggering a new build (with a noop rewrite or a rebase to the latest maint) would hopefully sort it out.  Sorry for the hassle.

adswa:
I have rebased to the latest maint

kyleam (approved):

PR #5203, opened 2020-12-03
TEMP: try most recent release of annex on conda-forge on maint

yarikoptic:

PR #5212, opened 2020-12-07
BFs: bug fixes for testing picked up from #5203

yarikoptic:
See individual commits for descriptions.  I decided to leave the `ls` in the helper script to avoid need to debug in the future (we never look at that output typically anyways)

### DIFF
--- a/datalad/core/distributed/tests/test_clone.py
+++ b/datalad/core/distributed/tests/test_clone.py
@@ -362,15 +362,23 @@ def check_reckless(annex, src_path, top_path, sharedpath):
         raise SkipTest("Remainder of test needs proper filesystem permissions")
 
     if annex:
-        # umask might be such (e.g. 002) that group write permissions are inherited, so
-        # for the next test we should check if that is the case on some sample file
-        dltmp_path = ds.pathobj / '.git' / "dltmp"
-        dltmp_path.write_text('')
-        default_grp_write_perms = dltmp_path.stat().st_mode & stat.S_IWGRP
-        dltmp_path.unlink()
-        # the standard setup keeps the annex locks following umask inheritance
-        eq_((ds.pathobj / '.git' / 'annex' / 'index.lck').stat().st_mode \
-             & stat.S_IWGRP, default_grp_write_perms)
+        if ds.repo.git_annex_version < "8.20200908":
+            # TODO: Drop when GIT_ANNEX_MIN_VERSION is at least 8.20200908.
+
+            # the standard setup keeps the annex locks accessible to the user only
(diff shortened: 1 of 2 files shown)


datalad/core/distributed/tests/test_clone.py
```diff
@@ -362,9 +362,15 @@ def check_reckless(annex, src_path, top_path, sharedpath):
         raise SkipTest("Remainder of test needs proper filesystem permissions")
 
     if annex:
-        # the standard setup keeps the annex locks accessible to the user only
-        nok_((ds.pathobj / '.git' / 'annex' / 'index.lck').stat().st_mode \
-             & stat.S_IWGRP)
+        # umask might be such (e.g. 002) that group write permissions are inherited, so
+        # for the next test we should check if that is the case on some sample file
```
kyleam, line 366: I think that git-annex does try to tweak these permissions via `setAnnexFilePerm` and that there may have been an unintended behavior change in a recent annex.  I'm trying to pinpoint that.

https://travis-ci.com/github/datalad/datalad/jobs/455429618#L2315

yarikoptic: Good point! I just blindly assumed that we were just somehow skipping this test with older annex, I should have checked, indeed ideally should be fixed in git annex

kyleam: The change in behavior comes with with 6e9a4f50f (make viaTmp honor umask, 2020-09-02).  Based on my reading of that commit, the related eed20fe3b (fix some file modes in calls to withTmpFileIn to honor umask, 2020-09-02), and the [related bug report][1], it's intentional.

I've pushed a commit that uses the previous check for older git-annex versions.  Of course feel free to drop it if you don't think that's the way to go.

I think `test_clone:test_reckless` should now be in a passing state for all combinations.  Local results:


| datalad version | git-annex version     | umask=022 | umask=002 |
|--------------- |--------------------- |--------- |--------- |
| master          | 7.20190819-g908476a9b | passes    | passes    |
| master          | 8.20201127            | passes    | fails     |
| pr (a8d269aa01) | 7.20190819-g908476a9b | passes    | fails     |
| pr (a8d269aa01) | 8.20201127            | passes    | passes    |
| pr (current)    | 7.20190819-g908476a9b | passes    | passes    |
| pr (current)    | 8.20201127            | passes    | passes    |

[1]: https://git-annex.branchable.com/bugs/directory_special_remote_export_file_mode/

kyleam (approved):
Looks good to me.

Thanks for figuring out why we couldn't trigger the `test_reckless` failure locally.

yarikoptic:
Thank you @kyleam for figuring out annex change and fixing up this PR, appreciated!  Let's proceed!

Issue #5211, opened 2020-12-07
Windows: create-sibling is dysfunctional


adswa:
When taking a closer look at the test failure of ``datalad.distribution.tests.test_publish.test_since_empty_and_unsupported`` in #5210 I noticed that there are ``mkdir -p`` calls in the code base of "create sibling". Windows, unfortunately, has [its own mkdir command](https://docs.microsoft.com/de-de/windows-server/administration/windows-commands/mkdir) that does not take any flags. Consequently, when calling ``create sibling`` in a dataset, the command fails and leaves behind an empty ``-p/`` directory:

```
(base) C:\Users\mih\adinatests>datalad create blabla
[INFO] Creating a new annex repo at C:\Users\mih\adinatests\blabla
[INFO] Detected a filesystem without fifo support.
[INFO] Disabling ssh connection caching.
[INFO] Detected a crippled filesystem.
[INFO] Scanning for unlocked files (this may take some time)
[INFO] Entering an adjusted branch where files are unlocked as this filesystem does not support locked files.
[INFO] Switched to branch 'adjusted/master(unlocked)'
create(ok): C:\Users\mih\adinatests\blabla (dataset)

(base) C:\Users\mih\adinatests>cd blabla

(base) C:\Users\mih\adinatests\blabla>datalad create-sibling ../onebla
[INFO] Considering to create a target dataset C:\Users\mih\adinatests\blabla at C:\Users\mih\adinatests\onebla of localhost
[INFO] Fetching updates for Dataset("C:\Users\mih\adinatests\blabla")
[INFO] Adjusting remote git configuration
[INFO] Running post-update hooks in all created siblings
create_sibling(error): C:\Users\mih\adinatests\blabla (dataset) [failed to run post-update hook under remote path C:\Users\mih\adinatests\onebla (CommandError: 'cd "C:\Users\mih\adinatests\onebla\.git" && ( [ -x hooks/post-update ] && hooks/post-update || : )' failed with exitcode 1 [err: 'Der Befehl ":" ist entweder falsch geschrieben oder
konnte nicht gefunden werden.'] [cmd.py:run:514])]

(base) C:\Users\mih\adinatests\blabla>ls
-p
```

adswa:
oh there is so much on windows that makes "create sibling" doomed to fail :(

adswa:
There are several things that do not work on windows. One is this part of the code:
[...]

PR #5214, opened 2020-12-07
BF/Win: Towards a working creating sibling command

adswa:
There are a range of issues with create sibling (see #5211) and I'm not sure if it is possible possible to make it work, but I at least wanted to start.

- [x] use ``mkdir`` without -p on Windows
- [x] prevent failure due to not recognizing ``|| :`` as a command

### DIFF
--- a/datalad/distribution/create_sibling.py
+++ b/datalad/distribution/create_sibling.py
@@ -795,7 +795,7 @@ class CreateSibling(Interface):
             lgr.debug("Running hook for %s (if exists and executable)", path)
             try:
                 shell("cd {} "
-                      "&& ( [ -x hooks/post-update ] && hooks/post-update || : )"
+                      "&& ( [ -x hooks/post-update ] && hooks/post-update || true )"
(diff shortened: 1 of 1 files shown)


yarikoptic:
Looks lgtm to me. I already forgot - nobody should care about windows in maint? If we still do, rebase is desired

adswa:
> nobody should care about windows in maint? If we still do, rebase is desired

ah, damn, I keep forgetting to think about which base to use :-) I will rebase

kyleam:
@adswa I believe the Travis builds failures are because you rebased/pushed then changed the base on the web interface, rather than the other way around.  A force-push of a noop change should sort it out.

adswa:
> @adswa I believe the Travis builds failures are because you rebased/pushed then changed the base on the web interface, rather than the other way around. A force-push of a noop change should sort it out.

I swear at one point I manage to do it in the correct order. Thx! Pushed an empty commit.

kyleam (approved):

PR #5216, opened 2020-12-08
TEMP: try most recent release of annex on conda-forge on master

yarikoptic:
Follow up to #5203 (on maint)


kyleam:
Green here too.  Nice :)

yarikoptic:
Nice indeed! ;)

PR #5218, opened 2020-12-08
RF+BF: github - remove user/password authentication, use only tokens

yarikoptic:
- minor __doc__ tuneup for typos, and consistent GitHub (not Github)
- also removes storing of a (now "entered", not generated) token into
  git configuration.  If desired to do so - do manually.  To
  centralize entered credentials handling, regular credential store is
  used
- VCR tapes were updated (with token manually replaced after) for two
  tests and 3rd commented out was removed completely (Closes #4079)
- in some cases a little better error reporting now (e.g., giving a
  hint that token likely lack permissions to find a repository while
  creating ;))

Overall, it is not yet "much better" but it should be working and
usable again.

If merged, closes #3947

PR #5219, opened 2020-12-08
ENH(UX): do not repeat entry for token and aws key/secret

yarikoptic:
I failed to find an issue but I know that it annoyed many people, me included,
whenever long secrets such as tokens and keys for AWS were asked to be repeated.
There is IMHO no reason for that since they are never typed in and validation
of correct entry is beyond human capabilities without much straining

A followup to #5218


kyleam (approved):
Sounds good to me.

PR #5222, opened 2020-12-09
RF+ENH: travis - do not explicitly test for annex v 5 vs 6 - rely on default of annex

### LINKED ISSUE #4000: travis: Adjust DATALAD_REPO_VERSION in light of auto-upgrades
This issue is being split off from gh-2969, specifically this comment by @bpoldrack:

> Subsequently we should also change the Travis builds, so they don't suggest to test sth, they actually do not test.

---

The .travis.yml in master sets `DATALAD_REPO_VERSION` to 5 and 6, but, with the minimum git-annex version on master (7.20190503), v6 will get auto-upgraded to the v7.  So, as of now, it makes sense to move the `DATALAD_REPO_VERSION=6` runs to `=7`.

And when neurodebian's git-annex (i.e. what we test with on Travis) is at least 7.20190912, v5 will also auto-upgrade to v7.  I think this becomes a little tricky because, if our minimum git-annex version doesn't auto-upgrade, we'll want to still have a run that tests v5 repos.  So we'd either need to add a run that gets an older git-annex from somewhere else or bump the minimum required version.

Once the minimum git-annex version is above 7.20190912, we should still make sure that we have AnnexRepo tests that cover the auto-upgrading (I think we already do), but I don't see any reason we need dedicated runs for different `DATALAD_REPO_VERSION` (which means we can drop two builds from the matrix).

yarikoptic:
I am not sure how much we benefit from testing against elderly version 5, and 6 is also
not "current".  Depending on version of git-annex installed we would have older or newer
version used by default, so I decided to instead switch to
- test by default against recently found to be "good" recent version of git-annex
- instead of explicit repo version 6, just use older version in those 2 split runs

this way I think we would just gain better coverage of annex versions without dropping
code/logic coverage (yet to be seen)

I have positioned it against `maint` since I feel it could only benefit from testing against more recent annex, but we might want to make it for `master` instead

Closes #4000

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,10 +27,7 @@ env:
     - DATALAD_DATASETS_TOPURL=http://datasets-tests.datalad.org
     # How/which git-annex we install.  conda's build would be the fastest, but it must not
     # get ahead in PATH to not shadow travis' python
-    - _DL_ANNEX_INSTALL_SCENARIO="conda-forge 7.20190819"
-  matrix:
-##1    - DATALAD_REPO_VERSION=6
-    - DATALAD_REPO_VERSION=5
+    - _DL_ANNEX_INSTALL_SCENARIO="conda-forge 8.20201007"
 
 # Disabled since old folks don't want to change workflows of submitting through central authority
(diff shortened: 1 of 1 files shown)


kyleam (approved):
This sounds very good to me.

In my view, this aligns with and closes gh-4000.  Back then I said:

> And when neurodebian's git-annex (i.e. what we test with on Travis) is at least 7.20190912, v5 will also auto-upgrade to v7. [...] So we'd either need to add a run that gets an older git-annex from somewhere else or bump the minimum required version.

conda's in the picture now and is the obvious "somewhere".

yarikoptic:
Great! added closes for that issue and let's proceed!

PR #5236, opened 2020-12-10
datalad/install.py: Don't crash when called without arguments

### LINKED ISSUE #5230: `python -m datalad.install` crashes on master
```
(datalad3-dev) mih@meiner /tmp % python -m datalad.install 
  File "/usr/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/mih/hacking/datalad/git/datalad/install.py", line 432, in <module>
    main()
  File "/home/mih/hacking/datalad/git/datalad/install.py", line 89, in main
    args.version, miniconda_path=args.path_miniconda, batch=args.batch
AttributeError: 'Namespace' object has no attribute 'version'
```

I guess it should say that it needs an argument.

jwodder:
Fixes #5230.

### DIFF
--- a/datalad/install.py
+++ b/datalad/install.py
@@ -71,8 +71,6 @@ def main():
     scm_datalad.add_argument("-b", "--batch", action="store_true")
     scm_datalad.add_argument("--path-miniconda")
     args = parser.parse_args()
-    if args.schema is None:
-        args.schema = "conda-forge"
     if args.env_write_file is not None:
         with open(args.env_write_file, "w"):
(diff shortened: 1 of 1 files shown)


yarikoptic:
should be good for now, but later (as part of the ultimate solution for #5181) we should default to installing `datalad`, not just `git-annex`

Issue #5232, opened 2020-12-10
SSH ControlPath too long


mih:
Discovered when playing with Appveyor: https://ci.appveyor.com/project/mih/appveyor-datalad/build/job/smhtn8ewrnjrskjy

```
======================================================================
ERROR: datalad.support.tests.test_gitrepo.test_GitRepo_ssh_push
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/appveyor/venv3.8.3/lib/python3.8/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/Users/appveyor/venv3.8.3/lib/python3.8/site-packages/datalad/tests/utils.py", line 200, in _wrap_skip_if_on_windows
    return func(*args, **kwargs)
  File "/Users/appveyor/venv3.8.3/lib/python3.8/site-packages/datalad/tests/utils.py", line 274, in _wrap_skip_ssh
    return func(*args, **kwargs)
  File "/Users/appveyor/venv3.8.3/lib/python3.8/site-packages/datalad/tests/utils.py", line 293, in _wrap_skip_nomultiplex_ssh
    return func(*args, **kwargs)
  File "/Users/appveyor/venv3.8.3/lib/python3.8/site-packages/datalad/tests/utils.py", line 755, in _wrap_with_tempfile
    return t(*(arg + (filename,)), **kw)
  File "/Users/appveyor/venv3.8.3/lib/python3.8/site-packages/datalad/tests/utils.py", line 755, in _wrap_with_tempfile
    return t(*(arg + (filename,)), **kw)
  File "/Users/appveyor/venv3.8.3/lib/python3.8/site-packages/datalad/support/tests/test_gitrepo.py", line 690, in test_GitRepo_ssh_push
    pushed = list(repo.push(remote="ssh-remote", refspec="ssh-test"))
  File "/Users/appveyor/venv3.8.3/lib/python3.8/site-packages/datalad/support/gitrepo.py", line 2560, in push
    return list(
  File "/Users/appveyor/venv3.8.3/lib/python3.8/site-packages/datalad/support/gitrepo.py", line 2571, in push_
    yield from self._fetch_push_helper(
  File "/Users/appveyor/venv3.8.3/lib/python3.8/site-packages/datalad/support/gitrepo.py", line 2669, in _fetch_push_helper
    out = self._git_runner.run(
[...]

Issue #4075, opened 2020-01-23
We likely need to sense filesystem on where to establish ssh sockets


yarikoptic:
See https://git-annex.branchable.com/bugs/2_ssh_connection_prompts_for_password/#comment-1d5d35f76f3711f856e415eb097066b0

I guess we would need to sense if default location is good enough, and if not -- check if TMPDIR could be used, where we could establish a directory with sockets and point git-annex to it via `GIT_ANNEX_TMP_DIR` environment variable.

mih:
Here is another scenario for which this is critical: In our own tests we place HOME in a tempdir. If that is on a crippled FS, the following happens unconditionally for all SSH functionality:

```
datalad.support.exceptions.CommandError: CommandError: 'git push --progress --porcelain target master:master' failed with exitcode 128 under /media/mih/Samsung_T5/tmp/datalad_temp_tree_test_publish_target_urlypdrqc63 [err: 'unix_listener: cannot bind to path /media/mih/Samsung_T5/tmp/datalad_temp__nty059y/.cache/datalad/sockets/6258b3a7.MewfydtoaZa4Iuj9: Input/output error
ConnectionOpenFailedError: ''['"'"'ssh'"'"', '"'"'-fN'"'"', '"'"'-o'"'"', '"'"'ControlMaster=auto'"'"', '"'"'-o'"'"', '"'"'ControlPersist=15m'"'"', '"'"'-o'"'"', '"'"'ControlPath=/media/mih/Samsung_T5/tmp/datalad_temp__nty059y/.cache/datalad/sockets/6258b3a7'"'"', '"'"'localhost'"'"']'' failed with exitcode 255 [Failed to open SSH connection (could not start ControlMaster process)]
```

(please give the quoting the recognition it deserves!)

mih:
And consider: https://github.com/datalad/datalad/issues/4198

PR #5238, opened 2020-12-10
RF: Use datalad.locations.sockets as dedicated configuration

mih:
Instead of putting sockets into the cache dir. This makes it possible to
place sockets specifically, rather than all cache content. This will
make it easier to workaround filesystem limitations (gh-4075), and
path length issues (gh-5232).

Keep default location unchanged `<cachedir>/sockets`

However, using this location seems suboptimal. $HOME (which contains `cachedir` typically), may be mounted from another machine. `/tmp` is much more commonly used for this on Linux (X Server, KDE, ...), and more likely to be a local drive.

Using this new config makes things work on OSX re pathlength https://ci.appveyor.com/project/mih/appveyor-datalad/builds/36771014/job/r7yif4oct81baee7

### DIFF
--- a/datalad/interface/common_cfg.py
+++ b/datalad/interface/common_cfg.py
@@ -58,6 +58,13 @@ definitions = {
         'destination': 'global',
         'default': opj(expanduser('~'), 'datalad'),
     },
+    'datalad.locations.sockets': {
+        'ui': ('question', {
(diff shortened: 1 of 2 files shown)


yarikoptic (approved):
Looks LGTM

Issue #5240, opened 2020-12-11
install.py: flexible/modular and easy on user (next round)


yarikoptic:
# Preamble:

This is another attempt to formalize idea, previously done in https://github.com/datalad/datalad/issues/5181 but it already grew too long. With #5181 in mind:

- I felt "schema" is "off" in its meaning, so I will use "method"
- if we are not to use argparse, then reflecting on your [analysis above](https://github.com/datalad/datalad/issues/5181#issuecomment-732344347) I think the easiest/least confusing setup would be `-e|--extra-args "--opt1 --opt2 val ..."`, so consider `[-e "..."]` below to correspond to it.  Every method would pass them into underlying "main" command.

## Note about possible implementation/future extensions

My thinking (might be not without flaws, please voice your concerns): make "methods"s generic (with those methods specific options listed below), so overall setup at large could be just a "specification" which would be fed into methods.  All the components, in particular it is obvious with  `git-annex` and `datalad` components, are just "packages" which could be installed through the same underlying commands for a given methods (well - distribution), and often matching the name in the methods/distribution to the component  (not always the case, thus in datalad-extensions we similarly allow for mapping of pip/github names, see https://github.com/datalad/datalad-extensions/blob/master/extensions.yaml).

So each method "handler" would have signature `def methodsname(component, option1=somedefault, ... extra_args=None)` where those `option*`s are specific for each method (per description above), defaults are provided in the spec (see below), and `extra_args` would be the extra arguments which are passed  from cmdline directly into underlying command  as is (having `conda-env` case in mind, but applies also to ). Data structure (below) would provide defaults for `parameters` . `methodsname` could match above methods (just having `-` replaced with `_`).

Many methods are OS-specific so should not be considered if current OS is not "matching" (so may be 'methods' should be a class to bind relevant aspects/parameters together).

Define a data structure which would provide parameters per component for various methods, and then corresponding "methods handlers" would just use those parameters and options provided on cmdline to install them.  Later we will add other similar components (`datalad-neuroimaging`, `datalad-container`, ...) and this approach should generalize to them as well. So it could be something like below, where empty dict (`{}`) just means that no overrides of defaults, but that method is supported.
```
{
 'datalad' : {
   'apt': {},
   'conda': {
     'channel': 'conda-forge'
   },
   'deb-url': {},
   ...
   'pip': {
     'git': "https://github.com/datalad/datalad",
   }
 },
 'miniconda': {
   'miniconda-installer': {}
 },
 'venv': {
[...]

PR #5243, opened 2020-12-11
Delete datalad/install.py (Moved to its own repository)

jwodder:
This code has been moved to https://github.com/datalad/datalad-installer/ as part of #5240.

NOTE: Some workflows in this repository depend on datalad/install.py.  Do not merge until the new version is stable.


yarikoptic:
probably we should just wait for the datalad-installer to get "released" so we could may be use a versioned release instead of master and so changes to datalad-installer's master would not break datalad CI setup?

yarikoptic:
There is now https://pypi.org/project/datalad-installer/ (thanks @jwodder ) so removing the label

yarikoptic:
@jwodder would you be kind to rebase please, so we have a fresh tests run, time to merge IMHO

jwodder:
@yarikoptic Rebased.

PR #5246, opened 2020-12-14
Make diffstatus decision making more accessible

mih:
This is a key piece of decision making, but it is notoriously hard to read and not well designed (all blame on me). Here I am trying to dissect it and put it back together in a more accessible fashion.

The focus is on being more gentle to humans, but it should not suffer a performance-loss.

The diff will look like a rewrite, but I tried to make the individual commits tell a story.


mih:
Travis error is unrelated and will be fixed via #5247

kyleam (approved):
Thanks, all of these look like improvements to me.

mih:
Thx much!

PR #5229, opened 2020-12-10
Remove old Runner

mih:
Fixed #5003

PR #5247, opened 2020-12-14
Few fixes that address runner behavior

mih:
For example fix the leaking of git-annex debug output that makes the travis run error. All pulled from #5229


kyleam (approved):
Thanks.  

I've also cherry picked and pushed my fixup commit from that PR: 4b00337491 (TST: annexrepo: Fix one-off error in last commit)

Issue #5253, opened 2020-12-15
test_bare_git_version_{1,2} fails on datalad/git-annex  maint (and may be others)


yarikoptic:
```
======================================================================
ERROR: datalad.distributed.tests.test_ria_git_remote.test_bare_git_version_1(None,)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/site-packages/datalad/tests/utils.py", line 750, in _wrap_with_tempfile
    return t(*(arg + (filename,)), **kw)
  File "/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/site-packages/datalad/tests/utils.py", line 750, in _wrap_with_tempfile
    return t(*(arg + (filename,)), **kw)
  File "/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/site-packages/datalad/distributed/tests/test_ria_git_remote.py", line 92, in _test_bare_git_version_1
    ds.repo.enable_remote('bare-git')
  File "/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/site-packages/datalad/support/annexrepo.py", line 1851, in enable_remote
    raise e
  File "/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/site-packages/datalad/support/annexrepo.py", line 1844, in enable_remote
    env=env)
  File "/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/site-packages/datalad/support/annexrepo.py", line 1104, in _run_annex_command
    **kwargs)
  File "/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/site-packages/datalad/cmd.py", line 144, in run_gitcommand_on_file_list_chunks
    results.append(func(cmd, *args, **kwargs))
  File "/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/site-packages/datalad/cmd.py", line 513, in run
    **results,
datalad.support.exceptions.CommandError: CommandError: 'git annex enableremote -c annex.dotfiles=true -c annex.retry=3 bare-git' failed with exitcode 1 under /tmp/datalad_temp__test_bare_git_version_15m7abgnw [out: 'enableremote bare-git 
[...]

PR #5255, opened 2020-12-16
BF: RIA cloning

### LINKED ISSUE #5186: `datalad clone "ria+file:/some/store/#~myds"` leads to undesired establishment of an "annex"
After such a clone, the config of the dataset in the store gets an `[annex]` section that it should not have:

```
% cat config
[core]
        repositoryformatversion = 0
        filemode = true
        bare = true
[datalad "ora-remote"]
        uuid = 08ba549c-2991-4b7f-94c2-0fbfed0914d3
[annex]
        uuid = 37f33fa8-4eff-4c90-bc83-597b3aa16224
        version = 8
```

I suspect this is `git-annex` automatically doing this, because the clone is on the same filesystem and the user had write permissions. As a consequence, the `config` file also has a different owner now (the cloner).

This is a problem. `git-annex` should not become aware of that remote as having an annex directly, because its availability is recorded in the corresponding ORA special remote. In such a configuration, git-annex will now push to the remote directly, and not via ORA. This leads to confusion at minimum, and to data loss in the worst case.

The full list of modifications is here:

```
./14a/5e5b6-d30d-11ea-8836-b4969157768c/annex
./14a/5e5b6-d30d-11ea-8836-b4969157768c/annex/sentinal
./14a/5e5b6-d30d-11ea-8836-b4969157768c/annex/index
./14a/5e5b6-d30d-11ea-8836-b4969157768c/annex/journal
./14a/5e5b6-d30d-11ea-8836-b4969157768c/annex/journal/uuid.log
./14a/5e5b6-d30d-11ea-8836-b4969157768c/annex/othertmp
./14a/5e5b6-d30d-11ea-8836-b4969157768c/annex/sentinal.cache
./14a/5e5b6-d30d-11ea-8836-b4969157768c/annex/journal.lck
./14a/5e5b6-d30d-11ea-8836-b4969157768c/annex/othertmp.lck
./14a/5e5b6-d30d-11ea-8836-b4969157768c/config
```

bpoldrack:
Fixes #5186 

Sits on top of PR #5254 due to #5253, which is not the same but a somewhat related issue.

Despite @mih's hesitance, I'd like to support RIA stores without any ORA remote. Since an ID based structure for standard, bare annex repos is perfectly valid and requires all the same stuff (URL resolution for example). In such a case this fix would not fix but sabotage the intended behavior. With that in mind, this one-liner is a function, that can be enhanced to first query the store for a to be introduced dataset level layout, that would disable the `annex-ignore` setting.

### DIFF
--- a/datalad/core/distributed/clone.py
+++ b/datalad/core/distributed/clone.py
@@ -526,6 +526,10 @@ def clone_dataset(
         lgr.debug('Reinit %s to enable shared access permissions', destds)
         destds.repo.call_git(['init', '--shared={}'.format(reckless[7:])])
 
+    # In case of RIA stores we need to prepare *before* annex is called at all
+    if result_props['source']['type'] == 'ria':
+        postclone_preannex_cfg_ria(destds)
+
     yield from postclonecfg_annexdataset(
         destds,
         reckless,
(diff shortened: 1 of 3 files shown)


datalad/core/distributed/clone.py
```diff
@@ -918,9 +933,6 @@ def postclonecfg_annexdataset(ds, reckless, description=None):
     yield from configure_origins(ds, ds)
 
 
-_handle_possible_annex_dataset = postclonecfg_annexdataset
```
kyleam, line 921: There's also an import in the deprecated `datalad.distribution.clone`.

kyleam (commented):
Thanks.  Looks good to me aside from the import issue I noted.

Even if we expect 0.14 out before another maintenance release, I'm guessing you intended to choose maint rather than master as the base for this PR?

bpoldrack:
Re `maint`: True!

Re import: Indeed. How did git-grep not find it? Anyway, will remove, too.

Thx!

bpoldrack:
crippledFS failures look unrelated. Do they seem famililar, @datalad/developers ?

kyleam:
> crippledFS failures look unrelated. Do they seem famililar, @datalad/developers ?

At least the `test_cached_dataset` one is a regular.  The build is red frequently enough that I'm guilty of not paying much attention to it (especially when I can't confirm with `tools/eval_under_testloopfs` locally).

mih:
I agree that it should go into `maint`. Other than that it looks good to me. The crippled FS failures are old friends.

mih:
Thx!

mih:
I was too quick and did not acknowledge `Remove obsolete alias`. This isn't desired in `maint` and largely unnecessary in `master` (already removed file).

I have cleanup up `maint` by removing the commit and then merged into `master`.

PR #5259, opened 2020-12-17
Restore progress logging behavior when non-interactive

mih:
When JSON progress by git-annex was handled with `ProcessAnnexProgressIndicators` there was no progress logging in non-interactive mode, unless log level was set to 1.

With the switch to `AnnexJsonProtocol` this behavior changed and now non-interactive logging comes out at the default INFO level.

This change standardizes the non-interactive log level to 5, as used in other places too.

The second commit removes the already obsolete `ProcessAnnexProgressIndicators` and its tests.


kyleam (approved):
Sounds fine for me.

For the `ProcessAnnexProgressIndicators` removal, it seems unlikely to be used outside the code base and a search doesn't bring up anything.

mih:
Thx!

adswa:
this fixes the observation I shared in matrix, but I believe it led to #5260

PR #5267, opened 2021-01-02
BF: filter out Nones while estimating total size to download + safeguard dict access

### LINKED ISSUE #5266: `get -J10` TypeError on master but not maint
In a clone of https://github.com/datalad-datasets/machinelearning-books

```
% datalad get -J 10  .
[ERROR  ] unsupported operand type(s) for +: 'int' and 'NoneType' [annexrepo.py:get:1493] (TypeError) 
```

yarikoptic:
Key might be relaxed, so there would be no size information.  Example could be found in
https://github.com/datalad-datasets/machinelearning-books
dataset for  H.DaumeIII-A_Course_in_Machine_Learning.pdf  file

Closes #5266

Might not be the best (re)solution but should make code more robust even if total progress indicator would no longer report correct total whenever it is not known.

Also fixed up that https://github.com/datalad-datasets/machinelearning-books dataset for `G.James_D.Witten_T.Hastie_R.Tibshirani-An_Introduction_to_Statistical_Learning_with_Applications_in_R.pdf` file URL for which no longer worked, so updated to new one.

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -1488,7 +1488,8 @@ class AnnexRepo(GitRepo, RepoInterface):
             files=files_arg,
             jobs=jobs,
             progress=True,
-            total_nbytes=sum(expected_downloads.values()),
+            # filter(bool,   to avoid trying to add up None's when size is not known
+            total_nbytes=sum(filter(bool, expected_downloads.values())),
         )
         results_list = list(results)
         # TODO:  should we here compare fetch_files against result_list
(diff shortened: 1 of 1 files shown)


datalad/support/annexrepo.py
```diff
@@ -3508,15 +3509,15 @@ def _proc_json_record(self, j):
                     # do not crash if no command is reported
                     label=j['action'].get('command', '').capitalize(),
                     unit=' Bytes',
-                    total=float(j['total-size']),
+                    total=float(j.get('total-size', None)),
```
mih, line 3512: I guess that should also be `s/None/0/` as it is done below to prevent `float(None)`, right?

mih (changes requested):
Thanks much! I think there is an issue though, or I am misreading the code.

datalad/support/annexrepo.py
yarikoptic: yes, I guess should be `0` as well as we are not having special handling for None

```suggestion
                    total=float(j.get('total-size', 0)),
```

kyleam (approved):

PR #5108, opened 2020-10-30
Fix test failures on macOS

jwodder:
Fixes #4941.

PR #5269, opened 2021-01-03
BF: create_sibling cp -- use -R for --recursive and -p for --preserve

yarikoptic:
So works also on OSX

could be seen among fails against released version on https://github.com/datalad/git-annex/runs/1638321268?check_suite_focus=true#step:12:2515


mih (approved):
LGTM!

kyleam:
Note to aid any future digging: gh-5108 made these same changes on master.

yarikoptic:
damn... indeed, should have just adopted that PR changes for maint :-/

Issue #4644, opened 2020-06-19
test_local_url_with_fetch: fails on crippled fs


yarikoptic:
As present in #4620 https://github.com/datalad/datalad/pull/4620/checks?check_run_id=788686809 and possibly in other PRs etc (didn't check).
```shell
======================================================================
ERROR: datalad.core.distributed.tests.test_clone.test_local_url_with_fetch
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.5.9/x64/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/opt/hostedtoolcache/Python/3.5.9/x64/lib/python3.5/site-packages/datalad/tests/utils.py", line 541, in newfunc
    return t(*(arg + (d,)), **kw)
  File "/opt/hostedtoolcache/Python/3.5.9/x64/lib/python3.5/site-packages/datalad/tests/utils.py", line 714, in newfunc
    return t(*(arg + (filename,)), **kw)
  File "/opt/hostedtoolcache/Python/3.5.9/x64/lib/python3.5/site-packages/datalad/core/distributed/tests/test_clone.py", line 582, in test_local_url_with_fetch
    ds_cloned.repo.fetch()
  File "/opt/hostedtoolcache/Python/3.5.9/x64/lib/python3.5/site-packages/datalad/support/gitrepo.py", line 2271, in fetch
    git_options=git_options,
  File "/opt/hostedtoolcache/Python/3.5.9/x64/lib/python3.5/site-packages/datalad/support/gitrepo.py", line 2288, in fetch_
    git_options=git_options)
  File "/opt/hostedtoolcache/Python/3.5.9/x64/lib/python3.5/site-packages/datalad/support/gitrepo.py", line 2429, in _fetch_push_helper
    "from nor a tracking branch is set up.".format(action))
ValueError: Neither a remote is specified to fetch from nor a tracking branch is set up.
```

yarikoptic:
reproduces locally with python 3.6 conda environment:
```
$> tools/eval_under_testloopfs 'for s in {1..30}; do python -m nose -s -v datalad/core/distributed/tests/test_clone.py:test_local_url_with_fetch; done'
I: vfat of 10:  /home/yoh/.tmp/datalad-fs-VokNp
10+0 records in
10+0 records out
10321920 bytes (10 MB, 9.8 MiB) copied, 0.00425128 s, 2.4 GB/s
[...]

PR #5275, opened 2021-01-05
BF: reload config of the cloned dataset whenever finished "cloning"

yarikoptic:
It helps to mitigate but does not 100% resolve the
https://github.com/datalad/datalad/issues/4644 -- it still might fail once in a rare while

### DIFF
--- a/datalad/core/distributed/clone.py
+++ b/datalad/core/distributed/clone.py
@@ -547,6 +547,12 @@ def clone_dataset(
             'datalad.clone.reckless', reckless,
             where='local',
             reload=True)
+    else:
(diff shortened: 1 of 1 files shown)


yarikoptic:
a more complete solution is in #5276 but that one is aiming at master, not sure if would be adopted for `maint`.  Thus I will proceed merging this one for now into maint

kyleam:
> a more complete solution is in #5276

But that one doesn't change the potential need for a `reload` here, does it?  As far as I can tell, it just changes how `_need_reload` decides its answer, affecting when this `reload(force=False)` will actually do anything but not whether it's needed.

PR #5278, opened 2021-01-05
RM: install - remove --nosave/save=False

yarikoptic:
As it is of no effect for awhile -- I have not done investigation when it
started to be of no effect, but it was reported on the matrix forum that
it has no effect in 0.13.5-0.13.7

Since 0.14.0 seems to be not "revolutionary" enough , and option is of no effect, I think it would be ok to just kill it

### DIFF
--- a/datalad/distribution/install.py
+++ b/datalad/distribution/install.py
@@ -19,7 +19,6 @@ from datalad.interface.common_opts import (
     recursion_limit,
     location_description,
     jobs_opt,
-    nosave_opt,
     reckless_opt,
 )
 from datalad.interface.results import (
@@ -158,7 +157,6 @@ class Install(Interface):
         description=location_description,
         recursive=recursion_flag,
         recursion_limit=recursion_limit,
-        save=nosave_opt,
         reckless=reckless_opt,
         jobs=jobs_opt,
     )
@@ -174,7 +172,6 @@ class Install(Interface):
             description=None,
             recursive=False,
             recursion_limit=None,
-            save=True,
             reckless=None,
             jobs="auto"):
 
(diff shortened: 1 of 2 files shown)


kyleam (approved):
> As it is of no effect for awhile -- I have not done investigation when it
> started to be of no effect

Perhaps 21f64ead2b (RF+BK: Non-functional `install` RF, documents general approach, 2017-04-23)?

Anyway, sounds good to me.  Thanks.

kyleam:
> Anyway, sounds good to me.

Though perhaps someone has scripts that call `install` with `save=True` or `save=False`, and they didn't notice it had no effect.  Rather than crashing their scripts, it might be nice to use a dummy parameter along the lines of `_NoAnnexDefault` and issue a warning saying the parameter has been a noop for a long time and will be removed in the next release.

yarikoptic:
> > Anyway, sounds good to me.
> 
> Though perhaps someone has scripts that call `install` with `save=True` or `save=False`, and they didn't notice it had no effect. Rather than crashing their scripts, it might be nice to use a dummy parameter along the lines of `_NoAnnexDefault` and issue a warning saying the parameter has been a noop for a long time and will be removed in the next release.

what kind of a revolution would that be? ;) I would better wait for corpses in the basement (https://github.com/datalad/datalad/issues/3496, and https://github.com/datalad/datalad/issues/3496#issuecomment-636873354) to come alive, i.e. not going to bother for this unlikely to be heavily used "feature". Feel welcome to take over

kyleam:
> not going to bother for this unlikely to be heavily used "feature"

It's not about it being a feature.  It's about avoiding potential script breakage when we have an easy way to.

But sure, I agree that perhaps nobody has every done `install(..., save=True)` without thinking, and I'm fine with your judgment call that this isn't worth worrying about.

Issue #5277, opened 2021-01-05
testrepo setup key issue


mih:
Debugging why the basic testrepo setup on windows still does not work, which ultimately led us to disable tons of tests:

```
        self.repo.add_url_to_file("test-annex.dat", fileurl)
        self.repo.commit("Adding a rudimentary git-annex load file")
        self.repo.drop("test-annex.dat")  # since available from URL
```

Either the `add_url_to_file()` call fails, or the `drop()` call fails in `populate()` of the test repo.

On the machine that I am at right now, it is the `drop()` command. The file is present, the source at the end of the URL is present to. Both files are readable. I can delete the annexed file (hence no process seems to be blocking it). Yet the `drop()` fails.

Calling a dedicated `git annex drop test-annex.dat` gives:

```
C:\Users\datalad\AppData\Local\Temp\datalad_temp_testrepo_h0ro_zp8>git annex drop test-annex.dat
drop test-annex.dat
git-annex: content is locked
failed
git-annex: drop: 1 failed
```

nothing seems to change that
```
C:\Users\datalad\AppData\Local\Temp\datalad_temp_testrepo_h0ro_zp8>git annex fsck test-annex.dat
fsck test-annex.dat (checksum...) ok

C:\Users\datalad\AppData\Local\Temp\datalad_temp_testrepo_h0ro_zp8>git annex unlock test-annex.dat
(recording state in git...)

C:\Users\datalad\AppData\Local\Temp\datalad_temp_testrepo_h0ro_zp8>git annex drop test-annex.dat
drop test-annex.dat
git-annex: content is locked
failed
git-annex: drop: 1 failed
```

<details>
<summary>Nothing insightful in the debug log either</summary>

```
C:\Users\datalad\AppData\Local\Temp\datalad_temp_testrepo_h0ro_zp8>git annex --debug drop test-annex.dat
[2021-01-05 21:37:37.6167558] process [5316] read: git ["--git-dir=.git","--work-tree=.","--literal-pathspecs","symbolic-ref","-q","HEAD"]
[2021-01-05 21:37:37.6357331] process [5316] done ExitSuccess
[2021-01-05 21:37:37.6377555] process [12636] read: git ["--git-dir=.git","--work-tree=.","--literal-pathspecs","show-ref","refs/heads/adjusted/dl-test-branch(unlocked)"]
[2021-01-05 21:37:37.660762] process [12636] done ExitSuccess
[...]

PR #5280, opened 2021-01-06
TST: Stop using file:// URLs in test repo annex setup

mih:
git-annex is having issues with those:
https://github.com/datalad/datalad/issues/5277

But even if that is fixed, it is still suboptimal to use a URL scheme as
primary testing target that not only needs to be specifically enabled
first, but is also explicitly advised against.

From https://git-annex.branchable.com/git-annex:

> List of URL schemes that git-annex is allowed to download content from. The default is "http https ftp".
> Think very carefully before changing this; there are security implications. For example, if it's changed to allow "file" URLs, then anyone who can get a commit into your git-annex repository could git-annex addurl a pointer to a private file located outside that repository, possibly causing it to be copied into your repository and transferred on to other remotes, exposing its content.

This change introduces a global HTTP server `datalad.test_http_server`
that is persistently available at test runtime. It could be used to
replace all usage of file:// URLs with http:// URLs. However, in the
scope of this change, only git-annex related usage is modified, but
the use of file:// URLs for Git (submodules) is kept. Argueably, those
are more relevant in practice.

### DIFF
--- a/datalad/__init__.py
+++ b/datalad/__init__.py
@@ -92,6 +92,8 @@ _test_states = {
     'HOME': None,
 }
 
+# handle to an HTTP server instance that is used as part of the tests
+test_http_server = None
 
 def setup_package():
     import os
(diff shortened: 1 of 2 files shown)


datalad/__init__.py
```diff
@@ -92,6 +92,8 @@ def test(module='datalad', verbose=False, nocapture=False, pdb=False, stop=False
     'HOME': None,
 }
 
+# handle to an HTTP server instance that is used as part of the tests
+test_http_server = None
```
yarikoptic, line 96: why not to make it at least `_test_http_server` or even better just place within `_test_states['http_server']` - I see no value in polluting top level namespace

yarikoptic (changes requested):
In general - I do not mind getting rid of testing `file://` since indeed too niche of a use-case while having the issues you pointed out.

datalad/__init__.py
mih: Yes, I agree, we can tune that.

mih:
FTR: I will likely close this again. It does not fix the testrepo tests properly. I was spending a few hours by ripping them out in full, instead. And now I have all core and gitrepo tests running in full and green on windows too.

mih:
I am closing this now. With the associated changes in #5303 it is much less relevant now.

Issue #4958, opened 2020-09-24
maint: failing test  test_ExtractedArchive on maint on NFS


yarikoptic:
eh heh -- merging of https://github.com/datalad/datalad/pull/4931 (before merging #4953 but after #4949 for which it was green) triggered test failure under NFS:
https://travis-ci.org/github/datalad/datalad/jobs/729943926
```
======================================================================
FAIL: datalad.tests.test_archives.test_ExtractedArchive
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/datalad/tests/utils.py", line 554, in _wrap_with_tree
    return t(*(arg + (d,)), **kw)
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/datalad/tests/test_archives.py", line 202, in test_ExtractedArchive
    op.join(fn_archive_obscure, '3.txt')
AssertionError: [' b|;& ΔЙקم๗あ.dbtc / a|;& ΔЙקم๗あ.datc '] != [' b|;& ΔЙקم๗あ.dbtc / a|;& ΔЙקم๗あ.datc ', ' b|;& ΔЙקم๗あ.dbtc /3.txt']
```
when ran in the `maint` branch on `DATALAD_TESTS_SSH=0 _DL_TMPDIR="/tmp/nfsmount" TESTS_TO_PERFORM="datalad.tests datalad.support"` matrix run. 

Let's see if current build after merging #4953 "fixes it"... but I just wanted to make a note since may be some flaky test somehow

yarikoptic:
issue is still persistent, e.g. was present on https://github.com/datalad/datalad/pull/5108

yarikoptic:
ha -- this oldie is a duplicate of a good old #4101 and there also reported seems to be exclusively against maint

PR #5297, opened 2021-01-07
BF(DIRTY): ensure that we never allow for gc to run detached

yarikoptic:
Otherwise subsequent  git  command invocation might fail to find
objects etc, as Adina has found in her example on her laptop:

	#!/usr/bin/sh

	set -e -x

	datalad create ima
	cd ima
	datalad download-url \
		 --archive \
		 --message "Download Imagenette dataset" \
		 'https://s3.amazonaws.com/fast-ai-imageclas/imagenette2-160.tgz'

	cd ../
	datalad create -c text2git -c yoda ml
	cd ml

	mkdir -p data
	datalad clone -d . ../ima data/raw

leading to errors like

	CommandError: 'git clone --progress ../ima /tmp/ml/data/raw' failed with exitcode 128 [err: 'Cloning into '/tmp/ml/data/raw'...
	fatal: failed to copy file to '/tmp/ml/data/raw/.git/objects/2f/f2467efde6304a84a1a2f52eee3f5a5fa5448d': No such file or directory']]

We have numerous spots were we invoke git without some centralization and
common place is the GitWitlessRunner.  That is why in this DIRTY hack I just
overloaded `run` to ensure us passing that configuration setting to git, since
AFAIK it cannot be passed through env var (or am I wrong?)

Since we would be always passing it, I have effectively "disabled"
background gc in GitRepo.gc used in a single script among tools

@adswa - does this diff help (or did you use `maint`?)

### DIFF
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -581,6 +581,18 @@ class GitWitlessRunner(WitlessRunner, GitRunnerBase):
         return results
 
 
+    def run(self, cmd, **kwargs):
+        """Execute a command and communicate with it.
+
+        TODO
+        """
+        assert isinstance(cmd, list), "we do not do strings here"
+        assert cmd[0] == 'git'
+        # We need to ensure that git does not engage gc in the background
+        cmd_ = [cmd[0], '-c', 'gc.autodetach=0'] + cmd[1:]
+        return super().run(cmd_, **kwargs)
+
+
 def readline_rstripped(stdout):
     #return iter(stdout.readline, b'').next().rstrip()
     return stdout.readline().rstrip()
--- a/datalad/core/distributed/tests/test_clone.py
+++ b/datalad/core/distributed/tests/test_clone.py
@@ -580,7 +580,7 @@ def test_cfg_originorigin(path):
     with chpwd(path), swallow_logs(new_level=logging.DEBUG) as cml:
         clone_lev3 = clone('clone_lev2', 'clone_lev3')
         # we called git-annex-init; see gh-4367:
-        cml.assert_logged(msg=r"[^[]*Async run \[('git', 'annex'|'git-annex'), "
+        cml.assert_logged(msg=r"[^[]*Async run \[('git', '-c', 'gc.autodetach=0', 'annex'|'git-annex'), "
                               r"'init'",
                           match=False,
                           level='DEBUG')
--- a/datalad/core/local/tests/test_diff.py
+++ b/datalad/core/local/tests/test_diff.py
@@ -58,14 +58,18 @@ from datalad.api import (
 )
 
 
-def test_magic_number():
+@with_tempfile(mkdir=True)
+def test_magic_number(path):
     # we hard code the magic SHA1 that represents the state of a Git repo
     # prior to the first commit -- used to diff from scratch to a specific
     # commit
     # given the level of dark magic, we better test whether this stays
     # constant across Git versions (it should!)
-    out = GitWitlessRunner().run(
-        'cd ./ | git hash-object --stdin -t tree',
+    runner = GitWitlessRunner(cwd=path)
+    runner.run(['git', 'init'])
+    runner.run(['git', 'commit', '--allow-empty', '-m', 'empty'])
+    out = runner.run(
+        ['git', 'show', '--pretty=format:%T'],
         protocol=StdOutCapture)
     eq_(out['stdout'].strip(), PRE_INIT_COMMIT_SHA)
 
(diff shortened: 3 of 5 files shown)


kyleam:
I'm confused.  In what realistic scenario (i.e. non-foot-shooting `gc.*` config values) would `git gc` prune just-created objects?  Also, if this is the explanation, why would the problem go away when Adina added a sleep?

kyleam:
> Ah, I think you suspect it's a race between the packing and fetch.

My quick attempts to trigger that suggest that it's guarded okay.

<details>
<summary>example</summary>

```sh
#!/bin/sh

set -eu

cd "$(mktemp -d "${TMPDIR:-/tmp}"/pack-race-XXXX)"

git init a
(
    cd a

    git config gc.autoPackLimit 1
    git config gc.auto 1

    echo "abc\ndef\nghi\njkl\nmno" >f0
    for i in $(seq 1 500)
    do
        cp f0 "f$i"
        echo "more $i" >>"f$i"
    done
    git add f*
    git commit -mc --quiet
    echo "~~ before gc"
    find .git/objects | wc
    git gc --auto
)

echo "~~ before clone"
find a/.git/objects | wc
git clone a b
echo "~~ after clone"
find a/.git/objects | wc
echo "~~ after sleep"
sleep 5
find a/.git/objects | wc
```

output:

```
Initialized empty Git repository in /tmp/pack-race-IXHO/a/.git/
~~ before gc
    728     728   31266
Auto packing the repository in background for optimum performance.
See "git help gc" for manual housekeeping.
~~ before clone
    729     729   32758
Cloning into 'b'...
done.
~~ after clone
find: ‘a/.git/objects/57’: No such file or directory
[... 91 lines ...]
find: ‘a/.git/objects/45’: No such file or directory
    304     304   10578
~~ after sleep
      7       7     255
```

</details>

yarikoptic:
> I'm confused. In what realistic scenario (i.e. non-foot-shooting `gc.*` config values) would `git gc` prune just-created objects? Also, if this is the explanation, why would the problem go away when Adina added a sleep?
> (edit: Ah, I think you suspect it's a race between the packing and fetch.)

smth like that, although I did not even try to pin point that well ;)  I think it would be worth reporting to git folks, although I am not sure if we would be able to produce a pure git reproducer, unless may be yours behaves differently on her laptop? (after all we could not reproduce anywhere else)

yarikoptic:
I did not bother fixing up benchmarks for now where git runner is used to run `echo` for timing of overhead... first need to decide is that is the way to go ;-)

kyleam:
> first need to decide is that is the way to go

FWIW my vote is against these changes with the information that's been presented at this point.

yarikoptic:
@adswa BTW -- in your global git config do you have any datalad specific settings? in particular the one to allow abandoning "hanging" batch processes? ;)

yarikoptic:
FTR: the only failing travis test is https://github.com/datalad/datalad/issues/4958 so unrelated

adswa:
> @adswa BTW -- in your global git config do you have any datalad specific settings? in particular the one to allow abandoning "hanging" batch processes? ;)

not aware of it. <details><summary>my global config </summary>

```
user.name=Adina Wagner
user.email=adina.wagner@t-online.de
user.signingkey=61DSS30NA7DCAE36
push.default=simple
color.status=auto
color.branch=auto
core.whitespace=trailing-space,space-before-tab
core.editor=vim
apply.whitespace=fix
alias.co=checkout
alias.st=status
alias.wdiff=diff --color-words
alias.sync=!sh -c 'git checkout --quiet HEAD && git fetch upstream master:maste>
alias.fap=fetch --all --prune
alias.ls-subtrees=!git log | grep git-subtree-dir | tr -d ' ' | cut -d : -f2 | >
annex.security.allowed-url-schemes=http https file
uploadarchive.allowunreachable=true
grep.linenumber=true
pull.rebase=false
datalad.log.level=info
hub.oauthtoken=cd2a3bde239034894734r2r5e9f36454f73b5a404dc
```
</details>

Kyle's script works fine:
```
bash kylesscript.sh
Initialized empty Git repository in /tmp/pack-race-bicK/a/.git/
~~ before gc
    727     727   31250
Auto packing the repository in background for optimum performance.
See "git help gc" for manual housekeeping.
~~ before clone
    728     728   32740
Cloning into 'b'...
done.
~~ after clone
      6       6     222
~~ after sleep
      7       7     255
```

On top of this branch, the error does not surface. But this error seems to be such an unlikely and irreproducible event (only I, on one computer, get it in one not-so-often-used workflow, I have never seen a bug report with this error before), is it really necessary to do something about it in here?

yarikoptic:
ok, nobody got excited about this change as a fix to workaround "documented git shortcoming", so let's let it RiP until cases resurface (e.g. in some automated pipeline)

Issue #5302, opened 2021-01-08
FAIL: datalad.distributed.tests.test_ria_basics.test_remote_layout('datalad-test',)


mih:
```
Traceback (most recent call last):
2017  File "/Users/appveyor/venv3.8.6/lib/python3.8/site-packages/nose/case.py", line 198, in runTest
2018    self.test(*self.arg)
2019  File "/Users/appveyor/venv3.8.6/lib/python3.8/site-packages/datalad/tests/utils.py", line 291, in _wrap_skip_ssh
2020    return func(*args, **kwargs)
2021  File "/Users/appveyor/venv3.8.6/lib/python3.8/site-packages/datalad/tests/utils.py", line 263, in _wrap_skip_if
2022    return func(*args, **kwargs)
2023  File "/Users/appveyor/venv3.8.6/lib/python3.8/site-packages/datalad/tests/utils.py", line 772, in _wrap_with_tempfile
2024    return t(*(arg + (filename,)), **kw)
2025  File "/Users/appveyor/venv3.8.6/lib/python3.8/site-packages/datalad/tests/utils.py", line 772, in _wrap_with_tempfile
2026    return t(*(arg + (filename,)), **kw)
2027  File "/Users/appveyor/venv3.8.6/lib/python3.8/site-packages/datalad/tests/utils.py", line 772, in _wrap_with_tempfile
2028    return t(*(arg + (filename,)), **kw)
2029  File "/Users/appveyor/venv3.8.6/lib/python3.8/site-packages/datalad/distributed/tests/test_ria_basics.py", line 289, in _test_remote_layout
2030    assert_equal(len(ds.repo.whereis('one.txt')), len(whereis) + 1)
2031AssertionError: 2 != 3
```

mih:
I cannot replicate the failure on the macos appveyor workers (anymore). Closing for now, maybe it resurfaces.

mih:
And here it is: https://ci.appveyor.com/project/mih/datalad/builds/37191972/job/7ollki4j3m3l6ocj#L2018

Sadly, "random" behavior...

mih:
This test made negative news before: https://github.com/datalad/datalad/issues/5145

mih:
When I go into a failing run. I see the following `fsck` output for the final and unsuccessful `fsck`:

```
(venv3.8.6) worker-628-001:datalad_temp__test_remote_layouth5lsnix8 appveyor$ git annex fsck -f archive
fsck one.txt 
[...]

Issue #5301, opened 2021-01-08
ERROR: datalad.customremotes.tests.test_archives.test_basic_scenario


mih:
```
Traceback (most recent call last):
848  File "/Users/appveyor/venv3.8.6/lib/python3.8/site-packages/nose/case.py", line 198, in runTest
849    self.test(*self.arg)
850  File "/Users/appveyor/venv3.8.6/lib/python3.8/site-packages/datalad/tests/utils.py", line 593, in _wrap_with_tree
851    return t(*(arg + (d,)), **kw)
852  File "/Users/appveyor/venv3.8.6/lib/python3.8/site-packages/datalad/tests/utils.py", line 772, in _wrap_with_tempfile
853    return t(*(arg + (filename,)), **kw)
854  File "/Users/appveyor/venv3.8.6/lib/python3.8/site-packages/datalad/customremotes/tests/test_archives.py", line 143, in test_basic_scenario
855    cloned_annex.get(fn_extracted)
856  File "/Users/appveyor/venv3.8.6/lib/python3.8/site-packages/datalad/support/gitrepo.py", line 325, in _wrap_normalize_paths
857    result = func(self, files_new, *args, **kwargs)
858  File "/Users/appveyor/venv3.8.6/lib/python3.8/site-packages/datalad/support/annexrepo.py", line 1485, in get
859    results = self._call_annex_records(
860  File "/Users/appveyor/venv3.8.6/lib/python3.8/site-packages/datalad/support/annexrepo.py", line 1139, in _call_annex_records
861    raise e
862  File "/Users/appveyor/venv3.8.6/lib/python3.8/site-packages/datalad/support/annexrepo.py", line 1091, in _call_annex_records
863    out = self._call_annex(
864  File "/Users/appveyor/venv3.8.6/lib/python3.8/site-packages/datalad/support/annexrepo.py", line 987, in _call_annex
865    return runner.run_on_filelist_chunks(
866  File "/Users/appveyor/venv3.8.6/lib/python3.8/site-packages/datalad/cmd.py", line 570, in run_on_filelist_chunks
867    res = self.run(
868  File "/Users/appveyor/venv3.8.6/lib/python3.8/site-packages/datalad/cmd.py", line 400, in run
869    raise CommandError(
[...]

Issue #5300, opened 2021-01-08
FAIL: datalad.plugin.tests.test_plugins.test_no_annex


mih:
```
Traceback (most recent call last):
786  File "C:\Python38-x64\lib\site-packages\nose\case.py", line 198, in runTest
787    self.test(*self.arg)
788  File "C:\Python38-x64\lib\site-packages\datalad\tests\utils.py", line 772, in _wrap_with_tempfile
789    return t(*(arg + (filename,)), **kw)
790  File "C:\Python38-x64\lib\site-packages\datalad\plugin\tests\test_plugins.py", line 196, in test_no_annex
791    assert_repo_status(ds.path)
792  File "C:\Python38-x64\lib\site-packages\datalad\tests\utils.py", line 1747, in assert_repo_status
793    eq_(state_files, oktobefound,
794AssertionError: unexpected content of state "modified": [WindowsPath('C:/DLTMP/datalad_temp_11in84lz/code/inannex')] != []
```

https://ci.appveyor.com/project/mih/appveyor-datalad/builds/37166446/job/5spymsp3vosq3cvf

I cannot replicate this failure on another windows box. And the test passes on a re-run.

Smells like a race-condition of the git filters.

mih:
I tried various way to reproduce this failure inside and outside of appveyor. So far I failed.

Given there is nothing that really can be done about this, I will close this issue, merely leaving a record that it has happened before.

yarikoptic:
it resurfaced again in https://github.com/datalad/datalad/pull/5010 for windows run  : ID=WinP39a2, DTS=datalad.downloaders datalad.interface datalad.plugin , APPVEYOR_BUILD_WORKER_IMAGE=Visual Studio 2019, PY=39-x64 https://ci.appveyor.com/project/mih/datalad/builds/38344401/job/783s60nrfellnkak
```
======================================================================
FAIL: datalad.plugin.tests.test_plugins.test_no_annex
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Python39-x64\lib\site-packages\nose\case.py", line 198, in runTest
    self.test(*self.arg)
  File "C:\Python39-x64\lib\site-packages\datalad\tests\utils.py", line 757, in _wrap_with_tempfile
    return t(*(arg + (filename,)), **kw)
  File "C:\Python39-x64\lib\site-packages\datalad\plugin\tests\test_plugins.py", line 196, in test_no_annex
    assert_repo_status(ds.path)
[...]

Issue #5299, opened 2021-01-08
FAIL: datalad.cmdline.tests.test_main.test_help_np


mih:
```
Traceback (most recent call last):
1483  File "/home/appveyor/dlvenv/lib/python3.8/site-packages/nose/case.py", line 198, in runTest
1484    self.test(*self.arg)
1485  File "/home/appveyor/dlvenv/lib/python3.8/site-packages/datalad/cmdline/tests/test_main.py", line 127, in test_help_np
1486    raise AssertionError(
1487AssertionError: Following lines in --help output were longer than 80 chars:
1488183                         set logging verbosity level. Choose among critical, error, warning, info, debug. Also you can specify an integer <10 to provide even more debugging information
1489119                         execute command by scheduling it via available PBS. For settings, config file will be consulted
1490237   -C PATH               run as if datalad was started in <path> instead of the current working directory. When multiple -C options are given, each subsequent non-absolute -C <path> is interpreted relative to the preceding -C <path>. This
...
```
Linux: https://ci.appveyor.com/project/mih/appveyor-datalad/build/job/cq8t16bcpw02lw4d

Mac: https://ci.appveyor.com/project/mih/appveyor-datalad/build/job/nyba8paqts086vsv

mih:
The failure needs a disconnected `/dev/stdin` to surface, plus a wide terminal session ;-)

It passes on a small terminal, or in an interactive session.

PR #5303, opened 2021-01-08
New cross-platform appveyor setup

mih:
The PR substantially broadens the use of appveyor for testing. The old setup is fully removed and a new one is put in place -- with the hope that it is more structured, and easier to maintain. Key features:

- (almost) full set of tests on Windows, Linux, and MacOS tests (incl SSH connectivity)
- possibility for interactive debugging for all CI runs enabled by default
- per CI-run configurable Python version (a consistent 3.8 at the moment)
- `datalad-installer` integration for flexible deployment of git-annex variant (see examples inside)
- ability to install "system packages" via apt or homebrew by simply listing their names
- Test run environment configured to have reasonably short path names (to avoid mysterious glitches from such failures)
- coverage submission for all platforms
- retains all log files (at least for now) so linking them in issues is meaningful (unlike with github actions). Exhibit A: [First ever appveyor run](https://ci.appveyor.com/project/mih/datalad/build/1.0.1)

The test suite is split into 9 separate runs across all platforms. Each run aims for a duration of ~30min to leave time for debugging without extra fiddling. When appveyor is paid in full (10 workers), this should give a turn-around latency of <45min.

In order to get here (and the expansion of the windows tests in particular) I had to make a number of fixes (split into dedicated commits). The most notable is the discontinuation of the `with_testrepos` decorator (and associated machinery) in the tests of core functionality.

There are still a few tests failing. I have filed dedicated issues for them:

- [x] `git-annex cannot create directory` #5291 (osx)
- [x] `datalad.support.tests.test_parallel.test_gracefull_death` #5296 (win)
- [x] `datalad.distribution.tests.test_create_sibling.test_target_ssh_simple` #5298 (linux)
- [x] `datalad.cmdline.tests.test_main.test_help_np` #5299 (osx)
- [x] `datalad.plugin.tests.test_plugins.test_no_annex` #5300 (win)
- [x] `datalad.customremotes.tests.test_archives.test_basic_scenario` #5301 (osx)
- [x] `datalad.distributed.tests.test_ria_basics.test_remote_layout` #5302 (osx)

In order to be able to use the RDP connection to the windows working, you will need a password that I will provide. For the SSH connection, you will need a dedicated SSH private key. Using dedicated aliases makes the connection relatively simple. I suggest:

```
alias appveyor_rdp="xfreerdp /cert:ignore /dynamic-resolution /u:appveyor /p:PASSWORDFROMME"
alias appveyor_ssh="ssh -i ~/.ssh/id_KEYFROMME_rsa -o StrictHostKeyChecking=no"
```

With these two, one can log in with the credentials shown at the (relatively) top of each CI run log. For example:

```
% appveyor_ssh appveyor@67.225.164.41 -p 22033

% appveyor_rdp /v:67.225.164.53:33816
```

When logging in, you can `touch BLOCK` (linux/mac), or `right-click-new-textfile-BLOCK[.txt]` (win) to block the worker from exiting -- up to the maximum run duration (60min). Please remove blocks in order to make slots free, as soon as you no longer need them.

FTR: Here is the last CI run log from the repository I was working in: https://ci.appveyor.com/project/mih/appveyor-datalad/builds/37168575


.appveyor.yml
```diff
@@ -0,0 +1,283 @@
...
+  - cmd: ssh-keygen -f C:\Users\appveyor\.ssh\id_rsa -N ""
+  - sh: ssh-keygen -f ~/.ssh/id_rsa -N ""
+  # deploy the datalad installer
+  - appveyor DownloadFile https://raw.githubusercontent.com/datalad/datalad-installer/master/src/datalad_installer.py -FileName datalad_installer.py
+
+
+install:
+  # deploy standard SSH config for localhost access on Windows
+  - cmd: tools\ci\appveyor_ssh2localhost.bat
+  - sh: sh tools/ci/appveyor_ssh2localhost.sh
```
yarikoptic, line 212: so it seems that those `ci/appveyor_` aren't really appveyor specific.  Why no to strip `appveyor_` prefix to avoid possible confusion on the purpose?

.appveyor.yml
```diff
@@ -0,0 +1,283 @@
...
+      DTS: datalad.core
+      APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2019
+      # Python version specification is non-standard on windows
+      PY: 38-x64
+    - ID: WinP38a1
+      # ~30min
+      DTS: >
+          datalad.cmdline
+          datalad.customremotes
+          datalad.distribution
```
yarikoptic, line 71: I would have really preferred if all matrix runs tested entire `datalad` package and not subselection of modules.  That is what we do in datalad/git-annex already and seems to be "almost there" for both windows (just few failing and/or unmarked as known fail/skip) and OSX (some odd issues relating to symlinks in the path AFAIK).  See https://github.com/datalad/git-annex/blob/master/.github/workflows/template/build-%7B%7Bostype%7D%7D.yaml.j2#L602 and actions results there for inspiration

- If splitting of runs is needed, best to split like how we do in .travis.yml separating out slow/integration tests into separate runs
  - that would 1) guarantee that we would not miss tests whenever new test files added; 2) needs no CI adjustment if any test file is removed/moved etc

yarikoptic (changes requested):
awesome -- thanks for pushing it forward!

I thought that ultimate goal was to replace travis entirely. If it was so -- `.travis.yml` should have  been removed.  But I guess it is not the case (none of the "special TEMPDIR" setups etc), at least for this PR.

.appveyor.yml
mih: Why do you think that they are not specific? It may be that, right now, nothing special is done in them, but declaring them generic has very limited benefits, and causes entanglement that is undesired.

mih:
> thought that ultimate goal was to replace travis entirely. If it was so -- `.travis.yml` should have been removed. But I guess it is not the case (none of the "special TEMPDIR" setups etc), at least for this PR.

Hmm, I stated in the description that the goal you stated is not the goal.

.appveyor.yml
yarikoptic: https://www.appveyor.com/pricing/ says

All plans have maximum build job execution time of 60 minutes. For longer build times contact us.

Since we are paying, I think it would be very reasonable to contact them and ask to extend e.g. to 180 min, which would also provide better/sufficient time for interactive sessions whenever needed etc.

> but only specific flavors somehow declared "inside"

what do you mean beyond tags which we already have (`slow`, `turtle`, etc)?

.appveyor.yml
yarikoptic: ok

.appveyor.yml
mih: I don't mind asking for longer times. But I very much prefer a complete test run in ~30min, hence the aim for 30min chunks and the commitment to pay 10 workers. If the pro is fast turn around and the con is only the risk of not running some test in a newly introduced module, than I'd say the risk is negligible.

yarikoptic: FWIW my 1c: I would prefer a complete run under a minute, and we could achieve it by running even less tests... but the point is to test well, and speed is IMHO a factor to minimize in code, not by tests sub-selection as we do regularly discover problems in niche tests, and delays in their discovery waste our time later to identify how/when they really emerged.
But oh well, as long as this PR does not drop already existing testing, at least it would not be worse than before, and it would be under 30 minutes making you happy.

mih: I am not sure how you got the impression that tests are dropped. The number of tests that do run has gone up substantially. In fact, never before was that ever even attempted for any windows test setup. It is standard now, bringing it much closer to the level of other OSes.

mih:
Windows is all green now, and a superset of tests that ran on the old appveyor setup now runs in full.

.appveyor.yml
yarikoptic: > In fact, never before was that ever even attempted for any windows test setup

I am confused, since as I have mentioned we just run on entire `datalad` in datalad/git-annex.  BTW -- for the first time in history all marked to be not-skipped tests are passing now against maint/release/master! (git-annex itself has an issue in its tests, [already reported](https://git-annex.branchable.com/bugs/fresh_test_fails_for___34__trust__58____34___-_trust_failed_/)) . Hopefully git-annex itself gets a fix soon and we would turn green to establish a baseline there!

FTR, https://github.com/datalad/git-annex/actions/runs/473493901 

maint: Ran 1175 tests in 5706.491s OK (SKIP=318)
master: Ran 1210 tests in 4514.621s OK (SKIP=320)

yarikoptic: > I am not sure how you got the impression that tests are dropped. 

I never stated that something is dropped. We already did similar sub-selection before, I just expressed my opinion/desire to avoid such subselection, and that this PR is ok with me as long as no tests are dropped from before (for some reason I do not see coverage report now, so cannot attest one way or another)

yarikoptic:
great, we just need to make it all green and get coverage report back to get it merged (I guess could be just pruning other OSes if not easy to fix). travis seems to start stalling on some runs.

mih:
All green now, except MacOSX. And all failures are either directly https://git-annex.branchable.com/bugs/git-annex_get__58___createDirectory__58___does_not_exist/ or have that underneath the surface.

mih:
Plan from conference call: Manually restart travis to overcome docker pull failure, and then merge when travis is green.

mih:
I had no luck with travis. No the docker issue is gone, but 6/8 runs stall. AFAICS they are finished and only coverage submission is missing.

mih:
OK, we seem to have a catch22 situation here. All was green, except the CI that is the subject of replacement. Now we are trying to make it green to be able to not care abut it anymore.

I see two approaches:

- Break this PR up to get all the actual code changes merged first. This will break the old appveyor tests, because the corresponding fixes in the test setup do not come at the same time.
- Take the remaining travis run properties and port them to appveyor, and discontinue travis.

In the conference call we have ruled out (2), so we are left with (1), and have to redo the patches.

mih:
We can close this now. I have started breaking this PR up into smaller pieces, which may or may not get merged. If the teardown issues resurfaces, it can be investigated in that new (hopefully more narrow) context.

PR #5311, opened 2021-01-11
DOC: Include information on subdataset clone sources in `get` docs

mih:
So far this information is only available in the docstring of an
internal helper, and in the handbook

http://handbook.datalad.org/en/latest/beyond_basics/101-148-clonepriority.html

However, it is an important piece of info that should be easily
accessible, especially when the `get` of a subdataset has just failed.
This specific use case also motivates, in part, the placement of this
information in the `get` documentation.

I will cancel travis and appveyor due to the nature of the diff.

### DIFF
--- a/datalad/distribution/get.py
+++ b/datalad/distribution/get.py
@@ -107,8 +107,9 @@ def _get_remotes_having_commit(repo, commit_hexsha, with_urls_only=True):
 def _get_flexible_source_candidates_for_submodule(ds, sm):
     """Assemble candidate locations from where to clone a submodule
 
-    The following locations candidates are considered. For each candidate a
-    cost is given in parenthesis, lower values indicate higher cost:
+    The following location candidates are considered. For each candidate a
+    cost is given in parenthesis, higher values indicate higher cost, and
+    thus lower priority:
 
     - URL of any configured superdataset remote that is known to have the
       desired submodule commit, with the submodule path appended to it.
@@ -672,7 +673,7 @@ class Get(Interface):
     """Get any dataset content (files/directories/subdatasets).
 
     This command only operates on dataset content. To obtain a new independent
-    dataset from some source use the `install` command.
+    dataset from some source use the `clone` command.
 
     By default this command operates recursively within a dataset, but not
     across potential subdatasets, i.e. if a directory is provided, all files in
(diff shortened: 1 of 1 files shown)


datalad/distribution/get.py
```diff
@@ -685,6 +685,54 @@ class Get(Interface):
     and possibly assigned remote priorities), unless a specific source is
     specified.
 
+    *Getting subdatasets*
+
+    Just a DataLad supports getting file content from more than one location,
```
adswa, line 690: s/a/as/

datalad/distribution/get.py
```diff
@@ -685,6 +685,54 @@ class Get(Interface):
...
+    the same supported for subdatasets, including a ranking of individual
```
adswa, line 691: s/the same/the same is/

datalad/distribution/get.py
```diff
@@ -685,6 +685,54 @@ class Get(Interface):
...
+    sources for prioritization.
+
+    The following locations candidates are considered. For each candidate a
```
adswa, line 694: s/locations/location/

datalad/distribution/get.py
```diff
@@ -685,6 +685,54 @@ class Get(Interface):
...
+    cost is given in parenthesis, lower values indicate higher cost:
```
adswa, line 695: maybe add something along the lines of "and thus lower priority"?

adswa (commented):
I believe its useful to have this documented in the sourcecode. But maybe its useful to add a link to the handbook section to this docstring, too? I can add an identifier for redirection, if needed.

mih:
@adswa Thx for the review! Yes adding a link to a handbook redirect would be good!

datalad/distribution/get.py
mih: Done

datalad/distribution/get.py
mih: Done.

datalad/distribution/get.py
mih: Fixed here, and in the original source.

datalad/distribution/get.py
mih: Yeah, it is even more broken. Fixed.

adswa:
http://handbook.datalad.org/r.html?clone-priority should work in a bit

mih:
Thx @adswa

kyleam (approved):
Looks good to me.

My only suggestion would be to strip the text from `_get_flexible_source_candidates_for_submodule` and have it point to `Get`s docstring to avoid duplicating the text, but I'm okay with it as is.

PR #5316, opened 2021-01-12
BF: PWD doesn't exist on windows

### LINKED ISSUE #5315: ERROR: datalad.customremotes.tests.test_archives.test_get_git_environ_adjusted
```
573  File "C:\Python38-x64\lib\site-packages\nose\case.py", line 198, in runTest
574    self.test(*self.arg)
575  File "C:\Python38-x64\lib\site-packages\datalad\customremotes\tests\test_archives.py", line 192, in test_get_git_environ_adjusted
576    assert_equal(sys_env["PWD"], os.environ.get("PWD"))
577KeyError: 'PWD'
```

https://ci.appveyor.com/project/mih/datalad/builds/37223213/job/1qfy2dxj6bihru2i#L569

bpoldrack:
Test is about getting system env correctly. So, instead of relying on
PWD being there, set an artificial one and see it passed on.

Closes #5315

### DIFF
--- a/datalad/customremotes/tests/test_archives.py
+++ b/datalad/customremotes/tests/test_archives.py
@@ -174,7 +174,6 @@ def test_annex_get_from_subdir(topdir):
         assert_true(annex.file_has_content(fpath))              # and verify if file got into directory
 
 
-@known_failure_githubci_win
 def test_get_git_environ_adjusted():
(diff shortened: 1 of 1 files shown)


kyleam (approved):
Looks good.

(As a side note, I was surprised to see that this test lives in test_archives.)

PR #5322, opened 2021-01-12
RF(TST): invoke setup-docker-ssh without explicit --from

### LINKED ISSUE #5288: address rate limiting on "docker pull" of our ssh image
here and there we now keep getting following errors:

```
latest: Pulling from dataladtester/docker-ssh-target
anyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
The command "tools/ci/prep-travis-forssh.sh" failed and exited with 1 during .
Your build has been stopped.
```

which is due to 

    The rate limits of 100 container image requests per six hours for anonymous usage, 
    and 200 container image requests per six hours for free Docker accounts are now in effect

I guess possible ways to mitigate:
- :dollar: pay
- :rocket: retry a few times with some delays.  our CI runs long enough that we would likely "escape" the exceeding the limit 
- :-1: setup to use free account to double the limit 
- :heart: add images to https://github.com/datalad-tester/dev-containers , host those images ourselves (e.g. within github LFS, OSF, and/or datasets.datalad.org)  and use datalad-container
  -  my main concern is that to achieve that we need to have datalad and datalad-container installed first and it might "taint" our testing :-/ possible mitigation: lean helper which would just use git-annex directly to `get` that container files, and run the same `docker import` etc as datalad-container does.
- :eyes: : there is also caching, which boils down to the same docker save/load: https://stackoverflow.com/questions/32866599/can-travis-ci-cache-docker-images but travis advises against caching them: https://docs.travis-ci.com/user/caching/#things-not-to-cache 

WDYT (can vote, sorry -- no votes for :dollar: ;)) -- may be there is some better way I do not see?

yarikoptic:
This should make it first download our tarball export of the docker container
and load it, thus avoiding docker pull AFAIK

Corresponding changes were directly pushed to https://github.com/datalad-tester/docker-ssh-target and mirror of that repository with the actual tarball was shared at http://datasets.datalad.org/?dir=/datalad/docker-ssh-target from which now `setup` script would fetch the tarball

If works, Closes #5288 , and similar changes (just removing `--from`) should be introduced in other CI setups

### DIFF
--- a/tools/ci/prep-travis-forssh.sh
+++ b/tools/ci/prep-travis-forssh.sh
@@ -33,8 +33,7 @@ ssh-keygen -f /tmp/dl-test-ssh-id -N ""
 curl -fSsL \
      https://raw.githubusercontent.com/datalad-tester/docker-ssh-target/master/setup \
      >setup-docker-ssh
-sh setup-docker-ssh --key=/tmp/dl-test-ssh-id.pub -2 \
-   --from=dataladtester/docker-ssh-target:latest
+sh setup-docker-ssh --key=/tmp/dl-test-ssh-id.pub -2
 
 until nc -vz localhost 42241 && nc -vz localhost 42242
 do sleep 1


yarikoptic:
crap, submitted with the wrong base (master) and only then changed to `maint`. not sure if travis would pick up the correct one

kyleam:
> not sure if travis would pick up the correct one

It didn't.  I'll push a no-change rewrite to retrigger.

kyleam (approved):

PR #5330, opened 2021-01-14
Like #5303, but just the minimum to make appveyor run

mih:
This is a strict and unmodified subset of commits from #5303. In contrast to that full PR, it does not broaden the testing coverage on windows, but merely introduces the new appveyor setup, and the minimal number of fixes to make the appveyor CI runs pass.

If travis does not stall on this one, it could be merged to break the appveyor spell on windows. Subsequent improvements could be fetched from a broken up or stripped down #5303


yarikoptic:
- Coverage, where are you???

mih:
On first attempt, one if the macos runs stalls on appveyor. I am OK to disable this platform entirely.

mih:
Rerunning the test was much faster, and passed.

yarikoptic:
coverage reporting is odd... [it seems to be submitted](https://ci.appveyor.com/project/mih/datalad/builds/37263456/job/phycvj1n7ec23elx#L906) from a (sample) windows run, and seems for the correct commit (https://codecov.io/github/datalad/datalad/commit/46feb827e47cfef4585195dd8a3f7731eafb9665) but [reports drop in a (sample) windows-specific spot](https://codecov.io/gh/datalad/datalad/pull/5330/changes) so either coverage is indeed dropped (some code/tests which ran before do no longer run) or codecov/coverage misbehaving?

mih:
> coverage reporting is odd... [it seems to be submitted](https://ci.appveyor.com/project/mih/datalad/builds/37263456/job/phycvj1n7ec23elx#L906) from a (sample) windows run, and seems for the correct commit (https://codecov.io/github/datalad/datalad/commit/46feb827e47cfef4585195dd8a3f7731eafb9665) but [reports drop in a (sample) windows-specific spot](https://codecov.io/gh/datalad/datalad/pull/5330/changes) so either coverage is indeed dropped (some code/tests which ran before do no longer run) or codecov/coverage misbehaving?

I have no idea. However, the aim is to bring up the actual coverage, and this is just an intermediate step. I would prefer to be able to do things in small step rather than larger chunks. This is the smallest that I can produce. If that is not acceptable, I cannot help ATM, and we can close this again.

mih:
I will not investigate the coverage mystery. The only cases where tests are no longer executed are due to the test operation under wrong assumptions. The diff rectifies the assumption tests. No test is disabled unconditionally.

Let's move on.

PR #5331, opened 2021-01-15
Less with_testrepos for a more test running on all platforms

mih:
This is a tuned and trimmed spin-off from #5303

### DIFF
--- a/datalad/core/distributed/tests/test_clone.py
+++ b/datalad/core/distributed/tests/test_clone.py
@@ -265,7 +265,6 @@ def test_clone_isnot_recursive(path_src, path_nr, path_r):
         {'subm 1', '2'})
 
 
-@slow  # 23.1478s
 @with_tempfile
 @with_tempfile
 def test_clone_into_dataset(source_path, top_path):
--- a/datalad/core/distributed/tests/test_clone.py
+++ b/datalad/core/distributed/tests/test_clone.py
@@ -247,10 +247,14 @@ def test_clone_dataladri(src, topurl, path):
     ok_file_has_content(ds.pathobj / 'test.txt', 'some')
 
 
-@with_testrepos('submodule_annex', flavors=['local', 'local-url', 'network'])
 @with_tempfile(mkdir=True)
 @with_tempfile(mkdir=True)
-def test_clone_isnot_recursive(src, path_nr, path_r):
+@with_tempfile(mkdir=True)
+def test_clone_isnot_recursive(path_src, path_nr, path_r):
+    src = Dataset(path_src).create()
+    src.create('subm 1')
+    src.create('2')
+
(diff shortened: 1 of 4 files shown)


yarikoptic:
observation: #5330 dropped coverage (from 90.35% to  89.86%), this one bring is back up a little toward what it was before.

datalad/core/distributed/tests/test_clone.py
```diff
@@ -262,14 +266,10 @@ def test_clone_isnot_recursive(src, path_nr, path_r):
 
 
 @slow  # 23.1478s
```
yarikoptic, line 268: what timing do you observe now? may be without testrepos it got speedy and does not need `@slow`?

datalad/support/tests/test_annexrepo.py
```diff
@@ -694,11 +706,12 @@ def test_AnnexRepo_get_file_backend(src, dst):
     eq_(ar.get_file_backend('test-annex.dat'), 'SHA1')
 
 
-@known_failure_windows
 @with_tempfile
 def test_AnnexRepo_always_commit(path):
 
     repo = AnnexRepo(path)
+    if repo.is_managed_branch():
+        raise SkipTest('Test logic invalid for adjusted repos')
```
yarikoptic, line 714: @kyleam in #5325 introduced `@skip_if_adjusted_branch` which should be used here instead of the conditional

yarikoptic (commented):
Left a comment about a decorator... all initial steps for populating a repo really begs for some `@with_repo` which could be just using `@with_tree` and do init for a desired Annex or Git class.

datalad/core/distributed/tests/test_clone.py
mih: 7.3s on a cold run of just this single test:

`python -m nose -s -v datalad.core.distributed.tests.test_clone:test_clone_into_dataset`

What is the desired threshold for slow?

datalad/support/tests/test_annexrepo.py
mih: Done.

datalad/core/distributed/tests/test_clone.py
mih: Removed

yarikoptic:
```
def slow(f):
    """Mark test as a slow, although not necessarily integration or usecase test

    Rule of thumb cut-off to mark as slow is 10 sec
    """
    return attr('slow')(f)
```
I will push some quick tune up to CONTRIBUTING.md to reflect that

yarikoptic: pushed 193bd76b19beaf02fcdef984febdec58f1cbcb38 to maint

mih:
OK, I addressed all issues raised.

PR #5332, opened 2021-01-15
TST: Stop using file:// URLs in test repo annex setup

mih:
git-annex is having issues with those:
https://github.com/datalad/datalad/issues/5277

But even if that is fixed, it is still suboptimal to use a URL scheme as
primary testing target that not only needs to be specifically enabled
first, but is also explicitly advised against.

From https://git-annex.branchable.com/git-annex:

> List of URL schemes that git-annex is allowed to download content from. The default is "http https ftp".
> Think very carefully before changing this; there are security implications. For example, if it's changed to allow "file" URLs, then anyone who can get a commit into your git-annex repository could git-annex addurl a pointer to a private file located outside that repository, possibly causing it to be copied into your repository and transferred on to other remotes, exposing its content.

This change introduces a global HTTP server `datalad.test_http_server`
that is persistently available at test runtime. It could be used to
replace all usage of file:// URLs with http:// URLs. However, in the
scope of this change, only git-annex related usage is modified, but
the use of file:// URLs for Git (submodules) is kept. Argueably, those
are more relevant in practice.

### DIFF
--- a/datalad/__init__.py
+++ b/datalad/__init__.py
@@ -97,6 +97,8 @@ _test_states = {
     'env': {},
 }
 
+# handle to an HTTP server instance that is used as part of the tests
+test_http_server = None
 
 def setup_package():
     import os
@@ -144,6 +146,21 @@ def setup_package():
 	email = test@example.com
 """)
         _TEMP_PATHS_GENERATED.append(new_home)
+
+    # in order to avoid having to fiddle with rather uncommon
+    # file:// URLs in the tests, have a standard HTTP server
+    # that serves an 'httpserve' directory in the test HOME
+    # the URL will be available from datalad.test_http_server.url
+    from datalad.tests.utils import HTTPPath
+    import tempfile
+    global test_http_server
+    serve_path = tempfile.mkdtemp(
+        dir=cfg.get("datalad.tests.temp.dir"),
+        prefix='httpserve',
+    )
+    test_http_server = HTTPPath(serve_path)
+    test_http_server.start()
+    _TEMP_PATHS_GENERATED.append(serve_path)
 
     # Re-load ConfigManager, since otherwise it won't consider global config
     # from new $HOME (see gh-4153
@@ -234,6 +251,9 @@ def teardown_package():
     ui.set_backend(_test_states['ui_backend'])
     if _test_states['loglevel'] is not None:
         lgr.setLevel(_test_states['loglevel'])
+
+    if test_http_server:
+        test_http_server.stop()
 
     from datalad.tests import _TEMP_PATHS_GENERATED
     if len(_TEMP_PATHS_GENERATED):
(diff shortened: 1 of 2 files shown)


mih:
And I have isolated the change that originally derailed #5303  @bpoldrack

mih:
I will rebase this on the current master and get a fresh run. I am not seeing any feedback on this PR yet, although I think I remember some discussion of it -- not remembering any conclusions, though.

datalad/__init__.py
```diff
@@ -145,6 +147,21 @@ def set_envvar(v, val):
...
+    # the URL will be available from datalad.test_http_server.url
+    from datalad.tests.utils import HTTPPath
+    import tempfile
+    global test_http_server
+    serve_path = tempfile.mkdtemp(
+        dir=cfg.get("datalad.tests.temp.dir"),
+        prefix='httpserve',
+    )
+    test_http_server = HTTPPath(serve_path)
+    test_http_server.start()
```
kyleam, line 162: `HTTPPath.start` patches the environment, so this interacts badly with the rest of the setup/teardown logic.

For example, running a random test locally (without explicitly setting `DATALAD_LOG_LEVEL` in the environment) gives this error at teardown:

```
ERROR: test suite for <module 'datalad.core.local.tests.test_save' from '/home/kyle/src/python/datalad/datalad/core/local/tests/test_save.py'>
  File "/home/kyle/src/python/venvs/datalad/lib/python3.7/site-packages/nose/suite.py", line 229, in run
    self.tearDown()
  File "/home/kyle/src/python/venvs/datalad/lib/python3.7/site-packages/nose/suite.py", line 352, in tearDown
    self.teardownContext(ancestor)
  File "/home/kyle/src/python/venvs/datalad/lib/python3.7/site-packages/nose/suite.py", line 368, in teardownContext
    try_run(context, names)
  File "/home/kyle/src/python/venvs/datalad/lib/python3.7/site-packages/nose/util.py", line 471, in try_run
    return func()
  File "/home/kyle/src/python/datalad/datalad/__init__.py", line 272, in teardown_package
    os.environ.pop(v)
  File "/usr/lib/python3.7/_collections_abc.py", line 795, in pop
    value = self[key]
  File "/usr/lib/python3.7/os.py", line 678, in __getitem__
    raise KeyError(key) from None
KeyError: 'DATALAD_LOG_LEVEL'
```

I think the most basic fix would be to move the above code to the end of `setup_package`.  That way, `setup_package` modifies the environment before it is patched by `HTTPPath.start`, and `HTTPPath.stop` restores the original environment before `teardown_package` reverse `setup_package`s modifications.

mih: Thx for the suggestion! Done.

mih:
Test failures are the current `brew` issue on the mac and the familiar #5300

kyleam (approved):
Looks good to me.

mih:
Thx @kyleam ! Let's see where this will take us...

Issue #5281, opened 2021-01-06
CommandError not informative for failing JSON-based annex commands


mih:
```diff
Date:   Wed Jan 6 10:42:03 2021 +0100

    ENH: Report error messages from JSON records in str(CommandError)

    Not doing that leaves users largely clueless as to why a JSON-based annex
    command might have failed, without interactively debugging it.

    This change turn the previous

    ```
    CommandError:
     '"git" "annex" "drop" "--json" "--json-error-messages" --" "test-annex.dat"'
     failed with exitcode 1 under C:\Users\...\datalad_temp_testrepo_08ix2heg
     [err: 'git-annex: drop: 1 failed']
     [info keys: stdout_json]
    ```

    into

    ```
    CommandError:
     '"git" "annex" "drop" "--json" "--json-error-messages" --" "test-annex.dat"'
     failed with exitcode 1 under C:\Users\...\datalad_temp_testrepo_08ix2heg
     [err: 'git-annex: drop: 1 failed']
     [errors from JSON records: [['git-annex: content is locked']]]
     [info keys: stdout_json]
    ```

    The markup is a bit convoluted, but there could be any number of error messages
    per result and also any number of results.

diff --git a/datalad/support/exceptions.py b/datalad/support/exceptions.py
index 19d6b7e4d..b9aa4688d 100644
--- a/datalad/support/exceptions.py
+++ b/datalad/support/exceptions.py
@@ -60,6 +60,11 @@ class CommandError(RuntimeError):
         if self.stderr:
             to_str += " [err: '{}']".format(ensure_unicode(self.stderr).strip())
         if self.kwargs:
+            if 'stdout_json' in self.kwargs:
+                to_str += " [errors from JSON records: {}]".format(
+                    [m['error-messages'] for m in self.kwargs['stdout_json']
+                     if 'error-messages' in m]
+                )
             to_str += " [info keys: {}]".format(
                 ', '.join(self.kwargs.keys()))
         return to_str
```

PR #5333, opened 2021-01-15
ENH: Report error messages from JSON records in str(CommandError)

mih:
Not doing that leaves users largely clueless as to why a JSON-based annex
command might have failed, without interactively debugging it.

This change turn the previous

```
CommandError:
 '"git" "annex" "drop" "--json" "--json-error-messages" --" "test-annex.dat"'
 failed with exitcode 1 under C:\Users\...\datalad_temp_testrepo_08ix2heg
 [err: 'git-annex: drop: 1 failed']
 [info keys: stdout_json]
```

into

```
CommandError:
 '"git" "annex" "drop" "--json" "--json-error-messages" --" "test-annex.dat"'
 failed with exitcode 1 under C:\Users\...\datalad_temp_testrepo_08ix2heg
 [err: 'git-annex: drop: 1 failed']
 [errors from JSON records: [['git-annex: content is locked']]]
 [info keys: stdout_json]
```

The markup is a bit convoluted, but there could be any number of error messages
per result and also any number of results.

"Error" messages are taken from either/or 'note' or 'error-messages' fields.

Fixes gh-5281

### DIFF
--- a/datalad/support/exceptions.py
+++ b/datalad/support/exceptions.py
@@ -60,6 +60,17 @@ class CommandError(RuntimeError):
         if self.stderr:
             to_str += " [err: '{}']".format(ensure_unicode(self.stderr).strip())
         if self.kwargs:
+            if 'stdout_json' in self.kwargs:
+                from datalad.utils import unique
+                json_errors = unique([
+                    '{}{}'.format(
+                        m['note'] if m.get('note') else m.get('error-messages'),
+                        m['error-messages'] if m.get('error-messages') else '')
+                    for m in self.kwargs['stdout_json']
+                    if m.get('error-messages') or m.get('note')
+                ])
+                if json_errors:
+                    to_str += " [errors from JSON records: {}]".format(json_errors)
             to_str += " [info keys: {}]".format(
                 ', '.join(self.kwargs.keys()))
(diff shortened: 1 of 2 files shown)


datalad/support/exceptions.py
```diff
@@ -60,6 +60,17 @@ def to_str(self, include_output=True):
         if self.stderr:
             to_str += " [err: '{}']".format(ensure_unicode(self.stderr).strip())
         if self.kwargs:
+            if 'stdout_json' in self.kwargs:
+                from datalad.utils import unique
+                json_errors = unique([
+                    '{}{}'.format(
+                        m['note'] if m.get('note') else m.get('error-messages'),
+                        m['error-messages'] if m.get('error-messages') else '')
```
kyleam, line 68: In the case that 'note' is false and 'error-messages' is not, wouldn't the string formatted as '{error-messages}{error-messages}'?  Admittedly, I can't come up with a git-annex call that has an error-messages field but no note.

Aside from being confused by that, this PR looks like a clear improvement to me.

mih: Yeah, that is probably an unlikely event, but still needs fixing IMHO. Will do. Thx!

Issue #5338, opened 2021-01-18
test_expanduser fails on master/windows on github CI


yarikoptic:
Just spotted within datalad/git-annex sweep of tests (ok on released, maint):
Tests were passing 5 days back, so possibly triggered by #5310

https://github.com/datalad/git-annex/runs/1722165033?check_suite_focus=true

```
======================================================================
ERROR: datalad.core.distributed.tests.test_clone.test_expanduser
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\site-packages\nose\case.py", line 198, in runTest
    self.test(*self.arg)
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\site-packages\datalad\tests\utils.py", line 757, in _wrap_with_tempfile
    return t(*(arg + (filename,)), **kw)
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\site-packages\datalad\tests\utils.py", line 757, in _wrap_with_tempfile
    return t(*(arg + (filename,)), **kw)
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\site-packages\datalad\core\distributed\tests\test_clone.py", line 545, in test_expanduser
    cloneds = clone(op.join('~', 'src'), 'dest')
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\site-packages\datalad\interface\utils.py", line 482, in eval_func
    return return_func(generator_func)(*args, **kwargs)
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\site-packages\datalad\interface\utils.py", line 470, in return_func
    results = list(results)
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\site-packages\datalad\interface\utils.py", line 457, in generator_func
    msg="Command did not complete successfully")
datalad.support.exceptions.IncompleteResultsError: Command did not complete successfully. 1 failed:

[{'action': 'install',
  'message': ('Failed to clone from any candidate source URL. Encountered '
              'errors per each url were:\n'
              '- %s',
              '~\\src\n'
              '  CommandError: \'"git" "clone" "--progress" '
              '"C:\\Users\\runneradmin\\AppData\\Local\\Temp\\datalad_temp_zl9rmwun\\src" '
[...]

PR #5339, opened 2021-01-18
TEMP appveyor: test with python 3.6 on windows (instead of 3.8)

yarikoptic:
inspired by https://github.com/datalad/datalad/issues/5338


mih:
That "worked". Looking into it.

mih:
I can replicate the particular test failure, but I cannot determine the source of the problem. The repository labeled as "does not appear to be a git repository" is perfectly clonable via `git clone` as well as `datalad clone`. The path is also expanded to the correct location.

That being said: Is there any specific reason for using PY36 on windows? It is approaching its EOL, and IMHO energy is better spent on achieving reasonable functionality on a recent Python, rather than on all Python versions.

yarikoptic:
It's our minimal supported version of python ATM according to setup.py

mih:
> It's our minimal supported version of python ATM according to setup.py

That is true. But I was wondering whether that makes any sense. AFAIK that number is motivated by Debian stable and HPC platform availability concerns. Neither of which have any implications on Windows. The Q for me is: Do we have anyone that cares for PY 3.6/7/8 on Windows, and anyone that cares enough about the functionality that may or may not be available in this specific scenario. I personally do not care, and I am also not aware of anything that would justify spending time on this.

Happy to tag 'wontfix' and move on, from my side.

PR #5340, opened 2021-01-19
Expand appveyor setup

mih:
Broaden the coverage of older Pythons, without going orthogonal across, platforms.

- Windows 3.9 (hard to see why anything other than the latest would make more sense here)
- Mac 3.8
- Ubuntu 3.6, 3.7, 3.8 (because CI workers have max performance on this platform)

Also introduce a small debug setup helper to ease interactive sessions on windows.


mih:
More test runs -> less coverage. Ok.

I will nevertheless merge.

PR #5341, opened 2021-01-19
TST: Test uuids to exist before comparison

### LINKED ISSUE #4621: Something is flaky in crippled FS workflow

bpoldrack:
In some cases a few test runs end up comparing uuids that fail to exist in the first place, leading to a comparison that evaluates to `None == None`. Make the assumption that there actually is an uuid explicit in order to fail in the spot where it's going wrong rather than a much less telling point later on.

- Closes #4621

### DIFF
--- a/datalad/core/distributed/tests/test_clone.py
+++ b/datalad/core/distributed/tests/test_clone.py
@@ -197,6 +197,7 @@ def test_clone_simple_local(src, path):
         # no content was installed:
         ok_(not ds.repo.file_has_content('test-annex.dat'))
         uuid_before = ds.repo.uuid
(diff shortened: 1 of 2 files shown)


mih (approved):
Makes sense!

kyleam (approved):
Makes sense to me too.  Thanks.

Issue #5309, opened 2021-01-10
FAIL: datalad.support.tests.test_parallel.test_gracefull_death


mih:
Failed on 5th re-run of identical code.

```
Traceback (most recent call last):
2096  File "/Users/appveyor/venv3.8.6/lib/python3.8/site-packages/nose/case.py", line 198, in runTest
2097    self.test(*self.arg)
2098  File "/Users/appveyor/venv3.8.6/lib/python3.8/site-packages/datalad/tests/utils.py", line 263, in _wrap_skip_if
2099    return func(*args, **kwargs)
2100  File "/Users/appveyor/venv3.8.6/lib/python3.8/site-packages/datalad/support/tests/test_parallel.py", line 215, in test_gracefull_death
2101    assert_equal(sorted(inner()), [0, 1])
2102AssertionError: Lists differ: [1, 2] != [0, 1]
2103
2104First differing element 0:
21051
21060
2107
2108- [1, 2]
2109+ [0, 1]
```
- https://ci.appveyor.com/project/mih/datalad/build/job/alrk337e10quig3c#L2093
- https://ci.appveyor.com/project/mih/datalad/builds/37220881/job/hmnqn4ij7iqb5tgf#L2081
- https://ci.appveyor.com/project/mih/datalad/builds/37291931/job/1uqu1bw1sa7secdo#L2088

yarikoptic:
FWIW: failed to reproduce on local OSX (yet to try to enter appveyor), or 1 core Linux VM (suspected that may be it is the reason) :-/  not sure if I will embark on appveyor debug route, so I will submit a PR to just announce it as a known/skip on OSX.  It isn't critical (just is testing expected behavior in case of being killed) so I think it is ok

yarikoptic:
also could fail differently
```
FAILED ../datalad/support/tests/test_parallel.py::test_gracefull_death - assert 100 > 107

```

<details>
<summary>some grep to get some idea on how often happened and with what in the past month</summary> 

```shell
(git)smaug:/mnt/datasets/datalad/ci/logs/2022/08[master]git
$> git grep -l '^FAIL.*test_gracefull_dea' | xargs grep -B4 ': Assertion' 
03/pr/6915/964b1f4/travis-14378-failed/2.txt-        if msg is None:
03/pr/6915/964b1f4/travis-14378-failed/2.txt->           assert first > second
03/pr/6915/964b1f4/travis-14378-failed/2.txt-E           assert 100 > 121
[...]

PR #5342, opened 2021-01-19
RM: remove url from GitRepo.__init__ -- was deprecated in 0.5.0

yarikoptic:
Note: no other uses of DeprecatedError exception ATM


kyleam (approved):
Sounds good to me, and I think it's probably safe to assume that nobody was relying on passing `runner` or anything beyond as a positional argument.

kyleam:
The remaining failure is the known test_parallel one from gh-5309.

https://ci.appveyor.com/project/mih/datalad/builds/37343210/job/i32hq1h5gfnhrmj5#L2095

mih (approved):
Very happy to see that!

PR #5343, opened 2021-01-19
ENH: introduce use of @deprecated from deprecated

yarikoptic:
If we decide to proceed forward with smth like that

- [ ] provide `@deprecated` support for `Parameter`
- [ ] add `sphinx.versionadded` and `sphinx.versionchanged` from deprecated


yarikoptic:
Nobody got excited, let's postpone

PR #5345, opened 2021-01-20
BF+RF: Consistent and more "thorough" handling of fake HOMEs (including Windows)

yarikoptic:
Windows brings us many joys, including also having not just USERPROFILE but also HOMEDRIVE and HOMEPATH, so resetting USERPROFILE itself is not totally kosher alone.  Also, we could have PATH (e.g. in git shell) defined too even on Windows, so it is not sufficient to just reset USERPROFILE.

So in the first commit I have added a function to set only the variables known to current os.environ, and use of that one instead of hardcoded use of USERPROFILE vs HOME in test_clone seems to make that test pass for me for both py 3.7 and 3.8 on windows.  If confirmed that other Windows gurus are happy, then Closes #5338 (merge conflict will be minimal when merging into master for that test, didn't check overall though).   I have tuned up a few other tests where I found `HOME` to be used without paying due respect to glorious Windows land, so hopefully it would make it all a bit more consistent.

While at it I also made handling of memoized env variables in setup/teardown more consistent and concise. It is a separate commit, so if any objections -- could probably later do it against master

PR #5347, opened 2021-01-20
TEMP RM travis and github to see if appveyor is ok on maint

yarikoptic:
To see if changes in https://github.com/datalad/datalad/pull/5345
broke appveyor or it is just life


yarikoptic:
so, appveyor is happy in current maint state, khm

PR #5352, opened 2021-01-20
TEMP: again attempting to fix coverage from appveyor

yarikoptic:


yarikoptic:
observation, still no coverage report even after I added Ubuntu, so it is not OS but new appveyor setup specific.

yarikoptic:
out of immediate ideas, so asking codecov: https://codecov.freshdesk.com/support/tickets/4173 (not sure if public, so just FTR)

yarikoptic:
ran out of time/motivation

PR #5354, opened 2021-01-21
BF: All params in ORA's response to LISTCONFIGS

bpoldrack:
Used to be optional, but recent git-annex will not accept parameters to
initremote that are not listed in its response to LISTCONFIGS if there
was a response to begin with.

### DIFF
--- a/datalad/distributed/ora_remote.py
+++ b/datalad/distributed/ora_remote.py
@@ -631,6 +631,8 @@ class RIARemote(SpecialRemote):
         if hasattr(self, 'configs'):
             # introduced in annexremote 1.4.2 to support LISTCONFIGS
             self.configs['url'] = "RIA store to use"
+            self.configs['archive-id'] = "Dataset ID. Should be set " \
+                                         "automatically by datalad"
         # machine to SSH-log-in to access/store the data
         # subclass must set this
         self.storage_host = None


kyleam:
> Used to be optional, but recent git-annex will not accept parameters to
> initremote that are not listed in its response to LISTCONFIGS if there
> was a response to begin with.

I think that behavior (initremote checking the input based on LISTCONFIGS out, if any, and aborting with "Unexpected parameters: ...") came with the addition of LISTCONFIGS itself in 7.20200202.7.

Fwiw the proposed change looks unproblematic to me, though I feel like I'm missing some wider context because I don't see where in the code base we pass archive-id to initremote:

```
$ git grep archive-id
datalad/distributed/ora_remote.py:        self.archive_id = self.annex.getconfig('archive-id')
datalad/distributed/ora_remote.py:        self.annex.setconfig('archive-id', self.archive_id)
datalad/distributed/tests/test_ora_http.py:    assert_in("archive-id={}".format(ds.id), remote_log)
datalad/distributed/tests/test_ria_basics.py:    assert_in("archive-id={}".format(ds.id), remote_log)
datalad/distributed/tests/test_ria_basics.py:        assert_in("archive-id={}".format(ds.id), remote_log)
datalad/distributed/tests/test_ria_basics.py:    assert_in("archive-id={}".format(ds.id), remote_log)
```

bpoldrack:
> Fwiw the proposed change looks unproblematic to me, though I feel like I'm missing some wider context because I don't see where in the code base we pass archive-id to initremote

We don't. `annexremote.SpecialRemote` does based on that `config` property.

PR #5356, opened 2021-01-21
prep-travis-forssh.sh: Fail nc loop after 60 tests

### LINKED ISSUE #5326: prep-travis-forssh.sh: timeout (e.g. 30 sec) instead of sleep waiting forever
we are having an issue probably relating to having multiple docker instances on the same VM instance: https://github.com/datalad/git-annex/issues/42  which leads to very lengthy sleep, fill up of the logs, and eventual time out.

loop at https://github.com/datalad/datalad/blob/master/tools/ci/prep-travis-forssh.sh#L38  should be adjusted to exit 1 upon reaching e.g. 60 loop cycles

jwodder:
Closes #5326.

### DIFF
--- a/tools/ci/prep-travis-forssh.sh
+++ b/tools/ci/prep-travis-forssh.sh
@@ -35,8 +35,16 @@ curl -fSsL \
      >setup-docker-ssh
 sh setup-docker-ssh --key=/tmp/dl-test-ssh-id.pub -2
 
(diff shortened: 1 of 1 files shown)


yarikoptic:
Thanks! I think we need it in `maint`, so I will cherry-pick to it and merge to master locally, and push... done

PR #5358, opened 2021-01-22
BF: Flooding INFO log with messages non-consequential to users

### LINKED ISSUE #5349: search: flooding INFO log with non-consequential to user messages
```
(git-annex)lena:/tmp/datasets.datalad.org[master]
$> datalad search haxby
                                      [INFO   ] Falling back to standard indexer for metadata format: @context                                                                                                                                                     
[INFO   ] Falling back to standard indexer for metadata format: datalad_core                                                                                                                                                                                       
[INFO   ] Falling back to standard indexer for metadata format: frictionless_datapackage 
[INFO   ] Falling back to standard indexer for metadata format: @context 
[INFO   ] Falling back to standard indexer for metadata format: datalad_core 
[INFO   ] Falling back to standard indexer for metadata format: @context 
...
```
and there are 100s of them... IMHO they should go to DEBUG level and probably adjusted to inform what dataset they are talking about

christian-monch:
Fixes #5349

Missing indexers are now reported on DEBUG level instead of INFO level. In addition,
each missing indexer is reported only once.

To support reduced reporting without changing the existing signatures too much and
without carrying unrelated objects into functions, the InstanceStorage class was created.
It provides singleton storage objects, addressed by their purpose, i.e. an entry in an
Enum, that exist for the duration of the datalad invocation, i.e. as long as the
current python process is executed.

### DIFF
--- a/datalad/metadata/search.py
+++ b/datalad/metadata/search.py
@@ -203,7 +203,9 @@ def _meta2autofield_dict(meta, val2str=True, schema=None, consider_ucn=True):
                             indexer.name,
                             str(indexer.dist),
                             exc_str(e))
-        lgr.info('Falling back to standard indexer for metadata format: %s', metadata_format_name)
+        lgr.debug(
+            'Falling back to standard indexer for metadata format: %s',
+            metadata_format_name)
         return lambda metadata: _deep_kv('', metadata)
 
     if val2str:


yarikoptic:
I think to fix the issue, dropping to `.debug` level should be sufficient.  If we were to introduce "log once", we should just work it out at the level of the logger one way or another (ideally so we could just do `lgr.warn("msg", *args, log_once=True)` -- I feel a bit lukewarm about the proposed new construct which isn't really convenient (and thus unlikely to become used elsewhere) in current form.

christian-monch:
I removed the InstanceStore commits.

Setting up the logger to ignore repeated lines is possible, but I am not sure about the best solution yet. In the interest of not wasting time, let us keep it to the DEBUG-level commit and open another issue (which could be assigned to me ;-) if we believe that repeated messages should be filtered.

Issue #5361, opened 2021-01-22
Support 'git-annex config' in ConfigManager/`configuration`


mih:
As means to supply branch-independent config across clones.

Simultaneously, we should rename `--scope dataset` to `--scope branch` in `configuration` immediately. And later (or even at the same time) introduce the `git-annex` configuration as the actual scope `dataset` (or for authenticity reasons maybe indeed as `git-annex`, which would avoid future confusion.)

bpoldrack:
Despite the potential confusion with the former notion of `dataset`, I'm pretty much in favor of `branch`+`dataset`. While `git-annex` is technically correct, it's less telling if you didn't grasp a lot about how git-annex works yet. If you're using datalad looking at it as a kind of abstraction layer, you can do a lot w/o ever getting what that `git-annex` branch is about and what storing something there would imply. I don't think there's a whole lot of multi-branch usage of datalad datasets with branch specific configs out there yet.

An alternative could be something like `branch`+`all-branches`.

kyleam:
My knee-jerk reaction is that this may not be worth exposing.  DataLad's config situation is already pretty complicated to reason about (e.g., gh-5217), and this adds another layer for readers/users to digest.  On top of that, there's the complication that git-annex won't look at the git-annex branch for most options, only these:
  
  * `annex.addunlocked`
  * `annex.autocommit`
  * `annex.dotfiles`
  * `annex.largefiles`
  * `annex.resolvemerge`
  * `annex.securehashesonly`
  * `annex.synccontent`
  * `annex.synconlyannex`

Grepping around, I wasn't able to find any spots in the code base where we need inspect the git-annex branch for these.

mih:
I confirmed that git-annex prevents setting any non-whitelisted configuration. This makes it pointless to support this mechanism. Thx @kyleam

bpoldrack:
I'm not entirely convinced. This only makes the point to not expose the functionality of `git-annex-config`. Agree on that aspect. But: At a quick test it doesn't seem like annex is confused by adding non-whitelisted configs to `git-annex:config.log`. So, if we are interested in branch-independent configs it still is a place we could use instead of coming up with yet another mechanism. We would just not use `git--annex-config` to read/write those configs.

PR #5363, opened 2021-01-22
RF: configuration --scope dataset -> --scope branch

mih:
Following the spirit of gh-5361, this change renames the 'dataset'
scope to 'branch', because with a future introduction of other
"dataset" scopes (e.g. git-annex based persistent configuration),
such a generic name will be a problem.

'branch' is a compact and more accurate label.

This changes does not alter labeling in the ConfigManager yet, but
focuses on the about-to-be introduced API addition. Further changes
have less urgency.

### DIFF
--- a/datalad/local/configuration.py
+++ b/datalad/local/configuration.py
@@ -69,20 +69,21 @@ class Configuration(Interface):
     Query and modification of three distinct configuration scopes is
     supported:
 
-    - 'dataset': the persistent configuration in .datalad/config of a dataset
+    - 'branch': the persistent configuration in .datalad/config of a dataset
+      branch
     - 'local': a dataset clone's Git repository configuration in .git/config
     - 'global': non-dataset-specific configuration (usually in $USER/.gitconfig)
 
-    Modifications of the persistent 'dataset' configuration will not be saved
+    Modifications of the persistent 'branch' configuration will not be saved
     by this command, but have to be committed with a subsequent `save`
     call.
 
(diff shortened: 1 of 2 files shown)


yarikoptic:
IMHO scope should be aligned with the scopes of .config to avoid confusion etc. For git-annex config logical scope is git-annex. So overall I don't see a benefit from the change proposed in this PR.

mih:
> IMHO scope should be aligned with the scopes of .config to avoid confusion etc. For git-annex config logical scope is git-annex. So overall I don't see a benefit from the change proposed in this PR.

I was afraid you would say that. Then I would propose to delay the release of the `configuration` command until everything is aligned.

Even though the `git-annex` scope doesn't even make sense to support (https://github.com/datalad/datalad/issues/5361#issuecomment-765709327), because git-annex actively prevents this mechanism from being used for arbitrary settings. The need for a branch-independent `dataset`-scope configuration is not new. For example the `ukbiobank` extension dances around this absent feature quite a bit.

I do not see why I would want to introduce a new command with a parameterization that will be in the way sooner than later. Let's remove it from master.

yarikoptic:
> The need for a branch-independent dataset-scope configuration is not new. For example the ukbiobank extension dances around this absent feature quite a bit.

Could be dataset-any-branch, dataset-global, any-branch.
I know that it is suboptimal, but I value consistency (through time) and alignment more.

mih:
Superseded by #5370

PR #5367, opened 2021-01-23
BF: in a new process, test first if we can run any command, and if not - old ways

### LINKED ISSUE #5309: FAIL: datalad.support.tests.test_parallel.test_gracefull_death

### LINKED ISSUE #5362: FOI: wtf blow on windows / master / appveyor / python 3.7
since we unofficially do not care about 3.7 for windows, not considering it a bug but wanted to record:  while trying to figure out new appveyor + codecov gotcha in #5352 and temporarily adding 3.7 on windows got a failed run due to

```
datalad wtf
  File "C:\Python37-x64\Scripts\datalad-script.py", line 33, in <module>
    sys.exit(load_entry_point('datalad', 'console_scripts', 'datalad')())
  File "C:\Python37-x64\Scripts\datalad-script.py", line 25, in importlib_load_entry_point
    return next(matches).load()
  File "c:\python37-x64\lib\site-packages\importlib_metadata\__init__.py", line 96, in load
    module = import_module(match.group('module'))
  File "c:\python37-x64\lib\importlib\__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1006, in _gcd_import
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 953, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1006, in _gcd_import
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 953, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1006, in _gcd_import
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 967, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 677, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 728, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "c:\projects\datalad\datalad\__init__.py", line 47, in <module>
    cfg = ConfigManager()
  File "c:\projects\datalad\datalad\config.py", line 274, in __init__
    self.reload(force=True)
  File "c:\projects\datalad\datalad\config.py", line 327, in reload
    self._stores[store_id] = self._reload(runargs)
  File "c:\projects\datalad\datalad\config.py", line 359, in _reload
    encoding='utf-8',
  File "c:\projects\datalad\datalad\config.py", line 717, in _run
    out = self._runner.run(self._config_cmd + args, **kwargs)
  File "c:\projects\datalad\datalad\cmd.py", line 390, in run
    env=env,
  File "c:\python37-x64\lib\asyncio\base_events.py", line 587, in run_until_complete
    return future.result()
  File "c:\projects\datalad\datalad\cmd.py", line 103, in run_async_cmd
    transport, protocol = await proc
  File "c:\python37-x64\lib\asyncio\base_events.py", line 1544, in subprocess_exec
    bufsize, **kwargs)
  File "c:\python37-x64\lib\asyncio\base_events.py", line 462, in _make_subprocess_transport
    raise NotImplementedError
NotImplementedError
Command exited with code 1
```
which could also be due to very recent change #5350 (heh)

yarikoptic:
To overcome a problem with unruly super processes, ref: https://github.com/ReproNim/testkraken/issues/95 may be just on OSX.

We will track if we are trying to run in a new PID, and if new - we will first test if we can run a command in the default loop, if not -- we revert to "old ways" for that PID to start loop for each process

@djarecka please check if works for you.  Tried on OSX with 3.7  on testkraken -- seems to be "good as old" ;)

- [x] ideally needs a test

Also adds a known failure for parallel on osx, closes #5309

Checked on a local VM, it also seems to Closes #5362 (although I am not yet 100% sure why/how since we are not catching NotImplementedError ;))

### DIFF
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -273,7 +273,7 @@ class WitlessRunner(object):
     # new loop instantiation in a child process
     # https://bugs.python.org/issue21998
     _loop_pid = None
-    _loop_need_new = False
+    _loop_need_new = None
 
     def __init__(self, cwd=None, env=None):
         """
@@ -365,51 +365,16 @@ class WitlessRunner(object):
         # with our own event loop management
         # this is how ipython does it
         try:
-            pid = os.getpid()
-            is_new_proc = WitlessRunner._loop_pid is None or WitlessRunner._loop_pid != pid
-            if WitlessRunner._loop_need_new and not is_new_proc:
-                raise RuntimeError("we know we need a new loop")
+            is_new_proc = self._check_if_new_proc()
             event_loop = asyncio.get_event_loop()
             if is_new_proc:
-                WitlessRunner._loop_pid = pid
-                # We need to check if we can any command
-                try:
-                    event_loop.run_until_complete(
-                        run_async_cmd(
-                            event_loop,
-                            [sys.executable, "--version"],
-                            protocol,
-                            stdin,
-                            protocol_kwargs=kwargs,
-                            cwd=cwd,
-                            env=env,
-                        )
-                    )
-                    WitlessRunner._loop_need_new = False
-                except OSError as e:
-                    # due to https://bugs.python.org/issue21998
-                    # exhibits in https://github.com/ReproNim/testkraken/issues/95
-                    lgr.debug("It seems we need a new loop when running our commands: %s", exc_str(e))
-                    WitlessRunner._loop_need_new = True
-                    raise RuntimeError("the loop is not reusable")
+                self._check_if_loop_usable(event_loop, stdin)
             if event_loop.is_closed():
                 raise RuntimeError("the loop was closed - use our own")
             new_loop = False
         except RuntimeError:
+            event_loop = self._get_new_event_loop()
             new_loop = True
-            # start a new event loop, which we will close again further down
-            # if this is not done events like this will occur
-            #   BlockingIOError: [Errno 11] Resource temporarily unavailable
-            #   Exception ignored when trying to write to the signal wakeup fd:
-            # It is unclear to me why it happens when reusing an event looped
-            # that it stopped from time to time, but starting fresh and doing
-            # a full termination seems to address the issue
-            if sys.platform == "win32":
-                # use special event loop that supports subprocesses on windows
-                event_loop = asyncio.ProactorEventLoop()
-            else:
-                event_loop = asyncio.SelectorEventLoop()
-            asyncio.set_event_loop(event_loop)
         try:
             # include the subprocess manager in the asyncio event loop
             results = event_loop.run_until_complete(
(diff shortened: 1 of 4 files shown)


djarecka:
works great :) Thank you!

yarikoptic:
Ftr: no notable effect on benchmarks run

yarikoptic:
dedicated test added in 3e403fbd512fcb32b8c7310b7607aa9b1c2180ff

yarikoptic:
Sleeping more grief but mac consistently fails one parallel test, seemingly unrelated. Will need to handle it here as well

yarikoptic:
As it seems to resolve also #5362 marked it for 0.14.0

bpoldrack:
TBH, I'm quite confused about the design of this. If we look at `_check_if_new_proc` for example:

It says it's about figuring out whether or not we are in a new process. But then it raises an exception that is not about something like "I couldn't figure whether to respond with `True` or `False`", but instead is there to communicate that we found something other than the answer to the question and this other thing is also interesting to the intended caller of the function. So, the answer would validly be `False`, but screw that - we raise. ;-)

So - not sure. Weird to me and I see my future self massively struggling with making meaningful changes to this if it turns out to be needed.

datalad/cmd.py
```diff
@@ -418,6 +416,72 @@ def run(self, cmd, protocol=None, stdin=None, cwd=None, env=None, **kwargs):
...
+          in this pid
+        """
+        pid = os.getpid()
+        is_new_proc = cls._loop_pid is None or cls._loop_pid != pid
+        if is_new_proc:
+            # We need to check if we can run any command smoothly
+            cls._loop_pid = pid
+            cls._loop_need_new = None
+        elif cls._loop_need_new:
+            raise RuntimeError("we know we need a new loop")
```
mih, line 436: Not sure when this would happen, but maybe the error needs to be specific about which loop it is talking about and why this knowledge must be expressed as a runtimeerror.

datalad/cmd.py
```diff
@@ -418,6 +416,72 @@ def run(self, cmd, protocol=None, stdin=None, cwd=None, env=None, **kwargs):
...
+                    stdin,
+                )
+            )
+            cls._loop_need_new = False
+        except OSError as e:
+            # due to https://bugs.python.org/issue21998
+            # exhibits in https://github.com/ReproNim/testkraken/issues/95
+            lgr.debug("It seems we need a new loop when running our commands: %s", exc_str(e))
+            cls._loop_need_new = True
+            raise RuntimeError("the loop is not reusable")
```
mih, line 466: Same spirit as the comment above: Why is this a runtime error? Why not a new loop?

mih (commented):
I left comments re the nature of the exceptions raised, but otherwise I view this as "if it works, why not?" My own explorations of this topic have not been particularly informative or even effective ;-)

yarikoptic:
> but screw that - we raise. ;-)

it is just coding paradigm (like  `StopIteration` for the loops which could come from break or just finishing iterating): raise there is just to fit the try/except paradigm above which is also to catch exceptions which might be raised from `get_event_loop`.  So it is pretty much a "`False` with a message" so while debugging you can quickly assess in `except` why we start to mint our own loops (instead of going line by line until one of them evaluates to `False`).

datalad/cmd.py
yarikoptic: it is because of outside `try/except RuntimeError`.  This message will not be seen anywhere in outside code, so notion of a `loop` is clear as to me in this limited context.

datalad/cmd.py
yarikoptic: see above, and it will be a new loop since we are raising an exception here.

datalad/cmd.py
bpoldrack: Well, that's the design I struggle with (although it is of course "valid if it works"). The function implements assumptions about how exactly and for what purpose it is called from within what context. Why not just return `True`/`False` and check for `_loop_need_new` outside of it?

yarikoptic:
just for clarity, since I think diff is making it harder to grasp, here is the code on deciding either to start a new loop or not

```python
        try:
            is_new_proc = self._check_if_new_proc()
            event_loop = asyncio.get_event_loop()
            if is_new_proc:
                self._check_if_loop_usable(event_loop, stdin)
            if event_loop.is_closed():
                raise RuntimeError("the loop was closed - use our own")
            new_loop = False
        except RuntimeError:
            event_loop = self._get_new_event_loop()
            new_loop = True
```

unrolling it into a sequence of `if`s (if I RFed those `_check` to return bool instead of rising) and try/excepts could have looked like

```python
        new_loop = False
        is_new_proc = self._check_if_new_proc()
        try:
            event_loop = asyncio.get_event_loop()
           if is_new_proc:
                if not self._check_if_loop_usable(event_loop, stdin):
                     new_loop = True
           if event_loop.is_closed():
              new_loop = True
        except RuntimeError:
            new_loop = True
        if new_loop:
            event_loop = self._get_new_event_loop()
```
and not sure that it would have made it more readable etc

bpoldrack:
> So it is pretty much a "False with a message"

Exactly. It's actually about returning three different values, but we use the return value for two of them and an exception for the third one. Why not return 1 of three values then instead of a boolean?

It's not the same as `StopIteration`. That would be raised if you can't validly iterate anymore. That's what an exception is for: Can't return a regular response for reason x. In opposition to that, here we can still validly tell whether the "answer" is `True` or `False` but raise nevertheless.

kyleam:
Verdict on the call was general approval and that we'd wait on Ben's fixups.  Those have come, and I think are unlikely to be a point of contention... merging.

PR #5368, opened 2021-01-23
More pains/trials with appveyor coverage submission troubleshooting

yarikoptic:
paralelizing with #5352


yarikoptic:
ran out of time/motivation

PR #4965, opened 2020-09-25
Don't move .coverage out of __testhome__

jwodder:
Fixes #4956.

Issue #4956, opened 2020-09-24
Bring back combining coverage reports from special remotes


yarikoptic:
Discovered in the course of https://github.com/datalad/datalad/pull/4955/files that we fail to merge coverage reports from special remotes:  https://travis-ci.org/github/datalad/datalad/jobs/729780153#L2438
```
6.32s$ if [ -z "${_DL_CRON:-}" ]; then cp __testhome__/.coverage .; coverage combine -a /tmp/.coverage-entrypoints-*; codecov; fi
Traceback (most recent call last):
  File "/home/travis/virtualenv/python3.7.1/bin/coverage", line 8, in <module>
    sys.exit(main())
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/coverage/cmdline.py", line 865, in main
    status = CoverageScript().command_line(argv)
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/coverage/cmdline.py", line 588, in command_line
    self.coverage.combine(data_dirs, strict=True)
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/coverage/control.py", line 696, in combine
    combine_parallel_data(self._data, aliases=aliases, data_paths=data_paths, strict=strict)
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/coverage/data.py", line 117, in combine_parallel_data
    data.update(new_data, aliases=aliases)
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/coverage/sqldata.py", line 706, in update
    for (file, context), numbits in lines.items()
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/coverage/sqldata.py", line 706, in <listcomp>
    for (file, context), numbits in lines.items()
KeyError: '/home/travis/build/datalad/datalad/datalad/api.py'
```

so that step fails and we just submit main coverage.

@jwodder - since you dealt with coverage reporting recently, could you please also have a look at this, more obscure case as well?  I have not yet tried to reproduce locally etc.  Look into .travis.yml for more pointers on how we do it (helper scripts etc)

jwodder:
@yarikoptic What creates the `/tmp/.coverage-entrypoints-*` files?  They're not appearing in my initial tests to reproduce the issue.

yarikoptic:
[...]

PR #5371, opened 2021-01-25
BF: make ORA aware of sameas

### LINKED ISSUE #4259: git-annex no longer provides remote name with GETCONFIG name

bpoldrack:
Special remotes being set up via --sameas don't have a 'name' config to
query but only a 'sameas-name'.

(Closes #4259)

### DIFF
--- a/datalad/distributed/ora_remote.py
+++ b/datalad/distributed/ora_remote.py
@@ -777,6 +777,8 @@ class RIARemote(SpecialRemote):
         # try loading all needed info from (git) config
         name = self.annex.getconfig('name')
         if not name:
+            name = self.annex.getconfig('sameas-name')
+        if not name:
             raise RIARemoteError(
                 "Cannot determine special remote name, got: {}".format(
                     repr(name)))
--- a/datalad/distributed/tests/test_ria_basics.py
+++ b/datalad/distributed/tests/test_ria_basics.py
@@ -145,6 +145,17 @@ def _test_initremote_basic(host, ds_path, store, link):
         [assert_in(c, remote_log) for c in common_init_opts]
         assert_in("archive-id={}".format(ds.id), remote_log)
 
+    # we can deal with --sameas, which leads to a special remote not having a
+    # 'name' property, but only a 'sameas-name'. See gh-4259
+    try:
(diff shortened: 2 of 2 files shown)


mih (changes requested):
Makes sense. But the coverage report seems to indicate that the test never actually engages `--sameas`, hence we do not know whether the fix works given that the tests pass.

Or am I reading it wrong?

PS: That wanted to be a "comment", not necessarily a "request changes".

bpoldrack:
> But the coverage report seems to indicate that the test never actually engages --sameas, hence we do not know whether the fix works given that the tests pass.
> Or am I reading it wrong?

Yes and no. The test call is executed with `--sameas` passed to `initremote`, hence the `try ... except` (introduced after I saw that failure). But if all the annex versions we test against that contribute to coverage report are too old and don't know the `--sameas` option, those calls will fail before the special remote is executed. Hence the coverage report might be technically correct, too.

bpoldrack:
Having a closer look, there might be more to it ... putting on hold

kyleam:
> It doesn't look like annex is that old everywhere, leaving me with the assumption, that coverage report isn't accurate unless someone can point out how exactly I screwed up.

Running with 8.20201129 locally, everything looks good to me:

- The new test fails with the error from gh-4259 without the code change and passes with it.
- Tweaking the new code (`print`, `assert 0`, etc) verifies that it's being executed.

My guess is that the lack of coverage has something to do with it being a separate process (e.g. gh-4956, though that should have been resolved on Travis by gh-4965).

kyleam (approved):

PR #5375, opened 2021-01-26
download-latest-artifact: Match "Location:" header case-insensitively

jwodder:
See https://github.com/datalad/datalad-extensions/pull/77 for more information.

I'm not actually sure if the script in this repository is actually used, but just in case...


bpoldrack:
Seems indisputable. Thanks, @jwodder!

PR #5377, opened 2021-01-27
BF(+skip): r- path strings, skip test_asyncio_loop_noninterference1 on windows for python < 3.8

yarikoptic:
First it was failing because paths strings on windows contain \ so I had
to announce those strings as r(aw).

But then test still fails since apparently on windows (with 3.7, not
3.8) there is no event loop for some reason, and get_event_loop() would
just raise RuntimeError.  Since "tuning" the reproducer might actually
change the use case we are trying to test (which works on other
environements), I decided to just announce it a skip.

First observed during datalad/git-annex  repo daily tests against master (this should be the last one -- otherwise all clear!)

### DIFF
--- a/datalad/tests/test_witless_runner.py
+++ b/datalad/tests/test_witless_runner.py
@@ -165,6 +165,10 @@ def test_runner_parametrized_protocol():
 @with_tempfile(mkdir=True)
 @with_tempfile()
 def test_asyncio_loop_noninterference1(path1, path2):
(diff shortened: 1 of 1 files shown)


kyleam (approved):

kyleam:
Hmm, a macOS build on appveyor failed, but there is no log (either shown on the web or via download).  I'll restart it.

https://ci.appveyor.com/project/mih/datalad/builds/37473287/job/p2uthmr5tie7fi3i

Issue #5383, opened 2021-01-28
Stop using .gitattributes for annex.largefiles config


mih:
A `create` leaves the following config in the default branch:

```
% cat .gitattributes 
* annex.backend=MD5E
**/.git* annex.largefiles=nothing%                                                         

% cat .datalad/.gitattributes 
config annex.largefiles=nothing
metadata/aggregate* annex.largefiles=nothing
metadata/objects/** annex.largefiles=(anything)%                                           
```

The fact that this config is branch-specific is generally a disadvantage for datalad.

I propose to stop using `.gitattributes` for this and switch to `git annex config`. This will yield a subjectively cleaner dataset on create, translates to all branches, and causes less pain when additional configuration is done by users (or dataset procedures) that need to take a datalad-created starting point into account.

Unfortunately, there is no way to replace `annex.backend` settings in this way.

yarikoptic:
> The fact that this config is branch-specific is generally a disadvantage for datalad.

#5390 is going forward to achieve this, but I kept thinking if that is what we want really, and either it is possible of an advantage.  I do see benefit for some use cases (and possible tiny OPT for `create`) to have such configuration to be just global and not require re-merging `master` branch where the `.gitattributes` setting was adjusted.  Also, I had to do that a good number of times while preparing openfmri and other datasets for datasets.datalad.org, so it is not a new "procedure" for me.

BUT, unlike files availability and special remotes configuration, I think that keeping such configuration within git-annex's branch config could lead to surprises without clear idea on why and when a dataset started to behave differently all of a sudden after you do `datalad update` and thus merge changes into git-annex branch for this setting (possibly from some semi-trustful remote where malicious user set largefiles to nothing, and you kept your passwords in the annexed file)... Since git-annex branch has no relationship to any other branch (besides temporal sorting by the date), it would make it hard and eventually impossible (after `annex forget`) to figure out the "provenance" of points in time where datalad starts to behave differently.

So, in general, I would actually prefer to keep largefiles configuration visible in .gitattributes and for it to be less "viral" as it would happen within `git annex config`. In a typical dataset, even if it gets branches, datalad operations would remain more consistent.  But I do see for some use cases an advantage of having it "global". So it might be up to the user to decide which way he/she would like such configuration to be handled.  

NB `.gitattributes` overrides `git-annex config` setting, so my desire of keeping it *as is* could be achieved with some additional command/change, but I do not see why it should not stay default.

[...]

PR #5390, opened 2021-01-29
Stop using .gitattributes for annex.largefiles config

mih:
Only the test commit is concerned with the change that motivated this PR (incomplete still, but I had to stop for now -- still not fully comprehending the necessary expression). The rest are minor, but nevertheless useful RFs.

ATM (at least), it seems that this approach is also slightly faster than the `.gitattributes` creation and addition to Git (~4-5% of the total `create` runtime on my system)

TODO:
- [ ] Address `datalad.metadata.create-aggregate-annex-limit`

Ref #5383

### DIFF
--- a/datalad/core/local/create.py
+++ b/datalad/core/local/create.py
@@ -377,70 +377,11 @@ class Create(Interface):
 
         # create and configure desired repository
         if no_annex:
-            lgr.info("Creating a new git repo at %s", tbds.path)
-            tbrepo = GitRepo(
-                tbds.path,
-                create=True,
-                create_sanity_checks=False,
-                git_opts=initopts,
-                fake_dates=fake_dates)
-            # place a .noannex file to indicate annex to leave this repo alone
-            stamp_path = Path(tbrepo.path) / '.noannex'
-            stamp_path.touch()
-            add_to_git[stamp_path] = {
-                'type': 'file',
-                'state': 'untracked'}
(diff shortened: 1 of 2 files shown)


kyleam:
There are a few failures on the build with 7.20190819.  Testing `test_plugins:test_no_annex` locally (and assuming the other failures in that build are for the same reason), it starts to pass with 7.20191230, specifically 559162273:

    git-annex-config --set/--unset: No longer change the local git config setting
    
    e53070c1f quietly made it set the local git config too, but that was never
    documented anywhere, and it had surprising results. If I set
    annex.largefiles globally in a repo, I would expect to be able to change it
    in another repo, and the original repo would get the change and use it,
    rather than being stuck on the old value set there.
    
    And, if I have a local annex.largefiles and set a different global default,
    I'd be surprised to have my local setting overwritten.
    
    annex.securehashesonly does need to be set locally, since it's a security
    feature and the global is only a default until it gets set locally. So
    special cased.

So, the `git annex config` calls before 7.20191230 also set largefiles in .git/config, taking precedence over the .gitattributes settings.

https://travis-ci.com/github/datalad/datalad/jobs/478205111

yarikoptic:
@mih planning to push forward with this? I am still not sure if it is worth the effort (https://github.com/datalad/datalad/issues/5383#issuecomment-770108778).

mih:
Ok, closing. Salvaged pieces in #5523

PR #5398, opened 2021-02-02
RF: travis - jump to 8.20201129 of git-annex from conda from 8.20201007

yarikoptic:
It is with some hope that it might help addressing tests "stalling" issues


kyleam:
No luck: https://travis-ci.com/github/datalad/datalad/jobs/479645323#L1749

yarikoptic:
was of no help, so screw it

PR #5400, opened 2021-02-02
TEMP: to troubleshoot nfs stall

yarikoptic:


yarikoptic:
So it is a "known" but ignored by us (since failed to reproduce/figure out) issue observed during testing then recent git-annex in Aug 2020: https://github.com/datalad/datalad/pull/4804
```
datalad.support.tests.test_annexrepo.test_commit_annex_commit_changed(True,) ... ERROR
datalad.support.tests.test_annexrepo.test_commit_annex_commit_changed(False,) ... 
```
but we do not see what ERROR since killed.  Just FYI @joeyh ATM, hopefully we will pin it down more specifically :-/

yarikoptic:
no time, might pick up later

Issue #5379, opened 2021-01-27
datalad.tests.test_auto.test_proxying_open_testrepobased stalls while test building 0.14.0rc1 for debian


yarikoptic:
I thought it was still building (since yesterday) across releases but apparently it just stalled at

```
datalad.tests.test_auto.test_proxying_open_testrepobased ...
```

it isn't CPU busy etc... `ps` shows no child processes:

```
neurode+ 23455  0.0  0.0  19720 16988 pts/2    S+   Jan26   0:00                         /usr/bin/perl /usr/bin/dpkg-buildpackage -us -uc
neurode+ 23503  0.0  0.0   2696  2012 pts/2    S+   Jan26   0:00                           /usr/bin/make -f debian/rules build
neurode+ 23504  0.0  0.0  17632 15028 pts/2    S+   Jan26   0:00                             /usr/bin/perl /usr/bin/dh build --with python3 --buildsystem=pybuild
neurode+ 23538  0.0  0.0   2700  2020 pts/2    S+   Jan26   0:00                               /usr/bin/make -f debian/rules override_dh_auto_test
neurode+ 23539  0.0  0.0   2420   580 pts/2    S+   Jan26   0:00                                 /bin/sh -c PATH=/build/datalad-0.14.0~rc1/bin:$PATH \     dh_auto_test -- --test-nose
neurode+ 23540  0.0  0.0  20720 18168 pts/2    S+   Jan26   0:00                                   /usr/bin/perl /usr/bin/dh_auto_test -- --test-nose
neurode+ 23548  0.0  0.0  17056 13696 pts/2    S+   Jan26   0:00                                     /usr/bin/python3 /usr/bin/pybuild --test --test-nose -i python{version} -p 3.9 --test-nose
neurode+ 23557  0.0  0.0   2420   524 pts/2    S+   Jan26   0:00                                       /bin/sh -c cd '/build/datalad-0.14.0~rc1/.pybuild/cpython3_3.9_datalad/build'; python3.9 -m nose -v -s -v -e test_basics
neurode+ 23558  0.7  0.1 899136 137836 pts/2   Sl+  Jan26   8:08                                         python3.9 -m nose -v -s -v -e test_basics
```

edit 1: reran just in case "as is" -- the same, so not just a fluke

yet to troubleshoot... worse comes to worse -- might need a skip for debian package building so **even though I am adding 0.14.0 milestone, it should not hold the release**

mih:
[...]

PR #5404, opened 2021-02-03
BF: auto - do not proxy anything while trying to establish an AnnexRepo instance

yarikoptic:
Closes: #5379

Not yet sure how it was not triggered before or why not triggered with other pythons, so did not come up with some new dedicated test.

Also may be there is more levels of inception here and a single `bool` would not be enough (or I might somehow enter multiple times and thus better keep a counter or alike), but at least it resolve the hanging I replicated on my laptop.

### DIFF
--- a/datalad/auto.py
+++ b/datalad/auto.py
@@ -99,6 +99,7 @@ class AutomagicIO(object):
         self._patch = patch
         self._paths_cache = set() if check_once else None
         self._repos_cache = {} if check_once else None
+        self._getting_repo_instance = False
         if activate:
             self.activate()
 
@@ -125,6 +126,8 @@ class AutomagicIO(object):
         try:
             if self._in_open:
                 raise _EarlyExit("within open already")
+            if self._getting_repo_instance:
+                raise _EarlyExit("we are opening an annex repo - no files must be annexed")
             self._in_open = True  # just in case someone kept alias/assignment
             # return stock open for the duration of handling so that
             # logging etc could workout correctly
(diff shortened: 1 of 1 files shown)


yarikoptic:
restarted 2 travis jobs, one was failing to download from docker completely, and another

```
FAIL: datalad.downloaders.tests.test_docker_registry.test_download_docker_blob
  File "/tmp/dl-miniconda-2njdf908/lib/python3.8/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/tmp/dl-miniconda-2njdf908/lib/python3.8/site-packages/datalad/tests/utils.py", line 182, in _wrap_skip_if_no_network
    return func(*args, **kwargs)
  File "/tmp/dl-miniconda-2njdf908/lib/python3.8/site-packages/datalad/tests/utils.py", line 757, in _wrap_with_tempfile
  File "/tmp/dl-miniconda-2njdf908/lib/python3.8/site-packages/datalad/downloaders/tests/test_docker_registry.py", line 46, in test_download_docker_blob
    assert_in(DATALAD_SPECIAL_REMOTES_UUIDS[DATALAD_SPECIAL_REMOTE],
AssertionError: 'cf13d535-b47c-5df6-8590-0793cb08a90a' not found in ['a9bd62d5-29cc-4864-9900-f2d4db5f0b76']
```
both unrelated to this PR changes

datalad/auto.py
```diff
@@ -125,6 +126,8 @@ def _proxy_open_name_mode(self, origname, origfunc, *args, **kwargs):
         try:
             if self._in_open:
                 raise _EarlyExit("within open already")
+            if self._getting_repo_instance:
+                raise _EarlyExit("we are opening an annex repo - no files must be annexed")
```
kyleam, line 130: Yeah, still no luck figuring out how to reach this on my end, but these changes look safe to me (though I'm not really familiar with the inner workings of auto.py).

Thanks

kyleam (approved):

yarikoptic:
errors are due to docker (filed https://github.com/datalad/datalad/issues/5405) and that NFS stalling :-/  
thanks @kyleam for the review/feedback.  Just as an additional "test" I will add this patch to https://github.com/conda-forge/datalad-feedstock/pull/53 and see if it resolves the stall there . Note: there it stalls for all pythons, including 3.6 .

kyleam:
> just as an additional "test" I will add this patch to [conda-forge/datalad-feedstock#53](https://github.com/conda-forge/datalad-feedstock/pull/53) and see if it resolves the stall there

Ah, excellent.  And I see it already passed.

yarikoptic:
ok, lets proceed and i will also merge the feedstock

PR #5415, opened 2021-02-09
Fix RIA/ORA publication dependencies

bpoldrack:
Setting the publication dependency for 'origin' on an ORA remote in
postclone routine didn't properly account for multiple accessible ORA
remotes or the fact that none of them may be associated with the RIA
store we are cloning from.

Now, first check for an ORA remote with the same store URL registered
and if there is none check whether the store has a remote's uuid stored
in the .git/config of the dataset. If we have that remote enabled, set
the dependency on it.

### DIFF
--- a/datalad/core/distributed/clone.py
+++ b/datalad/core/distributed/clone.py
@@ -723,48 +723,17 @@ def postclonecfg_ria(ds, props):
     repo = ds.repo
     RIA_REMOTE_NAME = 'origin'  # don't hardcode everywhere
 
-    # chances are that if this dataset came from a RIA store, its subdatasets
-    # may live there too. Place a subdataset source candidate config that makes
-    # get probe this RIA store when obtaining subdatasets
-    ds.config.set(
-        # we use the label 'origin' for this candidate in order to not have to
-        # generate a complicated name from the actual source specification.
-        # we pick a cost of 200 to sort it before datalad's default candidates
-        # for non-RIA URLs, because they prioritize hierarchical layouts that
-        # cannot be found in a RIA store
-        'datalad.get.subdataset-source-candidate-200origin',
-        # use the entire original URL, up to the fragment + plus dataset ID
-        # placeholder, this should make things work with any store setup we
-        # support (paths, ports, ...)
-        props['source'].split('#', maxsplit=1)[0] + '#{id}',
-        where='local')
-
-    # setup publication dependency, if a corresponding special remote exists
-    # and was enabled (there could be RIA stores that actually only have repos)
-    # make this function be a generator
-    ora_remotes = [s for s in ds.siblings('query', result_renderer='disabled')
-                   if s.get('annex-externaltype') == 'ora']
-    if not ora_remotes and any(
-            r.get('externaltype') == 'ora'
-            for r in (repo.get_special_remotes().values()
-                      if hasattr(repo, 'get_special_remotes')
-                      else [])):
-        # no ORA remote autoenabled, but configuration known about at least one.
-        # Let's check origin's config for datalad.ora-remote.uuid as stored by
-        # create-sibling-ria and enable try enabling that one.
-        lgr.debug("Found no autoenabled ORA special remote. Trying to look it "
-                  "up in source config ...")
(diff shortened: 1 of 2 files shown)


mih (commented):
My superficial comment is re performance: Can it happen that we clone from store X, and we have an ORA URL pointing to the same X, but we would make a mistake when we configure a publication dependency. I will like the additional query for a UUID makes things slower, but not more reliable.

bpoldrack:
> I will like the additional query for a UUID makes things slower, but not more reliable.

I think you misread things (or I made a mistake). Intention is to only check the uuid if we have *no* URL match.

datalad/core/distributed/clone.py
```diff
@@ -847,19 +859,47 @@ def postclonecfg_ria(ds, props):
...
+            # No matches but we have successfully autoenabled ORA remotes. Could
+            # be the same store accessed by different method (cloning via HTTP
+            # but special remote access via SSH). We can confidently set
+            # publication dependency if the store knows the UUID.
+            org_uuid = get_uuid_from_store(props['giturl'])
+            uuid_matching_remotes = [r for r in ora_remotes
+                                     if r['annex-uuid'] == org_uuid]
+            if uuid_matching_remotes:
+                # we can't have multiple
+                assert len(uuid_matching_remotes) == 1
```
kyleam, line 888: Challenge accepted :)

<details>
<summary>assertion trigger</summary>

```sh
set -eu

cd "$(mktemp -d /tmp/dl-sameas-XXXXXXX)"
mkdir store
store="$(pwd)/store"

datalad create a
(
    cd a
    datalad create-sibling-ria -s s1 ria+ssh://datalad-test"$store"
    git annex initremote s1-store-local --sameas=s1-storage \
        type=external externaltype=ora autoenable=true \
        url=ria+file://"$store"
    datalad push --to=s1
)

dsid=$(git -C a config --file=.datalad/config datalad.dataset.id)
datalad clone ria+file://"$store#$dsid" b || :
(
    cd b
    echo
    datalad -f json siblings query | \
        jq 'select(."annex-externaltype" == "ora") |
            {name, "annex-externaltype", "annex-uuid"}'
    echo
    git config --get-regexp remote.s1
)
```

```
[...]
[ERROR] AssertionError() [clone.py:postclonecfg_ria:888] (AssertionError) 

{
  "name": "s1-store-local",
  "annex-externaltype": "ora",
  "annex-uuid": "beee8cb5-090a-4f4c-b121-e4342f89ea73"
}
{
  "name": "s1-storage",
  "annex-externaltype": "ora",
  "annex-uuid": "beee8cb5-090a-4f4c-b121-e4342f89ea73"
}

remote.s1-store-local.annex-externaltype ora
remote.s1-store-local.annex-uuid beee8cb5-090a-4f4c-b121-e4342f89ea73
remote.s1-store-local.annex-config-uuid a2e7a24e-8fe1-42d6-9358-6be571f813fc
remote.s1-store-local.annex-cost 100.0
remote.s1-store-local.annex-availability GloballyAvailable
remote.s1-storage.annex-externaltype ora
remote.s1-storage.annex-uuid beee8cb5-090a-4f4c-b121-e4342f89ea73
remote.s1-storage.annex-cost 200.0
remote.s1-storage.annex-availability GloballyAvailable
remote.s1-storage.annex-ignore false
remote.s1-storage.skipfetchall true
```

</details>

`--sameas` is probably pretty rare and even rarer with ORA special remotes, but it's still probably worth filtering out sameas remotes here.

bpoldrack: You win! I agree.

bpoldrack: Amend and force-pushed.

kyleam: Thanks.

The PR in general looks reasonable to me, though given how little I know about ria/ora, that doesn't count for much.

mih (approved):
Sorry for taking so long. I now had the chance to battle-test this in a problematic scenario, different from my original issue, and also there it works as advertised!

Code itself LGTM too.

Thx much!

Issue #3844, opened 2019-10-30
`save` doesn't complain when non-existing path are given as arguments


mih:
```
datalad create some
cd some
touch this
datalad save not
```

return with success and not messages, and there is nothing in debug logs either.

The reason is the behavior of `git ls-files`, which is silent for non-matching/non-existing paths.

If we want to fix this, we have to inspect the yielded results, and check if we have something for any input arg. This is similar to a problem in #3836

bpoldrack:
I think we should fix it indeed and yield a proper `impossible` result.

yarikoptic:
Just ran into this myself when forgot to add `-m`:
```shell
$> touch 123; datalad save -r "my message"
$> echo $?
0
$> datalad status
untracked: 123 (file)
```
I really do not think it is just a "UX" issue, it is a really unexpected behavior of succeeding when it should have failed.  My script might have relied on `save` and then `rm` command and I would have lost the data.  FWIW `datalad add` 0.12 fails nicely
```shell
$> datalad add nonexisting
[WARNING] path does not exist: /home/yoh/.tmp/dl-YKKbnGB/src/nonexisting [add(/home/yoh/.tmp/dl-YKKbnGB/src/nonexisting)] 
add(impossible): /home/yoh/.tmp/dl-YKKbnGB/src/nonexisting [path does not exist: /home/yoh/.tmp/dl-YKKbnGB/src/nonexisting]
```

yarikoptic:
added it to 0.13.0, but sure thing feel free to remove if you consider this not really release important.

yarikoptic:
related: https://git-annex.branchable.com/bugs/annex_metadata___40__not_--batch__39__ed__41___is_not_aware_of_files_added_via_addurls_--batch/

yarikoptic:
in particular last [comment](http://git-annex.branchable.com/bugs/annex_metadata___40__not_--batch__39__ed__41___is_not_aware_of_files_added_via_addurls_--batch/#comment-e2fb6309c32cfe75f870f07c809ae3c6)

Implemented annex.skipunknown git config, that will make it error out when given a file that git doesn't know about.

Not default yet, will be in a couple of years. complete annex.skipunknown transition in 2022

adswa:
The linked annex comment says "early 2022", so maybe this problem is soon gone...

adswa:
[...]

PR #5416, opened 2021-02-09
BF/RF: push - get diff depth-first not breadth-first

### LINKED ISSUE #5410: push: goes top down, need bottom up for web ui stats updates
In this brave new world decided to give a shot to push on my regular update of datasets.datalad.org, and spotted that
```shell
(git)smaug:/mnt/btrfs/datasets/datalad/crawl[master]git
$> datalad push '--since=^' --to=datalad-public -r --data auto  
publish(ok): . (dataset) [refs/heads/master->datalad-public:refs/heads/master 6bf9121..41e9bf7]                                                                                                                                                                    
publish(ok): abide2 (dataset) [refs/heads/master->datalad-public:refs/heads/master b181e7f..f7486d4]                                                                                                                                                               
publish(ok): abide2/RawData (dataset) [refs/heads/master->datalad-public:refs/heads/master 6c0ab6e..d595725]                                                                                                                                                       
publish(ok): abide2/RawData/BNI_1 (dataset) [refs/heads/git-annex->datalad-public:refs/heads/git-annex 6211913..2dab00c]    
...
```
it goes from super dataset down to subdatasets.  
I am yet to check if there is some reason for that, but I know that it would interfere with web-ui functioning in "pulling up" the file stats for web UI: it relies on super-dataset seeing already up-to-date subdatasets whenever its `post-update` hook is run -- the only way AFAIK to have file stats efficiently pulled up the tree upon updates.

So I wonder if it could be may be flipped to go bottom-up or if not, may be there should be an option (yet another thing to remember) or configuration item (but then would need to be set in all subdatasets, uff)

yarikoptic:
Remote repositories must exist for push already, and might have hooks enabled to process the "clean" state of the repository. In hooks we deploy with web-ui e.g. we "aggregate" information about sizes within submodules.  For that to work correctly, submodules should be pushed first, and not after their "superdatasets".
    
No reason/use case comes to mind why we might want to push super datasets first (besides that they are more "readily" analyzable, so `push` might start pushing sooner), and original commit 9796a5ff seems to not state specific choice for the breadth-first behavior.

Closes #5410

TODOs:
- [x] do full tests sweep (here in this PR) to spot more side-effects
- ~~fix up the test (or code) which I marked to be skipped to do the sweep~~ magically "Fixed" itself and @yarikoptic wonders if may be related to git-annex version boost
- [x] if decided to proceed, add a dedicated test to establish this order
- [x] figure out proper naming... [see below](https://github.com/datalad/datalad/pull/5416#pullrequestreview-624688045)

### DIFF
--- a/datalad/core/distributed/push.py
+++ b/datalad/core/distributed/push.py
@@ -317,7 +317,7 @@ def _datasets_since_(dataset, since, paths, recursive, recursion_limit):
             # TODO?: expose order as an option for diff and push
             # since in some cases breadth-first would be sufficient
             # and result in "taking action faster"
-            # reporting_order='breadth-first'
+            reporting_order='bottom-up'
     ):
         if res.get('action', None) != 'diff':
             # we don't care right now
--- a/datalad/core/distributed/push.py
+++ b/datalad/core/distributed/push.py
@@ -314,10 +314,11 @@ def _datasets_since_(dataset, since, paths, recursive, recursion_limit):
             recursion_limit=recursion_limit,
             # make it as fast as possible
             eval_file_type=False,
-            # we relay on all records of a dataset coming out
-            # in succession, with no interuption by records
-            # concerning subdataset content
-            reporting_order='breadth-first'):
+            # TODO?: expose order as an option for diff and push
+            # since in some cases breadth-first would be sufficient
+            # and result in "taking action faster"
+            # reporting_order='breadth-first'
+    ):
         if res.get('action', None) != 'diff':
             # we don't care right now
             continue
@@ -331,9 +332,9 @@ def _datasets_since_(dataset, since, paths, recursive, recursion_limit):
                 'Cannot handle diff result without a parent dataset '
                 'property: {}'.format(res))
         if res.get('type', None) == 'dataset':
-            # a subdataset record in another data
+            # a subdataset record in another dataset
             # this could be here, because
-            # - this dataset with explicitely requested by path
+            # - this dataset was explicitly requested by path
             #   -> should get a dedicated dataset record -- even without recursion
             # - a path within an existing subdataset was given
             # - a path within an non-existing subdataset was given
@@ -386,7 +387,7 @@ def _push(dspath, content, target, data, force, jobs, res_kwargs, pbars,
 
     res_kwargs.update(type='dataset', path=dspath)
 
-    # content will be unique for every push (even on the some dataset)
+    # content will be unique for every push (even on the same dataset)
     pbar_id = 'push-{}-{}'.format(target, id(content))
     # register for final orderly take down
     pbars[pbar_id] = ds
--- a/datalad/core/distributed/tests/test_push.py
+++ b/datalad/core/distributed/tests/test_push.py
@@ -19,6 +19,7 @@ from datalad.support.exceptions import (
     InsufficientArgumentsError,
 )
 from datalad.tests.utils import (
+    assert_false,
     assert_in,
     assert_in_results,
     assert_not_in,
@@ -45,8 +46,9 @@ from datalad.tests.utils import (
     SkipTest,
 )
 from datalad.utils import (
-    chpwd,
     Path,
+    chpwd,
+    path_startswith,
 )
 from datalad.support.gitrepo import GitRepo
 from datalad.support.annexrepo import AnnexRepo
@@ -245,6 +247,28 @@ def test_push():
     yield check_push, True
 
 
+def check_datasets_order(res, order='depth-first'):
+    """Check that all type=dataset records not violating the expected order
+
+    it is somewhat weak test, i.e. records could be produced so we
+    do not detect that order is violated, e.g. a/b c/d would satisfy
+    either although they might be neither depth nor breadth wise.  But
+    this test would allow to catch obvious violations like a, a/b, a
+    """
+    prev = None
+    for r in res:
+        if r.get('type') != 'dataset':
+            continue
+        if prev and r['path'] != prev:
+            if order == 'depth-first':
+                assert_false(path_startswith(r['path'], prev))
+            elif order == 'breadth-first':
+                assert_false(path_startswith(prev, r['path']))
+            else:
+                raise ValueError(order)
+        prev = r['path']
+
+
 @slow  # 33sec on Yarik's laptop
 @with_tempfile
 @with_tempfile(mkdir=True)
@@ -272,6 +296,7 @@ def test_push_recursive(
     target_top = mk_push_target(top, 'target', dst_top, annex=True)
     # subdatasets have no remote yet, so recursive publishing should fail:
     res = top.push(to="target", recursive=True, on_failure='ignore')
+    check_datasets_order(res)
     assert_in_results(
         res, path=top.path, type='dataset',
         refspec=DEFAULT_REFSPEC,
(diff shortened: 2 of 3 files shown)


yarikoptic:
sweep results:

- mac fail is unrelated - connectivity issue "error: Failed to connect to datasets-tests.datalad.org port 80: "
- travis is still that dreaded NFS stall (I keep forgetting that we have not pinned it down and  arrogantly was praising how green we are today, shame! ;) )

and otherwise I do not see any side-effects, so unless I am stopped, I will try to tackle that failing test some time soon.

mih:
My immediate response would be that I cannot think of a strong reason to prefer one over the other. But I feel like it deserves some more thinking. The reason for giving `breadth-first` is likely that this code evaluating the `diff` records assumes this order (as hinted by `we rely on all records of a dataset coming out concerning subdataset content` (with even more typos fixed to communicate the intent better). The key would be this piece

```py
        if parentds != cur_ds:
            if ds_res:
                # we switch to another dataset, yield this one so outside
                # code can start processing immediately
                yield (cur_ds, ds_res)
```

So whenever the dataset changes (which I assume can now happen before the last report of the previous dataset was received) it will report a dataset to be pushed. As a consequence, I fear that we will now process one and the same dataset multiple times (with partial records). It is likely that the tests do not look for such a condition.

kyleam:
> so -- the first record is good since `gitshasum` is different from the `prev_gitshasum`. But then it is the same for '/home/yoh/datalad/openneuro/ds000001' ... shouldn't it be just diff's @mih?

Clean items are expected in the sense that `_diff_ds` uses `GitRepo.diffstatus`:

https://github.com/datalad/datalad/blob/b13ed99880a0070d173e6fead40b3483500b1a6f/datalad/support/gitrepo.py#L3610-L3613

Then `Diff.custom_result_renderer` filters them out from the standard reporting.

I'm not sure on the deeper question of why `.diffstatus` is used rather than `.diff`, though my guess is so that Python callers of the dataset-level command can have access to the full comparison.

mih:
I cannot comment on the implications of a switch from `.diffstatus()` to `.diff()`, but the principle motivation re reporting "clean" vs reporting nothing was/is that we had routine issues with deciding whether something evaluated "OK" or was not evaluated at all in the context of `git-annex` returns (still a major problem for `save` #3844). I generally switched to the report-clean paradigm in the -revolution times in order to improve on the guessing. Moreover, given that we need to come to a "clean" result anyways, it is relatively cheap to report it, and enable a wider set of use-cases for consumers of more than just a single piece of information. For me this leads to less duplication of very similar code pieces, and to an overall efficiency gain vs the previous model of having multiple more specialized methods to determine one single piece of information from a more complex pile (but haven to query twice, when two pieces from the pile are needed).

yarikoptic:
Appveyor unrelated. @mih please have a look again - I think I fixed underlying issue

kyleam (commented):
> RF: push - get diff depth-first not breadth-first

The rationale in the second commit for making `push` process deeper subdatasets first makes sense to me.  But...

> BF: diff._diff_ds - delay reporting within ds entries for depth-first
>
> To be truly depth-first there should be no records yielded for super-dataset
> before yielding all the records from the subdatasets. [...]

... in my opinion sticking with the "depth-first" terminology is confusing.

<details>
<summary>demo repo</summary>

```sh
set -eu

datalad create
datalad create -d . a
datalad create -d a a/b
datalad create -d . c
datalad save -r
touch a/b/foo
touch c/bar
```

</details>

On maint (2b3468ac7d), here's the depth-first order of `datalad diff`.

```
$ datalad -f json diff -r | jq -r '.path' | grep -v '.git'
/tmp/dl-j2jRQzN/.datalad/config
/tmp/dl-j2jRQzN/a
/tmp/dl-j2jRQzN/a/.datalad/config
/tmp/dl-j2jRQzN/a/b
/tmp/dl-j2jRQzN/a/b/foo
/tmp/dl-j2jRQzN/a/b/.datalad/config
/tmp/dl-j2jRQzN/c
/tmp/dl-j2jRQzN/c/bar
/tmp/dl-j2jRQzN/c/.datalad/config
```

I disagree with the commit message; that is a depth-first walk of the tree.  Child nodes are traversed before moving to a sister node.

This PR changes that to

```
$ datalad -f json diff -r | jq -r '.path' | grep -v '.git'
/tmp/dl-j2jRQzN/a/b/foo
/tmp/dl-j2jRQzN/a/b/.datalad/config
/tmp/dl-j2jRQzN/a/.datalad/config
/tmp/dl-j2jRQzN/a/b
/tmp/dl-j2jRQzN/c/bar
/tmp/dl-j2jRQzN/c/.datalad/config
/tmp/dl-j2jRQzN/.datalad/config
/tmp/dl-j2jRQzN/a
/tmp/dl-j2jRQzN/c
```

So, perhaps we should use the "bottom up" terminology (used in `datalad subdatasets`) for the behavior introduced by this PR?

yarikoptic:
GREAT analysis @kyleam, thank you!  Indeed, I make it into "bottom up", and it was a legit "depth-first".  I will introduce "bottom-up" to represent (and use for `push`) behavior I have introduced.

yarikoptic:
refactored, see the last commit (dc92bf85c61fce28bb7d995f399c4016a53b42e6 ATM) with examples from running @kyleam 's script in the commit message

kyleam:
> refactored, see the last commit (~[dc92bf8](https://github.com/datalad/datalad/commit/dc92bf85c61fce28bb7d995f399c4016a53b42e6)~ [13fbb07](https://github.com/datalad/datalad/commit/13fbb0731d4723969684bbe2f2fbe3dec6a437ff) ATM)

Thanks, looks very good to me.

It doesn't matter so much given that there's no exposed option, but just a note that `subdatasets` spells its value without a hyphen.

yarikoptic:
> It doesn't matter so much given that there's no exposed option, but just a note that `subdatasets` spells its value without a hyphen.

yeap, and there it is a flag (just a note) and it does have it hypened in a doc:
```
  --bottomup            whether to report subdatasets in bottom-up order along each branch in
                        the dataset tree, and not top-down.
```
so at some point (may be of some bigger RF) we might want to straighten it up and make it `--bottom-up` or even reflect on @mih's stated dislike of flags and convert to `--order top-down|bottom-up` for higher consistency and opening the road for other orders e.g. instead of current one:
```
$> datalad -f '{path}' subdatasets -r --bottomup
/home/yoh/.tmp/dl-WyTjJpB/a/b
/home/yoh/.tmp/dl-WyTjJpB/a
/home/yoh/.tmp/dl-WyTjJpB/c/a
/home/yoh/.tmp/dl-WyTjJpB/c
```
it could be 
```
/home/yoh/.tmp/dl-WyTjJpB/a/b
/home/yoh/.tmp/dl-WyTjJpB/c/a
/home/yoh/.tmp/dl-WyTjJpB/a
/home/yoh/.tmp/dl-WyTjJpB/c
```
for some kind of `depth-up` (with similar one for `depth-down`?) if any such need arises. 

> Thanks, looks very good to me.

if supported by an official "approval", I would merge it immediately ;)

yarikoptic:
I would also be ok to merge it into `master` instead of `maint` since it came out a bit more intrusive than originally planned, so feedback on this would be welcome too

yarikoptic:
eh, no approvals... I guess if no objections would be voiced (@mih?), I will merge in a day or so

mih:
Had no chance to look yet. Should be able to do so today. Sorry

mih (approved):
Sorry that it took so long, and thanks for the detailed discussions. The changes make sense to me and rational for doing them seems sound. I did not have the chance to look at potential implications re performance on datasets with deep hierarchies or many subdatasets. But I feel like those could be future performance optimizations, if needed at all.

yarikoptic:
Thank you @mih !

PR #5417, opened 2021-02-11
prep-travis-forssh: Support Docker installations using Docker Machine

jwodder:
This change is necessary in order for the script to work when used by datalad/git-annex on GitHub Actions' macOS environments.

cf. https://github.com/datalad/git-annex/issues/42

### DIFF
--- a/tools/ci/prep-travis-forssh.sh
+++ b/tools/ci/prep-travis-forssh.sh
@@ -3,20 +3,25 @@ set -eu
 
 mkdir -p "$HOME/.ssh"
 
-cat >>"$HOME/.ssh/config" <<'EOF'
+if command -V docker-machine &> /dev/null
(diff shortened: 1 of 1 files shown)


yarikoptic:
woohoo!!! Could you please rebase it against `maint` -- we better fix it there as well

jwodder:
@yarikoptic Base changed to maint.

yarikoptic:
(disregard travis - NFS, and github actions red -- probably initial pre-rebased one marked red)

appveyor is odd... one Mac run is green but then 3 for `MacP38a{1,2,3}` red and even didn't run...  may be it got confused by rebase somehow?  @jwodder, could you force push some change (unless someone has a better idea on what is going on there?)

jwodder:
@yarikoptic Trivial change force-pushed.

kyleam:
The only remaining failure is the nfs stall.

PR #5420, opened 2021-02-17
ENH: Support push-url with ORA special remotes

### LINKED ISSUE #5273: Support read URL separate from write URLs in ria special remotes
Inspired by a support request in the DataLad chat room, where one and the same dataset needs to be readily configured to copy to a storage sibling via one URL, and copy from via another one.

bpoldrack:
Introduce push-url config for ORA remotes. Like git itself `url` is the
required config, that is used for read and write operations by default,
while push-url is optional and only used for write operations.
Since a special remote can't tell what annex command it is being called
from, we don't know whether to use `url` or `push-url` during PREPARE.
Therefore always start using the (fetch-) url. However, whenever we
first encounter a request from annex suggesting a write operation to the
remote (TRANSFER-STORE and REMOVE for now), switch underlying IO class.

- Closes #5273

### DIFF
--- a/datalad/distributed/ora_remote.py
+++ b/datalad/distributed/ora_remote.py
@@ -686,14 +686,18 @@ class RIARemote(SpecialRemote):
         if hasattr(self, 'configs'):
             # introduced in annexremote 1.4.2 to support LISTCONFIGS
             self.configs['url'] = "RIA store to use"
+            self.configs['push-url'] = "URL for pushing to the RIA store. " \
+                                       "Optional."
             self.configs['archive-id'] = "Dataset ID. Should be set " \
                                          "automatically by datalad"
         # machine to SSH-log-in to access/store the data
         # subclass must set this
         self.storage_host = None
+        self.storage_host_push = None
         # must be absolute, and POSIX (will be instance of PurePosixPath)
         # subclass must set this
         self.store_base_path = None
+        self.store_base_path_push = None
         # by default we can read and write
         self.read_only = False
         self.force_write = None
@@ -707,7 +711,9 @@ class RIARemote(SpecialRemote):
         self.remote_git_dir = None
         self.remote_archive_dir = None
         self.remote_obj_dir = None
-        self._io = None  # lazy
+        # lazy IO:
+        self._io = None
+        self._push_io = None
 
         # cache obj_locations:
         self._last_archive_path = None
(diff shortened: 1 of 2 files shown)


datalad/distributed/ora_remote.py
```diff
@@ -851,6 +857,25 @@ def _verify_config(self, gitdir, fail_noid=True):
...
+                                     "".format(self.ria_store_pushurl))
+
+            # support URL rewrite without talking to a DataLad ConfigManager
+            # Q is why? Why not use the config manager?
+            url_cfgs = dict()
+            url_cfgs_raw = _get_gitcfg(gitdir, "^url.*", regex=True)
+            if url_cfgs_raw:
+                for line in url_cfgs_raw.splitlines():
+                    k, v = line.split()
+                    url_cfgs[k] = v
```
kyleam, line 874: It looks like the code from 867 to here is repeated from above.  Minor and subjective, but perhaps worth extracting into a function.

kyleam (commented):
> Since a special remote can't tell what annex command it is being called
> from, we don't know whether to use `url` or `push-url` during PREPARE.
> Therefore always start using the (fetch-) url. However, whenever we
> first encounter a request from annex suggesting a write operation to the
> remote (TRANSFER-STORE and REMOVE for now), switch underlying IO class.

Makes sense, and I think from a code standpoint that turned out pretty nice.

Overall the change looks good to me.  Is the idea that create-sibling-ria will eventualy expose a way to specify the push url?

sameas was mentioned in gh-5273, and I'm sure you considered it.  Mostly out of curiousity, why do you prefer this PR's approach?  I suppose it might just be that datalad's minimum git-annex version doesn't support sameas (which came in 7.20191017).

bpoldrack:
Thanks, @kyleam !

> Is the idea that create-sibling-ria will eventualy expose a way to specify the push url?

Yes, definitely. That piece is required for the usecases this mostly is about, since the push-url then is access method reqquired for setting up the sibling in the store.

> sameas was mentioned in gh-5273, and I'm sure you considered it. Mostly out of curiousity, why do you prefer this PR's approach? I suppose it might just be that datalad's minimum git-annex version doesn't support sameas (which came in 7.20191017).

That's one aspect, but it's also more of an hassle to setup several special remotes. `create-sibling-ria` would need another switch to disable `autoenable`, be called to setup the push remote, then an `annex-initremote --sameas autoenable=True` with the fetch url for consumption. And then `create-sibling-ria` and `postclonecfg_ria` wouldn't work the way they do now, since part of the logic is to register the remote's uuid in the store itself for postclone to discover. Would then be the push remote only, while the one for consumption wasn't detectable that way. I think there may still be valid usecases where you'd want to separate push-fetch access by different remotes, but for the usecases we currently have, the push-url approach seems more straight forward.

datalad/distributed/ora_remote.py
bpoldrack: Agree, will do.

bpoldrack: Didn't move that piece into a function, but no code duplication anymore. Executed unconditionally now and only once (calling `git-config` twice is pointless). The conditional call was about `url` being an alternative to the way older `base-path` and `host` configs, that I'll deprecate in another PR (there are several to come to clean up the mess that is `ora_remote.py`).

kyleam: > calling `git-config` twice is pointless

Ah, true.  Update looks good to me.  Thanks.

kyleam (approved):

bpoldrack:
Okay, known NFS stall only - merging. Thx, @kyleam

PR #5424, opened 2021-02-18
PEP-8 and comment cleanup in ora_remote.py

bpoldrack:
Non-functional PR. Separating it out to not pollute other PRs' diff.

### DIFF
--- a/datalad/distributed/ora_remote.py
+++ b/datalad/distributed/ora_remote.py
@@ -270,7 +270,8 @@ class SSHRemoteIO(IOBase):
     It doesn't even think about a windows server.
     """
 
-    # output markers to detect possible command failure as well as end of output from a particular command:
+    # output markers to detect possible command failure as well as end of output
+    # from a particular command:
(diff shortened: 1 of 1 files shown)


datalad/distributed/ora_remote.py
```diff
@@ -416,15 +428,19 @@ def get(self, src, dst, progress_cb):
         # actual get (that is 'cat').
         # Therefore check beforehand.
         if not self.exists(src):
-            raise RIARemoteError("annex object {src} does not exist.".format(src=src))
+            raise RIARemoteError("annex object {src} does not exist."
+                                 "".format(src=src))
```
kyleam, line 432: `s/"".format/.format/` would also work, though I'm fine with `"".format()` if you prefer it style-wise.

kyleam (approved):
Nothing objectionable in there :)

datalad/distributed/ora_remote.py
bpoldrack: Yeah, to me it's more obvious that way when skimming through code.

PR #5425, opened 2021-02-18
Deprecate old ORA configs

bpoldrack:
Configuring ssh-host and base-path for an ORA remote predates ria+ URLs
and the consideration of git's URL rewrites. Those configs are also not
capable of specifying an HTTP connection let alone a separation of `url`
and `push-url`. Respect them for now if we don't find a URL, but issue a
deprecation warning. Will be removed entirely eventually.

### DIFF
--- a/datalad/distributed/ora_remote.py
+++ b/datalad/distributed/ora_remote.py
@@ -831,15 +831,6 @@ class RIARemote(SpecialRemote):
                 raise NoLayoutVersion
 
     def _load_cfg(self, gitdir, name):
-        # for now still accept the configs, if no ria-URL is known:
-        if not self.ria_store_url:
-            self.storage_host = _get_gitcfg(
-                gitdir, 'annex.ora-remote.{}.ssh-host'.format(name))
-
-            store_base_path = _get_gitcfg(
-                gitdir, 'annex.ora-remote.{}.base-path'.format(name))
-            self.store_base_path = store_base_path.strip() \
-                if store_base_path else None
         # Whether or not to force writing to the remote. Currently used to
         # overrule write protection due to layout version mismatch.
         self.force_write = _get_gitcfg(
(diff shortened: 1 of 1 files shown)


datalad/distributed/ora_remote.py
```diff
@@ -883,6 +874,39 @@ def _verify_config(self, gitdir, fail_noid=True):
...
+            self.storage_host = None if host == '0' else host
+
+            path = _get_gitcfg(gitdir,
+                               'annex.ora-remote.{}.base-path'.format(name)) or \
+                   self.annex.getconfig('base-path')
+            self.store_base_path = path.strip() if path else path
+
+            if path or host:
+                self.message("WARNING: base-path + ssh-host configs are "
+                             "deprecated and won't be considered in the future."
```
kyleam, line 892: Missing a space between "future." and "Use".

datalad/distributed/ora_remote.py
```diff
@@ -883,6 +874,39 @@ def _verify_config(self, gitdir, fail_noid=True):
             self.storage_host, self.store_base_path, self.ria_store_url = \
                 verify_ria_url(self.ria_store_url, url_cfgs)
 
+        else:
+            # for now still accept the configs, if no ria-URL is known, but
+            # issue deprecation warning:
+            host = _get_gitcfg(gitdir,
+                               'annex.ora-remote.{}.ssh-host'.format(name)) or \
+                   self.annex.getconfig('ssh-host')
+            self.storage_host = None if host == '0' else host
```
kyleam, line 883: I think moving the deleted comment about '0' (line 916) to here could avoid some head scratching from readers, but it's obviously not very important given this code is set to be removed.

kyleam (commented):
Aside from a missing space in the warning, looks good as far as I can tell.

mih (approved):
LGTM! Thx! Comments by @kyleam make sense.

datalad/distributed/ora_remote.py
bpoldrack: Thx.

datalad/distributed/ora_remote.py
bpoldrack: True. Intended to move it, but obv. forgot about it. Thx.

bpoldrack:
Thx! Addressed the comments, only NFS stall left - hhus merging.

Issue #5423, opened 2021-02-18
RFC: datalad.save.subdataset config sections to define when/how to automagically establish subdatasets


yarikoptic:
Initial use case is to automate per-month subdatasets generation for https://github.com/con/tinuous/ .

I see this feature also potentially useful for addurls as a more flexible (than the `//` in the target filepath it currently supports) way to define when to create a new subdataset, and potentially later to be used for datalad-crawler.

```bash
[datalad.save.subdataset "allsubdstext2git"]
	regexes = ^[^/]*/
	cfg = text2git
```

there could be multiple such sections with custom `"names"` to provide most of the flexibility.

May be something like that is already achievable with the hooks somehow?

cons:
- this setting will reside only in the parent super-dataset and not propagate into subdatasets.  Coding/testing should be done so that behavior is well defined, e.g. if subdatasets define their own settings, they are respected as well and probably are in effect before super-dataset consideration (but that might be tricky)

bpoldrack:
I'm not confident I understand what this is supposed to achieve. Just triggering the config procedure whenever a subdataset is created or "automagically establishing a subdataset", i.e. I have new a new folder with some files, call `save` and it creates a subdataset?

If the former - should be doable already.

yarikoptic:
>  i.e. I have new a new folder with some files, call save and it creates a subdataset?

this one.

PR #5426, opened 2021-02-19
NF: cfg_noannex procedure to convert annexed repo into the one without annex

yarikoptic:
Initial motivation is to be able to use  `datalad create -c noannex` ,
in particular in the light of https://github.com/datalad/datalad/issues/5423
RFC.

Yes, we do have  --no-annex  option but it could not be applied "post-fact".
Also I like the consistency of having dataset configuration via  -c option
to create.

TODOs
- [x] seek initial feedback
- [x] test
- [x] also removed `no_annex` Python option for `create` which was deprecated as of 0.13.0 .
- possibly options? (--force, --ensure-content, ...?) -- decided not to bother for now (until a use case)

### DIFF
--- a/datalad/core/local/create.py
+++ b/datalad/core/local/create.py
@@ -168,11 +168,6 @@ class Create(Interface):
             doc="""enforce creation of a dataset in a non-empty directory""",
             action='store_true'),
         description=location_description,
-        no_annex=Parameter(
-            # hide this from the cmdline parser, replaced by `annex`
-            args=tuple(),
-            doc="""this option is deprecated, use `annex` instead""",
-            action='store_true'),
         annex=Parameter(
             args=("--no-annex",),
             dest='annex',
@@ -208,24 +203,10 @@ class Create(Interface):
             force=False,
             description=None,
             dataset=None,
-            no_annex=_NoAnnexDefault,
             annex=True,
             fake_dates=False,
             cfg_proc=None
     ):
-        # TODO: The current release of datalad-metalad (v0.2.1) still uses
-        # no_annex in its tests. Remove this compatibility kludge once a
-        # release is made, which will include 16a170e (2020-09-08).
-        if no_annex is not _NoAnnexDefault:
-            # the two mirror options do not agree and the deprecated one is
-            # not at default value
-            warnings.warn("datalad-create's `no_annex` option is deprecated "
-                          "and will be removed in a future release, "
-                          "use the reversed-sign `annex` option instead.",
-                          DeprecationWarning)
-            # honor the old option for now
-            annex = not no_annex
-
         # we only perform negative tests below
         no_annex = not annex
 
--- a/datalad/resources/procedures/cfg_noannex.py
+++ b/datalad/resources/procedures/cfg_noannex.py
@@ -1,0 +1,36 @@
+"""Procedure to uninit Git-annex (if initialized), and create/save .noannex file to prevent annex initialization
+
+If there are git-annex'ed files already, git annex uninit and this procedure will fail.
+
+"""
+
+import sys
+
+from datalad.distribution.dataset import require_dataset
+from datalad.support.annexrepo import AnnexRepo
+from datalad import lgr
+
+ds = require_dataset(
+    sys.argv[1],
+    check_installed=True,
+    purpose='configuration')
+
+if isinstance(ds.repo, AnnexRepo):
+    repo = ds.repo
+    # TODO: if procedures can have options -- add --force handling/passing
+    #
+    # annex uninit unlocks files for which there is content (nice) but just proceeds
+    # and leaves broken symlinks for files without content.  For the current purpose
+    # of this procedure we just prevent "uninit" of any annex with some files already
+    # annexed.
+    if any(repo.call_annex_items_(['whereis', '--all'])):
+        raise RuntimeError("Annex has some annexed files, unsafe")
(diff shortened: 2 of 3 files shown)


bpoldrack:
Agree with the utility of applying such a procedure later!

mih:
Same here. Ideally with the ambition to get rid of --noannex

yarikoptic:
travis -- NFS 

appveyor -- as in another PR for some reason just marked one of the Macs red without details (so unlikely related)

bpoldrack:
> appveyor -- as in another PR for some reason just marked one of the Macs red without details (so unlikely related)

Happened to me recently too with different builds. "Rerun incomplete" did the job. Since there's no log whatsoever I assume there's something flaky going on on startup of the workers.

yarikoptic:
ok, we are as green as we could be.  Please bless/merge this PR @kyleam @bpoldrack @mih ;)

datalad/core/local/create.py
```diff
@@ -208,24 +203,10 @@ def __call__(
...
-        # TODO: The current release of datalad-metalad (v0.2.1) still uses
-        # no_annex in its tests. Remove this compatibility kludge once a
-        # release is made, which will include 16a170e (2020-09-08).
-        if no_annex is not _NoAnnexDefault:
-            # the two mirror options do not agree and the deprecated one is
-            # not at default value
-            warnings.warn("datalad-create's `no_annex` option is deprecated "
-                          "and will be removed in a future release, "
-                          "use the reversed-sign `annex` option instead.",
-                          DeprecationWarning)
```
kyleam, line 225: It's not particularly important, so I have no objections to removing this, but note that the description in the comment isn't stale: metalad still hasn't been released, so the current release still uses `no_annex` in its tests.

datalad/resources/procedures/cfg_noannex.py
```diff
@@ -0,0 +1,41 @@
...
+    if isinstance(ds.repo, AnnexRepo):
+        repo = ds.repo
+        # TODO: if procedures can have options -- add --force handling/passing
+        #
+        # annex uninit unlocks files for which there is content (nice) but just proceeds
+        # and leaves broken symlinks for files without content.  For the current purpose
+        # of this procedure we just prevent "uninit" of any annex with some files already
+        # annexed.
+        if any(repo.call_annex_items_(['whereis', '--all'])):
+            raise RuntimeError("Annex has some annexed files, unsafe")
```
kyleam, line 27: [ just a non-actionable thought while reading ]

The particular output isn't important, so `whereis`s pretty verbose output could be reduced to something like `"--format=.\n"`.  This isn't supported by datalad's minimum git-annex version and would only speedup the error case (and even then, only when there are many files), though.

kyleam (approved):
LGTM

datalad/core/local/create.py
yarikoptic: oh, I missed that.  I guess I will just merge without this commit ? or may be we should just encourage datalad-metalad to release? ;) WDYT @christian-monch @mih

datalad/resources/procedures/cfg_noannex.py
yarikoptic: and in principle should not have notable effect here since hopefully annex starts producing records as soon as it has any and our `call_annex_items_` generator yielding them, so `any()` would just stop iterating as soon as it gets at least one

datalad/core/local/create.py
yarikoptic: TODO for the merge to add `#!/usr/bin/env python3` shebang in the light of #5436 (don't want to torture CI for no reason)

datalad/resources/procedures/cfg_noannex.py
kyleam: > since hopefully annex starts producing records as soon as it has any and our `call_annex_items_` generator yielding them

Perhaps your "in principle" meant you realize this is the case and are looking forward to some future enhancement, but fwiw `call_annex_items_` operates on the entire output of the command.

yarikoptic: oh...  ["как обманчива природа" сказал ёжик слезая с кактуса](https://translate.google.com/?sl=auto&tl=en&text=%22%D0%BA%D0%B0%D0%BA%20%D0%BE%D0%B1%D0%BC%D0%B0%D0%BD%D1%87%D0%B8%D0%B2%D0%B0%20%D0%BF%D1%80%D0%B8%D1%80%D0%BE%D0%B4%D0%B0%22%20%D1%81%D0%BA%D0%B0%D0%B7%D0%B0%D0%BB%20%D1%91%D0%B6%D0%B8%D0%BA%20%D1%81%D0%BB%D0%B5%D0%B7%D0%B0%D1%8F%20%D1%81%20%D0%BA%D0%B0%D0%BA%D1%82%D1%83%D1%81%D0%B0&op=translate)

datalad/core/local/create.py
kyleam: > I guess I will just merge without this commit ?

All right, if you're okay with that, that'd be my preference.

> TODO for the merge to add `#!/usr/bin/env python3`

Sounds good.

Both done with the merge in 76d9c36eac.

kyleam:
As suggested [above](https://github.com/datalad/datalad/pull/5426#discussion_r582135959), fixed up and merged locally (76d9c36eac).

PR #5428, opened 2021-02-19
ENH: optional push-url for create-sibling-ria

bpoldrack:
Expose ORA's new `push-url` in `create-sibling-ria`. This is used to configure the ORA special remote as well as to create the sibling in store.


kyleam:
> Should `create-sibling-ria` use the `push-url` for the git remote, too (same way: `url` + `push-url`, of course)?

Seems like a good idea to me as long as there isn't an obvious scenario where that wouldn't be desirable.  I can't think of one, but then again I've mostly been using git's `remote.<name>.pushurl` to frame/think about this new ria option, so perhaps I'm missing a use case where they don't overlap.

mih (approved):
It looks good and works as advertised for me. Thx!

Issue #5412, opened 2021-02-05
Number of commits for subdataset addition doubled


mih:
https://github.com/datalad/datalad/pull/5346 by @bpoldrack  added an additional commit to record another URL for a subdataset. However, there is no need to keep that in a dedicated commit, and it adds a lot of commits in datasets with a lot of subdatasets.

All URLs should be consolidated in a single commit.

bpoldrack:
Looked into this and it turns out a bit more complicated than it sounds.
#5346 added a `subdatasets` call in `clone` to set the `datalad-url` for the cloned subdataset. This call is invoked right after `clone`'s call to `save`, thus creating an additional commit after `save`.
I thought to just move that `subdatasets` call up and add a `--nosave` option to `subdatasets` in order to resolve this issue. However, it's not that easy. If such a `clone` call doesn't come from a `get` but actually creates the subdataset,  `subdatasets` has another issue: `datalad subdatasets not-yet-existing-submodule --set-property some value` does nothing. Zero-exit with no results whatsoever. Now, if you look at `subdatasets` as a standalone command, this should probably yield an error. Using it the way I intended to, we would need it to just create an entry in `.gitmodules` for a submodule, that doesn't exist yet, suggesting some kind of `--force` option for `subdatasets`. I'm not sure though what exactly this `force` should do under what circumstances if we look at the `subdatasets` command as a whole. Opinions, @datalad/developers?
An alternative could be to call `git-config` directly from within `clone` instead of `subdatasets` in order to set `datalad-url`. But this seems more troublesome as it requires to duplicate internal logic from `subdatasets`. Especially properly parsing `.gitmodules` if it already exists and figuring the correct submodule entry if `clone` was called by `get` (submodule name vs path).

Any suggestion for a different approach?

mih:
For a minimal change, I would move the `subdatasets` call to immediately before the `save` call, `git reset HEAD`1 the result, if there was a change in response to the call, add `.gitmodules` to the `path` that is given to `save`.

For a more involved change: RF `subdatasets` to provide an internal helper that performs the modifications, but not the `save`.

I think the minimal change would be sufficient to achieve the immediate goal without causing much cost.

bpoldrack:
> For a minimal change, I would move the subdatasets call to immediately before the save call, git reset HEAD1 the result, if there was a change in response to the call, add .gitmodules to the path that is given to save.

[...]

Issue #3244, opened 2018-12-29
save --amend


mih:
This or something that has the same effect is needed to be able to combine subsequent changesets. Otherwise, we would have to support some kind of (interactive) rebasing/squashing.

After 30s of thinking about it, it might be as simple as passing `--amend` to the respective commit calls.

Here is the (now closed) original PR that aimed at making this happen: https://github.com/datalad/datalad-revolution/pull/71

PR #5430, opened 2021-02-22
ENH: Amending commits

bpoldrack:
Looking into a resolution of #3244 (and subsequently #5412).

Introducing `--amend` for `save`. By default this keeps the previous commit message if none was passed to save. `--amend` can not be applied recursively for now (see #5455 ).

- minor fix: `GitRepo._call_git` didn't respect its `env` parameter
- minor fix: `save` exited zero w/o a result when called on empty repo (no commit or empty commit only). Now: either `notneeded`     or proceeds in case of `--amend`.
- minor enh: `GitRepo.update_ref` now optionally takes `oldvalue` parameter

### DIFF
--- a/datalad/core/local/save.py
+++ b/datalad/core/local/save.py
@@ -101,6 +101,10 @@ class Save(Interface):
         dict(text="Tag the most recent saved state of a dataset",
              code_py="save(version_tag='bestyet')",
              code_cmd="datalad save --version-tag 'bestyet'"),
+        dict(text="Save a specific change but integrate into last commit keeping "
+                  "the already recorded commit message",
+             code_py="save(path='myfile.txt', amend=True)",
+             code_cmd="datalad save myfile.txt --amend")
     ]
 
     _params_ = dict(
@@ -147,6 +151,18 @@ class Save(Interface):
             (see https://git-annex.branchable.com/tips/largefiles).
             """),
         jobs=jobs_opt,
+        amend=Parameter(
+            args=('--amend',),
+            action='store_true',
+            doc="""if set, changes are not recorded in a new, separate
+            commit, but are integrated with the changeset of the previous
+            commit, and both together are recorded by replacing that
+            previous commit. A commit message and authorship information
+            are re-used, if needed. Note, that the previous commits in all
+            subdatasets are amended if this is used recursively. Hence all
+            commit messages are changed if you amend recursively while also
+            passing a message.
+            """),
     )
 
     @staticmethod
@@ -159,6 +175,7 @@ class Save(Interface):
                  message_file=None,
                  to_git=None,
                  jobs=None,
+                 amend=False,
                  ):
         if message and message_file:
             raise ValueError(
@@ -267,6 +284,17 @@ class Save(Interface):
                             type='dataset')
                 paths_by_ds[superds] = superds_status
 
+        # We need to anticipate empty (amend) commits in subdatasets that
+        # currently appear to be clean. Change state of any subdataset to
+        # modified in order for it to be committed upwards after amending.
+        # Note, that at least a message is required in order to actually commit
+        # in a subdataset that otherwise is clean.
+        if amend and recursive and message:
+            for d in paths_by_ds:
+                for p in paths_by_ds[d]:
+                    if paths_by_ds[d][p]['type'] == 'dataset':
+                        paths_by_ds[d][p]['state'] = 'modified'
+
         def save_ds(args, version_tag=None):
             pdspath, paths = args
 
@@ -281,7 +309,8 @@ class Save(Interface):
                 pds_repo.pathobj / p.relative_to(pdspath): props
                 for p, props in paths.items()}
             start_commit = pds_repo.get_hexsha()
-            if not all(p['state'] == 'clean' for p in pds_status.values()):
+            if not all(p['state'] == 'clean' for p in pds_status.values()) or \
+                    (amend and message):
                 for res in pds_repo.save_(
                         message=message,
                         # make sure to have the `path` arg be None, as we want
@@ -294,7 +323,8 @@ class Save(Interface):
                         # we are supplying the full status already, do not
                         # detect anything else
                         untracked='no',
-                        _status=pds_status):
+                        _status=pds_status,
+                        amend=amend):
                     # TODO remove stringification when datalad-core can handle
                     # path objects, or when PY3.6 is the lowest supported
                     # version
--- a/datalad/core/local/save.py
+++ b/datalad/core/local/save.py
@@ -157,11 +157,8 @@ class Save(Interface):
             doc="""if set, changes are not recorded in a new, separate
             commit, but are integrated with the changeset of the previous
             commit, and both together are recorded by replacing that
-            previous commit. A commit message and authorship information
-            are re-used, if needed. Note, that the previous commits in all
-            subdatasets are amended if this is used recursively. Hence all
-            commit messages are changed if you amend recursively while also
-            passing a message.
+            previous commit. This is mutually exclusive with recursive
+            operation.
             """),
     )
 
@@ -180,6 +177,9 @@ class Save(Interface):
         if message and message_file:
             raise ValueError(
                 "Both a message and message file were specified for save()")
+
+        if amend and recursive:
+            raise ValueError("Cannot amend a commit recursively.")
 
         path = ensure_list(path)
 
@@ -283,18 +283,6 @@ class Save(Interface):
                             state='untracked',
                             type='dataset')
                 paths_by_ds[superds] = superds_status
-
-        # We need to anticipate empty (amend) commits in subdatasets that
-        # currently appear to be clean. Change state of any subdataset to
-        # modified in order for it to be committed upwards after amending.
-        # Note, that at least a message is required in order to actually commit
-        # in a subdataset that otherwise is clean.
-        if amend and recursive and message:
-            for d in paths_by_ds:
-                for p in paths_by_ds[d]:
-                    if paths_by_ds[d][p]['type'] == 'dataset' and \
-                            paths_by_ds[d][p]['state'] == 'clean':
-                        paths_by_ds[d][p]['state'] = 'modified'
 
         def save_ds(args, version_tag=None):
             pdspath, paths = args
--- a/datalad/core/local/save.py
+++ b/datalad/core/local/save.py
@@ -292,7 +292,8 @@ class Save(Interface):
         if amend and recursive and message:
             for d in paths_by_ds:
                 for p in paths_by_ds[d]:
-                    if paths_by_ds[d][p]['type'] == 'dataset':
+                    if paths_by_ds[d][p]['type'] == 'dataset' and \
+                            paths_by_ds[d][p]['state'] == 'clean':
                         paths_by_ds[d][p]['state'] = 'modified'
 
         def save_ds(args, version_tag=None):
@@ -370,6 +371,22 @@ class Save(Interface):
                     status='error',
                     message=('cannot tag this version: %s', e.stderr.strip()))
                 yield dsres
+
+        if not paths_by_ds:
+            # Special case: empty repo. There's either an empty commit only or
+            # none at all. An empty one we can amend otherwise there's nothing
+            # to do.
+            if amend and ds.repo.get_hexsha():
+                yield from save_ds((ds.pathobj, dict()), version_tag=version_tag)
+
+            else:
+                yield dict(action='save',
+                           type='dataset',
+                           path=ds.path,
+                           refds=ds.path,
+                           status='notneeded',
+                           logger=lgr)
+            return
 
         # TODO: in principle logging could be improved to go not by a dataset
         # but by path(s) within subdatasets. That should provide a bit better ETA
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -3356,15 +3356,14 @@ class AnnexRepo(GitRepo, RepoInterface):
             old_sha = self.get_hexsha(corresponding_branch)
 
             org_commit_pointer = corresponding_branch + "~1"
-            author_name = self.format_commit("%an", org_commit_pointer)
-            author_email = self.format_commit("%ae", org_commit_pointer)
-            author_date = self.format_commit("%ad", org_commit_pointer)
-            old_parent = self.format_commit("%P", org_commit_pointer)
+            author_name, author_email, author_date, \
+            old_parent, old_message = self.format_commit(
+                "%an%x00%ae%x00%ad%x00%P%x00%B", org_commit_pointer).split('\0')
             new_env = (self._git_runner.env
                        if self._git_runner.env else os.environ).copy()
             # `message` might be empty - we need to take it from the to be
             # amended commit in that case:
-            msg = message or self.format_commit("%B", org_commit_pointer)
+            msg = message or old_message
             new_env.update({
                 'GIT_AUTHOR_NAME': author_name,
                 'GIT_AUTHOR_EMAIL': author_email,
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -3350,6 +3350,7 @@ class AnnexRepo(GitRepo, RepoInterface):
 
             adjusted_branch = self.get_active_branch()
             corresponding_branch = self.get_corresponding_branch()
+            old_sha = self.get_hexsha(corresponding_branch)
 
             org_commit_pointer = corresponding_branch + "~1"
             author_name = self.format_commit("%an", org_commit_pointer)
@@ -3376,8 +3377,10 @@ class AnnexRepo(GitRepo, RepoInterface):
                                     read_only=False)
             new_sha = out.strip()
 
-            self.update_ref("refs/heads/" + corresponding_branch, new_sha)
-            self.update_ref("refs/basis/" + adjusted_branch, new_sha)
+            self.update_ref("refs/heads/" + corresponding_branch,
+                            new_sha, old_sha)
+            self.update_ref("refs/basis/" + adjusted_branch,
+                            new_sha, old_sha)
             self.localsync(managed_only=True)
 
     def localsync(self, remote=None, managed_only=False):
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -3356,7 +3356,7 @@ class AnnexRepo(GitRepo, RepoInterface):
             author_name = self.format_commit("%an", org_commit_pointer)
             author_email = self.format_commit("%ae", org_commit_pointer)
             author_date = self.format_commit("%ad", org_commit_pointer)
-
+            old_parent = self.format_commit("%P", org_commit_pointer)
             new_env = (self._git_runner.env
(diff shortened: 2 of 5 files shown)


datalad/support/annexrepo.py
```diff
@@ -3327,12 +3327,62 @@ def _save_add(self, files, git=None, git_opts=None):
...
+            org_commit_pointer = corresponding_branch + "~1"
+            author_name = self.format_commit("%an", org_commit_pointer)
+            author_email = self.format_commit("%ae", org_commit_pointer)
+            author_date = self.format_commit("%ad", org_commit_pointer)
+
+            new_env = (self._git_runner.env
+                       if self._git_runner.env else os.environ).copy()
+            from datalad.utils import make_tempfile
+
+            with make_tempfile() as index_file:
```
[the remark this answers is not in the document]
bpoldrack: Good point. Thx!

datalad/support/annexrepo.py
```diff
@@ -3327,12 +3328,65 @@ def _save_add(self, files, git=None, git_opts=None):
...
+                out, _ = self._call_git(["commit-tree",
+                                         corresponding_branch + "^{tree}",
+                                         "-p", corresponding_branch + "~2",
+                                         "-m", msg],
+                                        env=new_env,
+                                        read_only=False)
+                new_sha = out.strip()
+
+            self.update_ref("refs/heads/" + corresponding_branch, new_sha)
+            self.update_ref("refs/basis/" + adjusted_branch, new_sha)
```
kyleam, line 3388: I'm just looking at this bit and haven't tried to take in any of the larger change, but do you need to use a temporary index here?  It looks like you already have the tree you want.

<details>
<summary>example</summary>

```sh
set -eu

cd "$(mktemp -d "${TMPDIR:-/tmp}"/g-XXXXXXX)"
git init
git commit --allow-empty -m c1

branch=$(git symbolic-ref HEAD)

echo one >f1
git add f1
git commit -mf1

echo two >f2
git add f2
git commit -mf2

# Go elsewhere to test an out-of-tree amend.
git checkout HEAD~2
git ls-files

git update-ref "$branch" "$(git commit-tree -m amended -p "$branch"~2 "$branch^{tree}")" "$branch"

git log --oneline --stat --decorate master
```

```
[...]
c6b9e14 (master) amended
 f1 | 1 +
 f2 | 1 +
293b68e (HEAD) c1
```

</details>

kyleam, line 3388: I know our `update_ref` wrapper doesn't currently support it, but if we're doing this sort of dance, I'd really like to see these `update-ref` calls use `<oldvalue>` as a safety measure.

bpoldrack: No, don't need it here anymore, agree. Which is why it's an indentation less than the lines above. So, one of us is reading it wrong.

bpoldrack: Agreed. Will do.

kyleam: Sorry, by "here" I didn't mean right here at this line.  Let's try dropping "here": do you need to use a temporary index?

bpoldrack: Ah, I see. You mean specifying the parent and referencing branch^{tree} should give all that's needed. You're probably right. 
The idea stemmed from thinking about amending in the adjusted branch, where cahnges would be based on a different state. But we don't. It's all committed already.
Will try.

bpoldrack: Done. Thank you for spotting this, @kyleam !

bpoldrack:
Rebased and cleaned up history after tests passed.

datalad/core/local/save.py
```diff
@@ -267,6 +284,17 @@ def __call__(path=None, message=None, dataset=None,
...
+        # We need to anticipate empty (amend) commits in subdatasets that
+        # currently appear to be clean. Change state of any subdataset to
+        # modified in order for it to be committed upwards after amending.
+        # Note, that at least a message is required in order to actually commit
+        # in a subdataset that otherwise is clean.
+        if amend and recursive and message:
+            for d in paths_by_ds:
+                for p in paths_by_ds[d]:
+                    if paths_by_ds[d][p]['type'] == 'dataset':
+                        paths_by_ds[d][p]['state'] = 'modified'
```
kyleam, line 296: Hmm, amending recursively seems messsy.  As a naive user, I think I'd expect it to only amend subdatasets that changed in the last commit.  Perhaps `--amend` and `--recursive` should just be marked as incompatible, at least initially.  If I'm thinking about gh-5412 correctly, recursive amending isn't needed there.

datalad/support/annexrepo.py
```diff
@@ -3327,12 +3327,61 @@ def _save_add(self, files, git=None, git_opts=None):
                     if 'error-messages' in r else None,
                     logger=lgr)
 
-    def _save_post(self, message, status, partial_commit):
+    def _save_post(self, message, status, partial_commit, amend=False):
+
+        if amend and self.is_managed_branch() and \
+                self.format_commit("%B").strip() == "git-annex adjusted branch":
```
kyleam, line 3333: Hard coding this should be safe :)

Annex/AdjustedBranch.hs:

```
{- This message should never be changed. -}
adjustedBranchCommitMessage :: String
adjustedBranchCommitMessage = "git-annex adjusted branch"
```

datalad/support/annexrepo.py
```diff
@@ -3327,12 +3327,61 @@ def _save_add(self, files, git=None, git_opts=None):
...
+            # `message` might be empty - we need to take it from the to be
+            # amended commit in that case:
+            msg = message or self.format_commit("%B", org_commit_pointer)
+            new_env.update({
+                'GIT_AUTHOR_NAME': author_name,
+                'GIT_AUTHOR_EMAIL': author_email,
+                'GIT_AUTHOR_DATE': author_date
+            })
+            # TODO: What about fake_dates in this context?
+            #       ATM this is applied on top by _call_git().
```
kyleam, line 3371: It should just continue to override, as it does, I think.

datalad/support/gitrepo.py
```diff
@@ -3023,7 +3033,7 @@ def update_submodule(self, path, mode='checkout', init=False):
                 )
         # TODO: return value
 
-    def update_ref(self, ref, value, symbolic=False):
+    def update_ref(self, ref, value, oldvalue=None, symbolic=False):
```
kyleam, line 3036: Repositioning `symbolic` is a breaking change but probably okay.  I don't see any spots in our code (core or extensions) that pass it as a positional argument, and it's probably not used in third-party code.  Then again, perhaps the more natural positioning of `oldvalue` after `value` isn't worth any potential breakage, and `oldvalue` should just be moved to the end.

(In my view it'd be good to use keyword-only arguments more for newly added functions to avoid these sorts of issues.  It looks like `_get_dot_git` is the only function in the code base that currently does.)

datalad/support/gitrepo.py
```diff
@@ -1674,13 +1674,20 @@ def commit(self, msg=None, options=None, _datalad_msg=False, careless=True,
 
         orig_msg = msg
         if not msg:
-            msg = 'Recorded changes'
-            _datalad_msg = True
+            if '--amend' in options and '--no-edit' not in options:
+                # don't overwrite old commit message with our default message
+                # by default, but re-use old one. In other words: Make --no-edit
+                # the default:
+                options += ["--no-edit"]
```
kyleam, line 1681: Shouldn't the `'--no-edit' not in options` check be moved within the `'--amend' in options` arm?  Otherwise, if `commit` is called directly with `--amend` _and_  `--no-edit`, the generic 'Recorded changes' message would be used.

```diff
diff --git a/datalad/support/tests/test_gitrepo.py b/datalad/support/tests/test_gitrepo.py
index 8deb1f1396..b62b58b870 100644
--- a/datalad/support/tests/test_gitrepo.py
+++ b/datalad/support/tests/test_gitrepo.py
@@ -335,6 +335,11 @@ def test_GitRepo_commit(path):
     assert_equal(len(list(gr.get_branch_commits_())), 2)
     assert_equal(gr.format_commit("%B").strip(), "amend message")

+    gr.commit(options=["--amend", "--no-edit"])
+    assert_repo_status(gr)
+    assert_equal(len(list(gr.get_branch_commits_())), 2)
+    assert_equal(gr.format_commit("%B").strip(), "amend message")
+
     # nothing to commit doesn't raise by default:
     gr.commit()
     # but does with careless=False:
```

```
$ python -m nose -vs datalad.support.tests.test_gitrepo:test_GitRepo_commit                                                                                   datalad.support.tests.test_gitrepo.test_GitRepo_commit ... FAIL
[...]
FAIL: datalad.support.tests.test_gitrepo.test_GitRepo_commit
[...]
    assert_equal(gr.format_commit("%B").strip(), "amend message")
AssertionError: '[DATALAD] Recorded changes' != 'amend message'
- [DATALAD] Recorded changes
+ amend message
```

datalad/support/annexrepo.py
```diff
@@ -3327,12 +3327,61 @@ def _save_add(self, files, git=None, git_opts=None):
...
+            out, _ = self._call_git(["commit-tree",
+                                     corresponding_branch + "^{tree}",
+                                     "-p", corresponding_branch + "~2",
+                                     "-m", msg],
+                                    env=new_env,
+                                    read_only=False)
```
kyleam, line 3377: There's an edge case to worry about here:

```sh
set -eu

datalad create
git annex adjust --unlock
echo one >one
datalad save -ma --amend
```

```
[...]
CommandError: 'git commit-tree 'master^{tree}' -p 'master~2' -m a' failed with exitcode 128 under /tmp/dl-u1rhTW7
fatal: not a valid object name master~2
add(ok): one (file)
```

kyleam (commented):
Neat.  Testing it out on my end, that seems to work nicely.  Here are my comments from a mostly complete read-through.  (I didn't look over the tests very closely.)

datalad/support/annexrepo.py
bpoldrack: The question comes from the fact, that `commit --amend` and therefore this implementation take the authoring date from the previous commit. So, I thought either that one was faked already then it doesn't matter or it wasn't but now it's enabled. Not obvious to me what's the right thing in the latter case.

datalad/core/local/save.py
bpoldrack: Not needed there indeed.

> As a naive user, I think I'd expect it to only amend subdatasets that changed in the last commit.

I think that expectation isn't naive at all. ;)
I'm not sure what any user would expect, though. Not even sure what I want it to be. ;-)
Fulfilling your expectation instead seems perfectly reasonable, since the only "real" usecase I can think of is a previous `save --recursive`.

Would like to hear other @datalad/developers view on this.

datalad/support/annexrepo.py
bpoldrack: :)

I'm always uncomfortable with that kind of thing (and I didn't look it up), but on the other hand coming up with something to match instead that may or may not be more robust against a change in annex is arbitrary too considering a sample size of 1. And actually analyzing what kind of changes the commit makes to figure its "nature" is that of an annex commit in the branch ... well, seems more expensive with no real gain in robustness.

datalad/support/gitrepo.py
bpoldrack: Yeah I couldn't find a spot using `symbolic` positional either. My rational here was the other way around: Future change should most likely remove that - in my view weird - janus headed functionality and move symbolic-ref into its own function. If I had appended it, we would then encounter the situation where we'd remove an argument from inbetween.

datalad/support/annexrepo.py
bpoldrack: Right!

datalad/support/gitrepo.py
bpoldrack: Oops. Correct!

datalad/support/annexrepo.py
kyleam: Yeah, I understand why you're asking the question.  When we're faking dates I think it's fair to assume time information has been entirely handed over to datalad, so I'd just keep it simple and continue to tick.

datalad/support/gitrepo.py
kyleam: > and move symbolic-ref into its own function.

True, or just delete the method, as its only extra logic/handling on top of a straight `call_git` is looking at `symbolic` and deciding the command.

Anyway, your choice/rationale is fine by me.

bpoldrack:
So, the test for fixing the case where we amend a commit w/o parent fails on crippled FS on windows. The failure is real:

```
C:\Users\appveyor\test2>git log
commit 3faa0c63c6a3e4b6c48862fa53fde1f57f0f2161 (HEAD -> adjusted/master(unlocked))
Author: Appveyor Almighty <test@appveyor.land>
Date:   Fri Feb 26 14:25:35 2021 +0000

    git-annex adjusted branch

commit a8e13e782900d286caa816948be061144bb3efc9 (master, refs/basis/adjusted/master(unlocked))
Author: Appveyor Almighty <test@appveyor.land>
Date:   Fri Feb 26 14:25:35 2021 +0000

    commit before entering adjusted branch

C:\Users\appveyor\test2>datalad save -m new --amend
C:\Users\appveyor\test2>git log
commit 3faa0c63c6a3e4b6c48862fa53fde1f57f0f2161 (HEAD -> adjusted/master(unlocked))
Author: Appveyor Almighty <test@appveyor.land>
Date:   Fri Feb 26 14:25:35 2021 +0000

    git-annex adjusted branch

commit a8e13e782900d286caa816948be061144bb3efc9 (master, refs/basis/adjusted/master(unlocked))
Author: Appveyor Almighty <test@appveyor.land>
Date:   Fri Feb 26 14:25:35 2021 +0000

    commit before entering adjusted branch
```

However, that is when amending w/o a modification. With a change to commit it works and once that is done it does so w/o further change:

```
C:\Users\appveyor\test2>echo some > some
C:\Users\appveyor\test2>datalad save -m new --amend
add(ok): some (file)
save(ok): . (dataset)
action summary:
  add (ok: 1)
  save (ok: 1)

C:\Users\appveyor\test2>git log
commit d5e1ec2a442f790284f9a3c5c0874eee2760b059 (HEAD -> adjusted/master(unlocked))
Author: Appveyor Almighty <test@appveyor.land>
Date:   Fri Feb 26 14:25:35 2021 +0000

    git-annex adjusted branch

commit 7a71da7f1456c5e16ed774316671a48dd1cdd6fb (master, refs/basis/adjusted/master(unlocked))
Author: Appveyor Almighty <test@appveyor.land>
Date:   Fri Feb 26 14:25:35 2021 +0000

    new

C:\Users\appveyor\test2>datalad save -m brandnew --amend
save(ok): . (dataset)

C:\Users\appveyor\test2>git log
commit 49f8fec3a38dbf7d0ff4352caa29d2ce63ad8e3d (HEAD -> adjusted/master(unlocked))
Author: Appveyor Almighty <test@appveyor.land>
Date:   Fri Feb 26 14:25:35 2021 +0000

    git-annex adjusted branch

commit e3c5ec6dfeb24fd78c2179edc7681105b98d770a (master, refs/basis/adjusted/master(unlocked))
Author: Appveyor Almighty <test@appveyor.land>
Date:   Fri Feb 26 14:25:35 2021 +0000

    brandnew
```

So, I guess the issue here is not the adjusted branch (generally amending the first commit works w/ adjusted branch, too), but the fact, that the to be amended commit as well as the amendment are empty. Therefore there's no tree object to commit.

There seems to be no `--allow-empty` for `git commit-tree`, though. Hmmm.

kyleam:
> With a change to commit it works and once that is done it does so w/o further change

I believe this is because with an empty tree, the `Dataset.save`s `paths_by_ds` is empty and `GitRepo.save` is never called.

bpoldrack:
> I believe this is because with an empty tree, the Dataset.saves paths_by_ds is empty and GitRepo.save is never called.

Yes, that seems to be it. It exits zero with no message and no result, though. So, even w/o amend needs a fix, I guess.

```
❱ mkdir empty-repo                                                                                                                         128 !
(datalad-dev) ben@tree in /tmp
❱ git -C empty-repo init
Initialized empty Git repository in /tmp/empty-repo/.git/
(datalad-dev) ben@tree in /tmp
❱ git -C empty-repo annex init
init  (scanning for unlocked files...)
ok
(recording state in git...)
(datalad-dev) ben@tree in /tmp
(datalad-dev) ben@tree in /tmp
❱ datalad -f json_pp save -d empty-repo -m empty                                                                                             2 !
(datalad-dev) ben@tree in /tmp
❱ 
```

bpoldrack:
Changed back to draft until everything (including CI mess) is sorted again.

datalad/core/local/save.py
bpoldrack: Okay, so I think, we should go with @kyleam's expectation: On `save --amend --recursive`, amend only in subdatasets, that were changed in the previous commit. Modifications in other subdatasets should lead to a separate commit in them, but the changed commit should then be amended upwards. This seems to be the most intuitive thing to aim for, in my mind. I expect it to be a mess implementation-wise, though. Will have a look.

bpoldrack: Moved to #5455. Making `--amend` and `--recursive` incompatible for now.

datalad/support/annexrepo.py
```diff
@@ -3327,12 +3327,62 @@ def _save_add(self, files, git=None, git_opts=None):
...
+            org_commit_pointer = corresponding_branch + "~1"
+            author_name = self.format_commit("%an", org_commit_pointer)
+            author_email = self.format_commit("%ae", org_commit_pointer)
+            author_date = self.format_commit("%ad", org_commit_pointer)
+            old_parent = self.format_commit("%P", org_commit_pointer)
+            new_env = (self._git_runner.env
+                       if self._git_runner.env else os.environ).copy()
+            # `message` might be empty - we need to take it from the to be
+            # amended commit in that case:
+            msg = message or self.format_commit("%B", org_commit_pointer)
```
mih, line 3367: I would prefer, if we could limit the amount of git calls to collect this type of info. What about this:

```py
author_name, author_email, author_date, old_parent, old_message = \
    self.format_commit("%an%x00%ae%x00%ad%x00%P%x00%B").split('\0')
msg = message or old_message
```

mih (changes requested):
Very cool! Thx for implementing this!

I took a look at the coverage report -- which seems invalid. The lines reported missing here are reported covered when running on a crippledfs locally.

I gave it a try with a couple of datasets that I am working with and it works smoothly in easy-peasy-mode and with adjusted mode!

Going through the code, I only spotted one piece that seems needlessly wasteful, and suggested a change that should be leaner and still good enough.

Thx again!

datalad/support/annexrepo.py
bpoldrack: Yes, I wondered what to use as separator that is certain to not be included in such fields as author_name. `\0` seems like a reasonable solution. Will do.

bpoldrack: Hehe, nope. `format_commit` already splits by `\0`.

mih (approved):
A conflict appeared now, but otherwise LGTM! Thx!

PR #5433, opened 2021-02-23
BF(TST,workaround): skip stalling on travis NFS test

yarikoptic:
and only on Travis and only the stalling run

### DIFF
--- a/datalad/support/tests/test_annexrepo.py
+++ b/datalad/support/tests/test_annexrepo.py
@@ -2194,6 +2194,10 @@ def test_annexjson_protocol(path):
     assert_not_in('errors from JSON records', str(e.exception))
 
 
+# see https://github.com/datalad/datalad/pull/5400 for troubleshooting
+# for stalling with unlock=False, and then with unlock=True it took >= 300 sec
+# https://github.com/datalad/datalad/pull/5433#issuecomment-784470028
+@skip_if(on_travis and 'nfs' in os.getenv('TMPDIR', ''))  # TODO. stalls
 # http://git-annex.branchable.com/bugs/cannot_commit___34__annex_add__34__ed_modified_file_which_switched_its_largefile_status_to_be_committed_to_git_now/#comment-bf70dd0071de1bfdae9fd4f736fd1ec
 # https://github.com/datalad/datalad/issues/1651
 @known_failure_githubci_win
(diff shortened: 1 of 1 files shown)


yarikoptic:
odd... appveyor errorred out to create a job for some (not all) Mac environments?

kyleam:
`test_commit_annex_commit_changed` failed as too slow on the nfs run.  I guess it should be marked with `@turtle`?  Or perhaps these checks shouldn't apply to the nfs build?

```
ERROR: datalad.support.tests.test_annexrepo.test_commit_annex_commit_changed(True,)
  File "/tmp/dl-miniconda-5cts7vfx/lib/python3.8/site-packages/nose/case.py", line 134, in run
    self.runTest(result)
  File "/tmp/dl-miniconda-5cts7vfx/lib/python3.8/site-packages/nose/case.py", line 152, in runTest
    test(result)
AssertionError: Test was too slow (took 309.3363s, threshold was 30.0000s)
```

https://travis-ci.com/github/datalad/datalad/jobs/485574427

kyleam:
> odd... appveyor errorred out to create a job for some (not all) Mac environments?

The empty log failure with some macos builds has been happening occasionally, though the rate seems to have increased.

yarikoptic:
> AssertionError: Test was too slow (took 309.3363s

yikes... takes ~8sec for both runs on my laptop.  And that test (I assume since we do not run nose with `-v`) used to pass just fine not so long ago... smth is really odd.  But oh well -- this PR is to close our eyes on odd anyways, so let me skip the entire test not just one run. heh

datalad/support/tests/test_annexrepo.py
```diff
@@ -2194,6 +2194,10 @@ def test_annexjson_protocol(path):
     assert_not_in('errors from JSON records', str(e.exception))
 
 
+# see https://github.com/datalad/datalad/pull/5400 for troubleshooting
+# for stalling with unlock=False, and then with unlock=True it took >= 300 sec
+# https://github.com/datalad/datalad/pull/5433#issuecomment-784470028
+@skip_if(on_travis and 'nfs' in os.getenv('TMPDIR', ''))  # TODO. stalls
```
yarikoptic, line 2200: doh -- bloody generators... thank you @kyleam !

PR #5441, opened 2021-02-25
RM: deprecated no_annex Python API option

yarikoptic:
leftover from #5426 

needs datalad-metalad release IIRC


mih:
I think it is OK to merge this, and have metalad be fixed separately.

mih (approved):
LGTM. Thx for thinking off it.

mih:
LOL, the last test run was too old. `master` itself was incompatible with this change. No idea how that could happen. Will fix.

PR #5442, opened 2021-02-25
Temp test devel annex maint

yarikoptic:
should have fresh git-annex-standalone.  I believe our allowances for time should have been increased so it should not exceed allowed and must be green!


yarikoptic:
woohoo -- travis has passed, and runs were not that long -- 51 min top.  So I am now uploading this version to neurodebian to replace an elderly 7.20190819+git2-g908476a9b-1~ndall+1 .

PR #5445, opened 2021-02-26
BF: need to eval debian_installer call, and now we have --sudo option to allow sudo ops

yarikoptic:
Without --sudo ok  it now will ask user to confirm sudo operations by default.
We do have 1 matrix run which installs a .deb so it would need sudo.

Without eval it is impossible to pass  -e  options like I needed in a
https://github.com/datalad/datalad/pull/5442  so I decided to just have it
by default here

OSX tune up might have not been needed -- did not check. but better be safe than sorry ;)

PS resubmitting now against `maint`

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -250,8 +250,8 @@ install:
   - sh: "[ -n \"$INSTALL_SYSPKGS\" ] && ( [ \"x${APPVEYOR_BUILD_WORKER_IMAGE}\" = \"xmacOS\" ] && brew install -q ${INSTALL_SYSPKGS} || sudo apt-get install --no-install-recommends -y ${INSTALL_SYSPKGS} ) || true"
   # Install git-annex on windows, otherwise INSTALL_SYSPKGS can be used
   # deploy git-annex, if desired
-  - cmd: IF DEFINED INSTALL_GITANNEX python datalad_installer.py %INSTALL_GITANNEX%
-  - sh: "[ -n \"${INSTALL_GITANNEX}\" ] && python datalad_installer.py ${INSTALL_GITANNEX}"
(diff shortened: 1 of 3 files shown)


.appveyor.yml
```diff
@@ -250,8 +250,8 @@ install:
   - sh: "[ -n \"$INSTALL_SYSPKGS\" ] && ( [ \"x${APPVEYOR_BUILD_WORKER_IMAGE}\" = \"xmacOS\" ] && brew install -q ${INSTALL_SYSPKGS} || sudo apt-get install --no-install-recommends -y ${INSTALL_SYSPKGS} ) || true"
   # Install git-annex on windows, otherwise INSTALL_SYSPKGS can be used
   # deploy git-annex, if desired
-  - cmd: IF DEFINED INSTALL_GITANNEX python datalad_installer.py %INSTALL_GITANNEX%
-  - sh: "[ -n \"${INSTALL_GITANNEX}\" ] && python datalad_installer.py ${INSTALL_GITANNEX}"
+  - cmd: IF DEFINED INSTALL_GITANNEX python datalad_installer.py --sudo ok %INSTALL_GITANNEX%
```
mih, line 253: Windows users also need to allow sudo? Or does it use a different privilege elevation?

yarikoptic: no, I just added it "because why not" and if we discover that we need some privilege elevation we could add it under `--sudo` option "umbrella" in datalad-installer without needing to adjust our CI scripts here.

yarikoptic:
oh, forgot that we had not merged this one. #5448 could bring further tune ups but this one at least should resolve `--sudo ok` necessity for now.  Let's proceed with a merge

PR #5454, opened 2021-03-03
BF: log - accept non-str (but str()-able) instances

yarikoptic:
I have ran into a use case where external library logs an exception, but then
our logger starts puking.  Default formatter tollerates them just fine, so we
better do too.


kyleam (approved):

kyleam:
The one appveyor failure is a setup failure, but I think all of the other builds sufficiently exercise the added test.

PR #5458, opened 2021-03-03
BF: Unlock used to pass resolved paths to status

### LINKED ISSUE #5456: BUG: Unlock '.' within an untracked subdataset yields "dataset containing given paths is not underneath the reference dataset"
Found while digging this issue: https://github.com/datalad/datalad-ukbiobank/issues/64
See comment therein: https://github.com/datalad/datalad-ukbiobank/issues/64#issuecomment-789615058

Probably related: https://github.com/nipy/heudiconv/issues/456, https://github.com/datalad/datalad/issues/3928

Problem: `unlock` resolves the given path before passing it on to `status`.

bpoldrack:
This destroys the logic within status that is based on "raw" input. In the particular case triggering this PR, `status` wouldn't realize the original call was passing `.` as the `path` argument.

Closes #5456

### DIFF
--- a/datalad/interface/tests/test_unlock.py
+++ b/datalad/interface/tests/test_unlock.py
@@ -242,3 +242,11 @@ def test_unlock_cant_unlock(path):
             action="unlock",
             status=status,
             path=str(ds.pathobj / f))
+
+
+@with_tree(tree={'subdir': {'sub': {}}})
+def test_unlock_gh_5456(path):
+    path = Path(path)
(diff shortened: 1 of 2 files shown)


kyleam (approved):
Thanks, looks good.

The branch is based off of maint, but it looks like you forgot to change the base for the PR.  Assuming nobody else chimes in requesting changes, I'd be fine with skipping the "change base on PR, trigger rebuild" step, instead just merging this to maint locally and waiting for the PR to be closed when maint hits master.

PR #5459, opened 2021-03-03
Overhaul ORA remote's HTTP access to RIA stores

### LINKED ISSUE #5451: enableremote ORA doesn't fail on invalid/inaccessible HTTP URL
In opposition to `ria+(ssh|file):// ...` URLs, ORA's `initremote` method doesn't properly test it can actually find a RIA store at the given HTTP URL. This in turn leads to `clone` not reconfiguring, when cloning via a different scheme.

bpoldrack:
Restructure ORA remote's `HTTPRemoteIO` to remove special cases in ORA's code, where HTTP is treated somewhat differrent than SSH or local access.
For RIA stores, this leads to the `ria-layout-version` files being required to be served as well (which technically wasn't the case before). It also means that the request URLs now use paths for annexed objects according to the specified layout, whereas before we'd always send the same request and it was up to the server to consider its layout and deliver the right thing.

HTTP stores are validated on `annex initremote`/`annex enableremote` the same way they are when using `ssh:` or `file:` scheme URLs. Therefore, autoenabling will fail under the same conditions, making `clone`'s reconfiguration business consistent across access methods.

Closes #5451

### DIFF
--- a/datalad/distributed/ora_remote.py
+++ b/datalad/distributed/ora_remote.py
@@ -630,13 +630,12 @@ class HTTPRemoteIO(object):
     # NOTE: For now read-only. Not sure yet whether an IO class is the right
     # approach.
 
-    def __init__(self, ria_url, dsid, buffer_size=DEFAULT_BUFFER_SIZE):
-        assert ria_url.startswith("ria+http")
-        self.base_url = ria_url[4:]
-        if self.base_url[-1] == '/':
-            self.base_url = self.base_url[:-1]
+    def __init__(self, url, buffer_size=DEFAULT_BUFFER_SIZE):
+        if not url.startswith("http"):
+            raise RIARemoteError("Expected HTTP URL, but got {}".format(url))
 
-        self.base_url += "/" + dsid[:3] + '/' + dsid[3:]
+        self.store_url = url.rstrip('/')
+
         # make sure default is used when None was passed, too.
         self.buffer_size = buffer_size if buffer_size else DEFAULT_BUFFER_SIZE
 
@@ -644,17 +643,32 @@ class HTTPRemoteIO(object):
         # Note, that we need the path with hash dirs, since we don't have access
         # to annexremote.dirhash from within IO classes
 
-        url = self.base_url + "/annex/objects/" + str(key_path)
-        response = requests.head(url, allow_redirects=True)
-        return response.status_code == 200
+        return self.exists(key_path)
 
     def get(self, key_path, filename, progress_cb):
         # Note, that we need the path with hash dirs, since we don't have access
         # to annexremote.dirhash from within IO classes
 
-        url = self.base_url + "/annex/objects/" + str(key_path)
+        url = self.store_url + str(key_path)
         from datalad.support.network import download_url
         download_url(url, filename, overwrite=True)
+
+    def exists(self, path):
+        # use same signature as in SSH and Local IO, although validity is
+        # limited in case of HTTP.
+        url = self.store_url + path.as_posix()
+        try:
+            response = requests.head(url, allow_redirects=True)
+        except Exception as e:
+            raise RIARemoteError(str(e))
+
+        return response.status_code == 200
+
+    def read_file(self, file_path):
+
+        from datalad.support.network import download_url
+        content = download_url(self.store_url + file_path.as_posix())
+        return content
 
 
 def handle_errors(func):
@@ -1041,12 +1055,7 @@ class RIARemote(SpecialRemote):
                 # fall back on the UUID for the annex remote
                 self.archive_id = self.annex.getuuid()
 
-        if not isinstance(self.io, HTTPRemoteIO):
-            self.get_store()
-
-        # else:
-        # TODO: consistency with SSH and FILE behavior? In those cases we make
-        #       sure the store exists from within initremote
+        self.get_store()
 
         self.annex.setconfig('archive-id', self.archive_id)
         # make sure, we store the potentially rewritten URL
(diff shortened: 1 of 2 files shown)


datalad/distributed/ora_remote.py
```diff
@@ -1023,9 +1032,23 @@ def initremote(self):
...
+            #       underlying structure on FS and validate it isn't really
+            #       possible. However, may be we shouldn't do that much for SSH
+            #       and Local IO either.
+            #
+            #       In any case, do a minimal check for whether we can connect
+            #       to the store, since initremote/enableremote have to fail
+            #       otherwise. Note, that we shouldn't expect the webserver to
+            #       serve things like ria-layout-version file. This is mainly to
+            #       verify that we can in principle connect via that URL.
+            if not self.io.exists('HEAD'):
```
kyleam, line 1049: * It looks like `.exists` will fail if this isn't a pathlib object

 * Is it safe to assume the bare repo `HEAD` exists in a store?  It won't with, e.g., stores created with `create-sibling-ria --storage-sibling=only`, right?

bpoldrack: > Is it safe to assume the bare repo HEAD exists in a store? It won't with, e.g., stores created with create-sibling-ria --storage-sibling=only, right?

Good point, forgot we allow for the absence of a git repo. All that is actually required for the remote to work is correct response to requests (HEAD|GET) to annex keys. But for initremote/enableremote there's no key we could generally assume to be there. So, I wonder what to test for. We could execute a checkpresent for an arbitrary key. Issue is: What does a 404 really tell us? We can tell host was resolved and there's a webserver responding. That's at least something. On the other hand that doesn't verify we can get anything related to that dataset.
May be we need to just define something else, a valid RIA store needs to serve over HTTP. `ria-layout-version` (at least at dataset level, not necessarily top-level) would be an obvious choice that would also increase flexibility for URL building (mapping to a different layout could then be done client-side just like with SSH/local). What do you think, @mih?

mih: I hope I was not blocking this PR for 5 months, because I have nothing substantial to say. #5459 is rather terse and does not have any information on the context of the error. Taking what is provided it is unclear to me what `initremote` could do to prevent a future `enableremote` from failing. A server could vanish any time, or become misconfigured. To me this sounds like a UX of `enableremote` rather than a problem of `initremote`. Of course it would be good to check sanity of URLs given to `initremote`.

Looking at the code at the code I see that such checks are made in ~`get_store()`, but are explicitly disabled for HTTP access in `initremote()`. I'd say that it needs no change in behavior or logic, just a removal of the shortcuts made for HTTP access.

bpoldrack: @mih : There's no difference between `initremote` and `enableremote` from the point of view of a special remote. `INITREMOTE` is what it gets from annex it both cases.

> Looking at the code at the code I see that such checks are made in ~get_store(), but are explicitly disabled for HTTP access in initremote(). I'd say that it needs no change in behavior or logic, just a removal of the shortcuts made for HTTP access.

Yes and no. Needs an equivalent. But in opposition to local/SSH access we don't really need the same things to be checked, since the special remote doesn't actually operate on the store's FS tree.

I will change this to request the `ria-layout-version` files both from the store- and dataset level. This will be a more general change, as ORA will then use that file to determine the key URL, just like in the case of local FS and SSH. So, abandon the idea that we approach it differently via HTTP in that we only ever send uniform requests independent of any `version` f the store and leave it to the server to map the requests differently, if needed. Instead we get that version and build the URL from it exactly like we'd do it for non-HTTP access.
This not only gives us something to verify but also enables another change in the future: A version file, that doesn't necessarily give a number/label referring to a predefined-in-datalad layout, but alternatively specifies the layout directly, so that a store's maintainer has all the flexibility to set it up w/o a need for the special remote to know the layout. I wanted to go there anyway, so I think this is going to solve more than one problem.

bpoldrack:
Changed the approach and updated PR description accordingly.

datalad/distributed/ora_remote.py
mih: I am resolving this old conversation as it seems no longer relevant.

datalad/distributed/ora_remote.py
```diff
@@ -630,32 +630,48 @@ class HTTPRemoteIO(object):
...
-        self.base_url = ria_url[4:]
-        if self.base_url[-1] == '/':
-            self.base_url = self.base_url[:-1]
+    def __init__(self, url, buffer_size=DEFAULT_BUFFER_SIZE):
+        if not url.startswith("http"):
+            raise RIARemoteError("Expected HTTP URL, but got {}".format(url))
+
+        self.store_url = url
+        if self.store_url[-1] == '/':
+            self.store_url = self.store_url[:-1]
```
mih, line 639: Unrelated to this PR, but if downstream code requires a URL to not have a trailing slash, this better be an unconditional `.rstrip('/')`.

datalad/distributed/ora_remote.py
```diff
@@ -630,32 +630,48 @@ class HTTPRemoteIO(object):
...
 
-        url = self.base_url + "/annex/objects/" + str(key_path)
+        url = self.store_url + str(key_path)
         from datalad.support.network import download_url
         download_url(url, filename, overwrite=True)
 
+    def exists(self, path):
+        # use same signature as in SSH and Local IO, although validity is
+        # limited in case of HTTP.
+        url = self.store_url + str(path.as_posix())
```
mih, line 661: `str()` is redundant. `as_posix()` returns a `str`.

datalad/distributed/ora_remote.py
```diff
@@ -630,32 +630,48 @@ class HTTPRemoteIO(object):
...
+            response = requests.head(url, allow_redirects=True)
+        except Exception as e:
+            raise RIARemoteError(str(e))
+
+        return response.status_code == 200
+
+    def read_file(self, file_path):
+
+        from datalad.support.network import download_url
+        content = download_url(self.store_url + str(file_path.as_posix()))
```
mih, line 672: Another redundant `str()`

datalad/distributed/ora_remote.py
```diff
@@ -1103,9 +1114,18 @@ def io(self):
             if self._local_io():
                 self._io = LocalIO()
             elif self.ria_store_url.startswith("ria+http"):
-                self._io = HTTPRemoteIO(self.ria_store_url,
-                                        self.archive_id,
-                                        self.buffer_size)
+                # TODO: That construction of "http(s)://host/" should prob be
```
mih, line 1117:
```suggestion
                # TODO: That construction of "http(s)://host/" should probably be
```

mih (changes requested):
Thanks for the changes. I left a few minor comments.

There should be a test though. Given that the original issue is about not failing, in case of an invallid URL such a test requires no particular infrastructure, just a random URL with no RIA store behind it.

mih (approved):
LGTM. Thanks for the changes.

Issue #4244, opened 2020-03-05
Fate of AnnexRepo.sync()


mih:
`sync()` is used twice in the entire code base (one usage is in `merge_annex()`. This method is a trivial frontend to a `git-annex sync` call that is easily done with `call_git` -- and is done in this fashion in `AnnexRepo._save_post()`. `AnnexRepo.merge_annex()` is actively being avoid (e.g. in `update()`), because it leaves behind the `synced/...` branches.

I propose to dissolve both methods and replace them with `call_git()` calls, or a local sync helper from #4243 that also handles the removed of the `synced/...`branches.

yarikoptic:
I am ok to stop using `annex sync` (to avoid `synced/` branches proliferation), but it is not clear to me why in general `call_git` is preferable over having a `sync` method possibly just calling out to `call_git` if that is all it does.

mih:
The existing `AnnexRepo.sync()` method offers access to only a subset of the functionality of `annex sync` and merely converts Python kwargs into a command arg list. It is untested, and barely used. As such it can be considered "dead weight", and consequently should be removed.

mih:
https://github.com/datalad/datalad/pull/4243 replaces all usage of `merge_annex()`

mih:
Retitled after #4243 was merged.

PR #5461, opened 2021-03-04
RF: Remove last use of AnnexRepo.sync() and deprecate method

mih:
Fixes gh-4244

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -1972,7 +1972,9 @@ class AnnexRepo(GitRepo, RepoInterface):
 
     def sync(self, remotes=None, push=True, pull=True, commit=True,
              content=False, all=False, fast=False):
-        """Synchronize local repository with remotes
+        """This method is deprecated, use call_annex(['sync', ...]) instead.
+
+        Synchronize local repository with remotes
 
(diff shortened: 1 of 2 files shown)


datalad/support/annexrepo.py
```diff
@@ -1972,7 +1972,9 @@ def merge_annex(self, remote=None):  # do not use anymore, use localsync()
 
     def sync(self, remotes=None, push=True, pull=True, commit=True,
              content=False, all=False, fast=False):
-        """Synchronize local repository with remotes
+        """This method is deprecated, use call_annex(['sync', ...]) instead.
+
+        Synchronize local repository with remotes
```
bpoldrack, line 1977: I think we should develop a habit of this:
`warnings.warn("AnnexRepo.sync() is deprecated, use call_annex(['sync', ...]) instead.", DeprecationWarning)`

bpoldrack (changes requested):

datalad/support/annexrepo.py
mih: I added such a warning.

bpoldrack (approved):

kyleam (approved):

Issue #5460, opened 2021-03-03
we are breeding ssh control path processes which persist for days


yarikoptic:
```
$> ps auxw | grep ssh.*Contro[l] | nl            
     1  yoh        920  0.0  0.0  16000     8 ?        Ss   Mar01   0:00 ssh -fN -o ControlMaster=auto -o ControlPersist=15m -o ControlPath=/home/yoh/.cache/datalad/sockets/b0da704a datasets.datalad.org
     2  yoh       2024  0.0  0.0  16008     4 ?        Ss   Mar01   0:00 ssh -fN -o ControlMaster=auto -o ControlPersist=15m -o ControlPath=/home/yoh/.cache/datalad/sockets/b0da704a datasets.datalad.org
     3  yoh       2069  0.0  0.0  16000     8 ?        Ss   Mar01   0:00 ssh -fN -o ControlMaster=auto -o ControlPersist=15m -o ControlPath=/home/yoh/.cache/datalad/sockets/b0da704a datasets.datalad.org
     4  yoh       2627  0.0  0.0  16000     4 ?        Ss   Mar01   0:00 ssh -fN -o ControlMaster=auto -o ControlPersist=15m -o ControlPath=/home/yoh/.cache/datalad/sockets/b0da704a datasets.datalad.org
     5  yoh      10628  0.0  0.0  16000     4 ?        Ss   Feb28   0:00 ssh -fN -o ControlMaster=auto -o ControlPersist=15m -o ControlPath=/home/yoh/.cache/datalad/sockets/b0da704a datasets.datalad.org
     6  yoh      21818  0.0  0.0  16000     8 ?        Ss   Mar01   0:00 ssh -fN -o ControlMaster=auto -o ControlPersist=15m -o ControlPath=/home/yoh/.cache/datalad/sockets/b0da704a datasets.datalad.org
     7  yoh      21849  0.0  0.0  16008     4 ?        Ss   Mar01   0:00 ssh -fN -o ControlMaster=auto -o ControlPersist=15m -o ControlPath=/home/yoh/.cache/datalad/sockets/b0da704a datasets.datalad.org
     8  yoh      22436  0.0  0.0  16000     8 ?        Ss   Mar01   0:00 ssh -fN -o ControlMaster=auto -o ControlPersist=15m -o ControlPath=/home/yoh/.cache/datalad/sockets/b0da704a datasets.datalad.org
[...]

Issue #5465, opened 2021-03-04
ssh: we are "holding" stdout/stderr of the ssh mux process forbidding a clean exit if piped


yarikoptic:
dedicated extract from https://github.com/datalad/datalad/issues/5460#issuecomment-790923286

repoducer demonstrating the issue is a stalling script like

```
$> cat ./chronic-halt
#!/bin/bash

set -eu

(
pstree -p $$
echo "starting ssh mux"
ssh -fN -o ControlMaster=auto -o ControlPersist=15m -o ControlPath=/home/yoh/.cache/datalad/sockets/chronic-stall datasets.datalad.org
echo "done, exiting"
pstree -p $$
) 2>&1 | tee /tmp/chronic-stall.log

```
in interaction with #5464 we ending up with a stalled process.  Since it is interaction with #5464 and we do take care about closing all the ssh mux processes we started, this issue should not manifest itself.  But I still wonder if we should somehow take care about not dragging those pipes of `ssh -fN` with us (e.g. if we decide to change policy for those ssh processes to outlast our datalad process life time)

yarikoptic:
couldn't the WitlessRunner with Capture protocol be used here instead of a plain `Popen`?

yarikoptic:
@mih @christian-monch @bpoldrack @jwodder - any ideas on the best ways to address this one?

mih:
I would support a move to a runner-based execution. That it is not done already is just a long ignored relic.

yarikoptic:
Ha!  I was about to email

<details>
<summary>following question/email to openssh-unix-dev  mindrot.org </summary> 

```
Dear OpenSSH developers,

Originally posted/troubleshooted on our DataLad project issue tracker:
https://github.com/datalad/datalad/issues/5465

In DataLad we establish "control socket" for subsequent connections (primarily
by git or git-annex) via ssh to a remote host.   The problem came up from CRON
jobs which used  chronic command (by Joey Hess, shipped within moreutils in
Debian)  around our datalad command invocation.  I think we had our datalad
process stalling (forgot detail, sorry).  I have troubleshooted it to the fact
that the process/shell with  ssh -fN  invocation never exits if ssh's
stderr is not redirected to a file.  Here is a "reproducer" which would never
exit:


    #!/bin/bash

    set -eu

    # to clean up from previous run
    ps auxw | grep chronic-[s]tall | awk '{print $2;}' | xargs -r kill
[...]

Issue #5464, opened 2021-03-04
ssh: somewhere not closing ssh mux


yarikoptic:
dedicated issue extract from https://github.com/datalad/datalad/issues/5460#issuecomment-790912418

Issue #4101, opened 2020-01-28
failing test_archives.test_ExtractedArchive of 0.12.2 on NFS mount


yarikoptic:
https://travis-ci.org/datalad/datalad/jobs/642979829?utm_medium=notification&utm_source=email
building state as of debian pkg for the release
https://travis-ci.org/datalad/datalad/builds/642979821?utm_medium=notification&utm_source=email :
```
======================================================================
FAIL: datalad.tests.test_archives.test_ExtractedArchive
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/datalad/tests/utils.py", line 559, in newfunc
    return t(*(arg + (d,)), **kw)
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/datalad/tests/test_archives.py", line 196, in test_ExtractedArchive
    op.join(fn_archive_obscure, '3.txt')
AssertionError: [' "\';b&b&cΔЙקم๗あ `| / "\';a&b&cΔЙקم๗あ `| '] != [' "\';b&b&cΔЙקم๗あ `| / "\';a&b&cΔЙקم๗あ `| ', ' "\';b&b&cΔЙקم๗あ `| /3.txt']
```
so smells like similar (but different) indigestion to https://github.com/datalad/datalad/issues/3769 of plain 7z invocation on tricky filenames?

yarikoptic:
strangely it passed on master https://travis-ci.org/datalad/datalad/jobs/642975399 or https://travis-ci.org/datalad/datalad/jobs/642747907

bpoldrack:
This test failed again the same way here: https://app.travis-ci.com/github/datalad/datalad/jobs/587461343

PR #5466, opened 2021-03-04
BF: interprocess lock for checking/opening ssh socket connections

yarikoptic:
This should prevent race condition between testing if socket path
already exists and establishing it.

Closes: #5460

Related issues discovered while troubleshooting for this one are
  https://github.com/datalad/datalad/issues/5465
  https://github.com/datalad/datalad/issues/5464

### DIFF
--- a/datalad/support/sshconnector.py
+++ b/datalad/support/sshconnector.py
@@ -423,7 +423,11 @@ class MultiplexSSHConnection(BaseSSHConnection):
         ]
         self.ctrl_path = Path(ctrl_path)
         self._opened_by_us = False
-        self._lock = threading.Lock()
+        # used by @fasteners.locked
+        self._lock = [
+            threading.Lock(),
+            fasteners.process_lock.InterProcessLock(self.ctrl_path.with_suffix('.lck'))
+        ]
 
     def __call__(self, cmd, options=None, stdin=None, log_output=True):
         """Executes a command on the remote.
@@ -520,12 +524,6 @@ class MultiplexSSHConnection(BaseSSHConnection):
         ------
         ConnectionOpenFailedError
           When starting the SSH ControlMaster process failed.
-        """
-        with fasteners.process_lock.InterProcessLock(self.ctrl_path.with_suffix('.lck')):
-            return self._open()
-
-    def _open(self):
-        """The actual body of 'open' but without any locking primitives
         """
         # the socket should vanish almost instantly when the connection closes
         # sending explicit 'check' commands to the control master is expensive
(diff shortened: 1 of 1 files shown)


yarikoptic:
@bpoldrack @mih you might like to test this on your ssh use cases to see if it is not adding a notable slow down -- our benchmarking wouldn't test for this.

yarikoptic:
ha -- travis hits old #4101 (and again on maint). appveyor is just unhappy and this time against more than those Macs... even ubuntu freaks out.

datalad/support/sshconnector.py
```diff
@@ -521,6 +521,12 @@ def open(self):
         ConnectionOpenFailedError
           When starting the SSH ControlMaster process failed.
         """
+        with fasteners.process_lock.InterProcessLock(self.ctrl_path.with_suffix('.lck')):
```
kyleam, line 524: Could `datalad.support.locking.lock_if_check_fails` be used here to avoid creating a lock in many cases?

kyleam (approved):
Makes sense to me.

datalad/support/sshconnector.py
yarikoptic: oh right 
1. we should just have a lock instance bound to the instance and reuse it (should shave off some cycles).  
2. Indeed we should be able to use `lock_if_check_fails` but it would loose that (reuse instance) so I might need to RF it to allow to reuse a lock provided from outside to take advantage of `1.`

yarikoptic: Decided against using `lock_if_check_fails`: it was coded up with use cases in mind where check would check for something you expect to not disappear (e.g. downloaded annex key for an archive) between you checking and then doing some next action... so for faster paced use cases I am afraid there might be a race condition that we would happily not bother establishing the socket while it disappears from us.  So, to be safer than sorry,

I think there might still be benefit from reusing a lock instance so I have pushed a RF which does that while relying on a feature of fasteners.locked to have a list of locks instead of a single one.  So the diff becomes shorter ;)

datalad/support/sshconnector.py
```diff
@@ -423,7 +423,11 @@ def __init__(self, ctrl_path, sshri, **kwargs):
         ]
         self.ctrl_path = Path(ctrl_path)
         self._opened_by_us = False
-        self._lock = threading.Lock()
+        # used by @fasteners.locked
+        self._lock = [
+            threading.Lock(),
+            fasteners.process_lock.InterProcessLock(self.ctrl_path.with_suffix('.lck'))
+        ]
```
kyleam, line 430: Oh, that's fancy.  Thanks for leaving the comment to point future confused readers in the right direction.

kyleam (approved):

Issue #5467, opened 2021-03-04
appveyor needs "assistance"


yarikoptic:
recently the number of failed to even start OSX workers increased.  see e.g. https://ci.appveyor.com/project/mih/datalad/builds/38069462 where even ubuntu freaked out.  IMHO it needs to be troubleshooted.  Since a paid plan, @mih - could you please reach out to appveyor support?

mih:
https://help.appveyor.com/discussions/problems/29047-cannot-assign-job-to-cloudsgroups-macos-pro

mih:
I think we can close this. The story is behind the support forum link. The issue still shows sporadicly, but there is nothing else TODO.

PR #5468, opened 2021-03-05
BF(TEMP): temporarily disable a set of appveyor mac builds due to errorring out

yarikoptic:
AFAIK there were no single recent PR where those do not error out.

should be reverted after resolved (ref: #5467)

edit: manually canceled travis build


yarikoptic:
#5467 is presumably resolved, so let's close this for now

PR #5476, opened 2021-03-09
Make `--output-format default` accessible

### LINKED ISSUE #2481: Default result format is inaccessible via 'datalad -f default ...'
If a command sets a custom `result_renderer`, I don't see a way to get the non-custom format. Looking at `datalad --help`, I'd expect `-f default` to always give me this format.

    [...]
      -f {default,json,json_pp,tailored,'<template>'}, --output-format {default,json,json_pp,tailored,'<template>'}
                            select format for returned command results. 'default'
                            give one line per result reporting action, status,
                            path and an optional message;
    [...]

However, "default" instead uses whatever is specified by the class's `result_renderer`.

This is because we ignore `common_output_format` if it is "default":

<https://github.com/datalad/datalad/blob/43e37edc6f00b6edf250f1edc0116b9bd62f3f87/datalad/interface/base.py#L403-L405>

So I'd say that the problem is that we're conflating the default renderer for a command and the "default" output format.  Perhaps we should give the output format a name other than "default", so that `-f default` means "use the command's default renderer" and `-f <other name>` means "use the format described in the help message above".

Thoughts?

---

Here's an example of what I'm talking about.

```python
from datalad.interface.base import Interface
from datalad.interface.base import build_doc
from datalad.interface.results import get_status_dict


@build_doc
class Foo(Interface):
    from datalad.distribution.dataset import EnsureDataset
    from datalad.distribution.dataset import datasetmethod
    from datalad.interface.utils import eval_results
    from datalad.support.constraints import EnsureNone
    from datalad.support.param import Parameter

    result_renderer = "json_pp"

    _params_ = dict(
        dataset=Parameter(
            args=("-d", "--dataset"),
            doc=""""dataset to foo.""",
    )

    @staticmethod
    @eval_results
    def __call__(dataset=None):
        yield get_status_dict("foo",
                              message="did foo",
                              type="dataset",
                              path=dataset.path if dataset else "nowhere",
                              status="ok")


__datalad_plugin__ = Foo
```

```shell
% datalad foo
{
  "action": "foo",
  "path": "nowhere",
  "status": "ok",
  "type": "dataset"
}

% datalad -f json foo
{"action": "foo", "path": "nowhere", "status": "ok", "type": "dataset"}

% datalad -f default foo
{
  "action": "foo",
  "path": "nowhere",
  "status": "ok",
  "type": "dataset"
}
```

mih:
Running something like `datalad -f default status` does not actually
switch to the standard result renderer, because we use the "tailored"
one by default. As pointed out in #2481 this not only conflates two
different meanings of *default*, but also doesn't really make sense.

From my POV the implementation really wants `"tailored"` to be the default
mode, and not the "default" renderer to be enabled by default.

Instead of getting creative with new synonyms for these terms, I went
for the minimal change that achieves the desired result (make it
possible to have a command with a custom result renderer run with the
standard one instead).

I switch the default value of `common_output_format` to `"tailored"`,
which already is the effective default behavior. This makes a
specification of an alternative `"default"` value actually switch the
renderer to the default one.

One could argue that `"default"` would be better named `"standard"`
(renderer), but I will not walk on this battle field ;-)

Closes #2481

### DIFF
--- a/datalad/cmdline/main.py
+++ b/datalad/cmdline/main.py
@@ -121,15 +121,17 @@ def setup_parser(
         variables in the process environment.""")
     parser.add_argument(
         '-f', '--output-format', dest='common_output_format',
-        default='default',
+        default='tailored',
         type=ensure_unicode,
         metavar="{default,json,json_pp,tailored,'<template>'}",
-        help="""select format for returned command results. 'default' give one line
+        help="""select format for returned command results. 'tailored'
+        enables a command-specific rendering style that is typically
+        tailored to human consumption, if there is one for a specific
+        command, or otherwise falls back on the the 'default' output
+        format (this is the standard behavior); 'default' give one line
         per result reporting action, status, path and an optional message;
         'json' renders a JSON object with all properties for each result (one per
-        line); 'json_pp' pretty-prints JSON spanning multiple lines; 'tailored'
-        enables a command-specific rendering style that is typically
-        tailored to human consumption (no result output otherwise),
+        line); 'json_pp' pretty-prints JSON spanning multiple lines;
         '<template>' reports any value(s) of any result properties in any format
         indicated by the template (e.g. '{path}'; compare with JSON
         output for all key-value choices). The template syntax follows the Python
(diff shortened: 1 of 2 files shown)


kyleam (approved):
Thanks for digging this one back up.  Looks like a good solution to me.

PR #5479, opened 2021-03-10
Tst nometalad

mih:
This extension is undergoing a substantial change. There is no point
in testing it. Given that the list of tested extension here is
also an abitrary subset of existing extension, there is also little
point in changing the workflow to use the latest extension release.

Targeting `maint`, but needs pickup in `master` too.

### DIFF
--- a/.github/workflows/test_extensions.yml
+++ b/.github/workflows/test_extensions.yml
@@ -13,7 +13,6 @@ jobs:
         extension: [
             datalad-neuroimaging,
             datalad-container,
-            datalad-metalad,
             datalad-crawler,
         ]
     steps:


bpoldrack:
I think I'd prefer to test released metalad instead, like suggested in https://github.com/datalad/datalad-metalad/issues/70.
If datalad core is to break that one, I need to know (at least), since this would lead to datalad-hirni not being able to use such a version of datalad core until there's a compatible metalad.

mih:
@bpoldrack Why not replace metalad tests with hirni tests then? From my point it is suboptimal to test a proxy that we know needs to change.

bpoldrack:
Wouldn't really consider it a proxy since what's broken in that case really is metalad, but I agree anyway - failure won't be addressed in metalad, but would either lead to reconsideration in datalad or required action in hirni. Hence - retract my proposal ;-)

mih:
Alright, proceeding with this one for now. Adding hirni as a test run can be its own PR.

PR #5480, opened 2021-03-10
Reduce number of commits when cloning into a dataset

bpoldrack:
When a dataset is cloned into another one as a subdataset, we record the originally given URL in `.gitmodules`. This used to be a dedicated commit in addition to the one adding the new dataset as a subdataset. The former is now amended to the latter.

### DIFF
--- a/datalad/core/distributed/clone.py
+++ b/datalad/core/distributed/clone.py
@@ -316,12 +316,40 @@ class Clone(Interface):
                     result_filter=None,
                     result_xfm=None,
                     on_failure='ignore'):
-                yield r
 
-            # Modify .gitmodules to contain originally given url. This is
-            # particularly relevant for postclone routines on a later `get` for
-            # that subdataset. See gh-5256.
-            ds.subdatasets(path, set_property=[("datalad-url", source)])
+                # Modify .gitmodules to contain originally given url. This is
+                # particularly relevant for postclone routines on a later `get`
+                # for that subdataset. See gh-5256.
+                if r['action'] == 'save' and r['type'] == 'dataset':
+                    if r['status'] == 'ok':
+                        # New subdataset actually saved. Amend the modification
+                        # of .gitmodules. Note, that we didn't allow to deviate
+                        # from git default behavior WRT a submodule's name vs
+                        # its path when we made this a new subdataset.
+                        subds_name = path.relative_to(ds.pathobj)
+                        ds.repo.call_git(
+                            ['config',
+                             '--file',
+                             '.gitmodules',
+                             '--replace-all',
+                             'submodule.{}.{}'.format(subds_name,
+                                                      "datalad-url"),
+                             source]
+                        )
+                        yield from ds.save('.gitmodules',
+                                           amend=True, to_git=True)
+                        # We don't need to pollute outside world with two save
+                        # results, when the "real" result will be one commit.
+                        continue
+                    else:
+                        # We didn't really commit. Yield and call `subdatasets`
+                        # in that case.
+                        yield r
+                        ds.subdatasets(path,
+                                       set_property=[("datalad-url", source)])
+                        continue
+
+                yield r
 
 
 def clone_dataset(
--- a/datalad/core/distributed/clone.py
+++ b/datalad/core/distributed/clone.py
@@ -308,6 +308,10 @@ class Clone(Interface):
             # -> make submodule
             for r in ds.save(
                     path,
+                    # Note, that here we know we don't save anything but a new
+                    # subdataset. Hence, don't go with default commit message,
+                    # but be more specific.
+                    message="[DATALAD] Added subdataset",
                     return_type='generator',
                     result_filter=None,
                     result_xfm=None,
--- a/datalad/core/distributed/clone.py
+++ b/datalad/core/distributed/clone.py
@@ -306,7 +306,7 @@ class Clone(Interface):
         if ds is not None:
             # we created a dataset in another dataset
             # -> make submodule
-            save_results = []
+            actually_saved_subds = False
             for r in ds.save(
                     path,
                     # Note, that here we know we don't save anything but a new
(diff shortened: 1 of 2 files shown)


mih:
Restarted Appveyor

bpoldrack:
There appears to be a real issue with parallel execution - converting to draft until resolved.

datalad/core/distributed/clone.py
```diff
@@ -308,16 +308,48 @@ def __call__(
...
+                        # We don't need to pollute outside world with two save
+                        # results, when the "real" result will be one commit.
+                        continue
+                    else:
+                        # We didn't really commit. Yield and call `subdatasets`
+                        # in that case.
+                        yield r
+                        ds.subdatasets(path,
+                                       set_property=[("datalad-url", source)])
+                        continue
```
kyleam, line 350: Stepping through a simple `datalad clone -d. /path/to/a b`, the logic under `status=ok` looks correct to me.

The logic under the `else` arm may need some more thought.  For example, it'd be triggered by the `notneeded` record that comes out of a failed attempt to clone a dataset (say `datalad clone -d. http://nothere.datalad.org/abc sub`).  While the `subdatasets` call is a noop there, it seems confusing to do, and in general I wonder if saving via the `subdatasets` call is a good action if we know the `type=dataset, action=save` call had a non-ok status.

A couple of more thoughts:

 * A regression test for the "more than one commit" behavior is probably in order.

  * Relying on result records to decide to amend makes me a little uneasy (mostly in terms of changes upstream having unanticipated effects).  Do you think it's worth adding a `get_hexsha` call pre and post `ds.save(path, ...)` as a sanity check that HEAD moved?

bpoldrack: > While the subdatasets call is a noop there, it seems confusing to do, and in general I wonder if saving via the subdatasets call is a good action if we know the type=dataset, action=save call had a non-ok status.

Agree - I did wonder, too. The `else` is there to maintain previous behavior when there wasn't an `ok` result. However, I'm happy to remove it entirely and say: Either saving did something and we amend-record datalad-url or something went wrong and we don't want to put a commit on top of something broken or no action is needed anyway.

> A regression test for the "more than one commit" behavior is probably in order.

I'll add a test.

> Relying on result records to decide to amend makes me a little uneasy (mostly in terms of changes upstream having unanticipated effects). Do you think it's worth adding a get_hexsha call pre and post ds.save(path, ...) as a sanity check that HEAD moved?

I don't think so. Relying on return values is no special case when it is a result record in my view. We do that everywhere. Plus: If we think, we can't rely on it internally, we should urgently remove the result-hook mechanism. IMO it's the other way around: We need to consider the result dicts as part of the interface that we shouldn't easily break.
Regarding `get_hexsha`: It's more expensive and not that simple either. We need to consider adjusted branches. `save` calls `AnnexRepo.localsync` internally. If we would come from not yet synced state, we would get a different hexsha due the `annex-sync` even if the save itself was `notneeded`, I think.

kyleam: > I don't think so. Relying on return values is no special case when it is a result record in my view. We do that everywhere.

Right, result records are used everywhere.  I think the thing that is making me pause here is a combination of this being an operation that rewrites history and the specific result check (`action=save, type=dataset, status=ok` for each result yielded by `save(path, ...)`).  In the context of this call, I can't think of a future scenario where, say, there would be two `action=save, type=dataset, status=ok` results yielded, and I think risk of breakage is probably pretty low.  I'm just undecided about whether the code should be more defensive given the operation (and am okay with your answer of "no").

> Plus: If we think, we can't rely on it internally, we should urgently remove the result-hook mechanism.

I think it depends on what specific feature of the result records are being relied on.  For example, I'd guess we have had to adjust `assert_result_count` tests enough to know relying on a specific number of records matching a certain set of conditions is prone to become stale.

> IMO it's the other way around: We need to consider the result dicts as part of the interface that we shouldn't easily break.

I don't think anyone disagrees that result dicts should be considered as part of the interface (and this was one of the aspects [discussed][1] when it was introduced).

[1]: https://github.com/datalad/datalad/pull/3903#pullrequestreview-329207283

> Regarding `get_hexsha`: It's more expensive and not that simple either. [...]

Hmm, I haven't hit any issues quickly testing with `./tools/eval_under_testloopfs`, but either way I was proposing "head pre != head post" as an additional guard/check, so `notneeded` would still be passed over as it is now.

> `save` has the `get_hexsha` calls anyway and could include them in the result, which might be useful elsewhere, too.

If it's readily available, going that route sounds fine to me.

But as I mentioned above, I'm on the fence about the additional check in general, and it's okay with me if you decide to leave it as is.

mih (changes requested):
I have mixed feelings about this. Somewhat echoing @kyleam on the reliability of the result records, I'd say a few things:

- it is OK to rely on result records, but that reliance needs to be specific
- it is not reasonable to assume that order and number of records will never change

With that in mind, some observations:

- amending is triggered by a {status=ok, type=dataset, action=save)
- there is no check on `ds` actually being the subject of that record
- further modification of `ds` is done *while* the original `ds.save` could still be running -- the knowledge that the target result is (presently) the last one is implicit
- whenever any of these assumptions change, problems could arise

My request:

- at the very minimum we should check whether the dataset being reported as saved is actually `ds`
- if find it much safer to delay the amend until we know for sure that no further modification of `ds` can take place before we attempt an amend, or other change -- this would be after a `return` of `ds.save`
- the result record inspection can still be used to decided between amend and property change.

mih:
It would be very helpful to have this merged. @bpoldrack can you please have a look at what is left todo? Thx!

datalad/core/distributed/clone.py
```diff
@@ -306,18 +306,48 @@ def __call__(
...
                 yield r
 
             # Modify .gitmodules to contain originally given url. This is
-            # particularly relevant for postclone routines on a later `get` for
-            # that subdataset. See gh-5256.
-            ds.subdatasets(path, set_property=[("datalad-url", source)])
+            # particularly relevant for postclone routines on a later `get`
+            # for that subdataset. See gh-5256.
+            if any(r['action'] == 'save' and r['type'] == 'dataset' and \
+                   r['refds'] == ds.path for r in save_results):
```
mih, line 327: It seems that this is the only place where `save_results` is used. I am not sure how many it could be, but it seems a simpler approach would be to build the boolean indicator that triggers this conditional in the loop above. So something like

```
actually_saved_subds = None
...
actually_saved_subds = actually_saved_subds or r['status'] if \
    r['action'] == 'save' and r['type'] == 'dataset' and r['refds'] == ds.path else None
...
if actually_saved_subds:
    ...
```

This seems leaner and more clear as to why and how information is retained.

datalad/core/distributed/clone.py
```diff
@@ -306,18 +306,48 @@ def __call__(
...
+                if r['status'] == 'ok':
```
mih, line 328:
```suggestion
                if actually_saved_subdataset == 'ok':
```

mih (commented):
I made a proposal to simplify this change a bit. Please have a look.

datalad/core/distributed/clone.py
bpoldrack: Agree.

bpoldrack: Done. With slight variation.

mih:
Thx! Works for me now. Please add a proper title and top comment, so we will end up with something intelligible in the changelog.

bpoldrack:
Like this, @mih?

mih (approved):
Yes, thx!

PR #5481, opened 2021-03-10
Test against released metalad only

bpoldrack:
Alternative proposal for #5479.

datalad-hirni depends on metalad. If we break released metalad in datalad core, I still need to know, as hirni then cannot use both - released metalad and newest datalad.


mih:
Responded in the other PR with why I do not follow the logic completely.

PR #5484, opened 2021-03-11
TST: Add datalad-hirni extension

bpoldrack:
As suggested in PR #5479, test against datalad-hirni, which became more vital to do with metalad being gone from the list for now.


mih (approved):
LGTM

PR #5485, opened 2021-03-11
TST: Make appveyor codecov submission functional across all platforms

mih:
Curious if it will impact the report in this PR...

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -288,12 +288,14 @@ test_script:
 
 
 after_test:
-  # prepare coverage.xml in a separate invocation.
-  # if invoked directly with nose - do not include test_ files themselves
-  - python -m coverage xml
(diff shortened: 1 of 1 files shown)


mih:
Seems to work: https://app.codecov.io/gh/datalad/datalad/compare/5485/changes

yarikoptic:
NICE!!!! I had spent hours trying to figure out awhile back and I guess I have not tried this now obvious thing to try. Not sure why codecov map above has all the red squares though.  For paranoid me, may be worth pushing some additional commit to get another clean thorough run?

mih:
It still took several hours to get here ;-)

I think this run is as clean as it gets. The coverage map is very different from the one shown at the codecov website for this PR -- broken right now, but was working before.

I will merge this now, and we will see how it behaves across the next PRs.

PR #5491, opened 2021-03-12
Leave a log trace which env vars are queried for credentials

mih:
Before one would need to spend 10-15min to discover how exactly this is
working. Now `--log-level 5` will give a hint:

```
[Level 5] Credentials lookup attempt DATALAD_webdav.4shared.com_password
[Level 5] Credentials lookup attempt DATALAD_webdav.4shared.com_user
```

Should not hurt anyone. Isn't beautiful either, but having to override
credentials is a relatively frequent task in development, but also in
environments without credentials stores (HPC...). Knowing the names of
the variables is key here.

### DIFF
--- a/datalad/support/keyring_.py
+++ b/datalad/support/keyring_.py
@@ -48,7 +48,7 @@ class Keyring(object):
     def get(self, name, field):
         # consult environment, might be provided there and should take precedence
         env_var = ('DATALAD_%s_%s' % (name, field)).replace('-', '_')
-        lgr.log(5, 'Credentials lookup attempt %s', env_var)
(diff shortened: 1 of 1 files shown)


datalad/support/keyring_.py
```diff
@@ -46,6 +48,7 @@ def _get_service_name(cls, name):
     def get(self, name, field):
         # consult environment, might be provided there and should take precedence
         env_var = ('DATALAD_%s_%s' % (name, field)).replace('-', '_')
+        lgr.log(5, 'Credentials lookup attempt %s', env_var)
```
yarikoptic, line 51:
```suggestion
        lgr.log(5, 'Credentials lookup attempt via env var %s', env_var)
```
to make it more explicit?

PR #5494, opened 2021-03-13
[DATALAD RUNCMD] BF: use snapshot repo url for git-annex for .deb

yarikoptic:
Decided to use stock debian build instead of ours on datasets.datalad.org
since those are standalone and thus slower

=== Do not change lines below ===
{
 "chain": [],
 "cmd": "sed -i -e s,ftp.us.debian.org/debian/,snapshot.debian.org/archive/debian/20201228T023115Z/,g .appveyor.yml",
 "exit": 0,
 "extra_inputs": [],
 "inputs": [
  ".appveyor.yml"
 ],
 "outputs": [
  ".appveyor.yml"
 ],
 "pwd": "."
}
^^^ Do not change lines above ^^^

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -70,7 +70,7 @@ environment:
       APPVEYOR_BUILD_WORKER_IMAGE: Ubuntu2004
       INSTALL_SYSPKGS: python3-virtualenv
       # system git-annex is way too old, use better one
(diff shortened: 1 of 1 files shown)


yarikoptic:
I think this is a nobrainer and fixes up appveyor runs, so I will merge

Issue #5496, opened 2021-03-14
https://reuse.software/ spec metadata extractor


ypid:
#### What is the problem?

I noticed that you provide a way to specify a license for datasets using [RFC822-compliant metadata](http://docs.datalad.org/en/stable/metadata.html#rfc822-compliant-metadata-datalad-rfc822). I was wondering if this task would not be better offloaded to the dedicated https://reuse.software/ specification and tooling.

#### Have you had any success using DataLad before? (to assess your expertise/prior luck.  We would welcome your testimonial additions to https://github.com/datalad/datalad/wiki/Testimonials as well)

Long time git and git-annex user. My main use-case is reproducible execution (`datalad run`) currently.

yarikoptic:
Isn't https://reuse.software/ only about copyright/licenses, like dep5 of debian, and thus doesn't include any other metadata like description, etc?

ypid:
I fixed the issue title that was to unspecific before. Yes, that is also my suggestion. Only use REUSE for the licensing aspect.

bpoldrack:
At a first glance, addressing this sounds to me more like an additional (and certainly desirable) metadata extractor than a change to the rfc822 extractor.

yarikoptic:
Adding a dedicated extractor for reuse.software would indeed be useful.  But using SPDX license name in the rfc822 would also be useful IMHO.

yarikoptic:
I will consider this resolved by #5497

ypid:
#5497 is just related to this issue. That is why I linked it. In my eyes #5497 can resolve this issue.

yarikoptic:
Ok, adjusted title to match what I think it could be. ATM so metadata extractors are "made equal" pretty much, we don't do metadata harmonization. It is up to a user too decide which one to use or develop some harmonizer. I guess it would be for an interested user/developer working with REUSE specification to develop it.

PR #5497, opened 2021-03-14
DOC: Use SPDX License Identifiers/Expression for License metadata

ypid:
This provides an unambiguous and machine readable way to reference licenses.

I am not sure if this way of specificaing the license is the right way (see https://github.com/datalad/datalad/issues/5496), but switching to SPDX is a quick win.

### DIFF
--- a/datalad/metadata/extractors/tests/test_rfc822.py
+++ b/datalad/metadata/extractors/tests/test_rfc822.py
@@ -24,7 +24,7 @@ Description: Basic summary
  A text with arbitrary length and content that can span multiple
  .
  paragraphs (this is a new one)
-License: CC0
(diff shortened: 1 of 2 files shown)


yarikoptic:
Since it was not regulated before, it might make some existing metadata files "non-compliant" but since extractor doesn't care (yet; we did not version it/its schema) IMHO it is ok.  Ideally (later) it could have even provided validation and issued a warning whenever license field is not following SPDX (but then we would need to fetch/cache recent listing).
@mih - WDYT?

mih:
I have no objections. Thx!

yarikoptic:
appveyor fails unrelated (and afaik addressed since then), so let's proceed!

PR #5506, opened 2021-03-16
BF: AnnexJsonProtocol - memoize the last line which failed to json load

### LINKED ISSUE #5393: AnnexJsonProtocol cannot handle large JSON lines
Due to a coding error I have (erm, it turns out there is no coding error -- see below).

```
[git-annex] % wc -l < uuid.log
34869
```

This file is 3.8M. git-annex in general, and `git annex info --json` in particular, happily works with this, but not DataLad.

```
% python3 -c "import datalad.api as dl; dl.Dataset('.').repo.repo_info()"
...
[ERROR] Received undecodable JSON output
...
  File "/usr/lib/python3/dist-packages/datalad/support/annexrepo.py", line 2460, in repo_info
    assert(len(json_records) == 1)
AssertionError
```

This seems due to the implementation of the `AnnexJsonProtocol`. `git annex info --json` will report all of this information in a single line. With 3.8MB of text, we seem to hit a buffer limit and only attempt to process an incomplete JSON records, and eventually blow when nothing intelligible comes back from the protocol.

yarikoptic:
Protocol's buffer could get filled up before the entire json line is
actually received, so the AnnexJsonProtocol could then receive only a
part of it, and fail to parse such a partial json entry.

The patch is a bit overdone: in principle I think

    not data.endswith(os.linesep.encode())

condition could be dropped, but I think it should not hurt since we are not
expecting new lines in the output within records, thus if the trailing symbol
in the entire data is a newline, it would mean that we did read the entire line
and thus we should not try to join it with the next one.

I am not sure if similar "I care about lines" behavior is relevant to other
WitlessProtocols and might need to be adjusted similarly.

Closes #5393

TODO:
- [x] test (just feed with some large and short json records interleaved and see if all good)

individual commits (adjusted more while crafting the tests) might have more information.

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -3455,7 +3455,7 @@ class AnnexJsonProtocol(WitlessProtocol):
                         # it is the last line and fails to parse -- it can/likely
                         # to happen that it was not a complete line and that buffer
                         # got filled up/provided before the end of line.
-                        # So - memoize it and pass to the next call
+                        # Store it so that it can be prepended to data in the next call.
                         self._unprocessed = line
                         break
                     # TODO turn this into an error result, or put the exception
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -3410,6 +3410,7 @@ class AnnexJsonProtocol(WitlessProtocol):
         super().__init__(done_future)
         self._global_pbar_id = 'annexprogress-{}'.format(id(self))
         self.total_nbytes = total_nbytes
+        self._remainder = None
 
     def connection_made(self, transport):
         super().connection_made(transport)
@@ -3435,17 +3436,30 @@ class AnnexJsonProtocol(WitlessProtocol):
             # let the base class decide what to do with it
             super().pipe_data_received(fd, data)
             return
+        if self._remainder:
+            data = self._remainder + data
+            self._remainder = None
         # this is where the JSON records come in
         # json_loads() is already logging any error, which is OK, because
         # under no circumstances we would expect broken JSON
-        for line in data.splitlines():
+        lines = data.splitlines()
+        for iline, line in enumerate(lines):
             try:
                 j = json_loads(line)
-            except Exception:
-                # TODO turn this into an error result, or put the exception
-                # onto the result future -- needs more thought
+            except Exception as exc:
                 if line.strip():
                     # do not complain on empty lines
+                    if iline == len(lines) - 1 and not data.endswith(os.linesep.encode()):
+                        lgr.debug("Caught %s while trying to parse JSON line %s which might "
+                                  "be not yet a full line", exc, line)
+                        # it is the last line and fails to parse -- it can/likely
+                        # to happen that it was not a complete line and that buffer
+                        # got filled up/provided before the end of line.
+                        # So - memoize it and pass to the next call
+                        self._remainder = line
+                        break
+                    # TODO turn this into an error result, or put the exception
+                    # onto the result future -- needs more thought
                     lgr.error('Received undecodable JSON output: %s', data)
                 continue
             self._proc_json_record(j)
@@ -3579,6 +3593,11 @@ class AnnexJsonProtocol(WitlessProtocol):
                 pbar_id,
                 'Finished',
                 noninteractive_level=5,
+            )
+        if self._remainder:
+            lgr.error(
+                "%d bytes of received undecodable JSON output remain: %s",
+                len(self._remainder), self._remainder
             )
         super().process_exited()
 
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -3460,7 +3460,7 @@ class AnnexJsonProtocol(WitlessProtocol):
                         break
                     # TODO turn this into an error result, or put the exception
                     # onto the result future -- needs more thought
-                    lgr.error('Received undecodable JSON output: %s', data)
+                    lgr.error('Received undecodable JSON output: %s', line)
                 continue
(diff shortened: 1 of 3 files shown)


yarikoptic:
Codecov you can't be right! At least you report diff coverage 100% ;-)

datalad/support/annexrepo.py
```diff
@@ -3435,18 +3436,31 @@ def pipe_data_received(self, fd, data):
...
                 if line.strip():
                     # do not complain on empty lines
-                    lgr.error('Received undecodable JSON output: %s', data)
+                    if iline == len(lines) - 1 and not data.endswith(os.linesep.encode()):
+                        lgr.debug("Caught %s while trying to parse JSON line %s which might "
+                                  "be not yet a full line", exc, line)
+                        # it is the last line and fails to parse -- it can/likely
+                        # to happen that it was not a complete line and that buffer
+                        # got filled up/provided before the end of line.
+                        # So - memoize it and pass to the next call
```
kyleam, line 3458: Perhaps not something that will trip up other readers, but I found "memoize" and "pass" in this context confusing.  Maybe "store it so that it can be prepended to `data` in the next call"?

datalad/support/tests/test_annexrepo.py
```diff
@@ -2194,6 +2199,60 @@ def test_annexjson_protocol(path):
...
+    seed = _seed if _seed is not None else random.randint(0, 1000)
+    # Inform for ease of debugging.
+    # TODO: if other test like this appears, craft @with_rng helper
+    print("RNG seed: %s" % seed)
+    rng = random.Random(seed)
+    records = [
+        {('field%d' % f): ("a" * rng.randint(0, 5000))
+             for f in range(rng.randint(1, 100))
+         }
+        for i in range(10)
```
kyleam, line 2215: I think it'd be better to avoid using random lengths here.  Without the annexrepo changes from this PR, this test failed for me only 6 out of the 10 times I ran it.

It looks like `Lib.asyncio.unix_events._UnixReadPipeTransport` reads with a max size of `256 * 1024`.  So perhaps something like this instead:

```diff
diff --git a/datalad/support/tests/test_annexrepo.py b/datalad/support/tests/test_annexrepo.py
index d04715122c..a5b7fedc75 100644
--- a/datalad/support/tests/test_annexrepo.py
+++ b/datalad/support/tests/test_annexrepo.py
@@ -2201,18 +2201,11 @@ def test_annexjson_protocol(path):

 @with_tempfile
 def test_annexjson_protocol_long(path):
-    # Inform about the seed here and use dedicated RNG so we are not
-    # effected by any other use of random before this test
-    seed = _seed if _seed is not None else random.randint(0, 1000)
-    # Inform for ease of debugging.
-    # TODO: if other test like this appears, craft @with_rng helper
-    print("RNG seed: %s" % seed)
-    rng = random.Random(seed)
     records = [
-        {('field%d' % f): ("a" * rng.randint(0, 5000))
-             for f in range(rng.randint(1, 100))
-         }
-        for i in range(10)
+        {"k": "v" * 20},
+        # Value based off of
+        # Lib.asyncio.unix_events._UnixReadPipeTransport.max_size.
+        {"k": "v" * 256 * 1024}
     ]
     with open(path, 'w') as f:
         for record in records:
```

datalad/support/tests/test_annexrepo.py
```diff
@@ -2194,6 +2199,60 @@ def test_annexjson_protocol(path):
...
+        [sys.executable, path],
+        protocol=AnnexJsonProtocol
+    )
+    eq_(res['stdout'], '')
+    eq_(res['stderr'], '')
+    eq_(res['stdout_json'], records)
+
+
+def test_annexjson_protocol_incorrect():
+    yield check_annexjson_protocol_incorrect, ''
```
kyleam, line 2231: Mostly just a note for myself: This is a test of existing behavior; it passes before and after the annexrepo changes here.

kyleam (approved):
These changes look good to me.  I've left a suggestion for `test_annexjson_protocol_long`, though I think it's okay as is.

> I am not sure if similar "I care about lines" behavior is relevant to other
> WitlessProtocols and might need to be adjusted similarly.

`GitProgress` already has similar logic to what you're adding here.  Checking the other protocol classes, I don't spot any remaining ones that need to be adjusted.

yarikoptic:
> `GitProgress` already has similar logic to what you're adding here.

thanks!  I will call attribute `_unprocessed` to match.  I will not try to unify common logic at this point, but 2 is a proof there might eventually be 3 ;)

datalad/support/tests/test_annexrepo.py
yarikoptic: ok, will take yours but will compliment with bigger and shorter ones... just to make sure

kyleam: > just to make sure

Of course that's fine, though I'm not quite getting how that'd make you more sure than reverting your changes and seeing that it does reliably fail without the changes from this PR (as opposed to the, say, 6/10 I saw with the current version of the test).  IIRC from my local testing, the tests start to pass without the `AnnexJsonProtocol` changes from this PR at `(256 * 1024) - 9` (with 9 coming from the json quotes, etc).

kyleam (approved):
Thanks for the updates.  LGTM.

yarikoptic:
so codecov is consistently off.  to different degree but similar in nature in other PRs (https://github.com/datalad/datalad/pull/5505).

I guess coverage in `maint` was dropped in PRs with some tests/code been run now only in CRON jobs, so it is doomed to report under-coverage all the time and keep all PRs red. heh.

PR #5509, opened 2021-03-18
Test, documentation, fixes and public introduction of parse_gitconfig_dump()

### LINKED ISSUE #5502: git-config produces output that _parse_gitconfig_dump cannot handle
This is still a mystery, but this issue is about some aspect that might trigger a `get` failure that was reported to me. The environment is an HPC system with GPFS. Correction: the dataset is located in `/dev/shm/...` (tmpfs), while the user HOME is on GPFS.

```
$ datalad get -n inputs/ukb/sub-3192110
[WARNING] Skip unrecognized .gitmodule specification: rm: cannot remove '/p/home/users/.../.cache/git-annex/locales/97d8e8e226139eee06ab02ccdf2f4c84': Directory not empty=submodule.code/pipeline.path 
[WARNING] Failed to get 'code/pipeline.path', skipping this submodule
```

The specific composition of this warning message makes me suspect that 

```
        out = self.call_git(
            ['config', '-z', '-l', '--file', '.gitmodules'],
            read_only=True)
 ```

in `get.py:_parse_gitmodules()` produces output 

```
rm: cannot remove '/p/home/users/.../.cache/git-annex/locales/97d8e8e226139eee06ab02ccdf2f4c84': Directory not empty
```

That `_parse_gitconfig_dump()` cannot handle, as a result gets mistaken for a misconfiguration, subsequently leading to critical submodule information being ignores, and ultimately failing the entire operation.

### PR DESCRIPTION
**Pull request by mih**
This PR (in order of importance):

- fixes handling of multi-line values (all but the first line were ignored before)
- introduces a unit test for this function
- adds more documentation
- robustifies the parser to be able to work with contaminated output (fixes #5502)

Although this series slightly complicates the processing in this critical function, I do not observe a performance impact (see commit message).

### DIFF
--- a/datalad/config.py
+++ b/datalad/config.py
@@ -131,16 +131,9 @@ def parse_gitconfig_dump(dump, cwd=None, multi_value=True):
                 break
             # try getting key/value pair from the present chunk
             k, v = _gitcfg_rec_to_keyvalue(line)
-            if v or cfg_k_regex.match(k):
-                # if we have a value, we also must have a good key
-                # otherwise check the key
+            if k is not None:
                 # we are done with this chunk when there is a good key
                 break
-            else:
-                # no value, no good key, this chunk is not usable as is
-                # reset the key to cleanly discard the entire chunk
-                # should it not contain anything usable
-                k = None
             # discard the first line and start over
             ignore, line = line.split('\n', maxsplit=1)
             lgr.debug('Non-standard git-config output, ignoring: %s', ignore)
@@ -174,15 +167,19 @@ def _gitcfg_rec_to_keyvalue(rec):
     Returns
     -------
     str, str
-      Parsed key and value. Value could be None.
+      Parsed key and value. Key and/or value could be None
+      if not snytax-compliant (former) or absent (latter).
     """
     kv_match = cfg_kv_regex.match(rec)
     if kv_match:
         k, v = kv_match.groups()
-    else:
+    elif cfg_k_regex.match(rec):
         # could be just a key without = value, which git treats as True
         # if asked for a bool
         k, v = rec, None
+    else:
+        # no value, no good key
+        k = v = None
     return k, v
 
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: datalad/config.py [hunk 1]
```diff
@@ -71,44 +79,75 @@ def _where_reload(obj):
...
+                if not fname.is_absolute():
+                    fname = Path(cwd) / fname if cwd else Path.cwd() / fname
+                fileset.add(fname)
+                break
+            if line.startswith('command line:'):
+                # no origin that we could as a pathobj
+                break
+            # try getting key/value pair from the present chunk
+            k, v = _gitcfg_rec_to_keyvalue(line)
+            if v or cfg_k_regex.match(k):
```
bpoldrack, line 134: Shouldn't the `cfg_k_regex.match(k)` happen within `_gitcfg_rec_to_keyvalue` instead?
I find it a bit weird to have the result of the latter be checked like that afterwards. Internally, that function already makes the assumption that it must be a key-only line, so just check it within ...

bpoldrack (approved):
LGTM. Left a minor wish.

datalad/config.py
mih: The expression used in `_gitcfg_rec_to_keyvalue()` contains `cfg_k_regex`, so what you seem to ask for is already the case.

bpoldrack: Then there would be no need to test here again. See below.

datalad/config.py
```diff
@@ -120,6 +159,33 @@ def _parse_gitconfig_dump(dump, cwd=None, multi_value=True):
...
+    Returns
+    -------
+    str, str
+      Parsed key and value. Value could be None.
+    """
+    kv_match = cfg_kv_regex.match(rec)
+    if kv_match:
+        k, v = kv_match.groups()
+    else:
+        # could be just a key without = value, which git treats as True
```
bpoldrack, line 183: This is the branch resulting in an additional test outside, no, @mih?

datalad/config.py
mih: Ah, I get it now. Thx for the explanation below.

datalad/config.py
mih: Fixed. thx

PR #5515, opened 2021-03-18
OPT: Avoid unconditional string expansion in log messages

mih:
Use the loggers ability to delay expansion till a message is actually
being rendered.

Some messages are generated so frequently, it may even impact the performance.

### DIFF
--- a/datalad/support/network.py
+++ b/datalad/support/network.py
@@ -334,7 +334,7 @@ def _guess_ri_cls(ri):
     # We assume that it is a URL and parse it. Depending on the result
     # we might decide that it was something else ;)
     fields = URL._pr_to_fields(urlparse(ri))
-    lgr.log(5, "Parsed ri %s into fields %s" % (ri, fields))
+    lgr.log(5, "Parsed ri %s into fields %s", ri, fields)
     type_ = 'url'
     # Special treatments
     # file:///path should stay file:
@@ -347,7 +347,7 @@ def _guess_ri_cls(ri):
         # dl+archive:... or just for ssh   hostname:path/p1
         elif '+' not in fields['scheme']:
             type_ = 'ssh'
-            lgr.log(5, "Assuming ssh style ri, adjusted: %s" % (fields,))
+            lgr.log(5, "Assuming ssh style ri, adjusted: %s", (fields,))
 
     if not fields['scheme'] and not fields['hostname']:
         parts = _split_colon(ri)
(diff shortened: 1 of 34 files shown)


bpoldrack:
> Some messages are generated so frequently, it may even impact the performance.

I think this aspect of logging is underappreciated! Not completely solved here, though. The only thing delayed should be the (implicit) conversion to `str`. If we do something like `lgr.debug("%s", function(some))`, `function` is still called.
The real impact would be to pass callables, that are then called when the message is actually rendered. Not sure how to do this, though.

bpoldrack:
The only thing coming to my mind is a dedicated class, that overrides `__str__`:

```

import logging
import datalad.api
import time
from functools import partial

lgr = logging.getLogger('datalad.test_logger')


def test_func(value):

    print("test_func executed with %s" % value")
    return value


class ToBeLogged(object):

    def __init__(self, value):
        self.value = value

    @property
    def logged(self):

        return test_func(self.value)

    def __str__(self):

        return str(self.logged)


def log_things():

    lgr.debug("pass function call: %s", test_func(1))
    obj = ToBeLogged(2)
    lgr.debug("pass property: %s", obj.logged)
    obj = ToBeLogged(3)
    lgr.debug("delay: %s", obj)


log_things()

```

=> 

```
❱ DATALAD_LOG_LEVEL=DEBUG python test-log.py
...
test_func executed with 1
[DEBUG  ] pass function call: 1 
test_func executed with 2
[DEBUG  ] pass property: 2 
test_func executed with 3
[DEBUG  ] delay: 3 
```

```
❱ DATALAD_LOG_LEVEL=INFO python test-log.py 
test_func executed with 1
test_func executed with 2
```

datalad/support/network.py
```diff
@@ -347,7 +347,7 @@ def _guess_ri_cls(ri):
         # dl+archive:... or just for ssh   hostname:path/p1
         elif '+' not in fields['scheme']:
             type_ = 'ssh'
-            lgr.log(5, "Assuming ssh style ri, adjusted: %s" % (fields,))
+            lgr.log(5, "Assuming ssh style ri, adjusted: %s", (fields,))
```
kyleam, line 350: Doesn't functionally matter and perhaps not even worth a fixup commit, but this is introducing extra brackets.

kyleam (approved):
I believe we've been pretty good about not introducing new instances of pre-formatted log messages, and I'm glad to see a tree-wide cleanup.  I'm not sure I prevented my eyes from glazing over while going through this, but all of the conversions except for one look correct to me (edit: of course the tests disagree).

datalad/support/network.py
mih: Done

mih:
Thx for the reviews!

Issue #5513, opened 2021-03-18
datalad/support/configparserinc.py contains PY3.2 compatibility stuff


mih:
Contains PY3.2 compatibility stuff, which could be removed.

Overall it would be nice to get rid of this entire file. The `SafeConfigParserWithIncludes` class is only used by `datalad/downloaders/providers,py`. Do we actually need that additional functionality for providers @yarikoptic ?

PR #5516, opened 2021-03-18
RM: SafeConfigParserWithIncludes

mih:
I could not determine where and how the functionality added by this
class over `configparser.ConfigParser` is needed. It used to be used
in datalad.config, but that was loooong ago. This change simply
removes the class (no tests to remove, because there are none), and
replaces its usage with `configparser.ConfigParser`.

Closes gh-5513


bpoldrack:
Crawler! ;-)

yarikoptic:
yeap... the cradle of the datalad... and it stems from fail2ban where I hacked it in previous lifetime.  I am not sure though if even current crawler needs any added functionality (I do remember that in inception there were need to be able to include config files within config files), so would need a look at the crawler and possibly a simple RF and release.

mih:
If it is just the crawler, we could move the class, if needed.

mih:
OK, so the crawler is affected, but only the crawler.

mih:
I created companion PR https://github.com/datalad/datalad-crawler/pull/93 to see what action is best.

mih:
Crawler tests suggest that the crawler would work fine without this class.

yarikoptic:
FWIW, pushed 0.8.2 of the crawler out, thanks @mih for PR

yarikoptic:
somehow conflicts came about

mih:
Resolved conflict (modification of the file deleted here).

mih:
This seems good to go now. I would merge unless objections are voiced.

PR #5369, opened 2021-01-25
Demonstrating an issue reported by @djarecka (TODO: address)

yarikoptic:
Sits on top of #5367

troubleshooting https://github.com/ReproNim/testkraken/issues/93 -- while everyone is asleep, want to see where would fail.  Passes on Linux, fails on OSX (reproduced locally).

Ideally should be addressed for 0.14.0 but not "mandatory"

Issue #4539, opened 2020-05-15
Introduce support for subcommand interfaces


yarikoptic:
#### What is the problem?

This is an old issue for which I did not find an open issue in a quick search.  Our `Interface`s API generation goes only one level -- we can define commands but not subcommands which would have possibly additional different from each other options.  That manifests itself in having no uniform API and thus at times consistent `--help` across commands to support what we desire. We ended up with

- different `create-sibling*`, `export*` flavors
-  `siblings [OPTIONS] [COMMAND]` so that some OPTIONS is superset across all COMMANDs
- `search [--mode MODE] QUERY` with search API of specific `MODE` implementations accepting and swallowing options it does not care about, and then incorporating some `MODE` specific options within `QUERY`
- `metadata`, `aggregate-metadata`, `extract-metadata`, so if command completion worked, it requires reading documentation or `--help` to figure what commands operate with metadata (unlike if it is `metadata`, `metadata aggregate`, ... to also be inline with `siblings COMMAND` etc 

I think that eventually we should mitigate this situation and come up with some solution for consistent API. Existing approach in `click` allows to define and nest  groups of commands: https://click.palletsprojects.com/en/7.x/commands/ . And I think we might want to do something similar.   On Python level subcommands could be just methods of the parent command e.g. `siblings.create`.

May be (but not really necessarily) while at it we could also completely RF the approach on how we are defining interfaces to avoid IMHO somewhat non-Pythonic groupping via classes with static `__call__`.

I first wanted to seek opinions and feedback before possibly even trying to come up with a solution.

kyleam:
[ Related previous issues include gh-2729 and gh-3276 ]

My opinion is that at this point tree-wide subcommand restructuring is not worth the churn, necessary compatibility kludges, and end-user pain that would come with it.

yarikoptic:
Thanks for the references @kyleam!

> at this point

do you have in mind upcoming 0.13 release, or at this point of the entire project mature age?

kyleam:
> do you have in mind upcoming 0.13 release, or at this point of the entire project mature age?

The latter.

mih:
I am with @kyleam on the restructuring of existing commands. But having the ability to do this in the future (and especially for extensions) would be very valuable.

yarikoptic:
[...]

PR #5517, opened 2021-03-18
NF: foreach-dataset

### LINKED ISSUE #749: [wishlist] foreach command
thin proxy around `git submodule foreach`   but with an option to not fail entirely if one submodule fails

Additional features which might be added
- [ ] option to "filter" datasets, e.g. `--filter dirty|modified|git|git-annex`.  Not sure yet on how/if bool logic between those should be implemented, and/or such filtering could be relatively easily implemented differently (piping in cli, looping over a subdatasets generator in python)

yarikoptic:
Closes #749 

- amalgamation of `subdatasets` and `run`
- supports parallelization (threads, so might be not as efficient for "Python" code) via `-J|--jobs` using the same ConsumerProducer construct as the rest
- supports running external commands or eval'ing Python statements (the same placeholders passed in)
- different from `subdatasets` in that it would also operate on the current (`-d`) refds unless `--subdatasets-only`:
  - reason: while using `git submodules` as quick and dirty workaround I more often wanted to include the top dataset than not
- different from `run` in that output is swallowed (and passed in yielded records + shortened in `message`) unless `-p|--passthrough` is given:  
  - reason: especially in parallel mode dumping output to screen isn't very useful

Potential future enhancements:
- custom renderer, e.g. groupping outputs even if received in parallel (thus should not be printed in `throughput` mode)

Examples: well -- go wild with anything you would like to throw at it... e.g. could be used as a temp "parallelization" helper for commands which do not yet have parallelization "built-in".  Use `--bottomup` where you need to go from the bottom, and should still be safe in that super-dataset is not considered until all subdatasets are handled.

TODOs:
- [x] seek and account for feedback
   - [x] add support for `cmd`  to be a python function?
- [x] decide on what to do about chdir'ing for python eval/exec since cannot do in parallel: see "can't chpwd for Python eval/exec when parallelizing using threads"
- [x] expand DOC
- [x] address TODO for avoiding log progress if passthrough
- [x] address or remove other TODOs in the diff
- ? (never enough!) add tests
- [x] Address failures
   - [x] "I/O operation on closed file" -- we can't do capture in parallel threads due to use of thread-unsafe swallow_outputs
- [x] doc: add foreach to a toctree
- ? thread safe output by the runner protocol in pass-through and parallel mode? (or should not bother?) see NB in #5641 
- Changes required or desired for this functionality
  - ~~#5640 (required to make --fulfilled to default to True)~~ - made it into non-bool `--state {present|absent|any}`
  - #5642 - just desired
- **later-maybe**: expose hardcoded protection during the traversal. Could be `--safe-to-consume=auto,all-subds-done,superds-done,always` (name  with `auto` choosing between `all-subds-done` and `superds-done` depending on traversal order) and `always` to not do any protection.

Fails

<details>
<summary>I/O operation on closed file. </summary> 

```shell
$> datalad foreach -d ~/datalad/openneuro --fulfilled true --cmd-type exec  -J12 'import os; same=ds.path == os.getcwd(); print(f"{same}  {ds.path}: {os.getcwd()}")'
foreach(ok): /home/yoh/datalad/openneuro (dataset) [{'stdout': 'True  /home/yoh/datalad/openneuro: /home/yoh/datalad/openneuro\n', 'stderr': ''}]                                                                    
[ERROR  ] I/O operation on closed file. [dialog.py:message:78] (ValueError)                                                                                                                                          
foreach:  17%|██████████████████████████▏ 
```
</details>

<details>
<summary>can't chpwd for Python eval/exec when parallelizing using threads</summary> 

```shell
(git)lena:~datalad/datalad-master[nf-foreach]git
$> datalad foreach -d ~/datalad/openneuro --fulfilled true --cmd-type exec -J12 --s-o pass-through 'import os, random, time; time.sleep(random.randint(0, 10)/10.); same=ds.path == os.getcwd(); print(f"{same}  {ds.path}: {os.getcwd()}")' | cat
True  /home/yoh/datalad/openneuro: /home/yoh/datalad/openneuro
foreach(ok): /home/yoh/datalad/openneuro (dataset)
True  /home/yoh/datalad/openneuro/ds000006: /home/yoh/datalad/openneuro/ds000006
foreach(ok): /home/yoh/datalad/openneuro/ds000006 (dataset)
False  /home/yoh/datalad/openneuro/ds000003: /home/yoh/datalad/openneuro/ds000009
foreach(ok): /home/yoh/datalad/openneuro/ds000003 (dataset)
False  /home/yoh/datalad/openneuro/ds000105: /home/yoh/proj/datalad/datalad-master
foreach(ok): /home/yoh/datalad/openneuro/ds000105 (dataset)
False  /home/yoh/datalad/openneuro/ds000009: /home/yoh/datalad/openneuro/ds000008
foreach(ok): /home/yoh/datalad/openneuro/ds000009 (dataset)
False  /home/yoh/datalad/openneuro/ds000002: /home/yoh/datalad/openneuro/ds003027
False  /home/yoh/datalad/openneuro/ds003027: /home/yoh/datalad/openneuro/ds000005
foreach(ok): /home/yoh/datalad/openneuro/ds000002 (dataset)
foreach(ok): /home/yoh/datalad/openneuro/ds003027 (dataset)
False  /home/yoh/datalad/openneuro/ds000005: /home/yoh/datalad/openneuro/ds000007
False  /home/yoh/datalad/openneuro/ds000008: /home/yoh/proj/datalad/datalad-master
foreach(ok): /home/yoh/datalad/openneuro/ds000005 (dataset)
foreach(ok): /home/yoh/datalad/openneuro/ds000008 (dataset)
False  /home/yoh/datalad/openneuro/ds001785: /home/yoh/datalad/openneuro/ds000001
foreach(ok): /home/yoh/datalad/openneuro/ds001785 (dataset)
False  /home/yoh/datalad/openneuro/ds000001: /home/yoh/datalad/openneuro/ds000105
False  /home/yoh/datalad/openneuro/ds000007: /home/yoh/datalad/openneuro/ds000002
foreach(ok): /home/yoh/datalad/openneuro/ds000001 (dataset)
foreach(ok): /home/yoh/datalad/openneuro/ds000007 (dataset)
```
</details>
- [ ] fix up test in -deprecated where `published` was moved

<details>
<summary>relevant diff </summary> 

```shell
$> git diff master...HEAD -- datalad/distribution/tests/test_publish.py
diff --git a/datalad/distribution/tests/test_publish.py b/datalad/distribution/tests/test_publish.py
index b6dfecc98..6949bbc58 100644
--- a/datalad/distribution/tests/test_publish.py
+++ b/datalad/distribution/tests/test_publish.py
@@ -88,7 +88,7 @@ def test_invalid_call(origin, tdir):
     dummy = Dataset(tdir).create()
     dummy_sub = dummy.create('sub')
     dummy_sub.uninstall()
-    assert_in('sub', dummy.subdatasets(fulfilled=False, result_xfm='relpaths'))
+    assert_in('sub', dummy.subdatasets(state='absent', result_xfm='relpaths'))
     # now an explicit call to publish the unavailable subdataset
     assert_result_count(
         dummy.publish('sub', on_failure='ignore'),

```
</details>


datalad/local/foreach.py
```diff
@@ -0,0 +1,275 @@
...
+
+lgr = logging.getLogger('datalad.local.foreach')
+
+
+@build_doc
+class ForEach(Interface):
+    r"""Run a command on the dataset and/or each of its sub-datasets.
+
+    Each dataset's repo runner is reused to execute the commands, so they
+    are are executed in the top directory of a corresponding dataset.
```
kyleam, line 66:
```suggestion
    are executed in the top directory of a corresponding dataset.
```

datalad/local/foreach.py
```diff
@@ -0,0 +1,275 @@
...
+            args=("cmd",),
+            nargs=REMAINDER,
+            metavar='COMMAND',
+            doc="""command for execution. A leading '--' can be used to
+            disambiguate this command from the preceding options to DataLad."""),
+        python=Parameter(
+            args=("--python",),
+            action="store_true",
+            doc="""if given, must be a boolean flag indicating whether
+            the `cmd` is Python expressions to be evaluated, instead of
```
kyleam, line 87:
```suggestion
            the `cmd` is a Python expression to be evaluated, instead of
```

datalad/local/foreach.py
```diff
@@ -0,0 +1,275 @@
...
+        # Following options are taken from subdatasets
+        dataset=Parameter(
+            args=("-d", "--dataset"),
+            doc="""specify the dataset to operate on.  If
+            no dataset is given, an attempt is made to identify the dataset
+            based on the input and/or the current working directory""",
+            constraints=EnsureDataset() | EnsureNone()),
+        # TODO: assume True since in 99% of use cases we just want to operate
+        #  on present subdatasets?  but having it explicit could be good to
+        #  "not miss any".  But `--fulfilled true` is getting on my nerves
```
kyleam, line 98: Hmm, yeah, I see both points but would lean toward a true default.

datalad/local/foreach.py
```diff
@@ -0,0 +1,275 @@
...
+            doc="""if given, must be a boolean flag indicating whether
+            to report either only locally present or absent datasets.
+            By default subdatasets are reported regardless of their
+            status""",
+            constraints=EnsureBool() | EnsureNone()),
+        recursive=recursion_flag,
+        recursion_limit=recursion_limit,
+        # does without any outputs handling
+        # Not sure yet if worth supporting here
+        # contains=Parameter(
```
kyleam, line 110: It seems like it would be useful, but of course it's easy to tack on later if you're not sure.

datalad/local/foreach.py
```diff
@@ -0,0 +1,275 @@
...
+                yield dict(
+                    status_rec,
+                    status="impossible",
+                    message="not installed"
+                )
+                return
+            # For consistent environment (Python) and formatting (command) similar to `run` one
+            # But for Python command we provide actual ds and refds not paths
+            placeholders = dict(
+                pwd=pwd,
```
kyleam, line 204: At least for the non-python command, I'd find it confusing that `pwd` doesn't correspond to the cwd for execution.

datalad/local/foreach.py
```diff
@@ -0,0 +1,275 @@
...
+                ds=ds if python else ds.path,
```
kyleam, line 205: For consistency with `run`, perhaps it's worth adding `ds_path` too?

datalad/local/foreach.py
```diff
@@ -0,0 +1,275 @@
...
+                        status_rec.update(out)
+                        # provide some feedback to user in default rendering
+                        if any(out.values()):
+                            status_rec['message'] = shortened_repr(out, 100)
+                status_rec['status'] = 'ok'
+            except Exception as exc:
+                # TODO: option to not swallow but reraise!
+                status_rec['status'] = 'error'
+                # TODO: there must be a better place for it since this one is not
+                # output by -f json_pp ... a feature or a bug???
```
kyleam, line 235: It's intentional:

https://github.com/datalad/datalad/blob/3f5b53c0de72fa826c0cb4d2ef0acc86dd19aa28/datalad/interface/utils.py#L633-L635

`get_status_dict` would put this in "exception_traceback".

kyleam (commented):
My initial reaction is that this is neat, and that it offers useful functionality that goes beyond `submodule foreach`.

datalad/local/foreach.py
```diff
@@ -0,0 +1,275 @@
...
+            recursive=False,
+            recursion_limit=None,
+            # contains=None,
+            bottomup=False,
+            subdatasets_only=False,
+            passthrough=False,
+            jobs=None
+            ):
+        if not cmd:
+            lgr.warning("No command given")
```
bpoldrack, line 160: I think we have `InsufficientArgumentsError` or sth for that.

bpoldrack (commented):
First impression: I like! :)

datalad/local/foreach.py
yarikoptic: yeah, I might as well add it, thank you!

datalad/local/foreach.py
yarikoptic: so in Python case it would differ (`ds` would be instance and `ds_path` a string) while in CMD they would be identical, right?

kyleam: Yes, that was my thinking (though maybe that'd be confusing?)

datalad/local/foreach.py
mih: If the default would not be `True`, any given command would need to check if presence of a dataset.

datalad/local/foreach.py
mih: For large datasets (think UKbiobank), this could make all the difference in terms of performance.

datalad/local/foreach.py
```diff
@@ -0,0 +1,275 @@
...
+class ForEach(Interface):
+    r"""Run a command on the dataset and/or each of its sub-datasets.
+
+    Each dataset's repo runner is reused to execute the commands, so they
+    are executed in the top directory of a corresponding dataset.
+    In contrast, Python commands (expressions) are evaluated without changing directory.
+    TODO: unify via explicit --chdir? but not sure if worth it
+
+    WARNING: Python expressions are `eval`ed within the scope of this process
+    and provided existing objects.
```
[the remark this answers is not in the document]
mih: Update: I now saw that `eval()` is called with `globals`, so I don't really understand the warning anymore.

mih (changes requested):
I like! I also left a few comments ;-)

datalad/local/foreach.py
yarikoptic: yeah, I think I am of the same opinion now -- will make them consistently `chpwd` although (later) we might like to add an option `--chpwd=True|False` to facilitate possible cases where it is desired to operate from current directory?

datalad/local/foreach.py
yarikoptic: stroke me (didn't test/check): can't we have `{ds}` placeholder with `{ds.path}` etc (by `.format` accessing that `ds` attributes) in `run` etc?  then if we add `{ds}` (and `{refds}`) to `run` I would just may be drop `ds_path` here completely so people could just use `{ds.path}` if so desired regardless if it is CLI command or python eval.

datalad/interface/__init__.py
```diff
@@ -92,6 +92,7 @@
         ('datalad.distribution.siblings', 'Siblings', 'siblings'),
         ('datalad.support.sshrun', 'SSHRun', 'sshrun'),
         ('datalad.local.subdatasets', 'Subdatasets', 'subdatasets'),
+        ('datalad.local.foreach', 'ForEach', 'foreach'),
```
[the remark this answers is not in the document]
yarikoptic: indeed, we could workout `--entity-type={dataset,file,sibling,branch}`! For Python mode might want to establish helper classes to encapsulate those notions or otherwise just be records returned by e.g. `status`, `siblings` (and ?) correspondingly?  
actually -- all of those would be related to `ds` and `refds` anyways, so it might be just a matter of adding extra fields ?

As for command(s) name -- I am still hopeful that we eventually implement and RF to https://github.com/datalad/datalad/issues/4539 , and then it would be `dataset foreach`, `sibling foreach` etc (and I guess get `file` and `branch` top level commands)

datalad/local/foreach.py
yarikoptic: I think I misunderstood/misplaced my comment -- I thought about actual directory under which we run.  @mih raised comment about that [below](https://github.com/datalad/datalad/pull/5517#discussion_r598808830) so I will move my answer there

datalad/local/foreach.py
```diff
@@ -0,0 +1,275 @@
...
+
+
+@build_doc
+class ForEach(Interface):
+    r"""Run a command on the dataset and/or each of its sub-datasets.
+
+    Each dataset's repo runner is reused to execute the commands, so they
+    are executed in the top directory of a corresponding dataset.
+    In contrast, Python commands (expressions) are evaluated without changing directory.
+    TODO: unify via explicit --chdir? but not sure if worth it
```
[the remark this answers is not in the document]
yarikoptic: not sure what "normal rule" here is but yeah, I think I am of the same opinion now -- I will make them consistently `chpwd` . Later we might like to add an option `--chpwd=True|False` to facilitate possible cases where it is desired to operate from current directory.

datalad/local/foreach.py
kyleam: note: I used the wrong name above.  It's `dspath` for `run`.  

We could expose add `ds=` to `run`, and use `ds.path`.  I don't think we should, though, unless there's some other attribute of `ds` that we're actually interested in exposing.

datalad/local/foreach.py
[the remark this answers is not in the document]
yarikoptic: unlike with external execution, I thought you could potentially alter the provided instance (`ds`) or just inflict some other change which would persist in the same process (e.g. whenever `foreach` used through Python API)...  I guess situation is not much different from having a command doing changes or evil externally but still feels that it is worth noting somehow that  we are just evaluating in the same process.  I guess it could be `FYI:` or just `Note:` instead of a `WARNING:`

datalad/local/foreach.py
```diff
@@ -0,0 +1,275 @@
...
+            nargs=REMAINDER,
+            metavar='COMMAND',
+            doc="""command for execution. A leading '--' can be used to
+            disambiguate this command from the preceding options to DataLad."""),
+        python=Parameter(
+            args=("--python",),
+            action="store_true",
+            doc="""if given, must be a boolean flag indicating whether
+            the `cmd` is a Python expression to be evaluated, instead of
+            an external command to be executed"""),
```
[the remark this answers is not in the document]
yarikoptic: by name change you mean `--python` -> `--eval` or `refds` -> `rootds`, or both?

now I am thinking that may be we would come up with additional "ways " or types of "CMD", so may be worthwhile right away to RF `--python` --> `--cmd-type={external,eval}`?  in Python API it could even be `--cmd-type={auto,external,eval,call}` if `CMD` is a callable function to which we would pass all those variables as kwargs, and `auto` would choose `external` if str, and `call` if callable is provided.

datalad/local/foreach.py
```diff
@@ -0,0 +1,275 @@
...
+        # TODO: should we just introduce --lower-recursion-limit aka --mindepth of find?
+        subdatasets_only=Parameter(
+            args=("-s", "--subdatasets-only"),
+            action="store_true",
+            doc="""whether to exclude top level dataset."""),
+        passthrough=Parameter(  # TODO  could be of use for `run` as well
+            args=("-p", "--passthrough"),
+            action="store_true",
+            doc="""For command line commands, pass-through their output to 
+            the screen instead of capturing/returning as part of the result records."""),
```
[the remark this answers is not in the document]
yarikoptic: more of `={return,passthrough}`... with a stretch, I could indeed come up with a point to have a `log` for another possible mode.
I am a bit uncertain about `--output` for the name though.  I think a similar (well - identical) option could eventually be added to `run/rerun` (so many times I wanted that but always needs to come up with a shim... uff), and there we have `--output` to point to output file(s). So I would prefer to avoid `--output` here. Could be `--outputs` (since stdout and stderr) but still would be too close/too confusing. May be `--standard-outputs|-o` ?

datalad/local/foreach.py
```diff
@@ -0,0 +1,275 @@
...
+    """
+    # TODO:     _examples_ = [], # see e.g. run
+
+    _params_ = dict(
+        cmd=Parameter(
+            args=("cmd",),
+            nargs=REMAINDER,
+            metavar='COMMAND',
+            doc="""command for execution. A leading '--' can be used to
+            disambiguate this command from the preceding options to DataLad."""),
```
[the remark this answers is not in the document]
yarikoptic: yes... I probably is missing the point of the comment.

datalad/local/foreach.py
yarikoptic: I do not have an immediate pragmatic use case for `run` as well ATM, but I can see how the ability to access `{ds.pathobj}` for cross platform paths specifications etc or `{ds.config}` for configuration variables (which `git config` would not be sufficient due to not be aware of all additional locations) would be of benefit at least for `foreach`.  
I just do not see why then `run` should not be allowed to do the same (i.e., any possible side-effects?) and thus reduce user cognitive load in remembering which placeholders are available in `run/rerun` and which in `foreach`.

kyleam: If there's not an immediate use case, I'd rather not add it.  When you're dealing with `run` commands, you're talking about supporting things indefinitely.  I don't think it's a good place to add things just because.

datalad/local/foreach.py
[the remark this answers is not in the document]
yarikoptic: got an immediate other use-case, `exec` so it is possible if so desired to provide a bit more complete python source code with imports etc, so I will do the `--cmd-type`

datalad/local/foreach.py
yarikoptic: ok, "enabling" the --contains option

datalad/local/foreach.py
yarikoptic: @kyleam re `pwd`.  
- I do not see a value ATM to make `pwd` correspond to cwd for execution since that one would be identical to `{ds.path}`
- I think it is good to have it consistent with `run` which seems to have it to correspond to the current directory from which `run` is initiated: ""{pwd}" will be replaced with the full path of the current working directory. ":
```shell
(git)lena:~datalad/datalad-master[nf-foreach]git
$> datalad run -d ~/datalad/openneuro --explicit echo '{pwd}'            
/home/yoh/proj/datalad/datalad-master
```

datalad/local/foreach.py
yarikoptic: but it would leave user wondering on what has actually happened 
<details>
<summary>since no message would be provided</summary> 

```shell
$> datalad -f json_pp foreach -d ~/datalad/openneuro --fulfilled true --bottomup touch /dev/123 | head -n 14
{
  "action": "foreach",
  "command": [
    "touch",
    "/dev/123"
  ],
  "exception_traceback": "[foreach.py:run_cmd:244,cmd.py:run:408]",
  "path": "/home/yoh/datalad/openneuro/ds000001",
  "status": "error",
  "type": "dataset"
}
{
  "action": "foreach",
  "command": [
[ERROR] [Errno 32] Broken pipe [ansitowin32.py:write_plain_text:196] (BrokenPipeError) 
Exception ignored in: <_io.TextIOWrapper name='<stdout>' mode='w' encoding='utf-8'>
BrokenPipeError: [Errno 32] Broken pipe
(dev3) 1 67232 [1].....................................:Mon 22 Mar 2021 07:22:07 PM EDT:.
(git-annex)lena:~/datalad[master].datalad
$> datalad  foreach -d ~/datalad/openneuro --fulfilled true --bottomup touch /dev/123         
foreach(error): /home/yoh/datalad/openneuro/ds000001 (dataset)                                            
foreach(error): /home/yoh/datalad/openneuro/ds000002 (dataset)   
...
```
</details>

so I guess I will use `get_status_dict` but also pass `str(exc)` as message

datalad/local/foreach.py
```diff
@@ -0,0 +1,275 @@
...
+)
+
+lgr = logging.getLogger('datalad.local.foreach')
+
+
+@build_doc
+class ForEach(Interface):
+    r"""Run a command on the dataset and/or each of its sub-datasets.
+
+    Each dataset's repo runner is reused to execute the commands, so they
```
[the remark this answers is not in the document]
yarikoptic: I will just remove this doc piece for now ;-)

datalad/local/foreach.py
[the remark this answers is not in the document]
yarikoptic: Decided not to bother with any notes etc.

datalad/local/foreach.py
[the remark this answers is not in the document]
yarikoptic: added `--standard-outputs|--s-o`

datalad/local/foreach.py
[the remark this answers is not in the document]
yarikoptic: as of now -- no difference, and in both (external vs python) execution should happen while in that directory...
BUT it stroke me while writing that it would not work AFAIK with `-J` since we are parallelizing via threads, and `chpwd` would change PWD (AFAIK) of the entire process.  Added to TODOs and an example of "fail" under "can't chpwd for Python eval/exec when parallelizing using threads" in the description.
So would need to either revert to prior behavior (easy, and for python since `ds.` operations do not require chdir'ing I think it is generally not that bad), or figure out may be a way to run them in a subprocess, but then see yet another [known async gotcha](https://github.com/datalad/datalad/pull/5369/files#diff-4aa1a9688db72f7a475890e0ee6a343ea98af70ccc3bfa66f17da82c819dfa8dR188) so might be not that trivial ATM

datalad/local/foreach.py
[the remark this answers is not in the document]
bpoldrack: `--output-streams` might be a clearer distinction from the notion of `outputs` as used w/ `run`.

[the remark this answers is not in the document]
yarikoptic: sounds better indeed, I will rename to `--output-streams|--o-s` if no other alternatives suggested, thank you

datalad/local/foreach.py
kyleam: > with `run` which seems to have it to correspond to the current directory from which `run` is initiated

Unless a dataset instance is given, the directory in which `run` is called and the execution directory are one and the same.  The `pwd` passed as `WitlessRunner`s `cwd` and the placeholder are the same value.

datalad/local/foreach.py
kyleam: gh-5536 proposes to retain the message in the `json{,_pp}` output.

datalad/local/foreach.py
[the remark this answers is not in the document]
mih: I believe my point was the non-obvious fact that `--python` (or now `--cmd-type`) impact on what exactly REMAINDER could be -- calling for a more comprehensive docstring, I guess.

mih:
FTR: `create-sibling-github` could be RF'ed to use this, once it has landed.

datalad/local/foreach.py
yarikoptic: done

datalad/local/foreach.py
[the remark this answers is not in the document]
yarikoptic: I mentally boiled down to
- have `--chpwd=True|False` (so consistent between Python and CLI)
- crash with `NotImplementedError` if python and jobs discovered to be > 1: we have other non-thread safe calls (as e.g. `getcwd()` stumbled upon in #5606), and supporting parallel execution through threads could lead to surprises (and I just now realized that the whole threads-parallelism might be in jeopardy! need to check :-/).  so we need to approach it only with careful analysis etc.

datalad/local/foreach.py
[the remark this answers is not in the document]
yarikoptic: done

datalad/local/foreach.py
yarikoptic: https://github.com/datalad/datalad/pull/5640 is a pre-cursor to allow fulfilled to default to True here.

datalad/local/foreach.py
yarikoptic: wasdone

datalad/local/foreach.py
yarikoptic: > Unless a dataset instance is given, the directory in which `run` is called and the execution directory are one and the same

you might be in a subdirectory as well:
```shell
(git)lena:~datalad/datalad-master[nf-foreach]git
$> datalad -C datalad run --explicit 'echo {pwd} {dspath}' 
/home/yoh/proj/datalad/datalad-master/datalad /home/yoh/proj/datalad/datalad-master
```
and since `foreach` (if `chdir`ing - yet to add option) would run at the top directory of a (sub)dataset, `{ds.path}` would provide that path, but without `{pwd}` there would be no way to perform any operation which needs `cwd` without relying on some env var like `PWD` which might be set or not.
So I still do not see a point of 1. breaking analogie to `run`; 2. making `{pwd} == {ds.path}` and thus a duplicate; 3. removing possibility to have access to the initial directory from which `foreach` is executed.

datalad/local/foreach.py
yarikoptic: ok, to reduce user mental load I think I will add `{dspath}` as `= {ds.path}`  for `foreach`, even though as discussed/agreed exposing `{ds}` to `run` isn't .

datalad/local/foreach.py
yarikoptic: good.

datalad/interface/__init__.py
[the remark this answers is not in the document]
yarikoptic: FWIW: ATM not yet sure if renaming to anything else than `foreach` is worthwhile since `foreach` makes clearer analog to `git submodule foreach` and it is short... moreover, one of the main goals here is to help "crossing datasets boundaries", and it could indeed be then mixed in with also sweeping through various elements within those datasets, so in principle could do what it does + additional loop inside if (optional) `--entity-type` would be given (defaulting to dataset, which is just one, ... but could be "subdatasets", thus giving ability to act on a pair of `{ds}` and `{subds}` for each `{ds}` to be traversed)?

datalad/local/foreach.py
kyleam: > > Unless a dataset instance is given, the directory in which `run` is called and the execution directory are one and the same
> 
> you might be in a subdirectory as well

Yes, you can use `-C` to, in the option's own words, "run as if datalad was started in \<path> instead of the current working directory", and that of course affects run's pwd/the directory of execution.  Sorry, I'm not following what your point is by mentioning that, though.  

> 1. breaking analogie to run

Again, `run`s `pwd` is the execution directory for the command.  And, if a dataset instance is passed, that corresponds to `ds.path`.   If it's not, it's `getpwd()` (either `ds.path` or a subdirectory).  Either way, it's the actual `cwd` value passed to `WitlessRunner`.

I see a discrepancy with `run`, not an analogy. But...

> 2. making `{pwd} == {ds.path}` and thus a duplicate; 3. removing possibility to have access to the initial directory from which `foreach` is executed.

... fair enough. Perhaps it's worth coming up with another name, maybe `{pwd_initial}`?

datalad/local/foreach.py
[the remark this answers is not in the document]
yarikoptic: Have ATM
```
  --cmd-type {external|exec|eval}
                        type of the command. EXTERNAL: to be run in a child
                        process using dataset's runner; 'exec': Python source
                        code to execute using 'exec(), no value returned;
                        'eval': Python source code to evaluate using 'eval()',
                        return value is placed into 'result' field.
                        Constraints: value must be one of ('external', 'exec',
                        'eval') [Default: 'external']
```
May be later we would like to add some Python API specific mode to just pass a callable which will be given all the placeholders as kwargs.  But I do not think we even support such ability ATM for Python-specific constraint, do we?

datalad/local/foreach.py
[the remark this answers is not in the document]
yarikoptic: I have now 
```
            doc="""command for execution. [CMD: A leading '--' can be used to
            disambiguate this command from the preceding options to DataLad.
            For --cmd-type 'exec' or 'eval' only a single
            command argument (Python code) is supported. CMD]
            [PY: For `cmd_type='exec'` or `cmd_type='eval'` (Python code) should
            be either a string or a list with only a single item. PY]
            """),
```
could always be improved I guess

datalad/local/foreach.py
yarikoptic: After adding `--chpwd` I am now tending to agree with you @kyleam :  it becomes odd to have `--chpwd true` and then `{pwd}` still be the original one.  I also had difficult time to get the "effective" pwd while testing it out with or without --chpwd ;)
- I also dislike naming of `--chpwd` and tend to think about following @mih mantra of no boolean flags: we might want eventually to add `relds` or alike setting to invoke in the matching subdirectory of a dataset (e.g. when I am under some `.datalad/metadata/objects` or alike).  So may be it should become `--pwd ds,cwd` or alike?
- with the above -- may be `{cwd}` for the `{pwd_initial}` as for "current working directory"?  It might be a bit confusing since generally cwd and pwd are for the same thing, but IMHO "nothing the documentation can't address" ;)

WDYT on `--pwd` and `{cwd}`?

datalad/local/foreach.py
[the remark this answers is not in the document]
yarikoptic: decided against NotImplementedError to not cause pain and suffering whenever default `jobs` is set globally -- went for just issuing a warning and falling back to a threadless run.  Also the same in the case of `capture` mode for outputs -- we can't use `swallow_outputs` across threads really since they all use the same std streams.

yarikoptic:
gy -- codespell shows to be useful. Fixed locally, will push whenever.

yarikoptic:
> FTR: `create-sibling-github` could be RF'ed to use this, once it has landed.

For that we should then allow for direct python callables. I think those could be allowed under "eval" type (or could get their own "call" type but I do not see big need for that). Then, we could add `auto` mode (default) which would be "external" if a string (or list of strings) and "eval" if a callable (possible only in PY mode) to avoid necessity to always specify "eval" whenever we pass callable (kinda dubious since we can figure it out from the type).  WDYT?

datalad/local/foreach.py
kyleam: > WDYT on `--pwd` and `{cwd}`?

I'm a bit turned around, but that sounds okay to me.  Given your points 2 and 3 [above][1], I'm also okay leaving it at `pwd`, though I think using a different name is probably better.  Either way, I agree that this area is confusing, and we'll need some good documentation :)

[1]: https://github.com/datalad/datalad/pull/5517#discussion_r626916216

datalad/interface/__init__.py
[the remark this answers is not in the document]
yarikoptic: I will consider "parallelizing" across other entities out of scope for this PR -- not yet sure how it would be of help even, in particular at user API/interactions.

datalad/local/foreach.py
yarikoptic: pushed changes to have it 
```
  --chpwd {ds|pwd}      'ds' will change working directory to the top of the corresponding dataset. With 'pwd' no change of
                        working directory will happen. Note that for Python commands, due to use of threads, we do not
                        allow chdir=ds to be used with jobs > 1. Hint: use 'ds' and 'refds' objects' methods to execute
                        commands in the context of those datasets. Constraints: value must be one of ('ds', 'pwd')
                        [Default: 'ds']
```

datalad/interface/common_opts.py
```diff
@@ -42,6 +43,36 @@
...
+    constraints=EnsureStr() | EnsureNone())
+
+fulfilled = Parameter(
+    args=("--fulfilled",),
+    doc="""DEPRECATED: use [CMD: --state CMD][PY: `state` PY]
+    instead. If given, must be a boolean flag indicating whether
+    to consider either only locally present or absent datasets.
+    By default all subdatasets are considered regardless of their
+    status.""",
+    constraints=EnsureBool() | EnsureNone())
```
mih, line 65: Does it make sense the centralize deprecated argument specifications? Doesn't it increase the locations that need adjustment?

yarikoptic: As you know I am all for centralization of all deprecation annotations!  Not sure though if it would be folded into this PR which is already sizeable and elderly.

yarikoptic:
looking at this PR I wondered: may be `cmd` (command) should not be the positional argument but rather `--cmd` or alike kwarg (which would be mandatory to be specified).  This would allow later (or right away) to add `path` positional argument which is present in majority of other commands and at large it is specifying/limiting the "scope" of the operation.  Here it would be "not all subdatasets" but rather some (if specified).  WDYT @datalad/developers ?

yarikoptic:
> looking at this PR I wondered: may be `cmd` (command) should not be the positional argument but rather `--cmd` ...

implemented that but then decided to keep it KISS even though somewhat inconsistent with majority of other interfaces (since we typically provide `path` pos arg to limit/specify scope of command operation).  We have similarish `--contains` so may be eventually if need arises we would indeed just get `--path` or alike.

christian-monch:
> looking at this PR I wondered: may be `cmd` (command) should not be the positional argument but rather `--cmd` or alike kwarg (which would be mandatory to be specified). This would allow later (or right away) to add `path` positional argument which is present in majority of other commands and at large it is specifying/limiting the "scope" of the operation. Here it would be "not all subdatasets" but rather some (if specified). WDYT @datalad/developers ?

I would definitely second the path parameter.

christian-monch:
A small issue:

For example, on the longnow-podcast dataset the following command
```
datalad foreach -r -d <...>/longnow-podcasts/ "echo '{ds} {refds}'"
```

yields:

```
Dataset(<...>/longnow-podcasts) Dataset(<...>/longnow-podcasts)
```

I think that an external command should only receive the paths since it has no (at least no realistic) chance to make sense of the `Datalad(...)` predicate and has to remove it anyway.

yarikoptic:
> I think that an external command should only receive the paths since it has no (at least no realistic) chance to make sense of the `Datalad`-predicate and has to remove it anyway.

even though the `{ds}` itself is indeed not that useful in shell commands, its attributes could well be, e.g.

```shell
$> datalad foreach -d ~/datalad/ --o-s pass-through 'echo {ds.repo.dot_git}'          
subdataset(ok): abide (dataset)
subdataset(ok): allen-brain-observatory (dataset)
subdataset(ok): centerforopenneuroscience (dataset)
subdataset(ok): conp-dataset (dataset)
subdataset(ok): crcns (dataset)
subdataset(ok): dandi (dataset)
subdataset(ok): dbic (dataset)
subdataset(ok): hcp-openaccess (dataset)
subdataset(ok): labs (dataset)
subdataset(ok): openfmri (dataset)
subdataset(ok): openneuro (dataset)
/home/yoh/datalad/.git
foreach(ok): /home/yoh/datalad (dataset)
/home/yoh/datalad/abide/.git
foreach(ok): /home/yoh/datalad/abide (dataset)
/home/yoh/datalad/allen-brain-observatory/.git
```
and to keep it consistent across different types of commands/invocations.

christian-monch:
> even though the `{ds}` itself is indeed not that useful in shell commands, its attributes could well be, e.g.
> [...]
> and to keep it consistent across different types of commands/invocations.

Very good.

christian-monch:
I was wondering whether `foreach` might include files as well?

I am asking because I need content traversal for metadata extraction and metadata traversal for metadata filtering. That might be sufficiently separate concerns to keep them separate. But if we believe that they are closely related, we might try to envision/specify a common interface for that. Such a common interface might define a `foreach`-front and a traversal-implementer, e.g. subdataset, subdataset-and-content, metadata, etc. Of course, such an internal common interface would make the most sense, if the remaining arguments are orthogonal to the implementer, i.e. they should remain the same in syntax and semantics.

Just putting this out there.

[This message is partially copied from a DM exchange]

mih (commented):
I said it before, but I am going on record here too: This is a useful command, we should have it! Thanks for implementing it.

Resonating with https://github.com/datalad/datalad/pull/5517#issuecomment-982604861 my main thoughts are re future outlook and possible API extensions necessary to address extension of functionality.

With its name `foreach` the command takes a front-row seat -- and triggers my `create-sibling` warning.

I can easily see some of the API to generalize to use cases where the traversal entity is not a dataset, but I am not convinced that all of it will.

My suggestion would be to merge this command as `run-foreach-dataset`. This would tighten the connection to `run`, and take a more humble approach to API slot demands. Then we can gather real-world usage feedback and explore how this model can be translated to other use cases.

It certainly should no longer life in this PR. If that continues for whatever reason, it would be best to pull out the unrelated improvements and get the merged quickly.

WDYT?

bpoldrack:
> My suggestion would be to merge this command as `run-foreach-dataset`. This would tighten the connection to run, and take a more humble approach to API slot demands.

I like that suggestion.

bpoldrack (commented):
Exception handling needs adaption to https://github.com/datalad/datalad/blob/master/docs/source/design/exception_handling.rst

yarikoptic:
> My suggestion would be to merge this command as `run-foreach-dataset`. This would tighten the connection to `run`, and take a more humble approach to API slot demands.

I like this idea/line of thought!  Indeed we kept falling into the trap of "reserving" concise API interfaces only to later "redo" them and use another new name... My concern with proposed name would be making "connection" to `run` "too strong":
- this command was not intended to `save` and provide `run` commit records as `run` does.  But such suggestion actually brings the question either it may be should allow for such mode of operation (of `*foreach*`) or may be `run` itself should allow for that? (i.e. running the command in the (sub)datasets instead of a singular execution in the top). But then - why we need two commands...?
- (minor con) with placing command into `--cmd` (instead of a posarg like `run` has it) we would bring some "dichotomy" between `run` and this interfaces.

Let me dwell on this a bit.

datalad/local/foreach.py
```diff
@@ -0,0 +1,392 @@
...
+# ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+"""Plumbing command for running a command on each (sub)dataset"""
+
+__docformat__ = 'restructuredtext'
+
+
+import inspect
+import logging
+
+from argparse import REMAINDER
```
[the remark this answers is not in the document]
yarikoptic: relating to [this and referenced there comments](https://github.com/datalad/datalad/pull/5517#issuecomment-985574268) we could avoid it if we RF to have command  as `--cmd` with `nargs='+'`...  so, would be yet another `+1` on why to move into `--cmd`?

datalad/support/constraints.py
```diff
@@ -31,6 +31,19 @@ def _type_str(t):
...
+    """A helper construct to ease migrations for option=None.
+
+    Should be instantiated, so it gains `__repr__` and thus would render nicely
+    in docs etc.
+    """
+    def __repr__(self):
+        return 'None(DEPRECATED)'
+
+
+NoneDeprecated = _NoneDeprecated()
```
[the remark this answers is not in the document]
yarikoptic: doh, typed response and clicked Resolve instead of "Comment".

should I extract it into a separate PR? (would be ok with me)

mih:
> My concern with proposed name would be making "connection" to `run` "too strong": [...]

That makes sense. So `foreach-subdataset` seems to be a possible compromise of a name.

yarikoptic:
pushed rename, retitling

mih:
Fixing up the docs...

mih:
Huuiiiiiii.....

yarikoptic:
woohoo -- thanks @mih !  I am excited to see others trying it out and see how it would break in real life use cases

yarikoptic:
and thanks for the docs build fix up @mih -- my evening brain did not manage to catch that and I blamed some magic I didn't know about ;)

PR #5518, opened 2021-03-19
No tmpfile path resolving in make_temp()

mih:
This could be slow, seems unnecessary, and is undocumented behavior.

### DIFF
--- a/datalad/utils.py
+++ b/datalad/utils.py
@@ -1789,7 +1789,7 @@ def make_tempfile(content=None, wrapped=None, **tkwargs):
 
     filename = {False: tempfile.mktemp,
                 True: tempfile.mkdtemp}[mkdir](**tkwargs_)
-    filename = Path(filename).resolve()
+    filename = Path(filename)
 
     if content:
         (filename.write_bytes


mih:
Interesting failure pattern. Not sure I want to pursue this further, although I am a bit shocked that so much breaks.

kyleam:
> although I am a bit shocked that so much breaks

Oy... I was hoping the situation wouldn't be much worse than what it was in gh-2641 (two test failures).

yarikoptic:
FWIW, seems to come from

commit 8a8b0fc46208c288b990dc92abbc034475d46503
Author: Michael Hanke <michael.hanke@gmail.com>
Date:   Mon Jan 20 12:18:49 2020 +0100

    RF: Replace op.realpath with Path.resolve()
    
    Importantly, this makes the with_tempfile decorator use this
    new approach, which provides any test with proper longpaths
    that afford lexical comparisions on Windows.

so I guess Windows might like to have it?

mih:
No, if you check that change, you will see that it realpath'ed before.

kyleam:
I think a lot of the new failures are related to the unresolved directory not existing in the Docker SSH target.  Based on a comment in that repo, I think I didn't bother because it wasn't needed at the time:

https://github.com/datalad-tester/docker-ssh-target/blob/539f1922d5b321cf6227da28b92c7bf9a8648344/setup#L105-L108

So, something to be revisited.

mih:
This is too big for me to handle ATM -- nobody else jumped on it too -- let's close for now

PR #5522, opened 2021-03-22
3rd attempt to establish a core GitRepo class

mih:
This replaces #5335 with an even less ambitious goal of simply having a class and a set of tests to extend later. The reason for the continued downscaling are two previous attempt to wanted more, but achieved nothing.

In contrast to #5335 there is not functionality on state reporting or comparison in this PR anymore. It is merely the code needed to point to a local repo and run `call_git*()` methods.

As discussed before, this is no-longer targeting `.core` until more complete and more stable.

### DIFF
--- a/datalad/dataset/__init__.py
+++ b/datalad/dataset/__init__.py
@@ -1,0 +1,13 @@
+# emacs: -*- mode: python; py-indent-offset: 4; tab-width: 4; indent-tabs-mode: nil -*-
+# ex: set sts=4 ts=4 sw=4 noet:
+# ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+#
+#   See COPYING file distributed along with the datalad package for the
+#   copyright and license terms.
+#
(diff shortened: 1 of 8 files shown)


bpoldrack:
Fine with me. I think we should outline the ultimate goal and possible intermediate states in a dedicated project or sth. I'm increasingly convinced the aim should be to get to a single Repo class in `.core` eventually. But this requires a lot of planning ahead, I guess.

mih:
> Fine with me. I think we should outline the ultimate goal and possible intermediate states in a dedicated project or sth. I'm increasingly convinced the aim should be to get to a single Repo class in `.core` eventually. But this requires a lot of planning ahead, I guess.

At this point I have nothing to add beyond the original plan #3988 and the strategic direction voted upon in #4589.

kyleam (approved):
I have no objections to the more manageable scope, and, glancing through the `--color-moved` output, all the changes look sensible to me.

mih:
Thx @kyleam !

PR #5523, opened 2021-03-22
Minor beautifications salvaged from #5390

mih:
I consider those useful, even when no longer going forward with #5390

### DIFF
--- a/datalad/core/local/create.py
+++ b/datalad/core/local/create.py
@@ -498,7 +498,7 @@ def _setup_annex_repo(path, add_to_git, initopts=None, fake_dates=False,
                       description=None):
     """Create and configure a repository at `path`
 
-    This includes a default setup which files to never annex.
+    This includes a default setup of annex.largefiles.
 
     Parameters
     ----------
--- a/datalad/core/local/create.py
+++ b/datalad/core/local/create.py
@@ -377,70 +377,11 @@ class Create(Interface):
 
         # create and configure desired repository
         if no_annex:
-            lgr.info("Creating a new git repo at %s", tbds.path)
-            tbrepo = GitRepo(
-                tbds.path,
-                create=True,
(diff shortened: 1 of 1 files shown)


datalad/core/local/create.py
```diff
@@ -516,3 +457,115 @@ def __call__(
...
+        'type': 'file',
+        'state': 'untracked'}
+    return tbrepo
+
+
+def _setup_annex_repo(path, add_to_git, initopts=None, fake_dates=False,
+                      description=None):
+    """Create and configure a repository at `path`
+
+    This includes a default setup which files to never annex.
```
kyleam, line 501: s/which/of &/ ?

Or perhaps "This includes a default setup of `annex.largefiles`."

datalad/core/local/create.py
```diff
@@ -516,3 +457,115 @@ def __call__(
...
+
+
+def _setup_git_repo(path, add_to_git, initopts=None, fake_dates=False):
+    """Create and configure a repository at `path`
+
+    Parameters
+    ----------
+    path: str or Path
+      Path of the repository
+    add_to_git: dict
```
kyleam, line 469: `_setup_git_repo` and `_setup_annex_repo` don't care about what information is already in `add_to_git`, so I'd find it clearer if the side channel was avoided by each function dropping the parameter and returning a freshly created dictionary alongside the repo instance.

kyleam (approved):
Thanks, I think this makes `create` much more digestible.

Left a comment about `add_to_git`, but I have no objections if things are left as is.

datalad/core/local/create.py
mih: Yes, true, done (the full replacement).

datalad/core/local/create.py
mih: Done!

Issue #3067, opened 2018-12-05
run: cmdline args processing forbids passing --version to the cmd


yarikoptic:
```shell
$> datalad run heudiconv --version 
datalad 0.11.1.dev55

$> datalad run --cmd heudiconv --version 
datalad 0.11.1.dev55

$> datalad run -- heudiconv --version   
[INFO   ] == Command start (output follows) ===== 
/bin/sh: 0: Illegal option --
[INFO   ] == Command exit (modification check follows) ===== 
[INFO   ] The command had a non-zero exit code. If this is expected, you can save the changes with 'datalad save -r -F .git/COMMIT_EDITMSG .' 
CommandError: command '-- heudiconv --version' failed with exitcode 2
Failed to run u'-- heudiconv --version' under '/tmp/testds'. Exit code=2.
```
I guess it is the side-effect of some of my "optimizations" to make cmdline invocations of basic commands tolerably not-slow.  Seems to be --version specific since doesn't happen for `--help`.

Also -- may be it is worthwhile to add handling of `--` as an indicator for "do not consider any subsequent argument as a possible option and treat as a positional argument" which is I believe a common convention (don't remember where it stems from)

kyleam:
> Seems to be --version specific since doesn't happen for `--help`.

Odd.

> Also -- may be it is worthwhile to add handling of `--` as an indicator for "do not consider any subsequent argument as a possible option and treat as a positional argument" which is I believe a common convention (don't remember where it stems from)

I think it'd be worthwhile because it would address datalad/datalad-container#46.  I don't particularly like promoting it as a solution to this issue, which I think this really should be viewed as an issue with datalad's parser and fixed there.

To add a working variant to your list:

    $ datalad run 'heudiconv --version'

yarikoptic:
FWIW - argparse seems to do handling of `--` internally (at least in `_parse_known_args`):
```
$> datalad run blah --version   
datalad 0.11.1.dev60

$> datalad run git --version    
datalad 0.11.1.dev60

$> datalad run git -- --version 
[INFO   ] == Command start (output follows) ===== 
unknown option: --
usage: git [--version] [--help] [-C <path>] [-c <name>=<value>]
           [--exec-path[=<path>]] [--html-path] [--man-path] [--info-path]
           [-p | --paginate | -P | --no-pager] [--no-replace-objects] [--bare]
           [--git-dir=<path>] [--work-tree=<path>] [--namespace=<name>]
           <command> [<args>]
[...]

PR #5543, opened 2021-04-03
ENH: Make --version output just version, and CMD --version output module and its version

### LINKED ISSUE #5527: `datalad extension-command --version` to provide version of the extension, not of datalad
inspired by #5526  on multiple occasions I needed that. `datalad wtf` is a bit too much overhead to get that very "bite" of information.

yarikoptic:
I might have went a little too wild, but I felt that having clean output
of datalad --version is nice -- it could facilitate use by scripts since
would no longer need parsing etc (but also might break them now).

So here is how it would look now with these changes:

	$> datalad --version
	0.14.1.dev129

	$> datalad install --version
	datalad 0.14.1.dev129

	$> datalad containers-list --version
	datalad_container 1.1.2

	$> git-annex-remote-datalad --version
	datalad 0.14.1.dev129

Closes #5527

edit: a side effect, it also addresses #3067 which was worked around via adding `--` to separate out command for `datalad run`. Now would even work in the `datalad run whatever --version` without `--`

### DIFF
--- a/datalad/cmdline/helpers.py
+++ b/datalad/cmdline/helpers.py
@@ -16,6 +16,7 @@ import os
 import re
 import sys
 import gzip
+import warnings
 from tempfile import NamedTemporaryFile
 from textwrap import wrap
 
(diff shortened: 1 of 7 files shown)


datalad/cmdline/helpers.py
```diff
@@ -132,6 +133,55 @@ def __call__(self, parser, namespace, values, option_string=None):
...
 
 
+def parser_add_version_opt(parser, mod_name, include_name=False, delay=False):
+    """Setup --version option
+
+    Parameters
+    ----------
+    parser:
+    mod_name: str, optional
+    include_name: bool, optional
```
kyleam, line 143: I don't feel strongly about it, but my vote would be to just stick with the previous output and always include the name.

kyleam (approved):
Looks good to me, and it's nice that it fixed the deeper issue in gh-3067.

datalad/cmdline/helpers.py
yarikoptic: I am also luke warm about this change -- benefit is minimal and it might break someone's (unlikely but possible to exist) script.  May be @mih has an opinion on this minute aspect?

yarikoptic:
ok, added a commit to revert back to the prior behavior for `datalad --version` (left possibility/option in the function to exclude just in case).  So should be ready to go if no tests broken

PR #5544, opened 2021-04-03
ENH: datalad shell-completion  helper

yarikoptic:
This PR continues the streak of "silly" UX ones (following #5543)

Well, we have `tools/cmdline-completion` helper which nobody sees and
which we do not distribute via pypi, conda-forge, etc.
To make it easy to enable completion on any installation (yet to try
on Windows), this command could provide easy to access way.

As this is a new minor command, duplicating what we shipped under tools/ already, positioned it against `maint`.


bpoldrack:
Fine with me as a way to make it more discoverable.

datalad/interface/base.py
```diff
@@ -651,7 +651,7 @@ class Interface(object):
 
     _OLDSTYLE_COMMANDS = (
         'AddArchiveContent', 'CrawlInit', 'Crawl', 'CreateSiblingGithub',
-        'CreateTestDataset', 'Export', 'Ls', 'SSHRun', 'Test')
+        'CreateTestDataset', 'Export', 'Ls', 'SSHRun', 'ShellCompletion', 'Test')
```
mih, line 654: This tuple should shrink and never grow. It is fine for me to have outlier commands like this, but they should be treated as their own special case further down, and not populate a TODO list.

Unless, I got it wrong, and this new command is actually meant to be RF'ed into a normal datalad command.

datalad/interface/shell_completion.py
```diff
@@ -0,0 +1,64 @@
...
+    _no_eval_results = True
+    _params_ = {}
+
+    @staticmethod
+    def __call__():
+        """
+        """
+        print("""\
+# Universal completion script for DataLad with the core autogenerated by
+# python-argcomplete and only slightly umproved to work for ZSH if sourced under ZSH.
```
mih, line 42:
```suggestion
# python-argcomplete and only slightly improved to work for ZSH if sourced under ZSH.
```

mih (changes requested):
Good idea!

datalad/interface/base.py
yarikoptic: To say the truth I forgot the differences between plugins, "oldstyle" commands" and "normal datalad commands".  What I know is that this command output to stdout should be the only thing produced to the stdout.  I do not see a point of knowing the `json` representation etc.  But I guess it could still be just for tailored renderer to "present" it, but that just makes it like a "hello world" in java ;)

mih: That is my point. Why make this new command an old-style command, when it is not even completely understood what that means. As stated above, I am not concerned about the behavior of the command, only about how it is injected into the API.

yarikoptic: ok, now it is fancy new, but I followed up with #5575 for consideration

yarikoptic:
d;oh -- need to add some minimal test...

kyleam (approved):
It seems all comments are addressed, and it looks good to me.

PR #5554, opened 2021-04-09
Dissolve plugins

### LINKED ISSUE #4519: Dissolve plugins (and supporting functionality)
We have interfaces, plugins, procedures, and extensions now.
IIRC plugin mechanism was introduced before procedures and pretty much morphed into an interfaces:
```
$> grep Interface datalad/plugin/*py   
datalad/plugin/add_readme.py:from datalad.interface.base import Interface
datalad/plugin/add_readme.py:class AddReadme(Interface):
datalad/plugin/addurls.py:from datalad.interface.base import Interface
datalad/plugin/addurls.py:class Addurls(Interface):
datalad/plugin/check_dates.py:from datalad.interface.base import Interface
datalad/plugin/check_dates.py:class CheckDates(Interface):
datalad/plugin/export_archive.py:from datalad.interface.base import Interface
datalad/plugin/export_archive.py:class ExportArchive(Interface):
datalad/plugin/export_to_figshare.py:from datalad.interface.base import Interface
datalad/plugin/export_to_figshare.py:class ExportToFigshare(Interface):
datalad/plugin/__init__.py:    # TODO check all symbols whether they are derived from Interface
datalad/plugin/no_annex.py:from datalad.interface.base import Interface
datalad/plugin/no_annex.py:class NoAnnex(Interface):
datalad/plugin/wtf.py:from datalad.interface.base import Interface
datalad/plugin/wtf.py:class WTF(Interface):
```

should we just remove the notion of plugins (e.g. in 0.14) and move existing ones around into corresponding spots?  

or am I missing some unique feature of plugins ?

mih:
- [x] Converted all plugins into registered commands
- [x] Moved code and tests out of `plugin` and into `local` or `distributed`
- [x] Split up tests to have one file for each command
- [x] Adjust documentation
- [x] Adjusted logger names
- [x] Leave deprecation warnings and compatibility kludges
- [x] Try removing the plugin loading code

Fixes #4519


mih:
For all ex-plugins `datalad.tests.test_api.test_consistent_order_of_args` fails. They don't have the "right" order -- and sure thing I still cannot infer what that order would be. The test already has traces of my struggle from earlier:

```py
    elif intf.__name__ == 'ExtractMetadata':
        # MIH I was never sure what this test enforces and it takes
        # me ages each time to try to wrap my head around it
        # I am confident that I do not want to change the API of this
        # command now that it is a command and no longer a plugin
        # hence this exception
        eq_(spec_posargs, spec_posargs)
 ```

What ever these commands where doing, it was wrong in some way. However, I think I will just extend the list of exceptions -- I doubt someone prefers to break API for sudden compliance with a rule whose violation has never been a problem.

mih:
I have to detach from this project now. Feel free to have at it

bpoldrack:
> What ever these commands where doing, it was wrong in some way. However, I think I will just extend the list of exceptions -- I doubt someone prefers to break API for sudden compliance with a rule whose violation has never been a problem.

I think it's about the order in the `_params_` dict only (positionals first). So - no API breakage needed to comply. However, I don't think it has any actual effect (anymore?), not even on building the docs.

mih:
> I think it's about the order in the `_params_` dict only (positionals first).

Hu? `dict` has no order.

bpoldrack:
> Hu? dict has no order.

It does. Didn't we have that topic recently? Since 3.6 or something.

But I misread the code nevertheless, so ... what the test does seem to try to ensure is, that positional arguments in python interface are first in signature and match what's declared positional for CLI in that dict.

mih:
> > Hu? dict has no order.
> 
> It does. Didn't we have that topic recently? Since 3.6 or something.
> 
> But I misread the code nevertheless, so ... what the test does seem to try to ensure is, that positional arguments in python interface are first in signature and match what's declared positional for CLI in that dict.

Ha, true! So, any necessary change (which I am still not sure what it needs to be) would change the API. If that is true, I maintain my attitude and approach.

bpoldrack:
> Ha, true! So, any necessary change (which I am still not sure what it needs to be) would change the API. If that is true, I maintain my attitude and approach.

Yeah - I'm trying to decide whether to remove the test or make it more clear, since I'm struggled with that several times, too.

Example: AddReadme: `dataset` is first in python signature and positional - this seems wrong indeed. Moving it to first position for binding is done by the `datasetmethod` decorator - no need for that in `__call__`'s signature. But that way `dataset` is mandatory in unbound call, while `filename` becomes a kw arg in python but is positional in CLI. Compliance and consitency could be achieved by mainting CLI and changing python API (slightly). Not sure, what to make of that really, but seems to me it actually is an annoyance, that simply didn't spring to the eye due to a lack of examples pointing out, how the different possible calls in python are confusing and do not as directly translate in a CL call like other commands do.

kyleam:
CrippledFS failure looks like gh-5300.

mih:
@bpoldrack If you have a vision re API changes, go for it. I consider this a change separate from this PR, which only aims to remove the infrastructure for plugins. The commands are only touched insofar as they still relied on that infrastructure.

bpoldrack:
> I consider this a change separate from this PR

Agreed. Moved into two issues.

kyleam (approved):
Looks good to me.  Thanks for doing the tedious conversion.

mih:
Thx @kyleam for the review and the fix!

PR #5555, opened 2021-04-09
RF: Move web UI into an extension

### LINKED ISSUE #5376: How to deal with the javascript blobs in the code base
Github tells that 20% of the codebase are javascript https://github.com/datalad/datalad/search?l=javascript These are largely static, sometimes extremely outdated versions of javascript libraries:

- https://github.com/datalad/datalad/blob/7c1219d17036d5b22cc69d0185ab0a03609467db/datalad/resources/website/tests/qunit/sinon-1.17.5.js (version not even listed on upstream website anymore, current is 9.2.4!)
- https://github.com/datalad/datalad/blob/7c1219d17036d5b22cc69d0185ab0a03609467db/datalad/resources/website/assets/js/jquery-2.2.4.js (2.x branch is no longer supported upstream, 3.5.1 is current)
- https://github.com/datalad/datalad/blob/7c1219d17036d5b22cc69d0185ab0a03609467db/datalad/resources/website/assets/js/md5-2.3.0.js (uptodate)
- https://github.com/datalad/datalad/blob/7c1219d17036d5b22cc69d0185ab0a03609467db/datalad/resources/website/tests/qunit/qunit-2.0.1.js (~4 years old, 2.14.0 is current)
-  https://github.com/datalad/datalad/blob/7c1219d17036d5b22cc69d0185ab0a03609467db/datalad/resources/website/assets/js/jquery.dataTables-1.10.12.js (1.10.23 is current, unclear if it could be even made to work with jquery 2x)
- https://github.com/datalad/datalad/blob/7c1219d17036d5b22cc69d0185ab0a03609467db/datalad/resources/website/assets/js/showdown-1.8.5-33-g0c6ce94.js (~3 years old, 1.9.1 is current)
- https://github.com/datalad/datalad/blob/7c1219d17036d5b22cc69d0185ab0a03609467db/datalad/resources/website/tests/qunit/sinon-qunit-1.0.0.js (~10 years old, 2.0 is newer and already 6 years old)

So with few exceptions these are very outdated. Moreover, they just sit in the source repo and have received no maintenance after 2018 (some no a single change after the initial inclusion).

We should discuss how to handle this code in the future.

An initial step could be to pull the (large) static js library blobs from the repo, and replace them with adhoc downloads from an alternative source.

mih:
The change from the user perspective is that they will have to
install https://github.com/datalad/datalad-deprecated to keep
using the web UI. They will be prompted, if that is not the case.

The tests are kept as-is, but testing web UI related functionality
has been made conditional on the presence of the extension package.

Fixes #5376 

A sibling PR with the web UI will come up soon.


datalad/distribution/tests/test_create_sibling.py
```diff
@@ -78,44 +78,12 @@
...
-    with open(hook_path) as f:
-        assert_not_in(target_path, f.read())
-    # correct ls_json command in hook content (path wrapped in "quotes)
-    ok_file_has_content(hook_path,
-                        '.*datalad ls -a --json file \..*',
-                        re_=True,
-                        flags=re.DOTALL)
-
+try:
+    from datalad_deprecated.tests.test_sibling_webui \
```
yarikoptic, line 82: could there be at least a single matrix run (which would cover these tests) with `datalad_deprecated` installed?  or may be the entire test just copied over to deprecated and test here would not test for this aspect at all.  And then we add that dedicated github action for `datalad_deprecated` extension

mih: ATM they are very intertwined. I'd say it would make sense in the extension to simply import the test suite and run it in full, in a dedicated extension test action, as you suggested.

mih: Done

mih:
This is ready.

mih:
Verdict from call: "I will survive!"

PR #5565, opened 2021-04-13
RF: test that ex-plugins have "dataset" as the first posarg

yarikoptic:
echoing the meetup discussion

I took liberty of removing MIH comment, but we could bring it back if so desired


kyleam (approved):

yarikoptic:
hm, appveyor isn't happy with `brew` install of git-annex on OSX due to libgpg download fail.  I would not bother digging out if needs to be reported at this point

Issue #5562, opened 2021-04-12
(re)move `jsmin` dependency


mih:
https://github.com/datalad/datalad/pull/5555 move the web UI, and it seems `jsmin` is just for that. `setup.py` says "actually also involved in `install`" but I cannot see that.

I'd say: remove in -core and add as a dependency of -deprecated with similar conditional setup as the web UI upload.

PR #5568, opened 2021-04-14
Remove javascript-related dependencies and tests

mih:
This makes -core a little lighter. All removed bits have an equivalent in -deprecated now, incl. the tests.

Closes: #5562


mih:
All test failures are known flukes

mih:
I think this is good to go.

PR #5569, opened 2021-04-14
RF: Move ls() and ls_webui to -deprecated

### LINKED ISSUE #5359: Discontinue or RF `ls` command
This code has not seen any work in 3-4 years. It is either perfect or not used. Several open issues are associated with it (most for several years):

- https://github.com/datalad/datalad/issues/2875
- https://github.com/datalad/datalad/issues/2780
- https://github.com/datalad/datalad/issues/2217
- https://github.com/datalad/datalad/issues/2061
- https://github.com/datalad/datalad/issues/1145
- https://github.com/datalad/datalad/issues/858

The aspect of `ls` functionality that is focused on local datasets could be cheaply reimplemented (and with it almost automatically addressing some of the issues above) using the tooling that fuels `status()`. However, `ls` also aims to be a frontend for S3 (and in the future other platforms -- although we now live in the future and it did not happen ;-), and for that uses an abstraction that is unique to `ls` and not used elsewhere. This special nature makes it unattractive to work on `ls`.

Personally, I think we should:

- discontinue the S3-related functionality in datalad-core -- and maybe move it to the crawler (where it makes kinda sense as an intro/inspection tool
- either discontinue `ls` entirely (what feature does it provide without S3?) or reimplement it using more standard building blocks

### LINKED ISSUE #5563: Move `ls` and `ls_webui` to deprecated
Intermediate approach to #5359. This proposal has received a tentative "kinda make sense in the light of and after #5555".

Some related issue(s):

- [x] `save_print()` is only used in this code and can be moved with it

mih:
They landed there with https://github.com/datalad/datalad-deprecated/pull/2

Closes #5563
Closes #5359


mih:
So this is a more complex transition. The crawler also needs this.

mih:
I have made a release of -deprecated to pypi, such that the crawler can simply depend on it, to maintain availability of `ls()`, whether or not the installed datalad provides it.

mih:
I have put up https://github.com/datalad/datalad-crawler/pull/94 for the crawler

mih:
Minus the macos git-annex installation issue all is good now. The -crawler release 0.8.4 ensures availability of `ls` regardless of the state of the datalad core installation state. @yarikoptic gave feedback on adding -deprecated as a benchmark CI run dependency in order to maintain the normal set of benchmarks.

Bottom line: done!

PR #5570, opened 2021-04-14
TMP: Test remove unused utilities

mih:
Some may move, some may die -- let's see

Moving TODO:

- [ ] `lmtime` -> crawler
- [ ] `get_logfilename` -> crawler

### DIFF
--- a/datalad/utils.py
+++ b/datalad/utils.py
@@ -299,14 +299,6 @@ def md5sum(filename):
     return Digester(digests=['md5'])(filename)['md5']
 
 
-# unused in -core
-def sorted_files(dout):
-    """Return a (sorted) list of files under dout
-    """
-    return sorted(sum([[opj(r, f)[len(dout) + 1:] for f in files]
-                       for r, d, files in os.walk(dout)
-                       if not '.git' in r], []))
-
 _encoded_dirsep = r'\\'  if on_windows else r'/'
 _VCS_REGEX = r'%s\.(?:git|gitattributes|svn|bzr|hg)(?:%s|$)' % (
     _encoded_dirsep, _encoded_dirsep)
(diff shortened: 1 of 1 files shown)


datalad/utils.py
```diff
@@ -1097,74 +993,6 @@ def  _wrap_line_profile(*args, **kwargs):
     return  _wrap_line_profile
 
 
-# unused in -core
-@optional_args
-def collect_method_callstats(func):
```
bpoldrack, line 1102: I wouldn't delete this one because it's unused. To me it's meant to be used adhoc when assessing things locally, not something to decorate anything permanently with. There may be a better place to move it to, but deletion doesn't seem to be it.

mih: `datalad.tests.utils.py` sounds like a better place then.

How do we know that it works? ;-)

yarikoptic: ha -- and isn't it something you @mih wanted (to figure out what actually uses some function)? ;)
I would keep it around.  It did come handy in the original use case (183b0202723aed3b1e2fa32ab5c55833312434ec) and to say the truth I did forget about it. Indeed I do not expect it to be used in the main code UNLESS we actually decide to establish some optional (triggered e.g. by env var) collection of usage stats, and per se it is not really about helping tests machinery.  As for "how do we know" - I guess someone is yet to compose a test for it , and may be indeed meanwhile move such unused but potentially useful into something like `datalad._utils` which would disappear from the public API surface?

datalad/utils.py
```diff
@@ -1193,13 +1021,6 @@ def wrapped_func(*args, **kwargs):
 #
 
 
-# unused in -core
-@contextmanager
-def nothing_cm():
```
yarikoptic, line 1198: oh -- here you are my little helper I am going to use in the #749 to avoid code duplication for either doing chdir or not!
I guess I could reintroduce it there if keeping it here is a big ask

datalad/utils.py
```diff
@@ -2105,31 +1913,6 @@ def safe_print(s):
 # IO Helpers
 #
 
-# unused in -core
-def open_r_encdetect(fname, readahead=1000):
```
yarikoptic, line 2109: FWIW should be good to go. The original/only use was RFed in 5e07eadd76f7dabef81eaa8ec0310d42867af89c

datalad/utils.py
bpoldrack: > may be indeed meanwhile move such unused but potentially useful into something like datalad._utils which would disappear from the public API surface?

Agree with the reasoning and `datalad._utils` sounds better to me than `datalad.tests.utils`. And alternative may be to have a module for such debugging/analyzing under `datalad/tools`?

yarikoptic: ha - I was always wondering on what difference is between `utils` and `tools`.  I guess it could be it -- `utils` are for use in the codebase, and `tools` -- for internal use/debugging/etc.  Albeit easy to mix up, I think a clear docstring describing that  and stating that those interfaces might change without due DataLad version change should be be enough.

mih:
Closing due to low-priority and lack of time.

Issue #4593, opened 2020-05-30
Strip/RF "OLDSTYLE" interface commands 


mih:
`Interface` still contains special casing for commands that never made the transition to the result handling that had been standard for quite some time now: https://github.com/datalad/datalad/blob/master/datalad/interface/base.py#L643

It needs to be investigate how much can simply be stripped and how much needs actual RF.

mih:
Related #5548

yarikoptic:
FWIW, we no longer have an explicit list of such commands
```
datalad/interface/tests/test_base.py:    # with dissolution of _OLD_STYLE_COMMANDS and just relying on having
datalad/interface/tests/test_base.py-    # @eval_results, no result_filter is added, since those commands are
datalad/interface/tests/test_base.py-    # not guaranteed to return/yield any record suitable for filtering.
datalad/interface/tests/test_base.py-    # The effect is the same -- those "common" options are not really applicable
datalad/interface/tests/test_base.py-    # to Interface's which do not return/yield expected records
datalad/interface/tests/test_base.py-    assert_not_in("result_filter",
```

in 0034c81e03e8c861ea49d68368e2677d1c449767 comment I gave an alternative command to find such commands
```
$> git grep -l 'class.*(Interface)' | xargs grep -L @eval_results | grep -v -e '/test_'
datalad/distribution/create_test_dataset.py
datalad/interface/add_archive_content.py
datalad/interface/test.py
datalad/support/sshrun.py
```
so we still have a few. And I believe @adswa is working on add_archive_content in https://github.com/datalad/datalad/pull/6105

mih:
Retitled. Tagging @adswa's #6105

mih:
Thanks to @adswa we have one less command on this list.

PR #5575, opened 2021-04-15
RF: centralize logic on either @eval_results used, remove need for _OLDSTYLE_COMMANDS

yarikoptic:
ATM it is needed to define _no_eval_results attribute to signal docbuilder to
not populate docs for all the additional kwargs provided by @eval_results.
Some commands even acquired @eval_results (e.g. create_sibling) but still had
that attribute still set to True for unknown reason (probably just an effect of
the decentralized logic during RFing).  The effect of this was not
mentioned probably because outside users/developers  do not care to look up for
those possible additional kwargs, and "core" developers simply know about them.

I believe a similar situation was with common options defined at top level for
datalad (--on-error, etc) -- since old interfaces did not care to return/yield
structured records, they could (and still can) be provided but have no real
effect on anything.  So even though I had to tune up some tests in
interfaces/tests/test_base.py I expect them to be of no negative effect.

This change ENHs @eval_results to add _eval_results attriute to the wrapper, so
outside logic could consistently decide to either add docstrings AND to handle
all those additional kwargs.  Even though we would still have some common
options which are not applicable to interfaces which have no @eval_results, I
expect situation to not differ pragmatically in that regard.

As a result it should no longer require listing of OLDSTYLE commands.

I can see how some developers might take it as undesired effect since now there
would be no hardcoded listing of commands to RF or get rid of.  Since we
define a single Interface per file, analogous check could be done e.g.  via

	$> git grep -l 'class.*(Interface)' | xargs grep -L @eval_results | grep -v -e '/test_'
	datalad/distribution/create_sibling_github.py
	datalad/distribution/create_test_dataset.py
	datalad/interface/add_archive_content.py
	datalad/interface/ls.py
	datalad/interface/test.py
	datalad/support/sshrun.py

in the core codebase (so we would indeed loose some hits from extensions).

But to me, and with shell-completion
(https://github.com/datalad/datalad/pull/5544) and sshrun helpers as use cases,
it seems more of a matter of the fact that not all Interfaces should provide
fancy results handling: they might be leaf interfaces not intended for nesting
in a larger logic, not operate on datasets or be generators, have no associated
with their action paths, etc.  I really see no reason to complicate their code
with more code to accomplish the same mission and possibly only to complicate
debugging.  I think we should allow for those to co-exist in peace with
fancy @eval_result'ed interfaces.

If we decide to allow for "simple" interfaces (for such helpers as above), we
might want to RF further to derive an InterfaceWithResults to avoid ad-hoc
_has_eval_results_call, which this PR introduces, and/or to avoid explicit
addition of the @eval_results for every __call__ of every InterfaceWithResults.
The above witch hunt also could be carried through a simple grep
'class.*(Interface)'.

TODOs
- [x] see if any functionality broken (didn't do a full sweep of tests) - seems all as kosher as could be
- [x] discussion on desire/feasibility
- [ ] decide on either to introduce `InterfaceWithResults` here


mih:
I haven't read or thought in depth yet, but this is related to https://github.com/datalad/datalad/issues/4593

yarikoptic:
yes, it is related, but not contingent on the decision there and only provides some argumentation in description here on my opinion that not all commands have to be `InterfaceWithResults` -- IMHO we should still allow for "pure" `Interface`s which do not use `@eval_results`.   But overall -- this could be merged to address incorrect docstrings for some existing interfaces and centralize the logic.

bpoldrack:
I think I'm fine with this change. If things can be simplified they generally should. In other words: This

> I really see no reason to complicate their code
with more code to accomplish the same mission and possibly only to complicate
debugging.

is a perfectly reasonable argument. Additionally, whether or not to include those parameters in the docs should not be decided within command implementation but outside for consistency.


On the other hand: I don't think I really want commands not yielding result records in the first place. (I know - we have them) We should aim for unification. Currently, I consider the use cases for not having any result invalid. If we find, that `path` can't be mandatory, then `path` shouldn't be mandatory. If we don't want terminal output to be "polluted", we can have a "NullRenderer". But a python caller should be able to expect one way of assessing what happened across commands. And with that aim in the back of my head, this does feel a bit like hiding the inconsistency rather than cleaning up something.

However, this concern is technically independent from this change and a simplification of the interface (what does someone writing a command has to declare?) might be more important/urgent to not propagate the idea into ever more extensions.
Hence, overall +1 from me.

yarikoptic:
Thank you @bpoldrack !  Input from other @datalad/developers would also be appreciated so I do not proceed if there is strong dislike.
And WDYT about introducing `InterfaceWithResults`?  or should it suffice for now to just have that ducktyping helper?

mih:
I think this change makes sense. It forcibly connects pieces that must be connected.

yarikoptic:
ok, let's not bother with `InterfaceWithResults` for now and let's proceed

Issue #5521, opened 2021-03-20
save -d . subds/subsubds takes forever for no good reason


yarikoptic:
might be the same underlying issue as of #5283 (didn't analyze details yet)

the intention from running this command, without `-r`, is just to commit as possibly updated state of the sub/subds within hierarchy.  In majority of the cases it might not even be changed in my use case.  But it takes really forever IMHO for no valid reason

```shell
$> datalad --version
datalad 0.14.0.dev46

$> time datalad save -d . -J1  -m 'Fast forward psychoinformatics-de/hirni-toolbox recursively' psychoinformatics-de/
add(ok): psychoinformatics-de (file)
Total (1 skipped):  50%|████████████████████▌                    | 1.00/2.00 [00:00<00:00, 806 datasets/s]
.... was stuck at this already for the duration of me opening new tab, and composing this issue ... will later report output of time
```
NB rate of progress (806 datasets/s is really "funny" given it been stuck for so long having accomplished virtually nothing ;) )

oh -- finally it is "done" with a considerable effort on user and system:

```shell
$> time datalad save -d . -J1  -m 'Fast forward psychoinformatics-de/hirni-toolbox recursively' psychoinformatics-de/
add(ok): psychoinformatics-de (file)
action summary:                                                                                           
  add (ok: 1)
  save (notneeded: 2)
datalad save -d . -J1 -m  psychoinformatics-de/  49.94s user 102.12s system 52% cpu 4:51.77 total
```

yarikoptic:
<details>
<summary>some observations/notes</summary> 

the culprit is that at the end it does run
```
[DEBUG  ] Async run:
|  cwd=/mnt/btrfs/datasets/datalad/crawl
|  cmd=['git', '-c', 'diff.ignoreSubmodules=none', 'commit', '-m', 'Fast forward psychoinformatics-de/hirni-toolbox recursively', '--']
[DEBUG  ] Launching process ['git', '-c', 'diff.ignoreSubmodules=none', 'commit', '-m', 'Fast forward psychoinformatics-de/hirni-toolbox recursively', '--']
[DEBUG  ] Process 8095 started
```
i.e. commit without specifying any path, so as if nothing to be committed (there is "no staging area" for datalad, right?) which takes that while (>2 minutes cold, >1 warm)...

trying to debug why we are saving and got to
```
> /home/yoh/proj/datalad/datalad-master/datalad/support/gitrepo.py(3593)save_()
-> amend = kwargs.get('amend', False)
[...]

PR #5576, opened 2021-04-16
WiP: trying to get a better status than just "untracked" for a subdataset

yarikoptic:
In a very dirty state -- just want to do a quick sweep through the tests.
It does seem to address #5521 (runs for 0.9 sec instead of >4 minutes; edit: took 20 minutes on a cold `git`) with the same effect of accomplishing nothing.
Actually it even removes result record on `add(ok): psychoinformatics-de (file)` for intermediate dataset which is not really added anyhow (no changes) and reports only a correct `save (notneeded: 2)`

Edit: most likely it is a wrong place for the fix add it should be done in status, to return records for intermediate datasets if refds was provided

### DIFF
--- a/datalad/core/local/save.py
+++ b/datalad/core/local/save.py
@@ -43,6 +43,7 @@ from datalad.support.parallel import (
 )
 from datalad.utils import (
     ensure_list,
+    path_startswith,
 )
 import datalad.utils as ut
 
@@ -222,9 +223,9 @@ class Save(Interface):
         # use status() to do all discovery and annotation of paths
         paths_by_ds = {}
         for s in Status()(
-                # ATTN: it is vital to pass the `dataset` argument as it,
-                # and not a dataset instance in order to maintain the path
-                # semantics between here and the status() call
+                # ATTN: it is vital to pass the `dataset` argument as is,
+                # and not cast to a dataset instance in order to maintain the
+                # path semantics between here and the status() call
                 dataset=dataset,
                 path=path,
                 untracked=untracked_mode,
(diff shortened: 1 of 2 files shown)


yarikoptic:
ha -- confirming my suspicion that most likely adjusting status should be the right way forward

<details>
<summary>if I provide intermediate subdataset path into `save` - it goes quickly</summary> 

```shell
$> time datalad save -d . -J0 -m 'Fast forward psychoinformatics-de/hirni-toolbox recursively' psychoinformatics-de/hirni-toolbox psychoinformatics-de
action summary:                                                                                                                                                                                                      
  save (notneeded: 2)
datalad save -d . -J0 -m  psychoinformatics-de/hirni-toolbox   0.65s user 0.35s system 106% cpu 0.936 total

$> time datalad save -d . -J0 -m 'Fast forward psychoinformatics-de/hirni-toolbox recursively' psychoinformatics-de/hirni-toolbox                     
add(ok): psychoinformatics-de (file)                                                                                                                                                                                 
Total (1 skipped):  50%|██████████████████████████████████████████████████████████████████████████                                                                          | 1.00/2.00 [00:00<00:00, 771 datasets/s]ERROR:                                                                                                                                                                                                               
Interrupted by user while doing magic: KeyboardInterrupt() [selectors.py:select:468]
Total (1 skipped):  50%|█████████████████████████████████████████████████████████████████████████▌                                                                         | 1.00/2.00 [00:02<00:02, 2.80s/ datasets]
datalad save -d . -J0 -m  psychoinformatics-de/hirni-toolbox  1.01s user 2.04s system 82% cpu 3.697 total

```
</details>

mih:
It seems that despite what the title is suggesting, this is not about "better" but about "faster" (the unchanged tests ran fine back than). I am not confident that I actually comprehend what it does, based on the diff. However, I would very much like to get this understood, finalized, and merged. @yarikoptic could you elaborate a bit more on what is done here. Once I can understand, I can take this over the finish line.

mih:
I am looking into this now

yarikoptic:
> It seems that despite what the title is suggesting, this is not about "better" but about "faster" (the unchanged tests ran fine back than).

not really.  It being slow IIRC was a side effect of not providing status really describing the situation, and rather just saying "untracked" (or smth like that) thus causing an expensive evaluation
 
>  I am not confident that I actually comprehend what it does, based on the diff. However, I would very much like to get this understood, finalized, and merged. @yarikoptic could you elaborate a bit more on what is done here. Once I can understand, I can take this over the finish line.

yes, diff probably could be sacrificed.  If I remember problem correctly (didn't look in detail in a while), for `datalad save -d . subds/subsubds`, instead of returning status for `subsubds` and `subds` (if needed to be saved if eg subsubds is modified), there was only a single record returned, IIRC with status "untracked" and IIRC only for `subsubds`.  That caused save to "reanalyze" the entire `subds` status to see what needs to be done to `subsubds`.  Which was leading to the slowness.  Hopefully my memory of the situation is at least somewhat correct ;)

mih:
I cannot yet make such a thing happen, but I do have a test now that fails in master and your branch, because an intermediate dataset is not considered properly. This sounds at least similar to the scenario you describe.

mih:
I will not push into this PR. I somehow believe that my approach would differ form the one here. However, it seems that the following test can trigger the problematic behavior:

```python
@with_tempfile
def test_dummy(path):
    # create three levels of datasets
    ds = Dataset(path).create()
    sub = ds.create('sub')
    subsub = sub.create('subsub')
    ds.save(recursive=True)
    # clean hierarchy
    assert_repo_status(ds.repo)

    # modify deepest level, clean save, but only that one dataset
    (subsub.pathobj / 'file').write_text('some')
    subsub.save()
    # no try to save the entire tree
    # this one emits {'state': 'untracked', 'type': 'dataset'} internally
    # which makes no sense
    ds.save(subsub.pathobj)
    # it does the job in master (possibly in an expensive fashion),
    # but fails for this PR
    assert_repo_status(ds.repo)
```

mih:
Alrighty, I will close this PR now. The rational for this is the assessment following https://github.com/datalad/datalad/issues/5521#issuecomment-971864725 and in particular https://github.com/datalad/datalad/issues/5521#issuecomment-971898039.

I do not know yet, what would be the best fix for https://github.com/datalad/datalad/issues/5521 but I am confident that the knob outlined in that issue (`need_partial_commit`) is the right one to turn -- rather than changing `save()` as done here.

Let's continue the investigation in #5521

Issue #4742, opened 2020-07-17
(Re)move automagic IO functionality?


mih:
It has issues and future plans:

- https://github.com/datalad/datalad/issues/4724
- https://github.com/datalad/datalad/issues/2062

Looking at https://github.com/datalad/datalad/blame/master/datalad/auto.py the code has not been touch in 3-5 years (minus small fixes motivated by outside changes). This either means it is perfect, or unused.

I have personally not used this functionality and I am not aware that it is being used. Given the nature of it, I feel it is more suitably placed in a dedicated package, rather than at the core of datalad.

How do you feel about this?

bpoldrack:
I feel similarly - an extension package seems more appropriate to me given how the ecosystem evolved since this was written.

mih:
I am preparing a move of code and tests to -deprecated -- analog to the webUI transition.

Issue #2062, opened 2018-01-05
ENH: enable AutomagicIO upon import by setting an environment variable


yarikoptic:
#### What is the problem?

I want to be able to trigger DataLad's AutomagicIO e.g. by simple enabling an environment variable (`DATALAD_AUTO_ENABLE=1` ?), so a tool which imports datalad (e.g. heudiconv) could then trigger its activation for the duration of that run.

adelavega:
Given that AutomagicIO is deprecated, it seems this issue can be closed?

yarikoptic:
Yeah, let's just close. Thanks @adelavega for the buzz

PR #5577, opened 2021-04-16
RF: Move AutomagicIO functionality to -deprecated

mih:
Analog to the WebUI transition, all functionality stays as-is, given the
extension is installed. If the functionality is requested , but the
extension is not installed, users are prompted to install it.

Tests are now placed in the extension too.

Closes gh-4742

when merged, move:

- #2062
- #4742


mih:
#5578 fixes the travis failure that is already in master. The appveyor failure is the mac installation issue. The crippled FS failure is #5300

datalad/__main__.py
```diff
@@ -73,6 +72,21 @@ def main(argv=None):
...
+    except Exception as e:
+        # we could just test for ModuleNotFoundError (which should be
+        # all that would happen with PY3.6+, but be a little more robust
+        # and use the pattern from duecredit
+        if type(e).__name__ not in ('ImportError', 'ModuleNotFoundError'):
+            lgr.error("Failed to import datalad_deprecated.auto "
+                      "due to %s", str(e))
+        sys.stderr.write("{}\n".format(
+            "The DataLad AutomagicIO functionality has been moved to an "
+            "extension package. Please install the Python package "
```
bpoldrack, line 86:
```suggestion
            "extension package. Please install or update the Python package "
```

bpoldrack (approved):
LGTM

datalad/__main__.py
mih: I think this is fine without the change. There is only a single release of -deprecated, and it has all that is needed. If there is a problem, the message issued by `create-sibling` also needs to be adjusted.

mih:
thx @bpoldrack for the review.

datalad/__main__.py
bpoldrack: Yeah, it's okay. Just thought might be slightly better to make that a "habit" to not require adjustment.

Issue #5546, opened 2021-04-07
Git-annex 8.20210223 is broken in a Condor job 


adswa:
I haven't pinpointed where the issue is, so the title of the issue may be misleading, and it could be something on our cluster that is at fault.

After an upgrade to git-annex 8.20210223 a previously working workflow of mine crashes. The workflow executes a shell script with a datalad based analysis through condor. Once basically any datalad-related command is executed, ~git-annex~ git complains about not finding basic shell functionality:

```
+ datalad clone ria+file:///data/group/psyinf/myinputstore#32296733-7386-4e81-afcc-37eb33f34144 ds
Traceback (most recent call last):
  File "/usr/bin/datalad", line 11, in <module>
    load_entry_point('datalad==0.14.0', 'console_scripts', 'datalad')()
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 489, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 2793, in load_entry_point
    return ep.load()
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 2411, in load
    return self.resolve()
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 2417, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "/usr/lib/python3/dist-packages/datalad/__init__.py", line 47, in <module>
    cfg = ConfigManager()
  File "/usr/lib/python3/dist-packages/datalad/config.py", line 274, in __init__
    self.reload(force=True)
  File "/usr/lib/python3/dist-packages/datalad/config.py", line 327, in reload
    self._stores[store_id] = self._reload(runargs)
  File "/usr/lib/python3/dist-packages/datalad/config.py", line 359, in _reload
    encoding='utf-8',
  File "/usr/lib/python3/dist-packages/datalad/config.py", line 717, in _run
    out = self._runner.run(self._config_cmd + args, **kwargs)
  File "/usr/lib/python3/dist-packages/datalad/cmd.py", line 413, in run
    **results,
datalad.support.exceptions.CommandError: CommandError: 'git --git-dir= config -z -l --show-origin' failed with exitcode 1 [err: '/usr/lib/git-annex.linux/git: 2: /usr/lib/git-annex.linux/git: readlink: not found
[...]

PR #5591, opened 2021-04-21
ENH(UX,DX): raise RuntimeError if PATH env var is not set

yarikoptic:
I think it is very reasonable to require for it to be set across all OSes
(IIRC).  Even though we would set PATH to point to the location where we find
git-annex using `distutils.spawn.find_executable`, I think it just buries the
issue of inadequate environment, and IMHO it is better to error out with a
clear message right away.

This should prevent needing to troubleshoot odd execution errors such as
in the case of https://github.com/datalad/datalad/issues/5546 .

### DIFF
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -514,6 +514,8 @@ class GitRunnerBase(object):
         if found.  If it is empty (but not None), we do nothing
         """
         if GitRunnerBase._GIT_PATH is None:
+            if not os.environ.get('PATH', None):
+                raise RuntimeError("PATH environment variable is either undefined or set empty.")
             from distutils.spawn import find_executable
             # with all the nesting of config and this runner, cannot use our
             # cfg here, so will resort to dark magic of environment options


bpoldrack:
I think a warning should be fine. A whole lot of things might still work just fine - no reason to fail from my POV.

yarikoptic:
warning can be not displayed (BTW, logging at that stage might not be even enabled yet fully), suppressed, etc.
I do not think I would prefer to rely on "might work". I would prefer to get a run in a broken environment to error out as early as possible so would be alerted and fix the environment, instead of e.g. running various analyses to later on just realize that apparently it worked only because of "might", fixing it up, and now using some other versions of tools because they appear in the different directory pointed by fixed `PATH`.

bpoldrack:
> I do not think I would prefer to rely on "might work". 

Well, nothing is "relying" on it. We just don't know at this point that we actually need any `PATH`. As long as we find the executables we are calling, nothing is broken really. Why am I forced to "fix" an env that works for everything I want to do? What does "fix" even mean in such a case?

> BTW, logging at that stage might not be even enabled yet fully

That may indeed be true in case of `datalad.cfg` instantiation.

I don't like failing on something b/c it may or may not become a problem. As far as better tracking of real failures goes, it may be an option to just add that notice when we really do encounter an `CommandError`?

bpoldrack:
As a compromise `run` might check on that and fail. I can't see a reason why all commands should. For `run` I kinda see why it might be desirable to not even try.

yarikoptic:
> As a compromise `run` might check on that and fail. ... For `run` I kinda see why it might be desirable to not even try. 

by `run` do you mean `datalad run` or `WitlessRunner.run`?

>  I can't see a reason why all commands should.

I can't see why shouldn't - we cannot guarantee that any and every command manages to work nicely in misconfigured environment. With all wonderful shell scripts out there without use of `set -eu` I can even come up with some nice recipes for a disaster if we are to invoke some command which does smth like ```rm -rf `dirname $PWD`/*``` and `dirname` is "not found" for the reason we discuss.

bpoldrack:
> by run do you mean datalad run or WitlessRunner.run?

I meant `datalad run`.

> I can't see why shouldn't - we cannot guarantee that any and every command manages to work nicely in misconfigured environment.

We don't need to guarantee such a thing. If we actually make call down the road, that requires something in `PATH` we would still fail. Everything that ultimately needs to fail, fails either way. You just make additional commands fail for no reason.
Why would you force me to let `PATH` point to something, that I don't need and "you" don't need and "you" don't even check what it is, only so I can run `datalad status`, `datalad subdatasets` or whatever?
How is `PATH=/tmp/foo datalad status` anyhow more sane than `PATH= datalad status`?
To me this is rather paternalistic: DataLad says "Yes, I could execute what you demanded, but I won't until you cleaned your room. Basta!" Violation of Azimov's second law, if you aks me. :-D

yarikoptic:
> How is PATH=/tmp/foo datalad status anyhow more sane than PATH= datalad status?

BOTH are insane.  Can we catch all insane cases? no.  But we can catch insane PATH= or `unset PATH` cases which must not happen in a non-insane env.

>To me this is rather paternalistic: DataLad says "Yes, I could execute what you demanded, but I won't until you cleaned your room. Basta!" Violation of Azimov's second law, if you aks me. :-D

it is not paternalistic -- If we didn't do pixie dust fixup of PATH as we do, `datalad` would have just failed as any other command, since it would have not found `git` or anything else.

For completeness, here is the pixie dust of `find_executable`, where the effect of using os.defpath is not even documented in docstring:

```python
def find_executable(executable, path=None):
    """Tries to find 'executable' in the directories listed in 'path'.

    A string listing directories separated by 'os.pathsep'; defaults to
    os.environ['PATH'].  Returns the complete filename or None if not found.
    """
...
    if path is None:
        path = os.environ.get('PATH', None)
        if path is None:
            try:
                path = os.confstr("CS_PATH")
            except (AttributeError, ValueError):
                # os.confstr() or CS_PATH is not available
                path = os.defpath
...
```

so a willing soul could even become a Python contributor but adjusting docstring to reflect the reality (use of `os.defpath`).

kyleam:
In practice, I'd guess this would help flag unintentional errors while not affecting existing and working code.  My issue with that assessment is the last part: I really don't have a sense for how likely that is to be true.

It does seem noteworthy that Python handles the unset `PATH` case with a `CS_PATH` fallback.  On my system, `confstr(3)` says

```
_CS_PATH
       A value for the PATH variable which indicates  where  all  the  POSIX.2
       standard utilities can be found.
```

`exec(3)` says

```
The file is sought in the colon-separated list
of directory pathnames specified in the PATH environment  variable.   If  this
variable isn't defined, the path list defaults to a list that includes the di‐
rectories returned by confstr(_CS_PATH) (which  typically  returns  the  value
"/bin:/usr/bin")  and  possibly  also the current working directory; see NOTES
for further details.

[...]

NOTES
       The  default search path (used when the environment does not contain the vari‐
       able PATH) shows some variation across systems.  It  generally  includes  /bin
       and  /usr/bin  (in that order) and may also include the current working direc‐
       tory.  [...]

```

---

A variation of what you show above, with perl too:

```
(unset PATH; /usr/bin/python3 -c 'import subprocess as sp; sp.run(["git", "status"])'; )
On branch master

No commits yet

nothing to commit

$ (unset PATH; /usr/bin/perl -e 'print `git status`'; )
On branch master

No commits yet
```

---

The information above is enough to make me worried that we could break scripts/analyses there were working, and I'm uneasy about applying this to maint.  I'm on the fence with whether we should try it on master and then roll it back if someone has a use case where it causes issues for them.

yarikoptic:
Thank you @kyleam .
I especially like the "and **may** also include the current working directory" part in the core doc like this ;)

yarikoptic:
since seemed there were no consensus, I will just let this PR RiP for now

PR #5595, opened 2021-04-22
Updating mailmap

vsoch:
I'm using this now for GitHub commits and would like to remove the others that are not in use. Thank you!


mih:
You're welcome!

yarikoptic:
eh, not so fast since it obliterates the fix for double counting
```
$> git shortlog -sn | grep Van
     4	Vanessa Sochat
     3	Vanessasaurus
```
will submit a follow up PR

PR #5601, opened 2021-04-25
Use the noreply account for Vanessa 

yarikoptic:
Still all previous emails used in git commits history are doomed to be
listed AFAIK to have git shortlog -sn  working

Attn @vsoch

A "redo" for #5595

With this we get
```
$> git shortlog -sn | grep Van                       
    17	Vanessa Sochat
```
so the one and only Vanessa

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -18,8 +18,9 @@ Neuroimaging Community <committer@example.com> <test@example.com>
 Neuroimaging Community <committer@example.com> unknown <test@nowhere.com> 
 Taylor Olson <to.overlake@gmail.com> Taylor Olson <taylor.m.olson.21@dartmouth.edu>
 Torsten Stoeter <torsten.stoeter@lin-magdeburg.de>
-Vanessa Sochat <vsochat@gmail.com> Vanessasaurus <vsochat@gmail.com>
-Vanessa Sochat <vsochat@gmail.com> <vsochat@stanford.edu> 
+Vanessa Sochat <vsoch@github.noreply.users.com> Vanessasaurus <vsochat@gmail.com>
+Vanessa Sochat <vsoch@github.noreply.users.com> Vanessasaurus <814322+vsoch@users.noreply.github.com>
+Vanessa Sochat <vsoch@github.noreply.users.com> <vsochat@stanford.edu>
 Yaroslav Halchenko <debian@onerussian.com> 
 Yaroslav Halchenko <debian@onerussian.com> <site-github-private@onerussian.com>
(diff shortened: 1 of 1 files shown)


vsoch:
Nope not okay with me - please remove the gmail.

.mailmap
```diff
@@ -18,7 +18,9 @@ Neuroimaging Community <committer@example.com> <test@example.com>
 Neuroimaging Community <committer@example.com> unknown <test@nowhere.com> 
 Taylor Olson <to.overlake@gmail.com> Taylor Olson <taylor.m.olson.21@dartmouth.edu>
 Torsten Stoeter <torsten.stoeter@lin-magdeburg.de>
-Vanessa Sochat <vsoch@users.noreply.github.com> 
+Vanessa Sochat <vsoch@github.noreply.users.com> Vanessasaurus <vsochat@gmail.com>
```
vsoch, line 21:
```suggestion
```

yarikoptic: If I remove it, since it is already part of the git history (so anybody could see it) for some of your commits then we will get Vanessasaurus back!:

```
$> sed -e '/@gmail.com/d' -i .mailmap 
$> git shortlog -s | grep Van        
    15	Vanessa Sochat
     2	Vanessasaurus
```

vsoch: That's fine - only the main branch is indexed in search.

yarikoptic: not sure if I am following - is it fine to merge as is?
For my use case (to see who contributed) I would prefer a single entry/line per contributor -- and that is why we keep .mailmap with all dis-ambiguations.

vsoch: Please remove the line with the gmail address / just the gmail address and then you can merge (see my suggestion).

yarikoptic: we are going in circles -- as I showed in https://github.com/datalad/datalad/pull/5601#discussion_r619887396 above, if remove that entry with gmail address - I will get 2 Vanessasaurus which are you as well .  I would strongly prefer to keep one Vanessa.
If your whole intention is to remove your gmail address - I think it is like fighting with the windmill since it is already in the git history, and google search for it claims to return > 20k hits.  So - what problem are you trying to solve?

vsoch: It’s my personal information and I have the right to remove it. As I already stated, GitHub search only indexes the main branch so it is a tiny step of progress to remove it.

yarikoptic:
Let it then be as is, this PR would not make it better with the proposed suggestion

vsoch:
Thank you, I really appreciate it. 🥰

PR #5605, opened 2021-04-26
ENH: external_versions - add lookup table for github -> pygithub

yarikoptic:
Inspired by https://github.com/datalad/datalad/pull/5603#pullrequestreview-645080883

Positioned against `maint` since it addresses us returning non-informative UNKNOWN for `github` module version whenever it can be discovered.


kyleam (approved):
LGTM

PR #5606, opened 2021-04-26
A collection of a minor OPT, typo fixes, and extra TODO(OPT)? comments

yarikoptic:
I kept those in various throw away branches while looking at redoing #5576, but I guess they might better be committed on their own first.  I am ok to drop `TODO(OPT)` comments if so desired


kyleam (approved):
The code change looks obviously correct, and the added "TODO(OPT)?" comments seem worth answering (though I didn't review the associated functions closely).

yarikoptic:
no objections voiced, 1 approval, changes are minor, so I will proceed

PR #5611, opened 2021-04-28
Fix syntax of update-contributors workflow

jwodder:
The `update-contributors.yml` workflow has been broken since December, as strings in workflow context expressions such as "`if:`" keys must use single quotes, not double-quotes.


yarikoptic:
oh hoh -- thank you @jodder!

PR #5642, opened 2021-05-05
ENH: get_status_dict - add exit_code if CommandError

yarikoptic:
Immediate use-case is foreach since it feels worthwhile channeling
upstairs the details of the failed execution, but I can imagine
it being useful for other cases too.

WDYT?

TODO:
- [x] test

### DIFF
--- a/datalad/interface/results.py
+++ b/datalad/interface/results.py
@@ -27,6 +27,7 @@ from datalad.utils import (
     path_is_subpath,
     PurePosixPath,
 )
+from datalad.support.exceptions import CommandError
 from datalad.support.path import robust_abspath
 from datalad.support.exceptions import (
     format_oneline_tb,
@@ -106,6 +107,8 @@ def get_status_dict(action=None, ds=None, path=None, type=None, logger=None,
             if isinstance(exception, CapturedException) \
             else format_oneline_tb(
                 exception, include_str=False)
+        if isinstance(exception, CommandError):
+            d['exit_code'] = exception.code
     if kwargs:
         d.update(kwargs)
     return d
--- a/datalad/interface/results.py
+++ b/datalad/interface/results.py
@@ -99,7 +99,7 @@ def get_status_dict(action=None, ds=None, path=None, type=None, logger=None,
     if exception is not None:
         d['exception'] = exception
         d['exception_traceback'] = \
-            exception.format_oneline_tb(limit=1000, include_str=False)
+            exception.format_oneline_tb(limit=None, include_str=False)
     if kwargs:
         d.update(kwargs)
     return d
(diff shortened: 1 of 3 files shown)


kyleam:
Looks useful to me.

yarikoptic:
Thank you @kyleam . Other @datalad/developers - any feedback before I proceed?

mih:
My 2ct: There is very little cost to adding useful information to a result record. The only downside is that with https://github.com/datalad/datalad/issues/6054 it is the wild-west re naming the fields. Field names will have to be static forever, because hooks depend on them, once written.

mih:
After having looked into this, we need to change the approach to make this happen. As discovered in https://github.com/datalad/datalad/pull/6167#discussion_r746504495 and documented in https://github.com/datalad/datalad/issues/6171 `get_status_dict()` can no longer handle `CommandError` exceptions -- only `CapturedException`.

@bpoldrack can you please advice how to handle this case reliably, ie. check that the captured exception was `CommandError` (possible via `.name`), and then extract the original `.code` property? I believe this is not possible (by design), but I might be wrong. And if I am correct, which approach would be a suitable alternative (excluding a long list of special cases for particular captured exception types)?

bpoldrack:
> I believe this is not possible (by design), but I might be wrong. And if I am correct, which approach would be a suitable alternative (excluding a long list of special cases for particular captured exception types)?

I think you are right. There are two ways, I can see:

1. The "ugly" one:
    When capturing the exception get that property to be able to put in a result record directly rather than "reconstructing" it after the fact within `get_status_dict`.
    
2. The cleaner, generic way in my view:
     Enhance `CapturedException` to capture attributes of the original exception and make them available (via `__getattr__` or something, I guess). In that case the PR would work as is. After all `CapturedException` is supposed to capture the relevant information about it (w/o keeping references to stack frames, etc). I didn't think of such attributes, but arguably they simpy should be part of what is captured.

mih:
I think I do not like either scenario. Any additional logic that is put into `CapturedException` will slows things down. I think I will RF this stack and rearrange functionality:

- there is no real need for `get_status_dict()` to require `CapturedException`, only that it doesn't know how to format a traceback without it -- but this can change
- take `CapturedException.format_oneline_tb()` and turn it into a standalone helper that is also accessible to `get_status_dict()`
- install a switch in `get_status_dict()`that uses this new helper properly with any passed exception
- put the change in this PR on top of it

mih:
I pushed my proposal to this PR.

mih:
This is now complete with a test and ready to merge.

bpoldrack (approved):

mih:
Thx @bpoldrack

Issue #5631, opened 2021-05-03
datalad get throws a KeyError for (sub)datasets without dataset ID


yarikoptic:
```
$> datalad --version
datalad 0.14.2.dev265

$> git -C /home/yoh/proj/datalad/datalad-master describe           
0.14.2-265-g485b769da
```

so might be a bit dated master branch:

```
(git)smaug:/mnt/datasets/datalad/crawl[master]git
$> datalad install -d . -s https://github.com/psychoinformatics-de/studyforrest-data -r studyforrest
[INFO   ] Scanning for unlocked files (this may take some time)
[INFO   ] Remote origin not usable by git-annex; setting annex-ignore
install(ok): studyforrest (dataset)
[WARNING] Got jobs=6 but we cannot use threads with Pythons versions prior 3.8.0. Will run serially
add(ok): studyforrest (file)
add(ok): .gitmodules (file)
save(ok): . (dataset)
[INFO   ] Installing Dataset(/mnt/datasets/datalad/crawl/studyforrest) to get /mnt/datasets/datalad/crawl/studyforrest recursively
[INFO   ] Fetching 'https://datapub.fz-juelich.de/studyforrest/studyforrest.ria/0f6/6b1ba-e9a9-46fd-b9d9-2e64fe94d307/config'
install(ok): studyforrest/code/conversion_qa (dataset)
[INFO   ] Scanning for unlocked files (this may take some time)
[ERROR  ] Target commit 2ccaa115 does not exist in the clone, and a fetch that commit from origin failed [install(/mnt/datasets/datalad/crawl/studyforrest/derivative/aggregate_fmri_timeseries)]
                                                                                                                                                                                                                     install(error): studyforrest/derivative/aggregate_fmri_timeseries (dataset) [Target commit 2ccaa115 does not exist in the clone, and a fetch that commit from origin failed]
                                                                                                                                                                                                                     [INFO   ] Scanning for unlocked files (this may take some time)
                                                                                                                                                                                                                     [ERROR  ] Target commit 74cd7ec0 does not exist in the clone, and a fetch that commit from origin failed [install(/mnt/datasets/datalad/crawl/studyforrest/derivative/aligned_mri)]
...
install(error): studyforrest/original/7T_multiresolution_fmri (dataset) [Target commit 10e23aaf does not exist in the clone, and a fetch that commit from origin failed]
                                                                                                                                                                                                                     [INFO   ] Scanning for unlocked files (this may take some time)                                                                                                                                                      
[...]

PR #5644, opened 2021-05-06
BF: Catch key error when a submodule lacks a required property

adswa:
During flexible clone candidate generation, it can happen that a submodule does not contain
a property that the url template requires. One example for this arises in #5631 where a
superdataset that is cloned from a RIA store passes its subdataset-source-candidate configuration
to its subdatasets, and one of the subdatasets is a pure Git repository that lacks the dataset
id property required for the url template from the superds' configuration.
In these cases, we run into a key error and crash.

This change catches any key error (but logs it at the debug level)

@bpoldrack , do you think that this is sufficient?

- [x] add a test

### DIFF
--- a/datalad/core/distributed/clone.py
+++ b/datalad/core/distributed/clone.py
@@ -327,7 +327,8 @@ def clone_dataset(
         description=None,
         result_props=None,
         cfg=None,
-        checkout_gitsha=None):
+        checkout_gitsha=None,
+        misbuilt_url=None):
     """Internal helper to perform cloning without sanity checks (assumed done)
 
     This helper does not handle any saving of subdataset modification or adding
@@ -356,6 +357,11 @@ def clone_dataset(
       did not obtain the commit object. Should the checkout of the target commit
       cause a detached HEAD, the previously active branch will be reset to the
       target commit.
+    misbuilt_url : dict, optional
+      A dictionary of custom submodule source candidate locations that resulted
+      in a key error during flexible source candidate generation. Should cloning
+      fail eventually, the information on unsuccessfully built clone candidates
+      is reported in the result summary.
 
     Yields
     ------
@@ -463,8 +469,12 @@ def clone_dataset(
 
         except CommandError as e:
             e_stderr = e.stderr
-
             error_msgs[cand['giturl']] = e
+            # we might have gotten here because a misbuilt clone candidate
+            # for a submodule left no other choice than to patch a broken git
+            # url together
+            if misbuilt_url:
+                _add_hint_on_misbuilt_urls(misbuilt_url, error_msgs)
             lgr.debug("Failed to clone from URL: %s (%s)",
                       cand['giturl'], exc_str(e))
             if dest_path.exists():
@@ -1324,6 +1334,36 @@ def decode_source_spec(spec, cfg=None):
         # if we still have no good idea on where a dataset could be cloned to if no
         # path was given, do something similar to git clone and derive the path from
         # the source
-        props['default_destpath'] = _get_installationpath_from_url(props['giturl'])
+        props['default_destpath'] = _get_installationpath_from_url(
+            props['giturl'])
 
     return props
+
+
+def _add_hint_on_misbuilt_urls(urls, error_msgs):
+    """If cloning failed, but a KeyError was caught in the generation of a
+    submodule clone candidate URL, the misbuilt submodule URL template is
+    reported to the user
+
+    Parameters
+    ----------
+    urls : dict
+      contains a submodule and unmatched URL template
+
+    Returns
+    -------
+    error_msgs : OrderedDict
+      encountered cloning errors
+    """
+    # import pdb; pdb.set_trace()
+    from datalad.ui import ui
+    from datalad.support import ansi_colors
+    intro = ansi_colors.color_word(
+        "Hint: A configuration for custom subdataset clone candidates exists, "
+        "but some key(s) used in the template were missing: ",
+        ansi_colors.YELLOW)
+    hint = ansi_colors.color_word('\n- '.join(
+        '{} {}'.format(subds, tmpl)
+        for subds, tmpl in urls.items()),
+        ansi_colors.YELLOW)
+    error_msgs[intro] = hint
--- a/datalad/core/distributed/clone.py
+++ b/datalad/core/distributed/clone.py
@@ -1355,15 +1355,12 @@ def _add_hint_on_misbuilt_urls(urls, error_msgs):
     error_msgs : OrderedDict
       encountered cloning errors
     """
-    # import pdb; pdb.set_trace()
-    from datalad.ui import ui
-    from datalad.support import ansi_colors
-    intro = ansi_colors.color_word(
-        "Hint: A configuration for custom subdataset clone candidates exists, "
-        "but some key(s) used in the template were missing: ",
-        ansi_colors.YELLOW)
-    hint = ansi_colors.color_word('\n- '.join(
-        '{} {}'.format(subds, tmpl)
-        for subds, tmpl in urls.items()),
-        ansi_colors.YELLOW)
-    error_msgs[intro] = hint
+    for remote in urls.keys():
+        from datalad.support import ansi_colors
+        formatted_msg = ansi_colors.color_word(
+                             urls[remote]['msg'],
+                             ansi_colors.YELLOW)
+        msg = '{} {}'.format(
+                 formatted_msg,
+                 urls[remote]['error'])
+        error_msgs[remote] = msg
--- a/datalad/distribution/get.py
+++ b/datalad/distribution/get.py
@@ -247,6 +247,16 @@ def _get_flexible_source_candidates_for_submodule(ds, sm):
                         ds_repo.config[c])
                        for c in ds_repo.config.keys()
                        if c.startswith(candcfg_prefix)]:
+        # ensure that there is only one template of the same name
+        if type(tmpl) == tuple and len(tmpl) > 1:
+            lgr.debug('Found %i URL templates for the submodule clone '
+                      'candidate %s, but only one is allowed. Check your '
+                      'dataset configuration!', len(tmpl), name)
+            raise ValueError(
+                "There are multiple URL templates for submodule clone "
+                "candidate '{}', but only one is allowed. "
+                "Check your configuration!".format(name)
+            )
         try:
             url = tmpl.format(**sm_candidate_props)
         except KeyError as e:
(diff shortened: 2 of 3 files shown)


datalad/distribution/get.py
```diff
@@ -243,7 +243,15 @@ def _get_flexible_source_candidates_for_submodule(ds, sm):
...
                        if c.startswith(candcfg_prefix)]:
-        url = tmpl.format(**sm_candidate_props)
+        try:
+            url = tmpl.format(**sm_candidate_props)
+        except KeyError:
+            # don't crash if the submodule lacks properties required by the
+            # template - this can happen if a pure Git repository is queried
+            # for a dataset id, for example.
+            lgr.debug('Caught a key error in the generation of a submodule '
+                      'URL using the template {}'.format(tmpl))
```
yarikoptic, line 253: Minor OPT: don't use format for logging messages, use % interpolation and pass arguments to debug call. See other invocations of logging for examples

yarikoptic: Also it is usually a good manner to log exception itself, helps to debug

adswa: thanks for the tips!

bpoldrack (commented):
Yes, I think that will do.
However, I think we should RF `get` a bit, so, that we that we get a hold on that "error" outside of that function and can decide to include it in the error result if `get` ultimately doesn't succeed.

adswa:
> Yes, I think that will do.
> However, I think we should RF `get` a bit, so, that we that we get a hold on that "error" outside of that function and can decide to include it in the error result if `get` ultimately doesn't succeed.
> 
> Edit: So, depending on your eagerness to try or not, you could either have a look into that or just make that an "UX" issue for someone else to solve, @adswa

Sorry, I missed your response. Thanks for your thoughts. Ultimately, (if I see it correctly), if ``get`` fails, it is ``clone`` that reports the final ``"Failed to clone from any candidate source URL. "`` So I should collect the mis-built URLs in get, pass them on into clone, and provide them as a hint in ``clone``'s result rendering, right?

adswa:
I have added a sketch on how these errors could be propagated to the user here. 
Here is the difference in output when trying to retrieve a subdataset with a URL template that uses a key that can't be matched (and no other working URL)...

... before catching the KeyError:
Screenshot from 2021-05-31 21-45-32

... after catching the KeyError, but without reports of it
Screenshot from 2021-05-31 21-46-11

... with reports of the key error:
Screenshot from 2021-05-31 21-44-20

When a working URL exists, the key error is not reported unless one is in debug mode:
Screenshot from 2021-05-31 22-00-35

datalad/core/distributed/clone.py
```diff
@@ -1311,6 +1321,28 @@ def decode_source_spec(spec, cfg=None):
...
 
     return props
+
+
+def _add_hint_on_misbuilt_urls(urls, error_msgs):
+    """If cloning failed, but a KeyError was caught in the generation of a
+    submodule clone candidate URL, the misbuilt submodule URL template is
+    reported to the user
+    :param urls: dict; contains a submodule and unmatched URL template
+    :return: error_msgs; OrderedDict of encountered cloning errors
```
yarikoptic, line 1335: Let's do consistently numpy style docstrings

datalad/core/distributed/clone.py
```diff
@@ -1311,6 +1321,28 @@ def decode_source_spec(spec, cfg=None):
...
+    """
+    # import pdb; pdb.set_trace()
+    from datalad.ui import ui
+    from datalad.support import ansi_colors
+    intro = ansi_colors.color_word(
+        "Hint: A configuration for custom subdataset clone candidates exists, "
+        "but no key matched the template: ",
```
yarikoptic, line 1342: Might be that some keys did match. 
```suggestion
        "but some key(s) used in the template were missing: ",
```
and unfortunately we don't channel to user here which key

datalad/core/distributed/clone.py
adswa: done, thx

datalad/core/distributed/clone.py
adswa: this is better, I added it. I'm also reporting the key now.

adswa:
rebased onto master to get the fix in #5705

bpoldrack:
Generally I'm kinda fine with that approach. However, it seems a bit too specific to the `KeyError` to me. There are all kinds of URLs that may be tried and fail for different reasons. I think a more generalized approach would be fine. A result dict could just get a `hints` entry in addition and whenever something goes wrong the respective routine can add such suggestions. And result renderers can decide whether or not to make use of that field. That could be done after this PR - no need to do it right now, but that's what I'd really like :)

adswa:
Thanks for your thoughts, @bpoldrack! :) Just to make sure I understand: 

> Right now it's the result renderer, that has explicit knowledge about this being a KeyError. What I'm saying is, that this should stay where we caught the error, whereas the renderer does nothing more than considering the hints field it got in the dict.

Does that mean I shouldn't pass the ``exc_str(e)`` to the ``misbuilt_url`` dict?

bpoldrack:
@adswa:

> Does that mean I shouldn't pass the exc_str(e) to the misbuilt_url dict?

Not neccessarily, but yes -  I think the exception isn't useful. This is:

> "Hint: A configuration for custom subdataset clone candidates exists ..."

And this could be formed and passed to that dict right where the exception occurs. It's not like you later check on additional info to decide whether that is the hint for that URL. So you could form that message right there and have the result renderer simply get the message from that dict. That way we could have similar things elsewhere. A different kind of malforming with a different hint passed into the dict, no change to the result renderer.

adswa:
will do, thanks!

datalad/distribution/get.py
```diff
@@ -264,8 +264,15 @@ def _get_flexible_source_candidates_for_submodule(ds, sm):
             # template - this can happen if a pure Git repository is queried
             # for a dataset id, for example.
             lgr.debug('Caught a key error in the generation of a submodule '
-                      'URL using the template %s (%s)', tmpl, exc_str(e))
-            misbuilt_url[name] = (tmpl, e)
+                      'URL using the template %s (%s). The URL will not be '
+                      'constructed.', tmpl, exc_str(e))
+            from datalad.support import ansi_colors
+            msg = ansi_colors.color_word(
```
adswa, line 270: @bpoldrack : unsure if it is good practice to do the ansi colors at this stage rather than at result rendering. Could you let me know?

bpoldrack: In my view, it should be done at rendering. Here we don't know yet (or shouldn't) whether there even is a terminal output, etc.
So, I'd just have the renderer rendering this particular field in that way, while field itself only contains text.

adswa: ok, thx

datalad/core/distributed/clone.py
```diff
@@ -327,7 +327,8 @@ def clone_dataset(
         description=None,
         result_props=None,
         cfg=None,
-        checkout_gitsha=None):
+        checkout_gitsha=None,
+        misbuilt_url=None):
```
yarikoptic, line 331: although it is just an internal (albeit without `_` prefix) function, so we could stick anything into its signature, this `misbuilt_url` "memo"  or "info" of a kind makes me wonder, and please pardon if you already explored that - I failed to find in current comments, why just not to yield some "error" record with some annotation which higher level logic could then collect/filter and report on if so desired?

[*] btw , docstring for it isn't explicit that clone_dataset just adds to it and nohow uses it otherwise

mih: I echo the notion of "rather use the results" instead of in-place modification of some internal construct. But I am not in favor of "some error result", if that means changing the *number* of results reported (see additional comment of mine in a second).

mih:
Without having looked too closely at the current implementation, here are a few comment re decision making and reporting that are partially informed by other contexts (e.g. #5709). Nothing of what I write here is meant to be taken as "instead of what is done here, I'd do". I might later get the chance to actually make this contrast.

- I believe we want to use results as primary means of reporting (not log messages), because we can act on them in downstream code most easily
- I believe there should be a single result per subdataset, not a result per subdataset clone attempt -- because `get <subds>` is looking to get that subdataset, and `impossible` results should only happen when it is actually impossible, and not "not yet successfull"
- Information a user could be interested in are:
    -  all attempt *made*, when no attempt worked out
    - any attempt *not made*, because the stars did not align properly (configuration prevents it, incompatible with particular subdataset properties (i.e. the original issue) -- possibly terse summary as a note even when the `get` succeeded nevertheless (?), but certainly a more detailed report on failure


Taken together I think this calls for an extension of the result record with all the information that a post-hoc inspection and result rendering might benefit from. That does not and should not preclude additional runtime logging/progress info.

datalad/core/distributed/clone.py
adswa: Sorry, I lost track of this PR. Can you help me understand how I could implement what you mean with:

>  why just not to yield some "error" record with some annotation which higher level logic could then collect/filter and report on if so desired

Is it that I should attempt cloning with the URL where the key error would originate, and add something to the result record that hints at the missing property in the template?

adswa:
btw 4dd9aac contains a fix for #5715. Would you prefer this to be a separate PR?

datalad/core/distributed/clone.py
yarikoptic: I was thinking about smth like
```
                yield get_status_dict(
                    status='warning',
                    message="URL was misbuilt: {some_details}"
                    **result_props)
                continue
```
but given @mih 's comments "above and [below](https://github.com/datalad/datalad/pull/5644#issuecomment-860576337)  " on an expectation that this particular call should have only a singular status record, then indeed could/should go into some other extra field within the resultant record.

but then the question is also -- why should we try misbuilt urls at all?  so this spot doesn't seem to be the right one to handle it IMHO, should be done upstairs (thus not even calling `clone_dataset` if we know that URL is not good) and then just resultant record extended with that extra information.  Or am I missing some use case?

adswa: I don't think you have missed a usecase, I think I just lacked insight on how to do it more cleverly. :) But how can I include one or more bad url(s) in the result record "upstairs" - isn't the final result rendered only in clone? How can I pass it into the record without passing it to clone?
If it would speed things up, I can also force push my earlier commits away, and leave it to someone else to commit a better approach, result rendering is still one of the many mysteries to me...

adswa:
This PR is stale, but contains two useful fixes, one just probably in a bad implementation. If someone wants to take this over, please do. Else, I'd propose to cherry-pick 4dd9aac into its own PR (this change didn't seem to be controversial), and close this one unmerged.

adswa:
> I'd propose to cherry-pick [4dd9aac](https://github.com/datalad/datalad/commit/4dd9aacef62d05f0a1c429234721360c213d6071) into its own PR (this change didn't seem to be controversial)

Ha, I didn't realized I already did this in #5839. Will close this PR then. The issue remains unaddressed, sadly.

Issue #5409, opened 2021-02-04
Jupyter lab error


k-sap:
#### What is the problem?
I try to import datalad package in Jupyter Lab Notebook. And this results in `RuntimeError: Cannot run the event loop while another loop is running`.
#### What steps will reproduce the problem?
Create new Python Jupyter Lab Notebook and try to import datalad.
#### What version of DataLad are you using (run `datalad --version`)? On what operating system (consider running `datalad wtf`)?
datalad 0.14.0
Ubuntu 20.04
#### Is there anything else that would be useful to know in this context?
This happened to me after update to Datalad 0.14. Before, everything was ok.
#### Have you had any success using DataLad before? (to assess your expertise/prior luck.  We would welcome your testimonial additions to https://github.com/datalad/datalad/wiki/Testimonials as well)
I'm figuring it out.

kyleam:
Thanks for reporting.  We've been fumbling around with asyncio stuff.  gh-5106 (0.13.6) prevented an "Event loop is closed" issue when running from IPython.  In 0.14, the two asyncio related changes were gh-5350 and gh-5367.

> This happened to me after update to Datalad 0.14. Before, everything was ok.

Do you recall what version of DataLad you were on before?  0.13.7?

Also, what version of Python and Jupyter are you using?  And do you see a traceback?  That could help pinpoint where this is happening.

k-sap:
Thanks for response.

Python 3.8.5.
JupyterLab 3.0.7

'
/home/user/.local/lib/python3.8/site-packages/datalad/cmd.py:377: RuntimeWarning: coroutine 'run_async_cmd' was never awaited
  new_loop = True
RuntimeWarning: Enable tracemalloc to get the object allocation traceback

---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-1-2677c3f1f0d7> in <module>
----> 1 import datalad

~/.local/lib/python3.8/site-packages/datalad/__init__.py in <module>
     45 
     46 from .config import ConfigManager
---> 47 cfg = ConfigManager()
     48 
     49 from .log import lgr

~/.local/lib/python3.8/site-packages/datalad/config.py in __init__(self, dataset, overrides, source)
    272             self._runner = GitWitlessRunner(**run_kwargs)
    273 
--> 274         self.reload(force=True)
    275 
    276         if not ConfigManager._checked_git_identity:

~/.local/lib/python3.8/site-packages/datalad/config.py in reload(self, force)
[...]

PR #5667, opened 2021-05-17
NF: implement a non asyncio-based runner

### LINKED ISSUE #5422: NotImplementedError: nest_asyncio.apply() for datalad.api not sufficient
Continuation of issue #5409 

Suggested workaround:
```python
import nest_asyncio
nest_asyncio.apply()
import datalad.api as dl
```
...did not work (in this case on Windows 10)

```
NotImplementedError
<ipython-input-3-4f86fa9c5d49> in <module>
      3 import nest_asyncio
      4 nest_asyncio.apply()
----> 5 import datalad.api as dl
```

<details>
<summary>file from the link below</summary> 

```shell

!pip install nest_asyncio
Requirement already satisfied: nest_asyncio in c:\programdata\anaconda3\lib\site-packages (1.4.2)


import os
import glob
import nest_asyncio
nest_asyncio.apply()
import datalad.api as dl

NotImplementedError                       Traceback (most recent call last)
<ipython-input-3-4f86fa9c5d49> in <module>
      3 import nest_asyncio
      4 nest_asyncio.apply()
----> 5 import datalad.api as dl

C:\ProgramData\Anaconda3\lib\site-packages\datalad\__init__.py in <module>
     45 
     46 from .config import ConfigManager
---> 47 cfg = ConfigManager()
     48 
     49 from .log import lgr

C:\ProgramData\Anaconda3\lib\site-packages\datalad\config.py in __init__(self, dataset, overrides, source)
    272             self._runner = GitWitlessRunner(**run_kwargs)
    273 
--> 274         self.reload(force=True)
    275 
    276         if not ConfigManager._checked_git_identity:

C:\ProgramData\Anaconda3\lib\site-packages\datalad\config.py in reload(self, force)
    325         while to_run:
    326             store_id, runargs = to_run.popitem()
--> 327             self._stores[store_id] = self._reload(runargs)
    328 
    329         # always update the merged representation, even if we did not reload

C:\ProgramData\Anaconda3\lib\site-packages\datalad\config.py in _reload(self, run_args)
    353     def _reload(self, run_args):
    354         # query git-config
--> 355         stdout, stderr = self._run(
    356             run_args,
    357             protocol=StdOutErrCapture,

C:\ProgramData\Anaconda3\lib\site-packages\datalad\config.py in _run(self, args, where, reload, **kwargs)
    715         if '-l' in args:
    716             # we are just reading, no need to reload, no need to lock
--> 717             out = self._runner.run(self._config_cmd + args, **kwargs)
    718             return out['stdout'], out['stderr']
    719 

C:\ProgramData\Anaconda3\lib\site-packages\datalad\cmd.py in run(self, cmd, protocol, stdin, cwd, env, **kwargs)
    378         try:
    379             # include the subprocess manager in the asyncio event loop
--> 380             results = event_loop.run_until_complete(
    381                 run_async_cmd(
    382                     event_loop,

C:\ProgramData\Anaconda3\lib\site-packages\nest_asyncio.py in run_until_complete(self, future)
     94                 raise RuntimeError(
     95                     'Event loop stopped before Future completed.')
---> 96             return f.result()
     97         finally:
     98             events._set_running_loop(old_running_loop)

C:\ProgramData\Anaconda3\lib\asyncio\futures.py in result(self)
    176         self.__log_traceback = False
    177         if self._exception is not None:
--> 178             raise self._exception
    179         return self._result
    180 

C:\ProgramData\Anaconda3\lib\asyncio\tasks.py in __step(***failed resolving arguments***)
    278                 # We use the `send` method directly, because coroutines
    279                 # don't have `__iter__` and `__next__` methods.
--> 280                 result = coro.send(None)
    281             else:
    282                 result = coro.throw(exc)

C:\ProgramData\Anaconda3\lib\site-packages\datalad\cmd.py in run_async_cmd(loop, cmd, protocol, stdin, protocol_kwargs, **kwargs)
    102     try:
    103         lgr.debug('Launching process %s', cmd)
--> 104         transport, protocol = await proc
    105         lgr.debug('Waiting for process %i to complete', transport.get_pid())
    106         # The next wait is a workaround that avoids losing the output of

C:\ProgramData\Anaconda3\lib\asyncio\base_events.py in subprocess_exec(self, protocol_factory, program, stdin, stdout, stderr, universal_newlines, shell, bufsize, encoding, errors, text, *args, **kwargs)
   1628             debug_log = f'execute program {program!r}'
   1629             self._log_subprocess(debug_log, stdin, stdout, stderr)
-> 1630         transport = await self._make_subprocess_transport(
   1631             protocol, popen_args, False, stdin, stdout, stderr,
   1632             bufsize, **kwargs)

C:\ProgramData\Anaconda3\lib\asyncio\base_events.py in _make_subprocess_transport(self, protocol, args, shell, stdin, stdout, stderr, bufsize, extra, **kwargs)
    489                                          extra=None, **kwargs):
    490         """Create subprocess transport."""
--> 491         raise NotImplementedError
    492 
    493     def _write_to_self(self):

NotImplementedError: 

```
</details>

[Datalad_Asyncio_NotImplemented_20210216.txt](https://github.com/datalad/datalad/files/6002091/Datalad_Asyncio_NotImplemented_20210216.txt)

christian-monch:
PR for an non asyncio-based runner.

This patch implements a thread-based runner, that uses the asyncio protocol interface to communicate
process output to protocol instances. 

This change was done to support running datalad commands from processes within a concurrent.futures.ProcessPoolExecutor. This did not work reliably if the datalad command invoked subprocesses, and if asyncio-based subprocess were used. In the asyncio-based implementation processes would randomly wait forever on some synchronization primitive.

Motivation: I am creating a *pipeline*/*orchestration* command for metalad that allows a user to specify an information source element, e.g. a dataset-traverser, and a number of processing elements, e.g. a metadata-extractor and a metadata-adder, which would be run on the output of the previous element. So the pipeline would look something like this:

   [dataset-traverser->metadata-extractor->metadata-adder]

And I would like to use concurrent.futures (or some other process-based parallelization) to distribute pipeline element processing to multiple cores.

This PR is marked as draft-pull request, since it changes a central piece of datalad, and since I did not yet test it on Windows or MacOS

Closes #5422, #5409, #5100

TODOs:
- [x] docstrings for added functions
- [x] add a `test_popen_invocation` from https://github.com/datalad/datalad/pull/5369/files#diff-4aa1a9688db72f7a475890e0ee6a343ea98af70ccc3bfa66f17da82c819dfa8dR230
- [x] add an explicit test to verify its operation within an asyncio event_loop (ref  #5100)

### DIFF
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -535,6 +535,9 @@ class GitWitlessRunner(WitlessRunner, GitRunnerBase):
 
 
 def readline_rstripped(stdout):
+    warnings.warn("the function `readline_rstripped()` is deprecated "
+                  "and will be removed in a future release",
+                  DeprecationWarning)
     return stdout.readline().rstrip()
 
 
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -15,19 +15,13 @@ import subprocess
 import sys
 import logging
 import os
-import functools
 import tempfile
 from locale import getpreferredencoding
 import asyncio
-from collections import (
-    namedtuple,
-)
+import warnings
 
 from .consts import GIT_SSH_COMMAND
-from .dochelpers import (
-    borrowdoc,
-    exc_str,
-)
+from .dochelpers import borrowdoc
 from .support import path as op
 from .support.exceptions import CommandError
 from .utils import (
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -22,7 +22,6 @@ import warnings
 
 from .consts import GIT_SSH_COMMAND
 from .dochelpers import borrowdoc
-from .nonasyncrunner import run_command
 from .support import path as op
 from .support.exceptions import CommandError
 from .utils import (
@@ -290,6 +289,10 @@ class WitlessRunner(object):
         FileNotFoundError
           When a given executable does not exist.
         """
+
+        # imported here to avoid recursive importing
+        from .nonasyncrunner import run_command
+
         if protocol is None:
             # by default let all subprocess stream pass through
             protocol = NoCapture
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -144,11 +144,11 @@ class WitlessProtocol(asyncio.SubprocessProtocol):
             self.process.pid, return_code)
         # give captured process output back to the runner as string(s)
         results = {
-            self.fd_infos[fileno][0]: (
-                bytes(self.fd_infos[fileno][1]).decode(self.encoding)
-                if self.fd_infos[fileno][1] is not None
+            name: (
+                bytes(byt).decode(self.encoding)
+                if byt is not None
                 else '')
-            for fileno in (self.stdout_fileno, self.stderr_fileno)
+            for name, byt in self.fd_infos.items()
         }
         results['code'] = return_code
         return results
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -137,7 +137,8 @@ class WitlessProtocol(asyncio.SubprocessProtocol):
         'cmd' and 'cwd' on error, if they are present in the result.
         """
         return_code = self.process.poll()
-        assert return_code is not None
+        if return_code is None:
+            raise CommandError("Got None as a return_code for the process %i", self.process.pid)
         lgr.debug(
             'Process %i exited with return code %i',
             self.process.pid, return_code)
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -22,6 +22,7 @@ import warnings
 
 from .consts import GIT_SSH_COMMAND
 from .dochelpers import borrowdoc
+from .nonasyncrunner import run_command
 from .support import path as op
 from .support.exceptions import CommandError
 from .utils import (
@@ -138,7 +139,9 @@ class WitlessProtocol(asyncio.SubprocessProtocol):
         """
         return_code = self.process.poll()
         if return_code is None:
-            raise CommandError("Got None as a return_code for the process %i", self.process.pid)
+            raise CommandError(
+                "Got None as a return_code for the process %i",
+                self.process.pid)
         lgr.debug(
             'Process %i exited with return code %i',
             self.process.pid, return_code)
@@ -296,8 +299,6 @@ class WitlessRunner(object):
             env or self.env,
             cwd=cwd,
         )
-
-        from .nonasyncrunner import run_command
 
         lgr.debug('run:\n cwd=%s\n cmd=%s', cwd, cmd)
         results = run_command(
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -67,14 +67,21 @@ class WitlessProtocol(asyncio.SubprocessProtocol):
     proc_out = None
     proc_err = None
 
-    def __init__(self, encoding=None):
+    def __init__(self, done_future=None, encoding=None):
         """
         Parameters
         ----------
+        done_future: Any
+          Ignored parameter, kept for backward compatibility (DEPRECATED)
         encoding : str
           Encoding to be used for process output bytes decoding. By default,
           the preferred system encoding is guessed.
         """
+
+        if done_future is not None:
+            warnings.warn("`done_future` argument is ignored "
+                          "and will be removed in a future release",
+                          DeprecationWarning)
 
         self.fd_infos = {}
 
@@ -532,7 +539,6 @@ class GitWitlessRunner(WitlessRunner, GitRunnerBase):
 
 
 def readline_rstripped(stdout):
-    #return iter(stdout.readline, b'').next().rstrip()
     return stdout.readline().rstrip()
 
 
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -333,36 +333,6 @@ class WitlessRunner(object):
         results.pop('code', None)
         return results
 
-    @classmethod
-    def _check_if_new_proc(cls):
-        """Check if WitlessRunner is used under a new PID
-
-        Note that this is a function that is meant to be called from within a
-        particular context only. The RuntimeError is expected to be caught by
-        the caller and is meant to be more like a response message than an
-        exception.
-
-        Returns
-        -------
-        bool
-
-        Raises
-        ------
-        RuntimeError
-          If it is not a new proc and we already know that we need a new loop
-          in this pid
-        """
-        pid = os.getpid()
-        is_new_proc = cls._loop_pid is None or cls._loop_pid != pid
-        if is_new_proc:
-            # We need to check if we can run any command smoothly
-            cls._loop_pid = pid
-            cls._loop_need_new = None
-        elif cls._loop_need_new:
-            raise RuntimeError("we know we need a new loop")
-        return is_new_proc
-
-
 class GitRunnerBase(object):
     """
     Mix-in class for Runners to be used to run git and git annex commands
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -50,70 +50,6 @@ _TEMP_std = sys.stdout, sys.stderr
 _MAGICAL_OUTPUT_MARKER = "_runneroutput_"
 
 
-async def run_async_cmd(loop, cmd, protocol, stdin, protocol_kwargs=None,
-                        **kwargs):
-    """Run a command in a subprocess managed by asyncio
-
-    This implementation has been inspired by
-    https://pymotw.com/3/asyncio/subprocesses.html
-
-    Parameters
-    ----------
-    loop : asyncio.AbstractEventLoop
-      asyncio event loop instance. Must support subprocesses on the
-      target platform.
-    cmd : list or str
-      Command to be executed, passed to `subprocess_exec` (list), or
-      `subprocess_shell` (str).
-    protocol : WitlessProtocol
-      Protocol class to be instantiated for managing communication
-      with the subprocess.
-    stdin : file-like or None
-      Passed to the subprocess as its standard input.
-    protocol_kwargs : dict, optional
-      Passed to the Protocol class constructor.
-    kwargs : Pass to `subprocess_exec`, will typically be parameters
-      supported by `subprocess.Popen`.
-
-    Returns
-    -------
-    undefined
-      The nature of the return value is determined by the given
-      protocol class.
-    """
-    if protocol_kwargs is None:
-        protocol_kwargs = {}
-    cmd_done = asyncio.Future(loop=loop)
-    factory = functools.partial(protocol, cmd_done, **protocol_kwargs)
-    kwargs.update(
-        stdin=stdin,
-        # ask the protocol which streams to capture
-        stdout=asyncio.subprocess.PIPE if protocol.proc_out else None,
-        stderr=asyncio.subprocess.PIPE if protocol.proc_err else None,
-    )
-    if isinstance(cmd, str):
-        proc = loop.subprocess_shell(factory, cmd, **kwargs)
-    else:
-        proc = loop.subprocess_exec(factory, *cmd, **kwargs)
-    transport = None
-    result = None
-    try:
-        lgr.debug('Launching process %s', cmd)
-        transport, protocol = await proc
-        lgr.debug('Waiting for process %i to complete', transport.get_pid())
-        # The next wait is a workaround that avoids losing the output of
-        # quickly exiting commands (https://bugs.python.org/issue41594).
-        await asyncio.ensure_future(transport._wait())
-        await cmd_done
-        result = protocol._prepare_result()
-    finally:
-        # protect against a crash when launching the process
-        if transport:
-            transport.close()
-
-    return result
-
-
 class WitlessProtocol(asyncio.SubprocessProtocol):
     """Subprocess communication protocol base class for `run_async_cmd`
 
@@ -128,29 +64,28 @@ class WitlessProtocol(asyncio.SubprocessProtocol):
     and a list with bytestrings of all captured output streams.
     """
 
-    FD_NAMES = ['stdin', 'stdout', 'stderr']
-
     proc_out = None
     proc_err = None
 
-    def __init__(self, done_future, encoding=None):
+    def __init__(self, encoding=None):
         """
         Parameters
         ----------
-        done_future : asyncio.Future
-          Future promise to be fulfilled when process exits.
         encoding : str
           Encoding to be used for process output bytes decoding. By default,
           the preferred system encoding is guessed.
         """
-        self.done = done_future
+
+        self.fd_infos = {}
+
+        self.process = None
+        self.stdout_fileno = 1
+        self.stderr_fileno = 2
+
         # capture output in bytearrays while the process is running
-        Streams = namedtuple('Streams', ['out', 'err'])
-        self.buffer = Streams(
-            out=bytearray() if self.proc_out else None,
-            err=bytearray() if self.proc_err else None,
-        )
-        self.pid = None
+        self.fd_infos[self.stdout_fileno] = ("stdout", bytearray()) if self.proc_out else ("stdout", None)
+        self.fd_infos[self.stderr_fileno] = ("stderr", bytearray()) if self.proc_err else ("stderr", None)
+
         super().__init__()
         self.encoding = encoding or getpreferredencoding(do_setlocale=False)
 
@@ -169,9 +104,9 @@ class WitlessProtocol(asyncio.SubprocessProtocol):
         pass
 
     def _log_summary(self, fd, data):
-        fd_name = self.FD_NAMES[fd]
+        fd_name = self.fd_infos[fd]
         lgr.log(5, 'Read %i bytes from %i[%s]%s',
-                len(data), self.pid, fd_name, ':' if self._log_outputs else '')
+                len(data), self.process.pid, fd_name, ':' if self._log_outputs else '')
         if self._log_outputs:
             log_data = ensure_unicode(data)
             # The way we log is to stay consistent with Runner.
@@ -179,16 +114,16 @@ class WitlessProtocol(asyncio.SubprocessProtocol):
             # fd_name prefix
             lgr.log(5, "%s| %s ", fd_name, log_data)
 
-    def connection_made(self, transport):
-        self.transport = transport
-        self.pid = transport.get_pid()
-        lgr.debug('Process %i started', self.pid)
+    def connection_made(self, process):
+        self.process = process
+        lgr.debug('Process %i started', self.process.pid)
 
     def pipe_data_received(self, fd, data):
         self._log(fd, data)
         # store received output if stream was to be captured
-        if self.buffer[fd - 1] is not None:
-            self.buffer[fd - 1].extend(data)
+        fd_name, buffer = self.fd_infos[fd]
+        if buffer is not None:
+            buffer.extend(data)
 
     def _prepare_result(self):
         """Prepares the final result to be returned to the runner
@@ -200,23 +135,24 @@ class WitlessProtocol(asyncio.SubprocessProtocol):
         this exception class as kwargs on error. The Runner will overwrite
         'cmd' and 'cwd' on error, if they are present in the result.
         """
-        return_code = self.transport.get_returncode()
+        return_code = self.process.poll()
+        assert return_code is not None
         lgr.debug(
             'Process %i exited with return code %i',
-            self.pid, return_code)
+            self.process.pid, return_code)
         # give captured process output back to the runner as string(s)
         results = {
-            name:
-            bytes(byt).decode(self.encoding)
-            if byt else ''
-            for name, byt in zip(self.FD_NAMES[1:], self.buffer)
+            self.fd_infos[fileno][0]: (
+                bytes(self.fd_infos[fileno][1]).decode(self.encoding)
+                if self.fd_infos[fileno][1] is not None
+                else '')
+            for fileno in (self.stdout_fileno, self.stderr_fileno)
         }
         results['code'] = return_code
         return results
 
     def process_exited(self):
-        # actually fulfill the future promise and let the execution finish
-        self.done.set_result(True)
+        pass
 
 
 class NoCapture(WitlessProtocol):
@@ -256,7 +192,7 @@ class KillOutput(WitlessProtocol):
             lgr.log(
                 5,
                 'Discarded %i bytes from %i[%s]',
-                len(data), self.pid, self.FD_NAMES[fd])
+                len(data), self.process.pid, self.fd_infos[fd][0])
 
 
 class WitlessRunner(object):
@@ -359,41 +295,17 @@ class WitlessRunner(object):
             cwd=cwd,
         )
 
-        # rescue any event-loop to be able to reassign after we are done
-        # with our own event loop management
-        # this is how ipython does it
-        try:
-            is_new_proc = self._check_if_new_proc()
-            event_loop = asyncio.get_event_loop()
-            if is_new_proc:
-                self._check_if_loop_usable(event_loop, stdin)
-            if event_loop.is_closed():
-                raise RuntimeError("the loop was closed - use our own")
-            new_loop = False
-        except RuntimeError:
-            event_loop = self._get_new_event_loop()
-            new_loop = True
-        try:
-            lgr.debug('Async run:\n cwd=%s\n cmd=%s', cwd, cmd)
-            # include the subprocess manager in the asyncio event loop
-            results = event_loop.run_until_complete(
-                run_async_cmd(
-                    event_loop,
-                    cmd,
-                    protocol,
-                    stdin,
-                    protocol_kwargs=kwargs,
-                    cwd=cwd,
-                    env=env,
-                )
-            )
-        finally:
-            if new_loop:
-                # be kind to callers and leave asyncio as we found it
-                asyncio.set_event_loop(None)
-                # terminate the event loop, cannot be undone, hence we start a fresh
-                # one each time (see BlockingIOError notes above)
-                event_loop.close()
+        from .nonasyncrunner import run_command
+
+        lgr.debug('Async run:\n cwd=%s\n cmd=%s', cwd, cmd)
+        results = run_command(
+            cmd,
+            protocol,
+            stdin,
+            protocol_kwargs=kwargs,
+            cwd=cwd,
+            env=env,
+        )
 
         # log before any exception is raised
         lgr.log(8, "Finished running %r with status %s", cmd, results['code'])
@@ -443,57 +355,6 @@ class WitlessRunner(object):
         elif cls._loop_need_new:
             raise RuntimeError("we know we need a new loop")
         return is_new_proc
-
-    @classmethod
-    def _check_if_loop_usable(cls, event_loop, stdin):
-        """Check if given event_loop could run a simple command
-
-        Sets _loop_need_new variable to a bool depending on what it finds
-
-        Note that this is a function that is meant to be called from within a
-        particular context only. The RuntimeError is expected to be caught by
-        the caller and is meant to be more like a response message than an
-        exception.
-
-        Raises
-        ------
-        RuntimeError
-          If loop is not reusable
-        """
-        # We need to check if we can run any command smoothly
-        try:
-            event_loop.run_until_complete(
-                run_async_cmd(
-                    event_loop,
-                    [sys.executable, "--version"],  # fast! 0.004 sec and to be ran once per process
-                    KillOutput,
-                    stdin,
-                )
-            )
-            cls._loop_need_new = False
-        except OSError as e:
-            # due to https://bugs.python.org/issue21998
-            # exhibits in https://github.com/ReproNim/testkraken/issues/95
-            lgr.debug("It seems we need a new loop when running our commands: %s", exc_str(e))
-            cls._loop_need_new = True
-            raise RuntimeError("the loop is not reusable")
-
-    @staticmethod
-    def _get_new_event_loop():
-        # start a new event loop, which we will close again further down
-        # if this is not done events like this will occur
-        #   BlockingIOError: [Errno 11] Resource temporarily unavailable
-        #   Exception ignored when trying to write to the signal wakeup fd:
-        # It is unclear to me why it happens when reusing an event looped
-        # that it stopped from time to time, but starting fresh and doing
-        # a full termination seems to address the issue
-        if sys.platform == "win32":
-            # use special event loop that supports subprocesses on windows
-            event_loop = asyncio.ProactorEventLoop()
-        else:
-            event_loop = asyncio.SelectorEventLoop()
-        asyncio.set_event_loop(event_loop)
-        return event_loop
 
 
 class GitRunnerBase(object):
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -11,17 +11,24 @@ Wrapper for command and function calls, allowing for dry runs and output handlin
 
 """
 
-import subprocess
-import sys
 import logging
 import os
+import subprocess
+import sys
 import tempfile
-from locale import getpreferredencoding
-import asyncio
 import warnings
 
+from .cmd_protocols import (
+    KillOutput,
+    NoCapture,
+    StdErrCapture,
+    StdOutCapture,
+    StdOutErrCapture,
+    WitlessProtocol,
+)
 from .consts import GIT_SSH_COMMAND
 from .dochelpers import borrowdoc
+from .nonasyncrunner import run_command
 from .support import path as op
 from .support.exceptions import CommandError
 from .utils import (
@@ -42,161 +49,6 @@ _TEMP_std = sys.stdout, sys.stderr
 # in Runner so we take care about their removal, in contrast to those
 # which might be created outside and passed into Runner
 _MAGICAL_OUTPUT_MARKER = "_runneroutput_"
-
-
-class WitlessProtocol(asyncio.SubprocessProtocol):
-    """Subprocess communication protocol base class for `run_async_cmd`
-
-    This class implements basic subprocess output handling. Derived classes
-    like `StdOutCapture` should be used for subprocess communication that need
-    to capture and return output. In particular, the `pipe_data_received()`
-    method can be overwritten to implement "online" processing of process
-    output.
-
-    This class defines a default return value setup that causes
-    `run_async_cmd()` to return a 2-tuple with the subprocess's exit code
-    and a list with bytestrings of all captured output streams.
-    """
-
-    proc_out = None
-    proc_err = None
-
-    def __init__(self, done_future=None, encoding=None):
-        """
-        Parameters
-        ----------
-        done_future: Any
-          Ignored parameter, kept for backward compatibility (DEPRECATED)
-        encoding : str
-          Encoding to be used for process output bytes decoding. By default,
-          the preferred system encoding is guessed.
-        """
-
-        if done_future is not None:
-            warnings.warn("`done_future` argument is ignored "
-                          "and will be removed in a future release",
-                          DeprecationWarning)
-
-        self.fd_infos = {}
-
-        self.process = None
-        self.stdout_fileno = 1
-        self.stderr_fileno = 2
-
-        # capture output in bytearrays while the process is running
-        self.fd_infos[self.stdout_fileno] = ("stdout", bytearray()) if self.proc_out else ("stdout", None)
-        self.fd_infos[self.stderr_fileno] = ("stderr", bytearray()) if self.proc_err else ("stderr", None)
-
-        super().__init__()
-        self.encoding = encoding or getpreferredencoding(do_setlocale=False)
-
-        self._log_outputs = False
-        if lgr.isEnabledFor(5):
-            try:
-                from . import cfg
-                self._log_outputs = cfg.getbool('datalad.log', 'outputs', default=False)
-            except ImportError:
-                pass
-            self._log = self._log_summary
-        else:
-            self._log = self._log_nolog
-
-    def _log_nolog(self, *args):
-        pass
-
-    def _log_summary(self, fd, data):
-        fd_name = self.fd_infos[fd]
-        lgr.log(5, 'Read %i bytes from %i[%s]%s',
-                len(data), self.process.pid, fd_name, ':' if self._log_outputs else '')
-        if self._log_outputs:
-            log_data = ensure_unicode(data)
-            # The way we log is to stay consistent with Runner.
-            # TODO: later we might just log in a single entry, without
-            # fd_name prefix
-            lgr.log(5, "%s| %s ", fd_name, log_data)
-
-    def connection_made(self, process):
-        self.process = process
-        lgr.debug('Process %i started', self.process.pid)
-
-    def pipe_data_received(self, fd, data):
-        self._log(fd, data)
-        # store received output if stream was to be captured
-        fd_name, buffer = self.fd_infos[fd]
-        if buffer is not None:
-            buffer.extend(data)
-
-    def _prepare_result(self):
-        """Prepares the final result to be returned to the runner
-
-        Note for derived classes overwriting this method:
-
-        The result must be a dict with keys that do not unintentionally
-        conflict with the API of CommandError, as the result dict is passed to
-        this exception class as kwargs on error. The Runner will overwrite
-        'cmd' and 'cwd' on error, if they are present in the result.
-        """
-        return_code = self.process.poll()
-        if return_code is None:
-            raise CommandError(
-                "Got None as a return_code for the process %i",
-                self.process.pid)
-        lgr.debug(
-            'Process %i exited with return code %i',
-            self.process.pid, return_code)
-        # give captured process output back to the runner as string(s)
-        results = {
-            name: (
-                bytes(byt).decode(self.encoding)
-                if byt is not None
-                else '')
-            for name, byt in self.fd_infos.values()
-        }
-        results['code'] = return_code
-        return results
-
-    def process_exited(self):
-        pass
-
-
-class NoCapture(WitlessProtocol):
-    """WitlessProtocol that captures no subprocess output
-
-    As this is identical with the behavior of the WitlessProtocol base class,
-    this class is merely a more readable convenience alias.
-    """
-    pass
-
-
-class StdOutCapture(WitlessProtocol):
-    """WitlessProtocol that only captures and returns stdout of a subprocess"""
-    proc_out = True
-
-
-class StdErrCapture(WitlessProtocol):
-    """WitlessProtocol that only captures and returns stderr of a subprocess"""
-    proc_err = True
-
-
-class StdOutErrCapture(WitlessProtocol):
-    """WitlessProtocol that captures and returns stdout/stderr of a subprocess
-    """
-    proc_out = True
-    proc_err = True
-
-
-class KillOutput(WitlessProtocol):
-    """WitlessProtocol that swallows stdout/stderr of a subprocess
-    """
-    proc_out = True
-    proc_err = True
-
-    def pipe_data_received(self, fd, data):
-        if lgr.isEnabledFor(5):
-            lgr.log(
-                5,
-                'Discarded %i bytes from %i[%s]',
-                len(data), self.process.pid, self.fd_infos[fd][0])
 
 
 class WitlessRunner(object):
@@ -289,10 +141,6 @@ class WitlessRunner(object):
         FileNotFoundError
           When a given executable does not exist.
         """
-
-        # imported here to avoid recursive importing
-        from .nonasyncrunner import run_command
-
         if protocol is None:
             # by default let all subprocess stream pass through
             protocol = NoCapture
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -148,7 +148,7 @@ class WitlessProtocol(asyncio.SubprocessProtocol):
                 bytes(byt).decode(self.encoding)
                 if byt is not None
                 else '')
-            for name, byt in self.fd_infos.items()
+            for name, byt in self.fd_infos.values()
         }
         results['code'] = return_code
         return results
@@ -299,7 +299,7 @@ class WitlessRunner(object):
 
         from .nonasyncrunner import run_command
 
-        lgr.debug('Async run:\n cwd=%s\n cmd=%s', cwd, cmd)
+        lgr.debug('run:\n cwd=%s\n cmd=%s', cwd, cmd)
         results = run_command(
             cmd,
             protocol,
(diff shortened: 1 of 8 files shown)


mih (commented):
Had a first look. Cool!

From my POV this needs some real-world exposure. I will try to use this branch in some of the more tricky tasks, and report back.

THX big time!

datalad/support/annexrepo.py
```diff
@@ -3408,10 +3408,10 @@ class AnnexJsonProtocol(WitlessProtocol):
     proc_out = True
     proc_err = True
 
-    def __init__(self, done_future, total_nbytes=None):
+    def __init__(self, total_nbytes=None):
```
yarikoptic, line 3411: I wonder if we should avoid breaking API for a cycle here and still allow for `done_future=None` and issue DeprecationWarning if specified?

datalad/nonasyncrunner.py
```diff
@@ -0,0 +1,138 @@
...
+        self.quit = True
+        self.file.close()
+
+    def run(self):
+        logger.debug(f"ReaderThread({self.file}, {self.queue}) started")
+        while not self.quit:
+            try:
+                data = os.read(self.file.fileno(), 1024)
+            except BrokenPipeError as exc:
+                logger.debug(f"{self} exiting (broken pipe)")
```
yarikoptic, line 56: although just a minor opt, in general - avoid unconditional formatting here and elsewhere of log messages -- delay until/if line is actually logged:

```suggestion
                logger.debug("%s exiting (broken pipe)", self)
```

yarikoptic:
Woooohooo!  I see the light!  I am going to it!!!  thank you @christian-monch !

<details>
<summary>only 2 failing tests out of >1k in one of the travis runs</summary> 

```shell
FAIL: datalad.dataset.tests.test_gitrepo.test_gitrepo_call_git_methods
  File "/tmp/dl-miniconda-q7xiqj3h/lib/python3.8/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/tmp/dl-miniconda-q7xiqj3h/lib/python3.8/site-packages/datalad/tests/utils.py", line 585, in _wrap_with_tree
    return t(*(arg + (d,)), **kw)
  File "/tmp/dl-miniconda-q7xiqj3h/lib/python3.8/site-packages/datalad/dataset/tests/test_gitrepo.py", line 242, in test_gitrepo_call_git_methods
    check("fatal: bad source", cml.out)
AssertionError: 'fatal: bad source' unexpectedly found in "[DEBUG] Async run:\n cwd=/tmp/datalad_temp_tree_test_gitrepo_call_git_methodsfoymy59e\n cmd=['git', '-c', 'diff.ignoreSubmodules=none', 'mv', '--', 'notthere', 'dest']\n[DEBUG] Process 25881 started\n[DEBUG] ReaderThread(<_io.FileIO name=14 mode='rb' closefd=True>, <queue.Queue object at 0x7f4f52ff5bb0>) started\n[DEBUG] ReaderThread(<_io.FileIO name=10 mode='rb' closefd=True>, <queue.Queue object at 0x7f4f52ff5bb0>) started\n[DEBUG] ReaderThread(<_io.FileIO name=10 mode='rb' closefd=True>, <queue.Queue object at 0x7f4f52ff5bb0>) exiting (stream end)\n[DEBUG] ReaderThread(<_io.FileIO name=14 mode='rb' closefd=True>, <queue.Queue object at 0x7f4f52ff5bb0>) got data: b'fatal: bad source, source=notthere, destination=dest\\n'\n[DEBUG] ReaderThread(<_io.FileIO name=14 mode='rb' closefd=True>, <queue.Queue object at 0x7f4f52ff5bb0>) exiting (stream end)\n[DEBUG] Process 25881 exited with return code 128\n"
datalad.cmd: DEBUG: Async run:
 cwd=/tmp/datalad_temp_tree_test_gitrepo_call_git_methodsfoymy59e
 cmd=['git', '-c', 'diff.ignoreSubmodules=none', 'mv', '--', 'notthere', 'dest']
datalad.cmd: DEBUG: Process 25878 started
datalad.runner: DEBUG: ReaderThread(<_io.FileIO name=25 mode='rb' closefd=True>, <queue.Queue object at 0x7f4f52ef9550>) started
datalad.runner: DEBUG: ReaderThread(<_io.FileIO name=23 mode='rb' closefd=True>, <queue.Queue object at 0x7f4f52ef9550>) started
datalad.runner: DEBUG: ReaderThread(<_io.FileIO name=23 mode='rb' closefd=True>, <queue.Queue object at 0x7f4f52ef9550>) exiting (stream end)
datalad.runner: DEBUG: ReaderThread(<_io.FileIO name=25 mode='rb' closefd=True>, <queue.Queue object at 0x7f4f52ef9550>) got data: b'fatal: bad source, source=notthere, destination=dest\n'
datalad.runner: DEBUG: ReaderThread(<_io.FileIO name=25 mode='rb' closefd=True>, <queue.Queue object at 0x7f4f52ef9550>) exiting (stream end)
datalad.cmd: DEBUG: Process 25878 exited with return code 128
datalad.dataset.gitrepo: Level 11: CommandError: 'git -c diff.ignoreSubmodules=none mv -- notthere dest' failed with exitcode 128 under /tmp/datalad_temp_tree_test_gitrepo_call_git_methodsfoymy59e [err: 'fatal: bad source, source=notthere, destination=dest']
datalad.cmd: DEBUG: Async run:
 cwd=/tmp/datalad_temp_tree_test_gitrepo_call_git_methodsfoymy59e
 cmd=['git', '-c', 'diff.ignoreSubmodules=none', 'mv', '--', 'notthere', 'dest']
datalad.cmd: DEBUG: Process 25881 started
datalad.runner: DEBUG: ReaderThread(<_io.FileIO name=14 mode='rb' closefd=True>, <queue.Queue object at 0x7f4f52ff5bb0>) started
datalad.runner: DEBUG: ReaderThread(<_io.FileIO name=10 mode='rb' closefd=True>, <queue.Queue object at 0x7f4f52ff5bb0>) started
datalad.runner: DEBUG: ReaderThread(<_io.FileIO name=10 mode='rb' closefd=True>, <queue.Queue object at 0x7f4f52ff5bb0>) exiting (stream end)
datalad.runner: DEBUG: ReaderThread(<_io.FileIO name=14 mode='rb' closefd=True>, <queue.Queue object at 0x7f4f52ff5bb0>) got data: b'fatal: bad source, source=notthere, destination=dest\n'
datalad.runner: DEBUG: ReaderThread(<_io.FileIO name=14 mode='rb' closefd=True>, <queue.Queue object at 0x7f4f52ff5bb0>) exiting (stream end)
datalad.cmd: DEBUG: Process 25881 exited with return code 128
--------------------- >> end captured logging << ---------------------
FAIL: datalad.support.tests.test_gitrepo.test_gitrepo_call_git_methods
  File "/tmp/dl-miniconda-q7xiqj3h/lib/python3.8/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/tmp/dl-miniconda-q7xiqj3h/lib/python3.8/site-packages/datalad/tests/utils.py", line 585, in _wrap_with_tree
    return t(*(arg + (d,)), **kw)
  File "/tmp/dl-miniconda-q7xiqj3h/lib/python3.8/site-packages/datalad/support/tests/test_gitrepo.py", line 1632, in test_gitrepo_call_git_methods
    check("fatal: bad source", cml.out)
AssertionError: 'fatal: bad source' unexpectedly found in "[DEBUG] Async run:\n cwd=/tmp/datalad_temp_tree_test_gitrepo_call_git_methodsxtvk6i7c\n cmd=['git', '-c', 'diff.ignoreSubmodules=none', 'mv', '--', 'notthere', 'dest']\n[DEBUG] Process 4319 started\n[DEBUG] ReaderThread(<_io.FileIO name=14 mode='rb' closefd=True>, <queue.Queue object at 0x7f4f42da0430>) started\n[DEBUG] ReaderThread(<_io.FileIO name=10 mode='rb' closefd=True>, <queue.Queue object at 0x7f4f42da0430>) started\n[DEBUG] ReaderThread(<_io.FileIO name=14 mode='rb' closefd=True>, <queue.Queue object at 0x7f4f42da0430>) got data: b'fatal: bad source, source=notthere, destination=dest\\n'\n[DEBUG] ReaderThread(<_io.FileIO name=10 mode='rb' closefd=True>, <queue.Queue object at 0x7f4f42da0430>) exiting (stream end)\n[DEBUG] ReaderThread(<_io.FileIO name=14 mode='rb' closefd=True>, <queue.Queue object at 0x7f4f42da0430>) exiting (stream end)\n[DEBUG] Process 4319 exited with return code 128\n"
datalad.cmd: DEBUG: Async run:
 cwd=/tmp/datalad_temp_tree_test_gitrepo_call_git_methodsxtvk6i7c
 cmd=['git', '-c', 'diff.ignoreSubmodules=none', 'mv', '--', 'notthere', 'dest']
datalad.cmd: DEBUG: Process 4316 started
datalad.runner: DEBUG: ReaderThread(<_io.FileIO name=125 mode='rb' closefd=True>, <queue.Queue object at 0x7f4f4291e280>) started
datalad.runner: DEBUG: ReaderThread(<_io.FileIO name=123 mode='rb' closefd=True>, <queue.Queue object at 0x7f4f4291e280>) started
datalad.runner: DEBUG: ReaderThread(<_io.FileIO name=125 mode='rb' closefd=True>, <queue.Queue object at 0x7f4f4291e280>) got data: b'fatal: bad source, source=notthere, destination=dest\n'
datalad.runner: DEBUG: ReaderThread(<_io.FileIO name=125 mode='rb' closefd=True>, <queue.Queue object at 0x7f4f4291e280>) exiting (stream end)
datalad.runner: DEBUG: ReaderThread(<_io.FileIO name=123 mode='rb' closefd=True>, <queue.Queue object at 0x7f4f4291e280>) exiting (stream end)
datalad.cmd: DEBUG: Process 4316 exited with return code 128
datalad.dataset.gitrepo: Level 11: CommandError: 'git -c diff.ignoreSubmodules=none mv -- notthere dest' failed with exitcode 128 under /tmp/datalad_temp_tree_test_gitrepo_call_git_methodsxtvk6i7c [err: 'fatal: bad source, source=notthere, destination=dest']
datalad.cmd: DEBUG: Async run:
 cwd=/tmp/datalad_temp_tree_test_gitrepo_call_git_methodsxtvk6i7c
 cmd=['git', '-c', 'diff.ignoreSubmodules=none', 'mv', '--', 'notthere', 'dest']
datalad.cmd: DEBUG: Process 4319 started
datalad.runner: DEBUG: ReaderThread(<_io.FileIO name=14 mode='rb' closefd=True>, <queue.Queue object at 0x7f4f42da0430>) started
datalad.runner: DEBUG: ReaderThread(<_io.FileIO name=10 mode='rb' closefd=True>, <queue.Queue object at 0x7f4f42da0430>) started
datalad.runner: DEBUG: ReaderThread(<_io.FileIO name=14 mode='rb' closefd=True>, <queue.Queue object at 0x7f4f42da0430>) got data: b'fatal: bad source, source=notthere, destination=dest\n'
datalad.runner: DEBUG: ReaderThread(<_io.FileIO name=10 mode='rb' closefd=True>, <queue.Queue object at 0x7f4f42da0430>) exiting (stream end)
datalad.runner: DEBUG: ReaderThread(<_io.FileIO name=14 mode='rb' closefd=True>, <queue.Queue object at 0x7f4f42da0430>) exiting (stream end)
datalad.cmd: DEBUG: Process 4319 exited with return code 128
```
</details>

Left a few tiny individual comments.  Would be nice if tried within jupyter and also absorbed the test `test_popen_invocation` demonstrating the issue https://github.com/datalad/datalad/pull/5369/files#diff-4aa1a9688db72f7a475890e0ee6a343ea98af70ccc3bfa66f17da82c819dfa8dR230

datalad/support/annexrepo.py
christian-monch: Good point. I am not fully aware, where it is used and assumed we use it only internally. Do we expose it externally, i.e. do we document it and encourage its use? To me either way is fine.

christian-monch:
> only 2 failing tests out of >1k in one of the travis runs

Those should be fixed with commit d64e61d334ca59e7a4a485c174dc3b4a84eb2443 (the reader threads debug-logged the data coming from the sub-process, which confuses some tests that capture all output).

yarikoptic:
FWIW, tried on windows VM conda install jupyter lab -- although I have not exhibited all the issues others saw, for me it just worked well with this PR. COOL!

datalad/support/annexrepo.py
yarikoptic: well -- wonders of Python that as long as it is not `_`'ed somewhere in the namespace path to it -- it is "public".  But it is so specific and I do not see hits for https://github.com/search?q=AnnexJsonProtocol&type=code at least outside of datalad, and tests for extensions are green -- I guess we could just jump and not care about deprecation cycle.

datalad/nonasyncrunner.py
```diff
@@ -0,0 +1,137 @@
...
+        have any effect, because the thread might be waiting in
+        os.read() or queue.put(). We are closing the file that read
+        is reading from here, but the queue has to be emptied in
+        another thread in order to ensure thread-exiting.
+        """
+        self.quit = True
+        self.file.close()
+
+    def run(self):
+        logger.debug(f"ReaderThread({self.file}, {self.queue}) started")
```
yarikoptic, line 51: same here 
```suggestion
        logger.debug("ReaderThread(%s, %s) started", self.file, self.queue)
```

datalad/nonasyncrunner.py
```diff
@@ -0,0 +1,137 @@
...
+        while not self.quit:
+            try:
+                data = os.read(self.file.fileno(), 1024)
+            except BrokenPipeError as exc:
+                logger.debug("%s exiting (broken pipe)", self)
+                self.queue.put((self.file.fileno(), exc, time.time()))
+                return
+
+            if data == b"":
+                logger.debug(f"{self} exiting (stream end)")
```
yarikoptic, line 61:
```suggestion
                logger.debug("%s exiting (stream end)", self)
```

datalad/nonasyncrunner.py
```diff
@@ -0,0 +1,137 @@
...
+                return
+
+            self.queue.put((self.file.fileno(), data, time.time()))
+
+
+def run_command(cmd,
+                protocol_class,
+                stdin,
+                protocol_kwargs=None,
+                **kwargs) -> Any:
```
yarikoptic, line 72: docstring describing options and what `kwargs` about would be nice

datalad/nonasyncrunner.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def run(self):
+        logger.debug(f"ReaderThread({self.file}, {self.queue}) started")
+        while not self.quit:
+            try:
+                data = os.read(self.file.fileno(), 1024)
+            except BrokenPipeError as exc:
+                logger.debug("%s exiting (broken pipe)", self)
+                self.queue.put((self.file.fileno(), exc, time.time()))
+                return
```
yarikoptic, line 58: since this is such a core functionality for the datalad, would be nice to have exception handling tested as well.  E.g. in this case could be a process closing its stdin pipe prematurely or smth like that?

yarikoptic (changes requested):
I know it is just a draft PR, but I am too excited ;-)  Left a few minor comments.

yarikoptic:
I have adjusted description with issues this PR would close and TODOs in my view to accomplish (easier to keep track of them when present in the top description) - they could always be addressed or argued against and crossed out ;)

Re `[dataset-traverser->metadata-extractor->metadata-adder]` motivation -- I think it aligns very well with `ProducerConsumer` pattern we already have (https://github.com/datalad/datalad/blob/master/datalad/support/parallel.py#L71) and if bottleneck is IO and not CPU - it could possibly be reused as is, or RFed to
- support sub-process not Thread mode of operation
- enhanced/subclassed? to have some "Reducer" for the final singular `metadata-adder`, although I think it could just sit on top of it and consume all produced records from `ProducerConsumer` to do its "reduce/addition/whatnot" operation in the same process (unless indeed desired to have a separate thread or subprocess)

datalad/nonasyncrunner.py
```diff
@@ -0,0 +1,137 @@
...
+    # We pass process as transport-argument. It does not have the same
+    # semantics as the asyncio-signature, but since it is only used in
+    # WitlessProtocol, we can change it there.
+    protocol.connection_made(process)
+
+    # Map the pipe file numbers to stdout and stderr file number, because
+    # the latter are hardcoded in the protocol code
+    # TODO: the fixed file numbers seem to be a side-effect of using
+    #  SubprocessProtocol. Some datalad-code relies on this. Shall
+    #  we replace hard coded stdout, stderr-file numbers with parameters?
```
kyleam, line 102: Given that the protocols derive from `asyncio.SubprocessProtocol`, I think sticking to the standard 1/2 make sense, so I'd vote for handling the translation in `run_command` as you currently do.

kyleam:
This is great to see.  Thanks, @christian-monch.

Obviously this addressing outstanding problems is more important, but it's worth noting that the benchmarks are looking good too:

<details>
<summary>benchmarks</summary>

```
2021-05-17T10:23:45.3349457Z -      3.41±0.1ms      2.74±0.03ms     0.80  core.witlessrunner.time_echo
2021-05-17T10:23:45.3350512Z        4.03±0.2ms       3.33±0.3ms    ~0.83  core.witlessrunner.time_echo_gitrunner
2021-05-17T10:23:45.3351661Z -     3.95±0.05ms      3.57±0.06ms     0.90  core.witlessrunner.time_echo_gitrunner_fullcapture
```

```
2021-05-17T13:21:00.1262025Z        2.97±0.2ms       2.73±0.2ms     0.92  core.witlessrunner.time_echo
2021-05-17T13:21:00.1263348Z        3.60±0.5ms       3.06±0.4ms    ~0.85  core.witlessrunner.time_echo_gitrunner
2021-05-17T13:21:00.1264621Z        3.81±0.2ms      3.38±0.06ms    ~0.89  core.witlessrunner.time_echo_gitrunner_fullcapture
```

```
2021-05-17T13:28:29.8606181Z -      2.40±0.1ms      2.00±0.02ms     0.83  core.witlessrunner.time_echo
2021-05-17T13:28:29.8607182Z -      2.78±0.2ms      2.50±0.06ms     0.90  core.witlessrunner.time_echo_gitrunner
2021-05-17T13:28:29.8609269Z        2.92±0.2ms      2.64±0.06ms    ~0.90  core.witlessrunner.time_echo_gitrunner_fullcapture
```

https://github.com/datalad/datalad/runs/2599712075
https://github.com/datalad/datalad/runs/2600966706
https://github.com/datalad/datalad/runs/2601062427

</details>

datalad/nonasyncrunner.py
christian-monch: Ok, Thanks :-)

datalad/nonasyncrunner.py
```diff
@@ -0,0 +1,137 @@
...
+    catch_stderr = protocol_class.proc_err is not None
+
+    kwargs = {
+        **kwargs,
+        **dict(
+            bufsize=0,
+            stdin=stdin,
+            stdout=subprocess.PIPE if catch_stdout else None,
+            stderr=subprocess.PIPE if catch_stderr else None,
+            shell=True if isinstance(cmd, str) else False
```
bpoldrack, line 84: `shell=True`: Not sure if we ever really want that. It's considered a security hazard, and we should always be able to properly `shlex.split()` into a list.

I think we did this somewhere before, so it's not really this PR introducing it, but I'd like to use the occasion to get rid of it.

datalad/nonasyncrunner.py
bpoldrack: Hm. In my mind - the protocols don't *have* to derive from `asyncio.SubprocessProtocol` anymore with this PR.
What's important is to keep the concept, not the heritage. However - fine with me to not do that with the the first shot at it.

bpoldrack (commented):
Cool, that's what I wanted to see.
What I'd love to see in addition, is a `WriterThread` as well. ;-)

datalad/nonasyncrunner.py
mih: Please don't use `shlex.split()`, it is POSIX-only and a significant chunk of my life went into getting its use out of the command execution code. Whether or not the use of `shell=True` is a security risk should be left to the caller to decide.

mih: and FTR: If any splitting is to be performed, `datalad.utils.split_cmdline()` is to be used for cross-platform compatibility.

datalad/nonasyncrunner.py
christian-monch: Thanks for the comment, We actually should not check for BrokenPipeErrors, those should only be raised on writing.

datalad/nonasyncrunner.py
christian-monch: > Cool, that's what I wanted to see.
> What I'd love to see in addition, is a `WriterThread` as well. ;-)

That was in a previous version, together with another queue (and it worked nicely to remote control an interactive python session). I removed it for now, because it seems that the existing caller of the runner only provide stdin-content when calling the runner. That seems to work for now.

If we want to change that (the queue and writer-thread can easily be re-added), we have IMHO three options:

1. [Not good] Change the runner interface to expose a pipe, queue, or similar which the caller could use to send data (That could be a major change in the interface). This would allow the main thread to send data whenever required. But the caller interface would be something like: `start process; while process is not done: write_to_process`.

2. [Ok, works with processes the generate output] The caller implements a protocol in which the receive callbacks deposit data in the writer-queue. This requires some kind of "interactive" behavior of the process, i.e. read from stdin, write something to stdout/stderr.

3. [Best, more complicated] Extend WitlessProtocol with a input callback, that is executed on the main thread and allows to provide input. That would be called whenever the input queue is not-full, therefore some flow-control primitives might also be necessary. Then the caller can decide whether to provide input to the subprocess in the input-callback or in other callbacks by putting elements into the input queue.

I think it is nice to have, but I am wondering whether there is a need for "interactive" input to the subprocess.

christian-monch: > `shell=True`: Not sure if we ever really want that. It's considered a security hazard, and we should always be able to properly `shlex.split()` into a list.
> 
> I think we did this somewhere before, so it's not really this PR introducing it, but I'd like to use the occasion to get rid of it.

I kept the original semantics in this one

bpoldrack: > Please don't use shlex.split(), it is POSIX-only and a significant chunk of my life went into getting its use out of the command execution code.
> If any splitting is to be performed, datalad.utils.split_cmdline() is to be used for cross-platform compatibility.

True. Forgot about that.

> Whether or not the use of shell=True is a security risk should be left to the caller to decide.

Fine with me, but here we decide based on what `cmd` is an instance of. That's the "wrong" part, I think.

bpoldrack: > Therefore I determined, that an "interactive" stdin-content provision is not required.

The part where it is required at the moment, is the implementation of the ORA special remote. I don't go through the runner there and have an ugly construct to talk to that subprocess. Always wanted to build a proper threaded solution like this one, but never got to it.

Re 1., 2. or 3.:

The usecase with ORA is indeed an "interactive" remote shell. The special remote waits for git-annex to request something, then it goes ahead, writes a command to that remote shell's stdin, reads from stdout/stderr and proceeds according to the output. Then the shell's subprocess is "idle" until the process that runs the special remote receives the next request from annex and so on. So, an input callback sounds suboptimal at a first glance (need to think that through). The caller process needs to be able to decide, when to send the next input and must not be blocked by the subprocess waiting for an input.

bpoldrack: Just to emphasize: Not sure about an input callback not being a good solution. Need to have a calm moment to envision how that would look like. Not having that moment right now :)

mih: > Fine with me, but here we decide based on what cmd is an instance of. That's the "wrong" part, I think.

Wrong or not, this is how it is done, and that assumption is baked into the present usage. If it needs to be changed now to become "right", please book the time to assess the situation in-full, rather than in the narrow scope of what would be good for this method.

christian-monch: > The usecase with ORA is indeed an "interactive" remote shell. The special remote waits for git-annex to request something, then it goes ahead, writes a command to that remote shell's stdin, reads from stdout/stderr and proceeds according to the output. Then the shell's subprocess is "idle" until the process that runs the special remote receives the next request from annex and so on. So, an input callback sounds suboptimal at a first glance (need to think that through). The caller process needs to be able to decide, when to send the next input and must not be blocked by the subprocess waiting for an input.

The current implementation passes the Popen instance of the subprocess in the connection_made-callback. That should have a pipe, if you override the stdin-keyword argument with { "stdin": subprocess.PIPE } in the kwargs to run_command.

The stdin-attribute of the Popen instance can always be used to send bytes to the stdin of the subprocess. But: only when the main thread runs in your code, which is currently (while the subprocess is running) only in the Protocol callbacks, mainly in pipe_data_received.

If that is not sufficient, we can add a provide_pipe_data-callback that is called whenever we can send data to the subprocess. This can be throttled by the size of an input queue (which holds elements that the subprocess should consume from stdin), otherwise, it would be called very often

datalad/nonasyncrunner.py
christian-monch: done

datalad/nonasyncrunner.py
kyleam: > the protocols don't _have_ to derive from `asyncio.SubprocessProtocol` anymore with this PR.

I don't disagree with that, though I also don't see a benefit of decoupling.  And, on second thought, regardless of the parent class/interface, I consider mapping to the standard 1,2 fds inside `run_command` a bit nicer from the standpoint of isolating the logic/implementation details.

datalad/nonasyncrunner.py
christian-monch: In the first version, no specific writer support will be implemented. It is possible to pass `subprocess.PIPE` as stdin-argument to run_command and write to this pipe, for example, in the `connection_made` or `pipe_data_received` callbacks of the protocol

datalad/nonasyncrunner.py
christian-monch: We keep it as it is for now

christian-monch:
@yarikoptic: I checked the ability of pycrunch-engine to execute test_ria_basic with this PR in a python 3.6 environment. Seems fine. If I understand task 3 correctly, you would like a test that runs uses an async context and executed the runner successfully within this context?

datalad/nonasyncrunner.py
christian-monch: We go with the proposed implementation for now.

datalad/support/annexrepo.py
christian-monch: Deprecation cycle added

datalad/nonasyncrunner.py
```diff
@@ -0,0 +1,162 @@
...
+
+"""
+
+import logging
+import os
+import queue
+import subprocess
+import threading
+import time
+from typing import Any
```
yarikoptic, line 20: xoxoxo -- I thought to complain with "as much as I would have liked to add typing, let's postpone for this PR: it would impact CLI startup time notably" but apparently we were impacted already for a while due to `asyncio` itself (at least with python3 3.9.1):

```
  File "/home/yoh/proj/datalad/datalad-master/datalad/__init__.py", line 47, in <module>
    from .config import ConfigManager
  File "/home/yoh/proj/datalad/datalad-master/datalad/config.py", line 22, in <module>
    from datalad.cmd import (
  File "/home/yoh/proj/datalad/datalad-master/datalad/cmd.py", line 21, in <module>
    import asyncio
  File "/usr/lib/python3.9/asyncio/__init__.py", line 8, in <module>
    from .base_events import *
  File "/usr/lib/python3.9/asyncio/base_events.py", line 45, in <module>
    from . import staggered
  File "/usr/lib/python3.9/asyncio/staggered.py", line 6, in <module>
    import typing
  File "/tmp/badimport/typing.py", line 1, in <module>
    raise ImportError
ImportError
```

So finally we can resurrect https://github.com/datalad/datalad/pull/4513 and see other issue referenced therein, and introduce type hints through out the code!!! 

FWIW -- I verified that `import datalad` and `datalad --help-np` aren't notably impacted by this PR for me since we already tolerated import of typing via asyncio , so I would just welcome its addition here and elsewhere ;-)

datalad/nonasyncrunner.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+logger = logging.getLogger("datalad.runner")
+
+STDOUT_FILENO = 1
+STDERR_FILENO = 2
+
+
+class ReaderThread(threading.Thread):
+    def __init__(self, file, a_queue, command):
```
yarikoptic, line 30: docstring (or even type hints?;-)) please. what `a_queue` and `file` should be ?

datalad/nonasyncrunner.py
```diff
@@ -0,0 +1,162 @@
...
+    def __str__(self):
+        return f"ReaderThread({self.file}, {self.queue}, {self.command})"
+
+    def request_exit(self):
+        """
+        Request the thread to exit. This is not guaranteed to
+        have any effect, because the thread might be waiting in
+        os.read() or queue.put().
+        """
+        self.quit = True
```
yarikoptic, line 46: could we test this scenario/code block?

datalad/nonasyncrunner.py
```diff
@@ -0,0 +1,162 @@
...
+            if data == b"":
+                logger.debug("%s exiting (stream end)", self)
+                self.queue.put((self.file.fileno(), None, time.time()))
+                return
+
+            self.queue.put((self.file.fileno(), data, time.time()))
+
+
+def run_command(cmd,
+                protocol_class,
```
yarikoptic, line 63:
```suggestion
                protocol,
```
to match the doc or adjust the doc?

datalad/tests/test_nonasyncrunner.py
```diff
@@ -0,0 +1,159 @@
+# emacs: -*- mode: python-mode; py-indent-offset: 4; tab-width: 4; indent-tabs-mode: nil; coding: utf-8 -*-
+# ex: set sts=4 ts=4 sw=4 noet:
+# ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+#
+#   See COPYING file distributed along with the datalad package for the
+#   copyright and license terms.
+#
+# ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+"""Test the thread based runner (aka. non asyncio based runner).
```
yarikoptic, line 9: hm... codecov is bugging out? how this file could be not test covered???

yarikoptic (changes requested):
comments, notes and recommendations.  And with this PR let's consciously welcome typing hints! ;-)

datalad/cmd.py
```diff
@@ -359,41 +296,17 @@ def run(self, cmd, protocol=None, stdin=None, cwd=None, env=None, **kwargs):
...
-        finally:
-            if new_loop:
-                # be kind to callers and leave asyncio as we found it
-                asyncio.set_event_loop(None)
-                # terminate the event loop, cannot be undone, hence we start a fresh
-                # one each time (see BlockingIOError notes above)
-                event_loop.close()
+        from .nonasyncrunner import run_command
+
+        lgr.debug('Async run:\n cwd=%s\n cmd=%s', cwd, cmd)
```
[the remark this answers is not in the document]
christian-monch: It ain't no longer. :-) That is true. And I did remove it earlier, but at least one test does rely on the string "Async run:" being present in the output :-/ That has to be changed as well (`datalad.core.distributed.tests.test_clone.py:test_cfg_originorigin`)

Shall I do it?

datalad/cmd.py
```diff
@@ -359,41 +296,17 @@ def run(self, cmd, protocol=None, stdin=None, cwd=None, env=None, **kwargs):
...
-                )
-            )
-        finally:
-            if new_loop:
-                # be kind to callers and leave asyncio as we found it
-                asyncio.set_event_loop(None)
-                # terminate the event loop, cannot be undone, hence we start a fresh
-                # one each time (see BlockingIOError notes above)
-                event_loop.close()
+        from .nonasyncrunner import run_command
```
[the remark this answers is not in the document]
christian-monch: will move

datalad/nonasyncrunner.py
christian-monch: Gladly type hints. I am all for those (tried to be a little light-footed due to previously uttered reservations ;-)

datalad/nonasyncrunner.py
christian-monch: The non blocking part is tested and we can wrap the blocking part into that (or do it seperately)

datalad/nonasyncrunner.py
christian-monch: Good one. Let's adjust the doc

datalad/tests/test_nonasyncrunner.py
christian-monch: I don't know. I get 100% coverage locally when I execute the tests (on linux, python 3.6). Are we sure that this issue will still exist after merging?

datalad/cmd.py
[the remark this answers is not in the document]
yarikoptic: sure -- tests are under our control

datalad/nonasyncrunner.py
yarikoptic: sorry -- I should have provided more context and explicit desire right away:  we have it as a `protocol` for WitlessRunner as well and just document that it should be a class and not an instance.   It would be odd to use `protocol=` there and `protocol_class=` here. IMHO it would be better to stay consistent and just keep it as a `protocol` in the signature (and you could make it into `protocol_instance` or alike in the body).
FWIW  IIRC in typing should be smth like `: Type[WitlessProtocol]`.

datalad/tests/test_nonasyncrunner.py
yarikoptic: let's consider it a codecov etc fluke, sorry for the noise

datalad/tests/test_nonasyncrunner.py
```diff
@@ -0,0 +1,159 @@
...
+from time import sleep, time
+
+from datalad.tests.utils import assert_true, eq_, known_failure_windows, \
+    known_failure_osx, with_tempfile
+
+from ..cmd import WitlessProtocol, WitlessRunner, StdOutCapture
+from ..nonasyncrunner import ReaderThread, run_command
+
+
+@known_failure_windows
```
yarikoptic, line 25: could you add a comment on what is so special about windows here?

yarikoptic:
sorry that my suggestion for bd7bb5929eb04e68483d95f19e06c7e7a085ee6f was not entire kosher . it is what broke CI and running tests locally. let me know if you need a hand to mitigate

datalad/tests/test_nonasyncrunner.py
christian-monch: > sorry that my suggestion for [bd7bb59](https://github.com/datalad/datalad/commit/bd7bb5929eb04e68483d95f19e06c7e7a085ee6f) was not entire kosher . it is what broke CI and running tests locally. let me know if you need a hand to mitigate

Don't worry, its fine. It's a small fix. I combine it with the logging improvements. :-)

datalad/cmd.py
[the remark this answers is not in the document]
christian-monch: fixed

datalad/cmd.py
```diff
@@ -671,7 +533,6 @@ def run_on_filelist_chunks(self, cmd, files, protocol=None,
 
 
 def readline_rstripped(stdout):
```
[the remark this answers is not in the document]
christian-monch: done

datalad/nonasyncrunner.py
christian-monch: Done

datalad/nonasyncrunner.py
christian-monch: I test the blocking part, although it is not a feature ;-)

datalad/tests/test_nonasyncrunner.py
christian-monch: added

datalad/cmd.py
```diff
@@ -169,26 +105,26 @@ def _log_nolog(self, *args):
...
+        fd_name = self.fd_infos[fd]
         lgr.log(5, 'Read %i bytes from %i[%s]%s',
-                len(data), self.pid, fd_name, ':' if self._log_outputs else '')
+                len(data), self.process.pid, fd_name, ':' if self._log_outputs else '')
         if self._log_outputs:
             log_data = ensure_unicode(data)
             # The way we log is to stay consistent with Runner.
             # TODO: later we might just log in a single entry, without
             # fd_name prefix
             lgr.log(5, "%s| %s ", fd_name, log_data)
```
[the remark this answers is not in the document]
christian-monch: Hard to say. The logger has to be enabled for level `5`, and the datalad configuration should set `datalad.log.outputs` to `True`. I do not see that in a quick glance over the tests.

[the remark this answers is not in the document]
kyleam: > The logger has to be enabled for level `5`, and the datalad configuration should set `datalad.log.outputs` to `True`. I do not see that in a quick glance over the tests.

https://github.com/datalad/datalad/blob/2fd46ee0dcd710ba99b1101e8172eace47dd8235/.travis.yml#L92-L100

yarikoptic:
oh no!!! the plague of stalling is upon us? please NO!  looking at red appveyor run:
```
datalad.distribution.tests.test_publish.test_publish_simple ... ok
datalad.distribution.tests.test_publish.test_publish_with_data ... 
```

datalad/cmd.py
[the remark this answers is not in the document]
christian-monch: Thanks @kyleam Didn't look there! :-)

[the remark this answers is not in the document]
christian-monch: I think `DATALAD_LOG_OUTPUTS=1` does not work because `1` is interpreted as `str`. It should be `yes` or similar to be interpreted as `True`. Changing it now, which should the execution of the hitherto untested lines.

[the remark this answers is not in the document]
christian-monch: Fixed in a456f2bab1028127d7ed6616946e2683d3986db3. See issue #5678

yarikoptic:
FTR: benchmarks are still happy and suggestive of improvement for quickly running commands:
```
      2.55±0.06ms      2.20±0.07ms    ~0.86  core.witlessrunner.time_echo
       3.08±0.1ms      2.65±0.04ms    ~0.86  core.witlessrunner.time_echo_gitrunner
```

datalad/cmd.py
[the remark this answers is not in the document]
yarikoptic: FWIW, changes highlighting says that it is covered now.  Might had been a fluke or smth fixed/extended.  I will consider this resolved.

christian-monch:
@yarikoptic: from my point of view this should work. It seems that the tests that are execute with DATALAD_LOG_LEVEL=2 do not contribute to the coverage database. @kyleam Maybe you know why?

datalad/cmd.py
```diff
@@ -349,6 +289,10 @@ def run(self, cmd, protocol=None, stdin=None, cwd=None, env=None, **kwargs):
         FileNotFoundError
           When a given executable does not exist.
         """
+
+        # imported here to avoid recursive importing
```
yarikoptic, line 293: note: is happening likely due to `from .cmd import WitlessProtocol` there used solely for typing annotation.  Could have been resolved by moving protocols into their own `cmd_protocols.py` breaking the loop and making it all a bit more kosher.  I will leave to @christian-monch to decide on his cuisine .

datalad/nonasyncrunner.py
```diff
@@ -0,0 +1,184 @@
...
+            if data == b"":
+                logger.debug("%s exiting (stream end)", self)
+                self.queue.put((self.file.fileno(), None, time.time()))
+                return
+
+            self.queue.put((self.file.fileno(), data, time.time()))
+
+
+def run_command(cmd: Union[str, List],
+                protocol_class: Type[WitlessProtocol],
```
yarikoptic, line 85: uff -- spent some time chasing it in "resolved" comments.  My feeling grow stronger to keep it consistent here with runner itself and have just `protocol` not `protocol_class`.  My comments were:

 I should have provided more context and explicit desire right away: we have it as a protocol for WitlessRunner as well and just document that it should be a class and not an instance. It would be odd to use `protocol=` there and `protocol_class=` here. IMHO it would be better to stay consistent and just keep it as a protocol in the signature (and you could make it into protocol_instance or alike in the body).

yarikoptic:
Great job @christian-monch !  This PR is nearly perfectly done, and could only be made a bit more kosher (left 2 comments on that) so we do not need to deal with indigestion later on ;-)

datalad/nonasyncrunner.py
christian-monch: Very well, `protocol` it is.

datalad/cmd.py
christian-monch: We do it the proper way, as you suggested: extract the protocols into their own file.

christian-monch:
> Great job @christian-monch ! This PR is nearly perfectly done, and could only be made a bit more kosher (left 2 comments on that) so we do not need to deal with indigestion later on ;-)

Both dealt with :-)

yarikoptic (approved):

yarikoptic:
FWIW, note on benchmarks -- short commands get consistently faster but surprisingly the beast of a studyforrest one is consistently slower a bit:

```shell
(tinuous-dev) datalad@smaug:/mnt/datasets/datalad/ci/logs/2021/05$ grep 'usecases.study_forrest.time_make_studyforrest_mockup' */pr/*5667/*/*success/*/8*txt
19/pr/5667/d1888cf/github-Benchmarks-3085-success/vs-master/8_Compare.txt:2021-05-19T20:42:29.0192372Z         45.5±0.1s       48.7±0.04s     1.07  usecases.study_forrest.time_make_studyforrest_mockup
20/pr/5667/1424902/github-Benchmarks-3108-success/vs-master/8_Compare.txt:2021-05-20T19:50:02.2174723Z         40.4±0.1s        43.3±0.2s     1.07  usecases.study_forrest.time_make_studyforrest_mockup
20/pr/5667/144f8f9/github-Benchmarks-3099-success/vs-master/8_Compare.txt:2021-05-20T11:25:36.2577054Z         43.0±0.1s        47.0±0.1s     1.09  usecases.study_forrest.time_make_studyforrest_mockup
20/pr/5667/7f18e9f/github-Benchmarks-3092-success/vs-master/8_Compare.txt:2021-05-20T08:15:15.2595110Z         44.2±0.2s        47.0±0.5s     1.06  usecases.study_forrest.time_make_studyforrest_mockup
20/pr/5667/7f6cc85/github-Benchmarks-3094-success/vs-master/8_Compare.txt:2021-05-20T10:19:06.5116745Z         45.2±0.6s        51.2±0.5s    ~1.13  usecases.study_forrest.time_make_studyforrest_mockup
20/pr/5667/e8e43ff/github-Benchmarks-3100-success/vs-master/8_Compare.txt:2021-05-20T12:59:23.9873648Z        42.9±0.03s       45.9±0.02s     1.07  usecases.study_forrest.time_make_studyforrest_mockup
(tinuous-dev) datalad@smaug:/mnt/datasets/datalad/ci/logs/2021/05$ grep 'core.witlessrunner.time_echo_gitrunner' */pr/*5667/*/*success/*/8*txt
19/pr/5667/d1888cf/github-Benchmarks-3085-success/vs-master/8_Compare.txt:2021-05-19T20:42:29.0181216Z -     3.32±0.08ms       2.80±0.1ms     0.84  core.witlessrunner.time_echo_gitrunner
19/pr/5667/d1888cf/github-Benchmarks-3085-success/vs-master/8_Compare.txt:2021-05-19T20:42:29.0182450Z       3.51±0.03ms      3.27±0.05ms     0.93  core.witlessrunner.time_echo_gitrunner_fullcapture
20/pr/5667/1424902/github-Benchmarks-3108-success/vs-master/8_Compare.txt:2021-05-20T19:50:02.2164221Z       2.79±0.05ms      2.45±0.04ms    ~0.88  core.witlessrunner.time_echo_gitrunner
20/pr/5667/1424902/github-Benchmarks-3108-success/vs-master/8_Compare.txt:2021-05-20T19:50:02.2165249Z        3.06±0.1ms      2.77±0.05ms    ~0.90  core.witlessrunner.time_echo_gitrunner_fullcapture
20/pr/5667/144f8f9/github-Benchmarks-3099-success/vs-master/8_Compare.txt:2021-05-20T11:25:36.2564968Z       2.98±0.05ms       2.62±0.1ms    ~0.88  core.witlessrunner.time_echo_gitrunner
20/pr/5667/144f8f9/github-Benchmarks-3099-success/vs-master/8_Compare.txt:2021-05-20T11:25:36.2566224Z       3.26±0.04ms      3.10±0.09ms     0.95  core.witlessrunner.time_echo_gitrunner_fullcapture
20/pr/5667/7f18e9f/github-Benchmarks-3092-success/vs-master/8_Compare.txt:2021-05-20T08:15:15.2581174Z -     3.09±0.09ms      2.76±0.04ms     0.89  core.witlessrunner.time_echo_gitrunner
20/pr/5667/7f18e9f/github-Benchmarks-3092-success/vs-master/8_Compare.txt:2021-05-20T08:15:15.2582573Z        3.37±0.2ms      3.11±0.09ms     0.92  core.witlessrunner.time_echo_gitrunner_fullcapture
20/pr/5667/7f6cc85/github-Benchmarks-3094-success/vs-master/8_Compare.txt:2021-05-20T10:19:06.5105150Z       3.17±0.06ms      2.94±0.04ms     0.93  core.witlessrunner.time_echo_gitrunner
20/pr/5667/7f6cc85/github-Benchmarks-3094-success/vs-master/8_Compare.txt:2021-05-20T10:19:06.5106280Z       3.31±0.06ms       3.18±0.2ms     0.96  core.witlessrunner.time_echo_gitrunner_fullcapture
20/pr/5667/e8e43ff/github-Benchmarks-3100-success/vs-master/8_Compare.txt:2021-05-20T12:59:23.9861584Z        3.08±0.1ms      2.65±0.04ms    ~0.86  core.witlessrunner.time_echo_gitrunner
20/pr/5667/e8e43ff/github-Benchmarks-3100-success/vs-master/8_Compare.txt:2021-05-20T12:59:23.9862804Z       3.21±0.06ms       2.98±0.2ms     0.93  core.witlessrunner.time_echo_gitrunner_fullcapture
```

yarikoptic:
ok, @mih supported my motion, so totals to 2 approvals, let's proceed and fix if there is anything to fix after

datalad/cmd.py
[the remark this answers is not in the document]
jwodder: @yarikoptic `readline_rstripped()` is actually used as the default `output_proc` by `BatchedCommand`, so now using `BatchedAnnex` produces a warning.

Issue #5628, opened 2021-05-03
Undesired RIA URL reconfiguration


mih:
I have a superdataset with

```
% datalad x-configuration | grep subdataset | head -1
datalad.get.subdataset-source-candidate-100origin=ria+ssh://bulk1.htc.inm7.de/ds/studyforrest/srv#{id}
```

because I want all subdataset to be cloned from an internal store, despite `.gitmodules` pointing to public sources.

That in turn causes `clone` to reconfigure the access URL for an ORA special remote to use this store URL. However, the ORA remote is already configured properly have this as `pushurl`, but uses a more widely accessible, different `http` URL from read-access. This configuration is replaced by a uniform SSH setup. Whenever I push back from this clone to the store, all `http` access to this remote is lost.

ping @bpoldrack

bpoldrack:
FTR: Two things instantly coming to mind:
1. Reconfiguration needs double-check to account for the "new" url/push-url distinction
2. Still strange, since reconfiguration is supposed to only happen, if the stored access method fails.

mih:
Sounds sensible. From the top of my head I do not know when we would ever want to commit.

bpoldrack:
Cool. Since local special remote configs are now officially supported by annex, will switch entirely.

mih:
It will be critical to check which annex version is required for that.

Also, a simple and _quick_ fix first, and an improvement later would be much appreciated. I already spend hours undoing the goodness

bpoldrack:
Not sure there is a quick fix, though. We explicitly introduced the idea, that we do want a reconfiguration, if the URL of the ORA remote does not match the clone URL, despite both belonging to the same store. I could introduce a special case where we say: "If one of `url`/`push-url` matches, don't touch it", but while this may solve this particular case, it's still undesireable in another. Say you clone that dataset locally, then you would still go through HTTP to retrieve data in that clone (while ATM you'd get a reconfiguration giving you local access).

bpoldrack:
There's another aspect. This doesn't seem true to me:

> Whenever I push back from this clone to the store, all http access to this remote is lost.

If anyone clones the changed dataset from that store via HTTP, they should get a reconfiguration to HTTP, too. While that committed ping-pong remains annoying, there's no data access loss, that I can see.

bpoldrack:
> I could introduce a special case where we say: "If one of url/push-url matches, don't touch it"

[...]

PR #5672, opened 2021-05-19
Fix post-clone reconfiguration for ORA remote

bpoldrack:
See #5628. We did not consider the recently introduced `push-url` of ORA
special remotes for matching against the clone URL. With this patch we
stick to the logic that tries a reconfiguration if either all known ORA
remotes failed to autoenable or none of the enabled ORA remotes have a
matching URL on record.

However, as pointed out in #5628, this only is a quick fix for a
particular usage pattern. A general solution for the reconfiguration
mess appears to be to switch to local configs only.

- [x] add test

### DIFF
--- a/datalad/core/distributed/clone.py
+++ b/datalad/core/distributed/clone.py
@@ -832,14 +832,27 @@ def postclonecfg_ria(ds, props, remote="origin"):
     srs = repo.get_special_remotes() \
         if hasattr(repo, 'get_special_remotes') else dict()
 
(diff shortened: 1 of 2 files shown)


mih (approved):
LGTM! Thx MUCH...this pained me substantially over the past weeks.

kyleam (approved):
Read through gh-5628 and stepped through the changed code (as triggered by the new test).  Looks good to me too.

Issue #5668, opened 2021-05-17
cannot `drop --from ORAREMOTE`


mih:
`git annex drop --unused --from joc-storage`, where `joc-storage` is an ORA special remote. Talking via `ria+file://` i.e. `LocalIO()`

```
...
drop joc-storage MD5E-s2779420--0250bcb1c57ad101113dee93c5a29022.tar.gz 
  [Errno 13] Permission denied: '/data/project/ukb_vbm/outputstore/893/8de76-0302-45b5-9825-3c6ce3f3fffe/annex/objects/08/V4/MD5E-s2779420--0250bcb1c57ad101113dee93c5a29022.tar.gz/MD5E-s2779420--0250bcb1c57ad101113dee93c5a29022.tar.gz'
failed
```

I own that remote store:

```
% ls -l /data/project/ukb_vbm/outputstore/893/8de76-0302-45b5-9825-3c6ce3f3fffe/annex/objects/08/V4/MD5E-s2779420--0250bcb1c57ad101113dee93c5a29022.tar.gz/MD5E-s2779420--0250bcb1c57ad101113dee93c5a29022.tar.gz
```

annex debug

```
drop joc-storage MD5E-s45810--00cd85eb2bed0909e7334ceddfce98d4.tar.gz [2021-05-17 12:58:16.383804023] Dropping from remote Remote { name ="joc-storage" } proof: Nothing                 
[2021-05-17 12:58:16.386627744] process [201873] chat: /home/mih/venvs/datalad/bin/git-annex-remote-ora []
[2021-05-17 12:58:16.575321454] /home/mih/venvs/datalad/bin/git-annex-remote-ora[1] --> VERSION 1
[2021-05-17 12:58:16.575448616] /home/mih/venvs/datalad/bin/git-annex-remote-ora[1] <-- EXTENSIONS INFO GETGITREMOTENAME ASYNC                                                           
[2021-05-17 12:58:16.575578517] /home/mih/venvs/datalad/bin/git-annex-remote-ora[1] --> EXTENSIONS
[2021-05-17 12:58:16.575647338] /home/mih/venvs/datalad/bin/git-annex-remote-ora[1] <-- PREPARE
[2021-05-17 12:58:16.575916161] /home/mih/venvs/datalad/bin/git-annex-remote-ora[1] --> GETGITDIR
[2021-05-17 12:58:16.576018682] /home/mih/venvs/datalad/bin/git-annex-remote-ora[1] <-- VALUE .git
[...]

PR #5673, opened 2021-05-19
Drafting an approach to #5668

bpoldrack:
As mentioned in #5668, I think the main issue is to get into the situation in the first place. Generally, we should respect whatever permissions are set in-store, instead of obtaining write permissions (although I see that it's kind of fine, if we can obtain them).
The assumption that git-annex "touched" the object tree and that we need to deal with that seems dangerous to me, as git-annex should not be aware of the object tree in the store. It seems better to come up with appropriate solutions to make it much less likely for anything like that to happen. However, here is a "fix". For said reasons I intentionally limited the scope in which it is applied (only things directly involved in writing to a potentially protected object tree, as opposed to all write operations on the remote end). ORA/RIA-related stuff is undergoing some rewriting/refactoring and I think this is to be revisited afterwards.

Please check, whether this (check the test) is sufficient for now from your point of view, @mih!

- [x] add test
- [x] equivalent approach for SSH

### DIFF
--- a/datalad/distributed/ora_remote.py
+++ b/datalad/distributed/ora_remote.py
@@ -3,10 +3,12 @@ from annexremote import RemoteError
 from annexremote import ProtocolError
 
 import os
+import stat
 from pathlib import (
     Path,
     PurePosixPath
 )
+from contextlib import contextmanager
 import requests
 import shutil
 from shlex import quote as sh_quote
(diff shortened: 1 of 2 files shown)


mih:
I am not sure what convinced me, the repeated non-objection-notes, or the relief that git-annex would still not print "NOT CONVINCED" on each drop ;-)

In any case, I will close the original issue, and leave it to the store maintainers to fix the permissions.

thx for the resistance

PR #5676, opened 2021-05-19
RF: just pip install datalad-installer

yarikoptic:
master branch has its own ideas and binds to some older version
```
$> git grep datalad.installer
.appveyor.yml:  # deploy the datalad installer, override version via DATALAD_INSTALLER_VERSION
.appveyor.yml:      python -m pip install "datalad-installer%DATALAD_INSTALLER_VERSION%"
.appveyor.yml:      python -m pip install datalad-installer
.appveyor.yml:  - sh: python -m pip install datalad-installer${DATALAD_INSTALLER_VERSION:-}

```
needs yet to be addressed, decided to just go with maint for now and simplify (not versioned depends etc)

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -79,8 +79,9 @@ environment:
       # Python version specification is non-standard on windows
       PY: 39-x64
       # TODO: use datalad/git-annex (github release packages) but
-      # it would need setup of a GITHUB_TOKEN to access
-      INSTALL_GITANNEX: git-annex -m datalad/packages
+      # it would need setup of a GITHUB_TOKEN to access.
+      # This one is set in master but kept without change in maint for now
+      # INSTALL_GITANNEX: git-annex -m datalad/packages
     # MacOS core tests
     - ID: MacP38core
       # ~25min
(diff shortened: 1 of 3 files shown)


mih:
Just commenting, have no checked yet. It would be good to know if a diff to the extension setup exists, and if so why, and if all extensions need to be modified.

mih:
It is at least missing the ability to switch the installer version without editing .appveyor.yml. The feature was introduces in the extension setup the last time the installer behavior broke, and all extensions had to be fix before any CI would work again.


```yml
install:
...
  # deploy the datalad installer, override version via DATALAD_INSTALLER_VERSION
  - cmd:
      IF DEFINED DATALAD_INSTALLER_VERSION (
      python -m pip install "datalad-installer%DATALAD_INSTALLER_VERSION%"
      ) ELSE (
      python -m pip install datalad-installer
      )
  - sh: python -m pip install datalad-installer${DATALAD_INSTALLER_VERSION:-}

```

it would be great if the setups converge.

yarikoptic:
> it would be great if the setups converge.

it would be great if the setup didn't diverge ;-)  copied `master`'s .appveyor with only slight tune up for maint, and then improved upon.  So should be re-converging now

yarikoptic:
hm, codecov offset is too big. #5688 against maint has only 0.1% . looking at diff in coverage - might relate to travis runs for high log level to not provide results, but also might relate to added `INSTALL_GITANNEX: git-annex -m datalad/packages`. so I will comment it out and give another full round to see if codecov report comes cleaner.

yarikoptic:
ok, had to restart one appveyor on mac. now all green and codecov divergence is small. so we should be good to go

mih (approved):
Wonderful, thx @yarikoptic for the updates!

PR #5683, opened 2021-05-20
Enhance permission error message for ORA

bpoldrack:
With PR #5673 closed, address #5668 by improving error message.

Ping @mih

### DIFF
--- a/datalad/distributed/ora_remote.py
+++ b/datalad/distributed/ora_remote.py
@@ -205,7 +205,12 @@ class LocalIO(IOBase):
         src.rename(dst)
 
     def remove(self, path):
-        path.unlink()
+        try:
+            path.unlink()
+        except PermissionError as e:
+            raise RIARemoteError(str(e) + os.linesep +
+                                 "Note: Write permissions for a key's parent"
+                                 "directory are also required to drop content.")
 
     def remove_dir(self, path):
         path.rmdir()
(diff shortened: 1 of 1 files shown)


datalad/distributed/ora_remote.py
```diff
@@ -205,7 +205,13 @@ def rename(self, src, dst):
...
 
     def remove(self, path):
-        path.unlink()
+        try:
+            path.unlink()
+        except PermissionError as e:
+            raise RIARemoteError(str(e) + os.linesep +
+                                 "No support for fixing your permissions. Note,"
+                                 " that write permissions to the directory are "
+                                 "required to drop remote content.")
```
mih, line 214:
```suggestion
                                 "Note: Write permissions for a key's parent directory are "
                                 "also required to drop content.")
```

mih: I suggested a change. It is not at all given that the permissions are "mine", or that they could be fixed with my level of authorization.

datalad/distributed/ora_remote.py
```diff
@@ -486,7 +492,17 @@ def rename(self, src, dst):
...
+            self._run('rm {}'.format(sh_quote(str(path))))
+        except RemoteCommandFailedError as e:
+            if "permission denied" in str(e).lower():
+                raise RIARemoteError(
+                    str(e) + os.linesep +
+                    "No support for fixing your permissions. Note,"
+                    " that write permissions to the directory are "
+                    "required to drop remote content.")
+            else:
+                raise
```
mih, line 505: In addition to the wording change (see above), I would prefer to remove the conditional. The exception is given in both cases anyways and the additional not is just that. The chances to miss it because of an `rm` that reports slightly different are bigger than the confusion, IMHO.

mih (changes requested):
Thx, I left two comments, please have a look.

mih:
@bpoldrack Can you take this over the finish line. Thx!

bpoldrack:
@mih  Done, I guess.

mih (approved):
Looks good, thanks!

PR #5686, opened 2021-05-20
Record the use of --explicit in a run record 

### LINKED ISSUE #5547: record --explicit flag of a run call?
At present a `run` record does not report whether it was produced with `--explicit` or not. That forces a user to decide on `rerun` whether or not to use `--explicit` (and its docstring rightfully suggests to be careful with that).

However, besides running on a dirty dataset, the `--explicit` flag is instrumental in saving only select outputs. In that case, running `rerun` without explicit will lead to different results, even if all outcomes are bit-identical. Rerun only allows to put an an entire rerun run ;-) into explicit mode, but in practice the explicitness can vary from record to record.

Maybe

1. starting on a dirty dataset
2. only saving explicitly listed outputs

are two different things that should not be triggered together with a single option. (2) looks like a candidate for inclusion into a run record itself.

adswa:
In today's labmeeting we agreed to tackle UX issues more heavily. In an attempt to fix #5547, this PR records the use of ``--explicit`` in a run record.
In an attempt to start a conversation about how to handle the presence of ``explicit = true`` records during rerun, the second commit in this PR also (in a quite encroaching manner) lets the original ``--explicit`` flag "override" a non-set ``--explicit`` flag in a rerun. The second commit is not sufficient, because it completely lacks critical thoughts or solutions on what @kyleam noted in the original issue:

> At least one thing that would need to be carefully thought through is the interaction between rerunning explicit and non-explicit commits. For example, if rerun first processes an explicit=True commit followed by an explict=False one, rerunning without any further checks will result in that commit absorbing unrelated changes.

I secretly hope that someone is so offended by this bad commit that it starts a conversation on how to handle this. I couldn't come up with a solution, but maybe someone else has a good idea. At the very least, with the first commit ``--explicit`` is recorded, even if we decide to not yet act on it.


kyleam:
@adswa Thanks for starting to work on this.

Is your second commit doing a bit more than you intend?  Once an explicit run is encountered, `run` is called with `explict=True` for each following run commit, even those that weren't originally created with `explicit=True`.  A minimal change for that would be something like

<details>
<summary>diff</summary>

```diff
diff --git a/datalad/interface/rerun.py b/datalad/interface/rerun.py
index 6673b7ea33..da76099970 100644
--- a/datalad/interface/rerun.py
+++ b/datalad/interface/rerun.py
@@ -502,22 +502,24 @@ def _rerun(dset, results, assume_ready=None, explicit=False):
             # if the original run command was invoked with --explicit, but the
             # rerun is not, the original explicit takes precedence, but we
             # report it to the user.
             if original_explicit and not explicit:
-                explicit = original_explicit
+                explicit_arg = original_explicit
                 lgr.info(
                     "Run commit %s was invoked using --explicit. The "
                     "rerun of this commit will set --explicit, too.",
                     res_hexsha[:7]
                 )
+            else:
+                explicit_arg = explicit
             message = res["rerun_message"] or res["run_message"]
             for r in run_command(run_info['cmd'],
                                  dataset=dset,
                                  inputs=run_info.get("inputs", []),
                                  extra_inputs=run_info.get("extra_inputs", []),
                                  outputs=outputs,
                                  assume_ready=assume_ready,
-                                 explicit=explicit,
+                                 explicit=explicit_arg,
                                  rerun_outputs=auto_outputs,
                                  message=message,
                                  rerun_info=run_info):
                 yield r
```

</details>

---

I don't have any great ideas on how to handle unsaved output from an explicit run leaking into the commit for the next non-explicit run.  The main thing I can think of is to check whether the working tree is dirty (at least between rerunning an explicit command and a non-explicit one) and call `save` to create a dedicated commit.  That's ugly, but I think it's better than unrelated output leaking into the next run.

Fwiw here's a demo showing the problem.

<details>
<summary>demo</summary>

```sh
set -eu

datalad create
datalad run -o f1 --explicit "echo f1 >f1 && touch f2"
rm f2
datalad run "echo f3 >f3"

datalad rerun --onto= --since=
git log --oneline --stat
```

Results (this PR plus my patch above).  Notice f2 leaked into the last run commit.

```
[...]
45e3ffc (HEAD) [DATALAD RUNCMD] echo f3 >f3
 f2 | 1 +
 f3 | 1 +
20facde [DATALAD RUNCMD] echo f1 >f1 && touch f2
 f1 | 1 +
 1 file changed, 1 insertion(+)
00cc088 [DATALAD] new dataset
 .datalad/.gitattributes | 4 ++++
 .datalad/config         | 2 ++
 .gitattributes          | 3 +++
 3 files changed, 9 insertions(+)
```

</details>

yarikoptic:
FTR, a sample failed travis run was failing "rightfully" (in that it was not some spurious error or fail, didn't analyze changes):
```
FAIL: datalad.interface.tests.test_rerun.test_rerun_explicit
  File "/tmp/dl-miniconda-gqpfkekb/lib/python3.8/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/tmp/dl-miniconda-gqpfkekb/lib/python3.8/site-packages/datalad/tests/utils.py", line 1897, in _wrap_skip_if_adjusted_branch
    return func(*args, **kwargs)
  File "/tmp/dl-miniconda-gqpfkekb/lib/python3.8/site-packages/datalad/tests/utils.py", line 560, in _wrap_with_tree
    return t(*(arg + (d,)), **kw)
  File "/tmp/dl-miniconda-gqpfkekb/lib/python3.8/site-packages/datalad/interface/tests/test_rerun.py", line 793, in test_rerun_explicit
    eq_(orig_head, ds.repo.get_hexsha(DEFAULT_BRANCH + "~1"))
AssertionError: '4a04c769fc71b34c4009c980f0fad78c9dbff296' != '3fdb7ec73e6e5a6ebafc8eaff5a28e827adc05bc'
```
@adswa  - planing to bring this over the finish line?

adswa:
I failed with this PR, closing as stale

Issue #5685, opened 2021-05-20
Add LICENSE to the repo


jkanche:
I notice the license in the `README`, but would be great to explicitly add a `LICENSE` file to the repo

kyleam:
It's there, just with a different name: COPYING.

adswa:
Hey @jkanche, I think this comes in the context of the JOSS review, right? As @kyleam said, we keep the license (MIT) in the ``COPYING`` file - how strict do you think is that item on the review checklist to be interpreted? :smile: I actually don't know where difference in naming comes from, but judging from https://stackoverflow.com/questions/5678462/should-i-provide-a-license-txt-or-copying-txt-file-in-my-project  both are in standard use.

adswa:
hey @jkanche, just a quick ping on this. I feel like the presence of a license in a text file with a commonly used name should suffice, but we could also ask the editor in the JOSS issue about this.

arokem:
From the point of view of the JOSS review, it's fine for it to be where it is.

@jkanche : can this issue be closed? Does this check your "License" box in the review?

jkanche:
this should work, Thank you!

PR #5687, opened 2021-05-20
DOC: up-to-date COPYING

yarikoptic:
inspired by #5685 

old mentions are being moved to -deprecated: https://github.com/datalad/datalad-deprecated/pull/25 where they are still (?) relevant


yarikoptic:
Let's proceed, improve more if I miss anything

PR #5691, opened 2021-05-21
Introduce datalad.log.exc and disable exception logging by default

### LINKED ISSUE #4478: Make exc_str() silent by default

bpoldrack:
This is originally motivated by wanting to close #4478. So, there's a config `datalad.log.exc` now, that controls whether exceptions are "rendered" in log messages (and it's off by default).

However, once at it, I took the occasion to somewhat address our exception handling in general. When we catch and report on an exception, we pretty much always would use `exc_str` (or now `CapturedException`), providing an opportunity to let that also do the clean up of stack frame references and thereby somewhat enforcing it. Hence, decided against the proposed approach in the issue to extend the to-be-logged exception and pass it on, but introduced a class that represents the textual information about it (and its traceback) without any stack references. Remember: circular references. Furthermore, since @yarikoptic introduced result reporting on exceptions as well (using `exc_str`, too), there's more to be done with that class - thinking of have a central place to enhance a result record with (possibly several) fields about an exception, so that a renderer can decide what parts to make use of. Also: we can have python's standard output of an exception traceback, too - might be useful for a better "debug mode". But all that should be a follow-up. Here I only address the logging directly and turn `exc_str` into a "dummy". Last remark: In the issue the idea was to have the config be handled by the log formatter. That makes sense, but isn't that straightforward really. There can be different formatters including stdlib ones, that still should be able to deal with an object passed to a log record. That's why it needs a `__str__` anyway. We also use the standard formatter class with `swallow_log` and all the log output testing. So, moving behavior control to `ColorFormatter` (or several formatters?) should be aimed for, I think, but that implies some more reworking of log testing (and we do that a lot with way too precise matches IMO).

Remaining question from my POV: If exception logging is disabled, should it really be an empty string? From a UX standpoint, I think even an exception name can be confusing (see example in original issue - no exc info needed whatsoever), but with most(?) our current messages there's probably a `:` or `()` remaining that way. I'd argue that we should rewrite those messages, so that they make sense either way. If they don't it's probably not yet a good message anyway. WDYT, @datalad/developers?

- Closes #4478

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -51,6 +51,8 @@ shallow_clone: true
 
 environment:
   DATALAD_TESTS_SSH: 1
+  # The following is currently necessary as we test a lot of logging output in high detail:
+  DATALAD_LOG_EXC: 1
 
   # Do not use `image` as a matrix dimension, to have fine-grained control over
   # what tests run on which platform
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -569,8 +569,7 @@ Refer datalad/config.py for information on how to add these environment variable
 - *DATALAD_LOG_VMEM*:
   Reports memory utilization (resident/virtual) at every log line, needs `psutil` module
 - *DATALAD_EXC_STR_TBLIMIT*: 
-  This flag is used by the datalad extract_tb function which extracts and formats stack-traces.
-  It caps the number of lines to DATALAD_EXC_STR_TBLIMIT of pre-processed entries from traceback.
+  This flag is used by datalad to cap the number of traceback steps included in exception logging and result reporting to DATALAD_EXC_STR_TBLIMIT of pre-processed entries from traceback.
 - *DATALAD_SEED*:
   To seed Python's `random` RNG, which will also be used for generation of dataset UUIDs to make
   those random values reproducible.  You might want also to set all the relevant git config variables
--- a/datalad/customremotes/base.py
+++ b/datalad/customremotes/base.py
@@ -320,7 +320,7 @@ class AnnexCustomRemote(object):
             try:
                 method(*req_load)
             except Exception as e:
-                self.error("Problem processing %r with parameters %r: %r"
+                self.error("Problem processing %r with parameters %r: %s"
                            % (req, req_load, exc_str(e)))
                 from traceback import format_exc
                 lgr.error("Caught exception detail: %s", format_exc())
--- a/datalad/dochelpers.py
+++ b/datalad/dochelpers.py
@@ -18,7 +18,9 @@ import textwrap
 import os
 import sys
 import traceback
+from pathlib import Path
 
+from datalad.support.exceptions import CapturedException
 
 lgr = logging.getLogger("datalad.docutils")
 
@@ -314,7 +316,7 @@ def borrowkwargs(cls=None, methodname=None, exclude=None):
 
 
 # TODO: make limit respect config/environment parameter
-def exc_str(exc=None, limit=None, include_str=True):
+def exc_str_old(exc=None, limit=None, include_str=True):
     """Enhanced str for exceptions.  Should include original location
 
     Parameters
@@ -372,3 +374,10 @@ def exc_str(exc=None, limit=None, include_str=True):
         # https://docs.python.org/2/library/sys.html#sys.exc_info
         del tb
     return out
+
+
+def exc_str(exc=None, limit=None, include_str=True):
+    """Temporary test shim, for finding issues re refactoring for
+    using CapturedExcpetion instead."""
+
+    return str(CapturedException.capture_exc(exc))
--- a/datalad/dochelpers.py
+++ b/datalad/dochelpers.py
@@ -378,6 +378,10 @@ def exc_str_old(exc=None, limit=None, include_str=True):
 
 def exc_str(exc=None, limit=None, include_str=True):
     """Temporary test shim, for finding issues re refactoring for
-    using CapturedExcpetion instead."""
+    using CapturedException instead.
 
-    return str(CapturedException.capture_exc(exc))
+    See gh-5716
+    """
+
+    return str(CapturedException(exc))
+
--- a/datalad/support/exceptions.py
+++ b/datalad/support/exceptions.py
@@ -10,8 +10,152 @@
 """
 
 import re
+import sys
+import traceback
 from os import linesep
+from pathlib import Path
 from pprint import pformat
+
+
+class CapturedException(object):
+    """This class represents information about an occurred exception (including
+    its traceback), while not holding any references to the actual exception
+    object or its traceback, frame references, etc.
+
+    Just keep the textual information for logging or whatever other kind of
+    reporting.
+    """
+
+    @classmethod
+    def capture_exc(cls, exc=None, limit=None, capture_locals=False):
+        """Capture an exception and its traceback for logging.
+
+        Clears the exception's traceback frame references afterwards.
+
+        Parameters
+        ----------
+        exc: Exception or None
+          If None, rely on sys.exc_info() to get the latest.
+        limit: int
+          Note, that this is limiting the capturing of the exception's
+          traceback depth. Formatting for output comes with it's own limit.
+        capture_locals: bool
+          Whether or not to capture the local context of traceback frames.
+
+        Returns
+        -------
+        CapturedException
+        """
+        # Note, that with lookup_lines=False the lookup is deferred,
+        # not disabled. Unclear to me ATM, whether that means to keep frame
+        # references around, but prob. not. TODO: Test that.
+
+        if exc is None:
+            exctype, value, tb = sys.exc_info()
+            tb_exc = traceback.TracebackException(
+                exctype, value, tb,
+                limit=limit,
+                lookup_lines=True,
+                capture_locals=capture_locals
+            )
+            traceback.clear_frames(tb)
+        else:
+            tb_exc = traceback.TracebackException.from_exception(
+                exc,
+                limit=limit,
+                lookup_lines=True,
+                capture_locals=capture_locals
+            )
+            traceback.clear_frames(exc.__traceback__)
+
+        return CapturedException(tb_exc)
+
+    def __init__(self, tb_exc):
+        """
+
+        Parameters
+        ----------
+        tb_exc: traceback.TracebackException
+        """
+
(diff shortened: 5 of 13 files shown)


bpoldrack:
FTR: It seems a change like this requires a change in `swallow_logs` as that construction circumvents our Formatter.

bpoldrack:
FTR: There also is an improvement over `exc_str`: Traceback now includes chained exceptions ("during handling ...", "the above is the direct cause ..")

.appveyor.yml
```diff
@@ -51,6 +51,8 @@ shallow_clone: true
 
 environment:
   DATALAD_TESTS_SSH: 1
+  # The following is currently necessary as we test a lot of logging output in high detail:
+  DATALAD_LOG_EXC: 1
```
mih, line 55: This would be needed across all extensions, right?

datalad/dochelpers.py
```diff
@@ -372,3 +374,10 @@ def exc_str(exc=None, limit=None, include_str=True):
         # https://docs.python.org/2/library/sys.html#sys.exc_info
         del tb
     return out
+
+
+def exc_str(exc=None, limit=None, include_str=True):
+    """Temporary test shim, for finding issues re refactoring for
+    using CapturedExcpetion instead."""
```
mih, line 381: Maybe add a note which condition must be fulfilled in order for this shim to be removed.

datalad/customremotes/base.py
```diff
@@ -321,7 +321,7 @@ def _loop(self):
                 method(*req_load)
             except Exception as e:
                 self.error("Problem processing %r with parameters %r: %r"
-                           % (req, req_load, exc_str(e)))
+                           % (req, req_load, str(exc_str(e))))
```
mih, line 324: What makes this call to `exc_str()` (together the few additional ones below) different from the ~150 other calls that are not adjusted?

mih (commented):
I think this is a good way and intermediate step. Thx! Some aspects are not fully clear to me yet, though. I left a few comments.

CONTRIBUTING.md
```diff
@@ -569,8 +569,7 @@ Refer datalad/config.py for information on how to add these environment variable
 - *DATALAD_LOG_VMEM*:
   Reports memory utilization (resident/virtual) at every log line, needs `psutil` module
 - *DATALAD_EXC_STR_TBLIMIT*: 
-  This flag is used by the datalad extract_tb function which extracts and formats stack-traces.
-  It caps the number of lines to DATALAD_EXC_STR_TBLIMIT of pre-processed entries from traceback.
+  This flag is used by datalad to cap the number of traceback steps included in exception logging and result reporting to DATALAD_EXC_STR_TBLIMIT of pre-processed entries from traceback.
```
yarikoptic, line 572: minor: not that we are 100% consistent in this file, but generally we do wrap lines in this file.

datalad/customremotes/base.py
yarikoptic: why `str()` here -- if needed, may be just replace last `%r` with `%s`?
or overall -- why to not chase all `exc_str` invocations to be placed under `str()` and just make the new `exc_str` to return that string representation and have e.g.

```
def exc_cap(e: Exception) -> CapturedException:
   return CapturedException.capture_exc(e)
```
to use where non-str form is desired.  This would minimize the diff I guess making it easier to grasp etc.

datalad/support/exceptions.py
```diff
@@ -10,9 +10,153 @@
...
+        else:
+            tb_exc = traceback.TracebackException.from_exception(
+                exc,
+                limit=limit,
+                lookup_lines=True,
+                capture_locals=capture_locals
+            )
+            traceback.clear_frames(exc.__traceback__)
+
+        return CapturedException(tb_exc)
```
yarikoptic, line 71: if this is the only way to instantiate this class (from an exception), why bother with this `capture_exc` `@classmethod` and not just to make it into `__init__(self, exc, ...)`?

yarikoptic (changes requested):
Overall -- I think it would be nice to have a class to "absorb" details about an exception. Having it though is orthogonal to desire to close #4478 which could be done simply by tuning default level for current `exc_str`.

I have left a few comments/questions on the design. The main one -- why not just to make `exc_str` return the string representation of your `CapturedException` thus avoiding all those `str(exc_str(exc))`?

.appveyor.yml
bpoldrack: If they test logging output to the same extend - yes.

datalad/dochelpers.py
bpoldrack: Yeah. If we agree on this PR, I'll put "plans" into an issue and reference it here.

datalad/customremotes/base.py
bpoldrack: > may be just replace last %r with %s

Yeah, good point.

datalad/support/exceptions.py
bpoldrack: >  why not just to make exc_str return the string representation of your CapturedException thus avoiding all those str(exc_str(exc))

Mainly to make sure, that I discover what exactly (Formatters - what else?) would need to deal with an instance of `CapturedException`. Furthermore - thinking of further RF'ing - from my POV those all almost always the wrong places to decide how something should be formatted. So, I do want to break that pattern, when I replace the calls to `exc_str` entirely. Formatting decisions should ultimately be made higher up the chain, while in those places just the object should be passed.

bpoldrack: > why bother with this capture_exc @classmethod and not just to make it into __init__(self, exc, ...)?

Right. Original idea was somewhat different, but now - will do.

datalad/dochelpers.py
yarikoptic: as I have mentioned, this PR is IMHO a composition of two largely orthogonal changes:

- encapsulation of exception information in a handy `CapturedExcpetion` (just noted a typo -- should be `CapturedException`).  `exc_str` could just use it to provide formatting (thus no invocations or even logging output needs to change), and `return` records have its instance and add rendering of it etc
- changing output of `exc_str` used throughout codebase in logging to be more contextual (if there is desire/need; and thus may be more silent in some spots)

I think the first one would be easier to review if not clumped together with the 2nd one which might need discussion on its own.

bpoldrack: @yarikoptic The aim is not to change output of `exc_str`, but move decisions about output changes elsewhere. Please also see #5716 - I think the spots where `exc_str` is used are the wrong spots to make a decision about formatting. The introduction of this class is supposed to help change that. I could do it all at once to make that clear, but that'll be even harder to review. I made this PR to not have it all in one go. Separating further and only having `exc_str` account for a config or something is just inventing yet another intermediate step. I think that'll make it more likely to not get it finished above anything else.
The need for being more silent is discussed in #4478.

CONTRIBUTING.md
bpoldrack: Looking at the file, I really don't know what you're talking about. What do you want me to do?

datalad/support/exceptions.py
bpoldrack: Did remove `capture_exc`.

Re `%s` vs `%r`: Did that, too. However, regarding reducing the diff this is not gonna do much. All other places are status dicts where a string is needed (as opposed to something that can be converted to a string).

bpoldrack: > why to not chase all exc_str invocations

I *do* want to chase them, but with more in mind and therefore in a separate PR - see #5716.

CONTRIBUTING.md
yarikoptic: I was thinking about simple split into two lines:
```suggestion
  This flag is used by datalad to cap the number of traceback steps included in exception logging and
  result reporting to DATALAD_EXC_STR_TBLIMIT of pre-processed entries from traceback.
```

bpoldrack:
With #5716 and the repeated wondering about `str(exc_str)` vs `exc_str` ATM, I do want to make it clear:
I *want* to chase down all those invocations and they are two separate things wrt to a follow up, that I want to be able to easily distinguish and grep for in a follow-up PR. Whereever `exc_str` is used, an instance of `CapturedException` is passed onto the logger, which should be able to deal with it and I do want to see it working. Where `str(exc_str)` is used it's about a status/result dict that actually needs strings (but should ultimately get them in a different way - again see #5716).

bpoldrack:
Re failing crawler test:

At a quick look, seems to me `fasteners` changed (released three days ago): https://github.com/harlowja/fasteners/commit/ade69f8b3cf2b08b92ab21843bca69649a4ebef1

Ping @yarikoptic

yarikoptic:
> Re failing crawler test:
> 
> At a quick look, seems to me `fasteners` changed (released three days ago): [harlowja/fasteners@ade69f8](https://github.com/harlowja/fasteners/commit/ade69f8b3cf2b08b92ab21843bca69649a4ebef1)

yeap, filed #5717 and #5718 (but we might not need to rush since AFAIK 0.16.1 will be yanked)

bpoldrack:
Any feedback on my responses and push, @mih @yarikoptic ?

bpoldrack:
@mih @yarikoptic  Changed as promised in the call, rebased and added a change to `datalad test` so it would (for now) enable exception logging (just like it's currently needed for the CI runs).

mih (approved):
Thx, this works for me now! @yarikoptic unless you object, I will merge this soonish.

mih:
As the main request by @yarikoptic is addressed, and further work is coming in a separate PR, I will merge this now.

Issue #5228, opened 2020-12-10
How to install a nested dataset from gitlab


xarvshar:
Hi 

I am trying to install a nested dataset which was published to a gitlab remote earlier using *flat layout*. By doing a simple datalad install -r "path to git repo" it will not install the subdatasets data. 
It tries to install data from a below url

**- https:<somepath>/datalad/datalad_poc.git/leaf_ds**

where as the actual data will be present on 
**https://<somepath>/datalad/datalad_poc-leaf_ds.git**

Is there any flag where we can tell a datalad install command as what layout it has to use to extract data?

yarikoptic:
what is the content of `.gitmodules` there?  unless there are dark secrets there, actual commands invoked and values might help to assess situation ;)

xarvshar:
Hi @yarikoptic 

Thanks for your response ;)

Please find the below commands and contents which I used to implement this

Content of .gitmodules
**
[submodule "leaf_ds"]
        path = leaf_ds
        url = ./leaf_ds
        datalad-id = dfb16704-c403-45b2-b142-84ecd9e893a8
**

contents of python-gitlab.cfg file

 1 [mygit]
  2 url = https://gitlab.com/
  3 api_version = 4
  4 private_token = mytoken

script snippet to create GitLab remote sibling on the nested dataset 

**root_ds.config.set('datalad.gitlab-default-site', 'mygit')
root_ds.config.set('datalad.gitlab-mygit-siblingname', 'mygit')
root_ds.config.set('datalad.gitlab-mygit-layout', 'flat')
root_ds.config.set('datalad.gitlab-mygit-project', 'datalad/datalad_poc')
root_ds.save('Saving changes', recursive=True)
root_ds.create_sibling_gitlab(recursive=True,existing='reconfigure',publish_depends='gcsremote')** 

tree structure of the nested datalad dataset

![image](https://user-images.githubusercontent.com/16451564/101789260-16f74800-3b27-11eb-9e69-69dd5fd8e0bd.png)

Install command to install the nested dataset 

datalad install -r   <gitlab path where the root folder is published >

error what I am getting 

**[ERROR  ] Failed to clone from any candidate source URL. Encountered errors per each URL were:                      
| - https://gitlab.com/datalad/datalad_poc.git/leaf_ds
[...]

Issue #5033, opened 2020-10-14
install/get: ideally should add to candidates a URL based on super's 


yarikoptic:
If I clone/install a super-dataset, which has **only** immediate subdatasets installed, but not subdatasets (although available from their original location with "relative path URLs", `datalad install` in such a clone would fail since it nohow considers urls "built up" from the url present (only) in super dataset's .gitmodules

<details>
<summary>full dump from an example</summary> 

```shell
> git remote show origin
* remote origin
  Fetch URL: /tmp/tttt
  Push  URL: /tmp/tttt
  HEAD branch: master
  Remote branches:
    git-annex tracked
    master    tracked
  Local branch configured for 'git pull':
    master merges with remote master
  Local refs configured for 'git push':
    git-annex pushes to git-annex (fast-forwardable)
    master    pushes to master    (up to date)
(dev3) 1 29608 [1].....................................:Tue 13 Oct 2020 10:42:04 PM EDT:.
(git-annex)lena:/tmp/tttt-clone[master]
$> cat .gitmodules 
[submodule "haxby"]
	path = haxby
	url = smaug:/mnt/btrfs/datasets/datalad/crawl/labs/haxby
	datalad-id = 01cfb952-8439-11e6-af53-002590f97d84
[submodule "abide"]
	path = abide
	url = falkor.datalad.org:/srv/datasets.datalad.org/www/abide/
	datalad-id = 064527ea-92b3-11e7-8836-002590f97d84
(dev3) 1 29609 [1].....................................:Tue 13 Oct 2020 10:42:08 PM EDT:.
(git-annex)lena:/tmp/tttt-clone[master]
$> datalad install haxby
(dev3) 1 29610 [1].....................................:Tue 13 Oct 2020 10:42:25 PM EDT:.
(git-annex)lena:/tmp/tttt-clone[master]
$> ls haxby
total 16
4 attention/  4 dataset_description.json  4 life/  4 raiders/
(dev3) 1 29611 [1].....................................:Tue 13 Oct 2020 10:42:27 PM EDT:.
(git-annex)lena:/tmp/tttt-clone[master]
$> cat haxby/.gitmodules 
[submodule "raiders"]
	path = raiders
	url = ./raiders
[submodule "attention"]
	path = attention
	url = ./attention
[submodule "life"]
	path = life
	url = ./life
(dev3) 1 29612 [1].....................................:Tue 13 Oct 2020 10:42:49 PM EDT:.
[...]

Issue #4822, opened 2020-08-13
clone/install: introduce url mapping convenience?


yarikoptic:
#### What is the problem?

ATM for a user visiting some datalad-osf published dataset on osf.io it would require either user knowledge or every dataset having a description such as "this component can be git or datalad cloned from a 'osf://ID' URL, where 'ID' is the OSF node ID that shown in the OSF HTTP URL, e.g. https://osf.io/q8xnk/ can be cloned from osf://q8xnk", and then do all url tune up manually to get `datalad` `clone` or `install` working.

### Suggested solution:

I think datalad should make it easier and allow for a simple "copy paste URL I see in the browser bar", thus making `datalad clone https://osf.io/q8xnk/` work as well, while internally mapping it into `osf://q8xnk` for `git clone` operation.

Possible cons:
- if OSF changes their url schema -- would break our mapping.  But given historical stability and presence of those URLs in the wild already, it is very unlikely to happen

Possible implementation:
- datalad internally and also allowing extensions to register new "URL mappers"
- such "mappers" possibly should allow for different/absent context/purpose behavior
  - e.g. for OSF it would only matter for `clone`
  - in a somewhat related (waiting for merge) #4816, where within datalad itself we introduce "mapping" of `shub://` URLs into (minted by singularity hub with timeouts) http urls so that those images could be downloaded from the singularity hub; context would be `download`
- we already hardcode `///` to be a mapping to `http://datasets.datalad.org` -- implementation could generalize it
- some mappers could be as simple as regex based transformation (like in the case of OSF, `///`)

mih:
I think anything that is specific to a particular system should stay in the code that interfaces that system. I would consider it suboptimal, if datalad-core start accumulating mappings for all kinds of 3rd-party services, given that there is an extension to deal with OSF. Whatever mechanism will be implemented, it should be extensible/updatable without having to modify datalad-core.

Another use case for this is the RIA URL decoding that is presently implemented in `clone`.

yarikoptic:
Another usecase might be -- docker registries/hubs, see #5129

mih:
I am toying around with the idea to implement this through a config setting (and possibly some entrypoint that extensions can provide, in addition). It seems as if this could be as simple as

```
[datalad "url"]
	substitute = (.*)github.com/([^/]*)$\n\\1gitlab.com/\\2
	substitute = mike\nsue
```

so a multiply defined setting that separates match and substitution by a newline char (which should be rare in  a URL).

[...]

PR #5709, opened 2021-06-03
ENH+RF+BF:  github - sanitize top level reponame + make hierarchies datasets install/get-able from github

yarikoptic:
Please see the first commit which RFs and IMHO BFs our github repo name sanitization - it should address detected issue of us incorrectly assessing either top level repository is "existing" if its name was sanitized by github (or us) when it was created.

The second commit (details from commit message are pasted below) is the main purpose for this PR. The immediate use-case is to finally deploy https://github.com/con/tinuous/ to collect/archive CI logs for DataLad etc.  I think I will just publish them as hierarchies of the private datasets (one per month) under some dedicated organization (e.g. datalad-logs) so all team members could have access to them. But the `create-sibling-github -r` created hierarchies are not immediately available to `install -r` or `get [-r]` operations, hence this PR to address it.
 
    This change introduces special handling for github URLs.  Since we do
    have `create-sibling-github` I think it is Ok to also provide special
    handling for such dataset hierarchies created on github (by DataLad)
    within DataLad:
    
    - before this change we were using POSIX-style composition for all URL
      + path, even though we know that it would not work for GitHub
    
    - we do our "sanitization" of repository names we provide to GitHub
      upon create-sibling-github.  So we are the "central" place with
      that logic on sanitization (see previous commit -- our logic is
      different from github's)
    
    An alternative solution I considered:
    
    Add an option to `create-sibling-github` to populate full fledged urls
    within .gitmodules.  I've decided against it for a number of reasons:
    
    - per above reasoning -- we are already doing the wrong thing
      by trying POSIX-style composition on github urls, so it should
            go even if for the purpose of not wasting CPU/bandwidth
    
    - GitHub could be one of many remotes, thus it might not be
      desirable to "hard-code" the full url to github within .gitmodules
    
    - already created hierarchies would need fix ups got .gitmodules to
      provide full urls

With this PR I can do full round-trip like

```shell
#!/bin/bash

export PS4='> '
set -x
set -eu

datalad create "test  gh4"; 
cd "test  gh4"; 
datalad create -d . "sub _1"
datalad create -d . "sub _1/d/sub_sub- 1"

datalad create-sibling-github -s github --github-organization dl-throwaway --existing replace -r "test  gh4"
datalad push --to github -r;

datalad install -r -s git://github.com/dl-throwaway/test_gh4 ../dest
```

<details>
<summary>output of the interesting part</summary> 

```shell
> datalad create-sibling-github -s github --github-organization dl-throwaway --existing replace -r 'test  gh4'
[INFO   ] Successfully obtained information about organization dl-throwaway using token 765... 
.: github(-) [https://github.com/dl-throwaway/test_gh4.git (git)]
.: github(-) [https://github.com/dl-throwaway/test_gh4-sub__1.git (git)]
.: github(-) [https://github.com/dl-throwaway/test_gh4-sub__1-d-sub_sub-_1.git (git)]
'https://github.com/dl-throwaway/test_gh4.git' configured as sibling 'github' for Dataset('/home/yoh/.tmp/dl-g82eogS/test  gh4')
'https://github.com/dl-throwaway/test_gh4-sub__1.git' configured as sibling 'github' for Dataset('/home/yoh/.tmp/dl-g82eogS/test  gh4/sub _1')
'https://github.com/dl-throwaway/test_gh4-sub__1-d-sub_sub-_1.git' configured as sibling 'github' for Dataset('/home/yoh/.tmp/dl-g82eogS/test  gh4/sub _1/d/sub_sub- 1')
> datalad push --to github -r
publish(ok): sub _1/d/sub_sub- 1 (dataset) [refs/heads/master->github:refs/heads/master [new branch]]     
publish(ok): sub _1/d/sub_sub- 1 (dataset) [refs/heads/git-annex->github:refs/heads/git-annex [new branch]]
publish(ok): sub _1 (dataset) [refs/heads/master->github:refs/heads/master [new branch]]                  
publish(ok): sub _1 (dataset) [refs/heads/git-annex->github:refs/heads/git-annex [new branch]]            
publish(ok): . (dataset) [refs/heads/master->github:refs/heads/master [new branch]]                       
publish(ok): . (dataset) [refs/heads/git-annex->github:refs/heads/git-annex [new branch]]                 
> datalad install -r -s git://github.com/dl-throwaway/test_gh4 ../dest                                    
[INFO   ] Remote origin uses a protocol not supported by git-annex; setting annex-ignore                  
install(ok): /home/yoh/.tmp/dl-g82eogS/dest (dataset)
[INFO   ] Installing Dataset(/home/yoh/.tmp/dl-g82eogS/dest) to get /home/yoh/.tmp/dl-g82eogS/dest recursively 
[INFO   ] Remote origin uses a protocol not supported by git-annex; setting annex-ignore                  
install(ok): /home/yoh/.tmp/dl-g82eogS/dest/sub _1 (dataset)                                              
[INFO   ] Remote origin uses a protocol not supported by git-annex; setting annex-ignore                  
install(ok): /home/yoh/.tmp/dl-g82eogS/dest/sub _1/d/sub_sub- 1 (dataset)                                 
action summary:                                                                                           
  install (ok: 3)
bash gh-recursive  12.14s user 3.39s system 41% cpu 37.848 total

```
</details>

I am also ok to rebase this on  top of master (seems to be not merging cleanly) -- I think we should release `master` quite soon with its new WitlessRunner anyways ;-)

### DIFF
--- a/datalad/distribution/create_sibling_github.py
+++ b/datalad/distribution/create_sibling_github.py
@@ -44,9 +44,28 @@ from .siblings import Siblings
 
 lgr = logging.getLogger('datalad.distribution.create_sibling_github')
 
-# presently only implemented method to turn subdataset paths into Github
-# compliant repository name suffixes
+
+# Sanitization of a repository name
 template_fx = lambda x: re.sub(r'\s+', '_', re.sub(r'[/\\]+', '-', x))
+
+
+def _get_gh_reponame(reponame, sub=''):
+    """Return a name suitable for github, flattening via - for an optional sub(dataset).
+
+    We also do sanitization of the full name of the repository.  Divergence from
+    github behavior: it replaces spaces with "-" (so "test  _- whatever" becomes
+    "test-_--whatever.git"), and we replace with "_" and use "-" for joining different levels
+
+    Parameters
+    ----------
+    reponame: str
+      Name of the repository.  Must not contain any /, i.e. be just the name of the
+      repository and not "organization/name"
+    sub: str, optional
+      Name of the subdataset repository.
+    """
+    assert '/' not in reponame
+    return template_fx('-'.join(filter(bool, (reponame, sub))))
 
 
 @build_doc
@@ -200,10 +219,7 @@ class CreateSiblingGithub(Interface):
                         raise ValueError(msg)
                 elif existing == 'skip':
                     continue
-            gh_reponame = '{}{}{}'.format(
-                reponame,
-                '-' if mp else '',
-                template_fx(mp))
+            gh_reponame = _get_gh_reponame(reponame, mp)
             filtered.append((d, gh_reponame))
 
         if not filtered:
--- a/datalad/distribution/tests/test_install.py
+++ b/datalad/distribution/tests/test_install.py
@@ -946,7 +946,7 @@ def _create_test_install_recursive_github(path):  # pragma: no cover
 @with_tempfile(mkdir=True)
 def test_install_recursive_github(path):
     # test recursive installation of a hierarchy of datasets created on github
-    # using datalad create-sibling-github.  Following invocation was used to poplate it
+    # using datalad create-sibling-github.  Following invocation was used to populate it
     #
     # out = _create_test_install_recursive_github(path)
 
--- a/datalad/distribution/tests/test_install.py
+++ b/datalad/distribution/tests/test_install.py
@@ -921,3 +921,41 @@ def test_relpath_semantics(path):
         sub = install(
             dataset='super', source='subsrc', path=op.join('super', 'sub'))
         eq_(sub.path, op.join(super.path, 'sub'))
+
+
+def _create_test_install_recursive_github(path):  # pragma: no cover
+    # to be ran once to populate a hierarchy of test datasets on github
+    # Making it a full round-trip would require github credentials on CI etc
+    ds = create(opj(path, "testrepo  gh"))
+    # making them with spaces and - to ensure that we consistently use the mapping
+    # for create and for get/clone/install
+    ds.create("sub _1")
+    ds.create("sub _1/d/sub_-  1")
+    import datalad.distribution.create_sibling_github  # to bind API
+    ds.create_sibling_github(
+        "testrepo  gh",
+        github_organization='datalad',
+        recursive=True,
+        # yarik forgot to push first, "replace" is not working in non-interactive IIRC
+        # existing='reconfigure'
+    )
+    return ds.push(recursive=True, to='github')
+
+
+@skip_if_no_network
+@with_tempfile(mkdir=True)
+def test_install_recursive_github(path):
+    # test recursive installation of a hierarchy of datasets created on github
+    # using datalad create-sibling-github.  Following invocation was used to poplate it
+    #
+    # out = _create_test_install_recursive_github(path)
+
+    # "testrepo  gh" was mapped by our sanitization in create_sibling_github to testrepo_gh, thus
+    for i, url in enumerate([
+        'https://github.com/datalad/testrepo_gh',
+        # optionally made available to please paranoids, but with all takes too long (22sec)
+        #'https://github.com/datalad/testrepo_gh.git',
+        #'git@github.com:datalad/testrepo_gh.git',
+    ]):
+        ds = install(source=url, path=opj(path, "clone%i" % i), recursive=True)
+        eq_(len(ds.subdatasets(recursive=True, fulfilled=True)), 2)
--- a/datalad/distribution/utils.py
+++ b/datalad/distribution/utils.py
@@ -17,10 +17,15 @@ from os.path import isabs
 from os.path import normpath
 import posixpath
 
-from datalad.support.network import DataLadRI
-from datalad.support.network import URL
-from datalad.support.network import RI
-from datalad.support.network import PathRI
+from datalad.support.network import (
+    PathRI,
+    RI,
+    URL,
+    is_url,
+    urlunquote,
+)
+
+from .create_sibling_github import _get_gh_reponame
 
 
 lgr = logging.getLogger('datalad.distribution.utils')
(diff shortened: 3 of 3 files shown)


yarikoptic:
Echoing our debate today, I wonder if I should add `--set-upstream` option (or should it become `--set-upstream branch,gitmodules,all` or alike?) which would instruct to make github to be the default upstream remote and adjust .gitmodules `url` to point to github (instead of just keeping the `path`).  That would make those indeed more kosher for github.  I do not think it should be the default behavior, so it doesn't anyhow eliminate the need of this (or alike) PR. WDYT @mih @bpoldrack et al?

mih:
I am familiarizing myself with the approach we have implemented, but already considering the proposed update, ie. working on the branch of this PR. I am using my clone of the main code repo, and I have added a dummy subdataset. If I try to get it, I see these URLs being assembled:

```
{
  'cost': 500,
  'name': 'gh-yarik',
  'url': 'git@github.com:yarikoptic/datalad-subdstest.git'},
{
  'cost': 600,
  'name': 'gh-yarik',
  'url': 'git@github.com:datalad-datasets/kdenlive-what-is-datalad.git'},
{
  'cost': 900,
  'name': 'local',
  'url': 'git@github.com:datalad-datasets/kdenlive-what-is-datalad.git'},
{
  'cost': 590,
  'name': 'dl-url',
  'url': 'git@github.com:datalad-datasets/kdenlive-what-is-datalad.git'}
```

After sorting and unique'ing only two remain:

```
{
  'cost': 500,
  'name': 'gh-yarik',
  'url': 'git@github.com:yarikoptic/datalad-subdstest.git'},
{
  'cost': 590,
  'name': 'dl-url',
  'url': 'git@github.com:datalad-datasets/kdenlive-what-is-datalad.git'}
```

The first URL does not work, it succeeds on the second attempt.

(I will continue with my attempt to refresh my mind)

mih:
Trying the test case for this PR:

```
% datalad clone https://github.com/datalad/testrepo_gh
```

The final list of candidates is

```
{
  'cost': 500,
  'name': 'origin',
  'url': 'https://github.com/datalad/testrepo_gh-sub__1'},
{
  'cost': 600,
  'name': 'origin',
  'url': 'https://github.com/datalad/testrepo_gh-.-sub__1'}
```

The first one works (as expected), the second makes sense if one mentally undoes the github-specific normalization introduced by this PR.

mih:
> Echoing our debate today, I wonder if I should add `--set-upstream` option (or should it become `--set-upstream branch,gitmodules,all` or alike?) which would instruct to make github to be the default upstream remote and adjust .gitmodules `url` to point to github (instead of just keeping the `path`).  That would make those indeed more kosher for github.  I do not think it should be the default behavior, so it doesn't anyhow eliminate the need of this (or alike) PR. WDYT @mih @bpoldrack et al?

This makes sense to me.

yarikoptic:
For tuning `.gitmodules` -- filed https://github.com/datalad/datalad/issues/5720 . For "flat hiearchies", ran into a related https://github.com/datalad/datalad/issues/5228 where the desire was also for gitlab.  Use-case slightly different since for gitlab we do not have a dedicated `sanitizer` as we do for github, so it would not be the same handler.

bpoldrack:
> Echoing our debate today, I wonder if I should add --set-upstream option (or should it become --set-upstream branch,gitmodules,all or alike?) which would instruct to make github to be the default upstream remote and adjust .gitmodules url to point to github (instead of just keeping the path). That would make those indeed more kosher for github. I do not think it should be the default behavior, so it doesn't anyhow eliminate the need of this (or alike) PR. WDYT @mih @bpoldrack et al?

Sounds reasonable. Leaning towards an optional value for that option, so - `--set-upstream [branch | gitmodules | all]` where `branch` is default (that is - *if* `--set-upstream` is given at all, of course). That's because I find `all`/`gitmodules` surprising as a default.

yarikoptic:
@bpoldrack - "moved" your comment to https://github.com/datalad/datalad/issues/5720#issuecomment-857662701 since implementation of that option should be IMHO at `siblings` level and thus IMHO OOS (out of scope) for this PR.

yarikoptic:
@mih could you please elaborate on your [comments above](https://github.com/datalad/datalad/pull/5709#issuecomment-856832397) on either you are just confirming that this PR already works as you expected, or that you want some changes on top of it? if changes -- which ones?

mih:
Here is a full alternative diff that also achieves successful installation of the test case, with no initial failed attempts:

```diff
diff --git a/datalad/distribution/get.py b/datalad/distribution/get.py
index b43d786d4..747a66b32 100644
--- a/datalad/distribution/get.py
+++ b/datalad/distribution/get.py
@@ -202,6 +202,9 @@ def _get_flexible_source_candidates_for_submodule(ds, sm):
         for k, v in sm.items()
         if k.startswith('gitmodule_')
     }
+    # add prop matching the name component that create-sibling-github would generate
+    from datalad.distribution.create_sibling_github import normalize_reponame
+    sm_candidate_props['github_name'] = normalize_reponame(sm_path)
 
     for remote in unique(candidate_remotes):
         remote_url = ds_repo.get_remote_url(remote, push=False)
```

Call then requires config to be set (e.g. by `create-sibling-github`, or user config, or a future catalog of substitutions):

```sh
datalad \
  -c datalad.get.subdataset-source-candidate-100usethis={remoteurl-origin}-{github_name} \
  get -n -r .
```

yarikoptic:
> Here is a full alternative diff that also achieves successful installation of the test case, with no initial failed attempts ...

This is nice!  I think we should indeed add `github_name` into `_props`!   It would allow to address #5228 use case (assuming that the same sanitization logic was used for gitlab repos as well).

It is not really a `"full alternative diff"` though since 
- IMHO we must not require users to give such an elaborate `-c` for a mere `install`/`clone`-ing created with `create-sibling-github` hierarchies on github.  IMHO that would not be any good UX! To resolve that
  - the "full diff" would require making `create-sibling-github` adjusting `.datalad/config` for the repos it to share on github.
    - ... and thus introduce a diff, so can't be used to just reshare clones on github
  - nobody would be able to `clone`/`install` already populated on github hierarchies without such elaborate `-c` 
  - priorities are nice etc, but those "break" default git's logic on taking the super-ds's location as the one to consider first. But that what would happen if hard-coded high priority setting would be in `.datalad/config`, thus possibly affecting installation from local clones of the hierarchy which would then first consider github instead of just a local location
- IMHO we should still stop even trying to clone from github URLs we know we do not support - we know they would not work.

bpoldrack:
> To resolve that the "full diff" would require making   ...

There's an option missing. Ultimately it's all about subdatasets. The superdataset you can clone just fine from Github. We can simply have another `postclone_cfg` that - in case we just cloned from Github - does set that config locally in order to consider subdatasets from that location and be done. Would work with existing ones and w/o active user intervention at any point. IIRC we already do that for RIA stores.
Also solves the recursive case, since by the time a clone on a subdataset is called, the config would be present.

> IMHO we should still stop even trying to clone from github URLs we know we do not support - we know they would not work.

I have nothing against skipping URL guesses that we know can't possibly work. I just wouldn't want to introduce a special case guessing of possibly existing alternatives on top of that.

yarikoptic:
> simply have another `postclone_cfg` that - in case we just cloned from Github - does set that config locally in order to consider subdatasets from that location and be done. Would work with existing ones and w/o active user intervention at any point. IIRC we already do that for RIA stores.

sounds interesting! but what if superdataset was cloned already and I am just `install`/`get`-ing a subdataset -- if it is `postclone_cfg`, sounds like it would not apply.

yarikoptic:
> > IMHO we should still stop even trying to clone from github URLs we know we do not support - we know they would not work.
> 
> I have nothing against skipping URL guesses that we know can't possibly work. I just wouldn't want to introduce a special case guessing of possibly existing alternatives on top of that.

IMHO we can't avoid "special case"ing entirely, it is just a matter how/where we can do it. This PR places it directly at the point of submodule URL construction, which as is it is now, we know would not work for github (or any other portal with flattened hierarchy for that matter).  If we could do that somewhere in higher logic (as e.g. it sounds was done for RIA) - I don't mind moving it there, but then may be the entire submodule URL construction logic might need to go there as well.

mih:
FTR: I am presently looking into https://github.com/datalad/datalad/issues/5033 -- which is closely related to the generation of adequate candidates.

mih:
Here is another issue that is somewhat related https://github.com/datalad/datalad/issues/4822 It talks about taking a URL, figuring out that it actually should be another URL, and mapping it.

yarikoptic:
what is your feeling @mih on this PR -- do you see some better approach which would replace it?  My feeling that this this PR we would only be better off -- it is an internal change which would fix the broken use-case, doesn't require any chance on user-side, and later we could RF to take advantage if applicable of any rewriting we might add while also accounting for this use-case.

mih:
> assuming that the same sanitization logic was used for gitlab repos as well

I am looking into this, but it is painful #5761 #5763 #5764

mih:
Ok, I think I found an approach to implement this without the need to configure individual dataset, or change the code with service-specific conditionals. Stay tuned. I will in all likelihood adjust #5749 to make this happen.

mih:
I have an implementation now. Will push later on -- needs a change of docs.

yarikoptic:
I believe this PR is largely (still can do useless tries on 3rd level from github) replaced by https://github.com/datalad/datalad/pull/5749 so I will close this one.  Note, that #5749 was against master since adding new functionality etc, and not `maint` like this one, so will be present only for 0.15.0

PR #5718, opened 2021-06-07
BF: do not use .acquired - just get state from acquire()

### LINKED ISSUE #5717: fasteners 0.16.1 removed .acquired and thus broke datalad
complained upstream https://github.com/harlowja/fasteners/issues/71

spotted while observing avalanche of CI failures for https://github.com/datalad/datalad/pull/5709 which should have been green.

crashes look like this
```
ERROR: datalad.tests.test_archives.test_ExtractedArchive
  File "/tmp/dl-miniconda-1newyg4m/lib/python3.8/site-packages/datalad/support/locking.py", line 91, in lock_if_check_fails
    lgr.debug("Acquired? lock %s: %s", lock_filename, lock.acquired)
AttributeError: '_FcntlLock' object has no attribute 'acquired'
  File "/tmp/dl-miniconda-1newyg4m/lib/python3.8/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/tmp/dl-miniconda-1newyg4m/lib/python3.8/site-packages/datalad/tests/utils.py", line 585, in _wrap_with_tree
    return t(*(arg + (d,)), **kw)
  File "/tmp/dl-miniconda-1newyg4m/lib/python3.8/site-packages/datalad/tests/test_archives.py", line 192, in test_ExtractedArchive
    extracted_ = earchive.get_extracted_file(fpath)
  File "/tmp/dl-miniconda-1newyg4m/lib/python3.8/site-packages/datalad/support/archives.py", line 427, in get_extracted_file
    self.assure_extracted()
  File "/tmp/dl-miniconda-1newyg4m/lib/python3.8/site-packages/datalad/support/archives.py", line 315, in assure_extracted
    with lock_if_check_fails(
  File "/tmp/dl-miniconda-1newyg4m/lib/python3.8/contextlib.py", line 113, in __enter__
    return next(self.gen)
  File "/tmp/dl-miniconda-1newyg4m/lib/python3.8/site-packages/datalad/support/locking.py", line 100, in lock_if_check_fails
    if lock.acquired:
AttributeError: '_FcntlLock' object has no attribute 'acquired'
```

we could have added upper bound for fasteners version, but I guess the best would be to look into removing use of `.acquired` -- will look into it now.

yarikoptic:
Somewhat ad-hoc and makes an already confusing lock_if_check_fails even more so
by optionally changing its return signature. So eventually we might want
to RF this whole piece of code.  But I think it should work ok for now.
Anyways -- fasteners 0.16.x series might re-gain .acquired (see
https://github.com/harlowja/fasteners/issues/71) and we might want to see to RF
this PR properly if so instead of merging/releasing it right away.  But FWIW
Closes #5717

### DIFF
--- a/datalad/support/locking.py
+++ b/datalad/support/locking.py
@@ -30,7 +30,7 @@ def lock_if_check_fails(
     lock_path,
     operation=None,
     blocking=True,
-    return_acquired=False,
+    _return_acquired=False,
     **kwargs
 ):
     """A context manager to establish a lock conditionally on result of a check
@@ -66,8 +66,9 @@ def lock_if_check_fails(
     blocking: bool, optional
       If blocking, process would be blocked until acquired and verified that it
       was acquired after it gets the lock
-    return_acquired: bool, optional
-      Return also if lock was acquired
+    _return_acquired: bool, optional
+      Return also if lock was acquired.  For "private" use within DataLad (tests),
+      do not rely on it in 3rd party solutions.
     **kwargs
       Passed to `.acquire` of the fasteners.InterProcessLock
 
@@ -97,7 +98,7 @@ def lock_if_check_fails(
             assert acquired
         check2 = _get(check)
         ret_lock = None if check2 else lock
-        if return_acquired:
+        if _return_acquired:
             yield check2, ret_lock, acquired
         else:
             yield check2, ret_lock
--- a/datalad/support/locking.py
+++ b/datalad/support/locking.py
@@ -30,6 +30,7 @@ def lock_if_check_fails(
     lock_path,
     operation=None,
     blocking=True,
+    return_acquired=False,
     **kwargs
 ):
     """A context manager to establish a lock conditionally on result of a check
@@ -65,12 +66,14 @@ def lock_if_check_fails(
     blocking: bool, optional
       If blocking, process would be blocked until acquired and verified that it
       was acquired after it gets the lock
+    return_acquired: bool, optional
+      Return also if lock was acquired
     **kwargs
       Passed to `.acquire` of the fasteners.InterProcessLock
 
     Returns
     -------
-    result of check, lock
+    result of check, lock[, acquired]
     """
     check1 = _get(check)
     if check1:  # we are done - nothing to do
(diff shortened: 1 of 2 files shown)


bpoldrack:
Since it seems like fasteners will not only regain the attribute but actually remove the release from pypi, we might want to give them 24h or so, before reacting with a quick fix, I think (much longer doesn't really work with entire CI broken).

yarikoptic:
0.16.2 is released, tested locally that our "locking" tests now back to normal.  But I think we should still proceed with this PR, even if without major rush ;-)  As mentioned in https://github.com/harlowja/fasteners/issues/71#issuecomment-856094384 it is pretty much not advised to rely on it, and our code can be easily RFed as done here to not.  It is only the tests are kinda troublesome and I forgot all this locking kitchen to possibly find a better way to fix up the tests without changing the API.

bpoldrack:
> I forgot all this locking kitchen to possibly find a better way to fix up the tests without changing the API.

Never really looked into it, but it seems to me, that the solution is not to not return `aquired`, but get that information in a different way (not using this particular `.acquired`). Quick look is suggesting to me that something like `.lockfile is not None` will do as a replacement.

yarikoptic:
> Never really looked into it, but it seems to me, that the solution is not to not return `aquired`, but get that information in a different way (not using this particular `.acquired`). Quick look is suggesting to me that something like `.lockfile is not None` will do as a replacement.

IIRC I had considered that option too but for some reason decided to just return... Looking at the code:

```python
        self._do_open()
        watch = _utils.StopWatch(duration=timeout)
        r = _utils.Retry(delay, max_delay,
                         sleep_func=self.sleep_func, watch=watch)
        with watch:
            gotten = r(self._try_acquire, blocking, watch)
        if not gotten:
            self.acquired = False
            return False
        else:
            self.acquired = True
            self.logger.log(_utils.BLATHER,
                            "Acquired file lock `%s` after waiting %0.3fs [%s"
                            " attempts were required]", self.path,
                            watch.elapsed(), r.attempts)
            return True
```
where `.lockfile` is set/opened in `_do_open`, and then tried to get locked in `_try_acquire` - it seems that it is not guaranteed that .lockfile would be `None` if `acquired = False`.   That is why I thought that the cleanest way is to somehow bubble up the `acquired` bool.

bpoldrack:
Yeah. So - just store the return value of `acquire()` in `lock_if_check_fails`?

yarikoptic:
> Yeah. So - just store the return value of `acquire()` in `lock_if_check_fails`?

well, I already store it in a variable now in that function: https://github.com/datalad/datalad/pull/5718/files#diff-c4170b6a46f8d90d2bdcb206966fe9f16e8a431395245d39434edb681ab03ca6R94 . The question is how to channel back to the test on either lock was acquired or not (if not blocking). Having slept on it, I will just document that the `_return_acquired` is "private" and should not be used by 3rd party code, and be done with this one. pushed now - I still think we should go with this RFing to avoid future breakage-by-fasteners

PR #5722, opened 2021-06-08
WiP Q&D AnnexRepo.add_ batched (by default) if more than a single file passed

### LINKED ISSUE #5721: Use annex add --batch  instead of a sequence of `annex add ...chunk-of-files`
Not yet sure if it would provide a remedy, but I have attributed a [fresh performance issue of git-annex](https://git-annex.branchable.com/bugs/significant_performance_regression_impacting_datal/#comment-4df4242f85746baef8afd32e0632c37c) (manifesting on OSX... may be shorter max cli size triggering it, didn't check) to a growth in how long `annex add` takes when we chop up our long list of files into separate `annex add` calls.  I think (didn't time) it should be faster if we just use a single `annex add --batch` process there.  This also reflects on  https://github.com/datalad/datalad/issues/4630#issuecomment-759854665 .  Given those observations, switching to `annex add --batch` could provide a notable performance benefit.

yarikoptic:
If ever finished, Closes #5721 

If to proceed
- ideally some hero of the Runners, like @christian-monch  , would just RF/fold that BatchedAnnexes deep into the guts of `_call_annex*` so reduce duplication etc, but it shouldn't be the goal of this PR
- [x] see if any positive effect on OSX with bleeding edge annex in relation to [recent regression on our tests on OSX](https://git-annex.branchable.com/bugs/significant_performance_regression_impacting_datal/#comment-4df4242f85746baef8afd32e0632c37c)
   - yeap, seems to mitigate it well, see the git-annex issue above for fresh comments on that.
- [x] no effect on benchmarks... if anything only `0.92  plugins.addurls.addurls1.time_addurls('*')` is happier 
- [ ] see what fails and either it could be addressed here or there is some roadblock
- [ ] I think added `batch=None` logic handling is sound (taken from `get_file_key`), so I would just keep it but could be improved to just reuse a batched `add` if already one running.
- [ ] interface that `total_nbytes` (I did not even look into what it is about)

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -895,23 +895,7 @@ class AnnexRepo(GitRepo, RepoInterface):
         cmd.append('annex')
         cmd += args
 
-        if lgr.getEffectiveLevel() <= 8:
-            cmd.append('--debug')
-
-        if self._annex_common_options:
-            cmd += self._annex_common_options
-
-        if jobs == 'auto':
-            # Limit to # of CPUs (but at least 3 to start with)
-            # and also an additional config constraint (by default 1
-            # due to https://github.com/datalad/datalad/issues/4404)
-            jobs = self._n_auto_jobs or min(
-                self.config.obtain('datalad.runtime.max-annex-jobs'),
-                max(3, cpu_count()))
-            # cache result to avoid repeated calls to cpu_count()
-            self._n_auto_jobs = jobs
-        if jobs and jobs != 1:
-            cmd.append('-J%d' % jobs)
+        cmd += self._get_annex_options(jobs)
 
         runner = self._git_runner
         env = None
@@ -970,6 +954,27 @@ class AnnexRepo(GitRepo, RepoInterface):
 
             # we don't know how to handle this, just pass it on
             raise
+
+    def _get_annex_options(self, jobs):
+        """Common logic to populate annex call options across our spectrum of annex invocations
+        """
+        cmd = []
+        if lgr.getEffectiveLevel() <= 8:
+            cmd.append('--debug')
+        if self._annex_common_options:
+            cmd += self._annex_common_options
+        if jobs == 'auto':
+            # Limit to # of CPUs (but at least 3 to start with)
+            # and also an additional config constraint (by default 1
+            # due to https://github.com/datalad/datalad/issues/4404)
+            jobs = self._n_auto_jobs or min(
+                self.config.obtain('datalad.runtime.max-annex-jobs'),
+                max(3, cpu_count()))
+            # cache result to avoid repeated calls to cpu_count()
+            self._n_auto_jobs = jobs
+        if jobs and jobs != 1:
+            cmd.append('-J%d' % jobs)
+        return cmd
 
     def _call_annex_records(self, args, files=None, jobs=None,
                             git_options=None,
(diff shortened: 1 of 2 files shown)


bpoldrack:
FTR: Approach looks good to me ;-)

mih:
This PR has not seen an update re the given TODOs in several months. At this point it is unclear to me how it would relate to the related work in #6125 -- which seems to be nearing completion. I propose to close this for now. WDYT @yarikoptic

datalad/support/annexrepo.py
```diff
@@ -1604,12 +1608,29 @@ def _get_to_be_added_recs(paths):
         else:
             if backend:
                 options.extend(('--backend', backend))
-            for r in self._call_annex_records(
+            yield from self._maybe_batched_add(
+                files, options, jobs, expected_additions, batch=None)
+
+    def _maybe_batched_add(self, files, options, jobs, expected_additions, batch=None):
+        # Common code to be reused across existing implementations of "annex add" invocation
+        if batch or (batch is None and len(files) > 1):
```
christian-monch, line 1616: This means that the else-path is executed if, for example, `b == ""` and `len(files) > 1`. I guess that is intended?

mih:

datalad/support/annexrepo.py
yarikoptic: well, `batch` should be a bool, so `""` would be taken as `False` and thus going into `else` since "explicitly" instructed to not batch. So IMHO that was ok.

yarikoptic:
FTR, fails seems to be just "few" which is great

```
======================================================================^M
FAIL: datalad.support.tests.test_annexrepo.test_annexjson_protocol^M
----------------------------------------------------------------------^M
Traceback (most recent call last):^M
  File "/tmp/dl-miniconda-wva2dikc/lib/python3.8/site-packages/nose/case.py", line 198, in runTest^M
    self.test(*self.arg)^M
  File "/tmp/dl-miniconda-wva2dikc/lib/python3.8/site-packages/datalad/tests/utils.py", line 560, in _wrap_with_tree^M
    return t(*(arg + (d,)), **kw)^M
  File "/tmp/dl-miniconda-wva2dikc/lib/python3.8/site-packages/datalad/support/tests/test_annexrepo.py", line 2176, in test_annexjson_protocol^M
    assert_repo_status(path)^M
  File "/tmp/dl-miniconda-wva2dikc/lib/python3.8/site-packages/datalad/tests/utils.py", line 1619, in assert_repo_status^M
    eq_(state_files, oktobefound,^M
AssertionError: unexpected content of state "untracked": [PosixPath('/tmp/datalad_temp_tree_test_annexjson_protocolii_p7wb4/dir1')] != []^M
^M
======================================================================^M
FAIL: datalad.support.tests.test_annexrepo.test_find_batch_equivalence^M
----------------------------------------------------------------------^M
Traceback (most recent call last):^M
  File "/tmp/dl-miniconda-wva2dikc/lib/python3.8/site-packages/nose/case.py", line 198, in runTest^M
    self.test(*self.arg)^M
  File "/tmp/dl-miniconda-wva2dikc/lib/python3.8/site-packages/datalad/tests/utils.py", line 560, in _wrap_with_tree^M
    return t(*(arg + (d,)), **kw)^M
  File "/tmp/dl-miniconda-wva2dikc/lib/python3.8/site-packages/datalad/support/tests/test_annexrepo.py", line 551, in test_find_batch_equivalence^M
    eq_(set(ar.find(["subdir"])["subdir"]), {"subdir/d", "subdir/e"})^M
AssertionError: set() != {'subdir/e', 'subdir/d'}^M
```
and might be worth resurrecting (I will try quickly now...)

PR #5723, opened 2021-06-09
ENH: addreadme: improve default template

loj:
This improves the default template for `add-readme` using the content that @adswa created for the [DataLad Handbook](http://handbook.datalad.org/en/latest/basics/101-180-FAQ.html#how-can-i-help-others-get-started-with-a-shared-dataset).

Closes #5714

PR #5725, opened 2021-06-09
Do not annex the README that add-readme generates

### LINKED ISSUE #5713: `add-readme` should put the README into git (by default)
ATM it depends on prior config to do the right thing -- I cannot think of a situation where a README should be annex'ed by default.

adswa:
This PR sits on top of @loj's #5723, and configures the dataset to not annex the README, unless a user overrides this default with an ``--annex`` flag.
With this, it fixes #5713. The previous behavior (having a README file ~placed in Git~ annexed) can be achieved by running ``datalad add-readme --annex``.

### DIFF
--- a/datalad/local/add_readme.py
+++ b/datalad/local/add_readme.py
@@ -52,19 +52,12 @@ class AddReadme(Interface):
             'skip': do nothing; 'append': append information to the existing
             file; 'replace': replace the existing file with new content.""",
             constraints=EnsureChoice("skip", "append", "replace")),
-        annex=Parameter(
-            args=("--annex",),
-            action='store_true',
-            default=False,
-            doc="""By default, the dataset will be configured to store a newly
-            created README in Git. If this parameter is given, the README will
-            be annexed."""),
     )
 
     @staticmethod
     @datasetmethod(name='add_readme')
     @eval_results
-    def __call__(dataset, filename='README.md', existing='skip', annex=False):
+    def __call__(dataset, filename='README.md', existing='skip'):
         from os.path import lexists
         from os.path import join as opj
         from io import open
@@ -77,10 +70,10 @@ class AddReadme(Interface):
         dataset = require_dataset(dataset, check_installed=True,
                                   purpose='add README')
 
-        filename = opj(dataset.path, filename)
-        res_kwargs = dict(action='add_readme', path=filename)
+        fpath = opj(dataset.path, filename)
+        res_kwargs = dict(action='add_readme', path=fpath)
 
-        if lexists(filename) and existing == 'skip':
+        if lexists(fpath) and existing == 'skip':
             yield dict(
                 res_kwargs,
                 status='notneeded',
@@ -88,31 +81,20 @@ class AddReadme(Interface):
             return
 
         # unlock, file could be annexed
-        if lexists(filename):
-            dataset.unlock(filename)
-        if not lexists(filename):
-            if isinstance(dataset.repo, AnnexRepo):
-                # if we have an annex repo, shall the README go to Git or annex?
-                attr_state = dataset.status(
-                        '.gitattributes',
-                        result_renderer='disabled',
-                        return_type='item')
-                if attr_state:
-                    # we have made sure that .gitattributes exists. We only fail
-                    # now if the state of the .gitattributes file is not clean
-                    if not attr_state[0].get('state') == 'clean':
-                        raise RuntimeError(
-                            'Stopping, as the .gitattributes file is modified'
-                            )
+        if lexists(fpath):
+            dataset.unlock(fpath)
+        if not lexists(fpath):
+            # if we have an annex repo, shall the README go to Git or annex?
 
-                # unless --annex is set, configure the README to go into Git
+            if isinstance(dataset.repo, AnnexRepo) \
+                and 'annex.largefiles' not in \
+                    dataset.repo.get_gitattributes(filename).get(filename, {}):
+                # configure the README to go into Git
                 dataset.repo.set_gitattributes(
-                    [('README.md', {'annex.largefiles': 'anything' if annex
-                                                        else 'nothing'})])
+                    [(filename, {'annex.largefiles': 'nothing'})])
                 dataset.save(
                     path='.gitattributes',
-                    message="[DATALAD] Configure README to be in {}".format(
-                        "annex" if annex else "Git"),
+                    message="[DATALAD] Configure README to be in Git",
                     to_git=True
                 )
 
@@ -236,16 +218,16 @@ files by whom, and when.
             id=u' (id: {})'.format(dataset.id) if dataset.id else '',
             )
 
-        with open(filename, 'a' if existing == 'append' else 'w', encoding='utf-8') as fp:
+        with open(fpath, 'a' if existing == 'append' else 'w', encoding='utf-8') as fp:
             fp.write(default_content)
             yield dict(
                 status='ok',
-                path=filename,
+                path=fpath,
                 type='file',
                 action='add_readme')
 
         for r in dataset.save(
-                filename,
+                fpath,
                 message='[DATALAD] added README',
                 result_filter=None,
                 result_xfm=None):
--- a/datalad/local/add_readme.py
+++ b/datalad/local/add_readme.py
@@ -13,6 +13,7 @@ __docformat__ = 'restructuredtext'
 
 from datalad.interface.base import Interface
 from datalad.interface.base import build_doc
+from datalad.support.annexrepo import AnnexRepo
 
 
 @build_doc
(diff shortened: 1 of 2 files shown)


bpoldrack:
> The previous behavior (having a README file placed in Git) can be achieved by running datalad add-readme --annex.

You mean place it in annex, I assume? ;-)

adswa:
> > The previous behavior (having a README file placed in Git) can be achieved by running datalad add-readme --annex.
> 
> You mean place it in annex, I assume? ;-)

whoops, yes. :P

datalad/local/add_readme.py
```diff
@@ -82,6 +89,20 @@ def __call__(dataset, filename='README.md', existing='skip'):
         # unlock, file could be annexed
         if lexists(filename):
             dataset.unlock(filename)
+        if not lexists(filename) and not annex:
```
bpoldrack, line 92: Actually, you're not doing what the doc reads: `--annex` doesn't make sure it's annexed (while not giving it *does* make sure it's going into git). Instead, with `--annex` what happens is, whatever the existing configuration of the datasets implies. That's not exactly the same thing.

adswa: That is true! Do you have an idea for a better parameter name?

adswa: ``--honor-config``?

bpoldrack: Well, honoring the config should be expected default anyway. Would be strange to have that as an option. Implementation just needs an `else` where we record the opposite in `.gitattributes` explicitly.

bpoldrack: Or: Don't make it an `else` really, but have the block `if not lexists(filename):` and make the value for `annex.alrgefiles` conditional:

` ... {'annex.largefiles': 'anything' if annex else 'nothing'} ...`

adswa: ok, I have implemented this. Weirdest .gitattribute configuration I have written so far :D

datalad/local/add_readme.py
```diff
@@ -82,6 +89,23 @@ def __call__(dataset, filename='README.md', existing='skip'):
         # unlock, file could be annexed
         if lexists(filename):
             dataset.unlock(filename)
+        if not lexists(filename):
+            # if it does not exists, make sure it is saved into Git
```
bpoldrack, line 93: Well, now the comment is kinda outdated in this spot. :-P

bpoldrack (approved):
Apart from the now not really fitting comment - LGTM.

datalad/local/add_readme.py
```diff
@@ -82,6 +89,23 @@ def __call__(dataset, filename='README.md', existing='skip'):
...
+                    'Stopping, because the .gitattributes file is modified'
+                    )
+            dataset.repo.set_gitattributes(
+                [('README.md', {'annex.largefiles': 'anything' if annex
+                                                    else 'nothing'})])
+            dataset.save(
+                path='.gitattributes',
+                message="[DATALAD] Configure README to be in {}".format(
+                    "annex" if annex else "Git")
+            )
```
yarikoptic, line 108: might be worth conditioning this block on `isinstance(ds.repo, AnnexRepo)` to not pester git-puritans with all the git-annex goodness (docstring might need to be adjusted to state that the option is in effect only for annex'ed repos, which is due anyways since if repo is a pure GitRepo we will not save to annex even if annex=True)

datalad/local/add_readme.py
adswa: haha, fixed ;-)

datalad/local/add_readme.py
adswa: good idea - done.

datalad/local/add_readme.py
```diff
@@ -82,6 +90,25 @@ def __call__(dataset, filename='README.md', existing='skip'):
...
+                if dataset.status(
+                        '.gitattributes',
+                        result_renderer='disabled',
+                        return_type='list',)[0]['state'] != 'clean':
+                    raise RuntimeError(
+                        'Stopping, because the .gitattributes file is modified'
+                        )
+                # unless --annex is set, configure the README to go into Git
+                dataset.repo.set_gitattributes(
+                    [('README.md', {'annex.largefiles': 'anything' if annex
```
yarikoptic, line 105: Well, you want to do this even if .gitattributes doesn't exist - datalad does create it to force md5e backend, but a pure annex repo might not have it

yarikoptic: Overall - why to condition on. gitattributes existence? We can have it or not, but we want to enforce this fine too go to git or annex

adswa: > Overall - why to condition on. gitattributes existence? We can have it or not, but we want to enforce this fine too go to git or annex

mhh, maybe I misunderstood what ``set_gitattributes`` does - does it add a condition on the existence of this file? I thought it rather creates one if it can't find it?

adswa: (or maybe I misunderstood your second comment?)

bpoldrack: Missed that. The issue is, that `status` on a not existing path would return an empty list. That's an indirect conditioning, since the call will fail with `index out of range` (your `[0]`). Don't see another kind of conditioning here.

adswa: Ah, I see, thanks! But some check would need to be in place to make sure that there are no uncommitted changes in gitattributes before applying and saving the configuration for the README, right?

bpoldrack: So, safeguard needed. First get the status, and then check it only if something was returned. Otherwise there is no `.gitattributes` and you can safely add as well.

On that note another nitpick:
`['state'] != 'clean'` doesn't exactly imply "modified". Could be untracked. Probabyl to the same consequence - we don't exactly know whether it's ok to add something.

bpoldrack: > But some check would need to be in place to make sure that there are no uncommitted changes in gitattributes before applying and saving the configuration for the README, right?

Right. Just get the status in variable and do the check on `clean` only if there was a result. If not - all good, go ahead and write to it.

datalad/local/add_readme.py
```diff
@@ -82,6 +90,31 @@ def __call__(dataset, filename='README.md', existing='skip'):
...
+                attr_state = dataset.status(
+                        '.gitattributes',
+                        result_renderer='disabled',
+                        return_type='item')
+                if attr_state:
+                    # we have made sure that .gitattributes exists. We only fail
+                    # now if the state of the .gitattributes file is not clean
+                    if not attr_state[0].get('state') == 'clean':
+                        raise RuntimeError(
+                            'Stopping, as the .gitattributes file is modified'
```
bpoldrack, line 105:
```suggestion
                            'Stopping, as the .gitattributes file is %s', attr_state[0].get('state')
```

bpoldrack: ^ Nitpicky, but while at it ... would become another UX issue eventually if we reported "modified" here, when it's "untracked".

datalad/local/add_readme.py
adswa: thanks, this was helpful! I have pushed a commit that should implement it. If a .gitattributes file does not exist, it will be created

mih (changes requested):
Late to the party with a potentially controversial opinion....

First of all: thx for addressing this UX issue!

I think adding the `--annex` parameter is needless complexity that maintains the possibility to behave in an undesirable fashion. As stated in https://github.com/datalad/datalad/issues/5713#issue-913131529 I cannot come up with a scenario where one would put a README in the annex and make it less accessible by default.

- If this concern is sensible data: we should recommend to not put sensible data in a README.
- If the concern is "yeah, but this one time at bandcamp": we can state that this simple helper to drop a default template into a dataset is not suitable for special cases.

Overall, I believe the previous behavior is almost always undesirable, and certainly does not justify an extension of the API to maintain it.

Re conditional on when to configure: I am thinking something like this might be best:

```py
if isinstance(dataset.repo, AnnexRepo) \
        and ds.repo.get_gitattributes(filename).get(
            filename, {}).get('annex.largefiles') == 'nothing':
   ...
```

This will add explicit config, unless README.md would already already go into Git unconditionally. Importantly, it leaves the figuring out to Git itself. There is no need to run possible expensive `status` calls, and this approach works with non-DataLad annex repos just as fine.

Alternatively:

```py
'annex.largefiles' not in ds.repo.get_gitattributes(filename).get(filename, {})
```
To avoid interference with *any* existing `annex.largefiles` config that matches `README.md`.

Again, thx for pushing this forward.

bpoldrack:
> I think adding the `--annex` parameter is needless complexity that maintains the possibility to behave in an undesirable fashion. As stated in https://github.com/datalad/datalad/issues/5713#issue-913131529 I cannot come up with a scenario where one would put a README in the annex and make it less accessible by default.
>
> - If this concern is sensible data: we should recommend to not put sensible data in a README.
> - If the concern is "yeah, but this one time at bandcamp": we can state that this simple helper to drop a default template into a dataset is not suitable for special cases.

May be not so controversial. I can certainly get behind that take.

adswa:
thanks for your thoughts, @mih . Could you check if my edit captures what you described?
I am unsure about dropping the check against a modified .gitattributes file - when this file contains unsaved modifications, they will be saved together with the largefile config for the README.

mih:
> thanks for your thoughts, @mih . Could you check if my edit captures what you described?

This works for me, thx!

> I am unsure about dropping the check against a modified .gitattributes file - when this file contains unsaved modifications, they will be saved together with the largefile config for the README.

I think, again given the scope of a convenience helper, and not a low-level building block, it is sufficient to state this clearly in the documentation and stay simple in the implementation.

Apart from the addition to the docs that describe what will happen in an annex repo for the README this LGTM! Thx!

adswa:
Added a short note in the docstring

mih (approved):
LGTM! Thx for the edits!

PR #5726, opened 2021-06-09
Better hint when a sibling is provided without --to

### LINKED ISSUE #5712: `datalad push <something>` is most likely wrong
This is happening to me all the time. And might be a similar experience for other git users.

```
% datalad push github
<no output>
```

The reason is that we have paths as positional args, and it needs `--to` to achieve what I desired here. `push` could check if the positional matches a remote name (before or only when nothing was pushed), and issue a message to suggest to use `--to`.

adswa:
This is an attempt to fix #5712. If
- a path is given, 
- but not ``--to``, 
- (a part of) the path matches a known sibling,
- and no matching files are found, 
 
``push`` reports "impossible" instead of "notneeded", and issues a hint to the user.

Before:
Screenshot from 2021-06-09 16-29-15

After:
Screenshot from 2021-06-09 16-28-58

### DIFF
--- a/datalad/core/distributed/push.py
+++ b/datalad/core/distributed/push.py
@@ -229,7 +229,6 @@ class Push(Interface):
                     if sr
                     else 'No targets configured in dataset.'))
             return
-
         if since == '^':
             # figure out state of remote branch and set `since`
             since = _get_corresponding_remote_state(ds_repo, to)
@@ -266,6 +265,23 @@ class Push(Interface):
             for i, ds in pbars.items():
                 log_progress(lgr.info, i, 'Finished push of %s', ds)
         if not matched_anything:
+            potential_remote = False
+            if not to and path:
+                # if we get a remote name without --to, provide a hint
+                sr = ds_repo.get_remotes(**get_remote_kwargs)
+                potential_remote = [
+                    p for p in ensure_list(path) if p in sr
+                ]
+            if potential_remote:
+                msg = "It seems like you specified a sibling name {} " \
+                      "as a path. Forgot to use --to?".format(potential_remote)
+                yield dict(
+                    res_kwargs,
+                    status='impossible',
+                    message=msg,
+                    type='dataset',
+                    path=ds.path,
+                )
             yield dict(
                 res_kwargs,
                 status='notneeded',
--- a/datalad/core/distributed/push.py
+++ b/datalad/core/distributed/push.py
@@ -298,6 +298,7 @@ class Push(Interface):
 
     @staticmethod
     def custom_result_summary_renderer(results, action_summary):  # pragma: more cover
+        render_action_summary(action_summary)
         # report on any hints at the end
         # get all unique hints
         hints = set([r.get('hints', None) for r in results])
@@ -306,14 +307,13 @@ class Push(Interface):
             from datalad.ui import ui
             from datalad.support import ansi_colors
             intro = ansi_colors.color_word(
-                "Potential hints to solve encountered errors: ",
+                "Hints: ",
                 ansi_colors.YELLOW)
             ui.message(intro)
             [ui.message("{}: {}".format(
                 ansi_colors.color_word(id + 1, ansi_colors.YELLOW), hint))
                 for id, hint in enumerate(hints)]
 
-        render_action_summary(action_summary)
 
 
 def _datasets_since_(dataset, since, paths, recursive, recursion_limit):
--- a/datalad/core/distributed/push.py
+++ b/datalad/core/distributed/push.py
@@ -273,12 +273,16 @@ class Push(Interface):
                     p for p in ensure_list(path) if p in sr
                 ]
             if potential_remote:
-                msg = "{} matches a sibling name and not a path. " \
+                hint = "{} matches a sibling name and not a path. " \
                       "Forgot --to?".format(potential_remote)
                 yield dict(
                     res_kwargs,
-                    status='impossible',
-                    message=msg,
+                    status='notneeded',
+                    message=hint,
+                    hints=hint,
+                    type='dataset',
+                    path=ds.path,
+                )
                     type='dataset',
                     path=ds.path,
                 )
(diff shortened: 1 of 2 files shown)


datalad/core/distributed/push.py
```diff
@@ -229,7 +229,13 @@ def __call__(
                     if sr
                     else 'No targets configured in dataset.'))
             return
-
+        potential_remote = False
+        if not to and path:
```
bpoldrack, line 233: May be that should be relevant only, if the path doesn't actually exist?

adswa: good point. I moved it down into the ``if not matched_anything`` section.

bpoldrack (commented):
Messaging is fine that way. I'm not convinced of the result change, though. What's specified remains (technically) a path that couldn't be matched. I wouldn't know why this should yield `impossible` just because it happens to (also) be a remote name, while it's `notneeded` if it's an arbitrary not matched path.

adswa:
> Messaging is fine that way. I'm not convinced of the result change, though. What's specified remains (technically) a path that couldn't be matched. I wouldn't know why this should yield `impossible` just because it happens to (also) be a remote name, while it's `notneeded` if it's an arbitrary not matched path.

I oriented myself at how the case of a ``push`` without ``--to`` is handled, which also yields an "impossible" result.

bpoldrack:
> I oriented myself at how the case of a push without --to is handled, which also yields an "impossible" result.

I see. But that's not (necessarily) the case here. The target could be auto-detectable (one existing remote only). In fact, if it wasn't we would fail `impossible` before (the `yield from _push` above your changes) and not get to that part of the code.
So, I'd stay in-line with the other `if not matched_anything`-result.

bpoldrack:
Arguably, we could distinguish the cases where `not matched` actually means - doesn't exist from "no changes" and yield impossible in the former case, but this should then apply generally, not just in the case where the not existing path happens to also be a remote name.

datalad/core/distributed/push.py
```diff
@@ -229,7 +229,6 @@ def __call__(
                     if sr
                     else 'No targets configured in dataset.'))
             return
-
```
mih, line 232: Noooo... my favorite empty line...why did it have to die....life is not fair...it feels so meaningless... ;-)

datalad/core/distributed/push.py
```diff
@@ -266,6 +265,23 @@ def __call__(
...
+            potential_remote = False
+            if not to and path:
+                # if we get a remote name without --to, provide a hint
+                sr = ds_repo.get_remotes(**get_remote_kwargs)
+                potential_remote = [
+                    p for p in ensure_list(path) if p in sr
+                ]
+            if potential_remote:
+                msg = "It seems like you specified a sibling name {} " \
+                      "as a path. Forgot to use --to?".format(potential_remote)
```
mih, line 277: I am unsure about the communication style. I would personally do:

```
'{} matches a sibling name and not a path. Forgot --to ?'.format(potential_remote)
```

Just statement of a fact, plus suggestion. Less (unnecessary) speculation on intent. WDYT?

datalad/core/distributed/push.py
```diff
@@ -266,6 +265,23 @@ def __call__(
...
+                yield dict(
+                    res_kwargs,
+                    status='impossible',
+                    message=msg,
+                    type='dataset',
+                    path=ds.path,
+                )
```
mih, line 284: Unclear to me, if the next `yield` should not move into an `else:`. If not, a comment on why not might help the next reader.

datalad/core/distributed/push.py
```diff
@@ -266,6 +265,23 @@ def __call__(
             for i, ds in pbars.items():
                 log_progress(lgr.info, i, 'Finished push of %s', ds)
         if not matched_anything:
+            potential_remote = False
+            if not to and path:
```
mih, line 269:
```suggestion
            if not to and len(paths) == 1:
```
This is a readily available list of positional args, and we would only have an idea, if there is exactly one arg that may turn out to match a sibling name. We can rule out all other cases here and save cycles.

mih (commented):
Thx! I think this will improve UX!

I left a few comments on some possible optimizations.

datalad/core/distributed/push.py
adswa: I like this suggestion. Thanks!

datalad/core/distributed/push.py
adswa: I appreciate hints on what is considered best practice in this case. The next yield, I think, should only be conditional on ``not matched_anything``. The previous conditionals (`ìf not to and path`` and ``potential_remote``) only exist to capture the UX issue this PR tries to tackle before the "notneeded" of the next yield statement is issued.
I could do an ``else [no potential_remote]``, but to me this feels like I would unnecessarily confine myself into a conditional (i.e., somewhat implying that ``potential_remote`` is more related to unmatched paths)... What would you do? Is it more readable to use ``else``?

datalad/core/distributed/push.py
adswa: I will bring it back to life ;-)

datalad/core/distributed/push.py
bpoldrack: Looking at this again, I'd say there are two valid things:
- `else`
- `return` after the `impossible` result

Currently both results are yielded - which we didn't catch before, since it's not tested and a command line output wouldn't show it b/c `notneeded` isn't rendered by default.

Re style: I'd probably go with `return` and a comment before the next yield that states the condition we are in at that point (not matched anything). That's b/c of the same feeling: The `notneeded` has a relation to the conditional for `impossible` only "by chance" at this point in time (could change). But that's no strong opinion.

adswa: thx, done

adswa:
I have reused the ``hints`` framework that ``push``'s custom result renderer uses now. Here is how it looks like:

Screenshot from 2021-06-14 13-43-38

adswa:
The failure of one AppVeyor run seems unrelated

bpoldrack:
> The failure of one AppVeyor run seems unrelated

Failed during setup. Restarted.

mih (approved):
This looks good to me now! Thx!

PR #5730, opened 2021-06-10
Add config datalad.remote.archives.autoclean

### LINKED ISSUE #5519: datalad-archive leaves temp. files behind by default
Problem is: Build a dataset with `add-archive-content` to annex the archive as well as the extracted content using the datalad-archive special remote for extraction.
Now, in a fresh clone that only knows a source for the archive (as opposed to knowing a remote with the extracted content) run a `datalad get .`. This effectively leaves that clone requiring up-to 3 times the space of the content we wanted to get:
- the archive
- the extracted content in `.git/annex/objects`
- an extraction cache in `.git/datalad/tmp`

I think at least the cache shouldn't be persistent by default. The demand on space seems too surprising for a default. Arguably the same is true for keeping the archive around, when the `get` was called on the extracted content. In any case, the special remote should communicate how to drop the archive and how to use `datalad clean` to remove the cache, if any of those was kept (INFO message to annex).

Stumbled upon this in https://github.com/psychoinformatics-de/datalad-hirni/issues/183

Ping @yarikoptic 

Ping @pvavra, so you know where to follow ...

bpoldrack:
This flag determines whether or not the datalad-archives
special remote keeps extracted files under .git as an
extraction cache. Default set to False, since this can
be expensive to keep around, it's not obvious for users
that it's there and how to remove it, and generally the
defaults should scale. True doesn't scale here.

Closes #5519

There's a little more TODO - see https://github.com/datalad/datalad/issues/5519#issuecomment-802883995

@yarikoptic : What do you think of this approach?

### DIFF
--- a/datalad/customremotes/main.py
+++ b/datalad/customremotes/main.py
@@ -21,6 +21,7 @@ from ..cmdline import helpers
 
 from ..utils import setup_exceptionhook
 from ..ui import ui
+from datalad import cfg
 
 backends = ['archive']
 
@@ -77,7 +78,10 @@ def _main(args, backend=None):
     assert(backend is not None)
     if backend == 'archive':
         from .archives import ArchiveAnnexCustomRemote
-        remote = ArchiveAnnexCustomRemote()
+        autoclean = cfg.obtain("datalad.remote.archives.autoclean")
+        remote = ArchiveAnnexCustomRemote(
+            persistent_cache=not autoclean
+        )
     elif backend == 'datalad':
         from .datalad import DataladAnnexCustomRemote
         remote = DataladAnnexCustomRemote()
--- a/datalad/customremotes/tests/test_archives.py
+++ b/datalad/customremotes/tests/test_archives.py
@@ -338,3 +338,48 @@ def test_link_file_load(tempfile):
         assert_equal(f.read(), "LOAD")
     assert_equal(stats(tempfile, times=False), stats(tempfile2, times=False))
     unlink(tempfile2)  # TODO: next two with_tempfile
+
+
+@with_tree(tree=
+    {'1.tar.gz':
+         {
+             'bu.dat': '52055957098986598349795121365535' * 10000,
+             'bu3.dat': '8236397048205454767887168342849275422' * 10000
+          },
+    }
+)
+@with_tempfile
+def test_autoclean_cache(dspath, clonepath):
+    from datalad.consts import ARCHIVES_TEMP_DIR
+    from datalad.api import add_archive_content
+    from datalad.api import create
+    from datalad.api import clone
+
+    ds = create(dspath, force=True)
+    ds.save()
+    # NOTE: add_archive_content isn't proper datalad command ATM. It doesn't
+    # operate on a Dataset instance!
+    with chpwd(dspath):
+        add_archive_content('1.tar.gz', delete=True, drop_after=True)
+
+    clone_ds = clone(ds.path, clonepath)
+    cache_path = clone_ds.pathobj / ARCHIVES_TEMP_DIR
+
+    try:
+        with patch.dict("os.environ",
+                        {"DATALAD_REMOTE_ARCHIVES_AUTOCLEAN": "1"}):
+            dl_cfg.reload()
+            clone_ds.get()
+            assert_false(cache_path.exists())
+
+        clone_ds.drop('.')
+
+        with patch.dict("os.environ",
+                        {"DATALAD_REMOTE_ARCHIVES_AUTOCLEAN": "0"}):
+            dl_cfg.reload()
+            clone_ds.get()
+            assert_true(cache_path.exists())
+            # contains .stamp + dir
+            assert_equal(len([p for p in cache_path.iterdir()]), 2)
+    finally:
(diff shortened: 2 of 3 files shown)


datalad/customremotes/tests/test_archives.py
```diff
@@ -338,3 +338,48 @@ def raise_AttributeError(*args):
...
+        add_archive_content('1.tar.gz', delete=True, drop_after=True)
+
+    clone_ds = clone(ds.path, clonepath)
+    cache_path = clone_ds.pathobj / ARCHIVES_TEMP_DIR
+
+    try:
+        with patch.dict("os.environ",
+                        {"DATALAD_REMOTE_ARCHIVES_AUTOCLEAN": "1"}):
+            dl_cfg.reload()
+            clone_ds.get()
```
yarikoptic, line 372: Since we haven't tested ever functionality with such non-persistent cache, I would recommend
- make archive contain lots (100?) of (small) files
- add `jobs=5` or so here to see if there is no some kind of interaction between separate get jobs while trying to use non-persistent cache

datalad/customremotes/main.py
```diff
@@ -77,7 +78,10 @@ def _main(args, backend=None):
     assert(backend is not None)
     if backend == 'archive':
         from .archives import ArchiveAnnexCustomRemote
-        remote = ArchiveAnnexCustomRemote()
+        autoclean = cfg.obtain("datalad.remote.archives.autoclean")
+        remote = ArchiveAnnexCustomRemote(
+            persistent_cache=not autoclean
```
yarikoptic, line 83: although I do like `autoclean`, it is not really corresponding to the underlying mechanism, smth like `cache.persistent = bool` would have been more reflecting of what is going on.

gave me also idea for `autodrop` where we would also auto drop the underlying archive if that process did `get` it (not that it should be done here, just an idea).  But with the comment below on `jobs` -- for `autodrop` might be even trickier.

yarikoptic (changes requested):
IMHO needs testing to be adjusted to verify that works ok with some `jobs!=1`.  Also not 100% sure about the config name.

datalad/customremotes/tests/test_archives.py
bpoldrack: Ok, thanks!

datalad/customremotes/main.py
bpoldrack: Yes, wondered about better naming as well, just went with what you proposed in the issue for now.

> But with the comment below on jobs -- for autodrop might be even trickier.

Yes, I'm not sure how to address this. The special remote itself might be the wrong place.

bpoldrack:
Thx, @yarikoptic. Just wondered, whether the way of addressing it is missing something, since you originally suggested to handle this explicitly at exit rather than just enabling/disabling the persistent cache. Looking through, it seems `clean()` would be properly called, but I may have missed some pecularities.

yarikoptic:
well, clean is called in `__del__`
```
    def __del__(self):
        try:
            # we can at least try
            if not self.persistent:
                self.clean()
        except:  # MIH: IOError?
            pass
```
and with python3 that `__del__` might get there whenever `.persistent` is not even accessible any longer AFAIK... that is why probably I was mumbling about an explicit `.clean` call upon `atexit` , but again not sure if it would not cause disturbance in case of multiple `datalad-archive` special remotes spawn by `annex get --jobs` since then it is a race condition and if annex stops some worker whenever it sees that no more work to be done, but before all workers are done (might not be the case, would need to ask @joeyh) then one of the workers might cleanup before all workers done copying the content from the extracted archive.

datalad/customremotes/main.py
yarikoptic: > Yes, I'm not sure how to address this. The special remote itself might be the wrong place.

may be you are hitting it on the nail -- may be it could be `datalad.runtime.get.autoclean` which just calls out to `clean` upon completion, and thus taking care about cleaning everything (archives cache and also possibly some annex tmp files etc)?

yarikoptic: oh no -- also a tricky case: I remember having nested tarballs !  Then we might end up in some nested `datalad.get` which would then react on `autoclean` and clean up after itself **and** parent process... not good.  tricky business it is all this "cleaning" ;)

bpoldrack: > may be it could be datalad.runtime.get.autoclean which just calls out to clean upon completion, and thus taking care about cleaning everything (archives cache and also possibly some annex tmp files etc)?

Currently thinking in the same direction, yes.

>  Then we might end up in some nested datalad.get which would then react on autoclean and clean up after itself and parent process... not good.

Hm. What about changing the cache dir layout? `.git/datalad/tmp/archives/<key of the archive itself>/...`? Then a get theoretically could clean only the for the key(s) it was supposed to get. May need some more serious RF'ing then it sounds, though.

bpoldrack:
Ok. Going through the code base of that special remote, I think addressing the actual issue and making this cache non-persistent requires some substantial changes, that I don't have the time for right now.
So, will close and only propose PR #5738 to at least improve on UX for now.

PR #5708, opened 2021-06-03
ENH(UX): state original purpose in NoDatasetFound exception + detail it for get

yarikoptic:
I have seen those reported by users, and I must confess that some of those stem from the confusing interface of `install` + RFing which lead `install`/`get` to unable to operate on arbitrary paths (#3759).  But I think that such error might come up in other use-cases as well.  So before this PR users might get:

```shell
$> datalad install git://github.com/dl-throwaway/test_gh3 dest                                   
[INFO   ] scanning for unlocked files (this may take some time)                                           
[INFO   ] Remote origin uses a protocol not supported by git-annex; setting annex-ignore                  
install(ok): /home/yoh/.tmp/test_gh3 (dataset)
[ERROR  ] No dataset found at '/home/yoh/.tmp'.  Specify a dataset to work with by providing its path via the `dataset` option, or change the current working directory to be in a dataset. [dataset.py:require_dataset:569] (NoDatasetFound) 
usage: datalad install [-h] [-s SOURCE] [-d DATASET] [-g] [-D DESCRIPTION] [-r] [-R LEVELS]
                       [--reckless [auto|ephemeral|shared-...]] [-J NJOBS]
                       [PATH ...]
```

which leaves them wondering on WTF (error message nohow points to `dest`) and needing to ask/RTFM etc to figure out that they needed `-s`.  With this change it would look like

```shell
$> datalad install git://github.com/dl-throwaway/test_gh3 dest
[ERROR  ] No dataset found at '/home/yoh/.tmp' for the purpose "get content of ['dest']".  Specify a dataset to work with by providing its path via the `dataset` option, or change the current working directory to be in a dataset. [dataset.py:require_dataset:569] (NoDatasetFound) 
usage: datalad install [-h] [-s SOURCE] [-d DATASET] [-g] [-D DESCRIPTION] [-r] [-R LEVELS]
                       [--reckless [auto|ephemeral|shared-...]] [-J NJOBS]
                       [PATH ...]
```
it is still not perfect but it immediately points to the underlying culprit.

PR #5733, opened 2021-06-10
ENH: harmonize "purpose" statements to imperative form

yarikoptic:
We had a mix, and even though "ing form was probably more prevalent, I felt that shorter
imperative form would be better, even though it would be not "for {purpose}" but
"to {purpose}" if to be displayed. (as eg in #5708)


bpoldrack:
Nice. Don't give anything about which form, but harmonizing it certainly is desirable. :)

mih (approved):
I like this! thx!

adswa (approved):
I like it too!

PR #5735, opened 2021-06-10
BF+ENH+RF: logic and code support for --jobs='auto', make AnnexRepo. interfaces use 'auto' by default

yarikoptic:
The first commit is a true BF which kinda started it all.

And then I think I went a little "too broad" in my changes, even if we decide that it is a good time to return getting annex operations to do some parallelization by default ('auto') it might be a better fit for `master` then.  Alternatively I could kick out that commit which does set default to `auto` for AnnexRepo interfaces, and keep the rest, but then I think it would loose the point of all other RFings... uff...  At least for `maint` we could take the BF and handling for `0` value (as to "do not bother annex") which is now the effect of setting it to `1` (prior default) but then there is no way to overload our or annex config setting (if set > 1) via env var as should be possible.

Please see individual commits  for more reasoning etc.

In any case, I am interested to see if
- [x] we gain run time speed ups for CIs
   - not spotted any gain
- [x] anything fails
   - quite enough -- see below/referenced to this PR
- [ ] decide on how to proceed if at all (besides a tiny bugfix PR)

edit 1: in `maint` we still have 7.20190503 min annex version, and in `master` 8.20200309.  This [annex issue](https://git-annex.branchable.com/bugs/get_-J8_on_OSX_leads_to_git-annex__58___git__58___createProcess__58___runInteractiveProcess__58___pipe__58___resource_exhausted___40__Too_many_open_files__41__/) was fixed before (according to the date of changes, didn't git describe) 20190503 so users should not run into it again even with older annexes. 

edit 2: having thought about it a bit more, I think this PR is a bust and we might just keep as it was (which in effect was "defer to git-annex config" albeit without being able to change it via env var) besides any bug fixes we need to CP or to do.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -51,6 +51,7 @@ shallow_clone: true
 
 environment:
   DATALAD_TESTS_SSH: 1
+  DATALAD_RUNTIME_MAX__ANNEX__JOBS: 5
 
   # Do not use `image` as a matrix dimension, to have fine-grained control over
   # what tests run on which platform
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,8 @@ env:
     - DATALAD_TESTS_SSH=1
     - DATALAD_LOG_ENV=GIT_SSH_COMMAND
     - TESTS_TO_PERFORM=datalad
+    # To potentially speed up testing, let's do annex jobs in parallel by default
+    - DATALAD_RUNTIME_MAX__ANNEX__JOBS=5
     # Should be an array, travis breaks on it, define/adjust in the "before_install"
     #- NOSE_OPTS=( -v )
     # Note, that there's "turtle" as well, which is always excluded from
@@ -38,6 +40,8 @@ matrix:
     # We cannot have empty -A selector, so the one which always will be fulfilled
     - NOSE_SELECTION=
     - NOSE_SELECTION_OP=not
+    # A single run where we do not instruct annex on number of jobs
+    - DATALAD_RUNTIME_MAX__ANNEX__JOBS=0
     # To test https://github.com/datalad/datalad/pull/4342 fix.
     # From our testing in that PR seems to have no effect, but kept around since should not hurt.
     - LC_ALL=ru_RU.UTF-8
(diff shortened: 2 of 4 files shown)


mih:
I am linking https://github.com/datalad/datalad/issues/4404 here, which is the last investigation we had on possible side-effects of going parallel by default.

yarikoptic:
> I am linking #4404 here, which is the last investigation we had on possible side-effects of going parallel by default.

those I think (as described in `edit 1` of this PR description) were mitigated at git-annex level and should no longer be encountered.

Overall one of the goals (should split it apart) for this PR was to bring back testing parallel operation of git-annex which we stopped doing and thus possibly ending up slowly brewing some regressions or new issues which only users would encounter if they do some `--jobs auto` or alike. And it seems we did get a few of those: #5742  #5743 #5745 :-(.  But I guess for that we should not need this PR code RF and could just set `annex.jobs` somehow in the global git config env (but we create one in our fixture!).

not sure what's up with hirni since seems got stuck and logs aren't suggesting anything to me.

Overall -- I think a 2nd thought is needed on how to RF this PR so we would retain old behavior (no parallelism by default) but be able to adjust it via env var config setting.

yarikoptic:
FWIW, I do not see any consistent improvement in travis CI run time (well -- who knows, may instance was busier/slower etc) overall in this PR if compared to another run of maint:
```
(tinuous-dev) datalad@smaug:/mnt/datasets/datalad/ci/logs/2021/06$ grep Ran 14/push/maint/09dcf7c/travis-12068-success/*  | sed -e 's,.*/,,g' | sort -n
1.txt:Ran 1271 tests in 2000.884s
2.txt:Ran 1271 tests in 1960.163s
3.txt:Ran 1148 tests in 778.720s
4.txt:Ran 132 tests in 1060.977s
5.txt:Ran 132 tests in 1068.617s
6.txt:Ran 1148 tests in 800.113s
7.txt:Ran 1148 tests in 968.214s
8.txt:Ran 540 tests in 323.074s

(tinuous-dev) datalad@smaug:/mnt/datasets/datalad/ci/logs/2021/06$ grep Ran 10/pr/5735/8e5398a/travis-12054-*/*  | sed -e 's,.*/,,g' | sort -n
1.txt:Ran 1271 tests in 2270.288s
2.txt:Ran 1271 tests in 1910.976s
3.txt:Ran 1148 tests in 660.163s
4.txt:Ran 132 tests in 1132.893s
5.txt:Ran 132 tests in 1218.141s
6.txt:Ran 1148 tests in 857.430s
7.txt:Ran 1148 tests in 1023.897s
8.txt:Ran 540 tests in 296.677s
```

yarikoptic:
IIRC I decided not to bother, at least this way, so let's let it RiP for now

PR #4338, opened 2020-03-24
ENH: Capture hints from Git

adswa:
Sometimes, Git has "hints" why pushing fails. 
For example, if I attempt a push without integrating remote changes, it says

```
error: failed to push some refs to 'git@github.com:datalad-handbook/DataLad-101.git'
hint: Updates were rejected because the tip of your current branch is behind
hint: its remote counterpart. Integrate the remote changes (e.g.
hint: 'git pull ...') before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.
```

`datalad push` swallowed these hints. In this PR I'm attempting to catch them and report all hints at the end of the result rendering.

#### A few examples:

```
datalad push --to blubb43
publish(error): . (dataset) [refs/heads/git-annex->blubb43:refs/heads/git-annex [rejected] (fetch first)]
Potential hints to solve encountered errors: 
1: Updates were rejected because the remote contains work that you do not have locally. This is usually caused by another repository pushing to the same ref. You may want to first integrate the remote changes (e.g., 'git pull ...') before pushing again. See the 'Note about fast-forwards' in 'git push --help' for details.

```

A situation with lots of the same hints 

```
datalad push --to public                                                                     1 ↵
publish(error): . (dataset) [refs/heads/master->public:refs/heads/master [rejected] (non-fast-forward)]
publish(error): . (dataset) [refs/heads/sct_back_and_forth_in_time->public:refs/heads/sct_back_and_forth_in_time [rejected] (non-fast-forward)]
publish(error): . (dataset) [refs/heads/sct_computational_reproducibility->public:refs/heads/sct_computational_reproducibility [rejected] (non-fast-forward)]
publish(error): . (dataset) [refs/heads/sct_create_a_dataset->public:refs/heads/sct_create_a_dataset [rejected] (non-fast-forward)]
publish(error): . (dataset) [refs/heads/sct_datalad_rerun->public:refs/heads/sct_datalad_rerun [rejected] (non-fast-forward)]
publish(error): . (dataset) [refs/heads/sct_hide_content->public:refs/heads/sct_hide_content [rejected] (non-fast-forward)]
publish(error): . (dataset) [refs/heads/sct_input_and_output->public:refs/heads/sct_input_and_output [rejected] (non-fast-forward)]
publish(error): . (dataset) [refs/heads/sct_install_datasets->public:refs/heads/sct_install_datasets [rejected] (non-fast-forward)]
[...]

PR #5738, opened 2021-06-11
Dry run option for clean command

### LINKED ISSUE #5834: Make users aware of (the need for) datalad clean
Creating this issue to separate aspects of #5519.

Not only, but particularly the extraction cache of `datalad-archives` is a surprising and not easy to discover location requiring disc space. A `--dry-run` for `clean` would help discover it and in addition the special remote itself shouldn't be silent about it, but announce the creation of that extraction cache.

bpoldrack:
There are several places where temporary files could be left by datalad operations. `datalad clean` now comes with a `--dry-run` option to just report on what could be cleaned.
Furthermore, the `datalad-archives` special remote now issues an INFO message to annex when it's actually extracting an archive in order to let a user know they may want to call `datalad clean` afterwards. This should make the presence of an extraction cache less surprising to users and make them aware of the `clean` command.

The issue leading to this PR and the discussion therein is #5519, although this is not actually solving it. However, at least this is an improvement of the UX aspect of the issue.

Closes #5834

### DIFF
--- a/datalad/customremotes/archives.py
+++ b/datalad/customremotes/archives.py
@@ -32,6 +32,7 @@ from ..utils import ensure_bytes
 from ..utils import unlink
 from .base import AnnexCustomRemote
 from .main import main as super_main
+from datalad.consts import ARCHIVES_SPECIAL_REMOTE
 
 
 # ####
@@ -393,8 +394,16 @@ class ArchiveAnnexCustomRemote(AnnexCustomRemote):
                 # so
                 pwd = getpwd()
                 lgr.debug(u"Getting file {afile} from {akey_path} while PWD={pwd}".format(**locals()))
+                was_extracted = self.cache[akey_path].is_extracted
                 apath = self.cache[akey_path].get_extracted_file(afile)
                 link_file_load(apath, path)
+                if not was_extracted and self.cache[akey_path].is_extracted:
+                    self.info("%s special remote is using an extraction cache "
+                              "under %s. Remove it with DataLad's 'clean' "
+                              "command to save disk space." %
+                              (ARCHIVES_SPECIAL_REMOTE,
+                               self.cache[akey_path].path)
+                              )
                 self.send('TRANSFER-SUCCESS', cmd, key)
                 return
             except Exception as exc:
--- a/datalad/interface/clean.py
+++ b/datalad/interface/clean.py
@@ -24,7 +24,9 @@ from ..consts import (
 )
 
 from datalad.support.gitrepo import GitRepo
-from datalad.support.constraints import EnsureNone
+from datalad.support.constraints import (
+    EnsureNone,
+)
 from datalad.distribution.dataset import (
     EnsureDataset,
     require_dataset,
@@ -37,7 +39,6 @@ from datalad.interface.common_opts import (
 from datalad.interface.results import get_status_dict
 from datalad.interface.utils import eval_results
 from datalad.interface.base import build_doc
-
 from logging import getLogger
 lgr = getLogger('datalad.api.clean')
 
@@ -56,6 +57,8 @@ class Clean(Interface):
       $ datalad clean
     """
 
+    result_renderer = 'tailored'
+
     _params_ = dict(
         dataset=Parameter(
             args=("-d", "--dataset"),
@@ -63,15 +66,20 @@ class Clean(Interface):
                 no dataset is given, an attempt is made to identify the dataset
                 in current working directory""",
             constraints=EnsureDataset() | EnsureNone()),
-        # TODO:  --info  -- which performs dry run just summarizing what is to be cleaned up
+        dry_run = Parameter(
+            args=("--dry-run",),
+            doc="""Report on cleanable locations - not actually cleaning up
+            anything.""",
+            action="store_true",),
         # TODO: Python only???
         what=Parameter(
             args=("--what",),
             dest='what',
-            choices=('cached-archives', 'annex-tmp', 'annex-transfer', 'search-index'),
+            choices=('cached-archives', 'annex-tmp', 'annex-transfer',
+                     'search-index'),
             nargs="*",
-            doc="""What to clean.  If none specified -- all known targets are
-            cleaned"""),
+            doc="""What to clean. If none specified -- all known
+            targets are considered."""),
         recursive=recursion_flag,
         recursion_limit=recursion_limit,
     )
@@ -79,9 +87,14 @@ class Clean(Interface):
     @staticmethod
     @datasetmethod(name='clean')
     @eval_results
-    def __call__(dataset=None, what=None, recursive=False, recursion_limit=None):
-        ds = require_dataset(dataset, purpose='clean-up')
-        res_kwargs = dict(action='clean', logger=lgr, refds=ds.path)
+    def __call__(dataset=None, what=None, dry_run=False,
+                 recursive=False, recursion_limit=None):
+
+        ds = require_dataset(dataset,
+                             purpose="report on cleanable locations"
+                             if dry_run else "clean dataset")
+        res_kwargs = dict(action='clean [dry-run]' if dry_run else 'clean',
+                          logger=lgr, refds=ds.path)
         for wds in itertools.chain([ds], ds.subdatasets(
                 fulfilled=True,
                 recursive=recursive,
@@ -115,10 +128,62 @@ class Clean(Interface):
                         path=topdir, status='notneeded', type='directory', **res_kwargs)
                     continue
                 pl = len(paths) > 1
-                message = ("Removed %d %s %s: %s",
+                message = ("%s %d %s %s: %s",
+                           "Discovered" if dry_run else "Removed",
                            len(paths), msg, sing_pl[int(pl)],
                            ", ".join(sorted([x[len(topdir) + 1:] for x in paths])))
-                rmtree(topdir)
-                yield get_status_dict(
-                    path=topdir, status='ok', type='dir', message=message,
-                    **res_kwargs)
+
+                if not dry_run:
+                    rmtree(topdir)
+
+                yield get_status_dict(path=topdir,
+                                      status='ok',
+                                      type='directory',
+                                      message=message,
+                                      **res_kwargs)
+
+    @staticmethod
+    def custom_result_renderer(res, **kwargs):  # pragma: more cover
+        # Don't render things like 'status' for clean-info messages -
+        # seems rather meaningless.
+
+        from datalad.interface.utils import default_result_renderer
+        import datalad.support.ansi_colors as ac
+        from datalad.utils import Path
+        from os import getcwd
+
+        if res['action'] == 'clean':
+            # default renderer is just fine
+            return default_result_renderer(res)
+        elif res['action'] != 'clean [dry-run]':
+            # Result didn't come from within `clean`.
+            # Should be handled elsewhere.
+            return
+
+        assert res['action'] == 'clean [dry-run]'
+
+        if res.get('status', None) == 'ok':
+            from datalad.ui import ui
+
+            # when to render relative paths:
+            #  1) if a dataset arg was given
+            #  2) if CWD is the refds
+
+            refds = res.get('refds', None)
+            refds = refds if kwargs.get('dataset', None) is not None \
+                or refds == getcwd() else None
+            path = res['path'] if refds is None \
+                else str(Path(res['path']).relative_to(refds))
+
+            ui.message(u"{path}: {message}".format(
+                path=ac.color_word(path, ac.BOLD),
+                message=(res['message'][0] % res['message'][1:]
+                         if isinstance(res['message'], tuple)
+                         else res['message'])
+                        if res.get('message', None) else ''
+                )
+            )
+
+        else:
+            # Any other status than 'ok' is reported the default way.
+            return default_result_renderer(res)
--- a/datalad/interface/clean.py
+++ b/datalad/interface/clean.py
@@ -10,11 +10,12 @@
 __docformat__ = 'restructuredtext'
 
 
-from os.path import join as opj
-from glob import glob
 import itertools
 from .base import Interface
-from ..utils import rmtree
+from ..utils import (
+    rmtree,
+    Path
+)
 from ..support.param import Parameter
 from ..consts import (
     ARCHIVES_TEMP_DIR,
@@ -23,7 +24,6 @@ from ..consts import (
     SEARCH_INDEX_DOTGITDIR,
 )
 
-from datalad.support.gitrepo import GitRepo
 from datalad.support.constraints import (
     EnsureNone,
 )
@@ -50,14 +50,34 @@ import datalad.distribution.subdatasets
 class Clean(Interface):
     """Clean up after DataLad (possible temporary files etc.)
 
-    Removes extracted temporary archives, etc.
+    Removes temporary files and directories left behind by DataLad and
+    git-annex in a dataset.
 
-    Examples:
-
-      $ datalad clean
     """
 
     result_renderer = 'tailored'
+
+    _examples_ = [
+        dict(text="Clean all known temporary locations of a dataset",
+             code_py="clean()",
+             code_cmd="datalad clean"),
+        dict(text="Report on all existing temporary locations of a dataset",
+             code_py="clean(dry_run=True)",
+             code_cmd="datalad clean --dry-run"),
+        dict(text="Clean all known temporary locations of a dataset and all "
+                  "its subdatasets",
+             code_py="clean(recursive=True)",
+             code_cmd="datalad clean -r"),
+        dict(text="Clean only the archive extraction caches of a dataset and "
+                  "all its subdatasets",
+             code_py="clean(what='cached-archives', recursive=True)",
+             code_cmd="datalad clean --what cached-archives -r"),
+        dict(text="Report on existing annex transfer files of a dataset and "
+                  "all its subdatasets",
+             code_py="clean(what='annex-transfer', recursive=True, "
+                     "dry_run=True)",
+             code_cmd="datalad clean --what annex-transfer -r --dry-run"),
+    ]
 
     _params_ = dict(
         dataset=Parameter(
@@ -66,7 +86,7 @@ class Clean(Interface):
                 no dataset is given, an attempt is made to identify the dataset
                 in current working directory""",
             constraints=EnsureDataset() | EnsureNone()),
-        dry_run = Parameter(
+        dry_run=Parameter(
             args=("--dry-run",),
             doc="""Report on cleanable locations - not actually cleaning up
             anything.""",
@@ -102,41 +122,57 @@ class Clean(Interface):
                 return_type='generator',
                 result_renderer='disabled',
                 result_xfm='datasets') if recursive else []):
-            d = wds.path
-            gitdir = GitRepo.get_git_dir(d)
+            d = wds.pathobj
+            gitdir = wds.repo.dot_git
             DIRS_PLURAL = ("directory", "directories")
             FILES_PLURAL = ("file", "files")
+            discover_or_remove = "Discovered" if dry_run else "Removed"
+
             for dirpath, flag, msg, sing_pl in [
-                (ARCHIVES_TEMP_DIR, "cached-archives",
+                (Path(ARCHIVES_TEMP_DIR), "cached-archives",
                  "temporary archive", DIRS_PLURAL),
-                (ANNEX_TEMP_DIR, "annex-tmp",
+                (Path(ANNEX_TEMP_DIR), "annex-tmp",
                  "temporary annex", FILES_PLURAL),
-                (ANNEX_TRANSFER_DIR, "annex-transfer",
+                (Path(ANNEX_TRANSFER_DIR), "annex-transfer",
                  "annex temporary transfer", DIRS_PLURAL),
-                (opj(gitdir, SEARCH_INDEX_DOTGITDIR), 'search-index',
+                (gitdir / Path(SEARCH_INDEX_DOTGITDIR), 'search-index',
                  "metadata search index", FILES_PLURAL),
             ]:
-                topdir = opj(d, dirpath)
+                topdir = wds.pathobj / dirpath
                 lgr.debug("Considering to clean %s:%s", d, dirpath)
                 if not ((what is None) or (flag in what)):
                     yield get_status_dict(
-                        path=topdir, status='notneeded', type='directory', **res_kwargs)
+                        path=str(topdir), status='notneeded', type='directory',
+                        **res_kwargs)
                     continue
-                paths = glob(opj(topdir, '*'))
+
+                paths = [p for p in topdir.glob('*')]
                 if not paths:
-                    yield get_status_dict(
-                        path=topdir, status='notneeded', type='directory', **res_kwargs)
-                    continue
-                pl = len(paths) > 1
-                message = ("%s %d %s %s: %s",
-                           "Discovered" if dry_run else "Removed",
-                           len(paths), msg, sing_pl[int(pl)],
-                           ", ".join(sorted([x[len(topdir) + 1:] for x in paths])))
+                    if not topdir.exists():
+                        yield get_status_dict(
+                            path=str(topdir), status='notneeded',
+                            type='directory', **res_kwargs)
+                        continue
+                    else:
+                        # we empty topdir only
+                        message = ("%s empty %s directory",
+                                   discover_or_remove,
+                                   msg)
+                else:
+                    pl = len(paths) > 1
+                    message = ("%s %d %s %s: %s",
+                               discover_or_remove,
+                               len(paths), msg, sing_pl[int(pl)],
+                               ", ".join(sorted([str(p.relative_to(topdir))
+                                                 for p in paths if p != topdir]
+                                                )
+                                         )
+                               )
 
                 if not dry_run:
-                    rmtree(topdir)
+                    rmtree(str(topdir))
 
-                yield get_status_dict(path=topdir,
+                yield get_status_dict(path=str(topdir),
                                       status='ok',
                                       type='directory',
                                       message=message,
@@ -187,3 +223,13 @@ class Clean(Interface):
         else:
             # Any other status than 'ok' is reported the default way.
             return default_result_renderer(res)
+
+    @staticmethod
+    def custom_result_summary_renderer(results):
+        # Since 'notneeded' results aren't rendered by default, give
+        # a nothing-to-clean-message if all results were "notneeded",
+        # to not remain entirely silent.
+
+        if all(r['status'] == 'notneeded' for r in results):
+            from datalad.ui import ui
+            ui.message("nothing to clean, no temporary locations present.")
--- a/datalad/customremotes/tests/test_archives.py
+++ b/datalad/customremotes/tests/test_archives.py
@@ -122,7 +122,12 @@ def test_basic_scenario(d, d2):
     in_('[%s]' % ARCHIVES_SPECIAL_REMOTE, list_of_remotes)
 
     assert_false(annex.file_has_content(fn_extracted))
-    annex.get(fn_extracted)
+
+    with swallow_logs(new_level=logging.INFO) as cml:
+        annex.get(fn_extracted)
+        # Hint users to the extraction cache (and to datalad clean)
+        cml.assert_logged(msg="datalad-archives special remote is using an "
+                              "extraction", level="INFO", regex=False)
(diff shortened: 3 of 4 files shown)


bpoldrack:
Note, that this is for now replacing PR #5730, as an actual solution turns out to be more of an effort than I can currently afford. So, not closing #5519 really, but at least let the user know what's going on and how to remove that cache.

@mih : Guess, with that the issue can be regarded as addressed within context of https://github.com/datalad/datalad/projects/13, since the "real solution" is actually significantly more than a "simple UX issue".

bpoldrack:
Looking into more reporting on this. Thus proposing `--info` for `clean`, which in turn could be called by `status`.
Proposal in https://github.com/datalad/datalad/pull/5738/commits/73d97d47286f133bea7f11197522941ed11f5b24. WDYT, @datalad/developers ?

Yes - needs tests. I'm aware. :)

mih (commented):
Thx, I think linking this into `status` is a reasonable way to give this aspect the visibility it needs in order to avoid accumulation of cruft.

That being said: I personally see the reporting on cleanable locations not as a core function of `status`, hence I do not think it should actually yield such results (unconditionally). A programmatic user that is concerned with cleaning things up, can, will, and has to make a dedicated call to `clean`. With the present change, `clean` will be called twice for them (once with `--info` and once to actually clean). -- and once for everybody else, unconditionally.

I think it would make sense to move the call to clean into `custom_result_summary_renderer()` and only report on cleanable locations, if there are any. This setup ensures that the cycles necessary are only spent when there is a chance that a (human) consumer catches the notification, and, e.g. not in pipes with `datalad -f json`.

bpoldrack:
> I think it would make sense to move the call to clean into custom_result_summary_renderer() and only report on cleanable locations, if there are any. 

Makes sense to me. Thx.

bpoldrack:
> only report on cleanable locations, if there are any.

FTR: That was the case already.

However, changed to what you suggested, I think, @mih. This now includes a commit to pass original kwargs to custom summary renderers (as we already do for custom result renderers).

datalad/core/local/status.py
```diff
@@ -495,3 +495,24 @@ def custom_result_summary_renderer(results):  # pragma: more cover
                for r in results):
             from datalad.ui import ui
             ui.message("nothing to save, working tree clean")
+
+        # ### Report on possibly cleanable temporary locations
+
+        # Infer the dataset we ran on from the results and get original
+        # recursive (+limit) options:
+        refds = set(r['refds'] for r in results)
+        assert len(refds) == 1
```
mih, line 504: If for whatever reason the assertion is wrong, it will crash the process. I think this is a bit extreme for a convenience report. Maybe just `return`?

datalad/interface/clean.py
```diff
@@ -56,32 +57,44 @@ class Clean(Interface):
...
             doc="""specify the dataset to perform the clean operation on.  If
                 no dataset is given, an attempt is made to identify the dataset
                 in current working directory""",
             constraints=EnsureDataset() | EnsureNone()),
-        # TODO:  --info  -- which performs dry run just summarizing what is to be cleaned up
+        info=Parameter(
+            args=("-i", "--info"),
+            doc="""Report on cleanable locations - not actually cleaning up
+            anything.""",
+            action='store_true'),
```
mih, line 73: Thinking about this more, I feel that adopting an already established pattern could have benefits here. `run` uses `run [dry-run]` as action specification in result records. It also uses `--dry-run <mode>` as an argument, not `--info`. I cannot immediately come up with a scenario for more than a boolean flag, but it would still make sense to adopt the naming from my POV. Ultimately it is pretty much a dry-run.

datalad/interface/clean.py
```diff
@@ -56,32 +57,44 @@ class Clean(Interface):
...
         what=Parameter(
             args=("--what",),
             dest='what',
-            choices=('cached-archives', 'annex-tmp', 'annex-transfer', 'search-index'),
+            choices=('cached-archives', 'annex-tmp', 'annex-transfer',
+                     'search-index'),
             nargs="*",
-            doc="""What to clean.  If none specified -- all known targets are
-            cleaned"""),
+            doc="""What to clean or report on. If none specified -- all known
```
mih, line 81: If `--dry-run` is adopted, this first half of the doc change can be removed again, I think.

datalad/interface/utils.py
```diff
@@ -450,7 +450,7 @@ def generator_func(*_args, **_kwargs):
                     summary_args = (results, action_summary)
                 else:
                     summary_args = (results,)
-                wrapped_class.custom_result_summary_renderer(*summary_args)
+                wrapped_class.custom_result_summary_renderer(*summary_args, **allkwargs)
```
mih, line 453: This change breaks the API. I could not find an actual case where this matters (across extensions), but still worth pointing out. The unification with `custom_result_renderer()` makes sense to me in general.

datalad/core/local/status.py
```diff
@@ -495,3 +495,24 @@ def custom_result_summary_renderer(results):  # pragma: more cover
...
+        refds = Dataset(refds.pop())
+        recursive = kwargs.pop('recursive', False)
+        recursion_limit = kwargs.pop('recursion_limit', None)
+        clean_infos = refds.clean(info=True,
+                                  recursive=recursive,
+                                  recursion_limit=recursion_limit,
```
mih, line 510: I admittedly did not think about recursion until now. But maybe we should:

```
/ds/hcp/super (git)-[master] % datalad clean -r 
action summary:
  clean (notneeded: 18000)
datalad clean -r  22.20s user 13.67s system 22% cpu 2:38.55 total
```

I think this calls for some kind of switch/config setting, or we decide that switching the renderer to something else is good enough. However, worth acknowledging explicitly one way or another.

datalad/core/local/status.py
```diff
@@ -495,3 +495,24 @@ def custom_result_summary_renderer(results):  # pragma: more cover
...
+                                  return_type='list')
+        if any(r['action'] == 'clean-info' and r['status'] == 'ok'
+               for r in clean_infos):
+            # If status reported on cleanable temp locations, add a hint to
+            # datalad-clean:
+            from datalad.ui import ui
+            ui.message("Note: You can remove temporary files anytime by using "
+                       "datalad's 'clean' command.")
```
mih, line 518: We should have a style guide, as each time I am unsure what tone to use. Passive, active, imperative?

What about simply `Temporary data can be removed with DataLad's 'clean' command`?

I personally think these messages should be as concise and interpretation independent as possible.

datalad/customremotes/archives.py
```diff
@@ -393,8 +394,16 @@ def _transfer(self, cmd, key, path):
                 # so
                 pwd = getpwd()
                 lgr.debug(u"Getting file {afile} from {akey_path} while PWD={pwd}".format(**locals()))
+                was_extracted = self.cache[akey_path].is_extracted
                 apath = self.cache[akey_path].get_extracted_file(afile)
                 link_file_load(apath, path)
+                if not was_extracted and self.cache[akey_path].is_extracted:
+                    self.info("%s special remote is using an extraction cache "
+                              "under %s. You may want to call 'datalad clean' "
+                              "to remove it afterwards." %
```
mih, line 403:
```suggestion
                    self.info("%s special remote created an extraction cache "
                              "under %s. Remove it with DataLad's 'clean' command "
                              "to save disk space." %
```

There can be performance loss when it is removed. I would prefer concrete advice when to do what, rather than a more nebulous statement.

mih (commented):
As always, the more one thinks, the more thoughts come out ;-)

Below are a bunch of additional comments. Also this PR has already shed some light for me personally. I am afraid, this kind of output will be with us frequently:

```
% datalad clone https://github.com/psychoinformatics-de/studyforrest-data-structural
install(ok): /tmp/studyforrest-data-structural (dataset)
% cd studyforrest-data-structural
studyforrest-data-structural (git)-[master] % datalad
studyforrest-data-structural (git)-[master] % datalad get sub-01/anat/sub-01_T1w.nii.gz
get(ok): sub-01/anat/sub-01_T1w.nii.gz (file) [from mddatasrc...]                                                          
studyforrest-data-structural (git)-[master] % datalad status
.git/annex/transfer: Discovered 1 annex temporary transfer directory: download
Note: You can remove temporary files anytime by using datalad's 'clean' command.
```

datalad/core/local/status.py
mih: Git does not seem to use 'hint:' or 'note:' markup:

```

[...]

It took 13.57 seconds to enumerate untracked files. 'status -uno'
may speed it up, but you have to be careful not to forget to add
new files yourself (see 'git help status').
```

datalad/core/local/status.py
bpoldrack: True. Actually only wanted to see whether that holds throughout the test battery. Replacement kinda depends on the outcome. Either `return` or failures point to how I got it wrong and can adjust.

datalad/interface/clean.py
bpoldrack: No strong opinion on my end. I'm fine with `--dry-run`.

datalad/interface/clean.py
bpoldrack: Agree.

datalad/core/local/status.py
mih: @adswa used yet another (establish) way to give hints https://github.com/datalad/datalad/pull/5726#issuecomment-860619676

which should declare one as *the one*.

datalad/interface/utils.py
bpoldrack: Sufficient to point it out in the changelog or do you think this actually needs some sort of "technical" deprecation?

datalad/core/local/status.py
bpoldrack: I wondered about that, too. If the reporting should become configurable the "problem" appears to be the default. On the one hand: That hint is kinda important to be there by default - if you know you want to enable it, you probably don't really need it. On the other hand: Being more expensive by default is kinda wrong, too.

datalad/core/local/status.py
bpoldrack: > We should have a style guide, as each time I am unsure what tone to use. Passive, active, imperative?

Agree. I struggle everytime.

datalad/customremotes/archives.py
bpoldrack: Yeah, I see that. However, this is the special remote itself. Technically it's kinda wrong to point to datalad commands to begin with + long message would be there all the time. Wondering: We usually call annex with `--json` by now, giving us an info object that we decide to spit out (rather than annex itself). So, we could use that for a short message that we catch in datalad and can potentially decide to enhance or swallow from within a context where we know much more about the situation than within the special remote. Not too clear to me at this point, just thinking about principles how to deal with messaging better in general.

datalad/core/local/status.py
bpoldrack: > @adswa used yet another (established?) way to give hints #5726 (comment)

Not established until merged, I'd say ;-)
However, agree - like the result record enhancement much better. Will use that, too.

datalad/core/local/status.py
```diff
@@ -464,7 +464,7 @@ def custom_result_renderer(res, **kwargs):  # pragma: more cover
                 ac.color_word(type_, ac.MAGENTA) if type_ else '')))
 
     @staticmethod
-    def custom_result_summary_renderer(results):  # pragma: more cover
+    def custom_result_summary_renderer(results, **kwargs):  # pragma: more cover
```
yarikoptic, line 467: FWIW, looking at the diff I am already confused on how  it was
```
 def custom_result_summary_renderer(results, action_summary):
```
in `push` and just 
```
def custom_result_summary_renderer(results):
```
here was supposed to work (different signatures), and with addition of `**kwargs` (exemplified by absent docstrings describing expected parameters) it just makes them even more of "dark magic" to figure out what they expect to be passed into them.

If we are to make those signatures "flexible", as one renderer might want `recursive` and another not, may be we should at least aim to list expected kwargs listed explicitly, and have some `**_` or alike to swallow those it doesn't care about but which might be passed, with `**_` vs `**kwargs` to hint that "I do not care about other options".

datalad/core/local/status.py
```diff
@@ -495,3 +495,24 @@ def custom_result_summary_renderer(results):  # pragma: more cover
...
+        # ### Report on possibly cleanable temporary locations
+
+        # Infer the dataset we ran on from the results and get original
+        # recursive (+limit) options:
+        refds = set(r['refds'] for r in results)
+        assert len(refds) == 1
+        refds = Dataset(refds.pop())
+        recursive = kwargs.pop('recursive', False)
+        recursion_limit = kwargs.pop('recursion_limit', None)
+        clean_infos = refds.clean(info=True,
```
yarikoptic, line 508: I really don't like such an approach:
- "renderer" which is also a "do house-keeping" action, possibly even failing along its operation, is IMHO "too much"
- `status` operation could have been limited to some subdatasets, and how this would go through the entire hierarchy of 1000s I have? thank you not
- in my view `status` should be a read-only operation, and cleanup shouldn't be its side-effect
- despite it might be a nice thing to have for some cases, but clearly not for all (e.g. I might not like to cleanup anything).

yarikoptic: d'oh -- ok, blind me now mentioned this new `info=True` so operation would be a read-only, crossed out some items ;-).  Still not sure about the last one wondering if it is wise to add more extra run time on top of already not that speedy `status` when operating on heavy hierarchies.

bpoldrack: > status operation could have been limited to some subdatasets, and now this would go through the entire hierarchy of 1000s I have?

You mean it should consider `path` passed into `status` in addition to `recursion_limit`, @yarikoptic ?

bpoldrack: Re "extra run time":

Consider responding to https://github.com/datalad/datalad/pull/5738#discussion_r650906698? ;-)

yarikoptic: > You mean it should consider `path` passed into `status` in addition to `recursion_limit`, @yarikoptic ?

well, it sounds like that -- or may be any of those extra kwargs aren't needed?  may be paths for relevant subdatasets could be taken from the `results` records passed into `status`? then `clean` would be checked on those datasets (without any recursion)

bpoldrack: >  may be paths for relevant subdatasets could be taken from the results records passed into status? 

True. Will do that. Makes the signature change independent from this PR (still think a unification w/ result renderer makes sense).

yarikoptic: continuing on "renderer" taking on new tasks, I wondered: wouldn't it be better to have `status` code to actually do the clean test internally and just enhancing the returned/yielded dataset status record with e.g. extra field
```
"extra_statuses": {
  "clean": [ ... records from clean  ... ]
}
```
or alike, which we could later enhance possibly with other "hints" or information (aren't we getting more info from `git annex info` which might just get here into `annex-info` ), and then leave it to renderer to just take care about rendering them (instead of running those extra checks)?

datalad/core/local/status.py
adswa: I "established" it in #4338 I think, but simply copied Git ;-)

bpoldrack: Huh. I forgot that. :)

datalad/core/local/status.py
bpoldrack: @yarikoptic  Hm. Two things speak against that from my POV:
1. After @mih chimed in, I explicitly changed from yielding anything about that to only call `clean --info` from within the summary renderer. See https://github.com/datalad/datalad/pull/5738#pullrequestreview-682620241. I think the argument makes sense.
2. There are clean results coming from a direct invocation of `clean`. I think it's strange that - when called from within `status` - you'd have the info in an extra field of the status-result, while when calling directly you get it as `clean-info` results. That seems inconsistent with how we deal with results and such an approach would generally make it harder to act upon results via hooks - you'd need know all the forms in which it can appear. So, I think if it's something that genuinely yielded from another command it should stay that way. Otherwise I'm fine with extra fields.

datalad/core/local/status.py
bpoldrack: I don't think listing expected arguments in a docstring (of that function) makes a lot of sense. Those functions are called from only one internal place. Furthermore it's not a bout a specific function but all of them. It's an interface - "what do I need to implement". So, may be the renderers should just become classes derived from an abstract class that defines the signatures. That would be a place to document it ... WDYT?

yarikoptic: recording my thoughts from our discussion during weekly meeting: such interface(s) should probably established/documented within `Interface` class itself.  But my comment (if still to be used) on `**kwargs` vs `**_` stands .  But defining interfaces etc probably shouldn't be a part of this PR.

Also, this `**kwargs` here is really the `**allkwargs` (from https://github.com/datalad/datalad/pull/5738/files#diff-37fba60e98afe87a28f73afa6b5bbc3410bad59901363e26256da5584cec265cR331) which, if I got it right, contains all arguments passed into the call.  
By passing it like it is done in this PR via `**kwargs`, would forbid to have `results` kwarg for an `Interface` `__call__`, and makes it really unclear from just looking at the code -- what are those (hence instigated my comment on docstrings, and discussion on interfaces).

I think a better strategy would be to not `**` them and pass as a dedicated `call_kwargs` argument.  This would then:
- not shadow `results` or any other argument to renderers we ever decide to add
- make it explicit on what is that thing we are passing in

datalad/interface/utils.py
yarikoptic: Do we have  a test which would go through all `Interface`'s and test that invocation in command line works as expected? if not -- we wouldn't even notice until we break things up.

Also see my comment [above](https://github.com/datalad/datalad/pull/5738/files#r654449041): I think they should be passed as 

```suggestion
                wrapped_class.custom_result_summary_renderer(*summary_args, call_kwargs=allkwargs)
```
here.

Also, in principle, we could use then `inspect` (and cache the check) if `custom_result_summary_renderer` expects `call_kwargs` or has a swallow all `**` (if that is possible to check -- didn't look): if not --  issue a DeprecationWarning to alert that interface is changing.

datalad/core/local/status.py
bpoldrack: Changed to just issue a debug log message and proceed with the first discovered refds (which now is only needed to retrieve a config).

datalad/interface/clean.py
bpoldrack: Done

datalad/interface/clean.py
bpoldrack: Done

bpoldrack:
@mih @yarikoptic :

Since I didn't see a good solution to performance issues as discussed above, changed to the following: Simply let `status` display a hint to `clean` in the summary by default - no actual probing is performed. Hint can be disabled by a config. This should address the UX aspect, as users become aware that way.

If that's fine with you guys, will edit PR+first post accordingly.
I would not close the original issue, since autocleaning would still be a desirable option, but just take out the UX label and consider this done for the UX project. Agree, @mih?

Otherwise one question remains for me: `datalad.status.hints.clean` may not be ideal. If we want to generalize that approach, we should decide on the config "hierarchy": Are such flags to be collected per command (as I did here), or should we go for a broader collection like `datalad.hints.status-clean`?

mih:
I cannot get behind some of the rational that led to the current state of the PR, but in its present form I would prefer to not merge it. From my POV it adds to the problem of silently growing repositories the problem of output noise:

```
% datalad create sofresh
[INFO   ] Creating a new annex repo at /tmp/sofresh 
create(ok): /tmp/sofresh (dataset)

% datalad -C sofresh status
Hint: Run 'datalad clean [--dry-run]' to find and remove temporary files used by datalad and git-annex.
Disable 'datalad.status.hints.clean' to not display this message anymore.
```

If we ignore that the hint is visually masking the important information, users will see this unconditionally every time they execute status (which is arguably one, if not the, most frequently executed command). They will have learned to ignore it (or disable it), before they actually get into the situation that `datalad clean --dry-run` would report anything cleanable.

Given that the PR does not longer address the original issue, I'd say that the modification of `status` should be reverted, and only the updates to `clean` be kept. I wanted to look into how the messaging of the new archive remote behaves, but I tripped over https://github.com/datalad/datalad/issues/5800

bpoldrack:
So, just `clean --dry-run` and a message from `datalad-archives` special remote when it creates an extraction cache.
The latter looks like this ATM:

```
❱ datalad get .
[INFO   ] To obtain some keys we need to fetch an archive of size 2.6 kB                                                                         
[INFO   ] datalad-archives special remote is using an extraction cache under /tmp/clone/.git/datalad/tmp/archives/34d49c45cd. Remove it with DataLad's 'clean' command to save disk space. 
get(ok): .ssh/id_appveyor_rsa (file) [from datalad-archives...]
get(ok): .ssh/id_appveyor_rsa.pub (file) [from datalad-archives...]
get(ok): appveyor_key.tar.gz (file) [from origin...]
action summary:
  get (ok: 3)
```

adswa (commented):
I haven't looked into this in detail, but came here to say that I like the UX message where it is put now much more than unconditionally within ``status``.
I also like the addition of a ``dry-run`` flag. 

There were a few things I noticed:
When I run it in a random directory without a cache, it just returns nothing. What do you think: would it make sense to report back a "cacheless" state, e.g., "No temporary clutter found"? This behavior would somehow match that of ``status`` (which reports ``nothing to save, working tree clean``) and ``rerun`` (which warn ``No command given`` when nothing would be run). But it would be different from, e.g. ``git clean -dfn``, which also returns nothing if nothing would be cleaned away.

I tried to trigger the hint by downloading and extracting an archive, but didn't see the hint even though there is a temporary download directory left behind:

```
(handbook2) adina@muninn in /tmp/abcd on git:master
❱ datalad download-url --archive -m test https://www.7-zip.org/a/7za920.zip
[INFO   ] Downloading 'https://www.7-zip.org/a/7za920.zip' into '/tmp/abcd/' 
download_url(ok): /tmp/abcd/7za920.zip (file)                                   
add(ok): 7za920.zip (file)                                                      
save(ok): . (dataset)                                                           
[INFO   ] Adding content of the archive /tmp/abcd/7za920.zip into annex AnnexRepo(/tmp/abcd) 
[INFO   ] Initiating special remote datalad-archives 
[INFO   ] Finished adding /tmp/abcd/7za920.zip: Files processed: 4, +annex: 4 
action summary:
  add (ok: 1)
  download_url (ok: 1)
  save (ok: 1)
(handbook2) adina@muninn in /tmp/abcd on git:master
❱ datalad clean --dry-run
.git/annex/transfer: Discovered 1 annex temporary transfer directory: download
```

It may make sense to add it here, too?

I think it would be nice to add ``--dry-run`` to the examples of the command, even if it is kinda straightforward.

bpoldrack:
> I tried to trigger the hint by downloading and extracting an archive, but didn't see the hint even though there is a temporary download directory left behind

That's a different thing, not the extraction cache of the special remote, though.
`.git/annex/transfer/...` is created by git-annex, not by datalad. So, we can't really issue such a hint when it's happening. We'd need to investigate whether annex left something behind, thus running into similar issues as with the original approach here - probing can be costly.


Re no result:
1. FTR - There should be `notneeded` results (which aren't rendered by default)
2. I guess I'm with you - should behave like `status`. Will do.

mih:
> That's a different thing, not the extraction cache of the special remote, though.

I see 

```
% tree -d .git/datalad
.git/datalad
└── tmp
    └── archives
```

after `download-url`, but no report on it, and `clean` not acting on it without `--dry-run`.

bpoldrack:
> I see 

> ```
> % tree -d .git/datalad
> .git/datalad
> └── tmp
>     └── archives
> ```


@mih : I see that, too. However, it's empty. So, there's nothing to clean.

bpoldrack:
@mih , @adswa :

Added a "nothing-to-do" message, added examples, `clean` now deals with empty directories as well and switched to make use of `Path` objects instead.

```
/tmp/some on git:master❱ datalad clean --dry-run
.git/datalad/tmp/archives: Discovered empty temporary archive directory
.git/annex/tmp: Discovered empty temporary annex directory
.git/annex/transfer: Discovered 1 annex temporary transfer directory: download
/tmp/some on git:master❱ datalad clean
clean(ok): .git/datalad/tmp/archives (directory) [Removed empty temporary archive directory]
clean(ok): .git/annex/tmp (directory) [Removed empty temporary annex directory]
clean(ok): .git/annex/transfer (directory) [Removed 1 annex temporary transfer directory: download]
/tmp/some on git:master❱ datalad clean --dry-run
nothing to clean, no temporary locations present.
```

adswa:
I like it! 
But one of your tests fails on Windows:

```
ERROR: datalad.interface.tests.test_clean.test_clean
    self.test(*self.arg)
  File "C:\Python39-x64\lib\site-packages\datalad\tests\utils.py", line 739, in _wrap_with_tempfile
  File "C:\Python39-x64\lib\site-packages\datalad\interface\tests\test_clean.py", line 55, in test_clean
    res = clean(dataset=ds, return_type='item-or-list',
  File "C:\Python39-x64\lib\site-packages\datalad\interface\utils.py", line 483, in eval_func
  File "C:\Python39-x64\lib\site-packages\datalad\interface\utils.py", line 476, in return_func
    results = list(results)
  File "C:\Python39-x64\lib\site-packages\datalad\interface\utils.py", line 396, in generator_func
    for r in _process_results(
  File "C:\Python39-x64\lib\site-packages\datalad\interface\utils.py", line 574, in _process_results
    for res in results:
  File "C:\Python39-x64\lib\site-packages\datalad\interface\clean.py", line 173, in __call__
    rmtree(topdir)
  File "C:\Python39-x64\lib\site-packages\datalad\utils.py", line 476, in rmtree
    path = r'\\?\ '.strip() + path
TypeError: can only concatenate str (not "WindowsPath") to str
````

**Comment by bpoldrack** [comment]
Ah - stupid me. Force-pushed.

**Comment by adswa** [comment]
Cool, works for me locally on Windows now!

**Review by mih — CHANGES_REQUESTED** [review]
I resolved half a dozen conversation on topics that are no longer associated with the changes proposed in here. It would probably have been better to close this PR and start a new one, instead. It is needlessly hard to review these winding stories.

Given that this PR closes #5834 I think a key change is the special remote info message. However, there is no test for it actually showing up in user output, and I cannot make it show manually either. 

```
% datalad create archive
[INFO   ] Creating a new annex repo at /tmp/archive 
create(ok): /tmp/archive (dataset)
% cd archive
% datalad download-url --archive -m test https://www.7-zip.org/a/7za920.zip
[INFO   ] Downloading 'https://www.7-zip.org/a/7za920.zip' into '/tmp/archive/' 
download_url(ok): /tmp/archive/7za920.zip (file)                                           
add(ok): 7za920.zip (file)                                                                 
save(ok): . (dataset)                                                                      
[INFO   ] Adding content of the archive /tmp/archive/7za920.zip into annex AnnexRepo(/tmp/archive) 
[INFO   ] Initiating special remote datalad-archives 
[INFO   ] Finished adding /tmp/archive/7za920.zip: Files processed: 4, +annex: 4 
action summary:
  add (ok: 1)
  download_url (ok: 1)
  save (ok: 1)
(datalad-dev) mih@meiner /tmp/archive (git)-[master] % git annex info
...
        c04eb54b-4b4e-5755-8436-866b043170fa -- [datalad-archives]
```

Is it not working? Or is it not supposed to be working in this case?

**Comment by bpoldrack** [comment]
> Is it not working? Or is it not supposed to be working in this case?

Not supposed. `download_url` doesn't leave an extraction cache. However, if you drop the extracted content, clone that ds and `get` the content from origin, the cache should be established and the message be issued.

Will add a test.

**Comment by mih** [comment]
> Not supposed. `download_url` doesn't leave an extraction cache. However, if you drop the extracted content, clone that ds and `get` the content from origin, the cache should be established and the message be issued.

Thanks for the clarification. I can confirm that it does not need a clone to see it working:

(continued from above)
```
(datalad-dev) mih@meiner /tmp/archive (git)-[master] % datalad drop license.txt
drop(ok): /tmp/archive/license.txt (file)                                                  
(datalad-dev) mih@meiner /tmp/archive (git)-[master] % datalad get license.txt
[INFO   ] datalad-archives special remote is using an extraction cache under /tmp/archive/.git/datalad/tmp/archives/5ca17613fa. Remove it with DataLad's 'clean' command to save disk space. 
get(ok): license.txt (file) [from datalad-archives...]
```

bpoldrack:
> I can confirm that it does not need a clone to see it working:

Yeah, just reread my own comment and noticed the cloning was nonsense. 

Pushed a test enhancement.

mih (approved):
Thx! From my POV this is ready.

Issue #5711, opened 2021-06-07
`create-sibling`(`-github`) result renderer does not mention sibling name


mih:
```
% datalad create-sibling-github -d . kdenlive-gource-ukbiobank  
[INFO   ] Successfully obtained information about organization psychoinformatics-de using token ghp... 
create_sibling_github(ok): . (dataset) [project at git@github.com:psychoinformatics-de/kdenlive-gource-ukbiobank.git]
configure-sibling(ok): . (sibling)
action summary:
  configure-sibling (ok: 1)
  create_sibling_github (ok: 1)
```

The `configure-sibling` result rendering could mention the name of the sibling (the default `github` in this case). ATM I have to investigate, or remember "how it does it".

adswa:
fixed in #5744

PR #5744, opened 2021-06-14
UX: Report the sibling name in create-sibling-github's results record

adswa:
Here's my attempt at reporting a sibling name when a GitHub sibling is created, as desired in #5711. Its not a separate key in the result, but merely added into the message, but I found that it looked quite crowded when I rendered several messages.
Here is the current look:
Screenshot from 2021-06-14 15-37-24


bpoldrack (approved):
LGTM.

Just would like to see a different title since it appears in the changelog. THis is not about result rendering. You just changed the message in the possibly-to-be-rendered result record.

mih (approved):
Cool, thx!

PR #5747, opened 2021-06-14
BF(TST): parallel - take longer for producer to produce

### LINKED ISSUE #5746: test_gracefull_death: Assertion error
continuation to #5742 found while testing #5735 with , while cmd:annex: 8.20201127 but not sure if related to having -J5 there or not

https://ci.appveyor.com/project/mih/datalad/builds/39550675/job/ns726sivdltwl9tu

```

FAIL: datalad.support.tests.test_parallel.test_gracefull_death
  File "/home/appveyor/dlvenv/lib/python3.8/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/appveyor/dlvenv/lib/python3.8/site-packages/datalad/tests/utils.py", line 266, in _wrap_skip_if
    return func(*args, **kwargs)
  File "/home/appveyor/dlvenv/lib/python3.8/site-packages/datalad/support/tests/test_parallel.py", line 197, in test_gracefull_death
AssertionError: 0 not greater than or equal to 2
```

most likely just fragility of that test

yarikoptic:
so that consumers have a better chance to finish up at least some
jobs (we got assert_greater_equal(len(results), 2) failing), before
the whole thing fails and we raise right away (that is what this
portion of the test tests).

If this does not show to be a reliable workaround, since we do not
use  reraise_immediately=True  mode yet anywhere, I gues we could just
disable this portion of the test.

Closes #5746

### DIFF
--- a/datalad/support/tests/test_parallel.py
+++ b/datalad/support/tests/test_parallel.py
@@ -179,7 +179,7 @@ def test_gracefull_death():
 
     def producer():
         for i in range(10):
-            sleep(0.0001)
+            sleep(0.0003)
             yield i
         raise ValueError()
     # by default we do not stop upon producer failing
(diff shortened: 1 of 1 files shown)


yarikoptic:
eh, in case of the heavy logging we still get it to fail
```
FAIL: datalad.support.tests.test_parallel.test_gracefull_death
  File "/tmp/dl-miniconda-f10lrp_8/lib/python3.8/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/tmp/dl-miniconda-f10lrp_8/lib/python3.8/site-packages/datalad/tests/utils.py", line 266, in _wrap_skip_if
    return func(*args, **kwargs)
  File "/tmp/dl-miniconda-f10lrp_8/lib/python3.8/site-packages/datalad/support/tests/test_parallel.py", line 202, in test_gracefull_death
    assert_greater_equal(8, len(results))
AssertionError: 8 not greater than or equal to 9
```
... with that effect -- I feel that I better just disable (I will comment out) that portion of the test so it stops annoying us ;-)

yarikoptic:
to pacify `Test for semver label / test (pull_request) ` where we test only for `patch` or `internal` and not `tests` (attn @jwodder), I added `patch`.  Let's see now what would happen in the release workflow upon merge!  BRACE FOR IMPACT!

PR #5749, opened 2021-06-16
Clone URL mapping ability

### LINKED ISSUE #4822: clone/install: introduce url mapping convenience?

mih:
This implements a simple feature and fix for #4822. It allows for configuring an arbitrary number of URL mapping specification. With this change one can configure datalad to transfer URLs like `https://osf.io/q8xnk/` (copied from a browser window), which do not work, but can be reliably mapped into working URLs (`osf://q8xnk` in this case) via a regular expression specification.

In the example above, this would do the trick:

```
datalad -c 'datalad.url.substitute.osf=,^https://osf.io/([^/]+)[/]*$,osf://\1' clone https://osf.io/q8xnk/
```

as usual, configuration is read from all supported places. Here is an example of a mapping rule
for GitHub

```
% datalad x-configuration
...
# GitHub URL substitution rule
# Substitution specification defined as a string with a match and
# substitution expression, each following Python's regular expression
# syntax. Both expressions are concatenated to a single string with an
# arbitrary delimiter character. The delimiter is defined by prefixing
# the string with the delimiter. Prefix and delimiter are stripped
# from the expressions (Example: ",^http://(.*)$,https://\1"). This
# setting can be defined multiple times. Substitutions will be applied
# incrementally, in order of their definition. The first substitution
# in such a series must match, otherwise no further substitutions in a
# series will be considered. However, following the first match all
# further substitutions in a series are processed, regardless whether
# intermediate expressions match or not.
datalad.clone.url-substitute.github=(',https?://github.com/([^/]+)/(.*)$,\\1###\\2', ',[/\\\\]+,-', ',\\s+|(%2520)+|(%20)+,_', ',([^#]+)###(.*),https://github.com/\\1/\\2')
...
```

It would be possible to implement a mechanism to read additional specifications from an entrypoint that can be provided by extensions. However, it would make also sense to finally support extension-provided (default) configuration and use that mechanism.

TODO:

- [x] Implement sketch for discussion
- [x] Documentation
- [x] Test
- [x] should it rather be `datalad.clone.url-substitute`? (to limit the scope)
- [x] Institutionalize OSF-related mapping from example above?

Closes #4822

### DIFF
--- a/datalad/core/distributed/clone.py
+++ b/datalad/core/distributed/clone.py
@@ -45,6 +45,7 @@ from datalad.support.constraints import (
 )
 from datalad.support.exceptions import DownloadError
 from datalad.support.param import Parameter
+from datalad.support.strings import get_replacement_dict
 from datalad.support.network import (
     get_local_file_url,
     download_url,
@@ -60,6 +61,7 @@ from datalad.dochelpers import (
 )
 from datalad.utils import (
     ensure_bool,
+    ensure_list,
     knows_annex,
     make_tempfile,
     Path,
@@ -119,6 +121,18 @@ class Clone(Interface):
     an installed dataset, regardless of whether it was newly installed ('ok'
     result), or found already installed from the specified source ('notneeded'
     result).<< PYTHON ||
+
+    URL mapping configuration
+
+    'clone' supports the transformation of URL via (multi-part) substitution
+    specifications. Substitutions are defined as a string with a match and
+    substitution expression, each following Python's regular expression syntax.
+    Both expressions are concatenated to a single string with an arbitrary
+    delimiter character. The delimiter is defined by prefixing the string with
+    the delimiter. Prefix and delimiter are stripped from the expressions.
+    Example: ,^http://(.*)$,https://\\1
+    TODO continue once match preference (longest match?) is decided and
+    implemented.
 
     .. seealso::
 
@@ -349,6 +363,63 @@ class Clone(Interface):
                 ds.subdatasets(path, set_property=[("datalad-url", source)])
 
 
+def _get_url_mappings(cfg):
+    return [
+        # decode the rule specifications
+        get_replacement_dict(
+            # one or more could come out
+            ensure_list(
+                cfg.get(
+                    k,
+                    default=[],
+                    # we specifically support declaration of multiple
+                    # settings to build replacement chains
+                    get_all=True)))
+        for k in cfg.keys()
+        if k.startswith('datalad.clone.url-substitute.')
+    ]
+
+
+def _map_urls(cfg, urls):
+    mapping_specs = _get_url_mappings(cfg)
+    if not mapping_specs:
+        return urls
+
+    mapped = []
+    # we process the candidate in order to maintain any prioritization
+    # encoded in it (e.g. _get_flexible_source_candidates_for_submodule)
+    # if we have a matching mapping replace the URL in its position
+    for u in urls:
+        # we only permit a single match
+        # TODO we likely want to RF this to pick the longest match
+        mapping_applied = False
+        # try one mapping set at a time
+        for mapping_spec in mapping_specs:
+            # process all substitution patterns in the specification
+            mu = u
+            matched = False
+            for match_ex, subst_ex in mapping_spec.items():
+                if not matched:
+                    matched = re.match(match_ex, mu) is not None
+                if not matched:
+                    break
+                # try to map, would return unchanged, if there is no match
+                mu = re.sub(match_ex, subst_ex, mu)
+            if mu != u:
+                lgr.debug("URL substitution: '%s' -> '%s'", u, mu)
+                mapped.append(mu)
+                # we could consider breaking after the for effective mapping
+                # specification. however, that would mean any generic
+                # definition of a broadly matching substitution would derail
+                # the entroe system. moreover, suddently order would matter
+                # substantially
+                mapping_applied = True
+        if not mapping_applied:
+            # none of the mappings matches, go with the original URL
+            mapped.append(u)
+    return mapped
+
+
 def clone_dataset(
         srcs,
         destds,
@@ -407,6 +478,11 @@ def clone_dataset(
         reckless = cfg.get('datalad.clone.reckless', None)
 
     dest_path = destds.pathobj
+
+    # check for configured URL mappings, either in the given config manager
+    # or in the one of the destination dataset, which is typically not existent
+    # yet and the process config manager is then used effectively
+    srcs = _map_urls(cfg or destds.config, srcs)
 
     # decode all source candidate specifications
     candidate_sources = [decode_source_spec(s, cfg=cfg) for s in srcs]
--- a/datalad/core/distributed/clone.py
+++ b/datalad/core/distributed/clone.py
@@ -124,13 +124,21 @@ class Clone(Interface):
 
     URL mapping configuration
 
-    'clone' supports the transformation of URL via (multi-part) substitution
-    specifications. Substitutions are defined as a string with a match and
-    substitution expression, each following Python's regular expression syntax.
-    Both expressions are concatenated to a single string with an arbitrary
-    delimiter character. The delimiter is defined by prefixing the string with
-    the delimiter. Prefix and delimiter are stripped from the expressions.
-    Example: ,^http://(.*)$,https://\\1
+    'clone' supports the transformation of URLs via (multi-part) substitution
+    specifications. Substitution specification is defined as a configuration
+    setting 'datalad.clone.url-substition.<seriesID>' with a string containing
+    a match and substitution expression, each following Python's regular
+    expression syntax. Both expressions are concatenated to a single string
+    with an arbitrary delimiter character. The delimiter is defined by
+    prefixing the string with the delimiter. Prefix and delimiter are
+    stripped from the expressions (Example: ",^http://(.*)$,https://\\1").
+    This setting can be defined multiple times, using the same '<seriesID>'.
+    Substitutions will be applied incrementally, in order of their definition.
+    The first substitution in such a series must match, otherwise no further
+    substitutions in a series will be considered. However, following the
+    first match all further substitutions in a series are processed,
+    regardless whether intermediate expressions match or not.
+
     TODO continue once match preference (longest match?) is decided and
     implemented.
 
@@ -364,6 +372,13 @@ class Clone(Interface):
 
 
 def _get_url_mappings(cfg):
+    cfg_prefix = 'datalad.clone.url-substitute.'
+    # figure out which keys we should be looking for
+    # in the active config
+    subst_keys = set(k for k in cfg.keys() if k.startswith(cfg_prefix))
+    # and in the common config specs
+    from datalad.interface.common_cfg import definitions
+    subst_keys.update(k for k in definitions if k.startswith(cfg_prefix))
     return [
         # decode the rule specifications
         get_replacement_dict(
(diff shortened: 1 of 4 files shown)


datalad/core/distributed/clone.py
```diff
@@ -410,6 +410,33 @@ def clone_dataset(
...
+        for s in srcs:
+            for match_ex, subst_ex in mspecs.items():
+                # try to map, would return unchanged, if there is no match
+                mapped = re.sub(match_ex, subst_ex, s)
+                if mapped != s:
+                    _srcs.append(mapped)
+                    # MIH: unclear if readding the original URL as a fallback,
+                    # in case the mapping was stupid is desirable.
+                    # if the candidate URL was hopeless to begin with, it would
+                    # result in two failed attempts instead of one
```
bpoldrack, line 434: Re comment - I agree. Don't "fix" explicitly given configs. However, we could use the hint mechanism to point out, that a candidate failed that was the result of such a replacement.

bpoldrack: To be clear: By agreeing I mean: Do NOT add the original one.

mih: > Re comment - I agree. Don't "fix" explicitly given configs. However, we could use the hint mechanism to point out, that a candidate failed that was the result of such a replacement.

I am not so sure about that. Config could come from all kinds of places, also sources that are not under a user's control. If any of those substitutions is too loose, what then? File a bug and wait for a release?

bpoldrack: > If any of those substitutions is too loose, what then? File a bug and wait for a release?

Put a high-priority noop substitution in your confgi? ;-)
Pointing to one issue: If the config is read from everywhere (as it should), explicit priorities seem to be needed (rather than just implicitly by `git-config`-return order.

yarikoptic:
> * should it rather be `datalad.clone.url-substitute`? (to limit the scope)

FWIW - +1 on that. I just wonder if we would have some other 'clone.url' related config options (`.backends`, etc) and might want to make it `datalad.clone.url.substitute` although we do have only few of such 4 level'ed configs and generally they are more of the 3 level form with last level composing into a single word.

yarikoptic:
> > * should it rather be `datalad.clone.url-substitute`? (to limit the scope)
> 
> FWIW - +1 on that. ...

FWIW, I thought that may be we could/should parallel how git does it, e.g. I have
```
[url "git@github.com:"]
    pushinsteadOf = https://yarikoptic@github.com/
    pushinsteadOf = https://github.com/
    pushinsteadOf = http://github.com/
    pushInsteadOf = git://github.com/
```
but trying it out, with 
```
[datalad "url.osf://\1"]
    cloneinsteadOf = "https://osf.io/([^/]*)/?"
```

points to shortcoming of swallowed backslash

```
$> git config -f /tmp/urls.config --get-regex 'url\.*'
datalad.url.osf://1.cloneinsteadof https://osf.io/([^/]*)/?
```

but also if target pattern has `.` in it, I guess splitting apart would also be tricky

```shell
$> cat /tmp/urls.config
[datalad "url.http://something.com/\1/download"]
    cloneinsteadOf = "https://buga.io/([^/]*)/?"

$> git config -f /tmp/urls.config --get-regex 'url\.*'
datalad.url.http://something.com/1/download.cloneinsteadof https://buga.io/([^/]*)/?
```
although we could always assume that it is `datalad.url.<str>.whatinsteadof` and thus `.rsplit(1)` everything after `datalad.url.`

mih:
#### re config style @yarikoptic

I was toying with the idea to adopt that pattern a bit more

```config
[url "osf://\\1"]
        datalad-clone-insteadOf = "https://osf.io/([^/]*)/?"
```

```sh
% git config -l | grep osf
url.osf://\1.datalad-clone-insteadof=https://osf.io/([^/]*)/?
```

(also note the escaped backslash)

I don't see why we could not use the `url.` namespace directly in this fashion, and it would make things look nicer and better grouped, per URL (i.e. github handling, OSF handling ...).

The counter argument is that we need/require(?) a different URL specification style -- which actually makes it a similar, but different beast.

However, which could extend (instead of replacing) the specification, such that if we see something like this:

```
[url "git@github.com:"]
    datalad-clone-insteadof = "https://yarikoptic@github.com/"
```

we replace it internally with:

```
[url "git@github.com:\\1"]
    datalad-clone-insteadof = "https://yarikoptic@github.com/(.*)$"
```

i.e. mimic the base URL replacement done by Git.

#### re priorities @bpoldrack 

> Pointing to one issue: If the config is read from everywhere (as it should), explicit priorities seem to be needed (rather than just implicitly by git-config-return order. I think that is true independently of the question at hand. There could be conflicting substitutions and you don't want to try the wrong one first everytime.

I am struggling to come up with a real use case for this -- if you had some in mind, please share. If there is no real use case, I think we should not implement prioritization. If we find out later that we need it, we could easily extend the setting name scheme ala `datalad.get.subdataset-source-candidate-00...`:

```
[url "osf://\\1"]
        datalad-clone-insteadof-100onlythisworks = "https://osf.io/([^/]*)/?"
```

My rational for "we don't need this" is:

- for subdataset clone candidates we already have a mechanism that can do this -- no need to have two alternatives -- so we are talking about initial clones only
- for initial clones the cost of a failed attempt is much lower (i.e. not attempted wrongly for each of 10k subdataset, but only once for the root)
- we have a prioritization of config sources already, a system/user/repo-provided suboptimal configuration can already by overridden

So Q is: which use case cannot be handled by the levers that are already in place?

yarikoptic:
> I don't see why we could not use the `url.` namespace directly in this fashion, and it would make things look nicer and better grouped, per URL (i.e. github handling, OSF handling ...).

primarily for consistency since all datalad-specific settings reside under `datalad.` .  What I also wonder -- may be should be a hybrid like

```
[datalad "url.datalad-osf"]
    cloneinsteadOf = ",https://osf.io/([^/]*)/?,url.osf://\1"
```
or generally
```
[datalad "url.<mapid>"]
    cloneinsteadOf = "<substitution-pair>"
```
which might help to avoid any possible oddity from sticking regexes into section names, easier overrides (we would need to disallow multiple entries per section I guess), etc?

mih:
> ```
> [datalad "url.<mapid>"]
>    cloneinsteadOf = "<substitution-pair>"
> ```

TBH, I don't really like it. The only thing is has in common with the git-construct is the name. I do like the `<substitution-pair>` syntax better than what git does. Unless there is a good use of the `mapid`, I think I came almost full circle to `datalad.clone.url-substitute = "<substitution-pair>"`.

mih:
I have updated the PR with a new implementation that is capable of providing a fix analog to https://github.com/datalad/datalad/pull/5709, but without custom service dependent conditional. Instead, we can define and ship a default mapping for Github URL, as demo'd in the included unit test at https://github.com/datalad/datalad/pull/5749/files#diff-9789036616f2c57b1a1882ce86a6dc36cd5611c24e73a17eb1305da8335d98d8R1604

A full documentation update is still pending (which is pending a decision of match prioritization, see notes inside), but maybe this is already enough for you @yarikoptic to get an idea where I am going.

In contrast to #5709 this is a fairly generic feature with relatively wide applicability (e.g. #4822), but pretty much any kind of URL sanitization, redirection, reconfiguration. It also makes my suggestion in https://github.com/datalad/datalad/pull/5709#issuecomment-860260616 obsolete, as it offers the same functionality without the custom building block.

mih:
This is stalled by https://github.com/datalad/datalad/issues/5767

datalad/core/distributed/clone.py
```diff
@@ -351,6 +371,74 @@ def __call__(
...
+    if not mapping_specs:
+        return urls
+
+    mapped = []
+    # we process the candidate in order to maintain any prioritization
+    # encoded in it (e.g. _get_flexible_source_candidates_for_submodule)
+    # if we have a matching mapping replace the URL in its position
+    for u in urls:
+        # we only permit a single match
+        # TODO we likely want to RF this to pick the longest match
```
yarikoptic, line 411: Not sure why longest would be desired

datalad/core/distributed/clone.py
```diff
@@ -351,6 +371,74 @@ def __call__(
...
+            matched = False
+            for match_ex, subst_ex in mapping_spec.items():
+                if not matched:
+                    matched = re.match(match_ex, mu) is not None
+                if not matched:
+                    break
+                # try to map, would return unchanged, if there is no match
+                mu = re.sub(match_ex, subst_ex, mu)
+            if mu != u:
+                lgr.debug("URL substitution: '%s' -> '%s'", u, mu)
```
yarikoptic, line 427: Debug statement! :heart:

datalad/core/distributed/clone.py
```diff
@@ -351,6 +371,74 @@ def __call__(
...
+def _get_url_mappings(cfg):
+    cfg_prefix = 'datalad.clone.url-substitute.'
+    # figure out which keys we should be looking for
+    # in the active config
+    subst_keys = set(k for k in cfg.keys() if k.startswith(cfg_prefix))
+    # and in the common config specs
+    from datalad.interface.common_cfg import definitions
+    subst_keys.update(k for k in definitions if k.startswith(cfg_prefix))
+    # TODO a potential sorting of substitution series could be implemented
+    # here
```
yarikoptic, line 383: I guess we would need to introduce `datalad.clone.url-substitute.<series>.priority` or alike?

yarikoptic (approved):
Didn't try, left comments on TODO comments, but otherwise - love it

bpoldrack:
@mih 

> My rational for "we don't need this" is:
>  -  for subdataset clone candidates we already have a mechanism that can do this -- no need to have two alternatives -- so we are talking about initial clones only
>  -  for initial clones the cost of a failed attempt is much lower (i.e. not attempted wrongly for each of 10k subdataset, but only once for the root)
>  -  we have a prioritization of config sources already, a system/user/repo-provided suboptimal configuration can already by overridden

Generally agree. What I had in mind is admittedly mostly hypothetical. There could be different handlers from different extensions that want to handle URL from some platform. I might then want to curate the order since I know that I'll mostly want A to handle it and only fall back to B if A can't succeed. Something like that. But I guess we can stick with current ways of overriding for now. I'm fine with deferring that until we find it's actually needed.

datalad/core/distributed/clone.py
mih: ;-)

datalad/core/distributed/clone.py
mih: This is what Git does, when deciding which `insteadof` setting to apply. The rational is probably, the longer the match the more specific. In the regex world things will not be that simple, but the analog "more specific before more generic" could still make sense and be implementable.

datalad/core/distributed/clone.py
mih: That, or sorting based on series ID, or inspection of the regex match specification (see "longest match" comment). I think there are multiple possible ways forward.

mih:
Thanks for the comments and the approval. I guess this is good to go now, and we will see when the need for priorities arises.

PR #5757, opened 2021-06-21
BF: update changelog conversion from .md to .rst (for sphinx)

yarikoptic:
Seems to need quite a bit of workarounds to get an .rst version. See diff comment in Makefile -- sphinx isn't that ready for unicode in section headings, eh heh

TODOs
- [x] @jwodder -- we need to call `make update-changelog` (requires `iconv` and `pandoc`) after a new entry for CHANGELOG.md  is generated but before it is committed.  Adjustments to the changelog.rst in docs/ should be committed as well.  If you see how to improve that ugly Makefile rule , feel welcome ;)
- [x] added allowing  for "tests" label without patch or minor

### DIFF
--- a/.autorc
+++ b/.autorc
@@ -8,6 +8,7 @@
         [
             "exec",
             {
+                "afterChangelog": "make update-changelog && git add docs/source/changelog.rst && git commit -m '[skip ci] Update RST changelog'",
                 "afterRelease": "python -m build && twine upload dist/*"
             }
         ]
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -30,6 +30,9 @@ jobs:
       - name: Install Python dependencies
         run: python -m pip install build twine
 
+      - name: Install pandoc
+        run: sudo apt-get update && sudo apt-get install -y pandoc
+
       - name: Create release
         run: ~/auto shipit
         env:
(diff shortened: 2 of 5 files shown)


jwodder:
@yarikoptic Rather than futzing with pull requests on pull requests, I'll just tell you what changes need to be made.

* Modify `.github/workflows/release.yml` to install pandoc before running `~/auto shipit` (iconv is already installed)
* Add the following line to `.autorc` above the `"afterRelease"` line:

        "afterChangelog": "make update-changelog && git add docs/source/changelog.rst && git commit -m '[skip ci] Update RST changelog'",

  It might be cleaner to add the `git add && git commit` to the Makefile rule instead, though.

Also, note the following:
* Assuming that the Unicode problem is entirely due to the emojis in the label sections, an alternative to using iconv would be to modify the labels' `changelogTitle` strings in the configuration.
* If we ever run into the E2BIG error again, I believe it would now occur at the `afterChangelog` stage, killing the release entirely.  The best way around this (as far as I know) would be [this plugin](https://github.com/intuit/auto/issues/1294#issuecomment-643403558), but the devs haven't responded to my comment about how exactly to use it.  I'm going to try it out in a test repository and get back to you.

yarikoptic:
Thank you @jwodder !!! I will meanwhile merge as is, since we might need to kick out release soonish to accompany joss publication.

PR #5760, opened 2021-06-23
BF: Prefer committed subdataset location info over guesses

mih:
gh-5033 describes a situation where the installation of a sub-subdataset
fails due to the particular condition of the _checkout_ the subdataset
was cloned from. The issues originally expressed the desire for a more
in-depth exploration of possible superdatasets and their subdataset
configurations. However, the undesirable failure can also be avoided by
simply preferring the location information that is committed in each
superdataset, at all clone iterations on the way to a target subdataset.

Effectively, the change reorders the locations priorities from

1. Assumed availability within the assumed existing checkout of the
   superdataset
2. Location info from .gitmodules

to

1. Location info from .gitmodules
2. Other candidates and guesses

Specifically, the following URLs will be tried by default:

1. original URL given to datalad when registering the subdataset
2. `url` from .gitmodules
3. other candidates

Changes to this default order are possible via configuration changes, of
course.

Resolves datalad/datalad#5033

### DIFF
--- a/datalad/distribution/tests/test_publish.py
+++ b/datalad/distribution/tests/test_publish.py
@@ -285,6 +285,10 @@ def test_publish_recursive(pristine_origin, origin_path, src_path, dst_path, sub
     # we will be publishing back to origin, so to not alter testrepo
     # we will first clone it
     origin = install(origin_path, source=pristine_origin, recursive=True)
+    # uncouple subdataset from testrepo sources after recursive install
+    # to make this clone the source of all `get` attempts
+    for sub in origin.subdatasets(result_xfm=lambda x: x['gitmodule_name']):
+        origin.subdatasets(path=sub, set_property=[('url', './{}'.format(sub))])
     # prepare src
     source = install(src_path, source=origin.path, recursive=True)
     # we will be trying to push into this later on, need to give permissions...
@@ -366,14 +370,12 @@ def test_publish_recursive(pristine_origin, origin_path, src_path, dst_path, sub
 
     # still nothing gets pushed, because origin is up to date
     res_ = publish(dataset=source, recursive=True, since='HEAD^')
-    assert_result_count(
-        res_, 3, status='notneeded', type='dataset')
+    assert_status('notneeded', res_)
 
     # and we should not fail if we run it from within the dataset
     with chpwd(source.path):
         res_ = publish(recursive=True, since='HEAD^')
-        assert_result_count(
-            res_, 3, status='notneeded', type='dataset')
+        assert_status('notneeded', res_)
 
     # Let's now update one subm
     with open(opj(sub2.path, "file.txt"), 'w') as f:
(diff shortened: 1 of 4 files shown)


datalad/distribution/tests/test_publish.py
```diff
@@ -285,6 +285,10 @@ def test_publish_recursive(pristine_origin, origin_path, src_path, dst_path, sub
     # we will be publishing back to origin, so to not alter testrepo
     # we will first clone it
     origin = install(origin_path, source=pristine_origin, recursive=True)
+    # uncouple subdataset from testrepo sources after recursive install
+    # to make this clone the source of all `get` attempts
+    for sub in origin.subdatasets(result_xfm=lambda x: x['gitmodule_name']):
+        origin.subdatasets(path=sub, set_property=[('url', './{}'.format(sub))])
```
yarikoptic, line 291: was the intention to provide a "new" test scenario or due to code changes it became required to alter `url` entries?

datalad/distribution/tests/test_update.py
```diff
@@ -79,6 +79,10 @@ def test_update_simple(origin, src_path, dst_path):
     # forget we cloned it by removing remote, which should lead to
     # setting tracking branch to target:
     source.repo.remove_remote(DEFAULT_REMOTE)
+    # also forget the declared absolute location of the submodules, and turn them
+    # relative to this/a clone
+    for sub in source.subdatasets(result_xfm=lambda x: x['gitmodule_name']):
+        source.subdatasets(path=sub, set_property=[('url', './{}'.format(sub))])
```
yarikoptic, line 85: same here -- a new scenario or needed (i.e. something what worked before, no longer works)?

yarikoptic (commented):
I like the minimalism of the change but have not yet fully grasped (and thus left comments) on either the priorities tune up requires tune up of `.gitmodules` in the test(s)?

mih:
FTR: Now that https://github.com/datalad/datalad/pull/5749 is merged, this PR is needed in master in order to not break scenarios like the remodnav executable paper presented in the handbook. Current master still prefers a "local" checkout of a subdataset to clone from. Now that URL is mapped to a Github project URL (with #5749), and datalad asks for credentials to a project that does not exist (in the case of that paper). The relevant submodule record is:

```
[submodule "remodnav/tests/data/anderson_etal"]
        path = remodnav/tests/data/anderson_etal
        url = https://github.com/richardandersson/EyeMovementDetectorEvaluation.git
```

and this URL is the one that must be cloned.

With this PR on top of master, the desirable behavior is maintained/restored.

I added the 0.15 milestone. We should not release with such a change in behavior.

datalad/distribution/tests/test_publish.py
mih: No, this is a direct consequence of https://github.com/datalad/datalad/issues/5033. The testrepos configure an explicit URL in `.gitmodules`. But here we want to test what happens if there is a relative URL (it shall clone from the initial clone target done in this test). With the change introduced by the PR we would clone from the true origin. So in order to maintain the premise of the test, we need to reconfigure to a relative URL

datalad/distribution/tests/test_update.py
mih: and same answer as above. The very nature of this PR makes this change necessary.

yarikoptic:
thanks @mih. let's proceed and see where sea takes us ;-)

PR #5766, opened 2021-06-26
[JOSS review] Change bib classification of BIDS paper

arokem:
The current bibliography item that gets rendered into the article is a bit wonky (including the publisher name, for example). I think that reclassifying this explicitly as an article may make the rendered bibliography item more conformant.

### DIFF
--- a/paper.bib
+++ b/paper.bib
@@ -174,7 +174,7 @@
     year={2021}
 }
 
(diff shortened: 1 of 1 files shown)


yarikoptic:
yeap, cleans it up a bit:
image

Thank you @arokem!

Issue #5775, opened 2021-06-30
Provide means to force re-entry of credentials


mih:
I am not aware of a procedure to update credentials, other than

- find out where they are stored, using largely undocumented means, or platform expertise
- find out how to delete or update them there

The handbook does another thing for the HCP S3 credential special case, suggesting people to code something themselves with python `keyring`.

This situation is rather sad. DataLad should simply be enabled to treat credentials as unknown, if instructed.

I will propose a PR.

Issue #396, opened 2016-04-06
provide user accessible (API) interface to credentials


yarikoptic:
Should allow to list, add, remove, modify credentials "managed" by datalad one way or another.

mih:
Seems as if this could be a plugin.

mih:
As discussed in #4981, the key issue that prevents work on this for 5 years now is the inability of the credential system to list all known credentials, and by extension making it hard for users to even know what credential should be edited in order to impact a particular service/functionality.

https://github.com/datalad/datalad/issues/4981#issuecomment-705706777 argues that this is not necessary, because `Provider`s can report the IDs of the credentials they require.

However, this leaves out any credential relevant beyond plain download authentication (pretty much any service auth: github/lab and friends, OSF, webdav).

mih:
https://github.com/datalad/datalad-next now provides a `credentials` command, which does exactly that.

PR #5777, opened 2021-06-30
New config `datalad.credentials.force-ask` to force interactive credential updates

mih:
This change introduces a new configuration setting:

```
 % datalad x-configuration
 ...
 # Force (re-)entry of credentials
 # Should DataLad prompt for credential (re-)entry? This can be used to
 # update previously stored credentials.
 datalad.credentials.force-ask=False
 ...
```

It enables seamless updates of stored credentials. Importantly, users
are no longer required to learn where credentials are stored, and how to
update them in that store. It makes instructions provided in the
handbook on how to learn python `keyring` superfluous. Credential
updates are simply entered at the same place where the original
credentials were provided.

This is not a perfect solution (can get messy when there are a
number of credentials involved in a single datalad operation, and only a
single one needs updating), but it addresses a substantial usability
issue for a broad range of use cases in a straightforward manner.

Besides being compatible with the usual ways of DataLad configuration
handling, this also works for plain git-annex calls that trigger
DataLad's credential handling, e.g.:

```
git -c datalad.credentials.force-ask=no annex get T1w_acpc_dc.nii.gz
```

Fixes datalad/datalad#5775

This is related to #396 by addressing the "update" use case.

### DIFF
--- a/datalad/interface/common_cfg.py
+++ b/datalad/interface/common_cfg.py
@@ -80,6 +80,15 @@ definitions = {
         'destination': 'local',
         'type': bool,
     },
+    # this is actually used in downloaders, but kept cfg name original
+    'datalad.credentials.force-ask': {
+        'ui': ('yesno', {
+               'title': 'Force (re-)entry of credentials',
+               'text': 'Should DataLad prompt for credential (re-)entry? This '
+                       'can be used to update previously stored credentials.'}),
+        'type': bool,
+        'default': False,
+    },
     'datalad.externals.nda.dbserver': {
         'ui': ('question', {
                'title': 'NDA database server',
--- a/datalad/downloaders/credentials.py
+++ b/datalad/downloaders/credentials.py
@@ -32,6 +32,7 @@ from ..utils import auto_repr
 from ..support.network import iso8601_to_epoch
 
 from datalad import cfg as dlcfg
+from datalad.config import anything2bool
 
 from logging import getLogger
 lgr = getLogger('datalad.downloaders.credentials')
@@ -157,8 +158,14 @@ class Credential(object):
     def __call__(self):
         """Obtain credentials from a keyring and if any is not known -- ask"""
         fields = {}
+        # check if we shall ask for credentials, even if some are on record
+        # already (but maybe they were found to need updating)
+        force_reentry = dlcfg.obtain(
+            'datalad.credentials.force-ask',
+            valtype=anything2bool)
         for f in self._FIELDS:
-            v = self._get_field_value(f)
+            # don't query for value if we need to get a new one
+            v = None if force_reentry else self._get_field_value(f)
             if not self._is_field_optional(f):
                 while v is None:  # was not known
                     v = self._ask_and_set(f)
--- a/datalad/downloaders/tests/test_credentials.py
+++ b/datalad/downloaders/tests/test_credentials.py
@@ -27,7 +27,7 @@ from ..credentials import (
     CompositeCredential,
     UserPassword,
 )
-from datalad.config import ConfigManager
+from datalad import cfg as dlcfg
 
 
 @with_testsui(responses=[
@@ -59,7 +59,7 @@ def test_cred1_enter_new():
     assert_equal(keyring.get('name', 'user'), 'user2')
     assert_equal(keyring.get('name', 'password'), 'newpassword')
 
-@with_testsui(responses=['password1'])
+@with_testsui(responses=['password1', 'newuser', 'newpassword'])
 def test_cred1_call():
     keyring = MemoryKeyring()
     cred = UserPassword("name", keyring=keyring)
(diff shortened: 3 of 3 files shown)


datalad/interface/common_cfg.py
```diff
@@ -80,6 +80,15 @@
         'destination': 'local',
         'type': bool,
     },
+    # this is actually used in downloaders, but kept cfg name original
+    'datalad.credential.force-ask': {
```
bpoldrack, line 84: Minor, but I feel it's somewhat strange/inconsistent to use `credential` instead of `credentials` in that name. No strong objection, though.
However, if I think ahead imagining that we may want to enhance this, addressing a particular set of credentials by a configuration, singular at that level feels even more strange ;-)

bpoldrack (dismissed):
Looks straight forward to me.

datalad/interface/common_cfg.py
mih: If someone can provide the rules for switching between singular and plural, please write them down. Our use of that is a wild ride. I hardly ever (consciously) see use of plural outside datalad. I am not aware of a single instance of a Git config setting that uses plural (NB. Git has `credential. ...`).

I have been guilty of introducing the use of plural in the past myself. I regret all of them.

yarikoptic:
can't lookup atm, but I think we contemplated a command like `datalad credentials` which would list known providers/credentials and interface forced re-entry.   But I guess it is orthogonal/complimentary to this PR, which in general I do not mind, but haven't thought yet about possible corner cases etc.

mih:
OK, I will put this back in draft mode. It does not work: parallel operation!

```
git -c datalad.credential.force-ask=yes annex get -J2 T1w_acpc_dc.nii.gz ribbon.nii.gz
```

asks twice.

mih:
So thinking about this for a bit more. It might just be a matter of documenting that parallel operations and this flag do not work conveniently with each other. Any other solution would yield in a undesirably complex entangling of credential queries and download (or other) operations.

christian-monch:
> [...]
> Bottom line: I think this is a useful feature, even with a coming `credential` command.

Without going into technical details, I agree with the rationale given here. Especially since the `credential` command might not be here before a few months have passed

bpoldrack:
> Bottom line: I think this is a useful feature, even with a coming credential command.

I tend to agree, but it may need reconsideration of config naming. We have command specific configs like `datalad.clone.something`. If we are introducing `datalad.credential.some` and then later a command `credential`, then the config better be something that makes sense for that command to respect or not for the config to not "conflict" in the first place. 
Not quite sure how to avoid future user confusion.

The feature itself seems to be worthwhile with or w/o the command, though.

mih:
> I tend to agree, but it may need reconsideration of config naming. We have command specific configs like `datalad.clone.something`. If we are introducing `datalad.credential.some` and then later a command `credential`, then the config better be something that makes sense for that command to respect or not for the config to not "conflict" in the first place.
> Not quite sure how to avoid future user confusion.

Do you have a suggestion for a change?

datalad/interface/common_cfg.py
yarikoptic: TL;DR:  I think going for plural at this point in time would be more consistent with current state of affairs.

<details>
<summary>TL:</summary> 

- we have `datalad.externals`, `datalad.tests` and `datalad.locations` already... the singular ones kinda make sense singular 
<details>
<summary>full grep</summary> 

```shell
      1     'datalad.annex
      1     'datalad.crawl
      1     'datalad.exc
      1     'datalad.externals
      1     'datalad.github
      1     'datalad.install
      8     'datalad.locations
      9     'datalad.log
      4     'datalad.metadata
      3     'datalad.repo
      5     'datalad.runtime
      1     'datalad.save
      3     'datalad.search
      2     'datalad.ssh
     16     'datalad.tests
      4     'datalad.ui
      1                     "See 'datalad.ui
      1                'title': 'Include exceptions and their traceback in log messages. If set, \'datalad.exc

```
</details>

so it feels like `datalad.credentials` would be more consistent with that

- also `datalad.credential` somewhat suggests to have some ultimate singular credential, thus confusing
- and our commands ATM are plural, e.g. `datalad subdatasets`, `siblings`; so if we are to come up with a helper command (like e.g. sketched [here](https://github.com/datalad/datalad/issues/4981#issuecomment-705706777)), it should be `credentials` to stay consistent.

"When" for singular
- if we had it like `datalad.credential.<credential-id>.force-ask` then it might have become indeed more consistent with `git` having them singular. 
- If we ever to RF to establish "hierarchical commands" mimicing `git` more closely in API design, like suggested [here](https://github.com/datalad/datalad/issues/4539#issuecomment-756204514), that is where singular `datalad credential` (and thus config section) would be more appropriate.

</details>

yarikoptic:
> > I tend to agree, but it may need reconsideration of config naming. We have command specific configs like `datalad.clone.something`. If we are introducing `datalad.credential.some` and then later a command `credential`, then the config better be something that makes sense for that command to respect or not for the config to not "conflict" in the first place.
> > Not quite sure how to avoid future user confusion.
> 
> Do you have a suggestion for a change?

make it `datalad.credentials` config and `datalad credentials` command. At least until some brave soul would do big UI RF towards something like https://github.com/datalad/datalad/issues/4539#issuecomment-756204514 . More argumentation is [above](https://github.com/datalad/datalad/pull/5777#discussion_r670693042)

mih:
Bowed to the majority of expressed opinions, and force-pushed an amendment that renames the config setting to plural.

mih:
Took out of draft mode, ready to merge. Test failure is unrelated #5794

yarikoptic (approved):
I think it might be worth to tune up PR title to become more specific, before merge

mih:
Title is updated.

PR #5778, opened 2021-06-30
fix a bug in descriptor name rendering

### LINKED ISSUE #5774: New Runner swamps logging with growing? messages
I have a hard time understanding what is happening, but I believe the Runner keeps logging output that it has read from a process with every update BUT always starts at the beginning. In other words, each log message contains all previously logged output of a process again

I ran `datalad -l 5 get HCP1200/143426/T1w/T1w_acpc_dc.nii.gz &> log` to get a single file from the HCP dataset. The captured `log` is 22 MB (!) with less than 600 log lines.

This makes the log incomprehensible.

christian-monch:
This fixes #5774 

Thanks to @mih for catching the bug before the 0.15 release.

Due to a missing array dereferencing the complete hitherto buffered output data of a command execution was written into each log message that logged received data.

### DIFF
--- a/datalad/cmd_protocols.py
+++ b/datalad/cmd_protocols.py
@@ -83,7 +83,7 @@ class WitlessProtocol(asyncio.SubprocessProtocol):
         pass
 
     def _log_summary(self, fd, data):
-        fd_name = self.fd_infos[fd]
+        fd_name = self.fd_infos[fd][0]
         lgr.log(5, 'Read %i bytes from %i[%s]%s',
                 len(data), self.process.pid, fd_name, ':' if self._log_outputs else '')
         if self._log_outputs:


mih:
I can confirm that this dramatically reduced the output. However, instead of seeing the git-annex chatter multiple times, I no longer see it at all? This seems to match the behavior of `maint`, though. I am a bit confused.

christian-monch:
> I can confirm that this dramatically reduced the output. However, instead of seeing the git-annex chatter multiple times, I no longer see it at all? This seems to match the behavior of `maint`, though. I am a bit confused.

The original goal was to match behavior of the async-runner. Subprocess output is now logged iff datalad.log.outputs is True.

christian-monch:
> [...] Subprocess output is logged iff datalad.log.outputs is True [...]

In addition the logger has to be enabled for level 5

mih (approved):
Thx, confirmed!

datalad/cmd.py
```diff
@@ -254,7 +254,6 @@ def get_git_environ_adjusted(env=None):
         if 'GIT_SSH_COMMAND' not in git_env:
             git_env['GIT_SSH_COMMAND'] = GIT_SSH_COMMAND
             git_env['GIT_SSH_VARIANT'] = 'ssh'
-        git_env['GIT_ANNEX_USE_GIT_SSH'] = '1'
```
yarikoptic, line 257: Why is it "rogue" and removed?
Seems was added for a reason: https://github.com/datalad/datalad/commit/6320c3f826a1335c301b57889a472685f21b82d7

yarikoptic: Definitely OOS for the purpose/title of the PR

christian-monch: Yepp. I put it back. 
Thx

PR #5781, opened 2021-07-03
shell-completion: add argcomplete to 'misc' extra_depends, log an ERROR if argcomplete fails to import

yarikoptic:
$> datalad shell-completion > /dev/null
    [ERROR] argcomplete failed to import - completions unlikely to work. Check if it is installed: No module named 'argcomplete' [shell_completion.py:__call__:51]

Motivator: ruedin as of https://neurostars.org/t/tab-completion-not-working-with-datalad-commands/19683/3?u=yarikoptic


AKSoo:
I was also wondering why the script from `datalad shell-completion` was not working.

I think `argcomplete` should be included as an optional dependency in setup and/or explicit mentioned as a separate dependency in shell-completion docs.

mih:
I agree. I don't see why it should not be a declared dependency, given the current list of deps.

yarikoptic:
indeed! I have not realized it was not listed.  Added to `misc` (so part of `pip install datalad[full]`) in 2c74fd973b2e00911a8e29e4611b7480622a3484

yarikoptic:
I think fails are unrelated (due to newer annex on osx) and I will proceed with the merge

Issue #5785, opened 2021-07-07
2 new failures in `maint`  on datalad/git-annex


yarikoptic:
A sattelite to #5759 : two other failures creeped in  while those prior 2 regressions (?) were not attended to

```
ERROR: datalad.support.tests.test_external_versions.test_external_versions_basic
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.7.10/x64/lib/python3.7/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/opt/hostedtoolcache/Python/3.7.10/x64/lib/python3.7/site-packages/datalad/support/tests/test_external_versions.py", line 57, in test_external_versions_basic
    assert_greater(ev[our_module], '0.1')
  File "/opt/hostedtoolcache/Python/3.7.10/x64/lib/python3.7/unittest/case.py", line 1249, in assertGreater
    if not a > b:
  File "/opt/hostedtoolcache/Python/3.7.10/x64/lib/python3.7/distutils/version.py", line 64, in __gt__
    c = self._cmp(other)
  File "/opt/hostedtoolcache/Python/3.7.10/x64/lib/python3.7/distutils/version.py", line 337, in _cmp
    if self.version < other.version:
TypeError: '<' not supported between instances of 'str' and 'int'
```

super fresh
```
$> git grep 'test_external_versions_basic.*ERR'
cron-20210707/build-macos.yaml-344-ac85ac50-failed/2_test-datalad (maint).txt:2021-07-07T04:07:10.3761420Z datalad.support.tests.test_external_versions.test_external_versions_basic ... ERROR
cron-20210707/build-macos.yaml-344-ac85ac50-failed/test-datalad (maint)/11_Run datalad tests.txt:2021-07-07T04:07:10.3761400Z datalad.support.tests.test_external_versions.test_external_versions_basic ... ERROR
cron-20210707/build-ubuntu.yaml-336-ac85ac50-failed/2_test-datalad (maint).txt:2021-07-07T04:02:34.5914734Z datalad.support.tests.test_external_versions.test_external_versions_basic ... ERROR
cron-20210707/build-ubuntu.yaml-336-ac85ac50-failed/test-datalad (maint)/12_Run datalad tests.txt:2021-07-07T04:02:34.5914720Z datalad.support.tests.test_external_versions.test_external_versions_basic ... ERROR
[...]

PR #5786, opened 2021-07-07
Tests: Allow for version to contain + as a separator and provide more information for version related comparisons

yarikoptic:
To partially address and provide more information to figure out details of https://github.com/datalad/datalad/issues/5785


yarikoptic:
if no surprises, will merge to make datalad/git-annex fails more informative

Issue #5763, opened 2021-06-25
create-sibling-gitlab does not talk and does not do (at times)


mih:
Running the following in a dataset comes back quickly, with no output and nothing done.

```
% datalad create-sibling-gitlab . 
```

Not just `.`, anything

PR #5789, opened 2021-07-14
BF: create-sibling-gitlab: do not skip root dataset when path is given

mih:
Before `datalad create-sibling-gitlab .` would just do nothing.
Now the root dataset is processed if a given path matches it.

Fixes datalad/datalad#5763

### DIFF
--- a/datalad/distributed/tests/test_create_sibling_gitlab.py
+++ b/datalad/distributed/tests/test_create_sibling_gitlab.py
@@ -77,15 +77,17 @@ def test_dryrun(path):
         site='dummy', sibling='dummy',
     )
     # now a working, fully manual call
-    res = ctlg['root'].create_sibling_gitlab(
-        dryrun=True, on_failure='ignore',
-        site='dummy', project='here',
-    )
-    assert_result_count(res, 1)
-    assert_result_count(
-        res, 1, path=ctlg['root'].path, type='dataset', status='ok',
-        site='dummy', sibling='dummy', project='here',
-    )
+    for p in (None, ctlg['root'].path):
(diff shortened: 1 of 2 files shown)


mih:
A whole range of unrelated, heterogeneous failures on MacOS.

datalad/distributed/tests/test_create_sibling_gitlab.py
```diff
@@ -77,15 +77,16 @@ def test_dryrun(path):
...
-    res = ctlg['root'].create_sibling_gitlab(
-        dryrun=True, on_failure='ignore',
-        site='dummy', project='here',
-    )
-    assert_result_count(res, 1)
-    assert_result_count(
-        res, 1, path=ctlg['root'].path, type='dataset', status='ok',
-        site='dummy', sibling='dummy', project='here',
-    )
+    for p in (None, ctlg['root'].path):
```
yarikoptic, line 80: `p` seems to be not used below -- shouldn't it be?

mih: LOL, indeed! Thx for catching that!

mih: Test fixed. Pointed to a flaw in the implementation. Adjusted and added a comment pointing out the (avoided) pitfall.

mih: Force pushed an amendment.

mih:
Test failure is unrelated #5794

mih:
Happy all around!

Issue #5028, opened 2020-10-13
Support `git-credentials` as a credential provider 


mih:
It provides a nice and extensible, that is uniform across platforms.

`man 7 gitcredentials`

Particularly attractive are the following aspects:

- a single `git` command to support (`git credential`), all others can simply comply with the API defined by `git -- this includes datalad's own credential support
- separation of `fill` and `approve` functionality that allows for testing credentials before they are stored

However, it is sufficiently different from our credential setup to make things difficult to integrate.

It can only do user/pass, which may suggest that we should only amend the `UserPassword` credential to query the git credentials.

However, our use of the python keyring pretty much abuses it to store user/password credentials with actually two credentials that are similarly named. So it could also be seen a more capable, by providing credential pairs.

So there are pretty much two ways.

1. Implement support for git credentials within `Credential` in DataLad (either as a fallback or a priority)
2. Or redoing `Credential` sitting on top of `git credential` and providing the python keyring support as a git credential helper

I am presently leaning towards (1).

A bunch of questions need answers to determine the most sensible implementation

- Can `git credential` be used to probe for the presence of a credential? If not, it cannot provide the `known()` feature of `Credential()` and we can only use it as a fallback.

  It seems this is possible! Prompting can be disabled like so:
  
  ```
  % echo url=http://example.com | GIT_ASKPASS=true git credential fill  
  protocol=http
  host=example.com
  username=
  password=
  ```
  This could not distinguish between an unset and a credential set to empty.

- ATM we implement are own credential entry via `ui.question()`. It only works in the terminal, and has usability issues (Ctrl-C behavior is flaky). OTOH `git credential fill` supports launching credential entry helpers via all the standard setups, incl `SSH_ASKPASS`. Is this a feature worth adopting?

mih:
Fixed by #5796

PR #5796, opened 2021-07-17
git-credential interfaces (datalad's in Git and vice versa)

mih:
Towards #5028

This description is updated by me (@bpoldrack), since I took over.

This implements an integration with `git-credential` in both directions:
1. A `git-credential-datalad` script that allows git to query (and store to)
   Datalad's credential system, by adding it as potential backend by setting 
   `credential.helper=datalad`. This credential helper will query datalad for a 
   provider matching the URL in question and retrieve the linked credentials.
   The other way around, storing will create such a provider config file and the 
   respective credential entries via `keyring`.
2. `Gitcredential` class, that can be referred to by provider configs as the 
   `type` of credential ('git'). This allows datalad to use `git-credential` 
   instead of `keyring`. Credentials provided via datalad config (or environment 
   variables) are still taking precedence over what is stored in that backend 
   (same as with keyring: no change).
   This class uses the same internal `FIELDS` as the `UserPassword` class and 
   hence should be useable with any authenticator that works with `UserPassword`.

Tests for the credential helper in `datalad/local/tests/test_gitcredential.py`
and for the new `Credential` object in
`datalad/downloaders/tests/test_credentials.py:test_gitcredential_read` and
`datalad/downloaders/tests/test_credentials.py:test_gitcredential`

While 1.) is a rather seemless integration in my view, 2.) isn't ideal yet, as 
it's sort of a special case, where the provider config would need to state it
wants to query `git-credential` rather than `keyring` by setting `type=git` for
the credential (rather than `type=userpassword`). However, I think this is good
enough to enable these usecases but should be developed further and make
`git-credential` just an additional backend to query (in addition rather than
instead of keyring). This requires some more overhauling of our credential
system as a whole, though. This view may raise concerns WRT introducing a value
for such - possibly committed - configs, that may be deprecated in the
foreseeable future again. In my view, this would only concern those 
configuration files and I think they should be replaced in the process of
overhauling the credential system. They are an unnecessary exception to how
configurations are dealt with by datalad everywhere else. Those files should
become a legacy format, that we keep support for reading. But the to-be default
approach should be regular, datalad configuration variables (probably in
standard config file locations). The switch will then not be "Hey use credential 
type git" to "Just kidding, use credential type userpassword", but instead from
"This is how you do it with provider config files" to "You can still do it that
way, but better do it like everywhere else. The new datalad config variables for
this are: ...". Hence, that anticipated deprecation doesn't really matter, I think.

- DataLad's credential system is now able to query `git-credential` by specifying credential type `git` in the respective provider configuration [#5796](https://github.com/datalad/datalad/pull/5796) (by @bpoldrack)
- DataLad now comes with a git credential helper `git-credential-datalad` allowing Git to query DataLad's credential system [#5796](https://github.com/datalad/datalad/pull/5796) (by @bpoldrack and @mih)

#### 📝 Documentation

- A new [document](http://docs.datalad.org/credentials.html) describes the credential system from a user's perspective [#5796](https://github.com/datalad/datalad/pull/5796) (by @bpoldrack)
- Enhance the [design document](http://docs.datalad.org/design/credentials.html) on DataLad's credential system [#5796](https://github.com/datalad/datalad/pull/5796) (by @bpoldrack)

#### 🏠 Internal

- Add internal method for non-interactive provider/credential storing [#5796](https://github.com/datalad/datalad/pull/5796) (by @bpoldrack)
- Allow credential classes to have a context set, consisting of a URL they are to be used with and a dataset DataLad is operating on, allowing to consider "local" and "dataset" config locations [#5796](https://github.com/datalad/datalad/pull/5796) (by @bpoldrack)

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -53,6 +53,9 @@ shallow_clone: false
 
 environment:
   DATALAD_TESTS_SSH: 1
+  # Prevent interactive credential entry (note "true" is the command to run)
+  # See also the core.askPass setting below
+  GIT_ASKPASS: true
 
   # Do not use `image` as a matrix dimension, to have fine-grained control over
   # what tests run on which platform
@@ -220,6 +223,8 @@ init:
   # Identity setup
   - git config --global user.email "test@appveyor.land"
   - git config --global user.name "Appveyor Almighty"
+  # Companion to the GIT_ASKPASS environment setting above
+  - git config --global core.askPass ""
   # globally setting filter.annex.process (needs git-annex 8.20211117) to reduce git-add runtime
   # https://git-annex.branchable.com/bugs/Windows__58___substantial_per-file_cost_for___96__add__96__/
   - cmd: git config --system filter.annex.process "git-annex filter-process"
--- a/datalad/downloaders/credentials.py
+++ b/datalad/downloaders/credentials.py
@@ -35,6 +35,8 @@ from ..support.network import iso8601_to_epoch
 
 from datalad import cfg as dlcfg
 from datalad.config import anything2bool
+from datalad.distribution.dataset import Dataset
+from datalad.local.gitcredential import GitCredentialInterface
 
 from logging import getLogger
 lgr = getLogger('datalad.downloaders.credentials')
@@ -43,6 +45,11 @@ lgr = getLogger('datalad.downloaders.credentials')
 @auto_repr
 class Credential(object):
     """Base class for different types of credentials
+
+    Note: While subclasses can define their own `_FIELDS`, they are actually
+    assumed to have particular keys by the implementation of (some)
+    authenticators. `HTTPRequestsAuthenticator` and its subclasses for example,
+    assume `user` and `password` to be valid keys.
     """
 
     # Should be defined in a subclass as an OrderedDict of fields
@@ -54,7 +61,8 @@ class Credential(object):
         'optional',  # Not mandatory thus not requested if not set
     }
 
-    def __init__(self, name, url=None, keyring=None):
+    def __init__(self, name, url=None, keyring=None, auth_url=None,
+                 dataset=None):
         """
         Parameters
         ----------
@@ -66,9 +74,16 @@ class Credential(object):
         keyring : a keyring
             An object providing (g|s)et_password.  If None, keyring module is used
             as is
+        auth_url : str, optional
+            URL string this credential is going to be used with. This context
+            may be needed to query some credential systems (like git-credential).
+        dataset : str, Path or Dataset
+            The dataset datalad is operating on with this credential. This may
+            be needed for context in order to query local configs.
         """
         self.name = name
         self.url = url
+        self.set_context(auth_url=auth_url, dataset=dataset)
         self._keyring = keyring or keyring_
         self._prepare()
 
@@ -206,6 +221,50 @@ class Credential(object):
         """Deletes credential values from the keyring"""
         for f in self._FIELDS:
             self._keyring.delete(self.name, f)
+
+    def set_context(self, auth_url=None, dataset=None):
+        """Set URL/dataset context after instantiation
+
+        ATM by design the system of providers+downloaders+credentials doesn't
+        necessarily provide access to that information at instantiation time of
+        `Credential` objects. Hence, allow to provide this whenever we can.
+
+        Arguments are only applied if provided. Hence, `None` does not overwrite
+        a possibly already existing attribute.
+
+        Note
+        ----
+        Eventually, this is going to need a major overhaul. `Providers` etc. are
+        built to be mostly unaware of their context, which is why
+        `get_dataset_root()` tends to be the only way of assessing what dataset
+        we are operating on. This will, however, fail to detect the correct
+        dataset, if it can'T be deduced from PWD, though.
+
+        Parameters
+        ----------
+        auth_url : str, optional
+            URL string this credential is going to be used with. This context
+            may be needed to query some credential systems (like git-credential).
+        dataset : str, Path or Dataset, optional
+            The dataset datalad is operating on with this credential. This may
+            be needed for context in order to query local configs.
+        """
+
+        # TODO: change of context should probably not be allowed. When context
+        #       is actually needed for a particular credential store, this
+        #       object represents such associated creds.
+        #       Allowing to switch context within the same instance leads to
+        #       trouble determining when exactly a reload is needed and what is
+        #       to be overwritten or not.
+
+
+
+        if auth_url:
+            self.auth_url = auth_url
+        if isinstance(dataset, Dataset):
+            self.ds = dataset
+        else:
+            self.ds = Dataset(dataset) if dataset else None
 
 
 class UserPassword(Credential):
@@ -388,10 +447,71 @@ class LORIS_Token(CompositeCredential):
         super(CompositeCredential, self).__init__(name, url, keyring)
 
 
-CREDENTIAL_TYPES = {
-    'user_password': UserPassword,
-    'aws-s3': AWS_S3,
-    'nda-s3': NDA_S3,
-    'token': Token,
-    'loris-token': LORIS_Token
-}
+class GitCredential(Credential):
+    """Credential to access git-credential
+    """
+
+
+    _FIELDS = OrderedDict([('user', {}),
+                           ('password', {'hidden': True})])
+
+    # substitute keys used within datalad by the ones used by git-credential
+    _FIELDS_GIT = {'user': 'username',
+                   'password': 'password'}
+
+    is_expired = False  # no expiration provisioned
+
+    def __init__(self, name, url=None, keyring=None,
+                 auth_url=None, dataset=None):
+        super().__init__(name, url=url, keyring=keyring,
+                         auth_url=auth_url, dataset=dataset)
+
+    def _get_field_value(self, field):
+
+        from datalad import cfg as dlcfg
+        cfg = self.ds.config if self.ds else dlcfg
+        cfg.reload()
+        from_cfg = cfg.get('datalad.credential.{name}.{field}'.format(
+            name=self.name,
+            field=field.replace('_', '-')
+        ))
+
+        if from_cfg:
+            # config takes precedence
+            return from_cfg
+
+        # Note:
+        # In opposition to the keyring approach of other `Credential`s,
+        # we don't query for single values, but for an entire "description".
+        # Currently required methods have to return single values, though.
+        # Hence, calls to `git credential fill` could be optimised. Not easy to
+        # assess when exactly this class can know whether it's context is yet to
+        # be completed, though, so that another `fill` would actually yield
+        # something different than before.
+
+        self._git_cred.fill()
+        git_field = self._FIELDS_GIT[field]  # translate to git-credential terms
+        if git_field in self._git_cred and self._git_cred[git_field]:
+            return self._git_cred[git_field]
+
+        # we got nothing
+        return
+
+    def set(self, **kwargs):
+        for f, v in kwargs.items():
+            if f not in self._FIELDS:
+                raise ValueError("Unknown field %s. Known are: %s"
+                                 % (f, self._FIELDS.keys()))
+
+        mapped = {self._FIELDS_GIT[k]: v for k, v in kwargs.items()}
+        self._git_cred = GitCredentialInterface(url=self.auth_url, repo=self.ds,
+                                                **mapped)
+        self._git_cred.approve()
+
+    def delete(self):
+        """Deletes credential"""
+        self._git_cred.reject()
+
+    def set_context(self, auth_url=None, dataset=None):
+        super().set_context(auth_url, dataset)
+        self._git_cred = GitCredentialInterface(url=auth_url, repo=dataset)
--- a/datalad/downloaders/providers.py
+++ b/datalad/downloaders/providers.py
@@ -12,7 +12,6 @@
 from glob import glob
 from logging import getLogger
 
-import os
 import re
 from os.path import dirname, abspath, join as pathjoin
 from urllib.parse import urlparse
@@ -32,12 +31,15 @@ from .http import (
 from .s3 import S3Authenticator, S3Downloader
 from .shub import SHubDownloader
 from configparser import ConfigParser as SafeConfigParserWithIncludes
-from ..support.external_versions import external_versions
-from ..support.network import RI
-from ..support import path
-from ..utils import auto_repr
-from ..utils import ensure_list_from_str
-from ..utils import get_dataset_root
+from datalad.support.external_versions import external_versions
+from datalad.support.network import RI
+from datalad.support import path
+from datalad.utils import (
+    auto_repr,
+    ensure_list_from_str,
+    get_dataset_root,
+    Path,
+)
 
 from ..interface.common_cfg import dirs
 
@@ -59,7 +61,7 @@ AUTHENTICATION_TYPES = {
     'none': NoneAuthenticator,
 }
 
-from .credentials import CREDENTIAL_TYPES
+from datalad.downloaders import CREDENTIAL_TYPES
 
 
 @auto_repr
@@ -156,6 +158,23 @@ class Providers(object):
 
     _DEFAULT_PROVIDERS = None
     _DS_ROOT = None
+    _CONFIG_TEMPLATE = """\
+# Provider configuration file created to initially access
+# {url}
+
+[provider:{name}]
+url_re = {url_re}
+authentication_type = {authentication_type}
+# Note that you might need to specify additional fields specific to the
+# authenticator.  Fow now "look into the docs/source" of {authenticator_class}
+# {authentication_type}_
+credential = {credential_name}
+
+[credential:{credential_name}]
+# If known, specify URL or email to how/where to request credentials
+# url = ???
+type = {credential_type}
+"""
 
     def __init__(self, providers=None):
         """
@@ -261,6 +280,8 @@ class Providers(object):
                     raise ValueError("Unknown credential %s. Known are: %s"
                                      % (provider.credential, ", ".join(credentials.keys())))
                 provider.credential = credentials[provider.credential]
+                # TODO: Is this the right place to pass dataset to credential?
+                provider.credential.set_context(dataset=cls._DS_ROOT)
 
         providers = Providers(list(providers.values()))
 
@@ -363,10 +384,74 @@ class Providers(object):
                   scheme, provider)
         return provider
 
-    def enter_new(self, url=None, auth_types=[]):
+    def _store_new(self, url=None, authentication_type=None,
+                   authenticator_class=None, url_re=None, name=None,
+                   credential_name=None, credential_type=None, level='user'):
+        """Stores a provider and credential config and reloads afterwards.
+
+        Note
+        ----
+        non-interactive version of `enter_new`.
+        For now non-public, pending further refactoring
+
+        Parameters
+        ----------
+        level: str
+          Where to store the config. Choices: 'user' (default), 'ds', 'site'
+
+        Returns
+        -------
+        Provider
+          The stored `Provider` as reported by reload
+        """
+
+        # We don't ask user for confirmation, so for this non-interactive
+        # routine require everything to be explicitly specified.
+        if any(not a for a in [url, authentication_type, authenticator_class,
+                               url_re, name, credential_name, credential_type]):
+            raise ValueError("All arguments must be specified")
+
+        if level not in ['user', 'ds', 'site']:
+            raise ValueError("'level' must be one of 'user', 'ds', 'site'")
+
+        providers_dir = Path(self._get_providers_dirs()[level])
+        if not providers_dir.exists():
+            providers_dir.mkdir(parents=True, exist_ok=True)
+        filepath = providers_dir / f"{name}.cfg"
+        cfg = self._CONFIG_TEMPLATE.format(**locals())
+        filepath.write_bytes(cfg.encode('utf-8'))
+        self.reload()
+        return self.get_provider(url)
+
+    def enter_new(self, url=None, auth_types=[], url_re=None, name=None,
+                  credential_name=None, credential_type=None):
+        # TODO: level/location!
+        """Create new provider and credential config
+
+        If interactive, this will ask the user to enter the details (or confirm
+        default choices). A dedicated config file is written at
+        <user_config_dir>/providers/<name>.cfg
+
+        Parameters:
+        -----------
+        url: str or RI
+          URL this config is created for
+        auth_types: list
+          List of authentication types to choose from. First entry becomes
+          default. See datalad.downloaders.providers.AUTHENTICATION_TYPES
+        url_re: str
+          regular expression; Once created, this config will be used for any
+          matching URL; defaults to `url`
+        name: str
+          name for the provider; needs to be unique per user
+        credential_name: str
+          name for the credential; defaults to the provider's name
+        credential_type: str
+          credential type to use (key for datalad.downloaders.CREDENTIAL_TYPES)
+        """
+
         from datalad.ui import ui
-        name = None
-        if url:
+        if url and not name:
             ri = RI(url)
             for f in ('hostname', 'name'):
                 try:
@@ -398,7 +483,10 @@ class Providers(object):
             else:
                 break
 
-        url_re = re.escape(url) if url else None
+        if not credential_name:
+            credential_name = name
+        if not url_re:
+            url_re = re.escape(url) if url else None
         while True:
             url_re = ui.question(
                 title="New provider regular expression",
@@ -439,42 +527,30 @@ class Providers(object):
             title="Credential",
             text="What type of credential should be used?",
             choices=sorted(CREDENTIAL_TYPES),
-            default=getattr(authenticator_class, 'DEFAULT_CREDENTIAL_TYPE')
+            default=credential_type or getattr(authenticator_class,
+                                               'DEFAULT_CREDENTIAL_TYPE')
         )
 
-        # Just create a configuration file and reload the thing
-        if not path.lexists(providers_user_dir):
-            os.makedirs(providers_user_dir)
-        cfg = """\
-# Provider configuration file created to initially access
-# {url}
-
-[provider:{name}]
-url_re = {url_re}
-authentication_type = {authentication_type}
-# Note that you might need to specify additional fields specific to the
-# authenticator.  Fow now "look into the docs/source" of {authenticator_class}
-# {authentication_type}_
-credential = {name}
-
-[credential:{name}]
-# If known, specify URL or email to how/where to request credentials
-# url = ???
-type = {credential_type}
-""".format(**locals())
+        cfg = self._CONFIG_TEMPLATE.format(**locals())
         if ui.yesno(
             title="Save provider configuration file",
             text="Following configuration will be written to %s:\n%s"
                 % (filename, cfg),
             default='yes'
         ):
-            with open(filename, 'wb') as f:
-                f.write(cfg.encode('utf-8'))
+            # Just create a configuration file and reload the thing
+            return self._store_new(url=url,
+                                   authentication_type=authentication_type,
+                                   authenticator_class=authenticator_class,
+                                   url_re=url_re,
+                                   name=name,
+                                   credential_name=credential_name,
+                                   credential_type=credential_type,
+                                   level='user'
+                                   )
         else:
             return None
-        self.reload()
-        # XXX see above note about possibly multiple matches etc
-        return self.get_provider(url)
+
 
     # TODO: avoid duplication somehow ;)
     # Sugarings to get easier access to downloaders
--- a/datalad/local/gitcredential_datalad.py
+++ b/datalad/local/gitcredential_datalad.py
@@ -1,0 +1,168 @@
+import sys
+
+from datalad.downloaders import UserPassword
+from datalad.downloaders.providers import (
+    AUTHENTICATION_TYPES,
+    Providers,
+)
+from datalad.local.gitcredential import _credspec2dict
+from datalad.utils import Path
+
+git_credential_datalad_help = """\
+Git credential interface to DataLad's credential management system.
+
+Only DataLad's UserPassword-type credentials are supported. This helper
+passes standard 'get', 'store' actions on to the respective interfaces in
+DataLad.
+The 'erase' action is not supported, since this is called by Git, when the
+credentials didn't work for a URL. However, DataLad's providers store a regex
+for matching URLs. That regex-credential combo may still be valid and simply
+too broad. We wouldn't want to auto-delete in that case. Another case is a
+somewhat circular setup: Another credential-helper provided the credentials and
+is also used by DataLad. The provider config connecting it to DataLad could be
+intentionally broad. Deleting credentials from keyring but keeping the provider
+config pointing to them, on the other hand,  would be even worse, as it would
+invalidate the provider config and also means rejecting credentials w/o context.
+Hence, this helper doesn't do anything on 'erase' at the moment.
+
+usage: git credential-datalad [option] <action>
+
+options:
+    -h                   Show this help.
+    --non-interactive    Don't ask for user confirmation when storing to
+                         DataLad's credential system. This may fail if default
+                         names result in a conflict with existing ones.
+                         This mode is used for DataLad's CI tests.
+"""
+
+
+def git_credential_datalad():
+    """Entrypoint to query DataLad's credentials via git-credential
+    """
+
+    if not 2 <= len(sys.argv) <= 3 \
+            or sys.argv[1] == "-h" \
+            or sys.argv[1] != "--non-interactive" \
+            or sys.argv[-1] not in ['get', 'store', 'erase']:
+        help_explicit = sys.argv[1] == "-h"
+        print(git_credential_datalad_help,
+              file=sys.stdout if help_explicit else sys.stderr)
+        sys.exit(0 if help_explicit else 1)
+
+    interactive = False if sys.argv[1] == "--non-interactive" else True
+    action = sys.argv[-1]
+    attrs = _credspec2dict(sys.stdin)
+
+    # This helper is intended to be called by git. While git-credential takes a
+    # `url` property of the description, what it passes on is `protocol`, `host`
+    # and potentially `path` (if instructed to useHttpPath by config).
+    # For communication with datalad's credential system, we need to reconstruct
+    # the url, though.
+    assert 'protocol' in attrs.keys()
+    assert 'host' in attrs.keys()
+    if 'url' not in attrs:
+        attrs['url'] = "{protocol}://{host}{path}".format(
+            protocol=attrs.get('protocol'),
+            host=attrs.get('host'),
+            path="/{}".format(attrs.get('path')) if 'path' in attrs else ""
+        )
+
+    # Get datalad's provider configs.
+    providers = Providers.from_config_files()
+
+    if action == 'get':
+        _action_get(attrs, providers)
+    elif action == 'store':
+        _action_store(attrs, interactive, providers)
+
+
+def _action_store(attrs, interactive, providers):
+    # Determine the defaults to use for storing. In non-interactive mode,
+    # this is what's going to be stored, in interactive mode user is
+    # presented with them as default choice.
+    # We don't really know what authentication type makes sense to store in
+    # the provider config (this would be relevant for datalad using those
+    # credentials w/o git).
+    # However, pick 'http_basic_auth' in case of HTTP(S) URL and 'none'
+    # otherwise as the default to pass into store routine.
+    if attrs.get('protocol') in ['http', 'https']:
+        authentication_type = 'http_basic_auth'
+    else:
+        authentication_type = 'none'
+    # If we got a `path` component from git, usehttppath is set and thereby
+    # git was instructed to include it when matching. Hence, do the same.
+    url_re = "{pr}://{h}{p}.*".format(pr=attrs.get('protocol'),
+                                      h=attrs.get('host'),
+                                      p="/" + attrs.get('path')
+                                      if "path" in attrs.keys() else "")
+    name = attrs.get('host')
+    credential_name = name
+    credential_type = "user_password"
+    if not interactive:
+
+        # TODO: What about credential labels? This could already exist as
+        #       well. However, it's unlikely since the respective
+        #       "service name" for keyring is prepended with "datalad-".
+        #       For the same reason of how the keyring system is used by
+        #       datalad it's not very transparently accessible what labels
+        #       we'd need to check for. Rather than encoding knowledge about
+        #       datalad's internal handling here, let's address that in
+        #       datalad's provider and credential classes and have an easy
+        #       check to be called from here.
+        if any(p.name == name for p in providers):
+            print(f"Provider name '{name}' already exists. This can't be "
+                  "resolved in non-interactive mode.",
+                  file=sys.stderr)
+
+        authenticator_class = AUTHENTICATION_TYPES[authentication_type]
+        saved_provider = providers._store_new(
+            url=attrs.get('url'),
+            authentication_type=authentication_type,
+            authenticator_class=authenticator_class,
+            url_re=url_re,
+            name=name,
+            credential_name=credential_name,
+            credential_type=credential_type,
+            level='user'
+        )
+    else:
+        # use backend made for annex special remotes for interaction from a
+        # subprocess whose stdin/stdout are in use for communication with
+        # its parent
+        from datalad.ui import ui
+        ui.set_backend('annex')
+
+        # ensure default is first in list (that's how `enter_new` determines
+        # it's the default)
+        auth_list = AUTHENTICATION_TYPES.copy()
+        auth_list.pop(authentication_type, None)
+        auth_list = [authentication_type] + list(auth_list.keys())
+
+        saved_provider = providers.enter_new(
+            url=attrs.get('url'),
+            auth_types=auth_list,
+            url_re=url_re,
+            name=name,
+            credential_name=credential_name,
+            credential_type=credential_type)
+    saved_provider.credential.set(user=attrs['username'],
+                                  password=attrs['password'])
+
+
+def _action_get(attrs, providers):
+    # query datalad and report if it knows anything, or be silent
+    # git handles the rest
+    provider = providers.get_provider(attrs['url'],
+                                      only_nondefault=True)
+    if provider is None \
+            or provider.credential is None \
+            or not isinstance(provider.credential, UserPassword):
+        # datalad doesn't know or only has a non UserPassword credential for
+        # this URL - create empty entry
+        dlcred = UserPassword()
+    else:
+        dlcred = provider.credential
+    for dlk, gitk in (('user', 'username'), ('password', 'password')):
+        val = dlcred.get(dlk)
+        if val is not None:
+            print('{}={}'.format(gitk, val))
--- a/datalad/local/gitcredential_datalad.py
+++ b/datalad/local/gitcredential_datalad.py
@@ -1,6 +1,9 @@
 import sys
 
-from datalad.downloaders import UserPassword
+from datalad.downloaders import (
+    GitCredential,
+    UserPassword,
+)
 from datalad.downloaders.providers import (
     AUTHENTICATION_TYPES,
     Providers,
@@ -169,6 +172,15 @@ def _action_get(attrs, providers):
         dlcred = UserPassword()
     else:
         dlcred = provider.credential
+        # Safeguard against circular querying. We are a git-credential-helper.
+        # If we find a datalad credential that tells DataLad to query Git, we
+        # need to ignore it. Otherwise we'd end up right here again.
+        if isinstance(provider.credential, GitCredential):
+            # Just return the unchanged description we got from Git
+            for k, v in attrs.items():
+                print('{}={}'.format(k, v))
+            return
+
     for dlk, gitk in (('user', 'username'), ('password', 'password')):
         val = dlcred.get(dlk)
         if val is not None:
--- a/datalad/local/gitcredential_datalad.py
+++ b/datalad/local/gitcredential_datalad.py
@@ -6,7 +6,7 @@ from datalad.downloaders.providers import (
     Providers,
 )
 from datalad.local.gitcredential import _credspec2dict
-from datalad.utils import Path
+from datalad import ConfigManager
 
 git_credential_datalad_help = """\
 Git credential interface to DataLad's credential management system.
@@ -35,10 +35,11 @@ usage: git credential-datalad [option] <action>
 
 options:
     -h                   Show this help.
-    --non-interactive    Don't ask for user confirmation when storing to
-                         DataLad's credential system. This may fail if default
-                         names result in a conflict with existing ones.
-                         This mode is used for DataLad's CI tests.
+
+If DataLad's config variable 'datalad.credentials.githelper.noninteractive' is
+set: Don't ask for user confirmation when storing to DataLad's credential
+system. This may fail if default names result in a conflict with existing ones.
+This mode is used for DataLad's CI tests.
 """
 
 
@@ -46,16 +47,16 @@ def git_credential_datalad():
     """Entrypoint to query DataLad's credentials via git-credential
     """
 
-    if not 2 <= len(sys.argv) <= 3 \
+    if len(sys.argv) != 2 \
             or sys.argv[1] == "-h" \
-            or sys.argv[1] != "--non-interactive" \
-            or sys.argv[-1] not in ['get', 'store', 'erase']:
+            or sys.argv[1] not in ['get', 'store', 'erase']:
         help_explicit = sys.argv[1] == "-h"
         print(git_credential_datalad_help,
               file=sys.stdout if help_explicit else sys.stderr)
         sys.exit(0 if help_explicit else 1)
 
-    interactive = False if sys.argv[1] == "--non-interactive" else True
+    cfg = ConfigManager()
+    interactive = not cfg.obtain("datalad.credentials.githelper.noninteractive")
     action = sys.argv[-1]
     attrs = _credspec2dict(sys.stdin)
 
--- a/datalad/local/gitcredential_datalad.py
+++ b/datalad/local/gitcredential_datalad.py
@@ -11,6 +11,12 @@ from datalad.utils import Path
 git_credential_datalad_help = """\
 Git credential interface to DataLad's credential management system.
 
+In order to use this, one needs to configure git to use the credential helper
+'datalad'. In the simplest case this can be achieved by 'git config --add
+--global credential.helper datalad'. This can be restricted to apply to
+certain URLs only. See 'man gitcredentials' and
+http://docs.datalad.org/credentials.html for details.
+
 Only DataLad's UserPassword-type credentials are supported. This helper
 passes standard 'get', 'store' actions on to the respective interfaces in
 DataLad.
--- a/datalad/local/gitcredential_datalad.py
+++ b/datalad/local/gitcredential_datalad.py
@@ -42,7 +42,10 @@ options:
 If DataLad's config variable 'datalad.credentials.githelper.noninteractive' is
 set: Don't ask for user confirmation when storing to DataLad's credential
 system. This may fail if default names result in a conflict with existing ones.
-This mode is used for DataLad's CI tests.
+This mode is used for DataLad's CI tests. Note, that this config can not
+reliably be read from local configs (a repository's .git/config or
+.datalad/config) as this credential helper when called by Git doesn't get to
+know what repository it is operating on.
 """
 
 
--- a/docs/source/credentials.rst
+++ b/docs/source/credentials.rst
@@ -33,3 +33,7 @@ Valid locations for these files are listed in :ref:`chap_design_credentials`.
 In opposition to Git's approach, `url_re` is a regular expression that matches the entire URL including the scheme.
 The name of the provider section (`data_example_provider`)needs to match the file name.
 The name of the credential section (`data_example_cred`) needs to match the `credential` entry in the provider section.
+
+DataLad will create a provider configuration interactively, whenever it first encounters an URL that requires authentication and no credentials can be matched yet.
+This behavior extends to the git credential helper and can therefore even be triggered by a `git clone`.
+However, `git-credential-datalad` will not do this, if the config variable `datalad.credentials.githelper.noninteractive` is set.
--- a/docs/source/credentials.rst
+++ b/docs/source/credentials.rst
@@ -1,0 +1,35 @@
+Credentials
+***********
+
+Integration with Git
+====================
+
+Git and DataLad can use each other's credential system.
+In order to allow Git to query credentials from DataLad, the git credential helper delivered with DataLad needs to be configured.
+That is, a section like this needs to be part of one's git config file::
+
+  [credential "https://*.data.example.com"]
+    helper = "datalad"
+
+Note:
+
+- This most likely only makes sense at the user or system level (options `--global`|`--system` with `git config`), since cloning of a repository needs the credentials before there is a local repository.
+- The name of that section is a URL matching expression - see `man gitcredentials`.
+- The URL matching does NOT include the scheme! Hence, if you need to match `http` as well as `https`, you need two such entries.
+- Multiple different git credential helpers can be configured - Git will ask them one after another. For example on OSX, Git comes with a helper to use the OSX keychain which is configured at system level. This does not conflict with setting up DataLad's credential helper.
+
+The other way around, DataLad can ask Git for credentials it knows by other means (other git credential helpers).
+For that a datalad provider config needs to be set up in similar fashion::
+
+  [provider:data_example_provider]
+    url_re = http.*://.*data\.example\.com
+    authentication_type = http_basic_auth
+    credential = data_example_cred
+  [credential:data_example_cred]
+    type = git
+
+Such a config lives in a dedicated file named after the provider name (:file:`data_example_provider.cfg` in this case).
+Valid locations for these files are listed in :ref:`chap_design_credentials`.
+In opposition to Git's approach, `url_re` is a regular expression that matches the entire URL including the scheme.
+The name of the provider section (`data_example_provider`)needs to match the file name.
+The name of the credential section (`data_example_cred`) needs to match the `credential` entry in the provider section.
--- a/docs/source/credentials.rst
+++ b/docs/source/credentials.rst
@@ -5,7 +5,14 @@ Integration with Git
 ====================
 
 Git and DataLad can use each other's credential system.
-In order to allow Git to query credentials from DataLad, the git credential helper delivered with DataLad needs to be configured.
+Both directions are independent of each other and none is necessarily required.
+Either direction can be configured based on URL matching patterns.
+In addition, Git can be configured to always query DataLad for credentials without any URL matching.
+
+Let Git query Datalad
+=====================
+
+In order to allow Git to query credentials from DataLad, Git needs to be configured to use the git credential helper delivered with DataLad (an executable called `git-credential-datalad`).
 That is, a section like this needs to be part of one's git config file::
 
   [credential "https://*.data.example.com"]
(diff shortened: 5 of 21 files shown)


mih:
OK, closing for now, until there is time again.

mih:
Makes sense. I am removing the tag, given that you reopened this.

bpoldrack:
@yarikoptic : Can you have a look at this? Specifically, I'm passing the URL we want to use a `Credential` object for to those objects. It's used only by the new `GitCredential`, but the important question is: Did I miss some call path where we could end up asking a `Credential` object when can't know that URL yet? Anything that comes to mind in that regard?

bpoldrack:
Just leaving a record here:

SSH'ing into appveyor's macos build yielded the realization that - at least the currently used version of git on there - doesn't respect `GIT_ASKPASS` at a apparently, but things work out fine via `core.askpass` config.

bpoldrack:
This seems to now work in both directions on al platforms - datalad using credentials provided by `git-credential` and git using credentials provided by datalad. However, there are a couple of restrictions:

1. Only reading the respective credentials is currently tested
2. In order to tell datalad to use credentials provided by git, the respective provider config needs to set a dedicated credential type "git".
3. The other way around, the credential helper `git-credential-datalad` expects the credential label to be the URL, since git doesn't use the concept of a label.

Re 1:
Reasoning for this is, that I think the current system isn't really ready for that usecase and first and foremost useful for using already existing credentials.
a) For datalad using git-credential storing of new credentials isn't straightforward. git expects credentials to be stored for a URL. However, when a potential "store" is triggered via datalad, that would currently mean, something like `datalad download-url` figured there are no credentials yet and ask. It is now not clear at all, what would be the desired context for storing those credentials. Right now it would be that exact URL. That may not be useful at all, since the "next" file has a different one. But we also do not know whether those credentials are meant to apply to the entire host. I think our current credential system isn't up to make that direction really useful.
b) The other way around would be possible, as we are storing via datalad and could get to ask for the entire provider setup. However, testing and documenting this should be dependent on a potential design change to `git-credential-datalad` - see 3.) below

Re 2:
As the usecase for this to say: "Hey, datalad, git knows - just use it", I think this is fine for now. In order to make that less of a special case we need some redesign of our credential approach, I think.

Re 3:
This will likely need to change. For one thing, the different approaches of datalad and git need to be made more compatible. That includes, I think, making our credentials better aware of context by design. I started to allow for this in this PR, but to do it properly requires substantial changes. There is, however, another aspect: Even w/o redesigning our credential system, that helper could/should not only try to query credentials via datalad, but via providers. Meaning: Match the URL in question against the provider definitions and only thereby get a credential label. 


Lastly, and somewhat tangentially relevant: Since any test involving `git-credential` would otherwise lead to hanging git commands on OSX, I changed the test setup to NOT manipulate `HOME` on OSX, but set `GIT_CONFIG_GLOBAL` to point to a different config file. Reason is, that OSX comes with a system level `credential.helper=osxkeychain`, that will be invoked. This one hangs when it can't find the Keychain in its expected pace under `HOME`. I suggest to let us try this approach on OSX only for now, observe whether there are side-effects of not changing `HOME` that the tests don't currently cover and possibly to switch to that approach entirely if we find nothing to be problematic, thus easing things on Windows as well.

bpoldrack:
Failing tests as far as I saw are addressed by PR #6229

bpoldrack:
Verdict from call:

- separate out the change to the test setup on OSX
- support store/delete in both directions

Changing back to draft mode until done.

bpoldrack:
FTR: Current failure on windows is a bit odd. It didn't happen a few days ago. Culprit: location of the to-be created, user-level provider config is off between the test itself and the `git-credential-datalad` helper. Both are querying via the `Providers` class, which in turn is using datalad's AppDirs/platformdirs instance.
From within the test the path is (correctly) `C:\Users\appveyor\AppData\Local\datalad.org\datalad\providers\example.org.cfg`, whereas from within the environment of the credential helper it resolves to `datalad.org\datalad\providers\example.org.cfg`.

adswa:
just a comment from my explorations to run the tests:

The ``git credential-datalad``'s ``--help`` (but not ```-h```) is broken. 

```
(env39) C:\Users\adina\repos\datalad>git credential-datalad --help
fatal: 'C:/Program Files/Git/mingw64/share/doc/git-doc/git-credential-datalad.html': documentation file not found.

(env39) C:\Users\adina\repos\datalad>git credential-datalad -h
[DEBUG] Run ['git', 'version'] (cwd=None)
[DEBUG] Finished ['git', 'version'] with status 0
[DEBUG] Not retro-fitting GitRepo with deprecated symbols, datalad-deprecated package not found
[DEBUG] parseParameters: Given "exclude_special_remotes: bool, optional
|   if True, don't return annex special remotes", we split into [('exclude_special_remotes', "exclude_special_remotes: bool, optional\n  if True, don't return annex special remotes")]
[DEBUG] parseParameters: Given "with_urls_only : bool, optional
|   return only remotes which have urls", we split into [('with_urls_only', 'with_urls_only : bool, optional\n  return only remotes which have urls')]
[DEBUG] Failed to import requests_ftp, thus no ftp support: ModuleNotFoundError(No module named 'requests_ftp')
[DEBUG] parseParameters: Given "headers: dict, optional
|   Header fields to be provided to the session. Unless User-Agent provided, a custom
|   one, available in `DEFAULT_USER_AGENT` constant of this module will be used.", we split into [('headers', 'headers: dict, optional\n  Header fields to be provided to the session. Unless User-Agent provided, a custom\n  one, available in `DEFAULT_USER_AGENT` constant of this module will be used.')]
[DEBUG] parseParameters: Given "credential: Credential, optional
|   Provides necessary credential fields to be used by authenticator
| authenticator: Authenticator, optional
|   Authenticator to use for authentication.", we split into [('credential', 'credential: Credential, optional\n  Provides necessary credential fields to be used by authenticator'), ('authenticator', 'authenticator: Authenticator, optional\n  Authenticator to use for authentication.')]
[DEBUG] parseParameters: Given "", we split into []
[DEBUG] parseParameters: Given "credential: Credential, optional
|   Provides necessary credential fields to be used by authenticator
| authenticator: Authenticator, optional
|   Authenticator to use for authentication.", we split into [('credential', 'credential: Credential, optional\n  Provides necessary credential fields to be used by authenticator'), ('authenticator', 'authenticator: Authenticator, optional\n  Authenticator to use for authentication.')]
[DEBUG] parseParameters: Given "", we split into []
[DEBUG] parseParameters: Given "headers: dict, optional
|   Header fields to be provided to the session. Unless User-Agent provided, a custom
|   one, available in `DEFAULT_USER_AGENT` constant of this module will be used.
| credential: Credential, optional
|   Provides necessary credential fields to be used by authenticator
| authenticator: Authenticator, optional
|   Authenticator to use for authentication.", we split into [('headers', 'headers: dict, optional\n  Header fields to be provided to the session. Unless User-Agent provided, a custom\n  one, available in `DEFAULT_USER_AGENT` constant of this module will be used.'), ('credential', 'credential: Credential, optional\n  Provides necessary credential fields to be used by authenticator'), ('authenticator', 'authenticator: Authenticator, optional\n  Authenticator to use for authentication.')]
[DEBUG] parseParameters: Given "", we split into []
[DEBUG] parseParameters: Given "method : callable
|   A callable, usually a method of the same class, which we decorate
|   with access handling, and pass url as the first argument
| url : string
|   URL to access
| *args, **kwargs
|   Passed into the method call", we split into [('method', 'method : callable\n  A callable, usually a method of the same class, which we decorate\n  with access handling, and pass url as the first argument'), ('url', 'url : string\n  URL to access'), ('*args, **kwargs', '*args, **kwargs\n  Passed into the method call')]
Git credential interface to DataLad's credential management system.

Only DataLad's UserPassword-type credentials are supported. This helper
passes standard 'get', 'store', and 'erase' actions on to the respective
interfaces in DataLad.

usage: git credential-datalad [option] <action>

options:
    -h/--help            Show this help.
    --non-interactive    Don't ask for user confirmation when storing to
                         DataLad's credential system. This may fail if default
                         names result in a conflict with existing ones.
                         This mode is used for DataLad's CI tests.
```

adswa:
As noted in the chat, the current failure that is based on configurations leaking into the test dataset instead of ``%APPDATA_LOCAL%`` is fixed locally for me by #6260.

adswa:
Also as noted in the chat, there ``provider.credentials.delete()`` has a problem that I think prevents the credentials from being removed from the keyring. The traceback looks like this:
```Traceback (most recent call last):
  File "c:\users\adina\repos\datalad\datalad\support\keyring_.py", line 83, in delete
    return self.__keyring.delete_password(self._get_service_name(name), field)
AttributeError: 'NoneType' object has no attribute 'delete_password'

  File "C:\Users\adina\env39\Scripts\git-credential-datalad-script.py", line 33, in <module>
    sys.exit(load_entry_point('datalad', 'console_scripts', 'git-credential-datalad')())
  File "c:\users\adina\repos\datalad\datalad\local\gitcredential_datalad.py", line 164, in git_credential_datalad
    provider.credential.delete()
  File "c:\users\adina\repos\datalad\datalad\downloaders\credentials.py", line 223, in delete
    self._keyring.delete(self.name, f)
  File "c:\users\adina\repos\datalad\datalad\support\keyring_.py", line 84, in delete
    except self.__keyring_mod.errors.PasswordDeleteError as exc:
AttributeError: 'NoneType' object has no attribute 'errors'
```

Locally, I see a final AssertionError in this test for the final two checks that check for blank username and password after their deletion, but I believe due to the error above they are kept, and the Assertions fail.

```
c:\users\adina\repos\datalad\datalad\local\tests\test_gitcredential.py(161)test_datalad_credential_helper()
-> eq_(dlcred.get('user'), None)
(Pdb) n
AssertionError: 'dl-user' != None
```

bpoldrack:
@adswa 

> The git credential-datalad's --help (but not -h) is broken.

Thx! I didn't notice so far, that `git` expects an html to be there. (In other words: It's not exactly `--help` of the helper that's broken, but that it doesn't get to see that option in reality.)

They keyring issue was fixed already. I draw it out of this PR into its own.

> As noted in the chat, the current failure that is based on configurations leaking into the test dataset instead of %APPDATA_LOCAL% is fixed locally for me by #6260.

Confirmed. (This is seriously fucked up, though!)
Will not merge that into this PR for now, but make it wait for the fate of yours.



> sorry, I just force pushed my branch to cf3f005. Two tests that passed locally failed on Appveyor, so I disabled them again. You'll probably see them fail in appveyor... sorry

Hehe. I was just wondering what the hell did I do when I tried to apply your branch locally :-D
Had the original branch in my fork, though - just a reset - all good.

docs/source/design/credentials.rst
```diff
@@ -28,11 +33,57 @@ These locations are tried in the following order, and the first successful looku
    As with any other specification of configuration items, environment variables can be used to set or override credentials.
    Variable names take the form of ``DATALAD_CREDENTIAL_<NAME>_<COMPONENT>``, and standard replacement rules into configuration variable names apply.
 
-2. DataLad uses the `keyring` package https://pypi.org/project/keyring to connect to any of its supported back-ends for setting or getting credentials.
-   This provides support for credential storage on all major platforms, but also extensibility, providing 3rd-parties to implement and use specialized solutions.
+2. DataLad uses the `keyring` package https://pypi.org/project/keyring to connect to any of its supported back-ends for setting or getting credentials,
+   via a wrapper in ``datalad.support.keyring_``. This provides support for credential storage on all major platforms, but also extensibility, providing 3rd-parties to implement and use specialized solutions.
```
yarikoptic, line 37: minor: should we use sphinx's cross-referencing while listing modules? e.g.
```suggestion
   via a wrapper in :mod:`~datalad.support.keyring_`. This provides support for credential storage on all major platforms, but also extensibility, providing 3rd-parties to implement and use specialized solutions.
```
so we get hyperlinks and also failed doc build if some module gets moved/deprecated?

docs/source/design/credentials.rst
```diff
@@ -28,11 +33,57 @@ These locations are tried in the following order, and the first successful looku
...
 
 When a credential is required for operation, but could not be obtained via any of the above approaches, DataLad can prompt for credentials in interactive terminal sessions.
 Interactively entered credentials will be stored in the active credential store available via the ``keyring`` package.
+Note, however, that the keyring approach is somewhat abused by datalad. The wrapper only uses ``get_/set_password`` of ``keyring`` with the credential's ``FIELDS`` as the name to query (essentially turning the keyring into a plain key-value store) and "datalad-<CREDENTIAL-LABEL>" as the "service name".
+With this approach it's not possible to use credentials in a system's keyring that were created by other means.
```
yarikoptic, line 42: not entirely clear what "other means" means here.  I can go and add credential with that name using regular mechanisms.  I guess more specific would be
```suggestion
With this approach it's not possible to use credentials in a system's keyring that were defined by other, datalad unaware software.
```
or alike?

docs/source/design/credentials.rst
```diff
@@ -28,11 +33,57 @@ These locations are tried in the following order, and the first successful looku
...
 
 When a credential value is known but invalid, the invalid value must be removed or replaced in the active credential store.
 By setting the configuration flag ``datalad.credentials.force-ask``, DataLad can be instructed to force interactive credential re-entry to effectively override any store credential with a new value.
+
+Providers
+=========
+
+Providers are associating credentials with a context for using them and are defined by configuration files. A single provider is represented by :class:`Provider` object and the list of available providers is represented by the :class:`Providers` class.
```
yarikoptic, line 50: unrelated to content -- good strategy (as we did e.g. for joss paper) is to have one sentence per line. E.g. here
```suggestion
Providers are associating credentials with a context for using them and are defined by configuration files.
A single provider is represented by :class:`Provider` object and the list of available providers is represented by the :class:`Providers` class.
```

docs/source/design/credentials.rst
```diff
@@ -28,11 +33,57 @@ These locations are tried in the following order, and the first successful looku
...
+A provider is identified by a label and stored in a dedicated config file per provider named `LABEL.cfg`. Such a file can reside in a dataset (under `.datalad/providers/`), at the user level (under `user_config_dir/providers`), at the system level (under `site_config_dir/providers`) or come packaged with the datalad distribution (in directory `configs` next to `providers.py`).
```
yarikoptic, line 51:
```suggestion
A provider is identified by a label and stored in a dedicated config file per provider named `LABEL.cfg`.
Such a file can reside in a dataset (under `.datalad/providers/`), at the user level (under `{user_config_dir}/providers`), at the system level (under `{site_config_dir}/providers`) or come packaged with the datalad distribution (in directory `configs` next to `providers.py`).
```
split lines, and put those `*config_dir` into `{}` to signal that they are not actual names of the directories.  This seems to be the first instance of mentioning `config_dir` in the `docs/`.  I wonder if we may be should define a `:term:` for each one of them or anyhow describe them somewhere and reference to here?

docs/source/design/credentials.rst
```diff
@@ -28,11 +33,57 @@ These locations are tried in the following order, and the first successful looku
...
+Such a provider specifies a regular expression to match URLs against and assigns authenticator abd credentials to be used for a match.
```
yarikoptic, line 52:
```suggestion
Such a provider specifies a regular expression to match URLs against and assigns authenticator and credentials to be used for a match.
```

docs/source/design/credentials.rst
```diff
@@ -28,11 +33,57 @@ These locations are tried in the following order, and the first successful looku
...
+Credentials are referenced by their label, which in turn is the name of another section in such a file specifying the type of the credential.
+References to credential and authenticator types are strings that are mapped to classes by the following dict definitions:
+
+- ``datalad.downloaders.AUTHENTICATION_TYPES``
+- ``datalad.downloaders.CREDENTIAL_TYPES``
+
+Available providers can be loaded by ``Providers.from_config_files`` and ``Providers.get_provider(url)`` will match a given URL against them and return the appropriate `Provider` instance.
+A :class:`Provider` object will determine a downloader to use (derived from :class:`BaseDownloader`), based on the URL's protocol.
+
+Note, that the provider config files are not currently following datalad's general config approach. Instead they are special config files, read by ``SafeConfigParser`` that are not compatible with `git-config` and hence the :class:`ConfigManager`.
```
yarikoptic, line 62:
```suggestion
Note, that the provider config files are not currently following datalad's general config approach.
Instead they are special config files, read by :class:`configparser.ConfigParser` that are not compatible with `git-config` and hence the :class:`ConfigManager`.
```
?

docs/source/design/credentials.rst
```diff
@@ -28,11 +33,57 @@ These locations are tried in the following order, and the first successful looku
...
+Since git's way to deal with credentials doesn't involve labels but only matching URLs, it is - in some sense - the equivalent of datalad's provider layer. However, providers don't talk to a backend, credentials do.
+Hence, a more seemless integration requires some changes in the design of datalad's credential system as a whole.
+
+In the opposite direction - making git aware of datalad's credentials, there's no special casing, though. Datalad comes with a `git-credential-datalad` executable. Whenever git is configured to use it by setting `credential.helper=datalad`,
+it will be able to query datalad's credential system for a provider matching the URL in question and retrieve the referenced by this provider credentials. This helper is also capable of storing new provider+credentials when asked to do so by git.
+It can this interactively, asking a user to confirm/change that config or - if `credential.helper=datalad --non-interactive` - try to non-interactively store with its defaults.
```
yarikoptic, line 76:
```suggestion
In the opposite direction - making git aware of datalad's credentials, there's no special casing, though.
Datalad comes with a `git-credential-datalad` executable.
Whenever git is configured to use it by setting `credential.helper=datalad`, it will be able to query datalad's credential system for a provider matching the URL in question and retrieve the referenced by this provider credentials.
This helper can also store a new provider+credentials when asked to do so by git.
It can do this interactively, asking a user to confirm/change that config or - if `credential.helper='datalad --non-interactive'` - try to non-interactively store with its defaults.
```
am I right in adding `'` around the value for the helper?

docs/source/design/credentials.rst
bpoldrack: Yes! Thanks.

docs/source/design/credentials.rst
bpoldrack: Well, `git` wouldn't be the best example, since we now have a way to circle through git-credential even if that is ending up in a keyring again. But agree - will change wording.

docs/source/design/credentials.rst
bpoldrack: > I wonder if we may be should define a :term: for each one of them or anyhow describe them somewhere and reference to here?

True, that would be good.

docs/source/design/credentials.rst
bpoldrack: You are.

docs/source/design/credentials.rst
bpoldrack: Will do.

bpoldrack:
This still needs an update of the desription and confirmation by the Appveyor builds. Somehow AppVeyor currently fails to pick up this PR - trying to trigger it.

mih:
Appveyor build has completed

mih:
@bpoldrack has this been tested on juseless and confirmed to work for clone+get from an password-protected RIA store?

datalad/downloaders/providers.py
```diff
@@ -363,10 +366,90 @@ def get_provider(self, url, only_nondefault=False, return_all=False):
...
+        url_re: str
+          regular expression; Once created, this config will be used for any
+          matching URL; defaults to `url`
+        name: str
+          name for the provider; needs to be unique per user
+        credential_name: str
+          name for the credential; defaults to the provider's name
+        credential_type: str
+          credential type to use (key for datalad.downloaders.CREDENTIAL_TYPES)
+        """
```
mih, line 449: Thank you!

mih (commented):
I have to interupt my review for now. I'll pick it up later.

.appveyor.yml
```diff
@@ -220,6 +227,13 @@ init:
   # Identity setup
   - git config --global user.email "test@appveyor.land"
   - git config --global user.name "Appveyor Almighty"
+  # Don't let git ask for credentials in CI runs. Note, that this variable
+  # technically is not a flag, but an executable (which is why name and value
+  # are a bit confusing here - we just want a no-op basically). The environment
+  # variable GIT_ASKPASS overwrites this, but neither env var nor this config
+  # are supported by git-credential on all systems and git versions (most recent
+  # ones should work either way, though). Hence use both across CI builds.
```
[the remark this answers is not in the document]
bpoldrack: I'm fine with this shortened version. Will do.

bpoldrack:
@mih 

> @bpoldrack has this been tested on juseless and confirmed to work for clone+get from an password-protected RIA store?

Yes, https://github.com/datalad/datalad/pull/5796#issuecomment-948195372 stands. Obviously, I could only test a user-level provider config, but system-level as described there should work, too. Just confirmed on current tip of branch w/ that provider config at user-level:

```
❱ datalad clone "ria+http://studyforrest-test.ds.inm7.de#3a8648b3-7df8-413f-8efb-4d39040ac174" studyforrest_test_auth
[INFO   ] [INFO] Fetching 'http://studyforrest.ds.inm7.de/ria-layout-version'                                                                                            
[INFO   ] [INFO] http session: Authenticating into session for http://studyforrest.ds.inm7.de/ria-layout-version 
[INFO   ] [INFO] Fetching 'http://studyforrest.ds.inm7.de/3a8/648b3-7df8-413f-8efb-4d39040ac174/ria-layout-version' 
[INFO   ] Fetching 'http://studyforrest-test.ds.inm7.de/3a8/648b3-7df8-413f-8efb-4d39040ac174/config' 
[INFO   ] http session: Authenticating into session for http://studyforrest-test.ds.inm7.de/3a8/648b3-7df8-413f-8efb-4d39040ac174/config 
[INFO   ] Reconfigured inm7-storage for ria+http://studyforrest-test.ds.inm7.de 
[INFO   ] Fetching 'http://studyforrest-test.ds.inm7.de/3a8/648b3-7df8-413f-8efb-4d39040ac174/config' 
[INFO   ] Configure additional publication dependency on "inm7-storage" 
install(ok): /home/bpoldrack/studyforrest_test_auth (dataset)
action summary:
  configure-sibling (ok: 1)
  install (ok: 1)

❱ git annex whereis sub-04/ses-r08/func/sub-04_ses-r08_task-coverage_bold.nii.gz
whereis sub-04/ses-r08/func/sub-04_ses-r08_task-coverage_bold.nii.gz [INFO] Fetching 'http://studyforrest-test.ds.inm7.de/ria-layout-version' 
[INFO] http session: Authenticating into session for http://studyforrest-test.ds.inm7.de/ria-layout-version 
[INFO] Fetching 'http://studyforrest-test.ds.inm7.de/3a8/648b3-7df8-413f-8efb-4d39040ac174/ria-layout-version' 
(3 copies) 
  	38df1f7e-a0b9-418d-a164-6ba5b00d871f -- [inm7-storage]
   	bca0599f-7ec4-4bdc-b96b-fc5d0f5dedfa -- studyforrest multires7T original source repository
   	ddead121-b69b-4399-b981-8275a40ce4ef -- [mddatasrc]

  inm7-storage: http://studyforrest-test.ds.inm7.de/3a8/648b3-7df8-413f-8efb-4d39040ac174/annex/objects/4x/Xv/MD5E-s281718642--2593650da696d2f84ceae20209e0608a.nii.gz/MD5E-s281718642--2593650da696d2f84ceae20209e0608a.nii.gz
ok

❱ git annex get -f inm7-storage sub-04/ses-r08/func/sub-04_ses-r08_task-coverage_bold.nii.gz
get sub-04/ses-r08/func/sub-04_ses-r08_task-coverage_bold.nii.gz (from inm7-storage...) 
[INFO] Fetching 'http://studyforrest-test.ds.inm7.de/ria-layout-version' 
[INFO] http session: Authenticating into session for http://studyforrest-test.ds.inm7.de/ria-layout-version 
[INFO] Fetching 'http://studyforrest-test.ds.inm7.de/3a8/648b3-7df8-413f-8efb-4d39040ac174/ria-layout-version' 
[INFO] Downloading 'http://studyforrest-test.ds.inm7.de/3a8/648b3-7df8-413f-8efb-4d39040ac174/annex/objects/4x/Xv/MD5E-s281718642--2593650da696d2f84ceae20209e0608a.nii.gz/MD5E-s281718642--2593650da696d2f84ceae20209e0608a.nii.gz' into '.git/annex/tmp/MD5E-s281718642--2593650da696d2f84ceae20209e0608a.nii.gz' 
(checksum...) ok
(recording state in git...)
```

Note, that the point is ORA using authentication and there are additional sources available, hence testing with `git-annex-get` rather than just `datalad get`, which may get the content from elsewhere.

datalad/downloaders/providers.py
```diff
@@ -439,12 +525,13 @@ def enter_new(self, url=None, auth_types=[]):
...
+            default=credential_type or getattr(authenticator_class,
+                                               'DEFAULT_CREDENTIAL_TYPE')
         )
 
-        # Just create a configuration file and reload the thing
-        if not path.lexists(providers_user_dir):
-            os.makedirs(providers_user_dir)
+        # TODO: This string is duplicated in self._store_new ATM. However,
+        #       that entire business of dealing with these config files should
+        #       go into a more sophisticated config parser/writer
```
mih, line 534: Why not define the template in a variable and use it twice? Seems to improve maintainability without requiring a config syntax/parser change.

datalad/local/gitcredential_datalad.py
```diff
@@ -0,0 +1,181 @@
...
+interfaces in DataLad.
+
+usage: git credential-datalad [option] <action>
+
+options:
+    -h                   Show this help.
+    --non-interactive    Don't ask for user confirmation when storing to
+                         DataLad's credential system. This may fail if default
+                         names result in a conflict with existing ones.
+                         This mode is used for DataLad's CI tests.
```
mih, line 25: I would prefer to not expose such switches in the API. An envvar should be as, or even more, flexible in terms of triggering test-related special modes inside some other application.

datalad/local/gitcredential_datalad.py
```diff
@@ -0,0 +1,181 @@
...
+        attrs['url'] = "{protocol}://{host}{path}".format(
+            protocol=attrs.get('protocol'),
+            host=attrs.get('host'),
+            path="/{}".format(attrs.get('path')) if 'path' in attrs else ""
+        )
+
+    # Get datalad's provider configs.
+    providers = Providers.from_config_files()
+
+    if action == 'get':
```
mih, line 63: What do you think about de-spaghettifying this monolithic helper and put its actions into dedicated helper functions that could have a dedicated unittest without having to go through the complexity of the entire system for testing?

docs/source/design/credentials.rst
```diff
@@ -12,13 +12,18 @@ Credential management
    This specification describes the current implementation.
 
 Various components of DataLad need to be passed credentials to interact with services that require authentication. 
-This includes downloading files, but also things like REST API usage.
+This includes downloading files, but also things like REST API usage or authenticated cloning.
+Relevant components of Datalad's credential management to look at are credentials types, providers, authenticators and downloaders.
```
mih, line 16:
```suggestion
Key components of Datalad's credential management are credentials types, providers, authenticators and downloaders.
```

docs/source/design/credentials.rst
```diff
@@ -12,13 +12,18 @@ Credential management
...
+
+Credentials
+===========
 
 Supported credential types include basic user/password combinations, access tokens, and a range of tailored solutions for particular services.
 All credential type implementations are derived from a common :class:`Credential` base class.
+When referenced by a string - such as an entry from a config file - ``datalad.downloaders.CREDENTIAL_TYPES`` defines a mapping of those strings to credential classes.
```
mih, line 23:
```suggestion
A mapping from string labels to credential classes is defined in ``datalad.downloaders.CREDENTIAL_TYPES``.
```

docs/source/design/credentials.rst
```diff
@@ -28,11 +33,67 @@ These locations are tried in the following order, and the first successful looku
...
+A :class:`Provider` object will determine a downloader to use (derived from :class:`BaseDownloader`), based on the URL's protocol.
+
+Note, that the provider config files are not currently following datalad's general config approach.
+Instead they are special config files, read by :class:`configparser.ConfigParser` that are not compatible with `git-config` and hence the :class:`ConfigManager`.
+
+There are currently two ways of storing a provider and thus creating its config file: ``Providers.enter_new`` and ``Providers._store_new``.
+The former will only work interactively and provide the user with options to choose from, while the latter is non-interactive and can therefore only be used, when all properties of the provider config are known and passed to it.
+There's no way at the moment to store an existing :class:`Provider` object directly.
+
+Integration with git
```
mih, line 73:
```suggestion
Integration with Git
```

mih: and elsewhere...

docs/source/design/credentials.rst
```diff
@@ -28,11 +33,67 @@ These locations are tried in the following order, and the first successful looku
...
+
+Authenticators are used by downloaders to issue authenticated requests.
+They are not easily available to directly be applied to requests being made outside of the downloaders.
+
+TODO
+
+Downloaders
+===========
+
+TODO
```
mih, line 99: I am not confident that there is value in leaving these TODO markers. We could have 10k of them. Why here?

mih (commented):
I did a full pass now. This is an important PR that sheds light into the darker corners of the code base. The added documentation alone is worth their weight in gold. Thx!

ND: At some point I would like to understand, why Python `keyring` is so crippled. It does support more than `set|get_password()` (which is the piece we use), but it is hard to comprehend where the counterpart to `get_credential()` can be found.

I would personally prefer to cut the history of this PR short. It primarily records an iterative approach to the final result. I am almost under the impression that the entire change set could be three commits. Plumbing changes, plus the two interfaces.

loj:
I gave this a try in the context of the clone+get from a password-protected RIA store on juseless, and it seems to work. I tested with the user-level provider config that @bpoldrack provided in #5796 (comment), and setup non-interactive in my user-level `.gitconfig` via
```
[credential "http://*.ds.inm7.de"]
    helper = "/usr/local/bin/inm7_datastore_pw"
    helper = "datalad --non-interactive"
[credential "https://*.ds.inm7.de"]
    helper = "/usr/local/bin/inm7_datastore_pw"
    helper = "datalad --non-interactive"
```
I also tested what it would look like if a user does not have access to a particular RIA store. It states that authentication failed, but the messaging is a bit confusing, especially when it talks about trying to delete the credential.
```
❱ datalad clone "ria+http://studyforrest-test.ds.inm7.de#3a8648b3-7df8-413f-8efb-4d39040ac174" studyforrest_test
install(error): /home/loj/tmp/studyforrest_test (dataset) [Failed to clone from any candidate source URL. Encountered errors per each url were:
- http://studyforrest-test.ds.inm7.de/3a8/648b3-7df8-413f-8efb-4d39040ac174
  CommandError: 'git -c diff.ignoreSubmodules=none clone --progress http://studyforrest-test.ds.inm7.de/3a8/648b3-7df8-413f-8efb-4d39040ac174 /home/loj/tmp/studyforrest_test' failed with exitcode 128 [err: 'Cloning into '/home/loj/tmp/studyforrest_test'...
fatal: refusing to work with credential missing host field
Unable to delete provider config /home/loj/.config/datalad/providers/inm7_ds.cfg: [Errno 2] No such file or directory: '/home/loj/.config/datalad/providers/inm7_ds.cfg'
Unable to delete credential: CommandError: 'git credential reject' failed with exitcode 128
fatal: Authentication failed for 'https://studyforrest-test.ds.inm7.de/3a8/648b3-7df8-413f-8efb-4d39040ac174/'']
- http://studyforrest-test.ds.inm7.de/3a8/648b3-7df8-413f-8efb-4d39040ac174/.git
  CommandError: 'git -c diff.ignoreSubmodules=none clone --progress http://studyforrest-test.ds.inm7.de/3a8/648b3-7df8-413f-8efb-4d39040ac174/.git /home/loj/tmp/studyforrest_test' failed with exitcode 128 [err: 'Cloning into '/home/loj/tmp/studyforrest_test'...
fatal: refusing to work with credential missing host field
Unable to delete provider config /home/loj/.config/datalad/providers/inm7_ds.cfg: [Errno 2] No such file or directory: '/home/loj/.config/datalad/providers/inm7_ds.cfg'
Unable to delete credential: CommandError: 'git credential reject' failed with exitcode 128
fatal: Authentication failed for 'https://studyforrest-test.ds.inm7.de/3a8/648b3-7df8-413f-8efb-4d39040ac174/.git/'']]
```

bpoldrack:
Thx, @loj ! Didn't catch the deletion issue. Looking into it.

docs/source/design/credentials.rst
bpoldrack: Right. Wasn't meant to be committed. Wanted to enhance those aspects outside of this PR.

datalad/local/gitcredential_datalad.py
bpoldrack: Yes, I agree. Will do.

datalad/downloaders/providers.py
bpoldrack: Honestly forgot why I decided against it. Will do (or re-discover why).

datalad/downloaders/providers.py
```diff
@@ -261,6 +262,8 @@ def from_config_files(cls, files=None, reload=False):
                     raise ValueError("Unknown credential %s. Known are: %s"
                                      % (provider.credential, ", ".join(credentials.keys())))
                 provider.credential = credentials[provider.credential]
+                # TODO: Is this the right place to pass dataset to credential?
+                provider.credential.set_context(dataset=cls._DS_ROOT)
```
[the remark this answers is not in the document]
bpoldrack: Wasn't sure about two things:
1. There could be call paths from high-level code that wouldn't even go through this function. In that case: Not a wrong place but insufficient.
2. There could be call paths where a dataset is figured out in more up-to-date style already (here it basically relies on PWD), and I wouldn't want to overwrite here.

This is all about employing the credential system from crawler, download-url and whatever else and what exactly are this code's entry points to that system. It's not obvious from in here whether `from_config_files` would be called from within a datalad context, when a dataset was already figured. That's one aspect of that part of the code base that is quite outdated and needs overhauling.

datalad/downloaders/providers.py
```diff
@@ -38,6 +38,7 @@
 from ..utils import auto_repr
 from ..utils import ensure_list_from_str
 from ..utils import get_dataset_root
+from ..utils import Path
```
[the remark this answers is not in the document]
bpoldrack: Yes, will add an import clean-up.

datalad/downloaders/credentials.py
```diff
@@ -388,10 +447,71 @@ def __init__(self, name, url=None, keyring=None):
...
+            return from_cfg
+
+        # Note:
+        # In opposition to the keyring approach of other `Credential`s,
+        # we don't query for single values, but for an entire "description".
+        # Currently required methods have to return single values, though.
+        # Hence, calls to `git credential fill` could be optimised. Not easy to
+        # assess when exactly this class can know whether it's context is yet to
+        # be completed, though, so that another `fill` would actually yield
+        # something different than before.
```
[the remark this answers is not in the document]
bpoldrack: I don't think so, no. A dataset may need to be set (and in fact is) at a point where there's no URL to be set (yet). We don't have that one point in the code where al required information is available to be passed into a `Credential` object ATM.

mih:
What are the chances that we can get this merged before the holidays?

mih:
This is now officially the oldest, still open, PR

bpoldrack:
@loj : The "unable to delete" messages should be gone now. The `CommandError` would need addressing elsewhere.

@mih : I removed the 'erase' action entirely from the git credential helper. Rationale: https://github.com/datalad/datalad/pull/5796/commits/ae9bde33b1f8f41db477a26608a38d27143fc37d. Objections?
Meanwhile I'll clean up and squash ...

datalad/downloaders/credentials.py
```diff
@@ -388,10 +447,71 @@ def __init__(self, name, url=None, keyring=None):
...
+
+
+    _FIELDS = OrderedDict([('user', {}),
+                           ('password', {'hidden': True})])
+
+    # substitute keys used within datalad by the ones used by git-credential
+    _FIELDS_GIT = {'user': 'username',
+                   'password': 'password'}
+
+    is_expired = False  # no expiration provisioned
```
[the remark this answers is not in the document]
bpoldrack: Stuck with the wording of other `Credential` classes. Don't think it matters really. These classes don't provision it, or when using these classes no such functionality is supported.

bpoldrack:
Hopefully didn't screw up the history rewrite. If tests pass, should be ready now, @mih

datalad/local/gitcredential_datalad.py
mih: @bpoldrack Are you planning on changing this? Or is there a need for it?

mih (commented):
I still think this PR would benefit from importing information from the PR discussion into the diff. In comparison to the last time I read it, it has much more background info, but I feel it is still hard to understand how to actually use the features in here. The PR discussion has concrete examples that would help here.

Along the same line, the help of the git credential helper would benefit from info what it will actually do. It is lacking information on what a user needs to pay attention to when things need to be altered/erased. From the description it is clear that this is a complex problem, but there is no indication where to start looking for answers -- or even a hint that this helper further complicates the situation by also adding provider configs somewhere.

mih:
Any updates @bpoldrack ?

mih:
For some reason, one conversation in the diff panel is not showing up here, and probably therefore doesn't have a response. Can you please take a look @bpoldrack 

image

docs/source/credentials.rst
```diff
@@ -0,0 +1,37 @@
...
+====================
+
+Git and DataLad can use each other's credential system.
+In order to allow Git to query credentials from DataLad, the git credential helper delivered with DataLad needs to be configured.
+That is, a section like this needs to be part of one's git config file::
+
+  [credential "http://*.ds.inm7.de"]
+    helper = "datalad"
+  [credential "https://*.ds.inm7.de"]
+    helper = "datalad"
```
mih, line 14: This makes the impression there is always a need to have two config section pairs, but this just for this particular case

```suggestion
  [credential "https://*.data.example.com"]
    helper = "datalad"
```

docs/source/credentials.rst
```diff
@@ -0,0 +1,37 @@
...
+
+The other way around, DataLad can ask Git for credentials it knows by other means (other git credential helpers).
+For that a datalad provider config needs to be set up in similar fashion::
+
+  [provider:inm7_ds]
+    url_re = http.*://.*ds\.inm7\.de
+    authentication_type = http_basic_auth
+    credential = inm7_ds
+  [credential:inm7_ds]
+    type = git
```
mih, line 31:
```suggestion
  [provider:data_example_com]
    url_re = http.*://.*data\.example\.com
    authentication_type = http_basic_auth
    credential = data_example_com
  [credential:data_example_com]
    type = git
```

docs/source/credentials.rst
```diff
@@ -0,0 +1,37 @@
...
+
+Such a config lives in a dedicated file named after the provider name (`inm7_ds.cfg` in this case).
```
mih, line 33:
```suggestion
Such a config lives in a dedicated file named after the provider name (`data_example_com.cfg` in this case).
```

docs/source/credentials.rst
```diff
@@ -0,0 +1,37 @@
...
+Valid locations for these files are in a dataset (under `.datalad/providers/`), at the user level (under `{user_config_dir}/providers`), at the system level (under `{site_config_dir}/providers`) or come packaged with the datalad distribution (in directory `configs` next to `providers.py`).
```
mih, line 34: I would vote for not duplicating this info here, but to reference the design doc for locations.

mih (commented):
Thx for the updated. I left a few comments and suggestions for changes.

datalad/local/gitcredential_datalad.py
```diff
@@ -0,0 +1,174 @@
...
+from datalad.utils import Path
+
+git_credential_datalad_help = """\
+Git credential interface to DataLad's credential management system.
+
+In order to use this, one needs to configure git to use the credential helper
+'datalad'. In the simplest case this can be achieved by 'git config --add
+--global credential.helper datalad'. This can be restricted to apply to
+certain URLs only. See 'man gitcredentials' and
+http://docs.datalad.org/en/stable/credentials.html for details.
```
mih, line 18: http://docs.datalad.org/credentials.html would be better. It would take a while till that URL starts working.

bpoldrack:
Thx, @mih

> For some reason, one conversation in the diff panel is not showing up here, and probably therefore doesn't have a response. Can you please take a look @bpoldrack

Huh. Seeing this for the first time indeed.
My thinking was, that there may be more usecases than just the CI, that may make it necessary to be part of the API, since that way one could specify that mode specifically for the helper calls per URL matching pattern. This would allow to act (scripted and) non-interactive on some but in regular, interactive mode on other services. Whether that's needed in any real setup - I don't know. And, obviously, it's still achievable via the env var approach. So, yeah - I'm fine taking it out of there.

mih:
> Huh. Seeing this for the first time indeed. My thinking was, that there may be more usecases than just the CI, that may make it necessary to be part of the API, since that way one could specify that mode specifically for the helper calls per URL matching pattern. This would allow to act (scripted and) non-interactive on some but in regular, interactive mode on other services. Whether that's needed in any real setup - I don't know. And, obviously, it's still achievable via the env var approach. So, yeah - I'm fine taking it out of there.

I see the motivation, but not the conclusion. This is an extension of the credential-helper API to communicate a specific mode of operation. But we have the entire config system to make arbitrarily complex configurations accessible without a custom API extension. The helper already imports from `datalad`, hence the config is fully operational at no additional cost. `git` provides `git -c ... credential-datalad` to cover the cmdline usecase, and datalad add the ENV vars on top. Given all that already available what does the custom extension add?

docs/source/credentials.rst
```diff
@@ -0,0 +1,39 @@
+Credentials
+***********
+
+Integration with Git
+====================
+
+Git and DataLad can use each other's credential system.
```
aqw, line 7: It's probably good to mention up from whether either method can be used, or if /both/ are needed and work in tandem.

After reading all of this file, I realized I was unsure about whether both configurations are needed, or if it's one vs the other. I came back to the beginning hoping for a clarification, but didn't find it.

By now, I'm pretty convinced that both are needed, but it feels circular (git says talk to DataLad and DataLad says talk to git). The opening text should make the situation clear.

docs/source/credentials.rst
```diff
@@ -0,0 +1,39 @@
...
+- The name of that section is a URL matching expression - see `man gitcredentials`.
+- The URL matching does NOT include the scheme! Hence, if you need to match `http` as well as `https`, you need two such entries.
+- Multiple different git credential helpers can be configured - Git will ask them one after another. For example on OSX, Git comes with a helper to use the OSX keychain which is configured at system level. This does not conflict with setting up DataLad's credential helper.
+
+The other way around, DataLad can ask Git for credentials it knows by other means (other git credential helpers).
+For that a datalad provider config needs to be set up in similar fashion::
+
+  [provider:data_example_provider]
+    url_re = http.*://.*data\.example\.com
+    authentication_type = http_basic_auth
```
aqw, line 26: Because of this, I /really/ think we should not be providing an example that matches `http`. People tend to copy and paste without groking.

docs/source/credentials.rst
```diff
@@ -0,0 +1,39 @@
...
+    credential = data_example_cred
+  [credential:data_example_cred]
+    type = git
```
aqw, line 29: Just to make sure I understand this right, this is where DataLad says "go ask git and see if it has a credential helper for this situation".

If that is the case, I'm pretty convinced that both configurations are needed in order for this to be functional (please correct me if I'm wrong; second time reading through).

It also makes me wonder what kind of weird bugs/corner-cases there could be from having two different regex patterns to match URLs (git's vs Python's). They won't always match exactly the same URLs. Cascading could get very, very weird.

aqw (changes requested):
I have only read docs/source/credentials.rst. A few key points that are unclear to me:

- whether both configurations are needed
  - if so: how they work without being circular (they look as if they are)
  - if not: why would one use one vs the other (git looks simple; datalad looks complicated)
- clarification of how http->https redirections are handled/matched
- if both configuration files are needed, the problem space should be explored of two different regex schemes in use

Some of my comments assume that both configs are needed. Other comments assume only one can be in use at a time. My apologies if that causes any confusion for others. I am thoroughly confused on that point, and have changed my mind repeatedly.

docs/source/credentials.rst
```diff
@@ -0,0 +1,39 @@
+Credentials
+***********
+
+Integration with Git
+====================
+
+Git and DataLad can use each other's credential system.
+In order to allow Git to query credentials from DataLad, the git credential helper delivered with DataLad needs to be configured.
+That is, a section like this needs to be part of one's git config file::
```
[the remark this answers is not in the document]
bpoldrack: Yes, that's what I mean.

docs/source/credentials.rst
```diff
@@ -0,0 +1,39 @@
...
+
+  [credential "https://*.data.example.com"]
+    helper = "datalad"
+
+Note:
+
+- This most likely only makes sense at the user or system level (options `--global`|`--system` with `git config`), since cloning of a repository needs the credentials before there is a local repository.
+- The name of that section is a URL matching expression - see `man gitcredentials`.
+- The URL matching does NOT include the scheme! Hence, if you need to match `http` as well as `https`, you need two such entries.
```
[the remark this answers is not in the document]
bpoldrack: > The necessity of this note is debatable, as it should be covered in man gitcredentials.

I'd argue: Evidently it's sufficiently unclear in that man page, that *you* didn't get it the first time (remember the initial git-credential config you had on juseless?), which is exactly why I put it here. ;-)

> If you /do/ feel it is noteworthy enough, then I propose warning people that they should avoid the http scheme, as credentials will be sent unencrypted (bad™).

I don't think this belongs here. A datalad user does not per se have a choice what URL(s) need(s) to be used. This is about infrastructure, setting up redirects, etc. - not a matter of how to use datalad.

docs/source/credentials.rst
```diff
@@ -0,0 +1,39 @@
...
+- Multiple different git credential helpers can be configured - Git will ask them one after another. For example on OSX, Git comes with a helper to use the OSX keychain which is configured at system level. This does not conflict with setting up DataLad's credential helper.
```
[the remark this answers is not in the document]
bpoldrack: > "Git will ask them one after another": does this mean all will always be asked? Or will it query them starting with the first, and attempt them sequentially until one matches and succeeds?

This. Which relates to the suspicion of cycles. Will change wording accordingly.

> OSX -> macOS (they've dropped the OSX for years now, as they are now on version 11.x)

Well. But the helper I am referring to is still called `git-credential-osxkeychain`. Will try to reword whie aowing to identify the one I am talking about.

> "This does not conflict with setting up DataLad's credential helper." I don't quite know what this means (and confuses me more than it helps me). How would the keychain "conflict"? It's just yet another credential helper that can be invoked by DataLad/git, yes?

Yes. I thought it may be a natural question: "There already is a helper configured. Am I supposed to overwrite that config?", because in a lot of cases, a git config you set up, will take precedence over the system config. Not in this case: It's an addition.

docs/source/credentials.rst
```diff
@@ -0,0 +1,39 @@
...
+
+The other way around, DataLad can ask Git for credentials it knows by other means (other git credential helpers).
+For that a datalad provider config needs to be set up in similar fashion::
+
+  [provider:data_example_provider]
+    url_re = http.*://.*data\.example\.com
```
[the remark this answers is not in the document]
bpoldrack: > What happens if the scheme is https only, but an http connection is initiated and the server redirects to https. Will it match this regex and provide credentials only for the https attempt (as git [correctly] does)?

Not currently. DataLad matches the given URL, not the one it figures in the preocess (when being redirected). But this is outside scope of this PR and a general issue of the cred/downloader system in DataLad.

docs/source/credentials.rst
```diff
@@ -0,0 +1,39 @@
...
+    authentication_type = http_basic_auth
+    credential = data_example_cred
+  [credential:data_example_cred]
+    type = git
+
+Such a config lives in a dedicated file named after the provider name (:file:`data_example_provider.cfg` in this case).
+Valid locations for these files are listed in :ref:`chap_design_credentials`.
+In opposition to Git's approach, `url_re` is a regular expression that matches the entire URL including the scheme.
+The name of the provider section (`data_example_provider`)needs to match the file name.
```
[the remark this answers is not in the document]
bpoldrack: There's no doc style guide.

docs/source/credentials.rst
```diff
@@ -0,0 +1,39 @@
...
+The name of the credential section (`data_example_cred`) needs to match the `credential` entry in the provider section.
+
+DataLad will create a provider configuration interactively, whenever it first encounters an URL that requires authentication and no credentials can be matched yet.
+This behavior extends to the git credential helper and can therefore even be triggered by a `git clone`.
+However, `git-credential-datalad` will not do this, if the config variable `datalad.credentials.githelper.noninteractive` is set.
```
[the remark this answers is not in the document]
bpoldrack: > Why?

Hm. That's actually a bit tricky as a serious answer (as opposed to "To not be asked!") requires to point out, how this interactivity is different from within such a helper than from within DataLad as a toplevel process (in which case it's auto-determined based on whether we are running in an interactive session).
Don't think that's helpful here. Will just mention that it's intended use is DataLad's CI tests.

docs/source/credentials.rst
```diff
@@ -0,0 +1,39 @@
...
+  [credential:data_example_cred]
+    type = git
+
+Such a config lives in a dedicated file named after the provider name (:file:`data_example_provider.cfg` in this case).
+Valid locations for these files are listed in :ref:`chap_design_credentials`.
+In opposition to Git's approach, `url_re` is a regular expression that matches the entire URL including the scheme.
+The name of the provider section (`data_example_provider`)needs to match the file name.
+The name of the credential section (`data_example_cred`) needs to match the `credential` entry in the provider section.
+
+DataLad will create a provider configuration interactively, whenever it first encounters an URL that requires authentication and no credentials can be matched yet.
```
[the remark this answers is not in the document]
bpoldrack: > Do you mean that it will prompt the user?

Yes.

> Can that behavior be suppressed?

By providing it beforehand.

docs/source/credentials.rst
```diff
@@ -0,0 +1,39 @@
...
+
+Integration with Git
+====================
+
+Git and DataLad can use each other's credential system.
+In order to allow Git to query credentials from DataLad, the git credential helper delivered with DataLad needs to be configured.
+That is, a section like this needs to be part of one's git config file::
+
+  [credential "https://*.data.example.com"]
+    helper = "datalad"
```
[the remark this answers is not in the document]
bpoldrack: > This surprises me. The command datalad is enough to act as a git credential helper? I don't understand how that would work.

See `man gitcredentials` (value string does not begin with an absolute path). It's not a command. It results in calling `git-credential-datalad`. 

> Also, this example assumes that datalad is always in the path, which might not be the case with scripts, cron, etc.

At least the helper (technically not datalad itself) has to be in the path, yes. Otherwise one would need to give the absolute path to that helper. Not good for an example, as it may depend on what system you installed DataLad (and how). So, I wouldn't suggest any absolute path here.

docs/source/credentials.rst
bpoldrack: Depends on what you mean by "needed". Both directions work independently and can stand on their own. What's needed depends on what you want. If you want to seemlessly use the credentials provided by both systems, no matter whether that's a `git clone` or a `datalad get`, then yes - you likely want both. But probably not both for the same URL.

In case of our datastores, you areadly provide credentials via a git-credential helper. So, all that's needed to make it complete, is telling DataLad to ask Git (when downloading content) when it encounters such URLs.

Will try to make that more clear.

docs/source/credentials.rst
bpoldrack: I am not sure, where you get the idea from, that both is needed and particularly for the same URL.

However, there probably are cases where a circular setup looks like what one wants. What I can imagine would be a not URL specific config `credential.helper="datalad"` for Git and then encounter a URL for which DataLad is configured to ask Git. In any case: Only makes sense if there's another git-credential-helper that can provide something. That one would need to be configured first, so Git doesn't ask DataLad when it has something else to ask, that only talks to Git.

I will have another look into consequences of circular setups.

bpoldrack:
Thx, @aqw !

I will try to reword some things and do some testing with circular setups

mih:
Any ETA for the doc update -- it would be great to finally get this to land.

bpoldrack:
@mih, (@aqw )
Just pushed a change preventing the possibility of setting up a circular config.

Doc update tomorrow latest. Just wanted to see CI results ASAP.

mih:
Thanks for adding a changelog. I took the respective issue tag off, to communicate that properly.

docs/source/credentials.rst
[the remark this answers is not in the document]
aqw: Cool. Then feel free to copy my rewording and use it instead.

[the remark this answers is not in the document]
aqw: Ahh; I /just/ now found the view where GitHub shows me your changes /and/ our comments. It seems like you already lifted the text. LGTM.

docs/source/credentials.rst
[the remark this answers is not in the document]
aqw: > See man gitcredentials (value string does not begin with an absolute path). It's not a command. It results in calling git-credential-datalad.

Ahh yeah, I remember reading that and disliking it. Now I am reminded why. ;-)

docs/source/design/credentials.rst
```diff
@@ -28,11 +33,60 @@ These locations are tried in the following order, and the first successful looku
...
+There's no way at the moment to store an existing :class:`Provider` object directly.
+
+Integration with Git
+====================
+
+In addition, there's a special case for interfacing `git-credential`: A dedicated :class:`GitCredential` class is used to talk to Git's ``git-credential`` command instead of the keyring wrapper.
+This class has identical fields to the :class:`UserPassword` class and thus can be used by the same authenticators.
+Since Git's way to deal with credentials doesn't involve labels but only matching URLs, it is - in some sense - the equivalent of datalad's provider layer.
+However, providers don't talk to a backend, credentials do.
+Hence, a more seemless integration requires some changes in the design of datalad's credential system as a whole.
```
aqw, line 80: seemless -> seamless

I would also recommend saying that it /would/ require changed to the design. Make it clear that it's hypothetical.

loj:
I tested this out again on our infrastructure with the following setup:

I setup my user level `.gitconfig` with
```
[credential "https://*.ds.inm7.de"]
    helper = "/usr/local/bin/inm7_datastore_pw"
[datalad "credentials.githelper"]
    noninteractive = true
```

And a user level provider config:
```
❱ cat ~/.config/datalad/providers/inm7_ds.cfg
[provider:inm7_ds]
  url_re = https://.*ds\.inm7\.de
  credential = inm7_ds

[credential:inm7_ds]
  type = git
```

But it fails to pick up the appropriate authentication post redirection. 
```
❱ datalad clone "ria+http://studyforrest-test.ds.inm7.de#3a8648b3-7df8-413f-8efb-4d39040ac174" studyforrest_test
[WARNING] redirecting to https://studyforrest-test.ds.inm7.de/3a8/648b3-7df8-413f-8efb-4d39040ac174/                                                                                                                
[INFO   ] [INFO] Fetching 'http://studyforrest.ds.inm7.de/ria-layout-version' 
[INFO   ] [INFO] Fetching 'http://studyforrest.ds.inm7.de/3a8/648b3-7df8-413f-8efb-4d39040ac174/ria-layout-version' 
[INFO   ] Fetching 'http://studyforrest-test.ds.inm7.de/3a8/648b3-7df8-413f-8efb-4d39040ac174/config' 
Access to http://studyforrest-test.ds.inm7.de/3a8/648b3-7df8-413f-8efb-4d39040ac174/config has failed.
Would you like to setup a new provider configuration to access url? (choices: [yes], no): no
```

<details><summary>With this setup, it works:</summary>

```
[credential "http://*.ds.inm7.de"]
    helper = "/usr/local/bin/inm7_datastore_pw"
[credential "https://*.ds.inm7.de"]
    helper = "/usr/local/bin/inm7_datastore_pw"
[datalad "credentials.githelper"]
    noninteractive = true
```
```
[provider:inm7_ds]
  url_re = http.*://.*ds\.inm7\.de
  credential = inm7_ds

[credential:inm7_ds]
  type = git
```

```
❱ datalad clone "ria+http://studyforrest-test.ds.inm7.de#3a8648b3-7df8-413f-8efb-4d39040ac174" studyforrest_test
[WARNING] redirecting to https://studyforrest-test.ds.inm7.de/3a8/648b3-7df8-413f-8efb-4d39040ac174/                                                                                                                
[INFO   ] [INFO] Fetching 'http://studyforrest.ds.inm7.de/ria-layout-version' 
[INFO   ] [INFO] http session: Authenticating into session for http://studyforrest.ds.inm7.de/ria-layout-version 
[INFO   ] [INFO] Fetching 'http://studyforrest.ds.inm7.de/3a8/648b3-7df8-413f-8efb-4d39040ac174/ria-layout-version' 
[INFO   ] Fetching 'http://studyforrest-test.ds.inm7.de/3a8/648b3-7df8-413f-8efb-4d39040ac174/config' 
[INFO   ] http session: Authenticating into session for http://studyforrest-test.ds.inm7.de/3a8/648b3-7df8-413f-8efb-4d39040ac174/config 
[INFO   ] Reconfigured inm7-storage for ria+http://studyforrest-test.ds.inm7.de 
[INFO   ] Fetching 'http://studyforrest-test.ds.inm7.de/3a8/648b3-7df8-413f-8efb-4d39040ac174/config' 
[INFO   ] Configure additional publication dependency on "inm7-storage" 
install(ok): /home/loj/tmp/studyforrest_test (dataset)
action summary:
  configure-sibling (ok: 1)
  install (ok: 1)
```
</details>

docs/source/credentials.rst
```diff
@@ -0,0 +1,50 @@
...
+- The name of that section is a URL matching expression - see `man gitcredentials`.
+- The URL matching does NOT include the scheme! Hence, if you need to match `http` as well as `https`, you need two such entries.
+- Multiple git credential helpers can be configured - Git will ask them one after another until it got a username and a password for the URL in question. For example on macOS, Git comes with a helper to use the system's keychain and Git is configured system-wide to query `git-credential-osxkeychain`. This does not conflict with setting up DataLad's credential helper.
+- The example configuration requires `git-credential-datalad` to be in the path in order for Git to find it. Alternatively, the value of the `helper` entry needs to be the absolute path of `git-credential-datalad`.
+- In order to make Git always consider DataLad as a credential source, one can simply not specify any URL pattern (so it's `[credential]` instead of `[credential "SOME-PATTERN"]`)
+
+The other way around, DataLad can ask Git for credentials (which it will acquire via other git credential helpers).
```
aqw, line 27: I recommend making this its own section with a heading.

docs/source/credentials.rst
aqw: Ok. I think I see how this works now. The intro could benefit by making one thing explicit: these methods are useful if either DataLad or Git are already configured to discover credentials to access a remote resource. Then the following steps explain how to direct one to the other to get that info.

Obviously that is not exactly the smoothest way of saying it.

I recommend having the intro explain the overall concept. Then one section for each method.

aqw (approved):

aqw:
@bpoldrack I'm confused whether the "Review" is still open or not according to GitHub. And there's no comment threading, so I'm just commenting here. :-)

I reviewed your changes, and most of my concerns are addressed. I feel like I better understand how it is designed.

I /do/ think that overall it could benefit from the use of sections/headers to organize the content, and a pass to reduce sentence complexity of the use of "note" statements.

None of this is a blocker IMO, and instead is nice to have.

The only blocker I see is the [redirection issue](https://github.com/datalad/datalad/pull/5796#issuecomment-1028072189) that @loj found.

bpoldrack:
Thanks, @aqw and @loj.

The redirection issue is documented, though and shouldn't block this PR, since it has nothing to do with this PR.

bpoldrack:
FTR: Canceled CI builds, since I only force-pushed doc updates

PR #5797, opened 2021-07-18
(Re-)define public python interface

bpoldrack:
This is not done. At least should come with some more adaptions within `Dataset` to the basic idea of this PR.
The "flavor" of `call_git/annex` methods in `Dataset` may need to be slightly different  as a high-level representation (error handling, may be documenting that this is not a "command" and doesn't come with the general options or add some wrapper code that does (part of) that.

For discussion please see #5798

- [ ] include proper documentation
- [ ] accompanying PR to extensions-template linking the relevant doc

### DIFF
--- a/datalad/distribution/dataset.py
+++ b/datalad/distribution/dataset.py
@@ -22,12 +22,8 @@ import wrapt
 
 from datalad import cfg
 from datalad.config import ConfigManager
-from datalad.core.local.repo import repo_from_path
 from datalad.support.annexrepo import AnnexRepo
 from datalad.support.constraints import Constraint
-# DueCredit
-from datalad.support.due import due
-from datalad.support.due_utils import duecredit_dataset
 from datalad.support.exceptions import (
     NoDatasetFound,
 )
@@ -53,7 +49,8 @@ from datalad.utils import (
     ensure_list,
     quote_cmdlinearg,
 )
-
+from datalad.interface.vcs import VCS
+from datalad.dochelpers import borrowdoc
 
 lgr = logging.getLogger('datalad.dataset')
 lgr.log(5, "Importing dataset")
@@ -143,10 +140,10 @@ class Dataset(object, metaclass=PathBasedFlyweight):
         if isinstance(path, ut.PurePath):
             path = str(path)
         self._path = path
-        self._repo = None
         self._id = None
         self._cfg = None
         self._cfg_bound = None
+        self._vcs = VCS(self.pathobj)
 
     @property
     def pathobj(self):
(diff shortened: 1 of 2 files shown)


datalad/distribution/dataset.py
```diff
@@ -453,6 +387,52 @@ def subds_contains_path(ds, path):
...
+
+    @borrowdoc(AnnexRepo)
+    def call_annex_items_(self, args, files=None, sep=None):
+
+        return self._vcs.call_annex_items_(args, files=None, sep=None)
+
+    @borrowdoc(AnnexRepo)
+    def call_annex_oneline(self, args, files=None):
+
+        return self._vcs.call_annex_oneline(args, files=None)
```
mih, line 434: All `call_annex_*()` methods would fail if the underlying `VCS` instance points to a `GitRepo`.

datalad/interface/vcs.py
```diff
@@ -0,0 +1,193 @@
...
+
+    @borrowdoc(AnnexRepo)
+    def call_annex_items_(self, args, files=None, sep=None):
+
+        return self._repository.call_annex_items_(args, files=None, sep=None)
+
+    @borrowdoc(AnnexRepo)
+    def call_annex_oneline(self, args, files=None):
+
+        return self._repository.call_annex_oneline(args, files=None)
```
mih, line 193: All `call_annex_*()` methods would fail if the underlying `_repository` instance points to a `GitRepo`.

mih (changes requested):
I am not convinced that a single VCS class works nicely.

I am feeling pretty strong about neither fiddling with code under `distribution/` nor `interface/` and certainly not introducing new stuff in there.

I feel that the `VCS` class and some other aspects of this PR overlap with the starting point that was made in `datalad/dataset/` (intentionally not under `core/` at the time).

Rather than just starting from scratch it would be useful to identify why this existing code cannot be used. This code also identifies the within `*Repo` dependencies to make Git calls via class instances. I consider this a useful approach to achieve an incremental transition.

datalad/distribution/dataset.py
bpoldrack: Yes, this draft so far is meant to illustrate what I wrote in the issue in principle, not be a full implementation yet. I'd like to first get some agreement on the goal. However - yes: At the `Dataset` level, those would have to be properly guarded.

bpoldrack:
@mih 

I was about to follow your suggestion in #5798 and turn the issue into a "discussion-PR". That may be better than here, since I'd like to get some view on the principles/goals I layed out. So far I don't even know whether and to what extend there's agreement on what's the problem to be solved.
Hence, will do that and copy your comments over.

bpoldrack:
Closing in favor of #5906

Issue #5759, opened 2021-06-22
two fresh failures for maint/release in datalad/git-annex


yarikoptic:
see e.g. https://github.com/datalad/git-annex/runs/2881114136?check_suite_focus=true

```
ERROR: datalad.distribution.tests.test_create_sibling.test_target_ssh_inherit('backup', False, False)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.7.10/x64/lib/python3.7/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/opt/hostedtoolcache/Python/3.7.10/x64/lib/python3.7/site-packages/datalad/tests/utils.py", line 764, in _wrap_with_tempfile
    return t(*(arg + (filename,)), **kw)
  File "/opt/hostedtoolcache/Python/3.7.10/x64/lib/python3.7/site-packages/datalad/tests/utils.py", line 764, in _wrap_with_tempfile
    return t(*(arg + (filename,)), **kw)
  File "/opt/hostedtoolcache/Python/3.7.10/x64/lib/python3.7/site-packages/datalad/distribution/tests/test_create_sibling.py", line 624, in _test_target_ssh_inherit
    ds.publish(to=remote)
  File "/opt/hostedtoolcache/Python/3.7.10/x64/lib/python3.7/site-packages/datalad/distribution/dataset.py", line 503, in apply_func
    return f(**kwargs)
  File "/opt/hostedtoolcache/Python/3.7.10/x64/lib/python3.7/site-packages/datalad/interface/utils.py", line 481, in eval_func
    return return_func(generator_func)(*args, **kwargs)
  File "/opt/hostedtoolcache/Python/3.7.10/x64/lib/python3.7/site-packages/datalad/interface/utils.py", line 474, in return_func
    results = list(results)
  File "/opt/hostedtoolcache/Python/3.7.10/x64/lib/python3.7/site-packages/datalad/interface/utils.py", line 461, in generator_func
    msg="Command did not complete successfully")
datalad.support.exceptions.IncompleteResultsError: Command did not complete successfully. 1 failed:
[{'action': 'publish',
  'message': ('failed to push to %s: %s;%s',
              'magical',
              'refs/heads/master -> refs/heads/master [rejected] '
              '(non-fast-forward)',
              " pushed: ['43c9324..5600e27']"),
  'path': '/tmp/datalad_temp__test_target_ssh_inheritdk_6i6mv',
[...]

PR #5799, opened 2021-07-19
ENH: Supply stdin to process without a tmp file

mih:
So far, a temporary file was necessary to pass input to a process's `stdin`
in `WitlessRunner` (need for `.fileno` in `subprocess.Popen`). This
enhancement adopts the subprocess `input=` implementation, and allows
for feeding input from a `str|bytes`, by writing its content to a stdin
pipe.

This enables a simplifications of the `push` implementation and reduces
the overall technical complexity of passing `stdin`.

As `WitlessRunner` presently operates the pipes in binary mode, input
must be given as bytes. However, the underlying implementation in
`run_command()` could handle strings as well, should `Popen` be
parametrized accordingly using custom arguments.

This PR include two cleanup commits that are essentially unrelated.

### DIFF
--- a/datalad/nonasyncrunner.py
+++ b/datalad/nonasyncrunner.py
@@ -106,32 +106,20 @@ class _StdinWriterThread(threading.Thread):
         self.stdin_fileno = stdin_fileno
         self.queue = q
         self.command = command
-        self.quit = False
 
     def __str__(self):
-        return f"WriterThread({self.stdin_data}, {self.process}, {self.stdin_fileno}, {self.command})"
-
-    def request_exit(self):
-        """
-        Request the thread to exit. This is not guaranteed to
-        have any effect, because the thread might be waiting in
-        `process._stdin_write(...)`.
-        """
-        self.quit = True
+        return (
+            f"WriterThread(stdin_data[0 ... {len(self.stdin_data) - 1}], "
+            f"{self.process}, {self.stdin_fileno}, {self.command})")
 
     def run(self):
         logger.debug("%s started", self)
 
-        try:
-            # (ab)use internal helper that takes care of a bunch of corner cases
-            # and closes stdin at the end
-            self.process._stdin_write(self.stdin_data)
-        except BrokenPipeError:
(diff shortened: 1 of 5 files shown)


mih:
Test failures are unrelated #5759

datalad/nonasyncrunner.py
```diff
@@ -154,6 +165,11 @@ def run_command(cmd: Union[str, List],
         process_stderr_fileno: STDERR_FILENO
     }
 
+    if input is not None:
+        # (ab)use internal helper that takes care of a bunch of corner cases
+        # and closes stdin at the end
+        process._stdin_write(input)
+
```
christian-monch, line 172: I think we should probably do that after the output reader threads are started, otherwise we might end up with a deadlock: subprocess blocked on stdout/stderr, and, as a result of that, a blocked main thread, if input is large and the input-fifo runs full. That might only happen with inputs larger than 4k, but we do not restrict the size, as far as I know.

Still, it might also be good to do the writing in an extra thread, since the main thread should read the output queues to ensure that the sub-process does not block on its output.

I would propose to introduce a writer thread (or re-introduce it, it existed in a previous non-async-commit) to ensure that we do not end up with blocked subprocesses or deadlocks.

christian-monch (changes requested):

datalad/nonasyncrunner.py
mih: I tried that (without a dedicated thread), but it did not work. If you see how this could be done, please take over. Thx!

PS: I am already using this feature in the implementation of the git-credential support.

christian-monch: Will give it a try, and get back to you.

mih:
Thx @christian-monch for the improvements! They look great to me! If you are also satisfied with your work, please change the review status ;-)

christian-monch (approved):

christian-monch:
> Thx @christian-monch for the improvements! They look great to me! If you are also satisfied with your work, please change the review status ;-)

Ops. Done!

Issue #5729, opened 2021-06-10
Debian package dependency issue with python3-github


mih:
`python3-github` is recommended by `python3-datalad` and is installed on this Debian buster machine:

```
% apt-cache policy python3-github
python3-github:
  Installed: 1.43.7-1~nd100+1
  Candidate: 1.43.7-1~nd100+1
  Version table:
 *** 1.43.7-1~nd100+1 500
        500 http://neurodebian.mirror.htc.inm7.de/debian buster/main amd64 Packages
        100 /var/lib/dpkg/status
     1.40-1 500
        500 http://debian.mirror.htc.inm7.de/debian buster/main amd64 Packages
```

but it is not functional

```
mih@juseless ~ % python3 -c "import github"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/usr/lib/python3/dist-packages/github/__init__.py", line 42, in <module>
    from github.MainClass import Github, GithubIntegration
  File "/usr/lib/python3/dist-packages/github/MainClass.py", line 74, in <module>
    from . import RateLimit
  File "/usr/lib/python3/dist-packages/github/RateLimit.py", line 33, in <module>
    from deprecated import deprecated
ModuleNotFoundError: No module named 'deprecated'
```

and this leads to

```
% datalad create-sibling-github some
[ERROR  ] PyGitHub is missing. GitHub-related functionality is unavailable without this package [github_.py:<module>:35] (MissingExternalDependency) 
```

which is an invalid interpretation of the situation.

Installing `python3-deprecated` manually resolves the issue.

I have not yet filed a bug against the `python3-github` Debian package, because this is a NeuroDebian backport, and may just be our bug.

yarikoptic:
I loosely recall this: IIRC dependency is autopopulated by some dh_python3 but it is not autodetected properly on nd100, so there is no issue to file against package in debian, since there it has it all working fine, and I guess we are just doomed to add it to our Recommends along with pygithub or carry patched version of packaging within neurodebian. but then -- python3-github is only recommends, so optional and thus we would not be able to guarantee that `apt-get install python3-github` would just fix it.

An alternative is to make python3-deprecated into Depends... well -- I have tried to make it used https://github.com/datalad/datalad/pull/5343 ;-)

mih:
Thanks for the info. With that in mind, we should at least update the error message to acknowledge these possible alternative causes for "missing.

PR #5803, opened 2021-07-21
Adjust wording of MissingExternalDependency error

mih:
The new phrasing should better cover the case of an existing installation
that is nevertheless broken, and fails an import test.

Fixes datalad/datalad#5729


adswa (approved):
This looks better than the previous version to me.

mih:
Thx @adswa !

PR #5805, opened 2021-07-22
UX: Check argument validity of --as-common-datasrc name

### LINKED ISSUE #5621: `siblings` should check argument validity of `--as-common-datasrc`
```
% datalad siblings add -d . \
  --name gin \
  --pushurl git@gin.g-node.org:/studyforrest/visual-areas.git \
  --url https://gin.g-node.org/studyforrest/visual-areas.git \
  --as-common-datasrc gin

[INFO   ] Could not enable annex remote gin. This is expected if gin is a pure Git remote, or happens if it is not accessible. 
CommandError: 'git -c diff.ignoreSubmodules=none annex initremote gin type=git location=https://gin.g-node.org/studyforrest/visual-areas.git autoenable=true -c annex.dotfiles=true -c annex.retry=3' failed with exitcode 1 under /data/project/studyforrest/superds/derivative/visual_areas
git-annex: There is already a remote named "gin"
```

adswa:
This PR adds a ValueError that is raised as a validity check when a user supplies identical names for a remote sibling name and the autoenabled as_common_datasrc special remote, as done in #5621.

Fixes #5621

### DIFF
--- a/datalad/distribution/siblings.py
+++ b/datalad/distribution/siblings.py
@@ -379,6 +379,9 @@ def _add_remote(
             message=("sibling is already known: %s, use `configure` instead?", name),
             **res_kwargs)
         return
+    if as_common_datasrc == name:
+        raise ValueError('Sibling name ({}) and common data source name ({}) '
+                         'can not be identical.'.format(name, as_common_datasrc))
     if isinstance(RI(url), PathRI):
         # make sure any path URL is stored in POSIX conventions for consistency
         # with git's behavior (e.g. origin configured by clone)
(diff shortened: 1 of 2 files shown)


mih (approved):
I was torn re exception vs error. The existing code uses both (I could not detect a clear pattern). Given that the name clash can only occur based on input arguments (rather than dataset properties), I think it is preferable to use an exception -- like you did. I also agree with the choice of exception type (`ValueError` rather than `InsufficientArgumentsError` which seems prevalent in this code, but often doesn't fit IMHO).

yarikoptic:
logic sounds logical, and I agree with @mih that `ValueError` as an exception for a misspecified entry (as for no entry at all) is the correct one .  Let's proceed. Thank you @adswa !

yarikoptic:
but may be this and such fixes should go against `maint` still to possibly benefit "more"

Issue #5807, opened 2021-07-22
`CommandError` info hidden (likely wide-spread issue)


mih:
Working on #3127, I tried what would happen when calling `push` on a special remote with `exporttree=yes`:

```
% datalad push --to expdir
Transfer data to 'expdir':  50%|█████████▌         | 2.00/4.00 [00:00<00:00, 7.06k Steps/sCommandError: 'git -c diff.ignoreSubmodules=none annex copy --batch -z --to expdir --fast --json --json-error-messages --json-progress -c annex.dotfiles=true -c annex.retry=3' failed with exitcode 1 under /tmp/testds
git-annex: copy: 1 failed
```

That is suboptimal error reporting.

Here is what git-annex is actually reporting:

```
% git annex copy --to expdir
copy 123 (to expdir...) 
  remote is configured with exporttree=yes; use `git-annex export` to store content on it
failed
git-annex: copy: 1 failed
```

But this seems to have little to do with `push`. There is no specific error handling. This is a plain `CommandError`:

```
> /home/mih/hacking/datalad/git/datalad/cmd.py(168)run()
-> raise CommandError(
(Pdb) l
163             if results.get('code', True) not in [0, None]:
164                 # the runner has a better idea, doc string warns Protocol
165                 # implementations not to return these
166                 results.pop('cmd', None)
167                 results.pop('cwd', None)
168  ->             raise CommandError(
169                     # whatever the results were, we carry them forward
170                     cmd=cmd,
171                     cwd=self.cwd,
172                     **results,
173                 )
```

and there is plenty of information given to it:

```
(Pdb) pprint(results)
{'code': 1,
 'stderr': 'git-annex: copy: 1 failed\n',
 'stdout': '',
 'stdout_json': [{'command': 'copy',
                  'error-messages': ['  remote is configured with '
                                     'exporttree=yes; use `git-annex export` '
                                     'to store content on it',
                                     '  remote is configured with '
                                     'exporttree=yes; use `git-annex export` '
                                     'to store content on it',
                                     '  remote is configured with '
                                     'exporttree=yes; use `git-annex export` '
                                     'to store content on it',
                                     '  This could have failed because --fast '
                                     'is enabled.'],
                  'file': '123',
                  'input': ['123'],
                  'key': 'MD5E-s4--ba1f2511fc30423bdbb183fe33f3dd0f',
                  'note': 'to expdir...',
                  'success': False}]}
```

[...]

PR #5809, opened 2021-07-22
BF: Report info from annex JSON error message in CommandError

mih:
When a `CommandError` bubbles up to the main entrypoint it is rendered
differently than anywhere else (calls `to_str()` with
`include_output=False`. This is done to be able to write the
`stdout|err` output of a failed command to the proper channels. However,
this also caused all information extraction from JSON records to be
disabled.

This change, makes sure that JSON records are processed even in this
case. It also RFs the exception rendering code to make it less
repr()-like.

From:

```
CommandError: 'git -c diff.ignoreSubmodules=none annex copy --batch -z --to expdir --fast --json --json-error-messages --json-progress -c annex.dotfiles=true' failed with exitcode 1 under /tmp/testds
git-annex: copy: 2 failed
```

To:

```
CommandError: 'git -c diff.ignoreSubmodules=none annex copy --batch -z --to expdir --fast --json --json-error-messages --json-progress -c annex.dotfiles=true' failed with exitcode 1 under /tmp/testds [info keys: stdout_json]
> to expdir...
  remote is configured with exporttree=yes; use `git-annex export` to store content on it
  This could have failed because --fast is enabled. [2 times]
git-annex: copy: 2 failed
```

Fixes datalad/datalad#5807

### DIFF
--- a/datalad/cmdline/tests/test_main.py
+++ b/datalad/cmdline/tests/test_main.py
@@ -29,7 +29,10 @@ from datalad.ui.utils import (
     get_console_width,
     get_terminal_size,
 )
-from datalad.api import create
+from datalad.api import (
+    create,
+    Dataset,
+)
 from datalad.utils import (
     chpwd,
     Path,
@@ -47,6 +50,7 @@ from datalad.tests.utils import (
     slow,
     SkipTest,
 )
+from datalad.support.exceptions import CommandError
 
 
 def run_main(args, exit_code=0, expect_stderr=False):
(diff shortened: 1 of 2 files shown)


mih:
The new test in a single travis run is failing (does not see the desired exception output). ATM I have no idea why this would only happen in this run, and nowhere else. I restarted it, but that is mostly a desperate move.

yarikoptic:
> Edit: Yeah, deterministic failure. So it seems that the git-annex version 7.20190819 that is specifically tested there, does not provide this kind of error. So I guess I should add a version check to the test.

Thanks for nailing it down! great that we have that run with that old annex. Note that in `maint` we have
```
$> grep MIN datalad/support/annexrepo.py
    GIT_ANNEX_MIN_VERSION = '7.20190503
```
and in master
```
$> grep MIN datalad/support/annexrepo.py
    GIT_ANNEX_MIN_VERSION = '8.20200309'
```
so if you could check that it passes on `8.20200309` - then we should strip that check away whenever merging into master

mih:
OK, all green. Good to go!

Issue #5810, opened 2021-07-22
test_runner_stdin_no_capture errors out while testing master on datalad/git-annex (github actions)


yarikoptic:
e.g. https://github.com/datalad/git-annex/runs/3130248890?check_suite_focus=true recently added in 0.14.4-477-gcdde06b73
seems to be failing only on ubuntu

```
======================================================================
ERROR: datalad.tests.test_witless_runner.test_runner_stdin_no_capture
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.7.11/x64/lib/python3.7/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/opt/hostedtoolcache/Python/3.7.11/x64/lib/python3.7/site-packages/nose/tools/nontrivial.py", line 97, in newfunc
    result = func(*arg, **kw)
  File "/opt/hostedtoolcache/Python/3.7.11/x64/lib/python3.7/site-packages/datalad/tests/test_witless_runner.py", line 162, in test_runner_stdin_no_capture
    protocol=None
  File "/opt/hostedtoolcache/Python/3.7.11/x64/lib/python3.7/site-packages/datalad/cmd.py", line 172, in run
    **results,
datalad.support.exceptions.CommandError: CommandError: '/opt/hostedtoolcache/Python/3.7.11/x64/bin/python -c 'import sys; print(sys.stdin.read())'' failed with exitcode 1
```

and consistently so

```shell
(git)smaug:/mnt/btrfs/datasets/datalad/ci/git-annex[master]builds/2021
$> cd 07
cron-20210701/  cron-20210703/  cron-20210705/  cron-20210707/  cron-20210709/  cron-20210711/  cron-20210713/  cron-20210715/  cron-20210717/  cron-20210719/  cron-20210721/  pr-70/  pr-UNK/
cron-20210702/  cron-20210704/  cron-20210706/  cron-20210708/  cron-20210710/  cron-20210712/  cron-20210714/  cron-20210716/  cron-20210718/  cron-20210720/  cron-20210722/  pr-72/

$> git grep test_runner_stdin_no_capture
cron-20210721/build-macos.yaml-360-69466103-failed/1_test-datalad (master).txt:2021-07-21T05:03:24.6620130Z datalad.tests.test_witless_runner.test_runner_stdin_no_capture ... ok
[...]

PR #5813, opened 2021-07-23
Fix log flooding during witless-runner tests

christian-monch:
Stops printing of 240000 characters during witless-runner tests.

### DIFF
--- a/datalad/tests/test_witless_runner.py
+++ b/datalad/tests/test_witless_runner.py
@@ -157,7 +157,7 @@ def test_runner_stdin_no_capture():
     # Ensure that stdin writing alone progresses
     runner = Runner()
     runner.run(
-        py2cmd('import sys; print(sys.stdin.read())'),
+        py2cmd('import sys; _ = sys.stdin.read()'),
         stdin=('ABCDEFGHIJKLMNOPQRSTUVWXYZ-' * 10000).encode('utf-8'),
         protocol=None
     )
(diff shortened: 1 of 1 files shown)


adswa:
The travis failure seems to be an unrelated connection issue

mih (approved):
Thx!

datalad/tests/test_witless_runner.py
```diff
@@ -157,7 +157,7 @@ def test_runner_stdin_no_capture():
     # Ensure that stdin writing alone progresses
     runner = Runner()
     runner.run(
-        py2cmd('import sys; print(sys.stdin.read())'),
+        py2cmd('import sys; _ = sys.stdin.read()'),
```
yarikoptic, line 160: now looking at the test, nothing in it really "Ensure that stdin writing alone progresses" does it?

I guess it could have tried to `swallow_outputs` and analyze them but not sure if that would interfere or not.  May be (if only for ensuring via visual inspection) it should print something at the end, e.g.

```suggestion
        py2cmd('import sys; print(sys.stdin.read()[-10:])'),
```
?

yarikoptic:
ref https://github.com/datalad/datalad/issues/5810 I assume.  Thank you @christian-monch !

datalad/tests/test_witless_runner.py
christian-monch: Thanks for the suggestion

christian-monch:
> ref #5810 I assume. Thank you @christian-monch !

Will have to look into that in more detail. Purpose of this PR was to stop the log-flooding

Issue #5812, opened 2021-07-23
Plenty of unused imports


adswa:
There are plenty of imports that are not used in the source code:

```
(handbook2) adina@muninn in ~/repos/datalad on git:master
❱ pyflakes datalad
datalad/cmd.py:21:1 '.cmd_protocols.KillOutput' imported but unused
datalad/cmd.py:21:1 '.cmd_protocols.StdErrCapture' imported but unused
datalad/cmd.py:21:1 '.cmd_protocols.StdOutCapture' imported but unused
datalad/cmd.py:21:1 '.cmd_protocols.StdOutErrCapture' imported but unused
datalad/cmd.py:21:1 '.cmd_protocols.WitlessProtocol' imported but unused
datalad/version.py:12:1 'sys' imported but unused
datalad/api.py:11:1 'from datalad.coreapi import *' used; unable to detect undefined names
datalad/api.py:11:1 'datalad.coreapi.*' imported but unused
datalad/coreapi.py:12:1 '.distribution.dataset.Dataset' imported but unused
datalad/utils.py:12:1 'hashlib' imported but unused
datalad/utils.py:45:1 'os.path.islink' imported but unused
datalad/utils.py:197:5 local variable 'environ' is assigned to but never used
datalad/utils.py:222:5 local variable 'e' is assigned to but never used
datalad/utils.py:390:1 'pathlib.PurePath' imported but unused
datalad/utils.py:390:1 'pathlib.PurePosixPath' imported but unused
datalad/utils.py:607:9 local variable 'exc' is assigned to but never used
datalad/__init__.py:41:1 local variable 'e' is assigned to but never used
datalad/__init__.py:67:1 '.version.__version__' imported but unused
datalad/dochelpers.py:21:1 'pathlib.Path' imported but unused
datalad/__main__.py:105:9 local variable 'aio' is assigned to but never used
datalad/ui/dialog.py:321:9 local variable 'backend' is assigned to but never used
datalad/ui/dialog.py:323:13 '.progressbars.tqdmProgressBar' imported but unused
datalad/ui/dialog.py:325:17 'tqdm.tqdm_notebook' imported but unused
datalad/ui/progressbars.py:356:13 local variable 'exc' is assigned to but never used
datalad/ui/tests/test_base.py:19:1 '...tests.utils.ok_' imported but unused
datalad/ui/tests/test_dialog.py:70:13 local variable 'choices_str' is assigned to but never used
[...]

PR #5814, opened 2021-07-23
MNT: Reduce unused imports and harmonize existing imports for a subset of the code base

adswa:
I went through the list of unused imports in #5812, and found plenty of unused imports in the code base. This PR removes them. I hope I didn't kill relevant imports - the tests will tell. 
Whenever I found inconsistent import formats, I tried to harmonize them into the current import statement standard, i.e.
```
from xy import (
   A,
   C, 
   Q,
   Z,
)
```

Within datalad/tests/utils.py  I stumbled about a mix between ``opj`` and ``op.join`` and I tried to unify this to ``opj``.

(There is still much to clean up, but this is all I got so far.)

### DIFF
--- a/datalad/distribution/dataset.py
+++ b/datalad/distribution/dataset.py
@@ -51,7 +51,6 @@ from datalad.utils import (
     Path,
     PurePath,
     ensure_list,
-    quote_cmdlinearg,
 )
 
 
@@ -405,7 +404,6 @@ class Dataset(object, metaclass=PathBasedFlyweight):
         -------
         Dataset or None
         """
-        from datalad.coreapi import subdatasets
         path = self.path
         sds_path = path if topmost else None
 
--- a/datalad/tests/utils.py
+++ b/datalad/tests/utils.py
@@ -398,7 +398,7 @@ def _prep_file_under_git(path, filename):
     # do absolute() in addition to always get an absolute path
     # even with non-existing paths on windows
     path = str(Path(path).resolve().absolute())  # intentional realpath to match GitRepo behavior
-    file_repo_dir = op.relpath(path, repo.path)
+    file_repo_dir = relpath(path, repo.path)
     file_repo_path = filename if file_repo_dir == curdir else opj(file_repo_dir, filename)
     return annex, file_repo_path, filename, path, repo
 
@@ -1679,26 +1679,26 @@ def get_convoluted_situation(path, repocls=AnnexRepo):
         ds.save(to_git=True)
         ds.drop([
             'file_dropped_clean',
-            op.join('subdir', 'file_dropped_clean')],
+            opj('subdir', 'file_dropped_clean')],
             check=False)
     # clean and proper subdatasets
     ds.create('subds_clean')
-    ds.create(op.join('subdir', 'subds_clean'))
+    ds.create(opj('subdir', 'subds_clean'))
     ds.create('subds_unavailable_clean')
-    ds.create(op.join('subdir', 'subds_unavailable_clean'))
+    ds.create(opj('subdir', 'subds_unavailable_clean'))
     # uninstall some subdatasets (still clean)
     ds.uninstall([
         'subds_unavailable_clean',
-        op.join('subdir', 'subds_unavailable_clean')],
+        opj('subdir', 'subds_unavailable_clean')],
         check=False)
     assert_repo_status(ds.path)
     # make a dirty subdataset
     ds.create('subds_modified')
-    ds.create(op.join('subds_modified', 'someds'))
-    ds.create(op.join('subds_modified', 'someds', 'dirtyds'))
+    ds.create(opj('subds_modified', 'someds'))
+    ds.create(opj('subds_modified', 'someds', 'dirtyds'))
     # make a subdataset with additional commits
-    ds.create(op.join('subdir', 'subds_modified'))
-    pdspath = op.join(ds.path, 'subdir', 'subds_modified', 'progressedds')
+    ds.create(opj('subdir', 'subds_modified'))
+    pdspath = opj(ds.path, 'subdir', 'subds_modified', 'progressedds')
     ds.create(pdspath)
     create_tree(
         pdspath,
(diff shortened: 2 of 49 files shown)


datalad/distribution/dataset.py
```diff
@@ -405,7 +404,6 @@ def get_superdataset(self, datalad_only=False, topmost=False,
         -------
         Dataset or None
         """
-        from datalad.coreapi import subdatasets
```
yarikoptic, line 408: I thought we still need those "dummy" imports from `.api` and `.coreapi` to get the methods of those bound to dataset instance.  But it seems no longer be the case:
```
$> python -c 'from datalad.distribution.dataset import Dataset; ds = Dataset("."); print(ds.subdatasets())'
[]
```
and code archaeology lead me to my own PR https://github.com/datalad/datalad/pull/3156 which resolved that - interfaces should be autoimported, thus no need for explicit imports, nice and damn that I forgot. cool ;-)

But note that bugs might resurface since those imports weren't really "unused", and when running a full sweep of tests it would likely to import everything so effects might be hidden as if compared to a singular command or test execution in a new process

datalad/tests/utils.py
```diff
@@ -1733,18 +1733,18 @@ def get_convoluted_situation(path, repocls=AnnexRepo):
...
-    os.remove(op.join(ds.path, 'file_deleted'))
-    os.remove(op.join(ds.path, 'subdir', 'file_deleted'))
+    os.remove(opj(ds.path, 'file_deleted'))
+    os.remove(opj(ds.path, 'subdir', 'file_deleted'))
     # staged deletion
     ds.repo.remove('file_staged_deleted')
     # modified files
     if isinstance(ds.repo, AnnexRepo):
-        ds.repo.unlock(['file_modified', op.join('subdir', 'file_modified')])
+        ds.repo.unlock(['file_modified', opj('subdir', 'file_modified')])
```
yarikoptic, line 1747: FWIW, ok going `opj` way since we have a mixed bag of `op.join` and `opj`, but overall I started to prefer using `op.join` form since more explicit etc

yarikoptic (approved):
Did only cursory review, and left only few comments.
Thank you @adswa for doing this!  Since tests are green I have high hopes for it to not reveal some ugly corners, but I also wouldn't be surprised that some would show themselves.  That should not stop us from proceeding this cleaner path - after all it is against master.

Hopefully we manage to kick out next "major" release soon or conflicts with maint might weight us down

datalad/distribution/dataset.py
adswa: > But note that bugs might resurface since those imports weren't really "unused", and when running a full sweep of tests it would likely to import everything so effects might be hidden as if compared to a singular command or test execution in a new process

That doesn't sound desirable! Shall I put those back in?

datalad/tests/utils.py
adswa: I'd be fine with using ``op.join`` to. I will change this. :)

adswa: force-pushed it back to ``op.join``

datalad/distribution/dataset.py
yarikoptic: nah! we better make it work without them, and #3156 promised that it must work that way! ;-)
if anything - paranoid someone could check if an import(s) was added after that PR merge -- may be commit message shared some special knowledge I am missing, but IMHO if there was any - should have been a comment in the code for the import.

datalad/tests/utils.py
yarikoptic: but now tests seems to be unhappy :-/

note also -- we also have `datalad/support/path.py` which provides few extra functions but primarily a replacement for `os.path` imports with some guards added for unicode filenames handling.  I guess neither me nor anyone else switched to use it systematically and was doing imports `import datalad.support.path as op` only whenever running into unicode issues... surprisingly I find only a single import of such kind!
```shell
$> git grep 'import .*\.path.*as op' | grep -v 'os\.path'
datalad/plugin/addurls.py:import datalad.support.path as op
```
but we do few direct ones:

```shell
$> git grep support.path -- datalad
datalad/customremotes/archives.py:from ..support.path import exists
datalad/customremotes/base.py:from ..support.path import exists, join as opj, dirname, lexists
datalad/distribution/tests/test_siblings.py:from datalad.support.path import (
datalad/interface/results.py:from datalad.support.path import robust_abspath
datalad/metadata/aggregate.py:from datalad.support.path import split_ext
datalad/plugin/addurls.py:import datalad.support.path as op
datalad/plugin/addurls.py:from datalad.support.path import split_ext
datalad/plugin/export_to_figshare.py:    from ..support.path import basename
datalad/support/archives.py:from datalad.support.path import (
``` 

doesn't need to be a part of this PR, but I think it might be worth to "solidify" that we use that helper submodule instead of `os.path`

datalad/distribution/dataset.py
adswa: Ah, I see the issue. ``test_utils_catched_dataset.py`` relied on ``utils.py`` to import ``opj``, because it imported it from there. Hopefully my push from a few seconds ago resolves this. :crossed_fingers:

datalad/tests/utils.py
adswa: Ah, I see the issue. ``test_utils_catched_dataset.py`` relied on ``utils.py`` to import ``opj``, because it imported it from there. Hopefully my push from a few seconds ago resolves this. :crossed_fingers:

adswa: I recorded your note about the alternative to ``os.path`` in a new issue, and tagged it as ``good-first-issue``

adswa:
ok, it seems the tests are happy again. Given that this is approved, I'll merge it in a few hours, given that no one voices objections

datalad/tests/utils.py
yarikoptic: > force-pushed it back to op.join

FWIW I still see all changes to opj, odd

yarikoptic:
Its green as we like it. Thank you @adswa

Issue #4097, opened 2020-01-28
Even remove cannot remove without using `-d` way to specify


yarikoptic:
Relates to #TODO-find-that-issue

```shell
$> datalad remove --nocheck -r 346a3ae0-2c2e-11ea-a27d-002590496000/
[ERROR  ] No dataset found [dataset.py:require_dataset:566] (NoDatasetArgumentFound) 
usage: datalad remove [-h] [-d DATASET] [-r] [--nocheck] [--nosave]
                      [-m MESSAGE] [--if-dirty {fail,save-before,ignore}]
                      [PATH [PATH ...]]

$> datalad remove --nocheck -r -d 346a3ae0-2c2e-11ea-a27d-002590496000/
uninstall(ok): unprocessed (dataset)
uninstall(ok): T1w (dataset)
uninstall(ok): MNINonLinear (dataset)
remove(ok): . (dataset)
action summary:
  remove (ok: 1)
  uninstall (ok: 3)
```

so I cannot  `remove ds*` or alike... (although --help suggests that and I hope that we retain such functionality).

effigies:
This is probably better slotted into #TODO-find-that-issue, but the loss of ability to use a lot of commands without an explicit `-d` or `chdir` into a repository has been causing quite a lot of failures on CIs that use datalad to fetch their data. Was this an intentional choice? I'm not sure what was wrong with the old method of finding the containing dataset of a given path. In many cases, what this means is changing:

```Shell
datalad get $DATASET/sub-0{1,2,3}
```

to

```Shell
datalad get -d $DATASET $DATASET/sub-0{1,2,3}
```

Which gets duplicative, especially when you don't have a DATASET variable to use, or have multiple things you're trying to fetch.

mih:
Just quickly: This should do -- like with git

`datalad get $DATASET/sub-0{1,2,3}`

to

`datalad -C $DATASET get sub-0{1,2,3}`

adswa:
I think even within the context of ``remove`` we are inconsistent in handling positional paths/datasets. When I remove a single dataset without use of ``-d``, things work nicely:

```
(handbook2) adina@muninn in /tmp
❱ datalad create superdataset
[INFO   ] Creating a new annex repo at /tmp/superdataset 
[INFO   ] scanning for unlocked files (this may take some time) 
create(ok): /tmp/superdataset (dataset)
(handbook2) adina@muninn in /tmp
❱ datalad remove superdataset
[...]

Issue #4750, opened 2020-07-20
datalad uninstall drops from ephemeral origin


yarikoptic:
Thanks to git-annex tracking it shouldn't lead to data loss but I wonder if that is really a desired behavior e.g. in cases we make ephemeral installs to satisfy YODA principles etc: it leads to wiping out possibly expensive to re-obtain data from the original annex:

```shell
$> datalad create testds; cd testds; datalad install -d . -s ~/datalad/openneuro/ds000001 --reckless=ephemeral
[INFO   ] Creating a new annex repo at /tmp/testds 
[INFO   ] Scanning for unlocked files (this may take some time) 
create(ok): /tmp/testds (dataset)
[INFO   ] Scanning for unlocked files (this may take some time)                                                                                                                                                                    
[INFO   ] access to 1 dataset sibling s3-PRIVATE not auto-enabled, enable with:
| 		datalad siblings -d "/tmp/testds/ds000001" enable -s s3-PRIVATE 
install(ok): /tmp/testds/ds000001 (dataset)
add(ok): /tmp/testds/ds000001 (file)
add(ok): /tmp/testds/.gitmodules (file)
save(ok): /tmp/testds (dataset)
action summary:
  add (ok: 2)
  install (ok: 1)
  save (ok: 1)

$> ls -l ds000001/.git/annex -d
lrwxrwxrwx 1 yoh yoh 47 Jul 20 09:36 ds000001/.git/annex -> /home/yoh/datalad/openneuro/ds000001/.git/annex/

$> datalad uninstall ds000001
drop(ok): /tmp/testds/ds000001/sub-01/anat/sub-01_T1w.nii.gz (file) [checking s3-PUBLIC...]
drop(ok): /tmp/testds/ds000001/sub-01/anat/sub-01_inplaneT2.nii.gz (file) [checking s3-PUBLIC...]
drop(ok): /tmp/testds/ds000001/sub-01/func/sub-01_task-balloonanalogrisktask_run-01_bold.nii.gz (file) [checking s3-PUBLIC...]
drop(ok): /tmp/testds/ds000001 (directory)
uninstall(ok): /tmp/testds/ds000001 (dataset)
action summary:
  drop (ok: 4)
  uninstall (ok: 1)

$> ls -L ~/datalad/openneuro/ds000001/sub-01/anat/sub-01_T1w.nii.gz 
[...]

Issue #3887, opened 2019-11-22
remove() should announce `annex dead`


mih:
Did not investigate, but it occurred to me that with short-lived one-off clones of a dataset, we might track tons of annexes that are long gone and can never come back.

yarikoptic:
yes, it will be the case.  It might be impossible to "announce" back in origin (e.g. I am offline, I should be able to "remove" without being able to talk back).  There should be analysis if "origin" (its git-annex branch) or actually any sibling, knows about this UUID. If not -- it should be ok to just kill it without announcing it dead to anyone.

In addition to announcing removed dataset, I wondered if we could provide `--throw-away` option to `install` which could help preventing proliferation of `git-annex` branch with records for all those deadly ones somehow.

May be we could 
- mint a new git annex UUID for the clone deterministically (so - all such clones would have the same UUID) based on the UUID of the original one
- give that UUID to annex instead of the one it would create
- pronounce that clone untrusted

There is a flaw in above -- git-annex might become unhappy if from some other clone it inherits information about availability of that clone.

Just an idea

mih:
FTR: We are already working on an implementation of such a thing, but using a different approach. Will report once complete and found successful.

mih:
Ping https://github.com/datalad/datalad/issues/3528

PR #5815, opened 2021-07-23
UX: Make remove work without -d if the path matches a dataset

adswa:
This is a pretty naive attempt at fixing #4097, and the fact that ``datalad remove someds`` works while ``datalad remove somesuperds`` requires ``-d``/``--dataset``. It adds a basic check whether a single supplied path is a dataset instead of a file to remove, and if it is, it treats the path as the dataset.

Since ``remove`` seems to do this anyway for datasets without subdatasets, I hope this change isn't too controversial. But check if I have overlooked something here, after all, its a quite destructive command. I threw all possible checks into the conditional in hopes to really only catch datasets disguised as paths - let me know if you know a more elegant way to do this.

### DIFF
--- a/datalad/distribution/remove.py
+++ b/datalad/distribution/remove.py
@@ -120,6 +120,12 @@ class Remove(Interface):
             # act on the whole dataset if nothing else was specified
             # TODO i think that would happen automatically in annotation?
             path = refds_path
+        if path and len(path) == 1 and not os.path.isfile(path[0]) and \
+                not refds_path and GitRepo.is_valid_repo(path[0]):
+            # The path we were given is actually a dataset
+            refds_path = Interface.get_refds_path(path[0])
+            res_kwargs['refds'] = refds_path
+            path = None
 
         to_process = []
 


adswa:
ahh, here is where my list of conditions was too naive:

```
  File "/tmp/datalad_temp_test_spec2bidsggltb_gy/code/hirni-toolbox/procedures/hirni-dicom-converter.py", line 175, in <module>
    message="[HIRNI] Remove empty *_event.tsv "
  File "/opt/hostedtoolcache/Python/3.7.11/x64/lib/python3.7/site-packages/datalad/distribution/dataset.py", line 505, in apply_func
    return f(**kwargs)
  File "/opt/hostedtoolcache/Python/3.7.11/x64/lib/python3.7/site-packages/datalad/interface/utils.py", line 483, in eval_func
  File "/opt/hostedtoolcache/Python/3.7.11/x64/lib/python3.7/site-packages/datalad/interface/utils.py", line 476, in return_func
    results = list(results)
  File "/opt/hostedtoolcache/Python/3.7.11/x64/lib/python3.7/site-packages/datalad/interface/utils.py", line 408, in generator_func
    allkwargs):
  File "/opt/hostedtoolcache/Python/3.7.11/x64/lib/python3.7/site-packages/datalad/interface/utils.py", line 574, in _process_results
    for res in results:
  File "/opt/hostedtoolcache/Python/3.7.11/x64/lib/python3.7/site-packages/datalad/distribution/remove.py", line 123, in __call__
    if path and len(path) == 1 and not os.path.isfile(path[0]) and \
TypeError: 'set' object is not subscriptable
ERROR
```

mih:
This is not to discourage such a move, but merely to ensure awareness: `remove` needs a rewrite, because it blocks https://github.com/datalad/datalad/issues/3368

mih:
Started thinking about this in more depth and I am hesitant to make this move. Not because I object, but because the situation is more complex. I think we first need to actually discuss a sensible behavior. This affects not only `remove` but also `uninstall` and to some degree `drop`. Traditionally, we have been more focused on getting data, not very much on carefully getting rid of datasets (not the focus on datasets, not files).

We uninstall datasets easily, even if they contain unique changes https://github.com/datalad/datalad/issues/1142, debate whether `remove` or `uninstall` should operate on top-level datasets https://github.com/datalad/datalad/issues/2967, and do not communicate gone clones properly https://github.com/datalad/datalad/issues/3887. Dropping or not dropping is a non-trivial question (#1282 vs #1823), and sometimes even dangerous ATM #4750.

This PR would make it even easier to make things worse IMHO.

It would be extremely helpful, if someone would come up with a proper behavior concept for this kind of functionality that we could check against all these issues for sanity and applicability.

adswa:
In light of our discussions in #5832 and #5842, which try to tackle a variety of issues with `remove`, `uninstall`, and `drop` - including this one - in a more holistic manner, I will close this PR.

PR #5816, opened 2021-07-23
merge maint into master (brings auto releasing)

yarikoptic:
If all is good, will push the merge as master (if in time) to avoid too "deep" merges... if not -- could as well merge here

    Merge commit '0.14.6-16-g3bb1e9387' into enh-merge-maint
    
    Main change causing conflicts etc merged is having 'auto' releases
    
    * commit '0.14.6-16-g3bb1e9387': (38 commits)
      BF(UX): do not claim to be "Authenticated" access if we do not know
      BF: provide DataLad User-Agent to please wikipedia
      ENH(UX,DX): inform user with a warning if version is 0+unknown
      ENH: refresh Singularity.git-annex-dev to use fresh bullseye etc
      ENH(TST): explicit test that .version is string or list
      ENH(TST): make assertion fail message descriptive
      BF: split also on + not only .
      ENH: add argcomplete to misc requires
      RF internal tools/cmdline-completion to use datalad shell-completion
      ENH: log an ERROR if argcomplete fails to import
      add python-gitlab dependency
      [skip ci] Update RST changelog
      Update CHANGELOG.md [skip ci]
      Run update-changelog as part of auto release
      BF: we should also allow for "tests" label without "patch" or "minor"
      [DATALAD RUNCMD] make update-changelog
      fix update-changelog to provide conversion of changelog to rst
      Update CHANGELOG.md [skip ci]
      RF(TST): disable test on not consuming all produced items
      BF(TST): parallel - take longer for producer to produce
      ...
    
    Conflicts:
            CHANGELOG.md -- all due to auto-ing done in maint
            CONTRIBUTING.md
            datalad/version.py


yarikoptic:
hm... something went wrong since only appveyor ran, no travis etc... hm...

yarikoptic:
linting is codespell: `datalad/_version.py:277: unparseable ==> unparsable` .  Sent https://github.com/python-versioneer/python-versioneer/pull/260 and will skip this file from codespell.  I will not push here unless some other fix to come up.

yarikoptic:
pushed a fresh merge, shouldn't forget to push it as master if all green...

yarikoptic:
grr you linters! will fix `noone` -> `nobody` locally

yarikoptic:
pushed merge into master... hopefully nothing got screwed up

PR #5817, opened 2021-07-23
RF(TST): do not test for no EASY and pkg_resources in shims

### LINKED ISSUE #5384: FAIL: datalad.cmdline.tests.test_main.test_script_shims
Fails in build of Debian package on bullseye

```
FAIL: datalad.cmdline.tests.test_main.test_script_shims
  File "/usr/lib/python3/dist-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/build/datalad-0.14.0~rc1/.pybuild/cpython3_3.9_datalad/build/datalad/cmdline/tests/test_main.py", line 213, in test_script_shims
    assert_not_in('EASY', content) # NOTHING easy should be there
AssertionError: 'EASY' unexpectedly found in "#!/usr/bin/python3.9\n# EASY-INSTALL-ENTRY-SCRIPT: 'datalad','console_scripts','datalad'\nimport re\nimport sys\n\n# for compatibility with easy_install; see #2198\n__requires__ = 'datalad'\n\ntry:\n    from importlib.metadata import distribution\nexcept ImportError:\n    try:\n        from importlib_metadata import distribution\n    except ImportError:\n        from pkg_resources import load_entry_point\n\n\ndef importlib_load_entry_point(spec, group, name):\n    dist_name, _, _ = spec.partition('==')\n    matches = (\n        entry_point\n        for entry_point in distribution(dist_name).entry_points\n        if entry_point.group == group and entry_point.name == name\n    )\n    return next(matches).load()\n\n\nglobals().setdefault('load_entry_point', importlib_load_entry_point)\n\n\nif __name__ == '__main__':\n    sys.argv[0] = re.sub(r'(-script\\.pyw?|\\.exe)?$', '', sys.argv[0])\n    sys.exit(load_entry_point('datalad', 'console_scripts', 'datalad')())\n"
```

yarikoptic:
Awhile back we did switch away from creating our own entry point
scripts, and just succumbed to all the goodness default mechanisms
give us.  We had to skip this test on debian builds, where we have
to run something like   setup.py develop  to produce those entry points
for the tests.  That added up burden etc.  So I decided we should no
longer bother with such tests, also because pkg_resources now used
just as a fall back in those produced by setuptools, so typically there
should be no big penalty in start time.

Closes #5384

### DIFF
--- a/datalad/cmdline/tests/test_main.py
+++ b/datalad/cmdline/tests/test_main.py
@@ -189,12 +189,6 @@ def test_incorrect_options():
 
 def test_script_shims():
     runner = Runner()
-    # The EASY-INSTALL checks below aren't valid for editable installs. Use the
-    # existence of setup.py as an indication that install is _probably_
-    # editable. The file should always exist for editable installs, but it can
-    # also exist for non-editable installs when the tests are being executed
-    # from the top of the source tree.
-    setup_exists = (Path(datalad.__file__).parent.parent / "setup.py").exists()
     for script in [
         'datalad',
         'git-annex-remote-datalad-archives',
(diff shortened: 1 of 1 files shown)


mih:
Filed https://github.com/datalad/datalad/issues/5819 for the test failure. Remaining failure is known and unrelated (mac os CI fragility).

yarikoptic:
ok, no objections, tests fails unrelated, let's proceed

PR #5825, opened 2021-07-26
BF: RIARemote - set UI backend to annex to make it interactive

### LINKED ISSUE #5790: Design flaw using downloaders in ORA special remote?
I have a dataset with a configured ORA remoted that points to a password-protected HTTP server.

```
% git annex get Simulation2GenerateRun.m
get Simulation2GenerateRun.m (from inm7-storage...) 
[INFO] Downloading 'https://archive.ds.inm7.de/c27/71672-96c4-11ea-be6a-a0369f287950/annex/objects/GX/88/MD5E-s1749--847057de59b55cf34e98757525ba7809.m/MD5E-s1749--847057de59b55cf34e98757525ba7809.m' into '.git/annex/tmp/MD5E-s1749--847057de59b55cf34e98757525ba7809.m' 
[INFO] http session: Authenticating into session for https://archive.ds.inm7.de/c27/71672-96c4-11ea-be6a-a0369f287950/annex/objects/GX/88/MD5E-s1749--847057de59b55cf34e98757525ba7809.m/MD5E-s1749--847057de59b55cf34e98757525ba7809.m 

  A non-interactive silent UI was asked for a response to a question: user. Title: You need to authenticate with 'ds.inm7.de' credentials..
```

There is no way for me to enter these credentials interactively, because interactivity is not possible at this stage (git-annex talks to the special remote binary).

I have

```
[provider:ds.inm7.de]
url_re = https.*://.*ds\.inm7\.de
credential = ds.inm7.de

[credential:ds.inm7.de]
type = user_password
```

but I do not think this matters or is suboptimal. I believe this is a design flaw. The only way out would be to prime the credential store prior to the first `get`. The necessary credentials are known to git (git clone works fine), but #5028 is not implemented. This only leaves #5680

yarikoptic:
Otherwise our UI would consider session non interactive and would not
prompt for credentials etc.

Closes #5790

### DIFF
--- a/datalad/distributed/ora_remote.py
+++ b/datalad/distributed/ora_remote.py
@@ -1217,7 +1217,9 @@ class RIARemote(SpecialRemote):
 def main():
     """cmdline entry point"""
     from annexremote import Master
+    from datalad.ui import ui
     master = Master()
     remote = RIARemote(master)
+    ui.set_backend('annex')  # interactive, stdin/stdout will be used for interactions with annex
     master.LinkRemote(remote)
     master.Listen()


datalad/distributed/ora_remote.py
```diff
@@ -1217,7 +1217,9 @@ def _get_obj_location(self, key):
 def main():
     """cmdline entry point"""
     from annexremote import Master
+    from datalad.ui import ui
     master = Master()
     remote = RIARemote(master)
+    ui.set_backend('annex')  # interactive, stdin/stdout will be used for interactions with annex
```
bpoldrack, line 1223: Isn't something like this needed as well? https://github.com/datalad/datalad/blob/66fd8be2cadeda3b85665ae63d407e2241e0bb1a/datalad/customremotes/base.py#L119

Not sure, but if so - that's a problem.

yarikoptic: good catch -- missed that.  I assume you (`AnnexRemote`?) already interfaces progress indication somehow - probably directly for RIA operations without exposing any possible progressbar?  I guess may be a similar subclass/backend needs to be crafted which would either not care about progress bars or provide adapter for `AnnexRemote` interface and not built-in customremotes/base class?.

bpoldrack: > I assume you (AnnexRemote?) already interfaces progress indication somehow - probably directly for RIA operations without exposing any possible progressbar?

Yes, reporting progress directly to git-annex via `annexremote`.

yarikoptic: I guess the easiest then is to establish a dedicated UI for it (e.g. "called" `annexremote`) which would somehow swallow or just raise and exception in cat any possible progress bar requested (e.g. if ora invokes some command which it doesn't know leads to establishing some progress bar).
Anyways -- I would probably better get some "authenticated" ORA setup to mimic your case to actually test it all up... or interested @bpoldrack to take over?
also filed https://github.com/datalad/datalad/issues/5873 which might be of help/interest here

bpoldrack: > I would probably better get some "authenticated" ORA setup to mimic your case to actually test it all up... or interested @bpoldrack to take over?

I hope to come up with HTTPS setup in a docker image for that next week and rewrite/enhance ORA tests. We'll see where that leads ... Once such setup is running, will look into this again.

mih:
It seems there is no interest in fixing this issue. With #6165 the other special remote implementations are now also AnnexRemote based, but the continue to use a different approach to enabling this feature (set in constructor). However, I think this one here (set in main) makes more sense, and I will standardize all three on this behavior.

PR #5839, opened 2021-07-29
UX: When two or more clone URL templates are found, error out more gracefully

adswa:
This is cherry-picked from #5644 because it fixes an issue unrelated to that PR. If someone ends up with two configurations for source candidates with identical names, one would see
```
❱ datalad get HCP1200/100206                                                1 !
[ERROR  ] 'tuple' object has no attribute 'format' [get.py:_get_flexible_source_candidates_for_submodule:240] (AttributeError) 
```

With this change, a user gets a more helpful error message:
```
❱ datalad get HCP1200/100206                                                2 !
[ERROR  ] There are multiple URL templates for submodule clone candidate 'origin', but only one is allowed. Check your configuration! [get.py:_get_flexible_source_candidates_for_submodule:245] (ValueError) 
```

### DIFF
--- a/datalad/distribution/get.py
+++ b/datalad/distribution/get.py
@@ -243,6 +243,13 @@ def _get_flexible_source_candidates_for_submodule(ds, sm):
                         ds_repo.config[c])
                        for c in ds_repo.config.keys()
                        if c.startswith(candcfg_prefix)]:
+        # ensure that there is only one template of the same name
+        if type(tmpl) == tuple and len(tmpl) > 1:
+            raise ValueError(
+                f"There are multiple URL templates for submodule clone "
+                f"candidate '{name}', but only one is allowed. "
+                f"Check your configuration!"
+            )
         url = tmpl.format(**sm_candidate_props)
         # we don't want "flexible_source_candidates" here, this is
         # configuration that can be made arbitrarily precise from the
--- a/datalad/distribution/get.py
+++ b/datalad/distribution/get.py
@@ -248,7 +248,7 @@ def _get_flexible_source_candidates_for_submodule(ds, sm):
             raise ValueError(
                 f"There are multiple URL templates for submodule clone "
                 f"candidate '{name}', but only one is allowed. "
-                f"Check your configuration!"
+                f"Check datalad.get.subdataset-source-candidate-* configuration!"
             )
         url = tmpl.format(**sm_candidate_props)
         # we don't want "flexible_source_candidates" here, this is
--- a/datalad/distribution/tests/test_get.py
+++ b/datalad/distribution/tests/test_get.py
@@ -173,6 +173,17 @@ def test_get_flexible_source_candidates_for_submodule(t, t2, t3):
          for i in f(clone3, clone3.subdatasets(return_type='item-or-list'))]
     )
 
+    # check #5839: two source configs with the same name should raise an error
+    clone3.config.add(
+        f"datalad.get.subdataset-source-candidate-{DEFAULT_REMOTE}",
+        "should-not-work"
+    )
+    clone3.config.add(
+        f"datalad.get.subdataset-source-candidate-{DEFAULT_REMOTE}",
+        "should-really-not-work"
+    )
+    assert_raises(ValueError, clone3.get, 'sub')
+
     # TODO: check that http:// urls for the dataset itself get resolved
     # TODO: many more!!
 


datalad/distribution/get.py
```diff
@@ -237,6 +237,16 @@ def _get_flexible_source_candidates_for_submodule(ds, sm):
                         ds_repo.config[c])
                        for c in ds_repo.config.keys()
                        if c.startswith(candcfg_prefix)]:
+        # ensure that there is only one template of the same name
+        if type(tmpl) == tuple and len(tmpl) > 1:
+            lgr.debug('Found %i URL templates for the submodule clone '
```
bpoldrack, line 242: There's no need for that log message. We are trying to get rid of the various forms of double-reporting ;-)
If you throw an exception, then it either bubbles up, leading to its message being logged at ERROR level or it's catched and dealt with upstairs, in which case a proper logging of what's going on should happen at that spot (where we know the relevance of it).

adswa: ah yeah, makes sense!

adswa: removed it

datalad/distribution/get.py
```diff
@@ -237,6 +237,13 @@ def _get_flexible_source_candidates_for_submodule(ds, sm):
                         ds_repo.config[c])
                        for c in ds_repo.config.keys()
                        if c.startswith(candcfg_prefix)]:
+        # ensure that there is only one template of the same name
+        if type(tmpl) == tuple and len(tmpl) > 1:
+            raise ValueError(
+                "There are multiple URL templates for submodule clone "
+                "candidate '{}', but only one is allowed. "
+                "Check your configuration!".format(name)
```
yarikoptic, line 245:
```suggestion
                f"candidate '{name}', but only one is allowed. "
                "Check your configuration!"
```
since we are `>= 3.6` in master so we could take advantage from a modern python to make code more readable etc.

also, would it be too much to ask for a test to ensure that issue doesn't resurface and that we don't fall afoul of some variables renaming? ;)

adswa: Oh, I always thought fstrings where python 3.7 :upside_down_face: - will do!

adswa: >  would it be too much to ask for a test to ensure that issue doesn't resurface and that we don't fall afoul of some variables renaming? ;)

no, will add one :)

adswa:
oh no the semverlabel check is here...

datalad/distribution/get.py
adswa: test is added.

adswa: f-string too

yarikoptic:
Looks good to me @adswa and thank you for the test! I changed label from `internal` (for internal changes/tools/utils some CI setup, which aren't delivered to the user) to `patch` (e.g. a fix to a bug), to trigger CI fail, only to realize that 
- my desires I expressed to @jwodder to ensure that `master` always has `minor` might be flawed in case of submitting a PR fixing a bug introduced in master and not yet released
- that this PR might actually better fit to be for `maint` branch as what it is, isn't it? ;-)

yarikoptic:
eh, so we are starting to hit limits on how many OSX builders are available to github actions:
```
2021-07-29T15:17:38.9242652Z The agent pool assigned to this job has hit their MacOs concurrency limits
2021-07-29T15:17:38.9475063Z Found online and idle hosted runner(s) in the current repository's organization account that matches the required labels: 'macos-latest'
2021-07-29T15:17:38.9475164Z Waiting for a hosted runner in 'organization' to pick this job...
```

as for semver, just pushed directly to maint and master allowing for `patch` within `master` branch. See 7f543d00563e8abbcd74742a59c356ac9b34ed87 . But still -- this PR might be better of to target `maint`

datalad/distribution/get.py
```diff
@@ -237,6 +237,13 @@ def _get_flexible_source_candidates_for_submodule(ds, sm):
                         ds_repo.config[c])
                        for c in ds_repo.config.keys()
                        if c.startswith(candcfg_prefix)]:
+        # ensure that there is only one template of the same name
+        if type(tmpl) == tuple and len(tmpl) > 1:
+            raise ValueError(
+                f"There are multiple URL templates for submodule clone "
+                f"candidate '{name}', but only one is allowed. "
+                f"Check your configuration!"
```
mih, line 245:
```suggestion
                f"Check datalad.get.subdataset-source-candidate-* configuration!"
```

So users have an immediate idea what to look for.

mih (approved):
Great, thx. I hope it would not matter if this ends up in maint or master. I am fine with either choice.

datalad/distribution/get.py
adswa: that makes sense, thx for the suggestion

adswa:
rebased to ``maint`` :+1:

adswa:
A single failing test:

```

FAIL: datalad.tests.test_witless_runner.test_asyncio_forked

  File "/tmp/dl-miniconda-60pqdy8v/lib/python3.9/site-packages/nose/case.py", line 198, in runTest

    self.test(*self.arg)

  File "/tmp/dl-miniconda-60pqdy8v/lib/python3.9/site-packages/datalad/tests/utils.py", line 764, in _wrap_with_tempfile

  File "/tmp/dl-miniconda-60pqdy8v/lib/python3.9/site-packages/datalad/tests/test_witless_runner.py", line 221, in test_asyncio_forked

    raise AssertionError("Child process did not create a file we expected!")

AssertionError: Child process did not create a file we expected!
```

have never seen this one before

mih:
That test failure is unrelated.

datalad/distribution/get.py
adswa: Thank you for noticing this! I meant to commit it, and thought I did, but it must have gotten lost somewhere - I still had the commit message for it cached

adswa: it should be applied now

datalad/distribution/get.py
```diff
@@ -243,6 +243,13 @@ def _get_flexible_source_candidates_for_submodule(ds, sm):
                         ds_repo.config[c])
                        for c in ds_repo.config.keys()
                        if c.startswith(candcfg_prefix)]:
+        # ensure that there is only one template of the same name
+        if type(tmpl) == tuple and len(tmpl) > 1:
+            raise ValueError(
+                f"There are multiple URL templates for submodule clone "
+                f"candidate '{name}', but only one is allowed. "
+                f"Check datalad.get.subdataset-source-candidate-* configuration!"
```
yarikoptic, line 251: just a note -- (although I like consistency of `f"` prefix too) - `f"` is not needed on the lines where no `{format}` was used

yarikoptic (approved):
If no surprises from the tests -- looks good to me, thanks @adswa

yarikoptic:
whenever tests pass, lets merge to release (added a label)

yarikoptic:
appveyor -- macs + ubuntu freaked out 
```
git clone -q https://github.com/datalad/datalad.git /home/appveyor/projects/datalad
git fetch -q origin +refs/pull/5839/merge:
git checkout -qf FETCH_HEAD
A WebException with status SendFailure was thrown.
```

Let's proceed

datalad/distribution/get.py
adswa: ha, good to know! thanks!

PR #5841, opened 2021-07-29
DOC: Move man pages for push from 'plumbing' into 'dataset operations'

### LINKED ISSUE #5840: "Push" is still documented under plumbing in the docs
I keep failing to find the doc page for push because it is so far down. It is probably best positioned in the section "dataset operations", especially in light of #5837

adswa:
Fixes #5840


mih (approved):
Thx!

PR #2912, opened 2018-10-11
RF: Replace custom .gitattributes handling with standard method

mih:

PR #5847, opened 2021-07-30
Append (not prepend) a newline character(s) to .gitattributes for prettier appearance

AKSoo:
Current `.gitattributes` files have this weird structure of an empty line at start and no newline at end. This looks ugly when the files are viewed in terminal.

The solution is very simple and doesn't seem to break anything.

### DIFF
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -2836,7 +2836,7 @@ class GitRepo(CoreGitRepo):
                         attrline += ' -{}'.format(a)
                     else:
                         attrline += ' {}={}'.format(a, val)
-                f.write('\n{}'.format(attrline))
+                f.write('{}\n'.format(attrline))
 
     def get_content_info(self, paths=None, ref=None, untracked='all',
                          eval_file_type=True):
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -2819,8 +2819,8 @@ class GitRepo(CoreGitRepo):
         with open(git_attributes_file, mode + '+') as f:
             # for append, fix existing files that do not end with \n
             if mode == 'a' and f.tell():
-                f.seek(f.tell() - 1)
-                if f.read() != '\n':
+                f.seek(max(0, f.tell() - len(os.linesep)))
+                if not f.read().endswith('\n'):
                     f.write('\n')
 
             for pattern, attr in sorted(attrs, key=lambda x: x[0]):
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -2821,7 +2821,7 @@ class GitRepo(CoreGitRepo):
             if mode == 'a':
                 f.seek(0)
                 s = f.read()
-                if len(s) > 1 and s[-1] != '\n':
+                if s and not s.endswith('\n'):
                     f.write('\n')
 
             for pattern, attr in sorted(attrs, key=lambda x: x[0]):
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -2818,10 +2818,9 @@ class GitRepo(CoreGitRepo):
 
         with open(git_attributes_file, mode + '+') as f:
             # for append, fix existing files that do not end with \n
-            if mode == 'a':
-                f.seek(0)
-                s = f.read()
-                if s and not s.endswith('\n'):
+            if mode == 'a' and f.tell():
+                f.seek(f.tell() - 1)
+                if f.read() != '\n':
                     f.write('\n')
 
             for pattern, attr in sorted(attrs, key=lambda x: x[0]):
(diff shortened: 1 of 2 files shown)


datalad/support/gitrepo.py
```diff
@@ -2836,7 +2836,7 @@ def set_gitattributes(self, attrs, attrfile='.gitattributes', mode='a'):
                         attrline += ' -{}'.format(a)
                     else:
                         attrline += ' {}={}'.format(a, val)
-                f.write('\n{}'.format(attrline))
+                f.write('{}\n'.format(attrline))
```
yarikoptic, line 2839: Unfortunately it wouldn't be as safe: default mode is 'a' so we aren't rewriting the file but just adding a new record. What is it lacked final new line to start with?

yarikoptic: But may be we could avoid writing an initial new line if eg file didn't exist or had size 0

AKSoo: We could check whether the file (if it exists) ends with \n. If not, start with \n.

yarikoptic: could be a windows one AFAIK (was it `\r\n` or `\n\r`? has been a while) so care should be taken.
worth `git blame` and looking at the original comments/PRs on motivation why we have done it this way

yarikoptic: an alternative - `if mode=='a'` `readlines` it in full first (and start with empty list otherwise), prepare all extra lines, and then write it out consistently in full.  I do not think that performance hit would be notable... again -- worth checking original PR(s) on either such idea was ruled out for some reason (beyond "performance" which is IMHO is nonissue)

AKSoo: This was the motivation behind the change. https://github.com/datalad/datalad/pull/2912#issuecomment-429224540

Windows is `\r\n`, so there shouldn't be an issue with my suggestion.

AKSoo: I'm not familiar with how datalad tests, but it seems there is no test for the case where file exists and last character isn't `\n`?

AKSoo:
Aesthetics aside, it is apparently a [bad practice](https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline) in general to not end with newline. Several command line tools break. There is a reason git often points out missing newlines at EOF.

datalad/support/gitrepo.py
```diff
@@ -2815,7 +2815,15 @@ def set_gitattributes(self, attrs, attrfile='.gitattributes', mode='a'):
...
         if not op.exists(attrdir):
             os.makedirs(attrdir)
-        with open(git_attributes_file, mode) as f:
+
+        with open(git_attributes_file, mode + '+') as f:
+            # for append, fix existing files that do not end with \n
+            if mode == 'a':
+                f.seek(0)
+                s = f.read()
+                if len(s) > 1 and s[-1] != '\n':
```
yarikoptic, line 2824:
```suggestion
                if s and not s.endswith('\n'):
```
is a bit more pythonic to me

datalad/support/gitrepo.py
```diff
@@ -2815,7 +2815,15 @@ def set_gitattributes(self, attrs, attrfile='.gitattributes', mode='a'):
...
+                    f.write('\n')
```
yarikoptic, line 2825: add a test to ensure we handle it right?

yarikoptic (changes requested):
overall, I am ok with this approach if works as expected (needs a test)

datalad/support/gitrepo.py
AKSoo: Yeah, I think so too. It will take some time for me to figure out nosetests.

bpoldrack:
Your last assertion in the test is failing on windows. Need to double-check, but this suggests to me that the line ending handling is wrong. While python3 should by default translate `\n` properly when `open` is called with no specified `newline` option, this likely only applies to `read/write` methods. You are using `seek/tell/truncate` instead, but they referring to bytes, not "characters" like `\r\n`.

So, in your test, your cutting of the `\n` but not the `\r`. But the `f.read()` in your implementation reads the `\r` and still returns `\n` (somewhat relaxed auto-translation), failing your condition to add the missing `\n`. However, we are then reading the file via `git config -z` where git is responsible for determining what is a line and it only sees a `\r`.

I haven't double-checked all of that, but in any case: Simply because `seek` doesn't treat `\r\n` as a single byte, I'd prefer to only use functions, that include python's auto-translation. So, instead of `f.seek(f.tell() -1)`, just check `f.readlines()[-1][-1] == '\n'`. If so, prepend the to be written output with an `\n` and be done.

AKSoo:
> I haven't double-checked all of that, but in any case: Simply because `seek` doesn't treat `\r\n` as a single byte, I'd prefer to only use functions, that include python's auto-translation. So, instead of `f.seek(f.tell() -1)`, just check `f.readlines()[-1][-1] == '\n'`. If so, prepend the to be written output with an `\n` and be done.

`read` does auto-translate OS-specific newlines, so the problem was `f.tell() - 1` assuming 1 char line ending. I fixed that to use `len(os.linesep)` instead. Also rewrote the test to just strip whitespace at the end.

bpoldrack:
That's an option, too - I'm fine with that. :-)

yarikoptic:
thank you @AKSoo , let's proceed and see where it takes us! ;)

PR #5851, opened 2021-07-30
nf: export-archive-ora add options to filter files exported to the 7z archive

### LINKED ISSUE #5845: export-archive-ora filter files based on git-annex matching options.
#### What is the problem?

More a feature request / feasibility question: 
Currently, when exporting an archive for ORA, it will add all keys present in the current repo to the archive, while one might want to filter the archive content (eg. based on distribution-restriction metadata) similar to the `wanted` option for special-remotes.

Do you think it would be possible to replace the following glob:
https://github.com/datalad/datalad/blob/master/datalad/distributed/export_archive_ora.py#L139
with a call to `git-annex find --format='.git/annex/objects/${hashdirmixed}${key}/${key} '` with optional [git-annex-matching-options](https://git-annex.branchable.com/git-annex-matching-options/) passed as an option to the `export-archive-ora` command line?
If you think this is of interest, I could work on a PR.

Thanks!

bpinsard:
changes to command export-archive-ora:

- adds `--for <remote>` which will apply the remote `wanted` expression to select files to archive
- adds `--filters <git-annex-preferred-content>` to provide additional filters to select files to archive
- adds `--branch=<tree-ish>` to select files based on a specific repo state.

Closes #5845 

TODO:

- [x] Add tests
- [x] Handle missing files in current repo (raise or ignore).

### DIFF
--- a/datalad/distributed/export_archive_ora.py
+++ b/datalad/distributed/export_archive_ora.py
@@ -77,20 +77,23 @@ class ExportArchiveORA(Interface):
             constraints=EnsureStr() | EnsureNone()),
         remote=Parameter(
             args=("--for",),
+            dest="remote",
             metavar='LABEL',
             doc="""name of the target sibling, wanted/preferred settings
             will be used to filter the files added to the archives""",
             constraints=EnsureStr() | EnsureNone()),
-        filters=Parameter(
-            args=("--filters",),
+        annex_wanted=Parameter(
+            args=("--annex-wanted",),
             metavar="FILTERS",
             doc="""git-annex-preferred-content expression for
             git-annex find to filter files. Should start with
             'or' or 'and' when used in combination with `--for`"""),
-        branch=Parameter(
-            args=("--branch",),
-            metavar="BRANCH",
-            doc="""a branch, or tree-ish to use to select files"""),
+        froms=Parameter(
+            args=("--from",),
+            dest="froms",
+            metavar="FROM",
+            nargs="+",
+            doc="""one or multiple tree-ish from which to select files"""),
         opts=Parameter(
             args=("opts",),
             nargs=REMAINDER,
@@ -107,8 +110,8 @@ class ExportArchiveORA(Interface):
             opts=None,
             dataset=None,
             remote=None,
-            filters=None,
-            branch=None):
+            annex_wanted=None,
+            froms=None):
         # only non-bare repos have hashdirmixed, so require one
         ds = require_dataset(
             dataset, check_installed=True, purpose='export to ORA archive')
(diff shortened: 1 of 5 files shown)


mih:
Thanks for the PR

while still in draft mode, some thoughts on the API.

> * adds `--filters <git-annex-preferred-content>` to provide additional filters to select files to archive

"Filters" a very generic term, the proposed `--branch` also is a filter. I'd say it is better to limit the scope in the name. `create-sibling`, for example, uses `--annex-wanted`.

> * adds `--branch=<tree-ish>` to select files based on a specific repo state.

It would be good to get input from the others on this one. `--branch` is used by git-annex for this purpose, so it makes sense. However, it specially is not only about branches (tree-ish), so it may not be an optimal name.
Moreover, it seems to make sense to consider a (later) extension to support multiple tree-ishes.

What do you think about `--from <tree-ish>` as an alternative name, given that it defines the frame of reference to perform a key lookup for?

bpoldrack:
> What do you think about --from <tree-ish> as an alternative name, given that it defines the frame of reference to perform a key lookup for?

`--from` is in-line with `datalad diff`, so it makes sense to me.

bpinsard:
Thanks for your feedback, I changed the naming of the command-line args:
- `--annex-wanted` to provide extra filters
- `--from` to select tree-ishes (now allow multiple) from which to select files.

mih:
Sorry @bpinsard that this got ignored for so long, and thanks for the updates! I have approved the CI runs now. I will review again once they are finished.

mih:
The linter failure is a false-positive.

datalad/distributed/export_archive_ora.py
```diff
@@ -103,6 +137,9 @@ def __call__(
         else:
             archive.parent.mkdir(exist_ok=True, parents=True)
 
+        if isinstance(froms, str):
+            froms = [froms]
```
mih, line 141:
```suggestion
        froms = ensure_list(froms)
```

plus the `from datalad.utils import ensure_list`

mih:
@bpoldrack I did not comment on this PR after the last set of changes. Do you need more time, or are you OK with getting this merged?

bpinsard:
No worries @mih , also I am now on pat-leave so will not be able to follow up quickly until January.

bpoldrack:
@mih

I am OK with this.

mih:
Ok, in this case, I will take over here and get this merged. Thx!

Enjoy the time @bpinsard !

mih:
I am closing this PR. It is replaced by https://github.com/datalad/datalad/pull/6234 -- where I resolved the conflicts, and addressed my own comment from above. Thx @bpinsard !

PR #5861, opened 2021-08-03
mimic new logic of git-annex to split_ext

yarikoptic:
The change was introduced in git-annex today
in

[8.20210803-2-g899983058](https://git.kitenet.net/index.cgi/git-annex.git/commit/?id=899983058f77fbeb67a2ac04016ea26fec5468fa)

in response to my report

https://git-annex.branchable.com/bugs/should_not_assume_entire_name_to_be_an_extension__63__/

I have decided to make it git-annex version specific since most likely use is
"along side with annex operation" although I could be wrong. ATM we use this
one in only few spots:

	(git)lena:~datalad/datalad-maint[rf-annex-dotfile-extension]git-annex
	$> git grep split_ext | grep -v 'test_'
	datalad/metadata/aggregate.py:from datalad.support.path import split_ext
	datalad/metadata/aggregate.py:        split_ext(key).split('--', 1)[1] if key else key
	datalad/plugin/addurls.py:from datalad.support.path import split_ext
	datalad/plugin/addurls.py:    root, ext = split_ext(filename)
	datalad/support/path.py:def split_ext(filename):
	datalad/support/path.py:    >>> from datalad.plugin.addurls import split_ext
	datalad/support/path.py:    >>> split_ext("filename.py")
	datalad/support/path.py:    >>> split_ext("filename.tar.gz")
	datalad/support/path.py:    >>> split_ext("filename.above4chars.ext")

Yet to see with a fresh datalad/git-annex build of tomorrow if anything
got broken, before we introduce this "fix".

Dear @jwodder, please read the haskell diff of [8.20210803-2-g899983058](https://git.kitenet.net/index.cgi/git-annex.git/commit/?id=899983058f77fbeb67a2ac04016ea26fec5468fa) and let me know if the logic I added in Python corresponds (I with there were more tests in annex, which I could have read ;-))

### DIFF
--- a/datalad/support/path.py
+++ b/datalad/support/path.py
@@ -104,10 +104,21 @@ def split_ext(filename):
     >>> split_ext("filename.above4chars.ext")
     ('filename.above4chars', '.ext')
     """
+    # needs to be delayed due to circular
+    # . > config -> cmd -> path -> exernal_version -> log -> ansi_colors -> .
+    from .external_versions import external_versions
+
     parts = filename.split(".")
     if len(parts) == 1:
         return filename, ""
-
+    # stay in sync with annex'es treatment of .dot files to not consider them
+    # having only an extension, and thus including in the key name for E backends
+    # Change in behavior introduced in 8.20210803-2-g899983058
+    if (
+        external_versions['cmd:annex'] > '8.20210803' and
+        len(parts) == 2 and not parts[0]
+    ):
+        return filename, ""
     tail = list(dropwhile(lambda x: len(x) < 5,
                           reversed(parts[1:])))
 
--- a/datalad/support/tests/test_path.py
+++ b/datalad/support/tests/test_path.py
@@ -8,6 +8,7 @@
 # ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
 
 import os
+from ..external_versions import external_versions
 from ..path import (
     abspath,
     curdir,
@@ -57,6 +58,13 @@ def test_split_ext():
     eq_(split_ext("file.a.b.ccccc.d"), ("file.a.b.ccccc", ".d"))
 
     eq_(split_ext("file.a.b..c"), ("file", ".a.b..c"))
+    if external_versions['cmd:annex'] > '8.20210803':
+        eq_(split_ext(".dot"), (".dot", ""))
+        eq_(split_ext(".dot.ext"), (".dot", ".ext"))
+    else:
+        eq_(split_ext(".dot"), ("", ".dot"))
+        eq_(split_ext(".dot.ext"), ("", ".dot.ext"))
+
 
 
 def test_get_parent_paths():


jwodder:
@yarikoptic Is `datalad.support.path.split_ext()` intended to match the behavior of `selectExtension` in `Backend.Utilities`?  Because it doesn't.  If it's not supposed to match, exactly what is your change trying to do?

jwodder:
@yarikoptic Here's my translation of `selectExtension` into Python:

```python
def select_extension(f, maxlen = 4):
    if PATCHED_GIT_ANNEX:
        f = os.path.basename(f).lstrip(".")
    es = f.partition(".")[2].split(".")
    es.reverse()
    es = itertools.takewhile(lambda e: len(e) < maxlen, es)
    es = filter(lambda e: all(map(validInExtension, e)), es)
    es = list(es)[:2]
    es.reverse()
    es = [e for e in es if e]
    if es:
        return ".".join([""] + es)
    else:
        return ""

def validInExtension(c):
    return c.isalnum() or c == "." or ord(c) > 127
```

yarikoptic:
wow, thanks for the detailed cross-language translation -- yet to digest in full, it seems we might want to add additional filtering on which characters could be in an extension etc.  

>  exactly what is your change trying to do?

to reflect the change in "python", couldn't read
```
+	-- Avoid treating a file ".foo" as having its whole name as an
+	-- extension.
+	f' = S.dropWhile (== fromIntegral (ord '.')) (P.takeFileName f)
```
to comprehend -- from your python code - is it corresponding to `os.path.basename(f).lstrip(".")`? (now that you did it -- I kinda can read that haskell ;))

> Note that this technically differs from the Haskell code, as it operates on strings while the Haskell version operates on bytes.

so it would process each byte of eg. utf-8 encoded filename/extension separately??

jwodder:
@yarikoptic Yes, the `if PATCHED_GIT_ANNEX:` block is for applying the effects of the git-annex patch.

> so it would process each byte of eg. utf-8 encoded filename/extension separately??

I suspect the file system encoding would be used, not necessarily UTF-8.

yarikoptic:
> > so it would process each byte of eg. utf-8 encoded filename/extension separately??
> 
> I suspect the file system encoding would be used, not necessarily UTF-8.

sure, I noted `utf-8` just as an example "variable width" (i.e. potentially multi-byte for some characters) encoding... so you think Haskell code analyzes filenames per-byte (as encoded) instead of per-character?

yarikoptic:
anyways -- my logic is probably too far from just an initial stripping of `.`... might adjust or even take @jwodder 's version for this function

jwodder:
@yarikoptic Yes, the Haskell code operates on `ByteString`s instead of `String`s.

yarikoptic:
Dear @jwodder please take over this PR and finish it up to match git-annex'es current logic (I don't think it is worth to make it annex version dependent really), may be even make it operate on bytes instead of string and then decode upon return?

jwodder:
@yarikoptic So is `split_ext()` supposed to match `Backend.Utilities.selectExtension` (which the patch in the OP modified) or `Utility.Path.splitShortExtensions` (which the docstring for `split_ext()` says it's based on)?

yarikoptic:
d'oh -- so git-annex has/uses two ways to split the extension?  I wonder why...  We need the one which is used for key generation. may be @joeyh could clarify which one is used for that purpose?

jwodder:
@yarikoptic Side note: If we end up reimplementing `split_ext()` in terms of `selectExtension` instead of `splitShortExtensions`, we're going to run into ambiguities regarding what the "root" part of the return value of `split_ext()` should be, due to empty extensions being filtered out of `selectExtension`'s return value in cases like:

```
>>> select_extension("filename.txt..png")
'.png'
>>> select_extension("filename.txt.png.")
'.png'
```

jwodder:
@yarikoptic I don't mean the extension.  Currently, `split_ext()` returns a `(root, extension)` tuple, and the root is used [here](https://github.com/datalad/datalad/blob/5adfde7818a9582cd12c2d25fe6f008132de1b2f/datalad/local/addurls.py#L508).  The current definition of the "root" is the part before the extension, but this becomes ambiguous for files with empty extensions; how should that be handled?  (Note that git-annex doesn't seem to bother with calculating a "root", so we can't just copy what it does.)

mih:
It seems this month's update of this PR hasn't happened yet! I am dying to read how the plot evolves ;-)

mih:
This PR seems abandoned. Given the git-annex change is 6 months old, and nothing broke without this change, maybe we don't need it?

jwodder:
For the record, if this is ever re-opened/revisited, my implementation of git-annex's `selectExtension` in Python was as follows:

```python
import os
import os.path

def select_extension(f, maxlen = 4):
    """
    Extract a file path's extensions, selected using the same logic as
    git-annex's ``Backend.Utilities.selectExtension`` as of v8.20210903

    Parameters
    ----------
    f: str, bytes, or os.PathLike
      Filepath from which to extract the file extensions
    maxlen: int, optional
      Maximum length (in bytes) of individual file extensions to consider; if
      any extension is longer than this many bytes, only extensions after it
      will be considered; defaults to 4

    Returns
    -------
    a `str` containing zero, one, or two trailing file extensions of `f`, each
    preceded by a period

    Examples
    --------
    >>> select_extension("filename.py")
    '.py'
    >>> select_extension("filename.tar.gz")
    '.tar.gz'
    >>> select_extension("filename.above4chars.ext")
    '.ext'
    >>> select_extension(".ini")
    ''
    >>> select_extension(".ini.txt")
    '.txt'
    >>> select_extension("filename.txt..png")
    '.png'
    >>> select_extension("filename.")
    ''
    >>> select_extension("filename.txt.png.")
    '.png'
    >>> select_extension("filename.txt.png.gz")
    '.png.gz'
    """

    def validInExtension(b: int) -> bool:
        return bytes([b]).isalnum() or b == ord(".") or b > 127

    f = os.path.basename(os.fsencode(f)).lstrip(b".")
    es = []
    for e in reversed(f.partition(b".")[2].split(b".")):
        if len(e) > maxlen:
            break
        if all(map(validInExtension, e)):
            es.append(e)
            if len(es) == 2:
                break
    es = [e for e in reversed(es) if e]
    if es:
        return os.fsdecode(b".".join([b""] + es))
    else:
        return ""
```

Issue #5832, opened 2021-07-28
(Re)design of `drop`


mih:
Companion of #5842 (to be discussed in conjunction)

Context: This is about approaching an improved implementation of the `drop` command, not in isolation, but in the context of a reconsideration of the `drop/uninstall/remove` combo, with the partial goal to replace the old `remove` entirely, and with that make progress towards #3368

This is an opportunity to reconsider the role of `drop` in the API, and amend it with the missing functionality we have identified (issues are linked below). This draft description is not about making a contrast to what the present `drop` is doing (there are massive overlaps), but about a comprehensible concept of what `drop` does, and does not. The initial design draft is the result of a discussion of @adswa and @mih.

### Basic concept

`drop` is the antagonist of `get`. Whatever a `drop` can do, should be undoable by a subsequent `get` (given unchanged remote availability).

Like `get`, `drop` primarily operates on a mandatory path specification (to discover relevant files and sudatasets to operate on).

`drop` has `--type` parameter that serves as an extensible "mode-switch" to cover all relevant scenarios, like 'drop all file content in the work-tree' (e.g. `--type files`, default, #5858), 'drop all keys from any branch' (i.e. `--type allkeys`, #2328), but also '"drop" AKA uninstall entire subdataset hierarchies' (e.g. `--type all`), or drop preferred content (`--type preferred-content`, #3122).

`drop` prevents data loss by default (#4750). Like `get` it features a `--reckless` "mode-switch" to disable some or all potentially slow safety mechnism, i.e. 'key available in sufficient number of other remotes', 'main or all branches pushed to remote(s)' (#1142), 'only check availability of keys associated with the worktree, but not other branches'. "Reckless operation" can be automatic, when following a reckless `get` (#4744).

`drop` properly manages annex lifetime information, e.g. by announcing an annex as `dead` on removal of a repository (#3887).

Like `get`, drop supports parallization #1953 

`datalad drop` is not intended to be a comprehensive frontend to `git annex drop` (e.g. limited support for e.g. #1482 outside standard use cases like #2328).

Note: It is understood that the current `uninstall` command is largely or completely made obsolete by this `drop` concept.

Given the development in #5842  towards the complete obsolescence of `remove` it becomes necessary to import one of its proposed features:

`drop` should be able to recognize a botched attempt to delete a dataset with a plain rm -rf, and act on it in a meaningful way, even if it is just hinting at chmod + rm -rf.

### Use cases

[...]

PR #5867, opened 2021-08-04
Design proposal for dropping content

mih:
This is about approaching an improved implementation of the drop
command, not in isolation, but in the context of a reconsideration of
the drop/uninstall/remove combo, with the partial goal to replace the
old remove entirely, and with that make progress towards #3368

This is an opportunity to reconsider the role of drop in the API, and
amend it with the missing functionality we have identified (issues are
linked below). This draft description is not about making a contrast to
what the present drop is doing (there are massive overlaps), but about a
comprehensible concept of what drop does, and does not. The initial
design draft is the result of a discussion of @adswa and @mih.

A discussion of this already started in #5832. In favor of a more formal
discussion and, importantly, adjustment process of this proposal, it
will be resubmitted as a PR, not as an issue.

I have made an attempt to ease precise referencing particular aspects via exhaustive numbering. If possible, use the line-by-line commenting feature of PRs.


mih:
This PR failed to ignite a discussion beyond #5832. I will implement the rename `--type` -> `--what` and then merge/start the implementation.

mih:
Test failures are from `master` itself, and hopefully fixed by #5897

Issue #5862, opened 2021-08-04
Remove "API principles" section from docs


mih:
This is a very narrow set of "principles", most information on `install`. It also has a description of the `--dataset` argument, which is incorrect.

adswa:
This has been fixed by  #5868. Please reopen if I am mistaken.

Issue #3905, opened 2019-12-07
Establish top level behavior rules/principles of DataLad


yarikoptic:
Moved from https://github.com/datalad/datalad/issues/3759#issuecomment-559908687 which concerns change in `get` and `install` behavior in pre-0.12 release.

I am proposing us to discuss a possible list of rules/principles we could document as to describe how DataLad operates given a specific CWD, optional --dataset, and provided paths.  ATM there is no  such description, introduced changes break previous behaviors (e.g. as in #3759) under the premise of removing ambiguity, but to me everything is ambiguous since we have no clear description of expected behaviors like below.  So we [ourselves even get confused](https://github.com/datalad/datalad/issues/3759#issuecomment-562631994) either such a list of steps corresponds to current or proposed behavior. 

Here is the list of rules/steps I proposed.  I would welcome someone to add a changed set which corresponds to current (as of sample commands X, Y, Z) behaviors:

1. If no working dataset is explicitly specified (via `dataset` argument), a dataset current directory belongs to is taken as the "working dataset".  (note: If there is no "working dataset", that is ok (rules `4., 5.` below have "if"))
2. Any given command first operates on the path(s) within the dataset where path(s) belong to (clarification: a (sub)dataset path causes operation within that (sub)dataset; even `datalad install .` is ok).
3. `--recursive` flag could be used (where applicable) to make command operate within subdatasets of the provided path(s). 
4. If there is a "working dataset", changes on the paths performed in the sub-datasets of the "working dataset" are also saved in any sub-dataset on the path from the "working dataset" to the dataset specified path(s) belongs to (old behavior I think)
5. For the paths outside of the "working dataset" (if any), changes to their datasets are not saved in their possible super-datasets (old behavior)
6. Merely a note: some commands (e.g. `get` and `install`) might require necessary sub-datasets first being `install`ed to act on the path(s) specified on the command line.  Such sub-datasets will be installed regardless of relation to the working directory dataset (old bevavior, nothing really special here). `create` might require to create a directory in the specified path to act on.

mih:
This has failed to ignite the desired discussion. In the meantime the behavior of the joint set of commands has further developed. I think, at present, the most useful for to pick such a discussion up -- if desired -- is a PR with a design document.

Will close this issue for now.

PR #5868, opened 2021-08-04
DOC: Design document for the `dataset` argument

mih:
This is no revolutionary re-design. It merely moves existing information
into a standalone document -- and strips incorrect information.

This also works towards datalad/datalad#5862

### DIFF
--- a/docs/source/design/dataset_argument.rst
+++ b/docs/source/design/dataset_argument.rst
@@ -12,8 +12,9 @@
    This specification describes the current implementation.
 
 All commands which operate on datasets have a ``dataset`` argument (``-d`` or
-``--dataset`` for the command line API) to identify a single dataset as the
-context of an operation.
+``--dataset`` for the :term:`CLI`) to identify a single dataset as the
+If ``--dataset`` argument is not provided, the context of an operation is operation-specific (e.g., for `clone` command -- the dataset which is being cloned), but typically - a dataset which current working directory belongs to.
+In the latter case, if operation (e.g., `get`) does not find a dataset in current working directory, operation fails with an ``NoDatasetFound`` error.
 
 
 Impact on relative path resolution
--- a/docs/source/design/dataset_argument.rst
+++ b/docs/source/design/dataset_argument.rst
@@ -7,6 +7,10 @@
 ``dataset`` argument
 ********************
 
+.. topic:: Specification scope and status
+
+   This specification describes the current implementation.
+
 All commands which operate on datasets have a ``dataset`` argument (``-d`` or
 ``--dataset`` for the command line API) to identify a single dataset as the
 context of an operation.
--- a/docs/source/design/dataset_argument.rst
+++ b/docs/source/design/dataset_argument.rst
@@ -13,7 +13,10 @@
 
 All commands which operate on datasets have a ``dataset`` argument (``-d`` or
 ``--dataset`` for the :term:`CLI`) to identify a single dataset as the
-If ``--dataset`` argument is not provided, the context of an operation is operation-specific (e.g., for `clone` command -- the dataset which is being cloned), but typically - a dataset which current working directory belongs to.
+context of an operation.
+If ``--dataset`` argument is not provided, the context of an operation is command-specific.
+For example, `clone` command will consider the :term:`dataset` which is being cloned to be the context.
+But typically, a dataset which current working directory belongs to is the context of an operation.
 In the latter case, if operation (e.g., `get`) does not find a dataset in current working directory, operation fails with an ``NoDatasetFound`` error.
 
 
--- a/docs/source/design/dataset_argument.rst
+++ b/docs/source/design/dataset_argument.rst
@@ -1,0 +1,58 @@
+.. -*- mode: rst -*-
+.. vi: set ft=rst sts=4 ts=4 sw=4 et tw=79:
+
+.. _chap_design_dataset_argument:
+
+********************
+``dataset`` argument
+********************
+
+All commands which operate on datasets have a ``dataset`` argument (``-d`` or
+``--dataset`` for the command line API) to identify a single dataset as the
+context of an operation.
+
+
+Impact on relative path resolution
+==================================
+
+With one exception, the nature of a provided ``dataset`` argument does **not**
+impact the interpretation of relative paths. Relative paths are always considered
+to be relative to the process working directory.
+
+The one exception to this rule is passing a ``Dataset`` object instance as
+``dataset`` argument value in the Python API. In this, and only this, case, a
+relative path is interpreted as relative to the root of the respective dataset.
+
+
+Special values
+==============
+
+There are some pre-defined "shortcut" values for dataset arguments:
+
+``^``
+   Represents to the topmost superdataset that contains the dataset the current
+   directory is part of.
+``^.``
+   Represents the root directory of the dataset the current directory is part of.
+``///``
+   Represents the "default" dataset located under `$HOME/datalad/`.
+
+
+Use cases
+=========
+
+Save modification in superdataset hierarchy
+-------------------------------------------
+
+Sometimes it is convenient to work only in the context of a subdataset.
+Executing a ``datalad save <subdataset content>`` will record changes to the
+subdataset, but will leave existing superdatasets dirty, as the subdataset
+state change will not be saved there. Using the ``dataset`` argument it is
+possible to redefine the scope of the save operation. For example::
+
+  datalad save -d^ <subdataset content>
+
+will perform the exact same save operation in the subdataset, but additionally
+save all subdataset state changes in all superdatasets until the root of a
+dataset hierarchy. Except for the specification of the dataset scope there is
+no need to adjust path arguments or change the working directory.
--- a/docs/source/basics.rst
+++ b/docs/source/basics.rst
@@ -162,31 +162,6 @@ Finally, DataLad supports SSH login style resource identifiers, such as ``me@loc
 
 ``datalad install me@localhost:/path``
 
-`--dataset` argument
---------------------
-
-All commands which operate with/on datasets (practically all commands) have a
-``dataset`` argument (``-d`` or ``--dataset`` for the command line API) which takes a
-path to the dataset that the command should operate on. If a dataset is
-identified this way then any relative path that is provided as an argument to
-the command will be interpreted as being relative to the topmost directory of that
-dataset.  If no dataset argument is provided, relative paths are considered to be
-relative to the current directory.
-
-There are also some useful pre-defined "shortcut" values for dataset arguments:
-
-``///``
-   refers to the "default" dataset located under `$HOME/datalad/`.
-   So running ``datalad install -d/// crcns`` will install the ``crcns`` subdataset
-   under ``$HOME/datalad/crcns``.  This is the same as running
-   ``datalad install $HOME/datalad/crcns``.
-``^``
-   topmost superdataset containing the dataset the current directory is part of.
-   For example, if you are in ``$HOME/datalad/openfmri/ds000001/sub-01`` and want
-   to search metadata of the entire superdataset you are under (in this case
-   ``///``), run ``datalad search -d^ [something to search]``.
-``^.``
-   the dataset the current directory is part of.
 
 Commands `install` vs `get`
 ---------------------------
--- a/docs/source/design/index.rst
+++ b/docs/source/design/index.rst
@@ -13,4 +13,5 @@ subsystems in DataLad.
 .. toctree::
    :maxdepth: 2
 
+   dataset_argument
    miscpatterns
--- a/docs/source/glossary.rst
+++ b/docs/source/glossary.rst
@@ -44,6 +44,10 @@ them to the corresponding Git_/git-annex_ concepts.
     operations significantly and impair handling of such repositories in
     general).
 
+  CLI
+    A `Command Line Interface`_. Could be used interactively by executing
+    commands in a `shell`_, or as a programmable API for shell scripts.
+
   DataLad extension
     A Python package, developed outside of the core DataLad codebase, which
     (when installed) typically either provides additional top level `datalad`
@@ -54,3 +58,5 @@ them to the corresponding Git_/git-annex_ concepts.
 
 .. _Git: https://git-scm.com
 .. _Git-annex: http://git-annex.branchable.com
+.. _`Command Line Interface`: https://en.wikipedia.org/wiki/Command-line_interface
+.. _shell: https://en.wikipedia.org/wiki/Shell_(computing)
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -2,7 +2,7 @@ DataLad |---| data management and publication multitool
 *******************************************************
 
 Welcome to DataLad's **technical documentation**. Information here is targeting
-software developers and is focused on the Python and command line APIs, as well
+software developers and is focused on the Python API and :term:`CLI`, as well
 as software design, employed technologies, and key features.  Comprehensive
 **user documentation** with information on installation, basic operation,
 support, and (advanced) use case descriptions is available in the `DataLad


docs/source/design/dataset_argument.rst
```diff
@@ -0,0 +1,58 @@
...
+
+.. _chap_design_dataset_argument:
+
+********************
+``dataset`` argument
+********************
+
+All commands which operate on datasets have a ``dataset`` argument (``-d`` or
+``--dataset`` for the command line API) to identify a single dataset as the
+context of an operation.
```
bpoldrack, line 12: I think this section lacks a hint on what is considered the context, if no `--dataset` argument is given. And that probably is the part that makes it somewhat confusing. Context ds is either `-d`, derived from PWD or derived from a given `path` argument.

docs/source/design/dataset_argument.rst
```diff
@@ -0,0 +1,58 @@
...
+
+Use cases
+=========
+
+Save modification in superdataset hierarchy
+-------------------------------------------
+
+Sometimes it is convenient to work only in the context of a subdataset.
+Executing a ``datalad save <subdataset content>`` will record changes to the
+subdataset, but will leave existing superdatasets dirty, as the subdataset
```
bpoldrack, line 49: While I generally agree (and see the technical reason, of course), I wonder whether this approach really is, what we should do. "Sometimes it is convenient to work only in the context of a subdataset." True. And: Given the aim for a "mono-repo experience", it makes sense to *only sometimes* "work only in the context of a subdataset". However, since the context by default is derived from a path argument (as shown here in `datalad save <subdataset content>`) and pretty much all commands do operate on paths, the mono-repo view is *not* the default. Quite the opposite: In order to know what a "simple" (as in "no options specified") invocation like this will do, I need to be aware of the subdataset boundaries. Not sure about a solution, but for now I wonder whether we agree there's an issue. May be it's just me ...

mih: I am not sure what you are asking for? Remove the example? Change how `save` works?

bpoldrack: As far as documenting the status quo goes the only asking is my comment further up - pointing out how PWD and a possible `path` argument contribute to determining the context.

Other than that: Wondering whether we are satisfied with how `-d` works (it's not about `save`). The suggestion would be to take `path` out of the equation and make the context always `-d` with default `.`. This would be a breaking change - yes (But easy in implementation, I think). ATM I think it would be much more intuitive, though.

mih:
I would suggest we make this about documenting the status quo first. And postpone further proposal till after they can be described as a change from that.

docs/source/design/dataset_argument.rst
mih: I agree that this would be nice. But ultimately this section does not have such hint, because I am not aware of a general valid approach. The best I could come up with is "given no `dataset` declares no context, and all bets are off" i.e. the command will do whatever seems to make sense. But this is not a "design" and unknown whether actually true. Even your statement is speculation AFAICS. Let's start simple.

mih:
I added a statement on the scope and status.

docs/source/design/dataset_argument.rst
```diff
@@ -0,0 +1,62 @@
...
+********************
+``dataset`` argument
+********************
+
+.. topic:: Specification scope and status
+
+   This specification describes the current implementation.
+
+All commands which operate on datasets have a ``dataset`` argument (``-d`` or
+``--dataset`` for the command line API) to identify a single dataset as the
```
yarikoptic, line 15:
```suggestion
``--dataset`` for the :term:`CLI`) to identify a single dataset as the
```
and add CLI to glossary? After all Command Line Interface is not necessarily used as an API but rather as an "interactive" interface.

docs/source/design/dataset_argument.rst
yarikoptic: Reminded me about https://github.com/datalad/datalad/issues/3905 (refers to relevant issues there as well) which unfortunately didn't get any traction, but I see this PR as a step toward addressing it.

I echo @bpoldrack 's comment that  "here" might be the correct place to state what the "context of an operation" really means, and how does it "differ" whenever `--dataset` is not specified.  E.g. `save` would add/save present but not yet "known" subdataset even without `-d` but `clone` would not unless `-d` is given.

<details>
<summary>example</summary> 

```shell
# sorry -- I start "dirty"
(git-annex)lena:/tmp/testds[master]git-annex
$> datalad status      
 modified: hcp-100206-git-annex2 (dataset)

$> datalad clone ///
install(ok): /tmp/testds/datasets.datalad.org (dataset)                                                                                                                                                              

$> datalad status   
untracked: datasets.datalad.org (directory)
 modified: hcp-100206-git-annex2 (dataset)

$> datalad clone -d. /// another.datasets.datalad.org
install(ok): another.datasets.datalad.org (dataset)                                                                                                                                                                  
add(ok): another.datasets.datalad.org (file)                                                                                                                                                                         
add(ok): .gitmodules (file)                                                                                                                                                                                          
save(ok): . (dataset)                                                                                                                                                                                                
action summary:                                                                                                                                                                                                      
  add (ok: 2)
  install (ok: 1)
  save (ok: 1)
datalad clone -d. /// another.datasets.datalad.org  4.02s user 1.93s system 38% cpu 15.410 total

$> datalad status                                    
untracked: datasets.datalad.org (directory)
 modified: hcp-100206-git-annex2 (dataset)

$> datalad --version
datalad 0.14.6
```
</details>
and I feel that some other operations also might still have this duality in their operation depending on either `--dataset` is provided or not, while often still operating on a dataset deduced to be the "context" of operation. E.g. above with `clone` the "context" is the cloned dataset, not dataset we are currently in.

So may be smth like

```suggestion
If ``--dataset`` argument is not provided, context of an operation is operation-specific (e.g., for `clone` command -- the dataset which is being cloned), but typically - a dataset which current working directory belongs to.
In the latter case, if operation (e.g., `get`) does not find a dataset in current working directory, operation fails with an ``NoDatasetFound`` error.
```

If we see it needing to be formalized more later, we could get a dedicated section on that, but IMHO we should not just completely disregard situation whenever `--dataset` is **not** provided here.

docs/source/design/dataset_argument.rst
```diff
@@ -0,0 +1,62 @@
...
+Impact on relative path resolution
+==================================
+
+With one exception, the nature of a provided ``dataset`` argument does **not**
+impact the interpretation of relative paths. Relative paths are always considered
+to be relative to the process working directory.
+
+The one exception to this rule is passing a ``Dataset`` object instance as
+``dataset`` argument value in the Python API. In this, and only this, case, a
+relative path is interpreted as relative to the root of the respective dataset.
```
yarikoptic, line 28: note: I always thought about it "in reverse", that CLI was a special case, not Python interface, but then realized that unbound invocations (i.e., like `datalad.api.clone`, without `dataset` being provided) also present in Python API.  So this formulation is probably the most succinct indeed to reflect the current state of the affairs.  But with "explicit better than implicit" (and the fact that `Dataset().<operation>` is passing `dataset` argument to operation is IMHO implicit here) I wonder if we better lay it out explicitly in addition (or instead) here:

```suggestion
relative path is interpreted as relative to the root of the respective dataset.

More specifically:

- relative paths in CLI are treated relative to :term:`CWD`
- relative paths in Python API will be treated 
  - to :term:`CWD` in invocations like ``datalad.api.<operation>(<arguments>)`` if ``<arguments>`` does not include ``dataset``
  - to context dataset root if 
    - aforementioned ``<operation>``invocation includes ``dataset`` argument
    - ``<operation>`` is performed as a `Dataset` object instance bound method, i.e. as ``Dataset(...).<operation>(<arguments>)``. **Note:** specification of ``dataset`` among ``<arguments>`` in such invocation will result in error.
```

docs/source/design/dataset_argument.rst
```diff
@@ -0,0 +1,62 @@
...
+Use cases
+=========
+
+Save modification in superdataset hierarchy
+-------------------------------------------
+
+Sometimes it is convenient to work only in the context of a subdataset.
+Executing a ``datalad save <subdataset content>`` will record changes to the
+subdataset, but will leave existing superdatasets dirty, as the subdataset
+state change will not be saved there. Using the ``dataset`` argument it is
```
yarikoptic, line 54: echoing resolved discussion with @bpoldrack - I think the "context of operation" is still to vague, and there it seems to suggest something like 

```rst
Some (or all?) operations can "transfer" the context of operation to the subdatasets provided in the path, if no `--dataset` argument is provided.
```
which would possibly provide explanation on how come `save <subdataset content>` does not operate in the context of current dataset (which it might if one of the paths is within this dataset) but "in the context of a subdataset".

Actual reality is actually even a bit more tricky, since 

<details>
<summary>we do also operate in the context of CWD dataset - since we do `save` the change saved within subdataset </summary> 

```shell
(git-annex)lena:/tmp/testds[master]git
$> datalad status
untracked: datasets.datalad.org (directory)
 modified: hcp-100206-git-annex2 (dataset)
(dev3) 1 26057.....................................:Thu 05 Aug 2021 10:43:32 AM EDT:.
(git-annex)lena:/tmp/testds[master]git
$> touch sub1/newfile newfilehere1 newfilehere2
(dev3) 1 26058.....................................:Thu 05 Aug 2021 10:43:53 AM EDT:.
(git-annex)lena:/tmp/testds[master]git
$> datalad status
untracked: datasets.datalad.org (directory)
untracked: newfilehere1 (file)
untracked: newfilehere2 (file)
 modified: hcp-100206-git-annex2 (dataset)
 modified: sub1 (dataset)
(dev3) 1 26059.....................................:Thu 05 Aug 2021 10:43:59 AM EDT:.
(git-annex)lena:/tmp/testds[master]git
$> datalad save sub1/newfile newfilehere1
add(ok): newfile (file)                                                                                                                                                                                              
save(ok): sub1 (dataset)                                                                                                                                                                                             
add(ok): sub1 (file)                                                                                                                                                                                                 
add(ok): newfilehere1 (file)                                                                                                                                                                                         
save(ok): . (dataset)                                                                                                                                                                                                
action summary:                                                                                                                                                                                                      
  add (ok: 3)
  save (ok: 2)
(dev3) 1 26060.....................................:Thu 05 Aug 2021 10:44:11 AM EDT:.
(git-annex)lena:/tmp/testds[master]git
$> datalad status
untracked: datasets.datalad.org (directory)
untracked: newfilehere2 (file)
 modified: hcp-100206-git-annex2 (dataset)

(git-annex)lena:/tmp/testds[master]git
$> touch sub1/newfile2                          
(dev3) 1 26062.....................................:Thu 05 Aug 2021 10:45:06 AM EDT:.
(git-annex)lena:/tmp/testds[master]git
$> datalad save sub1/newfile2
add(ok): newfile2 (file)                                                                                                                                                                                             
save(ok): sub1 (dataset)                                                                                                                                                                                             
action summary:                                                                                                                                                                                                      
  add (ok: 1)
  save (ok: 1)

$> datalad --version
datalad 0.14.7+1.g721315d5d.dirty

```
</details>

so may be this statement is not correct altogether?

yarikoptic (changes requested):
Left some comments/suggestions.  It seems that the late `save` example doesn't live up to what it describes...

docs/source/design/dataset_argument.rst
mih: TBH I don't like that. I is probably saying the same thing, but even on third pass I am not confident that it is, or whether it add additional nuances.

mih:
I think I will detach from this PR. Please adopt it.

My intention was to fix the factually incorrect description in the docs, while moving the rest into the design format idea to see how this works. @bpoldrack @yarikoptic seem to have much more ambition to take this to higher grounds. I have no plans to fill in the gaps that exist in the document, because I don't know about these aspects of the "design".

Personally, I would very much prefer to set a really low bar for improving the docs. There is much false information in them and the structure is not suitable for small additions (just search the issues for "I wish we could document this"). Sadly I only have the capacity for small steps.

docs/source/design/dataset_argument.rst
yarikoptic: ok, let's not bother being explicit

docs/source/design/dataset_argument.rst
yarikoptic: I think I just got mislead by `save(ok): sub1 (dataset)` assuming that it was to inform that we saved updated state of `sub1` within current dataset

<details>
<summary>but it is not the case</summary> 

```shell
(git-annex)lena:/tmp/testds[master]git-annex
$> git status       
On branch master
	modified:   sub1 (new commits)

(dev3) 1 26407 [1].....................................:Fri 06 Aug 2021 09:25:21 AM EDT:.
(git-annex)lena:/tmp/testds[master]git-annex
$> datalad save sub1/
(dev3) 1 26408 [1].....................................:Fri 06 Aug 2021 09:25:42 AM EDT:.
(git-annex)lena:/tmp/testds[master]git-annex
$> touch sub1/newfile3
(dev3) 1 26409 [1].....................................:Fri 06 Aug 2021 09:25:48 AM EDT:.
(git-annex)lena:/tmp/testds[master]git-annex
$> datalad save sub1  
add(ok): newfile3 (file)                                                                                  
save(ok): sub1 (dataset)                                                                                  
action summary:                                                                                           
  add (ok: 1)
  save (ok: 1)
(dev3) 1 26410 [1].....................................:Fri 06 Aug 2021 09:25:55 AM EDT:.
(git-annex)lena:/tmp/testds[master]git-annex
$> git status         
On branch master
	modified:   sub1 (new commits)

(dev3) 1 26411 [1].....................................:Fri 06 Aug 2021 09:26:21 AM EDT:.
(git-annex)lena:/tmp/testds[master]git-annex
$> git show | head
commit d7e5959365f6bc575a89e1a49d8dc1d504680c6f
Date:   Thu Aug 5 10:44:11 2021 -0400

    [DATALAD] Recorded changes

diff --git a/newfilehere1 b/newfilehere1
new file mode 120000
index 0000000..ce38032
--- /dev/null
(dev3) 1 26412 [1].....................................:Fri 06 Aug 2021 09:26:41 AM EDT:.
(git-annex)lena:/tmp/testds[master]git-annex
$> touch sub1/newfile4                                             
(dev3) 1 26413 [1].....................................:Fri 06 Aug 2021 09:28:09 AM EDT:.
(git-annex)lena:/tmp/testds[master]git-annex
$> datalad save sub1/      
add(ok): newfile4 (file)                                                                                  
save(ok): sub1 (dataset)                                                                                  
action summary:                                                                                           
  add (ok: 1)
  save (ok: 1)
(dev3) 1 26414 [1].....................................:Fri 06 Aug 2021 09:28:28 AM EDT:.
(git-annex)lena:/tmp/testds[master]git-annex
$> touch sub1/newfile5
(dev3) 1 26415 [1].....................................:Fri 06 Aug 2021 09:28:31 AM EDT:.
(git-annex)lena:/tmp/testds[master]git-annex
$> datalad save sub1/newfile5
add(ok): newfile5 (file)                                                                                  
save(ok): sub1 (dataset)                                                                                  
action summary:                                                                                           
  add (ok: 1)
  save (ok: 1)
(dev3) 1 26416 [1].....................................:Fri 06 Aug 2021 09:28:38 AM EDT:.
(git-annex)lena:/tmp/testds[master]git-annex
$> git status
On branch master
	modified:   sub1 (new commits)

```
</details>

so the statement was correct.

yarikoptic: I will mark it as resolved and will not bother about clarifying "context" here

yarikoptic:
> I think I will detach from this PR. Please adopt it.

with your blessing, I will try with only minimal changes following some of the comments/suggestions I left

docs/source/design/dataset_argument.rst
```diff
@@ -0,0 +1,63 @@
...
+********************
+
+.. topic:: Specification scope and status
+
+   This specification describes the current implementation.
+
+All commands which operate on datasets have a ``dataset`` argument (``-d`` or
+``--dataset`` for the :term:`CLI`) to identify a single dataset as the
+If ``--dataset`` argument is not provided, the context of an operation is operation-specific (e.g., for `clone` command -- the dataset which is being cloned), but typically - a dataset which current working directory belongs to.
+In the latter case, if operation (e.g., `get`) does not find a dataset in current working directory, operation fails with an ``NoDatasetFound`` error.
```
yarikoptic, line 17: if we decide to not "migrate" toward a sentence per line in written text, I will wrap them.
WDYT @mih @bpoldrack ?

docs/source/design/dataset_argument.rst
```diff
@@ -0,0 +1,63 @@
...
+``dataset`` argument
+********************
+
+.. topic:: Specification scope and status
+
+   This specification describes the current implementation.
+
+All commands which operate on datasets have a ``dataset`` argument (``-d`` or
+``--dataset`` for the :term:`CLI`) to identify a single dataset as the
+If ``--dataset`` argument is not provided, the context of an operation is operation-specific (e.g., for `clone` command -- the dataset which is being cloned), but typically - a dataset which current working directory belongs to.
```
yarikoptic, line 16: and crap -- I have missed that my suggestion killed the ending of prior sentence, sorry @mih.
In either case, at least this one better be split into multiple sentences, e.g.
```suggestion
context of an operation.
If ``--dataset`` argument is not provided, the context of an operation is operation-specific.
For example, `clone` command will consider the :term:`dataset` which is being cloned to be the context.
But typically, a dataset which current working directory belongs to is the context of an operation.
```

mih: I have adjusted your suggestion: `s/operation-specific/command-specific/` will commit now.

mih:
I resolved the conflict, and commited @yarikoptic's suggestions. Merging now.

PR #5875, opened 2021-08-05
ENH(UX): warn user if keyring returned a "null" keyring

yarikoptic:
Happened to me: a number of days ago I had to workaround a pip issue https://github.com/pypa/pip/issues/7883 for which Q&D workaround was just
to export PYTHON_KEYRING_BACKEND=keyring.backends.null.Keyring .

Then I lived happily ever after (since 1-May?), probably even datalad download-url
was just reusing prior cookies etc.  But while troubleshooting earthdata
data access, I was at WTF stage that datalad kept asking for
credentials, and there were no visual feedback on a reason.

Diff is trivial and IMHO safe, so made it against `maint`

### DIFF
--- a/datalad/support/keyring_.py
+++ b/datalad/support/keyring_.py
@@ -37,7 +37,12 @@ class Keyring(object):
             lgr.debug("Importing keyring")
             import keyring
             self.__keyring = keyring
-
+            the_keyring = keyring.get_keyring()
+            if the_keyring.name.lower().startswith('null '):
+                lgr.warning(
+                    "Keyring module returned '%s', no credentials will be provided",
+                    the_keyring.name
+                )
         return self.__keyring
 
     @classmethod


adswa:
Makes sense to me, but I can't seem to trigger the warning. What would I need to do?

I ran ``export PYTHON_KEYRING_BACKEND=keyring.backends.null.Keyring``, and then I tried to get data from www.github.com/datalad-datasets/human-connectome-project-openaccess with ``datalad -l warning get ...`` - it keeps prompting me for my credentials, but it nevers displays the warning. What am I doing wrong?

yarikoptic:
oh, with `get` I guess we are getting too deep into the sandwich of external remotes, and since `logging` is not going through our `ui.message`, it never gets displayed :-/  We have `.message` and `.error` for UIs but no `.warning`.  I guess I could just make it `ui.message("WARNING: ...")` and then it should be displayed even if deep in the sandwich. WDYT?

adswa:
> edit: but I guess that would be too overfitting for this use-case, which is just minor and not worth it IMHO. We might better see more consistent propagation of any warning or error from logger if we decide to go that route

Yeah, makes sense. The current warning works for the problem at hand :+1:

yarikoptic:
ok, let's proceed then. Thank you @adswa !

Issue #5846, opened 2021-07-29
Authenticated access to http EarthData (ecs.nsidc.org) does not work


yarikoptic:
Fresh user case : https://neurostars.org/t/download-url-from-password-protected-site/19888

<details>
<summary>wget -S call</summary> 

```shell
$> wget -S https://n5eil01u.ecs.nsidc.org/ICEBRIDGE/IDBMG4.004/1993.01.01/BedMachineGreenland-2021-04-20.nc.xml
--2021-07-29 17:20:16--  https://n5eil01u.ecs.nsidc.org/ICEBRIDGE/IDBMG4.004/1993.01.01/BedMachineGreenland-2021-04-20.nc.xml
Resolving n5eil01u.ecs.nsidc.org (n5eil01u.ecs.nsidc.org)... 128.138.97.102
Connecting to n5eil01u.ecs.nsidc.org (n5eil01u.ecs.nsidc.org)|128.138.97.102|:443... connected.
HTTP request sent, awaiting response... 
  HTTP/1.1 302 Found
  Date: Thu, 29 Jul 2021 21:20:17 GMT
  Server: Apache
  Location: https://urs.earthdata.nasa.gov/oauth/authorize?app_type=401&client_id=_JLuwMHxb2xX6NwYTb4dRA&response_type=code&redirect_uri=https%3A%2F%2Fn5eil01u.ecs.nsidc.org%2FOPS%2Fredirect&state=aHR0cDovL241ZWlsMDF1LmVjcy5uc2lkYy5vcmcvSUNFQlJJREdFL0lEQk1HNC4wMDQvMTk5My4wMS4wMS9CZWRNYWNoaW5lR3JlZW5sYW5kLTIwMjEtMDQtMjAubmMueG1s
  Content-Length: 517
  Keep-Alive: timeout=15, max=100
  Connection: Keep-Alive
  Content-Type: text/html; charset=iso-8859-1
[...]

PR #5876, opened 2021-08-05
RF(BF?)+DOC: provide User-Agent to entire session headers + use those if provided

yarikoptic:
BTW -- `@borrowkwargs` worked out pretty nice!

Apparently the "headers" even if provided seems were never used.
Assigning User-Agent to the headers which are then assigned to the
requests Session as soon as it is created then makes User-Agent
used also during authentication requests (_post_credential). Before
User-Agent was not provided during authentication at all, and only
whenever GETing the load/files

Came to the realization of this deficiency while looking at https://github.com/datalad/datalad/issues/5846

### DIFF
--- a/datalad/downloaders/http.py
+++ b/datalad/downloaders/http.py
@@ -517,9 +517,20 @@ class HTTPDownloader(BaseDownloader):
     """
 
     @borrowkwargs(BaseDownloader)
-    def __init__(self, headers={}, **kwargs):
+    def __init__(self, headers=None, **kwargs):
+        """
+
+        Parameters
+        ----------
+        headers: dict, optional
+          Header fields to be provided to the session. Unless User-Agent provided, a custom
+          one, available in `DEFAULT_USER_AGENT` constant of this module will be used.
+        """
         super(HTTPDownloader, self).__init__(**kwargs)
         self._session = None
+        headers = headers.copy() if headers else {}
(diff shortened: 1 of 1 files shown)


datalad/downloaders/http.py
```diff
@@ -517,9 +517,20 @@ class HTTPDownloader(BaseDownloader):
...
+
+        Parameters
+        ----------
+        headers: dict, optional
+          Header fields to be provided to the session. Unless User-Agent provided, a custom
+          one, available in `DEFAULT_USER_AGENT` constant of this module will be used.
+        """
         super(HTTPDownloader, self).__init__(**kwargs)
         self._session = None
+        headers = headers and headers.copy() or {}
```
yarikoptic, line 531: explicit better than obscure, so let's make it
```suggestion
        headers = headers.copy() if headers else {}
```

bpoldrack (approved):
LGTM and agree with the suggestion to make that expression more explicit.

datalad/downloaders/http.py
mih: FTR: I would be fine with both patterns

mih (approved):
LGTM. Thx!

yarikoptic:
Thanks!  rewrote that line locally, merged into maint and pushed locally and then pushed into this branch so PR gets autoclosed and hopefully `auto` doesn't get confused ;)

PR #5882, opened 2021-08-09
RF: avoid RI types while deducing target path for cloning

yarikoptic:
A complimentary to #5881 (sits on top) fix/RF for deducing target `clone` path from a URL 

Would rely on RI to have non-degenerating .path, or .hostname, and only
then (ab)use entire URL as path, taking the last component
    
The only side effect I envision so far is that `clone http://something.git/` would result in `something` which might be different from git would do ;)

### DIFF
--- a/datalad/core/distributed/clone.py
+++ b/datalad/core/distributed/clone.py
@@ -1207,20 +1207,32 @@ def _get_installationpath_from_url(url):
     from a URL, analog to what `git clone` does.
     """
     ri = RI(url)
-    if isinstance(ri, (URL, DataLadRI, SSHRI)):  # decode only if URL
-        path = ri.path.rstrip('/')
-        path = urlunquote(path) if path else ri.hostname
-        if '/' in path:
-            path = path.split('/')
-            if path[-1] == '.git':
-                path = path[-2]
-            else:
-                path = path[-1]
+
+    def get_last_path(url):
+        "Return the latest part of the Path(url) which is not .git"
+        for part in Path(url).parts[::-1]:
+            if part != '.git':
+                return part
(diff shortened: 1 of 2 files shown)


datalad/core/distributed/clone.py
```diff
@@ -1206,20 +1207,32 @@ def _get_installationpath_from_url(url):
...
+        for part in Path(url).parts[::-1]:
+            if part != '.git':
+                return part
+        return ''
+
+    # consider path from RI, if degenerate - hostname, and if that
+    # one is not known -- URL itself as if it was a Path, but only
+    # the last (non .git) component
+    for cand in (
+        urlunquote(getattr(ri, 'path', '')),
```
yarikoptic, line 1222: well, thinking about it -- may be `urlunquote` should be URL/DataLadRI specific? but I wonder - don't we decode it there?    worth adding an explicit test and ensure that we do not alter precious /home/%e0 etc while decoding them when in URLs... may be later ;)

mih:
I am not 100% confident that I understand the motivation, but it seems to make sense to me. However, the approach does not seem to work on windows (see appveyor). What are your plans re this PR. It has been open for 3 months.

yarikoptic:
I will look into it today. I expect motivation to be revealed if running adjusted unit test with current master logic

yarikoptic:
oh well -- the 2nd commit , the only one specific to this PR, comes without dedicated test, so indeed it was just a RF. I guess motivation was to make deducing path generic while avoiding explicit check for type of the RI, and I did not add any test. Oh well -- no relevant issue was filed since then, so probably it is very safe to just forget about this PR.

PR #5883, opened 2021-08-10
RF Exception reporting - part 2

### LINKED ISSUE #5716: Plans for `exc_str` replacement
Putting the idea for follow ups for PR #5691 here to have the discussion separated.
In that PR I leave `exc_str` as simple "proxy" shim, instead of replacing its calls.

IMO replacing this (its calls) should be taken as an occasion to go through the places where we log exceptions and make sure we don't double log (log where an exception is "handled", not where it occurs (that's why we have partially both)- we have a traceback and don't need the original place for a dedicated log message.
Furthermore I think `CapturedException` should get method `log(logger, level, msg)`, that actually does *two* things:

- log the actual exception at a dedicated level (may be configurable) - probably below DEBUG
- use `level` and `msg` for a second log message after the above that should not include the exception itself. That is for two reasons:
    1. A message containing the exception will often look weird (at least ATM) with something like "Bad thing happened:" where nothing follows the colon with exception logging disabled.
    2. From my POV a log message - particularly since logging is to some degree user output in our case - should focus on a humanly readable and understandable problem description, not a technical one. This is for people actually debugging and it's telling only, if one is somewhat fanilar with datalad's code base.

There are a few other spots, where `exc_str` is used with result dicts. I think, there should be a similar approach. Have a dedicated method for adding a result entry for the exception to a dict. This can be several key value pairs with different formats. It's then up to the result renderer to decide what to make use of.

Last but not least: In both cases (results and logging) the ultimate formatting decision should be moved from the spots where we currently use `exc_str` to log formatter(s) and result renderer(s). I'm not a fan of yet another intermediate step ATM, so I wanted to keep the use of `exc_str` as a marker for what to go through.

### LINKED ISSUE #5848: CaptureException must be used correctly throughout the code base
I saw this in the context of re-rendering the handbook, and checking the diff between the previous and current outputs of datalad commands.
When a ``datalad update`` fails, current ``master`` does not display any hint whatsoever on why. ``maint`` on the other hand tells me so much that it is hard to find the meaningful part of the message:

```sh
(handbook2) adina@muninn in ~/repos/datalad on git:bf-5644
❱ git co master
Switched to branch 'master'
Your branch is ahead of 'gh-adswa/master' by 3 commits.
(handbook2) adina@muninn in ~/repos/datalad on git:master
❱ datalad update -s juselessadina        
[INFO   ] Fetching updates for Dataset(/home/adina/repos/datalad) 
[ERROR  ] Fetch failed:  [update(/home/adina/repos/datalad)] 
update(error): . (dataset) [Fetch failed: ]
```
```sh
(handbook2) adina@muninn in ~/repos/datalad on git:master
❱ git co maint                                                                                                 1 !
Switched to branch 'maint'
Your branch is up to date with 'upstream/maint'.
(handbook2) adina@muninn in ~/repos/datalad on git:maint
❱ datalad update -s juselessadina
[INFO   ] Fetching updates for Dataset(/home/adina/repos/datalad) 
[ERROR  ] Fetch failed: CommandError: 'git -c diff.ignoreSubmodules=none fetch --verbose --progress --no-recurse-submodules --prune juselessadina' failed with exitcode 128 under /home/adina/repos/datalad [err: 'fatal: '/home/adina/scratch/datala22d' does not appear to be a git repository
CommandError: 'ssh -o ControlPath=/home/adina/.cache/datalad/sockets/3b4eaa1d -o SendEnv=GIT_PROTOCOL adina@juseless.inm7.de 'export "PATH=/usr/bin:$PATH"; git-upload-pack '"'"'/home/adina/scratch/datala22d'"'"''' failed with exitcode 128

and the repository exists.'] [cmd.py:run:408] [update(/home/adina/repos/datalad)] 
update(error): . (dataset) [Fetch failed: CommandError: 'git -c diff.ignoreSubmodules=none fetch --verbose --progress --no-recurse-submodules --prune juselessadina' failed with exitcode 128 under /home/adina/repos/datalad [err: 'fatal: '/home/adina/scratch/datala22d' does not appear to be a git repository
CommandError: 'ssh -o ControlPath=/home/adina/.cache/datalad/sockets/3b4eaa1d -o SendEnv=GIT_PROTOCOL adina@juseless.inm7.de 'export "PATH=/usr/bin:$PATH"; git-upload-pack '"'"'/home/adina/scratch/datala22d'"'"''' failed with exitcode 128

and the repository exists.'] [cmd.py:run:408]]
```

The information also is not in the JSON records (there seems to be no message for "fetch failed"):

```sh
(handbook2) adina@muninn in ~/repos/datalad on git:maint
❱ git co master                                                                                                1 !
Switched to branch 'master'
Your branch is ahead of 'gh-adswa/master' by 75 commits.
(handbook2) adina@muninn in ~/repos/datalad on git:master
❱ datalad -f json_pp update -s juselessadina 
[INFO   ] Fetching updates for Dataset(/home/adina/repos/datalad) 
[ERROR  ] Fetch failed:  [update(/home/adina/repos/datalad)] 
{
  "action": "update",
  "message": [
    "Fetch failed: %s",
    ""
  ],
  "path": "/home/adina/repos/datalad",
  "refds": "/home/adina/repos/datalad",
  "status": "error",
  "type": "dataset"
}
```

```sh
(handbook2) adina@muninn in ~/repos/datalad on git:master
❱ git co maint                                                                                                 1 !
Switched to branch 'maint'
Your branch is up to date with 'upstream/maint'.
(handbook2) adina@muninn in ~/repos/datalad on git:maint
❱ datalad -f json_pp update -s juselessadina 
[INFO   ] Fetching updates for Dataset(/home/adina/repos/datalad) 
[ERROR  ] Fetch failed: CommandError: 'git -c diff.ignoreSubmodules=none fetch --verbose --progress --no-recurse-submodules --prune juselessadina' failed with exitcode 128 under /home/adina/repos/datalad [err: 'fatal: '/home/adina/scratch/datala22d' does not appear to be a git repository
CommandError: 'ssh -o ControlPath=/home/adina/.cache/datalad/sockets/3b4eaa1d -o SendEnv=GIT_PROTOCOL adina@juseless.inm7.de 'export "PATH=/usr/bin:$PATH"; git-upload-pack '"'"'/home/adina/scratch/datala22d'"'"''' failed with exitcode 128

and the repository exists.'] [cmd.py:run:408] [update(/home/adina/repos/datalad)] 
{
  "action": "update",
  "path": "/home/adina/repos/datalad",
  "refds": "/home/adina/repos/datalad",
  "status": "error",
  "type": "dataset"
}
```

I guess this is related to #5691? The output that ``maint`` provides is unhelpful because its hard to parse it - especially with the line breaks that the error message from Git adds to it. But the output that ``master`` provides is unhelpful, because now I have to idea at all what might be going on.

bpoldrack:
Follow-up on #5691.

This PR introduces some enhancements to the new `CapturedException` class and goes through every invocation of `exc_str` and `exc_str_old`, replacing (and deprecating) it and trying to achieve several things with respect to our exception reporting:

1. If we log an exception, do so only where an exception is catched, not where it is (re-)raised!
At some spots we still do log the same thing several times. I think the original idea was to produce some kind of a traceback leaving a log entry wherever an exception passes through. This is unnecessary and almost always just annoying double-(or multi)-reporting. The exception logging captures a full traceback including the `__context__` and `__cause__` if an exception was `raised from` and/or happened within an `except` clause. The latest spot has all the info - we only need to log it where we catch it (if only at the outermost layer) and react to it (if only by failing). That's the meaningful spot.

2. When we log or report on (results, or direct user message) exceptions, we should separate a message that is potentially meaningful for a mortal and the technical info on what exactly the exception was and what its traceback is. If really necessary, the former can still contain part of the latter, but the potentially scary and confusing depiction of the exception is a separate thing.
That's what `CapturedException.log()` is for: You pass a message that's supposed to tell what's going on in datalad and that is meaningful/actionable for users. If `datalad.log.exc` is enabled, that record gets enhanced w/ a message reading "Reason for message above:" and the full exception log. When we ever (and I think this is a step to make that more likely) get to properly iron out user messaging and logging, I think `datalad.log.exc` should then be enabled by default for logging into logfiles and by default be disabled for terminal output.
This separation also is the reason to change the default for the `limit` of the depth of the traceback that is included in the "technical" message from `1` to `None` - now that this is it's own message that needs to be enabled, it can as well be as complete as possible by default.

3. When we reraise with a new message (whether or not same type of exception), that message should NOT include the full exception logging. The meaningful thing is to enhance or change the message. Hence the new message may be based on the old message, but the inclusion of the old type and its traceback is wrong. Instead `NewError raise from OldError` injects a `__cause__` to `NewError` and wherever `NewError` is logged, it will have the traceback including that from `OldError`.

4. For result records the same principle applies: user and "debug" message should be separated. @yarikoptic already introduced a modification to `get_status_dict` that, if `exception` is passed, creates an `exception_traceback` entry. I stuck with that for now, so that `message` is formed the way a default log message would be and pass the `CapturedException` object to `exception`. `get_status_dict` then adds the `exception_traceback` field. I enhanced that to actually keep the `exception` filed (Yarik didn't), so that in python (or result renderers) you have access to the `CapturedException` instance, providing several ways of "stringifying" that exception. Ultimately, this should not live in `get_status_dict` (which isn't used everywhere), but be matter of result renderers. However, I think this is best done with an RF'ing to make renderers actual classes (that custom ones can inherit from).

---------------------------

To ease review/discussion, here come some points to look at.
It does make sense to get an idea of the current state of `CapturedException` first, of course: https://github.com/datalad/datalad/pull/5883/commits/ef6b868909e841d31e05ef91a4862ba2827f3219

There are several kinds of changes throughout all the files:

- Simply separating the "technical" message out. Have what's usually meaningful in the message and let `datalad.log.exc` add details if required. Example: https://github.com/datalad/datalad/pull/5883/commits/511b8be688a1ab388e561f1f387e5ef21e451707#diff-35dad0480a094dee25a5777749e74e0d983b4dc742f3b9834230dba46f3f6619L357

- Use `raise from` when we reraise, instead of including the original exception as text (see point 3). Example: https://github.com/datalad/datalad/pull/5883/commits/6c589c5bcc9f5cbb7b6a6752f6a5f74dc53799f9#diff-94f9b0e4cdbcfe44b244bbf4450719d1034680f27d3a1cef81393694ed7d6e08L520

- For result dicts, have the respective "default message" in `message` and add the `CapturedException` to the `exception` field. `get_status_dict` does then add `exception_traceback`. Example: https://github.com/datalad/datalad/pull/5883/commits/13bf1626c003c1b6c0e934f5591804383e5db70b#diff-e378e274f6a8ae458d174d8e6df74ceca070cc8266670ebd1d94c46ffe9ff17eR189

Change to `get_status_dict`: https://github.com/datalad/datalad/pull/5883/commits/13bf1626c003c1b6c0e934f5591804383e5db70b#diff-092358c734aa136a079bbd581698442c22878877b79687b310cd838a1bc82653L92

- One of the main issues, I think, is us being to broad and unspecific, just dumping shit for the user to deal with, instead of assessing what's going on. This comes mainly in two forms: `except Exception` and `except CommandError`. I didn't change any of that, since this is not what the PR is about really, but those two tend to be similar: We actually can (often) know better, but don't do it. For the `CommandError` dumping that is: Something called a subprocess that failed. The caller knows what was called and can usually make an assessment of possible consequences or tell what kind of thing went wrong how. Ideally, say `fetch` would catch `CommandError` and reraise a `GitFetchError` or something like that and that new Exception can come with a proper message that was parsed from git's output and took into account whatever may be relevant from within the knowledge of the caller. Instead, we tend to just pass on the generic `CommandError` and higher up it may be harder to know what to look for to form a huan-digestable message. Dumping `CommandError`'s message my contain the relevant output of the subprocess, but it also contains a lot of distracting, generic stuff, since it's raised by the runner and that one isn't supposed to know a lot about each and every possible subprocess we may want to call. Anyways - that's just something to take note of and to do better. For now, however, I usually went for generating the message from the output instead of taking the entire "message" of `CommandError`, by `e.stderr or e.stdout or e.to_str()`. This means to look at stderr first (it's almost always git/git-annex - they do report an issue on stderr `fatal: ....`. That's the only part that usually is informative for a user when something is failing - buried under lots of irrelevant things. `or e.to_str()` makes the formerly used `exc_str` or `str` of `CommandError` still the fallback if there was no output. Example: https://github.com/datalad/datalad/pull/5883/commits/511b8be688a1ab388e561f1f387e5ef21e451707#diff-2329d131e6de4ec3957da6b6122869ee6b8e490ac65162461ed767dc0d6cef59L593

- For some spots it's perfectly fine to use the standard `str(e)` in messages. That's another point: At some spots it seems that authors weren't aware what `exc_str` actually adds over just `str`. That is: A questionable magic to decide whether to show the exception's name or its message and the option to put the traceback in that string. The former is questionable, because it very much depends on the exception whether or not the exception's name carries relevant information and whether or not the message can stand on its own. And what is the case is not our under control for any exception coming from third-party code. The latter part was nice, but is something that shouldn't be in a result message for example. It's useful for logging only. So: If it's simply about the exception's message and the exception isn't logged at that spot (in which case one can use the already created `CapturedException`): `str(e)` is perfectly fine!
Example: https://github.com/datalad/datalad/pull/5883/commits/6c589c5bcc9f5cbb7b6a6752f6a5f74dc53799f9#diff-37fba60e98afe87a28f73afa6b5bbc3410bad59901363e26256da5584cec265cL623

- Things are a bit different for special remotes, since their messaging to annex isn't exactly "logging". They are also usually a bit painful to debug. Therefore, I went for always including the full detail for now, like here: https://github.com/datalad/datalad/pull/5883/files#diff-67b52fecf48cc364bc9f19345991f2d3f87c11bdbde3cfc9ea16526b422d08b5L393 This is independent of `datalad.log.exc`! Agree, @yarikoptic ? While at it I stumbled upon something else - @yarikoptic Please have a look at this comment please: https://github.com/datalad/datalad/pull/5883/files#diff-67b52fecf48cc364bc9f19345991f2d3f87c11bdbde3cfc9ea16526b422d08b5R326 If we agree, that should become an issue (or a PR ;-) )

- Finally, https://github.com/datalad/datalad/pull/5883/files#diff-eec184e266c9aa336ee741b532169a76e4e2f52759f03ab4ab2dfa9a4d69ddcaR2057
and https://github.com/datalad/datalad/pull/5883/files#diff-eec184e266c9aa336ee741b532169a76e4e2f52759f03ab4ab2dfa9a4d69ddcaR2235 is the same comment. @yarikoptic : Given that the `logger` parameter of those decorators isn't actually used anywhere, we could as well change to passing the logger, rather than a callable, I think. Objections?

TODO:
- [x] check outermost catch
- [x] ~Remove~ deprecate `exc_str` + `exc_old_str`
- [x]  rebase with proper commit messages
- [x] Ease review by pointing out example cases and linking into the diff. Nobody will go through > 60 files.
- [x] double check for other related issues than the already linked one

Closes #5848
Closes #5716

### DIFF
--- a/datalad/cmdline/main.py
+++ b/datalad/cmdline/main.py
@@ -22,7 +22,6 @@ import os
 
 import datalad
 
-from ..dochelpers import exc_str_old
 from ..support.exceptions import (
     InsufficientArgumentsError,
     IncompleteResultsError,
@@ -216,7 +215,7 @@ def main(args=None):
                 ret = cmdlineargs.func(cmdlineargs)
             except InsufficientArgumentsError as exc:
                 # if the func reports inappropriate usage, give help output
-                lgr.error('%s (%s)', exc_str_old(exc), exc.__class__.__name__)
+                lgr.error('%s (%s)', exc_str(exc), exc.__class__.__name__)
                 cmdlineargs.subparser.print_usage(sys.stderr)
                 sys.exit(2)
             except IncompleteResultsError as exc:
@@ -228,7 +227,7 @@ def main(args=None):
                 # in general we do not want to see the error again, but
                 # present in debug output
                 lgr.debug('could not perform all requested actions: %s',
-                          exc_str_old(exc))
+                          exc_str(exc))
                 sys.exit(1)
             except CommandError as exc:
                 # behave as if the command ran directly, importantly pass
@@ -247,7 +246,7 @@ def main(args=None):
                 # had no code defined
                 sys.exit(exc.code if exc.code is not None else 1)
             except Exception as exc:
-                lgr.error('%s (%s)', exc_str_old(exc), exc.__class__.__name__)
+                lgr.error('%s (%s)', exc_str(exc), exc.__class__.__name__)
                 sys.exit(1)
     else:
         # just let argparser spit out its error, since there is smth wrong
@@ -261,7 +260,7 @@ def main(args=None):
         if hasattr(cmdlineargs, 'result_renderer'):
             cmdlineargs.result_renderer(ret, cmdlineargs)
     except Exception as exc:
-        lgr.error("Failed to render results due to %s", exc_str_old(exc))
+        lgr.error("Failed to render results due to %s", exc_str(exc))
         sys.exit(1)
 
 
--- a/datalad/cmdline/main.py
+++ b/datalad/cmdline/main.py
@@ -43,6 +43,7 @@ from .helpers import (
     parser_add_common_opt,
     strip_arg_from_argv,
 )
+from datalad.dochelpers import exc_str
 
 
 # TODO:  OPT look into making setup_parser smarter to become faster
--- a/datalad/__init__.py
+++ b/datalad/__init__.py
@@ -133,7 +133,6 @@ def setup_package():
     # own HOME where we pre-setup git for testing (name, email)
     if 'GIT_HOME' in os.environ:
         set_envvar('HOME', os.environ['GIT_HOME'])
-        set_envvar('DATALAD_LOG_EXC', "1")
     else:
         # we setup our own new HOME, the BEST and HUGE one
         from datalad.utils import make_tempfile
--- a/datalad/cmdline/helpers.py
+++ b/datalad/cmdline/helpers.py
@@ -33,7 +33,7 @@ from ..utils import (
     get_suggestions_msg,
 )
 from ..version import __version__, __full_version__
-from ..dochelpers import exc_str, exc_str_old
+from ..dochelpers import exc_str
 
 from appdirs import AppDirs
 from os.path import join as opj
@@ -440,7 +440,7 @@ def _maybe_get_single_subparser(cmdlineargs, parser, interface_groups,
         lgr.debug("Command line args 1st pass for DataLad %s. Parsed: %s Unparsed: %s",
                   __full_version__, parsed_args, unparsed_args)
     except Exception as exc:
-        lgr.debug("Early parsing failed with %s", exc_str_old(exc))
+        lgr.debug("Early parsing failed with %s", exc_str(exc))
         need_single_subparser = False
         unparsed_args = cmdlineargs[1:]  # referenced before assignment otherwise
     # First unparsed could be either unknown option to top level "datalad"
@@ -564,7 +564,7 @@ def add_entrypoints_to_interface_groups(interface_groups):
             interface_groups.append((ep.name, spec[0], spec[1]))
             lgr.debug('Loaded entrypoint %s', ep.name)
         except Exception as e:
-            lgr.warning('Failed to load entrypoint %s: %s', ep.name, exc_str_old(e))
+            lgr.warning('Failed to load entrypoint %s: %s', ep.name, exc_str(e))
             continue
 
 
--- a/datalad/core/distributed/clone.py
+++ b/datalad/core/distributed/clone.py
@@ -43,7 +43,10 @@ from datalad.support.constraints import (
     EnsureStr,
     EnsureKeyChoice,
 )
-from datalad.support.exceptions import DownloadError
+from datalad.support.exceptions import (
+    CapturedException,
+    DownloadError,
+)
 from datalad.support.param import Parameter
 from datalad.support.strings import get_replacement_dict
 from datalad.support.network import (
@@ -293,10 +296,12 @@ class Clone(Interface):
             # spec is determined to be useless
             path.exists()
         except OSError as e:
+            ce = CapturedException(e)
             yield get_status_dict(
                 status='error',
                 path=path,
-                message=('cannot handle target path: %s', exc_str(e)),
+                message=('cannot handle target path: %s', ce),
+                exception=ce,
                 **result_props)
             return
 
@@ -587,11 +592,12 @@ def clone_dataset(
                 create=True)
 
         except CommandError as e:
+            ce = CapturedException(e)
             e_stderr = e.stderr
 
             error_msgs[cand['giturl']] = e
             lgr.debug("Failed to clone from URL: %s (%s)",
-                      cand['giturl'], exc_str(e))
+                      cand['giturl'], ce)
             if dest_path.exists():
                 lgr.debug("Wiping out unsuccessful clone attempt at: %s",
                           dest_path)
@@ -641,7 +647,7 @@ def clone_dataset(
                 error_msg = "Failed to clone from any candidate source URL. " \
                             "Encountered errors per each url were:\n- %s"
                 error_args = '\n- '.join(
-                    '{}\n  {}'.format(url, exc_str(exc))
+                    '{}\n  {}'.format(url, exc.to_str())
                     for url, exc in error_msgs.items()
                 )
         else:
--- a/datalad/core/local/run.py
+++ b/datalad/core/local/run.py
@@ -42,7 +42,10 @@ import datalad.support.ansi_colors as ac
 from datalad.support.constraints import EnsureChoice
 from datalad.support.constraints import EnsureNone
 from datalad.support.constraints import EnsureBool
-from datalad.support.exceptions import CommandError
+from datalad.support.exceptions import (
+    CapturedException,
+    CommandError
+)
 from datalad.support.globbedpaths import GlobbedPaths
 from datalad.support.param import Parameter
 from datalad.support.json_py import dump2stream
@@ -481,8 +484,10 @@ def _unlock_or_remove(dset_path, paths):
             try:
                 os.unlink(res["path"])
             except OSError as exc:
+                ce = CapturedException(exc)
                 yield dict(res, action="run.remove", status="error",
-                           message=("Removing file failed: %s", exc_str(exc)))
+                           message=("Removing file failed: %s", ce),
+                           exception=ce)
             else:
                 yield dict(res, action="run.remove", status="ok",
                            message="Removed file")
--- a/datalad/customremotes/tests/test_archives.py
+++ b/datalad/customremotes/tests/test_archives.py
@@ -234,7 +234,7 @@ def test_interactions(tdir):
             ('VALUE dl+archive://somekey3#path', None),
             ('VALUE',
              re.compile(
-                 'TRANSFER-FAILURE RETRIEVE somekey Failed to fetch any '
+                 'TRANSFER-FAILURE RETRIEVE somekey RuntimeError\(Failed to fetch any '
                  'archive containing somekey. Tried: \[\]')
              )
         ],
--- a/datalad/customremotes/tests/test_datalad.py
+++ b/datalad/customremotes/tests/test_datalad.py
@@ -92,7 +92,7 @@ def test_interactions(tdir):
     fetch_scenarios.append(
         ('VALUE',
          re.compile(
-             'TRANSFER-FAILURE RETRIEVE somekey Failed to download from any')))
+             'TRANSFER-FAILURE RETRIEVE somekey RuntimeError\(Failed to download from any')))
 
     for scenario in BASE_INTERACTION_SCENARIOS + [
         [
--- a/datalad/distributed/create_sibling_gitlab.py
+++ b/datalad/distributed/create_sibling_gitlab.py
@@ -30,6 +30,7 @@ from ..support.constraints import (
     EnsureNone,
     EnsureStr,
 )
+from datalad.support.exceptions import CapturedException
 from ..utils import (
     ensure_list,
 )
@@ -419,11 +420,13 @@ def _proc_dataset(refds, ds, site, project, remotename, layout, existing,
                 status='ok',
             )
         except Exception as e:
+            ce = CapturedException(e)
             yield dict(
                 res_kwargs,
                 # relay all attributes
                 status='error',
-                message=('Failed to create GitLab project: %s', exc_str(e))
+                message=('Failed to create GitLab project: %s', ce),
+                exception=ce
             )
             return
     else:
--- a/datalad/distributed/export_archive_ora.py
+++ b/datalad/distributed/export_archive_ora.py
@@ -34,6 +34,7 @@ from datalad.support.constraints import (
     EnsureNone,
     EnsureStr,
 )
+from datalad.support.exceptions import CapturedException
 from datalad.distribution.dataset import (
     EnsureDataset,
     datasetmethod,
@@ -188,11 +189,13 @@ class ExportArchiveORA(Interface):
                 status='ok',
                 **res_kwargs)
         except Exception as e:
+            ce = CapturedException(e)
             yield get_status_dict(
                 path=str(archive),
                 type='file',
                 status='error',
-                message=('7z failed: %s', exc_str(e)),
+                message=('7z failed: %s', ce),
+                exception=ce,
                 **res_kwargs)
             return
         finally:
--- a/datalad/distribution/create_sibling.py
+++ b/datalad/distribution/create_sibling.py
@@ -70,6 +70,7 @@ from datalad.support.constraints import (
     EnsureStr,
 )
 from datalad.support.exceptions import (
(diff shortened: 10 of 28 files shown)


mih:
Uff, this diff is a challenge. Main points for me:

It seems `str(CapturedException)` is no longer empty. Good!

This looks painful TBH:

```py
CapturedException(e).log(
    lgr,
    logging.WARNING,
    "Failed to load entrypoint %s",
    entry_point.name)
```

I think there could be a `.warning()` (and the others) to make it less wordy.

Looking through the use cases I am wondering why it would not be sufficient to have the following two cases

```
lgr.XXX("some message with %s custom placement of exception stringification, CapturedException(e))

lgr.XXX("detailed technical blurb: %s, CapturedException(e))
```

There would be no stringification happening, unless the message is actually logged (not the case right now, not in master, not with this PR, unless specifically implemented for a case).

The fact that an exception was captured can be reported independently of the specific handling in a generic logging (low-level) in `CapturedException.__init__()`.

How an exception is reported can be altered in code via `%s` and `%r`, and via config by reading the respective switches in `CaptureException.__str/repr__()`.

In order for this to work there should be clearly defined conditions, like `str(CaptureException(...))` always gives a non-empty string, to be able to compose tailored messages -- avoiding the clunky "Reason for message above: %s"

bpoldrack:
@mih Thanks very much for looking through already!

1.

> I think there could be a .warning() (and the others) to make it less wordy.

True. The reason, I didn't do that (yet), is that I wasn't sure, whether to optionally enhance the given log message with that appendix or to optionally add another log entry (lower level) instead. In the latter case I'd need access to the `Logger`, not just a callable like `lgr.warning`. And `CapturedException.warning(lgr, ...)` wouldn't save anything more than a `logging.`. But yes: If we ultimately agree on the main points, I'll change that accordingly.
However, this:

> The fact that an exception was captured can be reported independently of the specific handling in a generic logging (low-level) in CapturedException.__init__().

could be an alternative indeed!

2.

> Looking through the use cases I am wondering why it would not be sufficient to have the following two cases
> ```
> lgr.XXX("some message with %s custom placement of exception stringification, CapturedException(e))
> lgr.XXX("detailed technical blurb: %r, CapturedException(e))
> ```

Because then it's back to every contributor what is reported how. One might see that as an advantage in flexibility, but I'd argue, that's the problem with confusing, scary or superfluous messages that are not well suited for consumption by users. This `log()` method's intention is to somewhat force us, to think about what message makes sense w/o the exception (or its details). I think it's wrong to try to rely on the exception carrying the decisive information to base a message on. In a lot of cases we could do much better. Instead we - as developers - find the exception telling and don't phrase a proper message, because it's easy for us to relay it and we can tell ourselves, that that is the important part. However, that approach is where a lot of unfortunate user messaging originates from. So, I tried to not encourage ourselves to be lazy about it.
It also tends to generate redundant output. Say: "git-clone failed: `CommandError('git', 'clone', ...)`". We do know what call we made and that it failed the moment we catch that exception. And we can (and often do) put it in actual english wording, only to then repeat the exact same information via some generic representation of an exception that may or may not carry some additional information, may or may not repeat what we just said and may or may not be useful for anyone but us. The "real" solution for the `CommandError` madness would be much more work, but it shows some general principle that's true for other exceptions as well: When we catch it, we usually can make a better statement about what's going on, that would actually be useful for a non-developer.

Another aspect of why `CapturedException.log` instead: I think, `CapturedException` ultimately has no business interpreting log configs. Eventually it should be log formatters/handlers that deal with it. Originally (when coding the first part) I imagined to do that here in the second part, but that involves more regarding the logging and how we do and test it. Since this is big enough as is, I went for: "at least have a dedicated method for logging, so that things can be refactored without breaking something like `__str__` that may be used outside logging as well".

Does that make sense to you?

mih:
> Does that make sense to you?

Not really. If I condense your statement, you are saying: "The exception is in many cases not the thing that is most important, and therefore we make logging a function of that exception." This seems rather twisted.

I also do not agree with the premise "make things more complicated to force us to think". I think the logging is the important bit. If a developer considers it appropriate to leave a trace of an exception in a log message, this is either the right call, are not in compliance with a rule that was never written down -- and should be appropriately discussed. I seriously doubt that any central machinery will magically improve the utility of individual log messages. That being said, I totally agree that `exc_str()` is not a good tool in all but a few cases.

I still stand by my proposal. The only thing it is lacking is the case of a captured exception that should leave NO trace in a log message (or if there is no log message, although that case should probably not exist). This case could be addressed by

```
CapturedException(e)
lgr.XXX("some standalone blurb)
```

`CapturedException.__init__()` can take care of any standardized reporting, while the targeted logging can do whatever is best in this particular case.

bpoldrack:
>  "The exception is in many cases not the thing that is most important, and therefore we make logging a function of that exception."

Not exactly. I didn't mean to make a point of making it a function of `CapturedException` (I said it should ultimately be dealt with by the logger mechanism). I tried to make a point of NOT making that functionality part of a method of `CapturedException` that withstands refactoring (`__str__`, `__repr__`). The `.log()` we can move, when we got a shot on the log handling in general. If we put the config into `__str__`, etc. then that config effects any use of it, is not limited to logging and the likely later move does change the stringification/representation outside logging. At least that was the argument I was trying to convey.

> If a developer considers it appropriate to leave a trace of an exception in a log message, this is either the right call, are not in compliance with a rule that was never written down -- and should be appropriately discussed.

Not sure how I can bring that statement in coherence with you opening #4478. If it's a case by case decision where any developer decides about the level of detail for every log message, you can't silence anything. There may be no magic solution to every aspect of it, but everybody doing whatever they thought was sensible from within the context in which they wrote that piece is what got us to the kind of messages we have. Don't see how that's a solution.

> or if there is no log message, although that case should probably not exist

FTR: That case does exist. Whether it should exist - well, why not? The fact that an exception occurred can be absolutely expected and a regular thing. The exception is catched, leads to a different call path and that's it. No need for a dedicated log message. Just like we don't log the outcome of every conditional.

> I still stand by my proposal
> ...
> CapturedException.__init__() can take care of any standardized reporting, while the targeted logging can do whatever is best in this particular case.

Trying to sum up what I got of your proposal:
  - `__repr__` provides exception name, message, traceback?
  - `__str__` is some short form like `SomeError(some message)` ?
  - `__init__` does automatically (unconditionally?) do a low-level log of the exception (full detail I suppose)
  - `__str__` and `__repr__` somehow respect configs - it is unclear to me which ones. If they are supposed to provide the explicit means to the dev to decide whether or not to show the details, what switch would be there to respect?

datalad/cmdline/main.py
```diff
@@ -216,19 +216,22 @@ def main(args=None):
                 ret = cmdlineargs.func(cmdlineargs)
             except InsufficientArgumentsError as exc:
                 # if the func reports inappropriate usage, give help output
-                lgr.error('%s (%s)', exc_str_old(exc), exc.__class__.__name__)
+                ce = CapturedException(exc)
+                ce.log(lgr, logging.ERROR, "%s", ce.format_short())
```
yarikoptic, line 220: FWIW, I think we should aim for with  
```suggestion
                lgr.error("", exc_info=exc)
```
with all needed for logging as desired to happened by logger's Formatter

yarikoptic: or
```suggestion
                CapturedException(exc).log(lgr=lgr)
```
although thinking is needed about cases where custom msg etc is desired, then could be those
```
                CapturedException(exc).log(msg, *args, lgr=lgr)
```
so `.log` gets signature of logging's `_log` with extra `lgr`

mih:
First line in each `except` branch is `CapturedException(exception)`, unless an exception is raised.

`raise from` is to be used whenever a different exception is raised than the one that was caught.

The constructor generates a low-level log message that an exception was captured.

This log message is `lgr.log(level, '%r')`.

@yarikoptic does not object in the absence of full grasp.

@mih thinks this is alrighty.

bpoldrack:
For readability closing in favor of #5954

PR #5884, opened 2021-08-12
Download session/credentials locking -- inform user if locking is "failing" to be obtained, fail upon ~5min timeout

### LINKED ISSUE #5099: Stale downloader lock prevents user from `datalad get`
It took enabling DEBUG log to see

```
[DEBUG] Acquiring a currently existing lock to establish download session. If stalls - check which process holds /home/XXX/.cache/datalad/locks/downloader-auth.lck 
```

It is unclear where that lockfile came from, but it was two weeks old.

For me any Ctrl-C leaves a lockfile behind. I suspect that any terminated cluster job could have the same result.

At the moment I have too little insight into why that lock is needed, what for and when exactly. In the present case, the respective credentials came from an envvar.

If my speculations are correct, then placing a lock into a user-global location that could stall any download on a cluster is a fragile setup.

Orthogonal to that: Information for finding this rather impactful problem was hidden too deep, I think.

yarikoptic:
I think it could be considered to resolve #5099 - user will be given INFO level log messages (just now realized -- may be the last one at least should be a warning??? or should we have a progress bar? ;-)) with all possible information we can provide, even which processes have that lock ATM (well -- might not universally work... windows - I am talking to you).

I also decided that "heck with that locking" if we fail to get the lock within 5 minutes!  at least this way may be some processes would complete (but possibly hide the problem).

Please see individual commits for more information.  It seems to come out to be again yet another "one time use locking helper", but since we have not come up with a better way to address this locking "issue", I think it is a small price to pay ;)

- [-] I have positioned against `maint` but it grew a bit too big.  I will be fine to reposition it against `master` -- we should get one out of the door soon.
- [-] I wonder if I should add some randomization so that not all parallel processes which fail to acquire lock start asking for credentials at the same time ... ;-)

### DIFF
--- a/datalad/support/tests/test_locking.py
+++ b/datalad/support/tests/test_locking.py
@@ -1,33 +1,40 @@
 #emacs: -*- mode: python-mode; py-indent-offset: 4; tab-width: 4; indent-tabs-mode: nil -*- 
 #ex: set sts=4 ts=4 sw=4 noet:
-"""
- LICENSE: MIT
-
-  Permission is hereby granted, free of charge, to any person obtaining a copy
-  of this software and associated documentation files (the "Software"), to deal
-  in the Software without restriction, including without limitation the rights
-  to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
-  copies of the Software, and to permit persons to whom the Software is
-  furnished to do so, subject to the following conditions:
-
-  The above copyright notice and this permission notice shall be included in
-  all copies or substantial portions of the Software.
-
-  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
-  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
-  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
-  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
-  THE SOFTWARE.
-"""
+# ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+#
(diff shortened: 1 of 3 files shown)


datalad/support/tests/test_locking.py
```diff
@@ -98,3 +106,74 @@ def test_lock_if_check_fails(tempfile):
...
+        assert_in(f'Failed to acquire lock at {lock_path} in 0.15', res['stderr'])
+        assert_in('proceed without locking', res['stderr'])
+        assert_greater(time() - t0, 0.19999)  # should wait for at least 0.2
+        try:
+            # PID does not correspond
+            assert_in('Check following process: PID=', res['stderr'])
+            assert_in(f'CWD={os.getcwd()} CMDLINE=', res['stderr'])
+        except AssertionError:
+            # we must have had the other one then
+            assert_in('failed to determine one', res['stderr'])
```
yarikoptic, line 157: note: no coverage here is odd since it should be happening on OSX... aren't we pushing coverage from those???

yarikoptic: hm, looking at `/mnt/btrfs/datasets/datalad/ci/logs/2021/08/12/pr/5884/9827dc3/github-Test on macOS-1723-success/test (brew)/9_Upload coverage report to Codecov.txt` -- coverage is successfully uploaded, and points to https://codecov.io/github/datalad/datalad/commit/9827dc38edf72106d429ee78dfa8d632a92bed43 but that one is not "run specific".  I wonder if combining from multiple submissions works correctly, or just the last one to arrive (travis? ;)) wins it all?

yarikoptic:
I think I feel brave enough with this to proceed against `maint` without any further changes.

mih:
I posted an extensive review yesterday. But there is no trace of it, wtf!? I have no time right now to rewrite it, but the bottom line was a "request changes", because ignoring locking in this fashion might pave over actual problems that we better get fixed. That being said, I unconditionally like the more informative output!

yarikoptic:
> I posted an extensive review yesterday. But there is no trace of it, wtf!?

Happens to the best of us (did you grow to carry over a dozen of tabs in the browser?).

> ... ignoring locking in this fashion might pave over actual problems that we better get fixed.

I was on the edge in this but even though I thought to argue "but that is only for us to lock for credentials input!", experience in troubleshooting git-annex in #5400 (where git-annex tries to lock for quite a while to only give up eventually and proceed) kinda tilted me to the opinion, that yeah -- let's just error out so no such abnormal hard-to-debug cases emerge, and behavior would be somewhat consistent with current `maint` where it would hang/"error out by being killed".  Pushing 1 line change

mih (approved):
Cool, thx! Let's give it a go!

PR #5887, opened 2021-08-12
RF: use https:// for datasets*.datalad.org

### LINKED ISSUE #5885: Please use https instead of http on the command line program
It tries to connect with http://datasets.datalad.org but my firewall blocks insecure connections.

yarikoptic:
Closes #5885

- [x] might want to go against master to avoid future merge conflicts etc... we need to release!

also added some extra ignores to .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,5 @@ tmp
 .*.swp
 .travis.yml.evil-bd
 .asv
+bin/
+tools/bisect-git-annex.simg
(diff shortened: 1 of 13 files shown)


yarikoptic:
ok -- all is still green, so good to go. not sure if it is `internal` or any other `patch` or `minor` since someone could claim that it is some kind of a "public API" in either http or https gets cloned for `///`. decided to not think too much about that.

yarikoptic:
I am chickening out -- will change base to `master`.

yarikoptic:
restarted one failed on windows , failed on relevant change -- failed to clone from the https... tried locally - worked. let's see what rerun says, might still be an unstable connection although somewhat odd to coincide -- may be https connection to the server is a bit more flaky somehow (more interactions needed etc).

yarikoptic:
ok, green, since not really that dramatic change and is against master, will proceed

PR #5888, opened 2021-08-12
.travis.yml: use python -m {nose,coverage} invocations, and always show combined report

yarikoptic:
Individual commits say more.  I do not expect any effect on coverage.


yarikoptic:
hm... such total greenness is worrisome since we usually expect lower coverage -- did I fix something as a side effect? ;-)

yarikoptic:
since only got better etc, I will just proceed with merge

PR #5892, opened 2021-08-13
BF: do not crash, just skip whenever trying to delete non existing field in the underlying keyring

### LINKED ISSUE #5889: Need to guard for deletion of a credential field which is not known
Reported in https://github.com/ReproNim/datalad-nda/issues/1#issuecomment-897786493 .

we have
```
    def delete(self):
        """Deletes credential values from the keyring"""
        for f in self._FIELDS:
            self._keyring.delete(self.name, f)
```
```
    def delete(self, name, field=None):
        if field is None:
            raise NotImplementedError("Deletion of all fields associated with a name")
        return self._keyring.delete_password(self._get_service_name(name), field)
```
so if for a specific field value was not known (no need to delete) -- `delete_password` kabooms ... I don't see keyring module interface for `is_known`, and our `is_known` relies on ` self._keyring.get(self.name, f) is not None` so we could do the same here or catch exception and analyze for 

```
        if not self.connected(service):
            # the user pressed "cancel" when prompted to unlock their keyring.
            raise PasswordDeleteError("Cancelled by user")
        if not self.iface.hasEntry(self.handle, service, username, self.appid):
            raise PasswordDeleteError("Password not found")
```
so pretty much "not found" message in the exception since there is no dedicated exception...

yarikoptic:
Fixing bug leading to a crash: We have pairs or more in composite
credentials. If user saved some but not all of them, then upon renew whenever
we try to delete them all we would crash here since keyring cannot delete the
ones it does not yet know.  So the easiest (although adhoc due to need to match
msg) is to handle such an exception.

To get it tested, I had to use a "real" keyring's backend (so they are called),
and for that I had to RF our Keyring helper class to be capable of being
provided a keyring backend instance (or in other words -- a "keyring").  An odd
logic abit I left behind is that even though we are now explicitly asking for
what keyring will be used, so we could store/use that instance if no explicit
backend is provided, but for consistency with current behavior (since to go
against maint), decided to keep it as is and keep using module level bound
interfaces of keyring module in case if no backend instance was provided.

So now ._keyring  could be either keyring module (no backend instance provided,
or actual keyring backend if it was provided (used only in the tests).

Added test also tests basic assumption that keyring is storing the name
somewhere, and deletes entire section whenever we delete entire credential.
Relies on "ad-hoc" way to specify the filename. Somewhat inquired in
https://github.com/jaraco/keyrings.alt/issues/45 on a better way.

Closes #5889

### DIFF
--- a/datalad/downloaders/tests/test_credentials.py
+++ b/datalad/downloaders/tests/test_credentials.py
@@ -15,9 +15,13 @@ from datalad.tests.utils import (
     assert_in,
     assert_raises,
     assert_true,
+    ok_file_has_content,
     SkipTest,
+    skip_if,
+    with_tempfile,
     with_testsui,
 )
+from datalad.support.external_versions import external_versions
 from datalad.support.keyring_ import (
     Keyring,
     MemoryKeyring,
@@ -57,6 +61,7 @@ def test_cred1_enter_new():
     cred.enter_new(user='user2')
     assert_equal(keyring.get('name', 'user'), 'user2')
     assert_equal(keyring.get('name', 'password'), 'newpassword')
+
 
 @with_testsui(responses=['password1'])
 def test_cred1_call():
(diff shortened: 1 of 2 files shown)


mih (approved):
LGTM. The introduction of the backend switch is borderline invasive for `maint`, but it seems to be safe.

yarikoptic:
thank you @mih.  Yeah -- was thinking the same about `maint`... but probably will take the chances now instead of RFing (I guess we could mock needed things out from our `Keyring`). Let's proceed (since approved)

Issue #5893, opened 2021-08-14
CRON NFS tests stall forever


yarikoptic:
since the beginning of the con/tinuous collection of logs we have:

```shell
(git)smaug:/mnt/btrfs/datasets/datalad/ci/logs/2021/05[master]git
$> git grep -B2 'No output has been received in the last'
25/cron/20210525T224617/69bb6b1/travis-11967-errored/15.txt-..
25/cron/20210525T224617/69bb6b1/travis-11967-errored/15.txt-
25/cron/20210525T224617/69bb6b1/travis-11967-errored/15.txt:No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
--
28/cron/20210528T190442/1c6afc4/travis-11976-errored/15.txt-E..
28/cron/20210528T190442/1c6afc4/travis-11976-errored/15.txt-
28/cron/20210528T190442/1c6afc4/travis-11976-errored/15.txt:No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
```

NFS testing was allowed to fail LONG ago in 105560892765604056539d5fa418d5f30612da79 apparently since "they just take too long" and "they make master look broken for no good reason" ... 
meanwhile -- later we added testing fresh git-annex builds tested (but on git actions) on NFS and they generally were green, which is a bit odd (need to doublecheck if all is "kosher there") since it seems NFS issues aren't behind us (e.g. see fresh https://git-annex.branchable.com/bugs/__34__357_out_of_984_tests_failed__34___on_NFS_lustre_mount/ with "357 out of 984 tests failed", although might be lustre not NFS issue)

yarikoptic:
seems to stall on
```
datalad.interface.tests.test_run_procedure.test_text2git ... 
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received
```
will now try running only that test

yarikoptic:
FWIW, it is still happening apparently
```
26/cron/20220826T223348/dc77a08/travis-14500-errored/17.txt-../datalad/local/tests/test_run_procedure.py::test_text2git 
26/cron/20220826T223348/dc77a08/travis-14500-errored/17.txt-
[...]

PR #5895, opened 2021-08-14
TEMP: trying to see what is up with NFS

yarikoptic:
ref: #5893


yarikoptic:
uff, ok, FWIW -- now we have a helper script which could be used locally and is used on travis to the same effect of hanging (on travis but not locally; but different git-annex I bet).

yarikoptic:
eh, even with the same git-annex (installed the same way) and the same ` sudo git config --system annex.pidlock true` -- no stall locally. eh heh.  Next: will run in increasing debug mode to see what/where it actually stalls

yarikoptic:
so, on
```
[DEBUG  ] Async run:
|  cwd=/tmp/datalad-nfs-NR7GK.nfs/datalad_temp_tree_test_text2gitnk5_rrsw
|  cmd=['git', '-c', 'diff.ignoreSubmodules=none', 'annex', 'add', '--json', '--json-error-messages', '-c', 'annex.dotfiles=true', '-c', 'annex.retry=3', '--', '0blob', 'JSON', 'MARKDOWN', 'YAML', 'empty', 'emptyline'] 
[DEBUG  ] Launching process ['git', '-c', 'diff.ignoreSubmodules=none', 'annex', 'add', '--json', '--json-error-messages', '-c', 'annex.dotfiles=true', '-c', 'annex.retry=3', '--', '0blob', 'JSON', 'MARKDOWN', 'YAML', 'empty', 'emptyline'] 
[DEBUG  ] Process 17926 started 
[DEBUG  ] Waiting for process 17926 to complete 
```
will try now with freshish git-annex (8.20210803) -- there were various fixes etc since old times. If stalls there -- will do even heavier debug mode to see what git-annex spits out

yarikoptic:
ok, I think it is boiling down not really to a "stall" but parallel annex processes (batched? yet to check) locking via pidlock and thus
```
2021-08-15 22:33:35,025 [Level 5] Read 98 bytes from 15899[stderr]: 
2021-08-15 22:33:35,025 [Level 5] stderr|   waiting for pid lock file .git/annex/pidlock which is held by another process (or may be stale)
```
so `annex add` invocation then "times out" at travis level since takes over 10 minutes (600 sec)... even if we enable debug out -- test eventually completes and run (e.g. https://app.travis-ci.com/github/datalad/datalad/builds/235376207) is green.

Also ran `annex test` presumably under NFS in #5896 -- all good and green

yarikoptic:
re pidlock/locking -- may be https://github.com/datalad/datalad/issues/5993 relates... not going to dig this NFS issue in the near future so let's let this PR RiP

PR #5896, opened 2021-08-15
TEMP: Run git annex tests under NFS on travis

yarikoptic:
Sits on top of #5895


yarikoptic:
well, tests passed but umount failed due to
```
All 984 tests passed (922.53s)
+ret=0
+echo 'I: done, unmounting'
I: done, unmounting
+sudo umount /tmp/datalad-nfs-RiTr7.nfs
umount.nfs4: /tmp/datalad-nfs-RiTr7.nfs: device is busy
The command "$NOSE_WRAPPER git annex test" exited with 16.
```
so most likely some run away git-annex process!  will gather more info

PR #5899, opened 2021-08-16
Use platform-appropriate "echo" calls in tests

### LINKED ISSUE #5870: TST: wrong invocation of "echo" on Windows?
I ran tests under Windows and noticed a few failures where ``subprocess.Popen()`` would complain with a WinError2 ("Can't find the specified file"), such as

```
ERROR: datalad.tests.test_nonasyncrunner.test_inside_async
  File "C:\Users\datalads\AppData\Local\Programs\Python\Python39\lib\site-packages\nose\case.py", line 198, in runTest
    self.test(*self.arg)
  File "C:\Users\datalads\repos\datalad\datalad\tests\test_nonasyncrunner.py", line 141, in test_inside_async
    result = loop.run_until_complete(main())
  File "C:\Users\datalads\AppData\Local\Programs\Python\Python39\lib\asyncio\base_events.py", line 642, in run_until_complete
    return future.result()
  File "C:\Users\datalads\repos\datalad\datalad\tests\test_nonasyncrunner.py", line 138, in main
    return runner.run(["echo", "abc"], StdOutCapture)
  File "C:\Users\datalads\repos\datalad\datalad\cmd.py", line 149, in run
    results = run_command(
  File "C:\Users\datalads\repos\datalad\datalad\nonasyncrunner.py", line 194, in run_command
    process = subprocess.Popen(cmd, **kwargs)
  File "C:\Users\datalads\AppData\Local\Programs\Python\Python39\lib\subprocess.py", line 951, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "C:\Users\datalads\AppData\Local\Programs\Python\Python39\lib\subprocess.py", line 1420, in _execute_child
    hp, ht, pid, tid = _winapi.CreateProcess(executable, args,
FileNotFoundError: [WinError 2] Das System kann die angegebene Datei nicht finden
```

```
ERROR: datalad.tests.test_witless_runner.test_runner_no_stdin_no_capture
  File "C:\Users\datalads\AppData\Local\Programs\Python\Python39\lib\site-packages\nose\case.py", line 198, in runTest
    self.test(*self.arg)
  File "C:\Users\datalads\AppData\Local\Programs\Python\Python39\lib\site-packages\nose\tools\nontrivial.py", line 97, in newfunc
    result = func(*arg, **kw)
  File "C:\Users\datalads\repos\datalad\datalad\tests\test_witless_runner.py", line 170, in test_runner_no_stdin_no_capture
    runner.run(
  File "C:\Users\datalads\repos\datalad\datalad\cmd.py", line 149, in run
    results = run_command(
  File "C:\Users\datalads\repos\datalad\datalad\nonasyncrunner.py", line 194, in run_command
    process = subprocess.Popen(cmd, **kwargs)
  File "C:\Users\datalads\AppData\Local\Programs\Python\Python39\lib\subprocess.py", line 951, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "C:\Users\datalads\AppData\Local\Programs\Python\Python39\lib\subprocess.py", line 1420, in _execute_child
    hp, ht, pid, tid = _winapi.CreateProcess(executable, args,
FileNotFoundError: [WinError 2] Das System kann die angegebene Datei nicht finden
```

In these cases, the test would call ``echo`` with some arguments, e.g.,  ``runner.run(["echo", "abc"]``. On Windows, [``echo`` is a command within ``cmd.exe``](https://stackoverflow.com/questions/44371876/why-is-popencmd-exe-echo-a-shell-true-not-running-echo). When I replace ``echo`` with ``"C:\\WINDOWS\\system32\\cmd.exe", "/c", "echo"``, ``subprocess.Popen()`` can execute the command. 

I know that via some magic, some things replace the Unixy call with a Windows equivalent. E.g., during start-up of datalad, I see a ``'C:\\WINDOWS\\system32\\cmd.exe /c "ver"'`` call being executed via ``Subprocess.Popen()``. But I am not sure how that works, or if this is intended for the tests at hand, or even anyhow useful

mih:
This implements the suggestion made by @adswa in #5870. 'echo' is a
shell built-in, so call it with a shell.

Fixes #5870

Plus some minor changes to make other test code for the runner run on windows.

### DIFF
--- a/datalad/tests/test_nonasyncrunner.py
+++ b/datalad/tests/test_nonasyncrunner.py
@@ -15,11 +15,26 @@ import signal
 import subprocess
 from time import sleep
 
-from datalad.tests.utils import assert_false, assert_true, eq_, \
-    known_failure_windows, known_failure_osx, with_tempfile
+from datalad.tests.utils import (
+    assert_false,
+    assert_true,
+    eq_,
+    known_failure_windows,
+    known_failure_osx,
+    with_tempfile,
+)
 
-from ..cmd import WitlessProtocol, WitlessRunner, StdOutCapture
-from ..nonasyncrunner import _ReaderThread, run_command
+from datalad.cmd import (
+    StdOutCapture,
+    WitlessProtocol,
+    WitlessRunner,
+)
+from datalad.nonasyncrunner import (
+    _ReaderThread,
+    run_command,
+)
+
+from datalad.utils import on_windows
 
 
 @known_failure_windows  # Windows uses different signals and commands
@@ -135,11 +150,13 @@ def test_thread_exit():
 def test_inside_async():
     async def main():
         runner = WitlessRunner()
-        return runner.run(["echo", "abc"], StdOutCapture)
+        return runner.run(
+            (["cmd.exe", "/c"] if on_windows else []) + ["echo", "abc"],
+            StdOutCapture)
 
     loop = asyncio.get_event_loop()
     result = loop.run_until_complete(main())
-    eq_(result["stdout"], "abc\n")
+    eq_(result["stdout"], "abc" + os.linesep)
 
 
 @known_failure_osx
--- a/datalad/tests/test_nonasyncrunner.py
+++ b/datalad/tests/test_nonasyncrunner.py
@@ -37,7 +37,6 @@ from datalad.nonasyncrunner import (
 from datalad.utils import on_windows
 
 
-@known_failure_windows  # Windows uses different signals and commands
 def test_subprocess_return_code_capture():
 
     class KillProtocol(WitlessProtocol):
@@ -60,16 +59,20 @@ def test_subprocess_return_code_capture():
         def process_exited(self):
             self.result_pool["process_exited_called"] = True
 
-    signal_to_send = signal.SIGINT
+    # windows doesn't support SIGINT but would need a Ctrl-C
+    signal_to_send = signal.SIGTERM if on_windows else signal.SIGINT
     result_pool = dict()
-    result = run_command(["sleep", "10000"],
+    result = run_command(['timeout', '3'] if on_windows else ["sleep", "10000"],
                          KillProtocol,
                          None,
                          {
                              "signal_to_send": signal_to_send,
                              "result_pool": result_pool
                          })
-    eq_(result["code"], -signal_to_send)
+    if not on_windows:
+        # this one specifically tests the SIGINT case, which is not supported
+        # on windows
+        eq_(result["code"], -signal_to_send)
     assert_true(result_pool["connection_lost_called"][0])
     assert_true(result_pool["process_exited_called"])
 
@@ -159,6 +162,8 @@ def test_inside_async():
     eq_(result["stdout"], "abc" + os.linesep)
 
 
+# Both Windows and OSX suffer from wrapt's object proxy insufficiency
+# NotImplementedError: object proxy must define __reduce_ex__()
 @known_failure_osx
 @known_failure_windows
 @with_tempfile(mkdir=True)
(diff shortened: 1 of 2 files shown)


mih:
Test failures are exclusively caused by a previous issue in master that was fixed by #5897

datalad/tests/test_nonasyncrunner.py
```diff
@@ -135,18 +154,21 @@ def test_thread_exit():
...
+    eq_(result["stdout"], "abc" + os.linesep)
 
 
 @known_failure_osx
-@known_failure_windows
 @with_tempfile(mkdir=True)
 @with_tempfile
 def test_popen_invocation(src_path, dest_path):
+    if on_windows:
+        raise SkipTest("Test requires not implemented __reduce_ex__ of wrapt's object proxy")
```
yarikoptic, line 171: I would keep this test as is, and just add a comment near decorators -- it is not only windows but OSX as well, according to https://github.com/GrahamDumpleton/wrapt/issues/102#issuecomment-766833709 

```suggestion
```

datalad/tests/test_nonasyncrunner.py
```diff
@@ -135,18 +154,21 @@ def test_thread_exit():
...
+            (["cmd.exe", "/c"] if on_windows else []) + ["echo", "abc"],
+            StdOutCapture)
 
     loop = asyncio.get_event_loop()
     result = loop.run_until_complete(main())
-    eq_(result["stdout"], "abc\n")
+    eq_(result["stdout"], "abc" + os.linesep)
 
 
 @known_failure_osx
```
yarikoptic, line 166: per comment below

```suggestion
# Both Windows and OSX suffer from  NotImplementedError: object proxy must define __reduce_ex__()
# see https://github.com/GrahamDumpleton/wrapt/issues/102
@known_failure_osx
@known_failure_windows
```
or alike

datalad/tests/test_nonasyncrunner.py
```diff
@@ -15,14 +15,29 @@
...
+    StdOutCapture,
+    WitlessProtocol,
+    WitlessRunner,
+)
+from datalad.nonasyncrunner import (
+    _ReaderThread,
+    run_command,
+)
+
+from datalad.utils import on_windows
```
yarikoptic, line 38: at some point we should really agree on whether we use relative or absolute imports, and whether exclusively or "mixed" (e.g. absolute generally, and relative within tests of direct imports of "one level upstairs" thus making it easy to move needed code without requiring to rewrite all the imports)

yarikoptic (commented):
I would have left `test_popen_invocation` alone and not overfit the skip for Windows -- proposed a suggestion

datalad/tests/test_nonasyncrunner.py
bpoldrack: > thus making it easy to move needed code without requiring to rewrite all the imports

You use an IDE - couldn't be easier ;-)

bpoldrack (approved):

datalad/tests/test_nonasyncrunner.py
yarikoptic: Well, might then with teach git to recognize IDE moved code as a "move" then too (file content gets changed, not an exact/clean move). Use of absolute import for such use cases provides no advantage whatsoever AFAIK, or what is it?

mih: I think we agreed, and it was absolute and not mixed. I am working towards this goal for a long time. I change every code that I touch.

re code moving: The moving that I am doing is mostly between packages, and relative imports are not helping with that.

datalad/tests/test_nonasyncrunner.py
mih: OK, done in this fashion! Thx!

datalad/tests/test_nonasyncrunner.py
mih: Done.

mih:
I canceled one set of appveyor runs, given the trivial nature of the last change.

PR #5902, opened 2021-08-16
Support username, password and port specification in RIA URLs

### LINKED ISSUE #5475: create-sibling-ria changes ssh username
#### What is the problem?
When trying to create a ria sibling the user name gets switched from the name in the ssh url to the current user name

```
[rbc@cubic-login1 bidsdatasets]$ datalad create-sibling-ria \
    --name pmacs-ria ria+ssh://user2@scisub.pmacs.upenn.edu:/project/RBC/BIDS/CCNP
rbc@scisub.pmacs.upenn.edu's password:   
```

0.14.0
No luck on create-sibling-ria :(

Is there some place I need to specify the user name for the remote server? It's replacing the user in the connection string with the name of the current user that's running the commands.

mih:
Previously only the hostname was retained for performing, e.g. SSH-based
logins, making it required to use an appropriate SSH config.

While not using such properties in RIA URLs is advisable, silently
stripping such information is a bug. Moreover, other mechanisms such
as HTTP (or a future FTP support) may not have equivalent configuration
analogs. Using such a core helper as `verify_ria_url()` to silently
remove information seems inappropriate.

A more comprehensive test is included.

Fixes #5475

### DIFF
--- a/datalad/customremotes/ria_utils.py
+++ b/datalad/customremotes/ria_utils.py
@@ -79,6 +79,8 @@ def verify_ria_url(url, cfg):
     -------
     tuple
       (host, base-path, rewritten url)
+      `host` is not just a hostname, but may also contain username, password,
+      and port, if specified in a given URL.
     """
     from datalad.config import rewrite_url
     from datalad.support.network import URL
@@ -99,7 +101,16 @@ def verify_ria_url(url, cfg):
                          "Supported: ssh, file, http(s)" %
                          protocol)
 
-    return url_ri.hostname if protocol != 'file' else None, \
+    host = '{user}{pdelim}{passwd}{udelim}{hostname}{portdelim}{port}'.format(
+        user=url_ri.username or '',
+        pdelim=':' if url_ri.password else '',
+        passwd=url_ri.password or '',
+        udelim='@' if url_ri.username else '',
+        hostname=url_ri.hostname or '',
+        portdelim=':' if url_ri.port else '',
+        port=url_ri.port or '',
+    )
+    return host if protocol != 'file' else None, \
         url_ri.path if url_ri.path else '/', \
         url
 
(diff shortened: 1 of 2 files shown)


datalad/customremotes/ria_utils.py
```diff
@@ -99,7 +101,16 @@ def verify_ria_url(url, cfg):
...
-    return url_ri.hostname if protocol != 'file' else None, \
+    host = '{user}{pdelim}{passwd}{udelim}{hostname}{portdelim}{port}'.format(
+        user=url_ri.username or '',
+        pdelim=':' if url_ri.password else '',
+        passwd=url_ri.password or '',
+        udelim='@' if url_ri.username else '',
+        hostname=url_ri.hostname or '',
+        portdelim=':' if url_ri.port else '',
+        port=url_ri.port or '',
+    )
```
yarikoptic, line 112: not familiar with RIA enough and can't grasp from the docstring -- what is the purpose/use of this returned entry?
it seems no longer a URL (no `protocol://` prefix) and also might contain sensitive information (`password`) thus should probably not be logged anywhere  (although I just now realized that we might be logging such URLs left and right) and should not be served as some kind of a "annex remote description".

is that a `host` to ssh into? (I guess not since IIRC you can't even add `:port` not to say password like that)

bpoldrack: > is that a host to ssh into?

If it's SSH, then yes. Will be passed to our SSH machinery in that case. 

Overall it depends: It's the thing the `base_path` of the RIA store is referring to.

bpoldrack (approved):
Actually, I was planning to use the parsed URL within ORA instead, but I'm fine with that change - it's the easier quick fix and the added test helps. Also doesn't stand in the way of overall RF'ing plans - so: Thanks much!

datalad/customremotes/ria_utils.py
mih: > I guess not since IIRC you can't even add :port not to say password like that

Thx @yarikoptic for making this point. I think it is better to switch to URL style destination description right away to actually make this possible. I will push an update.

bpoldrack (approved):

mih:
Thx for the review @bpoldrack

Issue #5808, opened 2021-07-22
Unconditional use of `annex.retry` turns error reporting into a mess


mih:
Example:

```
git annex copy --to expdir --fast --json --json-error-messages -c annex.retry=3
{"command":"copy","note":"to expdir...","success":false,"input":["123"],"key":"MD5E-s4--ba1f2511fc30423bdbb183fe33f3dd0f","error-messages":["  remote is configured with exporttree=yes; use `git-annex export` to store content on it","  remote is configured with exporttree=yes; use `git-annex export` to store content on it","  remote is configured with exporttree=yes; use `git-annex export` to store content on it","  This could have failed because --fast is enabled."],"file":"123"}
git-annex: copy: 1 failed
```

The command execution fails due to a setup issue (#3127). However, we unconditionally configure any call to `git-annex` to have `annex.retry` set (and we have 3 by default). From the documentation one can guess that this was only meant to be used for data transfer:

```
On transfer failure, annex.retry (sans "datalad.") controls the number of times that git-annex retries. DataLad will call git-annex with annex.retry set to the value here unless the annex.retry is explicitly configured
```

But is presently used for ANY `git-annex` call, unless `annex.retry` is set directly already.

This is not only a UX issue due to error message inflation. Trying 3 times for non data transport related errors is likely also having a performance impact.

This needs to be RF'ed to take `annex.retry` out of `_annex_common_options` and turn into targeted usage for calls where this is actually meaningful.

adswa:
I feel like I don't have enough background information to tackle this. Can we established for which command a retry would be meaningful? The only one I can think of is ``get``, but I am sure there are more to be considered.

mih:
I think `get` is the primary one. `copy` and `move` maybe make sense. Doing it for `sync` is already nightmare-ish. I am not sure if git-annex even considers it for anything else.

In general I would just vote to turn it off by default. If people want to protect against flaky behavior, they can turn it on for their use cases. Masking a wide range of errors like this has no net benefit IMHO, and can have actual negative impact (think: hammering services DDOS style because of parallelization times retries).

From my POV adding it to the `get` docs as a supported tuning option and switching the default to off would be the best change.

yarikoptic:
[...]

Issue #5903, opened 2021-08-16
AnnexRepo.copy_to() cannot accept options


mih:
If given `options` must be a list, due to this handling:

```
options = options[:] if options else []
```

however, if it is a list, it will crash due to this follow-up handling:

```
annex_options.extend(split_cmdline(options))
```

This is a minor issue, because this method is only used in tests, and was deprecatsed in Apr 2021 already. Waiting for removal is likely the best option.

adswa:
> This is a minor issue, because this method is only used in tests, and was deprecatsed in Apr 2021 already. Waiting for removal is likely the best option.

@mih, with this being deprecated in the ``0.14`` series, can this function be removed in ``0.16``?  The commit 62c01cce1ba7fb12f8c576c0a51fdab091a4beba that deprecated this function says in the commit message that it can be removed together with ``publish``, which has been moved to ``datalad-deprecated`` with ``0.15``

mih:
Yes! Thx for spotting that!

adswa:
In ripping it out and adjusting the tests that use it, I think I found a missing piece in ``push``. Some tests rely on ``copy_to`` to only transfer annexed contents to a remote. I couldn't find a parametrization for ``push`` that would allow for this (i.e., only transfer data and not push Git branches). Maybe ``push`` needs a ``--data only`` mode or a ``--what`` parameter similar to the new ``drop``?

mih:
I agree with the assessment of @adswa . Given that https://github.com/datalad/datalad/issues/6417 exists as a dedicated issue, and there rest is documented (but simply not acted on), I think we can close this.

PR #5904, opened 2021-08-16
Limit use of default annex.retry=3 to `get` and `copy` operations

mih:
As discussed in gh-5808 the unconditional use can lead to problematic
error reporting. @yarikoptic's verdict was to maintain the behavior, but
limit it to data transfer commands.

This change implements this move and enables this datalad-specific
default only for `get` and `copy` operations. The latter is not
implemented in `AnnexRepo.copy_to()` because this method is only used
internally in tests, partially broken (gh-5903), and was deprecated in
Apr 2021. Instead, the change is implemented directly in `push`.

Fixes datalad/datalad#5808

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -301,12 +301,6 @@ class AnnexRepo(GitRepo, RepoInterface):
         if self._ALLOW_LOCAL_URLS:
             self._allow_local_urls()
 
-        if config.get("annex.retry") is None:
-            self._annex_common_options.extend(
-                ["-c",
-                 "annex.retry={}".format(
-                     config.obtain("datalad.annex.retry"))])
-
         # will be evaluated lazily
         self._n_auto_jobs = None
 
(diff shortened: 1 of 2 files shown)


datalad/support/annexrepo.py
```diff
@@ -1382,6 +1382,12 @@ def get(self, files, remote=None, options=None, jobs=None, key=False):
         """
         options = options[:] if options else []
 
+        if self.config.get("annex.retry") is None:
+            options.extend(
+                ["-c",
+                 "annex.retry={}".format(
+                     self.config.obtain("datalad.annex.retry"))])
+
```
bpoldrack, line 1390: Hmm - could it be you forgot to remove the respective part in `__init__` L304 or did I get the intention wrong?

mih: LOL, yes. thx!

mih: I pushed an update that fixes this.

PR #5906, opened 2021-08-17
Python API design proposal

bpoldrack:
Ease discussion by making use of threaded review functionality. Simply
copied the opening post of the issue into a document that is supposed to
become a well-defined design proposal, given we reach some agreement on
the goals.

I also included the originally separate draft PR that is meant to shape
some illustration of the where to go, introducing a class `VCS` and an
attribute `Dataset._vcs` so far.

### DIFF
--- a/datalad/distribution/dataset.py
+++ b/datalad/distribution/dataset.py
@@ -22,12 +22,8 @@ import wrapt
 
 from datalad import cfg
 from datalad.config import ConfigManager
-from datalad.core.local.repo import repo_from_path
 from datalad.support.annexrepo import AnnexRepo
 from datalad.support.constraints import Constraint
-# DueCredit
-from datalad.support.due import due
-from datalad.support.due_utils import duecredit_dataset
 from datalad.support.exceptions import (
     NoDatasetFound,
 )
@@ -52,7 +48,8 @@ from datalad.utils import (
     PurePath,
     ensure_list,
 )
-
+from datalad.interface.vcs import VCS
+from datalad.dochelpers import borrowdoc
 
 lgr = logging.getLogger('datalad.dataset')
 lgr.log(5, "Importing dataset")
@@ -142,10 +139,10 @@ class Dataset(object, metaclass=PathBasedFlyweight):
         if isinstance(path, ut.PurePath):
             path = str(path)
         self._path = path
-        self._repo = None
         self._id = None
         self._cfg = None
         self._cfg_bound = None
+        self._vcs = VCS(self.pathobj)
 
     @property
     def pathobj(self):
@@ -210,13 +207,10 @@ class Dataset(object, metaclass=PathBasedFlyweight):
         return super(Dataset, self).__getattribute__(attr)
 
     def close(self):
-        """Perform operations which would close any possible process using this Dataset
+        """Perform operations which would close any possible process using this
+        Dataset
         """
-        repo = self._repo
-        self._repo = None
-        if repo:
-            # might take care about lingering batched processes etc
-            del repo
+        self._vcs.close()
 
     @property
     def path(self):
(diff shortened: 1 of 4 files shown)


bpoldrack:
@mih said (https://github.com/datalad/datalad/pull/5797#pullrequestreview-730177534):

> I am not convinced that a single VCS class works nicely.
> 
> I am feeling pretty strong about neither fiddling with code under distribution/ nor interface/ and certainly not introducing new stuff in there.
>
> I feel that the VCS class and some other aspects of this PR overlap with the starting point that was made in datalad/dataset/ (intentionally not under core/ at the time).
>
> Rather than just starting from scratch it would be useful to identify why this existing code cannot be used. This code also identifies the within *Repo dependencies to make Git calls via class instances. I consider this a useful approach to achieve an incremental transition.

datalad/distribution/dataset.py
```diff
@@ -451,6 +385,52 @@ def subds_contains_path(ds, path):
...
+
+    @borrowdoc(AnnexRepo)
+    def call_annex_items_(self, args, files=None, sep=None):
+
+        return self._vcs.call_annex_items_(args, files=None, sep=None)
+
+    @borrowdoc(AnnexRepo)
+    def call_annex_oneline(self, args, files=None):
+
+        return self._vcs.call_annex_oneline(args, files=None)
```
bpoldrack, line 432: @mih said:

> All call_annex_*() methods would fail if the underlying VCS instance points to a GitRepo.

bpoldrack: Yes, this draft so far is meant to illustrate what I wrote in the issue in principle, not be a full implementation yet. I'd like to first get some agreement on the goal. However - yes: At the Dataset level, those would have to be properly guarded.

bpoldrack:
@yarikoptic, too, already had a comment:

quick one:
- documentation is nice, but since we are in Python land, what is public **is** instrumented through reachable `.public` attributes, hence it is IMHO not just an issue of documentation, thus:
- :+1: on "introduce a "non-public" `Dataset` property `_vcs`" 
- re `2.` on architecture/testing: well -- `*Repo` classes in principle provide that level of abstraction. It just due to their organic development through the ages, and us never really unifying interfaces and never providing helpers to "mock them out" for unit-testing of higher level -- we got where we are.  IMHO the addition of `call_` methods even further diluted the boundary of that layer leaking all details to higher levels, and complicating those higher-levels unittesting without then mocking out git/git-annex behavior (should be the duty at `*Repo` levels).  I do not see how levels separation is possible as long as those `call_` are used/leaked to higher levels. So if we are to go for `2.`, even if `call_*` would stay public, they should actually not be used directly by those higher (`Dataset.`-level) interfaces IMHO.
- bringing those together -- it feels that we would like some `*VCS`s to replace `*Repo`s , while taking into  account `*Repo`s experience, but providing a more consistent re-implementation of interfaces to git/annex.

bpoldrack:
@yarikoptic 
> - documentation is nice, but since we are in Python land, what is public **is** instrumented through reachable `.public` attributes, hence it is IMHO not just an issue of documentation, thus:
> - :+1: on "introduce a "non-public" `Dataset` property `_vcs`" 

As I mentioned: These are actually different notions of "public". Everything is "reachable". Just not from any class in any module. However, how is `from datalad.support.annexrepo import AnnexRepo` anyhow less publicly reachable just because there's no "public" attribute in `Dataset`?  The fact that `datalad.api.Dataset` is the "entry point" is equally arbitrary and mere convention (by documentation). There's nothing to declare the starting point(s) for the supposed chain of "reachable through `.public`".

I used `_vcs` to not encourage it's usage by third party code, while at the same time documenting that this layer actually promises deprecation cycles for its "public" attributes. Compromise. :)


> *Repo classes in principle provide that level of abstraction.

Not enough in the long run IMO. We may well need to at least distinguish the notions of "a repository" from the notion of "a checkout" or "a worktree" for example. Ultimately it's not so much about whether or to what extend the Repo classes are fine. I think they can't stay the way they are and we need to be able to easily introduce additional layers in between `VCS`/datalad-commands and `*Repo`. That's near impossible in reality ATM and that's what I want to get out of.

I don't agree on the `call_*` aspect. I think it's useful to have that available in the python interface to integrate datalad calls and calls to git/annex consistently (just like we point to git/git-annex for the CLI for some tasks). The fact that they would go rather directly deep down in terms of abstraction is kinda fine with me for a very limited set of functions, that comes with no further datalad magic. But I agree: It should be the thing to use at high level only if there really is nothing else. And any usage points to `VCS` being incomplete.

mih:
Quick comment on the form of the proposal in `docs/source/design/python_interface.rst`. I think this is not a good format. It mixes the expressions of feelings and opinions with the pieces of a design proposal. That latter, arguably the critical pieces, are somewhat sprinkled across the text.

I seems as if the text touches upon two different aspects: The actual interface specification, and how to get it implemented. It may make sense to discuss these together (although it comes with the potential of distraction from the key ideas). However, I do think these aspects should be clearly separated.

It would be useful to reformat the document into clearly separated, concise statements of facts that can be commented on (rather than long blurbs). In the present form it will be hard to come to a result other than accept/reject in full, IMHO.

bpoldrack:
Agreed. Will try to structure it better.

mih:
> Agreed. Will try to structure it better.

Is there an ETA for the update? I do not expect much contribution until then.

mih:
Let's close this for now. There is no value in it being open before an update.

Issue #5764, opened 2021-06-25
Confusing error message


mih:
```
% datalad create-sibling-gitlab --site gitlab.com 
create_sibling_gitlab(error): . (dataset) [No project path specified, and no configuration to derive one]
```

This seems to suggest that a path must be specified (there is `PATH` as an argument in the docs). However it seems to want `--project` to be specified in reality.

PR #5907, opened 2021-08-17
Clarify error message re unspecified gitlab project

mih:
Issue
```
% datalad create-sibling-gitlab --site gitlab.com
create_sibling_gitlab(error): . (dataset) [No project path specified, and no configuration to derive one]

% datalad create-sibling-gitlab -h
Usage: datalad create-sibling-gitlab [-h] [--site SITENAME] [--project NAME/LOCATION]
                                     [--layout {hierarchy|collection|flat}] [--dataset DATASET] [-r] [-R LEVELS]
                                     [-s NAME] [--existing {skip|error|reconfigure}] [--access {http|ssh|ssh+http}]
                                     [--publish-depends SIBLINGNAME] [--description DESCRIPTION] [--dryrun]
                                     [PATH ...]
```

The error message is hinting at a missing PATH, but --project is
required instead.

This change standardizes on using the terms from the existing metavar
spec, in the docs and the error message to reduce the confusion:

```
create_sibling_gitlab(error): . (dataset) [No project name/location specified, and no configuration to derive one]
```

Fixes datalad/datalad#5764

### DIFF
--- a/datalad/distributed/create_sibling_gitlab.py
+++ b/datalad/distributed/create_sibling_gitlab.py
@@ -150,7 +150,7 @@ class CreateSiblingGitlab(Interface):
         project=Parameter(
             args=('--project',),
             metavar='NAME/LOCATION',
-            doc="""project path at the GitLab site. If a subdataset of the
+            doc="""project name/location at the GitLab site. If a subdataset of the
             reference dataset is processed, its project path is automatically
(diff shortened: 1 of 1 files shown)


mih:
OK, I complained about the message, I think it is better now. Will merge now-ish.

PR #5916, opened 2021-08-18
Provide a few more hints whenever user just runs `datalad`

yarikoptic:
Instead of 
```
$> datalad
error: too few arguments
usage: datalad [-l LEVEL] [--pbs-runner {condor}] [-C PATH] [--version] [--dbg] [--idbg] [-c KEY=VALUE]
               [-f {default,json,json_pp,tailored,'<template>'}]
               [--report-status {success,failure,ok,notneeded,impossible,error}]
               [--report-type {dataset,file}] [--on-failure {ignore,continue,stop}] [--cmd]
               command [command-opts]
```

and being puzzled on "what next" until they realize they need to run apparently non-listed above`--help`:

```
$> datalad
error: too few arguments, run with --help to get more information.
visit: https://www.datalad.org https://docs.datalad.org https://handbook.datalad.org
usage: datalad [-l LEVEL] [--pbs-runner {condor}] [-C PATH] [--version] [--dbg] [--idbg] [-c KEY=VALUE]
               [-f {default,json,json_pp,tailored,'<template>'}]
               [--report-status {success,failure,ok,notneeded,impossible,error}]
               [--report-type {dataset,file}] [--on-failure {ignore,continue,stop}] [--cmd] [-h]
               command [command-opts]
```

### DIFF
--- a/datalad/cmdline/helpers.py
+++ b/datalad/cmdline/helpers.py
@@ -433,7 +433,7 @@ def _maybe_get_single_subparser(cmdlineargs, parser, interface_groups,
         if not unparsed_args and getattr(parsed_args, 'version', None):
             parsed_args.version()  # will exit with 0
         if not (completing or unparsed_args):
-            fail_handler(parser, msg="too few arguments", exit_code=2)
+            fail_handler(parser, msg="too few arguments, run with --help to get more information.", exit_code=2)
         lgr.debug("Command line args 1st pass for DataLad %s. Parsed: %s Unparsed: %s",
                   __full_version__, parsed_args, unparsed_args)
     except Exception as exc:
--- a/datalad/cmdline/helpers.py
+++ b/datalad/cmdline/helpers.py
@@ -433,7 +433,10 @@ def _maybe_get_single_subparser(cmdlineargs, parser, interface_groups,
         if not unparsed_args and getattr(parsed_args, 'version', None):
             parsed_args.version()  # will exit with 0
         if not (completing or unparsed_args):
-            fail_handler(parser, msg="too few arguments, run with --help to get more information.", exit_code=2)
+            fail_handler(
+                parser,
+                msg="too few arguments, run with --help or visit https://handbook.datalad.org",
+                exit_code=2)
         lgr.debug("Command line args 1st pass for DataLad %s. Parsed: %s Unparsed: %s",
                   __full_version__, parsed_args, unparsed_args)
     except Exception as exc:
--- a/datalad/cmdline/helpers.py
+++ b/datalad/cmdline/helpers.py
@@ -591,6 +591,7 @@ def fail_with_short_help(parser=None,
         out.write("error: %s\n" % msg)
     if not known:
         if parser:
+            parser_add_common_opt(parser, 'help')
             # just to appear in print_usage also consistent with --help output
             parser.add_argument("command [command-opts]")
             parser.print_usage(file=out)


datalad/cmdline/helpers.py
```diff
@@ -433,7 +433,11 @@ def _maybe_get_single_subparser(cmdlineargs, parser, interface_groups,
         if not unparsed_args and getattr(parsed_args, 'version', None):
             parsed_args.version()  # will exit with 0
         if not (completing or unparsed_args):
-            fail_handler(parser, msg="too few arguments", exit_code=2)
+            fail_handler(
+                parser,
+                msg="too few arguments, run with --help to get more information.\n"
+                    "visit: https://www.datalad.org https://docs.datalad.org https://handbook.datalad.org",
```
mih, line 439: This is a user-facing message intended to aid datalad usage. I think it is important to keep it as short as possible.

https://www.datalad.org has infos on what datalad is (at best), users that see this message already have it.

https://docs.datalad.org has almost exclusively technical information, users do not need that, when they still have no idea how to run it.

That let's me conclude that only the last pointer should retained, and the first two stripped for clarity and conciseness. The handbook website does have links to the project website and the technical docs on the frontpage too.

mih (changes requested):
Great idea! Thx! I would very much prefer to make it more concise though.

datalad/cmdline/helpers.py
yarikoptic: Ok, I will drop 2nd commit

The idea was to provide extra information (potentially long but fitting into 80x40) for completely unparametrized invocation `datalad` (meaning that user really didn't know how to use datalad) but I guess it should indeed not be in this handler which might be triggered in other situations too

yarikoptic: And for such situations IMHO it would be useful to provide useful URLs which would describe what it is, have commands reference and provide tutorials etc

yarikoptic:
FWIW, dropped 2nd commit, so we will only inform about `--help` and also add `[-h]` to usage summary options

datalad/cmdline/helpers.py
mih: Ah, sorry, a misunderstanding! I am very much in favor of giving a URL -- just not 3 URLs at once.

datalad/cmdline/helpers.py
```diff
@@ -433,7 +433,7 @@ def _maybe_get_single_subparser(cmdlineargs, parser, interface_groups,
         if not unparsed_args and getattr(parsed_args, 'version', None):
             parsed_args.version()  # will exit with 0
         if not (completing or unparsed_args):
-            fail_handler(parser, msg="too few arguments", exit_code=2)
+            fail_handler(parser, msg="too few arguments, run with --help to get more information.", exit_code=2)
```
mih, line 436:
```suggestion
            fail_handler(
                parser,
                msg="too few arguments, run with --help or visit https://handbook.datalad.org",
                exit_code=2)
```

What about this? Fits in 80 chars on a single line. It has the pointer to `--help` and a URL. Giving the explicit mention of "help" and "handbook" I think we can get away without "to get more information", while still being fairly clear about why one would want to visit this page.

mih:
```sh
% datalad
error: too few arguments, run with --help or visit https://handbook.datalad.org
usage: datalad [-l LEVEL] [--pbs-runner {condor}] [-C PATH] [--version]
               [--dbg] [--idbg] [-c KEY=VALUE]
               [-f {default,json,json_pp,tailored,'<template>'}]
               [--report-status {success,failure,ok,notneeded,impossible,error}]
               [--report-type {dataset,file}]
               [--on-failure {ignore,continue,stop}] [--cmd] [-h]
               command [command-opts]
```

yarikoptic: Ok, let's go with this

yarikoptic:
appveyor fail is legit this time, but unrelated to this PR

```

FAIL: datalad.support.tests.test_parallel.test_gracefull_death
  File "/home/appveyor/dlvenv/lib/python3.8/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/appveyor/dlvenv/lib/python3.8/site-packages/datalad/tests/utils.py", line 266, in _wrap_skip_if
    return func(*args, **kwargs)
  File "/home/appveyor/dlvenv/lib/python3.8/site-packages/datalad/support/tests/test_parallel.py", line 224, in test_gracefull_death
    assert_equal(sorted(inner()), [0, 1])
AssertionError: Lists differ: [0, 2] != [0, 1]
First differing element 1:
2
1
- [0, 2]
?     ^
+ [0, 1]
?     ^
```
I fear I just need to disable those tests :-/

Let's proceed

Issue #4697, opened 2020-07-13
Q: can "datalad siblings add" just add a remote as  autoenabled type=git special remote?


yarikoptic:
I kept doing that dance manually, the one we already do I believe in `create-sibling`, smth like 

```shell
git remote add --fetch datasets.datalad.org_ http://datasets.datalad.org/labs/mvdm/Kilosort2_results/.git/
git annex initremote datasets.datalad.org type=git location=http://datasets.datalad.org/labs/mvdm/Kilosort2_results/.git/ autoenable=true 
git remote remove datasets.datalad.org_
```

I think it should be `datalad siblings --as-common-datasrc` but it seems to not yet working out that use case correctly:

```shell
tmp > datalad install https://github.com/datalad-datasets/hcp_wm_preprocessed                              
[INFO   ] Remote origin not usable by git-annex; setting annex-ignore                                            
install(ok): /tmp/hcp_wm_preprocessed (dataset)                                                                  
datalad install https://github.com/datalad-datasets/hcp_wm_preprocessed  3.76s user 2.25s system 80% cpu 7.450 total

/tmp > cd hcp_wm_preprocessed 

/tmp/hcp_wm_preprocessed > git remote
datalad
origin

/tmp/hcp_wm_preprocessed > datalad siblings
.: here(+) [git]
.: origin(-) [https://github.com/datalad-datasets/hcp_wm_preprocessed (git)]
.: datalad(+) [datalad]

/tmp/hcp_wm_preprocessed > git show git-annex:remote.log
6239b9e6-0b94-48f9-b9c1-15c5581d2524 autoenable=true encryption=none externaltype=datalad name=datalad type=external timestamp=1593790927.331819559s

/tmp/hcp_wm_preprocessed > datalad siblings --url https://datasets.datalad.org/.git --as-common-datasrc datasets.datalad.org add  
CommandError: 'git-annex initremote -c annex.dotfiles=true -c annex.retry=3 datasets.datalad.org type=git location=https://datasets.datalad.org/.git autoenable=true' failed with exitcode 1 under /tmp/hcp_wm_preprocessed
git-annex: There is already a remote named "datasets.datalad.org"

/tmp/hcp_wm_preprocessed > datalad siblings             
.: here(+) [git]
.: datalad(+) [datalad]
.: origin(-) [https://github.com/datalad-datasets/hcp_wm_preprocessed (git)]
.: datasets.datalad.org(+) [https://datasets.datalad.org/.git (git)]

[...]

Issue #5914, opened 2021-08-18
`siblings configure` attempts to configure non-existing remotes, and leaves things in a limbo state


mih:
```
% datalad siblings configure -s mike
[INFO   ] Could not enable annex remote mike. This is expected if mike is a pure Git remote, or happens if it is not accessible. 
.: mike(?) [git]
```

where `mike` is a random string with no relationship to any existing remote.

I would have expected an error saying "Unkown remote, use 'add' to configure a new remote" or something like that. Instead, I got

```
[remote "mike"]
        fetch = +refs/heads/*:refs/remotes/mike/*
```

barfed into `.git/config`.

mih:
So, it seems to be doing that ... on purpose!?

```py
    if name != 'here':
        # do all configure steps that are not meaningful for the 'here' sibling
        # AKA the local repo
        if name not in known_remotes and url:
            # this remote is fresh: make it known
            # just minimalistic name and URL, the rest is coming from `configure`
            ds_repo.add_remote(name, url)
            known_remotes.append(name)
 
```

If `configure()` does not recognize the label to belong to a remote, it just adds it -- bypassing all processing that `add` would do. The comment seems to indicate that this is actually a C&P result from `add`.

So either `add` is superfluous (which it is not, lots of checks), or `configure` must call `add` in this case (which it should not, unless `add` is removed from the API), or it should produce an error.

I think the latter is the sane choice for a fix. Everything else is more of a re-design.

mih:
Hours later I slowly get the impression that this is a conceptual weakness of the `add/configure` separation. When there is *any* `remote.<name>...` config setting, a remote `<name>` is considered to be "known" and `add` refuses to work with it, referring to `configure`. `remote.<name>.crazy-made-up-stuff` is perfectly sufficient to trigger this situation.

This seems to suggest that it would be better to move most functionality from `add` into `configure`, and acknowledge the fact that a "known" remote could be substantially different from a "working/sensible" remote setup.

PR #5913, opened 2021-08-18
Fix argument-spec for `siblings` and improve usage synopsis

mih:
`None` was declared as supported `action`, but if given, would cause a crash.

Synopsis went from

```
Usage: datalad siblings [-h] [-d DATASET] [-s NAME] [--url [URL]] [--pushurl PUSHURL] [-D DESCRIPTION] [--fetch]
                        [--as-common-datasrc NAME] [--publish-depends SIBLINGNAME] [--publish-by-default REFSPEC]
                        [--annex-wanted EXPR] [--annex-required EXPR] [--annex-group EXPR] [--annex-groupwanted EXPR]
                        [--inherit] [--no-annex-info] [-r] [-R LEVELS] [--version]
                        [ACTION]
```

to

```
Usage: datalad siblings [-h] [-d DATASET] [-s NAME] [--url [URL]] [--pushurl PUSHURL] [-D DESCRIPTION] [--fetch]
                        [--as-common-datasrc NAME] [--publish-depends SIBLINGNAME] [--publish-by-default REFSPEC]
                        [--annex-wanted EXPR] [--annex-required EXPR] [--annex-group EXPR] [--annex-groupwanted EXPR]
                        [--inherit] [--no-annex-info] [-r] [-R LEVELS] [--version]
                        [{query|add|remove|configure|enable}]
```

Fixes datalad/datalad#5912

PR #1500, opened 2017-05-08
New `siblings` command

mih:
Naming is analog to the `subdatasets` command that we already have. The purpose of this new command is to concentrate all code for sibling registration and meta data manipulation of siblings into a single command (set/change name, description, wanted policies, ....). Moreover, I would propose to treat "here" as another sibling and have the very same command also manipulate a standalone dataset.

In essence, all this command does are actions bound to a single dataset, plus a for-each-style recursive iteration of a dataset tree. For the latter we need to support a templating mechanism to e.g. transform a base URL to match a particular subdataset (so nothing new or fancy).

List of related, unaddressed issues and previous discussions -- not necessarily TODO items for this PR, but the changes here should make those simpler to implement:

- rewrite description #1484 
- sibling/dataset configuration via templates #1462 
- wanted policy setup #925 (related #921)
- registration via datalad RIs #1438 
- annex-only sibling tracking/update #1363
- figure out if we want to treat "here" as yet another sibling and have the very same command also manipulate a standalone dataset (e.g. for description, wanted, ....). If so, how to indicate this best? `--name` no name could be `here`, but the we would need something for "give me all info about all remotes".
- check (maybe via coverage) which parts of the legacy `add_sibling` became obsolete and gut the code accordingly
- insufficient documentation #1017 
- scary low-level logs #1519 

Actual TODOs:

- [x] generalize `add_sibling` to `siblings` command #1235 - [ ] 
- [x] `siblings` is new-style generator
- [x] `siblings` can remove a sibling (`remote remove`) #1483
- [x] all internal use of `add_sibling` is via `siblings`
- [x] tests for `siblings remove`

PR #5921, opened 2021-08-19
Some love and fixes for `siblings`

### LINKED ISSUE #5626: `siblings` crash when running in subdir with `-d .`
```
datalad siblings add -d . --name gin-update --pushurl git@gin.g-node.org:/studyforrest/aligned-mri.git --url https://gin.g-node.org/studyforrest/aligned-mri.git --as-common-datasrc gin
```

The call is invalid, but it should not crash.

mih:
The diff looks massive, but looking at individual commits should tell a simpler story.

Key changes:

- coverage boost from 77% to 90% (see below)
- `--as-common-datasrc` working again, and tested from now (fixes: #4697)
- fixed `GitRepo.remove_remote()` to issue the expected exception when trying to remove a non-existing remote (also covered by a test now)
- fix crash with invalid `dataset` argument (fixes #5626)

Individual fixes could have been PR'ed against `master`. However, this whole thing sparks no joy and I am happy to have come to this point. I do not plan on further improvements. There is more to do, even after #5913 and #5915. Importantly #5914. The changes included here should make incremental improvements simpler. But I have to detach for now.

```
DATALAD_TESTS_SSH=1 python -m nose -s -v --with-coverage --cover-erase --cover-package datalad.distribution.siblings datalad.distribution.tests.test_siblings
...
Name                               Stmts   Miss  Cover   Missing
----------------------------------------------------------------
datalad/distribution/siblings.py     341     78    77%   225, 308-313, 351, 355, 371, 419-422, 508, 513-522, 525-532, 561-577, 593, 597-600, 623-624, 645-648, 661, 685, 700, 716, 726-731, 753-757, 760-764, 771-775, 781-807, 815-827, 834, 847, 865, 874
----------------------------------------------------------------
TOTAL                                341     78    77%
Ran 8 tests in 24.917s
```

```
Name                               Stmts   Miss  Cover   Missing
----------------------------------------------------------------
datalad/distribution/siblings.py     345     35    90%   520, 608, 634-635, 672, 696, 711, 722, 781-807, 815-827, 834, 847, 865, 874
----------------------------------------------------------------
TOTAL                                345     35    90%
Ran 10 tests in 26.504s
```

### DIFF
--- a/datalad/distribution/siblings.py
+++ b/datalad/distribution/siblings.py
@@ -178,7 +178,13 @@ class Siblings(Interface):
             args=("--fetch",),
             action="store_true",
             doc="""fetch the sibling after configuration"""),
-        as_common_datasrc=as_common_datasrc,
+        as_common_datasrc=Parameter(
+            args=("--as-common-datasrc",),
+            metavar='NAME',
+            doc="""configure a sibling as a common data source of the
+            dataset that can be automatically used by all consumers of the
+            dataset. The sibling must be a regular Git remote with a
+            configured HTTP(S) URL."""),
         publish_depends=publish_depends,
         publish_by_default=publish_by_default,
         annex_wanted=annex_wanted_opt,
@@ -393,6 +399,8 @@ def _add_remote(ds, repo, name, known_remotes, url, pushurl, as_common_datasrc,
             message=("sibling is already known: %s, use `configure` instead?", name),
             **res_kwargs)
         return
+    # XXX this check better be done in configure too
+    # see https://github.com/datalad/datalad/issues/5914
     if as_common_datasrc == name:
         raise ValueError('Sibling name ({}) and common data source name ({}) '
                          'can not be identical.'.format(name, as_common_datasrc))
@@ -568,7 +576,11 @@ def _configure_remote(
                 lgr.debug("Exception was: %s", exc_str(exc))
 
             if as_common_datasrc:
-                ri = RI(url)
+                # we need a fully configured remote here
+                # do not re-use `url`, but ask for the remote config
+                # that git-annex will use too
+                remote_url = repo.config.get(f'remote.{name}.url')
+                ri = RI(remote_url)
                 if isinstance(ri, URL) and ri.scheme in ('http', 'https'):
                     # XXX what if there is already a special remote
                     # of this name? Above check for remotes ignores special
@@ -581,12 +593,11 @@ def _configure_remote(
                         'initremote',
                         as_common_datasrc,
                         'type=git',
-                        'location={}'.format(url),
+                        'location={}'.format(remote_url),
                         'autoenable=true'])
                 else:
                     yield dict(
                         status='impossible',
-                        name=name,
                         message='cannot configure as a common data source, '
                                 'URL protocol is not http or https',
                         **result_props)
--- a/datalad/distribution/siblings.py
+++ b/datalad/distribution/siblings.py
@@ -246,18 +246,21 @@ class Siblings(Interface):
         # at the top-level and vice versa
         worker = action_worker_map[action]
 
-        dataset = require_dataset(
-            dataset, check_installed=False, purpose='configure sibling')
-        refds_path = dataset.path
+        ds = require_dataset(
+            dataset,
+            # it makes no sense to use this command without a dataset
+            check_installed=True,
+            purpose='configure sibling')
+        refds_path = ds.path
 
         res_kwargs = dict(refds=refds_path, logger=lgr)
 
-        ds_name = op.basename(dataset.path)
+        ds_name = op.basename(ds.path)
 
         # do not form single list of datasets (with recursion results) to
         # give fastest possible response, for the precise of a long-all
         # function call
-        ds = dataset
+
         # minimize expensive calls to .repo
         ds_repo = ds.repo
 
--- a/datalad/distribution/siblings.py
+++ b/datalad/distribution/siblings.py
@@ -286,8 +286,7 @@ class Siblings(Interface):
             get_annex_info=get_annex_info,
             res_kwargs=res_kwargs,
         )
-        for r in worker(**worker_kwargs):
-            yield r
+        yield from worker(**worker_kwargs)
         if not recursive:
             return
 
@@ -318,8 +317,7 @@ class Siblings(Interface):
                 url=subds_url,
                 pushurl=subds_pushurl,
             )
-            for r in worker(**worker_kwargs):
-                yield r
+            yield from worker(**worker_kwargs)
 
     @staticmethod
     def custom_result_renderer(res, **kwargs):
--- a/datalad/distribution/siblings.py
+++ b/datalad/distribution/siblings.py
@@ -12,71 +12,72 @@ __docformat__ = 'restructuredtext'
 
 
 import logging
-
 import os
 import os.path as op
-
 from urllib.parse import urlparse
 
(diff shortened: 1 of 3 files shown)


yarikoptic:
> * coverage boost from 77% to 90% (see below)

woohoo -- appreciated, better later than never ;)

datalad/distribution/siblings.py
```diff
@@ -239,30 +246,47 @@ def __call__(
         # at the top-level and vice versa
         worker = action_worker_map[action]
 
-        dataset = require_dataset(
-            dataset, check_installed=False, purpose='configure sibling')
-        refds_path = dataset.path
+        ds = require_dataset(
+            dataset,
+            # it makes no sense to use this command without a dataset
+            check_installed=True,
```
yarikoptic, line 252: FWIW, it was `=False` in the original PR https://github.com/datalad/datalad/pull/1500/files#diff-a45159f0547073e97cc9b38ee5d4fb94ecf3b6ec6d70250e5017d074d976deccR184 so not like it was set to `=False` for some explained in some later commit/PR reason, and I agree with the comment.

datalad/distribution/siblings.py
```diff
@@ -239,30 +246,47 @@ def __call__(
...
+            annex_wanted=annex_wanted,
+            annex_required=annex_required,
+            annex_group=annex_group,
+            annex_groupwanted=annex_groupwanted,
+            inherit=inherit,
+            get_annex_info=get_annex_info,
+            res_kwargs=res_kwargs,
+        )
+        for r in worker(**worker_kwargs):
             yield r
```
yarikoptic, line 290: if all workers are generators (as probably they should be, but didn't check), here and below could be just

```suggestion
        yield from worker(**worker_kwargs)
```

datalad/distribution/siblings.py
```diff
@@ -239,30 +246,47 @@ def __call__(
...
-                as_common_datasrc, publish_depends, publish_by_default,
-                annex_wanted, annex_required, annex_group, annex_groupwanted,
-                inherit, get_annex_info,
-                **res_kwargs):
+
+        # minimize expensive calls to .repo
+        ds_repo = ds.repo
+
+        # prepare parameterization package for all worker calls
+        worker_kwargs = dict(
```
yarikoptic, line 268: FWIW, I would have preferred to just identify common `worker_kwargs` and provide specific ones explicitly, instead of `.update()`ing them for the next invocation.  Rationale -- if later decided to re-use such kwargs dict for yet another worker invocation, extra care should be taken to overload all inplace modifications for subsequent invocations.  Without inplace modification
- no such care needs to be take
- makes it clear right here which options are common for top level and subdatasets

datalad/distribution/siblings.py
yarikoptic: While adding "love" may be start adding type annotations? pandora's box of importing typing is already open -- so there should be no extra run time penalty, smth like
```
diff --git a/datalad/distribution/siblings.py b/datalad/distribution/siblings.py
index f3129f31f..9d4dd9dbd 100644
--- a/datalad/distribution/siblings.py
+++ b/datalad/distribution/siblings.py
@@ -16,6 +16,7 @@ import logging
 import os
 import os.path as op
 
+from typing import Iterator
 from urllib.parse import urlparse
 
 from datalad.interface.base import Interface
@@ -338,7 +339,7 @@ def _add_remote(
         as_common_datasrc, publish_depends, publish_by_default,
         annex_wanted, annex_required, annex_group, annex_groupwanted,
         inherit, get_annex_info,
-        **res_kwargs):
+        **res_kwargs) -> Iterator[dict]:
     # TODO: allow for no url if 'inherit' and deduce from the super ds
     #       create-sibling already does it -- generalize/use
     #  Actually we could even inherit/deduce name from the super by checking
```
for those workers, and that should (AFAIK) immediately give an idea if all workers are providing Iterators (mypy though seems to not alert if I change here to `yield from` and do not annotate all workers)

datalad/distribution/siblings.py
```diff
@@ -558,7 +579,11 @@ def _configure_remote(
                 lgr.debug("Exception was: %s", exc_str(exc))
 
             if as_common_datasrc:
-                ri = RI(url)
+                # we need a fully configured remote here
+                # do not re-use `url`, but ask for the remote config
+                # that git-annex will use too
+                remote_url = repo.config.get(f'remote.{name}.url')
```
yarikoptic, line 585: to say the truth I am not a "strong user of --as-common-datasrc NAME" and its "directionality" here in relation to `--name`  its own `NAME` value and `--url`... the question is -- can remote `name` be a git annex special remote? if so -- then it most likely would not have a `url` (some like type `git` ones would)

yarikoptic (commented):
just some minor comments.

overall -- I welcome `ds.repo` -> `ds_repo` local bindings etc and expect us to become rocket fast ;)

datalad/distribution/siblings.py
mih: I know too little about typing to be the one doing that, sorry. Quickly glancing at the docs, it also tells me:

> class typing.Iterator(Iterable[T_co])
A generic version of collections.abc.Iterator.
> Deprecated since version 3.9: collections.abc.Iterator now supports []. See PEP 585 and Generic Alias Type.

So this may not be the right path, who knows. Anyways, I do not see myself doing it in here.

mih: That being said, yes, `yield from` works, will change.

datalad/distribution/siblings.py
mih: I know that this would be better, but it would require more substantial changes. This change only wants to aid code readability -- it does not make things worse than they were before. I do not claim a beauty price for this PR. As expressed elsewhere, I am perfectly fine with ripping `siblings` apart and rewrite it from scratch. However, this would need to involve a debate about "inherit" and "delayed super" that I do not have the juice for now.

datalad/distribution/siblings.py
mih: No, special remotes do not work. This is creating a `type=git` special remote, which requires a fully set up regular Git remote for `initremote` to work. If there is no URL, it will fail (and should).

The directionality is

```
siblings configure -s GITREMOTE --as-common-datasrc SEPCIALREMOTE
```

`SPECIALREMOTE` will be created based on `GITREMOTE`. You can also use `add` (see tests for both examples), in which case `GITREMOTE` will be created first (requires additional spec of a URL), and then `SPECIALREMOTE` will be created based on it.

mih:
Windows failure is real:

```
790ERROR: datalad.distribution.tests.test_siblings.test_arg_missing
791----------------------------------------------------------------------
...
799  File "C:\Python39-x64\lib\site-packages\datalad\distribution\tests\test_siblings.py", line 367, in test_arg_missing
800    assert_raises(
...
829ValueError: Port could not be cast to integer value as '\\DLTMP\\datalad_temp_mni6o9m4'
```

The `file://` URL implementation trick in the tests does not work on windows, of course. We make it conditional.

mih:
OK, all green. Test fixed, suggested changes (minus some, see resolved converations) implemented.

I am good now. Take it or leave it ;-)

bpoldrack (approved):
LGTM. I'll take it. :)

datalad/distribution/siblings.py
yarikoptic: I had in mind just this: https://github.com/datalad/datalad/pull/5924

PR #5923, opened 2021-08-19
Add isort config to match current convention + run isort via pre-commit (if configured)

### LINKED ISSUE #5922: Configure pre-commit for isort
following https://github.com/datalad/datalad/pull/5909/files#diff-fce12c5612668baada175b52f097e84518731a399bf6c9d518a71d3ef7758998R47 to run it as `python -m isort -m3 -fgw 2 -tc <filename>` on the changed files.  Should be "optional" (i.e. not enforced yet/no CI -- just support for the tool if someone installs and configures pre-commit locally)

Since we have not really decided on feasible timeline for `master` release, I think it might be worth doing it for `maint` at this time point

jwodder:
Closes #5922.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,5 @@
+repos:
+  - repo: https://github.com/PyCQA/isort
+    rev: 5.9.3
+    hooks:
+      - id: isort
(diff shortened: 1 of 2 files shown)


yarikoptic:
awesome -- thank you @jwodder , I will test it out.  I wonder -- why `patch` and not `internal` label?

yarikoptic:
appveyor -- a Mac, As this change doesn't mandate anyone to anything, I will proceed with merge

PR #5924, opened 2021-08-20
RF: minor - define only common worker kwargs in a reused dict

yarikoptic:
Follow up to https://github.com/datalad/datalad/pull/5921/files#r692337399
discussion/suggestion


mih:
Ha, ok. That looks much more simple than what I read from your original comment ;-)

PR #5927, opened 2021-08-25
0.15.0 Changelog

mih:
The proposed format is a mix of what Kyle did and what `auto` does.

Please do not argue with individual phrasing and other aspects, simply suggest a concrete change.

FTR: I will get officially angry should any automated procedure replace this work of art! ;-)

The release subtitle is non-negotiable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,264 @@
+# ? (?) --  We miss you Kyle!
+
+#### Enhancements and new features
+
+- Command execution is now performed by a new `Runner` implementation that is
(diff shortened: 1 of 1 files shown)


yarikoptic:
thank you @mih !

yarikoptic:
Should we may be merge this PR and release an `rc1` so it could be tested etc... might make it easier also to track later what new changelog entries to add?

mih:
I don't mind, but I don't need the merge for better tracking.

mih:
This is now final from my POV. All remaining work for 0.15 is developer-focused or internal and no candidate for this changelog.

PR #5599, opened 2021-04-23
RF copy_file() with a cleaner (cached) Repo method access

mih:
This is a first example of an approximation of an idea that I sketched in https://github.com/datalad/datalad/issues/5598

Instead of using a complicated beast made of dicts and more dicts, I RF'ed all relevant Repo method access into an Interface helper that uses `lru_cache` and other means to minimize the actual interaction with the underlying `Repo` instance.

It does that with awareness of what type of access is being made and only works within the context of `copy_file()`.

I mostly did this to see if these task-focused/tailored API classes can make sense, and de-convolute primary command logic from performance aspects.

Happy to get feedback on the approach.

PR #5933, opened 2021-08-26
RF: `copy_file()` and fix handling of unavailable files

### LINKED ISSUE #5600: copy-file is silent when copying files without any known copy
IMHO this is surprising behavior. I think this warrants at least a warning. But more likely an `impossible` result.

This is no about file content being absent in the source dataset, but about `whereis` not having a single source on record.

Demo:

```
% 
git -C serv annex whereis noavail.dat
whereis noavail.dat (0 copies) failed
git-annex: whereis: 1 failed
% datalad copy-file -d . serv/noavail.dat 
copy_file(ok): /tmp/datalad_temp_test_copy_file_nourlr7iq11d6/serv/noavail.dat [/tmp/datalad_temp_test_copy_file_nourlr7iq11d6/noavail.dat]
save(ok): . (dataset)                                                                      
action summary:
  copy_file (ok: 1)
  save (ok: 1)
```

mih:
This PR resurrects #5599 which was met with fuzzy disregard several months ago, and the hope for a better, more general solution. However no such solution surfaced, while the proposed RF and fix still make sense.

So here we are again.

I recommend looking at individual commits for a simpler story, rather than the full diff at once.

Fixes #5600

TODO

- [ ] test failure indicates potential changes in the behavior wrt the previous base of the PR

### DIFF
--- a/datalad/local/copy_file.py
+++ b/datalad/local/copy_file.py
@@ -634,6 +634,15 @@ def _copy_file(src, dest, cache):
     # are there any URLs defined? Get them by special remote
     # query by key to hopefully avoid additional file system interaction
     urls_by_sr = src_repo.get_key_urls_by_specialremote(finfo['key'])
+    if not urls_by_sr:
+        yield dict(
+            path=str_src,
+            destination=str_dest,
+            message='no known location of file content',
+            status='impossible',
+        )
+        return
+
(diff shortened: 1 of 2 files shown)


mih:
OK, closing for now, until there is time again.

Issue #5910, opened 2021-08-17
Discontinue double-reporting of error results


mih:
Demo:

```
% datalad create-sibling-gitlab --site gitlab.com 
[ERROR  ] No project path specified, and no configuration to derive one [create_sibling_gitlab(/tmp/some)] 
create_sibling_gitlab(error): . (dataset) [No project path specified, and no configuration to derive one]
```

The reasons for this have been described in https://github.com/datalad/datalad/issues/3752.

https://github.com/datalad/datalad/pull/3754 was introduced to provide means for preventing it.

I have used `datalad.log.result-level=debug` since that time (Oct 2019), I have not had a single case, where I missed that duplicate log message.

I think it is time to stop issuing log messages from results.

bpoldrack:
It may be worth making it conditional on whether or not the log target is a file, since a log file wouldn't have any indication of an error otherwise. But apart from that aspect, absolutely agree.

mih:
I should have been more precise: I want to change the log level from "match the result status" to `datalad.log.result-level` and make `datalad.log.result-level=debug` by default. If anyone configures the log target to be a file, they can also change that default to `datalad.log.result-level=None`.

PR #3754, opened 2019-10-06
Introduce datalad.log.result-level config switch

mih:
Setting this will cause messages included in command results (which are independently rendered as part of a result) to go to the identified log level (debug, info, warning, error). This can be used to alter the default behavior to have 'impossible' results cause a 'warning' log message and 'error' results yield an 'error' log message.

The default setup leads to cluttered output and double messaging. Setting this switch to 'debug' hides all result logging and emphasizes final results over intermediate events (where an 'error' may or may not be an indication of a problem in the big picture of a command execution).

- [x] fixes gh-3752

Exemplary impact:

was
```
% datalad install http://example.com/nothere
[INFO   ] Cloning http://example.com/nothere [1 other candidates] into '/tmp/nothere' 
[ERROR  ] Failed to clone from any candidate source URL. Encountered errors per each url were: (OrderedDict([('http://example.com/nothere', "Cmd('/usr/lib/git-annex.linux/git') failed due to: exit code(128)\n  cmdline: /usr/lib/git-annex.linux/git clone --progress -v http://example.com/nothere /tmp/nothere [cmd.py:wait:412]"), ('http://example.com/nothere/.git', "Cmd('/usr/lib/git-annex.linux/git') failed due to: exit code(128)\n  cmdline: /usr/lib/git-annex.linux/git clone --progress -v http://example.com/nothere/.git /tmp/nothere [cmd.py:wait:412]")]),) [install(/tmp/nothere)] 
install(error): /tmp/nothere (dataset) [Failed to clone from any candidate source URL. Encountered errors per each url were: (OrderedDict([('http://example.com/nothere', "Cmd('/usr/lib/git-annex.linux/git') failed due to: exit code(128)\n  cmdline: /usr/lib/git-annex.linux/git clone --progress -v http://example.com/nothere /tmp/nothere [cmd.py:wait:412]"), ('http://example.com/nothere/.git', "Cmd('/usr/lib/git-annex.linux/git') failed due to: exit code(128)\n  cmdline: /usr/lib/git-annex.linux/git clone --progress -v http://example.com/nothere/.git /tmp/nothere [cmd.py:wait:412]")]),)]
```

is

```
% datalad -c datalad.log.result-level=debug install http://example.com/nothere
[INFO   ] Cloning http://example.com/nothere [1 other candidates] into '/tmp/nothere' 
[...]

Issue #3752, opened 2019-10-05
Fate of logging of results


mih:
At present, any result that passes through the core event loop and has a logger instance attached, get a corresponding log message. This leads to generally duplicate reporting. For example:

```
% datalad save
[ERROR  ]   that3: setFileMode: permission denied (Operation not permitted) [add(/tmp/perm/that3)] 
add(error): that3 (file) [  that3: setFileMode: permission denied (Operation not permitted)]
```

I propose to demote all log messages triggered like this to a level less than `info`, hence remove them from the default user experience.

Rational: "error/impossible"-type events can be caused by non-faulty probing behavior and those should not clutter the terminal with scary inconsequential messages. Still logging any of such events at a lower level keeps them accessible for debugging. In addition, any faulty behavior that only leads to an "error" log message, without a corresponding "error" result is likely a bug (e.g. not leading to non-zero exit code, on-failure handling not controllable, etc.).

If that is not acceptable, an alternative solution to a cleaner UX is to no longer add a logger to the result records -- but that is clearly inferior for a developer perspective.

yarikoptic:
Duplication is indeed unpleasant but it is better than silent failing it just operating not as expected. It is not always possible to just rerun work a different log level.
I feel that there should either be explicit control (khe khe - option to the run, expect_error alike expect_stderr) or handling (alike swallow_logs)

mih:
What is "run" in this context?

yarikoptic:
Execution of any datalad command. They aren't atomic, so re- retuning might start from a different state.

mih:
Hmm, dont get it then. You want to have a cmdline arg that enables users to disable the double logging of results?

Just to be clear, any result gets a log message, regardless where it is coming from, and what is the originating command. Do you also propose that any disabling arg must be handed of to any called command inside any other command?

PR #5934, opened 2021-08-26
Log result record message at DEBUG level by default

mih:
Previously, messages with 'impossible' and 'error' status were logged
at WARNING and ERROR level respectively (by default). gh-3754 introduced
the config switch `datalad.log.result-level` to change this behavior,
but kept the default behavior.

This change makes DEBUG-level logging the default, and keeps the
previous default accessible as a 'match-status' mode.

Rational: Logging behavior can be configured in many ways, including
making it disappear from a user's point of view. Therefore it cannot be
relied on for critical messaging of _results_. Consequently, result
renderers perform their own messaging, ultimately leading to
double-reporting of errors and warnings in many default setups (see
gh-3752). Moreover, higher-level commands may later act on non-OK
results with recovery or mitigation actions. In such cases an
unconditional (prominent) logging of results emitted by a lower-level
command can be undesirable, but it is presently impossible to disable
from a calling command.

This change alters the requirements for result renderers to assume the
sole responsibility for visualizing (or not) any and all results that are
passed to them, without being able to rely on a particular logging
behavior.

It is unknown in how many edge cases critical messages will disappear
with this change. However, @mih has been running with
`datalad.log.result-level=debug` since Oct 2019, and has not noticed
undesirable behavior (or fixed it).

Fixes datalad/datalad#5910

### DIFF
--- a/datalad/interface/utils.py
+++ b/datalad/interface/utils.py
@@ -362,7 +362,7 @@ def eval_results(func):
         if not result_renderer:
             result_renderer = dlcfg.get('datalad.api.result-renderer', None)
         # look for potential override of logging behavior
-        result_log_level = dlcfg.get('datalad.log.result-level', None)
+        result_log_level = dlcfg.get('datalad.log.result-level', 'debug')
 
         # query cfg for defaults
         # .is_installed and .config can be costly, so ensure we do
@@ -595,7 +595,7 @@ def _process_results(
                 res_lgr = getattr(
                     res_lgr,
                     default_logchannels[res['status']]
-                    if result_log_level is None
+                    if result_log_level is 'match-status'
                     else result_log_level)
             msg = res['message']
             msgargs = None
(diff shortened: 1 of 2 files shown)


mih:
I am planing to merge this by the end of this week.

datalad/interface/utils.py
```diff
@@ -595,7 +595,7 @@ def _process_results(
                 res_lgr = getattr(
                     res_lgr,
                     default_logchannels[res['status']]
```
yarikoptic, line 597: FTR, its value is 
```
default_logchannels = {Michael Hanke, 4 years ago: • RF: Plain move. @build_doc is now base, not u…
    '': 'debug',
    'ok': 'debug',
    'notneeded': 'debug',
    'impossible': 'warning',
    'error': 'error',
}
```

yarikoptic:
no objections, although I neither  grasp neither have experience (did not use custom setting for logging AFAIK, trying to be like mortals) on either valuable logs would disappear.

mih:
Ok, then let's find out. If something disappears, it needs to be fixed.

Issue #5942, opened 2021-08-31
unicode encoding fiascos on a sample CentOS system 


yarikoptic:
Well, the system is somewhat unique, which lead to a number of recent issues/workarounds at git-annex level, see e.g. [xattr issue](https://git-annex.branchable.com/bugs/__34__357_out_of_984_tests_failed__34___on_NFS_lustre_mount/) but these particular ones seems to be all relating to some assumptions we have where we either need to instrument tests a bit differently or address them some other way

<details>
<summary>selection of tests failures all of some UnicodeEncodeError</summary> 

```shell
======================================================================
ERROR: datalad.support.tests.test_globbedpaths.test_globbedpaths
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/dartfs-hpc/rc/home/v/d31548v/.conda/envs/datalad/lib/python3.9/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/dartfs-hpc/rc/home/v/d31548v/.conda/envs/datalad/lib/python3.9/site-packages/datalad/tests/utils.py", line 583, in _wrap_with_tree
    create_tree(d, tree, archives_leading_dir=archives_leading_dir)
  File "/dartfs-hpc/rc/home/v/d31548v/.conda/envs/datalad/lib/python3.9/site-packages/datalad/utils.py", line 2420, in create_tree
    with open_func(full_name, "wb") as f:
UnicodeEncodeError: 'latin-1' codec can't encode character '\u03b2' in position 90: ordinal not in range(256)

======================================================================
ERROR: datalad.support.tests.test_network.test_get_local_file_url
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/dartfs-hpc/rc/home/v/d31548v/.conda/envs/datalad/lib/python3.9/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/dartfs-hpc/rc/home/v/d31548v/.conda/envs/datalad/lib/python3.9/site-packages/datalad/support/tests/test_network.py", line 477, in test_get_local_file_url
    print("D: %s" % path)
UnicodeEncodeError: 'latin-1' codec can't encode characters in position 12-17: ordinal not in range(256)

======================================================================
ERROR: datalad.tests.test_config.test_something
----------------------------------------------------------------------
Traceback (most recent call last):
[...]

Issue #2929, opened 2018-10-17
Any test uses `get_most_obscure_supported_name()` fails on old windows


mih:
It seems that the most obscure name is too obscure on windows vista, win7 seems fine. Not sure what to do about it.

No longer sure about the 'win7' is fine.

yarikoptic:
Seeing example of failure in eg test_gitrepo could have helped to provide informed feedback.

mih:
There is no need for feedback. Quoting breaks down. This is just a pointer for a useful starting point once a push for full windows-compat is made. Any test  that uses the default OBSCURE filename in any command execution fails.

yarikoptic:
cool, so it is insufficient or incorrect quoting/escaping? Seeing example of failure still could have helped my hungry mind even if nobody cares about my feedback.

yarikoptic:
What is vista? ;)
I do not think that "Any test that uses the default OBSCURE filename in any command execution fails." any longer but quoting indeed seems still to cause some issues, e.g. https://github.com/datalad/datalad/pull/5117#issuecomment-721456122 . But altogether I feel that this issue could be closed, and more specific ones should be created/tackled.

PR #5944, opened 2021-09-01
BF: do use OBSCURE_FILENAME instead of hardcoded unicode

yarikoptic:
OBSCURE_UNICODE should be used since otherwise tests would
just keep crashing while running on file systems/environment
which do not support those fancy characters.

Initial intention here was to address some Windows fails on Vista and
other elderly systems.  https://github.com/datalad/datalad/issues/2929
If tests pass on current CI I would assume that we are past that
stone age.

Closes a part of #5942.  ~~Wanted first to see if there are side-effects~~

### DIFF
--- a/datalad/support/tests/test_globbedpaths.py
+++ b/datalad/support/tests/test_globbedpaths.py
@@ -11,17 +11,21 @@
 
 __docformat__ = 'restructuredtext'
 
-from itertools import product
 import logging
-from unittest.mock import patch
 import os.path as op
+from itertools import product
+from unittest.mock import patch
 
-from datalad.support.globbedpaths import GlobbedPaths
-from datalad.tests.utils import assert_in
-from datalad.tests.utils import eq_
-from datalad.tests.utils import swallow_logs
-from datalad.tests.utils import with_tree
-from datalad.tests.utils import known_failure_windows
+from datalad.tests.utils import (
+    OBSCURE_FILENAME,
+    assert_in,
+    eq_,
+    known_failure_windows,
+    swallow_logs,
+    with_tree,
+)
+
+from ..globbedpaths import GlobbedPaths
 
 
 def test_globbedpaths_get_sub_patterns():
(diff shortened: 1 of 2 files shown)


yarikoptic:
besides codecov all green, good, so I assume that appveyor is our gold standard for minimal version of windows to be compatible and will proceed fixing other similar failures in the same fashion.

yarikoptic:
Windows doesn't like my questionable fix

```
TypeError: bytes args is not allowed on Windows
```

datalad/cmd.py
```diff
@@ -91,10 +90,19 @@ async def run_async_cmd(loop, cmd, protocol, stdin, protocol_kwargs=None,
...
+        # do that: TypeError: bytes args is not allowed on Windows
+        encode_ = lambda s: s
+    else:
+        encode_ = lambda s: s.encode('utf-8')
     if isinstance(cmd, str):
-        proc = loop.subprocess_shell(factory, cmd, **kwargs)
+        proc = loop.subprocess_shell(factory, encode_(cmd), **kwargs)
     else:
-        proc = loop.subprocess_exec(factory, *cmd, **kwargs)
+        proc = loop.subprocess_exec(factory, *map(encode_, cmd), **kwargs)
```
yarikoptic, line 105: this change (from the 2nd commit) is the most questionable in this PR.  According to tests it seems to do the right thing, and the fact that it otherwise fails on "encode" step suggests that a similar action is done by Python std library itself.  But here we would "hardcode" us into "utf-8" as the "one and only" we support.
An alternative would be to address failing tests by making them skip if system locale is not utf-8 or not use unicode (since might fail) in the test values -- but the point of some of those tests is exactly that -- test that unicode could be provided, and according to results of the test run on those systems with non utf-8 locale -- it is possible (with this explicit `encode` here).

WDYT?

yarikoptic: no longer pertinent here - decided it is worth a separate PR/discussion: #5961

yarikoptic:
Extracted a more "core" change to #5961 so this PR now concerns only with tests.

adswa:
I can't spot anything wrong with this PR, looks sensible to my not-so-expert eyes :)

yarikoptic:
Thank you @adswa for feedback on all PRs staged for the release. Let's cut out the release by merging this PR, adding the label

Issue #5946, opened 2021-09-01
`remove-sibling-*`


mih:
At present we do a bad job at maintaining the integrity of a network of git-annex repo clones. Pretty much nothing announces the death of a clone. In a paradigm where we have a "central" clone on which `create-sibling-*` commands operate, but where we require users to individually remove such siblings using unspecified platform-specific means (delete repo on (github|osf|..), log into a machine and run `remove` in a shell, etc.) this is hard to implement.

What about changing the paradigm to also perform sibling removal from a "central" management clone?

* and by central I obviously do not mean any significant centralization, but rather a more persistent clone that knows about all relevant siblings.

In such a paradigm, we would also have a series of `remove-sibling-*` commands to parallel their `create-sibling-*` companions. Or we become fancy and learn the ability to have subcommands for `create` and `remove` actions of a `sibling-<type>` proxy command series.

mih:
Closing this old issue of mine. The idea remains recorded, but there are clearly no plans to make this a reality any time soon.

Issue #5939, opened 2021-08-30
`create-sibling-github` cannot configure https-pull/ssh-push setup


mih:
All information is present internally, but the API can only handle uniform HTTPS vs SSH choices.

This will become even more relevant with

```
# TODO let submodule URLs point to GitHub (optional)
```

Issue #5932, opened 2021-08-26
FYI: How to obtain a Gin token for authentification via curl


adswa:
@mih and I have toyed around with https://github.com/G-Node/gin-cli to find out how the gin-client is able to obtain tokens for authentication that are required in GET requests for authentication.
Here is a ``curl`` call that does it:

```
curl -v \
   -H "Content-Type: application/json" \
   -H "Authorization: Basic $(echo -n adswa:<secretpassword> | base64)" \
  https://gin.g-node.org/api/v1/users/adswa/tokens
```

The result is a token that looks like this:
```
[{"name":"gin-cli","sha1":"super-secret-40ish-char-sha"}]
```

Subsequent HTTP headers to anything but the ``token`` url use 
``Authorization: token super-secret-40ish-char-sha`` and  ``Content-Type:application/jsonAuthorization``

adswa:
This can be closed, we decided to leave it to the user to retrieve and supply one.

Issue #4079, opened 2020-01-24
Reenable github integration test


mih:
A failing test of special-interest functionality was a blocker to an urgent global fix #4078

I disabled the test with cad795b0f73a952e38214c62fe1a35eef43d258a:

Please bring it back @yarikoptic

yarikoptic:
the fact that it started to fail suggests that the interaction sequence with github has changed, and I wonder why it has changed (unless the #4078 fixed something which resulted e.g. in a new url or credentials being used)

adswa:
It looks like this has been resolved by #5218?

yarikoptic:
I think so, at least at that time. thank you @adswa ! left a relevant note on https://github.com/datalad/datalad/pull/5949#issuecomment-918164954 which provides a fresh generic reimplementation of github support.

PR #5949, opened 2021-09-02
create-sibling-{github,gogs,gitea,gin} as instances of GitHub-like platforms

### LINKED ISSUE #2680: Interoperability with GIN-hosting
Datasets hosted with/at GIN https://web.gin.g-node.org should be working nicely with DataLad. However, there are a few things that needs to polish:

- the GIN webUI only offers "GIN" and "SSH" access (not sure if anonymous HTTP is intentionally unavailable). This means that DataLad users need to register, generate, and upload an SSH key to be able to `datalad install` from GIN. That seems needlessly complicated, but we might not be able to do anything about it at our end.

- the "public" datasets use datacite as metadata standard (dataset-level), but in the YAML file format flavor. At present DataLad only understands the XML flavor.

- some datasets contain the actual data files as ZIP archives (e.g. https://web.gin.g-node.org/bendalab/Henninger2017_electrosensory), maybe we can have a helper that sifts through such a dataset, and unpacks the archives (in the spirit of add-archive).

New dev contact is @achilleas-k

Tasks (from must-have to nice-to-have)

- [x] enable read access to annex'ed files (see below for failure report)
- [x] enable annex `copy --to` operations
- [x] add CI test to make sure that datalad and GIN remain interoperable (we have approval to host small test datasets and access them for CI purposes)
- [x] enable anonymous read access (see https://github.com/datalad/datalad/issues/2680#issuecomment-412152760)
- [ ] enable programmatic creation of repos (analog to what we have for SSH server, GitHub and GitLab, ie. `create-sibling...`)
- [ ] check what of the GIN client functionality we should wrap (or reimplement) to make it fit into datalad's API
- [ ] check to which degree DataLad can work (and be advertised) as an alternative GIN client: https://gin.g-node.org/G-Node/Info/wiki/GinClientHelp

### LINKED ISSUE #3755: create-sibling-github: Allow to specify {organization,login}/reponame
A report came from our tireless pilot users, who tried `create-sibling-github hislogin/repo` to get `hislogin-repo` generated on github instead of expected `repo` under his `github.com/login`.  I think it would be nice from UX perspective to allow for `reponame` specification to carry `organization/` or `login/` prefix.  The tricky part on our side would be necessity to figure out, through interaction with github, either it is a login or organization since subsequent codepath is a bit different between the two, so it is not just a matter of splitting into `github_organization, reponame` as I have blindly tried first with 

<details>
<summary>this patch</summary> 

```patch
diff --git a/datalad/distribution/create_sibling_github.py b/datalad/distribution/create_sibling_github.py
index fc8d74125..6b37da58c 100644
--- a/datalad/distribution/create_sibling_github.py
+++ b/datalad/distribution/create_sibling_github.py
@@ -86,7 +86,9 @@ class CreateSiblingGithub(Interface):
             args=('reponame',),
             metavar='REPONAME',
             doc="""Github repository name. When operating recursively,
-            a suffix will be appended to this name for each subdataset""",
+            a suffix will be appended to this name for each subdataset. Could
+            be used to specify both organization and repository name if
+            specified as 'org/name'""",
             constraints=EnsureStr()),
         recursive=recursion_flag,
@@ -154,6 +156,16 @@ class CreateSiblingGithub(Interface):
         # unnecessary dependencies
         from datalad.support.github_ import _make_github_repos
 
+        # If user provides "login/repo" (as they do), split based on '/'
+        if '/' in reponame and not github_organization:
+            github_organization, reponame_ = reponame.split('/', 1)
+            lgr.info(
+                "Split original '%s' into '%s' organization and '%s' repo name",
+                reponame, github_organization, reponame_
+            )
+            reponame = reponame_
+            del reponame_
+
         # what to operate on
         ds = require_dataset(
             dataset, check_installed=True, purpose='create Github sibling')
```
</details>

### LINKED ISSUE #5324: create-sibling-github fails for enterprise base URL
#### What is the problem?
`datalad create-sibling-github` can only send requests to `github.com` not enterprise URLs like `git.mycompany.edu`. 

Enabling an optional parameter like `--enterprise-url` with the `datalad create-sibling-github` command for custom URL input would fix this issue. I believe this approach would require modifying `datalad/datalad/distribution/create_sibling_github.py` and `datalad/datalad/support/github.py` to accept and utilize an additional parameter.

Datalad version 0.13.7 on Mac OS Catalina 10.15.7.

Nothing else to note.

I love Datalad! To make sure that this issue was not user error on my part, I tried `datalad create-sibling-github` on my personal account, which worked perfectly, so I am pretty certain the issue is with the requesting URL to github.

### LINKED ISSUE #5935: Implement `create-sibling-gogs`
https://gogs.io is a fairly popular solution for Git "self-hosting", i.e. my own github. It seems as if the API has all that is needed to implement a `create-sibling` command. As far as I can tell, and #5932 seems to confirm that, the GIN services does not deviate much (at all?) from this API, so such a command would also effectively be a `create-sibling-gin`, and thereby complete #2680.

git-annex functionality (added to GIN, but not gogs) should be enabled transparently, based on git-annex's normal operation.

I have created an account on https://try.gogs.io to test an implementation and maybe later on run the CI against..

### LINKED ISSUE #5937: Decomplexify/unify `create-sibling-github`
Looking into #5935 I realized that the pieces of the `gogs` API we would need in order to implement `create-sibling-gogs` are fairly minimal. This made me wonder how that would translate to our use of the github API in `create-sibling-github`, given that one is modeled after the other.

This brought two realizations:

1. The `create-sibling-github` implementation is fairly sophisticated  (in the bad way). After the removal of user/pass auth, it is hard to see why we would need to be using this credential-test-cycle approach anymore -- or even the support of these `hub` tokens. It seems just taking a personal access token is all it needs.
2. If (1) is implemented, we will be performing (very) few different API calls. All of them trivial (no sessions management or alike). Just adding the token to the header of any request is all that is needed in order to make these requests with something generic as `requests`. If we would consider getting rid of pygithub, we could implement `create-sibling-gogs` with almost the exact same code as a reworked `create-sibling-github` -- we would just need to make the API endpoint mapping explicit and use targeted error handlers.

(1) and (2) combined would pretty much automatically also handle #3755 and #5324, as both a features of the less-centralized nature of gogs too.

### LINKED ISSUE #5941: `create-sibling-github --existing skip` configures a sibling
Not sure how I feel about this behavior. I have a dataset that I want to create a github sibling for. I (accidentally) chose a repo name that conflicts with an already existing repo on github. So this behavior is good:

```
% datalad create-sibling-github funkydummy3
create_sibling_github(error): . (dataset) [repository "funkydummy3" already exists on Github]
```

but this one, I cannot easily get behind:

```
% datalad create-sibling-github funkydummy3 --existing skip
action summary:
  configure-sibling (ok: 1)
  create_sibling_github (notneeded: 1)
```

I does skip the creation of the repo (to avoid the conflict), but it also proceeds to configure the (totally unrelated repo, in my case) as a sibling.

I would have expected the `existing='skip'` to mean "if there already is a sibling configured, skip", not "if there already is a repo, skip creation, but configure a sibling".

I think I would expect such a behavior with `existing='reconfigure'`, but not with `skip`.

mih:
This is a large PR that adds to new supported platforms (GIN, fixes #2680, and GOGS, fixes #5935), and reimplements the GitHub-platform support using the same code (fixes #5937).

Moreover, it comes with the following changes:

- [x] personal access tokens are now the only supported auth method. They can be provided via an option, or all supported mean to provide credentials.
- [x] there is no support for automatically obtaining access tokens using other means of authentication, in order to keep the complexity of the implementation low (ping #5932)
- [x] most code can be tested without network using a uniform `--dry-run` mode
- [x] the same test code evaluates all three platform directly, using a particular instance each with a dedicated account. Respective token are deposited on appveyor.
- [x] the previous `create-sibling-github` implementation has been replaced almost in full. In particular, the complex github-only credential handling was trimmed down to match the others. There is no dependency on `PyGithub` anymore.
- [x]  enables direct specification of organization repos via a `[<org>/]repo` syntax. Fixes #3755 
- [x]  makes API URLs configurable for all platforms, incl. GitHub. Fixes #5324
- [x] no longer reconfigure an existing repo with `existing=skip` when there is no local conflicting sibling. Do so only with `existing=reconfigure`. Fixes #5941
- [x]  implement `existing=replace` in a consistent manner, or stop supporting it (see related #5946). From discussion: only implement for github.
- [x] robustify handling of deprecated parameters of `-github`, but @yarikoptic appears to argue for a hard rip-out in https://github.com/datalad/datalad/pull/5949#discussion_r701121819 Conlusion from call: implement a simple mapping, as planned
- [x] investigate what causes the SSH connection attempt in the gin tests that stall the CI on mac and linux, e.g. https://ci.appveyor.com/project/mih/datalad/builds/40620305/job/36yo7wifel4164ld
- [x] change `--auth` to something like `--credential <name>`, and document cfg and env spec possibilities in docstring
- [x]  gitea was added with minimal effort
- [x] run `isort`
- [x] make pass over documentation
- [x] decide on a (common) default for `--access-protocol` (see #5939). With GIN a sensible setup is a push URL with SSH (to enable annex-copy) and a pull URL with http (to enable --as-common-datasrc). For all other platforms pure-http or http/ssh can make equal sense (on windows pure-http would be a better choice due to limited SSH support, think credentials, think no multiplexing)


datalad/distributed/create_sibling_ghlike.py
```diff
@@ -0,0 +1,549 @@
...
+    toprocess = [ds]
+    if recursive:
+        for sub in ds.subdatasets(
+                # we want to report on missing dataset in here
+                fulfilled=None,
+                recursive=recursive,
+                recursion_limit=recursion_limit,
+                result_xfm='datasets'):
+            if not sub.is_installed():
+                lgr.info('Ignoring unavailable subdataset %s', sub)
```
yarikoptic, line 546: I thought that using logger (INFO level in particular) for such purposes is not in fashion any longer... should it return annotated dict of what to process and what to "skip" for some reason, so code above could yield corresponding result records?

datalad/distributed/create_sibling_ghlike.py
```diff
@@ -0,0 +1,549 @@
...
+
+class _GitHubLike(object):
+    name = None
+    fullname = None
+    create_org_repo_endpoint = None
+    create_user_repo_endpoint = None
+    get_authenticated_user_endpoint = None
+    get_repo_info_endpoint = None
+    response_code_repo_created = requests.codes.created
+    response_code_unauthorized = requests.codes.forbidden
```
yarikoptic, line 110:
```suggestion
    # "parametrization" derived classes should provide
    name: str = None
    fullname: str = None
    create_org_repo_endpoint: str = None
    create_user_repo_endpoint: str = None
    get_authenticated_user_endpoint: str = None
    get_repo_info_endpoint: str = None
    response_code_repo_created: int = requests.codes.created
    response_code_unauthorized: int = requests.codes.forbidden
```
type annotation (comes for free since `str` and `int` is built-in) helps to immediately get an idea on what are those to be values to be.  First reading this I was not sure if may be you tried to define `@absractmethods` this ad-hoc way.
Strictly speaking, `""` not `None` should be used or `Optional[str]` with import of Optional from typing should be used, but since we are not using mypy yet to do types checking, I think it would be fine and provide desired DX effect without that.

datalad/distribution/create_sibling_github.py
```diff
@@ -180,98 +183,22 @@ def __call__(
...
-        ds = require_dataset(
-            dataset, check_installed=True, purpose='create GitHub sibling(s)')
-
-        res_kwargs = dict(
-            action='create_sibling_github [dry-run]' if dry_run else
-            'create_sibling_github',
-            logger=lgr,
-            refds=ds.path,
+        # TODO robustify the deprecation mappings
+        auth = auth or github_login
```
yarikoptic, line 187: why `login` should be here? to tell apart one of the credentials from another? (IIRC we had something like that in current implementation for github to support choosing between possibly multiple tokens available)

datalad/distribution/create_sibling_github.py
```diff
@@ -80,41 +133,9 @@ class CreateSiblingGithub(Interface):
...
-            or process interrupts with error ('error'). DANGER ZONE: If 'replace'
-            is used, an existing github repository will be irreversibly removed,
-            re-initialized, and the sibling (re-)configured (thus implies 'reconfigure').
-            `replace` could lead to data loss, so use with care.  To minimize
-            possibility of data loss, in interactive mode DataLad will ask for
-            confirmation, but it would raise an exception in non-interactive mode.
-            """,),
+    _params_ = _GitHub.create_sibling_params
+    # deprecated option
+    _params_.update(
```
yarikoptic, line 138: all of them (since above `_params_`) or just `--github-login` (move one line down)? I think `--github-organization` and others remains valid. If so:
```suggestion
    _params_.update(
        # deprecated option
```
but even then -- not sure if to be deprecated, per concern stated in another comment -- how to choose between multiple credentials, where knowing the login would help to disambiguate

yarikoptic (commented):
Initial quick pass with minor comments. Didn't try yet and didn't comprehend the full scale of changes due to amount of code changes

mih:
I have no idea yet how to tackle this:

```
/home/appveyor/dlvenv/lib/python3.8/site-packages/urllib3/connectionpool.py:1013: InsecureRequestWarning: Unverified HTTPS request is being made to host 'gin.g-node.org'. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/1.26.x/advanced-usage.html#ssl-warnings
984  warnings.warn(
985/home/appveyor/dlvenv/lib/python3.8/site-packages/urllib3/connectionpool.py:1013: InsecureRequestWarning: Unverified HTTPS request is being made to host 'gin.g-node.org'. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/1.26.x/advanced-usage.html#ssl-warnings
986  warnings.warn(
987The authenticity of host 'gin.g-node.org (141.84.41.219)' can't be established.
988ECDSA key fingerprint is SHA256:E35RRG3bhoAm/WD+0dqKpFnxJ9+yi0uUiFLi+H/lkdU.
989Are you sure you want to continue connecting (yes/no/[fingerprint])? 
```

The warning comes from `verify=False` which I tried as a test. The real problem is the interactive prompt -- which I do not see locally, nor on Windows, only appveyor Linux and Mac. Maybe a missing dependency?

datalad/distributed/create_sibling_ghlike.py
```diff
@@ -0,0 +1,549 @@
...
+            irreversibly removed, re-initialized, and the sibling
+            (re-)configured (thus implies 'reconfigure').
+            `replace` could lead to data loss! In interactive a confirmation
+            prompt is shown, an exception is raised in non-interactive
+            sessions.""",),
+        auth=Parameter(
+            args=('--auth',),
+            constraints=EnsureStr() | EnsureNone(),
+            metavar='TOKEN',
+            doc="""personal access token"""),
```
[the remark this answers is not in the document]
mih: Env vars and config items are already supported. If you want to disallow the token specification via cmdline option, we can take a token "label" with the same option, and use it to look up a particular credential. Right now I am using the `api_url` `netloc` to do this, but an explicit label would be even better.

datalad/distributed/create_sibling_ghlike.py
```diff
@@ -0,0 +1,549 @@
...
+                'skip', 'error', 'reconfigure', 'replace'),
+            doc="""behavior when already existing or configured
+            siblings are discovered: skip the dataset ('skip'), update the
+            configuration ('reconfigure'), or fail ('error').
+            DANGER ZONE: With 'replace', an existing repository will be
+            irreversibly removed, re-initialized, and the sibling
+            (re-)configured (thus implies 'reconfigure').
+            `replace` could lead to data loss! In interactive a confirmation
+            prompt is shown, an exception is raised in non-interactive
+            sessions.""",),
```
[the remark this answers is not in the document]
mih: C&P from github code.thx.

datalad/distributed/create_sibling_ghlike.py
mih: Thx

datalad/distributed/create_sibling_ghlike.py
```diff
@@ -0,0 +1,549 @@
...
+        self._user_info = None
+
+        # maybe a token can be retrieved?
+        if not self.auth:
+            token_name = urlparse(url).netloc
+            try:
+                # TODO platform-specific doc URL for token generation
+                self.auth = Token(token_name, url=url)()['token']
+            except Exception as e:
+                lgr.debug('Token retrieval failed: %s', e)
```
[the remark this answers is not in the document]
mih: Noted!

datalad/distributed/create_sibling_ghlike.py
mih: Good point. This is a C&P from github code.

datalad/distribution/create_sibling_github.py
mih: This is purely for compatibility with the present API. We can rip it out.

datalad/distribution/create_sibling_github.py
mih: It refers to all of them.

datalad/distributed/create_sibling_ghlike.py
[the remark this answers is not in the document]
mih: Done

mih:
SSH issues is fixed. Only failure that remains is https://github.com/datalad/datalad/issues/5967 (which is already present in master).

mih:
This is now technically complete. Only beautification and documentation is TODO.

yarikoptic:
VCR tapes based integration tests were always a bit of a pain (since we didn't fully automate their sanitization), so not sure if worth the pains here.  For anything on current list, but github, it could be real integration test against some docker container the actual service.  But I guess could be considered out of scope of this PR.  FWIW, relevant past issue  ref https://github.com/datalad/datalad/issues/4079 which I will close now.

datalad/distributed/create_sibling_github.py
```diff
@@ -0,0 +1,266 @@
...
+
+    In order to be able to use this command, a personal access token has to be
+    generated on the platform (Account->Settings->Developer Settings->Personal
+    access tokens->Generate new token).
+    """
+
+    _examples_ = [
+        dict(text="Use a new sibling on GIN as a common data source that is "
+                  "auto-available when cloning from GitHub",
+             code_py="""\
```
adswa, line 136: I think, as a first basic example, a single command to just create a GitHub sibling would be helpful. This would also help to give the examples section a structure that makes it clear that the long, multi-command example is indeed a single example, not four. I just tried the command, and I was quite confused because with the lack of another example, it looks like there are four example commands, two of which use "create-sibling-gin".

adswa:
UX: When I don't have a token set up, the information that each command provides isn't really helpful. 
For Github I get pointed to [https://api.github.com](https://api.github.com), which is wrong and confusing. It should be [https://github.com/settings/tokens](https://github.com/settings/tokens), which can be hard to find in the web interface as it is reachable only via "Developer Settings".
```
❱ datalad create-sibling-github new-create-sibling-command-gh
You need to authenticate with None credentials. https://api.github.com provides information on how to gain access
token: 
```

For Gin I get pointed to [https://gin.g-node.org](https://gin.g-node.org), which is better, but it still takes me some time to find the correct page ([https://gin.g-node.org/user/settings/applications](https://gin.g-node.org/user/settings/applications))
```
❱ datalad create-sibling-gin new-create-sibling-command-gin -s gin
You need to authenticate with 'gin' credentials. https://gin.g-node.org provides information on how to gain access
token: 
```

datalad/distributed/create_sibling_github.py
```diff
@@ -0,0 +1,266 @@
...
+    @eval_results
+    def __call__(
+            reponame,
+            dataset=None,
+            recursive=False,
+            recursion_limit=None,
+            name='github',
+            existing='error',
+            github_login=None,
+            credential=None,
```
adswa, line 211: Why is the credential name "None" here? In case of 'gin' and 'gogs' it has sensible default names. I just copy-pasted a wrong token, and it took me a while to find the credential under the name "datalad-None" to delete it.

datalad/distributed/create_sibling_github.py
mih: Yes, I pretty much made a protocol of some testing I did. All these commands should have more.

datalad/distributed/create_sibling_github.py
mih: I would not mind making it `'github'`. However, the previous implementation had so many ways to pull credentials from, I had no good concept. Probably safe.

mih:
> For Github I get pointed to https://api.github.com, which is wrong and confusing. It should be https://github.com/settings/tokens, which can be hard to find in the web interface as it is reachable only via "Developer Settings".

Yes, that is suboptimal. However, the URL you gave is only correct for github.com, not for any other github deployment. Moreover, it is not always possible to construct the desired info URL from the API url.

I think the main issue here is that the credential system requires a URL here, instead of just textual instructions.

mih:
I will shortly push an update that addresses the instruction problem:

```sh
% datalad create-sibling-gitea --credential no test 
An access token is required for https://gitea.com. Visit https://gitea.com/user/settings/applications to create a token

% datalad create-sibling-gogs --api https://mygogs.me.com --credential no test 
An access token is required for https://mygogs.me.com. Visit https://mygogs.me.com/user/settings/applications to create a token

% datalad create-sibling-gin --credential no test 
An access token is required for https://gin.g-node.org. Visit https://gin.g-node.org/user/settings/applications to create a token

% datalad create-sibling-github --credential no test 
An access token is required for https://api.github.com. Visit https://github.com/settings/tokens to create a token.

% datalad create-sibling-github --api https://api.mygithub.de --credential no test 
An access token is required for https://api.mygithub.de. Log into the platform, and visit [Account->Settings->Developer Settings->Personal access tokens->Generate new token]to create a new token.

```

datalad/distributed/create_sibling_github.py
mih: I decided to implement a uniform policy of using the hostname part of the API url as the default credential name. So `None` cannot occur anymore.

mih:
This is complete now, and I have to detach. Please feel most welcome to add more examples, and doc tweaks.

bpoldrack:
Before a code review, collecting observations from trying:

- `--dry-run` doesn't seem to apply to implicit actions. That's not necessarily wrong, but also not necessarily obvious. Concrete case: dry-run with `create-sibling-gin` asked for token, subsequent dry-run doesn't anymore. It was stored the first time. Again that may even be desirable, but for a moment I was surprised, since my mental concept of a `dry-run` suggests no changes on disc.

- GIN doesn't appear to work with a repo under an organization for me, but that's described in internal chat and looks more like an issue with GIN rather than datalad.

datalad/distributed/create_sibling_ghlike.py
```diff
@@ -0,0 +1,638 @@
...
+                message='repository already exists',
+                preexisted=True,
+            )
+        elif r.status_code == self.response_code_unauthorized:
+            return dict(
+                status='error',
+                message=('unauthorized: %s', response.get('message')),
+            )
+        # make sure any error-like situation causes noise
+        r.raise_for_status()
```
bpoldrack, line 505: As pointed out in https://github.com/datalad/datalad/pull/5949#issuecomment-924713051, with that `raise_for_status` we kind of throw away the message from the platform (`str(r)` would have it), turning it into a generic HTTPError. Not sure how to it best, but we could simply enhance the message. Something along those lines:

```
try:
    r.raise_for_status()
except HTTPError as e:
    raise HTTPError(str(e) + '%s responded with %s %s'.format(self.fullname, r, response)) from e
```

`requests.Response` would also allow to overwrite the default `.reason` that is used to build the `HTTPError`, but that seems to much reliance on the reponses  body to be more meaningful.

Alternatively: Catch and return an error result? Not sure about implications down the line yet.

Last thing I can see ATM: `HTTPError` does have the `Response` linked already it seems. Then it could also be up to rendering the exception to display the actual response. Either way, just letting the generic HTTPError bubble up, swallows meaningful information.

datalad/distributed/create_sibling_gin.py
```diff
@@ -0,0 +1,133 @@
...
+        """
+        return dict(
+            reponame=response.get('name'),
+            private=response.get('private'),
+            # GIN reports the SSH URL as 'clone_url', but we need
+            # a HTML URL (without .git suffix) for setting up a
+            # type-git special remote (if desired)
+            clone_url=response.get('html_url'),
+            ssh_url=response.get('ssh_url'),
+            html_url=response.get('html_url'),
```
bpoldrack, line 44: Saw that in a couple of spots, I think, and I'm a bit confused. What's an HTML-URL? Shouldn't that be an HTTP-URL? If it's a about a browsable site to report on, that's okay, but the comment reads we need such to setup git type remote, so possible HTML content at the URL doesn't seem to be what we are after here.

datalad/distributed/create_sibling_gin.py
```diff
@@ -0,0 +1,133 @@
...
+                 % datalad siblings configure -s gin --as-common-datasrc gin-storage
+                 # announce availability (redo for other siblings)
+                 % datalad push --to gin
+                 """,
+             ),
+    ]
+
+    _params_ = _GIN.create_sibling_params
+    _params_['api']._doc = """\
+        URL of the GIN instance without a 'api/<version>' suffix"""
```
bpoldrack, line 102:
```suggestion
        URL of the GIN instance without an 'api/<version>' suffix"""
```

datalad/distributed/create_sibling_gitea.py
```diff
@@ -0,0 +1,135 @@
...
+                message='repository already exists',
+                preexisted=True,
+            )
+        elif r.status_code == self.response_code_unauthorized:
+            return dict(
+                status='error',
+                message=('unauthorized: %s', response.get('message')),
+            )
+        # make sure any error-like situation causes noise
+        r.raise_for_status()
```
bpoldrack, line 76: See above (same spot in base class)

bpoldrack (commented):
Generally, I like it. Left a bunch of comments, though. From my point of view can be addressed (if applicable) after a merge as well, but at least the error reporting part should become an issue in that case.

datalad/distributed/create_sibling_ghlike.py
```diff
@@ -0,0 +1,638 @@
...
+            args=("--dry-run",),
+            action="store_true",
+            doc="""if set, no repository will be created, only tests for
+            name collisions will be performed, and would-be repository names
+            are reported for all relevant datasets"""),
+    )
+
+    def __init__(self, url, credential, require_token=True, token_info=None):
+        if not url:
+            raise ValueError(f'API URL required for {self.fullname}')
```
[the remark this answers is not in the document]
mih: I'd rather crash TBH

datalad/distributed/create_sibling_ghlike.py
```diff
@@ -0,0 +1,638 @@
...
+                'Cannot determine authorization token for %s', credential)
+            if require_token:
+                raise ValueError(
+                    f'Authorization required for {self.fullname}, '
+                    f'cannot find token for a credential {credential}.')
+            else:
+                lgr.warning("No token found for credential '%s'", credential)
+                self.auth = 'NO-TOKEN-AVAILABLE'
+
+        self.request_headers = {
```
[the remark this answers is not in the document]
mih: I have no idea what other implications this has. I would leave such changes to someone else, with more experience. I found it helpful to see that there is no hidden magic required.

datalad/distributed/create_sibling_ghlike.py
```diff
@@ -0,0 +1,638 @@
...
+                    and existing not in ('reconfigure', 'replace'):
+                # nothing to do anymore, when no reconfiguration is desired
+                continue
+
+            # lastly configure the local datasets
+            for var_name, var_value in \
+                    self.extra_remote_settings.items():
+                var = 'remote.{}.{}'.format(siblingname, var_name)
+                if var not in d.config:
+                    d.config.add(var, var_value, where='local')
```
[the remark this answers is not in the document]
mih: Good catch!

datalad/distributed/create_sibling_ghlike.py
```diff
@@ -0,0 +1,638 @@
...
+            doc="""dataset to create the publication target for. If not given,
+            an attempt is made to identify the dataset based on the current
+            working directory""",
+            constraints=EnsureDataset() | EnsureNone()),
+        reponame=Parameter(
+            args=('reponame',),
+            metavar='[<org-name>/]<repo-(base)name>',
+            doc="""repository name, optionally including an '<organization>/'
+            prefix if the repository shall not reside under a user's namespace.
+            When operating recursively, a suffix will be appended to this name
```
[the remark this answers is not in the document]
mih: Will adjust

mih:
Thx for the review @bpoldrack ! I will fix some of the things you reported. I consider the implementation of a dry-run mode for credentials management out-of-scope for this PR. If necessary, I can disable the dry-run, and with it disable all tests that so far require no network. I consider that a waste, though.

datalad/distributed/create_sibling_ghlike.py
mih: I decided to specifically support the error condition that you faced, rather than adding more sophisticated (re)raising of exceptions, that would nevertheless lead to debugging. Using a token without permissions to create a repo under an existing org now looks like:

```
create_sibling_gin(error): [unauthorized: given user is not owner of organization]
```

datalad/distributed/create_sibling_gitea.py
mih: And solved same way as described above.

datalad/distributed/create_sibling_ghlike.py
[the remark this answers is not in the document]
mih: This will now raise `NotImplementedError`

mih:
OK, let's do this. All CI failures are unrelated issues present in master already. Thanks for all the feedback!

bpoldrack:
> I consider the implementation of a dry-run mode for credentials management out-of-scope for this PR.

Yes, perfectly fine with that. I'm also not sure we'd want a change in that regard. Just noticed that it can be seen as a little weird.

PR #5952, opened 2021-09-02
Pass `--no-changelog` to `auto shipit` if changelog already has entry

### LINKED ISSUE #5782: Enhance auto CHANGELOG generation
During last weekly meeting we discussed the desire to see auto entries in changelog to be improved. Two main questions aspects came to mind

- Ability to provide custom summary for the release.  Something like what we did (sometimes) before:  e.g. https://github.com/datalad/datalad/blob/master/CHANGELOG.md#0130-june-23-2020---- . So would be nice to somehow be able to modify some perspective entry within CHANGELOG which would get plugged in into next release on top.  If not possible with auto "natively", we might need to hackup some auto extension.
- Ability to provide custom changelog entry for each PR in the PR.  ATM it is a PR title (if I get right), but that one is a bit restrictive since should be short. For some changes it is desired to provide a more elaborate description beyond title.  Is that possible?

jwodder:
Closes #5782.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -34,7 +34,24 @@ jobs:
         run: sudo apt-get update && sudo apt-get install -y pandoc
 
       - name: Create release
-        run: ~/auto shipit
+        run: |
+          new_version="$(~/auto shipit --dry-run --quiet)"
+          if [ -z "$new_version" ]
+          then echo "Nothing to release"
+               exit
+          fi
+          changelog_version="$(perl -nle '/^# ([0-9]+\.\S*) \(.*\)/ and print $1 and exit' CHANGELOG.md)"
+          opts=
+          if dpkg --compare-versions "$new_version" lt "$changelog_version"
+          then echo "[ERROR] New version is less than latest version in changelog"
+               exit 1
+          elif dpkg --compare-versions "$new_version" eq "$changelog_version"
+          then opts=--no-changelog
+          elif ! dpkg --compare-versions "$new_version" gt "$changelog_version"
+          then echo "[ERROR] Could not compare new version '$new_version' to changelog version '$changelog_version'"
+               exit 1
+          fi
+          ~/auto shipit $opts
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           TWINE_USERNAME: __token__


.github/workflows/release.yml
```diff
@@ -34,7 +34,24 @@ jobs:
...
+          if dpkg --compare-versions "$new_version" lt "$changelog_version"
+          then echo "[ERROR] New version is less than latest version in changelog"
+               exit 1
+          elif dpkg --compare-versions "$new_version" eq "$changelog_version"
+          then opts=--no-changelog
+          elif ! dpkg --compare-versions "$new_version" gt "$changelog_version"
+          then echo "[ERROR] Could not compare new version '$new_version' to changelog version '$changelog_version'"
+               exit 1
+          fi
+          ~/auto shipit $opts
```
yarikoptic, line 54: elaborate but looks kosher... I was fearing that the empty `$opts` would freak out `auto` but my shell foo was weak -- should be ok unless it is `"$opts"`:
```
lena:/tmp
$> opts=                   

$> ./printargs.sh 1 2 $opts
> 1
> 2

$> ./printargs.sh 1 2 "$opts"
> 1
> 2
> 

$> cat printargs.sh 
#!/bin/bash
for a in "$@"; do
    echo "> $a"
done
```

yarikoptic (approved):
@mih et al - please have a look too.  With this custom-changelog-makers will not be ignored! ;)

yarikoptic:
ah -- it was rebased against maint, that is why probably linters complain!

mih:
So the approach will be to commit a changelog (or define its version) just before a release? Sounds good to me. Thx!

yarikoptic:
> So the approach will be to commit a changelog (or define its version) just before a release? Sounds good to me. Thx!

yes. Can as well work if that CHANGELOG adding PR is the one which triggers the release (has `release` label) but note that IIRC it should be against `maint` branch.  So pretty much it could be the PR which proposes to merge `master` into `maint` and has such a custom CHANGELOG entry and `release` label. IMHO kinda neat ;)

.github/workflows/release.yml
```diff
@@ -34,7 +34,24 @@ jobs:
...
+          opts=
+          if dpkg --compare-versions "$new_version" lt "$changelog_version"
+          then echo "[ERROR] New version is less than latest version in changelog"
+               exit 1
+          elif dpkg --compare-versions "$new_version" eq "$changelog_version"
+          then opts=--no-changelog
+          elif ! dpkg --compare-versions "$new_version" gt "$changelog_version"
+          then echo "[ERROR] Could not compare new version '$new_version' to changelog version '$changelog_version'"
+               exit 1
+          fi
```
yarikoptic, line 53: an idea!  @jwodder -- would it be possible to move this checking into a separate step, make this workflow run also on PRs which have `release` label, but without actually doing `~/auto shipit`?
Then we could be certain that `release` PR is a kosher one.
Workflow should react on change to labels (like the other one).

jwodder: @yarikoptic No, if the workflow is running on an open `release` PR, `auto` won't consider the project eligible for a release (as the PR isn't merged yet), and so `auto shipit --dry-run --quiet` will output nothing.

yarikoptic: damn, it was a good idea! ;)  just to make sure -- it wouldn't work even if we adjust `"baseBranch": "maint"` to point to current to PR merge branch (I guess there is none on github) in `.autorc`, right?

jwodder: @yarikoptic I *think* that wouldn't work, as then `auto` wouldn't pick up the PRs that had already been merged into `maint`, and so the version bump level would be potentially inaccurate.

yarikoptic:
ok, let's proceed on this in preparation for 0.15 . Should be glorious! ;)

PR #5954, opened 2021-09-06
Change CapturedException's behavior and intended usage

### LINKED ISSUE #5848: CaptureException must be used correctly throughout the code base

bpoldrack:
Replaces PR #5883 as discussed and summarized in https://github.com/datalad/datalad/pull/5883#issuecomment-900348975

The linked comment suggests how to use that. Except from places where `exc_str` was used to create a `message` in a (result) dict, this PR only provides the means - it does NOT fully implement the usage pattern. This would again be a huge diff and (I think) should involve similar changes to the higher level log messages as its predecessor. From my POV it's better to get this merged as is and I can then provide those changes in reviewable and discussable batches against `maint` - no need to include them in a particular release. Therefore `exc_str` still exists (but doesn't return empty) and is still used. However, there's no need for `exc_str_old` anymore, hence removed.

Also removed config `datalad.log.exc` again. Since the detailed logging is at low level, I think there's no need to additionally enable/disable it.

Closes #5848 

UPDATE: Please consider the included design document.

### DIFF
--- a/datalad/support/exceptions.py
+++ b/datalad/support/exceptions.py
@@ -9,12 +9,16 @@
 """ datalad exceptions
 """
 
+import logging
 import re
 import sys
 import traceback
 from os import linesep
 from pathlib import Path
 from pprint import pformat
+
+
+lgr = logging.getLogger('datalad.support.exceptions')
 
 
 class CapturedException(object):
@@ -26,7 +30,8 @@ class CapturedException(object):
     reporting.
     """
 
-    def __init__(self, exc=None, limit=None, capture_locals=False):
+    def __init__(self, exc=None, limit=None, capture_locals=False,
+                 level=8, logger=None):
         """Capture an exception and its traceback for logging.
 
         Clears the exception's traceback frame references afterwards.
@@ -62,6 +67,10 @@ class CapturedException(object):
             )
             traceback.clear_frames(exc.__traceback__)
 
+        # log the captured exception
+        logger = logger or lgr
+        logger.log(level, "%r", self)
+
     def format_oneline_tb(self, limit=None, include_str=True):
         """Format an exception traceback as a one-line summary
 
@@ -76,21 +85,11 @@ class CapturedException(object):
         from datalad import cfg
 
         if include_str:
-            # try exc message
-            leading = str(self.tb)
-            if not leading:
-                # go with type
-                leading = self.tb.exc_type.__qualname__
+            # try exc message else exception type
+            leading = self.message() or self.name()
             out = "{} ".format(leading)
         else:
             out = ""
-
-        if limit is None:
-            # TODO: config logging.exceptions.traceback_levels = 1
-            #       ^ This is taken from exc_str(). What exactly does it mean?
-            #         Controlling the tblimit differently for logging, result
-            #         reporting, whatever else?
-            limit = int(cfg.obtain('datalad.exc.str.tblimit', default=1))
 
         entries = []
         entries.extend(self.tb.stack)
@@ -99,6 +98,9 @@ class CapturedException(object):
         elif self.tb.__context__ and not self.tb.__suppress_context__:
             entries.extend(self.tb.__context__.stack)
 
+        if limit is None:
+            limit = int(cfg.obtain('datalad.exc.str.tblimit',
+                                   default=len(entries)))
         if entries:
             tb_str = "[%s]" % (','.join(
                 "{}:{}:{}".format(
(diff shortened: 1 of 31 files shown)


bpoldrack:
FTR: Appveyor failure unrelated

mih (changes requested):
Thx for the update.

There is a lot of valuable information scattered across the PR description and the commit messages. I think these should go into a design document. I do think that without such a description will will not manage to adopt the new system consistently.

mih (changes requested):
Thanks. The design doc makes it much more clear.

I think it should be mentioning the associated config settings that are available for tweaking the reporting (traceback details, etc), to have all things in one place.

christian-monch:
Thanks Benjamin for the work. I like it.

I would like to suggest something that might be added in the future in another issue. When we store exception information we might also provide some "context" (not defined in detail here). Contexts could later be used by an exception "explainer" module to describe the exception and its definitive or probable cause.  That would separate the exception catching from the exception explanation (which can be implemented at a different point in time)

What do you think?

datalad/support/exceptions.py
```diff
@@ -62,6 +67,10 @@ def __init__(self, exc=None, limit=None, capture_locals=False):
             )
             traceback.clear_frames(exc.__traceback__)
 
+        # log the captured exception
```
mih, line 70: I can only attach the comment here, but it really is about the optional nature of `exc` argument. Why is this done? Just because it is possible? What are the performance implications of `exc_info`? Under which conditions would I pass or not pass `exc`? My gut feeling is that `exc` should be a mandatory parameter, and exceptions should be caught with something like `Exception as e` and pass in, instead of promoting (double) lookups.

datalad/support/exceptions.py
```diff
@@ -112,31 +114,51 @@ def format_oneline_tb(self, limit=None, include_str=True):
...
+    def message(self):
+        """Returns only the message of the original exception
+
+        Returns
+        -------
+        str
+        """
+        return str(self.tb)
+
+    def name(self):
```
mih, line 148:
```suggestion
    @property
    def name(self):
```

Maybe?

mih (commented):
Great! I went through the full diff and only minor aspects caught my attention. I'd say unless objections are expressed, let's merge this Friday EOB.

mih (approved):

datalad/support/exceptions.py
bpoldrack: TBH, that's a remainder from previous iteration with the purpose of being "compatible" with old `exc_str` behavior. I do agree, though, and I'm happy to remove it.

datalad/support/exceptions.py
bpoldrack: Yes

datalad/support/exceptions.py
bpoldrack: According to git grep there's only one spot, where `exc_str` is called without argument. So - will adjust that spot and make it mandatory.

Issue #5938, opened 2021-08-30
Deprecate/remove `--pbs-runner`


mih:
This core option claims that _any_ command can be submitted to a batch system. I think we should remove it from the API. Rational:

- for most commands it makes no sense to submit them to a batch system
- for those commands where submitting to a batch system is meaningful, there is not sufficient parameterization capability to work with anything other than NFS-style filesystem setups across the distributed resource
- there is no support for prevent conflicting operations done to the datasets that are subject to a command, when submitted in  parallel
- only a single system is "supported" (condor), and in the many years of datalad not a single one was added. I consider this an indication for under-use.
- the name is suboptimal: PBS is a particular and ancient batch system, and not even supported
- current implementation has little or no error handling (e.g. plain crash when `condor_submit` is not found
- current implementation does not work on my test condor system (see log below)

#### Conclusion

I cannot come up with a use case, where simply wrapping a (tuned) datalad command call into `condor_run` would not be superior to the `-pbs-runner` option use.

```
% datalad --pbs-runner condor siblings
Submitting job(s).
1 job(s) submitted to cluster 613835.
[INFO   ] Scheduled execution via condor.  Logs will be stored under /tmp/datalad-condor-ctzfh3d9.log 

...
<long time passed, job long finished>
% cat /tmp/datalad-condor-ctzfh3d9.log
cat: /tmp/datalad-condor-ctzfh3d9.log: No such file or directory

% condor_run datalad siblings -d $(pwd)
.: here(?) [git]
.: gh-ben(?) [git@github.com:bpoldrack/datalad.git (git)]
.: gh-master(?) [git@github.com:datalad/datalad.git (git)]
.: origin(?) [git@github.com:mih/datalad.git (git)]
```

Note the `-d` switch is only needed, because not all condor deployments put a job's PWD to be the submit PWD.

yarikoptic:
I have used it extensively in the past for crawling large collections of datasets, but haven't used it for ages.  As such, unused and under-tested and designed on the sun rise of the prior decade, I think indeed we can just remove it.  As always, never know if someone is using it, so I guess we should announce it deprecated in 0.15.0 and remove in 0.16.0

PR #5956, opened 2021-09-07
Deprecate --pbs-runner

mih:
Fixes datalad/datalad#5938

### DIFF
--- a/datalad/cmdline/common_args.py
+++ b/datalad/cmdline/common_args.py
@@ -50,7 +50,7 @@ pbs_runner = (
     'pbs-runner', ('--pbs-runner',),
     dict(choices=['condor'],
          default=None,
-         help="""execute command by scheduling it via available PBS.  For settings, config file will be consulted""")
+         help="""DEPRECATED, will be removed in a future release.""")
 )
 
 change_path = (
--- a/datalad/cmdline/helpers.py
+++ b/datalad/cmdline/helpers.py
@@ -313,6 +313,10 @@ def strip_arg_from_argv(args, value, opt_names):
 
 
 def run_via_pbs(args, pbs):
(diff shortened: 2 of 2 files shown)


datalad/cmdline/helpers.py
```diff
@@ -313,6 +313,10 @@ def strip_arg_from_argv(args, value, opt_names):
 
 
 def run_via_pbs(args, pbs):
+    warnings.warn("Job submission via --pbs-runner is deprecated."
+                  "Use something like condor_run",
```
yarikoptic, line 317:
```suggestion
                  "Use something like condor_run from datalad-htcondor extension",
```

mih: Err, no. I meant the standard HTCondor command. I do not want to advertise `datalad-htcondor`.

yarikoptic: ah, misunderstood. ok

datalad/cmdline/common_args.py
```diff
@@ -50,7 +50,7 @@
     'pbs-runner', ('--pbs-runner',),
     dict(choices=['condor'],
          default=None,
-         help="""execute command by scheduling it via available PBS.  For settings, config file will be consulted""")
+         help="""DEPRECATED, will be removed in a future release.""")
```
yarikoptic, line 53: useful would have been to add a comment with version (0.16) when it is ok to remove, or someone would need to do archaeological expedition to figure out for sure if it is ok time to remove

don't want to hold this PR with this, so let's proceed

mih: Filed https://github.com/datalad/datalad/issues/5962 with milestone 0.16

Issue #5959, opened 2021-09-07
#5934 broke master (syntax warning)


mih:
Windows tests revealed it 

```
C:\Python39-x64\Lib\site-packages\datalad\interface\utils.py:598: SyntaxWarning: "is" with a literal. Did you mean "=="?
918  if result_log_level is 'match-status'
```

Saw it in https://github.com/datalad/datalad/pull/5956, but it seems @bpoldrack has already spotted it in https://github.com/datalad/datalad/pull/5954#issuecomment-913581131 but filed no issue.

Likely broken by https://github.com/datalad/datalad/pull/5934 Update: confirmed broken by #5934

Somehow `None` is assigned a better value of `'match-status'`

PR #5960, opened 2021-09-07
BF: use `==` not `is`

mih:
Introduced by an incomplete replacement in 3a8fd827a8bb21bee91815fe133422356aedf777

Fixes datalad/datalad#5959


yarikoptic (approved):
LGTM although not sure how it could lead to any issue (besides a warning) since afaik strings are singletons internally in python

PR #5963, opened 2021-09-07
ENH: Add ability to expand downloaders from entry points

yarikoptic:
Would be needed if an extension to introduce some new downloader support

### DIFF
--- a/datalad/downloaders/providers.py
+++ b/datalad/downloaders/providers.py
@@ -96,6 +96,19 @@ class Provider(object):
         self.authenticator = authenticator
         self._downloader = downloader
 
+    @classmethod
+    def _extend_from_entry_points(cls):
+        from pkg_resources import iter_entry_points
+        # TODO: it seems we are hitting this multiple times, why?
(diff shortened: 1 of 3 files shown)


mih (approved):
Makes sense!

mih:
Any reason this has not made it out of draft mode yet?

yarikoptic:
added/adjusted docs on entry_points listing.  Taking out of the draft

mih:
Only checked on the phone, but there was still a `pdb` call in the diff. Shouldn't be there, right?

yarikoptic:
d'oh -- thanks! I am afraid it was there for a reason that it was not quite finished and I submitted just to seek feedback first.

well -- I think I will take it back into draft mode: I thought I already had a downloader to hook up, and most likely it was the one for fsspec (within datalad-fuse) but can't locate it ATM.  Need to craft that one and then tune up unittest to make sure all is good.

mih:
Let's close this PR until work on it resumes. We have almost 20 PRs open.

PR #5966, opened 2021-09-08
Consolidate appveyor setup (fewer jobs, recent git-annex on linux)

mih:
Each build now uses 9 jobs/workers, out of 10 available with a target runtime of ~45min.


mih:
Overall runtime is <50min with a single job determining the max time (same as before, but two jobs less).

Failure is unrelated #5967

Issue #5967, opened 2021-09-08
appveyor windows error: test_clone.test_invalid_args


mih:
```
======================================================================
870ERROR: datalad.core.distributed.tests.test_clone.test_invalid_args
871----------------------------------------------------------------------
...
881  File "C:\Python39-x64\lib\site-packages\datalad\core\distributed\tests\test_clone.py", line 109, in test_invalid_args
882    res = clone('Zoidberg', path='ssh://mars:Zoidberg', on_failure='ignore')
...
892    yield from clone_dataset(
893  File "C:\Python39-x64\lib\site-packages\datalad\core\distributed\clone.py", line 505, in clone_dataset
894    srcs = _map_urls(cfg or destds.config, srcs)
895  File "C:\Python39-x64\lib\site-packages\datalad\distribution\dataset.py", line 344, in config
896    repo = self.repo
897  File "C:\Python39-x64\lib\site-packages\datalad\distribution\dataset.py", line 290, in repo
898    self._repo = repo_from_path(self._path)
899  File "C:\Python39-x64\lib\site-packages\datalad\core\local\repo.py", line 56, in repo_from_path
900    if not cls.is_valid_repo(path, **ckw):
901  File "C:\Python39-x64\lib\site-packages\datalad\support\annexrepo.py", line 670, in is_valid_repo
902    return initialized_annex or GitRepo(path, create=False, init=False).is_with_annex()
903  File "C:\Python39-x64\lib\site-packages\datalad\dataset\repo.py", line 146, in __call__
904    id_, new_args, new_kwargs = cls._flyweight_id_from_args(*args, **kwargs)
905  File "C:\Python39-x64\lib\site-packages\datalad\dataset\repo.py", line 234, in _flyweight_id_from_args
906    path_postproc = cls._flyweight_postproc_path(localpath)
907  File "C:\Python39-x64\lib\site-packages\datalad\dataset\repo.py", line 193, in _flyweight_postproc_path
908    resolved = str(ut.Path(path).resolve().absolute())
909  File "C:\Python39-x64\lib\pathlib.py", line 1215, in resolve
910    s = self._flavour.resolve(self, strict=strict)
911  File "C:\Python39-x64\lib\pathlib.py", line 215, in resolve
912    s = self._ext_to_normal(_getfinalpathname(s))
913OSError: [WinError 123] The filename, directory name, or volume label syntax is incorrect: 'C:\\projects\\datalad\\__testhome__\\ssh:\\mars:Zoidberg'
[...]

PR #5968, opened 2021-09-08
`Path.exist()` no longer raises OSError for non-existing malformed paths p…

mih:
Use `Path.resolve(strict=False)` as a more reliable, but also more
expensive test.

Fixes datalad/datalad#5967

### DIFF
--- a/datalad/core/distributed/clone.py
+++ b/datalad/core/distributed/clone.py
@@ -291,7 +291,8 @@ class Clone(Interface):
             # we are running on -- we don't care if the path actually
             # exists at this point, but we want to abort early if the path
             # spec is determined to be useless
-            path.exists()
+            # we can do strict=False since we are 3.6+
+            path.resolve(strict=False)
         except OSError as e:
             yield get_status_dict(
                 status='error',


mih:
Test failure is unrelated (slow CI worker)

yarikoptic:
> Path.exist() no longer raises OSError for non-existing malformed paths

yikes -- is that a new feature? 

NB https://github.com/datalad/datalad/pull/5968/commits/716e248b3a7c72f6fac9d99215add58d8a6e2607 commit msg didn't provide the answer 

if not a feature -- smells like a bug and thus better be reported/fixed instead of us working around it "always"!

yarikoptic:
d'oh -- looked at the diff/code. -- nevermind above comment 

FWIW, use of `.resolve` indeed provides extra testing which could be of benefit (even on 3.8 and may be before), e.g.:

```
>>> sys.version
'3.8.5 (default, Sep  3 2020, 21:29:08) [MSC v.1916 64 bit (AMD64)]'
>>> import pathlib
>>> pathlib.Path('con').exists()
True
>>> pathlib.Path('con').resolve()
  File "<stdin>", line 1, in <module>
  File "C:\Users\DataLad\miniconda3\envs\datalad\lib\pathlib.py", line 1177, in resolve
    s = self._flavour.resolve(self, strict=strict)
  File "C:\Users\DataLad\miniconda3\envs\datalad\lib\pathlib.py", line 205, in resolve
    s = self._ext_to_normal(_getfinalpathname(s))
OSError: [WinError 87] The parameter is incorrect: 'con'
>>> pathlib.Path('contr').resolve()
WindowsPath('contr')
```

Issue #5489, opened 2021-03-12
Migrate `configuration` command into datalad proper


mih:
It is presently provided by an extension and has to be installed separately. However, I find it rather handy, and consider it a candidate for inclusion into core.

https://github.com/mih/datalad-mihextras

https://datalad-mihextras.readthedocs.io/en/latest/generated/datalad.api.x_configuration.html

It was briefly already available, but was removed https://github.com/datalad/datalad/pull/5370

mih:
Opinions have not changed. Taking it off the 0.15 milestone.

yarikoptic:
> The alternatively requested renaming of parameters and their values would only establish superficial alignment of the APIs

Re `2` and `3` which are IMHO a singular aspect. As we have discussed on the call, renaming could as well be done at `ConfigManager` level  (`where` -> `scope`; `dataset` -> `branch`) through a deprecation cycle (e.g. add for 0.15.0 and deprecate/remove old style in 0.16.0 or later ;) ), thus gently transitioning/preparing users to use `configuration` options and not requiring them to remember what `option=value` to use with what API endpoint. IMHO it would not be a "superficial alignment" since AFAIK they refer to exactly the same thing, and I do like `scope` and `branch` better (even surprised I did not complain/suggest it whenever `where="dataset"` was introduced in #922).  And FWIW I don't care about `4` (`system`-level) and also do not see how to avoid `1` in a reasonable effort.

mih:
Ok, so someone should do this ConfigManager PR within the next 6 month.

PR #5969, opened 2021-09-08
RF: config  'where[="dataset"]'->'scope[="branch"]

yarikoptic:
A follow up to the call https://github.com/datalad/datalad/issues/5489#issuecomment-915404977 -- as I see it could potentially be done

- [x] datalad.tests.test_config pass
- [x] decide on the destiny of `source` which underlies where/scope
- [x] comb through docstrings/comments on what to revert back to use word "where"
- [x] investigate if there is any notable effect on performance due to decoration/checks
- [x] extend tests to check (and likely fix) deprecated usage still working out

note: initial replacements are done using `datalad run` so in principle this PR might be `datalad rerun`able if we start breeding conflicts... would be fun to check -- is opening up a DX usage scenario for `run/rerun`.


datalad/config.py
```diff
@@ -118,7 +147,7 @@ def parse_gitconfig_dump(dump, cwd=None, multi_value=True):
         # line is a null-delimited chunk
         k = None
         # in anticipation of output contamination, process within a loop
-        # where we can reject non syntax compliant pieces
+        # scope we can reject non syntax compliant pieces
```
yarikoptic, line 150: yet another victim in text of the replace. TODO:
```suggestion
        # where we can reject non syntax compliant pieces
```

datalad/config.py
```diff
@@ -74,12 +80,35 @@ def get_git_version(runner=None):
...
 
+#
+# TODO: remove in/after 0.16.0
+#
+def _where_to_scope(func):
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        if 'where' in kwargs:
+            if 'scope' in kwargs:
+                raise ValueError("Do not specify both 'scope' and DEPRECATED 'wraps'")
```
mih, line 97:
```suggestion
                raise ValueError("Do not specify both 'scope' and DEPRECATED 'where'")
```

datalad/config.py
```diff
@@ -74,12 +80,35 @@ def get_git_version(runner=None):
...
+            kwargs = kwargs.copy()
+            where = kwargs.pop('where')
+            if where == 'dataset':
+                warnings.warn("'where=\"dataset\"' is being deprecated as of 0.15. Use 'scope=\"branch\"' instead",
```
mih, line 101:
```suggestion
                warnings.warn("'where=\"dataset\"' is deprecated, use 'scope=\"branch\"' instead",
```

I think we stick with simple statements.

datalad/config.py
```diff
@@ -74,12 +80,35 @@ def get_git_version(runner=None):
...
+                              DeprecationWarning)
+                where = 'branch'
+            else:
+                warnings.warn("'where' is being deprecated as of 0.15. Use 'scope' instead",
```
mih, line 105:
```suggestion
                warnings.warn("'where' is deprecated, use 'scope' instead",
```

datalad/config.py
```diff
@@ -265,6 +294,7 @@ class ConfigManager(object):
       not global or system configuration are considered; if 'any'
       all possible sources of configuration are considered.
     """
+    # TODO: above dataset* -> branch*?
```
mih, line 297: and `dataset-local` -> `branch-local`

datalad/config.py
```diff
@@ -74,12 +80,35 @@ def get_git_version(runner=None):
...
-def _where_reload(obj):
+def _scope_reload(obj):
     """Helper decorator to simplify providing repetitive docstring"""
-    obj.__doc__ = obj.__doc__ % _where_reload_doc
+    obj.__doc__ = obj.__doc__ % _scope_reload_doc
     return obj
 
 
+#
+# TODO: remove in/after 0.16.0
```
mih, line 90: This is way too short.  Superficially looking through code on my laptop I see usage in -metalad, -container, -crawler, -webapp, -ukbiobank, -xnat, -hirni,  -osf, -mihextra, -neuroimaging, -deprecated. Plus hits in internal code, including individual analyses implementations. Given the breadth, I think we must assume that we break most non-trivial code somewhere when we stop supporting `where=`. I'd say 2-3 years is a more reasonable *minimum* threshold.

Realistically speaking, we can expect client library/tool code to change over time. But archived one-off implementations will be unlikely to be adjusted. This is of course not different from any other deprecation, but it touches such a core piece of code that the impact is more substantial in comparison. What pains me most is that any code that uses the metadata system (where config must be altered in `.datalad/config`) will be broken.

mih (changes requested):
First set of comments. Will be interesting to see performance impact estimates.

datalad/config.py
```diff
@@ -372,7 +403,7 @@ def reload(self, force=False):
         # 2-step strategy:
         #   - load datalad dataset config from dataset
         #   - load git config from all supported by git sources
-        # in doing so we always stay compatible with where Git gets its
+        # in doing so we always stay compatible with scope Git gets its
```
adswa, line 406:
```suggestion
        # in doing so we always stay compatible with where Git gets its
```

mih:
re peformance; it seems to be kosher:

Using `timeit cfg.set('mike.this', 'value', where='global')`

```py
# with decorator, old call
>>> timeit cfg.set('mike.this', 'value', where='global')
10.4 ms ± 202 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

# with decorator, new call
timeit cfg.set('mike.this', 'value', scope='global')
10.1 ms ± 45.3 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

# without decorator, new call
timeit cfg.set('mike.this', 'value', scope='global')
9.96 ms ± 10.9 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

IOW: <5% performance slowdown for old unchanged code, <2% in comparison to a hard transition (breakage) to a new API.

Given that only *write* access is affected, this is tolerable IMHO.

datalad/config.py
yarikoptic: ok, let's do those as well then.  And thank you for the timing results!

datalad/config.py
yarikoptic: BTW, touches on the topic where a better formalized support for deprecations would help -- it is useful to know (without discovery via commits etc analysis) when did we deprecated the feature, so when you encounter such a warning and know current version you can right away assess that may be it is time to remove it (and if you don't know the version when deprecated -- you would just skip/postpone/ignore since might have been in the same "major" release)

datalad/config.py
yarikoptic:
```suggestion
# TODO: remove when deprecated 'where' is removed
```

yarikoptic: agree -- deprecation->removal should indeed be more "temporal" and not semver based!

yarikoptic:
@mih -- we should either strive to adhere to pep8 from now on (e.g. in diffs at least; may be with a quick pep8-ification maraphone after 0.16.0) or disable pep8 checks on codeclimate. What would you prefer?

datalad/config.py
```diff
@@ -73,12 +79,35 @@ def get_git_version(runner=None):
...
 
 
+#
+# TODO: remove when deprecated 'where' is removed
+#
+def _where_to_scope(func):
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        if 'where' in kwargs:
+            if 'scope' in kwargs:
```
yarikoptic, line 95: todo -- add a unittest for this deprecation logic so it is test covered and that there is no surprises when/if it is used

mih:
> @mih -- we should either strive to adhere to pep8 from now on (e.g. in diffs at least; may be with a quick pep8-ification maraphone after 0.16.0) or disable pep8 checks on codeclimate. What would you prefer?

I don't care. I consider this really minor in comparison to all other issues. But I know that without a check being done, nothing will ever change.

mih:
Given the config-related PR will not land in 0.16, there is no pressure to revive this stalled PR. Let's close it.

yarikoptic:
Well, there is no pressure I guess, but I did revive it and updated to master only few days ago. Still hopeful to see it reviewed/merged

yarikoptic:
pushed resolved conflicts and tune ups needed which I didn't miss (see e.g. 280f6f408ae87071e8160f5febcf160a2b7106e6 which shows "alignment" brought by this PR between `configuration` and `config`.

datalad/config.py
yarikoptic: done in 05a909c449aa7e3c44ed5f0b2e4c4ab7d20fc516

yarikoptic:
@mih:
> Hence I will not object this PR, if this is the only way forward.

There are always choices, e.g. here "merge" or "close". Since no stronger objection against "merge", I will take it.

@bpoldrack :
> Ultimately I am OK with a merge.

Coolio, thank you.

Overall, I think such harmonization is benefitial. Due to deprecation compatibility layer it should be not effecting users much, even if some rare user uses `scope` explicitly.  So I feel that it is a good way forward regardless either we ever see fusion of `config` and `configuration` or "privatization" of `config`.  Let's proceed so it doesn't hang over any longer.

PR #5974, opened 2021-09-10
Avoid using auto's exec plugin as much as possible

### LINKED ISSUE #5972: Release workflow failed
https://github.com/datalad/datalad/runs/3568118676?check_suite_focus=true

error     Received E2BIG from execSync.

I guess we could now just cut paste that changelog and retry to release with that PR, right @jwodder ?

jwodder:
Fixes #5972 (hopefully).

See also https://github.com/intuit/auto/discussions/2072, where I have asked about an alternative to enabling & disabling the `afterChangelog` hook.

### DIFF
--- a/.autorc
+++ b/.autorc
@@ -8,8 +8,7 @@
         [
             "exec",
             {
-                "afterChangelog": "make update-changelog && git add docs/source/changelog.rst && git commit -m '[skip ci] Update RST changelog'",
-                "afterRelease": "python -m build && twine upload dist/*"
+                "afterChangelog": "make update-changelog && git add docs/source/changelog.rst && git commit -m '[skip ci] Update RST changelog'"
             }
         ]
     ]
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,6 +10,8 @@ jobs:
   auto-release:
     runs-on: ubuntu-latest
     if: "!contains(github.event.head_commit.message, 'ci skip') && !contains(github.event.head_commit.message, 'skip ci')"
+    outputs:
+      auto-version: ${{ steps.auto-version.outputs.version }}
     steps:
       - name: Checkout source
         uses: actions/checkout@v2
@@ -22,13 +24,13 @@ jobs:
           wget -O- "$auto_download_url" | gunzip > ~/auto
           chmod a+x ~/auto
 
-      - name: Set up Python
-        uses: actions/setup-python@v2
-        with:
-          python-version: '^3.8'
-
-      - name: Install Python dependencies
-        run: python -m pip install build twine
+      - name: Check whether a release is due
+        id: auto-version
+        run: |
+          version="$(~/auto version)"
+          echo "::set-output name=version::$version"
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Install pandoc
         run: sudo apt-get update && sudo apt-get install -y pandoc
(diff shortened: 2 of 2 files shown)


yarikoptic:
ouch, quite an extensive diff. But since it now would not need a workaround for large changelogs to be copy/pasted (I assume) then I guess we better try it out.  So, if I guess we need to add `release` label and merge to see if works, right?

jwodder:
@yarikoptic 

> it now would not need a workaround for large changelogs to be copy/pasted

If a release is made without pre-supplying the changelog, and it fails due to the E2BIG problem, you'll have to supply the changelog manually (possibly by copying it from the output from the failed release) and then release again in order for the release to succeed.  I assumed that's what you were getting at with [your workaround idea](https://github.com/datalad/datalad/issues/5972#issuecomment-917014847).

Avoiding this fail-and-retry entirely would require either (a) always pre-supplying the changelog (which defeats most of the point of `auto`), (b) replacing the `exec` plugin with a local plugin ([which I can't get to work](https://github.com/intuit/auto/issues/2019)), or (c) rewriting `auto shipit` in terms of constituent commands (See the link in the comment above).

yarikoptic:
If we are to go for the former (still needing manual cut paste), where is my workaround idea was insufficient - some other command uses exec?

jwodder:
@yarikoptic When `--no-changelog` is not in effect, the `auto`-generated changelog still needs to be converted to rST after the Markdown version is generated and before the release is released.  Under this PR, that's handled by an `afterChangelog` hook of the `exec` plugin.

yarikoptic:
Sorry, failing to grasp the difference from what I proposed ... will look later when having a clearer head

yarikoptic:
For now I just manually published 0.14.8 while copy/pasting that change log from the failed actions run.  We should just not delay releasing any other maint if any to come before (or after) 0.15.0 so changelog stays small, and then can release "manually" as in old times when fails... I still hope that `auto` would fix it up one way or another.

mih:
I am not at all familiar with the involved technology, but this seems stuck without any activity from the people that could make it happen. Unless stopped, I would close this PR in a few days.

mih:
I cannot see any record of it, but @bpoldrack tried to release 0.15.4 via auto and failed. This means that but major and minor released are no longer possible via this system. Should we go back to normal?

yarikoptic:
my wild guess is that it is due to the renaming of labels to have `semver-` prefix but I could be wrong.  This is what I see
```
(git)lena:~datalad/datalad-maint[maint]git
$> ~/bin/intuit-auto shipit --dry-run        
✖  error     Found configuration errors:

"labels.3.releaseType"

Expected ""none"" or ""skip"" or ""major"" or ""minor"" or ""patch"" or ""skip"" or ""release"" but got: dependencies

"labels.4.releaseType"

Expected ""none"" or ""skip"" or ""major"" or ""minor"" or ""patch"" or ""skip"" or ""release"" but got: documentation

"labels.5.releaseType"

Expected ""none"" or ""skip"" or ""major"" or ""minor"" or ""patch"" or ""skip"" or ""release"" but got: internal

"labels.6.releaseType"

Expected ""none"" or ""skip"" or ""major"" or ""minor"" or ""patch"" or ""skip"" or ""release"" but got: performance

"labels.7.releaseType"

Expected ""none"" or ""skip"" or ""major"" or ""minor"" or ""patch"" or ""skip"" or ""release"" but got: tests
 

⚠  warning   These errors are for the fully loaded configuration (this is why some paths might seem off).
(dev3) 1 29645 ->1.....................................:Wed 15 Dec 2021 04:09:57 PM EST:.
(git)lena:~datalad/datalad-maint[maint]git
$> git grep semver
.autorc:    { "releaseType": "major", "name": "semver-major" },
.autorc:    { "releaseType": "minor", "name": "semver-minor" },
.autorc:    { "releaseType": "patch", "name": "semver-patch" },
.autorc:    { "releaseType": "dependencies", "name": "semver-dependencies" },
.autorc:    { "releaseType": "documentation", "name": "semver-documentation" },
.autorc:    { "releaseType": "internal", "name": "semver-internal" },
.autorc:    { "releaseType": "performance", "name": "semver-performance" },
.autorc:    { "releaseType": "tests", "name": "semver-tests" }
```

@jwodder - any ideas on what actual issue is?

jwodder:
@yarikoptic Looks like using "dependencies", "documentation", "internal", "performance", and "tests" as `releaseType` values in `.autorc` is the wrong thing to do, and those values should all be `"none"`.

yarikoptic:
Thanks @jwodder. Let's see -- submitted https://github.com/datalad/datalad/pull/6320

yarikoptic:
@mih -- https://pypi.org/project/datalad/ shows new 0.15.4, so #6320 fixed it all!

mih:
So what is the status here? All is working, but this is just hanging here? I assume it is no longer necessary, hence closing. Feel free to reopen, if that is a false assessment.

Issue #5962, opened 2021-09-07
Remove deprecated --pbs-runner option


mih:
Deprecation in https://github.com/datalad/datalad/pull/5956

PR #5981, opened 2021-09-15
Remove deprecated --pbs-runner cmdline option

mih:
Was deprecated with the 0.15.0 release, see gh-5956.

Fixes datalad/datalad#5962

### DIFF
--- a/datalad/cmdline/common_args.py
+++ b/datalad/cmdline/common_args.py
@@ -46,13 +46,6 @@ log_level = (
               % ', '.join(_log_level_names))
 )
 
-pbs_runner = (
-    'pbs-runner', ('--pbs-runner',),
-    dict(choices=['condor'],
-         default=None,
-         help="""DEPRECATED, will be removed in a future release.""")
-)
-
 change_path = (
     'change-path', ('-C',),
     dict(action='append',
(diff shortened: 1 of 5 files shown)


adswa (approved):
LGTM

yarikoptic:
Thank you but:
 I guess it is great to have those PRs so we do not "forget" to remove deprecated clutter, but since they would be just hanging here cluttering the list of PRs.  If we establish `next` branch -- we could merge them there but then we would need to deal with conflicts etc...  But also I thought we agreed on that removal of deprecated functionality should happen more on "time" than on "release" basis to provide users enough time.  We don't know when we would release 0.16.0 .

mih:
My intention was to have to merged now-ish. This is positioned against master. The next release coined from master will be 0.16.0. And with what I learned from the 0.15.0 release actually no release will ever come from master.

Are you saying that there should be a minimum time after a release for changes in master scheduled for the next non-maint release? If so, what would be the delay?

Moreover, given that deprecations are not visible enough, would it not call for removals to happen at the earliest time in a release cycle to actually catch breakages during development, rather than post-release?

yarikoptic:
sorry, I was slow -- right, it will then go to 0.16.0 . all good indeed!

PR #5986, opened 2021-09-15
BF(TST): use sys.executable, mark test_ria_basics.test_url_keys as requiring network

yarikoptic:
Detected while building Debian package .

Also did isort on those tests files


mih (approved):
Thx!

PR #5988, opened 2021-09-16
Restructure cmdline API presentation

mih:
Introduce the notion of core commands, remove the notion of 'plugins'.
It is now also no longer a plain list of command names.


mih:
Unless there are objections, I will merge this documentation update later today.

PR #5991, opened 2021-09-17
BF: downloader - fail to download even on non-crippled FS if symlink exists

yarikoptic:
Observed initially by @jwodder on
https://github.com/datalad/datalad-fuse/pull/3#issuecomment-921934544
since datalad-extensions-template default workflow runs on a crippled FS.

I think it makes sense to fail regardless either it is existing file or a
(broken) symlink since the idea is not to override a (possibly annexed) file

### DIFF
--- a/datalad/downloaders/base.py
+++ b/datalad/downloaders/base.py
@@ -436,9 +436,9 @@ class BaseDownloader(object, metaclass=ABCMeta):
         else:
             filepath = downloader_session.filename
 
-        existed = exists(filepath)
+        existed = op.lexists(filepath)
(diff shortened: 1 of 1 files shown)


yarikoptic:
crawler fails are likely unrelated: https://github.com/datalad/datalad-crawler/issues/104 .

If no objections or other feedback -- will merge in a day or so.  To make `auto` auto-mint release, might even merge with `release` label unless some other PR would "release" us sooner.

mih (approved):
LGTM! I will leave the merge to @yarikoptic to preserve the possibility to add the release tag.

PR #5999, opened 2021-09-22
ENH: import datalad.api to bind extensions methods for discovery of dataset methods

### LINKED ISSUE #5990: Traverse entry points if some Dataset.<method> is not found
Observed by @jwodder while working on https://github.com/datalad/datalad-fuse/pull/3#issuecomment-921822862

Even though we declared interfaces to be added via entry points, and e.g. here if we move some functionality to -deprecated (BTW -- we need a release since `publish` move was not released in datalad-deprecated, known: https://github.com/datalad/datalad-deprecated/issues/29)

```shell
$> python -c 'from datalad.distribution.dataset import Dataset; ds=Dataset("."); print(ds.publish)'     
  File "/home/yoh/proj/datalad/datalad-master/datalad/distribution/dataset.py", line 210, in __getattr__
    return super(Dataset, self).__getattribute__(attr)
AttributeError: 'Dataset' object has no attribute 'publish'
```

<details>
<summary>but it works if we import `datalad.api` as well, which then traverses the entry points:</summary> 

```shell
$> python -c 'from datalad.distribution.dataset import Dataset; from datalad.api import publish; ds=Dataset("."); print(ds.publish)'
<bound method Publish.__call__ of Dataset('/home/yoh/proj/datalad/datalad-deprecated')>
```
</details>

so I think we should add to  `Dataset.__getattr__` we already have traversal of entry points (code should be reused or RFed so not to be needed as triggered in datalad.api support code).

yarikoptic:
This might be not the most elegant but the "easiest" in DX approach to resolve
this issue IMHO.  It might cause run time penalty though in small scripts which
import only a few dedicated modules, thus do not require import of the entire
API.

An alternative would be to e.g. do 2 rounds of that loop, importing datalad.api
only upon the end of the first loop if no meth is found.

Closes #5990

### DIFF
--- a/datalad/distribution/dataset.py
+++ b/datalad/distribution/dataset.py
@@ -189,6 +189,9 @@ class Dataset(object, metaclass=PathBasedFlyweight):
             from datalad.interface.base import (
                 get_interface_groups, get_api_name, load_interface
             )
+            # load entire datalad.api which will also bind datasetmethods
+            # from extensions. Although adds overhead, good for UX
+            import datalad.api
             groups = get_interface_groups()
             for group, _, interfaces in groups:
                 for intfspec in interfaces:
--- a/datalad/distribution/dataset.py
+++ b/datalad/distribution/dataset.py
@@ -180,36 +180,14 @@ class Dataset(object, metaclass=PathBasedFlyweight):
             return self.pathobj == other.pathobj
 
     def __getattr__(self, attr):
-        # Assure that we are not just missing some late binding
-        # @datasetmethod . We will use interface definitions.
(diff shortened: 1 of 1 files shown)


yarikoptic:
crawler issue is unrelated. filed https://github.com/datalad/datalad-crawler/issues/104

mih (approved):
OK, let's do this. However, I want to put on record that I believe we might end up revisiting this. More extensions, will ultimately require more careful, less unconditional importing to avoid uncontrolled slowdown. But for now the reduced complexity is attractive!

mih:
Travis issue seems unrelated.

Issue #5995, opened 2021-09-20
HTTPDownloader not yet working with redirected http-auth


bpoldrack:
While working on providing http-auth testing via enhanced docker image from https://github.com/datalad-tester/docker-ssh-target, I looked into `download_url` in order to combine the test setup with an actually meaningful test (and possibly necessary first fix).

The setup in the docker I have reflects the setup for RIA stores we use at INM-7: HTTP is redirected to HTTPS. Against that setup `download_url` works fine if pointed to the HTTPS address. It fails with the redirected HTTP address, though. That made me stumble upon several questionable aspects:

The downloader actually sends two requests to the same URL: A POST request to establish authentication and a subsequent GET request to actually download. Since with simple default apache serving files the response to the POST request already contains the content of the file, the GET request seems superfluous (or the other way around depending on how you look at it).

- Leading to the first question in this regard: Under what circumstances is a dedicated authentication request actually needed? We use `requests.Session` object and it seems to me, that the authenticator should be assigned to the `Session` (instead of being passed to the request call) and we can simply go for an authenticated GET. Why is there a POST to begin with? Was that written with some kind of webservice in mind, that would actually require that? May be in context of the crawler? @yarikoptic any recollection? It doesn't seem to make sense to me, since this is happening within the http downloader processing a single URL.

- That the redirected download fails is due to the fact, that we auto-follow a redirect for the initial POST request (which in this case is all we need and that succeeds), but we do not do that for the subsequent GET (which then therefore is what fails). `requests` does strip the credentials from redirected request, since it's considered a security hazard to auto-send credentials to a redirect. Now, this raises two questions: Should we really do that automatically for the POST (as we do right now)? If so, why not do it with the GET then? It's the exact same URL. If we trust the one request we can trust the other one, I think. If we shouldn't do it at all, then what could be our strategy to deal with such server setups? Asking for confirmation on every redirect? Would users even know whether or not that's fine? What could a config look like, that allows to authenticate against the redirected URL? Just the target URL that is whitelisted for being the target of such a redirect? Or does it make a difference where we are redirected from? (I don't think so).

- Finally, when we follow the redirect for the initial POST, what we send to the new target suddenly becomes a GET. Why is that? It makes it even more weird to send this as mere authentication establishing request, only to now have the intended download request to be *completely* identical (with POST/GET being the only difference to begin with - that's now gone, too).

It seems to me, what we should do is:
[...]

PR #6002, opened 2021-09-23
Address #5995

bpoldrack:
For now just looking for failures in the entire test battery.

Locally there's `datalad.downloaders.tests.test_http.test_access_denied` failing with its last assertion. However, that one is a convoluted mess. The reason it's passing to begin with (in `master` that is) is a bit absurd. The POST request I ripped out, does fail with `Unsupported request` by the http server used in the test, leading to a different call path that ultimately makes the assertion pass. So, sending an unsupported POST is required to correctly fail with `AccessDeniedError`. W/O that POST request, there still is an `AccessDeniedError`, but it ends up being handled by trying to ask for credentials which then somehow fails with an `AssertionError` instead.
I think that test is simply wrong. Not yet clear how it should look like, though.

### DIFF
--- a/datalad/downloaders/http.py
+++ b/datalad/downloaders/http.py
@@ -350,21 +350,24 @@ class HTTPAuthAuthenticator(HTTPRequestsAuthenticator):
         authenticator = self.REQUESTS_AUTHENTICATOR(
             *[credentials[f] for f in self.REQUESTS_FIELDS])
         session.auth = authenticator
-        response = session.post(post_url, data={},
-                                auth=authenticator)
-        auth_request = response.headers.get('www-authenticate')
-        if response.status_code == 401 and auth_request:
-            if auth_request.lower().split(' ', 1)[0] == 'basic':
-                if response.url != post_url:
-                    # was instructed to authenticate elsewhere
-                    # TODO: do we need to loop may be??
-                    response2 = session.get(response.url, auth=authenticator)
-                    return response2
-            else:
-                lgr.warning(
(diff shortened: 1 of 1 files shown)


yarikoptic:
> For now just looking for failures in the entire test battery.

as [I have mentioned](https://github.com/datalad/datalad/issues/5995#issuecomment-924146607) -- interaction with relevant websites might not be tested on CI since it would lack such credentials.

bpoldrack:
@yarikoptic 

> as I have mentioned -- interaction with relevant websites might not be tested on CI since it would lack such credentials.

Yes, but then again, even if a particular site would need a dedicated POST, this seems to be the wrong spot to address it, since this is unconditionally called with the same url as the one we are trying to download. It would only make sense in a context where it is clear, that there are multiple URLs to deal with (including a dedicated login URL of sorts). This call path, however, is invoked even if we all we have is a single URL for a single file download.


However, For some reason I only now saw your response to #5995 - thx! Will check out that other issue.

mih:
Please note that the change in d818cf1 reduces the behavior of `HTTPAuthAuthenticator` to the exact same as the one in its base class `HTTPRequestsAuthenticator`. This could have been made more obvious by actually calling the base method.

However, if the proposed patch is an actual solution to a problem, just changing the class would achieve the same thing.

yarikoptic:
> However, if the proposed patch is an actual solution to a problem, just changing the class would achieve the same thing.

exactly. Using `http_basic_auth` (not `http_auth`) should just do that AFAIK.

PR #6004, opened 2021-09-23
Ensure there's a blank line between the class `__doc__` and "Parameters" in `build_doc` docstrings

jwodder:
Without this change, applying `@build_doc` to a class whose docstring doesn't end in a newline, like so:

```python
@build_doc
class Foo(Interface):
     """Foo all the bars"""

    @staticmethod
    @eval_results
    def __call__():
        pass
```

will result in `__call__` getting a docstring that starts out with:

```
Foo all the bars
Parameters
----------
```

and the lack of a newline between the class docstring and "Parameters" will result in Sphinx producing a warning when trying to build the documentation for the class.  This PR fixes this behavior by ensuring that, if the class docstring doesn't already end with a newline, one is added, so that there are always at least two newlines (i.e., one blank line) before the "Parameters" header.

### DIFF
--- a/datalad/interface/base.py
+++ b/datalad/interface/base.py
@@ -349,6 +349,8 @@ def update_docstring_with_parameters(func, params, prefix=None, suffix=None,
     doc = prefix if prefix else u''
     if len(args) > 1:
         if len(doc):
+            if not doc.endswith('\n'):
+                doc += '\n'
             doc += '\n'
         doc += "Parameters\n----------\n"
         for i, arg in enumerate(args):
--- a/datalad/interface/tests/test_base.py
+++ b/datalad/interface/tests/test_base.py
@@ -216,6 +216,21 @@ def test_update_docstring_with_parameters_no_kwds():
     assert_in("3", fn.__doc__)
 
 
+def test_update_docstring_with_parameters_single_line_prefix():
+    from datalad.support.param import Parameter
+
+    def fn(pos0, pos1):
+        pass
+
+    update_docstring_with_parameters(
+        fn,
+        dict(pos0=Parameter(doc="pos0 param doc"),
+             pos1=Parameter(doc="pos1 param doc")),
+        prefix="This is a single line.",
+    )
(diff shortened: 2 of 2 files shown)


yarikoptic:
would you be so kind to rebase against `maint`?  I think we could even try to release since there were a few fixes (we are at 0.15.0-14-geb7a43b09) and we must not grow changelog for `auto` to succeed.

yarikoptic:
also if you could add a basic test to datalad/interface/tests/test_base.py which would demonstrate current issue (and test the fix) so we do not undo it -- would be great

yarikoptic (approved):

yarikoptic:
Let's wait for the tests to pass, and then we could proceed with merge.

yarikoptic:
Test seems to fail (didn't analyze it always)
```
FAIL: datalad.interface.tests.test_base.test_update_docstring_with_parameters_single_line_prefix
  File "/tmp/dl-miniconda-q0b2i3i0/lib/python3.9/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/tmp/dl-miniconda-q0b2i3i0/lib/python3.9/site-packages/datalad/interface/tests/test_base.py", line 230, in test_update_docstring_with_parameters_single_line_prefix
    assert_in("This is a single line.\n\nParameters\n", fn.__doc__)
AssertionError: 'This is a single line.\n\nParameters\n' not found in 'This is a single line.'
```

jwodder:
@yarikoptic The test should be fixed now.  It turns out `update_docstring_with_parameters()` doesn't add a "Parameters" block if the function has only one argument, presumably based on the assumption that the argument will always be `self`.  Might want to change that.

yarikoptic:
crawler fail is likely unrelated. one run of travis -- inability to download conda, unlikely to bring surprises, but I restarted the job -- better be safe than sorry ;)  if would turn green -- I think we should proceed with merge. Thank you @jwodder

yarikoptic:
Travis is happy now

PR #6006, opened 2021-09-24
RF: Promote internal status() helper with docs, and breadth-first reporting

mih:
Matching the diff_dataset() helper.

I had this change lying around locally. it makes using status-functionality internally more convenient.

### DIFF
--- a/datalad/core/local/status.py
+++ b/datalad/core/local/status.py
@@ -104,8 +104,47 @@ STATE_COLOR_MAP = {
 }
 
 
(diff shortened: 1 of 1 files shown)


mih:
I tried to convince the one mac runner on appveyor to actually do something but failed 5 times. The nature of the change makes it highly unlikely to fail only in this one particular run. Hence I'll ignore it.

mih:
OK, more docs should be more good

PR #6008, opened 2021-09-24
Large code-reorganization of everything runner-related

mih:
Basically, this re-organizes the code to establish `datalad.runner` as a self-contained subpackage that include all runner-related functionality, and no other functionality. Legacy imports are put in place to minimize likelihood of API breakage.

- [x] clear separation of runner from non-runner related code (ie. BatchedCommand is a different animal)
- [x] clearly separated tests
- [x] clear separation of runner interface, from protocol interface, from their various implementations, and underlying helpers
- [x] clear indication of necessary helpers from the rest of datalad
- [x] place `CommandError` with the runner code, as they are tightly linked
- [x] adjust appveyor/travis setup to consider the new test location

This is proposed against `maint` to avoid carrying a large diff for the full 0.16 release cycle. it should cause no breakage.


mih:
crawler failure is same old https://github.com/datalad/datalad-crawler/issues/104

bpoldrack (commented):
Don't think I can meaningfully double check the whole diff, but I like the aim/approach.

christian-monch (approved):
Looks good IMO. Nice cleanup of dependencies and better separation of sub-systems

mih:
Thanks for the feedback, will merge now.

PR #6013, opened 2021-09-27
Harmonize usage of --dry-run vs --dryrun in create-sibling-gitlab

adswa:
This adds a ``--dry-run`` parameter to ``create-sibling-gitlab`` to match the corresponding parameter in ``create-sibling-{github,gin,gogs,gitea}``, and deprecates the current ``--dryrun`` option.
The PR also includes a small doc fix to the URL pointing to the documentation of python gitlab

### DIFF
--- a/datalad/distributed/create_sibling_gitlab.py
+++ b/datalad/distributed/create_sibling_gitlab.py
@@ -142,7 +142,7 @@ class CreateSiblingGitlab(Interface):
             doc="""name of the GitLab site to create a sibling at. Must match an
             existing python-gitlab configuration section with location and
             authentication settings (see
-            https://python-gitlab.readthedocs.io/en/stable/cli.html#configuration).
+            https://python-gitlab.readthedocs.io/en/stable/cli-usage.html#configuration).
             By default the dataset configuration is consulted.
             """,
             constraints=EnsureNone() | EnsureStr()),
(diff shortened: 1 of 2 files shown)


yarikoptic:
+1 on harmonizing into `--dry-run`  (instead of `--dryrun`)

datalad/distributed/create_sibling_gitlab.py
```diff
@@ -190,14 +191,17 @@ class CreateSiblingGitlab(Interface):
...
         dryrun=Parameter(
             args=("--dryrun",),
             action="store_true",
-            doc="""If this flag is set, no communication with GitLab is
-            performed, and no repositories will be created. Instead
-            would-be repository names and configurations are reported for all
-            relevant datasets
-            """),
+            doc="""Deprecated. Use the renamed
+        [CMD: --dry-run CMD][PY: `dry_run` PY] parameter""")
```
mih, line 204: Is this a whitespace issue?

mih (changes requested):
Makes sense! Thx!

It would be good to discontinue the use of `dryrun` in the tests, though.

adswa:
> It would be good to discontinue the use of `dryrun` in the tests, though.

Right, thanks much for catching this. I have replaced them, and force-pushed.

datalad/distributed/create_sibling_gitlab.py
adswa: Thanks for catching this - a left over from copy-pasting. Should be resolved now.

adswa:
The Appveyor failure is a network glitch, I think

> Error making request with Error Code NotFound and Http Status Code NotFound. No further error information was returned by the service.

mih (approved):
Thanks for the update!

Issue #6014, opened 2021-09-27
Create-sibling-gitlab seems to overwrite existing remotes 


adswa:
I created two different siblings of a dataset without supplying a sibling name. The second sibling creation overwrote the first sibling:

```
$ datalad create-sibling-gitlab  -d . --site jugit --project datalad --layout flat                            1 !
create_sibling_gitlab(ok): . (dataset)
configure-sibling(ok): . (sibling)
action summary:
  configure-sibling (ok: 1)
  create_sibling_gitlab (ok: 1)
sys:1: ResourceWarning: unclosed <ssl.SSLSocket fd=3, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=6, laddr=('192.168.178.23', 56870), raddr=('134.94.130.31', 443)>
ResourceWarning: Enable tracemalloc to get the object allocation traceback
```

```
$ datalad siblings
.: here(?) [git]
[...]
.: jugit(-) [https://jugit.fz-juelich.de/adswa/datalad.git (git)]
```

Now I created a second sibling in the same dataset:

```
$ datalad create-sibling-gitlab  -d . --site jugit --project datalad2 --layout flat --access ssh 
create_sibling_gitlab(ok): . (dataset)
configure-sibling(ok): . (sibling)
action summary:
  configure-sibling (ok: 1)
  create_sibling_gitlab (ok: 1)
sys:1: ResourceWarning: unclosed <ssl.SSLSocket fd=3, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=6, laddr=('134.94.63.192', 40794), raddr=('134.94.130.31', 443)>
ResourceWarning: Enable tracemalloc to get the object allocation traceback
```
...and my previous sibling is gone, replaced by the new one

```
datalad siblings
.: here(?) [git]
[...]
.: jugit(-) [git@jugit.fz-juelich.de:adswa/datalad2.git (git)]

```

PR #6015, opened 2021-09-27
BF: Do not overwrite existing sibling configurations with create-sibling-gitlab

adswa:
This tries to be a fix for #6014. It adds a safe-guard to check whether a sibling-name to-be already exists, and errors out instead of overwriting it. I believe that this should be non-controversial, as it mirrors the way the other ``create-sibling-``s work. However, the comments seem to indicate that there is something special going on with GitLab, and the tests required some larger changes. I would therefore be very grateful if @mih (who has written code I have now changed) could check my changes carefully.


mih (approved):
LGTM!

Even if it changes the behavior, I think this is OK, because the behavior was the problem.

Thx!

PR #6020, opened 2021-09-28
BF: ProgressHandler - close the other handler if was specified

yarikoptic:
First commit actually enhances the test to optionally (if env var set) ensure that and reporting about open files for `DATALAD_ASSERT_NO_OPEN_FILES=assert` runs. 2nd commit then fixes the underlying issue of ProgressHandler not closing the "other" handler upon close.


yarikoptic:
restarted benchmarks. IMHO this change should have not effected them!

mih (approved):
LGTM, thx!

Issue #5975, opened 2021-09-11
Discontinue use of `exc_str()` throughout the code base


mih:
ATM 125 instances remain, after #5954 made it obsolete.  The new approach is described in http://docs.datalad.org/en/latest/design/exception_handling.html

mih:
45 files to go until completion. This is trivial work.

PR #6022, opened 2021-09-28
Next wave of exc_str() removals

mih:
Ping #5975


datalad/__init__.py
```diff
@@ -243,8 +244,8 @@ def teardown_package():
         print("Obscure filename: str=%s repr=%r"
                 % (OBSCURE_FILENAME.encode('utf-8'), OBSCURE_FILENAME))
     except UnicodeEncodeError as exc:
-        from .dochelpers import exc_str
-        print("Obscure filename failed to print: %s" % exc_str(exc))
+        ce = CapturedException(exc)
+        print("Obscure filename failed to print: %s" % ce)
```
yarikoptic, line 248: sorry if I forgot, why 
1. not 
```suggestion
        print("Obscure filename failed to print: %s" % CapturedException(exc))
```
2. and not just overload `exc_str = CapturedException` and not bother with any code changes whenever it is used? (AFAIK it is always used/expected to provide a str)

PR #6027, opened 2021-10-02
BF: disable outdated SSL root certificate breaking chain on older/buggy clients

### LINKED ISSUE #6026: Lots of `server certificate verification failed. CAfile: none CRLfile` errors on ubuntu CI
Sample https://ci.appveyor.com/project/mih/datalad/build/job/begrlqetcsw4f3ra

Seems unspecific to this PR.

Reported for

- https://datasets-tests.datalad.org
- https://gin.g-node.org

and possibly CRCNS.org, but hidden in a more complex download failure.

Given that it is only the linux workers, and that it is not a single host for which it fails, I blame the certificate setup in the worker.

yarikoptic:
Closes #6026


yarikoptic:
since all green CIs, I would assume that it is indeed fixing a possible issue, so will proceed with merge, we can always revert ;)

yarikoptic:
FTR: merged maint into master and pushed (there were only two PRs merged, both BFs)

PR #6031, opened 2021-10-04
BF: undeprecate readline_rstripped + adjust mock to mock actual warnings library

### LINKED ISSUE #6030: `readline_rstripped()` should be un-deprecated
In [this exchange](https://github.com/datalad/datalad/pull/5667#discussion_r635214578) on #5667, it was decided to add a deprecation warning to `datalad.cmd.readline_rstripped()` on the assumption that it was no longer used.  However, it *is* used: [it's the default `output_proc` for `BatchedCommand`](https://github.com/datalad/datalad/blob/91ed701ddc36a97d76c2f976f996157959f73606/datalad/cmd.py#L96), and so, with the release of 0.15.0, using `BatchedCommand`, `BatchedAnnex`, `BatchedAnnexes`, etc. produces a deprecation warning.

yarikoptic:
I was wrong in my statement that this function is no longer used, it is used in the Batched*
classes.  It is a trivial function and could be moved I guess to not appear in the interface
but since we do not really sanitize those, I decided that undeprecating would be the easiest
way forward.  Closes #6030.

Upon removal of "warnings" import test_done_deprecation started to fail since it patches
the local import of warnings. IMHO it should mock patch the actual warnings library, so I have
fixed that test accordingly

### DIFF
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -16,7 +16,6 @@ import os
 import subprocess
 import sys
 import tempfile
-import warnings
 
 # start of legacy import block
 # to avoid breakage of code written before datalad.runner
@@ -62,9 +61,6 @@ _MAGICAL_OUTPUT_MARKER = "_runneroutput_"
 
 
 def readline_rstripped(stdout):
-    warnings.warn("the function `readline_rstripped()` is deprecated "
-                  "and will be removed in a future release",
-                  DeprecationWarning)
     return stdout.readline().rstrip()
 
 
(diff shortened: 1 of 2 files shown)


mih:
I understand the motivation, but I think it is the wrong direction. It is true that module API overall is a mess, but there is a clear trajectory to more consistent use of "private" indicators in names. I think the better way would be to keep the deprecation, and adjust the previous usage to a `_readline_rstripped()` which is also called by the deprecated method.

mih:
I proposed https://github.com/datalad/datalad/pull/6032 as an alternative fix.

yarikoptic:
I thought about such approach too, let's go with it, thank you!

Issue #5044, opened 2020-10-15
misleading suggestion in the error message - datalad.runtime.use-patool is in fact ignored on windows


yarikoptic:
```python
external_versions.check(
    "cmd:7z",
    msg='The 7z binary (7-Zip) is required for archive handling, but is missing. '
        "Setting the config flag 'datalad.runtime.use-patool' enables an "
        "alternative implementation that may not need 7z.")
```
but that code is reached via
```python
# fall back on patool, if a functional implementation is available
# (i.e. not on windows), it is requested, or 7z is not found
if not on_windows and (
        cfg.obtain(
            'datalad.runtime.use-patool', default=False,
            valtype=anything2bool) or not external_versions['cmd:7z']):
    from datalad.support.archive_utils_patool import (
        decompress_file as _decompress_file,
        # other code expects this to be here
        compress_files
    )
else:
    from datalad.support.archive_utils_7z import (
        decompress_file as _decompress_file,
        # other code expects this to be here
        compress_files
    )
```
so `on_windows` you can break your knuckles trying to set that config variable and it would have no effect.  IMHO if that variable is explicitly set, it should be warranted.  If not set -- then on_windows or not should be the decision maker

mih:
So we came to the conclusion that patool is not working on windows. However, upstream thinks otherwise. Given this fact, I think it is sensible to go with @yarikoptic's suggestion and check whether the setting is enabled and honor this. I think it would be sufficient to remove `not on_windows and` from the conditional.

PR #6033, opened 2021-10-05
Honor datalad.runtime.use-patool if set regardless of OS (was Windows only)

mih:
Previously, windows was a declared special case in which the flag was
effectively ignored. This is confusing, and may be unjustified even.

Fixes datalad/datalad#5044

### DIFF
--- a/datalad/support/archives.py
+++ b/datalad/support/archives.py
@@ -47,12 +47,10 @@ from datalad.utils import (
 from datalad import cfg
 from datalad.config import anything2bool
 
-# fall back on patool, if a functional implementation is available
-# (i.e. not on windows), it is requested, or 7z is not found
-if not on_windows and (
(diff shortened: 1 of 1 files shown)


yarikoptic:
and let's push a release with this simple one, we are already at 0.15.1-33-gba32ae216

yarikoptic:
hm, I do not see github actions for release to ever trigger after merging this PR: https://github.com/datalad/datalad/actions/workflows/release.yml  so it was not released.  @jwodder -- any clue on what have might happened?
meanwhile I guess I will remove release label from this PR and we could apply it to some other PR to see if release action would finally kick in

jwodder:
@yarikoptic GitHub Actions was having some problems yesterday; I experienced a run for datalad/git-annex that waited forever and never started.

PR #6037, opened 2021-10-06
Align installation setup with Travis

mih:
In response to my observation that the set of tests executed by travis and appveyor is not the same.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -266,8 +266,10 @@ install:
 
 
 build_script:
-  - python -m pip install ".[tests]"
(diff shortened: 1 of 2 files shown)


mih:
Needs more:

```
libxmp.ExempiLoadError: Exempi library not found.
```

Although only required for two metadata-related tests on linux. Maybe these tests should rather be skipped when it is not around.

mih:
I decided to not install libxmp on appveyor. These tests will be phased out soonish, and this is absolutely marginal functionality, not worth impacting the runtime of each CI run.

Issue #5982, opened 2021-09-15
Creating a ria sibling only if store exists


mih:
From https://neurostars.org/t/creating-a-ria-sibling-only-if-store-exists/20181

> Is there a way of creating a sibling of a given dataset in a ria store but only if that store already exists? I use datalad create-sibling-ria to create both the store and a sibling in the store, but I want a command that attempts to create a sibling in a store and fails if the store doesn’t exist. Is there a way of achieving this?

There is no builtin to do this conveniently that I am aware of.

PR #6045, opened 2021-10-06
UX: Allow create-sibling-ria to only act on existing store

adswa:
This implements a basic solution to #5982.
The issue is a feature request for the ability to only act on
create-sibling-ria calls if the specified store already
exists. This change adds a parameter to create-sibling-ria,
``'--construct-store'`` (False by default), which, if set, creates a new
store if it doesn't exist under the specified url, and otherwise fails to
create a RIA sibling, warning the user that there is no store at the
specified path. This is an API change to the current release, which
will create a new store if it doesn't find one.

The parameter is intentially not named  ``create-store`` because this would
lead to a name conflict with the ``create_store`` method of ``ria_utils``.

### DIFF
--- a/datalad/distributed/create_sibling_ria.py
+++ b/datalad/distributed/create_sibling_ria.py
@@ -41,6 +41,8 @@ from datalad.distribution.dataset import (
 )
 from datalad.distributed.ora_remote import (
     LocalIO,
+    RIARemoteError,
+    RemoteCommandFailedError,
     SSHRemoteIO,
 )
 from datalad.utils import (
@@ -407,7 +409,7 @@ class CreateSiblingRia(Interface):
             # Because this raises a FileNotFound error if non-existent, we need
             # to catch it
             io.read_file(Path(base_path) / 'ria-layout-version')
-        except FileNotFoundError as e:
+        except (FileNotFoundError, RIARemoteError, RemoteCommandFailedError) as e:
             if not new_store_ok:
                 # we're instructed to only act in case of an existing RIA store
                 res = get_status_dict(


datalad/distributed/create_sibling_ria.py
```diff
@@ -394,7 +400,18 @@ def __call__(url,
         #       - more generally consider store creation a dedicated command or
         #         option
 
-        create_store(SSHRemoteIO(ssh_host) if ssh_host else LocalIO(),
+        io = SSHRemoteIO(ssh_host) if ssh_host else LocalIO()
+        if not io.exists(Path(base_path)) and expect_store:
```
bpoldrack, line 404: For consistency with how the ORA remote asses the existence of a store, a valid `ria-layout-version` file at that path is needed, too. Mere existence of the path itself doesn't make it a useable store. Also: `base_path` could be a file and would pass this test.

So, I think change condition to "`io.read_file(Path(base_path)  / 'ria-layout-version')` doesn't raise anything".

bpoldrack (changes requested):
Unless there are strong objections to my suggestion I would prefer it to be the other way around. Only the first dataset per store would need to actually create it. Hence, in most cases creation isn't needed. I would go with a `--create-store` flag rather than `expect-store`.
For `maint` its default should probably be `true`, but switch to `false` in `master`. (or not have it in `maint` at all)

adswa:
Thanks for the review, @bpoldrack!

> Only the first dataset per store would need to actually create it. Hence, in most cases creation isn't needed. I would go with a --create-store flag rather than expect-store.

Sure, also makes sense. In your opinion, what should happen in case of ``--create-store`` at a path of an existing store? Shall we error out, shall we ignore the flag and continue placing a sibling in it, or a "mix" between completely ignoring and erroring with some form of logging? 


> For maint its default should probably be true, but switch to false in master. (or not have it in maint at all)

I'd personally prefer to not have it in maint at all

mih:
I also see this not targeting `maint` at all.

adswa:
ok, noted. Just let me know how you would like me to change it.

PR #6053, opened 2021-10-07
Revert "TST: Disable tests against metalad"

yarikoptic:
This reverts commit 96014ecc9ccfca3ddddba2e04c2a9d1520f5ae1d.

metalad is under active development but starts to be tested/used in
use cases, so would be useful to have datalad cocompatible with it
(might need a release).  Relevant issues

https://github.com/datalad/datalad-metalad/issues/124
https://github.com/datalad/datalad-metalad/issues/66

### DIFF
--- a/.github/workflows/test_extensions.yml
+++ b/.github/workflows/test_extensions.yml
@@ -13,6 +13,7 @@ jobs:
         extension: [
             datalad-neuroimaging,
             datalad-container,
+            datalad-metalad,
             datalad-crawler,
             datalad-deprecated,
         ]


mih:
I disabled it, because it is failing and will fail for some time. I very much enjoy the theoretical chance that a PR can actually become "green". If this is merged, we will not learn much more than what we already do now, but burn lots of coal for it.

I can promise to turn it on again, as soon as there is a chance that it would be helpful.

mih:
> isn't it just a matter of tagging/uploading some release since its tests are currently passing https://ci.appveyor.com/project/mih/datalad-metalad ?

That may be true right now. I wanted to point to the intent to be non-broken. Basically thinking ahead to next week when something is broken, but nobody cares.

However, I will leave this decision to @christian-monch

yarikoptic:
Replaced by #6060

Issue #6036, opened 2021-10-05
ORA remote fails (inconsequentually) spectacularly on SSH connection failure


mih:
```
% datalad get -n MNINonLinear   
[INFO   ] scanning for unlocked files (this may take some time)                                                            
The authenticity of host 'hcp_a.ds.inm7.de (10.0.8.13)' can't be established.                                              
ED25519 key fingerprint is SHA256:lgHgWH1yXQmDidLyJ43ngKD1nIMbv5xKSItftk0dgVg.
Are you sure you want to continue connecting (yes/no)? no
[INFO   ] Host key verification failed. 
The authenticity of host 'hcp_a.ds.inm7.de (10.0.8.13)' can't be established.
ED25519 key fingerprint is SHA256:lgHgWH1yXQmDidLyJ43ngKD1nIMbv5xKSItftk0dgVg.
Are you sure you want to continue connecting (yes/no)? no
[INFO   ] Host key verification failed. 
[INFO   ] Traceback (most recent call last):
|   File "/usr/lib/python3/dist-packages/datalad/distributed/ora_remote.py", line 694, in _wrap_handle_errors 
[INFO   ] return func(self, *args, **kwargs)
|   File "/usr/lib/python3/dist-packages/datalad/distributed/ora_remote.py", line 1064, in initremote 
[INFO   ] self.get_store() 
[INFO   ] File "/usr/lib/python3/dist-packages/datalad/distributed/ora_remote.py", line 1021, in get_store 
[INFO   ] self.verify_store() 
[INFO   ] File "/usr/lib/python3/dist-packages/datalad/distributed/ora_remote.py", line 800, in verify_store 
[INFO   ] self._get_version_config(dataset_tree_version_file) 
[INFO   ] File "/usr/lib/python3/dist-packages/datalad/distributed/ora_remote.py", line 983, in _get_version_config 
[INFO   ] file_content = self.io.read_file(path).strip().split('|') 
[INFO   ] File "/usr/lib/python3/dist-packages/datalad/distributed/ora_remote.py", line 1134, in io 
[INFO   ] self._io = SSHRemoteIO(self.storage_host, self.buffer_size) 
[INFO   ] File "/usr/lib/python3/dist-packages/datalad/distributed/ora_remote.py", line 314, in __init__ 
[INFO   ] self.ssh.open() 
[INFO   ] File "/usr/lib/python3/dist-packages/fasteners/lock.py", line 302, in wrapper 
[...]

PR #6063, opened 2021-10-08
BF: ORA shouldn't crash while handling a failure

bpoldrack:
Closes datalad/datalad#6036


mih:
Restarted the tests

mih:
I filed https://github.com/datalad/datalad/issues/6069 to keep track of the logging problem. I will merge this now, with the reasoning that now logging may be better than crash -- although I am not totally convinced that this is actually true.

Issue #5617, opened 2021-04-29
A successful `datalad get` can still look quite sad


mih:
Here is the raw `git-annex get` call:

```
% git -C derivative/aggregate_fmri_timeseries annex get .
get sub-10/atlases/bold3Tp2/shen_fconn_atlas_150.nii.gz (from psydata...) 

  download failed: Not Found

  failed to download content
(from inm7-storage...) 
[INFO] Downloading 'https://datapub.fz-juelich.de/studyforrest/studyforrest.ria/7fc/d8812-d0fe-11e7-8db2-a0369f7c647e/annex/objects/2G/0J/MD5E-s32338--3cb12b68abd59160dffdf4830e1de8bb.nii.gz/MD5E-s32338--3cb12b68abd59160dffdf4830e1de8bb.nii.gz' into '.git/annex/tmp/MD5E-s32338--3cb12b68abd59160dffdf4830e1de8bb.nii.gz' 
(checksum...) ok
(recording state in git...)
```

So it tried one thing, failed, and then succeeded. Here is datalad:

```
% datalad get derivative/aggregate_fmri_timeseries/sub-10/atlases/bold3Tp2/shen_fconn_atlas_150.nii.gz
get(ok): derivative/aggregate_fmri_timeseries/sub-10/atlases/bold3Tp2/shen_fconn_atlas_150.nii.gz (file) [download failed: Not Found
failed to download content
download failed: Not Found
failed to download content
download failed: Not Found
failed to download content]
action summary:
  get (notneeded: 1, ok: 1)
```

So all worked, but the number of "failed" statements has doubled, and they shadow the happy core message quite effectively.

Due to the particular circumstances in my original run I  did not notice that each failing download attempt is also accompanied by ERROR log messages. So it is even worse.

mih:
Oh, and depending where the error is happening and how exactly it looks like, this can be REALLY ugly (thx @adswa):

```
adina@head1:~$ datalad clone https://github.com/psychoinformatics-de/studyforrest-data-retinotopy.git
[INFO   ] Remote origin not usable by git-annex; setting annex-ignore           
[INFO   ] ssh: Could not resolve hostname bulk1.htc.inm7.de: Name or service not known 
[INFO   ] ssh: Could not resolve hostname bulk1.htc.inm7.de: Name or service not known 
[INFO   ] Traceback (most recent call last):
[...]

Issue #3939, opened 2019-12-17
Distinguish and support multiple messages per result


mih:
Came up in https://github.com/datalad/datalad/pull/3931

In some cases (git-annex) provides multiple pieces of information with different severity levels (e.g. error messages and minor notes). ATM we have no ability to mark them up and ensure that result renderers reflect the severity differences.

mih:
Thank you for reporting this issue!

However, at this point I am closing it at a time of ~600 open issues, in an attempt to regain control over the issue tracker. This particular issue was posted more than 6 months ago, has not received a single response, and does not describe an immediate software defect. *Closing this issue does not imply that it is not relevant*. If any reader objects the closing of this issue, please feel more than free to reopen it with an update on how it is relevant for the current state of the project, and which concrete actions to address it are ongoing or planned.

PR #6070, opened 2021-10-12
BF: Retain git-annex error messages & don't show them if operation successful

DisasterMo:
### Description

Instead of being mashed into a 'normal' message, git-annex reported error messages will now be retained in a separate dict entry.
This allows for more differentiated handling of them. Examples implemented in this PR:

1. To avoid confusion, git-annex error messages will only be rendered if they are meaningful, i.e. when the operation actually failed.
2. Error messages are now grouped separately from notes & coloured red.

### Related Issues

1. #5617 
2. #3939

### DIFF
--- a/datalad/core/local/tests/test_results.py
+++ b/datalad/core/local/tests/test_results.py
@@ -35,8 +35,8 @@ def test_default_result_renderer():
         (dict(type='funky'),
          ['<action-unspecified>(<status-unspecified>): (funky)']),
         # plain message makes it through
-        (dict(message='funky'),
-         ['<action-unspecified>(<status-unspecified>): [funky]']),
+        (dict(message='funky', error_message='extra-funky'),
+         ['<action-unspecified>(<status-unspecified>): [funky] [extra-funky]']),
     ]
     if on_windows:
         testcases.extend([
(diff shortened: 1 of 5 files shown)


DisasterMo:
> Some tests expect the error messages to be rendered/provided specifically as they were. Those need to be adjusted.
On it. I should probably run the tests myself before I submit a PR in the future, huh.

> The PR also includes a number of unrelated formatting changes.
Yeah, I had PyCharm do it's magic a couple of times out of habit, sorry.

yarikoptic:
Windows?

DisasterMo:
> Also I would personally prefer no preemptive linebreaks in the reporting (i.e. before the [)

Why? From what I can tell, it is pretty obvious that any message is supposed to be a multi-line block. Personally, I find it incredibly ugly to see a multi-line block not start on a new line.

DisasterMo:
> what are the semantics here? Do we know that this is one message with several lines, or several messages.

The original error-messages contains, in general, multiple messages each with at least one line. We can distinguish lines, but not messages. So the general form of `d['error-messages']` looks something like this:
`["err1:line1", "err1:line2", "err2:line1", ... , "errN:lineM"]`

PR #6073, opened 2021-10-13
Remove uses of `__full_version__` and `datalad.version`

jwodder:
The import of `__full_version__` in `datalad.local.wtf` was causing a deprecation warning to be emitted from just importing `datalad.api`.  This PR fixes that and also removes all the remaining imports of `datalad.version` I could find.


mih:
OK, no need to keep this dangling here. Thx again!

PR #6075, opened 2021-10-14
BF ensure lists in AnnexRepo and GitRepo

christian-monch:
This PR ensures that paths that are converted into lists if list parameters are expected.
This prevents some status()-functions from yielding information about a complete repo when information about just a single file is requested.

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -3130,7 +3130,7 @@ class AnnexRepo(GitRepo, RepoInterface):
         """
         Parameters
         ----------
-        paths : list or None
+        paths : str, list of str, or None
           Specific paths to query info for. In `None`, info is reported for all
           content.
         init : 'git' or dict-like or None
@@ -3171,6 +3171,8 @@ class AnnexRepo(GitRepo, RepoInterface):
             pathlib.Path of the content object in the local annex, if one
             is available (with `eval_availability`)
         """
+        paths = ensure_list(paths)
+
         if init is None:
             info = OrderedDict()
         elif init == 'git':
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -3087,7 +3087,7 @@ class GitRepo(CoreGitRepo):
 
         Parameters
         ----------
-        paths : list or None
+        paths : str, list of str, or None
           If given, limits the query to the specified paths. To query all
           paths specify `None`, not an empty list. If a query path points
           into a subdataset, a report is made on the subdataset record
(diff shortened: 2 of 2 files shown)


mih:
> Well, this isn't translating anything. From my POV it's just consistency. We do that particular thing pretty much everywhere else. 

There are two conceptual places where this is done

1. the high-level command interface, where it is a technical necessity to by robust regarding a variety of inputs. Here, typically an extensive (although often still insufficient) input validation is done
2. residual components of the earliest `*Repo` interface

Just because we continuously fail to remove old functionality like `normalize_path`, it should not be used as an argument to continue with the throw-me-anything-and-I-will-do-my-best approach, IMHO.

> I don't see why developers should need to keep track of all the functions that handle lists of paths and remember which are the exceptions that can't deal with a single one not passed as a list.

I think it is a developer's task to call functions according to specification, and to suffer if not ;-)

> I would prefer assertations about input parameters, although there seems to be some raise-avoiding scent on the code (less than 100 asserts in non-test datalad-code). I think that asserts are conforming to this "in-official rule", if the parameter content is not directly determined by user-input. Or am I misinterpreting something?

I am not confident that I fully understand what you said. But if you said that you would prefer to see more `assert` statements, I agree. 

I am not aware of inofficial rules (who could be sure that they exist?). I doubt that inofficial rules contribute something valuable. If they do, they should be made explicit, and discussed. I am very much enjoying the growing collection of design documents that help me understand how something is *meant* to work, to be able to tell when it is broken.

One thing to keep in mind re input validation in low-level code like `*Repo` is performance. A type check should be OK. An in-depth *upfront* deep-inspection of complex inputs will like represent a performance impact. Over the past years the development trend is a shift of flexibility from low to high-level code. Simultaneously, a shift of error *interpretation* from low to high-level code. This is a complex issue that I was not yet able to put into writing. So from my POV we are doomed to keep having these discussions, until a clear, simple, verbalizable pattern emerges. But the direction is exactly that: more clear, simpler, more explainable. Everytime the entropy in low-level code goes up, there is a problem.

mih:
Preaching aside: the change proposed here actually breaks the behavior of the method for all callers, because:

```
>>> ensure_list(None)
[]
```

which means that `None` input not longer causes a report on *all* paths, but a report on *no* paths -- see the test failures for the impact.

christian-monch:
> Preaching aside: the change proposed here actually breaks the behavior of the method for all callers, because:
> 
> ```
> >>> ensure_list(None)
> []
> ```
> 
> which means that `None` input not longer causes a report on _all_ paths, but a report on _no_ paths -- see the test failures for the impact.

Definitely an error.

Shall we close this PR unmerged @bpoldrack ? Possibly opening a new PR with an approriate assert?

christian-monch:
Not a valid fix, not an agreed upon strategy. Created issue #6087 to continue the discussion and possibly create an alternative PR

PR #6078, opened 2021-10-14
BF: Add dashes to 'force' option in non-empty directory error message

DisasterMo:
### Description

In my opinion all mentioned options should be displayed 'in full', i.e. with any and all dashes.

### DIFF
--- a/datalad/core/local/create.py
+++ b/datalad/core/local/create.py
@@ -335,7 +335,7 @@ class Create(Interface):
                 'status': 'error',
                 'message':
                     'will not create a dataset in a non-empty directory, use '
-                    '`force` option to ignore'})
+                    '`--force` option to ignore'})
             yield res
             return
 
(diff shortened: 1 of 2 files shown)


mih:
Ah! Sorry! I misread the location where this is happening. The macro-handling I mentioned only runs for API doc generation. See https://github.com/datalad/datalad/blob/dfa956984f6e737e11518f4a9b230b6a6eadc97a/datalad/interface/base.py#L199 and https://github.com/datalad/datalad/blob/dfa956984f6e737e11518f4a9b230b6a6eadc97a/datalad/interface/base.py#L249.

I think it would be lovely to have the same feature for result/error messages -- which would require pulling this handling out of these functions into a more targeted helper.

However, right now I am not sure how we could tell whether a result renderer is called in the context of a cmdline invocation vs a Python session.

Issue #828, opened 2016-09-14
Related effort(s): bdbag, bagit, etc


yarikoptic:
http://bd2k.ini.usc.edu/tools/bdbag/
pointing to 
- https://github.com/ResearchObject/bagit-ro describing data structure to describe collections of data, in large minicing
  -  git (tags)
  -  annex (file checksums) entities
  - datalad (json-ld meta-data description, serialization for delivery through tarballs, access to portals with authentication)
    Apparently BagIt format is (they say) widely adopted already for digital preservation etc. Wiki: https://en.wikipedia.org/wiki/BagIt 

possible integration is digestion or export into 'bagit's.  Didn't look into mapping their meta-data descriptions at all yet, but we might want to keep in mind for our 'meta data aggregators' to provide mapping so it could be bi-directional.
- https://github.com/ini-bdds/bdbag (and larger collection of related tools under that organization) -- set of Python (2+3) utilities to interface with BDDS (???) structured BagIts
  - defines "exchange" profiles json description to harmonize exchange of such bagits between consumer/producer software

above projects also mention `ark:/` urls,  https://en.wikipedia.org/wiki/Archival_Resource_Key, and use http://n2t.net (by california digital library) for resolving those ... (yoh) wasn't aware of ARK

mih:
Cool. We should be able to import a BagIt and turn in into a dataset....

yarikoptic:
oh well, we know about them for a while apparently, but it is not an issue ;)

mih:
Started looking into this.

The Python implementation of `bagit` itself is not very useful for us. It only seems to work with all data being present locally, and does not seem to handle `fetch.txt` which is a key feature for us.

https://github.com/fair-research/bdbag seems to be a better base.

mih:
I think we can close this. Pointers to an implementation are in the posts above.

Issue #6086, opened 2021-10-15
ORA remote `whereis` reports wrong HTTP URL


mih:
Such a URL is reported

https://datapub.fz-juelich.de/studyforrest/studyforrest.ria/annex/objects/Kz/7V/MD5E-s37132--f8a50b1462dbb3eb403013d28c197a91.nii.gz/MD5E-s37132--f8a50b1462dbb3eb403013d28c197a91.nii.gz

but the actual URL is 

https://datapub.fz-juelich.de/studyforrest/studyforrest.ria/7fc/d8812-d0fe-11e7-8db2-a0369f7c647e/annex/objects/Kz/7V/MD5E-s37132--f8a50b1462dbb3eb403013d28c197a91.nii.gz/MD5E-s37132--f8a50b1462dbb3eb403013d28c197a91.nii.gz

(dataset ID part inserted)

edit by @yarikoptic : fixed the "actual" by adding `/` before `annex`

mih:
This seems to be a reincarnation of #4979.

mih:
Discussed in a call. There was basic agreement that broken URLs should not be reported.

@bpoldrack suggested to included a hardcoded `dsid[:3]/dsid[3:]` in the output, and veto'ed the use of `dsobj_dir`. Likewise, he objected the use of `key_path`.

@mih still doesn't understand why this would be a sensible move. Hardcoding `dsid[:3]/dsid[3:]` would perform layout version independent inference, while `dsobj_dir` is based on the actual store configuration. Using `key_path` also is based on actual store configuration, whereas hardcoding a directory based on `dirhash` is not. It is unclear why `ssh-based` access can work out of the box, while http-based access should required particular server-side URL mapping, beyond mere exposure of a store, in order to make HTTP. To make my confusion complete, the `transfer_retrieve()` implementation does `dsobj_dir / key_path` -- which is conceptually exactly what the veto'ed change does.

NB: This path handling is POSIX-only, and would yield incorrect HTTP URLs on non-Posix systems.

Because these aspects remain unresolved, a maintenance fix is to disable the reporting of incorrect URLs, by not reporting URLs. A correct fix can then be introduced later.

bpoldrack:
@mih 
> @bpoldrack suggested to included a hardcoded dsid[:3]/dsid[3:] in the output, and veto'ed the use of dsobj_dir. Likewise, he objected the use of key_path.

Not exactly. It's about `key_path` only. 


[...]

PR #6091, opened 2021-10-18
Report correct HTTP URL for RIA store content

mih:
These URL do not only have annecdotal value. In efforts like #828 incorrect URLs invalidate the result.

The fix is not pretty (due to #6089 and similar issues), but it comes with a test that hopefully indicates breakage when these are fixed.


mih:
I am reopening this PR. Issue #6086 continues to be a problem, but no better solution materialized in the past month. If this fix is broken, it is certainly less broken than the completely disfunctional "URL" that is presently reported.

mih:
I will merge as soon as the tests pass.

Issue #6084, opened 2021-10-15
Status/Diff performance boost (for large-N datasets)


mih:
I am hunting for performance. With #6034 in mind, the default mode of `status(report_filetype='eval')` (leading to `eval_file_type=True` in `GitRepo.diffstatus()` and friends is a HUGE slowdown in large dataset (many files). For a `status()` is will resolve ANY symlink (i.e. all annexed files) on the file system (on call each), and for a `diff()` it will use a batched `git cat-file`.

This is doing the correct thing by default (well #6034), but has to be turned off for any performance critical functinality.

Demo setup

```
% datalad create fast
% cd fast
% echo ingit > ingit
% datalad save ingit --to-git
% echo inannex > inannex
% datalad save inannex
% ln -s ingit somelink
% datalad save
```

```
In [3]: ds.repo.diffstatus('HEAD', None, eval_file_type=False, paths=['inannex', 'ingit', 'somelink'])
Out[3]: 
OrderedDict([(PosixPath('/tmp/fast/inannex'),
              {'type': 'symlink',
               'gitshasum': '33cc8c5f11621d19c212ebf4ff8b45b3fc4169a9',
               'prev_gitshasum': '33cc8c5f11621d19c212ebf4ff8b45b3fc4169a9',
               'state': 'clean'}),
             (PosixPath('/tmp/fast/ingit'),
              {'type': 'file',
               'gitshasum': '4ede24e7074a994ff45e1bf0e04eb7fda254052f',
               'bytesize': 6,
               'prev_gitshasum': '4ede24e7074a994ff45e1bf0e04eb7fda254052f',
               'state': 'clean'}),
             (PosixPath('/tmp/fast/somelink'),
              {'type': 'symlink',
               'gitshasum': 'ec28779b36c1a65a3fb8ca1b1cad32c8b1f0fd45',
               'prev_gitshasum': 'ec28779b36c1a65a3fb8ca1b1cad32c8b1f0fd45',
               'state': 'clean'})])
In [4]: ds.repo.diffstatus('HEAD', None, eval_file_type=True, paths=['inannex', 'ingit', 'somelink'])
Out[4]: 
OrderedDict([(PosixPath('/tmp/fast/inannex'),
              {'type': 'file',
               'gitshasum': '33cc8c5f11621d19c212ebf4ff8b45b3fc4169a9',
               'bytesize': 108,
               'prev_gitshasum': '33cc8c5f11621d19c212ebf4ff8b45b3fc4169a9',
[...]

Issue #6034, opened 2021-10-05
Dataset content type/size reporting broken in subdatasets


mih:
Demo:

```sh
% datalad create demo
% cd demo
% datalad create -d . subds
% echo 123 > file1; echo 123 > subds/file1
% datalad save -r
% datalad -f json status -r
...
{"bytesize": 108, "gitshasum": "670d89de1eaf38bd9740fdc274ef214e0f63408f", "path": "/tmp/demo/file1", "state": "clean", "type": "file", ...}
{"gitshasum": "670d89de1eaf38bd9740fdc274ef214e0f63408f", "path": "/tmp/demo/subds/file1", "state": "clean", "type": "symlink", ...}
```

I stripped a bunch of properties, but notably `bytesize` is missing, and the `type` info differs for the two identical files.

mih:
I believe the reason for that is that it is type evaluation is forcefully disable...

https://github.com/datalad/datalad/blob/561b175b33de1171a5fc8e70597dbfd3188d090c/datalad/support/gitrepo.py#L3142

It has been like this since the introduction of the flag in https://github.com/datalad/datalad/commit/e9314f1917b45a9b6a15e9a028593d7f0f7f4524

Issue #5643, opened 2021-05-06
`diff` performance impaired by absence of `status -t raw` equivalent


mih:
I have a sizable dataset (single dataset, no present subdatasets, ~150k annexed files):

```
% datalad status --untracked no -t raw
nothing to save, working tree clean
datalad status --untracked no -t raw  8.80s user 3.12s system 61% cpu 19.256 total                                                                                

% datalad status --untracked no
nothing to save, working tree clean
datalad status --untracked no  1126.54s user 8.31s system 100% cpu 18:54.29 total
```

`diff` has no `-t raw`, and on top of that needs to run things twice, hence:

```
datalad -f '{path}::{state}' diff -f master -t jureca-master  2096.43s user 11.43s system 100% cpu 35:07.01 total                                                 
```

PR #6097, opened 2021-10-18
Boost default performance of status/diff commands

mih:
This is done by removing the file-type evaluation as a default operation, following the rational given in https://github.com/datalad/datalad/issues/6084.

Moreover, the type reporting now follows a simpler rule. When asked for a plain status, file-types are reported as known to git. Only in `--annex` reporting mode the type info is corrected for symlinks that are actually annexed files, and the links only exist for technical reasons in a particular clone.

This generally leads to more robust and valid behavior across supported environments, and to more performance, because this kind of type inference can be done without expensive file system
interaction to resolve symlinks of read-out file content.

Using this simpler switching logic (--annex yes/no), also avoids further expansion of the `diff` API to tune file-type evaluation for diffs.

Fixes datalad/datalad#6084
Fixes datalad/datalad#6034
Fixes datalad/datalad#5643

#### Some benchmarks

All benchmarks based on https://github.com/datalad-datasets/hcp_wm_preprocessed with ~45k files.

##### Current master as baseline

```
# old default
% time datalad status
datalad status  17.98s user 3.12s system 90% cpu 23.296 total

# old tuned
% time datalad status -t raw
datalad status -t raw  2.93s user 0.52s system 105% cpu 3.274 total

# old annex-mode
% time datalad status --annex
datalad status --annex  108.49s user 12.88s system 185% cpu 1:05.38 total

# old annex-tuned
% time datalad status --annex -t raw
datalad status --annex -t raw  96.52s user 11.58s system 206% cpu 52.395 total
```

##### This PR

```
# default
% time datalad status
datalad status  3.08s user 0.44s system 105% cpu 3.338 total

# annex-mode
% time datalad status --annex
datalad status --annex  93.26s user 11.37s system 210% cpu 49.793 total
```

Bottom-line: Both main reporting modes are substantially faster (700% and 25% respectively), and a previously half-broken implementation of a feature of questionable utility was removed.

### DIFF
--- a/datalad/interface/rerun.py
+++ b/datalad/interface/rerun.py
@@ -708,6 +708,7 @@ def new_or_modified(diff_results):
     """Filter diff result records to those for new or modified files.
     """
     for r in diff_results:
-        if r.get('type') == 'file' and r.get('state') in ['added', 'modified']:
+        if r.get('type') in ('file', 'symlink') \
+                and r.get('state') in ['added', 'modified']:
             r.pop('status', None)
             yield r


yarikoptic:
diff looks surgical enough, seeing the CI (metalad known/reported) to be green is good news, well done!  I wish we had more targetted benchmark but even without such we suggestively observe a possible improvement among those we have:
```
         90.6±5ms         84.4±2ms     0.93  api.supers.time_subdatasets_recursive_first
...
         3.85±0.06s       3.56±0.08s     0.92  api.testds.time_create_test_dataset2x2
```

PR #6104, opened 2021-10-19
Introduce AnnexRepo.get_file_annexinfo(), deprecate AnnexRepo.get_file_key()

mih:
This is another step in a slow paradigm change to assemble more information with fewer calls to Git. Here it is `get_file_key()` which is often used in conjunction with `file_has_content()` (or a local `is_available()` which does the same thing with a different implementation), or `get_contentlocation()`, or `get_key|file_backend()`.

All this information is readily provided via `get_content_annexinfo()`. However, in most places, a query for a single files is made (that this can be wasteful is a problem, but also a reality right now). The new `get_file_annexinfo()` makes the single file query use case more convenient, but otherwise relies on `get_content_annexinfo()` as the work horse.

Beyond this addition the previous `get_file_key()` was reimplemented using `get_content_annexinfo()` alone. Simultaneously it is being deprecated, because better means exist now, to perform its tasks. For example the OpenNeuro usage referred to in https://github.com/datalad/datalad/pull/6104#issuecomment-946807095 is fully replaced by a single `get_file_annexinfo()` call now (not just the `get_file_key() `method call, but the entire loop body). Moreover, the behavior of that function is extremely complex (different error behavior, dependent on parameter types). It's exception behavior (3 different exceptions for why a key could not be obtained) also led to abuse of this method for type-checking of repository file types (call that didn't even bother checking the return value of a `get_*` method). And lastly, it uses `normalize_paths`.

TODO:
- [x] see if current tests pass
- [x] implement dedicated tests for `get_file_annexinfo()` pulling together test code for all the related methods listed above.
- follow up PR will then remove their usage too

Developer notes:

This is furthering the goal of finally getting rid of `normalize_paths()` #4595

This is a second and different attempt to this. The last one was https://github.com/datalad/datalad/pull/5069

Final progress ping to #3333

### DIFF
--- a/datalad/customremotes/archives.py
+++ b/datalad/customremotes/archives.py
@@ -114,7 +114,7 @@ class ArchiveAnnexCustomRemote(AnnexCustomRemote):
         if archive_file is not None:
             if archive_key is not None:
                 raise ValueError("Provide archive_file or archive_key - not both")
-            archive_key = self.repo.get_file_key(archive_file)
+            archive_key = self.repo.get_file_annexinfo(archive_file)['key']
         assert(archive_key is not None)
         attrs = OrderedDict()  # looking forward for more
         if file:
--- a/datalad/customremotes/tests/test_base.py
+++ b/datalad/customremotes/tests/test_base.py
@@ -46,7 +46,8 @@ def test_get_contentlocation(tdir):
     repo.add('file.dat')
     repo.commit('added file.dat')
 
-    key = repo.get_file_key('file.dat')
+    # TODO contentlocation would come with eval_availability=True
+    key = repo.get_file_annexinfo('file.dat')['key']
     cr = AnnexCustomRemote(tdir)
     key_path = cr.get_contentlocation(key, absolute=False)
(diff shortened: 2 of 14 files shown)


mih:
The need comes from the need to get rid of `normalize_paths` (#4595).

Right now, I am exploring what migration path makes sense. The usage in OpenNeuro code is a common one (ie., operation on a single file). Once I cataloged all our usage, I will know whether my current plan hold. This plan is to implement something like `.get_file_props(path)` as a convenience front-end for `.get_content_(annex)info()` for the single file use case.

The OpenNeuro usage transition would then become:

`get_file_key(name)` -> `get_file_props(name)['key']`

Instead of handling three types of exceptions (not in annex, not in git, not there), the full set up file properties can always be obtained for any known file and acted on, including the key.

Alternatively, we could consider just removing the normalize path decorator. But this would either require to deprecate all its usage at once (this is already taking years, and will still take more), or to silently break the behavior of the method. I consider my proposal less problematic, right now.

Issue #5919, opened 2021-08-19
Hint the user if lots of content was committed directly to git


yarikoptic:
#### What is the problem?

use cases: many in my own history, and a recent one https://neurostars.org/t/datalad-dataset-cleaning-questions/20052/9 

If at the end of save we could assess how much content was committed directly into git, and if it is notable (e.g. committed > 100MB or (.git/objects became > 100MB and commit increased it by over 100%)) -- inform the user

- if it is a pure git repo -- inform to use git-annex
- if it is an annexed repo -- inform the user to adjust .gitattributes to fit their case better

along side -- provide some extra guidance e.g. show largest file committed; point to ways to mitigate (somewhere in handbook?)

yarikoptic:
<details>
<summary>I guess it might be easiest if extra field `size` was provided to returned records, or we could interrogate files upon analysis/preparing the summary from those records, skipping annex'ed keys:
</summary> 

```shell
$> datalad -f json_pp save 123
{                                                                                                         
  "action": "add",                                                                                        
  "key": "MD5E-s4--ba1f2511fc30423bdbb183fe33f3dd0f",
  "message": "",
  "path": "/tmp/testds/123",
  "refds": "/tmp/testds",
  "status": "ok",
  "type": "file"
}
$> datalad -f json_pp save --to-git 125
{                                                                                                         
  "action": "add",
  "key": null,
  "path": "/tmp/testds/125",
  "refds": "/tmp/testds",
  "status": "ok",
  "type": "file"
}

```
</details>

PR #6108, opened 2021-10-21
NF: Hint the user if lots of content was committed directly to git (WIP)

DisasterMo:
### Description

As suggested by #5919, this implements a notification triggered by a large amount of data being saved directly to git (current default threshold 100 MiB). The note currently displays the total number and size of files saved to git as well as the largest file(s) and their (individual) size. It can be disabled via `datalad.ui.git-commit-notification` and the threshold can be adjusted via `datalad.ui.git-commit-notification-threshold`.

Any comments and suggestions are very welcome.

### TODO

- [ ] check if dataset is pure git repo
- [x] inform the user to adjust .gitattributes
- [ ] point to ways to mitigate
- [ ] add unit test(s)  (item added by @yarikoptic)

### DIFF
--- a/datalad/core/local/save.py
+++ b/datalad/core/local/save.py
@@ -381,18 +381,65 @@ class Save(Interface):
         # and more "dynamic" feedback than jumpy datasets count.
         # See addurls where it is implemented that way by providing agg and another
         # log_filter
-        yield from ProducerConsumerProgressLog(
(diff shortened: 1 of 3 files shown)


mih:
Unless someone expresses interest in picking up this PR, I will archive it in a few days, to be picked up later, when this issue bubbles up again. Thx for the investigation so far.

mih:
Closing now as announced a month ago.

PR #6111, opened 2021-10-25
Reimplementation of drop/(uninstall)/remove

### LINKED ISSUE #1142: uninstall must check whether a dataset has a non-local remote and all changes are pushed

### LINKED ISSUE #1282: bring --kill back?
ATM the only "fast" way to remove some temp dataset is 
```shell
datalad remove --nocheck --if-dirty=ignore PATH
```
as far as I see it.  Could/should we have something more concise and easy to discover, e.g. --kill which would set those necessary options to achieve desired behavior? ;)

### LINKED ISSUE #1823: remove  --nocheck is not just "nocheck" but also "nodrop"
also it implies that I cannot remove the file (well -- a "handle"), while avoiding dropping its content. So now

- if I have two files pointing to the same content, removing one of it would drop the load (unless I use `--nocheck`)

I think it would be more coherent and transparent if

- change default behaviour by adding  `--drop` flag to explicitly signal that content should be dropped 
- `--nocheck` would imply only absent checking while dropping content, so drop would still happen, just by calling drop with `--nocheck` argument

### LINKED ISSUE #1953: -J support for datalad drop
#### What is the problem?
nothere

### LINKED ISSUE #2229: `remove` confuses
Tried to remove a directory that I accidentally confused with a subdataset root. The actual subdataset is in the `dicoms` subdirectory. That could probably be read from the error message below, but I couldn't immediately decipher it.

```
% datalad remove -d . wz29_1206
drop(notneeded): /home/mih/pd/scratch/tempelmann/wz29_1206 (directory) [nothing to drop from /home/mih/pd/scratch/tempelmann/wz29_1206]
Failed to run ['git', '-c', 'receive.autogc=0', '-c', 'gc.auto=0', 'rm', '-r', '--', 'wz29_1206'] under '/home/data/psyinf/scratch/tempelmann'. Exit code=1. out= err=error: the following submodule (or one of its nested submodules)
uses a .git directory:
    wz29_1206/dicoms
(use 'rm -rf' if you really want to remove it including all of its history)
```

### LINKED ISSUE #2328: --all (or just --git-annex-options?) for datalad drop
#### What is the problem?
ATM we provide no interface to drop content from previous versions etc.  I thought that either we would need to provide some "helpers" (`--all`, `--unused`, `--not-referenced`) which would translate one-to-one to some pre-crafted git-annex drop commands, or just make it possible to specify options for git-annex drop for maximum flexibility, while benefiting from our `-r` etc functionality

### LINKED ISSUE #2655: `remove --if-dirty` default handling is misleading
Default is `save-before`. Hence any modification is saved before a remove it attempted. However, this is independent of a possible `check=False`. If that is given, any modification will be saved, but then everything is wiped out with no further checks. This will effectively cause unopposed removal of locally modified  datasets.

One could argue whether this is a bug, or a more general flaw -- we are presently not checking, whether a dataset is pushed to any remote (or if any of present subdatasets are not pushed).

### LINKED ISSUE #2673: --no-drop for datalad remove
#### What is the problem?
ATM `datalad remove` would also drop the content.  In my usecase, I just want `git rm` but with all our paths handling across subdatasets, and I think it is not possible (according to review of the cmdline options)

### LINKED ISSUE #2967: Make uninstall uninstall top-level datasets

### LINKED ISSUE #3887: remove() should announce `annex dead`

### LINKED ISSUE #4097: Even remove cannot remove without using `-d` way to specify

### LINKED ISSUE #4115: Just a suggestion: datalad remove (make default --if-dirty fail)
#### What is the problem?
http://docs.datalad.org/en/latest/generated/man/datalad-remove.html
Looks like `datald remove` will by default save a dataset before starting the process. I would suggest the default value of `--if-dirty` is switched to fail, this way the user by default is faced with the problem rather than the software making the decision. It would then make sense that a user faced with the problem selects either `--if-dirty save` or `--if-dirty ignore` to proceed. I know changes in default values are not easy decisions, but the defaults usually should (1) make selections for naive users which all of us are at the beginning, and (2) follow the most common use (which does not seem the right thing to do for most users; trigger saving all pending changes when removing a single file). Just a thought/suggestion. 

NA

0.12.2 (github today)

Nope.

Sure, love is in the air :)

### LINKED ISSUE #4784: remove: fails to remove in --recursive, demands --recursive option
```shell
smaug:~/proj/repronim/containers
*$> datalad remove --nocheck -r /home/yoh/.tmp/repro-Ut8tPTv/ds000003-qc/
[WARNING] path constraints did not match an installed subdataset: /home/yoh/.tmp/repro-Ut8tPTv/ds000003-qc/sourcedata 
[WARNING] path constraints did not match an installed subdataset: /home/yoh/.tmp/repro-Ut8tPTv/ds000003-qc/containers 
[ERROR  ] to be uninstalled dataset Dataset(/home/yoh/.tmp/repro-Ut8tPTv/ds000003-qc) has present subdatasets, forgot --recursive? [remove(/home/yoh/.tmp/repro-Ut8tPTv/ds000003-qc)] 
remove(error): /home/yoh/.tmp/repro-Ut8tPTv/ds000003-qc (dataset) [to be uninstalled dataset Dataset(/home/yoh/.tmp/repro-Ut8tPTv/ds000003-qc) has present subdatasets, forgot --recursive?]

$> datalad --version
datalad 0.13.1

# is it a symlink?
$> readlink -f /home/yoh/.tmp/repro-Ut8tPTv/ds000003-qc/
/home/yoh/.tmp/repro-Ut8tPTv/ds000003-qc
```

The script used to produce that dataset is https://github.com/ReproNim/containers#a-typical-workflow and to reproduce you could cut/paste or

```
git clone http://github.com/ReproNim/containers
bash <(sed -n -e '/^ *#!/,/^ *$/p' containers/README.md) 
```

### LINKED ISSUE #5842: (Re)design of `remove` as `purge`
Companion of #5832 (to be discussed in conjunction)

Context: This is about approaching an improved implementation of the `remove` command, not in isolation, but in the context of a reconsideration of the drop/uninstall/remove combo, with the partial goal to replace the old remove entirely, and with that make progress towards #3368

This is an opportunity to reconsider the role of `remove` in the API, and amend it with the missing functionality we have identified (issues are linked below). This draft description is not about making a contrast to what the present `remove` is doing (there are massive overlaps), but about a comprehensible concept of what as companion of a refurbished `drop` (#5832) would need to do, and not do. The initial design draft is the result of a discussion of @adswa and @mih.

### Basic concept

`purge` is the companion of `drop` with a focus on "destructive" operations. It could be thought of as the antagonist to `clone`. It is used fully wipe out datasets, including top-most superdatasets.

`purge` is called `purge` not `remove` to indicate the severity of the operation, and set it semantically apart from operations like `git rm`, which are typically reversible. This simplifies user expectations and avoid clunky alternative indicator (#4097).

`purge` is needed for two things:
1. to delete (sub-)datasets, because repositories and submodules are under special protection by git-annex
2. to trigger a adequate `drop` call (the intention get rig of file content is semantically more clear when calling `drop`, than when saving the deletion of a file. Actual dropping behavior can be tweaked in `purge` (see below), not in `save`)

`purge` prevents data loss by default. It inherits a number of parameters from `drop`, for which it typically acts as a frontend to benefit from its data security mechanisms (this includes primarily `--reckless`, and `--jobs`). `purge` supports a dedicated `--reckless` mode to immediately wipe out a dataset (hierarchy) with not checks whatsoever (#1282, #2673).

`purge` should be able to recognize a botched attempt to remove a dataset with a plain `rm -rf`, and act on it in a meaningful way, even if it is just hinting at `chmod + rm -rf`.

### Use cases

The following use cases operate in the dataset hierarchy depicted below.

```
super
├── dir
│   ├── fileD1
│   └── fileD2
├── fileS1
├── fileS2
├── subA
│   ├── fileA
│   ├── subsubC
│   │   ├── fileC
│   └── subsubD
└── subB
    └── fileB
```

Unless explicitly stated, all command are assumed to be executed in the root of `super`.

- `datalad purge subB`
   Wipe out `subB`, leave `super` dirty (analog to `datalad clone` behavior)
- `datalad purge -d . subB`
   Wipe out `subB`, save removal in `super` (analog to `datalad clone -d .` behavior)

- `datalad purge fileS2`
   commits deletion, like `rm fileS2 + datalad save`, but with a prior `drop`!
- `datalad purge dir`
   like above, commits deletion of the entire directory and its content, unless subdatasets are found underneath (errors in this case)

- `datalad purge subA`
   errors because of present subdatasets
- `datalad purge subA -r`
   wipes out subdataset hierarchy, leaves `super` dirty

- `datalad purge .` (purge in `super`)
  errors, because of subdataset presence

- `cd .. && datalad purge super -r`
   eradicates super and its subdatasets entirely

The following use case are executed in `subB/` (equivalent to a `super` without subdatasets present)

- `datalad purge .`
  removes all files in the dataset, does *not* purge the dataset
- `datalad purge -d .`
  refuses to wipe out the entire dataset that is in CWD (like `rm -rf` does)
- `cd .. && datalad purge subB`
  wipes out `subB` with all its content and the dataset itself too

### LINKED ISSUE #6123: Dataset.get_superdataset() calls subdatasets() with result renderer
It should be silent.

mih:
Sits on top of #6110 and will grow further. For now just a demo of the state for the curious.

Please help with naming! We need labels for

- `--what`
   - keys detetermined by paths in the worktree (default operation): `filecontent`?
   - to engage `git annex drop --auto`: `unwanted`?
   - to engage `git annex drop --all`: `allkeys`?
   - to uninstall entire datasets: `all`?

- `--reckless`
   - to disable minimum key redundancy checks: `availability`
   - to disable test if all branches are pushed: `availability` (yes, same as above -- same recklessness different subject only. one matters for file drop, the other for dataset uninstall), @adswa gave example, when the two should be selectable individually, though
   - to disable checks for unsaved modifications: `modification`
   - to disable announcement/push of `git annex dead here` outcome: `undead`

- Fixes #1142
- Fixes #1282
- Fixes #1823
- Fixes #1953
- Fixes #2229
- Fixes #2328
- Fixes #2655
- Fixes #2673
- Fixes #2967
- Fixes #3887
- Fixes #4097
- Fixes #4115
- Fixes #4784
- Fixes #5842
- Fixes #6123 

TODO:

- [x] drop in replacement for the previous `drop` implementation
- [x] remove old `drop` implementation
- [x] 95+% percent test coverage, using a new set of tests that are confirmed to work cross-platform
- [x] support passing `--jobs` to `git annex drop` #1953
- [x] support dropping all keys from all branches #2328 (criticial precondition for safely uninstalling datasets)
- [x] drop in replacement for `uninstall`
- [x] decide if with `drop --what all --reckless availability` we should even attempt to drop all keys. We could simply wipe out the repo, and be done (I feel like we discussed this before). already possible to bypass with `kill`
- [x] should the outcome of `git annex dead here` be pushed to the default remote (if any), or all remotes? What if one of them fails? What if it needs a merge first?
       No! If, and only if, the existence of a local annex was ever communicated elsewhere (most would not), an error message must inform users what to do to communicate its pending death. But not automatically inside `drop`.
- [x] check for potential undead annex records #3887
- [x] same for all branches: should they be pushed to the default remote (if any), or all remotes? What if one of them fails? What if it needs a merge first?
      Same no! An error message should inform about this, and instruct what to do (i.e. datalad push), but not do anything automatically inside `drop`
- [x] verify branch state availability before dropping an entire dataset
- [x] can drop toplevel datasets #2967
- [x] go through use case list at http://docs.datalad.org/en/stable/design/drop.html and confirm/deny support for each
- [x] document commands
- [x] decide on "unpushed" detection with adjusted branches. I think we should `localsync()` before testing this, and then only report the corresponding branches. The adjusted ones are not meant to be pushed anyways.
- [x] bring back `kill`
   image
- [x] `uninstall` is a thin shim of a command around `drop`, old tests pass with minimal adjustments for increased safety
- [x] rewrite `remove` as a thin wrapper around `drop` and `save`, and remove all old helper code.
- [x] made the last usage of path annotation obsolete and dropped the entire code collection that was required to do this previously
- [x] rename `remove(what=...`) to `remove(drop=...)`
- [x] PR against -deprecated to move `AnnotatePaths` there (`publish` is built on it) https://github.com/datalad/datalad-deprecated/pull/46
- [x] `reckless=kill` should maybe be `kill-dataset`, it won't actually kill anything, when given a file path to drop, or actually affect file dropping too


datalad/distributed/tests/test_drop.py
```diff
@@ -0,0 +1,432 @@
...
+    cloneds.push(data='nothing')
+
+    res = cloneds.drop(what='all', on_failure='ignore')
+    assert_in_results(res, action="drop", status="error")
+    assert_true(res[0]['message'][0].startswith(
+        "to-be-deleted local annex not declared 'dead'"))
+    # some windows test setup is not very robust, explicitly
+    # include the default name "origin" in the test success
+    # conditions to make this more robust
+    assert_in(res[0]['message'][1], ['dl-test-remote', 'origin'])
```
adswa, line 378: I think the reason for the test failure is that your testing something like
```
>>> ['ab'] in ['ab', 'cd']
False
```
when it actually should be
```
>>> 'ab' in ['ab', 'cd']
True
```

adswa:
I hopefully pushed a fix for the failing test

datalad/distributed/tests/test_drop.py
adswa: 4c88098 hopefully fixed this

adswa:
Judging from the currently running tests, the test in question is fixed. But I just tried on windows and 

```
eq_(res[0]['message'][1], ['dl-test-remote'])
```

(without the origin) works just as well. maybe this works across platforms, too

datalad/distributed/tests/test_drop.py
mih: Argh, stupid me. Thx!

mih:
So the reason underlying the failing `remove` test is that `status` cannot see the removal, hence `save` doesn't act on it.

After `remove` ran, the subdataset in question is gone, and the subproject removal 
```
C:\DLTMP\datalad_temp__eocjulu>git status
On branch adjusted/master(unlocked)
  (use "git add/rm <file>..." to update what will be committed)
        modified:   three (new commits)
        deleted:    two

C:\DLTMP\datalad_temp__eocjulu>git diff HEAD -- two
diff --git a/two b/two
deleted file mode 160000
index beed3a2..0000000
--- a/two
+++ /dev/null
@@ -1 +0,0 @@
-Subproject commit beed3a2f14cbc57f8efe67de9a375eb81c809933

C:\DLTMP\datalad_temp__eocjulu>datalad status
```

This needs to be investigated.

On appveyor we have `git version 2.26.2.windows.1`, on another windows box with `git version 2.33.0.windows.2` the exact same test passes.

adswa:
I can trigger the failure, but onlz bz downgrading Git

```
C:\Users\datalad\repos\datalad>git --version
git version 2.33.0.windows.2

C:\Users\datalad\repos\datalad>python -m nose -s -v datalad/distribution/tests/test_uninstall.py:test_clean_subds_removal
datalad.distribution.tests.test_uninstall.test_clean_subds_removal ... subdataset(ok): one (dataset)
subdataset(ok): two (dataset)
subdataset(ok): two (dataset)
subdataset(ok): three (dataset)
subdataset(ok): two (dataset)
subdataset(ok): three (dataset)
subdataset(ok): two (dataset)
subdataset(ok): three (dataset)
ok
Versions: annexremote=1.5.0 appdirs=1.4.4 boto=2.49.0 cmd:7z=19.00 cmd:annex=8.20210804-g1d3f59a9d cmd:bundled-git=UNKNOWN cmd:git=2.33.0.windows.2 cmd:system-git=2.33.0.windows.2 cmd:system-ssh=for humanize=3.2.0 iso8601=0.1.14 keyring=23.0.0 keyrings.alt=UNKNOWN msgpack=1.0.2 requests=2.25.1 wrapt=1.12.1
Obscure filename: str=b";&%b5{}'.datc" repr=";&%b5{}'.datc"
Encodings: default='utf-8' filesystem='utf-8' locale.prefered='cp1252'
Environment: PATH='C:\\Program Files\\Docker\\Docker\\Resources\\bin;C:\\WINDOWS\\system32;C:\\WINDOWS;C:\\WINDOWS\\System32\\Wbem;C:\\WINDOWS\\System32\\WindowsPowerShell\\v1.0\\;C:\\WINDOWS\\System32\\OpenSSH\\;C:\\ProgramData\\chocolatey\\bin;C:\\Program Files (x86)\\Gpg4win\\..\\GnuPG\\bin;C:\\Program Files\\PuTTY\\;C:\\Program Files\\Git\\cmd;C:\\Users\\datalad\\AppData\\Roaming\\local\\bin;C:\\Users\\datalad\\AppData\\Local\\Programs\\Python\\Python39\\Scripts\\;C:\\Users\\datalad\\AppData\\Local\\Programs\\Python\\Python39\\;C:\\Users\\datalad\\AppData\\Local\\Microsoft\\WindowsApps;C:\\Program Files (x86)\\GnuWin32\\bin;C:\\Program Files\\7-Zip;;C:\\Users\\datalad\\AppData\\Local\\Programs\\Microsoft VS Code\\bin' GIT_CONFIG_PARAMETERS="'init.defaultBranch=dl-test-branch' 'clone.defaultRemoteName=dl-test-remote'"

Ran 1 test in 137.071s

OK

C:\Users\datalad\repos\datalad>git --version
git version 2.26.0.windows.1

C:\Users\datalad\repos\datalad>python -m nose -s -v datalad/distribution/tests/test_uninstall.py:test_clean_subds_removal
datalad.distribution.tests.test_uninstall.test_clean_subds_removal ... subdataset(ok): one (dataset)
subdataset(ok): two (dataset)
subdataset(ok): two (dataset)
subdataset(ok): three (dataset)
subdataset(ok): two (dataset)
subdataset(ok): three (dataset)
subdataset(ok): two (dataset)
FAIL
Versions: annexremote=1.5.0 appdirs=1.4.4 boto=2.49.0 cmd:7z=19.00 cmd:annex=8.20210804-g1d3f59a9d cmd:bundled-git=UNKNOWN cmd:git=2.26.0.windows.1 cmd:system-git=2.26.0.windows.1 cmd:system-ssh=for humanize=3.2.0 iso8601=0.1.14 keyring=23.0.0 keyrings.alt=UNKNOWN msgpack=1.0.2 requests=2.25.1 wrapt=1.12.1
Obscure filename: str=b";&%b5{}'.datc" repr=";&%b5{}'.datc"
Encodings: default='utf-8' filesystem='utf-8' locale.prefered='cp1252'
Environment: PATH='C:\\Program Files\\Docker\\Docker\\Resources\\bin;C:\\WINDOWS\\system32;C:\\WINDOWS;C:\\WINDOWS\\System32\\Wbem;C:\\WINDOWS\\System32\\WindowsPowerShell\\v1.0\\;C:\\WINDOWS\\System32\\OpenSSH\\;C:\\ProgramData\\chocolatey\\bin;C:\\Program Files (x86)\\Gpg4win\\..\\GnuPG\\bin;C:\\Program Files\\PuTTY\\;C:\\Program Files\\Git\\cmd;C:\\Users\\datalad\\AppData\\Roaming\\local\\bin;C:\\Users\\datalad\\AppData\\Local\\Programs\\Python\\Python39\\Scripts\\;C:\\Users\\datalad\\AppData\\Local\\Programs\\Python\\Python39\\;C:\\Users\\datalad\\AppData\\Local\\Microsoft\\WindowsApps;C:\\Program Files (x86)\\GnuWin32\\bin;C:\\Program Files\\7-Zip;;C:\\Users\\datalad\\AppData\\Local\\Programs\\Microsoft VS Code\\bin' GIT_CONFIG_PARAMETERS="'init.defaultBranch=master' 'clone.defaultRemoteName=origin'"

FAIL: datalad.distribution.tests.test_uninstall.test_clean_subds_removal
  File "C:\Users\datalad\AppData\Local\Programs\Python\Python39\lib\site-packages\nose\case.py", line 198, in runTest
    self.test(*self.arg)
  File "C:\Users\datalad\repos\datalad\datalad\tests\utils.py", line 737, in _wrap_with_tempfile
  File "C:\Users\datalad\repos\datalad\datalad\distribution\tests\test_uninstall.py", line 113, in test_clean_subds_removal
    assert_in_results(
  File "C:\Users\datalad\repos\datalad\datalad\tests\utils.py", line 1325, in assert_in_results
    _check_results_in(True, results, **kwargs)
  File "C:\Users\datalad\repos\datalad\datalad\tests\utils.py", line 1318, in _check_results_in
    raise AssertionError(msg.format(_format_res(kwargs),
AssertionError: Desired result
  {
   "action": "remove",
   "path": "C:\\Users\\datalad\\AppData\\Local\\Temp\\datalad_temp_xahpplwj\\two"
  }
not found among
  [
   {
    "action": "drop",
    "message": [
     "nothing to drop from %s",
     "C:\\Users\\datalad\\AppData\\Local\\Temp\\datalad_temp_xahpplwj\\two"
    ],
    "path": "C:\\Users\\datalad\\AppData\\Local\\Temp\\datalad_temp_xahpplwj\\two",
    "refds": "C:\\Users\\datalad\\AppData\\Local\\Temp\\datalad_temp_xahpplwj",
    "status": "notneeded",
    "type": "directory"
   },
   {
    "action": "save",
    "path": "C:\\Users\\datalad\\AppData\\Local\\Temp\\datalad_temp_xahpplwj",
    "refds": "C:\\Users\\datalad\\AppData\\Local\\Temp\\datalad_temp_xahpplwj",
    "status": "notneeded",
    "type": "dataset"
   }
  ]
datalad.utils: Level 5: Importing datalad.utils
datalad.utils: DEBUG: Failed to query or got useless SC_ARG_MAX sysconf, will use hardcoded value: module 'os' has no attribute 'sysconf'
datalad.utils: DEBUG: Maximal length of cmdline string (adjusted for safety margin): 16383
datalad.utils: Level 5: Done importing datalad.utils
datalad.runner.gitrunner: Level 9: Will use git under '' (no adjustments to PATH if empty string)
datalad.runner.runner: DEBUG: Run ['git', '--git-dir=', 'config', '-z', '-l', '--show-origin'] (cwd=None)
datalad.runner.protocol: Level 8: Process 2164 started
datalad.runner.nonasyncrunner: Level 5: ReaderThread(<_io.FileIO name=4 mode='rb' closefd=True>, <queue.Queue object at 0x000002A9EBF6ECD0>, ['git', '--git-dir=', 'config', '-z', '-l', '--show-origin']) started
datalad.runner.nonasyncrunner: Level 5: ReaderThread(<_io.FileIO name=3 mode='rb' closefd=True>, <queue.Queue object at 0x000002A9EBF6ECD0>, ['git', '--git-dir=', 'config', '-z', '-l', '--show-origin']) started
datalad.runner.protocol: Level 5: Read 1024 bytes from 2164[stdout]
datalad.runner.protocol: Level 5: Read 604 bytes from 2164[stdout]
datalad.runner.nonasyncrunner: Level 5: ReaderThread(<_io.FileIO name=4 mode='rb' closefd=True>, <queue.Queue object at 0x000002A9EBF6ECD0>, ['git', '--git-dir=', 'config', '-z', '-l', '--show-origin']) exiting (stream end)
datalad.runner.nonasyncrunner: Level 5: ReaderThread(<_io.FileIO name=3 mode='rb' closefd=True>, <queue.Queue object at 0x000002A9EBF6ECD0>, ['git', '--git-dir=', 'config', '-z', '-l', '--show-origin']) exiting (stream end)
datalad.runner.protocol: Level 8: Process 2164 exited with return code 0
datalad.runner.runner: DEBUG: Finished ['git', '--git-dir=', 'config', '-z', '-l', '--show-origin'] with status 0
datalad.ui: Level 5: Starting importing ui
datalad.ui.dialog: Level 5: Starting importing ui.dialog
datalad.ui.dialog: Level 5: Done importing ui.dialog
datalad.ui: Level 5: Initiating UI switcher
datalad.ui: DEBUG: UI set to DialogUI(out=<StreamWrapper>)
datalad.ui: Level 5: Done importing ui
--------------------- >> end captured logging << ---------------------

Ran 1 test in 136.252s

FAILED (failures=1)
```

mih:
Yes, same here -- not on Windows, but also on my Debian laptop. With Git 2.26.2 `status` does not see the change. Investigating...

adswa:
with ``--what allkeys`` I don't see any path reporting - is this intentional? I makes it hard to guess which file only has one copy

```
❱ datalad drop --what allkeys
drop(ok): . (key)
drop(error): . (key) [unsafe
Could only verify the existence of 0 out of 1 necessary copy
(Use --reckless to override this check, or adjust numcopies.)]
drop(ok): . (key)
action summary:
  drop (error: 1, ok: 2)
```

Compared to 

```
❱ datalad drop .
drop(error): archive.tar.gz (file) [unsafe; Could only verify the existence of 0 out of 1 necessary copy; (Use --reckless to override this check, or adjust numcopies.)]
drop(ok): archive/3 (file)
drop(ok): archive/5 (file)
drop(impossible): . (directory) [could not drop some content in /tmp/abcd ['/tmp/abcd/archive.tar.gz']]
action summary:
  drop (error: 1, impossible: 1, ok: 2)
```

mih:
> with `--what allkeys` I don't see any path reporting - is this intentional? I makes it hard to guess which file only has one copy

True. The key drop results are associated with the path of the dataset. In full, they look like this

```
% dl -f json_pp drop --what allkeys
{
  "action": "drop",
  "key": "MD5E-s4--ba1f2511fc30423bdbb183fe33f3dd0f",
  "message": "unsafe\nCould only verify the existence of 0 out of 1 necessary copy\n(Use --reckless to override this check, or adjust numcopies.)",
  "path": "/tmp/top",
  "refds": "/tmp/top",
  "status": "error",
  "type": "key"
}
```

We could think about a dedicated result renderer that searches for key usage. However, I general it is not guaranteed that the is just one file associated with a key, or any file with a path in the current worktree, or even any file at all (unused key).

adswa:
> We could think about a dedicated result renderer that searches for key usage. However, I general it is not guaranteed that there is just one file associated with a key, or any file with a path in the current worktree, or even any file at all (unused key).

Would it be an option to just not report paths then? Just a count of successful operations, for example?

The linebreak after ``unsafe`` looks a bit funky as well:

```
drop(error): HCP1200/996782 (key) [unsafe
Could only verify the existence of 0 out of 1 necessary copy
(Note that these git remotes have annex-ignore set: origin)
(Use --reckless to override this check, or adjust numcopies.)]
```

adswa:
Sorry for all the edge cases I'm digging up, feel free to ignore if they get too obscure:

The automatic hint to use ``--reckless`` when file availability checks need to be overridden reads a bit weird when I've used the ``--reckless parameter`` for a different type of reckless behavior:

e.g., here is a dataset with  untracked content, and unique content, and the check that warns about the lack of copies for the unique key points to the parameter I'm already using. 

```
❱ datalad drop HCP1200/996782/ --reckless modification --what datasets              
[WARNING] Failed to load entrypoint crawler: DistributionNotFound(The 'datalad_deprecated' distribution was not found and is required by the application) 
drop(error): HCP1200/996782 (key) [unsafe
Could only verify the existence of 0 out of 1 necessary copy
(Note that these git remotes have annex-ignore set: origin)
(Use --reckless to override this check, or adjust numcopies.)]
```

Would it make sense to specify ``--reckless availability`` in this hint?

datalad/distributed/drop.py
```diff
@@ -0,0 +1,791 @@
...
+                    f"{repr(value)} [{str(e)}]") from e
+
+        # TODO deprecation warning on `check`
+        if check is False:
+            # TODO check for conflict with new reckless parameter
+            reckless = 'availability'
+
+        if what in ('all', 'datasets') and reckless == 'kill' and not recursive:
+            raise ValueError(
+                'A reckless kill is requested by no recursion flag is set. '
```
adswa, line 201:
```suggestion
                'A reckless kill is requested but no recursion flag is set. '
```

mih:
> Would it be an option to just not report paths then? Just a count of successful operations, for example?

The `path` property is the only indication to know from which dataset keys are being dropped (this could be a recursive operation).

> The linebreak after `unsafe` looks a bit funky as well:

This is a message from git-annex. it has the newline already. Not sure if we want to mess with it further, but we could.

> Would it make sense to specify `--reckless availability` in this hint?

It would, yes.

adswa:
Just a note to myself: I see that the new drop became even more careful. Previously, a newly created, empty dataset could be just removed, now I get a warning that the ``master`` branch isn't pushed anywhere:

```
(handbook2) adina@muninn in /tmp/abcd on git:master
❱ datalad drop --what datasets acb 
uninstall(error): acb (dataset) [to-be-dropped dataset has the following revisions that are not available at any known sibling. Use `datalad push --to ...` to push these before dropping the local dataset, or ignore via `--reckless availability`: ['master']]
(handbook2) adina@muninn in /tmp/abcd on git:master
❱ datalad remove -d acb                                                                                                           
uninstall(error): . (dataset) [to-be-dropped dataset has the following revisions that are not available at any known sibling. Use `datalad push --to ...` to push these before dropping the local dataset, or ignore via `--reckless availability`: ['master']]

```

mih:
> Just a note to myself: I see that the new drop became even more careful. Previously, a newly created, empty dataset could be just removed, now I get a warning that the `master` branch isn't pushed anywhere:

Yes, and on top of that: If you ever pushed a dataset elsewhere, it will complain about undead annexes too, and refuse dropping.

datalad/distributed/drop.py
```diff
@@ -0,0 +1,791 @@
...
+            potential loss of unsaved or untracked dataset components.
+            With 'availability', detection of dataset/branch-states that are
+            only available in the local dataset, and detection of an
+            insufficient number of file-content copies will be disabled.
+            Especially the latter is a potentially expensive check which might
+            involve numerous network transactions.
+            With 'undead', detection whether a to-be-removed local annex is
+            still known to exists in the network of dataset-clones is disabled.
+            This could cause zombie-records of invalid file availability.
+            With 'kill', all safety-checks are disabled.""",
```
adswa, line 132: When I use ``kill``, I get a value error for not having used ``--recursive`` as well. Maybe its useful to document this here, too?

Also, I learned that kill only operates on datasets, and wouldn't kill filecontent:

```
❱ datalad save
add(ok): file (file)                                                                                                                  
save(ok): . (dataset)                                                                                                                 
action summary:                                                                                                                       
  add (ok: 1)
  save (ok: 1)
(handbook2) adina@muninn in /tmp/abcd on git:master
❱ datalad drop --what filecontent --reckless kill file 
(handbook2) adina@muninn in /tmp/abcd on git:master
❱ ls
acb  file
```
Maybe this could be stated in the docstring, too? Or it could yield an "impossible" result when the path to be dropped is a file?

datalad/distributed/drop.py
```diff
@@ -0,0 +1,791 @@
...
+                action='uninstall',
+                path=ds.path,
+                type='dataset',
+                status='error',
+                message=(
+                    "to-be-dropped dataset has the following revisions that "
+                    "are not available at any known sibling. Use "
+                    "`datalad push --to ...` to push "
+                    "these before dropping the local dataset, "
+                    "or ignore via `--reckless availability`: %s",
```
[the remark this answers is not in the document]
mih: Makes sense! Thx!

datalad/distributed/drop.py
mih: I am afraid the only clean approach would be to change the logic to also have `kill` impact dropping file content.

datalad/distributed/drop.py
```diff
@@ -0,0 +1,792 @@
...
+                 "remote sources for dropped data",
+         'code_py': "drop(path='path/to/content', reckless='availability')",
+         'code_cmd': "datalad drop <path/to/content> --reckless availability"},
+        {'text': 'Drop (uninstall) an entire dataset '
+                 '(will fail with subdatasets present)',
+         'code_py': "drop(what='all')",
+         'code_cmd': "datalad drop --what all"},
+        {'text': 'Kill a dataset recklessly with any existing subdatasets too'
+                 '(this will be fast, but will disable any and all safety '
+                 'checks)',
```
adswa, line 98:
```suggestion
        {'text': 'Kill a dataset recklessly with any existing subdatasets too '
                 '(this will be fast, but will disable any and all safety '
                 'checks)',
```

There is a missing space:

```
Kill a dataset recklessly with any existing subdatasets too(this will
be fast, but will disable any and all safety checks)::
```

datalad/distributed/drop.py
mih: Done

datalad/local/remove.py
```diff
@@ -0,0 +1,229 @@
...
+    # if_dirty and check as deprecated
+    for p in ('reckless', 'if_dirty', 'check',):
+        _params_[p] = Drop._params_[p]
+
+    _examples_ = [
+        dict(text="Permanently remove a subdataset from a dataset",
+             code_py="remove(dataset='path/to/dataset', path='path/to/subds')",
+             code_cmd="datalad remove -d <path/to/dataset> <path/to/subds>"),
+    ]
+
```
adswa, line 123: maybe emphasize the inherent recursiveness of operations in the examples? And maybe also mention removal of a superdataset as an example (it is my most used usecase for remove)

```suggestion
    _examples_ = [
        dict(text="Permanently remove a subdataset (and all further subdatasets contained in it) from a dataset",
             code_py="remove(dataset='path/to/dataset', path='path/to/subds')",
             code_cmd="datalad remove -d <path/to/dataset> <path/to/subds>"),
             dict(text="Permanently remove a superdataset (with all subdatasets) from the filesystem",
             code_py="remove(dataset='path/to/dataset')",
             code_cmd="datalad remove -d <path/to/dataset>"),
    ]

```

datalad/local/remove.py
```diff
@@ -0,0 +1,229 @@
...
+    behavior can be changed to always drop content prior removal, for cases
+    where a minimal storage footprint for local datasets installations is
+    desirable.
+
+    Removing a dataset component is always a recursive operation. Removing a
+    directory, removes all content underneath the directory too. If
+    subdatasets are located under a to-be-removed path, they will be
+    uninstalled entirely, and all their content dropped. If any subdataset
+    can not be uninstalled safely, the remove operation will fail and halt.
+    """
```
adswa, line 74: Maybe this is the right place to mention that remove won't operate on ``.``?

datalad/local/remove.py
```diff
@@ -0,0 +1,229 @@
...
+            if res.get('status') not in ('ok', 'notneeded'):
+                drop_success = False
+            yield res
+
+        if not drop_success:
+            # there will be 'rm -rf' below, so play safe
+            lgr.debug('Observed drop failure, will not attempt remove')
+            return
+
+        for dpath, paths in paths_by_ds.items():
```
adswa, line 176: Not yet sure where this problem is rooted, but ``remove`` fails to save a subdataset deletion if it is a hierarchy if datasets:

```
# remove a single level of subdatasets is fine:
❱ datalad get -n HCP1200/996782   
[INFO   ] Fetching 'http://store.datalad.org/89d/4899e-1fe6-11ea-9e62-002590496000/config' 
install(ok): /home/adina/repos/human-connectome-project-openaccess/HCP1200/996782 (dataset) [Installed subdataset in order to get /home/adina/repos/human-connectome-project-openaccess/HCP1200/996782]
❱ datalad remove HCP1200/996782                                                                                                   
uninstall(ok): HCP1200/996782 (dataset)
remove(ok): HCP1200/996782 (file)                                                                                                     
save(ok): . (dataset)                                                                                                                 
action summary:                                                                                                                       
  remove (ok: 1)
  save (ok: 1)
  uninstall (ok: 1)
```


```
# removal of many levels of subdatasets leaves the superdataset dirty:
❱ datalad get -n -r HCP1200/996782

[INFO   ] Installing Dataset(/home/adina/repos/human-connectome-project-openaccess/HCP1200/996782) to get /home/adina/repos/human-connectome-project-openaccess/HCP1200/996782 recursively 
[INFO   ] scanning for annexed files (this may take some time) 
[INFO   ] Fetching 'http://store.datalad.org/937/b2e76-1fe6-11ea-8fe3-002590496000/config' 
install(ok): /home/adina/repos/human-connectome-project-openaccess/HCP1200/996782/MNINonLinear (dataset)                              
[INFO   ] Fetching 'http://store.datalad.org/94f/c881c-1fe6-11ea-8fe3-002590496000/config'                                            
install(ok): /home/adina/repos/human-connectome-project-openaccess/HCP1200/996782/T1w (dataset)                                       
[INFO   ] Fetching 'http://store.datalad.org/96a/2d630-1fe6-11ea-8fe3-002590496000/config'                                            
install(ok): /home/adina/repos/human-connectome-project-openaccess/HCP1200/996782/unprocessed (dataset)                               
action summary:                                                                                                                       
  get (notneeded: 1)
  install (ok: 3)
datalad get -n -r HCP1200/996782  5.83s user 3.39s system 32% cpu 27.943 total
(handbook2) adina@muninn in ~/repos/human-connectome-project-openaccess on git:master
❱ datalad remove -d HCP1200/996782
uninstall(ok): MNINonLinear (dataset)
uninstall(ok): T1w (dataset)
uninstall(ok): unprocessed (dataset)
uninstall(ok): . (dataset)
action summary:
  uninstall (ok: 4)
(handbook2) adina@muninn in ~/repos/human-connectome-project-openaccess on git:master!
❱ git st
On branch master
Your branch is ahead of 'origin/master' by 1 commit.

  (use "git add/rm <file>..." to update what will be committed)
	deleted:    HCP1200/996782
```

datalad/local/remove.py
mih: Not sure. My view was that removing CWD would be extremely unexpected behavior -- hence confirming that it won't happen is not needed. Same rational as for UNIX `rm`. It is also not in the docs that it won't remove CWD, but still wouldn't.

datalad/local/remove.py
mih: > `datalad remove -d HCP1200/996782`

You are not removing *from* the superdataset, hence it stays unsaved. Without `-d`:

```sh
(datalad-dev) mih@meiner /tmp/hcp (git)-[master] % datalad remove HCP1200/996782
uninstall(ok): HCP1200/996782/MNINonLinear (dataset)
uninstall(ok): HCP1200/996782/T1w (dataset)
uninstall(ok): HCP1200/996782/unprocessed (dataset)
uninstall(ok): HCP1200/996782 (dataset)
remove(ok): HCP1200/996782 (file)                                                                                           
save(ok): . (dataset)                                                                                                       
action summary:                                                                                                             
  remove (ok: 1)
  save (ok: 1)
  uninstall (ok: 4)
(datalad-dev) mih@meiner /tmp/hcp (git)-[master] % git st
On branch master
Your branch is ahead of 'origin/master' by 1 commit.

```

datalad/local/remove.py
mih: Makes sense. Thx!

datalad/local/remove.py
adswa: ehhh right :facepalm:

datalad/local/remove.py
adswa: ok

mih:
Good FS:

```
% python -m nose -s -v datalad.local.tests.test_remove datalad.distributed.tests.test_drop datalad.distribution.tests.test_uninstall  --with-coverage --cover-erase --cover-package datalad.local.remove --cover-package datalad.distributed.drop
Name                          Stmts   Miss  Cover   Missing
-----------------------------------------------------------
datalad/distributed/drop.py     209      8    96%   189-190, 324-331, 403, 562, 607, 716
datalad/local/remove.py          56      0   100%
-----------------------------------------------------------
TOTAL                           265      8    97%
Ran 30 tests in 109.508s

OK
```

Crippled FS (hence by-passing the often inappropriate `if-on-windows` conditionals:

```
FAIL: datalad.distribution.tests.test_uninstall.test_remove_file_handle_only
  File "/home/mih/env/datalad-dev/lib/python3.9/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/mih/hacking/datalad/git/datalad/tests/utils.py", line 558, in _wrap_with_tree
    return t(*(arg + (d,)), **kw)
  File "/home/mih/hacking/datalad/git/datalad/distribution/tests/test_uninstall.py", line 292, in test_remove_file_handle_only
    eq_(rpath_one, (ds.pathobj / 'two').resolve())
AssertionError: PosixPath('/media/mih/Samsung_T5/tmp/datalad_temp_tree_test_remove_file_handle_onlyonew6nnh/one') != PosixPath('/media/mih/Samsung_T5/tmp/datalad_temp_tree_test_remove_file_handle_onlyonew6nnh/two')

FAIL: datalad.distribution.tests.test_uninstall.test_uninstall_multiple_paths
  File "/home/mih/env/datalad-dev/lib/python3.9/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/mih/hacking/datalad/git/datalad/tests/utils.py", line 558, in _wrap_with_tree
    return t(*(arg + (d,)), **kw)
  File "/home/mih/hacking/datalad/git/datalad/distribution/tests/test_uninstall.py", line 248, in test_uninstall_multiple_paths
    ok_(all([f.endswith('keep') for f in files_left if exists(f) and not isdir(f)]))
AssertionError: None

FAIL: datalad.distribution.tests.test_uninstall.test_uninstall_recursive
  File "/home/mih/env/datalad-dev/lib/python3.9/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/mih/hacking/datalad/git/datalad/tests/utils.py", line 558, in _wrap_with_tree
    return t(*(arg + (d,)), **kw)
  File "/home/mih/hacking/datalad/git/datalad/distribution/tests/test_uninstall.py", line 339, in test_uninstall_recursive
    ok_(not exists(lname))
AssertionError: None

Name                          Stmts   Miss  Cover   Missing
-----------------------------------------------------------
datalad/distributed/drop.py     209      7    97%   189-190, 324-331, 403, 562, 716
datalad/local/remove.py          56      0   100%
-----------------------------------------------------------
TOTAL                           265      7    97%
Ran 30 tests in 317.609s

FAILED (failures=3)
```

3x slower, some test assumptions broken.

mih:
Consolidated the tests and ported for cross-platform compatibility. Now approximately same speed and identical near-complete coverage:

Good FS
```
Name                          Stmts   Miss  Cover   Missing
-----------------------------------------------------------
datalad/distributed/drop.py     209      5    98%   197, 324-331, 403, 607
datalad/local/remove.py          56      1    98%   142
-----------------------------------------------------------
TOTAL                           265      6    98%
Ran 22 tests in 56.195s

OK
```

Crippled FS:

```
Name                          Stmts   Miss  Cover   Missing
-----------------------------------------------------------
datalad/distributed/drop.py     209      5    98%   197, 324-331, 403, 607
datalad/local/remove.py          56      1    98%   142
-----------------------------------------------------------
TOTAL                           265      6    98%
Ran 22 tests in 77.139s

OK
```

bpoldrack:
Not a huge issue, since it's not destroying data, but I wonder whether the availability check for revisions is ideal.
Example: I have a dataset w/ a subdataset. I clone that thing and in the clone I remove `origin` from the subdataset. If I now want to drop that subds from the clone, `drop` is complaining about no known remote. On the one hand, that's technically right, since that dataset doesn't know about a remote. On the other hand: If I drop it with `--reckless availability`, I can `get` it again w/o issues. At some level we know that it's fine. Looks to me as if the check is running on the knowledge of the to-be-dropped dataset only, while the reverse operation (`get`) would use the info as available from the superdataset.

Don't immediately see a better solution, but someone else might.

```
(datalad-dev) ben@tree in /tmp/test_drop_clone/subds on git:master
❱ git remote remove origin
(datalad-dev) ben@tree in /tmp/test_drop_clone/subds on git:master
❱ cd ..
(datalad-dev) ben@tree in /tmp/test_drop_clone on git:master
❱ datalad -f json_pp drop -d. subds --what datasets                                                                                                                                                            1 !
{
  "action": "uninstall",
  "message": [
    "to-be-dropped dataset has revisions that are not available at any known sibling. Use `datalad push --to ...` to push these before dropping the local dataset, or ignore via `--reckless availability`. Unique revisions: %s",
    [
      "master"
    ]
  ],
  "path": "/tmp/test_drop_clone/subds",
  "refds": "/tmp/test_drop_clone",
  "status": "error",
  "type": "dataset"
}
(datalad-dev) ben@tree in /tmp/test_drop_clone on git:master
❱ datalad -f json_pp drop -d. subds --what datasets --reckless availability                                                                                                                                    1 !
{
  "action": "uninstall",
  "path": "/tmp/test_drop_clone/subds",
  "refds": "/tmp/test_drop_clone",
  "status": "ok",
  "type": "dataset"
}
(datalad-dev) ben@tree in /tmp/test_drop_clone on git:master
❱ datalad get subds
install(ok): /tmp/test_drop_clone/subds (dataset) [Installed subdataset in order to get /tmp/test_drop_clone/subds]
get(ok): subds/another (file) [from origin...]                                                                                                                                                                     
action summary:                                                                                                                                                                                                    
  get (ok: 1)
  install (ok: 1)
```

bpoldrack:
Dropping a regular subdirectory w/ `--what datasets` returns absolutely nothing. No `notneeded`, no error - just silence.

```
❱ datalad -f json_pp drop somedir --what datasets
```

bpoldrack:
Similar: Dropping an already dropped subds, doesn't return anything either. Shoudn't that be a `notneeded`?

datalad/distributed/drop.py
```diff
@@ -0,0 +1,804 @@
...
+            With 'modification', unsaved modifications in a dataset will not be
+            detected. This improves performance at the cost of permitting
+            potential loss of unsaved or untracked dataset components.
+            With 'availability', detection of dataset/branch-states that are
+            only available in the local dataset, and detection of an
+            insufficient number of file-content copies will be disabled.
+            Especially the latter is a potentially expensive check which might
+            involve numerous network transactions.
+            With 'undead', detection whether a to-be-removed local annex is
+            still known to exists in the network of dataset-clones is disabled.
```
loj, line 131: A few small typos in this sentence:

"With 'undead', detection of whether a to-be-removed local annex is still known to exist in the network of dataset-clones is disabled."

add "of" and change "exists" -> "exist"

mih:
> Dropping a regular subdirectory w/ `--what datasets` returns absolutely nothing. No `notneeded`, no error - just silence.

Doing this for this simple case is easy, for the general case not. How would you determine that for any given input path whether it had absolutely no consequences -- keeping in mind:

- recursive operation
- multiple path arguments
- multiple path arguments pointing to different datasets

The desire for this type of reporting originally led to `AnnotatePaths`. We should keep in mind that `notneeded` results are inconsequential in general, and typically not even displayed. I doubt that it is worth complicating the code for it, or even invest runtime to determine whether this is happinging.

Same here ...

> Similar: Dropping an already dropped subds, doesn't return anything either. Shoudn't that be a notneeded?

How much time should we invest? An empty dir could be an uninstalled subdataset, or an empty dir. A non-existing path could be an uninstalled subdataset, or not exist at all. Do we report notneeded for first-level uninstalled datasets? But why not for next-level? -- they also do not exist. So generating these inconsequential results always require inspection, slow normal operations, and complicate the code, because we need to track numerous combinations when a report qualifies as "having said something already". The current implementation reports notneeded whenever it is cheap, or when required for backward compatibility. But it tries hard to avoid runtime costs for beautification.

> Example: I have a dataset w/ a subdataset. I clone that thing and in the clone I remove origin from the subdataset. If I now want to drop that subds from the clone, drop is complaining about no known remote. On the one hand, that's technically right, since that dataset doesn't know about a remote. On the other hand: If I drop it with --reckless availability, I can get it again w/o issues. At some level we know that it's fine. Looks to me as if the check is running on the knowledge of the to-be-dropped dataset only, while the reverse operation (get) would use the info as available from the superdataset.

What are you proposing? That on-drop the superdataset is inspected whether it can generate any candidate URL from which a repository can be cloned that is then inspected whether it has all the commits needed by the to-be-dropped subdataset?

loj:
This feels good! I tried it out for a couple of different workflows/situations that I commonly used drop/uninstall/remove, and I have no complaints. :-)

I just found the small typo in the docs for `--reckless undead`.

datalad/distributed/drop.py
mih: Fixed locally. Will push when nothing else is reported. Thx!

bpoldrack:
> How would you determine for any given input path whether it had absolutely no consequences

I might be wrong, but shouldn't such a case be discoverable right after determining `paths_by_ds`? A given path pointing to a regular subdir should be associated with `dataset`, rather than generate its own entry, right? So, `_drop_dataset` should be able to tell. In fact, I think right here: 

https://github.com/datalad/datalad/pull/6111/files#diff-f0ac845bc6391c474eb58b0c1448b7f6bc00cb5585dab58777d8d9512f99c06bR371

(link doesn't work properly. Line 371 in that diff's `drop.py`)

Comment right above that `return` seems to realize that basically. We have a path of which we know that ony possibly keys can be dropped, but `what` was `datasets`, so we don't do anything. Fine. Why not make that a `notneeded` result with respective message?

datalad/distributed/drop.py
```diff
@@ -0,0 +1,804 @@
...
+    if had_fatality:
+        return
+
+    # next check must come AFTER the modification checks above, otherwise
+    # remove() could not rely on the modification detection above
+    if paths is not None and paths != [ds.pathobj] and what == 'datasets':
+        # so we have paths constraints that prevent dropping the full dataset
+        # there is nothing to do here, but to drop keys, which we must not
+        # done
+        return
```
bpoldrack, line 371: @mih

Something along these lines is what I mean. When do we end up here? AFAICS, when we have path(s) that are part of THIS dataset (as opposed to a subdataset) and we have `what='datasets'`. So, phrasing it like the comment ("constraining the drop of THIS dataset") is one way, but "trying to drop subdataset(s) where there are none" is another.

May be I just don't see how we can end up here in a way that this message would be misleading. I don't know.

```suggestion
        for p in paths:
            yield dict(action="drop",
                       path=p,
                       status="impossible",
                       message=f"what=datasets was given, but no dataset can be dropped from {p}",
                       type="file")
        return
```

mih: I am afraid this is not correct. It is perfectly possible that the paths you are reporting on have been originally used to drop datasets underneath it. `impossible` would be wrong, because something might have been done already, and `file` is a guess.

bpoldrack:
FTR: I'm testing my suggestion.

As for this:

> What are you proposing? That on-drop the superdataset is inspected whether it can generate any candidate URL from which a repository can be cloned that is then inspected whether it has all the commits needed by the to-be-dropped subdataset?

No - too expensive, obviously. As I said: No immediate idea. Just stumbled upon it. If you don't see a solution either - it's fine with me as is. I agree it's not worth making things more expensive.

> We should keep in mind that notneeded results are inconsequential in general, and typically not even displayed. I doubt that it is worth complicating the code for it, or even invest runtime to determine whether this is happinging.

Agree with the latter. Wouldn't phrase the former exactly like that. Via python interface I'd consider `notneeded` results much nicer to deal with than having to account for results sometimes, exceptions in other cases and absolutely nothing as yet another way of how datalad commands can behave. So - not worth paying a significant cost, but worth giving a second thought from my POV.

mih:
Conclusion from chat: Let's put the exploration and improvements possibly coming from @bpoldrack's comments in a separate PR. This one is already huge, and the suggested behavior was also not exhibited by the implementation that is being replaced it.

datalad/distributed/drop.py
bpoldrack: > It is perfectly possible that the paths you are reporting on have been originally used to drop datasets underneath it.

Hm. I fail to produce that, but thanks for the hint. Trying to figure that out.

PR #6112, opened 2021-10-25
UX: If a clean repo is dirty after a failed run, give clean-up hints

### LINKED ISSUE #2924: Give hints on how to revert/cleanup whenever "run" fails
#### What is the problem?
Follow up to https://github.com/datalad/datalad/pull/2922#issuecomment-429996043

If `run` fails, repo could be left in a dirty state with some new  and/or changed (unlocked) files.  User would need to figure out not only how to fix his command to run, but also how to revert back to the clean state.  So it would be nice to give either specific instruction or hints ("Use `git reset --hard && git clean -dfx` to get to the clean state before running the command again")

adswa:
This is an implementation of #2924.
In case where a run command failed in a formerly clean repo, and the repo is
left in a dirty state, log general hints to use 'git clean' and 'git reset'
at INFO level.
Closes #2924.

### DIFF
--- a/datalad/core/local/run.py
+++ b/datalad/core/local/run.py
@@ -807,8 +807,9 @@ def run_command(cmd, dataset=None, inputs=None, outputs=None, expand=None,
             if repo.dirty and not explicit:
                 # Give clean-up hints if a formerly clean repo is left dirty
                 lgr.info(
-                    "The commands 'git reset --hard' and/or 'git clean -df' "
-                    "could be used to get to a clean dataset state again."
+                    "The commands 'git clean -di' and/or 'git reset' "
+                    "could be used to get to a clean dataset state again. "
+                    "Consult their man pages for more information."
                 )
         raise exc
     elif do_save:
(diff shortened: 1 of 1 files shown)


adswa:
I have reworded the message to be a bit more careful

adswa:
any opinions on this? We can vote:
- :-1: close this PR altogether
- :+1: merge with the current message
- :confused: close this PR but make a note in #6114 to include a similar warning in an overhaul of how we present hints
- :eyes: reword the current warning further (please include a suggestion)

bpoldrack:
> "The commands 'git reset --hard' and/or 'git clean -dfi' could be used to get to a clean dataset state again. Consult their man pages for more information."

Generally fine with that, but if we are to suggest concrete command options for the "usual/general" case, make it either `git clean -dffi` (two `f`) or `git clean -di`, I think. That's because the second `f` makes `git clean` consider currently untracked subdatasets for removal (but still ask b/c `-i`). Hence, either make a suggestion that allows to clean up everything or a suggestion that is particularly careful about not removing too much (no `f` - AFAIK there's no point in a single  `f` if `i` is given).

So, my vote is on `git clean -dffi`, since creation of (to-be-added) subdatasets within a `run` call doesn't seem particularly outlandish to me.

mih:
Waiting for @bpoldrack to put conclusion into https://github.com/datalad/datalad/issues/6114, such that compliance of this PR can be evaluated

bpoldrack:
FTR: Wrote down the idea in #6114, but I don't think it makes sense to wait for an actual implementation.

So, apart from my preference for minimal rewording, can be merged from my POV.

PR #6113, opened 2021-10-25
Stop using distutils

jwodder:
[Distutils was officially deprecated in Python 3.10 and will be removed from the standard library completely in 3.12.](https://www.python.org/dev/peps/pep-0632/)  This PR thus eliminates almost all of the uses of distutils from the code.

Note that I had to leave in some uses of `LooseVersion` because I wasn't sure whether the stricter `packaging.version.Version` would accomodate all of its uses.  If we're only using `LooseVersion` for versions of the form `N(.N)*`, then `Version` will work out fine, but anything more exotic is not guaranteed to work.  If we need support for more exotic versions, we may have to roll our own version class.

### DIFF
--- a/datalad/distribution/create_sibling.py
+++ b/datalad/distribution/create_sibling.py
@@ -11,6 +11,7 @@
 
 __docformat__ = 'restructuredtext'
 
+from distutils.version import LooseVersion
 import logging
 import os
 from os.path import (
@@ -19,8 +20,6 @@ from os.path import (
     normpath,
     relpath,
 )
-
-from packaging.version import Version
 
 from datalad import ssh_manager
 
@@ -350,7 +349,7 @@ def _create_dataset_sibling(
 
     # check git version on remote end
     lgr.info("Adjusting remote git configuration")
-    if shell.get_git_version() and Version(shell.get_git_version()) >= Version("2.4"):
+    if shell.get_git_version() and shell.get_git_version() >= LooseVersion("2.4"):
         # allow for pushing to checked out branch
         try:
             shell("git -C {} config receive.denyCurrentBranch updateInstead".format(


_datalad_build_support/setup.py
```diff
@@ -13,13 +13,11 @@
...
 
 
-from distutils.core import Command
-from distutils.errors import DistutilsOptionError
-from distutils.version import LooseVersion
 from genericpath import exists
 from os import linesep, makedirs
 from os.path import dirname, join as opj, sep as pathsep, splitext
-from setuptools import findall, find_packages, setup
+from setuptools import Command, DistutilsOptionError, findall, find_packages, setup
```
mih, line 19: Please follow the format yielded by `python -m isort -m3 -fgw 2 -tc`, apart from other aspects, one import per line.

datalad/cmdline/tests/test_main.py
```diff
@@ -211,8 +211,8 @@ def test_script_shims():
             with open(which.rstrip()) as f:
                 content = f.read()
         else:
-            from distutils.spawn import find_executable
-            content = find_executable(script)
+            from shutil import which
+            content = which(script)
```
mih, line 215: Cool!, I was not aware that `shutil` provides `which` on windows!

What would be the advantage of having non-windows be a special case now? `which` is not even a POSIX command, so using the `shutil` implementation should work more robustly in general.

mih (commented):
This looks all sane to me. Thanks! I left a few comment which are all relatively minor.

_datalad_build_support/setup.py
jwodder: Done.

datalad/cmdline/tests/test_main.py
jwodder: I was only looking at the `distutils` import when I made the change, not at the surrounding code.  I've added a commit to use `shutil.which()` on all platforms.

mih: Cool, thx!

mih:
Test failure seems legitimately caused by these changes.

jwodder:
@mih I've pushed a fix.

mih:
OK, this seems ready. I have kicked off the errored test again, but I expect no failure. The only thing holding me back from merging is the draft state of this PR. @jwodder is this intentionally not meant to be merged?

jwodder:
@mih It should be merged at some point.  The only thing this PR is currently lacking is an alternative to the use of `distutils.version.LooseVersion`, though it could be merged as-is right now and a separate issue created for that.

mih:
@jwodder I would prefer if we get this done here, and leave the `LooseVersion` issue for later. Right now it seems there is no strict requirement to keep it in -core, but maybe I am missing some case.

jwodder:
@mih

> I would prefer if we get this done here, and leave the `LooseVersion` issue for later.

Then feel free to merge.

> Right now it seems there is no strict requirement to keep it in -core, but maybe I am missing some case.

I don't know what you're referring to here.

mih:
Thanks! 

I was referring to keeping `LooseVersion` or a replacement in this package (vs an extension, for example).

Issue #5126, opened 2020-11-04
Need to reenable some (32) tests on windows


yarikoptic:
I have forced `@skip_if_windows` to not skip on windows and then ran with (no ssh testing yet) on maint:

```
DATALAD_TESTS_KNOWNFAILURES_SKIP=0 DATALAD_TESTS_KNOWNFAILURES_PROBE=1 python -m nose -s -v --traverse-namespace datalad 2>&1 | tee /tmp/nose-maint-all-probe.log

```
with the result file uploaded to http://www.onerussian.com/tmp/nose-maint-all-probe.log . Among all the fails and errors of
```
FAILED (SKIP=132, errors=20, failures=63)
```
among which there are 32 which seems no longer be failing on windows for me locally.  Sure thing appveyor might think differently ;-) but we need to first bring it up to current state of things (#5125)

```shell
$> grep 'AssertionError: Exception not raised by test_' nose-maint-all-probe.log | nl
     1	AssertionError: Exception not raised by test_clone_dataladri
     2	AssertionError: Exception not raised by test_inputs_quotes_needed
     3	AssertionError: Exception not raised by test_py2_unicode_command
     4	AssertionError: Exception not raised by test_subdatasets
     5	AssertionError: Exception not raised by test_get_recurse_subdatasets
     6	AssertionError: Exception not raised by test_failed_install_multiple
     7	AssertionError: Exception not raised by test_install_dataset_from_just_source
     8	AssertionError: Exception not raised by test_install_into_dataset
     9	AssertionError: Exception not raised by test_install_known_subdataset
    10	AssertionError: Exception not raised by test_install_recursive
    11	AssertionError: Exception not raised by test_install_simple_local
    12	AssertionError: Exception not raised by test_uninstall_subdataset
    13	AssertionError: Exception not raised by test_reobtain_data
    14	AssertionError: Exception not raised by test_add_archive_content_absolute_path
    15	AssertionError: Exception not raised by test_rerun_ambiguous_revision_file
    16	AssertionError: Exception not raised by test_rerun_cherry_pick
    17	AssertionError: Exception not raised by test_rerun_just_one_commit
    18	AssertionError: Exception not raised by test_rerun_exclude_side
    19	AssertionError: Exception not raised by test_rerun_fastforwardable
    20	AssertionError: Exception not raised by test_rerun_fastforwardable_mutator
    21	AssertionError: Exception not raised by test_rerun_left_right_runs
[...]

PR #6120, opened 2021-10-29
TST: Enable all disabled tests listed in  #5126 and more

adswa:
#5126 had a list of to-be-reenabled-under-Windows tests. Since the CI's seem to be idle atm, I've re-enabled them to see which of them still fail.


mih (approved):
Wonderful! Thx much!

Issue #6085, opened 2021-10-15
Witless-based BatchedCommand (but generally generator-style command execution)


mih:
We need this dearly. It is the primary reason for datalad's slowness (especially on big datasets). Basically what we need is a way to execute a command, and start yielding outputs as quickly as possible, *not* waiting for the command to finish. When we have that, we also have batched command -- without the need for a dedicated implementation, because we also want to start yielding as quickly as possible for non-batched commands.

My under-performing brain thinks:

- if we want to start acting on outputs, while we are still in a `run()` method, we need multithreading (at least)
- we need to receive process output in the thread, turn it into a desirable structure (maybe JSON decode), and hand it over to the main thread for acting on it while more process output is received
- it seems https://docs.python.org/3/library/queue.html is the standard exchange paradigm for this stuff

It would be ideal, if we could wrap this entire construct into something that acts as an iterable, is parameterized with a runner instance, and exposes the signature of `Runner.run()`

adswa:
Has #6244 fixed this issue?

christian-monch:
> Has https://github.com/datalad/datalad/pull/6244 fixed this issue?

I would say it did, closing this issue

PR #6125, opened 2021-10-29
Add a generator style run() usage with example GeneratorAnnexJsonProtocol

christian-monch:
This PR provides:

- [x] Allows using a queue.Queue() as stdin (in addition to previous types), for convenient  stdin writing.
- [x] Generator style run() semantics if the provided protocol is derived from `GeneratorMixIn`-class.
- [x] Timeout capabilities
- [x] Define a GeneratorAnnexJsonProtocol
- [x] Add basic protocol tests

This PR provided some more ground work for issue #6085 and demonstrates the use of the `GeneratorProtocolMixIn`

Note: this PR includes PR #6124 (because it was forked from the branch)


datalad/runner/nonasyncrunner.py
```diff
@@ -10,251 +10,369 @@
...
     Dict,
+    IO,
     List,
     Optional,
     Type,
     Union,
 )
 
-from .protocol import WitlessProtocol
+from datalad.runner.protocol import (
```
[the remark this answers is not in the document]
christian-monch: changed to relative import

datalad/runner/nonasyncrunner.py
```diff
@@ -10,251 +10,369 @@
...
+        self.return_code = None
+        self.state = self.GeneratorState.process_running
+
+    def send(self, _):
+        if self.state == self.GeneratorState.process_running:
+
+            # if the result queue is empty and no more reading/writing
+            # threads are active, progress to next state
+            if len(self.result_queue) == 0 and not self.runner.active_file_numbers:
+                self.runner.process.wait()
```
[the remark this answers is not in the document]
christian-monch: Test added

datalad/runner/nonasyncrunner.py
```diff
@@ -10,251 +10,369 @@
...
+                self.stdin_queue = Queue()
+                self.stdin_queue.put(self.stdin)
+                self.stdin_queue.put(None)
+            else:
+                self.stdin_queue = self.stdin
+        else:
+            # indicate that we will not write anything to stdin, that
+            # means the user can pass None, or he can pass a
+            # file-like and write to it from a different thread.
+            lgr.warning(f"unknown instance class: {type(self.stdin)}, "
```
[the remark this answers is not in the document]
christian-monch: Added suggestion

datalad/runner/nonasyncrunner.py
```diff
@@ -10,251 +10,369 @@
...
+    def process_queue(self):
+        """
+        Get a single event from the queue
+        """
+        data = None
+        while True:
+            try:
+                file_number, state, data = self.output_queue.get(timeout=self.timeout)
+            except Empty:
+                lgr.warning(f"TIMEOUT on output queue")
```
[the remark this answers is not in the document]
christian-monch: Logging is removed. Timeouts will lead to timeout callbacks (except from the internal progress-timeout), which appear according to the specified timeout interval.

datalad/runner/nonasyncrunner.py
```diff
@@ -10,251 +10,369 @@
...
-        # and closes stdin at the end
-        self.process._stdin_write(self.stdin_data)
+            if state == IOState.ok:
+                break
 
-        lgr.log(5, "%s exiting (write completed or interrupted)", self)
-        self.queue.put((self.stdin_fileno, None, time.time()))
+            # Handle timeouts
+            if state == IOState.timeout:
+                lgr.warning(f"TIMEOUT on {self.fileno_mapping[file_number]}")
```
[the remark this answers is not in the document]
christian-monch: Modified according to the above timeout issue  (no logging, protocol timeout-handlers called)

datalad/runner/nonasyncrunner.py
```diff
@@ -10,251 +10,369 @@
...
+                self.protocol.pipe_data_received(self.fileno_mapping[file_number], data)
 
 
 def run_command(cmd: Union[str, List],
                 protocol: Type[WitlessProtocol],
                 stdin: Any,
                 protocol_kwargs: Optional[Dict] = None,
+                timeout: Optional[float] = None,
                 **kwargs) -> Any:
-    """
```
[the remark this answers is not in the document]
christian-monch: Docstrings added

datalad/runner/protocol.py
```diff
@@ -20,6 +21,17 @@
 lgr = logging.getLogger('datalad.runner.protocol')
 
 
+class GeneratorMixIn:
+
```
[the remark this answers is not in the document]
christian-monch: Docstring added

datalad/runner/tests/test_nonasyncrunner.py
```diff
@@ -122,32 +119,35 @@ def pipe_data_received(self, fd, data):
...
-    sleep(5)
-    assert_true(reader_thread.is_alive())
-
-    # Check actual exit
-    os.write(write_descriptor, b"more data")
-    reader_thread.join()
-    data = read_queue.get()
-    eq_(data[1], b"more data")
-    assert_true(read_queue.empty())
+# def test_blocking_thread_exit():
```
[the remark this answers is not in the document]
christian-monch: Test is back

datalad/runner/nonasyncrunner.py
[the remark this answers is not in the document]
christian-monch: Working on full coverage

christian-monch:
@yarikoptic : considered all your suggestions. Still working on 100 % test coverage, besides that, the PR is more or less ready.

mih:
Restarted MacOS tests

docs/source/design/threaded_runner.rst
```diff
@@ -0,0 +1,68 @@
...
+.. _chap_threaded_runner:
+
+
+****************
+Threaded runner
+****************
+
+.. topic:: Specification scope and status
+
+   This specification provides an overview over the current implementation.
```
mih, line 13: YES! Thank you! This makes all the difference!

docs/source/design/threaded_runner.rst
```diff
@@ -0,0 +1,68 @@
...
+
+Threads
+=======
+
+Datalad often requires the execution of subprocesses. While subprocesses are executed, datalad, i.e. its main thread, should be able to read data from stdout and stderr of the subprocess as well as write data to stdin of the subprocess. This requires a way to efficiently multiplex reading from stdout and stderr of the subprocess as well as writing to stdin of the subprocess.
+
+Since non-blocking IO and waiting on multiple sources (poll or select) differs wastly in terms of capabilities and API on different OSs, we decided to use blocking IO and threads to multiplex reading from different sources.
```
mih, line 20:
```suggestion
Since non-blocking IO and waiting on multiple sources (poll or select) differs vastly in terms of capabilities and API on different OSs, we decided to use blocking IO and threads to multiplex reading from different sources.
```

docs/source/design/threaded_runner.rst
```diff
@@ -0,0 +1,68 @@
...
+
+The method `WitlessProtocol.timeout(fd)` is also called if all of stdout, stderr and stdin are closed and the process does not exit within the given interval. In this case `fd` is set to `None`. If `WitlessProtocol.timeout(fd)` returns `True` the process is terminated.
+
+
+Object and Generator Results
+================================
+
+If the protocol that is provided to `run()` does not inherit `datalad.runner.protocol.GeneratorMixIn`, the final result that will be returned to the caller is determined by calling `WitlessProtocol._prepare_result()`. Whatever object this method returns will be returned to the caller.
+
+If the protocol that is provided to `run()` does inherit `datalad.runner.protocol.GeneratorMixIn`, `run()` will return a `Generator`. This generator will yield the elements that sent to it in the protocol-implementation by calling `GeneratorMixIn.send_result()` in the order in which the method `GeneratorMixIn.send_result()` is called. For example, if `GeneratorMixIn.send_result(43)` is called, the generator will yield `43`, and if `GeneratorMixIn.send_result({"a": 123, "b": "some data"})` is called, the generator will yield `{"a": 123, "b": "some data"}`.
```
mih, line 66:
```suggestion
If the protocol that is provided to `run()` does inherit `datalad.runner.protocol.GeneratorMixIn`, `run()` will return a `Generator`. This generator will yield the elements that were sent to it in the protocol-implementation by calling `GeneratorMixIn.send_result()` in the order in which the method `GeneratorMixIn.send_result()` is called. For example, if `GeneratorMixIn.send_result(43)` is called, the generator will yield `43`, and if `GeneratorMixIn.send_result({"a": 123, "b": "some data"})` is called, the generator will yield `{"a": 123, "b": "some data"}`.
```

mih:
So if I read this correctly, the implementation is complete, but it is not actually used outside the test. Is that right? If so, any showstoppers for doing so?

christian-monch:
> So if I read this correctly, the implementation is complete, but it is not actually used outside the test. Is that right? If so, any showstoppers for doing so?

That is correct. I am working on using it in AnnexRepo together with @bpoldrack in this branch: https://github.com/christian-monch/datalad/tree/nf_annex_repo_yielding. Should be a quick thing, but we are pairing up in order to increase (reduce?) the bus-factor.

.travis.yml
```diff
@@ -178,7 +178,7 @@ matrix:
 #    - UNSET_S3_SECRETS=1
 
 before_install:
-  - NOSE_OPTS=( )
+  - NOSE_OPTS=( -v )
```
yarikoptic, line 181: `git blame` or log might tell why we removed `-v` generally (may be was getting too long?)

datalad/runner/nonasyncrunner.py
```diff
@@ -35,226 +48,499 @@
...
+            the result created by the protocol method `_generate_result`.
+
+        stdin : file-like, string, bytes, Queue, or None
+            If stdin is a file-like, it will be directly used as stdin for the
+            subprocess. The caller is responsible for writing to it and closing
+            it. If stdin is a string or bytes, those will be fed to stdin of the
+            subprocess. If all data is written, stdin will be closed.
+            If stdin is a Queue, all elements (bytes) put into the Queue will
+            be passed to stdin until None is read from the queue. If None is
+            read, stdin of the subprocess is closed.
```
yarikoptic, line 169: and what happens if value is `None` - `stdin` is used? (docstring doesn't describe that)

datalad/runner/nonasyncrunner.py
```diff
@@ -35,226 +48,499 @@
...
+            of the runner. So you could write:
+
+               gen = runner.run()
+               for file_descriptor, data in gen:
+                   ...
+
+               # get the return code of the process
+               result = gen.return_code
+        """
+        if isinstance(self.stdin, (int, IO, type(None))):
```
yarikoptic, line 256: May be more straightforward and explicit

```suggestion
        if self.stdin is None or isinstance(self.stdin, (int, IO)):
```
?

datalad/runner/runnerthreads.py
```diff
@@ -0,0 +1,298 @@
...
+class BlockingOSReaderThread(ExitingThread):
+    """
+    A blocking OS file reader. If it reads
+    anything, it stores its data in a queue.
+    That allows a consumer to block
+    on OS-reads with timeout, independent from
+    the OS read capabilities.
+    It enqueues bytes if something is read. If
+    the file is close, it will enqueue None and
+    exit.
```
yarikoptic, line 56: should it rhyme if so narrow? ;)

datalad/runner/tests/test_nonasyncrunner.py
```diff
@@ -184,3 +330,204 @@ def test_popen_invocation(src_path, dest_path):
...
+    for i in range(72):
+        stdin_queue.put(b"\x00" * 1024)
+    stdin_queue.put(None)
+
+    timeout_queue = []
+    run_command(
+        py2cmd("import time; time.sleep(5)\n"),
+        stdin=stdin_queue,
+        protocol=TestProtocol,
+        timeout=1,
```
yarikoptic, line 384: could
```suggestion
        timeout=0.1,
```
for the same effect?

datalad/runner/tests/test_nonasyncrunner.py
```diff
@@ -184,3 +330,204 @@ def test_popen_invocation(src_path, dest_path):
...
```
yarikoptic, line 413: could timeout be reduced here as well?

yarikoptic (commented):
may be time outs could be reduced to not delay tests more than really needed.

left other minor comments.  Waiting anxiously for seeing the WiP on this code used in the code base .

christian-monch:
> Waiting anxiously for seeing the WiP on this code used in the codebase.
I have started to use it in the code, for example, to implement a generator-get in AnnexRepo (https://github.com/christian-monch/datalad/tree/nf_generator_get).

Another demo uses a generator version of `AnnexRepo._call_annex`, to demonstrate equality of the approaches (https://github.com/christian-monch/datalad/tree/nf_annex_repo_yielding) (one test is still failing). It uses a generator-based `_call_annex()` implementation and adapts to the "complete execution" style in `_call_annex_records`.

While the second demo has value in showing the use of the generator-based runner, it does however not add to the overall "agility" of the execution. By that I mean, that we do not use generators on all calling levels. Instead, we revert to the old "catch complete executions"-style at some point. In the generator-based `_call_annex` demo quite early, namely in `_call_annex_records`. Exposing the generator-style on the interface of `AnnexRepo` may require huge amounts of change throughout datalad. I am investigating this, and we should probably have a discussion on the topic with at least @yarikoptic, @mih, @bpoldrack. Looking forward to that.

datalad/support/annexrepo.py
```diff
@@ -3638,6 +3642,17 @@ def process_exited(self):
...
 
+class GeneratorAnnexJsonProtocol(GeneratorMixIn, AnnexJsonProtocol):
+    def __init__(self,
+                 done_future=None,
+                 total_nbytes=None):
+        GeneratorMixIn.__init__(self)
+        AnnexJsonProtocol.__init__(self, done_future, total_nbytes)
+
+    def add_to_output(self, json_object):
+        self.send_result(json_object)
```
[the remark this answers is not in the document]
christian-monch: You are right. I actually introduced `add_to_output` for the sole reason of overwriting it in `GeneratorAnnexJsonProtocol` in order to _demonstrate_ the minimal necessary changes that are required to create `GeneratorAnnexJsonProtocol` and accepted the "uncleanliness".

In the future `AnnexJsonProtocol` might just be a `GeneratorMixIn` protocol and there will be no more need for `GeneratorAnnexJsonProtocol`. So I would leave it as is for now. If we already see that we need both protocols, we should fix it though.

.travis.yml
christian-monch: Needed it to debug the remote tests. It is removed now.

datalad/runner/nonasyncrunner.py
christian-monch: Description added

datalad/runner/nonasyncrunner.py
christian-monch: If actually find the original version more straightforward, but I am not dogmatic about that. Feel free to commit.

datalad/runner/runnerthreads.py
christian-monch: You mean something like this: 

    It enqueues bytes if something's read.
    If source is closed, it ends the thread,
    not without enqueuing None,
    so you might see that it is done.

;)

datalad/runner/tests/test_generatormixin.py
```diff
@@ -0,0 +1,110 @@
...
+
+    # Expect process exited and connection lost to be called.
+    assert_equal(result_generator.send(None), 3)
+    assert_equal(result_generator.send(None), 4)
+    assert_raises(StopIteration, result_generator.send, None)
+
+
+def test_exiting_process():
+
+    result = run_command(py2cmd("import time\ntime.sleep(3)\nprint('exit')"),
```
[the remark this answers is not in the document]
christian-monch: No, a shorter sleep is sufficient. Changed to 0.1 seconds

datalad/runner/tests/test_nonasyncrunner.py
christian-monch: I changed `sleep(5)` in line 381 to `sleep(.6)` and added your change for line 384

datalad/runner/tests/test_nonasyncrunner.py
christian-monch: Reduced to `.6` and `.1`

datalad/runner/runnerthreads.py
mih: LOL! 😎

christian-monch:
W.r.t. performance regression:

I have a fix for that (get rid of the blocking OS-threads and add timeout-threads) in the branch fix-performance-generator-runner (https://github.com/christian-monch/datalad/tree/fix-performance-generator-runner). Did not want to add it to this PR in order to  have a chance to finish this one.

christian-monch:
> So if I read this correctly, the implementation is complete, but it is not actually used outside the test. Is that right? If so, any showstoppers for doing so

I created a drop-in replacement for BatchedCommand, which is based on generator-style Runner (https://github.com/christian-monch/datalad/tree/enh-generator-batched-command-step-1), all tests pass.

There is no PR for the new BatchedCommand yet because there is a step in between, i.e. a branch with performance improvements over this PR, on which the Generator-BatchedCommand branch is based. I would really like to first close this before creating new PRs and I don't want to extend this further, it is already huge.

mih:
@christian-monch I would need a little guidance on how to approach this. You would prefer to have this merged, but the actual usage is in another branch. What I can try out (not just look at), is not what will be merged. Moreover, you mention that the actual usage additionally requires a performance optimization that is also not in this PR.

From my POV a 2k line diff is already beyond what can be reasonably inspected line-by-line. Runtime testing and introspection is what I would do to evaluate the impact of such a change. However, when the PR does not actually contain the changes that must be there, it seems like an expensive intermediate step that adds little value.

What is the advantage of an "academic" review of this PR in isolation?

christian-monch:
> I would need a little guidance on how to approach this. You would prefer to have this merged, but the actual usage is in another branch. What I can try out (not just look at), is not what will be merged. Moreover, you mention that the actual usage additionally requires a performance optimization that is also not in this PR.
>
> From my POV a 2k line diff is already beyond what can be reasonably inspected line-by-line. Runtime testing and introspection is what I would do to evaluate the impact of such a change. However, when the PR does not actually contain the changes that must be there, it seems like an expensive intermediate step that adds little value.

I see your point. My intention was to allow interested parties to follow with the changes, but I can see, that this might be unrealistic.

Here is my suggestion: I will take what was learned in this PR and in the follow-up work and bundle it into a neat little PR (the interface of the classes introduced in this PR would not change in the new PR). The new PR would also contain the BatchedProcess implementation based on the ThreadedRunner, which reduces code duplication. We then abandon this PR and focus on the new PR. This should not take long, i.e. one to two days

christian-monch:
This PR has accumulated too many changes and is therefore closed. A smaller PR that supports the API described here will be opened soon.

PR #6127, opened 2021-10-29
BF(TST): remove reuse of the same tape across unrelated tests

yarikoptic:
originally that tape was added for test_install_crcns (hence the name of the tape)
which later was commented out in 941e51ffc8c506b51678e2fe3b2bf334ff3299ab and never
brought back.  Somehow that decorator was copy/pasted into other tests.

I guess because we no longer use GitPython, no online interactions through
python http modules were made, so tape was not really used, and it did not
lead to any failures.

My attention to that was brought by failures on Windows, such as
https://github.com/datalad/git-annex/runs/4042500337?check_suite_focus=true
which also should be unrelated, thus probably would not provide any
remedy, but there remains no reason to keep those "uses" of the same vcr
tape around.

### DIFF
--- a/datalad/core/distributed/tests/test_clone.py
+++ b/datalad/core/distributed/tests/test_clone.py
@@ -124,7 +124,6 @@ def test_invalid_args(path, otherpath, alienpath):
 
 @integration
 @skip_if_no_network
-@use_cassette('test_install_crcns')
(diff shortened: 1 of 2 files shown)


mih:
Thx! Restarted macos ci, should be unrelated by sounded too suspiciously close.

PR #6128, opened 2021-10-29
Replace AnnotatePaths in metadata commands

### LINKED ISSUE #3368: Retire `annotate_paths()`

mih:
This is the second-to-last stop to finally resolve #3368

There is only a minimal behavior change: A rather peculiar reporting
behavior, where metadata of the "nearest" available dataset is chosen
when no explicit dataset is given.

This has no practical value and would complicate this temporary fix
substantially. This whole stuff only needs to last until metalad is
fully functional.

### DIFF
--- a/datalad/interface/annotate_paths.py
+++ b/datalad/interface/annotate_paths.py
@@ -10,7 +10,7 @@
 
 __docformat__ = 'restructuredtext'
 
-from collections import defaultdict
+from pathlib import Path
 
 import logging
 import textwrap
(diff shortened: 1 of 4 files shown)


mih:
Thanks @bpoldrack 

Will merge tomorrow

PR #6132, opened 2021-11-02
TEMP: trying to meld Interface and ConfigManager

yarikoptic:


mih:
I believe the mood was that this will not make it in time for 0.16. Hence I am archiving it now.

PR #6105, opened 2021-10-19
RF: Transform add-archive-content into a Dataset method

adswa:
This is a very crude and work-in-progress refactoring effort for
add-archive-content. It could fix #2122, because it transforms 
``add-archive-content`` into a dataset method.
It doesn't yet deal with simplifying the code,
but only includes transitions to current dataset methods (e.g., proper
result rendering, a --dataset argument) and transitions away from things
that are so outdated or fragile that even I can spot them (e.g., use of
pathlib functionality instead of os.path functionality).

Wrt to the tests, the goal is to make the original tests pass with minimally invasive
changes. Thus, the current changes do not RF the test setup in general
but rather make it work as is. This includes transforming pure annex
repos into datasets, calling add-archive-content as a dataset method,
and switching from 'add' to 'save', etc. Many suboptimal aspects of this
test setup are not fixed by this. I hope I can RF them once they have
helped me determine whether my RFing of add-archive-content has broken
something that shouldn't be broken.

``add-archive-content`` has improved slightly with this refactoring:
- It has progress bars (fixes #4876)
- It checks also for existing but not enabled special remotes now (fixes #1693)
- It doesn't crash when ``--key`` is given but ``--use-current-dir`` is not (fixes #3995)
- 7 more Windows tests pass now

PR #1644, opened 2017-07-17
Better support for special remotes (webdav in particular)

mih:
- [x] `AnnexRepo.get_special_remotes()`
- [x] add support for enabling remotes when annex itself will not do it. One case is the webdav special remotes that requires user and password to be supplied in addition -- we could get then from the cred store and supply them: `AnnexRepo.enable_special_remote()`
- [x] probably worth supporting this low-level functionality through the `siblings` command -- come up with a good concept
- [x] check that auto-enabling a special remote during install doesn't fail the entire process if the password is mistyped (or a wrong one is given due to #1779)
 Related issues
- [x] fixes #1775 
- [x] fixes #1779 
- [x] fixes #1780 
- [x] new `siblings enable` subcommand can enable webdav special remotes, either with credential entry or automatic when credentials are already known
- [x] https://asciinema.org/a/136015

PR #6133, opened 2021-11-02
BF: Force git-annex to commit the changes before we interrogate git-annex branch

yarikoptic:
The use case was detected while working on
https://github.com/datalad/datalad-crawler/pull/112
to accompany
https://github.com/datalad/datalad/pull/6105 .

Due to use of alwayscommit = False to minimize number of initial commits,
the repository ended up in a "peculiar" state with no useful commits
in git-annex branch -- all changes were staged within git-annex to be
committed "eventually".

Although we could have easily "fixed" the test in datalad-crawler,
I consider the approach of going around git-annex commands to interrogate
git-annex metadata to be fragile, in particular for such a case as here:
some other command could have been ran in alwayscommit=false mode, and
information in git-annex branch might not yet reflect actual state of
git-annex.  Ideally we should avoid doing such "going around git-annex".
As a quick resolution I have decided to just trigger a possible commit
by running  git-annex info --fast.  But may be there is a better/cheaper
way to trigger git-annex to commit, @joeyh?

Questionable change is the expansion of call_annex interface.  Could be
avoided through monkey patching of self.always_commit for that call,
but that would be more ugly IMHO

FWIW
- I have looked at few other uses of `call_git` within annexrepo -- those seems to be ok, since do not stick their fingers into git-annex'es "privates".

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -783,6 +783,11 @@ class AnnexRepo(GitRepo, RepoInterface):
         argspec = re.compile(r'^([^=]*)=(.*)$')
         srs = {}
         try:
+            # git-annex might have some changes "staged" to be committed to
+            # git-annex branch.  Before interrogating git-annex's metadata without
+            # going through git annex, we need to possibly commit staged changes.
+            if not self.always_commit:
+                self.call_annex(['info', '--fast'], always_commit=True)
             for line in self.call_git_items_(
                     ['cat-file', 'blob', 'git-annex:remote.log'],
                     read_only=True):
@@ -813,6 +818,7 @@ class AnnexRepo(GitRepo, RepoInterface):
 
     def _call_annex(self, args, files=None, jobs=None, protocol=StdOutErrCapture,
                     git_options=None, stdin=None, merge_annex_branches=True,
+                    always_commit=None,
                     **kwargs):
         """Internal helper to run git-annex commands
 
@@ -847,6 +853,8 @@ class AnnexRepo(GitRepo, RepoInterface):
           If False, annex.merge-annex-branches=false config will be set for
           git-annex call.  Useful for operations which are not intended to
           benefit from updating information about remote git-annexes
+        always_commit: bool, optional
+          If specified, will take precedence over self.alwayscommit
         **kwargs:
           Additional arguments are passed on to the WitlessProtocol constructor
 
@@ -879,7 +887,9 @@ class AnnexRepo(GitRepo, RepoInterface):
         if git_options:
             cmd += git_options
 
-        if not self.always_commit:
+        if always_commit is None:
+            always_commit = self.always_commit
+        if not always_commit:
             cmd += ['-c', 'annex.alwayscommit=false']
 
         if not merge_annex_branches:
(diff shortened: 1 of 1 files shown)


mih (commented):
I feel a bit dizzy.

We had a repo property (set by `__init__`).

Now we have `call_annex()` method parameters to override this, for some methods, not all.

And we now unconditionally enable it for an additional call to `info --fast`, only to discard its output in `get_special_remotes()`.

This all seems to complement the approach taken in `addurls` to override the `always_commit` value (assuming that that implementation also wanted to have this repo behavior be configurable?).

This seems rather twisted.

If I understand the use case description correctly, the situation that this PR is trying to address is intentionally created by a user. Which makes me wonder why that same user cannot stop this non-committing mode, as soon as they seem to start require commits to exist?

joeyh:
Oh, I see you're reading git-annex:remote.log directly.

So, options are:

* Use git-annex merge or something like that to get the journal committed to
  the branch before reading from it. (A git-annex merge --only-annex
  that avoids merging the master branch would be ideal, IIRC I've
  considered adding that before.)
* Read .git/annex/journal/remote.log yourself, if it's present,
  and concat it to git-annex:remote.log to see the same thing git-annex
  will later commit.
* Add a new query command to git-annex to get whatever information you
  are reading from remote.log.

-- 
see shy jo

yarikoptic:
>  This seems rather twisted.

yeah, I also dislike it.

> the situation that this PR is trying to address is intentionally created by a user. Which makes me wonder why that same user cannot stop this non-committing mode, as soon as they seem to start require commits to exist?

every state of *Repo is created by a user via "standard" interfaces of *Repos.  Whenever some method is not providing correct return value due to how it is implemented (e.g. by digging through git-annex branch "manually" without awareness of `.git/annex/journal`, instead of using `git annex` commands to provide that information) -- it is IMHO a bug in the software implementation of those interfaces.

> * Read .git/annex/journal/remote.log yourself, if it's present, and concat it to git-annex:remote.log to see the same thing git-annex will later commit.

I guess if we are to keep digging through git-annex privates, this might be as a *gentle* approach indeed.  Some archaeological expedition is probably due to see why we aren't just calling out to `git annex info --json --fast` (yet to check if it would provide expected output even if ran with `-c alwayscommit=false`)

joeyh:
Your parsing of remote.log is also slightly buggy in that it does not consider
timestamps, so it may choose an old version of a remote's config rather
than the current version.

Also reading .git/annex/journal/ will miss configs of private remotes,
which are stored in a separate journal and never reach the git-annex
branch.

Extending git-annex info to get the information you need seems like the
best way to me, although we'd need to discuss just what that information
is and the best way to do it.

-- 
see shy jo

yarikoptic:
FWIW, use of `cat-file` to get a list of special remotes was introduced in 4a4b2245bbf75fb274acc0bec2a1c69dcc439087 as part of #1644 in 2017. `annex info` information is insufficient replacement since `get_special_remotes` returns record with all the parameters of the remote (encryption, type, etc).

@joeyh - is there another `git-annex` command which would show details of the special remotes git annex knows about?

joeyh:
There is no other command, but you can git-annex info about specific uuids of remotes and get information like "type" and "encryption".

One thing I don't want git-annex info uuid to do is dump out sensitive information like cipher= from remote.log.

yarikoptic:
> Read .git/annex/journal/remote.log yourself, if it's present, and concat it to git-annex:remote.log to see the same thing git-annex will later commit.

implemented in https://github.com/datalad/datalad/pull/6135 which replaces this PR

Issue #5835, opened 2021-07-29
`clone --reckless ephemeral` should set `annex.private=true`


mih:
See here for more info: https://git-annex.branchable.com/tips/cloning_a_repository_privately/

This was already mentioned in https://github.com/datalad/datalad/issues/4903#issuecomment-832444161 but not acted upon. Filing this dedicated issue to get it done.

This feature was implemented in git-annex back in April 2021 (bfa2db922281349cfc37264ee91e1c861266e208).

@bpoldrack Can you take this on?

mih:
Hey, is there any ETA for this enhancement yet? Should I take over?

bpoldrack:
@mih Next on my list after authenticated HTTP testing.

PR #6135, opened 2021-11-02
BF: obtain information about annex special remotes also from annex journal

yarikoptic:
An alternative to https://github.com/datalad/datalad/pull/6133 implementation
which first forces git-annex to commit.

The use case was detected while working on https://github.com/datalad/datalad-crawler/pull/112
to accompany https://github.com/datalad/datalad/pull/6105 .

Although we could have easily "fixed" the test in datalad-crawler, I consider
the approach of going around git-annex commands to interrogate git-annex
metadata to be fragile, in particular for such a case as here: some other
command could have been ran in alwayscommit=false mode, and information in
git-annex branch might not yet reflect actual state of git-annex.  Ideally we
should avoid doing such "going around git-annex", but ATM there is no git-annex
command which would "expose" all this information.

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -782,33 +782,51 @@ class AnnexRepo(GitRepo, RepoInterface):
         """
         argspec = re.compile(r'^([^=]*)=(.*)$')
         srs = {}
+
+        # We provide custom implementation to access this metadata since ATM
+        # no git-annex command exposes it on CLI.
+        #
+        # Information will potentially be obtained from remote.log within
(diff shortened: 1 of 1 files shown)


mih (commented):
This LGTM. Thx!

However, I want to remind about a comment from Joey

> Also reading .git/annex/journal/ will miss configs of private remotes, which are stored in a separate journal and never reach the git-annex branch.

This should be mentioned in the docs, at least. We are not using this yet, but soon "ephemeral clones" will be private annex repos, and then any special remote configured in them would still be invisible to the method. Ping #5835

PR #6136, opened 2021-11-03
TST: Enable more Windows tests

adswa:
A mix of newly enabled tests because they passed on an actual Windows box, and a few adjusted tests that pass after they got a Window equivalent to a Unix command. Let's see if the CIs agree.


mih (approved):
Awesome, thx!

Issue #4496, opened 2020-05-07
test_decompress_file: fails on NFS on master


yarikoptic:
in master (didn't spot in maint) test fails on our cron job: https://travis-ci.org/github/datalad/datalad/jobs/682067828

```
======================================================================
ERROR: datalad.tests.test_archives.test_decompress_file('strip',)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/datalad/tests/utils.py", line 518, in newfunc
    return t(*(arg + (d,)), **kw)
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/datalad/tests/test_archives.py", line 77, in check_decompress_file
    leading_directories=leading_directories)
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/datalad/support/archives.py", line 99, in decompress_file
    rmdir(widow_dir)
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/datalad/utils.py", line 471, in rmdir
    os.rmdir(path)
OSError: [Errno 39] Directory not empty: '/tmp/nfsmount/datalad-fhwhrw_1/simple-extracted/ "\';b&b&cΔЙקم๗あ `| '
```

adswa:
I fail to find this test in the logs. Do we run this test battery under NFS currently?

yarikoptic:
@adswa there are back references from other PRs/issues now, and 

```
(git)smaug:/mnt/datasets/datalad/ci/logs/2021[master]git
$> git submodule foreach git grep 'ERROR: datalad.tests.test_archives.test_decompress_file'
Entering '05'
26/pr/5698/d719809/travis-11973-failed/8.txt:ERROR: datalad.tests.test_archives.test_decompress_file('strip',)
Entering '06'
17/pr/5749/adc72ce/travis-12088-failed/8.txt:ERROR: datalad.tests.test_archives.test_decompress_file('strip',)
18/push/master/99bac03/travis-12090-failed/8.txt:ERROR: datalad.tests.test_archives.test_decompress_file('strip',)
29/pr/5749/19278eb/travis-12118-failed/8.txt:ERROR: datalad.tests.test_archives.test_decompress_file('strip',)
[...]

PR #6137, opened 2021-11-03
OPT: do not deduce the ds path over and over again for files in a dir

yarikoptic:
I had some concerns due about symlinks but it seems that get_dataset_root doesn't bother much about them, so I think it should be kosher -- let's see what CI says ;-)

edit: may be I should just add a locally decorated `lru_cache(get_dataset_root)` instead of "manual" try/excepts?

edit: on a trial attempt benefit seems to be minor but present:

```shell
(git-annex)lena:~/datalad/dandi/dandisets/000026[draft]rawdata/sub-I46/ses-SPIM
$> multitime -n 5 datalad status *  # has 1100 files in that directory
```

this branch first line, master - 2nd (both are warm runs)
```
real        0.825       0.018       0.792       0.834       0.842   
real        0.848       0.031       0.816       0.830       0.897
```

### DIFF
--- a/datalad/core/local/status.py
+++ b/datalad/core/local/status.py
@@ -532,22 +532,31 @@ def get_paths_by_ds(refds, dataset_arg, paths, subdsroot_mode='rsync'):
     # step 2: resolve
     # for later comparison, we need to preserve the original value too
     paths = [(resolve_path(p, dataset_arg), str(p)) for p in paths]
+    # OPT: store cache for dataset roots for each directory directly
+    #      listed in paths, or containing the path (if file)
+    roots_cache = {}
     # sort any path argument into the respective subdatasets
     # sort by comparing the resolved Path instances, this puts top-level
     # paths first, leading to their datasets to be injected into the result
     # dict first
     for p, orig_path in sorted(paths, key=lambda x: x[0]):
-        # TODO(OPT)? YOH does not spot any optimization for paths under the same
-        # directory: if not isdir(path) - files would all have the same
-        # "root", and we could avoid doing full `get_dataset_root` check for
-        # those. Moreover, if some path points UNDER that path which isdir, and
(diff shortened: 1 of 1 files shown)


yarikoptic:
"interesting" -- on appveyor seems to stall in different runs on different tests (collated below)
```
datalad.core.local.tests.test_diff.test_diff ... 
datalad.core.distributed.tests.test_clone.test_notclone_known_subdataset ... subdataset(ok): subm 1 (dataset)
subdataset(ok): subm 1 (dataset)
datalad.local.tests.test_copy_file.test_copy_file ... 
datalad.core.local.tests.test_create.test_create_dataset_same_as_path ... 
..
```

on travis -- fails one test
```
FAIL: datalad.tests.test_archives.test_decompress_file('strip',)
  File "/tmp/dl-miniconda-5hkyo5ef/lib/python3.9/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/tmp/dl-miniconda-5hkyo5ef/lib/python3.9/site-packages/datalad/tests/utils.py", line 558, in _wrap_with_tree
    return t(*(arg + (d,)), **kw)
  File "/tmp/dl-miniconda-5hkyo5ef/lib/python3.9/site-packages/datalad/tests/test_archives.py", line 94, in check_decompress_file
    assert_true(op.exists(op.join(testpath, '3.txt')))
AssertionError: False is not true
```
on NFS.  probably another manifestation of https://github.com/datalad/datalad/issues/4496 (@adswa asked if to be closed after #6064 which might have changed how it fails).

yarikoptic:
hm, on rerun it is all green (besides codecov but diff is fully green there). So in principle there is no reason to not merge but prior stalls make me worry that indeed some race condition might come to light.

mih (approved):
LGTM, Thx!

Sadly in conflict with #6111, but oh well...

PR #6142, opened 2021-11-04
Remove all usage of exc_str()

mih:
Fully implementing http://docs.datalad.org/en/latest/design/exception_handling.html

Fixes datalad/datalad#5975


yarikoptic:
FWIW: I still can't fathom the value of adding a single use `ce` variable (longer code etc) and not just aliasing `exc_str` to be `CapturedException` besides rare cases of debugging whenever the exception manages to bubble up through all the generators and results handling in `--dbg` mode.

mih:
> FWIW: I still can't fathom the value of adding a single use `ce` variable (longer code etc) and not just aliasing `exc_str` to be `CapturedException` besides rare cases of debugging whenever the exception manages to bubble up through all the generators and results handling in `--dbg` mode.

I was expecting this comment (again). I will respond in the same fashion as before. I just implemented the concept that was discussed and put in writing. With this PR the concept is fully implemented. We can now start the discussion what the next concept will be, and then we can implement that too. ;-)

PR #6143, opened 2021-11-04
BF produce an error in create and run_procedure if procedure does not exist

### LINKED ISSUE #6129: No warning / error when incorrect config procedure is given to datalad create -c
### What is the problem?

When datalad create is used with an incorrect run procedure name, it generates no warnings or errors. The exit status reported by zsh is not zero, but there is no explicit information why (no `create(ok)` message either). The dataset gets created, just without applying the configuration.

Verify expected operation when a correct name is given:
```
$ datalad create -c text2git my-dataset
[INFO   ] Creating a new annex repo at /home/mszczepanik/Documents/rdm-warmup/example/my-dataset 
create(ok): /home/mszczepanik/Documents/rdm-warmup/example/my-dataset (dataset)
```
Reproduce by using an incorrect name:
```
datalad create -c texttogit my-other-dataset
[INFO   ] Creating a new annex repo at /home/mszczepanik/Documents/rdm-warmup/example/my-other-dataset 
```

### Datalad information

Datalad 0.15.0 on Debian 11 (bullseye)

### Additional context

This message is shown when `--log-level debug` is used 
```
$ datalad --log-level debug create -c texttogit my-other-dataset
(...)
[DEBUG  ] Cannot find procedure with name 'cfg_texttogit' [run_procedure(cfg_texttogit)] 
[DEBUG  ] could not perform all requested actions: IncompleteResultsError(Command did not complete successfully. 1 failed:
[{'action': 'run_procedure',
  'message': "Cannot find procedure with name 'cfg_texttogit'",
  'path': 'cfg_texttogit',
  'refds': '/home/mszczepanik/Documents/rdm-warmup/example/my-other-dataset',
  'status': 'impossible'}]) 
```

The message comes from `datalad/interface/run_procedure.py` [here](https://github.com/datalad/datalad/blob/3478fefef9f20a665108a74e2ab16aca48eb5495/datalad/interface/run_procedure.py#L398). If there's agreement that  this situation should be elevated to error/warning, I think I could try submitting a PR.

Yes, though mostly tutorial-level stuff.

mslw:
This PR changes the behaviour of `create` and `run_procedure` so that using an incorrect (unavailable) procedure name results in an error. Closes #6129 

Specifically, the following changes were made:

1. The `run_procedure` on its own was changed to raise a `ValueError` instead of reporting a result with `status='impossible'`
2. A call to `run_procedure(discover=True)` was introduced in `datalad create` before any creation takes place to check whether the requested `cfg_proc(s)` are available.
3. To avoid going through a discovery process again when the procedure is applied to a newly created dataset, results of the call described above are reused. To make it possible, a new behaviour was introduced to `run_procedure(spec, ...)` when `spec` is a dictionary.

The third change seems the most "invasive" to me so I'd be happy to have your opinions on whether this is the correct way.

Resulting behaviour:
```
datalad create -c unknown my-dataset                                                             `129` !
[ERROR  ] ValueError(Cannot find procedure with name 'unknown') (ValueError)
```

Remaining TODO items:
- [x] Add a test for `create` asserting that `ValueError` is raised and the dataset is not created
- [x] Add a test for `run_procedure` asserting that it can be done with a (proper) dictionary input.

### DIFF
--- a/datalad/core/local/create.py
+++ b/datalad/core/local/create.py
@@ -339,18 +339,22 @@ class Create(Interface):
             yield res
             return
 
-        # Check if specified cfg_proc(s) can be discovered. If not,
-        # raise an error to prevent creating the dataset.
+        # Check if specified cfg_proc(s) can be discovered, storing
+        # the results so they can be used when the time comes to run
+        # the procedure. If a procedure cannot be found, raise an
+        # error to prevent creating the dataset.
+        cfg_proc_specs = []
         for cfg_proc_ in cfg_proc:
             for discovered_proc in tbds.run_procedure(
                     discover=True,
                     result_renderer='disabled',
                     return_type='generator'):
                 if discovered_proc['procedure_name'] == 'cfg_' + cfg_proc_:
+                    cfg_proc_specs.append(discovered_proc)
                     break
             else:
-                raise ValueError("Cannot find procedure with name"
-                                 "'%s'" % cfg_proc)
+                raise ValueError("Cannot find procedure with name "
+                                 "'%s'" % cfg_proc_)
 
         if initopts is not None and isinstance(initopts, list):
             initopts = {'_from_cmdline_': initopts}
@@ -424,8 +428,8 @@ class Create(Interface):
             _status=add_to_git,
         )
 
-        for cfg_proc_ in cfg_proc:
-            for r in tbds.run_procedure('cfg_' + cfg_proc_):
+        for cfg_proc_spec in cfg_proc_specs:
+            for r in tbds.run_procedure(cfg_proc_spec):
                 yield r
 
         # the next only makes sense if we saved the created dataset,
--- a/datalad/core/local/create.py
+++ b/datalad/core/local/create.py
@@ -344,17 +344,17 @@ class Create(Interface):
         # the procedure. If a procedure cannot be found, raise an
         # error to prevent creating the dataset.
         cfg_proc_specs = []
-        for cfg_proc_ in cfg_proc:
-            for discovered_proc in tbds.run_procedure(
-                    discover=True,
-                    result_renderer='disabled',
-                    return_type='generator'):
-                if discovered_proc['procedure_name'] == 'cfg_' + cfg_proc_:
-                    cfg_proc_specs.append(discovered_proc)
-                    break
-            else:
-                raise ValueError("Cannot find procedure with name "
-                                 "'%s'" % cfg_proc_)
+        if cfg_proc:
+            discovered_procs = tbds.run_procedure(
+                discover=True, result_renderer='disabled')
+            for cfg_proc_ in cfg_proc:
+                for discovered_proc in discovered_procs:
+                    if discovered_proc['procedure_name'] == 'cfg_' + cfg_proc_:
+                        cfg_proc_specs.append(discovered_proc)
+                        break
+                else:
+                    raise ValueError("Cannot find procedure with name "
+                                     "'%s'" % cfg_proc_)
 
         if initopts is not None and isinstance(initopts, list):
             initopts = {'_from_cmdline_': initopts}
--- a/datalad/core/local/create.py
+++ b/datalad/core/local/create.py
@@ -339,6 +339,19 @@ class Create(Interface):
             yield res
             return
 
+        # Check if specified cfg_proc(s) can be discovered. If not,
+        # raise an error to prevent creating the dataset.
+        for cfg_proc_ in cfg_proc:
+            for discovered_proc in tbds.run_procedure(
+                    discover=True,
+                    result_renderer='disabled',
+                    return_type='generator'):
+                if discovered_proc['procedure_name'] == 'cfg_' + cfg_proc_:
+                    break
+            else:
+                raise ValueError("Cannot find procedure with name"
+                                 "'%s'" % cfg_proc)
+
         if initopts is not None and isinstance(initopts, list):
             initopts = {'_from_cmdline_': initopts}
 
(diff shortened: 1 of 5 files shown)


datalad/core/local/create.py
```diff
@@ -339,6 +339,23 @@ def __call__(
...
+        # Check if specified cfg_proc(s) can be discovered, storing
+        # the results so they can be used when the time comes to run
+        # the procedure. If a procedure cannot be found, raise an
+        # error to prevent creating the dataset.
+        cfg_proc_specs = []
+        for cfg_proc_ in cfg_proc:
+            for discovered_proc in tbds.run_procedure(
+                    discover=True,
+                    result_renderer='disabled',
+                    return_type='generator'):
```
mih, line 351: If I read this correctly, then this will call `run_procedure` once for each given config procedure name. This seems wasteful, as the result of each call will be identical. Why not call it once, and then search for each given `cfg_proc`?

datalad/interface/run_procedure.py
```diff
@@ -280,10 +280,11 @@ class RunProcedure(Interface):
             args=("spec",),
             metavar='NAME [ARGS]',
             nargs=REMAINDER,
-            doc="""Name and possibly additional arguments of the
-            to-be-executed procedure. [CMD: Note, that all options to
-            run-procedure need to be put before NAME, since all ARGS get
-            assigned to NAME CMD]"""),
+            doc="""Name and possibly additional arguments of the to-be-executed
+            procedure. [PY: Can also be a dictionary coming from
+            run-procedure(discover=True).] [CMD: Note, that all options
```
mih, line 285: Possibly:

```suggestion
            run-procedure(discover=True).][CMD: Note, that all options
```

mih (changes requested):
Thx for the PR!

Apart from a performance-related aspect, the approach looks good and clean to me. Once the multi-call- issue is fixed, we can merge this IMHO. Thx!

datalad/core/local/create.py
mslw: I assumed there won't be many cases when multiple cfg_procs are given, but indeed I wasn't sure about this part. Probably better to create a list once than a generator (possibly) many times. I'm changing according to your suggestion.

adswa:
The failure on travis seems unrelated:

```
=====================================================================

ERROR: datalad.interface.tests.test_download_url.test_download_url_need_datalad_remote

  File "/tmp/dl-miniconda-duu7yni7/lib/python3.9/site-packages/nose/case.py", line 198, in runTest

    self.test(*self.arg)

  File "/tmp/dl-miniconda-duu7yni7/lib/python3.9/site-packages/datalad/tests/utils.py", line 190, in _wrap_skip_if_no_network

    return func(*args, **kwargs)

  File "/tmp/dl-miniconda-duu7yni7/lib/python3.9/site-packages/datalad/tests/utils.py", line 737, in _wrap_with_tempfile

  File "/tmp/dl-miniconda-duu7yni7/lib/python3.9/site-packages/datalad/interface/tests/test_download_url.py", line 237, in test_download_url_need_datalad_remote

    ds_a.download_url([url], path="foo")

  File "/tmp/dl-miniconda-duu7yni7/lib/python3.9/site-packages/datalad/distribution/dataset.py", line 485, in apply_func

    return f(**kwargs)

  File "/tmp/dl-miniconda-duu7yni7/lib/python3.9/site-packages/datalad/interface/utils.py", line 484, in eval_func

  File "/tmp/dl-miniconda-duu7yni7/lib/python3.9/site-packages/datalad/interface/utils.py", line 476, in return_func

    results = list(results)

  File "/tmp/dl-miniconda-duu7yni7/lib/python3.9/site-packages/datalad/interface/utils.py", line 461, in generator_func

    raise IncompleteResultsError(

datalad.support.exceptions.IncompleteResultsError: Command did not complete successfully. 1 failed:

[{'action': 'download_url',

  'exception': BotoServerError: 503 Slow Down

 [download_url.py:__call__:199,base.py:download:517,base.py:access:162,s3.py:_establish_session:228,utils.py:_wrap_try_multiple_dec:2053,s3.py:authenticate:121,s3.py:get_bucket:131,utils.py:_wrap_try_multiple_dec:2053,connection.py:get_bucket:509,connection.py:head_bucket:528,connection.py:make_request:667,connection.py:make_request:1070,connection.py:_mexe:1028],

  'exception_traceback': '[download_url.py:__call__:199,base.py:download:517,base.py:access:162,s3.py:_establish_session:228,utils.py:_wrap_try_multiple_dec:2053,s3.py:authenticate:121,s3.py:get_bucket:131,utils.py:_wrap_try_multiple_dec:2053,connection.py:get_bucket:509,connection.py:head_bucket:528,connection.py:make_request:667,connection.py:make_request:1070,connection.py:_mexe:1028]',

  'message': 'BotoServerError(BotoServerError: 503 Slow Down\n)',

  'path': '/tmp/datalad_temp_test_download_url_need_datalad_remotenico1j3p/a/foo',

  'status': 'error',

  'type': 'file'}]
```

You can fix the ``semver`` test by adding a versioning label, e.g., ``patch`` (see https://github.com/datalad/datalad/blob/master/CONTRIBUTING.md#labelling-pull-requests). Should you not have permissions to add a label, I think this should be fixed :) 

I tried your change, works nicely for me - thanks!

mslw:
Thank you for taking a look at the proposed changes and the checks @adswa! 

> The failure on travis seems unrelated

Guess so, unless that's a weird collateral - unfortunately, I have nothing to add.

> You can fix the semver test by adding a versioning label, e.g., patch (see https://github.com/datalad/datalad/blob/master/CONTRIBUTING.md#labelling-pull-requests). Should you not have permissions to add a label, I think this should be fixed :)

Thanks for pointing it out (we also talked about it with @bpoldrack yesterday). Indeed, I do not have permissions - no button saying "Add label" where it would normally be. Whether that's something to be fixed, I don't know. I think that's tied to repository write access?

adswa:
I think @mih should add you to the datalad organization - do you know, @mih?

mih (approved):
Thx for the update, LGTM now.

Issue #5833, opened 2021-07-28
Rename auto-related issue/pr labels


mih:
I think they are hard to find in the slew of labels we have, and a common prefix like `version-` would substantially aid discoverability for me -- as someone who is actively developing, but not into daily release management operations.

yarikoptic:
I think `auto` doesn't support "custom labels", @jwodder ?  I just remember them "by heart" now, it is only a few.

jwodder:
`auto` does support changing the label names; see <https://intuit.github.io/auto/docs/configuration/autorc#labels>.

mih:
Now that these labels are also required for working on master, I would really appreciate this rename to aid findability.

I also feel like the descriptions could be clarified. There is only the list in CONTRIBUTION.md, which also has other labels (so there is no standalone list that identifies the full set of relevant choices).

- minor — for changes corresponding to an increase in the minor version component
- patch — for changes corresponding to an increase in the patch/micro version component; this is the default label for unlabelled PRs
- internal — for changes only affecting the internal API
- tests — for changes to tests

`patch` is described as "default", but it seems it is actually necessary to be selected manually, if no others are selected.

It seems that `internal` and `tests` semantically mean `...-only`, as in it is not necessary to bump patch or minor version.

Is there any practical use for the additional `tests` label vs. just `internal`? If yes, do you now expect any PR (i.e. practically all of them) to have an additional `tests` label?

A common label prefix would also clarify the semantics of such generic terms as "internal" and "tests".

Again, these labels might be obvious in standard repos with a handful of labels, but we are presently operating a model with ~100 labels.

yarikoptic:
> *  `patch` is described as "default", but it seems it is actually necessary to be selected manually, if no others are selected.

just because to help us to not forget etc, I asked @jwodder to add that action which would do extra checks that we didn't forget to label our PRs. `auto` would just use `patch` by default indeed

> It seems that `internal` and `tests` semantically mean `...-only`, as in it is not necessary to bump patch or minor version.

correct

> Is there any practical use for the additional `tests` label vs. just `internal`? If yes, do you now expect any PR (i.e. practically all of them) to have an additional `tests` label?

[...]

PR #6151, opened 2021-11-08
infra: Provide custom prefix to auto-related labels

adswa:
In looking through the issues closed over the weekend, I felt compelled to act on the suggestion in #5833 to prefix a common, versioning-related identifier to the PR labels for the auto-workflow.
I think there was some general agreement in the original issue, but not enough momentum to implement the change. I personally keep forgetting which labels exist, and would be in favor of having them in one group to ease discoverability. 
I'm suggesting ``semver`` as a prefix, because the corresponding test is called "Test for semver label". I'm happy to change this into anything else, though.

- [x] if this change is accepted, the label names need to be changed

### DIFF
--- a/.autorc
+++ b/.autorc
@@ -12,5 +12,15 @@
                 "afterRelease": "python -m build && twine upload dist/*"
             }
         ]
-    ]
+    ],
+  "labels": [
+    { "releaseType": "major", "name": "semver-major" },
+    { "releaseType": "minor", "name": "semver-minor" },
+    { "releaseType": "patch", "name": "semver-patch" },
+    { "releaseType": "dependencies", "name": "semver-dependencies" },
+    { "releaseType": "documentation", "name": "semver-documentation" },
+    { "releaseType": "internal", "name": "semver-internal" },
+    { "releaseType": "performance", "name": "semver-performance" },
+    { "releaseType": "tests", "name": "semver-tests" }
+  ]
 }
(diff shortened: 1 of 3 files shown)


yarikoptic:
I wonder on what disturbance that would cause interim since if we rename the labels and workflow in master -- `maint` from which we do cut releases wouldn't be able to `auto` release, so it would break that component.
If we are to do such a change, I feel it should happen right after some release so both `maint` and `master` could have a chance to auto-release without problems and collate perspective changelog based on labels of already merged PRs etc.

mih:
I am all for such a change. I am totally unaware of what disturbances it would cause. But I do know that a release from `master` didn't work already with 0.15.0 (so no breakage possible here), and trying right *after* a release will only delay a reliable test of whether there is really no breakage. The worth case scenario -- an auto release does not work -- is nothing that concerns me at all.

adswa:
With @mih's statement and @bpoldrack's thumbs up this kinda sounds like we could proceed here? AFAIK we could rename the current labels by editing them

yarikoptic:
FWIW, on a 2nd thought, since `auto` analyzes PR labels only when running, and renaming (not creating new) hopefully would remain them in already labeled PRs, I think it should be safe to proceed **but** this PR most likely need to be done against `maint`, and then `maint` merged into master so everything is consistent and nice.

upon rename -- would be worthwhile to check recently merged PRs if their labels were updated correspondingly.

mih:
Sound like a plan! Thx!

adswa:
I rebased to ``maint``, and will edit the labels now.

adswa:
I had to re-attach the label to the PR to have it picked up by the semver test

mih:
I killed the appveyor tests -- nothing in here is relevant for appveyor.

mih:
OK, let's do it.

Issue #6150, opened 2021-11-07
Enhance httpserve test helpers with auth and ssl/tls


mih:
These are both common aspects of contemporary http usage, and it would be beneficial to be able to use them in tests easily.

Generating certificates can be done with openssl, and serving https seems doable with stdlib tooling. Not sure if self-signed certificates are a good starting point.

Basic auth requires a relatively straightforward customization of the requesthandler. Also doable.

mih:
Coming via #6153

PR #6158, opened 2021-11-08
RF a first RIA test to run across platforms

mih:
The previous strict linkage between access protocols and IO
implementation to generate test stores was pointless.
This is a general pattern that needs to be altered throughout the
codebase.


mih:
Closing and merging with #6150

Issue #6054, opened 2021-10-07
Create design document on result records


mih:
See https://github.com/datalad/datalad/issues/5940#issuecomment-937687710 for an example issue that stems from the lack of an explicit description pf what result records are and what requirements they must meet.

PR #6167, opened 2021-11-10
Design document on result records

mih:
Developers are encouraged to amend this initial document with additional information, in particular field names that are in active use.

- Fixes datalad/datalad#6054
- Fixes datalad/datalad#6171

Please suggest concrete changes/addition using the reviewing functionality of by pushing the this branch. Given the documentation-nature of this PR, I will merge this tomorrow, unless objections are expressed.

### DIFF
--- a/docs/source/design/result_records.rst
+++ b/docs/source/design/result_records.rst
@@ -1,0 +1,216 @@
+.. -*- mode: rst -*-
+.. vi: set ft=rst sts=4 ts=4 sw=4 et tw=79:
+
+.. _chap_design_result_records:
+
+**************
+Result records
+**************
+
+.. topic:: Specification scope and status
+
+   This specification describes the current implementation.
+
+Result records are the standard return value format for all DataLad commands.
+Each command invocation yields one or more result records. Result records are
+routinely inspected throughout the code base, and are used to inform generic
+error handling, as well as particular calling commands on how to proceed with
+a specific operation.
+
+The technical implementation of a result record is a Python dictionary.  This
+dictionary must contain a number of mandatory fields/keys (see below). However,
+an arbitrary number of additional fields may be added to a result record.
+
+The ``get_status_dict()`` function simplifies the creation of result records.
+
+.. note::
+   Developers *must* compose result records with care! DataLad supports custom
+   user-provided hook configurations that use result record fields to
+   decide when to trigger a custom post-result operation. Such custom hooks
+   rely on a persistent naming and composition of result record fields.
+   Changes to result records, including field name changes, field value changes,
+   but also timing/order of record emitting potentially break user set ups!
+
+
+Mandatory fields
+================
+
+The following keys *must* be present in any result record. If any of these
+keys is missing, DataLad's behavior is undefined.
+
+
+``action``
+----------
+
+A string label identifying which type of operation a result is associated with.
+Labels *must not* contain white space. They should be compact, and lower-cases,
+and use ``_`` (underscore) to separate words in compound labels.
+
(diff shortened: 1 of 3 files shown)


docs/source/design/result_records.rst
```diff
@@ -0,0 +1,209 @@
...
+- ``impossible``: a requested operation cannot be performed, possibly because
+  its preconditions are not met
+- ``error``: an error occurred while performing an operation
+
+Based on the ``status`` field, a result is categorized into *success* (``ok``,
+``notneeded``) and *failure* (``impossible``, ``error``). Depending on the
+``on_failure`` parameterization of a command call, any failure-result emitted
+by a command can lead to an ``IncompleteResultsError`` being raised on
+command exit, or a non-zero exit code on the command line.
+
```
adswa, line 82: In addition to raising errors, isn't it also "nothing" when we set ``on_failure='ignore'``? Could we maybe add the possible values that `on_failure` can take (sorry if I have missed them)?

adswa:
I find this very helpful! Left a few comments

docs/source/design/result_records.rst
mih: Will do, good point, thx!

docs/source/design/result_records.rst
```diff
@@ -0,0 +1,209 @@
...
+------------------
+
+List of any error messages that were captured or produced while achieving a
+result.
+
+
+``exception``
+-------------
+
+An exception that occurred while achieving the reported result.
```
bpoldrack, line 179:
```suggestion
A ``CapturedException`` that occurred while achieving the reported result. If provided, ``get_status_dict`` will fill in the ``exception_traceback`` field automatically.
```

docs/source/design/result_records.rst
```diff
@@ -0,0 +1,209 @@
...
+If a result record has a ``message`` field, then a given `Logger` instance
+(typically from ``logging.getLogger()``) will be used to automatically log
+this message. The log channel/level is determined based on
+``datalad.log.result-level`` configuration setting. By default, this is
+the ``debug`` level. When set to ``match-status`` the log level is determined
+based on the ``status`` field of a result record:
+
+- ``debug`` for ``'ok'``, and ``'notneeded'`` results
+- ``warning`` for ``'impossible`` results
+- ``error`` for ``'impossible`` results
```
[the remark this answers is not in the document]
mih: Thx, it also has a quote too much. Will fix.

docs/source/design/result_records.rst
mih: Is that true? It seems the code takes any exception. The feature is undocumented.

bpoldrack: Result rendering may take any, but `get_status_dict` doesn't. So, at least when taking advantage of that convenience feature `CapturedException` would be required.

bpoldrack: FWIW: I'm fine with declaring that `get_status_dict` *should* be able to deal with any. However, that's defining current behavior as a bug and I'd argue it's still worth documenting that feature here.

mih: I have no desire to declare anything. I am documenting what I think I know is going on. It seems that this feature was exactly how I had it in mind and described it (added by @yarikoptic in https://github.com/datalad/datalad/commit/b737726673c4e47ad3e15d78654eb981146bbe79), but was silently changed (now restricted to `CapturedException`) by @bpoldrack with the recent https://github.com/datalad/datalad/commit/6d4f1ebb901010725b9dba4dd4d2e75af474dbfe.

PR #6168, opened 2021-11-10
BF: Don't overwrite subdataset source candidates

### LINKED ISSUE #6159: datalad get fails on subdataset with different ria URL than superdataset
### What is the problem?

I created a Datalad dataset with subdatasets for every subject for a BIG neuroimaging dataset: Ds - manyDs 
Then I add this dataset with many subdataset as input dataset to another superdataset: SDs -> Ds - manyDs
When I push this setup to a RIA URL with another ria url than the first Ds containing all subdatasets, I cannot datalad get the subdatasets form the supersubdataset, while it works with the identical ria url for SDs & Ds.  
When I check the git remote in the subdataset the RIA does not resolve to the hash in the case of different URLs for SDs and manyDs.
```
RIA superdataset:    /data/project/inputstore
RIA subdataset:	 /data/project/dataladstore/alias/3T_data
```
```
RIA superdataset:  /data/project/dataladstore
RIA subdataset:     /data/project//dataladstore/5e7/1a5e6-95c7-4b38-9099-5c46f2379007
```

Should be rather clear from above:
Add (clone by alias link) a dataset with subdatasets to a superdataset. Then create a sibling with another URL than the added dataset. Uninstall the added dataset. Push the superdataset to the URL (different than the added one). Clone the Superdataset and datalad get the dataset and subdatasets failed then for me. 

### Datalad information

datalad 0.15.0

## python 
  - implementation: CPython
  - version: 3.7.3
## system 
  - distribution: debian/10.11
  - encoding: 
    - default: utf-8
    - filesystem: utf-8
    - locale.prefered: UTF-8
  - max_path_length: 297
  - name: Linux
  - release: 4.19.0-18-amd64
  - type: posix
  - version: #1 SMP Debian 4.19.208-1 (2021-09-29)

### Additional context

I am using our workflow:  
https://github.com/psychoinformatics-de/fairly-big-processing-workflow

Yes, absolutely, great things 💯 
🏄🏼

bpoldrack:
datalad-get used to overwrite possibly existing source candidate configs
in newly cloned subdatasets. While inheriting is generally fine, it
didn't take into account that postclone_cfg routines may already have
had a better idea.

"Better idea" in that I think the principle of looking up subdatasets in the same RIA store their super came from should apply at all levels recursively. Hence, when `super` came from storeA, storeA is also considered for `sub`. If it, however, turns out that `sub` ultimately came from storeB, then `subsub` should be looked for in storeB. And that's the situation where that overwrite matters AFAICT.

Closes #6159

### DIFF
--- a/datalad/distribution/get.py
+++ b/datalad/distribution/get.py
@@ -369,13 +369,15 @@ def _install_subds_from_flexible_source(ds, sm, **kwargs):
         for rec in cand_cfg:
             # check whether any of this configuration originated from the
             # superdataset. if so, inherit the config in the new subdataset
-            # clone. if not, keep things clean in order to be able to move with
-            # any outside configuration change
+            # clone unless that config is already specified in the new
+            # subdataset which can happen during postclone_cfg routines.
+            # if not, keep things clean in order to be able to move with any
+            # outside configuration change
             for c in ('datalad.get.subdataset-source-candidate-{}{}'.format(
                           rec['cost'], rec['name']),
                       'datalad.get.subdataset-source-candidate-{}'.format(
                           rec['name'])):
-                if c in super_cfg.keys():
+                if c in super_cfg.keys() and c not in subds.config.keys():
                     subds.config.set(c, super_cfg.get(c), where='local',
                                      reload=False)
                     need_reload = True
--- a/datalad/distribution/tests/test_get.py
+++ b/datalad/distribution/tests/test_get.py
@@ -28,6 +28,7 @@ from datalad.support.exceptions import (
     InsufficientArgumentsError,
     RemoteNotAvailableError,
 )
+from datalad.support.network import get_local_file_url
 from datalad.tests.utils import (
     ok_,
     eq_,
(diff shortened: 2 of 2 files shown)


bpoldrack:
FTR: There are failing tests, b/c I based that fix on `maint` while calling `create-sibling-ria` with the new `new_store_ok` option.
So, I guess I will make that two PRs instead, since it's still a bug fix that should go into `maint` but not reintroduce unadjusted calls when merging back into `master`.

mih:
Travis failure is a known fluke, but windows appveyor looks real.

mih:
What is the status here? The fix it there, but the test issues seem unaddressed? Having this PR just sitting in this limbo state is not a good thing.

bpoldrack:
@mih 
Verdict from call was to skip that test on windows, since RIA+ORA are not working on windows anyway and this is the only scenario deemed worth testing. Pushed accordingly.

Otherwise it remained somewhat unclear to me whether there are objections to the fix itself.

bpoldrack:
Edited description to point the reasoning behind this fix.

An alternative would be to take the config from super and try to find an available spot with higher cost within whatever sub already has set. Given that so far only clones from RIA store would get into that situation to begin with, that feels like too big a gun to me, though.

mih (approved):
Thanks, LGTM!

Here as in all other places, it would be so important to have a design document, where this could simply be added to the description of the implemented behavior. I hope we will get there...

Issue #6164, opened 2021-11-09
Stop using Appdirs


adswa:
We rely on [appdirs](https://github.com/ActiveState/appdirs) to figure out appropriate config, data, and cache dirs for any given platform. However, the implementation in appdirs for Windows is for Windows 7 at latest. It relies on [Windows special folders CSIDL IDs](https://www.nirsoft.net/articles/find_special_folder_location.html) and retrieves them for example like this:

```
path = os.path.normpath(_get_win_folder("CSIDL_LOCAL_APPDATA"))
```
where ``CSIDL_LOCAL_APPDATA`` is an identifier for a special folder. This is necessary because Windows versions change the location and name of these special folders.
However, since Windows Vista, these IDs have been replaced by [``KNOWNFOLDERID`` values](https://docs.microsoft.com/en-us/windows/win32/shell/csidl), and Windows advises to use them instead.

I believe this is the underlying cause for the test failure observed in #6160. 

I filed [an issue with appdirs](https://github.com/ActiveState/appdirs/issues/167) this summer because certain CSIDL ID's simply don't exist (they return an empty string on the windows 10/11 test boxes) but did not receive a reply, and it doesn't look like there is much development anymore. 

I think, at least for Windows-related things, we should switch away from appdirs.

jwodder:
For the record, the modern, maintained replacement for appdirs is [platformdirs](https://github.com/platformdirs/platformdirs).

adswa:
> For the record, the modern, maintained replacement for appdirs is [platformdirs](https://github.com/platformdirs/platformdirs).

Thanks a lot! I guess we should at least switch to that one. They also seem to use CSIDL values instead of KNOWNFOLDER IDs under Windows, sadly. I have a rather lengthy alternative that I might PR as a draft later...

Issue #6160, opened 2021-11-09
Windows test setup overwrites ``USERPROFILE`` variable, resulting in leaked caches etc in test datasets


mih:
Two items `datalad` and `datalad.org`. And it seems it uses the dataset root as a cache directory. It contains the cookie DB, and also the downloader locks.

For now I can only see this behavior with an HTTP store URL, all other access methods are even more severely broken.

`initremote` arguments are `['encryption=none', 'type=external', 'externaltype=ora', 'autoenable=true', 'url=ria+http://127.0.0.1:54420/']`

A possible cause is the use of `download_url` in `HTTPRemoteIO.get|read_file()` (not connected to any dataset). And careful: this `download_url` is not related to the datalad command, but a helper from `datalad/support/network.py`. This helper is exclusively used in RIA code (dunno why). That being said, `download_url` cannot be blamed on its own, its tests (also on windows) work fine, and with the tested usage, it is not showing this behavior.

adswa:
Called from the command line, git-annex initremote behaves as expected in the relevant part of the test:

```

C:\Users\datalads\AppData\Local\Temp\datalad_temp_lebhxtnu>git status
On branch adjusted/dl-test-branch(unlocked)
nothing to commit, working tree clean

# should fail, but not deposit any files
C:\Users\datalads\AppData\Local\Temp\datalad_temp_lebhxtnu>git annex initremote ria-remote encryption=none type=external externaltype=ora autoenable=true url=ria+http://127.0.0.1:53436/
initremote ria-remote [INFO] Fetching 'http://127.0.0.1:53436/ria-layout-version'

git-annex: Access to http://127.0.0.1:53436/ria-layout-version has failed: not found
failed
initremote: 1 failed

C:\Users\datalads\AppData\Local\Temp\datalad_temp_lebhxtnu>git status
On branch adjusted/dl-test-branch(unlocked)
nothing to commit, working tree clean
```

Calling the part of the test that should do the exact same thing, the problematic behavior shows:

```
(Pdb) n
> c:\users\datalads\repos\datalad\datalad\distributed\tests\test_ria_basics.py(76)_test_initremote_basic()
-> ds.repo.init_remote, 'ria-remote', options=init_opts)
(Pdb) p init_opts
['encryption=none', 'type=external', 'externaltype=ora', 'autoenable=true', 'url=ria+http://127.0.0.1:53436/']
(Pdb) l
 71
 72         init_opts = common_init_opts + ['url={}'.format(url)]
[...]

PR #6169, opened 2021-11-10
New helper `tools/adhoc-httpd` to facilitate manual HTTP debugging

mih:
This helper exposes `serve_path_via_http` via a cmdline utility that is
capable of deploying an ad-hoc instance of the same HTTP server that we
are using for internal testing (with SSL and auth, if desired).

This lowers the threshold for debugging real usage, by avoiding the need
to develop a dedicated test for a scenario. It also makes it possibly to
bypass to additional implications of a datalad test implementation, like
the fake homedir (see #6164, #6160).

### DIFF
--- a/datalad/tests/utils.py
+++ b/datalad/tests/utils.py
@@ -713,9 +713,12 @@ class HTTPPath(object):
             # if this fails, check datalad/tests/ca/prov.sh for ca_bundle
             try:
                 import requests
-                requests.get(
+                from requests.auth import HTTPBasicAuth
+                r = requests.get(
                     self.url,
-                    verify=True)
+                    verify=True,
+                    auth=HTTPBasicAuth(*self.auth) if self.auth else None)
+                r.raise_for_status()
             # be robust and skip if anything goes wrong, rather than just a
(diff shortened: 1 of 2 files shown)


bpoldrack:
Nice!

yarikoptic:
sweet -- will come handy.
Note though that PR is not just an addition of a helper but a tune up to the `HTTPPath` itself with not explained reasoning for the changes there ( I thought that auth was already implemented, so not clear why these extra changes).

mih:
> Note though that PR is not just an addition of a helper but a tune up to the `HTTPPath` itself with not explained reasoning for the changes there ( I thought that auth was already implemented, so not clear why these extra changes).

yes, good catch. I should have added this to the message. I discovered that the check for proper function of an SSL enabled webserver *that also had basic auth enabled* would not pass, because the check itself did not perform authentication properly. This is inconsequential for datalad as of now, because I am only adding such tests in a local branch that improves the RIA/ORA testing. But given that the helper can actually trigger the problem, I thought to include the fix here. More comprehensive machinery on this will come.

Issue #6114, opened 2021-10-25
Harmonize "hints" provided to the user


yarikoptic:
ATM we do not have consistent/agreed upon way/interface to provide "hints" to the users. 
- #6108 uses `ui.message` and relies on `datalad.ui.git-commit-notification*` config variables
- #6112 uses `lgr.info`

We also provide hints via `message` of the result record and then display them via `ui.message` in the renderer. 
But it is also not easy to make a clear cut, since e.g. some of the git-annex messages could be considered to be hints, and in general we do not want to parse/specially handle those.

Placing channeled from git-annex messages aside, I think it would be useful to 

- decide on consistent config prefix (e.g. `datalad.hints`) to aggregate all possible hints and possibly provide centralized configuration (`datalad.hints.enabled:bool=True`) to enable/disable them
- add `hints:list[str]` to result records
  - cons: will not be displayed when Python API used
- centralize their display in results renderer

adswa:
@yarikoptic and @bpoldrack discussed this in a devcall, but we failed to take any notes. Could you remember the main points you concluded? Else, I'd propose to discuss this issue again in a (more crowded) devcall.

mih:
@bpoldrack agreed to provide a summary of the discussion here

bpoldrack:
So, the somewhat rough idea is to have a central (sort of) table, that defines hints unambiguously with an ID. This is supposed to help with two things:

- Different places may want to add hints that are semantically the same thing. With such a table we can avoid to have several different wordings for the same thing, which may be confusing when one actually encounters both.
- When deciding what to do with hints in rendering we can use IDs rather than string parsing to find out we have the same thing several times

Furthermore, hints should be made part of result records, so renderers can decide whether and how to show them, since low-level code encountering a potential issue has little knowledge about context (command, interactivity, previously issued hints and so on). Likely that would mean to simply refer to the hint ID in the result records.

PR #6173, opened 2021-11-10
ENH(UX): log a hint to use ulimit command in case of "Too long" exception

### LINKED ISSUE #6106: "Argument list too long: 'git' " while saving datalad dataset
I have a issue very similar to #4778. I wanted to datalad save all the files in a datalad dataset, and `datalad -l debug save .` returns the errors:
```
[DEBUG  ] Future for '/cbica/projects/RBC/RBC_RAWDATA/bidsdatasets/HCP_D' raised [Errno 7] Argument list too long: 'git'.  Re-raising to trigger graceful shutdown etc 
[WARNING] Received an exception [Errno 7] Argument list too long: 'git' [subprocess.py:_execute_child:1704].
| Canceling not-yet running jobs and waiting for completion of running.
| You can force earlier forceful exit by Ctrl-C. 
[DEBUG  ] Shutting down <concurrent.futures.thread.ThreadPoolExecutor object at 0x7fa0658d27c0> with 0 futures. Reason: [Errno 7] Argument list too long: 'git' 
[INFO   ] Canceled 0 out of 0 jobs. 0 left running. 
[DEBUG  ] Finished shutdown with force=False due to exception=OSError(7, 'Argument list too long') 
[DEBUG  ] Shutting down <concurrent.futures.thread.ThreadPoolExecutor object at 0x7fa0658d27c0> with 0 futures. Reason: [Errno 7] Argument list too long: 'git' 
[ERROR  ] [Errno 7] Argument list too long: 'git' [subprocess.py:_execute_child:1704] (OSError) 
[DEBUG  ] Shutting down None with 0 futures. Reason: None 
[DEBUG  ] Finished shutdown with force=True due to exception=None 
```

Checking the maximum argument length returns
```
In [1]: import os

In [2]: os.sysconf('SC_ARG_MAX')
Out[2]: 63963136
```
FYI, Datalad version: 0.14.6 and I am using it on a Linux based HPC system. Thanks so much for your time!

yarikoptic:
Closes #6106

Fragile part of the test -- we might not hit that limit may be?  Let's see what CI says. Alternative would be to mock that Popen call but that would get us away from reality.

Sure thing could be improved etc, but overall I am giving up on trying to set/control that CMD_MAX_ARG and decided just to hint (related: #6114) user on how to mediate.

### DIFF
--- a/datalad/runner/nonasyncrunner.py
+++ b/datalad/runner/nonasyncrunner.py
@@ -26,6 +26,7 @@ from typing import (
     Union,
 )
 
+from datalad.utils import on_windows
 from .protocol import WitlessProtocol
 
 lgr = logging.getLogger("datalad.runner.nonasyncrunner")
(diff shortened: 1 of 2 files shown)


mih (commented):
Wouldn't this be limited to non-localized machines?

yarikoptic:
we already rely on parsing error messages from `git` IIRC, and for that
```
datalad/runner/gitrunner.py:        git_env['LC_MESSAGES'] = 'C'
```
so I would assume that even in your case our GitRunner (which will use the base runner) would report in English I would assume. Would have been nice to check.  Sure thing if runner is used directly, and without tune up of env -- no hint would be produced.  But since it is just a hint -- better to have a hint in some (if not most) of the cases  than none IMHO.

mih:
Thanks for the explanation. Let's merge.

PR #6174, opened 2021-11-10
Decomplexify/document result-rendering switches

### LINKED ISSUE #6038: Describe result_renderer kwarg semantics
Presently the (Python API) docs say

```
result_renderer : {'default', 'json', 'json_pp', 'tailored'} or None, optional
  format of return value rendering on stdout. [Default: None]
```

This has issues:

1. It does not mention the `'disabled'` value, which turns result rendering off
2. It does not mention that `None` is NOT turning result rendering off, but will trigger a query of `datalad.api.result-renderer` to determine a result renderer

This is not at all obvious, especially given the fact that `datalad.api.result-renderer` itself is not documented.

I have the suspicion that we will find `result_renderer=None` in the code base, where it should say `result_renderer='disabled'`.

It could be argued that the `datalad.api.result-renderer` config setting should be removed, because we already have enough ambiguity with the `default` result renderer not necessarily being the "default" (because `tailored` might be), and a config setting `datalad.api.result-renderer` configuring another default, which is neither "default", not `default`, but only get's its turn when yet another condition is reached. I am not aware of any use case for this setting either.

It should also be mentioned somewhere that there seems to be support for an `Interface` attribute `result_renderer_cmdline`, which may alter the behavior in ways that I do not understand yet.

Looking further, I see this code in `datalad/cmdline/main.py`:

```py
        if hasattr(cmdlineargs, 'result_renderer'):
            cmdlineargs.result_renderer(ret, cmdlineargs)
```

which will try render results whenever an interface ends up having a `result_renderer` attribute. This seems weird, as a result renderer can only handle a single result, but there might be multiple.

The effective result renderer *on the command line* is determined by this logic:

```py
            # allow commands to override the default, unless something other than
            # default is requested
            kwargs['result_renderer'] = \
                args.common_output_format if args.common_output_format != 'tailored' \
                else getattr(cls, 'result_renderer', 'default')
            if '{' in args.common_output_format:
                # stupid hack, could and should become more powerful
                kwargs['result_renderer'] = DefaultOutputRenderer(args.common_output_format)
```

First, at this point it will not be a surprise that `DefaultOutputRenderer` has nothing to do with `default_result_renderer`, and is in no way a default.

Second, the interface attribute `result_renderer` cannot `None` from the cmdline (rendering `datalad.api.result-renderer` useless, except for internal calls), but it can become any other str value, hence not necessarily a callable. However, there is code in `datalad/interface/utils.py` that will eventually raise a `ValueError('unknown result renderer "{}"'.format(result_renderer))` for any unsopported value.

Ufff...

mih:
The full story is in #6038

Concretely, this:

- documents the 'disabled' renderer mode
- strip supported for the undocumented datalad.api.result-renderer
  config setting
- homogeneous documentation for `--output-format` and `result_renderer`
- rename `default[_result_renderer]` to `generic[_result_renderer` to
  resolve the multi-level ambiguity of "default" for a renderer that,
  ironically, is not the default. Shims for legacy code are in place,
  and will likely never be removed
- replace internal usage of `output_format` with a consistent
  `result_renderer` for homogeneity. Command line API is kept unchanged
  though -- might be worth a follow-up discussion in a separate issue/PR
- discontinue the undocumented use of `result_renderer=None` and replaces
  it with `result_renderer='disabled'`

Fixes #6038

### DIFF
--- a/datalad/cmdline/helpers.py
+++ b/datalad/cmdline/helpers.py
@@ -24,7 +24,10 @@ from textwrap import wrap
 
 from datalad import __version__
 from ..cmd import WitlessRunner as Runner
-from ..interface.common_opts import eval_defaults
+from ..interface.common_opts import (
+    eval_defaults,
+    eval_params,
+)
 from ..log import is_interactive
 from datalad.support.exceptions import CapturedException
 from ..ui.utils import get_console_width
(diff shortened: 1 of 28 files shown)


mih:
This is ready. Will merge EOB today, unless objections are expressed.

bpoldrack (approved):
No objections. :)

PR #6175, opened 2021-11-10
RM: (re)move get_func_kwargs_doc and some python2 compat code path

yarikoptic:
There might be more but largely orthogonal so decided to submit this short quick PR.

It is a companion to https://github.com/datalad/datalad-crawler/pull/113 which I marked for the release and just merged... https://pypi.org/project/datalad-crawler/ has 0.9.3 now

I didn't check other extensions if they use this elderly function -- will see with this PR


mih (approved):
Thanks for digging this up!

PR #6178, opened 2021-11-11
Clean up `datalad.interface`

mih:
- [x] Move commands according to #3192 
- [x] Move tests
- [x] Run `isort` on all moved code
- [x] Place backward compatibility shims. Announce deprecation of old locations
- [x] Introduce sorting of command groups
- [x] Cleanup `datalad/interface/__init__.py`
- [ ] Fix typo
- [ ] Decide on how to deal with the massive breakage on crippled FS. These are not new, but more tests are executed this way due to the move. This makes it more obvious how much of the code base or its tests are still inherently incompatible with adjusted branches on POSIX machines.


mih:
OK, this is not manageable. Too many conflicts, too quickly, to many per-design-broken tests. I need to brake this into smaller pieces (one PR per command/commit in this PR), or the thing will fall apart quicker than I can handle it.

Issue #6186, opened 2021-11-12
`datalad-archives` special remote broken with adjusted branches


mih:
As revealed by https://github.com/datalad/datalad/pull/6178 pretty much all tests fail when executed on a crippled FS. This suggests a fundamental incompatibility of this functionality (or some key part of it) with adjusted branches.

mih:
It seems the `ExtractedArchive` class is to be blamed here. On a standard file system:

```
> /home/mih/hacking/datalad/git/datalad/local/add_archive_content.py(601)__call__()
-> out_json = annex.add_url_to_file(
(Pdb) p extracted_file
'1'
```

Crippled FS:

```
> /home/mih/hacking/datalad/git/datalad/local/add_archive_content.py(601)__call__()
-> out_json = annex.add_url_to_file(
(Pdb) p extracted_file
'MD5E-s60--fb50bc26544b88e2864209d9bb53c3dd'
```

mih:
This is the cause and solution

```diff
diff --git a/datalad/customremotes/archives.py b/datalad/customremotes/archives.py
index 7e754c059..c9613c04d 100644
--- a/datalad/customremotes/archives.py
+++ b/datalad/customremotes/archives.py
@@ -64,7 +64,12 @@ def link_file_load(src, dst, dry_run=False):
 
     try:
         os.link(src_realpath, dst)
-    except AttributeError as e:
+    except (OSError, AttributeError) as e:
+        # we need to catch OSError too, because Python's own logic
+        # of not providing link() where it is known to be unsupported
+        # (e.g. Windows) will not cover scenarios where a particular
+        # filesystem simply does not implement it on an otherwise
+        # sane platform (e.g. exfat on Linux)
         lgr.warning("Linking of %s failed (%s), copying file" % (src, e))
         shutil.copyfile(src_realpath, dst)
         shutil.copystat(src_realpath, dst)
```

Issue #6187, opened 2021-11-12
`AnnexRepo.drop_key(batch=True)` does not drop keys in adjusted mode


mih:
Can be reproduced by executing `datalad.local.tests.test_add_archive_content:TestAddArchiveOptions.test_add_delete_after_and_drop` on a crippled FS.

The test immediately passes with

```diff
diff --git a/datalad/local/add_archive_content.py b/datalad/local/add_archive_content.py
index 25f9d78ac..c0bbb255b 100644
--- a/datalad/local/add_archive_content.py
+++ b/datalad/local/add_archive_content.py
@@ -609,7 +609,7 @@ class AddArchiveContent(Interface):
                     # we need to maintain a list of those to be dropped files
                     if drop_after:
                         # drop extracted files after adding to annex
-                        annex.drop_key(out_json['key'], batch=True)
+                        annex.drop_key(out_json['key'], batch=False)
                         stats.dropped += 1
                     stats.add_annex += 1
                 else:
```

which makes me think it is related to batch-mode operation.

However, when executed, in batch-mode, manually on the command line, it actually drops. So this is not about git-annex, and not about batch-mode in general. It is specifically about batch-mode execution from within DataLad and using its implementation.

PR #6188, opened 2021-11-12
Move AddArchiveContent in the spirit of #3192

mih:
Follow-up to the failed #6178

- [x] Backward-compatibility shim is in place.
- [x] Fixes datalad/datalad#6186
- [x] Make tests run on adjusted mode repos
- [x] Disable test on adjusted mode due to #6187


mih:
Ok, let's go

PR #6176, opened 2021-11-10
ENH/RF: Interface-s - add *  to explicitly separate posargs from kwargs

yarikoptic:
- Sits on top of tiny https://github.com/datalad/datalad/pull/6175 which I thought would be relevant but apparently not really.  Can rebase
- inspired by #5517 when I thought to RF `cmd` into `--cmd` and thus possibly opening the way for ppl to shoot themselves in the foot whenever later I would decide to add `path` pos arg (in CLI etc)
- adding explicit `*` separation would mimic how CLI is specified - you can't specify `--options` as positional arguments.
- may be would give another reason to harmonize interfaces of older interfaces (moved from plugins) -- in most of the cases it is just move of `dataset` from the 0th arg into its due `kwarg` position.
- [x] make sure tests pass
- [x] decide to either issue a DeprecationWarning if an Interface without `*` is encountered: decided not to bother. Now we have some test which sould fail in such case of having some specargs but no kwonly args
- [x] decide what to do if anything about those `TODO*` marked interfaces
  - [x] RFed signature of `clone` to place optional last (REMAINER) posarg to be a kwarg right before `*`
  - [x] RFed signature of `addurls` to have `dataset` as a kwarg as all other interfaces do!
  - [x] RFed other elderlies: add_readme, export_to_figshare, export_archive to also have `dataset` as a kwarg. Rationale: if they are used, most likely via CLI anyways. There it should not matter. Uniformity is good to have
  - [-] Left alone `extract_metadata`  ....
- [x] pay attention to change in the test for `save` where we did allow `message` to be used as posarg (I think - bad idea). But may be do allow for posarg use?  decided against -- no reason to make it different from `git commit -m ...`.
- [ ] (after there is an agreement) Add perspective CHANGELOG.md record to describe this change and breaking changes (if any left).

PR #6190, opened 2021-11-13
RF: de-wrapt

yarikoptic:
A quick chat with @mih inspired me to look into it.... in a nutshell (more to come if feasible):

- use `inspect.signature` instead of `inspect.getargspec` since that one manages to get args of `functools.wraps`ed functions. Source of information: https://stackoverflow.com/questions/147816/preserving-signatures-of-decorated-functions#comment68410839_147878

- ~~remaining gotcha (TEMPorarily disabled that test ) -- not yet "binding" `__doc__` to `Dataset.method` -- `@build_doc` assigns `__doc__` to `__call__` later than `@datasetmethod` is invoked. So I guess `wrapt` does some magic for `__doc__` to be always be queried for the `wrapt`ed functions.~~ -- I think I came up with a good enough workaround

- reworked `getargspec` is based on the one in #6176 but redone on top - will conflict. If we go with this -- this one should take over ;-)

oh well -- let's see if anything else breaks ;-)

### DIFF
--- a/datalad/distribution/dataset.py
+++ b/datalad/distribution/dataset.py
@@ -10,6 +10,7 @@
 """
 
 import logging
+from functools import wraps
 from os.path import (
     curdir,
     exists,
@@ -18,7 +19,6 @@ from os.path import (
     pardir,
 )
 from weakref import WeakValueDictionary
-import wrapt
 
 from datalad import cfg
 from datalad.config import ConfigManager
@@ -447,11 +447,12 @@ def datasetmethod(f, name=None, dataset_argname='dataset'):
 
     The decorator has no effect on the actual function decorated with it.
     """
+
     if not name:
         name = f.__name__
 
-    @wrapt.decorator
-    def apply_func(wrapped, instance, args, kwargs):
+    @wraps(f)
+    def apply_func(instance, *args, **kwargs):
         # Wrapper function to assign arguments of the bound function to
         # original function.
         #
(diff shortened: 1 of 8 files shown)


yarikoptic:
Woohoo, not yet sure what docs are complaining about, but the rest seems to be happy, so it might be viable

mih:
YES! I knew it. This is really cool. Thanks for taking it on!

yarikoptic:
Ha - that test said fails on Windows and Mac so may be it wasn't really wrapt to blame!

```
ERROR: datalad.runner.tests.test_nonasyncrunner.test_popen_invocation
    self.test(*self.arg)
  File "c:\projects\datalad\datalad\tests\utils.py", line 856, in _wrap_with_tempfile
  File "c:\projects\datalad\datalad\tests\utils.py", line 856, in _wrap_with_tempfile
  File "c:\projects\datalad\datalad\runner\tests\test_nonasyncrunner.py", line 180, in test_popen_invocation
    fetching_data.start()
  File "C:\Python39-x64\lib\multiprocessing\process.py", line 121, in start
    self._popen = self._Popen(self)
  File "C:\Python39-x64\lib\multiprocessing\context.py", line 224, in _Popen
    return _default_context.get_context().Process._Popen(process_obj)
  File "C:\Python39-x64\lib\multiprocessing\context.py", line 327, in _Popen
    return Popen(process_obj)
  File "C:\Python39-x64\lib\multiprocessing\popen_spawn_win32.py", line 93, in __init__
    reduction.dump(process_obj, to_child)
  File "C:\Python39-x64\lib\multiprocessing\reduction.py", line 60, in dump
    ForkingPickler(file, protocol).dump(obj)
TypeError: cannot pickle '_thread.lock' object
```

yarikoptic:
ok, dropped that last TEMP optimistic commit (so test will remain skipped on OSX and windows).  Fixed up docstring so it builds.  Should be ready for tries/review/ and even merge unless some oddity detected

mih (approved):
I think this should fly. Let's try it it! Thx!

PR #6191, opened 2021-11-14
Move Clean in the spirit of #3192

mih:

Follow-up to the failed #6178


mih:
FTR: the fact that the `clean` command is not tested well is not a problem introduced by this PR.

PR #6192, opened 2021-11-14
Move Unlock in the spirit of #3192

mih:

Follow-up to the failed #6178


mih:
OK, good to go -- with more coverage now.

PR #6166, opened 2021-11-09
MNT/Windows: More modern replacement for appdirs under Windows

adswa:
WIP

(This is ugly and just pushed/PRed for @yarikoptic to see, please ignore it)

PR #6198, opened 2021-11-16
MNT: Replace appdirs with platformdirs

### LINKED ISSUE #6164: Stop using Appdirs

adswa:
The original appdirs repository is unmaintained. This change replaces it
with a currently maintained fork of the project. Closes #6164, and for now replaces #6166

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -134,7 +134,7 @@ since we use it to provide backports of recent fixed external modules we depend 
 
 ```sh
 apt-get install -y -q git git-annex-standalone
-apt-get install -y -q patool python3-scrapy python3-{appdirs,argcomplete,git,humanize,keyring,lxml,msgpack,progressbar,requests,setuptools}
+apt-get install -y -q patool python3-scrapy python3-{argcomplete,git,humanize,keyring,lxml,msgpack,progressbar,requests,setuptools}
 ```
 
 and additionally, for development we suggest to use tox and new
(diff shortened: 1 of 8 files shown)


adswa:
rebased to ``master`` and added ``semver-minor`` label as per Bens suggestion

mih (approved):
Thx!

PR #6200, opened 2021-11-16
TST: redistribute windows tests across workers to harmonize runtime

### LINKED ISSUE #6193: Rebalance runtime of windows CI runs on appveyor
image

Almost 20min difference between the fastest and the slowest. This is the result of running more tests, and moving tests the other locations in this codebase.

adswa:
This is an attempt to fix #6193, but we will have to take a look at the actual runtimes with this change

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -105,6 +105,7 @@ environment:
           datalad.cmdline
           datalad.customremotes
           datalad.distribution
+          datalad.distributed
           datalad.downloaders
           datalad.interface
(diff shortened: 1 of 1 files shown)


adswa:
I have cancelled all CIs except for Appveyor, which should be the only thing affected by this change.

adswa:
Screenshot from 2021-11-16 13-30-41

moving one more test around to improve this further

mih (approved):
Thanks for doing this!

PR #6203, opened 2021-11-17
Document `BatchedCommand` and `BatchedAnnex`

christian-monch:
This PR adds design documentation for  `BatchedCommand` and `BatchedAnnex`, as well as a very basic test for `BatchedCommand`. This PR is a fallout of my work on implementing  `BatchedCommand` and `BatchedAnnex` with the new generator-based runner.  Although this PR might be outdated in the next or a later release, I think it provides some value.

- [X] Add documentation for `BatchedCommand` and `BatchedAnnex`
- [X] Add a basic test

### DIFF
--- a/datalad/tests/test_cmd.py
+++ b/datalad/tests/test_cmd.py
@@ -8,10 +8,15 @@
 # ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
 """Test WitlessRunner
 """
-
+import sys
 import unittest.mock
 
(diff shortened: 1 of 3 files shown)


mih (approved):
God, yes! Thank you!

mih:
Maybe a little side-note that I got initially confused by the terminology. "command" is not the thing that is "batched" (as in batchedCommand). When the term "command" is used, it is one of the `stdin` items that are passed to the running command-batch-process. But I am not sure how to do this better. It is not an argument either.

Issue #6210, opened 2021-11-18
Python 3.10/Windows: Formatters throws uncaught ModuleNotFoundError


adswa:
Running the tests under Windows 10 with a fresh Python 3.10 and an editable install of DataLad yields a ModuleNotFoundError, but the code only catches an ImportError. Thus, the complete testsuite crashes

```
C:\Users\adina\repos\datalad>python -m nose -s -v --traverse-namespace datalad
Traceback (most recent call last):
  File "C:\Users\adina\repos\datalad\datalad\cmdline\tests\test_formatters.py", line 21, in <module>
    import formatters as fmt
ModuleNotFoundError: No module named 'formatters'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Users\adina\AppData\Local\Programs\Python\Python310\lib\site-packages\nose\loader.py", line 417, in loadTestsFromName
    module = self.importer.importFromPath(
  File "C:\Users\adina\AppData\Local\Programs\Python\Python310\lib\site-packages\nose\importer.py", line 47, in importFromPath
    return self.importFromDir(dir_path, fqname)
  File "C:\Users\adina\AppData\Local\Programs\Python\Python310\lib\site-packages\nose\importer.py", line 94, in importFromDir
    mod = load_module(part_fqname, fh, filename, desc)
  File "C:\Users\adina\AppData\Local\Programs\Python\Python310\lib\imp.py", line 235, in load_module
    return load_source(name, filename, file)
  File "C:\Users\adina\AppData\Local\Programs\Python\Python310\lib\imp.py", line 172, in load_source
    module = _load(spec)
  File "<frozen importlib._bootstrap>", line 719, in _load
  File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 883, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "C:\Users\adina\repos\datalad\datalad\cmdline\tests\test_formatters.py", line 27, in <module>
    raise SkipTest
unittest.case.SkipTest

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Users\adina\AppData\Local\Programs\Python\Python310\lib\runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\Users\adina\AppData\Local\Programs\Python\Python310\lib\runpy.py", line 86, in _run_code
[...]

PR #6212, opened 2021-11-18
BF: Also catch ModuleNotFoundErrors for formatters

adswa:
A quick fix for #6210.

### DIFF
--- a/datalad/cmdline/tests/test_formatters.py
+++ b/datalad/cmdline/tests/test_formatters.py
@@ -16,15 +16,11 @@ from datalad.tests.utils import (
     assert_in,
     assert_not_in,
     SkipTest,
+    skip_if_no_module
 )
-try:
-    import formatters as fmt
-except ImportError:  # pragma: no cover
-    # must be running from installed version where formatters is not present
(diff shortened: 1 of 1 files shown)


datalad/cmdline/tests/test_formatters.py
```diff
@@ -19,7 +19,7 @@
 )
 try:
     import formatters as fmt
-except ImportError:  # pragma: no cover
+except ImportError or ModuleNotFoundError:  # pragma: no cover
```
mih, line 22: Rather than that, the entire `try-except` block could become:

```
skip_if_no_module('formatters')
```

see for example `datalad/downloaders/tests/test_s3.py` for a similar usage.

mih (changes requested):
Thanks, I would suggest a different approach for more homogeneity with other cases like this, though.

datalad/cmdline/tests/test_formatters.py
adswa: thanks!

mih (approved):
Thx!

PR #6213, opened 2021-11-18
API mode indicator and library-mode enabling switch

### LINKED ISSUE #6081: Application vs library use distinction (also Python vs Cmdline)
Seeing https://github.com/datalad/datalad/pull/6078#issuecomment-944040232 reminded me, that we had several cases already where the information whether datalad was called via CLI or python would be desirable.
Currently that is not assessible from within most of the code base. However, the outer layer in `interface`, the arg parsing, etc. do actually have that information. We could simply set a module attribute in `datalad`, `datalad.api` or may be `datalad.interface` to store a flag, that can then be used by code deeper down.

WDYT, @datalad/developers ?

mih:
This PR lays the foundation to introduce "library-mode" behavior, and to distinguish via which API datalad functionality is being accessed. These aspects are related but orthogonal.

Summary:

- Added `datalad.get_apimode()` to query for the effective API in use
- Added `datalad.enable_librarymode()` to put DataLad in lib-mode
- Added `datalad.in_librarymode()` to test for active lib-mode
- Added design document on this

This is the foundation to further tailor messaging, logging behavior, etc. None of which is planned for this PR.

Additional changes:

- Unit tests for the mode queries (limited to deviation from the default)
- Disabled Python API doc generation in lib-mode (shaves off 150ms process startup time on my machine)

- Fixes #6081

### DIFF
--- a/datalad/__init__.py
+++ b/datalad/__init__.py
@@ -25,10 +25,6 @@ import os
 
 # this is not to be modified. for querying use get_apimode()
 __api = 'python'
-# this is not to be modified. see enable/in_librarymode()
-# by default, we are in application-mode, simply because most of
-# datalad was originally implemented with this scenario assumption
-__runtime_mode = 'application'
 
 
 def get_apimode():
@@ -83,6 +79,8 @@ def enable_librarymode():
     """
     global __runtime_mode
     __runtime_mode = 'library'
+    # export into the environment for child processes to inherit
+    os.environ['DATALAD_RUNTIME_LIBRARYMODE'] = '1'
 
 
 # For reproducible demos/tests
@@ -112,6 +110,13 @@ from .config import ConfigManager
 cfg = ConfigManager()
 
 # must come after config manager
+# this is not to be modified. see enable/in_librarymode()
+# by default, we are in application-mode, simply because most of
+# datalad was originally implemented with this scenario assumption
+__runtime_mode = 'library' \
+    if cfg.getbool('datalad.runtime', 'librarymode', False) \
+    else 'application'
+
 from .log import lgr
 from datalad.support.exceptions import CapturedException
 from datalad.utils import (
--- a/datalad/__init__.py
+++ b/datalad/__init__.py
@@ -23,6 +23,28 @@ import atexit
 import os
 
 
+# this is not to be modified. for querying use get_apimode()
+__api = 'python'
+
+
+def get_apimode():
+    """Returns the API mode label for the current session.
+
+    The API mode label indicates whether DataLad is running in "normal"
+    mode in a Python session, or whether it is used via the command line
+    interface.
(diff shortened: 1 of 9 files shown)


datalad/__init__.py
```diff
@@ -19,6 +19,68 @@
...
+    Library mode can only be enabled once. Switching it on and off within
+    the runtime of a process is not supported.
+
+    Example::
+
+        >>> import datalad
+        >>> datalad.enable_librarymode()
+    """
+    global __runtime_mode
+    __runtime_mode = 'library'
```
yarikoptic, line 81: I worry a little of placing "library" etc into function names .
I wonder if it wouldn't be better to encapsulate it in a helper class (e.g. `_DataLadMode`) with `.current`, `.supported` (returns available -- we might need to add later some other, e.g. `jupyter-shell`, `web-app` although since it relates to `ui`s - not yet sure if separation would be clear since that is where we also we might need to adjust UI - e.g. use jupyter widgets for progress bars), `.switch_to(mode: str)` and have an instance `datalad.mode` exposing it? Could also have `__eq__` overloaded to implement `self.current == other`.

mih: I am not confident I get your argument properly.

> I worry a little of placing "library" etc into function names

I am OK with a different name, which name would you suggest?

> I wonder if it wouldn't be better to encapsulate it in a helper class

I proposed three simple functions to be able to have maximum freedom for such changes, once they turn out to be necessary. They can *internally* do what you proposed.

However, so far the concept is only developed to the degree that we know we need to distinguish an API mode (with a presently fixed set of two possibilities, without preventing future extension) and standalone vs library-mode (whatever labels we end up using for them). We have not yet talked about any mode switching, additional modes, etc. If we need to add them having a `in_librarymode()` helper is not an obstacle.

All this would have implications on how to test for such conditions throughout the code-base (and other people's code). I feel confident to be able to support these three functions indefinitely. It is not clear to me whether you have maintenance concerns. From my reading I'd say that everything you write can be done in the future on top of these changes. But if I am missing something here, we likely need to close this PR again, and work on a better concept. I only started implementing this, after going through the issues and finding that we did not come up with additional needs for some years.

bpoldrack:
I like it. As far as I'm concerned: Let's get some experience with this before (if at all) developing something more complex.

mih:
Conclusion from call: Replace the entire API with config manager calls. Do inheritance to subprocesses via ENV variables.

mih:
Despite the conclusion from the call yesterday, I did not base the facility on the configmanager entirely. It caused too many ripples, import dependencies and expensive checks. That being said, the cmdline switch `--library-mode` is gone (no longer needed).

mih:
Resolved conflict. Will merge when tests are done.

Issue #3832, opened 2019-10-24
Inconsistent argument update -s/--sibling


mih:
Other commands (`siblings`, `create-sibling-gitlab`) use `-s/--name`.

yarikoptic:
Would be nice to unify. 

There is also related overlap in `--to`: `--to=REVISION` being used by `diff`, while `publish` and now `push` use `--to=sibling` .  I have not researched yet but I would not be surprised if we would like to get `--to=REVISION` for publish/push to publish specific ranges of changes instead of the overall state.

I somewhat dislike having `-s` but `--name`, and even though tautological for those commands having `--sibling` would be consistent with other commands. Otherwise I need to remember to use short version (`-s`) to not use an incorrect long one.

yarikoptic:
FWIW, I grew to like (actually I think the API grew to be that as well) conda's consistent use of `-n` to describe name of the environment.  Besides `conda activate NAME` all other commands IIRC ([chatsheet](https://docs.conda.io/projects/conda/en/4.6.0/_downloads/52a95608c49671267e40c689e0bc00ca/conda-cheatsheet.pdf)) have it as `-n|--name NAME` so I do not need to think twice what it could be (`-e vs -n`) or positional or an option. I feel that ideally we should unify similarly, that it is always `-s NAME` in some cases even if it iseems to be a good idea to make it a positional argument.  In tricky cases (I am not sure if we filed an issue) where we might eventually want something like `push  --from X --to Y` it could still be `push --from|--from-sibling X -s Y`.

jsheunis:
@yarikoptic @mih I'm looking into this. So do we want to purely change the argument for `update` to `-s/--name` ([here](https://github.com/datalad/datalad/blob/b3acf7ca3dadeacc15aa6308d4d51ff7a973011b/datalad/distribution/update.py#L131)) or do we want to reconsider the whole `-s/--name` vs `-n/--name`. In the latter case, should that be a separate issue? In the former, should I make the PR against the `maint` branch?

yarikoptic:
no options rename should be for `maint` .  While at it we might want to go for uniform `-s/--sibling-name` ?

jsheunis:
Is that a question or a suggestion to implement? I don't have a specific preference and would defer to datalad elders on this :)

I've done searches for `--name` and `--sibling` and looks like changes will need to be made in these modules:
- `update`
- `siblings`
- `create_sibling`
- `create_sibling_github`
- `create_sibling_gitlab`
[...]

PR #6214, opened 2021-11-18
Design document on standard parameters

adswa:
A start on a design document about standard parameters following #3832. It documents the current status quo, but adds the suggestion to use ``-s``/``--sibling-name`` raised by @yarikoptic in #3832.


mih (approved):
LGTM, thx!

mih:
It seems there are no other comments.

PR #6215, opened 2021-11-19
MNT: Harmonize sibling parameter to -s/--sibling-name

adswa:
This PR implements the suggestion in #3832 and the proposed design document in #6214 by harmonizing the ``name`` parameter for dataset siblings to be ``-s``/``--sibling-name``. Let's see what breaks with this.

### DIFF
--- a/datalad/distribution/create_sibling.py
+++ b/datalad/distribution/create_sibling.py
@@ -452,7 +452,7 @@ class CreateSibling(Interface):
                 URL and path on the server.""",
             constraints=EnsureStr()),
         name=Parameter(
-            args=('-s', '--name',),
+            args=('-s', '--sibling-name',),
             metavar='NAME',
             doc="""sibling name to create for this publication target.
                 If `recursive` is set, the same name will be used to label all
--- a/datalad/distribution/update.py
+++ b/datalad/distribution/update.py
@@ -129,14 +129,20 @@ class Update(Interface):
             operation.""",
             nargs="*",
             constraints=EnsureStr() | EnsureNone()),
-        sibling=Parameter(
-            args=("-s", "--sibling",),
+        name=Parameter(
+            args=("-s", "--sibling-name",),
             doc="""name of the sibling to update from. When unspecified,
             updates from all siblings are fetched. If there is more than one
             sibling and changes will be brought into the working tree (as
             requested via [CMD: --merge, --how, or --how-subds CMD][PY:
             `merge`, `how`, or `how_subds` PY]), a sibling will be chosen based
             on the configured remote for the current branch.""",
+            constraints=EnsureStr() | EnsureNone()),
+        sibling=Parameter(
+            args=("--sibling",),
+            doc="""DEPRECATED, will be removed in a future release.
+            Please use  the renamed parameter [CMD: -s/--sibling-name CMD][PY:
+            name PY] instead""",
             constraints=EnsureStr() | EnsureNone()),
         dataset=Parameter(
             args=("-d", "--dataset"),
@@ -215,6 +221,7 @@ class Update(Interface):
     def __call__(
             path=None,
             sibling=None,
+            name=None,
             merge=False,
             how=None,
             how_subds=None,
@@ -226,6 +233,13 @@ class Update(Interface):
             reobtain_data=False):
         if fetch_all is not None:
             lgr.warning('update(fetch_all=...) called. Option has no effect, and will be removed')
+        if sibling is not None:
+            import warnings
+            warnings.warn("The --sibling parameter has been renamed to -s/"
+                          "--sibling-name.",
+                          DeprecationWarning)
+        else:
+            sibling = name
         if path and not recursive:
             lgr.warning('path constraints for subdataset updates ignored, '
                         'because `recursive` option was not given')
--- a/datalad/distributed/create_sibling_gitlab.py
+++ b/datalad/distributed/create_sibling_gitlab.py
@@ -111,7 +111,7 @@ class CreateSiblingGitlab(Interface):
         Name of the default GitLab site (see --site)
     "datalad.gitlab-SITENAME-siblingname"
         Name of the sibling configured for the local dataset that points
-        to the GitLab instance SITENAME (see --name)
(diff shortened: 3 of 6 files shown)


adswa:
(side question: Why does the semver check forbid ``minor`` labels for ``maint``?)

adswa:
Oh there is a flaw in this PR. I'm harmonizing the parameter name for the CLI, but now there is a mismatch between CLI and Python API...

datalad/distribution/create_sibling.py
```diff
@@ -452,7 +452,7 @@ class CreateSibling(Interface):
                 URL and path on the server.""",
             constraints=EnsureStr()),
         name=Parameter(
-            args=('-s', '--name',),
+            args=('-s', '--sibling-name',),
```
yarikoptic, line 455: no deprecation cycle at all? (here and elsewhere)

I would vote for keeping both `--name` and `--sibling-name` **but** is there any sensible way to identify that it was used in the CLI...

also if we are renaming CLI arg, ideally the name of the Python API should also change (and also deprecation cycle) since otherwise we would have dichotomy (`--sibling-name` in CLI but python code needs `name`)

Having both of those in mind, I think we could

- keep original `name` and remove `-s` from it
- add new `sibling_name` with `-s`, `--sibling-name`
- function would gain both `name` and `sibling_name`.  If `name` is specified explicitly -- we issue a deprecation warning and assign (after checking) to `sibling_name`.  I believe we already do similar dances elsewhere.  Once again -- IMHO worth centralizing somehow

datalad/distribution/update.py
```diff
@@ -226,6 +233,13 @@ def __call__(
             reobtain_data=False):
         if fetch_all is not None:
             lgr.warning('update(fetch_all=...) called. Option has no effect, and will be removed')
+        if sibling is not None:
```
yarikoptic, line 236: ideally, for those deprecated we should use some dedicated (not just `None`) singleton... in #5517 I introduced NoneDeprecated, check it out.  Then even if someone used explicitly `sibling=None` for some obnoxious reason, we would "detect" that

datalad/distribution/update.py
```diff
@@ -226,6 +233,13 @@ def __call__(
...
+            import warnings
+            warnings.warn("The --sibling parameter has been renamed to -s/"
+                          "--sibling-name.",
```
yarikoptic, line 239: I believe similar question came, and the context is related to #6213 -- `--sibling` is how we describe it in CLI, whenever in Python API would be `sibling`.

<details>
<summary>we don't have yet consistency (a sample grep showing another effect of absent centralized helper - duplication)</summary> 

```shell
datalad/core/local/status.py-            warnings.warn(
datalad/core/local/status.py-                "status(repor_filetype=) no longer supported, and will be removed "
datalad/core/local/status.py-                "in a future release",
datalad/core/local/status.py:                DeprecationWarning)
--
datalad/distributed/create_sibling_github.py-                "datalad-create-sibling-github's `dryrun` option is "
datalad/distributed/create_sibling_github.py-                "deprecated and will be removed in a future release, "
datalad/distributed/create_sibling_github.py-                "use the renamed `dry_run/--dry-run` option instead.",
datalad/distributed/create_sibling_github.py:                DeprecationWarning)
--
datalad/distributed/create_sibling_github.py-                "datalad-create-sibling-github's `github_login` option is "
datalad/distributed/create_sibling_github.py-                "deprecated and will be removed in a future release, "
datalad/distributed/create_sibling_github.py-                "use the `credential` option instead.",
datalad/distributed/create_sibling_github.py:                DeprecationWarning)
--
datalad/distributed/create_sibling_github.py-                "datalad-create-sibling-github's `github_organization` "
datalad/distributed/create_sibling_github.py-                "option is deprecated and will be removed in a future "
datalad/distributed/create_sibling_github.py-                "release, prefix the reposity name with `<org>/` instead.",
datalad/distributed/create_sibling_github.py:                DeprecationWarning)
--
datalad/distributed/create_sibling_gitlab.py-                "datalad-create-sibling-github's `dryrun` option is "
datalad/distributed/create_sibling_gitlab.py-                "deprecated and will be removed in a future release, "
datalad/distributed/create_sibling_gitlab.py-                "use the renamed `dry_run/--dry-run` option instead.",
datalad/distributed/create_sibling_gitlab.py:                DeprecationWarning)
--
datalad/distributed/create_sibling_ria.py-            import warnings
datalad/distributed/create_sibling_ria.py-            warnings.warn("datalad-create-sibling-ria --no-storage-sibling "
datalad/distributed/create_sibling_ria.py-                          "is deprecated, use --storage-sibling off instead.",
datalad/distributed/create_sibling_ria.py:                          DeprecationWarning)
--
datalad/distributed/drop.py-            warnings.warn(
datalad/distributed/drop.py-                "The `check` argument of `datalad drop` is deprecated, "
datalad/distributed/drop.py-                "use the `reckless` argument instead.",
datalad/distributed/drop.py:                DeprecationWarning)
--
datalad/distributed/drop.py-                "The `if_dirty` argument of `datalad drop` is ignored, "
datalad/distributed/drop.py-                "it can be removed for a safe-by-default behavior. For "
datalad/distributed/drop.py-                "other cases consider the `reckless` argument.",
datalad/distributed/drop.py:                DeprecationWarning)

```
</details>

so may be following the `create_sibling*` trend, make it `sibling/--sibling` and `sibling_name/--sibling-name` for now?

mih: I made a concrete proposal for a uniform specification in https://github.com/datalad/datalad/pull/6249

It also implements support for API-specific doc formatting for it. I went for `||` instead of `/`, because the latter is sometimes used for alternatives within an API (e.g. long v short), and I went for `,` to separate long v short, because that is what the manpages do.

datalad/distribution/update.py
mih: The `NoneDeprecated` type is now actually merged and available.

mih:
This PR is stalled. I am planning to archive it before the holidays, unless activity resumes.

mih:
Closing as announced a month ago.

PR #6216, opened 2021-11-19
NF: Teach clone and install to name siblings other names than origin

### LINKED ISSUE #5320: install/clone: add -s/--name
re option name: we are yet to unify, e.g. see https://github.com/datalad/datalad/issues/3832

I guess it is not a frequent use case but I did run into it before and again now: we have no way to specify the name of the sibling (other than default "origin") while clone/installing a dataset.  `git clone` has `-o|--origin <name>`

adswa:
This is an attempt to implement #5320: Giving ``clone`` and `install` the ability to custom name the sibling it was cloned from. It is implemented as a new common ``-o``/``--origin`` parameter, and uses Git's own ``-o``/``--origin`` parameter under the hood.
I stayed consistent with Gits naming scheme, and referenced the configuration ``clone.defaultRemoteName`` (Git version 2.30 and up) in the parameter docstring as well.

Fixes #5320

### DIFF
--- a/datalad/core/distributed/clone.py
+++ b/datalad/core/distributed/clone.py
@@ -22,6 +22,7 @@ from datalad.interface.results import get_status_dict
 from datalad.interface.common_opts import (
     location_description,
     reckless_opt,
+    origin_opt,
 )
 from datalad.log import log_progress
 from datalad.support.gitrepo import (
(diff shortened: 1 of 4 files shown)


adswa:
closing in favor of #6218

PR #6218, opened 2021-11-19
`clone` supports passing options to `git-clone` (Replacement of #5662)

### LINKED ISSUE #4005: get() should rename 'origin' of a subdataset to match superds remote is was cloned from
Aspect of original #1956 reported by @mvdoc on behalf of @yarikoptic 

If a subdataset is found to be available as a subdataset underneath a superdataset's clone/remote, and ends up being installed from that location, its origin remote should match the one of the superdataset's remote.

Implementing such a behavior will counteract the fix for https://github.com/datalad/datalad/issues/3926 in #3966

### LINKED ISSUE #5286: `clone` must support optimization arguments
I have a case where the source dataset contains MANY branches. Cloning it takes forever. With `git clone` I can to `--singlebranch` or `--depth 1`. With DataLad I am forced to wait.

### LINKED ISSUE #5320: install/clone: add -s/--name

### LINKED ISSUE #5586: Missing ability to tweak the clone process
I am doing a twisted mass-rebase of several 10k branches. This involves re-annexing 100k files (i.e. change their content and put it back into the annex). This starts with a fresh clone:

```
datalad clone --reckless ephemeral ria+file:///...ce3f3fffe fixup
```

I need the `ephemeral` for efficiency reasons.

Now I am facing this problem:

```
% git commit -a --amend --no-edit
error: invalid object 100644 91e15dc904225126970a8cb35912e5b0db053875 for '000/021/MD5E-s53
32849--bae29652c989fa1234daf4b7c44d1360.tar.gz.log'                                       
fatal: git-write-tree: error building trees                                               
git-annex: failed to read sha from git write-tree                                          
CallStack (from HasCallStack):                                                            
  error, called at ./Git/Sha.hs:23:15 in main:Git.Sha                                     
```

I am not sure where this is coming from, but I suspect it is happening, because I have more than one clone like this, and maybe the sharing of git objects (default with local clones) breaks the operation due to undesired interactions.

Anyways, this particular issue points to a more general issue: there is no way to customize the `clone` call (e.g. with `--no-local`, but there are more that can have relevance in special cases).

Option specification seems to be implemented all the way up to `clone_dataset()`, so it is accessible for internal use, but not in general.

I do not necessarily see the need to expand the API, but maybe a config option could be useful.

mih:
This substitutes #5662, because I cannot push to Kyle's fork to resolve conflicts. Please see that PR for extensive information on the motivation for this change. Beyond the original purpose "Support cloning a specified branch or tag" it does:

- Makes #6216 obsolete, but still fixes #5320
- Exposes `git clone`'s optimization abilities and fixes #5286
- Fixes #5586
- While it doesn't fully address #5835, it makes it possible for users to not have to wait for a resolution: `datalad clone some private --config annex.private=true`, with a similar approach now working for any potential setting that should become effective *before* `git annex init` runs
- It can be used internally to trivially fix #4005


bpoldrack:
Going through the original discussion, stumbled upon this: https://github.com/datalad/datalad/pull/5662#issuecomment-850451548 by @mih 

> If there is need for homogenization, we could investigate to strip the support for embedded versions from the RIA code. WDYT @bpoldrack ?

There's one potential issue I can see with stripping support for this: By now we store the originally given URL in `.gitmodules` under `datalad-url`. This would be the complete RIA URL in such a case, hence stripping support would invalidate existing entries in datasets. That's not necessarily an issue in the sense that we wouldn't know what to clone (git knows the commit of a submodule), but we must not fail on parsing such a RIA URL.


Other than that: LGTM.

mih:
This PR does not do stripping, and I have no plans to do such thing without a dedicated issue and discussion. Promised!

mih:
How about this (going from https://github.com/datalad/datalad/issues/5286#issuecomment-756210908)

What if we execute the proposed

```
git fetch --depth 1 origin +refs/heads/git-annex:refs/remotes/origin/git-annex
```

whenever `git_clone_opts` where provided, but no `git-annex` branch manifested locally after a clone? This would make it possible to use `--depth 1` with `datalad clone` and make for pretty speedy clones for large datasets.

PR #6220, opened 2021-11-20
Move ReRun in the spirit of #3192

mih:

Follow-up to the failed #6178


mih:
I will not wait for 6 hours to let the metalad tests stall out in peace ;-)

PR #6223, opened 2021-11-22
Cleanup interface list, introduce group sort, new group structure

mih:
The new structure accents on the API structure proposal in #3192 

This PR concludes the recovery from the failed #6178


mih:
Killed the metalad tests after 3h

mih:
Unrelated test errors

Issue #6209, opened 2021-11-18
`create-sibling-gin` does not yield setup that is immediately usable by `push`


mih:
Previously  described in https://github.com/datalad/datalad/issues/6204#issuecomment-971869080

Demo:

```sh
# create a dataset and a GIN sibling
(datalad-dev) mih@meiner /tmp % datalad create ginny
[INFO   ] Creating a new annex repo at /tmp/ginny 
create(ok): /tmp/ginny (dataset)
(datalad-dev) mih@meiner /tmp % cd ginny
(datalad-dev) mih@meiner /tmp/ginny (git)-[master] % DATALAD_CREDENTIAL_GIN_TOKEN=... datalad create-sibling-gin ginny --credential gin
create_sibling_gin(ok): [sibling repository 'gin' created at https://gin.g-node.org/mih/ginny]
configure-sibling(ok): . (sibling)
action summary:
  configure-sibling (ok: 1)
  create_sibling_gin (ok: 1)

# add payload
(datalad-dev) mih@meiner /tmp/ginny (git)-[master] % echo 123 > file
(datalad-dev) mih@meiner /tmp/ginny (git)-[master] % datalad save
add(ok): file (file)                                                                                                        
save(ok): . (dataset)                                                                                                       
action summary:                                                                                                             
  add (ok: 1)
  save (ok: 1)

# ISSUE: payload is not pushed (just commits)
(datalad-dev) mih@meiner /tmp/ginny (git)-[master] % datalad push --to gin
publish(ok): . (dataset) [refs/heads/master->gin:refs/heads/master [new branch]]                                            
publish(ok): . (dataset) [refs/heads/git-annex->gin:refs/heads/git-annex [new branch]]                                      
action summary:                                                                                                             
  publish (ok: 2)
(datalad-dev) mih@meiner /tmp/ginny (git)-[master] % git annex whereis file
whereis file (1 copy) 
        2e209a3f-fc2a-405f-b1ca-7fb0fe9c46cd -- mih@meiner:/tmp/ginny [here]
ok

# underlying cause: annex-ignore
(datalad-dev) mih@meiner /tmp/ginny (git)-[master] % git config --list | grep gin
remote.gin.annex-ignore=true
remote.gin.url=https://gin.g-node.org/mih/ginny
remote.gin.fetch=+refs/heads/*:refs/remotes/gin/*
remote.gin.annexurl=git@gin.g-node.org:/mih/ginny.git
[...]

Issue #6204, opened 2021-11-17
Investigating type git special remotes


bpoldrack:
I am investigating our suggested usage of type git special remotes in tests, help pages (like `create-sibling-gin`) and the handbook.
This relates to https://github.com/datalad-handbook/book/issues/779, https://github.com/datalad/datalad/issues/6199, https://github.com/datalad/datalad/issues/5144, https://github.com/datalad/datalad/issues/6195 and also somewhat https://github.com/datalad/datalad/issues/6184.

I'll use this issue to keep track of what I discover.

1. To correctly `initremote` such a special remote the respective git remote needs to already be an accessible, valid annex with a uuid. Otherwise the special remote will end up with no uuid and is unusable.

2. That invalidates the description in the handbook (https://handbook.datalad.org/en/latest/basics/101-139-gin.html#using-gin-as-a-data-source-behind-the-scenes), since it suggests to add a freshly created repo on GIN with `siblings add ... -as-common-datasource`, but such a new repo is a plain git repo. In addition that description doesn't account for the fact, the subsequent push needs to deal with unrelated histories.

3. The test `distribution.tests.test_siblings.test_as_common_datasource` is partially broken, since it creates two such special remotes, one of which is invalid, not autoenabled and not used anywhere in the test. The clone on which the `get` is tested has this:
```
❱ git cat-file blob "git-annex:remote.log"
 autoenable=true location=http://127.0.0.1:34537/ name=fresh-sr type=git timestamp=1637150426.567013487s
0116253d-d1a2-458b-b6ba-e69180149a2b autoenable=true location=http://127.0.0.1:34537/.git name=mike2 type=git timestamp=1637150360.267307216s
```
Hence, `fresh-sr` is invalid (no uuid) and `mike2` works correctly and is in fact autoenabled. Note, that the two end up using different URLs (once with and once w/o `.git`)

4. `git-annex-init` doesn't complain about such autoenabled special remote w/o uuid. It simply doesn't do anything. A manual `enableremote` afterwards reveals that annex thinks there's no special remote.

bpoldrack:
One more confusing thing about that test (`distribution.tests.test_siblings.test_as_common_datasource`).
There's this line cloning a dataset: https://github.com/datalad/datalad/blob/02f6914ba856333f00ab26af97b6e51fb8792104/datalad/distribution/tests/test_siblings.py#L542

Dropping into debugger right there, gives me:
```
(Pdb) p url
'http://127.0.0.1:44847/'
```
[...]

PR #6230, opened 2021-11-22
Make outcome of `create-sibling-gin` readily usable for a `push` data transfer

mih:
This change corrects and overrules a premature conclusion by git-annex on the non-availability of an annex at the GIN platform -- which is merely an effect of happening prior any branch availability on GIN.

I decided to not pursue the alternative approach from https://github.com/datalad/datalad/issues/6204#issuecomment-971869080 which could potentially be more expensive with no clear advantage.

Fixes datalad/datalad#6209


bpoldrack (approved):

mih:
Thx!

PR #6242, opened 2021-11-24
Fix special remote messaging

### LINKED ISSUE #5773: S3 credential issue not communicated and breaks special remote protocol in master
This is similar to #5469, but a different issue.

I have (intentionally) broken S3 credentials in my keyring. Now trying to download a file from the HCP dataset

```
% datalad get HCP1200/143426/T1w/T1w_acpc_dc.nii.gz
get(error): HCP1200/143426/T1w/T1w_acpc_dc.nii.gz (file) [TRANSFER failed with no reason given                             
TRANSFER failed with no reason given
TRANSFER failed with no reason given]
action summary:
  get (error: 1, notneeded: 1)
```

Because I created the invalid credentials, I no what the reason is. However, in the general case it seems unresonable to not be able to tell that a download failed because of invalid authentication. Even turning on `-l debug` does not reveal more information.

### LINKED ISSUE #6231: ORA initremote cryptic and confusing (and error messaging in general)
```
% git annex initremote ora1 encryption=none type=external externaltype=ora url=ria+ssh://localhost/tmp/store 
initremote ora1 
  Remote doesn't report any dataset tree version. Consider upgrading datalad or add a fitting 'ria-layout-version' file at the RIA store's root.Treating remote as read-only in order toprevent damage by putting things into an unknown version of the target layout. You can overrule this by setting 'annex.ora-remote.<name>.force-write=true'.

git-annex: INITREMOTE failed with no reason given
failed
initremote: 1 failed
```

It makes a nebulous statement about upgrading datalad which can sent people off on a long journey, where it is unclear what the rational for it would be. Also there is no remote at this point that could be considered read or write. It simply fails. `--debug` gives no additional information

When I "fit" a version file (although no instructions for this are available), so using innate knowledge. The error goes away, and is replaced by a plain failure, with no reason given.

```
% echo 1 > /tmp/store/ria-layout-version
% git annex initremote ora1 encryption=none type=external externaltype=ora url=ria+ssh://localhost/tmp/store  
initremote ora1 
git-annex: INITREMOTE failed with no reason given
failed
initremote: 1 failed
```

Again `--debug` gives no additional information. The reason for the failure is simply that there is no location for the dataset in the store yet.

I was worried for a moment that because I have not provided a dataset ID, it would not look at the right location, because the help says

```
archive-id
        Dataset ID. Should be set automatically by datalad
```

and I am not using datalad directly -- but this is apparently not the case. Another dead end.

The reason why there is no explanation whatsoever on the failure is the use of the following pattern for exceptions

```
raise RIARemoteError(str(<exception>))
```

where `<exception>` is a plain `FileNotFoundError`, `NoLayoutVersion` etc in key parts of the verification code. These all render to an empty string, hence users are completely clueless.

After manually creating the dataset location, I can progress to the next problem:

```
% git annex initremote ora1 encryption=none type=external externaltype=ora url=ria+ssh://localhost/tmp/store
initremote ora1 

  Remote doesn't report any object tree version. Consider upgrading datalad or add a fitting 'ria-layout-version' file at the remote dataset root. Treating remote as read-only in order toprevent damage by putting things into an unknown version of the target layout. You can overrule this by setting 'annex.ora-remote.<name>.force-write=true'.
ok
(recording state in git...)
```

which is not longer an error with respect to special remote configuration.

mih:
- Prevention of multi-line messages breaking the special remote protocol
- Generalize multi-line message protection for `RemoteError` from just ORA to all special remotes
- Report on the cause for download error in the `datalad` special remote; fixes #5773 
- Introduce new helper to generator compact messages from an exception with underlying causes
- Alter ORA exception raising to report underlying causes; fixes #6231

Together this change improves from `maint` behavior:

`RuntimeError(Failed to download from any of 1 locations)`

from `master`

```
  external special remote protocol error, unexpectedly received ")" (unable to parse command)
  unable to use special remote due to protocol error
```

to 

`Failed to download from any of 1 locations [S3ResponseError: 403 Forbidden\n]`

### DIFF
--- a/datalad/customremotes/datalad.py
+++ b/datalad/customremotes/datalad.py
@@ -13,15 +13,14 @@ __docformat__ = 'restructuredtext'
 import logging
 from urllib.parse import urlparse
 
-from annexremote import RemoteError
-
 from datalad.downloaders.providers import Providers
 from datalad.support.exceptions import (
     CapturedException,
     TargetFileAbsent,
 )
 
-from .base import AnnexCustomRemote
+from datalad.customremotes import RemoteError
+from datalad.customremotes.base import AnnexCustomRemote
 
 lgr = logging.getLogger('datalad.customremotes.datalad')
 
--- a/datalad/customremotes/datalad.py
+++ b/datalad/customremotes/datalad.py
@@ -18,6 +18,7 @@ from datalad.support.exceptions import (
     CapturedException,
     TargetFileAbsent,
 )
+from datalad.utils import unique
 
 from datalad.customremotes import RemoteError
 from datalad.customremotes.base import AnnexCustomRemote
@@ -38,6 +39,7 @@ class DataladAnnexCustomRemote(AnnexCustomRemote):
 
     def transfer_retrieve(self, key, file):
         urls = []
+        error_causes = []
         # TODO: priorities etc depending on previous experience or settings
         for url in self.gen_URLS(key):
             urls.append(url)
@@ -49,10 +51,17 @@ class DataladAnnexCustomRemote(AnnexCustomRemote):
                 return
             except Exception as exc:
                 ce = CapturedException(exc)
-                self.message("Failed to download url %s for key %s: %s"
-                             % (url, key, ce))
-        raise RemoteError(
-            f"Failed to download from any of {len(urls)} locations")
+                cause = getattr(exc, '__cause__', None)
+                debug_msg = f"Failed to download {url} for key {key}: {ce}"
+                if cause:
+                    debug_msg += f' [{cause}]'
+                self.message(debug_msg)
+                error_causes.append(cause)
+
+        error_msg = f"Failed to download from any of {len(urls)} locations"
+        if error_causes:
+            error_msg += f' {unique(error_causes)}'
+        raise RemoteError(error_msg)
 
     def checkurl(self, url):
         try:
--- a/datalad/customremotes/datalad.py
+++ b/datalad/customremotes/datalad.py
@@ -50,8 +50,8 @@ class DataladAnnexCustomRemote(AnnexCustomRemote):
                 return
             except Exception as exc:
                 ce = CapturedException(exc)
-                self.annex.debug("Failed to download url %s for key %s: %s"
-                                 % (url, key, ce))
+                self.message("Failed to download url %s for key %s: %s"
+                             % (url, key, ce))
         raise RemoteError(
             f"Failed to download from any of {len(urls)} locations")
 
@@ -64,7 +64,7 @@ class DataladAnnexCustomRemote(AnnexCustomRemote):
             return [props]
         except Exception as exc:
             ce = CapturedException(exc)
-            self.annex.debug("Failed to check url %s: %s" % (url, ce))
+            self.message("Failed to check url %s: %s" % (url, ce))
             return False
 
     def checkpresent(self, key):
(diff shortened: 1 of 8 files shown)


datalad/distributed/ora_remote.py
```diff
@@ -1097,27 +1098,21 @@ def _local_io(self):
         #       + just isinstance(LocalIO)?
         return not self.storage_host
 
+    # XXX this is pretty much obsolete, we should use self.message()
+    # which does debug messaging by default. however, I don't know if
+    # these prefixes are used for anything
```
bpoldrack, line 1103: Re question in comment: No, prefix is not used anywhere, only exists to make it recognizable in a heavy debug output (datalad + annex +ORA).

bpoldrack (approved):

datalad/distributed/ora_remote.py
mih: I removed it now. Thx!

mih:
@bpoldrack thanks for the quick review. I took this a little further. Maybe you could have another look.

datalad/support/exceptions.py
```diff
@@ -185,6 +185,26 @@ def format_oneline_tb(exc, tb=None, limit=None, include_str=True):
...
+def format_exception_with_cause(e):
+    """Helper to recursively format an exception with all underlying causes
+
+    For each exception in the chain either the str() of it is taken, or the
+    class name of the exception, with the aim to generate a simple and
+    comprehensible description that can be used in user-facing messages.
+    It is explicitly not aiming to provide a detailed/comprehensive source
+    of information for in-depth debugging.
+
+    '-caused by-' is used a separator between exceptions to be human-readable
```
bpoldrack, line 197: Note, that there is `__context__` as well. `__cause__` is set when `raise ... from ...` is used, `__context__`, when an exception happens during the handling of another one.

bpoldrack (commented):
GTM. However, note, that there is `__context__` as well. `__cause__` is set when `raise ... from ...` is used, `__context__`, when an exception happens during the handling of another one. Since you are at it - might be worth adding as well.

datalad/customremotes/datalad.py
```diff
@@ -50,10 +51,17 @@ def transfer_retrieve(self, key, file):
                 return
             except Exception as exc:
                 ce = CapturedException(exc)
-                self.annex.debug("Failed to download url %s for key %s: %s"
-                                 % (url, key, ce))
-        raise RemoteError(
-            f"Failed to download from any of {len(urls)} locations")
+                cause = getattr(exc, '__cause__', None)
```
bpoldrack, line 54: Another suggestion (no objection to PR really):

This string formatting could actually go into `CapturedException` as an alternative (there's `format_standard` and `format_short` already. Could be `format_long`.

mih: True, thx! I will keep in mind for now, though. The location here is minor in the context of a special remote. In the big picture, we might want to investigate the reporting of exceptions in general (same viewpoint, but the implications for an implementation might differ). The details are not clear to me yet.

mih:
__context__ doesnt seem as relevant for user reporting. At least, I cannot see a scenario where this would not be about a software defect.

bpoldrack:
@mih 

> __context__ doesnt seem as relevant for user reporting. At least, I cannot see a scenario where this would not be about a software defect.

Agree. Would be relevant for debugging.

PR #6245, opened 2021-11-25
TST: Enable filter.annex.process on windows to speed things up

mih:
See https://git-annex.branchable.com/bugs/Windows__58___substantial_per-file_cost_for___96__add__96__/ and https://github.com/datalad/datalad/issues/5994 for the saga.

Let's see what effect it would have.

Pre:

image

Post:

image

Pretty interesting. A substantial improvement (16% overall), but not evenly across jobs (between 5% and 32%)... No idea what is happening here.

The job that saw the massive speed-up is running:

```
local/
test_add_archive_content.py
test_add_readme.py
test_addurls.py
test_check_dates.py
test_clean.py
test_copy_file.py
test_download_url.py
test_export_archive.py
test_no_annex.py
test_remove.py
test_rerun_merges.py
test_rerun.py
test_run_procedure.py
test_subdataset.py
test_unlock.py
test_wtf.py

support/
test_annexrepo.py
test_ansi_colors.py
test_cache.py
test_captured_exception.py
test_cookies.py
test_digests.py
test_due_utils.py
test_external_versions.py
test_fileinfo.py
test_gitrepo.py
test_globbedpaths.py
test_json_py.py
test_locking.py
test_network.py
test_parallel.py
test_path.py
test_repodates.py
test_repo_save.py
test_sshconnector.py
test_sshrun.py
test_stats.py
test_status.py
test_vcr_.py

ui/
test_base.py
test_dialog.py
```

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -220,6 +220,11 @@ init:
   # Identity setup
   - git config --global user.email "test@appveyor.land"
   - git config --global user.name "Appveyor Almighty"
+  # globally setting filter.annex.process (needs git-annex 8.20211117) to reduce git-add runtime
+  # https://git-annex.branchable.com/bugs/Windows__58___substantial_per-file_cost_for___96__add__96__/
+  - cmd: git config --system filter.annex.process "git-annex filter-process"
+  # might also bring a slight speed-up on Mac, but Ubuntu git-annex is currently too old
+  #- sh: sudo git config --system filter.annex.process "git-annex filter-process"
   # Scratch space
   - cmd: md C:\DLTMP
   # we place the "unix" one into the user's HOME to avoid git-annex issues on MacOSX


mih:
Hmm, no effect. It seems running `git config --global` in the appveyor setup has no effect at all. I guess, this is because we internally replace the config effective for the tests.

I will replace with `--system` for now, but if that works, it would make sense to customize the actual test config in `setup_package()`

mih:
I am taking this out of draft-mode. It is clearly beneficial. Also it is unclear how exactly it is manifesting itself.

mih:
Runtimes replicate

image

mih:
Ubuntu's git-annex is too old for this (8.20210903), but the Mac tests remain more or less, where they were (~5%). So setting this variable is not doing global magic to our tests, but specifically boosts the windows/adjusted mode performance. I will now revert the change back to appveyor-win only.

image

adswa:
The huge speed up in WinP39a2 is impressive, and just for the record, I can replicate it. On my Windows computer I see substantial speed ups with the config compared to without the config in ``support`` and ``local`` (don't mind the overall long run times, they are faster on appveyor - the relative difference is important)

| Command | With config  | Without config| Speed-up by a factor of|
| ------------- | ------------- | ------------- |------------- |
| support    | 829  / 834     sec    | 2500 / 1890 sec |   2.2-3           |
| local         | 931 / 1115/ 804 sec  | 1999 / 1740 sec|  1.7-2.2 |
| local         | 2.6 / 2.6 sec | 2.6 / 2.5 sec | - |

mih:
That is impressive. Also interesting is the range from 75min down to 30 min. Much more impact than on appveyor, it seems the per-process cost is much more expensive on that machine than on appveyor, but the overall performance is higher!

adswa:
Another anecdote: I ran the complete test suite under windows in 106 minutes on this machine just now. It would not be an exaggeration to say that this easily took a full night on the previous system and without the new configuration

mih:
I am in favor of datalad setting this automatically on windows, unless it is explicitely set otherwise.... But not yet. First we need to get some more experience with it.

adswa:
I was thinking to add it as an optional configuration into the handbooks installation instructions for now.

mih:
It would be hard to justify a default slowdown of 3x without any downsides.

adswa:
> It would be hard to justify a default slowdown of 3x without any downsides.

Is this a "yes" to include it in the handbook?

mih:
Sorry, yes, inclusion in the handbook is definitely useful. But on top of that we should aim for an optimized default. It is planned for git-annex anyways (v9 or late v8). We might want to jump earlier.

adswa:
> Sorry, yes, inclusion in the handbook is definitely useful. 

https://github.com/datalad-handbook/book/pull/791 done

> But on top of that we should aim for an optimized default. It is planned for git-annex anyways (v9 or late v8). We might want to jump earlier.

I agree

da5nsy:
@mih 
> It is planned for git-annex anyways (v9 or late v8)

Did this happen? (I don't know the system well enough to know where to look)

PR #6253, opened 2021-11-26
BF: Keyring.delete called uninitialized attribute

bpoldrack:
Likely a typo, delete() called an uninitialized private attribute instead
of the lazy loading, protected one.

Note, that this particular bug was only recently introduced to master, so "patch" but against master.

Introduced by: 3cddd7c4ac832e89241d5b396abf47fea994fc71


mih:
If you know what introduced it, please record it here.

bpoldrack:
> If you know what introduced it, please record it here.

Updated description.

mih (approved):
Thx!

PR #6261, opened 2021-11-30
DOC: positional vs kwargs design doc

yarikoptic:
This is a satellite to #6176 to first (before resolving conflicts and updating that PR) establish agreed upon design specification.


mih (approved):
Thanks!

bpoldrack (approved):

yarikoptic:
Unprecedented agreement without any comments is alarming ;-)

bpoldrack:
@yarikoptic 

Let's take advantage of it as long as it stands. :-D

PR #6262, opened 2021-11-30
ENH: enable recursive glob (**) for `run`

### LINKED ISSUE #6241: Support recursive wildcards for `run` output specifications
An issue brought up by @AKSoo in the DataLad channel. Basically this means supporting `**` for glob, with `recursive=True` -- available since PY3.5

AKSoo:
Closes #6241

### DIFF
--- a/datalad/core/local/run.py
+++ b/datalad/core/local/run.py
@@ -174,7 +174,17 @@ class Run(Interface):
              code_cmd="""\
              datalad run -m 'run my script' -i 'data/*' \\
              -i 'datafile.txt' -o 'output_dir/*' -o \\
-             'outfile.txt' 'code/script.sh'""")
+             'outfile.txt' 'code/script.sh'"""),
+        dict(test="Use ** to match any file at any directory depth recursively. "
+                  "Single * does not check files within matched directories.",
+             code_py="""\
+             run(cmd='code/script.sh',
+                 message='run my script',
+                 inputs=['data/**/*.dat'],
+                 outputs=['output_dir/**'])""",
+             code_cmd="""\
+             datalad run -m 'run my script' -i 'data/**/*.dat' \\
+             -o 'output_dir/**' 'code/script.sh'""")
     ]
 
     result_renderer = "tailored"
--- a/datalad/core/local/run.py
+++ b/datalad/core/local/run.py
@@ -175,7 +175,7 @@ class Run(Interface):
              datalad run -m 'run my script' -i 'data/*' \\
              -i 'datafile.txt' -o 'output_dir/*' -o \\
              'outfile.txt' 'code/script.sh'"""),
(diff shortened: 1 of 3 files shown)


datalad/core/local/run.py
```diff
@@ -174,7 +174,17 @@ class Run(Interface):
              code_cmd="""\
              datalad run -m 'run my script' -i 'data/*' \\
              -i 'datafile.txt' -o 'output_dir/*' -o \\
-             'outfile.txt' 'code/script.sh'""")
+             'outfile.txt' 'code/script.sh'"""),
+        dict(test="Use ** to match any file at any directory depth recursively. "
```
adswa, line 178:
```suggestion
        dict(text="Use ** to match any file at any directory depth recursively. "
```

I believe this is the cause for the current failures in the CI

adswa (commented):
Awesome! Thanks a lot! I just quickly glanced at the test failures for now, I believe they stem from a misspelled key in the examples - left a quick suggestion for that.

adswa:
added ``semver-patch`` for the semver test, please change or say something if someone disagrees :)

datalad/core/local/run.py
AKSoo: Oops! Thanks for spotting this.

adswa (approved):
This works nicely for me! Thanks a lot! I will wait a bit to give others a chance to spot things I may have missed, though :)

mih (approved):
Great, thx!

yarikoptic:
Looks great!  My only "fear"s are
- that it might change reproducibility of prior runs. E.g. if someone used `**` just by doubletyping `*`, which resulted in non-recursive match, now behavior would change, so `rerun` might find more
- in presence of subdatasets, something like `sub/**/*.nii.gz` could find even git annex keys themselves within sub/.git/annex/objects . But it seems to be not the case, in that `**` I guess does not include dot dirs:

```shell
$> python3 -c "from glob import glob; print([x for x in glob('ds*/.git/**/*.nii.gz', recursive=True) if 'annex' in x])"
['ds003027/.git/annex/objects/ZJ/k5/MD5E-s89045275--1625ff12d9c2d4e35f50df32601a7bb2.nii.gz', 'ds003027/.git/annex/objects/ZJ/k5/MD5E-s89045275--1625ff12d9c2d4e35f50df32601a7bb2.nii.gz/MD5E-s89045275--1625ff12d9c2d4e35f50df32601a7bb2.nii.gz']

$> python3 -c "from glob import glob; print([x for x in glob('ds*/**/*.nii.gz', recursive=True) if 'annex' in x])"     
[]
```
so I guess we should be safe on that end.

mih:
OK, let's proceed! A changelog will come.

adswa:
Its mentioned in the handbook now as well :+1:

mih:
Great, thx!

PR #6263, opened 2021-11-30
MNT: no tabs for vim/ex

AKSoo:
The modelines do not match the code style (space indentation).


adswa:
My vim foo isn't strong enough to review this, I'll leave it to @mih. But I'll cancel the remaining CIs because this change shouldn't affect the tests and there's no need to spin those machines up. I'll also add a ``semver-internal`` label.

mih:
Cool! Thanks!

PR #6266, opened 2021-11-30
ENH(TST): add some "pragma: no cover" wherever test code is not expected to run

yarikoptic:
Just spotted while staring in a stale browser tab of
https://github.com/datalad/datalad/pull/6190/files
which is now merged


yarikoptic:
all green, no need to pester precious developers with the review. Let's proceed

PR #6272, opened 2021-12-01
Do not run tests where test setup isn't safe

bpoldrack:
As lined out in PR #6260, I think we should not run the tests on OSX and Windows with too old a git for the changed approach in that PR to be applied.

Not only may tests fail or hang, they can also have unintended side-effects on user's files.

ATM, I'm thinking just raising from within `setup_package` in this case may do. WDYT, @datalad/developers ?

### DIFF
--- a/datalad/__init__.py
+++ b/datalad/__init__.py
@@ -172,6 +172,12 @@ def setup_package():
         return new_home, cfg_file
 
     if external_versions['cmd:git'] < "2.32":
+        if (on_osx or on_windows) and \
+                not cfg.obtain("datalad.tests.force-fake-home", default=False):
+            raise RuntimeError("Refuse to execute datalad tests with git<2.32, "
+                               "see config datalad.tests.force-fake-home for "
+                               "details.")
+
         # To overcome pybuild overriding HOME but us possibly wanting our
         # own HOME where we pre-setup git for testing (name, email)
         if 'GIT_HOME' in os.environ:
--- a/datalad/__init__.py
+++ b/datalad/__init__.py
@@ -129,6 +129,7 @@ def setup_package():
     from datalad.utils import (
         make_tempfile,
         on_osx,
+        on_windows
     )
 
     if on_osx:
--- a/datalad/interface/common_cfg.py
+++ b/datalad/interface/common_cfg.py
@@ -202,7 +202,7 @@ definitions = {
     },
     'datalad.tests.force-fake-home': {
         'ui': ('yesno', {
-               'title': 'Force running tests on OSX and window with manipulated HOME or USERPROFILE respectively.',
+               'title': 'Force running tests on OSX and window with manipulated HOME or USERPROFILE respectively',
                'text': "This is potentially dangerous and only applies on OSX and Windows if the git version used by datalad is older than 2.32."
                        "By default, tests are not executed under these circumstances, since the legacy setup for the test environment can have unintended side-effects:{nl}"
                        "1. On Windows, sub-processes launched by the tests fail to use the desired, isolated locations for configs, caches, etc."
--- a/datalad/interface/common_cfg.py
+++ b/datalad/interface/common_cfg.py
@@ -12,7 +12,10 @@
 
 __docformat__ = 'restructuredtext'
 
-from os import environ
+from os import (
+    environ,
+    linesep
+)
 from os.path import expanduser
 from os.path import join as opj
 
@@ -196,6 +199,18 @@ definitions = {
         'ui': ('yesno', {
                'title': 'Does not execute teardown_package which cleans up temp files and directories created by tests if this flag is set'}),
         'type': EnsureBool(),
+    },
+    'datalad.tests.force-fake-home': {
+        'ui': ('yesno', {
+               'title': 'Force running tests on OSX and window with manipulated HOME or USERPROFILE respectively.',
+               'text': "This is potentially dangerous and only applies on OSX and Windows if the git version used by datalad is older than 2.32."
+                       "By default, tests are not executed under these circumstances, since the legacy setup for the test environment can have unintended side-effects:{nl}"
+                       "1. On Windows, subprocesses launched by the tests (incl. indirectly started ones), may end up reading from and writing to wrong locations, as any subprocess will fail to retrieve correct standard locations for configs, caches, etc. This issue is discussed in gh-6160 and gh-6260."
+                       "See also: https://github.com/datalad/datalad/issues/6160 and https://github.com/datalad/datalad/pull/6260 for more details.{nl}"
+                       "2. OSX, most notably the system-level osxkeychain git-credential helper, which will lead to hanging git commands if it fails to find the keychain in its expected location under $HOME/Library/Keychain."
+                       "Do you want to run tests nevertheless?".format(nl=linesep)}),
+        'type': EnsureBool(),
+        'default': False
     },
     'datalad.tests.dataladremote': {
         'ui': ('yesno', {
--- a/datalad/interface/common_cfg.py
+++ b/datalad/interface/common_cfg.py
@@ -205,8 +205,8 @@ definitions = {
                'title': 'Force running tests on OSX and window with manipulated HOME or USERPROFILE respectively.',
                'text': "This is potentially dangerous and only applies on OSX and Windows if the git version used by datalad is older than 2.32."
                        "By default, tests are not executed under these circumstances, since the legacy setup for the test environment can have unintended side-effects:{nl}"
-                       "1. On Windows, subprocesses launched by the tests (incl. indirectly started ones), may end up reading from and writing to wrong locations, as any subprocess will fail to retrieve correct standard locations for configs, caches, etc. This issue is discussed in gh-6160 and gh-6260."
-                       "See also: https://github.com/datalad/datalad/issues/6160 and https://github.com/datalad/datalad/pull/6260 for more details.{nl}"
+                       "1. On Windows, sub-processes launched by the tests fail to use the desired, isolated locations for configs, caches, etc."
+                       "See https://github.com/datalad/datalad/issues/6160 and https://github.com/datalad/datalad/pull/6260 for more details.{nl}"
                        "2. OSX, most notably the system-level osxkeychain git-credential helper, which will lead to hanging git commands if it fails to find the keychain in its expected location under $HOME/Library/Keychain."
                        "Do you want to run tests nevertheless?".format(nl=linesep)}),
         'type': EnsureBool(),


datalad/__init__.py
```diff
@@ -142,6 +142,19 @@ def set_envvar(v, val):
...
+            # This approach of changing HOME (or USERPROFILE respectively)
+            # has unintended side-effects on:
+            #   1. Windows, generally, as any subprocess will fail to retrieve
+            #      correct standard locations for configs, caches, etc.
+            #   2. OSX, most noteably the system-level osxkeychain
+            #      git-credential helper, which will lead to hanging git
+            #      commands if it fails to find the keychain in its expected
+            #      location under $HOME/Library/Keychain.
+            raise RuntimeError("Refuse to execute datalad tests with git < 2.32"
+                               " on this system, as this may have unintended "
```
yarikoptic, line 155: I like uniformity
```suggestion
            raise RuntimeError("Refuse to execute datalad tests with git < 2.32 "
                               "on this system, as this may have unintended "
```

yarikoptic:
I don't mind, hopefully we do not run into such systems (with elderly git) where we build/test datalad, since it would add some pains.

bpoldrack:
> I don't mind, hopefully we do not run into such systems (with elderly git) where we build/test datalad, since it would add some pains.

Before 2.32 actually isn't that old. However, this is only about win+osx, so I think as "consumer" OSes, only running tests on recent-ish git should be fine. We would still want to be able to run those tests on linux server with older git, I guess.

datalad/__init__.py
bpoldrack: Thou shall have uniformity

datalad/__init__.py
```diff
@@ -142,6 +142,19 @@ def set_envvar(v, val):
...
+            #   1. Windows, generally, as any subprocess will fail to retrieve
+            #      correct standard locations for configs, caches, etc.
+            #   2. OSX, most noteably the system-level osxkeychain
+            #      git-credential helper, which will lead to hanging git
+            #      commands if it fails to find the keychain in its expected
+            #      location under $HOME/Library/Keychain.
+            raise RuntimeError("Refuse to execute datalad tests with git < 2.32 "
+                               "on this system, as this may have unintended "
+                               "side-effects.")
+
```
adswa, line 157: Without the knowledge acquired over the course of #6160 and #6260, this error message is pretty confusing. It would be fair to add a link or reference. Same goes for the comment above.

adswa (commented):
What about a config switch to be able to run the tests nevertheless? A system with an old version of Git wouldn't be able to run any tests, even if the tests that are run wouldn't not suffer from these side effects.

datalad/__init__.py
bpoldrack: Agreed

bpoldrack:
> What about a config switch to be able to run the tests nevertheless?

Yes, will do.

bpoldrack:
Updated accordingly, @adswa

mih:
I think this change has a rather suboptimal UX. People will not be aware of this PR (#6272). The info given in the message, sends them to #6160 and #6260 to start a journey to hopefully discover this PR, to be able to inspect the diff, to conclude that the added config switch can help them out.

Why not say all of this directly in the message instead of riddles, or use `obtain()` to get the informative text for the switch?

bpoldrack:
@mih 
Done

mih:
Ok, so with this change, the behavior is:

```
RuntimeError: Refuse to execute datalad tests with git<2.32, see config datalad.tests.force-fake-home for details.
```

With the following patch

```diff
diff --git a/datalad/__init__.py b/datalad/__init__.py
index ad973a878..fe940cf62 100644
--- a/datalad/__init__.py
+++ b/datalad/__init__.py
@@ -174,7 +174,7 @@ def setup_package():
 
     if external_versions['cmd:git'] < "2.32":
         if (on_osx or on_windows) and \
-                not cfg.obtain("datalad.tests.force-fake-home", default=False):
+                not cfg.obtain("datalad.tests.force-fake-home"):
             raise RuntimeError("Refuse to execute datalad tests with git<2.32, "
                                "see config datalad.tests.force-fake-home for "
                                "details.")
diff --git a/datalad/interface/common_cfg.py b/datalad/interface/common_cfg.py
index ff8d155a0..0459ee74e 100644
--- a/datalad/interface/common_cfg.py
+++ b/datalad/interface/common_cfg.py
@@ -210,7 +210,6 @@ definitions = {
                        "2. OSX, most notably the system-level osxkeychain git-credential helper, which will lead to hanging git commands if it fails to find the keychain in its expected location under $HOME/Library/Keychain."
                        "Do you want to run tests nevertheless?".format(nl=linesep)}),
         'type': EnsureBool(),
-        'default': False
     },
     'datalad.tests.dataladremote': {
         'ui': ('yesno', {
```

it will be:

```
Force running tests on OSX and window with manipulated HOME or USERPROFILE respectively.
This is potentially dangerous and only applies on OSX and Windows if the git version used by datalad is older than 2.32.By default, tests are not executed under these circumstances, since the legacy setup for the test environment can have unintended side-effects:
1. On Windows, subprocesses launched by the tests (incl. indirectly started ones), may end up reading from and writing to wrong locations, as any subprocess will fail to retrieve correct standard locations for configs, caches, etc. This issue is discussed in gh-6160 and gh-6260.See also: https://github.com/datalad/datalad/issues/6160 and https://github.com/datalad/datalad/pull/6260 for more details.
2. OSX, most notably the system-level osxkeychain git-credential helper, which will lead to hanging git commands if it fails to find the keychain in its expected location under $HOME/Library/Keychain.Do you want to run tests nevertheless? (choices: yes, no): 
```

It is worth discussing which behavior is more desirable.

I will comment on the particular message separately.

datalad/interface/common_cfg.py
```diff
@@ -197,6 +200,18 @@
                'title': 'Does not execute teardown_package which cleans up temp files and directories created by tests if this flag is set'}),
         'type': EnsureBool(),
     },
+    'datalad.tests.force-fake-home': {
```
mih, line 203: Somehow this name is suboptimal. The whole thing seems to be more about which approach to configuration/cache isolation we are using. The new behavior is to no longer isolate the user HOME from the tests. It only "protects" the git config. User cache/config locations are in the normal HOME and leave their garbage in it.

`force-fake-home` feels like something I would want to set in order to avoid all this. But it is actually considered a fallback for outdated systems.

datalad/interface/common_cfg.py
```diff
@@ -197,6 +200,18 @@
...
+        'ui': ('yesno', {
+               'title': 'Force running tests on OSX and window with manipulated HOME or USERPROFILE respectively.',
```
mih, line 205: See comment above. This somewhat conflates what the setting is about, and how it is put to action.

```suggestion
               'title': 'Force running tests on OSX and window with manipulated HOME or USERPROFILE respectively',
```

datalad/interface/common_cfg.py
```diff
@@ -197,6 +200,18 @@
...
+               'text': "This is potentially dangerous and only applies on OSX and Windows if the git version used by datalad is older than 2.32."
+                       "By default, tests are not executed under these circumstances, since the legacy setup for the test environment can have unintended side-effects:{nl}"
+                       "1. On Windows, subprocesses launched by the tests (incl. indirectly started ones), may end up reading from and writing to wrong locations, as any subprocess will fail to retrieve correct standard locations for configs, caches, etc. This issue is discussed in gh-6160 and gh-6260."
+                       "See also: https://github.com/datalad/datalad/issues/6160 and https://github.com/datalad/datalad/pull/6260 for more details.{nl}"
```
mih, line 209:
```suggestion
                       "1. On Windows, sub-processes launched by the tests fail to use the desired, isolated locations for configs, caches, etc."
                       "See https://github.com/datalad/datalad/issues/6160 and https://github.com/datalad/datalad/pull/6260 for more details.{nl}"
```

datalad/interface/common_cfg.py
```diff
@@ -197,6 +200,18 @@
...
+                       "2. OSX, most notably the system-level osxkeychain git-credential helper, which will lead to hanging git commands if it fails to find the keychain in its expected location under $HOME/Library/Keychain."
```
mih, line 210:
```suggestion
                       "2. On OSX, the system-level `osxkeychain` git-credential helper will cause hanging Git commands, because it fails to find the keychain under $HOME/Library/Keychain."
```

I never asked, but I assume it is not an option to just copy that keychain?

datalad/interface/common_cfg.py
bpoldrack: Agree. I am not liking the name either

datalad/interface/common_cfg.py
bpoldrack: It is. But: I have no idea whether that path is fix or something that we'd need to dynamically find out, possibly opening just another rabbit hole. Plus, that approach would need to be repeated for every such third-party software that we discover over time to have a similar issue. Therefore, that approach felt to me like starting a never-ending story.

bpoldrack: But on a sample size of 1, testing with a plain copy worked as well.

bpoldrack:
@mih 
Re not giving a default: I simply didn't fully realize the effect. So, the behavior w/o seems better to me. Would raise the question to what extend we are concerned with non-interactive behavior, though.

datalad/interface/common_cfg.py
bpoldrack: To add one more aspect to this: I don't know whether the copied keychain would be fully functioning. I only tested that such a copy would not lead to a hanging credential helper. If it contains credentials to clone from, say, github, and would actually need to be used during other tests obtaining test repositories or something - I don't know whether there would need to be an additional step of registering/verifying the new location.

datalad/interface/common_cfg.py
bpoldrack: I am failing to come up with something better. `force-outdated-isolation`? I don't know.

datalad/interface/common_cfg.py
mih: Before discussing these details, I think it is worth discussing the general approach of removing any isolations. They were introduced to prevent pollution. And now we are saying

- we must not pollute the git config
- we will pollute everything else

it feels a bit schizophrenic to me.

datalad/interface/common_cfg.py
mih: re "outdated" see also https://github.com/datalad/datalad/pull/6272#discussion_r761105214

datalad/interface/common_cfg.py
bpoldrack: Yes, from my point of view it's just about removing the wrong approach of isolation (since it actually breaks stuff + is partially dysfunctional).

I thought about patching the `AppDirs` object from within `setup_package` to make it refer to an application "datalad-test", rather than "datalad" and register the resulting dirs with `_TEMP_PATHS_GENERATED` for clean up.
But that leaves the issue of subprocesses, that are datalad processes, too, and wouldn't have access to the patched one. Like special remotes and credential helper.

mih:
This PR has stalled. I am planning to close it before the holidays, unless activity resumes.

bpoldrack:
@mih 
It's stalled, b/c you wanted to discuss the underlying approach first according to https://github.com/datalad/datalad/pull/6272#discussion_r761105214. Where should that happen?

mih:
I am closing this PR now. There are open questions, and no drive towards a resolution.

Issue #6275, opened 2021-12-02
Unnecessarily restrictive `eval_results` decorator limits `-d`/`--dataset` parameter to file-URLs


christian-monch:
### What is the problem?

The eval_results-decorator (in `datalad/interface/utils.py`) attempts to instantiate a `Dataset`-object with the dataset argument, which is provided via `-d`/`--dataset`. This leads to an error, if anything else then a locally resolvable file-URL is provided as dataset argument. The `Dataset`-instance is only created to query the dataset config manager for JSON-hooks, which are, should they have been defined, applied to the results.

The current restrictions prevent the use of the dataset parameter to specify a remote repository for `datalad meta-dump`. For example, the following command  (with the installed master-version of `datalad/datalad-metalad`) should dump metadata stored in the github repository `https://github.com/datalad/test_metadata.git`:

```shell
> datalad meta-dump -d https://github.com/datalad/test_metadata.git -r
```

But instead, it yields an error:

```
[ERROR  ] ValueError(Non 'file://' URL cannot be resolved to a local path) (ValueError)
```

This behavior should be changed in order to allow "arbitrary" dataset source definitions. One possible modification would look for hooks in the dataset configuration, if the dataset is locally available, and query the default config manager, i.e. `datalad.cfg` otherwise.

### What steps will reproduce the problem?

Any non-locally resolvable file argument to `-d`/`--datasets` For example:

```shell
> datalad status -d https://github.com/psychoinformatics-de/studyforrest-data.git
```
leads to:
```
[ERROR  ] ValueError(Non 'file://' URL cannot be resolved to a local path) (ValueError)
```

### Datalad information

`datalad 0.15.3`

PR #6276, opened 2021-12-02
Allow other URLs than local file as dataset parameter

christian-monch:
This PR fixes issue #6275.

The PR removes the restriction of dataset parameters to locally resolvable file-URLs. This is used, for example, in the new metalad version where you can dump metadata from remote repositories.

The main difference introduced is that the decorator `eval_results` only looks for hooks in the default configuration, if the dataset parameter does not specify a locally resolvable file-URL.

### DIFF
--- a/datalad/interface/utils.py
+++ b/datalad/interface/utils.py
@@ -48,7 +48,7 @@ from datalad.support.exceptions import (
 )
 from datalad import cfg as dlcfg
 from datalad.dochelpers import single_or_plural
-from datalad.support.network import URL
+
 from datalad.ui import ui
 import datalad.support.ansi_colors as ac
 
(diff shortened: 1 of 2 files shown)


bpoldrack (approved):

mih (changes requested):
This doesn't work as-is. A plain path is also enough to get a dataset instance, but

```py
In [1]: from datalad.support.network import URL

In [2]: URL('/mike').is_local()
Out[2]: False
```

christian-monch:
> This doesn't work as-is. A plain path is also enough to get a dataset instance, but
> 
> ```python
> In [1]: from datalad.support.network import URL
> 
> In [2]: URL('/mike').is_local()
> Out[2]: False
> ```

True (I misremembered `urlparse`). The validity determination is now delegated to the Dataset-constructor.

mih (approved):
This should do it now. Thx!

mih:
Appveyor failure is unrelated

PR #6244, opened 2021-11-24
Generator-style runner with application in BatchedCommand and AnnexRepo

christian-monch:
This PR provides a new threaded runner implementation that supports timeouts and generator-based subprocess-communication, Which means stdout and stderr data are available from a generator. Our standard protocol-approach is also supported with generators, i.e. every data packet from stdout and stderr is passed through an instance of `WitlessProtocol` (or a subclass), by calling `WitlessProtocol.pipe_data_received()`.

This PR:

- [x] Allows using a queue.Queue() as stdin (in addition to previously supported types), for convenient stdin writing.
- [x] Implements generator-style run() semantics if the provided protocol is derived from `GeneratorMixIn`-class.
- [x] Adds timeout capabilities for output file traffic and for process exit
- [x] Adds basic protocol tests
- [x] Adds tests for runner exceptions
- [x] Adds documentation for the threaded runners
- [x] Adds tests for `BatchedCommand`

The PR uses and demonstrates the generator-capabilities in `AnnexRepo` and in `BatchedCommand`, it:

- [x] Adds a GeneratorAnnexJsonProtocol to demonstrate how to send data to the generator.
- [x] Supports a generator-style runner in `AnnexRepo._call_annex_items_()`. To support the new `_call_annex_items_()` the method `AnnexRepo._call_annex()` has been extended to support generator-style runner, i.e. it uses a generator-style `GitRunner.run_on_filelist_chunks_items_()`.
- [x] Implements `BatchedCommand` based on a generator-style runner, which consolidates our codebase and demonstrates the application of the generator-style runner interaction
- [x] Adds documentation for `BatchedCommand`

Although there are quite a number of changes in this PR, the modifications are mainly encapsulated in the runner code and in `BatchedCommand`. Both have a slightly extended, backward compatible API. That means the new implementations are drop-in replacements.

The "application"-code that changed is mostly in four dozen lines in `datalad/support/annexrepo.py`. Due to the structure of the existing code, large parts of `AnnexRepo` are not generator-based, i.e. they expect and return lists or tuples. This required to "unwind" the generator-style results in the functions that return to legacy code. I think this PR lays the groundwork for us to convert our code iteratively to "all the way" generator-based code (where it is appropriate).

> This PR is the cleaned-up, improved, and optimized version of PR #6125 which became convoluted. All reviewer comments on PR #6125 have been considered and their resolution is part of this PR

PR #6278, opened 2021-12-03
Generator-based GitRepo.call_git_items_()

christian-monch:
This PR uses a generator-based runner to, i.e. a runner with a `GeneratorMixIn`-Protocol subclass, to implement `GitRepo.call_git_items_()`.

This is a follow up to PR #6244, especially the following reviewer comment: https://github.com/datalad/datalad/pull/6244#discussion_r760831590

The interface of  `GitRepo.call_git_items_()` does not change, but results will be yielded as soon as they are available.

### DIFF
--- a/datalad/dataset/gitrepo.py
+++ b/datalad/dataset/gitrepo.py
@@ -20,7 +20,6 @@ lifetime of a singleton.
 __all__ = ['GitRepo']
 
 import logging
-import os
 import re
 import threading
 import time
@@ -57,6 +56,7 @@ from datalad.support.exceptions import (
 )
 from datalad.utils import (
     ensure_list,
+    lock_if_required,
     Path,
 )
 
@@ -64,18 +64,6 @@ from datalad.utils import (
 lgr = logging.getLogger('datalad.dataset.gitrepo')
 
 preferred_encoding = getpreferredencoding(do_setlocale=False)
-
-
-@contextmanager
-def lock_if_required(lock_required, lock):
-    """Acquire/ and release the provided lock if indicated by a flag"""
-    if lock_required:
-        lock.acquire()
-    try:
-        yield lock
-    finally:
-        if lock_required:
-            lock.release()
 
 
 @contextmanager
@@ -377,7 +365,8 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
         runner = self._git_runner
         stderr_log_level = {True: 5, False: 11}[expect_stderr]
 
-        if not read_only and self._fake_dates_enabled:
+        read_write = not read_only
+        if read_write and self._fake_dates_enabled:
             env = self.add_fake_dates_to_env(env if env else runner.env)
 
         output = {
@@ -385,7 +374,7 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
             STDERR_FILENO: [],
         }
 
-        with lock_if_required(read_only, self._write_lock), \
+        with lock_if_required(read_write, self._write_lock), \
              git_ignore_check(expect_fail, output[STDOUT_FILENO], output[STDERR_FILENO]):
 
             for file_no, line in self._generator_call_git(args,
@@ -470,13 +459,14 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
         CommandError if the call exits with a non-zero status.
         """
 
-        if not read_only and self._fake_dates_enabled:
+        read_write = not read_only
+        if read_write and self._fake_dates_enabled:
             env = self.add_fake_dates_to_env(
                 env if env else self._git_runner.env)
 
         stderr_lines = []
 
-        with lock_if_required(read_only, self._write_lock), \
+        with lock_if_required(read_write, self._write_lock), \
              git_ignore_check(expect_fail, None, stderr_lines):
 
             for file_no, line in self._generator_call_git(
--- a/datalad/dataset/gitrepo.py
+++ b/datalad/dataset/gitrepo.py
@@ -20,6 +20,7 @@ lifetime of a singleton.
 __all__ = ['GitRepo']
 
 import logging
+import os
 from os import environ
 from os.path import lexists
 import re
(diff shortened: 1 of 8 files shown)


mih:
Are you planning to re-consolidate the additions to `call_git_items_()` that are modified from `_call_git()`?

christian-monch:
> Are you planning to re-consolidate the additions to `call_git_items_()` that are modified from `_call_git()`?

Yes, I will do that now.

christian-monch:
@mih: The latest commit consolidates `_call_git`, `call_git`, and `call_git_items` in `GitRepo`. All three methodes are now based on `_generator_call_git`.

The changes implement a drop-in replacement in order to keep the PR small, i.e. only very few changes outside of `GitRepo`. Having said that, I will create an issue to clean up the usage of `call_git` and `_call_git`. The latter returns a two-tuple, but only the first component is used if any.

mih:
I do not fully understand your last post. I can confirm that this code works, and is at least as fast as it was before. So all these signs point to a quick merge.

What I do not understand yet (and I assume this is what your last post was about), it why we have `_call_git()` and `call_git_items_()` being near-copies of each other.

This leads to possibly undesirable situations, like `call_git_success()` calling `_call_git()`, which now executes `_generator_call_git()`, but will only ever return a boolean. I assume this must be a more complex operation now than it was before.

FTR: unlike the other methods, `_call_git()` need not maintain its behavior, it is specifically not part of the public API -- and only used in 7 places throughout the code base.

datalad/support/gitrepo.py
```diff
@@ -419,13 +419,13 @@ def process_exited(self):
             )
         super().process_exited()
 
-    def pipe_data_received(self, fd, byts):
+    def pipe_data_received(self, fd, bytes):
```
mih, line 422: The reason for the previous name was to avoid shadowing the built-in `bytes`. I still consider it better style to not reuse built-in names, even though this particular could bit might be short enough to realize that this is not a type.

christian-monch: I agree

christian-monch:
> I do not fully understand your last post. I can confirm that this code works, and is at least as fast as it was before. So all these signs point to a quick merge.

I should have been clearer. Top priority in this PR was to keep it small and therefore keep internal and external APIs unmodified. 

> What I do not understand yet (and I assume this is what your last post was about), it why we have `_call_git()` and `call_git_items_()` being near-copies of each other.

With the aim of keeping `_call_git()` (due to the priority stated above), the similarities stem from the necessity to release the write_lock and from the `GitIgnoreError`-detection. The former is relatively easy to achieve with the try-finally-clause. Therefore it is present within `_call_git` and `call_git_items_()` (and could have been done simpler with context managers, as pushed in b6ff536a4c8f5704a84e09f4872a8a8ea5f7c2d5).

> This leads to possibly undesirable situations, like `call_git_success()` calling `_call_git()`, which now executes `_generator_call_git()`, but will only ever return a boolean. I assume this must be a more complex operation now than it was before.

Although the general complexity (O-Notation) remains the same for the same inputs, the additional method call might very well introduce some delay, i.e. increase a constant factor. That can be mitigated by cutting out `_call_git()` and using `_generator_call_git()`, which is probably what you had in mind (I did not do that in the first approach due to the small-diff-goal, but it is pushed now in ).


> FTR: unlike the other methods, `_call_git()` need not maintain its behavior, it is specifically not part of the public API -- and only used in 7 places throughout the code base.

Its return value is directly returned by `datalad.support.gitrepo.GitRepo.add_remote`, although the return values are not documented. If we can change the signature if `add_remote`, I will remove it.

datalad/dataset/gitrepo.py
```diff
@@ -53,6 +63,48 @@
...
+@contextmanager
+def lock_if_required(read_only,
+                     write_lock):
+    if not read_only:
+        write_lock.acquire()
+    try:
+        yield write_lock
+    finally:
+        if not read_only:
+            write_lock.release()
```
mih, line 78: What about the following change, and moving this into `datalad.utils`?
```suggestion
@contextmanager
def lock_if_required(lock_required, lock):
    """Acquire/ and release the provided lock if indicated by a flag"""
    if lock_required:
        lock.acquire()
    try:
        yield lock
    finally:
        if lock_required:
            lock.release()
```

mih (commented):
This is a pretty cool pattern to hide-away the try-except complexities! Awesome!

I left a suggestion to generalize the lock-helper a bit further to be able to take it out of the context of GitRepo. WDYT?

datalad/dataset/gitrepo.py
christian-monch: Very good

christian-monch: Will move it

mih (approved):
From my POV we can run with this! Thx!

mih:
The MacOS failure is more fundamental https://github.com/datalad/datalad/issues/6285

adswa:
Hey @christian-monch *waving for attention* :wave: 
The failures reported in datalad/datalad-ukbiobank#82 and the wrong fix and subsequent failure in PR datalad/datalad-ukbiobank#85 revealed that the output of any ``repo.call_git()`` call is stripped of the last line's trailing slash. This seems to have been introduced by you in this PR, and we debated in the chat that using ``call_git_()`` would fix this issue. But I'm not familiar enough with this code base to touch it, and as you changed it last and probably had good reasons to switch, could you take a look at how we can best fix this?
FTR, this patch (thx to @yarikoptic) would fix it for me, but unsure of side effects or better approaches:

```diff
diff --git a/datalad/dataset/gitrepo.py b/datalad/dataset/gitrepo.py
index a1e32b450..d7b831bca 100644
--- a/datalad/dataset/gitrepo.py
+++ b/datalad/dataset/gitrepo.py
@@ -429,12 +429,12 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
         ------
         CommandError if the call exits with a non-zero status.
         """
-        return "\n".join(
-            self.call_git_items_(args,
-                                 files,
-                                 expect_stderr=expect_stderr,
-                                 expect_fail=expect_fail,
-                                 read_only=read_only))
+
+        return self._call_git(args,
+                 files,
+                 expect_stderr=expect_stderr,
+                 expect_fail=expect_fail,
+                 read_only=read_only)[0]
 
     def call_git_items_(self,
                         args,
```

There were also a few comments by @yarikoptic in the chat:

> BTW _call_git docstring needs fixing:

```
        The parameters, return value, and raised exceptions match those
        documented for `call_git`.
```
and 
> that is odd -- worth checking if call_git atm just passes stderr to screen?

Issue #6118, opened 2021-10-27
[Suggestion] add --jobs/-J option to `datalad run`


AKSoo:
`run` involves `get` on the inputs. However, `run` cannot make use of parallel jobs for `get` because there is no option for it.

bpinsard:
I third that! ;)  And also for `datalad-containers-run`.
It would be critical when using datalad for parallel computing on HPC to maximize the usage of requested resources when cloning and getting input data and repushing results to a central repo (or ria-store).
I think for now the only way to achieve that is by setting the 2 following config values either globally or at a dataset level (which is not super convenient for nested ones), though I have not tested that yet. 

> datalad.runtime.max-annex-jobs
> Maximum number of git-annex jobs to request when “jobs” option set to “auto” (default): Set this value to enable parallel annex jobs that may speed up certain operations (e.g. get file content). The effective number of jobs will not exceed the number of available CPU cores (or 3 if there is less than 3 cores). Default: 1
> [value must be convertible to type ‘int’]

> datalad.runtime.max-jobs
> Maximum number of jobs DataLad can run in “parallel”: Set this value to enable parallel multi-threaded DataLad jobs that may speed up certain operations, in particular operation across multiple datasets (e.g., install multiple subdatasets, etc). Default: 1
> [value must be convertible to type ‘int’]

PR #6279, opened 2021-12-03
ENH: Add parallel jobs options for `run` and `rerun` (applies to get, save)

AKSoo:
Implements #6118 . I have 2 questions.

- [x] How do I edit the docstring for `jobs_opt`? I think we need to make it clear that the parallelization only applies to input `get` and output `save`, or some people might get the idea that datalad can parallelize scripts for them.
- [ ] How do we test this option? It seems like the `ProducerConsumer` class for general parallelization is tested, but not the specific parallel options for `get` and `save`.

### DIFF
--- a/datalad/core/local/run.py
+++ b/datalad/core/local/run.py
@@ -266,6 +266,10 @@ class Run(Interface):
             constraints=EnsureChoice(None, "basic", "command")),
         jobs=jobs_opt
     )
+    _params_['jobs']._doc += """\
+        NOTE: This option can only parallelize input retrieval (get) and output
+        recording (save). DataLad does NOT parallelize your scripts for you.
+    """
 
     @staticmethod
     @datasetmethod(name='run')
--- a/datalad/core/local/run.py
+++ b/datalad/core/local/run.py
@@ -33,7 +33,10 @@ from datalad.interface.utils import generic_result_renderer
 from datalad.interface.utils import eval_results
 from datalad.interface.base import build_doc
 from datalad.interface.results import get_status_dict
-from datalad.interface.common_opts import save_message_opt
+from datalad.interface.common_opts import (
+    save_message_opt,
+    jobs_opt
+)
 
 from datalad.config import anything2bool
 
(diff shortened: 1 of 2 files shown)


bpoldrack:
> How do I edit the docstring for jobs_opt? I think we need to make it clear that the parallelization only applies to input get and output save, or some people might get the idea that datalad can parallelize scripts for them.

Good question. I don't think we do that anywhere (other than may be mentioning applicability in the command's docstring rather than the parameter's). The `Parameter` class isn't fit for modifications, indeed. However, it might be better to come up with a clearer wording for the general option instead. While for `run` it may be particularly misleading, it's a bit vague in general.

mih:
> * How do I edit the docstring for `jobs_opt`? I think we need to make it clear that the parallelization only applies to input `get` and output `save`, or some people might get the idea that datalad can parallelize scripts for them.

Take a look at: https://github.com/datalad/datalad/blob/dbc34478980c808a86b5531316c986abac953e37/datalad/distributed/create_sibling_github.py#L187

You can use this approach to easily amend a docstring with a follow-up statement. Maybe that is already sufficient.

AKSoo:
I appended a NOTE to the docstring.

As for my second question, I guess testing multiprocessing might be unnecessary and/or too difficult to unit test? If so, I would appreciate if someone could personally try the PR and double-check that both `get` and `save` use multiple jobs. I could only check by skimming debug logs.

mih (commented):
The diff looks good to me, thanks! I will give it a spin later today.

mih:
I can confirm that both `get` and `save` receive the `--jobs` specification and properly act on it.

Brief note on the impact:

The `get` operation works as one would expect, it gets multiple files at once. This is not the case for save (i.e., it does not save/add multiple files at once). In https://github.com/datalad/datalad/commit/a0bb0ef60086ea4a617549085a88a16215863119 @yarikoptic implemented parallelization for `save`, but only across datasets. In other words, if there are to-be-saved files in multiple subdatasets, these subdatasets would be saved in parallel.

Unfortunately, this is a bit of a special case. For the common case (read from subdatasets, write to top-level dataset) no parallelization is performed.

This could be easily added (by altering `save` to pass on the jobs value to the underlying `Repo` methods). However, with that dataset-parallelization in place we cannot do that in a straightforward fashion, because the `--jobs` would multiply (e.g., 4 files in 4 datasets simultaneously). I am afraid that this will need a bit of coordination first.

However, this is outside the scope of this PR (needs a `save`-internal change), hence I will merge it now.

Thanks again!

PR #6283, opened 2021-12-06
Add a missing import of CapturedException

christian-monch:
Fix an issue where a missing import would break exception handling during AnnexRepo.add_()

- [X] add import of CapturedException
- [X] add regression test to ensure correct exception path


mih (approved):
Thx

Issue #6269, opened 2021-12-01
Undocumented: push operation is needed before a GIN repo can be configured as a data source


mslw:
As established in #6204 if a new GIN repository is created and configured as a common data source (using `siblings configure` or `siblings add` with the `--as-common-datasrc` option) before the first push operation, the special remote gets configured without an annex UUID, resulting in problematic behaviour later on. One way to go about it would be to change the the behaviour of `siblings` so that it checks for the presence of annex at the remote location, and, if it's missing, either refuses to work or configures the remote through *some* mechanism. The other is to leave the behaviour of `siblings` as is.

If I read it correctly, the conclusion was that configuring the remote is best left out of `siblings` to keep it lightweight, and an explicit `push` should be an expected part of the workflow.

If that is the case, the following changes in documentation are needed:
- [x] The docstring in `create_siblings_gin` suggests that it can be followed directly by `siblings configure` https://github.com/datalad/datalad/blob/a9e52013429a35e1986b0299c1e5b0249b68bf34/datalad/distributed/create_sibling_gin.py#L98-L103
- [x] The handbook does not specify push before `siblings configure` either https://github.com/datalad-handbook/book/issues/779
- [x] other?

Please tag with cmd-siblings

mslw:
Closing because the docstring & handbook were addressed by the two pull requests above and there seem to be no other places which would require an update.

PR #6289, opened 2021-12-08
DOC: Update examples in create_sibling_gin

mslw:
This is a small change. The example demonstrating how to create a sibling repo on GIN and register it as a common data source now suggests `push` as an additional step before `configure` to create git-annex branch remotely and obtain annex UUID. Otherwise, configuration would not happen properly. Addresses part of #6269


mih (approved):
Makes sense! Thx!

PR #6291, opened 2021-12-09
Warn when saving Windows-incompatible file names

### LINKED ISSUE #5123: Issue warning when encountering windows-unfriendly filenames?
I think it could annoy and also be appreciated by some users if `save` (since the one introducing such names) would warn if names which cause windows indigestion, e.g. `con`, `prn` and others, would be alerted about.

adswa:
Windows has a set of illegal characters that arenot allowed  in file names and a list of reserved
file names that refer(ed) to system actions or devices:

https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file

Files with illegal characters or files using reserved file names can not be created on Windows
or received by Windows, and Git repositories containing such files will fail to be cloned.

Because non-Windows systems or even the Git bash on Windows have less restrictive rules
for file names, users can inadvertently create Windows-incompatible files that make their
dataset not interoperable with Windows systems.

To prevent this, based on a config switch ``datalad.save.windows-compat-warning``, this
change emits either a log message at warning level (``warning``), or renders an impossible 
result without saving (``error``) or ignores problems (``none``) for three potential incompatibilities
of a file name:

- reserved file names
- illegal characters
- file name ending with a dot or space

Fixes  #5123


datalad/support/gitrepo.py
```diff
@@ -3627,3 +3639,33 @@ def _fixup_submodule_dotgit_setup(ds, relativepath):
...
+                "Consider renaming %s to ensure cross-platform compatibility.",
+                chars, file)
+        # check for file name ending in a dot or space
+        if file.endswith('.') or file.endswith(' '):
+            lgr.warning(
+                "Interoperability warning: Windows systems can not "
+                "receive or work with files that end with spaces or dots. "
+                "Consider renaming %s ", file
+            )
+
```
yarikoptic, line 3671: That might lead to a flood of warnings  (e.g. in case of a single illegal char used consistently across many files).  I wonder if it would be better to collect all problematic files for each case, and then report a single warning per each case with a number and a list of relevant files.

datalad/support/gitrepo.py
```diff
@@ -98,6 +104,7 @@
 
 
 lgr = logging.getLogger('datalad.gitrepo')
+cfg = ConfigManager()
```
mih, line 107: This is not needed. Just use `self.config` below.

datalad/support/gitrepo.py
```diff
@@ -3424,6 +3431,11 @@ def save_(self, message=None, paths=None, _status=None, **kwargs):
             if (props.get('state', None) in ('modified', 'untracked') and
                 not (f in to_add_submodules or f in to_stage_submodules))}
         if to_add:
+            if not on_windows and \
+                    cfg.obtain("datalad.save.windows-compat-warning",
+                               default=True):
```
mih, line 3436: `self.config`

datalad/support/gitrepo.py
```diff
@@ -3627,3 +3639,33 @@ def _fixup_submodule_dotgit_setup(ds, relativepath):
...
+
+
+def _check_for_win_compat(files):
+    """Check file names for illegal characters or reserved names on Windows
+    
+    In the case that a non-Windows-compatible file is detected, warn users
+    about potential interoperability issues.
+    """
+    for file in files.keys():
+        f = Path(file).stem
```
mih, line 3651: I wonder whether this should cover directories too? The could also have illegal chars.

mih (commented):
Thx for approaching this. Left a few comments to consider.

datalad/support/gitrepo.py
adswa: oh you're right, I thought it would only be files, but indeed it applies to directories as well

datalad/support/gitrepo.py
adswa: true, I'll try to think of something more clever

datalad/support/gitrepo.py
adswa: thanks!

datalad/support/gitrepo.py
adswa: thanks, done!

datalad/support/gitrepo.py
adswa: done

datalad/support/gitrepo.py
adswa: please take a look at the current implementation.

adswa:
The Windows failure seems unrelated, but I have never seen it before:
```
FAIL: datalad.runner.tests.test_nonasyncrunner.test_timeout
    self.test(*self.arg)
  File "c:\projects\datalad\datalad\runner\tests\test_nonasyncrunner.py", line 427, in test_timeout
    assert_true(len(TestProtocol.received_timeouts) > 0)
AssertionError: False is not true
```

adswa:
I updated this PR since the last review according to the suggestions and comments, I would appreciate a second review but there's nothing urgent. :)

datalad/support/gitrepo.py
```diff
@@ -3424,16 +3429,35 @@ def save_(self, message=None, paths=None, _status=None, **kwargs):
...
-                                else tuple())}):
-                yield r
+            if to_add:
+                for r in self._save_add(
+                        to_add,
+                        git_opts=None,
+                        **{k: kwargs[k] for k in kwargs
+                           if k in (('git',) if hasattr(self, 'annexstatus')
+                                    else tuple())}):
+                    yield r
```
yarikoptic, line 3446: a minor pythonisation:

```suggestion
                yield from self._save_add(
                        to_add,
                        git_opts=None,
                        **{k: kwargs[k] for k in kwargs
                           if k in (('git',) if hasattr(self, 'annexstatus')
                                    else tuple())}):
```
? since _save_add  is a generator

datalad/support/gitrepo.py
```diff
@@ -3444,6 +3468,49 @@ def save_(self, message=None, paths=None, _status=None, **kwargs):
...
 
+    def _check_for_win_compat(self, files):
+        """Check file names for illegal characters or reserved names on Windows
+
+        In the case that a non-Windows-compatible file is detected, warn users
+        about potential interoperability issues.
+        """
+        from collections import defaultdict
+        problems = defaultdict(list)
+        for file in files.keys():
```
yarikoptic, line 3479: assuming files is a dict, just
```suggestion
        for file in files:
```
would make it to follow "generic programming" principle and make code more "generic" (e.g. files could be a list or a dict and still work)

datalad/support/gitrepo.py
```diff
@@ -3444,6 +3468,49 @@ def save_(self, message=None, paths=None, _status=None, **kwargs):
         # TODO yield result for commit, prev helper checked hexsha pre
         # and post...
 
+    def _check_for_win_compat(self, files):
+        """Check file names for illegal characters or reserved names on Windows
+
+        In the case that a non-Windows-compatible file is detected, warn users
+        about potential interoperability issues.
+        """
+        from collections import defaultdict
```
yarikoptic, line 3477: I guess if `'datalad.save.windows-compat-warning'` is `'none'` there is no point to go through all of this at all, right? then better just to return here if `none`

datalad/support/gitrepo.py
```diff
@@ -3444,6 +3468,49 @@ def save_(self, message=None, paths=None, _status=None, **kwargs):
...
+        problems = defaultdict(list)
+        for file in files.keys():
+            for part in Path(file).parts:
```
yarikoptic, line 3480: I worry a bit about performance overhead from this loop on heavy in number of files datasets.  I wonder if any timing was done?

yarikoptic (commented):
Minor suggestions and performance concern

datalad/support/gitrepo.py
adswa: No I haven't. What number of files should I test? Do you by any chance have an example benchmarking procedure with an increasing number of files?

datalad/support/gitrepo.py
adswa: I'm probably misunderstanding, but I thought we wouldn't go through this in that case because of the 
```
 if self.config.obtain("datalad.save.windows-compat-warning") in \
                    ('warning', 'error'):
 ``` 
 
 conditional above? Or should I add an explicit 
 
 ```
 if  self.config.obtain("datalad.save.windows-compat-warning") == "none":
    return
``` 

at the start of the method?

datalad/support/gitrepo.py
yarikoptic: we do not have anything formal unfortunately at the needed scale within our benchmarks :-/  I guess it would have been valuable to add a benchmark to "save 1000 files" or alike -- could come handy in any other use case when we change `save` functionality.

datalad/support/gitrepo.py
yarikoptic: oh, right -- it is done outside of this function.  I guess that is what confused me -- spread around logic.  Might be worth indeed "centralize" by
- bind `self.config.obtain("datalad.save.windows-compat-warning")` to a local variable within function (unlikely to feel effect on performance but still -- why to ask 3 times for the same thing?)
- moving this conditional inside  and `return to_add, None`
?

adswa: thanks, I have applied your suggestions

datalad/support/gitrepo.py
adswa: Also no formal comparison, but I ran the following script to general 10000 files and one incompatible file and save everything a few times:

```
#! /bin/bash
DIR=$(mktemp -d)
cd $DIR && datalad create -f .
echo "Writing 10000 files..."
for n in {1..10000}; do
    dd if=/dev/urandom of=file$( printf %03d "$n" ).bin bs=1 count=$(( RANDOM + 1024 )) status=none
done
echo "incompatible" > COM1.txt
time datalad save .
```

Outside of this branch it yields:

```
[INFO   ] Creating a new annex repo at /tmp/tmp.zf6J7c3bg0 
create(ok): /tmp/tmp.zf6J7c3bg0 (dataset)
Writing 10000 files...
add(ok): COM1.txt (file)                                                        
add(ok): file001.bin (file)                                                     
add(ok): file002.bin (file)                                                     
add(ok): file003.bin (file)                                                     
add(ok): file004.bin (file)                                                     
add(ok): file005.bin (file)                                                     
add(ok): file006.bin (file)                                                     
add(ok): file007.bin (file)                                                     
add(ok): file008.bin (file)                                                     
add(ok): file009.bin (file)                                                     
  [1 similar message has been suppressed; disable with datalad.ui.suppress-simil  [9991 similar messages have been suppressed; disable with datalad.ui.suppress-similar-results=off]
save(ok): . (dataset)                                                           
action summary:                                                                 
  add (ok: 10001)
  save (ok: 1)

real	0m23.287s
user	0m18.608s
sys	0m9.586s
bash wincompat  139.60s user 555.50s system 100% cpu 11:31.80 total
```

With the config set to ``warning`` I saw
```
action summary:                                                                                                   
  add (ok: 10001)
  save (ok: 1)

real	0m22.583s
user	0m18.283s
sys	0m9.228s
DATALAD_SAVE_WINDOWS__COMPAT__WARNING='warning' bash wincompat  147.58s user 593.18s system 95% cpu 12:56.23 total
```

with ``error``

```

real	0m23.890s
user	0m19.235s
sys	0m10.133s
DATALAD_SAVE_WINDOWS__COMPAT__WARNING='error' bash wincompat  148.41s user 895.40s system 100% cpu 17:20.15 total
```

and with ``none``

```

real	0m29.357s
user	0m23.892s
sys	0m12.222s
DATALAD_SAVE_WINDOWS__COMPAT__WARNING='none' bash wincompat  154.34s user 907.61s system 100% cpu 17:37.29 total
```

(a bit strange that the time increases without the check, but I hope that's just chance - I think the important benchmark is that there is no increase in time between the first and 2nd/3rd timings?)

adswa:
btw here is how this looks with the default level of 'warning' to a user:

```
(handbook2) adina@muninn in /tmp
❱ datalad create incompat && cd incompat
[INFO   ] Creating a new annex repo at /tmp/incompat 
create(ok): /tmp/incompat (dataset)
(handbook2) adina@muninn in /tmp/incompat on git:master
❱ echo 12345 | tee endswith. 'has?illegal:chars' COM1.txt 'should be good' >/dev/null
(handbook2) adina@muninn in /tmp/incompat on git:master
❱ datalad save .
[WARNING] Some elements of your dataset are not compatible with Windows systems. Disable this check by changing datalad.save.windows-compat-warning or consider renaming the following elements: 
| Elements using a reserved filename: ['COM1.txt']
| Elements ending with a dot: ['endswith.']
| Elements with illegal characters: ['has?illegal:chars'] 
add(ok): COM1.txt (file)                                                        
add(ok): endswith. (file)                                                       
add(ok): has?illegal:chars (file)                                               
add(ok): should be good (file)                                                  
save(ok): . (dataset)                                                           
action summary:                                                                 
  add (ok: 4)
  save (ok: 1)
```

and this is with ``error``:

```
(handbook2) adina@muninn in /tmp
❱ datalad create incompat && cd incompat
[INFO   ] Creating a new annex repo at /tmp/incompat 
create(ok): /tmp/incompat (dataset)
(handbook2) adina@muninn in /tmp/incompat on git:master
❱ echo 12345 | tee endswith. 'has?illegal:chars' COM1.txt 'should be good' >/dev/null
(handbook2) adina@muninn in /tmp/incompat on git:master
❱ DATALAD_SAVE_WINDOWS__COMPAT__WARNING='error' datalad save .
add(ok): should be good (file)                                                  
save(impossible): endswith. (file) [Incompatible name for Windows systems; disable with datalad.save.windows-compat-warning.]
save(impossible): COM1.txt (file) [Incompatible name for Windows systems; disable with datalad.save.windows-compat-warning.]
save(impossible): has?illegal:chars (file) [Incompatible name for Windows systems; disable with datalad.save.windows-compat-warning.]
save(ok): . (dataset)                                                           
action summary:                                                                 
  add (ok: 1)
  save (impossible: 3, ok: 1)
```

datalad/support/gitrepo.py
yarikoptic: those trailing `... bash wincompat  ... 11:31.80 total` to be ignored? (they seems to be high -- minutes, right?)

datalad/support/gitrepo.py
```diff
@@ -3444,6 +3466,63 @@ def save_(self, message=None, paths=None, _status=None, **kwargs):
...
+          list of files to add
+        config
+          value of self.config.obtain("datalad.save.windows-compat-warning"),
+          used to choose appropriate behavior. "none" performs no check,
+          "warning" warns in case of incompatibilities, and "error" results in
+          an error result in case of incompatibilities
+        """
+        # don't perform any check when the configuration is set to 'none'
+        if config == 'none':
+            return files, None
```
yarikoptic, line 3487: just for annoying me, please add some test run for the `none` so we ensure that there is no side effect and everything functions as expected?

datalad/support/gitrepo.py
adswa: I just wrote a (probably pretty inefficient) script that takes so long

yarikoptic (changes requested):
minor optimization which might make some difference requested and one question left.  I think we should proceed -- unlikely this extra processing would be of critical slow down in case of heavy trees.

bpoldrack:
Just a comment on your description of the result in https://github.com/datalad/datalad/pull/6291#issuecomment-1025458727, @adswa:

The effect of warning seems questionable to me. By the time a user would see it, it's too late to avoid a problem. Renaming as suggested by the message wouldn't solve it as the history already contains invalid names. It would require throwing that commit away in addition. So, I guess either the warning message should point that out by suggesting an actual fix or warning shouldn't be a mode of operation to begin with. I'm not sure, but I don't see a usecase for warning. Either you want to be compatible then you don't want to create a commit that can't be checked out, or you don't care.

yarikoptic:
although I agree that it is "too late", I do see having a warning better than to not have any feedback that possibly a problematic commit was created.
Situation is not unlike `git` complaining when committing whenever `git.user` config settings aren't yet setup.  Only that in case of `git` warning, it might be easier to "redo" that single commit, and in case of `datalad` it might be "more involved" (depending on either subdatasets were involved, or any file annexed) or as easy (just `git reset HEAD^`, mitigate, and `save` again).
What I wouldn't mind though is to switch from `warning` to `error` as the default value ;)

datalad/support/gitrepo.py
adswa: that makes sense, sure :)__

datalad/consts.py
```diff
@@ -73,3 +73,11 @@
 
 # format of git-annex adjusted branch names
 ADJUSTED_BRANCH_EXPR = re.compile(r'^adjusted/(?P<name>[^(]+)\(.*\)$')
+
+# Reserved file names on Windows machines
+RESERVED_NAMES_WIN = ['CON', 'PRN', 'AUX', 'NUL', 'COM1', 'COM2', 'COM3',
+                      'COM4', 'COM5', 'COM6', 'COM7', 'COM8', 'COM9', 'LPT1',
+                      'LPT2', 'LPT3', 'LPT4', 'LPT5', 'LPT6', 'LPT7', 'LPT8',
+                      'LPT9']
```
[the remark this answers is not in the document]
adswa: thx for the tip!

datalad/support/gitrepo.py
```diff
@@ -3444,6 +3466,63 @@ def save_(self, message=None, paths=None, _status=None, **kwargs):
...
+
+        from collections import defaultdict
+        problems = defaultdict(list)
+        for file in files:
+            for part in Path(file).parts:
+                # check every component of the path for incompatibilities
+                if Path(part).stem.upper() in RESERVED_NAMES_WIN:
+                    problems['Elements using a reserved filename:'].append(part)
+                    problems['paths'].append(file)
+                if re.findall(ILLEGAL_CHARS_WIN, part):
```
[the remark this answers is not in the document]
adswa: I tried to run the same test with multitime, with either ``findall`` or ``search``:

``re.findall`` yields

```
===> multitime results
1: python -m nose -s -v datalad/core/local/tests/test_save.py:test_windows_incompatible_names
real        4.245       0.291       3.761       4.268       4.696       
user        3.187       0.160       2.946       3.161       3.481       
sys         2.008       0.206       1.652       2.062       2.300       
multitime -n 15 python -m nose -s -v   47.81s user 30.11s system 88% cpu 1:27.59 total
```

``re.search`` yields

```
===> multitime results
1: python -m nose -s -v datalad/core/local/tests/test_save.py:test_windows_incompatible_names
real        4.306       0.257       3.952       4.297       4.909       
user        3.204       0.122       3.005       3.200       3.400       
sys         2.081       0.217       1.780       2.103       2.591       
multitime -n 15 python -m nose -s -v   48.06s user 31.22s system 95% cpu 1:23.33 total
```

datalad/support/gitrepo.py
```diff
@@ -3444,6 +3466,63 @@ def save_(self, message=None, paths=None, _status=None, **kwargs):
...
+          an error result in case of incompatibilities
+        """
+        # don't perform any check when the configuration is set to 'none'
+        if config == 'none':
+            return files, None
+
+        from collections import defaultdict
+        problems = defaultdict(list)
+        for file in files:
+            for part in Path(file).parts:
```
yarikoptic, line 3492: another concern/question came to my mind -- since you are going through parts, what if we have a folder which has forbidden name -- this will report for every file under that folder instead of the folder itself right? (e.g. folder `con/` having good number of files under it).  I think only the folder should be reported (once), so below not `file` should be added to `paths` but rather the path composed from the parts considered, and only if was not seen before, right?

datalad/support/gitrepo.py
yarikoptic: I will mark this conversation resolved.  We have optimized it a bit, so I should sleep better ;)

datalad/support/gitrepo.py
adswa: From the top of my head I *think* this is handled below in 

```
for k, v in problems.items():

+ # use the key as an explanation, and report filenames only once

+ msg += f"\n{k} {[*{*v}]}" if k != 'paths' else ''
```

But I will check 👍

adswa: Its all behaving nicely:

```
(handbook2) adina@muninn in /tmp/123 on git:master
❱ tree
.
└── con
    ├── file1
    ├── file2
    ├── file3
    └── illegal?

1 directory, 4 files
(handbook2) adina@muninn in /tmp/123 on git:master
❱ datalad save .
[WARNING] Some elements of your dataset are not compatible with Windows systems. Disable this check by changing datalad.save.windows-compat-warning or consider renaming the following elements: 
| Elements using a reserved filename: ['con']
| Elements with illegal characters: ['illegal?'] 
add(ok): con/file1 (file)                                                       
add(ok): con/file2 (file)                                                       
add(ok): con/file3 (file)                                                       
add(ok): con/illegal? (file)                                                    
save(ok): . (dataset)      
```

yarikoptic: oh sweet! well done!

datalad/support/gitrepo.py
```diff
@@ -3424,16 +3429,35 @@ def save_(self, message=None, paths=None, _status=None, **kwargs):
             if (props.get('state', None) in ('modified', 'untracked') and
                 not (f in to_add_submodules or f in to_stage_submodules))}
         if to_add:
+            compat_config = \
+                self.config.obtain("datalad.save.windows-compat-warning")
+            to_add, problems = self._check_for_win_compat(to_add, compat_config)
```
mih, line 3434: I think the encapsulation isnt optimal here. `compat_config` is localized to the helper function, but must be passed in from the outside.

mih (approved):
Left a minor note, but I understand that we might have different preferences here. Not worth holding up progress for a discussion. Thx!

PR #6292, opened 2021-12-09
CI: Update environment for windows CI builds

bpoldrack:


adswa (approved):
Maybe we also want to bump the git-annex version on ``maint``, though?

PR #6293, opened 2021-12-09
Fix invalid escape sequences

jwodder:
The code currently contains a number of invalid escape sequences (mostly escapes of special characters in regexes).  These result in Python emitting DeprecationWarnings, and some future Python version will instead treat them as syntax errors.  This PR fixes them.


mih (approved):
Thx!

Issue #6302, opened 2021-12-13
CI broken due to recent merges


mih:
```
======================================================================
4366FAIL: datalad.tests.test_api.test_consistent_order_of_args(<class 'datalad.local.foreach_dataset.ForEachDataset'>, {'cmd'})
4367----------------------------------------------------------------------
4368Traceback (most recent call last):
4369  File "/home/appveyor/dlvenv/lib/python3.8/site-packages/nose/case.py", line 198, in runTest
4370    self.test(*self.arg)
4371  File "/home/appveyor/projects/datalad/datalad/tests/test_api.py", line 62, in _test_consistent_order_of_args
4372    assert False
4373AssertionError
```

mih:
Fixed by #6305

PR #6304, opened 2021-12-13
BF: add positional args delimiter to `foreach_dataset`

christian-monch:
This PR fixes issue #6302 

The merged `foreach_dataset` was missing the positional argument delimiter, i.e. `*`, in its argument list


mih (approved):
Thx!

PR #6311, opened 2021-12-14
Fix runner timeout test in windows

christian-monch:
This PR fixes the runner timeout tests in windows. It uses "WAITFOR" instead of "TIMEOUT" since the latter does not work in non-interactive shells, e.g. in our test settings.


mih (approved):
Thx, LGTM!

Issue #6090, opened 2021-10-18
`download_url` cannot handle Path object input


mih:
```
  File "/home/mih/hacking/datalad/git/datalad/interface/download_url.py", line 132, in __call__
    dir_is_target = not path or path.endswith(op.sep)
AttributeError: 'PosixPath' object has no attribute 'endswith'
```

adswa:
fixed by #6317

PR #6317, opened 2021-12-15
BF: When download-url gets Pathobject as path convert it to a string

adswa:
Tiny fix for #6090

### DIFF
--- a/datalad/local/download_url.py
+++ b/datalad/local/download_url.py
@@ -144,7 +144,7 @@ class DownloadURL(Interface):
                          "ds": ds}
 
         got_ds_instance = isinstance(dataset, Dataset)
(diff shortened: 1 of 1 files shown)


adswa:
Just commenting to not let this PR become stale, but there is no urgency: This is ready for someone to look at, and a quick 4-character-change to approve or dismiss :)

yarikoptic:
merged, and CPed (had to manually fix path since file moved)/pushed to `maint` (what could go wrong, right? ;)).
Although indeed resolves the issue, IMHO it is just a band-aid for a more general issue of our code (in particular Interfaces) in many places expecting paths to be passed as string not as Path object.  May be in some bright future we would get more type annotations done to catch them in CI times

PR #6319, opened 2021-12-15
BF(Q&D): do not crash - issue warning - if template fails to format

yarikoptic:
This is to address an elderly https://github.com/datalad/datalad/issues/5631
which I keep running into over and over again.  Unfortunately a more thorough
fix attempt in https://github.com/datalad/datalad/pull/5644 was not finalized.
Since I keep running into this issue over and over again leading to process
crash, I decided that a "crude" fix with issuing a warning would be
better than just crashing entirely (thus possibly leaving other many subdatasets not even be attempted to be installed).

### DIFF
--- a/datalad/distribution/get.py
+++ b/datalad/distribution/get.py
@@ -47,6 +47,7 @@ from datalad.support.gitrepo import (
     _fixup_submodule_dotgit_setup,
 )
 from datalad.support.exceptions import (
+    CapturedException,
     CommandError,
     InsufficientArgumentsError,
 )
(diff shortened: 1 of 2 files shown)


adswa (approved):
I liked your suggestions on how to do this better in #5644 and it would still be nice to make it proper, but I definitely agree that it is better to not crash now than to have a perfect solution later.

yarikoptic:
Thank you @adswa ! With that approval, let's proceed then (appveyor single fail was unrelated)

PR #6320, opened 2021-12-16
BF: autorc - replace incorrect releaseTypes with "none"

yarikoptic:
Just blindly (although makes sense) following suggestion of @jwodder
in https://github.com/datalad/datalad/pull/5974#issuecomment-995219083

I will add release label and merge shortly (no need to wait for CI on this)


mih:
Great, thx!

Issue #6324, opened 2021-12-16
`Repo.dot_git` inconsistent


mih:
It promisses:

```
    Returns
    -------
    Path
      Absolute path to the resolved .git/ directory
```

but that is only True for a non-bare repo. It will not resolve for a bare repo.

Issue #6321, opened 2021-12-16
`Repo.dot_git` inefficient


mih:
I'd say the standard case for a repo setup is that it is a directory inside the repo, named `.git`.

However, the order of tests for this case is:

1. is it a file?
2. is it a symlink?
3. does it not exist?
4. does it not exist? (yes, twice)

For a bare repo (mostly unsupported) it is a bit shorter:

1. is it a file?
2. is it a symlink?
3. does it not exist? (plus to tests for files)

All other cases it is testing for are typically normalized away when datasets were established via datalad.

PR #6325, opened 2021-12-16
Reimplement _get_dot_git() to be more efficient

mih:
It now tests the common scenarios first, rather then leaving them for last.

Moreover, it introduces an explicit `resolved` flag for result path reporting, and applies it consistently when enabled -- rather than just for symlinked directories, but not files, or gitdirs specified in .git files. Searching through the code, I did not spot a case where this features is needed, nor desirable.

Fixes datalad/datalad#6321
Fixes datalad/datalad#6324

### DIFF
--- a/datalad/dataset/gitrepo.py
+++ b/datalad/dataset/gitrepo.py
@@ -671,7 +671,9 @@ def _get_dot_git(pathobj, *, ok_missing=False, resolved=False):
     Returns
     -------
     Path
-      Absolute path to the (resolved) .git directory
+      Path to the (resolved) .git directory. If `resolved` is False, and
+      the given `pathobj` is not an absolute path, the returned path will
+      also be relative.
     """
     dot_git = pathobj / '.git'
     if dot_git.is_dir():
--- a/datalad/dataset/gitrepo.py
+++ b/datalad/dataset/gitrepo.py
@@ -651,8 +651,8 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
 #
 # Internal helpers
 #
-def _get_dot_git(pathobj, *, ok_missing=False):
-    """Given a pathobj to a repository return path to .git/ directory
+def _get_dot_git(pathobj, *, ok_missing=False, resolved=False):
+    """Given a pathobj to a repository return path to the .git directory
 
     Parameters
     ----------
@@ -660,6 +660,8 @@ def _get_dot_git(pathobj, *, ok_missing=False):
     ok_missing: bool, optional
       Allow for .git to be missing (useful while sensing before repo is
       initialized)
+    resolved : bool, optional
+      Whether to resolve any symlinks in the path, at a performance cost.
 
     Raises
     ------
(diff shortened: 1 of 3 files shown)


yarikoptic:
FWIW, IIRC I think resolving any path was "accepted" as a practice a while ago for obscure setups like mine where an entire TMPDIR might be a symlink to somewhere else and then comparing resolved paths to non-resolved ones.  I think we have some travis matrix runs simulating some scenario like that ... one of the succeeds and another one fails, but there are other fails as well, so I guess those should be addressed first

mih:
> FWIW, IIRC I think resolving any path was "accepted" as a practice a while ago for obscure setups like mine where an entire TMPDIR might be a symlink to somewhere else and then comparing resolved paths to non-resolved ones. I think we have some travis matrix runs simulating some scenario like that ... one of the succeeds and another one fails, but there are other fails as well, so I guess those should be addressed first

Hmm, I thought this was all about `Repo.path` being resolved, and `Dataset.path` not. But what is happening here is resolving `Repo.dot_git` beyond `Repo.path`.

I am not saying that it has to be one or the other. However the current state is broken. The documentation says

```
Given a pathobj to a repository return path to .git/ directory
```

and later

```
Absolute path to resolved .git/ directory
```

(let's ignore misleading POSIX notation)

So the return value must be a resolved path. However the code looks like this (note that I labeled the conditionals for easier reference and also note that they are mutually exclusive):

```py
    dot_git = pathobj / '.git'
    # 1
    if dot_git.is_file():
        with dot_git.open() as f:
            line = f.readline()
            if line.startswith("gitdir: "):
                dot_git = pathobj / line[7:].strip()
            else:
                raise InvalidGitRepositoryError("Invalid .git file")
    # 2
    elif dot_git.is_symlink():
        dot_git = dot_git.resolve()
    # 3
    elif not dot_git.exists() and \
            (pathobj / 'HEAD').exists() and \
            (pathobj / 'config').exists():
        # looks like a bare repo
        dot_git = pathobj
    # 4
    elif not (ok_missing or dot_git.exists()):
        raise RuntimeError("Missing .git in %s." % pathobj)
    return dot_git
```

(1) IF we have an *existing* (symlinked) .git file that points to a symlinked location: *not resolved*

(2) ELIF we have a symlinked directory or non-existing file: resolved (and since PY3.6 no longer crash in case of the symlink to a non-existing file)

(3) ELIF we have a bare repo: *not resolved*

(4) ELIF `.git` exists or it is OK not to exist: *not resolved*

(also not that `ok_missing=False` with a symlinked .git pointing to a non-existing path will not raise an exception).

So when we are talking about acceptance here, it is the acceptance implied by "my uncle is a lunatic, there is nothing I can do about it". ;-) I would prefer to sort this out and fix it up, even if that means a behavior change.

mih:
> From chat with @bpoldrack 

I don't currently think it's [resolving the path] necessary. But: If we ever switch the design, to represent the repository and a/the worktree separately, then the identifier for the actual Repo needs to become the resolved .git I think

Don't see another reason ATM

I think it's OK to change the test accordingly. It may have only be done for

- sense of consistency: If it's bare, it's resolved already (== path), so have it resolved always.
- Anticipation of changing *Repo's flyweight ID to the resolved .git, rather than the worktree path.

PR #6336, opened 2022-01-04
Install all data files under `datalad/`

jwodder:
* The contents of the `datalad/test/ca/` folder are currently not included when datalad is installed; this PR fixes that.
* `datalad/metadata/tests/data/nifti1.nii.gz` is currently not included when datalad is installed; this PR fixes that.
* The `package_data` specification in `setup.py` listed a number of file extensions to include under `datalad/resources/` that are no longer present; this PR removes them.
* The package data specification is simplified so that all files under `datalad/` are included when installing datalad without needing to list all of the desired files' file extensions.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,11 +6,9 @@
 #
 # ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
 
-from os.path import (
-    dirname,
-    join as opj,
-)
 import sys
+from os.path import dirname
+from os.path import join as opj
 
 # This is needed for versioneer to be importable when building with PEP 517.
 # See <https://github.com/warner/python-versioneer/issues/193> and links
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -196,13 +187,6 @@
...
     cmdclass=cmdclass,
-    package_data={
-        'datalad':
-            findsome('resources',
-                     {'sh', 'html', 'js', 'css', 'png', 'svg', 'txt', 'py'}) +
-            findsome(opj('downloaders', 'configs'), {'cfg'}) +
-            findsome(opj('distribution', 'tests'), {'yaml'}) +
-            findsome(opj('metadata', 'tests', 'data'), {'mp3', 'jpg', 'pdf'})
-    },
+    include_package_data=True,
```
yarikoptic, line 190: I wondered if we had it filtered on purpose.  But probably not really on purpose.  I guess it should be safe and we would not distribute something we do not actually like to distribute in the binary distribution.

yarikoptic (approved):

mih (approved):
Good to go! Thx!

Issue #6237, opened 2021-11-23
1 ERROR and 1 FAIL on windows while testing recent git-annex on master


yarikoptic:
Started to happen awhile back: example recent run  https://github.com/datalad/git-annex/runs/4294977555?check_suite_focus=true

1.

```
======================================================================
FAIL: datalad.distribution.tests.test_install.test_install_dataset_from_just_source
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\site-packages\nose\case.py", line 198, in runTest
    self.test(*self.arg)
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\site-packages\datalad\tests\utils.py", line 1157, in _wrap_with_testrepos
    t(*(arg + (uri,)), **kw)
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\site-packages\datalad\tests\utils.py", line 856, in _wrap_with_tempfile
    return t(*(arg + (filename,)), **kw)
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\site-packages\datalad\distribution\tests\test_install.py", line 248, in test_install_dataset_from_just_source
    assert_repo_status(ds.path, annex=None)
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\site-packages\datalad\tests\utils.py", line 1738, in assert_repo_status
    % (state, state_files, oktobefound))
AssertionError: unexpected content of state "deleted": [WindowsPath('C:/Users/runneradmin/AppData/Local/Temp/datalad_temp_m6cumsrj/testrepo--basic--r1/test-annex.dat')] != []
```

has been failing for a while (didn't check prior month)

```shell
$> git grep 'datalad.distribution.tests.test_install.test_install_dataset_from_just_source\>' | grep windows | grep master
cron-20211101/build-windows.yaml-439-280a49ed-failed/1_test-datalad (master).txt:2021-11-01T04:41:06.7132507Z datalad.distribution.tests.test_install.test_install_dataset_from_just_source ... FAIL
cron-20211101/build-windows.yaml-439-280a49ed-failed/1_test-datalad (master).txt:2021-11-01T06:06:16.3347737Z FAIL: datalad.distribution.tests.test_install.test_install_dataset_from_just_source
[...]

PR #6352, opened 2022-01-18
BF(?WIN): provide relpath relative to dataset not default CWD

yarikoptic:
Trying to address the `#2` fail listed in https://github.com/datalad/datalad/issues/6237 and happening somehow only on master . Due to absent comments I assume that `relpath` was used to provide a shortened path to the message file.  Given that `dataset` could point somewhere else (including another drive on windows) and `os.curdir` is used by relpath if source is not specified , I decided to just provide `start` to be the dataset path explicitly.  
I failed to reproduce #6237 locally, so it is somewhat a wild shot, but what could go wrong?  only reporting could be effected.

### DIFF
--- a/datalad/core/local/run.py
+++ b/datalad/core/local/run.py
@@ -819,7 +819,9 @@ def run_command(cmd, dataset=None, inputs=None, outputs=None, expand=None,
     if not rerun_info and cmd_exitcode:
         if do_save:
             repo = ds.repo
-            msg_path = relpath(opj(str(repo.dot_git), "COMMIT_EDITMSG"))
+            msg_path = opj(str(repo.dot_git), "COMMIT_EDITMSG")
+            # shorten to the relative path for a more concise message
+            msg_path = relpath(msg_path, ds_path)
             with open(msg_path, "wb") as ofh:
                 ofh.write(ensure_bytes(msg))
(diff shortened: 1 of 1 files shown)


mih:
Thx!

I cannot submit a proper suggesting, so doing it like this:

was:

```py
    msg_path = opj(str(repo.dot_git), "COMMIT_EDITMSG")
    # shorten to the relative path for a more concise message
    msg_path = relpath(msg_path, ds_path)
    with open(msg_path, "wb") as ofh:
        ofh.write(ensure_bytes(msg))
    lgr.info("The command had a non-zero exit code. "
                "If this is expected, you can save the changes with "
                "'datalad save -d . -r -F %s'",
                msg_path)
```

could be

```py
    msg_path = repo.dot_git / "COMMIT_EDITMSG"
    msg_path.write_text(msg)
    lgr.info("The command had a non-zero exit code. "
                "If this is expected, you can save the changes with "
                "'datalad save -d . -r -F %s'",
                # shorten to the relative path for a more concise message
                msg_path.relative_to(ds.pathobj)
```

yarikoptic:
thanks @mih, applied

yarikoptic:
ok, changed code is tests covered (not sure how much checked) and I took recommended changes. Let's proceed and improve/fix if needed after. Cheers!

Issue #6333, opened 2021-12-21
test_ria_push: fails on datalad/git-annex windows tests of datalad master


yarikoptic:
Complimentary to https://github.com/datalad/datalad/issues/6237 where two first ones were reported

```
======================================================================
ERROR: datalad.core.distributed.tests.test_push.test_ria_push
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\site-packages\nose\case.py", line 198, in runTest
    self.test(*self.arg)
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\site-packages\datalad\tests\utils.py", line 851, in _wrap_with_tempfile
    return t(*(arg + (filename,)), **kw)
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\site-packages\datalad\tests\utils.py", line 556, in _wrap_with_tree
    return t(*(arg + (d,)), **kw)
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\site-packages\datalad\core\distributed\tests\test_push.py", line 527, in test_ria_push
    "datastore", new_store_ok=True))
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\site-packages\datalad\distribution\dataset.py", line 502, in apply_func
    return f(*args, **kwargs)
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\site-packages\datalad\interface\utils.py", line 494, in eval_func
    return return_func(*args, **kwargs)
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\site-packages\datalad\interface\utils.py", line 486, in return_func
    results = list(results)
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\site-packages\datalad\interface\utils.py", line 473, in generator_func
    msg="Command did not complete successfully")
datalad.support.exceptions.IncompleteResultsError: Command did not complete successfully. 1 failed:

[{'action': 'create-sibling-ria',
  'message': 'initremote failed.\n'
             'stdout: initremote datastore-storage \n'
             'failed\n'
             '\n'
             'stderr: git-annex: Configured RIA store not found at '
             'ria+file:///C/Users/runneradmin/AppData/Local/Temp/datalad_temp_tree_6dfju520 \r\n'
             'initremote: 1 failed\r\n',
[...]

PR #6353, opened 2022-01-18
BF(workaround,TST): mark two tests as @known_failure_githubci_win

yarikoptic:
Those had been failing for a while, see
https://github.com/datalad/datalad/issues/6333
https://github.com/datalad/datalad/issues/6237
ruining the baseline testing of datalad/git-annex.  Unfortunately I have no
sufficient foo or time to track them down.

### DIFF
--- a/datalad/core/distributed/tests/test_push.py
+++ b/datalad/core/distributed/tests/test_push.py
@@ -511,6 +511,7 @@ def test_force_checkdatapresent(srcpath, dstpath):
                       message='Slated for transport, but no content present')
 
 
+@known_failure_githubci_win
 @with_tempfile(mkdir=True)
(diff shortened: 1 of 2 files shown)


yarikoptic:
even codecov is happy ;)

FWIW here is current stats
```
$> git grep -h '^ *@known_failure_' | f1 | sort | uniq -c | sort -n -r
     36 @known_failure_windows
     36 @known_failure_githubci_win
      2 @known_failure_osx
      2 @known_failure_githubci_osx
      1 @known_failure_direct_mode
```

yarikoptic:
ok, I think this one is a nobrainer and I would like to make datalad/git-annex green again on windows. Let's proceed.

PR #6355, opened 2022-01-19
BF: ORA over HTTP tried to check archive

### LINKED ISSUE #6354: `fsck` fails: 'HTTPRemoteIO' object has no attribute 'in_archive') failed
This is a ria/ora remote configured with `type=external url=ria+http://enki.ds.inm7.de`.

```
% git annex fsck --fast -f inm7-storage
fsck dataset_description.json [INFO] Fetching 'http://enki.ds.inm7.de/ria-layout-version' 
[INFO] Fetching 'http://enki.ds.inm7.de/6e6/61818-b617-11ea-ae2d-b4969157768c/ria-layout-version' 
('HTTPRemoteIO' object has no attribute 'in_archive') failed
fsck desc-aparcaseg_dseg.tsv ('HTTPRemoteIO' object has no attribute 'in_archive') failed
...
```

Whether or not `--fast` is used is irrelevant.

Source of the error is

```
[2022-01-19 11:17:54.833641188] /home/mih/venvs/datalad/bin/git-annex-remote-ora[1] --> CHECKPRESENT-UNKNOWN MD5E-s49926--3c24235cafe15cbf27fe67b55bbe7815.tsv 'HTTPRemoteIO' object has no attribute 'in_archive'
('HTTPRemoteIO' object has no attribute 'in_archive') failed
```

Also happening with 0.15.3

And the reason is that `HTTPRemoteIO` is used as an `IOBase` instance -- which it is simply not.

bpoldrack:
Commit a4b87f0c58047c9337e0cd1604f2b3eaeaae99a7 broke this
by removing to much. CHECKPRESENT and TRANSFER-RETRIEVE would
try to address a potential archive if key couldn't be
found in annex object tree.

This is wrong. In opposition to remote execution variants
(SSH and local access), HTTP is supposed to only send one request
per key - archives have to be dealt with server-side.

(Closes #6354)

### DIFF
--- a/datalad/distributed/tests/test_ora_http.py
+++ b/datalad/distributed/tests/test_ora_http.py
@@ -115,7 +115,17 @@ def test_read_access(store_path, store_url, ds_path):
     create_store(io, store_path, '1')
     create_ds_in_store(io, store_path, ds.id, '2', '1')
     ds.repo.init_remote('ora-remote', options=init_opts)
-    ds.repo.fsck(remote='ora-remote', fast=True)
+    fsck_results = ds.repo.fsck(remote='ora-remote', fast=True)
+    # Note: Failures in the special remote will show up as a success=False
+    # result for fsck -> the call itself would not fail.
+    for r in fsck_results:
+        if "note" in r:
+            # we could simply assert "note" to not be in r, but we want proper
+            # error reporting - content of note, not just its unexpected
+            # existence.
+            assert_equal(r["success"], "true",
+                         msg="git-annex-fsck failed with ORA over HTTP: %s" % r)
+        assert_equal(r["error-messages"], [])
     store_uuid = ds.siblings(name='ora-remote',
                              return_type='item-or-list',
                              result_renderer='disabled')['annex-uuid']
--- a/datalad/distributed/ora_remote.py
+++ b/datalad/distributed/ora_remote.py
@@ -1286,6 +1286,12 @@ class RIARemote(SpecialRemote):
         try:
             self.io.get(abs_key_path, filename, self.annex.progress)
         except Exception as e1:
+            if isinstance(self.io, HTTPRemoteIO):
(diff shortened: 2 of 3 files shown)


bpoldrack:
FTR: Converted draft, since the HTTP test has bug, too (which is why the issue wasn't caught).

datalad/distributed/tests/test_ora_http.py
```diff
@@ -115,7 +115,17 @@ def test_read_access(store_path, store_url, ds_path):
...
+    # Note: Failures in the special remote will show up as a success=False
+    # result for fsck -> the call itself would not fail. Hence, assert
+    # nothing special to be in "note" or "error-messages".
+    for r in fsck_results:
+        if "note" in r:
+            # we could simply assert "note" to not be in r, but we want proper
+            # error reporting - content of note, not just its unexpected
+            # existence.
+            assert_equal(r["note"], "")
+        assert_equal(r["error-messages"], [])
```
mih, line 128: Why is it not checking for the `success=true` result property? This is the direct indicator, while `note` and `error-messages` are more like proxies.

mih: ?

bpoldrack: I wanted the message to show up, when the assertion fails. But you're right - can be achieved differently.

yarikoptic:
I was surprised that diff code is not coverage covered, recalled yesterdays discussion of coverage of subprocesses, pushed 5927b9f7e7ee31c237704ae3d602363debb6bd44 to see if "shimming" I had for other custom remotes still works. If does -- diff lines in ora code should get green. Free free to drop if of no effect, but might be worth keeping even if just for consistency.

mih:
Change seems to have no effect on the coverage with the present setup, but thx for trying nevertheless!

PR #6359, opened 2022-01-20
BF(TST): clean command should be invoked within a dataset

yarikoptic:
Likely this test never failed to run (until it was ran during
Debian package build) since was always invoked within git clone
of datalad source code base.  So "clean" was able to find the dataset
to "work on".  When building a package, there were no git repository
and test failed since "clean" exited with 2 instead of 0.


yarikoptic:
trivial and only fixing a test, will merge (and drop patch from perspective debian package for upcoming 0.16.0)

PR #6371, opened 2022-01-25
Allow multiple threads in ProducerConsumer with python <= 3.7

christian-monch:
The new runner implementation allows for threads in ProducerConsumer in python versions <= 3.7 because we
do not use asyncio any longer in the runner

### DIFF
--- a/datalad/support/parallel.py
+++ b/datalad/support/parallel.py
@@ -115,10 +115,6 @@ class ProducerConsumer:
     codebase `addurls.py`, `get.py`, `save.py`, etc.
     """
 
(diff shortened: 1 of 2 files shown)


christian-monch:
> We are about to have 3.7 as minimal supported version in master

Makes sense then for the 3.7 versions then

yarikoptic:
Thank you @christian-monch !
looks good to me and I guess could even go against `maint`  (I would have liked it since I kept running into this warning for quite a while and it annoyed me, but I upgraded at least one of the hosts since then). I am ok either way.  To not wait for it and since CI is happy, will merge as is for now but will add a label for potential CP to maint.

PR #6375, opened 2022-01-26
Annotate API with ONLY-USED-IN-CLI

mih:
These functions/methods are exclusively used by the CLI, but are located
in other parts of the codebase.

I will pull them into my effort to produce a self-contained CLI code
base. If that effort is finished, these can be removed.


mih:
Closing in favor of a more comprehensive demo https://github.com/datalad/datalad/pull/6378

PR #6381, opened 2022-01-27
Make room for dedicated CLI benchmarks + cleanup

mih:
This is only to establish a set of benchmarks to be able to evaluate a follow-up PR with a major CLI refactoring.

### DIFF
--- a/benchmarks/cli.py
+++ b/benchmarks/cli.py
@@ -1,0 +1,48 @@
+# ex: set sts=4 ts=4 sw=4 noet:
+# ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+#
+#   See COPYING file distributed along with the datalad package for the
+#   copyright and license terms.
+#
+# ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+"""Benchmarks for DataLad CLI"""
+
+import os
+import sys
+import os.path as osp
+
+from subprocess import call
+from .common import SuprocBenchmarks
+
+
+class startup(SuprocBenchmarks):
+    """
+    Benchmarks for datalad command startup
+    """
+
+    def setup(self):
+        # we need to prepare/adjust PATH to point to installed datalad
+        # We will base it on taking sys.executable
+        python_path = osp.dirname(sys.executable)
+        self.env = os.environ.copy()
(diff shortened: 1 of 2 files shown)


benchmarks/cli.py
```diff
@@ -0,0 +1,48 @@
...
+    """
+    Benchmarks for datalad command startup
+    """
+
+    def setup(self):
+        # we need to prepare/adjust PATH to point to installed datalad
+        # We will base it on taking sys.executable
+        python_path = osp.dirname(sys.executable)
+        self.env = os.environ.copy()
+        self.env['PATH'] = '%s:%s' % (python_path, self.env.get('PATH', ''))
```
yarikoptic, line 28: I see that it is a copy/paste from the core's `setup`.  Ideally we should avoid that. May be `setup` in SuprocBenchmarks should just do that for all benchmarks?

And even though it is a copy, and probably benchmarks code is *nix specific ATM, but might be worth using `os.pathsep` instead of hardcoded `:`.  

```suggestion
        self.env['PATH'] = f'{python_path}{os.pathsep}{self.env.get("PATH", "")}')
```
assuming that `os` is imported above.

benchmarks/cli.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def time_usage_advice(self):
+        call(["datalad"], env=self.env)
+
+    def time_short_help(self):
+        call(["datalad", "-h"], env=self.env)
+
+    def time_help_np(self):
+        call(["datalad", "--help-np"], env=self.env)
```
yarikoptic, line 37: we already have some of those in another location

```
(git)lena:~datalad/datalad-master[master]benchmarks
$> git grep -e --help
core.py:        call(["datalad", "--help-np"], env=self.env)

```

I guess it would be good to keep them there to be able to do continuous benchmarking? or should they be removed from there?

benchmarks/cli.py
mih: I am not sure what these lines do. I just copied them to have something working. I am fine to revert the entire PR after the CLI is merged. Will file an issue.

mih:
Thx @yarikoptic for the review. I am running out of time, so I decided to ignore the benchmakrs but focus on the CLI instead. I hope this is OK, otherwise we can also approach the CLI merge without expanded benchmarks.

PR #6391, opened 2022-01-29
Key interface documentation

mih:
- Fixes #6377

#### 📝 Documentation

-  The `datalad.interface.base.Interface` class, the basis of all DataLad command implementations, has been extensively documented to provide an overview of basic principles and customization possibilities (by @mih)

#### 🪓 Deprecations and removals

- `datalad.interface.common_opts.eval_default` has been deprecated. All (command-specific) defaults for common interface parameters can be read from `Interface` class attributes (by @mih)

#### 🏠 Internal

- `datalad.interface.base.Interface` is now an abstract class [#6391](https://github.com/datalad/datalad/pull/6391) (by @mih)

PR #6394, opened 2022-02-01
Simplify result rendering, and fix #6393

### LINKED ISSUE #6393: #6391 broke CLI generic result rendering
After https://github.com/datalad/datalad/pull/6391 (but not the other CLI changes), CLI-based result rendering is broken for the default generic renderer.

I suspect that the CLI is employing the same result transformation now, e.g. converting a created dataset's result record into an actual dataset instance -- which can then not be rendered anymore.

No, the cause is that the global default was `tailored`, which is now written to all classes, hence is the active default for all commands, unless overwritten. This class attribute is now determining the result renderer default stored in the CLI parser. But later this value is only considered by the CLI's execution handler if it is NOT `tailored`, otherwise the class is queried again for the value, which is again determined to be `tailored`. The last puzzle piece is that the result rendering handler in `datalad.interface.utils._process_results()` will do nothing, when `tailored` is requested, but a class does not actually have a `custom_result_renderer` attribute.

This is ... complex .... "logic".

Intuitively, I'd say that `tailored` default should only be in effect, if there is a `custom_result_renderer` in a class.

mih:
#### 🐛 Bug Fixes
- Restore default result rendering behavior broken by #6391. Fixes #6393
- An unused code path for result rendering was removed from the CLI `main()`
#### 🏠 Internal
- Simplified the decision making for result rendering, and reduced code complexity.

### DIFF
--- a/datalad/cli/main.py
+++ b/datalad/cli/main.py
@@ -123,20 +123,12 @@ def main(args=sys.argv):
         sys.exit(2)
 
     # execute the command, either with a debugger catching
-    # a crash, or with a simplistic exception handler
+    # a crash, or with a simplistic exception handler.
+    # note that result rendering is happening in the
+    # execution handler, when the command-generator is unwound
     ret = _run_with_debugger(cmdlineargs) \
         if cmdlineargs.common_debug or cmdlineargs.common_idebug \
         else _run_with_exception_handler(cmdlineargs)
-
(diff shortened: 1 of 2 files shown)


adswa:
I've tried this, and it fixes the result rendering and messaging for me

mih:
Thx @adswa ! This could be merged from my POV.

PR #6405, opened 2022-02-02
Fix fallout from CLI update

### LINKED ISSUE #6404: `datalad -h/--help` does not list extension command
But they are functional and their own help is accessible via the CLI. Likely fallout from the CLI update.

mih:
- Fixes #6404
- Add compact-shims for old CLI entrypoints (needed for extension doc builds)

Does not need to appear in a release changelog.

### DIFF
--- a/datalad/cli/parser.py
+++ b/datalad/cli/parser.py
@@ -116,6 +116,15 @@ def setup_parser(
         completing,
     ) if not (return_subparsers or help_ignore_extensions) else None
 
+    if target_subparser_name is None:
+        # we cannot be lean, we must built the full parser
+        # and for that must load all extensions too, e.g., to
+        # be able to list extension command in --help output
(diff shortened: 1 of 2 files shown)


adswa:
I tried this, and it fixes the issue for me

mih:
The benchmark impact is unavoidable. The issue here is that it did not process the extensions before, and now does.

yarikoptic:
appveyor fail seems to be due to running out of time since aligns to an hour:
```
$> tail -n 1 /tmp/log.txt
[00:59:58] uninstall(ok): subdir/subds_unavailable_clean (dataset)
```
?  shouldn't be related this much though since we rarely invoke in tests via CLI (besides special remotes, not sure if those are anyhow affected)

mih:
Yeah, that run pains me, need to split. But should be good to merge.

Issue #6410, opened 2022-02-03
Status missing in Python API docs


adswa:
http://docs.datalad.org/en/stable/modref.html doesn't list it.

adswa:
Fixed by  #6411

PR #6411, opened 2022-02-03
DOC: include status in Python API overview

adswa:
a quick fix for #6410, hopefully.

(no changelog entry needed)


mih (approved):
Ultimately, we should reorganize the modref layout to match the CLI one again. But this will do for now. thx!

Issue #6076, opened 2021-10-14
Remove or fix AnnexRepo.annexstatus()


christian-monch:
### What is the problem?

1. The function `AnnexRepo.annexstatus` is not used outside of tests.
2. The function has an unused `untracked`-parameter
3. The function is missing  doc-string

adswa:
I'm currently trying to remove this function, but its surprisingly difficult. The gotcha that Michael found in save is addressed easily, and I can also adjust the tests to rely on ``get_content_annexinfo`` instead.

But if I remove ``annexstatus`` at a time where *nothing* uses or calls this function anymore everything falls to pieces. WTF?

adswa:
Just for posterity, here's the diff that leads to ``annexstatus`` becoming unused:

```diff
commit 97821f73bc5d37a4ab65019dda0614910d947a3a (HEAD -> mnt-6076)
Author: Adina Wagner <adina.wagner@t-online.de>
Date:   Thu Feb 3 09:39:13 2022 +0100
 
    MNT: Determine ds instance type based on uuid property instead of annexstatus
    
    The latter is a negligably used, to-be deprecated function, while UUID should
    be a reliable indicator of AnnexRepo instances as it is a property of
    AnnexRepo
 
diff --git a/datalad/core/local/save.py b/datalad/core/local/save.py
index fa8718fdd..e3684d974 100644
--- a/datalad/core/local/save.py
+++ b/datalad/core/local/save.py
@@ -309,7 +309,7 @@ class Save(Interface):
                         # calls
                         paths=None,
                         # prevent whining of GitRepo
-                        git=True if not hasattr(ds.repo, 'annexstatus')
+                        git=True if not hasattr(ds.repo, 'uuid')
                         else to_git,
                         # we are supplying the full status already, do not
                         # detect anything else

```
```diff
commit 0d829c1cdbf3b2178c71e2caf0a7cdc2c2c703bd
Author: Adina Wagner <adina.wagner@t-online.de>
Date:   Thu Feb 3 09:27:02 2022 +0100
 
    RF: Deprecate annexstatus, move it from a class method into a test helper
    
    This function was formerly used only in tests, with one exception in save
    that relied on its precence as an attribute in ds.repo to determine
    whether or not ds.repo was an AnnexRepo or GitRepo instance.
    
    Tests that desperately rely on this, or test its function, have been
    adjusted to use the test helper instead.
    
    Work towards #6067
 
[...]

PR #6413, opened 2022-02-03
MNT: Deprecate annexstatus

adswa:
This PR works towards #6076, which calls out the ``AnnexRepo``'s ``annexstatus`` method is undocumented and mostly unused with the exception of tests. 

Its remaining usecases are addressed as follows:
- A few spots (in ``save`` and ``GitRepo``) rely on the existence of this method in ``ds.repo`` to assert whether or not we're dealing with an AnnexRepo. These spots were changed to instead check for a ``uuid`` property, which should exist in AnnexRepos but not in GitRepos
- A few tests were using ``annexstatus``. Some were able to change to use ``get_content_annexinfo``. For those that didn't, I moved ``annexstatus`` into ``tests/utils`` as a new test helper ``get_annexstatus``.

#### 🏠 Internal
- AnnexRepo's internal ``annexstatus`` is deprecated. In its place, a new test helper assists the few tests that rely on it.

### DIFF
--- a/datalad/core/local/save.py
+++ b/datalad/core/local/save.py
@@ -309,7 +309,7 @@ class Save(Interface):
                         # calls
                         paths=None,
                         # prevent whining of GitRepo
-                        git=True if not hasattr(ds.repo, 'annexstatus')
+                        git=True if not hasattr(ds.repo, 'uuid')
                         else to_git,
                         # we are supplying the full status already, do not
                         # detect anything else
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -3334,6 +3334,12 @@ class AnnexRepo(GitRepo, RepoInterface):
         return info
 
     def annexstatus(self, paths=None, untracked='all'):
+        """
+        .. deprecated:: 0.16
+
+        Use get_content_annexinfo() or the test helper
+        datalad/utils/get_annexstatus() instead.
+        """
         info = self.get_content_annexinfo(
             paths=paths,
             eval_availability=False,
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -3336,9 +3336,8 @@ class AnnexRepo(GitRepo, RepoInterface):
     def annexstatus(self, paths=None, untracked='all'):
         """
         .. deprecated:: 0.16
-
-        Use get_content_annexinfo() or the test helper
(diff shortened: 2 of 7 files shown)


adswa:
Travis failure seems unrelated
```

ERROR: datalad.distribution.tests.test_get.test_source_candidate_subdataset
  File "/tmp/dl-miniconda-chiq2wk1/lib/python3.9/site-packages/nose/case.py", line 134, in run
    self.runTest(result)
  File "/tmp/dl-miniconda-chiq2wk1/lib/python3.9/site-packages/nose/case.py", line 152, in runTest
    test(result)
AssertionError: Test was too slow (took 30.1101s, threshold was 30.0000s)
```

and Appveyor failure is a coverage glitch:
```
python -m coverage xml
C:\Python310\python.exe: No module named coverage
Command exited with code 1
```

datalad/support/annexrepo.py
```diff
@@ -3334,6 +3334,12 @@ def get_content_annexinfo(
         return info
 
     def annexstatus(self, paths=None, untracked='all'):
+        """
+        .. deprecated:: 0.16
+
+        Use get_content_annexinfo() or the test helper
+        datalad/utils/get_annexstatus() instead.
```
mih, line 3341:
```suggestion
        .. deprecated:: 0.16
            Use get_content_annexinfo() or the test helper
            datalad/utils/get_annexstatus() instead.
```

mih (commented):
LGTM, thx! Small indentation issue, fix suggested.

mih (approved):
I merged the doc-fix. Once tests passed, this can move in. Thx!

datalad/core/local/save.py
```diff
@@ -309,7 +309,7 @@ def save_ds(args, version_tag=None):
                         # calls
                         paths=None,
                         # prevent whining of GitRepo
-                        git=True if not hasattr(ds.repo, 'annexstatus')
+                        git=True if not hasattr(ds.repo, 'uuid')
```
bpoldrack, line 312: I'd prefer to use that occasion to unify this kind of test:

```suggestion
                        git=True if not isinstance(ds.repo, AnnexRepo)
```

bpoldrack: Not a blocker, though.

mih: Let's do this elsewhere.

PR #6416, opened 2022-02-03
DOC: run - adjust description for -i/-o to mention that it could be a directory

yarikoptic:
AFAIK, those could point to directories since otherwise such option
would not scale for many realistic scenarios.  But @djarecka mentioned
that docs explicitly mention file only, not directories, which led her
to believe that it must point to a file.

May be wording could be adjusted, I was aiming for the shortest way to express myself.  Or should we switch to use `path` instead of `file`?

### DIFF
--- a/datalad/core/local/run.py
+++ b/datalad/core/local/run.py
@@ -200,7 +200,7 @@ class Run(Interface):
             metavar=("PATH"),
             action='append',
             doc="""A dependency for the run. Before running the command, the
-            content of this file will be retrieved. A value of "." means "run
+            content of this file (or directory) will be retrieved. A value of "." means "run
             :command:`datalad get .`". The value can also be a glob. [CMD: This
             option can be given more than once. CMD]"""),
         outputs=Parameter(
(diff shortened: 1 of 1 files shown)


adswa:
I like ``path``, actually

mih:
Same here. Maybe we even use the opportunity to remove the redundancy in the input docstring.

Anyways, good point and definitely worth approaching. Thx

mih:
I cannot provide a committable suggestion in this PR. I will likely open a new one.

datalad/core/local/run.py
```diff
@@ -200,15 +200,15 @@ class Run(Interface):
             metavar=("PATH"),
             action='append',
             doc="""A dependency for the run. Before running the command, the
-            content of this file will be retrieved. A value of "." means "run
+            content of this file (or directory) will be retrieved. A value of "." means "run
```
mih, line 203:
```suggestion
            content for this relative path will be retrieved. A value of "." means "run
```

datalad/core/local/run.py
```diff
@@ -200,15 +200,15 @@ class Run(Interface):
...
             :command:`datalad get .`". The value can also be a glob. [CMD: This
             option can be given more than once. CMD]"""),
         outputs=Parameter(
             args=("-o", "--output"),
             dest="outputs",
             metavar=("PATH"),
             action='append',
-            doc="""Prepare this file to be an output file of the command. A
+            doc="""Prepare this file (or directory) to be an output file of the command. A
```
mih, line 211:
```suggestion
            doc="""Prepare this relative path to be an output file of the command. A
```

mih (commented):
Here is a variant with "path", but also emphasizing the relative nature of the path argument.

mih:
Good to go, thx!

PR #6418, opened 2022-02-04
WIP: Teach push --data 'only' for transferring only annexed contents

adswa:
This change is pretty crude, and probably not suited for a ``core`` command, but I'm eager for advice on what the best approach is. For now, I'm PRing this only to see if it would break anything.

I'm trying to teach ``push`` a new mode ``--data only`` that solely performs transfers of annexed contents and no git pushes whatsoever. I believe there is a need for this as our own tests rely on this functionality, which has so far been fulfilled by the deprecated ``AnnexRepo.copy_to()``. 

In the diff, the change looks more extensive than it is.
Its basically the following three changes:

- I made pushing of the refspecs conditional on ``data != 'none'``. This is most of the change, just an indentation of ~100 lines of code
```
    # perform git push unless only data is to be transferred
    if data != 'only':
       [now push refs as it has always been done]
```
And to make everything work in case ``data='none'`` 
- I moved caching of repo types a bit up:
```
        # cache repo type
        is_annex_repo = isinstance(ds.repo, AnnexRepo)

        # and lastly the primary push target
        target_is_git_remote = repo.config.get(
            'remote.{}.url'.format(target), None) is not None
```
- and initialized ``refspecs2push`` so that ``data='only'`` would not perform a push of any branch. 
```
refspecs2push = None
```
 
Lets see what breaks.

- [TODO - not so optimisitic it would make it yet :P]

### DIFF
--- a/datalad/core/distributed/push.py
+++ b/datalad/core/distributed/push.py
@@ -125,9 +125,11 @@ class Push(Interface):
             or "numcopies" settings for the remote (thus "nothing" otherwise).
             'auto-if-wanted' would enable '--auto' mode only if there is a 
             "wanted" setting for the remote, and transfer 'anything' otherwise.
+            'only' would result in a pure 'git annex copy', transferring nothing
+            but annexed contents.
             """,
             constraints=EnsureChoice(
-                'anything', 'nothing', 'auto', 'auto-if-wanted')),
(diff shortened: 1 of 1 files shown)


adswa:
closing this as I didn't take it any further, and recent PRs went in a similar direction

Issue #6401, opened 2022-02-02
`run --record-reproducible`


mih:
[Sketch for now, will flesh out later]

Undrlying goal: be able to automatically recomoute individual file content on demand, when confirmed reproducible.

Confirming reproducibility implies computing at least twice.

The two computation need not be identical, e.g. an expensive initial computation might yield all pieces to recompute individual file content much cheaper on a 2nd run.

The `run` command seems to have all maschinery to implement this feature, it only needs to be amended with a new last processing step that can record the fact that a command reproduced a declared output, but did NOT modify the dataset. Implementing this in `run` makes this feature easily accessible for `containers-run`, too.

Given that in general no new dataset state will be commited, the resulting additional run-record cannot live in a commit message. We do have sidecar based records as a possible alternative already.

Given that fine grained recomputability for individual files is desirable, one dedicated run-record for each file is not optimal, due to the large number of record file it would generate. We do have the ability to use custom placeholders already. Their expansion may need to be extended to input and output specs too.

When such run records need to be expanded per file, the respective parameter value must be store somewhere. I seems to make sense to annotate the respective annex keys that can be recomputed with this information.

This makes it possible to implement am annex special remote, that can act on this information, and "retrieve" file content via recomputation.

For each key, the following information must be stored:

- reference commit defining the worktree state required for a recomputation
- identifier of the run record template
- list of parameters for expanding all placeholders in the template

yarikoptic:
I am still lacking good understanding of the target use-case you keep in mind (and mentioned in https://github.com/datalad/datalad/issues/2850#issuecomment-1018255469 as well), where somehow some cheap recomputation (and on per-file basis) is possible and so it would be claimed to be a proof of reproducibility.  Could you please make that use case more specific? Otherwise I am still missing why we need "per key" (or per file) record beyond being able to identify commit(s) with a `run` record which produced that content.

mih:
Let's discuss next Tue, ok?

mih:
Everything necessary to do this is now implemented in https://github.com/datalad/datalad/pull/6424 and https://github.com/mih/datalad-mihextras/pull/25

Issue #2850, opened 2018-09-20
datalad-run external special remote


notestaff:
It'd be useful if, when datalad run adds a computed output to git-annex, it also did git annex addurl for that output, where the URL would tell a new external special remote (say, datalad-run) how to recompute that output.  The URL could be e.g. datalad-run:<git hash of the commit recorded by run>, or datalad-run:<git-annex hash of a recipe file sufficient to rerun the computation>.  Then one could use standard 'git annex get' to reconstruct the file, if all its copies have been dropped.  This would work best if the inputs to the run could be explicitly specified, or maybe inferred from the atimes of files that are later than the run start, or even from running under strace.   Then the datalad-run remote could git-annex-get just the needed inputs before a re-run, and check their availability with git-annex-checkpresentkey when answering a CHECKURL request.

yarikoptic:
my 1c: "fancy!" but probably not that trivial (often multiple files produced, might require chaining of multiple runs, so full dependency graph, etc)
So not sure if worthwhile the effort for the majority of scientific applications where outputs are often depend heavily on the software version/environment... although for containerized runs should be quite bit-to-bit reproducible (unless computation is not deterministic, and doesn't compress outputs where header would include some time stamp).
Do you have some realistic use-cases?

notestaff:
chaining of multiple runs would be done by git-annex, when you git-annex-get the input files of the computation and some of them are only "present" in the datalad-run remote (i.e. are recomputable), git-annex would call the remote to get these files.

One use case is that you can use standard git-annex machinery to safely drop all copies of a file as long as it's still recomputable.  So you'd only need to store the initial source files, but all derived files would still be accessible with the standard git-annex-get.  Or could drop heavy raw result files but keep smaller summary/metrics files.

Might work for files produced by something like https://cromwell.readthedocs.io/en/stable/ .  A recipe for computing contents with a given git-annex key is just a git hash for a git tree in which, if you do 'git annex get . && ./run', some file under ./ will have that git-annex key.

yarikoptic:
chaining - indeed...  With our `datalad-archives` some times going deeper and deeper I should have seen it! ;)
Although the entire idea sounds kinda cool or even "sick" (in a good way), I'm still not sure if would be worthwhile beyond having it as a cool exercise, i.e. again -- what real usecase it would address otherwise not addressable via already existing, and somewhat "cleaner" means such as `datalad rerun`?  
[...]

PR #6424, opened 2022-02-07
`run` improvements towards #6401

### LINKED ISSUE #3075: run: be able to say --output={inputs}
Might be RTFM, then please label as a `question`.  I want to be able to avoid specifying the same path(s) whenever inputs are the same as outputs (inplace modifications):
```shell
$> datalad run --explicit -m "Replace leftover datalad_revolution in the API bindings spec" --input datalad/revolution/__init__.py --output '{inputs}' sed -i -e 's,datalad_revolution,datalad.revolution,g' '{inputs}'
[INFO   ] Making sure inputs are available (this may take some time) 
[WARNING] path does not exist: /home/yoh/proj/datalad/datalad/{inputs} [add(/home/yoh/proj/datalad/datalad/{inputs})] 
add(impossible): {inputs} [path does not exist: /home/yoh/proj/datalad/datalad/{inputs}]
```

### LINKED ISSUE #6100: Design document on "run-records"
This is a high-urgency issue. Alternative `run` implementations are likely, and we need standardization, and a comprehensive description.

mih:
Work towards

- #6401
- #2850

This adds a bunch of features to the internal `run`/`rerun` workhorse `run_command()` that are necessary to be able to implement a solution to #6401 efficiently, without reaching further into the guts of the machinery.

I will keep this open, until such a solution is fully implemented, and testable -- however, I am fairly confident at this point that this is pretty much the entire change. I am hoping to be able to add a design document on run records and `run`'s results at the very end. Reviews welcome now!

TODO:

- [x] rather than including just `declared_outputs` in the `run` result, I feel like we should also yield `declared_inputs`. With `remove_outpus=True`, a caller may need to know which files where actually generated (not inputs/outputs intersection). Yielding them would be negligible cost, because expansion was already done.
- [x] Design document on prov-capture

- `run` now supports placeholders in input and output specifications. This includes all standard and custom placeholders, and enables use cases like `--inputs '{outputs}'` to conveniently declare all outputs to also be inputs for a command execution. Fixes #3075 
#### 📝 Documentation
- A design document with basic information on the provenance capture approach was added. Fixes #6100
#### 🏠 Internal
- The `run` and `rerun` workhorse function `run_command()` has gained additional flags. `parametric_record` prevents the expansion of any placeholder in a provenance record. This enables creating a single record that be parameterized to implement more than one (re)computation. `skip_dirtycheck` prevents any test for dataset modifications prior a command execution, regardless of other mode specifications. This aids callers that already need to perform analog checks before executing `run_command()`. The new option `yield_expanded` can be used to add lists of expanded inputs and outputs to the result record to avoid re-processing this information in callers. Lastly, `remove_outputs` alters the default, content availability dependent, output preparation. If enabled, all declared outputs that are not also declared inputs will be deleted prior a command execution. Files that are both input and output will be unlocked. This helps to avoid #6441

### DIFF
--- a/datalad/core/local/run.py
+++ b/datalad/core/local/run.py
@@ -710,7 +710,8 @@ def run_command(cmd, dataset=None, inputs=None, outputs=None, expand=None,
                 rerun_outputs=None,
                 inject=False,
                 parametric_record=False,
-                remove_outputs=False):
+                remove_outputs=False,
+                skip_dirtycheck=False):
     """Run `cmd` in `dataset` and record the results.
 
     `Run.__call__` is a simple wrapper over this function. Aside from backward
@@ -747,6 +748,11 @@ def run_command(cmd, dataset=None, inputs=None, outputs=None, expand=None,
     remove_outputs : bool, optional
         If enabled, all declared outputs will be removed prior command
         execution, except for paths that are also declared inputs.
+    skip_dirtycheck : bool, optional
+        If enabled, a check for dataset modifications is unconditionally
+        disabled, even if other parameters would indicate otherwise. This
+        can be used by callers that already performed analog verififcations
+        to avoid duplicate processing.
 
     Yields
     ------
@@ -777,7 +783,8 @@ def run_command(cmd, dataset=None, inputs=None, outputs=None, expand=None,
 
     lgr.debug('tracking command output underneath %s', ds)
 
-    if not (rerun_info or inject):
+    # skip for callers that already take care of this
+    if not (skip_dirtycheck or rerun_info or inject):
         # For explicit=True, we probably want to check whether any inputs have
         # modifications. However, we can't just do is_dirty(..., path=inputs)
         # because we need to consider subdatasets and untracked files.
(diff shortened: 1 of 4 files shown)


mih:
I have added this PR to the 0.16 milestone. I do not (or no longer) plan to actuall implement #6401 within `run` or within this PR. But it would be nice to get `run` improved enough to make this possible via a separate implementation.

datalad/core/local/run.py
```diff
@@ -11,66 +11,64 @@
...
-from argparse import REMAINDER
+import logging
 import os
 import os.path as op
-from os.path import join as opj
-from os.path import normpath
-from os.path import relpath
+from pathlib import Path
+import warnings
+from argparse import REMAINDER
```
yarikoptic, line 20: run isort?

mih: Done

mih:
I think this can be merged now. I will do so once the tests have passed after my conflict resolution push

mih:
Nope, this is now blocked by https://github.com/datalad/datalad/issues/6482

PR #6428, opened 2022-02-07
MNT: remove deprecated mktemp from add-archive-content

adswa:
This function bears a potential security risk and has been deprecated
since Python version 2.3. This change replaces its use with mkdtemp,
which creates a temporary directory in a more secure manner

As mkdtemp does not clean the temporary directory up afterwards, this
change also adds a removal of any leftover empty temporary directories.

#### 🏠 Internal
- ``add-archive-content`` does not rely on the deprecated ``tempfile.mktemp`` anymore, but uses the more secure ``tempfile.mkdtemp``


mih (approved):
I somehow feel the urge to put this in a try-finally, but no such protection existed before, so this should still be a net improvement. Thx!

Issue #6412, opened 2022-02-03
Travis tripping over AssertionError: Test was too slow


mih:
This has happened a few times over the past days. Here is one example:

https://app.travis-ci.com/github/datalad/datalad/jobs/558196552
```
======================================================================
4717ERROR: datalad.distribution.tests.test_get.test_source_candidate_subdataset
4718----------------------------------------------------------------------
4719Traceback (most recent call last):
4720  File "/tmp/dl-miniconda-s23vc9zy/lib/python3.9/site-packages/nose/case.py", line 134, in run
4721    self.runTest(result)
4722  File "/tmp/dl-miniconda-s23vc9zy/lib/python3.9/site-packages/nose/case.py", line 152, in runTest
4723    test(result)
4724AssertionError: Test was too slow (took 31.3677s, threshold was 30.0000s)
```

yarikoptic:
Was it happening in the same PR? Then likely it is the one which introduced slow down. I think I never saw this test to exceed its time before

yarikoptic:
FWIW

<details>
<summary>in the past year it seemed always was snappy enough to not hit 30sec mark and had about 5 sec in reserve even in the worst cases</summary> 

```shell
(git)smaug:/mnt/datasets/datalad/ci/logs/2021/12[master]git
$> git grep 'test_source_candidate_subdataset:' | grep travis | sed -e 's,.*: ,,g'| sort -n | nl | tail
   355	23.4473s
   356	23.4852s
   357	23.6095s
   358	23.6279s
   359	23.7064s
   360	23.9314s
   361	24.3871s
   362	24.8944s
   363	25.0311s
   364	25.4870s
1 2796.....................................:Thu 03 Feb 2022 02:47:45 PM EST:.
(git)smaug:/mnt/datasets/datalad/ci/logs/2021/12[master]git
$> cd ../11 
02/  03/  04/  05/  07/  08/  09/  10/  11/  12/  13/  14/  15/  16/  17/  18/  19/  20/  21/  22/  23/  24/  25/  26/  28/  29/  30/
1 2797.....................................:Thu 03 Feb 2022 02:47:53 PM EST:.
(git)smaug:/mnt/datasets/datalad/ci/logs/2021/11[master]git
$> git grep 'test_source_candidate_subdataset:' | grep travis | sed -e 's,.*: ,,g'| sort -n | nl | tail
   199	21.0065s
   200	21.2269s
   201	21.2856s
   202	21.3108s
   203	21.3109s
[...]

PR #6429, opened 2022-02-07
TST: mark test_source_candidate_subdataset as @slow

yarikoptic:
See https://github.com/datalad/datalad/issues/6412 for more background.
The actual reason was not discovered, but the test became slower in late Jan,
crossing the line of 30seconds.  We are marking it as slow to mitigate
false positive CI fails on PRs.


mih:
Thx!

PR #6433, opened 2022-02-08
Simplify pattern 'EnsureChoice(..) | EnsureNone'

mih:
By including None in the choice list. This makes for a better
documentation and enables auto-completion, provided by #6415

This accompanies https://github.com/datalad/datalad/pull/6415#issuecomment-1032559152 but it independent of it.

### Changelog
This does not need to show up in the changelog.


yarikoptic:
failed travis run is https://github.com/datalad/datalad/issues/4496

mih:
Thanks @yarikoptic !

PR #6435, opened 2022-02-09
Use `self.repo` in `RegisterUrl`-constructor

christian-monch:
The constructor of `datalad.local.addurls.RegisterUrl` uses the parameter `repo` wich defaults to `None` during comparison. It should instead use the attribute `self.repo`, which is initialized to `ds.repo` if `repo` is `None`.

This PR uses `self.repo` in the constructor and adds a regression test.

This was not reported in an issue, but is definitively a bug in my opinion.

### DIFF
--- a/datalad/local/addurls.py
+++ b/datalad/local/addurls.py
@@ -875,9 +875,9 @@ class RegisterUrl(object):
         self.repo = repo or ds.repo
         self._err_res = get_status_dict(action="addurls", ds=self.ds,
                                         type="file", status="error")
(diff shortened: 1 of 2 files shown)


yarikoptic (approved):
I can only wonder why `, repo=None` if used only with the one provided and in general this interface AFAIK is not intended for "outside use".  But as for motivation and the fix -- looks LGTM, so good with me.

Windows run on appveyor seemed problematic for some other reasons -- had errors and then stalled I guess exceeding its hour. will restart just for the sake of it

PR #6436, opened 2022-02-09
OPT+RF: create-sibling --since=^ -- diff to HEAD, use "^" instead of ""

### LINKED ISSUE #6399: create-sibling -r --since   is likely investigating too much/long
Switching from `publish` (which had option to create sibling if missing) to `create-sibling` to precede `push` for a set of changes where some number of new subsubdatasets were created:
 

<details>
<summary>`create-sibling -r --since` took over 8 hours to complete</summary> 

```shell
(git)smaug:/mnt/datasets/datalad/crawl[master]git
$> time datalad create-sibling -s datalad-public -d . -r --inherit --existing=skip --since datalad-public/master
[INFO   ] Connecting ...
[INFO   ] Considering to create a target dataset /mnt/datasets/datalad/crawl/openneuro/ds000232 at /srv/datasets.datalad.org/www/openneuro/ds000232 of datasets.datalad.org
subdataset(ok): ds000232 (dataset)
[INFO   ] Fetching updates for Dataset(/mnt/datasets/datalad/crawl/openneuro/ds000232)
subdataset(ok): ds000232 (dataset)
[INFO   ] Inherited publish_depends from Dataset(/mnt/datasets/datalad/crawl/openneuro): refs/heads/*
...
create_sibling(ok): /mnt/datasets/datalad/crawl/openneuro/ds001371 (dataset)
create_sibling(ok): /mnt/datasets/datalad/crawl/openneuro/ds000232 (dataset)
action summary:
  create_sibling (notneeded: 135, ok: 42)
datalad create-sibling -s datalad-public -d . -r --inherit --existing=skip    1670.84s user 26386.11s system 90% cpu 8:35:55.34 total
```
</details>

I started `time datalad diff -r --from=datalad-public/master -R 2` -- will let know what timing would be , but likely it is the culprit

datalad 0.15.4

yarikoptic:
Two commits:

    OPT: create_sibling - diff to HEAD not None (tree)
    
    Primary motivation is very slow --since= mode of operation of create-sibling
    on a heavy superdataset (see #6399).  Although it could be argued that there might be
    a benefit from create-sibling to be able to create siblings for not yet committed
    submodules, I think such use cases are sparse if any and thus diff"ing to the committed
    state (HEAD) is actually the desired effect.
    
    Fixes #6399 (although I did not check explicitly)

-----

    RF: create-sibling --since to take "^" instead of ""
    
    to be consistent with "push" and thus reduce cognitive load.  Old value of an
    empty string will remain supported.  While at it, added a tiny bit of testing
    to ensure that --since actually works out the way we expect and only deals with
    that subdataset (or notneeded)

Moreover, apparently `--since=` was not even documented before, this fixes it.

First commit I would not even mind to destine against `maint` but decided not to bother since 0.16.0 is imminent, right?

- `create-sibling --since=^` mode will now be as fast as `push --since=^` to figure out for which subdatasets to create siblings.
- `create-sibling` will require now "^" instead of an empty string for `since` option. 
#### 📝 Documentation
- `--since=^` mode of operation of `create-sibling` is documented now

### DIFF
--- a/datalad/distribution/create_sibling.py
+++ b/datalad/distribution/create_sibling.py
@@ -673,7 +673,7 @@ class CreateSibling(Interface):
                 for r in diff_dataset(
                     ds,
                     fr=since,
-                    to=None,
+                    to='HEAD',
                     # make explicit, but doesn't matter, no recursion in diff()
                     constant_refs=True,
                     # contrain to the paths of all locally existing subdatasets
--- a/datalad/distribution/create_sibling.py
+++ b/datalad/distribution/create_sibling.py
@@ -551,7 +551,9 @@ class CreateSibling(Interface):
             constraints=EnsureStr() | EnsureNone(),
             doc="""limit processing to subdatasets that have been changed since
             a given state (by tag, branch, commit, etc). This can be used to
-            create siblings for recently added subdatasets."""),
+            create siblings for recently added subdatasets.
+            If '^' is given, the last state of the current branch at the sibling
+            is taken as a starting point."""),
     )
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


yarikoptic:
It is magically green! Even nearly all benchmarks under 1 - can't be bad ;-)

mih (approved):
Let's not let this unicorn event slip!

yarikoptic:
note FTR: damn, burnt myself -- thought that it was against `maint`, spent time tracking down in "maint". uff.

PR #6439, opened 2022-02-10
Demote BatchedCommand remaining content-warning to debug message

### LINKED ISSUE #6431: Runner emits warnings on batched annex commands
I am using `git annex registerurl` in batch mode. Apparently there's one more line of output than expected (namely some `registerurl (ok)`  at the end). This results in the runner emitting a WARNING about remaining, unprocessed output.
Either that means, that batch commands need to change how they call the runner to make it clear in some way that this is expected. But I think of that as a maintenance burden as it requires to specifically find out about every annex command how exactly it behaves in that regard in batch mode (and in what annex version).
Or it should not be a WARNING to begin with. I lean towards that, since a user can't do anything about it anyway. What do you think, @christian-monch ?

christian-monch:
This PR changes a warning message to a debug message. The message indicates that there is still unprocessed command-output in a BatchedCommand object, when the BatchedCommand object is closed.

This PR fixes #6431 

#### 🏠 Internal
- Remove a warning message that does not necessarily indicate an error, but might confuse users.

### DIFF
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -119,7 +119,7 @@ class BatchedCommandProtocol(GeneratorMixIn, StdOutErrCapture):
         if fd == STDOUT_FILENO:
             remaining_line = self.line_splitter.finish_processing()
             if remaining_line is not None:
-                lgr.warning(f"unterminated line: {remaining_line}")
+                lgr.debug(f"unterminated line: {remaining_line}")
                 self.send_result((fd, remaining_line))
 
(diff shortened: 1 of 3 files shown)


mih:
Thx @christian-monch. Can you please add a changelog item, or a statement why you think there should be none?

christian-monch:
> Thx @christian-monch. Can you please add a changelog item, or a statement why you think there should be none?

Done

bpoldrack (approved):
Thx!

Issue #6445, opened 2022-02-11
Something introduced widespread double reporting of results


adswa:
I noticed in #6258 that some commands result rendering contains duplicate entries. For example, this is the output of ``clone`` to clone a subdataset on current ``master``:

```
(handbook2) adina@muninn in /tmp/ab on git:master
❱ datalad clone -d . https://gin.g-node.org/adswa/bids-data input15             
install(ok): input15 (dataset)                                                  
add(ok): input15 (file)                                                         
add(ok): input15 (file)                                                         
add(ok): .gitmodules (file)                                                     
add(ok): .gitmodules (file)                                                     
save(ok): . (dataset)                                                           
save(ok): . (dataset)                                                           
add(ok): .gitmodules (file)                                                     
save(ok): . (dataset)                                                           
add(ok): .gitmodules (file)                                                     
save(ok): . (dataset)
```

This specific flood of results seems to be due to ``save`` -  a previous not set ``result_renderer='disabled'`` became necessary. 
#6421, for example, already added one to fix up the duplicate result rendering of ``run`` a few days ago. While its not confusing that ``result_renderer='disabled'`` would need to be set, its confusing to me why it only now showed the double-reporting and this was not observed before.

I tried to bisect this change, but due to the changes in CLI this proved to be quite difficult - I couldn't evaluate good or bad bisection states for some commits because at one point, result rendering was completely turned off, or because in one state result rendering led to internal errors. The outcome of this bisection is thus not really trustworthy in my opinion:

```
(handbook2) adina@muninn in ~/repos/datalad on bisect/bad~1 (bisect)
❱ git bisect bad 
83c9270305778884b6a480ef7c0875811d44948d is the first bad commit
commit 83c9270305778884b6a480ef7c0875811d44948d
Author: Michael Hanke <michael.hanke@gmail.com>
Date:   Sat Jan 29 13:24:36 2022 +0100

    Deprecate datalad.interface.common_opts.eval_defaults
    
    The default are more productively put into the Interface class directly.
    The makes the documentation more accessible, makes more obvious where
    overrides need to go, reduces the number of places that need to be
    queried in order to determine the effective value.

 datalad/cli/common_args.py       | 7 ++-----
 datalad/cli/tests/test_exec.py   | 2 +-
 datalad/interface/base.py        | 5 +++--
 datalad/interface/common_opts.py | 6 ++++++
 datalad/interface/utils.py       | 7 +------
[...]

Issue #6258, opened 2021-11-26
Certain commands don't have an action summary anymore


adswa:
I'm reviewing the diff between the handbook build before the last 0.15 release and the current master. I'm not sure whether its a problem, so please close if its intended, but I noticed that some commands don't have an `action summary` anymore, and some commands have gained one. Here is what I am seeing:

``run`` and ``subdatasets`` and ``uninstall`` lost it:

```diff
diff --git a/docs/basics/_examples/DL-101-108-105 b/docs/basics/_examples/DL-101-108-105
index 3cf5f53c..6166703c 100644
--- a/docs/basics/_examples/DL-101-108-105
+++ b/docs/basics/_examples/DL-101-108-105
@@ -4,6 +4,3 @@ $ datalad run -m "create a list of podcast titles" \
 [INFO] == Command exit (modification check follows) ===== 
 add(ok): recordings/podcasts.tsv (file)
 save(ok): . (dataset)
-action summary:
-  add (ok: 1)
-  save (notneeded: 1, ok: 1)

diff --git a/docs/basics/_examples/DL-101-110-103 b/docs/basics/_examples/DL-101-110-103
index 94d53b0d..aaabcceb 100644
--- a/docs/basics/_examples/DL-101-110-103
+++ b/docs/basics/_examples/DL-101-110-103
@@ -11,7 +11,3 @@ get(ok): recordings/longnow/.datalad/feed_metadata/logo_salt.jpg (file) [from we
 [INFO] == Command exit (modification check follows) ===== 
 add(ok): recordings/salt_logo_small.jpg (file)
 save(ok): . (dataset)
-action summary:
-  add (ok: 1)
-  get (notneeded: 1, ok: 1)
-  save (notneeded: 1, ok: 1)
diff --git a/docs/basics/_examples/DL-101-111-105 b/docs/basics/_examples/DL-101-111-105
index 2b15528e..9a734045 100644
--- a/docs/basics/_examples/DL-101-111-105
+++ b/docs/basics/_examples/DL-101-111-105
@@ -14,7 +14,3 @@ get(ok): recordings/longnow/.datalad/feed_metadata/logo_interval.jpg (file) [fro
 [INFO] == Command exit (modification check follows) ===== 
 add(ok): recordings/interval_logo_small.jpg (file)
 save(ok): . (dataset)
-action summary:
-  add (ok: 1)
-  get (notneeded: 1, ok: 1)
-  save (notneeded: 1, ok: 1)
[...]

PR #6446, opened 2022-02-11
BF/RF: Silence save result rendering in clone and download url

adswa:
A recent change has introduced double reporting for a number of commands (#6445). I haven't figured out why this change has suddenly happened, but this PR silences (``result_renderer='disabled'``) internal calls to ``save`` and ``get`` that should be silenced anyway.

In addition, it contains a small RF that centralizes general result renderer switches outside of a loop. Coincidentally, this RF reinstates action summaries for a few commands that had lost it (#6258)

#### 🐛 Bug Fixes
- A number of commands stopped to double-report results (by @adswa)

### DIFF
--- a/datalad/core/distributed/clone.py
+++ b/datalad/core/distributed/clone.py
@@ -367,6 +367,7 @@ class Clone(Interface):
                     return_type='generator',
                     result_filter=None,
                     result_xfm=None,
+                    result_renderer='disabled',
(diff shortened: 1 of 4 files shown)


mih (approved):
Thanks for digging through this, and the further logic simplification.

IMHO, all internal command calls that lack `result_renderer='disabled'` are a bug.

PR #6447, opened 2022-02-11
ENH: run: Provide result record for command execution

### LINKED ISSUE #3071: Option to force run (or containers-run) to save even if {cmd} exits with non-0
I am scripting use of `containers-run`. I know that the command I am to run might fail, and I do want the `run` to fail if it does. But I do want to save outputs (log file from bids-validator) regardless of the output status. ATM I do not see explicit option for that.  I am only instructed in the output to
```
[INFO   ] The command had a non-zero exit code. If this is expected, you can save the changes with 'datalad save -r -F .git/COMMIT_EDITMSG .'
```
But that is cumbersome for scripting -- `run` could have exited with non-0 for some other reason (wrong container specified etc), and I would need to parse log to figure out either it was the actual command of interest which has failed.
So it would be nice to have some `run` specific option, such as `--always-save` or alike which would save even when command fails.

mih:
Note that this contains a breaking change:

  * it changes a CommandError to a result record with status="error"

As pointed out in https://github.com/datalad/datalad/issues/3071#issuecomment-1034648920 we could continue to raise a CommandError for Python-API users. But @mih is not convinced that this is worth it.

This is the last piece from the closed https://github.com/datalad/datalad/pull/5695

Demo:

```
% datalad -f json_pp run "echo 12"
12
{
  "action": "run",
  "exitcode": 0,
  "explicit_outputs": null,
  "message": "Executed command",
  "msg_path": null,
  "path": "/tmp/runny",
  "record_id": "49f818a6e4ab1d9c70610db52da25daf",
  "run_info": {
    "chain": [],
    "cmd": "echo 12",
    "dsid": "2f0b7d17-f5d2-4659-8da8-8bd9dd781e14",
    "exit": 0,
    "extra_inputs": [],
    "inputs": [],
    "outputs": [],
    "pwd": "."
  },
  "status": "ok",
  "type": "dataset"
}
{                                                                                                                           
  "action": "save",
  ...
```

TODO:

- [x] Add sidecar info to result record
- [x] Avoid duplicate exit_code record in the error case (use `exit_code`)

- `run` now yields a result record immediately after executing a command.  This allows callers to use the standard `--on-failure
ignore` switch to control whether dataset modifications will be saved for a command that exited with an error. Fixes #3071 (edited postfact by @yarikoptic to fix HI slop)
- `run` no longer raises a `CommandError` exception for failed commands, but yields an `error` result that includes a superset of the information provided by the exception. This change impacts command line usage insofar as the exit code of the underlying command is no longer relayed as the exit code of the `run` command call -- although `run` continues to exit with a non-zero exit code in case of an error. For Python API users, the nature of the raised exception changes from `CommandError` to `IncompleteResultsError`, and the exception handling is now configurable using the standard `on_failure` command argument. The original `CommandError` exception remains available via the `exception` property of the newly introduced result record for the command execution, and this result record is available via `IncompleteResultsError.failed`, if such an exception is raised.

### DIFF
--- a/datalad/core/local/run.py
+++ b/datalad/core/local/run.py
@@ -15,6 +15,8 @@ import logging
 import json
 import warnings
 
+from pathlib import Path
+
 from argparse import REMAINDER
 import os
 import os.path as op
@@ -322,8 +324,8 @@ class Run(Interface):
                res.get("action") == "run" and res.get("status") == "error":
                 msg_path = res.get("msg_path")
                 if msg_path:
+                    ds_path = res["path"]
                     if datalad.get_apimode() == 'python':
-                        ds_path = res["path"]
                         help = f"\"Dataset('{ds_path}').save(path='.', " \
                                "recursive=True, message_file='%s')\""
                     else:
@@ -334,7 +336,7 @@ class Run(Interface):
                         f"{help}",
                         # shorten to the relative path for a more concise
                         # message
-                        msg_path.relative_to(ds.pathobj))
+                        Path(msg_path).relative_to(ds_path))
             generic_result_renderer(res)
 
 
@@ -832,9 +834,11 @@ def run_command(cmd, dataset=None, inputs=None, outputs=None, expand=None,
     if not rerun_info and cmd_exitcode:
         if do_save:
             repo = ds.repo
-            msg_path = relpath(opj(str(repo.dot_git), "COMMIT_EDITMSG"))
-            with open(msg_path, "wb") as ofh:
-                ofh.write(ensure_bytes(msg))
+            # must record path to be relative to ds.path to meet
+            # result record semantics (think symlink resolution, etc)
+            msg_path = ds.pathobj / \
+                repo.dot_git.relative_to(repo.pathobj) / "COMMIT_EDITMSG"
+            msg_path.write_text(msg)
 
     expected_exit = rerun_info.get("exit", 0) if rerun_info else None
     if cmd_exitcode and expected_exit != cmd_exitcode:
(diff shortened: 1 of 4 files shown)


bpoldrack:
> As pointed out in #3071 (comment) we could continue to raise a CommandError for Python-API users. But @mih is not convience that this is worth it.

I agree. Especially since a python caller should somehow deal with possible error results anyway.

datalad/core/local/run.py
```diff
@@ -823,30 +828,34 @@ def run_command(cmd, dataset=None, inputs=None, outputs=None, expand=None,
...
-            if repo.dirty and not explicit:
-                # Give clean-up hints if a formerly clean repo is left dirty
-                lgr.info(
-                    "The commands 'git clean -di' and/or 'git reset' "
-                    "could be used to get to a clean dataset state again. "
-                    "Consult their man pages for more information."
-                )
-        raise exc
-    elif do_save:
+            msg_path = relpath(opj(str(repo.dot_git), "COMMIT_EDITMSG"))
```
bpoldrack, line 835: Why turn the `Path` in `repo.dot_git` into a str to then use it with `os.path` rather than just using `pathlib`?

mih: This is the original change that Kyle made. Any suggested changes are most welcome!

mih: It actually seems to be redundant with the post-processing done in the result renderer. So I guess I resolved the merge conflicts in a suboptimal fashion. Will correct.

mih: I have addressed this now. It still needs to be recoded from Repo to Dataset semantics, though.

mih:
I added another changelog item with more details on the nature of the breaking change.

PR #6458, opened 2022-02-17
Silence result rendering in run's use of drop()

adswa:
This prevents double reporting of unlock.
Prior to this change, a run that unlocks a file looks like this:
```
datalad run --input api.py --output some 'cat api.py >> some'
[INFO   ] Making sure inputs are available (this may take some time)
unlock(ok): some (file)
unlock(ok): some (file)
run(ok): /tmp/mamam (dataset) [Executed command]
add(ok): some (file)
save(ok): . (dataset)
```
With this change:
```
 datalad run --input api.py --output some 'cat api.py >> some'
[INFO   ] Making sure inputs are available (this may take some time)
unlock(ok): some (file)
run(ok): /tmp/mamam (dataset) [Executed command]
add(ok): some (file)
save(ok): . (dataset)
```

I also set the return_type of ``drop()`` to ``generator`` because I think it makes sense there - let me know if that's wrong.

The changes should not conflict with #6424

### Changelog
not needed

### DIFF
--- a/datalad/core/local/run.py
+++ b/datalad/core/local/run.py
@@ -519,8 +519,11 @@ def _unlock_or_remove(dset_path, paths):
         # Note: If Unlock() is given a directory (including a subdataset) as a
         # path, files without content present won't be reported, so those cases
         # aren't being covered by the "remove if not present" logic below.
-        for res in Unlock()(dataset=dset_path, path=existing,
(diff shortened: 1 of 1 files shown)


mih (approved):
Thx!

Note to self: ignore is ok here, because the results are always yielded or handled otherwise.

Issue #6461, opened 2022-02-18
Systematic review of result handling in internal command calls.


mih:
Following #6445, we decided to review the code base for its result handling practices. In particular the `return_type` and `result_renderer` switches, which should be `generator` and `disabled` respectively, unless there is an actual reason for a deviation from the pattern. Likewise, `on_failure='ignore'` in most cases.

Please extend this list with the respective PRs or commands that were reviewed or were good already. Please tick the box, when a command is considered done completely (it may need more than one PR):

- [x] add-archive-content
   - (no adjustments necessary)
- [x] add-readme
  - #6469 
- [x] addurls
  - #6465
- [x] aggregate-metadata
  - #6475 
- [x] check-dates
   - (no adjustments necessary) 
- [x] clean
   - (no adjustments necessary)
- [x] clone
  - #6446
  - #6466
- [x] configuration
  - #6468
- [x] copy-file
   - #6470 
- [x] `create`
  - #6457
- [x] create-sibling
  - #6471
- [x] create-sibling-gin
  - #6471
- [x] create-sibling-gitea
  - #6471
- [x] create-sibling-github
  - #6471
- [x] create-sibling-gitlab
  - #6471
- [x] create-sibling-gogs
  - #6471
- [x] create-sibling-ria
  - #6471
- [x] create-test-dataset
   - (no adjustments necessary)
- [x] diff
   - (no adjustments necessary)
- [x] download-url
  - #6446
- [x] drop
   - (no adjustments necessary)
- [x] export-archive
- [x] export-archive-ora
- [x] export-to-figshare
- [x] extract-metadata
- [x] foreach-dataset
  - #6475 
- [x] get
   - #6476 
- [x] install
   - #6472 
- [x] metadata
  - #6473 
- [x] no-annex
  - #6475 
- [x] push
- [x] remove
  - #6475 
- [x] rerun
- [x] `run`
  - #6458
  - #6463 
- [x] `save`
  - #6462
- [x] search
 - #6489 
- [x] shell-completion
- [x] siblings
- [x] sshrun
- [x] status
- [x] subdatasets
  - #6467
- [x] test
  - (no adjustment needed)
- [x] unlock
- [x] update
  - #6464 
- [x] wtf

Note on executing this review:

- Case 1: A command does not make any internal calls to other datalad command -- just tick the box, done
- Case 2: another datalad command is called inside a loop, but return_type='generator' is missing -- add it
[...]

PR #6466, opened 2022-02-18
MNT: use generator return_type in internal save call in clone

adswa:
Update after #6446, which did not set the return type explicitly as outlined in #6461.
This change specifies the return type of a save call to save .gitmodules to be a generator.

no changelog needed


mih (approved):
LGTM, thx!

christian-monch (commented):
Hi Adina, could it be that `ds.subdataset` in line 405 should also be changed?

adswa:
> Hi Adina, could it be that `ds.subdataset` in line 405 should also be changed?

@mih needs to correct me if I'm wrong, but I don't think so because no result is yielded

PR #6467, opened 2022-02-18
Update `subdatasets()` in the spirit of #6461

mih:
No changelog needed.


yarikoptic:
FWIW, looks legit, thank you @mih

PR #6469, opened 2022-02-18
MNT: Silence internal command calls' result rendering in add-readme

adswa:
This updates `add-readme`  according to #6461. Previously, 'datalad add-readme' resulted in double reporting that looked
like this:
```
datalad add-readme
[WARNING] Found no aggregated metadata info file /tmp/abcd/more/.datalad/metadata/aggregate_v1.json. You will likely need to either update the dataset from its original location or reaggregate metadata locally.
metadata(impossible): . (dataset) [Dataset at . contains no aggregated metadata on this path]
[WARNING] Could not obtain dataset metadata, proceeding without
add_readme(ok): /tmp/abcd/more/README.md (file)
add(ok): README.md (file)
save(ok): . (dataset)
action summary:
  add (ok: 1)
  save (ok: 1)
add(ok): README.md (file)
save(ok): . (dataset)
action summary:
  add (ok: 1)
  add_readme (ok: 1)
  save (ok: 1)
```
With this change applied, it looks like this:
```
datalad add-readme
[WARNING] Found no aggregated metadata info file /tmp/abcd/more/more/more/.datalad/metadata/aggregate_v1.json. You will likely need to either update the dataset from its original location or reaggregate metadata locally.
[WARNING] Could not obtain dataset metadata, proceeding without
add_readme(ok): /tmp/abcd/more/more/more/README.md (file)
add(ok): README.md (file)
save(ok): . (dataset)
action summary:
  add (ok: 1)
  add_readme (ok: 1)
  save (ok: 1)
```

### DIFF
--- a/datalad/local/add_readme.py
+++ b/datalad/local/add_readme.py
@@ -98,15 +98,20 @@ class AddReadme(Interface):
                 # configure the README to go into Git
                 dataset.repo.set_gitattributes(
                     [(filename, {'annex.largefiles': 'nothing'})])
-                dataset.save(
+                yield from dataset.save(
                     path='.gitattributes',
                     message="[DATALAD] Configure README to be in Git",
-                    to_git=True
+                    to_git=True,
+                    return_type='generator',
+                    result_renderer='disabled'
                 )
 
         # get any metadata on the dataset itself
         dsinfo = dataset.metadata(
-            '.', reporton='datasets', return_type='item-or-list',
+            '.',
+            reporton='datasets',
+            return_type='item-or-list',
+            result_renderer='disabled',
             on_failure='ignore')
         meta = {}
         if not isinstance(dsinfo, dict) or dsinfo.get('status', None) != 'ok':
@@ -232,9 +237,11 @@ files by whom, and when.
                 type='file',
                 action='add_readme')
 
-        for r in dataset.save(
+        yield from dataset.save(
                 fpath,
                 message='[DATALAD] added README',
                 result_filter=None,
-                result_xfm=None):
-            yield r
+                result_xfm=None,
+                return_type='generator',
+                result_renderer='disabled'
+        )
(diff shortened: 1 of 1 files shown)


christian-monch (commented):
There might be an unlock in line 91 the should be changed as well

datalad/local/add_readme.py
```diff
@@ -101,12 +101,17 @@ def __call__(filename='README.md',
                 dataset.save(
                     path='.gitattributes',
                     message="[DATALAD] Configure README to be in Git",
-                    to_git=True
+                    to_git=True,
+                    return_type='generator',
+                    result_renderer='disabled'
```
mih, line 106: needs to be wrapped into `yield from` or would stop doing anything

datalad/local/add_readme.py
```diff
@@ -236,5 +241,8 @@ def __call__(filename='README.md',
                 fpath,
                 message='[DATALAD] added README',
                 result_filter=None,
-                result_xfm=None):
+                result_xfm=None,
+                return_type='generator',
+                result_renderer='disabled'
+        ):
```
mih, line 247: The whole call could be reduced to `yield from` instead of the `for` loop

mih (changes requested):
I made a few remarks, one points to a bug. Thx!

datalad/local/add_readme.py
adswa: oh thanks for catching this! it is a bit worrisome that the tests did not pick this up...

datalad/local/add_readme.py
adswa: thanks, done!

mih (changes requested):
LGTM. But as @christian-monch mentioned, there is an unlock call on line 91. It should likely yield is results with rendering disabled.

adswa:
Oh sorry, I did not see this comment!

yarikoptic:
I didn't spot anything else, looks LGTM, let's proceed

PR #6471, opened 2022-02-18
Update `create_sibling*()` in the spirit of #6461

mih:
No changelog needed.


mih:
Nobody complained....

PR #6476, opened 2022-02-18
Unify keyword in the spirit of issue #6461

christian-monch:
Add a missing result_renderer-spec in call to subdatasets

#### 🏠 Internal
- disable result renderer in internal call  to subdataset in get


mih (approved):
LGTM, thx!

Is that all there was for `get`?

christian-monch:
> Is that all there was for `get`?

I am taking another look at it, there are subdataset calls that can be converted to generator based calls.

PR #6484, opened 2022-02-22
Replace custom entrypoint with standard for special remotes

mih:
New pattern introduced with ecd4577753d9e82cdb097d35368eadcf79e24e21

- `git-annex-remote-ora` now provides the same standardized CLI interface as other special remotes and the core package with support for help and version reporting.


yarikoptic:
CI is happy, I am happy with such a change

Issue #6482, opened 2022-02-22
`datalad.local.tests.test_foreach_dataset.test_basic_resilience` not resilient


mih:
From https://ci.appveyor.com/project/mih/datalad/builds/42653693/job/bj1sa5ow0st1ketd testing https://github.com/datalad/datalad/pull/6424
```
======================================================================
4066ERROR: datalad.local.tests.test_foreach_dataset.test_basic_resilience(<function get_deeply_nested_structure at 0x0000021B32B80040>,)
4067----------------------------------------------------------------------
4068Traceback (most recent call last):
4069  File "c:\projects\datalad\datalad\support\parallel.py", line 368, in _iter_threads
4070    raise _FinalShutdown()
4071datalad.support.parallel._FinalShutdown
4072
4073During handling of the above exception, another exception occurred:
4074
4075Traceback (most recent call last):
4076  File "C:\Python39-x64\lib\site-packages\nose\case.py", line 198, in runTest
4077    self.test(*self.arg)
4078  File "c:\projects\datalad\datalad\tests\utils.py", line 874, in _wrap_with_tempfile
4079    return t(*(arg + (filename,)), **kw)
4080  File "c:\projects\datalad\datalad\local\tests\test_foreach_dataset.py", line 50, in check_basic_resilience
4081    res_python = ds.foreach_dataset("ds.path", cmd_type='eval', **kwargs)
4082  File "c:\projects\datalad\datalad\distribution\dataset.py", line 502, in apply_func
4083    return f(*args, **kwargs)
4084  File "c:\projects\datalad\datalad\interface\utils.py", line 447, in eval_func
4085    return return_func(*args, **kwargs)
4086  File "c:\projects\datalad\datalad\interface\utils.py", line 439, in return_func
4087    results = list(results)
4088  File "c:\projects\datalad\datalad\interface\utils.py", line 357, in generator_func
4089    for r in _process_results(
4090  File "c:\projects\datalad\datalad\interface\utils.py", line 544, in _process_results
4091    for res in results:
4092  File "c:\projects\datalad\datalad\local\foreach_dataset.py", line 387, in __call__
4093    yield from pc_class(
4094  File "c:\projects\datalad\datalad\support\parallel.py", line 265, in __iter__
4095    yield from self._iter_threads(self._jobs)
4096  File "c:\projects\datalad\datalad\support\parallel.py", line 417, in _iter_threads
4097    self.shutdown(force=True, exception=self._producer_exception or interrupted_by_exception)
[...]

PR #6473, opened 2022-02-18
Unify interface keyword args of metadata commands according to issue #6461

christian-monch:
This PR contains keyword unification work (issue #6461) of the metadata modules

#### 🏠 Internal
- ensure that internal datalad command calls do write results to stdout

PR #6489, opened 2022-02-22
Unify keyword arguments in search in spirit of issue #6461

christian-monch:
Disable result renderer in internal datalad API calls in search-code.

#### 🏠 Internal
- silence output from internal calls in `datalad search`

### DIFF
--- a/datalad/metadata/search.py
+++ b/datalad/metadata/search.py
@@ -301,7 +301,9 @@ def _search_from_virgin_install(dataset, query):
             "Performing search using DataLad superdataset %r",
             default_ds.path
         )
-        for res in default_ds.search(query):
+        for res in default_ds.search(query,
+                                     return_type="generator",
+                                     result_renderer="disabled"):
             yield res
         return
     else:
@@ -466,13 +468,7 @@ class _WhooshSearch(_Search):
         if not exists(index_dir):
             os.makedirs(index_dir)
 
-        # this is a pretty cheap call that just pull this info from a file
-        dsinfo = self.ds.metadata(
-            get_aggregates=True,
-            return_type='list',
-            result_renderer='disabled')
-
-        self._mk_schema(dsinfo)
+        self._mk_schema()
 
         idx_obj = widx.create_in(index_dir, self.schema)
         idx = idx_obj.writer(
@@ -495,7 +491,6 @@ class _WhooshSearch(_Search):
             lgr.info,
             'autofieldidxbuild',
             'Start building search index',
-            total=len(dsinfo),
             label='Building search index',
             unit=' Datasets',
         )
@@ -654,7 +649,7 @@ class _BlobSearch(_WhooshSearch):
                 val2str=True,
                 schema=None).items()))
 
-    def _mk_schema(self, dsinfo):
+    def _mk_schema(self):
         from whoosh import fields as wf
         from whoosh.analysis import StandardAnalyzer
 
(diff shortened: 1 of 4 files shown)


datalad/metadata/search.py
```diff
@@ -298,7 +301,9 @@ def _search_from_virgin_install(dataset, query):
             "Performing search using DataLad superdataset %r",
             default_ds.path
         )
-        for res in default_ds.search(query):
+        for res in default_ds.search(query,
+                                     return_type="generator",
+                                     result_renderer="disabled"):
```
mih, line 306: could be `yield from...`

mih (commented):
I am ok with removing the progress total, but maybe the commit message should make the rational explicit?

datalad/metadata/search.py
```diff
@@ -286,7 +286,10 @@ def _search_from_virgin_install(dataset, query):
                      "superdataset at %r?"
                      % DEFAULT_DATASET_PATH):
             from datalad.api import install
-            default_ds = install(DEFAULT_DATASET_PATH, source='///')
+            default_ds = install(
+                DEFAULT_DATASET_PATH,
+                source='///',
+                result_renderer='disabled')
```
yarikoptic, line 292: I actually think that in such cases it is ok to let commands render or we should `yield` up and render upstairs -- it is useful to communicate back what is going on/was done -- a large superdataset was installed.  So should either be "caught" and dataset extracted from result record and then re-yielded or could be left as it was IMHO

mih: Personally, I think always yielding is a good default.

datalad/metadata/search.py
christian-monch: Thx, changed

datalad/metadata/search.py
christian-monch: > 

True, true! Due to haste, that was left out.

mih:
This PR now has conflicts after https://github.com/datalad/datalad/pull/6473 was merged.

datalad/metadata/search.py
christian-monch: > "[...] always yielding is good"

I call `install` now with the `return_type='generator'` and disabled result rendering. All non-dataset results are yielded upwards.

christian-monch:
> I am ok with removing the progress total, but maybe the commit message should make the rational explicit?

I ammended the commit message

christian-monch:
Rebased, comments processed. Test failures are on Mac OS and one unrelated failure on Windows (as far as I can see).

Took a little longer, sorry for the delay.

mih:
Windows failure is #6482

mih (commented):
I tried to pin my comments on a specific change, but I give up now in the interest of time.

This PR introduces a different, somewhat new, paradigm for filtering results. It is related but not identical to the established result filtering via the standard `result_filter=` feature of any command. It also implements a different solution for `return_type=`.

Given that this new solution is only used in a single place, it is hard to judge whether it is functionally equivalent to what we have, or superior. I have the feeling that `return_type='item-or-list'` and `result_filter=is_ok_dataset` would achieve the same without the introduction of a new helper. But I do not have the resources right now to compare. I found no rational in the commit message or the PR description.

datalad/utils.py
```diff
@@ -2689,3 +2695,89 @@ def check_symlink_capability(path, target):
...
+
+    def __iter__(self):
+        # Imported here to prevent import loops
+        from .interface.results import YieldDatasets
+
+        results = yield from FilteringIterator.__iter__(self)
+        conversions = [
+            (YieldDatasets(self.success_only)(result), result)
+            for result in results
+        ]
```
mih, line 2771: Still digesting the desired behavior here, but wouldn't the list-comprehension here prevent a generator-style processing, because all results must be consumed, even when only a single dataset would be returned in the end?

Issue #6492, opened 2022-02-22
4 (maint)/6 (master) test fails with super fresh (3 days old) git-annex


yarikoptic:
https://github.com/datalad/git-annex/runs/5292823639?check_suite_focus=true

```
======================================================================
FAIL: datalad.core.local.tests.test_save.test_partial_unlocked
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/datalad/tests/utils.py", line 742, in _wrap_with_tempfile
    return t(*(arg + (filename,)), **kw)
  File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/datalad/core/local/tests/test_save.py", line 648, in test_partial_unlocked
    assert_repo_status(ds.path)
  File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/datalad/tests/utils.py", line 1624, in assert_repo_status
    % (state, state_files, oktobefound))
AssertionError: unexpected content of state "modified": [PosixPath('/tmp/datalad_temp_test_partial_unlockedgjlcmggh/culprit.txt')] != []

======================================================================
FAIL: datalad.core.local.tests.test_status.test_status_symlinked_dir_within_repo
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/datalad/tests/utils.py", line 742, in _wrap_with_tempfile
    return t(*(arg + (filename,)), **kw)
  File "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/datalad/core/local/tests/test_status.py", line 302, in test_status_symlinked_dir_within_repo
    call()
AssertionError: CommandError not raised

======================================================================
FAIL: datalad.local.tests.test_addurls.TestAddurls.test_addurls_metafail
----------------------------------------------------------------------
Traceback (most recent call last):
[...]

PR #6493, opened 2022-02-23
Relax permissions needed in RIA stores

### LINKED ISSUE #6254: ORA require more relaxed permissions than a local annex
Here is the situation:

I want to drop from an ORA remote

```
% git annex drop -f joc-storage sub-...
  [Errno 13] Permission denied: '/data/project/ukb_vbm/outputstore/893/8de76-0302-45b5-9825-3c6ce3f3fffe/annex/objects/w4/2V/MD5E-s5303257--d2dd1136a5770958323d713b131c85e7.tar.gz/MD5E-s5303257--d2dd1136a5770958323d713b131c85e7.tar.gz'\nNote: Write permissions for a key's parentdirectory are also required to drop content.
```

It literally means that. `w4` and `w4/2V` have write permissions, but `w4/2V/MD5E-s5303257--d2dd1136a5770958323d713b131c85e7.tar.gz` does not.

And indeed, a key pushed via ORA remote has a writable parent directory. But why?

```
% ls -l 21f/b481c-5e02-40ab-a036-1d4d62cc3385/annex/objects/pF/Zf                                                                                  
total 1
drwxrws--- 2 mih ukb 3 Nov 26 13:42 MD5E-s4--ba1f2511fc30423bdbb183fe33f3dd0f/
```

With an ephemeral clone, a direct non-ORA access would go directly to this key-dir and be able to delete the key, bypassing git-annex protections.

Is this justifiable?

PS: ORA access via `file://` URL.

bpoldrack:
Fixes #6254

Previously the ORA special remote relied on whatever the store end
is set up to with respect to permissions and accepted them as is,
meaning it didn't try to obtain write permissions on the parent
directory when dropping a key from store. However, there are good
reasons to revoke write permissions in store, particularly when the
object tree is meant to be used with ephemeral clones.
    
Hence, do what git-annex would do in the local case: View that as
protection against direct user interaction and try to obtain permission
on drop.

#### 🐛 Bug Fixes
- ORA special remote tries to obtain permissions when dropping a key from a RIA store rather than just failing. Thus having the same permissions in the store's object trees as one directly managed by git-annex would have, works just fine now. Fixes #6254 [#6493](https://github.com/datalad/datalad/pull/6493) (by @bpoldrack )

### DIFF
--- a/datalad/distributed/ora_remote.py
+++ b/datalad/distributed/ora_remote.py
@@ -568,9 +568,8 @@ class SSHRemoteIO(IOBase):
             with self.ensure_writeable(path.parent):
                 self._run('rm {}'.format(sh_quote(str(path))), check=True)
         except RemoteCommandFailedError as e:
-            raise RIARemoteError(f"Unable to remove {path}. Could not "
-                                 "obtain write permission for containing"
-                                 "directory.") from e
+            raise RIARemoteError(f"Unable to remove {path} "
+                                 "or to obtain write permission in parent directory.") from e
 
     def remove_dir(self, path):
         with self.ensure_writeable(path.parent):
--- a/datalad/distributed/ora_remote.py
+++ b/datalad/distributed/ora_remote.py
@@ -1,8 +1,12 @@
+import functools
 import os
+import stat
+import sys
 from pathlib import (
     Path,
     PurePosixPath
 )
+from contextlib import contextmanager
 import requests
 import shutil
 from shlex import quote as sh_quote
@@ -21,6 +25,7 @@ from datalad.customremotes.ria_utils import (
     UnknownLayoutVersion,
     verify_ria_url,
 )
+from datalad.utils import on_osx
 
 
 lgr = logging.getLogger('datalad.customremotes.ria_remote')
@@ -170,6 +175,39 @@ class IOBase(object):
 
 class LocalIO(IOBase):
     """IO operation if the object tree is local (e.g. NFS-mounted)"""
+
+    @staticmethod
+    @contextmanager
+    def ensure_write_permission_parent(destination):
+        """Context manager to get permissions on parent directory and restore
+        afterwards
+
+        If git-annex ever touch this key store, the keys will
+        be in mode 444 directories, and we need to obtain permission
+        first.
+
+        Parameters
+        ----------
+        destination: Path
+          path to the target file
+        """
+
+        par = destination.parent
+        # remember original mode -- better than to prescribe a fixed mode
+        par_mode = par.stat().st_mode
+        # only IWRITE works on windows, in principle
+        if not par_mode & stat.S_IWRITE:
+            par.chmod(par_mode | stat.S_IWRITE)
+            changed = True
+        else:
+            changed = False
+        try:
+            yield
+        finally:
+            if changed:
+                # restore original mode
+                par.chmod(par_mode)
+
     def mkdir(self, path):
         path.mkdir(
             parents=True,
@@ -212,18 +250,21 @@ class LocalIO(IOBase):
         #         Set output stream for output/error/progress line
 
     def rename(self, src, dst):
-        src.rename(dst)
+        with self.ensure_write_permission_parent(dst):
+            src.rename(dst)
 
     def remove(self, path):
         try:
-            path.unlink()
+            with self.ensure_write_permission_parent(path):
+                path.unlink()
         except PermissionError as e:
             raise RIARemoteError(
                 "Write permissions for a key's parent directory are "
                 "also required to drop content.") from e
 
     def remove_dir(self, path):
-        path.rmdir()
+        with self.ensure_write_permission_parent(path):
+            path.rmdir()
 
     def exists(self, path):
         return path.exists()
@@ -438,6 +479,53 @@ class SSHRemoteIO(IOBase):
             raise RIARemoteError("{}: {}".format(call, "".join(lines)))
         return "".join(lines[:-1])
 
+    @contextmanager
+    def ensure_write_permission_parent(self, destination):
+        """Context manager to get permissions on directory and restore
+        afterwards
+
+        If git-annex ever touch this key store, the keys will
+        be in mode 444 directories, and we need to give permission
+        first.
+
+        Parameters
+        ----------
+        destination: Path
+          path to the target file
+        """
+
+        par = sh_quote(str(destination.parent))
+        # remember original mode -- better than to prescribe a fixed mode
+
+        if on_osx:
+            format_option = "-f%Dp"
+            # on macOS this would return decimal representation of mode (same
+            # as python's stat().st_mode
+            conversion = int
+        else:  # win is currently ignored anyway
+            format_option = "--format=\"%f\""
+            # in opposition to the above form for macOS, on debian this would
+            # yield the hexadecimal representation of the mode; hence conversion
+            # needed.
+            conversion = functools.partial(int, base=16)
+
+        output = self._run(f"stat {format_option} {par}",
+                           no_output=False, check=True)
+        par_mode = conversion(output)
+        if not par_mode & stat.S_IWRITE:
+            new_mode = oct(par_mode | stat.S_IWRITE)[-3:]
+            self._run(f"chmod {new_mode} {par}")
+            changed = True
+        else:
+            changed = False
+        try:
+            yield
+        finally:
+            if changed:
+                # restore original mode
+                self._run("chmod {mode} {file}".format(mode=oct(par_mode)[-3:],
+                                                       file=destination))
+
     def mkdir(self, path):
         self._run('mkdir -p {}'.format(sh_quote(str(path))))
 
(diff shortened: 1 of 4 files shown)


bpoldrack:
FTR:

- There are issue with the macOS builds - see #6492
- code climate is mostly right, but out of scope here (plus a bunch of warnings about using `subprocess` and the need to be sure to only use trusted input)

datalad/distributed/ora_remote.py
```diff
@@ -167,9 +170,64 @@ def write_file(self, file_path, content, mode='w'):
...
+        # `pathlib.Path.chmod` (local IO).
+        raise NotImplementedError
+
 
 class LocalIO(IOBase):
     """IO operation if the object tree is local (e.g. NFS-mounted)"""
+
+    @staticmethod
+    @contextmanager
+    def ensure_write_permission_parent(destination):
```
mih, line 194: I think such a helper should live outside this class and outside `ora_remote.py`

datalad/distributed/ora_remote.py
```diff
@@ -167,9 +170,64 @@ def write_file(self, file_path, content, mode='w'):
...
+        -----
+
+        As pointed out in gh-5668, Ben is not convinced that this should really
+        be solved client-side, although it seems kind of fine, if we are able to
+        obtain the permission. However, store-side permissions should be set for
+        a reason, git-annex actually shouldn't touch the object tree and it
+        seems better to make sure that unaware people/programs don't
+        accidentally mess with a store's internal settings.
+        Hence, address the particular issue for the object tree only, not
+        enhancing the approach for all write operations ATM.
```
mih, line 212: This note is only relevant for a particular usage pattern in a particular context. Moreover, I think we should make it a habit of including personal message in documentation -- they can be placed in code comments.

datalad/distributed/ora_remote.py
```diff
@@ -167,9 +170,64 @@ def write_file(self, file_path, content, mode='w'):
...
+        par = destination.parent
+        # remember original mode -- better than to prescribe a fixed mode
+        par_mode = par.stat().st_mode
+        # only IWRITE works on windows, in principle
+        par.chmod(par_mode | stat.S_IWRITE)
+        try:
+            yield
+        finally:
+            # restore original mode
+            par.chmod(par_mode)
```
mih, line 229: I think the documentation should state that 3 additional FS calls are implied here -- this can mean a 3x slow-down for simple operations. Moreover, I think this helper should check after the first `stat` whether a permisson modification is actually necessary in order to achieve the goal -- it may just be writable already.

datalad/distributed/ora_remote.py
```diff
@@ -438,6 +507,51 @@ def _run(self, cmd, no_output=True, check=False):
...
+        -----
+
+        As pointed out in gh-5668, Ben is not convinced that this should really
+        be solved client-side, although it seems kind of fine, if we are able to
+        obtain the permission. However, store-side permissions should be set for
+        a reason, git-annex actually shouldn't touch the object tree and it
+        seems better to make sure that unaware people/programs don't
+        accidentally mess with a store's internal settings.
+        Hence, address the particular issue for the object tree only, not
+        enhancing the approach for all write operations ATM.
```
mih, line 529: See note on personal messages above.

mih (changes requested):
I left a bunch of comments. This PR goes beyond fixing #6254 and unconditionally performs 6 additional calls per stored key in order to ensure read-only permissions.

The motivation is to safe people from the troubles of potentially jeopardizing strict permissions when using a RIA store dataset annex in an ephemeral clone. The is certainly a laudable goal, but the cost seems substantial, and it is solving a problem that is only relevant for a subset of users (there is no possibility to make non-local ephemeral clones).

In the interest of a quick merge, it would be best to just fix #6254 here.

datalad/distributed/ora_remote.py
bpoldrack: This will require substantial RF'ing (it needs the `IO` classes to do shit). That should happen, yes, but not in this PR. This wi need to be a dedicated one (likely not just one).

mih: How does it need the `IO` class? It makes no single reference to it. And being here makes it harder to test, and less visible. Is there any advantage of having it in `LocalIO`?

bpoldrack:
Remaining CI failures are #6492 and #6482 AFAIK

datalad/distributed/ora_remote.py
bpoldrack: Well, yes - in case of the `LocalIO` it doesn't directly depend on it (as in doesn't use `self`). But it has a twin in the SSH class, which needs to use its `self`. Higher-level functions are not supposed to distinguish which IO they are using. Hence, nominally this function needs to stay. So - I could factor out the local case in the sense that the core of it becomes a helper outside and then this method here is just a wrapper around it. And the SSH/shell variant stays where it is. I thought you meant both. If you only want to have this one separate - can do.

bpoldrack: In that case, I'd go for a general "request write permission for `path`" and leave the "parent thereof" notion in here. Agree?

bpoldrack: @mih 

Pushed Proposal: https://github.com/datalad/datalad/pull/6493/commits/60151778b59005c9039d8c37b84fd65f7e63da6c

mih (commented):
Thx for the update!

What do you think about going one step further and convert `ensure_write_permission_parent(path)` to `ensure_write_permission(path)` and to adjust this pattern (and the similar ones)

```
with self.writeable_parent(path):
   path.rmdir()
```

to

```
with self.ensure_writeable(path.parent):
   path.rmdir()
```

There is nothing special about writable parents in the general case, and AFAICS the usage within the IO classes would not be limited by such a change.

bpoldrack:
@mih 

> There is nothing special about writable parents in the general case, and AFAICS the usage within the IO classes would not be limited by such a change.

To me the "special thing" about it is, that this contextmanager communicates that the target of whatever operation is to be performed within that context, is underneath of what is being manipulated by that contextmanager. That's because restoring the original mode would only be possible if the path still exists. The more general case requires either testing (more calls/disc IO - wouldn't want that in the special remote) within the contextmanager or knowledge how it's used (only the caller has that). That's why I added `obtain_write_permission`, which is not a contextmanager. So, your suggestion is fine but requires more awareness by callers.

mih:
I cannot really follow the claim of superior communication or less need for testing -- AFAICS it the needs are the same and so is the communication -- only the applicability of the tool is severely limited.

Feel free to revert or change, and merge. I should have communicated better. In a quick scan of the codebase I could not find a use case for having to make a parent dir writeable, but not a file inside. Moreover, the general implementation would also likely take the target mode as an argument, rather than making it specific for write. Sorry.

bpoldrack:
@mih 

I changed my mind. :)

So, contextmanager now takes the path it's supposed to obtain write permissions for. Saved the `exists` test on exiting the contextmanager I thought of. Wouldn't be as straightforward in the SSH case, but that's irrelevant for now.

> the general implementation would also likely take the target mode as an argument, rather than making it specific for write. 

Not sure about that. I think it's advantageous to aim for helpers that provide a higher abstraction than modes in order to take of different FS/systems at that lower level. Also: at some point there may be a need to deal with ACLs. Overall the notion of "mode" may not be that useful to propagate to higher-level functions.

Anyways, until there are more usecases in the code base, not worth spending more time right now, I think.

mih (approved):
Thx!

mih:
It seems this PR does not solve the problem underlying the original report. Investigating....

datalad/distributed/ora_remote.py
```diff
@@ -498,18 +560,21 @@ def get(self, src, dst, progress_cb):
...
-            self._run('rm {}'.format(sh_quote(str(path))))
+            with self.ensure_writeable(path.parent):
+                self._run('rm {}'.format(sh_quote(str(path))), check=True)
         except RemoteCommandFailedError as e:
-            raise RIARemoteError(
-                "Write permissions for a key's parent"
-                "directory are also required to drop content.") from e
+            raise RIARemoteError("Write permissions for a key's parent "
+                                 "directory are also required to drop "
+                                 "content.") from e
```
mih, line 573: Keeping this message unchanged is now confusing. The problem is no longer that there are no write permissions. The problem is that write permissions could not be obtained. This can happen when the path is owned by another user, for example. This should be corrected. Right now it would error like so:

```
Write permissions for a key's parent directory are also required to drop content. caused by [Errno 1] Operation not permitted: '/data/project/ukb_vbm/outputstore/893/8de76-0302-45b5-9825-3c6ce3f3fffe/annex/objects/j6/Pq/MD5E-s45608--c22377262d535937cc36659c70bb36a3.tar.gz'
```

the shown path is the parent directory.

mih (changes requested):
Sorry, @loj found an issue. Reporting on her behalf.

datalad/distributed/ora_remote.py
bpoldrack: Right, thx. Will fix.

bpoldrack: Pushed a changed message.

datalad/distributed/ora_remote.py
```diff
@@ -498,18 +560,21 @@ def get(self, src, dst, progress_cb):
...
-            self._run('rm {}'.format(sh_quote(str(path))))
+            with self.ensure_writeable(path.parent):
+                self._run('rm {}'.format(sh_quote(str(path))), check=True)
         except RemoteCommandFailedError as e:
-            raise RIARemoteError(
-                "Write permissions for a key's parent"
-                "directory are also required to drop content.") from e
+            raise RIARemoteError(f"Unable to remove {path}. Could not "
+                                 "obtain write permission for containing"
+                                 "directory.") from e
```
mih, line 573:
```suggestion
            raise RIARemoteError(f"Unable to remove {path} "
                                 "or to obtain write permission in parent directory.") from e
```

Thinking about this more, I'd say this is the better message. Without further inspection, we do not know whether this failure is about write permissions. Unlike before, the path reported in the exception will be the problematic one directly, and not just some proxy that requires expert knowledge to make the connection to interpret.

PR #6495, opened 2022-02-23
BF: No object tree in RIA w/o special remote

bpoldrack:
`create-sibling-ria` and the underlying helper created a dataset's `annex/objects` directory unconditionally.
However, only having the git repositories in store with no special remote and no data whatsoever is valid and supported usecase. The existence of that (empty) directory increases the inode count per dataset, is superfluous and leads to failures of annex commands being executed directly in store. While the latter is something to generally be very careful about, there are valid usecases and with the previous behavior we'd cause annex to pointlessly fail on discovery of that dir, indicating to annex that the repo is not annex-init'ed but was apparently used with annex before.

#### 🐛 Bug Fixes
- `create-sibling-ria` no longer creates an `annex/objects` directory in-store, when called with `--no-storage-sibling`. [#6495](https://github.com/datalad/datalad/pull/6495) (by @bpoldrack )

### DIFF
--- a/datalad/customremotes/ria_utils.py
+++ b/datalad/customremotes/ria_utils.py
@@ -181,7 +181,8 @@ def create_store(io, base_path, version):
     io.mkdir(error_logs)
 
 
-def create_ds_in_store(io, base_path, dsid, obj_version, store_version, alias=None):
+def create_ds_in_store(io, base_path, dsid, obj_version, store_version,
+                       alias=None, init_obj_tree=True):
     """Helper to create a dataset in a RIA store
 
     Note, that this is meant as an internal helper and part of intermediate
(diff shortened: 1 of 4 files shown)


datalad/customremotes/ria_utils.py
```diff
@@ -203,6 +204,7 @@ def create_ds_in_store(io, base_path, dsid, obj_version, store_version, alias=No
       layout version of the dataset itself (object tree)
     alias: str, optional
       alias for the dataset in the store
+    init_obj_tree: bool
```
mih, line 207: Worth the additional line to define what `obj_tree` is

bpoldrack: No idea why I didn't. Thx.

mih (approved):
LGTM

PR #6501, opened 2022-02-24
DOC: Add remark to old credential env vars

bpoldrack:
This outdated way of specifying credentials needs to be kept while at
the same time it is not documented (only indirectly in context of HCP OA
for example). Hence, the added note is important to stumble upon on
refactoring.

Just an internal comment - no changelog here.


bpoldrack:
Canceled builds. This is just a comment for future RF'ing of the credential system.

Issue #6490, opened 2022-02-22
CLI parser changes cause manpage build skip in extensions


mih:
I switched https://github.com/mih/datalad-mihextras over to build docs against upcoming 0.16 and it breaks the doc build. Reproduces locally, for other extensions too. Must be investigated prior release.

mih:
The reason is that the argparser used to determine for which commands to build manpages does not actually have any information on extension commands. This is the result of a behavior change after the CLI update that started with https://github.com/datalad/datalad/pull/6378.

I will investigate which would be the best way to restore the manpage generation -- it may be possible to do this without any general parser construction.

mih:
Doable, but complicated. A trivial fix that enables us to remove this release blocker is found.

PR #6502, opened 2022-02-24
Replace left-over (unused) mode label in CLI parser

mih:
This is used to support the weird parser construction mode that is
only used to construct parsers for manpage generation during extension
package documentation builds. Eventually this must all be factored
out into dedicated helpers that do this, and only only this -- instead
of being shoehorned into code that must run a quick as possible.

That being said, this fix removes this 0.16 release blocker, and makes
the broken extension package doc build work again.

Fixes datalad/datalad#6490

### Changelog
No entry needed, only broken in unreleased code.

### DIFF
--- a/datalad/cli/parser.py
+++ b/datalad/cli/parser.py
@@ -124,7 +124,7 @@ def setup_parser(
 
     command_provider = 'core'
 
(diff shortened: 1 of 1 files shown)


mih:
Failure pattern is #6492

Issue #5598, opened 2021-04-22
Repo method cache


mih:
I am planning to implement a replacement for manual caching of `Repo` method result caching. The type of caching I am talking about can be seen in the `copy_file()` code, but is also prevalent elsewhere.

The "cache" that I am aiming for is a class that:

- wraps a `Repo` instance
- implements any method call (in a generic fashion)
- keeps a record of any output in a dict with a key that combines (method name, and arguments)
- returns the cached result on a repeated request

In essence this creates a static variant of a `Repo` (not really static, because time is only frozen from the time of the first request with a particular parameterization). It needs to be used with care, but it has its use, I think.

Returned generators will need special handling, such that they can be unrolled multiple times.

yarikoptic:
wouldn't it work only
- in case of no external changes to the repository
- until any non-read-only invocation of external command is made
unless analysis of file stats is done on `.git/` directory files content as we do in fscacher?

if there were no aforementioned concerns, I guess a simple `@lru_cache` could have been just used for such methods?

mih:
Yes, that is what I wanted to say by calling it "static". That doesn't cover all use cases, but some, where we assume static already and have custom caching implemented:

- `diffstatus()` caches `get_content_info()` (which is actually coordinated from `diff()` and `status()`)
- `copy_file()` caches `get_dataset_root()`, `get_special_remotes()`

What these implementations do is to assume a static situation on disk in order to perform some kind of query or assessment to inform subsequent processing. And they do it for performance reasons.

I think it would be instrumental if we could establish some generic machinery for that: treat all repos as static and behave `lru_cache`-like.

I don't think `lru_cache` per se does what is needed. I think it is only meant to be used for functions that are always behaving in ways that keeps their previous return values valid for a later call. Here we have the situation that this should only ever happen in some specific context. To me this requires either a switch within `*Rep` (ugly), or a dedicated class.

I don't have a solid understanding yet, but I think such a tool would offer a good opportunity to decomplicate code.

bpoldrack:
If I understand the idea correctly, and instance of such a thing would be explicitly created for use in a context where we assume the repo to be "static" already and throw it away when that context is left.

Sounds to me like something that could be more generic (not specific to `*Repo`). 
A `CachedObject` class that:
 - defines a decorator function, that would deal with the cache dict
 - has a constructor that takes any object
 - implements `__getattr__` to wrap any method call with that decorator
 
 Something like that, @mih?

[...]

PR #6505, opened 2022-02-25
RF copy_file() with a cleaner (cached) Repo method access

### LINKED ISSUE #6504: Copy-file is slow
I am in the process of updating HCP subsets that were ``copy-filed`` from the large HCP dataset, such as  https://github.com/datalad-datasets/hcp-structural-preprocessed. Here is one exemplary run commit used to find the relevant files and copyfile them: https://github.com/datalad-datasets/hcp-structural-preprocessed/commit/cfdced1d00300c3637f7de2396f42cf9cc699a81

Rerunning these run commands, and thus ``copy-file``, to update files and availability information works nicely. However, it is slow. On juseless, it takes more than a second to process one file, and some datasets require updates 50k or 70k files. I found https://github.com/datalad/datalad/pull/5599, which looks like it could substantially speed things up with improved caching but was closed mainly due to a lack of interest.

adswa:
This reincarnates #5599 in new hopes to fix #6504.

@mih's original description was:

> This is a first example of an approximation of an idea that I sketched in https://github.com/datalad/datalad/issues/5598
Instead of using a complicated beast made of dicts and more dicts, I RF'ed all relevant Repo method access into an Interface helper that uses lru_cache and other means to minimize the actual interaction with the underlying Repo instance.
It does that with awareness of what type of access is being made and only works within the context of copy_file().
I mostly did this to see if these task-focused/tailored API classes can make sense, and de-convolute primary command logic from performance aspects.
Happy to get feedback on the approach.


datalad/local/copy_file.py
```diff
@@ -42,10 +42,151 @@
...
+        # TODO this could also be the place to stop lingering batch processes
+        if not self._tmpdir:
+            return
+
+        try:
+            self._tmpdir.rmdir()
+        except OSError as e:
+            lgr.warning(
+                'Failed to clean up temporary directory: %s',
+                exc_str(e))
```
mih, line 137: This still needs to be converted to use CapturedException

adswa: thanks for catching this! I pushed a fix

adswa:
So far, the pattern of failures is the known one of #6492 plus a few github connection errors:

```
 ERROR: Could not install packages due to an OSError: HTTPSConnectionPool(host='github.com', port=443): Max retries exceeded with url: /datalad/datalad-deprecated/archive/master.zip (Caused by ConnectTimeoutError(<pip._vendor.urllib3.connection.HTTPSConnection object at 0x7f7861d1b150>, 'Connection to github.com timed out. (connect timeout=15)'))
```

mih (approved):
LGTM! Thx!

PR #6508, opened 2022-02-25
More updates to get in the spirit if #6461

christian-monch:
This PR modifies get to use generator-based datalad.api-calls.

No changelog required.

### DIFF
--- a/datalad/distribution/get.py
+++ b/datalad/distribution/get.py
@@ -32,7 +32,6 @@ from datalad.interface.common_opts import (
     jobs_opt,
     reckless_opt,
 )
-from datalad.interface.results import is_ok_dataset
 from datalad.support.constraints import (
     EnsureInt,
     EnsureChoice,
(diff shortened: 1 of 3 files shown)


christian-monch:
Tests are failing on MacOS only. This seems unrelated to this PR

datalad/distribution/get.py
```diff
@@ -414,10 +415,21 @@ def _install_necessary_subdatasets(
...
+        result_renderer='disabled',
+        return_type='generator')
+
+    subds_trail = yield from FilteringIterator(
+        iterable=iterable,
+        patterns=[{
+            'action': 'subdataset',
+            'type': 'dataset',
+            'status': 'ok'}])
+
```
yarikoptic, line 432: Dunno enough, but isn't it what `result_filter=is_ok_dataset` in prior call for?

Also isn't condition before would always be false since we would always have an operator? (On the phone ATM, can't see coverage)

christian-monch: Good point. I should have removed "result_filter".

Result filter is different from FilteringIterator because the filtering iterator yields everything that is not filtered and keeps the filtered elements. A result filter, on the other hand, would not yield elements that do not pass the filter.

christian-monch:
Does not improve behavior in the current state

PR #6510, opened 2022-02-25
Adjust for recent git-annex reporting change on non-existent files

adswa:
This is a fix for one of the failing tests reported in #6492. 

A recent git-annex release changed the result of running 'git annex
find' on non-existing paths from an error to exit-code 0 with a more
informative hint:
```
$ git annex find --copies 0 --json --json-error-messages qwdqwd csdc
error: pathspec 'qwdqwd' did not match any file(s) known to git
error: pathspec 'csdc' did not match any file(s) known to git
Did you forget to 'git add'?
```
Previously, in order to determine whether non-existent paths were
queried, datalad relied on catching and parsing the command error.
```
CommandError: 'git -c diff.ignoreSubmodules=none annex find --copies 0
--json --json-error-messages -c annex.dotfiles=true -- nothere' failed
with exitcode 1
[info keys: stdout_json] [err: 'git-annex: nothere not found
find: 1 failed']
```
Because such a CommandError is not thrown anymore, this change parses
the stderror of the output depending on the git-annex version, looking
for files reported as not existing in a similar way to how the command
error is parsed, and assembled an output dictionary to report those
files as missing.

Happy to get feedback on this approach. I feel like it isn't particularly graceful.

## Changelog:

- Corresponding to the change in git-annex that a query of non-existent files will not result in a command error anymore, ``call_annex_records`` will parse the `stderr` of recent git-annex versions to identify operations on non-existing files.

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -1063,7 +1063,7 @@ class AnnexRepo(GitRepo, RepoInterface):
             )
         except CommandError as e:
             # Note: Workaround for not existing files as long as annex doesn't
-            # report it within JSON response:
+            # report it within JSON response (for annex versions < 10.20220128):
             # see http://git-annex.branchable.com/bugs/copy_does_not_reflect_some_failed_copies_in_--json_output/
             not_existing = [
                 # cut the file path from the middle, no useful delimiter
@@ -1116,7 +1116,24 @@ class AnnexRepo(GitRepo, RepoInterface):
             #        "Running %s resulted in stderr output: %s",
             #        args, shorten(e.stderr)
             #    )
-
+        if external_versions['cmd:annex'] >= '10.20220128':
+            if 'did not match any file(s) known to git' in out['stderr']:
+                not_existing = [
+                    line[17:-70] for line in out['stderr'].splitlines()
+                    if line.startswith('error:') and
+                    line.endswith('to git')
+                ]
+                if not_existing:
+                    out = {'stdout_json': []}
+                out['stdout_json'].extend(
+                    {
+                    "command": args[0],
+                    "file": f,
+                    "note": "not found",
+                    "success": False,
+                    }
+                    for f in not_existing
+                )
         json_objects = out.pop('stdout_json')
 
         if out.get('stdout'):


yarikoptic:
Thank you @adswa!  yet to review but meanwhile filed [an issue](https://git-annex.branchable.com/bugs/no_longer_non-0_exit_for___34__find_..._nonexisting__34__/?updated) for @joeyh since I believe that there ideally might be fixes on git-annex side.

joeyh:
The nonzero exit status is fixed in git-annex, but you will still need to look for new error messages when it does exit nonzero.

I'm a bit surprised that datalad is parsing error messages though. It would be better to have an API for this. Error messages can change at any time, or new ones be added that cover additional cases (part of what happened here really), or due to translation. Anywhere that datalad parses error messages feels to me like technical debt that perhaps you would want to address.

Indeed, those messages from git-ls-files that you're making it parse are in fact localized..

I've opened this to discuss such an API: https://git-annex.brachable.com/todo/api_for_telling_when_nonexistant_or_non_git_files_passed/

Also, switching on the git-annex version is not quite right, the new output happens when annex.skipunknown is set to false, which can be the case with older versions of git-annex as well. Only the default value of that config changed in the recent git-annex version.

adswa:
Thanks for fixing the error code!

> Also, switching on the git-annex version is not quite right, the new output happens when annex.skipunknown is set to false, which can be the case with older versions of git-annex as well. Only the default value of that config changed in the recent git-annex version.

Thanks much for this, I will adjust this conditional.

> I'm a bit surprised that datalad is parsing error messages though. It would be better to have an API for this. Error messages can change at any time, or new ones be added that cover additional cases (part of what happened here really), or due to translation. Anywhere that datalad parses error messages feels to me like technical debt that perhaps you would want to address.
Indeed, those messages from git-ls-files that you're making it parse are in fact localized..
I've opened this to discuss such an API: https://git-annex.brachable.com/todo/api_for_telling_when_nonexistant_or_non_git_files_passed/

Thanks a lot! I think we protect ourselves against localizations with
https://github.com/datalad/datalad/blob/c87cf8265a16fa02048b1ba78f564d68c6f07d53/datalad/runner/gitrunner.py#L107-L113 because there are a few places where we need to do it, but it would be great to not have to parse error messages anymore. Thanks for creating the issue, much appreciated!

bpoldrack:
@joeyh 
> I'm a bit surprised that datalad is parsing error messages though. It would be better to have an API for this. Error messages can change at any time, or new ones be added that cover additional cases (part of what happened here really), or due to translation. Anywhere that datalad parses error messages feels to me like technical debt that perhaps you would want to address.

Agreed. To put that into context: That piece of code is part of a generic function calling annex with `--json` and `--json-error-messages` and returning respective dictionaries in the python context. That function is not specific to a particular git-annex command. And what it does with that parsing is pretending that git-annex reported those errors as JSON for the cases where annex doesn't (yet). So, this parsing leads to the creation of something like `{"command": "find", "file": <whatever path was discovered by parsing>, "note": "not found", "success": False}` in order to have higher-level functions deal with annex' output uniformly. If annex could deliver something like that directly, that would, of course, be way better. Ideally, there'd be some indication of the problem that isn't meant to be a message to the user and not (so easily) subject to change - pretty much something that's akin to an exception name. That's because there are possible errors that we can deal with from within datalad if we can "understand" the problem - not everything needs to be passed onto the user.

bpoldrack:
Closing in favor of #6550

yarikoptic:
FTR: "error" ID is coming to git-annex for "not found" https://git-annex.branchable.com/todo/api_for_telling_when_nonexistant_or_non_git_files_passed/?updated#comment-c45982817c12f74bd1bf8d85f68cc92d

PR #6513, opened 2022-03-01
Disable internal save-result rendering in update()

mih:
Move `save()` call into a dedicated helper to reduce the cognitive
complexity of the monster Update.__call__() method.

In the spirit of datalad/datalad#6461

No changelog needed.


mih:
Most failures where "normal". Restarted one spurious run. Will merge when passed.

bpoldrack (approved):

Issue #6199, opened 2021-11-16
Unable to add a public gin repo as common data source


mslw:
It appears that there is a problem with configuring a gin repository as a common data source _if the repository is public_. Additionally, the behaviour is different depending on whether `.git` is included at the end of the `https` URL given to `datalad siblings add` (but problematic in both cases).

This seems related to #5144 (and also seems to be the initial source of the problem misrepresented in #6195 ).

To reproduce, I started by manually creating two repositories on gin, and making one of them public (the other is private by default).

Then, using the private one, the `datalad siblings add (...) --as-common-datasrc (...)` works as expected, producing only the typical warning:
```
mszczepanik@bnbnb64 in ~/Documents/tmp
❱ datalad create priv                                                                                 
[INFO   ] Creating a new annex repo at /home/mszczepanik/Documents/tmp/priv 
create(ok): /home/mszczepanik/Documents/tmp/priv (dataset)
mszczepanik@bnbnb64 in ~/Documents/tmp
❱ cd priv && cp ~/Downloads/*.jpg ./ && datalad save                                                  
(...)
action summary:                                                                                                                             
  add (ok: 3)
  save (ok: 1)
mszczepanik@bnbnb64 in ~/Documents/tmp/priv on git:main
❱ datalad siblings add -d . --name gin-update \
> --pushurl git@gin.g-node.org:/msz/my-private-repo.git \
> --url https://gin.g-node.org/msz/my-private-repo \
> --as-common-datasrc gin
[INFO   ] Could not enable annex remote gin-update. This is expected if gin-update is a pure Git remote, or happens if it is not accessible. 
[WARNING] Could not detect whether gin-update carries an annex. If gin-update is a pure Git remote, this is expected. Remote was marked by annex as annex-ignore. Edit .git/config to reset if you think that was done by mistake due to absent connection etc 
.: gin-update(-) [https://gin.g-node.org/msz/my-private-repo (git)]
```

However, with the public repo, the same command generates a git annex error:
```
mszczepanik@bnbnb64 in ~/Documents/tmp
❱ datalad create pub
[INFO   ] Creating a new annex repo at /home/mszczepanik/Documents/tmp/pub 
create(ok): /home/mszczepanik/Documents/tmp/pub (dataset)
mszczepanik@bnbnb64 in ~/Documents/tmp
❱ cd pub && cp ~/Downloads/*.jpg ./ && datalad save
(...)
action summary:                                                                                                                             
  add (ok: 3)
[...]

PR #6517, opened 2022-03-01
Refuse to configure common data source if annex-uuid is missing

mslw:
This WIP PR introduces an additional check / warning for `datalad siblings [add|configure] (...) --as-common-datasrc` to prevent unusable configuration. The update focuses on GIN use case, but I think it should be at least neutral, if not positive, for others.

### Overview

- For example, prior to this PR, trying to configure a sibling pointing to an empty GIN repo as a common data source produces an invalid `git-annex` configuration and may or may not produce an error: #6199 
- Adding / configuring a sibling involves a series of steps which add or change configuration entries as a result. Those related to the common data source are done last.
- The best indicator for validity of a sibling seems to be the presence of `remote.{name}.annex-uuid` in `.git/config`
  - Note: `create_sibling_gin` yields a configuration with `annex.ignore` unset, so `annex.ignore` alone cannot be checked
  - Note: adding a non-empty repo with `siblings add` will produce the `annex.uuid` entry before the proposed checks

### Proposed solution

- I propose to refuse configuring the sibling as a common data source if `annex.uuid` is missing, and to raise some kind of a warning rather then an error, because:
  - configuration steps prior to the common data source part may still be successful / meaningful
  - hard to determine in advance if the configuration will succeed (see note on adding a non-empty repo above)

### Introduced behaviour (example)

Compared to what happened previously, an attempt to configure an empty GIN repo additionally yields (impossible) result with an explanation. Behind the scenes, no entry is made in `git-annex:remote.log`.

```
datalad siblings add --url https://gin.g-node.org/msz/empty --pushurl git@gin.g-node.org:/msz/empty.git --name gin --as-common-datasrc gin-src
[INFO   ] Could not enable annex remote gin. This is expected if gin is a pure Git remote, or happens if it is not accessible. 
add-sibling(impossible): . (sibling) [cannot configure as a common data source; the remote has no annex UUID, which may indicate absence of remote git-annex branch; consider running push first]
[WARNING] Could not detect whether gin carries an annex. If gin is a pure Git remote, this is expected. Remote was marked by annex as annex-ignore. Edit .git/config to reset if you think that was done by mistake due to absent connection etc 
.: gin(-) [https://gin.g-node.org/msz/empty (git)]
```

### Open questions (feedback welcome!)

- What's the best way to communicate refusal to configure to users? I used `yield dict(status='impossible', message=...)` but maybe something else would be preferred?
- Are there other cases for which checking `remote.{name}.annex-uuid` wouldn't be the thing to do?
- What should a test for the behaviour look like?

### TODO
- [ ] Add a separate check for `annex.ignore=True`
- [ ] Add a test, if possible
- [ ] Write a changelog entry
- ...

#### 🐛 Bug Fixes
- Description... Fixes #issue
-
-
#### 📝 Documentation
-
#### 🏠 Internal
-
#### 🛡 Tests
-

### DIFF
--- a/datalad/distribution/siblings.py
+++ b/datalad/distribution/siblings.py
@@ -584,7 +584,21 @@ def _configure_remote(
                 # that git-annex will use too
                 remote_url = repo.config.get(f'remote.{name}.url')
                 ri = RI(remote_url)
-                if isinstance(ri, URL) and ri.scheme in ('http', 'https'):
+                if not (isinstance(ri, URL) and ri.scheme in ('http', 'https')):
+                    yield dict(
+                        status='impossible',
+                        message='cannot configure as a common data source, '
+                                'URL protocol is not http or https',
+                        **result_props)
+                elif f'remote.{name}.annex-uuid' not in repo.config:
+                    yield dict(
+                        status='impossible',
+                        message='cannot configure as a common data source; '
+                                'the remote has no annex UUID, which may '
+                                'indicate absence of remote git-annex branch; '
+                                'consider running push first',
+                        **result_props)
+                else:
                     # XXX what if there is already a special remote
                     # of this name? Above check for remotes ignores special
                     # remotes. we need to `git annex dead REMOTE` on reconfigure
@@ -598,12 +612,7 @@ def _configure_remote(
                         'type=git',
                         'location={}'.format(remote_url),
                         'autoenable=true'])
-                else:
-                    yield dict(
-                        status='impossible',
-                        message='cannot configure as a common data source, '
-                                'URL protocol is not http or https',
-                        **result_props)
+
     #
     # place configure steps that also work for 'here' below
     #
--- a/datalad/distribution/siblings.py
+++ b/datalad/distribution/siblings.py
@@ -590,20 +590,23 @@ def _configure_remote(
                         message='cannot configure as a common data source, '
                                 'URL protocol is not http or https',
                         **result_props)
-                elif f'remote.{name}.annex-uuid' not in repo.config:
-                    yield dict(
-                        status='impossible',
-                        message='cannot configure as a common data source; '
-                                'the remote has no annex UUID, which may '
-                                'indicate absence of remote git-annex branch; '
-                                'consider running push first',
-                        **result_props)
                 else:
                     # XXX what if there is already a special remote
                     # of this name? Above check for remotes ignores special
                     # remotes. we need to `git annex dead REMOTE` on reconfigure
                     # before we can init a new one
                     # XXX except it is not enough
+
+                    # Warn if configuring something that has no annex UUID
+                    # which will produce a weird entry in git-annex:remote.log
+                    if f'remote.{name}.annex-uuid' not in repo.config:
+                        lgr.warning(
+                            f'Configuring the remote {name} as a common data'
+                            f'source {as_common_datasrc}, but the resulting '
+                            f'special remote configuration may be incomplete. '
+                            f'The {name} remote has no annex UUID, which may '
+                            f'indicate absence of git-annex branch on the '
+                            f'remote end; consider pushing / updating first.')
 
                     # make special remote of type=git (see #335)
                     repo.call_annex([


datalad/distribution/siblings.py
```diff
@@ -584,7 +584,21 @@ def _configure_remote(
...
-                if isinstance(ri, URL) and ri.scheme in ('http', 'https'):
+                if not (isinstance(ri, URL) and ri.scheme in ('http', 'https')):
+                    yield dict(
+                        status='impossible',
+                        message='cannot configure as a common data source, '
+                                'URL protocol is not http or https',
+                        **result_props)
+                elif f'remote.{name}.annex-uuid' not in repo.config:
+                    yield dict(
+                        status='impossible',
```
bpoldrack, line 595: See general comment on this approach.

bpoldrack (commented):
I think that approach is a bit problematic. That's b/c a sibling configuration can be a completely local operation. Hence, a config can be valid w/o `datalad sibling` knowing whether that's the case, since the remote needs to be accessed in order to figure whether it has an annex-uuid. I may simply be offline while set things up. It would then be annoying to be denied just because datalad couldn'T connect and double-check me. A WARNING seems fine to me, but failing not so much.

datalad/distribution/siblings.py
```diff
@@ -584,7 +584,21 @@ def _configure_remote(
                 # that git-annex will use too
                 remote_url = repo.config.get(f'remote.{name}.url')
                 ri = RI(remote_url)
-                if isinstance(ri, URL) and ri.scheme in ('http', 'https'):
+                if not (isinstance(ri, URL) and ri.scheme in ('http', 'https')):
+                    yield dict(
+                        status='impossible',
```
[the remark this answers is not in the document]
mih: I don't think that is true. `git-annex` itself will complain about any URL that is not http(s).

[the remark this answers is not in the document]
mslw: In either case, that check was already present, and I only restructured the if/else.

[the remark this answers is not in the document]
bpoldrack: You're right. I take that back. Mistook the URL this is talking about.

datalad/distribution/siblings.py
bpoldrack: It could turn out to work later, I think. Even in the case you are targeting: Just because our last knowledge of that GIN repo is, that it is empty and has no annex-uuid, it could get one and that doesn't necessarily have to happen via THIS clone.

mslw: Would the warning be just `warnings.Warn(...)` or is there a dedicated method?

bpoldrack: There should be logger instance in that file (usually called `lgr`). It's then `lgr.warn()`.

mslw:
@bpoldrack I tried he scenario you suggested with code that only shows a warning:

In "original" location:
1. Create a gin sibling with `datalad create-sibling-gin somename --name gin --access-protocol https-ssh --credential gin --private`
2. Configure as common data source with `datalad siblings configure --name gin --as-common-datasrc gin-src` (in addition to my warning, this set `annex-ignore` config value to True and warned about it), and produced an entry in `git-annex:remote.log` without UUID)
3. Set `annex-ignore` back to true

In a local clone:
4. Made a local clone.
5. Added the same gin repo as a sibling (just a regular one, not a common data source)
6. Added files and pushed (at this moment, the git-annex branch was created on GIN)

Back in the "original" location:
7. Updated from gin sibling
8. Tried pushing without adding files
9. Added files and pushed

The `uuid.log` file seems ok, but still, the `remote.log` is incomplete (no UUID), which I believe is a sign of trouble
```
git show git-annex:remote.log
 autoenable=true location=https://gin.g-node.org/msz/somename name=gin-src type=git timestamp=1647334900.504500391s
```

And the way to fix it is by repeating the same `datalad siblings configure --name gin --as-common-datasrc gin-src` command  (i.e. it doesn't seem to fix itself through pushing and pulling) - this, then, produces 
```
git show git-annex:remote.log
06a6beb5-7f77-49ad-bc53-5e6b1aa9d244 autoenable=true location=https://gin.g-node.org/msz/somename name=gin-src type=git timestamp=1647337716.387780317s
```

Regarding the proposed approach: I think an explicit warning is better than no warning at all. However, I see it as a small improvement rather than a solution. Partially because the configuration step needed to be redone, partially because this warning comes early, and problems would manifest only later, when trying to "propagate" the autoenabled remote to other clones.

So I may go with the warning to clean up messaging a little, but ATM I don't see what would be a complete solution.

mslw:
Going back to:
> Just because our last knowledge of that GIN repo is, that it is empty and has no annex-uuid, it could get one and that doesn't necessarily have to happen via THIS clone.

If the scenario is: configure sibling (not yet common data source) - populate from another clone - configure as common data source, then my error/warning wouldn't trigger... because siblings calls `repo.enable_remote` (unless annex.ignore is set) just before it gets to the common data source part.

So if I understand correctly, our last knowledge is just from moments ago.

bpoldrack:
>  (i.e. it doesn't seem to fix itself through pushing and pulling)

Yes. Since in the other clone you explicitly didn't configure the type-git special remote (commondatasrc), there's no change to `remote.log` that could have been pulled.

At a first glance I can't see what exactly is different w/ public/private repos on GIN. But it seems that the public one can't be reached by that URL. Did you try to use URLs with a trailing `.git`?


As for the failure vs. warning:

I'm torn. I see your point. The thing is, normally I'd want a config to be set if I say so. It may fail when it's being used but that shouldn't necessarily prevent me from setting it. However, `datalad siblings` (or `git annex initremote/enableremote`) kinda mix up setting a config and using it. I'm honestly not sure what to do here.

I haven't looked into `siblings` code, but we might be able to work out how to fail with the respective `enableremote`/`initremote` cals, yielding an error result while still succeeding with possible other parts that only involve a `git config` call.

mslw:
Thanks for the comments. Answering partially:
> At a first glance I can't see what exactly is different w/ public/private repos on GIN. But it seems that the public one can't be reached by that URL. Did you try to use URLs with a trailing .git?

From a quick check, if `.git/config` has a URL with trailing .git, direct call to `git annex initremote` does not error. But at this stage I'm confused whether initremote checks just the config file, or tries to reach that url - because for private repos it doesn't matter.

mslw:
I'm closing this, because solving the issue will require another, fresh look at a) the conditions we want to address b) the desired behaviour c) whether it's an issue with error handling already present in the siblings command. This all probably goes beyond a single warning call.

However, there's a lot of useful comments in this thread, and I hope to go back to siblings' messaging in unspecified future.

PR #6518, opened 2022-03-02
Support next generation metadata in search

christian-monch:
This PR extends search to use next generation metadata, if a next generation metalad extension is found.

This PR is marked as work in progress and exists to support experiments with the new metadata system. One shortcoming of this patch is that legacy and NG-metadata results are both used and not de-duplicated. Also, the mapping of NG-metadata on the expected result structure, i.e. the legacy metadata-results, is probably sub-optimal.

- [x] possibly add a flag to `search` that allows the selection of the metadata that should be searched.
- [x] add test

- use next generation metadata code in search, if it is availble

### DIFF
--- a/datalad/metadata/metadata.py
+++ b/datalad/metadata/metadata.py
@@ -1045,10 +1045,10 @@ class Metadata(Interface):
 
 
 def gen4_query_aggregated_metadata(reporton: str,
-                                 ds: Dataset,
-                                 aps: List[Dict],
-                                 recursive: bool = False,
-                                 **kwargs):
+                                   ds: Dataset,
+                                   aps: List[Dict],
+                                   recursive: bool = False,
+                                   **kwargs):
     """Query metadata in a metadata store
 
     Query paths (`aps["path"]`) have to be contained in the poth of the ds.
--- a/datalad/metadata/metadata.py
+++ b/datalad/metadata/metadata.py
@@ -1140,7 +1140,7 @@ def query_aggregated_metadata(reporton: str,
                               ds: Dataset,
                               aps: List[Dict],
                               recursive: bool = False,
-                              use_metadata: Optional[str] = None,
+                              metadata_source: Optional[str] = None,
                               **kwargs):
     """Query legacy and NG-metadata stored in a dataset or its metadata store
 
@@ -1156,11 +1156,11 @@ def query_aggregated_metadata(reporton: str,
     recursive : bool
       Whether or not to report metadata underneath all query paths
       recursively.
-    use_metadata : Optional[str]
-      Metadata generation that should be used. If set to "legacy", only
-      prior metalad version 0.3.0 metadata will be queried, if set to "gen4",
-      only metalad version 0.3.0 and beyond metadata will be queried, if set
-      to 'None', both will be queried.
+    metadata_source : Optional[str]
+      Metadata source that should be used. If set to "legacy", only metadata
+      prior metalad version 0.3.0 will be queried, if set to "gen4", only
+      metadata of metalad version 0.3.0 and beyond will be queried, if set
+      to 'None', all known metadata will be queried.
     **kwargs
       Any other argument will be passed on to the query result dictionary.
 
@@ -1170,7 +1170,7 @@ def query_aggregated_metadata(reporton: str,
       Of result dictionaries.
     """
 
-    if use_metadata in (None, "legacy"):
+    if metadata_source in (None, "legacy"):
         yield from legacy_query_aggregated_metadata(
             reporton=reporton,
             ds=ds,
@@ -1179,7 +1179,7 @@ def query_aggregated_metadata(reporton: str,
             **kwargs
         )
 
-    if use_metadata in (None, "gen4") and next_generation_metadata_available:
+    if metadata_source in (None, "gen4") and next_generation_metadata_available:
         yield from gen4_query_aggregated_metadata(
             reporton=reporton,
             ds=ds,
--- a/datalad/metadata/metadata.py
+++ b/datalad/metadata/metadata.py
@@ -23,6 +23,7 @@ from pathlib import Path
 from typing import (
     Dict,
     List,
+    Optional,
 )
 
 from datalad import cfg
@@ -1120,6 +1121,7 @@ def query_aggregated_metadata(reporton: str,
                               ds: Dataset,
                               aps: List[Dict],
                               recursive: bool = False,
+                              use_metadata: Optional[str] = None,
                               **kwargs):
     """Query legacy and NG-metadata stored in a dataset or its metadata store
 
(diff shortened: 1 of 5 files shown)


yarikoptic:
do you have some dataset(s) available publicly to try it out on?

christian-monch:
And my venv: `~cristian/venv/datalad-metalad-python3.9/`

datalad/metadata/metadata.py
```diff
@@ -70,6 +76,14 @@
 from datalad.log import log_progress
 from datalad.core.local.status import get_paths_by_ds
 
+# Check availability of new-generation metadata
+try:
+    from datalad_metalad.dump import Dump
+    next_generation_metadata_available = True
+except ImportError:
+    next_generation_metadata_available = False
```
yarikoptic, line 84: I think it would not hurt if we start "versioning" the metadata approach ;) I would count this as `_gen4`

- 1st gen: `.datalad/meta` (old implementation within `datalad`)
- 2nd gen: `.datalad/metadata/aggregate_v1.json` (currently in `datalad`)
- 3rd gen: the "old" one of metalad
- 4th gen: this "next_generation" - current as in `metalad`, right?
- 
May be even add that info into `docs/source/metadata.rst` to reach clarity? WDYT @mih -- I think having some codenames for metadata approaches would be helpful, at least to me.

datalad/metadata/metadata.py
```diff
@@ -172,11 +186,11 @@ def _get_containingds_from_agginfo(info, rpath):
     return dspath
 
 
-def query_aggregated_metadata(reporton, ds, aps, recursive=False,
-                              **kwargs):
+def legacy_query_aggregated_metadata(reporton, ds, aps, recursive=False,
```
yarikoptic, line 189: correspondingly `gen2_query_...`

datalad/metadata/metadata.py
```diff
@@ -1022,3 +1036,127 @@ def custom_result_renderer(res, **kwargs):
                  if meta else ' -' if 'metadata' in res else ' aggregated',
             tags='' if 'tag' not in meta else ' [{}]'.format(
                  ','.join(ensure_list(meta['tag'])))))
+
+
+def ng_query_aggregated_metadata(reporton: str,
```
yarikoptic, line 1041:
```suggestion
def gen4_query_aggregated_metadata(reporton: str,
```

yarikoptic (changes requested):
Add a test.  Could be a matrix run (in travis or appveyor, I don't care) which installs  metalad, extracts/aggregates metadata, and performs search.  Or do you think we should publish some sample dataset like that or may be even "move" search into `metalad` altogether?

Also I left a comment about possibly bringing some "identification" to various metadata approaches datalad went through since current "next_generation" might be "echo_from_the_past" in the future.

Adding an option on which metadata to search sounds like a useful thing given present multiplicity of metadatas

mih:
I have not talked with @christian-monch about his plans with this code, but personally I'd like to see all of `datalad/metadata` be removed from the core package as soon as possible. It is one of these pieces that connects large chunks of otherwise obsolete functionality. Therefore adding more layers of backward compatibility is not attractive too me, except when it come with the consideration for maintenance load.

yarikoptic:
I think it will indeed make sense to remove all metadata handling and move `search` into -metalad (or should it be `datalad-metalad-based-search` extension?) whenever metalad provides feature parity to replace metadata handling in the core.  Does it ATM? (IIRC it also has already at least 2 levels of different metadata formats and not yet sure if does provide handling in the most recent reincarnation for extractors in extensions such as neuroimaging).

NB: originally posted this comment in another PR by mistake ;)

christian-monch:
My plans for metadata beyond this PR are to remove metadata code from datalad core and rely solely on the datalad-metalad extension. This would support the execution of old and new extractors, and the handling of new git-stored metadata out of the box. 

### Metadata content
To support old metadata content, i.e `aggregate_v1.json` plus its object files, there are a number of possibilities that can co-exist:

1. Have a datalad-metalad-old extension, that would basically be the datalad-metalad-0.2.x. This would allow to switch between processing old and new metadata.

2. Convert old metadata to new metadata (issue datalad-metalad/#118 https://github.com/datalad/datalad-metalad/issues/118). This would allow unified processing with the new metadata extension.


### Search
Since search is orthogonal to metadata storage and handling (if it uses a high level metadata-access API), it can be altered independently. I think we might want to take search out of datalad core and move it either into the datalad-metalad extension, or into its own extension. It makes sense to combine it with datalad-metalad because it has a strong dependency on the metadata-access API. On the other hand, the one size fits all approach to searching might not be able to tackle specific search requirements, e.g. graph-based search. For such cases a specific search-extension might be better suited.

I would like to put these ideas out for discussion here. I will move them to individual issues if we have some basic agreement on the general direction.

datalad/metadata/metadata.py
christian-monch: Good point. `_gen4` would be appropriate, if we discriminate between `_gen2` and `_gen3` based on the interface, since the storage is the same (at least in `_gen2` and `_gen3`).

I will start using "generation four".

datalad/metadata/metadata.py
christian-monch: The `query_aggregated_metadata` method from master would actually be compatible with at least `gen2` and `gen3` (if I am not mistaken, @mih could clarify that).

I would therefore still rename it to `legacy_query_aggregated_metadata` because it covers two generations (due to the the shared storage model),

datalad/metadata/metadata.py
```diff
@@ -70,6 +76,21 @@
 from datalad.log import log_progress
 from datalad.core.local.status import get_paths_by_ds
 
+# Check availability of new-generation metadata
+try:
+    from datalad_metalad.dump import Dump
+    from datalad_metalad.exceptions import NoMetadataStoreFound
```
yarikoptic, line 82: this line is claimed to be not test covered -- so this "new" thing is not yet released or is there another reason for ImportError being raised?... oh well -- we do not submit coverage from extensions testing at all, filed https://github.com/datalad/datalad/issues/6780

datalad/metadata/search.py
```diff
@@ -1304,6 +1314,14 @@ class Search(Interface):
             doc="""if given, the formal query that was generated from the given
             query string is shown, but not actually executed. This is mostly useful
             for debugging purposes."""),
+        use_metadata=Parameter(
```
yarikoptic, line 1317: let me think about it more but ATM I dislike `use_metadata` which immediately suggested me being `bool` and then it makes little sense since search works only on metadata. Needed me to get to definition of `use_metadata` here (from points of use above) to understand what it actually means.
Why such confusion: Even though we do have `@use_casssette(cassette_name)`, typically anything `use_` is `bool`.  So may be something like `metadata_type` or `metadata_format` would be a better name and align better with its docstring?

christian-monch: Changed the parameter name from `use_metadata` to `metadata_source`.

christian-monch:
> I have restarted that run

Thx. Haven't seen that before. Will keep an eye on it

christian-monch:
Git annex version was updated to fix the stalling tests in appveyor.

The newly selected version of git-annex comes together with ssh version  9, which triggers an error in the ssh copy test (see issue #6655).

yarikoptic:
ok, let's see where it would bring us! ;)

PR #6521, opened 2022-03-07
require_dataset() uniformly raises NoDatasetFound

### LINKED ISSUE #6503: datalad-configuration dump not working without dataset
```
/tmp % datalad configuration
[ERROR  ] ValueError(No installed dataset found at /tmp.) (ValueError) 
```

This should not happen, and likely rather dump the non-dataset config.

Compounded by

```
(datalad-dev) mih@meiner ~ % datalad configuration --scope local
[ERROR  ] ValueError(Scope selection is not supported for dumping) (ValueError) 
(datalad-dev) 1 mih@meiner ~ % datalad configuration --scope global
[ERROR  ] ValueError(Scope selection is not supported for dumping) (ValueError) 
```

mih:
Before it claimed to raise InsufficientArgumentsError uniformly,
but actually raised NoDatasetFound or ValueError, depending
on the exact error condition.

This change consolidates and documents raising NoDatasetFound,
which seems to make most sense.

#### 🐛 Bug Fixes
- `require_dataset()` now uniformly raises `NoDatasetFound` when no dataset was found. Implementations that catch the previously documented `InsufficientArgumentsError` or the actually raised `ValueError` will continue to work, because `NoDatasetFound` is derived from both types. Fixes #6503

### DIFF
--- a/datalad/distribution/dataset.py
+++ b/datalad/distribution/dataset.py
@@ -562,7 +562,12 @@ def require_dataset(dataset, check_installed=True, purpose=None):
     Returns
     -------
     Dataset
-      Or raises an exception (InsufficientArgumentsError).
+      If a dataset could be determined.
+
+    Raises
+    ------
+    NoDatasetFound
+      If not dataset could be determined.
     """
     if dataset is not None and not isinstance(dataset, Dataset):
         dataset = Dataset(dataset)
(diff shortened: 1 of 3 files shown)


mih:
Metalad's `meta_extract()` is specifically catching a `NoDatasetFound` exception and yield error results, but two of its tests do not expect this in all cases. If filed https://github.com/datalad/datalad-metalad/issues/218 which is describing the conditions surrounding this behavior.

PR #6525, opened 2022-03-07
ORA: discontinue use of duplicate SSHManager

### LINKED ISSUE #6523: ORA remote maintains its own SSHManager instance
Maybe this made sense when its implementation did not import any datalad code, but at present it is just a duplicate instance, where `datalad.ssh_manager could/should be used instead.

mih:
Instead use the one that already exists in any process. Which also
already has `atexit` setup.

#### 🏠 Internal
- ORA remote no longer uses its on `SSHManager` instance, but the standard process-wide one provided by the `datalad` module. Fixes #6523

### DIFF
--- a/datalad/distributed/ora_remote.py
+++ b/datalad/distributed/ora_remote.py
@@ -14,6 +14,7 @@ import subprocess
 import logging
 from functools import wraps
 
+from datalad import ssh_manager
 from datalad.customremotes import (
     RemoteError,
     SpecialRemote,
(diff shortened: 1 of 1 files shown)


mih:
I will merge this in a couple of hours, unless stopped.

bpoldrack (approved):
That should be fine. I think the original reason was to not import `datalad` at all for fast startup, but that reason is kinda obsolete.

mih:
Thx for the merge. FTR: the original reason is, in fact, obsolete, because there is already an import and with it, there is also the SSHManager instance unconditionally present.

PR #6532, opened 2022-03-10
Stop including annex-bundle info in SSH connection socket name

### LINKED ISSUE #6531: SSH socket hash name depends on bundle setup
`get_connection_hash()` takes `bundled=`.

I don't understand why this was done, and I left no note to myself on this topic in https://github.com/datalad/datalad/pull/3402

As far as I can see it, the socket provides an authenticated connection, but no persistent shell. Any command adjustments for use of a bundled Git are per-call, and do not anyhow impact the SSH control master process.

I think this was a mistake. It has implications on code complexity (any code that needs to find that socket must be aware of a bundle config flag), and in the worst case, this would lead to opening superfluous connections to the same host, based on an unrelated setting.

mih:
As far as I can see it, the socket provides an authenticated connection,
but no persistent shell. Any command adjustments for use of a bundled
Git are per-call, and do not anyhow impact the SSH control master
process.

The `bundled` parameter of `get_connection_hash()` is ignored from now,
but kept to not break user code. Deprecation messages have been added.

- The `bundled` parameter of `get_connection_hash()` is now ignored and will be removed with a future release. Fixes #6531.

### DIFF
--- a/datalad/support/sshconnector.py
+++ b/datalad/support/sshconnector.py
@@ -53,7 +53,7 @@ lgr = logging.getLogger('datalad.support.sshconnector')
 
 
 def get_connection_hash(hostname, port='', username='', identity_file='',
-                        bundled='', force_ip=False):
+                        bundled=None, force_ip=False):
     """Generate a hash based on SSH connection properties
 
     This can be used for generating filenames that are unique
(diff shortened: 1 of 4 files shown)


bpoldrack (approved):
Makes sense and LGTM.

Issue #6524, opened 2022-03-07
Determine fate of `datalad.ssh_manager`


mih:
This SSH-related code in DataLad is not cross-platform ready. Following up in https://github.com/datalad/datalad/issues/5829#issuecomment-1060640263 this is an exploration where it is actually used, to determine possible migration strategies and (hidden) failures.

In the core package, `datalad.support.sshconnector` is unused, except for an import of `SSHManager` in the top-level package in order to instantiate `ssh_manager = SSHManager()`.

The `ssh_manager` instance is used in the following four places to open SSH connections, which are subsequently used to execute remote commands.

```
datalad/distributed/create_sibling_ria.py:        ssh = ssh_manager.get_connection(
datalad/distributed/ora_remote.py:        self.ssh = ssh_manager.get_connection(
datalad/distribution/create_sibling.py:            shell = ssh_manager.get_connection(sshurl)
datalad/support/sshrun.py:        ssh = ssh_manager.get_connection(sshurl, force_ip=force_ip)
```

The only other usage is in `GitRepo`, where it is anticipating the opening of SSH connections by underlying commands, and enables connection sharing to minimize re-authentication and enable parallelization.

In particular the latter use case is broken, because this code does not allow for any interactivity in its present form (i.e. typing in ssh (key) passwords).

I wonder if any of these use cases is actually needed, and could not be replaced by calling `Dataset.sshrun()`?

No `Repo` method makes any use of `SSHManager` functionality.

`create_sibling_ria()` only makes straightup calls to `ssh`, which is exactly what `sshrun` is for.

`ora_remote` used it to open a connection and read out the parameters how it was done, in order to bypass it subsequently with a dedicated `ssh` process that it manages on its own. The exception is the execution of `scp` for `.get()` and `.put()`.

`create_sibling()` uses an opened SSH connection interchangeably with a local shell via a proxy class, again primarily for remote command execution, but also `put`.

#### Thoughts

Superficially it seems to make sense to consolidate all SSH-related functionality into dedicated commands, rather than having them fused into the core. I think two developments are needed:

- provide a slim wrapper around `scp` in a similar manner as `sshrun` wraps `ssh`
- implement a strategy to have multiple simultaneous `sshrun` command reliably create and use the same authenticated connection, likely via some kind of inter-process lock

Moreover, the present unconditional inspection of remote hosts for git-annex executables must stop and be done in layer of code that actually expect git-annex to be around.

yarikoptic:
[...]

PR #6537, opened 2022-03-11
Stop forcefully disabling keyboard-interactive SSH auth

mih:
We did this initially for non-multiplex connections, assuming that it
would not work "by-design" when the SSH process is controlled by a
process that is not directly connected to a terminal on windows.

However, explorations documented in https://github.com/datalad/datalad/issues/5829
and https://github.com/datalad/datalad/issues/6524 revealed that this is
wrong, and the reason for it appearing to not work is in the protocol
handling within `Runner`.

This patch merely removed the forceful disabling, and thereby
immediately makes `sshrun` functional on windows.

#### 🐛 Bug Fixes
- Keyboard-interactive authentication is now possibly with non-multiplexed SSH connections (i.e., when no connection sharing is possible, due to lack of socket support, for example on Windows). Previously, it was disabled forcefully by DataLad for no valid reason.

### DIFF
--- a/datalad/support/sshconnector.py
+++ b/datalad/support/sshconnector.py
@@ -347,27 +347,6 @@ class NoMultiplexSSHConnection(BaseSSHConnection):
     The connection is opened for execution of a single process, and closed
     as soon as the process end.
     """
-    def __init__(self, sshri, **kwargs):
-        """Create a connection handler
-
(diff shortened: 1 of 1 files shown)


mih:
Windows failure continues to be https://github.com/datalad/datalad/issues/6482

PR #6540, opened 2022-03-11
DOC: sshconnector - remove duplicate docstrings in subclasses

### LINKED ISSUE #6536: make (more) use of @borrowdoc
while reviewing https://github.com/datalad/datalad/pull/6533/files and then 
```
$> git grep '@borrow' | grep -v test_
datalad/dochelpers.py:       @borrowdoc(Mapper)
datalad/dochelpers.py:           @borrowkwargs(A)
datalad/downloaders/base.py:    # @borrowkwargs(BaseDownloader, 'get_status')
datalad/downloaders/http.py:    @borrowkwargs(BaseDownloader)
datalad/downloaders/s3.py:    @borrowkwargs(BaseDownloader)
datalad/downloaders/shub.py:    @borrowkwargs(HTTPDownloader)
datalad/downloaders/shub.py:    @borrowkwargs(HTTPDownloader)
datalad/runner/gitrunner.py:    @borrowdoc(WitlessRunner)
datalad/support/annexrepo.py:    @borrowdoc(GitRepo, 'is_valid_git')
datalad/support/annexrepo.py:    @borrowkwargs(GitRepo)
datalad/tests/utils.py:@borrowkwargs(methodname=make_tempfile)
```
I realized that for some reason we aren't using it much whenever we should!  @mih -- any objections to apply it post #6533 ?

yarikoptic:
Besides providing some convenience to developer to see docstring right by
the signature
- they are duplicate
- make harder to see code between all the docstrings
- they are duplicate (copy pasted) and thus at times incorrect/not
  adjusted for the subclass (e.g. nomultiplex one was claiming to get a singleton
for a shared ssh connection

pydoc and pycharm are just fine without docstrings in subclasses -- they pick
them up from the base class, so we do not even need an explicit
@borrowdoc I thought we would need

Closes #6536 

#### 📝 Documentation
- Remove duplicate docstrings in subclasses of sshconnector

### DIFF
--- a/datalad/support/sshconnector.py
+++ b/datalad/support/sshconnector.py
@@ -151,12 +151,45 @@ class BaseSSHConnection(object):
         self._remote_props = {}
 
     def __call__(self, cmd, options=None, stdin=None, log_output=True):
+        """Executes a command on the remote.
+
+        It is the callers responsibility to properly quote commands
+        for remote execution (e.g. filename with spaces of other special
+        characters).
+
+        Parameters
+        ----------
+        cmd: str
+          command to run on the remote
+        options : list of str, optional
+          Additional options to pass to the `-o` flag of `ssh`. Note: Many
+          (probably most) of the available configuration options should not be
+          set here because they can critically change the properties of the
+          connection. This exists to allow options like SendEnv to be set.
+
+        Returns
+        -------
+        tuple of str
+          stdout, stderr of the command run.
+        """
         raise NotImplementedError
 
     def open(self):
+        """Opens the connection.
+
+        Returns
+        -------
+        bool
+          To return True if connection establishes a control socket successfully.
+          Return False otherwise
+        """
+
         raise NotImplementedError
 
     def close(self):
+        """Closes the connection.
+        """
+
         raise NotImplementedError
 
     @property
@@ -358,27 +391,7 @@ class NoMultiplexSSHConnection(BaseSSHConnection):
     as soon as the process end.
     """
     def __call__(self, cmd, options=None, stdin=None, log_output=True):
-        """Executes a command on the remote.
 
-        It is the callers responsibility to properly quote commands
-        for remote execution (e.g. filename with spaces of other special
-        characters).
-
-        Parameters
-        ----------
-        cmd: str
-          command to run on the remote
-        options : list of str, optional
-          Additional options to pass to the `-o` flag of `ssh`. Note: Many
-          (probably most) of the available configuration options should not be
-          set here because they can critically change the properties of the
-          connection. This exists to allow options like SendEnv to be set.
-
-        Returns
-        -------
-        tuple of str
-          stdout, stderr of the command run.
-        """
         # there is no dedicated "open" step, put all args together
         ssh_cmd = [self.ssh_executable] + self._ssh_open_args + self._ssh_args
         return self._exec_ssh(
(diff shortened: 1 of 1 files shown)


mih:
TBH in general I don't like this at all. If the documentation is bad, it should be fixed, not removed. Requiring additional layers of technology (like IDEs) to make code minimally accessible is bad too. If an IDE provides features to superimpose documentation from elsewhere, but "make harder to see code between all the docstrings" I think the blame is on the IDE, not the code.

In this particular case, I don't mind any changes. I don't think this code will survive for much longer.

yarikoptic:
> If the documentation is bad, it should be fixed, not removed.

no documentation is removed -- only duplicate documentation is removed.  And I think that duplication (in code or documentation) is a source of many bugs/issues.   I think that a practice of copy pasting docstrings into derived classes is detrimental.  That is why we even have those `@borrowdocs` and `@borrowkwargs` to remove the need for copy/pasting and maintaining redundant copies of the text. It seems that we might want to establish (after possibly some research into other projects procedures) some generic approach to "what do we do for docstrings in derived classes methods" to harmonize code-base in that regard going forward.

> Requiring additional layers of technology (like IDEs) to make code minimally accessible is bad too ...

no code is modified and if anything -- only made more accessible. FWIW, I would say that removing documentation clutter makes code more accessible since I don't need to scroll as much to actually see the code.  I would not blame IDEs for lacking some super-features of docstrings invisibility etc.

bpoldrack:
Just to be clear: What I am saying WRT to this PR is: The move seems right to me. Those docstrings are describing the super-class' intend for those (abstract) methods. But: The subclasses shouldn't "inherit" the docstring. They should have their own (doesn't need to be this PR), that would describe what they do (the concrete implementation in that case). If the implementation is identical it shouldn't exist twice. If it's not then having the same docstring simply means not taking documentation seriously.

yarikoptic:
> If the implementation is identical it shouldn't exist twice. If it's not then having the same docstring simply means not taking documentation seriously.

I would not agree generally -- docstrings should not generally describe *how* they implement desired functionality -- and *how* is the implementation in subclasses. *what* they do should be descriptive enough at the base/abstract class - that is why there is a point of having subclasses and abstract methods that the choice of the subclass to obtain a generic operation should not matter, thus documentation most often is ok to stay identical.

note: for borrowing docstrings on specific Parameters (whenever derived interface e.g. extends with new kwargs and thus does need a docstring) - we have `@borrowkwargs` .

yarikoptic:
Just a little more although largely reiterating previously stated:

> If there's a need for the method to be implemented or overwritten in a subclass to begin with, than something about it is special and should be documented.

Disagree strongly --  base class which defines interface should describe the desired effect/returned value and parameters. Ideally a specific implementation (in subclass) should have no side-effects worth mentioning, and just fulfill that interface defined/documented in base class.  Examples could be multitude and go to the basics of OOP.
Indeed subclass implementation could have extra options added which need to be documented and thus warranting method's docstring. 

Side-effects (eg. persisting ControlPath socket) might better be described in the docstring of the entire class since likely multiple methods would be effected and it is better to provide higher level overview of the class specifics in the class docstring than overloading docs of every method which potentially deals with those side-effects.

Meanwhile, updated this PR to account for ff1d3e473c5b407179188666aef9c22119e56268 where `sh_quote` was removed from being recommended.

yarikoptic:
> For this reason I never found that borrow_doc approach very helpful or convincing.

yeah - `@borrow_doc` is boring and probably not needed. But `@borrow_kwargs` (currently no longer used at all) could come really handy in subclasses extending the signature of the function to avoid re-listing all args/kwargs or just saying `*kwargs: all passed to paren'ts blah` and thus making it bad DX.

ok, since @mih did not have major objections :

> In this particular case, I don't mind any changes. I don't think this code will survive for much longer.

and @bpoldrack generally agreed:

> Just to be clear: What I am saying WRT to this PR is: The move seems right to me.

let's proceed!

Issue #6551, opened 2022-03-15
different git version when switching between datalad versions


djarecka:
### What is the problem?

When I use my older `datalad` conda environment on a cluster (openmind) it changes my git to a higher version (2.3), but when I use the latest version of `datalad` (0.15.6) it doesn't do it. Was this change done on purpose?

### What steps will reproduce the problem?

This is what I run:
```
[19:09][194.73][-36%]djarecka@openmind7:~$ git --version
git version 1.8.3.1
[19:10][194.88][-35%]djarecka@openmind7:~$ conda activate datalad
[19:10][194.89][-35%]djarecka@openmind7:~$ git --version
git version 2.30.0
[19:10][194.81][-35%]djarecka@openmind7:~$ datalad --version
datalad 0.15.2
[19:10][194.75][-35%]djarecka@openmind7:~$ conda activate datalad_0.15.6
[19:10][194.97][-35%]djarecka@openmind7:~$ git --version
git version 1.8.3.1
```

### Datalad information

Ha! I can't run any datalad commands: getting `datalad.runner.exception.CommandError: CommandError: 'git --git-dir= config -z -l --show-origin' failed with exitcode 129 [err: 'error: unknown option 'show-origin' `, but I just run `pip install datalad`, so should be `0.15.6`.

<details>
<summary> But I can run `datalad wtf` with my old environment </summary>

[19:24][196.74][-38%]djarecka@openmind7:~$ datalad --version
datalad 0.15.2
[19:25][196.39][-38%]djarecka@openmind7:~$ datalad wtf
# WTF
## credentials 
  - keyring: 
    - active_backends: 
      - PlaintextKeyring with no encyption v.1.0 at /home/djarecka/.local/share/python_keyring/keyring_pass.cfg
    - config_file: /home/djarecka/.config/python_keyring/keyringrc.cfg
    - data_root: /home/djarecka/.local/share/python_keyring
## datalad 
  - full_version: 0.15.2
  - version: 0.15.2
## dependencies 
  - annexremote: 1.5.0
  - appdirs: 1.4.4
  - boto: 2.49.0
  - cmd:7z: 16.02
  - cmd:annex: 8.20201129-geb388e6
  - cmd:bundled-git: UNKNOWN
  - cmd:git: 2.30.0
  - cmd:system-git: 2.30.0
  - cmd:system-ssh: 8.4p1
  - humanize: 3.2.0
  - iso8601: 0.1.13
  - keyring: 22.0.1
[...]

PR #6553, opened 2022-03-16
Enforce system ssh client usage on Windows

### LINKED ISSUE #5829: Cannot clone on Windows via ssh - no password prompt, just errors out
might relate #5031 

image

you can see that straight `git clone` works, but we fail - not asking for any password, just saying "kaboom".

<details>
<summary>works ok on linux, besides interference with progress bar</summary> 

```shell
$> datalad clone  smaug.datalad.org:/mnt/btrfs/datasets/datalad/ci/git-annex
yoh@smaug.datalad.org's password:                     | 0.00/1.00 [00:00<?, ? Candidate locations/s]
install(ok): /home/test3/git-annex (dataset)                                                        

$> datalad --version
datalad 0.14.6
```
</details>

and here is with -l debug (and displaying PID just in case):

image

and it seems it is our ssh connector which sets those up

```shell
$> git grep -3 PasswordAuth
datalad/support/sshconnector.py-        self._ssh_open_args += [
datalad/support/sshconnector.py-            # we presently do not support any interactive authentication
datalad/support/sshconnector.py-            # at the time of process execution
datalad/support/sshconnector.py:            '-o', 'PasswordAuthentication=no',
datalad/support/sshconnector.py-            '-o', 'KbdInteractiveAuthentication=no',
datalad/support/sshconnector.py-        ]
datalad/support/sshconnector.py-
```

in effect disabling interactive prompting... didn't look in greater detail but I wonder how it works for any windows mortal user

<details>
<summary>
edit1: disabling those settings we do **only** for `NoMultiplexSSHConnection` (on windows I guess)
</summary>

still results in a failed clone, just differently:
image
</details>

<details>
<summary>edit 2: native git submodule update --init  also fails!</summary> 

image

</details>

so unfortunately there is more to the story :-/

christian-monch:
This PR fixes #5829 

This PR enforces the use of the system windows ssh client (OpenSSH for Windows) in `sshconnector.py`. The reason for this restriction is that other ssh clients, namely the ssh client that is bundled with git for windows, do not properly work when called by git (see the discussions in issue #5829).

The PR introduces the new environment variable `DATALAD_WINDOWS_SSH_CLIENT` that can be set to inform `datalad` about an alternative location of the system ssh client.

Possible future improvements:
- We do not currently verify that the ssh-client is indeed the system windows ssh-client.
- This PR requires the installation of OpenSSH for Windows in order to use datalad with ssh on windows. We might want to investigate why other ssh clients do not properly work and fix the root cause.

#### 🐛 Bug Fixes
- Description... Fixes #5829 by allowing to explicitly specify a compatible ssh client-executable
- Introduces the configuration key `datalad.ssh.executable`. This key allows specifying an ssh-client executable that should be used by datalad to establish ssh-connections. The default value is `ssh` unless on a Windows system where `$WINDIR\System32\OpenSSH\ssh.exe` exists. In this case, the value defaults to `$WINDIR\System32\OpenSSH\ssh.exe`.

### DIFF
--- a/datalad/interface/common_cfg.py
+++ b/datalad/interface/common_cfg.py
@@ -35,9 +35,11 @@ def get_default_ssh():
     from datalad.utils import on_windows
     from pathlib import Path
 
-    windows_openssh_path = r'C:\Windows\System32\OpenSSH\ssh.exe'
-    if on_windows and Path(windows_openssh_path).exists():
-        return windows_openssh_path
+    if on_windows:
+        windows_openssh_path = \
+            environ.get("WINDIR", r"C:\Windows") + r"\System32\OpenSSH\ssh.exe"
+        if Path(windows_openssh_path).exists():
+            return windows_openssh_path
     return "ssh"
 
 
--- a/datalad/interface/common_cfg.py
+++ b/datalad/interface/common_cfg.py
@@ -579,7 +579,7 @@ definitions = {
         'ui': ('question', {
             'title': "Name of ssh executable for 'datalad sshrun'",
             'text': "Specifies the name of the ssh-client executable that"
-                    "'datalad sshrun' will use. This might be an absolute "
+                    "datalad will use. This might be an absolute "
                     "path. On Windows systems it is currently by default set "
                     "to point to the ssh executable of OpenSSH for Windows, "
                     "if OpenSSH for Windows is installed. On other systems it "
--- a/datalad/interface/common_cfg.py
+++ b/datalad/interface/common_cfg.py
@@ -30,6 +30,17 @@ from datalad.utils import on_windows
 
 dirs = AppDirs("datalad", "datalad.org")
 
+
+def get_default_ssh():
+    from datalad.utils import on_windows
+    from pathlib import Path
+
+    windows_openssh_path = r'C:\Windows\System32\OpenSSH\ssh.exe'
+    if on_windows and Path(windows_openssh_path).exists():
+        return windows_openssh_path
+    return "ssh"
+
+
 subst_rule_docs = """\
 A substitution specification is a string with a match and substitution
 expression, each following Python's regular expression syntax. Both expressions
@@ -564,16 +575,18 @@ definitions = {
         'default': 'warning',
 
     },
-    'datalad.windows.ssh.client': {
+    'datalad.ssh.executable': {
         'ui': ('question', {
-            'title': "Location of OpenSSH for Windows' ssh.exe",
-            'text': "To use git with ssh on Windows the Windows system ssh"
-                    "client, i.e. OpenSSH for Windows, has to be installed. "
-                    "If it is not installed in the default location, its "
-                    "location can be set in this variable."}),
+            'title': "Name of ssh executable for 'datalad sshrun'",
+            'text': "Specifies the name of the ssh-client executable that"
+                    "'datalad sshrun' will use. This might be an absolute "
+                    "path. On Windows systems it is currently be default set "
+                    "to point to the ssh executable of OpenSSH for Windows, "
+                    "if OpenSSH for Windows is installed. On other systems it "
+                    "defaults to 'ssh'."}),
         'destination': 'global',
         'type': EnsureStr(),
-        'default': r'C:\Windows\System32\OpenSSH\ssh.exe',
+        'default_fn': get_default_ssh,
     }
 }
 
--- a/datalad/interface/common_cfg.py
+++ b/datalad/interface/common_cfg.py
@@ -580,7 +580,7 @@ definitions = {
             'title': "Name of ssh executable for 'datalad sshrun'",
             'text': "Specifies the name of the ssh-client executable that"
                     "'datalad sshrun' will use. This might be an absolute "
-                    "path. On Windows systems it is currently be default set "
+                    "path. On Windows systems it is currently by default set "
                     "to point to the ssh executable of OpenSSH for Windows, "
                     "if OpenSSH for Windows is installed. On other systems it "
                     "defaults to 'ssh'."}),
--- a/datalad/interface/common_cfg.py
+++ b/datalad/interface/common_cfg.py
@@ -564,6 +564,17 @@ definitions = {
         'default': 'warning',
 
     },
+    'datalad.windows.ssh.client': {
+        'ui': ('question', {
+            'title': "Location of OpenSSH for Windows' ssh.exe",
+            'text': "To use git with ssh on Windows the Windows system ssh"
+                    "client, i.e. OpenSSH for Windows, has to be installed. "
+                    "If it is not installed in the default location, its "
+                    "location can be set in this variable."}),
+        'destination': 'global',
+        'type': EnsureStr(),
+        'default': r'C:\Windows\System32\OpenSSH\ssh.exe',
+    }
 }
 
 
--- a/datalad/support/sshconnector.py
+++ b/datalad/support/sshconnector.py
@@ -167,12 +167,9 @@ class BaseSSHConnection(object):
         """determine which ssh client should be used.
 
         On Windows OpenSSH for Windows is required. It is searched in its
-        default location. This can be overriden by setting the datalad
+        default location. This can be overridden by setting the datalad
         config variable 'datalad.windows.ssh.client' to the path of the
         system ssh client.
-
-        If a location is stored in the config variable, it takes precedence
-        over the default location.
         """
         if self._ssh_command:
             return self._ssh_command
@@ -183,9 +180,7 @@ class BaseSSHConnection(object):
         if on_windows:
             from datalad import cfg
 
-            cmd = cfg.get(self.cfg_datalad_windows_ssh_client)
-            if cmd is None:
-                cmd = r"C:\Windows\System32\OpenSSH\ssh.exe"
+            cmd = cfg.obtain(self.cfg_datalad_windows_ssh_client)
             if not Path(cmd).exists():
                 lgr.error(
                     f"SSH client ({cmd}) was not found.\nPlease "
--- a/datalad/support/sshconnector.py
+++ b/datalad/support/sshconnector.py
@@ -163,7 +163,7 @@ class BaseSSHConnection(object):
         raise NotImplementedError
 
     @property
-    def ssh_command(self):
+    def ssh_executable(self):
         """determine which ssh client should be used.
 
         On Windows OpenSSH for Windows is required. It is searched in its
--- a/datalad/support/sshconnector.py
+++ b/datalad/support/sshconnector.py
@@ -15,6 +15,7 @@ git calls to a ssh remote without the need to reauthenticate.
 
 import fasteners
 import os
+import platform
 import logging
 from socket import gethostname
 from hashlib import md5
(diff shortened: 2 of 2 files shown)


datalad/support/sshconnector.py
```diff
@@ -97,6 +98,9 @@ def get_connection_hash(hostname, port='', username='', identity_file='',
 class BaseSSHConnection(object):
     """Representation of an SSH connection.
     """
+
+    env_name_datalad_windows_ssh_client = "DATALAD_WINDOWS_SSH_CLIENT"
```
bpoldrack, line 102: Is there a reason to not use the standard config mechanism to define `datalad.windows.ssh-client` instead of limiting it to env var?

christian-monch: No, I am just not used to datalad's standard config mechanism ;-), and there were a few other places that used env-vars.

Will change it to config. Good idea!

mih (commented):
Awesome!

Quick comment prior review: please have a look at `cfg.obtain()` and `datalad/interface/common_cfg.py`. When the new switch is placed in there, it will be nicely presented to `configuration` users. Thx!

yarikoptic:
> The reason for this restriction is that other ssh clients, namely the ssh client that is bundled with git for windows, do not properly work when called by git (see the discussions in issue #5829).

was this issue raised with "git for windows" developers? IMHO we should aim to have it addressed properly instead of relying on using system wide git which is known on other systems to get outdated (see e.g. recent #6551) .

mih (commented):
Thanks for switching to a config def!

I wonder whether it should be `datalad.ssh.executable` and be `None` on all platforms by default, and `r'C:\Windows\System32\OpenSSH\ssh.exe'` on windows?

We do have a bunch of `ssh` related variables, and only `datalad.save.windows-compat-warning` otherwise. Maybe bundling it with SSH is more accessible?

datalad/interface/common_cfg.py
```diff
@@ -564,6 +564,17 @@
...
+    'datalad.windows.ssh.client': {
+        'ui': ('question', {
+            'title': "Location of OpenSSH for Windows' ssh.exe",
+            'text': "To use git with ssh on Windows the Windows system ssh"
+                    "client, i.e. OpenSSH for Windows, has to be installed. "
+                    "If it is not installed in the default location, its "
+                    "location can be set in this variable."}),
+        'destination': 'global',
+        'type': EnsureStr(),
+        'default': r'C:\Windows\System32\OpenSSH\ssh.exe',
```
yarikoptic, line 576: why bother making it windows-specific?
Could be just `datalad.ssh.client` defaulting to `"ssh"` on non-windows and this specific path on windows.  If abs path is given, but points to non-existing file, we could even be nice and inform on which paths do have that executable within PATH, something akin to what we have in https://github.com/datalad/datalad/blob/master/tools/which.py ?

note: we do not yet have any other `datalad.windows.` config variable.

datalad/support/sshconnector.py
```diff
@@ -158,6 +162,40 @@ def open(self):
...
+    def ssh_command(self):
+        """determine which ssh client should be used.
+
+        On Windows OpenSSH for Windows is required. It is searched in its
+        default location. This can be overridden by setting the datalad
+        config variable 'datalad.windows.ssh.client' to the path of the
+        system ssh client.
+        """
+        if self._ssh_command:
+            return self._ssh_command
```
yarikoptic, line 175: see https://github.com/datalad/datalad/issues/6556 -- might be time to start using `@cached_property` but we better centralize its import (needs try/except for python < 3.8) e.g. in `datalad.utils`?

christian-monch:
> was this issue raised with "git for windows" developers? IMHO we should aim to have it addressed properly instead of relying on using system wide git which is known on other systems to get outdated (see e.g. recent https://github.com/datalad/datalad/issues/6551) .

It was not raised. But I agree we should raise it with git-windows developer. I can do that.

Then the question is what we do now: I see four options:

- Do nothing (not good).
- No changes in the code, We document that on Windows you should not (for now) use the git-bundled ssh. This documentation can be changed/removed later.
- We add a general `datalad.ssh.client`-config key. We document that it should not point to the git-bundled ssh on Windows. We can change/remove the documentation later.
- We add a temporary `datalad.windows.ssh.client`-config key, which would be in effect until the root cause is resolved? We might be able to remove the key later, but that could require deprecation.

Did I forget an option?

mih:
I think it should be option 3.

But I thinking using the term `executable` in the name is better. We already do client "variant" selection inside. Just "client" is to ambiguous IMHO.

The fact that it could be an absolute path or relative is not a problem with semantics IMHO.

I also disagree that "it should no point to any other". There are good reasons for choosing others. The problem being fixed here can be solved by users with other means and bundled ssh version can have desired features.

christian-monch:
> I think it should be option 3.
> 
> But I thinking using the term `executable` in the name is better. We already do client "variant" selection inside. Just "client" is to ambiguous IMHO.

Will implement option 3 with `datalad.ssh.executale`

> The fact that it could be an absolute path or relative is not a problem with semantics IMHO.

I agree


> I also disagree that "it should no point to any other". There are good reasons for choosing others. The problem being fixed here can be solved by users with other means and bundled ssh version can have desired features.

Generally, I agree. But I would nitpickingly ;-) add: the "should not point ..." part refers to the initial intention to provide a Windows-specific work-around. On other systems, we would rely on the path-mechanism.

So I will implement the agreed-upon config key and we add a "demand"-driven default value for windows and/or documentation of the error case.

christian-monch:
> > The reason for this restriction is that other ssh clients, namely the ssh client that is bundled with git for windows, do not properly work when called by git (see the discussions in issue #5829).
> 
> was this issue raised with "git for windows" developers? IMHO we should aim to have it addressed properly instead of relying on using system wide git which is known on other systems to get outdated (see e.g. recent #6551) .

On further investigation, I am not sure that it is a git-for-windows problem. The command `git clone some.host.de:/path/to/repo localrepo` works in Powershell (PS) with redirected stderr. It does not work from within a python interpreter when stderr is redirected. The following snippet exhibits the same error (no password-prompt and faulty authenitcation):

```
import subprocess
subprocess.run(["git", "clone"; "some.host.de:/path/to/repo", "localrepo"], stderr=subprocess.PIPE)
``` 

So, at least it does not seem to be related to datalad. I will look closer into pythons subprocess and maybe give python `3.10` a try, before turning to the git-for-windows folks.

datalad/interface/common_cfg.py
christian-monch: I introduced a generic "datalad.ssh.executable" configuration. On non-Windows systems it defaults to `ssh`. On Windows systems it defaults to `\Windows\System32\OpenSSH\ssh.exe`, if that path exists, otherwise also to `ssh`.

datalad/support/sshconnector.py
christian-monch: I would do that in a different PR!?

datalad/support/sshconnector.py
```diff
@@ -158,6 +162,40 @@ def open(self):
...
+                    f"SSH client ({cmd}) was not found.\nPlease "
+                    f"install OpenSSH for Windows. If you have installed "
+                    f"OpenSSH for Windows in another location than the default "
+                    f"location, please set the datalad configuration key "
+                    f"'{self.cfg_datalad_windows_ssh_client}' to the "
+                    f"location of the installed 'ssh.exe' file.")
+                raise RuntimeError("ssh-executable not found")
+            self._ssh_command = cmd
+        else:
+            self._ssh_command = "ssh"
```
mih, line 195: This is now duplicating the logic in the cfg default setup. Moreover, the switching and redeclaration of a default should not be platform-specific here, for the same reason.

mih (changes requested):
Thx! I have tried this a windows box and it works nicely!

I nevertheless left a bunch of comments, because I think the PR still contains a few traces of the exploration for a fix and should be consolidated on the final solution.

mih:
I just noticed that the changelog also needs an update. I think it is sufficient to present this as an improvement (ssh exec configurability) that also happens to address #5829

datalad/support/sshconnector.py
christian-monch: Thx, I overlooked that in the torrent of activities and changes

datalad/support/sshconnector.py
```diff
@@ -158,6 +162,40 @@ def open(self):
...
 
+    @property
+    def ssh_command(self):
+        """determine which ssh client should be used.
+
+        On Windows OpenSSH for Windows is required. It is searched in its
+        default location. This can be overridden by setting the datalad
+        config variable 'datalad.windows.ssh.client' to the path of the
+        system ssh client.
+        """
```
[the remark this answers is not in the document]
christian-monch: Agreed. I removed it

datalad/support/sshconnector.py
```diff
@@ -97,6 +97,9 @@ def get_connection_hash(hostname, port='', username='', identity_file='',
 class BaseSSHConnection(object):
     """Representation of an SSH connection.
     """
+
+    cfg_datalad_windows_ssh_client = "datalad.ssh.executable"
```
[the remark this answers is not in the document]
christian-monch: Thx. Fixed by using the string in the remaining single usage.

datalad/interface/common_cfg.py
```diff
@@ -564,6 +575,19 @@
         'default': 'warning',
 
     },
+    'datalad.ssh.executable': {
+        'ui': ('question', {
+            'title': "Name of ssh executable for 'datalad sshrun'",
+            'text': "Specifies the name of the ssh-client executable that"
+                    "'datalad sshrun' will use. This might be an absolute "
+                    "path. On Windows systems it is currently be default set "
```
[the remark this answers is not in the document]
christian-monch: Thx, fixed

datalad/interface/common_cfg.py
```diff
@@ -564,6 +575,19 @@
...
+                    "to point to the ssh executable of OpenSSH for Windows, "
```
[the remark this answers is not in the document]
christian-monch: I agree. Will add a link to the git4win issue.

datalad/interface/common_cfg.py
```diff
@@ -30,6 +30,17 @@
...
 dirs = AppDirs("datalad", "datalad.org")
 
+
+def get_default_ssh():
+    from datalad.utils import on_windows
+    from pathlib import Path
+
+    windows_openssh_path = r'C:\Windows\System32\OpenSSH\ssh.exe'
+    if on_windows and Path(windows_openssh_path).exists():
+        return windows_openssh_path
```
christian-monch, line 40: This is related to the [git-for-windows issue #3746](https://github.com/git-for-windows/git/issues/3748). Once this issue is resolved. this function can be removed and `"ssh"` can be used as default value in the definition of the config key `datalad.ssh.executable`.

christian-monch:
Changelog is addressed.

christian-monch:
Looking at https://github.com/git-for-windows/git/issues/3748#issuecomment-1072243174, it does not seem that the problem with the ssh-client that is bundled with git-for-windows will be solved any time soon.

mih (approved):
Thx!

Issue #3126, opened 2019-01-23
Credentials should be tested before being stored


effigies:
#### What is the problem?

When creating a sibling on GitHub for the first time, the username and password provided are stored. If someone has 2FA enabled, then they actually need a token, and not their password. However, the username and password are now stored in a keychain. The default behavior on the next round is not to ask again, but to attempt to reuse the bad credentials.

#### What steps will reproduce the problem?

Start from a fresh environment with no stored credentials.

```Shell
$ datalad create newrepo
$ pushd newrepo
$ datalad create-sibling-github <github-url>
You need to authenticate with 'github' credentials. https://github.com/login provides information on how to gain access
user: USERNAME

password: 
password (repeat): 
[ERROR  ] 401 {'message': 'Must specify two-factor authentication OTP code.', 'documentation_url': 'https://developer.github.com/v3/auth#working-with-two-factor-authentication'} [Requester.py:__check:275] (GithubException)
$ datalad create-sibling-github <github-url>
[ERROR  ] 401 {'message': 'Must specify two-factor authentication OTP code.', 'documentation_url': 'https://developer.github.com/v3/auth#working-with-two-factor-authentication'} [Requester.py:__check:275] (GithubException)
```

#### What version of DataLad are you using (run `datalad --version`)? On what operating system (consider running `datalad wtf`)?

```
# WTF
- path: /Users/markiewicz/Projects/crn/fitlins-test-ds000003
- type: dataset
## configuration <SENSITIVE, report disabled by configuration>
## datalad 
  - full_version: 0.11.1
  - version: 0.11.1
## dataset 
  - metadata: <SENSITIVE, report disabled by configuration>
  - path: /Users/markiewicz/Projects/crn/fitlins-test-ds000003
  - repo: AnnexRepo
## dependencies 
  - appdirs: 1.4.3
  - boto: 2.49.0
  - cmd:annex: 7.20181212-g2f8c5bd4c
  - cmd:git: 2.18.0
  - cmd:system-git: 2.19.1
  - cmd:system-ssh: 7.9p1
  - git: 2.1.11
  - gitdb: 2.0.5
  - humanize: 0.5.1
  - iso8601: 0.1.12
  - keyring: 17.0.0
  - keyrings.alt: 3.1.1
  - msgpack: 0.6.0
  - requests: 2.21.0
  - six: 1.12.0
  - tqdm: 4.28.1
  - wrapt: 1.10.11
## environment 
[...]

PR #6555, opened 2022-03-16
New credentials() command and CredentialManager

### LINKED ISSUE #396: provide user accessible (API) interface to credentials

### LINKED ISSUE #2721: wishlist: Disable repeated entry of "hidden" values

### LINKED ISSUE #6519: Redesign credential system
`[The top post is repeatedly updated in response to comments below]`

### Problem
The current credential subsystem is tightly coupled to "providers" and consequently mostly targeting "downloads". However, since its inception other uses of credentials were needed (API tokens, etc), and support for those was bolted onto this system, yielding the present state.

One key problem is that we do not know which credentials are know/necessary until a particular credential request is made. This is a limitation of the `keyring` system that underlies our credential setup, and it used as its main (only) data store. Not knowing which credentials exist (because of a lack of identifiers), prevent associating additional information with a credential, such as a priority (#322), and complicates/prevents credential maintenance (update, removal).

The "provider" concept amends the credential store with a configuration file that is used to associated additional information with a credential (but actually in reverse, it associates a credential with a download scenario). I believe we could improve the design by making these systems more or fully separate. A credential can be used in more than one scenario. It requires a dedicated API for query and manipulation (#396).

### Proposal

It would make sense to me to start using the configuration mechanism as the main registry for credentials too. It would provide the complete list of credentials known to datalad. It would be the place to annotate credentials with additional aspects, such as scope (e.g. URL regex), priority, expiration date, and linkage to credential storage subsystems.

http://docs.datalad.org/en/stable/design/credentials.html already defines the config variable setup

   `datalad.credential.<name>.<component>`

only that `component` would no longer be limited to `token`, `username`, `password`, but could be amended with `priority`, `expiration`, etc. as needed.

*On top* of what is defined in the configuration, a management API could list credential names and types known to the provider subsystem. On accessing any unknown credential in any backend, the management layer would leave a trace about a new credential in the config, such that any custom credential becomes accessible over time.

Looking at actual credentials in my local credential store, here are a few properties to consider:

- description
- type (token, password, ...)
- secret (the piece that needs protection)
- username/identity (whatever would be necessary in addition to the secret)
- usage cateory (login, ...)
- date created/synced
- realm (some label to identify a matching credential)
- times-used, last-used
- action-url (location to use it with)
- original-url (location/service that provided it)

Beyond the basic secret, all other properties could be useful for making decisions on what credential is applicable or matches in a particular context. Only a user of the credential system can know what makes sense in the general case. The credential system should ease flexible queries, but make no decisions.

The ability for users to indicate which credential should be used will be a crucial feature. We need `name` from  `datalad.credential.<name>.<component>` already to bind properties from secret storage and config mechanism, so this will be a given identification method. It is largely identical to the value semantics of the various `--credential` arguments, and is similar to what secretstorage calls "label".

Moreover, it makes sense to support other (generic) specifications like a realm. This could conceptually replace the direct specification of a single credential label in a provider configuration. Something like `s3_hcp` could be a realm that several credentials are associated with (think: I know multiple credential sets, maybe one read-only, and one for admin-access, different gihub tokens, etc.). This would make it possible to better separate the user-defined namespace of credential labels from datalad-provider defined credential scopes. A default for a `--credential` argument of a command that interacts with the github API could be `realm:github-api`, and would automatically find a suitable credential. If there is only one, everything is fine. If there are multiple ones, a user can identify a precise one via a label.

### API

The primary (possibly only) API to interact with the credential system is a dedicated `credentials` command (plural name to follow established pattern: `subdatasets`, `containers-run`, `siblings`, `osf_credentials`). It implements the following operations:

- `set <name> <spec>`
- `get <name> [type=...]`
- `remove <name> [type=...]`
- `query [<spec>]`

where `<name>` is a unique identifier/label, and `<spec>` are key-value pairs that annotate a credential. For `set` the `<spec>` are a complete list of all credential properties, for `query` it is a list that any matching credential must have. For maximum usability, multiple formats for specifying `<spec>` will be supported (Python iterable, k/v-pairs like `configuration`, JSON-formatted string). Property names must be valid Git-config variable names. They are used for the `<component>` slot in the `datalad.credential.<name>.<component>` naming convention, whereas `<name>` is used as `<name>`.

The API does not define a finite list of credential types (token, user-pass, etc). The type is merely a property. For reasons of backward compatibility only, `get` supports a type-hint specification using the same syntax as `<spec>` (see below).

The number and identify of properties is not limited.

Each credential has a single secret. Its value is not constrained to a particular nature. An additional credential property could (optionally) signal different types of encodings (e.g., a multi-part credential encoded as a JSON-array/object). If a secret is not part of the `<spec>` given to `set`, interactive secret entry is performed. Secret always go into whatever the available/active secret store is for a particular installation.

`remove` is used to wipe out an entire credential with all its properties.

In order to modify or remove individual properties of a credential, `set` is used while specifying only the key, and no value (in case of removing a property), or an alternative value (for a modification of a property). In Python, "no value" can be `None` (as a non-string value), for a CLI call the equivalent could be `:key`  (followed by no `=` and no value, akin to deleting a remote refspec with git-push). An alternative CLI notation could be just a key name (no prefix), but this might cause more accidents (e.g. unintentional whitespace between key and `=<value>` etc.).

##### Set a credential
`credentials set mycred username=mike` will ask for secret interactively.

`credentials set mycred username=mike secret=precious` will not ask for secret interactively.

`-c datalad.credential.mycred.secret=precious credentials set mycred username=mike` will also not ask for secret interactively.

##### Update a credential

`credentials set mycred :secret` will delete a previously defined secret, and because afterwards there is none defined, we ask for a new one interactively.

`credentials set mycred secret=update` will replace a previously defined secret, non-interactively.

`-c datalad.credential.mycred.secret=precious credentials set mycred :secret` will also delete a previously defined secret, but because there is one provided via cfg, will not ask interactively.

##### Remove a credential entirely

`credentials remove mycred`

##### Find credential candidates

`credential query 'realm=http://api.github.com'` report all credentials for the given realm. Credentials are yielded as results, with all their properties. Client code can decide which one to use, if there are many, e.g. by `last-used` timestamp, etc.

### Implementation details

Given that credentials are typically personal information, non-secret information is put into the `global` configuration scope (i.e. user-config), whenever no alternative implementation is available. Secrets are always stored in and retrieved from a secret store (via `keyring`). On systems where alternative implementations are available (e.g. `secretstorage` in the future), secrets *and* additional properties can be placed together in a secret store.

`get` and `remove` can be given a credential `type` hint, in order to be able to retrieve and delete credentials stored by the previous credential system, where type-information is only available within the context of a credential request.

mih:
This is a new subsystem aiming to factor out the management of credentials from the downloaders/providers concept into a
standalone system.
    
Importantly, the management of downloaders/providers credentials is supported by the new system, albeit requiring "type hints" for some operations, because downloader/provider credentials are only fully discoverable in the context of a provider (identified by the context of a request).
    
In contrast, the new system considers credentials as a standalone, request independent entity.
    
A credential is discoverable for any number of abitrarily named properties, but is typically (and right now exclusively) identified
by a name.
        
The system of provided operations/actions enables the implementation
of a confirm-function-function-before-storage pattern (see next commit,
ping datalad/datalad#3126).
    
### Demo

The second commit switches `create-sibling-ghlike` to new credential system.
    
The switch also makes the storage of a newly entered credential conditional on a successful authorization, in the spirit of
datalad/datalad#3126.
    
Moreover, stored credentials now contain a `realm` property that identified the API endpoint. This makes it possible to identify
candidates of suitable credentials without having to specific their name, similar to a request context url used by the old
providers setup. Such a lookup is now also implemented and makes the specification of a credential obsolete in most standard scenarios.

- A new credential management system is introduced that enables storage and query of credentials with any number of properties associated with a secret. These properties are stored as regular configuration items, following the scheme `datalad.credential.<name>.<property>`. The special property `secret` lives in a keystore, but can be overriden using the normal configuration mechanisms. The new system continues to support the previous credential storage setup. Fixes #6519
- A new `credentials` command enables query, modification and storage of credentials. Legacy credentials are also supported, but may require the specification of a `type`, such as (`token`, or `user_password`) to be discoverable. Fixes #396
- Two new configuration settings enable controlling how the interactive entry of credential secrets is conducted for the new credential manager: `datalad.credentials.repeat-secret-entry` can be used to turn off the default double-entry of secrets, and `datalad.credentials.hidden-secret-entry` can turn off the default hidden entry of secrets. Fixes #2721

### DIFF
--- a/datalad/credman.py
+++ b/datalad/credman.py
@@ -1,0 +1,710 @@
+# emacs: -*- mode: python; py-indent-offset: 4; tab-width: 4; indent-tabs-mode: nil -*-
+# ex: set sts=4 ts=4 sw=4 noet:
+# ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+#
+#   See LICENSE file distributed along with the datalad_osf package for the
+#   copyright and license terms.
+#
+# ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+"""Credential management and query"""
+
+__docformat__ = 'restructuredtext'
+
+__all__ = ['CredentialManager']
+
+from datetime import datetime
+import logging
+import re
+
+import datalad
+from datalad.support.exceptions import (
+    CapturedException,
+    CommandError,
+)
+from datalad.ui import ui
+
+lgr = logging.getLogger('datalad.credman')
+
+
+class CredentialManager(object):
+    """Facility to get, set, remove and query credentials.
+
+    A credential in this context is a set of properties (key-value pairs)
+    associated with exactly one secret.
+
+    At present, the only backend for secret storage is the Python keyring
+    package, as interfaced via a custom DataLad wrapper. Store for credential
+    properties is implemented using DataLad's (i.e. Git's) configuration
+    system. All properties are stored in the `global` (i.e., user) scope
+    under configuration items following the pattern::
+
+      datalad.credential.<name>.<property>
+
+    where ``<name>`` is a credential name/identifier, and ``<property>`` is an
+    arbitrarily named credential property, whose name must follow the
+    git-config syntax for variable names (case-insensitive, only alphanumeric
+    characters and ``-``, and must start with an alphabetic character).
+
+    Create a ``CredentialManager`` instance is fast, virtually no initialization
+    needs to be performed. All internal properties are lazily evaluated.
+    This facilitate usage of this facility in code where it is difficult
+    to incorporate a long-lived central instance.
+
+    API
+
+    With one exception, all parameter names of methods in the main API
+    outside ``**kwargs`` must have a ``_`` prefix that distinguishes credential
+    properties from method parameters. The one exception is the ``name``
+    parameter, which is used as a primary identifier (albeit being optional
+    for some operations).
+    """
+    valid_property_names_regex = re.compile(r'[a-z0-9]+[a-z0-9-]*$')
+
+    def __init__(self, cfg=None):
+        """
+
+        Parameters
+        ----------
+        cfg: ConfigManager, optional
+          If given, all configuration queries are performed using this
+          ``ConfigManager`` instance. Otherwise ``datalad.cfg`` is used.
+        """
+        self.__cfg = cfg
+        self.__cred_types = None
+        self.__keyring = None
+
+    # main API
+    #
+    def get(self, name=None, _prompt=None, _type_hint=None, **kwargs):
+        """Get properties and secret of a credential.
+
+        This is a read-only method that never modifies information stored
+        on a credential in any backend.
+
+        Credential property lookup is supported via a number approaches.  When
+        providing ``name``, all existing corresponding configuration items are
+        found and reported, and an existing secret is retrieved from name-based
+        secret backends (presently ``keyring``). When providing a ``type``
+        property or a ``_type_hint`` the lookup of additional properties in the
+        keyring-backend is enabled, using predefined property name lists
+        for a number of known credential types.
+
+        For all given property keys that have no value assigned after the
+        initial lookup, manual/interactive entry is attempted, whenever
+        a custom ``_prompt`` was provided. This include requesting a secret.
+        If manually entered information is contained in the return credential
+        record, the record contains an additional ``_edited`` property with a
+        value of ``True``.
+
+        If no secret is known after lookup and a potential manual data entry,
+        a plain ``None`` is returned instead of a full credential record.
+
+        Parameters
+        ----------
+        name: str, optional
+          Name of the credential to be retrieved
+        _prompt: str or None
+          Instructions for credential entry to be displayed when missing
+          properties are encountered. If ``None``, manual entry is disabled.
+        _type_hint: str or None
+          In case no ``type`` property is included in ``kwargs``, this parameter
+          is used to determine a credential type, to possibly enable further
+          lookup/entry of additional properties for a known credential type
+        **kwargs:
+          Credential property name/value pairs. For any property with a value
+          of ``None``, manual data entry will be performed, unless a value
+          could be retrieved on lookup, or prompting was not enabled.
+
+        Returns
+        -------
+        dict or None
+          Return ``None``, if no secret for the credential was found or entered.
+          Otherwise returns the complete credential record, comprising all
+          properties and the secret. An additional ``_edited`` key with a
+          value of ``True`` is added whenever the returned record contains
+          manually entered information.
+
+        Raises
+        ------
+        ValueError
+          When the method is called without any information that could be
+          used to identify a credential
+        """
+        if name is None and _type_hint is None and not kwargs:
+            # there is no chance that this could work
+            raise ValueError(
+                'CredentialManager.get() called without any identifying '
+                'information')
+        if name is None:
+            # there is no chance we can retrieve any stored properties
+            # but we can prompt for some below
+            cred = {}
+        else:
+            # if we have a chance to query for stored legacy credentials
+            # we do this first to have the more modern parts of the
+            # system overwrite them reliably
+            cred = self._get_legacy_field_from_keyring(
+                name, kwargs.get('type', _type_hint)) or {}
+
+            var_prefix = _get_cred_cfg_var(name, '')
+            # get related info from config
+            cred.update({
+                k[len(var_prefix):]: v
+                for k, v in self._cfg.items()
+                if k.startswith(var_prefix)
+            })
+
+        # final word on the credential type
+        _type_hint = cred.get('type', kwargs.get('type', _type_hint))
+        if _type_hint:
+            # import the definition of expected fields from the known
+            # credential types
+            cred_type = self._cred_types.get(
+                _type_hint,
+                dict(fields=[], secret=None))
+            for k in (cred_type['fields'] or []):
+                if k == cred_type['secret'] or k in kwargs:
+                    # do nothing, if this is the secret key
+                    # or if we have an incoming value for this key already
+                    continue
+                # otherwise make sure we prompt for the essential
+                # fields
+                kwargs[k] = None
+
+        prompted = False
+        for k, v in kwargs.items():
+            if k == 'secret':
+                # handled below
+                continue
+            if _prompt and v is None and cred.get(k) is None:
+                # ask if enabled, no value was provided,
+                # and no value is already on record
+                v = self._ask_property(k, None if prompted else _prompt)
+                if v is not None:
+                    prompted = True
+            if v:
+                cred[k] = v
+
+        # start locating the secret at the method parameters
+        secret = kwargs.get('secret')
+        if secret is None and name:
+            # get the secret, from the effective config, not just the keystore
+            secret = self._get_secret(name, type_hint=_type_hint)
+        if _prompt and secret is None:
+            secret = self._ask_secret(
+                type_hint=self._cred_types.get(
+                    _type_hint, {}).get('secret'),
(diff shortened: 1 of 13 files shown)


datalad/local/credentials.py
```diff
@@ -0,0 +1,519 @@
...
+                title=f"Please enter secret for credential {name!r}",
+                repeat=self._cfg.obtain(
+                    'datalad.credentials.repeat-secret-entry'),
+                hidden=self._cfg.obtain(
+                    'datalad.credentials.hidden-secret-entry'),
+            )
+        # at this point we will have a secret. it could be from ENV
+        # or provided, or entered. we always want to put it in the
+        # store
+        self._keyring.set(name, 'secret', cred['secret'])
```
bpoldrack, line 285: Since the entire thing seems to aim at migrating existing credentials, I think we may want to also try and get rid of using `datalad.support.keyring_._get_service_name` (which `set` does use). I don't see any value in a mandatory `datalad-` prefix, ruling out to use credentials from an existing keychain, just b/c they lack this prefix. I think we'd want to be able to point to any existing one that is accessible to us.

datalad/credman.py
```diff
@@ -0,0 +1,621 @@
...
+        value of `True`.
+
+        If no secret is known after lookup and a potential manual data entry,
+        a plain `None` is returned instead of a full credential record.
+
+        Parameters
+        ----------
+        name: str, optional
+          Name of the credential to be retrieved
+        _prompt: str or None
```
bpoldrack, line 100: Not exactly essential, but seems like `_prompt` should be a dict `{<name of missing property to ask for>: <string to use for prompting>}`, no?

mih: Can you elaborate on why that is needed?

bpoldrack: > For all given property keys that have no value assigned after the
> initial lookup, manual/interactive entry is attempted, whenever
> a custom `_prompt` was provided.

If there's only one string, then all those property keys can only be prompted for by the exact same question it seems. If this is not supposed to only cover a missing `secret`, then several may be needed.

mih: Do you have any example where this would be needed? The current system also has a single one and seems fine.

datalad/local/credentials.py
```diff
@@ -0,0 +1,390 @@
...
+        spec=Parameter(
+            args=("spec",),
+            doc="""specification of[CMD:  a credential name and CMD]
+            credential properties. Properties are[CMD:  either CMD] given as
+            name/value pairs[CMD:  or as a property name prefixed
+            by a colon CMD].
+            Properties with [CMD: no CMD][PY: a `None` PY] value
+            indicate a property to be deleted (action 'set'), or a
+            property to be entered interactively, when no value is set
+            yet, and a prompt text is given (action 'get').
```
adswa, line 179: I find this piece of the docs for the CMD confusing, and I think I found a bug while trying out what I thought was suggested by it.

I read this doc as if this is the correct way to remove the user name from a stored credential:
```
datalad credentials set nitrc user=    
```

But I think from the command line it should rather be
```
datalad credentials set nitrc :user 
```

right? I think it the doc string (here for the parameter and also for the docsting paragraph on set) should say something like
```suggestion
            Properties [CMD: prefixed with a colon CMD][PY: with a `None` value PY] 
            indicate a property to be deleted (action 'set'), or a
            property to be entered interactively, when no value is set
            yet, and a prompt text is given (action 'get').
```

The command above with the empty value actually did lead to a removal of the ``user`` property though (this is a "legacy" credential, unsure if it matters)
```
datalad credentials set nitrc user= 
nitrc(secret): [:user]
```

I think I find that in line with the docs, but it isn't the intended behavior, right?

adswa (commented):
I have played with this command, and I definitely think it is a huge improvement over the current credential system. It seems incredibly flexible and versatile - thanks a lot for implementing this!

Nevertheless, from the user facing side there are a few UX things I found. Most of them boil down to two things:

1) the flexibility of this feature makes it API quite complex, and even when I read the doc strings and helps several times it took me many attempts to get a command right. It also wasn't trivial to translate a CMD command to a Python command in my first attempts without being familiar with the API. I suggested an edit that made it more clear how properties are deleted, I pushed a commit that concretely showed the format of the ``spec`` argument of the Python API, and I pushed a few basic examples that e.g., displayed that ``--prompt`` needs to be in front of the action specification or made extra clear that ``get`` will not save anything in the credential manager even though it returns the user input. I think the more concrete/explicit the docstring and the more examples there are, the better - please add more if you have any.

2) The simultaneous backend character of this command caused me some confusion when I tried to figure out what the added value of each command for a user is. E.g., I personally find it weird that the result rendering of ``query`` tells me the properties of each credential but not their value (I added an example with ``-f json_pp`` for CMD to help people find those values), or that `get` prompts me but then doesn't store stuff (made that clear in the example as well). I think those behaviors are all ok and cool from a "backend"/"internal use" perspective, I just think that users might be surprised so examples help here. If you have any "user story" in mind that can be translated to an example, please consider adding it.
I would appreciate a review and additions to the examples that I added.

adswa:
> Two new configuration settings enable controlling how the interactive entry of credential secrets is conducted for the new credential manager: ``datalad.credentials.repeat-secret-entry`` can be used to turn off the default double-entry of secrets, and ``datalad.credentials.hidden-secret-entry`` can turn off the default hidden entry of secrets. 

it would be so cool if those switches could be mentioned somewhere in the docs for this...

mih:
Thx the review, and in particular for the examples! I extracted a few questions from your review

#### Does it make sense to split the command into a plumbing one and a user-facing one?

Users need all actions, except for `get`. The plumbing command would need all actions, except for (possibly) `remove`. Sounds like too much overlap for a split.

#### `--prompt` placement in CLI can be weird ... must that be?

The command synopsis is

```
datalad credentials [-h] [--prompt PROMPT] [-d DATASET] [--version] [{query|get|set|remove}] [[name] [:]property[=value] ...]
```

This prompt placement is OK. Moreover, `get` is not something that users would run (much). The weird case would be:

```
% datalad credentials get --prompt some :mike
[ERROR  ] unknown argument: :mike 
```

which is a argparse error that cannot determine a destination arg for the `:mike` string. This is hard to prevent without a change in the API. This change would be (CLI) only:

- make `--name` and explicit *option*, that is always mandatory (despite being an option), except for `query`
- make `action` and option that must to be used for 3 out of 4 operations
- OR make `--spec` an option that must never be given as anything other than the last option

None of these trajectories sound particularly attractive to me, in comparison to the proposed API.

#### Why is the result renderer secretive?

It boils down to what to do with the secrets? When using `-f json` they are displayed (no other way). But when people type `datalad credentials` and hit return -- would they appreciate seeing all their passwords verbatim and at once on the screen (or in a logfile)? We could add some kind of marker instead (maybe `***` independent of secret nature and length).

mih:
Closing due to lack of interest.

mih:
Closing due to lack of interest.

PR #6557, opened 2022-03-16
BF: account for "No such password" in addition to not found returned by keyring

### LINKED ISSUE #6552: Keyring field deletion crashes with unknown field
In https://github.com/datalad/datalad/blob/b23174877533ffac3c552f92412ef5398aa20c59/datalad/support/keyring_.py#L88 the code tries to specifically protect against attempting to delete a field that doesn't exist, but it is looking for an exception message (`not found`) that is not part of the actual exception:

`PasswordDeleteError(No such password!)`

yarikoptic:
Closes #6552


mih (approved):

mih:
Metalad failure seems unrelated.

PR #6562, opened 2022-03-17
Make `datalad.cli.tests.test_main.run_main()` also capture UI output

### LINKED ISSUE #6561: `datalad.cli.tests.test_main.run_main()` does not return stdout stderr as claimed
I cannot make it return anything but `('', '')`.

```py
In [1]: from datalad.cli.tests.test_main import run_main

In [2]: run_main(['wtf', '-S', 'python'])
# WTF
## python 
  - implementation: CPython
  - version: 3.9.9
Out[2]: ('', '')

```

This would be a great helper to get the CLI better tested for specific commands too -- not just general CLI functionality.

Update: it can indeed capture any output to stdout/stderr via something like `print()` just fine. But it fails to capture anything from `ui.message()` -- which is what the result renderers use.

And this is not an artifact of the IPython backend:

```
% python -c 'from datalad.cli.tests.test_main import run_main; run_main(["wtf", "-S", "python"]); from datalad.ui import ui; print(ui.backend)' 
# WTF
## python 
  - implementation: CPython
  - version: 3.9.9
dialog
```

mih:
This is necessary, because the UI holds its own handle of `sys.stdout`
that was not affected by the previous patch approach.

With this change, the same StringIO instance is used for both
`sys.stdout` and the `UI.message()` outlet.

With this change, `run_main()` is no suitable for testing more CLI
aspects, such as result renderer output in response to CLI calls.

The resulting three context manager calls have been collapsed into
a single statement, rather than continuing the pattern from before.

#### 🛡 Tests
- The `run_main()` test helper is now able to capture `UI.message()` output too. Fixes #6561

### DIFF
--- a/datalad/cli/tests/test_main.py
+++ b/datalad/cli/tests/test_main.py
@@ -69,22 +69,27 @@ def run_main(args, exit_code=0, expect_stderr=False):
     """
     was_mode = datalad.__api
     try:
-        with patch('sys.stderr', new_callable=StringIO) as cmerr:
-            with patch('sys.stdout', new_callable=StringIO) as cmout:
-                with assert_raises(SystemExit) as cm:
-                    main(["datalad"] + list(args))
-                eq_('cmdline', datalad.get_apimode())
-                assert_equal(cm.exception.code, exit_code)
-                stdout = cmout.getvalue()
-                stderr = cmerr.getvalue()
-                if expect_stderr is False:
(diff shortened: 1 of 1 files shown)


christian-monch:
LGTM

christian-monch (approved):

mih:
Thx for the review!

FTR: Test failures are the "normal" macos ones.

PR #6563, opened 2022-03-17
Deduplicate top-level CLI exception handler output

### LINKED ISSUE #6549: CLI exception handler duplicated exception name
When a `ValueError` reaches the handler it looks like this:

```
[ERROR  ] ValueError(<msg>) (ValueError) 
```

either one of the two exception type names should be removed.

mih:
Whenever an exception reached the `main()` exception handler, it is
converted to a `CapturedException()` and `str(CapturedException)`,
plus the exception type is logged. However, the exception type
is already contained in the `__str__` output:

```
[ERROR  ] ValueError(Value or unset flag ":" missing for property [('ddd',)]) (ValueError)
```

With this change, this duplication is removed and changes the output to

```
[ERROR  ] Value or unset flag ":" missing for property [('ddd',)] (ValueError)
```

This is specifically not identical to the output of
`str(CapturedException())`, in order to maintain the previous spirit of
putting more emphasis on the user-facing message, rather than the
technicality of the specific exception name.

#### 🐛 Bug Fixes
- Remove duplicate exception type in reporting of top-level CLI exception handler. Fixes #6549

### DIFF
--- a/datalad/cli/main.py
+++ b/datalad/cli/main.py
@@ -182,7 +182,7 @@ def _run_with_exception_handler(cmdlineargs):
             exit_code = 3
         else:
             # some unforeseen problem
-            lgr.error('%s (%s)', ce, exc.__class__.__name__)
+            lgr.error('%s (%s)', ce.message, exc.__class__.__name__)
         sys.exit(exit_code)
 
 
--- a/datalad/cli/main.py
+++ b/datalad/cli/main.py
@@ -182,7 +182,7 @@ def _run_with_exception_handler(cmdlineargs):
             exit_code = 3
         else:
             # some unforeseen problem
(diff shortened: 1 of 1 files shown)


datalad/cli/main.py
```diff
@@ -182,7 +182,7 @@ def _run_with_exception_handler(cmdlineargs):
             exit_code = 3
         else:
             # some unforeseen problem
-            lgr.error('%s (%s)', ce, exc.__class__.__name__)
+            lgr.error('%s (%s)', ce.message, exc.__class__.__name__)
```
bpoldrack, line 185: May be this would be a bit more straightforward?

```suggestion
            lgr.error('%s (%s)', ce.message, ce.name)
```

bpoldrack (approved):

mih:
Thx!

Issue #6545, opened 2022-03-14
Downloaders not suitable for internal usage


bpoldrack:
The current implementation of downloaders wasn't aimed at providing download capabilities for somewhat internal operations. An example for the need for that would be the ORA special remote (HTTP case). Instead the error handling as well as the logging of downloaders clearly aim to provide user-facing messaging rather than provide higher-level code with an assessment of what went wrong.
Examples for that include:

- an INFO level log message "Fetching <URL> ...": that level is only useful if that download was something the user would be aware or intended to download of to begin with. For internal usage within ORA or `clone` (checking `origin`'s config when trying to enable special remote(s)) INFO level seems inappropriate
- `check_response_status` in `http.py` does not make use of `raise_from_status`. Hence, we don't, for example, get an `HTTPError` when encountering a 404. For higher-level code to know what's going on we'd currently need to parse the string of a generic `DownloadError` instead.

yarikoptic:
yes, downloaders were written before any notion of ORA and possibly of return result records existed. They are used internally (thus issue title is a bit misleading) IIRC but indeed might need a face lift (or completely redo) to fit better into current internal design paradigm.

bpoldrack:
Yes, I know this is old.
But apart from any legacy, the way exceptions and logging is shaped/worded seems clearly intended for user-messaging. This is only valid if that layer "knows" how and where it is used. That's what I mean by "not suitable". Internal use makes the downloaders a too low level layer to decide about user messaging.

mih:
From my POV this can be closed. The logging issue is gone. The error behavior is just one aspect, why using `download_url` with the context of a special remote is undesirable. The inability to report progress to git-annex is another https://github.com/datalad/datalad/issues/4500#issuecomment-1074241474

bpoldrack:
I don't think the logging issue is gone really. Just that one spot is lowered to DEBUG. But downloaders in general have a bunch of them and if they are used for any internal purpose that can result in the same problem in similar situations.

The use of `download_url` can be replaced (or that function adjusted, since replacement would be near identical), but this would likely reveal more such spots (like `BaseDownloader.download` emitting ` lgr.info("Downloading %r into %r", url, path)`). However, lowering to DEBUG should probably include to double-check where higher-level code may need to emit an INFO instead to not lose the message entirely (or confirm that it is OK to lose it, obv.)

mih:
[...]

PR #6564, opened 2022-03-17
Prevent internal download functionality from logging at INFO level

mih:
See #6545 for the rational. #6545 was unfortunately filed as compound issue, hence
cannot be solved here.

Logging is maintained, but at DEBUG level. This enables using the
downloader functionality when the download is a technical detail (fetch
remote state info) rather then the primary goal (as with data downloads)
-- without unconditionally polluting user-facing output with traced of
internal processing.

It is unclear for which functionality the INFO level message would be or
was possibly essential, but any such functionality should either use
progress logging, or result reporting nowadays.

This remove a line like:

```
[INFO   ] [INFO] Fetching '<URL>/ria-layout-version' 
```

from every attempt to enable a `ria+http` special remote.

- `BaseDownloader.fetch()` is logging download attempts on DEBUG (previously INFO) level to avoid polluting output of higher-level commands. See #6545

### DIFF
--- a/datalad/downloaders/base.py
+++ b/datalad/downloaders/base.py
@@ -632,7 +632,7 @@ class BaseDownloader(object, metaclass=ABCMeta):
         bytes
           content
         """
-        lgr.info("Fetching %r", url)
+        lgr.debug("Fetching %r", url)
         # Do not return headers, just content
         out = self.access(self._fetch, url, **kwargs)
         return out[0]


bpoldrack (approved):

mih:
Failure pattern is #6492

adswa:
ohhh thanks for this!

Issue #6522, opened 2022-03-07
Inappropriate use of sh_quote/quote_cmdlinearg


mih:
`create_sibling()` use this helper to quote commands that will be executed on the remote end. Only POSIX systems are supported on the remote end. However, `quote_cmdlinearg()` will switch quote mode depending on the local platform, not the remote platform. This means that windows quote will be applied.

This is not a serious problem right now, because the SSH code that is needed to actually make use of this is also broken on windows.

PR #6565, opened 2022-03-17
Stop advertising a broken quoting solution for remote commands.

mih:
`quote_cmdlinearg()` uses the local platform to decide on the quoting
style, not the remote.

This change also adjusts an import to make more obvious what helper is being
used.

See datalad/datalad#6522 for details.

### Changelog
Not needed


mih:
Failures unrelated.

PR #6575, opened 2022-03-21
Make sure a download failure reason is always communicated

mih:
Previously, the datalad special remote would report the `__cause__` of an exception, if there was one set -- leading to uninformative error messages of the type:

```
  Failed to download from any of 1 locations [None]
```

for any exception without a declared `__cause__`.

This change falls back on the corresponding `CapturedException` for reporting, whenever no `__cause__` is available. This avoid non-specific error messages.

### Changelog
No needed, likely introduced with change in `master` only.

### DIFF
--- a/datalad/customremotes/datalad.py
+++ b/datalad/customremotes/datalad.py
@@ -52,7 +52,10 @@ class DataladAnnexCustomRemote(AnnexCustomRemote):
                 return
             except Exception as exc:
                 ce = CapturedException(exc)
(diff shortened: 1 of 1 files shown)


yarikoptic:
seems to be working on code which isn't test covered, so I am afraid we will just keep chasing it ... I believe that 11038c2a3c68ec4da9b04ada0db7e9c0944aad82 might have removed the previously test for this (well, original) code path judging on presence of that message in the test body.

Issue #6572, opened 2022-03-21
`datalad` special remote lost progress reporting during retrieval


adswa:
Originally noticed in the context of ``datalad containers-run``: A ``get`` of data from a ``datalad`` special remote does not report any transfer progress.
Example file from a ```datalad``` special remote:
```
(handbook2) adina@muninn in /demo/bids-data on git:master
❱ git -C code/containers annex whereis images/repronim/repronim-reproin--0.9.0.sing
whereis images/repronim/repronim-reproin--0.9.0.sing (4 copies) 
  	23456fab-05d4-441e-8952-b8c6d90ad785 -- yoh@smaug:~/proj/repronim/containers
   	58140b89-2abc-43eb-b89e-7c4cd82acab2 -- [datalad]
   	71c620b5-997f-4849-bb30-c42dbb48a51e -- [datasets.datalad.org]
   	7c7289ab-ad08-41eb-9318-f28e4fd957e7 -- yoh@smaug:/mnt/btrfs/datasets/datalad/crawl/repronim/containers

  datalad: shub://ReproNim/reproin:0.9.0
ok
```

This is how things look on ``master`` during the ``get``:
```
(handbook2) adina@muninn in /demo/bids-data on git:master
❱ datalad containers-run  --container-name code/containers/repronim-reproin
[INFO   ] Making sure inputs are available (this may take some time) 
Total:   0%|                                    | 0.00/374M [00:00<?, ? Bytes/s]
(no updates at all for the complete duration of the download)
```

It used to be like this (on ``maint``):

```
datalad containers-run  --container-name code/containers/repronim-reproin 1 !
[INFO   ] Making sure inputs are available (this may take some time) 
Total:   0%|                                    | 0.00/374M [00:00<?, ? Bytes/s]
Get images/r .. --0.9.0.sing:   3%|▏    | 12.6M/374M [00:16<07:21, 819k Bytes/s]
```

adswa:
I think `AnnexRemote` supports progress reporting: https://lykos153.github.io/AnnexRemote/annexremote/annexremote.html

>  def progress(self, progress)

>    Indicates the current progress of the transfer (in bytes). May be repeated any number of times during the transfer process, but it's wasteful to update the progress until at least another 1% of the file has been sent. This is highly recommended for _store(). (It is optional but good for _retrieve().)
    Parameters

>    progress : int
        The current progress of the transfer in bytes.

It just looks like we haven't yet used it

PR #6573, opened 2022-03-21
UX/BF: Poking in the dark to enable transfer progress from datalad remote

adswa:
I tried to get back a progress bar for file retrieval from ``datalad`` special remotes. This was a purely blind operation (I really don't know special remotes well), but somehow worked, at least for my usecase in #6572.

```
datalad containers-run  --container-name code/containers/repronim-reproin
[INFO   ] Making sure inputs are available (this may take some time) 
Total:   0%|                                    | 0.00/374M [00:00<?, ? Bytes/s]
Get images/r .. --0.9.0.sing:   2%|    | 5.76M/374M [00:04<03:54, 1.57M Bytes/s]
```

Could @yarikoptic and/or @bpoldrack take a look at this, provided it doesn't break the tests?

PR #6165, opened 2021-11-09
Reimplement the datalad(-archives) special remotes with AnnexRemote

mih:
This aims to be a faithful reimplementation of both special remotes. It should not change behavior in any way, even though the implementation changes significantly, capitalizing on the protocol implementation readily provided by `AnnexRemote`.

- Fixes #5006 
- Fixes #5800

TODO:

- [x] "Rebase" `datalad` special remote on `AnnexRemote`
- [x] "Rebase" `datalad-archive` special remote on `AnnexRemote`
- [x] Strip obsolete annex protocol implementation
- [x] Consolidate, deduplicate

PR #6576, opened 2022-03-21
Channel progress logging by AnnexSpecialRemoteProgressBar to git-annex

mih:
(again)

This was removed (unintentionally) when the `datalad` and `-archives` special remotes were reimplemented in https://github.com/datalad/datalad/pull/6165

I only now understood (for a brief moment in time) how this actually works.

This change places a dedicated method in the `AnnexCustomRemote` base class (again) that serves as a receiver and place of documentation.

Fixes datalad/datalad#6572

Replaces #6573

### Changelog
Not needed


adswa:
I can confirm that this fixes things for me, and the failure pattern in the CI is known

PR #6550, opened 2022-03-15
Fixing test failures with recent annex

bpoldrack:
As of annex 10.20220127 we observed several failures on macOS, which at this point is the only build with that recent an git-annex.
These failures boil down to two issues:

- One issue became apparent (actually older), where the reporting of annex commands with respect to paths that are
unknown to git or not existent at all happens in different ways. This depends on annex' internal git calls (git itself appears to be inconsistent WRT this message) as well as on the value of 'annex.skipunknown' the default of which changed in said version. This results in both styles of reporting being used and as long as we rely on parsing that output, we need to account for both.
Moreover, annex commands don't always exit non-zero when that happens, which is why we need to check stderr output when the command doesn't fail, too. At least for `git-annex-find` but probably more commands the zero exit is fixed but unreleased in annex ([10.20220222-30-gce91f1013](https://git.kitenet.net/index.cgi/git-annex.git/commit/?id=ce91f10132805d11448896304821b0aa9c6d9845), Feb 28, 2022 - "fix annex.skipunknown false error propagation").
None of that actually depends on annex' version exactly. It's mostly just that the zero-exit bug in combination with the switch of default made those issues surface.

- The other issue seems to be an annex bug: https://git-annex.branchable.com/bugs/Change_to_annex.largefiles_leaves_repo_modified/
This patch simply skips the respective part of the test for this situation conditioned on the annex version.

Closes #6492
Closes #6570
Closes #6571

Please, review especially this change, @datalad/developers :

This will change behavior of some datalad commands. The change partially
only applies to annex version >= 10.20220322 and `annex.skipunknown=false`
(default since 10.20220127).

Assume following setup:

.
├── bar -> foo
├── bla
├── foo
│   └── f -> ../.git/annex/objects/ ...
└── untracked

Then, previously we had (all zero-exit):

```
$ datalad status --annex availability notexistent foo/f
1 annex'd file (0.0 B/0.0 B present/total size)
nothing to save, working tree clean
$ datalad status --annex availability bar/f foo/f
1 annex'd file (0.0 B/0.0 B present/total size)
nothing to save, working tree clean
$ datalad drop notexistent
$ datalad drop bar/f
$ datalad diff -f HEAD~2 --annex availability bar/f notexistent foo/f
    added: foo/f (file)
```

Now, with non-zero exit:

```
$ datalad status --annex availability notexistent foo/f
status(error): notexistent [File unknown to git]
1 annex'd file (0.0 B/0.0 B present/total size)
[...]

PR #6580, opened 2022-03-22
Use yielding annex record calls for `drop`

### LINKED ISSUE #6578: `drop` feels unresponsive
I run `datalad -l debug  drop -r --reckless kill` (because I felt it was doing nothing). The last message I see before "stalling" is

```
[DEBUG  ] Run ['git', '-c', 'diff.ignoreSubmodules=none', 'annex', 'drop', '--force', '--json', '--json-error-messages', '-c', 'annex.dotfiles=true', '--', '.'] (cwd=/data/project/trusted_computing/phantom_prisma) 
```

When I run `git annex drop --force --json .`, however, I see immediate output of the kind

```
git-annex: failed to lock content: .git/annex/objects/2K/xK/MD5E-s623478--7e4eb97751914bd8a5ebca2f93728bc1.dcm/MD5E-s623478--7e4eb97751914bd8a5ebca2f93728bc1.dcm: setFileMode: permission denied (Operation not permitted)
```

None of these results are anyhow communicated to a user until the very end.

Maybe @christian-monch 's yielding runner can be used here?

christian-monch:
This PR introduces the method `AnnexRepo._call_annex_items_()`, which returns a generator that yields the same results as `AnnexRepo._call_annex()`, but as soon as they are available. The new method is used by `datalad drop`.

Fixes #6578 

- [x] verify error handling is compatible with `AnnexRepo._call_annex()`.

- Improve responsiveness of `datalad drop` in datasets with a large annex.

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -1143,6 +1143,93 @@ class AnnexRepo(GitRepo, RepoInterface):
 
         return return_objects
 
+    def _call_annex_records_items_(self,
+                                   args,
+                                   files=None,
+                                   jobs=None,
+                                   git_options=None,
+                                   stdin=None,
+                                   merge_annex_branches=True,
+                                   progress=False,
+                                   **kwargs):
+        """Yielding git-annex command execution with JSON result processing
+
+        `_call_annex()` is used for git-annex command execution, using
+        GeneratorAnnexJsonProtocol. This means _call_annex() will yield
+        results as soon as they are available.
+
+        For a description of the parameters and raised exceptions, please
+        refer to _call_annex_records().
+
+        Returns
+        -------
+        Generator(something)
+        list(dict)
+          List of parsed result records.
+        """
+        protocol_class = GeneratorAnnexJsonProtocol
+
+        args = args[:] + ['--json', '--json-error-messages']
+        if progress:
+            args += ['--json-progress']
+
+        json_objects_received = False
+        try:
+            for json_object in self._call_annex(
+                                      args,
+                                      files=files,
+                                      jobs=jobs,
+                                      protocol=protocol_class,
+                                      git_options=git_options,
+                                      stdin=stdin,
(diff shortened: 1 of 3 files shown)


datalad/support/annexrepo.py
```diff
@@ -1153,6 +1153,93 @@ def _call_annex_records(self, args, files=None, jobs=None,
...
+            for json_object in self._call_annex(
+                                      args,
+                                      files=files,
+                                      jobs=jobs,
+                                      protocol=protocol_class,
+                                      git_options=git_options,
+                                      stdin=stdin,
+                                      merge_annex_branches=merge_annex_branches,
+                                      **kwargs):
+                if tuple(json_object) == 1 and json_object.get('info', None):
```
yarikoptic, line 1197: What does `tuple(jsin_object)` == 1 means to achieve???

christian-monch: Thanks, that is a mistake, it should be: len(json_object)

datalad/support/annexrepo.py
```diff
@@ -1153,6 +1153,93 @@ def _call_annex_records(self, args, files=None, jobs=None,
...
+            yield from (
+                {
+                    "command": args[0],
+                    "file": f,
+                    "note": "not found",
+                    "success": False,
+                }
+                for f in not_existing
+            )
+
```
yarikoptic, line 1224: the entire body of except handling above here is duplicated now with the one in `_call_annex_records` .  worth centralizing it into a generator so here it would just become

```suggestion
            yield from _handle_not_found(e.stderr)
```
or alike, and in  _call_annex_records just make it into the list

bpoldrack: I am gonna take care of that, following a merge of #6550, since resolving conflict that way will be easier (and either way likely easiet for me ATM)

bpoldrack:
Rebased and added a commit to account for PR #6550

bpoldrack:
Thx much, @christian-monch!

PR #6582, opened 2022-03-23
TMP: formalize extending common_cfg.definitions

### LINKED ISSUE #6314: formalize extending common_cfg.definitions (and other things) in extensions?
Looking at http://docs.datalad.org/en/latest/customization.html (filed separate #6313), and then skimming through extension template and extensions, it seems we have not formalized how extensions could (or should at all) expand the common_cfg.definitions (https://github.com/datalad/datalad/blob/master/datalad/interface/common_cfg.py#L46).  ATM it seems just be a matter of "documenting" those added/used config variables somewhere in docstrings, but IMHO they should be added to aforementioned `definitions`. It would be particularly useful in the scope of the `configuration` command (#6306).

I guess we could just tell people to add entries to that (`common_cfg.definitions`) dictionary.  But I start feeling that we need something like `datalad.support.extensions` module which would provide formalized and centralized interfaces for extensions to extend datalad.  E.g. in the case of extending configuration, could be 

```
def add_config_variable(name, ui, *, destination='global', default=None, type=None)
```  
if we decide to just mimic internal structure present ATM (may be name it `add_common_cfg` to bring closer to current setup).  If we later decide to formalize/change internal config structure or interface to add config variables, we would be able to provide deprecation cycle, while adjusting initial (above) version to map into a new internal data structure, etc.  So, benefits are multiple.

In the scope of #5963 we could also `add_downloader`.

mih:
See what would break

#### 🐛 Bug Fixes
- Description... Fixes #issue
- Fixes #6314
-
#### 📝 Documentation
-
#### 🏠 Internal
-
#### 🛡 Tests
-

### DIFF
--- a/datalad/__init__.py
+++ b/datalad/__init__.py
@@ -183,7 +183,7 @@ def setup_package():
     import tempfile
 
     from datalad import consts
-    from datalad.interface.common_cfg import compute_cfg_defaults
+    from datalad.interface.common_cfg import definitions as cfg_defs
     from datalad.support.annexrepo import AnnexRepo
     from datalad.support.external_versions import external_versions
     from datalad.tests import _TEMP_PATHS_GENERATED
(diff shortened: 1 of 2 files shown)


mih:
I am leaving this unfinished, maybe someone else wants to pick it up.

PR #6583, opened 2022-03-23
add an auto_repr_long decorator to BatchedAnnex

### LINKED ISSUE #6479: Debug output for BatchedAnnex incomprehensible
It looks like this:

```
[DEBUG] Initiating a new process for BatchedAnnex(cmd=<<['git', '-c', ++113 chars++ug']>>, output_proc=<function>, path=<<'/data/group/p++44 chars++sed'>>)
```

for pretty much any call (whichever actual command), presumably, because it has been length-optimized, thereby removing the interesting bits.

This is also an issue for reporting errors from BatchedAnnex runs:

```
[ERROR] BatchedAnnex(command=<<['git', '-c', ++106 chars++ch']>>, encoding='UTF-8', exception_on_timeout=False, generator=<_ResultGenerator>, output_proc=<function>, path=<<'/data/group/p++44 chars++sed'>>, return_code=None, runner=<WitlessRunner>, stderr_output=b'', stdin_queue=<Queue>, timeout=None, wait_timed_out=None): command error: CommandError: 'git -c diff.ignoreSubmodules=none annex addurl --with-files --json --json-error-messages --batch' failed with exitcode -2 
[WARNING] One more exception was received while trying to finish gracefully: AttributeError('NoneType' object has no attribute 'get') 
```

christian-monch:
This PR fixes #6479 

The PR adds a new decorator `auto_repr_long` which provides a similar representation as `auto_repr` but does not abbreviate attribute values.

I have chosen to use `auto_repr_long` instead of changing `auto_repr` to a function with a `abbreviate`-flag in order to keep changes to a minimum

#### 🏠 Internal
- Show full `BatchedAnnex`-attribute values in debug messages

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -60,7 +60,7 @@ from datalad.support.exceptions import CapturedException
 from datalad.ui import ui
 import datalad.utils as ut
 from datalad.utils import (
-    auto_repr,
+    auto_repr_long,
     ensure_list,
     on_windows,
     Path,
@@ -3800,7 +3800,7 @@ class AnnexInitOutput(WitlessProtocol, AssemblingDecoderMixIn):
             lgr.info(line.strip())
 
 
-@auto_repr
+@auto_repr_long
 class BatchedAnnex(BatchedCommand):
     """Container for an annex process which would allow for persistent communication
     """
--- a/datalad/tests/test_utils.py
+++ b/datalad/tests/test_utils.py
@@ -42,6 +42,7 @@ from datalad.utils import (
     all_same,
     any_re_search,
     auto_repr,
+    auto_repr_long,
     better_wraps,
     CMD_MAX_ARG,
     Path,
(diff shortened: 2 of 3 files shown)


yarikoptic:
In the style of the "Survive Style 5+" I first thought to ask: "How/who will decide between using `@auto_repr` and `@auto_repr_long`"?  and thought that it might be better just to make shortening in the repr "optional"/configurable which could be expressed in quick and dirty

```patch
diff --git a/datalad/utils.py b/datalad/utils.py
index d76ca3b51..1fe0e7e14 100644
--- a/datalad/utils.py
+++ b/datalad/utils.py
@@ -114,7 +114,7 @@ def not_supported_on_windows(msg=None):
                                   + (": %s" % msg if msg else ""))
 
 
-def shortened_repr(value, l=30):
+def shortened_repr(value, l=int(os.environ.get('DATALAD_AUTO_REPR_LEN', '30'))):
     try:
         if hasattr(value, '__repr__') and (value.__repr__ is not object.__repr__):
             value_repr = repr(value)
```
so it becomes tunable depending on the need, e.g. with `DATALAD_AUTO_REPR_LEN=10000` making it clearly long enough to include full value whenever/where needed, while keeping logs and our sanity at bay by default.

But then I changed by mind -- for this BatchedAnnexes use case, I think it makes total sense to have those unabbreviated, so I am all for such a change.  **BUT** I dislike breeding decorators whenever it is a matter of parametrizing them.  Just make `@auto_repr` use `@optional_args`... ended up submitting an alternative PR: https://github.com/datalad/datalad/pull/6588

christian-monch:
Closed in favor of #6588

PR #6591, opened 2022-03-25
Consolidate entrypoint processing on importlib.metadata

### LINKED ISSUE #6584: Entrypoint discovery comparison: switch to importlib!!
At present we use `pkg_resources` which is quite dated and feels slow. I implemented an discovery/load benchmark for three solutions:

- `pkg_resources` (what we have)
- `importlib` (what comes with Python)
- `reentry` (claims to have speed as a feature)

The benchmark is this (only the start is different):

```py
# ----- pkg_resources
from pkg_resources import iter_entry_points
for ep in iter_entry_points('datalad.extensions'):
# ----- importlib
import sys
if sys.version_info < (3, 10):
    from importlib_metadata import entry_points
else:
    from importlib.metadata import entry_points
for ep in entry_points(group='datalad.extensions'):
# ----- reentry
from reentry import manager
for ep in manager.iter_entry_points(group='datalad.extensions'):
# ------ common
    # just discovery
    print(ep)
    # incl import
    print(ep.load())
```

The benchmark ran on current mainline, with exactly one extension installed and called as `multitime -n 5 python <script>`. Here is what the timing looks like:

#### Discovery

```
# ----- pkg_resources
real        0.126       0.002       0.124       0.125       0.128       
user        0.117       0.003       0.112       0.117       0.120       
sys         0.009       0.002       0.008       0.008       0.012       

# ----- importlib
real        0.074       0.003       0.068       0.075       0.077       
user        0.060       0.006       0.048       0.062       0.066       
sys         0.014       0.004       0.008       0.012       0.020       

# ----- reentry
real        0.063       0.001       0.061       0.063       0.063       
user        0.048       0.005       0.042       0.050       0.055       
sys         0.014       0.005       0.008       0.012       0.021       
```

Both alternatives are twice as fast than what we presently use.

#### Discovery and import

Results from `multitime -n 20 python <script>` due to more variability in the performance.

```
# ----- pkg_resources
real        0.369       0.020       0.327       0.369       0.414       
user        0.314       0.018       0.274       0.316       0.350       
sys         0.058       0.010       0.038       0.056       0.075       

# ----- importlib
real        0.179       0.013       0.159       0.178       0.199       
user        0.135       0.010       0.117       0.136       0.153       
sys         0.049       0.010       0.031       0.051       0.068       

# ----- reentry
real        0.166       0.015       0.128       0.167       0.185       
user        0.126       0.013       0.103       0.125       0.161       
sys         0.045       0.008       0.028       0.046       0.055       
```

The median difference suggests saving of ~200ms!

#### Discovery and import for 11 extensions (11x increase)

Results from `multitime -n 20 python <script>` due to more variability in the performance.

There are no results for `reentry`. My code was not able to discover more than one extension.

```
# ----- pkg_resources
real        0.885       0.018       0.861       0.883       0.937       
user        0.739       0.024       0.684       0.732       0.794       
sys         0.162       0.016       0.141       0.159       0.212       

# ----- importlib
real        0.498       0.021       0.463       0.496       0.541       
user        0.369       0.021       0.323       0.368       0.427       
sys         0.146       0.019       0.120       0.144       0.183       
```

~400ms absolute and ~50% faster.

I believe it is a no-brainer to make the switch to `importlib`. It would be possible to explore why `reentry` is not functioning properly, as it was slightly faster on the simple test case. However, in contrast to `importlib` this would be an external dependency, hence slightly less attractive.

Neither approach would require any adjustments in extension code.

### LINKED ISSUE #6585: `iter_entrypoints` helpers
We have about 5 different copies of an implementation to iterate over entry points and safely (or not safely) load them. In the light of https://github.com/datalad/datalad/issues/6584 we should consolidate them to a uniform (set of) helper(s).

mih:
This approach is now an offiical part of the standard lib (since 3.10), and is available for earlier version in a provisional module and also a separate package.

This change introduces `datalad.support.entrypoints` with `iter_entrypoints()` that is now used as a uniform implementation.

As previously shown in https://github.com/datalad/datalad/issues/6584 this brings significant performance improvements for internal processing. Moreover, they translate to outward-facing speed gains. On a system with 10 extensions installed a `datalad -h` now completes in ~300ms, whereas current mainline after the CLI update needs ~650ms, and the latest release 0.15.6 needs ~1200ms -- so overall a 4x speed-up.

Importantly, the new `iter_entrypoints()` helper does not return complex entrypoint instances anymore, which previously contaminated DataLad code with the internals of `pkg_resources`. This is also the reason for one occurrence of a `pkg_resources`-based entrypoint iteration being left in the code (in `search` and its tests). It specifically relies on these
internals and even mocks them in the tests.

This code is likely to change entirely soon, and is also slow enough to not make the require effort to switch it over felt by users.

This change implies no necessary modifications of any extensions.

- Loading DataLad extension packages has been sped-up leading to between 2x and 4x faster run times for loading individual extensions and reporting help output across all installed extensions. Fixes #6584
#### 🏠 Internal
- Entrypoint processing for extensions and metadata extractors has been consolidated on a uniform helper that is about twice as fast as the previous implementations. Fixes #6585

### DIFF
--- a/datalad/support/entrypoints.py
+++ b/datalad/support/entrypoints.py
@@ -1,0 +1,65 @@
+# emacs: -*- mode: python; py-indent-offset: 4; tab-width: 4; indent-tabs-mode: nil -*-
+# ex: set sts=4 ts=4 sw=4 et:
+# ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+#
+#   See COPYING file distributed along with the datalad package for the
+#   copyright and license terms.
+#
+# ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+"""Core utilities"""
+
+import logging
+import sys
+
+from datalad.support.exceptions import CapturedException
+
+lgr = logging.getLogger('datalad.support.entrypoints')
+
+
+def iter_entrypoints(group, load=False):
+    """Iterate over all entrypoints of a given group
+
+    Parameters
+    ----------
+    group: str
+      Name of the entry point group to iterator over, such as
+      'datalad.extensions'.
+    load: bool, optional
+      Whether to execute the entry point loader internally in a
+      protected manner that only logs a possible exception and emits
+      a warning, but otherwise skips over "broken" entrypoints.
+      If False, the loader callable is returned unexecuted.
+
+    Yields
+    -------
+    (name, module, loade(r|d))
+      The first item in each yielded tuple is the entry point name (str).
+      The second is the name of the module that contains the entry point
+      (str). The type of the third items depends on the load parameter.
+      It is either a callable that can be used to load the entrypoint
+      (this is the default behavior), or the outcome of executing the
+      entry point loader.
+    """
+    lgr.debug("Processing entrypoints")
+
+    if sys.version_info < (3, 10):
+        # 3.10 is when it was no longer provisional
+        from importlib_metadata import entry_points
+    else:
+        from importlib.metadata import entry_points
+    for ep in entry_points(group=group):
+        if not load:
+            yield ep.name, ep.module, ep.load
+            continue
+
+        try:
+            lgr.debug('Loading entrypoint %s from %s', ep.name, group)
+            yield ep.name, ep.module, ep.load()
+            lgr.debug('Loaded entrypoint %s from %s', ep.name, group)
+        except Exception as e:
+            ce = CapturedException(e)
+            lgr.warning(
+                'Failed to load entrypoint %s from %s: %s',
+                ep.name, group, ce)
+            continue
+    lgr.debug("Done processing entrypoints")
--- a/datalad/api.py
+++ b/datalad/api.py
@@ -43,32 +43,20 @@ if not datalad.in_librarymode():
 def _generate_extension_api():
     """Auto detect all available extensions and generate an API from them
     """
-    from importlib import import_module
-    from pkg_resources import iter_entry_points
-    from .interface.base import get_api_name
-    from datalad.support.exceptions import CapturedException
+    from datalad.support.entrypoints import iter_entrypoints
+    from datalad.interface.base import (
+        get_api_name,
+        load_interface,
+    )
 
     import logging
     lgr = logging.getLogger('datalad.api')
 
-    for entry_point in iter_entry_points('datalad.extensions'):
-        try:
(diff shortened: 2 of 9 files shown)


datalad/support/entrypoints.py
```diff
@@ -0,0 +1,65 @@
...
+      (this is the default behavior), or the outcome of executing the
+      entry point loader.
+    """
+    lgr.debug("Processing entrypoints")
+
+    if sys.version_info < (3, 10):
+        # 3.10 is when it was no longer provisional
+        from importlib_metadata import entry_points
+    else:
+        from importlib.metadata import entry_points
```
yarikoptic, line 49: line is not covered, which is odd since we do have 3.10 run in travis

datalad/support/entrypoints.py
```diff
@@ -0,0 +1,65 @@
...
+    for ep in entry_points(group=group):
+        if not load:
+            yield ep.name, ep.module, ep.load
+            continue
+
+        try:
```
yarikoptic, line 55: there is no test/tests matrix run which triggers this with `load=True`

yarikoptic (commented):
I wholeheartedly support this PR: faster, less code duplication etc.  I first time mentioned that deprecated being failing, but it seems to not be the first time if I look at https://github.com/datalad/datalad/actions/workflows/test_extensions.yml -- started to fail some time within past 2 days -- run 2 days ago was green and then run on this (and then other PRs) PR 8 hours ago was red :-/  And failure doesn't look related to this PR changes.

I am not 100% certain though on what @mih you wanted to say by

> This code is likely to change entirely soon, and is also slow enough to not make the require effort to switch it over felt by users.

Also it seems that there is no test coverage at all for the `load=True` case which is odd - we do not send out coverage from extensions testing runs or how else could we make it cover that (crucial piece of) code?

datalad/support/entrypoints.py
mih: I am not sure how much these coverage reports can be trusted. This code is exercises by both the construction of the Python API (`datalad.api`) and the CLI. 

The latter is chronically undertested -- something that this PR won't fix. I started using a seemingly attractive pattern for such tests in #6555 for testing the proposed `credentials` command.

yarikoptic:
ok, still a bit concerned about coverage, but let's proceed -- will be tested in real life ;) Thanks @mih !

PR #6594, opened 2022-03-25
Enable loading select extensions on CLI startup or API import 

### LINKED ISSUE #6589: Add `--with` to CLI to selectively enable extension support "outside" extension commands
There are various shortcomings re how extensions can influence the core package: https://github.com/datalad/datalad/issues/5996

However, for DataLad use in a Python session most of them are more of a nuisance than a strict limitation, because users can simply import an extension directly in such a session, enabling the extension to arbitrarily customize the core package.

There is no equivalent in the CLI. In contrast, the CLI tries hard to avoid extension imports that are not needed for execution a specific command -- only once it is clear that an extensions needs to be imported, this is finally done. Consequently, running core commands like `save` or `run` through the CLI will not import any extensions. This, more than anything else, limits the effective impact of extension code.

I propose to implement a `--with` option for the main DataLad CLI entrypoint. If given (could be multiple times), an extension entrypoint for the given name will be loaded, which causes the extension to be imported. A matching configuration setting would allow for configuring a set of extensions to be always loaded unconditionally for a user and/or dataset. This puts extension code into the position to customize behavior and capabilities of core commands, arbitrarily. The name is chosen, because the mechanism resembles a `with patch(....):` context manager use for a DataLad session.

Such a feature would enable use cases like:

- make clone run something other than `git-clone` to establish a repo https://github.com/datalad/datalad-ebrains/issues/21, e.g. 

  `datalad --with ebrains clone https://search.kg.ebrains.eu/instances/e472a8c7-d9f9-4e75-9d0b-b137cecbc6a2`

- make `clone` and `push` deploy necessary credentials to enable or push to special remotes of particular types https://github.com/mih/datalad-mihextras/issues/30

If this acknowledged as a recognized approach to customize datalad, existing internal helpers could be promoted to a more "public" status with a more careful/stable API choice for reliably supporting this kind of "remote decoration".

If this feature would have existed, it would likely have avoided the need to (prematurely) merge the RIA code base into the core package -- which was only needed, because `clone` could not be customized with the necessary pre and post annex-init hooks.

mih:
Performance impact when this feature is unused, is negligible. Otherwise it is determined by the import time of a particular extension's `__init__.py`.

- A new configuration switch `datalad.extensions.load` can be used to load particular DataLad extension packages on startup. This enables extensions to also enhance core commands, such as teaching the `clone` command to handle additional types of URLs. To continue to use commands provided by extensions, this new configuration is not necessary. Fixes #6589

### DIFF
--- a/datalad/support/entrypoints.py
+++ b/datalad/support/entrypoints.py
@@ -63,3 +63,31 @@ def iter_entrypoints(group, load=False):
                 ep.name, group, ce)
             continue
     lgr.debug("Done processing entrypoints")
+
+
+def load_extensions():
+    """Load entrypoint for any configured extension package
+
+    Log a warning in case a requested extension is not available, or if
+    a requested extension fails on load.
+
+    Extensions to load are taken from the 'datalad.extensions.load'
+    configuration item.
+    """
+    from datalad import cfg
+    load_extensions = cfg.get('datalad.extensions.load', get_all=True)
+    if load_extensions:
+        from datalad.utils import ensure_list
+        exts = {
+            ename: eload
+            for ename, _, eload in iter_entrypoints('datalad.extensions')
+        }
+        for el in ensure_list(load_extensions):
+            if el not in exts:
+                lgr.warning('Requested extension %r is not available', el)
+                continue
+            try:
+                exts[el]()
+            except Exception as e:
+                ce = CapturedException(e)
+                lgr.warning('Could not load extension %r: %s', el, ce)
--- a/datalad/cli/main.py
+++ b/datalad/cli/main.py
@@ -73,6 +73,13 @@ def main(args=sys.argv):
         from .helpers import _fix_datalad_ri
         args = [_fix_datalad_ri(s) for s in args]
 
+    from datalad.support.entrypoints import load_extensions
+    # load extensions requested by configuration
+    # analog to what coreapi is doing for a Python session
+    # importantly, load them prior to parser construction, such
+    # that CLI tuning is also within reach for extensions
+    load_extensions()
+
     # PYTHON_ARGCOMPLETE_OK
     # TODO possibly construct a dedicated parser just for autocompletion
     # rather than lobotomizing the normal one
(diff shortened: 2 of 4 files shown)


datalad/support/entrypoints.py
```diff
@@ -0,0 +1,93 @@
...
+        }
+        for el in ensure_list(load_extensions):
+            if el not in exts:
+                lgr.warning('Requested extension %r is not available', el)
+                continue
+            try:
+                exts[el]()
+            except Exception as e:
+                ce = CapturedException(e)
+                lgr.warning('Could not load extension %r: %s', ce)
```
bpoldrack, line 93:
```suggestion
                lgr.warning('Could not load extension %r: %s',el, ce)
```

mih: Thx!

bpoldrack (approved):
Looks good to me. I have played with it, too.

Just one thing, that isn't clear to me: `datalad.api` still loads extensions unconditionally. Is that intentional? I have no strong feelings about it, but I suspect that can be surprising and therefore unintentionally triggered.

mih:
I did not claim any modification of `datalad.api` behavior. If no change in behavior surprises, then I don't know what to do about it.

bpoldrack:
That's ok. I just wanted to be sure it's not an oversight and am wondering what to document WRT to when extensions are loaded if one uses the python API (not a request for this PR, but `datalad.api` needs a proper doc not only for this reason).

mslw:
I played with it a little bit too. I think I like the idea.

At first I was worried that if I use an extension for a single project (and therefore have it installed in a virtualenv). and set `datalad.extensions.load something` globally, then I'd be seeing a warning for every single DataLad command used outside that virtualenv (ie. in other envs / globally). But then I realized that it's a non-issue, or actually a good feature - if I don't intend to use this extension globally, then I shouldn't set the option globaly, or set/unset it when activating the virtual environment (eg hooks in virtualenvwrapper), am I right?. So in any case it's about correct usage, and the warning is actually what it should be, and I wouldn't change anything.

PR #6600, opened 2022-03-30
disable metalad tests

christian-monch:
This PR will disable metalad tests. Currently metalad is changing a lot and CI does not seem to respect pinned versions of required pypi packages. That leads to environements with non-working metalad installations.

### DIFF
--- a/.github/workflows/test_extensions.yml
+++ b/.github/workflows/test_extensions.yml
@@ -13,7 +13,6 @@ jobs:
         extension: [
             datalad-neuroimaging,
             datalad-container,
-            datalad-metalad,
             datalad-crawler,
             datalad-deprecated,
         ]


yarikoptic:
> This PR will disable metalad tests. Currently metalad is changing a lot and CI does not seem to respect pinned versions of required pypi packages. That leads to environements with non-working metalad installations.

I don't see any pinned versions in https://github.com/datalad/datalad-metalad/blob/master/setup.cfg#L16 .  I think it would be very valuable to bring metalad to some consistent/stable state either with most recent datalad release in 0.15.x series or current master, and have it released on pypi .

christian-monch:
> > This PR will disable metalad tests. Currently metalad is changing a lot and CI does not seem to respect pinned versions of required pypi packages. That leads to environements with non-working metalad installations.
> 
> I don't see any pinned versions in https://github.com/datalad/datalad-metalad/blob/master/setup.cfg#L16 . I think it would be very valuable to bring metalad to some consistent/stable state either with most recent datalad release in 0.15.x series or current master, and have it released on pypi .

On it. Planned to happen tomorrow or friday

PR #6601, opened 2022-03-30
NF: register_config()

### LINKED ISSUE #6314: formalize extending common_cfg.definitions (and other things) in extensions?

mih:
This PR requires and sits on top of #6594. The first commit is not unique to this PR.

This also replaces the previously very flexible data structure for such items with a more rigid set of classes that implement lazy evaluation of defaults (avoiding their possibly expensive computation, even when they are not actually needed). They otherwise implement all elements of the previously used API, such that they work as a drop-in replacement.

Only register_config() (and its friend `has_config()` is intended to be used publicly. Like so:

```py
from datalad.support.extensions import register_config
register_config(
    'datalad.mihextras.blows',
    title="Does it?",
    default=True,
    scope='global',
    type=bool,
    description="I could tell you, but I won't!",
)

```
- A new module `datalad.support.extensions` offers the utility functions `register_config()` and `has_config()` that allow extension developers to announce additional configuration items to the central configuration management. Fixes #6314


yarikoptic:
+1 on `register_config` since in-line with what I was looking after in https://github.com/datalad/datalad/issues/6314 . I would only also "expose" it within some centralized `datalad.support.extensions` module.

> This PR requires and sits on top of https://github.com/datalad/datalad/pull/6594.

but in principle this functionality is not needing #6594, right? i.e. extension can only define config variable **it** uses, thus would need to be imported by then one way (as via `datalad.api` or CLI) or another (the #6594), most likely using that `register_config` within that extension's `__init__.py`, correct?

bpoldrack:
> This PR is currently in draft mode to gather feedback on the API, before tests are added. 

Feedback: I like it.

mih:
Can be seen in action here: https://github.com/datalad/datalad-next/blob/master/datalad_next/__init__.py (running across platforms)

mih:
> > This PR requires and sits on top of #6594.
> 
> but in principle this functionality is not needing #6594, right? i.e. extension can only define config variable **it** uses, thus would need to be imported by then one way (as via `datalad.api` or CLI) or another (the #6594), most likely using that `register_config` within that extension's `__init__.py`, correct?

#6594 is already needed to give the `configuration` command the chance to see such definitions.

datalad/interface/common_cfg.py
```diff
@@ -28,9 +29,185 @@
...
 
+lgr = logging.getLogger('datalad.interface.common_cfg')
 dirs = AppDirs("datalad", "datalad.org")
 
 
+class _NotGiven():
+    pass
+
+
+def register_config(
```
bpoldrack, line 40: I think, this lacks a conflict prevention. While it's okay to override existing configs from core (such as changing a default value), name conflict with other extensions is likely unintentional. May be override should require a `force=True` to make clear that it is intended.

datalad/interface/common_cfg.py
```diff
@@ -28,9 +29,185 @@
...
+      explanatory inline comment just prior the item definitions.
+    default: optional
+      A default value that is already known at the time of registering the
+      configuration items. Can be of any type.
+    default_fn: callable, optional
+      A callable to compute a default value lazily on access. The can be
+      used, if the actual value is not yet known at the time of registering
+      the configuration item, or if the default is expensive to compute
+      and its evaluation need to be deferred to prevent slow startup
+      (configuration items are typically define as one of the first things
```
mslw, line 86:
```suggestion
      and its evaluation needs to be deferred to prevent slow startup
      (configuration items are typically defined as one of the first things
```

datalad/interface/common_cfg.py
mih: I cannot see this having a net-positive effect. The person to set `force=True` is a developer of an extension. They, as any other person, will not see the big picture of all extensions and future versions of core. So not setting `force=True` has the "benefit" of informing *users* at some point that they cannot use an extension (anymore). Their only cause of action is to inform developers of the affected packages and hope for a timely release of an update. As an extension developer, i can prevent this inconvenience by always setting `force=True` (which is what I implemented), and which is not change from the status quo. IMHO it is an illusion that 'conflict prevention' is done in the scope of this function. If there ever is a conflict, the resolution will need to come with a (large-scale) migration of existing configuration users have lying around on any system and clone, or in any dataset branch.

I believe the only approach here to avoid conflicts is proper monitoring of developers by developers, and good changelogs.

That being said, if anyone has a good idea how to approach this in a meaningful way, please propose a solution in a dedicated PR.

datalad/interface/common_cfg.py
mih: Thanks! I will apply this patch in the next update.

datalad/interface/common_cfg.py
bpoldrack: Good points. Agree with addressing elsewhere if and when there's a better idea.

mih:
I took this out of draft mode. Test coverage is now 100%, and missing functionality for the doc builder was added. I also added `has_config()` to mitigate the concern expressed in https://github.com/datalad/datalad/pull/6601#discussion_r839340666

bpoldrack:
Everyone in agreement it seems. Merging.

Issue #6599, opened 2022-03-30
Flaky SSH cloning failure in tests


bpoldrack:
Extracted from https://github.com/datalad/datalad/pull/6550#issuecomment-1073024568:

This only happens on AppVeyor (macOS again):

```
======================================================================
ERROR: datalad.core.distributed.tests.test_clone.test_ria_postclonecfg('ssh://datalad-test:/Users/appveyor/DLTMP/datalad_temp_ix8umpb9', '07c27167-6fef-443c-bbb7-3eec35daddc3')
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/appveyor/venv3.8.12/lib/python3.8/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/Users/appveyor/projects/datalad/datalad/tests/utils.py", line 288, in _wrap_skip_ssh
    return func(*args, **kwargs)
  File "/Users/appveyor/projects/datalad/datalad/tests/utils.py", line 874, in _wrap_with_tempfile
    return t(*(arg + (filename,)), **kw)
  File "/Users/appveyor/projects/datalad/datalad/tests/utils.py", line 874, in _wrap_with_tempfile
    return t(*(arg + (filename,)), **kw)
  File "/Users/appveyor/projects/datalad/datalad/core/distributed/tests/test_clone.py", line 958, in _test_ria_postclonecfg
    riaclone = clone('ria+{}#{}'.format(url, dsid), clone_path)
  File "/Users/appveyor/projects/datalad/datalad/interface/utils.py", line 447, in eval_func
    return return_func(*args, **kwargs)
  File "/Users/appveyor/projects/datalad/datalad/interface/utils.py", line 439, in return_func
    results = list(results)
  File "/Users/appveyor/projects/datalad/datalad/interface/utils.py", line 424, in generator_func
    raise IncompleteResultsError(
datalad.support.exceptions.IncompleteResultsError: Command did not complete successfully. 1 failed:
[{'action': 'install',
  'message': ('Failed to clone from any candidate source URL. Encountered '
              'errors per each url were:\n'
              '- %s',
              'ssh://datalad-test/Users/appveyor/DLTMP/datalad_temp_ix8umpb9/07c/27167-6fef-443c-bbb7-3eec35daddc3\n'
              "  CommandError: 'git -c diff.ignoreSubmodules=none clone "
              '--progress '
[...]

PR #6606, opened 2022-03-31
Sanitize AppVeyor's known_hosts

bpoldrack:
Remove AppVeyor's world-matching default entries for gitlab.com in `known_hosts`, that led to `Remote host identification changed` with pretty much every SSH invocation.

See https://github.com/datalad/datalad/issues/6599#issuecomment-1085744290


yarikoptic (approved):
Unfortunately I think we never actually saw that msg in the CI logs
```
(git)smaug:/mnt/datasets/datalad/ci/logs/2022/04[master]git
$> git grep 'Remote host identification changed'
```
so can't confirm that it is gone now. But since it is green -- let's proceed even if I can't confirm that it has any effect ;)

Issue #3127, opened 2019-01-25
publish should automagically "annex export" instead of "annex copy" to the special remote configured for export


yarikoptic:
ignorant me thought that annex export happens automagically upon "annex copy" if remote is configured with `exporttree=yes`. [Apparently](https://git-annex.branchable.com/git-annex-export/) and unfortunately it is not the case and it requires use of dedicated `annex export BRANCH` command to export the tree to such special remote.  I think we could/should simplify publishing to the "exported" special remotes by calling `annex export` (not copy) if the annex special remote has exporttree configured.  The possible gotcha (I didn't work with exported remotes much) is confusion if export is attempted from different branches.

adswa:
This would be a cool feature. FWIW, it would also be helpful for ``datalad-osf``. There is a dedicated sibling mode ``export-only`` that so far requires plain git annex operations for publishing data (and remembering that you had to do them). Having push/publish infer it automatically would simplify things!

yarikoptic:
number of usecases / user inquiries grows (e.g. https://neurostars.org/t/configuring-remotes-on-and-pushing-to-github-gin/20244/9) so I think we should work that support out. Could be as simple as just using `export` instead of `copy` if remote is `exporttree` remote

mih:
Thanks for the demonstrator. I have implemented a related test.

mih:
Looking into the implementation logic further, there are a few complications.

The current `push` discovers datasets to push via a `diff`. In doing so it does not have any information on whether there were particular path constraints relevant for a specific subdataset readily available, at the time when it comes to making the decision to `annex copy`. This is relevant here, because `annex export` can export a `treeish`. This could be the root of a dataset, but also just a subdir. Furthermore, `git annex` cannot detect export conflicts immediately.

We could take the exported treeish from `export.log` and try to find the associated commit, which is a plain loop over `rev-list`, while inspection the result of `cat-file -p <commit>`, matching the `tree` property -- but only for top-level trees.

Only in one simple case of all combinations we could act sensibly: no path constraints given, root tree exported. Everything else would require sophisticated analysis in order to figure out, whether we are messing up a previous export -- of course all already assuming that this particular repository was the one from which exports were done before.

Proposal:

Rather than (incomplete) verification and inference, we could introduce configuration that simply defines what `push` shall do for an `exporttree=yes` special remote.

- `datalad.push.export-root`
[...]

PR #6618, opened 2022-04-07
Factor out actual data transfer in push

christian-monch:
This commit delegates the actual data transfer in
push to a new method: _transfer_data. This is done
to simplify monkey-patching via "datalad_next", in
order to implement export to git-annex special remotes
with type webdav and exporttree enabled.

This is a first step to address issue #3127 
#### 🏠 Internal
- factor out actual data transfer in push, in order to make it "patchable".

### DIFF
--- a/datalad/core/distributed/push.py
+++ b/datalad/core/distributed/push.py
@@ -45,6 +45,7 @@ from datalad.support.exceptions import CommandError
 from datalad.utils import (
     Path,
     ensure_list,
+    todo_interface_for_extensions,
 )
 
 from datalad.distribution.dataset import (
@@ -401,18 +402,19 @@ def _datasets_since_(dataset, since, paths, recursive, recursion_limit):
         yield (cur_ds, ds_res)
 
 
+@todo_interface_for_extensions
 def _transfer_data(repo, ds, target, content, data, force, jobs, res_kwargs,
                    got_path_arg):
-        yield from _push_data(
-            ds,
-            target,
-            content,
-            data,
-            force,
-            jobs,
-            res_kwargs.copy(),
-            got_path_arg=got_path_arg,
-        )
+    yield from _push_data(
+        ds,
+        target,
+        content,
+        data,
+        force,
+        jobs,
+        res_kwargs.copy(),
+        got_path_arg=got_path_arg,
+    )
 
 
 def _push(dspath, content, target, data, force, jobs, res_kwargs, pbars,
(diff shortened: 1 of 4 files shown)


datalad/core/distributed/push.py
```diff
@@ -401,6 +401,20 @@ def _datasets_since_(dataset, since, paths, recursive, recursion_limit):
         yield (cur_ds, ds_res)
 
 
+def _transfer_data(repo, ds, target, content, data, force, jobs, res_kwargs,
```
yarikoptic, line 404: please add a comment or docstring for thing function then on its purpose in life, or the next PR from someone could be "collapsing not really necessary function".

But altogether -- such monkey patching from extensions is opening a precedence, or are we doing it already in other spots?  

We should at least come up with some consistent annotation so we could `git grep` for such locations: let's add a "do nothing" decorator `@todo_interface_for_extensions` to annotate such methods.  Could come handy later to identify such hidden treasures.

christian-monch: It was introduced by @mih in `datalad-next` and was already used to patch the method `_set_request_headers` of `datalad.distributed.create_sibling_ghlike._GitHubLike`.

There is a discussion of the approach in another issue that I do not find yet. The monkey-patching approach is new and we have to see what can of worms it opens. The decorator is definitely a good idea! I will add it

yarikoptic (approved):
Since we have nothing better ATM, I guess it is Ok, although I am afraid could be quite fragile "long term" -- extensions could then fight over who monkey patches first, leading to odd/hard to troubleshoot behaviors etc.

yarikoptic:
if it isn't WiP, please adjust the title.

yarikoptic:
@christian-monch Let's rebase this against maint. Since it is just factoring out, it is IMHO ok. While at it please annotate with the decorator just merged into `maint` https://github.com/datalad/datalad/pull/6627/files#diff-124eede0287a62652d5fa856d4b11d371ac5484cab67bbc096a0b6656e6f8555R155  . If all kosher with CI after rebase etc, we would then mark this PR for `release`.

yarikoptic:
eh, now we need another PR with a `release` label to kick out the release -- what would it be!? ;)

mih:
Here is a new candidate: https://github.com/datalad/datalad/pull/6635

I doubt there will be a shortage of them over the coming days. ;-)

PR #6624, opened 2022-04-11
BF: Fix wrong key name of annex' JSON records

### LINKED ISSUE #6621: Annexrepo: `error_message` vs `error-messages` in `get_content_info`
While grepping for `error_message` to investigate #6620 I came upon this recently merged piece, and I'm not sure if there isn't a misnomer:

https://github.com/datalad/datalad/blob/6b4280dcb3e31a2efc6b8d8be544655c08992982/datalad/support/annexrepo.py#L3397-L3402

In this case, `j` comes from `self.call_annex_records(cmd, files=files)` I'm not sure about the context, but it looks like:
- all key/value pairs, except `error_messages` get copied from git-annex record to `rec`
- accordingly, comment then says "change annex' `error_messages` into singular to match result records
- then the content of `error-messages` is pieced together and written to `error_message`

AFAIK git-annex returns `error-messages`, and I'm not sure if `error_messages` can be present at all. and if I'm reading the code correctly, then the `error-messages` field (containing a list) will pass the initial check, and will be included alongside the `error_message` field (containing a string). The intention may have been to do so, but it looks to me like the conditional and comment were meant to say `error-messages`.

@bpoldrack sorry to bring this comment a week after the PR was merged, but since you worked with this code recently, could you verify what was the intention? In case I misunderstood, sorry for the buzz.

bpoldrack:
Annex uses a `-` instead of `_` in its JSON key names.
This correction should only have a minor impact on behavior. Instead of
the intended replacement of a list of messages under `error-messages`
with a joined message under `error_message`, the created record
unintentionally had both.

Fixes #6621

#### 🏠 Internal
- Fix misspelled use of a key name in annex' JSON records.

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -3395,8 +3395,8 @@ class AnnexRepo(GitRepo, RepoInterface):
                 else:
                     rec = {}
             rec.update({'{}{}'.format(key_prefix, k): j[k]
-                       for k in j if k != 'file' and k != 'error_messages'})
-            # change annex' `error_messages` into singular to match result
+                       for k in j if k != 'file' and k != 'error-messages'})
+            # change annex' `error-messages` into singular to match result
             # records:
             if j.get('error-messages', None):
                 rec['error_message'] = '\n'.join(m.strip() for m in j['error-messages'])


mslw:
It definitely addresses the issue - code looks less ambiguous now. As we discussed in another thread, keeping `error-messages` is not bad *per se*, but they probably wouldn't be used elsewhere and keeping just `error_message` eliminates the need for second guessing.

I see two test runs fail on mac, but on a cursory glance I don't think they're related (also the failure seems very selective).

bpoldrack:
Thx, @mslw. Yes, those failures are the known issue with the snapshot (annex version that that test cannot properly match) and #6599

PR #6627, opened 2022-04-15
Factor out credential handling of create-sibling-ghlike

mih:
This achieves the same modularization of the code as in #6555, but without making any changes to the actual credential handling.

This really should have made it into 0.16, but I seem to have lost track of this one missing commit. It is a mere code reorganization. Without this patch being part of a release, I need to switch the `next` extension back to a Git snapshop dependency again.


datalad/distributed/create_sibling_ghlike.py
```diff
@@ -145,27 +145,30 @@ def __init__(self, url, credential, require_token=True, token_info=None):
...
-        if token_info:
-            token_instructions += f'. {token_info}'
+    def _set_request_headers(self, credential_name, auth_info, require_token):
+        if credential_name is None:
+            credential_name = urlparse(self.api_url).netloc
 
         try:
-            self.auth = Token(credential)(
+            self.auth = Token(credential_name)(
                 instructions=token_instructions)['token']
```
bpoldrack, line 161: Isn't `token_instructions` undefined now?

bpoldrack: As in: Should be `auth_info`, right?

mih: Yes, correct. Interesting test setup that tolerates such problems.

bpoldrack: Amended accordingly.

bpoldrack (approved):

yarikoptic:
since just factoring out, and minor, I am ok to have it in maint.  I would prefer to have the same decoration as in https://github.com/datalad/datalad/pull/6618/files#diff-0fd2c7ca659a7033dd85d2f12232b418ac44effa3514aa2fa98e2ebdb47ad97dR405  so we have such code paths annotated.  Let's proceed with merge for now, and also see #6618 to go against maint and add annotation to this RF and release then?

PR #6631, opened 2022-04-19
MNT: Require importlib-metadata >=3.6 for Python < 3.10 for entry_points taking kwargs

effigies:
Datalad now depends on entry_points taking kwargs, introduced in https://github.com/python/importlib_metadata/commit/8320adef797d5f14d9fff7b58ebc2a31a2a6a437, first released in importlib-metadata 3.6.

Apologies for ignoring your PR template, but it's unclear where this fits in your changelog set.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,7 +29,7 @@ requires = {
         'chardet>=3.0.4',      # rarely used but small/omnipresent
         'colorama; platform_system=="Windows"',
         'distro; python_version >= "3.8"',
-        'importlib-metadata; python_version < "3.10"',
+        'importlib-metadata >=3.6; python_version < "3.10"',
         'iso8601',
         'humanize',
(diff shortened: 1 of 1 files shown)


effigies:
> **THANK YOU** @effigies . Would you be so kind to rebase it against `maint`? It should be included in the patch release.

Should have read the CONTRIBUTORS... I feel like there needs to be a tl;dr for those things.

yarikoptic:
Just contribute more often and it will be in your fingertips ;)

effigies:
I can't see an obvious connection between this patch and the test failures.

yarikoptic:
yeah, that mac snapshot (likely) test failure is known... we just thought to wait for a new release of annex but might just need to tune up our skips.  Let's proceed. We might kick out maint release in upcoming days.

PR #6635, opened 2022-04-21
Demote (to level 1 from DEBUG) and speed-up API doc logging (parseParameters)

mih:
Previously, log messages like this would appear at DEBUG level:

```
[DEBUG] parseParameters: Given "exclude_special_remotes: bool, optional
  if True, don't return annex special remotes", we split into [('exclude_special_remotes', "exclude_special_remotes: bool, optional\n  if True, don't return annex special remotes")]
```

These are very verbose and large in number, making debug
output needlessly complicated to read.

Moreover, they log the outcome of a trivial operation that practically
never fails in production code, at a comparatively high level (DEBUG),
and they did it in an expensive way that actually built the log string
unconditionally of the effective log level.

This change demote the log level to the lowest possible, and uses a proper
logging pattern that avoid needless string operations.

### Changelog
Not needed

### DIFF
--- a/datalad/dochelpers.py
+++ b/datalad/dochelpers.py
@@ -173,8 +173,11 @@ def _parse_parameters(paramdoc):
     entries = __re_spliter1.split(paramdoc)
     result = [(__re_spliter2.split(e)[0].strip(), e)
               for e in entries if e != '']
-    lgr.debug('parseParameters: Given "%s", we split into %s' %
-              (paramdoc, result))
+    lgr.log(
+        1,
+        'parseParameters: Given "%s", we split into %s',
+        paramdoc, result,
+    )
     return result
 
 


datalad/dochelpers.py
```diff
@@ -173,8 +173,11 @@ def _parse_parameters(paramdoc):
     entries = __re_spliter1.split(paramdoc)
     result = [(__re_spliter2.split(e)[0].strip(), e)
               for e in entries if e != '']
-    lgr.debug('parseParameters: Given "%s", we split into %s' %
-              (paramdoc, result))
+    lgr.log(
+        1,
```
yarikoptic, line 177: I remember there was some interaction between `1` and "something else" . That is why we hardly ever have logging at that low level

<details>
<summary></summary> 

```shell
$> git grep -A1 '\.log($'
datalad/cmd.py:                lgr.log(
datalad/cmd.py-                    5,
--
datalad/core/local/repo.py:            lgr.log(
datalad/core/local/repo.py-                5,
--
datalad/runner/coreprotocols.py:            lgr.log(
datalad/runner/coreprotocols.py-                5,
--
datalad/runner/protocol.py:        lgr.log(
datalad/runner/protocol.py-            8,
--
datalad/runner/runnerthreads.py:        lgr.log(
datalad/runner/runnerthreads.py-            5,
--
datalad/support/sshconnector.py:            lgr.log(
datalad/support/sshconnector.py-                5,

$> git grep '\.log(1'    
datalad/dataset/gitrepo.py:        lgr.log(1, "Finalizer called on: GitRepo(%s)", path)
datalad/support/annexrepo.py:        lgr.log(1, "Finalizer called on: AnnexRepo(%s)", path)

```
</details>

but IIRC it was relating to custom special remotes and custom implementation of those is gone, so might no longer be pertinent.  I personally would have not made it `1` but something like 5 (HEAVYDEBUG) but I also do not care enough.  Let's proceed.  I will release `maint` with this

PR #6648, opened 2022-04-25
RF: get rid of asyncio import by defining few noops from SubprocessProtocol

### LINKED ISSUE #6645: Avoid `import asyncio`
This is the only place outside the tests where it is imported:

```py
datalad/runner/protocol.py:import asyncio
```

This is a prominent place, pretty much runs for any functionality, hence the cost is unconditional, and amounts to ~17ms on my machine, according to `python -X importtimes`.

According to the design doc, this is only for the Protocol interface definition. Can you confirm that @christian-monch ?

Given that we do not aim for any explicit interoperability with `asyncio`, I'd say this is a substantial cost, and we should aim for defining the interface independently from `asyncio`.

yarikoptic:
Closes #6645. Alternative to heavier #6647

#### 🏠 Internal
- Get rid of `asyncio` import by defining few noops methods from asyncio.protocols.SubprocessProtocol
   directly in `WitlessProtocol`.

### DIFF
--- a/datalad/runner/protocol.py
+++ b/datalad/runner/protocol.py
@@ -9,7 +9,6 @@
 """Base class of a protocol to be used with the DataLad runner
 """
 
-import asyncio
 import logging
 import warnings
 from collections import deque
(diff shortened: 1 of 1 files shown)


yarikoptic:
fail on OSX seems to be a new flaky one -- filed https://github.com/datalad/datalad/issues/6653

PR #6273, opened 2021-12-01
Minimalistic (adapters, no asserts changes etc) conversion from nose to pytest

jwodder:
This PR converts the module-wide fixtures in `datalad/__init__.py` and the code in `datalad/tests/utils.py` and `datalad/tests/test_tests_utils.py` to use pytest instead of nose, as instructed [here](https://github.com/datalad/datalad/issues/6225#issuecomment-975946834) and in following comments.  The conversion can be confirmed to work by running `tox -- -k test_tests_utils`.

Though this PR is finished in the sense of doing everything it set out to do, it should not be merged in its current state, and so I am marking it as a draft.

### All tests conversion 

mark the one you want to work on, add your name(s), and cross out when test module is done

`find datalad -type d -iname tests | while read f; do echo "- [ ] $f"; done`

- [x] all converted etc, some are marked as xfail and need to be dealt with here or later

### Notes

- @yarikoptic Have pushed some automated `sed`ing changes (DATALAD RUNCMD commits for details) -- there could be some false positives or incomplete changes.
- At large it is refactoring of parametric tests (from nose's `yield` generators to pytest.mark.parametrize). Even though pytest does still support them, they are being deprecated so we better RF them away while at it (our pytest setup would error on discovering those parametric tests in a file)

### Procedure/Extra TODOs

- expand `TESTS_TO_PERFORM` in `.travis.yml` with individual modules for that sub-package, and whenever entire sub-package is ready -- just leave full subpackage
- [x] ensure no `nose` imports or dependency
- [x] after all sub-packages ready -- bring back .appveyor.yml and all the workflows under .github/workflows
- (~~ideally, but could be~~ postponed) address added `xfails` which were not there - dedicated https://github.com/datalad/datalad/issues/6662
- [ ] check that coverage remains the same and check for places where dropped -- may be tests did not run etc
   - codecov isn't happy with us: sent support request https://codecov.freshdesk.com/support/tickets/6481 (ticket was closed without any reply/support :-/)
- [x] adjust https://github.com/datalad/datalad/blob/HEAD/CONTRIBUTING.md#tests  and provide summary instructions for similar migrations (e.g. in extensions)
- [x] Changes for/to datalad-extension-template and existing extensions
  - :heart: **Choice 1**: strive to maintain no immediate effect on extensions:  
[...]

PR #6660, opened 2022-04-27
OPT: AnnexJsonProtocol - avoid dragging possibly long data around

yarikoptic:
Identified this place while troubleshooting "out of memory" primarily due to pytest also swallowing all logs, and thus our "DATALAD_LOG_LEVEL=2" run on travis was causing out of memory for some tests. Using the --memray pytest plugin, identified this line of code where up to 1GB was kept at some point in time, and realized that code would hold to it in addition to lines because there was that check inside the loop.  So I have moved out the condition and explicitly `del data`.  Shouldn't hurt and shouldn't consume twice more RAM (`lines` + `data`) for long period of time in some cases ;)

CPed from #6273

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -3684,13 +3684,15 @@ class AnnexJsonProtocol(WitlessProtocol):
         # json_loads() is already logging any error, which is OK, because
         # under no circumstances we would expect broken JSON
         lines = data.splitlines()
+        data_ends_with_eol = data.endswith(os.linesep.encode())
+        del data
         for iline, line in enumerate(lines):
             try:
                 j = json_loads(line)
             except Exception as exc:
                 if line.strip():
                     # do not complain on empty lines
-                    if iline == len(lines) - 1 and not data.endswith(os.linesep.encode()):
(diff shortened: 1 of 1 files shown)


yarikoptic:
fails on travis (only!) are curious:
```
FAIL: datalad.tests.test_version.test__version__
  File "/tmp/dl-miniconda-s9q2xzkg/lib/python3.9/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/build/datalad/datalad/datalad/tests/test_version.py", line 61, in test__version__
    ok_startswith(__version__, changelog_version)
  File "/home/travis/build/datalad/datalad/datalad/tests/utils.py", line 451, in ok_startswith
    ok_(s.startswith(prefix),
AssertionError: String '0.16.1+27.g42d6384' doesn't start with '0.16.2'
```
so it is like if it didn't fetch 0.16.2 tag, and thus getting based on 0.16.1.  Not sure why is that so, but it is unrelated.  Since patch is trivial, definetely no the cause of the above, and IMHO worthwhile -- merging.

PR #6664, opened 2022-04-29
NF: Teach WTF to report the filesystem of CWD, TMP, and HOME

### LINKED ISSUE #6294: WTF: add information about the filesystem of CWD etc.
Many issues we run into (e.g. freshish https://github.com/ReproNim/containers/issues/62) with git-annex stem from "exotic" filesystems, often used on HPC, and which could be proprietary and cannot be setup within our CI to be tested against.  When users report problems, and even share WTF information, it requires inquiring more to discover specific of the filesystem in question and possibly relevant details (e.g. mount options).

I think it would be great if we added information about file systems (of $HOME, $TMPDIR, and $CWD) into WTF. Not sure how it could be done in reliable/consistent way, but it could be done quite "easily" in ad-hoc fashion at least for Linux and OSX by parsing output of `mount` command which provides consistent'ish output AFAIK. E.g.  on Linux looks like

```
$> mount | grep /home
/dev/md16 on /home type btrfs (rw,noatime,compress=lzo,space_cache=v2,subvolid=257,subvol=/home)
```
and on OSX
```
yoh@datalads-imac2 ~ % mount
/dev/disk1s5s1 on / (apfs, sealed, local, read-only, journaled)
devfs on /dev (devfs, local, nobrowse)
/dev/disk1s4 on /System/Volumes/VM (apfs, local, noexec, journaled, noatime, nobrowse)
/dev/disk1s2 on /System/Volumes/Preboot (apfs, local, journaled, nobrowse)
/dev/disk1s6 on /System/Volumes/Update (apfs, local, journaled, nobrowse)
/dev/disk1s1 on /System/Volumes/Data (apfs, local, journaled, nobrowse)
map auto_home on /System/Volumes/Data/home (autofs, automounted, nobrowse)
/dev/disk1s5 on /System/Volumes/Update/mnt1 (apfs, sealed, local, journaled, nobrowse)
```
which is the very similar (noted absence of leading `/` in device path, but that is unrelated as to information we care about) format as on FreeBSD, e.g.

```
$ mount | grep BIDS
inbox/BIDS on /inbox/BIDS (zfs, NFS exported, local, nfsv4acls)
```

So they are different "formats" and would require different regexes/parsers to parse. We can even just go through a set of regexes we know, and thus possibly make it just work even if `mount` comes from different implementations (i.e. not make it platform specific in deciding which parser to use).

adswa:
This PR teaches ``datalad wtf`` to report the file systems of ``TMP``, ``CWD``, and ``HOME`` as suggested in  #6294 using ``psutil``. On my Debian system, the output looks like this

```
(handbook) adina@muninn in ~/Documents
❱ datalad wtf -S system
# WTF
## system 
  - distribution: debian/11-updates/n/a
  - encoding: 
    - default: utf-8
    - filesystem: utf-8
    - locale.prefered: UTF-8
  - filesystem: 
    - /home/adina/Documents: ext4
    - /tmp: ext4
    - /home/adina: ext4
  - max_path_length: 277
  - name: Linux
  - release: 5.16.0-3-amd64
  - type: posix
  - version: #1 SMP PREEMPT Debian 5.16.11-1 (2022-02-25)
```

On my Windows machine it looks like this:

```
C:\Users\adina\repos\datalad>datalad wtf -S system
# WTF
## system
  - distribution:   10/10.0.19044/SP0/Multiprocessor Free
  - encoding:
    - default: utf-8
    - filesystem: utf-8
    - locale.prefered: cp1252
  - filesystem:
    - C:\Users\adina\repos\datalad: NTFS
    - C:\Users\adina\AppData\Local\Temp: NTFS
    - C:\Users\adina: NTFS
  - max_path_length: 284
  - name: Windows
  - release: 10
  - type: nt
  - version: 10.0.19044
```

We might need to consider moving ``psutil`` from a development requirement into a standard requirement if we want to ensure this is reported everywhere (if it isn't installed, the file system info is an empty tuple). However, I've read that psutil is quite "heavy", so I wanted to discuss this here first.

Fixes #6294

- ``datalad wtf`` includes a report on file system types of commonly used paths

### DIFF
--- a/datalad/local/wtf.py
+++ b/datalad/local/wtf.py
@@ -140,24 +140,31 @@ def _describe_system():
         'max_path_length': get_max_path_length(getpwd()),
         'encoding': get_encoding_info(),
         'filesystem': {l: _get_fs_type(l, p) for l, p in
-                       [('CWD', os.getcwd()),
-                        ('TMP', tempfile.gettempdir()),
-                        ('HOME', str(Path.home()))]}
+                       [('CWD', Path.cwd()),
+                        ('TMP', Path(tempfile.gettempdir())),
+                        ('HOME', Path.home())]}
     }
 
 
 def _get_fs_type(loc, path):
+    """Return file system info for given Paths. Provide pathlib path as input"""
     res = {'path': path}
     try:
         from psutil import disk_partitions
-        match = ""
-        fs = ""
-        for part in disk_partitions():
-            if path.startswith(part.mountpoint) and \
-                    len(match) < len(part.mountpoint):
-                fs = part.fstype
-                match = part.mountpoint
-        res['type'] = fs
+        parts = {Path(p.mountpoint): p  for p in disk_partitions()}
+        match = None
+        for mp in parts:
+            # if the mountpoint is the test path or its parent
+            # take it, whenever there is no match, or a longer match
+            if (mp == path or mp in path.parents) and (
+                    match is None or len(p.parents) > len(match.parents)):
+                match = mp
+        match = parts[match]
+        for sattr, tattr in (('fstype', 'type'),
+                             ('maxpath', 'max_pathlength'),
+                             ('opts', 'mount_opts')):
+            if hasattr(match, sattr):
+                res[tattr] = getattr(match, sattr)
     except Exception as exc:
         ce = CapturedException(exc)
         # if an exception occurs, leave out the fs type. The result renderer can
--- a/datalad/local/wtf.py
+++ b/datalad/local/wtf.py
@@ -14,6 +14,7 @@ import logging
 import os
 import os.path as op
 import sys
+import tempfile
 from functools import partial
 from collections import OrderedDict
 
@@ -138,7 +139,27 @@ def _describe_system():
                                   _t2s(pl.win32_ver())]).rstrip(),
         'max_path_length': get_max_path_length(getpwd()),
         'encoding': get_encoding_info(),
+        'filesystem': [_get_fs_type(p) for p in [os.getcwd(),
+                                                 tempfile.gettempdir(),
+                                                 str(Path.home())]]
     }
+
+
+def _get_fs_type(path):
+    try:
+        from psutil import disk_partitions
+        match = ""
+        fs = ""
+        for part in disk_partitions():
+            if path.startswith(part.mountpoint) and \
+                    len(match) < len(part.mountpoint):
+                fs = part.fstype
+                match = part.mountpoint
+    except Exception as exc:
+        ce = CapturedException(exc)
+        lgr.warning("Failed to get filesystem information: %s", ce)
+        fs = tuple()
+    return f'{path}: {fs}'
 
 
 def _describe_environment():
--- a/datalad/local/wtf.py
+++ b/datalad/local/wtf.py
@@ -512,13 +512,23 @@ class WTF(Interface):
         from datalad.ui import ui
         out = _render_report(res)
         ui.message(out)
-        # if the record lacks file system information, hint at a psutil problem
-        if res['infos']['system']['filesystem']['TMP'].get('type') is None:
-            from datalad.support import ansi_colors
-            ui.message("{}".format(
-                ansi_colors.color_word(
-                    'Hint: Failed to file system information with psutil',
-                    ansi_colors.YELLOW)))
+        # add any necessary hints afterwards
+        maybe_show_hints(res)
+
+
+def maybe_show_hints(res):
+    """Helper to add hints to custom result rendering"""
+    # check for missing file system info
+    # if the system record lacks file system information, hint at a psutil
+    # problem. Query for TMP should be safe, is included in system info
+    from datalad.ui.utils import show_hint
+
+    if 'system' in res.get('infos', {}) and \
+        'type' not in res['infos']['system']['filesystem']['TMP']:
+        try:
+            import psutil
+        except ImportError:
+            show_hint('Hint: install psutil to get filesystem information')
 
 
 def _render_report(res):
--- a/datalad/local/wtf.py
+++ b/datalad/local/wtf.py
@@ -139,13 +139,14 @@ def _describe_system():
                                   _t2s(pl.win32_ver())]).rstrip(),
         'max_path_length': get_max_path_length(getpwd()),
         'encoding': get_encoding_info(),
-        'filesystem': [_get_fs_type(p) for p in [os.getcwd(),
-                                                 tempfile.gettempdir(),
-                                                 str(Path.home())]]
+        'filesystem': {l: _get_fs_type(l, p) for l, p in
+                       [('CWD', os.getcwd()),
+                        ('TMP', tempfile.gettempdir()),
+                        ('HOME', str(Path.home()))]}
     }
 
 
-def _get_fs_type(path):
+def _get_fs_type(loc, path):
     try:
         from psutil import disk_partitions
         match = ""
@@ -159,7 +160,8 @@ def _get_fs_type(path):
         ce = CapturedException(exc)
         lgr.warning("Failed to get filesystem information: %s", ce)
         fs = tuple()
-    return f'{path}: {fs}'
+    return {'path': path,
+            'type': fs}
 
 
 def _describe_environment():
--- a/datalad/local/wtf.py
+++ b/datalad/local/wtf.py
@@ -147,8 +147,7 @@ def _describe_system():
 
 
 def _get_fs_type(loc, path):
-    global fs_hint
-    fs_hint = None
+    res = {'path': path}
     try:
         from psutil import disk_partitions
         match = ""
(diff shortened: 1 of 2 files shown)


datalad/local/wtf.py
```diff
@@ -138,9 +139,29 @@ def _describe_system():
...
+        fs = ""
+        for part in disk_partitions():
+            if path.startswith(part.mountpoint) and \
+                    len(match) < len(part.mountpoint):
+                fs = part.fstype
+                match = part.mountpoint
+    except Exception as exc:
+        ce = CapturedException(exc)
+        lgr.warning("Failed to get filesystem information: %s", ce)
+        fs = tuple()
```
bpoldrack, line 161: How about something like this instead?

```suggestion
        fs = "Install 'psutils' to get filesystem information"
```

bpoldrack (commented):
Nice. I think this doesn't warrant making `psutils` a strict requirement. But it could be pointed out in the record instead of showing an empty entry (see below).

datalad/local/wtf.py
adswa: I'm fine with removing the warning. I initially implemented it to match the behavior when other system information queries caused an exception.

adswa: But I'm unsure what you mean with "infuse the message in the result rendering"?

bpoldrack: > But I'm unsure what you mean with "infuse the message in the result rendering"?

The renderer could - upon encountering such an empty entry - print that message instead. That would leave the result record itself intact as a structure that says "I got nothing on this", while conveying the suggestion to users.

However, I haven't looked into other cases. If you say that's how it's handled elsewhere, too, then this PR may not be the place to introduce a different approach.

mih:
Thx for adding that. I propose a slightly different reporting style, more structure, less stringy ;-)

Status quo is

```json
      "filesystem": [
        "/home/mih/hacking/datalad/git: ext4",
        "/tmp: ext4",
        "/home/mih: ext4"
      ],
```
What about this:

```json
      "filesystem": {
        "CWD": {
          "path": "/home/mih/hacking/datalad/git",
          "type": "ext4"
        },
        "TMP": ...,
        "HOME": ...,
      }
```

When there is no `psutil`, there is no `type` field in the dict. Instead of a warning, I'd be content with a statement documentation.

datalad/local/wtf.py
adswa: Thx! I've pushed one attempt to implement what you suggest in 33036ddcb95ad

adswa:
Thx for the suggestion, @mih, I'm now returning a nested dictionary with location names as keys and fs type and path information.

datalad/local/wtf.py
adswa:
```
(handbook) adina@muninn in ~/Documents
❱ datalad wtf -S system
# WTF
## system 
  - distribution: debian/11-updates/n/a
  - encoding: 
    - default: utf-8
    - filesystem: utf-8
    - locale.prefered: UTF-8
  - filesystem: 
    - CWD: 
      - path: /home/adina/Documents
      - type: ext4
    - HOME: 
      - path: /home/adina
      - type: ext4
    - TMP: 
      - path: /tmp
      - type: ext4
  - max_path_length: 277
  - name: Linux
  - release: 5.16.0-3-amd64
  - type: posix
  - version: #1 SMP PREEMPT Debian 5.16.11-1 (2022-02-25)
(handbook) adina@muninn in ~/Documents
❱ pip uninstall psutil
Found existing installation: psutil 5.9.0
Uninstalling psutil-5.9.0:
  Would remove:
    /home/adina/env/handbook/lib/python3.9/site-packages/psutil-5.9.0.dist-info/*
    /home/adina/env/handbook/lib/python3.9/site-packages/psutil/*
Proceed (y/n)? y 
  Successfully uninstalled psutil-5.9.0
(handbook) adina@muninn in ~/Documents
❱ datalad wtf -S system
# WTF
## system 
  - distribution: debian/11-updates/n/a
  - encoding: 
    - default: utf-8
    - filesystem: utf-8
    - locale.prefered: UTF-8
  - filesystem: 
    - CWD: 
      - path: /home/adina/Documents
      - type: 
    - HOME: 
      - path: /home/adina
      - type: 
    - TMP: 
      - path: /tmp
      - type: 
  - max_path_length: 277
  - name: Linux
  - release: 5.16.0-3-amd64
  - type: posix
  - version: #1 SMP PREEMPT Debian 5.16.11-1 (2022-02-25)
Hint: Install psutils to get file system type information
```

bpoldrack: May be not most elegant solution, but I'm fine with that. We'll need to iron out the system for results, rendering and hints anyway. This is easy to find later, works and seems unlikely to cause trouble due to the global var - so, all good from my POV.
Thanks!

datalad/local/wtf.py
```diff
@@ -481,7 +506,10 @@ def custom_result_renderer(res, **kwargs):
         from datalad.ui import ui
         out = _render_report(res)
         ui.message(out)
-
+        if fs_hint is not None:
```
bpoldrack, line 509: `global fs_hint` is missing in this function (see CI failure).

datalad/local/wtf.py
```diff
@@ -138,9 +139,33 @@ def _describe_system():
...
+    fs_hint = None
+    try:
+        from psutil import disk_partitions
+        match = ""
+        fs = ""
+        for part in disk_partitions():
+            if path.startswith(part.mountpoint) and \
+                    len(match) < len(part.mountpoint):
+                fs = part.fstype
+                match = part.mountpoint
```
yarikoptic, line 160: looking at this trove of information:
```
In [10]: disk_partitions()[0]
Out[10]: sdiskpart(device='/dev/nvme0n1p5', mountpoint='/', fstype='ext4', opts='rw,relatime,errors=remount-ro', maxfile=255, maxpath=4096)
```
I want to say -- **I want it all** (besides `device` may be -- don't think it is relevant).
- `maxfile`/`maxpath` -- we even try to deduce max filename length, could be critical to know "actual" ones known for file system
- `opts` - most of the times I report anything filesystems related to @joeyh - we end up wanting to see the mount options

datalad/local/wtf.py
```diff
@@ -138,9 +139,33 @@ def _describe_system():
...
+def _get_fs_type(loc, path):
+    global fs_hint
+    fs_hint = None
+    try:
+        from psutil import disk_partitions
+        match = ""
+        fs = ""
+        for part in disk_partitions():
+            if path.startswith(part.mountpoint) and \
+                    len(match) < len(part.mountpoint):
```
yarikoptic, line 158: such comparison I think could fail to differentiate `/mnt` and `/mntpoint`... since not working with pathlib Paths here to use  smth like `pathlib.Path(part.mountpoint) in pathlib.Path(path).parents` use our  `path_is_subpath`.

```shell
$> python3 -c 'from datalad.utils import path_is_subpath; print(path_is_subpath("/mnt/subpath","/mnt"))'
True
```

adswa:
Thanks to @mih, the approach to get fs information and provide hints is a bit more robust and a bit more generic now. The output looks like this:

```
❱ datalad wtf -S system
# WTF
## system 
  - distribution: debian/11-updates/n/a
  - encoding: 
    - default: utf-8
    - filesystem: utf-8
    - locale.prefered: UTF-8
  - filesystem: 
    - CWD: 
      - max_pathlength: 4096
      - mount_opts: rw,relatime,errors=remount-ro
      - path: /home/adina/repos/datalad
      - type: ext4
    - HOME: 
      - max_pathlength: 4096
      - mount_opts: rw,relatime,errors=remount-ro
      - path: /home/adina
      - type: ext4
    - TMP: 
      - max_pathlength: 4096
      - mount_opts: rw,relatime,errors=remount-ro
      - path: /tmp
      - type: ext4
  - max_path_length: 281
  - name: Linux
  - release: 5.16.0-3-amd64
  - type: posix
  - version: #1 SMP PREEMPT Debian 5.16.11-1 (2022-02-25)
```
It includes @yarikoptic 's desired extra infos on pathlength and mount_opts.
Without `psutil` you get

```
❱ pip uninstall psutil
Found existing installation: psutil 5.9.0
Uninstalling psutil-5.9.0:
  Would remove:
    /home/adina/env/handbook/lib/python3.9/site-packages/psutil-5.9.0.dist-info/*
    /home/adina/env/handbook/lib/python3.9/site-packages/psutil/*
Proceed (y/n)? y
  Successfully uninstalled psutil-5.9.0
(handbook) adina@muninn in ~/repos/datalad on git:wtf-fs
❱ datalad wtf -S system                               
# WTF
## system 
  - distribution: debian/11-updates/n/a
  - encoding: 
    - default: utf-8
    - filesystem: utf-8
    - locale.prefered: UTF-8
  - filesystem: 
    - CWD: 
      - path: /home/adina/repos/datalad
    - HOME: 
      - path: /home/adina
    - TMP: 
      - path: /tmp
  - max_path_length: 281
  - name: Linux
  - release: 5.16.0-3-amd64
  - type: posix
  - version: #1 SMP PREEMPT Debian 5.16.11-1 (2022-02-25)
Hint: install psutil to get filesystem information
```

The hint is only rendered if we can confirm that a lack of psutil is the cause for failing to retrieve fs information. If other things go wrong, there is no hint and the captured exception would need to be consulted.

bpoldrack (approved):

bpoldrack:
Ok, no objections, clear improvement - let's go. Thanks, @adswa !

PR #6663, opened 2022-04-29
BF: Sanitize keys before checking content availability

adswa:
In #6630 I noticed wrong reporting of ``datalad status --annex all`` for https://github.com/datalad-datasets/machinelearning-books - even though all content was retrieved, only a fraction of it was reported as being locally available by DataLad. git-annex' own reports were fine, and matched file system reports.

```
adina@muninn in /tmp/machinelearning-books on git:master
❱ git annex info
trusted repositories: 0
semitrusted repositories: 8
	00000000-0000-0000-0000-000000000001 -- web
 	00000000-0000-0000-0000-000000000002 -- bittorrent
 	51d4ba52-edfc-4505-bd36-21233a06f8cb -- yoh@lena:~/proj/datalad/datasets/machinelearning-books
 	7f1346f6-a5dd-485d-96dc-563d0a221a95 -- yoh@hopa:/tmp/mlbooks
 	c6bcd625-0a68-49bb-acd5-51c28e2e3073 -- yoh@hopa:~/proj/datalad/repos/mlbooks
 	d5f231e1-6901-456a-9398-39299242baf6 -- mih@meiner:/tmp/machinelearning-books
 	dcfeee85-513e-43fb-bbac-cbbeb8ccad30 -- adina@muninn:/tmp/machinelearning-books [here]
 	fb0fcdc2-d42d-44e9-beac-74c684962ddd -- adina@muninn:/tmp/machinelearning-books
untrusted repositories: 0
transfers in progress: none
available local disk space: 156.44 gigabytes (+1 megabyte reserved)
local annex keys: 10
local annex size: 77.97 megabytes (+ 1 unknown size)
annexed files in working tree: 10
size of annexed files in working tree: 77.97 megabytes (+ 1 unknown size)
bloom filter size: 32 mebibytes (0% full)
backend usage: 
	MD5E: 2
	URL: 8
adina@muninn in /tmp/machinelearning-books on git:master
❱ datalad status --annex all
9 annex'd files (28.8 MB/74.4 MB present/total size)
nothing to save, working tree clean
adina@muninn in /tmp/machinelearning-books on git:master
❱ du -sh .       
78M	.
adina@muninn in /tmp/machinelearning-books on git:master
❱ du -shLc *.pdf
684K	A.Shashua-Introduction_to_Machine_Learning.pdf
8.5M	B.Efron_T.Hastie-Computer_Age_Statistical_Inference.pdf
[...]

PR #6665, opened 2022-04-29
Sanitize keys before checking content availability to ensure correct value for keys with URL or custom backend

yarikoptic:
CP #6663 to `maint`


adswa:
argh, thx @yarikoptic - completely forgot to base it against maint.

bpoldrack:
Ping @yarikoptic: Failures unrelated to the code changed here, but they are "real".

bpoldrack:
Restarted failing builds. Should pass now.

PR #6666, opened 2022-04-30
Make `push` avoid refspec handling for special remote push targets

### LINKED ISSUE #6657: `push.py:_push()` unconditionally calls `git-push --dry-run` even for special remotes
This leads to

```
[DEBUG  ] Run ['git', '-c', 'diff.ignoreSubmodules=none', 'push', '--progress', '--porcelain', '--dry-run', 'fz-juelich.sciebo.de-storage'] (cwd=/tmp/dum2)                                                                                                                                                    
[DEBUG  ] Non-progress stderr: b"fatal: 'fz-juelich.sciebo.de-storage' does not appear to be a git repository\n"                                                                                                                                                                                               
[DEBUG  ] Non-progress stderr: b'fatal: Could not read from remote repository.\n'                                                                                                                                                                                                                              
[DEBUG  ] Non-progress stderr: b'\n'                                                                                                                                                                                                                                                                           
[DEBUG  ] Non-progress stderr: b'Please make sure you have the correct access rights\n'                                                                                                                                                                                                                        
[DEBUG  ] Non-progress stderr: b'and the repository exists.\n'                                                                                                                                                                                                                                                 
[DEBUG  ] Finished ['git', '-c', 'diff.ignoreSubmodules=none', 'push', '--progress', '--porcelain', '--dry-run', 'fz-juelich.sciebo.de-storage'] with status 128                                                                                                                                               
[DEBUG  ] Dry-run push to check push configuration failed, assume no configuration: CommandError: 'git -c diff.ignoreSubmodules=none push --progress --porcelain --dry-run fz-juelich.sciebo.de-storage' failed with exitcode 128 under /tmp/dum2 [err: 'fatal: 'fz-juelich.sciebo.de-storage' does not appear to be a git repository                                                                                                                                                                                                                                                                                         

and the repository exists.'] 
```

It does not imply much slowdown (10-15ms), but maybe it can be avoided.

mih:
This PR refactors `push.py:_push()` to make the flow more intelligible.

Importantly, it makes one behavior change, which is desirable IMHO. Instead of rejecting to git-push any refspec for a repo with a detached HEAD, it will attempt to push a `git-annex` branch for an `AnnexRepo`. The respective test that ensured this behavior beyond the particular conditions the original problem occurred in was adjusted accordingly.

#### 🐛 Bug Fixes
- `push` now only evaluates `refspecs` to push to a remote when the target remote is actually a Git remote. Fixes #6657

### DIFF
--- a/datalad/core/distributed/push.py
+++ b/datalad/core/distributed/push.py
@@ -439,56 +439,24 @@ def _push(dspath, content, target, data, force, jobs, res_kwargs, pbars,
         total=4,
     )
 
+    #
+    # First we must figure out where to push to, if needed
+    #
+
     # will contain info to determine what refspecs need to be pushed
     # and to which remote, if none is given
     wannabe_gitpush = None
     # pristine input arg
     _target = target
-    # verified or auto-detected
-    target = None
-    if not _target:
-        # let Git figure out what needs doing
-        # we will reuse the result further down again, so nothing is wasted
-        wannabe_gitpush = _get_push_dryrun(repo)
-        # we did not get an explicit push target, get it from Git
-        target = set(p.get('remote', None) for p in wannabe_gitpush)
-        # handle case where a pushinfo record did not have a 'remote'
-        # property -- should not happen, but be robust
-        target.discard(None)
-        if not len(target):
-            yield dict(
-                res_kwargs,
-                status='impossible',
-                message='No push target given, and none could be '
-                        'auto-detected, please specify via --to',
-            )
-            return
-        elif len(target) > 1:
-            # dunno if this can ever happen, but if it does, report
-            # nicely
-            yield dict(
-                res_kwargs,
-                status='error',
-                message=(
-                    'No push target given, '
-                    'multiple candidates auto-detected: %s',
-                    list(target),
-                )
-            )
-            return
-        else:
-            # can only be a single one at this point
-            target = target.pop()
-
-    if not target:
-        if _target not in repo.get_remotes():
-            yield dict(
-                res_kwargs,
-                status='error',
-                message=(
-                    "Unknown target sibling '%s'.", _target))
-            return
-        target = _target
+    # verified or auto-detected target sibling name
+    target, status, message, wannabe_gitpush = _get_push_target(repo, target)
+    if target is None:
+        yield dict(
+            res_kwargs,
+            status=status,
+            message=message,
+        )
+        return
 
     log_progress(
         lgr.info, pbar_id, "Push refspecs",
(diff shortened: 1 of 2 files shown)


datalad/core/distributed/push.py
```diff
@@ -572,19 +489,28 @@ def _push(dspath, content, target, data, force, jobs, res_kwargs, pbars,
...
-    # to know better than any git config
-    must_have_branches = [active_branch] if active_branch else []
-    if is_annex_repo:
-        must_have_branches.append('git-annex')
-    for branch in must_have_branches:
-        _append_branch_to_refspec_if_needed(ds, refspecs2push, branch)
+    #
+    # We know where to push to, honor dependencies
+    # XXX we could do this right after we know the value of `target`,
+    # but this owuld mean we would also push to dependencies
```
bpoldrack, line 495:
```suggestion
    # but this would mean we would also push to dependencies
```

datalad/core/distributed/push.py
```diff
@@ -572,19 +489,28 @@ def _push(dspath, content, target, data, force, jobs, res_kwargs, pbars,
...
+    # even when no actual push to the primary target is needed
```
bpoldrack, line 496: > but this would mean we would also push to dependencies even when no actual push to the primary target is needed

But wouldn't that be desired? If I configure a publication dependency, isn't the aim to have both targets "in sync"? The dependency could require a push to make it so even if the primary target doesn't. May be because I only just added the dependency and now want to make sure the secondary target catches up?

To be sure, just confirmed, that this would currently work as I described.

datalad/core/distributed/push.py
```diff
@@ -925,3 +808,212 @@ def _get_corresponding_remote_state(repo, to):
...
             since = '%s/%s' % (tracked_remote, tracked_refspec)
     return since
+
+
+def _get_push_dryrun(repo, remote=None):
+    """
+    Returns
+    -------
+    list
+      The result of the dry-run. Will be an empty list of the dry-run
```
bpoldrack, line 818:
```suggestion
      The result of the dry-run. Will be an empty list if the dry-run
```

bpoldrack (commented):
Overall looks good. However, besides typos left a comment where I am not sure about desirable behavior.

datalad/core/distributed/push.py
mih: I am not sure what the question is. Yes, it is desired. But it has nothing to do with this PR. The comment is about making more obvious that the structure of the function could be further simplified, but a potential behavior change is implied -- but again, nothing that this PR (against maint!) wants to change.

datalad/core/distributed/push.py
```diff
@@ -438,131 +438,48 @@ def _push(dspath, content, target, data, force, jobs, res_kwargs, pbars,
...
-    if not refspecs2push and not active_branch:
-        # nothing was set up for push, and we have no active branch
-        # this is a weird one, let's confess and stop here
-        # I don't think we need to support such a scenario
-        if not active_branch:
+    # handling pure special remotes is a lot simpler
+    target_is_git_remote = repo.config.get(
+        f'remote.{target}.url', None) is not None
+    # TODO would is be useful to also check whether the
+    # target is set 'annex-ignore' right here?
```
yarikoptic, line 471: not sure... it could be a git remote with annex-ignore set (github) or not (ssh) -- doesn't change the fact that it is a  remote git can interact with (I guess).  If I got it right:
```suggestion
```

yarikoptic (commented):
Started to review but ran out of time
- would have been easier to review if there was a separate RF (code moving) PR easy to review/merge
- actual bug fix PR (smaller diff than here) with a dedicated test which would have demonstrated what is being fixed.

given the number of commits (even though nicely annotated) makes it hard to get a "holistic" view of the diff online. will try again later with local gitk/git

mih:
Closing. Seems to be too complex for `maint` and raises conceptual questions. Will move this change to an extension to get the desired speed-up in the context where it is needed most.

PR #6667, opened 2022-05-02
Deprecate AnnexRepo.add_urls()

### LINKED ISSUE #5153: Why do we have `AnnexRepo.add_urls()`?
This is what it advertises:

```
    def add_urls(self, urls, options=None, backend=None, cwd=None,
                 jobs=None,
                 git_options=None, annex_options=None):
        """Downloads each url to its own file, which is added to the annex.

        Parameters
        ----------
        urls: list of str
        options: list, optional
            options to the annex command
        cwd: string, optional
            working directory from within which to invoke git-annex
        """
```

but we do have `add_url_to_file()` too:

```
    def add_url_to_file(self, file_, url, options=None, backend=None,
                        batch=False, git_options=None, annex_options=None,
                        unlink_existing=False):
        """Add file from url to the annex.

        Downloads `file` from `url` and add it to the annex.
        If annex knows `file` already,
        records that it can be downloaded from `url`.

        Note: Consider using the higher-level `download_url` instead.

        Parameters
        ----------
        file_: str
        url: str
        options: list
            options to the annex command
        batch: bool, optional
            initiate or continue with a batched run of annex addurl, instead of just
            calling a single git annex addurl command
        unlink_existing: bool, optional
            by default crashes if file already exists and is under git.
            With this flag set to True would first remove it.

        Returns
        -------
        dict
          In batch mode only ATM returns dict representation of json output returned
          by annex
        """
```

which seems to be a superset of `add_urls()`. Also `add_urls()` is not used outside the tests. ``add_url_to_file()` is used in:

- `add_archive_content`
- `download_url`
- `addurls`

Both functions score equal on advertising options that are not implemented and just lead to warnings.

I'd say we should get rid of `AnnexRepo.add_urls()`.

This was prompted by the discovery in #5152 that this is pretty much the last annex run function to go through a non-witless execution path.

mih:
This function contains more warnings which parts of the promised API do not actually work, than actual code.

- `AnnexRepo.add_urls()` is deprecated in favor of `AnnexRepo.add_url_to_file()` or a direct call to `AnnexRepo.call_annex()`. Fixes #5153

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -2188,6 +2188,9 @@ class AnnexRepo(GitRepo, RepoInterface):
                  git_options=None, annex_options=None):
         """Downloads each url to its own file, which is added to the annex.
 
+        .. deprecated:: 0.17
+            Use add_url_to_file() or call_annex() instead.
+
         Parameters
         ----------
         urls: list of str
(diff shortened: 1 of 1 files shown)


bpoldrack:
Thanks.
Failures unrelated and change not controversial, I think. Merging.

PR #6668, opened 2022-05-02
Remove deprecated `uninstall` shim

### LINKED ISSUE #4594: Strip handle_dirty_dataset() and if_dirty_opt
Still (and only) used in `remove/uninstall/drop`. This is pre `save()`, and a relict left behind after #1175 was closed 3+ years ago.

mih:
- The deprecated `uninstall` command was removed. The `drop` command has learned to safely drop (sub)datasets with DataLad 0.16.
- The long obsolete `handle_dirty_dataset()` helper function was removed. Fixes #4594


datalad/distribution/uninstall.py
```diff
@@ -1,157 +0,0 @@
...
-    doc="""whether to perform checks to assure the configured minimum
-    number (remote) source for data.[CMD:  Give this
-    option to skip checks CMD]""",
-    action="store_false",
-    dest='check')
-
-
-@build_doc
-class Uninstall(Interface):
-    """DEPRECATED: use the `drop` command"""
```
yarikoptic, line 67: I wondered when was it deprecated... Had to consult changelog, seems in
0.16.0 (Fr Apr 8 2022), so not even a month ago . I wonder now if removing not too early?

mih: Ok, closing. Please let me know when a better time would be.

PR #6669, opened 2022-05-02
Remove `datalad test`

### LINKED ISSUE #6651: datalad test -- remove or RF
a follow up to https://github.com/datalad/datalad/pull/6273 : we still have `datalad test` and `datalad/support/third/nose*.py`

mih:
Thanks @jwodder for implementing the removal. I am proposing this here in a dedicated PR alongside other removals.

The removal of `test`also reduces the list of ancient, neglected command a bit further #4593

- Remove the obsolete and outdated helper command `test` that could be used to execute some of the unit test battery. It interfaced the `nose` package, which will be replaced in a future release. Fixes #6651


yarikoptic:
I guess this PR would become obsolete if #6273 is merged first, right?  for this one someone needs to triage CI fails

yarikoptic:
I think this PR is no longer needed since all those actions were done during RF into pytest. I guess github is not complaining since master has exactly those changes in already.  I also spotted one more minor left over which I pushed directly in 84f979ae759d1118f56494529e4fa9a9fd71bc79

Issue #6144, opened 2021-11-05
Fully discontinue the use of `testrepos`


mih:
There is a total of 37 tests that still use this facility. The complete list of files still making use of it is

- [x] `datalad/cmdline/tests/test_helpers.py`
- [x] `datalad/core/distributed/tests/test_clone.py` #6224
- [x] `datalad/core/local/tests/test_run.py` #6224
- [x] `datalad/core/local/tests/test_save.py` #6224
- [x] `datalad/distribution/tests/test_create_sibling.py` #6690
- [x] `datalad/distribution/tests/test_dataset.py` #6670 
- [x] `datalad/distribution/tests/test_get.py` #6674
- [x] `datalad/distribution/tests/test_install.py` #6690
- [x] `datalad/distribution/tests/test_siblings.py` #6671
- [x] `datalad/distribution/tests/test_uninstall.py`
- [x] `datalad/distribution/tests/test_update.py` #6672
- [x] `datalad/interface/tests/test_unlock.py`
- [x] `datalad/support/tests/test_annexrepo.py` done in https://github.com/christian-monch/datalad/tree/rf-testannexrepo
- [x] `datalad/tests/test_testrepos.py` #6690
- [x] `datalad/tests/test_tests_utils.py` #6690

yarikoptic:
Could you remind me why we want to discontinue their use?

mih:
This is not about starting a removal. This issue is just tracking the tail end of a process that started ~2 years ago, when we seriously approached cross-platform operation. Testrepos were established long before proper path handling, managed branch handling, or even URL handling was implemented. Over the past years the discontinuation of this facility has broad cross-platform coverage and faster test.

PR #6670, opened 2022-05-02
Modernize test

mih:
Ping datalad/datalad#6144


yarikoptic:
it is all green, good -- let's proceed! Thanks!

PR #6675, opened 2022-05-03
Use annexremote's LogHandler in special remotes

bpoldrack:
This change adds `annexremote`s LogHandler to all loggers when running
in a special remote process to the effect of sending the logging output
as DEBUG messages on stdout to git-annex.
Note, that this only applies when the special remote's `main` function
is executed, not when instantiating a special remote class.

#### 🐛 Bug Fixes
- Description... Fixes #issue
-
-
#### 📝 Documentation
-
#### 🏠 Internal
-
#### 🛡 Tests
-


bpoldrack:
FTR: No idea how code climate manages to complain about `push` here.

yarikoptic (approved):

Issue #6623, opened 2022-04-09
test_datalad_credential_helper  test wants interaction


yarikoptic:
while building debian package (unattended), next day spotted that it wants to talk to me

```
datalad.local.tests.test_gitcredential.test_datalad_credential_helper ... Please set a password for your new keyring: 
Please confirm the password: 
Error: blank passwords aren't allowed.
Please set a password for your new keyring: 
```

shouldn't happen, and any needed for that settings (if any) changes should be done in setup_package fixture or prevented via mocking in the test.

TheChymera:
Just ran into the exact same issue running the tests for `0.16.1`

```
datalad.local.tests.test_gitcredential.test_datalad_credential_helper ... Please set a password for your new keyring:
```

bpoldrack:
FTR: Didn't look closely yet, but it's likely `GIT_ASKPASS` (or the config `core.askpass`) that need to be set to not let Git ask interactively. Which is currently done in our CI setups. I see the argument to move this to `setup_package`.

yarikoptic:
seting GIT_ASKPASS to true is of no effect:
```
root@smaug:/build/datalad-0.17.0/.pybuild/cpython3_3.9_datalad/build# GIT_ASKPASS=true PYTHONPATH=$PWD PATH=/build/datalad-0.17.0/bin:$PATH python3 -m pytest -s -v datalad/local/tests/test_gitcredential.py::test_datalad_credential_helper
It is highly recommended to configure Git before using DataLad. Set both 'user.name' and 'user.email' configuration variables.
======================================================================================================================= test session starts =======================================================================================================================
platform linux -- Python 3.10.5, pytest-7.1.2, pluggy-1.0.0 -- /usr/bin/python3
cachedir: .pytest_cache
rootdir: /build/datalad-0.17.0, configfile: tox.ini
collected 1 item                                                                                                                                                                                                                                                  

datalad/local/tests/test_gitcredential.py::test_datalad_credential_helper create(ok): . (dataset)
Please set a password for your new keyring: 
```
and we seems to set it already in fixtures:

```shell
$> git grep ASKPASS
datalad/conftest.py:            # variable GIT_ASKPASS overwrites this, but neither env var nor this config
datalad/conftest.py:        m.setenv('GIT_ASKPASS', 'true')
datalad/local/gitcredential.py:        # TODO we could prevent prompting by setting GIT_ASKPASS=true
[...]

PR #6682, opened 2022-05-05
RF special remote base classes

bpoldrack:
This is refactoring our `SpecialRemote` and `AnnexCustomRemote` classes. The latter only serves as base class for the `datalad` and `datalad-archives` special remotes, whereas the former is the more general one. Hence, move functionality that isn't specific to those two.
Two minor fixes along the way:
- Implementation of availability was a bit inconsistent
- Possible failure at the outermost layer wasn't sending an ERROR message to annex, but only logging.

I don't think this needs a changelog.

### DIFF
--- a/datalad/customremotes/main.py
+++ b/datalad/customremotes/main.py
@@ -20,6 +20,9 @@ from datalad.cli.parser import (
 from datalad.cli.utils import setup_exceptionhook
 from datalad.ui import ui
 
+import logging
+lgr = logging.getLogger('datalad.customremotes')
+
 
 def setup_parser(remote_name, description):
     # setup cmdline args parser
@@ -74,6 +77,8 @@ def main(args=None, cls=None, remote_name=None, description=None):
         try:
             _main(args, cls)
         except Exception as exc:
+            lgr.debug('%s (%s) - passing ERROR to git-annex and exiting',
+                      str(exc), exc.__class__.__name__)
             # `SpecialRemote` classes are supposed to catch everything and
             # turn it into a `RemoteError` resulting in an ERROR message to
             # annex. If we end up here, something went wrong outside of the
--- a/datalad/customremotes/main.py
+++ b/datalad/customremotes/main.py
@@ -18,7 +18,6 @@ from datalad.cli.parser import (
     parser_add_version_opt,
 )
 from datalad.cli.utils import setup_exceptionhook
-from datalad.log import lgr
 from datalad.ui import ui
 
 
@@ -75,5 +74,12 @@ def main(args=None, cls=None, remote_name=None, description=None):
         try:
             _main(args, cls)
         except Exception as exc:
-            lgr.error('%s (%s)', str(exc), exc.__class__.__name__)
+            # `SpecialRemote` classes are supposed to catch everything and
+            # turn it into a `RemoteError` resulting in an ERROR message to
+            # annex. If we end up here, something went wrong outside of the
+            # `master.Listen()` call in `_main`.
+            # In any case, exiting the special remote process should be
+            # accompanied by such an ERROR message to annex rather than a log
+            # message.
+            print("ERROR %s (%s)" % (str(exc), exc.__class__.__name__))
             sys.exit(1)
--- a/datalad/customremotes/tests/test_main.py
+++ b/datalad/customremotes/tests/test_main.py
@@ -10,26 +10,31 @@
 
 import logging
 import os
+import pytest
 
 from ..main import main
 
 from ..base import AnnexCustomRemote
 
 from datalad.tests.utils_pytest import (
+    patch,
     swallow_logs,
     swallow_outputs,
 )
-
-import pytest
+from datalad.ui import ui
 
 
 def test_erroring_out():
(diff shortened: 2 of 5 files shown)


datalad/customremotes/main.py
```diff
@@ -75,5 +74,12 @@ def main(args=None, cls=None, remote_name=None, description=None):
...
         except Exception as exc:
-            lgr.error('%s (%s)', str(exc), exc.__class__.__name__)
+            # `SpecialRemote` classes are supposed to catch everything and
+            # turn it into a `RemoteError` resulting in an ERROR message to
+            # annex. If we end up here, something went wrong outside of the
+            # `master.Listen()` call in `_main`.
+            # In any case, exiting the special remote process should be
+            # accompanied by such an ERROR message to annex rather than a log
+            # message.
+            print("ERROR %s (%s)" % (str(exc), exc.__class__.__name__))
```
yarikoptic, line 84: I would say we better have both since if git-annex just swallows that ERROR and might die some other way (may be it was the first one to be killed?) then we might not see this error anywhere... Greedy me would keep the log (it will be the main datalad logger right?) and  channel ERROR to annex.   or am I missing the whole point? ;)

bpoldrack: Well, several things, @yarikoptic:

1. I have never seen annex swallow those. And if it does, that would be something to address in annex from my POV. Special remote protocol explicitly expects this kind of message and considers it a violation to exit non-zero w/o it. If anything, datalad swallows it by not considering stderr of the annex process it calls.
2. Within a datalad context where we don't swallow errors from annex (which should be most cases), this would lead to double reporting again, because what do we do with an error message we got from annex? We log it.

I think having both only makes sense when we finally manage to log to files by default and separate logging from user output.

adswa:
I'm not an expert on special remotes, but overall those changes look little controversial to me. In the interest of progressing with existing PRs so that they do not get forgotten, what is @yarikoptic and @bpoldrack's current verdict about the discussion on whether or not to log in addition to printing errors? I don't feel like I could confidently judge all implications of either solution so I'm secretly hopeful that the weeks in between now and that discussion led to a consensus ;-)

mih (approved):
I think these changes are sane, largely moving code around into more fundamental locations. The introduction of the ERROR message is needed, thx. I personally follow the line of argument presented by @bpoldrack re not logging it, in particular the double-logging avoidance.

datalad/customremotes/main.py
yarikoptic: implemented the promise in 1441891f7f4c163b2fcad234a3288c4ddc738e66 , accompanied with a test for that block (was not unittested at all)

yarikoptic:
it was approved, comments addressed... appveyor though so red, can't tell if anything relates to these PR changes. Clicked "rerun incomplete". Travis for some reason is not present at all!

bpoldrack:
Despite approvals, not merging yet. There are failures where tests start querying interactively, which doesn't happen on master apparently. That would suggest I somehow messed with the detection. Postponing until after 0.17.0

yarikoptic:
FTR, I thought that may be it is https://github.com/datalad/datalad/issues/6623 but it is different test

https://ci.appveyor.com/project/mih/datalad/builds/44097916/job/8ctjnacmj36d10a0

```
../datalad/distributed/tests/test_create_sibling_ghlike.py::test_invalid_call create(ok): . (dataset)
An access token is required for https://gin.g-node.org. Visit https://gin.g-node.org/user/settings/applications to create a token
token: 
```

adswa:
what shall we do with this one? I checked the actions status at all commits in this PR, and they all show the failure. Does anyone remember if that failure occurred before, e.g., when https://github.com/datalad/datalad/commit/f95d1887d337bcfe2877c52f71c74ff677362bb6 was the only commit this PR (I think it was force-pushed over at some point)? I vaguely remember to see a green CI

mih:
Nobody has worked on this for a month. There is virtually no cost to closing a PR to stop exploiting people's attention span. If interest ever rekindles, reopening is a button push. Tag 'stale' and close. No activity for a week or two should be enough of a threshold, because, again, there is no cost. And the CI that would need to rerun is a good thing.

And yes, feel free to close any and all of my PRs that match this too. I would support adding a bot doing this mercilessly.

mih:
This is again inactive for more than a month. Let's close inactive drafts, ok?

datalad/customremotes/tests/test_main.py
```diff
@@ -0,0 +1,39 @@
...
+"""Tests for the "main" driver of the special remotes"""
+
+import logging
+import os
+
+from ..main import main
+
+from ..base import AnnexCustomRemote
+
+from datalad.tests.utils import (
```
bpoldrack, line 18:
```suggestion
from datalad.tests.utils_pytest import (
```

bpoldrack:
Fixed the interactivity issue (it was the introduced test, not an actual code change), rebased and removed the disabling of parallel test execution again.
Only remaining "failures" before that rebase were about `test_add_archive_content` (not actually failing the test, but complaining the test took to long).

PR #6690, opened 2022-05-10
Discontinue usage of testrepos

bpoldrack:
#### 🐛 Bug Fixes
- Description... Fixes #issue
-
-
#### 📝 Documentation
-
#### 🏠 Internal
-
#### 🛡 Tests
-


bpoldrack:
FTR: AppVeyor Failure is on win "no module named coverage". I consider this unrelated and it's after the actual test runs which are fine.

datalad/distribution/tests/test_create_sibling.py
```diff
@@ -37,6 +37,7 @@
 )
 from datalad.support.gitrepo import GitRepo
 from datalad.support.network import urlquote
+from ...tests.utils_testdatasets import _mk_submodule_annex
```
adswa, line 40: why is this import relative, shouldn't it be absolute, or am I missing something? It is also absolute in ``test_get.py`` further down

datalad/distribution/tests/test_install.py
```diff
@@ -36,6 +37,11 @@
     InsufficientArgumentsError,
 )
 from datalad.support.gitrepo import GitRepo
+from datalad.support.network import get_local_file_url
+from ...tests.utils_testdatasets import (
```
adswa, line 41: again relative versus absolute imports here

adswa (commented):
I noticed a few inconsistent relative imports - is that intentional?

datalad/distribution/tests/test_install.py
yarikoptic: even I would agree to use abs imports here. Design doc has it -- use relative based on "semantic" binding of the code

datalad/tests/test_tests_utils.py
```diff
@@ -155,27 +151,6 @@ def t():  # pragma: no cover
         t()
 
 
-def test_with_testrepos():
```
yarikoptic, line 158: I would have kept this test for the duration of having `@with_testrepos` in the codebase.

datalad/tests/test_tests_utils.py
```diff
@@ -186,11 +161,6 @@ def test_get_resolved_values():
     with patch_config({'datalad.tests.nonetwork': '1'}):
         eq_(_get_resolved_flavors(flavors), ['local'])
 
-        # and one more to see the exception being raised if nothing to teston
-        @with_testrepos(flavors=['network'])
-        def magical():
-            raise AssertionError("Must not be ran")
-        assert_raises(SkipTest, magical)
```
yarikoptic, line 193: @yarikoptic would have appreciated to have this tested as long as we have debian packages built for extensions which still use `@with_testrepos`

datalad/tests/test_tests_utils.py
mih: I agree.

datalad/tests/test_tests_utils.py
mih: From my POV the only extension using it is `deprecated`, which would also be the place to migrate this functionality to. @yarikoptic are you aware of other extensions using it?

datalad/distribution/tests/test_create_sibling.py
mih: Done

datalad/distribution/tests/test_install.py
mih: Done

datalad/tests/test_tests_utils.py
mih: Done.

mih:
This has been hanging here for too long, hence I made an attempt to unstuck it be addressing the previous comments: absolute imports, keeping the `with_testrepo` tests.

If the tests pass, this closes #6144 from my POV.

yarikoptic:
I believe my comments were addressed, I have no objections, thank you @bpoldrack - let's proceed and the rest (full deprecation) could be in a follow up PR if so desired

Issue #6617, opened 2022-04-07
8E+7F=15 problems with recent git-annex on windows


yarikoptic:
as reported in testing of fresh snapshot build on https://github.com/datalad/git-annex/runs/5845108035?check_suite_focus=true

<details>
<summary>cut and paste from the full log</summary> 

```shell
2022-04-06T06:17:03.2886770Z ======================================================================
2022-04-06T06:17:03.2887137Z ERROR: datalad.distribution.tests.test_get.test_get_greedy_recurse_subdatasets
2022-04-06T06:17:03.2887546Z ----------------------------------------------------------------------
2022-04-06T06:17:03.2887861Z Traceback (most recent call last):
2022-04-06T06:17:03.2889030Z   File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\site-packages\nose\case.py", line 198, in runTest
2022-04-06T06:17:03.2889646Z     self.test(*self.arg)
2022-04-06T06:17:03.2890239Z   File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\site-packages\datalad\tests\utils.py", line 1157, in _wrap_with_testrepos
2022-04-06T06:17:03.2890748Z     testrepos_uris = _get_testrepos_uris(regex, flavors_)
2022-04-06T06:17:03.2891264Z   File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\site-packages\datalad\tests\utils.py", line 1090, in _get_testrepos_uris
2022-04-06T06:17:03.2891725Z     _basic_annex_test_repo.create()
2022-04-06T06:17:03.2892235Z   File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\site-packages\datalad\tests\utils_testrepos.py", line 85, in create
2022-04-06T06:17:03.2892665Z     self.populate()
2022-04-06T06:17:03.2894010Z   File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\site-packages\datalad\tests\utils_testrepos.py", line 110, in populate
2022-04-06T06:17:03.2895886Z     self.repo.add_url_to_file("test-annex.dat", remote_file_url)
2022-04-06T06:17:03.2896533Z   File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\site-packages\datalad\support\gitrepo.py", line 240, in _wrap_normalize_path
2022-04-06T06:17:03.2897015Z     return func(self, file_new, *args, **kwargs)
[...]

PR #6692, opened 2022-05-11
No change for a PR to trigger release

yarikoptic:


yarikoptic:
will wait for CI just to make sure no surprises, will merge whennever all is ok

yarikoptic:
Actually -- I might replace it with another PR , please don't merge

yarikoptic:
well -- my idea was to just mark all those failing or erroring on windows with recent git-annex (attn @joeyh -- might be regressions) tests in https://github.com/datalad/datalad/issues/6617  as known.  But those errors/fails seems grave enough to not just mark and forget (we have over 30 uses of `@known.*windows`), so although we might miss new extra fails, at least I will be constantly reminded (and will remind all of us) that we need to address those!!!

yarikoptic:
- travis - some apt server issues in 2 runs, restarted
- appveyor. 2 issues with starting OSX,  1 - ERROR of test_ria_postclonecfg. #6599 , commented on.

Anyways -- I think it is good, merging

PR #6693, opened 2022-05-13
BF: require argcomplete version at least 1.12.3 to test/operate correctly

yarikoptic:
On debian stable bullseye and ubuntu 22.04 we have 1.8.1 and test_completion fails.

I will add corresponding version depend also for debian/control in debian branch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,15 @@
 # vim ft=yaml
 # travis-ci.org definition for DataLad build
 language: python
+
+# Do full clone (~10 extra seconds) to fetch all tags.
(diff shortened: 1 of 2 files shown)


bpoldrack (commented):
Looks alright, but can you tell anything about those failures, @yarikoptic ? Just noticed them in #6704, too.

yarikoptic:
ok, mystery resolved and problems averted.  Fails  are AFAIK known on macs... one is the known test_ria_postclonecfg and another one is also something freshish git-annexy I whined about recently.  Let's proceed.

Issue #6700, opened 2022-05-23
create-sibling-github lacks feedback when no dataset can be determined


adswa:
I'm used to the ``create-sibling-ghlike`` command API, and accidentally tried to call ``create-sibling-gitlab``'s ``--project`` parameter without the flag, as a path:

```
# my incorrect invocation:
datalad create-sibling-gitlab --site jugit some
# correct one would be
datalad create-sibling-gitlab --site jugit --project some
```

In case of my incorrect command invocation, ``create-sibling-gitlab`` attempts to create a sibling for the non-existing subdataset ``some`` but returns zero feedback when it doesn't find it - I believe it would be helpful if the command would yield an impossible result instead or a ``NoDatasetFound`` error like ``create-sibling-ghlike`` does when pointed to a wrong path to create a sibling for.

PR #6701, opened 2022-05-23
ENH: Better feedback when create-sibling-gitlab is called with non-existing path

adswa:
When create-sibling-gitlab receives a path to a non-existing or non-dataset location, it does not provide any feedback to the user about this (#6700). This PR makes ``create-sibling-gitlab`` yield an impossible result whenever a provided path does not point to a subdataset:

```
$ datalad create-sibling-gitlab --site https://jugit.fz-juelich.de thisismyproject                    1 !
create_sibling_gitlab(impossible): thisismyproject (dataset) [No dataset found under /tmp/some/thisismyproject]
```

- When ``create-sibling-gitlab`` is called on non-existing subdatasets or paths it now returns an impossible result instead of no feedback at all

### DIFF
--- a/datalad/distributed/create_sibling_gitlab.py
+++ b/datalad/distributed/create_sibling_gitlab.py
@@ -260,7 +260,7 @@ class CreateSiblingGitlab(Interface):
         # to avoid undesired level-1 recursion into subdatasets
         if any(p != ds.pathobj for p in (path or [])) or recursive:
             # also include any matching subdatasets
-            for subds in ds.subdatasets(
+            subds = ds.subdatasets(
                     path=path,
                     # we can only operate on present datasets
                     state='present',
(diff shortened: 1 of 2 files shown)


bpoldrack (commented):
Hm. This change looks to me as if the actual problem is `subdatasets` that should yield an `impossible` or `error` result when passing such a path. Then here that result could just be yielded upwards. WDYT, @adswa? Do I miss something?

adswa:
mhhh, looking into ``subdatasets`` seems to suggest that yielding an impossible or error result when a given path does not match a subdataset is against the way the ``path`` parameter is used, though. Its docstring says ``path/name to query for subdatasets. Defaults to the current directory[PY: , or the entire dataset if called as a dataset method PY]`` and internal logging confirms that it is a constrain for where to search underneath for subdataset, not where specifically a subdataset should be:

```py
        if paths is not None:
            lgr.debug('Query subdatasets underneath paths: %s', paths)
```
Making ``subdatasets()`` yield ``impossible`` would probably be a quite severe API change with mostly bad consequences (I suspect this argument is especially useful in large datasets), so I wouldn't go down that road.

bpoldrack (approved):
Ah, you're right. Apologies. Approved then. :)

adswa:
Thanks for the review and approval, @bpoldrack! Merging, since no-one has complained for a week :)

PR #6705, opened 2022-05-24
ENH: Rudimentary CODEOWNERS file to test assigning code responsibility

yarikoptic:
See https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners
for more info.  I have spotted that file in

https://github.com/w3c-ccg/markdown-to-spec  template

and decided to give it a shot

Suggest your additions etc but might be worth first merging and seeing what it would give us if anything

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,13 @@
+# These owners will be the default owners for everything in
+# the repo. Unless a later match takes precedence,
+# they will be requested for review when someone opens a 
+# pull request.
+*       @datalad/developers
+
+# The Runner
+*cmd.py             christian.moench@web.de
+/datalad/runner/*   christian.moench@web.de
+
+*foreach_dataset.py @yarikoptic
+
+# See CODEOWNERS syntax here: https://help.github.com/articles/about-codeowners/#codeowners-syntax


bpoldrack:
Hm. I don't know. We only recently introduced the team-* labels that are aiming in the same direction. If we want to make use of this feature, may be we should formalize those teams into actual github teams and assign them here instead of invidual people? I think this is otherwise kind of reinforcing the bottlenecks we should get rid of.

yarikoptic:
yeah, indeed... and it seems that issues cannot be assigned to teams, only to individuals which doesn't allow for ultimate "one thing to rule them all".  But also not sure if `team-` labels are the right medium since nohow integrate with announcement/assignment. Do they work out for you?

yarikoptic:
d'oh -- sorry @adswa -- indeed my bad.  The question remains -- either it is worth it and how/if would align with existing `team-` labels and what else to extend with right away?

adswa:
> The question remains -- either it is worth it and how/if would align with existing `team-` labels and what else to extend with right away?

I think it would be an addition to the team labels because its not about fixing issues, but reviewing PRs, and if I understand it correctly, it even comes with automation, right? Automation is especially cool, self-assignment based on teams hasn't worked well so far and automation could help with keeping up.

As for the latter, its definitely easier to suggest code pieces for others than finding code pieces for myself. But I would see @bpoldrack to everything in ria/ora related (``create_sibling_ria``, ``ora_remote``), I would love to see you @yarikoptic in the downloaders-related stuff (which is completely black magic to me), and @mih maybe for outcome of the recent CLI refactoring and the configuration command?

PR #6709, opened 2022-05-26
Ensure that assertion helpers in `utils_pytest.py` get rewritten by pytest

jwodder:
In order to provide useful information when a test fails, pytest rewrites `assert` statements in test code so that failures will show the values involved.  However, this only applies to code in test files, which, by default, are files with names of the form `test_*.py` or `*_test.py`.  This PR adds `utils_pytest.py` to the set of files considered test files so that the assertions in the assertion helpers within will be rewritten; without this change, a failing, say, `assert_equal(x, y)` will just show "`AssertionError`" but not `x` or `y`.

Note that this only affects tests run on datalad itself; other projects that use assertion helpers in `utils_pytest.py` will not benefit from assertion rewriting (unless we instead rename `utils_pytest.py`).

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -68,6 +68,8 @@ markers =
     with_sameas_remotes
     with_testrepos
     without_http_proxy
+# Ensure that assertion helpers in utils_pytest.py get rewritten by pytest:
+python_files = test_*.py *_test.py utils_pytest.py
 
 [flake8]
 #show-source = True


yarikoptic:
Thank you @jwodder!  I have missed that.  https://stackoverflow.com/a/38138244/1265472 suggests to look into [register_assert_rewrite](http://docs.pytest.org/en/latest/writing_plugins.html?highlight=pytest.register_assert_rewrite#pytest.register_assert_rewrite) and https://pytest.org/en/7.1.x/how-to/assert.html#advanced-assertion-introspection suggests "You can manually enable assertion rewriting for an imported module by calling [register_assert_rewrite](https://pytest.org/en/7.1.x/how-to/writing_plugins.html#assertion-rewriting) before you import it (a good place to do that is in your root conftest.py)." which seems to imply that again each extension would need to do the same and thus effect would be similar to what you have done in tox.ini .

So may be indeed we should just rename it e.g. into `utils_py_test.py` and add into its docstring that it must have `_test.py` ending for this reason?  I would not mind... WDYT @datalad/developers ? I don't think we even need a deprecation cycle in  `utils_pytest.py` adapter since we haven't released yet migration to pytest so no extensions could use it.

yarikoptic:
Well, this improves it. Further improvement could be in some other PR, let's proceed to benefit

PR #6712, opened 2022-05-30
BF: Catch KeyErrors from unavailable WTF infos

### LINKED ISSUE #6711: `wtf` crashed with `-S dataset` executed with no dataset
```sh
/tmp % datalad wtf -S dataset
[ERROR  ] 'dataset' (KeyError) 

/tmp/mike (git)-[master] % datalad wtf -S dataset
# WTF
## dataset 
  - branches: 
    - git-annex@91bd416
    - master@5d8a666
  - id: 7a8e52e7-387a-4361-a810-6935e1519e91
  - metadata: <SENSITIVE, report disabled by configuration>
  - path: /tmp/mike
  - repo: AnnexRepo
```

It should say that this cannot work, because there is no dataset.

adswa:
When unavailble WTF sections are requested, e.g., '-S dataset'
outside of any dataset, the command errors with a KeyError
(#6711). This change catches this and reports an impossible
result for the affected section, such that the problem is
communicated orderly, and that the reporting of other sections
is not impacted

PS: Is `maint` the correct target for this?

#### 🐛 Bug Fixes
- ``wtf`` now reports on unavailable info sections with an 'impossible' result record instead of failing with a KeyError. Fixes #6711

### DIFF
--- a/datalad/local/wtf.py
+++ b/datalad/local/wtf.py
@@ -463,8 +463,18 @@ class WTF(Interface):
                 raise ValueError(flavor)
 
         for s in sections:
-            infos[s] = section_callables[s]()
-
+            try:
+                infos[s] = section_callables[s]()
+            except KeyError:
+                yield get_status_dict(
+                    action='wtf',
+                    path=Path.cwd(),
+                    status='impossible',
+                    message=(
+                      'Requested section <%s> was not found among the '
+                      'available infos. Skipping report.', s),
+                    logger=lgr
+                    )
         if clipboard:
             external_versions.check(
                 'pyperclip', msg="It is needed to be able to use clipboard")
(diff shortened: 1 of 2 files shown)


datalad/local/wtf.py
```diff
@@ -463,8 +463,18 @@ def __call__(*, dataset=None, sensitive=None, sections=None, flavor="full", deco
                 raise ValueError(flavor)
 
         for s in sections:
-            infos[s] = section_callables[s]()
-
+            try:
+                infos[s] = section_callables[s]()
+            except KeyError:
```
yarikoptic, line 468: I know I am annoying, but could there be basic unittest invocation of wtf with incorrect section and getting impossible status -- should be 1 line, right?

Asking because those errors where "while handling this exception , such exception happened" are really annoying UX ;)

yarikoptic (changes requested):
> PS: Is maint the correct target for this?

oh yes. Greedy me just wants a unittest for the goodness of UX ;)  should be trivial to craft, so I do not see a reason why to skip

datalad/local/wtf.py
adswa: good call, done

yarikoptic:
LGTM -- thank you @adswa !

PR #6719, opened 2022-05-31
Restore proper double dash rendering in the docs/manpages

### LINKED ISSUE #4032: Double-dash options get escaped in manpages

### LINKED ISSUE #6699: Docs render double dashes with slash (again?)
I think this has come up before already: instead of ``--help`` it renders ``-\-help`` etc on readthedocs, e.g., [here](http://docs.datalad.org/en/stable/generated/man/datalad-create-sibling-gitlab.html)

Screenshot from 2022-05-23 13-46-39

adswa:
The docs started to render double dashes with a slash (e.g., -\-help)
(see #6699). Through trial and error, I discovered that changing the
doubledash escape sequence in build support seems to fix this issue
without impacting the command-line rendering of the parameters or the
Python help. Originally, the escape sequence that is changed here was
introduced in 96370608b0564 to fix #4032 - which seems to be identical
to the current problem. I haven't out why the slashes were reintroduced,
why the old fix does not suffice anymore, and reverting the old fix
partially seems to fix it -- BUT as things seems to be fixed with it
¯\_(ツ)_/¯... additional insights welcome :)

Fixes #6699

Before this change:
Screenshot from 2022-05-31 09-58-26
After this change: 
Screenshot from 2022-05-31 09-58-50

#### 📝 Documentation
- Proper rendering of double dashes was restored.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -120,7 +120,7 @@ release = datalad.__version__
 #
 # This is also used if you do content translation via gettext catalogs.
 # Usually you set "language" from the command line for these cases.
-language = None
+language = 'en'
 
 # There are two options for replacing |today|: either, you set today to some
 # non-false value, then it is used:


adswa:
> also shouldn't it be against `maint`?

yep, true. Additional PR or would you just cherry-pick?

yarikoptic:
if you were to add a test anyways, IMHO better a replacement PR against `maint`.  CP would require extra memory to do that.

adswa:
whhoops rebase went wrong but need to jump in a meeting and can't fix right away - will get to it later

adswa:
Oh I see - the commit was on maint already....

PR #6720, opened 2022-05-31
Color update

### LINKED ISSUE #6713: `datalad.ui.__init__` logging ineffective, runner before logger is set up
The code runs as the more or less first thing, hence none of the infos it wants to log actually make it out, or can be made to make it out with any of the tooling that datalad provides.

```
-> from .log import lgr
  <frozen importlib._bootstrap>(1007)_find_and_load()
  <frozen importlib._bootstrap>(986)_find_and_load_unlocked()
  <frozen importlib._bootstrap>(680)_load_unlocked()
  <frozen importlib._bootstrap_external>(850)exec_module()
  <frozen importlib._bootstrap>(228)_call_with_frames_removed()
  /home/mih/hacking/datalad/git/datalad/log.py(579)<module>()
-> lgr = LoggerHelper().get_initialized_logger()
  /home/mih/hacking/datalad/git/datalad/log.py(560)get_initialized_logger()
-> ColorFormatter(use_color=use_color,
  /home/mih/hacking/datalad/git/datalad/log.py(154)__init__()
-> msg = colors.format_msg(self._get_format(log_name, log_pid),
  /home/mih/hacking/datalad/git/datalad/support/ansi_colors.py(64)format_msg()
-> if color_enabled() and use_color:
  /home/mih/hacking/datalad/git/datalad/support/ansi_colors.py(59)color_enabled()
-> ui_color == 'auto' and os.getenv('NO_COLOR') is None and ui.is_interactive)
```

### LINKED ISSUE #6714: Color-logging setup decision making twisted
I am looking into decoupling UI setup from logging setup, but it is fighting back:

`LoggerHelper.get_initialized_logger()` calls `datalad.utils.is_interactive()` to check whether to enable color logging. The result of this evaluation (`use_color`) go into `ColorFormatter.__init__(use_color=...)`.

NB: `ColorFormatter` disabled color logging on windows under all circumstances, no rational provided....

`ColorFormatter.__init__`'s `use_color` goes into `datalad.support.ansi_colors.format_msg`. Even if `use_color` is `False`, this function will call `datalad.support.ansi_colors.color_enabled()` first (before disabling color formatting due to `use_color` being `False`.

`datalad.support.ansi_colors.color_enabled()` will query `datalad.ui.ui.is_interactive`, whenever the config switch `datalad.ui.color` is set to `auto` and the env var `NO_COLOR` is not set. This is the default, so `datalad.ui.ui.is_interactive` is pretty much always queries.

`datalad.ui.ui.is_interactive` returns whether the effective UI is a subclass of `InteractiveUI`.

In order to set up an effective UI, `datalad.ui._UI_Switcher.set_backend()` runs, and if no particular backend is requested (the default), it selects `dialog` or `no-progress` after running `datalad.utils.is_interactive()` (again).

This is ... complex...

I think this should be properly dissected and simplified.

It seems desirable to decouple logging setup from UI setup, with two clearly defined lines of decision making/switches.

It seems undesirable to forcefully and unconditionally disable color logging with one switch, but render this switch ineffective by another runtime check.

It also seems desirable to avoid a full UI setup, before the first non-logging message needs to be displayed.

It seems to be undesirable to have `ansi_color` support code depend on a datalad `ui`. The link is only through `color_enabled()`, which is called for pretty much any use of the ansi coloring code, without and caching of the outcome.

### LINKED ISSUE #6718: Reconsider `datalad.support.ansi_colors`
This code is used by the loggers, and by `datalad.ui` internally, but also with `color_word()` directly in a number of commands. However, this is the only function with any significant usage.

The functions in this module implement a complicated pattern that enables either per call enabling or disabling of colorization, or combines it with additional external switches that alter their behavior. These external switches are repeatedly evaluated upon each and every call.

There is no evidence in the code base that any other usage than the default behavior (colorize unless prevented by `datalad.ui.color` (and friends)) is anyhow used.

As described in https://github.com/datalad/datalad/issues/6714 the usage of `ui.isinteractive` in this code tightly couples the `ui` to loggers (I verified that this is the only connection point).

I think this should be rewritten.

There should be an `AnsiFormatter` class that decides whether or not to colorize once, and remembers the outcome.

This close should use `is_interactive()` rather than `ui.isinteractive` to orthogonalize code and decision making better.

It should provide a compatibility shim for `color_word()`, and consolidate everything else into that class.

mih:
With this PR, the previous decision making for color loggin outlined in https://github.com/datalad/datalad/issues/6714 is simplified to:

`LoggerHelper.get_initialized_logger()` calls `datalad.utils.is_interactive()` to check whether to enable color logging. The result of this evaluation (`use_color`) go into `ColorFormatter.__init__(use_color=...)`.

`ColorFormatter.__init__`'s use_color goes into `datalad.support.ansi_colors.formatter.format()`, which is an instance method of `AnsiFormatter`. This instance will check for `datalad.utils.is_interactive()` 
 and env var `NO_COLOR` once in a datalad process.

**Bottom line:** Color logging will be performed if requested by the logging code, and deemed possible by `AnsiFormatter`.

Importantly, the latter no longer requires an initialized UI to function. Therefore, one can now see the UI-related debug messages that are emitted when the UI is actually needed for messaging, e.g.:

```
[DEBUG  ] UI set to DialogUI(out=<TextIOWrapper>) 
```

- Closes #6714 
- Closes #6718 
- Closes #6713

### DIFF
--- a/datalad/core/distributed/push.py
+++ b/datalad/core/distributed/push.py
@@ -305,13 +305,13 @@ class Push(Interface):
         hints = [hint for hint in hints if hint is not None]
         if hints:
             from datalad.ui import ui
-            from datalad.support import ansi_colors
-            intro = ansi_colors.color_word(
+            from datalad.support.ansi_colors import formatter as ansi
+            intro = ansi.colorize(
                 "Hints: ",
-                ansi_colors.YELLOW)
+                ansi.color.YELLOW)
             ui.message(intro)
             [ui.message("{}: {}".format(
-                ansi_colors.color_word(id + 1, ansi_colors.YELLOW), hint))
+                ansi.colorize(id + 1, ansi.color.YELLOW), hint))
                 for id, hint in enumerate(hints)]
 
 
--- a/datalad/core/local/run.py
+++ b/datalad/core/local/run.py
@@ -21,7 +21,7 @@ from pathlib import Path
 from tempfile import mkdtemp
 
 import datalad
-import datalad.support.ansi_colors as ac
+from datalad.support.ansi_colors import formatter as ansi
 from datalad.config import anything2bool
 from datalad.core.local.save import Save
 from datalad.core.local.status import Status
@@ -339,11 +339,11 @@ class Run(Interface):
 
 
 def _display_basic(res):
-    ui.message(ac.color_word("Dry run information", ac.MAGENTA))
+    ui.message(ansi.colorize("Dry run information", ansi.color.MAGENTA))
 
     def fmt_line(key, value, multiline=False):
         return (" {key}:{sep}{value}"
-                .format(key=ac.color_word(key, ac.BOLD),
+                .format(key=ansi.colorize(key, ansi.color.BOLD),
                         sep=os.linesep + "  " if multiline else " ",
                         value=value))
 
--- a/datalad/core/local/status.py
+++ b/datalad/core/local/status.py
@@ -35,7 +35,7 @@ from datalad.interface.utils import (
     eval_results,
     generic_result_renderer,
 )
-import datalad.support.ansi_colors as ac
+from datalad.support.ansi_colors import formatter as ansi
 from datalad.support.param import Parameter
 from datalad.support.constraints import (
     EnsureChoice,
@@ -100,11 +100,11 @@ _common_diffstatus_params = dict(
 
 
 STATE_COLOR_MAP = {
-    'untracked': ac.RED,
-    'modified': ac.RED,
-    'deleted': ac.RED,
-    'added': ac.GREEN,
-    'unknown': ac.YELLOW,
+    'untracked': ansi.color.RED,
+    'modified': ansi.color.RED,
+    'deleted': ansi.color.RED,
+    'added': ansi.color.GREEN,
+    'unknown': ansi.color.YELLOW,
 }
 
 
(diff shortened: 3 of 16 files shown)


yarikoptic (commented):
FWIW, not to not approve this PR, but possibly to gain some additional insight for rationale for such a change, or may be to avoid future RFing which would make code only more "elaborate" (needing legacy deprecation code, trickier imports/patches, etc) -- I do not see really a reason for encapsulating these interfaces into a class here:
- for all means and purposes encapsulation of all those methods at the module level was IMHO just as fine
  - lazy evaluation still possible to do, just define global to module `_variables` if so desired 
- `AnsiFormater` is a single class and there is no another 
- there will be only a single instance of it (`formater`), so altogether with above -- no really point for a class instance IMHO
- now imports became quite involved `from datalad.support.ansi_colors import formatter as ansi` in that it is not `ansi` module/object, it is `formatter`, and from `ansi_colors` module. So - need to remember the import pattern for consistency (instead of just "import this module").  
 
So IMHO such RFing just increases cognitive load for no real need -- OOP is good but there are code patterns where procedural code is just as fine and might be even easier to understand.

mih:
This is the only way I can contribute in a way that the referenced issues are closed. Happy to have this PR replaced with a better one. I don't know why it is better to have a module act as a singleton, or have the need to force this class be a singleton. But if this is required, someone with a better understanding should take over.

yarikoptic:
FTR  - benchmarks were failing due to

```
          169±1ms          186±7ms    ~1.10  core.startup.time_import
```

which could be a fluke -- didn't check

PR #6722, opened 2022-05-31
MNT: Remove unused helper for listing sorted files

### LINKED ISSUE #6432: datalad.utils.sorted_files() unused, untested

adswa:
There are no uses of this function in core (as the comment indicated) or in any of the extensions (as determined by a quick git grep - but tests will tell if I have missed anything).
Removing it to not drag it further along with us.

closes #6432

### DIFF
--- a/datalad/utils.py
+++ b/datalad/utils.py
@@ -344,14 +344,6 @@ def md5sum(filename):
     return Digester(digests=['md5'])(filename)['md5']
 
 
-# unused in -core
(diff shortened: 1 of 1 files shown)


yarikoptic:
FTR, seems was used in `check_page2annex_same_incoming_and_public` of a crawler tests, but I don't think that one even migrated to datalad-crawler extension in those extraction times... it seems indeed should be safe to kill. Thanks @adswa !

PR #6738, opened 2022-06-02
MNT: Remove ``eval_file_type()`` from code base

### LINKED ISSUE #6652: remove  eval_file_type  from "internal" interfaces? -- deprecated at AnnexRepo level, unused
became annoying in #6273

adswa:
This is a maintenance change to fix #6652. It removes all remaining uses of the deprecated and unused ``eval_file_type()`` parameter, and the parameter itself. As far as I was able to grep around, only ``datalad-ukbiobank`` needs an adjustment for this change, but let's see if any unit tests break.


adswa:
No-one has complained and @yarikoptic approved, so I guess I'll merge :)

Issue #4028, opened 2020-01-15
siblings needs to be able to set annex-ignore


bpoldrack:
Ran into this in https://github.com/datalad/git-annex-ria-remote/pull/38

`datalad-siblings-configure` doesn't allow to specify `annex-ignore`. If I now configure a sibling to a still empty repo (in order to be able to push to it), it tries to fetch and complains with a warning, that it isn't able to determine whether that remote is carrying an annex. But I can configure `annex-ignore` only afterwards. Using this in an extensions command means, there's no way to disable that warning, which makes no sense in that particular context, since it's the entire point of adding that sibling to push to it.

bpoldrack:
Should this rather be addressed by utilizing the new `configure` command, @datalad/developers ?

kyleam:
> Should this rather be addressed by utilizing the new `configure` command

I'm not sure if there should be a nicer way, but configuring the remote's annex-ignore value before running `siblings` seems like an okay approach to me and squelches the following messages on my end:

```
[INFO   ] Could not enable annex remote sib. This is expected if sib is a pure Git remote, or happens if it is not accessible. 
[WARNING] Could not detect whether sib carries an annex. If sib is a pure Git remote, this is expected.  
```



<details>
<summary>script</summary>

```sh
set -eu

cd "$(mktemp -d "${TMPDIR:-/tmp}"/dl-XXXXXXX)"
tdir=$(pwd)

git init --bare sib
datalad create a
(
    cd a
    datalad configuration --scope=local set remote.sib.annex-ignore=true
    datalad siblings configure --name=sib --url="$tdir/sib"
    datalad push --to=sib
)
```

```
Initialized empty Git repository in /tmp/dl-OHh2tIw/sib/
[INFO   ] Creating a new annex repo at /tmp/dl-OHh2tIw/a 
[INFO   ] Scanning for unlocked files (this may take some time) 
create(ok): /tmp/dl-OHh2tIw/a (dataset)
set_configuration(ok): . [remote.sib.annex-ignore=true]
.: sib(-) [/tmp/dl-OHh2tIw/sib (git)]
publish(ok): . (dataset) [refs/heads/master->sib:refs/heads/master [new branch]]        
publish(ok): . (dataset) [refs/heads/git-annex->sib:refs/heads/git-annex [new branch]]
```
</details>

mih:
Given that `siblings` already has:

- `--annex-wanted`
- `--annex-required`
- `--annex-group`
[...]

PR #6742, opened 2022-06-03
ENH: Allow siblings to set annex.ignore of siblings

adswa:
This PR tries to implement #4028. It adds a ``--annex-ignore`` parameter to ``siblings`` that can add or set an `annex.ignore` config in the datasets ``.git/config``. For this, it also adds an internal helper ``set_remote_ignore`` in ``Annexrepo``. When the parameter is set, it will be added to a remote configuration before annex has a chance to emit the ``Could not enable annex remote %s. This is expected if %s is a pure Git remote, or happens if it is not accessible.``-warning

I'm not sure though if this is sufficient to address the use case in  #4028 so I would like to as @bpoldrack for clarification. Was something simple like this what you had envisioned? Or does it need to go deeper?
I'm also not sure about the parameter specification (suggestions for what to change ``metavar`` and the value defaults into are welcome - I think there are too many defaults, and I'm not sure if its a valid metavar name)

### Changelog
TBD

### DIFF
--- a/datalad/interface/common_opts.py
+++ b/datalad/interface/common_opts.py
@@ -269,6 +269,11 @@ annex_groupwanted_opt = Parameter(
     See https://git-annex.branchable.com/git-annex-groupwanted/ for more information""",
     constraints=EnsureStr() | EnsureNone())
 
+annex_ignore_opt = Parameter(
+    args=("--annex-ignore",),
+    metavar='BOOL',
+    doc="Whether or not a remote shall be ignored or not",
+    constraints=EnsureChoice(None, True, False, 'true', 'false'))
 
 inherit_opt = Parameter(
     args=("--inherit",),
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -699,6 +699,16 @@ class AnnexRepo(GitRepo, RepoInterface):
             default=False
         )
 
+    def set_remote_ignore(self, remote, ignored=False):
+        """Add or set an annex-ignore section for a given remote"""
+        # check if there is a section already
+        if self.config.has_section('remote.{}.annex-ignore'.format(remote)):
+            # overwrite (set) config
+            self.config.set('remote.{}.annex-ignore'.format(remote), ignored, scope='local')
+        else:
+            # add it
+            self.config.add('remote.{}.annex-ignore'.format(remote), ignored, scope='local')
+
     def is_special_annex_remote(self, remote, check_if_known=True):
         """Return whether remote is a special annex remote
 
--- a/datalad/distribution/siblings.py
+++ b/datalad/distribution/siblings.py
@@ -32,6 +32,7 @@ from datalad.interface.common_opts import (
     annex_groupwanted_opt,
     annex_required_opt,
     annex_wanted_opt,
+    annex_ignore_opt,
     as_common_datasrc,
     inherit_opt,
     location_description,
@@ -191,6 +192,7 @@ class Siblings(Interface):
         annex_required=annex_required_opt,
         annex_group=annex_group_opt,
         annex_groupwanted=annex_groupwanted_opt,
+        annex_ignore=annex_ignore_opt,
         inherit=inherit_opt,
         get_annex_info=Parameter(
             args=("--no-annex-info",),
@@ -221,6 +223,7 @@ class Siblings(Interface):
             annex_required=None,
             annex_group=None,
             annex_groupwanted=None,
+            annex_ignore=None,
             inherit=False,
             get_annex_info=True,
             recursive=False,
@@ -277,6 +280,7 @@ class Siblings(Interface):
             annex_required=annex_required,
             annex_group=annex_group,
             annex_groupwanted=annex_groupwanted,
+            annex_ignore=annex_ignore,
             inherit=inherit,
             get_annex_info=get_annex_info,
             res_kwargs=res_kwargs,
@@ -428,7 +432,7 @@ def _configure_remote(
         ds, repo, name, known_remotes, url, pushurl, fetch, description,
         as_common_datasrc, publish_depends, publish_by_default,
         annex_wanted, annex_required, annex_group, annex_groupwanted,
-        inherit, res_kwargs, **unused_kwargs):
+        annex_ignore, inherit, res_kwargs, **unused_kwargs):
     result_props = dict(
         action='configure-sibling',
         path=ds.path,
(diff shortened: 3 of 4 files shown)


datalad/support/annexrepo.py
```diff
@@ -699,6 +699,16 @@ def is_remote_annex_ignored(self, remote):
             default=False
         )
 
+    def set_remote_ignore(self, remote, ignored=False):
+        """Add or set an annex-ignore section for a given remote"""
+        # check if there is a section already
+        if self.config.has_section('remote.{}.annex-ignore'.format(remote)):
```
yarikoptic, line 705: well, indeed it returns True

```shell
$> python3 -c 'import datalad.api as dl; print(dl.Dataset(".").config.has_section("remote.origin.annex-ignore"))'  
True
```
which somewhat surprises me since this setting is not an entire section but just one option, so wouldn't a more semantically correct :?

```shell
$> python3 -c 'import datalad.api as dl; print(dl.Dataset(".").config.has_option("remote.origin", "annex-ignore"))' 
True
```

datalad/support/annexrepo.py
```diff
@@ -699,6 +699,16 @@ def is_remote_annex_ignored(self, remote):
...
+            # overwrite (set) config
+            self.config.set('remote.{}.annex-ignore'.format(remote), ignored, scope='local')
+        else:
+            # add it
+            self.config.add('remote.{}.annex-ignore'.format(remote), ignored, scope='local')
```
yarikoptic, line 710: `.set` seems to work just fine for unknown yet options, even if the entire section doesn't exist
```
(git-annex)lena:~/datalad/openneuro/ds000001[master]git
$> python3 -c 'import datalad.api as dl; print(dl.Dataset(".").config.set("remote.origin.brand-new", "true", scope="local"))'
None

$> git config remote.origin.brand-new                                                                                        
true

$> python3 -c 'import datalad.api as dl; print(dl.Dataset(".").config.set("remote.unknown.brand-new", "true", scope="local"))' 
None

$> git config remote.unknown.brand-new                                                                                  
true
```
but how would we "unset" or remove that setting entirely -- e.g. if I would prefer to leave it for git-annex to *possibly* set. I.e. may be if the `ignored` is `None` it would remove that setting if it is set (to any - true or false)?

mih:
After loading the issue in my brain, a few high-level notes to put everything in one place.

This underlying issue is #4028 by @bpoldrack which states "siblings needs to be able to set annex-ignore". It is not clear to me why it *needs* to do this, in particular because the statement "there's no way to disable that warning" is invalidated by the demo at https://github.com/datalad/datalad/issues/4028#issuecomment-763196929

So from my POV this is not a necessity, but rather a convenience. It think it is perfectly fine for (extension) developers to account for this situation with a dedicated `configuration` call, rather than having to have an option for `siblings`.

The situation is a bit different for a user adding a sibling manually. However, they would need to know/anticipate that a particular sibling does not or shall not have an annex. I am not able to come up with an actual use case where a users knows that, cares about the warning not to be displayed, and would not want to set the configuration upfront. Again pointing to the convenience nature of the issue.

With that conclusion in mind, I stand by my comment in https://github.com/datalad/datalad/issues/4028#issuecomment-849382862 that we cannot make the current state of `siblings` worse by adding such a convenience option (frankly, the command is already overloaded beyond repair).

So far here, I will make a few comments on the diff soon.

datalad/interface/common_opts.py
```diff
@@ -269,6 +269,11 @@
     See https://git-annex.branchable.com/git-annex-groupwanted/ for more information""",
     constraints=EnsureStr() | EnsureNone())
 
+annex_ignore_opt = Parameter(
```
mih, line 272: Is there another envision use for this option in another command? If not, I think this should move into the `siblings` code. It is hard to get this description right if the usage patterns are unclear. And it is easy to break semantics when later fixing up the description from one angle, without seeing other potential usage. UNtil this is clear there is little value in centralization IMHO.

datalad/interface/common_opts.py
```diff
@@ -269,6 +269,11 @@
...
+    args=("--annex-ignore",),
+    metavar='BOOL',
```
mih, line 274: It is generally not beneficial to declare `metavar` AND `constraints`. Here is the help resulting from this implementation:

```
Usage: datalad siblings [--annex-ignore BOOL]
```

when `metavar` is removed:

```
Usage: datalad siblings [--annex-ignore {true|false}]
```

In many cases the constraints are capable of explaining themselves.

datalad/interface/common_opts.py
```diff
@@ -269,6 +269,11 @@
...
+    doc="Whether or not a remote shall be ignored or not",
+    constraints=EnsureChoice(None, True, False, 'true', 'false'))
```
mih, line 276: This is less general than `EnsureBool() | EnsureNone()`, but the resulting composite constraint looses the ability to generate a meaningful help. Although it would be =relatively simple and useful to finally fix this (out of scope here). This this implementation is (sadly) optimal.

mih (changes requested):
Thanks for the PR!

Following up on my previous comment, I think this is a convenience-related change -- which is nevertheless good.

I believe the new option should not be a "common" one, though.

adswa:
Thanks much for your comments and suggestions @mih and @yarikoptic!
> This underlying issue is https://github.com/datalad/datalad/issues/4028 by @bpoldrack which states "siblings needs to be able to set annex-ignore". It is not clear to me why it needs to do this, in particular because the statement "there's no way to disable that warning" is invalidated by the demo at https://github.com/datalad/datalad/issues/4028#issuecomment-763196929

> So from my POV this is not a necessity, but rather a convenience. It think it is perfectly fine for (extension) developers to account for this situation with a dedicated configuration call, rather than having to have an option for siblings.

I agree - FTR, I would also be completely fine with closing this. I'm not yet an avid user of ``datalad configuration``, but it seems similarly convenient as a siblings call

bpoldrack:
I largely agree with @mih assessment.

It is convenience, but I still find a bit strange having to set a config for a remote that doesn't exist yet. So, I am in favor of this convenience. 

I also agree, that the parameter definition should probably not be centralized (yet). Do you intend to address this, @adswa ?

yarikoptic:
my 1c: there was support in favor of the convenience but AFAIK it seems "too big of a convenience" for what seems to happen so far only in a single type of special remote and nobody really bothered to workaround with `config`/`configure` setting before `siblings` were to be invoked. So it suggests that this feature/extension of interface is not really in need.  So, despite that @adswa already invested so much time to code it up, I do not really think it is needed.

adswa:
Then let's close this :+1:

PR #5010, opened 2020-10-10
BF/RF: GitRepo.push -- push first refspec separately first for github (or if remote.{remote}.datalad-push-default-first is set)

yarikoptic:
This is a central place for logic of pushing refspecs used by push and publish,
so I decided to fix in a single spot instead of duplicating logic or providing
yet another additional wrapper, and thus possibly of help to any other code which
uses .push().

Note:
- logic is in effect only if remote and refspecs are provided. So it will be
  of no effect if either of those not provided, but most likely there would be
  no relevant use case since it would require manual configuration first of the
  refspecs and default remote for a branch within config before submitting a push

Qs:
-  may be logic should move even deeper into  push_ generator?
-  ideas on how to establish testing for this, without interacting with github, since it seemed we encountered such behavior on pushing multiple refspecs at once only with github. ATM tested "manually" with both push and publish using reproducer in #4997
- [x] redo following the proposal in https://github.com/datalad/datalad/pull/5010#issuecomment-712997679 to explicitly annotate config while `create-sibling-github` to do separate initial push(es) only in that case (not always upon initial push).  I am not 100% satisfied in such approach since would make it github specific, and might be needed for other (gitlab?) if they start behaving the same way

Closes: #4997

PR #6750, opened 2022-06-07
BF: push to github - remove datalad-push-default-first config only in non-dry run to ensure we push default branch separately in next step

### LINKED ISSUE #6749: create-sibling-github + push could still end up with `git-annex` branch being default
- if user/organization has different (e.g. `main`) branch being default from the one we push
- described in handbook http://handbook.datalad.org/en/latest/basics/101-180-FAQ.html#gitannexdefault
- tireless @dnkennedy kept running into it: https://github.com/ReproNim/OHBMEducation-2022/issues/10#issuecomment-1148558908
- we had a fix in #5010 which was supposed to fix it (released 0.14.1)
- the fix is not really in effect in `datalad push --to=github` at https://github.com/datalad/datalad/pull/5010/files#diff-bfc1c47d032fc72b15b6a79db3d49e4c4334cf96318c5efe48a3403744b32fb1R2453 since `refspec` is `None` so we do not push current branch first.
- in that PR added test https://github.com/datalad/datalad/pull/5010/files#diff-f340e678f6f2a72a87e905265328cea97fe63b307e450c99b47bb006128ee55dR1722 we operate on clone and provide refspec explicitly so test checks that logic, but unfortunately nohow checks operation on github
- FWIW I did test with 0.14.1 (and specific that PR 519e7dd8ffc7dede5739e3d17321ce6ab72b98a6) and ATM it would make `git-annex` branch default
- so it is only up to anecdotal evidence of deteriorating memory of an elderly that it actually provided remedy before somehow.

I am digging into it

yarikoptic:
Fixes #6749. Ref: original fix in #5010  which was incomplete

- I am not 100% confident and did not carry full archeological expedition on
  how it could have worked ever since due to that --dry-run it was "always"
  there
- 2nd call to that push comes with refspec since it is that initial --dry-run
  which provides them. And that is when we should separate them out
- Unfortunately since there is no dedicated dry_run option to push,
  I had to rely on matching it in git-options
- We stopped using VCR for playing out the integration tests with github,
  and mocking out interactions would provide no point since we want to ensure
  correct operation against github.  So I do not see an easy way to
  develop a unittest, so I will leave it at this

### DIFF
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -1983,7 +1983,8 @@ class GitRepo(CoreGitRepo):
                 )
             )
         # note: above push_ should raise exception if errors out
-        if cfg.get_from_source('local', cfg_push_var) is not None:
+        if '--dry-run' not in git_options \
+            and cfg.get_from_source('local', cfg_push_var) is not None:
             lgr.debug("Removing %s variable from local git config after successful push", cfg_push_var)
             cfg.unset(cfg_push_var, 'local')
         return push_res


yarikoptic:
travis -- known #4101 (Restarted). appveyor -- test_ria_postclonecfg (attn @bpoldrack )

if no objections and no further comments, since fix is IMHO straightforward, I will merge (and thus release) tomorrow

PR #6751, opened 2022-06-07
TEMP: do only 1 test on NFS but 100 times to reproduce #4101

yarikoptic:


yarikoptic:
Dang, sleeping seems didn't help, so it is something more peculiar

yarikoptic:
HA! with the sleep before creating tarball a full run succeeded! restarted it to ensure that it stays so.

I guess the most logical fix would be to ensure that the entire "tree" does "list" as created, and if not -- sleep a little

yarikoptic:
#6912 CPs the last commit adding xfail

PR #6754, opened 2022-06-08
Preserve final trailing slash in ``call_git()`` output

adswa:
A failure in ``datalad-ukbiobank`` (datalad/datalad-ukbiobank#82 and a misguided fix in datalad/datalad-ukbiobank#85 has revealed that a change introduced in https://github.com/datalad/datalad/pull/6278 caused final trailing newlines to be removed from the final stdout of our ``call_git`` calls. In datalad-ukb, unit tests started to fail as the contents of files that were written with `git cat-file -p <sourcebranch>:<sourcefile>` lacked the trailing newlines that the source files had.

I'm poking in the dark and PRing a WIP PR in hopes of improvements and feedback to fix this by switching back to ``_call_git()`` from ``call_git_items()``. The PR also contains half of a docstring fix @yarikoptic idenitified - input on the description of the `env` parameter (see commit) would be appreciated.

### DIFF
--- a/datalad/dataset/gitrepo.py
+++ b/datalad/dataset/gitrepo.py
@@ -366,7 +366,8 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
         Internal helper to the call_git*() methods.
         Unlike call_git, _call_git returns both stdout and stderr.
         The parameters, return value, and raised exceptions match those
-        documented for `call_git`, with the exception of env, which is TODO?.
+        documented for `call_git`, with the exception of env, which allows to
+        specify the custom environment (variables) to be used.
         """
         runner = self._git_runner
         stderr_log_level = {True: 5, False: 11}[expect_stderr]
--- a/datalad/dataset/gitrepo.py
+++ b/datalad/dataset/gitrepo.py
@@ -429,12 +429,12 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
         ------
         CommandError if the call exits with a non-zero status.
         """
-        return "\n".join(
-            self.call_git_items_(args,
-                                 files,
-                                 expect_stderr=expect_stderr,
-                                 expect_fail=expect_fail,
-                                 read_only=read_only))
+
+        return self._call_git(args,
+                 files,
+                 expect_stderr=expect_stderr,
+                 expect_fail=expect_fail,
+                 read_only=read_only)[0]
 
     def call_git_items_(self,
                         args,
--- a/datalad/dataset/gitrepo.py
+++ b/datalad/dataset/gitrepo.py
@@ -364,9 +364,9 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
         """Allows for calling arbitrary commands.
 
         Internal helper to the call_git*() methods.
-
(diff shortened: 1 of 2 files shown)


yarikoptic:
I have rebased onto maint, extended tuneup of the docstring, force pushed and changed the base to be `maint` here.

If no failures and no objections are stated (inviting @christian-monch for review), I think we should merge it and take ukbiobank out of hostage situation.

bpoldrack (commented):
I don't think this is the way to go.

This changes the underlying call from `_generator_call_git` to `_call_git` if I'm not mistaken. Hence, the change does something substantially different from its actual aim.

How about `call_git_items_` simply getting an option to not swallow an empty line at the end?

Alternatively, one could think about a more specialized helper for `cat-file`, since it's a somewhat different usecase to read a file content from getting other kinds of results from git calls.

adswa:
I guess we really need insights from @christian-monch

yarikoptic:
I think it didn't survive change of the base neatly... I have force pushed rewritten last commit to get CI redo it

yarikoptic:
```
====================================================================
FAIL: datalad.tests.test_config.test_bare
  File "/tmp/dl-miniconda-d53tixnp/lib/python3.9/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/build/datalad/datalad/datalad/tests/utils.py", line 874, in _wrap_with_tempfile
  File "/home/travis/build/datalad/datalad/datalad/tests/utils.py", line 874, in _wrap_with_tempfile
  File "/home/travis/build/datalad/datalad/datalad/tests/test_config.py", line 660, in test_bare
    assert_equal(
AssertionError: '725ae47b529a6627b987523e19b3c3981ba44264\n' != '725ae47b529a6627b987523e19b3c3981ba44264'
- 725ae47b529a6627b987523e19b3c3981ba44264
?                                         -
+ 725ae47b529a6627b987523e19b3c3981ba44264
```
seems like legit/intended fail here since now we would have a trailing new line

christian-monch:
> This changes the underlying call from `_generator_call_git` to `_call_git` if I'm not mistaken. Hence, the change does something substantially different from its actual aim.
> 

I think @adswa s PR is fine. The previous state was:  

- Inside `call_git` we call `call_git_items_` (which does call `_generator_call_git`) in `call_git` and unwind the generator in `call_git`

The current state is:  
- Inside `call_git` we call `_call_git` (which does call `_generator_call_git`) and unwinds the generator

Both are almost equivalent besides the fact that `_call_git` assembles the stderr-output which is ignored in `call_git`.

I think `call_git_items_`  should probably be modified to faithfully recreate the output of `git`, if the results are joined by `"".join(call_git_items_(....))`. Will create an issue for that.

christian-monch (approved):
IMO this is a good solution for the problem. The only fly in the ointment is that stderr-output is recorded and ignored, but correctness and PR simplicity should trump performance considerations.

The tests should be modified to reflect the updated behavior

christian-monch:
Pushed a fix for the tests

christian-monch:
> seems like legit/intended fail here since now we would have a trailing new line

Indeed, pushed an updated test

PR #6757, opened 2022-06-09
Properly include "data packages" in project

jwodder:
There are several folders in `datalad/` that contain data files but no Python source code, and we want these to be included when datalad is installed.  Currently, these folders are automatically included by the combination of `graft datalad` in `MANIFEST.in` and `include_package_data=True` in `setup.py`, but because we set `packages` in `setup()` to `find_packages()` instead of `find_namespace_packages()`, the folders themselves are not recognized by setuptools as packages.  [This combination of settings is deprecated](https://github.com/pypa/setuptools/issues/3340), and the setuptools developers recommend using `find_namespace_packages()` instead of `find_packages()` for such situations.

### DIFF
--- a/_datalad_build_support/setup.py
+++ b/_datalad_build_support/setup.py
@@ -25,7 +25,7 @@ from packaging.version import Version
 from setuptools import (
     Command,
     DistutilsOptionError,
-    find_packages,
+    find_namespace_packages,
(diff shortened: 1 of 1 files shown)


yarikoptic:
Thank you @jwodder !

Issue #6759, opened 2022-06-10
can't store token for github -- operation fails


yarikoptic:
trying on distant mac with datalad from conda-forge 0.16.5

```
(datalad) yoh@datalads-imac2 testbranch % datalad -l debug create-sibling-github testbranch
[DEBUG  ] Command line args 1st pass for DataLad 0.16.5. Parsed: Namespace() Unparsed: ['create-sibling-github', 'testbranch'] 
[DEBUG  ] Building doc for <class 'datalad.distributed.create_sibling_github.CreateSiblingGithub'> 
[DEBUG  ] Parsing known args among ['/Users/yoh/miniconda3-2/envs/datalad/bin/datalad', '-l', 'debug', 'create-sibling-github', 'testbranch'] 
[DEBUG  ] Determined class of decorated function: <class 'datalad.distributed.create_sibling_github.CreateSiblingGithub'> 
[DEBUG  ] Importing keyring 
An access token is required for https://api.github.com. Visit https://github.com/settings/tokens to create a token.
token: 
[DEBUG  ] Token retrieval failed: Can't store password on keychain: (-25308, 'Unknown Error') 
[WARNING] Cannot determine authorization token for api.github.com 
[ERROR  ] Authorization required for GitHub, cannot find token for a credential api.github.com. (ValueError) 
```

narclab:
Is this workaround something a user should be able to implement? I am following this guide to set up a special remote linked to github: https://handbook.datalad.org/en/latest/basics/101-139-s3.html

Running `datalad configuration` shows me the tokens I have installed, and all other git and gh related operations indicate successful authentication, yet I still receive the same (or very similar) error:  

`datalad create-sibling-github -d . more-test --publish-depends more-test
[WARNING] Cannot determine authorization token for api.github.com 
[ERROR  ] Authorization required for GitHub, cannot find token for a credential api.github.com. 
jgray@narc-PowerEdge-R720xd:/mnt/minio/hdd8tb/disk1/ntfs/more$ datalad -l debug create-sibling-github -d . more-test --publish-depends more-test
[DEBUG  ] Command line args 1st pass for DataLad 0.18.3. Parsed: Namespace() Unparsed: ['create-sibling-github', '-d', '.', 'more-test', '--publish-depends', 'more-test'] 
[DEBUG  ] Failed to import requests_ftp, thus no ftp support: ModuleNotFoundError(No module named 'requests_ftp') 
[...]

PR #6762, opened 2022-06-10
BF(workaround to #6759): if saving credential failed, just log error and continue

yarikoptic:
This way user can still continue and achieve the goal without the entire
datalad process crashing and have no clear way forward.  User would just
need to keep entering credential every time I guess.

Ref (not really a fix IMHO): #6759

### DIFF
--- a/datalad/downloaders/credentials.py
+++ b/datalad/downloaders/credentials.py
@@ -140,7 +140,10 @@ class Credential(object):
 
     def _ask_and_set(self, f, instructions=None):
         v = self._ask_field_value(f, instructions=instructions)
-        self.set(**{f: v})
+        try:
+            self.set(**{f: v})
+        except Exception as e:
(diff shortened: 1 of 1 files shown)


yarikoptic:
appveyor - mac - test_ria_postclonecfg attn @bpoldrack . Otherwise - no side effects

mih:
Only tangentially related here, but the pattern established for the new-style credential handling is more different. It will only ever attempt to set a credential *after* it has proven to actually work. It would be great to figure out, how this could work with downloaders. So far, I failed to figure this out, which is blocking the introduction of REALM-based credential identification for this core functionality.

yarikoptic:
Indeed would be nice to make that happen. As for this particular change 
- would new credential mechanism crash process of unable to set credential?
- does this change look ok?

christian-monch (approved):
Good strategy

PR #6773, opened 2022-06-23
WiP: dump of a provider config/auth to support downloads from brainlife

yarikoptic:
Attn @soichih :

```shell
datalad --dbg download-url https://brainlife.io/api/warehouse/dataset/download/605b91bbbc89fe21415e1958
```
worked for me and I got 

```shell
$> tar -tvf 605b91bbbc89fe21415e1958.tar 
-rw-r--r-- brlife/brlife   766 2021-03-24 15:23 microperimetry.tsv
```

### DIFF
--- a/datalad/downloaders/__init__.py
+++ b/datalad/downloaders/__init__.py
@@ -12,6 +12,7 @@
 
 from datalad.downloaders.credentials import (
     AWS_S3,
+    BrainLife_Token,
     LORIS_Token,
     NDA_S3,
     Token,
@@ -32,5 +33,6 @@ CREDENTIAL_TYPES = {
     'nda-s3': NDA_S3,
     'token': Token,
     'loris-token': LORIS_Token,
+    'brainlife-token': BrainLife_Token,
     'git': GitCredential,
 }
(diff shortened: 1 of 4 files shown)


adswa:
any reason why this is a draft?

yarikoptic:
> any reason why this is a draft?

not yet finished - was a quick draft while at OHBM.  @soichih should we finish it up? interested in some quality hours of quick workathon or smth like that? ;)

yarikoptic:
will close for now, and resurrect whenever we get back to it

soichih:
Sorry I didn't respond back to this sooner. Yes, let's work on it when we get together again.

My only comment to this right now is that we probably don't want to store user/pass on Datalad end. We should let user login to brainlife with user/pass using brainlife CLI tool (or datalad could delegate it?) and datalad can just use the .jwt token from `~/.config/brainlife.io/.jwt` to download data.

yarikoptic:
Deal!

Is brainlife CLI tool `pip install`able so we could just depend on it? if not -- might be better to reimplement that logic/API call so we do not complicate installation. Counter-argument: if someone wants to interact with brainlife via datalad, they are likely to have brainlife CLI already installed.

soichih:
`npm install brainlife` is how it's currently installed.. but I am not sure if there is a way to somehow wrap that into `pip` command? But yes, from datalad point of view, it can just look for that file `~/.config/brainlife.io/.jwt` and if it doesn't exists just show error message saying they need to venture into the magical land of brainlife.io.

PR #6781, opened 2022-06-30
Upload coverage from extension tests to Codecov

### LINKED ISSUE #6780: submit codecov report from testing against extensions as well
apparently we are not submitting coverage in https://github.com/datalad/datalad/blob/HEAD/.github/workflows/test_extensions.yml#L61 as we IMHO should. @jwodder please help with this little drill

jwodder:
Closes #6780.

#### 🛡 Tests
- Upload coverage from extension tests to Codecov

### DIFF
--- a/.github/workflows/test_extensions.yml
+++ b/.github/workflows/test_extensions.yml
@@ -38,7 +38,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install -r requirements.txt
+        pip install -r requirements-devel.txt
     - name: Install ${{ matrix.extension }} extension
       run: |
         pip install https://github.com/datalad/${{ matrix.extension }}/archive/master.zip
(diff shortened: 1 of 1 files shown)


.github/workflows/test_extensions.yml
```diff
@@ -38,7 +38,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install -r requirements.txt
+        pip install -r requirements-devel.txt
```
yarikoptic, line 41: note -- the difference between the two besides `[full]` and `[devel]` flavors of install is that in `-devel` we install with `pip -e` i.e. for development.  Should be ok, but just worth noting.

jwodder: Note that I changed it to `-devel` in order to get `coverage` installed.

PR #6783, opened 2022-07-01
Remove superfluous `AnnexRepo.remove()` implementation

mih:
By moving the `Repo.precommit()` call into the actual
implementation in `GitRepo`. No change in API or functionality.

This scores a point in datalad/datalad#4595

#### 🏠 Internal
- Consolidate `GitRepo.remove()` and `AnnexRepo.remove()` into a single implementation.

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -2683,24 +2683,6 @@ class AnnexRepo(GitRepo, RepoInterface):
             self._batched.close()
         super(AnnexRepo, self).precommit()
 
-
-    @normalize_paths(match_return_type=False)
-    def remove(self, files, force=False, **kwargs):
-        """Remove files from git/annex
-
-        Parameters
-        ----------
-        files
-        force: bool, optional
-        """
-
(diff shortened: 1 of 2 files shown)


bpoldrack (commented):
While I generally agree with having `GitRepo` handle this, you drop the `force` option from signature. It should still function (due to `to_options`), but now it's invisible to the caller. Maybe just make it explicit in `GitRepo.remove`?

yarikoptic (approved):
Makes total sense to me !  Checked other methods in AnnexRepo and it is only `set_metadata_` which is calling `.precommit`, and no other ... In principle AFAIK, all parallel batched git-annex'es are operating with some awareness of each other due to journaling their operation and that is why things seems to remain working even though we do not have `.precommit` in e.g. `_call_annex`.  So to some degree comment about "manifest ... on disk" is also incomplete/off. ATM it is indeed for changes to manifest in the tree and git index AFAIK.

yarikoptic:
CI run was done probably before chardet versioning workaround was introduced so travis is full of fails... I would assume they aren't related and appveyor is largely green, so let's just proceed

PR #6785, opened 2022-07-01
Fix "common data source" test to use a valid URL

### LINKED ISSUE #6779: initremote type=git fresh "regression"
This is a sattelite of originally filed only in git-annex for @joeyh at https://git-annex.branchable.com/bugs/initremote_type__61__git_fresh___34__regression__34__/  and which nobody yet troubleshooted to resolution yet, but which started to appear in our CI due to recent release of git-annex.

mih:
Previously the test relied on git-annex to be more tolerant to
invalid repository URLs than Git itself.

Fixes datalad/datalad#6779

#### 🛡 Tests
- Fix "common data source" test to use a valid URL to avoid failure with git-annex later than 10.20220525-64-g14584e7a3. Fixes #6779

### DIFF
--- a/datalad/distribution/tests/test_siblings.py
+++ b/datalad/distribution/tests/test_siblings.py
@@ -548,7 +548,11 @@ def test_as_common_datasource(testbed=None, viapath=None, viaurl=None, remotepat
     res = ds_fromurl.siblings(
         'add',
         name='fresh',
-        url=url,
+        # we must amend the URL given by serve_path_via_http, because
+        # we are serving the root of a non-bare repository, but git-annex
+        # needs to talk to its .git (git-clone would also not eat
+        # `url` unmodified).
+        url=url + '.git',
         as_common_datasrc='fresh-sr',
         result_renderer='disabled',
     )


yarikoptic:
why not `maint` since it is a bug fix to issue present there as well AFAIK?

yarikoptic:
ha! with annex 10.20220322-1~ndall+1, while testing locally "extended" test and rebased onto `maint` I got into

```
datalad.runner.exception.CommandError: CommandError: 'git -c diff.ignoreSubmodules=none annex initremote fresh-sr type=git location=http://127.0.0.1:38045/.git autoenable=true -c annex.dotfiles=true' failed with exitcode 1 under /home/yoh/.tmp/datalad_temp_test_as_common_datasource49kq5qqu [out: 'initremote fresh-sr 
failed'] [err: 'git-annex: found multiple git remotes with specified location
initremote: 1 failed']
```
so some newer (but not "brand new" since the bleeding edge one swallows it just fine; I have not established the range) would start puking at the test fix :-/

yarikoptic:
IMHO superseded by extended version in #6788 (summary of rationale: adding .git to url makes test pass, but not for all git-annexes and does not really fix the test to test for that other added remote)

PR #6788, opened 2022-07-01
Fix "common data source" test to use a valid URL (maint-based & extended edition)

yarikoptic:
Rebased #6785 against `maint` and then extended the test to actually test that we can get data from that remote for which we specified the now fixed URL.  While at it, as described in https://github.com/datalad/datalad/pull/6785#issuecomment-1172659104 I have ran into another peculiarity of git-annex which would have made #6785 fail on some versions of git-annex ~~, and thus made that remote be served from another URL.  Now seems all to work with both bleeding edge and not so versions.~~ . It is "fun" -- most recent version works it out fine - just replaces one URL with another since UUID is the same (thus makes sense):

```
$> git show git-annex:remote.log
d7d23c22-a9d6-4f88-987d-8b671db81d4e autoenable=true location=http://127.0.0.1:43859/.git name=mike2 type=git timestamp=165670508

$> git show git-annex:remote.log
d7d23c22-a9d6-4f88-987d-8b671db81d4e autoenable=true location=http://127.0.0.1:40665/.git name=fresh-sr type=git timestamp=1656705139.816543473s
```

With that, again, we would not be testing those both cases ('configured' vs 'added' with url), so evolved test to do have two different repos (UUIDs) with two urls and test explicitly that we can get from any of them (so they are autoenabled etc)

### DIFF
--- a/datalad/distribution/tests/test_siblings.py
+++ b/datalad/distribution/tests/test_siblings.py
@@ -581,10 +581,11 @@ def test_as_common_datasource(testbed, viapath, viaurl, remotepath, url, remotep
     # and the other one
     assert_status('ok', testbed.get('testfile2'))
 
-    # Let's get explicitly from 'fresh' remote which would not work if URL
-    # above is wrong etc
-    assert_status('ok', testbed.drop('testfile'))
-    assert_status('ok', testbed.get('testfile', source="fresh-sr"))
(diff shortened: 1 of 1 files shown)


mih (approved):
LGTM. Thx!

yarikoptic:
fails unrelated

Issue #6791, opened 2022-07-02
`status` reports typechange of prev subds record as "clean"


mih:
Discovered when investigating  #5418

Demo

```sh
datalad create ds
cd ds
datalad create -d . subds
rm -rf subds
touch subds
```

at this point `git-status` reports

```
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        typechange: subds
```

but `datalad-status`

```
{
  "action": "status",
  "gitshasum": "deb687a2fa360d4df134dae95ac552d5c59fd994",
  "parentds": "/tmp/ds",
  "path": "/tmp/ds/subds",
  "prev_gitshasum": "deb687a2fa360d4df134dae95ac552d5c59fd994",
  "refds": "/tmp/ds",
  "state": "clean",
  "status": "ok",
  "type": "dataset"
}
```

mih:
I found that in `GitRepo._diffstatus_get_state_props()` (the status decision-maker) such a typechange looks like this

```
from_state: {'gitshasum': '1271e2ab506ccadeb6072817cc748f68fa626aa1', 'type': 'dataset'}
to_state: {'gitshasum': '1271e2ab506ccadeb6072817cc748f68fa626aa1', 'type': 'dataset'}
modified_in_worktree: True
```

So it seems possible to detect a `typechange` here, possible as easy as including a further conditional

```
elif to_sha == from_sha and modified_in_worktree:
```

mih:
Now, that is not the solution. A simple worktree modification without a typechange looks exactly the same.

It may just be that `GitRepo.get_content_info()` needs to perform an additional `git diff-files` call in order to be able to report types reliably.

mih:
> It may just be that GitRepo.get_content_info() needs to perform an additional git diff-files call in order to be able to report types reliably. Or, likely much cheaper, the worktree modification detection in GitRepo.diffstatus() could switch from ls-files to diff-files.

https://github.com/datalad/datalad/pull/6797 is an attempt to make this happen (switch from `ls-files` to `diff-files`), but it turns out for adjusted branches, it opens a new dimension for corner cases.

In the following scenario `file1` is an empty, annexed file on an adjusted branch:

```sh
% git diff -- file1
diff --git a/file1 b/file1
--- a/file1
+++ b/file1
@@ -1 +0,0 @@
-/annex/objects/MD5E-s0--d41d8cd98f00b204e9800998ecf8427e

[...]

PR #6796, opened 2022-07-02
Enable `status/diff` to detect type-changes

mih:
Previously, type-changes (when a nature of a repository component
changes in the worktree, e.g., a subdataset is suddenly a file) where
not detected by `status/diff()`, because the `gitsha` report in the
does not differ between the "from" and "to" state.

This change enables this detection by combination this particular
characteristic with a "modified in working tree" indicator to
identify typechanges.

In such cases the previous `type` is not reported as the `prev_type`
property. The new type is determined by a dedicated call to `git
diff-files`, one per each typechange -- under the assumption that such
dataset modifications are rare.

For such cases, the `gitsha` property is no longer included, because
it is not provided by Git and only expensive to obtain. Previously,
the `gitsha` of the previous object of the old type was wrongly
reported.

Fixes datalad/datalad#6791

This will break tests left and right for now, because our code now longer reports wrong gitshas. If it should ever get merged, #6793 should be revisited for a potential simplification.

#### 🐛 Bug Fixes
- Description... Fixes #issue
-

### DIFF
--- a/datalad/consts.py
+++ b/datalad/consts.py
@@ -81,3 +81,14 @@ RESERVED_NAMES_WIN = {'CON', 'PRN', 'AUX', 'NUL', 'COM1', 'COM2', 'COM3',
                       'LPT9'}
 # Characters that can't be a part of a file name on Windows
 ILLEGAL_CHARS_WIN = "[<>:/\\|?*\"]|[\0-\31]"
+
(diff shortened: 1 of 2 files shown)


mih:
I am closing this, because the approach taken is invalid: https://github.com/datalad/datalad/issues/6791#issuecomment-1173021184

PR #6797, opened 2022-07-03
Introduce typechange detection for status reporting and saving

mih:
The proposed changes improve the accuracy of `status/diff` reporting, and consequently the reliability with which `save` can act on datasets modified in more complex ways (files replaced by subdatasets, etc.)

I cannot anticipate to which degree this change is appreciated in datalad-core, given that the added accuracy comes with a runtime cost. I nevertheless wanted to develop and proposed it here.

Should this change be rejected, it may still make sense to ingest the included test, to document which dataset modification scenarios are not supported.

And https://github.com/datalad/datalad/pull/6797/commits/990bc16d42295ebbef496e8135d45ac95949d80a should likely be adopted nevertheless.

I will now detach here and extract this essence of this PR in a patch for inclusion in `datalad-next`. Further work on `diff --patch` and diff drivers will continue outside `-core`.

TODO:

- [x] check whether changes possibly already merged from #6793 require an update or could now be done cheaper/simpler
- [x] add test for typechange reporting from `diffstatus()` as suitable test to add this to sits in #6793
- [x] expand type annotation of diff results, record previous and new type
- [x] the current implementation leaves the working tree modified after a safe on adjusted branches. This is an unintentional change and needs to be investigated

### DIFF
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -2989,16 +2989,18 @@ class GitRepo(CoreGitRepo):
             _cache[key] = from_state
 
         status = OrderedDict()
-        for f, to_state_r in to_state.items():
+        # for all paths we know now or knew before
+        for f in set(from_state).union(to_state):
             props = self._diffstatus_get_state_props(
                 f,
-                from_state.get(f, None),
-                to_state_r,
+                from_state.get(f),
+                to_state.get(f),
                 # are we comparing against a recorded commit or the worktree
                 to is not None,
-                # if we have worktree modification info, report if
-                # path is reported as modified in it
-                modified and f in modified,
+                # if we have worktree modification info, pass type(change)
+                # report on, if there is any
+                False if modified is None
+                else modified.get(f, False),
                 eval_submodule_state)
             # potential early exit in "global" eval mode
             if eval_submodule_state == 'global' and \
@@ -3006,21 +3008,6 @@ class GitRepo(CoreGitRepo):
                 # any modification means globally 'modified'
                 return 'modified'
             status[f] = props
-
-        for f, from_state_r in from_state.items():
-            if f not in to_state:
-                # we new this, but now it is gone and Git is not complaining
-                # about it being missing -> properly deleted and deletion
-                # stages
-                status[f] = dict(
-                    state='deleted',
-                    type=from_state_r['type'],
-                    # report the shasum to distinguish from a plainly vanished
-                    # file
-                    gitshasum=from_state_r['gitshasum'],
-                )
-                if eval_submodule_state == 'global':
-                    return 'modified'
 
         if to is not None or eval_submodule_state == 'no':
             # if we have `to` we are specifically comparing against
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -3083,20 +3083,51 @@ class GitRepo(CoreGitRepo):
             return status
 
     def _get_worktree_modifications(self, paths=None):
-        # from Git 2.31.0 onwards ls-files has --deduplicate
-        # by for backward compatibility keep doing deduplication here
-        modified = set(
-            self.pathobj.joinpath(ut.PurePosixPath(p))
-            for p in self.call_git_items_(
-                # we must also look for deleted files, for the logic
-                # below to work. Only from Git 2.31.0 would they be
-                # included with `-m` alone
-                ['ls-files', '-z', '-m', '-d'],
-                # low-level code cannot handle pathobjs
-                files=[str(p) for p in paths] if paths else None,
-                sep='\0',
-                read_only=True)
-            if p)
+        """Report working tree modifications
+
+        Parameters
+        ----------
+        paths : list or None
+          If given, limits the query to the specified paths. To query all
+          paths specify `None`, not an empty list.
+
+        Returns
+        -------
+        dict
+          Mapping of Paths to type labels from GIT_MODE_TYPE_MAP.
+          Deleted paths have type `None` assigned.
+        """
+        # ask diff-files, gives a list with interspersed diff info and
+        # filenames
+        mod = list(self.call_git_items_(
+            ['diff-files',
+             # without this, diff-files would run a full status (recursively)
+             # but we are at most interested in a subproject commit
+             # change within the scope of this repo
+             '--ignore-submodules=dirty',
+             # hopefully making things faster by turning off features
+             # we would not benefit from (at least for now)
+             '--no-renames',
+             '-z'
+            ],
+            # low-level code cannot handle pathobjs
+            files=[str(p) for p in paths] if paths else None,
+            sep='\0',
+            read_only=True))
+        # convert into a mapping path to type
+        modified = dict(zip(
+            # paths are every other element, starting from the second
+            mod[1::2],
+            # mark `None` for deletions, and give raw mode otherwise
+            (None if l.endswith('D') else l.split(' ', maxsplit=2)[1]
+             for l in mod[::2])
+        ))
+        # convenience-map to type labels, leave raw mode if unrecognized
+        # (which really should not happen)
+        modified = {
+            self.pathobj / ut.PurePosixPath(k):
+            GIT_MODE_TYPE_MAP.get(v, v) for k, v in modified.items()
+        }
         return modified
 
     def _diffstatus_get_state_props(self, f, from_state, to_state,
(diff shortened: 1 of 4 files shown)


datalad/support/gitrepo.py
```diff
@@ -3109,12 +3154,14 @@ def _diffstatus_get_state_props(self, f, from_state, to_state,
         Parameters
         ----------
         f : Path
-        from_state : dict
-        to_state : dict
+        from_state : dict or None
+        to_state : dict or None
```
yarikoptic, line 3158: what does it mean for state to be `None` -- current worktree? might be worth a docstring

datalad/support/gitrepo.py
```diff
@@ -3123,85 +3170,137 @@ def _diffstatus_get_state_props(self, f, from_state, to_state,
...
+        elif to_state and 'type' in to_state:
+            props['type'] = to_state['type']
+
+        if state == 'modified':
+            # for modifications we want to report types for both states
+            props['prev_type'] = from_state['type']
+
+        if state == 'deleted':
+            # report the type that was deleted
+            props['type'] = from_state['type']
```
yarikoptic, line 3225: shouldn't "current" state just be None since it was deleted?

datalad/support/gitrepo.py
```diff
@@ -3109,12 +3154,14 @@ def _diffstatus_get_state_props(self, f, from_state, to_state,
...
         f : Path
-        from_state : dict
-        to_state : dict
+        from_state : dict or None
+        to_state : dict or None
         against_commit : bool
           Flag whether `to_state` reflects a commit or the worktree.
-        modified_in_worktree : bool
-          Flag whether a worktree modification is reported. This is ignored
+        modified_in_worktree : False or str
```
yarikoptic, line 3161:
```suggestion
        modified_in_worktree : False or str or None
```

datalad/support/gitrepo.py
```diff
@@ -3109,12 +3154,14 @@ def _diffstatus_get_state_props(self, f, from_state, to_state,
...
+          False if there is no modification of `f` in the working tree,
+          or a type label from GIT_MODE_TYPE_MAP indicating the type
+          of the modified path `f`. This is ignored
```
yarikoptic, line 3164:
```suggestion
          of the modified path `f`. This is ignored. Could be `None` to signal that file was deleted.
```
correct (judging from the [code below](https://github.com/datalad/datalad/pull/6797/files#diff-bfc1c47d032fc72b15b6a79db3d49e4c4334cf96318c5efe48a3403744b32fb1R3214)?

datalad/support/gitrepo.py
```diff
@@ -3123,85 +3170,137 @@ def _diffstatus_get_state_props(self, f, from_state, to_state,
...
+            self, f, from_state, to_state, modified_in_worktree,
+            against_commit, eval_submodule_state):
+        """Determine the state of `f` from from_state and to_state records
+
+        Parameters
+        ----------
+        f: Path
+        from_state: dict or None
+        to_state: dict or None
+        modified_in_worktree: bool
```
yarikoptic, line 3244:
```suggestion
        modified_in_worktree: bool or None or str?
```

yarikoptic (commented):
just did some diff code reading with hope to understand it and contribute to the number of homo sapiens on planet Earth who at least understand this core pieces of DataLad core code ;) 
I left some comments/questions from a very initial review.  I think (in the long term) we could greatly benefit from code type annotations, especially in the cases of such core and complicated logic, where values could also be a `Union[str, bool, None]` where `None` is not just "not specified" but seems to really carry a semantic load ("vanished" i.e. "removed").  Filed related https://github.com/datalad/datalad/issues/6884 . If you would like, I could give a shot to initial code annotation within `support/gitrepo.py`.

datalad/support/gitrepo.py
```diff
@@ -3007,16 +2989,18 @@ def _get_cache_key(label, paths, ref, untracked=None):
...
+                from_state.get(f),
+                to_state.get(f),
                 # are we comparing against a recorded commit or the worktree
                 to is not None,
-                # if we have worktree modification info, report if
-                # path is reported as modified in it
-                modified and f in modified,
+                # if we have worktree modification info, pass type(change)
+                # report on, if there is any
+                False if modified is None
```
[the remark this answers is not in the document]
mih: It is right there, on line 2978 in the `else` clause. It indicates that there cannot be modifications in the worktree, because the worktree is not part of the comparison.

datalad/support/gitrepo.py
mih: `None` means that there is not state for this path in either the `to` or `from` part, because that path is added or removed, for example.

datalad/support/gitrepo.py
mih: I don't know what makes you suggest this change. If it is based on a condition you saw happen in the code, that condition could be a bug. The only time this helper is called ensures that `None` is not passed as an argument.

datalad/support/gitrepo.py
```diff
@@ -3123,85 +3170,137 @@ def _diffstatus_get_state_props(self, f, from_state, to_state,
...
             # assign previous gitsha to any record
             # state==None can only happen for subdatasets that
             # already existed, so also assign a sha for them
             props['prev_gitshasum'] = from_sha
+
+        # current type reporting
+        if modified_in_worktree is not False:
+            # we have a report of a modified type, include if
+            # not None (i.e. vanished)
+            if modified_in_worktree is not None:
```
mih, line 3214: This is probably the line that mode you think it should also take `None`, but I currently think this conditional is always true.

datalad/support/gitrepo.py
mih: Absolutely! https://github.com/datalad/datalad/pull/6878/commits/3fa23fcd288af6714f451c18c03d8e19aa152694 in https://github.com/datalad/datalad/pull/6878

datalad/support/gitrepo.py
mih: No, for the same reason as for `_diffstatus_get_state_props()`

mih (commented):
> just did some diff code reading with hope to understand it and contribute to the number of homo sapiens on planet Earth who at least understand this core pieces of DataLad core code ;)

Thanks A LOT for this, and even more to do it quickly, we I still remember somehow what I did and why!

I responded to your comments.

> If you would like, I could give a shot to initial code annotation within support/gitrepo.py.

If that means "in this PR", then please not. But in general, I think this would be a good idea. Although I think that taken this file apart in more manageable pieces (probably ~8-10) would be a worthwhile prior step.

datalad/support/gitrepo.py
```diff
@@ -3100,6 +3069,82 @@ def _get_cache_key(label, paths, ref, untracked=None):
...
+             for spec in mod[::2])
+        ))
+        # `diff-files` cannot give us the full answer to "what is modified"
+        # because it won't consider what smudge filters could do, for this
+        # we need `ls-files --modified` to exclude any paths that are not
+        # actually modified
+        modified_files = set(
+            p for p in self.call_git_items_(
+                # we need not look for deleted files, diff-files did that
+                ['ls-files', '-z', '-m'],
```
yarikoptic, line 3124: possible enhancement to do since there was a concern of now having two calls instead of one:  this call (`ls-files`) seems to be independent of prior `diff-files`.  . Could at least partially be mitigated by multi-threading first execution until the 2nd one is complete (AKA poor man "async" ;-) ).  We have `parallel.ProducerConsumer` but I think it would be too heavy for such one.  It is more of really a simply do it directly here without bothering with a helper until we have another use case like this.

datalad/support/gitrepo.py
yarikoptic: then it better becomes an `assert modified_in_worktree is not None` ?

bpoldrack (approved):
From my POV, this looks alright and for me @mih's answers to @yarikoptic's review are satifying :)

However, it seems @yarikoptic did not respond further or resolve conversations. So, is it OK from your POV, too, @yarikoptic ?

yarikoptic:
I am yet to rereview while having @mih insights in mind. Some comments though might still be worth addressing https://github.com/datalad/datalad/pull/6797#discussion_r929432518 since code is IMHO is far from trivial

mih:
I tried to load this in my head again after vacation and failed so far. Maybe it is better to close it, and have another attempt at some point in the future.

mih:
OK, two weeks past. I do not think I will get back to this. Let it RIP.

Issue #6798, opened 2022-07-04
Fix a race condition in ssh copy test


christian-monch:
The test `datalad.support.tests.test_sshconnector.test_ssh_copy` introduces a race condition by copying a file onto itself. This sometimes leads to an empty file and a failing test.

yarikoptic:
Part of https://github.com/datalad/datalad/issues/6655 . Emptying is a behavior of new versions of ssh

bpoldrack:
I think this aspect is actually addressed in #6799

PR #6799, opened 2022-07-04
Fix a race condition in ssh copy test

christian-monch:
Fixes issue #6798 

This PR modifies the test `datalad.support.tests.test_sshconnector::test_ssh_copy` so that it does not longer populate a source directory by copying existing files onto themselves, thereby remove a race condition that left a source file empty.

#### 🛡 Tests
- remove a race condition in ssh copy test

### DIFF
--- a/datalad/support/tests/test_sshconnector.py
+++ b/datalad/support/tests/test_sshconnector.py
@@ -204,8 +204,8 @@ def test_ssh_manager_close_no_throw(bogus_socket=None):
 @skip_if_on_windows
 @skip_ssh
 @with_tempfile(mkdir=True)
-@with_tempfile(content="one")
-@with_tempfile(content="two")
+@with_tempfile(content='one')
+@with_tempfile(content='two')
 def test_ssh_copy(sourcedir=None, sourcefile1=None, sourcefile2=None):
     port = get_ssh_port('datalad-test')
     remote_url = 'ssh://datalad-test:{}'.format(port)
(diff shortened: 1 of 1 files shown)


datalad/support/tests/test_sshconnector.py
```diff
@@ -212,14 +212,18 @@ def test_ssh_copy(sourcedir=None, sourcefile1=None, sourcefile2=None):
...
+    obscure_path = opj(sourcedir, obscure_file)
+    with open(obscure_path, 'w') as f:
         f.write("three")
 
-    # copy tempfile list to remote_url:sourcedir
-    sourcefiles = [sourcefile1, sourcefile2, obscure_file]
+    # copy first two temp files to remote_url:sourcedir
+    sourcefiles = [sourcefile1, sourcefile2]
     ssh.put(sourcefiles, opj(remote_url, sourcedir))
+    # copy obscure file to remote_url:sourcedir/'<obscure_file_name>.c opy'
```
yarikoptic, line 224: Clarify in the comment why you got into another name

yarikoptic:
BTW why not against `maint`?

bpoldrack:
> I don't see it being related to this PR but we neither saw it before happening -- what has triggered it?

Not sure, what triggers it, but happens elsewhere, too. Example: PR #6793 

> BTW why not against maint?

Would agree to put it `maint`, but I'm preparing to release 0.17, so not making a difference really.

bpoldrack (approved):

PR #6802, opened 2022-07-05
DOC: fix and move docstring reference into design docs

### LINKED ISSUE #6801: DOC: Add docstring formats into design docs
The contributing file specifies that we follow the "numpy docstring standard", and points to a link that doesn't resolve ([this one](https://github.com/numpy/numpy/blob/main/doc/HOWTO_DOCUMENT.rst.txt#docstring-standard)). We should clarify the docstring format in the [design document on docstrings](http://docs.datalad.org/en/stable/design/docstrings.html)

adswa:
This change moves information on how docstrings are
to be designed from CONTRIBUTING to the Design docs
as the Design docs have become a place to consolidate
information like this and already contain a section
on docstrings. At the same time, it fixes the link
to the numpy docstring standard that did not resolve
anymore, and replaces the 'Documentation' section in
CONTRIBUTING, which only contained docstring info, with
a more general pointer to the design documents further
up in the file.
Fixes #6801

#### 📝 Documentation
- Different aspects of information on docstrings were consolidated in the design documentation

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -122,6 +122,9 @@ we outline the workflow used by the developers:
 
 (If any of the above seems like magic to you, then look up the
 [Git documentation](http://git-scm.com/documentation) on the web.)
+Our [Design Docs](http://docs.datalad.org/en/stable/design/index.html) provide a
+growing collection of insights on the command API principles and the design of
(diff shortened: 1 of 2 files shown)


yarikoptic:
Sounds good, let's proceed, thanks @adswa !

PR #6789, opened 2022-07-01
BF: use --python-match minor option in new datalad-installer release to match outside version of python

yarikoptic:
@bpoldrack spotted that we are not actually testing against the python version
we set in our matrix runs of travis CI setup.  with datalad-installer 0.9.0 we
can now match to minor (or try even more detailed match) portion of the version
of system wide python. Again -- it would not be system wide python intallation,
but conda installation of python matching in minor portion of the version, which
is what we care in our matrix

PR #6803, opened 2022-07-05
BF: use --python-match minor option in new datalad-installer release to match outside version of python (master version)

yarikoptic:
Master based version of #6789 since we might be unable to install nose in fresh python envs.

we set in our matrix runs of travis CI setup.  with datalad-installer 0.9.0 we
can now match to minor (or try even more detailed match) portion of the version
of system wide python. Again -- it would not be system wide python intallation,
but conda installation of python matching in minor portion of the version, which
is what we care in our matrix

This one is positioned against master since maint we run into difficulty
installing nose in env with python 3.10

TODOs
- [ ] we might need to seek completely remove `nose` from `devel` installation scenario
- [x] waiting on #6808 to be accepted to maint and merged into matser
- [ ] waiting on #6806 to be resolved (due to newer ssh, gets installed here with newer git-annex in python 3.10 env)

#### 🛡 Tests
- Fix Travis CI setup to test against specific versions of Python while using datalad-installer to install git-annex in miniconda.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,6 +60,9 @@ matrix:
     env:
       - PYTEST_SELECTION=
       - PYTEST_SELECTION_OP=not
+      # current "default" version of git-annex 8.20201007  is not co-installable
+      # with python 3.10, so let's just allow for any git-annex version
+      - _DL_ANNEX_INSTALL_SCENARIO="miniconda --python-match minor --batch git-annex -m conda"
   - if: type = cron
     python: 3.7
     # Single run for Python 3.7
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,7 @@ env:
     - DATALAD_DATASETS_TOPURL=https://datasets-tests.datalad.org
     # How/which git-annex we install.  conda's build would be the fastest, but it must not
     # get ahead in PATH to not shadow travis' python
-    - _DL_ANNEX_INSTALL_SCENARIO="miniconda --batch git-annex=8.20201007 -m conda"
+    - _DL_ANNEX_INSTALL_SCENARIO="miniconda --python-match minor --batch git-annex=8.20201007 -m conda"
 
 matrix:
   include:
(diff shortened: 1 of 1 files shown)


yarikoptic:
Finally some "sensible" test fails, on 3.10:

```
______________________________ test_inside_async _______________________________
    def test_inside_async():
        async def main():
            runner = Runner()
            return runner.run(
                (["cmd.exe", "/c"] if on_windows else []) + ["echo", "abc"],
                StdOutCapture)
    
>       loop = asyncio.get_event_loop()
E       DeprecationWarning: There is no current event loop
../../../../.local/lib/python3.10/site-packages/datalad/runner/tests/test_nonasyncrunner.py:378: DeprecationWarning
________________________________ test_ssh_copy _________________________________
sourcedir = '/tmp/datalad_temp_test_ssh_copy0_mtjaxj'
sourcefile1 = '/tmp/datalad_temp_test_ssh_copye_0vqnzz'
sourcefile2 = '/tmp/datalad_temp_test_ssh_copyh3e9voze'
    @skip_if_on_windows
    @skip_ssh
    @with_tempfile(mkdir=True)
    @with_tempfile(content="one")
    @with_tempfile(content="two")
    def test_ssh_copy(sourcedir=None, sourcefile1=None, sourcefile2=None):
        port = get_ssh_port('datalad-test')
        remote_url = 'ssh://datalad-test:{}'.format(port)
        manager = SSHManager()
        ssh = manager.get_connection(remote_url)
    
        # write to obscurely named file in sourcedir
        obscure_file = opj(sourcedir, get_most_obscure_supported_name())
        with open(obscure_file, 'w') as f:
            f.write("three")
    
        # copy tempfile list to remote_url:sourcedir
        sourcefiles = [sourcefile1, sourcefile2, obscure_file]
        ssh.put(sourcefiles, opj(remote_url, sourcedir))
        # docs promise that connection is auto-opened in case of multiplex
        if _ssh_manager_is_multiplex:
            ok_(ssh.is_open())
    
        # recursive copy tempdir to remote_url:targetdir
        targetdir = sourcedir + '.c opy'
        ssh.put(sourcedir, opj(remote_url, targetdir),
                recursive=True, preserve_attrs=True)
    
        # check if sourcedir copied to remote_url:targetdir
>       ok_(isdir(targetdir))
../../../../.local/lib/python3.10/site-packages/datalad/support/tests/test_sshconnector.py:233: 
expr = False, msg = None
    def assert_true(expr, msg=None):
        if msg is None:
>           assert expr
E           assert False
../../../../.local/lib/python3.10/site-packages/datalad/tests/utils_pytest.py:191: AssertionError
```

note that it also installs version of git-annex in that environment different from other matrix runs, so fail might not be specifically 3.10 but also relating to git-annex.  @christian-monch since those are runner/ssh related -- any clues?

yarikoptic:
now that we released replaced by https://github.com/datalad/datalad/pull/6827 against maint

PR #6805, opened 2022-07-06
Definition of a release manager role

mih:
Based on the discussion in the developer call on 2022-07-5, I am proposing this definition of the role of a release manager.

Given that we have not done such thing before, I propose the following procedure

- Discussion phase for 7 days (until 2022-07-13, noon UTC), during which any comments or suggestions for changes are recorded and discussed.
- After the discussion phase, the proposer (@mih in this case) compiles a revision of the document in this PR, and it is put up for a vote.
- Voting happens publicly by posting a response to this PR. Choices:
  - **accept**: adopt the role as defined at the time of the vote
  - **reject**: there should be no release manager role
  - **further discussion**: there should be a release manager role, but not as defined in the description at the time of the vote

### DIFF
--- a/policy/release_manager.md
+++ b/policy/release_manager.md
@@ -1,0 +1,49 @@
+# Release manager role
+
+## Appointment
+
+A release manager (RM) is appointed for a specific upcoming release.
+
+A RM may and should step down at any time, when they cannot fulfill the associated obligations. In this case, a new RM is appointed.
+
+The appointment of an RM ends with the declaration of the respective release's end-of-life by the RM.
+
+## Responsibilities
+
+Immediately after the previous release was made, the RM takes over decision making regarding merges into the integration branch for all developments targeting the respective release. Such merges can be made as soon as, but no earlier, than the RM indicating approval.
+
+It is the RM's responsibility to act on any merge request in a timely manner.
+
+An RM is not expected to be an expert in all techniques, features, and parts of the code base. Consequently, an RM should seek feedback prior to approving merge requests whenever necessary.
+
+The RM declares the time of a feature freeze for the upcoming release.
+
+The RM decides on the length of the consolidation period following a feature freeze, and decides which kinds of changes are still acceptable.
+
+The RM decides on the number and timing of potential release candidates, and on the timing of the final release.
+
+After the initial release, the RM takes over the decision making for the respective maintenance branch in the same fashion as for the prior development towards the initial release. The RM decides on the nature of changes qualified for maintenance releases, and their timing.
+
+
+## Assistant release manager role
+
+In addition to an RM there is an assistant RM to prevent disruption in cases where the RM temporarily cannot act on their duties.
+
+The assistant RM role is identical to the RM, with the same responsibilities. 
+
+
+## Transition of release manager appointments
+
+The assistant RM (aRM) becomes the RM when the current RM completed a release cycle and took over the new maintenance series. 
+
+The aRM becomes the RM when the current RM steps down prior completion of the release cycle.
+
+The RM of an upcoming release serves as the aRM of the current maintenance series.
+
+An aRM for the upcoming release is appointed by the project at the start of a release cycle.
+
+Consequently, there is an RM and aRM for any given release series. A RM will therefore typically transition through the following stages
+
+- appointment as aRM for the upcoming release (R+1)
+- appointment as RM for release R+2 and continuing to be the aRM for the release R+1
+- RM for the maintenance phase of release R+2 until its end-of-life


policy/release_manager.md
```diff
@@ -0,0 +1,49 @@
+# Release manager role
+
+## Appointment
+
+A release manager (RM) is appointed for a specific upcoming release.
```
adswa, line 5: Maybe we should add info on a few edge cases or common cases on how this appointing works? Eg, you can't be made RM if you don't want it yourself. And what happens in the hypothetical situation that someone in the team objects an RM, or if there are more volunteers for the position than roles?

policy/release_manager.md
```diff
@@ -0,0 +1,49 @@
...
+
+A RM may and should step down at any time, when they cannot fulfill the associated obligations. In this case, a new RM is appointed.
+
+The appointment of an RM ends with the declaration of the respective release's end-of-life by the RM.
+
+## Responsibilities
```
adswa, line 11: is making the actual release, including Changelog generation, also part of the responsibilities?

policy/release_manager.md
```diff
@@ -0,0 +1,49 @@
...
+
+An RM is not expected to be an expert in all techniques, features, and parts of the code base. Consequently, an RM should seek feedback prior to approving merge requests whenever necessary.
+
+The RM declares the time of a feature freeze for the upcoming release.
+
+The RM decides on the length of the consolidation period following a feature freeze, and decides which kinds of changes are still acceptable.
+
+The RM decides on the number and timing of potential release candidates, and on the timing of the final release.
+
+After the initial release, the RM takes over the decision making for the respective maintenance branch in the same fashion as for the prior development towards the initial release. The RM decides on the nature of changes qualified for maintenance releases, and their timing.
```
adswa, line 25: Just so that I grasp the concept fully: That means that if I am release manager for 0.17 I act on all PRs targeting ``master``, and after 0.17 release I act on all PRs targeting ``maint``, correct?

policy/release_manager.md
mih: I would be reluctant to include it, but would not object if there happens to be an agreeable statement. I think this document has a lot of explicit rules already, in comparison to what we had before. I personally made bad experiences trying to upfront-regulate a process that has never been tried before. So in some sense, my proposal is a declarative description of something that should match was we had in the past too, but turns it into an explicit role that a person can know to have, and adds the concept of having two people for resilience at all times.

I would prefer to stay vague on the appointment process. I consider it unlikely that we what try to make a person an RM that does not want to be in that role. If there is a competition for that role, we a 1) very lucky, and 2) I am confident that we can find a solution -- we are still very few people involved.

Once this has proven to work, we could formalize it more.

policy/release_manager.md
mih: I'd say: There cannot be a release without a changelog. I think it would not be useful to state that a RM needs to personally craft one, but they obviously should make sure that one is made by some processes.

If that changelog is bad, it can be fixed, and complained about (like any other issue), but this is entirely orthogonal to the RM role definition.

policy/release_manager.md
mih: Yes.

(Although the names of these branches are not enshrined here. This would be more appropriate for the description of the actual technical release process, rather than the description of the person role. I'd say these two should be different)

adswa: I think then we should add information and authorization on making releases, right? Or is everything automated already, e.g., upload to PyPi?

mih: I cannot speak to automation -- I am largely unaware of it.

Can you give an example, what you mean by "information and authorization on making releases"?

I would also caution against an inflation of the RM role to including packaging.

From my POV a release is done by tagging a state in the Git repo. End of story.

Because we are presently using the GitHub platform, it makes sense to push that tag, and create a "GitHub release" too.

PyPi, to me, is a type of distribution, just like Conda, or Debian. I see updating all these channel in the responsibility of the respective package maintainers.

One could argue that an upload to PyPi is *that* simple, that a RM could also do it. But I think it is not a given implication.

adswa: > From my POV a release is done by tagging a state in the Git repo. End of story.

Cool, thx for clarifying.

policy/release_manager.md
```diff
@@ -0,0 +1,49 @@
...
+In addition to an RM there is an assistant RM to prevent disruption in cases where the RM temporarily cannot act on their duties.
+
+The assistant RM role is identical to the RM, with the same responsibilities. 
+
+
+## Transition of release manager appointments
+
+The assistant RM (aRM) becomes the RM when the current RM completed a release cycle and took over the new maintenance series. 
+
+The aRM becomes the RM when the current RM steps down prior completion of the release cycle.
```
jsheunis, line 39: prior *to*

policy/release_manager.md
```diff
@@ -0,0 +1,49 @@
...
+
+A release manager (RM) is appointed for a specific upcoming release.
+
+A RM may and should step down at any time, when they cannot fulfill the associated obligations. In this case, a new RM is appointed.
+
+The appointment of an RM ends with the declaration of the respective release's end-of-life by the RM.
+
+## Responsibilities
+
+Immediately after the previous release was made, the RM takes over decision making regarding merges into the integration branch for all developments targeting the respective release. Such merges can be made as soon as, but no earlier, than the RM indicating approval.
```
jsheunis, line 13:
```suggestion
Immediately after the previous release was made, the RM takes over decision making regarding merges into the integration branch for all developments targeting the respective release. Such merges can be made as soon as, but no earlier than, the RM indicates approval.
```

policy/release_manager.md
jsheunis, line 39:
```suggestion
The aRM becomes the RM when the current RM steps down prior to completion of the release cycle.
```

policy/release_manager.md
christian-monch: Again a question regarding `master` and `maint`. I am using the names here because I do not find a clear definition that would, for example, define "development-branch", "integration-branch", or "maintenance-branch". Is [issue #4238](https://github.com/datalad/datalad/issues/4238) still governing the release process? (If so, should it not say in the last comment before 4.i: "merge maint into master"?)

If understand the process outlined in #4238 correctly, new features go into `master` and bug-fixes into  `maint`. So should an RM not focus on both branches? Or do we assume that bug-fixes are always prioritized, so no RM decision is needed w.r.t. bug-fixes?

jsheunis:
I've read through the proposal and it all seems clear and useful to me, including the comments from @adswa and the resulting back and forth. Happy for this to start happening.

policy/release_manager.md
```diff
@@ -0,0 +1,49 @@
+# Release manager role
+
+## Appointment
+
+A release manager (RM) is appointed for a specific upcoming release.
+
+A RM may and should step down at any time, when they cannot fulfill the associated obligations. In this case, a new RM is appointed.
+
+The appointment of an RM ends with the declaration of the respective release's end-of-life by the RM.
```
mslw, line 9: I was initially confused by what *end-of-life* means, but it was probably due to my initial misconception that RM is related to release *preparation*, and not preparation and subsequent maintenance of the MINOR version.

On re-read, I think it should have been obvious to me that release = release life cycle, not the moment at which it is released; and in any case the rest of the document clarifies it sufficiently. So I'd keep it as is. But if someone else thinks it's not immediately clear, then I'd suggest adding a preamble statement (before appointment section) like "the RM oversees the release life cycle, including preparation and subsequent maintenance".

mslw:
I've read it and I have nothing to add. I think it's clear and concise. I left one comment, but I feel that the confusion was more on my side than the text's.

policy/release_manager.md
mih: I am not responding to the first paragraph here. I think whether or not #4238 should (also) stay/become policy is a worthwhile discussion, but a separate one.

> So should an RM not focus on both branches?

No, there is a dedicated RM for in charge of each of them.

> Or do we assume that bug-fixes are always prioritized, so no RM decision is needed w.r.t. bug-fixes?

The RM for the upcoming release is also the assistant RM for the current release (see last section in the document). So procedurally, the decision making is happening on a team of two, or within a single person. I think this is not too complex.

But I want to emphasize again that this document is written in the spirit of a volunteer-driven project. So if a PR is opened for the upcoming release, and the RM for the current release thinks it better goes there (first), this person can *ask* for the change -- but there is no guarantee that such a change will be made by the volunteer contributor. Both RMs in charge  can then act in any which way they think is best (as it is already the case: rebase, or cherrypick to maint, etc).

christian-monch: Thx for clarifying. I misunderstood the scope of release.

policy/release_manager.md
christian-monch, line 5:
```suggestion
A release manager (RM) is appointed for a specific upcoming minor or major release, i.e. `x.y.z` -> "x.y+1.0` or `x.y.z` -> `x+1.0.0`.
```

christian-monch:
I think the role is defined well and I believe it will benefit the project

policy/release_manager.md
mih: Makes sense. I wanted to stay away from it, because of possible changes after `1.0`, but I think in this way it just works.

mih:
Given that we confirmed in today's dev-call that everyone was heard, I think it is fine to shorten the discussion period.

I have pushed minor changes to the proposal

- a clarification: https://github.com/datalad/datalad/pull/6805/commits/4b37b773e52bfb80bd3ef04475ad7fd386d54aa1
- reformating: https://github.com/datalad/datalad/pull/6805/commits/9ce04f98ad536961aaf3565f303aa3e9da7b5407

I am hereby putting this proposal (https://github.com/datalad/datalad/pull/6805/commits/9ce04f98ad536961aaf3565f303aa3e9da7b5407) up for a vote. Please respond to this PR with a comment indicating your vote. Choices are

----
  - **accept**: adopt the role as defined at the time of the vote
  - **reject**: there should be no release manager role
  - **further discussion**: there should be a release manager role, but not as defined in the description at the time of the vote
----

The voting period ends at 2022-07-19, noon UTC.

After the vote, this policy addition is either adopted or not. In case of adoption, any changes can be proposed and voted on at any time, by posting a PR that changes the policy document in the desired fashion.

mih:
**accept**

jsheunis:
**accept**

policy/release_manager.md
```diff
@@ -0,0 +1,54 @@
...
+
+A RM may and should step down at any time, when they cannot fulfill the associated obligations.
+In this case, a new RM is appointed.
+
+The appointment of an RM ends with the declaration of the respective release's end-of-life by the RM.
+
+## Responsibilities
+
+Immediately after the previous release was made, the RM takes over decision making regarding merges into the integration branch for all developments targeting the respective release.
+Such merges can be made as soon as, but no earlier than, the RM indicates approval.
```
yarikoptic, line 15: just to make it clear -- merges could be made by anyone with merge permissions, not necessarily RM, as long as RM indicates approval (but not necessarily clicks "Merge").

It makes procedure vague, i.e. should I monitor my PR and merge it if it is approved, or could I just rely that RM would merge it.  Since it is kinda a procedure now, let's keep it as vague as it is now, can improve later.  On the other hand, why would RM not merge it if already approved?

policy/release_manager.md
```diff
@@ -0,0 +1,54 @@
...
+
+The RM decides on the number and timing of potential release candidates, and on the timing of the final release.
+
+After the initial release, the RM takes over the decision making for the respective maintenance branch in the same fashion as for the prior development towards the initial release.
+The RM decides on the nature of changes qualified for maintenance releases, and their timing.
+
+
+## Assistant release manager role
+
+In addition to an RM there is an assistant RM to prevent disruption in cases where the RM temporarily cannot act on their duties.
```
yarikoptic, line 34: who is assigning/deciding on being an assistant RM?  is it like a president+prime to come into position at the same time?

Should it just be "next RM" to serve as "assistant RM" for current cycle, thus removing this ambiguity and possibly gaining some needed training?  if so

```suggestion
RM appointed for the next cycle serves as an assistant RM to prevent disruption in cases where the RM temporarily cannot act on their duties.
```

policy/release_manager.md
```diff
@@ -0,0 +1,54 @@
...
+The assistant RM role is identical to the RM, with the same responsibilities. 
+
+
+## Transition of release manager appointments
+
+The assistant RM (aRM) becomes the RM when the current RM completed a release cycle and took over the new maintenance series. 
+
+The aRM becomes the RM when the current RM steps down prior to completion of the release cycle.
+
+The RM of an upcoming release serves as the aRM of the current maintenance series.
```
yarikoptic, line 45: ah -- so my comment/"suggestion" above was right on spot. So "aRM" is like  an "RM-elect" or "next-RM", but those names aren't better so I guess ok as is although flow of introduction of "aRM" role is a bit off: the point is that we need to always have 2 RMs, and I think may be that should be stated at the opening where we talk about `Appointments`.  I will try to suggest that now.

policy/release_manager.md
```diff
@@ -0,0 +1,54 @@
+# Release manager role
+
+## Appointment
+
+A release manager (RM) is appointed for a specific upcoming major or minor release (i.e., x.y.z -> x.y+1.0 or x.y.z -> x+1.0.0).
```
yarikoptic, line 5:
```suggestion
A release manager (RM) is appointed for a specific upcoming major or minor release (i.e., x.y.z -> x.y+1.0 or x.y.z -> x+1.0.0).

Because RM serves as an Assistant Release Manager (aRM) for prior maintenance release (see below), RM should be appointed 1 release cycle prior to the release he/she would be responsible for.
```

smth like that ;) i.e. if we consider @bpoldrack (as the one cutting 0.17.0 release) being responsible for `0.17` release cycle, we should have decided for RM for `0.18` right after `0.16.0` release since that is when RM-0.18 should serve as an aRM for 0.17 series whenever @bpoldrack is the RM.

yarikoptic: With that in mind, since both @bpoldrack did 0.17.0 and I did 0.17.1 both of us could be RMs for 0.17 series. I can do it, or would you @bpoldrack would like to ?  and then we need to decide on RM for 0.18 asap to follow this design

policy/release_manager.md
```diff
@@ -0,0 +1,54 @@
...
+The RM of an upcoming release serves as the aRM of the current maintenance series.
+
+An aRM for the upcoming release is appointed by the project at the start of a release cycle.
+
+Consequently, there is an RM and aRM for any given release series.
+A RM will therefore typically transition through the following stages
+
+- appointment as aRM for the upcoming release (R+1)
+- appointment as RM for release R+2 and continuing to be the aRM for the release R+1
+- RM for the maintenance phase of release R+2 until its end-of-life
```
yarikoptic, line 54: ok, this describes what I have tried to capture at the beginning since it talks about Appointment. Overall that Appointment section then is of no value if it is actually described here and above in how to "transition".

mih:
Given the comments arriving after the start of the voting, there seems to be little point in continuing here.

mih:
Given that there will be little chance to figure out what needs to be done here in a call, I am closing this PR. There is no point in leaving it hanging here. I had asked if someone wants to take the process over, but there was no expression of interest.

The PR can be reopened once the interest in establishing such a role rekindles.

PR #6807, opened 2022-07-06
Fix broken annex symlink after git-mv before saving

yarikoptic:
This is a single forgotten to be CPed commit from one of the PRs by @mih sent to `master`.  I think I might have already did some misdemeanor and pushed some cherry-picks directly to `maint` without a PR.

With this PR we would be at 0.16.6-26-gedfcfabdb and worth release'ing a quick bug fix release (would also help to bring datalad/git-annex testing into a bit more green land on 'release')

### DIFF
--- a/datalad/core/local/tests/test_save.py
+++ b/datalad/core/local/tests/test_save.py
@@ -938,6 +938,19 @@ def test_save_diff_ignore_submodules_config(path):
     assert_repo_status(ds.path)
 
 
+@with_tree({"subdir": {"foo": "foocontent"}})
+def test_save_git_mv_fixup(path=None):
+    ds = Dataset(path).create(force=True)
+    ds.save()
+    assert_repo_status(ds.path)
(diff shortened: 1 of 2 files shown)


yarikoptic:
travis - flaky test on nfs which I have not yet worked around, appveyor -- mac AFAIK it is a known git annex regression which was just fixed upstream https://git-annex.branchable.com/bugs/regression__58___annex_add_of_moved_file_errors_out/  . Let's proceed

yarikoptic:
ha -- I *screwed up* -- this was intended to go against `maint` but went against `master` FWIW. Since https://github.com/yarikoptic/datalad/blob/HEAD/.github/workflows/release.yml#L7 is reacting only to `maint` changes, no release was even tried.

CP of this commit was submitted as a part of https://github.com/datalad/datalad/pull/6809 (sorry -- no time for per PR cherry picking, and yes - changelog now will be even less optimal)

PR #6809, opened 2022-07-06
Fix broken annex symlink after git-mv before saving + fix a race condition in ssh copy test

yarikoptic:
Bundle of two cherry picked PRs.


yarikoptic:
you are so green to me! lovely, since only CPs from master, I guess no objections.

Issue #6804, opened 2022-07-05
../datalad/support/tests/test_annexrepo.py::test_is_available[False]  is stalling in master


yarikoptic:
e.g. https://ci.appveyor.com/project/mih/datalad/builds/44082389/job/mw5l35ntiak9xmaa of https://github.com/datalad/datalad/pull/6803 :

```
PASSED
../datalad/support/tests/test_annexrepo.py::test_is_available[False] 
```
but also came up in other scenarios recently in e.g. https://github.com/datalad/datalad/pull/6518#issuecomment-1171417710 

more hits from our logs archives which show that it is not a new problem
```
(git)smaug:/mnt/datasets/datalad/ci/logs/2022/07[master]git
$> git grep -B2 'No output has been received in the last' | grep ::test
01/cron/20220701T193907/797dde3/travis-14182-errored/16.txt-../datalad/local/tests/test_run_procedure.py::test_text2git 
01/pr/6518/3496925/travis-14176-errored/4.txt-../datalad/support/tests/test_annexrepo.py::test_is_available[False] 
02/pr/6790/73d48cc/travis-14189-errored/4.txt-../datalad/support/tests/test_annexrepo.py::test_is_available[False] 
05/pr/6803/fcf849c/travis-14211-errored/1.txt-../datalad/support/tests/test_annexrepo.py::test_is_available[False] 

$> cd ../06

(git)smaug:/mnt/datasets/datalad/ci/logs/2022/06[master]git
$> git grep -B2 'No output has been received in the last' | grep ::test
03/cron/20220603T193647/173dc6f/travis-14100-errored/16.txt-../datalad/local/tests/test_run_procedure.py::test_text2git 
08/pr/6690/ada1778/travis-14121-errored/4.txt-../datalad/support/tests/test_annexrepo.py::test_is_available[False] 
09/push/master/4d53077/travis-14129-errored/4.txt-../datalad/support/tests/test_annexrepo.py::test_is_available[False] 
10/cron/20220610T193717/029f839/travis-14136-errored/16.txt-../datalad/local/tests/test_run_procedure.py::test_text2git 
10/cron/20220610T193717/029f839/travis-14136-errored/5.txt-../datalad/support/tests/test_annexrepo.py::test_is_available[False] 
13/push/master/0654290/travis-14140-errored/4.txt-../datalad/support/tests/test_annexrepo.py::test_is_available[False] 
[...]

PR #6810, opened 2022-07-06
Troubleshoot stalling at times is_available test

yarikoptic:
ref: https://github.com/datalad/datalad/issues/6804

most likely -- addressed with recent git-annex, so I will make it 100% reproducible, update git-annex to recent, if gone -- test will be conditioned on having recent git-annex


yarikoptic:
observations
- the only "failed" are the ones which didn't run the test since I guess selection (which I forgot to remove) didn't select them
- so I failed to reproduce it "get stuck" and thus can't test if boosting git-annex version would resolve
- there is **huge** difference in run times between runs. First 3 took 11 minutes each. The last two (heavy debug, and under NFS) took 90 minutes! (so 700% penalty) There were 2 runs of about 26 minutes.

yarikoptic:
I failed to troubleshoot, might get back to it some time later

PR #6817, opened 2022-07-08
Avoid import of nose-based tests.utils, make  skip_if_no_module() and skip_if_no_network() allowed at module level

jwodder:
This PR addresses the test failures seen in the conda-forge feedstock:

* A remaining import of nose-based `datalad.test.utils` is updated so that nose is no longer needed during tests.
* Calling `skip_if_no_module()` and `skip_if_no_network()` at module level (as is done in `datalad/downloaders/tests/test_s3.py`) is now allowed.

I suggest releasing this in 0.17.1 soon so that a passing release can be made on conda-forge.

#### 🛡 Tests
- Fix failing tests on conda-forge

### DIFF
--- a/datalad/runner/tests/test_threadsafety.py
+++ b/datalad/runner/tests/test_threadsafety.py
@@ -8,11 +8,12 @@ from typing import (
     Tuple,
 )
 
+from datalad.tests.utils_pytest import assert_raises
+
 from ..coreprotocols import StdOutCapture
 from ..nonasyncrunner import ThreadedRunner
 from ..protocol import GeneratorMixIn
 from .utils import py2cmd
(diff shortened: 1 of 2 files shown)


yarikoptic:
> I suggest releasing this in 0.17.1 soon so that a passing release can be made on conda-forge.

in general I would not oppose it but since diff is trivial and tests related, might be better to just take this diff as a patch for now.  Meanwhile, adjusted the PR title to provide more details which would be taken into Changelog.

yarikoptic:
travis fail is the ssh copy one, known... unrelated. Let's proceed. thank you @jwodder . more fixes will pile up and then we will release. I think I accounted for all in conda feestock but may be there would be more

Issue #6307, opened 2021-12-13
Stop using `distutils.version.LooseVersion`


jwodder:
[Distutils was officially deprecated in Python 3.10 and will be removed from the standard library completely in 3.12.](https://www.python.org/dev/peps/pep-0632/)  Most references to it were removed from Datalad by #6113, but some uses of `distutils.version.LooseVersion` remain, and there is no clear alternative that handles all use-cases.

yarikoptic:
FWIW Debian build of 0.17.0 even fails now with following warnings turned to errors

```
DeprecationWarning: The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives
```

I am a little surprised we do not observe that elsewhere, even while trying to fix testing and do test against 3.10 in #6803

effigies:
FWIW we found use cases in Nipype where looseversion was needed, as `packaging.version.Version` is for PEP-440 versions only, and `packaging.version.LegacyVersion` is being deprecated. Rather than try to come up with something else that would be close enough, I just pulled `LooseVersion` into its own package and hacked it to compare seamlessly with `distutils.version.LooseVersion`: https://pypi.org/project/looseversion/

yarikoptic:
awesome! thank you @effigies!  Now we would just need to package it for debian and conda-forge. @jwodder - could you please create conda-forge recipe for https://pypi.org/project/looseversion/ ?

effigies:
It's on conda-forge.

yarikoptic:
oh lovely -- less work for us.  I guess we can switch to use it overall for 0.18 series (0.17 was just released). python 3.12 is to be released in 2023-10 so we still have some time ;)

yarikoptic-gitmate:
Issue fixed in [`0.18.0`](https://github.com/datalad/datalad/releases/tag/0.18.0)

PR #6819, opened 2022-07-08
Suppress DeprecationWarning to allow for distutils to be used

yarikoptic:
Needs to be gone for 3.12: https://github.com/datalad/datalad/issues/6307


yarikoptic:
ok, it is minor and I think it does what it needs to do, let's proceed

PR #6820, opened 2022-07-08
BF: provide /dev/null as --git-dir instead of no value

### LINKED ISSUE #6815: datalad error on branching git config
### What is the problem?

datalad errors when a directive such as the following is presetn in .gitconfig:

```
[includeIf "gitdir:**/devbgc/**"]
    path = ~/.gitconfig_bgc
```

With error that is confusing novice users:

```
(base) twutz@macadamia tmp> datalad --version
  File "/User/homes/twutz/scratch/twutz/miniconda3/bin/datalad", line 7, in <module>
  File "/User/homes/twutz/scratch/twutz/miniconda3/lib/python3.8/site-packages/datalad/__init__.py", line 110, in <module>
    cfg = ConfigManager()
  File "/User/homes/twutz/scratch/twutz/miniconda3/lib/python3.8/site-packages/datalad/config.py", line 394, in __init__
    self.reload(force=True)
  File "/User/homes/twutz/scratch/twutz/miniconda3/lib/python3.8/site-packages/datalad/config.py", line 469, in reload
    self._stores[store_id] = self._reload(runargs)
  File "/User/homes/twutz/scratch/twutz/miniconda3/lib/python3.8/site-packages/datalad/config.py", line 497, in _reload
    stdout, stderr = self._run(
  File "/User/homes/twutz/scratch/twutz/miniconda3/lib/python3.8/site-packages/datalad/config.py", line 878, in _run
    out = self._runner.run(self._config_cmd + args, **kwargs)
  File "/User/homes/twutz/scratch/twutz/miniconda3/lib/python3.8/site-packages/datalad/runner/runner.py", line 201, in run
    raise CommandError(
datalad.runner.exception.CommandError: CommandError: 'git --git-dir= config -z -l --show-origin' failed with exitcode 128 [err: 'fatal: The empty string is not a valid path']

```

- include a git directive such as above
- run `datalad --version`

### Datalad information

-     datalad version 0.16.5
-     linuxmint/20.2/uma, release 5.4.0-121-generic
-     git annex version: 10.20220525-gf1fdc90
-     git version 2.34.1

### Additional context

My usual .gitconfig is set so that I use a different name, email and certificates in work-related projects (in specific folders) versus private projects.

First reported in  [neurostars question](https://neurostars.org/t/beginners-problem-with-datalad/22995)

I was quite disappointed by this unlucky first encounter with datalad. However, by just reading along the 101 tutorial I am impressed by your work. Thanks.

yarikoptic:
As it was reported in https://github.com/datalad/datalad/issues/6815
and inquired from git folks for a better solution (no reply yet) at
https://lore.kernel.org/git/YscCKuoDPBbs4iPX@lena.dartmouth.edu/T/#u
whenever we provide an empty value for --git-dir, iff there is includeIf
statement in config, then git starts to demand a non-empty value for that
setting.  Unfortunately RTFM does not give any clear answer either that is
a legit or not. So I have decided to replace one questionable use with
another which does not cause an error.

Anther alternative I found we could have used two separate invocations of
--system and --global, but we cannot use them at the same time.

I would be intrigued to see how it works or not on Windows ;) (didn't run a full sweep of tests locally either)

Closes #6815

edit: only later mentioned that sent it from a `maint` branch, oops

### DIFF
--- a/datalad/config.py
+++ b/datalad/config.py
@@ -373,7 +373,7 @@ class ConfigManager(object):
             # The caller didn't specify a repository. Unset the git directory
             # when calling 'git config' to prevent a repository in the current
             # working directory from leaking configuration into the output.
-            self._config_cmd = ['git', '--git-dir=', 'config']
+            self._config_cmd = ['git', '--git-dir=/dev/null', 'config']
 
         self._src_mode = source
         run_kwargs = dict()
--- a/datalad/tests/test_config.py
+++ b/datalad/tests/test_config.py
@@ -289,6 +289,24 @@ def test_something(path=None, new_home=None):
 
 
 @with_tree(tree={
+    '.gitconfig': """\
+[includeIf "gitdir:**/devbgc/**"]
+    path = ~/.gitconfig_bgc
+
+[custom "datalad"]
+  variable = value
+"""})
+def test_includeif_breaking(new_home=None):
+    patched_env = os.environ.copy()
+    patched_env.pop('GIT_CONFIG_GLOBAL', None)
+    patched_env.update(get_home_envvars(new_home))
+    with patch.dict('os.environ', patched_env, clear=True):
+        cfg = ConfigManager()
+        # just want to make sure we read it and didn't crash
+        assert cfg.get('custom.datalad.variable') == "value"
+
+
+@with_tree(tree={
     'ds': {
         '.datalad': {
             'config': """\


yarikoptic:
Besides other untrusted oddities, in windows, added test (only) failed with

```
CommandError: '"git" "--git-dir=/dev/null" "config" "-z" "-l" "--show-origin"' failed with exitcode 128 [err: 'fatal: Invalid path '/dev': No such file or directory']
```
so the workaround is incomplete unfortunately

yarikoptic:
pushed this work as `bf-config-no-repo` (instead of errorneous `maint`), might work on it later again although so far this approach is incomplete

PR #6822, opened 2022-07-08
RM(TST): remove testing of datalad.test which was removed from 0.17.0

yarikoptic:
This test was always skipped since conditioned on having numpy installed
and we no longer keep it as a dependency of any kind so on all CIs it was
skipped.  But I ran into it while testing for debian build where for
some reason numpy was installed

Actually detected while looking at fails for conda: https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=533304&view=logs&j=517fe804-fa30-5dc2-1413-330699242c05&t=c10fa5f2-fdf6-5338-3bdb-c4bea7c23412


yarikoptic:
Fail is test_ssh_copy

PR #6812, opened 2022-07-07
WIP: Quote scp filenames depending on ssh version

christian-monch:
Fixes #6806 

This PR adds a very rough ssh-version detection to deal with different filename quoting requirements in ssh version <= 8 and in ssh version 9.

This is a rough first draft to verify the overall viability and to test the waters in CI. The version parsing will have to be improved.

#### 🐛 Bug Fixes
- quote filenames used in `scp` depending on the ssh version. Fixes #6806

PR #6826, opened 2022-07-09
Do not quote paths for ssh >= 9

yarikoptic:
This is the work by @christian-monch in #6812 with 2 little final tune ups to get it working and robustify against case where version is None for some reason (avoid quoting). And also submitted against `maint`.

I would appreciate quite review/merge since that test kept failing all our CI runs, bringing fatigue.


yarikoptic:
Oh you are so green, 💚 it

christian-monch:
@yarikoptic Thanks for fixing the last bugs. Niche surprise! Just sat down to fix and: voila, it's already done. Thx

christian-monch (approved):
LGTM. Thx for fixing the name-comparison bug

PR #6827, opened 2022-07-09
BF: use --python-match minor option in new datalad-installer release to match outside version of Python

yarikoptic:
Yet another replacement now of #6803 and sits on top of #6826 -- I am aiming for a clean CI without needing to lookup what failed.  #6826 should be merged first


yarikoptic:
Any brave soul to hit merge for two birds at once?

yarikoptic:
#6826 is merged, so this is no brainer

PR #6833, opened 2022-07-12
DOC: Add git-annex bug report tips to contributing

adswa:
As discussed in the devcall


yarikoptic:
Thank you , looks good to me!

PR #6836, opened 2022-07-12
Support older pytests

### LINKED ISSUE #6835: Testing on debian stable (10.0) errors due to older pytest having no MonkeyPatch
```
Preparing to unpack .../136-python3-pytest_6.0.2-2_all.deb ...
Unpacking python3-pytest (6.0.2-2) ..
...
```
and that feature was added in 6.2.0 according to https://docs.pytest.org/en/stable/changelog.html#pytest-6-2-0-2020-12-12 . Not sure if backporting of http://deb.debian.org/debian/pool/main/p/pytest/pytest_6.2.5-3.dsc from testing would be seamless

we have a single use

```
$> git grep MonkeyPa
datalad/conftest.py:    with pytest.MonkeyPatch().context() as m:

```

@jwodder  may be it could be just worked around so we just don't use that feature and testing just works with pytest 6.0?

jwodder:
Fixes #6835.

#### 🛡 Tests
- Support older versions of pytest

### DIFF
--- a/datalad/conftest.py
+++ b/datalad/conftest.py
@@ -1,6 +1,8 @@
 import logging
 import os
 import re
+from contextlib import ExitStack
+from unittest.mock import patch
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


yarikoptic:
Thank you!  I will test it within debian stable package build env and will report back.

yarikoptic:
FWIW, patching with the diff from this PR allowed debian package build -- will now do for a proper -2 debian pkg build. not sure why one CI run here https://github.com/datalad/datalad/runs/7304081650?check_suite_focus=true seems got stuck but should not relate IMHO.

yarikoptic (approved):
CI is finally done and green.  I will let others to provide feedback, and if no feedback - will merge tomorrow

yarikoptic:
worked out well, no objections, let's proceed

Issue #4921, opened 2020-09-14
0.13.3 on conda and fresh git-annex: duplicated output picked up within test_runner_parametrized_protocol?


yarikoptic:
@notestaff working on adding downstream testing back to git-annex package on conda-forge which is great!

datalad testing is not yet green.
[Full log](https://dev.azure.com/conda-forge/84710dde-1620-425b-80d0-4cf5baca359d/_apis/build/builds/209014/logs/13) (160MB huge  due to annexremote testing verbosity, unrelated), but then testing of datalad fails with a single (better than if many) test failing

```
2020-09-14T06:57:15.1594851Z Versions: annexremote=1.4.3 appdirs=1.4.3 boto=2.49.0 cmd:7z=16.02 cmd:annex=8.20200908-ga853039 cmd:bundled-git=2.28.0 cmd:git=2.28.0 cmd:system-git=2.28.0 cmd:system-ssh=8.3p1 exifread=2.1.2 humanize=2.6.0 iso8601=0.1.13 keyring=21.4.0 keyrings.alt=3.1 msgpack=1.0.0 mutagen=1.41.1 patoolib=1.12 requests=2.24.0 tqdm=4.49.0 wrapt=1.12.1
2020-09-14T06:57:15.1601769Z Obscure filename: str=b' "\';a&b&c\xce\x94\xd0\x99\xd7\xa7\xd9\x85\xe0\xb9\x97\xe3\x81\x82 `| ' repr=' "\';a&b&cΔЙקﻡ๗あ `| '
2020-09-14T06:57:15.1608345Z Encodings: default='utf-8' filesystem='utf-8' locale.prefered='UTF-8'
2020-09-14T06:57:15.1616112Z Environment: PYTHONNOUSERSITE='1' PYTHONUNBUFFERED='1' PATH='$PREFIX:$PREFIX/bin:$PREFIX/bin:/opt/conda/condabin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/home/conda/bin' LANG='en_US.UTF-8' GIT_BRANCH='merge' PYTHON='$PREFIX/bin/python' GIT_CONFIG_PARAMETERS="'init.defaultBranch=dl-test-branch'"
2020-09-14T06:57:15.2906560Z Exception ignored in: <function _Database.close at 0x7fa3c4c100d0>
2020-09-14T06:57:15.2907546Z Traceback (most recent call last):
[...]

PR #6837, opened 2022-07-12
XFAIL test_runner_parametrized_protocol on python3.8 when getting duplicate output

yarikoptic:
This should close (unresolved) but not fix https://github.com/datalad/datalad/issues/4921
which seems to happen so far rarely on conda on 3.8 python only. If some brave soul
manages to figure it out -- great. But I would not break our knuckles to see it fixed
since seems to be very particular.  I made xfail also very particular (not over entire
test, only for 3.8 and only for the 55 output) so we do not miss any other problem

### DIFF
--- a/datalad/runner/tests/test_witless_runner.py
+++ b/datalad/runner/tests/test_witless_runner.py
@@ -199,7 +199,13 @@ def test_runner_parametrized_protocol():
         # value passed to protocol constructor
         value=b'5',
     )
-    eq_(res['stdout'], '5')
+    try:
+        eq_(res['stdout'], '5')
+    except AssertionError:
+        # TODO: remove when dropping support for python3.8
+        if res['stdout'] == '55' and sys.version_info[:2] == (3, 8):
+            pytest.xfail("got 55 and not 5, see https://github.com/datalad/datalad/issues/4921")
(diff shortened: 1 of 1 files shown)


yarikoptic:
any objections/blessings @datalad/developers ?

yarikoptic:
ok, no objections then.

datalad/runner/tests/test_witless_runner.py
```diff
@@ -199,7 +199,13 @@ def pipe_data_received(self, fd, data):
         # value passed to protocol constructor
         value=b'5',
     )
-    eq_(res['stdout'], '5')
+    try:
+        eq_(res['stdout'], '5')
+    except AssertionError:
+        # TODO: remove when dropping support for python3.8
+        if res['stdout'] == '55' and sys.version_info[:2] == (3, 8):
```
yarikoptic, line 206: apparently not 3.8 specific -- happened for 3.9 https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=556819&view=logs&j=517fe804-fa30-5dc2-1413-330699242c05&t=c10fa5f2-fdf6-5338-3bdb-c4bea7c23412

PR #6839, opened 2022-07-14
RF: Replace distutils LooseVersion with looseversion

### LINKED ISSUE #6307: Stop using `distutils.version.LooseVersion`

effigies:
Fixes #6307

Followed the convention of stdlib -> dependency -> currentpackage imports, but since there weren't dependency imports in these two files, I'm not positive that you follow this convention. LMK if I should change anything.

#### 🏠 Internal
- Use `looseversion.LooseVersion` as drop-in replacement for `distutils.version.LooseVersion`

### DIFF
--- a/datalad/distribution/create_sibling.py
+++ b/datalad/distribution/create_sibling.py
@@ -11,7 +11,6 @@
 
 __docformat__ = 'restructuredtext'
 
-from distutils.version import LooseVersion
 import logging
 import os
 from os.path import (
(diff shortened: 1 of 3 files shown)


yarikoptic:
There is no rush but also no objections, so let's just proceed!

yarikoptic-gitmate:

PR #6869, opened 2022-07-22
Let push honor multiple publication dependencies declared via siblings 

### LINKED ISSUE #6867: when multiple publish-depends remotes, only one is used
### What is the problem?

`datalad push --to origin` with origin having multiple 'remote.origin.datalad-publish-depends' only takes into consideration one of the sibling it should push to when I try to push

The push is done correctly on one of the 'publish-depends' remote and on the origin remote but nothing is pushed to the other 'publish-depends' remote.

After looking up what could have gone wrong :
https://github.com/datalad/datalad/blob/4f7735f7937c36da37b1c5fa2b8fc24b599cca5f/datalad/core/distributed/push.py#L499-L501
To me, it looks like the `ds.config.get(depvar, [])` only returns one value.

So I think replacing the line with:
```python
publish_depends = ensure_list(ds.config.get(depvar, [], get_all=True))
```
would solve the problem

```bash
#create a new dataset
datalad create dataset
cd dataset

# link to origin , public and confidential remotes (I used only GIN repositories)
datalad siblings add -s origin --url git@gin.g-node.org:/Loann_Peurey/test.git
datalad siblings add -s public --url git@gin.g-node.org:/Loann_Peurey/test-public.git
datalad siblings add -s confidential --url git@gin.g-node.org:/Loann_Peurey/test-confidential.git

# set publish-depends for origin to push to public and confidential
datalad siblings configure -s origin --publish-depends public --publish-depends confidential

# push to origin
datalad push --to origin
# => this only pushed to origin and public in my case.
```

## datalad
  - datalad 0.15.6
## git annex
  - git-annex version: 10.20220222-g8b42e3e
## python 
  - implementation: CPython
  - version: 3.7.12
## system 
  - distribution: debian/bullseye/sid
  - encoding: 
    - default: utf-8
    - filesystem: utf-8
    - locale.prefered: UTF-8
  - max_path_length: 288
  - name: Linux
  - release: 5.15.0-41-generic
  - type: posix
  - version: #44~20.04.1-Ubuntu SMP Fri Jun 24 13:27:29 UTC 2022

### Additional context

content of `git/config`
```
[core]
        repositoryformatversion = 0
        filemode = true
        bare = false
        logallrefupdates = true
[annex]
        uuid = 5efda5a7-c095-4ada-8b39-0230e22b126f
        version = 8
[filter "annex"]
        smudge = git-annex smudge -- %f
        clean = git-annex smudge --clean -- %f
[remote "origin"]
        url = git@gin.g-node.org:/Loann_Peurey/test.git
        fetch = +refs/heads/*:refs/remotes/origin/*
        annex-ignore = false
        annex-uuid = 30cf6df3-8145-431b-aeb5-a157ffb32237
        datalad-publish-depends = confidential
        datalad-publish-depends = public
[branch "main"]
        remote = origin
        merge = refs/heads/main
[remote "confidential"]
        url = git@gin.g-node.org:/Loann_Peurey/test-confidential.git
        fetch = +refs/heads/*:refs/remotes/confidential/*
        annex-ignore = false
        annex-uuid = ab8b01c8-2de8-4551-b31b-419fa56b446d
[remote "public"]
        url = git@gin.g-node.org:/Loann_Peurey/test-public.git
        fetch = +refs/heads/*:refs/remotes/public/*
        annex-ignore = false
        annex-uuid = 5e1760fc-0023-425e-b276-8f738820e872
```
output of `git config`
```bash
~/dataset$ git config --get-all remote.origin.datalad-publish-depends
public
confidential
```

Have been using datalad for 6 months, managing datasets for a research lab.

mih:
The documentation for declaring publication dependencies allows for two different ways to declare multiple dependencies:

1. Declaring a single config item with a list of dependencies (space-separated)
2. Declaring a config item multiple times with a single dependency each.

Previously `push()` only supported (1), while `siblings()` produced (2).

This change lets `push()` handle both scenarios.

Fixes #6867

Thanks to @LoannPeurey for the report!

### DIFF
--- a/datalad/core/distributed/push.py
+++ b/datalad/core/distributed/push.py
@@ -13,6 +13,7 @@
 __docformat__ = 'restructuredtext'
 
 from collections import OrderedDict
+from itertools import chain
 import logging
 import re
 
@@ -498,7 +499,13 @@ def _push(dspath, content, target, data, force, jobs, res_kwargs, pbars,
     depvar = 'remote.{}.datalad-publish-depends'.format(target)
     # list of remotes that are publication dependencies for the
     # target remote
-    publish_depends = ensure_list(ds.config.get(depvar, []))
+    # multiple dependencies could come from multiple declarations
+    # of such a config items, but each declaration could also
+    # contain a white-space separated list of remote names
+    # see https://github.com/datalad/datalad/issues/6867
+    publish_depends = list(chain.from_iterable(
+        d.split() for d in ensure_list(
+            ds.config.get(depvar, [], get_all=True))))
     if publish_depends:
         lgr.debug("Discovered publication dependencies for '%s': %s'",
                   target, publish_depends)
--- a/datalad/core/distributed/push.py
+++ b/datalad/core/distributed/push.py
@@ -499,13 +499,7 @@ def _push(dspath, content, target, data, force, jobs, res_kwargs, pbars,
     depvar = 'remote.{}.datalad-publish-depends'.format(target)
     # list of remotes that are publication dependencies for the
     # target remote
-    # multiple dependencies could come from multiple declarations
-    # of such a config items, but each declaration could also
-    # contain a white-space separated list of remote names
-    # see https://github.com/datalad/datalad/issues/6867
-    publish_depends = list(chain.from_iterable(
-        d.split() for d in ensure_list(
-            ds.config.get(depvar, [], get_all=True))))
+    publish_depends = ensure_list(ds.config.get(depvar, [], get_all=True))
     if publish_depends:
         lgr.debug("Discovered publication dependencies for '%s': %s'",
                   target, publish_depends)
--- a/datalad/core/distributed/tests/test_push.py
+++ b/datalad/core/distributed/tests/test_push.py
@@ -62,6 +62,10 @@ from datalad.utils import (
 
 DEFAULT_REFSPEC = "refs/heads/{0}:refs/heads/{0}".format(DEFAULT_BRANCH)
 
+ckwa = dict(
+    result_renderer='disabled',
+)
+
 
 @with_tempfile(mkdir=True)
 @with_tempfile(mkdir=True)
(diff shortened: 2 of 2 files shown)


datalad/core/distributed/push.py
```diff
@@ -498,7 +499,13 @@ def _push(dspath, content, target, data, force, jobs, res_kwargs, pbars,
     depvar = 'remote.{}.datalad-publish-depends'.format(target)
     # list of remotes that are publication dependencies for the
     # target remote
-    publish_depends = ensure_list(ds.config.get(depvar, []))
+    # multiple dependencies could come from multiple declarations
+    # of such a config items, but each declaration could also
+    # contain a white-space separated list of remote names
+    # see https://github.com/datalad/datalad/issues/6867
```
yarikoptic, line 505: where do we document/allow for a white-space separated list of names?

Rationale for asking:  Indeed ATM git does not allow white-space in the name of the remote, and it seems we do not "disallowed" it explicitly since just allow for `git` to error out:

```shell
$> datalad siblings add -s "my name" --url https://example.com
CommandError: 'git -c diff.ignoreSubmodules=none remote add 'my name' https://example.com' failed with exitcode 128
fatal: 'my name' is not a valid remote name
```

so if in the future `git` for some reason decides to support that (unlikely, but not impossible IMHO) such splitting by space might become buggy.  So I wondered to ask where we already document/allow for that, and may be we should add some explicit checking for the name to not contain spaces (since could lead to fail of this added logic)

mih: Hmm, I cannot find a user-facing documentation that made me conclude that we allow(ed) for this. Maybe it was a comment, but I cannot find that either.

yarikoptic: then since we didn't have that documented, and looking at original code above and behavior of ensure_list:

```shell
$> python -c 'from datalad.utils import ensure_list; print(ensure_list("w1 w2"))'
['w1 w2']
```
that it doesn't split by space, I think we never had such a feature -- should we really add it along with this?  may be just a matter of adding `get_all=True` and be done?

yarikoptic:
2 travis fails known, one is something to note -- might need `@slow`

```
___________________ test_nested_pushclone_cycle_allplatforms ___________________
Test passed but took too long to run: Duration 30.47114537500005s > 30.0s
----------------------------- Captured stdout call -----------------------------
```

yarikoptic:
with all of the above in mind, let's proceed !

Issue #6862, opened 2022-07-20
Confusing INFO-message with `datalad get -r`


christian-monch:
### What is the problem?

When `datalad get -r .` is called inside an installed subdataset, the following information is printed to console:

```
2022-07-20 05:00:39,876 [INFO   ] Installing Dataset(/tmp/super1/sub1) to get /tmp/super1/sub1 recursively
```

This is confusing, since the subdataset in question, here: `sub1`, is already present and does not require installation.

### What steps will reproduce the problem?

The following is a transcript of a reproduction:

```shell
> cd /tmp
> datalad create super1
create(ok): /tmp/super1 (dataset)
> cd super1/
> datalad create -d . sub1
add(ok): sub1 (file)                                                                                                                                                                         
add(ok): .gitmodules (file)                                                                                                                                                                  
save(ok): . (dataset)                                                                                                                                                                        
create(ok): sub1 (dataset)                                                                                                                                                                   
action summary:
  add (ok: 2)
  create (ok: 1)
  save (ok: 1)
> echo "something" >sub1/some.txt
> datalad save -r
add(ok): some.txt (file)                                                                                                                                                                     
save(ok): sub1 (dataset)                                                                                                                                                                     
add(ok): sub1 (file)                                                                                                                                                                         
save(ok): . (dataset)                                                                                                                                                                        
action summary:                                                                                                                                                                              
  add (ok: 2)
  save (ok: 2)
> cd sub1/
> datalad get -d . -r .
2022-07-20 05:00:39,876 [INFO   ] Installing Dataset(/tmp/super1/sub1) to get /tmp/super1/sub1 recursively 
``` 

### Datalad information

```
> datalad wtf
# WTF
## configuration <SENSITIVE, report disabled by configuration>
## credentials 
  - keyring: 
    - active_backends: 
      - SecretService Keyring
      - PlaintextKeyring with no encyption v.1.0 at /home/cristian/.local/share/python_keyring/keyring_pass.cfg
    - config_file: /home/cristian/.config/python_keyring/keyringrc.cfg
    - data_root: /home/cristian/.local/share/python_keyring
## datalad 
  - version: 0.17.2
## dataset 
  - branches: 
    - git-annex@167a5d7
    - master@c5d4daa
  - id: 0091f26b-e480-4b59-86db-e614a466a8bc
  - metadata: <SENSITIVE, report disabled by configuration>
  - path: /tmp/super1/sub1
  - repo: AnnexRepo
## dependencies 
[...]

PR #6871, opened 2022-07-22
Clarify confusing INFO log message from get() on dataset installation

mih:
The confusion arises from a message that announces the installation
of a (sub)dataset, even when the dataset is already present.

This change rewords the message to be more valid in this case.

Fixes datalad/datalad#6862

### DIFF
--- a/datalad/distribution/get.py
+++ b/datalad/distribution/get.py
@@ -622,7 +622,7 @@ def _install_targetpath(
             yield res
         return
     lgr.info(
-        "Installing %s%s recursively",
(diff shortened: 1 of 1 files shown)


bpoldrack (approved):
Issue remains a bit weird in the toplevel case, since this was ensured already and `get` would fail before otherwise, but I agree - this wording removes the confusion about the message itself.

mih:
Thx for the review. I think all these log messages are a relic, and should be implemented properly as progress reporting. But I have no interest in working on this.

PR #6878, opened 2022-07-24
Report `prev_type` rather than `type` for `diff/status` of deletions

### LINKED ISSUE #6877: status/diff report (wrong) type for deletions
In the same spirit as #6875:

```
% git diff-files -- README.md
:100644 000000 0696c43208dbcc378ed4c0531899eadca6165ca3 0000000000000000000000000000000000000000 D      README.md
```

but (even with #6876)

```
% dl -f json_pp status -- README.md
{
  "action": "status",
  "parentds": "/home/mih/hacking/datalad/next",
  "path": "/home/mih/hacking/datalad/next/README.md",
  "prev_gitshasum": "0696c43208dbcc378ed4c0531899eadca6165ca3",
  "refds": "/home/mih/hacking/datalad/next",
  "state": "deleted",
  "status": "ok",
  "type": "file"
}
```

IMHO we should align the behavior to Git and report `prev_type` instead of `type` in case of a deletion.

The default result renderer can then be adjusted to display the content of `prev_type` when `state == 'deleted'` to maintain the same begavior, at least visually.

mih:
This PR sits on top of https://github.com/datalad/datalad/pull/6797 to enable a simple and clear fix. Only the last commit is relevant here.

This change aligns the behavior with that of `git diff-files`.
    
The default result renderer is adjusted to maintain the same (visual) behavior. This is also fixing the querying of an outdated result property.
    
Closes datalad/datalad#6877

PR against `master`, because of the implied behavior change.

#### 🐛 Bug Fixes
- Align status/diff reporting behavior with git-diff-files, by not reporting a `type` property for paths that are deleted. Instead `prev_type` is reported. Fixes #6877

### DIFF
--- a/datalad/consts.py
+++ b/datalad/consts.py
@@ -81,3 +81,14 @@ RESERVED_NAMES_WIN = {'CON', 'PRN', 'AUX', 'NUL', 'COM1', 'COM2', 'COM3',
                       'LPT9'}
 # Characters that can't be a part of a file name on Windows
 ILLEGAL_CHARS_WIN = "[<>:/\\|?*\"]|[\0-\31]"
+
+# mode identifiers used by Git (ls-files, ls-tree), mapped to
+# type identifiers as used in command results
+GIT_MODE_TYPE_MAP = {
+    '100644': 'file',
+    # we do not distinguish executables
+    '100755': 'file',
+    '040000': 'directory',
+    '120000': 'symlink',
+    '160000': 'dataset',
+}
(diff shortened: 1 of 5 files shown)


mih:
I am thinking of closing https://github.com/datalad/datalad/pull/6797#issuecomment-1214997911

adswa:
The premise of the last commit makes sense to me (but I did not look at the previous commits)

mih:
Closing, after https://github.com/datalad/datalad/pull/6797 was closed.

Issue #6874, opened 2022-07-24
Extension API generator cannot handle extension load failure


mih:
Related to, but different from #6844.

The critical code in `datalad.api` is

```py
       intf = load_interface(intfspec[:2])
        api_name = get_api_name(intfspec)
        if api_name in globals():
            lgr.debug(
                'Command %s from extension %s is replacing a previously loaded implementation',
                api_name,
                ename)
       globals()[api_name] = intf.__call__
```

The docs of `load_interface()` say

```rst
    Returns
    -------
    The interface class or, if importing the module fails, None
```

So it is very much expected that `intf` could be `None`, which will always make `intf.__call__` access crash the session.

PR #6879, opened 2022-07-24
Protect again failing to load a command interface from an extension

mih:
Previously DataLad could crash if an extension entrypoint was
importable, but a declared command suit item would not.

This change now skips each individual command interface with an
error message like this:

```
% python -c 'import datalad.api as dl'
[ERROR  ] Internal error, cannot import interface 'datalad_dataverse.create_sibling_dataverse': ModuleNotFoundError(No module named 'datalad_next')
[ERROR  ] Skipping unusable command interface 'datalad_dataverse.create_sibling_dataverse.CreateSiblingDataverse' from extension 'dataverse'
```

(I have an extension installed, where one command depends on an
unavailable package).

Closes datalad/datalad#6874

### DIFF
--- a/datalad/api.py
+++ b/datalad/api.py
@@ -57,6 +57,11 @@ def _generate_extension_api():
         for intfspec in interfaces:
             # turn the interface spec into an instance
             intf = load_interface(intfspec[:2])
+            if intf is None:
+                lgr.error(
+                    "Skipping unusable command interface '%s.%s' from extension %r",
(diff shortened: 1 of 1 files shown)


yarikoptic:
docs fail unrelated: https://github.com/datalad/datalad/issues/6882

yarikoptic:
Should be ok. Some brave soul might come up with a test some time in the bright future to ensure we do not error while handling an error ;)

PR #6880, opened 2022-07-24
RF,ENH(TST): future proof testing of git annex version upgrade + test annex init on all supported versions

yarikoptic:
Originally this test was crafted when 6 was still among supported versions.
We have gone past that point now, minimal supported version of git-annex
(from 202003) supports only 8, and current ones default to 8 but support also
9 and 10.  I have RF and extended test to to remain valid "forever" and test
that we can init git-annex of any supported version

### DIFF
--- a/datalad/support/tests/test_annexrepo.py
+++ b/datalad/support/tests/test_annexrepo.py
@@ -135,6 +135,9 @@ from datalad.utils import (
 )
 
 
+_GIT_ANNEX_VERSIONS_INFO = AnnexRepo.check_repository_versions()
+
+
 @assert_cwd_unchanged
 @with_tempfile
 @with_tempfile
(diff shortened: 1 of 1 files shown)


yarikoptic:
docs known, so likely unrelated. travis - one known and one annoying recent filed https://github.com/datalad/datalad/issues/6887 . These two tests passed just fine everywhere

```
$> git grep test_repo_version | nl | tail -n 2
    43	ca7460f/travis-14312-success/8.txt:../datalad/support/tests/test_annexrepo.py::test_repo_version_upgrade PASSED [ 59%]
    44	ca7460f/travis-14312-success/8.txt:../datalad/support/tests/test_annexrepo.py::test_repo_version_supported[8] PASSED [ 59%]

$> git grep test_repo_version | grep -v PASSED
```

PR #6886, opened 2022-07-24
BF: RF f-string uses in logger to %-interpolations

yarikoptic:
f-strings should not be used for logger since if any of the .format-ed
content contains some %, it would then be incorrectly or causing a crash
%-polated by the logger.  Although some of the "fixed" cases would be ok
(we know that it was an int etc) for consistency and to avoid needed
"is this particular use safe?" checks, better to avoid f-strings in
logger in general

### DIFF
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -119,7 +119,7 @@ class BatchedCommandProtocol(GeneratorMixIn, StdOutErrCapture):
         if fd == STDOUT_FILENO:
             remaining_line = self.line_splitter.finish_processing()
             if remaining_line is not None:
-                lgr.debug(f"unterminated line: {remaining_line}")
+                lgr.debug("unterminated line: %s", remaining_line)
                 self.send_result((fd, remaining_line))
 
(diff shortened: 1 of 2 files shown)


mih (approved):
I agree. The performance aspect is also not to be forgotten.

datalad/cmd.py
```diff
@@ -325,13 +325,13 @@ def __call__(self,
...
                         # code and restart the process.
-                        lgr.debug(f"{self}: command exited")
+                        lgr.debug("%s: command exited", self)
                         self.return_code = self.generator.return_code
                         self.runner = None
 
         except CommandError as command_error:
             # The command exited with a non-zero return code
-            lgr.error(f"{self}: command error: {command_error}")
+            lgr.error("%s: command error: command_error", self, command_error)
```
yarikoptic, line 334: dang, I did I screw up here, travis picked it up in https://app.travis-ci.com/github/datalad/datalad/jobs/577595841#L3864

Issue #6887, opened 2022-07-24
Travis: installs "wrong" version of datalad, gets 0+unknown version as a result


yarikoptic:
previously observed in https://github.com/datalad/datalad/pull/6842 and others,  now in https://github.com/datalad/datalad/pull/6880 with detected oddity:  seems known flukes with version (git describe does say [0.17.2-17-ga415c2946](https://app.travis-ci.com/github/datalad/datalad/jobs/577592702) in one where it is probably ok but fails due to known other unrelated; and then [0.17.2-13-g637eb3d01 in another run](https://app.travis-ci.com/github/datalad/datalad/jobs/577592696) where it becomes unknown.  That 637eb3d01 is nothing known locally .

Seems to be always happening in that matrix run for  python 3.10 where installation is a bit different from other runs:

```shell
(git)smaug:/mnt/datasets/datalad/ci/logs/2022/07[master]git
$> git grep -l '0+unknown' | grep -v Extensions
02/pr/6789/fb05b8b/travis-14194-failed/3.txt
11/push/maint/558c0c3/travis-14262-failed/3.txt
11/push/maint/9679312/travis-14264-failed/3.txt
15/pr/6841/ed686d2/travis-14273-failed/3.txt
15/pr/6842/d449c68/travis-14275-failed/3.txt
15/push/maint/8c65776/travis-14274-failed/3.txt
15/push/master/8c65776/travis-14277-failed/3.txt
16/push/maint/8787719/travis-14280-failed/3.txt
16/push/maint/e377e54/github-Auto-release on PR merge-166-success/1_auto-release.txt
16/push/maint/e377e54/github-Auto-release on PR merge-166-success/auto-release/7_Create release.txt
17/pr/6682/3e2b96c/travis-14282-errored/3.txt
18/push/master/e377e54/travis-14283-failed/3.txt
19/cron/20220719T215137/c29c2e1/travis-14291-failed/3.txt
19/pr/6784/5bec143/travis-14285-failed/3.txt
19/pr/6797/79c69c7/travis-14290-failed/3.txt
19/pr/6859/399688c/travis-14289-failed/3.txt
20/pr/6858/4e697f5/travis-14294-failed/3.txt
21/push/maint/cefecae/travis-14301-failed/3.txt
22/cron/20220722T194034/1c8944c/travis-14305-failed/3.txt
24/pr/6880/ca7460f/travis-14312-failed/3.txt
24/pr/6881/5ae1b46/travis-14313-failed/4.txt
[...]

PR #6888, opened 2022-07-26
ENH: upgrade versioneer from versioneer-0.20.dev0 to versioneer-0.23.dev0

yarikoptic:
In wild hopes of resolving https://github.com/datalad/datalad/issues/6887


yarikoptic:
ok -- good that we attended to benchmarks. All good now. failing travis is known NFS one

PR #6890, opened 2022-07-26
DX: Demote loglevel of message on url parameters to debug

### LINKED ISSUE #6872: `RI` unconditionally whines about obscure path
The tests use obscure path names for good reasons. Given that, I do not understand this behavior:

```py
>>> from datalad.support.network import RI
>>> RI("/tmp/datalad-fs-DAkJS/datalad_temp_tree_test_create_subdataset_hierarchy_from_topkfwtympw/origin/ds-;&%b5{}'ΔЙקم๗あ.datc")
[WARNING] ParseResults contains params "&%b5{}'ΔЙקم๗あ.datc", which will be ignored 
PathRI(path="/tmp/datalad-fs-DAkJS/datalad_temp_tree_test_create_subdataset_hierarchy_from_topkfwtympw/origin/ds-;&%b5{}'ΔЙקم๗あ.datc")
```

The path shown above is generated by our own test helpers. `RI` seems to be cabable of handling it correctly. But in the process `URL._pr_to_fields()` is whining loudly, and inappropriately in this context.

This is both a UX issue (users have to endure this with any repository that has `&` in the directory name), and a developer X issue (I spent a considerable amount of timing hunting this message down, assuming that it is related to a bug I am investigation (given the WARNING level), but turns out to be inconsequential).

adswa:
Our own obscure paths generate filepaths that RFC 1808 on URLs
schemes mistakes to have URL parameter specifications because
they contain a semicolon (see  #6872).
Internally, when urlparse detects parameters in a URL, we warn
that those parameters will not be honored or considered at all.
In general, I find this is a nice thing to do, but annoying when
its actually not a URL with parameters. My initial thought on
how to keep the information but make it less annoying is to
demote the log level of the message from warning to debug.
Alternatively, we could do more parsing and guessing on whether
or not something a URL where parameters make sense, but that
could probably easily introduce more problems or wrong assessments. So this change here could maybe fix #6872 

#### 🏠 Internal
- In order to reduce inconsequential warnings over obscure paths resembling URLs with parameters in tests, a warning about ignoring URL parameters has been demoted to debug log level

### DIFF
--- a/datalad/support/network.py
+++ b/datalad/support/network.py
@@ -631,7 +631,7 @@ class URL(RI):
         This function converts ParseResult into dict"""
 
         if pr.params:
-            lgr.warning("ParseResults contains params %r, which will be ignored"
+            lgr.debug("ParseResults contains params %r, which will be ignored"
                         % (pr.params,))
 
         hostname_port = pr.netloc.split('@')[-1]
--- a/datalad/support/tests/test_network.py
+++ b/datalad/support/tests/test_network.py
@@ -211,7 +211,7 @@ def test_url_base():
 
     assert_raises(ValueError, url._set_from_fields, unknown='1')
 
-    with swallow_logs(new_level=logging.WARNING) as cml:
+    with swallow_logs(new_level=logging.DEBUG) as cml:
         # we don't "care" about params ATM so there is a warning if there are any
         purl = URL("http://example.com/;param")
         eq_(str(purl), 'http://example.com/;param')  # but we do maintain original string


yarikoptic:
I think it is worth `maint` after a little "tune up". I think we should still issue warning whenever we explicitly parse for `URL`.  Hence I have resubmitted your PR with extra commit for that against `maint` in https://github.com/datalad/datalad/pull/6891 thus superseding this PR if accepted.

adswa:
> I think it is worth `maint` after a little "tune up". I think we should still issue warning whenever we explicitly parse for `URL`. Hence I have resubmitted your PR with extra commit for that against `maint` in #6891 thus superseding this PR if accepted.

nice, thanks! I'm closing this in favor of yours :+1:

PR #6892, opened 2022-07-27
Use pytest to test datalad-metalad in test_extensions-workflow

christian-monch:
This PR modifies the test_extensions-workflow to use `pytest` if the tested extension is `datalad-metalad`. The extension `datalad-metalad` recently switched from `nose` to `pytest` because `nose` does not work with python 3.10.

This should fix errors in `test (datalad-metalad)

#### 🛡 Tests
- switch test runner for `datalad-metalad` extension from `nose` to `pytest`.

### DIFF
--- a/.github/workflows/test_extensions.yml
+++ b/.github/workflows/test_extensions.yml
@@ -67,6 +67,14 @@ jobs:
         # TODO: later whenever some extensions would migrate to pytest -- use pytest
         # python -m pytest -c ../tox.ini -s -v --cov=datalad --pyargs $(echo ${{ matrix.extension }} | tr '-' '_')
         python -m coverage xml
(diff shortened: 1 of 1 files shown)


bpoldrack (commented):
Looks right overall. However, the `Install dependencies` step further up has an unconditional `pip install nose` ATM. I'd say make it conditional now on `matrix.extension != 'datalad-metalad'?

bpoldrack:
Travis failure unrelated, but: Ping @yarikoptic - `Versioneer` upgrade didn't do the trick apparently.

christian-monch:
@bpoldrack: affre

> Looks right overall. However, the `Install dependencies` step further up has an unconditional `pip install nose` ATM. I'd say make it conditional now on `matrix.extension != 'datalad-metalad'`?

@bpoldrack : addressed your comment

bpoldrack (approved):

PR #6893, opened 2022-07-27
Fix runner type annotations

### LINKED ISSUE #6889: `datalad/runner` already has lots of annotations but ATM fails `mypy` (@christian-monch interested? ;))

christian-monch:
This PR fixes #6889 

The runner-code passes validation tests of `mypy` version `0.971`, i.e. with strict optional checking.

#### 🏠 Internal
- fix type annotations in `datalad.runner` this resolves issue #6889

### DIFF
--- a/datalad/runner/exception.py
+++ b/datalad/runner/exception.py
@@ -82,7 +82,7 @@ class CommandError(RuntimeError):
         return self.to_str()
 
 
-def _format_json_error_messages(recs: List[Dict]):
+def _format_json_error_messages(recs: List[Dict]) -> str:
     # there could be many, condense
     msgs: Dict[str, Union[str, int]] = {}
     for r in recs:
--- a/datalad/runner/exception.py
+++ b/datalad/runner/exception.py
@@ -8,13 +8,9 @@
 # ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
 """Exception raise on a failed runner command execution
 """
+from __future__ import annotations
 
 import logging
-from typing import (
-    Dict,
-    List,
-    Union,
-)
 
 
 lgr = logging.getLogger('datalad.runner.exception')
@@ -82,9 +78,9 @@ class CommandError(RuntimeError):
         return self.to_str()
 
 
-def _format_json_error_messages(recs: List[Dict]) -> str:
+def _format_json_error_messages(recs: list[dict]) -> str:
     # there could be many, condense
-    msgs: Dict[str, Union[str, int]] = {}
+    msgs: dict[str, str | int] = {}
     for r in recs:
         if r.get('success'):
             continue
--- a/datalad/runner/exception.py
+++ b/datalad/runner/exception.py
@@ -13,6 +13,7 @@ import logging
 from typing import (
     Dict,
     List,
+    Union,
 )
 
 
@@ -83,7 +84,7 @@ class CommandError(RuntimeError):
 
 def _format_json_error_messages(recs: List[Dict]):
     # there could be many, condense
-    msgs = {}
+    msgs: Dict[str, Union[str, int]] = {}
     for r in recs:
         if r.get('success'):
             continue
@@ -93,6 +94,7 @@ def _format_json_error_messages(recs: List[Dict]):
         )
         if 'file' in r or 'key' in r:
             occur = msgs.get(msg, 0)
+            assert isinstance(occur, int)
             occur += 1
             msgs[msg] = occur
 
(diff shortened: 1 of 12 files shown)


jwodder:
@christian-monch DataLad requires Python 3.7+, correct?  If so, the recommended approach is to use "modern" type annotations by adding `from __future__ import annotations` to the top of annotated files, which lets you write, e.g., `dict[str, str | int]` instead of `typing.Dict[str, typing.Union[str, int]]`.

datalad/runner/exception.py
```diff
@@ -83,7 +84,7 @@ def __str__(self):
 
 def _format_json_error_messages(recs: List[Dict]):
```
jwodder, line 85:
```suggestion
def _format_json_error_messages(recs: List[Dict]) -> str:
```

jwodder (changes requested):

christian-monch:
> @christian-monch DataLad requires Python 3.7+, correct? If so, the recommended approach is to use "modern" type annotations by adding `from __future__ import annotations` to the top of annotated files, which lets you write, e.g., `dict[str, str | int]` instead of `typing.Dict[str, typing.Union[str, int]]`.

Good point, thanks. The annotations stem from the earlier versions. Will verify the requirements and use the "modern" style, which I prefer anyway :-)

yarikoptic:
woohoo, thanks @christian-monch . I pushed little commit to add entire runner into typechecking CI of tox

yarikoptic:
for your convenience, here is what errors CI reports
```
datalad/runner/tests/test_utils.py:140: error: Need type annotation for "decoded_chars"
datalad/runner/tests/test_threadsafety.py:61: error: Need type annotation for "args"
datalad/runner/tests/test_witless_runner.py:22: error: Cannot find implementation or library stub for module named "pytest"
datalad/runner/tests/test_witless_runner.py:22: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
Found 3 errors in 3 files (checked 22 source files)
ERROR: InvocationError for command /home/runner/work/datalad/datalad/.tox/typing/bin/mypy --follow-imports skip datalad/api.py datalad/downloaders/providers.py datalad/metadata/indexers/base.py datalad/runner datalad/support/collections.py (exited with code 1)
```
so only something funny left in the runner tests.

yarikoptic:
For
```
error: Cannot find implementation or library stub for module named "pytest"
```

added `pytest` to `deps` of typing (@jwodder is there a better way?) . So we are left only with the other two

tox.ini
```diff
@@ -29,6 +29,7 @@ commands = flake8 {posargs}
 [testenv:typing]
 deps =
     mypy~=0.900
+    pytest
```
jwodder, line 32: Rather than manually specifying test dependencies, it would be better to tell tox to install all of them via the `tests` extra:

```
[testenv:typing]
extras = tests
...
```

christian-monch:
> for your convenience, here is what errors CI reports
> 
> ```
> datalad/runner/tests/test_utils.py:140: error: Need type annotation for "decoded_chars"
> datalad/runner/tests/test_threadsafety.py:61: error: Need type annotation for "args"
> datalad/runner/tests/test_witless_runner.py:22: error: Cannot find implementation or library stub for module named "pytest"
> datalad/runner/tests/test_witless_runner.py:22: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
> Found 3 errors in 3 files (checked 22 source files)
> ERROR: InvocationError for command /home/runner/work/datalad/datalad/.tox/typing/bin/mypy --follow-imports skip datalad/api.py datalad/downloaders/providers.py datalad/metadata/indexers/base.py datalad/runner datalad/support/collections.py (exited with code 1)
> ```
> 
> so only something funny left in the runner tests.

Thx, forgot to check the `tests`-directory.

datalad/runner/tests/test_threadsafety.py
```diff
@@ -32,7 +33,7 @@ def pipe_data_received(self, fd, data):
             self.send_result((fd, line))
 
 
-def _runner_with_protocol(protocol) -> ThreadedRunner:
+def _runner_with_protocol(protocol: Type[WitlessProtocol]) -> ThreadedRunner:
```
jwodder, line 36: `from __future__ import annotations` lets you use the builtin `type` instead of `typing.Type` here.

jwodder (changes requested):

datalad/runner/tests/test_threadsafety.py
christian-monch: Thx. Changed it to `type[...]`

tox.ini
jwodder: Someone please address this.

jwodder (changes requested):

christian-monch:
Moved the type-test dependency definitions to `setup.py`

jwodder (approved):

PR #6895, opened 2022-07-27
Switch default access protocol for GH siblings from https to https-ssh

### LINKED ISSUE #6894: Reconsider default access protocol for create-sibling-github
By default a Github sibling is created with an HTTPS protocol. I keep forgetting this, and then get reminded of the deprecated user-password authentication of Github when I try and fail to push. Making ``https-ssh`` the default would make sense in my opinion. What do you think?

adswa:
``create-sibling-github`` by defaults configures the new github sibling by default with the https protocol. With deprecated user-password authentication, pushing contents via HTTPS is likely to fail for users. ``https-ssh`` would perform pushing via SSH and pull via HTTPS, and I think this is a more failsafe default.
Fixes #6894 

- ``create-sibling-github``'s access protocol default was changed from https to https-ssh


yarikoptic (approved):
> With deprecated user-password authentication

github still allows for a token to be used as a password, doesn't it?

In general "I approve" although still not 100% sure since if it does, then Windows users would have better chance to be able to push changes in likely absence of familiarity with `ssh`.

adswa:
> github still allows for a token to be used as a password, doesn't it?

Yes, it does, although I find it quite confusing when Git prompts for "password" ("token" would be much more unambiguous IMO). And good point about the Windows users. I'm also unsure now which way would be more convenient for most people now...

adswa:
given there is one approval and no one complained for three weeks, and there are two dozen open PRs, I'll merge this later today. Last chance to complain

yarikoptic:
> given there is one approval and no one complained for three weeks, and there are two dozen open PRs, I'll merge this later today. Last chance to complain

did you make up your mind about Windows ppl?  should it may be sense if there is `ssh` (not a strict requirement AFIAK) and then choose https-ssh?

PR #6898, opened 2022-07-29
ENH(TST): Disable coverage reporting on travis while running pytest

yarikoptic:
I found those reports spit out at the end of the pytest very annoying to avoid
while searching for failed tests.  And they aren't really useful since we

- at large anyways relying on codecov reporting while combining across
different CIs and runs

- have `report` at the end upon success when submitting to codecov. So we
can look at those if desired, and they are nicly folded in their own step

If desired -- can do for other runs (appveyor etc).


yarikoptic:
ok, no objections. CI fails known/unrelated, I like it, let's proceed.

PR #6899, opened 2022-07-29
BK: Initial set of type annotation fixes for cmd.py and some in runner.py

yarikoptic:
Decided to practice some type annotations.  Need help.  Will ask specific questions in resolvable threads on code.

### DIFF
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -12,6 +12,7 @@ via stdin. For each instruction send over stdin, a response is read and
 returned. The response structure is determined by "output_proc"
 
 """
+from __future__ import annotations
 
 import logging
 import os
@@ -29,7 +30,7 @@ from typing import (
 )
 from datetime import datetime
 from operator import attrgetter
-from weakref import WeakValueDictionary
+from weakref import WeakValueDictionary, ReferenceType, ref
 
 # start of legacy import block
 # to avoid breakage of code written before datalad.runner
@@ -174,7 +175,7 @@ class BatchedCommand(SafeDelCloseMixin):
 
     # Collection of active BatchedCommands as a mapping from object IDs to
     # instances
-    _active_instances = WeakValueDictionary()
+    _active_instances: WeakValueDictionary[int, BatchedCommand] = WeakValueDictionary()
 
     def __init__(self,
                  cmd: Union[str, Tuple, List],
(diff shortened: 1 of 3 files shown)


datalad/cmd.py
```diff
@@ -185,15 +186,15 @@ def __init__(self,
...
+        self.command: list = [command] if not isinstance(command, List) else command
+        self.path: Optional[str] = path
+        self.output_proc: Optional[Callable] = output_proc
+        self.timeout: Optional[float] = timeout
+        self.exception_on_timeout: bool = exception_on_timeout
 
         self.stdin_queue = None
         self.stderr_output = b""
-        self.runner = None
+        self.runner: Optional[WitlessRunner] = None
```
yarikoptic, line 197: this delayed lazy initialization becomes an issue. Since we allow for `None`, then type checking complains that we are possibly accessing some attributes from `None`:

```
$    mypy --follow-imports skip datalad/cmd.py datalad/runner/runner.py                          
datalad/cmd.py:281: error: "None" has no attribute "runner"
datalad/cmd.py:330: error: "None" has no attribute "return_code"
datalad/cmd.py:355: error: "None" has no attribute "put"
datalad/cmd.py:402: error: "None" has no attribute "send"
datalad/cmd.py:501: error: "None" has no attribute "runner"
Found 5 errors in 1 file (checked 2 source files)
```

how to solve that properly @jwodder @christian-monch ?

datalad/runner/runner.py
```diff
@@ -70,7 +72,7 @@ def run(self,
             env=None,
             timeout=None,
             exception_on_error=True,
-            **kwargs):
+            **kwargs) -> dict | Generator:
```
yarikoptic, line 75: @christian-monch it seems we have some possible non-uniformity.  Docstring here says `Union[Any, Generator]` but then nonasyncrunner..run has `def run(self) -> dict | Generator:` .  I wondered what it truly be -- `Any` sounds a bit too unspecific.

@jwodder in such cases where we have a composite type present/used in many spots, should we define some type to be reused?

jwodder: If it's just two short things like `dict | Generator`, I don't really see the need.

datalad/cmd.py
jwodder: Unless you want to completely restructure the code, the only real option is to add `assert self.runner is not None` once per method before any uses of `self.runner` where you, the human, can guarantee that `self.runner` has already been filled in.

datalad/runner/runner.py
christian-monch: I agree with @jwodder. Firstly, as he points out, it is just a short annotation, secondly, a telling name might not be shorter, and thirdly, we might clutter the name space if we define a lot of custom types

christian-monch: @yarikoptic: yes, the docstring is wrong. We should put `dict | Generator` as return type

datalad/cmd.py
christian-monch: There are two issues here:

1. Type checking fails
2. Initialisation

For the type checker we could just add pure declarations, e.g.:
```
from datalad.runner.nonasyncrunner import _ResultGenerator


    ...


        self.stdin_queue: Queue
        self.generator: _ResultGenerator
``` 
That would satisfy `mypy`.

Since the "declaration" does not create the attributes, we do not even need an `assert self.runner is not None`, because we would get an `AttributeError` if the attributes were accessed before they are initialized.

christian-monch: Will push a commit

christian-monch: Since I have no write access to your repo, I added PR #7006 which contains your changes and some changes from me on top. PR #7006, should obsolete this PR

yarikoptic: > Since I have no write access to your repo, 

you should have since I have marked "Allow edits and access to secrets by maintainers" on the right!  Did you try?

> I added PR #7006 which contains your changes and some changes from me on top. PR #7006, should obsolete this PR

works as well for me, thank you!

Issue #4598, opened 2020-05-30
embrace pathlib fully in .core?


yarikoptic:
Currently I see 19 uses in code
```
$> git grep -e '\<op[j.]' -- datalad/core | grep -v test_ | nl
     1	datalad/core/distributed/clone.py:                config_content = op.read_file(cfg_path)
     2	datalad/core/distributed/clone.py:                config_content = op.read_file(cfg_path)
     3	datalad/core/local/create.py:            op.normpath(op.join(str(path), os.pardir)))
     4	datalad/core/local/create.py:        if op.isdir(tbds.path) and listdir(tbds.path) != [] and not force:
     5	datalad/core/local/create.py:                [op.join('.datalad', i[0]) for i in attrs_cfg])
     6	datalad/core/local/create.py:                        op.join('.datalad', p), {}).get(k, None) == v:
     7	datalad/core/local/create.py:                    attrfile=op.join('.datalad', '.gitattributes'))
     8	datalad/core/local/diff.py:                orig_path.endswith(op.sep) or resolved_path == ds.pathobj
     9	datalad/core/local/run.py:from os.path import join as opj
    10	datalad/core/local/run.py:        rel_pwd = op.curdir
    11	datalad/core/local/run.py:        return [d for d in map(op.dirname, gpaths.expand(refresh=True))
    12	datalad/core/local/run.py:        if op.exists(path) or op.lexists(path):
    13	datalad/core/local/run.py:        pwd = normpath(opj(dataset.path, rel_pwd))
    14	datalad/core/local/run.py:        record_dir = ds.config.get('datalad.run.record-directory', default=op.join('.datalad', 'runinfo'))
    15	datalad/core/local/run.py:        record_path = op.join(ds_path, record_dir, record_id)
    16	datalad/core/local/run.py:        if not op.lexists(record_path):
    17	datalad/core/local/run.py:            msg_path = relpath(opj(str(repo.dot_git), "COMMIT_EDITMSG"))
    18	datalad/core/local/status.py:                            not (orig_path.endswith(op.sep) or
    19	datalad/core/local/status.py:                        super_root = get_dataset_root(op.dirname(root))
```
and about 159 in tests.  I think it would be beneficial to get `core` fully pathlib driven.

mih:
No change re state within the past 18 months. Sigh.

mih:
18 left!

yarikoptic:
[...]

PR #6907, opened 2022-08-02
RF: just rename internal variable from unclear "op" to "io"

yarikoptic:
came up while chatting with @AKsoo about https://github.com/datalad/datalad/issues/4598 .
That use of "op" (what does it stand for) was a "false positive" in trying to find
all uses of os.path


adswa (approved):
makes sense to me

yarikoptic:
CI - unrelated... damn them. merging

PR #6912, opened 2022-08-03
BF(TST,workaround): just xfail failing archives test on NFS

yarikoptic:
A workaround for https://github.com/datalad/datalad/issues/4101
with troubleshooting and trying various approaches done within
https://github.com/datalad/datalad/pull/6751

An ideal approach would be to check if create_tree does create a tree (and
archives with files it expect) matching the specification of the tree. Somehow
on NFS it seems to need some times a few seconds (sleep 3 before creating
archive helped according to testing in #6751) before filesystem actually reflect
creation of all files desired.

I found no robust solution which would not require "too much work" so decided
to just announce this an xfail.

For better or for worse, the loop of 100 invocations had them all PASS and none
XFAIL in https://app.travis-ci.com/github/datalad/datalad/builds/253793240 so
the effectiveness of this is not exactly confirmed yet ;)

### DIFF
--- a/datalad/tests/test_archives.py
+++ b/datalad/tests/test_archives.py
@@ -211,12 +211,18 @@ def test_ExtractedArchive(path=None):
 
     extracted_files = earchive.get_extracted_files()
     ok_generator(extracted_files)
-    eq_(sorted(extracted_files),
-        sorted([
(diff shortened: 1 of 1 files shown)


yarikoptic:
travis - 3.10 , another PR to fix it... all good for this xfail AFAIK

Issue #6917, opened 2022-08-04
datalad create returns git-annex SQLite3 Error


JohannesWiesner:
### What is the problem?

When trying  to create a dataset on a windows-based filesystem using a Linux OS, I get the following error:

```
[INFO   ] Detected a filesystem without fifo support.
|   Disabling ssh connection caching. 
[INFO   ] git-annex: SQLite3 returned ErrorBusy while attempting to perform close: unable to close due to unfinalized statements or unfinished backups 
[INFO   ] init: 1 failed 
CommandError: 'git -c diff.ignoreSubmodules=none annex init -c annex.dotfiles=true' failed with exitcode 1 under /zi-flstorage/group_csp/analyses/johannes.wiesner/projects/hcp_project/testing/foo
```

Looks like [this thread](https://git-annex.branchable.com/bugs/crippled_filesystem__58___SQLite3_returned_ErrorBusy/) deals with this problem and this error could possibly be out of scope (not something related to datalad itself?)

### What steps will reproduce the problem?

1.) Changing the working directory to a windows-based file system
2.) Executing `datalad create foo`

### Datalad information

Linux OS is: Linux Mint 19.3 "Tricia"
Datalad version is: 0.17.2

yarikoptic:
What is git-annex version and filesystem? `datalad wtf` output should have all interesting information

JohannesWiesner:
1.) The mounted file system uses cifs

2.) `datalad wtf`:

```
# WTF
## configuration <SENSITIVE, report disabled by configuration>
## credentials 
  - keyring: 
    - active_backends: 
      - SecretService Keyring
      - PlaintextKeyring with no encyption v.1.0 at /home/johannes.wiesner/.local/share/python_keyring/keyring_pass.cfg
    - config_file: /home/johannes.wiesner/.config/python_keyring/keyringrc.cfg
    - data_root: /home/johannes.wiesner/.local/share/python_keyring
## datalad 
  - version: 0.17.2
## dependencies 
  - annexremote: 1.5.0
  - boto: 2.49.0
  - cmd:7z: 16.02
  - cmd:annex: 10.20220504-g35cafb7
  - cmd:bundled-git: UNKNOWN
  - cmd:git: 2.35.3
  - cmd:ssh: 9.0p1
  - cmd:system-git: 2.35.3
  - cmd:system-ssh: 9.0p1
  - exifread: 3.0.0
  - humanize: 4.1.0
  - iso8601: 1.0.2
[...]

PR #6920, opened 2022-08-05
BF+ENH(TST): fix typo in code of wtf filesystems reports

yarikoptic:
Detected while looking at https://github.com/datalad/datalad/issues/6917#issuecomment-1205048868 and it was impossible to detect that there is an underlying issue in the code since no logging and swallowing of all exceptions silently.  Added logging, and some rudimentary testing (let's see if works on windows. edit: appveyor was green, so I guess it did!).

### DIFF
--- a/datalad/local/tests/test_wtf.py
+++ b/datalad/local/tests/test_wtf.py
@@ -37,6 +37,8 @@ from datalad.tests.utils_pytest import (
 )
 from datalad.utils import ensure_unicode
 
+from datalad.support.external_versions import external_versions
+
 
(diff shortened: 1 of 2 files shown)


yarikoptic:
FWIW I have CPed the actual fix (b5fd8cd0f3fe64f332401791dea3d305f18e68fe) into `maint` since there is nothing to discuss there. Will keep this PR for review of the rest (testing etc).

adswa (approved):
Thanks for catching this, and adding a test. And sorry for the vacation-related delay

Issue #6929, opened 2022-08-10
DataLad GitHub Action


thewtex:
### What is the problem?

Hi folks!

I am interested in a DataLad GitHub Action to provide CI testing data. Functionality

- Download / install DataLad
- Across platforms
- Possibly of a specific version
- Optionally `clone` and `get` datasets

### Have you had any success using DataLad before?

Yes, with a bit of help from @yarikoptic :-) I recently read [FAIRly big: A framework for computationally reproducible processing of large-scale data](https://www.nature.com/articles/s41597-022-01163-2) and was inspired. Going through the wonderful handbook now.

yarikoptic:
I think it indeed would make total sense to simplify life of others.  What "action API" (options etc) do you envision? We might want to start with some rudimentary design doc PR into a file under https://github.com/datalad/datalad/tree/master/docs/source/design ?

But I think it probably should be a separate repo since overall would have its own life cycle.  Do you have any experience with establishing github actions? is there any template to start from ? (attn @vsoch and @jwodder who might know more)

vsoch:
It's pretty easy to make actions - when y'all have a design I'd be happy to take an initial shot, or let me know if install/get is a good start!

yarikoptic:
Yeah, I think at large it is `install/get`, and I would say 

- should use datalad-installer for install needs. Command should be smth like `datalad-installer -E {miniconda_env_file:=~/miniconda_env.sh} miniconda --path {miniconda_path:=~/miniconda} miniconda git-annex -m conda datalad -m conda` (might need dedicated `git` for windows https://github.com/datalad/datalad-installer/issues/46 attn @jwodder). 
(I used bash-inspired {var:=DEFAULT} above)

- action configuration should expose those options above ({miniconda*})
```yaml
[miniconda_env_file: str ...]
[miniconda_path: str ... # if exists, could just reuse]
datasets: list[dict(source: str, [path: str], [recurisive:bool], [recursion_limit:int], [get_paths: list[str]] ]
  
```

with [] marking optional, and those dicts in `datasets` 
- `source` : url for dataset to install
- `path`: path to install under
- `recursive` either to install recursively
- `recursion_limit` - possibly limit recursion
- `get_paths` - paths to `get`

so for this step it is roughly:

```python
import datalad.api as dl
for ds_spec datasets_spec:
   # kw should be populated based on relevant options
   ds = dl.install(source=ds_spec.source, **kw)
   ds.get(ds_spec.get_paths) 
```

[...]

PR #6931, opened 2022-08-10
Describe a DataLad GitHub Action

thewtex:
#### 📝 Documentation
- Describe specifications for a DataLad GitHub Action re: #6929. These may find their way to the GitHub Action README.

[ci skip]
[appveyor skip]

### DIFF
--- a/docs/source/design/github_actions.rst
+++ b/docs/source/design/github_actions.rst
@@ -1,0 +1,126 @@
+.. -*- mode: rst -*-
+.. vi: set ft=rst sts=4 ts=4 sw=4 et tw=79:
+
+.. _chap_design_github_action:
+
+*******************************
+GitHub Action
+*******************************
+
+.. topic:: Specification scope and status
+
+   This specification describes a proposed interface to a DataLad GitHub Action.
+   https://github.com/datalad/datalad-action provides an implementation which loosely
+   followed this specification.
+
+The purpose of the DataLad GitHub Action is to support CI testing with DataLad datasets
+by making it easy to install ``datalad`` and ``get`` data from the datasets.
+
+
+Example Usage
+=============
+
+Dataset installed at ``${GITHUB_WORKSPACE}/studyforrest-data-phase2``,
+``get``'s all the data::
+
+    - uses: datalad/datalad-action@master
+      with:
+        datasets:
+          - source: https://github.com/psychoinformatics-de/studyforrest-data-phase2
+          - install_get_data: true
+
+Specify advanced options::
+
+    - name: Download testing data
+      uses: datalad/datalad-action@master
+      with:
+        datalad_version: ^0.15.5
+        add_datalad_to_path: false
+        datasets:
+          - source: https://github.com/psychoinformatics-de/studyforrest-data-phase2
+          - branch: develop
+          - install_path: test_data
+          - install_jobs: 2
+          - install_get_data: false
+          - recursive: true
+          - recursion_limit: 2
+          - get_jobs: 2
+          - get_paths:
+              - sub-01
+              - sub-02
+              - stimuli
+
+Options
+=======
+
+``datalad_version``
+-------------------
+
+``datalad`` version to install. Defaults to the latest release.
+
+``add_datalad_to_path``
+-------------------
+
+Add ``datalad`` to the ``PATH`` for manual invocation in subsequent steps.
+
+Defaults to ``true``.
+
+``source``
+----------
+
+URL for the dataset (mandatory).
+
+``branch``
+----------
+
+Git branch to install (optional).
+
+``install_path``
+----------------
+
+Path to install the dataset relative to `GITHUB_WORKSPACE`.
+
+Defaults to the repository name.
+
+``install_jobs``
+----------------
+
+Jobs to use for ``datalad install``.
+
+Defaults to ``auto``.
+
+``install_get_data``
+----------------
+
+Get all the data in the dataset by passing ``--get-data`` to ``datalad install``.
+
+Defaults to ``false``.
+
+``recursive``
+-------------
+
+Boolean defining whether to clone subdatasets.
+
+Defaults to ``true``.
+
+``recursion_limit``
+-------------------
+
+Integer defining limits to recursion.
+
+If not defined, there is no limit.
+
+``get_jobs``
+----------------
+
+Jobs to use for ``datalad get``.
+
+Defaults to ``auto``.
+
+
+``get_paths``
+-------------------
+
+A list of paths in the dataset to download with ``datalad get``.
+
+Defaults to everything.


docs/source/design/github_actions.rst
```diff
@@ -0,0 +1,84 @@
...
+
+The purpose of the DataLad GitHub Action is to support CI testing with DataLad datasets
+by making it easy to install ``datalad`` and ``get`` data from the dataset.
+
+
+Example Usage
+=============
+
+Dataset installed at ``${GITHUB_WORKSPACE}/studyforrest-data-phase2``,
+``get``'s all the data::
```
yarikoptic, line 22: are you sure that getting all the data should be default?  E.g. in this demo use case it would be 14GB! (will CI have enough time to fetch it all? ;-))

docs/source/design/github_actions.rst
```diff
@@ -0,0 +1,84 @@
...
+        datalad_version: ^0.15.5
+        datasets:
+          - source: https://github.com/psychoinformatics-de/studyforrest-data-phase2
+          - install_path: test_data
+          - recursive: true
+          - recursion_limit: 2
+          - get_paths:
+              - sub-01
+              - sub-02
+              - stimuli
```
yarikoptic, line 43: both `install` and `get` have `--jobs` options. We might want to add here as `jobs` and `get_jobs`.

Also, for the use-case of "get all data", I think we might better just add also `get_data` option to invoke `install` with `--get-data` and avoid individual `get_paths`. Also `get_paths` on itself might be "tricky" to ask for all data -- i.e. `.` would not be recursive.

docs/source/design/github_actions.rst
```diff
@@ -0,0 +1,84 @@
...
+          - source: https://github.com/psychoinformatics-de/studyforrest-data-phase2
+
+Specify advanced options::
+
+    - name: Download testing data
+      uses: datalad/datalad-action@master
+      with:
+        datalad_version: ^0.15.5
+        datasets:
+          - source: https://github.com/psychoinformatics-de/studyforrest-data-phase2
```
yarikoptic, line 36: I believe that actions could add to GITHUB_ENV file thus augmenting PATH. Let's add option 

```
add_datalad_to_path: bool
```

which should then add its miniconda-installed datalad path into PATH so subsequent actions could make use of the `datalad` command.  The fear is that conda env might interfere with the rest of the setup.  We might want to provide only some "datalad" shim which would then export PATH to conda for its duration.

Than we can allow for arbitrary desired flexibility of `datalad get` invocations in subsequent steps.

docs/source/design/github_actions.rst
thewtex: My use case involves a dedicated small unit testing dataset, but, yes I can see how result in a large download :-).

docs/source/design/github_actions.rst
thewtex: I made a pass at this -- please let me know if I did not follow correctly and more changes are needed!

docs/source/design/github_actions.rst
thewtex: Good idea, added!

docs/source/design/github_actions.rst
```diff
@@ -0,0 +1,118 @@
...
+Specify advanced options::
+
+    - name: Download testing data
+      uses: datalad/datalad-action@master
+      with:
+        datalad_version: ^0.15.5
+        add_datalad_to_path: false
+        datasets:
+          - source: https://github.com/psychoinformatics-de/studyforrest-data-phase2
+          - install_path: test_data
```
yarikoptic, line 39: Let's add optional `branch` option to be able to checkout a specific branch or tag.  Might also want `--depth` but not sure how that would play with `git-annex` branch -- didn't experiment.

A little more of history/details: we at large didn't uniformly address https://github.com/datalad/datalad/issues/2109 in being able to install dataset at a given committish.

There was https://github.com/datalad/datalad/pull/5662 which was superseded with https://github.com/datalad/datalad/pull/6218 so it became possible to pass git options to `datalad clone` so it is possible to use `datalad clone URL --branch BRANCH` to point to specific branch or tag. We have also explicit `--branch` option in `datalad install`. In either of those cases that still does not allow to point to specific commit hexsha.  Since besides `--branch` , the `--depth` and may be other git options could come handy, may  be it would be better to code action as sequence of `datalad clone {url} [{path}] [{git_options}]` , `datalad -C {path} get` with `--no-data` and  `-r` options depending on settings instead of a single `datalad install` command.

thewtex: :+1: done.

vsoch:
Question (and going to work on this tomorrow) - is there possibly another functionality aside "get"? E.g., for actions I like to design them as small modules, so you could do like:

```yaml
    - name: Install Datalad
      uses: datalad/datalad-action/install@main
```
```yaml
# This is the "get" idea
    - name: Download Dataset
      uses: datalad/datalad-action/download@main
```
```yaml
    - name: Something else?
      uses: datalad/datalad-action/??@main
```
Minimally if we can install / get, those should be separate (and the get can use the install) so that's my proposed modification to the above!

docs/source/design/github_actions.rst
```diff
@@ -0,0 +1,124 @@
...
+
+Example Usage
+=============
+
+Dataset installed at ``${GITHUB_WORKSPACE}/studyforrest-data-phase2``,
+``get``'s all the data::
+
+    - uses: datalad/datalad-action@master
+      with:
+        datasets:
```
vsoch, line 26: Just FYI (as far as I know) you cannot have nested fields like this.

vsoch:
@yarikoptic can you make me a member of the org or create datalad-action and make me maintainer? I have the install start ready and I'd like to push and test and then work on get.

vsoch:
Could someone show me an example of the python and/or bash you want for the get action? I see both as examples in the discussion so I'm not sure which to use - it took me 10 minutes to find (what I think) is the install function in core/local/run.py but I don't see any mention of recursion...

yarikoptic:
sorry @vsoch - I am not entirely sure what example you would like? but in general anything "datalad COMMAND [OPTIONS] ARGS" roughly corresponds to

```python
import datalad.api as dl
dl.COMMAND(*args, **OPTIONS)
```

and imho `bash` scripts are good too, just make them `set -eu` on the top to ensure them not struggling through the pile of errors 

In general -- as long as there is action which simplifies installation of datalad, and exposes it to `run` of subsequent steps/actions -- I do not think there is much need for really modular reflection of various datalad commands in the action -- `datalad COMMANDS` could be just coded in across platforms.  I think the main goal of having datalad action is to condense and simplify for simplistic use case  - get datasets with some data, and for that git-annex and datalad need to be installed and few installations targets given.

vsoch:
@yarikoptic I already figured it out.

yarikoptic:
@thewex, there is https://github.com/datalad/datalad-action now from @vsoch . Could you have a look at it and give it a shot?

docs/source/design/github_actions.rst
```diff
@@ -0,0 +1,124 @@
...
+
+.. _chap_design_github_action:
+
+*******************************
+GitHub Action
+*******************************
+
+.. topic:: Specification scope and status
+
+   This specification describes a proposed interface to a DataLad GitHub Action.
```
yarikoptic, line 12:
```suggestion
   This specification describes a proposed interface to a DataLad GitHub Action.
   https://github.com/datalad/datalad-action provides an implementation which loosely
   followed this specification.
```

yarikoptic (approved):
Let's adjust status (proposed suggestion), take out of draft, and get it merged - we have an implementation by @vsoch so any further discussion etc should happen in its issues.

docs/source/design/github_actions.rst
thewtex: :+1: done

docs/source/design/github_actions.rst
```diff
@@ -0,0 +1,126 @@
...
+Options
+=======
+
+``datalad_version``
+-------------------
+
+``datalad`` version to install. Defaults to the latest release.
+
+``add_datalad_to_path``
+-------------------
```
mih, line 62: I am not sure why the doc build CI did not run here, but this is invalid RST, and broke it:

```
Warning, treated as error:
167
/home/runner/work/datalad/datalad/docs/source/design/github_actions.rst:62:Title underline too short.
```

yarikoptic-gitmate:

PR #6932, opened 2022-08-10
ENH: GitRepo -- make use of --pathspec-from-file where possible and chunking needed

### LINKED ISSUE #6922: Avoid passing long lists of paths to `git` in CLI
ATM we just pass them on command line, quite often requiring to split into multiple invocations on "chunks" of file paths, and even `--amend` previous commit.  It can potentially lead to "odd" situations whenever such a "partial commit" is pushed to a sibling (https://github.com/dandi/dandisets/issues/230 ).

So ideally we should avoid doing that whenever possible.  Hence I asked git folks about how could we do that

<details>
<summary>A full reply to my question</summary> 

```
Date: Mon, 8 Aug 2022 14:12:56 +0100                                                                                                                  
From: Phillip Wood <xxx>                                                                                                        
To: Yaroslav Halchenko <xxx>, "git@vger.kernel.org" <git@vger.kernel.org>                                                              
Subject: Re: --batch or some --paths-file for very long lists of paths                                                                                
                                                                                                                                                      
Hi Yaroslav                                                                                                                                           
                                                                                                                                                      
On 04/08/2022 23:59, Yaroslav Halchenko wrote:                                                                                                        
> Dear Git Gurus,                                                                                                                                     
>                                                                                                                                                     
> In DataLad (https://datalad.org) we are doing "our own" analysis of what                                                                            
> specific files (not entire directories) should git and git-annex operate                                                                            
> on.  Obviously, in large repositories (and we have with >100k files)                                                                                
> that might require invoking  git add  or  git diff  etc with a long list                                                                            
> of paths specified in the command line.  For that we often split                                                                                    
> invocation into multiple and even resort to   git commit --amend  to                                                                                
> combine multiple commits then into a single one.                                                                                                    
>                                                                                                                                                     
> But I wondered if may be there is already some trick to make such                                                                                   
> commands as   status, diff, add, commit   to operate on arbitrarily long                                                                            
> list of paths passed to that git command somehow.                                                                                                   
                                                                                                                                                      
A number of porcelain commands have a --pathspec-from-file option that takes a                                                                        
file with a list of pathspecs or reads them from stdin. When combined with                                                                            
--pathspec-file-nul this handles paths containing newline correctly or you can                                                                        
quote them without this option. You can pass --literal-pathspecs if you have a                                                                        
list of paths rather than pathspecs.                                                                                                                  
                                                                                                                                                      
At the plumbing level you can use "update-index" to add/delete/update paths in                                                                        
the index which will read paths from stdin and "checkout-index" will also read                                                                        
paths from stdin.                                                                                                                                     
                                                                                                                                                      
The diff family do not have any support for --pathspec-from-file at the moment                                                                        
but I'd be happy to see someone implement it (I think it would be fairly                                                                              
straight forward).                                                                                                                                    
                                                                                                                                                      
> Note that gitglossary (at least in 2.35.1 git on debian) says that                                                                                  
>                                                                                                                                                     
>       Pathspecs are used on the command line of "git ls-files", "git                                                                                
>      ls-tree", "git add", "git grep", "git diff", "git checkout", and many other                                                                    
>      commands ...                                                                                                                                   
>                                                                                                                                                     
> but                                                                                                                                                 
>                                                                                                                                                     
>       $> git ls-tree -h | head -n1                                                                                                                  
>       usage: git ls-tree [<options>] <tree-ish> [<path>...]                                                                                         
>                                                                                                                                                     
> so it is <path> not the <pathspec> like (why in stderr this time?)                                                                                  
>                                                                                                                                                     
>       $> git commit -h 2>&1 | head -n1                                                                                                              
>       usage: git commit [<options>] [--] <pathspec>...                                                                                              
>                                                                                                                                                     
> So if in both cases it is pathspec, may be pathspec could support some                                                                              
> other magical keyword like :(filelist)/tmp/mylonglistofpaths ?                                                                                      
                                                                                                                                                      
I like that path magic idea, but as we already have --pathspec-from-file I                                                                            
think we'd be better improving support for that.                                                                                                      
                                                                                                                                                      
Best Wishes                                                                                                                                           
                                                                                                                                                      
Phillip                                                                                                                                               
                                                                                                                                                      
> Thanks in advance for your time and thoughts,   
```
</details>

so I was pointed to two "options"

- use --pathspec-from-file where possible (see below for list of commands supporting it, but might need to check since what versions of git!?)
- and plumbing level `update-index`  which can read from stdin (ATM we do use it for removal in `save_`)

```shell
(git)lena:~/proj/misc/git[master]git
$> git describe
v2.37.1-377-g679aad9e82

$> git grep -e '^--pathspec-from-file=' -- 'Documentation/git*'
Documentation/git-add.txt:--pathspec-from-file=<file>::
Documentation/git-checkout.txt:--pathspec-from-file=<file>::
Documentation/git-commit.txt:--pathspec-from-file=<file>::
Documentation/git-reset.txt:--pathspec-from-file=<file>::
Documentation/git-restore.txt:--pathspec-from-file=<file>::
Documentation/git-rm.txt:--pathspec-from-file=<file>::
Documentation/git-stash.txt:--pathspec-from-file=<file>::
```

and we should then pait with `--literal-pathspecs` whenever we know that it is a PATH and not pathspecs (dirty PR was sent as #6921 )

yarikoptic:
Chunking is a kludge to workaround the problem of DataLad needing to provide a
long list of files to git/git-annex commands.  Whenever git-annex has --batch mode
which we use typically for commands which could get lots of file paths arguments, we
have not used any special means like that for git commands.  Many commands support
--pathspec-file-nul and --pathspec-from-file=FILE options to pass a list of pathspec(s)
to operate on via a FILE instead of command line.

In this PR we decide to pass via such a file in cases where more than a single
chunk is necessary.

The main change is to `GitRepo.commit` which used to resort to sequence of commits with `--amend` to establish the desired commit.

- [x] rebase on master (I think branch is really on top of `maint`, might not matter)
- [x] establish "thorough testing".  Since tests do not operate on heavy lists of files, and not just one but a number of commands make use of `pathspec_from_file`, I will establish `datalad.runtime.prefer_pathspec_from_file: bool` which would cause use of `--pathspec-from-file` regardless on either chunking is needed. Then one of the matrix runs of travis will enable that config setting to get commands covered (probably the one for "bleeding edge everything")

Closes #6922

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -84,11 +84,13 @@ matrix:
     env:
     - PYTEST_SELECTION_OP=not
     - DATALAD_SSH_MULTIPLEX__CONNECTIONS=0
+    - DATALAD_RUNTIME_PATHSPEC__FROM__FILE=always
     - _DL_ANNEX_INSTALL_SCENARIO="miniconda --python-match minor --batch git-annex=10.20220525 -m conda"
   - python: 3.7
     env:
     - PYTEST_SELECTION_OP=""
     - DATALAD_SSH_MULTIPLEX__CONNECTIONS=0
+    - DATALAD_RUNTIME_PATHSPEC__FROM__FILE=always
     - _DL_ANNEX_INSTALL_SCENARIO="miniconda --python-match minor --batch git-annex=8.20210310 -m conda"
     # To test https://github.com/datalad/datalad/pull/4342 fix in case of no "not" for pytest.
     # From our testing in that PR seems to have no effect, but kept around since should not hurt.
--- a/.travis.yml
+++ b/.travis.yml
@@ -269,7 +269,7 @@ script:
   # Run tests
   #  Note: adding --log-cli-level=INFO would result in DATALAD_LOG_TARGET=/dev/null being not
   #  in effect, dumping too many logs.
-  - http_proxy=
+  - set -x; http_proxy=
     PATH=$PWD/../tools/coverage-bin:$PATH
     $PYTEST_WRAPPER python
       -m pytest "${PYTEST_OPTS[@]}"
--- a/datalad/dataset/gitrepo.py
+++ b/datalad/dataset/gitrepo.py
@@ -27,6 +27,7 @@ from contextlib import contextmanager
 from locale import getpreferredencoding
 from os import environ
 from os.path import lexists
+from typing import Optional
 from weakref import (
     finalize,
     WeakValueDictionary
@@ -271,8 +272,10 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
 
     def _generator_call_git(self,
                             args,
+                            *,
                             files=None,
                             env=None,
+                            pathspec_from_file: Optional[bool]=False,
                             sep=None):
         """
         Call git, yield stdout and stderr lines when available. Output lines
@@ -330,7 +333,9 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
                 cmd,
                 files,
                 protocol=GeneratorStdOutErrCapture,
-                env=env)
+                env=env,
+                pathspec_from_file=pathspec_from_file,
+            )
         else:
             generator = self._git_runner.run(
                 cmd,
@@ -360,6 +365,7 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
                   expect_stderr=False,
                   expect_fail=False,
                   env=None,
+                  pathspec_from_file: Optional[bool] = False,
                   read_only=False):
         """Allows for calling arbitrary commands.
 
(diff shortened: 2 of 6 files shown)


yarikoptic:
ok, the issue is reported https://github.com/pypa/setuptools/issues/3501 ... so seems to relate to versionner and workarounds seems to be suggested such as https://github.com/shapely/shapely/pull/1475/files  from https://github.com/pypa/setuptools/issues/3501#issuecomment-1212476025 .  I guess I will submit a dedicated PR with that one.

mih (approved):
This makes sense and is a verz useful feature. Thx!

yarikoptic-gitmate:

PR #6934, opened 2022-08-11
nothing: testing if it is a global wtf with setuptools

yarikoptic:
https://github.com/datalad/datalad/runs/7793747019?check_suite_focus=true
and others in #6932 are now failing all of a sudden!  Smells the like newest
and greatest setuptools added some deprecation


yarikoptic:
ok, appveyor is green and the rest seems tobe happily running without early crashing so all good back

PR #6935, opened 2022-08-11
BUG: fix versioneer.py to work with PEP 660

yarikoptic:
Patch from https://patch-diff.githubusercontent.com/raw/shapely/shapely/pull/1475.patch mentioned in https://github.com/pypa/setuptools/issues/3501 to workaround that issue which just came up and leading all our builds to error out.

### DIFF
--- a/versioneer.py
+++ b/versioneer.py
@@ -1826,6 +1826,8 @@ def get_cmdclass(cmdclass=None):
             cfg = get_config_from_root(root)
             versions = get_versions()
             _build_py.run(self)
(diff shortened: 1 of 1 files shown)


yarikoptic:
it seems there is a new release of setuptools https://github.com/pypa/setuptools/releases/tag/v64.0.2 which should include the https://github.com/pypa/setuptools/pull/3506 fix... will test it in #6934

yarikoptic:
yeap, no longer needed, resolved at setuptools level

PR #6936, opened 2022-08-12
DOC: fix capitalization of some service names

aqw:
I noticed a "Github" in a config file while helping someone with DataLad, and it caused me to twitch enough to fix it and a few other strings I found. ;-)

This is by no means comprehensive, but I did some `grep`ing around and tried to catch all I could find.

I intentionally chose to not fix occurrences in code comments and tests.

- Github -> GitHub
- openfmri -> OpenfMRI
- s3 -> S3
- travis -> Travis CI

This also fixes "Datalad" mis-capitalizations. For this, I did fix all occurrences, to make future `grep`ing easier.
- Datalad -> DataLad

#### 📝 Documentation
- fix capitalization of some service and software names

### DIFF
--- a/.github/ISSUE_TEMPLATE/issue_template.yml
+++ b/.github/ISSUE_TEMPLATE/issue_template.yml
@@ -16,7 +16,7 @@ body:
 
   - type: textarea
     attributes:
-      label: Datalad information
+      label: DataLad information
       description: |
         What version of DataLad and git-annex do you use (run `datalad --version`)? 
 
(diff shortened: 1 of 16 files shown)


aqw:
Well, it seems like most tests fail. But I don't know if that's significant or normal.

If it's atypical, let me know and I'll take a look. But these changes aren't all that important, so it's not worth a tooling struggle to get them in.

yarikoptic:
Looks good to me. I will need to cherrypick it into maint.. CI - hopefully we would fix it soon (not our fault)

yarikoptic-gitmate:

PR #6937, opened 2022-08-12
BF: git_ignore_check do not overload possible value of stdout/err if present

yarikoptic:
A little more discussion at https://github.com/datalad/datalad/pull/6930 and initially
identified while working on https://github.com/datalad/datalad/pull/6932 and being
unable to get CommandError.stdout for matching while using call_git (but works
with _call_git using different code path, which does not have this git_ignore_check).

One of the two locations where this function is used has stdout=None, so this
lead to wiping out of e.stdout even if it was present.  I think it was not intentional
but @christian-monch would know better

### DIFF
--- a/datalad/dataset/gitrepo.py
+++ b/datalad/dataset/gitrepo.py
@@ -76,8 +76,8 @@ def git_ignore_check(expect_fail,
     try:
         yield None
     except CommandError as e:
-        e.stdout = "".join(stdout_buffer) if stdout_buffer else ""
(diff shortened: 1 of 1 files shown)


yarikoptic:
woohoo -- it is green! only crawler wants something but I do not think it is related -- I saw it failing in other PRs.

christian-monch (approved):

christian-monch (approved):
Good extension to other use cases, thx

Issue #6941, opened 2022-08-13
`GitRepo.get_submodules_()` slow with many directories/files and few subdatasets


mih:
`GitRepo.get_content_info()` is called by `GitRepo.get_submodules_()` and also calls `GitRepo.get_submodules_()` in order to ensure compatibility with recent Git versions. "Introduced" by https://github.com/datalad/datalad/commit/8ff86134405dcd125e41fe3a2544ebc7454efc32

This is a grave performance issue: Any performance-conscious command would call `GitRepo.get_content_info()` with `paths` constraints. On any modern system with Git > 2.29.0 a call with `ref=None` (part of any `status()`) will trigger a call to `GitRepo.get_submodules_()`, which will in turn trigger an unconstrained call to `GitRepo.get_content_info()` https://github.com/datalad/datalad/issues/6940 Performance on any sizable dataset, in terms of numbers of files, with a least one subdataset will be severely impacted proportional to the number of files, regardless of whether the calling code uses path-contraints.

yarikoptic:
> `GitRepo.get_content_info()` is called by `GitRepo.get_submodules_()` and also calls `GitRepo.get_submodules_()` in order to ensure compatibility with recent Git versions. This can lead to
> 
> ```
> RecursionError: maximum recursion depth exceeded
> ```

I wish there was an example which demonstrated that. I thought to add a unittest to ensure that we do not get it but looking at the code I do not see how infinite recursion could happen since inside `get_content_info` we call back `get_modules_` only if `paths` were provided, but without providing `paths` into that recursive call, so not clean how this could happen to me :-/

mih:
I retitled and edited the top post to remove this aspect from this issue. It is not relevant. The performance is.

yarikoptic:
then duplicate of https://github.com/datalad/datalad/issues/6940 ?

mih:
Yes. closing

Issue #6940, opened 2022-08-13
`subdatasets` slow for no apparent reason


mih:
Scenario: dataset with two non-installed subdatasets and 40k top-level subdirectories:

```
% datalad -f disabled subdatasets 
datalad -f disabled subdatasets  91.01s user 11.94s system 100% cpu 1:42.54 total
% time git submodule
-dfa6d975ea8888ed33bf714c674371e1ab54e834 code/pipeline
-0c7f0b45140dde1d7291b15725e2016231a670b0 inputs/ukb
git submodule  0.02s user 0.02s system 105% cpu 0.035 total
% cat .gitmodules
[submodule "code/pipeline"]
        path = code/pipeline
        url = http://containers.ds.inm7.de/alias/cat
        datalad-id = 99300fb0-ea39-4fd2-b36d-2afc1e9ab7b2
[submodule "inputs/ukb"]
        path = inputs/ukb
        url = http://ukb.ds.inm7.de/alias/bids
        datalad-id = 14a5e5b6-d30d-11ea-8836-b4969157768c
% git submodule
-dfa6d975ea8888ed33bf714c674371e1ab54e834 code/pipeline
-0c7f0b45140dde1d7291b15725e2016231a670b0 inputs/ukb
```

So 35ms (git) vs 1m42s (datalad). :facepalm: 

And the reason seems to be in `GitRepo.get_submodules_()`:

https://github.com/datalad/datalad/blob/6b7a2f7cf480861ce44d162b4946f108acecf027/datalad/support/gitrepo.py#L2357-L2368

It calls `get_content_info()` to get *everything* about the dataset, only to discard anything that is not a `type='dataset'`. In this case of 2 subdatasets and 80k directories this is *highly* inefficient.

It seems to make sense to limit the call to `get_content_info()` to the paths obtain from `_parse_gitmodules()` already (or even a dedicated `git submodule` call).

`get_submodules_()` currently yields records like this:

```py
 {'gitmodule_datalad-id': '14a5e5b6-d30d-11ea-8836-b4969157768c',
  'gitmodule_name': 'inputs/ukb',
  'gitmodule_url': 'http://ukb.ds.inm7.de/alias/bids',
  'gitshasum': '0c7f0b45140dde1d7291b15725e2016231a670b0',
  'path': PosixPath('/home/mih/ukb-gource/ukb-derivatives/inputs/ukb'),
  'type': 'dataset'}
```

[...]

PR #6942, opened 2022-08-14
Reimplement GitRepo.get_submodules_() without get_content_info()

mih:
The previous implementation resulted in a circular dependency after
a check for a Git-version dependent reporting behavior was introduced
with 8ff86134405dcd125e41fe3a2544ebc7454efc32

Moreover, the previous implementation also queried a repository for
all content, merely to discard any record that is not a subdataset.
This could slow a responds dramatically for dataset with few subdatasets
but many files.

This change reimplements `get_submodules_()` with a plain call to
`git ls-files` parameterized with the paths of recorded submodules
taken from `.gitmodules`.

This changes the behavior to be more in-line with `git submodule`
by refusing to report on submodules that are not recorded in
.gitmodules.

Fixes datalad/datalad#6940
Fixes datalad/datalad#6941
Fixes datalad/datalad#6943

This PR goes against `master`, because it changes behavior, even if only in corner-cases. The fix for datalad/datalad#6941 is also cherry-picked into a PR against `maint` https://github.com/datalad/datalad/pull/6944

The speed-up is substantial for datasets with many files and at least one subdataset. I have a real-world test cased with a dataset with 160k files and 2 subdatasets showing a speed-up of 350x.

See here for another real-world performance change estimate https://github.com/datalad/datalad-next/pull/92#issuecomment-1214308010

### DIFF
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -2344,7 +2344,10 @@ class GitRepo(CoreGitRepo):
         Parameters
         ----------
         paths : list(pathlib.PurePath), optional
-            Restrict submodules to those under `paths`.
+            Restrict submodules to those under `paths`. Paths must be relative
+            to the resolved repository root, and must be normed to match the
+            reporting done by Git, i.e. no parent dir components
+            (ala "some/../this").
 
         Returns
         -------
@@ -2355,17 +2358,67 @@ class GitRepo(CoreGitRepo):
             return
 
         modinfo = self._parse_gitmodules()
-        for path, props in self.get_content_info(
-                paths=paths,
-                ref=None,
-                untracked='no').items():
-            if props.get('type', None) != 'dataset':
+        if not modinfo:
+            # we exit early, if there is nothing on record (even though
+            # a .gitmodules file exists).
+            # without this conditional exit, we would be able to discover
+            # subprojects even when they are not recorded in .gitmodules,
+            # but at the cost of running an unconstrained ls-files call
+            # below
+            return
+
+        # taken from 3.9's pathlib, can be removed once the minimally
+        # supported python version
+        def is_relative_to(self, *other):
+            """Return True if the path is relative to another path or False.
+            """
+            try:
+                self.relative_to(*other)
+                return True
+            except ValueError:
+                return False
+
+        if paths:
+            # ease comparison
+            paths = [self.pathobj / p for p in paths]
+            # contrain the report by the given paths
+            modinfo = {
+                # modpath is absolute
+                modpath: modprobs
+                for modpath, modprobs in modinfo.items()
+                # is_relative_to() also match equal paths
+                if any(is_relative_to(modpath, p) for p in paths)
+            }
+        for r in self.call_git_items_(
+            ['ls-files', '--stage', '-z'],
+            sep='\0',
+            files=[str(p.relative_to(self.pathobj)) for p in modinfo.keys()],
+            read_only=True,
+            keep_ends=True,
+        ):
+            if not r.startswith('160000'):
                 # make sure this method never talks about non-dataset
                 # content
                 continue
-            props["path"] = path
-            props.update(modinfo.get(path, {}))
-            yield props
+            props, rpath = r.split('\t')
+            mode, gitsha, stage = props.split(' ')
+            if stage not in ('0', '2'):
+                # we either have non-merge situation, or a simple merge
+                # situation (i.e. stage=0). the reported gitsha always
+                # matches what we have locally.
+                # or we are in a three-way merge, in which case stage=2
+                # is what we want to report, because it matches the
+                # current HEAD (see git-read-tree manpage).
+                # there is either a stage 2 or stage 0, never both
+                continue
+            # remove the expected line separator from the path
+            path = self.pathobj / PurePosixPath(rpath[:-1])
+            yield dict(
+                path=path,
+                type='dataset',
+                gitshasum=gitsha,
+                **modinfo.get(path, {})
+            )
 
     def get_submodules(self, sorted_=True, paths=None):
         """Return list of submodules.
--- a/datalad/tests/test_config.py
+++ b/datalad/tests/test_config.py
@@ -704,7 +704,7 @@ def test_write_config_section(path=None):
     # can we handle a bare repo?
     gr = GitRepo(path, create=True, bare=True)
 
-    obscure = "ds-;&%b5{}'ΔЙקم๗あ.datc"
+    obscure = "ds-; &%b5{}# some % "
     # test cases
     # first 3 args are write_config_section() parameters
     # 4th arg is a list with key/value pairs that should end up in a
@@ -721,16 +721,13 @@ def test_write_config_section(path=None):
         ('short', ' s p a c e ', {"a123": ' space all over '}, [
             ('short. s p a c e .a123', ' space all over '),
         ]),
+        ('submodule', obscure, {
+            'path': obscure,
+            'url': f"./{obscure}"}, [
+            (f"submodule.{obscure}.path", obscure),
+            (f"submodule.{obscure}.url", f"./{obscure}"),
+        ]),
     ]
-    if not on_windows:
-        # on windows our test setup lacks the full unicode fancy
-        testcfg.append(
-            ('submodule', obscure, {
-                'path': obscure,
-                'url': f"./{obscure}"}, [
-                (f"submodule.{obscure}.path", obscure),
-                (f"submodule.{obscure}.url", f"./{obscure}"),
-            ]))
 
     for tc in testcfg:
         # using append mode to provoke potential interference by
--- a/datalad/tests/test_config.py
+++ b/datalad/tests/test_config.py
@@ -50,6 +50,7 @@ from datalad.tests.utils_pytest import (
 from datalad.utils import (
     Path,
     get_home_envvars,
+    on_windows,
     swallow_logs,
 )
 
@@ -703,6 +704,7 @@ def test_write_config_section(path=None):
     # can we handle a bare repo?
     gr = GitRepo(path, create=True, bare=True)
 
+    obscure = "ds-;&%b5{}'ΔЙקم๗あ.datc"
     # test cases
     # first 3 args are write_config_section() parameters
     # 4th arg is a list with key/value pairs that should end up in a
@@ -720,6 +722,15 @@ def test_write_config_section(path=None):
             ('short. s p a c e .a123', ' space all over '),
         ]),
     ]
+    if not on_windows:
+        # on windows our test setup lacks the full unicode fancy
+        testcfg.append(
+            ('submodule', obscure, {
+                'path': obscure,
+                'url': f"./{obscure}"}, [
+                (f"submodule.{obscure}.path", obscure),
+                (f"submodule.{obscure}.url", f"./{obscure}"),
+            ]))
 
     for tc in testcfg:
         # using append mode to provoke potential interference by
--- a/datalad/config.py
+++ b/datalad/config.py
@@ -1114,12 +1114,14 @@ def quote_config(v):
       To-be-quoted string
     """
     white = (' ', '\t')
+    comment = ('#', ';')
     # backslashes need to be quoted in any case
     v = v.replace('\\', '\\\\')
     # must not have additional unquoted quotes
     v = v.replace('"', '\\"')
-    if v[0] in white or v[-1] in white:
+    if v[0] in white or v[-1] in white or any(c in v for c in comment):
         # quoting the value due to leading/trailing whitespace
+        # or occurrence of comment char
         v = '"{}"'.format(v)
     return v
 
--- a/datalad/support/tests/test_gitrepo.py
+++ b/datalad/support/tests/test_gitrepo.py
@@ -959,6 +959,16 @@ def test_GitRepo_get_submodules(path=None):
          for s in repo.get_submodules(sorted_=True)],
         ["s_abc", "s_xyz"])
 
+    # Limit by path
+    eq_([s["gitmodule_name"]
+         for s in repo.get_submodules(paths=["s_abc"])],
+        ["s_abc"])
+
+    # Limit by non-existing/non-matching path
+    eq_([s["gitmodule_name"]
+         for s in repo.get_submodules(paths=["s_unknown"])],
+        [])
+
 
 @with_tempfile
 def test_get_submodules_parent_on_unborn_branch(path=None):


mih:
Crawler test failure is unrelated and occurs elsewhere too.

yarikoptic:
> The fix for #6941 is also cherry-picked into a PR against `maint` #6944

#6941 is about circular dependency and #6944 about comment symbols quotation... although could be related - I do not see how. Could you please expain?

datalad/support/gitrepo.py
```diff
@@ -2355,17 +2358,67 @@ def get_submodules_(self, paths=None):
...
+            # a .gitmodules file exists).
+            # without this conditional exit, we would be able to discover
+            # subprojects even when they are not recorded in .gitmodules,
+            # but at the cost of running an unconstrained ls-files call
+            # below
+            return
+
+        # taken from 3.9's pathlib, can be removed once the minimally
+        # supported python version
+        def is_relative_to(self, *other):
```
yarikoptic, line 2372: note: I was curious if we have other uses, grepped to find added in 51e061135a

```
datalad/support/gitrepo.py-                # this could potentially be expensive if lots of files become
datalad/support/gitrepo.py-                # directories, but it is unlikely to happen often
datalad/support/gitrepo.py:                # Note: PurePath.is_relative_to was added in 3.9 and seems slowish
datalad/support/gitrepo.py-                # path_is_subpath faster, also if comparing to "in f.parents"
datalad/support/gitrepo.py-                f_str = str(f)
datalad/support/gitrepo.py-                if any(path_is_subpath(str(f2), f_str) for f2 in status_state['added']):
```

so if speed is of concern, might be worth timing and extracting into a reusable helper.

datalad/support/gitrepo.py
```diff
@@ -2355,17 +2358,67 @@ def get_submodules_(self, paths=None):
...
+            """Return True if the path is relative to another path or False.
+            """
+            try:
+                self.relative_to(*other)
+                return True
+            except ValueError:
```
yarikoptic, line 2378: I wonder if there is a guarantee that ValueError would be raised **only** if not relative, and not for some other (malformatted value etc) cases, thus possibly making it hard to debug.  I hope stdlib has lots of typing annotation and testing to guarantee that. Aforementioned code snippet might provide remedy.

datalad/support/gitrepo.py
```diff
@@ -2355,17 +2358,67 @@ def get_submodules_(self, paths=None):
...
+        if paths:
+            # ease comparison
+            paths = [self.pathobj / p for p in paths]
+            # contrain the report by the given paths
+            modinfo = {
+                # modpath is absolute
+                modpath: modprobs
+                for modpath, modprobs in modinfo.items()
+                # is_relative_to() also match equal paths
+                if any(is_relative_to(modpath, p) for p in paths)
```
yarikoptic, line 2390: Mutating input variables often makes it harder to debug that code later since causes confusion (docstring says `paths` is relative -- how they are absolute?), so I would have preferred:

```suggestion
            abspaths = [self.pathobj / p for p in paths]
            # contrain the report by the given paths
            modinfo = {
                # modpath is absolute
                modpath: modprobs
                for modpath, modprobs in modinfo.items()
                # is_relative_to() also match equal paths
                if any(is_relative_to(modpath, p) for p in abspaths)
```

yarikoptic: note that this is potentially `O(N^2)` operation (becomes expensive) if paths provided for all submodules (e.g smth like `subdatasets sub-*` where each sub- is a subdataset). It reminded me about [get_parent_paths](https://github.com/datalad/datalad/blob/HEAD/datalad/support/path.py#L119) solving similar problem -- please see if it could be used here as well.  If not -- we should be able to come up with at least `O(N*log(N))` solution here

datalad/support/gitrepo.py
```diff
@@ -2355,17 +2358,67 @@ def get_submodules_(self, paths=None):
...
+            mode, gitsha, stage = props.split(' ')
+            if stage not in ('0', '2'):
+                # we either have non-merge situation, or a simple merge
+                # situation (i.e. stage=0). the reported gitsha always
+                # matches what we have locally.
+                # or we are in a three-way merge, in which case stage=2
+                # is what we want to report, because it matches the
+                # current HEAD (see git-read-tree manpage).
+                # there is either a stage 2 or stage 0, never both
+                continue
```
yarikoptic, line 2413: Not familiar much with stages, but it is not clear why we would skip reporting submodule if it is in oms other `stage?  e.g. would it then be present in `stage` `0` and thus still be yielded below? 
is there a test which would demonstrate that may be?

yarikoptic (commented):
Some performance observations/recommendations.

datalad/support/gitrepo.py
yarikoptic: didn't check if that is due to this particular code path but here is comparative timing of master (<.3sec) and this branch (>3 sec):

```shell
$> datalad --version; (builtin cd /home/yoh/datalad/hcp-openaccess; time python -c "from datalad.support.gitrepo import *; from glob import glob; print('\n'.join(map(str,GitRepo('.').get_submodules_(glob('HCP1200/*')))))" | wc -l; )
datalad 0.17.2+102.g2c0073395
1113
python -c   0.26s user 0.04s system 107% cpu 0.277 total
wc -l  0.00s user 0.00s system 0% cpu 0.277 total

$> git co -
Switched to branch 'bf-6940'
Your branch is up to date with 'gh-mih/bf-6940'.

$> datalad --version; (builtin cd /home/yoh/datalad/hcp-openaccess; time python -c "from datalad.support.gitrepo import *; from glob import glob; print('\n'.join(map(str,GitRepo('.').get_submodules_(glob('HCP1200/*')))))" | wc -l; )
datalad 0.17.2+101.gb36148675
1113
python -c   3.78s user 0.02s system 100% cpu 3.777 total
wc -l  0.00s user 0.00s system 0% cpu 3.777 total

```

datalad/support/gitrepo.py
mih: This is the implementation shipped with Python 3.9. What can be done to resolve your concern?

datalad/support/gitrepo.py
mih: Yes, without this conditional. Our merge conflict tests in `update` fail.

datalad/support/gitrepo.py
mih: I am not interested in improving upon this function. I was planing on removing it whenever we reach PY3.9

mih:
> > The fix for #6941 is also cherry-picked into a PR against `maint` #6944
> 
> #6941 is about circular dependency and #6944 about comment symbols quotation... although could be related - I do not see how. Could you please expain?

What is the question? Why it is a BF PR against `maint`? Why it is also here? If the latter: The roundtrip until a patch against `maint` becomes available in `master` is often taking quite a bit.

datalad/support/gitrepo.py
adswa: https://git-scm.com/docs/git-read-tree#_3_way_merge
I read it like this: if stages 1 and 3 are present, they belong to the common ancestor and the remote submodule state - stage 2 is "our" submodule state. Stages 1-3 won't exist if stage 0 exists, and if stage 0 exists, its always the state of "our" submodule.

datalad/support/gitrepo.py
mih: I can replicate this. I have not looked into the reasons, but I will state upfront that "give me all subdatasets" is a more common use case than "give my all subdatasets for the paths of all subdatasets". Even if the slow-down for this use case is unavoidable it does not outweigh the substantial speed-up for the common case.

mih: I can confirm that applying the path-constraint is the source of the slowness. I have no idea how to fix this in a reasonable time frame. Please advice on how to proceed.

The remaining aspect in this conversation is

> docstring says paths is relative -- how they are absolute?

They are not absolute, as the docs say. But they have to become absolute to enable the comparison with the paths reported by `_parse_gitmodules()`.

The path conversion has no significant impact on the timing.

yarikoptic: >  I have no idea how to fix this in a reasonable time frame. Please advice on how to proceed.

I have advised in 

> It reminded me about [get_parent_paths](https://github.com/datalad/datalad/blob/HEAD/datalad/support/path.py#L119) solving similar problem -- please see if it could be used here as well.

I will check in detail if it could be used here now. Otherwise can come up with some other O(N*log(N)) strategies similar to annotate paths we used to do to assign paths into subdatasets paths (here it is assign submodule paths into `paths` -- kinda the same, and we probably already do that with some other functions as well...)

> The remaining aspect in this conversation is
> 
> > docstring says paths is relative -- how they are absolute?
> 
> They are not absolute, as the docs say. But they have to become absolute to enable the comparison with the paths reported by `_parse_gitmodules()`.

So, without the assignment to the dedicated `absname` if I end up at any line after you make them absolute within `paths` - the **are not absolute**, thus contradicting what the docs say and my expectation. that was the only concern. I have no concern on timing impact from such conversion.

mih: Ok, so it read from your statement that this PR is blocked until someone investigated whether this edge case can be made faster to some degree.

yarikoptic: here is a patch which seems to work as desired 
```patch
index a437fc5a2..0776f5805 100644
@@ -59,6 +59,7 @@ from datalad.dataset.gitrepo import (
     path_based_str_repr,
 )
+from datalad.support.path import get_parent_paths
 from datalad.support.due import (
     Doi,
     due,
@@ -2367,32 +2368,14 @@ class GitRepo(CoreGitRepo):
             # below
             return
 
-        # taken from 3.9's pathlib, can be removed once the minimally
-        # supported python version
-        def is_relative_to(self, *other):
-            """Return True if the path is relative to another path or False.
-            """
-            try:
-                self.relative_to(*other)
-                return True
-            except ValueError:
-                return False
-
+        mod_relpaths = [str(m.relative_to(self.pathobj)) for m in modinfo]
         if paths:
-            # ease comparison
-            paths = [self.pathobj / p for p in paths]
-            # contrain the report by the given paths
-            modinfo = {
-                # modpath is absolute
-                modpath: modprobs
-                for modpath, modprobs in modinfo.items()
-                # is_relative_to() also match equal paths
-                if any(is_relative_to(modpath, p) for p in paths)
-            }
+            # constrain the report by the given paths
+            mod_relpaths = get_parent_paths(paths, mod_relpaths, only_with_parents=True)
         for r in self.call_git_items_(
             ['ls-files', '--stage', '-z'],
             sep='\0',
-            files=[str(p.relative_to(self.pathobj)) for p in modinfo.keys()],
+            files=mod_relpaths,
             read_only=True,
             keep_ends=True,
         ):
```
and brings it down to 

```shell
$> datalad --version; (builtin cd /home/yoh/datalad/hcp-openaccess; time python -c "from datalad.support.gitrepo import *; from glob import glob; print('\n'.join(map(str,GitRepo('.').get_submodules_(glob('HCP1200/*')))))" | wc -l ; )
datalad 0.17.2+102.g8b49a71d7.dirty
1113
python -c   0.18s user 0.03s system 110% cpu 0.194 total
wc -l  0.00s user 0.00s system 0% cpu 0.194 total
```

so as well faster than current master and removing the need for borrowed `is_relative` [I questioned above](https://github.com/datalad/datalad/pull/6942/files#r947882244).  Known annotated tests for get_submodules pass

```
$> DATALAD_TESTS_SSH=1 python -m pytest -s -v -k get_submodule datalad
======================================================================= test session starts =======================================================================
platform linux -- Python 3.10.5, pytest-7.1.2, pluggy-1.0.0 -- /home/yoh/proj/datalad/datalad-master/venvs/dev3/bin/python
cachedir: .pytest_cache
rootdir: /home/yoh/proj/datalad/datalad-master, configfile: tox.ini
plugins: fail-slow-0.2.0, cov-3.0.0
collected 1304 items / 1301 deselected / 3 selected                                                                                                               

datalad/support/tests/test_annexrepo.py::test_AnnexRepo_get_submodules SKIPPED (TODO)
datalad/support/tests/test_gitrepo.py::test_GitRepo_get_submodules PASSED
datalad/support/tests/test_gitrepo.py::test_get_submodules_parent_on_unborn_branch PASSED
```

but I feel that we might benefit from more tests here!
```

File: datalad/support/gitrepo.py (hunk 4 above)
Comment (yarikoptic) [reply datalad/support/gitrepo.py:2413]: thank you @mih and @adswa !

File: datalad/tests/test_config.py [hunk 5]
```diff
@@ -50,6 +50,7 @@
 from datalad.utils import (
     Path,
     get_home_envvars,
+    on_windows,
```
yarikoptic, line 53: no longer needed AFAIK
```suggestion
```

datalad/support/gitrepo.py
mih: I think this would not work on Windows. At least the docs of that function say that it needs posix paths. So it seems an explicit conversion to posix is necessary, and in a way that is consistent with what git does for the report from ls-files.

datalad/support/gitrepo.py
yarikoptic: > path_is_subpath() is practically unused outside the tests,

```shell
$> git grep -p 'path_is_subpath(' | grep -v test_
datalad/interface/results.py=def results_from_annex_noinfo(ds, requested_paths, respath_by_status, dir_fail_msg,
datalad/interface/results.py:                if path_is_subpath(fp, p)]
datalad/interface/results.py:                    if path_is_subpath(fp, p)]
datalad/interface/utils.py=def discover_dataset_trace_to_targets(basepath, targetpaths, current_trace,
datalad/interface/utils.py:                       path_is_subpath(t, current_trace[-1]) and
datalad/metadata/aggregate.py=def _get_dsinfo_from_aggmetadata(ds_path, path, recursive, db):
datalad/metadata/aggregate.py:        if path_is_subpath(agginfo_path, seed_ds):
datalad/metadata/aggregate.py:            elif any(path_is_subpath(ep, rp) for ep in exclude):
datalad/metadata/metadata.py=def _get_containingds_from_agginfo(info, rpath):
datalad/metadata/metadata.py:             if path_is_subpath(rpath, subds)],
datalad/metadata/metadata.py=def legacy_query_aggregated_metadata(reporton, ds, aps, recursive=False,
datalad/metadata/metadata.py:                              path_is_subpath(sub, rpath)]
datalad/metadata/metadata.py=class Metadata(Interface):
datalad/metadata/metadata.py:                if parentds and not path_is_subpath(dspath, parentds[-1]):
datalad/support/gitrepo.py=class GitRepo(CoreGitRepo):
datalad/support/gitrepo.py:                if any(path_is_subpath(str(f2), f_str) for f2 in status_state['added']):
datalad/support/parallel.py=def no_parentds_in_futures(futures, path, skip=tuple()):
datalad/support/parallel.py:    return all(not path_is_subpath(path, p) or p in skip for p in futures)
datalad/support/parallel.py=def no_subds_in_futures(futures, path, skip=tuple()):
datalad/support/parallel.py:    return all(not path_is_subpath(p, path) or p in skip for p in futures)
datalad/utils.py=def path_startswith(path, prefix):
datalad/utils.py:def path_is_subpath(path, prefix):
datalad/utils.py=def import_module_from_file(modpath, pkg=None, log=lgr.debug):
datalad/utils.py:            if path_is_subpath(modpath, pkgpath):
```

and I think it is just a matter of fixing the tests, so submitted https://github.com/datalad/datalad/pull/6955 . Let's see how it behaves on windows

datalad/support/gitrepo.py
yarikoptic: concern is minor, but could be resolved through not using it, e.g. as proposed in https://github.com/datalad/datalad/pull/6942/files#r948226105 below

yarikoptic:
> > > The fix for #6941 is also cherry-picked into a PR against `maint` #6944
> > 
> > 
> > #6941 is about circular dependency and #6944 about comment symbols quotation... although could be related - I do not see how. Could you please expain?
> 
> What is the question? Why it is a BF PR against `maint`? Why it is also here? If the latter: The roundtrip until a patch against `maint` becomes available in `master` is often taking quite a bit. If the patch is removed tests will fail, because they exploit #6941

I was not asking about `maint` per se. You said that "The fix for #6941" (where #6941 is about circular dependency) "is also cherry-picked into a PR against `maint`" as "#6944" which is merely about comments parsing.  So I could not understand how comments parsing could resolve circular dependency. That is why was asking. So - is comments parsing resolving circular dependency issue somehow?  or what this sentence was supposed to mean?

yarikoptic:
travis - unrelated stall on is_available - known/not yet resolved. appveyor is all green.

datalad/support/gitrepo.py
yarikoptic: NB note that appveyor tests on windows pass. I feel that there might be a need for a dedicated test here for any type of fix.

indeed -- there is that docstring!  and more confusing:
- at https://github.com/datalad/datalad/blob/master/datalad/support/path.py#L183 we use hardcoded `'/'`
- in `_get_parent_paths_check` at https://github.com/datalad/datalad/blob/master/datalad/support/path.py#L202 we use `os.sep`

and both cases are on `path` of `paths`, and both are added within the same commit where that function was added 5fefda4370e4beac5e5687846e801c9f9f0dadf1 .

Looking at the code, I do not see any "POSIX-specifics" of any kind, so I think that the comment and hardcoded use of `/` *could go* and make POSIX assumption optional (so use on git output could still be used).  I will propose a PR for that because I do feel that currently that function could indeed be buggy. 

proposed patch could do what we do in the other use of that function within `get_content_info` -- `paths = map(ut.PurePosixPath,  paths)` first to ensure that they are POSIX, after all we are to compare to modules paths which should be POSIX right?

datalad/support/gitrepo.py
```diff
@@ -2344,7 +2344,10 @@ def get_submodules_(self, paths=None):
         Parameters
         ----------
         paths : list(pathlib.PurePath), optional
-            Restrict submodules to those under `paths`.
+            Restrict submodules to those under `paths`. Paths must be relative
+            to the resolved repository root, and must be normed to match the
+            reporting done by Git, i.e. no parent dir components
+            (ala "some/../this").
```
yarikoptic, line 2350: I wonder if below for `Returns` we could clarify that paths below for submodules are Posix paths, or they aren't?

datalad/support/gitrepo.py
yarikoptic: submitted https://github.com/datalad/datalad/pull/6963 . While looking at the code realized that above I am pretty much suggesting what `get_content_info` was doing when called from `get_submodules_()` so to some effect we are just moving that logic out. Indeed would need aforementioned `map(ut.PurePosixPath, paths)` to be added to the patch but otherwise -- should be fine.

datalad/support/gitrepo.py
mih: I have no definitive answer. I took this description from the documentation. I did not plan to make changes to the reporting or parameterization.

mih:
@yarikoptic feel free to take over here, you seem to have plans for a deeper dive. I was just here for a quick contribution.

yarikoptic:
> @yarikoptic feel free to take over here, you seem to have plans for a deeper dive. I was just here for a quick contribution.

Not much deeper  than I suggested above . Thanks for the blessing - I will push some commits along those lines later.

yarikoptic:
> > @yarikoptic feel free to take over here, you seem to have plans for a deeper dive. I was just here for a quick contribution.
> 
> Not much deeper than I suggested above . Thanks for the blessing - I will push some commits along those lines later.

submitted #6974 but made a "mistake" of adding a unittest which unearthed a problem needed fixing (`paths=[]` is not `None` but logic implied otherwise) to make my proposed solution  work properly (I checked `maint` -- was ok without it, didn't check this PR but most likely it was ok too).

yarikoptic:
@mih - I ran into need for more work on `get_parent_paths` since its goal is a bit different for matching paths within submodules not limiting some paths  by other paths... needs a bit more work.

so I am ok with you solution, even if later to be improved later. But the tests I added in #6974 fails here and passes on `master`, so we would need to at least fix it up to not introduce regression.

mih:
Closing due to personal overload.

Issue #6943, opened 2022-08-15
Datalad (save) writes invalid .gitmodules with obscure paths


mih:
While debugging https://github.com/datalad/datalad/pull/6942 I found that we would write `.gitmodules` files like this:

```
[submodule "ds-;&%b5{}'ΔЙקم๗あ.datc"]
        path = ds-;&%b5{}'ΔЙקم๗あ.datc
        url = ./ds-;&%b5{}'ΔЙקم๗あ.datc
        datalad-id = c0525452-e64b-44ac-b522-3f380484e8b8
```

Notice that `;` is a comment character. Consequently, reading this file, strips key information. For example, `GitRepo._parse_gitmodules()` essentially runs `git config -l --file .gitmodules`, which yields:

```
submodule.ds-;&%b5{}'ΔЙקم๗あ.datc.path=ds-
submodule.ds-;&%b5{}'ΔЙקم๗あ.datc.url=./ds-
submodule.ds-;&%b5{}'ΔЙקم๗あ.datc.datalad-id=c0525452-e64b-44ac-b522-3f380484e8b8
```

This is likely a problem in any place that writes config-like files without going through individual `git-config` calls.

Here is the essential bit from the docs

```
The # and ; characters begin comments to the end of line, blank lines are
ignored.

Leading whitespaces after name =, the remainder of the line after the
first comment character # or ;, and trailing whitespaces of the line are
discarded unless they are enclosed in double quotes.
```

A correct `.gitmodules` file for the example above would therefore be:

```
[submodule "ds-;&%b5{}'ΔЙקم๗あ.datc"]
        path = "ds-;&%b5{}'ΔЙקم๗あ.datc"
        url = "./ds-;&%b5{}'ΔЙקم๗あ.datc"
        datalad-id = c0525452-e64b-44ac-b522-3f380484e8b8
```

The responsible code is `datalad.config.quote_config()`. Fix is pending.

PR #6944, opened 2022-08-15
Acknowledge git-config comment chars

mih:
Previously we ignored them in values as mandatory reasons for quoting.
This change expands the `quote_config()` helper to perform this
additional test.

Fixes datalad/datalad#6943

### DIFF
--- a/datalad/tests/test_config.py
+++ b/datalad/tests/test_config.py
@@ -704,7 +704,7 @@ def test_write_config_section(path=None):
     # can we handle a bare repo?
     gr = GitRepo(path, create=True, bare=True)
 
-    obscure = "ds-;&%b5{}'ΔЙקم๗あ.datc"
+    obscure = "ds-; &%b5{}# some % "
     # test cases
     # first 3 args are write_config_section() parameters
     # 4th arg is a list with key/value pairs that should end up in a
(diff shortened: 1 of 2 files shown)


mih:
Crawler test failure is unrelated and occurs elsewhere too.

datalad/tests/test_config.py
```diff
@@ -703,6 +704,7 @@ def test_write_config_section(path=None):
     # can we handle a bare repo?
     gr = GitRepo(path, create=True, bare=True)
 
+    obscure = "ds-;&%b5{}'ΔЙקم๗あ.datc"
```
yarikoptic, line 707: might not be supported by the filesystem/locale etc. That is partially why we still carry https://github.com/datalad/datalad/blob/debian/debian/patches/deb_no_utf8 to patch away all such explicit uses. Please use `get_most_obscure_supported_name` or a `OBSCURE_FILENAME` const within desired obscure name.  Also if you use OBSCURE_FILENAME, like `f"ds-;#{OBSCURE_FILENAME}.datc` you likely be able to avoid needing ad-hoc `if on_windows:` condition below.  After all this fix is not about unicode but about `;` and `#` comments.

yarikoptic (changes requested):
Please do not hardcode unicode in the test. IMHO no need for windows conditioning right there.

datalad/tests/test_config.py
mih: Makes sense, will do! Thx!

mih: Done

datalad/tests/test_config.py
```diff
@@ -50,6 +50,7 @@
 from datalad.utils import (
     Path,
     get_home_envvars,
+    on_windows,
```
yarikoptic, line 53: uff, duplicated across PRs commits... 

```suggestion
```
and will accept it to expedite

yarikoptic (approved):

Issue #6950, opened 2022-08-17
Error result generation crashes with --existing=error (default)


mih:
Running a `create-sibling-ria` call twice makes it crash:

```
% datalad -C dist/packages/demo create-sibling-ria -s internal --shared group --group floppy --alias pkg-demo ria+file:///tmp/wt/internal                                                
[ERROR  ] datalad.interface.results.get_status_dict() got multiple values for keyword argument 'ds' (TypeError) 
```

Offending code in `datalad/distributed/create_sibling_ria.py`

```py
352  ->                 res = get_status_dict(
353                         status='error',
354                         message=(
355                             "a sibling %r is already configured in dataset %r",
356                             sname, dpath),
357                         type='sibling',
358                         name=sname,
359                         ds=ds,
360                         **res_kwargs,
361                     )
```

with

```py
296             res_kwargs = dict(
297                 ds=ds,
298                 action="create-sibling-ria",
299                 logger=lgr,
300             )

```

yarikoptic-gitmate:
Issue fixed in [`0.18.0`](https://github.com/datalad/datalad/releases/tag/0.18.0)

PR #6952, opened 2022-08-17
BF: Remove duplicate key in result record

adswa:
Should be a fix and testcase for #6950.

#### 🐛 Bug Fixes
-  When ``create-sibling-ria`` was invoked with a sibling name of a pre-existing sibling, a duplicate key in the result record caused a crashed. This was fixed.


adswa:
argh I can never remember to pick the right base...

yarikoptic-gitmate:

PR #6955, opened 2022-08-17
BF(TST): make tests use _path_ helper for Windows "friendliness" of the tests

yarikoptic:
b91b494b89 simply annotated those tests which were failing on windows to be skipped without
considering to fix them. Those tests were failing because hardcoded POSIX path separator (/).
Quick solution to have tests fixed is to use _path_ helper around those POSIX paths.
The underlying tested code AFAIK should be ok - CI is yet to prove or disprove it.

Closes #2564

PR #6956, opened 2022-08-17
Minor tuneups to please updated codespell

yarikoptic:
Initially lead to linting failure in https://github.com/datalad/datalad/pull/6955 . Mostly concerns with comments and other strings some of which are exposed to the user


mih (approved):
Thx for the fix, these show up everywhere now!

adswa (approved):

Issue #6953, opened 2022-08-17
`create-sibling-ria --post-update-hook` docs not helpful


mih:
Motivation and evidence for this from https://github.com/datalad/datalad/issues/6906#issuecomment-1217876993

```
Enable git's default post-update-hook for the created sibling.
```

Is all there is ATM, and nothing about why one would want to enable that hook, or that it is even more-or-less required in some cases to give a working sibling.

PR #6958, opened 2022-08-18
Explain purpose of create-sibling-ria's --post-update-hook

mih:
Fixes datalad/datalad#6953

### DIFF
--- a/datalad/distributed/create_sibling_ria.py
+++ b/datalad/distributed/create_sibling_ria.py
@@ -191,9 +191,11 @@ class CreateSiblingRia(Interface):
             With `recursive=True`, only the top dataset will be aliased.""",
             constraints=EnsureStr() | EnsureNone()),
         post_update_hook=Parameter(
-            args=("--post-update-hook",),
-            doc="""Enable git's default post-update-hook for the created
-            sibling.""",
+            args=("--post-update-hook", "--update-server-info"),
+            doc="""Enable Git's default post-update-hook for the created
+            sibling. This is useful when the sibling is made accessible via a
+            "dumb server" that requires running 'git update-server-info'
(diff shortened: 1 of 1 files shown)


datalad/distributed/create_sibling_ria.py
```diff
@@ -191,9 +191,11 @@ class CreateSiblingRia(Interface):
             With `recursive=True`, only the top dataset will be aliased.""",
             constraints=EnsureStr() | EnsureNone()),
         post_update_hook=Parameter(
-            args=("--post-update-hook",),
-            doc="""Enable git's default post-update-hook for the created
-            sibling.""",
+            args=("--post-update-hook", "--update-server-info"),
```
yarikoptic, line 194: is adding an alternative --long-option really called for?  IMHO -- no need, and moreover we do not even guarantee that update-server-info is ran within the hook, we just do it once while running this command.

If insisting on keeping -- please adjust the title which might be used for the CHANGELOG entry to reflect that addition in the next .patch release.

Otherwise:
```suggestion
            args=("--post-update-hook",),
```

mih: I don't care much. `--post-update-hook` also does not convey that it *enables* the hook. I don't know what the title needs to be adjusted to, I am fine with any line.

yarikoptic: I could then argue that neither original nor added option reflect entirely the action they are intended to accomplish.  As such I do not see a point of having one more suboptimally named option to be added. I will leave it to someone else to finish review/merge this PR then.

mih:
In the interest of incremental progress, I stripped the option from the diff and edited the top post.

Issue #6948, opened 2022-08-17
`drop` local keys from a `--reckless ephemeral` clone drops from origin


mih:
There is no warning, no confirmation request, it just deletes them -- and even if it is the last copy.

Essentially:

```
datalad clone --reckless ephemeral <url> ds
<anything>
datalad drop --what all -r -d ds
```

wipes out all locally known annex keys from origin.

This is bad!

mih:
A fix for this issue is impacted by the semantic unspecificity of `--reckless ephemeral`. Dropping from a private mode repo is no issue, but dropping from a symlinked annex is.

Until this ambiguity is resolved, it is impossible to simply act according to a set `datalad.clone.reckless=ephemeral` config flag. Instead, this precondition needs to be followed up with an inspection of a linked annex, in combination with a path-url of an existing remote.

Such a sophisticated conjunction of conditions as likely to leave gaps and residual data security issues. Hence I propose to ignore any an all configuration for decision making, and plainly to refuse dropping anything from `here`, whenever `GITDIR/annex` is found to be a symlink of any kind.

PR #6959, opened 2022-08-18
Forbid drop operation from symlink'ed annex (e.g. due to being cloned with --reckless=ephemeral) to prevent data-loss

mih:
This is a blunt, but effective measure to prevent a data security issue,
where dropping annex keys from 'here' is actually also dropping them
from 'origin' -- thereby potentially deleting the last existing copy.

Even once a more elegant error anticipation and handling is implemented,
it should be kept for safety reasons.

Closes datalad/datalad#6948

PS: I will be working on a more elegant wrapping of this fix. But given the severity, this should be merged and released ASAP. https://github.com/datalad/datalad/issues/4750 continues to remain as a reminder.

I came to the conclusion that it might be best to remove the entire feature. I filed https://github.com/datalad/datalad/issues/6961

Spellchecker errors are addressed in https://github.com/datalad/datalad/pull/6956

### DIFF
--- a/datalad/distributed/drop.py
+++ b/datalad/distributed/drop.py
@@ -727,6 +727,9 @@ def _kill_dataset(ds):
 def _drop_allkeys(ds, repo, force=False, jobs=None):
     """
     """
+    assert not (repo.dot_git / 'annex').is_symlink(), \
+        "Dropping from a symlinked annex is unsupported to prevent data-loss"
+
     cmd = ['drop', '--all']
     if force:
         cmd.append('--force')
@@ -757,6 +760,8 @@ def _drop_files(ds, repo, paths, force=False, jobs=None):
     ------
     dict
     """
+    assert not (repo.dot_git / 'annex').is_symlink(), \
+        "Dropping from a symlinked annex is unsupported to prevent data-loss"
     cmd = ['drop']
     if force:
         cmd.append('--force')
--- a/datalad/distributed/tests/test_drop.py
+++ b/datalad/distributed/tests/test_drop.py
@@ -9,6 +9,7 @@
 
 import os
 import os.path as op
+from pathlib import Path
 from unittest.mock import patch
 
 from datalad.api import (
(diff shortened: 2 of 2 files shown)


adswa:
>  But given the severity, this should be merged and released ASAP. 

I agree

yarikoptic:
Crippledfs where no symlinjs supported failure is relevant

datalad/distributed/drop.py
```diff
@@ -727,6 +727,9 @@ def _kill_dataset(ds):
 def _drop_allkeys(ds, repo, force=False, jobs=None):
     """
     """
+    assert not (repo.dot_git / 'annex').is_symlink(), \
+        "Dropping from a symlinked annex is unsupported to prevent data-loss"
```
yarikoptic, line 731: In general I am good with such a stop gap feature (after testing is fixed up on crippled fs).  But also such message could confuse some naive user a little.  We might better rely first on

```
[datalad "clone"]
	reckless = ephemeral
```
being present in repo.config -- and inform user that dropping in ephemeral clone (the term they must have heard about since they had to use it to get here) is not supported.  But indeed checking for symlink is more robust and could be the 2nd line of defense if so desired.

mih: I disagree. Whether or not this config item is still set makes little difference re the outcome -- it remains data loss.

I don't understand what "2nd line of defense if so desired" wants to say.

adswa: If I decode @yarikoptic's comment correctly, its about rewording the message under certain circumstances:
1: Check if the config item about an ephemeral clone
2: If it is present, reword the warning to talk about "ephemeral clone" instead of "symlinked annex"
The symlink check and warning would co-exist and catch a symlinked annex even without the ephemeral config.
Is that interpretation correct?

yarikoptic: For now let's just merge, and if @mih decides to improve UX, could be in a follow up PR.

yarikoptic: I will adjust the title to be little more "obvious" and reflective of the change.

PR #6963, opened 2022-08-18
ENH+BF: get_parent_paths - make / into sep option and consistently use "/" as path separator

yarikoptic:
Code review of this function came up while reviewing the
https://github.com/datalad/datalad/pull/6942 and where get_parent_paths,
which was pretty much "avoided" while resolving infinite recursion of
`get_submodules_ -> get_content_info -> get_submodules_`, was re-introduced
specifically within get_submodules_ .

BF: use consistently "sep" instead of / for check within the loop, and then
os.sep to check if points to up/down folders.

ENH:

- sep became an option so we could use it also with os.sep if so desired
  on paths which are not POSIXed.  By default it would remain "POSIX"
  (although checks were not checking for the POSIX endings before).

- The test became parametric to test for various scenarios of using \ vs /
  vs "the default" (/ - POSIX).

### DIFF
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -2720,22 +2720,23 @@ class GitRepo(CoreGitRepo):
             # any incoming path has to be relative already, so we can simply
             # convert unconditionally
             # note: will be list-ified below
-            paths = map(ut.PurePosixPath,  paths)
+            posix_paths = [ut.PurePath(p).as_posix() for p in paths]
         elif paths is not None:
             return info
+        else:
+            posix_paths = None
 
-        path_strs = list(map(str, paths)) if paths else None
-        if path_strs and (not ref or external_versions["cmd:git"] >= "2.29.0"):
+        if posix_paths and (not ref or external_versions["cmd:git"] >= "2.29.0"):
             # If a path points within a submodule, we need to map it to the
             # containing submodule before feeding it to ls-files or ls-tree.
             #
             # Before Git 2.29.0, ls-tree and ls-files differed in how they
             # reported paths within submodules: ls-files provided no output,
             # and ls-tree listed the submodule. Now they both return no output.
-            submodules = [str(ut.PurePosixPath(s["path"].relative_to(self.pathobj)))
+            submodules = [s["path"].relative_to(self.pathobj).as_posix()
                           for s in self.get_submodules_()]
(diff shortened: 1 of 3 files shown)


yarikoptic:
hm, this
```
..\datalad\support\gitrepo.py:2738: in get_content_info
    path_strs = get_parent_paths(path_strs, submodules)
..\datalad\support\path.py:191: in get_parent_paths
    _get_parent_paths_check(path, sep, asep)
path = 'lvl1\\file1', sep = '/', asep = '\\'
    def _get_parent_paths_check(path, sep, asep):
        """A little helper for get_parent_paths"""
        if isabs(path) or path.startswith(pardir + sep) or path.startswith(curdir + sep):
            raise ValueError("Expected relative within directory paths, got %r" % path)
        if sep+sep in path:
            raise ValueError(f"Expected normalized paths, got {path} containing '{sep+sep}'")
        if asep in path:
>           raise ValueError(f"Expected paths with {sep} as separator, got {path} containing '{asep}'")
E           ValueError: Expected paths with / as separator, got lvl1\file1 containing '\'
..\datalad\support\path.py:217: ValueError
============================== warnings summary ===============================

```
is odd since `paths` should be posix paths there, unless `str`ing a `PurePosixPath` produces native one... I will finally upgrade my windows VM tomorrow ;)  interestingly that it errors `test_copy_file_into_dshierarchy`!

yarikoptic:
Woohoo, it is ready to please windows ppl

mih:
One thing upfront: This change should make the code no worse than it was before. It should be fine to merge if that is the goal.

In general I very much dislike the idea of fixing up any path processing helper in datalad that needs to take `Path` objects, turns them into to strings for processing. The way forward should be to unconditionally replace them with proper `pathlib` use.

We have been hit by countless "there is more to path-processing-than-replacing-slash-with-backslash" issues over the years. Also this code is unlikely to properly handle encoding and other handling of unsupported chars etc.

yarikoptic:
I hear you @mih and would welcome RF/BF which would introduce proper use/manipulation of pathlib objects instead of path strings. It was not the goal of this PR though which was initially to fix inconsistency thus making it work properly on Windows, then due to added extra checks detected additional issues with taken approach and those were fixed.  It indeed should not make code worse from what I see, and it should make DataLad perform better on Windows.

Overall I will take your comment as blessing then and will proceed with the merge.

PR #6965, opened 2022-08-20
BF(TEMP): use git-annex from neurodebian -devel to gain fix for bug detected with datalad-crawler

yarikoptic:
the issue is https://git-annex.branchable.com/bugs/regression__58___fails_to___39__add_.__39___if_file_was_renamed/
and should be addressed in 10.20220724+git144-gc25ade4d9-1~ndall+1, which likely would not
be released for a bit more since would trigger upgrades to v9.  To make CI green again, built
curent snapshot and uploaed to -devel of neurodebian

[skip travis]
[skip appveyor]


yarikoptic:
Crawler is green now, let's proceed!

PR #6971, opened 2022-08-21
Faciliate backports to Debian bullseye

mih:
The `platformdirs` dependency (since 0.16.0) is a pain in the butt to backport to the present Debian stable release. Rather backporting chains of 2nd-level dependencies, this change reverts that move, and brings back `appdirs` on this platform.

This patch setup assume the good 'old `backport-dsc` tool and can be applied with

`backport-dsc -d bullseye -s dl11 datalad_0.17.1-2.dsc`

Where `datalad_0.17.1-2.dsc` is generated with `gbp`

```
gbp buildpackage --git-builder="dpkg-source -b ." --git-postbuild=
```

or whatever floats your boat.

The recorded (and included) run-record captures how the actual patch can be regenerated for future releases of DataLad (rather than having to fiddle with the diff directly).

Otherwise this change is entirely orthogonal to the rest of the Debian packing.

### DIFF
--- a/debian/patches/bullseye-dsc-patch
+++ b/debian/patches/bullseye-dsc-patch
@@ -1,0 +1,13 @@
+diff --git a/debian/control b/debian/control
+index d88d575da..b94e31430 100644
+--- a/debian/control
++++ b/debian/control
+@@ -34,7 +33,7 @@ Build-Depends: debhelper (>= 11),
+                python3-pytest,
+                python3-packaging,
+                python3-pil,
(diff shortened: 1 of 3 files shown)


yarikoptic:
FWIW the only possibly relevant change in platformdirs since appdirs was

```
platformdirs 2.2.0
------------------
- Unix: Fallback to default if XDG environment variable is empty
```

so performance wise indeed should be just as fine at large.  Lets do it, and I might pick it up for few other ubuntu releases

PR #6974, opened 2022-08-22
Reimplement GitRepo.get_submodules_() without get_content_info() and avoiding O(N^2) performance hit

yarikoptic:
This PR is based on work by @mih in https://github.com/datalad/datalad/issues/6941 so it 

Fixes #6940
Fixes #6941

but I took only 1 commit relevant to the stated in the title work, others were already merged in their own #6944 .  While working on the solution and adding a unittest I realized (well -- added test failed ;-) ) that we still have inconsistent handling of the invocation of git commands whenever we DO provide `paths` but the list is empty as if eg "nothing matched".  Awhile back we had already a fix introduced at higher level of get_content_info in https://github.com/datalad/datalad/commit/ec0243c92822f36ada5e87557eb9f5f53929c9ff . In this PR (last commit ATM) I am attempting to fix it in the core -- at the point of invocation of any git command via `_generator_call_git` and also fixing some conditions I did encounter in the code where IMHO it should be `if paths is not None` instead of just `if paths`.

I thought that #6941 was against `maint` and only now realized that it was against `master`.  Indeed the initial change (since code worked ok, just performance was bad) worth `master` and the fix for `paths` can also have "far going manifestations", so I would be ok to rebase it against `master`. But originally submitting as was "planned" against `maint` to see how CI reacts to those "core fixes"

PR #6976, opened 2022-08-24
BF: make logic more consistent for files=[] argument (which is False but not None)

yarikoptic:
Similar problem was resolved previously at higher level in
http://github.com/datalad/datalad/commit/ec0243c92822f36ada5e87557eb9f5f53929c9ff
for get_content_info.  But IMHO it could and should be resolved at the
lowest/common point of invocation, that is why I added it to
_generator_call_git which is underlying most (if not all, hard to tell now ;) )
git invocations.

I also fixed a few other invocations where it was pure `if files` or `if paths`.

This commit was initially proposed as a part of the #6974 but there it failed and later I realized that approach I proposed in that PR for `get_modules_` was flawed, so this PR is dedicated to only this change (although at once across different commands/levels). Want to see if we could distill it independently.

### DIFF
--- a/datalad/dataset/gitrepo.py
+++ b/datalad/dataset/gitrepo.py
@@ -330,6 +330,12 @@ class GitRepo(RepoInterface, metaclass=PathBasedFlyweight):
                 files,
                 protocol=GeneratorStdOutErrCapture,
                 env=env)
+        elif files is not None:
+            # it was an empty structure, so we did provide paths but "empty",
+            # then we must not return anything. For more reasoning see
+            # ec0243c92822f36ada5e87557eb9f5f53929c9ff which added similar code pattern
+            # within get_content_info
+            return
         else:
             generator = self._git_runner.run(
                 cmd,
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -3370,7 +3370,7 @@ class AnnexRepo(GitRepo, RepoInterface):
         else:
             cmd = ['find'] + cmd
             # stringify any pathobjs
-            if paths is not None:
+            if paths:  # we have early exit above in case of [] and not None
                 files = [str(p) for p in paths]
             else:
                 cmd += ['--include', '*']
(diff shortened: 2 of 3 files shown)


yarikoptic:
wow -- "all green" . The only confusing item is a flaky test failure on Travis - confused since I thought that we xfail'ed that test_ExtractedArchive recently in https://github.com/datalad/datalad/pull/6912/files . And code in traceback shows that `try` but not clear how the heck `if 'nfsmount' in fpath:` did not match or ` pytest.xfail(` didn't raise its own exception? I dont' know. I have restarted that job. And will take this PR our of draft since I think it is the right thing to do/fix here although we might run into cases where it would show changed behavior (but may be "for good")

yarikoptic:
eh, last commit adjusting for "early decision making" did introduce regression!

```
        assert_false(
            ds.repo.get_content_annexinfo(paths=[], ref="HEAD", init=None))
        # ... where whatever was passed for init will be returned as is.
>       assert_equal(
            ds.repo.get_content_annexinfo(
                paths=[], ref="HEAD", init={"random": {"entry": "a"}}),
            {"random": {"entry": "a"}})
../datalad/support/tests/test_fileinfo.py:323: 
first = OrderedDict(), second = {'random': {'entry': 'a'}}, msg = None
        if msg is None:
E           AssertionError: assert OrderedDict() == {'random': {'entry': 'a'}}
E             Right contains 1 more item:
E             {'random': {'entry': 'a'}}
E             Full diff:
E             - {'random': {'entry': 'a'}}
E             + OrderedDict()
../datalad/tests/utils_pytest.py:107: AssertionError
```
because there is apparently some extra semantic of `init` for `dict-like`.

```
        init : 'git' or dict-like or None
          If set to 'git' annex content info will amend the output of
          GitRepo.get_content_info(), otherwise the dict-like object
          supplied will receive this information and the present keys will
          limit the report of annex properties. Alternatively, if `None`
          is given, no initialization is done, and no limit is in effect.
```

which is imho likely something to be done outside or with a helper...  Hence I reverted that shortcut return change, leaving only the portion for that "heavily debated" `if paths` to stay as is.

yarikoptic:
appveyor fail is unrelated

```
FAILED ../datalad/support/tests/test_parallel.py::test_gracefull_death - assert 20 >= 23
```
Let's proceed with the release and someone could subsequently push on more of similar unification or adding shortcut returns for some performance benefit.

Issue #6918, opened 2022-08-04
Replace changelog generation with integration with scriv


yarikoptic:
Auto e2big bug keeps striking and now even for relatively small 0.17.3 release. We need a reliable solution. Also @mih and others arn't satisfied with automated collection of changelog entries and would like to be able to curate them before the release. I think we need a solution which would allow for both

datalad-next already used scriv to assemble changelog entries: https://pypi.org/project/scriv/ . But again it is done manually requiring PR author to use scriv locally to master changelog every PR.

I think it would be best to merge both worlds: allow for automation when entries come from PR titles and can be provided custom and could be curated before release.

What is there is GitHub workflow which reacts on PR merges to maint and master and if PR diff lacks changes to changelog.d -  scriv creates a note based on that PR and pushes that change to github. Release workflow will use scriv  collect to actually create changelog for the release. The rest is to be done by auto.

Interested parties could then curate changelog skippers before release to their liking. Lazy overs like me can just rely on automated ones but still capable of inspecting/fixing.

@jwodder do you think this could work?

yarikoptic:
this is in large a satellite for the https://github.com/datalad/datalad/issues/6856 and if it gets resolved with a PR - should close that one as well.

jwodder:
@yarikoptic First off, I feel like mentioning [towncrier](https://pypi.org/project/towncrier/) as a possible alternative to scriv.  The two tools seem to have the same basic behavior, and I'm not sure how they differ, but I'd never heard of scriv before this issue.

> What is there is GitHub workflow which reacts on PR merges to maint and master and if PR diff lacks changes to changelog.d - scriv creates a note based on that PR and pushes that change to github.

Note that there's no dedicated "run on PR merge" event for GitHub Actions; the closest thing is reacting on pushes to the base branch.  Given that, I'm not entirely certain if it's possible to retrieve the PR that was just merged in a GitHub Actions workflow.

Also, if we generate the CHANGELOG entry ourselves instead of using auto, I'm not sure what auto would do about filling in the GitHub release notes.

Also note that the only features of auto that are hard to reproduce ourselves are (a) determining what PRs went into a release (which scriv/towncrier should be able to help with) and (b) determining the version bump level (which, for datalad, corresponds to the branch being released).  If we can do those things without auto, we won't really have any need for auto left.

yarikoptic:
[...]

PR #6981, opened 2022-08-24
Add workflow for autogenerating changelog snippets

jwodder:
Part of #6918.  The workflow won't affect this PR itself, but you can see it in action in a test repository at https://github.com/datalad/temp-release-devel/pull/4.

#### 🏠 Internal
- Add a workflow for autogenerating changelog snippets

[travis skip]
[appveyor skip]

### DIFF
--- a/.github/workflows/add-changelog-snippet.yml
+++ b/.github/workflows/add-changelog-snippet.yml
@@ -26,8 +26,8 @@ jobs:
 
       - name: Configure Git identity
         run: |
-          git config --global user.name "github-actions"
-          git config --global user.email "github-actions@users.noreply.github.com"
+          git config --global user.name "DataLad Bot"
+          git config --global user.email "bot@datalad.org"
 
       - name: Set up Python
         uses: actions/setup-python@v4
--- a/.github/workflows/add-changelog-snippet.yml
+++ b/.github/workflows/add-changelog-snippet.yml
@@ -15,7 +15,7 @@ jobs:
   add:
     runs-on: ubuntu-latest
     if: >
-      github.repository == 'datalad/datalad'
+      github.repository_owner == 'datalad'
         && contains(github.event.pull_request.labels.*.name, 'CHANGELOG-missing')
     steps:
       - name: Check out repository
--- a/.github/workflows/add-changelog-snippet.yml
+++ b/.github/workflows/add-changelog-snippet.yml
@@ -2,7 +2,14 @@ name: Add changelog.d snippet
 
 on:
   pull_request_target:
-    types: [opened, synchronize, reopened, labeled]
+    # "labeled" is also triggered when a PR is created with one or more labels
+    types: [synchronize, labeled]
+
+# Prevent the workflow from running multiple jobs at once when a PR is created
+# with multiple labels:
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref_name }}
+  cancel-in-progress: true
 
 jobs:
   add:
--- a/.github/workflows/add-changelog-snippet.yml
+++ b/.github/workflows/add-changelog-snippet.yml
@@ -1,0 +1,56 @@
+name: Add changelog.d snippet
+
+on:
+  pull_request_target:
+    types: [opened, synchronize, reopened, labeled]
+
+jobs:
+  add:
+    runs-on: ubuntu-latest
+    if: >
+      github.repository == 'datalad/datalad'
+        && contains(github.event.pull_request.labels.*.name, 'CHANGELOG-missing')
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.ref }}
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+
+      - name: Configure Git identity
+        run: |
+          git config --global user.name "github-actions"
+          git config --global user.email "github-actions@users.noreply.github.com"
+
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '^3.7'
+
+      - name: Install Python dependencies
+        run: python -m pip install --upgrade requests
+
+      - name: Generate changelog.d snippet
+        run: |
+          python3 tools/ci/mkchlog-snippet.py \
+            ${{ github.repository_owner }} \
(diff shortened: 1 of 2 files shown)


yarikoptic:
Thanks @jwodder ! shouldn't we get it into `maint` instead of `master`?

jwodder:
@yarikoptic Fixed.

.github/workflows/add-changelog-snippet.yml
```diff
@@ -0,0 +1,63 @@
...
+      - name: Check out repository
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.ref }}
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+
+      - name: Configure Git identity
+        run: |
+          git config --global user.name "github-actions"
+          git config --global user.email "github-actions@users.noreply.github.com"
```
yarikoptic, line 30:
```suggestion
          git config --global user.name "DataLad Bot"
          git config --global user.email "bot@datalad.org"
```

as we already use in `.autorc`.

.github/workflows/add-changelog-snippet.yml
```diff
@@ -0,0 +1,63 @@
...
+# with multiple labels:
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref_name }}
+  cancel-in-progress: true
+
+jobs:
+  add:
+    runs-on: ubuntu-latest
+    if: >
+      github.repository == 'datalad/datalad'
```
yarikoptic, line 18: since this is to prevent running this workflow on forks, and since we are unlikely to fork into datalad organization itself but might decide to adopt this workflow elsewhere,  please make it allow for any repository under `datalad` organization.

tools/ci/mkchlog-snippet.py
```diff
@@ -0,0 +1,180 @@
...
+
+    @property
+    def category(self) -> str:
+        for lbl, cat in SEMVER_LABELS:
+            if lbl in self.labels:
+                return cat
+        sys.exit("Pull request lacks semver labels")
+
+    def as_snippet(self) -> str:
+        item = self.title
```
yarikoptic, line 57:
```suggestion
        item = self.title.strip()
```
to also ensure legit subsequent checks etc.

tools/ci/mkchlog-snippet.py
```diff
@@ -0,0 +1,180 @@
...
+        if not item.endswith((".", "!", "?")):
+            item += "."
+        if self.closed_issues:
+            item += f"  Fixes {', '.join(self.closed_issues)} via {self.url} (by @{self.author})"
```
yarikoptic, line 61: please make closed_issues not just pure URLs but rather `[#NUMBER](https://github.com/datalad/datalad/pull/NUMBER)` and similar to PR not just `self.url` but `[PR #NUMBER](self.url)` and `@{self.author}` into a link - would make changelog rendering more concise and inline with what we have now.

tools/ci/mkchlog-snippet.py
```diff
@@ -0,0 +1,180 @@
...
+        else:
+            item += f"  {self.url} (by @{self.author})"
```
yarikoptic, line 63: same here about PR self.url and author

yarikoptic (changes requested):
Looks great to me, suggested minor stylistic additions.

.github/workflows/add-changelog-snippet.yml
jwodder: Done: f5fe5aa54.

tools/ci/mkchlog-snippet.py
jwodder: Done: dd2824d62.

tools/ci/mkchlog-snippet.py
jwodder: Done.

tools/ci/mkchlog-snippet.py
```diff
@@ -0,0 +1,210 @@
...
+        "repo_name": repo_name,
+        "prnum": prnum,
+        "closing_cursor": None,
+        "label_cursor": None,
+    }
+    closed_issues: list[str] = []
+    labels: set[str] = set()
+    with requests.Session() as s:
+        s.headers["Authorization"] = f"bearer {token}"
+        while True:
```
yarikoptic, line 180: This loop necessity/logic is not clear to me: from reading the code - it loops if there are next pages for issues and/or labels, while overwriting `data` (used to get metadata for the PR itself) on next iteration.  
Does every "page" repeat all the same information about PR (so overloading of `data` with title/number/url would be ok) but with some extra metadata for issues/labels?  if so -- ok, if not -- might need fixup of logic.

In either case (either logic needs to be fixed or not), @jwodder could you please add some comment above the `while True` loop describing why it is needed?

tools/ci/mkchlog-snippet.py
```diff
@@ -0,0 +1,210 @@
...
+        if not item.endswith((".", "!", "?")):
+            item += "."
+        if self.closed_issues:
+            item += (
+                "  Fixes "
+                + ", ".join(i.as_link() for i in self.closed_issues)
+                + f" via {self.as_link()} (by {self.author.as_link()})"
+            )
+        else:
+            item += f"  {self.as_link()} (by {self.author.as_link()})"
```
yarikoptic, line 89: @mih -- overall, are you ok with trying out such an approach in `maint` (and we will merge into master as well)

tools/ci/mkchlog-snippet.py
yarikoptic: Thanks for the explanation and the comment. As long as it works -- fine with me as-is given that we should not get over 50 in any of those "categories" ;)

tools/ci/mkchlog-snippet.py
```diff
@@ -0,0 +1,213 @@
...
```
yarikoptic, line 89: Minor: it seems some minor duplication avoidance with centralized "two spaces away":

```suggestion
        item += "  "
        if self.closed_issues:
            item += (
                "Fixes "
                + ", ".join(i.as_link() for i in self.closed_issues)
                + " via "
            )
        item += f"{self.as_link()} (by {self.author.as_link()})"
```

PR #6983, opened 2022-08-24
RF(BK?): dissolve support for old style Interfaces without @eval_results

yarikoptic:
I saw TODO and "acted". But then at the end I decided to ask grep and found that we have 2 interfaces
which do that still. So it might be a bit early or we could just make those interfaces
also into @eval_results ones. WDYT?

TODOs
- [ ] see if we want to proceed with such change
- [ ] ? make `@eval_results` a no-op and just to become part of any `Interface` (somehow), and remove all of its references in this repo (and eventually in extensions through deprecation cycle)

- Remove explicit support for `Interface` subclasses without `@eval_results`

### DIFF
--- a/datalad/cli/exec.py
+++ b/datalad/cli/exec.py
@@ -22,7 +22,6 @@ from datalad.utils import (
 )
 
 # only imported during command execution
-# .interface._has_eval_results_call
 # from .utils import EnsureKeyChoice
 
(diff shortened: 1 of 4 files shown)


yarikoptic:
not yet I guess, have better things to do ;-)

PR #6915, opened 2022-08-03
Run pytests in parallel (-n 4) on CI (travis ATM)

yarikoptic:
[appveyor skip]

PR #6987, opened 2022-08-26
Run pytests in parallel (-n 2) on appveyor

yarikoptic:
https://github.com/datalad/datalad/pull/6915 has done that for Travis

Should cut tests run time by considerable %

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -346,8 +346,8 @@ test_script:
   - sh: mkdir __testhome__
   - cd __testhome__
     # run test selection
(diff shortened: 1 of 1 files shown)


yarikoptic:
hm -- one run seems stalled and thus was killed at test_download_url_dataset, another one - filed https://github.com/datalad/datalad/issues/6994 - some flaky. restarting incomplete

yarikoptic:
Since it is trivial, if no objections are stated, will merge tomorrow.

PR #6997, opened 2022-08-31
Make the manpages build reproducible via datalad.source.epoch (to be used in Debian packaging)

yarikoptic:
I think it is worth adding to our codebase to avoid posssible conflicts in the
patch and also to facilitate similar reproducibility in other distributions
where so would be desired.

Origin: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1010318

### DIFF
--- a/changelog.d/pr-6997.md
+++ b/changelog.d/pr-6997.md
@@ -1,0 +1,5 @@
+### Internal
+
+- Make the build reproducible (on Debian) via SOURCE_DATE_EPOCH for dates in
+  manpages.  [PR #6997](https://github.com/datalad/datalad/pull/6997) (by
+  [@yarikoptic](https://github.com/yarikoptic))
--- a/changelog.d/pr-6997.md
+++ b/changelog.d/pr-6997.md
@@ -1,5 +1,6 @@
 ### Internal
 
-- Make the build reproducible (on Debian) via SOURCE_DATE_EPOCH for dates in
-  manpages.  [PR #6997](https://github.com/datalad/datalad/pull/6997) (by
+- Introduce datalad.source.epoch configuration to make the build reproducible
+  (on Debian only ATM) by providing date to use in the manpages.
+  [PR #6997](https://github.com/datalad/datalad/pull/6997) (by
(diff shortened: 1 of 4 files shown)


yarikoptic:
hm, where is all the CIs?  I got notification that docs failed to build, so I pushed a fix 6a21dabeed62be22db756323c945b44abff6efed directly to `maint` (I know -- evil but minor change, only docs, and reproduced/tested locally).  I will rebase and force push now this PR to hopefully get CIs back

yarikoptic:
ha ha -- with force-push I "killed" the generated changelog!  will regen.

yarikoptic:
following @bpoldrack feedback: we probably should introduce some dedicated DATALAD_ config var and just set it to the value of SOURCE_DATE_EPOCH in debian/rules... I will RF

yarikoptic:
ok, ready for review again!  Testing added to building the docs worked and showed

```
Run # with custom date
running build_manpage
.TH "datalad" "1" "1973\-03\-03" "datalad 0.17.4+5.g393a702e8"
running build_manpage
.TH "datalad" "1" "2022\-09\-01" "datalad 0.17.4+5.g393a702e8"
```

I highly doubt that other failures (e.g. in `test_datalad_credential_helper` with "AssertionError: assert None == 'dl-user'") relate anyhow.

changelog.d/pr-6997.md
```diff
@@ -0,0 +1,6 @@
+### Internal
+
+- Introduce datalad.source.epoch configuration to make the build reproducible
+  (on Debian only ATM) by providing date to use in the manpages.
+  [PR #6997](https://github.com/datalad/datalad/pull/6997) (by
+  [@yarikoptic](https://github.com/yarikoptic))
```
yarikoptic, line 6: I even took time to tune up this ;)

yarikoptic:
oh, we are plagued by https://github.com/jaraco/keyring/issues/593 !  I will submit a blacklisting PR as I did for dandi-cli https://github.com/dandi/dandi-cli/pull/1112

yarikoptic:
ok, what could go wrong!? ;)

PR #7001, opened 2022-09-01
Do not use `gen4`-metadata methods in `datalad metadata`-command

### LINKED ISSUE #7000: `datalad metadata` fails with `metadata(impossible): <path> (('dataset', 'file')) [Dataset at <path> does not contain gen4 metadata]`
### What is the problem?

If a newer version of datalad-metalad is installed, i.e. >= 0.3.0, the legacy command `datalad metadata` will fail after reporting all existing legacy metadata, if no `gen4`-metadata, i.e. metadata created by `datalad-metalad>=0.3.0`,  is found in the dataset. The error will show a message like this: ` [Dataset at <path> does not contain gen4 metadata]`.

This is an error because the legacy commands should only care about legacy metadata.

```
> pip install datalad>=0.17.0
> pip install datalad-metalad>=0.3.0
> dataset create -c text2git t2
> cd t2
> echo "some content" >file1.txt
> datalad save -m "add content"
> datalad configuration set datalad.metadata.nativetype=datalad_core
> datalad aggregate-metadata
> datalad metadata
```

### DataLad information

The error occurs with datalad version 0.17,0 and higher.
The error only occurs if datalad-metalad >= 0.3.0 is installed as well.

christian-monch:
Fixes #7000 

This PR fixes issue #7000. It restricts legacy metadata code that implents `datalad metadata` to only use legacy metadata query methods.

#### 🐛 Bug Fixes
- Do not process gen4-metadata in `datalad metadata` command ... Fixes #7000

### DIFF
--- a/datalad/metadata/metadata.py
+++ b/datalad/metadata/metadata.py
@@ -1124,13 +1124,19 @@ def gen4_query_aggregated_metadata(reporton: str,
                     }
                 }
         except NoMetadataStoreFound:
+            if len(matching_types) == 2:
+                matching_type = "all"
+            elif len(matching_types) == 0:
+                matching_type = "none"
+            else:
+                matching_type = matching_types[0]
             yield {
                 **kwargs,
                 'path': str(ds.pathobj / relative_path),
                 'status': 'impossible',
                 'message': f'Dataset at {ds.pathobj} does not contain gen4 '
                            f'metadata',
-                'type': matching_types
+                'type': matching_type
             }
 
     return None
--- a/datalad/metadata/metadata.py
+++ b/datalad/metadata/metadata.py
@@ -1124,6 +1124,7 @@ def gen4_query_aggregated_metadata(reporton: str,
                     }
                 }
         except NoMetadataStoreFound:
+            lgr.warning(f"Found no gen4-metadata in dataset {dataset.pathobj}.")
             if len(matching_types) == 2:
                 matching_type = "all"
             elif len(matching_types) == 0:
--- a/datalad/metadata/metadata.py
+++ b/datalad/metadata/metadata.py
@@ -1142,7 +1142,7 @@ def query_aggregated_metadata(reporton: str,
                               recursive: bool = False,
                               metadata_source: Optional[str] = None,
                               **kwargs):
-    """Query legacy and NG-metadata stored in a dataset or its metadata store
+    """Query legacy and gen4-metadata stored in a dataset or its metadata store
 
     Parameters
     ----------
--- a/datalad/metadata/metadata.py
+++ b/datalad/metadata/metadata.py
@@ -347,7 +347,7 @@ def legacy_query_aggregated_metadata(reporton, ds, aps, recursive=False,
 def _query_aggregated_metadata_singlepath(
         ds, agginfos, agg_base_path, qap, reporton, cache, dsmeta,
         contentinfo_objloc):
-    """This is the workhorse of query_aggregated_metadata() for querying for a
+    """This is the workhorse of legacy_query_aggregated_metadata() for querying for a
     single path"""
     rpath = qap['rpath']
     containing_ds = qap['metaprovider']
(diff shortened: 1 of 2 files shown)


yarikoptic:
eh, I wish you have finished this PR @christian-monch ... Trying to make datalad-neuroimaging sane again I spent time solving this issue , and differently and IMHO suboptimally (I just was disabling `status="impossible"` bu default) :-/   Are you planing to finish this PR?

yarikoptic:
FTR: checked -- actually it is not solving it for neuroimaging . Since it is done here at the level of `metadata` command and not `query_aggregate_metadata` where IMHO it should be done -- tests in neuroimaging would still get the "impossible" record.

christian-monch:
@yarikoptic sorry for the delay. Will finish it now

christian-monch:
@yarikoptic  I added a test. looking into moving it into  `query_aggregate_metadata` now

christian-monch:
> FTR: checked -- actually it is not solving it for neuroimaging . Since it is done here at the level of `metadata` command and not `query_aggregate_metadata` where IMHO it should be done -- tests in neuroimaging would still get the "impossible" record.

But it is actually done in `query_aggregated_metadata`, more precisely in `gen4_query_aggregated_metadata`, which is called by `query_aggregated_metadata`. What did I get wrong?

christian-monch:
> FTR: checked -- actually it is not solving it for neuroimaging . Since it is done here at the level of `metadata` command and not `query_aggregate_metadata` where IMHO it should be done -- tests in neuroimaging would still get the "impossible" record.

I think the change should remain at the level of `metadata` because `metadata` is meant to be used with "legacy" metadata. When it was introduced gen4 metadata was not around and I think the limitation to "legacy" metadata should remain to keep the initial semantics.

Having said that, the default for the `metadata_source` parameter in `query_aggregated_metadata` should probably change. Since `query_aggregated_metadata` receives the metadata stores that it should search explicitly, It should yield an impossible-result when gen4-metadata is defined as `metadata_source` but does not exist. The thing is that the current default for `metadata_source` specifies querying both, "legacy" and "gen4" metadata. That changes the semantics of a call to `query_aggregated_metadata` that does not provide the `metadata_source` argument. I will change the default to only search in "legacy" metadata. After this change, existing code that does not specify the new argument `metadata_source`should behave properly.

changelog.d/pr-7001.md
```diff
@@ -0,0 +1,5 @@
+### Internal
+
+- Do not use `gen4`-metadata methods in `datalad metadata`-command.  [PR
+  #7001](https://github.com/datalad/datalad/pull/7001) (by
+  [@christian-monch](https://github.com/christian-monch))
```
yarikoptic, line 5: dewarp somewhat - fix is coming

```suggestion
- Do not use `gen4`-metadata methods in `datalad metadata`-command.
  [PR #7001](https://github.com/datalad/datalad/pull/7001) (by
  [@christian-monch](https://github.com/christian-monch))
```

datalad/metadata/search.py
```diff
@@ -1316,14 +1316,15 @@ class Search(Interface):
             for debugging purposes."""),
         metadata_source=Parameter(
             args=('--metadata-source',),
-            choices=('legacy', 'gen4'),
-            default=None,
+            choices=('legacy', 'gen4', 'all'),
+            default='legacy',
```
yarikoptic, line 1320: I believe duplication is not needed since we have it in the function signature
```suggestion
```

yarikoptic (approved):
In this PR you are suggesting to revert back `query_aggregated_metadata` to 
- replace `None` with `"all"`
- make "legacy" to be default for that function **and** `search`
As a result, `query_aggregated_metadata`, `metadata` and `search` would now not error out while encountering datasets only with legacy metadata.  Since it "fits my use-case" of immediate need to test datalad-neuroimaging, which I checked would pass now, I marked it as "approve".

**But** ideally I think it should have been more of a "any" instead of "all" which was added, and then sensing for any (legacy or gen4) was done or if indeed all are impossible, yield an impossible rec?

PS I will accept my suggestion on fixing warping of changelog entry to expedite

yarikoptic-gitmate:

PR #7002, opened 2022-09-01
Fix test helpers for results

bpoldrack:
The helpers for testing result records had a couple of bugs.
Most noteably, some (e.g. `assert_status`) would pass when no result was
given at all. Therefore command calls that did not return any result
were passing assertions about the `status` field of their results.

Furthermore, most of those helpers did not properly deal with all three
"representations" of results that are (and need to be) passed to them:
1. list of dict
2. single dict
3. generator

That's because `ensure_list` is used with the incoming results, but this
turns dicts that are not encapsulated by a list into a list of their
keys rather than a single-item list of dict. Reason for that is, that a
dict is iterable.

The falsely passing helpers also hid bugs in the tests - mostly command calls (`clone`) failing to set `result_xfm` correctly and therefore not getting result records back, but then proceeding to assertions about such result records.

This patch should fix this.

### DIFF
--- a/changelog.d/pr-7002.md
+++ b/changelog.d/pr-7002.md
@@ -1,0 +1,4 @@
+### Tests
+
+- Fix broken test helpers for result record testing that would falsely pass.
+  [PR #7002](https://github.com/datalad/datalad/pull/7002) (by [@bpoldrack](https://github.com/bpoldrack))
--- a/datalad/tests/utils_pytest.py
+++ b/datalad/tests/utils_pytest.py
@@ -1476,7 +1476,10 @@ def assert_status(label, results):
     in this sequence.
     """
     label = ensure_list(label)
-    results = ensure_list(results)
+    results = ensure_result_list(results)
+    if len(results) == 0:
+        # If there are no results, an assertion about all results must fail.
+        raise AssertionError("No results retrieved")
     for i, r in enumerate(results):
         try:
             assert_in('status', r)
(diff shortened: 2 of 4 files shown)


mih:
This change breaks a lot. Is there an issue that explains what is happening here, or why it needs doing? Or can this be closed?

changelog.d/pr-7002.md
```diff
@@ -0,0 +1,4 @@
+### Tests
+
+- Fix broken test helpers for result record testing that would falsely pass.
+  https://github.com/datalad/datalad/pull/7002 (by @bpoldrack)
```
yarikoptic, line 4: I think we should tune up our template ... filed #7084 . up to you to fix it up or not here.

yarikoptic:
why `master` and not `maint` - sounds like an important Fix to have in `maint`!

yarikoptic:
nice find -- thanks @bpoldrack  ! IMHO should go against  `maint` and ideally just get `ensure_results` helper to centralize "casting" ;)

changelog.d/pr-7002.md
bpoldrack: You mean like this: https://github.com/datalad/datalad/pull/7002/commits/4ce17b19d7a317d215c59711aedc78658bdcfb4e ?

datalad/tests/utils_pytest.py
```diff
@@ -1476,7 +1476,10 @@ def assert_status(label, results):
     in this sequence.
     """
     label = ensure_list(label)
-    results = ensure_list(results)
+    results = [results] if isinstance(results, dict) else ensure_list(results)
```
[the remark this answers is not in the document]
bpoldrack: Yes, done.

bpoldrack:
Thx, @yarikoptic. All addressed, I think.

yarikoptic-gitmate:

PR #7003, opened 2022-09-01
BF: blacklist 23.9.0 of keyring as introduces regression

yarikoptic:
Ref: https://github.com/jaraco/keyring/issues/593

We better release asap to not cause the chain reaction through downstreams.  May be without yet "scriv-edited" changelogs since not fully implemented so I will not bother adding one here either.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -40,7 +40,8 @@ requires = {
     ],
     'downloaders': [
         'boto',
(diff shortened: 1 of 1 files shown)


yarikoptic:
so -- containers do have their own new issue(s)... eh. still waiting on travis/appveyor

PR #7004, opened 2022-09-02
ENH(TEMP): set git core.quotePath to false in tests fixture

yarikoptic:
That causes paths "displayed" properly to the users, and I wonder if any
of our tests/interfaces which use utf-8 would be succeptable.  Might be worth
adding this config setting to one of the CI runs, but for that we would need
to make it optional, and not always false like here

### DIFF
--- a/datalad/conftest.py
+++ b/datalad/conftest.py
@@ -82,6 +82,10 @@ def setup_package():
         email = test@example.com
 [core]
 	askPass =
+	# "display" unicode filenames in status etc.
+	# discovered at https://stackoverflow.com/questions/34549040/git-not-displaying-unicode-file-names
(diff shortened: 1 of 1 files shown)


yarikoptic:
hm, there was some keyring related gotcha which forbidden a clean run.  I will push rebase to trigger a fresh run but it seems that it might all just work.  Might be worth some kind of a "hint" to the user.

yarikoptic:
forget about it for now

yarikoptic:
FWIW -- relevant recent change in git-annex: [10.20230407-6-gd689a5b338](https://git.kitenet.net/index.cgi/git-annex.git/commit/?id=d689a5b338c781a9994ee96e3d0511d303fa9a27)

PR #7006, opened 2022-09-02
Fix type annotations in `cmd.py` and `runner.py`

christian-monch:
This PR os based on the work and discussions in PR #6899 

#### 🏠 Internal
- Add type annotations to `cmd.py` and `runner.py`

### DIFF
--- a/datalad/runner/runner.py
+++ b/datalad/runner/runner.py
@@ -126,13 +126,14 @@ class WitlessRunner(object):
 
         Returns
         -------
-          Union[Any, Generator]
+          dict | _ResultGenerator
 
             If the protocol is not a subclass of `GeneratorMixIn`, the
             result of protocol._prepare_result will be returned.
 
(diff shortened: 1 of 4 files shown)


datalad/runner/runner.py
```diff
@@ -191,7 +196,7 @@ def run(self,
         if issubclass(protocol, GeneratorMixIn):
             return results_or_iterator
         else:
-            results = results_or_iterator
+            results = cast(dict, results_or_iterator)
```
yarikoptic, line 199: For myself a note that according to [docs](https://docs.python.org/3/library/typing.html#typing.cast) - "This returns the value unchanged." so should have no effect at runtime. good.

yarikoptic:
great, thanks. appeveyor is a known `FAILED ../datalad/support/tests/test_parallel.py::test_gracefull_death - assert [0, 2] == [0, 1]
` ... let's proceed

PR #7009, opened 2022-09-02
Use scriv for CHANGELOG generation in release workflow

jwodder:
Part of #6918.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -28,7 +28,7 @@ jobs:
           python-version: '^3.8'
 
       - name: Install Python dependencies
-        run: python -m pip install build twine
+        run: python -m pip install build scriv twine
 
       - name: Install pandoc
         run: sudo apt-get update && sudo apt-get install -y pandoc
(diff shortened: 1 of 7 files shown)


.github/workflows/release.yml
```diff
@@ -28,34 +28,29 @@ jobs:
...
-          elif ! dpkg --compare-versions "$new_version" gt "$changelog_version"
-          then echo "[ERROR] Could not compare new version '$new_version' to changelog version '$changelog_version'"
+          if [ "$auto_version" != "$changelog_version" ]
+          then echo "[ERROR] Auto version does not equal changelog version"
                exit 1
           fi
-          ~/auto shipit -vv $opts
+          scriv collect --add --version "$changelog_version"
+          git commit -m '[skip ci] Update CHANGELOG' CHANGELOG.md
+          ~/auto shipit -vv --no-changelog
```
yarikoptic, line 53: if `--no-changelog` -- would it trigger 
```
.autorc:                "afterChangelog": "make update-changelog && git add docs/source/changelog.rst && git commit -m '[skip ci] Update RST changelog'",
```
?  may be better to just move that call right here before `shipit`?

changelog.d/templates/entry_title.md.j2
```diff
@@ -0,0 +1 @@
+{{ version if version else "VERSION" }} ({{ date.strftime('%Y-%m-%d') }})
```
yarikoptic, line 1: when would that `"VERSION"` happen? should we just error may be to not end up with some "broken" changelog?

yarikoptic (commented):
A few minor questions/comments. Otherwise -- looks worth trying! Thank you!

changelog.d/templates/entry_title.md.j2
jwodder: With this setup, I don't believe it'll happen; I just copied the template from datalad-next.

.github/workflows/release.yml
jwodder: Done.

yarikoptic:
ok, and with this we will test if it all works -- added a `release` label. Thank you @jwodder

yarikoptic-gitmate:
PR released in [`ignore-me-testing`](https://github.com/datalad/datalad/releases/tag/ignore-me-testing)

yarikoptic-gitmate:

PR #7011, opened 2022-09-02
UX: push - provide specific error with details if push failed due to permission issue

yarikoptic:
I kept getting an error like

	$> datalad push
	publish(impossible): . (dataset) [No push target given, and none could be auto-detected, please specify via --to]
	action summary:
	  publish (impossible: 1)

and kept tuning the .git/config to specify extra options like push-remote etc,
but nothing helped.  Only when looking at -l debug I saw that the issue has
nothing to do with default target not being given but rather with permissions
issue.  IMHO "none could be auto-detected" is although technically true that
datalad FAILS to auto-detect it, is semantically wrong since it could be
autodetected but datalad can't.  That is why I decided that in cases whenever
it is a matter of permission issue, we can report that error in full.

edit: details on how such errors were triggered were added to commit message. ATM it is https://github.com/datalad/datalad/commit/9647b4906a02b4374599e700b0e0f86ead56455f

### DIFF
--- a/datalad/core/distributed/push.py
+++ b/datalad/core/distributed/push.py
@@ -443,14 +443,27 @@ def _push(dspath, content, target, data, force, jobs, res_kwargs, pbars,
     target = None
     if not _target:
         try:
-            # let Git figure out what needs doing
-            # we will reuse the result further down again, so nothing is wasted
-            wannabe_gitpush = repo.push(remote=None, git_options=['--dry-run'])
-            # we did not get an explicit push target, get it from Git
-            target = set(p.get('remote', None) for p in wannabe_gitpush)
-            # handle case where a pushinfo record did not have a 'remote'
-            # property -- should not happen, but be robust
-            target.discard(None)
+            try:
+                # let Git figure out what needs doing
+                # we will reuse the result further down again, so nothing is wasted
+                wannabe_gitpush = repo.push(remote=None, git_options=['--dry-run'])
+                # we did not get an explicit push target, get it from Git
+                target = set(p.get('remote', None) for p in wannabe_gitpush)
+                # handle case where a pushinfo record did not have a 'remote'
+                # property -- should not happen, but be robust
+                target.discard(None)
+            except CommandError as e:
+                if 'Please make sure you have the correct access rights' in e.stderr:
+                    # there is a default push target but we have no permission
+                    yield dict(
+                        res_kwargs,
+                        status='impossible',
+                        message='Attempt to push to default target resulted in following '
+                                'error.  Address the error or specify different target with --to: '
+                                + e.stderr,
+                    )
+                    return
+                raise
         except Exception as e:
             lgr.debug(
                 'Dry-run push to determine default push target failed, '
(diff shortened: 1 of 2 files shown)


mih:
A total of four different CI runs across two systems failed (plus the known container extension test run). I restarted them to get a sense of what is real here.

datalad/core/distributed/push.py
```diff
@@ -443,14 +443,27 @@ def _push(dspath, content, target, data, force, jobs, res_kwargs, pbars,
...
+                # let Git figure out what needs doing
+                # we will reuse the result further down again, so nothing is wasted
+                wannabe_gitpush = repo.push(remote=None, git_options=['--dry-run'])
+                # we did not get an explicit push target, get it from Git
+                target = set(p.get('remote', None) for p in wannabe_gitpush)
+                # handle case where a pushinfo record did not have a 'remote'
+                # property -- should not happen, but be robust
+                target.discard(None)
+            except CommandError as e:
+                if 'Please make sure you have the correct access rights' in e.stderr:
```
mih, line 456: It would be good for posterity to have at least some of these errors on record. In the commit message, here in the comment, doesn't matter. Personally, I have not seen such error, and based on the information given here, I have no idea what permissions where problematic. We know that these match-pieces-of-a-message approaches to error handling are fragile. Given that no test is added, we will never know whether this construct is still valid, and have to carry it forever, even when Git changes its behavior.

mih (changes requested):
It makes sense to catch such an error. But I have no idea how to trigger them.

datalad/core/distributed/push.py
yarikoptic: damn right you are! not sure why I have not pasted it in its entirety.  Now I have tried to reproduce by moving my ssh key aside, generating a new one and trying to push and got into another, very related but behaving entirely differently use case -- and apparently either we do not catch `stdout` there or it just goes directly to PTY somehow but we do not get it in stdout/stderr:

```
[DEBUG  ] Opening MultiplexSSHConnection(ctrl_path=<<PosixPath('/ho++36 chars++9b')>>, sshri=<<SSHRI(hostname++26 chars++it')>>) by calling ['ssh', '-fN', '-o', 'ControlMaster=auto', '-o', 'ControlPersist=15m', '-o', 'ControlPath=/home/yoh/.cache/datalad/sock
ets/1f30269b', 'git@github.com'] 
Push:   0%|                                                                                                                                                                                                                        | 0.00/4.00 [00:00<?, ? Steps/s]
git@github.com: Permission denied (publickey).
> /home/yoh/proj/datalad/datalad-maint/datalad/core/distributed/push.py(457)_push()
-> if 'Please make sure you have the correct access rights' in e.stderr:
(Pdb) p e.stderr
None
*(Pdb) p e.stdout
None
```

but it does go to stderr:
```
$> ssh git@github.com 
git@github.com: Permission denied (publickey).

$> ssh git@github.com 2>/dev/null

$>
```
so it started to question my whole fix!  But I reproduced it and adjusted commit message with all the gory details: 9647b4906a02b4374599e700b0e0f86ead56455f

mih:
Thanks for adding the additional info.

yarikoptic-gitmate:

Issue #6961, opened 2022-08-18
Remove `clone --reckless ephemeral` annex-symlinking feature


mih:
It serves a niche use case, but comes with a slew of (accumulating) unaddressed issues:

- https://github.com/datalad/datalad/issues/4144
- https://github.com/datalad/datalad/issues/4744
- https://github.com/datalad/datalad/issues/4750
- https://github.com/datalad/datalad/issues/6946
- https://github.com/datalad/datalad/issues/6948
- https://github.com/datalad/datalad/issues/6949
- https://github.com/datalad/datalad/issues/6960

It might be time to remove it, and limit `ephemeral` to be a private mode annex repo.

yarikoptic:
FWIW AFAIK insofar `clone --reckless ephemeral` seems to be the only generic (on a non-crippled filesystem; and not relying on CoW of e.g. BTRFS) way for quick "experimentation" throw away clones.  I do appreciate the fact that it is a source of many pains, but most seems to be relating to making such mode "read-only" (i.e. disabling `save`, `drop`, `remove` and possibly even `get`) on such clones, thus making them "efficient" only for `git` operations while reusing the same `git-annex` pool of objects.

yarikoptic:
Given @bpinsard I see some variations of ephemeral mode to fulfill his usecase:

- I think it is valid desire to "just disable git-annex" for some modes of operation in such clones. If only has unlocked git-annex files, their content would be available through symlinks and everything would work nicely. So might want to be `--reckless no-annex` mode where we prevent detecting that repo as AnnexRepo. Not sure those if some other functionality would not get broken.
- ATM `ephemeral` mode symlinks entire `.git/annex`, and that is what likely causes then trouble for @bpinsard . But since we care largely about making `.git/annex/objects` available across them all, may be we should just symlink that directory only instead e.g. as `--reckless=ephemeral-objects` mode? sure thing the same user permissions issues would apply if someone decides to `annex get` but in this mode the other `annex info` etc would work since `.git/annex/` would be local to that repo, and "git annex unlocked" files would probably work ok as well.

Issue #6408, opened 2022-02-03
`clone/create` could/should write `datalad.dataset.id` into local git config


mih:
This is a critical piece of information. Right now, it only sits in a branch config. More advanced branch manipulations might result in a branch that contains no `.datalad/config` file. However, I'd argue that even in this case, we would not want to stop treating such a repository as a DataLad dataset.

Possibly there are other types of information that should be treated in the same way, but the presence of a dataset ID is certainly used as a critical test.

Issue #5289, opened 2021-01-07
"python-space" installable hooks 


yarikoptic:
We have briefly touched on that topic this Tue, as a possible way forward to resolve https://github.com/datalad/datalad/issues/5256 more "flexibly" and avoid adding too many "domain"/"solution" specific code into datalad core.  The idea was, analogously to hooks, to provide at Python level entry points "namespace" so extensions could either register additional actions to be performed, or even to provide datalad core with additional "choices" (e.g. installation urls etc).

Adding such facility to datalad core would facilitate

- adding "registry" support for registering datasets: https://github.com/datalad/datalad-registry/issues/7

The difficulty/problem: IMHO it adds an additional "API" on top of the already "fluid" API where would need to regulate such entrypoints hooks -- on what args they would be provided and in some cases what they are supposed to return.  But I guess it would not be too radically different from what is already done in hooks.

mih:
https://github.com/datalad/datalad/issues/4903#issuecomment-1076004289 is an idea that is very appealing to me right now.

Issue #5094, opened 2020-10-26
Autoenable gin sibling when cloning from somewhere else


mikapfl:
#### What is the problem?

Publishing a dataset to gin is already quite easy; however, other places like github or community-specific git hosting services are often more visible places. Therefore, it would be great if datalad datasets could be published to gin (including annex'ed data) and somewhere else like github (sans annex'ed data) such that if the dataset is `datalad clone`d from the github repo, the gin sibling is automatically configured and data can be automatically fetched using `datalad get`.

#### Is there anything else that would be useful to know in this context?

Specifically, I tried this here: https://gin.g-node.org/mikapfl/read_di_unfccc/ and here: https://github.com/mikapfl/read_di_unfccc ; I ended up doing the "coupling" with a README for now.

yarikoptic:
rright -- I guess we need to interface functionality of `create-sibling --as-common-datasrc` also somewhere within `siblings` to "mint" a `type=git autoenable=true` special remote for an existing git remote accessible via http*.

Meanwhile I think something like this should work, if you do in your original/clone of the first repo:

```
git annex initremote gin type=git location=https://gin.g-node.org/mikapfl/read_di_unfccc/ autoenable=true
datalad push
```

mikapfl:
Thanks for the suggestion! I had to expand it a bit to get it work, but this did the trick:

```bash
$ git remote add gin-http https://gin.g-node.org/mikapfl/read_di_unfccc/
$ git fetch gin-http
From https://gin.g-node.org/mikapfl/read_di_unfccc
 * [new branch]      git-annex  -> gin-http/git-annex
 * [new branch]      main       -> gin-http/main
$ git annex initremote gin-http-annex type=git location=https://gin.g-node.org/mikapfl/read_di_unfccc/ autoenable=true
initremote gin-http-annex ok
(recording state in git...)
$ datalad push --to github
publish(ok): /home/pflueger/work/read_di_unfccc (dataset) [refs/heads/git-annex->gin:refs/heads/git-annex cf23fc4..3adb102]                                                                       
publish(ok): /home/pflueger/work/read_di_unfccc (dataset) [refs/heads/git-annex->github:refs/heads/git-annex cf23fc4..3adb102]                                                                    
```
Now a simple `datalad clone` of the github repo directly can `datalad get` the data. Sweet!

mih:
[...]

PR #7017, opened 2022-09-07
Make clone_dataset() patchable and less monolithic

mih:
This change is equivalent to the commuluative effort posted at https://github.com/datalad/datalad-next/pull/99. Its description is included below.

In addition to this original changeset:

- RIA-related helper are moved into `clone_ria.py` too
- internal patches `ria` and `ephemeral` are applied via functions that can be easily patched-out, in order to turn the patches off externally -- for example when an extension starts to provide a superior replacement.

The (adjusted for filenames) description of the original PR follows:

The series provides a patch to replace `clone_dataset()` the key workhorse of datalad-core's `clone` command. The patch causes no change to any functionality, but refactors the code from large monolithic functions into significantly smaller, more comprehensible units. Emphasis is made on separating function that control the flow of processing (`clone.py`) from utilities that perform focused tasks (`clone_utils.py`).

Key feature of the refactoring is the provisioning of a set of dedicated function that implement particular stages of the clone and post-clone setup processing:

- `_try_clone_candidates()` (the main clone attempt loop)
- `_try_clone_candidate()` (handle a single clone attempt)
- `_try_git_clone_candidate()` (only workhorse ATM with all git-clone specifics)
- `_post_gitclone_processing_()` (implements the logic to call all others)
- `_post_git_init_processing_()`
- `_pre_annex_init_processing_()`
- `_annex_init()`
- `_post_annex_init_processing_()`
- `_pre_final_processing_()`

All functions implement a keyword-argument-only API. All functions ending with the name `processing_` are implemented as generators that yield standard DataLad result records.

Any of these functions can be "patched" following the pattern used in `datalad-next` for any kind of modifications of datalad-core. This capability is demonstrated by pulling out two optional features, originally intermingled with essential code:

- `clone_ria.py`
- `clone_ephemeral.py` (related: https://github.com/datalad/datalad/issues/6961)

(Note that for both patches, the documentation is not patched right now, because it seems unnecessary, and can be left to when the patch is turned off externally)

These document how it is possible, with this new implementation, to provide such add-on features externally, without having to introduce them in datalad-core directly. They also show, how the patches can benefit from the keyword-argument-only API. They should continue to work as designed, as long as the particular keyword-arguments the require continue to be provided with the same semantics.

Taken together, this series provides a solution to a number of issues that are linked below with a brief comment:

- https://github.com/datalad/datalad/issues/4903 -- more-or-less arbitrary modifications can be implemented following `git-clone`. The latter is itself encapsulated in the `_try_clone_candidate()` utility, so it would even be possible to patch-in means to establish repository clones that do not involve calling `git-clone`

- This is related to https://github.com/datalad/datalad-registry/issues/7 -- while we already have flexibility to add handling of new URL-types (e.g., URL mapping, subdataset candidate URL configurations), now, either `_try_clone_candidates()` or another utility (`_generate_candidate_clone_sources()`) can be patched to query an external resolver of some kind, including a DataLad registry

- https://github.com/datalad/datalad/issues/5289 -- the datalad-next style patching has proven to be somewhat reliable, and continue to grow in usage. This series modifies datalad-core to enable more surgical patching for add-on functionality. Continuing this for other parts of the code base would lessen the need for (possibly expensive) hooks that may require additional setup and maintenance.

The patch-in-a-clone-feature approach should be applicable to the growling list of open issues asking for additional (special interest) capabilities. Examples:

- https://github.com/datalad/datalad/issues/6408
- https://github.com/datalad/datalad/issues/5094
- https://github.com/datalad/datalad/issues/4142

Lastly, this approach can help to expedite work on rejuvenating datalad components that struggle to keep up within the corset of datalad-core, with its ambition to remain as stable as possible. One example is the RIA functionality that alone is associated with 7% of the open issues, and would benefit from a few faster, less constrained development cycles. The previous impossibility of `clone` to be customized externally (like done here), was one of the main reasons for having to introduce RIA in datalad-core directly, rather than having it prove itself in a dedicated extension.

### DIFF
--- a/datalad/core/distributed/clone.py
+++ b/datalad/core/distributed/clone.py
@@ -9,56 +9,54 @@
 """Plumbing command for dataset installation"""
 
 
-from argparse import REMAINDER
 import logging
-from typing import (
-    Dict,
-)
+from argparse import REMAINDER
+from typing import Dict
 
+from datalad.cmd import CommandError
 from datalad.config import ConfigManager
-from datalad.interface.base import Interface
-from datalad.interface.utils import eval_results
-from datalad.interface.base import build_doc
-from datalad.interface.results import get_status_dict
+from datalad.distribution.dataset import (
+    Dataset,
+    EnsureDataset,
+    datasetmethod,
+    require_dataset,
+    resolve_path,
+)
+from datalad.interface.base import (
+    Interface,
+    build_doc,
+)
 from datalad.interface.common_opts import (
     location_description,
     reckless_opt,
 )
-from datalad.cmd import CommandError
+from datalad.interface.results import get_status_dict
+from datalad.interface.utils import eval_results
 from datalad.support.annexrepo import AnnexRepo
 from datalad.support.constraints import (
+    EnsureKeyChoice,
     EnsureNone,
     EnsureStr,
-    EnsureKeyChoice,
 )
 from datalad.support.exceptions import CapturedException
-from datalad.support.param import Parameter
 from datalad.support.network import (
-    PathRI,
     RI,
+    PathRI,
 )
+from datalad.support.param import Parameter
 from datalad.utils import (
     knows_annex,
     rmtree,
 )
 
-from datalad.distribution.dataset import (
-    Dataset,
-    datasetmethod,
-    resolve_path,
-    require_dataset,
-    EnsureDataset,
-)
-
-from .clone_utils import (
-    # needed because other code imports it from here
-    decode_source_spec,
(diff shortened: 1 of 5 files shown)


mih:
Doc build failure is fixed in https://github.com/datalad/datalad/pull/7018

mih:
Container extension test failures a
look same as in #7018 

I think this is ready to go

yarikoptic:
docs - unrelated, fixed AFAIK elsewhere.

yarikoptic (approved):

yarikoptic-gitmate:

PR #7020, opened 2022-09-07
Revert back to non-devel neurodebian in crippledfs/extensions, add testing against -next

yarikoptic:
per brief chat on matrix


mih:
Worked. Thx!

PR #7022, opened 2022-09-08
Remove RIA functionality

mih:
> Note: for now this is a demo. It needs to be updated after https://github.com/datalad/datalad/pull/7017. Moreover, it might make sense to address https://github.com/datalad/datalad/issues/6961 at the same time.

This changeset is part of an effort to strip datalad-core off non-essential components in order to reduce technical debt and improve maintainability.

RIA (the concept of a dataset store, the git-annex special remote the URL handling) is not an essential component, but an application built on the datalad core.

https://github.com/datalad/datalad/pull/7017 shows how it can be completely separated from the core, because it is only integrated in within the `clone()` command -- all other code is located in pieces that can be installed in parallel.

At the same time RIA is large (~6k lines of code), and has accumulated as significant number of unaddressed technical and conceptual issues. The amount to more than 7% of the total number of currently open issues in datalad-core. Moreoever, it is severly underdocumented.

This change documents the code pieces that make up RIA, and the impact of their removal from core. They would move into an extension package, and remain available to users.

The most significant impact of this removal is the removal of a set of tests of core functionality that made use of RIA in one form or another. This loss could trun out to be substantial and require compensatory actions. However, these would have been needed anyways (and or simply harder to ignore now), because the RIA tooling did not work across all target platforms, and the majority of these tests were skipped on one or more platform.


bpoldrack:
Since removing this from core somewhat "breaks" existing datasets using the special remote, while core can not pull in an extension as a dependency and we can't mark a particular command as deprecated or something like this, I propose to replace the special remote in core with a dummy, that would simply spit out an INFO message reading "This special remote requires datalad-ria to be installed" or alike when the extension isn't installed. It would be replaced by the extension when installed and the messaging would automatically be triggered by any interaction with it. That way we wouldn't need to keep and possibly maintain any code in core that would do that. AFAICT, the only thing not covered by that would be the usage of `ria+` URLs with core's `clone`. Skipping such a URL and issuing a message would be a one-liner to keep in core, though.

All of that would be intended as the means for a proper deprecation cycle. So, doesn't need to stay forever.

yarikoptic:
I personally do not use RIA functionality myself yet, but I would like to avoid keeping `master` non-releasable because of this PR being merged. To avoid that, echoing @bpoldrack, and to provide continuity to users of RIA you already have I think that
- first the `datalad-ria` should be born/released and be added into extensions to be tested 
- @bpoldrack ideas for shims to inform user are good

mih:
I was informed that only PRs in non-draft mode are meant to be merged. If I misunderstood that, apologies.

The extension will be born from this PR. Not by merging it, but by seeing what the removal does to the rest.

If you see a better approach than this one. please enlighten me. Thx!

mih:
#7017 has been merged. Work here can continue.

mih:
I am closing this inactive PR of mine.

PR #7014, opened 2022-09-07
Remove metadata-related code after its move to -deprecated

mih:
This changeset is part of the transition from the 2nd-gen metadata implementation to then gen3 code provided by the -metalad extension.

This removal has side-effect on other code, not directly related to metadata. This following list can be considered a TODO for further maintenance work:

Now unused code:

- `datalad/support/json_py.py:load_xzstream()` : `load_xzstream` and `dump_xzstream` _removed in 82631286443cddde9051bb32f00f4dfc878d237b_
- `datalad/interface/common_opts.py:reporton_opt`: _removed in 1c2607230dc65f7f4995dc0031810b5ab208db5a_
- `datalad/utils.py:all_same()`: _removed in 5e50038ef4f4182d22b19448b60d54bdc6b49721_
- `datalad/utils.py:as_unicode()`: _removed in 13514bad9c43fa73fdba39b5b3a5e151483bf351_
- `datalad/utils.py:unicode_srctypes` made obsolete by removal of `as_unicode()`: _removed in 13514bad9c43fa73fdba39b5b3a5e151483bf351 (same as above)_

Now barely used (technical debt?):

- `nosave_opt`: Residual use in `addurls()` and `download_url()`
- `datalad/support/path.py:split_ext()`: Residual use in a single spot in `addurls()` <- https://github.com/datalad/datalad/issues/4056
- `datalad/utils.py:path_startswith()`: 2 pieces of code still use this outside the tests
- `datalad/utils.py:get_suggestions_msg()`: CLI parser and `addurls()` make the total of 2 calls made to this helper
- `datalad/utils.py:shortened_repr()`: one call in `get()` and one in `foreach_dataset()`

Issues to file:

- [x] Investigate whether the dependency on `simplejson` is still needed, `simplejson` can be removed by PR datalad/datalad#7035 
- [x] **Won't fix now** `datalad/interface/utils.py:discover_dataset_trace_to_targets()` is now an internal helper of `save()`
- [x] Clean uses the removed `SEARCH_INDEX_DOTGITDIR`
- [x] Fix doc-building
- [x] Move code to deprecated (datalad-deprecated#62 (https://github.com/datalad/datalad-deprecated/issues/62))

Related TODOs:

- close or move all remaining metadata-related issues https://github.com/datalad/datalad/labels/metadata
- https://github.com/datalad/datalad-deprecated/issues/74
- https://github.com/datalad/datalad-deprecated/issues/75
- https://github.com/datalad/datalad-deprecated/issues/76
[...]

PR #7027, opened 2022-09-12
DOC: update strings for command categories

aqw:
The new DataLad Gooey-GUI-Gewwee uses the category strings for commands for the UI, which increases the visibility of this problem.

This PR intends to achieve the following:
- make the text consistent (the word "commands" alternate between the front or back of the text description)
- reduce redundancy (they are all commands, no need to use the word "command" to redefine them)

I looked at `git --help`, and I /really/ like the way that they organize their subcommands. However, such a change for DataLad would involve a pretty significant reorganization of the subcommand categories, which I anticipate would be controversial. I aim for this change to enter first, and if there's desire for a further refinement of organization, then that could follow.

Lastly, this PR removes an epilogue from the DataLad help. Accusations of me of being a curmudgeon would not be entirely out of line --- but I do find the current rendering to be confusing. If indeed we want people to be happy, we should not confuse them in the process.

If there's strong support for the retention of the terse-and-imperative-message-of-optimism, then I can amend this PR to adjust the surrounding whitespace.

#### 📝 Documentation
- Command categories in help text are more consistently named.

### DIFF
--- a/datalad/interface/__init__.py
+++ b/datalad/interface/__init__.py
@@ -20,7 +20,7 @@ __docformat__ = 'restructuredtext'
 # the name of the `_group_*` variable determines the sorting in the command overview
 # alphanum ascending order
 _group_0dataset = (
-    'Essential commands',
+    'Essential',
     [
         # source module, source object[, dest. cmdline name[, dest python name]]
         # src module can be relative, but has to be relative to the main 'datalad' package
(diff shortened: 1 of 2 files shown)


datalad/interface/__init__.py
```diff
@@ -64,7 +64,7 @@
     ])
 
 _group_2metadata = (
-    'Commands for metadata handling',
+    'Metadata',
```
yarikoptic, line 67: minor but might want to avoid modifying this to avoid conflict with https://github.com/datalad/datalad/pull/7014

yarikoptic (approved):
FWIW, good with me! Thanks!

mih (approved):
Thx!

yarikoptic:
note: CI was red for appveyor and travis for a reason! See e.g. https://app.travis-ci.com/github/datalad/datalad/builds/255439732 . I will submit a PR to fix.

yarikoptic-gitmate:

PR #7028, opened 2022-09-12
Fix datalad --help to not have *Global options* empty and list options in "options:" section

yarikoptic:
Detected while considering #7027 (note: that one is against `master`, this one -- `maint`)

I believe came up as a regression of e89ce439a3b24607e4809a664a961a678063cded and since we had no unittest. So I also added a test.

Without this fix  --help looked like

	options:
	  -c (:name|name=value)
	...
	*Global options*

and nothing at the end after *Global options*.

### DIFF
--- a/changelog.d/pr-7028.md
+++ b/changelog.d/pr-7028.md
@@ -1,0 +1,6 @@
+### Bug Fixes
+
+- Fix datalad --help to not have *Global options* empty with python 3.10 and
+  list options in "options:" section.
+  [PR #7028](https://github.com/datalad/datalad/pull/7028)
+  (by [@yarikoptic](https://github.com/yarikoptic))
--- a/datalad/cli/helpers.py
+++ b/datalad/cli/helpers.py
@@ -78,7 +78,7 @@ class HelpAction(argparse.Action):
         options = []
         in_options = False
         for line in helpstr.splitlines():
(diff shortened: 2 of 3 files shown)


yarikoptic:
although changelog commit no longer has `[ci skip]` it seems that github completely skips that updated state in its CI. @jwodder any ideas on why that might be happening?

unfortunately adjusted test fails on travis and appveyor... need to figure why (passes locally). Could be python (argparse) version or how we install for testing... will need to troubleshoot

jwodder:
@yarikoptic I think this may be because the snippet is pushed using `secrets.GITHUB_TOKEN`, so GitHub doesn't trigger any workflows in order to avoid a potential infinite loop of workflows.

yarikoptic:
> @yarikoptic I think this may be because the snippet is pushed using `secrets.GITHUB_TOKEN`, so GitHub doesn't trigger any workflows in order to avoid a potential infinite loop of workflows.

hm, may be, odd though.  I wish there was some log to check on github which might have given explanation. If we would not be able to overcome this limitation, it would be quite suboptimal :-/

yarikoptic:
failures on travis and appveyor unrelated now.

mih (approved):
Works for me! Thx

yarikoptic-gitmate:

PR #6777, opened 2022-06-29
Fix test errors caused by updated chardet v5 release

christian-monch:
Fixes #6776 

This PR limits the allowed `chardet` version to <5.0.0, because `requests` does only support `chardet` up to version 4, and version 5.0.0. was released on pypi on 25th of June 2022

#### 🏠 Internal
- limit chardet version to version 4 in order to fulfill `requests`-dependencies

PR #7029, opened 2022-09-13
Update chardet version limit from 5 to 6

kimsin98:
#6777 is obsolete with psf/requests#6179. Could also remove the upper limit since it seems requests wants to drop the dependency eventually (https://github.com/psf/requests/issues/6177#issuecomment-1168148782).

#### 🏠 Internal
- Update chardet version limit to allow ver 5, matching `requests` dependencies.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -26,7 +26,7 @@ from _datalad_build_support.setup import (
 requires = {
     'core': [
         'platformdirs',
-        'chardet>=3.0.4, <5.0.0',      # rarely used but small/omnipresent
+        'chardet>=3.0.4',      # rarely used but small/omnipresent
         'colorama; platform_system=="Windows"',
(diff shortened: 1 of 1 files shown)


adswa:
This PR somehow failed to retrieve the Appveyor build status and displays a failure, but the actual build succeeded: https://ci.appveyor.com/project/mih/datalad/builds/44759641

mih (approved):
This is what we agreed on, to LGTM!

yarikoptic:
eh, "we agreed" with reasoning but now I see that travis seems was not even ran , and now we get failed tests like https://github.com/datalad/datalad/pull/7049#issuecomment-1253898475 on travis only.  also nobody requested/created changelog entry before merging. overall: not kosher IMHO.

Issue #7034, opened 2022-09-14
Replace `simplejson` with `json` to reduce dependencies


christian-monch:
Datalad uses `simplejson` instead of `json`, which introduces additional dependencies. Since the functionality is identical, `simplejson` should be replaced with `json`.

This change should also be done in the following extensions:

- [x] datalad-metalad: https://github.com/datalad/datalad-metalad/issues/285
- [ ] datalad-neuroimaging: https://github.com/datalad/datalad-neuroimaging/issues/107
- [ ] datalad-container: https://github.com/datalad/datalad-container/issues/181

bpoldrack:
@christian-monch: Do you have plans to finish this?

bpoldrack:
I think this is done as far as the core package is concerned and datalad-container has its own issue about it.

PR #7035, opened 2022-09-14
Remove `simplejson` in favor of using `json`

christian-monch:
Fixes issue #7034 

This PR removes `simplejson` in favor of the "builtin" `json`-module. 

Remark: this changes are also contained in PR #7014 

In addition this PR fixes the test `datalad.cli.tests.test_main.test_help_np`, by adapting the group names used in the test to the group names of the command groups.

- replace `simplejson` with `json` and remove `simplejson`

### DIFF
--- a/datalad/support/json_py.py
+++ b/datalad/support/json_py.py
@@ -20,13 +20,7 @@ from os.path import (
 from os import makedirs
 import os
 import os.path as op
-
-# wrapped below
-from simplejson import load as jsonload
-from simplejson import dump as jsondump
-# simply mirrored for now
-from simplejson import loads as json_loads
-from simplejson import JSONDecodeError
+import json
 
 from datalad.support.exceptions import CapturedException
 
@@ -36,8 +30,7 @@ json_dump_kwargs = dict(
     indent=0,
     separators=(',', ':\n'),
     sort_keys=True,
-    ensure_ascii=False,
-    encoding='utf-8', )
+    ensure_ascii=False,)
 
 # achieve minimal representation, but still deterministic
 compressed_json_dump_kwargs = dict(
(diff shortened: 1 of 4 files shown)


yarikoptic:
Is there comment in commit where we started to use simplejson - was there some specific reason or it was so long ago that there were no built in json?

christian-monch:
> Is there comment in commit where we started to use simplejson - was there some specific reason or it was so long ago that there were no built in json?

Yes, there is:

```
commit 5513caae92e82d26f0e90b1e642d18f9cb43e0a1
Author: Michael Hanke <michael.hanke@gmail.com>
Date:   Tue Aug 9 20:57:39 2016 +0200

    RF: Switch to simplejson, with default handling of utf-8
    
    Doesn't require additional jumps like the stdlib json module
```

I think at least since python 3.6, utf-8 encoded JSON should be properly handled by the built-in json-module

datalad/support/json_py.py
```diff
@@ -20,13 +20,7 @@
...
 import os
 import os.path as op
-
-# wrapped below
-from simplejson import load as jsonload
-from simplejson import dump as jsondump
-# simply mirrored for now
-from simplejson import loads as json_loads
-from simplejson import JSONDecodeError
+import json
```
yarikoptic, line 23: To minimize impact and possibly avoid some part of this PR diff I would have:

```suggestion
import json
from json import JSONDecodeError
```

mih: This would not be sufficient. There would need to be replacements for `jsonload`, `jsondump`, `json_loads` too.

I personally think a clean cut is better than yet another intermediate import target to preserve forever.

mih (approved):
LGTM! Thx!

My memory also confirm that the reason for `simplejson` was UTF handling with PY2. No longer a necessity.

yarikoptic-gitmate:

PR #7036, opened 2022-09-15
Let `create` touch the dataset root, if not saving in parent dataset

mih:
This communicates the "nature change" of the directory (from plain to dataset) to systems like inotify.

Without this, a dataset creation in an existing, empty subdirectory would go undiscovered, unless the empty subdirectory is explicitly watched itself.

With this change, that comes at virtually no additional cost, it is sufficient to watch the parent directory if the previously empty target subdirectory.

### DIFF
--- a/changelog.d/pr-7036.md
+++ b/changelog.d/pr-7036.md
@@ -1,0 +1,5 @@
+### Bug Fixes
+
+- Let `create` touch the dataset root, if not saving in parent dataset.  [PR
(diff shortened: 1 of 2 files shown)


yarikoptic (approved):
Despite clean description I do not quite grasp why it needs to be done on datalad end instead of in a "watcher" to have some logic to this effect.  But if it helps -- I would not mind this gentle touch

mih:
Thx!

yarikoptic-gitmate:

PR #7037, opened 2022-09-15
Let get_status_dict() use exception message if none is passed

mih:
A result with an exception attached has no error message, unless one is explicitly given, even when the exception is a CapturedException and could provide a message.

This change reports such an exception-provided message in the result.

### DIFF
--- a/changelog.d/pr-7037.md
+++ b/changelog.d/pr-7037.md
@@ -1,0 +1,5 @@
+### Bug Fixes
+
+- Let get_status_dict() use exception message if none is passed.  [PR
+  #7037](https://github.com/datalad/datalad/pull/7037) (by
+  [@mih](https://github.com/mih))
--- a/datalad/interface/results.py
+++ b/datalad/interface/results.py
@@ -107,6 +107,8 @@ def get_status_dict(action=None, ds=None, path=None, type=None, logger=None,
             if isinstance(exception, CapturedException) \
             else format_oneline_tb(
                 exception, include_str=False)
+        if error_message is None and isinstance(exception, CapturedException):
+            d['error_message'] = exception.message
         if isinstance(exception, CommandError):
             d['exit_code'] = exception.code
     if kwargs:


adswa:
There was a clone failure in an Appveyor run, which I have restarted.
<details><summary>I've seen the Travis failure in other PRs already:</summary>

```
=================================== FAILURES ===================================
_____________________________ test_gracefull_death _____________________________
[gw1] linux -- Python 3.7.8 /tmp/dl-miniconda-v52fqo40/bin/python
    @known_failure_osx  # https://github.com/datalad/datalad/issues/5309
    def test_gracefull_death():
    
        def assert_provides_and_raises(pc, exception, target=None):
            """Helper to get all results before exception is raised"""
            results = []
            with assert_raises(exception):
                for r in pc:
                    results.append(r)
            # results should be sorted since we do not guarantee order
            results = sorted(results)
            if target is not None:
                assert_equal(results, target)
            return results
    
        def interrupted_producer():
            yield 1
            raise ValueError()
    
        def consumer(i):
            sleep(0.001)
            yield i
    
        assert_provides_and_raises(
            ProducerConsumer(interrupted_producer(), consumer, jobs=3), ValueError, [1])
    
        def faulty_consumer(i):
            sleep(0.001)
            if i == 1:
                raise ValueError()
            return i
    
        # so we do not get failed, but other parallel ones finish their job
        results = assert_provides_and_raises(
            ProducerConsumer(range(1000), faulty_consumer, jobs=5), ValueError)
        # and analysis of futures to raise an exception can take some time etc, so
        # we could get more, but for sure we should not get all 999 and not even a 100
        if info_log_level:
            assert_greater(100, len(results))
        assert_equal(results[:4], [0, 2, 3, 4])
    
        def producer():
            for i in range(10):
                sleep(0.0003)
                yield i
            raise ValueError()
        # by default we do not stop upon producer failing
        assert_provides_and_raises(
            ProducerConsumer(producer(), consumer, jobs=2), ValueError, list(range(10)))
        # if producer produces more than we can as quickly consume but then fails
        # ATM we do not proceed to consume other items, but fail when we finish
        # consuming until the time point when producer has failed
        # by default we do not stop upon producer failing
        results = assert_provides_and_raises(
            ProducerConsumer(producer(), consumer, reraise_immediately=True, jobs=2),
            ValueError)
        # we will get some results, seems around 4 and they should be "sequential"
        assert_equal(results, list(range(len(results))))
    
        # This test relies too much on threads scheduling to not hog up on handling
        # consumers, but if it happens so - they might actually consume all results
        # before producer decides to finally raise an exception.  As such it remains
        # flaky and thus not ran, but could be useful to test locally while
        # changing that logic.
        #
        # if info_log_level and not (on_windows or on_osx):
        #     # consumers should not be able to consume all produced items.
        #     # production of 10 should take 3 unites, while consumers 10/2 (jobs)
        #     # 5 units, so some should not have a chance.
        #     assert_greater_equal(8, len(results))
    
        # Simulate situation close to what we have when outside code consumes
        # some yielded results and then "looses interest" (on_failure="error").
        # In this case we should still exit gracefully (no GeneratorExit warnings),
        # not over-produce, and also do not kill already running consumers
        consumed = []
        def inner():
            def consumer(i):
                sleep(0.01)
                consumed.append(i)
                return i
            pc = iter(ProducerConsumer(range(1000), consumer, jobs=2))
            yield next(pc)
            yield next(pc)
>       assert_equal(sorted(inner()), [0, 1])
../datalad/support/tests/test_parallel.py:223: 
first = [0, 2], second = [0, 1], msg = None
        if msg is None:
E           assert [0, 2] == [0, 1]
E             At index 1 diff: 2 != 1
E             Full diff:
E             - [0, 1]
E             ?     ^
E             + [0, 2]
E             ?     ^
../datalad/tests/utils_pytest.py:107: AssertionError
```
</details>

yarikoptic:
appveyor is green , travis as `test_parallel.py::test_gracefull_death ` fail which was mitigated since then, let's proceed and also try adjusted release workflow .

yarikoptic-gitmate:

PR #7043, opened 2022-09-16
adapt tests to modified command group names

christian-monch:
This PR fixes the test that verifies the command group names. It basically adapts the verified command group names to the changes from 8decbdc7c63d869665c5711a21aa9e837eda6f9f

#### 🏠 Internal
- Fix an error in the comman group names-test


christian-monch:
Should not have been cherry-picked. Will create a new PR that addresses the problem

PR #7047, opened 2022-09-20
Give more information about used token and performed actions

yarikoptic:
We have an odd behavior in being unable to comment on issues, and looks like if GITHUB_TOKEN was set to an empty value. Here we will error out if empty and otherwise reveal the beginning of it which by itself should not be leaking too much of a secret to do any harm.

Also print about actions being performed since now it is just silently does its deeds

[ci skip]
[appveyor skip]

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -57,7 +57,7 @@ jobs:
           python -m build
           twine upload dist/*
         env:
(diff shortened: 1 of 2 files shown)


yarikoptic:
eh, revert commit I guess did trigger CI, ill cancel appveyor and travis since no point

jwodder (approved):

yarikoptic:
thanks, let's try this one

Issue #7048, opened 2022-09-20
Move release workflow from `pull_request` to `push`?


yarikoptic:
ok, I did more investigation on why workflow switching from use of `auto` in  https://github.com/datalad/datalad/pull/7024  doesn't have credentials -- neither default GITHUB_TOKEN works nor the GITMATE_GITHUB_TOKEN works. The last summary was in https://github.com/datalad/datalad/pull/7046#issuecomment-1252539797 followed by https://github.com/datalad/datalad/pull/7047 which confirmed that we do not have a value for GITMATE_GITHUB_TOKEN available.

I have noted in the log for the action run that `Secret source: None` and logically got that we get them set only for `push` and not `pr`s:

```shell
(git)smaug:/mnt/datasets/datalad/ci/logs/2022/09[master]git
$> git grep 'Secret source: ' | sed -e 's,^[^/]*/\(pr\|push\).*[0-9]Z Secret source:,\1,g' | sort | uniq -c
   2162 pr None
    580 push Actions
```
and in #7024 we switched from `push` to `pull_request`.  Thus now we do not have any secrets available since it seems to be treated as a PR, just on merge:

```yaml
name: Auto-release on PR merge

on:
  pull_request:
    branches:
      - maint
    types:
      - closed
```

I do not remember exact reason and 597572ef3d2454b26dedc0883aa2a014d1513913 commit message provides none either.  @jwodder -- do you recall why is that and either we can may be get back to reacting on `push` somehow?  Or what other way you see to mitigate it and make releases work agian?

PS It might be that for PRs based off branches in the repo it would work, but it would be unfortunate to need to limit being able to release and remember about that only for PRs from branches from this repo.

jwodder:
@yarikoptic We need to use a `pull_request` trigger in order to have access to information about the PR that was just merged (specifically, whether it had the "release" label).  It's possible that using `pull_request_target` instead of `pull_request` will fix the situation.

yarikoptic:
How did we work it out in case of `auto`-powered on simple `push` -- we did the same reaction to merged PR with `release` label, right? so I guess it might be just a matter of taking the most recent merge commit and deducing which PR they correspond to?

jwodder:
@yarikoptic `auto` did the determination itself.  I think it parses commit messages looking for GitHub-generated merge commits and deduces the PRs from those.

yarikoptic:
[...]

PR #7051, opened 2022-09-21
Trigger release workflow on `pull_request_target`

jwodder:
See #7048.

The change in this PR may or may not take effect in the workflow that is run when it is merged; to be safe, this PR should *not* be labelled with "release", and a separate "release" PR should be made after it is merged in order to test it.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,7 +1,7 @@
 name: Auto-release on PR merge
 
 on:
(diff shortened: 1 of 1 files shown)


yarikoptic:
did you try this in some lightweight repo?

jwodder:
@yarikoptic Yes: https://github.com/datalad/temp-release-devel/pull/5

PR #7056, opened 2022-09-22
DATALAD_TESTS_GITCONFIG env variable to provide custom settings for git during tests

yarikoptic:
Although could be just once in a lifetime bug, we just got into an odd issue with an unlocked file with content being marked as modified https://git-annex.branchable.com/bugs/reports_file___34__modified__34___whenever_it_is_not/#comment-4e61c4c0f493e98f9416e09de4cfb33f which seems to boil down to have a rarely used `annex.stalldetection` set.

By adding one more env variable for TESTS we can now add such specific setting to be tested in the tests.

TODOs
- [x] make sure that tests pass -- we do might trigger that issue

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,7 +56,7 @@ matrix:
     - PYTEST_SELECTION=
     - PYTEST_SELECTION_OP=not
   # Two matrix runs for "recent python and git-annex with the recent supported by git annex
-  # new version of repo"
+  # new version of repo" and various extra options/features enabled for git-annex
   - python: '3.10'
     dist: focal
     env:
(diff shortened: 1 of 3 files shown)


joeyh:
I'm not sure if you're using git-annex test, but if you are, it also ignores global gitconfigs. I've added a --test-with-git-config option to it that you could use to run the git-annex test suite with a git config set.

yarikoptic:
it will need to wait for #7057 to get merged to get a kosher travis run.

@joeyh thanks for the note! we do not do `git-annex test` here but I am adding a matrix run for git annex test with that option in https://github.com/datalad/git-annex/pull/133

yarikoptic:
Travis is happy, so either this PR is not in effect or we just do not have a test which `get`s file in unlocked mode?  not sure if I am to craft one now, so I will just give it to review.

PR #7024, opened 2022-09-09
Stop using auto

jwodder:
Closes #7016.  See https://github.com/jwodder/pr-merge-release-test for a test of the workflow in action.

Closes #6856 , Closes #6918

TODOs:

- [x] comments on released PRs & issues
- [ ] does not overwrite old release notes; see https://github.com/nedbat/scriv/issues/57 -- was tested to work - we just need scriv release

PR #7058, opened 2022-09-23
use GITMATE token for actual push of changelog

yarikoptic:
Without that we use default token and github actions still do not run, although there is no guarantee that this would change it

Supplement to #7024 (a73411af88186750d136a6aa465af76e6df13508)

### DIFF
--- a/.github/workflows/add-changelog-snippet.yml
+++ b/.github/workflows/add-changelog-snippet.yml
@@ -59,5 +59,7 @@ jobs:
                git push origin HEAD:${{ github.head_ref }}
           else echo "No changes to commit"
           fi
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITMATE_GITHUB_TOKEN }}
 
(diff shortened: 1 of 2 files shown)


yarikoptic:
well, this change would not get in effect until we merge it.  I will just merge one commit (no need for changelog entry imho) locally and push so we could see effect on some other PR.

yarikoptic:
attn @jwodder that I did this PR, may be it was wrong ;-)

jwodder:
@yarikoptic When pushing a repo checked out with the `actions/checkout` action, the `secrets.GITHUB_TOKEN` which the action embedded in the clone on checkout is used; setting the `GITHUB_TOKEN` envvar has no effect.

yarikoptic:
oh, that is "fun".  I thought it was the whole point to use some other TOKEN than default so github tries to run its github actions upon push of the CHANGELOG fragment for scriv.

PR #7060, opened 2022-09-24
Make pytest to be the default tester for extensions

yarikoptic:
Our testing started to fail for container and metalad and I believe they did get pytest based testing version released recently which caused this


yarikoptic:
appveyor -- can't relate and it is `FAILED ../datalad/distribution/tests/test_install.py::test_install_subds_from_another_remote - `

metalad -- https://github.com/datalad/datalad-metalad/issues/288 attn @christian-monch 

Merging

Issue #6834, opened 2022-07-12
test_command_fail_1 fails on debian testing py 3.9 (only)


yarikoptic:
```
_____________________________ test_command_fail_1 ______________________________

    def test_command_fail_1():
        # Expect that the return code of a failing command is caught,
        # that None is returned as result.
        bc = BatchedCommand(
            cmd=py2cmd(
                """
    print("something")
    exit(3)
                """))
    
        # Send something to start the process
        result = bc("line one")
        assert_equal(bc.return_code, None)
        assert_equal(result, "something")
        result = bc("line two")
>       assert_equal(bc.return_code, 3)

datalad/tests/test_cmd.py:165: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

first = None, second = 3, msg = None

    def assert_equal(first, second, msg=None):
        if msg is None:
>           assert first == second
E           assert None == 3
E             +None
E             -3

datalad/tests/utils_pytest.py:107: AssertionError
```

as was added in 9f066efa6a - @christian-monch any ideas ?

yarikoptic:
on a rebuild of 0.17.1-2 which is functionally the same it went through just fine. Thus adding a label for the best kinds of test failure.

christian-monch:
There is actually an error here. The test is assuming a behavior that does not match the control flow in `BatchedCommand`. A little more detail:`BatchedCommand`will restart the process when `result = bc("line two")` is executed and the process did already exit. This is not anticipated by the test.

christian-monch:
I looked at the history of the restart-code:

The restart code was introduced in PR #334 (https://github.com/datalad/datalad/pull/334), commit 74dada12ce70a5535eb59c926165e4433091436a. The commit was made 2016-01-15 and the PR was merged 2016-01-27

Besides the commit message. no rationale was given:
```
BF/ENH: restart --batched process if it crashes
```

On 2019-04-08 the restart code was factored out into `BatchedCommand`. It seems that the restart code is annex-specific and should probably live in `BatchedAnnex`.

yarikoptic:
Thank you @christian-monch for the investigation! I wish I left more information for that commit. Given the age when that was introduced I think it should indeed be safe to assume that we might now change and just bubble up the error instead of restarting. Use on the wild might point to few left corner cases

yarikoptic:
[...]

PR #7066, opened 2022-10-05
BF: disable testing for exit code for batched process. 

yarikoptic:
attn @christian-monch .  Proper solution is to be worked out but I think for now we should just disable that test (at least I would place that patch into conda build for now)

Ref: https://github.com/datalad/datalad/issues/6834


yarikoptic:
for some reason the job for adding CHANGELOG is just skipped... dunno that either but since hopefully the issue itself would be addressed (right @christian-monch ?) I will just proceed as is for now.

PR #7068, opened 2022-10-06
Ensure that `BatchedCommandError` is raised if the subprocesses of `BatchedCommand` fails or raises a `CommandError`

### LINKED ISSUE #6834: test_command_fail_1 fails on debian testing py 3.9 (only)

christian-monch:
Fixes #6834

This PR ensures that `BatchedCommand` raises a `BatchedCommandError` if the subprocesses of `BatchedCommand` exits with a non-zero value or raises a `CommandError`.

#### 🐛 Bug Fixes
- Adapt `BatchedCommand`-tests to actual behavior. Fixes #6834
- Ensure the errors from sub-processes of `BatchedCommand` are signaled to the caller via `BatchedCommandError`-exception

### DIFF
--- a/changelog.d/pr-7068.md
+++ b/changelog.d/pr-7068.md
@@ -1,6 +1,3 @@
 ### Internal
 
-- Ensure that `BatchedCommandError` is raised if the subprocesses of
-  `BatchedCommand` fails or raises a `CommandError`.  [PR
-  #7068](https://github.com/datalad/datalad/pull/7068) (by
-  [@christian-monch](https://github.com/christian-monch))
+- Ensure that `BatchedCommandError` is raised if the subprocesses of `BatchedCommand` fails or raises a `CommandError`.  [PR #7068](https://github.com/datalad/datalad/pull/7068) (by [@christian-monch](https://github.com/christian-monch))
--- a/changelog.d/pr-7068.md
+++ b/changelog.d/pr-7068.md
@@ -1,0 +1,6 @@
+### Internal
+
+- Ensure that `BatchedCommandError` is raised if the subprocesses of
+  `BatchedCommand` fails or raises a `CommandError`.  [PR
+  #7068](https://github.com/datalad/datalad/pull/7068) (by
+  [@christian-monch](https://github.com/christian-monch))
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -401,7 +401,7 @@ class BatchedCommand(SafeDelCloseMixin):
             if not self.process_running():
                 self._initialize()
 
-            # Remember request end send it to subprocess
+            # Remember request and send it to subprocess
             if not isinstance(request, str):
                 request = ' '.join(request)
             self.stdin_queue.put((request + "\n").encode())
--- a/datalad/cmd.py
+++ b/datalad/cmd.py
@@ -68,6 +68,30 @@ from datalad.utils import (
 )
 
 
+class BatchedCommandError(CommandError):
+    def __init__(self,
+                 cmd="",
+                 request="",
+                 msg="",
+                 code=None,
+                 stdout="",
+                 stderr="",
+                 cwd=None,
+                 **kwargs):
+
+        CommandError.__init__(
+            self,
+            cmd=cmd,
+            msg=msg,
+            code=code,
+            stdout=stdout,
+            stderr=stderr,
+            cwd=cwd,
+            **kwargs
+        )
+        self.request = request
+
+
 lgr = logging.getLogger('datalad.cmd')
 
 # TODO unused?
@@ -200,6 +224,7 @@ class BatchedCommand(SafeDelCloseMixin):
         self.wait_timed_out = None
         self.return_code = None
         self._abandon_cache = None
+        self.last_request: Optional[str] = None
 
         self._active = 0
         self._active_last = _now()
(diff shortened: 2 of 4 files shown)


datalad/cmd.py
```diff
@@ -335,10 +374,19 @@ def __call__(self,
                         self.runner = None
 
         except CommandError as command_error:
-            # The command exited with a non-zero return code
-            lgr.error("%s: command error: %s", self, command_error)
-            self.return_code = command_error.code
+            # Convert CommandError into BatchedCommandError
             self.runner = None
+            self.return_code = command_error.code
+            raise BatchedCommandError(
```
bpoldrack, line 380: I think this should be a `raise BatchedCommandError(...) from command_error` to not loose the original context/traceback.

bpoldrack (commented):
From looking at the code, I have one comment, otherwise looks right to me.

changelog.d/pr-7068.md
```diff
@@ -0,0 +1,6 @@
+### Internal
+
+- Ensure that `BatchedCommandError` is raised if the subprocesses of
+  `BatchedCommand` fails or raises a `CommandError`.  [PR
+  #7068](https://github.com/datalad/datalad/pull/7068) (by
+  [@christian-monch](https://github.com/christian-monch))
```
yarikoptic, line 6: @jwodder is reworking release helpers into a reusable action and it should get warping fixed, meanwhile

```suggestion
- Ensure that `BatchedCommandError` is raised if the subprocesses of `BatchedCommand` fails or raises a `CommandError`.  [PR #7068](https://github.com/datalad/datalad/pull/7068) (by [@christian-monch](https://github.com/christian-monch))
```

datalad/cmd.py
```diff
@@ -68,6 +68,30 @@
 )
 
 
+class BatchedCommandError(CommandError):
```
yarikoptic, line 71: please add docstring to describe why we need it and what is "request"... with all `requests` and/or weekend took me a bit to understand what "request" it is dealing with,

datalad/cmd.py
```diff
@@ -353,7 +401,7 @@ def process_request(self,
             if not self.process_running():
                 self._initialize()
 
-            # Send request to subprocess
+            # Remember request end send it to subprocess
```
yarikoptic, line 404:
```suggestion
            # Remember request and send it to subprocess
```
?

datalad/support/annexrepo.py
```diff
@@ -2761,9 +2761,14 @@ def is_available(self, file_, remote=None, key=False, batch=False):
...
-            out = self._batched.get(
-                ':'.join(annex_cmd), annex_cmd,
-                path=self.path)(key_)
+            try:
+                out = self._batched.get(
+                    ':'.join(annex_cmd), annex_cmd,
+                    path=self.path)(key_)
+            except CommandError:
+                # git-annex will signal an unknown remote by exiting with
+                # return code 1.
```
yarikoptic, line 2770: hm, then shouldn't return_code be checked here and if not `1` then just re-raise?
Is return code `1` used for anything else?

yarikoptic (approved):
I agree with @bpoldrack recommendation on raising with preserving "origin". I also have some odd gut feeling that we might see some odd fails in parallel executions, but it might be only for good since we might really need to look into making those batched commands (more) thread-safe.  But overall - I approve at this stage of our datalad development ;-) thanks @christian-monch

datalad/support/annexrepo.py
christian-monch: I am not sure, whether it is used for anything else. In non-batched mode `1` is used if the remote does not exist, or if the content is not present in the remote. There might be another error situation that is mapped onto returning `1`.

bpoldrack: I think to be inline with what this function did before, we want:

```
except CommandError as e:
    CapturedException(e)  # logs the exception
    out = e.stdout
```

and then proceed (potentially leading to a ValueError with a message). Any potential issue I can see with this, seems to be part of how this method behaves and has nothing to do with this PR.
Or do you see a case where this solution goes wrong, @christian-monch ?

(I seem unable to make this a committable "suggestion" as I don't get the right anchor point for it)

christian-monch: I updated the comment in c5419974d4a9c9c015f68145d03eacaec7ce46f3 to describe more precisely what was meant, i.e. git-annex might still exit with non-zero in batch mode instead of returning errors in JSON-output records

yarikoptic:
@christian-monch please add `CHANGELOG-missing` label so then workflow would produce a scriv changelog stub (based on title though, not the ones you added to description) -- you would then adjust it to your liking.

datalad/support/annexrepo.py
christian-monch: Adressed in datalad-internal call

yarikoptic-gitmate:

Issue #6848, opened 2022-07-15
test_push.py::test_nested_pushclone_cycle_allplatforms fails some times


yarikoptic:
in cron jobs happens only in a single matrix run

```
$> datalad foreach-dataset "git grep 'FAILED .*test_nested_pushclone_cycle_allplatforms'  | grep cron || :"            
foreach-dataset(ok): /mnt/datasets/datalad/ci/logs/2022 (dataset)
03/cron/20220603T193647/173dc6f/travis-14100-failed/14.txt:FAILED ../core/distributed/tests/test_push.py::test_nested_pushclone_cycle_allplatforms
foreach-dataset(ok): /mnt/datasets/datalad/ci/logs/2022/01 (dataset)
10/cron/20220610T193717/029f839/travis-14136-failed/14.txt:FAILED ../core/distributed/tests/test_push.py::test_nested_pushclone_cycle_allplatforms
20/cron/20220520T193556/84f979a/travis-14036-failed/14.txt:FAILED ../core/distributed/tests/test_push.py::test_nested_pushclone_cycle_allplatforms
27/cron/20220527T193600/310fd9f/travis-14055-failed/14.txt:FAILED ../core/distributed/tests/test_push.py::test_nested_pushclone_cycle_allplatforms
foreach-dataset(ok): /mnt/datasets/datalad/ci/logs/2022/06 (dataset)
foreach-dataset(ok): /mnt/datasets/datalad/ci/logs/2022/05 (dataset)
foreach-dataset(ok): /mnt/datasets/datalad/ci/logs/2022/04 (dataset)
15/cron/20220715T194006/52e822a/travis-14276-failed/14.txt:FAILED ../core/distributed/tests/test_push.py::test_nested_pushclone_cycle_allplatforms
foreach-dataset(ok): /mnt/datasets/datalad/ci/logs/2022/07 (dataset)
foreach-dataset(ok): /mnt/datasets/datalad/ci/logs/2022/03 (dataset)
foreach-dataset(ok): /mnt/datasets/datalad/ci/logs/2022/02 (dataset)
```
and recent case https://app.travis-ci.com/github/datalad/datalad/jobs/576842696#L3269 .  That is the matrix run which uses `PYTEST_WRAPPER="sudo -E" DATALAD_TESTS_SSH=0`

PR #4454, opened 2020-04-28
WiP: dirty draft for alternative paths reporting by default renderer

yarikoptic:
**Warnings**: Don't look at the diff, this is more of a "tech preview", tests guaranteed to fail, diff has unrelated changes etc. Authors of the handbook and/or documentation which hardcoded outputs most likely would start hating me.

Prior attempt was gh-2679 but didn't proceed on that one since then the main problem I was fighting was full paths, and then paths relative to `refds` came to resolve that aspect. I kept running into situations where "relative to refds" made little sense, if not to say that it was misleading since `subds` could differ -- see gh-3442 .  The reason is that although the behavior of "always to refds" is easy to describe, since the actual `refds` is not shown, user is likely to be provided with paths he cannot just cut/paste in majority of the cases IMHO, and would not even know that they are not really such.

So, with this PR output paths would be either relative
- to the CWD
- to the specified dataset, and thus prefixed with an explicit `-d/` prefix (could be anything, but would be nice to have some explicit indication) to signal that.

Behavior: The second case is excercised only if that dataset is outside of the CWD.

Appearance: we could use some other demarkation between the two, e.g. make all relative paths start with `os.curdir` and relative to the dataset not have any `os.curdir` prefix.

<details>
<summary>Here is a demo script which exercises a number of scenarios invoking basic commands with or without `-d` and possibly from within a subdirectory of a dataset</summary> 

```bash
#!/bin/bash

export PS4="> "
set -eu
cd "$(mktemp -d ${TMPDIR:-/tmp}/dl-XXXXXXX)"

which datalad
datalad --version

set -x
datalad create src
datalad create -d src src/subds
datalad create -d src src/subds/subsubds

(
cd src/subds
touch file
touch subsubds/file

mkdir d1 d2
touch d1/f1 d2/f2
cd d1
datalad save -d^  -r 
#file subsubds/file
)

# now we will try from some very outside directory which has nothing in
# common with the dataset path
mkdir outside

# introduce some changes to the dataset somewhere deep
rm src/subds/d2/f2
touch src/newfile

(
cd outside
datalad save -r -d ../src
)
```
</details>

<details>
<summary>Full output on master</summary> 

```shell
/home/yoh/proj/datalad/datalad-master/venvs/dev3/bin/datalad
datalad 0.12.6.dev611
> datalad create src
[INFO   ] Creating a new annex repo at /home/yoh/.tmp/dl-zHdjBPX/src 
[...]

Issue #3994, opened 2020-01-09
Support `datalad -f '{relpath}'`


mih:
It would be convenient to be able get relative paths from the cmdline results for programatic post-processing. I.e. like this:

```
% datalad diff -f incoming^ -t incoming                 
    added: 1008788_20227_2_0.zip (file)
    added: 1008788_20249_2_0.zip (file)
    added: 1008788_20252_2_0.zip (file)
```

but just the relative path. Absolute is working already, relative not:

```
% datalad -f '{path}' diff -f incoming^ -t incoming                                                                                      
/data/project/rehab_biobank/datalad_trial/1008788/.datalad/.gitattributes
/data/project/rehab_biobank/datalad_trial/1008788/.datalad/config
/data/project/rehab_biobank/datalad_trial/1008788/.gitattributes
/data/project/rehab_biobank/datalad_trial/1008788/1008788_20227_2_0.zip
/data/project/rehab_biobank/datalad_trial/1008788/1008788_20249_2_0.zip
/data/project/rehab_biobank/datalad_trial/1008788/1008788_20252_2_0.zip

% datalad -f '{relpath}' diff -f incoming^ -t incoming
N/A
N/A
N/A
N/A
N/A
N/A
```

mih:
I take that back. Adding this is less useful than I thought. The refds setting required for it is not very reliable, a `basename` is easy to produce on the cmdline, and it would still need result filtering in many places. I now believe that a `jq`-based solution is superior and already fully supported right now.

Example: Report the path of any dataset content that were added or modified by the last commit
```
datalad -f json diff -f HEAD^ -t HEAD | jq '. | select(.state == "added", .state == "modified") | .path'
```

yarikoptic:
note that there was https://github.com/datalad/datalad/pull/2679 for this goal, didn't finish up, but could be a starting point

PR #7071, opened 2022-10-07
NF: foreach-dataset --o-s relpath - capture and pass-through prefixing with path to subds

yarikoptic:
First commit is just a helper, 2nd is the actual changes with following ATM description:

Motivation for this feature is frequent use of  `foreach-dataset  git grep PATTERN` on e.g
captured by con/tinuous logs as I did in e.g. https://github.com/datalad/datalad/issues/6848 :

    $> datalad foreach-dataset "git grep 'FAILED .*test_nested_pushclone_cycle_allplatforms'  | grep cron || :"
    foreach-dataset(ok): /mnt/datasets/datalad/ci/logs/2022 (dataset)
    03/cron/20220603T193647/173dc6f/travis-14100-failed/14.txt:FAILED ../core/distributed/tests/test_push.py::test_nested_pushclone_cycle_allplatforms
    foreach-dataset(ok): /mnt/datasets/datalad/ci/logs/2022/01 (dataset)
    10/cron/20220610T193717/029f839/travis-14136-failed/14.txt:FAILED ../core/distributed/tests/test_push.py::test_nested_pushclone_cycle_allplatforms
    20/cron/20220520T193556/84f979a/travis-14036-failed/14.txt:FAILED ../core/distributed/tests/test_push.py::test_nested_pushclone_cycle_allplatforms
    27/cron/20220527T193600/310fd9f/travis-14055-failed/14.txt:FAILED ../core/distributed/tests/test_push.py::test_nested_pushclone_cycle_allplatforms
    foreach-dataset(ok): /mnt/datasets/datalad/ci/logs/2022/06 (dataset)
    foreach-dataset(ok): /mnt/datasets/datalad/ci/logs/2022/05 (dataset) 
    ...

As you could see - the default pass-through + default rendering there is "suboptimal" since

- main hurdle: paths from `grep` are relative to that subdataset and not
  readily cut-pasteable since we need to join with subdataset path first
- pollutes with default renderer 'ok' or not for  a given dataset (hence needed
  || : to avoid complains)
- prints all those (ok) which we do not quite care about -- we just want to see output
  as this hierarchy was a single dataset, so if nothing hit -- we just carry forward.

With this PR and "relpath" mode, we adjust renderer so that output looks like:

    (git)smaug:/mnt/datasets/datalad/ci/logs/2022[master]git
    $> datalad foreach-dataset --o-s relpath "git grep 'FAILED .*test_nested_pushclone_cycle_allplatforms'  | grep cron "
    06/03/cron/20220603T193647/173dc6f/travis-14100-failed/14.txt:FAILED ../core/distributed/tests/test_push.py::test_nested_pushclone_cycle_allplatforms
    06/10/cron/20220610T193717/029f839/travis-14136-failed/14.txt:FAILED ../core/distributed/tests/test_push.py::test_nested_pushclone_cycle_allplatforms
    05/20/cron/20220520T193556/84f979a/travis-14036-failed/14.txt:FAILED ../core/distributed/tests/test_push.py::test_nested_pushclone_cycle_allplatforms
    05/27/cron/20220527T193600/310fd9f/travis-14055-failed/14.txt:FAILED ../core/distributed/tests/test_push.py::test_nested_pushclone_cycle_allplatforms
    07/15/cron/20220715T194006/52e822a/travis-14276-failed/14.txt:FAILED ../core/distributed/tests/test_push.py::test_nested_pushclone_cycle_allplatforms
  
pros:

- readily usable path which melded relative path to the subdataset
  with output of  grep
- since we ignore status -- no errors reported even though I had no ||:
- no useless ok's
- with "smart" treatment of either "dataset" argument is provided or not
  IMHO reported relative paths are quite neat despite C1 (below)
- I thought it could be sufficiently closely matched simply via
  `datalad -f '{path}{stdout}' foreach-dataset --o-s capture` but "not":

        $> datalad -f '{path}/{stdout}' foreach-dataset --o-s capture "git grep 'FAILED .*test_nested_pushclone_cycle_allplatforms'  | grep cron ||:"
        /mnt/datasets/datalad/ci/logs/2022/
        /mnt/datasets/datalad/ci/logs/2022/01/
        /mnt/datasets/datalad/ci/logs/2022/06/03/cron/20220603T193647/173dc6f/travis-14100-failed/14.txt:FAILED ../core/distributed/tests/test_push.py::test_nested_pushclone_cycle_allplatforms
        10/cron/20220610T193717/029f839/travis-14136-failed/14.txt:FAILED ../core/distributed/tests/test_push.py::test_nested_pushclone_cycle_allplatforms

        /mnt/datasets/datalad/ci/logs/2022/05/20/cron/20220520T193556/84f979a/travis-14036-failed/14.txt:FAILED ../core/distributed/tests/test_push.py::test_nested_pushclone_cycle_allplatforms
        27/cron/20220527T193600/310fd9f/travis-14055-failed/14.txt:FAILED ../core/distributed/tests/test_push.py::test_nested_pushclone_cycle_allplatforms

        /mnt/datasets/datalad/ci/logs/2022/04/

  since we would get bunch of empty lines, those paths for datasets without hits etc.
  So - decided that it is worth adding this new mode

cons:

- C1: overfitting "grep" use case since we prefix with actual path and without
  any separation (like via :).  So may be we should name it `grep` instead of `relpath`.
- C2: someone might want full path! But hey -- there is a reason why I named it "relpath"

both cons somewhat point to may be choosing another, better describing name. **Please suggest one!**

Note: "relpath" idea in output rendering was more generally excercised but
never finalized in other PRs

- https://github.com/datalad/datalad/pull/2679
- https://github.com/datalad/datalad/pull/4454

and argued in https://github.com/datalad/datalad/issues/3994 to be "less useful" which I would disagree.

NB code is based off `maint` but positioning against master since a new feature, it would be a sin to propose to `maint`.

### DIFF
--- a/datalad/local/foreach_dataset.py
+++ b/datalad/local/foreach_dataset.py
@@ -13,6 +13,9 @@ __docformat__ = 'restructuredtext'
 
 import inspect
 import logging
+import os.path as op
+import sys
+from typing import Union
 
 from argparse import REMAINDER
 from itertools import chain
@@ -167,9 +170,11 @@ class ForEachDataset(Interface):
             `contains` is used"""),
         output_streams=Parameter(
             args=("--output-streams", "--o-s"),
-            constraints=EnsureChoice('capture', 'pass-through'),
-            doc="""whether to capture and return outputs from 'cmd' in the record ('stdout', 'stderr') or
-            just 'pass-through' to the screen (and thus absent from returned record)."""),
+            constraints=EnsureChoice('capture', 'pass-through', 'relpath'),
+            doc="""ways to handle outputs. 'capture' and return outputs from 'cmd' in the record ('stdout',
+            'stderr'); 'pass-through' to the screen (and thus absent from returned record); prefix with 'relpath'
+            and just pass-through (similar to like grep does), and if `dataset` is specified - relative to top
+            of that dataset, if not - to current directory."""),
         chpwd=Parameter(
             args=("--chpwd",),
             constraints=EnsureChoice('ds', 'pwd'),
@@ -316,7 +321,7 @@ class ForEachDataset(Interface):
                         if output_streams == 'pass-through':
                             res = cmd_f(*cmd_a, **cmd_kw)
                             out = {}
-                        elif output_streams == 'capture':
+                        elif output_streams in ('capture', 'relpath'):
                             with swallow_outputs() as cmo:
                                 res = cmd_f(*cmd_a, **cmd_kw)
                                 out = {
@@ -342,7 +347,7 @@ class ForEachDataset(Interface):
                         cmd_expanded,
                         cwd=ds.path if chpwd == 'ds' else pwd,
                         protocol=protocol)
-                if output_streams == 'capture':
+                if output_streams in ('capture', 'relpath'):
                     status_rec.update(out)
                     # provide some feedback to user in default rendering
                     if any(out.values()):
@@ -377,7 +382,7 @@ class ForEachDataset(Interface):
                     warning += \
                         "Execution of Python commands in parallel threads while changing directory " \
                         "is not thread-safe. "
-                if output_streams == 'capture':
+                if output_streams in ('capture', 'relpath'):
                     warning += \
                         "Execution of Python commands in parallel while capturing output is not possible."
                 if warning:
@@ -395,6 +400,33 @@ class ForEachDataset(Interface):
             **pc_kw
         )
 
+    @staticmethod
+    def custom_result_renderer(res, **kwargs):
+        from datalad.interface.utils import generic_result_renderer
+        if kwargs.get('output_streams') == 'relpath':
+            from datalad.log import no_progress
+            with no_progress():
+                ds: Union[str, Dataset] = kwargs.get('dataset')
+                if ds:
+                    if not isinstance(ds, Dataset):
+                        ds = Dataset(ds)  # so all ///, ^ etc get treated
+                    refpath = ds.path
+                else:
+                    refpath = getpwd()
+                for k in ('stdout', 'stderr'):
+                    v = res.get(k)
+                    if v:
+                        path = res.get('path')
+                        relpath = op.relpath(path, refpath) if path else ''
+                        if relpath == op.curdir:
+                            relpath = ''
+                        if relpath and not relpath.endswith(op.sep):
+                            relpath += op.sep
+                        out = getattr(sys, k)
+                        for l in v.splitlines():
+                            out.write(f"{relpath}{l}\n")
+        else:
+            generic_result_renderer(res)
 
 # Reduced version from run
 def format_command(command, **kwds):
--- a/datalad/local/foreach_dataset.py
+++ b/datalad/local/foreach_dataset.py
@@ -173,8 +173,9 @@ class ForEachDataset(Interface):
             constraints=EnsureChoice('capture', 'pass-through', 'relpath'),
             doc="""ways to handle outputs. 'capture' and return outputs from 'cmd' in the record ('stdout',
             'stderr'); 'pass-through' to the screen (and thus absent from returned record); prefix with 'relpath'
-            and just pass-through (similar to like grep does), and if `dataset` is specified - relative to top
-            of that dataset, if not - to current directory."""),
+            captured output (similar to like grep does) and write to stdout and stderr. In 'relpath', relative path
+            is relative to the top of the dataset if `dataset` is specified, and if not - relative to current
+            directory."""),
         chpwd=Parameter(
             args=("--chpwd",),
             constraints=EnsureChoice('ds', 'pwd'),
--- a/changelog.d/pr-7071.md
+++ b/changelog.d/pr-7071.md
@@ -1,0 +1,7 @@
+### 🚀 Enhancements and New Features
+
+- `foreach-dataset` command got a new possible value for the --output-streamns|--o-s
+  option 'relpath' to capture and pass-through prefixing with path to subds.  Very
+  handy for e.g. running `git grep` command across subdatasets.
+  [PR #7071](https://github.com/datalad/datalad/pull/7071)
+  (by [@yarikoptic](https://github.com/yarikoptic))
--- a/datalad/local/tests/test_foreach_dataset.py
+++ b/datalad/local/tests/test_foreach_dataset.py
@@ -7,6 +7,7 @@
 # ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
 """Test foreach-dataset command"""
 
+import os.path as op
 import sys
 from pathlib import Path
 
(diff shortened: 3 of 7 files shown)


bpoldrack (commented):
Hm. I understand the problem and hence the desire for that kind of result rendering. 

But I'm a bit confused by the appraoch. Why entangling a result rendering option with the output stream option? This kinda destroys a basic concept: That a `datalad` option like `-f` should not interfere with actual command behavior. But now it does. If I specify `-f`, I overrule `--output-stream` which is far from obvious. It also means, that if I disable result rendering altogether, I don't get any output "passed-through". The doc declaring that `relpath` essentially is like pass-through suggests otherwise and is technically a lie. It's `capture`+prefix+spit out again (as long as one doesn't specify a result renderer). That's a different concept.
I feel this is an abuse of the result renderer concept (including writing to sys.stdout/err directly instead of `ui`). And I agree - it's an overfit to `grep`. 

Not happy with it, but also not instantly clear to me, what's the better alternative, though. I think what you really want is disable result rendering and read+prefix+write the actual output pipes as things are coming in. That would suggest a runner protocol for it (but another construct for python callables, I guess). Hm.

yarikoptic:
An interesting analysis/concern @bpoldrack.   But even with `--o-s relpath` we would still have `-f` working as expected, thus overloading any `custom_result_renderer` and that is our design. And wouldn't your concern then apply to **any other** interface with `custom_result_renderer` (14 of those?), especially the one which uses `kwargs` to tune its output? 

But I am open for any pragmatic suggestion on how to tune it up. As I have mentioned in the original description I was even considering smth like `-f {path}{stdout}` but I have mentioned its shortcoming.   I guess we could have come up with some `-f {stdout.prefix_eachline(relpath)}` or smth like that, and I do want to get back to those issues I have mentioned and add/use `relpath` to any record, but I feel that due to need to process captured data in specific way (one per line), I still think proposed somewhat ad-hoc way might be the best way forward.

bpoldrack:
On second thought, @yarikoptic :

Actually, it seems to me we "simply" need to fix result filtering (#5581). With `--report-type file` actually working your `-f {path}{stdout}` approach should actually work, since the undesired stuff comes from dataset type results. WDYT?

bpoldrack:
On second thought, @yarikoptic:

I think, what we need to do instead is properly enabling result filtering. What's missing from your `-f {path}{stdout}` approach is to only use results that actually have a `stdout`, I think. Which brings up #5581 WRT result filtering.
That would seem to generalize much better. WDYT?

I would like to not conflate the "output stream of a command" with its rendered results. Those are two different concepts. While rendered results can be part of the output stream of a command, that's not generally true and an output stream of a command can contain things that are not rendered results (as is the case with `--o-s pass-through`).

Another aspect of this is the writing to stdout directly from within the result renderer. Not going through `ui`, disables other interfaces like `datalad-gooey`, where one would probably want to have the same result rendering (ready for c&p or drag and drop). Ignoring other `ui` backends would sabotage that.

yarikoptic:
> I think, what we need to do instead is properly enabling result filtering. What's missing from your `-f {path}{stdout}` approach is to only use results that actually have a `stdout`, I think. Which brings up #5581 WRT result filtering. That would seem to generalize much better. WDYT?

what about needing to prefix each line in `stdout` -- how would we instruct that?

> Another aspect of this is the writing to stdout directly from within the result renderer. Not going through `ui`, disables other interfaces like `datalad-gooey`, where one would probably want to have the same result rendering (ready for c&p or drag and drop). Ignoring other `ui` backends would sabotage that.

We would need to differ `stdout` and `stderr`. In `ui` we have  `message` and `error` but using `.error` would add undesired prefix `ERROR`. I guess we could make prefix configurable and then make it `STDERR` although also prefix would be displayed after, since I aim for stdout -- ok, with me.

On the other hand, the fact that everything centralizes via `ui` is good, but I bet gooey etc should still capture/report somehow stdout/stderr somehow to not swallow the output.

yarikoptic:
note: rebased on top of master (was on maint) since diff was poisoned with other maint diffs

yarikoptic:
I like the idea, thank you @bpoldrack . My concerns are
- forgot what I did re `python` commands, your solution covers only the externally executed commands
- the same concern as yours about partial `data`, so might need some internal accumulation if data comes without trailing newline...  I will wait for analysis by @christian-monch too ;)
- should write to the `fd` I guess not to `1` all the time

yarikoptic:
I liked the idea of PrefixStdout (despite that would need to also add `-f disabled` to `datalad` invocation), and started to implement but then:

> * forgot what I did re `python` commands, your solution covers only the externally executed commands

well, we just `eval` or `exec` them, so I do not think we have any direct control over stdout/err AFAIK and would need to monkey-patch sys.stdout/err and hope that those were not bound in that underlying function first , i.e. any monkey patching would work.
We could then make it also non-python only, but that would kind suck IMHO.

With that it begs to question if it is a worthwhile approach?   ATM I have been using this PR version successfully for my use case of invoking `git grep`, which is an external command.  So, may be give it another consideration?

bpoldrack:
@yarikoptic 
Yeah, I don't have an all-including solution either. So, let's go with it. However, may be try to rephrase the help at least? It's described as a prefixed pass-through, but that's not what it is. It's a prefixed `capture` + dedicated rendering to spit it out again (after execution in a ds finished).

datalad/local/foreach_dataset.py
```diff
@@ -167,9 +170,11 @@ class ForEachDataset(Interface):
...
         output_streams=Parameter(
             args=("--output-streams", "--o-s"),
-            constraints=EnsureChoice('capture', 'pass-through'),
-            doc="""whether to capture and return outputs from 'cmd' in the record ('stdout', 'stderr') or
-            just 'pass-through' to the screen (and thus absent from returned record)."""),
+            constraints=EnsureChoice('capture', 'pass-through', 'relpath'),
+            doc="""ways to handle outputs. 'capture' and return outputs from 'cmd' in the record ('stdout',
+            'stderr'); 'pass-through' to the screen (and thus absent from returned record); prefix with 'relpath'
+            and just pass-through (similar to like grep does), and if `dataset` is specified - relative to top
+            of that dataset, if not - to current directory."""),
```
yarikoptic, line 177: @bpoldrack, re your suggestion

> It's a prefixed capture + dedicated rendering to spit it out again (after execution in a ds finished).

what about 

```suggestion
            'stderr'); 'pass-through' to the screen (and thus absent from returned record); prefix with 'relpath'
            captured output (similar to like grep does) and write to stdout and stderr. In 'relpath', relative path
            is relative to the top of the dataset if `dataset` is specified, and if not - relative to current directory."""),
```

bpoldrack: Yes, fine with me.

bpoldrack:
Ok, I linked the failing archive test in the issue - don't think it's related. Let's proceed then. Thx, @yarikoptic!

yarikoptic-gitmate:

PR #7073, opened 2022-10-07
RF: remove unused code str-ing PurePath

yarikoptic:
Is of no use since we operate on path_ not path down below that block. Originally introduced in b0258db13236491c6e582a16ee4fc3129f240f87 (#3191).

Should be safe since RI which is coming next can take Path just fine:

	❯ python -c 'from pathlib import Path; from datalad.support.network import RI; print(repr(RI(Path("/home"))))'
	PathRI(path='/home')
	❯ python -c 'from pathlib import Path; from datalad.support.network import RI; print(repr(RI("/home")))'
	PathRI(path='/home')

### DIFF
--- a/changelog.d/pr-7073.md
+++ b/changelog.d/pr-7073.md
@@ -1,0 +1,5 @@
+### Bug Fixes
+
+- RF: remove unused code str-ing PurePath.  [PR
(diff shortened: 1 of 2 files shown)


bpoldrack (approved):

yarikoptic-gitmate:

PR #7076, opened 2022-10-07
datalad-deprecated moved to pytest

yarikoptic:


yarikoptic:
ok, what is relevant -- passes, so I will merge and will push merge into `master` as well to green up CI.

PR #7079, opened 2022-10-08
BF (regression fix): Do not yield "impossible" records by default while querying metadata

yarikoptic:
.... eh -- might not be needed/suboptimal since there is WiP https://github.com/datalad/datalad/pull/7001 but that one doesn't address neuroimaging fails since operates at "metadata" command level.  With this one i think neuroimaging would be "clean". But may be #7001 logic could/should be moved into `query_aggregated_metadata` instead @christian-monch ?

### DIFF
--- a/datalad/metadata/metadata.py
+++ b/datalad/metadata/metadata.py
@@ -1141,6 +1141,7 @@ def query_aggregated_metadata(reporton: str,
                               aps: List[Dict],
                               recursive: bool = False,
                               metadata_source: Optional[str] = None,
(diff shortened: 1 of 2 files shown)


yarikoptic:
note: https://github.com/datalad/datalad/pull/7001 by @christian-monch  would supersede it when merged

christian-monch:
> note: https://github.com/datalad/datalad/pull/7001 by @christian-monch  would supersede it when merged

@yarikoptic thx for the work and sorry for the delay in #7001 . Since #7001 is merged now, I will close this PR-

Issue #7078, opened 2022-10-07
test_get_subdataset_direct_fetch is flaky on OSX on appveyor


yarikoptic:
Failed in https://github.com/datalad/datalad/pull/7066#issuecomment-1268609480 and then later in other runs (con/tributing setup is waiting for github to get fixed, do not have recent fetches). Example from https://ci.appveyor.com/project/mih/datalad/builds/45005458/job/3v4wb1e356gxc3lq

```
test_get_subdataset_direct_fetch _______________________________________________________________________________________________________
[gw0] darwin -- Python 3.8.12 /Users/appveyor/venv3.8.12/bin/python
path = PosixPath('/Users/appveyor/DLTMP/datalad_temp_test_get_subdataset_direct_fetch5k3cq4nk')
    @skip_if_adjusted_branch
    @slow  # ~12s
    @skip_if_on_windows
    @skip_ssh
    @with_tempfile(mkdir=True)
    def test_get_subdataset_direct_fetch(path=None):
        path = Path(path)
        origin = Dataset(path / "origin").create()
        for sub in ["s0", "s1"]:
            sds = origin.create(origin.pathobj / sub)
            sds.repo.commit(msg="another commit", options=["--allow-empty"])
        origin.save()
        s0 = Dataset(origin.pathobj / "s0")
        s1 = Dataset(origin.pathobj / "s1")
        # Abandon the recorded commit so that it needs to be brought down by a
        # direct fetch.
        s0.repo.call_git(["reset", "--hard", "HEAD~"])
        s1.repo.call_git(["reset", "--hard", "HEAD~"])
    
        # Tweak the configuration of s0 to make the direct fetch fail.
        # Disallow direct oid fetch (default).
        s0.repo.config.set("uploadpack.allowAnySHA1InWant", "false",
                           scope="local")
        # Configure the fetcher to avoid v2, which allows fetching unadvertised
        # objects regardless of the value of uploadpack.allowAnySHA1InWant.
        s0.repo.config.set("protocol.version", "0", scope="local")
    
        # Configure s1 to succeed with direct fetch.
        s1.repo.config.set("uploadpack.allowAnySHA1InWant", "true",
                           scope="local")
    
        clone = install(
            str(path / "clone"),
            source="ssh://datalad-test:" + origin.repo.pathobj.as_posix())
    
        res = clone.get(["s0", "s1"], on_failure="ignore")
>       assert_result_count(res, 1,
                            action="install", type="dataset", status="error")
[...]

PR #7081, opened 2022-10-11
PR template: remove Changelog entries, refer to scriv

yarikoptic:
Since we are migrating to an automation for changelog management with scriv, suggesting to describe changelog entry in description is suboptimal

- adds yet another way to specify changelog entries (thus - confusion)
- those entries would not be used and would get forgotten
- unlike scriv entries in a file included in the diff they cannot be easily provided "suggestions" to
- duplicate listing of agreed upon sections for the 3rd time (in addition to scriv.ini and action to add changelog entries)

thus, following our discussion at meetup, this PR removes changelog template from the PR template.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,24 +1,10 @@
 ### PR checklist
 
-- [ ] Provide an overview of the changes you're making and explain why you're proposing them, ideally in the form of a changelog (template below)
-- [ ] Include `Fixes #NNN` somewhere in the description if this PR addresses an existing issue.
+- [ ] Provide an overview of the changes you're making and explain why you're proposing them, ideally in the form of a changelog (use `scriv create` or add `CHANGELOG-missing` label to generate one from title).
+- [ ] Include `Fixes #NNN` somewhere in the description and `scriv` changelog entry if this PR addresses an existing issue.
 - [ ] If this PR is not complete, select "Create Draft Pull Request" in the pull request button's menu.
   Consider using a task list (e.g., `- [ ] add tests ...`) to indicate remaining to-do items.
 - [ ] If you would like to list yourself as a DataLad contributor and your name is not mentioned please modify .zenodo.json file.
 - [ ] **Delete these instructions**. :-)
 
 Thanks for contributing!
-
-### Changelog
-#### 🐛 Bug Fixes
-- Description... Fixes #issue
-#### 💫 Enhancements and new features
--
-#### 🪓 Deprecations and removals
--
-#### 📝 Documentation
--
-#### 🏠 Internal
--
-#### 🛡 Tests
--
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,6 +1,8 @@
 ### PR checklist
 
-- [ ] Provide an overview of the changes you're making and explain why you're proposing them, ideally in the form of a changelog (use `scriv create` or add `CHANGELOG-missing` label to generate one from title).
+- [ ] Provide an overview of the changes you're making and explain why you're proposing them.
+- [ ] Create a changelog snippet (add the `CHANGELOG-missing` label to this pull request in order to have a snippet generated from its title;
+  or use `scriv create` locally and include the generated file in the pull request, see [scriv](https://scriv.readthedocs.io/)).
 - [ ] Include `Fixes #NNN` somewhere in the description and `scriv` changelog entry if this PR addresses an existing issue.
 - [ ] If this PR is not complete, select "Create Draft Pull Request" in the pull request button's menu.
   Consider using a task list (e.g., `- [ ] add tests ...`) to indicate remaining to-do items.


jwodder:
This should probably mention that, if `scriv create` is used, then the new fragment file should be renamed to a name of the form `pr-PRNUM.md` so that the release workflow can comment on it and take its version bump level into account.

yarikoptic:
oh, I thought that we are agnostic to the names. 
- Couldn't we just take all `changelog.d/*.md` files in the action? People who want to do `scriv create` "manually" would not know PR number until submitted and thus would just prefer to stick to default name it gives.
- I thought bump is based on the semver-labels of the PRs, but if I get it right those are used only to produce `scriv` changelog entry and not retrospected upon release then, correct?

jwodder:
@yarikoptic

* Do you want this as part of the current `add-changelog-snippet` action (with an adjusted trigger) or as a separate action?
* For "rename existing scriv fragment", should any scriv fragment not in the proper form be renamed, or should the Git history be inspected to see exactly what fragment the PR added?

yarikoptic:
> * Do you want this as part of the current `add-changelog-snippet` action (with an adjusted trigger) or as a separate action?

I thought to reuse that action

> For "rename existing scriv fragment", should any scriv fragment not in the proper form be renamed, or should the Git history be inspected to see exactly what fragment the PR added?

"git history" -- I thought of using `git diff` from the base branch to figure out what files were added, and only work on that file (should be single one, if multiple -- error out)

jwodder:
@yarikoptic 

> if CHANGELOG-missing label added, instead of creating new, just rename existing scriv fragment.

How strongly do we want this to be tied to the `CHANGELOG-missing` label?  Should it be configurable by the consumer of the action?  Note that the action already has an option for configuring what label(s) to remove after adding a changelog snippet, which is "recommended" to be set to `CHANGELOG-missing`; how should this interact with that option?

jwodder:
@yarikoptic Also, for detecting whether there's already a scriv-generated fragment, should the candidates just be any `.md` file in the fragment directory?  That would cause problems if someone wanted to, say, add a `README.md` file to the directory.

yarikoptic:
deprecatde -- known. appveyor too -- https://github.com/datalad/datalad/issues/7078 .

mih (approved):
Fine with me, but questionable in how far a new contributor would dive into the required tooling. But the threshold here is already so high that it doesn't matter much from my POV.

jwodder:
@yarikoptic Could you specify what you want the action to do with and without `CHANGELOG-missing` (or other configured label) in the following cases:
    * There are no changelog snippets in the PR
    * There is one changelog snippet in the PR, and its name is not in `pr-PRNUM.md` format
    * There are multiple changelog snippets in the PR

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,24 +1,10 @@
 ### PR checklist
 
-- [ ] Provide an overview of the changes you're making and explain why you're proposing them, ideally in the form of a changelog (template below)
-- [ ] Include `Fixes #NNN` somewhere in the description if this PR addresses an existing issue.
+- [ ] Provide an overview of the changes you're making and explain why you're proposing them, ideally in the form of a changelog (use `scriv create` or add `CHANGELOG-missing` label to generate one from title).
```
bpoldrack, line 3:
```suggestion
- [ ] Provide an overview of the changes you're making and explain why you're proposing them, ideally in the form of a changelog snippet (use `scriv create` or add the `CHANGELOG-missing` label to this pull request in order to generate one from its title).
```

yarikoptic:
@jwodder let's continue on that in a dedicated https://github.com/datalad/release-action/issues/9

bpoldrack (approved):
I think this is fine. Just trying to be a little clearer from the POV of a new contributor.

yarikoptic:
> Fine with me, but questionable in how far a new contributor would dive into the required tooling.

could be done by us by adding label, providing suggestion for the generated changelog, so to some degree actually making it better than just giving instructions on how to re-word description or going behind curtain and changing it ourselves.

bpoldrack:
@yarikoptic restarted appveyor, since I believe the issue you mentioned should be fixed by #7070

christian-monch:
Fine by me, I am happy to agree on one way and one way only. :-)

.github/PULL_REQUEST_TEMPLATE.md
mslw: Or, if we want to disassociate PR description and changelog entry, this could be:
```
- [ ] Provide an overview of the changes you're making and explain why you're proposing them
- [ ] Create a changelog snippet (use `scriv create` locally and include the generated file in the pull request, or add the `CHANGELOG-missing` label to this pull request in order to have a snippet generated from its title; see [scriv](https://scriv.readthedocs.io/)).
```
but the above is also fine.

bpoldrack: Agree. That would allow a "ask for forgiveness" approach by contributors, checking off the first one only and allow "us" to fix for the second if they didn't.

yarikoptic: I like @mslw proposed separation, and reordered to list label first so then scriv ref comes right on

yarikoptic:
Since there is agreement in general, let's proceed and if anything adjust more in a follow up PR.

PR #7082, opened 2022-10-12
Update GitHub Actions action versions

jwodder:
We are currently using older versions of the updated GitHub Actions actions, and these older versions still use Node 12, [which is deprecated](https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/).

This PR also configures Dependabot to create PRs updating workflows whenever a new version of a used action is released, as discussed in https://github.com/datalad/datalad-extensions/pull/105.

### DIFF
--- a/changelog.d/pr-7082.md
+++ b/changelog.d/pr-7082.md
@@ -1,0 +1,5 @@
+### Internal
+
+- Update GitHub Actions action versions.  [PR
+  #7082](https://github.com/datalad/datalad/pull/7082) (by
+  [@jwodder](https://github.com/jwodder))
--- a/changelog.d/pr-7082.md
+++ b/changelog.d/pr-7082.md
@@ -1,5 +1,5 @@
 ### Internal
 
-- Update GitHub Actions action versions.  [PR
-  #7082](https://github.com/datalad/datalad/pull/7082) (by
+- Update GitHub Actions action versions.
+  [PR #7082](https://github.com/datalad/datalad/pull/7082) (by
   [@jwodder](https://github.com/jwodder))
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: weekly
+    commit-message:
+      prefix: "[gh-actions]"
+      include: scope
+    target-branch: maint
+    labels:
+      - semver-internal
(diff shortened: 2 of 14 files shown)


changelog.d/pr-7082.md
```diff
@@ -0,0 +1,5 @@
+### Internal
+
+- Update GitHub Actions action versions.  [PR
+  #7082](https://github.com/datalad/datalad/pull/7082) (by
+  [@jwodder](https://github.com/jwodder))
```
yarikoptic, line 5:
```suggestion
- Update GitHub Actions action versions.
  [PR #7082](https://github.com/datalad/datalad/pull/7082) (by
  [@jwodder](https://github.com/jwodder))
```

yarikoptic:
hm, I dewarped the changelog snippet and then looked back. benchmarks are failing with something which might actually relate:

```
fatal: HEAD^1: not a valid SHA1
ram [7110660]: 
Error: Process completed with exit code 128.
```

rerunning : https://github.com/datalad/datalad/actions/runs/3236005986

yarikoptic:
seems to be consistently failing, @jwodder please check what could be the difference between two versions of checkout which would affect it

jwodder:
@yarikoptic The Benchmark workflow was previously using v1 of actions/checkout, which checks out the complete history by default, yet v2 and up only check out a single commit by default.  Setting `fetch-depth: 0` should fix it.

yarikoptic:
> @yarikoptic The Benchmark workflow was previously using v1 of actions/checkout, which checks out the complete history by default, yet v2 and up only check out a single commit by default. Setting `fetch-depth: 0` should fix it.

oh... then it is something to keep for **every** installation of any module which uses `versioneer`, `versioningit` and alike or version would not be deducable.  That lead to tests failure  on OSX, and it seems even more "fun" situation on testing extensions which then lead to "reinstallation" of datalad from pypi (instead of already reusing the one from checkout) while installing extension: from https://github.com/datalad/datalad/actions/runs/3237400351/jobs/5304414220

```
2022-10-12T19:51:20.2161287Z Collecting datalad>=0.17.6
2022-10-12T19:51:20.2447327Z   Downloading datalad-0.17.6-py3-none-any.whl (1.4 MB)
```
so I added to those and other workflows similar depth setting now and pushed.  I added even to workflows where it is not exactly needed (e.g. crippled fs - we do not run tests on version to be ok, or disabled windows) so we do not need to figure it out later. 
I guess it is something to keep attention to (well, I have merged most of your PRs by now ;))

yarikoptic:
I have checked logs of one of the extensions and datalad reported to be 
```
version: 0.17.6+62.g0d3f8e8ed
```
so all good now, let's proceed, but now with a better awareness of change in semantic of new version of checkout

yarikoptic-gitmate:

PR #7085, opened 2022-10-13
BF: Missing import in clone

bpoldrack:
Recent refactoring of `clone` (c57de3885cdbad488803241b63ab8d6b6e0738ec part of #7017) moved functions to `clone_utils.py` and forgot to import all of them.
Similarly, the introduction of `clone_ria.py` came with the same issue, as functions are moved. This led to breaking extension (datalad-next) importing from clone. (See https://github.com/datalad/datalad-next/issues/111)
    
This patch makes all those functions importable from clone again.
    
Note, that after this is released, datalad-next can drop its clone_ria entirely, since it's part of core now.
    
I don't think this needs a changelog, since it simply fixes https://github.com/datalad/datalad/commit/c57de3885cdbad488803241b63ab8d6b6e0738ec, which isn't released yet.


bpoldrack:
Note, that I'm not addressing the two "complexity issues" pointed to by code climate. That's b/c the move of ria functions into `clone_utils` is temporary anyway. We need this to have the extension CI running without requiring datalad-next to do anything. However, after this is released, the next move would be to move the RIA functionality out of core anyway (thus resolving the "complexity" in here).

datalad/core/distributed/clone.py
```diff
@@ -54,9 +54,15 @@
...
+    _get_tracking_source,
+    _map_urls,
     _test_existing_clone_target,
     _try_clone_candidates,
     decode_source_spec,
+    # RIA imports needed b/c datalad-next imports it from here ATM;
+    # Remove after core was released and next dropped the ria patch.
+    postclone_preannex_cfg_ria,
+    postclonecfg_ria,
 )
```
yarikoptic, line 66: wouldn't it be sufficient to adjust import to import here also from clone_ria instead of moving those RIA related functions into `clone_utils`?

yarikoptic:
FWIW I pushed fixes for linting to `maint`, merged `maint` (0.17.7-7-gc31cb2c3c) into master and pushed as well. So lint'ing is green on master now: https://github.com/datalad/datalad/actions/runs/3268501454

datalad/core/distributed/clone.py
bpoldrack: Nope - tried that first. That leads to a circular import.

PR #7087, opened 2022-10-13
Merge maint into master

yarikoptic:
all kinds of goodness from maint.  Testing on CI, expect failure of -deprecated (see https://github.com/datalad/datalad/pull/7002)


yarikoptic:
so it is only deprecated and next ... hm - next was failing 10 days ago and there is #7085 for it , deprecated is also known https://github.com/datalad/datalad-deprecated/issues/69 ... so let's proceed... hopefully someone gets to fix deprecated

PR #7093, opened 2022-10-17
[BF] Prevent double yielding of impossible get result

### LINKED ISSUE #5537: Duplicate `get` error result (reporting)
Get a file with invalid filename (extension missing):

```sh
$ datalad clone git@github.com:datalad-datasets/human-connectome-project-openaccess.git hcp
$ cd hcp
$ datalad -f json_pp get HCP1200/705341/MNINonLinear/Results/rfMRI_REST1_LR/rfMRI_REST1_LR_hp2000_clean.nii
# and again when the subdatasets are present
$ datalad -f json_pp get HCP1200/705341/MNINonLinear/Results/rfMRI_REST1_LR/rfMRI_REST1_LR_hp2000_clean.nii
{
  "action": "get",
  "message": "path does not exist",
  "path": "/tmp/hcp/HCP1200/705341/MNINonLinear/Results/rfMRI_REST1_LR/rfMRI_REST1_LR_hp2000_clean.nii",
  "refds": null,
  "status": "impossible"
}
{
  "action": "get",
  "message": "path does not exist",
  "path": "/tmp/hcp/HCP1200/705341/MNINonLinear/Results/rfMRI_REST1_LR/rfMRI_REST1_LR_hp2000_clean.nii",
  "refds": null,
  "status": "impossible"
}
```

Unclear why a second attempt is made, when first attempt finds "path does not exist".

jsheunis:
Fixes #5537.

This PR prevents double yielding of an impossible get result on a path that does not exist by keeping track of impossible get results and skipping additional yields inside `_install_targetpath()`.

The issue reported this problem occurring twice:
- when called from a superdataset _without_ interim subdatasets in the problematic path installed: this is addressed by changing the `bottomup` flag to `True` in the `Subdatasets.__call__()`.
- when called from a superdataset _with_ interim subdatasets in the problematic path already installed: this addressed by tracking the paths for which `_install_targetpath()` has already been called, using the new helper function `_check_error_reported()` and the tracking dict `error_reported`

These changes were rerun locally using the example provided in the issue, and succeeded to give single impossible results in both cases. Tests all succeeded locally with `python -m pytest -s -v --pyargs datalad/core/distributed datalad/distribution datalad/distributed`

#### 🐛 Bug Fixes
- Prevent double yielding of `impossible` `get` result for a path that does not exist. Fixes #5537

### DIFF
--- a/changelog.d/pr-7093.md
+++ b/changelog.d/pr-7093.md
@@ -1,5 +1,6 @@
 ### Bug Fixes
 
-- [BF] Prevent double yielding of impossible get result.  [PR
-  #7093](https://github.com/datalad/datalad/pull/7093) (by
+- [BF] Prevent double yielding of impossible get result 
+  Fixes [#5537](https://github.com/datalad/datalad/issues/5537).
+  [PR #7093](https://github.com/datalad/datalad/pull/7093) (by
   [@jsheunis](https://github.com/jsheunis))
--- a/changelog.d/pr-7093.md
+++ b/changelog.d/pr-7093.md
@@ -1,0 +1,5 @@
+### Bug Fixes
+
+- [BF] Prevent double yielding of impossible get result.  [PR
+  #7093](https://github.com/datalad/datalad/pull/7093) (by
+  [@jsheunis](https://github.com/jsheunis))
--- a/datalad/distribution/get.py
+++ b/datalad/distribution/get.py
@@ -890,9 +890,8 @@ class Get(Interface):
                 # maintain path argument semantics and pass in dataset arg
                 # as is
                 dataset=dataset,
-                # from the bottom prevents duplicate yielding of results
-                # for a dataset where subdatasets are not installed yet
-                bottomup=True,
+                # always come from the top to get sensible generator behavior
+                bottomup=False,
                 # when paths are given, they will constrain the recursion
                 # automatically, and we need to enable recursion so we can
                 # location path in subdatasets several levels down
--- a/datalad/distribution/get.py
+++ b/datalad/distribution/get.py
@@ -705,6 +705,23 @@ def _get_targetpaths(ds, content, refds_path, source, jobs):
         yield r
 
 
+def _check_error_reported(res: dict, error_dict: dict):
+    # Helper to check if an impossible result for a path that does
+    # not exist has already been yielded before. If not, add path
+    # to the error_dict.
+    if res.get('action', None) == 'get' and \
+        res.get('status', None) == 'impossible' and \
+        res.get('message', None) == 'path does not exist':
+            non_existing_path = res.get('path', None)
+            if non_existing_path not in error_dict.keys():
+                # if path not in dict, add it
+                error_dict[non_existing_path] = True
+                return False
+            else:
+                return True
+    return False
+
+
 @build_doc
 class Get(Interface):
     """Get any dataset content (files/directories/subdatasets).
@@ -850,6 +867,7 @@ class Get(Interface):
             reckless=None,
             jobs='auto',
     ):
+        
         if not (dataset or path):
             raise InsufficientArgumentsError(
                 "Neither dataset nor target path(s) provided")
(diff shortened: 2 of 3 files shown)


datalad/distribution/get.py
```diff
@@ -871,7 +891,7 @@ def __call__(
                 # as is
                 dataset=dataset,
                 # always come from the top to get sensible generator behavior
```
yarikoptic, line 893: so this change changes this behavior:
- I wonder why top-down was considered to be more "sensible" (may be smth in commit which added)
- comment should be adjusted according to the change if this change is needed (seems to be orthogonal to the actual fix described)

datalad/distribution/get.py
```diff
@@ -705,6 +705,23 @@ def _get_targetpaths(ds, content, refds_path, source, jobs):
         yield r
 
 
+def _check_error_reported(res: dict, error_dict: dict):
```
yarikoptic, line 708:
```suggestion
def _check_error_reported_before(res: dict, error_dict: dict):
```
(and change invocation below) to make it a bit clear(er) on either it was reported before or just now (and thus possibly having the opposite bool meaning)

yarikoptic (changes requested):
left some suggestions but I think it would warrant a test (wasn't some test triggering it already? just add pdb at `return True` and see if some test hits it, then add a test there for this fix.  If not a single test was hitting it -- I think it is worthwhile adding such a use case if we "value" this fix. Otherwise it would be tricky to see why it was added in retrospective.

changelog.d/pr-7093.md
```diff
@@ -0,0 +1,5 @@
+### Bug Fixes
+
+- [BF] Prevent double yielding of impossible get result.  [PR
+  #7093](https://github.com/datalad/datalad/pull/7093) (by
+  [@jsheunis](https://github.com/jsheunis))
```
yarikoptic, line 5:
```suggestion
- [BF] Prevent double yielding of impossible get result 
  Fixes [#5537](https://github.com/datalad/datalad/issues/5537).
  [PR #7093](https://github.com/datalad/datalad/pull/7093) (by
  [@jsheunis](https://github.com/jsheunis))
```

jsheunis:
Thanks for the review @yarikoptic. Responding individually to comments.

datalad/distribution/get.py
jsheunis: Thanks for picking this up, I missed the comment and will update

datalad/distribution/get.py
```diff
@@ -921,6 +941,9 @@ def __call__(
                             # all those messages on not having installed anything
                             # are a bit pointless
                             # "notneeded" for annex get comes below
+                            # prevent double yielding of impossible result
+                            if _check_error_reported(res, error_reported):
```
jsheunis, line 945:
```suggestion
                            if _check_error_reported_before(res, error_reported):
```

datalad/distribution/get.py
```diff
@@ -958,6 +981,9 @@ def __call__(
                     # installed by explorative installation to get to target
                     # paths, prior in this loop
                     if res.get('status', None) != 'notneeded' or not known_ds:
+                        # prevent double yielding of impossible result
+                        if _check_error_reported(res, error_reported):
```
jsheunis, line 985:
```suggestion
                        if _check_error_reported_before(res, error_reported):
```

bpoldrack:
@jsheunis, @yarikoptic: I added a test (fails on `maint` w/o this PR) and reverted the change to `bottomup`. Failing test in `datalad-deprecated` showed, that it was misguided.

yarikoptic:
hm, travis stalled with
```
[gw0] [ 98%] PASSED ../datalad/ui/tests/test_dialog.py::test_message_pbar_state_logging_is_demoted 
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received
The build has been terminated
```
in the first matrix run. For some reason travis refuses to log me in atm to restart, can't waste time on that . Hard to say where it stalled since we run with `-n 2` thus in parallel.  But it is odd since I do not think we saw such stalls before :-/

yarikoptic:
I don't see how this fix could relate to stalling, I found many of other stalled runs in the historical logs, didn't investigate, let's proceed!

yarikoptic-gitmate:

Issue #6967, opened 2022-08-21
create-sibling-ria --existing reconfigure --shared does not rerun `git-init`


mih:
Similar to https://github.com/datalad/datalad/issues/6951 running such a command will not actually do anything about a provided `--shared` flag.

I would have expected that it reruns `git init --shared` on the remote end, but nothing like that happens.

When I use the command with no pre-existing repo at the remote, I get the expected result.

bpoldrack:
PR #7095 fixes, that for local stores the re-initialization of the repo when `existing=reconfigure` wasn't executed (but was for SSH).

As described in that PR's description there's a remaining issue with that:

`git init --bare --shared=...` does not seem to change the permissions for existing files. So, the result of that call in an empty dir vs the result of two subsequent `git init` (one w/o the `shared` option and a second one with it) are not identical.

PR #7095, opened 2022-10-17
Make sure to re-initialise git repo

mslw:
The `create_sibling_ria` command used `GitRepo(repo_path, create=True)` constructor to initialise bare repositories locally, also when `existing=reconfigure` was requested. But for existing repositories, `GitRepo` does not execute `git init`, it just returns a `GitRepo` instance, so `reconfigure` was not actually handled. This commit introduces an aditional check, to make sure that `git init` is called when reconfiguring.

Partially addresses #6967 (`--existing reconfigure --shared` not working). However, another problem remains: although we now do handle the reconfiguration, calling `git annex init --bare --shared` sets the config variable `core.sharedRepository` (which would presumably affect future commits), but does not change the permissions for the already existing files in the bare git repo.

### DIFF
--- a/datalad/distributed/create_sibling_ria.py
+++ b/datalad/distributed/create_sibling_ria.py
@@ -684,10 +684,7 @@ def _create_sibling_ria(
             # skip & error have been handled at this point
             gr.init(
                 sanity_checks=False,
-                init_options=[
-                    "--bare",
-                    "--shared={}".format(shared) if shared else "",
-                ]
+                init_options=["--bare"] + ([f"--shared={shared}"] if shared else []),
             )
         if storage_sibling:
             # write special remote's uuid into git-config, so clone can
--- a/datalad/distributed/create_sibling_ria.py
+++ b/datalad/distributed/create_sibling_ria.py
@@ -518,7 +518,8 @@ def _create_sibling_ria(
             use_remote_annex_bundle=False)
         ssh.open()
 
-    if existing in ['skip', 'error']:
+    exists = False
+    if existing in ['skip', 'error', 'reconfigure']:
         config_path = repo_path / 'config'
         # No .git -- if it's an existing repo in a RIA store it should be a
         # bare repo.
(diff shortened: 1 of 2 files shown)


datalad/distributed/create_sibling_ria.py
```diff
@@ -674,6 +678,17 @@ def _create_sibling_ria(
...
                      shared=shared if shared else None)
+        if exists and existing == 'reconfigure':
+            # if the repo exists at the given path, the GitRepo would not
+            # (re)-run git init, and just return an instance of GitRepo;
+            # skip & error have been handled at this point
+            gr.init(
+                sanity_checks=False,
+                init_options=[
+                    "--bare",
+                    "--shared={}".format(shared) if shared else "",
```
yarikoptic, line 689: f-string can simply be used
```suggestion
                    f"--shared={shared}" if shared else "",
```

datalad/distributed/create_sibling_ria.py
```diff
@@ -674,6 +678,17 @@ def _create_sibling_ria(
...
+                ]
```
yarikoptic, line 690: not sure if empty string in init_options list is desired and would not lead to side effects, so better 
```suggestion
                init_options=["--bare"] + ([f"--shared={shared}"] if shared else []),
```

yarikoptic (commented):
don't know much about `ria` so can't judge if `git init` is safe to do upon reconfigure -- left just a single recommendation

datalad/distributed/create_sibling_ria.py
mslw: Nice, this is clever!

datalad/distributed/create_sibling_ria.py
mslw: Thanks, applying suggestion from below which also covers this bit.

bpoldrack:
> don't know much about ria so can't judge if git init is safe to do

FTR: "safe" doesn't even matter. Re-init is the explicitly described behavior of `existing=reconfigure` and implemented in the SSH case, it just wasn't done in the local case.

bpoldrack (approved):

bpoldrack:
Failure in datalad-deprecated happens on `maint` already - not related to this PR.

Merging.

PR #7102, opened 2022-10-21
RF `datalad.interface.utils.eval_results()`

mih:
This is an essential function that any command execution goes through, which is responsible for result generation and handling. This change refactors the code by

- [x] Moving the internal `generator_func()` helper into a regular function with documentation and type annotation
- [x] Established a no-op call to a new `_validate_cmd_call()` function that can perform universal validation of a command parameterization before a command execution is attempted

This second aspect is the main motivation for this change. It lays the foundation for homogeneous validation of parameters beyond the CLI-specific (partial) validation and the custom, per-command validation that is done after a command execution has started.

The work on such a validation has started in https://github.com/datalad/datalad-next/pull/116

### DIFF
--- a/datalad/interface/utils.py
+++ b/datalad/interface/utils.py
@@ -295,24 +295,6 @@ def eval_results(wrapped):
         # short cuts and configured setup for common options
         return_type = common_params['return_type']
 
-        # query cfg for defaults
-        # .is_installed and .config can be costly, so ensure we do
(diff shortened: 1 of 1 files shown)


bpoldrack (approved):
LGTM.

Appveyor is once again `test_gracefull_death`, which is not likely to be related.
Provided that Travis passes, good to go from my POV.

PR #7103, opened 2022-10-21
Report repository name in `create_sibling_ghlike` dry run

mslw:
This PR adds a different error message when `create_sibling_github` is used with `dry_run`. The new error message  actually gives repository names.

The docstring for `create_sibling_ghlike` `dry_run` option said:
```
if set, no repository will be created, only tests for
name collisions will be performed, and would-be repository names
are reported for all relevant datasets
```
However, it behaved like this (dataset with one subdataset):
```
❱ datalad create-sibling-gin --recursive --dry-run -d . ProjectX
create_sibling_gin [dry-run](ok): [sibling repository 'gin' created at None]
create_sibling_gin [dry-run](ok): [sibling repository 'gin' created at None]
action summary:
  create_sibling_gin [dry-run] (ok: 2)
```

Now it will behave like this:
```
❱ datalad create-sibling-gin --recursive --dry-run -d . ProjectX
create_sibling_gin [dry-run](ok): [would create sibling 'gin' and repository 'ProjectX']
create_sibling_gin [dry-run](ok): [would create sibling 'gin' and repository 'ProjectX-bar']
action summary:
  create_sibling_gin [dry-run] (ok: 2)
```

Moreover, the docstring wasn't clear on what name collision tests will be performed (sibling names vs repository names). The `dry_run` mode, in fact, does not make any requests to the hosting service, so it would only check for the former.

Because adding checks for repository names would be more complicated (need to process the response, and then adjust the message depending on what we chose to reconfigure), I assumed that the current behavior is intended and only adjusted the docstring.

### DIFF
--- a/datalad/distributed/create_sibling_ghlike.py
+++ b/datalad/distributed/create_sibling_ghlike.py
@@ -467,7 +467,6 @@ class _GitHubLike(object):
         headers = self.request_headers
 
         if dry_run:
-            res_msg = "would create {} sibling and repository named {}"
             return dict(
                 status='ok',
                 request_url=endpoint,
--- a/datalad/distributed/create_sibling_ghlike.py
+++ b/datalad/distributed/create_sibling_ghlike.py
@@ -134,8 +134,8 @@ class _GitHubLike(object):
             args=("--dry-run",),
             action="store_true",
             doc="""if set, no repository will be created, only tests for
-            name collisions will be performed, and would-be repository names
-            are reported for all relevant datasets"""),
+            sibling name collisions will be performed, and would-be repository
+            names are reported for all relevant datasets"""),
     )
 
     def __init__(self, url, credential, require_token=True, token_info=None):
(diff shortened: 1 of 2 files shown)


datalad/distributed/create_sibling_ghlike.py
```diff
@@ -458,6 +467,7 @@ def repo_create_request(self, reponame, organization, private,
         headers = self.request_headers
 
         if dry_run:
+            res_msg = "would create {} sibling and repository named {}"
```
bpoldrack, line 470: I am a bit confused. Where is `res_msg` used?

bpoldrack (commented):
Thx, @mslw! Agree with the intend, but if I'm not mistaken there's something missing in the diff (see below).

datalad/distributed/create_sibling_ghlike.py
mslw: Oops, well spotted! I accidentally left and committed part of my first attempt at the message. This line should go away.
```suggestion
```

bpoldrack (approved):

yarikoptic-gitmate:

PR #7107, opened 2022-10-24
Fix `PWD` environment entry if `cwd` provided to `Runner()` or `Runner.run()` is  a `Path`-object

### LINKED ISSUE #7040: Runner cannot handle `run(cwd=Path)`
Docs say: `cwd : path-like, optional`

Made me assume that a `Path` instance would work, but it does not. At least not when passed to `run()`. It does work, when passed to the constructor (same docs of the argument there).

On Windows, such a path obj ends up in the environment, leading to unexpected results.

christian-monch:
Fixes #7040 

This PR fixes #7040. It ensures that an environment that is provided to the subprocess contains a proper entry for `PWD`. That means, the entry contains a string that describes the path of the working directory, and not a python-representation of a Path object.

### DIFF
--- a/changelog.d/pr-7107.md
+++ b/changelog.d/pr-7107.md
@@ -1,0 +1,8 @@
+### Bug Fixes
+
+- Ensure subprocess environments have a valid path in `os.environ['PWD']`,
+  even if a Path-like object was given to the runner on subprocess creation
+  or invocation.
+  Fixes [#7040](https://github.com/datalad/datalad/issues/7040) via
+  [PR #7107](https://github.com/datalad/datalad/pull/7107)
+  (by [@christian-monch](https://github.com/christian-monch))
--- a/datalad/runner/tests/test_witless_runner.py
+++ b/datalad/runner/tests/test_witless_runner.py
@@ -343,3 +343,41 @@ def test_environment(temp_dir_path=None):
     results = runner.run(cmd=cmd, protocol=StdOutCapture)
     output = results['stdout'].splitlines()[0]
     assert output == temp_dir_path
+
+
+def test_argument_priority():
+    class X:
+        def __init__(self, *args, **kwargs):
+            self.args = args
+            self.kwargs = kwargs
+
+        def run(self):
+            return dict(
+                code=0,
+                args=self.args,
+                kwargs=self.kwargs,
+            )
+
+    test_path_1 = "a/b/c"
+    test_env_1 = dict(source="constructor")
+    test_path_2 = "d/e/f"
+    test_env_2 = dict(source="run-method")
+
+    with unittest.mock.patch('datalad.runner.runner.ThreadedRunner') as tr_mock:
+
+        tr_mock.side_effect = X
+        runner = Runner(cwd=test_path_1, env=test_env_1)
+
+        result = runner.run("first-command")
+        assert result['kwargs']['cwd'] == test_path_1
+        assert result['kwargs']['env'] == {
+            **test_env_1,
+            'PWD': test_path_1
+        }
+
+        result = runner.run("second-command", cwd=test_path_2, env=test_env_2)
+        assert result['kwargs']['cwd'] == test_path_2
+        assert result['kwargs']['env'] == {
+            **test_env_2,
+            'PWD': test_path_2
+        }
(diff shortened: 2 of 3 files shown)


datalad/runner/tests/test_witless_runner.py
```diff
@@ -311,3 +311,34 @@ def test_too_long():
...
+    runner = Runner()
+    adjusted_env = runner._get_adjusted_env(
+        cwd=Path("/a/b/c"),
+        env=dict(some_key="value")
+    )
+    assert "/a/b/c" == adjusted_env['PWD']
+
+
+@with_tempfile(mkdir=True)
+def test_environment(temp_dir_path=None):
```
bpoldrack, line 328: This test makes me think we have not covered the case of passing cwd to the constructor and (a different one) to the method. We probably should test that behavior, too.

changelog.d/pr-7107.md
```diff
@@ -0,0 +1,65 @@
+<!-- Uncomment the section that is right (remove the HTML comment wrapper).-->
```
bpoldrack, line 1: Would be nice to delete the out commented stuff, since this is going to pile up in the repo over time.

bpoldrack (approved):
Thx, @christian-monch. LGTM.

I have two suggestions, though.

datalad/runner/tests/test_witless_runner.py
christian-monch: Both are tested. Line 336 -337 test passing to method, line 341-342 test passing to the constructor. Or did I missunderstand you?

changelog.d/pr-7107.md
christian-monch: Good idea.

christian-monch: done in df3bb1dba316b32904aa48e0185370a76f7d86e1

datalad/runner/tests/test_witless_runner.py
bpoldrack: Yes, you did. We do not test what happens if the caller sets (possibly different) `cwd` via both - constructor *and* method. This is only testing the case where the caller does either one of those.

bpoldrack: In code: `Runner(cwd="one/path").run(cwd="another/path")` - what happens?

bpoldrack:
One more thing, @christian-monch: Looks like the command quoting in the test (`py2cmd( ...`) is wrong for windows. We do have a helper for that somewhere (trying to find).

datalad/runner/tests/test_witless_runner.py
christian-monch: The method provided variables take precedence over the constructor provided. Don't ask me what the original authors intended with this behavior, it might be suited to reusing "pre-configured" runners multiple times with minimal source code 

> One more thing, @christian-monch: Looks like the command quoting in the test (`py2cmd( ...`) is wrong for windows. We do have a helper for that somewhere (trying to find).

`py2cmd` is the helper. I copied it from an older version of `datalad.tests.test_witless_runner.py`, and just added the additional argumnets, which are not used in this context. It works on some windows environments, and not on others. I think the reason might be in the selection of string-delimiters within the python code-string. It seems that `Popen` on some windows environments uses `"` to delimit the arguments when they are passed down to the system. That conflicts with using them in the python code (the error message does not reflect the string that is passed down, this would only contain one double-quote around `PWD`, i.e. `["PWD"]`).

So I will leave `py2cmd` in its original state until we figure out what goes on.

christian-monch:
> One more thing, @christian-monch: Looks like the command quoting in the test (`py2cmd( ...`) is wrong for windows. We do have a helper for that somewhere (trying to find).

`py2cmd` is the helper. I copied it from an older version of `datalad.tests.test_witless_runner.py`, and just added the additional argumnets, which are not used in this context. It works on some windows environments, and not on others. I think the reason might be in the selection of string-delimiters within the python code-string. It seems that `Popen` on some windows environments uses `"` to delimit the arguments when they are passed down to the system. That conflicts with using them in the python code (the error message does not reflect the string that is passed down, but a garbled version. The passed down string only contains one double-quote around `PWD`, i.e. `["PWD"]`).

bpoldrack:
> I pushed an update that uses ' within the python code string. That should fix the issue on CI.
> I will not change py2cmd before we figure out what exactly is going on on Windows.

Agreed. Thx!

christian-monch:
Amended the PR with a test regarding the parameter priorities.

bpoldrack:
Thanks, @christian-monch!

yarikoptic-gitmate:

PR #7123, opened 2022-10-28
Fix RIA file URL reporting in exception handling

adswa:
Probing for the existence of a ria_layout_version file internally misconstructs a ``target_ri`` variable only used for reporting the path to the file. We noticed recently that this reporting was off, duplicating paths, e.g.:

```
[INFO   ] RIA store unavailable. 
-caused by- 
file:///tmp/myriastore/tmp/myriastore/ria-layout-version not found. 
-caused by- 
[Errno 2] No such file or directory: '/tmp/myriastore/ria-layout-version'
```
This duplication should be fixed with this small change. It extracts the file name (``ria_layout_version``) and appends it to the ria store location, instead of appending the entire path to it.

### DIFF
--- a/datalad/distributed/ora_remote.py
+++ b/datalad/distributed/ora_remote.py
@@ -1058,8 +1058,9 @@ class RIARemote(SpecialRemote):
         """ Get version and config flags from remote file
         """
         if self.ria_store_url:
-            # construct path to ria_layout_version file in store for reporting
-            target_ri = self.ria_store_url[4:] + '/' + path.name
+            # construct path to ria_layout_version file for reporting
+            target_ri = self.ria_store_url[4:] +\
+                        path.relative_to(Path(self.store_base_path)).as_posix()
         elif self.storage_host:
             target_ri = "ssh://{}{}".format(self.storage_host, path.as_posix())
         else:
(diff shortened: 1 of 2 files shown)


datalad/distributed/ora_remote.py
```diff
@@ -1057,9 +1057,9 @@ def _verify_config(self, gitdir, fail_noid=True):
     def _get_version_config(self, path):
         """ Get version and config flags from remote file
         """
-
         if self.ria_store_url:
-            target_ri = self.ria_store_url[4:] + path.as_posix()
+            # construct path to ria_layout_version file in store for reporting
+            target_ri = self.ria_store_url[4:] + '/' + path.name
```
bpoldrack, line 1062: This is insufficient, since it is not accounting for different locations of the config file. By default there's the `ria-layout-version` at the store level and one per dataset.
What is needed is: Turning `path` into a relative path to the path-part of `self.ria_store_url`. Note, that your assumption that `Path.relative_to` is relatively recent in `pathlib` is wrong, @adswa - you likely looked at `is_relative_to` instead.

bpoldrack (changes requested):
Thx, for taking it on, @adswa. Spot is right, solution not quite yet - see comment in diff.

datalad/distributed/ora_remote.py
adswa: indeed, that may have been the case. I pushed an alternative, relying on ``self.store_base_path`` - could you check if that approach is better?

bpoldrack:
Thx much, @adswa

bpoldrack (approved):

yarikoptic-gitmate:

yarikoptic-gitmate:

PR #7125, opened 2022-10-31
Unrestrict pytest and pytest-cov versions

jwodder:
pytest-cov was being restricted to an old version that produces warnings when combined with the latest version of pytest.  At the same time, it's doubtful that restricting the maximum version of pytest will ever help.


yarikoptic-gitmate:

yarikoptic-gitmate:

PR #7118, opened 2022-10-25
Fix all logging to use %-interpolation and not .format, sort imports in touched files, add pylint-ing for % formatting in log messages to `tox -e lint`

yarikoptic:
#7117 made me look into tightening our automated linting.  Adding pylinting for a single warning/error to start with:

    W1202: Use lazy % or % formatting in logging functions (logging-format-interpolation)

which I believe nobody would doubt to be a good thing to do.
- [ ] any objections?
- [x] fix the code
- later (other PRs): enable more checks while at it or may be postpone to follow up dedicated PRs, but worth listing here may be (I was looking for the one in #7117 like *use of undefined variable* but `pylint` didn't warn about that :-/ )

[appveyor skip]
[travis skip]

edit: had to cancel appveyor manually -- for some reason it ignored the annotation I gave

Issue #7116, opened 2022-10-25
Fresh failures in datalad/git-annex tests -- likely due to new git


yarikoptic:
Was looking specifically at `test_having_annex` which  fails ATM on Windows and OSX (not Ubuntu). OSX has 2.38.1 git. This seems to be the first time it started to fail and you can see that it has newer git than ubuntu (using 2.30.2 while system has 2.38.1 as well) where it doesn't fail:

```shell
(git)smaug:/mnt/datasets/datalad/ci/git-annex/builds/2022/10[master]git
$> grep -e 'Versions:' -e 'FAILED ' cron-20221022/build-macos.yaml-877-b1412209-failed/1_test-datalad\ \(master\).txt
2022-10-22T07:27:41.2638160Z PASSEDVersions: annexremote=1.6.0 boto=2.49.0 cmd:7z=17.04 cmd:annex=10.20221004-gcde2e6110 cmd:bundled-git=UNKNOWN cmd:git=2.38.1 cmd:ssh=8.1p1 cmd:system-git=2.38.1 cmd:system-ssh=8.1p1 datalad=0.17.7+83.g234632671 humanize=4.4.0 iso8601=1.1.0 keyring=23.9.3 keyrings.alt=UNKNOWN msgpack=1.0.4 patoolib=1.12 platformdirs=2.5.2 requests=2.28.1 tqdm=4.64.1
2022-10-22T07:27:41.2794250Z FAILED ../datalad/tests/test_testrepos.py::test_having_annex - datalad.runner...
2022-10-22T07:27:41.2794760Z FAILED ../datalad/tests/test_testrepos.py::test_clone - datalad.runner.except...
2022-10-22T07:27:41.2795250Z FAILED ../datalad/tests/test_tests_utils_pytest.py::test_with_testrepos - dat...

$> grep -e 'Versions:' -e 'FAILED ' cron-20221022/build-ubuntu.yaml-882-b1412209-success/1_test-datalad\ \(master\).txt 
[...]

PR #7126, opened 2022-10-31
XFAIL a check on length of results in test_gracefull_death

yarikoptic:
Still happens in some busy environments, like recently within https://app.travis-ci.com/github/datalad/datalad/jobs/586725682 of https://github.com/datalad/datalad/pull/7118

Attn @bpoldrack who was after all the flaky tests

### DIFF
--- a/changelog.d/pr-7126.md
+++ b/changelog.d/pr-7126.md
@@ -1,0 +1,3 @@
+### 🧪 Tests
+
+- XFAIL a check on length of results in test_gracefull_death.  [PR #7126](https://github.com/datalad/datalad/pull/7126) (by [@yarikoptic](https://github.com/yarikoptic))
--- a/datalad/support/tests/test_parallel.py
+++ b/datalad/support/tests/test_parallel.py
@@ -14,6 +14,8 @@ from time import (
     time,
 )
 
+import pytest
+
 # logging effects threading and causes some 'weak' tests to fail,
 # so we will just skip those (well, if happens again -- disable altogether)
 from datalad import lgr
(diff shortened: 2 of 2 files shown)


bpoldrack (approved):
I'm kinda fine with this approach. Hence the approval.

However, I see little utility in having the test executed in the first place. Nobody was able/willing to figure it out while it was making CI fail, I predict nobody will care about an XFAIL either. So, whatever the outcome of the test - there will be no consequences. IMHO it's just wasted runtime.

yarikoptic:
> Nobody was able/willing to figure it out while it was making CI fail

from my observation - busy system. And that was at large reason for majority of other fails in test_parallel -- consumer threads are not executed in the number/order expected by the test.

> So, whatever the outcome of the test - there will be no consequences. IMHO it's just wasted runtime.

In general (not this particular xfail assert for which takes virtually no time) I agree to disagree: the test has other conditions and possibly even contributes uniquely to code coverage. As with any test, "they are useless" after possibly initial TDD is done and until some changes in environment or code make them resurface problems again. But I do agree that XFAILs are largely forgotten about, as well as our `@known_failure`s.

Ok, overall, with this blessing let's proceed

yarikoptic:
note: travis - filed https://github.com/datalad/datalad/issues/7129 . appveyor -- seems like that new git/git-annex giving us issues we can no longer postpone looking at :-/  https://github.com/datalad/datalad/issues/7116

yarikoptic-gitmate:

yarikoptic-gitmate:

PR #7132, opened 2022-11-02
More informative AltConstraint error

### LINKED ISSUE #7092: `AltConstraint` error could be more informative
Here is the patch

```diff
diff --git a/datalad/support/constraints.py b/datalad/support/constraints.py
index 19ad29912..dd54132f5 100644
--- a/datalad/support/constraints.py
+++ b/datalad/support/constraints.py
@@ -434,8 +434,10 @@ class AltConstraints(_MultiConstraint):
                 return c(value)
             except Exception as e:
                 e_list.append(e)
-        raise ValueError("all alternative constraints (%s) violated while testing value %r"
-                         % (self.constraints, value))
+        raise ValueError(
+            f"{value} does not match any alternative: "
+            f"{self.constraints} {e_list}"
+        )
 
 
```

It turns:

```
ValueError: all alternative constraints ([constraint:<class 'list'>(constraint:path), constraint:None]) violated while testing value PosixPath('/home/mih/demo/myproject/inputs/penguins/Materials/HPP_scales_share_Final.xlsx')
```

into 

```
ValueError: /home/mih/demo/myproject/inputs/penguins/Materials/HPP_scales_share_Final.xlsx does not match any alternative: [constraint:<class 'list'>(constraint:path), constraint:None] [TypeError("'PosixPath' object is not iterable"), ValueError('value must be `None`')]
```

It is not an option to simply `raise ... from ....`, because there could be more than one exception to report.

bpoldrack:
Closes #7092

### DIFF
--- a/changelog.d/pr-7132.md
+++ b/changelog.d/pr-7132.md
@@ -1,0 +1,6 @@
+### Enhancements and New Features
+
+- Make validation failure of alternative constraints more informative.
+  Fixes [#7092](https://github.com/datalad/datalad/issues/7092) via
+  [PR #7132](https://github.com/datalad/datalad/pull/7132)
+  (by [@bpoldrack](https://github.com/bpoldrack))
--- a/datalad/support/constraints.py
+++ b/datalad/support/constraints.py
@@ -434,8 +434,8 @@ class AltConstraints(_MultiConstraint):
                 return c(value)
             except Exception as e:
                 e_list.append(e)
-        raise ValueError("all alternative constraints (%s) violated while testing value %r"
-                         % (self.constraints, value))
+        raise ValueError(f"{value} does not match any alternative: "
+                         f"{self.constraints} {e_list}")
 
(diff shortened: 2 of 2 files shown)


bpoldrack:
crippledFS failure: https://github.com/datalad/datalad/actions/runs/3376700932/jobs/5604739627

This one is strange. Apart from the fact that I don't see even a remote relationship to change here (and it didn't happen in https://github.com/datalad/datalad/pull/7115), I'm also wondering how the test could ever pass on crippled FS before. It relies on checking out `DEFAULT_BRANCH` which shouldn't work that way on adjusted branches.
I'm missing something. ATM neither the fact that it didn't fail before nor the fact that it now starts to fail makes any sense to me.

bpoldrack:
I think this is trivial. Since CI passes now, will go ahead and merge.

yarikoptic-gitmate:

PR #7133, opened 2022-11-02
Let EnsureDataset handle Path instances

### LINKED ISSUE #7069: `EnsureDataset` cannot handle `Path` values
`datalad-gooey` does full parameter validation not, and because of that fails on any parameter that has an `EnsureDataset` constraint, because it does not use string paths internally, but `Path` objects.

This is a bit ironic, because `EnsureDataset` does not actually do anything other than rejecting anything that is not a `str` or `Dataset` instance.

I will replace this constraint in `Gooey` with something more useful, because the rational in the code doesnot apply there:

```
            # we cannot convert to a Dataset class right here
            # - duplicates require_dataset() later on
            # - we need to be able to distinguish between a bound
            #   dataset method call and a standalone call for
            #   relative path argument disambiguation
            #return Dataset(path=value)
 ```

Ultimately, datalad-core will need to do something about it, when it would start validating Python command arguments.

https://github.com/datalad/datalad-gooey/issues/298

bpoldrack:
Previously this constraint allowed for strings only.

Closes #7069

### DIFF
--- a/changelog.d/pr-7133.md
+++ b/changelog.d/pr-7133.md
@@ -1,0 +1,6 @@
+### Internal
+
+- Allow EnsureDataset constraint to handle Path instances.
+  Fixes [#7069](https://github.com/datalad/datalad/issues/7069) via
+  [PR #7133](https://github.com/datalad/datalad/pull/7133)
+  (by [@bpoldrack](https://github.com/bpoldrack))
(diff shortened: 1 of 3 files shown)


yarikoptic (approved):
Looks good to me. appveyor tests should pass if rebased

yarikoptic-gitmate:

Issue #7100, opened 2022-10-19
Tests still require nose


TheChymera:
even though the test suite now works with pytest, nose is needed for testing, and not just for the docs (which is what `setup.py` indicates):

```console
chymera@decohost ~/src/datalad $ git rev-parse HEAD
b3ff763c8f1eff6ec6c21c837dd4c37136944fe0
chymera@decohost ~/src/datalad $ grep nose datalad/tests/utils.py
from nose import SkipTest
from nose.plugins.attrib import attr
from nose.tools import (
from nose.tools import assert_in
from nose.tools import assert_in as in_
from nose.tools import (
    from nose.ext import dtcompat
    from nose.plugins import (
    # Monkey patch nose so it does not ERROR out whenever code asks for fileno
    # of the output. See https://github.com/nose-devs/nose/issues/6
        from nose import SkipTest
    The only way would be to instruct testing framework (i.e. nose
def ignore_nose_capturing_stdout(func):
    Originally was intended as a decorator workaround for nose's behaviour
    with redirecting sys.stdout, but now we monkey patch nose now so no test
    https://code.google.com/p/python-nose/issues/detail?id=243&can=1&sort=-id&colspec=ID%20Type%20Status%20Priority%20Stars%20Milestone%20Owner%20Summary
        "@ignore_nose_capturing_stdout no longer does anything - nose should "
 ```

bpoldrack:
@TheChymera: Is it just that you saw `nose` being used there or that you could not run the tests without `nose` being installed?

The intention here was to keep a nose-dependent `tests/utils.py` for a while to provide it for extensions that may still rely on nose (including our own extensions that could not be switch all at once). The tests in `datalad` should actually not use anything in `tests/utils.py` anymore, but the equivalents from `tests/utils_pytest.py`, so that `tests/utils.py` can be deleted sometime in the not too distant future.

bpoldrack:
I'm closing this, since I think there's nothing to do ATM.

@TheChymera: Please feel free to reopen, if you think I misunderstood the point of it.

TheChymera:
The tests actually failed, because the code was loaded when running the test suite, but it's ok, I just listed nose as a dependency, will re-open if it's still needed by the next release.

bpoldrack:
Actually, that suggests there's something incomplete on our end.
This should be excluded from inspection by pytest. Plus: Since CI isn't failing here, we probably manage to still install `nose` in their environment somehow.

Thx, @TheChymera

[...]

PR #7136, opened 2022-11-02
Removing traces of nose

bpoldrack:
Addressing  #7100

(Need to figure how to deal with `py.path.local`s)

### DIFF
--- a/.github/workflows/test_extensions.yml
+++ b/.github/workflows/test_extensions.yml
@@ -60,12 +60,6 @@ jobs:
         case ${{ matrix.extension }} in
           datalad-metalad)
             DL_PIP_INSTALLS+=" nose";;
-          datalad-neuroimaging)
-            DL_TESTER=nose;;
-          datalad-hirni)
-            DL_REPO=psychoinformatics-de/${{ matrix.extension }}
-            DL_TESTER=nose
-            DL_NEED_SINGULARITY=1;;
           datalad-container)
             DL_NEED_SINGULARITY=1;;
         esac
--- a/changelog.d/pr-7136.md
+++ b/changelog.d/pr-7136.md
@@ -1,0 +1,6 @@
+### 🔩 Dependencies
+
+- Remove remaining references to `nose` and the implied requirement for building the documentation
+  Fixes [#7100](https://github.com/datalad/datalad/issues/7100) via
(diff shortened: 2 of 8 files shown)


yarikoptic:
FWIW -- I agree with the taken approach. Might have done a different check (that `(.parent / "tests" / "test_tests_utils.py").exists()`) but ok with suggsted. I would've only brought it into `maint` since @TheChymera would want it for a release, and it might come sooner from `maint`.

bpoldrack:
> I would've only brought it into maint since @TheChymera would want it for a release

Agree, will do.

.github/workflows/test_extensions.yml
```diff
@@ -60,12 +60,6 @@ jobs:
         case ${{ matrix.extension }} in
           datalad-metalad)
             DL_PIP_INSTALLS+=" nose";;
-          datalad-neuroimaging)
-            DL_TESTER=nose;;
-          datalad-hirni)
-            DL_REPO=psychoinformatics-de/${{ matrix.extension }}
-            DL_TESTER=nose
-            DL_NEED_SINGULARITY=1;;
```
yarikoptic, line 68: note: hirni is no longer tested and neuroimaging pytest-ified

yarikoptic:
looks good to me, and we still have http://github.com/datalad/datalad/blob/HEAD/.github/workflows/test-nose.yml which tests that nose code with nose.  I will mark this as a candidate to cut the release (we need it) if merged.

> (Need to figure how to deal with `py.path.local`s)

is that still a "TODO" item? (there is no checkbox available/marked)

bpoldrack:
> is that still a "TODO" item?

Nope. Ready from my point of view.

.github/workflows/test_extensions.yml
bpoldrack: Yes, that's why I wanted to remove it, since this file kinda suggests a different reality.

yarikoptic:
one (first) travis run got stuck... unclear why yet but unlikely, restarted for the sake of it. But given that the rest is all good -- let's proceed and cut release!

yarikoptic:
doh - I was too late with adding `release` label :-/

yarikoptic-gitmate:

yarikoptic-gitmate:

PR #7147, opened 2022-11-07
BF: read RIA config from stdin instead of temporary file

### LINKED ISSUE #6514: clone from RIA store (kinda) fails with invalid TMPDIR
A user reported an issue where she wouldn't be able to `get` a subdataset resulting in:

```
[INFO   ] Fetching 'http://store.datalad.org/d3f/a72e4-2c2b-11ea-948f-0025904abcb0/config' 
[ERROR  ] FileNotFoundError([Errno 2] No such file or directory: '/tmpstore/datalad_temp_9lpj5jqp') (FileNotFoundError) 
```

Turns out she had an invalid `TMPDIR=/tmpstore`. The call succeeded on second try, though and it looked like it would behave differently when called with DEBUG logging.

Culprit is: `clone` is putting the content of the store's `config` into a tempfile in order to let a call to `git config` parse it. This is happening post-clone, so the actual cloning has already succeeded, explaining the observed differences in behavior. However, this partial success is not recognizeable to the user.
Moreover, the need for the tempfile is questionable to begin with, as this comment in the code states:

```
# TODO: We might be able to spare the saving to a file.
#       "git config -f -" is not explicitly documented but happens
#       to work and would read from stdin. Make sure we know this
#       works for required git versions and on all platforms.
```

So, two things need to be done, from my POV:

1. Make sure we don't crash in a postclone routine, obfuscating the fact that we already succeeded with the cloning (or crash but then clean up the clone)

2. As the comment says: If we can confirm this to work reliably, we should switch and not bother with a tempfile at all.

adswa:
A RIA postclone config hook tried to read a special remote UUID from a stores config file by writing the config files contents to a temporary config file, and subsequently running ``git config -f <thatfile>``. As outlined in a comment by @bpoldrack in the code, ``git config -f -`` can read from stdin:

```
echo '[core]\n\trepositoryformatversion = 0\n\tfilemode = true\n\tbare = true\n' | git config -f - core.filemode
true
```

A quick chat in the Git IRC channel hinted that this is supported by git as its a unix convention. Trial and error confirms that this patch also runs successfully in a Windows CMD. I'll rely on the tests on MacOS to check if it runs there successfully, too. By switching to a git config call with stdin as the input file, we spare the need to write a temporary RIA config file. This fix could thus fix #6514, where a user error lead to a non-existent TMP path and a subsequent failure to create the temporary config file.

### DIFF
--- a/changelog.d/pr-7147.md
+++ b/changelog.d/pr-7147.md
@@ -1,0 +1,3 @@
+### 🐛 Bug Fixes
+
+- BF: read RIA config from stdin instead of temporary file.  Fixes [#6514](https://github.com/datalad/datalad/issues/6514) via [PR #7147](https://github.com/datalad/datalad/pull/7147) (by [@adswa](https://github.com/adswa))
--- a/datalad/core/distributed/clone.py
+++ b/datalad/core/distributed/clone.py
@@ -953,24 +953,20 @@ def postclonecfg_ria(ds, props, remote="origin"):
         # And read it
         uuid = None
         if config_content:
-            # TODO: We might be able to spare the saving to a file.
-            #       "git config -f -" is not explicitly documented but happens
-            #       to work and would read from stdin. Make sure we know this
-            #       works for required git versions and on all platforms.
-            with make_tempfile(content=config_content) as cfg_file:
-                runner = GitWitlessRunner()
-                try:
-                    result = runner.run(
-                        ['git', 'config', '-f', cfg_file,
-                         'datalad.ora-remote.uuid'],
-                        protocol=StdOutCapture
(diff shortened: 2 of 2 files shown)


bpoldrack (approved):
Great! Thanks double-checking with git, @adswa !

bpoldrack:
Note to myself: This will need some rectification in `master`, where the postclone routine moved into a dedicated RIA patch.

yarikoptic:
note: benchmarks error are likely unrelated --  filed https://github.com/datalad/datalad/issues/7149

yarikoptic-gitmate:

PR #7153, opened 2022-11-08
Fix for type checking

bpoldrack:
I don't think this needs a changelog.


bpoldrack:
Type-check passed, nothing else to wait for here, I think. Will merge right away.

PR #7155, opened 2022-11-08
Fix runner not dealing with str as stdin

### LINKED ISSUE #7145: Q: how to pass a string as stdin to Runners' run()?
The docstring of the Runner's [run](https://github.com/datalad/datalad/blob/66a8b77c09af3b0c8871ef541a061596e8257e44/datalad/runner/runner.py#L69) method state the following about the ``stdin=`` parameter:
https://github.com/datalad/datalad/blob/66a8b77c09af3b0c8871ef541a061596e8257e44/datalad/runner/runner.py#L91-L98

However, if I provide a string to it (I tried toying around in the context of https://github.com/datalad/datalad/issues/6514), I get
```
  File "/usr/lib/python3.10/threading.py", line 1016, in _bootstrap_inner
    self.run()
  File "/home/adina/repos/datalad/datalad/runner/runnerthreads.py", line 172, in run
    succeeded = self.write(data)
  File "/home/adina/repos/datalad/datalad/runner/runnerthreads.py", line 246, in write
    written += os.write(
TypeError: a bytes-like object is required, not 'str'
```

I feel like I might be missing something fairly obvious. How can I provide a string as stdin to a ``runner.run()`` call?

bpoldrack:
Fixes #7145

### DIFF
--- a/datalad/runner/nonasyncrunner.py
+++ b/datalad/runner/nonasyncrunner.py
@@ -173,11 +173,11 @@ class ThreadedRunner:
             `datalad.runner.protocol.GeneratorMixIn`, the function will return
             the result created by the protocol method `_generate_result`.
 
-        stdin : file-like, string, bytes, Queue, or None
+        stdin : file-like, bytes, Queue, or None
             If stdin is a file-like, it will be directly used as stdin for the
             subprocess. The caller is responsible for writing to it and closing
-            it. If stdin is a string or bytes, those will be fed to stdin of the
-            subprocess. If all data is written, stdin will be closed.
+            it. If stdin is a bytes, it will be fed to stdin of the subprocess.
+            If all data is written, stdin will be closed.
             If stdin is a Queue, all elements (bytes) put into the Queue will
             be passed to stdin until None is read from the queue. If None is
             read, stdin of the subprocess is closed.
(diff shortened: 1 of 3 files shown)


bpoldrack:
Does this look aright to you, @christian-monch ?

datalad/runner/nonasyncrunner.py
```diff
@@ -333,7 +333,8 @@ def _locked_run(self) -> dict | _ResultGenerator:
             # enqueue the input that is already provided.
             self.write_stdin = True
             self.stdin_queue = Queue()
-            self.stdin_queue.put(self.stdin)
+            self.stdin_queue.put(self.stdin if isinstance(self.stdin, bytes)
+                                 else self.stdin.encode(encoding='utf-8'))
```
christian-monch, line 337: Basically correct, but I would use the protocol_kwargs encoding value, if it is set. Besides that, the indentation in this file has a different style and strings here are always delimited with double-quotes.

```suggestion
            self.stdin_queue.put(
                self.stdin
                if isinstance(self.stdin, bytes)
                else self.stdin.encode(
                    encoding=self.protocol_kwargs.get("encoding", "utf-8")))
```

christian-monch (approved):
Thx @bpoldrack 
LGTM with the suggested changes

bpoldrack:
@christian-monch: Agree with your take in #7145. Force-pushed accordingly.

yarikoptic-gitmate:

Issue #7157, opened 2022-11-09
2 tests fail in cron travis run  


yarikoptic:
e.g. from https://app.travis-ci.com/github/datalad/datalad/jobs/587920121

```
=========================== short test summary info ============================
FAILED ../datalad/customremotes/tests/test_datalad.py::test_basic_scenario_local_url - AssertionError: assert '_' in '3versions-allversioned.txt'
FAILED ../datalad/distributed/tests/test_ria_basics.py::test_version_check - datalad.support.exceptions.IncompleteResultsError: Command did not complete successfully. 1 failed:
[{'action': 'copy',
  'annexkey': 'MD5E-s18--6beb3be625001976a17c6eb78f88c592',
  'error_message': 'Remote is treated as read-only. Set '
                   "'ora-remote.<name>.force-write=true' to overrule this.\n"
                   'This could have failed because --fast is enabled.',
  'path': '/tmp/datalad_temp__test_version_check8aj6f1ae/new_file',
  'refds': '/tmp/datalad_temp__test_version_check8aj6f1ae',
  'status': 'error',
  'target': 'store',
  'type': 'file'}]
= 2 failed, 1160 passed, 45 skipped, 4 xfailed, 17 warnings in 869.72s (0:14:29) =
The command "http_proxy= PATH=$PWD/../tools/coverage-bin:$PATH $PYTEST_WRAPPER python -m pytest "${PYTEST_OPTS[@]}" -c ../tox.ini -n 2 -m "${PYTEST_SELECTION:+$PYTEST_SELECTION_OP($PYTEST_SELECTION) and }not(turtle)" --doctest-modules --cov=datalad --cov-report= --pyargs $TESTS_TO_PERFORM" exited with 1.
```

<details>
<summary>had been happening for awhile, e.g. all previous month</summary> 

```shell
(git)smaug:/mnt/datasets/datalad/ci/logs/2022/10[master]git
$> git grep 'FAILED.*test_basic_scenario_local_url'
07/cron/20221007T213523/22c97cd/travis-14676-failed/12.txt:[gw1] [ 17%] FAILED ../datalad/customremotes/tests/test_datalad.py::test_basic_scenario_local_url 
07/cron/20221007T213523/22c97cd/travis-14676-failed/12.txt:FAILED ../datalad/customremotes/tests/test_datalad.py::test_basic_scenario_local_url
[...]

PR #7158, opened 2022-11-09
BF(TST): for older versions allow either replaced or not -

yarikoptic:
Our cron tests against git annex 8.20200309 were failing due to wrong assumption that some prior old versions MUST had it wrong, which apparently was not the case.

Addressed 1/2 of #7157


bpoldrack (approved):
Change makes sense, thx @yarikoptic

PR #7161, opened 2022-11-09
RF: Specify parameter choices uniformly

### LINKED ISSUE #7026: clean's parameter specification sticks out
The `--what` parameter to the `clean` command is an outlier in how the choices for it are specified. Most commands use `EnsureChoice` while this one specifies `choices` for argparse instead.

Not sure yet, whether we should aim to unify by using either one of them everywhere or both or rework the interface specification even more and come up with a new way. However, this makes parameter discovery harder for datalad-gooey.

bpoldrack:
`clean` was an outlier in that regard.

Closes #7026

### DIFF
--- a/changelog.d/pr-7161.md
+++ b/changelog.d/pr-7161.md
@@ -1,0 +1,6 @@
+### 🏠 Internal
+
+- Unify definition of parameter choices with `datalad clean`.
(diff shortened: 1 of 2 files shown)


adswa (approved):

adswa:
I guess nobody complained in a week, merging.

yarikoptic-gitmate:

PR #7166, opened 2022-11-10
BF: Prevent doomed annex calls on files we already know are untracked

### LINKED ISSUE #7032: `status --annex basic <untrackedfile>` fails with unnecessary annex call
I have an untracked file in a dataset:

```json
$ datalad -f json status -d .  this
{"action": "status", "parentds": "/tmp/top1", "path": "/tmp/top1/this",
"refds": "/tmp/top1", "state": "untracked", "status": "ok", "type": "file"}
```

When I ask for `--annex basic` it attempts to run annex for this file,
although it could not possibly be in annex.

```json
$ datalad -f json status -d . --annex basic this
{"action": "status", "command": "find",
"error_message": "File unknown to git", "note": "not found",
"parentds": "/tmp/top1", "path": "/tmp/top1/this", "refds": "/tmp/top1",
"state": "untracked", "status": "ok", "success": false, "type": "file"}
```

adswa:
Fixes #7032 

This change adds a short check if any file in a list of paths given to a ``datalad status --annex `` call is untracked. If it is, it is removed from the list of paths given to the annex query, as we already know that this annex call would fail. 
It does not prevent the status report of being untracked:

```
❱ datalad status --annex basic my some moreuntracked 
untracked: moreuntracked (file)
untracked: my (file)
1 annex'd file (7.0 B recorded total size)
```

If untracked paths are excluded from this query, a trace of it is left in the debug log. 

This PR also adds a small test.

### DIFF
--- a/changelog.d/pr-7166.md
+++ b/changelog.d/pr-7166.md
@@ -1,0 +1,3 @@
+### 🐛 Bug Fixes
+
+- BF: Prevent doomed annex calls on files we already know are untracked.  Fixes [#7032](https://github.com/datalad/datalad/issues/7032) via [PR #7166](https://github.com/datalad/datalad/pull/7166) (by [@adswa](https://github.com/adswa))
--- a/changelog.d/pr-7166.md
+++ b/changelog.d/pr-7166.md
@@ -1,3 +1,3 @@
 ### 🐛 Bug Fixes
 
-- BF: Prevent doomed annex calls on files we already know are untracked.  Fixes [#7032](https://github.com/datalad/datalad/issues/7032) via [PR #7166](https://github.com/datalad/datalad/pull/7166) (by [@adswa](https://github.com/adswa))
+- Prevent doomed annex calls on files we already know are untracked.  Fixes [#7032](https://github.com/datalad/datalad/issues/7032) via [PR #7166](https://github.com/datalad/datalad/pull/7166) (by [@adswa](https://github.com/adswa))
--- a/datalad/core/local/status.py
+++ b/datalad/core/local/status.py
@@ -168,6 +168,15 @@ def yield_dataset_status(ds, paths, annexinfo, untracked, recursion_limit,
         eval_submodule_state=eval_submodule_state,
         _cache=cache)
     if annexinfo and hasattr(repo, 'get_content_annexinfo'):
+        if paths:
+            # when an annex query has been requested for specific paths,
+            # exclude untracked files from the annex query (else gh-7032)
+            untracked = [k for k, v in status.items() if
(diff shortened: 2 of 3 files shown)


datalad/core/local/status.py
```diff
@@ -168,6 +168,15 @@ def yield_dataset_status(ds, paths, annexinfo, untracked, recursion_limit,
...
         _cache=cache)
     if annexinfo and hasattr(repo, 'get_content_annexinfo'):
+        if paths:
+            # when an annex query has been requested for specific paths,
+            # exclude untracked files from the annex query (else gh-7032)
+            untracked = [k for k, v in status.items() if
+                         v['state'] == 'untracked']
+            lgr.debug(
+                'Skipping %s.get_content_annexinfo() for untracked paths: %s',
+                repo, paths)
```
yarikoptic, line 178: even though only for `DEBUG` I wonder if it is wise to include all `paths` -- what is there is 1000s of them?  do we ever log all `paths` . I would just log number of them I guess, or may be we have some helper to show like first N and then `...` ?

adswa: Oh, I agree that this would be sensible. I guess it never occurred to me because the very next line in that piece of code is
```
 lgr.debug('Querying %s.get_content_annexinfo() for paths: %s', repo, paths)
```
which would probably also benefit from such a helper. But I'm not aware of one...

bpoldrack (approved):
Looks alright to me!

Are you planning to address the logging, @adswa? Otherwise I'm fine with merge - logging can be addressed independently.

adswa:
I don't have a good idea how to do the logging. As it would be necessary also in other parts of that code, I'd be in favour of addressing this outside of this PR.

changelog.d/pr-7166.md
```diff
@@ -0,0 +1,3 @@
+### 🐛 Bug Fixes
+
+- BF: Prevent doomed annex calls on files we already know are untracked.  Fixes [#7032](https://github.com/datalad/datalad/issues/7032) via [PR #7166](https://github.com/datalad/datalad/pull/7166) (by [@adswa](https://github.com/adswa))
```
bpoldrack, line 3:
```suggestion
- Prevent doomed annex calls on files we already know are untracked.  Fixes [#7032](https://github.com/datalad/datalad/issues/7032) via [PR #7166](https://github.com/datalad/datalad/pull/7166) (by [@adswa](https://github.com/adswa))
```

bpoldrack:
Minor change to changelog - will merge w/o rerunning tests.

yarikoptic-gitmate:

PR #7170, opened 2022-11-11
MNT: Move eval_results() into interface.base

### LINKED ISSUE #6694: Move `interface.utils.eval_results` to `interface(.base)`
There is pretty much no `Interface` implementation without it, just like with `build_doc`. IMHO both pieces should live in the same place.

This would make the common

```py
from datalad.interface.base import (
    Interface,
    build_doc,
)
from datalad.interface.utils import (
    eval_results,
)
```

just be 

```py
from datalad.interface.base import (
    Interface,
    build_doc,
    eval_results,
)
```

an argument could be made to consolidate all (more or less strictly) required pieces in `datalad.interface`. Looking like this:

```py
from datalad.interface.base import (
    Interface,
    Parameter,
    build_doc,
    datasetmethod,
    eval_results,
)
```

adswa:
If done right, this PR fixes #6694. It moves ``eval_results()`` from ``interface.utils`` into ``interface.base``.  It replaces it with a shim ``eval_results()`` in ``interface.utils`` which issues a deprecation warning and imports the moved function back from ``interface.base``, such that this hopefully doesn't break things left and right.

I moved two helper functions, ``_execute_command_`` and ``_validate_cmd_call`` together with ``eval_results``. I found it cleaner than importing underscore functions from ``utils`` into ``base``, and got entangled in a few circular imports when I left it in `utils`. 

Tests were ok locally, I hope the extensions manage, too :crossed_fingers:

### DIFF
--- a/changelog.d/pr-7170.md
+++ b/changelog.d/pr-7170.md
@@ -1,3 +1,3 @@
 ### 🏠 Internal
 
-- MNT: Move eval_results() into interface.base.  Fixes [#6694](https://github.com/datalad/datalad/issues/6694) via [PR #7170](https://github.com/datalad/datalad/pull/7170) (by [@adswa](https://github.com/adswa))
+- Move eval_results() into interface.base to simplify imports for command implementations. Deprecate use from interface.utils accordingly. Fixes [#6694](https://github.com/datalad/datalad/issues/6694) via [PR #7170](https://github.com/datalad/datalad/pull/7170) (by [@adswa](https://github.com/adswa))
--- a/changelog.d/pr-7170.md
+++ b/changelog.d/pr-7170.md
@@ -1,0 +1,3 @@
+### 🏠 Internal
+
+- MNT: Move eval_results() into interface.base.  Fixes [#6694](https://github.com/datalad/datalad/issues/6694) via [PR #7170](https://github.com/datalad/datalad/pull/7170) (by [@adswa](https://github.com/adswa))
--- a/datalad/interface/utils.py
+++ b/datalad/interface/utils.py
@@ -12,18 +12,9 @@
 
 __docformat__ = 'restructuredtext'
 
-import inspect
 import logging
 import sys
-from functools import wraps
 from time import time
-from typing import (
-    Callable,
-    Dict,
-    Generator,
-    TypeVar,
-    TYPE_CHECKING,
-)
 from os import listdir
 from os.path import join as opj
 from os.path import isdir
@@ -32,6 +23,9 @@ from os.path import sep
 
 import json
 
+from typing import (
+    TypeVar
+)
 # avoid import from API to not get into circular imports
 from datalad.utils import with_pathsep as _with_sep  # TODO: RF whenever merge conflict is not upon us
 from datalad.utils import (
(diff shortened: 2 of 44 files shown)


adswa:
Sadly, it broke metalad:
```
Error: [ERROR] Internal error, cannot import interface 'datalad_metalad.dump': DeprecationWarning(datalad.interface.utils.eval_results is obsolete. Use datalad.interface.base.eval_results instead) 
[...]
```
The other extensions are fine

bpoldrack:
Great! Thx, @adswa .

I am adding a changelog

adswa:
thx

changelog.d/pr-7170.md
```diff
@@ -0,0 +1,3 @@
+### 🏠 Internal
+
+- MNT: Move eval_results() into interface.base.  Fixes [#6694](https://github.com/datalad/datalad/issues/6694) via [PR #7170](https://github.com/datalad/datalad/pull/7170) (by [@adswa](https://github.com/adswa))
```
bpoldrack, line 3:
```suggestion
- Move eval_results() into interface.base to simply imports for command implementations. Deprecate use from interface.utils accordingly. Fixes [#6694](https://github.com/datalad/datalad/issues/6694) via [PR #7170](https://github.com/datalad/datalad/pull/7170) (by [@adswa](https://github.com/adswa))
```

datalad/interface/utils.py
```diff
@@ -241,289 +218,12 @@ def _fx(res, **kwargs):
...
-
-    This default implementation performs no validation.
-    """
-    pass
+    import warnings
+    from datalad.interface.base import eval_results as f
+    warnings.warn("datalad.interface.utils.eval_results is obsolete. "
+                  "Use datalad.interface.base.eval_results instead",
+                  DeprecationWarning)
+    return f(wrapped)
```
yarikoptic, line 226: minor:

```suggestion
    from datalad.interface.base import eval_results as eval_results_moved
    warnings.warn("datalad.interface.utils.eval_results is obsolete. "
                  "Use datalad.interface.base.eval_results instead",
                  DeprecationWarning)
    return eval_results_moved(wrapped)
```

to avoid confusing `f` in a possible traceback.

datalad/interface/utils.py
```diff
@@ -570,6 +270,7 @@ def render_action_summary(action_summary):
                     for act in sorted(action_summary))))
 
 
+
```
yarikoptic, line 273:
```suggestion
```
save the byte, plant a tree!

yarikoptic (approved):
generally approved but provided few minor suggestions.

datalad/interface/utils.py
adswa: whhoops

adswa:
thx for the review - merging

yarikoptic-gitmate:

yarikoptic:
Note to myself and others: we should keep in mind that such "core" moves cause extensions to break testing (warnings treated as errors) so if left not attended to right away - cause some confusion/rush to fix (e.g. for debian) etc. So best would be if such PRs are accompanied with PRs against extensions to bring them up to date (and also showing real DX cost for even a small API change).

adswa:
>  edit: probably worth patching distribution packages to not treat warnings as errors during testing!

Do you mean like the patch e.g., metalad got in ``tox.ini`` when testing of this PR showed that warnings are treated as errors (https://github.com/datalad/datalad-metalad/pull/295)?

yarikoptic:
yes. Although in datalad-container we do not even have tox.ini , so not sure why it was errorring out during build of debian pkg -- might be some specific to debian env setting .

PR #7174, opened 2022-11-14
Use regular dicts instead of OrderedDicts for speedier operations

### LINKED ISSUE #6566: Use dict instead of OrderedDict
`dict` is ordered now in versions of python we support. And according to https://realpython.com/python-ordereddict/ - `dict` seems to generally outperform `OrderedDict` in CPU and memory, and I don't think we use any of specific to OrderedDict functionality.

```
(git)lena:~datalad/datalad-maint[bf-sve]git
$> git grep OrderedDict | wc -l 
64
``` 
I think it would not hurt to just switch to just use `dict`.

adswa:
This is an attempt to fix #6566, which calls for a replacement of OrderedDicts with regular dicts to increase performance/speed. There was only one instance in ``status()`` where OrderedDict-specific functionality was used; I used a patch cooked up by @yarikoptic in #6566 as an alternative. 
Let's check if the CI is happy with this change or if it breaks something.

### DIFF
--- a/changelog.d/pr-7174.md
+++ b/changelog.d/pr-7174.md
@@ -1,0 +1,3 @@
+### 🏎 Performance
+
+- Use regular dicts instead of OrderedDicts for speedier operations.  Fixes [#6566](https://github.com/datalad/datalad/issues/6566) via [PR #7174](https://github.com/datalad/datalad/pull/7174) (by [@adswa](https://github.com/adswa))
--- a/datalad/core/distributed/push.py
+++ b/datalad/core/distributed/push.py
@@ -12,7 +12,6 @@
 
 __docformat__ = 'restructuredtext'
 
-from collections import OrderedDict
 from itertools import chain
 import logging
 import re
@@ -868,13 +867,13 @@ def _push_data(ds, target, content, data, force, jobs, res_kwargs,
     # input has type=dataset, but now it is about files
     res_kwargs.pop('type', None)
 
-    # A set and an OrderedDict is used to track files pointing to the
+    # A set and a dict is used to track files pointing to the
     # same key.  The set could be dropped, using a single dictionary
     # that has an entry for every seen key and a (likely empty) list
     # of redundant files, but that would mean looping over potentially
     # many keys to yield likely few if any notneeded results.
     seen_keys = set()
-    repkey_paths = OrderedDict()
+    repkey_paths = dict()
 
     # produce final path list. use knowledge that annex command will
     # run in the root of the dataset and compact paths to be relative
--- a/datalad/core/local/diff.py
+++ b/datalad/core/local/diff.py
@@ -13,7 +13,6 @@ __docformat__ = 'restructuredtext'
 
 import logging
 import os.path as op
-from collections import OrderedDict
 from datalad.utils import (
     ensure_list,
     ensure_unicode,
@@ -283,7 +282,7 @@ def diff_dataset(
             if recursion_limit is not None and recursive
             else -1 if recursive else 0,
             # TODO recode paths to repo path reference
-            origpaths=None if not path else OrderedDict(path),
+            origpaths=None if not path else dict(path),
             untracked=untracked,
             annexinfo=annex,
             cache=content_info_cache,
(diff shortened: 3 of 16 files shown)


adswa:
There is an Appveyor failure that I feel I had seen fail before already, but I can't find it in recently opened PRs:

```
____________________________ test_gracefull_death _____________________________
[gw0] win32 -- Python 3.9.13 C:\Python39-x64\python.exe
    def test_gracefull_death():
    
        def assert_provides_and_raises(pc, exception, target=None):
            """Helper to get all results before exception is raised"""
            results = []
            with assert_raises(exception):
                for r in pc:
                    results.append(r)
            # results should be sorted since we do not guarantee order
            results = sorted(results)
            if target is not None:
                assert_equal(results, target)
            return results
    
        def interrupted_producer():
            yield 1
            raise ValueError()
    
        def consumer(i):
            sleep(0.001)
            yield i
    
        assert_provides_and_raises(
            ProducerConsumer(interrupted_producer(), consumer, jobs=3), ValueError, [1])
    
        def faulty_consumer(i):
            sleep(0.001)
            if i == 1:
                raise ValueError()
            return i
    
        # so we do not get failed, but other parallel ones finish their job
        results = assert_provides_and_raises(
            ProducerConsumer(range(1000), faulty_consumer, jobs=5), ValueError)
        # and analysis of futures to raise an exception can take some time etc, so
        # we could get more, but for sure we should not get all 999 and not even a 100
        if info_log_level:
            assert_greater(100, len(results))
        assert_equal(results[:4], [0, 2, 3, 4])
    
        def producer():
            for i in range(10):
                sleep(0.0003)
                yield i
            raise ValueError()
        # by default we do not stop upon producer failing
        assert_provides_and_raises(
            ProducerConsumer(producer(), consumer, jobs=2), ValueError, list(range(10)))
        # if producer produces more than we can as quickly consume but then fails
        # ATM we do not proceed to consume other items, but fail when we finish
        # consuming until the time point when producer has failed
        # by default we do not stop upon producer failing
        results = assert_provides_and_raises(
            ProducerConsumer(producer(), consumer, reraise_immediately=True, jobs=2),
            ValueError)
        # we will get some results, seems around 4 and they should be "sequential"
        assert_equal(results, list(range(len(results))))
        try:
        except AssertionError:
            # Possible TODO: if tests below would start failing too, move xfail to the level
            # of the entire test
            pytest.xfail(f"Rarely but happens. Got only {len(results)} instead of at least 2")
    
        # This test relies too much on threads scheduling to not hog up on handling
        # consumers, but if it happens so - they might actually consume all results
        # before producer decides to finally raise an exception.  As such it remains
        # flaky and thus not ran, but could be useful to test locally while
        # changing that logic.
        #
        # if info_log_level and not (on_windows or on_osx):
        #     # consumers should not be able to consume all produced items.
        #     # production of 10 should take 3 unites, while consumers 10/2 (jobs)
        #     # 5 units, so some should not have a chance.
        #     assert_greater_equal(8, len(results))
    
        # Simulate situation close to what we have when outside code consumes
        # some yielded results and then "looses interest" (on_failure="error").
        # In this case we should still exit gracefully (no GeneratorExit warnings),
        # not over-produce, and also do not kill already running consumers
        consumed = []
        def inner():
            def consumer(i):
                sleep(0.01)
                consumed.append(i)
                return i
            pc = iter(ProducerConsumer(range(1000), consumer, jobs=2))
            yield next(pc)
            yield next(pc)
        # typically it should be [0, 1] but it does happen some times that
        # one other worker gets ahead and we get [0, 2]. As it is not per se the
        # purpose of this test to ensure absence of such race, we just allow for any
        # two from first 3 possible.
        assert len(set(inner()).intersection({0, 1, 2})) == 2
        consumed = sorted(consumed)
        assert_equal(consumed, list(range(len(consumed))))
>       assert_greater_equal(len(consumed), 4)  # we should wait for that 2nd batch to finish
..\datalad\support\tests\test_parallel.py:236: 
first = 3, second = 4, msg = None
    def assert_greater_equal(first, second, msg=None):
        if msg is None:
>           assert first >= second
E           assert 3 >= 4
..\datalad\tests\utils_pytest.py:125: AssertionError
```

I've restarted the failed run to see if it persists.

yarikoptic (approved):
looks good, thank you!

yarikoptic-gitmate:

PR #7175, opened 2022-11-14
Add \n possibility into most obscure file name

### LINKED ISSUE #7173: Newline symbol into oscure filename?
I was told it is supported by posix, but I don't know if supported by git and us... Just making note to check

yarikoptic:
Closes #7173 

TODOs
- [ ] make CI happy
- [ ] send out a question to `git` ppl to ensure that they are aware of problems with submodules having a newline in the name

### DIFF
--- a/datalad/tests/utils_pytest.py
+++ b/datalad/tests/utils_pytest.py
@@ -1452,7 +1452,7 @@ with_parametric_batch = pytest.mark.parametrize("batch", [False, True])
 # filesystems across different OSs.  Start with the most obscure
 OBSCURE_PREFIX = os.getenv('DATALAD_TESTS_OBSCURE_PREFIX', '')
 # Those will be tried to be added to the base name if filesystem allows
-OBSCURE_FILENAME_PARTS = [' ', r'\n', '/', '|', ';', '&', '%b5', '{}', "'", '"', '<', '>']
+OBSCURE_FILENAME_PARTS = [' ', '\n', '/', '|', ';', '&', '%b5', '{}', "'", '"', '<', '>']
 UNICODE_FILENAME = u"ΔЙקم๗あ"
 
 # OSX is exciting -- some I guess FS might be encoding differently from decoding
--- a/datalad/tests/utils_pytest.py
+++ b/datalad/tests/utils_pytest.py
@@ -1517,7 +1517,7 @@ def get_most_obscure_supported_name(tdir, return_candidates=False, directory_nam
 
     # incrementally build up the most obscure filename from parts
     for part in OBSCURE_FILENAME_PARTS:
-        if directory_name is True and part in DIRECTORY_EXCLUDED_PARTS:
+        if directory_name and part in DIRECTORY_EXCLUDED_PARTS:
             continue
         candidate = good + part
         if good_filename(candidate):
--- a/datalad/tests/utils_pytest.py
+++ b/datalad/tests/utils_pytest.py
@@ -1474,7 +1474,7 @@ DIRECTORY_EXCLUDED_PARTS = ['\n']
 
 
 @with_tempfile(mkdir=True)
-def get_most_obscure_supported_name(tdir, return_candidates=False, directory_name=False):
+def get_most_obscure_supported_name(tdir, return_candidates=False):
     """Return the most obscure filename that the filesystem would support under TEMPDIR
 
     Parameters
@@ -1482,13 +1482,6 @@ def get_most_obscure_supported_name(tdir, return_candidates=False, directory_nam
     return_candidates: bool, optional
       if True, return a tuple of (good, candidates) where candidates are "partially"
       sorted from trickiest considered
-
-    directory_name: bool, optional
-      if True, return an obscure directory name. Directories might have
-      different obscurity profiles. Currently, directory names will not
-      contain newline characters. The reason for that is that git-annex
-      will not work properly in an annexed repository, if the root-directory
-      of the repository has a newline in its name.
 
     TODO: we might want to use it as a function where we would provide tdir
     """
@@ -1504,23 +1497,17 @@ def get_most_obscure_supported_name(tdir, return_candidates=False, directory_nam
             # ATM we do not distinguish obscure filename and dirname.
             # So here we will test for both - being able to create dir
             # with obscure name and obscure filename under
-            if directory_name is True:
-                os.mkdir(opj(tdir, filename))
-            else:
-                with open(opj(tdir, filename), 'w') as f:
-                    f.write("TEST LOAD")
+            with open(opj(tdir, filename), 'w') as f:
+                f.write("TEST LOAD")
             return True
         except:
             lgr.debug(
-                f"{'Dir' if directory_name else 'File'}name %r is "
-                "not supported on %s under %s",
+                f"File name %r is not supported on %s under %s",
                 filename, system, tdir)
             return False
 
     # incrementally build up the most obscure filename from parts
     for part in OBSCURE_FILENAME_PARTS:
-        if directory_name and part in DIRECTORY_EXCLUDED_PARTS:
-            continue
         candidate = good + part
         if good_filename(candidate):
             good = candidate
@@ -1535,8 +1522,8 @@ def get_most_obscure_supported_name(tdir, return_candidates=False, directory_nam
         return good
 
 
-OBSCURE_FILENAME, OBSCURE_FILENAMES = get_most_obscure_supported_name(return_candidates=True)
-OBSCURE_DIRNAME, OBSCURE_DIRNAMES = get_most_obscure_supported_name(return_candidates=True, directory_name=True)
+OBSCURE_FILENAME, OBSCURE_FILENAMES = get_most_obscure_supported_name(
+    return_candidates=True)
 
 
 @optional_args
@@ -1849,7 +1836,7 @@ def get_deeply_nested_structure(path):
     |      ├── subdir
     |      │   └── annexed_file.txt -> ../.git/annex/objects/...
     |      └── subds_lvl1_modified
-    |          └── OBSCURE_DIRNAME_directory_untracked
+    |          └── OBSCURE_FILENAME_directory_untracked
     |              └── untracked_file
 
     When a system has no symlink support, the link2... components are not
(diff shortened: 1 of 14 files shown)


datalad/tests/utils_pytest.py
```diff
@@ -1632,57 +1632,72 @@ def ignore_nose_capturing_stdout(func):
...
     Parameters
     ----------
     return_candidates: bool, optional
       if True, return a tuple of (good, candidates) where candidates are "partially"
       sorted from trickiest considered
+
+    directory_name: bool, optional
+      if True, return an obscure directory name. Directories might have
+      different obscurity profiles. Currently, directory names will not
+      contain newline characters.
```
yarikoptic, line 1652: Why so definitive? It is that we don't support new lines in directory - if so, please add a note why

datalad/tests/utils_pytest.py
```diff
@@ -1632,57 +1632,72 @@ def ignore_nose_capturing_stdout(func):
...
+                os.mkdir(opj(tdir, filename))
+            else:
+                with open(opj(tdir, filename), 'w') as f:
+                    f.write("TEST LOAD")
             return True
         except:
-            lgr.debug("Filename %r is not supported on %s under %s",
-                      filename, system, tdir)
+            lgr.debug(
+                f"{'Filename' if directory_name is False else 'Dirname'} %r is "
```
yarikoptic, line 1676: May be
```suggestion
                f"{'Dir' if directory_name else 'File'}name %r is "
```
would be easier since shorter etc?

datalad/tests/utils_pytest.py
```diff
@@ -1632,57 +1632,72 @@ def ignore_nose_capturing_stdout(func):
...
+                "not supported on %s under %s",
+                filename, system, tdir)
             return False
 
     # incrementally build up the most obscure filename from parts
     for part in OBSCURE_FILENAME_PARTS:
+        if directory_name is True and part in DIRECTORY_EXCLUDED_PARTS:
```
yarikoptic, line 1683:
```suggestion
        if directory_name and part in DIRECTORY_EXCLUDED_PARTS:
```

christian-monch:
We are going to have issues with commands that use `--batch`. The reason is, the BatchedCommand-interface implicitly assumes the newline terminates an input record for a batched command. Therefore a file name with newline would prematurely terminate the input record. Although we could update BatchedCommand to use different record terminators, e.g. `\0`, not all commands that support batched mode might support switching to a different input record terminator.

Such a switch would be a rather large effort and should only be attempted, if we are certain that all commands that we execute in batch-mode would support an alternative record terminator, i.e. `\0`. That would still leave the problem that we would not support file names that contain both, `\0` and `\n`.

(BTW: should we add `\0` to the OBSCURE_FILENAME parts?)

datalad/tests/utils_pytest.py
christian-monch: Yes, it is that we do not support new lines in directories. I commented in detail in issue #7173 in https://github.com/datalad/datalad/issues/7173#issuecomment-1314968568. Basically `git-annex` does not work in annex repositories that have a newline in the name of the root directory. Will add a comment.

christian-monch: Comment added

yarikoptic: Thanks! Might be worth alerting @joeyh to that fact as well

yarikoptic:
> (BTW: should we add \0 to the OBSCURE_FILENAME parts?)

Nope ;-)

yarikoptic:
> Although we could update BatchedCommand to use different record terminators, e.g. `\0`, not all commands that support batched mode might support switching to a different input record terminator.

apparently git-annex claims to support `\n` in filenames since 2018: https://git-annex.branchable.com/devblog/day_521__newlines_in_filenames/ . For `--batch` there should be `-z` to instruct to use `\0` instead of `\n` for separator.  Which commands @christian-monch you had in mind that they do not support `-z`?

bpoldrack:
@christian-monch 

> Then I realized that git-annex has its own problems with its --batch option and files with newlines. So I added support for -z to every batchable command in git-annex, including a couple of commands that did batch input without a --batch option.

The annex post linked by @yarikoptic says "Then I realized that git-annex has its own problems with its --batch option and files with newlines. So I added support for -z to every batchable command in git-annex, including a couple of commands that did batch input without a --batch option."

So, I think it's fine to assume every command does until proven otherwise. ;)


> The only problem that I saw with git annex and "\n" was that git annex seems to not work properly if the root directory of the annex repo has a name that contains "\n"

Just out of curiosity, can you elaborate on this? As in: What wasn't working?

christian-monch:
@bpoldrack 

> Just out of curiosity, can you elaborate on this? As in: What wasn't working?

I mentioned it above. It is detailed in this comment: https://github.com/datalad/datalad/issues/7173#issuecomment-1314968568

adswa:
What is the fate of this PR? Is someone feeling strong enough about #7173 to take it over or continue it?

datalad/tests/utils_pytest.py
yarikoptic: filed https://git-annex.branchable.com/bugs/init_fails_in_a_folder_with_newline_in_its_name/?updated .
Overall I think that
- git-annex issue should be fixed
- we should not bother distinguishing obscure path from folder -- our detection here doesn't help for git-annex case, and otherwise we seems to have no other use case where obscure filename cannot be obscure directory name *given filesystem limitations* and not bugs in the software, right?

yarikoptic:
indeed, we forgot about this one .  For specific subitems, filed [issue against annex](https://git-annex.branchable.com/bugs/init_fails_in_a_folder_with_newline_in_its_name/?updated) and one for us https://github.com/datalad/datalad/issues/7315 which I believe would be required to be done to progress on this one. Will mark it blocked for now.

datalad/tests/utils_pytest.py
christian-monch: > 

@yarikoptic : using newline in directories is a snake-pit. Try, for example, the following in a datalad dataset:
```bash
> datalad create -d . "abc
def"
```

That leads to an unusable dataset because ".git/config" and ".gitmodules" become invalid (even adding a submodule with newline in git, does leave the repo in an inconsistent state, because a .gitmodule entry for a commit-object is missing). 


Using OBSCURE_FILENAME for directories introduces a problem with submodule configuration entries in
`.git/config` and `.gitmodules`.  For example, a submodule called "abc\ndef", i.e. "abc" plus a newlines plus "def", leads to the following invalid entry in `.git/config`:

```
[submodule "abc
def"]
        active = true
        url = ./abc
def
```

and the following incorrect entry in `.gitmodules`.

```
[submodule "abc
def"]
        path = abc
def
        url = ./abc
def
        datalad-id = d9f3edc7-a39d-44db-9e74-4bb37741e08d
```

To allow newlines in subdataset names, we would have to quote the submodule name requires quoting of submodule names, paths and URLs. The folliowing `.gitmodules` and `.git/config` entries would work

.gitmodules:
``` 
[submodule "abc\ndef"]
        path = "abc\ndef"
        url = "./abc\ndef"
        datalad-id = d9f3edc7-a39d-44db-9e74-4bb37741e08d
``` 

.git/config
``` 
[submodule "abc\ndef"]
        active = true
        url = "./abc\ndef"
``` 

But not even git creates the proper entries, when adding a submodule with a newline in the directory name.

So in conclusion, we should not try to use newlines in directories

christian-monch:
ping @yarikoptic posted a new reply to an old comment https://github.com/datalad/datalad/pull/7175#discussion_r1203643207

datalad/tests/utils_pytest.py
yarikoptic: Good digging.  I think it is 
- worth indeed just not bother about `\n` in directories for now
- report finding (allowed by filesystem and `git` per se - but doesn't work correctly) to `git` folks

christian-monch:
Hi @mih, @yarikoptic. I was trying to fix the CI errors and I get a few problems with git-annex. The problem is that zero-terminated batch input for `git annex addurl` reports success (via json-records), but leads to non-zero exit status, and leaves the repository in an inconsistent state.

The git-annex version: `git-annex version: 10.20230407`


For example (sorry for the german messages):
```
~/tmp/nl-test-1> echo -n -e "http://datasets.datalad.org add-4\x00http://datasets.datalad.org add-5\x00" | git -c diff.ignoreSubmodules=none annex addurl --with-files --json --json-error-messages --batch -z
{"command":"addurl","error-messages":[],"file":"add-4","input":["http://datasets.datalad.org add-4"],"key":"MD5E-s1953--b602c390894a197db3a524c843df4630","note":"to add-4","success":true}
{"command":"addurl","error-messages":[],"file":"add-5","input":["http://datasets.datalad.org add-5"],"key":"MD5E-s1953--b602c390894a197db3a524c843df4630","note":"to add-5","success":true}
git-annex: Batch input parse failure: parsed empty url or filename in input: 
~/tmp/nl-test-1> echo $?
1
~/tmp/nl-test-1> ls -l add-[45]
lrwxrwxrwx 1 cristian cristian 114  6. Jun 04:52 add-4 -> .git/annex/objects/2w/7V/MD5E-s1953--b602c390894a197db3a524c843df4630/MD5E-s1953--b602c390894a197db3a524c843df4630
lrwxrwxrwx 1 cristian cristian 114  6. Jun 04:52 add-5 -> .git/annex/objects/2w/7V/MD5E-s1953--b602c390894a197db3a524c843df4630/MD5E-s1953--b602c390894a197db3a524c843df4630
~/tmp/nl-test-1> git annex drop add-4
Fehler: Pfadspezifikation 'add-4' stimmt mit keinen Git bekannten Dateien überein
Did you forget to 'git add'?
drop: 1 failed
```

So, success is signaled, the links to annex's objects store are created, and the objects are created, But `git annex drop` does not work. Do I terminate the communication with `git annex --batch -z` wrongly? I just close the input stream (same as with newline-terminated batch mode). Thought I ask for your opinions before filing a bug report with git-annex.


In contrast to the above, it seems to work fine with new-line terminated batch mode:
```
~/tmp/nl-test-1> echo -n -e "http://datasets.datalad.org add-100\nhttp://datasets.datalad.org add-101\n" | git -c diff.ignoreSubmodules=none annex addurl --with-files --json --json-error-messages --batch
{"command":"addurl","error-messages":[],"file":"add-100","input":["http://datasets.datalad.org add-100"],"key":"MD5E-s1953--b602c390894a197db3a524c843df4630","note":"to add-100","success":true}
{"command":"addurl","error-messages":[],"file":"add-101","input":["http://datasets.datalad.org add-101"],"key":"MD5E-s1953--b602c390894a197db3a524c843df4630","note":"to add-101","success":true}
~/tmp/nl-test-1> echo $?
0
~/tmp/nl-test-1> git annex drop add-100
drop add-100 ok
(recording state in git...)
```

PR #7181, opened 2022-11-16
Comply to Posix-like clone URL formats on Windows

### LINKED ISSUE #7180: Erroneous git url construction with relative Windows paths for subdatasets
I found a Windows issue while looking into fixing #3538.
If we clone a local relative Windows path into another dataset on Windows, the remote git url of the resulting subdataset will be butchered - also with git on its own:

```
C:\Users\adina\AppData\Local\Temp\datalad_temp_1szhiauj\ds>git clone ..\origin dummy
Cloning into 'dummy'...
done.

C:\Users\adina\AppData\Local\Temp\datalad_temp_1szhiauj\ds>git -C dummy config --list
[...]
remote.origin.url=C:/Users/adina/AppData/Local/Temp/datalad_temp_1szhiauj/ds/..\origin
```

[Consulting git clone's manpage confirms that the desired format for clone urls are forward slashes only](https://git-scm.com/docs/git-clone#_git_urls).
Because we pass urls consisting of relative paths as Windows paths (with escaped backslashes) on Windows, the result of a DataLad clone is even more messed up:

```
'C:/Users/adina/AppData/Local/Temp/datalad_temp_frvczceh/ds/..\\origin'
```

adswa:
Fixes #7180.

Git clone's url format does not include backslashes.   Therefore, this commits ensures that paths provided to clone on  Windows are compliying to posix. When platform specific windows paths were passed on by clone as git urls, cloning succeeded, but resulting remote urls can became disfunctional. Specifically, in the case of cloning a subdataset from a local relative path, git will append the relative Windows path to an otherwise Posix-confirming absolute path in the remote url in the subdatasets .git/config. (see #7180). This patch should fix this, and add a test.

### DIFF
--- a/datalad/core/distributed/tests/test_clone.py
+++ b/datalad/core/distributed/tests/test_clone.py
@@ -741,6 +741,19 @@ def test_relative_submodule_url(path=None):
         '../../origin')
 
 
+def test_mushy_windows_submodule_url(tmp_path):
+    """Test that subdataset clones from relative urls
+    do not include backslashes (gh-7180)"""
+    Dataset(tmp_path / 'origin').create()
+    ds = Dataset(tmp_path / 'ds').create()
+    with chpwd(ds.path):
+        ds_cloned = ds.clone(
+            source=op.join(op.pardir, 'origin'),
+            path='sources')
+        assert '\\' not in ds_cloned.config.get('remote.dl-test-remote.url')
+
+
+
 @with_tree(tree={"subdir": {}})
 @with_tempfile(mkdir=True)
 def test_local_url_with_fetch(path=None, path_other=None):
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -968,7 +968,21 @@ class GitRepo(CoreGitRepo):
 
         # Massage URL
         url_ri = RI(url) if not isinstance(url, RI) else url
-        if not on_windows:
+        if on_windows:
+            # When we're cloning from a local path on Windows, the URL at
+            # this point is platform-specific (e.g., "..\\origin"). According 
+            # to Git clone's manpage, clone urls can't have backslashes.
+            # While Git does manage to clone a URL with backslashes, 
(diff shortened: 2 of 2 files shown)


datalad/core/distributed/tests/test_clone.py
```diff
@@ -741,6 +741,19 @@ def test_relative_submodule_url(path=None):
...
+def test_mushy_windows_submodule_url(tmp_path):
+    """Test that subdataset clones from relative urls
+    do not include backslashes (gh-7180)"""
+    Dataset(tmp_path / 'origin').create()
+    ds = Dataset(tmp_path / 'ds').create()
+    with chpwd(ds.path):
+        ds_cloned = ds.clone(
+            source=op.join(op.pardir, 'origin'),
+            path='sources')
+        assert '\\' not in ds_cloned.config.get('remote.dl-test-remote.url')
```
bpoldrack, line 753: `dl-test-remote` is a defined constant. Please use it instead of hardcoding names in tests.

bpoldrack (commented):
The `as_posix` approach is certainly right. I suppose the change in `gitrepo.clone` could be in the `RI` class(es) instead, but I'm fine with it at as. Thank you, @adswa!

Just one minor thing to adjust please (see below).

datalad/core/distributed/tests/test_clone.py
adswa: thx for catching that. I wanted to squash the fix into [6d1d129](https://github.com/datalad/datalad/pull/7181/commits/6d1d129280bc99590fe7a11c4981739ccb087102), but that messed up the merge commit in between - so now its a standalone commit [305a256](https://github.com/datalad/datalad/pull/7181/commits/305a256e1a2db1d7d8af59d3f5e4200012070c7f)

bpoldrack (approved):
Thank you, @adswa. Reran failing Travis build - all good, let's proceed.

yarikoptic-gitmate:

PR #7187, opened 2022-11-17
Fix broken submodule URLs generated from relative paths

adswa:
This change reincarnates parts of a changeset originally proposed in #4026, which did not make it in due to Windows errors I never saw (CI logs weren't preserved), but that I suspect to originate in https://github.com/datalad/datalad/issues/7180 (Git stitching together posix and windows paths into a non-functional URL). Thus, this change sits on top of https://github.com/datalad/datalad/pull/7181, which fixes
these URLs to be fully posix. Based on those fixes, this change wrangles absolute URLs originating from relative paths back into relative paths, to keep them functional.
This change also enables an old test for this problem, marked as a known
failure.

Fixes https://github.com/datalad/datalad/issues/3538

### DIFF
--- a/datalad/core/distributed/tests/test_clone.py
+++ b/datalad/core/distributed/tests/test_clone.py
@@ -722,7 +722,6 @@ def test_cfg_originorigin(path=None):
 
 
 # test fix for gh-2601/gh-3538
-@known_failure
 @with_tempfile()
 def test_relative_submodule_url(path=None):
     Dataset(op.join(path, 'origin')).create()
--- a/datalad/core/distributed/tests/test_clone.py
+++ b/datalad/core/distributed/tests/test_clone.py
@@ -740,19 +740,6 @@ def test_relative_submodule_url(path=None):
         '../../origin')
 
 
-def test_mushy_windows_submodule_url(tmp_path):
-    """Test that subdataset clones from relative urls
-    do not include backslashes (gh-7180)"""
-    Dataset(tmp_path / 'origin').create()
-    ds = Dataset(tmp_path / 'ds').create()
-    with chpwd(ds.path):
-        ds_cloned = ds.clone(
-            source=op.join(op.pardir, 'origin'),
-            path='sources')
-        assert '\\' not in ds_cloned.config.get(f'remote.{DEFAULT_REMOTE}.url')
-
-
-
 @with_tree(tree={"subdir": {}})
 @with_tempfile(mkdir=True)
 def test_local_url_with_fetch(path=None, path_other=None):
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -1062,6 +1062,22 @@ class GitRepo(CoreGitRepo):
                 gr.fsck()
             else:
                 lgr.warning("Experienced issues while cloning: %s", fix_annex)
+        # ensure that Git doesn't mangle relative paths into obscure absolute
+        # paths: https://github.com/datalad/datalad/issues/3538
+        if isinstance(url_ri, PathRI):
+            url_path = Path(url)
+            if not url_path.is_absolute():
+                # get git-created path
+                remote_url = 'remote.' + gr.get_remotes()[0] + '.url'
+                git_url = gr.config.get(remote_url)
+                if Path(git_url).is_absolute():
+                    # Git created an absolute path from a relative URL.
+                    git_url = op.relpath(git_url, gr.path)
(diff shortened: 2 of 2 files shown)


datalad/core/distributed/tests/test_clone.py
```diff
@@ -741,6 +740,19 @@ def test_relative_submodule_url(path=None):
...
+    Dataset(tmp_path / 'origin').create()
+    ds = Dataset(tmp_path / 'ds').create()
+    with chpwd(ds.path):
+        ds_cloned = ds.clone(
+            source=op.join(op.pardir, 'origin'),
+            path='sources')
+        assert '\\' not in ds_cloned.config.get('remote.dl-test-remote.url')
+
+
+
```
yarikoptic, line 755: one too many 
```suggestion
```

datalad/core/distributed/tests/test_clone.py
```diff
@@ -741,6 +740,19 @@ def test_relative_submodule_url(path=None):
...
+def test_mushy_windows_submodule_url(tmp_path):
+    """Test that subdataset clones from relative urls
+    do not include backslashes (gh-7180)"""
+    Dataset(tmp_path / 'origin').create()
+    ds = Dataset(tmp_path / 'ds').create()
+    with chpwd(ds.path):
+        ds_cloned = ds.clone(
+            source=op.join(op.pardir, 'origin'),
+            path='sources')
+        assert '\\' not in ds_cloned.config.get('remote.dl-test-remote.url')
```
yarikoptic, line 752: if we are to ensure that they are POSIX, can't we just compare against expected value fully thus ensuring also that we are still having relative path there, i.e.

```suggestion
        assert ds_cloned.config.get('remote.dl-test-remote.url') == '../origin'
```
or it is not the case?

yarikoptic: but why not just to add this/such a check into prior `test_relative_submodule_url` - why bother with a separate unittest -- just prolongs tests run whenever could be avoided while testing exactly that in the prior test which is succinct and has appropriate purpose/name already

yarikoptic (commented):
Looks LGTM but I don't see the point of a dedicated test -- condition could be placed into prior test

adswa:
Thx for your review, @yarikoptic - I will go apply it on #7181 and then rebase this PR on it.

datalad/core/distributed/tests/test_clone.py
adswa: The unit test is not meant as a dedicated unit test for https://github.com/datalad/datalad/issues/3538 (that one is indeed already present directly above), but a test for https://github.com/datalad/datalad/issues/7180. The reason that this unit test is part of the changeset in this PR is only because fixing #7180 is necessary before fixing #3538. As the fix for #7180 isn't yet merged, I needed to put 3018901b79bf9045c5c7de1de5d71ccbf753deb7 (the only commit relevant to fix #3538) on top of #7181 (which lays the necessary pre-fix on Windows).

The outcome of  ``ds_cloned.config.get(f'remote.{DEFAULT_REMOTE}.url')`` in the test ``test_mushy_windows_submodule_url`` will be different on this branch and on #7181: While on this branch it would indeed be ``../../origin`` (entirely correct and arguably superfluous given the test above ensures the same outcome), the test was originally added in #7181 to ensure that the URL is ``/tmp/pytest-of-adina/pytest-7/test_mushy_windows_submodule_u0/ds/../origin`` instead of ``/tmp/pytest-of-adina/pytest-7/test_mushy_windows_submodule_u0/ds/..\\origin`` (i.e., a testcase for #7180). So sorry for the confusion - all commits but the last one in this PR belong to #7181, and the test might be evaluated best in that PR instead of this one. 

To me it made sense in the context of #7181 to add this test (it was an unnoticed problem so far), but I do see that with the fix from this PR on top the test does not make as much sense anymore - so I would be fine with removing the test completely if you think its not so useful.

adswa: As https://github.com/datalad/datalad/pull/7181 was merged and I rebased the PR, this piece of code isn't part of this PR anymore. However the test was merged before we could discuss whether it is useful to have...

bpoldrack:
Ok, so apparently I missed the test discussion here when merging the PR this was based on.

I agree that the test isn't needed since it does the same thing as the one above. Pushed a removal of that test.
Since otherwise no objections, will go ahead and merge if CI is fine.

Thx, @adswa

Issue #4137, opened 2020-02-12
export-to-figshare would fail to cleanup for dataset with subdatasets


yarikoptic:
Tried on some test dataset laying around:

```shell
(git-annex)lena:/tmp/testds[master]git
$> git status
On branch master
nothing to commit, working tree clean

$> datalad export-to-figshare
[INFO   ] Exporting current tree as an archive under /tmp/testds since figshare does not support directories 
[INFO   ] Uploading /tmp/testds/datalad_None.zip to figshare 
Article
Would you like to create a new article to upload to?  If not - we will list existing articles (choices: yes, no): yes

[INFO   ] Created a new (private) article 11845443 at https://figshare.com/account/articles/11845443. Please visit it, enter additional meta-data and make public 
[INFO   ] 'Registering' /tmp/testds/datalad_None.zip within annex                                                                                                                                                                  
[INFO   ] Adding URL https://ndownloader.figshare.com/files/21708819 for it 
[INFO   ] Registering links back for the content of the archive 
[INFO   ] Adding content of the archive /tmp/testds/datalad_None.zip into annex <AnnexRepo path=/tmp/testds (<class 'datalad.support.annexrepo.AnnexRepo'>)> 
[INFO   ] Initiating special remote datalad-archives 
[INFO   ] Finished adding /tmp/testds/datalad_None.zip: Files processed: 2, removed: 2, +annex: 2 
[WARNING] Removing temporary directory under which extracted files were annexed and should have been removed: /tmp/testds/.dataladw4a0msxt 
CommandError: command '['git', 'rm', '--force', '-r', '--', '.dataladw4a0msxt']' failed with exitcode 128
Failed to run ['git', 'rm', '--force', '-r', '--', '.dataladw4a0msxt'] under '/tmp/testds'. Exit code=128. out= err=fatal: pathspec '.dataladw4a0msxt' did not match any files

fatal: pathspec '.dataladw4a0msxt' did not match any files

$> datalad subdatasets
subdataset(ok): d0/d1/r0 (dataset)

```
That was with datalad 0.12.2 and then reproduced with 0.12.2.dev117

<details><summary>DataLad 0.12.2.dev117 WTF (configuration, datalad, dataset, dependencies, environment, extensions, git-annex, location, metadata_extractors, python, system)</summary>

# WTF
## datalad 
  - full_version: 0.12.2.dev117-g66da5-dirty
  - version: 0.12.2.dev117
## dataset 
  - id: None
[...]

PR #7188, opened 2022-11-17
Bandaids for export-to-figshare to restore functionality

adswa:
I checked whether [this export-to-figshare bug](https://github.com/datalad/datalad/issues/4137) still reproduces. But in its current state, the command is dysfunctional, failing first in an attempt to initialize a progressbar with a pathobject as a label, and then further by giving a path object to ``repo.add()`` (which can't handle pathobjects). The command also still relied on calling ``add_archive_content`` with a deprecated ``annex`` parameter, which would eventually crash in the patch the next extension provides for ``get_allargs_as_kwargs``. This PR adds some very patchy bandaids to make it functional again. But I guess the future should bring slightly more sweeping refurbishments. It would probably also help to have proper tests for this command.

### DIFF
--- a/changelog.d/pr-7188.md
+++ b/changelog.d/pr-7188.md
@@ -1,0 +1,3 @@
+### 🐛 Bug Fixes
+
+- Bandaids for export-to-figshare to restore functionality.  [PR #7188](https://github.com/datalad/datalad/pull/7188) (by [@adswa](https://github.com/adswa))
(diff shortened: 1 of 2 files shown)


yarikoptic:
added "default" changelog. yeah, having a unittest would have been awesome ;) would need mocking away `FigshareRESTLaison.__call__` I guess-- would allow full test without interaction with figshare

yarikoptic (approved):

yarikoptic-gitmate:

PR #7191, opened 2022-11-18
TST: Re-Enable now-passing diff test on Windows

adswa:
``test_path_diff`` only failed on windows CI, not real machines, 3 years ago. I'm curious if this is still the case. Will cancel all tests but a shrunk appveyor suite

### DIFF
--- a/datalad/core/local/tests/test_diff.py
+++ b/datalad/core/local/tests/test_diff.py
@@ -332,8 +332,6 @@ def test_diff_recursive(path=None):
         action='diff', state='modified', path=sub.path, type='dataset')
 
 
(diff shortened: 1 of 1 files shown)


adswa:
wohoo:

```
[gw0] PASSED ..\datalad\core\local\tests\test_diff.py::test_path_diff 
```

yarikoptic:
it seems that the brunch is off `maint` so I will change the base for the PR since IMHO it is worth being in `maint`

adswa:
thanks much, and sorry for the wrong base branch, I'll close this in favour of your #7194. (also thanks for the reminder about the env var - I keep forgetting about this)

Issue #5085, opened 2020-10-24
Invalid path handling test_ria_postclonecfg


mih:
Revealed by #5080 

```
======================================================================
1541ERROR: Failure: OSError ([WinError 123] The filename, directory name, or volume label syntax is incorrect: '\\C:\\Users\\appveyor\\AppData\\Local\\Temp\\1\\datalad_temp_0e20xmuu\\ria-layout-version')
1542----------------------------------------------------------------------
1543Traceback (most recent call last):
1544  File "C:\Miniconda35\envs\test-environment\lib\site-packages\nose\failure.py", line 39, in runTest
1545    raise self.exc_val.with_traceback(self.tb)
1546  File "C:\Miniconda35\envs\test-environment\lib\site-packages\nose\loader.py", line 252, in generate
1547    for test in g():
1548  File "C:\Miniconda35\envs\test-environment\lib\site-packages\datalad\core\distributed\tests\test_clone.py", line 941, in test_ria_postclonecfg
1549    id = _postclonetest_prepare(lcl, store)
1550  File "C:\Miniconda35\envs\test-environment\lib\site-packages\datalad\tests\utils.py", line 750, in _wrap_with_tempfile
1551    return t(*(arg + (filename,)), **kw)
1552  File "C:\Miniconda35\envs\test-environment\lib\site-packages\datalad\core\distributed\tests\test_clone.py", line 904, in _postclonetest_prepare
1553    d.create_sibling_ria(upl_url, "store")
1554  File "C:\Miniconda35\envs\test-environment\lib\site-packages\datalad\distribution\dataset.py", line 503, in apply_func
1555    return f(**kwargs)
1556  File "C:\Miniconda35\envs\test-environment\lib\site-packages\datalad\interface\utils.py", line 495, in eval_func
1557    return return_func(generator_func)(*args, **kwargs)
1558  File "C:\Miniconda35\envs\test-environment\lib\site-packages\datalad\interface\utils.py", line 483, in return_func
1559    results = list(results)
1560  File "C:\Miniconda35\envs\test-environment\lib\site-packages\datalad\interface\utils.py", line 414, in generator_func
1561    allkwargs):
1562  File "C:\Miniconda35\envs\test-environment\lib\site-packages\datalad\interface\utils.py", line 562, in _process_results
1563    for res in results:
1564  File "C:\Miniconda35\envs\test-environment\lib\site-packages\datalad\distributed\create_sibling_ria.py", line 343, in __call__
1565    '1')
[...]

PR #5136, opened 2020-11-08
Fix path and URL handling in ORA remote and associated tests

mih:
This fixes #5085 but it does not make the relevant test complete due to #5134. However, with these changes, I can manually create RIA siblings, push them, clone them, and get data from them -- on windows.

```
(base) C:\Users\mih\TMP>datalad create someds
[INFO] Creating a new annex repo at C:\Users\mih\TMP\someds
[INFO] Detected a filesystem without fifo support.
[INFO] Disabling ssh connection caching.
[INFO] Detected a crippled filesystem.
[INFO] Scanning for unlocked files (this may take some time)
[INFO] Entering an adjusted branch where files are unlocked as this filesystem does not support locked files.
[INFO] Switched to branch 'adjusted/master(unlocked)'
create(ok): C:\Users\mih\TMP\someds (dataset)

(base) C:\Users\mih\TMP>datalad -C someds create-sibling-ria -s store ria+file://C/Users/mih/TMP/RIA
[INFO] create siblings 'store' and 'store-storage' ...
[INFO] Fetching updates for Dataset("C:\Users\mih\TMP\someds")
[INFO] Configure additional publication dependency on "store-storage"
create-sibling-ria(ok): C:\Users\mih\TMP\someds (dataset)

(base) C:\Users\mih\TMP>echo 123 > someds\dummy.txt

(base) C:\Users\mih\TMP>datalad -C someds save
add(ok): dummy.txt (file)
save(ok): . (dataset)
action summary:
  add (ok: 1)
  save (ok: 1)

(base) C:\Users\mih\TMP>datalad -C someds push --to store
copy(ok): dummy.txt (file) [to store-storage...]
publish(ok): . (dataset) [refs/heads/master->store:refs/heads/master [new branch]]
publish(ok): . (dataset) [refs/heads/git-annex->store:refs/heads/git-annex [new branch]]

(base) C:\Users\mih\TMP>cat someds\.datalad\config
[datalad "dataset"]
        id = 7b95bb8c-750f-4e2e-8e91-a083af8b2b13

(base) C:\Users\mih\TMP>datalad clone ria+file:///C/Users/mih/TMP/RIA#7b95bb8c-750f-4e2e-8e91-a083af8b2b13 from_ria
[INFO] Detected a filesystem without fifo support.
[INFO] Disabling ssh connection caching.
[INFO] Detected a crippled filesystem.
[INFO] Scanning for unlocked files (this may take some time)
[...]

Issue #5134, opened 2020-11-08
git-annex does not enable autoenable=true special remotes on windows


mih:
```
(Pdb) riaclone.siblings()
.: here(+) [git]
.: origin(-) [file:///C/Users/mih/AppData/Local/Temp/datalad_temp_t_l0fc4o/cb5/d3fcb-1391-4078-9039-0eb6b568c8a1 (git)]
[{'action': 'query-sibling', 'path': 'C:\\Users\\mih\\AppData\\Local\\Temp\\datalad_temp_xyg24gdd', 'type': 'sibling', 'refds': 'C:\\Users\\mih\\AppData\\Local\\Temp\\datalad_temp_xyg24gdd', 'name': 'here', 'annex-uuid': '228ec9f6-99aa-4b8c-a7d2-82f3d23efd75', 'annex-bare': 'false', 'annex-version': '8', 'available_local_disk_space': 459097722304, 'annex-description': 'mih@bnbdatalad:~/AppData\\Local\\Temp\\datalad_temp_xyg24gdd', 'status': 'ok'}, 
 {'action': 'query-sibling', 'path': 'C:\\Users\\mih\\AppData\\Local\\Temp\\datalad_temp_xyg24gdd', 'type': 'sibling', 'refds': 'C:\\Users\\mih\\AppData\\Local\\Temp\\datalad_temp_xyg24gdd', 'name': 'origin', 'url': 'file:///C/Users/mih/AppData/Local/Temp/datalad_temp_t_l0fc4o/cb5/d3fcb-1391-4078-9039-0eb6b568c8a1', 'fetch': '+refs/heads/*:refs/remotes/origin/*', 'annex-ignore': True, 'status': 'ok'}]
```

but it is configured

```
(Pdb) riaclone.repo.call_git(['cat-file', '-p', 'git-annex:remote.log'])
[...]

PR #7193, opened 2022-11-18
TMP/TST: check if ria test works under windows - blocking issue has b…

adswa:
https://github.com/datalad/datalad/pull/5136 set out to fix https://github.com/datalad/datalad/issues/5085, but noted that a critical test could not be enabled on windows because https://github.com/datalad/datalad/issues/5134 was blocking. Since the blocker was closed a while ago, I wanted to check if the tests runs now. Will cancel all non appveyor tests

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -66,24 +66,24 @@ environment:
     # is a need for debugging
 
     # Ubuntu core tests
-    - ID: Ubu20core
-      # ~30min
-      DTS: >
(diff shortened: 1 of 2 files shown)


adswa:
Mhh, it seems not. I briefly looked into the test on a Windows machine, and interestingly the special remote that doesn't report the "not autoenabled" info message did not get a .git/config section in the clone. Other ORA special remotes had a section, however.

PR #7203, opened 2022-11-28
DOC: Use standard key name 'error_message' in result record design docs

### LINKED ISSUE #6620: Design docs: `error_message` vs `error-messages` in result records
The [design docs on Result records ](https://docs.datalad.org/en/latest/design/result_records.html) specify `error-messages` as one of the optional fields in a result record. They also say that the `get_status_dict()` function can be used to simplify creation of result records. Yet, this function has `error_message` instead:

https://github.com/datalad/datalad/blob/6b4280dcb3e31a2efc6b8d8be544655c08992982/datalad/interface/results.py#L101-L102

I think that design docs should be saying `error_message` instead.

However, I'm not 100% sure, so I wanted to confirm with @datalad/developers before making a one-liner PR.

---
For context:
- Git-annex uses `error-messages` in its JSON outputs (at least for get & drop that I checked - [git annex docs](https://git-annex.branchable.com/tips/using_git-annex_from_your_program/) only say "The format of git-annex's JSON output is not documented in full, because it varies from command to command. The shape is typically the same, but a few commands have a more custom JSON. Try a command and see what JSON it outputs and go from there" - so I tried and went from there)
- I searched DataLad codebase with `git grep` and it is *my impression* that `error_message` is preferred by us:
  - `error_message` (variable / key) and `get_status_dict` are used in several places
  - `error-messages` (variable / key) is also used in several places, but for most of them the context is clearly dealing with raw-ish git annex output, either directly (e.g. checking output of `repo._call_annex_records`) or to join error messages and place them in the `error_message` field for result records down the line
  - other possible combinations luckily don't exist:
    - `error-message` (hyphen, singular) is not present anywhere
    - `error_messages` is present once but probably as a typo #6577

adswa:
This is a small documentation fix for an old issue I found fully dissected, but yet unaddressed:
As elaborated in detail by @mslw in #6620, the common keyname for an error message in a result record is ``error_message``. The Design document on result records however spells this key as ``error-messages``, not as ``error_message`` as understood by ``generic_result_renderer()``. This commit changes the design document to use error_message. Fixes #6620

I would appreciate a brief review by @mslw to check if I understood his issue correctly.

### DIFF
--- a/docs/source/design/result_records.rst
+++ b/docs/source/design/result_records.rst
@@ -173,8 +173,8 @@ A string label categorizing the state of an entity. Common values are:
 - ``present``
 
 
-``error-messages``
-------------------
+``error_message``
+-----------------
 
 List of any error messages that were captured or produced while achieving a
 result.


bpoldrack (approved):

bpoldrack:
It's rather trivial - I'll go ahead and merge. Thank you, @adswa!

Issue #7197, opened 2022-11-21
test_nodrop_symlinked_annex fails with recent git-annex on windows


yarikoptic:
In addition to fresh #7184 and #7185 now (e.g. in https://github.com/datalad/git-annex/actions/runs/3511326704/jobs/5882167483) we got one more test failing now

```
FAILED ..\datalad\distributed\tests\test_drop.py::test_nodrop_symlinked_annex - datalad.support.exceptions.IncompleteResultsError: Command did not complete successfully. 1 failed:

[{'action': 'drop',
  'gitshasum': 'c38a697f0f89f0ac4755e8ca2bd2e95667190472',
  'message': 'cannot drop modified content, save first',
  'parentds': 'C:\\Users\\runneradmin\\AppData\\Local\\Temp\\datalad_temp_azsnbmet',
  'path': 'C:\\Users\\runneradmin\\AppData\\Local\\Temp\\datalad_temp_azsnbmet\\file1',
  'prev_gitshasum': 'c38a697f0f89f0ac4755e8ca2bd2e95667190472',
  'refds': 'C:\\Users\\runneradmin\\AppData\\Local\\Temp\\datalad_temp_azsnbmet',
  'state': 'modified',
  'status': 'impossible',
  'type': 'file'}]
```

seems to happen only once for now

```
(git)smaug:/mnt/datasets/datalad/ci/git-annex/builds/2022/11[master]git
$> git grep 'FAIL.*test_nodrop_symlinked_annex'         
cron-20221121/build-windows.yaml-857-b1412209-failed/1_test-datalad (master).txt:2022-11-21T06:00:53.8107286Z FAILED ..\datalad\distributed\tests\test_drop.py::test_nodrop_symlinked_annex - datalad.support.exceptions.IncompleteResultsError: Command did not complete successfully. 1 failed:
cron-20221121/build-windows.yaml-857-b1412209-failed/test-datalad (master)/22_Run datalad tests.txt:2022-11-21T06:00:53.8107281Z FAILED ..\datalad\distributed\tests\test_drop.py::test_nodrop_symlinked_annex - datalad.support.exceptions.IncompleteResultsError: Command did not complete successfully. 1 failed:
```

Note: https://github.com/datalad/datalad/issues/6960 talks about this test origin

Issue #7185, opened 2022-11-17
test_ria_push fails with recent git-annex on windows


yarikoptic:
Reincarnation of #6333 which was recently closed (error message is a bit different) and sattelite to another fresh windows specific #7184 which started to happen at the same time

```
2022-11-16T05:58:33.9825892Z FAILED ..\datalad\core\distributed\tests\test_push.py::test_ria_push - datalad.support.exceptions.IncompleteResultsError: Command did not complete successfully. 1 failed:
2022-11-16T05:58:33.9826430Z 
2022-11-16T05:58:33.9826624Z [{'action': 'create-sibling-ria',
2022-11-16T05:58:33.9826996Z   'message': 'initremote failed.\n'
2022-11-16T05:58:33.9827353Z              'stdout: initremote datastore-storage \n'
2022-11-16T05:58:33.9827708Z              'failed\n'
2022-11-16T05:58:33.9828006Z              '\n'
2022-11-16T05:58:33.9828350Z              'stderr: git-annex: RIA store unavailable. -caused by- '
2022-11-16T05:58:33.9828873Z              'file:///C/Users/runneradmin/AppData/Local/Temp/datalad_temp_tree_d1u0uemvria-layout-version '
2022-11-16T05:58:33.9829386Z              'not found. -caused by- [Errno 2] No such file or directory: '
2022-11-16T05:58:33.9829981Z              "'\\\\C\\\\Users\\\\runneradmin\\\\AppData\\\\Local\\\\Temp\\\\datalad_temp_tree_d1u0uemv\\\\ria-layout-version'\r\n"
2022-11-16T05:58:33.9830508Z              'initremote: 1 failed\r\n',
2022-11-16T05:58:33.9830927Z   'path': 'C:\\Users\\runneradmin\\AppData\\Local\\Temp\\datalad_temp_vx98823b',
2022-11-16T05:58:33.9831400Z   'status': 'error',
2022-11-16T05:58:33.9831718Z   'type': 'dataset'}]
```
see e.g. https://github.com/datalad/git-annex/actions/runs/3476245614/jobs/5811450861

Issue #7184, opened 2022-11-17
test_nested_pushclone_cycle_allplatforms started to fail on windows


yarikoptic:
when tested against recent git-annex build - in the past two days (note: windows build was broken for awhile)
```shell
(git)smaug:/mnt/datasets/datalad/ci/git-annex/builds/2022/11[master]git
$> git grep 'FAILED.*test_nested_pushclone_cycle_allplatforms'
cron-20221115/build-windows.yaml-851-b1412209-failed/1_test-datalad (master).txt:2022-11-15T05:55:56.7195394Z FAILED ..\datalad\core\distributed\tests\test_push.py::test_nested_pushclone_cycle_allplatforms - AssertionError: assert {'adjusted/dl..., 'git-annex'} == {'dl-test-bra..., 'git-annex'}
cron-20221115/build-windows.yaml-851-b1412209-failed/test-datalad (master)/22_Run datalad tests.txt:2022-11-15T05:55:56.7195389Z FAILED ..\datalad\core\distributed\tests\test_push.py::test_nested_pushclone_cycle_allplatforms - AssertionError: assert {'adjusted/dl..., 'git-annex'} == {'dl-test-bra..., 'git-annex'}
cron-20221116/build-windows.yaml-852-b1412209-failed/1_test-datalad (master).txt:2022-11-16T05:58:33.9832314Z FAILED ..\datalad\core\distributed\tests\test_push.py::test_nested_pushclone_cycle_allplatforms - AssertionError: assert {'adjusted/dl..., 'git-annex'} == {'dl-test-bra..., 'git-annex'}
cron-20221116/build-windows.yaml-852-b1412209-failed/test-datalad (master)/22_Run datalad tests.txt:2022-11-16T05:58:33.9832268Z FAILED ..\datalad\core\distributed\tests\test_push.py::test_nested_pushclone_cycle_allplatforms - AssertionError: assert {'adjusted/dl..., 'git-annex'} == {'dl-test-bra..., 'git-annex'}
```

See e.g.
https://github.com/datalad/git-annex/actions/runs/3476245614/jobs/5811450861

```
FAILED ..\datalad\core\distributed\tests\test_push.py::test_nested_pushclone_cycle_allplatforms - AssertionError: assert {'adjusted/dl..., 'git-annex'} == {'dl-test-bra..., 'git-annex'}
  Extra items in the left set:
  'adjusted/dl-test-branch(unlocked)'
  Full diff:
[...]

PR #7205, opened 2022-11-28
Mark some tests as known fails on github CI windows

yarikoptic:
Ref: #7184, #7185, #7197

Need to do that since there is no human force available to investigate and address them currently and we might this way miss more failures on datalad/git-annex CI setup.

- Note: that I marked them as known only for github CI, so they might resurface on appveyor whenever a new release of git-annex comes out. Then either they would need to be addressed or marked as generally known to fail on windows

- Another note: PR is against `master` since seems there is no such fails on `maint`. See e.g. https://github.com/datalad/git-annex/actions/runs/3561508598 for comparison

- I think I will not bother with changelog entry for this change

### DIFF
--- a/datalad/core/distributed/tests/test_push.py
+++ b/datalad/core/distributed/tests/test_push.py
@@ -517,6 +517,7 @@ def test_force_checkdatapresent(srcpath=None, dstpath=None):
                       message='Slated for transport, but no content present')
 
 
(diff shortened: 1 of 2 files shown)


bpoldrack:
I'm fine with this for now. But let's then also close those issues. They are linked in the code, so whenever someone is working on reducing the skips on windows again, they will them.

PR #7206, opened 2022-11-29
Fix localpath handling for file-URLs on different OS, especially on Windows

### LINKED ISSUE #7186: `PathRI` behavior incorrect/not helpful on Windows
Demo:

```py
>>> pathlib.PureWindowsPath("C:\\Windows").as_uri()
'file:///C:/Windows'

>>> RI("file:///c:/Windows").localpath
'/c:/Windows'

>>> pathlib.PureWindowsPath('/c:/Windows')
PureWindowsPath('/c:/Windows')
```

Ergo, it get's transformed into an unusable string.

A better approach would be:

```
>>> nturl2path.url2pathname(urlparse("file:///c:/Windows").path)
'C:\\Windows'
```

`nturl2path.url2pathname` is called by `urllib.request.url2pathname` on windows.

christian-monch:
Fixes #7186 

The current code base treats the `path` component of `file`-URLs not according to the conventions described in the relevant standard, i.e. [RFC8089](https://www.ietf.org/rfc/rfc8089.html) in [Appendix D.2](https://www.ietf.org/rfc/rfc8089.html#appendix-D.2) and [Appendix E.2](https://www.ietf.org/rfc/rfc8089.html#appendix-E.2) (funnily enough Append E.2 defines some "nonstandard techniques for interacting with DOS- or Windows-like file paths and URIs").

As @mih pointed out, there is support for the conventions described in RFC8089 Appendix D.2 and RFC8089 Appendix E.2 in python, implemented in `nturl2path`, which is transparently used in `urllib.request.url2pathname` on Windows systems.

This PR uses `urllib.request.url2pathname` to convert the path of a file-URL to a local path. This changes the behavior of `RI.localpath` if an `RI` is an instance of `URL`, and the URL-scheme is `file`, In this case `localpath` will return a result from `urllib.request.url2pathname`. This means for example that the following holds:

* On Posix-like systems: `RI('file:/c:/Windows').localpath == '/c:/Windows'`
* On Windows-like systems: `RI('file:/c:/Windows').localpath == 'C:\\Windows'`

Note that the drive letter is converted to a canonical "upper case" representation.

### DIFF
--- a/datalad/support/network.py
+++ b/datalad/support/network.py
@@ -67,12 +67,28 @@ from datalad.utils import (
 
 
 def local_path_representation(path: str) -> str:
-    """Return an OS-specific representation of a "file:" URI-style path
+    """Return an OS-specific representation of a Posix-style path
 
     With a posix path in the form of "a/b" this function will return "a/b" on
     Unix-like operating systems and "a\\b" on Windows-style operating systems.
     """
     return str(Path(path))
+
+
+def local_url_path_representation(url_path: str) -> str:
+    """Return an OS-specific representation of the path component in a file:-URL
+
+    With a path component like "/c:/Windows" (i.e. from a URL that reads
+    "file:///c:/Windows"), this function will return "/c:/Windows" on a
+    Unix-like operating systems and "C:\\Windows" on Windows-like operating
+    systems.
+    """
+    return url2pathname(url_path)
+
+
+def local_path_from_url(url: str) -> str:
+    """Parse the url and extract an OS-specific local path representation"""
+    return local_url_path_representation(urlparse(url).path)
 
 
 def is_windows_path(path):
--- a/datalad/support/network.py
+++ b/datalad/support/network.py
@@ -38,7 +38,11 @@ from urllib.parse import (
     urlsplit,
     urlunparse,
 )
-from urllib.request import Request
+from urllib.request import (
+    Request,
+    pathname2url,
+    url2pathname,
+)
 
 import iso8601
 
@@ -61,6 +65,14 @@ from datalad.utils import (
 # !!! Lazily import requests where needed -- needs 30ms or so
 # import requests
 
+
+def local_path_representation(path: str) -> str:
+    """Return an OS-specific representation of a "file:" URI-style path
+
+    With a posix path in the form of "a/b" this function will return "a/b" on
+    Unix-like operating systems and "a\\b" on Windows-style operating systems.
+    """
+    return str(Path(path))
 
 
 def is_windows_path(path):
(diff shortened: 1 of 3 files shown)


datalad/support/network.py
```diff
@@ -706,7 +734,22 @@ def localpath(self):
...
             raise ValueError("file:// URL does not point to 'localhost'")
-        return self.path
+
+        # RFC1738 and RFC3986 both forbid unescaped backslash characters in
+        # URLs, and therefore also in the path-component of file:-URLs. We
+        # assume here that any backslash present in a file-URL is a relict of a
+        # verbatim copy of a Windows-style path.
+        unified_path = self.path.replace('\\', '/')
+        return url2pathname(unified_path)
+        print("---: self.path:", self.path, file=sys.stderr)
```
bpoldrack, line 744: I assume the `print`s are left-overs?

datalad/support/network.py
```diff
@@ -706,7 +734,22 @@ def localpath(self):
...
+        print("|1 : unified_path:", unified_path, file=sys.stderr)
+        print("|2 : url2pathname(unified_path):", url2pathname(unified_path), file=sys.stderr)
+        print("|3 : pathname2url(url2pathname(unified_path)):", pathname2url(url2pathname(unified_path)), file=sys.stderr)
+        print("|r : url2pathname(unified_path):", url2pathname(unified_path), file=sys.stderr)
+        pathname = url2pathname(unified_path)
+        if on_windows:
+            return pathname.replace("\\", "/")
```
bpoldrack, line 751: This replace confuses me. Shouldn't it be the other way around, returning the backslashes on windows? Not sure what I am missing here.

datalad/support/network.py
christian-monch: Yep, Thanks

datalad/support/network.py
christian-monch: Sorry, that code is not executed. There is however a similar call in line 747 with an explanation in the leading comment

bpoldrack: Ah, okay. There the replace is happening in the URL, not the returned path like this line here suggested.

All good then!

christian-monch:
> Just in case you are looking for more related stuff to iron out: There's `support.network.get_local_file_urlspecial` casing windows. It's more involved b/c it's also about what git vs git-annex can actually consume, but an educated look at it would be great!

Will take a look at it

bpoldrack:
Ok, cleaned up, all green - thank you @christian-monch!

yarikoptic-gitmate:

PR #7194, opened 2022-11-18
Reenable now passing test_path_diff test on Windows

yarikoptic:
Closes #3725

It is https://github.com/datalad/datalad/pull/7191 but against maint since IMHO it is worth there

PR #7208, opened 2022-11-30
Skip changelog commits on appveyor

bpoldrack:


yarikoptic:
Let's see how it would look...

yarikoptic:
hm, why changelog snippet addition workflow didn't run after I added the label. @jwodder - any immediate ideas?

yarikoptic:
Remove/adding cycle helped - we got a changelog.  But now it shows disadvantage of this approach -- appveyor is no longer shown at all  among checks. We would need to expand prior sweep of runs to check it.  Bad DX. So, may be we are doomed to not bother skipping this way :-/

bpoldrack:
@yarikoptic 
> But now it shows disadvantage of this approach -- appveyor is no longer shown at all among checks. We would need to expand prior sweep of runs to check it. 

Well, already the case for commits to docs/ only. Don't see a problem with that. "Exhausting" our CI resources is more of a pain for me (needing to wait for hours before I can come back to something I worked on).

yarikoptic:
`docs/` only  commits are rare to come. `changelog.d/` commit is to come nearly for every PR.  re wait: I think appveyor is configured already to automagically cancel prior run if next commit was pushed to the PR. I think it happens for both Travis and Appveyor -- check it out e.g. on https://github.com/datalad/datalad/pull/7194 . Only may be github actions do run twice in full.

bpoldrack:
Yes, but you adding the CHANGELOG-missing label in PR A adds a lot of jobs to the appveyor queue making me wait longer in PR B. The point is that a run on a plain changelog commit is superfluous. It may not be followed up by another push that immediately supercedes it.

yarikoptic:
> The point is that a run on a plain changelog commit is superfluous. It may not be followed up by another push that immediately supercedes it.

my point is that a typical situation is the other way around -- changelog commit follows up a commit we actually want to test.   In those cases those jobs in appveyor queue for the initial commit cancel pretty much right away.

bpoldrack:
Ok, then let's not do it.

PR #7210, opened 2022-11-30
Let CLI report on exception causes

### LINKED ISSUE #7163: CLI toplevel exception reporting hides "cause"
While this is not about a particular exception scenario, here is a concrete use case to help illustrate the problem: I have a command that performs input validation. If that fails a uniform `ValueError` is raised, and it is raised `from` the underlying exception representing the specific nature of the validation failure (e.g. `TypeError`, network down, water too cold, ...).

When such a `ValueError` reaches the CLI top-level handler, all users get to learn about the issue is:

```
[ERROR  ] Validation of parameter 'spec' failed (ValueError)
```

And that is despite the fact that `exc.__cause__` is `ValueError: URL is not a string` at this point.

Two approaches would improve this:

1. `CapturedException`'s `.message` would include information on a reported cause.
2. A patch like this:

```py
diff --git a/datalad/cli/main.py b/datalad/cli/main.py
index c5364f4b3..49b09913a 100644
--- a/datalad/cli/main.py
+++ b/datalad/cli/main.py
@@ -212,7 +212,9 @@ def _run_with_exception_handler(cmdlineargs):
             exit_code = 3
         else:
             # some unforeseen problem
-            lgr.error('%s (%s)', ce.message, ce.name)
+            lgr.error('%s (%s)%s',
+                      ce.message, ce.name,
+                      f': {exc.__cause__}' if exc.__cause__ else '')
         sys.exit(exit_code)
 ```

In both cases the output to a user would be more like this:

```
[ERROR  ] Validation of parameter 'spec' failed (ValueError): URL is not a string
```

### PR DESCRIPTION
**Pull request by bpoldrack**
Ensure the already existing `format_exception_with_cause` also works with `traceback.TracebackException` and use this to provide another format method for `CapturedException`. This is then used with unexpectedly bubbled up exceptions by the CLI.

Fixes #7163

FTR:

Actually hard to find a case where I know an exception would bubble up and actually have a `__cause__` in core.

Making one up like this: 

```
 raise ValueError("something is wrong") from RuntimeError("severe problem")
```

Would now be:

```
> datalad wtf
[ERROR  ] something is wrong -caused by- severe problem 
```

Whereas 

```
 raise ValueError("something is wrong") from RuntimeError
 ```
 
 Would lead to 
 
 ```
> datalad wtf
[ERROR  ] something is wrong -caused by- RuntimeError
```

Rational for sparing the class name when there's a message, is taken from the existing implementation used by special remotes: 

> For each exception in the chain either the str() of it is taken, or the
> class name of the exception, with the aim to generate a simple and
> comprehensible description that can be used in user-facing messages.
> It is explicitly not aiming to provide a detailed/comprehensive source
> of information for in-depth debugging.

### DIFF
--- a/changelog.d/pr-7210.md
+++ b/changelog.d/pr-7210.md
@@ -1,0 +1,7 @@
+### 🚀 Enhancements and New Features
+
+- Exceptions bubbling up through CLI are now reported on including their chain
+  of __cause__.
+  Fixes [#7163](https://github.com/datalad/datalad/issues/7163) via
+  [PR #7210](https://github.com/datalad/datalad/pull/7210)
+  (by [@bpoldrack](https://github.com/bpoldrack))
--- a/datalad/cli/main.py
+++ b/datalad/cli/main.py
@@ -216,7 +216,7 @@ def _run_with_exception_handler(cmdlineargs):
             exit_code = 3
         else:
             # some unforeseen problem
-            lgr.error('%s (%s)', ce.message, ce.name)
+            lgr.error('%s', ce.format_with_cause())
         sys.exit(exit_code)
 
 
--- a/datalad/support/exceptions.py
+++ b/datalad/support/exceptions.py
@@ -93,6 +93,12 @@ class CapturedException(object):
         """
         return self.name + '(' + self.message + ')'
 
+    def format_with_cause(self):
+        """Returns a representation of the original exception including the
+        underlying causes"""
+
+        return format_exception_with_cause(self.tb)
+
     @property
     def message(self):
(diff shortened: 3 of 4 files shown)


yarikoptic:
no objections, although would have been nice to see in comparison to how it would look now, but I would just assume that now it would just not say that exception was caused by another exception.  so, as such, providing information about chain would might come useful (although without traceback/location information -- of limited immediate use).

bpoldrack:
> I would just assume that now it would just not say that exception was caused by another exception

Yes, w/o the PR causes are not reported.

yarikoptic-gitmate:

PR #7216, opened 2022-12-07
Clean up conversion between local file system paths and file-URL path components

### LINKED ISSUE #7212: `create-sibling-ria` cannot handle `(ria+)file://` URLs on windows
It fails with

```
C:\Python39-x64\lib\site-packages\datalad\distributed\create_sibling_ria.py:369: in __call__
2651    io.read_file(Path(base_path) / 'ria-layout-version')
2652_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
2653
2654self = <datalad.distributed.ora_remote.LocalIO object at 0x0000014965EDAFA0>
2655file_path = WindowsPath('/C:/DLTMP/pytest-of-appveyor/pytest-0/test_uncurl_ria_access0/ria/ria-layout-version')
2656
2657    def read_file(self, file_path):
2658    
2659>       with open(str(file_path), 'r') as f:
2660E       OSError: [Errno 22] Invalid argument: '\\C:\\DLTMP\\pytest-of-appveyor\\pytest-0\\test_uncurl_ria_access0\\ria\\ria-layout-version'
2661
2662
```

The key is this path `\\C:\\DLTMP\\pytest-of-appveyor\\pytest-0\\test_uncurl_ria_access0\\ria\\ria-layout-version`, which it derived from the URL, and clearly is not a windows path. The input URL here was generated via `Path.as_uri()` (and prepending `ria+`).

See here for a small helper that perform such conversion correctly, based on `urllib`.

https://github.com/datalad/datalad-next/blob/ceeac85b4c03cc454c17f091594d923d28ecd20a/datalad_next/url_operations/file.py#L39-L43

christian-monch:
Fixes #7212 

This PR introduces functions to convert the path component of a file-URL to local file system paths, independent from the OS.

We should probably perform a proper check of the code base to find the places in which we handle file-URL path components naively, i.e. take them verbatim as local paths. Even with automated conversion of `/` in path components to `\` on Windows-like OS, e.g. when using `Path()`, a simple approach would fail because it does not take Windows-anchor encoding into account. For example the local Windows path `C:\\Windows` might be encoded as `/c:/Windows` in the path component of a file-URL.

### DIFF
--- a/datalad/customremotes/ria_utils.py
+++ b/datalad/customremotes/ria_utils.py
@@ -9,9 +9,18 @@
 """Helper for RIA stores
 
 """
+from __future__ import annotations
 
 import logging
 from pathlib import Path
+from urllib.parse import (
+    unquote,
+    urlparse,
+)
+from urllib.request import (
+    url2pathname,
+    pathname2url,
+)
 
 
 lgr = logging.getLogger('datalad.customremotes.ria_utils')
@@ -87,6 +96,7 @@ def verify_ria_url(url, cfg):
       (host, base-path, rewritten url)
       `host` is not just a hostname, but is a stub URL that may also contain
       username, password, and port, if specified in a given URL.
+      `base-path` is the path component of the url
     """
     from datalad.config import rewrite_url
     from datalad.support.network import URL
@@ -243,3 +253,16 @@ def create_ds_in_store(io, base_path, dsid, obj_version, store_version,
         except FileExistsError:
             lgr.warning("Alias %r already exists in the RIA store, not adding an "
                         "alias.", alias)
+
+
+def local_path2url_path(local_path: str) -> str:
+    """Convert a local path into a URL path part"""
+    if not local_path:
+        return '/'
+    return urlparse(unquote(pathname2url(local_path))).path
+
+
+def url_path2local_path(url_path: str) -> str:
+    if not url_path:
+        return str(Path('/'))
+    return url2pathname(url_path)
--- a/datalad/customremotes/ria_utils.py
+++ b/datalad/customremotes/ria_utils.py
@@ -9,22 +9,8 @@
 """Helper for RIA stores
 
 """
-from __future__ import annotations
-
 import logging
-import os
-import re
-from pathlib import (
-    Path,
-    PurePosixPath,
-)
-from urllib.parse import (
-    quote,
-    urlparse,
-)
-from urllib.request import url2pathname
-
-from datalad.utils import on_windows
+from pathlib import Path
 
 
 lgr = logging.getLogger('datalad.customremotes.ria_utils')
@@ -255,59 +241,3 @@ def create_ds_in_store(io, base_path, dsid, obj_version, store_version,
         except FileExistsError:
             lgr.warning("Alias %r already exists in the RIA store, not adding an "
                         "alias.", alias)
-
-
-def local_path2url_path(local_path: str,
-                        auto_resolve: bool = False
-                        ) -> str:
-    """Convert a local path into a URL path component"""
-    if not local_path:
-        if not auto_resolve:
-            raise ValueError("cannot convert empty local path to URL path")
-        local_path = os.getcwd()
-
-    url = urlparse(Path(local_path).as_uri())
-    if url.hostname:
-        raise ValueError(
-            f"cannot convert remote path to an URL path: {local_path}")
-    return url.path
-
-
-def url_path2local_path(url_path: str | PurePosixPath,
-                        make_absolute: bool = True,
-                        ) -> str | Path:
-
-    if isinstance(url_path, PurePosixPath):
-        return_path = True
-        url_path = str(url_path)
-    else:
-        return_path = False
-
-    if not url_path or not url_path.startswith("/"):
-        # We expect a 'path-absolute' as defined in RFC 3986, therefore the
-        # path must begin with a slash.
-        raise ValueError(
-            f"url path does not start with '/': {url_path}, and is therefore "
-            f"not an absolute-path as defined in RFC 8089")
-
-    if url_path.startswith("//"):
-        # We expect a 'path-absolute' as defined in RFC 3986, therefore the
-        # first segment must not be empty, i.e. the path must not start with
-        # two or more slashes.
-        raise ValueError(
-            f"url path has empty first segment: {url_path}, and is therefore "
-            f"not an absolute-path as defined in RFC 8089")
-
-    return (
-        Path(url2pathname(url_path))
-        if return_path
-        else url2pathname(url_path)
-    )
-
-
-def quote_path(path: str, safe: str = "/") -> str:
-    """quote the path component of a URL, takes OS specifics into account"""
-    if on_windows:
-        if re.match("^/[a-zA-Z]:/", path):
-            return path[:3] + quote(path[3:], safe=safe)
-    return quote(path, safe=safe)
--- a/datalad/customremotes/ria_utils.py
+++ b/datalad/customremotes/ria_utils.py
@@ -12,16 +12,14 @@
 from __future__ import annotations
 
 import logging
+import os
 import re
 from pathlib import Path
 from urllib.parse import (
     quote,
     urlparse,
 )
-from urllib.request import (
-    url2pathname,
-    pathname2url,
-)
+from urllib.request import url2pathname
 
 from datalad.utils import on_windows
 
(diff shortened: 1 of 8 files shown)


mih:
Quick comment after skimming the diff: It would be good to extend the tests to make sure that the URLs coming out of `Path.as_uri()` are compatible with what is being tests as supported input.

christian-monch:
> Quick comment after skimming the diff: It would be good to extend the tests to make sure that the URLs coming out of `Path.as_uri()` are compatible with what is being tests as supported input.

Will add tests

christian-monch:
> Quick comment after skimming the diff: It would be good to extend the tests to make sure that the URLs coming out of `Path.as_uri()` are compatible with what is being tests as supported input.

The function `datalad.support.network.get_local_file_url`  contains an assert that assures that the use of `local_path2url_path` and our logic, if 'git`-compatibility is chosen, yields the same results as `Path.as_uri()`.

christian-monch:
The following works for me in Windows:

```
from datalad.distribution.dataset import Dataset

ds = Dataset(f"C:\\TMP\\ddd-1")
ds.create(force=True)
ds.create_sibling_ria(f"ria+file:///C:/TMP/sss-1", "test-sss-1", new_store_ok=True)
```

It creates a RIA store in `C:\TEMP\sss-1` and adds the sibling to the dataset in `C:\TMP\ddd-1`, yielding the following output:

```
[INFO] Detected a filesystem without fifo support. 
[INFO] Disabling ssh connection caching. 
[INFO] Detected a crippled filesystem. 
[INFO] Entering an adjusted branch where files are unlocked as this filesystem does not support locked files. 
[INFO] Switched to branch 'adjusted/master(unlocked)' 
create(ok): . (dataset)
[INFO] Creating a new RIA store at C:\TMP\sss-1 
[INFO] create siblings 'test-sss-1' and 'test-sss-1-storage' ... 
[INFO] Fetching updates for Dataset("C:\TMP\ddd-1") 
update(ok): . (dataset)
update(ok): . (dataset)
[INFO] Configure additional publication dependency on "test-sss-storage" 
create-sibling-ria(ok): C:\TMP\ddd-1 (dataset)
action summary:
  configure-sibling (ok: 1)
  create-sibling-ria (ok: 1)
  update (ok: 1)

Process finished with exit code 0
```

christian-monch:
@mih. It is definitely possible now to create ria siblings. I have not yet completely gone through `ora_remote.py` since @bpoldrack  announced an imminent refactoring. I also did not generally remove `@skip_if_on_windows` from tests. But I think that the current patch is a proper first step to support Windows-like file systems better.

christian-monch:
I would still refactor some stuff in a follow-up PR. For example, introduce the type-alias `UrlPath` for `PurePosixPath`, reduce back and forth conversion by operating on `UrlPath` as long as possible, and add type annotations to prevent confusion when working with the code.

datalad/support/network.py
```diff
@@ -969,38 +1007,43 @@ def is_ssh(ri):
...
-            urlquote(
-                re.sub(r'([a-zA-Z]):', r'\1', path)
-            ))
-    else:
-        # TODO:  need to fix for all the encoding etc
-        furl = str(URL(scheme='file', path=str(path)))
-    return furl
+    path_obj = Path(fname)
+    if not path_obj.is_absolute():
+        if allow_relative_path is True:
```
yarikoptic, line 1032:
```suggestion
        if allow_relative_path:
```
to not suggest some other special `True` value (like non empty list etc) handling.

datalad/support/network.py
```diff
@@ -969,38 +1007,43 @@ def is_ssh(ri):
...
+
+    url_path = local_path2url_path(fname, auto_resolve=False)
+    if on_windows and compatibility != "git":
+        # Work around the way in which git-annex interprets file URLs on
+        # Windows. This code path will put the path anchor, e.g. `C:` of `fname`
+        # into the network location component of the resulting URL.
+        return "file:/" + url_path
+
+    result = "file://" + url_path
+    assert result == Path(fname).as_uri()
```
yarikoptic, line 1045: how strong believe in this here?  I would leave such assertion for tests, and here if really desired -- just added a warning if mismatch detected.  Do we want entire process to crash if this assertion is not fulfilled? (e.g. some trailing `/` gets in etc)

setup.py
```diff
@@ -92,7 +92,7 @@
         'pypandoc',
         # Documentation
         'sphinx>=4.3.0',
-        'sphinx-rtd-theme>=0.5.1"',
+        'sphinx-rtd-theme>=0.5.1',
```
yarikoptic, line 95: odd to see this in the diff since already fixed in `maint`:
```
❯ git describe
0.17.10-2-g23b376756
❯ grep rtd setup.py
        'sphinx-rtd-theme>=0.5.1',
```
but ok

yarikoptic (changes requested):
minor suggestions

setup.py
christian-monch: I was surprised as well but did not dive into it any further.

datalad/support/network.py
```diff
@@ -50,6 +49,7 @@
     cfg,
     consts,
 )
+from datalad.customremotes.ria_utils import local_path2url_path
```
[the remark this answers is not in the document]
christian-monch: Done in fdda723653594b0b63cf353e357bf8f17447ffc2

datalad/support/network.py
```diff
@@ -83,7 +83,8 @@ def local_url_path_representation(url_path: str) -> str:
     Unix-like operating systems and "C:\\Windows" on Windows-like operating
     systems.
     """
-    return url2pathname(url_path)
+    from urllib.parse import quote
```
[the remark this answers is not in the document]
christian-monch: Removed

datalad/support/network.py
```diff
@@ -611,13 +612,21 @@ class URL(RI):
...
         'query',
         'fragment',
     )
 
+    # Only interpreted on Windows. If set to `True`, UNC-names encoded in
+    # file-URLs, e.g. "file://server/share/path", would be considered local
+    # and mapped onto "\\server\share\path". If set to `False`, UNC-names
+    # encoded in file-URLs would not be considered local and cannot be resolved
+    # to a local path.
+    support_unc = False
```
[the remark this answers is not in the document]
christian-monch: The value is to be set by developers or some generic config processing. It should not yet be in a configuration because datalad does not support it yet, and  I am not sure about `git-annex` and Windows shares. So there is currently no point in activating UNC support. Also, if we should support it in the future, there is probably no need to deactivate it from a configuration.

With this in mind, the line documents that UNC could be supported by the class, if the users of the class can handle it. (And I think it is good to give a `False` a name that indicates its purpose. In addition, it allows us to be able to switch UNC support on and off in a single place if we decide to support it in general (no config-option)).

datalad/support/network.py
```diff
@@ -418,7 +419,7 @@ class variable.
     provided as output, e.g.
 
     >>> RI('http://example.com')
-    URL(hostname='example.com', scheme='http')
+    URL(hostname='example.com', netloc='example.com', scheme='http')
```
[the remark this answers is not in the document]
christian-monch: Done in 61992cdda010f848cd59870b2bf254fdc06105c7

datalad/support/network.py
christian-monch: I am quite confident, and I think it might be good to blow instead of continuing with a faulty URI, which might -by chance- lead to data loss. Also, an uncaught exception might expose problems in our code earlier.

But I will change it to a warning,

Done in a194ecc5195db1e636866fe6825e765a3f2bf1b9

datalad/customremotes/ria_utils.py
```diff
@@ -117,11 +117,9 @@ def verify_ria_url(url, cfg):
...
         port=url_ri.port or '',
     )
-    # this != file is critical behavior, if removed, it will ruin the IO selection
-    # in RIARemote!!
-    return host if protocol != 'file' else None, \
-        url_ri.path if url_ri.path else '/', \
-        url
+    # this ``!= 'file'´´ is critical behavior, if removed, it will ruin the IO
+    # selection in RIARemote!!
+    return host if protocol != 'file' else None, url_ri.path or '/', url
```
yarikoptic, line 122: Reading carefully I see that it is not just reformatting but changing `if` into `or` expression ... and I think it isn't correct since any non empty tuple eg of `None, url_ri.path`.  was this change needed and what was intervention for it?

datalad/support/network.py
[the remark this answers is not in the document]
yarikoptic: Still not sure why we would have it here then, but ok, let it be.

datalad/support/network.py
```diff
@@ -1103,4 +1152,60 @@ def download_url(url, dest=None, overwrite=False):
...
+
+    url = urlparse(Path(local_path).as_uri())
+    if url.netloc:
+        raise ValueError(
+            f"cannot convert remote path to an URL path: {local_path}")
+    return url.path
+
+
+def url_path2local_path(url_path: str | PurePosixPath,
+                        make_absolute: bool = True,
```
yarikoptic, line 1172: This option seems not used

datalad/support/network.py
```diff
@@ -1103,4 +1152,60 @@ def download_url(url, dest=None, overwrite=False):
         return providers.fetch(url)
 
 
+def local_path2url_path(local_path: str,
+                        auto_resolve: bool = False
+                        ) -> str:
+    """Convert a local path into a URL path component"""
```
yarikoptic, line 1158: Describe  auto_resolve. Also code uses only with False, so why bother having this option at all? Just document that path must be empty and done?!

datalad/support/network.py
```diff
@@ -1103,4 +1152,60 @@ def download_url(url, dest=None, overwrite=False):
...
+
+    return (
+        Path(url2pathname(url_path))
+        if return_path
+        else url2pathname(url_path)
+    )
+
+
+def quote_path(path: str, safe: str = "/") -> str:
+    """quote the path component of a URL, takes OS specifics into account"""
```
yarikoptic, line 1204: What is safe?

christian-monch: It is handed down to `urllib.parse.quote()` from the documentation of `quote`:

```
The optional safe parameter specifies additional ASCII characters that should not be quoted — its default value is '/'.
```

datalad/support/network.py
christian-monch: I changed `auto_resolve` to `allow_relative_path`. The path can be empty if `allow_relative_path` is `True`. That actually leads to some more simplifications.

datalad/support/network.py
christian-monch: Right. That is a leftover. Removed.

datalad/customremotes/ria_utils.py
christian-monch: I think that `if url_ri.path` triggers execution iff `url_ri.path` evaluates to `True`. And `url_ri.path or '/'` will evaluate to `url_ri.path` iff `bool(url_ri.path)` evaluates to `True`, else to `'/'`. So I think the two version have identical results. Am I missing something?

The reason for the change was that I was working in the method and thought, it would be a good refactoring, making the code shorter and more readable. The latter is a probably a matter of taste though.

datalad/support/network.py
yarikoptic: doesn't describe that to me here, describe that in the docstring above please, even if at least as an argument to be passed into `urllib.parse.quote`

datalad/support/network.py
yarikoptic: why bother having it if we don't use it in the code?

datalad/customremotes/ria_utils.py
yarikoptic: I am sorry -- I was misreading the code and was assuming that `,` is taking precedence and making `None, url_ri.path` into `(None, url_ri.path)` tuple.  I should have read original code better -- having them on different lines helped to accent that fact.

With that -- you should be correct with replacing that `if` with `or` here, sorry for the noise.

datalad/support/network.py
christian-monch: We use it.

datalad/customremotes/ria_utils.py
christian-monch: No worries 😉

christian-monch:
> I think it is ok now. The left over comments are nit picking

Thx for the feedback :)

will document sage and merge then

PR #7218, opened 2022-12-08
Fix setup.py for type-check action

bpoldrack:
Superfluous quotes made type-check action fail.

No changelog intended.


bpoldrack:
Trivial and does what it says (fixing the type-check run). Merging.

PR #7189, opened 2022-11-18
Reimplement get_submodules_() without get_content_info()  (Reincarnated 6942)

adswa:
A while ago, @mih proposed #6942. This effort was supposed to be super-seeded by #6974, which aimed to fix a short coming of #6942, namely an O(N²) performance hit in an edgecase ([when one provides paths to all subdatasets to a subdataset call](https://github.com/datalad/datalad/pull/6942/files#r947956976)). That fix is currently in draft mode, as it unexpectedly required more work than anticipated. @yarikoptic therefore stated in #6942 before it was closed

> @mih - I ran into need for more work on get_parent_paths since its goal is a bit different for matching paths within submodules not limiting some paths by other paths... needs a bit more work.
so I am ok with you solution, even if later to be improved later. But the tests I added in https://github.com/datalad/datalad/pull/6974 fails here and passes on master, so we would need to at least fix it up to not introduce regression. 

As I can still see the performance boost #6942 brought, I wanted to reincarnate the PR in its last state.
This is with ``master`` and on this branch with the same dataset @mih tested in #6942:

```
# on master
(gooey) adina@muninn in /tmp/ukb_cat on git:master
❱ time datalad subdatasets                                     
subdataset(ok): code/pipeline (dataset)
datalad subdatasets  177.08s user 16.30s system 99% cpu 3:13.39 total
# on this branch
(gooey) adina@muninn in /tmp/ukb_cat on git:master
❱ rehash
(gooey) adina@muninn in /tmp/ukb_cat on git:master
❱ datalad --version
datalad 0.17.9+158.g087ac2612
(gooey) adina@muninn in /tmp/ukb_cat on git:master
❱ time datalad subdatasets
subdataset(ok): code/pipeline (dataset)
datalad subdatasets  0.28s user 0.06s system 100% cpu 0.336 total
```

 I tested the tests that were failing before after merging ``master`` on my Debian system and on a Windows system, were they passed. Based on that, I'm slightly hopeful that our CI might turn green now as well :crossed_fingers: 
 
 I do see the performance hit in the edge case @yarikoptic discovered, though, still. Here are my timings in the human-connectome-project openaccess dataset:
 
 ```
 # on this branch
(gooey) adina@muninn in ~/repos/data/human-connectome-project-openaccess on git:master
[...]

PR #7220, opened 2022-12-08
RF: get_modules_ - avoid O(N^2) if paths provided via use of get_limited_paths

yarikoptic:
Redone #6974 and #7211 which attempted to reuse get_parent_paths with tune ups
-- was not sufficient since need to have a bit ad-hoc limiting by the path
WITHIN submodule to select that submodule (which would not work for plain directory).

Overall -- now that #7189 has finished the RFing to avoid some slow
operations at the cost of making some use cases slower, this should resolve it
back to avoid any slow operation.

#6974 commits avoided a hypothetical infinite recursion (my code analysis
says it was not possible, see #6941 (comment)) and performance issues by
introducing similar logic directly in the code of get_modules_.

get_limited_paths relies on sorting of the paths, should be O(N*log(N)) but
generally likely faster since likely to get them sorted to start with.

Timing this branch:

    ❯ datalad --version; (builtin cd /home/yoh/datalad/hcp-openaccess; /bin/time python -c "from datalad.support.gitrepo import *; from glob import glob; print('\n'.join(map(str,GitRepo('.').get_submodules_(glob('HCP1200/*')))))" | wc -l; )
    datalad 0.17.9+185.gd236b5a6d
    0.79user 0.14system 0:00.89elapsed 105%CPU (0avgtext+0avgdata 28896maxresident)k
    824inputs+0outputs (1major+8149minor)pagefaults 0swaps
    1113

and master

    ❯ datalad --version; (builtin cd /home/yoh/datalad/hcp-openaccess; /bin/time python -c "from datalad.support.gitrepo import *; from glob import glob; print('\n'.join(map(str,GitRepo('.').get_submodules_(glob('HCP1200/*')))))" | wc -l; )
    datalad 0.17.9+183.g34a28a622
    4.29user 0.11system 0:04.39elapsed 100%CPU (0avgtext+0avgdata 29092maxresident)k
    0inputs+0outputs (0major+8226minor)pagefaults 0swaps
    1113

and maint

    ❯ datalad --version; (builtin cd /home/yoh/datalad/hcp-openaccess; /bin/time python -c "from datalad.support.gitrepo import *; from glob import glob; print('\n'.join(map(str,GitRepo('.').get_submodules_(glob('HCP1200/*')))))" | wc -l; )
    datalad 0.17.9+82.g405ece550
    0.97user 0.12system 0:01.05elapsed 104%CPU (0avgtext+0avgdata 30248maxresident)k
    0inputs+0outputs (0major+8633minor)pagefaults 0swaps
    1113

so we are no worse (if not better) than maint and definetely gain over master
for such use cases of N submodules/N paths given. And for a single path we might even be not only faster but fixing some bug in maint:

    ❯ datalad --version; (builtin cd /home/yoh/datalad/hcp-openaccess; /bin/time python -c "from datalad.support.gitrepo import *; from glob import glob; print('\n'.join(map(str,GitRepo('.').get_submodules_(glob('HCP1200/*')[1]))))" | wc -l; )
    datalad 0.17.9+185.g8b21bfceb
    0.65user 0.11system 0:00.75elapsed 101%CPU (0avgtext+0avgdata 28040maxresident)k
    104inputs+0outputs (0major+7652minor)pagefaults 0swaps
    1

    ❯ git co master
    Switched to branch 'master'
    Your branch is up to date with 'origin/master'.

    ❯ datalad --version; (builtin cd /home/yoh/datalad/hcp-openaccess; /bin/time python -c "from datalad.support.gitrepo import *; from glob import glob; print('\n'.join(map(str,GitRepo('.').get_submodules_(glob('HCP1200/*')[1]))))" | wc -l; )
    datalad 0.17.9+183.g34a28a622
    0.75user 0.10system 0:00.83elapsed 102%CPU (0avgtext+0avgdata 28816maxresident)k
    0inputs+0outputs (0major+8203minor)pagefaults 0swaps
    1113

    ❯ git co maint
    Switched to branch 'maint'
    Your branch is up to date with 'origin/maint'.

    ❯ datalad --version; (builtin cd /home/yoh/datalad/hcp-openaccess; /bin/time python -c "from datalad.support.gitrepo import *; from glob import glob; print('\n'.join(map(str,GitRepo('.').get_submodules_(glob('HCP1200/*')[1]))))" | wc -l; )
    datalad 0.17.9+82.g405ece550
      File "/home/yoh/proj/datalad/datalad-master/datalad/support/gitrepo.py", line 2388, in get_submodules_
        for path, props in self.get_content_info(
      File "/home/yoh/proj/datalad/datalad-master/datalad/support/gitrepo.py", line 2769, in get_content_info
        posix_paths = get_parent_paths(posix_paths, submodules)
      File "/home/yoh/proj/datalad/datalad-master/datalad/support/path.py", line 191, in get_parent_paths
        _get_parent_paths_check(path, sep, asep)
      File "/home/yoh/proj/datalad/datalad-master/datalad/support/path.py", line 213, in _get_parent_paths_check
        raise ValueError("Expected relative within directory paths, got %r" % path)
    ValueError: Expected relative within directory paths, got '/'
    Command exited with non-zero status 1
    0.39user 0.04system 0:00.43elapsed 101%CPU (0avgtext+0avgdata 29832maxresident)k
    0inputs+0outputs (0major+8107minor)pagefaults 0swaps
    0


datalad/support/path.py
```diff
@@ -207,6 +208,58 @@ def get_parent_paths(paths, parents, only_with_parents=False, *, sep='/'):
     return res
 
 
+def get_limited_paths(paths: list[str|Path], limits: list[str|Path], *, include_within_path: bool = False) -> list[Path]:
```
yarikoptic, line 211: better name anyone? should I add trailing `_` to signal that it is a generator? (should adjust type annotation for that I guess)

datalad/support/tests/test_path.py
```diff
@@ -114,3 +115,13 @@ def _p(path):
...
+
+
+def test_get_limited_paths():
+    # just to avoid typing all the same
+    def glp(*args, **kwargs):
+        return list(get_limited_paths(*args, **kwargs))
+
+    assert glp(['a', 'b'], ['a', 'c']) == ['a']
+    assert glp(['a', 'b'], ['a/b', 'c']) == []  # a is not subpath of a/b
+    assert glp(['a', 'b'], ['a/b', 'c'], include_within_path=True) == ['a']  # a is not subpath of a/b
```
yarikoptic, line 127: Should add more tests I think

datalad/support/path.py
```diff
@@ -207,6 +208,58 @@ def get_parent_paths(paths, parents, only_with_parents=False, *, sep='/'):
...
+
+    # we will pretty much "scroll" through sorted paths and limits at the same time
+    for path_parts in paths_parts:
+        while limits_parts:
+            limit_parts = limits_parts[0]
+            l = min(len(path_parts), len(limit_parts))
+            # if common part is "greater" in the path -- we can go to the next "limit"
+            if limit_parts[:l] < path_parts[:l]:
+                # get to the next one
+                limits_parts = limits_parts[1:]
```
adswa, line 246:
```suggestion
                limits_parts = limits_parts[l:]
```
?

yarikoptic: Why?
l is applicable to a single path - length of common part among two parts - the original and the limiting one

changelog.d/pr-7220.md
```diff
@@ -0,0 +1,3 @@
+### 🏎 Performance
+
+- RF: get_modules_ - avoid O(N^2) if paths provided via use of get_limited_paths.  [PR #7220](https://github.com/datalad/datalad/pull/7220) (by [@yarikoptic](https://github.com/yarikoptic))
```
yarikoptic, line 3: doh -- we don't need changelog here since this is just an OPT on top of what we already have in master.
I better just adjust  http://github.com/datalad/datalad/blob/HEAD/changelog.d/pr-7189.md to reflect contribution/PRs and be done.  Will rewrite this commit

datalad/support/path.py
adswa: whoops then i misunderstood. thanks for the explanation :)

yarikoptic:
typing check fail seems to no longer have anything to do with type checking and rather with installation in the env
```
  File "/opt/hostedtoolcache/Python/3.7.15/x64/lib/python3.7/site-packages/packaging/_tokenizer.py", line 164, in raise_syntax_error
    span=span,
packaging._tokenizer.ParserSyntaxError: Expected closing RIGHT_PARENTHESIS
    sphinx-rtd-theme (>=0.5.1") ; extra == 'devel'
                     ~~~~~~~~^
```
since we have that typo fixed in 41d88cb9cad00c168021a76c18266f0b27f88c65 in maint... @bpoldrack beat me with #7222 for the PR to merge in maint -- thanks!

datalad/support/path.py
yarikoptic: may be
- `get_paths_subset`
- `subselect_paths`
- `get_selected_paths`
- `get_paths_under`
- `select_subpaths`
? (I like `get_` fashion though, although it is also a generator and value of being a generator is minimal here really)

yarikoptic:
green -- nice!  I will rebase now that master has maint thanks to @bpoldrack

datalad/support/path.py
```diff
@@ -207,6 +211,59 @@ def get_parent_paths(paths, parents, only_with_parents=False, *, sep='/'):
...
+def get_limited_paths(paths: list[str|Path], limits: list[str|Path], *, include_within_path: bool = False) \
+        -> Generator[str, None, None]:
+    """Given list of relative POSIX paths (or Path objects), select the ones within limits (also relative and POSIX).
+
+    In case of include_with_path=True, if limit points to some path under a 'path' within 'paths',
+    that path would be returned as well.
+
+    Yields
+    -------
+    list of paths, sorted (so order is not preserved), which reside under 'limits' paths or path within 'limits' is
```
yarikoptic, line 223: TODO: adjust 
```suggestion
    paths, sorted (so order is not preserved), which reside under 'limits' paths or path within 'limits' is
```

datalad/support/path.py
adswa: Not much creative genius from my part:
``get_overlapping_paths``
``get_paths_contained_in``
``get_paths_within_path``

datalad/support/gitrepo.py
```diff
@@ -2388,34 +2388,34 @@ def get_submodules_(self, paths=None):
...
+            # harmonize them into relative to the repository
+            posix_paths = []
+            for path in paths:
+                path = ut.PurePath(path)
+                if path.is_absolute():
+                    try:
+                        path = path.relative_to(self.pathobj)
+                    except ValueError as exc:
+                        lgr.debug(
+                            "Path %s it not relative to %s, skipping since nothing should match it: %s",
```
bpoldrack, line 2402:
```suggestion
                            "Path %s it not underneath %s, skipping since nothing should match it: %s",
```

I think the phrasing is otherwise misleading. The issue is not that path is not relative to something (we know it's absolute).

datalad/support/gitrepo.py
```diff
@@ -2388,34 +2388,34 @@ def get_submodules_(self, paths=None):
...
-        def is_relative_to(self, *other):
-            """Return True if the path is relative to another path or False.
-            """
-            try:
-                self.relative_to(*other)
-                return True
-            except ValueError:
-                return False
-
+        posix_mod_paths = [m.relative_to(self.pathobj).as_posix() for m in modinfo]
```
bpoldrack, line 2391: I think, this is a little weird. `get_submodules_` appears to be the only consumer of `_parse_gitmodules`. The latter transforms the recorded relative paths in `.gitmodules` into absolute ones, and the former transforms them back.

datalad/support/path.py
yarikoptic: I like to have consistent `get_` since we are to return adjusted result instead of filtering in place... then `get_filtered_paths_`, but how to call 2nd paths argument -- I would not like `filters` , so `filter_paths`?

datalad/support/gitrepo.py
yarikoptic: true, thanks!

datalad/support/path.py
bpoldrack: > then `get_filtered_paths_` [...] so `filter_paths`?

Yes. I like that better!

datalad/support/gitrepo.py
bpoldrack: > So, if nobody minds -- I should then just strip making them absolute within _parse_gitmodules.

No objection. :)

bpoldrack:
@yarikoptic: Not sure what's going in with this breaking all builds (not happening in other PRs against master as far as I can tell), but the kind of failure (`Value Error embedded null byte` from Popen) seems to suggest, that we somehow manage to smuggle the `\0` from here https://github.com/datalad/datalad/pull/7220/files#diff-bfc1c47d032fc72b15b6a79db3d49e4c4334cf96318c5efe48a3403744b32fb1L2417 into the paths returned and pass them on to another git call? Something like that?

yarikoptic:
yeap yeap, thanks @bpoldrack -- force pushed the fix to apply `[:-1]` to rpath before reusing it (I was reusing once with that stripping and once without)

yarikoptic:
@jwodder could you please look into how/why installation of git-annex fails on appveyor OSXes via datalad-installer -- e.g. https://ci.appveyor.com/project/mih/datalad/builds/45696242/job/h7a2890c5yyt5ekw -- if anything which could be done on our end?

yarikoptic:
@bpoldrack PR should be ready for re-review. appveyor gotchas are unrelated AFAIK

jwodder:
@yarikoptic The logs look like they might be truncated to me, as they end in the middle of a `brew install`.  Do you see anything past line 191?

Regardless, note that the output from Homebrew includes:

```
Warning: You are using macOS 10.15.
We (and Apple) do not provide support for this old version.
It is expected behaviour that some formulae will fail to build in this old version.
```

yarikoptic:
> Do you see anything past line 191?

I don't.  So, overall, it seems that it just needs newer OSX. May be appveyor script should just install from the datalad/git-annex built .dmg's and not bother with brew (filed https://github.com/datalad/datalad/pull/7233 attn @mih).

adswa:
> So, overall, it seems that it just needs newer OSX.

``maint`` already has them https://github.com/datalad/datalad/pull/7227

bpoldrack:
Ok, thank you @yarikoptic!

PR #7223, opened 2022-12-12
Bf fix installation link git annex

asmacdo:
I didn't have git annex installed. The error was helpful, but the link has changed:

```
datalad.support.exceptions.MissingExternalDependency: No working git-annex installation of version >= 8.20200309. Visit http://handbook.datalad.org/r.html?install for instructions on how to install DataLad and git-annex
```
### PR checklist

- [X] Provide an overview of the changes you're making and explain why you're proposing them.
- [ ] Create a changelog snippet (add the `CHANGELOG-missing` label to this pull request in order to have a snippet generated from its title; (**asmacdo) IMO no changelog necessary for this?**
- [X] If you would like to list yourself as a DataLad contributor and your name is not mentioned please modify .zenodo.json file.

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -126,6 +126,10 @@
             "affiliation": "Institute of Neuroscience and Medicine, Brain & Behaviour (INM-7), Research Centre Jülich, Jülich, Germany",
             "name": "Szczepanik, Michał",
             "orcid": "0000-0002-4028-2087"
+        },
+        {
+            "affiliation": "Dartmouth College, Hanover, NH, United States",
+            "name": "Macdonald, Austin"
         }
     ],
     "grants": [
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -476,7 +476,7 @@ class AnnexRepo(GitRepo, RepoInterface):
     def _check_git_annex_version(cls):
         ver = external_versions['cmd:annex']
         # in case it is missing
-        msg = "Visit http://handbook.datalad.org/r.html?install " \
+        msg = "Visit http://handbook.datalad.org/en/latest/intro/installation.html" \
               "for instructions on how to install DataLad and git-annex."
 
         exc_kwargs = dict(


bpoldrack:
Thank you, @asmacdo!

I agree - no changelog needed.

adswa:
Hey @bpoldrack and @asmacdo, thanks for your work on this, but the link hasn't changed. The [r.html page](handbook.datalad.org/en/latest/r.html) of the handbook contains references to pages based on their RST anchor, and is used to generate links on the fly that are more persistent than links based on the internal handbook repo file structure. If the installation page ever moves or gets renamed, a link based on directory structure goes stale. This new link is more fragile than the former, in my opinion.
I do see that something gets messed up during the redirection/resolving of the former link, likely during the two different redirection rules that are configured on readthedocs (the first redirection from "``$domain/path/file``" to "``$domain/$lang/$version/path/file``" might be at fault, I'm taking a look). I'll comment on this PR with what I find. I suspect, though, that this issue doesn't need a change here, and if it does I would prefer to not base the URL on directory paths.

adswa:
I managed to reproduce this issue - its indeed a redirection issue that occurs with specific URLs (those with a query string) and a specific form of automatic redirection (prefix redirection). I have opened an [issue on readthedocs](https://github.com/readthedocs/readthedocs.org/issues/9799) and will merge [a workaround in the JS script that does the redirection in a short while](https://github.com/datalad-handbook/book/pull/880). With that workaround, the former URL should work again. I'm quite sorry to request this @asmacdo as I'm very grateful for your effort in making and PRing this change, but I'd like to revert [598afe8](https://github.com/datalad/datalad/pull/7223/commits/598afe8bca6cc2955d9dd2dcd9760231cf5330c9). Any objections, @bpoldrack?

bpoldrack:
@adswa:

No objections. I wasn't aware of the idea behind `r`.

asmacdo:
TIL: r.html :)

PR #7227, opened 2022-12-14
Switch macOS appveyor builds to Monterey

bpoldrack:


bpoldrack:
AppVeyor failure unrelated - this is an instance of https://github.com/datalad/datalad/issues/7229

bpoldrack:
This should not be controversial. `brew` is complaining, that is doesn't support Catalina anymore and failures to build are expected - that's obviously nothing we want or need to deal with. As far as I remember, the only reason for running Catalina to begin with, is that it's AppVeyor's default.

So, I'm going to merge.

PR #7231, opened 2022-12-15
Move timeout processing out of exception handler processing

### LINKED ISSUE #7229: Various tests fail on occasion with timeout
Example:
https://ci.appveyor.com/project/mih/datalad/builds/45679300/job/x31hi28vfjn0et99#L2636

This failure is happening with different tests, IIRC, but not reliably. Rerunning the build usually does "solve" it.

@christian-monch: Can you have a look? I think you've addressed this somewhere, but I fail to find it.

christian-monch:
Fixes #7229 

This PR moves the timeout processing in ThreadedRunner out of the `Empty`-exception handler. This is done to prevent raising a `TimeoutError`-exception while handling an `Empty`-exception.

### DIFF
--- a/datalad/tests/test_cmd.py
+++ b/datalad/tests/test_cmd.py
@@ -67,21 +67,35 @@ def test_batched_close_abandon():
 
 @pytest.mark.filterwarnings("ignore:Exception ignored")
 def test_batched_close_timeout_exception():
-    # Expect a timeout if the process runs longer than timeout and the config
-    # for "datalad.runtime.stalled-external" is "abandon".
-    bc = BatchedCommand(
-        cmd=[sys.executable, "-i", "-u", "-q", "-"],
-        timeout=.5,
-        exception_on_timeout=True)
+    while True:
+        try:
+            # Expect a timeout at BatchedCommand.close() if the process runs
+            # longer than timeout and the config for
+            # "datalad.runtime.stalled-external" is "abandon".
+            # In most cases the next commands until `bc.close()` will execute
+            # faster than `timeout`. If not we just restart the process
+            bc = BatchedCommand(
+                cmd=[sys.executable, "-i", "-u", "-q", "-"],
+                timeout=.5,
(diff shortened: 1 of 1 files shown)


christian-monch:
appveyor ubuntu errors are due to problems installing `git-annex` via `deb`-tools

bpoldrack (commented):
@christian-monch: I may miss something here, but to me it's not clear that this addresses #7229.
It seems to me, that test is failing because there is an unexpected timeout in `response = bc("import time; print('a')")`. That the exception comes with a `__context__` (during handling of ...) doesn't look like a problem to me.

If that's wrong, can you elaborate?

christian-monch:
> @christian-monch: I may miss something here, but to me it's not clear that this addresses #7229. It seems to me, that test is failing because there is an unexpected timeout in `response = bc("import time; print('a')")`. That the exception comes with a `__context__` (during handling of ...) doesn't look like a problem to me.
> 
> If that's wrong, can you elaborate?

The timeout is expected. That is actually what we test for. The issue is that it is raised within a function that is ultimatively called by the Empty-handler in the old `nonasyncrunner.py:627`. That leas to the issue that an exception is raised while another exception is currently handled.

A similar thing could happen in the CommandError handler in the old "cmd.py:544" because when python is shutting down, the log call would trigger an exception. That is usually caught and hidden, but becomes visible within the CommandError-handler. This actually poses the question: "how many log calls do we have in exception handlers where we expect an interpreter shutdown?"

christian-monch:
> @christian-monch: I may miss something here, but to me it's not clear that this addresses #7229. It seems to me, that test is failing because there is an unexpected timeout in `response = bc("import time; print('a')")`. That the exception comes with a `__context__` (during handling of ...) doesn't look like a problem to me.
> 
> If that's wrong, can you elaborate?

You are right. The timeout is actually expected, I just assumed it would not show during the first call. It is a race condition though and depending on the scheduling, it might. So let me fix the patch

PR #5859, opened 2021-08-03
RF/NF: Overhaul ORA's config business

bpoldrack:
Clean up the entire config consideration code and introduce consistent local configs to overrule what is committed in the `git-annex` branch. This will then be used by auto-reconfiguration in `clone` to not commit reconfigs.

Not ready yet. Want to see tests (as insufficient as they are) in more environments than my laptop.

PR #7235, opened 2022-12-22
RF ORA

bpoldrack:
This is a revival of #5859 (which I thought was long merged).

Apart from some refactoring in ORA, it introduces the possibility to override the special remote configs (in `git-annex:remote.log`) by local configs. This is then used to change the reconfiguration of ORA remotes by `clone` to only apply locally instead of committing a changed RIA URL (which could lead to a constant ping pong between different access methods to a RIA store).

In addition, there is a shim special remote introduced here, that allows datasets using the old ria remote (https://github.com/datalad/git-annex-ria-remote) to still work. This doesn't come with a proper test, since there's little point in making the tests depend on a package that is deprecated and archived. Please note, that there are two more points about introducing this "second flavor" of the special remote:
1. One idea is to have this replaced by an implementation that would auto-migrate during `initremote`/`enableremote`.
2. Following up on #7022,  after 0.18 I want to move all the RIA/ORA stuff into https://github.com/datalad/datalad-ria. Those two special remotes here in core would then be replaced by a dummy implementation that simply issues an error message to annex, reading something like "please install datalad-ria extension".
So, the point here is to introduce that second "spot".

### DIFF
--- a/changelog.d/pr-7235.md
+++ b/changelog.d/pr-7235.md
@@ -1,0 +1,16 @@
+### 💥 Breaking Changes
+
+- Automatic reconfiguration of the ORA special remote when cloning from RIA 
+  stores now only applies locally rather than being committed.
+  [PR #7235](https://github.com/datalad/datalad/pull/7235)
+  (by [@bpoldrack](https://github.com/bpoldrack))
+
+### 🚀 Enhancements and New Features
+
+- ORA special remote now allows to override its configuration locally.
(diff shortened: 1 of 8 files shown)


bpoldrack:
Ok, no objections raised and the setup failure for the "Test old nose code" build clearly is unrelated.
Merging.

yarikoptic-gitmate:

Issue #6846, opened 2022-07-15
test_ria_basics.py::test_version_check has been failing against 8.20200309 for a while


yarikoptic:
spotted that recent cron job failed, looked at https://app.travis-ci.com/github/datalad/datalad/jobs/576842693 and confirmed historically:

```
$> datalad foreach-dataset "git grep 'FAILED ../datalad/distributed/tests/test_ria_basics.py::test_version_check'  | grep cron || :"
foreach-dataset(ok): /mnt/datasets/datalad/ci/logs/2022 (dataset)
03/cron/20220603T193647/173dc6f/travis-14100-failed/11.txt:FAILED ../datalad/distributed/tests/test_ria_basics.py::test_version_check - ...
foreach-dataset(ok): /mnt/datasets/datalad/ci/logs/2022/01 (dataset)
06/cron/20220506T193505/9c6b62c/travis-14007-failed/11.txt:FAILED ../datalad/distributed/tests/test_ria_basics.py::test_version_check - ...
01/cron/20220701T193907/797dde3/travis-14182-failed/11.txt:FAILED ../datalad/distributed/tests/test_ria_basics.py::test_version_check - ...
13/cron/20220513T195020/84f979a/travis-14027-failed/11.txt:FAILED ../datalad/distributed/tests/test_ria_basics.py::test_version_check - ...
10/cron/20220610T193717/029f839/travis-14136-failed/11.txt:FAILED ../datalad/distributed/tests/test_ria_basics.py::test_version_check - ...
20/cron/20220520T193556/84f979a/travis-14036-failed/11.txt:FAILED ../datalad/distributed/tests/test_ria_basics.py::test_version_check - ...
17/cron/20220617T193730/797dde3/travis-14152-failed/11.txt:FAILED ../datalad/distributed/tests/test_ria_basics.py::test_version_check - ...
27/cron/20220527T193600/310fd9f/travis-14055-failed/11.txt:FAILED ../datalad/distributed/tests/test_ria_basics.py::test_version_check - ...
24/cron/20220624T193828/797dde3/travis-14163-failed/11.txt:FAILED ../datalad/distributed/tests/test_ria_basics.py::test_version_check - ...
foreach-dataset(ok): /mnt/datasets/datalad/ci/logs/2022/05 (dataset)
foreach-dataset(ok): /mnt/datasets/datalad/ci/logs/2022/06 (dataset)
[...]

PR #7238, opened 2022-12-23
ENH(TST): add a test for assert_raises since concern was raised

yarikoptic:
May be CI version would work differently from local?

See https://github.com/datalad/datalad/issues/6846#issuecomment-1363878497 for the reference

attn @bpoldrack

### DIFF
--- a/datalad/tests/test_tests_utils_pytest.py
+++ b/datalad/tests/test_tests_utils_pytest.py
@@ -665,3 +665,27 @@ def test_ok_file_under_git_symlinks(path=None):
         ok_file_under_git(op.join(lpath, 'notingit'))
     with assert_raises(AssertionError):
         ok_file_under_git(op.join(lpath, 'nonexisting'))
+
+
+def test_assert_raises():
+    # rudimentary test of assert_raises shim prompted by suspicion in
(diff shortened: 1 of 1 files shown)


bpoldrack:
@yarikoptic Thanks, let's see what happens. But note, that the issue for some reason seemed to only appear in the build with older annex. While I don't believe it has anything to do with annex itself, may be enable that run like in https://github.com/datalad/datalad/pull/7236/commits/c54bcd91fe2d0322d694bfa70cce4c24815fa44e ?

Something has to trigger the problem after all.

adswa:
I've merged ``maint`` into this branch to update it. 
Appveyor says ``Error fetching YAML configuration file from the repository. Make sure AppVeyor is authorized to access your gitHub account: An error occurred while sending the request.``, though, which I couldn't resolve with a rebuild of that PR.

yarikoptic:
let me rebase and force push, may be appveyor has some odd permissions handling?

adswa:
I'd say we either go with this test, or close the PR as stale. I'll mark it as ready for review to facilitate discussions.

yarikoptic:
fails on the now outdated run of appveyor known and xfailed now fails -- not this test. So let's proceed

PR #7237, opened 2022-12-22
Merge maint

yarikoptic:

PR #7239, opened 2022-12-31
Release 0.18

bpoldrack:
Trying release process for minor release.  I suppose a PR against `maint` that comes from `master` with `maint` merged back and added changelog snippets for `master` PRs prior to introduction of `scriv` based release action should do.

- [x] merge `maint` into `master` (replacing #7237)
- [x] added changelogs, b/c `scriv` was introduced only halfway since 0.17


bpoldrack:
FTR: Two failures.

- something goes wrong with a conda-forge install of an old git-annex in the "old nose code" test build only. Don't think this should hinder a release
- semver-label explicitly tests that we don't label `minor` in a PR against `maint`. This is fine for regular PRs, but when I want to use the release action for a minor release, this test is irrelevant.


Hence, will merge and see whether this works out. Guess, worst case is another case of "wrong" version number.

PR #7232, opened 2022-12-15
Boost minimal git annex version

yarikoptic:
see individual commits, ref: https://github.com/datalad/datalad/issues/6847 (closed by a specific commit)

Issue #6847, opened 2022-07-15
boost minimal supported git-annex from 8.20200309


yarikoptic:
just now, while filing datalad/datalad-ria#49 I realized that we are still supporting > 2 yo git-annex. Well, we kinda support it since we do have some number of test skips etc AFAIK.  So may be for 0.18.0 we should boost minimal supported git-annex into some future version?

adswa:
Would make sense to me to update the minimal supported version. But why a future version? Even just taking the current release is a two-year jump from the current supported minimum. And if we manage to stick to ~6months between releases, we'll have the release at the end of the year, i.e., we'd have a~6 month time window to the minimal git-annex release.

yarikoptic:
by "future" I bet I meant future from 2 yo one ;-)  I would go for something like 8.20210223 present in debian stable bullseye. Ubuntus >= `21.10 ` have it. I will propose a PR.

PR #7247, opened 2023-01-09
NF Support annex.private for clone/get.

### LINKED ISSUE #6456: `clone/get` should support setting annex.private=true
This was already requested some time ago in https://github.com/datalad/datalad/issues/5835 in a slightly different context, but it has never been done. Being able to do it is a major improvement when updating large datasets where repeated temporary clones need to be made to fit in the storage allotment.

In the simplest case the `--reckless` switch gets a dedicated mode for this.

### LINKED ISSUE #7666: `datalad clone --reckless ephemeral` from anonymous https sets `annex-ignore` for origin
### What is the problem?

A `datalad clone --reckless ephemeral` using an anonymous https URL makes origin `annex-ignore`, even though a `git annex enableremote origin` immediately afterwards works just fine.

```
$ datalad clone --reckless ephemeral https://hub.datalad.org/distribits/recordings.git
[INFO   ] RIA store unavailable. -caused by- file:///data/group/psyinf/distribits/distribits.ria/ria-layout-version not found, self.ria_store_url: ria+file:///data/group/psyinf/distribits/distribits.ria, self.store_base_pass: /data/group/psyinf/distribits/distribits.ria, self.store_base_pass_push: None, path: <class 'pathlib.PosixPath'> /data/group/psyinf/distribits/distribits.ria/ria-layout-version -caused by- [Errno 2] No such file or directory: '/data/group/psyinf/distribits/distribits.ria/ria-layout-version' 
[WARNING] reckless=ephemeral mode: origin doesn't seem local: https://hub.datalad.org/distribits/recordings.git
| no symlinks being used 
install(ok): /home/icg149/Playground/recordings (dataset)
$ cd recordings/
$ datalad get distribits2024/2024-04-05_12_unconference.webm 
get(error): distribits2024/2024-04-05_12_unconference.webm (file) [not available; (Note that these git remotes have annex-ignore set: origin)]                                                                                                                      
[ble: exit 1]
$ git annex enableremote origin
enableremote (normal) origin ok
$ datalad get distribits2024/2024-04-05_12_unconference.webm 
get(ok): distribits2024/2024-04-05_12_unconference.webm (file) [from origin...]
```

### DataLad information

```
$ datalad --version
datalad 1.1.3
```

### Additional context

Slightly off-topic: what I am really looking for is just the annex.private part of an ephemeral clone, not necessarily the symlinking hacks, but this is rather convoluted compared to providing a single option when cloning:

```
datalad -c annex.private=true clone ...
cd ...
git config annex.private true
```

Maybe that should be made easier?

_No response_

mslw:
This adds a new reckless mode, called "private" to `clone` (and, implicitly, `get` and `install`). ~, and an analogous "private" option to create~. When used, it will set the `annex.private` config option to `true` before `git annex init` is called. As a consequence, information about the cloned / created dataset will not be recorded in the git-annex branch (the branch will still exist, and track information about other repositories). Closes #6456 and should also close #7666.

Private mode can be useful when working with temporary clones that are discarded after pushing (clone, make changes, push back to origin, discard clone). Compared to using `git annex dead here`, it does not clutter `uuid.log` and `trust.log` with information about dead clones; and compared with `reckless=ephemeral` it does not share the annex.

See [Cloning a repository privately](https://git-annex.branchable.com/tips/cloning_a_repository_privately/) and configuration section in git-annex's manpage.

Note 2024-08-10: this PR has been slimmed down considerably from the initial proposal, which also changed `create`, so not all of the discussion applies. Nevertheless, I chose to rebase and force push rather than open a new PR.


mslw:
Looking briefly at the test errors:

- "Test old nose code" fails on installing dependencies, more specifically conda install git annex
- On Travis, some runs fail on tests added by me. Runs that fail use git-annex 8.20201007, runs that pass use a more recent version. Older version does not support annex.private.
  
  Tests for ephemeral clone (which also sets annex.private) execute similar piece if `external_versions['cmd:annex'] >= "8.20210428"`. I don't think I should avoid the test for private though, as here effect of the config is the whole point. So probably I would have to also add annex version check to the clone and create commands. But maybe the required annex version will be increased for next minor release? For now, I won't modify create/clone further and wait for suggestions.

bpoldrack (commented):
Thanks very much, @mslw! Sorry for missing this one.

> Should the reckless mode be named "private" or maybe "temporary"?

I'm not sure either. Mostly b/c there's overlap with `ephemeral`. If `origin` isn't on the same fie system and therefore can't be linked, then both modes are identical (in intention - implementation should follow).
However, since `ephemeral` is used already, I think adding `temporary` is confusing. So, `private` and thus being a little more specific than other mode names, is the best I can see.

Re `untrust`:

> the only value is as an additional safeguard for issuing a command to drop from elsewhere when working in a private repository.

Good point, I think. But since we likely need to declare dead anyway to guard against older annex, `untrust` in addition seems superfluous.
On that note: DataLad's minimum version for annex doesn't matter for this. A repo can be "touched" by git-annex w/o any DataLad being involved.

> Should create also set datalad.clone.reckless config option so that clones made into the created private dataset respect that mode by default? Should create also check that config in potential parent datasets?

Yes, agree. Thanks!

Looking through the diff ...

mslw:
Thanks for looking into it, I am really curious about your opinion. This seems like a small change in principle, but there are some interactions and decisions along the way (see e.g. commit message for 24fd32d).

> Re `untrust`:
> 
> > the only value is as an additional safeguard for issuing a command to drop from elsewhere when working in a private repository.
> 
> Good point, I think. But since we likely need to declare dead anyway to guard against older annex, `untrust` in addition seems superfluous. On that note: DataLad's minimum version for annex doesn't matter for this. A repo can be "touched" by git-annex w/o any DataLad being involved.

See, but here's something I'm concerned about. I noticed a similar logic in the code of `clone_ephemeral` (comments, actually), and I'm not sure if I agree. When I declare `annex dead here` in a repository that is in private mode (which it would be in this case, as private needs to be declared very early on), the information goes into `.git/annex/journal-private/trust.log`, not the "ordinary" place (`trust.log` on git-annex branch). I did not check what, if anything, a git-annex version prior to introduction of private mode would do.

And if it wouldn't observe these settings, then there's little we can do about it other then make users aware of that. Both ephemeral and private setups exist for a good reason.

mslw:
I think this warrants a ping to #6847 and #7232 because this PR uses a relatively new annex feature (though approaching 2 years old now).

As this is an additional feature for `clone` and `create`, I think it's fine if the minimally supported annex version (and indeed, one that's currently in Debian stable) does not support the private feature. In this case, the `annex.private` config option would be set, but not acted upon.

This probably warrants a check / warning in the relevant commands though, that I didn't initially propose? Please advise.

bpoldrack:
> When I declare annex dead here in a repository that is in private mode (which it would be in this case, as private needs to be declared very early on), the information goes into .git/annex/journal-private/trust.log, not the "ordinary" place (trust.log on git-annex branch).

Good catch, I missed that! You're right, then, I suppose - we can't do a lot about it and `annex dead here` doesn't actually mitigate.
Somewhat tied to the question of raising annex' minimum version. Since it's just an additional feature I agree - not imperative to raise. I think, there are three somewhat reasonable options here:

  - Have all relevant commands check the installed annex version at the beginning if that option (or the config) is set and error out for an invalid parameter
  - Use an annex version kludge like we do elsewhere and set private only if supported, declare `git annex dead here` as fallback (as the next best option to achieve something that may be good enough for purpose) alongside a warning that the proper thing requires annex XXX instead
  - annex version kludge for setting private, but no fallback - error out instead.
  
Setting private while knowing that installed annex can't respect it, doesn't seem to make sense to me. Further operation on that dataset would lead to obviously undesired results (hence second option above as an alternative).

Note, that 1 isn't actually mutually exclusive with 3. It's probably 1+3 or 2. 3 could seem superfluous in that case, but I think having that protection at a lower level despite the intention to fail early is the safer approach WRT later code changes.
Got to admit I don't have a clear preference. I suppose 1+3 is cleanest. 2 could be convenient but also a bit misleading.

datalad/core/local/create.py
```diff
@@ -365,6 +377,10 @@ def __call__(
         if initopts is not None and isinstance(initopts, list):
             initopts = {'_from_cmdline_': initopts}
 
+        # Inherit annex.private=true if parent dataset is given explicitly
+        if ds is not None:
+            private = private or ds.config.get('annex.private', None) == 'true'
```
bpoldrack, line 382: Minor hint: You may want to use `ConfigManager`'s `getbool` here, instead of string comparison. That would allow for `True`, `true`, `1`, etc. - see `anything2bool`.

bpoldrack (commented):
So far, implementation looks good to me.
Not approving yet, b/c of unresolved question above.

mslw:
Thanks @bpoldrack for taking a closer look. Just a quick note that I didn't forget about this PR, but I focused on other things recently. I will make changes addressing your review (especially the thing when annex version doesn't support the private mode) at some point by the end of the week.

mslw:
Sorry, I ran out of steam. Will be unavailable in the coming week. @bpoldrack do you think you can take over?

bpoldrack:
@mslw 
> Sorry, I ran out of steam. Will be unavailable in the coming week. @bpoldrack do you think you can take over?

Yes, no worries.

yarikoptic:
@mslw do you think anything was left to be done for this PR?

adswa (approved):
to me this PR looks good, and the failing benchmarks are fixed outside of this PR already. I'd really like to have this PR merged! I have played around with the new functions in toy repos, and they work fine and as I expect judging from the git-annex docs. I haven't had real usecases for private mode, though - maybe a second pair of eyes can spot something I might have missed?

changelog.d/pr-7247.md
```diff
@@ -0,0 +1,8 @@
+### 🚀 Enhancements and New Features
+
+- Support for annex private mode.
+  A new `--reckless private` mode was added to the `clone`, `get` & `install` commands; a new `--private` option was added to `create`.
+  Using private mode will configure the new clone with `annex.private=true`, meaning that the clone won't store any information about itself in the git-annex branch (the branch will still exist and contain information about other clones).
+  This mode can be used for temporary clones, where changes are pushed (e.g. back to origin), and the temporary clone is promptly discarded.
```
yarikoptic, line 6: - Ideally we would like to harmonize how `private` is specified across commands.  But not yet sure if `--private` or `--reckless`:
- For `reckless` -- can't it be a combination of `shared-` and `private` or `ephemeral` and `private` or some other  in particular `wreckless` is we do want to quickly do some change in a clone to push back?

mslw: > Ideally we would like to harmonize how private is specified across commands. But not yet sure if --private or --reckless

I'm trying to remember what was the reason I chose the current way, but can't. Probably just following the suggestion in #6456 to reuse the existing switch for clone, and then not feeling comfortable with the calling the create operation "reckless".

> For reckless -- can't it be a combination of shared- and private or ephemeral and private or some other in particular wreckless is we do want to quickly do some change in a clone to push back?

Umm... For ephemeral & private: that wouldn't change anything because clone ephemeral already sets annex.private (which is the only DataLad usage of annex.private outside this PR). For shared & private, and yet-to-be-invented and private: I did not think of it, but I suppose it would technically be possible...

adswa: FWIW, I agree that it is useful to reuse the ``--reckless`` switch, especially when the combination ``--reckless ephemeral --private`` becomes possible but doesn't make sense. And I agree it makes sense to call it ``--private`` instead of reckless for ``create``

mih:
With datalad-next, this is how it would be done:

```
❯ datalad -c annex.private=true create some
create(ok): /tmp/some (dataset)
❯ ls --no-icons some/.git/annex
fsck   index.lck        journal.lck  keysdb.lck    sentinal        smudge.lck
index  journal-private  keysdb       othertmp.lck  sentinal.cache  smudge.log
```

(see the `journal-private` directory)

What it does not do is setting `annex.private` in the local gitconfig. I assume, but have not confirmed, that this is necessary for full compliance.

mih:
With https://hub.datalad.org in place I find myself using temporary clones of datasets a lot more. I am leaving the following observations to potentially inform progress here.

- I find `create --private` to be an edge case that does not immediately warrant an API extension. It would be sufficient to document the approach that enables this use case, i.e. `datalad -c annex.private=true create`
- I see the problems associated with implementing "private" as a reckless mode. Theoretically, combinations with `shared-` are desirable, and would require multiplexed modes. However, it would not be hard to implement these multiplex modes, but more importantly recent changes in key should make us re-consider the support for "shared access" entirely. Setting a repository to shared access is no longer sufficient for enabling shared access, because it requires additional setup on the side of the consuming user (see https://knowledge-base.psychoinformatics.de/kbi/0001).

My personal conclusion is to add `--reckless private` only, for now, and postponed any other action until a holistic concept for shared access is determined. It seems unproblematic to support the same `--reckless private` for both `create` and `clone` in a uniform fashion, even though the former is an edge case.

yarikoptic:
@mslw would you have time to proceed with this PR? (seems agreement boils down to make it `--reckless private` for `clone`)

mslw:
Thanks for the ping and communicating the consensus @yarikoptic ! If it is `clone --reckless private`, then I imagine not much changes will be needed. At this moment it looks like not this week, but I'll be happy to take it up next week.

As a side note regarding `datalad -c`: this currently does not have a direct equivalent in Python API (requires patching environ, e.g. with `datalad-next.utils.patched_env`), but my recent experiences also suggest that private mode is more useful more often for clone. So it all makes sense again. Thanks guys!

mslw:
@yarikoptic -- I rebased this PR aggresively, keeping only parts relevant for `clone` / `get`, and adding an explicit test for `get`. The interface is kept as `--reckless private`. I think it is done now!

yarikoptic:
@mslw, my bad to not come back to it later. rebased again to avoid conflicts (minor in doc strings... hopefully I didn't screw it up ;) ).

yarikoptic:
fails are true positive:
```
FAILED ../datalad/distribution/tests/test_get.py::test_private - AssertionError: path /tmp/datalad_temp_test_privatet8ld5r72/foo/.git/annex/journal-private/uuid.log does not exist (or dangling symlink)
assert False
 +  where False = exists()
 +    where exists = PosixPath('/tmp/datalad_temp_test_privatet8ld5r72/foo/.git/annex/journal-private/uuid.log').exists
 +      where PosixPath('/tmp/datalad_temp_test_privatet8ld5r72/foo/.git/annex/journal-private/uuid.log') = Path(PosixPath('/tmp/datalad_temp_test_privatet8ld5r72/foo/.git/annex/journal-private/uuid.log'))
= 1 failed, 1145 passed, 43 skipped, 2 xfailed, 1 xpassed, 10 warnings in 638.75s (0:10:38) =
```

mslw:
Thanks for spotting that!

It is related to annex versions. It succeeds for me locally but indeed fails in (some?) CI runs. Looking at one of the failing runs ([this one](https://github.com/datalad/datalad/actions/runs/12332452813/job/34420319031?pr=7247)), the environment had `cmd:annex: 8.20201007-g903b2f1`, so without the private feature.

This PR contains an annex version kludge, and it is used in `test_clone.py:test_ephemeral`, so probably the failing `test_get.py:test_private` should also use it. Currently I don't know whether it should be `assert_raises` or just skip.

Need to detach for now, will try to look closer today but can't promise.

yarikoptic:
I think we should move forward from very well aged

datalad/support/annexrepo.py:    GIT_ANNEX_MIN_VERSION = '8.20200309'

from CHANGELOG of annex

```
git-annex (8.20210428) upstream; urgency=medium
  
  * New annex.private and remote.name.annex-private configs that can
    prevent storing information about a repository and remotes in the
    git-annex branch.
```

so any version after would be good.  Debian stable has `10.20230126-3` so I would vote for that one as the minimal for the next "master".

meanwhile I would have just indeed added a check/skip .

yarikoptic:
yikes, where is all the other github CI runs? ;) but appveyor is happy at least.

changelog.d/pr-7247.md
yarikoptic: > when the combination --reckless ephemeral --private becomes possible but doesn't make sense.

why,  because every ephemeral is private?  

```shell
❯ rm -rf test-clone-; datalad clone --reckless=ephemeral test test-clone-; git -C test-clone- config annex.private
install(ok): /tmp/test-clone- (dataset)                                                                         
true
```
But that is **not** contradictory!  Moreover I might want 'shared-group` 

```shell
❯ rm -rf test-clone-; datalad clone --reckless=shared-group test test-clone-; git -C test-clone- config annex.private
install(ok): /tmp/test-clone- (dataset)     
```

and decide it to be `private` or not. And that might got for some other 'reckless' modes where I aim to avoid wasting space with `annex/objects` where I am to decide to either announce it private (or not).

Overall, IMHO there is nothing `reckless` in a private repo -- it is a different property/dimension even though for some reckless modes it might be 'default', as **we** decided, but others (e.g. Joey) did not for worktree handling which does a similar thing:

```
❯ rm -rf test-clone-; git -C test worktree add -B newbranch $PWD/test-clone-; git -C test-clone- config annex.private; ls -l test-clone-/.git/annex
Preparing worktree (resetting branch 'newbranch'; was at 4703f09)
HEAD is now at 4703f09 [DATALAD] Recorded changes
lrwxrwxrwx 1 yoh yoh 11 Apr  8 10:03 test-clone-/.git/annex -> ../../annex/
``` 

hence I guess it would be for the user to decide that it is desired (or not) to announce it private,

yarikoptic (changes requested):
on 2nd look, as I left in my elaborated comment, IMHO 'private' is not a reckless mode, even though some of our ad-hoc reckless modes might announce it as such.  Thus I would remain of an opinion that it is worth a separate, consistently (across commands) defined option, either bool `--private` or some other `--kind private`?!

yarikoptic:
moving back to draft and hope @mslw would get back to it to finalize

PR #7249, opened 2023-01-12
BF: Fix invalid version comparison

### LINKED ISSUE #7248: Invalid `git annex findref` call on Windows
```
CommandError: '"git" "-c" "diff.ignoreSubmodules=none" "annex" "findref" "--anything" "HEAD" "--json" "--json-error-messages" "-c" "annex.dotfiles=true"' failed with exitcode 1 under C:\DLTMP\datalad_temp_tree_em3s022d\ds [info keys: stdout_json] [err: 'Invalid option `--anything'
```

Seems to be due to a wrong version comparison result in datalad proper

bpoldrack:
Specifying 10.20221212+git18 as requirement doesn't work, when the reported version of an annex build is using a different scheme. It was used because the git-annex version required isn't released yet. See issue 7248.

However, we know the scheme of annex versions (it's a date after the dot), therefore setting the required version to one day after the previous release does the trick with respect to released annex versions. Hence, we will use this only with the coming release (not a current snapshot). But that is clearly more sensible than crashing on an annex version that doesn't support it yet.

Closes #7248

### DIFF
--- a/changelog.d/pr-7249.md
+++ b/changelog.d/pr-7249.md
@@ -1,0 +1,7 @@
+### 🐛 Bug Fixes
+
+- Fixes crashes on windows where DataLad was mistaking git-annex 10.20221212 for
+  a not yet released git-annex version and trying to use a new feature.
+  Fixes [#7248](https://github.com/datalad/datalad/issues/7248) via
+  [PR #7249](https://github.com/datalad/datalad/pull/7249)
+  (by [@bpoldrack](https://github.com/bpoldrack))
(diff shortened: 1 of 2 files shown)


bpoldrack:
I marked this as release, since this is a critical bug.

yarikoptic:
THANKS! I did feel that I will get us into trouble with that comparison but could not put finger to it... now I know and will try to remember.  That comparison by "date based" version is what I thought to use instead too, so there we agree.

yarikoptic:
travis stall -- filed and issue. let's proceed! Thanks @bpoldrack !

yarikoptic-gitmate:
PR released in [`0.18.1`](https://github.com/datalad/datalad/releases/tag/0.18.1)

PR #7252, opened 2023-01-12
Rf batched command

### LINKED ISSUE #7251: Ensure thread-safety in cleaning up inactive `BatchedCommand` instances
This is a follow-up to issue [#6179. ](https://github.com/datalad/datalad/issues/6197)

The current implementation of [#6197](https://github.com/datalad/datalad/issues/6197) contains a race condition that could lead to deleting the runner from an active `BatchedCommand`-instance.

Also, from within `BatchedCommand` instance- and class-code, we can only clean up elements internal to BatchedCommand, i.e. we can remove runner that are not actively used, In 0.18.0, there might be a problem with identifying all instances because elements are removed from the instance dictionary.

I would like to question the necessity of this specific garbage collection. If we want to keep it, I think it has to be modified to ensure that no runner is removed from an active instance. I will open a PR linked to this issue that uses locking to prevent the race condition mentioned above.

christian-monch:
Fixes #7251

This PR uses thread-locks to ensure thread-safe determination of whether a `BatchedCommand` instance is active or not.

TODO:
- [ ] add regression-tests

### DIFF
--- a/changelog.d/pr-7252.md
+++ b/changelog.d/pr-7252.md
@@ -1,0 +1,3 @@
+### 🏠 Internal
+
+- Rf batched command.  Fixes [#7251](https://github.com/datalad/datalad/issues/7251) via [PR #7252](https://github.com/datalad/datalad/pull/7252) (by [@christian-monch](https://github.com/christian-monch))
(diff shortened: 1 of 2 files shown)


christian-monch:
Hi @yarikoptic, I would appreciate your thoughts on this PR and the related issue #7251

PR #7262, opened 2023-01-16
PATCH: bundle looseversion instead of relying on proper packaged one

yarikoptic:
Copied from 1.0.3 of https://github.com/effigies/looseversion . Primary goal - prepare a tested patch to possibly be included in the patches to shortcut waiting for an approved package of looseversions (e.g. as in Debian)

This PR is to test that all works as expected whenever no looseversion is installed.


yarikoptic:
all good -- no side effects observed.

PR #7263, opened 2023-01-17
Revert "Revert "Remove chardet version upper limit""

yarikoptic:
This reverts commit 09cb41ce5b5d8800d00e2ead9c8a95cc2e4e767a. Original PR which removed upper bound:
https://github.com/datalad/datalad/pull/7029

Debians now have 5.1.0, and we can't package while demanding an outdated chardet. Let's see again if we really need to limit upper version.


yarikoptic:
all tests pass, debian build also passed. Let's merge. Third time is a charm, right?

yarikoptic-gitmate:

Issue #6149, opened 2021-11-07
`create-sibling` needs a substantial overhaul


mih:
This is a meta-issue linking a number of individual and interconnected issues. Most of them are close as stale, due to sustained neglect, but the underlying problems/defects remain and should be addressed in a holistic reimplementations of the command, rather then more carcinogenic patches to this dinosaur of a command.

- [ ]  #5857 
- [ ] #5836
- [ ] #5452
- [ ] #4969
- [ ] #3277
- [ ] #2614 
- [ ] #2230 
- [ ] #2028 
- [ ] #2027
- [ ] #1428 
- [ ] #1326

Issue #4969, opened 2020-09-28
locale-dependent create-sibling failure


kyleam:
`create_sibling._ls_remote_path` assumes English output.

```console
$ export LANG=bg_BG.UTF-8
$ python -m nose datalad.distribution.tests.test_create_sibling:test_replace_and_relative_sshpath
SE
======================================================================
ERROR: datalad.distribution.tests.test_create_sibling.test_replace_and_relative_sshpath(False,)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/kyle/src/python/venvs/datalad/lib/python3.7/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/kyle/src/python/datalad/datalad/tests/utils.py", line 731, in _wrap_with_tempfile
    return t(*(arg + (filename,)), **kw)
  File "/home/kyle/src/python/datalad/datalad/tests/utils.py", line 731, in _wrap_with_tempfile
    return t(*(arg + (filename,)), **kw)
  File "/home/kyle/src/python/datalad/datalad/distribution/tests/test_create_sibling.py", line 537, in check_replace_and_relative_sshpath
    ds.create_sibling(url, ui=True)
  File "/home/kyle/src/python/datalad/datalad/distribution/dataset.py", line 503, in apply_func
    return f(**kwargs)
  File "/home/kyle/src/python/datalad/datalad/interface/utils.py", line 481, in eval_func
    return return_func(generator_func)(*args, **kwargs)
  File "/home/kyle/src/python/datalad/datalad/interface/utils.py", line 469, in return_func
    results = list(results)
  File "/home/kyle/src/python/datalad/datalad/interface/utils.py", line 400, in generator_func
    allkwargs):
  File "/home/kyle/src/python/datalad/datalad/interface/utils.py", line 539, in _process_results
    for res in results:
  File "/home/kyle/src/python/datalad/datalad/distribution/create_sibling.py", line 761, in __call__
    inherit
  File "/home/kyle/src/python/datalad/datalad/distribution/create_sibling.py", line 196, in _create_dataset_sibling
    path_children = _ls_remote_path(shell, remoteds_path)
  File "/home/kyle/src/python/datalad/datalad/distribution/create_sibling.py", line 392, in _ls_remote_path
    out, err = ssh("ls -A1 {}".format(sh_quote(path)))
  File "/home/kyle/src/python/datalad/datalad/cmd.py", line 608, in __call__
[...]

PR #7265, opened 2023-01-18
Fix `create-sibling` for non-English SSH remotes

### LINKED ISSUE #7264: create-sibling fails with non-english language setting on the server
`create-sibling` fails with non-existing target directories, when the language setting on the server isn't english. That's because we catch the respective failure of `ls` by matching its output to `No such file or directory`. If we need to rely on such output to figure out what went wrong, we should explicitly run `LC_ALL=C ls ..` instead.

Thanks, @nobodyinperson, for reporting.

nobodyinperson:
Came up in [chat](https://app.element.io/#/room/#datalad:matrix.org):

`datalad create-sibling` failed when trying to check for the existance of the sibling because of a non-fixed locale. Example:

```bash
yann in machine in ~/code/datalad 
❯ datalad create ds
[WARNING] Requested extension 'next' is not available 
create(ok): /home/yann/code/datalad/ds (dataset)
yann in machine in ~/code/datalad took 4s220ms 
❮ datalad create-sibling -d ds machine.server.de:code/datalad/ds-sibling
[WARNING] Requested extension 'next' is not available 
[INFO   ] No sibling name given. Using URL hostname 'machine.server.de' as sibling name 
[INFO   ] Connecting ... 
[INFO   ] Considering to create a target dataset /home/yann/code/datalad/ds at code/datalad/ds-sibling of machine.server.de 
CommandError: 'ssh -o ControlPath=/home/yann/.cache/datalad/sockets/cb4db4c9 machine.server.de 'ls -A1 code/datalad/ds-sibling'' failed with exitcode 2
ls: Zugriff auf 'code/datalad/ds-sibling' nicht möglich: Datei oder Verzeichnis nicht gefunden 
# ☝️☝️☝️☝️ shouldn't be German as the error message is compared to the English equivalent in the code
```

This PR fixes this by setting the locale of the used `ls` command to the default `C` locale. 

Fixes #7264

### DIFF
--- a/changelog.d/pr-7265.md
+++ b/changelog.d/pr-7265.md
@@ -1,3 +1,3 @@
 ### 🐛 Bug Fixes
 
-- Fix `create-sibling` for non-English SSH remotes.  [PR #7265](https://github.com/datalad/datalad/pull/7265) (by [@nobodyinperson](https://github.com/nobodyinperson))
+- Fix `create-sibling` for non-English SSH remotes by providing `LC_ALL=C` for the `ls` call.  [PR #7265](https://github.com/datalad/datalad/pull/7265) (by [@nobodyinperson](https://github.com/nobodyinperson))
--- a/changelog.d/pr-7265.md
+++ b/changelog.d/pr-7265.md
@@ -1,0 +1,3 @@
+### 🐛 Bug Fixes
+
+- Fix `create-sibling` for non-English SSH remotes.  [PR #7265](https://github.com/datalad/datalad/pull/7265) (by [@nobodyinperson](https://github.com/nobodyinperson))
--- a/datalad/distribution/create_sibling.py
+++ b/datalad/distribution/create_sibling.py
@@ -392,7 +392,7 @@ def _create_dataset_sibling(
 def _ls_remote_path(ssh, path):
     try:
         # yoh tried ls on mac
-        out, err = ssh("ls -A1 {}".format(sh_quote(path)))
+        out, err = ssh("LC_ALL=C ls -A1 {}".format(sh_quote(path)))
         if err:
             # we might even want to raise an exception, but since it was
             # not raised, let's just log a warning
--- a/datalad/distribution/create_sibling.py
+++ b/datalad/distribution/create_sibling.py
@@ -392,7 +392,9 @@ def _create_dataset_sibling(
 def _ls_remote_path(ssh, path):
     try:
         # yoh tried ls on mac
-        out, err = ssh("LC_ALL=C ls -A1 {}".format(sh_quote(path)))
+        out, err = ssh('sh -c "LC_ALL=C; export LC_ALL; /bin/ls -A1 {}"'.format(
+            sh_quote(path)))
+
         if err:
             # we might even want to raise an exception, but since it was
             # not raised, let's just log a warning
--- a/datalad/distribution/create_sibling.py
+++ b/datalad/distribution/create_sibling.py
@@ -15,6 +15,7 @@ import logging
 import os
 from distutils.version import LooseVersion
 from os.path import curdir
+import shlex
 from os.path import join as opj
 from os.path import (
     normpath,
@@ -392,8 +393,13 @@ def _create_dataset_sibling(
 def _ls_remote_path(ssh, path):
     try:
         # yoh tried ls on mac
-        out, err = ssh('sh -c "LC_ALL=C; export LC_ALL; /bin/ls -A1 {}"'.format(
-            sh_quote(path)))
+        # escape path explicitly with shlex.quote as 'sh' is a POSIX shell and
+        # sh_quote could decide to quote Windows-style
+        ls_cmd = "LC_ALL=C; export LC_ALL; /bin/ls -A1 {}".format(shlex.quote(path))
+        # TODO: Using sh_quote here is also flawed as it checks whether the
+        # *local* machine is Windows. Doesn't help if the remote we're ssh'ing in is Windows.
+        ssh_cmd = "sh -c {}".format(sh_quote(ls_cmd))
+        out, err = ssh(ssh_cmd)
 
         if err:
             # we might even want to raise an exception, but since it was
--- a/datalad/distribution/tests/test_create_sibling.py
+++ b/datalad/distribution/tests/test_create_sibling.py
@@ -946,7 +946,7 @@ def test_preserve_attrs(src=None, dest=None):
         assert fp.read() == "This is test text."
 
 
-@skip_if_on_windows
+@known_failure_windows # backstory: https://github.com/datalad/datalad/pull/7265
 @with_tempfile(mkdir=True)
 def test_only_one_level_without_recursion(path=None):
     # this tests for https://github.com/datalad/datalad/issues/5614: accidental
--- a/datalad/distribution/tests/test_create_sibling.py
+++ b/datalad/distribution/tests/test_create_sibling.py
@@ -946,6 +946,7 @@ def test_preserve_attrs(src=None, dest=None):
         assert fp.read() == "This is test text."
 
 
+@skip_if_on_windows
 @with_tempfile(mkdir=True)
 def test_only_one_level_without_recursion(path=None):
     # this tests for https://github.com/datalad/datalad/issues/5614: accidental
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -130,6 +130,11 @@
         {
             "affiliation": "Institute of Energy and Climate Research - Stratosphere (IEK-7), Research Centre Jülich, Jülich, Germany",
             "name": "Riße, Matthias"
+        },
+        {
+            "affiliation": "Eberhard Karls Universität Tübingen",
+            "name": "Büchau, Yann Georg",
+            "orcid": "0000-0003-2472-2051"
         }
     ],
     "grants": [


changelog.d/pr-7265.md
```diff
@@ -0,0 +1,3 @@
+### 🐛 Bug Fixes
+
+- Fix `create-sibling` for non-English SSH remotes.  [PR #7265](https://github.com/datalad/datalad/pull/7265) (by [@nobodyinperson](https://github.com/nobodyinperson))
```
yarikoptic, line 3: my 1c on a bit more specific description:

```suggestion
- Fix `create-sibling` for non-English SSH remotes by providing `LC_ALL=C` for the `ls` call.  [PR #7265](https://github.com/datalad/datalad/pull/7265) (by [@nobodyinperson](https://github.com/nobodyinperson))
```

adswa:
There's a relevant test failure on Windows Powershell:

```
____________________ test_only_one_level_without_recursion ____________________
[gw0] win32 -- Python 3.9.13 C:\Python39-x64\python.exe
path = WindowsPath('C:/DLTMP/datalad_temp_57dlsyej')
    @with_tempfile(mkdir=True)
    def test_only_one_level_without_recursion(path=None):
        # this tests for https://github.com/datalad/datalad/issues/5614: accidental
        # recursion of one level by default
        path = Path(path)
        ds_main = Dataset(path / "main").create()
        ds_main.create('sub1')
    
>       ds_main.create_sibling(
            name="dummy",
            sshurl=str(path / "toplevelsibling"))
..\datalad\distribution\tests\test_create_sibling.py:957: 
..\datalad\distribution\dataset.py:502: in apply_func
    return f(*args, **kwargs)
..\datalad\interface\base.py:773: in eval_func
    return return_func(*args, **kwargs)
..\datalad\interface\base.py:763: in return_func
    results = list(results)
..\datalad\interface\base.py:873: in _execute_command_
    for r in _process_results(
..\datalad\interface\utils.py:319: in _process_results
    for res in results:
..\datalad\distribution\create_sibling.py:794: in __call__
    path = _create_dataset_sibling(
..\datalad\distribution\create_sibling.py:200: in _create_dataset_sibling
    path_children = _ls_remote_path(shell, remoteds_path)
..\datalad\distribution\create_sibling.py:395: in _ls_remote_path
    out, err = ssh("LC_ALL=C ls -A1 {}".format(sh_quote(path)))
..\datalad\distribution\create_sibling.py:103: in __call__
    out = self.run(cmd, protocol=StdOutErrCapture)
self = <datalad.distribution.create_sibling._RunnerAdapter object at 0x000001A9CCFA0840>
cmd = 'LC_ALL=C ls -A1 "C:\\DLTMP\\datalad_temp_57dlsyej\\toplevelsibling"'
protocol = <class 'datalad.runner.coreprotocols.StdOutErrCapture'>, stdin = None
cwd = None, env = None, timeout = None, exception_on_error = True, kwargs = {}
applied_cwd = None, applied_env = None
threaded_runner = <datalad.runner.nonasyncrunner.ThreadedRunner object at 0x000001A9CD2EC940>
results_or_iterator = {'code': 1, 'stderr': "'LC_ALL' is not recognized as an internal or external command,\r\noperable program or batch file.\r\n", 'stdout': ''}
results = {'code': 1, 'stderr': "'LC_ALL' is not recognized as an internal or external command,\r\noperable program or batch file.\r\n", 'stdout': ''}
```

I believe both the way to set environment variables is invalid in powershell (its something like ``$env:LC_ALL=C``), and I *think* the environment variable itself is even meaningless for Windows. 
I guess its more fail-safe to pass this environment variable through the `env=` parameter of the runner, but I didn't try yet. I can try this evening, unless someone has a better idea about it.

nobodyinperson:
I had a look around but got lost where exactly the SSH process is invoked. Maybe [here](https://github.com/datalad/datalad/blob/maint/datalad/support/sshconnector.py#L577)? Anway, if the local `ssh` process has the `LC_ALL` variable set that's probably not enough as something like `-o 'SendEnv LC_ALL'` might be needed to actually forward the variable to the remote program.

adswa:
> I had a look around but got lost where exactly the SSH process is invoked. Maybe [here](https://github.com/datalad/datalad/blob/maint/datalad/support/sshconnector.py#L577)? Anway, if the local `ssh` process has the `LC_ALL` variable set that's probably not enough as something like `-o 'SendEnv LC_ALL'` might be needed to actually forward the variable to the remote program.

Indeed, absolutely true. Thinking pragmatically about this, given that almost everything in ``create-sibling`` is [incompatible with Windows anyhow](#6149), and almost all of the ``create-sibling`` tests are skipped, I'd actually propose simply skipping this test on windows as well.

yarikoptic:
NB wrote long response and then it was gone (don't get surprised if gets double posted in different wording).

Since this fix makes test `test_only_one_level_without_recursion` to no longer pass on Windows I would consider this change to introduce a regression bug, and thus not a good fit for `maint` (if at all). Given that every "meter" of the Windows land is hard to concur over, we better not surrender it so easily!

If we are to keep this approach for testing for remote path, we then might want to either first figure out what shell we are talking to (PS or not) and then act accordingly to that shell.   Or in the worst case to not fix the bug for Windows non-English-speaking users only (i.e. not add LC_ALL to invocation if `datalad` itself runs on windows).

But alternatively may be there is some common to PS and POSIX shell other command/way to test presence of the remote path which would be consistent and not relying on exit code (since we might not tell error of ssh from the error of e.g. remote `test` command)?

adswa:
> Since this fix makes test test_only_one_level_without_recursion to no longer pass on Windows I would consider this change to introduce a regression bug, and thus not a good fit for maint (if at all). Given that every "meter" of the Windows land is hard to concur over, we better not surrender it so easily!

I completely hear you, but I tend to disagree nevertheless on some of it. What the test checks for is an accidental recursion into the first level of subdatasets - this isn't reintroduced here. I dislike disabling tests a lot, yet an accidental recursion of ``create-sibling`` should be caught on other systems as well. I agree on the importance of getting DataLad to work on Windows, but I do feel that its a somewhat lost cause for this current implementation of ``create-sibling``. Out of the 13 tests for it, 11 were skipped on Windows already. Those that are ran are the one I disabled in my commit to this PR and a test for an invalid command call. I think its more than fair to call out the bad practice of simply disabling the test, but in this specific command and situation I would personally and pragmatically say its better to prioritize  fixing the command for non-English Unix systems as the command barely functions at all on Windows.

> But alternatively may be there is some common to PS and POSIX shell other command/way to test presence of the remote path which would be consistent and not relying on exit code (since we might not tell error of ssh from the error of e.g. remote test command)?

Ha, you're comment got me thinking - not as elegant as what you'd wish for, but what if we wrap the entire command including the LC_ALL env in a ``bash -c`` call? I think this should get it to work on all Windows shells.

nobodyinperson:
I don't really understand what's the matter with Powershell here: Is the problem that...

1) we SSH into a Windows host (does that even work?) and on that remote host the default shell is Powershell which doesn't understand setting environment variables like `LC_ALL=C command ...`? 
2) Or is the problem that we execute `create-sibiling` on a Windows machine and SSH into a Linux host that for some reason then uses Powershell as default shell (why would anybody use Powershell as default shell on a Linux host?). 
3) Or does SSH somehow *transfer* the current local shell (Powershell apparently when executed from Windows) into the remote (how does that work? Or is just `$SHELL` forwarded and if the remote *happens* to have Powershell available, that is used? Can we in that case overwrite/not forward `$SHELL`?)

> Ha, you're comment got me thinking - not as elegant as what you'd wish for, but what if we wrap the entire command including the LC_ALL env in a `bash -c` call? I think this should get it to work on all Windows shells.

In any case, if it is a given that `bash` is available on *any* remote we SSH into (maybe plain `sh` is better?), wrapping the command in `bash -c` sounds like a good solution to me. Would also avoid the problem with people using fancy shells like `fish` that always source their configs and that might alias `ls` to `exa` which doesn't have the `-A` option.

adswa:
> I don't really understand what's the matter with Powershell here: Is the problem that...
> 
>     1. we SSH into a Windows host (does that even work?) and on that remote host the default shell is Powershell which doesn't understand setting environment variables like `LC_ALL=C command ...`?
> 
>     2. Or is the problem that we execute `create-sibiling` on a Windows machine and SSH into a Linux host that for some reason then uses Powershell as default shell (why would anybody use Powershell as default shell on a Linux host?).
> 
>     3. Or does SSH somehow _transfer_ the current local shell (Powershell apparently when executed from Windows) into the remote (how does that work? Or is just `$SHELL` forwarded and if the remote _happens_ to have Powershell available, that is used? Can we in that case overwrite/not forward `$SHELL`?)
> 
> 
> > Ha, you're comment got me thinking - not as elegant as what you'd wish for, but what if we wrap the entire command including the LC_ALL env in a `bash -c` call? I think this should get it to work on all Windows shells.

I believe in this specific case its 1), but a general problem of ``create-sibling`` is that it hard-codes quite a few non-cross-platform compatible commands, so whether its a Windows server or a local Windows machine, a likely result is a "XY is not recognized as an internal or external command" CommandError. ``LC_ALL=C`` adds one more incompatibility, but its on top of many others that existed for years and make the command unusable in practice on Windows. 
At the same time, ``create-sibling`` is an old-ish command that has suffered from a fair bit of neglect and wasn't part of the synchronized effort for more usability on Windows (this has so far focused primarily on the ``datalad/core/`` commands). Actually, in browsing through a range of stale issues about ``create-sibling``, I even realized that the bug this PR fixes has been known for quite a while (https://github.com/datalad/datalad/issues/4969) and is part of a larger TODO list that no-one yet had time tackling. 

And to be honest, I've been staring at this test for an hour now, and even though I wrote it and must made a conscious decision to have it run under Windows, its not clear to me how that test passed before your change. It must be something we do on Appveyor specifically; this test as it is in ``maint`` currently does not pass on the real Win 10 and Win 11 machines I have access to (``ls`` works on barely any Windows shell by default). 

> In any case, if it is a given that `bash` is available on _any_ remote we SSH into (maybe plain `sh` is better?), wrapping the command in `bash -c` sounds like a good solution to me. Would also avoid the problem with people using fancy shells like `fish` that always source their configs and that might alias `ls` to `exa` which doesn't have the `-A` option.

I'd stick with ``bash``. Neither ``sh`` nor ``bash`` are native to Windows, but Git for Windows comes with it (thx to @mih for that hint).

adswa:
does anyone else have thoughts on this? (@bpoldrack maybe?)

adswa:
Thanks for the additional fixup! I forgot to click 'comment' on a quick note I wrote: The verdict in the devcall was to disregard Windows in this fix, keep the test disabled, and opt for ``sh`` :+1:

adswa:
The failures on Travis and ``datalad-next`` are unrelated

datalad/distribution/create_sibling.py
```diff
@@ -392,7 +393,14 @@ def _create_dataset_sibling(
 def _ls_remote_path(ssh, path):
     try:
         # yoh tried ls on mac
-        out, err = ssh("ls -A1 {}".format(sh_quote(path)))
+        # escape path explicitly with shlex.quote as 'sh' is a POSIX shell and
+        # sh_quote could decide to quote Windows-style
+        ls_cmd = "LC_ALL=C; export LC_ALL; /bin/ls -A1 {}".format(shlex.quote(path))
```
yarikoptic, line 398:
```suggestion
        ls_cmd = "LC_ALL=C /bin/ls -A1 {}".format(shlex.quote(path))
```
isn't sufficient?

datalad/distribution/tests/test_create_sibling.py
```diff
@@ -946,6 +946,7 @@ def test_preserve_attrs(src=None, dest=None):
         assert fp.read() == "This is test text."
 
 
+@skip_if_on_windows
```
yarikoptic, line 949: I think we better keep this one (with some annotation) for when we never expect that test to not be ran on windows, and use `@known_failure_windows` which would skip them by default but otherwise so at some point we could easily trigger sweep over some issues which might no longer be failures etc.

datalad/distribution/create_sibling.py
```diff
@@ -392,7 +393,14 @@ def _create_dataset_sibling(
...
+        # TODO: Using sh_quote here is also flawed as it checks whether the
+        # *local* machine is Windows. Doesn't help if the remote we're ssh'ing in is Windows.
```
yarikoptic, line 400: so should we then just hardcode `""` wrapping since IIRC single quote wrapping (`'???'`) is not available on Windows?

datalad/distribution/create_sibling.py
nobodyinperson: No, some shellx don't support the `export VAR=value` directly AFAIK. I'd leave it like this.

datalad/distribution/create_sibling.py
nobodyinperson: No, hard-coding doesn't work because the filename could contain quotes that need to be dealt with. Using `shlex.quote` or `sh_quote` handle this.

I don't have access to a Windows SSH host to test this, maybe someone else can check if this works for normal non-exotic files with quotes in the name.

adswa: > I don't have access to a Windows SSH host to test this, maybe someone else can check if this works for normal non-exotic files with quotes in the name.

``"`` is considered an illegal character on Windows systems, one couldn't create a file with it in its name: https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file

datalad/distribution/tests/test_create_sibling.py
adswa: Right, I forgot I wanted to change this. Pushed a change.

datalad/distribution/create_sibling.py
adswa: It followed a suggestion by @mih from the dev call, maybe he can chime in?

mih: @nobodyinperson already mentioned that the safe approach (as in most widely compatible with particular shells) is the one currently implemented. I have nothing else to add.

datalad/distribution/create_sibling.py
mih: ... and single quotes are not a thing in `cmd.exe`

adswa (approved):
This PR can be merged from my point of view

bpoldrack:
Agree on merging. Thanks much, @nobodyinperson!

nobodyinperson:
Meh, just realized that umlauts in the target dir now don't work as the `C` locale only knows ASCII 🙄

nobodyinperson:
Best part is it fails silently. But to know whether the target even supports a Unicode locale or how it is called, we'd need to run `locale -a` and do some grepping. Sounds cumbersome.

nobodyinperson:
But the good news is: `ls` seems to silently fall back to the predictable `C` locale if `LC_ALL` is set to an unknown locale. So it should be safe to set `LC_ALL=C.UTF-8`.

yarikoptic-gitmate:

PR #7266, opened 2023-01-18
BF: Set language when parsing output

### LINKED ISSUE #7264: create-sibling fails with non-english language setting on the server

bpoldrack:
In `create-siblling` we rely on parsing the output of `ls`. This failed when the language on the target system isn't english.

Closes #7264

### DIFF
--- a/datalad/distribution/create_sibling.py
+++ b/datalad/distribution/create_sibling.py
@@ -392,7 +392,7 @@ def _create_dataset_sibling(
 def _ls_remote_path(ssh, path):
     try:
         # yoh tried ls on mac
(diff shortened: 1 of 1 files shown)


bpoldrack:
Closing again - #7265 already addresses it ;-)

PR #7271, opened 2023-01-28
Codespell more (CHANGELOGs etc) and remove custom CLI options from tox.ini

yarikoptic:
I was codespell'ing a few other projects and got scared that we do not codespell our own DataLad because there were no `.codespellrc`.  apparently we did codespell but with custom CLI options in tox.ini and only the main package.  This PR makes codespelling more thorough, and in particular with us using scriv for changelogs - makes more sense now since we can then spellcheck those entries too (I checked using strace that codespell does open files under `changelog.d/`).

### DIFF
--- a/.codespell-ignorewords
+++ b/.codespell-ignorewords
@@ -1,4 +1,0 @@
-commitish
-ba
-froms
-ro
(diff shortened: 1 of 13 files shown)


yarikoptic:
wow, noone got excited about reviewing typo fixes? ;) I skimmed again, seems good. Adding changelog and will merge tomorrow if no objections voiced.

yarikoptic:
it was green before adding changelog and fails are "too code related"  , just filed https://github.com/datalad/datalad/issues/7296 , will proceed as promised

yarikoptic-gitmate:

PR #7272, opened 2023-01-28
Compact badges

TheChymera:
@yarikoptic some additional fixes:

* the original badges had a lot of redundant text (like repeating the word package in each one), figured out how to remove it.
* previously the github version badge was repeated twice
* we were missing Fedora, which interestingly also ships up-to-date datalad 🥳
* the list was not sorted alphabetically


bpoldrack:
Nice. Thank you, @TheChymera!

PR #7273, opened 2023-01-30
create-sibling: Use C.UTF-8 locale instead of C

nobodyinperson:
- supports special characters in filenames (like umlauts, etc.)
- `ls` seems to fall back to predictable `C` locale if `LC_ALL` is set to an unknown locale, so this should be safe.
- still: `create-sibling` currently fails silently if an umlaut is in the `--target-dir` and the remote `ls` doesn't handle it properly.

follow-up on PR #7265

### DIFF
--- a/changelog.d/pr-7273.md
+++ b/changelog.d/pr-7273.md
@@ -1,0 +1,3 @@
+### 🐛 Bug Fixes
+
+- create-sibling: Use C.UTF-8 locale instead of C.  [PR #7273](https://github.com/datalad/datalad/pull/7273) (by [@nobodyinperson](https://github.com/nobodyinperson))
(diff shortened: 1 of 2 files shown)


bpoldrack:
Thanks, @nobodyinperson! I think this is fine.

changelog.d/pr-7273.md
```diff
@@ -0,0 +1,3 @@
+### 🐛 Bug Fixes
+
+- create-sibling: Use C.UTF-8 locale instead of C.  [PR #7273](https://github.com/datalad/datalad/pull/7273) (by [@nobodyinperson](https://github.com/nobodyinperson))
```
bpoldrack, line 3:
```suggestion
- create-sibling: Use C.UTF-8 locale instead of C on the remote end.  [PR #7273](https://github.com/datalad/datalad/pull/7273) (by [@nobodyinperson](https://github.com/nobodyinperson))
```

yarikoptic:
I wonder if this would not kick-back on systems/environments without UTF-8 support available somehow (may be some really minimalistic environment/filesystem) but I guess we will see. Thanks @nobodyinperson !

yarikoptic-gitmate:

PR #7289, opened 2023-02-17
Api docs patch

### LINKED ISSUE #7288: A few commands, including `create_sibling_ria` and `foreach_dataset`, are missing from python module reference
I noticed that these commands are not listed in the [Python module reference](https://docs.datalad.org/en/stable/modref.html) of our online docs:

- `create_sibling_ria`
- `export_archive_ora`
- `foreach_dataset`
- `shell_completion`
- `uninstall`

`uninstall` has been removed in 85b9edf5c due to deprecation, and `shell_completion` has questionable value in Python API usage, but I feel that the first three (or even four) should be included - and aren't probably due to simple omission from the listing. The first three are listed under the methods of `datalad.api.Dataset`, but that table has no links to the method description.

I'll make a PR with the small change.

mslw:
This PR adds `create_sibling_ria`, `foreach_dataset`, `export_archive_ora`, and `shell_completion` to the table of contents in the Python module reference, therefore including them in the html documentation.

This required a change to section headings in the `create_sibling_ria` docstring, because sphinx napoleon extension only allows a set of predefined headings (custom ones can be added, but I thought reformatting makes more sense). I also used the opportunity to change single quotes into double backticks in said docstring, but compared to headings this isn't necessary for the docbuild.

Details in commit messages. Closes #7288

### DIFF
--- a/datalad/distributed/create_sibling_ria.py
+++ b/datalad/distributed/create_sibling_ria.py
@@ -88,25 +88,26 @@ class CreateSiblingRia(Interface):
     the store or specific datasets inside of it.
 
     The general structure of a RIA URL pointing to a store takes the form
-    'ria+[scheme]://<storelocation>' (e.g.,
-    ria+ssh://[user@]hostname:/absolute/path/to/ria-store, or
-    ria+file:///absolute/path/to/ria-store)
+    ``ria+[scheme]://<storelocation>`` (e.g.,
+    ``ria+ssh://[user@]hostname:/absolute/path/to/ria-store``, or
+    ``ria+file:///absolute/path/to/ria-store``)
 
     The general structure of a RIA URL pointing to a dataset in a store (for
     example for cloning) takes a similar form, but appends either the datasets
-    UUID or a ~ symbol followed by the dataset's alias name:
-    'ria+[scheme]://<storelocation>#<dataset-UUID>' or
-    'ria+[scheme]://<storelocation>#~<aliasname>'.
+    UUID or a "~" symbol followed by the dataset's alias name:
+    ``ria+[scheme]://<storelocation>#<dataset-UUID>`` or
+    ``ria+[scheme]://<storelocation>#~<aliasname>``.
     In addition, specific version identifiers can be appended to the URL with an
-    additional @ symbol:
-    'ria+[scheme]://<storelocation>#<dataset-UUID>@<dataset-version>', where 'dataset-version' refers to a branch or tag.
+    additional "@" symbol:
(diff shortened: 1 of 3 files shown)


datalad/distributed/create_sibling_ria.py
```diff
@@ -78,32 +78,36 @@ class CreateSiblingRia(Interface):
     The store's base path is expected to not exist, be an empty directory,
     or a valid RIA store.
 
+    Notes
+    -----
+
+
     RIA URL format
```
yarikoptic, line 85: so there can be no subsection whatsoever?

could/should there be explicit formatting, e.g.
```suggestion
    **RIA URL format**
```
?

docs/source/modref.rst
```diff
@@ -84,8 +86,10 @@ Miscellaneous commands
    api.check_dates
    api.configuration
    api.export_archive
+   api.export_archive_ora
    api.export_to_figshare
    api.no_annex
+   api.shell_completion
```
yarikoptic, line 92: I know that would be much to ask but since it seems to be "common" to forget to include them in modref, a unittest could be added to https://github.com/datalad/datalad/blob/HEAD/datalad/tests/test_api.py#L81 or around to alert when interface is not present in modref.rst if it would find it present?

not sure though if wouldn't interfer with the cases of extensions.  so consider as a really weak wishlist

datalad/distributed/create_sibling_ria.py
mslw: I tried `^^^` and it still complained about "Unexpected section title" :( Napoleon [docs](https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html) don't talk about subsections.

Nothing wrong about bolding the headers - I simply replicated what @mih used in his longer docstrings in -next. Both ways look fine to me.

docs/source/modref.rst
mslw: Thanks for pointing out that test, I wasn't aware of it. So the added test would need to parse `modref.rst`?

With extensions though - if they are listed in interface groups, then I suppose the outcome would depend on what is installed in test environment (which doesn't have to be CI). Although taking a random virtualenv and quickly checking, that might not be the case.

I guess that could be done, question is whether it should be done ;)

mslw:
I don't see myself working on a test that compares the ReST document to the actual API in the coming days due to other backlogs, and since @yarikoptic  sees it as "a really weak wishlist", I'd like to propose this PR as is.

adswa (approved):
This looks good to me

yarikoptic-gitmate:

PR #7293, opened 2023-02-21
TEMP: Just to get tests going - want to test clean maint

yarikoptic:


yarikoptic:
https://github.com/datalad/datalad/issues/7296

PR #7298, opened 2023-02-23
Reduce test environment complexity by redirecting HOME

### LINKED ISSUE #7297: What are the conditions under which the CI runs (there is no test HOME)
I am still chasing https://github.com/datalad/datalad-next/issues/249 and continue to discover misconceptions of mine on how I thought we would be running the tests.

I thought the pytest setup prepares a fake HOME, and configures the environment such that all test code runs relatively isolated from a user's actual configuration.

This is wrong (at least since Git 2.32). A new directory is prepared, but it is not declared HOME. The only thing that is happening is GIT_CONFIG_GLOBAL is pointed to a new config file inside that dir.

Everything else is left untouched, it seems. This means that any other configuration is actually shared with the executing user's environment.

Importantly, the configuration established by `conftest.py` is not the same as that created by, e.g. `.appveyor.yml` (I think what is done there is entirely useless). Also the fake HOME that is created has nothing to do with the `__testhome__` that is created and cd'ed into by the various CI setups.

IMHO this complexity creates a needlessly complex situation that seemingly serves no purpose.

Here are some observations re the effective environment of a test run done by me personal user on my personal laptop:

```
(Pdb) os.environ['HOME']
'/home/mih'
(Pdb) os.environ['GIT_CONFIG_GLOBAL']
'/tmp/datalad_temp_qg2z6bei/.gitconfig'
(Pdb) Path.cwd()
PosixPath('/home/mih/hacking/datalad/next')
(Pdb) platformdirs.user_config_dir()
'/home/mih/.config'
```

So the only thing "isolated" is the global git config. everything else is shared.
With Git < 2.32 the result would be different, because HOME would be redefined:

```
(Pdb) os.environ['HOME'] = '/tmp/datalad_temp_qg2z6bei'
(Pdb) platformdirs.user_config_dir()
'/tmp/datalad_temp_qg2z6bei/.config'
(Pdb) platformdirs.user_cache_dir()
'/tmp/datalad_temp_qg2z6bei/.cache'
```

It seems this particular setup was established in https://github.com/datalad/datalad/commit/c0fab80374bc3e128453dd1bd5086d17c8e1186f by @bpoldrack. From the description I get the impression that an alternative solution could have been to set GIT_CONFIG_GLOBAL in addition to a change of HOME. This could have solved the issue on Mac without creating this complex situation across all test environments.

mih:
Since Git 2.32 we are still creating a dedicated temporary HOME directory for running tests, but no longer actually enable it to be used beyond the global git config (i.e., `.gitconfig`).

This change rectifies that and actually reconfigures the environment to use this temporary HOME.

This should more effectively avoid "carry-over effects" between test runs, and make debugging easier (because of more similar conditions with the CI systems that are destroyed entirely after each run).

Without this change, tests could still trigger changes in a real user's `$HOME/.config` or `$HOME/.cache` that are persistent across test runs.

A simple test is included to trigger breakage in case someone is making modifications to this setup.

Closes #7297

### DIFF
--- a/datalad/conftest.py
+++ b/datalad/conftest.py
@@ -116,6 +116,7 @@ def setup_package():
             cfg_file.write_text(gitconfig)
             return new_home, cfg_file
 
+        # Only Git 2.32 brought GIT_CONFIG_GLOBAL support
         if external_versions['cmd:git'] < "2.32":
             # To overcome pybuild overriding HOME but us possibly wanting our
             # own HOME where we pre-setup git for testing (name, email)
@@ -126,8 +127,17 @@ def setup_package():
                 new_home, _ = prep_tmphome()
                 m.enter_context(patch.dict(os.environ, get_home_envvars(new_home)))
         else:
-            _, cfg_file = prep_tmphome()
-            m.enter_context(patch.dict(os.environ, {'GIT_CONFIG_GLOBAL': str(cfg_file)}))
+            new_home, cfg_file = prep_tmphome()
+            if 'GIT_HOME' in os.environ:
+                home_env = {'HOME': os.environ['GIT_HOME']}
+            else:
+                home_env = get_home_envvars(new_home)
+            # Some software does not expect a non-standard HOME
+            # point GIT_CONFIG_GLOBAL to the new git user config
+            # specifically to mitigate that
+            # see c0fab80374bc3e128453dd1bd5086d17c8e1186f
+            home_env['GIT_CONFIG_GLOBAL'] = str(cfg_file)
+            m.enter_context(patch.dict(os.environ, home_env))
 
         # Re-load ConfigManager, since otherwise it won't consider global config
         # from new $HOME (see gh-4153
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -281,6 +281,11 @@ Additionally, [tools/testing/test_README_in_docker](tools/testing/test_README_in
 be used to establish a clean docker environment (based on any NtesteuroDebian-supported
 release of Debian or Ubuntu) with all dependencies listed in README.md pre-installed.
 
+The pytest setup creates a new temporary HOME directory for each test run in order to
+minimize interaction with a user's particular configuration. Via the `GIT_HOME`
+environment variable, the test setup can be instructed to use a different particular
+(possibly persistent) HOME directory.
+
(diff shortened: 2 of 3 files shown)


datalad/conftest.py
```diff
@@ -126,8 +127,17 @@ def prep_tmphome():
...
+            if 'GIT_HOME' in os.environ:
+                home_env = {'HOME': os.environ['GIT_HOME']}
+            else:
+                home_env = get_home_envvars(new_home)
+            # Some software does not expect a non-standard HOME
+            # point GIT_CONFIG_GLOBAL to the new git user config
+            # specifically to mitigate that
+            # see c0fab80374bc3e128453dd1bd5086d17c8e1186f
+            home_env['GIT_CONFIG_GLOBAL'] = str(cfg_file)
+            m.enter_context(patch.dict(os.environ, home_env))
```
yarikoptic, line 140: seems becomes at large duplication with code above...
can't do suggestion over unchanged codeblock above, so - 
Overall -- if to go this way, can't we just `if` on git for `GIT_CONFIG_GLOBAL` only? i.e. 

```python
            new_home, cfg_file = prep_tmphome()
            if 'GIT_HOME' in os.environ:
                home_env = {'HOME': os.environ['GIT_HOME']}
            else:
                home_env = get_home_envvars(new_home)
            # Some software does not expect a non-standard HOME
            # point GIT_CONFIG_GLOBAL to the new git user config
            # specifically to mitigate that
            # see c0fab80374bc3e128453dd1bd5086d17c8e1186f
            if external_versions['cmd:git'] >= "2.32":
                 home_env['GIT_CONFIG_GLOBAL'] = str(cfg_file)
            m.enter_context(patch.dict(os.environ, home_env))
```

but I am afraid that this would bring us back what was attempted to be fixed by #6260 et al

adswa: > but I am afraid that this would bring us back what was attempted to be fixed by https://github.com/datalad/datalad/pull/6260 et al

yes indeed, the change you propose above would cause ``USERPROFILE`` to be overwritten on any Windows machine - it would probably cause quite a few tests to fail, and add a major obfuscation to debugging.

mih:
I am closing this PR of mine. Rational https://github.com/datalad/datalad/issues/7297#issuecomment-1446041345

Issue #7296, opened 2023-02-22
A swarm of tests (RIA + smth about ssh version) started to fail


yarikoptic:
just a few times back, initially attributed to https://github.com/datalad/datalad/pull/7292 but then from fresh https://app.travis-ci.com/github/datalad/datalad/jobs/596482198#L7524 of #7293  which was done to check if we are still green -- saw that it is just some regression, possibly due to some new ssh version or whatnot:

```
FAILED ../datalad/distributed/tests/test_create_sibling_ria.py::test_create_simple - datalad.runner.exception.CommandError: CommandError: 'git -c diff.ignoreSubmodules=none annex copy --batch -z --to datastore-storage --fast --json --json-error-messages --json-progress -c annex.dotfiles=true' failed with exitcode 1 under /tmp/datalad_temp_tree__test_create_storeketqmloj [info keys: stdout_json]
> to datastore-storage...
  Unable to remove /tmp/datalad_temp__test_create_storeqz7w04c1/6fc/eb6aa-90e5-4a98-b3e4-34852dfb20ec/ora-remote-fb7a30b0-c9d0-4ab2-a4d1-76422baa1f6a/transfer/MD5E-s4--03d59e663c1af9ac33a9949d1193505a.txt or to obtain write permission in parent directory. -caused by- rm /tmp/datalad_temp__test_create_storeqz7w04c1/6fc/eb6aa-90e5-4a98-b3e4-34852dfb20ec/ora-remote-fb7a30b0-c9d0-4ab2-a4d1-76422baa1f6a/transfer/MD5E-s4--03d59e663c1af9ac33a9949d1193505a.txt failed: 
  This could have failed because --fast is enabled. [err: 'git-annex: copy: 1 failed']
FAILED ../datalad/core/distributed/tests/test_clone.py::test_no_ria_postclonecfg - datalad.runner.exception.CommandError: CommandError: 'git -c diff.ignoreSubmodules=none annex copy --batch -z --to teststore-storage --fast --json --json-error-messages --json-progress -c annex.dotfiles=true' failed with exitcode 1 under /tmp/datalad_temp_tree_test_no_ria_postclonecfgrd6yqigl [info keys: stdout_json]
> to teststore-storage...
[...]

PR #7303, opened 2023-02-24
TEMP : print exactly what we see there on CI to crash

yarikoptic:
troubleshooting one of the fails in #7296


yarikoptic:
eh, running that test only with "the patch" it is all clear

```
../datalad/support/tests/test_sshconnector.py::test_ssh_copy PASSED      [100%]
=============================== warnings summary ===============================
../datalad/distributed/ora_remote.py:1098
  /home/travis/build/datalad/datalad/datalad/distributed/ora_remote.py:1098: DeprecationWarning: invalid escape sequence \.
    "^remote\..*\.annex-uuid",
-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
============================== slowest durations ===============================
```

yarikoptic:
and entire files -- failed only one due to time out... odd.

yarikoptic:
ok, I will just consider the issue got scared and ran away - we can no longer reproduce it. the only failing test is test_create_simple for ria which runs out of time (so there is that too, eg).

Issue #7286, opened 2023-02-15
most recent release of git-annex breaks 9 datalad tests


yarikoptic:
### What is the problem?

2 last days daily builds of git-annex errored out with 9 tests failing, most recent has

- cmd:annex: 10.20230214+git4-g474ea248e-1~ndall+1
- cmd:bundled-git: 2.30.2
- cmd:git: 2.30.2

```
2023-02-15T04:33:39.0617009Z =========================== short test summary info ============================
2023-02-15T04:33:39.0618135Z FAILED ../datalad/core/distributed/tests/test_clone.py::test_gin_cloning - datalad.runner.exception.CommandError: CommandError: 'git -c diff.ignoreSubmodules=none -c annex.merge-annex-branches=false annex info --bytes --fast --json --json-error-messages -c annex.dotfiles=true -- ./git/one' failed with exitcode 1 under /tmp/datalad_temp_test_gin_cloning2bxo5ekw [info keys: stdout_json] [err: 'fatal: Not a valid object name ./git/one
2023-02-15T04:33:39.0618428Z git-annex: not a directory or an annexed file or a treeish or a remote or a uuid']
2023-02-15T04:33:39.0619808Z FAILED ../datalad/local/tests/test_add_archive_content.py::test_add_archive_content - datalad.runner.exception.CommandError: CommandError: 'git -c diff.ignoreSubmodules=none -c annex.alwayscommit=false -c annex.merge-annex-branches=false annex info --bytes --fast --json --json-error-messages -c annex.dotfiles=true -- './.git/datalad/tmp/archives/9e66946386/1/1 f.txt'' failed with exitcode 1 under /tmp/datalad_temp_test_add_archive_contentju_6g1u4 [info keys: stdout_json] [err: 'fatal: Not a valid object name ./.git/datalad/tmp/archives/9e66946386/1/1 f.txt
2023-02-15T04:33:39.0620099Z git-annex: not a directory or an annexed file or a treeish or a remote or a uuid']
[...]

PR #7305, opened 2023-02-25
BF: declare string r-aw for the \. escape

yarikoptic:
Minor "fix".  I don't think it is even worth a changelog entry.

But also sending to to reconfirm that we observe fails of #7296

### DIFF
--- a/datalad/distributed/ora_remote.py
+++ b/datalad/distributed/ora_remote.py
@@ -1095,7 +1095,7 @@ class ORARemote(SpecialRemote):
             # version.
             # Lets try to find ourselves: Find remote with matching annex uuid
             response = _get_gitcfg(self.gitdir,
-                                   "^remote\..*\.annex-uuid",
+                                   r"^remote\..*\.annex-uuid",
                                    regex=True)
             response = response.splitlines() if response else []
             candidates = set()


yarikoptic:
damn thing passes !
- travis fail - reported https://github.com/datalad/datalad/issues/7306 - spurious fail, not new (don't see a button to restart for some reason!).
- appveyor - macs - got new cmd:annex: 10.20230214 so subject to #7286. 
- next -
```
FAILED annexremotes/tests/test_uncurl.py::test_uncurl_checkurl - ValueError: unsupported URL 'garbledhttps://httpbin.org/status/404'
FAILED annexremotes/tests/test_uncurl.py::test_uncurl_addurl_unredirected - datalad.runner.exception.CommandError: CommandError: 'git -c diff.ignoreSubmodules=none annex addurl --file=dummy garbledhttps://httpbin.org/bytes/24 -c annex.dotfiles=true' failed with exitcode 1 under /tmp/pytest-of-runner/pytest-0/test_uncurl_addurl_unredirecte0/ds [out: 'addurl garbledhttps://httpbin.org/bytes/24 
failed'] [err: 'Configuration of annex.security.allowed-url-schemes does not allow accessing garbledhttps://httpbin.org/bytes/24
addurl: 1 failed']
```
and was failing for awhile but nobody cared to look, neither the case here.  Let's proceed with this trivial fix.

PR #7308, opened 2023-02-27
 BF: Don't skip subds underneath paths in reckless & recursive drop

### LINKED ISSUE #7013: drop --reckless kill   ignores subdatasets under folders pointed to by `drop` command
It might be something obvious but I am missing it: do we have a way to invoke `uninstall` or `drop` to uninstall/drop all subdatasets but leave actual dataset behind?  I have tried

```
(dandisets) dandi@drogon:/mnt/backup/dandi/dandisets/000108$ datalad drop --what datasets --reckless kill -r -J10 
[ERROR  ] refuse to perform actions that would remove the current working directory (RuntimeError) 
(dandisets) dandi@drogon:/mnt/backup/dandi/dandisets/000108$ datalad drop --what datasets --reckless kill -r -J10 .
[ERROR  ] refuse to perform actions that would remove the current working directory (RuntimeError) 
```

so "current one" gets in the way, limiting to subfolders doesn't work

```
(dandisets) dandi@drogon:/mnt/backup/dandi/dandisets/000108$ datalad drop --what datasets --reckless kill -r -J10 sub*
```
in that it does nothing really. So the only way I see is to explicitly glob all subdatasets I know about:

```
(dandisets) dandi@drogon:/mnt/backup/dandi/dandisets/000108$ datalad drop --what datasets --reckless kill -r -J10 ./sub-*/ses-*/*/*.ome.zarr
uninstall(ok): sub-MITU01h3/ses-20220313h17m18s08/micr/sub-MITU01h3_ses-20220313h17m18s08_sample-25_stain-SST_run-1_chunk-4_SPIM.ome.zarr (dataset)
uninstall(ok): sub-MITU01h3/ses-20220313h17m18s08/micr/sub-MITU01h3_ses-20220313h17m18s08_sample-25_stain-SST_run-1_chunk-3_SPIM.ome.zarr (dataset)
...
```

and then it "goes" but BTW **very slowly** with 100% CPU by datalad (0.17.2, something to figure out I guess, might relate to #6974 but unlikely since it is how slow it goes after each subdataset, not some initial analysis),

adswa:
This fixes #7013.
If a drop --recursive --what all --reckless kill was provided with paths
to directories, it silently skipped killing the contained subdatasets.
The problem arises when the internal resolving of paths to datasets
determines that the provided directory path belongs to the superdataset.
Instead of then recursing into the subdirectory's potential subdatasets,
drop restrained the scope of what was to be dropped to only
'filecontent' in order to limit the drop from the determined
superdataset to the provided path as an internal safety mechanism:

    if paths is not None and paths != [ds.pathobj] and what == 'all':
        # so we have paths constraints that prevent dropping the full dataset
        lgr.debug('Only dropping file content for given paths in %s, '
                  'allthough instruction was to drop %s', ds, what)
        what = 'filecontent'
    
This change lifts this safety mechanism partially: Whenever reckless is
set to 'kill', and the provided path does not match the currently
investigated dataset, it adds a subdataset call constrained to the path
in question. If this returns a dataset, we know that this dataset is
meant to be dropped as well, as recursive (which has to be set whenever
reckless==kill) must be set, too.

### DIFF
--- a/datalad/distributed/drop.py
+++ b/datalad/distributed/drop.py
@@ -346,6 +346,30 @@ def _drop_dataset(ds, paths, what, reckless, recursive, recursion_limit, jobs):
         return
 
     if paths is not None and paths != [ds.pathobj] and what == 'all':
+        if recursive and reckless == 'kill':
+            # check if any paths contains a subdataset, and if so, drop it to
+            # ensure its not left behind
+            for sub in ds.subdatasets(
+                    # just check for subds at the provided path
+                    path=paths,
+                    state='present',
+                    recursive=recursive,
+                    recursion_limit=recursion_limit,
+                    result_xfm='datasets',
+                    on_failure='ignore',
+                    return_type='generator',
+                    result_renderer='disabled'):
+                if sub is not None:
+                    # there is a subdataset underneath the given path
+                    yield from _drop_dataset(
+                        ds=sub,
+                        # everything underneath the subds can go
+                        paths=None,
+                        what=what,
+                        reckless=reckless,
+                        recursive=False,
+                        recursion_limit=None,
+                        jobs=jobs)
         # so we have paths constraints that prevent dropping the full dataset
         lgr.debug('Only dropping file content for given paths in %s, '
                   'allthough instruction was to drop %s', ds, what)
--- a/datalad/distributed/tests/test_drop.py
+++ b/datalad/distributed/tests/test_drop.py
@@ -462,7 +462,8 @@ def test_kill(path=None):
 
 @with_tempfile
 def test_kill_7013(path=None):
-    """check that a recursive kill does not silently skip subds"""
+    """check that a recursive kill does not silently skip subdatasets
+     contained in subdirectory: github.com/datalad/datalad/issues/7013"""
     ds = Dataset(path).create()
     (ds.pathobj / 'subdir').mkdir()
     ds.create('subdir/subds')
--- a/datalad/distributed/tests/test_drop.py
+++ b/datalad/distributed/tests/test_drop.py
@@ -460,6 +460,20 @@ def test_kill(path=None):
     eq_(False, ds.pathobj.exists())
 
 
+@with_tempfile
+def test_kill_7013(path=None):
+    """check that a recursive kill does not silently skip subds"""
+    ds = Dataset(path).create()
+    (ds.pathobj / 'subdir').mkdir()
+    ds.create('subdir/subds')
+    ds.create('subds')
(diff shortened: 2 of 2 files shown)


adswa:
The failure is unrelated: 
```
FAILED ../datalad/support/tests/test_sshconnector.py::test_ssh_copy - datalad.runner.exception.CommandError: CommandError: 'scp -o ControlPath=/home/travis/.cache/datalad/sockets/1d11e952 /tmp/datalad_temp_test_ssh_copyxs_2d2y5 /tmp/datalad_temp_test_ssh_copy8zhyfqs1 datalad-test:/tmp/datalad_temp_test_ssh_copywrh7glsq' failed with exitcode 1 [err: 'lost connection']
```

datalad/distributed/drop.py
```diff
@@ -346,6 +346,30 @@ def _drop_dataset(ds, paths, what, reckless, recursive, recursion_limit, jobs):
...
+                    result_renderer='disabled'):
+                if sub is not None:
+                    # there is a subdataset underneath the given path
+                    yield from _drop_dataset(
+                        ds=sub,
+                        # everything underneath the subds can go
+                        paths=None,
+                        what=what,
+                        reckless=reckless,
+                        recursive=False,
```
bpoldrack, line 370: Shouldn't `recursive` and `recursion_limit` be passed on here, too?

bpoldrack (commented):
Thanks @adswa!

Overall reads right, but I think it needs a minor change. Correct me if I'm wrong.

datalad/distributed/drop.py
bpoldrack: Seems to me we would run into the same issue one level down otherwise, as the condition for the entire change wouldn't be triggered.

adswa: Hmm, I think we wouldn't because ``path`` isn't set, in which case it would kill everything it finds under there

adswa: I have amended the test to show this

adswa:
I have restarted the appveyor build I cancelled on Friday. There is one Travis run that failed, but it looks like something in the test setup glitched and caused widespread failures

datalad/distributed/drop.py
bpoldrack: Thx!

bpoldrack:
AppVeyor fails are #7320

Travis failed with only one build and this one is - once more, sigh - just completely failing to start up the http server for the tests. However, that build should not be relevant WRT to correctness of this PR.

bpoldrack (approved):

Issue #6613, opened 2022-04-05
Design / developer docs for how&when to implement exception raises/ logging / results yielding / hints


jsheunis:
The developer docs currently do not contain specific suggestions and/or examples for how and when to use different forms of user messages. E.g. when to use `raise`, when to turn an exception into an error result or not, when to decide if an exception should be left to break the flow at some point or not, what to do when command arguments are misspecified, and loads more usage patterns that I can't think of now.

This would be very useful for both new developers of datalad itself, and those developing new datalad extensions and not wanting to reinvent the wheel for many of this functionality.

This issue is meant to gather comments on this topic, and specifically also start thinking about typical use cases and examples that can be used to illustrate best practices for user messaging.

@datalad/developers

jsheunis:
Some items that perhaps relate to the same topic and/or could provide context (non-exhaustive, feel free to add):
- https://github.com/datalad/datalad/issues/6114
- https://github.com/datalad/datalad/issues/6403
- https://github.com/datalad/datalad/issues/6258
- https://github.com/datalad/datalad/issues/5919
- https://github.com/datalad/datalad/issues/5784
- For datalad-catalog, I need to completely refactor the user messages once I have a good grasp of what to implement when

mslw:
Recording comments made during yesterday's dev call. Hope the comment author don't mind me repeating them, and I didn't twist anything.

@yarikoptic regarding yield vs raise when an error occurs:
- If a function has an interface to yield a result record, then yield an error result
- If it's a low level python function, then there is no other choice than to raise an error

@bpoldrack 
- Also depends on how recoverable the situation is
- DataLad has an `--on-failure continue` option - for example when operating on multiple datasets, even if there is an error in one, it may be desired to continue with others (therefore error result is preferable for such cases)
- However, mis-specified options should cause an error, because when it is unknown what to do, then it's hard to say there was a result

yarikoptic:
minor clarification:

> * If it's a low level python function, then there is no other choice than to raise an error

an error, or log a DEBUG, INFO or WARNING msg depending on the situation (not always it is an error, but could be a hint,  etc)

jsheunis:
Some notes based on today's discussion between @jsheunis, @mslw and @bpoldrack (please add/adapt)

---

### Considerations

There are two main things that are being considered:
1. **Documenting existing code** (in the design docs) in order to clarify how and when to use existing functionality. This functionality includes (not an exhaustive list):
   - creating result records
   - rendering results
[...]

PR #7310, opened 2023-03-01
DOC: Add a "User messaging" design doc

jsheunis:
This PR:
- Aims to address https://github.com/datalad/datalad/issues/6613
- Adds a design specification "User messaging: result records vs exceptions vs logging"
- This specification aims to delineate the applicable contexts for using result records,  exceptions, logging, or other types of user messaging methods.

TODOs:
- [x] @datalad/developers: all please review the suggested design, since it will set the specification for future changes to the code base (+ extensions), as well as refactoring efforts
- [x] Examples to be added to help developers apply the design specifications to their own extension code or to core contributions
- [x] A description of the UI module and its use to be added
- [x] Add changelog entry

### DIFF
--- a/docs/source/design/user_messaging.rst
+++ b/docs/source/design/user_messaging.rst
@@ -9,33 +9,15 @@ User messaging: result records vs exceptions vs logging
 
 .. topic:: Specification scope and status
 
-   This specification aims to delineate the applicable contexts for using result records,
-   exceptions, logging, or other types of user messaging methods.
+   This specification provides a partial overview of the current implementation.
 
 Motivation
 ==========
 
-Design specifications exist for :ref:`chap_design_result_records`,
-:ref:`chap_design_exception_handling`, :ref:`chap_design_log_levels`, and 
-:ref:`chap_design_progress_reporting`, and an ideal yet elusive goal is the consistent
-application of these methods in the applicable contexts. 
-
-From a **developer's perspective**, a specification for the proper use of specific user
-messaging methods creates a best practice standard that simplifies the implementation
-process, minimizes code duplication, and ensures consistency in the code base. This
-applies both to development in DataLad core as well as in DataLad extensions.
-
-From a **user's perspective**, it is imperative to receive the most appropriate and
-unambiguous message relating to the current context or action.
-
-Discrepancies (on the development side) and resulting ambiguity (on the user side) arise
-when the chosen user messaging methods are implemented interchangeably, inconsistently,
-within contexts that do not warrant them, and in ways that do not inform users
-appropriately.
-
-Consequently, the motivation of this specification is to delineate the applicable
-contexts for using result records, exceptions, logging, or other types of
-user messaging processes.
+This specification delineates the applicable contexts for using
+:ref:`result records <chap_design_result_records>`, :ref:`exceptions <chap_design_exception_handling>`,
+:ref:`progress reporting <chap_design_progress_reporting>`, specific :ref:`log levels <chap_design_log_levels>`,
+or other types of user messaging processes.
 
 
 Specification
@@ -44,23 +26,24 @@ Specification
 Result records
 --------------
 
-**Result records are the standard and preferred return value** format for all
-DataLad commands. Result records are routinely inspected throughout the code base,
-and are used to inform generic error handling, as well as particular calling commands
-on how to proceed with a specific operation.
+**Result records are the only return value format** for all DataLad interfaces.
 
-Yield result records when:
+Constrasting with classic Python interfaces that return specific non-annotated values,
+DataLad interfaces implement message passing by yielding :ref:`result records <chap_design_result_records>`
+that are associated with individual operations. Result records are routinely inspected throughout
+the code base and their annotations are used to inform general program flow and error handling.
 
-1. A generic and standard way of error handling is considered useful
-2. A process can reasonably be continued despite evident errors or shortcomings in
-   its subprocesses
+Command calls can include an ``on_failure`` parameterization to specify how to
+proceed with a particular operation if a returned result record is
+:ref:`classified as a failure result <target-result-status>`. Command calls can
+also can include a ``result_renderer`` parameterization to explicitly enable or
+disable the handling and rendering of result records.
 
-Based on the ``status`` field of a result record, a result is categorized into
-*success* (``ok``, ``notneeded``) and *failure* (``impossible``, ``error``).
-Depending on the ``on_failure`` parameterization of a command call (``on_failure='stop'``,
-``on_failure='continue'``, or ``on_failure='ignore'``), any failure-result
-emitted by a command can lead to an ``IncompleteResultsError`` being raised on command
-exit, or a non-zero exit code on the command line.
+Developers should be aware that external callers will use command call parameterizations
+that can selectively ignore or act on result records, and that the process should therefore
+yield meaningful result records. If, in turn, the process itself receives a set of result
+records from a sub-process, these should be inspected individually in order to identify result
+values that could require re-annotation or status re-classification.
 
 
 Exception handling
@@ -71,7 +54,7 @@ the offending action**.
 
 More specifically, raise an exception when:
 
-1. A command's parameter specifications are violated
+1. A DataLad interface's parameter specifications are violated
 2. An additional requirement (beyond parameters) for the successful running of a
    command, function, or process is not met
 
@@ -80,19 +63,14 @@ is, given the context within which the user/caller triggered the action.
 This is achieved directly via a (re)raised exception, as opposed to logging messages or
 results records which could be ignored or unseen by the user.
 
-If the relevant caller receives a set of result records, these should be inspected
-individually in order to identify result status values that could require an exception
-to be raised to the user/caller. Depending on developer-defined logic, an exception can
-then be raised with an unambiguous message that excludes internal and intermediate
-result messages.
-
-Additionally, in the case of a complex set of dependent actions it could be expensive to
-confirm parameter violations. In such cases, initial sub-processes might already generate
-result records that have to be inspected by the caller, and it could be practically better
-to yield a result record (with ``status=[error|impossible]``) to communicate the failure.
-It would then be up to the upstream caller to decide whether to specify
-``on_failure='ignore'`` or whether to inspect individual result records and turn them
-into exceptions or not.
+.. note::
+   In the case of a complex set of dependent actions it could be expensive to
+   confirm parameter violations. In such cases, initial sub-processes might already generate
+   result records that have to be inspected by the caller, and it could be practically better
+   to yield a result record (with ``status=[error|impossible]``) to communicate the failure.
+   It would then be up to the upstream caller to decide whether to specify
+   ``on_failure='ignore'`` or whether to inspect individual result records and turn them
+   into exceptions or not.
 
 
 Logging
@@ -101,20 +79,28 @@ Logging
 Logging provides developers with additional means to describe steps in a process,
 so as to **allow insight into the program flow during debugging** or analysis of e.g.
 usage patterns. Logging can be turned off externally, filtered, and redirected. Apart from
-the log-level and message, it is not inspectable and cannot be used to control the logic
-or flow of a program.
+the :ref:`log-level <chap_design_log_levels>` and message, it is not inspectable and
+cannot be used to control the logic or flow of a program.
 
-Importantly, logging is not a user messaging method. Therefore:
+Importantly, logging should not be the primary user messaging method for command outcomes,
+Therefore:
 
-1. No command should rely on logging for user communication.
+1. No command should rely solely on logging for user communication
+2. Use logging for in-progress user communication via the mechanism for :ref:`progress reporting <chap_design_progress_reporting>`
+3. Use logging to inform debugging processes
 
 
 UI Module
 ---------
 
-.. note::
-   TODO
+The :mod:`~datalad.ui` module provides the means to communicate information
+to the user in a user-interface-specific manner, e.g. via a console, dialog, or an iPython interface.
+Internally, all DataLad results processed by the result renderer are passed through the UI module.
 
+Therefore: in cases where existing user communication processes are not appropriate,
+developers should let explicit user communication happen through the UI module
+as it provides the flexibility to adjust to the present UI. Specifically,
+:py:func:`datalad.ui.message` allows passing a simple message via the UI module.
 
 
 Examples
--- a/docs/source/design/user_messaging.rst
+++ b/docs/source/design/user_messaging.rst
@@ -9,7 +9,7 @@ User messaging: result records vs exceptions vs logging
 
 .. topic:: Specification scope and status
 
-   This specification provides a partial overview of the current implementation.
+   This specification provides a partial overview of the implementation goal.
 
 Motivation
 ==========
@@ -98,10 +98,9 @@ The :mod:`~datalad.ui` module provides the means to communicate information
 to the user in a user-interface-specific manner, e.g. via a console, dialog, or an iPython interface.
 Internally, all DataLad results processed by the result renderer are passed through the UI module.
 
-Therefore: in cases where existing user communication processes are not appropriate,
-developers should let explicit user communication happen through the UI module
-as it provides the flexibility to adjust to the present UI. Specifically,
-:py:func:`datalad.ui.message` allows passing a simple message via the UI module.
+Therefore: unless the criteria for logging apply, developers should let explicit user communication
+happen through the UI module as it provides the flexibility to adjust to the present UI.
+Specifically, :py:func:`datalad.ui.message` allows passing a simple message via the UI module.
 
 
 Examples
--- a/docs/source/design/user_messaging.rst
+++ b/docs/source/design/user_messaging.rst
@@ -28,8 +28,9 @@ Result records
 
 **Result records are the only return value format** for all DataLad interfaces.
 
-Constrasting with classic Python interfaces that return specific non-annotated values,
-DataLad interfaces implement message passing by yielding :ref:`result records <chap_design_result_records>`
+Contrasting with classic Python interfaces that return specific non-annotated values,
+DataLad interfaces (i.e. subclasses of :py:class:`datalad.interface.base.Interface`)
+implement message passing by yielding :ref:`result records <chap_design_result_records>`
 that are associated with individual operations. Result records are routinely inspected throughout
 the code base and their annotations are used to inform general program flow and error handling.
 
@@ -55,7 +56,7 @@ the offending action**.
 More specifically, raise an exception when:
 
 1. A DataLad interface's parameter specifications are violated
-2. An additional requirement (beyond parameters) for the successful running of a
+2. An additional requirement (beyond parameters) for the meaningful continuation of a
    command, function, or process is not met
 
 It must be made clear to the user/caller what the exact cause of the exception
@@ -106,5 +107,15 @@ as it provides the flexibility to adjust to the present UI. Specifically,
 Examples
 ========
 
-.. note::
-   TODO
+The following links point to actual code implementations of the respective user
+messaging methods:
+
+- `Result yielding`_
+- `Exception handling`_
+- `Logging`_
+- `UI messaging`_
+
+.. _Result yielding: https://github.com/datalad/datalad/blob/a8d7c63b763aacfbca15925bb1562a62b4448ea6/datalad/core/local/status.py#L402-L426
+.. _Exception handling: https://github.com/datalad/datalad/blob/a8d7c63b763aacfbca15925bb1562a62b4448ea6/datalad/core/local/status.py#L149-L150
+.. _Logging: https://github.com/datalad/datalad/blob/a8d7c63b763aacfbca15925bb1562a62b4448ea6/datalad/core/local/status.py#L158
+.. _UI messaging: https://github.com/datalad/datalad/blob/a8d7c63b763aacfbca15925bb1562a62b4448ea6/datalad/core/local/status.py#L438-L457
--- a/docs/source/design/user_messaging.rst
+++ b/docs/source/design/user_messaging.rst
@@ -65,7 +65,7 @@ results records which could be ignored or unseen by the user.
 
 .. note::
    In the case of a complex set of dependent actions it could be expensive to
-   confirm parameter violations. In such cases, initial sub-processes might already generate
+   confirm parameter violations. In such cases, initial sub-routines might already generate
    result records that have to be inspected by the caller, and it could be practically better
    to yield a result record (with ``status=[error|impossible]``) to communicate the failure.
    It would then be up to the upstream caller to decide whether to specify
--- a/docs/source/design/user_messaging.rst
+++ b/docs/source/design/user_messaging.rst
@@ -36,7 +36,7 @@ the code base and their annotations are used to inform general program flow and 
 Command calls can include an ``on_failure`` parameterization to specify how to
 proceed with a particular operation if a returned result record is
 :ref:`classified as a failure result <target-result-status>`. Command calls can
-also can include a ``result_renderer`` parameterization to explicitly enable or
+also include a ``result_renderer`` parameterization to explicitly enable or
 disable the handling and rendering of result records.
 
 Developers should be aware that external callers will use command call parameterizations
--- a/docs/source/design/user_messaging.rst
+++ b/docs/source/design/user_messaging.rst
@@ -46,6 +46,10 @@ yield meaningful result records. If, in turn, the process itself receives a set 
 records from a sub-process, these should be inspected individually in order to identify result
 values that could require re-annotation or status re-classification.
 
+For user messaging purposes, result records can also be enriched with additional human-readable
+information on the nature of the result, via the ``message`` key, and human-readable hints to
+the user, via the ``hints`` key. Both of these are rendered via the `UI Module`_.
+
 
 Exception handling
 ------------------
@@ -86,7 +90,7 @@ cannot be used to control the logic or flow of a program.
 Importantly, logging should not be the primary user messaging method for command outcomes,
 Therefore:
 
-1. No command should rely solely on logging for user communication
+1. No interface should rely solely on logging for user communication
 2. Use logging for in-progress user communication via the mechanism for :ref:`progress reporting <chap_design_progress_reporting>`
 3. Use logging to inform debugging processes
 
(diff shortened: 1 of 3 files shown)


docs/source/design/user_messaging.rst
```diff
@@ -0,0 +1,124 @@
...
+
+**Result records are the standard and preferred return value** format for all
+DataLad commands. Result records are routinely inspected throughout the code base,
+and are used to inform generic error handling, as well as particular calling commands
+on how to proceed with a specific operation.
+
+Yield result records when:
+
+1. A generic and standard way of error handling is considered useful
+2. A process can reasonably be continued despite evident errors or shortcomings in
```
adswa, line 55: maybe this warrants further explanation? E.g., mention that "on_failure=ignore" or "result_renderer=disabled" provide means to purposefully ignore failures

docs/source/design/user_messaging.rst
```diff
@@ -0,0 +1,124 @@
...
+
+Logging provides developers with additional means to describe steps in a process,
+so as to **allow insight into the program flow during debugging** or analysis of e.g.
+usage patterns. Logging can be turned off externally, filtered, and redirected. Apart from
+the log-level and message, it is not inspectable and cannot be used to control the logic
+or flow of a program.
+
+Importantly, logging is not a user messaging method. Therefore:
+
+1. No command should rely on logging for user communication.
```
adswa, line 109: ouch, there are many commands that are guilty of that. 
I wonder whether the INFO messages that are emitted via logging could be an exception to this rule - they couldn't be communicated via an exception or a result record (because they are neither - they're in-progress communication, such as progress bars.

docs/source/design/user_messaging.rst
```diff
@@ -0,0 +1,124 @@
...
+
+
+UI Module
+---------
+
```
adswa, line 114: one earlier comment by @mih would probably fit in well to this paragraph (it came in the context of "we should never write to ``sys.stout`` explicitly):

> We have to use ui.message -- otherwise this message might never see the light of day, because there could be no terminal. ui.message is used to send a message to whatever the UI actually is. Switching to result renderers solves this automatically -- they do that


so maybe relevant points:

- result renderer rely internally on the UI module
- Communication should always happen through the UI module as it provides the flexibility to adjust to the present UI

adswa (commented):
Thanks a lot for writing this up, this summarizes a lot of principles and guidelines really well. I left a few comments. Overall, I would agree with almost everything you outlined here, and I'm also curious what @yarikoptic and @mih say.

mih (commented):
Thanks for writing this up. I think it will need a few iteration to bring us all on the same page. I have left a bunch of comments.

In general, I'd prefer if we try to find the strongest language that we can. I fear that with shoulds and mays, we and up with something that is hard to use as an actionable specification.

docs/source/design/user_messaging.rst
```diff
@@ -0,0 +1,124 @@
...
+.. _chap_design_user_messaging:
+
+*******************************************************
+User messaging: result records vs exceptions vs logging
+*******************************************************
+
+.. topic:: Specification scope and status
+
+   This specification aims to delineate the applicable contexts for using result records,
+   exceptions, logging, or other types of user messaging methods.
```
yarikoptic, line 13: what @mih above says is to use language similar to the already present entries, where now I think
```suggestion
   This specification provides a partial overview of the current implementation.
```

docs/source/design/user_messaging.rst
```diff
@@ -0,0 +1,124 @@
...
+Motivation
+==========
+
+Design specifications exist for :ref:`chap_design_result_records`,
+:ref:`chap_design_exception_handling`, :ref:`chap_design_log_levels`, and 
+:ref:`chap_design_progress_reporting`, and an ideal yet elusive goal is the consistent
+application of these methods in the applicable contexts. 
+
+From a **developer's perspective**, a specification for the proper use of specific user
+messaging methods creates a best practice standard that simplifies the implementation
```
yarikoptic, line 24:
```suggestion
messaging methods creates the best practice standard that simplifies the implementation
```
or
```suggestion
messaging methods creates a good practice standard that simplifies the implementation
```

docs/source/design/user_messaging.rst
```diff
@@ -0,0 +1,124 @@
...
+process, minimizes code duplication, and ensures consistency in the code base. This
+applies both to development in DataLad core as well as in DataLad extensions.
```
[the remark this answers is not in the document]
yarikoptic: I agree

docs/source/design/user_messaging.rst
```diff
@@ -0,0 +1,124 @@
...
+unambiguous message relating to the current context or action.
+
+Discrepancies (on the development side) and resulting ambiguity (on the user side) arise
+when the chosen user messaging methods are implemented interchangeably, inconsistently,
+within contexts that do not warrant them, and in ways that do not inform users
+appropriately.
+
+Consequently, the motivation of this specification is to delineate the applicable
+contexts for using result records, exceptions, logging, or other types of
+user messaging processes.
```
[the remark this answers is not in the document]
yarikoptic: I would keep initial paragraph bringing in related actors (exceptions, logging, result records). The rest is indeed just amplify the state of ambiguity.

docs/source/design/user_messaging.rst
```diff
@@ -0,0 +1,124 @@
...
+
+
+Specification
+=============
+
+Result records
+--------------
+
+**Result records are the standard and preferred return value** format for all
+DataLad commands. Result records are routinely inspected throughout the code base,
```
yarikoptic, line 48: we don't have a notion of *command*. (there was some recent discussion I fail to find quickly).  Until we formalize it -- might be worth making it more specific:

```suggestion
DataLad `Interface`s. Result records are routinely inspected throughout the code base,
```

docs/source/design/user_messaging.rst
```diff
@@ -0,0 +1,124 @@
...
+and are used to inform generic error handling, as well as particular calling commands
+on how to proceed with a specific operation.
+
+Yield result records when:
```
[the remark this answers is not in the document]
yarikoptic: Since seems to concentrate on "errors", is the intended meaning

```suggestion
Yield result records instead of raising an exception when:
```
?

docs/source/design/user_messaging.rst
yarikoptic: moreover even if it can't continue, it would then just yield and return, i.e. still not resort to an exception. I.e.  not necessarily it can continue but still might need to yield record instead of just raising an exception.

But then taken together - it remains unclear when to raise Exception vs when yield record(s). 

I feel that there is more of us having two notions of functions (or "logic flow") in the codebase:
- *classic Python interfaces* -- return actual values, expected to interrupt operation and raise an `Exception` in case of error, exception bubbles up in the traceback automatically until explicitly handled so logic flow is bubbling up too.
- *datalad interfaces* -- message passing by yielding result records associated with individual items, could potentially continue operation after an error (on one of the items, or just to the next steps in logic).

And then it is pretty much for a developer to choose classical vs the datalad-interface style depending 
- how close to `Interface` layer/code it resides
- that 2nd factor of needing to be able to continue operation in case of an error relating to a specific item, so pretty much for any function where there is a loop over items, OR
- usefulness to preparing full "message" with not only a value but some possible extra annotation (hints etc) which otherwise would not be assignable to a basic return type in the *classical* one.

So I think it is valuable to describe such separation of interfaces, and clarification that *datalad interfaces* is pretty much "message passing" flow of logic, while original *classic* logic results in non-annotated return values, and interrupted logic in case of individual item(s) erroring out.

docs/source/design/user_messaging.rst
```diff
@@ -0,0 +1,124 @@
...
+
+1. A generic and standard way of error handling is considered useful
```
yarikoptic, line 54: well, I can argue that exceptions in Python is a "more generic and standard way of error handling" than message passing.  So by itself such statement is IMHO simply misleading if not incorrect.

docs/source/design/user_messaging.rst
```diff
@@ -0,0 +1,124 @@
...
+
+Exception handling
+------------------
+
+In general, **exceptions should be raised when there is no way to ignore or recover from
+the offending action**.
+
+More specifically, raise an exception when:
+
+1. A command's parameter specifications are violated
```
yarikoptic, line 74: Interface level parameter or do you mean an argument to the function?

If top level Interface -- makes sense.

docs/source/design/user_messaging.rst
```diff
@@ -0,0 +1,124 @@
...
+2. An additional requirement (beyond parameters) for the successful running of a
+   command, function, or process is not met
```
yarikoptic, line 76: I think we have `"impossible"` result records returned in some  of such cases... E.g. `impossible` to install subdataset for some reason (out of space etc), and we want to provide a "hint"/"message" on why.  In principle exception could be raised and do that job indeed, but I would expect *datalad interface* to return a record instead of raising an exception.

yarikoptic (commented):
Some thoughts, haven't finished review though so may be some are premature

docs/source/design/user_messaging.rst
jsheunis: I like this suggestion of describing the datalad approach vs what one would expect with a classic Python approach, thanks @yarikoptic.

> But then taken together - it remains unclear when to raise Exception vs when yield record(s).

I agree, but perhaps that uncertainty is okay and decision should be left up to the developer implementing the specific code (and the subsequent review process) ?

docs/source/design/user_messaging.rst
jsheunis: Ok, makes sense to me to shift the focus to:

- logging should not be the primary a user messaging method for command outcomes
- use logging for in-progress user communication (e.g. info messages or progress bars)

docs/source/design/user_messaging.rst
jsheunis: Is there any useful documentation of the UI module, other than found in docstrings and comments, that we can link to?

docs/source/design/user_messaging.rst
```diff
@@ -0,0 +1,124 @@
...
+It must be made clear to the user/caller what the exact cause of the exception
+is, given the context within which the user/caller triggered the action.
+This is achieved directly via a (re)raised exception, as opposed to logging messages or
+results records which could be ignored or unseen by the user.
+
+If the relevant caller receives a set of result records, these should be inspected
+individually in order to identify result status values that could require an exception
+to be raised to the user/caller. Depending on developer-defined logic, an exception can
+then be raised with an unambiguous message that excludes internal and intermediate
+result messages.
```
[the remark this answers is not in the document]
jsheunis: I think the intended meaning is what you also interpreted, although the focus was supposed to be the improved annotation of the failure. But I understand now that this would not be a good place or way to raise an exception, rather a failure result.

docs/source/design/user_messaging.rst
```diff
@@ -0,0 +1,124 @@
...
+Logging
+-------
+
+Logging provides developers with additional means to describe steps in a process,
+so as to **allow insight into the program flow during debugging** or analysis of e.g.
+usage patterns. Logging can be turned off externally, filtered, and redirected. Apart from
+the log-level and message, it is not inspectable and cannot be used to control the logic
+or flow of a program.
+
+Importantly, logging is not a user messaging method. Therefore:
```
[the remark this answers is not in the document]
jsheunis: makes sense, thanks

docs/source/design/user_messaging.rst
jsheunis: yes indeed

docs/source/design/user_messaging.rst
jsheunis: I agree, that is the reason for this part:

> In the case of a complex set of dependent actions it could be expensive to
confirm parameter violations. In such cases, initial sub-processes might already generate
result records that have to be inspected by the caller, and it could be practically better
to yield a result record (with ``status=[error|impossible]``) to communicate the failure.
It would then be up to the upstream caller to decide whether to specify
``on_failure='ignore'`` or whether to inspect individual result records and turn them
into exceptions or not.

I can turn it into a note to amke it more visible.

docs/source/design/user_messaging.rst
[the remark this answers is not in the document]
jsheunis: Done in [721c3d9](https://github.com/datalad/datalad/pull/7310/commits/721c3d9bce724a35a3235edecb7e8815c584d7fc)

docs/source/design/user_messaging.rst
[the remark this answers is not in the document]
jsheunis: Done in [721c3d9](https://github.com/datalad/datalad/pull/7310/commits/721c3d9bce724a35a3235edecb7e8815c584d7fc)

docs/source/design/user_messaging.rst
```diff
@@ -0,0 +1,124 @@
...
+user messaging processes.
+
+
+Specification
+=============
+
+Result records
+--------------
+
+**Result records are the standard and preferred return value** format for all
```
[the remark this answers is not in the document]
jsheunis: Done in [721c3d9](https://github.com/datalad/datalad/pull/7310/commits/721c3d9bce724a35a3235edecb7e8815c584d7fc)

docs/source/design/user_messaging.rst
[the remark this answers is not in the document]
jsheunis: Restructured the section in [721c3d9](https://github.com/datalad/datalad/pull/7310/commits/721c3d9bce724a35a3235edecb7e8815c584d7fc), and perhaps this warrants a reassessment of these comments?

docs/source/design/user_messaging.rst
```diff
@@ -0,0 +1,124 @@
...
+on how to proceed with a specific operation.
+
+Yield result records when:
+
+1. A generic and standard way of error handling is considered useful
+2. A process can reasonably be continued despite evident errors or shortcomings in
+   its subprocesses
+
+Based on the ``status`` field of a result record, a result is categorized into
+*success* (``ok``, ``notneeded``) and *failure* (``impossible``, ``error``).
```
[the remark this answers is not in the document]
jsheunis: Done in [721c3d9](https://github.com/datalad/datalad/pull/7310/commits/721c3d9bce724a35a3235edecb7e8815c584d7fc)

docs/source/design/user_messaging.rst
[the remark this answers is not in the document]
jsheunis: Restructured the section in [721c3d9](https://github.com/datalad/datalad/pull/7310/commits/721c3d9bce724a35a3235edecb7e8815c584d7fc).

docs/source/design/user_messaging.rst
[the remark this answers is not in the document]
jsheunis: Done in [721c3d9](https://github.com/datalad/datalad/pull/7310/commits/721c3d9bce724a35a3235edecb7e8815c584d7fc)

docs/source/design/user_messaging.rst
jsheunis: Done in [721c3d9](https://github.com/datalad/datalad/pull/7310/commits/721c3d9bce724a35a3235edecb7e8815c584d7fc)

jsheunis:
Note: the PR is still in draft because of the section `Examples` that hasn't been added. This would contain code examples of how to use each user messaging method. However: I would like to know opinions on whether this is actually necessary? I'm leaning towards it being a nice-to-have, but not strictly necessary.

docs/source/design/user_messaging.rst
bpoldrack: I don't think there is.

bpoldrack: I think, given that the document says it's a partial description of what is implemented, this part needs to make clear that it is more about what to aim for than what is.
I very much support the aim "No command should rely solely on logging for user communication", but there are a lot violations of this. (I have a "migration" strategy in mind, but that's for another place)


What I would like to add to that section in some form is, that user messaging and logging generally have a different target audience. That's sort of covered by 3., but 1. somewhat blurs the line. Why that's important is: wording of the respective messages may need to be quite different. User-facing messages only devs can understand are training users to ignore output entirely. Which ties into another aspect of my first remark: I'd like to switch to have no logging being spit out by default (write to file instead). What is intended for "regular user" messaging should become a dedicated message via `datalad.ui`. What is not should go into the logger.

bpoldrack (commented):
Thx for taking that on, @jsheunis!

docs/source/design/user_messaging.rst
jsheunis: @bpoldrack your comments are made under the "UI module" section, but seem to refer to logging (referring to the numbers 1, 2 and 3). Can you clarify where exactly you think changes should be made?

jsheunis:
@datalad/developers thanks for the comments so far. I must still add minor updates after the review of @bpoldrack. And then I would like input on whether we need examples or not? And lastly, a last sanity check on the latest state would probably be a good idea, for those who care strongly about the content here.

docs/source/design/user_messaging.rst
```diff
@@ -0,0 +1,110 @@
...
+
+Specification
+=============
+
+Result records
+--------------
+
+**Result records are the only return value format** for all DataLad interfaces.
+
+Constrasting with classic Python interfaces that return specific non-annotated values,
```
yarikoptic, line 31:
```suggestion
Contrasting with classic Python interfaces that return specific non-annotated values,
```

jsheunis:
Comments by @yarikoptic in a dev call: examples could point to commit-tracked implementations in existing code.

docs/source/design/user_messaging.rst
```diff
@@ -0,0 +1,110 @@
...
+or other types of user messaging processes.
+
+
+Specification
+=============
+
+Result records
+--------------
+
+**Result records are the only return value format** for all DataLad interfaces.
```
yarikoptic, line 29: Please make it a little more specific (reference or description such as "subclasses of `Interface`") on what `DataLad interfaces` are

docs/source/design/user_messaging.rst
jsheunis: corrected in [ee1c43b](https://github.com/datalad/datalad/pull/7310/commits/ee1c43bd67650db725a78ab4f293e181e064e807)

docs/source/design/user_messaging.rst
bpoldrack: Indeed - was meant to refer to the 3 points above this section.

Changes:
If we agree on the notion, that this document is actually more about describing the goal than the status quo, then I would:
- strike the "solely" in point 1 under logging
- change "scope and status" of the entire document accordingly
- change 2nd paragraf in UI module:
   "Therefore: ~in cases where existing user communication processes are not appropriate,~
     developers should let explicit user communication happen through the UI module ..."
   Maybe adding "unless the criteria for logging apply"

Crucial part is "If we agree on the notion ...", though.

jsheunis: Thanks. IMO most if not all of the document describes the implementation goal, so I will update the scope accordingly and make the 

I don't think the "solely" should be removed, because that helps convey the point: logging can be used as a user messaging method (in addition to debugging), but not as the _sole_ method for a _command_.

docs/source/design/user_messaging.rst
```diff
@@ -0,0 +1,124 @@
...
+``on_failure='continue'``, or ``on_failure='ignore'``), any failure-result
+emitted by a command can lead to an ``IncompleteResultsError`` being raised on command
+exit, or a non-zero exit code on the command line.
+
+
+Exception handling
+------------------
+
+In general, **exceptions should be raised when there is no way to ignore or recover from
+the offending action**.
```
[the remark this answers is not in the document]
yarikoptic: If this statement is concerning only `Interface`s implementations, then I would agree. But then there is no clarity on either it concerns those or any part of the code, e.g. a function, which might be just `compute_y(x)` and which 
- if just a regular classical Python function -- would raise ValueError or TypeError exception
- if somehow is "an interface" -- yield "impossible" record.

how does developer to decide between the two?

Overall -- is this document specifically just about subclasses of Interfaces or only the first section on return records is?

docs/source/design/user_messaging.rst
```diff
@@ -0,0 +1,120 @@
...
+Logging provides developers with additional means to describe steps in a process,
+so as to **allow insight into the program flow during debugging** or analysis of e.g.
+usage patterns. Logging can be turned off externally, filtered, and redirected. Apart from
+the :ref:`log-level <chap_design_log_levels>` and message, it is not inspectable and
+cannot be used to control the logic or flow of a program.
+
+Importantly, logging should not be the primary user messaging method for command outcomes,
+Therefore:
+
+1. No command should rely solely on logging for user communication
```
yarikoptic, line 89:
```suggestion
1. No interface should rely solely on logging for user communication
```
since we talk about those and not commands

docs/source/design/user_messaging.rst
```diff
@@ -0,0 +1,120 @@
...
+
+
+UI Module
+---------
+
+The :mod:`~datalad.ui` module provides the means to communicate information
+to the user in a user-interface-specific manner, e.g. via a console, dialog, or an iPython interface.
+Internally, all DataLad results processed by the result renderer are passed through the UI module.
+
+Therefore: unless the criteria for logging apply, developers should let explicit user communication
```
yarikoptic, line 101: I feel that it might be worth also explicitly hint that it is when it cannot be just part of the result record, e.g.

```suggestion
Therefore: unless the criteria for logging apply, and necessary to deliver message to the user 
is not "message" of the result record,  developers should let explicit user communication
```

docs/source/design/user_messaging.rst
```diff
@@ -0,0 +1,120 @@
...
+proceed with a particular operation if a returned result record is
+:ref:`classified as a failure result <target-result-status>`. Command calls can
+also include a ``result_renderer`` parameterization to explicitly enable or
+disable the rendering of result records.
+
+Developers should be aware that external callers will use command call parameterizations
+that can selectively ignore or act on result records, and that the process should therefore
+yield meaningful result records. If, in turn, the process itself receives a set of result
+records from a sub-process, these should be inspected individually in order to identify result
+values that could require re-annotation or status re-classification.
```
yarikoptic, line 47: Should there be a small paragraph on "message" and "hints" keys in the record are to be used to provide human-readable feedback to the user.  I think it would be valuable here as logging and ui talks about interfacing to the user and we do not even mention here how result records can be used for that.

docs/source/design/user_messaging.rst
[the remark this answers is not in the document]
jsheunis: > Overall -- is this document specifically just about subclasses of Interfaces or only the first section on return records is?

I think it's both. The main theme is user messaging, and the reality is that most of that happens in response to a call via the `Interface` class. But also: we aren't giving general advice i.e. it should not be seen as a comment on what to do with general/classic Python code.

docs/source/design/user_messaging.rst
jsheunis: Added in fee7a8f7a107ffc44739935a84e4415acdf3f715

docs/source/design/user_messaging.rst
jsheunis: Edited in fee7a8f7a107ffc44739935a84e4415acdf3f715

docs/source/design/user_messaging.rst
jsheunis: Fixed in fee7a8f7a107ffc44739935a84e4415acdf3f715

docs/source/design/user_messaging.rst
```diff
@@ -0,0 +1,128 @@
...
+
+UI Module
+---------
+
+The :mod:`~datalad.ui` module provides the means to communicate information
+to the user in a user-interface-specific manner, e.g. via a console, dialog, or an iPython interface.
+Internally, all DataLad results processed by the result renderer are passed through the UI module.
+
+nd necessary to deliver message to the user 
+	is not "message" of the result record,  developers should let explicit user communication
```
adswa, line 106: I feel like this is a copy-paste victim?

jsheunis: LOL thanks for catching. indeed.

jsheunis: fixed in fc28212154d8f595de6d81d1fdcbe22466a81be0

docs/source/design/user_messaging.rst
```diff
@@ -0,0 +1,125 @@
...
+
+**Result records are the only return value format** for all DataLad interfaces.
+
+Contrasting with classic Python interfaces that return specific non-annotated values,
+DataLad interfaces (i.e. subclasses of :py:class:`datalad.interface.base.Interface`)
+implement message passing by yielding :ref:`result records <chap_design_result_records>`
+that are associated with individual operations. Result records are routinely inspected throughout
+the code base and their annotations are used to inform general program flow and error handling.
+
+Command calls can include an ``on_failure`` parameterization to specify how to
```
yarikoptic, line 37:
```suggestion
DataLad interface calls can include an ``on_failure`` parameterization to specify how to
```

docs/source/design/user_messaging.rst
```diff
@@ -0,0 +1,125 @@
...
+proceed with a particular operation if a returned result record is
+:ref:`classified as a failure result <target-result-status>`. Command calls can
+also include a ``result_renderer`` parameterization to explicitly enable or
+disable the rendering of result records.
+
+Developers should be aware that external callers will use command call parameterizations
```
yarikoptic, line 43:
```suggestion
Developers should be aware that external callers will use DataLad interface call parameterizations
```

docs/source/design/user_messaging.rst
```diff
@@ -0,0 +1,125 @@
...
+------------------
+
+In general, **exceptions should be raised when there is no way to ignore or recover from
+the offending action**.
+
+More specifically, raise an exception when:
+
+1. A DataLad interface's parameter specifications are violated
+2. An additional requirement (beyond parameters) for the meaningful continuation of a
+   command, function, or process is not met
```
yarikoptic, line 64:
```suggestion
   DataLad interface, function, or process is not met
```

docs/source/design/user_messaging.rst
```diff
@@ -0,0 +1,125 @@
...
+
+Contrasting with classic Python interfaces that return specific non-annotated values,
+DataLad interfaces (i.e. subclasses of :py:class:`datalad.interface.base.Interface`)
+implement message passing by yielding :ref:`result records <chap_design_result_records>`
+that are associated with individual operations. Result records are routinely inspected throughout
+the code base and their annotations are used to inform general program flow and error handling.
+
+Command calls can include an ``on_failure`` parameterization to specify how to
+proceed with a particular operation if a returned result record is
+:ref:`classified as a failure result <target-result-status>`. Command calls can
```
yarikoptic, line 39:
```suggestion
:ref:`classified as a failure result <target-result-status>`. DataLad interface calls can
```

yarikoptic (changes requested):
quickly spotted the `NameError: name 'command' is not defined` so left some suggestions ;-)

jsheunis:
@yarikoptic thanks, I committed the suggestions

yarikoptic:
Thank you @jsheunis for pushing it through!

PR #7314, opened 2023-03-08
Add scriv to devel-utils

yarikoptic:
I don't really think it is worth changelog


yarikoptic:
IMHO  a nobrainer. travis fail + stall -- known AFAIK, unrelated

Issue #6884, opened 2022-07-24
Introduce type hints through-out + mypy testing.


yarikoptic:
We have quite complex typing in places, and I think we would greatly benefit from type hints (instead of just documenting them in the docstrings in numpy notation) + adding mypy testing.

TODOs
- [ ] mypy on some files which are already annotated and/or clean (no errors) in mypy
- [ ] add to `CONTRIBUTING.md` instructions on "how" (since we support >= 3.7, some `__future__` imports etc could be needed etc).  Good example PRs:
    - https://github.com/datalad/datalad/pull/7317
    - https://github.com/datalad/datalad/pull/7318 
- [ ] type hints throughout the project
  - [x] #6889
  - [x] `datalad/utils.py`
  - [ ] `datalad/support` , in particular `*repo.py` classes since they are "the next core" in terms of importance IMHO
      - [x] `datalad/support/gitrepo.py`
      - [ ] `datalad/support/annexrepo.py`
  - [ ] `datalad/core` - the "core"
  - [ ] .... add yours ...
  - [ ] mypy plugin to support our dynamic "API interfaces" . [mypy docs](https://mypy.readthedocs.io/en/stable/extending_mypy.html)
- [ ] mypy entire `datalad`

Could be done slowly growing amount of type hints annotations. Probably should be done through commits to `maint` so we reduce possible conflicts in the long-to-release `master`. Alternatively/complimentary we could have a short "hinting hackathon".

WDYT @datalad/developers ?

jwodder:
Note that datalad's dynamic loading of methods will not play nice with mypy.  At a certain point (I'm not sure what triggers type-checking of method calls; it may or may not be automatic for all classes), calling any dynamically-loaded method on a Dataset object in an annotated function will produce an error in mypy unless a `# type: ignore[attr]` comment is added to the line.  The proper way to deal with this is to write a dedicated mypy plugin for datalad, which is something I do not know how to do and for which the official documentation is rather lacking.

yarikoptic:
Thank you @jwodder for the valuable feedback.  Added an item to the list. If you ever feel like discovering "how to" accomplish the mission on that, it would be awesome!

yarikoptic:
> The proper way to deal with this is to write a dedicated mypy plugin for datalad, which is something I do not know how to do and for which the official documentation is rather lacking.

[...]

PR #7317, opened 2023-03-08
Type-annotate almost all of `datalad/utils.py`

jwodder:
Part of #6884.

The only functions left unannotated are `map_items()` and `optional_args()`.  A few `ensure_*` functions have been marked as needing better annotations at some point.


yarikoptic:
FWIW - linters failed due to codespell picking up new typos now, I have pushed fix for that straight into maint, so rebase should resolve that .

travis failed tests -- seems indeed result of this PR changes, at least I confirmed that for `datalad/cli/tests/test_parser.py::test_fail_with_short_help` -- please check them out and provide fixes, ideally in the code, not tests.

datalad/utils.py
```diff
@@ -760,8 +820,15 @@ def ensure_result_list(r):
...
+@overload
+def ensure_list_from_str(s: str, sep: str='\n') -> Optional[list[str]]:
+    ...
 
-def ensure_list_from_str(s, sep='\n'):
+@overload
+def ensure_list_from_str(s: list[T], sep: str='\n') -> Optional[list[T]]:
+    ...
+
+def ensure_list_from_str(s: str | list[T], sep: str='\n') -> Optional[list[str]] | Optional[list[T]]:
```
yarikoptic, line 831: according to https://docs.python.org/3/library/typing.html#typing.overload type annotations of the actual implementation is ignored by type checking, thus IMHO it is better to not even bother type annotating it since it just mixes up more specific type annotations specified above in `@overload`'ed signatures and thus can somewhat misguide (and as you know I dislike duplication). Hence

```suggestion
def ensure_list_from_str(s, sep='\n'):
```

datalad/utils.py
```diff
@@ -776,8 +843,15 @@ def ensure_list_from_str(s, sep='\n'):
...
+@overload
+def ensure_dict_from_str(s: str, sep: str = '\n') -> Optional[dict[str, str]]:
+    ...
+
+@overload
+def ensure_dict_from_str(s: dict[K, V], sep: str = '\n') -> Optional[dict[K, V]]:
+    ...
 
-def ensure_dict_from_str(s, **kwargs):
+def ensure_dict_from_str(s: str | dict[K, V], sep: str = '\n') -> Optional[dict[str, str]] | Optional[dict[K, V]]:
```
yarikoptic, line 854: For similar reasons to above
```suggestion
def ensure_dict_from_str(s, sep='\n'):
```

jwodder:
Tests should be fixed now.

datalad/utils.py
```diff
@@ -610,7 +654,15 @@ def rmtemp(f, *args, **kwargs):
...
-def file_basename(name, return_ext=False):
+@overload
+def file_basename(name: str | Path, return_ext: Literal[True]) -> tuple[str, str]:
+    ...
+
+@overload
+def file_basename(name: str | Path, return_ext: Literal[False] = False) -> str:
+    ...
+
+def file_basename(name: str | Path, return_ext: bool =False) -> str | tuple[str, str]:
```
yarikoptic, line 665: Due to use of more specific `@overload`s

```suggestion
def file_basename(name, return_ext=False):
```

datalad/utils.py
```diff
@@ -867,7 +943,7 @@ def ensure_unicode(s, encoding=None, confidence=None):
         return s.decode(encoding)
 
 
-def ensure_bool(s):
+def ensure_bool(s: str | bool) -> bool:
```
yarikoptic, line 946: I wonder if it should be
```suggestion
def ensure_bool(s: str | Any) -> bool:
```

or at least

```suggestion
def ensure_bool(s: str | bool | int) -> bool:
```
since we do test for `int` in test_assure_bool

datalad/utils.py
```diff
@@ -906,12 +982,17 @@ def unique(seq, key=None, reverse=False):
...
-    trans = reversed if reverse else lambda x: x
+    if reverse:
+        def trans(x: Sequence[T]) -> Iterable[T]:
+            return reversed(x)
+    else:
+        def trans(x: Sequence[T]) -> Iterable[T]:
+            return x
 
-    if not key:
+    if key is None:
```
yarikoptic, line 995: curious: is such change needed for type annotation as otherwise it cannot figure out something?

Why it is the right thing to do in general, I think: I guess in principle there could be a callable instance of a class which has some `__bool__` which would evaluate it to False and thus this check would be incorrectly handled here. Is that the "functional" (or may be what type checking potentially complains) reason @jwodder ?

datalad/utils.py
```diff
@@ -921,6 +1002,8 @@ def unique(seq, key=None, reverse=False):
     return out[::-1] if reverse else out
 
 
+# TODO: Annotate (would be made easier if the return value was always a dict
+# instead of doing `v.__class__(...)`)
 def map_items(func, v):
```
yarikoptic, line 1007: well, true, but IMHO it would be nice to keep it generic. May be there is some `Mappable` type "behind" `dict` which could be used here so anything with `.items()` kinda would be usable with it?

if not -- we could restrict to `dict` in type annotation (while adding a comment that in principle could be applied to any mappable thing with `.items()`) -- there is only 1 use of this helper and it is on `dict`.

datalad/utils.py
```diff
@@ -79,20 +123,26 @@
 on_osx = platform_system == 'darwin'
 on_linux = platform_system == 'linux'
 
+T = TypeVar("T")
+K = TypeVar("K")
+V = TypeVar("V")
+P = ParamSpec("P")
```
yarikoptic, line 129: would it make sense to define some `datalad.typing` which would have above python version conditional imports and these variables defined so we could avoid needing to copy paste such things in every file needing to use similar constructs?

datalad/utils.py
```diff
@@ -2172,7 +2275,15 @@ def open_r_encdetect(fname, readahead=1000):
...
-def read_file(fname, decode=True):
+@overload
+def read_file(fname: str | Path, decode: Literal[True] =True) -> str:
+    ...
+
+@overload
+def read_file(fname: str | Path, decode: Literal[False]) -> bytes:
+    ...
+
+def read_file(fname: str | Path, decode: Literal[True, False] =True) -> str | bytes:
```
yarikoptic, line 2286:
```suggestion
def read_file(fname, decode=True):
```
due to overloads

datalad/utils.py
```diff
@@ -2430,7 +2551,7 @@ def create_tree(path, tree, archives_leading_dir=True, remove_existing=False):
         os.makedirs(path)
 
     if isinstance(tree, dict):
-        tree = tree.items()
+        tree = list(tree.items())
```
yarikoptic, line 2554: pragmatically doesn't matter but in principle -- why is this needed here?  ideally we should be ok with `tree` remaining some iterable instead of populated into a `list`

datalad/utils.py
```diff
@@ -1715,15 +1803,15 @@ def dlabspath(path, norm=False):
     if not isabs(path):
         # if not absolute -- relative to pwd
         path = op.join(getpwd(), path)
-    return normpath(path) if norm else path
+    return normpath(path) if norm else str(path)
```
yarikoptic, line 1806: this changes "functionality" since before it could have returned original `Path`.  I wonder if there would be no effect on any use...

datalad/utils.py
jwodder: I believe omitting the annotations on the actual implementation of the function would mean that mypy wouldn't type-check the body against the signature (in fact, the body wouldn't be type-checked at all).

datalad/utils.py
jwodder: `tree` is annotated in the function signature as a tuple, list, or dict, and the return value of `tree.items()` is none of those, so we have to convert it to one of those.

datalad/utils.py
jwodder: The original path would only have been returned if it was absolute and `norm` was false, which is a rather specific situation for the return type to suddenly change.  The only time you could rely on getting a `Path` back would be instances in which calling the function was pointless to begin with.

datalad/utils.py
jwodder: I don't think this was done for typing reasons; I think I just let my dislike of that kind of sloppy conditional (If you *mean* `is None`, *say* `is None`) get away from me for a bit.

datalad/utils.py
jwodder: Note that the tests for this function specifically check that it returns the input type when it's a custom class.

datalad/utils.py
jwodder: Probably.

datalad/utils.py
jwodder: Done in 43cb9f03d.

datalad/utils.py
yarikoptic: Please do it then in this or subsequent PR.

datalad/utils.py
yarikoptic: You are correct!

<details>
<summary>on this simple example based on the docs</summary> 

```python
from typing import overload,Union

@overload
def process(response: None) -> None:
    ...
@overload
def process(response: int) -> tuple[int, str]:
    ...
@overload
def process(response: bytes) -> str:
    ...
#def process(response:None|int|bytes) -> None|tuple[int,str]|str:
def process(response):
    # incompatible return e.g. in case of int or bytes
    return response
```
</details>

mypy   returns no errors, if I uncomment and use that line with combined signature - it picks it up but not because of overloads seems to me

<details>
<summary>on a more elaborate one where it is specifically violating overloads</summary> 

```python
from typing import overload,Union

@overload
def process(response: None) -> None:
    ...
@overload
def process(response: int) -> tuple[int, str]:
    ...
@overload
def process(response: bytes) -> str:
    ...
#def process(response:None|int|bytes) -> None|tuple[int,str]|str:
def process(response):
    # incompatible returns
    if isinstance(response, int):
        # specifically wrong for int -- must be tuple according to overload
        return str(int)
    elif isinstance(response, bytes):
        # specifically wrong for bytes
        return (1, str(bytes))
    # ok for remaining None
    return response


```
</details>

it seems to not bother finding problem in either case!  Seems like overloads are completely useless.

there is over 300 open issues mentioning `overload` https://github.com/python/mypy/search?q=overload&state=open&type=issues , so I think it is just mypy implementation is simply not there at all yet :-/ 

ok, let's proceed as is then, hopefully later mypy implementation would stick to what is written in the python docs and would simply ignore the annotation of the final implementation function and use overloads instead.

datalad/utils.py
jwodder: Done: b376ffca4

yarikoptic (approved):
Although the PR is quite large as for `maint`, but since it is really just internal refactoring and annotation and should not (at least according to review/testing) to have side effects, I would proceed with merging on Monday unless I hear otherwise.
It would be great to see DataLad core properly type annotated.

yarikoptic:
will do merge locally and push to address small conflict in listing.

yarikoptic-gitmate:

PR #7321, opened 2023-03-13
MNT: remove leftover distutils import

adswa:
Originally, #6839 replaced usage of distutils with Looseversion as a fix for #6307. Somehow, this one isolated import of distutils seems to have made it back. This is a quick PR to remove it.
A changelog entry shouldn't be needed for this.


mslw (approved):
Well spotted!

PR #7309, opened 2023-02-28
WTF - bring back and extend information on metadata extractors etc

yarikoptic:
Ref https://github.com/datalad/datalad-metalad/issues/340 .

In general I feel that implementation for that issue is a good fit for `datalad wtf` so I looked inside to discover a function (`_describe_metadata_elements`) which already did what we want but was not used anywhere!  I discovered that the use (but not definition/body) of it was removed in 616fb1b444c50a38f5ec5ff54b9850085cb540db, so as a first step I brought back use of it. Then

- extended it also with `datalad.metadata.indexers`
- throughout (not only in `metadata` but also `extensions`) removed populating `load_error` with `None` if there were no error. IMHO it is pointless waste of screen real-estate. As our formatter is generic, we probably do not want it always to not print every field with `None` as value, so decided to remove it here.
- added gathering information on possible `version` of extractor
- added gathering information on possible `generation` of extractor which is proposed in https://github.com/datalad/datalad-metalad/pull/351 - so this PR is contingent on arriving to conclusion there
  - we discovered that 'legacy-legacy' specification exists in 2 places -- in `metalad` and `datalad_deprecated` and ideally should be centralized.  Most likely within `metalad` and make `deprecated` depend on it (attn @christian-monch )
- moved from `section_subsection` to `section.subsection` so it matches better and added functionality to request an entire `section`, e.g. `-S metadata` returns all subsections `metadata.*`

Example of output which you would get after `yq '.extensions| keys_unsorted | .[]  | @sh' < extensions.yaml | sed -e "s,[\"'],,g" | xargs pip install`, i.e. installing all datalad extensions registed in datalad-extensions

<details>
<summary>❯ datalad wtf -S metadata</summary> 

# WTF
## metadata.extractors 
  - annex: 
    - distribution: datalad-metalad 0.4.12+45.g5f623bb
    - generation: legacy-legacy
    - module: datalad_metalad.extractors.legacy.annex
  - audio: 
    - distribution: datalad-metalad 0.4.12+45.g5f623bb
    - generation: legacy-legacy
    - module: datalad_metalad.extractors.legacy.audio
  - bids: 
    - distribution: datalad-neuroimaging 0.3.3
    - module: datalad_neuroimaging.extractors.bids
  - bids_dataset: 
    - distribution: datalad-neuroimaging 0.3.3
    - doc: Top level BIDS extractor class interfacing with metalad     Inherits from metalad's DatasetMetadataExtractor class
    - generation: gen4
[...]

PR #7322, opened 2023-03-13
Do not render (empty) WTF report on other records

yarikoptic:
+  + strengthen the test for unavail section (will come handy for #7309 etc, doing here to avoid conflicts etc)

Otherwise we get

	❯ python -c "import datalad.api as dl; out = dl.wtf(sections=['murkie'])"
	# WTF
	# WTF
	...


adswa:
The ``test_gin_cloning`` failures coincide with scheduled maintenance on Gin

adswa (approved):

yarikoptic-gitmate:

PR #7323, opened 2023-03-13
Adopt call_annex_success helper from next

### LINKED ISSUE #6316: Missing feature: `AnnexRepo.call_annex_success`
It is the analog of `GitRepo.call_git_success()`

```py
def call_annex_success(self, args, files=None):
    """Call git-annex and return true if the call exit code of 0.

    All parameters match those described for `call_annex`.

    Returns
    -------
    bool
    """
    try:
        self.call_annex(args, files)
    except CommandError:
        return False
    return True
```

Useful for, e.g., checking whether a remote has a key:

```py
self.call_annex_success(['fsck', '-f', 'origin', '--fast', '--key', ...])
```

adswa:
Originally implemented in datalad-next by @mih and also proposed in #6316, this change migrates a helper from datalad-next that wraps annexrepo.call_annex() with a convenience layer to return True or False depending on the success of the call. I found at least one usage in annexrepo where this helper is useful, and I suspect there are more. Fixes #6316

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -1269,6 +1269,21 @@ class AnnexRepo(GitRepo, RepoInterface):
             files=files,
             protocol=StdOutErrCapture)['stdout']
 
+    def call_annex_success(self, args, files=None):
+        """Call git-annex and return true if the call exit code of 0.
+
+        All parameters match those described for `call_annex`.
+
+        Returns
+        -------
+        bool
+        """
+        try:
+            self.call_annex(args, files)
+        except CommandError:
+            return False
+        return True
+
     def call_annex_items_(self, args, files=None, sep=None):
         """Call git-annex, splitting output on `sep`.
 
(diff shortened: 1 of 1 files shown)


datalad/support/annexrepo.py
```diff
@@ -2761,12 +2776,7 @@ def is_available(self, file_, remote=None, key=False, batch=False):
         annex_input = [key_,] if not remote else [key_, remote]
 
         if not batch:
-            try:
-                out = self.call_annex(['checkpresentkey'] + annex_input)
-                assert(not out)
```
yarikoptic, line 2766: not sure how important but we are loosing now this `assert`ed behavior since in call_annex_success there is no check of output.  It was originally added in the initial implementation in c5ba8b6baa24c379ae41dff0018b152196d243b7 without any special comment on its necessity etc, and we never triggered it AFAIK.  I guess it could be sacrificed.

yarikoptic:
if rebased -- CI should get green

datalad/support/annexrepo.py
adswa: This is what the manpage of this command says about its output:

> Exits 0 if the content is verified present in the remote, or 1 if it is
       verified to not be present in the remote. If there is  a  problem,  the
       special  exit  code  100  is  used,  and  an error message is output to
       stderr.

bpoldrack: FTR: I support not relying on output assumptions.

adswa:
green now

PR #7327, opened 2023-03-15
Type-annotate `datalad/support/globbedpaths.py`

jwodder:
Part of #6884.

### DIFF
--- a/datalad/support/globbedpaths.py
+++ b/datalad/support/globbedpaths.py
@@ -9,16 +9,24 @@
 """Wrapper for globbing paths.
 """
 
+from __future__ import annotations
+
 import glob
-from functools import lru_cache
-from itertools import chain
 import logging
 import os.path as op
+from functools import lru_cache
+from itertools import chain
+from typing import (
+    Iterable,
+    Optional,
+)
 
-from datalad.utils import chpwd
-from datalad.utils import ensure_unicode
-from datalad.utils import getpwd
-from datalad.utils import partition
+from datalad.utils import (
+    chpwd,
+    ensure_unicode,
+    getpwd,
+    partition,
+)
 
 lgr = logging.getLogger('datalad.support.globbedpaths')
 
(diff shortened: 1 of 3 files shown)


datalad/support/globbedpaths.py
```diff
@@ -39,27 +47,30 @@ class GlobbedPaths(object):
...
-            patterns, dots = partition(patterns, lambda i: i.strip() == ".")
+            pattern_strs = list(map(ensure_unicode, patterns))
+            pattern_iter, dots = partition(pattern_strs, lambda i: i.strip() == ".")
             self._maybe_dot = ["."] if list(dots) else []
             self._patterns = [op.relpath(p, start=pwd) if op.isabs(p) else p
-                              for p in patterns]
-        self._cache = {}
+                              for p in pattern_iter]
+        self._cache: dict[str, dict[str, list[str]]] = {}
+        self._expanded_cache: dict[tuple[str, bool, bool], list[str]] = {}
```
yarikoptic, line 66: Thank you for the PR!  But with this (and below) change it seems to go beyond just annotation but also storing expanded cache in a separate data structure.  That can change logic.
- was it introduced just for type annotation or resolved some issue?
- naming is a bit confusing since on https://github.com/datalad/datalad/pull/7327/files#diff-635a44b1a06b74c24841f2e71d5f4fe560d1e77783f59d9fcecc615fdb3b7178L153 we still assign to `_cache`  (not `_expanded_cache`) results from  `self._expand_globs()`.  Could you please elaborate on this change in the commit message so later on becomes clear why this change needed and what effects (or lack of such) are expected in functioning.

jwodder: * I needed to split the dict in two because some of the values are `dict[str, list[str]]` while others are `list[str]`.  If the values had all been kept in the same dict, then an `assert isinstance(...)` would have been required almost every time something was retrieved from the dict.
* I don't care what the new attribute is named; I just picked "`_expanded_cache`" because it's queried with a variable named `key_expanded`.
* There should be zero change in functioning.

yarikoptic: I reviewed it and could not come up with a better name, and indeed it makes it cleaner for type annotation although a little superficial but highlights that even type of stored structure is different.  So let's proceed with as is, thank you.

yarikoptic (approved):

yarikoptic-gitmate:

PR #7333, opened 2023-03-16
Fix datalad special remote error reporting

### LINKED ISSUE #7332: datalad special remote's reporting on error causes is suboptimal
Just observed a user getting `Failed to download from any of 2 locations [None]`.
This gives no idea what the `None` would be about. It's actual purpose is to record the `__cause__` of the original exception raised by the download call within the special remote's `transfer_retrieve`.

bpoldrack:
Utilize `CapturedException.format_with_cause` instead of a custom solution, potentially leading to an error message containing an uninformative `[None]`.
This would also include more than just the first-level `__cause__` (all the way down and include potential `__context__` as well)

Closes #7332

With this change the output reported in #7332 (`Failed to download from any of 2 locations [None]`) changes to:

```
[{'action': 'get',
  'annexkey': 'MD5E-s994556226--ffc1353e08d640c7ba3154d963573732.nii.gz',
  'error_message': 'Failed to download from any of 2 locations ["Failed to '
                   'acquire lock to establish download session at '
                   '/home/<REDACTED>.cache/datalad/locks/downloader-auth.lck '
                   'in 3 attempts.Check following process: PID=2946262 '
                   'CWD=/var/lib/condor/execute/dir_2944629/tmpwoojnx0_/datadir/HCP1200/567052/MNINonLinear '
... <REDACTED>
```

So, much more telling.

### DIFF
--- a/datalad/customremotes/datalad.py
+++ b/datalad/customremotes/datalad.py
@@ -52,12 +52,10 @@ class DataladAnnexCustomRemote(AnnexCustomRemote):
                 return
             except Exception as exc:
                 ce = CapturedException(exc)
-                cause = getattr(exc, '__cause__', None)
-                debug_msg = f"Failed to download {url} for key {key}: {ce}"
-                if cause:
-                    debug_msg += f' [{cause}]'
+                debug_msg = f"Failed to download {url} for key {key}: " \
+                            f"{ce.format_with_cause()}"
                 self.message(debug_msg)
-                error_causes.append(cause)
+                error_causes.append(ce.format_with_cause())
 
         error_msg = f"Failed to download from any of {len(urls)} locations"
         if error_causes:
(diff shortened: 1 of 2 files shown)


datalad/customremotes/datalad.py
```diff
@@ -52,12 +52,10 @@ def transfer_retrieve(self, key, file):
...
                 ce = CapturedException(exc)
-                cause = getattr(exc, '__cause__', None)
-                debug_msg = f"Failed to download {url} for key {key}: {ce}"
-                if cause:
-                    debug_msg += f' [{cause}]'
+                debug_msg = f"Failed to download {url} for key {key}: " \
+                            f"{ce.format_with_cause()}"
                 self.message(debug_msg)
-                error_causes.append(cause)
+                error_causes.append(ce.tb.__cause__)
```
yarikoptic, line 58: I don't see complaints from codecov which means that this code block is probably tests covered -- could you add checking for expected cause to appear somewhere where expected in the test?

bpoldrack: I don't see where this is covered. `test_datalad.py` doesn't appear to test retrieval at all.

yarikoptic-gitmate:

PR #7336, opened 2023-03-16
Extend type-annotations for `datalad/support/path.py`

jwodder:
Part of #6884.

### DIFF
--- a/datalad/support/path.py
+++ b/datalad/support/path.py
@@ -17,13 +17,13 @@ import os
 import os.path as op
 # to not pollute API importing as _
 from collections import defaultdict as _defaultdict
+from collections.abc import Iterator
 from functools import wraps
 from itertools import dropwhile
 from pathlib import (
     Path,
     PurePosixPath,
 )
-from typing import Generator
 
 from ..utils import (
     ensure_bytes,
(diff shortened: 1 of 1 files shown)


datalad/support/path.py
```diff
@@ -74,18 +74,17 @@ def robust_abspath(p):
...
         return abspath(p)
-    except OSError as exc:
+    except OSError:
         if not isabs(p):
             try:
                 os.getcwd()
-                # if no exception raised it was not the reason, raise original
-                raise
-            except:
+            except Exception:
```
yarikoptic, line 81: ha -- so even logic I added in `0.10.3~18^2` (53cfc2d8ef7d8d268031ef5aee1ae925bd89fd8f) was wrong since I was re-raising and then right away going into `except` for it... odd.  This fixes that now I think. BUT I would keep it 
```suggestion
            except:
```
or  at least
```suggestion
            except BaseException:
```
if possible to retain original effect from `getcwd` crashing one way or another.

jwodder: Bare `except:`/`except BaseException:` will also catch `KeyboardInterrupt` and other things that normally shouldn't be caught.

yarikoptic-gitmate:

PR #7341, opened 2023-03-20
Type-annotate `datalad/support/gitrepo.py`

jwodder:
Part of #6884.

### DIFF
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -10,16 +10,28 @@
 
 """
 
+from __future__ import annotations
+
 import logging
 import os
 import os.path as op
 import posixpath
 import re
+import subprocess
 import warnings
-from collections.abc import Mapping
+from collections.abc import (
+    Callable,
+    Iterable,
+    Iterator,
+    Mapping,
+    Sequence,
+)
 from functools import wraps
 from itertools import chain
-from os import linesep
+from os import (
+    PathLike,
+    linesep,
+)
 from os.path import (
     commonprefix,
     curdir,
@@ -32,6 +44,17 @@ from os.path import (
     pardir,
     relpath,
     sep,
+)
+from re import Pattern
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    List,
+    Optional,
+    Tuple,
+    TypeVar,
+    Union,
+    overload,
 )
 
 import datalad.utils as ut
@@ -61,6 +84,15 @@ from datalad.log import log_progress
 from datalad.support.due import (
     Doi,
     due,
+)
+from datalad.typing import (
+    Concatenate,
+    Literal,
+    P,
+    Protocol,
+    Self,
+    T,
+    TypedDict,
 )
 from datalad.utils import (
     Path,
@@ -92,7 +124,13 @@ from .network import (
     PathRI,
     is_ssh,
 )
-from .path import get_parent_paths, get_filtered_paths_
+from .path import (
+    get_filtered_paths_,
+    get_parent_paths,
+)
+
+if TYPE_CHECKING:
+    from datalad.distribution.dataset import Dataset
 
 # shortcuts
 _curdirsep = curdir + sep
@@ -101,9 +139,11 @@ _pardirsep = pardir + sep
 
 lgr = logging.getLogger('datalad.gitrepo')
 
+Option = Union[str, bool, None, List[Union[str, bool, None]], Tuple[Union[str, bool, None], ...]]
+
 
 # outside the repo base classes only used in ConfigManager
-def to_options(split_single_char_options=True, **kwargs):
+def to_options(split_single_char_options: bool = True, **kwargs: Option) -> list[str]:
     """Transform keyword arguments into a list of cmdline options
 
     Imported from GitPython.
@@ -123,10 +163,10 @@ def to_options(split_single_char_options=True, **kwargs):
     -------
     list
     """
-    def dashify(string):
+    def dashify(string: str) -> str:
         return string.replace('_', '-')
 
-    def transform_kwarg(name, value, split_single_char_options):
+    def transform_kwarg(name: str, value: str | bool | None, split_single_char_options: bool) -> list[str]:
         if len(name) == 1:
             if value is True:
                 return ["-%s" % name]
@@ -143,8 +183,7 @@ def to_options(split_single_char_options=True, **kwargs):
         return []
 
     args = []
-    kwargs = dict(sorted(kwargs.items(), key=lambda x: x[0]))
-    for k, v in kwargs.items():
+    for k, v in sorted(kwargs.items()):
         if isinstance(v, (list, tuple)):
             for value in v:
                 args += transform_kwarg(k, value, split_single_char_options)
@@ -153,7 +192,7 @@ def to_options(split_single_char_options=True, **kwargs):
     return args
 
 
-def _normalize_path(base_dir, path):
+def _normalize_path(base_dir: str, path: str) -> str:
     """Helper to check paths passed to methods of this class.
 
     Checks whether `path` is beneath `base_dir` and normalizes it.
@@ -214,8 +253,12 @@ def _normalize_path(base_dir, path):
     return relpath(path, start=base_dir)
 
 
+class _WithPath(Protocol):
+    path: str
+
+
 @optional_args
-def normalize_path(func):
+def normalize_path(func: Callable[Concatenate[_WithPath, str, P], T]) -> Callable[Concatenate[_WithPath, str, P], T]:
     """Decorator to provide unified path conversion for a single file
 
     Unlike normalize_paths, intended to be used for functions dealing with a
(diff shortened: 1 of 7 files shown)


datalad/support/gitrepo.py
```diff
@@ -214,8 +253,12 @@ def _normalize_path(base_dir, path):
     return relpath(path, start=base_dir)
 
 
+class _WithPath(Protocol):
+    path: str
```
yarikoptic, line 257: why not to move it to `datalad/typing.py` as plain `WithPath` and may be even `WithPathT` since it seems you use those whenever construct is not clearly a type, but could be some class.

datalad/support/gitrepo.py
```diff
@@ -397,21 +440,23 @@ class GitProgress(WitlessProtocol):
         CHECKING_OUT: ("Check out", "Things"),
     }
 
-    __slots__ = ('_seen_ops', '_pbars', '_encoding')
+    __slots__ = ('_unprocessed', '_seen_ops', '_pbars')
```
yarikoptic, line 443: I am surprised that it didn't error out before somehow since I was expecting that if `__slots__` specified then assignment of unknown ones to lead to an error. oh well

datalad/support/gitrepo.py
```diff
@@ -448,7 +493,7 @@ def pipe_data_received(self, fd, byts):
                     # it to become a recognizable progress report
                     self._unprocessed = line
 
-    def _parse_progress_line(self, line):
+    def _parse_progress_line(self, line: bytes) -> bool:
```
yarikoptic, line 496: since this is purely internal, lets  make it

```suggestion
    def _parse_progress_line(self, bytes_line: bytes) -> bool:
```

and decode into regular `line` (not `sline`) thus staying more in line with other str line's and minimizing the diff

datalad/support/gitrepo.py
```diff
@@ -1680,7 +1736,10 @@ def get_merge_base(self, commitishes):
         if not commitishes:
             raise ValueError("Provide at least a single value")
         elif len(commitishes) == 1:
-            commitishes = commitishes + [self.get_active_branch()]
+            branch = self.get_active_branch()
+            if branch is None:
+                raise ValueError("Single commitish provided and no active branch")
```
yarikoptic, line 1741: LGTM, Note: in prior cases we would have got `TypeError` though but since it is indeed illegit and unlikely invocation , should be ok. Just FTR

```
❯ python -c 'from datalad.support.gitrepo import GitRepo; print(GitRepo(".").call_git_oneline(["merge-base", "master", None]))'
  File "/home/yoh/proj/datalad/datalad-maint/datalad/dataset/gitrepo.py", line 550, in call_git_oneline
    lines = list(self.call_git_items_(args, files=files,
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/yoh/proj/datalad/datalad-maint/datalad/dataset/gitrepo.py", line 517, in call_git_items_
    for file_no, line in self._generator_call_git(
  File "/home/yoh/proj/datalad/datalad-maint/datalad/dataset/gitrepo.py", line 345, in _generator_call_git
    generator = self._git_runner.run(
                ^^^^^^^^^^^^^^^^^^^^^
  File "/home/yoh/proj/datalad/datalad-maint/datalad/runner/runner.py", line 206, in run
    results_or_iterator = threaded_runner.run()
                          ^^^^^^^^^^^^^^^^^^^^^
  File "/home/yoh/proj/datalad/datalad-maint/datalad/runner/nonasyncrunner.py", line 343, in run
    return self._locked_run()
           ^^^^^^^^^^^^^^^^^^
  File "/home/yoh/proj/datalad/datalad-maint/datalad/runner/nonasyncrunner.py", line 403, in _locked_run
    self.process = Popen(self.cmd, **kwargs)         # nosec
                   ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/subprocess.py", line 1024, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib/python3.11/subprocess.py", line 1834, in _execute_child
    self.pid = _fork_exec(
               ^^^^^^^^^^^
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

datalad/support/gitrepo.py
```diff
@@ -1897,19 +1956,19 @@ def _maybe_open_ssh_connection(self, remote, prefer_push=True):
...
+        namelst = [name] if name else []
         self.call_git(
-            ['remote'] + name + ['update'] + options,
+            ['remote'] + namelst + ['update'] + options,
             expect_stderr=True
         )
 
-    def fetch(self, remote=None, refspec=None, all_=False, git_options=None,
-              **kwargs):
+    def fetch(self, remote: Optional[str] = None, refspec: str | list[str] | None = None, all_: bool = False, git_options: Optional[list[str]] = None,
```
yarikoptic, line 1970: when are we to make it ` ... | None = None` vs ` Optional[...] = None` ?

datalad/support/gitrepo.py
```diff
@@ -3268,13 +3349,16 @@ def _diffstatus_get_state_props(self, f, from_state, to_state,
...
+            elif state == 'clean':
+                assert from_state is not None
+                if 'bytesize' in from_state:
+                    # no change, we can take this old size info
+                    props['bytesize'] = from_state['bytesize']
         if state in ('clean', 'modified', 'deleted', None):
             # assign previous gitsha to any record
             # state==None can only happen for subdatasets that
             # already existed, so also assign a sha for them
+            assert from_sha is not None
```
yarikoptic, line 3361: not 100% sure it must be not a `None`, but ok - let's hope it is so

datalad/support/gitrepo.py
```diff
@@ -772,6 +817,14 @@ def _from_line(cls, line):
         )
 
 
+InfoT = TypeVar("InfoT", FetchInfo, PushInfo)
+
+
+class GitAddOutput(TypedDict):
+    file: str
+    success: bool
```
yarikoptic, line 825: note to myself and other @datalad/developers  -- `TypedDict` is what would save our sanity in formalizing our return record dictionaries, which we produce using https://github.com/datalad/datalad/blob/HEAD/datalad/interface/results.py#L51 .  `TypedDict` allows for `NotRequired` (not sure why not `Optional`) annotation for keys which do not have to be present.

yarikoptic (changes requested):
few possible suggestions, otherwise looks great, thank you @jwodder !

datalad/support/gitrepo.py
jwodder: Are there functions in other files that operate on any type that has a `path` attribute?

datalad/support/gitrepo.py
yarikoptic: oh, ok. Might be better to stick to one or another since otherwise it makes some folks like me wonder on how those two might differ ;)

datalad/support/gitrepo.py
yarikoptic: I would guess so, in particular in `annexrepo.py` since that one is close companion of this one. But also lots of files listed in hit for `git grep -l '\.path\>' `

jwodder: * There are only two top-level functions in `annexrepo.py`, and they both operate on IO handles rather than `WithPath`.
* A large number of hits for that regex doesn't mean that `WithPath` will be useful; most of those can be expected to operate on specific classes.

datalad/support/gitrepo.py
jwodder: `GitProgress` inherits from `WitlessProtocol`, which does not use `__slots__`, and so `GitProgress` will still have a `__dict__` attribute and support assignment of arbitrary attributes.

datalad/support/gitrepo.py
jwodder: Done: 78e0a3890

datalad/support/gitrepo.py
```diff
@@ -3664,14 +3751,15 @@ def _save_add(self, files, git_opts=None):
                     # while there is no git-annex underneath here, we
                     # tend to fake its behavior, so we can also support
                     # this type of messaging
-                    message='\n'.join(r['error-messages'])
-                    if 'error-messages' in r else None,
+                    #message='\n'.join(r['error-messages'])
+                    #if 'error-messages' in r else None,
+                    message=None,
```
yarikoptic, line 3756: note: indeed `r` cannot have `error-messages` in its return since it is 

```
    @staticmethod
    def _process_git_get_output(stdout, stderr=None):
        """Given both outputs (stderr is ignored atm) of git add - process it

        Primarily to centralize handling in both indirect annex and direct
        modes when ran through proxy
        """
        return [{u'file': f, u'success': True}
                for f in re.findall("'(.*)'[\n$]", ensure_unicode(stdout))]
```
so functionally this change would be correct and without side-effect.

yarikoptic:
ok, let's proceed! I had another look and although there are some changes to code, I think overall it just would make it more robust and prevent some loose/buggy behavior possible (but unlikely to be triggered) before.

yarikoptic-gitmate:
PR released in [`0.18.4`](https://github.com/datalad/datalad/releases/tag/0.18.4)

PR #7344, opened 2023-03-22
Add config datalad.ui.interactive and allow non-interactive special remotes

### LINKED ISSUE #7345: UnderAnnexUI is always interactive
It seems, that interactivity detection is entirely dysfunctional for special remotes.

The proof is here: https://ci.appveyor.com/project/mih/datalad/builds/46572731/job/koim0h1pc7m86isa#L878

This seems to be rooted in the fact, that `UnderAnnexUI` is derived from `InteractiveUI` and its `is_interactive` is inherited from `ConsoleLog` where its implementation is to test for self being a subclass of `InteractiveUI`. That test relies on the assumption that interactivity was figured out before. Above link shows, that the result is, however, different from the "original" `is_interactive` utility function in a non-interactive job (assertion fails in CI, passes in interactive, local run).

The consequence of that is, that our special remotes may try to query for credentials in non-interactive jobs, leading to stalling.

### LINKED ISSUE #7349: Make is_interactive configurable
As concluded in https://github.com/datalad/datalad/issues/7345#issuecomment-1481546486, the current interactive session detection in `datalad.utils.is_interactive` is not only if very limited validity, but there's no way of working around it for a user.
This must be configurable.

bpoldrack:
- Makes `is_interactive` configurable via `datalad.ui.interactive`. Defaults to current behavior of detection via `isatty()`. The detection result is propagated to datalad subprocesses via env var (ping #7352).
- Adds a non-dialog annex backend variant and makes the UISwitcher use it, based on `is_interactive` when asked to set the `annex` backend.

Closes #7345
Closes #7349

### DIFF
--- a/datalad/__init__.py
+++ b/datalad/__init__.py
@@ -15,6 +15,7 @@ From an interactive Python session, import `datalad.api` and inspect its
 documentation with `help`.
 """
 
+
 if not __debug__:
     raise RuntimeError(
         'DataLad cannot run in "optimized" mode, i.e. python -O')
@@ -118,11 +119,18 @@ __runtime_mode = 'library' \
     if cfg.getbool('datalad.runtime', 'librarymode', False) \
     else 'application'
 
+
 from datalad.utils import (
     get_encoding_info,
     get_envvars_info,
     getpwd,
+    # is_interactive is imported here and supposed to be imported from here, b/c
+    # anything importing and calling it (like log and ui), must do so only after
+    # datalad.cfg has been instantiated. Therefore, ensure the correct order
+    # right here.
+    is_interactive,
 )
+
 
 from .log import lgr
 
--- a/datalad/interface/common_cfg.py
+++ b/datalad/interface/common_cfg.py
@@ -30,7 +30,10 @@ from datalad.support.constraints import (
     EnsureNone,
     EnsureStr,
 )
-from datalad.utils import on_windows
+from datalad.utils import (
+    is_interactive_failsafe,
+    on_windows,
+)
 
 lgr = logging.getLogger('datalad.interface.common_cfg')
 dirs = AppDirs("datalad", "datalad.org")
@@ -639,6 +642,19 @@ _definitions = {
                     "information."}),
         'default': 10,
         'type': EnsureInt(),
+    },
+    'datalad.ui.interactive': {
+        'ui': ('question', {
+            'title': 'Datalad user interface interactivity',
+            'text': "If enabled, datalad may prompt for user input. Disable for "
+                    "non-interactive usage. If not set, datalad will attempt to "
+                    "detect whether it is running in an interactive session "
+                    "based on python's sys.__stdXXX__.isatty(). This detection "
+                    "can be wrong, though, possibly making datalad wait for "
+                    "user input, even though it is impossible to receive such "
+                    "input."}),
+        'default': is_interactive_failsafe(),
+        'type': EnsureBool(),
     },
     'datalad.save.no-message': {
         'ui': ('question', {
--- a/datalad/ui/__init__.py
+++ b/datalad/ui/__init__.py
@@ -17,17 +17,18 @@ lgr = getLogger('datalad.ui')
 
 lgr.log(5, "Starting importing ui")
 
+from datalad import is_interactive
 from .dialog import (
     ConsoleLog,
     DialogUI,
     IPythonUI,
+    NoDialogUnderAnnexUI,
     UnderAnnexUI,
     UnderTestsUI,
     SilentConsoleLog,
     QuietConsoleLog,
 )
 from ..utils import (
-    is_interactive,
     get_ipython_shell,
 )
 
@@ -36,6 +37,7 @@ KNOWN_BACKENDS = {
     'dialog': DialogUI,
     'ipython': IPythonUI,
     'annex': UnderAnnexUI,
+    'annex-no-dialog': NoDialogUnderAnnexUI,
     'tests': UnderTestsUI,
     'tests-noninteractive': QuietConsoleLog,
     'no-progress': SilentConsoleLog,
@@ -82,6 +84,8 @@ class _UI_Switcher(object):
                     backend = 'dialog'
             else:
                 backend = 'dialog' if is_interactive() else 'no-progress'
+        if not is_interactive() and backend == 'annex':
+            backend = 'annex-no-dialog'
         self._ui = KNOWN_BACKENDS[backend]()
         lgr.debug("UI set to %s", self._ui)
         self._backend = backend
--- a/datalad/ui/dialog.py
+++ b/datalad/ui/dialog.py
@@ -332,9 +332,17 @@ class IPythonUI(DialogUI):
                 *args, frontend=self._tqdm_frontend, **kwargs)
 
 
-# poor man thingie for now
-@auto_repr
-class UnderAnnexUI(DialogUI):
+class UnderAnnexMixIn(object):
+    """Mix-in for UnderAnnex UI backends
+
+    In order for the super() calls to work out, this should come first in
+    inheritance order:
+
+    class MyAnnexBackend(UnderAnnexMixIn, WhatEverElse)
+
+    This should lead to super().x() calling WhatEverElse.x()
+    """
+
     def __init__(self, specialremote=None, **kwargs):
         if 'out' not in kwargs:
             # to avoid buffering
@@ -342,7 +350,7 @@ class UnderAnnexUI(DialogUI):
             #kwargs['out'] = os.fdopen(sys.stderr.fileno(), 'w', 0)
             # but wasn't effective! sp kist straogjt for now
             kwargs['out'] = sys.stderr
-        super(UnderAnnexUI, self).__init__(**kwargs)
+        super().__init__(**kwargs)
         self.specialremote = specialremote
 
     def set_specialremote(self, specialremote):
@@ -350,11 +358,23 @@ class UnderAnnexUI(DialogUI):
 
     def get_progressbar(self, *args, **kwargs):
         if self.specialremote:
+            # Report progress via git-annex
+            from .progressbars import AnnexSpecialRemoteProgressBar
             kwargs = kwargs.copy()
-            kwargs['backend'] = 'annex-remote'
             kwargs['remote'] = self.specialremote
-        return super(UnderAnnexUI, self).get_progressbar(
-                *args, **kwargs)
+            return AnnexSpecialRemoteProgressBar(*args, **kwargs)
+        else:
+            return super().get_progressbar(*args, **kwargs)
+
+
+@auto_repr
+class UnderAnnexUI(UnderAnnexMixIn, DialogUI):
+    pass
+
+
+@auto_repr
+class NoDialogUnderAnnexUI(UnderAnnexMixIn, SilentConsoleLog):
+    pass
 
 
 @auto_repr
--- a/datalad/ui/tests/test_dialog.py
+++ b/datalad/ui/tests/test_dialog.py
@@ -35,6 +35,8 @@ from ..dialog import (
     ConsoleLog,
     DialogUI,
     IPythonUI,
+    NoDialogUnderAnnexUI,
+    SilentConsoleLog,
 )
 
 
@@ -169,12 +171,13 @@ def test_IPythonUI():
     assert_in('notebook', str(pbar._tqdm))
 
 
-def test_silent_question():
+@pytest.mark.parametrize("backend", [SilentConsoleLog, NoDialogUnderAnnexUI])
+def test_silent_question(backend):
     # SilentConsoleLog must not be asked questions.
     # If it is asked, RuntimeError would be thrown with details to help
     # troubleshooting WTF is happening
-    from ..dialog import SilentConsoleLog
-    ui = SilentConsoleLog()
+
+    ui = backend()
     with assert_raises(RuntimeError) as cme:
         ui.question("could you help me", title="Pretty please")
     assert_in('question: could you help me. Title: Pretty please.', str(cme.value))
--- a/datalad/utils.py
+++ b/datalad/utils.py
@@ -369,14 +369,36 @@ def _is_stream_tty(stream: Optional[IO]) -> bool:
 
 
 def is_interactive() -> bool:
-    """Return True if all in/outs are open and tty.
+    from datalad import cfg
+    return cfg.obtain("datalad.ui.interactive")
 
-    Note that in a somewhat abnormal case where e.g. stdin is explicitly
-    closed, and any operation on it would raise a
-    `ValueError("I/O operation on closed file")` exception, this function
-    would just return False, since the session cannot be used interactively.
+
+def is_interactive_failsafe() -> bool:
+    """Detect interactive session
+
+    This catches any exception when assessing whether standard I/O streams are a
+    terminal, in order to not crash when determining the default value of
+    `datalad.ui.interactive`. If something goes wrong, assume non-interactive
+    and proceed.
+
+    This also sets the environment variable for the `datalad.ui.interactive`
+    config, in order to have subprocesses use the outcome the detection of the
+    superprocess rather than running their own detection, which is (sort of)
+    doomed to not be correct. Note, that detection is only supposed to be
+    triggered if the config wasn't set to begin with.
     """
-    return all(_is_stream_tty(s) for s in (sys.stdin, sys.stdout, sys.stderr))
+
+    try:
+        interactive = all(_is_stream_tty(s)
+                          for s in (sys.stdin, sys.stdout, sys.stderr))
+    except Exception as e:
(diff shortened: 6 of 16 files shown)


mih (commented):
Just flew over the code, so this may be a bit too rough to be useful

datalad/__init__.py
```diff
@@ -118,12 +121,19 @@ def enable_librarymode():
...
     get_envvars_info,
     getpwd,
+    is_interactive as ut_is_interactive,
 )
 
+# TODO: Interactivity auto-false when librarymode is enabled?
+is_interactive = ut_is_interactive()
+# Propagate to subprocesses:
+os.environ['DATALAD_UI_INTERACTIVE'] = 'true' if is_interactive else 'false'
+
```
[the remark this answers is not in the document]
bpoldrack: > but datalad code in a subprocess also runs this very code itself.

Yes, but suggestion is not entirely true. A python API user can set a config override. This would be lost on a subprocess. Very much like `enable_librarymode`.
Of course, such caller can set an env var themselves, but it requires awareness of possible datalad subprocesses (not only special remotes, but `sshrun`, git-credential-help, ..). I think, that's a needless burden.

datalad/interface/common_cfg.py
```diff
@@ -640,6 +643,14 @@ def get_default_ssh():
         'default': 10,
         'type': EnsureInt(),
     },
+    'datalad.ui.interactive': {
+        'ui': ('question', {
+            'title': 'Datalad is executed interactively',
+            'text': "If enabled, datalad may prompt for user input. Disable for"
+                    "non-interactive usage."}),
+        'default': is_interactive_failsafe(),
```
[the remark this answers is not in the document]
bpoldrack: Agree.

datalad/__init__.py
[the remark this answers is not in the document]
bpoldrack: Actually, just realized, I didn't actually solve my aim here. Will not resolve yet.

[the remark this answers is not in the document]
bpoldrack: I made #7352 to address the general case instead.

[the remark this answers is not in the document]
bpoldrack: Removed propagation from here in favor of above mentioned issue.

yarikoptic:
with such extensive change touching functionality I also wonder if may be worth/safer for master?

bpoldrack:
> with such extensive change touching functionality I also wonder if may be worth/safer for master?

Main reason for `maint` to me is, that we need this released ASAP.

bpoldrack:
I think, this is ready, @yarikoptic. There's the `git-annex died of signal 11` on macos github action, but this seems unrelated.

datalad/utils.py
```diff
@@ -369,14 +369,25 @@ def _is_stream_tty(stream: Optional[IO]) -> bool:
...
+    return cfg.obtain("datalad.ui.interactive")
 
-    Note that in a somewhat abnormal case where e.g. stdin is explicitly
-    closed, and any operation on it would raise a
-    `ValueError("I/O operation on closed file")` exception, this function
-    would just return False, since the session cannot be used interactively.
-    """
-    return all(_is_stream_tty(s) for s in (sys.stdin, sys.stdout, sys.stderr))
+
+def is_interactive_failsafe() -> bool:
```
yarikoptic, line 380: please add docstring to describe why we need this failsafe?

also again -- code/logic duplication.  Please e.g. rename original implementation into `_is_interactive` or `is_interactive_noconfig` or add `check_config=True` and overload for the config default?

the point is to not duplicate logic which might need fixing in multiple spots later.  also note that we have `@never_fail`  so may be of use directly within invoking in config default?

yarikoptic (changes requested):
- would break current interactions with users while installing/getting data from some datasets which do require authentication (e.g. i tried `datalad install -g ///crcns/aa-1`).  I feel that there should be some "deeper" fix for interactivity detection or `annex-no-dialog` should be used only when explicitly requested with config setting (e.g. if explicitly asked to be non-interactive)
  - note that we have `datalad.tests.ui.backend` config to control which UI backend to use during tests but we seems to have no `datalad.ui.backend`
- for a bug fix, PR IMHO unnecessarily changes API by flipping from `is_interactive` function to a module attribute, and now confusingly imported from various levels (`datalad` and `datalad.ui`).  Was there really a need for such a change if we are aiming for fixing some issue? (made review harder)
- code duplication should be reduced IMHO

datalad/ui/__init__.py
```diff
@@ -81,7 +83,9 @@ def set_backend(self, backend):
                 else:
                     backend = 'dialog'
             else:
-                backend = 'dialog' if is_interactive() else 'no-progress'
+                backend = 'dialog' if is_interactive else 'no-progress'
+        if not is_interactive and backend == 'annex':
+            backend = 'annex-no-dialog'
```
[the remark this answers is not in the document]
bpoldrack: > unfortunately this would break entirely ability of users to authenticate when possible

It doesn't break the ability, it changes it - you'd need to set the new config (as env var ATM). Because we are unable to correctly detect. And as far as I can tell, it's simply not possible. (see #7345)
Keeping the behavior of assuming interactivity by default, is not an option IMO, though. Because in non-interactive jobs, that just means stalling with no error reporting whatsoever. Users would have no clue at all.
Whereas in your case, there should at least be a hint that authentication wasn't possible. I'd rather err on the somewhat informative side than on the one that leaves users with nothing to act upon.

We could maintain the behavior to some extend, when we address #7352. Passing down the (accidently correct) detection from the super process, would lead to the same behavior in your case. I'd still argue it's wrong, since detection outside of a special remote process is not more correct. It's just slightly rarer to be an issue. (see #7354)

datalad/__init__.py
```diff
@@ -14,6 +14,9 @@
 From an interactive Python session, import `datalad.api` and inspect its
 documentation with `help`.
 """
+from __future__ import annotations
```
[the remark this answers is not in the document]
bpoldrack: That's a left-over. Will remove.

datalad/utils.py
bpoldrack: Was unaware of never_fail indeed. Thx.

bpoldrack: Not sure, what duplication you are referring to, though.

bpoldrack: Just double-checked, but there's no duplication, I can see. `is_interactive()` now returns the config and the config gets its default from the new function, which is based on the former logic of `is_interactive`. It's a move of code, not duplication. What's the value of having that single line factored out again?

bpoldrack:
@yarikoptic 
> I feel that there should be some "deeper" fix for interactivity detection 

I spend hours on this and couldn't find one. That's why this solution.

> or annex-no-dialog should be used only when explicitly requested with config setting

As I pointed out in the other spot: That means we are defaulting to stalling of non-interactive jobs. 

Also: Does that mean, you want "regular" datalad processes to default to (broken) detection, but special remote process to default to assuming interactivity? This would mean to completely ignore, that this issue is not something special about special remotes. Any datalad process can find itself in the exact same situation as a special remote process.

The mitigation I can see, is #7352 - passing config manager's state down to subprocesses. So, if the top-level datalad process is set to default (detection) the result of that detection would be set for the special remote process as well.

datalad/ui/dialog.py
```diff
@@ -357,6 +357,32 @@ def get_progressbar(self, *args, **kwargs):
...
+        self.specialremote = specialremote
+
+    def get_progressbar(self, *args, **kwargs):
+        if self.specialremote:
+            from .progressbars import AnnexSpecialRemoteProgressBar
+            kwargs = kwargs.copy()
+            kwargs['remote'] = self.specialremote
+            return AnnexSpecialRemoteProgressBar(*args, **kwargs)
+        else:
+            return super(NoDialogUnderAnnexUI, self).get_progressbar(
```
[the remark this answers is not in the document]
bpoldrack: Ok

datalad/utils.py
yarikoptic: I am sorry -- I think I have misread the diff, indeed no duplication, sorry about that noise.

bpoldrack: np

datalad/ui/__init__.py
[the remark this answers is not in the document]
yarikoptic: > It doesn't break the ability enitrely, it changes it - you'd need to set the new config (as env var ATM).

well, *not entirely* means that it still breaks it regardless how you name it.

Not breaking *at all* would be whenever the default behavior stays as is at least for `maint`. Eg. could be done via config variable staying "auto" by default and then switch to `True` or `False` based on the `utils.is_interactive()`, while keeping using `is_interactive()` function, and allowing all CIs or whatnot which knows that there is no agent behind keyboard to say so.  That would IMHO be better since it would be explicit and useful regardless on how we decide to trick/handle interactivity.    Then in `master` we could be deliberated further or even change behavior.

> Whereas in your case, there should at least be a hint that authentication wasn't possible.

why *wasn't possible* if it was and was working just fine before this change?

datalad/ui/dialog.py
```diff
@@ -332,29 +332,50 @@ def get_progressbar(self, *args, **kwargs):
...
-            kwargs['backend'] = 'annex-remote'
             kwargs['remote'] = self.specialremote
-        return super(UnderAnnexUI, self).get_progressbar(
-                *args, **kwargs)
+            return AnnexSpecialRemoteProgressBar(*args, **kwargs)
+        else:
+            return super().get_progressbar(*args, **kwargs)
+
+
+# poor man thingie for now
```
yarikoptic, line 370: in what sense it is `poor`? looks legit to me

datalad/ui/tests/test_dialog.py
```diff
@@ -173,25 +173,26 @@ def test_silent_question():
...
-
-    # additional kwargs, no title, choices
-    with assert_raises(RuntimeError) as cme:
-        ui.question("q", choices=['secret1'])
-    assert_in('secret1', str(cme.value))
-
-
-@patch("datalad.log.is_interactive", lambda: False)
+    from ..dialog import SilentConsoleLog, NoDialogUnderAnnexUI
+    for k in [SilentConsoleLog, NoDialogUnderAnnexUI]:
```
yarikoptic, line 177: best to move that outside of the test as `@pytest.mark.parametrize`:
- minimizes `diff`
- fail in one doesn't taint the run of the other
- allows to run those two runs in parallel or selecting one particular one

datalad/ui/dialog.py
bpoldrack: Not cleaned up, yet. That's your comment on the (moved) implementation of `UnderAnnexUI` ;-)

yarikoptic: ah, LOL ;) so it is one of the skeletons ! ;)

datalad/ui/__init__.py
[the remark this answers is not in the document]
bpoldrack: > Eg. could be done via config variable staying "auto" by default and then switch to True or False based on the utils.is_interactive(), while keeping using is_interactive() function

That is what is happening. The special remote process evaluates the detection in the state of this PR. But detection (`auto`) simply doesn't work. It only happens to do the right thing, when the process in question is the top-level process. But we are querying from within the special remote process.

> why wasn't possible if it was and was working just fine before this change?

I didn't "work" before that change. It happened to do the right thing in interactive sessions and the wrong thing in non-interactive ones, b/c it always assumed to be in an interactive one.

yarikoptic:
> The mitigation I can see, is #7352 - passing config manager's state down to subprocesses. So, if the top-level datalad process is set to default (detection) the result of that detection would be set for the special remote process as well (instead of running it's own detection which will always come back `False`).

yes -- if outside process `datalad.ui.interactive` set to `"auto"` , and it determines that it is interactive (or not), setting its env var `DATALAD_UI_INTERACTIVE` so any nested process also realizes that sounds like a viable way to go forward which would satisfy both of our desires?

> As I pointed out in the other spot: That means we are defaulting to stalling of non-interactive jobs.

... until realizing that the first stalled job is needing credentials (which it prompts for) and making sure that they have them in the future run which anyways would seems to be needed to do.

Again -- the other end of spectrum this PR introduces is that there is not prompt  to the user who (may be due to incorrect assumptions and defaults) was getting a prompt and now all of a sudden would just get a crash that it cannot get to that data, and no instructions even on how to mitigate, or to mitigate by explicitly saying "be interactive" in the interactive shell.

bpoldrack:
@yarikoptic 
> yes -- if outside process datalad.ui.interactive set to "auto" , and it determines that it is interactive (or not), setting its env var DATALAD_UI_INTERACTIVE so any nested process also realizes that sounds like a viable way to go forward which would satisfy both of our desires?

Here is a test commit: https://github.com/datalad/datalad/pull/7344/commits/5417429ed9f004091e1ec794aa793b4793f6d820

> would need to set DATALAD_UI_INTERACTIVE envvar in any case, so config based options are overriden too in the children processes uniformly.

Not sure, whether I parse that right. You mean, you'd need the env var to overrule anything set in config files? Yes, that, or `-c` or `datalad.cfg.set(..., scope='override')` (With above commit would be passed down either way)

yarikoptic:
> Not sure, whether I parse that right. You mean, you'd need the env var to overrule anything set in config files? Yes, that, or `-c` or `datalad.cfg.set(..., scope='override')` (With above commit would be passed down either way)

in effect - yes, it would overrule in the children processes, but not in the parent process where it would actually obtain initial value via any available means (config, env var, overwrites) and then set explicit bool value it into `DATALAD_UI_INTERACTIVE` so all subprocesses (and this parent process later on may be) use it. So,  I do not immediately see any problem with the fact that children processes will not use value from the config value, and do not see a reason to utilize `DATALAD_CONFIG_OVERRIDES_JSON` for this specific aspect.

Note: this config is asked from the general `datalad.cfg`, i.e. not per dataset `config`.

bpoldrack:
@yarikoptic 

> Note: this config is asked from the general datalad.cfg, i.e. not per dataset config.

Yes, I added this aspect to the changelog. `Dataset.repo` evaluation from within the runner seems to much of a performance hit for this feature. I guess, the need to have something passed to a subprocess, while requiring it to be specific to a dataset, is rare - can't think of anything ATM. So, currently I think this is the best  of all the imperfect solutions.

bpoldrack:
Verdict from devcall: Rip out generic passing down of configs again. Should only apply to `datalad.ui.interactive` for now.

bpoldrack:
FTR: Remaining failures in crippledFS and macos github actions look as if https://github.com/datalad/datalad/pull/7367 wasn't merged.

Could it be that restarting github actions is running on an outdated, cached merge commit while Travis and AppVeyor build anew?

mih:
Before this is merged (in particular to runner adjustment), I recommend looking at https://github.com/datalad/datalad-next/issues/325 for an alternative -- which seems simpler, and also more effective.

yarikoptic (commented):
some comments I found pending

datalad/utils.py
```diff
@@ -369,14 +369,36 @@ def _is_stream_tty(stream: Optional[IO]) -> bool:
...
+    superprocess rather than running their own detection, which is (sort of)
+    doomed to not be correct. Note, that detection is only supposed to be
+    triggered if the config wasn't set to begin with.
     """
-    return all(_is_stream_tty(s) for s in (sys.stdin, sys.stdout, sys.stderr))
+
+    try:
+        interactive = all(_is_stream_tty(s)
+                          for s in (sys.stdin, sys.stdout, sys.stderr))
+    except Exception as e:
```
yarikoptic, line 394: I would worry about overall capturing of all exceptions here -- e.g. if detection breaks with new python or something like that -- we would miss it. Needs investigation of git history if more information is provided or just go back to no exception handling for now IMHO.

datalad/utils.py
```diff
@@ -369,14 +369,36 @@ def _is_stream_tty(stream: Optional[IO]) -> bool:
...
+        # Catch any exception and assume non-interactive mode if anything goes
+        # wrong accessing stdin/stdout/stderr.
+        # Raise log level to DEBUG in this case, though.
+        CapturedException(e, level=logging.DEBUG)
+        interactive = False
+    os.environ['DATALAD_UI_INTERACTIVE'] = str(interactive)
```
yarikoptic, line 400: I do not think it should overload if it is already defined.

datalad/utils.py
```diff
@@ -369,14 +369,36 @@ def _is_stream_tty(stream: Optional[IO]) -> bool:
 
 
 def is_interactive() -> bool:
-    """Return True if all in/outs are open and tty.
+    from datalad import cfg
+    return cfg.obtain("datalad.ui.interactive")
```
yarikoptic, line 373: what if config is within local git repo config?

shouldn't we overload that DATALAD_UI_INTERACTIVE here? and if we are to decide for "auto" to be default -- then here call is_interactive_failsafe explicitly?

yarikoptic (commented):
more comments

PR #7346, opened 2023-03-23
Remove `__init__()` from `GeneratorMixIn` and `AssemblingDecoderMixIn` to convert them to "proper" mixin-classes

christian-monch:
This PR removes the `__init__()`-methods from the classes `GeneratorMixIn` and `AssemblingDecoderMixIn` in order to convert them to proper mixin-classes.

The PR also adds `includeds`, which stands for "include dataset(s)" to the words that are ignored by codespell.

### DIFF
--- a/changelog.d/pr-7346.md
+++ b/changelog.d/pr-7346.md
@@ -1,3 +1,4 @@
 ### 🏠 Internal
 
 - Remove `__init__()` from `GeneratorMixIn` and `AssemblingDecoderMixIn` to convert them to "proper" mixin-classes.  [PR #7346](https://github.com/datalad/datalad/pull/7346) (by [@christian-monch](https://github.com/christian-monch))
+- Fix spelling errors and add `includeds` to codespell's ignore words list .  [PR #7346](https://github.com/datalad/datalad/pull/7346) (by [@christian-monch](https://github.com/christian-monch))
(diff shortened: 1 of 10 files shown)


yarikoptic:
> This PR removes the `__init__()`-methods from the classes `GeneratorMixIn` and `AssemblingDecoderMixIn` in order to convert them to proper mixin-classes.

are you trying to address this issue or what inspired this refactoring?

christian-monch:
> > This PR removes the `__init__()`-methods from the classes `GeneratorMixIn` and `AssemblingDecoderMixIn` in order to convert them to proper mixin-classes.
> 
> are you trying to address this issue or what inspired this refactoring?


Mixin-classes should not have constructors or the need to call those constructors. Without the need to call a mixin constructor, the mixin class can just be added as another superclass to the class that uese the mixin, and there is no need to change constructors in the class that uses the mixin.  That simplfies the code that uses mixins.

yarikoptic:
FWIW I asked chatgpt and it says

> is the statement correct in relation to Python "Mixin-classes should not have constructors or the need to call those constructors."?
> The statement is generally correct, but it requires some clarification.
>
> Mixin classes are designed to add functionality to other classes by providing additional methods or attributes. They are not meant to be instantiated on their own, but instead, they are intended to be subclassed by other classes.
> Because of this, mixin classes usually do not need to define their own constructors. When a class that inherits from a mixin class is instantiated, its constructor will automatically call the constructors of its parent classes, including the mixin class.
> However, if a mixin class needs to initialize some of its own attributes, it may define an `__init__` method. In this case, the mixin class should take care not to overwrite any attributes defined in the classes that inherit from it. It should also avoid calling `super().__init__()` or any other constructor, as this may interfere with the initialization of the other classes in the inheritance hierarchy.
> In summary, while mixin classes can have constructors, it is generally not necessary and can lead to issues. If a mixin class needs to initialize its own attributes, it should do so in a way that does not interfere with the initialization of other classes in the inheritance hierarchy.

and indeed avoiding `__init__` in them would make sense.

christian-monch:
This work has been rebased on main and merged into the branch `enh-generator-mix` which is the code that is to be merged in PR #7503

Issue #7351, opened 2023-03-23
non-annex superdataset causes saving in subdataset to use git add not git-annex add


yarikoptic:
I will not report amount of time it left me and @afni-rickr confused on why .nii.gz and other monsters get committed to `git` instead of git-annex, and I was surprised to realize that we have such a bug 

<details>
<summary>this script reproduces the problem</summary> 

```bash
#!/bin/bash
set -ex

cd "$(mktemp -d ${TMPDIR:-/tmp}/dl-XXXXXXX)"

# if you remove --no-annex -- all is good

datalad create --no-annex super
cd super
datalad create  -d . -c text2git subds

dd if=/dev/random of=subds/data.dat  count=1 bs=1024

cp subds/data.dat subds/data2.dat

# lets save in subds first the same content
datalad save -d subds -m 'saved one in subds directly' subds/data.dat

# now save the same content but from the super
datalad save -m "Super with data in subdataset" -r -d . -r subds

# you will see that the one saved while referring to non-annex super
# is committed directly to git because we call  "git add" instead of
# "git annex"
ls -l subds/data*.dat

```
</details>

which at the end would result in

```
+ ls -l subds/data2.dat subds/data.dat
-rw------- 1 yoh yoh 1024 Mar 23 18:51 subds/data2.dat
lrwxrwxrwx 1 yoh yoh  122 Mar 23 18:51 subds/data.dat -> .git/annex/objects/zj/54/MD5E-s1024--867e9587850130c42bdf311b8e15493d.dat/MD5E-s1024--867e9587850130c42bdf311b8e15493d.dat
```
showing that `data2.dat` is committed  (while saving from super dataset) from identical in content `data.dat`.

yarikoptic-gitmate:
Issue fixed in [`0.18.3`](https://github.com/datalad/datalad/releases/tag/0.18.3)

PR #7353, opened 2023-03-24
Fix testing assumption - do create pure GitRepo superdataset and test against it

yarikoptic:
While preparing unittest to demonstrate #7351 problem I found that the good old "parametric" test copied long time from revolution was actually not parametrized at all and thus pretty much causing running the same testing (with AnnexRepo) twice.  This PR in this shape just fixes that  and must not result in new failures.

As a result we do get differential timing for those two tests:

```
11.45s call     datalad/support/tests/test_repo_save.py::test_annexrepo_save_all
8.14s call     datalad/support/tests/test_repo_save.py::test_gitrepo_save_all
```

from original idential

```
13.86s call     datalad/support/tests/test_repo_save.py::test_gitrepo_save_all
13.83s call     datalad/support/tests/test_repo_save.py::test_annexrepo_save_all
```

;)

### DIFF
--- a/changelog.d/pr-7353.md
+++ b/changelog.d/pr-7353.md
@@ -1,0 +1,3 @@
+### 🧪 Tests
+
+- Fix testing assumption - do create pure GitRepo superdataset and test against it.  [PR #7353](https://github.com/datalad/datalad/pull/7353) (by [@yarikoptic](https://github.com/yarikoptic))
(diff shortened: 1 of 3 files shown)


bpoldrack (approved):
Oops.

Thx!

yarikoptic:
appveyor was a rare now flaky 
```
FAILED ../datalad/support/tests/test_parallel.py::test_gracefull_death - assert 20 >= 24
```

I will add changelog and merge. I don't see codecov results - but I would expect it to increase if anything

yarikoptic-gitmate:

PR #7355, opened 2023-03-24
Fix save to not force committing into git if reference dataset is pure git (not git-annex)

### LINKED ISSUE #7351: non-annex superdataset causes saving in subdataset to use git add not git-annex add

yarikoptic:
Sits on top of #7353 to reuse existing testing setup. Actual fix is few character change in `save.py` and is present in the last non-changelog commit.

Closes #7351

### DIFF
--- a/changelog.d/pr-7355.md
+++ b/changelog.d/pr-7355.md
@@ -1,0 +1,3 @@
+### 🐛 Bug Fixes
+
+- Fix save to not force committing into git if reference dataset is pure git (not git-annex).  Fixes [#7351](https://github.com/datalad/datalad/issues/7351) via [PR #7355](https://github.com/datalad/datalad/pull/7355) (by [@yarikoptic](https://github.com/yarikoptic))
--- a/datalad/core/local/save.py
+++ b/datalad/core/local/save.py
@@ -314,7 +314,7 @@ class Save(Interface):
                         # calls
                         paths=None,
                         # prevent whining of GitRepo
-                        git=True if not hasattr(ds.repo, 'uuid')
+                        git=True if not hasattr(pds_repo, 'uuid')
                         else to_git,
                         # we are supplying the full status already, do not
                         # detect anything else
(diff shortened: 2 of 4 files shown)


yarikoptic:
I would even say that we as well release after this fix since it would already be useful to some folks I know ;-)

yarikoptic:
It is as green as you can ever see a PR here, with even +1.97% boost in coverage -- what could be better? ;)

I will merge unless objections voiced on Sunday.

yarikoptic-gitmate:

PR #7362, opened 2023-03-31
BF?: add more obscure parts into filename -- [] could be used in globs

yarikoptic:
With @michael-sun we ran into some odd case with git-annex, where a file

Yarik[8910].ssh.txt  failed to be  git annex get"ed  which gave me idea to test for these extra cases in the most obscure filename case

### DIFF
--- a/datalad/tests/utils_pytest.py
+++ b/datalad/tests/utils_pytest.py
@@ -1452,7 +1452,7 @@ with_parametric_batch = pytest.mark.parametrize("batch", [False, True])
 # filesystems across different OSs.  Start with the most obscure
 OBSCURE_PREFIX = os.getenv('DATALAD_TESTS_OBSCURE_PREFIX', '')
 # Those will be tried to be added to the base name if filesystem allows
-OBSCURE_FILENAME_PARTS = [' ', '/', '|', ';', '&', '%b5', '{}', "'", '"', '<', '>']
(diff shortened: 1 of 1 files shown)


yarikoptic:
eh, there are some hits:
```
FAILED ../datalad/local/tests/test_run_procedure.py::test_quoting - ValueError: Cannot find procedure with name 'just2args'
FAILED ../datalad/support/tests/test_gitrepo.py::test_gitattributes - AssertionError: assert {'sec.key': '... 'tag': False} == {'sec.key': '...this': 'that'}
  Omitting 2 identical items, use -vv to show
  Right contains 1 more item:
  {'this': 'that'}
  Full diff:
  - {'sec.key': 'val', 'tag': False, 'this': 'that'}
  ?                                ----------------
  + {'sec.key': 'val', 'tag': False}
FAILED ../datalad/support/tests/test_sshconnector.py::test_ssh_copy - datalad.runner.exception.CommandError: CommandError: 'scp -o ControlPath=/home/travis/.cache/datalad/sockets/8ef0c2f1 'datalad-test:/tmp/datalad_temp_test_ssh_copy7ujhgeoz.c\ opy/\ \|\;\&%b5{}\'"'"'\"\<\>[][12]ΔЙקم๗あ\ .datc\ 2' /tmp/datalad_temp_test_ssh_copy7ujhgeoz' failed with exitcode 1 [err: 'protocol error: filename does not match request']
= 3 failed, 1224 passed, 59 skipped, 4 xfailed, 19 warnings in 1719.62s (0:28:39) =
```

PR #7375, opened 2023-04-26
Log warning if AnnexRepo is found to have unsupported sharedRepository config

yarikoptic:
See https://git-annex.branchable.com/todo/sharedRepository_mode_not_supported_by_git-annex/ and https://github.com/psychoinformatics-de/knowledge-base/issues/15 

now it would look like

```
❯ datalad clone ///openneuro ds --config core.sharedRepository=0600
[WARNING] AnnexRepo(/tmp/ds) found to have numeric specification '0600' of shared mode which is not supported by git-annex. See https://git-annex.branchable.com/todo/sharedRepository_mode_not_supported_by_git-annex/                                           
install(ok): /tmp/ds (dataset)
```

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -288,7 +288,12 @@ class AnnexRepo(GitRepo, RepoInterface):
                 self,
                 "datalad.repo.direct configuration instructs to use direct mode"
             )
-
+        shared = config.get("core.sharedrepository", "")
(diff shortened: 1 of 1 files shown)


joeyh:
git-annex has been updated to support them, so this pull should not be needed..

yarikoptic:
Ha, awesome, thank you @joeyh !

Issue #7380, opened 2023-05-01
boost DATALAD_INSTALLER_VERSION on appveyor!


yarikoptic:
we have been experiencing a broken .deb installation https://github.com/datalad/datalad-installer/issues/154 and recently added even a check for content-size header but "nothing helped". Looking at the recent run https://ci.appveyor.com/project/mih/datalad/builds/46925948/job/o17iv83h4laedpk6 logs with magnifying glass I saw

```
python -m pip install datalad-installer${DATALAD_INSTALLER_VERSION:-}
Collecting datalad-installer==0.2.0
  Using cached datalad_installer-0.2.0-py3-none-any.whl (19 kB)
Installing collected packages: datalad-installer
Successfully installed datalad-installer-0.2.0
```
so we are installing a fixed very old version of it!!! Currently we have 0.11.2 !  I guess it is some env var defined at appveyor level since I dont see it in codebase. Please adjust or remove it (AFAIK installer is quite stable, and we just keep getting downloads more robust etc) @mih

adswa:
I found the environment variable:
![image](https://user-images.githubusercontent.com/29738718/235590315-462ef233-8399-49e8-b1a5-8223db1e614d.png)


I removed it.

PR #7379, opened 2023-05-01
Fix failing testing on CI: Use sample S3 url DANDI archive, our copy of old .deb, specific miniconda installer for py 3.7

yarikoptic:
In this PR I decided to absorb all fixes to get our CI green again

1. ref: https://github.com/OpenNeuroOrg/openneuro/issues/2809   - prior one stopped allowing for ListObjectVersions

I have better "control" over what happens on DANDI bucket, and this URL we share with other projects for testing ROS3 driver, so used in other projects tests

This test started to fail Apr 28th, taints other tests failures, so should be merged asap

2. Use our mirror for .deb instead of snapshots (replaces #7381)
3. Use specific version of miniconda installer (avail with fresh 0.12.0 datalad-installer)

PR #7381, opened 2023-05-01
Use our mirror instead of snapshot

yarikoptic:
Although snapshots might be more "official" and thus more "reliably present" than our own server,  snapshots. has all kinds of throttling settings which delay download or even cause it to fail:
https://github.com/datalad/datalad-installer/issues/154

although datalad-installer should retry now (after it gets upgraded within appveyor setup see https://github.com/datalad/datalad/issues/7380) if size changes, I think it would still be more robust to just get it from our server.

I also made a note in
https://github.com/datalad/datalad-installer/issues/160 so may be we gain possibility to specify multiple URLs thus to robustify.


yarikoptic:
absorbing it within #7379

PR #7388, opened 2023-05-15
ENH: minor tune up of addurls to be more tollerant and "informative"

yarikoptic:
I was trying addurls on a file from https://gpt4all.io/models/models.json  . 

First

```
datalad  addurls -t json --key 'et:MD5-s{size}--{md5sum}' models.json 'https://gpt4all.io/models/{filename}' '{filename}'
```

was just crashing since not all records have some fields.  I could have excluded them using `-x` regex, but I thought it would be suboptimal since that would result in not recording some metadata. So I decided just to issue a warning and proceed adding all other fields which have metadata present.

Then, surprisingly the `--key` simply had no effect and no warning of any kind was issued.  It was since I used `size` instead of `filesize`.  DEBUG level message was likely there but IMHO it is too much of a user mistake to just silently ignore since IMHO it is most likely a misspecification of the `--key`. So I decided to issue a warning there too. And I did not really see the point for CapturedException there, so removed that part.

### DIFF
--- a/changelog.d/pr-7388.md
+++ b/changelog.d/pr-7388.md
@@ -1,0 +1,3 @@
+### 🐛 Bug Fixes
+
+- ENH: minor tune up of addurls to be more tolerant and "informative".  [PR #7388](https://github.com/datalad/datalad/pull/7388) (by [@yarikoptic](https://github.com/yarikoptic))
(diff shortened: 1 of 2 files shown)


yarikoptic:
@mih please chime in.

adswa (approved):
looks good to me

yarikoptic-gitmate:

PR #7391, opened 2023-05-24
RF+OPT: parse config dump in one go, give option for explicit --show-origin mode

yarikoptic:
Inspired by needing to review suggested optimization in the https://github.com/datalad/datalad/pull/7377 .

Unfortunately no exact benchmarking invocation was given so I am not sure how much faster my solution is over original or proposed one, but my guess is that it must beat both. @mih - could you try or share the invocation?

edit: ha -- #7377 just inspired me to look inside but really this RF is orthogonal to #7377 (concentrating on delaying asbolutification of the paths) and would compliment it and would need to be merged with it.  And most likely it is not giving any notable performance boost but rather IMHO makes code a little more straightforward (no nested loops besides first line etc).

While analyzing that code and RTFM on git-config -z

       -z, --null
           For all options that output values and/or keys, always end values with the null character (instead of a
           newline). Use newline instead as a delimiter between key and value. This allows for secure parsing of
           the output without getting confused e.g. by values that contain line breaks.

and looking at the format of printed out sample dump

	*(Pdb) print('\n-------------\n'.join(dump.split('\0')))
	file:/etc/gitconfig
	-------------
	filter.lfs.clean
	git-lfs clean -- %f
	-------------
	file:/etc/gitconfig
	-------------
	filter.lfs.smudge
	git-lfs smudge -- %f
	-------------
	file:/etc/gitconfig
	-------------
	filter.lfs.process
	git-lfs filter-process

I felt the need to provide a more straightforward implementation which does not (since there is no need) provide some "mixed" with and without origin specification output.  The only complications I encountered:

- I think it might be a bit "better" in terms of handling the "contaminant prefix" handling of which was introduced to address https://github.com/datalad/datalad/issues/5502 but who knows (at least I added more testing for its invocation)
- I had to fix up the sample test hardcoded dump to be parsed -- IMHO it incorrectly contained an entry for a config file without any entries. From trying on a simple case -- there should be no origin printed at all

	❯ git config -l --show-origin -z -f empty | wc
		  0       0       0

  but I just kept that entry and added another explicit \0 to separate in declared above in RTFM fashion of separation of entries.

- [ ] TODO: analyze the invocations and explicitly provide with_origin option - since "explicit better than implicit".
- [ ] ATM  based on `maint` just to ease benchmarking -- if decide to proceed I think it better to got to `master`


mih:
I have recorded the code snippet for benchmarking in https://github.com/datalad/datalad/pull/7377#issue-1689653638

yarikoptic:
oh, thanks for benchmarking.  so -- degress, let's then forget about this one and let's finalize that one

PR #7398, opened 2023-06-02
BF: Run an actual git invocation with interactive commit config

### LINKED ISSUE #7397: Interactive commit message entry via configuration is broken
I tested out a patch for the credential manager (https://github.com/datalad/datalad-next/pull/399), and toyed around with different configuration variables. 
I found that ``datalad.save.no-message=interactive`` is broken:
```
(icf) adina@muninn in /tmp/something on git:master
❱ datalad -c datalad.save.no-message=interactive save
add(ok): file (file)                                                            
[WARNING] Received an exception FileNotFoundError([Errno 2] No such file or directory: 'commit'). Canceling not-yet running jobs and waiting for completion of running. You can force earlier forceful exit by Ctrl-C. 
[INFO   ] Canceled 0 out of 0 jobs. 0 left running.                             
[ERROR  ] [Errno 2] No such file or directory: 'commit'                         
Total:   0%|                                 | 0.00/1.00 [00:00<?, ? datasets/s](icf) adina@muninn in /tmp/something on git:master
❱ git st                                                                    1 !
On branch master
```

The reason is: The command that gets ran is "commit --amend --edit", but as it isn't run with ``call_git``, but the GitRunner, it lacks the ``git`` command in front. I'm not sure when this broke, but the fix is simple. I will do a PR right away.

adswa:
Fixes #7397

### DIFF
--- a/datalad/support/gitrepo.py
+++ b/datalad/support/gitrepo.py
@@ -1545,7 +1545,7 @@ class GitRepo(CoreGitRepo):
 
         # handle interactive message entry by running another `git-commit`
         self._git_runner.run(
-            cmd + ['--amend', '--edit'],
+            ['git'] + cmd + ['--amend', '--edit'],
             protocol=NoCapture,
             stdin=None,
             env=env,


datalad/support/gitrepo.py
```diff
@@ -1545,7 +1545,7 @@ def commit(self, msg: Optional[str] = None,
 
         # handle interactive message entry by running another `git-commit`
         self._git_runner.run(
-            cmd + ['--amend', '--edit'],
+            ['git'] + cmd + ['--amend', '--edit'],
```
yarikoptic, line 1548: yikes -- we nohow tested this since this is solely for interactive amending and we lazied out to make a test (e.g. may be could be done via overloading `VISUAL`  to something which would "serve" the editor etc), and all tested invocations returned above.

It seems that I am the one to blame for ruining it in c73033edf675a03655dba6ef3edac86b69371943 AKA `0.18.0~2^2~49^2~3` where in https://github.com/datalad/datalad/commit/c73033edf675a03655dba6ef3edac86b69371943#diff-bfc1c47d032fc72b15b6a79db3d49e4c4334cf96318c5efe48a3403744b32fb1L1376 I stripped away `self._git_cmd_prefix` from `cmd` since I started to use `self._call_git(` instead of `_git_runner.run(`  ... So indeed it would be correct to prefix, but likely we would want 

```suggestion
            self._git_cmd_prefix + cmd + ['--amend', '--edit'],
```
"just in case"

adswa:
I'm taking @yarikoptic's thumbs-up as an approval. I've tested locally that I get an interactive commit message entry when setting the config variable on the latest state of this PR.

PR #7399, opened 2023-06-02
Honor the copy parameter _get_adjusted_env()

christian-monch:
This PR modifies `WitlessRunner._get_adjusted_env()` to honor the `copy`-parameter and perform a copy only if `copy` is `True`.
In addition, it improves the docstring of the method.

### DIFF
--- a/datalad/runner/runner.py
+++ b/datalad/runner/runner.py
@@ -63,14 +63,33 @@ class WitlessRunner(object):
     def _get_adjusted_env(self,
                           env: dict | None = None,
                           cwd: str | PathLike | None = None,
-                          copy: bool = True):
-        """Return an adjusted copy of an execution environment
+                          copy: bool = True
+                          ) -> dict | None:
+        """Return an adjusted execution environment
 
-        Or return an unaltered copy of the environment, if no adjustments
-        need to be made.
+        This method adjusts the environment provided in `env` to
+        reflect the configuration of the runner. It returns
+        an altered copy or an altered original, if `copy` is
+        `False`.
+
+        Parameters
+        ----------
+        env: dict
+          The environment that should be adjusted
+
+        cwd: str | PathLike | None (default: None)
+          If not None, the content of this variable will be
+          put into the environment variable 'PWD'.
+
+        copy: bool (default: True)
+          if True, the returned environment will be a
+          copy of `env`. Else the passed in environment
+          is modified. Note: if `env` is not `None` and
+          `cwd` is `None` and `copy` is `True`, the
+          returned dictionary is still a copy
         """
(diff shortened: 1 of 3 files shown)


datalad/runner/runner.py
```diff
@@ -63,14 +63,33 @@ def __init__(self,
...
-        Or return an unaltered copy of the environment, if no adjustments
-        need to be made.
+        This method adjusts the environment provided in `env` to
+        reflect the configuration of the runner. It returns
+        an altered copy or an altered original, if `copy` is
+        `False`.
+
+        Parameters
+        ----------
+        env: dict
```
yarikoptic, line 77: I think we should stop adding type annotations in docstrings since it duplicates type hints in the signature and can quickly (like immediately in this case) diverge: signature allows for `None`.

Filed related https://github.com/datalad/datalad/issues/7404, so may be just 

```suggestion
        env
```
? but if you decide to keep type annotations then I don't think they are correct since see other uses in the codebase and https://numpydoc.readthedocs.io/en/latest/format.html  and below AFAIK should be

```
        cwd: str or PathLike, optional
```

datalad/runner/runner.py
```diff
@@ -63,14 +63,33 @@ def __init__(self,
...
+          if True, the returned environment will be a
+          copy of `env`. Else the passed in environment
+          is modified. Note: if `env` is not `None` and
+          `cwd` is `None` and `copy` is `True`, the
+          returned dictionary is still a copy
         """
-        env = env.copy() if env else None
-        if cwd and env is not None:
+        env = env.copy() if env is not None and copy is True else env
+        if cwd is not None and env is not None:
```
yarikoptic, line 92: the unit test for the fix is so easy to make that IMHO totally worth it!

datalad/runner/runner.py
yarikoptic: and now we use those hints, so best probably to just remove type annotations in the docstrings.

yarikoptic (changes requested):
minor comments, but since the change is quite profound but also easy to test -- would be great to see a dedicated unittest (if we had it to start with - unlikely we grew such a bug)

datalad/runner/runner.py
christian-monch: Done in e15e69c8ea5fa4e90afdc2418e8e740280889328

christian-monch:
Hi @yarikoptic , thanks for the review. Yes, a test is a good ideat. I added it. Thanks for the hint.

yarikoptic-gitmate:
PR released in [`0.19.3`](https://github.com/datalad/datalad/releases/tag/0.19.3)

PR #7401, opened 2023-06-02
WiP ENH: yield "impossible" result record instead of exception if remote is not known

yarikoptic:
In this repo, where we have 'smaug' sibling only known to the top dataset

    ❯ giturl
    http://github.com/con/opfvta-replication-2023/

if we request getting data from 'smaug'

	❯ datalad get -rs smaug .
	[INFO   ] Ensuring presence of Dataset(/home/yoh/proj/CON/opfvta-replication-2023) to get /home/yoh/proj/CON/opfvta-replication-2023
	Remote 'smaug' is not available. Command failed:
	RemoteNotAvailableError: 'annex get --from=smaug -c annex.retry=3 --json --json-error-messages --json-progress'

where the problems are:

- not clear where (which submodule) the remote is not available
- we immediately fail instead of going through other submodules
- 'impossible' seems to be more appropriate than 'error'

with current patch it gets to be

    ❯ datalad get -rs smaug .
    [INFO   ] Ensuring presence of Dataset(/home/yoh/proj/CON/opfvta-replication-2023) to get /home/yoh/proj/CON/opfvta-replication-2023
    get(impossible): (dataset) [Remote 'smaug' is not available in '/home/yoh/proj/CON/opfvta-replication-2023/inputs/mouse-brain-templates']
    get(impossible): (dataset) [Remote 'smaug' is not available in '/home/yoh/proj/CON/opfvta-replication-2023/inputs/opfvta_bidsdata']
    action summary:
      get (impossible: 2)

but it feels "incorrect" in how the records is created.  The comments in the code show what I attempted to do -- the main problem is that then that datalad doesn't show anything for me, or duplicates info if I provide exception, etc. @mih please advise on how to proceed in "optimal record" way.

TODOs
- [ ] Figure out the best way 
- [ ] Provide a test

### DIFF
--- a/datalad/distribution/get.py
+++ b/datalad/distribution/get.py
@@ -42,6 +42,7 @@ from datalad.support.constraints import (
     EnsureNone,
 )
 from datalad.support.collections import ReadOnlyDict
+from datalad.support.exceptions import RemoteNotAvailableError
 from datalad.support.param import Parameter
 from datalad.support.annexrepo import AnnexRepo
 from datalad.support.gitrepo import (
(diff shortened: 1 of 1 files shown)


yarikoptic:
dang -- made off maint... will redo

PR #7405, opened 2023-06-05
Fix documentation link to python-gitlab

adswa:
The old link 404'ed.
Changelog should not be needed.


adswa:
As this is only a docstring change, I have cancelled all CI runs I had access to, apart from the doc ones.

mslw (approved):

PR #7407, opened 2023-06-06
Address gitlab API special character restrictions

### LINKED ISSUE #6681: `create-sibling-gitlab` fails with error: path 'must not start or end with a special character and must not contain consecutive special characters.'
### What is the problem?

When creating GitLab siblings recursively using `--layout=flat` or `--layout=collection`, the creation of the subdatasets' siblings fails with the following error:

```
create_sibling_gitlab(error): code/common (dataset) [Failed to create GitLab project: GitlabCreateError(400: {'path': ['must not start or end with a special character and must not contain consecutive special characters.']})]
```

It seems that GitLab 14.9 [introduced a new regex for project name validation](https://gitlab.com/gitlab-org/gitlab/-/merge_requests/80055) (only for newly created projects).

So the following project name formats used by DataLad in `create-sibling-gitlab` are invalid:
* `_repo_` (starts with special character)
* `{}--{}` (has consecutive special characters)

1. Configure a GitLab site `gitlab` in `.python-gitlab.cfg`
2. Create a dataset with nested subdatasets
3. From the superdataset directory, run `datalad create-sibling-gitlab -r --site gitlab`

### Datalad information

datalad 0.16.2

### Additional context

_No response_

Still trying it out, but very excited about it!
Also, you did a truly amazing job on the handbook 👍 
(The link to the testimonials redirects me to the main repo -- needs to be fixed?)

jsheunis:
The main change is from `_repo_` to `repo`. This PR serves as a space to brainstorm an effective solution.

Closes #6681

### DIFF
--- a/datalad/distributed/create_sibling_gitlab.py
+++ b/datalad/distributed/create_sibling_gitlab.py
@@ -407,9 +407,9 @@ def _proc_dataset(refds, ds, site, project, remotename, layout, existing,
             "Unknown site access '{}' given or configured, "
             "known ones are: {}".format(access, known_access_labels))
 
-    pathsep = ds.config.get("datalad.gitlab-default-pathseparator", None) or "-"
+    pathsep = ds.config.get("datalad.gitlab-default-pathseparator", "-")
     project_stub = \
-        ds.config.get("datalad.gitlab-default-projectname", None) or "project"
+        ds.config.get("datalad.gitlab-default-projectname", "project")
     project_var = 'datalad.gitlab-{}-project'.format(site)
     process_root = refds == ds
     if project is None:
--- a/datalad/distributed/create_sibling_gitlab.py
+++ b/datalad/distributed/create_sibling_gitlab.py
@@ -79,19 +79,22 @@ class CreateSiblingGitlab(Interface):
 
     "hierarchy"
       Each dataset is placed into its own group, and the actual GitLab
-      project for a dataset is put in a project named "_repo_" inside
-      this group. Using this layout, arbitrarily deep hierarchies of
+      project for a dataset is put in a project named "project" inside
+      this group. This project name is configurable (see Configuration).
+      Using this layout, arbitrarily deep hierarchies of
       nested datasets can be represented, while the hierarchical structure
       is reflected in the project path. This is the default layout, if
       no project path is specified.
     "flat"
       All datasets are placed in the same group. The name of a project
       is its relative path within the root dataset, with all path separator
-      characters replaced by '--'.
+      characters replaced by '-'. This path separator is configurable
+      (see Configuration).
     "collection"
-      This is a hybrid layout, where the root dataset is placed in a "_repo_"
+      This is a hybrid layout, where the root dataset is placed in a "project"
       project inside a group, and all nested subdatasets are represented
-      inside the group using a "flat" layout.
+      inside the group using a "flat" layout. The project name is configurable
+      (see Configuration).
 
     GitLab cannot host dataset content. However, in combination with
     other data sources (and siblings), publishing a dataset to GitLab can
@@ -100,7 +103,7 @@ class CreateSiblingGitlab(Interface):
 
     *Configuration*
 
-    All configuration switches and options for GitLab sibling creation can
+    Many configuration switches and options for GitLab sibling creation can
     be provided arguments to the command. However, it is also possible to
     specify a particular setup in a dataset's configuration. This is
     particularly important when managing large collections of datasets.
@@ -119,6 +122,15 @@ class CreateSiblingGitlab(Interface):
         Project location/path used for a datasets at GitLab instance
         SITENAME (see --project). Configuring this is useful for deriving
         project paths for subdatasets, relative to superdataset.
+    "datalad.gitlab-default-projectname"
+        The hierarchy and collection layouts publish (sub)datasets as projects
+        with a custom name. The default name "project" can be overridden with
+        this configuration.
+    "datalad.gitlab-default-pathseparator"
+        The flat and collection layout represent subdatasets with project names
+        that correspond to the path, with the regular path separator replaced
+        with a "-": superdataset-subdataset. This configuration can override
+        this default separator.
 
     This command can be configured with
     "datalad.create-sibling-ghlike.extra-remote-settings.NETLOC.KEY=VALUE" in
--- a/datalad/distributed/create_sibling_gitlab.py
+++ b/datalad/distributed/create_sibling_gitlab.py
@@ -104,7 +104,7 @@ class CreateSiblingGitlab(Interface):
     *Configuration*
 
     Many configuration switches and options for GitLab sibling creation can
-    be provided arguments to the command. However, it is also possible to
+    be provided as arguments to the command. However, it is also possible to
     specify a particular setup in a dataset's configuration. This is
     particularly important when managing large collections of datasets.
     Configuration options are:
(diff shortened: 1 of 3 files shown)


adswa:
The verdict from the discussion in the chat was:

* rename ``_repo_`` to ``project`` (by default, and config item to redefine it. It should be config, because it would need to be set within a dataset, to avoid users having to remember it each and every time)
* replace the hierarchy separator ``--`` with ``-`` (again with config item)

It should be configurations so that users themselves can act when GitLabs requirements change.

adswa:
Also TODO:
- [ ] adjust test (it tests for a specific ``_repo_`` name

datalad/distributed/create_sibling_gitlab.py
```diff
@@ -395,6 +407,9 @@ def _proc_dataset(refds, ds, site, project, remotename, layout, existing,
             "Unknown site access '{}' given or configured, "
             "known ones are: {}".format(access, known_access_labels))
 
+    pathsep = ds.config.get("datalad.gitlab-default-pathseparator", None) or "-"
+    project_stub = \
+        ds.config.get("datalad.gitlab-default-projectname", None) or "project"
```
adswa, line 412: what are people's feelings about dataset-level configs versus e.g., global configs? I personally tend to forget configuring this for all levels in a dataset. But we could also just add an example "for-each-dataset" call that would take care of this in the doc(string).

yarikoptic:
> * replace the hierarchy separator `--` with `-` (again with config item)

FTR: I tried to create `test--name-with--dashes` project on salsa of debian and got "Path must not start or end with a special character and must not contain consecutive special characters."

adswa:
> FTR: I tried to create test--name-with--dashes project on salsa of debian and got "Path must not start or end with a special character and must not contain consecutive special characters."

yes, this was the motivation behind not using ``--`` as a path separator anymore. But if a user has a local dataset with this pattern, it would still fail. I think we should detect this and rename it, too...

datalad/distributed/create_sibling_gitlab.py
```diff
@@ -100,7 +103,7 @@ class CreateSiblingGitlab(Interface):
 
     *Configuration*
 
-    All configuration switches and options for GitLab sibling creation can
+    Many configuration switches and options for GitLab sibling creation can
     be provided arguments to the command. However, it is also possible to
```
mih, line 107:
```suggestion
    be provided as arguments to the command. However, it is also possible to
```

datalad/distributed/create_sibling_gitlab.py
```diff
@@ -119,6 +122,15 @@ class CreateSiblingGitlab(Interface):
         Project location/path used for a datasets at GitLab instance
```
mih, line 122:
```suggestion
        Project location/path used for datasets at GitLab instance
```

datalad/distributed/create_sibling_gitlab.py
```diff
@@ -119,6 +122,15 @@ class CreateSiblingGitlab(Interface):
...
         SITENAME (see --project). Configuring this is useful for deriving
         project paths for subdatasets, relative to superdataset.
```
mih, line 124:
```suggestion
        project paths for subdatasets, relative to a superdataset.
```

datalad/distributed/create_sibling_gitlab.py
```diff
@@ -119,6 +122,15 @@ class CreateSiblingGitlab(Interface):
...
+    "datalad.gitlab-default-projectname"
+        The hierarchy and collection layouts publish (sub)datasets as projects
```
mih, line 126:
```suggestion
        The hierarchy and collection layouts publish (super)datasets as projects
```

I believe this is what the gitlab group taking the namespace slot of a (super)data contains as a project for that (super)dataset.

datalad/distributed/create_sibling_gitlab.py
```diff
@@ -119,6 +122,15 @@ class CreateSiblingGitlab(Interface):
...
+        with a custom name. The default name "project" can be overridden with
+        this configuration.
+    "datalad.gitlab-default-pathseparator"
+        The flat and collection layout represent subdatasets with project names
+        that correspond to the path, with the regular path separator replaced
```
mih, line 131:
```suggestion
        that correspond to their path within the superdataset, with the regular path separator replaced
```

datalad/distributed/create_sibling_gitlab.py
```diff
@@ -119,6 +122,15 @@ class CreateSiblingGitlab(Interface):
...
+        with a "-": superdataset-subdataset. This configuration can override
```
mih, line 132:
```suggestion
        with a "-": superdataset-subdataset. This configuration can be used to override
```

datalad/distributed/create_sibling_gitlab.py
```diff
@@ -395,6 +407,9 @@ def _proc_dataset(refds, ds, site, project, remotename, layout, existing,
             "Unknown site access '{}' given or configured, "
             "known ones are: {}".format(access, known_access_labels))
 
+    pathsep = ds.config.get("datalad.gitlab-default-pathseparator", None) or "-"
```
mih, line 410:
```suggestion
    pathsep = ds.config.get("datalad.gitlab-default-pathseparator", "-")
```

maybe simpler?

datalad/distributed/create_sibling_gitlab.py
mih, line 412:
```suggestion
    project_stub = \
        ds.config.get("datalad.gitlab-default-projectname", "project")
```

mih: I assume this is a general comment that has nothing to do with this line, right?

If so: I thinking making it work again is good enough for this PR, and for now.

datalad/distributed/create_sibling_gitlab.py
```diff
@@ -412,15 +427,15 @@ def _proc_dataset(refds, ds, site, project, remotename, layout, existing,
...
-                project = '{}/{}/_repo_'.format(ref_project, rproject)
+                project = f'{ref_project}/{rproject}/{project_stub}'
             elif layout == 'collection':
                 project = '{}/{}'.format(
                     ref_project,
-                    rproject.replace('/', '--'))
+                    rproject.replace('/', pathsep))
             else:
-                project = '{}--{}'.format(
+                project = '{}-{}'.format(
```
mih, line 436:
```suggestion
                project = '{}{pathsep}{}'.format(
```

Not a change to be applied directly, but wouldn't it make more sense to use `pathsep` here too?

mih (commented):
I think this makes all sense. I left a bunch of comments and suggestions, some even on stuff that is only in the margin of the diffs.

The PR top-comment should mention that this closes https://github.com/datalad/datalad/issues/6681 (12 months old).

datalad/distributed/create_sibling_gitlab.py
jsheunis:
```suggestion
    project_stub_cfg = "datalad.gitlab-default-projectname"
    project_stub = ds.config.get(
        project_stub_cfg,
        refds.config.get(project_stub_cfg, "project"))
```

Same reasoning as above

jsheunis (commented):
## Comments

1. Specification of the path separator via configuration `datalad.gitlab-default-pathseparator` should currently be done recursively in order to work correctly for subdatasets. For some other arguments set via config, the existing code infers the default from the top-level dataset (`refds`), and I think this should be the case for `pathsep` as well. I have made code suggestions to this effect.
2. This is more a question and explanation of my thoughts: it is not clear to me where exactly the path separator should be injected in a project's path. E.g. should it be injected twice when representing the path for a 2nd-level subdataset? And twice when representing a subdataset that's inside a subdirectory of the parent dataset? I tested out some examples below with `separator="+"` (note that the datasets all have dashes in them, just to add unnecessary complexity :D)
   - It seems like the collection path separation is done correctly (although this should actually be tested with even deeper levels)
   - It looks like the flat path separation is wrong though, shouldn't all dashes after `BOB2` be replaced with the `+`?
 
### Dataset structure

```
BOB
.
├── gitlab-subtest (dataset)
│   └── gitlab-subsub (dataset)
└── subdir (directory)
    └── gitlab-subdirdataset (dataset)
```

### Collection example - BOB1
Screenshot 2023-06-07 at 16 27 27

### Flat example - BOB2
Screenshot 2023-06-07 at 16 27 16

jsheunis:
A note in case we have not seen this error before: I tried to specify the path separator as `~` and received the following error:

```
[Failed to create GitLab project: GitlabCreateError(400: {'name': ["can contain only letters, digits, emojis, '_', '.', '+', dashes, or spaces. It must start with a letter, digit, emoji, or '_'."]})]
```

datalad/distributed/create_sibling_gitlab.py
jsheunis: I agree. This would address the question I had below in https://github.com/datalad/datalad/pull/7407#pullrequestreview-1468323877. I've tested it locally and it adds the separator in  all the places I expect it to be, i.e. to separate both directories and subdatasets from their parents.

adswa: Makes sense to me! I think a test needs adjustments, but I'll commit the suggestion and then amend the test fix

adswa:
I have applied the suggestions from the review, adjusted the path-separator override test, and squashed a few commits together. Thanks, @mih and @jsheunis!

yarikoptic-gitmate:
PR released in [`0.19.0`](https://github.com/datalad/datalad/releases/tag/0.19.0)

Issue #7409, opened 2023-06-06
Fix `hierarchy` layout in `create-sibling-gitlab`


jsheunis:
### Summary

This problem has been around and known about for a long time. See relevant issues and discussion:

- https://github.com/datalad/datalad/issues/5242
- https://github.com/datalad/datalad/issues/3467
- https://github.com/datalad/datalad/pull/3447#pullrequestreview-241875329

What it comes down to is that there seems to be a discrepancy between the implementation in the code of `create-sibling-gitlab` and what the docs describe for the `hierarchy` layout.

### Problem description

These are the docs for the layouts:

> **“hierarchy”**
> Each dataset is placed into its own group, and the actual GitLab project for a dataset is put in a project named “_repo_” inside this group. Using this layout, arbitrarily deep hierarchies of nested datasets can be represented, while the hierarchical structure is reflected in the project path. This is the default layout, if no project path is specified.
> **“flat”**
> All datasets are placed in the same group. The name of a project is its relative path within the root dataset, with all path separator characters replaced by ‘–‘.
> **“collection”**
> This is a hybrid layout, where the root dataset is placed in a “_repo_” project inside a group, and all nested subdatasets are represented inside the group using a “flat” layout.

The docs say that "Each dataset is placed into its own group, and the actual GitLab project for a dataset is put in a project named `_repo_` inside this group", but this is not how the top-level dataset is handled in the code, which is created as a gitlab project. 

For a subdataset, the process should be to create a gitlab subgroup in (inside the parent aka superdataset group) with the name of the subdataset, while the subdataset itself goes into the _repo_ project inside the subgroup. An example of such a url:
`gitlab.com/maingroup/superdataset/subdataset/_repo_`.

Currently, this recursive process fails because the program logic does not create the subgroup `subdataset` before attempting to create the `_repo_` project.

Another point to note is that one cannot create top-level groups via the latest Gitlab API. A suggestion is to tell users to create a top-level group themselves, and then pass that top-level group as the base part of the path provided in the `--project` argument to the `create-sibling-gitlab` call.

### Example

A demo of what a working process would look like:

1. User creates top-level `mydatalad` group on their gitlab acount
2. User sets project argument via config:
```
> datalad configuration set datalad.gitlab-mygitlab-project=mydatalad/midterm_hierarchy
set_configuration(ok): . [datalad.gitlab-mygitlab-project=mydatalad/midterm_hierarchy]
```
3. User created sibling recursively:
```
[...]

PR #7410, opened 2023-06-06
BF: Remove create-sibling-gitlab's hierarchy layout, make collection the default

jsheunis:
~This PR:~
- ~addresses https://github.com/datalad/datalad/issues/7409 and related issues~
- ~applies a small change to remove the discrepancy between the docs description of the hierarchy layout and the actual implementation.~
- ~adds adjustments for tests~
 
~The result is that each dataset name in a nested structure, including the top-level dataset, will map onto a gitlab group, and the associated dataset repositories will be created as a projects inside these nested groups.~

An additional requirement is that a user needs to create a top-level group via the Gitlab UI (this cannot be done via the API) and this group needs to be provided as the base of the value in the `--project` argument, e.g. `newgroup/mydatasets`.

**EDIT by adina**: This PR started out trying to fix the hierarchy layout of create-sibling-gitlab, but eventually discovered the same problems others have discovered at several different times in the past. See #7411 and #7409.
Now, this PR implements a suggestion that has been made by Kyle already shortly after the original create-sibling-gitlab feature was merged: Make ``collection`` the default layout. The need for this arose because the hierarchy layout has a number of edge cases that make it break when it operates on hierarchies of datasets (see #7411 and #7409 as well for details). Rather than keeping the hierarchy layout around in a dysfunctional state, this PR removes it entirely, and adjusts documentation and tests accordingly. I would argue that there is little use if we keep it - it has been broken since it existed for recursive operations, and in non-recursive operations, the outcome that  the ``collection`` layout  provides is identical to what  ``hierarchy`` would provide.


adswa:
I rebased your changes to sit on top of #7407

changelog.d/pr-7410.md
```diff
@@ -0,0 +1,3 @@
+### 🚀 Enhancements and New Features
+
+- BF: gitlab hierarchy fix.  [PR #7410](https://github.com/datalad/datalad/pull/7410) (by [@jsheunis](https://github.com/jsheunis))
```
adswa, line 3:
```suggestion
- BF: Make the behavior of ``create-sibling-gitlab``'s ``--layout hierarchy`` parameter match its documentation. Previously, despite the documentation saying otherwise, the toplevel dataset in a recursive ``create-sibling-gitlab`` call became a GitLab project instead of being placed in a dedicated group. Now, ``--layout hierarchy`` creates a group per dataset, and places the dataset as a project inside this group, as the docstring promises.  [PR #7410](https://github.com/datalad/datalad/pull/7410) (by [@jsheunis](https://github.com/jsheunis))
```

jsheunis (commented):
Some notes:
- Made a very minor doc edit
- I've tested it with multiple nested datasets, within subdirectories, and the collection and flat layouts seem to be working as expected
- The tests all pass locally
- There is a shortcoming with the `pathsep` implementation, I believe:
   - Specification of the path separator via configuration `datalad.gitlab-default-pathseparator` should currently be done recursively in order to work correctly for subdatasets. For some other arguments set via config, the existing code infers the default from the top-level dataset (`refds`), and I think this should be the case for `pathsep` as well (but it's not currently like that)
   - This is more a question: it is not clear to me where exactly the path separator should be injected in a project's path, and whether this should be injected. E.g. should it be injected twice when representing the path for a 2nd-level subdataset? And twice when representing a subdataset that's inside a subdirectory of the parent dataset? I tested out some examples below with `separator="+"` (note that the datasets all have dashes in them, just to add unnecessary complexity :D)
      - It seems like the collection path separation is done correctly (although this should actually be tested with even deeper levels)
      - It looks like the flat path separation is wrong though, shouldn't all dashes after `BOB2` be replaced with the `+`?
 
### Dataset structure

```
BOB
.
├── gitlab-subtest (dataset)
│   └── gitlab-subsub (dataset)
└── subdir (directory)
    └── gitlab-subdirdataset (dataset)
```

### Collection example - BOB1
Screenshot 2023-06-07 at 16 27 27

### Flat example - BOB2
Screenshot 2023-06-07 at 16 27 16

datalad/distributed/create_sibling_gitlab.py
```diff
@@ -395,32 +407,33 @@ def _proc_dataset(refds, ds, site, project, remotename, layout, existing,
             "Unknown site access '{}' given or configured, "
             "known ones are: {}".format(access, known_access_labels))
 
+    pathsep = ds.config.get("datalad.gitlab-default-pathseparator", None) or "-"
```
jsheunis, line 410:
```suggestion
    pathsep_config = "datalad.gitlab-default-pathseparator"
    pathsep = ds.config.get(pathsep_config,
                            refds.config.get(pathsep_config, "-"))
```

jsheunis: This checks the dataset configuration, then the top-level dataset config, and lastly takes the "-" as default.

jsheunis: The same probably has to be done for the `project_stub`?

datalad/distributed/create_sibling_gitlab.py
```diff
@@ -395,32 +407,33 @@ def _proc_dataset(refds, ds, site, project, remotename, layout, existing,
...
+    project_stub = \
+        ds.config.get("datalad.gitlab-default-projectname", None) or "project"
```
jsheunis, line 412:
```suggestion
    project_stub_cfg = "datalad.gitlab-default-projectname"
    project_stub = ds.config.get(
        project_stub_cfg,
        refds.config.get(project_stub_cfg, "project"))
```

adswa:
woah, I haven't seen this much green in a while!

adswa:
btw I have created a dedicated walk-through (relies on this PR) in the handbook: https://datalad-handbook--963.org.readthedocs.build/en/963/basics/101-139-hostingservices.html#creating-a-sibling-on-gitlab

yarikoptic:
`master` is now up to date with fresh maint release and is 0.18.5-69-g6d5278172 . I have merged it in, and added `release` label.  If tests are all good -- someone hit "Merge" and see what happens ;)

adswa:
I think it didn't do anything...

yarikoptic-gitmate:
PR released in [`0.19.0`](https://github.com/datalad/datalad/releases/tag/0.19.0)

PR #7412, opened 2023-06-06
Use `sphinx_autodoc_typehints`

### LINKED ISSUE #7404: Investigate/use sphinx-autodoc-typehints
Relates to #6884 .
From https://github.com/numpy/numpydoc/issues/196 it follows that https://pypi.org/project/sphinx-autodoc-typehints/ might be the one to provide type annotations into documentation from type hints so we can avoid duplicating them in docstrings, which are of numpy style in our code base.

jwodder:
Closes #7404.

### DIFF
--- a/changelog.d/pr-7412.md
+++ b/changelog.d/pr-7412.md
@@ -1,0 +1,3 @@
+### 📝 Documentation
+
+- Use `sphinx_autodoc_typehints`.  Fixes [#7404](https://github.com/datalad/datalad/issues/7404) via [PR #7412](https://github.com/datalad/datalad/pull/7412) (by [@jwodder](https://github.com/jwodder))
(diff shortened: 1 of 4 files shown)


yarikoptic:
could you give an example on how it works -- i.e. shot of documentation before/after?
does it detect/warn about "conflicts" or incongruencies (type hint says one and docstring already says smth different), or what does it do -- replaces or leaves original?

yarikoptic:
hm, RTD doesn't report on builds ... and seems to not build for PRs even though enabled. Following the docs I "resynced the hook" (under https://readthedocs.org/dashboard/datalad/integrations/92281/). Let's see if helps.

yarikoptic (approved):
I think it is a good thing.

Issue #7400, opened 2023-06-02
Drop support for Python 3.7


adswa:
The end-of-life of Python 3.7 is at the end of June, and we will drop support for it then. What needs and can be done?

- Bump used Python versions in CIs. 
    - Appveyor currently runs Ubu20P37a, Ubu20P37b on 3.7, all other builds run Python 3.8. 
    - Github crippled FS, extensions, macos, and typing workflows runs Python 3.7
 
- Fix a few issues that have been waiting for 3.8 or got easier with it:
    - make use of ``strict_timestamps``: https://github.com/datalad/datalad/issues/3753
    - https://github.com/datalad/datalad/issues/6556
    
What else?

yarikoptic:
Was done long ago. Currently

requires-python = ">= 3.10"

PR #7413, opened 2023-06-06
Boost python to 3.8 during extensions testing

yarikoptic:
Currently testing of datalad-next is failing (and has been for awhile) due to

	AttributeError: module 'gzip' has no attribute 'BadGzipFile'

and that one was added in 3.8.  I have not researched how come appveyor with 3.7 testing in datalad-next is not running into this.  As 3.7 is about to EOL, I think it is ok to start testing downstream extensions with the next 3.8.

### DIFF
--- a/.github/workflows/test_extensions.yml
+++ b/.github/workflows/test_extensions.yml
@@ -39,10 +39,10 @@ jobs:
       with:
         fetch-depth: 0
 
-    - name: Set up Python 3.7
+    - name: Set up Python 3.8
       uses: actions/setup-python@v4
       with:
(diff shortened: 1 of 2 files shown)


adswa:
see also #7400 - I think we can bump all CI's Python versions given its only 20 days until 3.7 EOL's?

adswa (approved):

yarikoptic:
all extensions are green now. Thank you @adswa for the approval.
All other CIs - let's be aligned with overall drop of 3.7 support.

yarikoptic-gitmate:

PR #7420, opened 2023-06-08
RF: Provide typing.Self verbatim for older pythons to avoid needing more recent version of typing_extensions

### LINKED ISSUE #7419: Q: can we avoid needing
In https://github.com/datalad/datalad/commit/ddffd174b6fff7bc10b02f46f4062cffe00f878c#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7R38 we had gone

from

        'typing_extensions>=3.10.0.0; python_version < "3.10"',

to

        'typing_extensions>=4.0.0; python_version < "3.11"',

for what feature?  I wonder if we could stick to 3.10... would simplify building of packages for older debian releases

yarikoptic:
Debian stable (soon oldstable) has only 3.10.0.2 ATM, and backporting shows some challenges.

Closes #7419

### DIFF
--- a/datalad/typing.py
+++ b/datalad/typing.py
@@ -8,12 +8,32 @@
 # ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
 
 import sys
+import typing
 from typing import TypeVar
 
 if sys.version_info >= (3, 11):
     from typing import Self
 else:
-    from typing_extensions import Self
+    # needs typing_extensions >= 4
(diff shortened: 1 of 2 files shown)


jwodder:
@yarikoptic Type-checking is failing, indicating this doesn't work.

yarikoptic:
indeed, entirely wrong since older `typing` has it differently... quick attempts to fixup didn't come out, so screw it -- will need to look more into backporting of typing_extensions :-/

Issue #7421, opened 2023-06-08
3 tests fail while downstream testing git-annex on conda


yarikoptic:
```
FAILED customremotes/tests/test_archives.py::test_basic_scenario - datalad.runner.exception.CommandError: CommandError: 'git -c diff.ignoreSubmodules=none annex get -c annex.retry=3 --json --json-error-messages --json-progress -c annex.dotfiles=true -- ' |;&%b5{}'"'"'"<>ΔЙקم๗あ .dbtc '' failed with exitcode 1 under /tmp/datalad_temp_tree_test_basic_scenarioj0bbjq3y [info keys: stdout_json]
> from datalad-archives...
Unable to access these remotes: datalad-archives
  Failed to fetch any archive containing MD5E-s3--202cb962ac59075b964b07152d234b70. Tried: ['MD5E-s180--857c5fa7f53dadeb0f5963bf5073e969.tar.gz']
  Failed to fetch any archive containing MD5E-s3--202cb962ac59075b964b07152d234b70. Tried: ['MD5E-s180--857c5fa7f53dadeb0f5963bf5073e969.tar.gz', 'MD5E-s180--857c5fa7f53dadeb0f5963bf5073e969.tar.gz']
  Failed to fetch any archive containing MD5E-s3--202cb962ac59075b964b07152d234b70. Tried: ['MD5E-s180--857c5fa7f53dadeb0f5963bf5073e969.tar.gz', 'MD5E-s180--857c5fa7f53dadeb0f5963bf5073e969.tar.gz'] [err: 'get: 1 failed']
FAILED customremotes/tests/test_archives.py::test_annex_get_from_subdir - datalad.runner.exception.CommandError: CommandError: 'git annex get -- ' |;&%b5{}'"'"'"<>ΔЙקم๗あ .datc '' failed with exitcode 1
FAILED support/tests/test_external_versions.py::test_system_ssh_version - assert None
= 3 failed, 1101 passed, 105 skipped, 56 deselected, 2 xfailed, 1 xpassed, 13 warnings in 3985.62s (1:06:25) =
```
for the latter I guess we just have assumed having `ssh` although not necessarily the case...

PR #7422, opened 2023-06-08
Skip test_system_ssh_version if no ssh found + split parsing into separate test

yarikoptic:
There is some auxiliary commits as well which would have more information.

Part to address 3rd failing test on conda #7421

### DIFF
--- a/datalad/support/tests/test_external_versions.py
+++ b/datalad/support/tests/test_external_versions.py
@@ -11,6 +11,7 @@ import logging
 from os import linesep
 
 from datalad import __version__
+from datalad.cmd import StdOutErrCapture, WitlessRunner
 from datalad.support.annexrepo import AnnexRepo
 from datalad.support.exceptions import (
     CommandError,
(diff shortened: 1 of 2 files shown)


datalad/support/tests/test_external_versions.py
```diff
@@ -231,9 +232,15 @@ class thing_with_list_version:
 
 
 def test_system_ssh_version():
+    try:
+        WitlessRunner().run(['ssh', '-V'], protocol=StdOutErrCapture)
+    except FileNotFoundError as exc:
+        raise SkipTest(f"no ssh binary available: {exc}")
```
jwodder, line 238: I think it would be preferrable to use [`pytest.skip()`](https://docs.pytest.org/en/7.3.x/reference/reference.html#pytest-skip) instead.

jwodder (changes requested):

yarikoptic-gitmate:

PR #7449, opened 2023-07-03
Drop python EOLed 3.7 support (in setup.py and CI)

adswa:
Python 3.7 EOL'ed mid-June. This PR
- bumps all python 3.7 CI environments to Python 3.8
- bumps all other CI environments' Python versions by one minor version

I think this needs to be done also for master, but for now I make only one PR to see what fails.


adswa:
This PR still needs an update for Travis. I have two questions there:

- I vaguely recall issues? in another datalad-extension? that talked about removing Travis CI. Are there plans to stop using Travis for this repo? 
- If we keep Travis CI, it has many Python3.7 builds to update. What I don't parse just now, though, is the following extra-arg passed to the datalad-installer: ``_DL_ANNEX_INSTALL_SCENARIO="miniconda=py37_23.1.0-1 --python-match minor --batch git-annex=8.20200309 -m conda"``. What would this change to in order to bump it to Py3.8?

adswa:
There is one travis build that fails like this:
```
$ sudo pip install --user .

WARNING: The directory '/home/travis/.cache/pip' or its parent directory is not owned or is not writable by the current user. The cache has been disabled. Check the permissions and owner of that directory. If executing pip with sudo, you should use sudo's -H flag.

ERROR: Can not perform a '--user' install. User site-packages are not visible in this virtualenv.

The command "sudo pip install --user ." exited with 1.
```
I'm not sure whether its related to the changes here or spurious, but I can't re-trigger the build. 

The appveyor failure did not show up in the previous CI run after the first commit in this PR, so I'm assuming their spurious.

yarikoptic:
> * If we keep Travis CI, it has many Python3.7 builds to update. What I don't parse just now, though, is the following extra-arg passed to the datalad-installer: `_DL_ANNEX_INSTALL_SCENARIO="miniconda=py37_23.1.0-1 --python-match minor --batch git-annex=8.20200309 -m conda"`. What would this change to in order to bump it to Py3.8?

FTR this was resolved by using `py38_23.3.1-0`.

adswa:
The one travis run still fails, and I don't get why. The failure happens in the ``script`` part of the config file which is identical across runs, so maybe its something particular with ubuntu bionic and Python 3.10 ...

yarikoptic:
well, erroring starts early with
```
$ sudo pip install --user .
WARNING: The directory '/home/travis/.cache/pip' or its parent directory is not owned or is not writable by the current user. The cache has been disabled. Check the permissions and owner of that directory. If executing pip with sudo, you should use sudo's -H flag.
ERROR: Can not perform a '--user' install. User site-packages are not visible in this virtualenv.
The command "sudo pip install --user ." exited with 1.
```
which I checked happened only in that one indeed, no other matrix runs... let's try it with `focal`

adswa:
With the bump of the misbehaving travis run to Python 3.11 a very old requested git-annex conflicts with other software versions :/

yarikoptic:
dang... I still don't know how/why it gets installed in general: https://github.com/conda-forge/datalad-feedstock/issues/109 .

.travis.yml
```diff
@@ -35,13 +35,13 @@ env:
     - DATALAD_DATASETS_TOPURL=https://datasets-tests.datalad.org
     # How/which git-annex we install.  conda's build would be the fastest, but it must not
     # get ahead in PATH to not shadow travis' python
-    - _DL_ANNEX_INSTALL_SCENARIO="miniconda=py37_23.1.0-1 --python-match minor --batch git-annex=8.20201007 -m conda"
+    - _DL_ANNEX_INSTALL_SCENARIO="miniconda=py38_23.3.1-0 --python-match minor --batch git-annex=8.20201007 -m conda"
```
adswa, line 38:
```suggestion
    - _DL_ANNEX_INSTALL_SCENARIO="miniconda=py38_23.3.1-0 --python-match minor --batch git-annex=8.20210803 -m conda"
```

yarikoptic:
note: restarted 1 travis job which hanged in conda

adswa:
This PR has been sleeping for a while, mostly because I couldn't figure out the setup-related travis failure described in the comments. But #7492 is planning to remove Travis entirely, so I feel like this could be the opportunity to also move forward with dropping Py3.7 support in this PR.

adswa:
The linting has been fixed elsewhere already, and the macos and datalad-containers failures are unrelated. The travis build is almost through the board unhappy, but my suggestion would be to proceed with #7492, then this PR, and then, eventually, #7450

mslw (approved):
- Although the Travis setup is not removed, Travis is inactive, so it should no longer be a blocker.
- Most testing changes are duplicated in #7585 where we currently explore the consequences of macOS-latest image being switched (some version bumps are done slightly differently)
- I fully support merging this PR after we have clarity in #7585 (and I volunteer to resolve conflicts)
- I would rather see this with semver-minor and done against master. Again, I'm willing to perform the ceremony and resolve conflicts.

PR #7457, opened 2023-07-17
Update funding acknowledgements

mih:


mslw (approved):
Looks good to me. I verified that the German (BMBF, DFG) and European (H2020) grant numbers / links are searchable or resolve to corresponding grants.

PR #7472, opened 2023-08-09
Allow for reckless=ephemeral clone using relative path for the original location

### LINKED ISSUE #7469: clone --reckless=ephemeral jumps into .. folder for original repo
reproducer

```bash
#!/bin/bash
#

# do not overload so we could use it outside
# export PS4='> '
# set -x

set -eu

umask 022
cd "$(mktemp -d /tmp/dl-XXXXXXX)"

datalad create original
datalad clone --reckless=ephemeral original original-clone

```

results in

```
create(ok): /tmp/dl-cw5lgKz/original (dataset)
install(error): /tmp/dl-cw5lgKz/original-clone (dataset) [/tmp/original] [/tmp/original]    
```

which, besides providing no clue to the user or what went wrong, points to the fact that it looks fo `/tmp/original` not `/tmp/dl-cw5lgKz/original-clone`. 

I guess whoever used this feature always used absolute path for the source repo, since then it seems to work:

```
+ datalad clone --reckless=ephemeral /tmp/dl-VBAcyig/original original-clone
install(ok): /tmp/dl-VBAcyig/original-clone (dataset)            
```

yarikoptic:
The last commit is actual fix + test, and prior one is just RFing test to avoid duplication and then reuse common testing logic in the fix.  That fix fixes #7469

### DIFF
--- a/changelog.d/pr-7472.md
+++ b/changelog.d/pr-7472.md
@@ -1,0 +1,3 @@
+### 🐛 Bug Fixes
+
+- Allow for reckless=ephemeral clone using relative path for the original location.  Fixes [#7469](https://github.com/datalad/datalad/issues/7469) via [PR #7472](https://github.com/datalad/datalad/pull/7472) (by [@yarikoptic](https://github.com/yarikoptic))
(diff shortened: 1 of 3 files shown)


yarikoptic:
container -- new fail unrelated , filed https://github.com/datalad/datalad-container/issues/216

travis - archives tests... unrelated, need fixing.  So overall -- ready for review/merge

yarikoptic:
ok, I have fixed travis (see #7474) - it was due to neurodebian recent changes... we are all green. Reviews are welcome but since it is fixing an issue, I would proceed with the merge and cutting a release tomorrow if no objections are voiced.

yarikoptic-gitmate:
PR released in [`0.19.3`](https://github.com/datalad/datalad/releases/tag/0.19.3)

PR #7475, opened 2023-08-11
Replace %-interpolation in 3 log msgs to passing as arguments

yarikoptic:
This would be "more correct" although, as checked by going through logging code, as safe as before.  Detected those through trying pyupgrade which converted those to .format() invocations which triggered pylint to start complaining about

     W1202: Use lazy % or % formatting in logging functions (logging-format-interpolation)

which is still there.  Some elderly issue https://github.com/pylint-dev/pylint/issues/2395
has some extended discussion on either to keep that warning default or not. I think, since it does delay possible "str"-ification of passed args, it is a good thing since ideed might have notable (if str() is taking some time for some specific types) performance benefits.  But it also points to the need of having a CI run (like we do on travis) which has verbose logging enabled to smoke test all those code paths.

Due to very limited scope of the change, I do not think it warrants changelog.

### DIFF
--- a/datalad/core/local/resulthooks.py
+++ b/datalad/core/local/resulthooks.py
@@ -51,8 +51,8 @@ def get_jsonhooks_from_config(cfg):
         )
         if not call:
             lgr.warning(
-                'Incomplete result hook configuration %s in %s' % (
-                    hook_basevar, cfg))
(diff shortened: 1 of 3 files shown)


adswa (approved):
I just found this PR and I don't have anything against it

Issue #7468, opened 2023-08-08
add --want-get to get command .... and more preferred-content thinking 


bpinsard:
### What is the problem?

It would be great to allow passing the `--want-get` option to git-annex.

More generally, I am looking into mechanisms that would allow a dataset creator to set what is the normal content dataset users should get without involving more actions from the users. For instance, I would want to expose the full dataset structure (including restricted files) but when users run `datalad get .` it would, by default, only try to get the ones that are accessible to them (and avoid the restricted one causing errors).
 
One way would be to set [groupwanted](https://git-annex.branchable.com/git-annex-groupwanted/) for a group (eg. client) on the repos. eg.
`git submodule foreach --recursive git-annex groupwanted client  "not metadata=distribution-restrictions=*"`
and then ask the users add their local clones to that group after installing:
`git submodule foreach --recursive git-annex group here client`
and then have `datalad get` pass `--want-get` to `git-annex` either as an option or maybe even by default.

I don't think there is a way in git-annex to have a default wanted or default group set  for `here`  when the repo is cloned/init. @joeyh ? 
I guess that could also be implemented in datalad as a post-clone step. 

WDYT? 

### What steps will reproduce the problem?

na

### DataLad information

latest

### Have you had any success using DataLad before?

a lot of success :D

joeyh:
There is this todo, which I have not yet decided about. If this is
useful for datalad users, it would of course be helpful in making a
decision about this kind of thing.

https://git-annex.branchable.com/todo/Setting_default_preferred_content_expressions/

-- 
see shy jo

yarikoptic:
at datalad level it smells like we might want to 
- add ability to run "configuration procedures" similarly to how we allow for `datalad create` (e.g. `datalad create -c text2git`) so users could do something like `datalad install -c sets_particular_wanted` or alike which would set any desired expression
- similarly to how `datalad push` switches to `git annex copy --auto` mode if "wanted" is set for the remote, we could possibly make `datalad get` to respect `wanted` and invoke `annex get --auto` if "wanted" is set (e.g. using aforementioned `cfg_sets_particular_wanted`)

bpinsard:
Just to be sure: for the moment, it is not possible to pass arguments to procedures when called with `datalad create`, right?

yarikoptic:
not "directly" AFAIK. You can overcome limitation by channeling them via config `datalad -c ...` or via an env variables which you would then use in your procedure.

bpinsard:
[...]

PR #7476, opened 2023-08-15
ENH add --data option to get command, matching push one

bpinsard:
Partially addresses #7468 by adding a `--data` option to `get` command with values similar to `push` command.
Passes `--want-get here` to  `git-annex get` if `auto` or `auto-if-wanted` (and preferred content is set) values are passed.

### PR checklist

- [x] Write tests.


datalad/distribution/get.py
```diff
@@ -1000,7 +1025,7 @@ def __call__(
                             continue
                         yield res
 
-        if not get_data:
+        if data=='nothing' or not get_data:
```
yarikoptic, line 1028: I feel that introduction of `data` potentially makes `get_data` obsolete and thus require replacement/deprecation cycle and that could be taken as a good thing since `get_data` flag semantic is already shown to be "inconsistent" (`get_data` in python but `--no-data` in CLI).

But `-n` (AKA `--no-data`) is used heavily in handbook etc, and is a convenient shortcut to signal `--data=nothing`.

So, I think that is how `-n|--no-data` should be promoted - that it is a shortcut for `data="nothing"` and there should be explicit statement that `data` option is ignored if `--no-data` is provided.

But also let's wait for others to chime in

adswa:
I have placed this PR on the agenda of the next dev call for October 24th, hopefully we can revive the discussion. :)

datalad/distribution/get.py
```diff
@@ -855,6 +867,18 @@ class Get(Interface):
...
+            doc="""what to do with (annex'ed) data. 'anything' would cause
+            fetching of all annexed content, 'nothing' would avoid call to
+            `git annex get` altogether. 'auto' would use 'git annex get' with
+            '--auto' thus fetching only data which would satisfy "wanted"
+            or "numcopies" settings for the local copy (thus "nothing" otherwise).
+            'auto-if-wanted' would enable '--auto' mode only if there is a
+            "wanted" setting for the remote, and fetching 'anything' otherwise.
+            """,
+            constraints=EnsureChoice(
+                'anything', 'nothing', 'auto', 'auto-if-wanted')),
```
yarikoptic, line 892: wanna get back to this one ? please review/adjust for defining `_cfg` like we just did in 
 
- #7477 

or let me know if I should . There are also conflicts emerged (didn't look)

yarikoptic: while at it -- worth review/analysis of `update` since deals with data on either it has ability to do the `auto` (I don't remember... good job for claude ;) )

bpinsard: Sure let's finish that PR! Not sure to understand what you mean for `update`, are you talking about controlling `--reobtain-data`  behavior with a `--data` option ?

yarikoptic: YES! likely should be a separate PR.  It indeed has only g'old `--reobtain-data` and nothing like fuller fledged `--data` here. I guess it could get expanded with `"reobtain"` option which I guess more specific to `update` and default to 'auto-if-wanted'

yarikoptic-gitmate:

PR #7477, opened 2023-08-16
ENH: add --cfg-proc to install, clone, and get

bpinsard:
Adds the `--cfg-proc` option to install command to allow running commands on dataset post-installation. see #7468 
It runs procedure after recursive install.

It does not call the procedure recursively for now (recursion can be coded in the procedure itself).
Making it recursive would require way more changes to `get` and other parts of the code. 

### PR checklist

- [x] Write tests.
- [x] fix tests
- [x] refactor code for harmonious helper name/interface
- [x] @yarikoptic to add changelog entry later on

### DIFF
--- a/datalad/core/distributed/clone.py
+++ b/datalad/core/distributed/clone.py
@@ -217,6 +217,15 @@ class Clone(Interface):
             because both a regular branch and the git-annex branch are
             required. Note that a version in a RIA URL takes precedence over
             '--branch'."""),
+        cfg_proc=Parameter(
+            args=("-c", "--cfg-proc"),
+            metavar="PROC",
+            action='append',
+            doc="""Run cfg_PROC procedure(s) (can be specified multiple times)
+            on the installed dataset. Use
+            [PY: `run_procedure(discover=True)` PY][CMD: run-procedure --discover CMD]
+            to get a list of available procedures, such as cfg_text2git.
+            """),
         description=location_description,
         reckless=reckless_opt,
         message=save_message_opt,
@@ -234,6 +243,7 @@ class Clone(Interface):
             description=None,
             reckless=None,
             message=None,
+            cfg_proc=None,
         ):
         # did we explicitly get a dataset to install into?
         # if we got a dataset, path will be resolved against it.
@@ -327,6 +337,7 @@ class Clone(Interface):
                 result_props,
                 cfg=None if ds is None else ds.config,
                 clone_opts=git_clone_opts,
+                cfg_proc=cfg_proc,
                 ):
             if r['status'] in ['error', 'impossible']:
                 clone_failure = True
@@ -412,7 +423,8 @@ def clone_dataset(
         result_props=None,
         cfg=None,
         checkout_gitsha=None,
-        clone_opts=None):
+        clone_opts=None,
+        cfg_proc=None):
     """Internal helper to perform cloning without sanity checks (assumed done)
 
     This helper does not handle any saving of subdataset modification or adding
@@ -445,6 +457,8 @@ def clone_dataset(
       Options passed to git-clone. Note that for RIA URLs, the version is
       translated to a --branch argument, and that will take precedence over a
       --branch argument included in this value.
+    cfg_proc : list of str, optional
+      List of procedures to be run on the cloned datasets.
 
     Yields
     ------
@@ -542,6 +556,15 @@ def clone_dataset(
         )
         rmtree(destds.path, children_only=dest_path_existed)
         return
+
+    from datalad.local.run_procedure import _get_proc_configs
+    cfg_proc_specs = _get_proc_configs(cfg_proc, destds) if cfg_proc else []
+    for cfg_proc_spec in cfg_proc_specs:
+        yield from destds.run_procedure(
+            cfg_proc_spec,
+            result_renderer='disabled',
+            return_type='generator',
+        )
 
     # yield successful clone of the base dataset now, as any possible
     # subdataset clone down below will not alter the Git-state of the
--- a/datalad/core/distributed/clone.py
+++ b/datalad/core/distributed/clone.py
@@ -557,6 +557,7 @@ def clone_dataset(
         rmtree(destds.path, children_only=dest_path_existed)
         return
 
+    # import in function to circumvent circular import issue
     from datalad.local.run_procedure import _get_proc_configs
     cfg_proc_specs = _get_proc_configs(cfg_proc, destds) if cfg_proc else []
     for cfg_proc_spec in cfg_proc_specs:
--- a/datalad/core/distributed/clone.py
+++ b/datalad/core/distributed/clone.py
@@ -28,6 +28,7 @@ from datalad.interface.base import (
     eval_results,
 )
 from datalad.interface.common_opts import (
+    cfg_proc_opt,
     location_description,
     reckless_opt,
     save_message_opt,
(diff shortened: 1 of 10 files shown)


yarikoptic:
> It does not call the procedure recursively for now (recursion can be coded in the procedure itself).

I do think that procedure better to operate non-recursively and then indeed it be `get` to actually gain that option first -- the "older style" `install` is pretty much a wrapper around `get` and some folks are just using `datalad clone` and `datalad get` tandem instead of `install`.

yarikoptic:
Sorry @bpinsard we didn't follow up on this one!
Could you please ideally rebase and force-push so we have a fresh runs of CIs -- and let's aim to see it merged/released.

Other @datalad/developers -- any feedback on this one? looks good and usefull to me.

yarikoptic:
changelog addition commit got lost, re-added the label. but CI failed to add that commit

```
remote: Permission to courtois-neuromod/datalad.git denied to yarikoptic-gitmate.
fatal: unable to access 'https://github.com/courtois-neuromod/datalad/': The requested URL returned error: 403

```

@bpinsard please permit other maintainers to contribute (I think it is the case why it failed) to this PR (checkmark on the side to the right above IIRC). Or produce it manually using `scriv` tool.

bpinsard:
Not sure what is wrong in the config, I am pretty sure that I simply forked the repo, what access should I grant and to who and where?

yarikoptic:
on the right under Paricipants - do you see smth like

image

?  didn't look where generally that is configured in profile

yarikoptic:
tests fail BTW and need fixing, a sample from appveyor:

```
=========================================================================================================== short test summary info ============================================================================================================
FAILED ../datalad/distribution/tests/test_install.py::test_install_withcfg - TypeError: __call__() takes from 0 to 1 positional arguments but 2 positional arguments (and 1 keyword-only argument) were given
FAILED ../datalad/distribution/tests/test_install.py::test_recursive_install_withcfg - TypeError: __call__() takes from 0 to 1 positional arguments but 2 positional arguments (and 2 keyword-only arguments) were given
```

datalad/core/local/create.py
```diff
@@ -572,3 +554,24 @@ def _setup_annex_repo(path, initopts=None, fake_dates=False,
             'type': 'file',
             'state': 'untracked'}
     return tbrepo, add_to_git
+
+def _procedures_exists(ds, cfg_proc):
+    # Check if specified cfg_proc(s) can be discovered, storing
+    # the results so they can be used when the time comes to run
+    # the procedure. If a procedure cannot be found, raise an
+    # error to prevent creating the dataset.
```
yarikoptic, line 562: I think this function should get into datalad/local/run_procedure.py to be along with other procedure support functions:

```shell
❯ git grep 'def _*get.*proc'
datalad/local/run_procedure.py:def _get_proc_config(name, ds=None):
datalad/local/run_procedure.py:def _get_procedure_implementation(name='*', ds=None):
```
and should be named more inline with other function names (although yeah -- already inconsistent above IMHO) and a little more descriptive :

not also I swapped the order of args to be more inline with other 2 functions and renamed arg so body should change too

```suggestion
def _get_proc_configs(names, ds):
    """Get specifications for discovered procedures.
    
     If not a single procedure found, raises `ValueError`
     """
```

bpinsard:
> on the right under Paricipants - do you see smth like
> 
> ? didn't look where generally that is configured in profile
not it isnt' there.
image

yarikoptic:
interesting -- you are special @bpinsard ! ;-)

ok, nevermind that -- please address my request for harmonization of naming of the helper and address tests fails and I will take care about adding changelog entry later.

bpinsard:
https://github.com/orgs/community/discussions/5634 because the fork is in an org not an individual...

yarikoptic:
tests aren't happy
```
FAILED ../datalad/distribution/tests/test_install.py::test_install_withcfg - TypeError: __call__() takes from 0 to 1 positional arguments but 2 positional arguments (and 1 keyword-only argument) were given
FAILED ../datalad/distribution/tests/test_install.py::test_recursive_install_withcfg - TypeError: __call__() takes from 0 to 1 positional arguments but 2 positional arguments (and 2 keyword-only arguments) were given
```
and may be more -- didn't look in detail.

bpinsard:
Inconsistent failure of the `datalad/distribution/tests/test_install.py::test_invalid_args` depending on python version, or env or else
https://github.com/datalad/datalad/blob/d4ce9ce377bde2d59fc3a550938b51931dbfbb7b/datalad/distribution/tests/test_install.py#L143 

In my local env (platform linux -- Python 3.10.12, pytest-8.1.1)  it raises IncompleteResultsError rather than TypeError. 
any clue @yarikoptic ?

datalad/core/local/create.py
```diff
@@ -341,25 +336,8 @@ def __call__(
...
-            )
-            for cfg_proc_ in cfg_proc:
-                for discovered_proc in discovered_procs:
-                    if discovered_proc['procedure_name'] == 'cfg_' + cfg_proc_:
-                        cfg_proc_specs.append(discovered_proc)
-                        break
-                else:
-                    raise ValueError("Cannot find procedure with name "
-                                     "'%s'" % cfg_proc_)
+        from datalad.local.run_procedure import _get_proc_configs
```
yarikoptic: also this -- move to the top of the file unless required to be here

datalad/distribution/tests/test_install.py
```diff
@@ -1057,3 +1057,22 @@ def test_install_recursive_github(path=None):
...
+
+@with_tempfile(mkdir=True)
+@with_tempfile(mkdir=True)
+def test_recursive_install_withcfg(src_path=None, dest=None):
+    src = create(src_path)
+    src.create('subds')
+    ds = install(path=dest, source=src_path, recursive=True, cfg_proc=['yoda'])
+
+    assert (ds.pathobj / 'README.md').exists()
+    assert (ds.pathobj / 'subds' / 'README.md').exists()
```
yarikoptic: let's fold this into prior test -- our CI already runs LONG time . Can either unlink first install under dest or use two different subfolders

yarikoptic (changes requested):
Indeed might be handy, conflicts accumulated and I left a few comments

datalad/distribution/tests/test_install.py
yarikoptic: eh, I need to find some quality time to fix up our datalad-installer ... 
- https://github.com/datalad/datalad-installer/pull/219 and others... 
but that is for overall -- kinda independent of the question here. but ok - let's keep separate

yarikoptic:
you didn't allow pushes from collaborators or just invite me :-/

```shell
❯ git push
ERROR: Permission to courtois-neuromod/datalad.git denied to yarikoptic.

and the repository exists.
```

bpinsard:
> you didn't allow pushes from collaborators or just invite me :-/
> 
> ```shell
> ❯ git push
> ERROR: Permission to courtois-neuromod/datalad.git denied to yarikoptic.
> fatal: Could not read from remote repository.
> 
> Please make sure you have the correct access rights
> and the repository exists.
> ```

again the org vs user mechanisms, added write permissions to our fork.

yarikoptic:
coolio -- force pushed now!

datalad/core/distributed/clone.py
```diff
@@ -217,6 +217,15 @@ class Clone(Interface):
             because both a regular branch and the git-annex branch are
             required. Note that a version in a RIA URL takes precedence over
             '--branch'."""),
+        cfg_proc=Parameter(
```
yarikoptic: uff -- just now spotted duplicated copies of the `cfg_proc` -- as all other `_opt` reused we should define it centrally and reuse -- no point breeding duplications to maintain...

yarikoptic: pushed

yarikoptic:
still green, let's proceed! 2.5 years in making ! a record ;)

Issue #7478, opened 2023-08-18
fails to install submodule when relative path is given


bpinsard:
### What is the problem?

Got a weird error when trying to recursively install a dataset from a local relative path and in fact this is exactly the same problem as for `~` described in [#2232](https://github.com/datalad/datalad/pull/2232)

### What steps will reproduce the problem?

```
datalad create test_src
datalad create -d test_src test_src/subds
datalad install -s test_src/ -r test_dst

install(ok): /home/basile/data/tests/test_dst (dataset)                                                                                 
[INFO   ] Ensuring presence of Dataset(/home/basile/data/tests/test_dst) to get /home/basile/data/tests/test_dst 
install(error): /home/basile/data/tests/test_dst/subds (dataset) [Failed to clone from any candidate source URL. Encountered errors per each url were:                                                                                                                          
- ../test_src/subds
  CommandError: 'git -c diff.ignoreSubmodules=none -c core.quotepath=false clone --progress ../test_src/subds /home/basile/data/tests/test_dst/subds' failed with exitcode 128 [err: 'fatal: repository '../test_src/subds' does not exist']]
action summary:                                                                                                                         
  install (error: 1, ok: 1)
```

### DataLad information

datalad 0.19.3+5.geeb731380

adswa:
I looked into this again in the context of #7510, and sharing an interim observation before I need to switch to a meeting.

In the debugger, I'm seeing the following:

```
# superdataset is fine
(Pdb) n
install(ok): /tmp/test_dst (dataset)
[INFO   ] Ensuring presence of Dataset(/tmp/test_dst) to get /tmp/test_dst 
Installing: 0.00 datasets [00:00, ? datasets/s]> 

# subdataset passes down wrong url, need to dig here more
/home/adina/env/handbook/lib/python3.11/site-packages/datalad/core/distributed/clone_utils.py(329)_get_installationpath_from_url()
-> ri = RI(url)
(Pdb) p url
'../test_src/subds'
(Pdb) 

[...]

> /home/adina/env/handbook/lib/python3.11/site-packages/datalad/core/distributed/clone_utils.py(438)decode_source_spec()
-> return props
(Pdb) p props
{'source': '../test_src/subds', 'version': None, 'type': 'giturl', 'giturl': '../test_src/subds', 'default_destpath': 'subds'}

[...]

(Pdb) p candidate_sources
[...]

PR #7479, opened 2023-08-19
BF: failure to install recursively from a relative path

bpinsard:
Working on a fix for

- #7478 
 
that looks like a more generic case of 

- #2232 . 

TODOs

- [x] writing a test that reproduces the bug
- [ ] write a fix

### DIFF
--- a/datalad/distribution/tests/test_get.py
+++ b/datalad/distribution/tests/test_get.py
@@ -876,3 +876,14 @@ def test_get_non_existing(origin_path=None, clone_path=None):
 
     assert_result_count(res, 1, status="impossible",
                         message="path does not exist")
+
+
+@with_tempfile(mkdir=True)
+def test_get_recursive_relpath(local_path=None):
+    with chpwd(local_path):
+        ds = create(opj(local_path, 'src'))
+        ds.create('sub_ds')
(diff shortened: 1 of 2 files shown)


bpinsard:
Not sure about the right approach to fixing that issue:
- resolving source to absolute path for the root dataset and then candidate will be absolute too, simpler option
- trying to preserve a relative path by running `git-clone` wile in the appropriate pwd for it to resolve.

yarikoptic:
Thank you @bpinsard for looking into this!!!  
The former (keep relative) sounds like a more proper fix to do but might overall be much more involved. Looking at a sample run

```
(Pdb) p dest_path
'/home/yoh/.tmp/datalad_temp_test_install_recursive_relpathkh6xkzg6/dest/sub_ds'
(Pdb) p clone_urls_
['../src/sub_ds']
(Pdb) getpwd()
'/home/yoh/.tmp/datalad_temp_test_install_recursive_relpathkh6xkzg6'
```

suggests looking into why we are manually creating those relative leading `../` paths and now just keeping as from pwd... might look more into it later unless you beat me  to it

yarikoptic:
I will move it to draft, please make `Ready for Review` whenever ready

PR #7488, opened 2023-09-19
RF(TST): use setup_method and teardown_method in TestAddArchiveOptions

yarikoptic:
Prior one is coming from "nose support" layer which is about to be deprecated. So we were getting flood of warnings like

   local/tests/test_add_archive_content.py::TestAddArchiveOptions::test_override_existing_under_git
  /Users/runner/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/site-packages/_pytest/fixtures.py:911: PytestRemovedIn8Warning: Support for nose tests is deprecated and will be removed in a future release.
  local/tests/test_add_archive_content.py::TestAddArchiveOptions::test_override_existing_under_git is using nose-specific method: `teardown(self)`
  To remove this warning, rename it to `teardown_method(self)`
  See docs: https://docs.pytest.org/en/stable/deprecations.html#support-for-tests-written-for-nose
    next(it)

for a while but didn't spot them.  I think that is the only place in core.

### DIFF
--- a/changelog.d/pr-7488.md
+++ b/changelog.d/pr-7488.md
@@ -1,0 +1,3 @@
+### 🧪 Tests
+
+- RF(TST): use setup_method and teardown_method in TestAddArchiveOptions.  [PR #7488](https://github.com/datalad/datalad/pull/7488) (by [@yarikoptic](https://github.com/yarikoptic))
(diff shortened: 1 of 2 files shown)


yarikoptic:
benchmarks are generally failing, yet to be investigated. no brainer so merging

yarikoptic-gitmate:

PR #7492, opened 2023-09-22
RF: remove travis CI setup + badge from README.md

yarikoptic:
We are using paid travis CI, and funds for that line are about to end. Moreover @mih is also paying for appveyor.  So I think it would be reasonable to remove travis. The tricky part would be to get coverage parity since travis has AFAIK more extended battery of matrix runs to excercise different deployment scenarios ATM.

TODOs
- [ ] see what we can do to increase coverage back to some only minimal difference


mih (approved):
I support this move.

Given the runtime of the test battery, I suspect that the goal of reaching coverage parity is something that will require a long time.

adswa:
I support this, too. Do I see correctly that the coverage loss is already minimal (-0.28%)? From a cursory read, I guess we mostly loose the breadth of versions, e.g. Python 3.11 or minimal git-annex version

yarikoptic:
Loss in code % is indeed small, but as you can assess - it would loose coverage of "interesting usecases" (more logging, specific git-annex setups, etc).  I would really like to add at least 1 or 2 matrix runs to .appveyor to mimic at least some of those so we do not completely loose testing those code paths.

yarikoptic:
we are good already
```
❯ grep -i travis README.md
❯ git describe
1.1.0-8-gb885a669b
```

PR #7500, opened 2023-10-04
improve threaded-runner documentation

### LINKED ISSUE #7498: Design document on `ThreadedRunner` outdated
https://docs.datalad.org/en/stable/design/threaded_runner.html

I did not go through the entire text, but it claims that we use `asyncio.protocols.SubprocessProtocol`, which is no longer the case (since datalad v0.17, 2022-07-07).

Ping @christian-monch

christian-monch:
Fixes #7498 

This commit clarifies the description of the protocols that are used by the ThreadedRunner-class. It states prominently that we use the same interface as the asyncio-protocol classes, but that we do not make use of any asyncio functionality.

### DIFF
--- a/docs/source/design/threaded_runner.rst
+++ b/docs/source/design/threaded_runner.rst
@@ -85,7 +85,7 @@ number specifying the desired timeout in seconds. If no data is received from
 ``stdin``, or ``stderr`` (if those are supposed to be captured), the method
 ``WitlessProtocol.timeout(fd)`` is called with ``fd`` set to the respective
 file number, e.g. 1, or 2. If ``WitlessProtocol.timeout(fd)`` returns ``True``,
-the file descriptor will be closed and the associated threads will exit.
+only the corresponding file descriptor will be closed and the associated threads will exit.
 
 The method ``WitlessProtocol.timeout(fd)`` is also called if stdout, stderr
 and stdin are closed and the process does not exit within the given interval.
--- a/docs/source/design/threaded_runner.rst
+++ b/docs/source/design/threaded_runner.rst
@@ -56,25 +56,41 @@ The main thread waits on the ``output_queue`` for data or signals and handles th
 Protocols
 =========
 
-Due to its history the runner implementation uses the interface defined in ``SubprocessProtocol`` (asyncio.protocols.SubprocessProtocol) (although the sub process protocol interface is defined in the asyncio libraries, the current thread-runner implementation does not make use of ``async``).
+Due to its history datalad uses the protocol defined in ``asyncio.protocols.SubprocessProtocol`` and in ``asyncio.protocols.BaseProtocol``.
+To keep compatibility with the code base, the threaded-runner implementation uses
+the same interface. Please note, although we use the same interface and
+although the interface is defined in the asyncio
+libraries, the threaded-runner implementation does not make any use of ``asyncio``.
+The description of the interface nevertheless applies in the context of the threaded-runner.
+The following methods of the ``SubprocessProtocol`` are supported.
 
     - ``SubprocessProtocol.pipe_data_received(fd, data)``
     - ``SubprocessProtocol.pipe_connection_lost(fd, exc)``
     - ``SubprocessProtocol.process_exited()``
 
-In addition the methods of ``BaseProtocol`` are called, i.e.:
+In addition the following methods of ``BaseProtocol`` are supported:
 
     - ``BaseProtocol.connection_made(transport)``
     - ``BaseProtocol.connection_lost(exc)``
 
 
-The datalad-provided protocol ``WitlessProtocol`` provides an additional callback:
+The datalad-provided protocol ``datalad.runners.protocol.WitlessProtocol``
+provides an additional callback:
 
     - ``WitlessProtocol.timeout(fd)``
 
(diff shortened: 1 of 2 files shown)


docs/source/design/threaded_runner.rst
```diff
@@ -56,25 +56,41 @@ The main thread waits on the ``output_queue`` for data or signals and handles th
...
     - ``WitlessProtocol.timeout(fd)``
 
-The method ``timeout()`` will be called when the parameter ``timeout`` in ``WitlessRunner.run``, ``ThreadedRunner.run``, or ``run_command`` is set to a number specifying the desired timeout in seconds. If no data is received from ``stdin``, or ``stderr`` (if those are supposed to be captured), the method ``WitlessProtocol.timeout(fd)`` is called with ``fd`` set to the respective file number, e.g. 1, or 2. If ``WitlessProtocol.timeout(fd)`` returns ``True``, the file descriptor will be closed and the associated threads will exit.
+The method ``timeout()`` will be called when the parameter ``timeout`` in
+``WitlessRunner.run``, ``ThreadedRunner.run``, or ``run_command`` is set to a
+number specifying the desired timeout in seconds. If no data is received from
+``stdin``, or ``stderr`` (if those are supposed to be captured), the method
+``WitlessProtocol.timeout(fd)`` is called with ``fd`` set to the respective
+file number, e.g. 1, or 2. If ``WitlessProtocol.timeout(fd)`` returns ``True``,
+the file descriptor will be closed and the associated threads will exit.
```
yarikoptic, line 88:
```suggestion
only the corresponding file descriptor will be closed and the associated threads will exit.
```

so it is responsibility of what code to close the other fd?  e.g. if timeout happens on fd 2 but not fd 1, whenever both are captured, what would happen?

docs/source/design/threaded_runner.rst
```diff
@@ -56,25 +56,41 @@ The main thread waits on the ``output_queue`` for data or signals and handles th
...
 =========
 
-Due to its history the runner implementation uses the interface defined in ``SubprocessProtocol`` (asyncio.protocols.SubprocessProtocol) (although the sub process protocol interface is defined in the asyncio libraries, the current thread-runner implementation does not make use of ``async``).
+Due to its history datalad uses the protocol defined in ``asyncio.protocols.SubprocessProtocol`` and in ``asyncio.protocols.BaseProtocol``.
+To keep compatibility with the code base, the threaded-runner implementation uses
+the same interface. Please note, although we use the same interface and
+although the interface is defined in the asyncio
+libraries, the threaded-runner implementation does not make any use of ``asyncio``.
+The description of the interface nevertheless applies in the context of the threaded-runner.
+The following methods of the ``SubprocessProtocol`` are supported.
```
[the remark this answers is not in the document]
yarikoptic: actually -- because original sentence was a line -- best to adhere to that rule to start.

docs/source/design/threaded_runner.rst
```diff
@@ -56,25 +56,41 @@ The main thread waits on the ``output_queue`` for data or signals and handles th
...
+``stdin``, or ``stderr`` (if those are supposed to be captured), the method
+``WitlessProtocol.timeout(fd)`` is called with ``fd`` set to the respective
+file number, e.g. 1, or 2. If ``WitlessProtocol.timeout(fd)`` returns ``True``,
+the file descriptor will be closed and the associated threads will exit.
 
-The method ``WitlessProtocol.timeout(fd)`` is also called if stdout, stderr and stdin are closed and the process does not exit within the given interval. In this case ``fd`` is set to ``None``. If ``WitlessProtocol.timeout(fd)`` returns ``True`` the process is terminated.
+The method ``WitlessProtocol.timeout(fd)`` is also called if stdout, stderr
+and stdin are closed and the process does not exit within the given interval.
+In this case ``fd`` is set to ``None``. If ``WitlessProtocol.timeout(fd)``
+returns ``True`` the process is terminated.
```
yarikoptic, line 93: another place where a sentence per line could have been useful . Although original one was combination of two. Would help in future

docs/source/design/threaded_runner.rst
christian-monch: > if timeout happens on fd 2 but not fd 1, whenever both are captured, what would happen?

fd 1 would be left untouched. It will be closed latest if the runner exits. But it might be closed earlier by the user or if a timeout appears on fd 1 and the timeout handler returns `True` for fd 1

docs/source/design/threaded_runner.rst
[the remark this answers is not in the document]
christian-monch: Reformated the patch to keep paragraphs mostly identical.

mih (approved):
Thanks for the update.

FTR: I think it would improve the documentation, if all inconsequential statements are removed. It does not matter if the protocol is used "due to its history". If it would matter, we still don't know what history that would be. The key fact is that this protocol is partially API compatible with `asyncio`. FWIW, it could just say "the protcol is "inspired" by asyncio. Which would pretty much convey the same information, and be more to the point.

That being said: Any day where this stay unmerged is a lost day.

adswa:
The errors are unrelated and this PR is approved, thus I'm merging.

yarikoptic-gitmate:

PR #7505, opened 2023-10-09
Inform asv that there should be no warmup runs for time_remove benchmark

yarikoptic:
Return the `time_remove` benchmark back.

Thanks to @trexfeathers for the RTFM pointers in
https://github.com/airspeed-velocity/asv/issues/1347#issuecomment-1752652351


adswa (approved):
the benchmarks are green again! :)

yarikoptic-gitmate:

PR #7509, opened 2023-10-13
Enhance `run()` placeholder substitutions to honor configuration defaults

mih:
Previously, `run()` would not recognize configuration defaults for placeholder substitution. This means that any placeholders globally declared in `datalad.interface.common_cfg`, or via `register_config()` in DataLad extensions would not be effective.

This changeset makes run's `format_command()` helper include such defaults explicitly, and thereby enable the global declaration of substitution defaults.

Moreoever a `{python}` placeholder is now defined via this mechanism, and points to the value of `sys.executable` by default. This particular placeholder was found to be valueable for improving the portability of run-recording across (specific) Python versions, or across different (virtual) environments. See
https://github.com/datalad/datalad-container/issues/224 for an example use case.

A corresponding test is included.

The ability to keep run-records paramterizable, in particular, for interpreters can also aid longevity (think platform-specific choices to call a system Python executable `python3` for some years, and then switch back.

Related https://github.com/datalad/datalad-container/pull/250

### DIFF
--- a/datalad/core/local/run.py
+++ b/datalad/core/local/run.py
@@ -38,6 +38,7 @@ from datalad.interface.base import (
     build_doc,
     eval_results,
 )
+from datalad.interface.common_cfg import definitions as cfg_defs
 from datalad.interface.common_opts import (
     jobs_opt,
     save_message_opt,
@@ -623,7 +624,12 @@ def format_command(dset, command, **kwds):
     command = normalize_command(command)
     sfmt = SequenceFormatter()
 
-    for k, v in dset.config.items("datalad.run.substitutions"):
+    for k in set(cfg_defs.keys()).union(dset.config.keys()):
+        v = dset.config.get(
+            k,
+            # pull a default from the config definitions
+            # if we have no value, but a key
+            cfg_defs.get(k, {}).get('default', None))
         sub_key = k.replace("datalad.run.substitutions.", "")
         if sub_key not in kwds:
             kwds[sub_key] = v
--- a/datalad/core/local/run.py
+++ b/datalad/core/local/run.py
@@ -625,6 +625,8 @@ def format_command(dset, command, **kwds):
     sfmt = SequenceFormatter()
 
     for k in set(cfg_defs.keys()).union(dset.config.keys()):
+        if not k.startswith("datalad.run.substitutions."):
+            continue
         v = dset.config.get(
             k,
             # pull a default from the config definitions
--- a/changelog.d/pr-7509.md
+++ b/changelog.d/pr-7509.md
@@ -1,0 +1,3 @@
+### 🐛 Bug Fixes
+
+- Enhance `run()` placeholder substitutions to honor configuration defaults.  [PR #7509](https://github.com/datalad/datalad/pull/7509) (by [@mih](https://github.com/mih))
--- a/datalad/core/local/tests/test_run.py
+++ b/datalad/core/local/tests/test_run.py
@@ -21,6 +21,7 @@ from os import (
     mkdir,
     remove,
 )
+import pytest
 from unittest.mock import patch
 
 from datalad.api import (
@@ -43,6 +44,7 @@ from datalad.support.exceptions import (
 from datalad.tests.utils_pytest import (
     DEFAULT_BRANCH,
     OBSCURE_FILENAME,
+    SkipTest,
     assert_false,
     assert_in,
     assert_in_results,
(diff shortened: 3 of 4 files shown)


mih:
MacOS failure looks unrelated (web remote).

datalad/core/local/run.py
```diff
@@ -623,7 +624,12 @@ def format_command(dset, command, **kwds):
     command = normalize_command(command)
     sfmt = SequenceFormatter()
 
-    for k, v in dset.config.items("datalad.run.substitutions"):
+    for k in set(cfg_defs.keys()).union(dset.config.keys()):
+        v = dset.config.get(
```
yarikoptic, line 628: explicit better than implicit, so we do not have some surprises if something overlaps in the names subspace after we do `k.replace` below, could we have smth like:

```suggestion
        if not k.startswith("datalad.run.substitutions."):
            continue
        v = dset.config.get(
```
?

yarikoptic (commented):
I need to understand it better: isn't this just a singular case of a larger common (?) pattern that in many (most?) cases we would actually like something like this for any `Dataset.config` access since dataset's config should be a "union" with system/user/env-overloads settings?

mih:
I dont know if this is a common case. Typically, access that requires acknowledging a configurable default goes via `obtain()`. Here it is different, because it needs to look for matches and not for a specific item.

datalad/core/local/run.py
mih: Makes a lot of sense. Feel free to commit, when all issues are expressed. I want to avoid the expensive CI reruns.

mih: I took another look while updating the -next patch. I went for this https://github.com/datalad/datalad-next/pull/488

yarikoptic: shouldn't we have it here?

mih: I don't think it matters now. This is a technical detail that could be rectified at the end, if desired. You asked much broader questions to which I had no answers. I don't know if and in which direction this PR could move forward. Once that is clear to me, I will take another look.

I have released the patch with datalad-next 1.0.1, so the immediate needs are addressed and this PR can move in its own pace.

yarikoptic: in effect that next patch to do the same, so let's try this suggested one finally here now.

yarikoptic:
<details>
<summary>appveyor might have stalled (restarted). </summary> 

```shell
[gw1] PASSED ../datalad/support/tests/test_sshconnector.py::test_bundle_invariance 
../datalad/support/tests/test_sshrun.py::test_exit_code CommandError: 'ssh -o ControlPath=/home/appveyor/.cache/datalad/sockets/bad871d9 datalad-test 'exit 42'' failed with exitcode 42
[gw1] XPASS ../datalad/support/tests/test_sshrun.py::test_exit_code 
../datalad/support/tests/test_sshrun.py::test_no_stdin_swallow 
[gw1] PASSED ../datalad/support/tests/test_sshrun.py::test_no_stdin_swallow 
../datalad/support/tests/test_sshrun.py::test_fancy_quotes 
[gw1] PASSED ../datalad/support/tests/test_sshrun.py::test_fancy_quotes 
../datalad/support/tests/test_sshrun.py::test_ssh_option 
[gw1] SKIPPED ../datalad/support/tests/test_sshrun.py::test_ssh_option 
../datalad/support/tests/test_sshrun.py::test_ssh_ipv4_6_incompatible 
[gw1] PASSED ../datalad/support/tests/test_sshrun.py::test_ssh_ipv4_6_incompatible 
../datalad/support/tests/test_sshrun.py::test_ssh_ipv4_6 ssh: Could not resolve hostname 127.0.0.1: Address family for hostname not supported
[gw1] PASSED ../datalad/support/tests/test_sshrun.py::test_ssh_ipv4_6 
../datalad/support/tests/test_stats.py::test_ActivityStats_basic 
[gw1] PASSED ../datalad/support/tests/test_stats.py::test_ActivityStats_basic 
../datalad/support/tests/test_stats.py::test_ActivityStats_comparisons 
[gw1] PASSED ../datalad/support/tests/test_stats.py::test_ActivityStats_comparisons 
../datalad/support/tests/test_stats.py::test_add 
[gw1] PASSED ../datalad/support/tests/test_stats.py::test_add 
../datalad/support/tests/test_status.py::test_FileStatus_basic 
[gw1] PASSED ../datalad/support/tests/test_status.py::test_FileStatus_basic 
../datalad/support/tests/test_vcr_.py::test_use_cassette_if_no_vcr 
```
</details>

OSX restarted... travis seems no more (I ended subscription but was hoping that opensource thing would kick in... will check later)

mih:
Closing, before it hits the one year mark.

PR #7513, opened 2023-10-17
Copy script to install singularity from datalad-container and use it in the workflow

yarikoptic:
That script now installs also libfuse2 library needed by singularity-ce, which lead to failed installations in this repo CI.

Ideally we should establish central collection of CI tools at some point to reuse across various extensions and derivative projects. ATM we will have 3 copies of this script: here, in datalad-container and datalad-extensions.


adswa (approved):
looks good!

yarikoptic:
so green it is surreal. Thanks @adswa for the approval! let's proceed

PR #7522, opened 2023-10-25
Robustify update target detection for adjusted mode datasets

### LINKED ISSUE #7507: `update --how(-subds) (!=fetch)` cannot handle adjusted branches
Here is an example

```
C:\Users\mih\paper-remodnav>datalad update -r --how-subds reset
[INFO] Fetching updates for Dataset("C:\Users\mih\paper-remodnav")
update(ok): . (dataset)
[INFO] Fetching updates for Dataset("C:\Users\mih\paper-remodnav\data\raw_eyegaze")
CommandError: '"git" "-c" "diff.ignoreSubmodules=none" "-c" "core.quotepath=false" "rev-parse" "--symbolic-full-name" "--abbrev-ref=strict" "@{upstream}"' failed with exitcode 128
fatal: no upstream configured for branch 'adjusted/master(unlocked)'

C:\Users\mih\paper-remodnav>cd C:\Users\mih\paper-remodnav\data\raw_eyegaze

C:\Users\mih\paper-remodnav\data\raw_eyegaze>git branch -a
* adjusted/master(unlocked)
  git-annex
  master
  remotes/mddatasrc/git-annex
  remotes/mddatasrc/master
  remotes/mddatasrc/synced/git-annex
  remotes/mddatasrc/synced/master
  remotes/origin/HEAD -> origin/master
  remotes/origin/git-annex
  remotes/origin/master
```

It looks, as if it is simply not considering the presence of adjusted branches.

Looking at the scenarios affected: Any update mode other than `fetch` leads to this crash, and need not be a subdataset either.

mih:
The true cause of the problem is not well understood. https://github.com/datalad/datalad/issues/7507 documents that I was not able to capture the breakage in a test, despite approaching it from different angles.

Only the original reproducer documents that this patch fixes the problem. However, the involved datasets are too heavy to be used in an already overloaded test battery. Hence this fix is proposed without a corresponding test.

Closes #7507

### DIFF
--- a/changelog.d/20231025_194118_michael.hanke_bf_7507.md
+++ b/changelog.d/20231025_194118_michael.hanke_bf_7507.md
@@ -1,0 +1,6 @@
+### 🐛 Bug Fixes
+
+- Update target detection for adjusted mode datasets has been improved.
+  Fixes [#7507](https://github.com/datalad/datalad/issues/7507) via
+  [PR #7522](https://github.com/datalad/datalad/pull/7522)
+  (by [@mih](https://github.com/mih))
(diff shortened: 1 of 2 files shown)


yarikoptic:
Well, I am not familiar with the syntax used here and no explanation for the actual change was provided but it doesn't break anything, so that's reassuring thus let's proceed

PR #7529, opened 2023-11-03
Generic issue -> Report a Bug template, Add a "Report a Generic Issue", disable "Open a blank"

### LINKED ISSUE #7528: Creating a "Generic Issue" should not present the user with prompts for a bug report
### What is the problem?

When I go to create an issue in this repository, the only option applicable to my use cases is to create a "Generic Issue", but the titles of the fields in the form I am directed to all assume that I'm reporting a bug; the very first field is even titled "What is the problem?".  What if I'm not reporting a bug and instead am requesting a feature or (as I was just doing) suggesting ways to improve the quality of the codebase?

If I click "Generic Issue", the form I'm presented with (if I'm shown a form at all) should actually be for a generic issue (where "issue" is in the looser GitHub sense of "a thing to be worked on", not "a thing that's wrong").  The current "Generic Issue" setup should instead be used for a "Report a Bug" option on the "New Issue" page.

_No response_

### DataLad information

_No response_

### Additional context

_No response_

_No response_

yarikoptic:
- as reported by @jwodder (Closes #7528) the form is really about reporting a bug, not just opening some generic issue
- we also have "Report a Security Vulnerability" internal github issue reporting we enabled in response to #7369;  Taken together with default "Open a blank issue", we would make it consistent in being imperative

edit: I also renamed "Gooey Issue" to be imperative


jwodder:
This just changes the name of one of the options the user is shown when creating an issue.  I think you still need to add a dedicated template for generic issues.

yarikoptic:
> This just changes the name of one of the options the user is shown when creating an issue. I think you still need to add a dedicated template for generic issues.

What would you like to see there in comparison to the default one you get with [Open a blank issue.](https://github.com/datalad/datalad/issues/new)?

jwodder:
@yarikoptic 

> What would you like to see there in comparison to the default one you get with [Open a blank issue.](https://github.com/datalad/datalad/issues/new)?

I mainly want the "generic/blank issue" option to be given the same prominence as the other options, not for it to be just an easy-to-miss link at the bottom of the page.

yarikoptic:
Something like

image

? implemented here now

yarikoptic:
> @yarikoptic Tacking on "which is not necessarily a bug" seems a little excessive, but good enough.

I am open to an alternative wording if someone has a better one in mind, I just wanted to make sure that it makes it clear for submitter which ones to use among available. May be other @datalad/developers have an idea for better wording - ideally suggest in suggestion mode.

yarikoptic:
a side-effect from above is that the "Generic" issue would gain that mandatory **Description** section in it, so it would not be totally a reproduction of a blank issue , see e.g. https://github.com/yarikoptic/datalad/issues/15 . I didn't spot ability to just make it without any section.  IMHO it is ok, but wanted to make sure that we have no surprises.  I will ping other @datalad/developers on matrix now.

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1 @@
+blank_issues_enabled: false
```
adswa, line 1: I usually open a blank issue, do we really have to remove it entirely?

mih: Same

adswa, line 1: Can we keep this? :)

```suggestion
blank_issues_enabled: true
```

yarikoptic: "Report a generic issue" is pretty much that, that is why I disabled to avoid possible confusion. But as I don't care enough -- I now committed suggested making it available as well.

yarikoptic:
I accepted recommendation.  Would merge in a week unless there is further feedback/changes desired

PR #7541, opened 2023-12-11
BF(TST): Relax matching of git-annex error message about unsafe drop, which was changed in 10.20231129-18-gfd0b510573

yarikoptic:
What matters really is that we fail and that there is "unsafe" and some message from git-annex.  In [10.20231129-18-gfd0b510573](https://git.kitenet.net/index.cgi/git-annex.git/commit/?id=fd0b5105731b17899d79cade82436320ab85cb56) git-annex changed message to become better but I do not think we should overcomplicate test ot match both messages exactly since who knows what other improvements might come later again.

Marking it for release so we bring CI of git-annex builds back into green


yarikoptic:
travis single fail to run is unrelated,   it is trivial, merging

yarikoptic-gitmate:

Issue #7313, opened 2023-03-07
test_rerun_fastforwardable_mutator fails some times on CrippledFS


yarikoptic:
most recently ran into in https://github.com/datalad/datalad/pull/7312 but was seen before exactly in that workflow

```
(git)smaug:/mnt/datasets/datalad/ci/logs/2023[master]git
$> GIT_TERMINAL_PROMPT=0 datalad foreach-dataset --o-s relpath -r -J10 git grep 'error: Your local changes to the following files would be overwritten by merge:'                                                                      
03/07/pr/7312/0f61a79/github-CrippledFS-3896-failed/1_test (1).txt:2023-03-07T00:44:57.7948909Z E               ' --no-ff --allow-unrelated-histories 508340379dd243ed50b8ca95b511a12999e4b5aa' failed with exitcode 128 [err: 'error: Your local changes to the following files would be overwritten by merge:
03/07/pr/7312/0f61a79/github-CrippledFS-3896-failed/1_test (1).txt:2023-03-07T00:44:57.7986115Z ' --no-ff --allow-unrelated-histories 508340379dd243ed50b8ca95b511a12999e4b5aa' failed with exitcode 128 [err: 'error: Your local changes to the following files would be overwritten by merge:
03/07/pr/7312/0f61a79/github-CrippledFS-3896-failed/test/8_Run tests.txt:2023-03-07T00:44:57.7948904Z E               ' --no-ff --allow-unrelated-histories 508340379dd243ed50b8ca95b511a12999e4b5aa' failed with exitcode 128 [err: 'error: Your local changes to the following files would be overwritten by merge:
03/07/pr/7312/0f61a79/github-CrippledFS-3896-failed/test/8_Run tests.txt:2023-03-07T00:44:57.7986108Z ' --no-ff --allow-unrelated-histories 508340379dd243ed50b8ca95b511a12999e4b5aa' failed with exitcode 128 [err: 'error: Your local changes to the following files would be overwritten by merge:
02/23/pr/7301/f884340/github-CrippledFS-3880-failed/1_test (1).txt:2023-02-23T14:05:28.9753834Z E               ' --no-ff --allow-unrelated-histories 4f5f08136b567a66406c75ed68c50ed494ad9c5c' failed with exitcode 128 [err: 'error: Your local changes to the following files would be overwritten by merge:
[...]

PR #7546, opened 2024-01-02
Update `pytest_ignore_collect()` for pytest 8.0

jwodder:
[Pytest 8.0 is coming](https://github.com/pytest-dev/pytest/releases/tag/8.0.0rc1), and one of the breaking changes is [the removal of the `py.path.local` arguments to various hooks](https://docs.pytest.org/en/latest/deprecations.html#py-path-local-arguments-for-hooks-replaced-with-pathlib-path), such as `pytest_ignore_collect()`.  This PR updates our use of that hook.

### DIFF
--- a/changelog.d/pr-7546.md
+++ b/changelog.d/pr-7546.md
@@ -1,0 +1,3 @@
+### 🏠 Internal
+
+- Update `pytest_ignore_collect()` for pytest 8.0.  [PR #7546](https://github.com/datalad/datalad/pull/7546) (by [@jwodder](https://github.com/jwodder))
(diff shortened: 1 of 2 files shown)


yarikoptic:
crippledfs fail - known https://github.com/datalad/datalad/issues/7313 . The other is `test_download_ftp` which failed on two appveyor runs, relies on accessing external resource so I guess it is just bad connection and clearly unrelated to this change, so let's proceed. Thank you @jwodder

yarikoptic-gitmate:
PR released in [`0.19.6`](https://github.com/datalad/datalad/releases/tag/0.19.6)

PR #7549, opened 2024-01-23
Make ORA remote more aware of the remote OS when using SSH

### LINKED ISSUE #7536: Publishing to ria store
### What is the problem?

I created a remote RIA store using `ria+ssh` protocol. 
Creation of the remote works without errors and results in the appropriate structure on the ssh remote. 
When actually publishing to the ria store however, datalad fails due to invalid permission settings on the remote side: 
```
Unable to remove <my-remote-path>/datastore/bbc/62b4a-cdad-4b17-ba6e-47bb04bb12f0/ora-remote-ec45211b-96b8-48ce-8386-907725f6fac1/transfer/MD5E-s305407--4ed88684b9e0d8db36e433fc771ff7e3.png or to obtain write permission in parent directory. -caused by- stat -f%Dp <my-remote-path>/datastore/bbc/62b4a-cdad-4b17-ba6e-47bb04bb12f0/ora-remote-ec45211b-96b8-48ce-8386-907725f6fac1/transfer failed:
copy: 1 failed
```
Indeed, the png file in the transfer folder on the target, as well as its parent folder `transfer` have permissions `-r--r--r--`. The `ore-remote- ...` folder has permissions `drwxrwxr-x`. I have no influence on those permissions since the complete tree is auto-created by datalad. 

Using DataLad version 0.19.3 and git version 2.43.0 on a MacBook M1. 
I tried different ssh remotes, including the JUDAC system in Jülich and a local ubuntu box.
Done the same with `ria+file` to a local folder works fine.

1) Create test datalad repository
`datalad create test`

2) Add a test file to the repository and `datalad save`

3) Create ria remote on ssh machine
`datalad create-sibling-ria -s datastore --new-store-ok "ria+ssh://<user>@<server>/path/to/datastore"`

4) Try to push
`datalad push --to datastore`

### DataLad information

0.19.3
darwin/23.1.0 14.1.1/arm64

### Additional context

_No response_

_No response_

mslw:
The ORA remote (more specifically, its `SSHRemoteIO.ensure_writeable` context manager) runs the `stat` command on the remote end to check file permissions. The command format and its output differ between macOS and Linux ("It doesn't even think about a windows server").

Previously, the format was chosen based on the *local* OS, which makes no sense for a command that is executed on the *remote* end. This led to situations where it was impossible to push data from Mac to Linux because the stat command errored out, and permissions were left incorrect, when moving a file from `transfer` to its desired location - see #7536 

With this change, the ORA remote will run `uname -S` on the remote end (of an SSH connection) to determine where it is operating, if it needs to. To the best of my knowledge (https://en.wikipedia.org/wiki/Uname), checking for "Darwin" should be sufficient to detect macOS.

This PR adds `remote_uname` as a lazy property of ORA remote's `SSHRemoteIO` class, which is accessed before running said `stat` commands (replacing a *locally operating* check of `on_osx`).

Having access to the property avoids re-running `uname` for every file that needs to be touched, and having the property lazily
resolved avoids running `uname` for operations which don't need it.

This should fix #7536

### DIFF
--- a/changelog.d/pr-7549.md
+++ b/changelog.d/pr-7549.md
@@ -1,0 +1,7 @@
+### 🐛 Bug Fixes
+
+- Correct remote OS detection when working with RIA (ORA) stores: this
+  should enable RIA operations, including push, from Mac clients to
+  Linux hosts (and likely vice versa).
+  Fixes [#7536](https://github.com/datalad/datalad/issues/7536)
+  via [PR #7549](https://github.com/datalad/datalad/pull/7549) (by [@mslw](https://github.com/mslw))
--- a/datalad/distributed/ora_remote.py
+++ b/datalad/distributed/ora_remote.py
@@ -474,9 +474,12 @@ class SSHRemoteIO(IOBase):
         """
 
         path = sh_quote(str(path))
-        # remember original mode -- better than to prescribe a fixed mode
 
-        if on_osx:
+        # remember original mode -- better than to prescribe a fixed mode
+        # format depends on remote OS, use uname (works on Linux and macOS)
+        uname_out = self._run("uname -s", no_output=False, check=True).rstrip()
+
+        if uname_out == "Darwin":
             format_option = "-f%Dp"
             # on macOS this would return decimal representation of mode (same
             # as python's stat().st_mode
--- a/datalad/distributed/ora_remote.py
+++ b/datalad/distributed/ora_remote.py
@@ -356,6 +356,22 @@ class SSHRemoteIO(IOBase):
         # make sure default is used when None was passed, too.
         self.buffer_size = buffer_size if buffer_size else DEFAULT_BUFFER_SIZE
 
+        # lazy property to store the remote unix name
+        self._remote_uname = None
+
+    @property
+    def remote_uname(self):
+        """Remote unix system name, lazy resolution
+
+        If accessed for the first time, runs uname -s to find out
+
+        """
+        if self._remote_uname is None:
+            self._remote_uname = self._run(
+                "uname -s", no_output=False, check=True
+            ).rstrip()
+        return self._remote_uname
+
     def close(self):
         # try exiting shell clean first
         self.shell.stdin.write(b"exit\n")
@@ -476,10 +492,7 @@ class SSHRemoteIO(IOBase):
         path = sh_quote(str(path))
 
         # remember original mode -- better than to prescribe a fixed mode
-        # format depends on remote OS, use uname (works on Linux and macOS)
-        uname_out = self._run("uname -s", no_output=False, check=True).rstrip()
-
-        if uname_out == "Darwin":
+        if self.remote_uname == "Darwin":
             format_option = "-f%Dp"
             # on macOS this would return decimal representation of mode (same
             # as python's stat().st_mode


yarikoptic:
makes sense to me, rerunning failed jobs but might be related given the specific of failure  (both OSX) ;-)

mslw:
Looking at the results after the rerun, the only failing test is now:

1. `../distributed/tests/test_push.py::test_force_checkdatapresent` (that's `core/distributed/...`)

which did not fail before.

Looking at the recent macOS test runs for different PRs, I see that the same error cropped up at least once [^1] ([run 7645277233](https://github.com/datalad/datalad/actions/runs/7645277233/)) but passed in most of them (e.g. all green [run 7782008144](https://github.com/datalad/datalad/actions/runs/7782008144)). All these runs use the same git-annex version.

I don't see the connection to the code changes, and will therefore re-run the macOS tests once again :crossed_fingers: 

One observation in case the failure persists or happens elsewhere: why would the `test_force_checkdatapresent` produce spurious results but only sometimes? In the test's code I see comments addressing changes in how git-annex handles timestamps (nb. the comment includes a link, but the link target says "HEAD" instead of the full ID of [this commit](https://git.kitenet.net/index.cgi/git-annex.git/commit/?id=eb59da9dd204b6c15ffd3e1cfb66acde5f0e904a)), and the test asserts either "ok" or "notneeded" result depending on the git-annex version. I did not fully understand the nature of the changes, but maybe the behavior can be affected by when things happen (within a second or not), and the test does not sufficiently account for that?


[^1]: For the methodical log greppers: ` FAILED ../distributed/tests/test_push.py::test_force_checkdatapresent - AssertionError: Desired result`

yarikoptic:
indeed we had it failed elsewhere as well, but there it seems to fail consistently and only on OSX and this PR does concern only with OSX... didn't look deeper but it might be just that this PR unravels otherwise not previously observed issue

mslw:
Thanks for checking!

Alas, re-running test on macOS for this PR after my previous comment I got `test_force_checkdatapresent` failing with both snapshot and brew. I still don't think it's related (the test pushes between two local repos) but I don't have good arguments and I understand that you first want to investigate further for underlying issues, related or not.

I looked a little deeper but I'm on Debian. So far I can't reproduce locally (using latest autobuild git-annex), and still don't have a good clue of what might be going on.

mslw:
Hey, I rebased the pull request onto the 1.0 maint, and all tests are now passing, no doubt thanks to #7581  (see checklist for `HEAD~1` - I unwisely added a changelog commit with skip ci)

The only thing that is red is codecov/project, and I don't think I can do anything about it.

One big caveat is that the datalad test setup can not test operation between OSs, so we have to rely on manual testing. I don't have a Mac, but @christian-monch does and I heard from him (please confirm) that the patch indeed works as intended.

@yarikoptic would you consider merging on that basis?

adswa (approved):
It looks like all blockers are removed here

mslw:
Thanks! This fix has made its way to datalad-next in the meantime, but it also seems useful to have it in here.

yarikoptic-gitmate:
PR released in [`1.1.2`](https://github.com/datalad/datalad/releases/tag/1.1.2)

PR #7551, opened 2024-01-25
Add the "http_token" authentication mechanism which provides 'Authentication: Token {TOKEN}' header

yarikoptic:
It is pretty much the "Bearer TOKEN" method but which uses different keyword "Token".  It is e.g. the one provided by Django REST Framework. GitHub allows for both 'Bearer' and 'Token' keywords: https://docs.github.com/en/rest/authentication/authenticating-to-the-rest-api?apiVersion=2022-11-28

In our case of DANDI we ran into it only now that we decided to add handling of embargoed datasets, for which we need to authenticate to access the files.  We are using Django REST Framework and insofar decision was made to retain current approach (and fix WWW-Authentication header response), instead of an alternative to allow for both  'Bearer' and 'Token' as was suggested in https://github.com/dandi/dandi-archive/pull/1830

As it is ad-hoc (nothing AFAIK in W3C standard) Authentication response I do not think some automation would ever be created based on WWW-Authentication but we will be able to support such providers.

### PR checklist

### DIFF
--- a/changelog.d/pr-7551.md
+++ b/changelog.d/pr-7551.md
@@ -1,3 +1,3 @@
-### 🐛 Bug Fixes
+### 🚀 Enhancements and New Features
 
 - Add the "http_token" authentication mechanism which provides 'Authentication: Token {TOKEN}' header.  [PR #7551](https://github.com/datalad/datalad/pull/7551) (by [@yarikoptic](https://github.com/yarikoptic))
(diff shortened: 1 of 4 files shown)


changelog.d/pr-7551.md
```diff
@@ -0,0 +1,3 @@
+### 🐛 Bug Fixes
```
yarikoptic, line 1: More of

```suggestion
### 🚀 Enhancements and New Features
```

yarikoptic:
fails unrelated (test_gh2043p1 on OSX and then ftp ones on windows appveyor) -- restarting

yarikoptic-gitmate:
PR released in [`0.19.6`](https://github.com/datalad/datalad/releases/tag/0.19.6)

PR #7552, opened 2024-01-26
Escape the \ in a docstring

yarikoptic:
Otherwise lead to the warning

change is so minor that I do not think it is worth a changelog

### DIFF
--- a/datalad/distribution/install.py
+++ b/datalad/distribution/install.py
@@ -63,7 +63,7 @@ lgr = logging.getLogger('datalad.distribution.install')
 class Install(Interface):
     """Install one or many datasets from remote URL(s) or local PATH source(s).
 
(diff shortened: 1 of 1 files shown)


yarikoptic:
FTR Mac fail is 

```
FAILED ../tests/test_annexrepo.py::test_is_available[False] - datalad.runner.exception.CommandError: CommandError: 'git -c diff.ignoreSubmodules=none -c core.quotepath=false annex find --anything --json --json-error-messages -c annex.dotfiles=true -- test-annex.dat' failed with exitcode 139 under /private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/datalad_temp_test_is_available7vgluc6z [info keys: stdout_json] [err: 'error: git-annex died of signal 11']
```

PR #7559, opened 2024-02-07
adjust for new error message in git cat-file

adswa:
Since at least Git 2.39.2, the error reads:

❱ git cat-file -p git-annex:remote.log
fatal: invalid object name 'git-annex'.

This change adjusts for this in a similar way it has so far been accounted for


adswa:
The extension test for datalad-container failing is unrelated - its a missing dependency for the singularity installation.

yarikoptic:
I am promoting this PR to major advancement of datalad.

PR #7578, opened 2024-04-09
Raise exception if annex remote needs interaction

christian-monch:
This PR is an attempt to deal with the issue where `datalad get` stalls if an annex remote tries to ask for credentials.


christian-monch:
> I don't think this would be legit -- it was the whole point to be able to request credentials from e.g. nested `git-annex-remote-datalad` . Or am I missing the effect here? I guess can try with this patch on some ///crcns dataset to see if still works.

Thanks @yarikoptic  for the comment (this is still in an "experimental" state). I modified it to explicitly check for a controlling terminal on Linux and OSX via `open('/dev/tty')`. Have not yet determined an appropriate check on Windows. I will check with some ///crcns datasets and ping you.

christian-monch:
I tested `datalad install -g ///crcns/aa-1` on Linux and OSX. It properly prompts for passwords on both systems. If `datalad install -g ///crcns/aa-1` is executed via ssh-login, i.e. in a process without a terminal, it does not ask for credentials. As a result it will terminate after some time with the following result:
```
action summary:
  get (error: 5950, ok: 8)
  install (ok: 1)
```
(BTW: there is no further feedback during the process, which can give the impression that the process is hanging.)

yarikoptic (commented):
for now just comments/minor suggestions as trying also to get a good sense of what is going

christian-monch:
> [...]
> so overall -- sounds good, and I wonder if we should may be do smth like
> 
> ```diff
> diff --git a/datalad/ui/dialog.py b/datalad/ui/dialog.py
> index 4d217fd45..2c8f5a95a 100644
> --- a/datalad/ui/dialog.py
> +++ b/datalad/ui/dialog.py
> @@ -346,6 +346,10 @@ class UnderAnnexUI(DialogUI):
>          super(UnderAnnexUI, self).__init__(**kwargs)
>          self.specialremote = specialremote
>  
> +    @property
> +    def is_interactive(self):
> +        return super().is_interactive() and can_prompt()
> +
>      def set_specialremote(self, specialremote):
>          lgr.debug("Setting specialremote of UI %s to %s", self, specialremote)
>          self.specialremote = specialremote
> ```
>
> or may be even at DialogUI level. [...]

I am not sure that I completely understand the ui-code and I wanted to keep the PR small, therefore I tried to touch `DialogUI` and friends as little as possible. I am grateful for any help in this area, please propose any change you feel should be in there.

christian-monch:
ping @yarikoptic : shall we merge this PR?

yarikoptic (approved):

yarikoptic-gitmate:

PR #7580, opened 2024-04-16
Load extensions in annex remote entrypoint

mih:
The main entrypoint for annex remotes now also runs the standard extension load hook. This enables extensions to alter annex remote implementation behavior in the same way than other DataLad components.

I see no reason why this code should be an exception.


mih:
Putting in draft for now, still unsatisfactory behavior

mih:
I have verify that this approach is capable of patching a remote implementation from extension code "by hand". Patches will soon appear and exercise the mechanism routinely.

mih:
The failures are unrelated "old friends".

mih:
I have added a 'release' label to this PR. It would be beneficial to have this available quickly, because it simplifies test setups for patch development. Thx!

PR #7585, opened 2024-04-24
Stop testing on Python 3.7

### LINKED ISSUE #7584: MacOS tests fail to install Python 3.7 (which is EOL anyway)
Our GitHub actions for MacOS testing use `macos-latest`, and it looks like they just started failing to set up Python 3.7: see https://github.com/datalad/datalad/actions/runs/8817812378/job/24205131401#step:4:11

```
Error: The version '3.7' with architecture 'arm64' was not found for macOS 14.4.1.
```

This was not the case 2 days ago, when the test ran on macOS 12.

Given that [3.7 reached EOL 10 months ago](https://endoflife.date/python), I think we need to bump the Python version rather than change the macOS version. For macOS tests for sure, maybe also for others if still used.

Related: #7400

mslw:
This switches all tests which still used Python 3.7 to use Python 3.8 (or, in Appveyor, 3.11 - see below).

The only exception, and potentially controversial change, is Appveyor, where I went from 3.7 to 3.11 in e52e761. A move to 3.8 would sound logical, but... if I'm seeing correctly, the only difference between Ubu20P37a & Ubu20P37b versus Ubu20core & Ubu20a1 is that the former had "PY: 3.7", while the latter did not set PY at all. And if I'm reading the logs of the latest Appveyor run correctly, Ubu20core & Ubu20a1 run (implicitly?) on Python 3.8.10. I cound not find confirmation that this is the default, but it seems to be.

An alternative (for Appveyor) would be to set PY for Ubu20core & Ubu20a1 to the newest supported (3.11) and use the other two to test oldest supported (3.8). But maybe appveyor tends to stay on the lowest non-EOL version when PY is not specified, in which case the current setup would be preferred (just a guess - couldn't find confirmation)? Please advise.

Note: We still officially support Python 3.7 (`python_requires>=3.7` in setup.py) but I feel a change there is overdue, too, and an easy candidate for a next PR to main.

Fixes #7584 (where tests on latest macOS started failing outright on Python setup).
Follows in the footsteps of #7413.
Addresses part of #7400.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -168,9 +168,9 @@ environment:
       CODECOV_BINARY: https://uploader.codecov.io/latest/macos/codecov
 
     # Test alternative Python versions
-    - ID: Ubu20P37a
+    - ID: Ubu20P311a
       # ~35min
-      PY: 3.7
+      PY: 3.11
       DTS: >
           datalad.cli
           datalad.core
@@ -183,9 +183,9 @@ environment:
       CODECOV_BINARY: https://uploader.codecov.io/latest/linux/codecov
       # system git-annex is way too old, use better one
       INSTALL_GITANNEX: git-annex -m deb-url --url https://datasets.datalad.org/datalad/packages/neurodebian/git-annex_8.20210903-1_amd64.deb
-    - ID: Ubu20P37b
+    - ID: Ubu20P311b
       # ~25min
-      PY: 3.7
+      PY: 3.11
       DTS: >
           datalad.downloaders
           datalad.interface
--- a/.github/workflows/test_macos.yml
+++ b/.github/workflows/test_macos.yml
@@ -25,10 +25,10 @@ jobs:
       with:
         fetch-depth: 0
 
-    - name: Set up Python 3.7
+    - name: Set up Python 3.11
       uses: actions/setup-python@v5
       with:
-        python-version: 3.7
+        python-version: 3.11
 
     - name: Install git-annex
       run: |
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -26,10 +26,10 @@ jobs:
     - uses: actions/checkout@v4
       with:
         fetch-depth: 0
-    - name: Set up Python 3.7
+    - name: Set up Python 3.8
       uses: actions/setup-python@v5
       with:
-        python-version: 3.7
+        python-version: 3.8
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
--- a/.github/workflows/docbuild.yml
+++ b/.github/workflows/docbuild.yml
@@ -13,10 +13,10 @@ jobs:
         git config --global user.email "test@github.land"
         git config --global user.name "GitHub Almighty"
     - uses: actions/checkout@v4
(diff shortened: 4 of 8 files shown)


.github/workflows/test_macos.yml
```diff
@@ -25,10 +25,10 @@ jobs:
       with:
         fetch-depth: 0
 
-    - name: Set up Python 3.7
+    - name: Set up Python 3.8
       uses: actions/setup-python@v5
       with:
-        python-version: 3.7
+        python-version: 3.8
```
yarikoptic, line 31: I submitted separate PR for this (#7586) to expedite but I believe here we need at least 3.10

mslw: I'll go for 3.11, and rebase this branch in a few mins.

mslw:
Quick note regarding homebrew failure: if I'm reading the Internet correctly, homebrew uses /opt/homebrew (?) for installing packages on M1, different than on Intel, and datalad-installer hardcodes "/usr/local/bin" (with a TODO note): https://github.com/datalad/datalad-installer/blob/d35969dc599f9781a81788a43388543e2d656980/src/datalad_installer.py#L1614-L1616

And if we are on `macos-latest`, we are now on arm64 (M1).

yarikoptic:
eh -- our cron testing of datalad-installer was disabled "due to inactivity" in the repo, I reenabled it ... investigating details in https://github.com/datalad/datalad-installer/pull/197

mslw:
It looks to me like that the above PR to datalad-installer fixes installation on `macos-latest` (i.e. `macos-14-arm64`). I think once DataLad installer is released, tests for this PR can be re-run without further changes to test config. This PR would then remove all testing on Python 3.7 *and* give us tests on M1 mac.

The non-macOS tests were affected by "Connecting to neuro.debian.net (...) failed: Connection refused.", so I expect it was temporary.

.appveyor.yml
```diff
@@ -168,9 +168,9 @@ environment:
       CODECOV_BINARY: https://uploader.codecov.io/latest/macos/codecov
 
     # Test alternative Python versions
-    - ID: Ubu20P37a
+    - ID: Ubu20P311a
       # ~35min
-      PY: 3.7
+      PY: 3.11
```
yarikoptic, line 173: can we get to 3.8 wherever it was 3.7 here ? the point of using 3.7 was to ensure that code is compatible with the minimal python version we have supported. If we just crank up to 3.11 right away, we might miss some cases of incompatibility etc.

while eu is asleep I will add those changes for 3.8 -- if doesn't work, feel free to drop that commit

yarikoptic: actually screw it -- I kinda do not see global pattern there... but we now might be testing less with 3.8 than we were with 3.7... who knows ;-) green -- is good!

mslw: FTR, I think the "main" Ubuntu runs, (`Ubu20core` & `Ubu20a1` which have no python version specified) are on 3.8 - at least they were the last time I looked. So if setting only here, we might have ended up with testing on 3.8 twice @yarikoptic 

I made a note about that in the commit message 90ace3fe1188615bb812b58549178ca1754abedc

yarikoptic-gitmate:

PR #7590, opened 2024-05-07
Make github workflow which would ensure that PR against master would have everything from maint merged

yarikoptic:
I think we should retain the habit of releasing minor (or eventually next major) versions from `master` while patch fixes from `maint` (we might even keep releasing patch fixes for the prior minor I guess... but probably we shouldn't).

This way we can 'stage' for next minor release while releasing patch releases without bottleneck/force to release miors.

With this PR I want to make sure that when we do release next minor, we do include all fixes from `maint`.

As only pertinent to master branch -- I positioned it against master. Also would make it easier to check
 
### PR checklist

- Sophisticate workflow to trigger also on any other push to `master` or `maint` so it stays "up to date" even if no changes to PR happen -- decided to postpone for later... would need more work.

edit: I canceled appveyor since unrelated

### DIFF
--- a/changelog.d/pr-7590.md
+++ b/changelog.d/pr-7590.md
@@ -1,3 +1,3 @@
 ### 🏠 Internal
 
-- Make github workflow which would ensure that PR against master would have everything from maint merged.  [PR #7590](https://github.com/datalad/datalad/pull/7590) (by [@yarikoptic](https://github.com/yarikoptic))
+- Make github workflow which would ensure that a "release" PR against `master` would have everything from `maint` merged.  [PR #7590](https://github.com/datalad/datalad/pull/7590) (by [@yarikoptic](https://github.com/yarikoptic))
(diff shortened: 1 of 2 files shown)


yarikoptic:
ok, seems good. Let me merge now 
- https://github.com/datalad/datalad/pull/7578
into maint ;)

changelog.d/pr-7590.md
```diff
@@ -0,0 +1,3 @@
+### 🏠 Internal
+
+- Make github workflow which would ensure that PR against master would have everything from maint merged.  [PR #7590](https://github.com/datalad/datalad/pull/7590) (by [@yarikoptic](https://github.com/yarikoptic))
```
yarikoptic, line 3:
```suggestion
- Make github workflow which would ensure that a "release" PR against `master` would have everything from `maint` merged.  [PR #7590](https://github.com/datalad/datalad/pull/7590) (by [@yarikoptic](https://github.com/yarikoptic))
```

yarikoptic:
I clarified wording in the changelog entry. 
As it is just a "safe guard" for releases I think it would be ok to merge -- which I will do in a few days unless there is a push back.

yarikoptic:
I will merge, although now with months delay from original statement. Easy to revert back etc if doesn't do what was desired from it.

yarikoptic-gitmate:
PR released in [`1.2.0`](https://github.com/datalad/datalad/releases/tag/1.2.0)

PR #7601, opened 2024-05-20
Cancel lengthy running workflows if a new commit is pushed

jwodder:
With this change, if a branch is updated while any of the affected workflows are still running on the branch's previous `HEAD`, the outstanding runs are cancelled.  As a result, the runs for the updated `HEAD` will not have to wait as long for runners to become available, and resources will not be wasted on outdated code.


yarikoptic:
Thank you @jwodder !

yarikoptic-gitmate:

Issue #696, opened 2016-08-10
datalad install doesn't handle paths (urls?) with spaces correctly


yarikoptic:
a dump of struggle, but in general -- seems to not get it done and even removes empty submodule directory if existed before... quite ugly ;)

```
$> git co 'Cleveland CCF'
(venv-tests)2 10405.....................................:Wed 10 Aug 2016 01:35:05 PM EDT:.
(git)smaug:/mnt/datasets/datalad/test-crawled2/all-3/indi/fcon1000[master]git
$> git st
On branch master
Your branch is up-to-date with 'origin/master'.
nothing to commit, working directory clean
(venv-tests)2 10406.....................................:Wed 10 Aug 2016 01:35:07 PM EDT:.
(git)smaug:/mnt/datasets/datalad/test-crawled2/all-3/indi/fcon1000[master]git
$> datalad install 'Cleveland CCF'
2016-08-10 13:35:14,218 [INFO   ] Installing /mnt/datasets/datalad/test-crawled2/all-3/indi/fcon1000/Cleveland CCF (install.py:378)
2016-08-10 13:35:14,350 [WARNING] Parsed version of URL 'http://datasets.datalad.org/indi/fcon1000/Cleveland%20CCF/.git' differs from original 'http://datasets.datalad.org/indi/fcon1000/Cleveland CCF/.git' (network.py:391)
2016-08-10 13:35:14,350 [INFO   ] Installing <Dataset path=/mnt/datasets/datalad/test-crawled2/all-3/indi/fcon1000/Cleveland CCF> from http://datasets.datalad.org/indi/fcon1000/Cleveland CCF/.git (install.py:378)
2016-08-10 13:35:14,445 [WARNING] Parsed version of URL 'http://datasets.datalad.org/indi/fcon1000/./Cleveland%20CCF/.git' differs from original 'http://datasets.datalad.org/indi/fcon1000/./Cleveland CCF/.git' (network.py:391)
2016-08-10 13:35:14,445 [INFO   ] Installing <Dataset path=/mnt/datasets/datalad/test-crawled2/all-3/indi/fcon1000/Cleveland CCF> from http://datasets.datalad.org/indi/fcon1000/./Cleveland CCF/.git (install.py:378)
Nothing was installed
changes on filesystem:                                                                                                                                                                                                            
 Cleveland CCF | 1 -
(venv-tests)2 10407.....................................:Wed 10 Aug 2016 01:35:14 PM EDT:.
[...]

PR #7605, opened 2024-05-27
Ensure early patch-loading when running datalad-next extension tests

christian-monch:
Fixes issue [datalad-next #705](https://github.com/datalad/datalad-next/issues/705)

The code in this PR sets `DATALAD_EXTENSIONS_LOAD=next` when testing the datalad-next extension. This is necessary to ensure patches are loaded before the objects they patch are used. 

"Delayed" loading of patches might also be the cause of the errors that were fixed by PR [datalad-next #696](https://github.com/datalad/datalad-next/pull/696). 

I am proposing this patch because it lets datalad-extension tests pass in datalad-next version 1.4.0, and I believe that the previous fix, i.e. [datalad-next #696](https://github.com/datalad/datalad-next/pull/696), should be reverted in the future, if the associated debian-test failures are resolved. 

In the datalad-next extension tests [datalad-next #696](https://github.com/datalad/datalad-next/pull/696) masked a genuine error, i.e. calling new ria-code function from old ria-code functions that used faulty arguments.

ping @mih

### DIFF
--- a/.github/workflows/test_extensions.yml
+++ b/.github/workflows/test_extensions.yml
@@ -65,6 +65,10 @@ jobs:
           datalad-container)
             DL_NEED_SINGULARITY=1;;
         esac
+        case ${{ matrix.extension }} in
+          datalad-next)
+            DATALAD_EXTENSIONS_LOAD=next;;
+        esac
         {
         echo "DL_PIP_INSTALLS=$DL_PIP_INSTALLS"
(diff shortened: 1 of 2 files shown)


christian-monch:
Closing this. The proper fix is to skip these tests in `datalad` if `datalad-next` is loaded and add the modified tests to `datalad-next`.

Issue #7604, opened 2024-05-26
`_get_flexible_source_candidates_for_submodule` wild guessing 


bpinsard:
### What is the problem?

I finally managed to reproduce a somewhat puzzling user-reported behavior that originates from `_get_flexible_source_candidates_for_submodule`.

`datalad get` on a sub-dataset with an ssh URL and without ssh-key setup, will fallback to non-existent guessed https URL from the super-dataset tracked remote URL, if the super-dataset was installed with https.

For instance if:
super-dataset installed with url:  `https://github.com/user/superds.git`
sub-dataset url in `.gitmodules`:   `git@github.com:user/subds.git`

`datalad get subds_path` with try `git@github.com:user/subds.git` then try  `https://github.com/user/superds-subds_path.git` rather than rewriting to `https://github.com/user/superds-subds_path.git` first.

This can be fixed at the dataset level by adding a datalad get config such as ` subdataset-source-candidate-620userhttp = https://github.com/user/{name}.git` for each dataset that contains sub-dataset.

To simplify the reproduction of the behavior, it is simpler to run it in docker to avoid any existing ssh-key to allow ssh connection to github. 

### What steps will reproduce the problem?

docker run --entrypoint sh -it ghcr.io/courtois-neuromod/datalad:alpine
datalad install -s https://github.com/courtois-neuromod/cneuromod.git
cd cneuromod
datalad get -n anat

### DataLad information

0.16.7

### Have you had any success using DataLad before?

yes ;)

PR #7606, opened 2024-05-28
ENH: add submodule URL candidates by URL-rewrite to match parent protocol

bpinsard:
Attempt at fixing #7604 

This PR adds a new submodule URL candidate by simply matching the protocol to the parent dataset one. 

It could be extended to provide more alternative URLs, such as matching protocol and hostname (and git platform user) of the parent one, so if a dataset is recursive forked to another server it can be recursively installed from that new server without changing the `.gitmodules` URLs.

I added a dependency, but if there are internal code to flexibly rewrite git URLS I would rather use these.

There are many changes which are triggered by pre-commit isort, though it seems not applied to the master branch.

Mainly looking for feedback before moving further.


yarikoptic:
Thanks @bpinsard - an interesting idea!  I guess in part it could be mitigated by git's config ability to rewrite urls (`insteadof` etc).  I think it would still be valuable to extend somewhere documentation to exercise some particular use case to show when such functionality would be useful.  What is the particular itch you were addressing ?

Addition of dependency would indeed add some burden since not yet even in debian:

- [giturlparse isn't in debian yet](https://packages.debian.org/search?suite=default&section=all&arch=any&searchon=names&keywords=giturlparse)

I do not remember if we have some easy ways to mutate RIs, but we do their parsing so might be worth checking - there might be easy ways in https://github.com/datalad/datalad/blob/maint/datalad/support/network.py#L606 (e.g. decomposing into a record and then composing as a new RI of a different class), which would achieve the same without requiring additional dependency.

bpinsard:
@yarikoptic sorry for the late follow-up on that PR.

The issue we are having is illustrated in #7604 where dataset users will experience issues when installing recursively some validly linked sub-datasets. The fact that datalad wild-guesses some non-existent URL is puzzling for the end-user, it will not report failing to clone through the registered submodule URL, but will for report failure on incoherent malformed https URL.

`insteadof` can of course partially solve that issue , but AFAIK these need to be set-up on each install, for each submodule, and can't inherit from the repo.
Another option is to hard-code `subdataset-source-candidate-620userhttp = https://github.com/user/{name}.git` in the `.datalad/config`, and that should fix my issue.
I am more concern that this issue will happen to other users that might not have the skills to dive into the code to figure out what happens.

As for adding this new pure-python dependency: is that required for all dependencies to be an APT package in debian?
We could try to recreate the logic of giturlparse, but it seems to have built-in logic for the differences between git-hosting platforms which would make it more straightforwardly robust than any of my attempt at solving that issue.

yarikoptic:
> As for adding this new pure-python dependency: is that required for all dependencies to be an APT package in debian?

quick one: overall **no**, but just that it adds extra burden since we would then need to package it and then likely to maintain . FWIW it is already in [conda-forge](https://anaconda.org/conda-forge/giturlparse), maintained but not too actively (last release Sep 24, 2023 with some recent PRs awaiting any feedback), not that many open issues... so a viable candidate to be used.

bpinsard:
The rewrite is a fallback if the original submodule URL doesn't work, so it won't rewrite if the original URL worked straight out of the box. The current problem is that an already existing fallback implemented in datalad is generating nonsensical URLs for this usecase (might still work and be required for other cases).

yarikoptic:
notes
- requesting changelog (CI was fixed up recently, let's see how it goes)
- Joey implemented `url.<base>.annexInsteadOf` to supplement `pushInsteadOf` of git

yarikoptic:
@bpinsard please allow contributions from collaborators for this PR (checkmark on the right IIRC), since ATM our bot can't add changelog entry

```
Run actions-ecosystem/action-remove-labels@v1
Run git push origin HEAD:fix/source_candidate_https
remote: Permission to courtois-neuromod/datalad.git denied to yarikoptic-gitmate.
fatal: unable to access 'https://github.com/courtois-neuromod/datalad/': The requested URL returned error: 403
```

bpinsard:
Same issue as before: as it is from an organization fork this checkbox is not available.
I added @yarikoptic-gitmate  write permissions to the fork, let see if that helps to push.

bpinsard:
I think this would be useful as it fixes an issue we have with cneuromod datasets, where the current rewrite options from ssh to http are not working.  https://github.com/datalad/datalad/issues/7604
I had not though about the insecure case but:
- if the parent dataset was cloned from a non-secured (eg. http) source, the issue is already there with the weakpoint being the parent dataset that could compromise any submodule operation. 
- it is a fallback rewrite, it will still try with the originally defined protocol in the submodule.
- does datalad/git warn about insecure connection (http, deprecated TLS) when cloning from http(s), similar to what any web-browser does nowadays?
- I would expect any worthy git-hosting platforms to redirect to https and preventing any credentials to be used over http, but I might be mistaken.

yarikoptic:
for 
> if the parent dataset was cloned from a non-secured (eg. http) source, the issue is already there with the weakpoint being the parent dataset that could compromise any submodule operation

if could be http for some "http://local-to-me.site" whenever in submodule for some "https://farfaraway-across-many-ears.com", thus adopting insecure http where we hop domains. So may be we could compare the address/server name and do such swap of protocols only if address matches, would that still work for you?

bpinsard:
sorry, I am not entirely sure to follow what you are saying.
Is the scenario you are mentioning, something like:
 - super-dataset: http local so OK-ish
   |- subdataset set with SSH URL : if SSH unreachable/unauthorized -> rewrites as http instead of https -> security issue

yarikoptic:
yes!

bpinsard:
I removed the http case to address the security concern.

datalad/distribution/get.py
```diff
@@ -233,6 +239,9 @@ def _get_flexible_source_candidates_for_submodule(ds, sm):
                         remote_url,
                         alternate_suffix=False)
                 )
+                rewriten_url = rewrite_match_scheme(remote_url, sm_url)
+                if rewriten_url:
```
yarikoptic:
```suggestion
                if (rewriten_url := rewrite_match_scheme(remote_url, sm_url)):
```

yarikoptic (changes requested):
I like it! Almost there. Could you strip unrelated to this PR purpose changes (see the diff touching on files unrelated to this PR)?  Can just commit them and let's merge with the squash since I think history is not carrying useful info there.  I also provided one minor recommendation as we can now use `:=` assignments.

yarikoptic:
FWIW, apparently echoing what I already said above -- it is adding depends on https://github.com/nephila/giturlparse which is  avail through conda forge https://anaconda.org/channels/conda-forge/packages/giturlparse/overview but I do see now recent activities so should be ok there. Unfortunately nothing in debian yet: https://packages.debian.org/search?suite=default&section=all&arch=any&searchon=names&keywords=giturlparse , not even RFP or ITP :-/

bpinsard:
Not sure to follow. Do we want/require pure python packages to be conda and debian installable??

yarikoptic:
you do not worry about it -- it is more of "want" and info for me since otherwise I would not be able to upload it to those distributions until we gain those dependencies packaged. Thus reviewed.

docs/source/conf.py
```diff
@@ -14,13 +14,13 @@
...
 import sys
 from glob import glob
 from os import pardir
 from os.path import exists
 from os.path import join as opj
 from pathlib import Path
-import subprocess
 
 
 def setup(sphinx):
```
yarikoptic, line 26: ok, all such changes do not belong to this PR, but as no harm, and addressing some inconsistencies which otherwise would be fixed up when these files would be touched by related changes, let's just proceed

yarikoptic-gitmate:
PR released in [`1.6.0`](https://github.com/datalad/datalad/releases/tag/1.6.0)

PR #7618, opened 2024-06-17
rf(`CommandError`): now based on datasalad's `CommandError`

mih:
This change bases the essential `CommandError` class on the analog class in the new `datasalad` library. This primarily improves the user-facing messaging, and debugging. For example, a `CommandError.__repr__()` no longer returns an empty string, which has hindered investigations in the past.

However, most reporting improvements (e.g., alignment with Python's `subprocess` behavior) have been overwritten again, to maintain the original behavior.

The primary motivation for this change is that in countless places type comparisions with `CommandError` are made, and until all used `CommandError` implementations are using a common basis migration of other code is blocked.

Refs: https://github.com/datalad/datasalad/pull/17, https://github.com/datalad/datalad-next/pull/728

### DIFF
--- a/datalad/runner/exception.py
+++ b/datalad/runner/exception.py
@@ -18,33 +18,51 @@ from typing import (
     Optional,
 )
 
+from datasalad.runners import CommandError as _CommandError
+
 lgr = logging.getLogger('datalad.runner.exception')
 
 
-class CommandError(RuntimeError):
+class CommandError(_CommandError):
     """Thrown if a command call fails.
 
-    Note: Subclasses should override `to_str` rather than `__str__` because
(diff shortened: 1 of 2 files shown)


mih:
I guess it is OK to fail for 3.7. `datasalad` declares compatibility with 3.8+ only. The oldest Python version with support.

yarikoptic:
https://github.com/datalad/datalad-next/pull/729 was merged, so I guess there is no immediate need in this PR/adding new dependency for datalad?

mih:
Adopting the new implementation would end the era of reporting a CommandError as an empty string in tracebacks etc AFAICR.

PR #7621, opened 2024-06-18
Add codespell and minor fixuppers to pre-commit configuration and apply it to non-`datalad/` components

yarikoptic:
I added a few more checks/fixes to the pre-commit I will use more often to avoid adding typos etc.
Having added them, decided to apply to various pieces which are not main code, so conflicts are unlikely.


yarikoptic:
if no objections @datalad/developers , I will merge by the end of the week.

yarikoptic-gitmate:
PR released in [`1.1.1`](https://github.com/datalad/datalad/releases/tag/1.1.1)

PR #7622, opened 2024-06-19
Make test for presence of max_path in partitions not run for current psutil 6.0.0

yarikoptic:
See https://github.com/giampaolo/psutil/issues/2109 - reporting of max_path was removed

Should fix our CI which just turned red


yarikoptic:
Since no brainer and needed to bring CI into green, merging.

yarikoptic-gitmate:
PR released in [`1.1.1`](https://github.com/datalad/datalad/releases/tag/1.1.1)

PR #7628, opened 2024-07-01
Correct typo of `max_path_length`

### LINKED ISSUE #7627: `test_wtf` fails if `psutil` is installed
### What is the problem?

Due to a typo in `max_path_length` the following test-errors occur.

On `main`: `test_wtf` fails if `psutil<6.0.0` is installed.
On `master`: `test_wtf` fails if `psutil` is installed. This affects at least version 1.0.0 through 1.1.0

Execute `pip install --upgrade datalad` and `pip install --upgrade psutil` in your datalad-venv.

Execute `pytest --pyargs datalad.local.tests.test_wtf` in your datalad-venv

### DataLad information

_No response_

### Additional context

_No response_

_No response_

christian-monch:
Fixes #7627 

This commit corrects a typo that leads to a failing `test_wtf` on different versions of `datalad` and `psutil`.
is installed.

### DIFF
--- a/datalad/local/tests/test_wtf.py
+++ b/datalad/local/tests/test_wtf.py
@@ -83,7 +83,7 @@ def test_wtf(topdir=None):
             if external_versions['psutil'] < '6.0.0':
                 # filesystems detail should be reported, unless 6.0.0 where
                 # it was removed. See https://github.com/giampaolo/psutil/issues/2109
-                assert_in('max_pathlength:', cmo.out)
+                assert_in('max_path_length:', cmo.out)
         else:
             assert_in("Hint: install psutil", cmo.out)
 


mih (approved):
Thank you!

yarikoptic:
I am yet to to double check when get to laptop, but are you sure? (Original issue doesn't show how test fails)

I believe I was also confused why we have it this way when worked on #7622 but we do have in **our** code it without underscore:

https://github.com/datalad/datalad/blob/35564bf98231a63010cdd5650849c19d89b6d375/datalad/local/wtf.py#L168

yarikoptic:
more details in https://github.com/datalad/datalad/issues/7627#issuecomment-2200231845 -- this fix is not needed and is not a fix.  It fixes through an accidental side effect that we do have max_path_length reported in another WTF section. See (with older psutil so we do see both):

```shell
❯ datalad wtf | grep max.*path
      - max_pathlength: 4096
      - max_pathlength: 4096
      - max_pathlength: 4096
  - max_path_length: 292
datalad wtf  5.81s user 3.08s system 221% cpu 4.022 total

```

PR #7630, opened 2024-07-02
Update affiliation of Matthias Riße in .zenodo.json

matrss:
While adding myself to the .zenodo.json file of the handbook in https://github.com/datalad-handbook/book/pull/1229 I noticed that my recorded affiliation here is outdated, since our institute was renamed starting yesterday. This is just a quick fix of that.

Hopefully the new name will stick for a while...

@adswa could you take a look here as well?

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -128,7 +128,7 @@
             "orcid": "0000-0002-4028-2087"
         },
         {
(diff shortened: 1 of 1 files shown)


adswa (approved):
thanks for the update!

PR #7633, opened 2024-07-17
INFRA: Add RTD-recommended configuration ahead of deprecations

adswa:
In an email to maintainers of affected projects, readthedocs advised to make configuration changes ahead of the deprecation of injecting Sphinx- related configurations during build time.
For more info, see: https://about.readthedocs.com/blog/2024/07/addons-by-default/

With this deprecation, Read the Docs will stop defining html_baseurl, and it has to be defined on its own to keep the configured canonical custom domain properly configured.

This change adds the recommended configuration change.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -263,6 +263,15 @@ html_show_sourcelink = False
 # Output file base name for HTML help builder.
 htmlhelp_basename = 'dataladdoc'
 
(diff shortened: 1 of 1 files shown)


yarikoptic:
Thank you! I guess we will see if all is good when we deploy! Preview seems built fine too

PR #6977, opened 2022-08-24
Add support for annex add --batch and use it if more than a single file (and not directories) is provided

yarikoptic:
Resurrects #5722 which could not be reopened since I force-pushed.

- Closes #5721.

## Summary

`AnnexRepo.add_()` now uses `git annex add --batch` when adding more than one file,
avoiding per-file `git annex add` process startup — which was previously the main
overhead on large `datalad save` invocations. Falls back to the non-batched
invocation when any of the paths is a directory, since `git annex add --batch`
does not (yet) descend into directories
(see [git-annex todo](https://git-annex.branchable.com/todo/make_add_--batch_add_directories_content/)).

Also observed to mitigate the OSX-specific
[performance regression on our tests](https://git-annex.branchable.com/bugs/significant_performance_regression_impacting_datal/#comment-4df4242f85746baef8afd32e0632c37c).

## Checklist

- [x] detected/reported an inconsistency against git-annex: https://git-annex.branchable.com/todo/make_add_--batch_add_directories_content/ . Worked around in the code by still using non-batched version if any path is a directory. Adds "checking overhead" but there is no other way ATM. Most likely not relevant much since higher level interfaces likely "analyze" folders and point to specific files.
- [x] remove TEMP `if True` to always try batch mode — restored intended `batch or (batch is None and len(files) > 1)` semantics, so single-file adds still take the plain (non-batched) path.
- [x] seek @christian-monch  wisdom on properly using BatchedCommand as a generator — resolved: current `proc1`-per-file loop faithfully reproduces the old `yield_` semantics, per [Christian's comment](https://github.com/datalad/datalad/pull/6977#discussion_r957200637). Removed the stale TODO/note about it.
- [x] see if any positive effect on OSX with bleeding edge annex in relation to [recent regression on our tests on OSX](https://git-annex.branchable.com/bugs/significant_performance_regression_impacting_datal/#comment-4df4242f85746baef8afd32e0632c37c) — yes, mitigates it, see git-annex issue.
- [x] no effect on benchmarks... if anything only `0.92  plugins.addurls.addurls1.time_addurls('*')` is happier.
- [x] added a scriv changelog fragment (`changelog.d/pr-6977.md`).
- [ ] `total_nbytes` (progress-bar accounting) not yet wired through in the batched path — left as a TODO in `_maybe_batched_add`.

## Test plan

[...]

PR #7636, opened 2024-07-24
Allow only one thread in S3 downloader's progress report callback 

christian-monch:
This PR should fix an issue with faulty `PROGRESS` commands that were sent from the "datalad" git annex special remote. This error was reported in INM7's "data" matrix channel. Here is an excerpt from the related error messages:

```
external special remote protocol error, unexpectedly received "PROGRESS PROGRESS 69730304" (unable to parse command)
unable to use special remote due to protocol error
external special remote protocol error, unexpectedly received " 61865984" (unable to parse command)
unable to use special remote due to protocol error]
```

The PR ensures that only one thread is "in" the progress bar callback that is used in S3 downloads. IMHO the issue was `boto3.download_fileobj()` (which is used to download S3 objects) can create multiple threads. Therefore the callback might be called by multiple threads without synchronisation.

### DIFF
--- a/changelog.d/pr-7636.md
+++ b/changelog.d/pr-7636.md
@@ -1,0 +1,3 @@
+### 🐛 Bug Fixes
+
+- Allow only one thread in S3 downloader's progress report callback.  [PR #7636](https://github.com/datalad/datalad/pull/7636) (by [@christian-monch](https://github.com/christian-monch))
--- a/datalad/downloaders/s3.py
+++ b/datalad/downloaders/s3.py
@@ -11,6 +11,7 @@
 
 import re
 from urllib.parse import urlsplit, unquote as urlunquote
+from threading import Lock
 
 import boto3
 import botocore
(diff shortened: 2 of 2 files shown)


christian-monch:
> Why are the three commits from #6977 included in this PR? They introduce quite a range of failures that can also be seen in the original draft PR, and I think they shouldn't contain relevant changes for the thread error, right?

That is a good question, I think I did not update my copy of `maint`. Let me update and rebase the PR.

mih (approved):
This LGTM. User feedback from the person that reported the underlying problem indicates that this change fixes it for them.

yarikoptic (approved):
Note: this is about boto3 potentially using multiple threads to download a single file/key, as described in https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/download_fileobj.html , and not somehow multiple threads of our own making (so this fix would be just hiding the issue in such a case).

yarikoptic-gitmate:
PR released in [`1.1.2`](https://github.com/datalad/datalad/releases/tag/1.1.2)

PR #7638, opened 2024-07-25
Adjust Gitea demo URL to reenable create-sibling-gitea test in datalad-next

adswa:
I want to reenable the gitea tests in datalad-next (https://github.com/datalad/datalad-next/issues/608). However, the demo Gitea instance moved from try.gitea.io to demo.gitea.com. I've migrated the ``dataladtester`` account, and saved the new token in the Appveyor project settings of datalad-next. Locally, tests pass with this change.


mih (approved):
YES!! Good catch!

mih:
I will merge this now and here. These tests are not running here, but they will be in -next.

See https://github.com/datalad/datalad-next/issues/608

PR #7648, opened 2024-08-29
Provide detail on why CHECKURL failed for datalad and archive special remotes (would require new AnnexRemote release (above 1.6.5) to take advantage of this PR)

yarikoptic:
Inspired by

- https://github.com/datalad/datalad-container/issues/267

but to be of benefit, requires review/merge/release of

- https://github.com/Lykos153/AnnexRemote/pull/106

But should be safe otherwise since ATM (without the above PR) AnnexRemote simply ignores the message


yarikoptic:
IMHO this one is "obvious" and there is no rush, so I will leave it up for a review for a week but then would merge unless objections come. Attn @mih and @datalad/developers

yarikoptic-gitmate:
PR released in [`1.1.4`](https://github.com/datalad/datalad/releases/tag/1.1.4)

PR #7655, opened 2024-09-06
OPT: use set O(log(n)) instead of list O(n) for checking if modified in _check_files

yarikoptic:
Likely impact is minimal since likely is not used for heavy lists of files, but who knows


yarikoptic:
- appveyour is #7657 

benchmarks say "no effect" besides

```
|          | 588±80ms                              | 731±80ms                   | ~1.24   | plugins.addurls.addurls1.time_addurls('*')    
```
which is odd, for starters -- restarted benchmark run to see if persists.

yarikoptic:
I expect this trivial change to be of possible benefit if invoked on heavy lists of files, so unless objections -- will merge in a few days.

yarikoptic:
trivial, no objections, should be good, let's proceed.

yarikoptic-gitmate:
PR released in [`1.1.4`](https://github.com/datalad/datalad/releases/tag/1.1.4)

Issue #7393, opened 2023-05-30
Circular import


mih:
```
❯ python -c 'from datalad.support.extensions import has_config'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/mih/hacking/datalad/git/datalad/support/extensions.py", line 12, in <module>
    from datalad.interface.common_cfg import (
  File "/home/mih/hacking/datalad/git/datalad/interface/common_cfg.py", line 33, in <module>
    from datalad.utils import on_windows
  File "/home/mih/hacking/datalad/git/datalad/utils.py", line 98, in <module>
    from datalad.consts import TIMESTAMP_FMT
  File "/home/mih/hacking/datalad/git/datalad/consts.py", line 52, in <module>
    from datalad.runner.gitrunner import GIT_SSH_COMMAND
  File "/home/mih/hacking/datalad/git/datalad/runner/__init__.py", line 12, in <module>
    from .coreprotocols import (
  File "/home/mih/hacking/datalad/git/datalad/runner/coreprotocols.py", line 14, in <module>
    from .protocol import WitlessProtocol
  File "/home/mih/hacking/datalad/git/datalad/runner/protocol.py", line 24, in <module>
    from datalad.utils import ensure_unicode
ImportError: cannot import name 'ensure_unicode' from partially initialized module 'datalad.utils' (most likely due to a circular import) (/home/mih/hacking/datalad/git/datalad/utils.py)
```

I believe it is unnecessary and not beneficial to have `GIT_SSH_COMMAND` be imported from runner code in `consts.py`.

adswa:
Note from the call: @yarikoptic suggested that there should be no imports in ``datalad/consts.py``

PR #7656, opened 2024-09-06
Remove import of GIT_SSH_COMMAND within .consts (import directly from datalad.runner.gitrunner)

yarikoptic:
It was added there for compatibility in 24d463109ececaa2e9633daa7445516330f01f73 when moved deeper in hierarchy.

I think it would be best for us to avoid deeper imports in this file and thus to avoid possible circular imports observed in the past. Ref: https://github.com/datalad/datalad/issues/7393

Even though it would "break" external API, I think it is unlikely someone relies on this in extensions (I could be proven wrong), and thus I think it should be ok to release in minor release.


yarikoptic:
if no objections, will proceed with merge by the end of the week.

yarikoptic-gitmate:
PR released in [`1.2.0`](https://github.com/datalad/datalad/releases/tag/1.2.0)

PR #7667, opened 2024-10-14
Resolve CI issues and upgrade to py 3.9 from 3.8

yarikoptic:
TODOs

- [ ] review current fails and ideally address as much as possible
- [ ] switch to 3.9 (might be a follow up PR)


yarikoptic:
superseded by
- #7682

PR #7669, opened 2024-10-14
Try to address git-annex/cabal installation: Use "sonoma" (2023) instead of "monterey" (2021) on Appveyor for OSX

yarikoptic:
ATM our appveyor on OSX times out while running tests. Actually for running tests it has barely 5 minutes left since otherwise it (brew?) seems to be taking care about installing git-annex judging from timings like

    [00:02:54] ==> Installing git-annex
    [00:03:02] ==> cabal v2-update^M
    [00:03:36] ==> cabal v2-install --flags=+S3^M
    [00:55:45] ==> Caveats
    [00:55:45] To start git-annex now and restart at login:
    [00:55:45]   brew services start git-annex
    [00:55:45] Or, if you don't want/need a background service you can just run:
    [00:55:45]   /usr/local/opt/git-annex/bin/git-annex assistant --autostart

I see that in datasalad @mih is using sonoma so decided to check if it would be the same.

FWIW it is also something "new" since it used to be quick as e.g. last month:

	datalad@smaug:/mnt/datasets/datalad/ci/logs/2024/09$ git grep '\] To start git-annex now and restart at login' | tail
	10/push/maint/f71ff4f/appveyor-9849-success/u6sj7gppun9h9jnw.txt:[00:00:45] To start git-annex now and restart at login:
	10/push/maint/f71ff4f/appveyor-9849-success/u6sj7gppun9h9jnw.txt:[00:00:49] To start git-annex now and restart at login:
	10/push/maint/f71ff4f/appveyor-9849-success/yvlyi46pkcy7ew9u.txt:[00:00:50] To start git-annex now and restart at login:
	10/push/maint/f71ff4f/appveyor-9849-success/yvlyi46pkcy7ew9u.txt:[00:00:53] To start git-annex now and restart at login:
	24/pr/7643/7e7ad48/appveyor-9851-success/4snf59wyahu701g1.txt:[00:00:53] To start git-annex now and restart at login:
	24/pr/7643/7e7ad48/appveyor-9851-success/4snf59wyahu701g1.txt:[00:00:57] To start git-annex now and restart at login:
	24/pr/7643/7e7ad48/appveyor-9851-success/idu7mku8x7s01ts3.txt:[00:00:54] To start git-annex now and restart at login:
	24/pr/7643/7e7ad48/appveyor-9851-success/idu7mku8x7s01ts3.txt:[00:00:59] To start git-annex now and restart at login:
[...]

PR #7673, opened 2024-10-15
Only run scheduled tests on GitHub for datalad/datalad

### LINKED ISSUE #7672: Do not run scheduled tests in forks
DataLad has a GitHub workflow called [test](/datalad/datalad/blob/maint/.github/workflows/test.yml) which runs on a daily schedule:

```
on:
  pull_request:
  push:
  schedule:
    - cron: '0 6 * * *'
```

Recently, while working on a PR, I pushed up-to-date `master` and `maint` branches to my fork ([mslw/datalad](/mslw/datalad)). I now noticed (from e-mail notifications) that this has caused the scheduled test to run daily in my fork, which seems like a waste of computing time.

GitHub docs [Disabling and enabling a workflow](https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/disabling-and-enabling-a-workflow) show how to disable a workflow manually, so no big deal. The docs also say that:

> When a public repository is forked, scheduled workflows are disabled by default. In a public repository, scheduled workflows are automatically disabled when no repository activity has occurred in 60 days.

So again, no big deal. However, in this case my fork is old enough to have been created before the scheduled GitHub workflow was added, so when I pushed the main branch, GitHub saw the workflow as new and enabled it (at least it would seem so).

Another way from preventing the workflow from running is suggested in this [StackOverflow Answer](https://stackoverflow.com/a/77895239/21665899) and documented in [Workflow syntax for GitHub Actions](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idif): a job can be configured to only run for specific repository and be skipped otherwise with e.g. `if: github.repository == 'octo-org/octo-repo-prod'`. However, it is not clear to me whether that would prevent the entire *workflow* from running, or just have it run and do nothing... There seems to be no way to set a conditional on workflow level.

PR incoming.

mslw:
This adds a repository conditional to the jobs in the "Test" GitHub workflow, which is triggered on schedule. As such, it should be disabled by default in forks, but e.g. manually pushing the main branch to a fork created before the action was added will cause it to become enabled.

Based on:
https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#example-only-run-job-for-specific-repository

See also:
https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/disabling-and-enabling-a-workflow

Closes #7672

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,6 +29,7 @@ env:
 
 jobs:
   filter:
+    if: ${{ github.repository == 'datalad/datalad' || github.event_name != 'schedule' }}
     runs-on: ubuntu-latest
     outputs:
       jobs: ${{ steps.jobs.outputs.matrix }}
@@ -49,6 +50,7 @@ jobs:
           echo 'EOT' >> "$GITHUB_OUTPUT"
 
   test:
+    if: ${{ github.repository == 'datalad/datalad' || github.event_name != 'schedule' }}
     runs-on: ubuntu-latest
     needs: filter
     strategy:


yarikoptic:
I don't see any of the Test runs at all like in  https://github.com/datalad/datalad/pull/7669:

image

yarikoptic:
ping @mslw on above.

yarikoptic:
I will move to draft, please move back to "Ready for Review" whenever.

mslw:
Hey, thanks for the ping.

I do not understand your comment @yarikoptic :

> I don't see any of the Test runs at all like in #7669:
> 
> image

To be clear, this PR was about making sure that the "Test" battery running on schedule ([last run in DataLad repo](https://github.com/datalad/datalad/actions/runs/12367578188)), i.e. the one shown below, does not trigger itself in forks. The change should not affect anything for PRs. Did it?

image

Anyway, this was just a suggestion. The underlying issue #7672 is highly situational (I unintentionally triggered the daily tests in my fork by pushing the main branch containing an updated YAML actions file to my fork; scheduled actions are by default disabled in fresh forks) and could be addressed by disabling manually in my fork. For this reason I will close the PR and the issue instead.

mslw:
Oh, now I understand: it *did* affect PR runs because I broke the workflow YAML file (I did not put `schedule` in quotes) - this is what was shown in the checks panel:

image

My bad, should have noticed.

yarikoptic:
we should still do this, right @mslw ?

yarikoptic:
@mslw ping

PR #7680, opened 2024-11-14
fix: remove unnecessary and confusing .path from error message

yarikoptic:
Leads to such messages as

    ❯ datalad get . -r
     [WARNING] Failed to get 'tests/data-for-tests.path', skipping this submodule

which are misleading


yarikoptic:
I don't think we need changelog for this minor one, and if anyone has objections -- we can discuss in a subsequent PR.

PR #7686, opened 2024-11-19
Add 3.13 testing/support

yarikoptic:
Debian has 3.13 already and my attempt to build 1.1.4 for Debian has failed. Hence had to look.

Changes of `CapturedException` as minimal and likely do not capture prior difference in use of `.__str__` vs `.__qualname__`. But I do not think there is anything we could do about that now that there is no `exc_type` stored. May be @bpoldrack would find a moment to have a look. 

Meanwhile I will run this version of the patch into debian package.

TODO
- [x] test_ProducerConsumer (interestingly -- always fails but differently and never seems to succeed "by chance")

### DIFF
--- a/datalad/support/exceptions.py
+++ b/datalad/support/exceptions.py
@@ -11,6 +11,7 @@
 
 import logging
 import re
+import sys
 import traceback
 from os import linesep
 from pathlib import Path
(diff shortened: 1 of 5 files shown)


yarikoptic:
eh, conda might not yet be ready for 3.13?  failed in many spectacular ways 

meanwhile, I have uploaded to debian with the proposed here patch to CapturedException -- at least all the tests passed

yarikoptic:
Woohoo, even codecov, albeit unhappy, is back. Let's proceed

yarikoptic:
Oops, changelog was forgotten

Issue #7681, opened 2024-11-14
test_subsuperdataset_save fails on osx github actions 


yarikoptic:
from https://github.com/datalad/datalad/actions/runs/11840062307/job/32992997932?pr=7679 for 
- https://github.com/datalad/datalad/pull/7679

which I do not think could have affected this

```
Versions: annexremote=1.6.6 boto3=1.35.60 cmd:7z=17.05 cmd:annex=10.20241031 cmd:bundled-git=UNKNOWN cmd:git=2.47.0 cmd:ssh=9.7p1 cmd:system-git=2.47.0 cmd:system-ssh=9.7p1 datalad=1.1.3+30.g1f5c60197 duecredit=0.10.2 humanize=4.11.0 iso8601=2.1.0 keyring=25.5.0 keyrings.alt=5.0.2 msgpack=1.1.0 platformdirs=4.3.6 requests=2.32.3
```

```

=================================== FAILURES ===================================
__________________________ test_subsuperdataset_save ___________________________
...
        assert_raises(CommandError, parent.save, 'sub1', recursive=True)
        # and should not fail in the top level superdataset
        with chpwd(parent.path):
            save('sub1')
        # or in a subdataset above the problematic one
        with chpwd(sub1.path):
>           save('sub2')

/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/datalad/core/local/tests/test_save.py:278: 
...

    def _check_result(self):
        if self.exception_on_error is True:
            if self.return_code not in (0, None):
                protocol = self.protocol
                decoded_output = {
                    source: protocol.fd_infos[fileno][1].decode(protocol.encoding)
                    for source, fileno in (
                        ("stdout", protocol.stdout_fileno),
                        ("stderr", protocol.stderr_fileno))
                    if protocol.fd_infos[fileno][1] is not None
                }
>               raise CommandError(
                    cmd=self.cmd,
                    code=self.return_code,
                    stdout=decoded_output.get("stdout", None),
                    stderr=decoded_output.get("stderr", None)
                )
E               datalad.runner.exception.CommandError: CommandError: 'git -c diff.ignoreSubmodules=none -c core.quotepath=false ls-files -z -m -d' failed with exitcode 128

/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/datalad/runner/nonasyncrunner.py:297: CommandError
```

yarikoptic:
[...]

PR #7687, opened 2024-11-20
test: xfail test_subsuperdataset_save on newer gits

yarikoptic:
The whole point for that "permissions" trick was to ensure that git does not go into submodules and thus we would be "optimized" and not waste cycles where not desired.  But the very fact that it errors out points that it is not the case.

So this provides only taming down for
  https://github.com/datalad/datalad/issues/7681
and for a proper solution we better first wait on what git gurus say in
  https://lore.kernel.org/git/ZzyyQpJw-kxJA7IS@bilena/T/#u


yarikoptic:
Appveyer unrelated

yarikoptic-gitmate:
PR released in [`1.1.5`](https://github.com/datalad/datalad/releases/tag/1.1.5)

PR #7701, opened 2025-02-28
[DATALAD RUNCMD] Switch to a more recent py39 based miniconda installation

yarikoptic:
Done with the hope to overcome issues with installation of py 3.13 which we seems to experience ATM


yarikoptic:
seems to be of no help... need more thoughts

PR #7724, opened 2025-05-18
Merge maint (1.1.6) into master

yarikoptic:
For some reason there are differences in acknowledgements etc. The only functional difference we have is removal of import in `consts`


yarikoptic:
ha ha -- this was the opposite -- the merge of `master` into `maint`, uff.  I think it is time to release it too.  I will push this merge also into `master` as well.

PR #7726, opened 2025-05-23
[ENH ] allow using submodule {path_basename} as props for url rewrite

bpinsard:
When setting-up online rewrite rule for git submodule URL to facilitate inter-operability of different hosting platforms that do or dont allow hierarchical naming of repos (eg. gitlab vs github/forgejo/ ), it can be useful to use the basename of the submodule path.
This PR simply adds that `{basename}` tag to the rewrite props.

### DIFF
--- a/changelog.d/pr-7726.md
+++ b/changelog.d/pr-7726.md
@@ -1,0 +1,3 @@
+### 🚀 Enhancements and New Features
+
+- [ENH ] allow using submodule {path_basename} as props for url rewrite.  [PR #7726](https://github.com/datalad/datalad/pull/7726) (by [@bpinsard](https://github.com/bpinsard))


datalad/distribution/get.py
```diff
@@ -195,6 +195,7 @@ def _get_flexible_source_candidates_for_submodule(ds, sm):
         for k, v in sm.items()
         if k.startswith('gitmodule_')
     }
+    sm_candidate_props['basename'] = op.basename(sm_candidate_props["path"])
```
yarikoptic, line 198: not that familiar with code/logic here:

- is `path` 100% guaranteed to exist?
- is `basename` guaranteed to not be defined already? anyways, sounds better to be `path_basename` to relate to `path`

would need
- a test to demonstrate/test
- tune up to documentation to reflect is availability

PR #7730, opened 2025-05-27
Remove absolute path specification from `/bin/ls` invocation while interrogating remote path

malikwirin:
`/bin/ls` is not available on every unix system like on mine (NixOS).
This should fix the compatibility issue

### DIFF
--- a/changelog.d/pr-7730.md
+++ b/changelog.d/pr-7730.md
@@ -1,0 +1,3 @@
+### 🐛 Bug Fixes
+
+- Remove absolute path specification from `/bin/ls` invocation while interrogating remote path.  [PR #7730](https://github.com/datalad/datalad/pull/7730) (by [@malikwirin](https://github.com/malikwirin))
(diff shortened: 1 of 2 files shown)


yarikoptic:
originally we had it just `ls` but then in e1eb959222b3f97699149914511e3b35455c7ff4 which was part of the https://github.com/datalad/datalad/pull/7265

```
    Attempt to shell-robustify create-sibling's ls call
...
-        out, err = ssh("LC_ALL=C ls -A1 {}".format(sh_quote(path)))
+        out, err = ssh('sh -c "LC_ALL=C; export LC_ALL; /bin/ls -A1 {}"'.format(
+            sh_quote(path)))
```

which isn't entirely clear to me which of the two changes (if any) - wrapping into `sh` or going from `ls` to `/bin/ls` robustified anything.  From looking at [this comment with error](https://github.com/datalad/datalad/pull/7265#issuecomment-1397181995) I think it was the former and thus should be ok to drop `/bin/`. It is very unlikely given the time has passed, but may be @adswa has some memory of this?

if not, given above analysis, I would be inclined indeed to just drop `/bin/` there.

yarikoptic:
Tests generally pass, and I do not see why full path must be needed, and original commit also did not give rationale.  So, let's merge and proceed and see how it goes

yarikoptic-gitmate:
PR released in [`1.2.2`](https://github.com/datalad/datalad/releases/tag/1.2.2)

PR #7731, opened 2025-05-27
BF: Prevent infinite recursion in clone when a remote points to itself

### LINKED ISSUE #7721: Datalad recurses infinitely if a remote has itself as origin
### What is the problem?

While doing some automated dataset handling, I accidentially set a git repo's remote to itself. This causes datalad to run in an infinite loop while trying to traverse all the local instances of the repo. 

The culprit is 

https://github.com/datalad/datalad/blob/9f581a6c57bc0af31ceab6a76066210ecccaee3a/datalad/core/distributed/clone.py#L890-L917

where in L915 there is no check whether `probeds.pathobj / origin_url` is already contained in `known_remote_urls`. Adding something like `if( not str(probeds.pathobj / origin_url) ):` before L909 should fix this problem I think.

```
datalad create a
datalad create b

cd b
git remote add origin "$PWD"
cd ../a
datalad clone -d . ../b b

## update(ok): . (dataset)
## update(ok): . (dataset)
## configure-sibling(ok): . (sibling)
## install(error): b (dataset) [maximum recursion depth exceeded] [maximum recursion depth exceeded]
```

### DataLad information

datalad 0+untagged.17342.g9f581a6
Linux

### Additional context

_No response_

_No response_

yarikoptic:
When cloning a repository that has its origin remote set to point to itself, DataLad would enter an infinite recursion loop in configure_origins().

This fix adds a check to prevent recursing into the same dataset path, avoiding the infinite loop reported in issue #7721.

Fixes #7721

🤖 Generated with [Claude Code](https://claude.ai/code)

Potentially there is other, trickier scenarios (chain of symlinks etc) but at least this would address an obvious issue.

### DIFF
--- a/changelog.d/pr-7731.md
+++ b/changelog.d/pr-7731.md
@@ -1,0 +1,3 @@
+### 🐛 Bug Fixes
+
+- BF: Prevent infinite recursion in clone when a remote points to itself.  Fixes [#7721](https://github.com/datalad/datalad/issues/7721) via [PR #7731](https://github.com/datalad/datalad/pull/7731) (by [@yarikoptic](https://github.com/yarikoptic))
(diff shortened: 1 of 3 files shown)


yarikoptic-gitmate:
PR released in [`1.2.1`](https://github.com/datalad/datalad/releases/tag/1.2.1)

PR #7736, opened 2025-06-15
releasing package datalad version 1.1.5-2.1

hoexter:
I'm uploading this NMU to Debian to allow datalad to be released with the upcoming stable release trixie.
This fixes Debian bug https://bugs.debian.org/1099382

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,11 @@
+datalad (1.1.5-2.1) unstable; urgency=medium
+
+  * Non-maintainer upload.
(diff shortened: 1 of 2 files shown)


yarikoptic:
Hi @hoexter , I have missed your PR and did push gbp import-dsc of your nmu. I am surprised that GitHub still shows diff here, but AFAIK this pr is no longer needed, see eg https://github.com/datalad/datalad/blob/debian/debian/changelog

Please reopen if I am wrong

PR #7737, opened 2025-07-02
fix: submodule property `datalad-url` used wrong submodule name

mih:
Instead of the relative path in POSIX conventions, it used the relative path in platform conventions. On windows this led to a dangling submodule property, detached from the actual submodule record, whenever a submodule was added in a 2nd-level subdirectory. This led to warnings from routines that parsed the submodule.

This change force-converts the name/path to POSIX conventions.


mih:
Rolled out this fix on a win10 laptop of a workshop attendee and it solved the issue.

adswa:
this has worked in practice during a workshop - thanks!

Issue #7624, opened 2024-06-28
28 tests fail on NFS on github (cron job)


yarikoptic:
Filed #7623 thinking that it is just a single test which was failing, but apparently the summary is 

```
= 28 failed, 1107 passed, 46 skipped, 3 xfailed, 4 xpassed, 9 warnings in 2609.13s (0:43:29) =
```

in the recent run on NFS https://github.com/datalad/datalad/actions/runs/9691677704/job/26743722926

<details>
<summary>full paste of fails summary</summary> 

```shell
FAILED ../datalad/core/distributed/tests/test_clone.py::test_no_ria_postclonecfg - datalad.support.exceptions.IncompleteResultsError: Command did not complete successfully. 1 failed:
[{'action': 'create-sibling-ria',
  'message': 'initremote failed.\n'
             'stdout: initremote teststore-storage \n'
             'failed\n'
             '\n'
             'stderr: git-annex: Dataset unavailable from RIA store.\n'
             'git-annex: initremote: 1 failed\n',
  'path': '/tmp/nfsmount/datalad_temp_tree_test_no_ria_postclonecfg7ku3o59m',
  'status': 'error',
  'type': 'dataset'}]
FAILED ../datalad/core/distributed/tests/test_push.py::test_publish_target_url - datalad.support.exceptions.IncompleteResultsError: Command did not complete successfully. 1 failed:
[{'action': 'update',
  'error_message': "CommandError: 'git -c diff.ignoreSubmodules=none -c "
                   'core.quotepath=false fetch --verbose --progress '
                   "--no-recurse-submodules --prune target' failed with "
                   'exitcode 128 under '
                   '/tmp/nfsmount/datalad_temp_tree_test_publish_target_urlui11ixqy '
                   "[err: 'fatal: repository "
                   "'http://127.0.0.1:34739/subdir/.git/' not found']",
  'exception': CommandError: 'git -c diff.ignoreSubmodules=none -c core.quotepath=false fetch --verbose --progress --no-recurse-submodules --prune target' failed with exitcode 128 under /tmp/nfsmount/datalad_temp_tree_test_publish_target_urlui11ixqy [err: 'fatal: repository 'http://127.0.0.1:34739/subdir/.git/' not found'] [update.py:__call__:322,gitrepo.py:fetch:1976,gitrepo.py:fetch_:1987,gitrepo.py:_fetch_push_helper:2161,runner.py:run:242],
[...]

PR #7749, opened 2025-09-25
TST: Allow for tests under NFS (cron run only) for x20 more time.

yarikoptic:
To mitigate at least some failures due to timeouts, as discovered/noted in

-  https://github.com/datalad/datalad/issues/7624


yarikoptic-gitmate:
PR released in [`1.2.2`](https://github.com/datalad/datalad/releases/tag/1.2.2)

PR #7751, opened 2025-09-27
chrore: Upgrade/switch back to ubuntu-latest for github CI

yarikoptic:
now that we should have neurodebian usable for it again, and have recent git-annex standalone


yarikoptic:
Every affected CI setup feels good!

Issue #7774, opened 2026-01-01
install error for pure git / no-annex repo using datalad clone / install


just-meng:
### What is the problem?

Cannot clone or install a dataset that is explicitly a no-annex repo. Git clone works fine, however is less convenient if I want to install the dataset as subdataset. It would be nice to do it simply with datalad clone or datalad install, just like any annex repo. @yarikoptic not sure if you were aware of this, but now you are!

```
meng@meng-MS-7D06 /m/Data> datalad clone git@gin.g-node.org:/jwu/code_process2p.git code
[INFO   ] Initialization prevented by .noannex file (remove the file to override)                                                                               
[INFO   ] git-annex: Not initialized. 
[INFO   ] init: 1 failed 
install(error): /mnt/Data/code (dataset) [CommandError: 'git -c diff.ignoreSubmodules=none -c core.quotepath=false annex init -c annex.dotfiles=true' failed with exitcode 1 under /mnt/Data/code] [CommandError: 'git -c diff.ignoreSubmodules=none -c core.quotepath=false annex init -c annex.dotfiles=true' failed with exitcode 1 under /mnt/Data/code]
```

### DataLad information

<details><summary>DataLad 1.2.3 WTF (configuration, credentials, datalad, dataset, dependencies, environment, extensions, git-annex, location, metadata.extractors, metadata.filters, metadata.indexers, python, system)</summary>

# WTF
## credentials 
  - keyring: 
    - active_backends: 
      - SecretService Keyring
      - libsecret Keyring
      - EncryptedKeyring with [PBKDF2] AES256.CFB v.1.0 at /home/meng/.local/share/python_keyring/crypted_pass.cfg
      - PlaintextKeyring with no encyption v.1.0 at /home/meng/.local/share/python_keyring/keyring_pass.cfg
    - config_file: /home/meng/.config/python_keyring/keyringrc.cfg
    - data_root: /home/meng/.local/share/python_keyring
## datalad 
  - version: 1.2.3
## dataset 
  - branches: 
    - git-annex@47c1193
    - master@19cafec
  - id: 0fc2f3fc-353a-46b4-b1fe-b4a8fbbcad47
  - path: /mnt/Data/Meng_Psychedelics
  - repo: AnnexRepo
## dependencies 
  - annexremote: 1.6.0
  - boto3: 1.40.31
  - cmd:7z: 16.02
  - cmd:annex: 10.20230407-1~ndall+1
  - cmd:bundled-git: 2.30.2
[...]

PR #7775, opened 2026-01-02
fix: clone - do not run  'git annex init'   if there is .noannex file

yarikoptic:
ATM we would run annex init if there is any sign that repository has git-annex in it, e.g. if there is some git-annex branch etc. I think that more correct would be to avoid running annex init if cloned branch has .noannex since then annex init would simply fail.  But since situation is not fully kosher and there might be some cleanup to be done (e.g. removal of git-annex branch), we issue a warning like

  [WARNING] Dataset(/tmp/code_process2p_noannex) seems to know about git-annex, but has .noannex file present. We skip running 'git annex init'.

- Closes https://github.com/datalad/datalad/issues/7774


yarikoptic:
mac fails on appveyor are unrelated, code change is otherwise minimal and shown to help

Issue #3896, opened 2019-11-27
Consolidate commit message specification


mih:
https://github.com/datalad/datalad/issues/3316 is asking for `-m/--message` for all commiting commands. There are quite a few, hence this would be a project of some size. https://github.com/datalad/datalad/pull/3883 already started, and https://github.com/datalad/datalad/pull/3893 suggests to add special meaning to a `-m` without an option. There is also: https://github.com/datalad/datalad/issues/3882

It would be good to thing to think this through a bit more and then make one consolidated effort (or at least a plan for such).

**Keep it minimal**: we have plenty of command that need message opts, but `save` as two (`-m` and `-F`). I propose to kill `-F` and replace it with a `file:` argument prefix detection than is then accessible to all commands

**Centralize**: Message file handling should move out of `save` and be extended to a  common helper that deal with any message handling. I think it is worth pondering whether we want a class/function that can be used internally to signal and handle interactive message input, like approached in #3893, possibly with templating for subdatasets as done in https://github.com/datalad/datalad-revolution/pull/71. But there is another templating aspect too: https://github.com/datalad/datalad/issues/2100

**What about tagging?**: https://github.com/datalad/datalad/issues/3371

yarikoptic:
Not sure about killing -F. On one hand indeed it is less if passing things around, but in the other - obfuscation. I wonder if we could resolve it programmatically. So both are present, but -F just acts via callback to the option specification itself and populates `message` option, (may be even somehow that so code itself doesn't need to even list it in the signature).

mih:
Hmm, you did not convince me that the implied obfuscation is more of a problem than the implied complexity of introducing two options to a large number of commands (instead of just one).

If you don't like `file:` prefix, we can make it a proper `file://` URL, which would impose that any message files need to be given via their absolute path. I found that annoying, but if that mitigates your concerns, so be it.

bpoldrack:
First-off, since this is labeled a question, I agree that message (file-) handling should be centralized and minimal in terms of API pollution, if we want to be able to specify messages for more commands than just `save`.

This would need a - possibly interactive - direct message spec as well as predefined messages from a file, templates, tags and a mechanism to deal with subdatasets in case of multiple commits.

Re `-F`:
I don't quite get the obfuscation aspect. What gets obfuscated?
[...]

Issue #3316, opened 2019-04-10
All "committing" commands should have -m|--message argument


yarikoptic:
My use case currently:
```shell
datalad create -d . -m "Dataset with additionally preprocessed files" preproc
```
in which case ideally we should commit in super dataset the addition of the newly created subdataset with that message.  Messages for commits within preproc should be "automated" ones I guess (thus possibly relates to #3236 (automated commit messages)) but I would not even mind if the last one (currently "[DATALAD] new dataset") would be the one I have provided.

For such multi-commit executions I even wonder if solution could/should be similar, or may be even the same in implementation, to #3265 (`run` to cover multiple commits).  The idea of the "merge" commit to depict the end of the single action  kinda make sense in such scenarios.  In `create` it is a bit tricky (start from an "empty state commit") but doable AFAIK.

Similarly  `install` which is in the case of installing a new subdatset (e.g. `install -d . URL`) would generate commit within super dataset.

P.S. I though we had such an issue filed but couldn't find it. May be we discussed and closed already ;-)

mih:
From my POV it has little benefit to implement this without something like https://github.com/datalad/datalad-revolution/pull/71 -- which we cannot do ATM.

I also keep thinking why we actually make `create` and `install` perform superdataset commits. It is nice for scenarios that are like our tests: do one thing, double check that it actually happened, run git status again to be sure... But in any real-world, non-developer use that I am aware of datasets are assembled in various steps that are typically concluded with a final `save` call -- the more complicated test setups we have seem to confirm that (but it may also be that I wrote them ;-). So all those intermediate commits are somewhat useless and slow things down.

So if `save --amend` would come and further tweaks and features of the underlying `git commit` would be exposed, all those commands with just `--message` would be crippled in contrast. I think I'd prefer to have `save` be the only command to do handling of dataset tree operations, and leave others within the scope of a single dataset.

yarikoptic:
> So if `save --amend` would come and further tweaks and features of the underlying `git commit` would be exposed, all those commands with just `--message` would be crippled in contrast.

not sure what/who gets crippled.  It is like `git commit` is crippled since there is `git commit --amend` to be.

> leave others within the scope of a single dataset.

not sure how it relates to the "no auto-`save`" besides that they already are, e.g. not calling save recursively if I get it right, but could cause a "cascade" of saves to save all the way to the super ds with `-S` (probably gone now) or `-d ^`.

[...]

PR #7785, opened 2026-01-15
Add --message option to clone, create, and update

asmacdo:
## Summary

Add `--message/-m` option to `clone`, `create`, and `update` commands, allowing users to specify custom commit messages when registering subdatasets in parent datasets.

- `clone -d . -m "message"  <source>` - customizes the "[DATALAD] Added subdataset" commit
- `create -d .  -m "message" <path>` - customizes the "[DATALAD] Created subdataset" commit
- `update -r --how=merge -m "message"` - customizes the "[DATALAD] Save updated subdatasets" commit

The message only affects the parent dataset's commit, not any commits within the subdataset itself.

## Related

This partially addresses #3316. That issue requests `--message` for all committing commands. This PR covers the three main subdataset registration cases (`clone`, `create`, `update`), but does not address other commands like `addurls`, `add_readme`, `no_annex`, or `subdatasets` which also have hardcoded commit messages.

It might also address #3896, though I'm not sure that there was consensus on that.

## TODO

- [x] Add `CHANGELOG-missing` label
- [x] Add semver label

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -146,6 +146,11 @@
             "affiliation": "Institute of Biophysics, University of Natural Resources and Life Sciences, Vienna, Austria",
             "name": "Lukas Schrangl",
             "orcid": "0000-0002-0420-7872"
+        },
+        {
+            "affiliation": "Dartmouth College, Hanover, NH, United States",
+            "name": "Macdonald, Austin",
+            "orcid": "0000-0002-8124-807X"
         }
     ],
     "grants": [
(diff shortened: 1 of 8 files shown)


yarikoptic (changes requested):
our tests battery already runs for long time. I think it would be much better to just extend some selected test(s) with such a test since you are testing just adding one more attribute! also that way you could see its effect on potentially more elaborate use cases (like hierarchical updates etc).

yarikoptic (approved):
looks good to me... should try as well...

PR #7787, opened 2026-01-15
ci: avoid running various test workflows on BOTH PR and branch (unless master and maint)

yarikoptic:
I just spotted that we are wasting cycles this way running on a branch (unmerged) version too


yarikoptic-gitmate:
PR released in [`1.3.0`](https://github.com/datalad/datalad/releases/tag/1.3.0)

PR #7798, opened 2026-01-30
Add typing_extensions to build system requirements (Python < 3.11)

lschr:
Currently, building datalad fails with Python 3.10 since `typing_extensions` is not part of `build-system.requires` in `pyproject.toml`,


lschr:
- Benchmark fails because master cannot be built (it would need this fix)
- CrippledFS fails due to a 503 HTTP error
- test (3.10, …) timed out.

So I think this is fine.

PR #7799, opened 2026-02-01
BF: Ensure annex-ignore config is immediately written for RIA stores

### LINKED ISSUE #5186: `datalad clone "ria+file:/some/store/#~myds"` leads to undesired establishment of an "annex"

yarikoptic:
The test_ria_postclone_noannex was failing because git-annex was auto-initializing the remote bare repository during clone, creating an annex/ directory that shouldn't exist in RIA stores.

The issue was that ds.config.set() might not immediately flush the config to disk, so git-annex couldn't see the annex-ignore setting when it auto-initialized remotes during "git annex init".

Fix by using git config directly to ensure the setting is written to .git/config immediately before git-annex-init runs.

- Fixes #5186 regression with newer git-annex versions.

from @yarikoptic: fix investigated/done by claude code. not 100% sure it is right. But hope is that should eliminate this sporadic fails on our testing of git-annex builds (of `FAILED ../datalad/core/distributed/tests/test_clone.py::test_ria_postclone_noannex - assert not True`) so if tests pass I merge and see how it goes

### DIFF
--- a/changelog.d/pr-7799.md
+++ b/changelog.d/pr-7799.md
@@ -1,0 +1,3 @@
+### 🐛 Bug Fixes
+
+- BF: Ensure annex-ignore config is immediately written for RIA stores.  Fixes [#5186](https://github.com/datalad/datalad/issues/5186) via [PR #7799](https://github.com/datalad/datalad/pull/7799) (by [@yarikoptic](https://github.com/yarikoptic))


yarikoptic-gitmate:
PR released in [`1.3.1`](https://github.com/datalad/datalad/releases/tag/1.3.1)

PR #7804, opened 2026-02-20
BF(xfail): Mark test_ria_postclone_noannex for git-annex >= 10.20260213

yarikoptic:
git-annex 10.20260213 reordered case matching in configRead (Remote/Git.hs) for "Push to Create" support, causing local remotes to bypass the annex-ignore check.  This auto-initializes annex on bare remotes despite annex-ignore=true being set.

https://git-annex.branchable.com/bugs/annex-ignore_check_is_skipped_for_local_remotes/

If not objections I will merge/release in a day to get builds of datalad/git-annex into greener land

attn to all @datalad/german who use RIA on a daily basis

### DIFF
--- a/datalad/core/distributed/tests/test_clone.py
+++ b/datalad/core/distributed/tests/test_clone.py
@@ -81,6 +81,7 @@ from datalad.tests.utils_pytest import (
     with_sameas_remote,
     with_tempfile,
     with_tree,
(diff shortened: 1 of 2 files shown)


yarikoptic:
looked at fails -- flaky clones from datasets. or gitea

PR #7811, opened 2026-03-02
ENH: config option to skip files open for writing during `datalad save`

yarikoptic:
Add datalad.save.skip-openfiles config (none/skip/warning/error) to detect and optionally exclude files held open for writing by other processes.  Uses psutil (added to misc optional deps) to enumerate open file descriptors and filter by write mode.

New module datalad.support.openfiles provides get_files_open_for_writing() which resolves symlinks and excludes the caller's own process tree. GitRepo._check_for_openfiles() integrates this into save_() following the same pattern as the existing Windows-compat check.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

Original problem aimed to be solved described in

- https://git-annex.branchable.com/bugs/assistant__58___nothing_added_to_commit_but_untracked/

and is needed for automating of data logistics for ReproNim's ReproStim project. Aiming to use as

    datalad -c datalad.save.skip-openfiles=skip save -m "Saving all recently captured files" \
    && git annex sync

Additionally it introduces a helper  `python -m datalad.support.openfiles .` which I was not sure we should pollute CLI ATM with but which could come handy when needed.

### PR checklist

- [x] by default - currently the same old behavior of committing everything, including open files.  
   - We could have changed to `warn` or even `error` to make operation more robust but it might be extra run time penalty and then require psutil to be core dependency. For now decided to keep consistent with current behavior

Thanks for contributing!

### DIFF
--- a/datalad/core/local/tests/test_save.py
+++ b/datalad/core/local/tests/test_save.py
@@ -1133,7 +1133,7 @@ def test_save_sub_trailing_sep_bf6547(path=None):
 
 def _mock_open_for_writing(open_file_abspath):
     """Return a mock for get_files_open_for_writing that reports *open_file_abspath*."""
-    def _fake(paths, exclude_pids=None):
+    def _fake(paths):
         result = {}
         for p in paths:
             if str(Path(p).resolve()) == str(Path(open_file_abspath).resolve()):
--- a/datalad/support/openfiles.py
+++ b/datalad/support/openfiles.py
@@ -38,7 +38,6 @@ lgr = logging.getLogger('datalad.support.openfiles')
 
 def get_files_open_for_writing(
     paths: list[str | Path],
-    exclude_pids: set[int] | None = None,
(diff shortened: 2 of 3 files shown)


yarikoptic:
"interesting". github in general in a bad shape ATM so tests might fail to access it but I did not expect this one in 3.13 run

```
docker: Error response from daemon: failed to set up container networking: driver failed programming external connectivity on endpoint datalad-tests-ssh2 (698223ae729fb8507e4e5bc34b98661024a8a1b58dfed29b46f17c2f44c5aa5f): failed to bind host port for 0.0.0.0:42242:172.17.0.3:22/tcp: address already in use

Run 'docker run --help' for more information
Error: Process completed with exit code 125.
```

yarikoptic:
ok, I reenabled this PR for review since we are fully green now! FWIW, I use prior version of this PR on reproiner to overcome git-annex limitations and so far so good!

PR #7832, opened 2026-03-21
BF: update patoolib imports for 2.0 API change and coverage-bin/datalad to arbitrary path to execute from

jkonieczny2:
* import check_new_filename from patoolib.fileutil after 2.0.0, or patoolib.util prior to that
* fixes code that calls archive_utils_patool; prior to this, check_new_filename could not be found

I noticed this issue when running tests after initial setup.  Pip retrieved the latest version of patool (4.0.4), which led to the import error.  You can reproduce with the following script:

```python
#!/usr/bin/env python
from datalad.support.archive_utils_patool import compress_files

if __name__ == "__main__":
    files = ["test.foo"]
    archive = "test.foo.gz"
    overwrite = False

    print("compressing files")

    compress_files(
        files,
        archive,
        overwrite=False,
    )
```

And confirm that it throws with patoolib >= 2.0 and without this change:

```
((venv) ) joshua@joshua-XPS-13-7390:~/projects/datalad$ ./test.py                                                
compressing files                                                                                                
  File "/home/joshua/projects/datalad/./test.py", line 11, in <module>                                           
    compress_files(                                     
  File "/home/joshua/projects/datalad/datalad/support/archive_utils_patool.py", line 220, in compress_files      
    check_new_filename(archive)                                                                                  
    ^^^^^^^^^^^^^^^^^^                                                                                           
NameError: name 'check_new_filename' is not defined
```

### DIFF
--- a/changelog.d/pr-7832.md
+++ b/changelog.d/pr-7832.md
@@ -1,0 +1,3 @@
+### 🐛 Bug Fixes
+
+- BF: update patoolib imports for 2.0 API change and coverage-bin/datalad to arbitrary path to execute from.  [PR #7832](https://github.com/datalad/datalad/pull/7832) (by [@jkonieczny2](https://github.com/jkonieczny2))
--- a/datalad/support/archive_utils_patool.py
+++ b/datalad/support/archive_utils_patool.py
@@ -18,7 +18,7 @@ assert(external_versions["patoolib"] >= "1.7")
 # check_new_filename moved from patoolib.util -> patoolib.fileutil in 2.0.0
 # still exists in patoolib.util as of 1.7; given the above assert we only
 # need to consider those two locations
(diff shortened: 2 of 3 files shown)


jkonieczny2:
@yarikoptic i rebased on origin/maint and removed the test_ria_git_remote.py changes; will see if the tests pass

yarikoptic-gitmate:
PR released in [`1.4.1`](https://github.com/datalad/datalad/releases/tag/1.4.1)

jkonieczny2:
@yarikoptic - in response to your question, i'm not a datalad user per se, but I was looking to contribute to open source and saw datalad on a dartmouth job board.

patoolib came up because of some failed tests while I was setting up my dev environment, so it seemed like a good issue to start with.

yarikoptic:
thanks for the detail!

PR #7841, opened 2026-04-11
chore: deduplicate code across interface docs, create-sibling, s3, and benchmarks

yarikoptic:
- Extract _strip_rst_roles() in interface/base.py, used by both alter_interface_docs_for_api() and alter_interface_docs_for_cmdline() to remove RST role markup and enhance handbook links.

- Add _is_repo_already_exists() hook to _GitHubLike base class so GitHub and Gitea override only the "already exists" detection (3 lines each) instead of the entire repo_create_response method (~40 lines each). GOGS/GIN inherit the base behavior unchanged.

- Extract _gen_versioned_bucket() in support/s3.py to consolidate repeated versioned S3 bucket setup (versioning, website config, public policy) across three gen_bucket_test*() functions.

- Extract StartupBenchmarks class in benchmarks/common.py with shared PATH setup, used by both benchmarks/cli.py and core.py.

### DIFF
--- a/benchmarks/cli.py
+++ b/benchmarks/cli.py
@@ -7,25 +7,15 @@
 # ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
 """Benchmarks for DataLad CLI"""
 
-import os
-import os.path as osp
-import sys
 from subprocess import call
 
-from .common import SuprocBenchmarks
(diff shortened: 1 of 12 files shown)


yarikoptic:
Every CI is happy and I think it looks good too, so let's proceed.

yarikoptic-gitmate:

PR #7846, opened 2026-04-20
Test patool in CI

asmacdo:
My tests were failing locally sometimes, but not others, even when CI was green. Tracked the issue down: the difference was whether or not `7z` was installed (it's installed in my dev container on Ubuntu, but not on my Fedora system).

The tests currently fail when 7z is not installed, but every job (IIUC) installs it, so CI misses these failures. 

We don't need a separate environment though, we should be able to hit our missing coverage with `DATALAD_RUNTIME_USE__PATOOL=1`. That's what this PR adds: one matrix leg (py3.10) with that env var set.

### What's broken, by patool version

| patool | Layer 1 (kwargs leak, real usage) | Layer 2 (`.lzma` fixture, tests only) |
|---|---|---|
| ≤ 2.x, 4.0.0, 4.0.1 | absent | present |
| 4.0.2+ | **present** | present |

- **Layer 1**: patool 4.0.2 added `interactive=True` to `run_checked`, which leaks through the monkey-patched `_patool_run` into `WitlessProtocol.__init__` and raises `TypeError`. That gets swallowed by a broad `except Exception` and surfaces as `PatoolError: returned non-zero exit status 100`. Affects any non-ZIP extraction.
- **Layer 2**: `create_tree` writes xz-format bytes into a `.lzma` test fixture, which patool's `extract_lzma` (`xz --format=lzma`) rejects. Test-only — real `.lzma` files are fine.

### Note

This PR is a draft because it doesn't fix the underlying issues. Its purpose is to surface them: the new leg will show red CI on every PR until either the patool backend is fixed or the `archive_utils_patool` fallback is retired.

### PR checklist

- [ ] Provide an overview of the changes you're making and explain why you're proposing them.
- [ ] Create a changelog snippet (add the `CHANGELOG-missing` label to this pull request in order to have a snippet generated from its title;
- [ ] Include `Fixes #NNN` somewhere in the description and `scriv` changelog entry if this PR addresses an existing issue.
- [ ] If this PR is not complete, select "Create Draft Pull Request" in the pull request button's menu.
  Consider using a task list (e.g., `- [ ] add tests ...`) to indicate remaining to-do items.
- [ ] If you would like to list yourself as a DataLad contributor and your name is not mentioned please modify .zenodo.json file.
- [ ] **Delete these instructions**. :-)

Thanks for contributing!

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -636,6 +636,17 @@
             "avatar_url": [
                 "https://avatars.githubusercontent.com/u/108482432?v=4"
             ]
+        },
+        {
+            "login": "jkonieczny2",
+            "name": "jkonieczny2",
+            "contributions": [
+                "code"
+            ],
+            "profile": "https://github.com/jkonieczny2",
+            "avatar_url": [
+                "https://avatars.githubusercontent.com/u/14842795?v=4"
+            ]
         }
     ],
     "contributorsPerLine": 7
--- a/.tributors
+++ b/.tributors
@@ -329,5 +329,9 @@
     "just-meng": {
         "name": "Jiameng Wu",
         "blog": "https://github.com/just-meng"
+    },
+    "jkonieczny2": {
+        "name": "jkonieczny2",
+        "blog": "https://github.com/jkonieczny2"
     }
 }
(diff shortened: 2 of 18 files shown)


yarikoptic:
note that there are two "lead" branches -- `maint` and `master`. Thus you might need to choose base appropriately. I would recommend to **always** review diff on what you are submitting before creating any (draft or ready for review) PR

asmacdo:
Indeed,  my bad.

PR #7848, opened 2026-04-21
BF: fix AltConstraints.__or__() and Constraints.__and__() mutating operands

### LINKED ISSUE #7164: `AltConstraint.__or__()` behavior hard to comprehend in edge cases
Here is the code

```py
    def __or__(self, other):
        if isinstance(other, AltConstraints):
            self.constraints.extend(other.constraints)
        else:
            self.constraints.append(other)
        return self
```

So if an `AltConstraint` is or'ed with an `AltConstrained` the first object is modified in-place. This is fine in general and helps to decomplexify the logic. However, if the second `AltConstraint` anyhow references the first one (e.g. `a=EnsureInt() | EnsureStr(); b=a | EnsureDType(a)`), this in-place modification leads to an infinite loop (e.g.  with `str(b)`), because `b` is now part of `a`

yarikoptic:
These operators mutated the left-hand operand in-place and returned `self`, causing silent side effects when constraint objects defined at module level (e.g. `reckless_opt`) were reused across multiple `|` or `&` chains.

For example:
    a = EnsureStr() | EnsureNone()  # AltConstraints([Str, None])
    b = a | EnsureInt()             # mutated a! a == b == [Str, None, Int]

Now both operators return a new instance, preserving the original.

This bug was identified in datalad-next's constraint system overhaul (https://github.com/datalad/datalad-next) where the constraint framework was redesigned from scratch.

- Closes #7164

### DIFF
--- a/changelog.d/20260421_073108_yaroslav_halchenko_altconstraints_or_mutation.md
+++ b/changelog.d/20260421_073108_yaroslav_halchenko_altconstraints_or_mutation.md
@@ -1,0 +1,11 @@
+### 🐛 Bug Fixes
+
+- Fix `AltConstraints.__or__()` and `Constraints.__and__()` mutating
+  their left-hand operand in place and returning `self`. This caused
+  silent side effects when constraint objects defined at module level
+  (e.g. `reckless_opt`) were reused across multiple `|` or `&` chains.
+  Both operators now return a new instance, preserving the original.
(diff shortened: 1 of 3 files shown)


yarikoptic:
@mih inviting you to cut the ribbon on a first PR to come to adopt pieces from datalad-next. Care to review?

PR #7851, opened 2026-05-04
BF(TST): mark test_gin_cloning flaky on IncompleteResultsError

yarikoptic:
The test clones https://gin.g-node.org/datalad/datalad-ci-target and fetches an annexed file. Both the initial git clone and the subsequent auto-enable of the gin remote for annex content are subject to transient gin.g-node.org reachability issues from CI runners.

Two distinct failure modes have been observed in the same datalad/git-annex nightly cron run on 2026-05-03 (all flavors using the freshly-built git-annex 10.20260421+git36-g70d04496e4):

- maint / release datalad: ``git clone`` itself times out at the TCP layer ("Failed to connect to gin.g-node.org port 443 after 136658 ms"), exhausting datalad's clone-candidate loop.
- master datalad: clone limps through, but the auto-enable HEAD probe fails, so git-annex sets ``remote.dl-test-remote.annex-ignore=true`` in the local config and the subsequent ``ds.get(annex_path)`` errors out with "(Note that these git remotes have annex-ignore set: dl-test-remote)".

Both surface as ``IncompleteResultsError``. The same git-annex binary passed on macOS-arm64 within the same cron window (05:21 UTC) and on the prior day's ubuntu cron, ruling out a deterministic regression and pointing squarely at gin reachability flakes.

Use the existing project pattern (see test_create_sibling_gitea, test_http, test_rerun_merges) of ``@pytest.mark.flaky(retries=N, only_on=[ExceptionType])`` to give the test up to two extra attempts on IncompleteResultsError. ``with_tempfile`` allocates a fresh path per invocation so retries do not collide with the prior attempt's half-cloned tree.

### PR checklist

- [ ] Create a changelog snippet (add the `CHANGELOG-missing` label to this pull request in order to have a snippet generated from its title;  I will not bother
- [ ] Include `Fixes #NNN` somewhere in the description and `scriv` changelog entry if this PR addresses an existing issue.

### DIFF
--- a/datalad/core/distributed/tests/test_clone.py
+++ b/datalad/core/distributed/tests/test_clone.py
@@ -1557,6 +1557,7 @@ def test_clone_unborn_head_sub(path=None):
 
 
 @skip_if_no_network
+@pytest.mark.flaky(retries=2, only_on=[IncompleteResultsError])
 @with_tempfile
 def test_gin_cloning(path=None):
     # can we clone a public ds anoynmously from gin and retrieve content
--- a/datalad/core/distributed/tests/test_clone.py
+++ b/datalad/core/distributed/tests/test_clone.py
@@ -1557,11 +1557,30 @@ def test_clone_unborn_head_sub(path=None):
 
 
 @skip_if_no_network
-@pytest.mark.flaky(retries=2, only_on=[IncompleteResultsError])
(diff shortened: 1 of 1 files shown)


yarikoptic:
appveyor has two runs hitting 1h wallclock limit, can't relate

PR #7856, opened 2026-05-07
BF(TST): probe filesystem capability in test_py2_unicode_command

yarikoptic:
## Summary

Make `test_py2_unicode_command` filesystem-capability-aware so it stops failing on user-provided CrippledFS tempdirs (e.g. vfat).

The existing skip is guarded only by `FILESYSTEM_SUPPORTS_UTF8`, which is `sys.getfilesystemencoding().lower() == 'utf-8'` — a process-level *encoding announcement*, not a per-mount probe. On vfat the OS-level locale is UTF-8, so the gate passes, yet the filesystem still rejects characters and trailing spaces; the test then fails with `OSError([Errno 22] Invalid argument)` on names like `' β1 '`. The legacy `os.environ.get('TMPDIR', None) != '/crippledfs'` workaround at the inner block only caught the historical AppVeyor path.

## Changes

- `datalad/tests/utils_pytest.py` — new `fs_supports_filename(tdir, name)` helper. Attempts `os.mkdir` on the candidate name under `tdir`, cleans up, and reports the boolean. Catches `(OSError, ValueError)`. Pattern parallels the inner probe in `get_most_obscure_supported_name`.
- `datalad/core/local/tests/test_run.py` — after the `FILESYSTEM_SUPPORTS_UTF8` fast-path skip, also skip when the actual tempdir rejects `bβ0.dat` (the simplest unicode name used). Replace the hardcoded `'/crippledfs'` check with `fs_supports_filename(path, u" β1 ")` — the worst-case trailing-space name that vfat actually refuses.
- `datalad/tests/test_tests_utils_pytest.py` — unit test for the new helper covering accept-and-cleanup and a universally-rejected NUL-byte name.

## Why two filename probes, not one

`bβ0.dat` and `' β1 '` are not equivalent: some CrippledFS variants accept the former and reject the latter (this is what was observed on the vfat scratch dir used during development). Probing each independently allows the outer test body to run when only the trailing-space subset is rejected — strictly more coverage than a blanket skip.

## Test plan

- [x] tmpfs (default `DATALAD_TESTS_TEMP_DIR`): full `test_run.py` + `test_save.py` — passes (no change).
- [x] vfat (`DATALAD_TESTS_TEMP_DIR=<vfat mount>`): `test_py2_unicode_command` no longer fails; outer assertions run, inner trailing-space block correctly skipped.
- [x] NFS (`DATALAD_TESTS_TEMP_DIR=<nfs mount>`): full suite passes (no change).

No production code is touched — this is a test-infrastructure fix.

Since a minor internal switch I would not even bother with the changelog

### DIFF
--- a/datalad/core/local/tests/test_run.py
+++ b/datalad/core/local/tests/test_run.py
@@ -57,6 +57,7 @@ from datalad.tests.utils_pytest import (
     assert_status,
     create_tree,
     eq_,
+    fs_supports_filename,
     known_failure_windows,
     neq_,
     ok_,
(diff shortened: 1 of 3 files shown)


yarikoptic:
some appveyors keep timing out

image

Issue #7867, opened 2026-05-27
`-J auto` is either broken or resolves to 1 across all commands


just-meng:
### Expected behavior

`-J auto` is documented as automatically determining the number of parallel jobs:

> how many parallel jobs (where possible) to use. "auto" corresponds to the number defined by 'datalad.runtime.max-annex-jobs' configuration item

A user passing `-J auto` expects DataLad to pick a reasonable degree of parallelism. Instead, some commands crash, some silently fail, and the rest resolve `auto` to 1 — making `-J auto` indistinguishable from omitting `-J` entirely. Not a single command provides automatic parallelism with `-J auto`.

### Commands that crash or silently fail

`push`, `drop`, and `remove` embed `--jobs auto` directly into the git-annex command line without resolving `auto` to an integer. Git-annex does not accept `auto` as a value for `--jobs`.

**`push`** — crashes with a visible `CommandError`:

```
$ datalad push --to origin --jobs=auto
CommandError: 'git ... annex copy --batch -z --to origin --jobs auto ...'
  failed with exitcode 1
option --jobs: cannot parse value `auto'
```

The string `auto` is passed verbatim to git-annex because `push` manually does `cmd.extend(['--jobs', str(jobs)])` before calling `_call_annex_records(cmd)` without `jobs=` ([push.py:854-855, 901](https://github.com/datalad/datalad/blob/maint/datalad/core/distributed/push.py#L854-L855)). This bypasses the resolution logic in `_call_annex()` ([annexrepo.py:958-968](https://github.com/datalad/datalad/blob/maint/datalad/support/annexrepo.py#L958-L968)).

**`drop`** — git-annex crashes identically, but the `CommandError` is caught and silently swallowed ([drop.py:801-806](https://github.com/datalad/datalad/blob/maint/datalad/distributed/drop.py#L801-L806)). DataLad exits 0, reports no results, and **content is not dropped**. Same pattern in both `_drop_allkeys()` ([drop.py:760-764](https://github.com/datalad/datalad/blob/maint/datalad/distributed/drop.py#L760-L764)) and `_drop_files()` ([drop.py:792-799](https://github.com/datalad/datalad/blob/maint/datalad/distributed/drop.py#L792-L799)).

**`remove`** — delegates to `drop` ([remove.py:212-219](https://github.com/datalad/datalad/blob/maint/datalad/local/remove.py#L212-L219)), inheriting the silent failure.

### Commands that work but resolve `auto` to 1

[...]

PR #7868, opened 2026-05-28
BF: resolve -J auto for push/drop/remove instead of passing 'auto' to git-annex

yarikoptic:
## Summary

`push --jobs=auto` crashed with `option --jobs: cannot parse value 'auto'`, and `drop --jobs=auto` (plus `remove --jobs=auto`, which delegates to drop) silently swallowed the same `CommandError` and left content untouched.

Root cause: `_push_data`, `_drop_allkeys`, and `_drop_files` each appended `--jobs <jobs>` to the git-annex command manually before calling `_call_annex_records[_items_]` **without** `jobs=`, bypassing the `'auto'`-to-integer resolution that already lives in `AnnexRepo._call_annex` (annexrepo.py:958-968).

Fix: drop the three manual `cmd.extend(['--jobs', str(jobs)])` blocks; forward `jobs=jobs` to the `_call_annex_records*` calls so resolution happens uniformly in one place.

Refs #7867. **This PR only addresses the literal-`'auto'`-passed-verbatim half of the issue.** The other half — that `'auto'` resolves to `1` by default because `datalad.runtime.max-annex-jobs` / `datalad.runtime.max-jobs` default to `1`, making `-J auto` indistinguishable from omitting `-J` — is left for a follow-up.

## Commits

- 88e0c7665 `BF: resolve -J auto for push/drop/remove instead of passing 'auto' to git-annex`

<details><summary>Per-commit details</summary>

### 88e0c7665 `BF: resolve -J auto for push/drop/remove`

Three call sites built the git-annex command line themselves:

- `datalad/core/distributed/push.py:_push_data` — `cmd.extend(['--jobs', str(jobs)])` then `_call_annex_records(cmd)` with no `jobs=`. Crashed with a visible `CommandError`.
- `datalad/distributed/drop.py:_drop_allkeys` and `_drop_files` — same pattern, but each is wrapped in `except CommandError` that pulls captured results from `e.kwargs.get('stdout_json', [])`. With `'auto'` git-annex never produces any JSON before erroring, so the exception is caught and `drop` returns no results. `remove` inherits the bug via `Drop.__call__(jobs=jobs)`.

The fix is to delete the manual `cmd.extend` and pass `jobs=jobs` to `_call_annex_records` / `_call_annex_records_items_`. The existing logic at `annexrepo.py:958-968` resolves `'auto'` → `min(datalad.runtime.max-annex-jobs, max(3, cpu_count()))` and emits `-J<int>` only when the result is > 1.

Tests: tuned the existing `test_push.check_push`, `test_drop.test_drop_file_content`, and `test_drop.test_drop_allkeys` calls from `jobs=2` to `jobs='auto'`. Verified that with the manual handling restored, `test_push[True]` raises `CommandError` and `test_drop_allkeys` fails its post-drop key-count check, so the change provides regression coverage for both the crash and the silent-failure paths.

</details>

<details><summary>Manual reproduction (before vs after)</summary>

```text
$ datalad push --to=target --jobs=auto         # before
CommandError: 'git ... annex copy --batch -z --to target --jobs auto ...'
  failed with exitcode 1
option --jobs: cannot parse value `auto'

$ datalad drop --jobs=auto .                    # before
(no output, exit 0, files NOT dropped)
```

After:

```text
$ datalad push --to=target --jobs=auto
copy(ok): a_1.bin (file) [to target...]
copy(ok): a_2.bin (file) [to target...]
...
$ datalad drop --jobs=auto .
drop(ok): a_1.bin (file)
drop(ok): a_2.bin (file)
...
```

</details>

## Test plan

- [x] `pytest datalad/core/distributed/tests/test_push.py` — 19 passed, 1 skipped
- [x] `pytest datalad/distributed/tests/test_drop.py` — 16 passed
- [x] `pytest datalad/local/tests/test_remove.py` — 9 passed
- [x] Reverified with the patch stashed: `test_push[True]` raises `CommandError`, `test_drop_allkeys` fails its key-count assertion → modified tests do provide regression coverage.
- [ ] Follow-up PR for the `max-annex-jobs` / `max-jobs` default-of-1 half of #7867.

### DIFF
--- a/changelog.d/pr-7868.md
+++ b/changelog.d/pr-7868.md
@@ -1,0 +1,3 @@
+### 🐛 Bug Fixes
+
+- Resolve `--jobs auto` for `push`, `drop`, and `remove` so the literal string `auto` is no longer passed through to `git annex`.  Previously `push --jobs=auto` crashed with `option --jobs: cannot parse value 'auto'`, while `drop --jobs=auto` (and `remove --jobs=auto`, which delegates to drop) swallowed the same `CommandError` and silently left content in place.  Fixes [#7867](https://github.com/datalad/datalad/issues/7867).  [PR #7868](https://github.com/datalad/datalad/pull/7868) (by [@yarikoptic](https://github.com/yarikoptic))
--- a/datalad/core/distributed/push.py
+++ b/datalad/core/distributed/push.py
@@ -859,9 +859,6 @@ def _push_data(ds, target, content, data, force, jobs, res_kwargs,
 
     cmd = ['copy', '--batch', '-z', '--to', target]
 
-    if jobs:
-        cmd.extend(['--jobs', str(jobs)])
-
     # Since we got here - we already have some  data != "nothing"
     if (data == 'auto') or \
         (
@@ -908,6 +905,7 @@ def _push_data(ds, target, content, data, force, jobs, res_kwargs,
     # and go
     res = ds_repo._call_annex_records(
         cmd,
+        jobs=jobs,
         git_options=[
             "-c",
             "annex.retry={}".format(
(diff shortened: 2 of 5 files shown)


yarikoptic:
appveyor - windows 1 run timeout on 1hr, I do not think related

yarikoptic-gitmate:

PR #7871, opened 2026-05-31
NF: add -J cpus option for explicit CPU-based parallelism

just-meng:
## Summary

Following #7868 (which fixed `-J auto` crashing push/drop), this adds `cpus` as a new `-J`/`--jobs` value that resolves to the number of CPUs available to the process.

**Motivation:** `-J auto` is intentionally conservative — it's capped by `datalad.runtime.max-annex-jobs`, which defaults to `1` (see #4404, where unbounded CPU-count parallelism caused crashes). With the default config `auto` therefore resolves to `1` (effectively serial). Users who *want* CPU-based parallelism currently have no way to request it without knowing their CPU count and passing a literal number.

`-J cpus` fills this gap: an explicit opt-in to CPU-count-based parallelism, without changing the safe `auto` default.

**Safety note:** unlike `auto`, `cpus` is deliberately *not* bounded by `datalad.runtime.max-annex-jobs`. On many-core machines this can be resource-intensive and, for git-annex operations, can reach the regime that #4404 guarded against. This trade-off is documented in the `-J` help text so inexperienced users are warned.

**Related future work (not in this PR):** arguably `-J auto` should itself resolve to something `> 1` but below a sensible cap, instead of the current default of `1`. That is a larger change to the default-parallelism policy and is left to the maintainers; `cpus` gives the explicit "use all available CPUs" knob in the meantime.

## Changes

| File | Change |
|------|--------|
| `utils.py` | Add `available_cpu_count()`: uses `os.sched_getaffinity()` on Linux (respects cgroups/containers), falls back to `os.cpu_count()` on macOS/Windows, `1` if undeterminable |
| `common_opts.py` | Add `'cpus'` to `EnsureChoice` for `jobs_opt`; document it and warn that it is not bounded by `max-annex-jobs` |
| `annexrepo.py` | Handle `'cpus'` → `available_cpu_count()` in `_call_annex()`; also switch the existing `auto` path from `multiprocessing.cpu_count()` to `available_cpu_count()` (no behavior change under the default `max-annex-jobs=1`) |
| `parallel.py` | Handle `'cpus'` in `ProducerConsumer.get_effective_jobs()` |
| `tests/test_jobs.py` | Tests for get/push/drop with `jobs='cpus'`, plus unit tests for resolution and `available_cpu_count()` |
| `changelog.d/` | scriv fragment under "🚀 Enhancements and New Features" |

## Test plan

- [x] `test_jobs_cpus_get` — get with `jobs='cpus'` is accepted and runs
- [x] `test_jobs_cpus_push` — push with `jobs='cpus'` works
- [x] `test_jobs_cpus_drop` — drop with `jobs='cpus'` works
- [x] `test_jobs_cpus_resolves_to_cpu_count` — resolves to `available_cpu_count()` in the `ProducerConsumer` path
- [x] `test_available_cpu_count` — returns a positive integer
- [x] Existing push and drop tests pass without regressions

### Manual end-to-end verification

Tested in a real dataset (16 available CPUs) — `-J cpus` resolves to `-J16` and is passed through to git-annex, completing successfully:

```
$ datalad -l debug get . -J cpus 2>&1 | grep -E -- "-J[0-9]+"
[DEBUG] Run ['git', ..., 'annex', 'get', ..., '-J16', '--', '.'] (protocol_class=AnnexJsonProtocol) (cwd=.../figures)
[DEBUG] Finished ['git', ..., 'annex', 'get', ..., '-J16', '--', '.'] with status 0
```

Ref: #7867

### DIFF
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -19,7 +19,6 @@ import re
 import time
 import warnings
 from itertools import chain
-from multiprocessing import cpu_count
 from os import linesep
 from os.path import (
     curdir,
@@ -70,6 +69,7 @@ from datalad.utils import (
     Path,
     PurePosixPath,
     auto_repr,
+    available_cpu_count,
     bytes2human,
     ensure_list,
     on_windows,
@@ -957,13 +957,19 @@ class AnnexRepo(GitRepo, RepoInterface):
         if self._annex_common_options:
             cmd += self._annex_common_options
 
-        if jobs == 'auto':
+        if jobs == 'cpus':
+            jobs = available_cpu_count()
+            if jobs == 1:
+                lgr.warning(
+                    "'-J cpus' resolved to a single CPU available to this "
+                    "process; proceeding without parallelism")
+        elif jobs == 'auto':
             # Limit to # of CPUs (but at least 3 to start with)
             # and also an additional config constraint (by default 1
             # due to https://github.com/datalad/datalad/issues/4404)
             jobs = self._n_auto_jobs or min(
                 self.config.obtain('datalad.runtime.max-annex-jobs'),
-                max(3, cpu_count()))
+                max(3, available_cpu_count()))
             # cache result to avoid repeated calls to cpu_count()
             self._n_auto_jobs = jobs
         if jobs and jobs != 1:
--- a/datalad/support/annexrepo.py
+++ b/datalad/support/annexrepo.py
@@ -959,10 +959,6 @@ class AnnexRepo(GitRepo, RepoInterface):
 
         if jobs == 'cpus':
             jobs = available_cpu_count()
-            if jobs == 1:
-                lgr.warning(
-                    "'-J cpus' resolved to a single CPU available to this "
-                    "process; proceeding without parallelism")
         elif jobs == 'auto':
             # Limit to # of CPUs (but at least 3 to start with)
             # and also an additional config constraint (by default 1
--- a/datalad/support/parallel.py
+++ b/datalad/support/parallel.py
@@ -28,7 +28,10 @@ from threading import Thread
 from datalad.support.exceptions import CapturedException
 
 from ..log import log_progress
-from ..utils import path_is_subpath
+from ..utils import (
+    available_cpu_count,
+    path_is_subpath,
+)
 from . import ansi_colors as colors
 
 lgr = logging.getLogger('datalad.parallel')
@@ -248,7 +251,13 @@ class ProducerConsumer:
         It will account for configuration variable ('datalad.runtime.max-jobs') and possible
         other requirements (such as version of Python).
         """
-        if jobs in (None, "auto"):
+        if jobs == "cpus":
+            jobs = available_cpu_count()
+            if jobs == 1:
+                lgr.warning(
+                    "'-J cpus' resolved to a single CPU available to this "
+                    "process; proceeding without parallelism")
+        elif jobs in (None, "auto"):
             from datalad import cfg
 
             # ATM there is no "auto" for this operation, so in both auto and None
--- a/datalad/support/parallel.py
+++ b/datalad/support/parallel.py
@@ -253,10 +253,6 @@ class ProducerConsumer:
         """
         if jobs == "cpus":
             jobs = available_cpu_count()
-            if jobs == 1:
-                lgr.warning(
-                    "'-J cpus' resolved to a single CPU available to this "
-                    "process; proceeding without parallelism")
         elif jobs in (None, "auto"):
             from datalad import cfg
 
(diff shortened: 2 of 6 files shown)


datalad/tests/test_jobs.py
```diff
@@ -0,0 +1,120 @@
...
+        'receive.denyCurrentBranch', 'updateInstead', scope='local')
+    src.siblings('add', name='target', url=dstpath,
+                 result_renderer='disabled')
+
+    res = src.push(to='target', jobs='cpus')
+    assert_in_results(res, action='copy', status='ok',
+                      path=str(src.pathobj / 'file.txt'))
+
+
+@skip_if_adjusted_branch  # drop content semantics differ on Windows adjusted branches
```
yarikoptic, line 64: apparently I forgot to add to CONTRIBUTING.md about need 

```
/home/yoh/.claude/CLAUDE.md:- When adding tests generated by AI assistants to projects that use pytest, ALWAYS mark them with `@pytest.mark.ai_generated`:
/home/yoh/.claude/CLAUDE.md-
/home/yoh/.claude/CLAUDE.md-```python
/home/yoh/.claude/CLAUDE.md:@pytest.mark.ai_generated
/home/yoh/.claude/CLAUDE.md-async def test_my_new_feature() -> None:
/home/yoh/.claude/CLAUDE.md-    """Test description."""
/home/yoh/.claude/CLAUDE.md-    # test code
```

@copilot send a short PR on that !

datalad/support/annexrepo.py
```diff
@@ -955,13 +955,19 @@ def _call_annex(self, args, files=None, jobs=None, protocol=StdOutErrCapture,
         if self._annex_common_options:
             cmd += self._annex_common_options
 
-        if jobs == 'auto':
+        if jobs == 'cpus':
+            jobs = available_cpu_count()
+            if jobs == 1:
+                lgr.warning(
+                    "'-J cpus' resolved to a single CPU available to this "
+                    "process; proceeding without parallelism")
```
yarikoptic: not sure if worth a warning

datalad/utils.py
```diff
@@ -2858,3 +2858,24 @@ def ensure_write_permission(path: Path) -> Iterator[None]:
...
+    cpuset restrictions (e.g. ``taskset``, ``--cpuset-cpus``, and many HPC
+    job schedulers).  Note that this does not account for cgroup CPU
+    *quota* limits (e.g. Docker ``--cpus``), which throttle CPU time
+    rather than restrict which cores are available.  On macOS/Windows,
+    falls back to ``os.cpu_count``.  Returns 1 if the count cannot be
+    determined.
+    """
+    try:
+        return len(os.sched_getaffinity(0))
+    except (AttributeError, NotImplementedError):
```
yarikoptic: add logging at least at .debug level so we do not completely silently react here.

Also should be coded lazily IMHO then since no need to re-check on every call - not to be changed, some private global var should suffice I think

just-meng:
Done — added debug log + lazy caching in  available_cpu_count; marked the tests @pytest.mark.ai_generated

datalad/support/annexrepo.py
just-meng: It took me a while to understant that `auto` was not doing what I thought it'd do. So whenever the program behaves differently than what the user would expect, I'd personally want to be notified.

yarikoptic: I get you. Let me rephrase -- is it really abnormal and worth such complaint on a real system with a single CPU?  Even those might be uncommon -- they well might be present in virtual systems etc.  I think it is at most a DEBUG level message .

"You" (claude?) made it "Returns 1 if the count cannot be determined." , so it does mask/collide with legit cases of 1 cpu setup.  If you want -- can make it return `None` and do explicit handling, or do such warning issuing inside that function when you can't determine number of CPUs, and then state that can't determine that.  you already have `lgr.debug(` -- so change that one to the warning, and remove here

datalad/support/parallel.py
```diff
@@ -248,7 +251,13 @@ def get_effective_jobs(cls, jobs):
         It will account for configuration variable ('datalad.runtime.max-jobs') and possible
         other requirements (such as version of Python).
         """
-        if jobs in (None, "auto"):
+        if jobs == "cpus":
+            jobs = available_cpu_count()
+            if jobs == 1:
+                lgr.warning(
+                    "'-J cpus' resolved to a single CPU available to this "
+                    "process; proceeding without parallelism")
```
yarikoptic: one more case where warning must go away... centralizing also would remove potentially duplicate warnings issuing! (helper function uses global to memoize)

datalad/support/annexrepo.py
just-meng: Good point on both. Centralized warning, and a real single CPU case no longer triggers a warning.

datalad/support/parallel.py
just-meng: done

PR #7874, opened 2026-06-03
BF: make update --how=reset/merge work on adjusted branches (Windows)

just-meng:
## Make `update --how=reset` / `--how=merge` work on adjusted branches (Windows)

Fixes #7772 (resurrection / zombie `synced/`) and #7873 (B1 — "Could not determine update target"). Partial fix for #7872 — this is the `update`-side half; the dedicated `datalad reset` command is the remaining half and lands in a follow-up PR.

### Motivation

On crippled filesystems (Windows), DataLad uses **adjusted branches**. Three distinct defects made "discard local divergence / match the sibling" unreliable there — the exact workflow needed when helping Windows collaborators clean up and re-sync a repo.

### The three fixes (each its own RED→GREEN commit pair)

- **B1 — reset wouldn't run.** `_choose_update_target` built the update target from the *adjusted* branch name (e.g. `<remote>/adjusted/master(unlocked)`), which never exists, so `update --how=reset` aborted with *"Could not determine update target"*. It now resolves the corresponding branch first (no-op on non-adjusted branches).

- **B3 — reset left a zombie behind.** `_annex_reset_target` reset the corresponding branch but left `synced/<branch>` at the pre-reset state, so a later `git annex sync` (run internally by `save`/`push`) merged it back, resurrecting the discarded commits. It now force-points `synced/<branch>` to the reset target.

- **B2 — merge seeded the zombie.** `update --how=merge` ran `git annex sync` but, unlike `AnnexRepo.localsync`, never removed the `synced/<branch>` it created. It now applies the same delete-if-created hygiene (using `branch -D`, since after a `--pull` sync the ref can be unreachable from the adjusted HEAD).

B3 alone makes the resurrection regression tests pass; B2 additionally prevents the zombie at its source, protecting discard paths that don't go through `update --how=reset`.

### Tests

- `test_update_reset_adjusted` — reset runs and lands at the sibling (parametrized: local vs. remote divergence).
- `test_update_reset_adjusted_resets_synced` — `_annex_reset_target` leaves the corresponding branch *and* `synced/<branch>` at the target, surviving a real `git annex sync`.
- `test_update_reset_adjusted_no_resurrection_on_save` / `…_on_push` — a later `save`/`push` does not resurrect the discarded commit (and keeps fresh work).
- `test_update_merge_adjusted_no_leftover_synced` — `update --how=merge` leaves no `synced/<branch>` behind.
[...]

Issue #7872, opened 2026-06-01
Provide an adjusted-branch-safe datalad reset (discard local divergence / adopt a remote rewrite)


just-meng:
## TL;DR

On adjusted branches (the default on Windows / crippled filesystems) there is **no
working DataLad way to discard local divergence and match a sibling**: `datalad update
--how=reset` errors out, `--how=merge` cannot adopt a rewritten history, and the manual
git + git-annex workaround is a real pain in the ass — multi-step, easy to get subtly
wrong, and run by hand every session (#7782). I'd like to add a **`datalad reset`**
command (initially `--hard`) that does the right thing on both normal and adjusted
branches, and have `update --how=reset` delegate to it. **Fixing only `update
--how=reset` is not sufficient** (see below). I have working fixes + tests and would like
a go-ahead.

This would resolve the earlier reports **#7772** (zombie/resurrected commits) and
**#7782** (feature request: enable `update --how=reset` on Windows — which lists exactly
the manual workaround above).

## Why it matters

DataLad exists so people don't have to be git + git-annex experts — but on adjusted
branches today, they do. In a long-running cross-platform collaboration, getting Windows
colleagues' clones cleanly back in sync with the remote is recurring, time-consuming
friction: the manual workaround (switch to the corresponding branch, hard-reset, fix the
`synced/` ref, re-adjust), by hand, every session. It's the kind of papercut that makes a
team quietly give up on reproducibility.

And it's worth fixing because it *works*: despite this friction, the collaboration just
produced a **micropublication backed by a fully reproducible DataLad pipeline** — source
data → final figure — with completed runs on **both Linux and Windows** (check it out [here](https://gin.g-node.org/Las-Ninas/CFDHW_Peru)). Smoothing the
adjusted-branch reset path is most of what stands between that and being effortless for
the next cross-platform team.

## Root cause — where each layer falls short on adjusted branches

(Verified on git-annex 10.2026 + datalad 1.4.1.)

| layer | "discard my local commit, match origin" | "team rewrote history, adopt it" |
|-------|------------------------------------------|----------------------------------|
| **git** | `reset --hard` hits the *adjusted view*, not the corresponding branch → nothing actually discarded; git has no re-adjust step and is adjusted-unaware | raw `push` fails (no upstream); git doesn't know to publish the corresponding branch |
[...]

PR #7878, opened 2026-06-09
Nf: design of a new command `datalad reset`

just-meng:
As discussed at length in #7872, fixing `datalad update how=reset` (#7874) alone is not enough to enable comparative DataLad experience on Windows (or with Windows users). To date, Windows users are blocked from time-travels and thus the full range of VC,  due to the fact that they are forced onto an adjusted branch where files are unlocked (no symlinks). A plain `git reset` on an adjusted branch only resets the view, but not the underlying corresponding branch that keeps the history. Windows users can only move forward in time, but not backward  (which is much more fun, right? ;-) ). The ability to time travel to the past is essential for any rigorous VC efforts and any serious cross-platform collaboration I would argue (out of lived pain).

I am proposing the addition of a new command -- `datalad reset` that 
1. enables correct `git reset` on an adjusted branch,
2. enables recursive(ly correct) `git reset` across dataset hierarchy.

The implementation will follow in another PR. I want to push the design first to get everyone's inputs. @yarikoptic @mih

### DIFF
--- a/docs/source/design/index.rst
+++ b/docs/source/design/index.rst
@@ -21,6 +21,7 @@ subsystems in DataLad.
    dataset_argument
    log_levels
    drop
+   reset
    python_imports
    miscpatterns
    exception_handling
--- a/docs/source/design/reset.rst
+++ b/docs/source/design/reset.rst
@@ -1,0 +1,183 @@
+.. -*- mode: rst -*-
+.. vi: set ft=rst sts=4 ts=4 sw=4 et tw=79:
+
+.. _chap_design_reset:
+
+************************************
+Reset a dataset to a target revision
+************************************
+
+.. topic:: Specification scope and status
+
+   This specification is a proposal, subject to review and further discussion.
+   A new command :command:`reset` is introduced to enable correct `git reset --hard`
+   behavior on an adjusted branch. Recursive operation (``--recursive``/``--follow``)
+   allows reset across the dataset hierarchy on both normal and adjusted branches.
+
+§1 The :command:`reset` command discards local divergence and sets a dataset to
+a target revision. On a normal branch ``datalad reset <target>`` is equivalent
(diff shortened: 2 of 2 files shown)


just-meng:
Hmm something is wrong with the PR, need to fix it.
It landed on maint, just switched the base to master. Now it looks fine to me (as intended).

just-meng:
Good disucssion I had with myself (and Claude). We considered including/excluding other reset modes, i.e. `git reset --{soft/mixed}` and here are our thoughts:

`git reset <target>` always moves HEAD to `<target>`; the mode decides how far that propagates into the index and  working tree:
     
  | mode | HEAD / branch | index (staging) | working tree |
  |---|---|---|---|
  | `--soft` | → target | *untouched* | *untouched* |
  | `--mixed` (default) | → target | → target | *untouched* |
  | `--hard` | → target | → target | → target |
 
So the difference is where the undone changes end up: `--soft` leaves them staged, `--mixed` leaves them unstaged, `--hard` discards them.
  
Mnemonic: soft = *uncommit* · mixed = *uncommit + unstage* · hard = *uncommit + unstage + discard*.

Neither of `--soft` or `--mixed` fits into DataLad's existing landscape. DataLad doesn't expose the staging area per se; `datalad save` goes straight from the working copy to history, no staging visible. DataLad is also mostly concerned with ensuring fundamental aspects of data versioning (integrity, availability, provenance, reproducibility) and less with the cosmetics of the commit history. 

We conclude that `--soft/--mixed` are local commit-surgery tools — "rearrange my own commits while keeping the work." `datalad reset`'s purpose is the --hard notion — "discard divergence, match a revision" — a dataset-level operation. We updated the design document and will push the implementation soon.

just-meng:
The updated design doc is now part of #7884.

PR #7880, opened 2026-06-29
BF: resolve the update target's corresponding branch on adjusted branches

### LINKED ISSUE #7873: update --how=reset on adjusted branch impossible: Could not determine update target
### What is the problem?

Apparently I never reported this bug: 

```
❮ datalad update --how reset                                                                                      
[INFO   ] Fetching updates for Dataset(/home/wjm/.claude/datalad-adjusted-sandbox/tmp/clone) 
update(impossible): . (dataset) [Could not determine update target]
```

Referenced in #7872, should be fixed by the follow-up PR.

```
mkdir tmp && cd tmp                                                                                                   
datalad create src
datalad clone src clone
git -C clone annex adjust --unlock
datalad -C clone update --how=reset
```

### DataLad information

datalad 1.4.1

### Additional context

_No response_

_No response_

just-meng:
Closes #7873.

On an adjusted branch (Windows / crippled filesystem) the active branch is e.g. `adjusted/master(unlocked)`, which has no remote counterpart. `_choose_update_target` built the update target from the adjusted name (`<remote>/adjusted/master(unlocked)`, a nonexistent ref), so `datalad update --how=reset` (and `--how=merge`) aborted with "Could not determine update target".

**Fix:** resolve the corresponding branch when building the target, so it becomes `<remote>/master`. A one-liner.

**Test:** 
- `test_update_reset_adjusted`: the update runs and lands both the corresponding branch and the adjusted view at the sibling.
- helper `corresponding_hexsha`: reads the hexsha NOT from the view, but from the corresponding branch below the view, so corr-aware.
- helper `maybe_unadjust_repo`: checks out the corresponding branch of the source data (while the clone, the dataset under control, stays adjusted) to enable "normal-enough" branch behaviour on a crippled FS -- the reverse of the existing `maybe_adjust_repo`.

Note, `test_update_reset_adjusted` deliberately tests beyond the observed bug and carries a generic name for `datalad update --how=reset` on adjusted branch, because no test exits in the test suite that asserts the basic behaviour of this command.

### DIFF
--- a/changelog.d/adjusted_update_target.md
+++ b/changelog.d/adjusted_update_target.md
@@ -1,0 +1,9 @@
+### 🐛 Bug Fixes
+
+- `update --how=reset` (and `--how=merge`) now resolves the corresponding branch
+  when determining the update target on adjusted branches (Windows / crippled
+  filesystems). Previously the adjusted branch name was used to build a
+  nonexistent `<remote>/adjusted/...` ref, so the update aborted with
+  "Could not determine update target".
+  Fixes [#7873](https://github.com/datalad/datalad/issues/7873).
+  (by [@just-meng](https://github.com/just-meng))
--- a/datalad/distribution/tests/test_update.py
+++ b/datalad/distribution/tests/test_update.py
@@ -41,10 +41,12 @@ from datalad.tests.utils_pytest import (
     assert_repo_status,
     assert_result_count,
     assert_status,
+    corresponding_hexsha,
     create_tree,
     eq_,
     known_failure_windows,
     maybe_adjust_repo,
+    maybe_unadjust_repo,
     neq_,
     ok_,
     ok_file_has_content,
@@ -1039,6 +1041,65 @@ def test_update_reset_dirty(path=None):
     assert_repo_status(ds_clone.path,
                        modified=[ds_clone_s1.repo.path,
                                  ds_clone_s2.repo.path])
+
+
+@slow  # ~10s
+@pytest.mark.parametrize("divergence", ["local_discard", "remote_discard"])
+@with_tempfile(mkdir=True)
+def test_update_reset_adjusted(path=None, *, divergence):
+    """`update --how=reset` must actually run on an adjusted branch following a sibling.
+
+    Two ways the corresponding branch can differ from the sibling:
+    ``local_discard`` (clone has an extra local commit) and ``remote_discard``
+    (the sibling moved back without the clone, e.g. via --force push).
+    Either way, target determination and the reset itself must work on the adjusted branch.
+
+    Regression: target determination used the *adjusted* branch name (e.g.
+    ``adjusted/master(unlocked)``) and built a nonexistent
+    ``<remote>/adjusted/master(unlocked)`` ref, so the command aborted with
+    "Could not determine update target".  It must instead resolve the
+    corresponding branch and reset to ``<remote>/<corresponding>``.
+    """
+    path = Path(path)
+
+    ds_src = Dataset(path / "source").create()
+    (ds_src.pathobj / "remote.txt").write_text("base commit")
+    ds_src.save()
+
+    ds_clone = install(source=ds_src.path, path=path / "clone",
+                       result_xfm="datasets")
+    maybe_adjust_repo(ds_clone.repo)
+    if not ds_clone.repo.is_managed_branch():
+        pytest.skip("test requires adjusted branch support")
+
+    if divergence == 'local_discard':
+        (ds_clone.pathobj / "local.txt").write_text("local only, discard me by reset")
+        ds_clone.save()
+    elif divergence == 'remote_discard':
+        # A raw git reset --hard must move the sibling's corresponding branch
+        # -- the one rewrite that needs an unadjust here (see maybe_unadjust_repo).
+        maybe_unadjust_repo(ds_src.repo)
+        ds_src.repo.call_git(["reset", "--hard", "HEAD~"])
+
+    # Read both real tips with corresponding_hexsha (corr-aware)
+    target_hexsha = corresponding_hexsha(ds_src.repo)
+    pre_update_hexsha = corresponding_hexsha(ds_clone.repo)
+    neq_(pre_update_hexsha, target_hexsha)  # the local clone has diverged
+
+    # reset local clone to sibling (previously raised "Could not determine update target")
+    assert_in_results(
+        ds_clone.update(follow="sibling", how="reset"),
+        action="update.reset", status="ok")
+
+    # The corresponding branch matches the sibling again (local commit gone)...
+    post_update_hexsha = corresponding_hexsha(ds_clone.repo)
+    eq_(post_update_hexsha, target_hexsha)
+    # ... the adjusted view's anchor (basis) was re-anchored to the target too,
+    # otherwise the next git annex adjust would rebuild the view from the wrong basis
+    adjusted = ds_clone.repo.get_active_branch()
+    eq_(ds_clone.repo.get_hexsha(f"refs/basis/{adjusted}"), target_hexsha)
+    # ... and we are still on the adjusted branch.
+    ok_(ds_clone.repo.is_managed_branch())
 
 
 def test_process_how_args():
--- a/datalad/tests/utils_pytest.py
+++ b/datalad/tests/utils_pytest.py
@@ -1952,6 +1952,47 @@ def maybe_adjust_repo(repo):
         repo.adjust()
 
 
+def maybe_unadjust_repo(repo):
+    """Check out `repo`'s *corresponding* branch (the real history beneath
+    the adjusting commit), the inverse of `maybe_adjust_repo`.
+
+    Call this function before you need to modify the commit history of a source/
+    sibling repo that has been created on a crippled FS (e.g. Windows CI).
+    A crippled FS forces every repo onto the adjusted branch. A plain
+    ``git reset --hard`` must land on the corresponding branch (not the adjusted
+    view). On a repo that is not adjusted it is a no-op.
+
+    Caveats:
+
+    - only faithful for asserting SHA-identity after a history-rewrite. If you
+      try to assert content-identity, you'll need to materialize the content first
+      via e.g. ``git checkout -- .``.
+
+    - only plain ``git`` keeps the repo unadjusted. DataLad's ``save``, ``update``
+      and ``push`` run ``git annex sync`` internally, which re-adjusts the branch
+      on a crippled filesystem.
+    """
+    if repo.is_managed_branch():
+        corr_branch = repo.get_corresponding_branch()
+        adjusted = repo.get_active_branch()
+        repo.call_git(["checkout", corr_branch])
+        repo.call_git(["branch", "-D", adjusted])
+
+
+def corresponding_hexsha(repo, ref="HEAD"):
+    """Hexsha of ``ref``, read from the corresponding branch when adjusted.
+
+    A plain ``get_hexsha(HEAD)`` on an adjusted repo returns the SHA of the git-annex
+    "adjusting" commit which is just the view. To read the SHA of the real history,
+    HEAD-relative refs need to be resolved against the corresponding branch instead.
+    On a normal branch it is a plain ``get_hexsha(ref)``.
+    """
+    if (corr := repo.get_corresponding_branch()) and (
+            ref == "HEAD" or ref.startswith(("HEAD~", "HEAD^", "HEAD@"))):
+        ref = corr + ref[len("HEAD"):]
+    return repo.get_hexsha(ref)
+
+
 @lru_cache()
 @with_tempfile
 @with_tempfile
--- a/datalad/distribution/update.py
+++ b/datalad/distribution/update.py
@@ -485,6 +485,12 @@ def _choose_update_target(repo, branch, remote, cfg_remote):
              f"{repo.get_corresponding_branch(branch) or ''}" "@{upstream}"],
             read_only=True)
     elif branch:
+        # On an adjusted branch the active branch (e.g.
+        # 'adjusted/master(unlocked)') has no remote counterpart; the sibling
+        # tracks the corresponding branch ('master'). Resolve it so the target
+        # becomes e.g. 'origin/master' instead of the nonexistent
+        # 'origin/adjusted/master(unlocked)'.
+        branch = repo.get_corresponding_branch(branch) or branch
         remote_branch = "{}/{}".format(remote, branch)
         if repo.commit_exists(remote_branch):
             target = remote_branch
--- a/datalad/tests/test_tests_utils_pytest.py
+++ b/datalad/tests/test_tests_utils_pytest.py
@@ -42,6 +42,7 @@ from _pytest.outcomes import (
 )
 
 from datalad import cfg as dl_cfg
+from datalad.distribution.dataset import Dataset
 from datalad.support import path as op
 from datalad.support.gitrepo import GitRepo
 from datalad.tests.utils_pytest import (
@@ -56,6 +57,7 @@ from datalad.tests.utils_pytest import (
     assert_re_in,
     assert_str_equal,
     assert_true,
+    corresponding_hexsha,
     eq_,
     fs_supports_filename,
     get_most_obscure_supported_name,
@@ -63,6 +65,9 @@ from datalad.tests.utils_pytest import (
     known_failure_githubci_win,
     known_failure_windows,
     local_testrepo_flavors,
+    maybe_adjust_repo,
+    maybe_unadjust_repo,
+    neq_,
     nok_startswith,
     ok_,
     ok_broken_symlink,
@@ -82,6 +87,7 @@ from datalad.tests.utils_pytest import (
     serve_path_via_http,
     signal_timeout,
     skip_if,
+    skip_if_adjusted_branch,
     skip_if_no_module,
     skip_if_no_network,
     skip_if_on_windows,
@@ -771,3 +777,66 @@ def test_signal_timeout_noop_without_sigalrm(monkeypatch):
     # but with SIGALRM removed it must be a pure no-op.
     with signal_timeout(0.001):
         time.sleep(0.05)
+
+
+#
+# Test the adjusted-branch helpers
+#
+
+@with_tempfile(mkdir=True)
+def test_corresponding_hexsha(path=None):
+    ds = Dataset(path).create()
+    maybe_adjust_repo(ds.repo)
+    corr = ds.repo.get_corresponding_branch()
+    old = ds.repo.get_hexsha(corr)
+
+    (ds.pathobj / "new.txt").write_text("NEW")
+    ds.save()
+    new = ds.repo.get_hexsha(corr)
+
+    neq_(old, new)
+    neq_(ds.repo.get_hexsha("HEAD"), new)
+
+    # valid non-HEAD refs that contain "HEAD" in the string
+    ds.repo.call_git(["branch", "myHEAD", new])
+    ds.repo.call_git(["update-ref", "ORIG_HEAD", new])
+
+    eq_(corresponding_hexsha(ds.repo), new)
+    eq_(corresponding_hexsha(ds.repo, corr), new)
+    eq_(corresponding_hexsha(ds.repo, old), old)
+    eq_(corresponding_hexsha(ds.repo, "HEAD~"), old)
+    eq_(corresponding_hexsha(ds.repo, "HEAD^"), old)
+    eq_(corresponding_hexsha(ds.repo, "HEAD@{0}"), new)
+    eq_(corresponding_hexsha(ds.repo, "myHEAD"), new)
+    eq_(corresponding_hexsha(ds.repo, "ORIG_HEAD"), new)
+
+    for bogus in ("bogus", "bogusHEAD", "HEAD~99"):
+        assert_raises(ValueError, corresponding_hexsha, ds.repo, bogus)
+
+
+@skip_if_adjusted_branch
+@with_tempfile(mkdir=True)
+def test_maybe_unadjust_repo_noop(path=None):
+    repo = Dataset(path).create().repo
+    branch, hexsha = repo.get_active_branch(), repo.get_hexsha("HEAD")
+    maybe_unadjust_repo(repo)
+    eq_(repo.get_active_branch(), branch)
+    eq_(repo.get_hexsha("HEAD"), hexsha)
+
+
+@with_tempfile(mkdir=True)
+def test_maybe_unadjust_repo(path=None):
+    ds = Dataset(path).create()
+    maybe_adjust_repo(ds.repo)
+    corr = ds.repo.get_corresponding_branch()
+    corr_hexsha = ds.repo.get_hexsha(corr)
+
+    maybe_unadjust_repo(ds.repo)
+    assert_false(ds.repo.is_managed_branch())
+    eq_(ds.repo.get_active_branch(), corr)
+    eq_(ds.repo.get_hexsha("HEAD"), corr_hexsha)
+
+    # idempotent: a second call is the no-op case
+    maybe_unadjust_repo(ds.repo)
+    eq_(ds.repo.get_active_branch(), corr)
+    eq_(ds.repo.get_hexsha("HEAD"), corr_hexsha)


changelog.d/adjusted_update_target.md
```diff
@@ -0,0 +1,9 @@
+### 🐛 Bug Fixes
+
+- `update --how=reset` (and `--how=merge`) now resolve the corresponding branch
```
yarikoptic:
```suggestion
- `update --how=reset` (and `--how=merge`) now resolves the corresponding branch
```

datalad/tests/utils_pytest.py
```diff
@@ -1952,6 +1952,57 @@ def maybe_adjust_repo(repo):
...
+      (not the adjusted view) belongs here. For *read-only* SHA access prefer
+      `corresponding_hexsha`, which observes without touching the repo -- no
+      unadjust needed.
+    - File *content*: no. This is the corresponding branch, not a normal one;
+      annexed files on it may be unmaterialised pointers rather than bytes
+      (especially on a crippled fs), so do not read file content through it.
+
+    Caveat: only plain ``git`` keeps the repo unadjusted. DataLad's ``save``,
+    ``update`` and ``push`` run ``git annex sync`` internally, which re-adjusts
+    the branch on a crippled filesystem.
```
yarikoptic: I find such AI-sloppy descriptions increasingly less useful. Some portions/reasoning is appropriate more for a commit message, e.g. in particular behavior/expectations of code **outside** of this function (e.g. save, update etc).  Condence into may be a few lines overall motivation/description on why this function is needed.

datalad/tests/utils_pytest.py
```diff
@@ -1952,6 +1952,57 @@ def maybe_adjust_repo(repo):
...
+    """Hexsha of ``ref``, read from the corresponding branch when adjusted, otherwise from the active branch.
+
+    The *read* half of the adjusted-branch fixture principle (see
+    `maybe_unadjust_repo` for the *rewrite* half). On a crippled filesystem
+    (e.g. Windows CI) `create()`/`clone()` put a repo on an adjusted branch
+    whose HEAD carries an extra git-annex "adjusting" commit on top of the real
+    history, so a plain ``get_hexsha("HEAD")`` there is off by that commit. This
+    resolves HEAD-relative refs against the corresponding branch instead -- so a
+    test can *observe* a repo's real history without mutating the fixture. On a
+    normal branch it is a plain ``get_hexsha(ref)``.
```
yarikoptic: the same here about description -- condense please

datalad/tests/utils_pytest.py
```diff
@@ -1952,6 +1952,57 @@ def maybe_adjust_repo(repo):
         repo.adjust()
 
 
+def maybe_unadjust_repo(repo):
+    """Check out `repo`'s *corresponding* branch, the inverse of
+    `maybe_adjust_repo`.
```
yarikoptic: I feel like it should fit 1 line?


```suggestion
    """Check out `repo`'s *corresponding* branch, the inverse of `maybe_adjust_repo`.
```

datalad/tests/utils_pytest.py
```diff
@@ -1952,6 +1952,57 @@ def maybe_adjust_repo(repo):
...
+    `maybe_unadjust_repo` for the *rewrite* half). On a crippled filesystem
+    (e.g. Windows CI) `create()`/`clone()` put a repo on an adjusted branch
+    whose HEAD carries an extra git-annex "adjusting" commit on top of the real
+    history, so a plain ``get_hexsha("HEAD")`` there is off by that commit. This
+    resolves HEAD-relative refs against the corresponding branch instead -- so a
+    test can *observe* a repo's real history without mutating the fixture. On a
+    normal branch it is a plain ``get_hexsha(ref)``.
+    """
+    corr = repo.get_corresponding_branch()
+    if not corr:
```
yarikoptic: we can use walrus assignments now


```suggestion
    if not (corr := repo.get_corresponding_branch()):
```

yarikoptic (changes requested):
cosmetic changes only .

datalad/tests/utils_pytest.py
just-meng: This is not some "AI-sloppy description" and I find this comment rather dismissive and quite an underappreciation of the human effort behind it. 

Given the thin test coverage on adjusted branch behaviors in this codebase and the absence of Windows users among the maintainers, I deliberately inserted knowledge pieces and findings where I thought they would be helpful for a future fellow contributor testing adjusted branch behaviors. 

I think a good function description should tell me when to use the function and when not. When it works and when it breaks. If that's already known. Particularly relevant, when this is neither obvious nor trivial. 

My two cents. 

I gave my best to loose words without loosing relevant information. If you need it to be shorter, be so kind and make a suggestion please.

datalad/tests/utils_pytest.py
yarikoptic: this is what I would really call "dismissive" -- resolve without taking in and not commenting on ;-)  Anything against walrus'es?

datalad/tests/utils_pytest.py
yarikoptic: Sorry if I sounded dismissive - was not my intent.  We live in turbulent times , and with AI capable of producing tons of text, while my prior experience of pure-human performance is "being succinct", seeing such long descriptions in the tests raises my prior for them being AI-produced, and thus overall AI-sloppy EVEN if curated by a human.  

I know that you are a very prolific writer @just-meng , and I will adjust my priors for reviewing your contributions. But it remains the fact that we need to aim to be (more) succinct and to the point in our comments/in-code texts since otherwise it becomes hard to see the code trees behind the forest of text.  Ideally, the code itself should be descriptive enough to tell the story.

yarikoptic: and thanks again for pushing on making adjusted branches mode work for the strugglers!

datalad/tests/utils_pytest.py
```diff
@@ -1952,6 +1952,46 @@ def maybe_adjust_repo(repo):
...
+     view). On a repo that is not adjusted it is a no-op.
+
+    Caveats:
+    - only faithful for asserting SHA-identity after a history-rewrite. If you
+      try to assert content-identity, you'll need to materialize the content first
+      via e.g. ``git checkout -- .``.
+
+    - only plain ``git`` keeps the repo unadjusted. DataLad's ``save``, ``update``
+      and ``push`` run ``git annex sync`` internally, which re-adjusts the branch
+      on a crippled filesystem.
```
yarikoptic, line 1973: docs CI build is red  due to some mistakes around here

```
reading sources... [ 99%] publications
reading sources... [100%] related

/opt/hostedtoolcache/Python/3.10.20/x64/lib/python3.10/site-packages/datalad/tests/utils_pytest.py:docstring of datalad.tests.utils_pytest.maybe_unadjust_repo:8: ERROR: Unexpected indentation. [docutils]
/opt/hostedtoolcache/Python/3.10.20/x64/lib/python3.10/site-packages/datalad/tests/utils_pytest.py:docstring of datalad.tests.utils_pytest.maybe_unadjust_repo:12: ERROR: Unexpected indentation. [docutils]
looking for now-outdated files... none found
```

please make sure CI is green for it. Can test build docs locally via the same as on CI `make -C docs html doctest`

datalad/tests/utils_pytest.py
```diff
@@ -1952,6 +1952,46 @@ def maybe_adjust_repo(repo):
...
+
+    A plain ``get_hexsha(HEAD)`` on an adjusted repo returns the SHA of the git-annex
+    "adjusting" commit which is just the view. To read the SHA of the real history,
+    HEAD-relative refs need to be resolved against the corresponding branch instead.
+    On a normal branch it is a plain ``get_hexsha(ref)``.
+    """
+    corr = repo.get_corresponding_branch()
+    if not corr:
+        return repo.get_hexsha(ref)
+    return repo.get_hexsha(ref.replace("HEAD", corr))
```
yarikoptic: so in both cases we call out to `repo.get_hexsha` so we better get 1 of them and make logic easier to read by avoiding "not" as e.g.:

```suggestion
    if corr := repo.get_corresponding_branch():
        target_ref = ref.replace("HEAD", corr)
    else:
        target_ref = ref
    return repo.get_hexsha(target_ref)
```

WDYT?

datalad/tests/utils_pytest.py
just-meng: But the world is not binary, right? If AI's description eventually hits the sweet spot for me after many iterations, I'm not gonna make the prose sound more human just to please the maintainers. 

So are you happy with the pure human rewrite of the function docstring now?

Alternatively, we could try to find a place to document testing strategies for crippled FS and relocate the context content there. Maybe you could start reviewing #7884 to see the extent of testing strategies.

datalad/tests/utils_pytest.py
just-meng: That's not me being dismissive, but me having poor gh PR skills ... otherwise it was unconditional acceptance of your suggestion :) 

Strange, I did apply this change in commit f2505b8d530f50b49f8cd4a3a3e8948a3c9afff2!

datalad/tests/utils_pytest.py
```diff
@@ -1997,8 +1996,7 @@ def corresponding_hexsha(repo, ref="HEAD"):
     test can *observe* a repo's real history without mutating the fixture. On a
     normal branch it is a plain ``get_hexsha(ref)``.
     """
-    corr = repo.get_corresponding_branch()
-    if not corr:
+    if not (corr := repo.get_corresponding_branch()):
```
just-meng: @yarikoptic, here is proof. I must have messed up things while being handicapped on my Windows laptop without jj, and lying in a hammock at a campsite without benches. Lemme fix it.

datalad/tests/utils_pytest.py
just-meng: Done. Let's also add a line to CONTRIBUTING.md?

datalad/tests/utils_pytest.py
just-meng: Concurred.

datalad/tests/utils_pytest.py
just-meng: Resolved by your other suggestion, already applied.

just-meng:
Say the word and I will squash all review commits into the original commits.

datalad/distribution/tests/test_update.py
```diff
@@ -1041,6 +1043,65 @@ def test_update_reset_dirty(path=None):
...
+        pytest.skip("test requires adjusted branch support")
+
+    if divergence == 'local_discard':
+        (ds_clone.pathobj / "local.txt").write_text("local only, discard me by reset")
+        ds_clone.save()
+    elif divergence == 'remote_discard':
+        # A raw git reset --hard must move the sibling's corresponding branch
+        # -- the one rewrite that needs an unadjust here (see maybe_unadjust_repo).
+        maybe_unadjust_repo(ds_src.repo)
+        ds_src.repo.call_git(["reset", "--hard", "HEAD~"])
```
yarikoptic, line 1082: Is unadjusting really needed (as code below depends on that?). If not, and it was just the need to update the branch, that here I think you could have just used smth like

```suggestion
        src_cor_branch = ds_src.get_corresponding_branch()
        ds_src.repo.call_git(["branch", "-f", src_cor_branch, src_cor_branch + "~"])
```
or even via `update-ref` (see below), without "unadjusting" thus eliminating the need for that additional helper not accompanied by any test?

Also we have already `.update_ref`, so here it could better be I guess

```suggestion
        src_cor_branch = ds_src.get_corresponding_branch()
        ds_src.repo.update_ref(f"refs/heads/{src_cor_branch}", f"refs/heads/{src_cor_branch}~"])
```
?

datalad/tests/utils_pytest.py
```diff
@@ -1952,6 +1952,48 @@ def maybe_adjust_repo(repo):
...
+      on a crippled filesystem.
+    """
+    if repo.is_managed_branch():
+        corr_branch = repo.get_corresponding_branch()
+        adjusted = repo.get_active_branch()
+        repo.call_git(["checkout", corr_branch])
+        repo.call_git(["branch", "-D", adjusted])
+
+
+def corresponding_hexsha(repo, ref="HEAD"):
```
yarikoptic, line 1982: we do have https://github.com/datalad/datalad/blob/HEAD/datalad/tests/test_tests_utils_pytest.py to test the helpers, thus every added helper should be tested to ensure that when used it performs as expected. But now I even wonder if it could be avoided like the other one (see the other comment)

datalad/tests/utils_pytest.py
yarikoptic: if it is missing that -- sure, please add.

datalad/tests/utils_pytest.py
```diff
@@ -1952,6 +1952,48 @@ def maybe_adjust_repo(repo):
...
+    """Hexsha of ``ref``, read from the corresponding branch when adjusted, otherwise from the active branch.
+
+    A plain ``get_hexsha(HEAD)`` on an adjusted repo returns the SHA of the git-annex
+    "adjusting" commit which is just the view. To read the SHA of the real history,
+    HEAD-relative refs need to be resolved against the corresponding branch instead.
+    On a normal branch it is a plain ``get_hexsha(ref)``.
+    """
+    if corr := repo.get_corresponding_branch():
```
yarikoptic: what if `ref="refs/heads/adjusted/myfancy"` or alike?  the point is that ref is completely ignored here.  And that is where a dedicated test where you would have tested against different options might have helped to identify... I guess the "ref" should have been treated/passed into `get_corresponding_branch` here...


<details>
<summary>But overall, this analysis made me look through the code where this pattern is actually quite common and now has a number of "variations", at times just with already a condition of being on an adjusted branch:
</summary> 

```shell
datalad/support/gitrepo.py:            subrepo_commit = subrepo.get_hexsha(subrepo.get_corresponding_branch())

datalad/support/annexrepo.py-            adjusted_branch = self.get_active_branch()
datalad/support/annexrepo.py-            corresponding_branch = self.get_corresponding_branch()
datalad/support/annexrepo.py:            old_sha = self.get_hexsha(corresponding_branch)

datalad/distribution/update.py-            if follow_parent_lazy and \
datalad/distribution/update.py:               repo.get_hexsha(repo.get_corresponding_branch()) == revision:

datalad/core/local/tests/test_save.py-    sub_branch = subds.repo.get_corresponding_branch()
datalad/core/local/tests/test_save.py-
datalad/core/local/tests/test_save.py-    # amend in subdataset w/ new message; otherwise empty amendment:
datalad/core/local/tests/test_save.py:    last_sha = subds.repo.get_hexsha(sub_branch)

datalad/support/gitrepo.py:            subrepo_commit = subrepo.get_hexsha(subrepo.get_corresponding_branch())
```
and there could be more
</details>

questioning either we actually need a helper or `repo.get_hexsha(repo.get_corresponding_branch())` is sufficient for your test cases (since as I have mentioned, `ref` seems not to be used correctly here)?

yarikoptic (changes requested):
Sorry to be the pain, but after a bit more careful review, I think we do not need to introduce (potentially broken as untested) test helpers at all.  I could be proven wrong of cause.

Do not worry about squashing, we could squash on merge but you also can squash as you go as you like.

yarikoptic:
<details>
<summary>wow -- appveyor error was not even time out but some other curious error, unrelated to this PR AFAIK</summary> 

```shell
FAILED ..\datalad\tests\test_utils_cached_dataset.py::test_get_cached_dataset - datalad.runner.exception.CommandError: CommandError: '"git" "-c" "diff.ignoreSubmodules=none" "-c" "core.quotepath=false" "annex" "get" "-c" "annex.retry=3" "--key" "MD5E-s144625--4c458c62b7ac8ec8e19c8ff14b2e34ad.gif" "--json" "--json-error-messages" "--json-progress" "-c" "annex.dotfiles=true"' failed with exitcode 1 under C:\DLTMP\datalad_temp_n5ilfaui\https___github.com_datalad_testrepo--minimalds [info keys: stdout_json]
> from web...
Unable to access these remotes: web
(Note that these git remotes have annex-ignore set: dl-test-remote)
  https://github.com/datalad/datalad-testrepo/raw/6bf8a8b12871fbe16f9ce8f2c631e18231b8dafc/ingit/animated.gif download failed
  downloading from all 1 known url(s) failed
  https://github.com/datalad/datalad-testrepo/raw/6bf8a8b12871fbe16f9ce8f2c631e18231b8dafc/ingit/animated.gif download failed
  downloading from all 1 known url(s) failed
  https://github.com/datalad/datalad-testrepo/raw/6bf8a8b12871fbe16f9ce8f2c631e18231b8dafc/ingit/animated.gif download failed
  downloading from all 1 known url(s) failed [err: 'curl: (35) schannel: next InitializeSecurityContext failed: CRYPT_E_REVOCATION_OFFLINE (0x80092013) - The revocation function was unable to check revocation because the revocation server was offline.
curl: (35) schannel: next InitializeSecurityContext failed: CRYPT_E_REVOCATION_OFFLINE (0x80092013) - The revocation function was unable to check revocation because the revocation server was offline.
curl: (35) schannel: next InitializeSecurityContext failed: CRYPT_E_REVOCATION_OFFLINE (0x80092013) - The revocation function was unable to check revocation because the revocation server was offline.
get: 1 failed']
FAILED ..\datalad\tests\test_utils_cached_dataset.py::test_cached_dataset - datalad.runner.exception.CommandError: CommandError: '"git" "-c" "diff.ignoreSubmodules=none" "-c" "core.quotepath=false" "annex" "get" "-c" "annex.retry=3" "--key" "MD5E-s144625--4c458c62b7ac8ec8e19c8ff14b2e34ad.gif" "--json" "--json-error-messages" "--json-progress" "-c" "annex.dotfiles=true"' failed with exitcode 1 under C:\DLTMP\datalad_temp_2rwfdb9t\https___github.com_datalad_testrepo--minimalds [info keys: stdout_json]
> from web...
Unable to access these remotes: web
(Note that these git remotes have annex-ignore set: dl-test-remote)
  https://github.com/datalad/datalad-testrepo/raw/6bf8a8b12871fbe16f9ce8f2c631e18231b8dafc/ingit/animated.gif download failed
  downloading from all 1 known url(s) failed
  https://github.com/datalad/datalad-testrepo/raw/6bf8a8b12871fbe16f9ce8f2c631e18231b8dafc/ingit/animated.gif download failed
  downloading from all 1 known url(s) failed
  https://github.com/datalad/datalad-testrepo/raw/6bf8a8b12871fbe16f9ce8f2c631e18231b8dafc/ingit/animated.gif download failed
  downloading from all 1 known url(s) failed [err: 'curl: (35) schannel: next InitializeSecurityContext failed: CRYPT_E_REVOCATION_OFFLINE (0x80092013) - The revocation function was unable to check revocation because the revocation server was offline.
curl: (35) schannel: next InitializeSecurityContext failed: CRYPT_E_REVOCATION_OFFLINE (0x80092013) - The revocation function was unable to check revocation because the revocation server was offline.
curl: (35) schannel: next InitializeSecurityContext failed: CRYPT_E_REVOCATION_OFFLINE (0x80092013) - The revocation function was unable to check revocation because the revocation server was offline.
get: 1 failed']
FAILED ..\datalad\tests\test_utils_cached_dataset.py::test_cached_url - datalad.runner.exception.CommandError: CommandError: '"git" "-c" "diff.ignoreSubmodules=none" "-c" "core.quotepath=false" "annex" "get" "-c" "annex.retry=3" "--key" "MD5E-s144625--4c458c62b7ac8ec8e19c8ff14b2e34ad.gif" "--json" "--json-error-messages" "--json-progress" "-c" "annex.dotfiles=true"' failed with exitcode 1 under C:\DLTMP\datalad_temp_blh1x7i9\https___github.com_datalad_testrepo--minimalds [info keys: stdout_json]
> from web...
Unable to access these remotes: web
(Note that these git remotes have annex-ignore set: dl-test-remote)
  https://github.com/datalad/datalad-testrepo/raw/6bf8a8b12871fbe16f9ce8f2c631e18231b8dafc/ingit/animated.gif download failed
  downloading from all 1 known url(s) failed
  https://github.com/datalad/datalad-testrepo/raw/6bf8a8b12871fbe16f9ce8f2c631e18231b8dafc/ingit/animated.gif download failed
  downloading from all 1 known url(s) failed
  https://github.com/datalad/datalad-testrepo/raw/6bf8a8b12871fbe16f9ce8f2c631e18231b8dafc/ingit/animated.gif download failed
  downloading from all 1 known url(s) failed [err: 'curl: (35) schannel: next InitializeSecurityContext failed: CRYPT_E_REVOCATION_OFFLINE (0x80092013) - The revocation function was unable to check revocation because the revocation server was offline.
curl: (35) schannel: next InitializeSecurityContext failed: CRYPT_E_REVOCATION_OFFLINE (0x80092013) - The revocation function was unable to check revocation because the revocation server was offline.
curl: (35) schannel: next InitializeSecurityContext failed: CRYPT_E_REVOCATION_OFFLINE (0x80092013) - The revocation function was unable to check revocation because the revocation server was offline.
get: 1 failed']
===== 3 failed, 490 passed, 105 skipped, 2 xfailed in 2516.24s (0:41:56) ======
Command exited with code 1

```
</details>

just-meng:
Glad to see you engage with the "meat" of the PR, @yarikoptic  :) 

All three points you raised are really good, and I will address them individually in more detail. But since this is part one out of 3 stacked PRs, I think you'd get a better feeling of whether the test helpers are a good idea. Take a look at #7881, and I'll try to point out the specific calls that demonstrate that the suggested simplification from you might be insufficient.

datalad/tests/utils_pytest.py
just-meng: ah, that's great! will add tests for the helpers.

datalad/distribution/tests/test_update.py
just-meng: Very sharp. For this particular case, you are right; either of your suggestions would do. 

I've pointed you at another case in #7881 where we need the worktree to be clean afterwards so we can push back to the sibling. `git update-ref / branch -f ` would not touch the index and worktree; `git reset --hard` moves the ref AND leaves a clean worktree. For that, we need to "unadjust" the repo.

That's on the mechanics. Now on semantics, all three PRs #7880, #7881, #7884 focus on the `git reset` operation on adjusted branches. Replacing `git reset --hard` with either `git update-ref / branch -f ` could be missing the point, even if the mechanical outcome would be achieved. We want to understand the exact conditions under which a `git reset --hard` operation fails/succeeds on the continuum of normal <--> crippled FS.

datalad/tests/utils_pytest.py
just-meng: Very sharp, and I agree this is where the helper tests are needed.

1. So an invalid ref will just be passed through, not exactly ignored. 

    - On a normal branch, `corresponding_hexsha(repo, ref)` just passes everything through to `repo.get_hexsha(ref)`. Invalid ref should raise a loud guard there, not here. 

    - On adjusted branches, `"HEAD"` points to the git-annex adjusting commit, NOT the tip of the corresponding branch. ONLY `HEAD`-relative refs get a treatment; all other refs can be resolved without any problem, so they get passed on to `repo.get_hexsha(ref)`. Again, invalid refs should raise an error there, not here.

2. I pointed you at a case in #7881 to demonstrate when `repo.get_hexsha(repo.get_corresponding_branch())` is insufficient.

3. I think the `.replace("HEAD", corr)` is still fragile, though, will take a closer look for cases where HEAD is legitimately part of the ref like `myHEAD`.

4. In #7884 I actually move this helper into production code to achieve `git reset --hard` on crippled FS. If you want more "simplification/unification", we could make the existing `get_hexsha()` corr-aware, but this will have a big blast radius.

just-meng:
Tightened the logic inside `corresponding_hexsha` and added tests for the two test helpers. @yarikoptic Have you had a chance to review and confirm that the helpers are justified?

yarikoptic-gitmate (changes requested):
Reviewed on behalf of @yarikoptic (pending review request). I checked out the PR head (`5213140`) in an isolated worktree, installed `git-annex` (10.20260717) via `uv pip install git-annex`, and ran the tests independently.

## Core fix: correct and minimal

`datalad/distribution/update.py:493`:
```python
branch = repo.get_corresponding_branch(branch) or branch
```
This is the right, minimal fix, and it follows the existing idiom used elsewhere in the codebase (`gitrepo.py`: `subrepo.get_hexsha(subrepo.get_corresponding_branch())`, `annexrepo.py`, `update.py`'s own `follow_parent_lazy` branch) — `get_corresponding_branch()` returns `None` on a non-managed branch, so the `or branch` fallback correctly preserves current (non-adjusted) behavior.

I verified the regression and the fix directly: temporarily reverting just this one line and re-running `test_update_reset_adjusted` reproduces exactly the reported bug:
```
update(impossible): . (dataset) [Could not determine update target]
```
With the line restored, both parametrizations pass:
```
datalad/distribution/tests/test_update.py::test_update_reset_adjusted[local_discard] PASSED
datalad/distribution/tests/test_update.py::test_update_reset_adjusted[remote_discard] PASSED
```
I also confirmed the downstream reset/re-adjust path (`_annex_reset_target`, pre-existing, not touched by this PR) is exercised correctly and the adjusted view's `refs/basis/<branch>` anchor ends up correct, which the test explicitly checks. CI on the actual head commit is green (the `codecov/project` and `test (datalad-next)` failures are unrelated repo-wide/extension issues, not from this diff). The changelog entry and docstring RST issue that `yarikoptic` flagged earlier were both fixed in later commits (verified: `make -C docs html doctest`-style docstring content now parses without the earlier "Unexpected indentation" errors).

**This part is good and I would approve it on its own.**

## Outstanding concern from `yarikoptic` is not actually resolved

`yarikoptic`'s second `REQUEST_CHANGES` (review 4773013705) was explicit: *"I think we do not need to introduce (potentially broken as untested) test helpers at all."* Since then, `just-meng` added unit tests for `corresponding_hexsha`/`maybe_unadjust_repo` (addressing "untested"), but the deeper "do we need these at all" question — raised concretely in the still-open, **not outdated** thread on `test_update.py:1082` (`PRRT_kwDOANZqws6TiwB_`) — was never acted on in code:

- `yarikoptic` suggested for the `remote_discard` branch either `git branch -f <corr> <corr>~` or the already-existing `repo.update_ref(...)` (I confirmed `GitRepo.update_ref` already exists at `datalad/support/gitrepo.py:2516`), instead of adding a new `maybe_unadjust_repo` helper.
- `just-meng`'s own reply: *"For this particular case, you are right; either of your suggestions would do."* — but the code was **not** changed; `maybe_unadjust_repo(ds_src.repo)` is still called in the current diff (`test_update.py`, `remote_discard` branch). The justification given for keeping it is entirely about two other, separate, not-yet-merged/reviewed PRs (#7881, #7884) that reuse it — not this PR's own test.
- I independently checked the second helper, `corresponding_hexsha`, against the same bar. In `test_update_reset_adjusted` it is only ever called with the default `ref="HEAD"`. I verified directly that for that call shape it is exactly equivalent to the already-existing codebase idiom:
  ```python
  >>> repo.get_hexsha(None) == repo.get_hexsha('HEAD') == repo.get_hexsha(repo.get_corresponding_branch())
  True
  ```
  (`get_hexsha(None)` treats `None`/falsy the same as `'HEAD'`, per `gitrepo.py:1618-1620`.) So the extra generality in `corresponding_hexsha` (handling `HEAD~`, `HEAD^`, `HEAD@{...}`, arbitrary refs, raising on bogus input) is dead weight as far as *this* PR's own test is concerned — it's exercised only by the helper's own dedicated unit test, not by anything `test_update_reset_adjusted` needs. That's a second, independent confirmation of `yarikoptic`'s original DRY/YAGNI instinct.

Net effect: the "untested" half of the objection was fixed, but the "do we need it at all" half — which is the part `just-meng` explicitly agreed with for this PR — was not, and the thread is still open (not resolved, not outdated) as of the current head.

## Minor, secondary observations (not blocking on their own)
- `maybe_unadjust_repo` does `git branch -D <adjusted>` after checkout — this permanently deletes the adjusted branch, not just "checks it out". That's a bit more destructive than the docstring's "the inverse of `maybe_adjust_repo`" framing suggests; worth a one-line callout if the helper is kept.
- Given this PR is scoped/labeled `semver-patch` and is meant to be a targeted bugfix for #7873, pulling in generic test infrastructure whose payoff is future stacked PRs works against keeping this PR minimal/self-contained (per this project's general preference for small, reviewable changes — `CONTRIBUTING.md`).

## Verdict

`REQUEST_CHANGES`. The production fix (`update.py`) is correct, minimal, well-tested, and I'd approve it standalone. But `yarikoptic`'s explicit, twice-repeated objection to introducing these test helpers in this particular PR has not been addressed in code — only argued in review comments, with the contributor's own reply conceding the simpler alternative "would do" here — and the thread raising it is still open. Recommend either (a) inlining the trivial equivalent (`repo.get_hexsha(repo.get_corresponding_branch())` and a `branch -f`/`update_ref` one-liner) directly in `test_update_reset_adjusted` and dropping the two new helpers + their dedicated tests from this PR, deferring them to #7881 where they're actually load-bearing, or (b) an explicit maintainer sign-off that the helpers are justified here despite not being strictly needed by this PR's own test.

---

Issue #7882, opened 2026-06-30
"too smart" in a symlinked repo: considers a parent directory as the "context" repo to operate in


yarikoptic:
I'm in a repo which is  symlinked to another location:
```shell
❯ pwd
/home/yoh/doc/grants/2026-os4s-bids/github-repo
❯ readlink -f .
/home/yoh/proj/bids/2026-os4s
❯ ls -ld $PWD
lrwxrwxrwx 1 yoh yoh 29 Jun  8 17:30 /home/yoh/doc/grants/2026-os4s-bids/github-repo -> /home/yoh/proj/bids/2026-os4s/
❯ ls -dl ~/doc/grants/2026-os4s-bids/.git
ls: cannot access '/home/yoh/doc/grants/2026-os4s-bids/.git': No such file or directory
❯ ls -dl ~/doc/grants/.git/
drwxrwsr-x 1 yoh yoh 90 Jun 30 16:39 /home/yoh/doc/grants/.git//
```

and actually that top level `grants/` was initialized but never had a commit
```
❯ git -C ~/doc/grants log
fatal: your current branch 'master' does not have any commits yet
```

and `git` and `git-annex` are just fine to report their status within this repository:

```shell
❯ git status
On branch master
Your branch is ahead of 'origin/master' by 34 commits.
  (use "git push" to publish your local commits)

nothing to commit, working tree clean
❯ ls -ld .git
drwxrwxr-x 1 yoh yoh 240 Jun 30 16:36 .git/
❯ git annex info | head
trusted repositories: 0
semitrusted repositories: 3
	00000000-0000-0000-0000-000000000001 -- web
	00000000-0000-0000-0000-000000000002 -- bittorrent
	c03761fb-acb3-45ee-9eea-63dceaa14106 -- yoh@bilena:~/proj/bids/2026-os4s [here]
untrusted repositories: 0
transfers in progress: none
available local disk space: 105.89 gigabytes (+0 bytes reserved)
local annex keys: 5
local annex size: 710.72 megabytes
git-annex: <stdout>: hPutBuf: resource vanished (Broken pipe)
```

but datalad (`status`) confuses itself to work under top level `grants/` instead of current (symlinked) repo:

```shell
❯ source /home/yoh/proj/datalad/datalad-master/.venv/bin/activate
❯ datalad --version
datalad 1.6.0
❯ datalad rerun e5a1b200cb2cc94bf79cb558891085de6e01e5c5
[...]

PR #7885, opened 2026-07-01
NF: avoid crash on get_toppath  - return `None` on non-existing path

LiamDGray:
When get_toppath() is called on a nonexistent path, the OSError handler recurses to parent directories via dirname(). If no ancestor has a .git, the recursive call returns None, and str(Path(None)) raises TypeError.

Fix: return None early when the OSError fallback also returns None.

This is an invited (by @yarikoptic) follow-up to the gh-7882 symlinked-dataset fix: the OSError handler's upward walk was what caused the wrong parent .git to be found when the original path was a broken symlink.

### DIFF
--- a/datalad/support/tests/test_gitrepo.py
+++ b/datalad/support/tests/test_gitrepo.py
@@ -796,6 +796,16 @@ def test_GitRepo_get_toppath(repo=None, tempdir=None, repo2=None):
     eq_(GitRepo.get_toppath(tempdir), None)
 
 
+def test_GitRepo_get_toppath_nonexistent():
+    # gh-7882: get_toppath with a nonexistent path should return None,
+    # not crash with TypeError when the OSError recursion reaches root
+    # without finding any parent .git
+    eq_(GitRepo.get_toppath(
+        '/tmp/__datalad_crash_test__/a/b/c/d', follow_up=True), None)
+    eq_(GitRepo.get_toppath(
+        '/tmp/__datalad_crash_test__/a/b/c/d', follow_up=False), None)
+
+
 @with_tempfile(mkdir=True)
 def test_GitRepo_dirty(path=None):
 
--- a/datalad/support/tests/test_gitrepo.py
+++ b/datalad/support/tests/test_gitrepo.py
@@ -800,10 +800,12 @@ def test_GitRepo_get_toppath_nonexistent():
     # gh-7882: get_toppath with a nonexistent path should return None,
     # not crash with TypeError when the OSError recursion reaches root
     # without finding any parent .git
-    eq_(GitRepo.get_toppath(
-        '/tmp/__nonexisting__/a/b/c/d', follow_up=True), None)
-    eq_(GitRepo.get_toppath(
-        '/tmp/__nonexisting__/a/b/c/d', follow_up=False), None)
+    import tempfile
+    unique_base = tempfile.mkdtemp()
+    os.rmdir(unique_base)
+    nonexistent = op.join(unique_base, 'a', 'b', 'c', 'd')
+    eq_(GitRepo.get_toppath(nonexistent, follow_up=True), None)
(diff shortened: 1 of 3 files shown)


datalad/support/tests/test_gitrepo.py
```diff
@@ -796,6 +796,16 @@ def test_GitRepo_get_toppath(repo=None, tempdir=None, repo2=None):
...
 
 
+def test_GitRepo_get_toppath_nonexistent():
+    # gh-7882: get_toppath with a nonexistent path should return None,
+    # not crash with TypeError when the OSError recursion reaches root
+    # without finding any parent .git
+    eq_(GitRepo.get_toppath(
+        '/tmp/__datalad_crash_test__/a/b/c/d', follow_up=True), None)
+    eq_(GitRepo.get_toppath(
+        '/tmp/__datalad_crash_test__/a/b/c/d', follow_up=False), None)
```
yarikoptic: I guess this assumes that no bright mind gets `/tmp/.git` somehow.  Might be worth adding some skip if there is `/tmp/.git` but also might be ok as is... 

Also best to have it explicit in the name like

```suggestion
    eq_(GitRepo.get_toppath(
        '/tmp/__nonexisting__/a/b/c/d', follow_up=True), None)
    eq_(GitRepo.get_toppath(
        '/tmp/__nonexisting__/a/b/c/d', follow_up=False), None)
```

since someone else might have another bright idea for adhoc `/tmp/__datalad_crash_test__` to test some crash .

datalad/support/tests/test_gitrepo.py
```diff
@@ -796,6 +796,18 @@ def test_GitRepo_get_toppath(repo=None, tempdir=None, repo2=None):
     eq_(GitRepo.get_toppath(tempdir), None)
 
 
+def test_GitRepo_get_toppath_nonexistent():
+    # gh-7882: get_toppath with a nonexistent path should return None,
+    # not crash with TypeError when the OSError recursion reaches root
+    # without finding any parent .git
+    import tempfile
+    unique_base = tempfile.mkdtemp()
+    os.rmdir(unique_base)
```
yarikoptic: I think would be sufficient/consistent with the rest:
```suggestion
@with_tempfile
def test_GitRepo_get_toppath_nonexistent(unique_base=None):
    # gh-7882: get_toppath with a nonexistent path should return None,
    # not crash with TypeError when the OSError recursion reaches root
    # without finding any parent .git
```

datalad/support/tests/test_gitrepo.py
LiamDGray: Sorry, was busy with other things -- and am still learning how Github PRs, Issues, and Actions work together.  I wasn't sure if you noticed that I pushed a second commit which does generate guaranteed-unique path. And now it looks like AppVeyor build failed? If I should do something more, please let me know.

yarikoptic: appveyor timeout is unrelated , ignore please... I approved the CI run now, if all green we should be good to go I think

yarikoptic:
datalad-next -- oddly didn't fail on a recent https://github.com/datalad/datalad/actions/runs/29340144004 but here failed on rerun?! I will rerun once more but now I start wondering if somehow it is effected by the diff (doesn't look like it should be)

yarikoptic:
I see the same datalad-next crash elsewhere as e.g. in https://github.com/datalad/datalad-extensions/actions/runs/29900647863/job/88860393809 

```
FAILED commands/tests/test_download.py::test_download_no_credential_leak_to_http - AssertionError: assert '401' in '503 Server Error: Service Temporarily Unavailable for url: https://httpbin.org/redirect-to?url=http://httpbin.org/basic-auth/mike/dummy'
```

confirming that it is not this PR specific and rather seems like git or git-annex build specific???  anyways , let's proceed here

PR #7895, opened 2026-08-04
DOC: Point out where tests for test helpers live in CONTRIBUTING.md

just-meng:
see title

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -252,7 +252,9 @@ own setup/teardown overhead.
 
 Test helpers live in
 [datalad/tests/utils_pytest.py](datalad/tests/utils_pytest.py); import what
-you need from there.
+you need from there.  A helper you *add* there is code like any other, so pin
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -255,7 +252,9 @@ own setup/teardown overhead.
 
 Test helpers live in
 [datalad/tests/utils_pytest.py](datalad/tests/utils_pytest.py); import what
-you need from there.
+you need from there.  A helper you *add* there is code like any other, so pin
+its contract with a test in
+[datalad/tests/test_tests_utils_pytest.py](datalad/tests/test_tests_utils_pytest.py).
```
yarikoptic: Good idea ! but let's make it less AI-speak:

```suggestion
you need from there.  A helper you *add* there is code like any other, so accompany it
with a test in [datalad/tests/test_tests_utils_pytest.py](datalad/tests/test_tests_utils_pytest.py).
```

PR #7901, opened 2026-08-21
BF: run/rerun --explicit: dirty declared inputs, and nested runs

yarikoptic-gitmate:
### Overview

Two reported defects of `datalad run --explicit` share one theme: the run record does not describe what actually happened. They are fixed together because the fixes touch the same handful of places in `run_command()`.

Fixes #5312, fixes #3565, fixes #7900.

Design document: `docs/designs/run-explicit/01-inputs-and-nesting.md`.

> **Narrowed per review.** This PR was three concerns deep; the concurrent-`run` work (#7899) now lives in **#7925**, stacked on this branch, where the use case that motivated it is argued on its own merits. `save.py` and `gitrepo.py` have left this PR entirely — what remains touches `run.py` and `rerun.py` only. The `datalad.run.dirty-inputs` configuration variable is gone as well; see below.

#### 1. Declared inputs were not checked (#5312, #3565)

`--explicit` tolerates a dirty dataset, but it never looked at the declared *inputs*, so a command could be executed on an input with unsaved modifications. The record then names an input state that is nowhere recorded, and a `rerun` of it uses the committed state instead.

`run`/`rerun` now refuse such a call before any worktree preparation, so a rejected run leaves no trace:

```
run(impossible): ... [declared inputs have unsaved modifications:
  .datalad/config [modified]. Save them, or run with --assume-ready=inputs]
```

Untracked content counts only when it *is* a declared input, not when it merely sits inside a declared directory. The **only** escape hatch is `--assume-ready inputs|both`.

There is no configuration variable: neither issue asked for one, the only escape hatch raised in either discussion is mih's note on #5312 that `--assume-ready` already exists, and a persistent `ignore` setting would silently reintroduce exactly the records this check exists to prevent — for every future run in that dataset — where `--assume-ready` says it per call, in the call.

Only `inputs` are checked, not `extra_inputs`, which is an unexposed detail for wrappers.

#### 2. A nested `run` was treated as an undeclared side effect (#7900)

Variant B of the reproducer — outer run declares only its own output — now succeeds:

```
* [DATALAD RUNCMD] sweep          (merge)
|\
| * Remaining changes after command execution   summary.txt
| * [DATALAD RUNCMD] cell 2                     o2
| * [DATALAD RUNCMD] cell 1                     o1
```

[...]

PR #7884, opened 2026-07-01
NF: new reset command for correct reset on adjusted branches and reset across a dataset hierarchy

just-meng:
- Closes #7872, #7878, #7782.

The new command `datalad reset` adds two new functionalities: 

- correct `git reset --hard` behavior on adjusted branches, e.g., Windows / crippled FS
- reset a whole dataset hierarchy, either with the target resolved independently or following the parent (on both normal and adjusted branches)

An annexed repo is forced onto the unlocked adjusted branch by default on a crippled FS. However, a plain `git reset --hard` on an adjusted branch only resets the *view* without touching the real history. Upon the next `git annex sync` (routinely exercised by `save`/ `update` / `push`),  the discarded commit resurrects (via the underlying `refs/basis/<adjusted>` for `save` or the `synced/` branch otherwise), and worse: new work is permanently lost upon subsequent `datalad save` (see #7872). 

The cure is to reset the *corresponding* branch, reconcile the synced branch (to prevent resurrection, see #7772), and force-rebuild the *view* from the *corresponding* branch (which automatically reconciles the `refs/basis/<adjusted>` as well). In the core, it is the same workaround I suggested in #7782 and applied to BF #7881.

**Note: this PR is stacked on #7881 (`update --how=reset` and `reset` naturally share the same reset engine), please merge that one first.**

In addition, `-r/--recursive` provides a convenient way to reset across a whole hierarchy. This is useful if you want to reset a nested worktree to the main worktree, reset the hierarchy to a tag/sibling, get a clean state everywhere, or reset subdatasets to the pointers recorded by the superdataset (if they are deliberately out of sync).

### Testing / Highlights
The main effort of this PR was invested in writing tests that cover different branch / FS modes: normal, adjusted, and crippled. Tests are only skipped per leg, e.g., normal is skipped on a crippled FS. All three legs run on Linux with the limitation that the crippled mode is only simulated and fails to capture the crippled FS faithfully for a history-moving reset whose target retains annexed content (-> XFAIL on Linux, PASS on Windows).

I verified all tests on both Linux AND Windows.

### AI-in-the-loop
I designed the command; Claude wrote the prose for the design doc. Claude drafted both tests and production code; I rewrote or refined all tests and accepted the final version of Claude's production code unmodified. I managed the commit history; Claude assisted with the prose. I reviewed all code and prose. Claude wrote the changelog; I wrote this PR.

PR #7881, opened 2026-06-29
BF: clean up synced/ so discarded commits stay dead on adjusted branches

just-meng:
Closes #7772. Stacked on #7880 — please review/merge that one first; its commits show in this PR until it lands.

On an adjusted branch (Windows / crippled filesystem) a stale or leftover `synced/<branch>` ref is merged back by the next `git annex sync` (run internally by `save`/`push`), resurrecting commits the user intentionally discarded ("zombie commits"). Two code paths leave one behind:

- `update --how=reset` reset the corresponding branch but left `synced/<branch>` at the pre-reset tip.
- `update --how=merge` created a `synced/<branch>` and never removed it.

**Fix:**

- `_annex_reset_target`: force-point `synced/<branch>` at the reset target.
- `_annex_sync`: delete the `synced/<branch>` it created, matching `AnnexRepo.localsync`'s hygiene.

**Tests:** `test_update_reset_adjusted_resets_synced` (plus no-resurrection across `save` and `push`), `test_update_merge_adjusted_no_leftover_synced`, and `test_update_reset_dirty` un-skipped on adjusted branches.

Issue #7772, opened 2025-12-23
git annex sync automatically reverts git reset --hard operations on crippled FS


just-meng:
### What is the problem?

This is actually related to a problem I tried to troubleshoot with @yarikoptic on my Windows computer. I could not update my nested datasets, it complained about merge conflict in a submodule. Solving the merge conflict in the besaid submodule did not enable updating the superds. Also `git reset --hard` in both the submodule and the superds to a good commit before the merge conflict did not help. We ended up thinking that the problem was caused by the fact that I got myself on 'master' by accident some time ago and did not notice that, but I should be on the adjusted branch. We also started thinking maybe I should do `git annex sync` after installing and `datalad save` to keep the adjusted branch in sync with master, see #7769. However, I think `git annex sync` might have caused the original problem in some more convoluted manner. I'm really not sure about this, but this behavior is weird: `git annex sync` automatically reverts `git reset --hard` operations! It has nothing to do with submodules, can be reproduced in a not nested dataset, and it does not happen on my Linux Mint system.

```bash
jiame@RAZOR MINGW64 /d/test
$ bash git-annex-sync-reverts-git-reset.sh
=== Creating a datalad dataset ===
[INFO] Detected a filesystem without fifo support.
[INFO] Disabling ssh connection caching.
[INFO] Detected a crippled filesystem.
[INFO] Entering an adjusted branch where files are unlocked as this filesystem does not support locked files.
[INFO] Switched to branch 'adjusted/master(unlocked)'
create(ok): D:\test\minimal_dataset (dataset)

=== Add a file ===
add(ok): books_for_december.txt (file)
save(ok): . (dataset)
action summary:
  add (ok: 1)
  save (ok: 1)

=== Before git reset ===
commit 947ad3ab32dedba4edfc6f227f4448d673a78c19 (HEAD -> adjusted/master(unlocked))
Author: Jiameng Wu <jiameng.wu@gmail.com>
Date:   Tue Dec 23 20:12:19 2025 +0100

    git-annex adjusted branch

 books_for_december.txt | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

commit 3f594aa3fa2a66c2fd35ce4a4b0e2373e79a7a51 (master)
Author: Jiameng Wu <jiameng.wu@gmail.com>
Date:   Tue Dec 23 20:12:19 2025 +0100

    add text file

 books_for_december.txt | 1 +
 1 file changed, 1 insertion(+)

=== Reset to previous commit ===
[...]

PR #7904, opened 2026-08-22
BF: nested datalad run - fix crash when command commits in a subdataset below a plain dir

asmacdo:
When using the new `datalad run` functionality in 1.6+ and both conditions met: 
 - condition 1: subdatasets are separated by a plaindir (non-dataset), ie `root/plaindir//subrepo`,
 - condition 2: the command commits (ie datalad run)
 
The inner datalad run works, and is committed to the leaf, and the root.
The outer datalad run crashes, so its commit is missing from the root. 
`[ERROR  ] 'NoneType' object has no attribute 'get_hexsha' `

```
datalad run "datalad run \"echo 'broke' > plaindir/subrepo/t2\""
run(ok): /tmp/datalad-run-byREmV (dataset) [echo 'broke' > plaindir/subrepo/t2]
add(ok): t2 (file)                                                                                            
save(ok): plaindir/subrepo (dataset)                                                                          
add(ok): plaindir/subrepo (dataset)                                                                           
add(ok): .gitmodules (file)                                                                                   
save(ok): . (dataset)                                                                                         
run(ok): /tmp/datalad-run-byREmV (dataset) [datalad run "echo 'broke' > plaindir/sub...]
[ERROR  ] 'NoneType' object has no attribute 'get_hexsha' 
```

It can also leave the root repo dirty, if the command commits but doesnt propagate the commit.

## Detailed Reproducer 

<details>
<summary>script</summary>

```console
# datalad run: crash when the command commits in a subdataset below a plain directory
#   root / plaindir // subrepo     //  dataset boundary, /  ordinary directory
# Run with a datalad that still has the bug (1.6.0 - 1.6.2, or master before the fix).
# No `set -e`: both `datalad run` calls are EXPECTED to fail.
export PS4='> '
set -xu
cd "$(mktemp -d ${TMPDIR:-/tmp}/dl-nested-XXXXXXX)"

datalad create -c text2git root
datalad create -c text2git -d root root/plaindir/subrepo
cd root

# 1. nested run: inner run saves the whole chain, so only the OUTER record is lost
datalad run "datalad run \"echo 'broke' > plaindir/subrepo/t2\""
git log --oneline -1
datalad status -r

# 2. bare commit: nothing saves upward, so no run record at either level
datalad run -m outer -- git -C plaindir/subrepo commit --allow-empty -m inner
git log --all --grep=RUNCMD --oneline
datalad status -r
```
</details>

## Proposed change

`_inject_sub_info` now climbs to the nearest **dataset** ancestor instead of the immediate parent directory, via a new `_parent_ds_path` helper.  Plain directories in between are skipped rather than treated as datasets. 

<details>
<summary>Claude Details</summary>

The merge-commit machinery walks up by directory and assumes every ancestor is a dataset; a plain directory in the chain has no repo, so asking it for its HEAD raises.

```
[ERROR] 'NoneType' object has no attribute 'get_hexsha'
  File "datalad/core/local/save.py", line 120, in _inject_sub_info
    Dataset(parent_path).repo.get_hexsha())
```

The walk exists to record, at every level between the changed subdataset and the run's dataset, that a child moved. It stepped by filesystem parent and asked each stop for its HEAD, which only holds if every subdataset is a direct child of its superdataset.

Two further points on the change itself: the `child_repo ... else cur_sha` fallback is dropped, since with plain directories skipped every step of the walk is a dataset, and that fallback was only ever reached for the plain-directory case where it substituted the changed subdataset's own SHA for the directory. `_parent_ds_path` returning `None` absorbs the old defensive "not a descendant" loop termination.

### Which layouts are affected

Writing `//` for a dataset boundary and `/` for an ordinary directory:

| layout | affected |
| --- | --- |
| `root//leaf` | no |
| `root//mid//leaf` | no |
| `root/plaindir//leaf` | **yes** |
| `root//mid/plaindir//leaf` | **yes** |

Extra dataset levels are not the problem; a single ordinary directory anywhere in the chain is enough. This is why existing coverage missed it: the tests exercise a direct child and an all-datasets three-level chain, but never a plain directory in between.

### What it leaves behind

The run always exits 1, but how much is lost depends on what the command did.

In the nested-run case, the inner run saves the whole chain on its way out, so the working tree ends **clean** and the only casualty is the outer run's record: every run record present describes the *inner* command. Because the outer command was a pure wrapper, that record still happens to describe the net effect, which makes the failure easy to dismiss as noise. It is not: the point of the run-merge is that the first-parent chain reads as one commit per `run`, and here the mainline ends up describing something the user never typed.

With the fix, that nested run exits 0, the root's mainline records the outer command, and the inner run's own record is preserved on the merge's second parent:

```console
$ git log --first-parent --oneline -2
e4fcd2c [DATALAD RUNCMD] datalad run "echo 'broke' > plaindir/sub...
a77eda0 [DATALAD] Created subdataset

$ git log --oneline HEAD^2 -2
0453193 Remaining changes after command execution
7c555fa [DATALAD RUNCMD] echo 'broke' > plaindir/subrepo/t2
```

### Affected versions

Present in 1.6.0, 1.6.1, 1.6.2 and current `maint` / `master`.
Introduced by #7821: `6dfa285bd` added the walk with a guard on the HEAD lookup, and `69f1530ea` two days later removed that guard, turning a latent wrong-data path into a hard crash.
Restoring the guard alone would not be correct, since the walk would still enter plain directories into the save plan as datasets.

### Verification

- New test fails without the fix and passes with it.
- `datalad/core/local/tests/test_run.py` + `test_save.py`: 78 passed, 3 skipped, 1 xfailed.
- `datalad/local/tests/test_rerun.py`: 21 passed, 1 xfailed.
- Green against both git-annex 10.20250630 and 10.20260717.
- Checked end to end that the run now produces merge commits at both levels, a run record on the superdataset, and a clean recursive status, including with three plain directories in the chain.

</details>

## Known limitation on adjusted branches

Testing this led to finding another bug #7905, so I was not able to fully test crippled filesystem. (On adjusted branches, I've marked that test as xfail)

### DIFF
--- a/datalad/core/local/save.py
+++ b/datalad/core/local/save.py
@@ -75,8 +75,12 @@ def _parent_ds_path(path, ds_path):
 
     Returns None when `path` is not underneath `ds_path` -- which
     `_parse_sub_status` from ``git submodule status --recursive`` never
-    produces, but a future caller might.
+    produces, but a future caller might.  Checked up front, so a stray
+    path that happens to sit inside some *other* dataset does not send
+    the walk climbing an unrelated hierarchy.
     """
+    if not ut.Path(path).is_relative_to(ds_path):
+        return None
     parent = ut.Path(path).parent
     while True:
         parent_path = str(parent)
--- a/datalad/core/local/save.py
+++ b/datalad/core/local/save.py
@@ -65,6 +65,31 @@ def _status_props(result):
     return {k: v for k, v in result.items() if k not in _STATUS_SKIP_KEYS}
 
 
+def _parent_ds_path(path, ds_path):
+    """Nearest ancestor of `path` that is a dataset, at or under `ds_path`
+
+    A subdataset need not be a direct child of its superdataset -- plain
+    directories can sit in between (``study/derivatives/thing``).  Those
+    have no repo, hence no HEAD, and must be skipped rather than treated
+    as datasets.
+
+    Returns None when `path` is not underneath `ds_path` -- which
+    `_parse_sub_status` from ``git submodule status --recursive`` never
+    produces, but a future caller might.
+    """
+    parent = ut.Path(path).parent
+    while True:
+        parent_path = str(parent)
+        if parent_path == ds_path:
+            return parent_path
+        if parent == parent.parent:
+            # filesystem root reached without hitting ds_path
+            return None
+        if Dataset(parent_path).repo is not None:
+            return parent_path
+        parent = parent.parent
+
+
 def _inject_sub_info(paths_by_ds, since_map, _since_sub_info, ds_path):
     """Inject changed subdatasets that diff_dataset missed on adjusted branches.
 
@@ -91,22 +116,16 @@ def _inject_sub_info(paths_by_ds, since_map, _since_sub_info, ds_path):
             paths_by_ds[sub_path] = {}
 
         # Walk up to ds_path, registering each child as modified in
-        # its parent and ensuring since_map entries exist at each level.
-        # Termination: normally `parent_path == ds_path`.  Defensive
-        # second termination on `parent_path == child_path` covers the
-        # case where `sub_path` is not a descendant of `ds_path` (which
-        # `_parse_sub_status` from `git submodule status --recursive`
-        # never produces, but a future caller might) — without it,
-        # `Path('/').parent == Path('/')` would loop forever.
+        # its parent dataset and ensuring since_map entries exist at
+        # each level.
         child_path = sub_path
-        while True:
-            parent_path = str(ut.Path(child_path).parent)
-            if parent_path == child_path:
+        while child_path != ds_path:
+            parent_path = _parent_ds_path(child_path, ds_path)
+            if parent_path is None:
                 lgr.debug("sub %s not under ds %s; aborting ancestor walk",
                           sub_path, ds_path)
                 break
-            child_repo = Dataset(child_path).repo
-            child_sha = child_repo.get_hexsha() if child_repo else cur_sha
+            child_sha = Dataset(child_path).repo.get_hexsha()
             child_pre = _since_sub_info.get(child_path, child_sha)
 
             pds_status = paths_by_ds.setdefault(parent_path, {})
@@ -119,8 +138,6 @@ def _inject_sub_info(paths_by_ds, since_map, _since_sub_info, ds_path):
                     parent_path,
                     Dataset(parent_path).repo.get_hexsha())
 
-            if parent_path == ds_path:
-                break
             child_path = parent_path
 
 
--- a/datalad/core/local/tests/test_run.py
+++ b/datalad/core/local/tests/test_run.py
@@ -1235,6 +1235,32 @@ def test_run_merge_three_levels(path=None):
     ok_((leaf.pathobj / "foo").exists())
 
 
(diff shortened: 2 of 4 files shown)


asmacdo:
@yarikoptic could you have a look when you get a chance?

datalad/core/local/save.py
```diff
@@ -65,6 +65,31 @@ def _status_props(result):
...
+    A subdataset need not be a direct child of its superdataset -- plain
+    directories can sit in between (``study/derivatives/thing``).  Those
+    have no repo, hence no HEAD, and must be skipped rather than treated
+    as datasets.
+
+    Returns None when `path` is not underneath `ds_path` -- which
+    `_parse_sub_status` from ``git submodule status --recursive`` never
+    produces, but a future caller might.
+    """
+    parent = ut.Path(path).parent
```
asmacdo, line 84: drops the defense for "subdataset not actually in the parent"

datalad/core/local/tests/test_run.py
```diff
@@ -1235,6 +1235,34 @@ def test_run_merge_three_levels(path=None):
...
+    _assert_run_merge(sub)
+    _assert_run_merge(ds)
+    ok_((sub.pathobj / "foo").exists())
+
+    if ds.repo.is_managed_branch():
+        # On adjusted branches `git annex sync` propagates a submodule
+        # pointer update to the corresponding branch only for submodules at
+        # the repo's top level, so the super keeps reporting this one
+        # modified.
+        # https://github.com/datalad/datalad/issues/7905
```
yarikoptic, line 1259: @just-meng has done a lot of work chasing adjusted branches issues recently among

- https://github.com/datalad/datalad/pulls?q=is%3Aopen+is%3Apr+label%3Aadjusted-branches

so might be related or even addressed there already! might be worth checking with claude on this particular aspet

yarikoptic-gitmate: Checked — #7880, #7881, #7884 (@just-meng's adjusted-branches work) fix a different bug: `update`'s own `git annex sync` call leaving a stale `synced/<branch>` ref that a later sync can resurrect (gh-7772). They only touch `datalad/distribution/update.py`, `reset.py`, and `utils_pytest.py`.

That's not the same mechanism as #7905: there, plain `git annex sync` (no datalad `update`/`save` involved) fails to propagate a *nested* submodule pointer from the adjusted branch to the corresponding branch. None of those three PRs touch that path, and there's no other open PR addressing gh-7905 yet. So the `xfail` marker this PR adds for that case stands on its own — it's not already covered elsewhere.

Also checked drift against `master`/`maint`: since this PR was opened, `maint` has only gained one unrelated doc commit, and `save.py`/`test_run.py`/`test_save.py` are currently byte-identical between `maint` and `master` — nothing to rebase against.

One thing worth flagging for sequencing: #7901 ("run/rerun --explicit: dirty inputs, nested runs, concurrent runs") also touches `save.py` (a `_partial_commit` param, unrelated section) and appends tests to `test_run.py`. No logical overlap with this PR's `_inject_sub_info`/`_parent_ds_path` change, but whichever PR merges second will hit a trivial textual conflict at the end of `test_run.py` since both append new tests there.

---

yarikoptic-gitmate:
PR released in [`1.6.3`](https://github.com/datalad/datalad/releases/tag/1.6.3)

Issue #7902, opened 2026-08-21
run/rerun: add explicit --on-cmd-failure


yarikoptic:
- #6447 

addressed the good old

- #3071

allowing to proceed and `save` results when external command exits with non-0.

Overall then, `datalad run` also exits with `0` upon completion.

IMHO this solution is neat but suboptimal as

- it groups **all** potential failures into the same global `--on-failure` handling whenever we are interested in particular in the status of external command execution
- we cannot instruct to pass original exit code out

thus I think we could introduce `--on-cmd-failure error,save,all` which would not be exclusive but combination switch, so we could say `save` and it would be current global `--on-failure ignore` in effect (but without masking), without proxying error, or `all` which would do both save and proxy original error.

yarikoptic-gitmate:
Issue fixed in [`1.6.3`](https://github.com/datalad/datalad/releases/tag/1.6.3)

Issue #6427, opened 2022-02-07
`run` ignores its own input spec error and proceeds


mih:
```
% datalad run -i bogus "echo 123 > mike2"
[INFO   ] Making sure inputs are available (this may take some time) 
run(error): /tmp/runny (dataset) [Input did not match existing file: bogus]
[INFO   ] == Command start (output follows) ===== 
[INFO   ] == Command exit (modification check follows) ===== 
add(ok): mike2 (file)                                                                                                       
save(ok): . (dataset)  
-> exit 1                                                                                                     
```

I would argue that if I explicitly specify an input, I rely on `run` to actually deliver the file content. From the docs:

> A dependency for the run. Before running the command, the content of this file will be retrieved.

Complaining about it, but then still executing the command, and modifying the dataset is unlikely to be desired in the use cases that I can come up with.

A possible resolution of #6426 would allow `run` to declare `on_failure=stop` as its default behavior. But until that is possible, it might be the best to stop processing unconditionally.

This issue is not an unresolvable thread to reliability, because the command does eventually fail, but it is easy to go unnoticed. Even with `--explicit` the input declaration has no impact on the command execution.

`rerun` behaves in a similar fashion when operating on such an invalid run-record. It shows an error message, proceeds, and eventually exits non-zero.

For the analog situation of an invalid `--output` declaration there is https://github.com/datalad/datalad/issues/3449 -- calling for a warning.

Taken together, the current implementation conceptualizes `--input` as a nice-to-have and hope-your-script has `set -e` "dependency". It seems a previous report concluded in a similar fashion https://github.com/datalad/datalad/issues/5583

Critical input from @kyleam is at https://github.com/datalad/datalad/commit/91d63ef455f8a7457224e57215b35e847cc5ccf9 thx! Seem to point into a similar direction.

bpoldrack:
All cases I can think of where it's desirable to have something specified in the record that may or may not be there, seem covered by being able to specify (and not expand in the record) globs.

So - just yielding `impossible` seems reasonable to me.

mih:
I have implemented a fix for this issue.

PR #7906, opened 2026-08-24
BF+DOC: make rerun fail like run does, and document exit-code handling

yarikoptic-gitmate:
`rerun` is routinely used as a shortcut for repeating a previous `run`, so the two should fail the same way. They did not: `Run` sets `on_failure = 'stop'`, `Rerun` inherited the global `'continue'` default. This PR makes them congruent and documents the exit-code semantics that were undocumented on both sides.

## The divergence was not a decision

- **1042e588d** (2018, `ENH: run: Support runs with non-zero exit codes`) aborted both cases explicitly: a fresh run with a non-zero exit wrote `COMMIT_EDITMSG` and raised, and `_execute_command()` raised outright when a *rerun* exited differently than recorded.
- **9595f5e50** (2021, gh-3071) turned both aborts into yielded `status="error"` records, deliberately — *"This allows callers to control whether the save happens via the standard `--on-failure` switch."*
- **0f5abc49c** (2022, for gh-6427) then gave `Run` — and only `Run` — the `'stop'` default, so a command is not executed after a failed input/output preparation.

`Rerun` was never given a matching default, so both of the 2018 behaviors were silently lost on the replay path. The rerun tests in 9595f5e50 were rewritten to assert only the error result (with `on_failure="ignore"`), so nothing covered the dataset state afterwards.

Two consequences, both verified before the change:

```
% datalad run -i in.dat 'sh -c "echo run2 >> marker"'     # origin gone
get(error): in.dat (file) [Unable to access these remotes: origin]
# command not executed, tree clean -- correct, this is gh-6427

% datalad rerun                                           # same dataset
get(error): in.dat (file) [Unable to access these remotes: origin]
run(ok): ... ; add(ok): marker ; save(ok): .
# command executed without its input, new run commit created, datalad exits 0
```

and a re-execution whose exit code did not match the recorded one still saved the modifications, where the equivalent `run` leaves them unsaved for inspection.

## Changes

- `Rerun.on_failure = 'stop'`, with the reasoning in a comment. Both commands now document `[Default: 'stop']`.
- A run commit recorded in a *different* dataset is reported `notneeded` rather than `impossible`. `run` duplicates its record into every modified subdataset, so a subdataset's history can contain run commits whose `dsid` is the superdataset's; such a commit is skipped or cherry picked instead of re-executed — the outcome 85d54231a intended ("skip or cherry pick the commit using the same logic that's used for non-run commits", and that commit relaxed `_rerun()`'s abort from any non-ok status to `error` so it would not stop the replay). Since `on_failure` dispatches on status alone, leaving it `impossible` would now abort a `--since=` replay at the first such commit. `_report()` keys on the presence of `diff` instead of that status.
- **A replay that re-executes nothing at all now fails.** Per-commit `notneeded` keeps a mixed range going, but when *every* run commit in the range belongs to another dataset, `rerun` used to exit 0 having done nothing. It now ends with an `impossible` result — which restores the failure the pre-`stop` code produced for `datalad rerun` on a single such commit, without giving up on ranges. A range with no run commits at all is untouched; `_rerun_as_results()` already reports that.
- Docs: the exit code is always in the run record; `--on-failure ignore` preserves a non-zero exit as a regular provenance record; nothing is recorded when the command did not modify the dataset (also when it failed); a re-execution is `ok` when it reproduces the recorded exit code, `error` when it exits with a *different* non-zero code, and — the one asymmetry left — always `ok` when it exits 0, so a command that stopped failing is only detectable by comparing `exit` against the record it was derived from (last entry of its `chain`).

## Tests

- `test_run_failure`: a re-execution with a mismatching exit code records nothing.
- `test_rerun_unavailable_input` (new): a rerun whose input cannot be obtained does not execute the command and records nothing.
- `test_rerun_range_with_foreign_run_commit` (new): a range replay skips a superdataset's run commit and continues with the subdataset's own one — this fails if the status goes back to `impossible` — and a replay of that commit alone raises `IncompleteResultsError` with an `impossible` result.
- `test_rerun`: the foreign-dsid expectation updated to `notneeded`.

The new assertions deliberately avoid `repo.dirty` and "a new commit exists": on an adjusted branch an output is removed rather than unlocked before a re-execution, so an appending command recreates it identically. The first CrippledFS run caught exactly that; fixed in 565bd68.

All 33 GitHub Actions checks are green on the current head `e6d11e35`, CrippledFS and the macOS matrix included. Locally `datalad/local/tests/test_rerun.py` (23 passed, 1 xfailed), `datalad/core/local/tests/test_run.py`, `test_run_procedure.py` and `interface/tests/test_docs.py` pass; codespell clean, no new flake8 findings.

## Notes for review

- This is a behavior change on `maint`: a mismatching rerun no longer commits. That is what `run` has always done. Labeled `semver-patch`, changelog snippet in `changelog.d/pr-7906.md` under Bug Fixes.
- Known, pre-existing and untouched here: under `--onto=`/`--branch=`, a foreign-dsid commit is *cherry picked*, producing a commit that carries the run record verbatim (`cmd`, `exit`, `chain: []`) although the command was never executed — the only signal is that the record's `dsid` belongs to another dataset. Making the pick visible in the result stream, and recording the picked hexsha in `chain`, are both possible; see the review thread.
- `# TODO: add --on-foreign-run=error|auto if a need arises for more detailed control` sits next to the status, for the day this needs explicit control.
- Follow-up left out on purpose: the `COMMIT_EDITMSG` recovery record is still only written for non-rerun failures (`if not rerun_info and cmd_exitcode`), so a failed rerun leaves an uncommitted state without one. Dropping that guard would complete the symmetry, but `Run.custom_result_renderer()` is what points users at the file and `Rerun` has no such renderer.
- For #7902/#7903: `--on-cmd-failure` should apply to both commands with the same default. #7903 also carries over `if cmd_exitcode and expected_exit != cmd_exitcode`, so the "recorded non-zero, now exits 0" case stays unflagged — worth deciding there. It also targets `maint` although #7902 is labeled `enhancement`.

## PR checklist

- [x] Overview of the changes provided above
- [x] Changelog snippet in `changelog.d/pr-7906.md`
- [x] Relates to #7902 / #7903 and gh-6427; deliberately no `Fixes` keyword — the divergence has no issue of its own, and gh-6427 is only referenced, not closed by this PR

https://claude.ai/code/session_01TpYt6fWVTvbRn4CH9pPBb8

### DIFF
--- a/datalad/core/local/run.py
+++ b/datalad/core/local/run.py
@@ -129,6 +129,20 @@ class Run(Interface):
     input or output preparation. This default ``stop`` behavior can be
     overridden via [CMD: --on-failure ... CMD][PY: `on_failure=...` PY].
 
+    || REFLOW >>
+    The exit code of the command is always part of the run record (``exit``
+    field), also when it is zero. With
+    [CMD: --on-failure ignore CMD][PY: `on_failure='ignore'` PY] a non-zero
+    exit does not prevent the modifications from being saved, and the
+    resulting provenance record is a regular run record that carries that
+    exit code. Like any other run, this requires the command to have
+    actually modified the dataset -- a command that leaves the dataset
+    unmodified is not recorded, whether it failed or not.
+    :command:`datalad rerun` compares the exit code of a re-execution against
+    the recorded one, see that command's documentation for the exact
+    semantics.
+    << REFLOW ||
+
     In the presence of subdatasets, the full dataset hierarchy will be checked
     for unsaved changes prior command execution, and changes in any dataset
     will be saved after execution. Any modification of subdatasets is also
--- a/datalad/local/rerun.py
+++ b/datalad/local/rerun.py
@@ -414,11 +414,16 @@ def _rerun_as_results(dset, revrange, since, branch, onto, message):
             rerun_dsid = res["run_info"].get("dsid")
             if rerun_dsid is not None and rerun_dsid != dset.id:
                 skip_or_pick(hexsha, res, "was ran from a different dataset")
-                # not a failure: the commit is cherry picked or skipped
-                # rather than re-executed, which is the correct outcome
-                # here.  Reporting it as such also keeps a range replay
-                # going, rather than aborting it under the command's
-                # 'stop' on_failure default.
+                # `run` duplicates its record into every modified
+                # subdataset, so a subdataset's history can contain run
+                # commits recorded by the superdataset.  Such a commit is
+                # cherry picked or skipped instead of being re-executed
+                # here -- the correct outcome, hence 'notneeded' and not
+                # 'impossible'.  This also matters for the 'stop'
+                # on_failure default above: it is dispatched on status
+                # alone, so an 'impossible' here would abort a range
+                # replay (`--since=`) at the first such commit instead of
+                # continuing with the ones this dataset can re-execute.
                 res["status"] = "notneeded"
             else:
                 res["rerun_action"] = "run"
@@ -619,8 +624,11 @@ def _report(dset, results):
     ds_repo = dset.repo
     for res in results:
         if "run_info" in res:
-            # no "diff" for records that will not be re-executed
-            # (e.g. recorded in a different dataset)
+            # only records that will be re-executed have a "diff" (it is
+            # set right next to the 'notneeded' above).  This used to be
+            # tested as status != "impossible", which was merely a proxy
+            # for the same condition and stopped holding when that status
+            # became 'notneeded'.
             if "diff" in res:
                 res["diff"] = list(res["diff"])
                 # Add extra information that is useful in the report but not
--- a/datalad/local/rerun.py
+++ b/datalad/local/rerun.py
@@ -71,18 +71,12 @@ class Rerun(Interface):
     matches the one recorded in the run record. A command that is on record to
     have exited non-zero (see :command:`datalad run`) is therefore reported as
     "ok" when it fails again in the same way, and as "error" when it exits with
-    a different non-zero code. A re-execution that exits with 0 is always
+    a different non-zero code -- in which case, as for :command:`datalad run`,
+    no modifications are saved. A re-execution that exits with 0 is always
     considered successful, also when the record has a non-zero exit code; such
     a change in behavior can only be detected by comparing the ``exit`` field
     of the new run record with that of the record it was derived from (the last
     entry of its ``chain``).
-    << REFLOW ||
-
-    || REFLOW >>
-    Unlike :command:`datalad run`, which stops before saving when the command
-    fails, the default ``continue`` behavior of this command saves the
-    modifications also when the exit code does not match, and reports the error
-    at the end of the operation.
     << REFLOW ||
 
     *Report mode*
@@ -113,6 +107,13 @@ class Rerun(Interface):
       dataset to a previous state. The working trees of any subdatasets remain
       unchanged.
     """
+    # `rerun` is routinely used as a shortcut for repeating a previous `run`,
+    # so it must fail in the same way: do not re-execute a command when its
+    # inputs could not be obtained (or its outputs not be prepared), and do
+    # not save the results of a command that failed in a way that was not
+    # recorded.  Can be overridden via the common 'on_failure' parameter.
+    on_failure = 'stop'
+
     _params_ = dict(
         revision=Parameter(
             args=("revision",),
@@ -413,7 +414,12 @@ def _rerun_as_results(dset, revrange, since, branch, onto, message):
             rerun_dsid = res["run_info"].get("dsid")
             if rerun_dsid is not None and rerun_dsid != dset.id:
                 skip_or_pick(hexsha, res, "was ran from a different dataset")
-                res["status"] = "impossible"
+                # not a failure: the commit is cherry picked or skipped
+                # rather than re-executed, which is the correct outcome
+                # here.  Reporting it as such also keeps a range replay
+                # going, rather than aborting it under the command's
+                # 'stop' on_failure default.
+                res["status"] = "notneeded"
             else:
                 res["rerun_action"] = "run"
                 res["diff"] = diff_revision(dset, hexsha)
@@ -613,7 +619,9 @@ def _report(dset, results):
     ds_repo = dset.repo
     for res in results:
         if "run_info" in res:
-            if res["status"] != "impossible":
+            # no "diff" for records that will not be re-executed
+            # (e.g. recorded in a different dataset)
+            if "diff" in res:
                 res["diff"] = list(res["diff"])
                 # Add extra information that is useful in the report but not
                 # needed for the rerun.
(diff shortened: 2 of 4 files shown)


datalad/core/local/run.py
```diff
@@ -129,6 +129,20 @@ class Run(Interface):
...
+    [CMD: --on-failure ignore CMD][PY: `on_failure='ignore'` PY] a non-zero
+    exit does not prevent the modifications from being saved, and the
+    resulting provenance record is a regular run record that carries that
+    exit code. Like any other run, this requires the command to have
+    actually modified the dataset -- a command that leaves the dataset
+    unmodified is not recorded, whether it failed or not.
+    :command:`datalad rerun` compares the exit code of a re-execution against
+    the recorded one, see that command's documentation for the exact
+    semantics.
+    << REFLOW ||
```
yarikoptic:
```suggestion
    || REFLOW >>
    The exit code of the command is a part of the run record (``exit``
    field), also when it is zero. With
    [CMD: --on-failure ignore CMD][PY: `on_failure='ignore'` PY] a non-zero
    exit does not prevent the modifications from being saved.
    :command:`datalad rerun` compares the exit code of a re-execution against
    the recorded one, see that command's documentation for the exact
    semantics.
    A command that leaves the dataset
    unmodified is not recorded, whether it failed or not.
    << REFLOW ||
```

yarikoptic:
`continuous-integration/appveyor/branch` reports failed on `7b747ea` (build 54604235); `continuous-integration/appveyor/pr` (build 54604237) is still pending after ~38 h.

This does not look like this PR's failure. AppVeyor is currently red on every recent open PR here, across unrelated changes and authors:

| PR | AppVeyor | reported |
|---|---|---|
| #7881 adjusted-branch `synced/` cleanup | failure | Aug 4 |
| #7884 new `reset` command | failure | Aug 4 |
| #7903 `--on-cmd-failure` | failure | Aug 23 |
| #7904 nested-run subdataset crash | failure | Aug 25 |
| #7906 (this one) | failure | Aug 27 |

Everything else on this exact commit is green: all 33 GitHub Actions checks including CrippledFS and the macOS matrix, plus Read the Docs.

I could not confirm the root cause from the build log — `ci.appveyor.com` is blocked by the network egress policy of the environment I run in, so I can neither read the log nor re-run the build. If someone with access looks and the failure turns out to be specific to this change, I will fix it; until then I have not pushed anything for it. I will keep watching the PR.

---

datalad/local/rerun.py
```diff
@@ -592,7 +619,9 @@ def _report(dset, results):
     ds_repo = dset.repo
     for res in results:
         if "run_info" in res:
-            if res["status"] != "impossible":
+            # no "diff" for records that will not be re-executed
+            # (e.g. recorded in a different dataset)
+            if "diff" in res:
```
yarikoptic, line 652: explain why these two changes are needed (impossible -> notneeded, and this one for overall behavior change to stop on failure and uniform behavior with `run`?

yarikoptic: Both are forced by the `on_failure = 'stop'` default, and 1fb11cf now says so in the code.

**`impossible` → `notneeded`.** `run` duplicates its record into every modified subdataset, so a subdataset's history can contain run commits recorded by the *superdataset* (its `dsid` is the super's). Such a commit cannot be re-executed there and is cherry picked or skipped instead — the correct outcome, not a failed operation.

That distinction did not matter while `Rerun` inherited `on_failure='continue'`: the result was collected and iteration went on. `stop` breaks on the first `impossible`/`error` (`interface/utils.py:390`), and since `_rerun()` is a generator, everything after it is never produced. Measured in a subdataset holding a superdataset run commit followed by its own:

```
# before (continue)
run(impossible): ... [2ce98d2 was ran from a different dataset; skipping]
run(ok):         ... [sh -c "echo own >> ownfile"]        # continues

# with stop, status still 'impossible'
run(impossible): ... [2ce98d2 was ran from a different dataset; skipping]
                                                          # aborts, own commit never replayed
```

`on_failure` dispatches on status alone, so there is no way to exempt one result; either the status stops being a failure or `rerun --since=` breaks for subdataset histories. `test_rerun_range_with_foreign_run_commit` fails if it goes back to `impossible`.

**This line.** The guard was `res["status"] != "impossible"`, which was only ever a proxy for "this record has a `diff`" — `diff` is set in the sibling branch of the same `if`, right where the status is set. Once that status became `notneeded`, the old guard would let the foreign-dsid record through and `list(res["diff"])` would raise `KeyError`. It now tests the actual precondition.

The one visible cost, for the record: `rerun` on such a commit alone used to print `run(impossible)` and exit 1; it is now silent (the generic renderer skips `notneeded`) and exits 0. If you want that signal back, the clean way is a final `impossible` when *no* run commit in the requested range was re-executed — a range with real work still proceeds. Happy to add it here.

---

yarikoptic: Pushed in e6d11e3: a replay that re-executed **nothing** now fails at the end, instead of exiting 0 silently.

```
% datalad rerun <foreign-dsid commit>        # in the subdataset
[INFO] skip-or-pick commit af4386f; ... was ran from a different dataset; skipping or cherry picking
run(impossible): .../sub (dataset) [no run commit in the given revision range could be re-executed in this dataset]
exit=1
```

Per-commit results stay `notneeded` so a mixed range still replays the commits this dataset *can* re-execute; the terminal result is the only place left to say the request as a whole came to nothing. A range with no run commits at all is untouched — `_rerun_as_results()` already reports that, and the new result is suppressed there so the two do not double up. `# TODO: add --on-foreign-run=error|auto if a need arises for more detailed control` sits next to the status.

Separately, on the cherry-pick half — you are right that it is the worse case, and it predates this PR. `datalad rerun --onto= --since= --branch=verify` in the subdataset produces:

```
verify:  a88f9f0 [DATALAD RUNCMD] sh -c "echo own >> own"                   <- actually re-executed
         def1c92 [DATALAD RUNCMD] sh -c "echo super >> sub/from-super"      <- cherry-picked, never ran
```

`def1c92` carries the run record verbatim — `cmd`, `exit: 0`, `chain: []` — so nothing in the commit says it was applied as a static patch rather than executed; the only signal is that its `dsid` is the superdataset's. Two things could be done about it, both outside this PR's scope unless you want them here:

1. make the pick visible in the result stream (`ok` for the cherry-pick, `notneeded` for the skip) — right now only the `[INFO] skip-or-pick …` line shows it;
2. record the provenance of the pick, e.g. append the original hexsha to `chain` in the cherry-picked commit, so a reader can tell it was carried over rather than re-run.

Say which, if any, and I will add them.

---

yarikoptic:
I think I cycled on this enough and the discovered story shows some "surprises" in handling some run records, but I think overall this PR would make it more consistent so I will proceed (one TODO was added, but not sure I would just to have it implemented ATM)

yarikoptic-gitmate:
PR released in [`1.6.3`](https://github.com/datalad/datalad/releases/tag/1.6.3)

Issue #4700, opened 2020-07-13
`rerun` does not cherry pick regular commits before first RUNCMD


yarikoptic:
The use case I had tried to pursue is (ab)using `datalad rerun` to re-establish dataset history with fake dates enabled.   There is a number of commits which were created by `datalad run` but they come late in the history. I know that most likely it would have been a failure (cherry picking commits on top of possibly changed by `rerun` data would fail) but I just thought to try.  

It might be "by design" that it does not CP all those commits `--since` until it runs into a commit which it needs to rerun.  It makes sense if `onto` is some commit inbetween `since` and `onto`.  Documentation says

    The decision to skip rather than cherry pick a revision is based on whether
    the revision would be reachable from HEAD at the time of execution.

which is not 100% clear what actually decision would be (skip or not) and either it is applicable here.

I have made the entire dataset (with all annexed files content history) available on smaug under `/mnt/btrfs/scrap/tmp/budapest` and here is  sequence of commands I had ran (under @kyleam 's account to make sure that permissions are right etc):

NB not yet sure why it is trying to reach out to the very original clone of that dataset, to which kyle has no access to (happens in the initial `clone` or `install` invocation)

```shell
(dev3) kyle@smaug:/mnt/btrfs/scrap/tmp$ datalad --version
datalad 0.13.0
(dev3) kyle@smaug:/mnt/btrfs/scrap/tmp$ datalad clone budapest budapest-fakedates
[INFO   ] Fetching updates for Dataset(/mnt/btrfs/scrap/tmp/budapest-fakedates)
CommandError: 'git fetch --verbose --progress --prune --recurse-submodules=no origin-2' failed with exitcode 128 under /mnt/btrfs/scrap/tmp/budapest-fakedates
fatal: '/mnt/btrfs/datasets/datalad/crawl/labs/gobbini/budapest' does not appear to be a git repository
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
(dev3) kyle@smaug:/mnt/btrfs/scrap/tmp$ cd budapest-fakedates/
(dev3) kyle@smaug:/mnt/btrfs/scrap/tmp/budapest-fakedates$ datalad get .heudiconv/
install(ok): /mnt/btrfs/scrap/tmp/budapest-fakedates/.heudiconv (dataset) [Installed subdataset in order to get /mnt/btrfs/scrap/tmp/budapest-fakedates/.heudiconv]
[...]

PR #7908, opened 2026-08-26
BF: rerun --since - do not silently skip leading non-run commits

yarikoptic-gitmate:
## Summary

`datalad rerun --since=<sha> --branch=<name> <revision>` should replay
the commits in `(since, revision]`. When that range starts with
non-RUNCMD (regular) commits, the current implementation silently
*skips* them, and the subsequent run-command iteration then checks out
the **original source-branch parent by SHA**. That splices the
source's full history past `<sha>` — including `<sha>` itself — back
into the rerun branch's ancestry.

Practical impact: the rerun branch shares SHAs with the source
branch's pre-`<since>` region. Subsequent rebase / amend of source
implicitly mutates the rerun branch. The most easily-observed
symptom is that files introduced only by the `--since` boundary
commit show up in the rerun branch's worktree.

Regression from 1.4.1 (PR #7729, commit `8a8d49dd4`, fix for #4700).
Still present in 1.6.0.

## Verification

The TST commit strengthens
`test_rerun_skip_regular_commits_before_first_runcmd` (the previous
form was satisfied by both pre-#4700 and post-#4700-buggy code); the
BF commit makes it green. Run:

```
pytest -x datalad/local/tests/test_rerun.py::test_rerun_skip_regular_commits_before_first_runcmd
```

against each commit to see the transition.

Full post-fix: `test_rerun.py` 23 passed, 1 xfailed; `test_rerun_merges.py` 15 passed. Determinism probe (20 trials, no timestamp pinning): **20/20** fires on unfixed code.

---

<details><summary>Mechanism — what's happening in <code>_rerun()</code></summary>

The elif at `datalad/local/rerun.py` on current maint:

```python
elif parent != head and ds_repo.is_ancestor(onto, parent):
    if rerun_action == "run":
        ds_repo.checkout(parent)
        head = parent
    else:
        _mark_nonrun_result(res, "skip")
        yield res
        continue
```

The elif serves two roles:
1. **Run commits** — checks out source parent to recreate pre-command state.
2. **Non-run commits on merge side-branches** — silent-skip preserves the
   original SHA so the upcoming merge-replay step can use it as parent-2.

The bug is that the silent-skip `else` branch also fires for leading
non-run commits under explicit `--since`.  For those commits no
cherry-pick happens and `new_bases` is never populated.  The *next*
iteration (the RUN_COMMIT) finds its `parent` unmapped, the `"run"` arm
fires, and `ds_repo.checkout(parent); head = parent` checks out the
original source-branch parent — splicing source ancestry into the rerun
branch.

**Fix**: within the `else` arm, guard on `head != onto`.  When
`head == onto` the loop has not yet replayed any run commit (we are
processing a leading non-run commit); the commit must be cherry-picked
rather than silently skipped.  When `head != onto` a run has already
been replayed and the commit is on a merge side-branch; the original
silent-skip is correct.  Leading non-run commits that fall through
record `new_bases[parent] = head` so the subsequent run iteration
finds its remapped base via `new_base = new_bases.get(parent)`.

</details>

<details><summary>Regression evidence (three-criteria justification)</summary>

Per the standard checklist:

1. **Introducing commit identified**: `8a8d49dd4` (PR #7729) in 1.4.1.
2. **Prior version behaved differently**: pre-1.4.1 the `dropwhile()`
   filter discarded leading non-run commits before they entered
   `_rerun()`'s loop, so the silent-skip elif was never reached with a
   non-run commit. Pre-1.4.1 the observable was the *original* #4700
   bug — `rerun-branch` was missing the leading commits entirely. The
   1.4.1 fix removed the filter for the explicit-`since` path, which
   exposed this latent gap.
3. **Not mentioned in the commit message**: PR #7729's commit message
   claims only the intended effect (include regular commits before
   first RUNCMD). The ancestry-leak side-effect was not identified.

</details>

<details><summary>Post-iteration short-circuit — deliberately untouched</summary>

The post-iteration branch (`if new_head not in [head, res_hexsha]:
new_bases[res_hexsha] = new_head`) is left as-is. When a cherry-pick
truly produces `new_head == res_hexsha` (identical content + metadata),
the commits are graph-equivalent, so not recording the mapping is
harmless — subsequent `new_bases.get(parent)` returns None but
`parent == head` already, and no checkout is needed. This path is not
on the failing code trajectory.

</details>

## Test plan

- [x] `pytest -x datalad/local/tests/test_rerun.py::test_rerun_skip_regular_commits_before_first_runcmd` (fails at TST commit, passes at BF commit).
- [x] `pytest datalad/local/tests/test_rerun.py datalad/local/tests/test_rerun_merges.py` full suite: 38 passed, 1 xfailed.
- [ ] CI green on all supported Python / OS combinations.
- [ ] Reviewer sanity-check on adjusted-branch scenarios (the test is `@skip_if_adjusted_branch`, matching the surrounding tests).

### DIFF
--- a/datalad/local/tests/test_rerun.py
+++ b/datalad/local/tests/test_rerun.py
@@ -1026,36 +1026,73 @@ def test_rerun_commit_message_check():
 @skip_if_adjusted_branch
 @with_tempfile(mkdir=True)
 def test_rerun_skip_regular_commits_before_first_runcmd(path=None):
-    """Test for issue #4700: regular commits before first RUNCMD are not cherry-picked"""
+    """Regression test for `rerun --since=<sha>` leading-non-run ancestry leak.
+
+    Covers two related bugs:
+
+    #4700 (fixed in 1.4.1): leading non-RUNCMD commits were silently dropped
+    by ``dropwhile()`` and never appeared on the rerun branch.
+
+    Follow-up bug introduced by the 1.4.1 fix: with an explicit ``--since``
+    boundary, non-RUNCMD commits are silently skipped by the elif at
+    ``datalad/local/rerun.py:500-507``. The *subsequent* run-command
+    iteration then checks out the ORIGINAL source-branch parent via
+    ``ds_repo.checkout(parent); head = parent``, splicing the source's
+    history past the ``--since`` boundary — including the boundary commit
+    itself — into the rerun branch's ancestry. The most easily-observed
+    symptom is that ``file1.txt`` (introduced by the boundary commit)
+    appears in the rerun-branch worktree even though its introducing commit
+    is meant to be excluded by ``--since``.
+    """
     ds = Dataset(path).create()
     initial_commit = ds.repo.get_hexsha()
 
-    # Create regular commits before any run commands
-    create_tree(ds.path, {"file1.txt": "content1", "file2.txt": "content2"})
-    ds.save(["file1.txt", "file2.txt"], message="Add files before run")
+    # Leading non-RUNCMD commit that will become the --since boundary
+    # (must be EXCLUDED from the rerun range).
+    create_tree(ds.path, {"file1.txt": "content1"})
+    ds.save("file1.txt", message="Add files before run #1")
+    files1_commit = ds.repo.get_hexsha()
 
-    # Create a run command
+    # Leading non-RUNCMD commit inside the --since range (must be picked).
+    create_tree(ds.path, {"file2.txt": "content2"})
+    ds.save("file2.txt", message="Add files before run #2")
+
+    # The RUNCMD.
     ds.run('echo "run output" > run_output.txt')
+    run_commit = ds.repo.get_hexsha()
 
-    # Create another regular commit after the run
+    # Trailing non-RUNCMD commit (covers the original #4700 scenario).
     create_tree(ds.path, {"file3.txt": "content3"})
     ds.save("file3.txt", message="Add file after run")
-
     head_commit = ds.repo.get_hexsha()
 
-    # Go back to initial commit and rerun everything
+    # Rerun {files2, run, files3}; files1 is the --since boundary and
+    # must be excluded per the docs.
     ds.repo.checkout(initial_commit, options=["--detach"])
-
-    # Rerun from the saved HEAD (issue #4700: without fix, regular commits before
-    # first RUNCMD would be dropped by dropwhile())
-    ds.rerun(revision=head_commit, since=initial_commit, branch="rerun-branch")
-
-    # Verify all files exist (with fix, regular commits are no longer dropped)
+    ds.rerun(revision=head_commit, since=files1_commit, branch="rerun-branch")
     ds.repo.checkout("rerun-branch")
-    ok_exists(op.join(path, "file1.txt"))
+
+    # Files inside the --since range were applied.
     ok_exists(op.join(path, "file2.txt"))
     ok_exists(op.join(path, "run_output.txt"))
     ok_exists(op.join(path, "file3.txt"))
+
+    # Primary observable of the ancestry-leak bug: without the fix,
+    # rerun-branch's ancestry includes the --since boundary commit, so
+    # its file (file1.txt) shows up in the worktree. This assertion fires
+    # deterministically against unfixed code (verified 20/20 in an
+    # empirical probe without date pinning).
+    assert not op.exists(op.join(path, "file1.txt")), \
+        "file1.txt should be excluded (its commit is the --since boundary)"
+
+    # Belt-and-braces: neither the --since boundary nor the original run
+    # commit should appear in the rerun-branch ancestry.
+    rerun_ancestry = ds.repo.call_git(
+        ["log", "--format=%H", "rerun-branch"]).split()
+    assert files1_commit not in rerun_ancestry, \
+        f"--since boundary {files1_commit[:7]} leaked into rerun-branch"
+    assert run_commit not in rerun_ancestry, \
+        f"original run commit {run_commit[:7]} leaked into rerun-branch"
 
 
 # -- Tests for rerunning run-merge commits --
--- a/changelog.d/pr-7908.md
+++ b/changelog.d/pr-7908.md
@@ -1,0 +1,8 @@
+### 🐛 Bug Fixes
+
+- `rerun --since=<sha>` no longer leaks the source branch's pre-`<since>`
+  ancestry into the rerun branch when the replayed range starts with
+  non-`run` commits.  Regression introduced in 1.4.1 by the fix for
+  [#4700](https://github.com/datalad/datalad/issues/4700).
+  [PR #7908](https://github.com/datalad/datalad/pull/7908)
+  (by [@yarikoptic](https://github.com/yarikoptic))
--- a/datalad/local/rerun.py
+++ b/datalad/local/rerun.py
@@ -540,10 +540,20 @@ def _rerun(dset, results, assume_ready=None, explicit=False, jobs=None):
             if rerun_action == "run":
                 ds_repo.checkout(parent)
                 head = parent
-            else:
+            elif head != onto:
+                # Non-run commit on a merge side-branch: silent-skip to keep
+                # the original SHA available for the upcoming merge commit.
                 _mark_nonrun_result(res, "skip")
                 yield res
                 continue
+            else:
+                # Leading non-run commit under explicit --since (head still at
+                # onto, no run replay yet).  Previously silent-skipped here,
+                # which left new_bases unpopulated and caused the subsequent
+                # run iteration to checkout the source-branch parent, splicing
+                # source ancestry into the rerun branch.  Fix: record the base
+                # mapping and let skip-or-pick cherry-pick it below.
+                new_bases[parent] = head
         else:
             if parent != head:
                 new_bases[parent] = head


yarikoptic:
@yarikoptic-gitmate since sent out from a fork, please add proper scriv changelog record! See documentation in the repo on how to do that.

yarikoptic-gitmate (commented):
## Review

Checked out this PR's head (`refs/pull/7908/head`) locally against `maint` and ran the affected suites directly:

```
pytest datalad/local/tests/test_rerun.py datalad/local/tests/test_rerun_merges.py
38 passed, 1 xfailed
```

matching the PR description's own numbers. `test_rerun_skip_regular_commits_before_first_runcmd` correctly exercises the claimed regression (leading non-`run` commit under an explicit `--since` boundary no longer leaks into the rerun branch's ancestry).

**Changelog**: `changelog.d/pr-7908.md` (added in `63499e5`, *after* the "please add a changelog" comment above) already follows the project's scriv format correctly (category header, issue/PR links, byline) — matches sibling fragments like `changelog.d/pr-7906.md`. The `CHANGELOG-missing` label looks stale rather than reflecting a real gap.

**Correctness**: the `head != onto` guard correctly distinguishes "no run replayed yet" (leading non-run commit, must cherry-pick) from "on a merge side-branch after a run replay" (must keep silent-skipping to preserve the SHA for the merge-replay step) — traced through `_rerun()`'s control flow and confirmed the merge-focused suite (`test_rerun_merges.py`) still passes in full.

**Minor, non-blocking reuse suggestions** (two independent reviews converged on these):
- `rerun.py`: the new branch's `new_bases[parent] = head` duplicates the outer catch-all `else: if parent != head: new_bases[parent] = head` a few lines below. Adding `and head != onto` to the `elif parent != head and ds_repo.is_ancestor(onto, parent):` guard would let this case fall through to that existing assignment instead of introducing a second write site.
- `test_rerun.py`: `ds_repo.call_git(["log", "--format=%H", "rerun-branch"]).split()` reinvents `ds.repo.get_revisions("rerun-branch")`, which this same test file already uses elsewhere (and `rerun.py` itself uses internally). The bare `assert x not in y, "..."` calls could likewise use the already-imported `assert_not_in` helper for consistency with the rest of the file.
- A couple of the added comments/docstring are longer than this file's usual terse style and restate history already captured in the PR description/changelog — could be trimmed, purely cosmetic.

**Outstanding, not code-related**: the `continuous-integration/appveyor/pr` check is red on the current head while all 33 GitHub Actions jobs (including the full Linux test matrix) are green. I wasn't able to reach `ci.appveyor.com` from this environment to inspect the actual failure log, so I can't tell whether it's a genuine Windows-specific regression or an unrelated flake — worth a look/re-run before merging.

None of the above are blocking; happy to see this land once AppVeyor is sorted out.

https://claude.ai/code/session_01WEWA192TK14bhAc4iZLywd

---

yarikoptic-gitmate (commented):
Second independent review (correctness-focused, with its own empirical verification) came back **CONFIRMED** as well:

- Traced the control flow through `_rerun()` and confirmed the leading non-run commit correctly falls through to the shared skip-or-pick cherry-pick logic via `new_bases[parent] = head`.
- Reverted just the `rerun.py` hunk from `9000ed0` (keeping the strengthened test): reproduced the exact claimed failure (`file1.txt` leaking in). Restored the fix: green again, full suite 38 passed / 1 xfailed.
- Additional edge cases exercised directly (not just read): 3+ consecutive leading non-run commits, and a leading merge of two non-run side-branches before the first run commit — both correct, no ancestry leak, side-branch SHAs correctly preserved.

One **theoretical, unconfirmed** edge case flagged for awareness rather than as a blocker: the existing `head_to_restore, head = head, new_base` detour-and-restore mechanism (used together with the skip path a few lines up) could in principle restore `head` back to exactly `onto` after a temporary detour, which would make a genuine merge-side-branch commit evaluate as `head == onto` and get mis-cherry-picked by this fix instead of silently preserved. No concrete repro was found despite trying, and it looks like a pre-existing sharp edge in the surrounding merge-restore logic rather than something this PR introduces — worth a mention for a maintainer familiar with that code path, not something I'd hold the PR on.

This corroborates the first review's duplication note: the new `new_bases[parent] = head` write is confirmed to be redundant with the existing catch-all `else: if parent != head: new_bases[parent] = head` a few lines below (chaining across multiple leading commits is actually carried by the pre-existing end-of-loop `new_bases[res_hexsha] = new_head` update, not by this new write) — so consolidating into the guard condition (`and head != onto`) rather than a new branch remains a safe, minor cleanup opportunity.

https://claude.ai/code/session_01WEWA192TK14bhAc4iZLywd

---

yarikoptic-gitmate:
**CI status**: `continuous-integration/appveyor/pr` is red on the current head (`63499e5`), but this is not this PR's failure. Cross-checking other, unrelated, currently-open PRs (e.g. #7903) shows the exact same `continuous-integration/appveyor/pr`/`appveyor/branch` checks failing there too, against completely different diffs — pointing at an infra-wide AppVeyor issue rather than anything introduced by this change. All 33 GitHub Actions jobs (the full Linux/macOS/Windows-via-Actions test matrix) are green on this head, and both independent subagent reviews above confirm the fix and its test coverage.

I don't have network access to `ci.appveyor.com` from this environment (egress-blocked) to inspect the actual failure log or trigger a re-run, so I can't push a fix for it here. Given it reproduces identically on an unrelated PR, I'm treating it as a pre-existing/infra flake rather than something for this PR to address — happy to revisit if a maintainer confirms otherwise. This PR is otherwise ready (fix verified, tests green, changelog present).
Claude-Session: https://claude.ai/code/session_01J2S6eGdZyn3X7AmawfhJ4s

---

yarikoptic-gitmate:
## Status check + one more minor cleanup applied where I could

Re-checked this PR's current state (head `63499e5`):

- **CI**: unchanged from the last update — all 33 GitHub Actions checks and Read the Docs are green; `continuous-integration/appveyor/pr` is still red, and remains the same pre-existing infra-wide AppVeyor breakage documented above and tracked separately in #7912 (reproduces identically on unrelated PRs). Nothing new to fix on my side.
- **Reviews**: no new comments since the two independent review rounds above (both CONFIRMED); nothing outstanding from a human reviewer either.

Of the two non-blocking review suggestions, I could safely apply one:

- `test_rerun.py`'s `rerun_ancestry` check now uses the already-used-elsewhere `ds.repo.get_revisions("rerun-branch")` helper and the already-imported `assert_not_in()` instead of a raw `call_git(["log", ...])` + bare asserts. No behavior change; verified locally: `test_rerun_skip_regular_commits_before_first_runcmd` still passes, and the full `test_rerun.py` + `test_rerun_merges.py`: 38 passed, 1 xfailed (matching this PR's documented baseline).

I could **not** push that commit here, though: this session's GitHub access is scoped to `datalad/datalad` only, and this PR's branch lives on `yarikoptic/datalad` (a fork outside that scope). Patch below for whoever has push access to that branch to apply directly, if desired — it's cosmetic only, so not blocking either way:

```diff
--- a/datalad/local/tests/test_rerun.py
+++ b/datalad/local/tests/test_rerun.py
@@ -1087,12 +1087,9 @@ def test_rerun_skip_regular_commits_before_first_runcmd(path=None):
 
     # Belt-and-braces: neither the --since boundary nor the original run
     # commit should appear in the rerun-branch ancestry.
-    rerun_ancestry = ds.repo.call_git(
-        ["log", "--format=%H", "rerun-branch"]).split()
-    assert files1_commit not in rerun_ancestry, \
-        f"--since boundary {files1_commit[:7]} leaked into rerun-branch"
-    assert run_commit not in rerun_ancestry, \
-        f"original run commit {run_commit[:7]} leaked into rerun-branch"
+    rerun_ancestry = ds.repo.get_revisions("rerun-branch")
+    assert_not_in(files1_commit, rerun_ancestry)
+    assert_not_in(run_commit, rerun_ancestry)
```

I deliberately left the other suggestion (consolidating the `new_bases[parent] = head` write site in `rerun.py` by folding the `head != onto` check into the outer `elif` guard) unapplied: on closer look it isn't behavior-preserving — it would also gate the `rerun_action == "run"` checkout branch on `head != onto`, which the current code intentionally does not do. Given the second review's own flagged (unconfirmed) edge case around `head` being restored to `onto` by the detour mechanism, I'd rather not touch that condition without a maintainer confirming the intended behavior there. Leaving it as-is.

This PR otherwise looks ready from my side, pending only the AppVeyor infra issue and a maintainer merge decision.

---

yarikoptic-gitmate:
@yarikoptic re: the scriv changelog record you asked for — it's still not there (the `CHANGELOG-missing` label is still on the PR, and the auto-generate-from-title workflow can't push to a fork branch for the same reason I couldn't push the `test_rerun.py` cleanup in my earlier comment: this session's write access doesn't extend to `yarikoptic/datalad`).

Proposed `changelog.d/pr-7908.md`, matching the format already used by e.g. `pr-7906.md`:

```markdown
### 🐛 Bug Fixes

- `rerun --since=<sha>` no longer silently skips leading non-run commits
  when replaying a range, which previously spliced the source branch's
  history past `<sha>` into the rerun branch's ancestry.
  [PR #7908](https://github.com/datalad/datalad/pull/7908)
  (by [@yarikoptic](https://github.com/yarikoptic))
```

Please add this file to the PR branch directly (or apply it however you'd prefer) — I don't have push access to the fork to do it myself.

---

yarikoptic-gitmate:
Correction to my previous comment: the scriv changelog record is actually already present on this branch — `changelog.d/pr-7908.md` was added in `63499e5` (confirmed just now via `pull_request_read get_files` on the current head), and matches the project's format:

```markdown
### 🐛 Bug Fixes

- `rerun --since=<sha>` no longer leaks the source branch's pre-`<since>`
  ancestry into the rerun branch when the replayed range starts with
  non-`run` commits.  Regression introduced in 1.4.1 by the fix for
  [#4700](https://github.com/datalad/datalad/issues/4700).
  [PR #7908](https://github.com/datalad/datalad/pull/7908)
  (by [@yarikoptic](https://github.com/yarikoptic))
```

This was also already confirmed present in the first review round on 2026-09-05. My 2026-09-09 comment claiming it was still missing was mistaken (and the proposed replacement file in that comment is redundant — please disregard it). The `CHANGELOG-missing` label on the PR appears to just be stale bot state, not a real gap.

No other change in status: all 33 GitHub Actions checks remain green; `continuous-integration/appveyor/pr` remains red for the same infra-wide reason documented above (reproduces identically on unrelated PRs, tracked in #7912). Nothing further needed from this side.

---

datalad/local/tests/test_rerun.py
```diff
@@ -1026,37 +1026,74 @@ def test_rerun_commit_message_check():
...
+    boundary, non-RUNCMD commits are silently skipped by the elif at
+    ``datalad/local/rerun.py:500-507``. The *subsequent* run-command
+    iteration then checks out the ORIGINAL source-branch parent via
+    ``ds_repo.checkout(parent); head = parent``, splicing the source's
+    history past the ``--since`` boundary — including the boundary commit
+    itself — into the rerun branch's ancestry. The most easily-observed
+    symptom is that ``file1.txt`` (introduced by the boundary commit)
+    appears in the rerun-branch worktree even though its introducing commit
+    is meant to be excluded by ``--since``.
+    """
```
yarikoptic, line 1046: this long paragraph is not really appropriate here -- shrink and do not point to the specific lines!

But also logic here is flawed on "The most easily-observed symptom". Since we are not talking about `--onto` mode of operation, with pure `--since=<sha> --branch <branch>` that , `<sha>` will be the 'root' (common commit) for both current branch and that new `<branch>`, so the `file1.txt` will be present there!  FWIW, if command executions are reproducible, the end tree and list of commits (not their hexsha, since we cherry-pick or rerun) must be the same as of the original branch!  But it should be a separate branch as we are rerunning and cherry-picking without feeding the same dates etc metadata as of the original commits, thus not making them fully reproducible to their hexsha!

So, is the whole PR logic flawed here with such a test?

yarikoptic-gitmate: Good catch to double check — but the logic isn't flawed here, and neither is the test's expectation. The key fact: **without `--onto`, rerun roots the new branch at whatever is currently checked out (`HEAD`) at invocation time — not at `--since`'s `<sha>`.** From `_rerun_as_results()`:

```python
start_point = onto or "HEAD"
...
commit=ds_repo.get_hexsha(start_point),   # this becomes the "checkout" result
```

and in `_rerun()`, the checkout action does `ds_repo.checkout(res_hexsha, ...)` then `head = onto = res_hexsha`. So "root" = current `HEAD`, unconditionally, when `--onto` is not given. `--since` only selects *which commits get replayed* (`git log SINCE..REVISION`); it plays no role in picking the root.

The test checks out `initial_commit` (**before** `file1.txt`) right before calling `ds.rerun(...)`, so that becomes the root — functionally identical to passing `--onto=initial_commit` explicitly, just via pre-checkout instead of the flag. Since `initial_commit` predates `file1.txt`, `file1.txt` correctly must not appear.

I verified both directions empirically rather than just re-reading the code:

- **As the test has it** (checked out to `initial_commit` first, matching `--onto=initial_commit`): `file1.txt` does *not* appear — correct, and this is exactly the scenario the ancestry-leak bug corrupts without the fix.
- **Your scenario** — plain `--since=<sha> --branch=<name> <revision>` with **no pre-checkout**, i.e. `HEAD` sitting at the tip (`head_commit`) as it normally would: `file1.txt` *does* appear, but for a mundane reason — `head_commit`'s worktree already contains `file1.txt` as ordinary tracked content (it's a descendant of `files1_commit`), same as `git checkout -b new_branch` would from any commit. Nothing to do with the ancestry-leak bug at all.

So the two scenarios aren't in tension — they're rooted at different commits (`initial_commit` vs. `head_commit`), which is exactly why the outcome differs. `--onto` (explicit or via pre-checkout) is precisely what makes the boundary-exclusion of `--since` meaningful for the *tree contents*, not just the replayed-commit list.

On reproducibility: agreed in general (same tree/commit-list, different hexshas from re-dated cherry-picks) — but that's orthogonal here, since `file1.txt`'s introducing commit is outside the replayed range in both scenarios; the only question is whether it's already in the base tree.

Shrunk the docstring per your other comment (dropped the line-number pointer, trimmed the paragraph) and added a short, precise note on the `--onto`-defaults-to-`HEAD` point above, to head off exactly this question in the future:

```python
    """Regression test for `rerun --since=<sha>` leading-non-run ancestry leak.

    Covers two related bugs: #4700 (fixed in 1.4.1), where leading
    non-RUNCMD commits were silently dropped and never appeared on the
    rerun branch; and a follow-up where, with an explicit ``--since``
    boundary, such a leading commit was silently skipped without recording
    a base mapping, so the next RUNCMD's checkout fell back to its
    *original* source-branch parent -- splicing history past the
    ``--since`` boundary (including the boundary commit itself) into the
    rerun branch's ancestry.

    HEAD is checked out to ``initial_commit`` before calling rerun, which
    (absent an explicit ``--onto``) is where rerun roots the new branch --
    equivalent to ``--onto=initial_commit`` rather than some special mode --
    and initial_commit predates file1.txt, hence the assertion.
    """
```

Verified: `datalad/local/tests/test_rerun.py` + `test_rerun_merges.py` still 38 passed, 1 xfailed with this change.

I can't push it myself right now — this PR's head (`bf-rerun-since-ancestry-leak`) lives on `yarikoptic/datalad`, which is outside this session's authorized repo scope (`con/annextube`, `datalad/datalad` only), so the git proxy refuses the push (403). Happy to apply it if the branch moves to `datalad/datalad`, or feel free to take the diff above directly.

https://claude.ai/code/session_01J2S6eGdZyn3X7AmawfhJ4s

---

yarikoptic-gitmate:
**Update**: #7912 (the AppVeyor-outage issue this PR's red `continuous-integration/appveyor/{pr,branch}` checks were tracked against) is now closed — its fix, #7913, merged into `maint` today (`2bc08a64f`).

This PR's branch (`bf-rerun-since-ancestry-leak`) lives on `yarikoptic/datalad`, outside this session's write access, so I can't merge `maint` into it myself the way I just did for #7918 (same underlying failure). Merging current `origin/maint` into this branch and pushing should turn the AppVeyor check green without any other change needed — no conflicts are expected, since #7913 only touches test-gating/runner-decode files this PR doesn't.

---

PR #7918, opened 2026-09-05
NF: push: add -u/--set-upstream to mimic `git push -u`

yarikoptic-gitmate:
## Overview

`datalad push --to REMOTE` did not offer a way to configure `REMOTE` as
the tracking (upstream) remote for the active branch -- users had to
follow up with a plain `git push -u REMOTE branch` or
`git branch --set-upstream-to` after the fact.

This adds a `-u`/`--set-upstream` flag to `push` that mimics
`git push -u`: given together with `--to SIBLING`, it configures
`SIBLING` as the upstream remote for the active branch as part of the
same push. Without `--to`, `--set-upstream` errors out immediately with
an actionable message (`--set-upstream (-u) requires the target sibling
to be given via --to`), since there'd otherwise be no unambiguous
target to track.

Fixes #7917

## Implementation notes

- The active branch (and, for annex repos, the `git-annex` branch) are
  already unconditionally included in every push by `push`'s existing
  refspec-selection logic. After that regular push succeeds,
  `--set-upstream` triggers one more, explicit push of just the active
  branch's refspec with git's own `--set-upstream` flag. Since the
  branch was already pushed above, this second push is a cheap
  "everything up-to-date" negotiation -- it does not transfer anything,
  it only has git itself record the tracking configuration, the same
  way `git push -u` does when the branch happens to already be current.
- This reuses `GitRepo.push()`/`_push_refspecs()` (refspec push +
  result/status handling) instead of hand-rolling the
  `branch.<name>.remote`/`branch.<name>.merge` config writes, and keeps
  the tracking change scoped to the active branch only -- so internal
  branches that ride along in the same push (notably `git-annex`) are
  not also marked as tracking the target sibling.
- `repo.config.reload()` is called afterwards so the in-process config
  cache picks up what the `git push` subprocess just wrote directly to
  `.git/config`, following the same pattern already used elsewhere in
  `GitRepo`/`AnnexRepo` after config-affecting git subprocess calls
  (e.g. `add_remote`).

## Testing

- Added `test_push_set_upstream` (`datalad/core/distributed/tests/test_push.py`):
  verifies no tracking branch is configured beforehand, that
  `ds.push(to='target', set_upstream=True)` configures it, and that a
  subsequent bare `ds.push()` (relying on the now-configured tracking
  branch) succeeds.
- Added a case to `test_invalid_call` asserting `ds.push(set_upstream=True)`
  (without `--to`) raises `ValueError`.
[...]

PR #7919, opened 2026-09-06
NF: push: add -u/--set-upstream option to configure branch tracking

### LINKED ISSUE #7917: Add --set-upstream|-u to mimic git push
It is a frequent usecase to set it up, and now requires running some siblings command or git push -u first. annoying. I think we should just add similar option to push, so could be used with `--to REMOTE` (error otherwise)

yarikoptic-gitmate:
## Summary

Adds a `-u`/`--set-upstream` flag to `datalad push`, mirroring `git push
-u`. When given, the active branch is configured to track the target
sibling's branch (`branch.<name>.remote`/`.merge`), so subsequent `git
push`/`git pull` (and plain `datalad push`) know where to go without an
explicit `--to`.

Fixes #7917.

## Why

Setting up a push target currently requires either running `git push -u`
directly once, or a separate `datalad siblings` invocation, before `datalad
push` can be used without `--to`. This adds the option directly to `push`,
as suggested in the issue.

## Design notes

- `--set-upstream` requires an explicit `--to SIBLING` (errors out
  otherwise), since the branch to track would otherwise be ambiguous.
- `-u`/`--set-upstream` is handled by a small, independent follow-up step
  (`_set_upstream_for_dataset`) that runs after the main per-dataset push,
  doing a dedicated `git push --set-upstream` naming only the active
  branch's refspec. Pushing an already-just-pushed refspec with
  `--set-upstream` is a safe, idempotent no-op ("everything up-to-date")
  that still gets git to write the tracking config.
- This is deliberately *not* threaded through `_push()`/`_push_refspecs()`
  as an extra parameter: CI on this PR (`test (datalad-next)`) caught that
  doing so breaks the `datalad-next` extension, which fully replaces both
  functions with its own copies (`datalad_next.patches.push_optimize`).
  Keeping their signatures untouched avoids that, and as a side effect
  keeps `-u` from also configuring tracking for the `git-annex` branch that
  gets pushed alongside the active one in annex repos (git's own
  `--set-upstream` applies to *every* refspec in a single invocation, not
  just one).
- `GitRepo.push()` now reloads config after a push that used
  `-u`/`--set-upstream`, matching the existing convention in `checkout()`
  of reloading after git operations known to (re)write branch config —
  otherwise a subsequent `get_tracking_branch()` call in the same process
  would see stale, cached config.

## Testing

- `pytest datalad/core/distributed/tests/test_push.py -q` → 20 passed, 1
  skipped
- `pytest datalad/support/tests/test_gitrepo.py -k push -q` → 2 passed, 1
  skipped
- New tests cover: the `--to`-required validation error, tracking being set
  correctly for the active branch, and — since the default `create()` is
  annex-enabled — that the `git-annex` branch pushed alongside it does
  *not* get tracking config.
- Verified against `datalad-next` 1.6.0 locally (the extension CI runs
  against this PR): `pytest --pyargs datalad_next.patches.tests.test_push
  datalad_next.annexremotes.tests.test_uncurl::test_uncurl_ria_access
  datalad_next.patches.tests.test_patched_ria_ora -q` → all passing after
  the redesign above (they were failing with `TypeError: _push() got an
  unexpected keyword argument 'set_upstream'` before it).

This PR went through two rounds of independent review before being opened
(the git-annex-branch tracking leak was caught and fixed as a result), and
one more fix afterward in response to the `datalad-next` CI failure above.

## PR checklist

- [x] Overview of changes and rationale: see above.
- [x] Changelog snippet: `changelog.d/pr-7919.md`.
- [x] `Fixes #7917`.
- [x] Not a draft; test plan included.

---
Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

---
_Generated by [Claude Code](https://claude.ai/code/session_016bLZsKVnaYpVGX9oRNFiDn)_

### DIFF
--- a/changelog.d/pr-7919.md
+++ b/changelog.d/pr-7919.md
@@ -1,4 +1,4 @@
-### 💫 Enhancements and new features
+### 🚀 Enhancements and New Features
 
 - `push` now supports `-u`/`--set-upstream` to configure the pushed branch
   to track the target sibling, mirroring `git push -u`.
--- a/changelog.d/pr-7919.md
+++ b/changelog.d/pr-7919.md
@@ -1,0 +1,6 @@
+### 💫 Enhancements and new features
+
+- `push` now supports `-u`/`--set-upstream` to configure the pushed branch
+  to track the target sibling, mirroring `git push -u`.
+  [PR #7919](https://github.com/datalad/datalad/pull/7919)
+  (by [@yarikoptic-gitmate](https://github.com/yarikoptic-gitmate))
--- a/datalad/core/distributed/push.py
+++ b/datalad/core/distributed/push.py
@@ -266,8 +266,14 @@ class Push(Interface):
             pbars = {}
             yield from _push(
                 dspath, dsrecords, to, data, force, jobs, res_kwargs.copy(), pbars,
-                got_path_arg=True if path else False,
-                set_upstream=set_upstream)
+                got_path_arg=True if path else False)
+            if set_upstream:
+                # kept as an independent, follow-up step (rather than
+                # threaded through `_push()`/`_push_refspecs()`) so as to
+                # not alter the signature of functions that at least one
+                # widely-used extension (datalad-next) fully replaces --
+                # see gh-7917 discussion.
+                yield from _set_upstream_for_dataset(dspath, to)
             # take down progress bars for this dataset
             for i, ds in pbars.items():
                 log_progress(lgr.info, i, 'Finished push of %s', ds)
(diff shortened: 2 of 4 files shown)


yarikoptic-gitmate:
Closing as a duplicate: #7918 already implements the same fix for #7917 (opened earlier, independently reviewed, and already green across all 33 GitHub Actions jobs — same AppVeyor infra pending/flake as elsewhere, already noted there). This PR's own review process converged on essentially the same design (an independent follow-up `git push --set-upstream` for just the active branch, rather than threading a new parameter through `_push()`/`_push_refspecs()`, to avoid breaking extensions like datalad-next that fully replace those functions).

Apologies for the duplicate work — should have checked for existing open PRs against #7917 before starting. Follow-up work on this feature should happen on #7918.

---

